PR #7676, opened 2023-02-13
Avoid installing amazon_kclpy>=2.1.0

dominikschubert:
A new version was published recently ([2.1.1](https://pypi.org/project/amazon-kclpy/#history)).

The tests previously worked fine, but let's see if this will make it more stable again :thinking:


thrau (approved):
thanks for jumping on this! LGTM!

PR #7677, opened 2023-02-13
Fix CFn Route53 Record Set TTL

pinzon:
The TTL should be accepted either as a string or int.

Changes:
- Convert attribute to integer if it's a string
- Altered one test to validate the behavior
- mark the test as AWS validated


dominikschubert (approved):
LGTM :+1:

giograno (approved):

Issue #7665, opened 2023-02-11
bug: Wrong Internet Gateway Ref in CloudFormation stack


hrishi-hugo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We have two different CF stacks in which we create a VPC and IGW in each of them.
First we deploy the first stack which creates a VPCa, creates a IGW and attaches the IGW to the VPCa.
Now when we create the second stack, in this stack also we create a VPCb, create a IGW and attach to the VPCb.
But when I am creating the second stack I get Internet Gateway already associated error.
The second stack creates a new internet gateway but in the attachment step it uses the gateway created in the previous stack (which causes the already associated error).
This error happens for all the subsequent stacks created after the first stack.

This is the CF snippet 

``` 
InternetGatewaySvc:
    Type: 'AWS::EC2::InternetGateway'

  VPCSvcGatewayAttachment:
    Type: 'AWS::EC2::VPCGatewayAttachment'
    DependsOn: [InternetGatewaySvc]
    Properties:
      InternetGatewayId: !GetAtt 'InternetGatewaySvc.InternetGatewayId'
      VpcId: !Ref VPCSvc
```

The `!GetAtt 'InternetGatewaySvc.InternetGatewayId'` in the above snippet returns the id of the gateway created in the first stack instead of the one created in the second stack. (same happens with !Ref also) 

The following error is raised
```Running action "create" for resource type "AWS::EC2::VPCGatewayAttachment" id "VPCSvcGatewayAttachment"
2023-02-11T20:30:54.321 DEBUG --- [uncthread367] l.s.c.e.template_deployer  : Error applying changes for CloudFormation stack "hcservices": An error occurred (Resource.AlreadyAssociated) when calling the AttachInternetGateway operation: Resource igw-4eea8dbb is already associated. Traceback (most recent call last):
  File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1304, in _run
    self.do_apply_changes_in_loop(changes, stack)
  File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1381, in do_apply_changes_in_loop
    self.apply_change(change, stack=stack)
  File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1461, in apply_change
    result = execute_resource_action(resource_id, self, ACTION_CREATE)
[...]

PR #7679, opened 2023-02-13
fix internet gateway attrs

pinzon:
The Ref of the resource model is fixed now. This PR addresses #7665.

Changes:
-  Resource model implements the resource handler to store the InternetGatewayId as PhysicalResourceId
-  The resource model now uses the InternetGatewayId to fetch the state
- Test to validate the changes.

### DIFF
--- a/tests/integration/cloudformation/resources/test_ec2.py
+++ b/tests/integration/cloudformation/resources/test_ec2.py
@@ -61,3 +61,22 @@ def test_cfn_with_multiple_route_table_associations(ec2_client, deploy_cfn_templ
     assert subnet1["EnableDns64"] is True
     assert subnet1["MapPublicIpOnLaunch"] is True
     assert subnet2["PrivateDnsNameOptionsOnLaunch"]["HostnameType"] == "ip-name"
+
+
+def test_internet_gateway_ref_and_attr(deploy_cfn_template, cfn_client, snapshot):
+
+    stack = deploy_cfn_template(
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../templates/internet_gateway.yml"
+        )
+    )
+
+    response = cfn_client.describe_stack_resource(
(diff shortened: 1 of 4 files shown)


tests/integration/cloudformation/resources/test_ec2.py
```diff
@@ -61,3 +61,22 @@ def test_cfn_with_multiple_route_table_associations(ec2_client, deploy_cfn_templ
     assert subnet1["EnableDns64"] is True
     assert subnet1["MapPublicIpOnLaunch"] is True
     assert subnet2["PrivateDnsNameOptionsOnLaunch"]["HostnameType"] == "ip-name"
+
+
+@pytest.mark.skip_snapshot_verify(paths=["$..DriftInformation", "$..Metadata"])
+def test_internet_gateway_ref_and_attr(deploy_cfn_template, cfn_client, snapshot):
```
dominikschubert, line 67: Since we're using snapshots here I guess this was also executed against AWS and can be marked as such? (`aws_validated`)

tests/integration/templates/internet_gateway.yml
```diff
@@ -0,0 +1,21 @@
...
+    Type: AWS::EC2::VPCGatewayAttachment
+    Properties:
+      VpcId:
+        Ref: Vpc
+      InternetGatewayId:
+        !GetAtt [Gateway,InternetGatewayId]
+Outputs:
+  RefAttachment:
+    Value:
+      Ref: Gateway
```
dominikschubert, line 18: Something interesting to note: AWS claims this returns the "Name" which makes no sense for an `AWS::EC2::InternetGateway` :sweat_smile:

dominikschubert (approved):
LGTM, small nit regarding test marker

Kudos for adding both ref and GetAtt validation in the test :+1:

PR #7682, opened 2023-02-14
fix encoding issues in router, proxy, partition rewriting

alexrashed:
This PR addresses several encoding issues:
- `get_raw_base_url` now avoids multi-encoding by avoiding the usage of `werkzeug.sansio.utils.get_current_url`.
- `SimpleRequestsClient` now avoids multi-encoding by avoiding the usage of `werkzeug.sansio.utils.get_current_url`.
- `Proxy` uses the encoded path of the given request (instead of the decoded one) if no `forward_path` is given.
- `Router` now matches on the encoded path (instead of the decoded one).
  - This is a quite fundamental change, but it is necessary, because otherwise the encoding information is lost for parsed path parameters (f.e. when using `/<path:path>`).
- `ArnPartitionRewriteHandler` now uses the correct forward path (the modified one, instead of the original one).
- `ArnPartitionRewriteHandler` does not decode string values anymore. It uses the encoding-specific regex if necessary anyways.

/cc @whummer

### DIFF
--- a/localstack/aws/handlers/partition_rewriter.py
+++ b/localstack/aws/handlers/partition_rewriter.py
@@ -2,7 +2,7 @@ import logging
 import re
 from re import Match
 from typing import Optional
-from urllib.parse import unquote, urlparse
+from urllib.parse import urlparse
 
 from localstack import config
 from localstack.aws.api import RequestContext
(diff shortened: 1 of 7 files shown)


thrau (approved):
LGTM! i think all we need to document somewhere is that `@route` and `Router.add` behave differently from werkzeug/flask now, in that we have to specify URL encoded paths now.
so `@route("/foo bar/baz")` has to become `@route("/foo%20bar/baz")`, if i understood correctly.

PR #7687, opened 2023-02-14
Change behavior of VtlTemplate to modify variable values in-place

whummer:
Change behavior of the `VtlTemplate` utility class to modify variable values in-place. This PR is a prerequisite for an upstream PR that enhances support for AppSync pipeline resolvers.

### DIFF
--- a/localstack/utils/aws/templating.py
+++ b/localstack/utils/aws/templating.py
@@ -1,12 +1,12 @@
 import contextlib
-import copy
 import json
 import re
 from typing import Any, Dict
 
 import airspeed
 
-from localstack.utils.objects import recurse_object
+from localstack.utils.collections import DictProxy
+from localstack.utils.objects import Mock, recurse_object
 from localstack.utils.patch import patch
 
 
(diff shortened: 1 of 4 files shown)


localstack/utils/aws/templating.py
```diff
@@ -67,9 +85,8 @@ def apply(obj, **_):
                         obj[k] = ExtendedString(v)
             return obj
 
-        # loop through the variables and enable certain additional util
-        # functions (e.g., string utils)
-        variables = copy.deepcopy(variables or {})
+        # loop through the variables and enable certain additional util functions (e.g., string utils)
+        variables = {} if variables is None else variables
```
silv-io, line 89: Is copying not necessary here? What side effects does removing `deepcopy` cause here?

localstack/utils/collections.py
```diff
@@ -109,6 +112,70 @@ def __hash__(self):
         return hash(canonical_json(self._dict))
 
 
+class DictProxy(MutableMapping):
+    """
+    Generic dict proxy class that can be used to define subclasses of dict with additional methods.
+    It implements the MutableMapping interface and redirects all methods calls to an underlying dict which it wraps.
+
+    The DictProxy class can be extended with additional custom logic as follows:
```
silv-io, line 120: Afaik the regular built-in `dict` can already do this. Is there some extra reason for this apart from adding new methods?

silv-io (commented):
LGTM, but some comments regarding dicts and deepcopy.

localstack/utils/aws/templating.py
whummer: That's a great catch. This change is required to allow in-place modification of elements in VTL.

For example, in AppSync we have mutable elements like the `stash` which we want to be able to modify like this in resolver mapping templates:
```
$ctx.stash.put("email", $ctx.args.input.email)
```

The ability to do in-place modification is covered by a small unit test in this PR (`test_put_value_to_dict(..)`)

localstack/utils/collections.py
whummer: Indeed, the built-in airspeed dict already supports `put`, nice catch. This was more to anticipate other additional map methods like `putAll`, `entrySet`, etc. 

Just realized that it may be cleaner to patch `__additional_methods__` in airspeed directly for those cases. Will remove the DictProxy again for now, we can revert to it later once a concrete use case arises.

silv-io (approved):
LGTM!

Issue #7567, opened 2023-01-27
EventBridge PutTargets does not check the syntax of target IDs.


wunderbarb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

For instance providing a target ID such as the ARN of a SQS queue is accepted by localstack.  In real, the ID must follow regex [a-z0-9.-_]+ and be less than 64 characters.

### Expected Behavior

Issue an error message

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack-pro

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```golang
...
pti := &eventbridge.PutTargetsInput{
		Rule: aws.String(ruleName),
		Targets: []types2.Target{
			{
				Arn: aws.String(q.GetARN()),
				Id:  aws.String(ruleName + q.GetARN()),
			},
		},
		EventBusName: nil,
	}
	_, err = c.PutTargets(context.Background(), pti)
```

### Environment

```markdown
- OS: Ubuntu 20.04
- LocalStack: localstack-pro: latest
```

Datashri:
please give me access to this bug.
I have an idea to solve this.

PR #7688, opened 2023-02-14
fix event target id validation

pinzon:
This PR adds the validation for the Targets Ids in the Put Target Operation. Addresses #7567

Changes:
- validation
- tests

### DIFF
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -29,6 +29,7 @@ from localstack.aws.api.events import (
     RuleState,
     ScheduleExpression,
     TagList,
+    TargetList,
 )
 from localstack.constants import APPLICATION_AMZ_JSON_1_1
 from localstack.services.events.models import EventsStore, events_stores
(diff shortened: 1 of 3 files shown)


tests/integration/test_events.py
```diff
@@ -1596,3 +1596,22 @@ def cleanup(
                     logGroupName=log_group_name, logStreamName=log_stream["logStreamName"]
                 )
             logs_client.delete_log_group(logGroupName=log_group_name)
+
+    def test_put_target_id_validation(self, sqs_create_queue, events_client, sqs_client, snapshot):
```
dominikschubert, line 1600:
```suggestion
    @pytest.mark.aws_validated
    def test_put_target_id_validation(self, sqs_create_queue, events_client, sqs_client, snapshot):
```

I guess this was executed against AWS since it has a snapshot right?

tests/integration/test_events.py
```diff
@@ -1596,3 +1596,22 @@ def cleanup(
...
+        events_client.put_rule(Name=rule_name, EventPattern=json.dumps(TEST_EVENT_PATTERN))
+
+        with pytest.raises(ClientError) as e:
+            events_client.put_targets(
+                Rule=rule_name,
+                EventBusName=TEST_EVENT_BUS_NAME,
+                Targets=[
+                    {"Id": target_id, "Arn": queue_arn, "InputPath": "$.detail"},
+                ],
+            )
```
dominikschubert, line 1615: I'd suggest at least adding one explicitly validated positive case here as well that contains more or less all possible.

localstack/services/events/provider.py
```diff
@@ -199,6 +200,30 @@ def create_connection(
 
         return call_moto(context)
 
+    def put_targets(
```
dominikschubert, line 203: The original issue also mentioned an upper limit of 64 characters. Did you look into this as well? :thinking:

dominikschubert (changes requested):
Thanks for churning through these issues :pray: 

Some concerns regarding validation & test cases, then good to go :+1:

dominikschubert (approved):
LGTM! :rocket:

PR #7690, opened 2023-02-15
Allow configuration of docker image used for port checks

### LINKED ISSUE #7681: bug: PORTS_CHECK_DOCKER_IMAGE is not configurable which prevents ECS tasks from running on air-gapped environments

### Current Behavior

`PORTS_CHECK_DOCKER_IMAGE` is not configurable which prevents ECS tasks from running on air-gapped environment

### Expected Behavior

`PORTS_CHECK_DOCKER_IMAGE` can be configured via configuration option to allow use of any available image 

With a docker-compose file

### Steps To Reproduce

Run an ECS task in an environment where localstack/localstack is not available.

### Environment

```markdown
- OS: RHEL8
- Docker runtime version: Podman v4.2.0
- LocalStack: 1.3.1
```

### Anything else?

_No response_

dominikschubert:
Closes #7681

## Changes

Now allows overriding the default image that is used internally in ECS to verify we can open ports with a docker container. Previously this was statically set to `localstack/localstack` and some users might not have access to that image in their environment.

## Discussion

I've noticed we are also using the DOCKER_IMAGE_NAME (`localstack/localstack`) in other places as well.
Should we maybe make these constants configurable as well? :thinking:

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -16,6 +16,7 @@ from localstack.constants import (
     DEFAULT_PORT_EDGE,
     DEFAULT_SERVICE_PORTS,
     DEFAULT_VOLUME_DIR,
+    DOCKER_IMAGE_NAME,
     ENV_INTERNAL_TEST_COLLECT_METRIC,
     ENV_INTERNAL_TEST_RUN,
     FALSE_STRINGS,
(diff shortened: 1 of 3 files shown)


whummer (approved):
Nice catch, LGTM! 👍

PR #7693, opened 2023-02-15
refactor community ext imports

alexrashed:
This PR tries to decrease the coupling to our LocalStack Pro extensions.
It can only merged together with its companion PR.

The changes are separated in different commits:
- _remove DNS startup from community:_
  - This commit removes the DNS startup from the `edge.py`. It will be started as a Pro plugin (infra start hook) instead.
    /cc @dfangl 
- _refactor SSL cert handling:_
  - Moves the SSL cert handling to community, since it is used and supported by community since the HTTP/HTTPS multiplexing is used (https://github.com/localstack/localstack/pull/1823).
    /cc @whummer 
- _unify, move, document ext imports:_
  - Documents, moves, and unifies remaining (for now tolerated) ext imports.
- _migrate serve_flask_app to only remaining using module:_
  - A somewhat unrelated refactoring, moving a legacy function from the generic proxy to its only using module (a legacy service provider).
    /cc @dominikschubert

### DIFF
--- a/localstack/aws/app.py
+++ b/localstack/aws/app.py
@@ -2,6 +2,7 @@ from localstack.aws import handlers
 from localstack.aws.handlers.metric_handler import MetricHandler
 from localstack.aws.handlers.service_plugin import ServiceLoader
 from localstack.services.plugins import SERVICE_PLUGINS, ServiceManager, ServicePluginManager
+from localstack.utils.ssl import create_ssl_cert, install_predefined_cert_if_available
 
 from .gateway import Gateway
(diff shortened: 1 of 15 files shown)


dominikschubert (commented):
Just looked over the part for which you tagged me (lambda flask refactoring). Seems innocent enough with the move :+1: 

I'll leave the rest for @thrau though to approve.

thrau (approved):
looks great! thanks for untangling all this, will make it much easier to remove dead code.

Issue #7548, opened 2023-01-25
bug: Getting error when creating lambda function with dynamodb as event stream


ayeganyan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Getting error when trying to attach dynamodb to lambda as an event source:
> 2023-01-19T15:20:52.121 ERROR --- [functhread28] l.s.a.e.stream_event_sourc : 'StartingPosition'
> Traceback (most recent call last):
>   File "/opt/code/localstack/localstack/services/awslambda/event_source_listeners/stream_event_source_listener.py", line 345, in _monitor_stream_event_sources
>     source["StartingPosition"],
> KeyError: 'StartingPosition'

This worked fine on localstack:0.14.5

### How are you starting LocalStack?

With a `docker run` command

### Steps To Reproduce

* Run localstack:
> docker run --rm -d -v /Users/ayeganyan/workspace/sandbox/lambda-nodejs/artifacts:/artifacts -p 4566 -e PROVIDER_OVERRIDE_LAMBDA=asf -v "/var/run/docker.sock:/var/run/docker.sock:rw" localstack/localstack

* ssh to running localstack

* Executes the following commands, which basically creates lambda (provide anything as code and runtime), and dynamodb
> awslocal dynamodb create-table --table-name dx-ref-java-quarkus-im-lambda-ddb --attribute-definitions AttributeName=key,AttributeType=S --key-schema AttributeName=key,KeyType=HASH --billing-mode PAY_PER_REQUEST --stream-specification StreamEnabled=true,StreamViewType=NEW_AND_OLD_IMAGES
> export DYNAMO_DX_REF_JAVA_QUARKUS_IM_LAMBDA_DDB_STREAM_ARN=$(awslocal dynamodb describe-table --table-name dx-ref-java-quarkus-im-lambda-ddb --query 'Table.LatestStreamArn' --output text)
> awslocal lambda create-function --function-name ddblambda --role arn:aws:iam::000000000000:role/DummyRole --runtime python3.8 --handler python.main.handle --zip-file fileb:///artifacts/python.zip --memory-size 1024 --timeout 20 --environment '{"Variables": {"OUT_QUEUE_NAME": "dx-ref-java-quarkus-im-lambda-ddb-out"}}'

* Add event source:
[...]

PR #7696, opened 2023-02-15
validate StartingPosition during CreateEventSourceMapping operation

pinzon:
This PR makes the **CreateEventSourceMapping** operation validate the **StartingPosition** parameter. Addresses #7548

Changes:
- validation
- snapshoted test

### DIFF
--- a/localstack/services/awslambda/provider.py
+++ b/localstack/services/awslambda/provider.py
@@ -181,6 +181,7 @@ from localstack.services.awslambda.urlrouter import FunctionUrlRouter
 from localstack.services.edge import ROUTER
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.aws import aws_stack
+from localstack.utils.aws.arns import extract_service_from_arn
 from localstack.utils.collections import PaginatedList
 from localstack.utils.files import load_file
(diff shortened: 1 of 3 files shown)


localstack/services/awslambda/provider.py
```diff
@@ -1346,6 +1347,13 @@ def create_event_source_mapping(
         if "EventSourceArn" not in request:
             raise InvalidParameterValueException("Unrecognized event source.", Type="User")
 
+        service = extract_service_from_arn(request.get("EventSourceArn"))
+        if "StartingPosition" not in request and service in ["dynamodb", "kinesis", "kafka"]:
```
dominikschubert, line 1351:
```suggestion
        if service in ["dynamodb", "kinesis", "kafka"] and "StartingPosition" not in request:
```

nit: IMO, it's a bit more natural to read this way :thinking: 



Also, I think this should also be evaluated in update_event_source_mapping.

We'll need to extend these validations, especially for event source mappings at some point even further, since there's a lot of differences in which sets of values are valid depending on the source service.

dominikschubert (changes requested):
Mostly LGTM, but would be great if we could add the same validation in the update operation :pray:

dominikschubert (approved):
LGTM :rocket:

PR #7700, opened 2023-02-16
Add support for route.destination-cidr-block filter in DescribeRouteTables

dominikschubert:
Adds support for the `route.destination-cidr-block` filter in `EC2.DescribeRouteTables`

See: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html

Can be removed once https://github.com/getmoto/moto/pull/5932/files is merged upstream and we have updated our fork.


whummer (approved):
LGTM, thanks for the quick fix @dominikschubert! (based on @viren-nadkarni 's upstream PR) 🚀

Issue #7645, opened 2023-02-08
bug: SQS:  localstack accepts queue names that are not syntactically correct.


wunderbarb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Localstack accepts queue name `test.example.com`  

The regex for queue name is [a-zA-Z0-9_\-]+$

### Expected Behavior

It should only accept valid queue names [a-zA-Z0-9_\-]+$ with less than 80 characters.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS: Ubuntu 20.04
- LocalStack: pro latest
```

wunderbarb:
I am not anymore able to reproduce it.  Mea culpa.  Sorry

PR #7701, opened 2023-02-16
#7645 correct missing constraints on queue name and acceptable charac…

rjmcginness:
fix: correct missing constraints on queue name and acceptable characters

#7645

### DIFF
--- a/localstack/services/cloudformation/api_utils.py
+++ b/localstack/services/cloudformation/api_utils.py
@@ -5,7 +5,6 @@ from urllib.parse import urlparse
 from requests.structures import CaseInsensitiveDict
 
 from localstack import config, constants
-from localstack.services.cloudformation.engine.template_preparer import transform_template
 from localstack.services.s3 import s3_listener, s3_utils
 from localstack.utils.aws import aws_stack
 from localstack.utils.functions import run_safe
@@ -25,8 +24,7 @@ def prepare_template_body(req_data: dict) -> str | bytes | None:  # TODO: mutati
         if modified_template_body:
             req_data.pop("TemplateURL", None)
             req_data["TemplateBody"] = modified_template_body
-    body = get_template_body(req_data)
-    modified_template_body = transform_template(body)
+    modified_template_body = get_template_body(req_data)
     if modified_template_body:
         req_data["TemplateBody"] = modified_template_body
     return modified_template_body
--- a/localstack/services/cloudformation/engine/entities.py
+++ b/localstack/services/cloudformation/engine/entities.py
@@ -142,7 +142,9 @@ class Stack:
     def set_time_attribute(self, attribute, new_time=None):
         self.metadata[attribute] = new_time or timestamp_millis()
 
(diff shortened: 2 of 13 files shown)


tests/integration/test_sqs.py
```diff
@@ -136,6 +137,20 @@ def test_create_queue_and_get_attributes(self, sqs_client, sqs_queue):
         assert int(float(attrs["CreatedTimestamp"])) == pytest.approx(int(time.time()), 30)
         assert int(attrs["VisibilityTimeout"]) == 30, "visibility timeout is not the default value"
 
+    @pytest.mark.only_localstack
+    def test_queue_name_too_long(self, sqs_create_queue, snapshot):
+        with pytest.raises(InvalidParameterValue) as e:
+            sqs_create_queue(QueueName=(81 * "a"))
+
+        snapshot.match("queue_name_invalid_character", e.value.response)
```
thrau, line 145: this seems like something we could test with snapshotting. here are a couple of examples. if you don't have an AWS account you want to test it with, then i'm happy to help with tests. https://github.com/localstack/localstack/blob/27b4e328cd2fbc3c7d3a11ee5d88c771da641d08/tests/integration/test_sqs.py#L1408-L1422

localstack/services/sqs/provider.py
```diff
@@ -669,6 +671,12 @@ def create_queue(
 
         return CreateQueueResult(QueueUrl=queue.url(context))
 
+    def _verify_valid_queue_name(self, queue_name: str):
+        name_re = re.compile(sqs_constants.QUEUE_NAME_REGEX + "{" + f"{len(queue_name)}" + "}")
+        if not queue_name or len(queue_name) > 80 or name_re.match(queue_name) is None:
+            error_msg = "Can only include alphanumeric characters, hyphens, or underscores. 1 to 80 in length"
+            raise InvalidParameterValue(error_msg)
```
thrau, line 678: we actually already have a method for this: https://github.com/localstack/localstack/blob/44ff8c78f3864c694dc52a44e0bfad032105be0f/localstack/services/sqs/provider.py#L105-L119

thrau (changes requested):
thanks for the contribution @rjmcginness 

this one's a headscratcher to me because we already have a method for asserting the queue name, but it's apparently not in use. we do also have assert methods in the sqs queue model itself, which *are* used, so i'm not sure why these exceptions wouldn't show up https://github.com/localstack/localstack/blob/27b4e328cd2fbc3c7d3a11ee5d88c771da641d08/localstack/services/sqs/models.py#L462-L466

verifying the queue name directly in `create_queue` makes sense to me, but we could maybe re-use the method we already have (linked in it a comment). and then remove the validation from the `SqsQueue` class accordingly. i just want to avoid having the same code in three different ways.

and it would be great to add snapshot validated tests. if that's too much of a bother for you, i'm happy to help.

rjmcginness:
@thrau Thank you, Thomas. Yes, makes perfect sense. I can try to do this, though it might not be quick. I will try to investigate why the other method is not working and use that in the place I added the new one.

PR #7675, opened 2023-02-13
Fix docker runtime executor flags

joe4dev:
This PR adds support for additional Docker flags and fixes ARM issues with lambda.

## Changes

* Bump docker dependency version (required for supporting platform flag)
* Add user and platform flags to docker clients (SdkDockerClient and CmdDockerClient). Set platform flag based on the architecture of the lambda function
* Test ARM create function exceptions and enable ARM functions
* Respect new provider override flags (`v2`) in lambda tests

## Addresses

* https://github.com/localstack/localstack/issues/7637
  * Resolves this issue by adding basic support for ARM Lambdas (more testing planned)
* https://github.com/localstack/localstack/pull/6724#issuecomment-1227314083
  * Enables workaround `LAMBDA_DOCKER_FLAGS=--user nobody`

## Related Issues

* https://github.com/localstack/localstack/issues/4921
* Shared ARM layers: https://github.com/localstack/localstack/issues/4932

## Follow Up

- [ ] Re-enable the skipped test `tests.bootstrap.test_cli.TestCliContainerLifecycle.test_start_cli_within_container` as soon as the new localstack image is on DockerHub (see [comment here](tests.bootstrap.test_cli.TestCliContainerLifecycle.test_start_cli_within_container))

PR #7702, opened 2023-02-16
Re-enable cli container lifecycle test

joe4dev:
Re-enables a Docker CLI test that depends on the not-immediately-updated `localstack/localstack` image from Dockerhub.

Follow up from https://github.com/localstack/localstack/pull/7675

## Background

`tests.bootstrap.test_cli.TestCliContainerLifecycle.test_start_cli_within_container` depends on the latest [localstack/localstack](https://github.com/localstack/localstack/blob/ce0f939c228448eb0b4aedcd7ea6110c132243e2/tests/bootstrap/test_cli.py#L185) image from DockerHub. The test could not pass because bumping the PyPi `docker` version added an incompatibility. Specifically, a `platform` flag required for fixing ARM compatibility in lambda.

See internal [demo](https://www.notion.so/localstack/Releasing-a-change-with-conflicting-circular-dependency-8ca299173c5140bdb766b0a0512bd2d3?pvs=4)

### DIFF
--- a/tests/bootstrap/test_cli.py
+++ b/tests/bootstrap/test_cli.py
@@ -180,7 +180,6 @@ class TestCliContainerLifecycle:
         output = container_client.exec_in_container(config.MAIN_CONTAINER_NAME, ["ps", "-u", user])
         assert "supervisord" in to_str(output[0])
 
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
LGTM :+1:

PR #7704, opened 2023-02-16
fix S3 GetObjectAttributes and GetObject to return Checksum if requested

bentsku:
This PR implements missing behaviour in S3 regarding checksums. It's not implemented yet in moto and a TODO is written about it:

> TODO: AWS computes the provided value and verifies it's the same. Afterwards, it should be returned in every subsequent call

I've implemented the behaviour inside the ASF provider, with multiple AWS validated and snapshotted tests. 
As `moto` does not store the computed or given hash in the key, for minimal changes we can compute the hash when it's requested (aka not often). This PR does not implement additional checksum validation (we have some on `PutObject`), but simply return the object checksum where it was missing.

For multipart uploads, I added some AWS validated tests, but we need additional behaviour for this, and it's out of scope of this PR.

Fixes:

- [x] Add `Checksum` field to `GetObjectAttributes` if the object was created with a checksum
- [x] Add `Checksum` field to `GetObject` if the object was created with a checksum and `ChecksumMode` is `"ENABLED"`
- [x] Investigate `Content-Encoding: ''` from AWS response and copy behaviour
- [x] Set a hardcoded timestamp when creating `gzip` files so that tests are reproducible
- [x] Create tests for checksum for multipart uploads (out of scope of this PR to add functionality, would be best in moto)

See:
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#mpuchecksums
\cc @whummer


macnev2013 (approved):
LGTM 🚀

PR #7705, opened 2023-02-17
added terraform test case for dms: rds - kinesis integration

macnev2013:
## Description
This PR contains test case for migrating data from Mysql RDS to Kinesis data stream using AWS data migration service.

- Infrastructure is created using terraform script. (tried using CF as well which didn't work - It was erroring out while clearing up the infrastructure)
- After the infra creation we are asserting states of the services
  - TableStatistics in DMS
  - Data records in Kinesis


macnev2013:
Moving this PR to: https://github.com/localstack/localstack-ext/pull/1384

PR #7709, opened 2023-02-17
[SFN] [IntrinsicFunctions] base implementation Encode-Decode functions

MEPalma:
Adds base implementation for States Encode and Decode Intrinsic Functions


dominikschubert (approved):
LGTM :rocket:

PR #7711, opened 2023-02-17
Add initial support for CFn Fn:Transform intrinsic functions

whummer:
Add initial support for CFn `Fn:Transform` intrinsic functions. This requirement came up while playing around with a more complex AWS sample app using a SAM / CloudFormation template: https://workshop.serverlesscoffee.com

For reference - part of the SAM template in the sample is a construct like this:
```
  RESTApConfigService:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      DefinitionBody:
        'Fn::Transform':
          Name: 'AWS::Include'
          Parameters:
            Location: '../backends/2-config-service/RestAPIs/api.yaml'
```
(the `Location` parameter is replaced with an `s3://...` URL by SAM upon deployment)

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -1462,6 +1462,8 @@ class TemplateDeployer:
             res_change["_deployed"] = is_deployed
             if not is_deployed:
                 return True
+            if action == "Add":
+                return False
             if action == "Modify" and not self.is_updateable(resource):
                 LOG.debug(
                     'Action "update" not yet implemented for CF resource type %s',
--- a/localstack/services/cloudformation/engine/template_preparer.py
+++ b/localstack/services/cloudformation/engine/template_preparer.py
@@ -44,6 +44,10 @@ def template_to_json(template: str) -> str:
 def transform_template(template: dict, parameters: list) -> Dict:
     result = dict(template)
 
+    # apply 'Fn::Transform' intrinsic functions (note: needs to be applied before global
+    #  transforms below, as some utils - incl samtransformer - expect them to be resolved already)
+    result = apply_transform_intrinsic_functions(result)
+
     # apply global transforms
     transformations = format_transforms(result.get("Transform", []))
     for transformation in transformations:
@@ -65,9 +69,6 @@ def transform_template(template: dict, parameters: list) -> Dict:
                 macro=transformation,
                 stack_parameters=parameters,
             )
-
-    # apply 'Fn::Transform' intrinsic functions
-    result = apply_transform_intrinsic_functions(result)
 
     return result
 
--- a/localstack/services/cloudformation/engine/template_preparer.py
+++ b/localstack/services/cloudformation/engine/template_preparer.py
@@ -11,6 +11,9 @@ from localstack.aws.api import CommonServiceException
 from localstack.services.cloudformation.engine import yaml_parser
 from localstack.services.cloudformation.engine.entities import resolve_ssm_parameter_value
 from localstack.services.cloudformation.engine.policy_loader import create_policy_loader
+from localstack.services.cloudformation.engine.transformers import (
+    apply_transform_intrinsic_functions,
+)
 from localstack.services.cloudformation.stores import get_cloudformation_store
 from localstack.utils.aws import aws_stack
 from localstack.utils.json import clone_safe
@@ -37,9 +40,11 @@ def template_to_json(template: str) -> str:
     return json.dumps(template)
 
 
+# TODO: consider moving to transformers.py as well
 def transform_template(template: dict, parameters: list) -> Dict:
     result = dict(template)
 
+    # apply global transforms
     transformations = format_transforms(result.get("Transform", []))
     for transformation in transformations:
         if not isinstance(transformation["Name"], str):
@@ -60,6 +65,9 @@ def transform_template(template: dict, parameters: list) -> Dict:
                 macro=transformation,
                 stack_parameters=parameters,
             )
+
+    # apply 'Fn::Transform' intrinsic functions
+    result = apply_transform_intrinsic_functions(result)
 
     return result
 
@@ -144,7 +152,6 @@ def format_transforms(transforms: list | dict | str) -> list[dict]:
      - an object like {Name: transformation, Parameters:{}}
      - a transformation name
      - a list of objects defining a transformation
-
      so the objective of this function is to normalize the list of transformations to apply.
     """
     formatted_transformations = []
--- a/localstack/services/cloudformation/engine/template_preparer.py
+++ b/localstack/services/cloudformation/engine/template_preparer.py
@@ -41,12 +41,12 @@ def template_to_json(template: str) -> str:
 
 
 # TODO: consider moving to transformers.py as well
-def transform_template(template: dict, parameters: list) -> Dict:
+def transform_template(template: dict, parameters: list, stack=None) -> Dict:
     result = dict(template)
 
     # apply 'Fn::Transform' intrinsic functions (note: needs to be applied before global
     #  transforms below, as some utils - incl samtransformer - expect them to be resolved already)
-    result = apply_transform_intrinsic_functions(result)
+    result = apply_transform_intrinsic_functions(result, stack=stack)
 
     # apply global transforms
     transformations = format_transforms(result.get("Transform", []))
--- a/localstack/services/cloudformation/engine/transformers.py
+++ b/localstack/services/cloudformation/engine/transformers.py
@@ -31,7 +31,7 @@ class AwsIncludeTransformer(Transformer):
             content = parse_template(content)
             return content
         else:
-            LOG.warning("Unexpected ")
+            LOG.warning("Unexpected Location parameter for AWS::Include transformer: %s", location)
         return parameters
 
 
@@ -39,8 +39,9 @@ class AwsIncludeTransformer(Transformer):
 transformers: Dict[str, Type] = {"AWS::Include": AwsIncludeTransformer}
 
 
-def apply_transform_intrinsic_functions(template: dict) -> dict:
+def apply_transform_intrinsic_functions(template: dict, stack=None) -> dict:
     """Resolve constructs using the 'Fn::Transform' intrinsic function."""
+    from localstack.services.cloudformation.engine.template_deployer import resolve_refs_recursively
 
     def _visit(obj, **_):
         if isinstance(obj, dict) and obj.keys() == {"Fn::Transform"}:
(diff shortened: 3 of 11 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1431,6 +1431,8 @@ def prepare_should_deploy_change(self, resource_id, change, stack, new_resources
             res_change["_deployed"] = is_deployed
             if not is_deployed:
                 return True
+            if action == "Add":
```
whummer, line 1434: Adding this here to ensure we have proper logic and avoid re-creation of resources - a companion PR with a test will follow soon upstream in -ext.

whummer:
Test error in integration-tests-against-pro seems unrelated to these changes:

```
...
2023-02-17T21:26:54.1775824Z   File "/home/runner/work/localstack/localstack/localstack/localstack/packages/api.py", line 89, in install
2023-02-17T21:26:54.1776199Z     self._prepare_installation(target)
2023-02-17T21:26:54.1776843Z   File "/home/runner/work/localstack/localstack/localstack-ext/localstack_ext/packages/core.py", line 76, in _prepare_installation
2023-02-17T21:26:54.1777222Z     self._os_switch(
2023-02-17T21:26:54.1777812Z   File "/home/runner/work/localstack/localstack/localstack-ext/localstack_ext/packages/core.py", line 62, in _os_switch
2023-02-17T21:26:54.1778459Z     raise SystemNotSupportedException(
2023-02-17T21:26:54.1779221Z localstack.packages.core.SystemNotSupportedException: OS level packages are only installed within docker containers.
2023-02-17T21:26:54.1787826Z 2023-02-17T21:26:54.178  INFO --- [   asgi_gw_5] localstack.request.aws     : AWS redshift.CreateCluster => 500 (InternalError)
2023-02-17T21:26:54.2252216Z 2023-02-17T21:26:54.224  INFO --- [   asgi_gw_7] localstack.request.aws     : AWS dynamodb.Scan => 200
2023-02-17T21:26:54.3094671Z 2023-02-17T21:26:54.309  INFO --- [   asgi_gw_1] localstack.request.aws     : AWS redshift.CreateCluster => 400 (ClusterAlreadyExists)
2023-02-17T21:26:54.3376208Z RERUN
2023-02-17T21:26:54.3474312Z ../localstack/tests/integration/test_redshift.py::TestRedshift::test_create_clusters 
```

pinzon (approved):
LGTM :rocket:

localstack/services/cloudformation/engine/template_preparer.py
```diff
@@ -37,9 +40,15 @@ def template_to_json(template: str) -> str:
     return json.dumps(template)
 
 
+# TODO: consider moving to transformers.py as well
 def transform_template(template: dict, parameters: list) -> Dict:
     result = dict(template)
 
+    # apply 'Fn::Transform' intrinsic functions (note: needs to be applied before global
+    #  transforms below, as some utils - incl samtransformer - expect them to be resolved already)
+    result = apply_transform_intrinsic_functions(result)
```
dominikschubert, line 49: That might turn out to be a bit of a problem. As far as I understand this is actually the opposite of what CloudFormation does.

tests/integration/cloudformation/api/test_transformers.py
```diff
@@ -0,0 +1,45 @@
...
+    Resources:
+      RestApi:
+        Type: AWS::ApiGateway::RestApi
+        Properties:
+          Name: {api_name}
+          Body:
+            'Fn::Transform':
+              Name: 'AWS::Include'
+              Parameters:
+                Location: s3://{s3_bucket}/api.yaml
```
dominikschubert, line 37: nit: I'd suggest using a parameter here

tests/integration/cloudformation/api/test_transformers.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    deploy_cfn_template(template=template)
+
+    # assert REST API is created properly
+    result = apigateway_client.get_rest_apis()
+    matching = [api for api in result["items"] if api["name"] == api_name]
+    assert matching
```
dominikschubert, line 44: nit: would suggest using an Output (rest api id)

localstack/services/cloudformation/models/ec2.py
```diff
@@ -8,6 +8,8 @@
 from localstack.utils.aws import aws_stack
 from localstack.utils.strings import str_to_bool
 
+TAG_KEY_LOGICAL_ID = "aws:cloudformation:logical-id"
```
dominikschubert, line 11: Is there something that explicitly breaks on this missing? Because I'm not sure about only implementing it here when it should actually be included nearly everywhere :thinking:

localstack/services/cloudformation/engine/transformers.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _visit(obj, **_):
+        if isinstance(obj, dict) and obj.keys() == {"Fn::Transform"}:
+            transform = obj["Fn::Transform"]
+            transform_name = transform.get("Name")
+            transformer_class = transformers.get(transform_name)
+            if transformer_class:
+                transformer = transformer_class()
+                properties = transform.get("Parameters") or {}
+                return transformer.transform(properties)
```
dominikschubert, line 53:
```suggestion
                parameters = transform.get("Parameters") or {}
                return transformer.transform(parameters)
```

small nit for consistency

tests/integration/cloudformation/api/test_transformers.snapshot.json
```diff
@@ -0,0 +1,27 @@
...
+          "apiKeySource": "HEADER",
+          "endpointConfiguration": {
+            "types": [
+              "EDGE"
+            ]
+          },
+          "tags": {
+            "aws:cloudformation:logical-id": "RestApi",
+            "aws:cloudformation:stack-id": "arn:aws:cloudformation:<region>:111111111111:stack/stack-8a6438b2/51efa190-aec6-11ed-a8ce-0a8057042631",
+            "aws:cloudformation:stack-name": "stack-8a6438b2"
```
dominikschubert, line 20: These don't seem stable. I.e. when running this against AWS the validation will fail. 
The stack name & uuid would need to be handled via transformers here to avoid this. :thinking:

dominikschubert (changes requested):
Mostly small nits that don't block the merge, otherwise LGTM. I like the intrinsic function abstraction. It kinda resembles what I've seen so far in related tools and how I'd like us to approach the rest as well :+1: 

Request for change affects the recorded snapshot since it won't pass against AWS when re-running.

tests/integration/cloudformation/api/test_transformers.snapshot.json
whummer: Sounds good. I've added `@pytest.mark.skip_snapshot_verify(paths=["$..tags"])` to avoid verification of the tags, but sounds like a good idea to move it to a transformer 👍 (will update..)

dominikschubert: skip_snapshot_verify does (intentionally) not work when verifying against AWS. It should generally only be used for skipping verification between AWS & LocalStack

localstack/services/cloudformation/models/ec2.py
whummer: That's a great catch. This change actually hints at a general limitation we have in our approach (as you're well aware) - the way how we detect the resource state and the fact that we're allowing templates to be deployed on top of existing resources.

With the current implementation of our resource classes, in combination with the [change above](https://github.com/localstack/localstack/pull/7711/files#diff-95b30e7b954acaf70be05be62cb7404d8fef4ddf7538b924e12724f73f369bf0R1434-R1435), the `fetch_state` method can sometimes return false positives - i.e., report that the resource already exists when in fact it doesn't exist yet.

This issue in particular occurs if we have multiple instances of resource types whose deployment state cannot easily be determined based on the resource properties, e.g., consider the example below (we have a [similar example in our tests](https://github.com/localstack/localstack/blob/master/tests/integration/templates/template36.yaml#L15-L40)):
```
  routeTable1:
    Type: AWS::EC2::RouteTable
    Properties:
      VpcId: vpc-12345
  routeTable2:
    Type: AWS::EC2::RouteTable
    Properties:
      VpcId: vpc-12345
```

I've simplified the implementation - now the state lookup detects whether `PhysicalResourceId` has already been set (using the `result_handler` hook) and only then returns the resource details. This is quite a bit cleaner, and also should give us better parity with AWS (no additional tags). 👌

localstack/services/cloudformation/engine/template_preparer.py
whummer: Agreed, had a similar thought. We'll see how it goes over time, especially as we further increase the surface area of snapshot tests. If this is not in line with AWS, we might end up having to perform some patches in samtransformer or other transformer libraries, to be able to handle `Fn::Transform` intrinsic functions within the templates. 👍

tests/integration/cloudformation/api/test_transformers.snapshot.json
whummer: Gotcha, now I see what you're saying. Updated the transformers in the test. 👍

tests/integration/cloudformation/api/test_transformers.py
whummer: This coincidentally led to another required change - if we're using parameter refs inside the transformer, then we need to resolve the refs before fetching the file from `Location`. Nice pointer!

whummer:
Tests against pro failed with `TestRedshift::test_create_clusters` (seems unrelated to this PR), and then timed out. Should be good for a re-review @dominikschubert 👍 
```
2023-02-21T15:01:29.4233622Z =========================== short test summary info ============================
2023-02-21T15:01:29.4234226Z FAILED ../localstack/tests/integration/test_redshift.py::TestRedshift::test_create_clusters
2023-02-21T15:01:29.4234908Z = 1 failed, 1346 passed, 292 skipped, 67 xfailed, 12 xpassed, 4877 warnings, 2 rerun in 3691.38s (1:01:31) =
```

localstack/services/cloudformation/engine/transformers.py
```diff
@@ -0,0 +1,59 @@
...
+        else:
+            LOG.warning("Unexpected Location parameter for AWS::Include transformer: %s", location)
+        return parameters
+
+
+# maps transformer names to implementing classes
+transformers: Dict[str, Type] = {"AWS::Include": AwsIncludeTransformer}
+
+
+def apply_transform_intrinsic_functions(template: dict, stack=None) -> dict:
```
dominikschubert, line 42: While I understand why we currently need it, I'm not super happy about the stack class being in here again :grimacing:

dominikschubert (approved):
LGTM now :rocket: 

Thanks for addressing all the raised nits/concerns :+1:

PR #7713, opened 2023-02-19
extend persistence api

thrau:
This PR extends the persistence API to prepare localstack core for the persistence extension.

Most noteworthy, the PR adds a wrapper around pickle/dill, and exposes a similar API that is composable and extensible.
* Visitors in the persistence extensions use `Encoder` and `Decoder` instances (they give more flexibility and decouple higher-level logic like visitors from pickle), but `PickleEncoder` and `PickleDecoder` by default
* If you're lazy and you don't want to use encoders, you can also just use `from localstack.state import pickle; pickle.dump(...)`
* Wraps the global reducer function dispatch table in dill (`dill.Pickler.dispatch`) with a mechanism that can also match subclasses (apply certain reducers for all subclasses rather than just the specific type)
* A plugin interface for the snapshot-based persistence mechanism that allows one to plug in custom visitors for save and load operations

`pickle.dumps` can now be used in the store police, as it will be extended dynamically with custom serializers by the persistence plugin, and the underlying system is used for both pods and snapshot persistence.


localstack/state/pickle.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def register(self, subclasses=False):
+        """
+        Registers this ObjectStateReducer's reducer function. See ``pickle.register``.
+        """
+        add_dispatch_entry(self.cls, self._pickle, subclasses=subclasses)
+
+    def _pickle(self, pickler, obj: _T):
+        state = self.get_state(obj)
+        # this is a builtin.Certificate which cannot be pickled, but can be restored from data inside the CertBundle
```
whummer, line 318: nit: is this comment relevant in the context here? (probably a copy/pasta..?)

whummer (approved):
Awesome set of changes @thrau ! 🚀 This gives us a nice flexibility to keep the generic API in this repo, and extend it elsewhere with plugins (via `SnapshotPersistencePlugin`, and the visitor pattern).

Was initially thinking whether introducing our own `pickle` module (with the same convenience functions `pickle.loads`, `pickle.dumps` etc) could cause some confusion if users are expecting to be interacting with the built-in `pickle` library. But then my second reaction was that it actually can be beneficial - this way we make it more explicit that our custom mechanism is being used. (It's also not uncommon to see things like `import dill as pickle`, which would be a similar drop-in replacement as we're defining here.) 👌  I guess we only need to ensure that we communicate to devs that raw `pickle` should never be used directly (which should not be the case anyway, as the core logic is encapsulated in the core runtime).

Love the direction this is taking - can't wait to see it in action! 🥳 🚀

PR #7716, opened 2023-02-20
Use token in GitHubReleaseInstaller for increased rate-limits

dominikschubert:
We're often running into rate-limits in the `GitHubReleaseInstaller` which calls the GitHub Rest API to get a list of artifacts for a given release. This is by default capped at **60 calls per hour** (per IP) for **unauthenticated** calls and any further calls currently lead to pipeline failures.

In CircleCI we're now using a custom secret (`GITHUB_API_TOKEN` which is a personal access token from @localstack-bot without any permissions). This increases the rate-limit to **5_000 calls per hour** (independent of IP)

In GitHub we're now using the intrinsic `GITHUB_TOKEN` secret. This increases the rate-limit to **1_000 calls per hour** 
 (independent of IP). If we ever need a higher limit in GitHub we can also use the token, but as long as it's not necessary I feel using `GITHUB_TOKEN` is more reliable.

Follow-up TODO:
- [x] check for more GitHub API calls in LocalStack and add token
- [ ] Use token in -ext repo

### DIFF
--- a/.github/workflows/pro-integration.yml
+++ b/.github/workflows/pro-integration.yml
@@ -47,6 +47,8 @@ jobs:
       run:
         working-directory: localstack-ext
     environment: localstack-ext-tests
+    env:
+      GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }} # used for increased rate-limits when installing packages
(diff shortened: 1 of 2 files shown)


thrau (approved):
awesome! this solves the problem at least for our CI runs

Issue #7703, opened 2023-02-16
bug: [PRO] Long DBClusterIdentifier for AWS::RDS::DBCluster with MySQL engine in Cloudformation stack causes cluster creation error


hrishi-hugo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We have the following Cloudformation step in which we create a AuroraV2 cluster and a AuroraV2DBInstance.
But when the DBClusterIdentifier is long (I am not able to determine what the limit is), but when we have a long DBClusterIdentifier property (Ex: examplestackhscmn-account-cluster) it causes a cluster creation error with the following stack trace. When I put a short value for DBClusterIdentifier, the stack creates without any issue.

CF snippet: 

```
AuroraV2Cluster:
    Type: AWS::RDS::DBCluster
    DeletionPolicy: Delete
    Properties:
      DatabaseName: hs
      DBClusterIdentifier: examplestackhscmn-account-cluster
      EnableCloudwatchLogsExports:
        - error
        - audit
        - slowquery
      EnableIAMDatabaseAuthentication: true
      Engine: aurora-mysql
      KmsKeyId:
        Ref: AuroraDbKmsKey
      MasterUsername: user
      MasterUserPassword: password
      ServerlessV2ScalingConfiguration:
        MinCapacity: 0.5
        MaxCapacity: 1
      StorageEncrypted: true
      VpcSecurityGroupIds:
        - Fn::Sub: ${SG}
  AuroraV2DBInstance:
    Type: AWS::RDS::DBInstance
    DependsOn:
      - AuroraV2Cluster
    Properties:
      DBClusterIdentifier: examplestackhscmn-account-cluster
      DBInstanceClass: db.serverless
      Engine: aurora-mysql
      PubliclyAccessible: true
```

Stack Trace:

```2023-02-16T17:03:11.510 DEBUG --- [uncthread111] l.u.c.docker_sdk_client    : Executing command in container ls-mysql-1f15ccdb: ['bash', '-c', "mysql -u root -p$MYSQL_ROOT_PASSWORD -e 'SELECT 1'"]
2023-02-16T17:03:14.620 DEBUG --- [uncthread111] l.u.c.docker_sdk_client    : Executing command in container ls-mysql-1f15ccdb: ['bash', '-c', "mysql -u root -p$MYSQL_ROOT_PASSWORD -e 'SELECT 1'"]
[...]

PR #7717, opened 2023-02-20
Add retry cycles for deleting stack resources

dominikschubert:
@steffyP  found an issue with stack deletion when deleting a stack that contains resources with dependencies

https://github.com/localstack/localstack/issues/7703#issuecomment-1436728012

## Changes
- Add retry loop around deletion of stack resources and only try to delete stack resources that are not already in a `DELETE_COMPLETE` state.

The loop is similar to the core deployment loop we already have. This retry now allows that first an independent resource is deleted and the dependent one is retried up to 10 times which is usually enough for the old one to be deleted as well.

There are of course still a few issues but they are quite out of scope like doing this asynchronously and not primarily  relying on the loop, but instead using the resource graph.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -1007,13 +1007,18 @@ class TemplateDeployer:
             resource["Properties"] = resource.get("Properties", clone_safe(resource))
             resource["ResourceType"] = resource.get("ResourceType") or resource.get("Type")
 
+        def _safe_lookup_is_deleted(r_id):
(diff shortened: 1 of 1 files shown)


steffyP (commented):
Thanks for jumping on this so quickly! 🔝 
I tested the use case for RDS cluster/instance locally, and it works 😄 

however, some of the tests in the CI are failing now for `stack.destroy()`😞

steffyP (approved):
LGTM 👍 😄 
Running/Deleting the rds-stack sample locally seems to work, and there are no more test failures 🚀

PR #7509, opened 2023-01-17
Feature CFn Macros/Transformations

pinzon:
This PR adds supports for CFn Macros/Global Transformations. Several feature tests were added in a previous PR so the objective of this one is to make several of those tests pass.

Tests Passing:

- [x] test_macro_deployment
- [x] test_global_scope
- [x] test_capabilities_requirements
- [x] test_validate_lamba_internals
- [x] test_to_validate_template_limit
- [x] test_error_pass_macro_as_reference
- [x] test_functions_and_references_during_transformation.
- [x] test_failed_state

Notes: 
- the correct behaviours and errors messages need to be correctly defined for the update_stack and create_change_set operations.

PR #7718, opened 2023-02-20
Don't require CAPABILITY_AUTO_EXPAND for change sets

dominikschubert:
Fixes a regression introduced in #7509

CFn Changesets, in contrast to Stacks, don't require the `CAPABILITY_AUTO_EXPAND` capability set when they include transformations/macros.

```
          NOTE:
                 This capacity doesn't apply  to  creating  change  sets,  and
                 specifying it when creating change sets has no effect. If you
                 want to create a stack from a stack  template  that  contains
                 macros and nested stacks, you must create or update the stack
                 directly from the template using the  CreateStack or   Updat-
                 eStack action, and specifying this capability.
```

TODOs:

- [ ] add test covering change sets with transformation


pinzon (approved):
LGTM! :rocket:

PR #7719, opened 2023-02-20
Allow container port binding for privileged ports

dfangl:
## Motivation
Currently, we do not allow checking privileged ports to be bound by LS. This leads to a problem with a hostport configuration in ECS for privileged ports - even if they are free.

Also, we are checking for each port we check on the host, if the port is free in the container as well. We should not do this, as we might want to use ports bound inside LS on the host as well.

## Changes
* Introduce separate port ranges for ports returned by `reserve_available_container_port` and ports which can be reserved on request. Per default, the former only returns non-privileged ports, while the whole port range can be reserved on explicit request for a port
* Add additional option to override the port range for reserving available ports. This can be used to further limit the random port range, for example for ECS: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html (The default ephemeral port range for Docker version 1.6.0 and later is listed on the instance under /proc/sys/net/ipv4/ip_local_port_range. If this kernel parameter is unavailable, the default ephemeral port range from 49153 through 65535 is used.)

## Remaining Problems
If checking is_port_available_for_containers before reserving the port, with the new changes, the port check will be done twice, which can lead to reduced performance.
Where possible, we should try reserving and catching the exception, instead of checking if it is possible first.

### DIFF
--- a/localstack/utils/docker_utils.py
+++ b/localstack/utils/docker_utils.py
@@ -23,7 +23,6 @@ PORT_START = 0
 PORT_END = 65536
 RANDOM_PORT_START = 1024
 RANDOM_PORT_END = 65536
-reserved_docker_ports = PortRange(PORT_START, PORT_END)
 
 
 def is_docker_sdk_installed() -> bool:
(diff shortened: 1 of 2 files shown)


whummer (approved):
Kudos for tackling this @dfangl ! 🚀 As we also briefly discussed, it makes perfect sense to include privileged ports into the container range as well. Nice and clean implementation with the additional `_DockerPortRange` class! 👍

PR #7724, opened 2023-02-21
add rest api fixture

calvernaz:
This PR started by just adding the fixture to the testing package, but end up adding more baggage to it.

### DIFF
--- a/localstack/testing/pytest/fixtures.py
+++ b/localstack/testing/pytest/fixtures.py
@@ -1932,6 +1932,7 @@ def sample_backend_dict() -> BackendDict:
 
     return BackendDict(SampleBackend, "sns")
 
+
 @pytest.fixture
 def create_rest_apigw(apigateway_client):
     rest_api_ids = []
--- a/localstack/testing/pytest/fixtures.py
+++ b/localstack/testing/pytest/fixtures.py
@@ -1934,10 +1934,13 @@ def sample_backend_dict() -> BackendDict:
 
 
 @pytest.fixture
-def create_rest_apigw(apigateway_client):
+def create_rest_apigw():
     rest_api_ids = []
 
-    def _create_apigateway_function(*args, **kwargs):
+    def _create_apigateway_function(**kwargs):
+        region_name = kwargs.pop("region_name", None)
+        apigateway_client = _client("apigateway", region_name)
+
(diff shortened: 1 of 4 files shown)


localstack/testing/pytest/fixtures.py
```diff
@@ -1931,3 +1931,23 @@ def __init__(self, region_name, account_id):
...
+        rest_api_ids.append(rest_api_id)
+        resources = apigateway_client.get_resources(restApiId=rest_api_id)
+        root_id = next(item for item in resources["items"] if item["path"] == "/")["id"]
+
+        return response.get("id"), response.get("name"), root_id
+
+    yield _create_apigateway_function
+
+    for rest_api_id in rest_api_ids:
+        with contextlib.suppress(Exception):
```
bentsku, line 1952: Wouldn't suppressing the exception leave the resource existing in AWS if we hit the rate limit? I see the issue with moving `tests.integration.apigateway.test_apigateway_api.apigw_create_rest_api` in `fixtures.py` as it does not return the same data, and the new tests rely on it for snapshot. Not sure how we could proceed for that?

whummer: Good thinking. I actually suggested this change to Cesar - it's a pattern that we've been following more recently, to at least ensure that the rest of the list is processed if a single operation in the cleanup loop fails. 👍  

But I guess at some point we may need to optimize this even further, introducing retries, etc. - let's see how it goes over time (plus, we need to keep an eye on left-over resources in AWS).

Haven't seen the case of rate limiting very frequently, though - did you experience in your testing @bentsku ?

bentsku: Yes, I agree the pattern is nice! 👍

Yes, API GW RestAPI are extremely rate limited (first time I experienced that too), you can approximately delete 1 RestAPI per 8~10sec (even in the UI actually!), even boto retries are not enough.

During my snapshot testing session, I was left with over 15 apis in AWS 😭 that I deleted in UI to be sure, and rate limited there too.

calvernaz: UI is fussier for sure. I never saw them on tests, it doesn't mean they don't exist.

whummer (approved):
LGTM! 👍 Great to see this moving into a pytest fixture

PR #7728, opened 2023-02-22
Fix apigateway path encoding

joe4dev:
This PR fixes an rest api parity issue related to encoding paths in URL.

## Actual behavior

Given the following invocation URL against a rest api (e.g., with `{proxy+}` integration and a lambda function as backend):

`GET https://{restapi_id}.execute-api.{region}.amazonaws.com/{stage_name}/user/test%2Balias@gmail.com/plus/test+alias@gmail.com?email=test%2Balias@gmail.com`

* `path`: `"/{stage_name}/user/test+alias@gmail.com/plus/test+alias@gmail.com?email=test%2Balias@gmail.com"`
* `queryStringParameters.email`: `test+alias@gmail.com`

## Expected behavior

The base `path` should not decode URL encoded characters such as `%2B => +` and looks like this in AWS: `/user/test%2Balias@gmail.com/plus/test+alias@gmail.com`

## Implementation

Instead of relying on decoded `request.path` or `request.full_path` from Werkzeug, we need to compose the invocation `path` manually consisting of a non-decoded base part and a decode params part.

## Questions

* When can `url_params` be None?
* Can we assume defaults for non-present stage and path variables or should we fail fast?
* Does the current solution work with custom domain names where the `stage_name` is omitted?
* Any other special cases to consider?

### DIFF
--- a/localstack/services/apigateway/router_asf.py
+++ b/localstack/services/apigateway/router_asf.py
@@ -31,7 +31,13 @@ def to_invocation_context(
         url_params = {}
 
     method = request.method
-    path = request.full_path if request.query_string else request.path
+    url_params_stage = url_params.get("stage", "dev")
+    url_params_path = url_params.get("path", request.path)
+    # Base path is not URL-decoded. Example: test%2Balias@gmail.com => test%2Balias@gmail.com
+    path = f"{url_params_stage}/{url_params_path}"
+    if request.query_string:
+        # Parameters are URL-decoded. Example: test%2Balias@gmail.com => test+alias@gmail.com
+        path += f"?{request.query_string.decode()}"
     data = restore_payload(request)
     headers = Headers(request.headers)
 
--- a/localstack/services/apigateway/router_asf.py
+++ b/localstack/services/apigateway/router_asf.py
@@ -17,6 +17,9 @@ from localstack.utils.aws.aws_responses import LambdaResponse
 LOG = logging.getLogger(__name__)
 
 
+# TODO: with the latest snapshot tests, we might start moving away from the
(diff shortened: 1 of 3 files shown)


localstack/services/apigateway/router_asf.py
```diff
@@ -31,7 +31,13 @@ def to_invocation_context(
         url_params = {}
```
joe4dev, line 31: When can `url_params` be None?

localstack/services/apigateway/router_asf.py
```diff
@@ -31,7 +31,13 @@ def to_invocation_context(
...
 
     method = request.method
-    path = request.full_path if request.query_string else request.path
+    url_params_stage = url_params.get("stage", "dev")
+    url_params_path = url_params.get("path", request.path)
```
joe4dev, line 35: Can we assume defaults for non-present stage and path variables or should we fail fast?

localstack/services/apigateway/router_asf.py
```diff
@@ -31,7 +31,13 @@ def to_invocation_context(
...
+    # Base path is not URL-decoded. Example: test%2Balias@gmail.com => test%2Balias@gmail.com
+    path = f"{url_params_stage}/{url_params_path}"
```
joe4dev, line 37: Does the current solution work with custom domain names where the stage_name is omitted?

calvernaz (approved):
Thanks for adding the parity test @joe4dev, we should have a third-party approving this PR 👍

whummer (approved):
Awesome, great to see this covered with a snapshot @joe4dev @calvernaz ! 🚀

PR #7580, opened 2023-01-31
Improve support for request integration parameters

calvernaz:
Improve support for request parameters.

Terraform Sample - https://github.com/localstack/localstack-terraform-samples/tree/master/apigateway-appsync-integration

Links to: https://github.com/localstack/localstack/pull/7729

PR #7729, opened 2023-02-22
fix: api gateway headers request mapping

amitassaraf:
Currently, API gateway blindly passes all headers from the request to the integration it connects with, the function I added does 2 things to mimic AWS functionality:

1. Only mapped headers will be passed to the integration (of course only if the function is used, currently it's only used in appsync)
2. Adds support for static value in request parameter headers see the documentation attached below

image
https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html


calvernaz (commented):
Hi @amitassaraf, good timing! we are working on a generic implementation for the request parameters that will go out soon. So this code is being removed in a way. We'll keep this around for reference and close after we merge upstream

alexrashed:
Hi @amitassaraf and @calvernaz!
I think this PR should be obsolete by now, as the referenced PR has been merged already.
Could you please confirm if this is the case and maybe close this PR?
Thanks!

bentsku:
I'll go ahead and close the PR as this was superseded by #7580.

Thanks again for your contribution, sorry this was being worked on simultaneously.

PR #7730, opened 2023-02-22
restructure python distributions

alexrashed:
This PR is a preparation for upcoming changes with v2 concerning the packaging and Docker image structure.
With this PR, the python distribution which is published from this repository is renamed from `localstack` to `localstack-core`.
It loses its dependency on our "extensions" distribution (which will resolve the build and release issues due to our cyclic dependency, and will introduce a clear separation between community and pro).
Some of the CLI / bootstrap tests have been moved to the extensions repo, since they were depending on them.


localstack/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "1.4.1.dev"
+__version__ = "1.4.1.dev1"
```
thrau, line 1: this should remain `1.4.1.dev` i think. it signifies that this is not the concrete version number

setup.cfg
```diff
@@ -28,9 +28,10 @@ install_requires =
     boto3>=1.25.0
     click>=7.0
     cachetools~=5.0.0
-    #dnspython==1.16.0
+    cryptography
+    dill==0.3.2
+    dnspython>=1.16.0
     localstack-client>=1.37
-    localstack-ext>=1.4.1.dev,<1.5
```
thrau, line 33: :fire: :fire: :fire:

localstack/__init__.py
thrau, line 1: i think this should stay `1.4.1.dev` since it indicates a placeholder, whereas `dev1` is a concrete release number.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = localstack
+name = localstack-core
 description = LocalStack - A fully functional local Cloud stack
```
thrau, line 3: Not super critical right now, but I wonder whether we could maybe also reword the description of the distribution to something like "The core library and runtime of LocalStack". Similarly we could probably remove the long description from the distribution, since we have it in the `localstack` distribution.

thrau (commented):
Looks mostly good to me! Just a couple of comments.

I wonder what we'll do about the CLI tests to get the coverage information?

alexrashed:
> I wonder what we'll do about the CLI tests to get the coverage information?

@dfangl also raised concerns about the CLI tests (since we need to identify incompatibilities with Python < 3.10 asap).
This is why I re-introduced all CLI unit tests which do not rely on our extensions.

localstack/__init__.py
alexrashed: Thanks! That was actually a leftover from the initial publishing of `localstack-core` to resolve the cyclic dependency in the pipeline.

thrau (approved):
This looks pretty good now! Let's release and merge? ;)

Issue #7569, opened 2023-01-28
bug: dynamodb.UpdateContinuousBackups UnknownOperationException when using Terraform’s aws_dynamodb_table point_in_time_recovery


davecardwell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

Replacement issue for #1197

### Current Behavior

Terraform’s `aws_dynamodb_table` has a `point_in_time_recovery` attribute you can enable. Doing so with LocalStack results in `AWS dynamodb.UpdateContinuousBackups => 400 (UnknownOperationException)`.

The table creation succeeds, but the overall `terraform apply` fails and Terraform considers it to be a tainted resource since the backup enabling operation failed.

Terraform output:

> aws_dynamodb_table.example: Creating...
> │ Error: creating Amazon DynamoDB Table (example): enabling point in time recovery: updating PITR: UnknownOperationException: An unknown operation was requested.
> │ 	status code: 400, request id: UA1IUBFM6870CXD0XVRJ5S9DM37XSJ7FGNUCCYU2D5T251KKPQPE

LocalStack logs:

[...]

PR #7735, opened 2023-02-23
Minimal implementation for continuous backup

giograno:
This PR enables a minimal implementation for the `UpdateContinuousBackups` and `DescribeContinuousBackups` actions.

The feature of continuous backup is actually not implemented. We only return the minimum needed to achieve Terraform parity for https://github.com/localstack/localstack/issues/7569.

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -29,12 +29,15 @@ from localstack.aws.api.dynamodb import (
     BatchWriteItemInput,
     BatchWriteItemOutput,
     BillingMode,
+    ContinuousBackupsDescription,
+    ContinuousBackupsStatus,
     CreateGlobalTableOutput,
     CreateTableInput,
     CreateTableOutput,
     DeleteItemInput,
     DeleteItemOutput,
     DeleteTableOutput,
+    DescribeContinuousBackupsOutput,
     DescribeGlobalTableOutput,
     DescribeKinesisStreamingDestinationOutput,
     DescribeTableOutput,
@@ -55,6 +58,9 @@ from localstack.aws.api.dynamodb import (
     ListTagsOfResourceOutput,
     NextTokenString,
     PartiQLBatchRequest,
+    PointInTimeRecoveryDescription,
+    PointInTimeRecoverySpecification,
+    PointInTimeRecoveryStatus,
     PositiveIntegerObject,
     ProvisionedThroughputExceededException,
     PutItemInput,
@@ -73,6 +79,7 @@ from localstack.aws.api.dynamodb import (
     ScanInput,
     ScanOutput,
     StreamArn,
+    TableDescription,
     TableName,
     TagKeyList,
     TagList,
@@ -81,6 +88,7 @@ from localstack.aws.api.dynamodb import (
     TransactGetItemsOutput,
     TransactWriteItemsInput,
     TransactWriteItemsOutput,
+    UpdateContinuousBackupsOutput,
     UpdateGlobalTableOutput,
     UpdateItemInput,
     UpdateItemOutput,
@@ -109,7 +117,7 @@ from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.aws import arns, aws_stack
 from localstack.utils.aws.arns import extract_account_id_from_arn, extract_region_from_arn
 from localstack.utils.aws.aws_stack import get_valid_regions_for_service
-from localstack.utils.collections import select_attributes
+from localstack.utils.collections import select_attributes, select_from_typed_dict
 from localstack.utils.common import short_uid, to_bytes
 from localstack.utils.json import BytesEncoder, canonical_json
 from localstack.utils.strings import long_uid, to_str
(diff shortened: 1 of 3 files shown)


localstack/services/dynamodb/provider.py
```diff
@@ -545,7 +553,7 @@ def describe_table(self, context: RequestContext, table_name: TableName) -> Desc
                     "TableClass": table_definitions["TableClass"]
                 }
 
-        return result
+        return DescribeTableOutput(Table=select_from_typed_dict(TableDescription, result["Table"]))
```
whummer, line 556: nit: in a few lines above we're hedging with `result.get("Table", {})`, whereas here we're using `result["Table"]`.  Not sure if there can be a case where Table would not be present (`_forward_request(..)` would raise if the table doesn't exist), so perhaps we could change the other lines to use `result["Table"]` as well.. 👍 (just cosmetics, not critical at all..)

localstack/services/dynamodb/provider.py
```diff
@@ -1197,6 +1205,56 @@ def describe_kinesis_streaming_destination(
...
         )
 
+    #
+    # Continuous Backups
+    #
+
+    def describe_continuous_backups(
+        self, context: RequestContext, table_name: TableName
+    ) -> DescribeContinuousBackupsOutput:
+        self.get_global_table_region(context, table_name)
```
whummer, line 1215: Should we take the result of this call and use it as the region for the store? (same pattern we're also using for other methods in the provider..)
```
        global_table_region = self.get_global_table_region(context, table_name)
        store = get_store(context.account_id, region=global_table_region)
```

localstack/services/dynamodb/provider.py
```diff
@@ -1197,6 +1205,56 @@ def describe_kinesis_streaming_destination(
...
+            ContinuousBackupsDescription=continuous_backup_description
+        )
+
+    def update_continuous_backups(
+        self,
+        context: RequestContext,
+        table_name: TableName,
+        point_in_time_recovery_specification: PointInTimeRecoverySpecification,
+    ) -> UpdateContinuousBackupsOutput:
+        self.get_global_table_region(context, table_name)
```
whummer, line 1236: same as above, regarding assigning and using `global_table_region`..

localstack/services/dynamodb/provider.py
```diff
@@ -1197,6 +1205,56 @@ def describe_kinesis_streaming_destination(
...
+            if point_in_time_recovery_specification["PointInTimeRecoveryEnabled"]
+            else PointInTimeRecoveryStatus.DISABLED
+        )
+        continuous_backup_description = ContinuousBackupsDescription(
+            ContinuousBackupsStatus=ContinuousBackupsStatus.ENABLED,
+            PointInTimeRecoveryDescription=PointInTimeRecoveryDescription(
+                PointInTimeRecoveryStatus=pit_recovery_status
+            ),
+        )
+        store.table_properties[table_name] = {
```
whummer, line 1250: This would potentially overwrite other table properties - I believe this should be something like:

```
        table_props = store.table_properties.setdefault(table_name, {})
        table_props["ContinuousBackupsDescription"] = continuous_backup_description
```

tests/integration/test_dynamodb.py
```diff
@@ -1610,6 +1615,45 @@ def test_data_encoding_consistency(
...
+        def wait_for_continuous_backend():
+            try:
+                dynamodb_client.update_continuous_backups(
+                    TableName=table_name,
+                    PointInTimeRecoverySpecification=PointInTimeRecoverySpecification(
+                        PointInTimeRecoveryEnabled=True
+                    ),
+                )
+                return True
+            except ContinuousBackupsUnavailableException:
```
whummer, line 1640: Out of curiosity - under which circumstances would we run into this error here? Is this happening during creation of the DDB table, until the table status becomes `ACTIVE`? Wondering if we should add a flag to the `dynamodb_create_table` fixture to wait until the table becomes active. Also, we have a `dynamodb_wait_for_table_active` fixture, which may come in handy here?

whummer (approved):
Great to see this supported @giograno , further increasing our parity and TF compatibility! 🎉 Added a few minor comments, would be great if we can address them before merging 👍

tests/integration/test_dynamodb.py
giograno: By default, all the tables in DDB have continuous backup enabled. However, I found that AWS takes a few seconds to enable such backups after a table has been created and is active. This basically was needed to generate the snapshot.

localstack/services/dynamodb/provider.py
giograno: I am actually not sure about the behavior here. 
If I remember correctly our implementation of global tables, we actually keep in DynamoDB local a single table in a given region. When we receive a request from a global region, we check with `get_global_table_region` which is the region that actually holds the data in DDBLocal, and forwards the calls to that specific region with `self._forward_request(context=context, region=global_table_region)` (we basically modify the region in the context to hit the "original" region). All over the provider, we use this `global_table_region` only to patch the forwarded request, but we access the store with the actual region in the context.

localstack/services/dynamodb/provider.py
giograno: Same case as above.

localstack/services/dynamodb/provider.py
giograno: Good catch! You are absolutely right 🙏

whummer (approved):

PR #7738, opened 2023-02-23
improve parity for APIGW Resource

bentsku:
Work on APIGW Resource parity, improve UpdateResource parity (patch operations validation), remove moto patches to move them into provider directly (we still manipulates moto entities directly as its tightly integrated, but extend moto with our own store), and fix deleting child resources when deleting a resource that was used as a parent.

Example:
`/` -> `/pets` -> `/pets/dogs` 
We delete `/pets`, `/pets/dogs` should be deleted as well because it declared `/pets` to be its parent.
We do this by mapping children IDs when they are created with a parent ID. When deleting a resource, we check if it is in the parent list, and then recursively delete resources. 

- [x] remove patches on Resource entity
- [x] create AWS validated tests with snapshots for resource manipulation
- [x] fix parity for `UpdateResource` (validation)
- [x] fix recursively deleting all child `Resource` when deleting a parent `Resource` with `DeleteResource`

note: I might delete the fixture for creating resource in a following PR, as we clean up all related resources when deleting the RestAPI.
note2: more tests are coming for deleting the resource with more children in a following PR. (#7746)
note3: I'd like to do a store refactor to allow easier validation of resources, see #7750 

\cc @whummer

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -5,6 +5,7 @@ from copy import deepcopy
 from typing import IO, Any
 
 from moto.apigateway import models as apigw_models
+from moto.apigateway.models import Resource as MotoResource
 from moto.apigateway.models import RestAPI as MotoRestAPI
 from moto.core.utils import camelcase_to_underscores
 
@@ -22,6 +23,7 @@ from localstack.aws.api.apigateway import (
     Boolean,
     ClientCertificate,
     ClientCertificates,
+    ConflictException,
     CreateAuthorizerRequest,
     CreateRestApiRequest,
     DocumentationPart,
@@ -93,6 +95,16 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
     @staticmethod
     def _get_moto_backend(context: RequestContext) -> apigw_models.APIGatewayBackend:
         return apigw_models.apigateway_backends[context.account_id][context.region]
+
+    @staticmethod
+    def _remove_rest_api(context: RequestContext, rest_api_id: str) -> None:
+        store = get_apigateway_store(account_id=context.account_id, region=context.region)
+        # clean up this way until we properly encapsulate RestApi in the store
+        store.authorizers.pop(rest_api_id, None)
+        store.validators.pop(rest_api_id, None)
+        store.documentation_parts.pop(rest_api_id, None)
+        store.gateway_responses.pop(rest_api_id, None)
+        store.resources_children.pop(rest_api_id, None)
 
     @handler("TestInvokeMethod", expand=False)
     def test_invoke_method(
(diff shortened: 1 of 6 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -89,6 +93,16 @@ def on_after_init(self):
     def _get_moto_backend(context: RequestContext) -> apigw_models.APIGatewayBackend:
         return apigw_models.apigateway_backends[context.account_id][context.region]
 
+    @staticmethod
+    def _remove_rest_api(context: RequestContext, rest_api_id: str) -> None:
+        store = get_apigateway_store(account_id=context.account_id, region=context.region)
```
calvernaz, line 98: it removes from Localstack store, should we check on Moto's as well ?

tests/integration/apigateway/test_apigateway_api.py
```diff
@@ -73,6 +75,25 @@ def _factory(*args, **kwargs):
         delete_rest_api_retry(apigateway_client, rest_api_id)
 
 
+@pytest.fixture
+def apigw_create_rest_resource(apigateway_client):
```
calvernaz, line 79: I'm curious about how to create this fixture. Deleting the API container should remove most of the resources created under path, templates, etc - backend integration lingers tho. Any particular use case where this didn't happen? meaning deleting the API wasn't removing resources?

localstack/services/apigateway/provider.py
```diff
@@ -89,6 +93,16 @@ def on_after_init(self):
...
+        # clean up this way until we properly encapsulate RestApi in the store
+        store.authorizers.pop(rest_api_id, None)
```
calvernaz, line 100: Since we want to move the state from moto to localstack, in a next iteration we can functionality to the store to delete an API, basically wrapping a method to all the operation below.

localstack/services/apigateway/provider.py
```diff
@@ -844,6 +972,17 @@ def filter_function(item):
 # ---------------
 
 
+def get_moto_rest_api(context: RequestContext, rest_api_id: str) -> MotoRestAPI:
```
calvernaz, line 975: we do have this method `self._get_moto_backend(context)` on the provider similar to this one. This one seems to handle mis-lookups, should we highlight that and remove usages of the former?

calvernaz (approved):
Thanks for doing this, I know working with the split state doesn't make it easier. LGTM 🚀 

I have some comments, that can be handle in next iterations

localstack/services/apigateway/provider.py
bentsku: Right, I'm not sure about adding methods to the store, I've seen it done in some cases, not sure what the proper way of doing it is, any idea @thrau?

localstack/services/apigateway/provider.py
bentsku: You're right, the `self._get_moto_backend(context)` was from my first PR, but not sure it still makes sense now, this new method handles a bit more. If we ever need to access the moto backend to access other kind of resources and not RestAPI, we can create a new method for that resource with the proper exception message. Thanks for picking this up!

tests/integration/apigateway/test_apigateway_api.py
bentsku: I'm deleting it, it was me not exactly knowing how it would behave. Thanks!

localstack/services/apigateway/provider.py
bentsku: The `call_moto` should take care of that as the rest API is contained there, I'm not sure how this could happen in moto

PR #7740, opened 2023-02-23
Refactor authorizer info in API Gateway invocation context

whummer:
Refactor authorizer info in API Gateway invocation context. Companion to upstream PR.

The main changes in this PR is to rename the `auth_info` attribute and unify it with `auth_context` in the API GW invocation context

The PR also adds a few type hints across the codebase.


calvernaz (approved):
Thanks for the speedy reaction and the cleanup!

PR #7746, opened 2023-02-24
improve apigw parity work for Resource proxy path

bentsku:
When moving on to implement resource methods validation/removing patches, I've seen this page of the docs:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-request.html#api-gateway-proxy-resource
So I followed it to implement more validations on the previous PR #7738

Stumbled unto an interesting issue in the docs:
> An API can have more than one proxy resource. For example, the following proxy resources are allowed within an API.
```
/{proxy+}
/parent/{proxy+}
/parent/{child}/{proxy+}
```
But it's not actually allowed, as the 2 last paths have a variable part, which will throw an exception. Implemented all the validations around this example in the docs. 

Based on #7738

\cc @whummer


calvernaz (approved):
LGTM!

PR #7749, opened 2023-02-24
apigw parity work resource method

bentsku:
This PR reimplements the `Method` resource CRUD API, removing the patches from moto to integrate them in our provider. 
There was quite some validations on the json patch area, I will focus my time on resources where `Update*` is more often called, as those can be huge rabbit holes. 

- [x] remove all moto patches related to the `Method` resource
- [x] implement validations on `UpdateMethod`, snapshot tested, see https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html#UpdateMethod-Patch
- [x] fixes in json patch logic, especially while adding new field in nested properties
- [x] add validations about "linked" resources to the created/updates resources (Authorizers, Validators, Models, etc)

### DIFF
--- a/localstack/services/apigateway/patches.py
+++ b/localstack/services/apigateway/patches.py
@@ -11,7 +11,7 @@ from moto.apigateway.responses import APIGatewayResponse
 from moto.core.utils import camelcase_to_underscores
 
 from localstack.services.apigateway.helpers import TAG_KEY_CUSTOM_ID, apply_json_patch_safe
-from localstack.utils.common import str_to_bool, to_str
+from localstack.utils.common import str_to_bool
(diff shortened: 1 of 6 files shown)


localstack/services/apigateway/patches.py
```diff
@@ -36,44 +36,6 @@ def apigateway_models_Stage_init(
     apigateway_models_Stage_init_orig = apigateway_models.Stage.__init__
     apigateway_models.Stage.__init__ = apigateway_models_Stage_init
```
calvernaz, line 38: 🔥

calvernaz (approved):
LGTM 🔥

Issue #5025, opened 2021-11-29
bug: s3 ListObjectVersions with delimiter not returning file objects and returning wrong CommonPrefixes


rtkgeoff:
### Is there an existing issue for this?
No
- [X] I have searched the existing issues

### Current Behavior

In a bucket with two objects:
- `dir/test`
- `dir/subdir/test2`

The ListObjectVersions is returning unexpected results using delimiter '/'. Given any of these prefixes:
- `''`
- `'dir/'`
- `'dir/s'`
- `'dir/test'`
- `'dir/subdir'`
- `'dir/subdir/'`
- `'dir/subdir/te'`
- `'dir/subdir/test2'`

the result is

```
{
    "CommonPrefixes": [
        {
            "Prefix": "dir/"
        }
    ]
}
```
which is wrong for all but the first 2 prefixes above.

### Expected Behavior

When given a delimiter `ListObjectVersions` should return any file object versions whose key starts with the given prefix and contains no additional delimiters. The same is true for the directories listed in CommonPrefixes; they must start with the prefix but not contain any additional delimiters except for a final trailing one.

For a bucket with two objects:
- `dir/test`
- `dir/subdir/test2`

ListObjectVersions delimiter=`/` and prefix=`dir/` should return
```
{
  "CommonPrefix":[{"Prefix":"dir/subdir/"}],
  "Versions":[{"Key":"dir/test", ...}]
}
```

ListObjectVersions delimiter=`/` and prefix=`dir/t` should return
```
{
  "Versions":[{"Key":"dir/test", ...}]
}
```

ListObjectVersions delimiter=`/ `and prefix=`dir/s` or prefix=`dir/subdir` should return
```
{
  "CommonPrefix":[{"Prefix":"dir/subdir/"}]
}
```

ListObjectVersions delimiter=`/ `and prefix=`dir/subdir/` or prefix=`dir/subdir/te` or prefix=`dir/subdir/test2`should return
```
{
  "Versions":[{"Key":"dir/subdir/test2", ...}]
}
```

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run --rm -it -p 4566:4566 -e SERVICES=s3 localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    aws --endpoint-url http://localhost:4566 s3 mb s3://mybucket
[...]

PR #7760, opened 2023-02-27
test: list object version with prefix and delimiters

macnev2013:
This PR contains AWS validated test case for the issue mentioned in #5025.


bentsku (approved):
LGTM! So we have parity with AWS, was the change implemented upstream? Thanks for validating the issue of the user!

macnev2013:
Hey @bentsku, Yes I believe it was fixed [here](https://github.com/getmoto/moto/pull/5562).

PR #7762, opened 2023-02-27
Add lambda model and api updates

joe4dev:
## Contributions

* Add support for [runtime management controls](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-update.html) at API level
* Add support for [SnapStart](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html) at API level
* Add snapshot tests for SnapStart API
* Update all lambda-related snapshot tests due to model updates (only for `aws_validated` tests)
* Add platform flag to docker clients for `build_image`
* Fix Lambda tests on ARM hosts. This enables running Lambda tests on ARM hosts.
* Conditionally skip the concurrency tests if the AWS account has too low limits (improve error messages).

## Limitations

* Runtime management controls use a hard-coded sha256 value (`8eeff65f6809a3ce81507fe733fe09b835899b99481ba22fd75b5a7338290ec1`) instead of inspecting the Docker image.
* Rust builds on ARM (tested on M1 MacBook Pro) are slow (~5 minutes) due to emulation. See Makefile for potential performance improvements.
* The SnapStart tests are a little slow (~1-2min) against AWS

## Breaking changes

* This PR includes model changes in lambda and will break CloudPods for the new lambda provider

### DIFF
--- a/tests/integration/awslambda/functions/common/endpointinjection/provided/Makefile
+++ b/tests/integration/awslambda/functions/common/endpointinjection/provided/Makefile
@@ -1,8 +1,33 @@
+DOCKER_PLATFORM ?= linux/amd64
+# Using a sha256 hash breaks on multi-platform hosts (e.g., ARM Macs): https://github.com/moby/moby/issues/43188
+DOCKER_RUST_IMAGE ?= rust:1.67.1
+
+# x86 cross-build on ARM hosts are killed due to memory issues: https://github.com/rust-lang/cargo/issues/10781#issuecomment-1351670409
+# Slow workaround takes ~5 to build on an M1 MacBook Pro: --config net.git-fetch-with-cli=true
+ARM_CROSS_BUILD_WORKAROUND ?= --config net.git-fetch-with-cli=true
+
+# Overview: https://doc.rust-lang.org/nightly/rustc/platform-support.html
+# Targets: x86_64-unknown-linux-musl or x86_64-unknown-linux-gnu or add aarch64-unknown-linux-gnu
+# ARM cross-compile requires some more config and dependencies because cargo does not find cross-build tools itself:
+# https://stackoverflow.com/questions/69360099/apple-m1-to-linux-x86-64-unrecognized-command-line-option-m64
+# --env RUSTFLAGS='-C linker=x86_64-linux-gnu-gcc' yields: error: linker `x86_64-linux-gnu-gcc` not found
+# Alternative cross compilation tool: https://github.com/cross-rs/cross
+# cargo install cross --git https://github.com/cross-rs/cross
+# --env CROSS_CONTAINER_IN_CONTAINER=true -v /var/run/docker.sock:/var/run/docker.sock
+# cross build --target x86_64-unknown-linux-musl
+# "no container engine found" => requires installing Docker
+# ARM support for aarch64-unknown-linux-gnu pending: https://github.com/rust-lang/rust/issues/77071
+RUST_TARGET ?= x86_64-unknown-linux-musl
+
 build:
 	mkdir -p build && \
-	docker run --rm -v $$(pwd)/src:/app -v $$(pwd)/build:/out rust:1.64.0@sha256:922d814994d77f8e3ab8a7db45a277e9cebe41a557046eeef91a2e34b28b4962 bash -c "rustup target add x86_64-unknown-linux-musl && mkdir -p /app2 && cp -r /app/* /app2 && cd /app2 && cargo build --release --target x86_64-unknown-linux-musl && cp ./target/x86_64-unknown-linux-musl/release/bootstrap /out && chown $$(id -u):$$(id -g) /out/bootstrap"
+	docker run --rm --platform=$(DOCKER_PLATFORM) -v $$(pwd)/src:/app -v $$(pwd)/build:/out:cached $(DOCKER_RUST_IMAGE) bash -c "rustup target add $(RUST_TARGET) && mkdir -p /app2 && cp -r /app/* /app2 && cd /app2 && cargo build $(ARM_CROSS_BUILD_WORKAROUND) --release --target $(RUST_TARGET) && cp ./target/$(RUST_TARGET)/release/bootstrap /out && chown $$(id -u):$$(id -g) /out/bootstrap"
 	find ./build -exec touch -t 200001010100.00 {} \;
 	cd ./build && zip -r handler.zip .
 	mv ./build/handler.zip .
 
-.PHONY: build
+clean:
+	$(RM) -r build
+	$(RM) handler.zip
+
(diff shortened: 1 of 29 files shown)


tests/integration/cloudformation/resources/test_lambda.py
```diff
@@ -577,6 +587,14 @@ def wait_logs():
             "$..StartingPosition",
             "$..StateTransitionReason",
             "$..Topics",
+            # resource index mismatch due to SnapStart
```
joe4dev, line 590: That is an example where a new feature changes the indexing of snapshots because it matches a resource transformer. It is not ideal that we need to disable these snapshot verifications for the old provider due to an unrelated change.

tests/integration/awslambda/test_lambda_api.py
```diff
@@ -2412,7 +2434,15 @@ def test_lambda_provisioned_lifecycle(self, lambda_client, create_lambda_functio
...
     condition=is_old_provider,
-    paths=["$..RevisionId", "$..Policy.Statement", "$..PolicyName", "$..PolicyArn", "$..Layers"],
+    paths=[
+        "$..RevisionId",
+        "$..Policy.Statement",
+        "$..PolicyName",
+        "$..PolicyArn",
+        "$..Layers",
+        # mismatching resource index due to SnapStart
+        "$..Statement.Condition.ArnLike.'AWS:SourceArn'",
```
joe4dev, line 2444: `:` break the JsonPath snapshot comparison and need to be enquoted using single quotes

tests/integration/awslambda/test_lambda.py
```diff
@@ -171,12 +171,12 @@ def fixture_snapshot(snapshot):
...
             "$..Layers",
-            "$..SnapStart",  # FIXME
+            "$..SnapStart",
         ],
     )
 else:
     pytestmark = pytest.mark.skip_snapshot_verify(
-        paths=["$..CodeSize", "$..SnapStart"],  # FIXME
+        paths=["$..CodeSize"],  # FIXME
     )
```
dominikschubert, line 180: we can actually just remove codesize as well and add it to the "global" transformers above

tests/integration/awslambda/test_lambda_api.py
```diff
@@ -953,6 +959,7 @@ def test_publish_with_wrong_sha256(
         )
         snapshot.match("publish_result", publish_result)
 
+    @pytest.mark.aws_validated
```
dominikschubert, line 962: good catch :+1:

tests/integration/awslambda/test_lambda_api.py
```diff
@@ -1301,13 +1309,15 @@ def test_notfound_and_invalid_routingconfigs(
...
-        "$..SnapStart"
-    ]
-)
 class TestLambdaRevisions:
+    @pytest.mark.skip_snapshot_verify(
+        # The RuntimeVersionArn is currently a hardcoded id and therefore does not reflect the ARN resource update
+        # from python3.9 to python3.8 in update_function_configuration_response_rev5.
+        paths=[
+            "update_function_configuration_response_rev5..RuntimeVersionConfig.RuntimeVersionArn",
+            "get_function_response_rev6..RuntimeVersionConfig.RuntimeVersionArn",
```
dominikschubert, line 1318:
```suggestion
            "$..RuntimeVersionConfig.RuntimeVersionArn",
```

or was there a particular reason for only including those?

tests/integration/awslambda/test_lambda_api.py
dominikschubert: Maybe we should add some validation for this to avoid errors like this. Was this fairly obvious when debugging?

tests/integration/cloudformation/resources/test_lambda.py
dominikschubert: Agreed. I'm generally not super happy with the state of the resource transformer. I've added this as a reference in our internal issue tracker for snapshots

dominikschubert (approved):
LGTM! :partying_face: 

Great that you uncovered a lot of hidden issues that we'll need to tackle at some point. (e.g. tests missing the validated marker, rust compilation, resource transformer)

tests/integration/cloudformation/resources/test_lambda.py
joe4dev: Good to track that one. I think this might become a nasty source of flaky tests with automated test re-validation.

tests/integration/awslambda/test_lambda_api.py
joe4dev: The error was pretty specific (see below) but the stack trace wasn't particularly helpful in identifying the issue because the hook is executed after the test.
It could be nice to give some context when raising (e.g., which snapshot match or which verify skip path).

```
>       raise JsonPathParserError('Parse error at %s:%s near token %s (%s)'
                                  % (t.lineno, t.col, t.value, t.type))
E       jsonpath_ng.exceptions.JsonPathParserError: Parse error at 1:34 near token : (:)

.venv/lib/python3.10/site-packages/jsonpath_ng/parser.py:81: JsonPathParserError
```

tests/integration/awslambda/functions/common/endpointinjection/provided/Makefile
```diff
@@ -1,8 +1,33 @@
+DOCKER_PLATFORM ?= linux/amd64
+# Using a sha256 hash breaks on multi-platform hosts (e.g., ARM Macs): https://github.com/moby/moby/issues/43188
+DOCKER_RUST_IMAGE ?= rust:1.67.1
+
+# x86 cross-build on ARM hosts are killed due to memory issues: https://github.com/rust-lang/cargo/issues/10781#issuecomment-1351670409
+# Slow workaround takes ~5 to build on an M1 MacBook Pro: --config net.git-fetch-with-cli=true
```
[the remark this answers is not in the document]
joe4dev: yup, fixed

tests/integration/awslambda/test_lambda.py
joe4dev: Good catch 👍 
I just removed it because `CodeSize` is already part of the `lambda_api()` transformer.

tests/integration/awslambda/test_lambda_api.py
joe4dev: Thanks.
I filtered by `-m aws_validated`, hence all lambda tests with any drift should be marked now 🟢

tests/integration/awslambda/test_lambda_api.py
joe4dev: I still wanted the validation that the `RuntimeVersionArn` does not change between revisions and not ignore it completely for the entire test

PR #7764, opened 2023-02-28
[SFN] [ChoiceOperators] redesign, is operators, boolean equals operators, tests

MEPalma:
- Improves the choice operation delegation strategy
- Adds baseline implementation of `Is*` and `BooleanEquals*` operators
- Adds testing setup/env
- Adds tests for added operators

Note: logs a warning on invocations of `IsTimestamp` as AWS's behaviour can only be partially replicated at this time.


tests/integration/stepfunctions/v2/choice_operators/utils.py
```diff
@@ -0,0 +1,98 @@
...
+        execution_arn = exec_resp["executionArn"]
+
+        await_execution_success(
+            stepfunctions_client=stepfunctions_client, execution_arn=execution_arn
+        )
+
+        exec_hist_resp = stepfunctions_client.get_execution_history(executionArn=execution_arn)
+        output = JSONPathUtils.extract_json(
+            "$.events[*].executionSucceededEventDetails.output", exec_hist_resp
+        )
```
dominikschubert, line 96: currently causes an error in the CI pipeline here for empty lists

MEPalma: I could reproduce this on an older version of the verification refactoring, but can't reproduce it locally on head. I've made some changes and merged latest master

tests/integration/stepfunctions/v2/choice_operators/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    (" ", True),  # 12
+    (" ", False),  # 13
+    ("HelloWorld", True),  # 14
+    ("HelloWorld", False),  # 15
+    ("2012-10-09T19:00:55", True),  # 16
+    ("2012-10-09T19:00:55", False),  # 17
+    ("2012-10-09T19:00:55Z", True),  # 18
+    ("2012-10-09T19:00:55Z", False),  # 19
+    ("2023-02-24", True),  # 20
+    ("2023-02-24", False),  # 21
```
dominikschubert, line 34: might also want to add samples for valid timestamps with a timezone, e.g. `2012-10-09T19:00:55+01:00`

dominikschubert (approved):
LGTM!

I'd suggest adding a test for rfc3339 timestamps with a timezone as well, even if we don't yet support them

PR #7766, opened 2023-02-28
[SFN] [ChoiceOperators] numeric operators, tests

MEPalma:
Extends [MEP-sfn-choice-ops](https://github.com/localstack/localstack/pull/7764)
- Adds baseline implementation of  `Numeric*` operators
- Adds tests for added operators


localstack/services/stepfunctions/asl/component/state/state_choice/comparison/operator/implementations/numeric.py
```diff
@@ -0,0 +1,139 @@
+from typing import Any
```
dominikschubert, line 1: I really like the more condensed version here instead of splitting it into too many files :+1:

dominikschubert (approved):
LGTM :+1:

PR #7767, opened 2023-03-01
Catch Route53 Record Sets without TTL

QuinnyPig:
Alias records, for example, should explicitly not have a TTL set, as it's derived from the TTL of the target.

### DIFF
--- a/localstack/services/cloudformation/models/route53.py
+++ b/localstack/services/cloudformation/models/route53.py
@@ -38,9 +38,9 @@ class Route53RecordSet(GenericBaseModel):
                 "HealthCheckId",
             ]
             attrs = select_attributes(params, attr_names)
-
-            if isinstance(attrs["TTL"], str):
-                attrs["TTL"] = int(attrs["TTL"])
+            if "TTL" in attrs:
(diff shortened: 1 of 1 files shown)


localstack/services/cloudformation/models/route53.py
```diff
@@ -38,9 +38,9 @@ def param_change_batch(params, **kwargs):
                 "HealthCheckId",
             ]
             attrs = select_attributes(params, attr_names)
-
-            if isinstance(attrs["TTL"], str):
-                attrs["TTL"] = int(attrs["TTL"])
+            if "TTL" in attrs:
+                if isinstance(attrs["TTL"], str):
+                    attrs["TTL"] = int(attrs["TTL"])
```
giograno, line 43:
```suggestion
            if isinstance(ttl := attrs.get("TTL"), str):
                attrs["TTL"] = int(ttl)
```
We could avoid the nested ifs with something like this.

giograno (approved):
Thanks for the contribution @QuinnyPig! 🙏

thrau:
looked safe to merge, we'll have to check what's going on with the CI with community PRs

PR #7771, opened 2023-03-01
fix SNS storing lambda invocation logs + improve parity

### LINKED ISSUE #7768: bug: SNS Topic Publish to Lambda throws an Exception

### Current Behavior

I am running an SNS topic that publishes to a lambda function. The localstack code reads a type byte object from the lambda function invocation payload [here](https://github.com/localstack/localstack/blob/master/localstack/services/sns/publisher.py#L187) and then tries to dump the result into json [here](https://github.com/localstack/localstack/blob/master/localstack/services/sns/publisher.py#L863) which results in the following error:
```
sns_pub_0] l.services.sns.publisher   : Unable to run Lambda function on SNS message: Object of type bytes is not JSON serializable
```

### Expected Behavior

The payload bytes should be converted to a string before being dumped to json. The error should not be thrown.

With a docker-compose file

docker-compose up
(Using the standard docker-compose.yaml)

```
awslocal sns publish --message file://sns/message.json --topic-arn arn:aws:sns:us-east-1:000000000000:<topic_name>
```

### Environment

```markdown
- OS: macOS Ventura 13.2.1
- LocalStack: latest
```

### Anything else?

_No response_

bentsku:
This PR fixes #7768, a regression introduced by remove a call to `json_safe` on the payload.
However, there was no payload to get here, because the lambda invocation is asynchronous. The parity was wrong here, and it's nice to be able to introduce additional behaviour in SNS.
I've introduced a test by retrieving CloudWatch logs for this particular scenario and improved parity.

We would still need to check parity for all other "Publisher" (SQS, Kinesis, HTTP and Platform application endpoint), to see how they store their logs. See https://docs.aws.amazon.com/sns/latest/dg/sns-topic-attributes.html

We will also need to add a conditional on storing the logs, as they should be enabled first. 

Note: I'll introduce a scenario fixture regarding setting up the logs for Success/Failure in SNS when introducing the next parity check for another publisher!

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -3,8 +3,9 @@ import logging
 from typing import Dict, List
 
 from botocore.utils import InvalidArnException
+from moto.core.utils import camelcase_to_pascal, underscores_to_camelcase
 from moto.sns import sns_backends
-from moto.sns.models import MAXIMUM_MESSAGE_LENGTH
+from moto.sns.models import MAXIMUM_MESSAGE_LENGTH, SNSBackend
 from moto.sns.utils import is_e164
 
 from localstack.aws.accounts import get_aws_account_id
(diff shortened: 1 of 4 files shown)


thrau (approved):
LGTM! nice find, sns parity with publishers is tricky :-)

PR #7774, opened 2023-03-01
Unify hostnames in returned URLs

simonrw:
This PR centralises the way we decide what hostname is returned when requested by the resource.

This PR contains tests that ensure the _current_ behaviour. This is important as I consider changing the way the user configures returned names to be a backwards incompatible change. In a future PR when v2 will be released, we will update the new url function to be systematic, based on the new configuration.


localstack/testing/pytest/fixtures.py
```diff
@@ -1978,3 +1978,54 @@ def factory(**kwargs):
...
         except Exception as e:
             LOG.debug(f"Error cleaning up AppSync API: {api}, {e}")
+
+
+@pytest.fixture
+def assert_host_customisation(monkeypatch):
+    hostname_external = f"external-host-{short_uid()}"
+    localstack_hostname = f"localstack-hostname={short_uid()}"
+    monkeypatch.setattr(config, "HOSTNAME_EXTERNAL", hostname_external)
+    # monkeypatch.setattr(config, "LOCALSTACK_HOSTNAME", localstack_hostname)
```
simonrw, line 1988: I'm struggling with this one. If I uncomment this line, two tests fail:
* opensearch testing the path strategy, and
* lambda testing the legacy http api to generate the function url.

This is because
* opensearch starts the domain with the `LOCALSTACK_HOSTNAME` hostname as its url, and therefore the health check keeps failing since the nonsense  value generated here is not a valid hostname, so LS can't check the domain is up
* a similar story with the lambda endpoint - requests are made to the nonsense hostname as part of the test, which is not a valid hostname.

Though it's documented as read-only, we do allow the user to change `LOCALSTACK_HOSTNAME`, hence why I'm trying to test it here. Since it has to be a valid host (and `localhost` doesn't really count) I'm not sure how to allow this line to be uncommented. Any thoughts?

simonrw: One idea is to set the value to `socket.gethostname()` which should always return a host that's accessible, and is (probably) not `localhost`, which would probably fulfil this criterion.

localstack/services/s3/virtual_host.py
```diff
@@ -79,6 +83,13 @@ def _rewrite_url(url: str, bucket: str, region: str) -> str:
         if region:
             netloc = netloc.replace(f"{region}", "")
 
+        # if the user specifies a custom hostname for LocalStack, this name may not be resolvable by
+        # LocalStack. We are proxying the request to ourself, so replace their custom hostname with
+        # `localhost.localstack.cloud` which also matches the PATH matchers.
+        host_definition = localstack_host(use_hostname_external=True)
+        if host_definition.host != LOCALHOST:
```
bentsku, line 90: I was confused by this at first, but it's because `HOSTNAME_EXTERNAL` is set to `LOCALHOST` if not set, is that right? Now it makes more sense to me.

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+
+class TestS3:
+    @pytest.mark.skipif(
+        condition=config.LEGACY_S3_PROVIDER, reason="Not implemented for legacy provider"
+    )
+    def test_non_us_east_1_location(
+        self, monkeypatch, s3_resource, s3_client, cleanups, assert_host_customisation
+    ):
+        monkeypatch.setattr(config, "LEGACY_S3_PROVIDER", False)
+        monkeypatch.setenv("PROVIDER_OVERRIDE_S3", "asf")
```
bentsku, line 80: This won't work I believe, as the provider will be loaded when one call is going to S3, which might be before this test. I suppose this test will be running only once the new ASF provider will be default. We could also add it manually to the S3 test suite, because we have a custom CI job running the S3 test suite with `PROVIDER_OVERRIDE_S3=asf`. Or I don't know if we can add these tests to this CI job? Maybe @alexrashed could help us here.

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+            bucket = s3_resource.Bucket(bucket_name)
+            bucket.objects.all().delete()
+            bucket.object_versions.all().delete()
+            bucket.delete()
+
+        cleanups.append(cleanup)
+
+        assert_host_customisation(res["Location"], use_hostname_external=True)
+
+    def test_multipart_upload(self, s3_bucket, s3_client, assert_host_customisation):
```
bentsku, line 100: This multipart test made me think on where the `Location` was updated for these responses, and it seems like there's still a use of `HOSTNAME_EXTERNAL` in `s3_listener.py` here: https://github.com/localstack/localstack/blob/4914f4896e86df94e7b815c89815257b6a73c262/localstack/services/s3/s3_listener.py#L1366

Not sure if this needs to be update as well?

bentsku (approved):
LGTM for the S3 part! Just a minor thing with the tests that might be skipped and another usage of `HOSTNAME_EXTERNAL`, but the changes to the new provider look good to me 👍 also tagging @steffyP to keep you in the loop, as you were the one working on the virtual host addressing 😄 
Thanks a lot for tackling this! And it helps us to being to strict with virtual host addressing before v2 🎉

tests/integration/test_network_configuration.py
alexrashed: That's true, providers cannot be changed at runtime once they have been loaded once.
In order to add this test to be executed in the S3 ASF provider tests in CircleCI it has to be added to the test path in the workflow: https://github.com/localstack/localstack/blob/204e605e6fc47331718201724cb2732179ecb772/.circleci/config.yml#L175

simonrw: Thanks. I think I left that line in before I set the test to be skipped if using the legacy provider. Thanks for the tip @alexrashed I can add `tests/integration/test_network_configuration.py::TestS3` to this line and the network tests will run with the new provider, thanks!

simonrw: I'll do the same with the lambda tests as well

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+
+from localstack import config
+from localstack.aws.api.lambda_ import Runtime
+from localstack.testing.aws.lambda_utils import is_new_provider, is_old_provider
+from localstack.utils.files import new_tmp_file, save_file
+from localstack.utils.strings import short_uid
+
+# TODO: how do we test `localstack_hostname` - this variable configures the
+# host that services make requests to when starting up (e.g. opensearch) and
+# they won't start if we override the variable.
```
alexrashed, line 23: This seems to be resolved with the usage of `socket.gethostname` in the `fixtures.py`.

localstack/testing/pytest/fixtures.py
```diff
@@ -1978,3 +1979,54 @@ def factory(**kwargs):
             appsync_client.delete_graphql_api(apiId=api)
         except Exception as e:
             LOG.debug(f"Error cleaning up AppSync API: {api}, {e}")
+
+
+@pytest.fixture
+def assert_host_customisation(monkeypatch):
+    hostname_external = f"external-host-{short_uid()}"
+    localstack_hostname = socket.gethostname()
```
alexrashed, line 1987: I think this deserves a comment to make it understandable why `socket.gethostname` is necessary here.

alexrashed (approved):
The OpenSearch part looks good to me! It's a nice formalization of the current state and shows that the there are currently a lot of different ways how localstack currently creates the URLs returned to the client. It's going to be great (both from a UX and a developer's perspective) to see this getting unified! 🚀

tests/integration/test_network_configuration.py
simonrw: Thanks for pointing this out - yeah that did need an update as well. I've pushed a new commit.

localstack/testing/pytest/fixtures.py
simonrw: Great suggestion, thanks!

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+
+        queue_name = f"queue-{short_uid()}"
+        queue_url = sqs_create_queue(QueueName=queue_name)
+
+        assert_host_customisation(queue_url, use_hostname_external=True)
+        assert queue_name in queue_url
+
+    @pytest.mark.parametrize("external_port", [0, 12345])
+    def test_domain_strategy(
```
baermat, line 168: `SQS_PORT_EXTERNAL` has no effect outside of the `off` strategy, so no need to test it :)

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+
+        queue_name = f"queue-{short_uid()}"
+        queue_url = sqs_create_queue(QueueName=queue_name)
+
+        assert_host_customisation(queue_url, use_localstack_cloud=True)
+        assert queue_name in queue_url
+
+    @pytest.mark.parametrize("external_port", [0, 12345])
+    def test_path_strategy(
```
baermat, line 181: Same as before, can be removed

localstack/testing/pytest/fixtures.py
simonrw: Using `socket.gethostname` works nicely

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+        external_port = 12345
+        monkeypatch.setattr(config, "SQS_ENDPOINT_STRATEGY", "off")
+        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+
+        queue_name = f"queue-{short_uid()}"
+        queue_url = sqs_create_queue(QueueName=queue_name)
+
+        assert_host_customisation(queue_url, use_hostname_external=True)
+        assert queue_name in queue_url
+
```
baermat, line 166: for completeness' sake:
`assert f":{external_port}" in queue_url`
Since this is the one case where the port actually changes due to `SQS_PORT_EXTERNAL`

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+    @pytest.mark.parametrize("external_port", [0, 12345])
+    def test_domain_strategy(
+        self, external_port, monkeypatch, sqs_create_queue, assert_host_customisation
+    ):
+        monkeypatch.setattr(config, "SQS_ENDPOINT_STRATEGY", "domain")
+        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+
```
baermat, line 173: Same as before, can be removed

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+        assert_host_customisation(queue_url, use_localstack_cloud=True)
+        assert queue_name in queue_url
+
+    @pytest.mark.parametrize("external_port", [0, 12345])
+    def test_path_strategy(
+        self, external_port, monkeypatch, sqs_create_queue, assert_host_customisation
+    ):
+        monkeypatch.setattr(config, "SQS_ENDPOINT_STRATEGY", "path")
+        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+
```
baermat, line 186: Same as before, can be removed

baermat (changes requested):
Looks great! Small things remaining which can be fixed super quickly and then it's a go from my side :tada:

tests/integration/test_network_configuration.py
```diff
@@ -0,0 +1,248 @@
...
+                    "AuthType": "NONE",
+                }
+            ),
+            params={"Qualifier": "$LATEST"},
+        )
+        r.raise_for_status()
+
+        function_url = r.json()["FunctionUrl"]
+
+        assert_host_customisation(function_url, use_localstack_cloud=True)
```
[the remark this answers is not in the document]
simonrw: It was another place where the configuration variables were being used. I know it's not implemented in the new provider, but I wanted to be complete.

simonrw:
> Looks great! Small things remaining which can be fixed super quickly and then it's a go from my side tada

Thanks @baermat, I've implemented your changes

baermat (approved):

localstack/services/sqs/models.py
```diff
@@ -249,13 +250,18 @@ def url(self, context: RequestContext) -> str:
...
         elif config.SQS_ENDPOINT_STRATEGY == "path":
             # https?://localhost:4566/queue/us-east-1/00000000000/my-queue (us-east-1)
             host_url = f"{context.request.host_url}queue/{self.region}"
         else:
             if config.SQS_PORT_EXTERNAL:
-                host_url = external_service_url("sqs")
+                host_definition = localstack_host(
+                    use_hostname_external=True, custom_port=config.SQS_PORT_EXTERNAL
+                )
+                host_url = f"{get_protocol()}://{host_definition.host_and_port()}"
```
thrau, line 264: hey! i can see that `SQS_PORT_EXTERNAL` is still in the control path here although we decided to boot it. can we just remove this?

dfangl (approved):
I mostly looked at the lambda parts, LGTM!

localstack/services/sqs/models.py
simonrw: This PR is just matching existing behaviour. I'll get rid of it in a follow up PR targeting the `v2` branch once this is merged :+1:

PR #7785, opened 2023-03-02
add persistence API integration for community services

thrau:
This PR uses the persistence API to integrate community services into the persistence lifecycle. It's "basic" because it doesn't fully add restore functionality for every services.
* added `accept_state_visitor` methods to community services that have an asset directory
* added the `SAVE_STRATEGY` variable used by the persistence plugin to the config
* I was able to add the restore functionality for opensearch (tested manually), but not for others. Those will be part of follow-up PRs once we have a good persistence testing setup.
* I also snuck in the two parameters to the opensearch `create_domain`  method that were missing after the last ASF API update /cc @alexrashed

### DIFF
--- a/localstack/services/opensearch/provider.py
+++ b/localstack/services/opensearch/provider.py
@@ -1,10 +1,13 @@
 import logging
+import os
 import re
 import threading
 from datetime import datetime, timezone
 from random import randint
 from typing import Dict, Optional
+from urllib.parse import urlparse
 
+from localstack import config
 from localstack.aws.accounts import get_aws_account_id
 from localstack.aws.api import RequestContext, handler
 from localstack.aws.api.opensearch import (
@@ -55,6 +58,7 @@ from localstack.aws.api.opensearch import (
     NextToken,
     NodeToNodeEncryptionOptions,
     NodeToNodeEncryptionOptionsStatus,
+    OffPeakWindowOptions,
     OpensearchApi,
     OpenSearchPartitionInstanceType,
     OptionState,
@@ -66,6 +70,7 @@ from localstack.aws.api.opensearch import (
     ServiceSoftwareOptions,
     SnapshotOptions,
     SnapshotOptionsStatus,
+    SoftwareUpdateOptions,
     StringList,
     TagList,
     TLSSecurityPolicy,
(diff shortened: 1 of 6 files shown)


localstack/services/opensearch/provider.py
```diff
@@ -400,6 +406,51 @@ def get_store(region: str = None) -> OpenSearchStore:
...
+                for domain_name, domain_status in store.opensearch_domains.items():
+                    domain_key = DomainKey(domain_name, region, account_id)
+                    if cluster_manager().get(domain_key.arn):
+                        # cluster already restored in previous call to on_after_state_load
+                        continue
+
+                    LOG.info(f"Restoring domain {domain_name} in region {region}.")
+                    try:
+                        preferred_port = None
+                        if config.OPENSEARCH_ENDPOINT_STRATEGY == "port":
```
dominikschubert, line 427: That's an interesting problem actually that might bite us in a few other places as well when a feature flag changes between different persistence-enabled runs :thinking:

localstack/services/opensearch/provider.py
```diff
@@ -400,6 +406,51 @@ def get_store(region: str = None) -> OpenSearchStore:
...
+                            if "Endpoint" in domain_status:
+                                preferred_port = urlparse(
+                                    f"http://{domain_status['Endpoint']}"
+                                ).port
+
+                        engine_version = domain_status.get("EngineVersion")
+                        domain_endpoint_options = domain_status.get("DomainEndpointOptions", {})
+                        create_cluster(
+                            domain_key, engine_version, domain_endpoint_options, preferred_port
+                        )
```
dominikschubert, line 438: `preferred_port` is used as the 4th positional parameter (instead of the 5th) here which should actually be `security_options`.

```suggestion
                        security_options = SecurityOptions.from_input(domain_status.get("AdvancedSecurityOptions"))
                        create_cluster(
                            domain_key, engine_version, domain_endpoint_options, security_options, preferred_port
                        )
```

function definition:

```python
def create_cluster(
    domain_key: DomainKey,
    engine_version: str,
    domain_endpoint_options: Optional[DomainEndpointOptions],
    security_options: Optional[SecurityOptions],
    preferred_port: Optional[int] = None,
):
```

dominikschubert (changes requested):
Tiny issue with the `create_cluster` call, otherwise LGTM! :rocket:

localstack/services/opensearch/provider.py
thrau: indeed, already had to disable ports in rds snapshot tests https://github.com/localstack/localstack-ext/pull/1397/files#diff-9554b233d544d2d605d0a8067d419a1adc3ee3dfff6500f6524605217b7aaf65R7-R9

localstack/services/opensearch/provider.py
thrau: nice catch

PR #7789, opened 2023-03-02
add runtime hook that runs on pro infra startup

thrau:
Adds a runtime hook used in pro and plugins that is only run when pro is activated. This makes it easier to hook in functionality that's conditionally after license activation.


alexrashed (approved):
LGTM! 🚀

PR #7793, opened 2023-03-03
Add scenario tests with Fundamental level for Stepfunctions

dominikschubert:
Adds tests based on workflow samples from [serverlessland.com](https://serverlessland.com/workflows?type=Standard&simplicity=1+-+Fundamental)

- [x] path-based-on-data
- [x] wait-for-callback
- [x] step-functions-calling-api-gateway

I've had to drop a lot of the samples since they were a poor fit (either too complicated or too simple and some of them didn't even properly allow testing all paths defined in the workflow definition). The rest were all simplified/fixed.

All tests are designed to be able to take a single CloudFormation template as an input with lambda code & workflow definitions being included in-line in the template.

### DIFF
--- a/tests/integration/stepfunctions/v2/scenarios/templates/path-based-on-data.yaml
+++ b/tests/integration/stepfunctions/v2/scenarios/templates/path-based-on-data.yaml
@@ -1,0 +1,108 @@
+Resources:
+  statemachineRole52044F93:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Statement:
+          - Action: sts:AssumeRole
+            Effect: Allow
+            Principal:
+              Service:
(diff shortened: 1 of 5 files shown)


localstack/services/opensearch/provider.py
```diff
@@ -157,6 +157,9 @@ def create_cluster(
         preferred_port=preferred_port,
     )
 
+    for k, v in store.opensearch_domains.items():
+        v["AdvancedSecurityOptions"]
+
```
alexrashed, line 162: Since this isn't really doing anything, this line might be a leftover from testing?

dominikschubert: Yep, sorry for the unnecessary ping!

MEPalma (approved):

PR #7794, opened 2023-03-03
Update apt repositories when installing system dependencies

simonrw:
We have to keep the repositories updated or else we potentially get network errors.

Also because of the updated metadata we have to install a later version of postgres. This is unfortunate because it's now a moving target and likely to cause more breakages in the future, but leaving unpinned causes a dependency conflict with postgres-plpython3-14.

### DIFF
--- a/.github/workflows/asf-updates.yml
+++ b/.github/workflows/asf-updates.yml
@@ -16,6 +16,7 @@ jobs:
 
       - name: Set up system wide dependencies
         run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Thanks for jumping on this, and thanks for also updating the ASF update action (this would have hit us on Monday morning otherwise 😛)! 🚀 
Feel free to merge as soon as the pro-integration pipeline goes past the installation (which currently fails on every run).

simonrw:
Packages have been installed so I'll merge

PR #7803, opened 2023-03-04
refactor SNS store to facilitate lookups

bentsku:
As we're currently working on persistence, I'd like to refactor the SNS store to facilitate lookup and have its data structures more aligned with how we are using them in the provider and to how AWS stores them.

Added some validations where I had questions on how AWS would behave.

### DIFF
--- a/localstack/services/sns/models.py
+++ b/localstack/services/sns/models.py
@@ -88,11 +88,13 @@ class SnsSubscription(TypedDict):
 
 
 class SnsStore(BaseStore):
-    # maps topic ARN to topic's subscriptions
-    sns_subscriptions: Dict[str, List[SnsSubscription]] = LocalAttribute(default=dict)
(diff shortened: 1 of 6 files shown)


thrau (approved):
nice cleanup! :broom: LGTM

bentsku:
> nice cleanup! :broom: LGTM
> 
> 
> 
> but looks like pro tests are failing

Yes, looks like a lot of tests are failing 😅 I changed some logic and think I went a bit too much on some checks, will add some tests and validate this with AWS.

thrau:
@bentsku are we merging this?

bentsku:
@thrau Yes, I think so? If we're breaking persistence, now is the way to go. We can also wait a bit to be closer to `v2`, but it's something I wanted to be out before v2, that'd be great.

PR #7816, opened 2023-03-07
Add DELETE to the _localstack/ses/ endpoint

### LINKED ISSUE #7491: feature request: implement DELETE /_localstack/ses endpoint

### Feature description

Provide an URL where I can delete stored SES emails.
My motivation is to write integration tests much easier, if I could delete the stored emails from the localstack before each test case. 

I use the docker image to run localstack. Mount the `/var/lib/localstack` folder to the host and delete the files from the `tmp/state/ses` is cumbersome because of the file permission issue of docker.

I think the simplest way to achieve the empty SES emails state is to implement a new endpoint in a same way as to the GET /_localstack/ses has been implemented:

**path:** /_localstack/ses
**method:** DELETE

### 🧑‍💻 Implementation

I am not a python developer, but I think one solution is to extend the [SesServiceApiResource](https://github.com/localstack/localstack/blob/590de96b3dd65e349709300b668e70f1f9664e99/localstack/services/ses/provider.py#L121) with an `on_delete` method.

Open question for me is either delete the `/var/lib/localstack/tmp/state/ses` folder or just delete the files of the `ses` folder.
The folder does not exist when the container starts. It is created when the first email received and saved by the service.
I think both solutions have the same performance impact, so the real question is what is preferred by the maintainers.

As I mentioned I am not familiar with python ecosystem but I am happy to help implementing the feature if you accept the request.

### Anything else?

_No response_

justinwyer:
Resolves #7491 

Added the following endpoints:
- `DELETE _localstack/ses` requested by the issue
- `DELETE _localstack/ses?id=<message_id>` [requested in a comment](https://github.com/localstack/localstack/issues/7491#issuecomment-1414126917) 
- `GET       _localstack/ses?id=<message_id>` added to match the DELETE by id

### DIFF
--- a/localstack/services/ses/provider.py
+++ b/localstack/services/ses/provider.py
@@ -125,27 +125,28 @@ class SesServiceApiResource:
     This is registered as a LocalStack internal HTTP resource.
 
     This endpoint accepts:
-    - GET param `email`: filter for `source` field in SES message
+    - GET param `id`: filter for `id` field in SES message
+    - GET param `email`: filter for `source` field in SES message, when `id` filter is specified then filters on both
     """
 
-    def on_get(self, request, message_id=None):
-        if message_id is not None:
-            return EMAILS[message_id]
-
+    def on_get(self, request):
+        filter_id = request.args.get("id")
         filter_source = request.args.get("email")
         messages = []
 
         for msg in EMAILS.values():
-            if filter_source in (msg.get("Source"), None, ""):
-                messages.append(msg)
+            if filter_id in (msg.get("Id"), None, ""):
+                if filter_source in (msg.get("Source"), None, ""):
+                    messages.append(msg)
 
         return {
             "messages": messages,
         }
 
-    def on_delete(self, request, message_id=None):
-        if message_id is not None:
-            del EMAILS[message_id]
+    def on_delete(self, request):
+        filter_id = request.args.get("id")
+        if filter_id is not None:
+            del EMAILS[filter_id]
         else:
             EMAILS.clear()
         return Response(status=204)
@@ -161,12 +162,12 @@ def register_ses_api_resource():
         ses_service_api_resource = SesServiceApiResource()
         get_internal_apis().add(
             Resource(
-                "/_localstack/ses/<message_id>",
+                "/_localstack/ses",
                 DeprecatedResource(
                     ses_service_api_resource,
-                    previous_path="/_localstack/ses/<message_id>",
+                    previous_path="/_localstack/ses",
                     deprecation_version="1.4.0",
-                    new_path="/_aws/ses/<message_id>",
+                    new_path="/_aws/ses",
                 ),
             )
         )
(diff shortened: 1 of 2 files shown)


justinwyer:

justinwyer:
As part of the testing I was using the `_aws/ses` endpoint but had to switch to `_localstack/ses` as the `message_id` parameter in the route `_localstack/ses/<message_id>` parameter was not working with the `_aws` prefix. If you can shed some light as to why this might be I would be happy to fix that.

justinwyer:
I am not sure what Circle CI is upset about exactly, perhaps someone can rerun the workflow?

localstack/services/ses/provider.py
```diff
@@ -128,7 +128,10 @@ class SesServiceApiResource:
     - GET param `email`: filter for `source` field in SES message
     """
 
-    def on_get(self, request):
+    def on_get(self, request, message_id=None):
+        if message_id is not None:
+            return EMAILS[message_id]
```
viren-nadkarni, line 133: Let's keep the response schema the same as before, whether or not the `message_id` filter is specified.

```
{
  "messages": [
    {<the message>}
  ]
}
```

localstack/services/ses/provider.py
```diff
@@ -128,7 +128,10 @@ class SesServiceApiResource:
     - GET param `email`: filter for `source` field in SES message
     """
 
-    def on_get(self, request):
+    def on_get(self, request, message_id=None):
```
viren-nadkarni, line 131: I would make `message_id` a query parameter which is more backward compatible.

Could we also update the docstring? Also the behaviour when both `message_id` and `email` are provided.

viren-nadkarni (changes requested):

tests/integration/test_ses.py
```diff
@@ -181,9 +181,12 @@ def _read_message_from_filesystem(message_id: str) -> dict:
         assert "Body" not in contents2
 
         # Ensure all sent messages can be retrieved using the API endpoint
-        emails_url = config.get_edge_url() + EMAILS_ENDPOINT
-        api_contents = requests.get(emails_url).json()
-        api_contents = {msg["Id"]: msg for msg in api_contents["messages"]}
+        emails_url = config.get_edge_url() + "/_localstack/ses/"
```
viren-nadkarni, line 184:
```suggestion
        emails_url = config.get_edge_url() + EMAILS_ENDPOINT
```

tests/integration/test_ses.py
```diff
@@ -181,9 +181,12 @@ def _read_message_from_filesystem(message_id: str) -> dict:
         assert "Body" not in contents2
 
         # Ensure all sent messages can be retrieved using the API endpoint
-        emails_url = config.get_edge_url() + EMAILS_ENDPOINT
```
viren-nadkarni, line 184: We need to test the 'get all messages' functionality as before.

viren-nadkarni (changes requested):
Thanks for addressing my comments. Just a few minor points and we can merge this :+1:

tests/integration/test_ses.py
justinwyer: This is still tested in the `test_send_templated_email_can_retrospect` above.

viren-nadkarni (approved):

Issue #7158, opened 2022-11-10
[KMS] For Sign / Verify - confirm that it is necessary and implement digest calculation based on SigningAlgorithm.


taras-kobernyk-localstack:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Currently, we always [calculate](https://github.com/localstack/localstack/blob/256b0de1f1a6085d3994e8d1322c763962ded2f0/localstack/services/kms/models.py#L229-L230) SHA256 digests for messages with `RAW` [MessageType](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#API_Sign_RequestSyntax).

It is not clear if this is compatible with AWS. There are some reports like [this one](https://github.com/localstack/localstack/issues/6815#issuecomment-1309882776) about our implementation of Sign / Verify not working properly. Could be due to these digests.

What has to be done:
1. Should try to figure out how AWS calculates digests, whether the same algorithm is always used, or if digests are based on [SigningAlgorithm](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#API_Sign_RequestSyntax). For this we can try importing the same key material into both AWS and LocalStack, use the resulting keys to sign the same message and then compare the signatures. If that works. If not - we can try just creating a key in AWS, sign a message with AWS, export the public key and then go through different digest algorithms in LocalStack, trying to verify the signed message with the obtained public key.
2. If the issue is really in our implementation of digests - should fix the digest logic.

paullallier:
I've added a comment here which might relate to this:
https://github.com/localstack/localstack/issues/6815#issuecomment-1458573903

sannya-singal:
Thanks @taras-kobernyk-localstack for reporting the issue. We have implemented changes in #7878 that should resolve the issue. Could you please update to the `latest` LocalStack container image and let us know whether your issue is resolved? Thank you!

Issue #6815, opened 2022-09-04
bug: KMS provides invalid signature by ignoring MessageType


dissoupov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

It seems that Sign operation ignores MessageType. 
As per AWS documentation the MessageType can be "DIGEST", that means the digest is already computed and just needs to be signed. 
https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html

Localstack ignores this parameter:
https://github.com/localstack/localstack/blob/5fafd1b68f31f5948ec88b00a85966319c2ce8c3/localstack/services/kms/provider.py#L332

In GoLang the `Sign` interface takes the digest, not the raw message.
`Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts)`
Therefore Sign/Verify operations in Go fail by using localstack.

### Expected Behavior

Use `MessageType` parameter and do not compute the digest when MessageType==`DIGEST`

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Use Sign/Verify in GoLang

### Environment

```markdown
- OS: ubuntu
- LocalStack: latest
- GoLang 1.19
```

taras-kobernyk-localstack:
It doesn't look like we generate digests, just sign whatever we get.
And the AWS doc page you mention says that digests are only generated for messages longer than 4K: "To sign a larger message, generate a hash digest of the message, and then provide the hash digest in the Message parameter".

I am not familiar with the implementation in Go. But as a workaround, I imagine, it should work if you generate digests and then use them both in Go sign/verify calls and in your calls to LocalStack KMS.

As for a proper fix - we have a new version of our KMS code coming. Hopefully in a few days. After that I will add the logic for the MessageType parameter. Will update the ticket. And if you would be able to check afterwards if it works with Go - would be great!

dissoupov:
My existing code works with AWS KMS and with nsmithuk/local-kms
and fails with localstack.
I will gladly test your new implementation.
Thanks,
Denis

taras-kobernyk-localstack:
Hi Denis,

Apologies for the delay. The change is finally there. If you update your docker image for LocalStack, you should have it.

Let me know if it works for your setup. If it doesn't - which is quite a possibility, as I can't understand why it wasn't working with our old setup - could you please provide me with some simple code that uses these APIs, so I could run the code on my side and debug the issue further without bothering you?

Thanks,
Taras

[...]

PR #7817, opened 2023-03-07
KMS sign: Avoid rehashing the already hashed message

paullallier:
This may be an improvement on issue #6815 - hopefully.
May also relate to issue #7158.

### DIFF
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -14,7 +14,7 @@ from typing import Dict, List
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization as crypto_serialization
-from cryptography.hazmat.primitives.asymmetric import ec, padding, utils, rsa
+from cryptography.hazmat.primitives.asymmetric import ec, padding, rsa, utils
 
 from localstack.aws.accounts import get_aws_account_id
 from localstack.aws.api import CommonServiceException
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -14,7 +14,7 @@ from typing import Dict, List
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization as crypto_serialization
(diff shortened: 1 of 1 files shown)


paullallier:

paullallier:
Not sure why the CI tests fails on something that doesn't seem KMS related...  But is there a CI configuration error more generally?

As far as I can see it's not run the tests in tests/integration/test_kms.py - or several other files in that folder.  Is that intentional?  (I was trying to double-check that my changes had been tested by the existing test.)

paullallier:
I believe my fix is correct though - I've manually applied the change to the localstack docker container I'm running, restarted python, and my JWTs now verify after being signed by localstack KMS in 'RAW' messageType mode - as they do with AWS.

laszlovandenhoek:
I can confirm that running your branch locally (using `python -m localstack.cli.main start --host`) now allows me to create (externally) valid signatures using the `Sign` operation. This is a must-have for verifying JWT tokens. Thanks a lot for finding and fixing this problem just in time for me :) Now let's get this merged ASAP ;)

laszlovandenhoek:
recheck

paullallier:
Hmm.  That test quite failed again, but on a different test.  I guess it's possible that S3 and firehose (which I don't know) both use KMS sign internally for something, and my change has broken it?

viren-nadkarni:
The failures don't seem related to your change.

Thanks for contributing @paullallier !

viren-nadkarni:
@paullallier: We need to add some tests. I'm trying to reproduce the issue so that we can compare the behaviour against AWS. Since you mentioned you're not fluent with Python, I'm happy to take this over.

paullallier:
@viren-nadkarni: Sounds good to me.  There's some information about it here, which might be useful if you haven't seen it already.  https://github.com/localstack/localstack/issues/6815#issuecomment-1458573903

Since we can't extract the private key from AWS, I was struggling a bit to come up with a way to do a like-for-like test.  And I've, personally, only really tested it with EC256 keys so far, since that's what I'm interested in.

viren-nadkarni:
Closing in favour of https://github.com/localstack/localstack/pull/7878

Thanks again @paullallier for the findings!

PR #7821, opened 2023-03-08
fix forward slash validation in S3 pre-signed url

bentsku:
This PR mostly add a test to verify behaviour raised in a community slack issue. The PHP SDK allows the user to specify a bucket with a slash in the name while generating a pre-signed URL. However, this is not allowed by AWS, and we would not raise the proper exception (and there was a small issue in our calculation, as we were using the sanitized path to create the `CanonicalRequest` used for the signature). 

I just added some proper exceptions in that case for the test to pass like it would against AWS.


macnev2013 (approved):
Looks Good 👍🏻

PR #7825, opened 2023-03-08
refactor kinesis state lifecycle hook

thrau:
This PR refactors our Kinesis provider to use the new state lifecycle hooks.

* Encapsulate starting and stopping of account-specific kinesis-mock backends into `KinesisServerManager`
* Implements the state lifecycle hook methods to properly shut down the managed kinesis backends
  * `on_before_state_load`: before a state load, we only need to shut down already running kinesis backends. the kinesis backends are re-started lazily the next time they are requested.
  * same with `on_before_state_reset`
  * `on_before_stop` just shuts them down

tested manually with the new persistence plugin using localstack-ext


localstack/services/kinesis/provider.py
```diff
@@ -48,19 +49,35 @@ def find_stream_for_consumer(consumer_arn):
...
+        # no need to restart servers, since that happens lazily in `server_manager.get_server_for_account`.
+        self.server_manager.shutdown_all()
+
+    def on_before_state_reset(self):
+        self.server_manager.shutdown_all()
+
+    def on_before_stop(self):
+        self.server_manager.shutdown_all()
 
     def get_forward_url(self):
```
giograno, line 72:
```suggestion
    def get_forward_url(self) -> str: 
```

localstack/services/kinesis/provider.py
```diff
@@ -48,19 +49,35 @@ def find_stream_for_consumer(consumer_arn):
...
         """Return the URL of the backend Kinesis server to forward requests to"""
         account_id = get_aws_account_id()
-        starter.start_kinesis(account_id=account_id)
-        return f"http://{LOCALHOST}:{starter.get_server(account_id).port}"
+        server = self.server_manager.get_server_for_account(account_id)
+        return f"http://{LOCALHOST}:{server.port}"
+
+    @staticmethod
+    def get_store(account_id: str, region_name: str) -> KinesisStore:
```
giograno, line 79: Had a quick look at the kinesis stores. Are we using them somewhere?

giograno (approved):

localstack/services/kinesis/provider.py
thrau: good question, i didn't actually look into it. but may be useful to keep it

PR #7827, opened 2023-03-09
DDB - persistence plugin and server rework

giograno:
This PR refactors the DynamoDB provider to use the new lifecycle hooks.
In addition, it refactors the singleton instance of the server into the provider itself. Due to these changes, the methods to start the server are not accessible at module level anymore. See related [ext PR](https://github.com/localstack/localstack-ext/pull/1435).


thrau (approved):
looks great! tested manually with the new persistence plugin and works fine

PR #7828, opened 2023-03-09
wip: apigw parity work Model resource

bentsku:
wip: working on adjusting parity for `Model` resource.
I took the opportunity to move the `Model` resource out of moto and into our own store, as there was no dependency between any other resources, only CRUD API was implemented in moto. 
Not sure if we used the data directly from `moto_backends` or with internal client call, so the CI will tell. 

still missing `UpdateModel` implementation and validation


localstack/services/apigateway/provider.py
```diff
@@ -1192,6 +1198,121 @@ def filter_function(item):
...
+            if operation.get("op") != "replace":
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op 'add'. Please choose supported operations"
+                )
+            if path not in ("/schema", "/description"):
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op 'replace'. Must be one of: [/description, /schema]"
+                )
+
+            key = path.strip("/")
```
calvernaz, line 1297: nit: we are striping left and right slashes (`/`), but most likely the intent is just removing the leading slash, otherwise the check above `path in ("/schema" ...` would not work if it had a trailing slash.

localstack/services/apigateway/provider.py
```diff
@@ -1192,6 +1198,121 @@ def filter_function(item):
...
+            value = operation.get("value", "")
```
calvernaz, line 1298: do we need the default, we'll bail anyway.

calvernaz (approved):
LGTM, nice set of changes. It scares be a bit that amount of work to pull the model into Localstack, given the amount of work just for Model is this much 😅

localstack/services/apigateway/provider.py
bentsku: Right, yes, I'll just remove the first character of the string. Thanks! 👍

localstack/services/apigateway/provider.py
bentsku: The default was for the `description` field, as you can set it back to an empty string even though you can pass `null` / `None` but wanted it to stay a string. But as seen later on, if the description is an empty string or nothing, AWS won't return the field at all, so can remove the default I suppose.

bentsku:
> LGTM, nice set of changes. It scares be a bit that amount of work to pull the model into LocalStack, given the amount of work just for Model is this much 😅

Thanks! I guess pulling it into LocalStack looks like a lot of work because there was zero validation done on Moto side (so no work at all 😄), and now we have quite a good amount I'd say. Some resources will need even more work, this one wasn't so bad really, but then we can reuse some mechanisms we've encountered already.
And there's even more coming, with #7872 (this will be used by `RequestValidator` as well).

PR #7829, opened 2023-03-09
Fix error with threads not having _target attributes

dfangl:
## Motivation
Master build is currently red due to some threads not having the _target attribute, and the thread leakage detection therefore failing.

## Changes
Check if a thread has the _target attribute before accessing it, otherwise just print `None`.


dominikschubert (approved):
LGTM :+1:

PR #7831, opened 2023-03-09
Support mapping random host ports to container ports

orgito:
Ommiting the host port from the a -p flag added to LAMBDA_DOCKER_FLAGS will cause docker to map a random host port to the container port.

An example use case is to have multiple Node.js Lambdas having the Inspector enabled and accessible from the host on different ports.

For example:
LAMBDA_DOCKER_FLAGS: -e NODE_OPTIONS=--inspect=0.0.0.0:9230 -p :9230

**Please refer to the contribution guidelines in the README when submitting PRs.**


orgito:

PR #7832, opened 2023-03-09
Fix multiple AWS::Lambda::Permission for single function

dominikschubert:
A single lambda function can be associated with multiple `AWS::Lambda::Permission` resources. This was fundamentally incompatible with how the previous fetch_state "worked'. 

## Changes
* Added a test for adding multiple AWS::Lambda::Permission resources for a single function
* Fixed `fetch_state` & physical resource id for `AWS::Lambda::Permission`
* Fixed overwriting `Code` properties in `AWS::Lambda::Function` which lead to unnecessary updates in certain cases


pinzon (approved):
LGTM

PR #7835, opened 2023-03-10
Allow privileged mode to be passed as additional flag

viren-nadkarni:
Allows the `--privileged` flag to be passed in `additional_flags` when running Docker containers.


thrau (approved):
LGTM! but i'll let one of the codeowners sign off :-)

dfangl (approved):
LGTM! Will just ping @joe4dev here, as I think he is reworking this method. Feel free to merge though.

PR #7837, opened 2023-03-10
Remove extra slash in SQS path strategy

baermat:
This PR removes a duplicate slash when creating the queue url under the `path` endpoint strategy


thrau (approved):
LGTM! nice catch

PR #7843, opened 2023-03-13
Refactor API GW tests, assert correct JSON content-type for DDB integration

whummer:
Refactor API Gateway tests, assert correct JSON content-type for DynamoDB integration. 

Summary of changes:
* move the remaining API GW tests into the `tests/integration/apigateway` module
* move IAM policies required by fixtures in `conftest.py` into that same file
* minor change to initialize `response` in the invocation context (to avoid `AttributeError`)
* fix DynamoDB integration to return proper content-type (`application/json`)
* extend `test_rest_api_to_dynamodb_integration` snapshot test to also capture the response headers (to verify that the content-type matches)
* use `echo_http_server` fixture instead of making `httpbin.org` requests (should fix some flaky tests with 504 timeouts) - some TODOs still remaining in the codebase

Side note: As also mentioned in the PR comment below, in a future iteration we should unify the logic/interface of integration response handling. 👍  /cc @calvernaz @bentsku


tests/integration/apigateway/test_apigateway_api.py
bentsku, line 1: Would it possible to leave this one untouched, as it helps me separate new tests from older ones?
Maybe moving them into a new file with a different name?
It would also prevents from rebasing from every PR I'm going to merge 😅

whummer: Sure, no worries - thanks for the pointer @bentsku ! Let's coordinate - happy to wait until your PRs are merged, then I can rebase the changes in this PR 👍

bentsku: Thanks 🙏 I think if you move the older tests in a different files, it would not conflict so much anymore.
There might 2 (one ready #7828 and one draft #7846) touching `invocations.py`, but I can rebase as well, no problem with that.

I'm really glad to see some structures in the tests by the way, thanks a lot for tackling this 🙏 I have a feeling we might even need to split `test_apigateway_api.py` at some point as it's already more than 1000 lines.

whummer: Done @bentsku  - reorganized the tests into a new `test_apigateway_extended.py` file. The name is not great, but we can improve in the next iterations. 👍 (left a TODO comment in the file..)

localstack/services/apigateway/integration.py
```diff
@@ -449,12 +449,14 @@ def invoke(self, invocation_context: ApiInvocationContext):
 
         # apply response templates
         response_content = json.dumps(remove_attributes(response, ["ResponseMetadata"]))
-        response_obj = requests_response(content=response_content)
+        invocation_context.response = response_obj = requests_response(content=response_content)
```
whummer, line 452: Note: In one of the next refactoring iterations, we should unify the response handling for API GW integrations. Most of the `BackendIntegration` implementation are returning a `Response` from `invoke(..)` (even though it is not defined in the interface), and some of the implementations also update the `invocation_context.response` property on the context. 

We should think about how we can unify this (either having the response only on the context, or only returning it from `invoke`), to have a clear interface and make the request processing less ambiguous. 👍

tests/integration/cloudformation/resources/test_apigateway.py
```diff
@@ -99,14 +99,12 @@ def test_cfn_apigateway_aws_integration(
...
-@pytest.mark.skip_offline
 @pytest.mark.aws_validated
-def test_cfn_apigateway_swagger_import(deploy_cfn_template, apigateway_client):
+def test_cfn_apigateway_swagger_import(deploy_cfn_template, apigateway_client, echo_http_server):
     api_name = f"rest-api-{short_uid()}"
-    int_uri = "http://httpbin.org/post"
     deploy_cfn_template(
         template=TEST_TEMPLATE_1,
-        parameters={"ApiName": api_name, "IntegrationUri": int_uri},
+        parameters={"ApiName": api_name, "IntegrationUri": f"{echo_http_server}/post"},
```
bentsku, line 107: I like that much better that we're using the `http_server` fixture for our tests, it will make the tests much less flaky. However, we won't be able to test them anymore against AWS this way, or it would need a bit more setup. Should we leave some comment on how to AWS validate the test, maybe replacing manually the `int_uri` to the `httpbin` one?

localstack/services/apigateway/integration.py
```diff
@@ -449,12 +449,14 @@ def invoke(self, invocation_context: ApiInvocationContext):
...
         response = self.response_templates.render(invocation_context, response=response_obj)
 
         # construct final response
-        response = requests_response(response)
-        invocation_context.response = response
+        # TODO: set response header based on response templates
+        headers = {HEADER_CONTENT_TYPE: APPLICATION_JSON}
+        response = requests_response(response, headers=headers)
```
bentsku, line 458: So, if I understood well, for now, we're only returning JSON, so we're setting the headers. LGTM 👍 
But just a question, we now set the `response` before being rendered in the template to the `invocation_context`, whereas before we would set it after, is that wanted?

bentsku (commented):
LGTM! Just one question and a comment about using `httpbin` to validate the tests against AWS. Thanks a lot for this big refactoring and moving files, the structure will be much cleaner, hope we can split it further with time! 🎉

bentsku:
Might still need to update the pro tests here:
https://github.com/localstack/localstack/blob/31c28250879581150f57fd0dc3fcd4ff792cbdf1/.github/workflows/pro-integration.yml#L207-L222

localstack/services/apigateway/integration.py
calvernaz: Agreed, if the integrations are self-contained as they're leaning to im not sure we will have the `invocation_context.response` anymore because all the work is done, meaning the response from the integration and the response templating is done. Food for thought, I guess

localstack/services/apigateway/integration.py
calvernaz: I don't think we need it after, once we move all the templating inside the integrations, that's why we put the response inside the invocation_context, but we should agree on this, whatever it is.

calvernaz (approved):
LGTM, the separation is helpful and also good thinking around the responses - another PoV on it is apply the response templates on the integration response but leave the HTTP layer outside of the integrations, but would require some sort of data structure that contains the response dictionary, the content type, etc, so we can build an HTTP response after.

localstack/services/apigateway/integration.py
whummer: Great points. In my view, this is exactly a symptom of the current duality of response handling. In future iterations, we'll need the response details at some point of processing the integration requests (e.g., response mapping templates may contain variables like `"$context.responseLength"`) - agree with Cesar what we should agree on one approach and unify. 

Slightly updated the PR and removed the context.response assignment to address this point for now 👍

tests/integration/cloudformation/resources/test_apigateway.py
whummer: Done - added a `echo_http_server_post` fixture that can handle POST requests for both, local testing and AWS parity tests.

PR #7845, opened 2023-03-13
update s3 tests location in  CODEOWNERS

bentsku:
Just realised I wasn't tagged in a test modification in S3, so as seen with @alexrashed, updating it.

Also using a fixture in the s3 test to not leave a dangling S3 bucket.


alexrashed (approved):
LGTM! 🚀

PR #7846, opened 2023-03-13
fix validator and model check during invocation

bentsku:
We had a wrong assumption when using a `RequestValidator` and validating the body of a request. If a method did not have a `requestModel` set, we would reject the request. However, AWS uses the default `Empty` model. We now follow the same logic.

We also did not have validation for validators before, and would return an empty response if no validators existed with that ID. Now, in parity with AWS, it raises an exception instead of returning `None`.
However, that case should not happen because it's not possible to delete an existing `RequestValidator` if it's used by a `Method` in AWS. I'll try to implement this behaviour in following PR. 

We needed to change the code in `invocations.py` to reflect the change when checking if a request is valid.

note: created a `test_apigateway_common`, where we could test "integrations", or how different RestAPI resources work together, on a level above the CRUD one. If you have a better name than "common", I'm all ears 😄 

_fixes in part #7842_

### DIFF
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -10,6 +10,7 @@ from localstack.constants import APPLICATION_JSON
 from localstack.services.apigateway import helpers
 from localstack.services.apigateway.context import ApiInvocationContext
 from localstack.services.apigateway.helpers import (
+    EMPTY_MODEL,
     extract_path_params,
     extract_query_string_params,
     get_cors_response,
@@ -85,9 +86,9 @@ class RequestValidator:
         # if there's no model to validate the body, use the Empty model
         # https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-apigateway.EmptyModel.html
         if not (request_models := resource.get("requestModels")):
-            schema_name = "Empty"
+            schema_name = EMPTY_MODEL
         else:
-            schema_name = request_models.get(APPLICATION_JSON, "Empty")
+            schema_name = request_models.get(APPLICATION_JSON, EMPTY_MODEL)
 
         try:
             model = self.apigateway_client.get_model(
@@ -95,7 +96,6 @@ class RequestValidator:
                 modelName=schema_name,
             )
         except ClientError as e:
-            # This should not happen, as we validate that we cannot delete a Model if it's used
             LOG.exception("An exception occurred while trying to validate the request: %s", e)
             return False
 
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -102,7 +102,7 @@ class RequestValidator:
             validate(instance=json.loads(self.context.data), schema=json.loads(model["schema"]))
             return True
         except ValidationError as e:
-            LOG.warning("failed to validate request body", e)
+            LOG.warning("failed to validate request body %s", e)
             return False
 
     # TODO implement parameters and headers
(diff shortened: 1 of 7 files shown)


localstack/services/apigateway/invocations.py
```diff
@@ -78,25 +82,32 @@ def is_request_valid(self) -> bool:
...
+            schema_name = request_models.get(APPLICATION_JSON, "Empty")
 
-        schema_name = resource["requestModels"].get(APPLICATION_JSON)
         try:
             model = self.apigateway_client.get_model(
                 restApiId=self.context.api_id,
                 modelName=schema_name,
             )
         except ClientError as e:
+            # This should not happen, as we validate that we cannot delete a Model if it's used
```
whummer, line 98: I'm always a bit wary of "should not happen" comments - should we simply remove the `try`/`except` handling in this case, if we think it is not a valid code path?

localstack/services/apigateway/provider.py
```diff
@@ -513,6 +513,11 @@ def update_method(
                 continue
 
             elif path == "/requestValidatorId" and value not in rest_api.validators:
+                if not value:
+                    # you can remove a requestValidator by passing an empty string as a value
```
whummer, line 517: A wonderful demonstration of the weirdness of API GW patch operations! 😄

tests/integration/apigateway/test_apigateway_common.py
```diff
@@ -0,0 +1,234 @@
...
+        apigateway_client.put_method(
+            restApiId=api_id,
+            resourceId=resource_id,
+            httpMethod="POST",
+            authorizationType="NONE",
+            requestValidatorId=validator_id,
+            requestParameters={"method.request.path.test": True},
+        )
+
+        apigateway_client.put_integration(
```
whummer, line 93: nit: for simple integrations, we could try leveraging and extending the `create_rest_api_with_integration` fixture. It currently only supports `dynamodb` and `kinesis` integrations, but Lambda would be a great addition.. 👍

whummer (approved):
Looks great - as usual, kudos for adding more snapshots to our test suite @bentsku ! 🚀 Added a few minor comments/questions, nothing that should block the merge, though.. 👍

localstack/services/apigateway/invocations.py
```diff
@@ -78,25 +82,32 @@ def is_request_valid(self) -> bool:
         return True
 
     def validate_body(self, resource):
-        # we need a model to validate the body
-        if "requestModels" not in resource or not resource["requestModels"]:
-            return False
+        # if there's no model to validate the body, use the Empty model
+        # https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-apigateway.EmptyModel.html
+        if not (request_models := resource.get("requestModels")):
+            schema_name = "Empty"
```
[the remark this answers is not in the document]
bentsku: Good idea, will do! 👍

localstack/services/apigateway/invocations.py
bentsku: Yes, I think it can happen for now until #7872 is merged. I'd be for letting it be there, I'll mark it with a TODO instead, until we're sure this code path can't happen anymore.

tests/integration/apigateway/test_apigateway_common.py
bentsku: Makes sense! I need to take a look at it, and how we can customise it. I guess we can always create simple integrations and use `update_*` afterwards, for example the `requestParameters` in that case. There are so many parameters possible.
I'll take a good look a those once I'm starting with integrations. This PR was a bit outside my "normal" work for now, as I was fixing the change in logic now that we are raising exceptions for non-existing resources. 😄

Issue #7847, opened 2023-03-13
S3 EventBridge event triggers lambdas that do not subscribe to said event


jmschibsted:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have created a setup that deploys the following infrastructure:
The bucket
The eventbridge rule
The lambda

"The lambda" uploads a random file to "The bucket" on a scheduled event "The eventbridge rule". 

This works nicely as long as I do not configure on eventbridge events on the S3 bucket.

When configured on "The lambda" is triggered for every file uploaded - creating an endless loop. If I generate more than one file in the lambda I create a "fork bomb".

### Expected Behavior

"The eventbridge rule" (schedule) triggers one execution of "The lambda" - and one object gets uploaded to "The bucket" then finishes.

There are no extra invocations of the lambda based on events from the bucket.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS: macOS Monterrey version 12.6.1
- LocalStack: latest
```

### Anything else?

I have added a `tgz` file here with a minimal example to re-create the bug.

In the `tgz`:
`docker-compose.yml` to spin up `localstack` and to spin up a docker container that builds the lambda and sets up the infrastructure using terraform.

Start it by running 
```
docker compose up --build -d
```
To see the logs run 
```
docker compose logs localstack -f
```

After ~1 minute the lambda will trigger, and sleep for 5 seconds, then it uploads a file with uuid as name and content to S3. When the upload completes there is an event on the eventbus, and even though there is no event rule to trigger a lambda execution the lambda is executed on this event.

This means that the lambda will execute every 5 seconds for a minute, then another scheduled event happens and another file is created meaning that two files will be created every 5 seconds. It is wise not to let this run for too long.

[bug_exmple.tgz](https://github.com/localstack/localstack/files/10957210/bug_exmple.tgz)

dominikschubert:
Should be fixed with https://github.com/localstack/localstack/pull/7848

Please pull the latest image and let us know if your issue persists

jmschibsted:
Thank you @dominikschubert !

PR #7848, opened 2023-03-13
ignore scheduled rules when processing put_events

dominikschubert:
Addresses #7847

Since scheduled events don't have a pattern and it defaults to a "{}" pattern, they were always unintentionally triggering the targets of the scheduled event rule when calling `put_events` (which should only consider pattern-based event rules).

To fix this we now ignore any rules that have a `scheduled_expression` set (which would otherwise default to None).


dfangl (approved):
LGTM

PR #7851, opened 2023-03-13
Fix openapi authorization header for type token

calvernaz:
small fix to handle openapi `x-amazon-apigateway-authorizer` extension for authorizer type `TOKEN`.

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -591,11 +591,11 @@ def import_api_from_openapi_spec(
                     ):
                         authorizer["authorizerCredentials"] = authorizer_credentials
                     if authorizer_type == "TOKEN":
-                        authorizer[
-                            "identitySource"
-                        ] = f"method.request.header.{security_config.get('name')}"
+                        header_name = security_config.get('name')
+                        authorizer["identitySource"] = f"method.request.header.{header_name}"
                     elif identity_source := aws_apigateway_authorizer.get("identitySource"):
-                        # Applicable for the authorizer of the request and jwt type only.
+                        # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html
+                        # Applicable for the authorizer of the request and jwt type only
                         authorizer["identitySource"] = identity_source
                     if identity_validation_expression := aws_apigateway_authorizer.get(
                         "identityValidationExpression"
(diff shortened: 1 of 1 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -589,7 +590,12 @@ def create_authorizer(path_payload: dict) -> Optional[Authorizer]:
                         "authorizerCredentials"
                     ):
                         authorizer["authorizerCredentials"] = authorizer_credentials
-                    if identity_source := aws_apigateway_authorizer.get("identitySource"):
+                    if authorizer_type == "TOKEN":
+                        header_name = security_config.get("name")
+                        authorizer["identitySource"] = f"method.request.header.{header_name}"
```
whummer, line 595: Not sure if we can rely on the `name` of the security config to be the header name of the  `identitySource` here? I understand that this is part of the examples in the [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html#api-gateway-swagger-extensions-authorizer-example) , but seems to me that the `name` can be arbitrarily chosen (and hence may not necessarily map to the header name).

Maybe the default  `identitySource` should rather be `"method.request.header.Authorization"` here? As usual, a parity/snapshot test would help us get ultimate clarity on this question. 🤷

whummer (commented):
+1 on the refined logic based on `authorizer_type` 👍 - just not sure if we're selecting the default `identitySource` header name properly here..

localstack/services/apigateway/helpers.py
calvernaz: not sure I agree here, so the `name` is a specific field of the authorizer extension, and for the type `TOKEN` must be a header. 

> Maybe the default  identitySource should rather be "method.request.header.Authorization" here? As usual, a parity/snapshot test would help us get ultimate clarity on this question.

that's precisely one of the issues, we had always default to that and must use whatever the user defines - https://github.com/localstack/localstack-ext/blob/289bfe82cfa11f9b79412938e8c647c69aa2375c/localstack_ext/services/apigateway/authorizers.py#L387-L389

whummer: Actually, I think you're right - thanks for clarifying! Looking more closely at this line in the [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html#api-gateway-swagger-extensions-authorizer-example):


```
      "name" : "Authorization",                  // The name of the header containing the authorization token.
```

whummer (approved):

PR #7852, opened 2023-03-13
Fix transformation of falsy values in KeyValueBasedTransformer

dominikschubert:
Encountered this issue when I was trying to replace a number that flaked between 0 and non-0

## Changes
* Explicitly check for `None` with matched result


steffyP (approved):
Good catch 👍 Let's hope it doesn't break existing tests 🤞 😅

PR #7855, opened 2023-03-13
add logging for basic container information

thrau:
This adds some basic logging of information about the container. Currently it extracts the image variant if available (community, pro, full, ...), and includes a heuristic whether the docker socket is mounted into the container.
We can extend this as needed.


ackdav (approved):
LGTM 🚀 looking forward to the result...

PR #6827, opened 2022-09-06
S3 migration - set s3 v2 provider as default

steffyP:
## Summary
This PR should remain minimal in terms of code changes, and is mostly used for keeping track of the PRs regarding the new s3 provider going directly into master and for tracking test failure rates against the new provider.

## PRs
The following PRs have been created to fully implement the new s3 provider:
* #6552
* #6629
* #6836
* #6829 
* #6812 
* #6839
* #6859
* #6868 
* #6875 
* #6846
* #6903 
* #6906
* #6908
* #6937 
* #6959
* #6980
* #6987
* #7001
* #7027
* #7065
* #7070
* #7081
* #7152
* #7173

## Tasks
- [x] Migrate all s3 tests to pytest and validates them against AWS
- [X] Create the new ASF provider
- [X] Allow patching of botocore specs to match actual AWS behaviour
- [X] Reroute virtual host addressed requests as path addressed ones to take advantage of the builtin ASF routing
- [x] Implement major S3 features:
  - [X] S3 notifications
  - [X] All major basic OPs (GetObject, GetBucket, etc)
  - [X] Pre-signed URLs
  - [X] Pre-signed POST
  - [X] S3 Static Website Hosting
  - [x] CORS

- [x] Fix ongoing issues with current provider not tackled in above major features
  - [X] VersionId and Expires always present
  - [X] GetObjectAttributes not implemented
  - [X] Empty tagging lists
  - [X] Delete versioned objects
  - [X] Fix Terraform tests (fixed in #7027 + full test run green [here](https://github.com/localstack/localstack/actions/runs/3322125332/jobs/5490831947))
  - [x] remove legacy VirtualHostRewriter
  - [x] Fix issues with presigned urls

- [x] Implement the Pro provider

- [x] Implement S3 ASF test job for CircleCI

PR #7859, opened 2023-03-13
set s3 v2 provider as default

thrau:
This PR sets the s3 v2 provider as the default provider


thrau:
@bentsku actually we should maybe close this in favor of https://github.com/localstack/localstack/pull/7859 and rebase 7859 to v2

bentsku:
Sure, we'd still need to cherry pick the 2 commits in this branch and remove the one from #6827, but we'd keep the history and comments I suppose. I'll try to spend some time later today checking why one test is failing.

PR #7860, opened 2023-03-13
fix APIGW UpdateMethod authorizer validation

bentsku:
Follow up from #7749, it seems the validation was lacking and a bit too restrictive while updating Authorizers-related fields. I've added more AWS tests, this should solve our Pro pipeline (tested locally and it works).

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -471,6 +471,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         store = get_apigateway_store(account_id=context.account_id, region=context.region)
         rest_api = store.rest_apis[rest_api_id]
         applicable_patch_operations = []
+        modifying_auth_type = False
+        modified_authorizer_id = False
         for patch_operation in patch_operations:
             op = patch_operation.get("op")
             path = patch_operation.get("path")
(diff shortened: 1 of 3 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -492,11 +494,10 @@ def update_method(
                 continue
 
             if path == "/authorizationType" and value in ("CUSTOM", "COGNITO_USER_POOLS"):
```
calvernaz, line 496: so it either update the type or the id that points to, correct?

calvernaz (approved):
LGTM! 👍

localstack/services/apigateway/provider.py
bentsku: Yes, you can:
- Update the type, but if the type is either CUSTOM or COGNITO_USER_POOLS, you also need to provide a valid `AuthorizerId` in the call (so you'd need to provide both). 
- Update only the ID, it will be taken into account only if the existing authorisation type is CUSTOM or COGNITO_USER_POOLS

I can see there's one scenario missing here, if the type is already `CUSTOM` and we change it to `COGNITO_USER_POOLS`, technically I think we'd still need to update the `AuthorizerId`, but this hasn't been tested against AWS.

The validation logic is these `Update*` operation is pretty convoluted, and it takes a lot of testing to get all the possible combinations.

Issue #7861, opened 2023-03-14
bug: Topic not found, error message does not match


artur-ma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Im using AWS SDK v3
We use the `GetTopicAttributesCommand` to validate topic existence (there is not other way to do it), then we check the error code and error message.

When trying to get attributes of a topic by ARN with lockstack, it returns error with different error message than AWS:

```ts
{
    Code: 'NotFound',
    Message: 'Topic with arn <MY ARN> not found',
    message: 'Topic with arn <MY ARN> not found'
  }
```

The original AWS error

```ts
{
  Type: "Sender",
  Code: "NotFound",
  Message: "Topic does not exist",
  message: "Topic does not exist"
}
```

### Expected Behavior

Lockstack should return exact same error structure as AWS cloud

```ts
{
  Type: "Sender",
  Code: "NotFound",
  Message: "Topic does not exist",
  message: "Topic does not exist"
}
```

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

```ts
  const client = new SNSClient({ })
  try {
    await client.send(new GetTopicAttributesCommand({ TopicArn: '<ARN TO TEST>' }))
  } catch (err) {
    console.log(err)
  }
  ```
  
  Run this both with AWS and lockstack and u will see the diff

### Environment

```markdown
- OS: osX
- LocalStack:latest
```

PR #7862, opened 2023-03-14
fix NotFound exception for GetTopicAttributes

bentsku:
As reported in #7861, our error message was not in parity with AWS (error thrown from moto). We now check before and raise an exception if the topic does not exist. Added a snapshot validated test.

It also shows we have a bit of work to do regarding `TopicAttributes`, regarding `DeliveryPolicy` which seems malformed.

_fixes #7861_


baermat (approved):
LGTM!

PR #7863, opened 2023-03-14
Ignore AWS::SecretsManager-2020-07-03 transform

silv-io:
Currently, when trying to use the [global transform from SecretsManager](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-secretsmanager.html), the entire deployment crashes as it's assuming that we don't know that custom macro name. As the transform itself isn't entirely necessary for local development (as it's needed for secret rotation), we can currently just notify that it's not supported and continue deploying the remainder of the stack instead of crashing.


pinzon (approved):
LGTM :rocket:

PR #7868, opened 2023-03-14
Fix S3 routing in moto to allow upload of favicon.ico files

whummer:
Fix S3 routing in moto to allow upload of `favicon.ico` files.

This issue surfaces primarily for hostname-based bucket addressing for Pro users with DNS integration, if an S3 client is making a request to `<bucket>.s3.amazonaws.com` domains.

Moto defines the [following logic](https://github.com/getmoto/moto/blob/fe8d30ae6329878053a3ae980ed424c737fd8cec/moto/moto_server/werkzeug_app.py#L163-L164) to handle `/favicon.ico` paths as a special case:
```
if path_info.startswith("/moto-api") or path_info == "/favicon.ico":
    host = "moto_api"
```

On a general note, I think we should verify if the routing to `<bucket>.s3.amazonaws.com` domains (i.e., defined via the `Host` HTTP header) is properly supported in the new S3 provider (for the different combinations of path-/host-based addressing). This is a fairly common use case for CloudFormation custom resources (e.g., via CDK templates). Would be great to get your help on this for the v2 switch @bentsku @thrau 🙌

### DIFF
--- a/localstack/services/motoserver.py
+++ b/localstack/services/motoserver.py
@@ -47,8 +47,8 @@ def get_moto_server() -> MotoServer:
 @patch(DomainDispatcherApplication.get_application)
 def get_application(fn, self, environ, *args, **kwargs):
     """
-    Fix an upstream issue where moto treats "/favicon.ico" as a special path, which can
-    break clients attempting to upload favicon.ico files to S3 buckets.
+    Patch to fix an upstream issue where moto treats "/favicon.ico" as a special path, which
+    can break clients attempting to upload favicon.ico files to S3 buckets.
     """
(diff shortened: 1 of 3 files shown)


bentsku (approved):
LGTM! Thanks a lot for taking care of this, this feels like a proper edge case.
This new test will also reveal handy for testing the new provider matching on something else than `s3.localhost.localstack.cloud`, which I'm going to tackle next like you suggested. Thanks again! 

note: the test will fail against the new provider, but I'm currently working on a PR to fix it.

bentsku (approved):
LGTM! Thanks for adding the missing parameter in the docstring!

Issue #6942, opened 2022-09-27
Localstack's Lambda Service Does Not Match AWS File Descriptors Quota


m-whisler:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

User's of Localstack's Lambda service are unable to reproduce EMFILE issues observed in real AWS deployments. This is due to the differences in `ulimit` configuration that does not match real AWS quotas defined [here](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html)
`ulimit -a` on local machine -> -n: file descriptors 10240
`ulimit -a` from Localstack container -> nofiles 1048576
`ulimit -a` from `lambci/lambda:nodejs12.x` container after invoking Lambda function -> open files (-n) 1048576

When overriding Localstack's `ulimits` via docker-compose file the `ulimit` is not applied to Lambda service.
```
services:
  localstack:
    ulimits:
      nofile:
        soft: 1024
        hard: 1024
```
`ulimit -a` from Localstack container -> nofiles 1024
`ulimit -a` from `lambci/lambda:nodejs12.x` container after invoking Lambda function -> open files (-n) 1048576

### Expected Behavior

By default, Localstack's Lambda service should mirror and enforce [AWS limits](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html) for more accurate simulation. 
At a minimum, Localstack users should be able to apply overrides to services to better align with real AWS quotas to reproduce issues observed from real AWS deployments such as EMFILE.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    With `docker-compose.yml`...
    ```
    version: "3"
    services:
    localstack:
      image: "localstack/localstack:0.14.2"
      ports:
       - "4566:4566"
      environment:
        AWS_ACCESS_KEY_ID: "local-access-key-id"
        AWS_SECRET_ACCESS_KEY: "local-secret-access-key"
        HOSTNAME_EXTERNAL: "localstack"
        LAMBDA_DOCKER_NETWORK: "project-name-default-438332cd_default"
        LAMBDA_EXECUTOR: "docker"
        SERVICES: "lambda"
      volumes:
       - "/path/to/localstack:/docker-entrypoint-initaws.d"
       - "/var/run/docker.sock:/var/run/docker.sock"
       - "/path/to/project/lambdaArtifact.zip"
[...]

PR #7871, opened 2023-03-15
Add Lambda runtime parity configuration for ulimit

joe4dev:
Addresses https://github.com/localstack/localstack/issues/6942

This PR makes `ulimit` configurable for Lambda as an opt-in option and refactors the Docker flags parser:

* Make `ulimit` configurable through `LAMBDA_DOCKER_FLAGS=--ulimit nofile=1024:1024 --ulimit nproc=735:735 --ulimit core=-1:-1`
* Add parity test for `ulimit`
* Add unit tests for ulimit, platform, and user flags
* Refactor custom implementation of Docker flags parser using argparse. The custom subclass `NoExitArgumentParser` prevents the parser from exiting (default behavior for interactive CLI).
* Split monolithic unit test case into logical segments and improve structure (mostly alphabetical ordering)


localstack/utils/container_utils/container_client.py
```diff
@@ -103,6 +105,23 @@ class DockerPlatform(str):
     linux_arm64 = "linux/arm64"
 
 
+class Ulimit(TypedDict, total=False):
```
joe4dev, line 108: I guess we aim to keep the `container_client` generic and therefore added a `Ulimit` type instead of importing `docker.types.Ulimit`

localstack/utils/no_exit_argument_parser.py
```diff
@@ -0,0 +1,17 @@
...
+LOG = logging.getLogger(__name__)
+
+
+# Implements the `exit_on_error=False` behavior introduced in Python 3.9 to support older Python versions
+# and prevents further SystemExit for other error categories.
+# Limitations of error cases: https://stackoverflow.com/a/67891066/6875981
+# Subclassing workaround example: https://stackoverflow.com/a/59072378/6875981
+class NoExitArgumentParser(argparse.ArgumentParser):
+    def exit(self, status: int = ..., message: str | None = ...) -> NoReturn:
+        LOG.warning(f"Error in argument parser but preventing exit: {message}")
```
joe4dev, line 14: I guess that's not even needed 🤔 but better be safe than exiting here.

localstack/utils/container_utils/container_client.py
```diff
@@ -103,6 +105,24 @@ class DockerPlatform(str):
     linux_arm64 = "linux/arm64"
 
 
+@dataclasses.dataclass
+class Ulimit:
```
joe4dev, line 109: I guess we aim to keep the `container_client` generic and therefore added a Ulimit data class instead of importing `docker.types.Ulimit`

tests/integration/awslambda/test_lambda.py
```diff
@@ -377,6 +379,31 @@ def test_runtime_introspection_arm(self, lambda_client, create_lambda_function,
...
+    )
+    @pytest.mark.skip_snapshot_verify(condition=is_old_provider, paths=["$..LogResult"])
+    @pytest.mark.aws_validated
+    def test_runtime_ulimits(self, lambda_client, create_lambda_function, snapshot, monkeypatch):
+        """We consider ulimits parity as opt-in because development environments could hit these limits unlike in
+        optimized production deployments."""
+        monkeypatch.setattr(
+            config,
+            "LAMBDA_DOCKER_FLAGS",
+            "--ulimit nofile=1024:1024 --ulimit nproc=735:735 --ulimit core=-1:-1 --ulimit stack=8388608:-1",
```
joe4dev, line 394: minimal list to pass the tests but we could also be more verbose here and cover the [snapshotted limits](https://github.com/localstack/localstack/pull/7871/files#diff-fadc6153815bbe901d871c5602204e98fbfe7f8fa264f686b0fa89f55c90765fR2667) to reduce potential environment-dependent flakes

localstack/utils/container_utils/container_client.py
```diff
@@ -904,146 +931,193 @@ def untar_to_path(tardata, target_path):
...
-                        flag,
+        if args.add_hosts:
+            for add_host in args.add_hosts:
+                extra_hosts = extra_hosts if extra_hosts is not None else {}
+                hosts_split = add_host.split(":")
+                extra_hosts[hosts_split[0]] = hosts_split[1]
+
+        if args.envs:
+            for env in args.envs:
+                lhs, _, rhs = env.partition("=")
```
dfangl, line 1013: it seems we are doing this partitioning in some cases. We could define an argparse custom action for it https://docs.python.org/3/library/argparse.html#action , like demonstrated here: https://www.geeksforgeeks.org/python-key-value-pair-using-argparse/ (was the first link for it but it demonstrates what's possible).

Just a thought, nothing urgent.

localstack/utils/no_exit_argument_parser.py
dfangl: What exactly is not needed and why? The subclassing workaround is not needed due to no required arguments?

tests/integration/awslambda/test_lambda.py
```diff
@@ -377,6 +379,31 @@ def test_runtime_introspection_arm(self, lambda_client, create_lambda_function,
         invoke_result = lambda_client.invoke(FunctionName=func_name)
         snapshot.match("invoke_runtime_arm_introspection", invoke_result)
 
+    @pytest.mark.skipif(
+        not use_docker(),
+        reason="Monkeypatching of Docker flags not applicable if run locally",
+    )
```
dfangl, line 385: Same as for the other PR, why wouldn't it be applicable?

dfangl (approved):
LGTM!

tests/integration/awslambda/test_lambda.py
joe4dev: Monkeypatching works but Docker-flags are not used with the local executor in the old lambda provider because no new Docker container is started. Hence, DOCKER flags are not applicable and Docker tests with the local executor fail.

localstack/utils/container_utils/container_client.py
joe4dev: Fully agree 👍 
I saw it but just tackled the main argument parsing in the first iteration. This looks like a low-hanging fruit though.

localstack/utils/no_exit_argument_parser.py
joe4dev: ups sorry, the anchor is not that clear on GH.

I just wanted to refer to the `exit` method overriding. It appears that exit is never invoked so far but I thought it is better to protect us from an accidental `sys. exit()` (default behavior of argparse for CLI).

The subclassing workaround is definitely required because even the [exit_on_error](https://docs.python.org/3/library/argparse.html#exit-on-error) flag added in Python 3.9 only catches a subset of argument errors and just exits in other cases.

localstack/utils/container_utils/container_client.py
joe4dev: Just added a comment for now. A quick test revealed that more tests and refinements are needed to ensure equivalent parsing, especially given some flag-specific logic.

tests/integration/awslambda/test_lambda.py
joe4dev: I clarified the reason and added a new is_old_local_executor() util. This needs to be extended to the new local executor replacement (i.e., the self-managed workers) because they do not spawn new containers either.

PR #7874, opened 2023-03-15
Add request id to request context

dominikschubert:
Keeping this as a draft since we're not merging any new features until mid April.

## Changes
- Adds `request_id` to the `RequestContext` object that is passed through the handler chain. In `serialize.py` we're now using this context request_id instead of creating new IDs. (@alexrashed  @thrau)
- Format of request_id was adapted to conform to what we know from AWS for the majority of services (a uuid)
- Updates the v2 lambda provider to use the request_id from the context  (@dfangl  @joe4dev)
- Updates a few cases for s3 to use the request_id from the context (@bentsku)

## Discussion

S3 is so far the only one that seems to be behaving differently here.

There are actually 2 IDs for a request:
```json
{
    "x-amz-id-2": "k6Fag1MY0geEkMVqP1pi9o0WWmOms7zCOpt5hRXOUhgz+PoNgZOt+yFjc9e8b9WpPc+VlcShrZU=",
    "x-amz-request-id": "RZQ2PVWKR3H7EYGV"
}
```

See the comment below from @joe4dev for more details: https://github.com/localstack/localstack/pull/7874#issuecomment-1469653419

## Potential Follow-up work

- S3 needs a lot of special handling here since it has 2 (!) request IDs that both differ from the generic request ID here. (see discussion above)
- Identify services that internally make use of the request ID


joe4dev:
S3 docs about format: https://docs.aws.amazon.com/AmazonS3/latest/userguide/LogFormat.html

* **Request ID**: A string generated by Amazon S3 to uniquely identify each request.
  * Example: `3E57427F33A59F07`
* **Host Id**: The x-amz-id-2 or Amazon S3 extended request ID.
  * Example: `s9lzHYrFp76ZVxRcpX9+5cjAnEH2ROuNkd2BHfIa6UkFVdtjf5mKR3/eTPFvsiP/XV/VLi31234=`
  * It appears this extended request ID is used by AWS internally (possibly encrypted) as mentioned by this [SO post](https://stackoverflow.com/a/52581596/6875981)

More detailed AWS docs about [S3 Request IDs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/get-request-ids.html).

AWS Knowledge Center: [How do I gather the Amazon S3 request ID values requested by AWS Support?](https://repost.aws/knowledge-center/s3-request-id-values)
Screenshot 2023-03-15 at 10 25 33


This older blog post about [AWS Request IDs](https://kevinhakanson.com/2017-03-08-aws-request-ids) mentions that S3 is the exception and CloudFront also adds another header called `X-Amz-Cf-Id`

bentsku:
Awesome, thanks a lot for tackling this! This has been bugging me for a while too. I suppose you're doing it to have the proper `RequestId` linked to the lambda call? Really awesome, because we need it in SNS as well, I'll link where we should make the change as well.

About S3, I don't have much to add from @joe4dev answer. I think we can hardcode `x-amz-id-2`, because it represents the `HostId`, and we're really always the same host. We're making use of it in exception as well, in many places really, and should be consolidate in one place. 

Here the SNS making use of the lambda RequestId when invoking it.
https://github.com/localstack/localstack/blob/3cb506992733dea2c740f26cdef0f2657588ee32/localstack/services/sns/publisher.py#L183-L191

localstack/aws/protocol/serializer.py
```diff
@@ -1482,6 +1547,15 @@ def _node_to_string(self, root: Optional[ETree.ElementTree], mime_type: str) ->
...
+def gen_amzn_requestid():
+    """
+    Examples:
+    996d38a0-a4e9-45de-bad4-480cd962d208
+    b9260553-df1b-4db6-ae41-97b89a5f85ea
+    """
+    return long_uid()
+
+
 def gen_amzn_requestid_long():
```
thrau, line 1559: i guess we can remove this now?

thrau (approved):
thanks for adding this. looks like it was quite a bit of work. LGTM overall!

I'm not super happy how the serializer interface is developing, and having to change all invocations of the serializer by now adding a fake request id is not great. i have no good alternative solution though right now, and it seems like we're not going to evolve this further any time soon, so guess it's fine for now!

tests/integration/s3/test_s3_cors.py
```diff
@@ -109,9 +112,9 @@ def test_cors_http_get_no_config(self, s3_bucket, snapshot, aws_client):
         response = aws_client.s3.put_object(Bucket=s3_bucket, Key=key, Body=body, ACL="public-read")
         assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
 
-        # key_url = f"{_bucket_url_vhost(bucket_name=s3_bucket)}/{key}"
+        key_url = f"{_bucket_url_vhost(bucket_name=s3_bucket)}/{key}"
```
alexrashed, line 115: Are the changes here in the file some leftover from debugging or intentional? If they are intentional, I think the comment should be removed.

alexrashed, line 115: Are the changes here in the file some leftover from debugging or intentional? If they are intentional, I think the comment should be removed.

alexrashed (approved):
Awesome! This is pretty big (due to the cross-cutting nature of the request ID being forwarded in every invocation), but is going to be quite beneficial.
Next step: Tracing! 😄 🚀

tests/integration/s3/test_s3_cors.py
bentsku: Oh, those were a remnant from the CORS issue linked to thread local, and I think Dominik changed it again to be able to run against AWS. I guess all of those could be set back to use `_bucket_url_vhost` and remove all the comments. I can do in a subsequent PR as well

dfangl (approved):

tests/integration/s3/test_s3_cors.py
dominikschubert: @bentsku already kinda explained how that slipped in

will revert :+1:

bentsku (approved):
LGTM! Thanks for tackling this, it seemed quite the work. And awesome that we can now use the request id to map the lambda invocation! 🎉

PR #7875, opened 2023-03-15
Add v1 provider aliases for legacy providers

dominikschubert:
Since we currently have been advertising the newer provider implementations as `v2`, users might try to "revert" to the legacy provider by setting the `PROVIDER_OVERRIDE_<svc>` variable to `v1`. Unfortunately this will currently fail since they need to remove the variable completely.

This PR now introduces v1 aliases for the legacy (currently still default) providers with a `v2` provider, i.e. `lambda`, `s3` and `stepfunctions`

cc @bentsku @dfangl

### DIFF
--- a/localstack/services/providers.py
+++ b/localstack/services/providers.py
@@ -152,6 +152,19 @@ def awslambda():
     )
 
 
+@aws_provider(api="lambda", name="v1")
+def awslambda_v1():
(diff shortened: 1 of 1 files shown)


thrau (approved):
great idea! we'll probably see this pattern a few more times moving forward, so maybe makes sense to introduce this as first class citizen into the provider plugins.

PR #7876, opened 2023-03-15
chore: update base image to python:3.10.10-slim-bullseye

Mins:
WIP upgrade the base image in Dockerfile to Debian Bullseye

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG IMAGE_TYPE=community
 
 # java-builder: Stage to build a custom JRE (with jlink)
-FROM python:3.10.8-slim-buster@sha256:6f0a9332035a0268cdca0bfecb509ec17db855e3d079d134373b3bf5bfb9e98f as java-builder
+FROM python:3.10.10-slim-bullseye@sha256:875d13333257e9f7ac2caf391bc8182fdd43c9c19d6a49fe27166eb42142c68a as java-builder
 ARG TARGETARCH
 
 # install OpenJDK 11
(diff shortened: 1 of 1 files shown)


Mins:
The container image builds are currently failing when installing `timescaledb` using `localstack.cli.lpm`. I'm not really familiar with the packaging system, but it seems to be a private (localstack-ext?). 

@alexrashed do you happen to know how that package can be updated?

alexrashed:
Hi @Mins!
Thanks for your contribution! You are correct, timescaledb is used for our Pro extensions. Unfortunately, this will not allow to upgrade to the latest version of Debian so easily. I'm afraid we will stick to old-stable for a bit (bot not too long). Also, the Dockerfile and the package structure will be quite heavily refactored in a few days (we are currently working towards our next major release).
Is there a reason why you need to use Bullseye instead of Buster right now?

Mins:
Thanks for the clarification @alexrashed ! Our organization has a policy of using only container images with updated base OSes, and the minimum required version for Debian was recently bumped up to Bullseye.

> we will stick to old-stable for a bit (bot not too long).
Since there is already work planned for the upgrade, I'll be happy to close this PR. Can you kindly let me know if there is a rough timeline for the upgrade?

alexrashed:
Sorry I didn't explicitly post an update here earlier. We upgraded to bullseye some time ago with #8344.
LocalStack 2.1.0 is already based on Debian Bullseye (as mentioned [in the release notes](https://github.com/localstack/localstack/releases/tag/v2.1.0))

Issue #7074, opened 2022-10-22
feature request: allow the customization of the max lambda zip upload size


TheDumbfounds:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

Despite the fact that the AWS Lambda official docs say that the limit is 50MB, I'm deploying a larger function in production and it's working fine.

However, Localstack gives me an error when I try to create a lambda using the zip package I deployed to AWS, saying that the max limit was exceeded.

It would be nice to have a variable where I can control the max limit myself.

This would also be useful for people who want to make sure to not deploy
packages above a certain limit they set and want to limit themselves when testing using
localstack as well.

joe4dev:
Hi @TheDumbfounds. Thanks for reporting a potential mismatch and suggesting an interesting feature.

## AWS <=> LocalStack Parity regarding AWS Lambda Limit

I tested different file sizes (using `dd if=/dev/urandom of=large_file bs=1M count=50`) and found that the current LocalStack behavior matches AWS (us-east-1, eu-central-1) using a threshold of 69905067 bytes for the maximum request size (defined in `localstack/services/awslambda/invocation/lambda_models.py:571`). The behavior is consistent with AWS in the current old Lambda provider as well as the new Lambda provider, which is available since `1.3.0` through the feature flag `PROVIDER_OVERRIDE_LAMBDA=asf`.

Assuming you create a lambda via zip like this:

```bash
aws --endpoint-url=http://localhost:4566 \
lambda create-function --function-name my-large-lambda \
--zip-file fileb://function.zip \
--handler main.handler --runtime python3.9 \
--role arn:aws:iam::000000000000:role/lambda-role
```

1. Is your error message related to a maximum request size (rather than other [Lambda quotas](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html) such as the 50MB limit for zipped direct uploads)?

    > An error occurred (RequestEntityTooLargeException) when calling the CreateFunction operation: Request size (69916993) must be smaller than 69905067 bytes for the CreateFunction operation

3. Does your AWS account has any modified limits?

4. What is the exact size of your zip file in bytes (e.g., `ls -la large_file`)?

5. In which AWS region did your larger-than-50MB zipped function deploy successfully?

## Feature suggestion for a configurable deployment size limit

That's a nice suggestion and would be a good candidate for a [LocalStack extension](https://localstack.cloud/blog/2022-09-12-announcing-localstack-extensions/).

## Background

[...]

PR #7881, opened 2023-03-15
Add support for provisioned and reserved concurrency 

dominikschubert:
Account limit config addresses https://github.com/localstack/localstack/issues/7074

## Background

AWS Lambda concurrency main docs page: https://docs.aws.amazon.com/lambda/latest/dg/lambda-concurrency.html
image

## Assumptions

* [Lambda quota](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html) apply at different levels
  * Typically "per AWS region" (e.g., `Concurrent executions`)

## Implementation

* We keep a counter per region and function by storing the unqualified function ARN in a dict, protected by locks.
* The counter should only be updated for 'on-demand' invocation types because provisioned concurrency is always deduced immediately after adding a provisioned concurrency config.
* We calculate `get_available_fn_concurrency` before allowing or denying every potential invocation

## Open (will mostly come in later PRs

- Check if there are any provisioned concurrency execution environments available before raising `TooManyRequestsException`. We generally want to favor provisioned execution environments over on-demand.
- `get_available_fn_concurrency` for async invokes 🤔 
- [x] Make account limit configurable and monkeypatch the account settings test (opt-in parity)


localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -191,39 +213,119 @@ def stop(self) -> None:
...
+        TODO: implement update while in progress (see test_provisioned_concurrency test)
+        TODO: loop until diff == 0 and retry to remove/add diff environments
+        TODO: alias routing & allocated
+        TODO: ProvisionedConcurrencyStatusEnum.FAILED
+        TODO: status reason
+
+        :param provisioned_concurrent_executions: set to 0 to stop all provisioned environments
+        """
+
+        if self.provisioned_state == ProvisionedConcurrencyStatusEnum.IN_PROGRESS:
```
dfangl, line 250: Doesn't make too much sense comparing the container with the state here but actually comparing the state will break it as well, as IN_PROGRESS is the default state when creating the provisioned_state in provider.py.

We probably should check if it is not None, and set the container afterwards, instead of doing it in provider.py.

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -191,39 +213,119 @@ def stop(self) -> None:
...
+                    runtime_environment = RuntimeEnvironment(
+                        function_version=self.function_version,
+                        initialization_type="provisioned-concurrency",
+                        service_endpoint=self,
+                    )
+                    self.all_environments[runtime_environment.id] = runtime_environment
+
+                    def start_provisioning_env():
+                        runtime_environment.start()
+                        self.set_environment_ready(runtime_environment.id)
```
dfangl, line 277: This should be set by the /ready report of the LRI, not manually

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -191,39 +213,119 @@ def stop(self) -> None:
...
+                RuntimeStatus.STARTING,
+                RuntimeStatus.RUNNING,
+            ]
+        )
+
+        # TODO: invocation number instead of considered
+        if (
+            self.function.reserved_concurrent_executions is not None
+            and considered_envs >= self.function.reserved_concurrent_executions
+        ):
```
dfangl, line 318: We should honor that todo :)

localstack/services/awslambda/provider.py
```diff
@@ -2314,10 +2320,10 @@ def put_provisioned_concurrency_config(
         manager = self.lambda_service.get_lambda_version_manager(fn_arn)
 
         fn.provisioned_concurrency_configs[qualifier] = provisioned_config
-        manager.provisioned_state = ProvisionedConcurrencyState(
-            allocated=provisioned_concurrent_executions,
-            available=provisioned_concurrent_executions,
-            status=ProvisionedConcurrencyStatusEnum.READY,
+
+        manager.provisioned_state = ProvisionedConcurrencyState()
```
dfangl, line 2324: Probably should not set it externally here, but handle that inside the version manager.

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -261,9 +363,27 @@ def invocation_loop(self) -> None:
                     )
                     return
                 LOG.debug("Got invocation event %s in loop", queued_invocation.invocation_id)
+                # Assumption: Synchronous invoke should never end up in the invocation queue because we catch it earlier
```
dfangl, line 366: We discussed it, but this assumption is probably incorrect, as some synchronous invocations could still be left in the queue.

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -275,11 +395,20 @@ def invocation_loop(self) -> None:
                                 self.function_arn,
                             )
                             return
+
+                        # TODO: untested assumption(!)
+                        #   because provisioned concurrency affects the unreserved
```
dfangl, line 400: What exactly is the untested assumption here?

If I set my provisioned_concurrency to 50, and reserved_concurrency to 70, I could invoke 120 parallel lambdas as it is currently implemented, right?

dfangl (changes requested):
LGTM in general, just some things still unclear / missing TODO implementations.

localstack/services/awslambda/invocation/lambda_service.py
```diff
@@ -56,18 +58,36 @@
 LAMBDA_DEFAULT_MEMORY_SIZE = 128
 
 
+# TODO: scope to account & region instead?
```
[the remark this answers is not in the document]
dominikschubert: Because the reserved/unreserved concurrency is actually calculcated per account **and** region. :grimacing:

localstack/services/awslambda/invocation/lambda_service.py
```diff
@@ -315,6 +345,99 @@ def update_version_state(
...
+            tracker.function_concurrency[unqualified_function_arn] -= 1
+            if tracker.function_concurrency[unqualified_function_arn] < 0:
+                LOG.warning(
+                    "Invalid function concurrency state detected for function: %s | recorded concurrency: %d",
+                    unqualified_function_arn,
+                    tracker.function_concurrency[unqualified_function_arn],
+                )
+                raise Exception(
+                    "Invalid function concurrency state"
+                )  # TODO: remove this before merge
```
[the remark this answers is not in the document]
dominikschubert: Yeah we just added this temporarily to check if anything in the pipeline triggers it

localstack/services/awslambda/invocation/version_manager.py
dominikschubert: On the same function it should actually still only be max 70 that you can invoke concurrently. The provisioned concurrency counts towards the reserved concurrent executions of the function

dominikschubert: So the reason we're not tracking them here is because provisioned concurrency is always automatically deducted in full from the available concurrency (either Account/Region global or per function if it has reserved concurrency set).

dfangl (approved):
LGTM, big changes and challenging feature :)
One test is still red, but no need to re-review once this is fixed (I think).

localstack/services/awslambda/provider.py
```diff
@@ -2275,7 +2297,25 @@ def put_provisioned_concurrency_config(
         provisioned_config = self._get_provisioned_config(context, function_name, qualifier)
 
         if provisioned_config:
-            pass  # TODO: test
+            pass  # TODO: merge?
```
joe4dev, line 2300: would be nice to at least log this unhandled case

joe4dev (approved):
Good first iteration to support the lambda concurrency features 👍

✅ The 4 tests `tests.integration.awslambda.test_lambda.TestLambdaConcurrency` pass on macOS

Follow-up PRs will need to cover several edge cases. Ideally, we would introduce appropriate abstractions to tackle the rising complexity of the version manager.

PR #7883, opened 2023-03-16
Track additional host variables

simonrw:
These variables are

* `LOCALSTACK_HOSTNAME` which is not _meant_ to be settable by the user as it's used internally, but it is settable
* `LOCALSTACK_HOST` which is a new variable that we plan to use in exchange for `HOSTNAME_EXTERNAL` and `LOCALSTACK_HOSTNAME`.

These variables are added so that we can use the analytics to see the usage of the previous variables drop.


ackdav (approved):
LGTM 👍 thanks for adding this before v2

PR #7885, opened 2023-03-16
fix scheduler start/stop for persistence

steffyP:
CloudWatch is using a scheduler, that should check if an alarm state should change periodically.

* The scheduler should be restarted and check all existing metric-alarms for the `on_*_load` scenario.
* The scheduler should also be restarted for the `on_*_reset` scenario


giograno (approved):
Start/stop logic looks good 👍

PR #7888, opened 2023-03-16
[wip] Fix race condition in Server startup to enable is_running() loops

whummer:
**🚧 work in progress...**

Fix race condition in Server startup to enable `is_running()`-based server loops.

### DIFF
--- a/localstack/testing/pytest/fixtures.py
+++ b/localstack/testing/pytest/fixtures.py
@@ -1880,7 +1880,7 @@ def ses_verify_identity(ses_client):
 
 
 @pytest.fixture
-def cleanups(ec2_client):
+def cleanups():
     cleanup_fns = []
 
     yield cleanup_fns
--- a/localstack/utils/serving.py
+++ b/localstack/utils/serving.py
@@ -125,9 +125,9 @@ class Server(abc.ABC):
             if self._started.is_set():
                 return False
 
-            self._thread = self.do_start_thread()
-            self._started.set()
-            return True
+        _thread = self.do_start_thread()
+        self._set_thread_started(_thread)
+        return True
 
     def join(self, timeout=None):
         """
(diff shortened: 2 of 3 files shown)


thrau (commented):
to my understanding, this can only happen when `self.is_running()` is called from within the `do_run` method, correct? if so, that would fundamentally seem like a flawed implementation, since `do_run` is inherently meant to be the _blocking_ method that defines whether `is_running` is true or not.
i'm having a hard time understanding what the case is for calling `is_running` from within `do_run`. could you elaborate?

whummer:
Sorry, should have made myself more clear @thrau . Added a few more details to the example above (socket server accepting connections).

An alternative would be to check for "stopped" state - in that case we could add an `is_stopped()` method to the interface:
```
class MyServer(Server):
    def do_run(self):
        server_socket = ...
        while not self.is_stopped():
            socket = server_socket.accept()
            # handle connection ...
        ...
```

Either way is fine for me - using `is_running` sounds a bit more intuitive to me (and is also what the contract of the method indicates), but `is_stopped` may be easier/cleaner to add with our current implementation. 👍

thrau:
neither `is_running` or `is_stopped` will interrupt `server_socket.accept()` though, so it will just continue blocking.

it seems an implementation that conceptually works like this would remove the need for calling `is_running` (or any sort of derivative like `is_stopped`) from `do_run`:
```python
class MySocketServer(Server):
    def __init__(self, port, host="localhost"):
        super().__init__(port, host)
        self.socket = None

    def do_run(self):
        self.socket = sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        sock.bind((self.host, self.port))
        sock.listen()

        while True:
            try:
                (clientsocket, address) = sock.accept()
                # do stuff with the socket
            except OSError:
                break  # raised when the socket is closed

    def do_shutdown(self):
        self.socket.shutdown(socket.SHUT_RDWR)
        self.socket.close()
```

or am i missing something?

whummer:
That's actually a nice one - makes perfect sense, not sure how I could have missed it. 😄 Thanks for taking the time to spec it out! 🙌

PR #7890, opened 2023-03-16
Add lambda config to ignore architecture

joe4dev:
Add an environment variable `LAMBDA_IGNORE_ARCHITECTURE` to ignore the architecture (i.e., x86_64 or arm64) of a Lambda function and let Docker select a matching image based on the host operating system and architecture.

## Motivation

* Improve performance by natively running cross-platform compatible lambda functions.
* Facilitate cross-platform testing with different test runners. We might need to introduce pytest markers to indicate compatibility restrictions (e.g., `amd64_only` and `arm64_only`).

## Further changes

* Use standardized `get_arch()` to match `aarch64` platform on Linux
* Move `is_arm_compatible()` to platform utils
* Make the `PULLED_IMAGES` cache platform-aware to prevent running into timeouts upon invocation

### DIFF
--- a/tests/integration/awslambda/test_lambda.py
+++ b/tests/integration/awslambda/test_lambda.py
@@ -410,6 +410,68 @@ class TestLambdaBehavior:
         lambda_arch = standardized_arch(payload.get("platform_machine"))
         assert lambda_arch == native_arch
 
+    @pytest.mark.skipif(is_old_provider(), reason="unsupported in old provider")
+    @pytest.mark.skipif(
+        not is_arm_compatible() and not is_aws(),
+        reason="ARM architecture not supported on this host",
+    )
+    @pytest.mark.aws_validated
+    def test_mixed_architecture(self, lambda_client, create_lambda_function):
+        """Test emulation and interaction of lambda functions with different architectures.
+        Limitation: only works on ARM hosts that support x86 emulation.
+        """
+        func_name = f"test_lambda_x86_{short_uid()}"
+        create_lambda_function(
+            func_name=func_name,
+            handler_file=TEST_LAMBDA_INTROSPECT_PYTHON,
+            runtime=Runtime.python3_9,
+            Architectures=[Architecture.x86_64],
+        )
+
+        invoke_result = lambda_client.invoke(FunctionName=func_name)
+        assert "FunctionError" not in invoke_result
+        payload = json.loads(invoke_result["Payload"].read())
+        assert payload.get("platform_machine") == "x86_64"
+
+        func_name_arm = f"test_lambda_arm_{short_uid()}"
+        create_lambda_function(
+            func_name=func_name_arm,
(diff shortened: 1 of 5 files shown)


tests/integration/awslambda/test_lambda.py
```diff
@@ -377,6 +378,99 @@ def test_runtime_introspection_arm(self, lambda_client, create_lambda_function,
         invoke_result = lambda_client.invoke(FunctionName=func_name)
         snapshot.match("invoke_runtime_arm_introspection", invoke_result)
 
+    @pytest.mark.skipif(is_old_provider(), reason="unsupported in old provider")
+    @pytest.mark.skipif(
+        not use_docker(),
+        reason="Monkeypatching of Docker-related flag not applicable if run locally",
+    )
```
dfangl, line 385: Why wouldn't this work?

tests/integration/awslambda/test_lambda.py
```diff
@@ -377,6 +378,99 @@ def test_runtime_introspection_arm(self, lambda_client, create_lambda_function,
...
+        invoke_result = lambda_client.invoke(FunctionName=func_name)
+        payload = json.loads(to_str(invoke_result["Payload"].read()))
+        lambda_arch = standardized_arch(payload.get("platform_machine"))
+        assert lambda_arch == native_arch
+
+    @pytest.mark.skipif(is_old_provider(), reason="unsupported in old provider")
+    @pytest.mark.skipif(
+        not is_arm_compatible() and not is_aws(),
+        reason="ARM architecture not supported on this host",
+    )
```
dfangl, line 416: I think this is fine for now, but emulation is thinkable in both ways, also arm64 on amd64.

dfangl (approved):
LGTM, just some minor questions for tests, nothing to hold back the merge though.

tests/integration/awslambda/test_lambda.py
joe4dev: interesting point. That could even be added to `is_arm_compatible()` if it becomes relevant in the future.

tests/integration/awslambda/test_lambda.py
dfangl: Ah, now i understand. The `if run locally` confused me, as I thought of running the test locally, not running lambdas in the local executor. Overlooked the `use_docker` and thought of `in_docker`.

Still, use_docker might be problematic, as it would skip the tests if I set (by accident) something weird like `LAMBDA_EXECUTOR=local PROVIDER_OVERRIDE_LAMBDA=v2`.

joe4dev: I clarified the reason and added a new `is_old_local_executor()` util. This skip is currently redundant for `test_ignore_architecture` given the old provider is unsupported anyways but needs to be extended to the new local executor replacement (i.e., the self-managed workers) because they do not spawn new containers either.

PR #7893, opened 2023-03-17
Configure cosmetic and functional variables with LOCALSTACK_HOST and GATEWAY_LISTEN

simonrw:
We are switching to using `LOCALSTACK_HOST` and `GATEWAY_LISTEN` to configure LocalStack.

## Variable types

These variables are split into two types: _cosmetic_ and _functional_. Cosmetic are purely used as hostnames in returned URLs or addresses by LocalStack. LocalStack does not use these variables to configure services.

_functional_ variables are used to configure the networking stack of LocalStack.

## Roles

* `LOCALSTACK_HOST`: (cosmetic) this variable is interpolated into URLs and addresses. It has the form `<hostname>:<port>` and defaults to `localhost.localstack.cloud:4566`.
* `GATEWAY_LISTEN`: (functional) this configures the bind addresses of LocalStack. It has the form `<ip address>:<port>(,<ip address>:<port>)*` and defaults to `127.0.0.1:4566` on the host and `0.0.0.0:4566` in docker.

## `GATEWAY_LISTEN` advantages

Since `GATEWAY_LISTEN` supports multiple addresses (separated by `,`) we can provide configuration for pro users who wish to use SSL e.g. `GATEWAY_LISTEN=0.0.0.0:4566,0.0.0.0:443`.

## Legacy configuration

Existing configuration of the system is maintained with the existing legacy variables (`EDGE_PORT`, `EDGE_PORT_HTTP` and `EDGE_BIND_HOST`) as these are now derived from `GATEWAY_LISTEN` if absent. Some examples:

* `GATEWAY_LISTEN`: "" -> `EDGE_PORT=4566`,`EDGE_PORT_HTTP=4566`,`EDGE_BIND_HOST=127.0.0.1/0.0.0.0`
* `GATEWAY_LISTEN=0.0.0.0:9999` -> `EDGE_PORT=9999`,`EDGE_PORT_HTTP=9999`,`EDGE_BIND_HOST=0.0.0.0`

If `GATEWAY_LISTEN` is not given but legacy variables are, they pick up the supplied values rather than inheriting the default `GATEWAY_LISTEN`.

If `GATEWAY_LISTEN` includes multiple values, the first will be used to configure our `hypercorn` listener, and the rest will be transparent proxies (implementation in future PR).

Legacy variables have been marked as deprecated:
* `LOCALSTACK_HOSTNAME`
* `HOSTNAME_EXTERNAL`
* `EDGE_PORT`
* `EDGE_PORT_HTTP`
* `EDGE_BIND_HOST`

[Design document](https://www.notion.so/localstack/Networking-sync-9f34c512d964469689eff9e815281238)

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -317,10 +317,15 @@ DEFAULT_REGION = (
 )
 
 # expose services on a specific host externally
+# DEPRECATED:  since v2.0.0 as we are moving to LOCALSTACK_HOST
 HOSTNAME_EXTERNAL = os.environ.get("HOSTNAME_EXTERNAL", "").strip() or LOCALHOST
 
 # name of the host under which the LocalStack services are available
+# DEPRECATED: if the user sets this since v2.0.0 as we are moving to LOCALSTACK_HOST
 LOCALSTACK_HOSTNAME = os.environ.get("LOCALSTACK_HOSTNAME", "").strip() or LOCALHOST
+
+# configuration of hostnames the user can use to access LocalStack
+LOCALSTACK_HOST = os.environ.get("LOCALSTACK_HOST")
 
 # directory for persisting data (TODO: deprecated, simply use PERSISTENCE=1)
 DATA_DIR = os.environ.get("DATA_DIR", "").strip()
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -433,7 +433,6 @@ def populate_legacy_edge_configuration(
     # populate LOCALSTACK_HOST first since EDGE_BIND may be derived from LOCALSTACK_HOST
     localstack_host = localstack_host_raw
     if localstack_host is None:
-        localstack_host = f"{default_ip}:4566"
         localstack_host = f"{constants.LOCALHOST_HOSTNAME}:{constants.DEFAULT_PORT_EDGE}"
 
     def parse_edge_bind(value: str) -> str:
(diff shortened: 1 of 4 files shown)


localstack/config.py
```diff
@@ -411,12 +408,113 @@ def in_docker():
...
+def populate_legacy_edge_configuration(
+    environment: Dict[str, str]
+) -> Tuple[str, str, str, int, int]:
+    if is_in_docker:
+        default_ip = "0.0.0.0"
+    else:
+        default_ip = "127.0.0.1"
+
+    localstack_host_raw = environment.get("LOCALSTACK_HOST")
+    edge_bind_raw = environment.get("GATEWAY_LISTEN")
```
thrau, line 430: nit: variable naming could be updated

localstack/config.py
```diff
@@ -411,12 +408,113 @@ def in_docker():
...
+        else:
+            return f"{value}:4566"
+
+    edge_bind = edge_bind_raw
+    if edge_bind is None:
+        # default to existing behaviour
+        port = int(localstack_host.split(":")[-1])
+        edge_bind = f"{default_ip}:{port}"
+    else:
+        components = edge_bind.split(",")
```
thrau, line 456: nit: should we maybe output a warning here if there are >1 components in here for now, since they are ignored? we could however pretty soon update the logic to start the gateway and proxies.

thrau (approved):
this looks great! really nice deprecation path. just a couple of nits regarding naming, but nothing blocking

simonrw:
The `test_presigned_url_v4_signed_headers_in_qs` S3 integration test is failing, but will be fixed with an upcoming merge of the new lambda provider. Thanks @alexrashed for the suggestion. I will merge this even though the pipeline is failing.

PR #7894, opened 2023-03-17
fix validation in test for put-subscription-filter with lambda

steffyP:
While adding persistence test for logs, I found that the validation for this test is actually misleading, as it didn't check the lambda-logs which should ensure the actual invocation

### DIFF
--- a/tests/integration/test_logs.py
+++ b/tests/integration/test_logs.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import base64
 import gzip
 import json
 import re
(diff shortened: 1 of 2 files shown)


tests/integration/test_logs.py
```diff
@@ -277,14 +277,25 @@ def test_put_subscription_filter_lambda(
...
+                    #  "logStream":"log-stream","subscriptionFilters":["test"],
+                    #  "logEvents":[{"id":"7","timestamp":1679056073581,"message":"test"},
+                    #               {"id":"8","timestamp":1679056073581,"message":"test 2"}]}
+                    data = json.loads(e["message"])["awslogs"]["data"].encode("utf-8")
+                    decoded_data = gzip.decompress(base64.b64decode(data)).decode("utf-8")
+                    for log_event in json.loads(decoded_data)["logEvents"]:
+                        filtered_events.append(log_event)
+            assert len(filtered_events) == 2
+
+            filtered_events.sort(key=lambda k: k.get("message"))
```
dominikschubert, line 297: nit: could be replaced with a sorting transformer

dominikschubert (approved):
LGTM :+1:

PR #7895, opened 2023-03-17
Fix SortingTransformer for nested input

dominikschubert:
Fixes a small edge case encountered when writing tests for xray with a SortingTransformer where SubSegments can be nested. The value of the detected entry is now also further transformed.


steffyP (approved):
LGTM! 👍 Thanks for fixing and adding a unit-test for the transformation 🔝 😄

PR #7899, opened 2023-03-17
Update default container name from localstack_main to localstack-main

simonrw:
Docker can use a container name for DNS resolution, but hostnames cannot include underscores.

This change sets the default container name in the documentation to localstack-main to allow it to be used as a hostname.


dominikschubert (approved):
LGTM. :+1: 
Only thing that might be a bit confusing is the difference between `LOCALSTACK_DOCKER_NAME` and `MAIN_CONTAINER_NAME`

Issue #7897, opened 2023-03-17
bug: SQS Lambda Listener, message attributes are getting passed in body attribute of SQSEvent of lambda handler


bajajyog:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

TestContainer Local stack version **1.4.0**

**Input msg to SQS:-**
{
  "Records": [
    {
      "messageId": "d9144555-9a4f-4ec3-99a0-34ce359b4b54",
      "receiptHandle": "13e7f7851d2eaa5c01f208ebadbf1e72==",
      "body": "{ \"traceId\":\"8908118f-06e6-460f-8297-1be74d4ff03f\"},
      "attributes": {
        "ApproximateReceiveCount": "1",
        "SentTimestamp": "1601975706495",
        "SenderId": "AROAIFU437PVZ5L2J53F5",
        "ApproximateFirstReceiveTimestamp": "1601975706499"
      },
      "messageAttributes": {
      },
      "md5OfBody": "13e7f7851d2eaa5c01f208ebadbf1e72",
      "eventSource": "aws:sqs",
      "eventSourceARN": "arn:aws:sqs:eu-central-1:123456789012:TestLambda",
      "awsRegion": "eu-central-1"
    }
  ]
}

**Message received at Lambda Handler for SQSEvent body attribute**

:{"records":[{"messageId":"d9144555-9a4f-4ec3-99a0-34ce359b4b54","receiptHandle":"13e7f7851d2eaa5c01f208ebadbf1e72==","body":"{\n  \"traceId\":\"8908118f-06e6-460f-8297-1be74d4ff03f\"

### Expected Behavior

**Expected Result for SQSEvent Body attribute at java lambda handler**
****
"{\n  \"traceId\":\"8908118f-06e6-460f-8297-1be74d4ff03f\"

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    LocalStack localstack=new LocalStackContainer().withServices(Service.LAMBDA, Service.SQS)
                            .withEnv("DEFAULT_REGION", Region.US_EAST_1.toString())
                            .withNetwork(Network.SHARED)
                            .withNetworkAliases(localstackNetworkAlias)
                            .withEnv("LAMBDA_DOCKER_NETWORK", ((Network.NetworkImpl) Network.SHARED).getName())
[...]

PR #7900, opened 2023-03-17
add sqs lambda event source listener test for message attributes

thrau:
This PR adds a test to verify #7897. Basically it's just a normal "let the lambda echo the SQS message into another SQS queue and fetch it" test, but i added message attributes explicitly.

The test didn't really reproduce the issue, but it's useful as regression test I suppose. The only difference I could find is that AWS adds empty values for `stringListValues` and `binaryListValues` of message attributes, but 1) [those are not implemented in AWS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html), and 2) these empty values only show up when using the event source mapper, not in SQS in general. :thinking: :shrug: 

```
  "Author": {
    "stringValue": "John Grisham",
    "stringListValues": [],
    "binaryListValues": [],
    "dataType": "String"
  },
```

in localstack that would be
```
  "Author": {
    "stringValue": "John Grisham",
    "dataType": "String"
  },
```

### DIFF
--- a/tests/integration/awslambda/test_lambda_integration_sqs.py
+++ b/tests/integration/awslambda/test_lambda_integration_sqs.py
@@ -161,6 +161,94 @@ def test_failing_lambda_retries_after_visibility_timeout(
 
 @pytest.mark.skip_snapshot_verify(
     paths=[
+        # AWS returns empty lists for these values, even though they are not implemented yet
+        # https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html
+        "$..stringListValues",
+        "$..binaryListValues",
+    ]
+)
+def test_message_body_and_attributes_passed_correctly(
+    create_lambda_function,
+    lambda_client,
+    sqs_client,
+    sqs_create_queue,
+    sqs_queue_arn,
+    lambda_su_role,
+    snapshot,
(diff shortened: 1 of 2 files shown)


tests/integration/awslambda/test_lambda_integration_sqs.py
```diff
@@ -160,6 +160,94 @@ def test_failing_lambda_retries_after_visibility_timeout(
...
 
+@pytest.mark.skip_snapshot_verify(
+    paths=[
+        # AWS returns empty lists for these values, even though they are not implemented yet
+        # https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_MessageAttributeValue.html
+        "$..stringListValues",
+        "$..binaryListValues",
+    ]
+)
+def test_message_body_and_attributes_passed_correctly(
```
dominikschubert, line 171:
```suggestion
)
@pytest.mark.aws_validated
def test_message_body_and_attributes_passed_correctly(
```

dominikschubert (approved):
LGTM :+1: 
I agree, event if it might not have reproduced the issue, it's quite useful to have more validated tests here :)

Just needs to be rebased & clients need to be migrated to the `aws_client` fixture

PR #7554, opened 2023-01-25
Extend allowed CORS origins to include EDGE_PORT_HTTP endpoint

whummer:
Extend allowed CORS origins to include `EDGE_PORT_HTTP` endpoint. This change ensures that the `http://localhost:4566` endpoint is in the `ALLOWED_CORS_ORIGINS` list (for Community + Pro users). 

The issue surfaced when deploying our [demo sample app](https://github.com/localstack/localstack-demo/pull/47) Web UI (HTML file) as an S3 website, and testing against LocalStack Pro: In the default Pro setup, we have `EDGE_PORT=443` and `EDGE_PORT_HTTP=4566`, and since we were previously configuring the allowed origins only via `config.EDGE_PORT`, it was not possible to access the LocalStack APIs from a website running under `http://localhost:4566` (only when using port 443).

Touching the CORS setup is always a bit finicky, but this change (1) should have no negative security side effects (we're only allowing the additional localhost port 4566, in addition to localhost port 443), and (2) ensures that we can offer S3 websites (e.g., `http://localhost:4566/mybucket/index.html`) that can connect directly to the LocalStack APIs, working for both Community and Pro.

The PR adds a small integration test to cover the logic around `EXTRA_CORS_EXPOSE_HEADERS` (existing logic, currently not yet covered), as well as a unit test to cover the new functionality around allowed localhost origins with different ports (incl. `EDGE_PORT_HTTP`).

PR #7904, opened 2023-03-18
add dynamic check of origin for CORS

bentsku:
This is a follow up from #7554, from an idea proposed by @whummer ([in this comment](https://github.com/localstack/localstack/pull/7554#discussion_r1086872025)) , with dynamic handling of CORS origin for web resource served by LocalStack.
This is not mergeable as is, as it still has some logging regarding performance, but I'd like to have some comments / opinions.

### Scenario:

LocalStack is serving a resource meant to be accessed from the browser (CloudFront or S3 website).

But this resource needs to access LocalStack resources as well. Right now, the only way is to set the environment variables of LocalStack to allow the Origin (let’s say, `http://<bucket>.s3-website.localhost.localstack.cloud:4566`).

### Fix

This PR adds a dynamic check of the origin, to validate if it follows some regex matching 2 service domains known for serving content accessible from a user browser (thus enforcing CORS).  We can then add a regex for every resource subdomain in that situation. 

The performance right now is around 4µs on my local machine when a request would be either rejected or matching `s3-website` or `cloudfront`. The performance would not be affected for most requests, as we either don't match CORS because of no `Origin` header, or the request is normally allowed. 

\cc @whummer @dfangl @alexrashed

### DIFF
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -76,7 +76,7 @@ ALLOWED_CORS_RESPONSE_HEADERS = [
 ]
 
 
-def _get_allowed_cors_internal_domains() -> Set[int]:
+def _get_allowed_cors_internal_domains() -> Set[str]:
     """
     Construct the list of allowed internal domains for CORS enforcement purposes
     Defined as function to allow easier testing with monkeypatch of config values
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -112,7 +112,7 @@ def _get_allowed_cors_origins() -> List[str]:
     # Add allowed origins for localhost domains, using different protocol/port combinations.
     # If a different port is configured for EDGE_PORT_HTTP, add it to allowed origins as well
     for protocol in {"http", "https"}:
-        for port in _ALLOWED_INTERNAL_PORTS:
+        for port in _get_allowed_cors_ports():
             result.append(f"{protocol}://{LOCALHOST}:{port}")
             result.append(f"{protocol}://{LOCALHOST_HOSTNAME}:{port}")
 
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -3,7 +3,7 @@ A set of handlers which handle Cross Origin Resource Sharing (CORS).
 """
 import logging
 import re
-from typing import List
+from typing import List, Set
 from urllib.parse import urlparse
 
 from flask_cors.core import (
@@ -20,6 +20,7 @@ from localstack.config import EXTRA_CORS_ALLOWED_HEADERS, EXTRA_CORS_EXPOSE_HEAD
 from localstack.http import Response
 
 from ...constants import LOCALHOST, LOCALHOST_HOSTNAME, PATH_USER_REQUEST
+from ...utils.bootstrap import log_duration
 from ..api import RequestContext
 from ..chain import Handler, HandlerChain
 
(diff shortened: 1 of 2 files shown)


whummer (approved):
Cool! Nice approach that can further streamline the CORS UX. 🎉  

Was able to replicate with the following steps:

1. create a CloudFront distribution, as per our [docs](https://docs.localstack.cloud/user-guide/aws/cloudfront/):
```
$ awslocal s3 mb s3://test
$ echo 'Hello World' > /tmp/hello.txt
$ awslocal s3 cp /tmp/hello.txt s3://test/hello.txt --acl public-read

$ domain=$(awslocal cloudfront create-distribution --origin-domain-name test.s3.amazonaws.com | jq -r '.Distribution.DomainName')
$ echo $domain
ce6554a5.cloudfront.localhost.localstack.cloud
```

2. create a simple `index.html` file that connects to the enpoint:
```
<html>
<script>
fetch('http://ce6554a5.cloudfront.localhost.localstack.cloud:4566/hello.txt');
</script>
</html>
```

3. upload the file to a bucket and expose it as an S3 website:
```
awslocal s3 cp /tmp/index.html s3://test/index.html --content-type text/html
awslocal s3api put-bucket-website --bucket test --website-configuration 'ErrorDocument={Key=index.html},IndexDocument={Suffix=index.html}'
```

4. open http://test.s3-website.localhost.localstack.cloud:4566/index.html in the browser, observe the Network console (request to `ce6554a5.cloudfront.localhost.localstack.cloud:4566`)

Result:
* returns a `403` error on current master  🟥 
* returns `200` and expected file content with the changes in this PR  ✅ 

---
To me, this achieves a decent balance between improving convenience (works out of the box without extra configs), while not sacrificing security (requests still only allowed from `*.localhost.localstack.cloud` pages in the browser). 👍 As usual with CORS or security-critical configurations, would be good to get a second pair of eyes on it, though.. 👀 /cc @dfangl @alexrashed @thrau

localstack/aws/handlers/cors.py
```diff
@@ -75,6 +76,28 @@
 ]
 
 
+def _get_allowed_cors_internal_domains() -> Set[int]:
```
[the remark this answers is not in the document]
bentsku: oops, copy pasted 😄

localstack/aws/handlers/cors.py
```diff
@@ -88,9 +111,8 @@ def _get_allowed_cors_origins() -> List[str]:
     ]
     # Add allowed origins for localhost domains, using different protocol/port combinations.
     # If a different port is configured for EDGE_PORT_HTTP, add it to allowed origins as well
-    _ports = set([config.EDGE_PORT] + ([config.EDGE_PORT_HTTP] if config.EDGE_PORT_HTTP else []))
     for protocol in {"http", "https"}:
-        for port in _ports:
+        for port in _get_allowed_cors_ports():
```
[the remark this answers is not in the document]
bentsku: That's what I thought and did at first, but then it would be generated only once, and in your other tests where you monkey patch `ALLOWED_CORS_ORIGINS`, you'd have to monkey patch `_ALLOWED_INTERNAL_PORTS` before. I thought it'd make your tests a bit more complicated.

https://github.com/localstack/localstack/blob/be7178da606ec4b15b98bd287fc339ad0c870d0d/tests/unit/test_cors.py#L18-L21

bentsku:
Thanks @whummer for creating a reproducible sample!

Just a note:
> (requests still only allowed from *.localhost.localstack.cloud pages in the browser).
The requests are only allowed from `*.s3-website.localhost.localstack.cloud` and `*.cloudfront.localhost.localstack.cloud` for now. 
Also, the webpage served by those 2 "dynamic origins" would be able to not only access cloudfront, but also for example the SQS API, or any API using the default LocalStack configuration.
Just giving precision for the second pair of eyes 👀

localstack/aws/handlers/cors.py
```diff
@@ -166,11 +196,25 @@ def is_cors_origin_allowed(headers: Headers) -> bool:
         return True
 
     @staticmethod
+    @log_duration(min_ms=0)
```
thrau, line 199: nit: could we remove this again?

localstack/aws/handlers/cors.py
```diff
@@ -20,6 +20,7 @@
 from localstack.http import Response
 
 from ...constants import LOCALHOST, LOCALHOST_HOSTNAME, PATH_USER_REQUEST
+from ...utils.bootstrap import log_duration
```
thrau, line 23: nit: ideally we don't use relative imports of modules that are outside the scope of the component we are building. basically: relative imports are used when things belong together in a component, absolute imports are used when importing code that represents a dependency to another component

thrau (approved):
I think this is fantastic :tada: I don't see a big issue right now :+1:

localstack/aws/handlers/cors.py
bentsku: Right, I'm not sure why there are dynamic import here, I can change them 👍

PR #7905, opened 2023-03-18
Add support for VPC endpoints for private API Gateway REST APIs

whummer:
Add support for VPC endpoints for private API Gateway REST APIs. See docs [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html).

These endpoints allow users to invoke REST APIs under the following endpoints:
```
http(s)://{rest-api-id}-{vpce-id}.execute-api.localhost.localstack.cloud:4566/{stage}
```
... in addition to the standard API GW invocation endpoints:
```
http(s)://{rest-api-id}.execute-api.localhost.localstack.cloud:4566/{stage}
```

A snapshot test has been added to cover the functionality.

More detailed description following soon..

### DIFF
--- a/localstack/services/awslambda/lambda_utils.py
+++ b/localstack/services/awslambda/lambda_utils.py
@@ -185,7 +185,6 @@ def store_lambda_logs(
 
 
 def get_main_endpoint_from_container() -> str:
-    global DOCKER_MAIN_CONTAINER_IP
     if config.HOSTNAME_FROM_LAMBDA:
         return config.HOSTNAME_FROM_LAMBDA
     return get_endpoint_for_network(network=get_container_network_for_lambda())
--- a/localstack/services/awslambda/lambda_utils.py
+++ b/localstack/services/awslambda/lambda_utils.py
@@ -14,7 +14,6 @@ from flask import Response
 from localstack import config
 from localstack.aws.accounts import get_aws_account_id
 from localstack.aws.api.lambda_ import FilterCriteria, Runtime
-from localstack.constants import LOCALHOST
 from localstack.services.awslambda.lambda_models import AwsLambdaStore, awslambda_stores
 from localstack.utils.aws import aws_stack
 from localstack.utils.aws.arns import extract_account_id_from_arn, extract_region_from_arn
@@ -188,9 +187,6 @@ def store_lambda_logs(
 def get_main_endpoint_from_container() -> str:
     if config.HOSTNAME_FROM_LAMBDA:
         return config.HOSTNAME_FROM_LAMBDA
-    if config.LAMBDA_EXECUTOR == "local":
-        # special case: return localhost for local Lambda executor (TODO remove after full switch to v2 provider)
-        return LOCALHOST
     return get_endpoint_for_network(network=get_container_network_for_lambda())
 
 
--- a/localstack/services/awslambda/lambda_utils.py
+++ b/localstack/services/awslambda/lambda_utils.py
@@ -14,6 +14,7 @@ from flask import Response
 from localstack import config
 from localstack.aws.accounts import get_aws_account_id
 from localstack.aws.api.lambda_ import FilterCriteria, Runtime
+from localstack.constants import LOCALHOST
 from localstack.services.awslambda.lambda_models import AwsLambdaStore, awslambda_stores
 from localstack.utils.aws import aws_stack
 from localstack.utils.aws.arns import extract_account_id_from_arn, extract_region_from_arn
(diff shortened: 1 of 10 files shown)


tests/integration/apigateway/test_apigateway_integrations.py
```diff
@@ -652,6 +663,189 @@ def test_put_integration_validation(apigateway_client):
...
+    subdomain = f"{api_id}-{endpoint_id}"
+    endpoint = api_invoke_url(subdomain, stage=DEFAULT_STAGE_NAME, path="/test")
+    host_header = urlparse(endpoint).netloc
+
+    # create Lambda function that invokes the API GW (private VPC endpoint not accessible from outside of AWS)
+    if not is_aws_cloud():
+        endpoint = endpoint.replace(
+            host_header, f"{get_main_endpoint_from_container()}:{config.get_edge_port_http()}"
+        )
+    print("!endpoint", endpoint, host_header)
```
whummer, line 800: TODO remove

localstack/services/awslambda/lambda_utils.py
```diff
@@ -185,9 +186,11 @@ def store_lambda_logs(
 
 
 def get_main_endpoint_from_container() -> str:
-    global DOCKER_MAIN_CONTAINER_IP
     if config.HOSTNAME_FROM_LAMBDA:
         return config.HOSTNAME_FROM_LAMBDA
+    if config.LAMBDA_EXECUTOR == "local":
+        # special case: return localhost for local Lambda executor (TODO remove after full switch to v2 provider)
+        return LOCALHOST
```
dfangl, line 193: I am not a fan of this, as it impacts the v2 provider even though it shouldn't. 
Example: People setting `PROVIDER_OVERRIDE_LAMBDA=v2` + `LAMBDA_EXECUTOR=local`. Does not make sense to set it, but it will effectively brick the new provider, although it should not influence it.
I would at least check which provider is active, or better handle this in the calling method of `get_main_endpoint_from_container` in the old provider :)

whummer: Great catch @dfangl , will look into that. The main use case was coming from a [new integration test](https://github.com/localstack/localstack/pull/7905/files#diff-03f6635c4fa95c5cddc4edb72266f76f9962986897cb93eb0293f687022a7e02R798) which needs to pass the LocalStack endpoint to a Lambda function (currently, we're still running the tests with the `local` executor of the old provider). 

Afaict, we currently don't yet have a lookup mechanism that would determine the endpoint for all combinations of `PROVIDER_OVERRIDE_LAMBDA`/`LAMBDA_EXECUTOR` - but as you mentioned, we could probably control this from the calling method (integration test function) for the time being. 👍

whummer: Done @dfangl , removed the switch in `lambda_utils.py`, and added it directly to the test function itself.

tests/integration/apigateway/test_apigateway_integrations.py
```diff
@@ -652,6 +664,191 @@ def test_put_integration_validation(apigateway_client):
...
+                "Principal": "*",
+                "Action": "execute-api:Invoke",
+                "Resource": ["execute-api:/*"],
+            }
+        ],
+    }
+    patches = [{"op": "replace", "path": "/policy", "value": json.dumps(statement)}]
+    result = apigateway_client.update_rest_api(restApiId=api_id, patchOperations=patches)
+    result["policy"] = json.loads(to_bytes(result["policy"]).decode("unicode_escape"))
+    snapshot.match("api-details", result)
```
bentsku, line 837: nit: could we add this jsonpatch operation to the previous call to `update_rest_api`? Honestly, this does not really matter to me, was just a thought. 😄

localstack/services/ec2/provider.py
```diff
@@ -381,6 +395,67 @@ def modify_launch_template(
...
+            account_id=context.account_id, region=context.region
+        )
+
+        moto_backend = get_moto_backend(context)
+        service_names = [s["ServiceName"] for s in ep_services]
+        execute_api_name = f"com.amazonaws.{context.region}.execute-api"
+
+        if execute_api_name not in service_names:
+            # ensure that the service entry for execute-api exists
+            service_names.append(execute_api_name)
```
bentsku, line 414: nit: why do we append the `execute_api_name` here in a list that is dynamically created on every API call?

bentsku (approved):
LGTM! 🎉  Thanks a lot for tackling this! It's so great to see how we can add functionality using 2 different services and it being so clear with ASF now. And so nice to see it AWS validated, and making it easier to fix the parity. Only one minor nit in the ec2 provider, but nothing blocking for me. 🚀

tests/integration/apigateway/test_apigateway_integrations.py
whummer: Yep, could totally be added above as well. 👍 Just separated it out to group the setup into logical units..

localstack/services/ec2/provider.py
whummer: Great catch, thanks - that is totally not necessary here. (Will remove..)

tests/integration/apigateway/test_apigateway_integrations.py
bentsku: Makes sense 👍

PR #7906, opened 2023-03-19
fix config.dirs usage in the CLI

thrau:
This PR adds a band-aid to a long-standing issue that `config.dirs` is used when running CLI commands.

This caused localstack to create the `.filesystem` directory in the Python path, which was originally just designed for executing host mode in developer environments, even when simply using the CLI.

```
 % locate .filesystem     
/home/thomas/.pyenv/versions/3.10.9/lib/python3.10/site-packages/.filesystem
/home/thomas/workspace/localstack/localstack/.filesystem
/home/thomas/workspace/localstack/localstack-ext/.venv/lib/python3.10/site-packages/.filesystem
```

Because the CLI and the runtime share control paths, there are situations where values from `config.dirs` are used to  create directories or files, even though they may not make sense in the context of the CLI.

* `dirs.cache` is used by the analytics metadata (cache machine id file), and was used by the `get_service_catalog` call (now removed)
* `dirs.data` is used by `localstack_ext.config` (this will be removed)
* `dirs.logs` is used as target to put the logs of the docker run command when starting localstack
* `dirs.tmp` should be removed now, but left it in there just in case

There is now a special control path that populates `config.dirs` deliberately with values that make the above examples work without the `.filesystem` directory.
To reduce the number of calls to `config.dirs`, I also removed some unused code from the startup procedure.

Once the old edge proxy logic is removed, we can more easily split up `infra.py` and therefore `bootstrap.py`, which will hopefuly also make it easier to separate out the CLI code.

### DIFF
--- a/localstack/cli/main.py
+++ b/localstack/cli/main.py
@@ -1,3 +1,9 @@
+import os
+
+# indicate to the environment we are starting from the CLI
+os.environ["LOCALSTACK_CLI"] = "1"
+
+
 def main():
     # config profiles are the first thing that need to be loaded
     from .profiles import set_profile_from_sys_argv
--- a/localstack/cli/main.py
+++ b/localstack/cli/main.py
@@ -1,11 +1,11 @@
 import os
 
-# indicate to the environment we are starting from the CLI
-os.environ["LOCALSTACK_CLI"] = "1"
-
 
 def main():
-    # config profiles are the first thing that need to be loaded
+    # indicate to the environment we are starting from the CLI
+    os.environ["LOCALSTACK_CLI"] = "1"
+
+    # config profiles are the first thing that need to be loaded (especially before localstack.config!)
     from .profiles import set_profile_from_sys_argv
 
     set_profile_from_sys_argv()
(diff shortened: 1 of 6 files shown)


localstack/cli/main.py
```diff
@@ -1,3 +1,9 @@
+import os
+
+# indicate to the environment we are starting from the CLI
+os.environ["LOCALSTACK_CLI"] = "1"
```
alexrashed, line 4: This module should only be loaded if the CLI is started, but maybe this should still be moved to the main function body to ensure that this cannot be set accidentally by importing this module?

alexrashed (commented):
The changes are looking good, I only added a minor nitpick. It's great to see this behavior being fixed. I'm really looking forward to extracting the CLI from the runtime and properly isolate the two.
But it seems there is a small thing missing: The pipeline is red because the `LocalstackContainer` class expects the `config.dirs.tmp` to be present for writing the logs (which is now is not anymore). As soon as this is fixed, we're good to go from my perspective!

thrau:
good catch @alexrashed, pushed a fix for your comments

PR #7907, opened 2023-03-19
Disable moto S3 bucket policy enforcement to unblock CDK bootstraps

whummer:
Disable moto's S3 bucket policy enforcement to unblock CDK bootstraps and bump moto-ext to `4.1.5.post1`

Moto recently added more strict bucket policy enforcement: https://github.com/getmoto/moto/pull/5883

This can break `cdklocal bootstrap` commands, as CDK is creating a bucket policy that denies access to bucket objects over non-SSL connections (by enforcing the `aws:SecureTransport` attribute in the policy): https://github.com/aws/aws-cdk/blob/e8158af34eb6402c79edbc171746fb5501775c68/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L217-L233 

After the bootstrap, `cdklocal deploy` can then result in errors in the CloudFormation deployment:
```
  ...
  File ".../localstack/services/cloudformation/provider.py", line 415, in create_change_set
    api_utils.prepare_template_body(
  File ".../localstack/services/cloudformation/api_utils.py", line 23, in prepare_template_body
    modified_template_body = get_template_body(req_data)
  File ".../localstack/services/cloudformation/api_utils.py", line 54, in get_template_body
    result = client.get_object(Bucket=parts[0], Key=parts[2])
  File ".../.venv/lib/python3.10/site-packages/botocore/client.py", line 530, in _api_call
  File ".../.venv/lib/python3.10/site-packages/botocore/client.py", line 960, in _make_api_call
botocore.exceptions.ClientError: An error occurred (403) when calling the GetObject operation: Forbidden
```

This was discovered when deploying some CDK stacks recently, among others this Developer Hub sample here: https://github.com/localstack/microservices-apigateway-lambda-dynamodb-sqs-sample/pull/1 

Seems like a reasonable approach to disable the bucket enforcement entirely (without making it configurable), to unblock CDK users, and also given the recent enhancements in our own IAM enforcement - but open to suggestions! /cc @bentsku @dfangl @thrau


thrau:
@whummer i saw that @viren-nadkarni reverted this commit in moto-ext: https://github.com/localstack/moto/commit/ffca15290bd5431149a73453f98581b33c5f49ec

how do we plan to consolidate the changes in this PR and the commit revert? can we remove the reverted commit and just use the patch? is there some dependence between the two?

whummer:
@thrau Good catch, thanks! (looks like I didn't have that moto change locally yet 🙈 ). To me, either way is fine - undoing the commit revert maybe could make sense (to reduce the number of potential merge conflicts, going forward). 👍 In any case, this PR would only work/build if the commit revert is undone (otherwise it could also be closed.. 🤷 ). Thoughts @viren-nadkarni ?

bentsku:
Before doing a review I'll wait for the decision between the revert of the commit or patching, but I have opinions on the patching.

To make a clear difference between the legacy and the new ASF provider, patches have been duplicated. 
Some are in the `s3_listener/starter`, and other are in the new ASF provider.
I think we should not have a patch in `s3_utils`, as we're trying to sunset it in favour of the new `s3/utils.py`. The goal would be to move "legacy" things into a subfolder `s3/legacy/<files>`, and see what breaks (lots of imports I believe). 

The best would be to add this patch both in `s3_listener` and in `provider` if possible. I'll review once the choice have been made 😄

dfangl:
Viren and I decided to revert the patch for now, as it would have unforseen consequences and does not integrate with our IAM enforcement at all.
Plan was to check the PR in the future and make it interoperable with our IAM enforcement, but this has not happened so far, which is why the revert is still there.
I don't have a specific opinion on patching vs revert though, just to give some context.

viren-nadkarni:
I think this patch makes sense over maintaining the revert commit in moto-ext. There were already quite a few conflicts during the last moto-ext bump (4.1.4).

I will re-release moto-ext with latest 4.1.5 changes and remove the revert commit. We can merge this PR after the bump :+1:

viren-nadkarni:
moto-ext is now bumped to 4.1.5.post1

bentsku:
LGTM too for the code I haven't written 😄 that's one collaborative PR!

Issue #7910, opened 2023-03-20
bug: Cannot CreateGrant with the same name for two different KMS keys


BlueDragon23:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When attempting to CreateGrant on KMS key A, if there is an existing grant with the same name on a _different_ key B, the CreateGrant will return a reference to the already existing grant. Listing grants on key A will show no grants

### Expected Behavior

CreateGrant on key A should successfully create a new grant, instead of returning a reference to the existing one

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
    @pytest.mark.aws_validated
    def test_create_grant_with_same_name_two_keys(self, kms_client, kms_create_key, user_arn):
        first_key_id = kms_create_key()["KeyId"]
        second_key_id = kms_create_key()["KeyId"]

        grant_name = "TestGrantName"

        first_grant = kms_client.create_grant(
            KeyId=first_key_id,
            GranteePrincipal=user_arn,
            Name=grant_name,
            Operations=["Decrypt", "DescribeKey"],
        )
        assert "GrantId" in first_grant
        assert "GrantToken" in first_grant

        second_grant = kms_client.create_grant(
            KeyId=second_key_id,
            GranteePrincipal=user_arn,
            Name=grant_name,
            Operations=["Decrypt", "DescribeKey"],
        )
        assert "GrantId" in second_grant
        assert "GrantToken" in second_grant

        first_grants_after = kms_client.list_grants(KeyId=first_key_id)["Grants"]
        assert len(first_grants_after) == 1

        second_grants_after = kms_client.list_grants(KeyId=second_key_id)["Grants"]
        assert len(second_grants_after) == 1
```

Currently the second grants list is empty

### Environment

```markdown
- OS: MacOS
- LocalStack:1.4.0
```

### Anything else?

I have a patch prepared to fix this issue already along with unit tests. PR incoming shortly

bentsku:
Closed with #7911, thanks a lot for your contribution!

PR #7911, opened 2023-03-20
7910 support having a grant with the same name on different keys

BlueDragon23:
This change addresses this issue https://github.com/localstack/localstack/issues/7910
Previously, due to the `grant_names` only mapping from grant name to grant ID, it was impossible to have more than one grant with the same name across keys. By including the key ID in the mapping, it becomes possible. 

I've added a unit test to demonstrate the fix works, and allows this grant to be created on both keys. 

If any extra test coverage would help, just point me in the right direction!

### DIFF
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -8,7 +8,7 @@ import struct
 import uuid
 from collections import namedtuple
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 from cryptography.exceptions import InvalidSignature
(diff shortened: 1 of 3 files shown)


BlueDragon23:

BlueDragon23:
recheck

BlueDragon23:
It looks like the failing test is related to DynamoDB rest tests, which shouldn't be impacted by this change at all. Is it possible to have that test re-run?

viren-nadkarni (approved):
LGTM!

Thanks for contributing @BlueDragon23!

PR #7913, opened 2023-03-20
Implement basic persistence for lambda v2

dominikschubert:
Adds a first iteration on basic persistence for the new lambda v2 provider.

## Changes

- Start loaded lambda function versions. => Detects lambdas that need to start by checking if they have a VersionManager assigned
- Start event source listeners (note: restoring a dynamodbstreams event source mapping is currently broken because the dynamodbstreams provider doesn't properly support persistence yet) => test skipped for now

## Left to do (in a later PR)
- Implement state reset
- Implement restore for provisioned concurrency (Depends on https://github.com/localstack/localstack/pull/7881)

### DIFF
--- a/localstack/services/awslambda/provider.py
+++ b/localstack/services/awslambda/provider.py
@@ -253,9 +253,9 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                                 exc_info=True,
                             )
 
-                    # Restore event source listeners
-                    for esm in state.event_source_mappings.values():
-                        EventSourceListener.start_listeners_for_asf(esm, self.lambda_service)
+                # Restore event source listeners
+                for esm in state.event_source_mappings.values():
+                    EventSourceListener.start_listeners_for_asf(esm, self.lambda_service)
 
     def on_after_init(self):
         self.router.register_routes()
--- a/localstack/services/awslambda/provider.py
+++ b/localstack/services/awslambda/provider.py
@@ -204,19 +204,12 @@ LAMBDA_TAG_LIMIT_PER_RESOURCE = 50
 LAMBDA_LAYERS_LIMIT_PER_FUNCTION = 5
 
 
-@dataclasses.dataclass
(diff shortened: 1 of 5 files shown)


localstack/services/awslambda/provider.py
```diff
@@ -218,6 +219,38 @@ def __init__(self) -> None:
...
+                            new_config = dataclasses.replace(fn_version.config, state=new_state)
+                            new_version = dataclasses.replace(fn_version, config=new_config)
+                            fn.versions[fn_version.id.qualifier] = new_version
+                            self.lambda_service.create_function_version(fn_version)
+
+                            # Restore event source listeners
+                            for esm in state.event_source_mappings.values():
+                                EventSourceListener.start_listeners_for_asf(
+                                    esm, self.lambda_service
+                                )
```
dfangl, line 252: It seems like this should be independent from the function? Maybe wrong intendation?

dfangl (changes requested):
It seems only "light" merging on a per service state is available now, so we need to shutdown the lambda service in `before_state_load` and then start it up in `after_state_load`.

localstack/services/awslambda/provider.py
```diff
@@ -218,6 +219,44 @@ def __init__(self) -> None:
...
+                for fn in state.functions.values():
+                    for fn_version in fn.versions.values():
+                        # restore the "Pending" state for every function version and start it
+                        try:
+                            new_state = VersionState(
+                                state=State.Pending,
+                                code=StateReasonCode.Creating,
+                                reason="The function is being created.",
+                            )
+                            new_config = dataclasses.replace(fn_version.config, state=new_state)
```
joe4dev, line 243: minor side-effect: might create a new `revision_id` because the state is included in the model

joe4dev (approved):
Lambda persistence tests from branch `lambda-persistence-tests` succeed on macOS `tests.persistence.services.test_lambda` 👍

Issue #7842, opened 2023-03-12
bug: API Gateway V1 REST requests breaks when path parameter validation is enabled in serverless


mickfold:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When performing any requests with path parameters with validation enabled the following error message is returned

`GET http://localhost:4566/restapis/<apigw_id>/local/_user_request_/test/world`

returns

`{"Type": "User", "message": "Invalid request body", "__type": "InvalidRequest"}`

### Expected Behavior

Path parameter validation works like AWS, i.e. validation is performed, lambda is invoked and response returned to caller.

`GET http://localhost:4566/restapis/<apigw_id>/local/_user_request_/test/world`

returns

`"hello world"`

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

   See project in zip archive - [localstack-path-parameter-test.zip](https://github.com/localstack/localstack/files/10952340/localstack-path-parameter-test.zip). Also included below.

To deploy to localstack:

1. `yarn install`
2. `npx sls deploy --stage local --region eu-west-2`

This has two endpoints. The first, test, works but the second endpoint, broken-test, fails with the error `Invalid request body`.

It looks like the error is caused by localstack not being able to handle `AWS::ApiGateway::RequestValidator` correctly.

Also if you try to fix an endpoint by removing the validation, i.e. change the broken-test endpoint to:

```
  broken-test:
    handler: handler.test
    events:
      - http:
          path: broken-test/{value}
          method: get
```

Then after deploying and hitting this endpoint you get an error like

```
{
"__type": "InternalError",
"message": "exception while calling apigateway with unknown operation: An error occurred (NotFoundException) when calling the GetRequestValidator operation: Validator 06ce6f for API Gateway 5ajrq316ln not found"
}
```

**serverless.yml**
```
service: locastack-test
provider:
  name: aws
  runtime: nodejs16.x
  environment:
    NODE_ENV: dev
plugins:
  - serverless-localstack
custom:
  localstack:
    stages: [local]
    host: http://127.0.0.1
    debug: true
functions:
  test:
    handler: handler.test
    events:
      - http:
          path: test/{value}
          method: get
  broken-test:
    handler: handler.test
    events:
      - http:
          path: broken-test/{value}
[...]

PR #7916, opened 2023-03-20
fix APIGW request validation with empty request body

bentsku:
Following #7842, we were missing a step in the validation: if the request body was empty, we should consider the request body as an empty JSON object to allow the validation to take effect, and not throw an error for invalid JSON.

I've extended the existing test for the request validator integration to add the `GET` method, and test with a request with no request body.

### DIFF
--- a/tests/integration/apigateway/test_apigateway_common.snapshot.json
+++ b/tests/integration/apigateway/test_apigateway_common.snapshot.json
@@ -1,6 +1,6 @@
 {
   "tests/integration/apigateway/test_apigateway_common.py::TestApiGatewayCommon::test_api_gateway_request_validator": {
-    "recorded-date": "15-03-2023, 01:01:50",
+    "recorded-date": "20-03-2023, 17:18:54",
     "recorded-content": {
       "register-lambda": {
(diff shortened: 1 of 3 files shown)


tests/integration/apigateway/test_apigateway_common.snapshot.json
```diff
@@ -20,7 +20,7 @@
         "httpMethod": "POST",
         "methodIntegration": {
           "cacheKeyParameters": [],
-          "cacheNamespace": "5l4d79",
```
whummer, line 23: nit: we could add a transformer (e.g., `regex(..)` to store a placeholder for this value (to make the test repeatable against AWS (without updating the snapshot) as well). 👍

whummer (approved):
Great extension for the previous fix, kudos for jumping on this one so quickly! 🚀

tests/integration/apigateway/test_apigateway_common.snapshot.json
bentsku: Right, yes! Good catch. It is repeatable at the moment because we're skipping the verification of the snapshot because methodIntegration format is very not in parity, but I'll add the transformer right now!

PR #7917, opened 2023-03-21
Added test cases and fixed issues for RequestValidator apigateway

sannya-singal:
For `RequestValidator` API in apigateway : 
- Wrote create, list, delete, update, invalid operations, etc. test cases 
- Resolved issues for json patching, invalid messages, conflicts in responses, etc. 
- Fix existing test case- `test_request_validator`

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -1029,9 +1029,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         )
 
         if validator is None:
-            raise NotFoundException(
-                f"Validator {request_validator_id} for API Gateway {rest_api_id} not found"
-            )
+            raise NotFoundException("Invalid Request Validator identifier specified")
 
         result = to_validator_response_json(rest_api_id, validator)
         return result
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -1054,8 +1054,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         validator = RequestValidator(
             id=validator_id,
             name=name,
-            validateRequestBody=validate_request_body,
-            validateRequestParameters=validate_request_parameters,
+            validateRequestBody=validate_request_body or False,
+            validateRequestParameters=validate_request_parameters or False,
         )
 
         rest_api_container.validators[validator_id] = validator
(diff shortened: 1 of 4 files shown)


tests/integration/apigateway/test_apigateway_api.py
```diff
@@ -1528,3 +1528,241 @@ def test_update_model(
...
+        response = apigateway_client.create_request_validator(
+            restApiId=api_id, name=f"test-validator-{short_uid()}"
+        )
+        snapshot.match("create-request-validator", response)
+        validator_id = response["id"]
+
+        response = apigateway_client.get_request_validator(
+            restApiId=api_id, requestValidatorId=validator_id
+        )
+        snapshot.match("get-request-validators", response)
```
bentsku, line 1566: We could add this to `test_create_request_validator`, as it is basically the same test extended with only one operation? Same for `test_get_request_validators`.

I feel like we could condense some of the tests with the flow:
- create a validator
- get the validator
- get all validators
- delete the validator
- try getting the deleted validator
- get all validators
This would look a lot like `test_delete_request_validator`, which cover the same ground as `test_create_request_validator` and `test_get_request_validators`.
With snapshot, we can take advantage of the fact that we can setup the resource one and follow its lifecycle.
We have good examples of this flow in the new lambda tests (see here:)https://github.com/localstack/localstack/blob/814bbe8613c8e955ff32274c7857828139711ef5/tests/integration/awslambda/test_lambda_api.py#L73

Otherwise, it's impeccable and I like how you structured the tests, and it's very clear, thanks a lot for tackling this!

bentsku (commented):
Congratulations on your first contribution!
It seems you've forked LocalStack, could you please push your branch directly to the `localstack/localstack` repo, so that the `integration-tests-against-pro` action could run? (It can run only for branches inside the repo). Because of the nature of API Gateway, it would be safer to run those as we have quite a lot of integrations.

Otherwise, I only have some minor comments about the test structure and a nit about a logic condition. Great use of all the snapshots, this is really great to see! Thanks a lot for tackling this and getting `RequestValidator` in parity with AWS! 🎉

localstack/services/apigateway/provider.py
```diff
@@ -1083,11 +1081,36 @@ def update_request_validator(
                 f"Validator {request_validator_id} for API Gateway {rest_api_id} not found"
             )
 
-        patched_validator = apply_json_patch_safe(validator, patch_operations)
-        rest_api_container.validators[request_validator_id] = patched_validator
+        for operation in patch_operations:
```
calvernaz, line 1084: nit: what if we instead of `operation` use `patch_operation` and after instead of `operation_performed` we can use `operation`

calvernaz (approved):
LGTM! sweet set of snapshot tests 🚀

localstack/services/apigateway/provider.py
```diff
@@ -1083,11 +1081,36 @@ def update_request_validator(
...
+            if key == "name" and not value:
+                raise BadRequestException("Request Validator name cannot be blank")
 
-        result = to_validator_response_json(rest_api_id, patched_validator)
-        return result
+            if key in ["validateRequestParameters", "validateRequestBody"]:
+                if value.lower() == "true":
+                    value = True
+                else:
+                    value = False
```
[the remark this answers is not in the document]
sannya-singal: Thanks for the review @bentsku 🙏🚀🙂 So I had already checked this case before writing the logic above; if we don't specify the value in this case it takes default `False` value, which was covered in the above logic. Let me even write a test case for this.

localstack/services/apigateway/provider.py
sannya-singal: Thanks @calvernaz for the review 🚀🙏🙂

localstack/services/apigateway/provider.py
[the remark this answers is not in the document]
sannya-singal: @bentsku - I have pushed the branch directly to `localstack/localstack` repo: [https://github.com/localstack/localstack/pull/7926]

bentsku (approved):
LGTM! Awesome, thank you for addressing the comments and making the integration tests run! Good to merge 😄

Issue #7902, opened 2023-03-18
bug: Incorrect email count from SES:GetSendQuota after using SES:SendRawEmail


tech-122:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The email count returned from SES:GetSendQuota is double the expected count after using the SES:SendRawEmail  e.g. returns 2 instead of 1 when sending an email to 1 recipient. 

Issue not present when using SES:SendEmail.

### Expected Behavior

The email count returned from SES:GetSendQuota should be the expected count after using the SES:SendRawEmail  e.g. returns 1 instead of 2 when sending an email to 1 recipient.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

docker-compose.yml:
```yaml
version: '3.7'

services:

  localstack:
    image: localstack/localstack:latest
    environment:
      - AWS_DEFAULT_REGION=eu-west-1
      - LOCALSTACK_DEFAULT_REGION=eu-west-1
      - LOCALSTACK_SERVICES=ses
      - LOCALHOST_DOCKER_HOST=unix:///var/run/docker.sock
      - LOCALSTACK_LS_LOG=debug
    ports:
      - '4566:4566'
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock"
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```python
import boto3
from mypy_boto3_ses import Client

CLIENT: Client = boto3.client(service_name='ses', endpoint_url='http://localhost:4566')
EMAIL = 'foo@example.com'

def email_quota():
    resp = CLIENT.get_send_quota()
    print(resp['SentLast24Hours'])

if __name__ == '__main__':
    CLIENT.verify_email_address(EmailAddress=EMAIL)

    email_quota()
    # CLIENT.send_email(
    #     Source=EMAIL,
    #     Destination={
    #         'ToAddresses': ['bar@example.com'],
    #     },
    #     Message={
    #       'Subject': {
    #           'Data': 'string',
    #           'Charset': 'string',
    #       },
    #       'Body': {
    #           'Text': {
    #               'Data': 'string',
    #               'Charset': 'string'
    #           },
    #       },
    #     },
    # )
    # email_quota()

    CLIENT.send_raw_email(
        Source=EMAIL,
        Destinations=['bar@example.com'],
        RawMessage={
[...]

Issue #6287, opened 2022-06-16
Error on SES SendRawEmail without Destinations specified


mmastika:
Hi there,

We're in the process of isolating our build to use LocalStack. In the process we stumbled on an issue where LocalStack would fail with 500 error on SendRawEmail. 

We're able to isolate the issue to what appears to be a LocalStack issue. According to a [post on StackOverflow](https://stackoverflow.com/questions/27446357/how-to-add-cc-and-bcc-in-amazon-ses-sendrawemail), for us to use SendRawEmail and to specify TO, CC and BCC, we must specify it on the on the message/email header, then we MUST not set destinations. This works as expected when we have AWS to test as well as in Production. LocalStack however, produces the following error:

```
2022-06-16T08:26:27.149:WARNING:localstack.utils.server.http2_server: Error in proxy handler for request POST http://localhost:4566/: object of type 'NoneType' has no len() Traceback (most recent call last):
  File "/opt/code/localstack/localstack/utils/server/http2_server.py", line 185, in index
    result = await run_sync(handler, request, data)
  File "/opt/code/localstack/localstack/utils/asyncio.py", line 85, in run_sync
    return await loop.run_in_executor(thread_pool, copy_context().run, func_wrapped)
  File "/opt/code/localstack/localstack/utils/threads.py", line 39, in run
    result = self.func(self.params, **kwargs)
  File "/opt/code/localstack/localstack/utils/asyncio.py", line 30, in _run
    return fn(*args, **kwargs)
  File "/opt/code/localstack/localstack/services/generic_proxy.py", line 993, in handler
    return modify_and_forward(
  File "/opt/code/localstack/localstack/services/generic_proxy.py", line 534, in wrapper
    value = func(*args, **kwargs)
  File "/opt/code/localstack/localstack/services/generic_proxy.py", line 614, in modify_and_forward
    listener_result = listener.forward_request(
  File "/opt/code/localstack/localstack/services/edge.py", line 170, in forward_request
    result = do_forward_request(api, method, path, data, headers, port=port)
  File "/opt/code/localstack/localstack/services/edge.py", line 225, in do_forward_request
    result = do_forward_request_inmem(api, method, path, data, headers, port=port)
  File "/opt/code/localstack/localstack/services/edge.py", line 249, in do_forward_request_inmem
    response = modify_and_forward(
  File "/opt/code/localstack/localstack/services/generic_proxy.py", line 534, in wrapper
    value = func(*args, **kwargs)
[...]

PR #7919, opened 2023-03-21
SES: Fix sent email counter double incrementing

viren-nadkarni:
Remove Moto operation being invoked twice which caused sent email counter to be double incremented.

Fixes https://github.com/localstack/localstack/issues/7902

### DIFF
--- a/tests/integration/test_ses.py
+++ b/tests/integration/test_ses.py
@@ -14,11 +14,25 @@ from localstack.utils.strings import short_uid
 if TYPE_CHECKING:
     from mypy_boto3_ses import SESClient
 
-TEST_TEMPLATE_ATTRIBUTES = {
+SAMPLE_TEMPLATE = {
     "TemplateName": "hello-world",
     "SubjectPart": "Subject test",
     "TextPart": "hello\nworld",
     "HtmlPart": "hello<br/>world",
+}
+
+SAMPLE_SIMPLE_EMAIL = {
+    "Subject": {
+        "Data": "SOME_SUBJECT",
+    },
+    "Body": {
+        "Text": {
+            "Data": "SOME_MESSAGE",
+        },
+        "Html": {
+            "Data": "<p>SOME_HTML</p>",
+        },
+    },
 }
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_ses.py
```diff
@@ -239,6 +244,47 @@ def test_send_templated_email_can_retrospect(self, ses_client, create_template):
...
+        assert new_counter == counter + 2
+        counter = new_counter
+
+        raw_message_data = f"From: {email}\nTo: recipient@example.com\nSubject: test\n\nThis is the message body.\n\n"
+        ses_client.send_raw_email(
+            Destinations=["foo@bar.com"], RawMessage={"Data": raw_message_data}
+        )
+
+        new_counter = ses_client.get_send_quota()["SentLast24Hours"]
+        assert new_counter == counter + 2
```
bentsku, line 286: question: I am confused, wasn't the issue report that the counter should be only +1? Or am I missing something?
After checking in moto, it seems moto is adding recipients from the raw headers to the sum. However, what if the recipient in `Destinations` and in the headers is the same? It seems they would be counted twice. (this previous part was before finding out the answer in the post under). 

Also, after finding an old bug report in LocalStack and this SO post: 
https://github.com/localstack/localstack/issues/6287
https://stackoverflow.com/questions/27446357/how-to-add-cc-and-bcc-in-amazon-ses-sendrawemail

It seems that if we have `Destinations` set then we shouldn't take into account the headers. So the count should be +1 here?

bentsku (commented):
I have one question regarding maybe something out of scope of this PR.
If this is too out of scope, I guess we could simplify the test to not have `Destinations` recipients and only one header `To:`? Or the same example as in the issue, one `Destinations` and no headers.
And create an xfailed test for the wrong behaviour of moto.

tests/integration/test_ses.py
viren-nadkarni: Good catch! This will need fixes both in LS and Moto and a out of scope of this issue. For now I've updated the test and added a fixme in the SES provider

bentsku (approved):
LGTM! Thanks for adding the note 👍

PR #7921, opened 2023-03-21
Removing state serializer interfaces

giograno:
Follow up of https://github.com/localstack/localstack-ext/pull/1500


thrau (approved):
:+1:  :broom: 

i'll target this into v2 since there's some other code removal there as well

PR #7923, opened 2023-03-21
Add pro clients, allow setting of request params through a "proxy"

dfangl:
TBD

### DIFF
--- a/localstack/aws/connect.py
+++ b/localstack/aws/connect.py
@@ -8,8 +8,8 @@ import json
 import logging
 import threading
 from abc import ABC, abstractmethod
-from functools import cache
-from typing import TYPE_CHECKING, Any, Optional, TypedDict
+from functools import cache, partial
+from typing import TYPE_CHECKING, Any, Callable, Generic, Optional, TypedDict, TypeVar, Union
 
 from boto3.session import Session
 from botocore.client import BaseClient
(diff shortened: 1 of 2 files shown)


localstack/aws/connect.py
```diff
@@ -115,6 +161,35 @@ def load_dto(data: str) -> InternalRequestParameters:
...
+    def request_metadata(
+        self, source_arn: str | None = None, service_principal: str | None = None
+    ) -> Union[T, "MetadataRequestInjector[T]"]:
+        params = {}
+        if source_arn:
+            params["_SourceArn"] = source_arn
+        if service_principal:
+            params["_ServicePrincipal"] = service_principal
+        self._params = params
+        return self
```
dfangl, line 190: On second thought, I think it would be better to make this immutable and always return a new object

localstack/utils/aws/client_types.py
```diff
@@ -0,0 +1,194 @@
+import abc
+from typing import TYPE_CHECKING, Union
+
+if TYPE_CHECKING:
+    from mypy_boto3_acm import ACMClient
```
dominikschubert, line 5: Love that these are now separated into their own file. Will keep the churn here much better isolated :+1:

localstack/aws/connect.py
```diff
@@ -115,49 +80,55 @@ def load_dto(data: str) -> InternalRequestParameters:
...
+        else:
+            return target
+
+    def request_metadata(
+        self, source_arn: str | None = None, service_principal: str | None = None
+    ) -> T:
+        """
+        Provides request metadata to this client.
+        Identical as if providing _ServicePrincipal and _SourceArn directly to the operation arguments but typing
+        compatible.
```
dominikschubert, line 106:
```suggestion
        Identical to providing _ServicePrincipal and _SourceArn directly as operation arguments but typing
        compatible.
        
        Raw example: lambda_client.invoke(FunctionName="fn", _SourceArn="...")
        Injector example: lambda_client.request_metadata(source_arn="...").invoke(FunctionName="fn")
```

viren-nadkarni (approved):
LGTM!

PR #7928, opened 2023-03-22
Update PRO image warning

silv-io:
We will start localstack with community services enabled if the `localstack/localstack` image is used instead of `localstack/localstack-pro`. Also, we have a new version-tag file for the new bigdata tag. This PR adapts the existing warning to consider these changes.

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -7,16 +7,14 @@ then
   EDGE_PORT=4566
 fi
 
-# FIXME: remove with 2.0
-# the Dockerfile creates .pro-version file for the pro image. When trying to activate pro features with any other
-# version, an error is printed.
-if [[ $LOCALSTACK_API_KEY ]] && [[ ! -f /usr/lib/localstack/.pro-version ]]; then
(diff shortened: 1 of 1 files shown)


bin/docker-entrypoint.sh
```diff
@@ -7,16 +7,14 @@ then
   EDGE_PORT=4566
 fi
 
-# FIXME: remove with 2.0
-# the Dockerfile creates .pro-version file for the pro image. When trying to activate pro features with any other
-# version, an error is printed.
-if [[ $LOCALSTACK_API_KEY ]] && [[ ! -f /usr/lib/localstack/.pro-version ]]; then
+# the Dockerfile creates .pro-version file for the pro image and .bigdata-pro-version for the bigdata image.
+# When trying to activate pro features with any other version, a warning is printed.
+if [[ $LOCALSTACK_API_KEY ]] && ! compgen -G /usr/lib/localstack/.*pro-version >/dev/null; then
```
alexrashed, line 12: FYI: I just double-checked that the compgen approach with the glob works just fine. 👍🏽

PR #7939, opened 2023-03-23
add fix for CLI tests to properly set CLI context 

thrau:
without setting `LOCALSTACK_CLI=1`, the config.dirs are not set to the CLI directories, but the the host-mode directories using `.filesystem`.

https://github.com/localstack/localstack/blob/a12e673130f5a5c77f6f9a2049d7bfe7bff960a4/localstack/config.py#L1096-L1103

we are setting `LOCALSTACK_CLI=1` here, but this is never called in the tests:

https://github.com/localstack/localstack/blob/a12e673130f5a5c77f6f9a2049d7bfe7bff960a4/localstack/cli/main.py#L4-L6

### DIFF
--- a/tests/bootstrap/conftest.py
+++ b/tests/bootstrap/conftest.py
@@ -1,0 +1,10 @@
+import pytest
+
+from localstack import config
+
+
+@pytest.fixture(autouse=True)
+def _setup_cli_environment(monkeypatch):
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
LGTM! 🚀

Issue #7940, opened 2023-03-23
bug: Log output fails if lambda outputs more than 4000 bytes containing multibyte characters


yamachu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In a SQS Queue with a very long payload containing multibyte characters, the following lambda will fail with log output.

```js
exports.handler = async function(event, context) {
	console.log(event);
	return event;
};
```

docker-compose logs shows ...
```
my-service-localstack-1 | 2023-03-23T13:17:27.211 DEBUG --- [functhread29] l.s.a.lambda_executors     : Calling http://172.28.0.24:9001 to run invocation in docker-reuse Lambda container
my-service-localstack-1 | 2023-03-23T13:17:27.275  INFO --- [functhread29] l.s.awslambda.lambda_api   : Error executing Lambda function arn:aws:lambda:ap-northeast-1:000000000000:function:show-sqs-events-as-records: 'utf-8' codec can't decode byte 0x9f in position 0: invalid start byte Traceback (most recent call last):
my-service-localstack-1 |   File "/opt/code/localstack/localstack/services/awslambda/lambda_api.py", line 459, in run_lambda
my-service-localstack-1 |     result = LAMBDA_EXECUTOR.execute(
my-service-localstack-1 |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 494, in execute
my-service-localstack-1 |     return do_execute()
my-service-localstack-1 |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 484, in do_execute
my-service-localstack-1 |     return _run(func_arn=func_arn)
my-service-localstack-1 |   File "/opt/code/localstack/localstack/utils/cloudwatch/cloudwatch_util.py", line 183, in wrapped
my-service-localstack-1 |     raise e
my-service-localstack-1 |   File "/opt/code/localstack/localstack/utils/cloudwatch/cloudwatch_util.py", line 179, in wrapped
my-service-localstack-1 |     result = func(*args, **kwargs)
my-service-localstack-1 |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 471, in _run
my-service-localstack-1 |     raise e
my-service-localstack-1 |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 467, in _run
my-service-localstack-1 |     result = self._execute(lambda_function, inv_context)
[...]

PR #7941, opened 2023-03-23
Dont abort when failed to decode logs that contains multibyte characters

yamachu:
fix: https://github.com/localstack/localstack/issues/7940

This PR allows for correct handling of lambda runs that output more than 4000 bytes, including multibyte characters.


yamachu:

dfangl (approved):
This looks fine. In general, we do not try to fix things in the old provider, to avoid breaking expected behavior, but this seems like a no-brainer to me.

PR #7942, opened 2023-03-23
Add cache for common functions

dfangl:
Adds a cache to the common functions to avoid building them every build.
Uses a file with a list of all hashes of hash-key-files as hash-key (so hashing the file full of hashes), as CircleCI does not support hashing multiple files (via a glob or something) sadly.
Reduces amd64 build time by about 10 minutes.


alexrashed (approved):
Looks good to me! This will speed up the new pipeline by quite a few minutes for each run! 🚀

PR #7943, opened 2023-03-23
Add back s3 v2 provider alias

dominikschubert:
The intent here is to avoid bricking environments for users that are still setting `PROVIDER_OVERRIDE_S3=v2`. 
Default, `v2` and `asf` are then equivalent and will all start the v2 s3 provider.


bentsku (approved):
LGTM!

Issue #7946, opened 2023-03-23
feature request: Add support for arrays in EventBridge event patterns


bruno-imx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

AWS doc: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-arrays.html

Currently, the event pattern does not match when a message contains the values in an  array

Example:
```
pattern = {"detail": {"event": {"data": {"type": ["1", "2"]}}}}

# should match an event with the following format

{"event": {"data": {"type": ["3", "1"]}}} 

```

as per AWS documentation

> The preceding event pattern matches an event that includes the following field because the first item in the event pattern array matches the second item in the event array.

### 🧑‍💻 Implementation

Suggestion

```
def handle_prefix_filtering(event_pattern, value):
    for element in event_pattern:
        if isinstance(element, (int, str)):
            if str(element) == str(value):
                return True
            if element in value: # added
                return True      # added

       ....
```

if the current `element` belongs to the `value` array in the event-pattern, return true

giograno:
Hi @bruno-imx, was this feature implemented in #7947?

PR #7947, opened 2023-03-24
Adds support for arrays in event bridge rule pattern

bruno-imx:
Suggestion implementation for [issue](https://github.com/localstack/localstack/issues/7946) (Add support for arrays in EventBridge event patterns)

As per AWS [documentation](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-arrays.html), an event pattern that includes the following field

```
"type": ["A", "B"]
```
The preceding event pattern matches an event that includes the following field because the first item in the event pattern array matches the second item in the event array.

```
"type": [ "C",  "A" ]
```


bruno-imx:

bruno-imx:
recheck

tests/integration/test_events.py
```diff
@@ -240,6 +240,39 @@ def test_put_events_with_target_sqs(self, events_client, sqs_client):
             events_client, sqs_client, pattern=TEST_EVENT_PATTERN, entries_asserts=[(entries, True)]
         )
 
+    @pytest.mark.aws_validated
```
dominikschubert, line 243:
```suggestion
```

With this specific test setup, this unfortunately doesn't work against AWS right now, which is why I'm removing the marker here. This is more of an issue with `_put_events_with_filter_to_sqs` though. The event test cases themselves (the 3 entries) should be fine and worked as expected when I tested them in a different test setup.

I'll let this pass for now and will refactor the tests in here soon, so we can mark them all `aws_validated` :+1:

dominikschubert (approved):
Thank you for your contribution! :rocket: 

I'll try to improve the test suite in `test_events.py` soon to be more compatible with running the suite against AWS ( a requirement for the `aws_validated` marker).

PR #7948, opened 2023-03-24
add check for FakeDeleteMarker in case of versioned bucket

bentsku:
While looking at the tests run, I realised we were encountering issue when doing test cleanup. 

Trace:
```python
2023-03-23T21:22:27.7599773Z Traceback (most recent call last):
2023-03-23T21:22:27.7600672Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/chain.py", line 90, in handle
2023-03-23T21:22:27.7601377Z     handler(self, self.context, response)
2023-03-23T21:22:27.7602178Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/handlers/service.py", line 122, in __call__
2023-03-23T21:22:27.7602712Z     handler(chain, context, response)
2023-03-23T21:22:27.7603397Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/handlers/service.py", line 92, in __call__
2023-03-23T21:22:27.7603934Z     skeleton_response = self.skeleton.invoke(context)
2023-03-23T21:22:27.7604622Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/skeleton.py", line 153, in invoke
2023-03-23T21:22:27.7605130Z     return self.dispatch_request(context, instance)
2023-03-23T21:22:27.7605835Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/skeleton.py", line 165, in dispatch_request
2023-03-23T21:22:27.7606339Z     result = handler(context, instance) or {}
2023-03-23T21:22:27.7607016Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/forwarder.py", line 60, in _call
2023-03-23T21:22:27.7607483Z     return handler(context, req)
2023-03-23T21:22:27.7608151Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/aws/skeleton.py", line 117, in __call__
2023-03-23T21:22:27.7608616Z     return self.fn(*args, **kwargs)
2023-03-23T21:22:27.7609473Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/services/s3/provider.py", line 485, in delete_objects
2023-03-23T21:22:27.7610083Z     s3_notification_ctx = S3EventNotificationContext.from_request_context(
2023-03-23T21:22:27.7610903Z   File "/home/runner/work/localstack-ext/localstack-ext/localstack/localstack/services/s3/notifications.py", line 127, in from_request_context
2023-03-23T21:22:27.7611457Z     key_etag=key.etag.strip('"'),
2023-03-23T21:22:27.7612135Z AttributeError: 'FakeDeleteMarker' object has no attribute 'etag'
2023-03-23T21:22:27.7613013Z 2023-03-23T21:22:27.760  INFO --- [   asgi_gw_2] localstack.request.aws     : AWS s3.DeleteObjects => 500 (InternalError)
```

After some reproductions, I could reproduce the issue when the bucket is versioned, which we, it turns out, don't support very well. While migrating the S3 provider, we did not account very much for the concept of `VersionId`, and we did not have tests covering it (not a single operation was done with `VersionId` parameter).

So I've added some checks and fixed some direct usages of moto `FakeKey` object which can be `FakeDeleteMarker`, a case we didn't account for. We can fix a lot more when using versioned bucket, but at least now we shouldn't raise direct `AttributeError`. 

Concerning notifications, we do not support `s3:ObjectRemoved:DeleteMarkerCreated` at the moment and we will send a `s3:ObjectRemoved:Delete` instead, but this will need a bit of refactoring in the notification handling (in the map determining the service operation, it would depend if the bucket is versioned or not and if we delete a key or delete a key version). 

I guess we can see this as a bandaid, we don't implement versioned buckets in a better way for now (except for `PutObjectAcl`), but at least we don't raise exceptions linked to that. We will need to come back to this to better support the use case with some usage numbers.

Sources:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeleteMarker.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html#supported-notification-event-types

### DIFF
--- a/localstack/services/s3/utils.py
+++ b/localstack/services/s3/utils.py
@@ -1,6 +1,6 @@
 import datetime
 import re
-from typing import Dict, Union
+from typing import Dict, Literal, Union
 
 import moto.s3.models as moto_s3_models
 from botocore.exceptions import ClientError
@@ -13,6 +13,7 @@ from localstack.aws.api.s3 import (
     BucketName,
     ChecksumAlgorithm,
     InvalidArgument,
+    MethodNotAllowed,
     NoSuchBucket,
     NoSuchKey,
     ObjectKey,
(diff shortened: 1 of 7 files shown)


localstack/services/s3/utils.py
```diff
@@ -140,19 +141,41 @@ def get_bucket_from_moto(
...
 
 
 def get_key_from_moto_bucket(
-    moto_bucket: moto_s3_models.FakeBucket, key: ObjectKey
-) -> moto_s3_models.FakeKey:
-    fake_key = moto_bucket.keys.get(key)
+    moto_bucket: FakeBucket,
+    key: ObjectKey,
+    version_id: str = None,
+    raise_if_delete_marker_method: Literal["GET", "PUT"] = None,
```
bentsku, line 151: I really dislike this, but this is a quick fix for now and would need to be revisited to better implement versioning.

thrau (approved):
great find! and looks like a pretty reasonable fix for now! :medal_sports: for all the snapshot tests.

PR #7950, opened 2023-03-24
Fix test failures for community tests in PRO

dominikschubert:
Fixes community tests with PRO enabled.

- Skip transparent endpoint injection tests when PRO enabled (with lambda v2) and DNS is disabled
- Fix stepfunctions state machine to return correct payload structure.


dfangl (approved):
LGTM

Issue #7915, opened 2023-03-20
bug: Error creating route when using multiple subnets in one vpc


cgoIT:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

First of all I'm not a neither an AWS nor a terraform expert. So maybe I'm the problem and not localstack. ;)

I've created to vpc "foo" and "bar" where both vpcs have two subnets (a private and a public one). I want to setup vpc peering between those two vpcs. I'm able to create the peering connection. After that I would like to add the corresponding routes to the route tables where the subnets are associated to.

During the creation of the routes localstack complains about the fact, that the desired destination cidrs already exist. If I query the existing route tables I can see that this is not the case.

### Expected Behavior

Routes should be created.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
version: '3.1'
services:
   localstack:
      image: localstack/localstack-pro:latest
      environment:
         - TZ=${TZ:-Europe/Berlin}
         - LOCALSTACK_API_KEY=${LOCALSTACK_API_KEY:- }
         - REQUIRE_PRO=1
         - DEBUG=${DEBUG:-0}
      ports:
         - "127.0.0.1:4566-4583:4566-4583"
         - "127.0.0.1:53:53"
         - "127.0.0.1:443:443"
      volumes:
         - "/var/run/docker.sock:/var/run/docker.sock"
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

My terraform script:
```
###################### Peering Connection #####################

resource "aws_vpc_peering_connection" "foo_bar" {
  peer_vpc_id   = aws_vpc.bar.id
  vpc_id        = aws_vpc.foo.id
  auto_accept = true

  accepter {
    allow_remote_vpc_dns_resolution = true
  }

  requester {
    allow_remote_vpc_dns_resolution = true
  }
}



###################### Peering #####################

data "aws_route_table" "foo_public" {
  subnet_id = aws_subnet.foo_public.id
}

data "aws_route_table" "foo_private" {
  subnet_id = aws_subnet.foo_private.id
}

data "aws_route_table" "bar_public" {
  subnet_id = aws_subnet.bar_public.id
}

data "aws_route_table" "bar_private" {
  subnet_id = aws_subnet.bar_private.id
}


resource "aws_route" "foo_public_to_bar" {
[...]

Issue #6076, opened 2022-05-17
bug: EC2 DescribeInstances without any parameters does not work


trivelt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I'm trying to list EC2 instances without `InstanceIds` parameter (which is not mandatory) and I use for example:

```
aws --endpoint-url=http://localhost:4566 ec2 describe-instances --region=eu-west-2
```

Localstack returns the following error:

> An error occurred (InvalidSubnetID.NotFound) when calling the DescribeInstances operation: The subnet ID 'subnet-f4a71320' does not exist

It happens after terminating the instance and deleting a subnet used by that instance.

### Expected Behavior

The result should be similar to the output produced by the real AWS, which lists all EC2 instances in a specified region (using filters if specified), even if some subnets were deleted.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

1. Initially describe-instances works as expected
```
aws --endpoint-url=http://localhost:4566 ec2 describe-instances --region=eu-west-2
```
2. Create new VPC
```
aws --endpoint-url=http://localhost:4566 ec2 create-vpc --cidr-block "10.1.0.0/16"
# Created VPC with vpc-id = vpc-dea45193
```
3. Create new subnet
```
aws --endpoint-url=http://localhost:4566 ec2 create-subnet --vpc-id vpc-dea45193 --cidr-block "10.1.1.0/27" --tag-specifications "ResourceType=subnet,Tags=[{Key=Name,Value=my-ipv4-subnet}]"
# Created subnet with subnet-id = 9303cf84
```
4. Create new instance using the subnet created above
```
aws --endpoint-url=http://localhost:4566 ec2 run-instances --image-id ami-0abcdef1234567890  --instance-type t2.micro --key-name MyKeyPair --subnet-id subnet-9303cf84
# Created instance with instance-id = i-a74fe925c78dbb733
```
5. Terminate instance
```
aws --endpoint-url=http://localhost:4566 ec2 terminate-instances --instance-ids i-a74fe925c78dbb733
```
6. Delete subnet
```
aws --endpoint-url=http://localhost:4566 ec2 delete-subnet --subnet-id subnet-9303cf84
```
7. Now describe-instance raises an error:
```
aws --endpoint-url=http://localhost:4566 ec2 describe-instances --region=eu-west-2
[...]

PR #7952, opened 2023-03-24
Bump moto-ext to 4.1.5.post2

viren-nadkarni:
Moto is now aware of Event Buses in Eventbridge. This meant some of our tests needed updates.

Includes upstream changes that resolve:
- https://github.com/localstack/localstack/issues/6076
- https://github.com/localstack/localstack/issues/7915


thrau (approved):
LGTM! not sure right now how we can get a heads up on whether this will cause issue in localstack-pro, @alexrashed ?

PR #7955, opened 2023-03-24
Extend docker/net utils to allow port checks for UDP ports

whummer:
Extend docker/net utils to allow port checks for UDP ports. Addresses an issue from a support channel where starting up the LS container fails due to UDP port 53 already being bound. 

This PR extends `container_ports_can_be_bound(..)` to allow us to make port availability checks also for UDP ports, which will allow us to selectively add the port bind for UDP port 53, based on availability.

Tried to make the changes minimally invasive, and backwards-compatible. Ports can be specified as an `int` (in which case we assume TCP as the default protocol), or as a `Port` instance that contains the protocol explicitly.

The existing `TestDockerPorts` test has been parameterized to also cover UDP ports, and a new simple test has been added to cover the UDP-based logic for `port_can_be_bound(..)`.

Debatable whether this is a new feature or a fix 🤷 - but since it addresses a support case, arguably this could still be considered for the pre v2 merges (despite the feature freeze). To be discussed..

/cc @simonrw @dfangl

### DIFF
--- a/localstack/utils/net.py
+++ b/localstack/utils/net.py
@@ -3,7 +3,7 @@ import re
 import socket
 import threading
 from contextlib import closing
-from typing import List, Optional, Union
+from typing import Any, List, MutableMapping, NamedTuple, Optional, Union
 from urllib.parse import urlparse
 
 import dns.resolver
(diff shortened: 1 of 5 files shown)


localstack/utils/net.py
```diff
@@ -88,7 +108,7 @@ def is_port_open(
 def wait_for_port_open(
     port: int, http_path: str = None, expect_success=True, retries=10, sleep_time=0.5
 ):
-    """Ping the given network port until it becomes available (for a given number of retries).
+    """Ping the given TCP network port until it becomes available (for a given number of retries).
```
simonrw, line 111: Perhaps the functions in this file could all be updated to take `IntOrPort` (or `Port` if they are not part of the public API), then the entire utility file would be protocol agnostic. What do you think? Perhaps an update for a future PR.

simonrw (approved):
Looks good, a minor suggestion about the API surface of `localstack.utils.net` but not blocking

localstack/utils/net.py
whummer: Good point, agreed. Would probably leave that for a future iteration.. 👍

PR #7959, opened 2023-03-24
skip test_package_installer_custom_lock to unblock CI pipeline

thrau:
Skipping this test for now since there seems to be a race condition in the test that sometimes triggers. Was also able to reproduce locally, but not reliably.

/cc @alexrashed


bentsku (approved):
LGTM! I was a witness in the reproduced deadlock 😄

PR #7960, opened 2023-03-24
fix client metadata to always return an empty string for api_key

thrau:
in #7693 we changed the behavior of `read_client_metadata` to return `None` as key if pro is not available. previously it was always an empty string. our data schema, unfortunately, as api key as a non-nullable field, so this leads to rows landing in quarantine.


ackdav (approved):
great catch! lgtm 🚀

PR #7961, opened 2023-03-24
fix CORS issue for CreateBucket and ListBuckets

bentsku:
It seems we missed a scenario when creating the new CORS handling for S3:

If we do a `CreateBucket` request not targeting the S3 specific endpoint (like we do from the web app), we cannot know for sure that the request is targeting S3 (it is a simple `PUT` request with a path, nothing else), because the bucket does not exist yet. So, the pre-process handler does not add any headers, and the request is seen as "self-managed" by the default CORS enforcer / enricher because it targets S3, so nobody adds the right headers in the end. 

The solution for now is to let the handler chain execute before handing the request to the skeleton, and if the request is `CreateBucket` or `ListBuckets`, we will use the default LocalStack CORS handling to enrich the response with the right headers, only of the origin is allowed. Otherwise, we will reject the request. At least we won't get any side effect directly in the provider. 

\cc @lukqw, thanks for finding and helping in reproducing the issue!

### DIFF
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -118,10 +118,6 @@ def should_enforce_self_managed_service(context: RequestContext) -> bool:
     if context.service:
         service_name = context.service.service_name
         if not config.DISABLE_CUSTOM_CORS_S3 and service_name == "s3":
-            # ListBuckets is not concerned by S3 CORS handling, it should follow general LocalStack CORS rules.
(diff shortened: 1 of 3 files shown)


thrau (approved):
as discussed, i think this is a reasonable solution for now! getting CORS right with s3 in combination with `localhost:4566` and other routes is tricky in our handler chain, so having a bit of workarounds in s3 seems fine to me!

PR #7962, opened 2023-03-24
skip test_kinesis_firehose_opensearch_s3_backup + es equivalent + fix s3 flake

bentsku:
Firehose tests with OpenSearch/ES have been really flaky and we're trying to get a deployment out. Let's see if this goes further

Trying to fix flakiness of the s3 CORS tests that make extensive use of virtual host addressing... could be some issue there?


thrau (approved):
Thanks! Need to get the pipeline 💚

PR #7965, opened 2023-03-24
fix: dynamodb restart

giograno:
This PR fixes the restart of DynamoDB. 

I noticed this issue when working to re-enable the cloud pods tests. I likely introduced this regression in https://github.com/localstack/localstack/pull/7827.

In a nutshell, we were creating the assets directory in the `on_before_state_reset()` hook and deleting it when calling the visitor. This PR moves the creation to the server to the `on_after_state_reset()` hook.

One can simply test it by creating some resources and calling the endpoint with `curl -X POST http://localhost:4566/_localstack/state/reset`


thrau (approved):
LGTM!

PR #7968, opened 2023-03-25
fix race condition in service plugin loading

thrau:
This PR fixes a race condition in the service plugin loading, where multiple concurrent calls to `service_manager.require(...)` could lead to Service plugins being instantiated more than once, and subsequently `on_after_init` being called multiple times, whose implementation often mutates global state.

We haven't seen this manifest yet, other an exception in the logs when running `tests/integration/test_terraform.py::TestTerraform::test_bucket_exists`:

```
2023-03-24T17:22:07.962 ERROR --- [   asgi_gw_5] localstack.utils.functions : error calling function on_after_init
  File "/opt/code/localstack/localstack/utils/functions.py", line 41, in call_safe
    return func(*args, **kwargs)
  File "/opt/code/localstack/localstack/services/apigateway/provider.py", line 97, in on_after_init
    apply_patches()
  File "/opt/code/localstack/localstack/services/apigateway/patches.py", line 270, in apply_patches
    def get_rest_api(self, function_id):
  File "/opt/code/localstack/localstack/utils/patch.py", line 170, in wrapper
    fn.patch = Patch.function(target, fn, pass_target=pass_target)
  File "/opt/code/localstack/localstack/utils/patch.py", line 96, in function
    return Patch(obj, name, new)
  File "/opt/code/localstack/localstack/utils/patch.py", line 56, in __init__
    self.old = getattr(self.obj, name)
AttributeError: module 'localstack.services.apigateway.patches' has no attribute 'get_rest_api'
```

which was an obscure indicator that `on_after_init` was called twice.

The PR introduces a locking mechanism that makes sure that, while the service is being loaded and instantiated, other calls to `require` of the same service block. The `SynchronizedDefaultDict(RLock)` pattern is helpful here - and we make sure we delete the locks after using them (we only need them once).

### DIFF
--- a/localstack/services/plugins.py
+++ b/localstack/services/plugins.py
@@ -16,7 +16,7 @@ from localstack.state import StateLifecycleHook, StateVisitable, StateVisitor
 from localstack.utils.bootstrap import get_enabled_apis, log_duration
 from localstack.utils.functions import call_safe
 from localstack.utils.net import wait_for_port_status
-from localstack.utils.sync import poll_condition
+from localstack.utils.sync import SynchronizedDefaultDict, poll_condition
 
 # set up logger
 LOG = logging.getLogger(__name__)
@@ -245,7 +245,7 @@ class ServiceContainer:
 class ServiceManager:
     def __init__(self) -> None:
         super().__init__()
-        self._services = {}
+        self._services: Dict[str, ServiceContainer] = {}
         self._mutex = threading.RLock()
 
     def get_service_container(self, name: str) -> Optional[ServiceContainer]:
@@ -454,6 +454,11 @@ class ServicePluginManager(ServiceManager):
         self._api_provider_specs = None
         self.provider_config = provider_config or config.SERVICE_PROVIDER_CONFIG
 
+        # locks used to make sure plugin loading is thread safe - will be cleared after single use
+        self._plugin_load_locks: Dict[str, threading.RLock] = SynchronizedDefaultDict(
+            threading.RLock
+        )
+
     def get_active_provider(self, service: str) -> str:
         """
         Get configured provider for a given service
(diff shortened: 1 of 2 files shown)


localstack/services/plugins.py
```diff
@@ -542,24 +547,32 @@ def get_state(self, name: str) -> Optional[ServiceState]:
...
-                super().add_service(plugin.service)
+            # this is where we start lazy loading. we now know the PluginSpec for the API exists,
+            # but the ServiceContainer has not been created.
+            # this control path will be executed once per service
+            plugin = self._load_service_plugin(name)
+            if not plugin or not plugin.service:
+                return None
+
+            with self._mutex:
+                if plugin.service.name() not in self._services:
```
whummer, line 570: Wondering if there can be a situation where this condition is _not_ true, i.e., where an entry for this service name would already exist in `self._services` (since we're now locking on the service `name` in this critical section, and `_load_service_plugin(..)` in line 565 above is side-effect free). 

The only case where this could happen is if `plugin.service.name()` returns a string that is different from `name` - can this be the case?

If that's something we're not expecting to happen, maybe we could proactively raise an error, to avoid such cases from going unnoticed:
```
            with self._mutex:
                if plugin.service.name() in self._services:
                    raise ServiceStateException(
                        "Service {plugin.service.name()} unexpectedly already in list of loaded services")
                    )
                super().add_service(plugin.service)
```

whummer (approved):
Great catch @thrau 🚀 . Nice approach with the `SynchronizedDefaultDict`, and great to see it covered by a test. Only one question / suggestion around introducing explicit guards around `self._services`..

localstack/services/plugins.py
thrau: good point, can't happen anymore. fixed!

PR #7978, opened 2023-03-27
Don't source (LOCALSTACK_)HOST in docker entrypoint

simonrw:
Before this change, the entrypoint script would strip the `LOCALSTACK_` prefix from `LOCALSTACK_HOST` and try to set the `HOST` environment variable to something. This is bad as the `HOST` enviornment variable is reserved for system things, and should not be set.


thrau (approved):
LGTM!

PR #7981, opened 2023-03-27
add test for ListMultipartUploads

bentsku:
While checking https://docs.localstack.cloud/references/coverage/coverage_s3, it appeared `ListMultipartUploads` was not implemented. It seems it is, and I've added a test for coverage of that operation.

\cc @steffyP @joe4dev


steffyP (approved):
thanks for adding even more snapshot tests 🥳 🚀 🙏 📸 ❤️ 
test LGTM!

PR #7545, opened 2023-01-24
fix: rename list bucket/object reponse class

macnev2013:
# Issue
- #7541 

# Fix
- Patched `ListBucketsOutput` to `ListAllMyBucketsResult`.
  - In boto spec definition of output from `ListBuckets` is `ListBucketsOutput` whereas AWS is returning  `ListAllMyBucketsResult`.
- `list_objects` contains `ListBucketResult` object whereas we are returning `ListObjectsOutput` 

# Commands to Reproduce

```
Start localstack instance with env `PROVIDER_OVERRIDE_S3: "asf"`
```

Comparing the response from this commands will po
`list_bucket`
```
awslocal s3api list-buckets --profile ls --debug
```
```
aws s3api list-buckets --profile ls --debug
```

`list-objects`
```
awslocal s3 mb s3://test-bucket --profile ls --region us-east-1 --debug
awslocal s3api list-objects --bucket test-bucket --profile ls --region us-east-1 --debug
```
```
aws s3 mb s3://test-bucket --profile ls --region us-east-1 --debug
aws s3api list-objects --bucket test-bucket --profile ls --region us-east-1 --debug
```

# Tested Also with
https://github.com/Dowwie/localstack_s3_xml_issue

PR #7983, opened 2023-03-27
fix response structure for S3 ListMultipartUploads

bentsku:
This is a follow up from #7545, a user reported an issue in our Community Slack. The test is AWS validated, and once again the specs were not compliant with what S3 really returns. 

We also reapply the logic that was in a patch previously set, moto would have raised an error if using some request parameters that were not implemented. It would raise a `NotImplemented` exception. The previous provider was swallowing the exception and letting it pass anyway without applying the parameters. We can keep the same behaviour for now and come back later and actually implement the right behaviour. 

related: #7981, https://github.com/localstack/docs/pull/539

\cc @steffyP

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import os
 from typing import IO, Dict, List
@@ -56,9 +57,12 @@ from localstack.aws.api.s3 import (
     InvalidPartOrder,
     InvalidStorageClass,
     ListBucketResult,
+    ListMultipartUploadsRequest,
+    ListMultipartUploadsResult,
     ListObjectsRequest,
     ListObjectsV2Request,
     MissingSecurityHeader,
+    MultipartUpload,
     NoSuchBucket,
     NoSuchKey,
     NoSuchLifecycleConfiguration,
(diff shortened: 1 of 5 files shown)


localstack/services/s3/provider.py
```diff
@@ -556,6 +561,60 @@ def complete_multipart_upload(
...
+    @handler("ListMultipartUploads", expand=False)
+    def list_multipart_uploads(
+        self,
+        context: RequestContext,
+        request: ListMultipartUploadsRequest,
+    ) -> ListMultipartUploadsResult:
+
+        try:
+            response: ListMultipartUploadsResult = call_moto(context)
+        except NotImplementedError:
```
thrau, line 573: just a question: do we need this try/except? is this method implemented in moto, and are we only using it partially? or can we just replace `call_moto` with our implementation all together?

thrau (approved):
LGTM! another nice catch :-) just wondering about whether we need moto here

localstack/services/s3/provider.py
bentsku: It's partially implemented in moto, and raises only if some parameters are passed:
https://github.com/localstack/moto/blob/localstack/moto/s3/responses.py#L428-L432

We could replace it I suppose because I think if there's an exception, we're basically doing what it would do if there wasn't an exception. Good idea!

bentsku: I guess it was in the case they would add more things, because right now our implementation is incomplete, and if moto adds feature we wouldn't get them.

PR #7984, opened 2023-03-27
Introduce new AWS client test fixtures

dominikschubert:
Introduces 3 new fixtures that build upon another

1. `aws_session` .. can be used to modify the session for all test clients
2. `aws_client_factory` ... can be used if programmatic client creation, a specific region or credentials  are required.
3. `aws_client` ... use for the most common use-case where we previously had to use multiple different clients for each service.

```python
def my_test(sqs_client, lambda_client, logs_client, iam_client, ...):
    sqs_client.create_queue(...)
    lambda_client.create_function(...)
    ...
```

becomes

```python
def my_test(aws_client):
    aws_client.sqs.create_queue(...)
    aws_client.awslambda.create_function(...)
    ...
```

## Changes
- Add new client fixtures (`aws_session`, `aws_client_factory`, `aws_client`)
- Remove old `_client` and a few related methods
- Add new client factory for external AWS clients (non-internal AWS-targeting clients)

## TODO (other PRs)
- [ ] Remove all `x_client` fixtures and replace with single `aws_client` across integration tests.
- [ ] Remove all `create_external_boto_client` occurences
- [ ] Remove all usage of resource clients in favor of normal clients (e.g. dynamodb or s3)
- [ ] Refactor fixtures in PRO to use the new clients


localstack/aws/connect.py
```diff
@@ -289,11 +290,11 @@ def _get_client(
...
-        verify: bool,
-        endpoint_url: str,
-        aws_access_key_id: str,
-        aws_secret_access_key: str,
-        aws_session_token: str,
+        verify: Optional[bool],
+        endpoint_url: Optional[str],
+        aws_access_key_id: Optional[str],
+        aws_secret_access_key: Optional[str],
+        aws_session_token: Optional[str],
```
viren-nadkarni, line 297: I think it's a good idea to keep these args explicit as they form part of the cache key

dominikschubert: You'll still need to provide them in this case since they don't have default values. In some cases (e.g. against AWS) we need to set them to None for boto to set them automatically. The None value in this case will just be part of the cache key

viren-nadkarni (approved):
I can already feel the dopamine of removing all those fixtures!

localstack/aws/connect.py
```diff
@@ -396,16 +399,26 @@ def get_client(
...
 
+        if config is None:
+            config = self._config
+        else:
+            config = self._config.merge(config)
+
+        endpoint_url = endpoint_url or get_local_service_url(service_name)
+        if service_name == "s3":
+            if re.match(r"https?://localhost(:[0-9]+)?", endpoint_url):
+                endpoint_url = endpoint_url.replace("://localhost", "://%s" % get_s3_hostname())
```
dfangl, line 410: Why not use an fstring here?

localstack/testing/pytest/fixtures.py
```diff
@@ -47,68 +48,34 @@
...
-    from mypy_boto3_sts import STSClient
-    from mypy_boto3_transcribe import TranscribeClient
-
 LOG = logging.getLogger(__name__)
 
 # URL of public HTTP echo server, used primarily for AWS parity/snapshot testing
 PUBLIC_HTTP_ECHO_SERVER_URL = "http://httpbin.org"
 
 
-def _client(service, region_name=None, aws_access_key_id=None, *, additional_config=None):
```
dfangl, line 92: Did you make sure nothing in -ext uses this?

dfangl (approved):
LGTM! We have to be careful not to break -ext tests here, though, with the removal of the _client and related methods

localstack/aws/connect.py
dominikschubert: copypasta :spaghetti: 

I'll update it though :+1:

thrau (approved):
LGTM! i like the `aws_clients.<service>` pattern, which i think is going to make tests lot cleaner, and also the fixtures moving forward.

in principle it makes sense to me, though i'm wondering, since the boto3 session is associated with a particular set of credentials and region, whether re-using the same Session across the entire test suite may presents any type of limitation in the future?

localstack/aws/connect.py
```diff
@@ -396,16 +401,26 @@ def get_client(
...
 
+        if config is None:
+            config = self._config
+        else:
+            config = self._config.merge(config)
+
+        endpoint_url = endpoint_url or get_local_service_url(service_name)
+        if service_name == "s3":
+            if re.match(r"https?://localhost(:[0-9]+)?", endpoint_url):
+                endpoint_url = endpoint_url.replace("://localhost", f"://{get_s3_hostname()}")
```
bentsku, line 412: don't we all love S3? 😄 thanks for adding the fix here, like we already talked, not sure how we could make this cleaner, I think this will do for now

localstack/aws/connect.py
```diff
@@ -442,17 +457,77 @@ def get_client(
...
+
+
+class ExternalAwsClientFactory(ClientFactory):
+    def get_client(
+        self,
+        service_name: str,
+        region_name: Optional[str],
+        aws_access_key_id: Optional[str] = None,
+        aws_secret_access_key: Optional[str] = None,
+        aws_session_token: Optional[str] = None,
```
bentsku, line 490: question: we are not using those, right? As we explicitly set the client creation credentials to None to load them from our local configuration. Should we update the docs string to convey that?

bentsku (approved):
LGTM! just a small question concerning the doc strings for the external client, but nothing major. Thanks for tackling this, this will make things cleaner 👍

localstack/aws/connect.py
dominikschubert: :+1: 

I'll sneak this into a later PR :wink:

dominikschubert: actually did this now, since I'm re-running the tests in -ext as well

PR #7985, opened 2023-03-27
do not merge: debug flaky s3 cors

bentsku:
just running to CI with more debug logs to try to understand the flaky S3 CORS which return 404 after creating objects
edit: obviously doesn't fail at that time. Will relaunch tomorrow during the day?


bentsku:
Closing as it was fixed with #8134

PR #7990, opened 2023-03-28
Add utility for creating a client assuming a role, migrate some inter-service communication to new client

dfangl:
## Motivation
* Assuming a role and using a client with the returned credentials is a pattern we often need across the code base, especially for the use of service roles or service-linked roles. The new method provides a simple interface to do this in one call.
* To enforce IAM between some services, we need to use the new aws client, and also handle the exceptions (it should fail with a warning, but should not crash anything important if a request fails).

### connect_to.with_assumed_role
A new method for the aws client factories.

Usage:

```python
logs_client = connect_to
    .with_assumed_role(role_arn="arn:aws:iam::000000000000:role/test-role", service_principal="lambda")
    .logs
```

This call will assume the given role as the given service, and return a service level client factory, where various clients can be created using the returned credentials.

The assume role call will be done immediately, the other clients instantiated lazily as they are requested.

### Client migrations

Several important inter-service calls between lambda, sqs and sns are migrated to the new clients.
This migration is ongoing, and not complete yet.

During this, I silently killed `SYNCHRONOUS_SNS_EVENTS`, as it is scheduled to remove with 2.0 anyway.

### Fix check if installed check outside of installer lock
This one is a tricky one.
Basically, we are circumventing the lock if the `is_installed()` method already returns true. But, the default is just checking if a directory is created, which can be true during the installation as well. Moving this entirely behind the lock avoids one thread installing while the other one is checking, and therefore avoids starting incompletely installed dependencies.

This seems to be the root cause for the failing `test_kinesis_firehose_elasticsearch_s3_backup` test, which is so annoying. It is basically trying to start elasticsearch while we are still installing plugins -> class loading failure.

### DIFF
--- a/localstack/services/awslambda/invocation/version_manager.py
+++ b/localstack/services/awslambda/invocation/version_manager.py
@@ -96,14 +96,24 @@ class LogHandler:
         self._thread = None
 
     def run_log_loop(self, *args, **kwargs) -> None:
-        logs_client = connect_to(region_name=self.region).logs
+        logs_client = connect_to.with_assumed_role(
+            region_name=self.region, role_arn=self.role_arn, service_principal="lambda"
+        ).logs
         while not self._shutdown_event.is_set():
             log_item = self.log_queue.get()
             if log_item is QUEUE_SHUTDOWN:
                 return
-            store_cloudwatch_logs(
-                log_item.log_group, log_item.log_stream, log_item.logs, logs_client=logs_client
-            )
+            try:
+                store_cloudwatch_logs(
+                    log_item.log_group, log_item.log_stream, log_item.logs, logs_client=logs_client
+                )
+            except Exception as e:
+                LOG.warning(
+                    "Error saving logs to group %s in region %s: %s",
+                    log_item.log_group,
+                    self.region,
+                    e,
+                )
 
     def start_subscriber(self) -> None:
         self._thread = FuncThread(self.run_log_loop, name="log_handler")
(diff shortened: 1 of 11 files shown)


tests/unit/aws/test_connect.py
```diff
@@ -283,24 +283,17 @@ def echo_request_handler(_: HandlerChain, context: RequestContext, response: Res
         endpoint_url = create_dummy_request_parameter_gateway([echo_request_handler])
 
         # TODO this should be extracted into a utility for the next iteration
```
dominikschubert, line 285:
```suggestion
```

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -527,14 +537,20 @@ def process_event_destinations(
...
+                    source_arn=self.function_arn,
+                    dlq_arn=self.function_version.config.dead_letter_arn,
+                    event=json.loads(to_str(original_payload)),
+                    error=InvocationException(
+                        message="hi", result=to_str(invocation_result.payload)
+                    ),  # TODO: check message
+                    role=self.function_version.config.role,
+                )
+            except Exception as e:
+                LOG.warning(
```
dominikschubert, line 551: :+1: 

Definitely good to start catching a few more of these exceptions and properly log them. Especially for things started asynchronously in a separate thread.

dominikschubert (approved):
LGTM! :rocket: 

I like the new pattern. One issue I'm starting to see a bit is the manual specification of the service principal as a string literal. Maybe it would make sense to keep a list of all current AWS service principals and use an enum for these. Sometimes they really aren't as intuitive on AWS as they should be :grimacing:  (e.g. states for stepfunctions)

localstack/services/awslambda/invocation/version_manager.py
dfangl: Yes. I had some problems with the logging, as it used too many permissions, and it would just crash the whole thread, not good :)

alexrashed (approved):
The package installer API change looks good to me! Thanks for tackling this! 🚀

PR #7992, opened 2023-03-28
Improve Lambda deprecation warnings

joe4dev:
Add and improve lambda deprecation warnings ....


localstack/config.py
```diff
@@ -735,100 +737,128 @@ def get_gateway_listen(gateway_listen: str) -> List[HostAndPort]:
 # Disable the check for MaxNumberOfMessage in SQS ReceiveMessage
 SQS_DISABLE_MAX_NUMBER_OF_MESSAGE_LIMIT = is_env_true("SQS_DISABLE_MAX_NUMBER_OF_MESSAGE_LIMIT")
 
-# host under which the LocalStack services are available from Lambda Docker containers
+# DEPRECATED: only applies to old lambda provider
+# Endpoint host under which LocalStack APIs are accessible from Lambda Docker containers.
 HOSTNAME_FROM_LAMBDA = os.environ.get("HOSTNAME_FROM_LAMBDA", "").strip()
```
dominikschubert, line 742: not sure if we're not actually gonna have to keep that for some edge cases :thinking:

localstack/services/awslambda/invocation/lambda_service.py
```diff
@@ -247,6 +247,12 @@ def invoke(
         except ValueError:
             version = function.versions.get(version_qualifier)
             state = version and version.config.state.state
+            # TODO: make such developer hints optional or remove after initial v2 transition period
+            LOG.warning(
+                "Lambda functions are created and updated asynchronously in the new lambda provider. "
+                "Please wait until the function is in the Active state before invoking. Check out "
+                "https://docs.localstack.cloud/references/lambda-provider-v2"
```
dominikschubert, line 254: Love this


I'd go a step further and show them a sample like `aws lambda wait function-active-v2 --function-name <actual-fn-name>`. This obviously won't apply to all use cases (e.g. SDKs) but might give them a good starting point, even if the docs are broken for some reason.

localstack/services/awslambda/plugins.py
```diff
@@ -20,3 +26,18 @@ def lambda_java_libs() -> Package:
     from localstack.services.awslambda.packages import lambda_java_libs_package
 
     return lambda_java_libs_package
+
+
+@hooks.on_infra_start()
+def validate_configuration() -> None:
+    if LAMBDA_EXECUTOR == "local":
```
dominikschubert, line 33: any reason this only applies to `local` ?

localstack/config.py
```diff
@@ -914,6 +953,8 @@ def get_gateway_listen(gateway_listen: str) -> List[HostAndPort]:
 # TODO remove fallback to LAMBDA_DOCKER_NETWORK with next minor version
 MAIN_DOCKER_NETWORK = os.environ.get("MAIN_DOCKER_NETWORK", "") or LAMBDA_DOCKER_NETWORK
 
+# Add deprecated environment variables to deprecations.py
+
```
dominikschubert, line 957:
```suggestion
```

not sure if this was intentional

dominikschubert (approved):
Just a few minor nits, otherwise LGTM.

Looking forward to extending these kind of inline suggestions in the logs. We might have to think about adding a separate logger for those though, so we're able to disable them if some might find them too intrusive.

One thing that I'm also a bit curious about is how stable the links to the docs pages will be since we don't really have a strategy in place for redirecting older pages, etc. so this might be outdated quite quickly.

localstack/config.py
joe4dev: agreed. We might want adopt the new `host` terminology instead of `hostname` in that case 🤔

localstack/services/awslambda/plugins.py
joe4dev: `LAMBDA_EXECUTOR` is part of the regular deprecation warning and this is an emphasis for the local executor because it does not require the Docker socket mounting.
We can easily extend the additional warning for other values as well.

localstack/services/awslambda/invocation/lambda_service.py
joe4dev: Good idea to demonstrate the waiter solution as concretely as possible. Added an example:
Screenshot 2023-03-28 at 21 56 04

joe4dev:
> One thing that I'm also a bit curious about is how stable the links to the docs pages will be since we don't really have a strategy in place for redirecting older pages, etc. so this might be outdated quite quickly.

Valid point because that reference becomes implicit knowledge. I introduced aliases when doing the Lambda doc re-work for backwards compatibility. I typically try to support page renames but anchors within a page might break after some refactoring.

dfangl (approved):

Issue #7993, opened 2023-03-28
feature request: Let us use custom regions again


toadzky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

I prefer not to use the real regions when doing tests or local stuff to make sure that if it accidentally picks up real credentials, it won't actually create or interact with resources in AWS. This used to work find in localstack - I even dug through the code to find that there is an option in the `RegionStore` to disable validation of the region string. At some point, you removed any way to disable that validation though, which means it just fails all requests using a custom region.

### 🧑‍💻 Implementation

There's already a flag that is passed to the account and region store constructors to disable validation. Seems like an environment variable or configuration flag to disable validation could be propagated.

toadzky:
thanks

PR #7997, opened 2023-03-29
Support non-standard AWS regions

viren-nadkarni:
Introduces a new config option `ALLOW_NONSTANDARD_REGIONS` which disables strict validation of AWS regions.

Closes https://github.com/localstack/localstack/issues/7993

### DIFF
--- a/tests/integration/test_stores.py
+++ b/tests/integration/test_stores.py
@@ -1,0 +1,17 @@
+import unittest
+
+
+@unittest.mock.patch("localstack.config.ALLOW_NONSTANDARD_REGIONS", True)
+def test_nonstandard_regions(create_boto_client, monkeypatch):
+    """
+    Ensure that non-standard AWS regions can be used vertically.
+    """
+    monkeypatch.setenv("MOTO_ALLOW_NONEXISTENT_REGION", "true")
+
+    # Create a resource in Moto backend
+    ec2_client = create_boto_client("ec2", region_name="uranus-south-1")
+    ec2_client.create_key_pair(KeyName="foo")
+
+    # Create a resource in LocalStack store
+    sqs_client = create_boto_client("sqs", region_name="pluto-central-2a")
+    sqs_client.create_queue(QueueName="bar")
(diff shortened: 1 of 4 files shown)


tests/integration/test_stores.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+
+
+@unittest.mock.patch("localstack.config.ALLOW_NONSTANDARD_REGIONS", True)
+def test_nonstandard_regions(create_boto_client, monkeypatch):
+    """
+    Ensure that non-standard AWS regions can be used vertically.
+    """
+    monkeypatch.setenv("MOTO_ALLOW_NONEXISTENT_REGION", "true")
```
viren-nadkarni, line 9: Ideally we want to patch the config option before LS starts. What's a proper way to do this?

tests/integration/test_stores.py
```diff
@@ -0,0 +1,17 @@
...
+
+    # Create a resource in Moto backend
+    ec2_client = create_boto_client("ec2", region_name="uranus-south-1")
+    ec2_client.create_key_pair(KeyName="foo")
+
+    # Create a resource in LocalStack store
+    sqs_client = create_boto_client("sqs", region_name="pluto-central-2a")
```
thrau, line 16: :D

tests/integration/test_stores.py
thrau: unfortunately there is no good way of doing as at all. especially not in a test. config.py seems good for the runtime

thrau (approved):
nice addition!

toadzky:
thanks guys

tests/integration/test_stores.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+
+
+@unittest.mock.patch("localstack.config.ALLOW_NONSTANDARD_REGIONS", True)
```
dominikschubert, line 4: @viren-nadkarni 

we usually use `monkeypatch.setattr(config, "ALLOW_NONSTANDARD_REGIONS", True)` here.
Any reason you're using unittest in this case?

viren-nadkarni: No particular reason at all, it just didn't cross my mind. I'll fix it when I retouch this next time

PR #7998, opened 2023-03-29
Ignore empty provider overrides

silv-io:
Currently, when explicitly setting `""` for provider overrides (e.g., `PROVIDER_OVERRIDE_S3=""`) the ServiceProviderConfig assumes that a provider with the name `""` (i.e. an empty string) is requested. This happens, e.g., in the `bin/test-in-docker.sh` script and leads to `NotImplementedError`s. 
Since we can safely assume that we won't create new provider variants with the explicit name of an empty string, we can ignore all falsey values for provider overrides and thus assume a selection of the default provider. 

This PR also adds a unit test to verify this new behavior.

/cc @bentsku

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1226,7 +1226,7 @@ class ServiceProviderConfig(Mapping[str, str]):
         if env is None:
             env = os.environ
         for key, value in env.items():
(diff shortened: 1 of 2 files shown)


dfangl (approved):
LGTM

PR #7999, opened 2023-03-29
remove DISABLE_TERM_HANDLER in docker entrypoint

alexrashed:
Removes a small leftover from #7970: The `DISABLE_TERM_HANDLER` env variable handling in the `docker-entrypoint.sh`.
This feature seems to be a weird feature, LocalStack should just listen to all signals.

Docs PR: https://github.com/localstack/docs/pull/543


thrau (approved):
LGTM! I totally thought I had removed that :thinking:

PR #8003, opened 2023-03-29
cleanup coverage-docs scripts

steffyP:
* removes scripts that have prior been used to create coverage report and aggregate metrics
* adapt tinybird-upload script to take a directory with raw-metrics as input

TODO:

- [x] adapt docs script, as we now need to download + handle two raw-metrics-test files https://github.com/localstack/docs/pull/546
- [ ] merge docs pr after successful merge


alexrashed (approved):
Awesome! Great to see the legacy scripts and build workflow steps being removed and the scripts being cleaned up a bit! 🧹

PR #8004, opened 2023-03-29
Ignore publish version code hash if hot-reloading is active

dfangl:
## Motivation
Some tools, for example serverless, seem to send the local code hash with them when using publish version. This comparison will fail, since hot-reloading code does not have a proper hash (but a placeholder string).

Related comment: https://github.com/localstack/serverless-localstack/pull/215#issuecomment-1488142179

## Changes
* Skip code hash verification if code is hot reloading + test
* Sneaked in a opensearch test update to make sure the cluster is properly terminated after the test


dominikschubert (approved):
LGTM :+1:

PR #8005, opened 2023-03-29
create separate cache directory for cli

thrau:
This PR adds a separate cache dir for the CLI. I saw errors in the log:

```
[16:16:34] starting LocalStack in Docker mode 🐳                                                                                                                                                  localstack.py:142
2023-03-29T16:16:34.656  WARN --- [  MainThread] localstack.utils.functions : error calling function save: [Errno 13] Permission denied: '/home/thomas/.cache/localstack/cli'
```

because `/home/thomas/.cache/localstack` is created by the container. So instead, we now have `/home/thomas/.cache/localstack-cli` which is managed by the cli and won't have permission errors.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -161,7 +161,7 @@ class Directories:
         from localstack.utils import files
 
         tmp_dir = os.path.join(tempfile.gettempdir(), "localstack-cli")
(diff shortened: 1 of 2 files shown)


dfangl (approved):
Seems sensible, LGTM!

PR #8006, opened 2023-03-29
Add basic cloudwatch metrics for new lambda provider

dominikschubert:
SImply restores parity with the old provider in regards to the CloudWatch metrics we record in the service.

Records two metrics (same as the old provider):

 - `Invocations` ... a simple numeric counter. Records 1 for each invocation result (both  success & result)
 - `Errors` ... simple numeric counter. Records 1 for each invocation error. This also applies to failed retries

The only other change to the old provider here is that we now connect to the correct region when sending the records.

Would be a cool project to extend this and add proper parity with AWS for all the service level metrics.


steffyP (approved):
Thanks for jumping on this so quickly! 🙏 
LGTM :)

PR #8007, opened 2023-03-29
fix signal handling in localstack-supervisor

thrau:
@dfangl noticed that SIGINT is not terminating localstack correctly in docker. this PR makes sure that SIGINT to the supervisor is translated into a SIGTERM to localstack. we tried to forward SIGINT but couldn't quite figure it out. left a todo in there for future daniel and thomas :P

We also now use SIGUSR1 instead of SIGHUP to restart LS, to avoid restarting when the shell is closed (it will now just be ignored. We could start terminating, we can discuss if this is expected).

/cc @alexrashed

### DIFF
--- a/bin/localstack-supervisor
+++ b/bin/localstack-supervisor
@@ -3,7 +3,7 @@
 localstack. This can be used on the host or in the docker-entrypoint.sh.
 
 The supervisor behaves as follows:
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Wow, great catch! Looks good to me! 🚀

PR #8008, opened 2023-03-29
Limit retries when trying to start new lambda environments

dominikschubert:
Trying to increase stability of the environment spawn process. It's a bit of a bandaid until we get around to refactor the invocation loop, but should avoid some of the issues we've been seeing in CI and with infinite environment spawn retries.  

## Changes

- start_environment is sync again to avoid multiple spawns for a single invocation if spawning the environment takes longer than a second
- limit number of attempts for starting a new environment when no environment is available

### DIFF
--- a/localstack/services/awslambda/invocation/version_manager.py
+++ b/localstack/services/awslambda/invocation/version_manager.py
@@ -52,6 +52,8 @@ if TYPE_CHECKING:
 
 LOG = logging.getLogger(__name__)
 
+MAX_ENVIRONMENT_START_ATTEMPTS = 3
+
 
 @dataclasses.dataclass(frozen=True)
 class QueuedInvocation:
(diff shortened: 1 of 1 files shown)


localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -404,7 +406,15 @@ def invocation_loop(self, *args, **kwargs) -> None:
 
                 environment = None
                 # TODO avoid infinite environment spawning retrying
```
joe4dev, line 408: That TODO should be fixed with this PR

localstack/services/awslambda/invocation/version_manager.py
```diff
@@ -404,7 +406,15 @@ def invocation_loop(self, *args, **kwargs) -> None:
...
+                environment_start_attempt_nr = 0
                 while not environment:
+                    environment_start_attempt_nr += 1
```
joe4dev, line 411: That counter increases without actually creating a new environment. I only got one container despite `LAMBDA_REMOVE_CONTAINERS=0` and got the exception "Exceeded maximum retries ..."

PR #8009, opened 2023-03-29
Log reported runtime init error

dominikschubert:
Depends on https://github.com/localstack/lambda-runtime-init/pull/17

- [x] Update init after PR above is released


joe4dev (approved):
LGTM 👍 
✅ Tested against https://github.com/localstack/lambda-runtime-init/pull/17/

Very helpful addition.

👉  Update RIE before merging

PR #8012, opened 2023-03-30
Remove unneeded defaults for envars

simonrw:
Currently we get a deprecation warning shown in docker because we set
EDGE_BIND_HOST and LOCALSTACK_HOST in the dockerfile. We default to these
values anyway so we don't need them in the Dockerfile any more.


alexrashed (approved):
LGTM! 🚀

PR #8015, opened 2023-03-30
Add config option to override used lambda init version

dominikschubert:
Adds the `LAMBDA_INIT_RELEASE_VERSION` config variable. This value will override the default version which was previously hard-coded in the lambda service. We still only really support a 1-1 mapping of localstack version to init version (always whatever is set as `LAMBDA_RUNTIME_DEFAULT_VERSION` at the time of the localstack release), but sometimes it's helpful to test an existing container with a newer released init version without updating localstack first.


dfangl (approved):

thrau (approved):
Configure all the things!

Issue #8019, opened 2023-03-30
bug: AWS S3 GetObjectTaggingRequest randomly returning null


dsarlo-viso:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This morning, I upgraded to localstack version 2.0.0.

We have a test case that uploads a file to S3 (localstack) with tags and checks to make sure those tags are set properly by doing the following on the Java AWS S3 client `s3client.getObjectTagging(GetObjectTaggingRequest(bucketName, location)).tagSet`.

The call to grab object tags in S3 throws the following exception - 

```java.lang.NullPointerException: Cannot invoke "com.amazonaws.services.s3.model.GetObjectTaggingResult.getTagSet()" because the return value of "com.amazonaws.services.s3.AmazonS3.getObjectTagging(com.amazonaws.services.s3.model.GetObjectTaggingRequest)" is null

When running the application locally, it throws a NPE every single time.

Using the environment var `PROVIDER_OVERRIDE_S3=legacy` to use the legacy version of S3 works just fine.

### Expected Behavior

_No response_

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)
</br>
docker-compose.yml contains the following:

        localstack:
          container_name: localstack
          image: localstack/localstack:2.0.0
          restart: always
          ports:
              - 4566:4566
              - 4571:4571
          environment:
              - SERVICES=s3,sqs,lambda
              - AWS_DEFAULT_REGION=us-west-2

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    Using the following commands doesn't seem to trigger the issue. Only the AWS S3 Java SDK
    
    awslocal s3api get-object-tagging --bucket test --key "test.csv"


### Environment

```markdown
- OS: MacOS 13.2.1
- LocalStack: 2.0.0
```

bentsku:
Hey @dsarlo-viso and thanks for your report!

It seems our response format follows the specs of the response, but it's not what AWS really returns.
I'm currently working on a fix.
The issue is similar to this one: #7983

I'll link the PR to this issue! Thanks again for your report.

bentsku:
Hi @dsarlo-viso, the fix has been merged and should be available in the `latest` tag of the Docker image. Could you give it a try and see if that solves your issue with the new S3 provider? Thanks a lot again for the report!

[...]

PR #8021, opened 2023-03-30
fix GetObjectTagging and GetBucketTagging format

bentsku:
No need to introduce this one anymore, same issue as always. Extending #7983

I think we need to create a small script to test every single S3 operations to check for the XML of the response, and not only what boto returns, or check the documentation of every operation. 

This addresses #8019 

Format in the docs:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectTagging.html#API_GetObjectTagging_ResponseElements

Format from botocore specs:
https://github.com/boto/botocore/blob/a77b1766b2e34a55b3c9d3aed89a8d2376a8eed6/botocore/data/s3/2006-03-01/service-2.json#L4920-L4935


macnev2013 (approved):
LGTM 🚀

PR #8022, opened 2023-03-30
Refactor integration tests to use new client fixtures

dominikschubert:
Depends on #7984

Refactors fixture usage in integration tests from:

```python
def test_something(lambda_client, logs_client):
    lambda_client.list_functions()
    logs_client.filter_log_events(...)
    ...
```

to

```python
def test_something(aws_client):
    aws_client.awslambda.list_functions()
    aws_client.logs.filter_log_events(...)
    ...
```

Still need to adapt some of the fixtures in a next PR, but most of the normal test methods should be done by now

### DIFF
--- a/tests/integration/apigateway/test_apigateway_api.py
+++ b/tests/integration/apigateway/test_apigateway_api.py
@@ -77,7 +77,7 @@ def apigw_create_rest_api(apigateway_client):
 
 
 @pytest.mark.aws_validated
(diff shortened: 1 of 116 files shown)


thrau (approved):
nice! cleans up tests, a few look unchanged, but overall great improvement! :+1: 

we should get this over the line soon before conflicts pile up.

PR #8028, opened 2023-03-31
fix SNS Subscription Filter OR/AND combination on MessageBody

bentsku:
This issue came up in the support channel, we were not handling OR conditions properly.
We now properly handle combination of OR and AND filters for the `MessageBody` scope. 

I've also added a unit test to properly test the new filter. 

See: https://docs.aws.amazon.com/sns/latest/dg/and-or-logic.html


thrau (approved):
LGTM! thanks for jumping on this.

that unit test is a good candidate for pytest parameterization :-)

PR #8029, opened 2023-03-31
fix SNS FIFO DLQ + add SequenceNumber to FIFO

bentsku:
This PR adds the ability to pass a failed message from a FIFO topic down to a FIFO queue. It was failing before because we would not pass the required attributes needed for a FIFO queue, namely `MessageGroupId` and optionally `MessageDeduplicationId`.

I've also removed some snapshot skipping by adding the `SequenceNumber` attributes to messages published on FIFO topics. Needed to add a field on `SnsMessage`, not very clean to pass the context there, but it feels like the only solution to be able to return that `SequenceNumber` to the `PublishBatch` operation response, it needs to be pre-defined before actually creating the message body. 

I already added the test for this behaviour, so I just removed the `xfail` mark.

### DIFF
--- a/localstack/services/sns/models.py
+++ b/localstack/services/sns/models.py
@@ -1,3 +1,5 @@
+import itertools
+import time
 from dataclasses import dataclass, field
 from typing import Dict, List, Literal, Optional, TypedDict, Union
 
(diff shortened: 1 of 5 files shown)


thrau (approved):
LGTM!

Issue #8030, opened 2023-03-31
bug: CreateMultipartUpload should return InitiateMultipartUploadResult element


pguillory:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Initiating a multipart S3 upload to LocalStack returns an XML document with a root element CreateMultipartUploadOutput. This causes an error in the ex_aws_s3 library I'm using. We've used the combination of LocalStack + ex_aws_s3 successfully for a few years and the error started occurring a few days ago.

### Expected Behavior

The element should be called InitiateMultipartUploadResult.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run -p 4566:4566 localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    aws --debug --endpoint-url=http://localhost:4566 s3api create-multipart-upload --bucket my-bucket --key 'multipart/01'

### Environment

```markdown
- OS: Ubuntu 20.04.5 LTS
- LocalStack: latest
```

### Anything else?

S3 docs specifying InitiateMultipartUploadResult:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html

Reference in official AWS Java client:
https://github.com/aws/aws-sdk-java/blob/ef22b94750d4c155d74b09012b3be5d9d98bbd33/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/transform/XmlResponsesSaxParser.java#L1865

Where our particular error gets raised in ex_aws_s3:
https://github.com/ex-aws/ex_aws_s3/blob/7e12f2b0578b2620c047d2e44c40a6de8773584a/lib/ex_aws/s3/parsers.ex#L78

bentsku:
Hi @pguillory and thanks for your report! I've answered in your PR as well: https://github.com/localstack/localstack/pull/8032#issuecomment-1492774815

We are currently taking a look at all the operations returning a root XML node different that was is specified in the specs, and will try to correct most of them quickly. Thanks again for your report!

bentsku:
Hi @pguillory, the PR has now been merged, the fix should be available in the `latest` Docker image. You can pull it and give a try! Thanks a lot for the kind words on your PR as well 😄

pguillory:
Verified the fix works and our tests are passing. Thanks again for your help!

[...]

PR #8032, opened 2023-03-31
Rename CreateMultipartUploadOutput to InitiateMultipartUploadResult

pguillory:
The response to a CreateMultipartUpload request should contain a root element InitiateMultipartUploadResult. This will align with the AWS API docs and with client libraries that check the root element.

https://github.com/localstack/localstack/issues/8030

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -1194,7 +1194,7 @@ class CreateBucketRequest(ServiceRequest):
     ObjectOwnership: Optional[ObjectOwnership]
 
 
-class CreateMultipartUploadOutput(TypedDict, total=False):
+class InitiateMultipartUploadResult(TypedDict, total=False):
     AbortDate: Optional[AbortDate]
     AbortRuleId: Optional[AbortRuleId]
     Bucket: Optional[BucketName]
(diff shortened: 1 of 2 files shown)


pguillory:

bentsku:
Hi @pguillory and thanks for your contribution!

However, the code you are updating is auto-generated and updated every week, based on the botocore specs.
We need to modify this data in another way, through patches.

Here are some examples on how we've already updated some responses: #8021, #7983 and #7545.
You'd also need to regenerate the API with the following command inside LocalStack virtual environment: `python -m localstack.aws.scaffold generate --save s3`. 

I can also take a stab a it tomorrow if you'd like, as it is a bit tedious and also quite urgent. I'm currently planning to try to collect all XML responses from every operation to compare them to the specs and correct all of the remaining ones we are missing.
It seems `CompleteMultipartUploadResult` will also be wrong compared to `CompleteMultipartUploadOutput` from the specs.

bentsku:
There's now a PR which would fix all S3 operations root tag at #8037.

pguillory:
Great, closing in favor of #8037. Thanks for your help @bentsku and everyone maintaining LocalStack! It's been very helpful in enabling us to write high quality unit tests for systems that touch S3, of which we have a lot.

Issue #7963, opened 2023-03-24
feature request: Support Update operation for IAM Policy in Cloudformation


pinzon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Current CloudFormation implementation doesn't support the **UPDATE** operation on a CFn Iam::Policy Resource.

deepak2431:
@pinzon I have checked this out, as the current CFn Iam model doesn't have the update/add policy implemented in their configurations after the creation. We can use the IAM service CreatePolicyRequest to attach the policies through the cloud formation template.  Can I take this up to work on?

pinzon:
The solution to this issue has been merged.

PR #8036, opened 2023-04-01
Cfn::IAM::Policy update

deepak2431:
As requested by @pinzon in #7963 implement support for the Update operation on IAM policy. I was working on it along with the help of @pinzon. After a few discussions, we decided to make a test that would create a policy for the user/group/role and again deploy a template to update it.  I also tried a test where we are making use of `put_user_policy` for the same user, and that also works fine.

Performing this operation on the current implementation of Cfn::IAM models works perfectly fine. So, I think we can conclude that the current implementation supports the Update operation. 

What do you think @thrau @pinzon ?

### DIFF
--- a/tests/integration/cloudformation/resources/test_iam.py
+++ b/tests/integration/cloudformation/resources/test_iam.py
@@ -83,7 +83,7 @@ def test_policy_attachments(
 
 
 @pytest.mark.aws_validated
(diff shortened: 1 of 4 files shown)


deepak2431:
Closing this PR due to a lot of merge conflicts after rebase

PR #8037, opened 2023-04-01
fix S3 response xml root tag names

### LINKED ISSUE #8030: bug: CreateMultipartUpload should return InitiateMultipartUploadResult element

bentsku:
So following #8021, #7983 and #7545, I've investigated and made a list of all S3 operations with a different root node name than the specs. Basically, 41 of 44 operations returning an XML body do not follow the specs. Here's the list:
❌: wrong name, 🟠: fixed by patches, ✅: right from the specs
| | Operation | Spec shape | Root tag
|-|-|-|-|
| ❌| CompleteMultipartUpload | CompleteMultipartUploadOutput | CompleteMultipartUploadResult
| ❌| CopyObject | CopyObjectOutput | CopyObjectResult
| ❌| CreateMultipartUpload | CreateMultipartUploadOutput | InitiateMultipartUploadResult
| 🟠| DeleteObjects | DeleteObjectsOutput | DeleteResult
| ❌| GetBucketAccelerateConfiguration | GetBucketAccelerateConfigurationOutput | AccelerateConfiguration
| ❌| GetBucketAcl | GetBucketAclOutput | AccessControlPolicy
| ❌| GetBucketAnalyticsConfiguration | GetBucketAnalyticsConfigurationOutput | AnalyticsConfiguration
| ❌| GetBucketCors | GetBucketCorsOutput | CORSConfiguration
| ❌| GetBucketEncryption | GetBucketEncryptionOutput | ServerSideEncryptionConfiguration
| ❌| GetBucketIntelligentTieringConfiguration | GetBucketIntelligentTieringConfigurationOutput | IntelligentTieringConfiguration
| ❌| GetBucketInventoryConfiguration | GetBucketInventoryConfigurationOutput | InventoryConfiguration
| ❌| GetBucketLifecycle | GetBucketLifecycleOutput | LifecycleConfiguration
| ❌| GetBucketLifecycleConfiguration | GetBucketLifecycleConfigurationOutput | LifecycleConfiguration
| - | GetBucketLocation | GetBucketLocationOutput | LocationConstraint
| ❌| GetBucketLogging | GetBucketLoggingOutput | BucketLoggingStatus
| ❌| GetBucketMetricsConfiguration | GetBucketMetricsConfigurationOutput | MetricsConfiguration
| ❌| GetBucketNotification | NotificationConfigurationDeprecated | NotificationConfiguration
| ✅| GetBucketNotificationConfiguration | NotificationConfiguration | NotificationConfiguration
| ❌| GetBucketOwnershipControls | GetBucketOwnershipControlsOutput | OwnershipControls
| ❌| GetBucketPolicyStatus | GetBucketPolicyStatusOutput | PolicyStatus
| ❌| GetBucketReplication | GetBucketReplicationOutput | ReplicationConfiguration
| ❌| GetBucketRequestPayment | GetBucketRequestPaymentOutput | RequestPaymentConfiguration
| 🟠| GetBucketTagging | GetBucketTaggingOutput | Tagging
| ❌| GetBucketVersioning | GetBucketVersioningOutput | VersioningConfiguration
| ❌| GetBucketWebsite | GetBucketWebsiteOutput | WebsiteConfiguration
| ❌| GetObjectAcl | GetObjectAclOutput | AccessControlPolicy
| ✅| GetObjectAttributes | GetObjectAttributesOutput | GetObjectAttributesOutput
| ❌| GetObjectLegalHold | GetObjectLegalHoldOutput | LegalHold
| ❌| GetObjectLockConfiguration | GetObjectLockConfigurationOutput | ObjectLockConfiguration
| ❌| GetObjectRetention | GetObjectRetentionOutput | Retention
| 🟠| GetObjectTagging | GetObjectTaggingOutput | Tagging
| ❌| GetPublicAccessBlock | GetPublicAccessBlockOutput | PublicAccessBlockConfiguration
| ❌| ListBucketAnalyticsConfigurations | ListBucketAnalyticsConfigurationsOutput | ListBucketAnalyticsConfigurationResult
| ✅| ListBucketIntelligentTieringConfigurations | ListBucketIntelligentTieringConfigurationsOutput | ListBucketIntelligentTieringConfigurationsOutput
| ❌| ListBucketInventoryConfigurations | ListBucketInventoryConfigurationsOutput | ListInventoryConfigurationsResult
| ❌| ListBucketMetricsConfigurations | ListBucketMetricsConfigurationsOutput | ListMetricsConfigurationsResult
| 🟠| ListBuckets | ListBucketsOutput | ListAllMyBucketsResult
| 🟠| ListMultipartUploads | ListMultipartUploadsOutput | ListMultipartUploadsResult
| 🟠| ListObjects | ListObjectsOutput | ListBucketResult
| 🟠| ListObjectsV2 | ListObjectsV2Output | ListBucketResult
| ❌| ListObjectVersions | ListObjectVersionsOutput | ListVersionsResult
| ❌| ListParts | ListPartsOutput | ListPartsResult
| - | SelectObjectContent | SelectObjectContentOutput | Payload
| ❌| UploadPartCopy | UploadPartCopyOutput | CopyPartResult

Note: I've removed `GetBucketLocation` as it's a special case with no root node wrapper, and `SelectObjectContent` as it's a streaming response.

This PR implements a fix by directly renaming the root node while still taking advantages of the specs. Previously, modifying the specs would also have effects on the parser where, for example, the shape would already exists (see `GetBucketTagging` and `GetObjectTagging`, both returns a root node with `Tagging` name but the shape is different). Now we can still differentiate for each operations. 

Now, the issue is systematic testing of those operations. I'm confident the new root node names are the right ones, but it would be nice to have replicable tests against AWS. We already have a test in place (`tests.integration.s3.test_s3.TestS3.test_response_structure`) testing some operations (and we can see that removing the patches didn't make the test fail as the new serializer part is now taking care of it).
However, as 41 of 44 operations needs to be tested for their XML response, this makes it a bit difficult as you would need to setup the resources to be able to query them afterwards (ex. `GetBucketCorsConfiguration` would need a configuration to be able to be queried). 

would fix #8030

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -1372,6 +1372,32 @@ class DeleteObjectTaggingRequest(ServiceRequest):
     ExpectedBucketOwner: Optional[AccountId]
 
 
+class Error(TypedDict, total=False):
+    Key: Optional[ObjectKey]
+    VersionId: Optional[ObjectVersionId]
+    Code: Optional[Code]
+    Message: Optional[Message]
+
+
+Errors = List[Error]
+
+
+class DeletedObject(TypedDict, total=False):
+    Key: Optional[ObjectKey]
+    VersionId: Optional[ObjectVersionId]
+    DeleteMarker: Optional[DeleteMarker]
+    DeleteMarkerVersionId: Optional[DeleteMarkerVersionId]
+
+
+DeletedObjects = List[DeletedObject]
+
+
+class DeleteObjectsOutput(TypedDict, total=False):
+    Deleted: Optional[DeletedObjects]
+    RequestCharged: Optional[RequestCharged]
+    Errors: Optional[Errors]
+
+
 class DeleteObjectsRequest(ServiceRequest):
     Bucket: BucketName
     Delete: Delete
@@ -1385,16 +1411,6 @@ class DeleteObjectsRequest(ServiceRequest):
 class DeletePublicAccessBlockRequest(ServiceRequest):
     Bucket: BucketName
     ExpectedBucketOwner: Optional[AccountId]
-
-
-class DeletedObject(TypedDict, total=False):
-    Key: Optional[ObjectKey]
-    VersionId: Optional[ObjectVersionId]
-    DeleteMarker: Optional[DeleteMarker]
-    DeleteMarkerVersionId: Optional[DeleteMarkerVersionId]
-
-
-DeletedObjects = List[DeletedObject]
 
 
 class ReplicationTimeValue(TypedDict, total=False):
@@ -1438,18 +1454,8 @@ class EndEvent(TypedDict, total=False):
     pass
 
 
-class Error(TypedDict, total=False):
-    Key: Optional[ObjectKey]
-    VersionId: Optional[ObjectVersionId]
-    Code: Optional[Code]
-    Message: Optional[Message]
-
-
 class ErrorDocument(TypedDict, total=False):
     Key: ObjectKey
-
-
-Errors = List[Error]
 
 
 class EventBridgeConfiguration(TypedDict, total=False):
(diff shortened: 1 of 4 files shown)


thrau (approved):
I think this is a much better solution than patching all this in the spec :+1:, thanks for digging in to this. And also I like that we scraped this from the docs, so should be pretty confident now that it's correct.

maybe we could add a couple of regression unit tests to the serializer to make sure the XML is serialized as expected. unless we have some already :-)

bentsku:
I'll look for the unit tests for the serializer! We have one integration test testing previously patched operations: 
`ListBuckets`, `ListMultipartUploads`, `ListObjects`, `ListObjectsV2`, `GetBucketTagging` and `GetObjectTagging`. If the serializer suddenly stops doing fixing the root node tag, this test will fail.

We could have a test using the specs + the dict from the serializer with the right node names and check if the serializer is returning the right name for the operation? I'm not sure how I'd construct this unit test, I'll have a look tomorrow!

alexrashed (approved):
Great investigation and thanks for jumping on this! The fix is looking great. 🚀 🦸🏽 
I wonder how other clients (like the Java SDK) handle this in a generalized way (but we can take a closer look in the next few weeks).

PR #8040, opened 2023-04-01
Fix AttributeError for custom S3 handler

giograno:
This PR fixes the `AttributeError: 'NoneType' object has no attribute 'service_name'` error occurring when doing any pod operation after the initialization of the S3 provider.

I am simply adding a guard clause for `context.service` being None, similar to the other S3 handler for CORS (`s3_cors_request_handler`).

The error can be simply reproduced with:
```
awslocal s3 mb s3://localstack-bucket
localstack pod save file://my-pod
```


bentsku (approved):
LGTM! Sorry I didn't change it like I did for `s3_cors_request_handler` 😬 One day we will have service specific request handler too, like the responses one! 🎉
Thanks a lot for fixing this!

Issue #8035, opened 2023-04-01
bug: ASGI Framework Lifespan error when running Localstack Docker instance


byteshiva:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

While running a Localstack Docker instance, the following warning message is displayed:

```
2023-04-01T01:34:09.207  WARN --- [-functhread5] hypercorn.error            : ASGI Framework Lifespan error, continuing without Lifespan support
2023-04-01T01:34:09.207  WARN --- [-functhread5] hypercorn.error            : ASGI Framework Lifespan error, continuing without Lifespan support
```

This warning indicates that there is an issue with the ASGI Framework Lifespan, and that the instance will continue running without Lifespan support. While the instance appears to be running without issue, it would be helpful to have more information on why this warning is being displayed and if it could potentially cause any problems in the future.

### Expected Behavior

Expected behavior:

- The localstack instance should run without any warnings or errors, and the portal should be accessible at https://localhost:4566/ or https://127.0.0.1:4566/

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

docker run --rm -it -p 4566:4566 -p 4510-4559:4510-4559 localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS:


Linux mx-cloud 6.0.0-6mx-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.0.12-1~mx21+1 (2022-12-15) x86_64 GNU/Linux

- LocalStack:

LocalStack version: 2.0.1.dev
LocalStack build date: 2023-03-31
LocalStack build git hash: d1d118ab
```

simonrw:
Hi, thanks for reporting this.

The warnings should not be affecting the normal startup of LocalStack. If LocalStack is not starting for you, would you mind posting the entire log output?

The warning is unfortunate, but should have no effect on the startup of LocalStack. It's talking about a [part of the ASGI specification](https://asgi.readthedocs.io/en/latest/specs/lifespan.html) that we don't need.

byteshiva:
> Hi, thanks for reporting this.
> 
> The warnings should not be affecting the normal startup of LocalStack. If LocalStack is not starting for you, would you mind posting the entire log output?


logs

```
./localstack start 

[...]

PR #8042, opened 2023-04-02
add support for asgi lifespan protocol listener in asgi/wsgi bridge

thrau:
This PR adds simple support for the ASGI lifespan protocol for the ASGIAdapter. The change allows you to attach a listener to the ASGIAdapter that can react to lifespan protocol events. See https://asgi.readthedocs.io/en/latest/specs/lifespan.html

This is mostly to get rid of the warning messages that have bothered and mislead some users in issues like #8035

But in the future, we could potentially use this for improving startup/shutdown routines.

The failure case is a bit tedious to test with the fixtures we have, so I decided to skip it for now, since we don't really have a use case for it yet.

### DIFF
--- a/localstack/aws/serving/asgi.py
+++ b/localstack/aws/serving/asgi.py
@@ -1,10 +1,14 @@
+import asyncio
 import concurrent.futures.thread
 from asyncio import AbstractEventLoop
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from localstack.aws.gateway import Gateway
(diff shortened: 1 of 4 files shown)


simonrw (approved):
Looks good apart from my comment, but hopefully my commit should suffice :tada:

thrau:
> Looks good apart from my comment, but hopefully my commit should suffice tada

thanks @simonrw, great catch, totally missed that! i simplified the implementation in the `ASGIGateway` to delegate the call entirely to the underlying adapter

PR #8050, opened 2023-04-03
Configure CI on release branches

simonrw:
We have a release branch `release/v2.1` which was not being tested.

Companion PR in ext: https://github.com/localstack/localstack-ext/pull/1573


alexrashed (approved):
Awesome! Thanks for jumping on this! 🦸🏽

PR #8052, opened 2023-04-03
include link to service coverage page on not implemented error

lukqw:
This PR adjusts the links that are printed for `not implemented` errors. 

Instead of pointing to our general feature coverage https://docs.localstack.cloud/user-guide/aws/feature-coverage/ they now point to the specific service page of our docs, e.g. https://docs.localstack.cloud/references/coverage/coverage_sqs/ for `sqs`.

question: 
do all our services have a related page on the docs?
based on the fact that the docs are auto-generated, I'm assuming that the answer to the question is yes, hence allowing us to link directly to the coverage page of each service 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@
   <br/>
   <a href="https://docs.localstack.cloud" target="_blank">📖 Docs</a> •
   <a href="https://app.localstack.cloud" target="_blank">💻 Pro version</a> •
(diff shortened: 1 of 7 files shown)


steffyP (approved):
LGTM! 🎉 👍 thanks so much for tackling this 🙏

PR #8056, opened 2023-04-03
Move test client creation logic into util package for re-use across repos

dominikschubert:
Since PyCharm still can't properly detect the imported pytest plugins from localstack, we'll just duplicate the actual fixture definition across the two repos.

The core logic behind the client creation has been moved to `util.py` and is re-used across these client fixture definitions.


dfangl (approved):
Great! Looking forward to have proper type hinting in -ext :D

Issue #8053, opened 2023-04-03
bug: custom messages of NotImplementedErrors are not displayed


lukqw:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating glue schema via the UI with data format `JSON` prints:
```
2023-03-31T11:34:40.994  INFO --- [   asgi_gw_0] localstack.aws.skeleton    : API action 'CheckSchemaVersionValidity' for service 'glue' not yet implemented or pro feature - check https://docs.localstack.cloud/user-guide/aws/feature-coverage for further information
2023-03-31T11:34:40.995  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS glue.CheckSchemaVersionValidity => 501 (InternalFailure)
```
even though I am starting LocalStack with an API Key and the operation is supported

### Expected Behavior

Since LocalStack does not yet support the JSON format for the glue schema creation operation, it should print the following:

```
Data formats other than AVRO are currently not supported.
```

due to following code in the glue provider:

```
if data_format != "AVRO":
     raise NotImplementedError("Data formats other than AVRO are currently not supported.")
```

Yet it does not, as our `NotImplementedError` handling does not print the message of the original error, and rather prints logs directing to our coverage docs page.
This behaviour is not specific to glue, but rather all `NotImplementedError`s with custom messages.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    LOCALSTACK_API_KEY=xxx localstack start

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    go to https://app.localstack.cloud/resources/glue/schemas/new
    enter name for schema
    pick `JSON` as data format
    write anything in the Schema Definition
    click submit

### Environment

```markdown
- OS: macOS
- LocalStack: 2.0
```

PR #8057, opened 2023-04-03
Display custom error message for NotImplementedErrors

giograno:
Implements #8053

### DIFF
--- a/tests/unit/aws/test_skeleton.py
+++ b/tests/unit/aws/test_skeleton.py
@@ -1,6 +1,10 @@
 import sys
 from typing import Dict, List
 
+import pytest
+
+from localstack.aws.api.sqs import SendMessageRequest
+
 if sys.version_info >= (3, 8):
     from typing import TypedDict
 else:
@@ -127,6 +131,19 @@ class TestSqsApiNotImplemented:
         raise NotImplementedError
 
 
+class TestSqsApiNotImplementedWithMessage:
+    service = "sqs"
+    version = "2012-11-05"
+
+    @handler("SendMessage", expand=False)
+    def send_message(
+        self,
+        context: RequestContext,
+        request: SendMessageRequest,
+    ) -> SendMessageResult:
+        raise NotImplementedError("We will implement it soon, that's a promise!")
+
+
 """ Test implementations """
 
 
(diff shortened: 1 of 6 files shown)


lukqw (commented):
When I tried to look into this issue I noticed that the `... yet implemented or pro feature ...` message is also printed in `localstack/aws/handlers/service.py` and `localstack/utils/aws/request_context.py`. 
Do we need to adjust the logic in there as well?

steffyP:
> When I tried to look into this issue I noticed that the `... yet implemented or pro feature ...` message is also printed in `localstack/aws/handlers/service.py` and `localstack/utils/aws/request_context.py`.
> Do we need to adjust the logic in there as well?

I think it would make sense. [localstack/utils/aws/request_context.py](https://github.com/localstack/localstack/blob/master/localstack/utils/aws/request_context.py#L175) seems to patch moto responses -> moto sometimes adds additional information when a specific parameters or actions are not implemented. 

Not quite sure when we would hit [localstack/aws/handlers/service.py](https://github.com/localstack/localstack/blob/master/localstack/aws/handlers/service.py#L185), I guess there are other edge-cases.

thrau:
Hey folks! The code that's being modified here is code that's actually not used in the handler chain. The handler chain has its own dispatching mechanism and also its own exception serializer, which happens here:

https://github.com/localstack/localstack/blob/e1b80232a5aa08f5ccc2b3f7c0950297bbe498f1/localstack/aws/handlers/service.py#L180-L193

Sorry about the confusion, I can see this may needs some consolidation

tests/unit/aws/test_skeleton.py
```diff
@@ -186,9 +203,23 @@ def test_skeleton_e2e_sqs_send_message():
...
 
 
-def test_skeleton_e2e_sqs_send_message_not_implemented():
+@pytest.mark.parametrize(
+    "api_class, oracle_message",
+    [
+        (
+            TestSqsApiNotImplemented(),
+            "API action 'SendMessage' for service 'sqs' not yet implemented or pro feature"
+            " - check https://docs.localstack.cloud/user-guide/aws/feature-coverage for further information",
```
giograno, line 212: This string should change accordingly with #8052.

steffyP (approved):
LGTM 👍 thanks for implementing 🙏

lukqw:
@giograno I have merged latest master into this branch and adjusted the files accordingly 👍 
Feel free to merge as-is or make further adjustments to your liking.

PR #8064, opened 2023-04-04
add boto3-stubs to development deps

dominikschubert:
Adds the types we're using for our typed client factory as dev dependencies

Packages have been extracted from this file here:
https://github.com/localstack/localstack/blob/2e86662f01a450396027982aadceac0006f89bd6/localstack/utils/aws/client_types.py#L4-L10

They're now also installed when running `make install` but I'm also open to move it to a separate target instead :thinking: 

/cc @pinzon


thrau (approved):
agree with daniel. this looks good to me!

dfangl (approved):

dominikschubert:
not sure why this wasn't closed automatically :raised_eyebrow:

PR #7240, opened 2022-11-24
Add boto client creation via factory pattern

dfangl:
**Please refer to the contribution guidelines in the README when submitting PRs.**
TBD

PR #8066, opened 2023-04-04
Disable HTTPS(S) proxies for internal calls

joe4dev:
Enterprise environments often use HTTP(S) proxies but LocalStack-internal AWS clients fail since boto3 migrated from requests to urllib3 and dropped support for `NO_PROXY` (https://github.com/boto/boto3/issues/3179).

This PR disables proxies for LocalStack-internal AWS calls and migrates Lambda (new provider) to the new AWS client (introduced in https://github.com/localstack/localstack/pull/7240).

## Issue

The following minimal configuration now fails against LocalStack:

```
HTTP_PROXY=http://mycompany.us-east-1.mycorp.com:3587
NO_PROXY = .s3.localhost.localstack.cloud,127.0.0.1,*.localhost
```

For example when trying to create any Lambda function via `awslocal lambda create-function ...`, an exception occurs:
**Hint:** For testing the setup, use [mitmproxy](https://mitmproxy.org/) and `HTTP_PROXY=http://localhost:8080`

```
botocore.exceptions.ProxyConnectionError: Failed to connect to proxy URL: "http://mycompany.us-east-1.mycorp.com:3587"
```

* We recommend `NO_PROXY` in our FAQ to fix connectivity issues in the BigData image: https://docs.localstack.cloud/getting-started/faq/#how-do-i-resolve-connection-issues-with-proxy-blocking-access-to-localstacks-bigdata-image

The issue came up in LocalStack Pro support.

## Explored Solutions

`NO_PROXY` supports a list of hosts to selectively disable any proxy configured through HTTP_PROXY
Example:
`NO_PROXY = .s3.localhost.localstack.cloud,127.0.0.1,*.localhost`

However,  we can only disable proxies entirely through botocore configuration: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html

AFIAK, the underlying urllib3 also does not offer a configuration to disable selected hosts from active proxies: https://urllib3.readthedocs.io/en/stable/advanced-usage.html

## Implemented Solution

Disable all proxies for LocalStack-internal calls because proxying them never makes sense for customers.
If we want to use mitmproxy for debugging LocalStack-internal AWS calls, we would need to add `"http": "http://localhost:4566"` to the proxies config. We could make this configurable if there is high demand among LocalStack developers.

## Workaround

I tested the `NO_PROXY` configuration from our [FAQ](https://docs.localstack.cloud/getting-started/faq/#how-do-i-resolve-connection-issues-with-proxy-blocking-access-to-localstacks-bigdata-image) documentation again and discovered that even a more generic version (without the s3 subdomain) `NO_PROXY=localhost.localstack.cloud,127.0.0.1,*.localhost` works against both, the old stack and new AWS client without this fix 🤔 

I don't know exactly why because it got removed from [boto3](https://github.com/boto/boto3/issues/3179#issuecomment-1069709605) and is not implemented in [urllib3](https://github.com/search?q=repo%3Aurllib3%2Furllib3+no_proxy&type=code) used by boto3.

### DIFF
--- a/localstack/services/awslambda/invocation/executor_endpoint.py
+++ b/localstack/services/awslambda/invocation/executor_endpoint.py
@@ -124,7 +124,9 @@ class ExecutorEndpoint:
         if not self.container_address:
             raise ValueError("Container address not set, but got an invoke.")
         invocation_url = f"http://{self.container_address}:{self.container_port}/invoke"
-        response = requests.post(url=invocation_url, json=payload)
+        # disable proxies for internal requests
+        proxies = {"http": "", "https": ""}
+        response = requests.post(url=invocation_url, json=payload, proxies=proxies)
         if not response.ok:
             raise InvokeSendError(
                 f"Error while sending invocation {payload} to {invocation_url}. Error Code: {response.status_code}"
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -647,7 +647,7 @@ FORCE_SHUTDOWN = is_env_not_false("FORCE_SHUTDOWN")
 MOCK_UNIMPLEMENTED = is_env_true("MOCK_UNIMPLEMENTED")
 
 # set variables no_proxy, i.e., run internal service calls directly
-no_proxy = ",".join([LOCALSTACK_HOSTNAME, LOCALHOST, LOCALHOST_IP, "[::1]"])
+no_proxy = ",".join([constants.LOCALHOST_HOSTNAME, LOCALHOST, LOCALHOST_IP, "[::1]"])
 if os.environ.get("no_proxy"):
     os.environ["no_proxy"] += "," + no_proxy
 elif os.environ.get("NO_PROXY"):
--- a/localstack/services/awslambda/invocation/lambda_service.py
+++ b/localstack/services/awslambda/invocation/lambda_service.py
@@ -22,6 +22,7 @@ from localstack.aws.api.lambda_ import (
     ResourceNotFoundException,
     State,
 )
+from localstack.aws.connect import connect_to
 from localstack.services.awslambda import api_utils
 from localstack.services.awslambda.api_utils import (
     lambda_arn,
@@ -45,7 +46,6 @@ from localstack.services.awslambda.invocation.models import lambda_stores
 from localstack.services.awslambda.invocation.version_manager import LambdaVersionManager
 from localstack.services.awslambda.lambda_utils import HINT_LOG
 from localstack.utils.archives import get_unzipped_size, is_zip_file
-from localstack.utils.aws import aws_stack
 from localstack.utils.container_utils.container_client import ContainerException
 from localstack.utils.docker_utils import DOCKER_CLIENT as CONTAINER_CLIENT
 from localstack.utils.strings import to_str
@@ -536,7 +536,7 @@ def store_lambda_archive(
             Type="User",
         )
     # store all buckets in us-east-1 for now
-    s3_client: "S3Client" = aws_stack.connect_to_service("s3", region_name="us-east-1")
+    s3_client: "S3Client" = connect_to(region_name="us-east-1").s3
     bucket_name = f"awslambda-{region_name}-tasks"
     # s3 create bucket is idempotent
     s3_client.create_bucket(Bucket=bucket_name)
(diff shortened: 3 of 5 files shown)


localstack/services/awslambda/invocation/executor_endpoint.py
```diff
@@ -124,7 +124,9 @@ def invoke(self, payload: Dict[str, str]) -> None:
         if not self.container_address:
             raise ValueError("Container address not set, but got an invoke.")
         invocation_url = f"http://{self.container_address}:{self.container_port}/invoke"
-        response = requests.post(url=invocation_url, json=payload)
+        # disable proxies for internal requests
+        proxies = {"http": "", "https": ""}
+        response = requests.post(url=invocation_url, json=payload, proxies=proxies)
```
dominikschubert, line 129: We should in the future maybe write a small wrapper and make sure we're using the wrapper for internal requests. :thinking:

dominikschubert (approved):
LGTM

Also, thanks for also migrating a few of the internal lambda clients to the new client :+1:

localstack/aws/connect.py
```diff
@@ -319,6 +320,10 @@ def _get_client(
         :param config: Boto config for advanced use.
         :return: Boto3 client.
         """
+        # Disable proxies for LocalStack-internal calls because boto3 does not support NO_PROXY anymore and
+        # proxying internal calls never makes sense for customers.
+        config = config.merge(botocore.client.Config(proxies={}))
+
```
dfangl, line 326: It makes sense here to avoid cache misses but this is now also enabled for external + clients against aws.
We might need to find another solution if someone needs proxying for those, at some point but this is fine for now!

dfangl (approved):
LGTM! Thanks for tackling this!

joe4dev:
Thanks for pushing this over the line @simonrw 

* ✅ Disabling proxying for internal Lambda invocation was the missing part 👍 
* 🤔 Regarding internal invocations in general, the main problem was that `s3.localhost.localstack.cloud` was missing in the `no_proxy` config. Obviously, disabling proxies completely also works and was my initial attempt because I read/thought that boto3 does not support no_proxy anymore.

localstack/aws/connect.py
joe4dev: Mhmm, that basically breaks shared Lambda layer fetching in a proxy setup 😒 


What if we just patch the `no_proxy` config in `localstack/config.py` for now?

```
no_proxy = ",".join([LOCALHOST_HOSTNAME, LOCALHOST, LOCALHOST_IP, "[::1]"])
```

`LOCALSTACK_HOSTNAME` resolves to `localhost.localstack.cloud` and that seemed to be the missing part in this issue.

I still don't fully understand why `NO_PROXY` works because I thought it is not supported anymore but my tests showed that actually works 🤔 😕

joe4dev: I searched through the site-packages and these are the libs that read `NO_PROXY`:

```
cd localstack/.venv/lib/python3.10/site-packages
ag NO_PROXY
boto/connection.py
697:        self.no_proxy = os.environ.get('no_proxy', '') or os.environ.get('NO_PROXY', '')

websocket/_url.py
75:DEFAULT_NO_PROXY_HOST = ["localhost", "127.0.0.1"]
106:        v = os.environ.get("no_proxy", os.environ.get("NO_PROXY", "")).replace(" ", "")
110:        no_proxy = DEFAULT_NO_PROXY_HOST

docker/utils/proxy.py
53:            env['no_proxy'] = env['NO_PROXY'] = self.no_proxy

requests/sessions.py
306:        NO_PROXY, we strip the proxy configuration. Otherwise, we set missing

requests/utils.py
860:    such a NO_PROXY to strip proxy configurations.

pip/_vendor/requests/sessions.py
306:        NO_PROXY, we strip the proxy configuration. Otherwise, we set missing

pip/_vendor/requests/utils.py
860:    such a NO_PROXY to strip proxy configurations.

localstack/config.py
508:elif os.environ.get("NO_PROXY"):
509:    os.environ["NO_PROXY"] += "," + no_proxy

future/backports/urllib/request.py
2423:    no_proxy = os.environ.get('no_proxy', '') or os.environ.get('NO_PROXY', '')
```

simonrw: I think it's in `botocore.utils`: https://github.com/boto/botocore/blob/63d5d3503f1b2267a10433e42727395729b039d0/botocore/utils.py#L2952

PR #8068, opened 2023-04-04
fix APIGW RestAPI minimumCompressionSize

bentsku:
As reported in https://github.com/localstack/localstack/issues/7975#issuecomment-1495698212, we have an issue while returning a `minimumCompressionSize` when set to an empty string. The docs states:
> To enable compression, apply a replace operation with the accompanying value property set to a non-negative integer between 0 and 10485760. To disable compression, apply a replace operation with the value property set to null or omit the value property.

https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html#UpdateRestApi-Patch

Apparently, an empty string works as well. I've added the validation + type casting of either `int` or `None` when disabled, but we're not actually enforcing this as a feature, only mocking this value, so no changes down the line.

If we updated the RestAPI with an empty string, we would return it in the response and the client would fail parse that response. This is now fixed. 

_fix #7975_


whummer (approved):
Great catch, and awesome to see additional snapshots to cover this case! 🚀

PR #8071, opened 2023-04-04
Detect containerd containers

simonrw:
We currently don't detect containers running in containerd correctly. containerd does not set the same docker file or cgroup values as docker does/did.

This was triggered by a user issue: https://github.com/localstack/localstack/issues/8035

I was able to recreate this by running containerd on Linux, or adding a beta feature of Docker Desktop on macOS.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -291,6 +291,13 @@ def in_docker():
         os_hostname = socket.gethostname()
         if os_hostname and os_hostname in content:
             return True
+
+    # containerd does not set any specific file or config, but does use
+    # io.containerd.snapshotter.v1.overlayfs as the overlay filesystem
+    with open("/proc/1/mountinfo", "rt") as infile:
+        if "io.containerd" in infile.read():
+            return True
+
     return False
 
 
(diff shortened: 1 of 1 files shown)


thrau (commented):
looks like you're building on #8042 which was merged into `release/v2.1`

just the `config.py` change could go into master.

on a sidenote: we create these marker files in the Dockerfile which it seems to be could be the first thing we check, before we go into systems details.

simonrw:
> looks like you're building on #8042 which was merged into `release/v2.1`
> 
> just the `config.py` change could go into master.



Ugh my bad, I'll rebase. I'm still learning the github CLI! I think the changes I'm after _just_ refer to `config.py`. Thanks


> on a sidenote: we create these marker files in the Dockerfile which it seems to be could be the first thing we check, before we go into systems details.

Are you referring to `/usr/lib/localstack/.{community,pro}-version` and `/tmp/localstack/.marker`? I think that makes sense. It's pretty unlikely that a user would set these, and if we - for some reason - remove these files from the docker image, we can fall back to checking the environment.

dfangl (approved):
LGTM!

PR #8076, opened 2023-04-05
fix DockerHub description update action

alexrashed:
This PR is used to debug current issues with the "Update Docker Hub Description" action, which is currently failling.

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -7,6 +7,12 @@ on:
     paths:
       - README.md
       - .github/workflows/dockerhub-description.yml
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


alexrashed:
Turns out, that the issue is actually called by the limit for the short description:
image

This description is set to the repository's description here on GitHub: https://github.com/localstack/localstack/blob/48fe3e85040e27675bbae7fa60c850919afbd384/.github/workflows/dockerhub-description.yml#L30

Unfortunately, it seems that there is a limit of 100 bytes for the short description when being set via the API.
A detailed description and a possible explanation can be found here: https://github.com/peter-evans/dockerhub-description/issues/142

For now, we stripped down the description (by removing an unnecessary whitespace and the exclamation mark in the end).
/cc @HarshCasper @thrau

PR #8080, opened 2023-04-05
Replace (incorrect) regex with filter pattern

simonrw:
I had thought from the `v[0-9]+` entry above that this filter pattern is
a regex, but it isn't. It's annoyingly close though. The effect is that
CI was _not_ being run against the `release/v2.1` branch.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/cli-tests.yml
+++ b/.github/workflows/cli-tests.yml
@@ -15,7 +15,7 @@ on:
       - "bin/**"
     branches:
       - master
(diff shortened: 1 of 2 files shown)


steffyP (approved):
Thanks for jumping on this again @simonrw 🙏 
changes LGTM! 👍

PR #8084, opened 2023-04-06
fix returned security options for opensearch domain

dominikschubert:
Fixes a user-reported issue where the returned security options for the `describe_domain` call were always set to `Enabled=False`.


alexrashed:
@dominikschubert Do you know if this is still an issue? Is this something that should be picked up and fixed?

dominikschubert:
> @dominikschubert Do you know if this is still an issue? Is this something that should be picked up and fixed?

Honestly I don't know and it's hard to tell. Since this hasn't turned up again, I'd say  let's prefer to close this PR. There's not much lost here even if it surfaces again.

PR #8085, opened 2023-04-06
add Docker SDK default timeout config

alexrashed:
We have recently seen flaky tests (on a new test environment which uses private GitHub runners). These tests failed due to request timeout errors within the Docker SDK client.

Some research led to an issue report in `docker-py`: https://github.com/docker/docker-py/issues/2266
According to these reports, these issues might be associated with the Docker daemon on the machine, and that it sometimes can happen that spawning a container takes a long time (i.e. longer than the default timeout of 60 seconds).

Since the Docker client is quite essential, and maybe some users are also facing issues like this in high-load scenarios, this PR introduces a config variable which allows globally setting the default timeout (in seconds) for the Docker SDK client.


dfangl (approved):
Seems like a good addition, LGTM!

PR #8086, opened 2023-04-06
fix: KMS incorrect error messaging for invalid keys 

sannya-singal:
- fixed error messaging for encrypt, decrypt, sign, verify signatures, generate and verify mac, etc. 
- added test case for parity between AWS and LS to verify error message when invalid key id is passed to operations


localstack/services/kms/provider.py
```diff
@@ -454,7 +454,7 @@ def get_public_key(
 
     def _generate_data_key_pair(self, key_id: str, key_pair_spec: str, context: RequestContext):
         key = self._get_key(context, key_id)
-        self._validate_key_for_encryption_decryption(key)
+        self._validate_key_for_encryption_decryption(key, "GenerateDataKeyPair")
```
viren-nadkarni, line 457: Avoid passing the operation as strings. 

You get get the operation from `context.operation`, so pass the context to the validation function.

```py
self._validate_key_for_encryption_decryption(context, key)
```

viren-nadkarni (changes requested):

viren-nadkarni (approved):

Issue #8090, opened 2023-04-06
bug: 'pictures' a reserved word for an S3 bucket name?


kyllerss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I created an S3 bucket with the name `pictures`. Each time I tried to write to it, I would get a CORS error. 

I used the same code as I used for another bucket which was working fine (`originaluploads`). I even verified that the same code path that was causing issues under `pictures` worked fine when pointed against the `originaluploads` bucket. 

When I renamed the `pictures` bucket to `userpictures`, the code worked fine. 

Is `pictures` a reserved bucket name in LocalStack? If so, I couldn't find any documentation. Otherwise, seems like a bug. 

### Expected Behavior

`pictures`-named S3 buckets shouldn't reject with CORS errors.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Attempt to write to an S3 bucket named `pictures`.

### Environment

```markdown
- OS: Docker
- LocalStack: 2.0.0
```

bentsku:
Hello @kyllerss and thanks for your report! I'm going to start looking into it, but would you have a small reproducible sample or something I could follow to reproduce?

Do you have any CORS configuration on your bucket? Or do you use CORS settings of LocalStack (https://docs.localstack.cloud/references/configuration/#security)?
How are you writing to your bucket? Using a SDK from the browser?

I'll try to reproduce in the meantime, but it would be nice to be sure we're reproducing the same issue. Thanks!

bentsku:
Thanks a lot, this should be enough to look into it, thanks a lot for taking the time to share the data. I'm going to look into it! I'm not aware of a `pictures` specific thing on CORS issue on LocalStack side. This seems interesting! I'll come back once I have more informations, thanks again for the report!

bentsku:
I've tried to reproduce, and it doesn't seem there's an issue with the `pictures` bucket name and CORS. We will need the LocalStack logs when the issue happen, to know exactly what is happening on LocalStack side. It could also be that there's an unexpected error and we wouldn't populate CORS headers correctly after that, but it would be hidden by the browser because of CORS: often CORS is the first to fail and hide the real issue underneath.

You can get more verbose logs by setting `LS_LOG=trace` and `DEBUG=1` before starting LocalStack.

Sorry for asking more, I'd understand if that takes time! I've also tried to look and there isn't a trace of `pictures` hardcoded anywhere that would be linked to that specific issue.

[...]

PR #8091, opened 2023-04-06
fix wrong service mapping to s3 vhost bucket

bentsku:
This PR fixes the issue reported in #8090.

### Issue

The flow is the following:
- request received as virtual host addressed : Host = `pictures.s3.localhost.localstack.cloud`
- the S3 CORS handler does its job correctly
- service name parser wrongly attributes this request to the service named `pi` because the host starts with `pi`
- the general CORS handler thinks it's not an S3 request and reject the request without taking into account S3 CORS config

Note that this issue could arise as well with any bucket name starting with an AWS service with an endpoint prefix, like `kmstest` or anything (you can play with the added test to see everything that would break). This did not affect functionality directly, as the request would be correctly routed to the right service in the end because it would be matched on the service router, but it greatly interfere with the CORS handler.

### Solution

I've added some checks in the service name parser, to verify that the request is not an S3 request when trying to match on the host. I'm not fond of the `".s3." not in host`, but the only other solution would be to forbid the creation of any S3 bucket with a conflictual name with any AWS service endpoint prefix (which is possible to do too).
I'm open to suggestion as to how to make this better, but it seems to fix the issue for now.

_fix #8090_

### DIFF
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -344,7 +344,8 @@ def determine_aws_service_name(request: Request, services: ServiceCatalog = None
     if host:
         # iterate over the service spec's endpoint prefix
         for prefix, services_per_prefix in services.endpoint_prefix_index.items():
-            if host.startswith(prefix):
+            # this prevents a virtual host addressed bucket to be wrongly recognized
+            if host.startswith(f"{prefix}.") and ".s3." not in host:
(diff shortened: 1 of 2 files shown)


PsuedoJack (commented):
Thanks we are Collaborative Team

thrau (approved):
What a find 👌 lgtm

PR #8095, opened 2023-04-07
enforce semver labels, block major labeled PRs

alexrashed:
- It renames some of the existing GitHub workflow files (in order to somewhat categorize them).
- It adds two more workflows:
  - `.github/workflows/pr-enforce-pr-labels.yml`
    - This workflow enforces that **every single PR** has to be labeled with exactly one of the following three labels: `semver: major`, `semver: minor`, `semver: patch`.
  - `.github/workflows/pr-enforce-bugfix-master.yml`
    - This workflow enforces that (just like now) only PRs which contain patches (non-breaking, no new features) can be merged.
    - This means that when this workflow is enabled, PRs which have been labeled either with `semver: minor` or `semver: major` are **rejected**.
    - If we decide that we open the master branch for features, we can disable this workflow (or rewrite it such that we block PRs labeled with `semver: major`).


alexrashed:
FYI: The "enforce bugfix on master" has been changed to "enforce no major on master". This action will stay active until the next major release (to avoid unwanted breaking changes being merged into the master).
I am still looking for any feedback! Please let me know if you feel this is not necessary, useless, should be changed, is nice,...

PR #8102, opened 2023-04-09
simplify implementation of HTTPRequestEventStreamAdapter

thrau:
This PR simplifies the implementation of the asgi/wsgi bridge fundamental `HTTPRequestEventStreamAdapter`.

The adapter exposes and underlying `ASGIReceiveCallable` as a synchronous IO stream which we later plug into the werkzeug request object. Previously we had separate implementations of `read`, `readline`, and `readlines`, and did not use any base abstraction. However, `RawIOBase` already provides generic implementations of these methods over `readinto`. So I removed all of the special implementations and replaced them with a generic `readinto` implementation.

The drawback is that the generic implementations are not optimized, and require a buffered reader to be used (for instance, `readline` will always call `readinto` with a buffer size of 1, which causes very poor performance).

### DIFF
--- a/localstack/http/asgi.py
+++ b/localstack/http/asgi.py
@@ -115,9 +115,10 @@ def create_wsgi_input(
 ) -> t.IO[bytes]:
     """
     Factory for exposing an ASGIReceiveCallable as an IO stream.
-    :param receive:
-    :param event_loop:
-    :return:
+
+    :param receive: the receive callable
+    :param event_loop: the event loop used by the event stream adapter
+    :return: a new IO stream that wraps the given receive callable.
     """
     return BufferedReader(RawHTTPRequestEventStreamAdapter(receive, event_loop))
 
--- a/localstack/http/asgi.py
+++ b/localstack/http/asgi.py
@@ -181,12 +181,10 @@ class RawHTTPRequestEventStreamAdapter(RawIOBase):
 
         # in this case, we can read at max buf_size from the body into the buffer, and need to save the
         # rest for the next call
-        to_read = min(remaining, buf_size)
-        buf[:to_read] = body[pos : pos + to_read]
+        buf[:buf_size] = body[pos : pos + buf_size]
+        self._buffered_body_pos = pos + buf_size
 
-        self._buffered_body_pos = pos + to_read
-
-        return to_read
+        return buf_size
 
 
 class WsgiStartResponse:
--- a/localstack/http/asgi.py
+++ b/localstack/http/asgi.py
@@ -6,7 +6,7 @@ import math
 import typing as t
 from asyncio import AbstractEventLoop
 from concurrent.futures import Executor
-from tempfile import SpooledTemporaryFile
+from io import BufferedReader, RawIOBase
 from urllib.parse import quote, unquote, urlparse
 
 if t.TYPE_CHECKING:
@@ -110,10 +110,24 @@ async def to_async_generator(
         yield val
 
 
-class HTTPRequestEventStreamAdapter:
+def create_wsgi_input(
+    receive: "ASGIReceiveCallable", event_loop: t.Optional[AbstractEventLoop] = None
+) -> t.IO[bytes]:
     """
-    An adapter to expose an ASGIReceiveCallable coroutine that returns HTTPRequestEvent
-    instances, as a PEP 3333 InputStream for consumption in synchronous WSGI/Werkzeug code.
+    Factory for exposing an ASGIReceiveCallable as an IO stream.
+    :param receive:
+    :param event_loop:
+    :return:
+    """
+    return BufferedReader(RawHTTPRequestEventStreamAdapter(receive, event_loop))
+
+
+class RawHTTPRequestEventStreamAdapter(RawIOBase):
+    """
+    An adapter to expose an ASGIReceiveCallable coroutine that returns HTTPRequestEvent instances as an IO
+    stream for synchronous WSGI/Werkzeug code. The adapter is a Raw IO stream, meaning it does not have
+    optimized ``read``, ``readline``, or ``readlines`` methods. Make sure to use a ``BufferedReader`` around
+    the stream adapter.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


localstack/http/asgi.py
```diff
@@ -123,111 +137,56 @@ def __init__(
...
-
-            # ensure the buffer has at least `size` bytes (or all)
-            if size > 0:
-                if len(buf) >= size:
-                    break
-            _, more = self._read_into(buf)
             if not more:
-                break
+                self._more_body = False
+                return 0
```
bentsku, line 165: question: should we return here? doesn't `more_body` indicates that there won't be more, but couldn't there be data in that last chunk? I think in the previous implementation, we would still extend the buffer with the body from this event.

bentsku (commented):
This brings back memories from this deep debugging session leading to a Werkzeug bug report!

I don't have any big objection to the changes (except maybe the question about `_more_body`), and the `BufferedReader` will take care of optimising the `RawIOBase` generic implementations. Could it actually be faster in the end, as the buffer will be a `memoryview`? Reading the [implementation](https://github.com/python/cpython/blob/main/Lib/_pyio.py#L1026) is pretty interesting. 

It really simplifies our implementation, and it feels nice to be able to use python built-ins.

I might have a stupid question: if I understood well, the flow is the following, starting from zero:
- we try to get data from the `ASGIReceiveCallable`
- we get data, but more than what the buffer (from the `BufferedReader` in that case) might be able to take
- we save the whole data internally, even what we have written
- next call, if we have data saved, we don't even try to receive from the coroutine but just return what we had saved, which might be 1 byte even if the buffer can take way more.
- bad cases: next call, we actually might hit the end of the coroutine, and will just return 0. Or we just had 10 bytes that we could have fit in the previous call.
- kinda good case, we have just under the size of the buffer, and doing anything above would not have helped. Or we still have way more than the buffer can take, and continue.

I understand that it might complicate the implementation a lot, or that it shouldn't be at all what we should do 😅
I'm also not sure what limits we could have to not take too much from the coroutine, because we might buffer way too much into the raw stream, but I'd try to get data if I have space in the buffer (which we can know if the `_buffered_body` is not `None`).
I also don't have an idea in what cases we fall most often. Do we often get a big body from the coroutine, and a small buffer from the `BufferedReader`, which would make sense to drain with multiple calls? Or are we more often with a big buffer able to eat the whole body?
The buffer into a buffer is playing games with me. 😅

Anyway, love that we're using more primitives like this! 😄

thrau:
> Could it actually be faster in the end, as the buffer will be a `memoryview`?

it's more memory efficient as we don't copy buffers around, so yes it could be a bit faster!

> we get data, but more than what the buffer (from the BufferedReader in that case) might be able to take

potentially we get either less and more data actually. there are generally no guarantees in IO streams, which is why `read` and `readinto` always return the number of bytes that were actually returned into the buffer. the consumer of a raw stream (in this case the `BufferedReader`, needs to be able to deal with the case that fewer bytes than what it expects can be returned. this is actually something we had to fix in werkzeug https://github.com/pallets/werkzeug/issues/2558

> next call, if we have data saved, we don't even try to receive from the coroutine but just return what we had saved, which might be 1 byte even if the buffer can take way more.


exactly. we could in principle also have a loop that calls `receive()` as often is needed to fill the buffer, but there is no point to that. it would just make the implementation more complicated.

> I also don't have an idea in what cases we fall most often. Do we often get a big body from the coroutine, and a small buffer from the BufferedReader, which would make sense to drain with multiple calls?

that's a great question, that will depend mostly on the service we are calling. it's hard to generalize, but in most cases, HTTP requests are small anyway, and buffer size is 8k, so we will probably just do two `receive` calls (one for the body, one to get the `more_body=False`), and allocate just one buffer


> I understand that it might complicate the implementation a lot, or that it shouldn't be at all what we should do sweat_smile

from your comment i didn't really catch whether you're suggesting anything in particular? the implementation looks pretty much the way you described it. :-) just with some clarifications

localstack/http/asgi.py
```diff
@@ -123,111 +137,56 @@ def __init__(
...
+            # less than the buffer size allows, but that's ok for raw IO streams.
+            buf[:remaining] = body[pos:]
+            self._buffered_body = None
+            return remaining
 
-            lines.append(line)
-            size = size - len(line)
+        # in this case, we can read at max buf_size from the body into the buffer, and need to save the
+        # rest for the next call
+        to_read = min(remaining, buf_size)
```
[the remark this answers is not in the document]
thrau: yea, let's say one receive call returns 100 bytes, and the buffer is 90 bytes, then in the first receive call `to_read` will be 90, the buffer will be filled so we get a cleared buffer in the next call, and there `to_read` will be 10 (because from the 100 bytes we already read 90)

localstack/http/asgi.py
thrau: empirically i validated that there is always a distinct event with a `b''` body and `more_body=False`. in other words, there will _always_ be at least two calls to `receive()`: one to get the body (which will have `more_body=True`, even though there's nothing else to read), and then one with a zero body and `more_body=False`. :shrug:

bentsku: Oh okay, yeah I tried to read the ASGI documentation and they were not explicit about the fact that there would be a body in the event with `more_body=False`. Good to know!

thrau: yep same. i'm not 100% confident it's in the spec, but at least hypercorn behaves in this way.

localstack/http/asgi.py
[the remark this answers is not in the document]
bentsku: Okay, but right above we have a condition with `if remaining <= buf_size:` which return. So, wouldn't that make `buf_size` always smaller?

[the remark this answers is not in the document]
thrau: oh right! totally missed that, great catch :100:

i restructured the code two or three times and may have just been a leftover :-) will fix

bentsku:
Thanks for the clarification! 😄

> potentially we get either less and more data actually. there are generally no guarantees in IO streams, which is why read and readinto always return the number of bytes that were actually returned into the buffer. the consumer of a raw stream (in this case the BufferedReader, needs to be able to deal with the case that fewer bytes than what it expects can be returned. this is actually something we had to fix in werkzeug https://github.com/pallets/werkzeug/issues/2558

Right, yes I remember returning less data than expected. 

> exactly. we could in principle also have a loop that calls receive() as often is needed to fill the buffer, but there is no point to that. it would just make the implementation more complicated.

Of course yes, I didn't have a loop in mind, more like a check in the beginning, but it's like... _premature optimization is the root of all evil_ right! My idea was to check if we had a buffered body and the rest to send was something like very small, but the buffer had a lot of space, should we try to get data from the coroutine instead of just returning the small rest of the body. But the `BufferedReader` will just call once more, I don't think it's very important nor efficient either. I was mostly wondering! 

Thanks a lot for taking the time to answer my questions! I'm really glad, this implementation looks really cool!

localstack/http/asgi.py
bentsku: perfect, well, that simplifies it and we don't need to check if there's a body then. Good for us 😄

bentsku (approved):

localstack/http/asgi.py
```diff
@@ -110,10 +110,24 @@ def _next_sync():
...
+def create_wsgi_input(
+    receive: "ASGIReceiveCallable", event_loop: t.Optional[AbstractEventLoop] = None
+) -> t.IO[bytes]:
     """
-    An adapter to expose an ASGIReceiveCallable coroutine that returns HTTPRequestEvent
-    instances, as a PEP 3333 InputStream for consumption in synchronous WSGI/Werkzeug code.
+    Factory for exposing an ASGIReceiveCallable as an IO stream.
+    :param receive:
+    :param event_loop:
+    :return:
```
baermat, line 120: nit: This looks unfinished, so either we need some information here or remove the section all together :)

baermat (approved):
Seems like the bot pinged me due to some completely unrelated changes, so I didn't dive into this comparatively complex topic. For completeness, here is my "review"^^

thrau:
for the record: some preliminary benchmarking with `hey` suggests that there is no noticeable performance impact at least for simple requests with small payloads.

Issue #8083, opened 2023-04-05
bug: KMS keys has no check on the plaintext size to encrypt


jcanabarro:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Localstack KMS allows plaintext data larger than the maximum size allowed by the key specs to be encrypted, which is different from AWS KMS. This results in inconsistent behavior between Localstack and AWS.

### Expected Behavior

The KMS feature in Localstack should check the size of plaintext based on the key specs to encrypt, just like AWS KMS. This ensures that the encrypted data is consistent between Localstack and AWS.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

```yaml
version: "3.8"

services:
  localstack:
    container_name: localstack_main
    image: localstack/localstack:1.4.0
    restart: always
    ports:
      - "127.0.0.1:53:53"
      - "127.0.0.1:53:53/udp"
      - "127.0.0.1:443:443"
      - "127.0.0.1:4510-4559:4510-4559"
      - "127.0.0.1:4566:4566"
    environment:
      - DEFAULT_REGION=eu-west-3
      - LAMBDA_EXECUTOR=docker
      - LAMBDA_REMOTE_DOCKER=true
      - LAMBDA_REMOVE_CONTAINERS=true
      - DEBUG=1
      - PERSISTENCE=1
      - DNS_ADDRESS=0
      - DOCKER_HOST=unix:///var/run/docker.sock

    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
      - ./scripts/localstack_bootstrap:/docker-entrypoint-initaws.d/
    env_file:
      - .env-localstack
```
#### Python script to reproduce 

```py
import boto3

def assume_role():
    sts_client = boto3.client(
        "sts",
        endpoint_url="http://localhost:4566",
    )
    assumed_role_object = sts_client.assume_role(
        RoleArn="arn:aws:iam::123456789101:role/fake-role",
        RoleSessionName="fake-role",
    )

    credentials = assumed_role_object["Credentials"]
    resource = boto3.client(
        "kms",
        aws_access_key_id=credentials["AccessKeyId"],
        aws_secret_access_key=credentials["SecretAccessKey"],
        aws_session_token=credentials["SessionToken"],
        region_name="eu-west-3",
        endpoint_url="http://localhost:4566",
    )

    return resource


[...]

PR #8103, opened 2023-04-09
fix: Adding validation on kms plaintext based on the key type 

jcanabarro:
The issue was partially fixed here #8083, but I also need to check the context based on the key type, not only the max allowed plaintext size.
* Added method to validate the plaintext size based on the key type, all those extracted numbers are from [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/encrypt.html)
* Add tests cases to the written function

### DIFF
--- a/localstack/services/kms/provider.py
+++ b/localstack/services/kms/provider.py
@@ -131,8 +131,6 @@ VALID_OPERATIONS = [
     "GenerateDataKeyPair",
     "GenerateDataKeyPairWithoutPlaintext",
 ]
-# maximum size allowed by AWS
-PLAINTEXT_MAX_SIZE_BYTES = 4096
 
 
 class ValidationError(CommonServiceException):
--- a/localstack/services/kms/provider.py
+++ b/localstack/services/kms/provider.py
@@ -1018,6 +1018,41 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
                     f" constraint: [Member must satisfy enum value set: {VALID_OPERATIONS}]"
                 )
 
+    def _validate_content_size(
+        self,
+        plaintext: PlaintextType,
+        key: KmsKey,
+        encryption_algorithm: EncryptionAlgorithmSpec = None,
+    ):
+        # max size values extracted from AWS boto3 documentation
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/encrypt.html
+        max_bytes_size = 0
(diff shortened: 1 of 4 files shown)


jcanabarro:

jcanabarro:
recheck

jcanabarro:
recheck

viren-nadkarni:
@jcanabarro Some of your earlier commits are associated with `joao@querylayer.com`. Could you squash/amend all you commits and use `joaovictorcana@gmail.com` for all the commits?

localstack/services/kms/provider.py
```diff
@@ -1018,6 +1019,33 @@ def _validate_grant_request(self, data: Dict, store: KmsStore):
...
+    def _validate_plaintext_key_type_based(
+        self,
+        plaintext: PlaintextType,
+        key: KmsKey,
+        encryption_algorithm: EncryptionAlgorithmSpec = None,
+    ):
+        # max size values extracted from AWS boto3 documentation
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/encrypt.html
+        max_size_bytes = 4096  # max allowed size
+        if key.metadata["KeySpec"] == "RSA_2048" and encryption_algorithm == "RSAES_OAEP_SHA_1":
```
sannya-singal, line 1031: We can use the key spec values from https://github.com/localstack/localstack/blob/ed7e02f18253e22780131cf961304aa9e9df5c85/localstack/aws/api/kms/__init__.py#L158 and for encryption algorithm https://github.com/localstack/localstack/blob/ed7e02f18253e22780131cf961304aa9e9df5c85/localstack/aws/api/kms/__init__.py#L122.

tests/integration/test_kms.py
```diff
@@ -542,6 +542,28 @@ def test_encrypt_decrypt(self, kms_create_key, key_spec, algo, aws_client):
         )["Plaintext"]
         assert base64.b64decode(plaintext) == message
 
+    @pytest.mark.parametrize(
```
sannya-singal, line 545: We can add `pytest.mark.aws_validated` handler if it tested against AWS.

sannya-singal (changes requested):
Thanks for adding this validation 🎉.

jcanabarro:
@sannya-singal I'll abandon the PR and recreate because even with the rebase I'm still facing issues with the CLA.

PR #7274, opened 2022-12-02
remove pins for werkzeug, flask, and quart

alexrashed:
This PR removes the pins for `werkzeug`, `flask`, and `Quart` introduced in #6514 and #6593.

TODO:
- [x] Remove pins.
- [ ] Fix usage of removed request context stack (for both flask and quart).
- [ ] Ensure all tests are green.

PR #6514, opened 2022-07-24
Pin quart and werkzeug versions to fix recent changes in req ctx stack, URL Map

whummer:
Pin `quart` version to `0.17` to fix recent change in request context stack.

The `_*_ctx_stack` globals have been removed in `0.18` - see this commit for rationale/background: https://github.com/pallets/quart/commit/b2b614b336d99071a05a8525d4d0e18b9a1d66ec

Seems easy enough to pin the version for now, but we should also keep an eye on developments in flask (looks like `_request_ctx_stack` will be [removed there soon](https://github.com/pallets/flask/blob/cbebdae698272eeea2739988ccce54b1f14a1a67/src/flask/__init__.py#L64-L68) as well).

---
Update: Also pinning werkzeug to version `2.1.2` for now, as the latest released version `2.2.0` seems to have changed the logic of URL Map routing - our routes with `strict_slashes=False` in the Lambda API are now failing when sending a request with a trailing slash in the path.

---

Also sneaking in a small change to remove duplicate constant `MAVEN_BASE_URL`.

PR #8104, opened 2023-04-10
upgrade werkzeug to 2.2.3

thrau:
## Motivation
This PR upgrades werkzeug to 2.2.3, and the dependents flask and quart, that were pinned in https://github.com/localstack/localstack/pull/6514

The changes since werkzeug 2.1.2 had the following effects:
* groups in regex host matchers must be non-capturing, otherwise they break. this was a problem in opensearch and s3, and also the `PortConverter` regex. things like `(.*).` in rules matchers need to be changed to `(?:.*).`
* `_request_ctx_stack` used to set the request context for flask and quart was removed
* `/<path:path>` matchers no longer remove trailing slashes. previously `/foo/` and `/foo` would both lead to the `path` being passed to be `foo`, now it is `foo/` and `foo` respectively.

## Changes
The PR does the following:
* adds several tests that cover the changed behavior
* fixes our request context handlers to the flask/quart refactoring (thanks to @alexrashed for his solution in #7274, mine was wrong)
* simplifies the opensearch endpoint matchers using the corrected `PortConverter`, to review by @alexrashed 
* reworked the `S3VirtualHostProxyHandler` a bit, and made some other minor s3 changes, to review by @bentsku 
  * changes the `AWS_REGION_REGEX` to use non-capturing groups
  * simplifies the regex to match domains and ports
  * inject the `Proxy` instance as dependency so we can unit test the implementation
* use `PortConverter` for s3-website regex
* `rstrip("/")` the path before it goes into the op router matcher so we get the old behavior
* tiny fix in s3 website to remove a redundant trailing slash

## Learnings for the next time slashes come back to haunt us

* `GreedyPathConverter` is needed to make `/mybucket//mykey` match the rule `/<Bucket>/<path:Key>`, such that `Key=/mykey`. 
* the line `path = path.rstrip("/")` in `RestServiceOperationRouter.match` is needed because without it, in combination with the `GreedyPathConverter`, `/mybucket/` would match the rule `/<Bucket>/<path:Key>` instead of `/<Bucket>`, which lead to calls like `GET /mybucket/` being detected as `GetObject` with `Key=''` instead of `ListObjects` (bucket operation)
* URI matching in smithy is explained in https://smithy.io/2.0/spec/http-bindings.html#uri, but there are no examples for the greedy labels for the special cases with trailing slashes or prefix slashes in patterns (like the first example `/mybucket//mykey`). however it states that "trailing slashes are ignored", which substantiates that the `rstrip` call in match is OK

### DIFF
--- a/localstack/aws/protocol/op_router.py
+++ b/localstack/aws/protocol/op_router.py
@@ -13,6 +13,11 @@ from localstack.http.request import get_raw_path
 
 
 class GreedyPathConverter(PathConverter):
+    """
+    This converter makes sure that the path ``/mybucket//mykey`` can be matched to the pattern
+    ``<Bucket>/<path:Key>`` and will result in `Key` being `/mykey`.
+    """
+
     regex = ".*?"
 
 
@@ -258,14 +263,13 @@ def _create_service_map(service: ServiceModel) -> Map:
 
     return Map(
         rules=rules,
-        # don't be strict about trailing slashes
+        # don't be strict about trailing slashes when matching
         strict_slashes=False,
         # we can't really use werkzeug's merge-slashes since it uses HTTP redirects to solve it
         merge_slashes=False,
+        # get service-specific converters
+        converters={"path": GreedyPathConverter},
     )
-
-
-multiple_slash_pattern = re.compile(r"/+")
 
 
 class RestServiceOperationRouter:
@@ -297,7 +301,14 @@ class RestServiceOperationRouter:
             # specified. the specs do _not_ contain any operations on OPTIONS methods at all.
             # avoid matching issues for preflight requests by matching against a similar GET request instead.
             method = request.method if request.method != "OPTIONS" else "GET"
-            path = self.get_path_for_matching(request)
+
+            path = get_raw_path(request)
+            # trailing slashes are ignored in smithy matching,
+            # see https://smithy.io/1.0/spec/core/http-traits.html#literal-character-sequences and this
+            # makes sure that, e.g., in s3, `GET /mybucket/` is not matched to `GetBucket` and not to
+            # `GetObject` and the associated rule.
+            path = path.rstrip("/")
+
             rule, args = matcher.match(path, method=method, return_rule=True)
         except MethodNotAllowed as e:
             # MethodNotAllowed (405) exception is raised if a path is matching, but the method does not.
@@ -318,24 +329,3 @@ class RestServiceOperationRouter:
         operation: OperationModel = rule.endpoint
 
         return operation, args
-
-    def get_path_for_matching(self, request: Request) -> str:
-        """
-        Extracts the path from the request used for finding matching operations. This doesn't have to be the
-        path actually used in the router, since here we need a path that works for matching the werkzeug
-        rules generated from the specs.
-
-        An example where these paths can differ is for instance s3 paths with double slashes. This request::
-
-            GET /mybucket//mykey
-
-        Should match "GetObject", but the rule for this operation looks like ``/<Bucket>/<Key>``, which won't
-        match here because the path matcher doesn't match on prefixed slashes.
-
-        :param request: the service request
-        :return: the path used for matching
-        """
-        path = get_raw_path(request)
-        # merge slashes for matching.
-        path = multiple_slash_pattern.sub("/", path)
-        return path
--- a/localstack/aws/protocol/op_router.py
+++ b/localstack/aws/protocol/op_router.py
@@ -20,6 +20,11 @@ class GreedyPathConverter(PathConverter):
 
     regex = ".*?"
 
+    part_isolating = False
+    """From the werkzeug docs: If a custom converter can match a forward slash, /, it should have the
+    attribute part_isolating set to False. This will ensure that rules using the custom converter are
+    correctly matched."""
+
 
 class _HttpOperation(NamedTuple):
     """Useful intermediary representation of the 'http' block of an operation to make code cleaner"""
--- a/localstack/aws/protocol/op_router.py
+++ b/localstack/aws/protocol/op_router.py
@@ -258,7 +258,7 @@ def _create_service_map(service: ServiceModel) -> Map:
         rules=rules,
         strict_slashes=False,
         merge_slashes=False,
-        converters={"path": GreedyPathConverter},
+        # converters={"path": GreedyPathConverter},
     )
 
 
--- a/localstack/aws/protocol/op_router.py
+++ b/localstack/aws/protocol/op_router.py
@@ -248,7 +248,9 @@ def _create_service_map(service: ServiceModel) -> Map:
             # if there is only a single operation for a (path, method) combination,
             # the default Werkzeug rule can be used directly (this is the case for most rules)
             op = ops[0]
-            rules.append(_StrictMethodRule(string=rule_string, method=method, endpoint=op.operation))  # type: ignore
+            rules.append(
+                _StrictMethodRule(string=rule_string, method=method, endpoint=op.operation)
+            )  # type: ignore
         else:
             # if there is an ambiguity with only the (path, method) combination,
             # a custom rule - which can use additional request metadata - needs to be used
@@ -256,10 +258,14 @@ def _create_service_map(service: ServiceModel) -> Map:
 
     return Map(
         rules=rules,
+        # don't be strict about trailing slashes
         strict_slashes=False,
+        # we can't really use werkzeug's merge-slashes since it uses HTTP redirects to solve it
         merge_slashes=False,
-        # converters={"path": GreedyPathConverter},
     )
+
+
+multiple_slash_pattern = re.compile(r"/+")
 
 
 class RestServiceOperationRouter:
@@ -291,7 +297,8 @@ class RestServiceOperationRouter:
             # specified. the specs do _not_ contain any operations on OPTIONS methods at all.
             # avoid matching issues for preflight requests by matching against a similar GET request instead.
             method = request.method if request.method != "OPTIONS" else "GET"
-            rule, args = matcher.match(get_raw_path(request), method=method, return_rule=True)
+            path = self.get_path_for_matching(request)
+            rule, args = matcher.match(path, method=method, return_rule=True)
         except MethodNotAllowed as e:
             # MethodNotAllowed (405) exception is raised if a path is matching, but the method does not.
             # Our router handles this as a 404.
@@ -311,3 +318,24 @@ class RestServiceOperationRouter:
         operation: OperationModel = rule.endpoint
 
         return operation, args
+
+    def get_path_for_matching(self, request: Request) -> str:
+        """
+        Extracts the path from the request used for finding matching operations. This doesn't have to be the
+        path actually used in the router, since here we need a path that works for matching the werkzeug
+        rules generated from the specs.
+
+        An example where these paths can differ is for instance s3 paths with double slashes. This request::
+
+            GET /mybucket//mykey
+
+        Should match "GetObject", but the rule for this operation looks like ``/<Bucket>/<Key>``, which won't
+        match here because the path matcher doesn't match on prefixed slashes.
+
+        :param request: the service request
+        :return: the path used for matching
+        """
+        path = get_raw_path(request)
+        # merge slashes for matching.
+        path = multiple_slash_pattern.sub("/", path)
+        return path
--- a/localstack/services/s3/virtual_host.py
+++ b/localstack/services/s3/virtual_host.py
@@ -18,11 +18,13 @@ AWS_REGION_REGEX = r"(?:us-gov|us|ap|ca|cn|eu|sa)-[a-z]+-\d"
 # virtual-host style: https://{bucket-name}.s3.{region?}.{domain}:{port?}/{key-name}
 # ex: https://{bucket-name}.s3.{region}.localhost.localstack.cloud.com:4566/{key-name}
 # ex: https://{bucket-name}.s3.{region}.amazonaws.com/{key-name}
-VHOST_REGEX_PATTERN = f"<regex('.*'):bucket>.s3.<regex('(?:{AWS_REGION_REGEX}.)?'):region><domain>"
+VHOST_REGEX_PATTERN = (
+    f"<regex('.*'):bucket>.s3.<regex('(?:{AWS_REGION_REGEX}\\.)?'):region><domain>"
+)
 
 # path addressed request with the region in the hostname
 # https://s3.{region}.localhost.localstack.cloud.com/{bucket-name}/{key-name}
-PATH_WITH_REGION_PATTERN = f"s3.<regex('{AWS_REGION_REGEX}.'):region><domain>"
+PATH_WITH_REGION_PATTERN = f"s3.<regex('{AWS_REGION_REGEX}\\.'):region><domain>"
 
 
 class S3VirtualHostProxyHandler:
@@ -34,6 +36,7 @@ class S3VirtualHostProxyHandler:
     def __init__(self, proxy: Proxy = None):
         self.proxy = proxy or Proxy(
             forward_base_url=config.get_edge_url(),
+            # do not preserve the Host when forwarding (to avoid an endless loop)
             preserve_host=False,
             client=SimpleStreamingRequestsClient(),
         )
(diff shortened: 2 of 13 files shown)


thrau:
Sorry I totally missed https://github.com/localstack/localstack/pull/7274

thrau:
I think we're almost there, I think my changes may have broken these three tests:
```
FAILED tests/integration/s3/test_s3.py::TestS3::test_create_bucket_via_host_name
FAILED tests/integration/s3/test_s3.py::TestS3::test_bucket_name_with_dots - ...
FAILED tests/integration/s3/test_s3.py::TestS3::test_s3_delete_objects_trailing_slash
```

@bentsku could you take a look? :pray:

localstack/services/s3/virtual_host.py
```diff
@@ -9,20 +9,19 @@
...
 
 LOG = logging.getLogger(__name__)
 
+AWS_REGION_REGEX = r"(?:us-gov|us|ap|ca|cn|eu|sa)-[a-z]+-\d"
+
 # virtual-host style: https://{bucket-name}.s3.{region?}.{domain}:{port?}/{key-name}
 # ex: https://{bucket-name}.s3.{region}.localhost.localstack.cloud.com:4566/{key-name}
 # ex: https://{bucket-name}.s3.{region}.amazonaws.com/{key-name}
-VHOST_REGEX_PATTERN = f"<regex('.*'):bucket>.s3.<regex('({AWS_REGION_REGEX}\\.)?'):region><regex('.*'):domain><regex('(?::\\d+)?'):port>"
+VHOST_REGEX_PATTERN = f"<regex('.*'):bucket>.s3.<regex('(?:{AWS_REGION_REGEX}.)?'):region><domain>"
```
alexrashed, line 20: Did you remove the escaping of the `.` on purpose (i.e. change the semantics from "region followed by a ." to "region followed by any character")?

localstack/services/s3/virtual_host.py
```diff
@@ -42,20 +47,16 @@ def __call__(self, request: Request, **kwargs) -> Response:
         copied_headers["Host"] = forward_to_url.netloc
         copied_headers[S3_VIRTUAL_HOST_FORWARDED_HEADER] = request.headers["host"]
         # do not preserve the Host when forwarding (to avoid an endless loop)
```
alexrashed, line 49: nit: This comment refers to the Proxy instantiation. I think it should be moved to the init as well.

localstack/services/s3/virtual_host.py
```diff
@@ -9,20 +9,19 @@
 from localstack.runtime import hooks
 from localstack.services.edge import ROUTER
 from localstack.services.s3.utils import S3_VIRTUAL_HOST_FORWARDED_HEADER
-from localstack.utils.aws.request_context import AWS_REGION_REGEX
 
 LOG = logging.getLogger(__name__)
 
+AWS_REGION_REGEX = r"(?:us-gov|us|ap|ca|cn|eu|sa)-[a-z]+-\d"
```
alexrashed, line 15: Comment: While this regex is not as strict as the `AWS_REGION_REGEX` in the `request_context` module, it's definitely strict enough.

tests/unit/http_/test_router.py
```diff
@@ -176,6 +183,100 @@ def test_port_host_dispatcher(self):
                 )
             )
 
+    def test_path_converter(self):
```
alexrashed, line 186: The `path` converter comes with Werkzeug, which means these tests are more or less to define / explore the exact behavior of the builtin Werkzeug converter, right?

localstack/services/s3/virtual_host.py
```diff
@@ -86,43 +86,47 @@ def _rewrite_url(url: str, domain: str, bucket: str, region: str, port: str, **k
...
 
-@hooks.on_infra_ready(should_load=not config.LEGACY_S3_PROVIDER)
-def register_virtual_host_routes():
-    """
-    Registers the S3 virtual host handler into the edge router.
-
-    """
-    s3_proxy_handler = S3VirtualHostProxyHandler()
-    ROUTER.add(
+def add_s3_vhost_rules(router, s3_proxy_handler):
```
alexrashed, line 97: Nice refactoring of this proxy handler to allow unit testing 👍🏽

alexrashed (approved):
> groups in regex host matchers must be non-capturing, otherwise they break.

How exactly do they break? Is an exception raised? Or are they just not matching? Out of curiosity, is it possible to use named capture-groups and forward it to a kwarg of an endpoint function?

I only have a few other questions and nitpicks, and I haven't looked at why these S3 tests are failing, but the rest looks great! Thanks for jumping on this, such that we can finally use the latest version of this - quite essential - library again! 🚀

localstack/services/s3/virtual_host.py
thrau: whoopsie. no i didn't, good catch!

localstack/services/s3/virtual_host.py
```diff
@@ -31,6 +30,12 @@ class S3VirtualHostProxyHandler:
     addressed S3 bucket to a path addressed URL, to allow easy routing matching the ASF specs.
     """
 
+    def __init__(self, proxy: Proxy = None):
+        self.proxy = proxy or Proxy(
+            forward_base_url=config.get_edge_url(),
+            preserve_host=False,
+        )
```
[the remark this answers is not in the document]
thrau: that's a good point actually. may be better to introduce some sort of proxy factory and create a new one every time.

tests/unit/http_/test_router.py
thrau: essentially, yes.this was also the reason why we never added tests like this before, since it's just testing werkzeug. but they clearly document the behavior of the integration now.

thrau:
btw somewhat related i found the following


this works on aws:
```
aws s3api put-object --bucket test-bucket-xxxxxx --key //baz/bar
curl https://test-bucket-xxxxxx.s3.amazonaws.com///baz/bar
```

Whereas LocalStack says no. 

```
thomas@om ~ % curl https://test-bucket-xxxxxx.s3.localhost.localstack.cloud:4566///baz/bar
<?xml version='1.0' encoding='utf-8'?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><RequestId>1RUCBJV8EPH9DJLQRKOXAT4UHABQ5XEW1P9WCHVODPAB35R5OWIG</RequestId><Key>baz/bar</Key></Error>
```


even though it lists the keys correctly:
```
thomas@om ~ % awslocal s3api list-objects --bucket test-bucket-xxxxxx
{
    "Contents": [
        {
            "Key": "//baz/bar",
            "LastModified": "2023-04-19T18:34:56Z",
```

this probably has nothing to do with the GreedyPathMatcher (which explicitly merges slashes in this case: https://github.com/localstack/localstack/blob/579031e1f91b512cf2b90a6d3edcacb4d88d5412/tests/unit/aws/protocol/test_op_router.py#L40

however since this is only used for operation routing, not for extracting the key/bucket name, this shouldn't be the root cause.

conclusion: we should add s3 tests to validate the behavior /cc @bentsku

thrau:
It seems like the double slash problem is related to the firehose test in ext failing:
```
2023-04-19T13:25:33.2039809Z 2023-04-19T13:25:33.202 ERROR --- [   asgi_gw_1] localstack.aws.handlers.logging : exception during call chain
2023-04-19T13:25:33.2041489Z Traceback (most recent call last):
2023-04-19T13:25:33.2043030Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/parser.py", line 557, in parse
2023-04-19T13:25:33.2054525Z     operation, uri_params = self._operation_router.match(request)
2023-04-19T13:25:33.2055362Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/op_router.py", line 294, in match
2023-04-19T13:25:33.2057201Z     rule, args = matcher.match(get_raw_path(request), method=method, return_rule=True)
2023-04-19T13:25:33.2058270Z   File "/home/runner/work/localstack/localstack/localstack-ext/.venv/lib/python3.10/site-packages/werkzeug/routing/map.py", line 624, in match
2023-04-19T13:25:33.2059253Z     raise NotFound() from None
2023-04-19T13:25:33.2060035Z werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
2023-04-19T13:25:33.2060590Z 
2023-04-19T13:25:33.2063861Z The above exception was the direct cause of the following exception:
2023-04-19T13:25:33.2159101Z 
2023-04-19T13:25:33.2175487Z Traceback (most recent call last):
2023-04-19T13:25:33.2176007Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/chain.py", line 90, in handle
2023-04-19T13:25:33.2176430Z     handler(self, self.context, response)
2023-04-19T13:25:33.2176938Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/handlers/service.py", line 64, in __call__
2023-04-19T13:25:33.2177586Z     return self.parse_and_enrich(context)
2023-04-19T13:25:33.2178115Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/handlers/service.py", line 77, in parse_and_enrich
2023-04-19T13:25:33.2178747Z     operation, instance = parser.parse(context.request)
2023-04-19T13:25:33.2179265Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/parser.py", line 172, in wrapper
2023-04-19T13:25:33.2179645Z     return func(*args, **kwargs)
2023-04-19T13:25:33.2180115Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/parser.py", line 1061, in parse
2023-04-19T13:25:33.2180507Z     return super().parse(request)
2023-04-19T13:25:33.2180976Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/parser.py", line 172, in wrapper
2023-04-19T13:25:33.2181359Z     return func(*args, **kwargs)
2023-04-19T13:25:33.2181822Z   File "/home/runner/work/localstack/localstack/localstack/localstack/aws/protocol/parser.py", line 559, in parse
2023-04-19T13:25:33.2182264Z     raise OperationNotFoundParserError(
2023-04-19T13:25:33.2183562Z localstack.aws.protocol.parser.OperationNotFoundParserError: Unable to find operation for request to service s3: PUT /test-bucket-e230fc41//testdata/2023/04/19/13/fh-stream-5451c0f5-2023-04-19-13-25-33-f6d6afe7-b0b9-43b6-beb6-23f7049dad30
2023-04-19T13:25:33.2184963Z 2023-04-19T13:25:33.203  INFO --- [   asgi_gw_1] localstack.request.http    : PUT /test-bucket-e230fc41//testdata/2023/04/19/13/fh-stream-5451c0f5-2023-04-19-13-25-33-f6d6afe7-b0b9-43b6-beb6-23f7049dad30 => 500
```

bentsku:
thanks @thrau, I'll also take a look in master to see if that's an issue in both path style and vhost style, with "prefix" slashes. We already fixed an issue with the key and trailing slashes, which was fixed in the parser (it was removing extra slashes too). I'll add tests in both mode after the hackathon and check the behaviour! thanks for bringing this up.

thrau:
I think i fixed the op router, but there's an issue in the parser now:

These are two separate keys:

```
aws s3api put-object --bucket test-bucket-1f48676a --key test-key 
aws s3api put-object --bucket test-bucket-1f48676a --key /test-key
```

```
 % aws s3api list-objects --bucket test-bucket-1f48676a
```

```json
{
    "Contents": [
        {
            "Key": "/test-key",
            "LastModified": "2023-04-20T14:11:33.000Z",
            "Size": 0,
            "StorageClass": "STANDARD",
            "Owner": {
                "DisplayName": "thomas.rausch",
                "ID": "40331a3a79454597cdd6369f2101c5f25cc433c7a81b71463cc0e9d51ee1e44d"
            }
        },
        {
            "Key": "test-key",
            "LastModified": "2023-04-20T14:11:30.000Z",
            "Size": 0,
            "StorageClass": "STANDARD",
            "Owner": {
                "DisplayName": "thomas.rausch",
                "ID": "40331a3a79454597cdd6369f2101c5f25cc433c7a81b71463cc0e9d51ee1e44d"
            }
        }
    ]
}
```


But it seems this behavior is now broken in our parser, since this behavior fails:

```
tests/unit/aws/protocol/test_parser.py:1076 (test_s3_get_object_keys_with_slashes)
{'Bucket': 'test-bucket', 'Key': 'test-key'} != {'Bucket': 'test-bucket', 'Key': '/test-key'}

Expected :{'Bucket': 'test-bucket', 'Key': '/test-key'}
Actual   :{'Bucket': 'test-bucket', 'Key': 'test-key'}
```

thrau:
I was also anticipating some performance improvements in the op router, given that the werkzeug router refactoring was aimed at [implementing a more efficient routing algorithm](https://pgjones.dev/blog/faster-routing-2022/).

Sadly, it seems it's not really helping, or at least I wasn't able to reproduce the 5x speedup.
Here's a simple benchmark I ran against both the old and new version of werkzeug, both of which lead to similar results.

```python
import timeit

from localstack.aws.protocol.op_router import RestServiceOperationRouter
from localstack.http import Request

op_router = RestServiceOperationRouter(service=load_service("apigateway"))


def match_routes():
    op, params = op_router.match(Request("POST", "/restapis/myrestapi/deployments"))
    assert op.name == "CreateDeployment"
    assert params == {"restapi_id": "myrestapi"}


def main():
    print(timeit.timeit(stmt=match_routes, number=100000))


if __name__ == "__main__":
    main()
```

localstack/aws/protocol/op_router.py
```diff
@@ -13,6 +13,11 @@
 
 
 class GreedyPathConverter(PathConverter):
+    """
+    This converter makes sure that the path ``/mybucket//mykey`` can be matched to the pattern
+    ``<Bucket>/<path:Key>`` and will result in `Key` being `/mykey`.
+    """
+
     regex = ".*?"
```
alexrashed, line 21: Just for the record (and maybe as an extension to the "haunting" section in the PR description):
When I was looking at the issue I was initially suspecting it to originate from the fact that the converters now define if they are `part_isolating` or not, and it is implicitly determined by checking if the regex contains a `/` (see https://github.com/pallets/werkzeug/blob/e0afdfb0d549b4fea34e0f91270d784efe7b0e1b/src/werkzeug/routing/converters.py#L19-L36).
Here this custom converter does not have a `part_isolating` field defined, and does not contain a slash in the regex (contrary to the superclass), which leads to `part_isolating=False`.
In fact, this is the correct configuration, because this greedy path converter isn't only greedy for leading slashes, but can also match across multiple path segments.

I'm actually not really a fan of the implicit configuration here (it feels like we were lucky), but I also understand that Werkzeug does this to increase backwards compatibility. Maybe we should state this here explicitly?

alexrashed (approved):
Wow, that was quite a grind. It was quite a fundamental change in the routing in Werkzeug, we are tightly coupled to Werkzeug, and the AWS (and in particular S3) routing is quite complex!
Great change, and it's great to see that we are on the latest version of this core framework again! 🚀

tests/unit/aws/protocol/test_op_router.py
```diff
@@ -59,6 +59,20 @@ def test_s3_head_request():
...
+
+    op, params = router.match(Request("POST", "/restapis/myrestapi/deployments"))
+    assert op.name == "CreateDeployment"
+    assert params == {"restapi_id": "myrestapi"}
+
+    with pytest.raises(NotFound):
+        # note: this is to document the behavior of werkzeug, not necessarily what we want for the op router.
+        # the behavior of double slashes in parameters probably cannot be generalized and will be handled by
+        # services case-by-case
+        router.match(Request("POST", "/restapis/myrestapi//deployments"))
```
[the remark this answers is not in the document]
thrau: The GreedyPathConverter is only used in the operation router, not the default router, and should IMO also be contained there. This is using the default werkzeug behavior that also underlies the router, which I think is the correct way. Other opinions?

localstack/aws/protocol/op_router.py
thrau: great find. i added the explicit `part_isolating=False` and also a comment for future us :grin:

thrau:
the two failing tests seem flaky and work on my machine

PR #8109, opened 2023-04-10
Replace SES Client In SNS Publisher

pinzon:
This PR adds a small change to use the new AWS Internal Client.
It does not call `request_metadata` due to the Publisher being able to just send emails without needing an Identity Policy on the address.


bentsku (approved):
LGTM! Thanks for taking care of this! 👍

PR #8112, opened 2023-04-11
Fix containerd detection for host mode

simonrw:
Previously, if _any_ containerd container was running on the host, and LocalStack was started in host mode then it would detect the `containerd` string in the disk partitions, and return True. This is not a great default for host mode, as it binds LocalStack to 0.0.0.0.

This change searches through the `/proc/1/mountinfo` virtual file system looking for mount options that include `containerd`. This means that only the root mount point is considered, rather than any host mounts that correspond to other containers. After this update, while running a containerd container in the background and starting LocalStack in host mode, the docker status is correctly determined to be False, and LocalStack is bound to 127.0.0.1.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -9,8 +9,6 @@ import time
 from dataclasses import dataclass
 from typing import Any, Dict, List, Mapping, Tuple, TypeVar
 
-import psutil
-
 from localstack import constants
 from localstack.constants import (
     AWS_REGION_US_EAST_1,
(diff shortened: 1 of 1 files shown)


thrau (commented):
thanks for tackling this simon! i'm worried about importing psutil into the config. in general it should be free from third-party tools as much as possible.

is there a different way we could do this? also adding @dfangl as reviewer.

simonrw:
> thanks for tackling this simon! i'm worried about importing psutil into the config. in general it should be free from third-party tools as much as possible.
> 
> is there a different way we could do this? also adding @dfangl as reviewer.

That's fair. The `/proc/<pid>/mountinfo` file is not hard to parse. We should also check that it exists first :+1:

PR #8114, opened 2023-04-11
Use explicit region for test fixtures

viren-nadkarni:
This PR makes all test fixtures use a explicit region instead of falling back the region set in system config.

Earlier, when run locally, the fixtures used the system region to create resources. Later the assertions would query a different region (`us-east-1`) eg. via endpoints as in SQS and fail.


localstack/testing/aws/util.py
```diff
@@ -180,7 +181,10 @@ def base_aws_client_factory(session: boto3.Session) -> ClientFactory:
     config = None
     if os.environ.get("TEST_DISABLE_RETRIES_AND_TIMEOUTS"):
         config = botocore.config.Config(
-            connect_timeout=1_000, read_timeout=1_000, retries={"total_max_attempts": 1}
+            connect_timeout=1_000,
+            read_timeout=1_000,
+            retries={"total_max_attempts": 1},
+            region_name=TEST_AWS_REGION_NAME,
```
dominikschubert, line 187: Not sure about this one. I think it might be better if the region is only added as a fallback in the else branch below? (i.e. when `os.environ.get("TEST_TARGET") != "AWS_CLOUD"`)

localstack/testing/aws/util.py
```diff
@@ -180,12 +181,17 @@ def base_aws_client_factory(session: boto3.Session) -> ClientFactory:
...
+            retries={"total_max_attempts": 1},
         )
 
     if os.environ.get("TEST_TARGET") == "AWS_CLOUD":
         return ExternalAwsClientFactory(session=session, config=config)
     else:
+        if config:
+            # Prevent this fixture from using the region configured in system config
+            config = config.merge(botocore.config.Config(region_name=TEST_AWS_REGION_NAME))
         return ExternalClientFactory(session=session, config=config)
```
dominikschubert, line 195: otherwise this would only be set if TEST_DISABLE_RETRIES_AND_TIMEOUTS is set

```suggestion
        if not config:
            config = botocore.config.Config()
            
        # Prevent this fixture from using the region configured in system config
        config = config.merge(botocore.config.Config(region_name=TEST_AWS_REGION_NAME))
        return ExternalClientFactory(session=session, config=config)
```

dominikschubert (changes requested):

viren-nadkarni:
Failing tests are not related to the changes proposed in this PR

```
=========================== short test summary info ============================
FAILED ../localstack/tests/integration/docker_utils/test_docker.py::TestDockerClient::test_push_access_denied[CmdDockerClient]
FAILED ../localstack/tests/integration/docker_utils/test_docker.py::TestDockerClient::test_push_access_denied[SdkDockerClient]
= 2 failed, 778 passed, 173 skipped, 977 deselected, 18 xfailed, 6 xpassed, 2785 warnings, 4 rerun in 2857.78s (0:47:37) =
```

PR #8116, opened 2023-04-11
Introduce usage counters

dominikschubert:
Based upon the analytics logger, @ackdav and I prototyped a fairly simple usage logger a few weeks ago.

One of the constraints for the prototype was to not add further pressure on the analytics backend, so we aggregate all captured data & "flush the logs" only on shutdown. This can be somewhat unreliable though, since in some cases the LocalStack instance might not be properly terminated and thus never wait for the shutdown to actually execute.

A few examples of what kind of information we wanted to collect:

* Lambda runtimes
* Usage of certain features (e.g. Lambda hot-reloading)
* Supported CloudFormation models
* Unsupported CloudFormation models

Most of this data can also be captured via structured logging and log aggregation but for our purposes (forwarding to the analytics backend) this seemed impractical.

### DIFF
--- a/localstack/services/awslambda/invocation/docker_runtime_executor.py
+++ b/localstack/services/awslambda/invocation/docker_runtime_executor.py
@@ -9,6 +9,7 @@ from typing import Callable, Dict, Literal, Optional
 from localstack import config
 from localstack.aws.api.lambda_ import Architecture, PackageType, Runtime
 from localstack.services.awslambda import hooks as lambda_hooks
+from localstack.services.awslambda import usage
 from localstack.services.awslambda.invocation.executor_endpoint import (
     INVOCATION_PORT,
     ExecutorEndpoint,
@@ -264,6 +265,7 @@ class DockerRuntimeExecutor(RuntimeExecutor):
         return f"{container_name}-{self.id}"
 
     def start(self, env_vars: dict[str, str]) -> None:
+        start_time = time.perf_counter()
         self.executor_endpoint.start()
         network = self._get_network_for_executor()
         container_config = LambdaContainerConfiguration(
@@ -336,6 +338,8 @@ class DockerRuntimeExecutor(RuntimeExecutor):
             for source, target in container_config.copy_folders:
                 CONTAINER_CLIENT.copy_into_container(self.container_name, source, target)
 
+        end_time = time.perf_counter()
+        usage.initduration.record_value(end_time - start_time)
         CONTAINER_CLIENT.start_container(self.container_name)
         self.ip = CONTAINER_CLIENT.get_container_ipv4_for_network(
             container_name_or_id=self.container_name, container_network=network
--- a/localstack/services/awslambda/invocation/docker_runtime_executor.py
+++ b/localstack/services/awslambda/invocation/docker_runtime_executor.py
@@ -9,7 +9,6 @@ from typing import Callable, Dict, Literal, Optional
 from localstack import config
 from localstack.aws.api.lambda_ import Architecture, PackageType, Runtime
 from localstack.services.awslambda import hooks as lambda_hooks
-from localstack.services.awslambda import usage
 from localstack.services.awslambda.invocation.executor_endpoint import (
     INVOCATION_PORT,
     ExecutorEndpoint,
@@ -265,7 +264,6 @@ class DockerRuntimeExecutor(RuntimeExecutor):
         return f"{container_name}-{self.id}"
 
     def start(self, env_vars: dict[str, str]) -> None:
-        start_time = time.perf_counter()
         self.executor_endpoint.start()
         network = self._get_network_for_executor()
         container_config = LambdaContainerConfiguration(
@@ -338,8 +336,6 @@ class DockerRuntimeExecutor(RuntimeExecutor):
             for source, target in container_config.copy_folders:
                 CONTAINER_CLIENT.copy_into_container(self.container_name, source, target)
 
-        end_time = time.perf_counter()
-        usage.initduration.record_value(end_time - start_time)
         CONTAINER_CLIENT.start_container(self.container_name)
         self.ip = CONTAINER_CLIENT.get_container_ipv4_for_network(
             container_name_or_id=self.container_name, container_network=network
--- a/localstack/services/awslambda/usage.py
+++ b/localstack/services/awslambda/usage.py
@@ -1,0 +1,5 @@
+from localstack.utils.analytics.usage import UsageCounter, UsageSetCounter
+
+hotreload = UsageCounter("lambda:hotreload", aggregations=["sum"])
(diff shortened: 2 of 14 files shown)


localstack/utils/analytics/usage.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def aggregate(self) -> dict:
+        aggregated_payload = {}
+        for ns, collector in collector_registry.items():
+            aggregated_payload[ns] = collector.aggregate()
+        return aggregated_payload
+
+    def flush(self):
+        self.handler.handle(
+            Event(name="hackystacky:usage", metadata=self._metadata(), payload=self.aggregate())
```
thrau, line 97: :eyes: :technologist:

localstack/utils/analytics/__init__.py
```diff
@@ -10,3 +11,7 @@ def _create_global_analytics_bus():
 
 
 log = EventLogger(handler=_create_global_analytics_bus(), session_id=get_session_id())
+
+# our own "GlobalAnalyticsBus"
+usage_event_handler = UsageEventHandler()
+usage_log = UsageLogger(handler=usage_event_handler)
```
thrau, line 17: how does the usage logger compare to the `EventLogger`? is this an orthogonal or a competing construct? when should i use one or the other?

thrau (commented):
Conceptually this looks pretty good to me! We can start collecting these metrics at shutdown for now and maybe in the future think about how we can make it a bit more robust against ungraceful termination (but first investigate whether that's even worth it).

the `UsageSetCounter` seems pretty solid. The `UsageLogger` i'm not too sure about.

documenting the internal API to get a better grip on how the constructs are supposed to be used would be great!

localstack/utils/analytics/__init__.py
dominikschubert: It doesn't actually need to be a separate logger since it shouldn't ever be used directly. Only the counters or other similar constructs should be used here :thinking: 

So I could image just removing it here and keeping it only as a local concept in `localstack/utils/analytics/usage.py`, that way there shouldn't be any confusion hopefully.

If you meant this in a more general sense, I'd say anything that you want to use aggregations with and where we are OK with missing out on some values potentially, should use these constructs.

thrau (approved):
looks good now! a bit of documentation couldn't hurt, but other than that: let's give this a go :-) cc @ackdav

ackdav (approved):
🚀 thx for pushing this from prototype to integrating it. I'm sure this enables some really interesting insights!

By the way - is this respecting the `DISABLE_EVENTS` config flag?

one nit - I would like to introduce some namespacing in the analytics events. Something like `product:event_name` - this will make sifting through data much easier in the future. (also calling it "usage" isn't really telling us much 😉). So could we rename this event `ls:usage_analytics` ? (happy for better ideas - naming is hard and we're not enforcing naming schemes yet)

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -73,6 +74,8 @@ def get_deployment_config(res_type):
     resource_class = RESOURCE_MODELS.get(canonical_type)
     if resource_class:
         return resource_class.get_deploy_templates()
+    else:
+        usage.missing_resource_types.record(canonical_type)
```
joe4dev, line 78: very useful 👍

localstack/services/awslambda/invocation/docker_runtime_executor.py
```diff
@@ -336,6 +338,8 @@ def start(self, env_vars: dict[str, str]) -> None:
             for source, target in container_config.copy_folders:
                 CONTAINER_CLIENT.copy_into_container(self.container_name, source, target)
 
+        end_time = time.perf_counter()
+        usage.initduration.record_value(end_time - start_time)
```
joe4dev, line 342: I guess that's the easiest way to reliably measure container initialization overhead. Terminology-wise, this is something completely different compared to what AWS refers to `INIT` https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html

In the config, we use `LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT`. Potential alternatives are `container_init_duration` or more generic `env_init_duration`.

I understand `initduration`; just wanted to mention the different meaning compared to AWS terminology so we are aware of it.

localstack/services/awslambda/usage.py
```diff
@@ -0,0 +1,10 @@
+"""
+Usage reporting for Lambda service
+"""
+from localstack.utils.analytics.usage import UsageCounter, UsageSetCounter
+
+# usage of lambda hot-reload feature
```
joe4dev, line 6: nit: description could be more specific (e.g., number of functions created with hot-reload feature)

localstack/services/awslambda/usage.py
```diff
@@ -0,0 +1,10 @@
...
+hotreload = UsageCounter("lambda:hotreload", aggregations=["sum"])
+
+# used unique lambda runtimes (e.g. python3.7)
```
joe4dev, line 9: nit: Number of function invocations per Lambda runtime (e.g., python3.7)

joe4dev (approved):
Looking forward to these insights 🔮 🚀 

main comment: potential demo `hotreload` instrumentation in `list_functions` and tests 🔴 
otherwise just nits

localstack/services/infra.py
```diff
@@ -272,6 +273,9 @@ def stop_infra():
     if events.infra_stopping.is_set():
         return
 
+    usage.aggregate_and_send()
+    # analytics.log.event("infra_stop")
```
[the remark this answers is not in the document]
dominikschubert: yeah initially it was, but not required anymore :+1:

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -647,6 +650,7 @@ def get_resource_model_instance(resource_id: str, stack) -> Optional[GenericBase
     canonical_type = canonical_resource_type(resource_type)
     resource_class = RESOURCE_MODELS.get(canonical_type)
     if not resource_class:
+        # usage.missing_resource_types.record(canonical_type)
```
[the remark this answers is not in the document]
dominikschubert: Kinda. It's a bit unclear actually in CFn right now where it's best to put the logging, just because the template deployer is still so convoluted. I'll remove it for now though. We'll need to re-evaluate the placement after an initial rework of the deployer

localstack/services/awslambda/invocation/docker_runtime_executor.py
dominikschubert: Actually that should be removed right now, since I also removed the `initduration` counter in a previous commit :grimacing: 

But your point is still valid :+1:  We can think about adding this back at a later date

PR #8119, opened 2023-04-12
fix ASF operation signature mismatches

alexrashed:
One of the core foundations of ASF are the generated API classes based on the service specifications contained in the latest `botocore` package. These generated APIs are contained in the `localstack.aws.api` module and are updated on a weekly basis by the ["Update ASF APIs"](https://github.com/localstack/localstack/actions/workflows/asf-updates.yml) workflow.

Unfortunately, these updates can cause problems if the signature of the generated APIs change, but the corresponding provider method signatures - implementing the abstract generated API methods - are not adjusted.
This also caused issues with our parity metric collection in the past. /cc @steffyP 

For example, the latest run of the [Update ASF APIs last Monday](https://github.com/localstack/localstack/actions/runs/4654646951) created PR #8106 with recent API changes related to the Lambda Response Streaming (as mentioned by @joe4dev in the PR). This changes the signature of `LambdaApi.create_function_url_config` by adding a new parameter called `invoke_mode`:
https://github.com/localstack/localstack/blob/9debbcc15c83727a560c78f336b0cc44dc516cec/localstack/aws/api/lambda_/__init__.py#L1831

This breaks the invocation of the implementing provider function in `LambdaProvider.create_function_url_config` if this new parameter is used by a client.

- b50c0a1db990f53836787160aacc530cfae7581f adds a new unit test which tries to discover all provider classes based on our conventions (classes where the name is ending with `Provider`, contained in modules called `provider`).
  - I was playing around with discovering the provider classes based on the Plux entrypoints, but I was having a hard time collecting _all_ available provider classes. I'm happy for any feedback concerning the simplification of the provider and base class detection.
  - The current approach detects a total of 2383 operation signatures / tests.
  - The execution time seems to be ~15 seconds for all these tests (comparing [the current run of this branch - 2m48s](https://app.circleci.com/pipelines/github/localstack/localstack/14134/workflows/f2de07a1-1a3c-43ae-98c3-1ccd6fd1237d/jobs/105027) - with the [latest run on master - 2m33s](https://app.circleci.com/pipelines/github/localstack/localstack/14129/workflows/96c4e38f-1f84-4ad7-a0d4-de5e2e1fa224/jobs/104994)).
- 7a7f96307d4c8e0767299de9101875ba6f77486d fixes current issues with the provider signatures detected with the unit test.

### DIFF
--- a/localstack/services/awslambda/provider.py
+++ b/localstack/services/awslambda/provider.py
@@ -64,6 +64,7 @@ from localstack.aws.api.lambda_ import (
     InvocationResponse,
     InvocationType,
     InvokeAsyncResponse,
+    InvokeMode,
     LambdaApi,
     LastUpdateStatus,
     LayerName,
@@ -1675,6 +1676,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         auth_type: FunctionUrlAuthType,
         qualifier: FunctionUrlQualifier = None,
         cors: Cors = None,
+        invoke_mode: InvokeMode = None,
     ) -> CreateFunctionUrlConfigResponse:
         state = lambda_stores[context.account_id][context.region]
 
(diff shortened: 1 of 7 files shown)


thrau (approved):
great work @alexrashed! i think the approach to discover the service providers is fine for now. we're not really changing much in the way we structure the providers or organize the code now, so it seems fairly robust. when we iterate over the `Service` concept, we can think about making it easier to find just the provider classes to do some static analysis like the one you're doing, without having to instantiate them (like in the approach i posted above)
:100:

PR #8126, opened 2023-04-12
Send service principal when assuming role, check if lambda can assume role before allowing creation

dfangl:
## Motivation
IAM Enforcement for assuming roles can be enabled in -ext now, so it makes sense to actually send the relevant data with assume role requests.

For lambda in particular, the creation of a function already fails when assuming the role does not work, it does not fail later.

## Changes
* Send source service if provided when creating a client for a role using the new client factory
* Check if a role can be assumed by lambda before creating the function. This can only be properly tested with IAM enforcement on, so no point doing so in this repo.


joe4dev (approved):
Tested against companion PR in ext ✅ 
https://github.com/localstack/localstack-ext/pull/1610

dominikschubert (approved):
LGTM!

Issue #8118, opened 2023-04-12
bug: SNS Subscription Confirmation fails with: InvalidParameter: Token


dave-mclean:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you perform the SubscriptionConfirmation request after subscribing an HTTP endpoint to an SNS topic localstack responses with:

```
<?xml version='1.0' encoding='utf-8'?>
<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
    <Error>
        <Code>InvalidParameter</Code>
        <Message>Invalid parameter: Token</Message>
        <Type>Sender</Type>
    </Error>
    <RequestId>WC33COKFDOENIKQE90648TUZNUJFAISR71ZLRPFDAXGN1MTHHX02</RequestId>
</ErrorResponse>%
```

When you subscribe an HTTP endpoint to an SNS Queue you receive a SubscriptionConfirmation message as described in [the docs](https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.prepare.html). 

Endpoints should parse the message and extract the SubsbscriptionConfirmationURL, then perform a GET request to that URL to confirm the subscription. 

**In localstack/localstack:1.1.0 this process works well**

The following code is used to create a bucket and a topic and subscribe
```
#! /bin/sh

BUCKET=test-bucket
TOPIC=test-topic
CONFIG_FILE=local/usr/s3-config.json
ENDPOINT=http://webapp:8000/api/process_dataflow

# Create the bucket and topic
awslocal s3api create-bucket --bucket $BUCKET
awslocal sns create-topic --name $TOPIC

# Configure the bucket to notify the topic when an object is created in it.
awslocal s3api put-bucket-notification-configuration --bucket $BUCKET --notification-configuration file://$CONFIG_FILE

# Subscribe the Dataflow processing endpoint to the topic
awslocal sns subscribe --topic-arn arn:aws:sns:us-east-1:000000000000:$TOPIC --protocol http --notification-endpoint $ENDPOINT
```

After the command to subscribe the endpoint to the topic the endpoint receives an SNS SubscriptionConfirmation message like this:

```
{
    "Message": "You have chosen to subscribe to the topic arn:aws:sns:us-east-1:000000000000:test-topic.\nTo confirm the subscription, visit the SubscribeURL included in this message.",
    "MessageId": "fd13c8a3-9e86-4c93-8d1d-4557aea236a3",
[...]

PR #8129, opened 2023-04-13
fix ConfirmSubscription access to SNS store

bentsku:
As reported in #8118, there was an issue with confirming a subscription from a simple http client, because we would use the region from the context, which was using the default region of LocalStack (set from the auth handler `MissingAuthHeaderInjector`). However, we should use the region from the `TopicARN`. I've re-validated the tests against AWS with a local setup with https://localhost.run, and could validate different behaviours that were untested. This adds more parity into our HTTP subscriptions, and fix the linked issue. 

Also, I've sneaked a little store change to get the parity needed. This should not break persistence nor cloud pods as tokens are usually short lived, and it is more additive. We won't try to get a topic ARN anymore from this dict, so it would not create error per se. 

_fixes #8118_

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -404,14 +404,33 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         authenticate_on_unsubscribe: authenticateOnUnsubscribe = None,
     ) -> ConfirmSubscriptionResponse:
         # TODO: validate format on the token (seems to be 288 hex chars)
-        store = self.get_store(account_id=context.account_id, region=context.region)
-        topic_tokens = store.subscription_tokens.get(topic_arn, {})
+        # this request can come from any http client, it might not be signed (we would need to implement
+        # `authenticate_on_unsubscribe` to force a signing client to do this request.
+        # so, the region and account_id might not be in the request. Use the ones from the topic_arn
+        try:
+            parsed_arn = parse_arn(topic_arn)
+        except InvalidArnException:
+            raise InvalidParameterException("Invalid parameter: Topic")
 
-        subscription_arn = topic_tokens.pop(token, None)
+        store = self.get_store(account_id=parsed_arn["account"], region=parsed_arn["region"])
+
+        # it seems SNS is able to know what the region of the topic should be, even though a wrong topic is accepted
+        if parsed_arn["region"] != get_region_from_subscription_token(token):
+            raise InvalidParameterException("Invalid parameter: Topic")
+
+        subscription_arn = store.subscription_tokens.get(token)
         if not subscription_arn:
             raise InvalidParameterException("Invalid parameter: Token")
 
         subscription = store.subscriptions.get(subscription_arn)
+        if not subscription:
+            # subscription could have been deleted in the meantime
+            raise InvalidParameterException("Invalid parameter: Token")
+
+        # ConfirmSubscription is idempotent
+        if subscription.get("PendingConfirmation") == "false":
+            return ConfirmSubscriptionResponse(SubscriptionArn=subscription_arn)
+
         subscription["PendingConfirmation"] = "false"
         subscription["ConfirmationWasAuthenticated"] = "true"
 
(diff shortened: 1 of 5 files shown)


localstack/services/sns/provider.py
```diff
@@ -404,14 +404,33 @@ def confirm_subscription(
...
+        if parsed_arn["region"] != get_region_from_subscription_token(token):
+            raise InvalidParameterException("Invalid parameter: Topic")
 
-        subscription_arn = topic_tokens.pop(token, None)
+        subscription_arn = store.subscription_tokens.get(token)
         if not subscription_arn:
             raise InvalidParameterException("Invalid parameter: Token")
 
         subscription = store.subscriptions.get(subscription_arn)
+        if not subscription:
```
bentsku, line 426: I should add a todo here. It might be that confirming a deleted subscription might actually recreate it, from the description of `UnsubscribeURL` messages. However, this is out of scope of the PR.

tests/unit/test_sns.py
```diff
@@ -579,3 +583,19 @@ def test_filter_policy_on_message_body(self):
...
+        token = encode_subscription_token_with_region(region)
+        assert len(token) == 64
+        token_region = get_region_from_subscription_token(token)
+        assert token_region == region
+
+    @pytest.mark.parametrize(
+        "token", ["abcdef123", "mynothexstring", "us-west-2", b"test", b"test2f", "test2f"]
+    )
+    def test_decode_token_with_no_region_encoded(self, token):
+        with pytest.raises(Exception) as e:
```
dave-mclean, line 598:
```suggestion
        with pytest.raises(InvalidParameterException) as e:
```
I think being more specific about what exception is raised here might catch possible errors in future.

bentsku: I am matching on the exception under, and if the exception raised does not have the following message `"Invalid parameter: Token"`, the test will fail anyway. Agreed that it could be safer though, I'll add it after the second review, thanks for pointing this out!

thrau (commented):
Thanks for tackling this so quickly @bentsku. Amazing test coverage as always!

Just one comment/question I think needs clarification:
> This should not break persistence nor cloud pods as tokens are usually short lived, and it is more additive.

i couldn't find code that deletes the `subscription_tokens`. there was one instance of `subscription_tokens.pop` that was removed - so it seems to me that, as soon as you have one subscription in your SNS pod, the `subscription_tokens` will be populated with state, that will be serialized into the store.
persistence will definitely break in that case, since we have `Dict[str, Dict[str, str]]` in the attribute, but now expect `Dict[str, str]`

if i'm missing something here and am wrong then let's go ahead and merge for the patch release, otherwise i think it would be good to stage it for 2.1

bentsku:
Yes, the persistence was a bit of a pickle.
Before, we would delete the token right away, but it wasn't the right behaviour as we still need to be able to match a token to a subscription.
So in my head, in existent cloud pods / persistence, there should not be any token saved (famous last words), or it would be at an unfortunate moment, between creating a subscription and confirming it. Only newly created pods will always have subscription tokens. 

I guess it boils down to a question I didn't have the answer: when restoring persistence, will the new store keep the old shape, or be cast to the new shape? And does the shape really have an incidence?
Because let's say the old store looks like this:
```python
subscription_tokens = {
  "faketopicarn": {
    "token": "subscriptionarn"
  }
}
# this would be old shape that is still in the store after restore
```
When accessing that store, we would just try to get the token from it and it would not exist basically, because it's one level deeper. We would break this exact moment of creating a subscription, saving the pod, restoring the pod, validating the subscription. I guess that's a very specific scenario. 

But if that's still a risk, we can stage it for 2.1, no problem for me! Might be safer, to avoid any issue.

tests/unit/test_sns.py
dave-mclean: ah interesting, I didn't see the assert underneath - sorry!

I imagine you already know this (and therefore have a reason for not doing it!) but just in case you can combine that `match` with `pytest.raises` into one line like this:

```
with pytest.raises(InvalidParameterException, match="Invalid parameter: Token"):
    get_region_from_subscription_token(token)
```

bentsku: Thanks for tip! I guess I like having explicit `assert` throughout the test, to quickly spot what we are looking for. But that's a good tip, thanks!

PR #8134, opened 2023-04-13
fix invalid access to thread-local access key ID

alexrashed:
Recently, a CORS test continuously fails in CI, while it is not reproducible locally:
This PR aims at fixing this issue and unblocking the community pipeline.

This happens by disabling access to the threadlocal stored access key id before it is being set in the Authorization Header Injector.

Edit: It turns out that the flaky tests were caused by an invalid fallback to a thread-local. Details can be found in the great writeup of @dfangl: https://github.com/localstack/localstack/pull/8134#issuecomment-1512906567

PR #8135, opened 2023-04-13
xfail TestS3Cors to allow green build until investigation

bentsku:
Unblocking the pipeline until investigation of the CI failure


alexrashed (approved):
Thanks for jumping on this!
We just discussed that our next steps forward are:
- Merging this PR to unblock the pipeline. The issue is not caused on CORS / S3, but seems to be caused by another test influencing the now failing one.
- Combine #7985 and #8134, properly investigate on what causes the issue, fix it, and re-enable the tests with this PR asap.

bentsku:
Sure, let's do this! I'm a bit worried about merging our 2 PRs, as I was unable to reproduce with #7985, and again the tests passed. Should we just investigate on #8134?

PR #8143, opened 2023-04-14
sns firehose client update

pinzon:
Changes:
- [x] Update Firehose Client in SNS Publisher 
- [X] Update S3 Client in Firehose Provider


bentsku (approved):
LGTM 👍

PR #8144, opened 2023-04-14
remove SNS create_sqs_message_attributes and add DLQ for Fifo topic

bentsku:
This PR continues on #8029. I actually only fixed the DLQ behaviour for FIFO topics for `publish_batch` and not for a simple `publish`.

I've also refactored our way of passing argument to the SQS client in the publisher, to allow reuse between the regular SQS publisher and the batch one, also when sending it down the DLQ. 

I've also removed `create_sqs_message_attributes` as it was a remnant from the `ProxyListener`, and now ASF properly decodes and casts the `MessageBodyAttributeMap` into the right format. It might have been the issue for https://discuss.localstack.cloud/t/i-have-been-stuck-on-binascii-error-incorrect-padding-not-sure-if-it-is-an-error/297 but not sure yet, anyway, after testing, it was not used anymore.

There's a follow up PR building on this to test and implement deduplication for FIFO topic, see #8145

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -654,6 +654,11 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
             # TODO: test how this behaves in a FIFO context with a lot of threads.
             self._publisher.publish_to_topic(publish_ctx, topic_or_target_arn)
 
+        if is_fifo:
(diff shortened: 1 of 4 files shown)


thrau (approved):
nice! LGTM :+1:

PR #8145, opened 2023-04-14
SNS test and add deduplication for FIFO topic

bentsku:
This is building on top of #8144 and #8146, and adds test and implement the deduplication for FIFO topics. Right now, we are just properly using the SQS deduplication features, but SNS one is a bit different regarding the `MessageGroupId`, see https://docs.aws.amazon.com/sns/latest/dg/fifo-message-dedup.html, the deduplication is at the topic level. 
We are also not storing the deduplication ID yet, so we cannot return the proper `MessageId` and `SequenceNumber` from the publishing call, which we should do.


baermat (approved):
LGTM 👍

Issue #6766, opened 2022-08-26
bug: concurrent requests made to a `FIFO` queue return messages from the same `MessageGroupId`


tomazfernandes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When concurrent requests are made to a `FIFO` queue, requests return messages from the same `MessageGroupId`.

#### Scenario Example

Consider we have a FIFO queue with 10 messages from the same `MessageGroupId`.

Then we make two concurrent `receiveMessage` requests for `5 messages` each.

According to `SQS FIFO` contract, we should receive only 5 messages in the first request, and none on the second request, as long as `Visibility Timeout` is not expired.

With `LocalStack 1.0.4` wrongly we receive all 10 messages, 5 in each request.

When the same is done targeting `AWS`, the behavior is correct, the first request returns 5 messages and the second one no messages.

### Expected Behavior

We should not receive messages from the same `MessageGroupId` for concurrent requests to `FIFO` queues.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

With `TestContainers`
```java
private static final String LOCAL_STACK_VERSION = "localstack/localstack:1.0.4";

static LocalStackContainer localstack = new LocalStackContainer(DockerImageName.parse(LOCAL_STACK_VERSION))
	.withServices(SQS).withReuse(true);
```

I've tried with `reuse(false)` and the behavior is unchanged.

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

I created a `JUnit5` test that executes the steps I've described in the previous section and that reproduces the behavior. When targeting `AWS` the test passes, when targeting `LocalStack` it doesn't. All results are consistent throughout many test runs.

```java

@Test
void receivesTwoBatchesFromSameMessageGroup() {
	String messageGroupId = UUID.randomUUID().toString();
	String queueName = "my-fifo-queue.fifo";
	logger.info("Create queue {}", queueName);
	String queueUrl = sqsAsyncClient.createQueue(CreateQueueRequest.builder()
			.queueName(queueName)
			.attributes(Collections.singletonMap(QueueAttributeName.FIFO_QUEUE, "true"))
			.build())
		.join()
		.queueUrl();
	logger.info("Send 10 messages to url {} for message group {}", queueUrl, messageGroupId);
[...]

PR #8146, opened 2023-04-14
SQS fix deduplication at queue level and deleted messages

bentsku:
While working on #8144 and #8145 I've spotted 2 issues with deduplication on FIFO SQS queues.

First, it seemed we did not deduplicate deleted messages when the SQS documentation says they do.
See https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html

While reading this, I've also realised that what I thought was a feature of SQS was actually a bug, and that we should deduplicate on the queue level and not the `MessageGroupId`. However, this means breaking persistence as we're changing the structure of the `FifoQueue` object. Is this acceptable for 2.1?

I've added 2 AWS validated tests, and #8144 is building on this PR.

### DIFF
--- a/tests/integration/test_sns.py
+++ b/tests/integration/test_sns.py
@@ -2220,7 +2220,7 @@ class TestSNSProvider:
             _get_snapshot_requests_response(broken_token_confirm_subscribe_request),
         )
 
-        # a topic with a different region will fail than the subscription
+        # using the right topic name with a different region will fail when confirming the subscription
         parsed_arn = parse_arn(topic_arn)
         different_region = "eu-central-1" if parsed_arn["region"] != "eu-central-1" else "us-east-1"
         different_region_topic = topic_arn.replace(parsed_arn["region"], different_region)
--- a/tests/integration/test_sqs.py
+++ b/tests/integration/test_sqs.py
@@ -2108,6 +2108,50 @@ class TestSqsProvider:
         snapshot.match("get-messages-duplicate", response)
 
     @pytest.mark.aws_validated
+    @pytest.mark.parametrize("content_based_deduplication", [True, False])
+    def test_fifo_deduplication_not_on_message_group_id(
+        self,
+        sqs_create_queue,
+        aws_client,
+        content_based_deduplication,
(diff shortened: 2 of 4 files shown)


thrau (approved):
LGTM! thanks, fifo dedup could use some overhaul... we also still have #6766 to tackle..

thanks for adding tests!

tests/integration/test_sns.py
```diff
@@ -2179,9 +2199,57 @@ def test_subscribe_external_http_endpoint(
...
+        test_broken_token_confirm_url = (
+            f"{service_url}/?Action=ConfirmSubscription&TopicArn={topic_arn}&Token=abc123"
+        )
+        broken_token_confirm_subscribe_request = requests.get(test_broken_token_confirm_url)
+        snapshot.match(
+            "broken-token-confirm",
+            _get_snapshot_requests_response(broken_token_confirm_subscribe_request),
         )
 
+        # a topic with a different region will fail than the subscription
```
baermat, line 2223: nit; I am not quite sure what this means, please clarify :) (possibly typo, did you mean *will then fail the subscription*?)

tests/integration/test_sqs.py
```diff
@@ -2020,6 +2061,96 @@ def test_fifo_content_based_message_deduplication_arrives_once(
...
+        aws_client.sqs.delete_message(
+            QueueUrl=queue_url, ReceiptHandle=response["Messages"][0]["ReceiptHandle"]
+        )
+
+        # republish the same message
+        aws_client.sqs.send_message(
+            QueueUrl=queue_url, MessageBody=item, MessageGroupId=group, **kwargs
+        )
+
+        # second doesn't receive anything since the message has the same content
```
baermat, line 2105: nit: since we parametrize content_based_deduplication, I'd say that the "second doesn't receive anything because the deduplication id is the same" (which is derived from the content in on of the cases, but not in the other)

baermat (approved):
Small nits but LGTM, and sorry for the delay!

PR #8147, opened 2023-04-14
fix IAM client fixture

bentsku:
#8074 just reintroduced usage of the old client fixture, so fixing this to unblock the pipeline.


pinzon (approved):
Thanks 👍

Issue #8139, opened 2023-04-13
bug: S3 automatically decodes gzip requests (it should leave that to the client)


kyllerss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am storing a gzipped file in an S3 bucket. When retrieving the contents, I expect to have to decode the response. However, once I updated to `2.0.1` I am finding that the response already has the gzipped content decoded. 

The following is the response when localstack is run in `legacy` mode. Note the payload reflects the uncompressed contents and the `Content-Length` header reflects the length of the compressed content (ie. the length before it is decompressed).

```bash
$> wget -q -S -O - https://originaluploads.dev.mydomain.com/2023/4/13/79012447-af8a-4732-8595-b378d9ddb8ac/original_upload.gpx
  HTTP/1.1 200 
  Server: nginx/1.22.1
  Date: Thu, 13 Apr 2023 21:35:21 GMT
  Content-Type: application/gpx+xml
  Content-Length: 484
  Connection: keep-alive
  x-amz-version-id: 721618c3-d9d2-42bd-b433-c0530c4e3a09
  cache-control: no-cache
  content-encoding: gzip
  ETag: "1d1f83feb12d0f0185a12bb7909b3955"
  last-modified: Thu, 13 Apr 2023 21:34:31 GMT
  x-amz-server-side-encryption: AES256
  x-amz-server-side-encryption-bucket-key-enabled: false
  x-amzn-requestid: zeRF2ZIh8iI5WgtR21aoNDIMAgxatwkKndFENfFQc1dfl7tqffx4
  Access-Control-Allow-Credentials: true
  x-amz-request-id: 5BC4A0ADB1AC8913
  x-amz-id-2: MzRISOwyjmnup5BC4A0ADB1AC89137/JypPGXLh0OVFGcJaaO3KW/hRAqKOpIEEp
  accept-ranges: bytes
  content-language: en-US
gu8d͕ˎ�0��y%����\����AZm�5KX�^
...
$>
```

The following is what I get when I disable the legacy mode and use 2.0.1's new S3 provider. I have uploaded the same file (even though the file path is different in this second run). Note how that the payload is now the uncompressed file and the `Content-Length` value reflects the uncompressed file content length:
 
```bash
$> wget -q -S -O - https://originaluploads.dev.mydomain.com/2023/4/13/e01bcd66-f84b-44af-b4ba-5f6819a22ff2/original_upload.gpx
  HTTP/1.1 200 
[...]

PR #8148, opened 2023-04-14
fix auto decoding of gzip for s3 vhost proxied requests

bentsku:
As reported in #8139, the new S3 provider will decode gzip encoded requests before returning the data.
This was due to our Proxy and more specifically our `SimpleHttpClient`, which uses `requests` under the hood. Requests automatically decode the response, and we need the raw data underneath. I've duplicated our http client to not do a drastic change and only use it in the virtual host part of S3, we can then see if it creates issues or not 😄

I didn't add unit tests as I believe it could be added on top of #8104, and don't want to create too many conflicts. 

About requests and decoding gzip:
https://docs.python-requests.org/en/latest/community/faq/#encoded-data

_fixes #8139_

### DIFF
--- a/localstack/http/client.py
+++ b/localstack/http/client.py
@@ -47,6 +47,19 @@ class SimpleRequestsClient(HttpClient):
     def __init__(self, session: requests.Session = None):
         self.session = session or requests.Session()
 
+    @staticmethod
+    def _get_destination_url(request: Request, server: str | None = None) -> str:
+        if server:
+            # accepts "http://localhost:5000" or "localhost:5000"
+            if "://" in server:
+                parts = urlparse(server)
+                scheme, server = parts.scheme, parts.netloc
+            else:
+                scheme = request.scheme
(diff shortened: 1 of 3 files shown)


thrau (approved):
LGTM! I appreciate you tried to leave the base implementation unmodified! It seems we should probably add this as a transparent feature to the default client though. Definitely not part of this PR though :-)

:shipit:

bentsku:
> LGTM! I appreciate you tried to leave the base implementation unmodified! It seems we should probably add this as a transparent feature to the default client though. Definitely not part of this PR though :-)
> 
> :shipit:

Thanks! Yes I am a bit wary when modifying a central piece like this one, and thought we could try it on S3 first. I guess it's also a nice part in our new push towards streaming responses, and it's good to see it seems to work this way! I'll merge after `2.0.2` :shipit: but agreed about merging this in one implementation, maybe with the simple keyword `stream` like `requests`, or just by default. We will see then!

PR #8151, opened 2023-04-17
Always return bridge as fallback network

dominikschubert:
In some cases the returned network was actually None here, which wasn't really handled anywhere downstream and was also in conflict with the method description. Returning and trying to use the `bridge` network here instead should still be a more sensible default.

## Changes
- `get_main_container_network` now also returns "bridge" when `DOCKER_CLIENT.get_networks` fails


dfangl (approved):
This makes sense, thanks!

PR #8033, opened 2023-03-31
[SFN] Failure support for aws-sdk

MEPalma:
- adds support for States.TaskFailed error handling in aws-sdk task states
- fixes parsing of empty parameters declarations
- fixes assumption of iam roles for snapshot tests
- fixes event reporting for task workflows
- adds error events
- fixes encoding of environment input dumping
- adds small snapshot tests for aws-sdk failing tasks
- fixes the handling of States.Runtime errors,  which cannot be caught

PR #8158, opened 2023-04-18
[SFN] Failure support for lambda (service tasks)

MEPalma:
\* Extends https://github.com/localstack/localstack/pull/8033

This pr:
- Restructures `service` `state tasks`, with `lambda service state task` and `lambda state task` now benefitting of this change
- Fixes issues for which runtime failures not always lead to execution failure state events being logged
- Fixed issue with result path not updating its target by value
- Fixes relationship between `catch declaration`s and `catchers`
- Fixes `States.TaskFailed` matching task failure (see `todo` for further investigations)
- Adds custom exception decorators for lambda errors
- Adds baseline snapshot tests for lambda exceptions handling in catch declarations


tests/integration/stepfunctions/v2/error_handling/utils.py
```diff
@@ -0,0 +1,41 @@
+from localstack.testing.snapshots.transformer import RegexTransformer
+from localstack.utils.strings import short_uid
+from tests.integration.stepfunctions.utils import await_execution_success
+
+
+@staticmethod
+def _test_sfn_scenario(
```
dominikschubert, line 7:
```suggestion
def base_test_sfn_scenario(
```
nit: staticmethod isn't needed here and `_` should be reserved to denote "private" access in a module or class.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_lambda.py
```diff
@@ -0,0 +1,56 @@
...
+        if "taskFailedEventDetails" in event_details:
+            task_failed_details: TaskStartedEventDetails = event_details["taskFailedEventDetails"]
+            task_failed_details["resourceType"] = self._get_resource_type()
+            task_failed_details["resource"] = self.resource.api_action
+        return failure_event
+
+    def _eval_execution(self, env: Environment) -> None:
+        parameters = self._eval_parameters(env=env)
+        parameters_str = to_json_str(parameters)
+        env.event_history.add_event(
```
dominikschubert, line 36: The `HistoryEventType.TaskScheduled` might imply some changes needed in the future for the internals here where this isn't actually executed immediately but scheduled instead and might have to wait for resources to be available :thinking:

dominikschubert (approved):
LGTM!

PR #8159, opened 2023-04-18
Fix firehose / s3 client creation of clients with role

dfangl:
## Motivation
Currently, we completely ignore the roles provided with the sns subscription or firehose s3 records integration.

## Changes
This PR uses the roles where provided. If it is not provided, we will just fallback to the service principal. This is to remain non-breaking, but we need to increase parity for this at a later point.

### DIFF
--- a/localstack/services/cloudformation/engine/entities.py
+++ b/localstack/services/cloudformation/engine/entities.py
@@ -437,7 +437,10 @@ def resolve_ssm_parameter_value(parameter_type: str, parameter_value: str) -> st
     function will resolve the SSM parameter with name `test-param` and return the SSM parameter's value.
     """
     # TODO: support different parameter value types
-    if parameter_type == "AWS::SSM::Parameter::Value<String>":
+    if (
+        parameter_type == "AWS::SSM::Parameter::Value<String>"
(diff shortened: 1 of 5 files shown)


bentsku (approved):
LGTM! 👍
only a quick question: when using this `with_assumed_role`, what happens if the role is badly formatted? Because I don't think we have validation concerning these attributes.

Also, curiosity question: in AWS, if the role does not have the right to publish to firehose, where would that message would go, just not be sent? I guess right now it would fail and log something, but if we were to implement DLQ behaviour (is there even for Firehose, I have no idea, I guess), then it would go there except if we catch some authorisation exception?
I'll have to take a look someday about this! 😄

dfangl:
Great points! If the role is badly formatted, it will fail. If the role just does not exist, it will use some credentials which won't be enforced.

For the second question: This is something we are testing in -ext. It depends on the service, some will check on resource creation if the necessary permissions are there (e.g. lambda), but if you for example remove the permissions afterwards, it will just not work.
For SNS, it will just try to do it and not work otherwise, potentially putting it in the DLQ (if it has the permissions for that ;)).

PR #8160, opened 2023-04-18
fix default ProvisionedThroughput in DynamoDB

pinzon:
Conditional attributes are not completely/correctly resolved during the execution of `add_defaults` method of a GenericBaeseModel. The current patch allows to have a DynamoDB table that has multiple conditional attributes to be deployed.

There's more work required to completely fix the issue but with these small changes the amplify-localstack plugin can finally deploy a API stack to LocalStack.

Changes:
- [x] Test with a small example of what amplify tries to deploy.
- [x] Move the addition of default parameters to another phase where conditionals are resolved.

### DIFF
--- a/tests/integration/cloudformation/resources/test_dynamodb.py
+++ b/tests/integration/cloudformation/resources/test_dynamodb.py
@@ -81,6 +81,9 @@ def test_default_name_for_table(deploy_cfn_template, snapshot, aws_client):
         "$..Table.ProvisionedThroughput.LastDecreaseDateTime",
         "$..Table.ProvisionedThroughput.LastIncreaseDateTime",
         "$..Table.Replicas",
+        "$..Table.DeletionProtectionEnabled",
+        "$..Table.LatestStreamArn",
+        "$..Table.LatestStreamLabel",
     ]
 )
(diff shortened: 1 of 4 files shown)


tests/integration/cloudformation/resources/test_dynamodb.py
```diff
@@ -73,3 +73,29 @@ def test_default_name_for_table(deploy_cfn_template, snapshot, aws_client):
...
+        "$..Table.ProvisionedThroughput.LastDecreaseDateTime",
+        "$..Table.ProvisionedThroughput.LastIncreaseDateTime",
+        "$..Table.Replicas",
+        "$..Table.DeletionProtectionEnabled",
+        "$..Table.LatestStreamArn",
+        "$..Table.LatestStreamLabel",
+    ]
+)
+@pytest.mark.parametrize("billing_mode", ["PROVISIONED", "PAY_PER_REQUEST"])
+def test_billing_mode_as_conditional(deploy_cfn_template, snapshot, aws_client, billing_mode):
```
dominikschubert, line 90: nit: Would be great to have a short comment here explaining the context of why this test was added

dominikschubert (approved):
Let's unblock amplify :rocket: 

Thanks for adding the integration test as well :+1:

PR #8161, opened 2023-04-18
Add EC2 image ssm parameter type

dfangl:
## Motivation
Any deployment referencing this parameter type fails currently

## Changes
Add `AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>` parameter type and use ssm to retrieve it (it is stored in SSM).


dominikschubert (approved):
LGTM :ship:

Issue #8156, opened 2023-04-18
bug: terraform aws_acm_certificate data source runs indefinitely


tiagoposse:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When attempting to retrieve an ACM certificate using the terraform data source "aws_acm_certificate", the certificate is never read and the call runs indefinitely until cancelled.

### Expected Behavior

The certificate is retrieved and the terraform plan executes.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack-pro:2.0.1 ...

#### Terraform code

The following terraform module is sufficient to reproduce this issue:
```
resource "aws_route53_zone" "this" {
  name = "example.com"
}

module "acm" {
  source  = "terraform-aws-modules/acm/aws"
  version = "4.3.2"

  domain_name         = "*.example.com"
  wait_for_validation = true
  zone_id             = aws_route53_zone.this.zone_id
}

data "aws_acm_certificate" "this" {
  depends_on = [module.acm]

  domain = "*.example.com"
}
```

### Environment

```markdown
- Docker
- LocalStack: 2.0.1
```

### Anything else?

[terraform_exec.txt](https://github.com/localstack/localstack/files/11259059/terraform_exec.txt)
[localstack_log.txt](https://github.com/localstack/localstack/files/11259060/localstack_log.txt)

whummer:
Thanks for reporting @tiagoposse . We have now pushed a few enhancements in #8162 which should fix the Terraform deployment of ACM certificates. Can you please pull the `latest` Docker image and give it another try? Please keep us posted on how it goes.. Thanks!

tiagoposse:
Unfortunately the "latest" image returns
```
LocalStack supervisor: localstack process (PID 16) returned with exit code -11
LocalStack supervisor: exiting
```
when running this example

whummer:
Thanks for the update @tiagoposse . Exit code -11 usually means "segmentation fault", which indicates a fatal error in the execution, usually caused by an invalid memory access.

This type of error is usually quite specific to the environment you're running in (operating system, CPU architecture, Docker version, etc). Can you please share a few more details about your environment? 

Also, are you seeing any more details in the logs, perhaps with `DEBUG=1` or `LS_LOG=trace` (for detailed trace logging) enabled - can you please share the logs with these configs enabled? Thanks!

[...]

PR #8162, opened 2023-04-18
Enhance parity for ACM certificates for wildcard subdomains

whummer:
Enhance parity for ACM certificates issued for wildcard subdomains . Addresses https://github.com/localstack/localstack/issues/8156

With the changes in this PR, the following TF script deploys successfully (confirmed that it was hanging indefinitely before):
```
resource "aws_route53_zone" "this" {
  name = "example.com"
}

module "acm" {
  source  = "terraform-aws-modules/acm/aws"
  version = "4.3.2"

  domain_name         = "*.example.com"
  wait_for_validation = true
  zone_id             = aws_route53_zone.this.zone_id
}

data "aws_acm_certificate" "this" {
  depends_on = [module.acm]

  domain = "*.example.com"
}
```

The changes are covered by a snapshot test. Overall, the amount of patching we're performing in ACM is becoming a bit ridiculous ;) - probably a candidate for decoupling from moto in the near future as well.

### DIFF
--- a/localstack/services/acm/provider.py
+++ b/localstack/services/acm/provider.py
@@ -1,22 +1,42 @@
+from moto import settings as moto_settings
 from moto.acm import models as acm_models
 
 from localstack.aws.api import RequestContext, handler
-from localstack.aws.api.acm import AcmApi, RequestCertificateRequest, RequestCertificateResponse
+from localstack.aws.api.acm import (
+    AcmApi,
+    ListCertificatesRequest,
+    ListCertificatesResponse,
+    RequestCertificateRequest,
+    RequestCertificateResponse,
+)
 from localstack.services import moto
 from localstack.utils.objects import singleton_factory
 from localstack.utils.patch import patch
 
+# reduce the validation wait time from 60 (default) to 10 seconds
+moto_settings.ACM_VALIDATION_WAIT = 10
+
 
 @patch(acm_models.CertBundle.describe)
 def describe(describe_orig, self):
-    result = describe_orig(self)
+
+    # Terrible hack (for parity). Moto adds certain required fields only if status is PENDING_VALIDATION. TODO fix!
+    cert_status = self.status
+    self.status = "PENDING_VALIDATION"
+    try:
+        result = describe_orig(self)
+    finally:
+        self.status = cert_status
+
     cert = result.get("Certificate", {})
-    sans = cert.get("SubjectAlternativeNames", [])
+    cert["Status"] = cert_status
+    sans = cert.setdefault("SubjectAlternativeNames", [])
+    sans_summaries = cert.setdefault("SubjectAlternativeNameSummaries", [])
 
     # add missing attributes in ACM certs that cause Terraform to fail
     addenda = {
         "RenewalEligibility": "INELIGIBLE",
-        "KeyUsages": [{"Name": "DIGITAL_SIGNATURE"}],
+        "KeyUsages": [{"Name": "DIGITAL_SIGNATURE"}, {"Name": "KEY_ENCIPHERMENT"}],
         "ExtendedKeyUsages": [],
         "Options": {"CertificateTransparencyLoggingPreference": "ENABLED"},
     }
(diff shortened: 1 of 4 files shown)


tests/integration/test_acm.py
```diff
@@ -53,12 +57,70 @@ def test_import_certificate(self, aws_client):
...
+        dns_options = response["Certificate"]["DomainValidationOptions"][0]["ResourceRecord"]
+        snapshot.add_transformer(
+            snapshot.transform.regex(dns_options["Name"].split(".")[0], "<record-prefix>")
+        )
+        snapshot.add_transformer(snapshot.transform.regex(dns_options["Value"], "<record-value>"))
+        snapshot.match("describe-cert", response)
+
+        if is_aws_cloud():
+            # wait until DNS entry has been added (needs to be done manually!)
+            input()
```
alexrashed, line 111: I think this might be a bit dangerous. If we (sometime in the future) want to run the AWS tests automatically to re-verify, would this block or fail (due to a missing tty)?
What exactly needs to be done here?
/cc @dominikschubert

localstack/services/acm/provider.py
```diff
@@ -1,22 +1,42 @@
...
 
+# reduce the validation wait time from 60 (default) to 10 seconds
+moto_settings.ACM_VALIDATION_WAIT = min(10, moto_settings.ACM_VALIDATION_WAIT)
+
 
 @patch(acm_models.CertBundle.describe)
 def describe(describe_orig, self):
-    result = describe_orig(self)
+
+    # Terrible hack (for parity). Moto adds certain required fields only if status is PENDING_VALIDATION. TODO fix!
```
alexrashed, line 23: Maybe we could add the TODO to the beginning of the sentence to enable IDE highlighting?

And this basically means that we should do an upstream fix to also add these mandatory fields even if the current status is not `PENDING_VALIDATION`, in the following section:
https://github.com/localstack/moto/blob/576c9122ff9b25548d0dca0cf22d737ef681b3ff/moto/acm/models.py#L343-L364

In fact, it seems like we could provide the whole patch to upstream (since it does not depend on any `localstack-ext` internals), however, this is a difficult question (when it comes to the coupling, and the reasoning for ACM in ext).

localstack/services/acm/provider.py
```diff
@@ -33,26 +53,62 @@ def describe(describe_orig, self):
...
+        cert["HasAdditionalSubjectAlternativeNames"] = False
+
+    if not cert.get("ExtendedKeyUsages"):
+        cert["ExtendedKeyUsages"] = [
+            {"Name": "TLS_WEB_SERVER_AUTHENTICATION", "OID": "1.3.6.1.0.1.2.3.0"},
+            {"Name": "TLS_WEB_CLIENT_AUTHENTICATION", "OID": "1.3.6.1.0.1.2.3.4"},
+        ]
+
+    # remove attributes prior to validation
+    attrs = ["CertificateAuthorityArn", "IssuedAt", "NotAfter", "NotBefore", "Serial"]
```
alexrashed, line 105: I guess this line could move two lines below.

localstack/services/acm/provider.py
```diff
@@ -33,26 +53,62 @@ def describe(describe_orig, self):
...
+    if not cert.get("Status") == "ISSUED":
+        for attr in attrs:
+            cert.pop(attr, None)
+        cert["KeyUsages"] = []
+        cert["ExtendedKeyUsages"] = []
```
alexrashed, line 110: Here the defaults created right above are removed again, if it is not yet `ISSUED`.
This is a bit confusing, maybe we could (in a later iteration) try to change this such that the defaults are added depending on the status (analogous to the patched function).

alexrashed (approved):
And another service with snapshot tests being introduced, that's awesome!
I only have some smaller nitpicks or somewhat fundamental questions (which aren't to be solved / answered within the scope of this PR), besides that it looks great! 🚀

tests/integration/test_acm.py
whummer: True, 👍  should have been a bit more specific. What needs to be done here is essentially to go to our DNS config (on gandi.net), and add a CNAME record that AWS generates and requires for domain verification (the CNAME record can be retrieved from the AWS console, or is also present in the `DomainValidationOptions.ResourceRecord` in the DescribeCertificate API response).

Good point about tests requiring TTY - at least we could introduce a test marker that indicates that this test requires manual intervention (which, essentially, will be the case for probably all ACM integration tests if we want to parity test them).

localstack/services/acm/provider.py
whummer: Yep, good point - this is indeed a bit roundabout. Looking forward to streamlining this a bit more in the next iteration 👍

localstack/services/acm/provider.py
whummer: Moved the TODO comment 👍  

Good point regarding the upstreaming. In fact, I'd love to discuss with you a migration of the entire logic into LocalStack. The moto ACM core implementation (models.py) has ~550 lines of code (which is not a lot), and our own provider is already at ~180 LOC. Seems like this could be a good case for a migration/re-implementation, and then we can add more proper snapshot tests and get rid of the patches.. 🚀

PR #8177, opened 2023-04-21
fix tag updates of AWS::SSM::Parameter

dominikschubert:
This issue was encountered by a user when updating an `AWS::SSM::Parameter` resource which included a tag and resulted in the following exception:

```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the PutParameter operation: Invalid request: tags and overwrite can't be used together. To create a parameter with tags, please remove overwrite flag. To update tags for an existing parameter, please use AddTagsToResource or RemoveTagsFromResource.
```

### DIFF
--- a/localstack/services/cloudformation/models/ssm.py
+++ b/localstack/services/cloudformation/models/ssm.py
@@ -1,3 +1,4 @@
+from localstack.aws.connect import connect_to
 from localstack.services.cloudformation.deployment_utils import (
     merge_parameters,
     params_dict_to_list,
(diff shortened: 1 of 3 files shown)


localstack/services/cloudformation/models/ssm.py
```diff
@@ -42,17 +43,43 @@ def update_resource(self, new_resource, stack_name, resources):
...
+                    {"Key": tag_key, "Value": tag_value}
+                    for tag_key, tag_value in new_tags.items()
+                    if tag_key in tag_keys_to_add
+                ],
+            )
+
+        if tag_keys_to_remove:
+            ssm_client.remove_tags_from_resource(
+                ResourceType="Parameter", ResourceId=self.props.get("Name"), Tags=tag_keys_to_remove
+            )
```
dominikschubert, line 80: this part isn't actually working yet because the engine passes the old tag state for some reason when updating :shrug: 

Will be fixed in the course of the next CFn core projects

whummer (approved):
Looks great, thanks for quickly jumping on this issue @dominikschubert ! 🚀 

> Will be fixed in the course of the next CFn core projects

Indeed - looking forward to fixing the state passing in the engine 🎉

pinzon (approved):

PR #8179, opened 2023-04-21
Fix publishing versions if package type is image

dfangl:
## Motivation

Publishing image versions from image lambdas is currently broken, as we access the wrong code object.

## Changes
* Fix access to codesha256
* Add test for versioning image lambdas
* Add dependency of `test_image` fixture to `login_docker_client` fixture, to correctly login (login fixture was previously unused, it seems)

### DIFF
--- a/localstack/services/awslambda/provider.py
+++ b/localstack/services/awslambda/provider.py
@@ -385,16 +385,23 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                 "The Revision Id provided does not match the latest Revision Id. Call the GetFunction/GetAlias API to retrieve the latest Revision Id",
                 Type="User",
             )
-        current_hash = current_latest_version.config.code.code_sha256
-        if (
-            code_sha256
-            and current_hash != code_sha256
-            and not current_latest_version.config.code.is_hot_reloading()
-        ):
+
+        # check if code hashes match if they are specified
+        current_hash = (
+            current_latest_version.config.code.code_sha256
+            if current_latest_version.config.package_type == PackageType.Zip
+            else current_latest_version.config.image.code_sha256
+        )
+        is_hot_reloading = (
+            current_latest_version.config.package_type == PackageType.Zip
+            and current_latest_version.config.code.is_hot_reloading()
+        )
+        if code_sha256 and current_hash != code_sha256 and not is_hot_reloading:
             raise InvalidParameterValueException(
                 f"CodeSHA256 ({code_sha256}) is different from current CodeSHA256 in $LATEST ({current_hash}). Please try again with the CodeSHA256 in $LATEST.",
                 Type="User",
             )
+
         state = lambda_stores[account_id][region]
         function = state.functions.get(function_name)
         changes = {}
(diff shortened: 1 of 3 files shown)


localstack/services/awslambda/provider.py
```diff
@@ -385,16 +385,23 @@ def _create_version_model(
...
-        ):
+
+        # check if code hashes match if they are specified
+        current_hash = (
+            current_latest_version.config.code.code_sha256
+            if current_latest_version.config.package_type == PackageType.Zip
+            else current_latest_version.config.image.code_sha256
+        )
+        is_hot_reloading = (
+            current_latest_version.config.package_type == PackageType.Zip
```
joe4dev, line 396: nit: the nested hot reloading check is a bit cumbersome but that's because it only applies to code package_type

tests/integration/awslambda/test_lambda_api.py
```diff
@@ -794,6 +794,75 @@ def test_lambda_image_and_image_config_crud(
...
+        create_image_response = create_lambda_function_aws(
+            FunctionName=function_name,
+            Role=lambda_su_role,
+            Code={"ImageUri": image},
+            PackageType="Image",
+            Environment={"Variables": {"CUSTOM_ENV": "test"}},
+            Publish=True,
+        )
+        snapshot.match("create-image-response", create_image_response)
+        aws_client.awslambda.get_waiter("function_active_v2").wait(FunctionName=function_name)
```
joe4dev, line 815: The waiter is not strictly necessary because `create_lambda_function_aws` waits but it cannot harm.

sidenote: We also have two similar fixtures `create_lambda_function_aws` and `create_lambda_function` without any comment how they differ.

tests/integration/awslambda/test_lambda_api.py
```diff
@@ -794,6 +794,75 @@ def test_lambda_image_and_image_config_crud(
...
+        function_name = f"test-function-{short_uid()}"
+        create_image_response = create_lambda_function_aws(
+            FunctionName=function_name,
+            Role=lambda_su_role,
+            Code={"ImageUri": image},
+            PackageType="Image",
+            Environment={"Variables": {"CUSTOM_ENV": "test"}},
+            Publish=True,
+        )
+        snapshot.match("create-image-response", create_image_response)
```
joe4dev, line 814: nit: any reason for alternating `-` and `_` 🤔

joe4dev (approved):
LGTM 👍 

✅ test passes on ARM mac

tests/integration/awslambda/test_lambda_api.py
dfangl: Yes. I copied the first part of the test from one test, the second part of it from another, and then modified it, hence the different styles. Will fix!

tests/integration/awslambda/test_lambda_api.py
dfangl: That's also copied from another test, you are right!

dfangl: For the fixtures: The first one is basically a wrapper around the boto call, with a waiter and desctruction. The second one is made to be compatible with the testutil.create_lambda_function method, as a drop in replacement.

dominikschubert (approved):

PR #5298, opened 2022-01-18
fix ASF xml response encoding

thrau:
attempts a fix for #5297 by

* escaping strings using HTML encoding
* adding XML declarations (not complete yet, just for errors)
* adds Content-Encoding: text/xml headers

PR #8180, opened 2023-04-21
fix invalid SQS message decoding

### LINKED ISSUE #8002: bug: &quot; when sending replaced by quotes when receiving

### Current Behavior

Sending a message with body:
`"&quot;&quot;"`
is received on the consumer side with body:
`""""`

### Expected Behavior

Sending a message with body:
"&quot;&quot;"
is received on the consumer side with body:
"&quot;&quot;"

Custom (please describe below)

Running localstack in docker desktop on Mac using helm chart

https://localstack.github.io/helm-charts
version 0.3.7
Image version: 0.14.3

### Environment

```markdown
- OS: macOS Monterey
- LocalStack: 0.14.3
```

### Anything else?

_No response_

alexrashed:
This PR slightly adjusts the special message encoding for SQS which has been introduced with #5298.

With #5298, an SQS specific serializer has been introduced which performs a special escaping logic for the node text content in the XML responses. This is only done by SQS, and since it's a very rare practice, it's not supported by any popular XML serialization libraries.

#8002 raised the issue that this behavior causes issues when these escaping sequences are used in the sent message itself.
A message `"&quot;&quot;"` sent would be received as `""""`.

This was due to the nature of the initial implementation:
- These special characters have been escaped before the ElementTree serialization was performed: `"&quot;&quot;"` ==> `&quot;&quot;&quot;&quot;`
- The escape sequence itself contains characters which are escaped in XML (specifically `&`): `&quot;` ==> `&amp;quot;&amp;quot;&amp;quot;&amp;quot;`
- A post-serialization step removed the double-escaping: `&amp;quot;&amp;quot;&amp;quot;&amp;quot;` ==> `&quot;&quot;&quot;&quot;`
- The client performs the unescape: `&quot;&quot;&quot;&quot;` ==> `""""`

With this PR, the special escaping is done after the XML node is serialized:
- The special characters in the text are "marked" (which is just another form of internal escaping): `"&quot;&quot;"` ==> `__marker__"__marker__&quot;&quot;__marker__"__marker__`
  - The "marking" is necessary to identify exactly those special characters within the completely serialized message, without taking any other occurrences of these special characters into account.
- The message is serialized by the XML serializer: `__marker__"__marker__&quot;&quot;__marker__"__marker__` ==> `__marker__"__marker__&amp;quot;&amp;quot;__marker__"__marker__`
- A post-serialization step replaces the marked special characters with their escape sequence: `__marker__"__marker__&amp;quot;&amp;quot;__marker__"__marker__` ==> `&quot;&amp;quot;&amp;quot;&quot;`
- The client performs the unescape: `&quot;&amp;quot;&amp;quot;&quot;` ==> `"&quot;&quot;"`

This approach now would not work if the "marked" / internally escaped representation of the character is used in the raw message. f.e. a message `"__marker__"__marker"` would be received as `"""`.
However, the only approach I can think of right now to properly implement this would be to avoid the specific pre- and post-processing and this could only be done by fixing the serialization itself (i.e. implementing our own XML serialization).
I'm happy for any feedback or ideas!

Fixes #8002.

### DIFF
--- a/localstack/aws/protocol/serializer.py
+++ b/localstack/aws/protocol/serializer.py
@@ -1513,23 +1513,28 @@ class SqsResponseSerializer(QueryResponseSerializer):
     """
 
     def _default_serialize(self, xmlnode: ETree.Element, params: str, _, name: str, __) -> None:
-        """Ensures that XML text nodes use HTML entities instead of " or \r"""
+        """
+        Ensures that we "mark" characters in the node's text which need to be specifically encoded.
+        This is necessary to easily identify these specific characters later, after the standard XML serialization is
+        done, while not replacing any other occurrences of these characters which might appear in the serialized string.
+        """
         node = ETree.SubElement(xmlnode, name)
-        node.text = str(params).replace('"', "&quot;").replace("\r", "&#xD;")
+        node.text = (
+            str(params)
+            .replace('"', '__marker__"__marker__')
+            .replace("\r", "__marker__\r__marker__")
+        )
 
     def _node_to_string(self, root: Optional[ETree.ElementTree], mime_type: str) -> Optional[str]:
-        """
-        Replaces the double-escaped HTML entities with their correct HTML entity (basically reverts the escaping in
-        the serialization of the used XML framework).
-        """
+        """Replaces the previously "marked" characters with their encoded value."""
         generated_string = super()._node_to_string(root, mime_type)
         return (
             to_bytes(
                 to_str(generated_string)
                 # Undo the second escaping of the &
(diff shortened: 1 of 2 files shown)


thrau (approved):
This is both practical and ridiculous. love it :100: 

In seriousness though: I think this is the solution we had once discussed as a potential workaround and it seems fine for now. I also don't see a way around something like this right now without putting significant effort into our own serialization approach.

so LGTM :+1: 

my only comment would be to perhaps add a few more tests with different variants of strings so we're guarded against regressions, and that can maybe also document limitations. but not necessary in this PR imo :+1:

PR #7396, opened 2022-12-28
add in-memory ASF request dispatching for boto clients

thrau:
This PR adds code to make calls from boto clients directly to a Gateway without an HTTP server roundtrip. This can make internal calls much faster because they don't require IO.

The mechanism used to achieve this is botocore's `before-send` event, which can be used to short-circuit the request logic by returning a response from the event handler, before the HTTP request is even made.

I did some basic refactoring to make this easier, including:
* created `localstack.runtime.components` as a place for runtime-level singleton components
* added a new method to `Gateway` with the signature `def handle(self, context: RequestContext, response: Response)`, which now makes the Gateway a drop-in replacement for a handler chain (and easier to call if you want to invoke it with a pre-existing RequestContext).

### TODO:
- [X] basic PoC implementation
- [ ] integration with the client factory and `connect_to`
- [ ] more benchmarks to see whether/where this is really useful
- [ ] more unit tests for different protocols
- [ ] documentation

PR #8181, opened 2023-04-21
add http client for in-memory gateway request dispatching

thrau:
:construction: similar to #7396, this adds an in-memory request dispatching mechanism. it has a similar but slightly different approach to organizing the runtime components into the container. just playing around.

### DIFF
--- a/localstack/aws/client.py
+++ b/localstack/aws/client.py
@@ -6,9 +6,11 @@ from typing import Dict, Iterable, Optional
 
 from botocore.model import OperationModel
 from botocore.parsers import ResponseParser, ResponseParserFactory
-from werkzeug import Response
+from werkzeug import Request, Response
 
 from localstack.aws.api import CommonServiceException, ServiceException, ServiceResponse
+from localstack.aws.gateway import Gateway
+from localstack.http.client import HttpClient
 from localstack.runtime import hooks
 from localstack.utils.patch import patch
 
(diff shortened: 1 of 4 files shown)


alexrashed:
@thrau the botocore in-memory request dispatching was implemented with https://github.com/localstack/localstack/pull/9569. Do you think we should pick this one up again, and also implement a similar thing for internal HTTP requests?

thrau:
now with `get_current_runtime()` and the Components class, it's a lot easier to instantiate a client like this somewhere in the ls runtime. For example:

```python
from rolo import Request, Response
from rolo.client import HttpClient
from rolo.gateway import Gateway

from localstack.runtime import get_current_runtime


class GatewayClient(HttpClient):
    """
    HttpClient implementation that injects the request directly into the Gateway.
    """

    gateway: Gateway

    def __init__(self, gateway: Gateway):
        self.gateway = gateway

    def request(self, request: Request, server: str | None = None) -> Response:
        response = Response()
        self.gateway.process(request, response)
        return response

    def close(self):
        pass


runtime = get_current_runtime()
client = GatewayClient(runtime.components.gateway)

client.request(Request("GET", "/_localstack/health"))
```

do you think this type of client would be useful somewhere in ls, for proxying or internal dispatching @bentsku ?

thrau:
after discussing with ben, i decided to toss this PR. it seems to have limited usefulness, and if someone needs it, it's trivial to implement now (see comment above)

PR #8190, opened 2023-04-24
reintroduce vhost use in s3 cors tests

bentsku:
Remove comments and reintroduce the usage of virtual host addressing in S3 CORS tests. These were in place because at some point, it seemed like it reduced test failures that were finally fixed with #8134

Also remove some commented out code.


thrau (approved):
cool, thanks @bentsku!

Issue #8192, opened 2023-04-25
bug: Different dynamo behavior in newer localstack (1.4/2.0) vs older localstack (1.2)


adamdavis40208:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using localstack:1.2, we can create dynamo tables and quickly use them. Without a wait. We can also list tables with aws cli and see the table we created.

We recently tried updating our images to 1.4/2.0 and experience failing tests around dynamo. Not only does it seem we need to wait a bit before using the table, we're also unable to list the table with the aws cli. 

Re-running the code throws a `ResourceInUseException: Cannot create preexisting table`, so it is getting created "somewhere", I just can't list it with the CLI. 

### Expected Behavior

I'd expect localstack to return a 200 when the table has been created and is ready to use.

I'd expect the aws CLI to list dynamo tables. 

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

My intern and I spun up a repo: 

https://github.com/kevinvuong01/Local_Stack_Issue

### Environment

```markdown
- OS: monterey 12.6.3
- LocalStack: 1.2/2.0
```

adamdavis40208:
Big props to my intern @kevinvuong01 for learning go, docker, and localstack to reproduce this issue in a repo without all of our proprietary cruft in the repo :P

steffyP:
Hi @adamdavis40208,

thanks for reporting. The link to the reproducible sample does not work - could you please check if the repo is public? 🙏 

We are currently working on a fix for a related issue #8100, which might solve your issue as well.

adamdavis40208:
Lol! It was late, nice catch. 

Yell if you want me to test a fix.

Kevin I'm not at work today can you open up your repo to public access and let us know when you get that done?? (another intern having to bail out his mentor)

kevinvuong01:
Hello,

The repo should be public now.

kevinvuong01:
Hello @steffyP ,

I just changed my repo to public now. Let me know if it's not available.

adamdavis40208:
looks good! thanks Kevin <3

alexrashed:
Hi @kevinvuong01!
This issue could have been related to #8194. This PR re-adds a package necessary for DynamoDB to the default Docker image (which removes the necessity of downloading it on demand when the DynamoDB service in LocalStack is initialized).
[...]

Issue #8100, opened 2023-04-08
TimeoutError: gave up waiting for service dynamodb to start


egoson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I get the following error with version 2.0.0, with version 1.4.0 I don't see the error and the services are working. Still I have to say that the s3 service works well

```
TimeoutError: gave up waiting for service dynamodb to start
localstack           | , headers={'Content-Type': 'application/json', 'X-Amzn-Errortype': 'InternalError', 'Content-Length': '915', 'x-amzn-requestid': 'K9PDQZLBIDPJUCYD06CZHHFF27M6PF9D6QYNDA13IH0ONNCIFGV0', 'x-amz-crc32': '3843399666', 'Connection': 'close', 'Access-Control-Allow-Origin': '*', 'Access-Control-Allow-Methods': 'HEAD,GET,PUT,POST,DELETE,OPTIONS,PATCH', 'Access-Control-Allow-Headers': 'authorization,cache-control,content-length,content-md5,content-type,etag,location,x-amz-acl,x-amz-content-sha256,x-amz-date,x-amz-request-id,x-amz-security-token,x-amz-tagging,x-amz-target,x-amz-user-agent,x-amz-version-id,x-amzn-requestid,x-localstack-target,amz-sdk-invocation-id,amz-sdk-request', 'Access-Control-Expose-Headers': 'etag,x-amz-version-id'})
```

docker compose version 3
 ```
 localstack:
    image: localstack/localstack:2.0.0
    container_name: localstack
    environment:
      - AWS_DEFAULT_REGION=eu-central-1
      - LS_LOG=trace
    ports:
      - "4566:4566"
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

docker-compose up

aws --endpoint-url=http://localhost:4566 dynamodb list-tables --region=eu-central-1

### Environment

```markdown
- OS: Ubuntu 22
- LocalStack: 2.0.0
```

egoson:
```
LocalStack version: 2.0.0
2023-04-08T14:46:10.245055667Z LocalStack Docker container id: d8bb13071833
2023-04-08T14:46:10.245090756Z LocalStack build date: 2023-03-30
[...]

Issue #8058, opened 2023-04-04
bug: Dynamodb service has issue starting


mariamr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Dynamodb does not start throwing the bellow errors in the log. There is no /var/lib/localstack directory created on my machine.  

2023-04-04T03:32:26.129  INFO --- [functhread10] l.services.dynamodb.server : Error opening zip file or JAR manifest missing : /var/lib/localstack/lib/dynamodb-local/latest/ddb-local-loader-0.1.jar
2023-04-04T03:32:26.129  INFO --- [functhread10] l.services.dynamodb.server : Error occurred during initialization of VM
2023-04-04T03:32:26.129  INFO --- [functhread10] l.services.dynamodb.server : agent library failed to init: instrument
2023-04-04T03:32:26.130  INFO --- [functhread10] localstack.utils.run       : Restarting process (received exit code 1): ['java', '-Xmx256m', '-javaagent:/var/lib/localstack/lib/dynamodb-local/latest/ddb-local-loader-0.1.jar', '-Djava.library.path=/var/lib/localstack/lib/dynamodb-local/latest/DynamoDBLocal_lib', '-jar', '/var/lib/localstack/lib/dynamodb-local/latest/DynamoDBLocal.jar', '-port', '60909', '-dbPath', '/var/lib/localstack/tmp/state/dynamodb']

### Expected Behavior

I expect to see an empty list of tables when I run `awslocal dynamodb list-tables` but the command times out with the bellow error:

`An error occurred (InternalError) when calling the ListTables operation (reached max retries: 2): exception while calling dynamodb.ListTables: gave up waiting for service dynamodb to start`

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal dynamodb list-tables

### Environment

```markdown
- OS: macOS Ventura 13.2.1
- LocalStack: 2.0.0.post1
```

sbwhitney:
I'm experiencing this issue also.

Environment

- OS: macOS Monterey
- Localstack: 2.0.2dev

alexrashed:
Hi @mariamr!
[...]

Issue #8054, opened 2023-04-03
bug: LocalStack 2.0 create DynamoDB table just after start fails


xtaixe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating a table just after starting LocalStack fails with `An error occurred (ResourceInUseException) when calling the CreateTable operation: Table already exists: Music2`

### Expected Behavior

Table is created.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run --rm -it -p 4566:4566 -p 4510-4559:4510-4559 -e DEBUG=1 localstack/localstack:2.0.0

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    aws --endpoint-url=http://localhost:4566 dynamodb create-table \                                                                                                                                                               
    --table-name Music2 \
    --attribute-definitions \
        AttributeName=Artist,AttributeType=S \
        AttributeName=SongTitle,AttributeType=S \
    --key-schema \
        AttributeName=Artist,KeyType=HASH \
        AttributeName=SongTitle,KeyType=RANGE \
    --provisioned-throughput \
        ReadCapacityUnits=5,WriteCapacityUnits=5 \
    --table-class STANDARD

### Environment

```markdown
- OS: macOS 12.6.3
- LocalStack: 2.0.0
```

### Anything else?

Logs of first time:

```
2023-04-03T16:14:33.619 DEBUG --- [   asgi_gw_0] localstack.packages.api    : Starting installation of dynamodb-local...
2023-04-03T16:14:34.593 DEBUG --- [   asgi_gw_0] localstack.utils.http      : Starting download from https://s3-us-west-2.amazonaws.com/dynamodb-local/dynamodb_local_latest.zip to /tmp/localstack.ddb.zip (42849335 bytes)
2023-04-03T16:14:35.680 DEBUG --- [   asgi_gw_0] localstack.utils.http      : Written 1048576 bytes (total 1048576) to /tmp/localstack.ddb.zip
2023-04-03T16:14:35.900 DEBUG --- [   asgi_gw_0] localstack.utils.http      : Written 1048576 bytes (total 2097152) to /tmp/localstack.ddb.zip
[...]

PR #8194, opened 2023-04-25
re-install DynamoDB in Dockerfile, use cache for archive download

alexrashed:
## Issue
With the release of 2.0, specifically with #7927, we completely refactored the way LocalStack is built.
Together with this change, the pre-installation of LPM packages has been removed in order to further decrease the Docker image size (and lazy-loading them on-demand when the respective service needs a specific package).

Unfortunately, some users are facing issues with DynamoDB, one of the most used services of LocalStack:
- #8192
- #8100
- #8058
- #8054

This only affects the Community Docker image, `localstack/localstack-pro` always contained DynamoDB (as well as other packages).

## Solution
After some internal discussions, we came to the conclusion that it might be the best approach to:
- Pre-ship the LocalStack Community Docker image with all packages necessary to run the most used (top 15) services out-of-the-box (i.e. without the necessity to download any packages on demand).
- If there is a high request for a minimum Docker image, we might introduce a `slim` variant of the image in the future (which then does not contain any preinstalled packages).

## Implications
Analysis of the Docker image size increase due to this change:
```
$ make docker-build
...
$ docker tag localstack/localstack:latest localstack/localstack:ddb
$ docker pull localstack/localstack
...
$ docker inspect localstack/localstack | jq ".[0].Size"                                                                                                                                                                               1075345104
$ docker inspect localstack/localstack:ddb | jq ".[0].Size"                                                                                                                                                                          
1132873716
```
The uncompressed image size will increase by `1132873716 - 1075345104 = 57528612 bytes = ~55MB`.

```
$ docker save localstack/localstack:latest | gzip > image-wo-ddb.tar.gz
$ docker save localstack/localstack:ddb | gzip > image-w-ddb.tar.gz
$ ls -lh | grep image | awk -v OFS='\t' '{print $9, $5}'                                                                                                                                                                     
image-w-ddb.tar.gz	371M
image-wo-ddb.tar.gz	324M
```
The compressed image size increases by **~47MB**.
The lack of compression is not too surprising, considering the fact that the JAR files are already compressed ZIP files.

### DIFF
--- a/localstack/services/dynamodb/packages.py
+++ b/localstack/services/dynamodb/packages.py
@@ -1,7 +1,7 @@
 import os
-import tempfile
 from typing import List
 
+from localstack import config
 from localstack.constants import ARTIFACTS_REPO, MAVEN_REPO_URL
 from localstack.packages import InstallTarget, Package, PackageInstaller
 from localstack.utils.archives import (
@@ -43,7 +43,7 @@ class DynamoDBLocalPackageInstaller(PackageInstaller):
 
     def _install(self, target: InstallTarget):
         # download and extract archive
-        tmp_archive = os.path.join(tempfile.gettempdir(), "localstack.ddb.zip")
+        tmp_archive = os.path.join(config.dirs.cache, "localstack.ddb.zip")
         install_dir = self._get_install_dir(target)
         download_and_extract_with_retry(DYNAMODB_JAR_URL, tmp_archive, install_dir)
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -154,6 +154,15 @@ ADD localstack/ localstack/
 # Generate the plugin entrypoints
 RUN make entrypoints
 
+# Install packages which should be shipped by default
+RUN --mount=type=cache,target=/root/.cache \
+    --mount=type=cache,target=/var/lib/localstack/cache \
+    source .venv/bin/activate && \
+    python -m localstack.cli.lpm install \
+      dynamodb-local && \
+    chown -R localstack:localstack /usr/lib/localstack && \
+    chmod -R 777 /usr/lib/localstack
+
 # link the extensions virtual environment into the localstack venv
 RUN echo /var/lib/localstack/lib/extensions/python_venv/lib/python3.10/site-packages > localstack-extensions-venv.pth && \
     mv localstack-extensions-venv.pth .venv/lib/python*/site-packages/


localstack/services/dynamodb/packages.py
```diff
@@ -43,7 +43,7 @@ def __init__(self):
 
     def _install(self, target: InstallTarget):
         # download and extract archive
-        tmp_archive = os.path.join(tempfile.gettempdir(), "localstack.ddb.zip")
+        tmp_archive = os.path.join(config.dirs.cache, "localstack.ddb.zip")
```
viren-nadkarni, line 46: salt in the wound :sweat:

viren-nadkarni (approved):

xtaixe:
Hi.

Good news. A couple of questions:

- It says "with all packages necessary to run the most used (top 15) services out-of-the-box". Can you let us know which are those services/packages, since I only see DynamoDb in this PR?
- Any chance you could give an estimation of when this will be released?

We started looking into building our own image as suggested [here](https://github.com/localstack/localstack/issues/8100#issuecomment-1506530555) and knowing the answers to those questions would really help us on deciding how to move forward.

Thanks.

alexrashed:
Hi @xtaixe!
From the current standpoint, the next release will most likely be 2.1, where we haven't yet decided on a release date.

First, I would like to stress that LocalStack ships with all the service implementations. All services can be used with the Docker images. However, some services depend on certain packages, which is - in some cases - downloaded on demand.

This PR re-adds the installation of a package needed for dynamodb, which has been removed with `2.0`.
Previously this section in the Dockerfile installed a predefined set of packages: https://github.com/localstack/localstack/blob/dde691c875249512735750661ab294b26be9f5ad/Dockerfile#L210-L227

However, some of these packages were only used in LocalStack Pro (which has been decoupled with v2), and some packages are only used by deprecated provider / implementations not used by default anymore.

This leaves us with the packages for the following services which (1) were installed pre-v2, (2) are used by current Community providers, and (3) are not installed anymore by default (but rather installed in demand):
- `cloudformation`
  - While CloudFormation is also a widely used service, this particular package is only used for a very small subset of CloudFormation. Maybe @dominikschubert wants to chime in here if this package should be installed by default?
- `kinesis`
  - Kinesis is not in the list of our most used services.
- `stepfunctions`
  - Stepfunctions currently isn't either.

Again, this just means that certain packages are downloaded when the service is used the first time.
As outlined in https://github.com/localstack/localstack/issues/8100#issuecomment-1506530555, there are several ways to circumvent this, f.e.:
- Using `SERVICES` and `EAGER_SERVICE_LOADING` to trigger the initialization of the services (and with that the on-demand download of certain packages) right at the startup of LocalStack.
- Using the new file system to mount the "var libs" folder, where these packages are cached on the host machine and thereby survive container re-creations.
- Building your own image and installing the package you want to have in there by default.

I'm happy to discuss the next steps and any changes to this PR :)

xtaixe:
@alexrashed thanks.

1. I understand LocalStack supports all services, I was referring to be able to use them without downloads. When you have a final list of what will be prepackaged (no downloads) I would like to know about it. Any issue/discussion I can follow?
2. Would actually be also good to know/document what will require downloads and what packages to preinstall if you are building your own image, since it seems we might still need to build our own image due to `cloudformation`, `kinesis` and `stepfunctions`, which we use.
3. By the way does that mean that `postgres`, `redis` and `lambda-java-libs` are only used by Pro or only used by deprecated provider / implementations not used by default anymore?

> Using SERVICES and EAGER_SERVICE_LOADING to trigger the initialization of the services (and with that the on-demand download of certain packages) right at the startup of LocalStack.

I tried that before even raising any issue, but given it starts everything in `SERVICES` (quite a few things for us) and with the downloads, the start time was terrible. I think the lazy load is good, but without downloads.

Thanks.

alexrashed:
> When you have a final list of what will be prepackaged (no downloads) I would like to know about it. Any issue/discussion I can follow?

The packages which are installed, can be found in the Dockerfile (i.e. the only one is about to be reintroduced with this PR).
It is installed with a command called "LPM", which stands for "LocalStack Package Manager". We did not advertise this API publicly because there might be changes in the future (since we do have plans to improve the UX), but it also has a `list` command which shows all available plugins.

> Would actually be also good to know/document what will require downloads and what packages to preinstall if you are building your own image, since it seems we might still need to build our own image due to cloudformation, kinesis and stepfunctions, which we use.

A documentation on "how to build your localstack" is definitely something we plan for the future. This is one of the intentions between the "LPM". However, there is no concrete roadmap on when this will be extended.

> By the way does that mean that postgres, redis and lambda-java-libs are only used by Pro or only used by deprecated provider / implementations not used by default anymore?

Yes, exactly.

ayeganyan:
Hi @alexrashed 
We have modified the localstack image with pre-povisioned dynamodb (only).
After running our test suite (which includes cloudformation, step function and kinsesis), we did not notice any zip download in the logs (like it was). Is there any explanation for that?

alexrashed:
Hi @ayeganyan!
The packages are only installed when they are needed. The cloudformation package is only used for a very small subset, so you might just not use the one function that relies on it. For stepfunctions and kinesis, these packages are quite important, so you might have overlooked that in the logs or didn't trigger an invocation.
But to be honest, it's a bit hard to reason on this amount of info to be honest. If you feel like something's wrong, please create an issue with more information (like the Dockerfile, the requests you send, and the logs).

Issue #8174, opened 2023-04-20
bug: S3 returns PutObject => 404 (NoSuchKey) when creating a folder with special character


ramysiha:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When trying to create a folder with special characters ( e.g. a@a ) S3 will return 404 (NoSuchKey)  error.
what is interesting that the folder will actually be created eventually

` <ListBucketResult>
<IsTruncated>false</IsTruncated>
<Marker/>
<Contents>
<Key>a@a/</Key>
<LastModified>2023-04-20T12:05:04Z</LastModified>
<ETag>"d41d8cd98f00b204e9800998ecf8427e"</ETag>
<Size>0</Size>
<StorageClass>STANDARD</StorageClass>
<Owner>
<DisplayName>webfile</DisplayName>
<ID>
75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a
</ID>
</Owner>
</Contents>`

### Expected Behavior

S3 PutObject command shall return 200 instead of 404

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    import boto3
    from localstack_client.patch import enable_local_endpoints
    enable_local_endpoints()
    s3 = boto3.client("s3")
    buckets = s3.list_buckets()
    print(s3.put_object(Bucket='my-test-bucket', Key=('a@a/')))
    print(buckets)

#### Command output
>Traceback (most recent call last):
  File "/tmp/test.py", line 7, in <module>
    print(s3.put_object(Bucket='my-test-bucket', Key=('aaabbg%40derr/')))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rmagdy/.local/lib/python3.11/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rmagdy/.local/lib/python3.11/site-packages/botocore/client.py", line 960, in _make_api_call
    raise error_class(parsed_response, operation_name)
[...]

PR #8198, opened 2023-04-26
fix: s3 parsing key with trailing backslash and reserved chars

giograno:
Addresses #8174.

In a nutshell, the issue was caused by not re-adding a trailing backslash for keys with special quoted characters.


bentsku (approved):
Awesome, thanks for taking care of this! 👍 and providing a clean fix!

PR #8200, opened 2023-04-26
upgrade Werkzeug to 2.3.4

alexrashed:
Right after upgrading to `Werkzeug==2.2.3` with https://github.com/localstack/localstack/pull/8104, version `2.3` has been released: https://github.com/pallets/werkzeug/releases/tag/2.3.0
In the meantime, the current release is 2.3.4: https://github.com/pallets/werkzeug/releases/tag/2.3.4
This upgrade breaks a few tests, which will be fixed within this PR.

Steps:
- [x] Explicitly upgrade Werkzeug to version 2.3.4.
- [x] Analyze test results
- [x] Fix all failing tests
  - [x] `tests.unit.test_motoserver.test_moto_server`
    - This seems to be an issue with our MotoServer bridge (which is only used in legacy code).
    - This might be related to recent changes in dropping certain decoding steps in Werkzeug 2.3.0 (mentioned in the changeset https://werkzeug.palletsprojects.com/en/2.3.x/changes/).
    - This seems to be fixed with https://github.com/getmoto/moto/pull/6266.
    - The change will end up in localstack with https://github.com/localstack/localstack/pull/8270
  - [x] `tests.integration.test_events.TestEvents.test_api_destinations`
     - https://app.circleci.com/pipelines/github/localstack/localstack/14407/workflows/db9cd6c8-d3e1-4b76-bc66-0f2ee7eb8b24/jobs/107287
     - Caused by https://github.com/pallets/werkzeug/commit/5ee439a692dc4474e0311de2496b567eed2d02cf#diff-abeaeb33abe8954dc4551a4631392ac46ad65e8b9f8404cec0e276a0d345af3b
     - This is actually a bug / implication of a change in werkzeug, its implications in http_server, and the test implementation.
     - Header params with underscores are just removed and cannot be retrieved when using the development server of Werkzeug 2.3+.
     - Changed the test for now. Needs to be clarified with @dominikschubert and @dfangl if this is okay.
     - I created an issue in Werkzeug for this: https://github.com/pallets/werkzeug/issues/2693
     - However, it's effectively a limitation for `pytest-httpserver` (which uses the Werkzeug development server, we don't).
  - [x] `tests.unit.aws.protocol.test_parser.py::test_s3_put_object_keys_with_trailing_slash_and_special_characters`
    - Started failing after rebasing, such that it contains #8198.
    - This has been fixed by removing the encoding handling in the parser (since the path param encoding changed with Werkzeug 2.3).
  - [x] `tests.unit.http_.test_asgi.test_multipart_post_large_payload`
    - This is still to be investigated. This could either be an incompatibility in our ASGI/WSGI bridge, or in Werkzeug 2.3.0.
    - The multipart file streams suddenly have a leading `\r\n` when reading the file stream of the request. The test fails because this adds additional 2 characters, which causes the lengths to diverge.
    - Fixed with https://github.com/pallets/werkzeug/pull/2678
  - [x] `tests.unit.utils.test_http_utils.test_download_with_headers`
    - Issue in `pytest-httpserver`: https://github.com/csernazs/pytest-httpserver/issues/241
    - This test will be fixed in 2.3.1: https://github.com/pallets/werkzeug/issues/2665
- [x] Ensure compatibility with downstream deps
  - [x] `tests.integration.docker_utils.test_docker.TestDockerClient.test_push_access_denied`
    - https://github.com/localstack/localstack/runs/13030344607
    - These tests seem unrelated, because they also failed with Werkzeug 2.2.3 in #8202.
  - [x] Run external integration tests against this branch to catch integration issues early.
    - [External tests are 💚 ](https://github.com/localstack/localstack-ext/actions/runs/4923405174)
- [x] Prepare for 2.4 (fix some deprecation warnings)
   - [x] `localstack/http/router.py:137: DeprecationWarning: The 'charset' parameter is deprecated and will be removed in Werkzeug 2.4.`
     - Added a comment in the code to remove with 2.4 (we are copying the request, for now this is still necessary)
   - [x] `localstack/http/router.py:137: DeprecationWarning: The 'encoding_errors' parameter is deprecated and will be removed in Werkzeug 2.4.`
     - Added a comment in the code to remove with 2.4 (we are copying the request, for now this is still necessary)
   - [x] `localstack/testing/pytest/snapshot.py:41: DeprecationWarning: Use get_result() which forces correct exception handling`
   - [x] `localstack/testing/snapshots/transformer.py:145: DeprecationWarning: Flags not at the start of the expression '(?<=(?i)UnsubscribeU' (truncated) but at position 4`
   - [x] `localstack/testing/snapshots/transformer.py:145: DeprecationWarning: Flags not at the start of the expression '(?<=(?i)SubscribeURL' (truncated) but at position 4`

### DIFF
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -70,7 +70,6 @@ from abc import ABC
 from email.utils import parsedate_to_datetime
 from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 from typing.io import IO
-from urllib.parse import quote
 from xml.etree import ElementTree as ETree
 
 import cbor2
@@ -1075,7 +1074,7 @@ class S3RequestParser(RestXMLRequestParser):
             and uri_params is not None
             and shape.serialization.get("location") == "uri"
             and shape.serialization.get("name") == "Key"
-            and request.base_url.endswith(f"{quote(uri_params['Key'])}/")
+            and request.base_url.endswith(f"{uri_params['Key']}/")
         ):
             uri_params = dict(uri_params)
             uri_params["Key"] = uri_params["Key"] + "/"
--- a/localstack/http/router.py
+++ b/localstack/http/router.py
@@ -135,15 +135,16 @@ def call_endpoint(
 
 def _clone_map_without_rules(old: Map) -> Map:
     return Map(
-        default_subdomain=old.default_subdomain,
+        # TODO charset and encoding_errors are deprecated, remove with upgrade to Werkzeug 2.4.0
         charset=old.charset,
+        encoding_errors=old.encoding_errors,
+        default_subdomain=old.default_subdomain,
         strict_slashes=old.strict_slashes,
         merge_slashes=old.merge_slashes,
         redirect_defaults=old.redirect_defaults,
         converters=old.converters,
         sort_parameters=old.sort_parameters,
         sort_key=old.sort_key,
-        encoding_errors=old.encoding_errors,
         host_matching=old.host_matching,
     )
 
--- a/localstack/testing/pytest/snapshot.py
+++ b/localstack/testing/pytest/snapshot.py
@@ -38,7 +38,7 @@ def pytest_runtest_makereport(item: Item, call: CallInfo[None]) -> Optional[Test
     use_legacy_report = os.environ.get("SNAPSHOT_LEGACY_REPORT", "0") == "1"
 
     result: _Result = yield
-    report: TestReport = result.result
+    report: TestReport = result.get_result()
 
     if call.excinfo is not None and isinstance(call.excinfo.value, SnapshotAssertionError):
         err: SnapshotAssertionError = call.excinfo.value
--- a/localstack/testing/snapshots/transformer_utility.py
+++ b/localstack/testing/snapshots/transformer_utility.py
@@ -429,7 +429,7 @@ class TransformerUtility:
             # this will be able to use a KeyValue based once we provide a certificate for message signing in SNS
             # a match must be made case-insensitive because the key casing is different from lambda notifications
             RegexTransformer(
-                r"(?<=(?i)UnsubscribeURL[\"|']:\s[\"|'])(https?.*?)(?=/\?Action=Unsubscribe)",
+                r"(?i)(?<=UnsubscribeURL[\"|']:\s[\"|'])(https?.*?)(?=/\?Action=Unsubscribe)",
                 replacement="<unsubscribe-domain>",
             ),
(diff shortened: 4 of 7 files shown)


thrau:
> * While this one is reproducible, it might not be related (because it's locally reproducible but not fixed with installing the previous Werkzeug version).
> 
> * [ ]  `tests.unit.test_motoserver.test_moto_server`
>   
>   * This seems to be an issue with our MotoServer bridge (which is only used in legacy code).
>   * This might be related to recent changes in dropping certain decoding steps in Werkzeug 2.3.0 (mentioned in the changeset https://werkzeug.palletsprojects.com/en/2.3.x/changes/).
>   * Maybe @thrau can chime in here? Is this still necessary?

the moto server is still used for the legacy s3 provider, since it does not use `call_moto` but the old forwarding proxy logic.

davidism:
I saw this issue linked in a Werkzeug issue. I briefly investigated "The multipart file streams suddenly have a leading `\r\n` when reading the file stream of the request." with a simple example and could not reproduce it. If you find a self-contained example, please report it to Werkzeug, we've been refactoring our form parser a lot lately.

<details><summary>simple example</summary>

```python
from werkzeug import Request
from werkzeug import Response
from werkzeug import run_simple


@Request.application
def app(request: Request) -> Response:
    if request.method == "POST":
        return Response(str(request.files["data"].read().startswith(b"\r\n")))

    return Response("""<!doctype html>
<meta charset=utf-8>
<form method=post enctype=multipart/form-data>
<input type=text name=name>
<input type=file name=data>
<input type=submit>
</form>
""", mimetype="text/html")


run_simple("127.0.0.1", 5000, app)
```

</details>

alexrashed:
Thanks a lot, @davidism! This could also be caused by our internal infrastructure, which is quite tightly coupled to Werkzeug (thanks for the great framework you are maintaining!). If it turns out the root cause is actually in the new form parser, we'll create an issue in the Werkzeug repo.

davidism:
Multipart parsing was fixed https://github.com/pallets/werkzeug/pull/2678 and will go out in Werkzeug 2.3.3 later today.

dominikschubert (approved):
LGTM! Awesome write-up in the PR description. Made it quite easy to understand the context of these changes :+1: 

The test modification is fine by me for now. Need to go over the EventBridge tests soon anyway as they're not validated :+1:

bentsku (approved):
LGTM! Awesome, thanks a lot for taking care of this. Changes look good to me!
Was just a bit worried about the path param encoding and S3 keys, but I've quickly written an AWS tests and it's all good 🎉 awesome! and thanks for the deprecation fix for the SNS snapshot transformer.

Issue #7041, opened 2022-10-18
Multi-Account only working in a limited capacity


dominikschubert:
When using the new [multi-account feature](https://docs.localstack.cloud/tools/multi-account-setups/) you might experience issues when a LocalStack service performs internal calls to another service.

This affects all scenarios where multiple resources are linked together such as Lambda Event Source Mappings or Destinations, SNS notifications, EventBridge rules/targets, CloudFormation deployments and more. 

At this moment only single resource usage is working as intended. An example of this would be `awslocal sqs create-queue`  or `awslocal sns create-topic` with `AWS_ACCESS_KEY_ID=000000000001` set. 

We're working on finding a potential solution and will update this issue as soon as we have new information to share.

ptrhck:
Just checking if I have understood this correctly: Is Something like [VPC Sharing](https://aws.amazon.com/blogs/networking-and-content-delivery/vpc-sharing-a-new-approach-to-multiple-accounts-and-vpc-management/) due to that issue not possible?

viren-nadkarni:
@ptrhck VPC sharing is currently not implemented but this is unrelated to that.

1989shack:
How do I build a multi accounts system Admin flow Host Provider DNS servers host Whois Record hostname servers Domains Registrar Domains Registration Processes Domains WHOIS Records apps and softwares

lyderichti59:
Adding here, for instance, that I'm facing an issue related to these multi-account limitations : I tried to deploy a VPC an a account A, deploy a VPC on a account B, and deploy a peering connection between A and B on account B, but the VPC ID of account A is not found from account B.

I think some resources (such as VPCs), by nature, should be exposed to all accounts.

viren-nadkarni:
Hi @lyderichti59, could you please file this as a bug? That will prevent this issue from getting lost in the comments and help me track it.

HarshCasper:
Due to the forthcoming changes in our distribution and development model ([read more](https://blog.localstack.cloud/the-road-ahead-for-localstack/)), this issue tracker is no longer being maintained, and this issue is now being closed. If you'd like us to take a look or need further support, please reach out through one of our support channels:

-   💬 [GitHub Discussions](https://github.com/orgs/localstack/discussions)
-   📖 [Help & Support](https://docs.localstack.cloud/aws/getting-started/help-support/)

PR #7040, opened 2022-10-18
Execute CI tests with AWS_ACCESS_KEY_ID=000000000001

dominikschubert:
Exploring potential band aids for #7041

PR #8204, opened 2023-04-26
Use non-default formatted access key ID for tests

viren-nadkarni:
## Summary

This PR makes use of formatted access key IDs for tests that resolve to `000000000001` (for primary account) and `000000000002` (for secondary account). The primary account is used for all tests (via the `aws_client` fixture) whereas the latter is used for cross-account testing (via `secondary_aws_client`). This PR also changes to a non-default test region to make sure multi-region works correctly.

In effect, this assures that all services in LocalStack now can work with any arbitrary account ID, and furthermore that service integrations also use the correct account ID. Prior to this PR the fallback account ID `000000000000` and default region `us-east-1` was used across LocalStack. For example, when using a service integration with a non-fallback account ID, the called service would still use the fallback account ID.

## Related

Supersedes: https://github.com/localstack/localstack/pull/7040

Closes: #7041

### DIFF
--- a/localstack/constants.py
+++ b/localstack/constants.py
@@ -145,16 +145,20 @@ DEFAULT_AWS_ACCOUNT_ID = "000000000000"
 # Credentials used in the test suite
 # These can be overridden if the tests are being run against AWS
 # If a structured access key ID is used, it must correspond to the account ID
-TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
-TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
-TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
-TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-east-1"
+TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or "000000000001"
+TEST_AWS_ACCESS_KEY_ID = (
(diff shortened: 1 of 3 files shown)


tests/unit/aws/test_connect.py
```diff
@@ -77,25 +77,6 @@ def test_external_client_credentials_not_loaded_from_env_if_set_to_none(
...
-            region_name="us-east-1",
-            use_ssl=True,
-            verify=False,
-            endpoint_url="http://localhost:4566",
-            aws_access_key_id="test",
-            aws_secret_access_key="test",
-            aws_session_token=None,
-            config=connect_to._config,
-        )
-
```
viren-nadkarni, line 98: Removed because ExternalClientFactory no longer reads config from env

localstack/services/infra.py
```diff
@@ -188,15 +182,6 @@ def log_startup_message(service):
     LOG.info("Starting mock %s service on %s ...", service, gateway_listen_ports_info())
 
 
-def check_aws_credentials():
-    # Setup AWS environment vars, these are used by Boto when LocalStack makes internal cross-service calls
-    os.environ["AWS_ACCESS_KEY_ID"] = constants.DEFAULT_AWS_ACCOUNT_ID
-    os.environ["AWS_SECRET_ACCESS_KEY"] = constants.INTERNAL_AWS_SECRET_ACCESS_KEY
-    session = boto3.Session()
-    credentials = session.get_credentials()
-    assert credentials
```
viren-nadkarni, line 197: Boto credentials must be explicit and determined from the request context

localstack/constants.py
```diff
@@ -8,9 +8,6 @@
 # constant to represent the "local" region, i.e., local machine
 REGION_LOCAL = "local"
 
-# dev environment
-ENV_DEV = "dev"
-
```
viren-nadkarni, line 13: Dead code

PR #8207, opened 2023-04-26
add dynamodb utils to transform python values into DDB types

bentsku:
This is a utils extracted and modified from `botocore`, to allow simple transformation of python types into DDB types accepted by the SDK. This was used in my hackathon project for an IoT feature, and I'm extracting it now in community following @viren-nadkarni advice.

Also added a unit test that showed the first implementation was wrong in some aspects. 

also see https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html

### DIFF
--- a/localstack/services/dynamodb/utils.py
+++ b/localstack/services/dynamodb/utils.py
@@ -1,6 +1,7 @@
 import logging
 import re
-from typing import Dict, List, Optional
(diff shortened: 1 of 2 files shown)


giograno (approved):
Nice util, thank you 💯

PR #8208, opened 2023-04-26
fix hardcoded user agent for APIGW lambda proxy test

bentsku:
Following this test failure: https://app.circleci.com/pipelines/github/localstack/localstack/14435/workflows/accc329c-b2bb-46b3-ab23-17de16e2046a/jobs/107516/tests
It seems the pipeline will be red until this is merged.

It seems in the test we were using an hardcoded user agent to avoid this kind of issue when we update `requests`, but it seems one test was missing it.

I've also added a transformer to avoid updating the snapshot with the api id change on every regeneration.


alexrashed (approved):
Thanks a lot for jumping on this and unblocking / hardening the tests! 🦸🏽

PR #8217, opened 2023-04-27
Add SourceArn for Events Message Publisher

pinzon:
This PR makes the Events service provider pass the RuleArn to the send_event_to_target function. Now the function can put the correct value on the request metadata for the new internal clients. 

Changes:
- [x] Source ARN for Not Scheduled events
- [x] Source ARN for Scheduled Events
- [x]  ~~Scheduled events now handle alternate event buses~~ this should not be possible for parity sake

### DIFF
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -40,6 +40,7 @@ from localstack.services.moto import call_moto
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.aws import aws_stack
 from localstack.utils.aws.arns import event_bus_arn
+from localstack.utils.aws.client_types import ServicePrincipal
 from localstack.utils.aws.message_forwarding import send_event_to_target
 from localstack.utils.collections import pick_attributes
 from localstack.utils.common import TMP_FILES, mkdir, save_file, truncate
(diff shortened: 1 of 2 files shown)


localstack/services/events/provider.py
```diff
@@ -523,10 +539,12 @@ def events_handler_put_events(self):
...
             if filter_event_based_on_event_format(self, rule.name, event_bus_name, formatted_event):
+                rule_targets = self.events_backend.list_targets_by_rule(
+                    rule.name, event_bus_arn(event_bus_name)
+                ).get("Targets", [])
+
                 targets.extend(
-                    self.events_backend.list_targets_by_rule(
-                        rule.name, event_bus_arn(event_bus_name)
-                    )["Targets"]
+                    [target.update({"RuleArn": rule.arn}) or target for target in rule_targets]
```
dfangl, line 547: Might be easier to use dict unions (https://peps.python.org/pep-0584/) here to achieve this, at first I did not quite get what this is supposed to mean here :)

dfangl (changes requested):
Just some minor nits.
I think it would be nice to have some tests for the "Scheduled events now handle alternate event buses" part of the PR though, which seems like a great addition/fix!

pinzon:
@dfangl. while writing the test for alternate event buses I learnt that [scheduled expressions are not supported by buses other than the default](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html).

dfangl (approved):
LGTM! Thanks for verifying the changes for scheduled rules. The test failures are fixed by #8251, so no reason to hold back the merge.

PR #8220, opened 2023-04-28
Raise DependencyNotYetSatisfied instead of KeyError when secret is not yet populated

dfangl:
## Motivation
Currently, we just raise a KeyError when a key is not present in the secret we had a Ref to. However, it is possible that this key will be filled by another resource, for example [a SecretTargetAttachement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secrettargetattachment.html) in the future, which was prevented by the immediate error.

Also, the Execution Status used for the short circuit condition is not actually available in the return value of the API call: https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeChangeSet.html

## Changes
* Raise DependencyNotYetSatisfied instead of KeyError when secret is not yet populated, to be able to wait for another resource to populate it correctly. If this is never the case, the stack will still fail, but it will not fail immediately anymore on the first try.
* Raise a ShortCircuitWaitException when execution status is "EXECUTE_FAILED" instead of "ROLLBACK_COMPLETE".


dominikschubert (approved):
LGTM! Good catch :+1:

PR #8223, opened 2023-04-28
SQS: Cross-account enhancements

viren-nadkarni:
This is part of an ongoing effort to making credentials more explicit -- by use of well defined access key IDs for testing and internal calls, removing fallbacks and use of context variables. In turn this will improve cross-account functionality.

See https://github.com/localstack/localstack/pull/8204


localstack/services/sqs/provider.py
```diff
@@ -571,7 +571,8 @@ def on_before_stop(self):
         self._queue_update_worker.stop()
         self._stop_cloudwatch_metrics_reporting()
 
-    def _require_queue(self, account_id: str, region_name: str, name: str) -> SqsQueue:
+    @staticmethod
+    def _require_queue(account_id: str, region_name: str, name: str) -> SqsQueue:
```
viren-nadkarni, line 575: Because this is now a static method, the mutex is made a global variable `_STORE_LOCK`. This also makes sense as it is not bound to a Provider instance but the Store instance.

PR #8229, opened 2023-04-30
Add mediaconvert client type

simonrw:


thrau (approved):
looking forward to mediaconvert as a service :grin:

Issue #8232, opened 2023-05-01
enhancement request: Display a specified number of log lines on console based on tail option


rahulmlokurte:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

The `localstack logs` command currently displays the complete logs. The `--follow` option can be used to stream the log output to the console in real time. However, sometimes it's not necessary to see the entire log file, and it would be useful to be able to display only a specific number of log lines.

The proposed enhancement is to add a new `--tail` option to the `localstack logs` command. This option will allow users to specify the number of lines they want to see from the log file. For example, `localstack logs --tail 10` would display the last 10 lines of the log file.

This feature would be useful in situations where users only need to see a specific part of the log file, without having to scroll through the entire log output. It would make it easier to find the relevant information and help users save time.

Overall, adding the `--tail` option to the localstack logs command would be a valuable addition to the tool and improve the user experience.
real-time

### 🧑‍💻 Implementation

To implement the `--tail` option in the localstack logs command, we can modify the existing `cmd_logs` function in the
`localstack.py` file. The `tail` option will be passed as an argument to this function.

Note that this implementation will work even if the --follow option is used

### Anything else?

Current Behaviour:
```
localstack git:(master) ✗ localstack logs
2023-05-01T12:29:10.880 DEBUG --- [  MainThread] plugin.manager             : instantiating plugin PluginSpec(localstack.hooks.on_infra_ready.do_run_state_load_all = <function do_run_state_load_all at 0x7f3673be7250>)
2023-05-01T12:29:10.880 DEBUG --- [  MainThread] plugin.manager             : plugin localstack.hooks.on_infra_ready:do_run_state_load_all is disabled
2023-05-01T12:29:18.853 DEBUG --- [   asgi_gw_0] l.aws.serving.wsgi         : GET localhost:4566/_localstack/health
```

Enhanced Behaviour:
```
localstack git:(master) ✗ localstack logs --tail 1
2023-05-01T12:29:18.853 DEBUG --- [   asgi_gw_0] l.aws.serving.wsgi         : GET localhost:4566/_localstack/health
```

PR #8233, opened 2023-05-01
Display a specified number of log lines on console based on tail option

rahulmlokurte:
Issue: #8232


rahulmlokurte:
recheck

localstack/cli/localstack.py
```diff
@@ -193,8 +193,14 @@ def cmd_stop():
     help="Block the terminal and follow the log output",
     default=False,
 )
+@click.option(
+    "--tail",
```
thrau, line 197: could we maybe add `-n` as shorthand? this would be consistent with `tail` and also `docker logs`.

localstack/cli/localstack.py
```diff
@@ -211,10 +217,18 @@ def cmd_logs(follow: bool):
...
         for line in DOCKER_CLIENT.stream_container_logs(container_name):
             print(line.decode("utf-8").rstrip("\r\n"))
+            num_lines += 1
+            if tail is not None and num_lines >= tail:
+                break
+
     else:
-        print(DOCKER_CLIENT.get_container_logs(container_name))
+        logs = DOCKER_CLIENT.get_container_logs(container_name)
+        if tail is not None and isinstance(tail, int):
```
thrau, line 229: this `isinstance` check seems unnecessary since click already does the check for us. we also didn't add it to the `if follow` control path so it would be inconsistent also

thrau (changes requested):
hey @rahulmlokurte, thanks for the contribution! looks like a useful feature.

i have a couple of comments that we could address before merging. generally i think the approach with reading the entire log stream and then filtering after we've loaded everything into memory could be improved by passing the lines argument down to the actual docker invocations. to that end we'd have to modify `get_container_logs` and `stream_container_logs` of the docker client. but we can do that another time.

cc @dfangl

localstack/cli/localstack.py
rahulmlokurte: Thank you for the suggestion. I have added the `-n` option as a shorthand for `--tail` in the latest commit. It is consistent with the `tail` option and the `docker logs` command, as you suggested.

localstack/cli/localstack.py
rahulmlokurte: Thank you for your feedback. You are right that the click already checks the type for us, and it is unnecessary to have an additional check `isinstance`. I have removed it in the latest commit.

thrau (approved):
LGTM! thanks for the contribution @rahulmlokurte, and welcome to the list of contributors :-)

PR #8235, opened 2023-05-02
Lambda: Cross-accounts enhancements

viren-nadkarni:
This PR makes a few minor fixes to improve cross-account access in Lambda.

This ensures smooth functionality when using a non-default account ID.


dfangl (commented):
We might want to change the clients to the new one as well, this could remove the typing imports.

Anyway, in AWS, the lambda code is not stored in each account, but in a central, AWS managed bucket, not directly accessible by the user.
We made our lives easier by just saving it to the default account, but theoretically we can hardcode some account number and use that.
So in essence, there is no real need to always safe to the user account.

viren-nadkarni:
@dfangl: Updated to use new clients :+1: . I am leaving the bits to have the Lambda code in respective account buckets in the spirit of true multi-tenancy.

joe4dev (approved):
LGTM 👍 
It would be nice to have a test covering the new `account_id` in the lambda models and service and/or demonstrating the use of the extended testutil. Are they in ext?

also nice adoption of the new AWS client and refactoring of magic strings ("us-east-1") 👍

PR #8239, opened 2023-05-02
Move CFn utils and remove redundant code

dominikschubert:
Changes

* Move CFn utils from template_deployer to util files
* Remove fallback to "ResourceType" when "Type" was not available (didn't see where this would occur anyway)
* Remove reference to canonical resource type since all now already in their canonical form


simonrw (approved):
LGTM!

PR #8240, opened 2023-05-02
CloudFormation: add types and restructure

simonrw:
Gain understanding by refactoring internals and add static types.

In particular, the `func_details` type was difficult to nail down. 

Remove `configure_resource_via_sdk` since it had two purposes. Instead, inline the two behaviours into
* a  function that resolves the function properties, and
* code that deploys the resource via the SDK.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -696,7 +696,10 @@ def get_resource_model_instance(resource_id: str, resources) -> Optional[Generic
     return instance
 
 
-def execute_resource_action(resource_id: str, stack_name, resources, action_name: str):
+# yeah `Any | None` is a bit pointless, but I want to ensure that None values are represented
+def execute_resource_action(
+    resource_id: str, stack_name: str, resources: dict, action_name: str
+) -> List[Any | None]:
     resource = resources[resource_id]
     resource_type = get_resource_type(resource)
     if action_name == ACTION_CREATE and resource_type:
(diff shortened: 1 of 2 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -67,7 +110,7 @@ class NoStackUpdates(Exception):
 # ---------------------
 
 
-def get_deployment_config(res_type):
+def get_deployment_config(res_type: str) -> FuncDetailsValue | None:
```
dominikschubert, line 113: that doesn't seem to be right :thinking: 

It actually returns something like 

```python
def _create(...):
    ...

{
  "create": _create,
  "delete": {
    "function": "...",
    "parameters": "..."
  }
}
```

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -55,6 +55,49 @@
...
+    # - resources
+    # - resource_type
+    result_handler: Optional[Callable[[dict, str, list[dict], str], None]]
+    """Take the result of the operation and patch the state of the resources, yuck..."""
+    types: Optional[dict[str, Callable]]
+    """Possible type conversions"""
+
+
+# Type definition for func_details supplied to invoke_function and configure_resource_via_sdk
+FuncDetails = list[FuncDetailsValue] | FuncDetailsValue | Any
```
dominikschubert, line 99: Shouldn't that be a `dict[str, Union[FuncDetailsValue, Callable]]` ? :thinking: 

str could even be a union of the literals "create", "delete" for now, but I don't think that's necessary

simonrw: Thanks, I've added another type that's the result of `get_deploy_templates`.

PR #8245, opened 2023-05-03
CFn: Allow tests to exit early

simonrw:
This PR tackles two things:
1. The cloudformation test fixture `is_change_set_finished` does not handle a stack deploy error correctly, which means that failing tests wait for a minute before timing out. This is not good for iterative development.
2. The template deployer has many levels of exception handlers which catch exceptions and turn them into stack failures. This may swallow the exception (only presented then in debug logs, which may not be enabled by default) or otherwise ignore the error. 

In this PR, we handle the `EXECUTE_FAILED` exception (which is raised on error with a resource provider), and provide a configuration variable that enables exceptions to be raised at the point of exception, not caught and re-thrown/logged later. This should improve development on CloudFormation going forward.


localstack/testing/pytest/fixtures.py
```diff
@@ -1075,7 +1075,7 @@ def _inner():
 
             check_set = aws_client.cloudformation.describe_change_set(**kwargs)
 
-            if check_set.get("ExecutionStatus") == "ROLLBACK_COMPLETE":
+            if check_set.get("ExecutionStatus") in {"ROLLBACK_COMPLETE", "EXECUTE_FAILED"}:
```
dfangl, line 1078: `ROLLBACK_COMPLETE` does not seem to be a valid execution status ever, which is why I changed this already in #8220 . Sorry for the merge conflict :)

simonrw: Haha thanks, I'll remove this change then :+1:

simonrw:
Closing this PR as @dfangl 's changes are more useful

PR #8252, opened 2023-05-04
Refactor physical resource ID, Ref and GetAtt

dominikschubert:
Changes

* Introduces a `get_ref()` for the `GenericBaseModel`. This replaces the previous `get_physical_resource_id(attribute="Ref")` calls
* GetAtt now actually uses `cfn_get_attribute` instead of first trying to get the physical resource id
* Remove any usage of `get_physical_resource_id` with an attribute parameter
* Adapted models where necessary by adding missing values in `get_cfn_attribute`


localstack/services/cloudformation/service_models.py
```diff
@@ -92,10 +89,15 @@ def get_cfn_attribute(self, attribute_name):
         """Retrieve the given CF attribute for this resource"""
         return self.props.get(attribute_name)
 
+    # TODO: make this stricter
+    def get_ref(self):
+        return self.physical_resource_id or self.get_physical_resource_id()
```
simonrw, line 94: I think this relationship still needs exploring. It's probably the right thing to do for now, but I would like to know if the physical resource id is always what's returned when using the `Fn::Ref` intrinsic.

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -344,22 +365,34 @@ def resolve_ref(stack_name, resources, ref, attribute):
...
+        # 2. a pseudo-parameter (e.g. AWS::Region)
+        # 3. the "value" of a resource
+
+        resource = resources.get(ref)
+        if not resource:
+            raise Exception("Should be detected earlier")
+
+        # TODO: remove after refactoring parameter resolution
+        if resource["Type"] == "Parameter":
+            return resource["Properties"]["Value"]
```
simonrw, line 381: This is presumably after SSM parameter resolution?

PR #8253, opened 2023-05-04
KMS: implement cross-accounts access

viren-nadkarni:
### Summary

This PR adds support for cross-accounts access to following operations in KMS:

- CreateGrant                                                                                                                                                                                                                                
- DescribeKey                                                                      
- GetKeyRotationStatus                                                             
- GetPublicKey                                                                     
- ListGrants                                                                       
- RetireGrant                                                                      
- RevokeGrant                                                                      
- Decrypt                                                                   
- Encrypt                                                                          
- GenerateDataKey                                                                  
- GenerateDataKeyPair                                                              
- GenerateDataKeyPairWithoutPlaintext                                              
- GenerateDataKeyWithoutPlaintext                                                  
- GenerateMac                                                                       
- Sign                                                                             
- Verify                                                                           
- VerifyMac

In essence, this means that it is possible to use KMS keys belonging to other accounts. This works only when using key ARNs for above operations. Furthermore, this must happen in the same region.

### Notes

This PR changes the nature of Grant Tokens. This will therefore break existing pods/persisted states that have grants.


localstack/services/kms/models.py
```diff
@@ -543,8 +545,12 @@ class KmsGrant:
     # simplicity.
     token: str
 
-    def __init__(self, create_grant_request: CreateGrantRequest):
+    def __init__(self, create_grant_request: CreateGrantRequest, account_id: str, region_name: str):
         self.metadata = dict(create_grant_request)
+
+        if not is_valid_key_arn(self.metadata["KeyId"]):
+            self.metadata["KeyId"] = kms_key_arn(self.metadata["KeyId"], account_id, region_name)
```
viren-nadkarni, line 552: This is changed to always store the KMS key ARN rather than just the ID. 

This is required for cross-account key retiring.

localstack/services/kms/models.py
```diff
@@ -554,7 +560,11 @@ def __init__(self, create_grant_request: CreateGrantRequest):
         # The Name field is present with just an empty string value.
         self.metadata.setdefault("Name", "")
 
-        self.token = long_uid()
+        # Encode account ID and region in grant token.
+        # This way the grant can be located when being retired by grant principal.
+        # The token consists of account ID, region name and a UUID concatenated with ':' and encoded with base64
+        decoded_token = account_id + ":" + region_name + ":" + long_uid()
+        self.token = to_str(base64.b64encode(to_bytes(decoded_token)))
```
viren-nadkarni, line 567: Similary, the account ID and region is now encoded in the grant token to help locate the grant when being operated on across-accounts.

localstack/services/kms/provider.py
```diff
@@ -344,56 +416,33 @@ def list_grants(
...
-        key_id = key.metadata.get("KeyId")
+    def _delete_grant(store: KmsStore, grant_id: str, key_id: str):
+        grant = store.grants[grant_id]
 
-        if grant_id not in store.grants:
-            raise InvalidGrantIdException()
-        if store.grants[grant_id].metadata["KeyId"] != key_id:
+        _, _, grant_key_id = parse_key_arn(grant.metadata.get("KeyId"))
+        if key_id != grant_key_id:
             raise ValidationError(f"Invalid KeyId={key_id} specified for grant {grant_id}")
```
viren-nadkarni, line 426: This is refactored to simplify the control flow. Only the common logic is retained in this method.

While `RevokeGrant` and `RetireGrant` both delete a grant, their parameters they take are different.

localstack/services/kms/provider.py
```diff
@@ -404,12 +456,28 @@ def retire_grant(
...
+            decoded_token = to_str(base64.b64decode(grant_token))
+            grant_account_id, grant_region_name, _ = decoded_token.split(":")
+            grant_store = self._get_store(grant_account_id, grant_region_name)
+
+            if grant_token not in grant_store.grant_tokens:
+                raise NotFoundException(f"Unable to find grant token {grant_token}")
+
+            grant_id = grant_store.grant_tokens[grant_token]
+        else:
+            grant_store = self._get_store(context.account_id, context.region)
```
sannya-singal, line 470: This condition will never occur; in all the cases where `grant_token` is `None`, L458 will throw an exception.

sannya-singal (changes requested):

localstack/services/kms/provider.py
viren-nadkarni: @sannya-singal: If grant token is not set, but both grant ID and key ID is set, this condition will occur.

Indeed there was no test for this, I've added one.

This can be tested with following:
```

```

sannya-singal (approved):

Issue #8255, opened 2023-05-04
bug: CreateStackInstances after CreateStackInstances results in an error


pandomic:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`CreateStackInstances` after `CreateStackInstances` results in 500 error:
```
An error occurred (InternalError) when calling the CreateStackInstances operation (reached max retries: 2): exception while calling cloudformation.CreateStackInstances: Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/chain.py", line 90, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/handlers/service.py", line 123, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/handlers/service.py", line 93, in __call__
    skeleton_response = self.skeleton.invoke(context)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(context, instance)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/skeleton.py", line 166, in dispatch_request
    result = handler(context, instance) or {}
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/aws/api/core.py", line 167, in operation_marker
    return fn(*args, **kwargs)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/services/cloudformation/provider.py", line 859, in create_stack_instances
    result = cf_client.create_stack(StackName=stack_name, **kwargs)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/botocore/client.py", line 960, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateStack operation: Stack named "sset-test-stack-set-001-000000000000" already exists with status "CREATE_COMPLETE"
```

[...]

PR #8256, opened 2023-05-04
fix stack set instance creation, add deletion

pandomic:
## Summary

* fix "CreateStackInstances" operation where repetitive calls were resulting in 500 errors. Few runs against AWS have shown that AWS returns an OperationId in any case. This can be also reproduced with `aws copilot`
* update `StackSet` id format to match the one in AWS
* implement `DeleteStackInstances` operation to complete the parity test
* update template metadata on changeset execution
* Include template metadata into `GetTemplateSummary` response

## Linked Issues

https://github.com/localstack/localstack/issues/8255


simonrw (approved):
Awesome, I'm glad we're increasing our snapshot coverage of CFn! 🎉

PR #8264, opened 2023-05-05
fix PutEvents non-existent custom EventBus

### LINKED ISSUE #8261: bug: events.PutEvents fails with custom bus name using boto3

### Current Behavior

The LocalStack Docker image >=2.0.0 throws a 500 `InternalError` when calling the `events.PutEvents` operation via boto3 **with a custom event bus name** (i.e. the event bus name is not `default`).

(docker-compose logs)
```
ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: 'customBusName'
 INFO --- [   asgi_gw_0] localstack.request.aws     : AWS events.PutEvents => 500 (InternalError)
```

(boto3 error traceback)
```
  File "eb_publish.py", line 6, in <module>
    response = eb.put_events(
  File "/Users/sangeeta.jadoonanan/.pyenv/versions/hydra/lib/python3.8/site-packages/botocore/client.py", line 530, in _api_call
  File "/Users/sangeeta.jadoonanan/.pyenv/versions/hydra/lib/python3.8/site-packages/botocore/client.py", line 960, in _make_api_call
botocore.exceptions.ClientError: An error occurred (InternalError) when calling the PutEvents operation (reached max retries: 4): exception while calling events.PutEvents: 'customBusName'
```

### Expected Behavior

I expected the `events.PutEvents` operation to get called successfully and result in a 200 status code with a custom event bus name, like I've tested with **LocalStack 1.4**.

(docker-compose logs running LocalStack 1.4)
```
INFO --- [   asgi_gw_0] localstack.request.aws     : AWS events.PutEvents => 200
```

With a docker-compose file

Here's my `docker-compose.yml`:
```yaml
version: "3.4"

services:
  eventbridge:
    image: localstack/localstack
    container_name: hydra_eventbridge
    ports:
      - 4566:4566
    environment:
      - EAGER_SERVICE_LOADING=1
      - SERVICES=events
```

To start LocalStack, I run `docker-compose up`.

The [image versions `2.0.0`](https://hub.docker.com/layers/localstack/localstack/2.0.0/images/sha256-2d0861a7fd281bb4f8a8404d8249ab4aed278c5ac8bdc55f8c246399e4bffcb8?context=explore) and higher fail when passing a custom bus name. The highest working [image version is `1.4`](https://hub.docker.com/layers/localstack/localstack/1.4/images/sha256-4a966e42eff4bbeec44afe13c2d24a67954742b191caae7c15e56186cc0b9ed8?context=explore).

You can try toggling the versions in the `docker-compose.yml` to verify:
```
image: localstack/localstack:1.4
```

and

```
image: localstack/localstack:2.0.0
```

Here's my Python boto3 snippet to publish an event with a custom bus name:
```python
import boto3

eb = boto3.client(service_name="events", endpoint_url="http://localhost:4566")

eb.put_events(
    Entries=[{
        "Source": "test-source",
        "Detail": "test-detail",
        "DetailType": "test-detail-type",
        "EventBusName": "customBusName"
    }]
)
```

### Environment

```markdown
- OS: macOS Ventura 13.2
- LocalStack: 2.0.0
- Python: 3.8.13
- boto3==1.26.127
- botocore==1.29.127
```

### Anything else?

**TL;DR:** The `1.4` localstack Docker image works for me, and anything from `2.0.0` and higher fails.

I noticed this issue with the latest `localstack` Docker image and worked backwards until I found that the breaking image version was `2.0.0`.

Also worth noting that all versions work when I use the default bus name of `default`.

bentsku:
This PR would fix #8261
AWS allows publishing an Event to a nonexistent custom EventBus, it will just accept it and not do anything with it, as the test demonstrates (we even try to describe the nonexistent custom EventBus and it raises an error)

### DIFF
--- a/tests/integration/test_events.py
+++ b/tests/integration/test_events.py
@@ -367,7 +367,6 @@ class TestEvents:
         sqs_client.set_queue_attributes(
             QueueUrl=queue_url, Attributes={"Policy": json.dumps(policy)}
         )
-        # ALLOWHERE
 
         events_client.create_event_bus(Name=bus_name)
         events_client.put_rule(
--- a/tests/integration/test_events.py
+++ b/tests/integration/test_events.py
@@ -367,6 +367,7 @@ class TestEvents:
         sqs_client.set_queue_attributes(
             QueueUrl=queue_url, Attributes={"Policy": json.dumps(policy)}
         )
+        # ALLOWHERE
 
         events_client.create_event_bus(Name=bus_name)
         events_client.put_rule(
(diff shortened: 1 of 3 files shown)


tests/integration/test_events.py
```diff
@@ -367,6 +367,7 @@ def _put_events_with_filter_to_sqs(
         sqs_client.set_queue_attributes(
             QueueUrl=queue_url, Attributes={"Policy": json.dumps(policy)}
         )
+        # ALLOWHERE
```
dfangl, line 370: What is this supposed to mean?

dfangl (approved):
LGTM!

tests/integration/test_events.py
bentsku: That is a very good question, which I am currently unable to answer? I guess I wanted to use this method at first, but realised it didn't fit my need, took some note? Maybe I wanted to replace this with `events_allow_event_rule_to_sqs_queue` fixture, I think? Anyway I delete that comment 😂 thanks!

Issue #8267, opened 2023-05-05
bug: Creating SQS queue fails with botocore==1.29.127 & botocore==1.31.81


mijolabs:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating an SQS queue appears to fail with AWS SDK for Python ([Boto3](https://pypi.org/project/boto3)) >= 1.26.0.

### Using Boto3 1.26.0 (released Nov 1, 2022. Latest version is 1.26.127)

Installing Boto3 1.26.0.
```shell
pip install boto3==1.26.0
```

Code creating SQS queue:
```python
from boto3 import Session

client = Session().resource("sqs", endpoint_url="http://localhost:4566")
client.create_queue(QueueName="test-queue")
```

LocalStack container log:
```
localstack  | 2023-05-05T10:18:43.915 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed. Missing Action in request for query-protocol service ServiceModel(sqs).
localstack  | 2023-05-05T10:18:43.917  INFO --- [   asgi_gw_0] localstack.request.http    : POST / => 500
localstack  | 2023-05-05T10:18:44.200 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed. Missing Action in request for query-protocol service ServiceModel(sqs).
localstack  | 2023-05-05T10:18:44.204  INFO --- [   asgi_gw_0] localstack.request.http    : POST / => 500
localstack  | 2023-05-05T10:18:44.750 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed. Missing Action in request for query-protocol service ServiceModel(sqs).
localstack  | 2023-05-05T10:18:44.755  INFO --- [   asgi_gw_0] localstack.request.http    : POST / => 500
localstack  | 2023-05-05T10:18:46.953 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed. Missing Action in request for query-protocol service ServiceModel(sqs).
localstack  | 2023-05-05T10:18:46.958  INFO --- [   asgi_gw_0] localstack.request.http    : POST / => 500
localstack  | 2023-05-05T10:18:49.182 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed. Missing Action in request for query-protocol service ServiceModel(sqs).
[...]

PR #8268, opened 2023-05-05
fix SQS JSON protocol support in ASF

alexrashed:
This PR aims at providing JSON support for SQS. AWS supports both protocols, `query` and `json`, for SQS for some time.

- With version `1.27.127` of [botocore](https://github.com/boto/botocore) (the official AWS Python SDK clients) the protocol of the SQS service spec has changed from `query` to `json`: https://github.com/boto/botocore/commit/2395012bd369f89b0bcfbbf830718866fc6ae166
  - This broke the usage of SQS with the AWS Python SDK for that underlying botocore version as reported in #8267.
- With version `1.27.129` this change reverted it again: https://github.com/boto/botocore/pull/2931
- With version `1.31.81` the switch was reintroduced and the issue effects us again: https://github.com/boto/botocore/commit/84c784761d4cde0c0ce3b2b32382f970d23b3cc8
  - Since `2023-11-09`, the JSON protocol is generally available as announced in this [AWS blogpost](https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-sqs-support-json-protocol/)
### Initially explored solution:
- [x] Upgrade to the latest version of `botocore` internally.
- [x] Implement a parser and serializer facade which can dynamically select the right parser/serializer based on the content type / accept headers of the request.
- [x] Implement error code handling for `x-amzn-query-error`, as implemented in https://github.com/boto/botocore/pull/2768
- [x] Upgrade to a moto which contains a fix for https://github.com/getmoto/moto/issues/6286
- [ ] ~~Implement tests which ensure that both, query-protocol and json-protocol requests and responses can be sent for SQS.~~
- [ ] ~~Fix all other tests that might break with the upgrade of botocore.~~
  - ~~Try to make SQS tests protocol agnostic, and the parser tests protocol specific (try not to depend on botocore spec protocol)~~

As it turns out this solution is not feasible, because the JSON specification removed a lot of information from the spec which is used to properly parse and serialize the requests and responses.
This means that there is no other alternative than handling two different specifications.

### New solution
- [x] Upgrade to the latest version of `botocore` internally.
- [x] Add the latest available version of the SQS query specifications to `localstack.aws.data`.
  - This is a newly created folder which is configured as `extra search path` in `botocore` when loading the specifications.
  - This will add a newly created "alias" service called `sqs-query`.

What's left to do:
- [x] Check if it is necessary to support JSON requests for SQS Queue URLs (or implement content negotiation).
  - Not necessary per se, it is not used by the `json` protocol based SDK. Would be nice to implement soon.
- [x] Check if the internal dev endpoints for SQS should be migrated to JSON (or implement content negotiation).
  - those endpoint are not targeted by SDKs, so it is not necessary right now, and can be tackled in a second iteration
- [x] Fix failing tests.
  - Most of them should only fail because they assume an XML response structure, but due to the upgrade of botocore a JSON request (and with that a JSON response) is sent.
- [x] Assert that the implementation works with some tests explicitly using XML
- [x] Ensure compatibility with ext (-ext is mostly green, only 4 tests to fix there with wrong assert)


shadowhand:
As of version 3.285.2 of aws/aws-sdk (for PHP) SQS::GetQueueUrl calls to localstack fail. This means we cannot upgrade the SDK because our CI/CD servers use localstack for tests, and all builds with version 3.285.2 are failing.

alexrashed:
Hi @shadowhand! 
I am really sorry for the inconvenience. However, as you can see here, we are already working on this issue in this PR. 
If you want to discuss the severity or other aspects of this issue, please head over to the related issue: https://github.com/localstack/localstack/issues/8267
Thanks!

localstack/aws/protocol/service_router.py
```diff
@@ -354,7 +357,7 @@ def determine_aws_service_name(request: Request, services: ServiceCatalog = None
                 if len(services_per_prefix) == 1:
                     return services_per_prefix[0]
                 candidates.update(services_per_prefix)
-
+        print(f"{candidates=}")
```
bentsku, line 360: I have the commit ready to remove this, but the CI is green, so will push and merge straight after review is no change is necessary

localstack/aws/protocol/serializer.py
```diff
@@ -1621,6 +1621,28 @@ def _node_to_string(self, root: Optional[ETree.ElementTree], mime_type: str) ->
...
+        error: ServiceException,
+        response: HttpResponse,
+        shape: StructureShape,
+        operation_model: OperationModel,
+        mime_type: str,
+        request_id: str,
+    ) -> None:
+        """
+        Overrides _serialize_error as SQS has a special header for query API legacy reason: 'x-amzn-query-error',
+        which contatained the exception code as well as a Sender field.
```
bentsku, line 1636: will fix this typo, commit is ready

localstack/utils/coverage_docs.py
```diff
@@ -10,6 +10,10 @@ def get_coverage_link_for_service(service_name: str, action_name: str) -> str:
 
     available_services = SERVICE_PLUGINS.list_available()
 
+    # TODO remove this once the sqs-query API has been phased out
```
bentsku, line 13: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html#query-protocol

I don't think the SQS Query (XML) implementation is going to be phased out soon

alexrashed (commented):
Thanks so much @bentsku for finalizing the PR and figuring out the nitty gritty details of the remaining test failures! 🦸🏽 
From my perspective, this PR is ready to be merged as it is, but I would like to get another pair of eyes from @dominikschubert or @thrau before merging this.
I'm not happy with some of the changes in here (like the hacky "provider alias"), but I think we can iron these things out in a follow-up.

dominikschubert (approved):
LGTM, Thanks so much @alexrashed  and @bentsku  for getting this over the line so quickly!

It's probably not going to be the only case where they'll make the switch to JSON I guess :disappointed: 

As @bentsku  said, any typos, etc. can be removed in a follow-up. First let's get this merged :rocket:

ilia-beliaev-miro:
Looks like this change has broke the localstack SQS usage for us due to the service name change from `sqs` to `sqs-query`. We receive `Service 'sqs-query' is not enabled.` now, was it an expected outcome and we should now switch the service name when creating a localstack instance from `sqs` to `sqs-query`?

bentsku:
Hello @ilia-beliaev-miro, this is an unexpected outcome and we will fix it with #9607. Sorry for the issue, you can declare both service names for now, but it should be fixed by the end of the day!

Issue #8142, opened 2023-04-14
bug: Memory leak in SQS


michael-robertson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With a lot of SQS queues (>300 in our case), localstack python process exhibits a memory leak.

Simply creating the queues, without any activity, is enough.

The exact number of queues to trigger the behavior is unknown.

### Expected Behavior

Memory usage in usual patterns.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce
```
pip3 install localstack

export DISABLE_EVENTS=1
export EAGER_SERVICE_LOADING=1 SERVICES=sns,sqs,s3
export LS_LOG=warn
export GATEWAY_LISTEN=0.0.0.0:4566 EDGE_PORT=4566
export DNS_ADDRESS=0

localstack start
```
In another terminal:
```
for i in $(seq 300); do
  aws --endpoint-url http://localhost:4566 sqs create-queue --queue-name mprtest$(printf "%03d" $i)
done
```
Then watch the memory step up ~1MB/minute until system dies from starvation, or docker memory limit hit.

### Environment

```markdown
- OS: Ubuntu 22.04 LTS, GCP e2-standard x86/64
- LocalStack: latest (2.0.1)
```

### Anything else?

This also occurred in the previous Localstack 1.x versions, so it was not introduced in 2.0.

baermat:
Hi @michael-robertson , thank you for reporting this! I tried to reproduce this, and you might be right with the leak, however this will take time to track down.
Since you reported ~1MB/minute, let's assume it's 100MB/hour. What tasks are you running that this is blocking you? This shouldn't hit a memory limit any time soon, so it would be really interesting to know how you encountered this. Thank you!

michael-robertson:
Thanks @baermat  - I'm happy to provide more info.

So we are running our whole stack in isolated test environments, meant to be disposable, but complete enough for engineering teams to develop and test their applications. Part of the stack is S3/SQS/SNS, difficult to orchestrate with real AWS, so it is the perfect use case for localstack really, and it does work well. But we found that it regularly died after 3 days as the system runs out of memory, in that case 8GB. Of course we can increase the instance size to give it more memory runway, enough to support longer lifespans, but as you can guess, that puts some constraints on what we can deem "temporary". 

This is the pattern I am seeing, just a script that checks RSS every second, the last number is the positive diff:
```
[...]

PR #7182, opened 2022-11-15
send SQS metrics to cloudwatch

steffyP:
This PR addresses issue #7131 by sending SQS metrics to cloudwatch.
The metrics are defined in the [AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-available-cloudwatch-metrics.html).

Limitations:
* `ApproximateAgeOfOldestMessage` is currently not considered
* Other `Approximate[...]` metrics are scheduled to be send every minute (on AWS the timeframe might be longer, but for LocalStack it probably makes sense to make the metrics available sooner)

PR #8269, opened 2023-05-05
add config to disable cloudwatch metrics for SQS

steffyP:
Following up on issue #8142 where memory leaks have been reported when LocalStack runs for a longer period (hours/days).

The issue seems to be related to a PR #7182, resulting from a feature request. 
With the PR we also added reporting of `Approximate*` messages to provide AWS parity. On LocalStack those approximate messages are reported periodically, e.g. every minute.
This leads to more metrics being added to CloudWatch, even without any interaction.

We now added two config flags to prevent related issues:
* `SQS_CLOUDWATCH_METRICS_REPORT_INTERVAL` which allows to configure the interval of reporting of the `Approximate*` messages
* `SQS_DISABLE_CLOUDWATCH_METRICS` to disable metrics reporting completely for SQS

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -783,6 +783,14 @@ SQS_ENDPOINT_STRATEGY = os.environ.get("SQS_ENDPOINT_STRATEGY", "") or "off"
 # Disable the check for MaxNumberOfMessage in SQS ReceiveMessage
 SQS_DISABLE_MAX_NUMBER_OF_MESSAGE_LIMIT = is_env_true("SQS_DISABLE_MAX_NUMBER_OF_MESSAGE_LIMIT")
 
+# Disable cloudwatch metrics for SQS
+SQS_DISABLE_CLOUDWATCH_METRICS = is_env_true("SQS_DISABLE_CLOUDWATCH_METRICS")
+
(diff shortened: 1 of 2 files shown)


thrau (approved):
This looks great, thanks @steffyP. The config options are very clear and self-descriptive, and the way we do the initialization seems good to me as well. I guess the only drawback from this approach right now is that we can't enable the cloudwatch publishing behavior after SQS has been initialized, but that seems fine to me for now :-)

suzaku:
@thrau Will this performance fix be backported to 1.3 and 1.4?

thrau:
@suzaku no we currently don't maintain older versions of LocalStack. We encourage everyone to upgrade to the latest version. If you have issues migrating we are happy to help

suzaku:
> @suzaku no we currently don't maintain older versions of LocalStack. We encourage everyone to upgrade to the latest version. If you have issues migrating we are happy to help

According to the document, in LocalStack 2.x the only runtime one can use for Lambda is Docker.
But in the latest K8s clusters, Docker is no longer THE container runtime.
Does this mean we can't use Lambda in K8s if we upgrade to LocalStack 2.x?

PR #8271, opened 2023-05-05
CFn: scaffolding for generating getattr tests

simonrw:
WIP: scaffolding generation for GetAttr tests

These tests test accessing attributes from the schema. The idea is that if these tests are run against AWS, we will record a snapshot of which attributes are valid `GetAtt` targets.


dominikschubert:
One issue I can already see is that we're generating huge snapshots with this kind of test setup. We'll probably need to find a way to strip that down a bit to avoid too much bloat :thinking:

PR #8278, opened 2023-05-08
Migrate s3 notifications to new clients, send test events

dfangl:
## Motivation

To allow the IAM enforcement of s3 notifications, we need to use the new clients for its requests, and properly send the `s3:TestEvent` for sns and sqs targets (https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-enable-disable-notification-intro.html). For lambda, it seems like no such event is sent, so it uses a dryrun invoke there (which needed to be fixed) to check if the invocation passes through.

## Changes
* Fix lambda dryrun invocations
* Introduce a BucketVerificationContext to contain all data necessary for validation (in essence everything sent in the `s3:TestEvent` payload.
* Send `s3:TestEvent` payload (sns, sqs) / dry run invoke (lambda)
* Migrate to the new clients
* Some changes due to changed AWS error messages caused by snapshot update

### DIFF
--- a/localstack/services/awslambda/invocation/lambda_service.py
+++ b/localstack/services/awslambda/invocation/lambda_service.py
@@ -274,6 +274,8 @@ class LambdaService:
             payload = b"{}"
         if invocation_type is None:
             invocation_type = "RequestResponse"
+        if invocation_type == InvocationType.DryRun:
+            return None
         # TODO payload verification  An error occurred (InvalidRequestContentException) when calling the Invoke operation: Could not parse request body into json: Could not parse payload into json: Unexpected character (''' (code 39)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
         #  at [Source: (byte[])"'test'"; line: 1, column: 2]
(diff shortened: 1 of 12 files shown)


bentsku (approved):
LGTM! Thanks a lot for the refactoring, awesome! And especially for adding the missing `TestEvent`! 
The `InvalidArgument` error is a bit annoying, as we would need to patch the specs to have more "space" and allow more to be added. Do you think we should collect the exceptions then properly add them a the end while raising?

dfangl:
Regarding the `InvalidArgument` error:

It seems to first validate the request within itself, correct structure, format of parameters (e.g. ARNs) etc.
This errors are returned without the counter at the end, and only the first (even if multiple errors are there) are reported back.

And then, there are the verifications, which connect to the validation against "state", for example sending test event etc. There, it seems to collect the errors (like in the test I added), and report all of them back at once. For this, we should also have some kind of collecting of errors. 

In the end, it all depends how good we can serialize this exception, I guess.

PR #8279, opened 2023-05-08
fix SdkDockerClient recovery if docker socket not available initially

alexrashed:
Currently, if the Docker socket is not available at the start of LocalStack, the initialization of the `SdkDockerClient` fails and the client will not recover during the runtime of LocalStack.
This PR changes this behavior such that:
- The initialization is tried again if the client is accessed at runtime, but not already available (lazy init of the SDK client if it was unsuccessful before).
- Introduces a retry config `DOCKER_SDK_DEFAULT_RETRIES` which allows configuring the amount of retries when initializing the client.

This addresses https://github.com/localstack/helm-charts/issues/21.

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -1394,7 +1394,7 @@ class TestDockerNetworking:
         init_thread = FuncThread(func=on_demand_init)
         init_thread.start()
         # reset / fix the DOCKER_HOST config
-        # monkeypatch.delenv("DOCKER_HOST")
+        monkeypatch.delenv("DOCKER_HOST")
         # wait for the init thread to finish
         init_thread.join()
         # verify that the client is available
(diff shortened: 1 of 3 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -1370,6 +1371,58 @@ def test_docker_sdk_timeout_seconds(self, monkeypatch):
...
+        def on_demand_init(*args):
+            global sdk_client
+            sdk_client = SdkDockerClient()
+            assert sdk_client.docker_client is not None
+
+        # start initializing the client in another thread (with 10 retries)
+        init_thread = FuncThread(func=on_demand_init)
+        init_thread.start()
+        # reset / fix the DOCKER_HOST config
+        # monkeypatch.delenv("DOCKER_HOST")
```
dfangl, line 1397: I think this should be uncommented

alexrashed: Thanks, of course! That's a leftover from verifying the test case.

dfangl (approved):

PR #8281, opened 2023-05-08
fix GetBucketCORS if CORS does not exist in store

bentsku:
This PR would fix the case when the usr restore the state created with the legacy S3 provider (everything would be stored in moto). The assumption was that moto would raise an exception if the CORS configuration didn't exist. However, while restoring an older state, the data would exist in moto but not in our store, thus creating a value error when trying to access the `CORSRules` (`NoneType` etc etc).

This will properly raise the 404 exception if the state is not properly restored.

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1,7 +1,7 @@
 import datetime
 import logging
 import os
-from typing import IO, Dict, List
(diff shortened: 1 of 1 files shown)


macnev2013 (approved):
LGTM 🚀

PR #8286, opened 2023-05-09
Remove REF_*_ATTRS from service models

dominikschubert:
Follow-up to #8257 
Now removes the now unused variables in the `service_models.py` file


simonrw (approved):
Woot! :tada:

Issue #8183, opened 2023-04-23
bug: S3 (AccessDenied) while trying to remove the file from the versioned bucket with legal hold locking enabled


nickperov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

1. Create a bucket with object locking enabled.
2. Enable versioning in this bucket.
3. Upload file with **key**.  Received **version1** for this file.
4. Put a legal hold lock on the file (  **key** - **version1**).
5. Upload another file with the same **key**.  Received **version2** for this file.
6.  Put a legal hold lock on the file (  **key** - **version2**).
7. Remove the legal hold lock from the first version of the file  (  **key** - **version1**).
8. Try to delete the first version of the file (  **key** - **version1**) -> get An error occurred (AccessDenied) when calling the DeleteObject operation: Access Denied.

The script would run successfully if we change the order of commands (basically remove versions  of the file in reverse order):
(We have one file with version1 and version2, both locked)

- First, remove the lock from the version2.
- Delete version2.
- Remove the lock from version1.
- Delete version1.

### Expected Behavior

After the legal hold lock removed (Status=OFF) from the first uploaded version of the file, files should be accessible for deletion by its key and version.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

```
#!/bin/bash

echo -e "\033[0;32mCreate two text files\033[0m"

echo "Some text version 01..." > example-v1.txt
echo "Some text version 02..." > example-v2.txt

echo -e "\033[0;32mCreate bucket\033[0m"
awslocal s3api create-bucket --bucket my-bucket-001 --create-bucket-configuration LocationConstraint=ap-southeast-2 --region ap-southeast-2 --object-lock-enabled-for-bucket

echo -e "\033[0;32mEnable versioning\033[0m"
awslocal s3api put-bucket-versioning --bucket my-bucket-001 --versioning-configuration Status=Enabled

echo -e "\033[0;32mUpload first version of the file\033[0m"
version1=$(awslocal s3api put-object --bucket my-bucket-001 --key 123456789 --body example-v1.txt | jq -r '.VersionId')

echo -e "\033[0;32mPut lock on the version1 of the file\033[0m"
awslocal s3api put-object-legal-hold --bucket my-bucket-001 --key 123456789 --version-id $version1 --legal-hold Status=ON

echo -e "\033[0;32mUpload second version of the file\033[0m"
[...]

PR #8291, opened 2023-05-10
add test for S3 LegalHold when targeting object version

bentsku:
This PR adds test to check the behaviour and the fix of https://github.com/getmoto/moto/pull/6295

This PR builds on #8326 

This will confirm the fix for #8183

### DIFF
--- a/tests/integration/s3/test_s3.py
+++ b/tests/integration/s3/test_s3.py
@@ -4406,6 +4406,89 @@ class TestS3:
         response = aws_client.s3.get_object(Bucket=bucket_1, Key=key_name)
         snapshot.match("get-obj-default-kms-s3-key-from-bucket", response)
 
+    @pytest.mark.aws_validated
+    @pytest.mark.skip_snapshot_verify(paths=["$..ServerSideEncryption"])
+    def test_s3_legal_hold_lock_versioned(self, aws_client, s3_create_bucket, snapshot):
+        snapshot.add_transformer(snapshot.transform.s3_api())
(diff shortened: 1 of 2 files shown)


tests/integration/s3/test_s3.py
```diff
@@ -3887,6 +3887,89 @@ def test_s3_sse_default_kms_key(
         response = aws_client.s3.get_object(Bucket=bucket_1, Key=key_name)
         snapshot.match("get-obj-default-kms-s3-key-from-bucket", response)
 
+    @pytest.mark.aws_validated
+    @pytest.mark.skip_snapshot_verify(paths=["$..ServerSideEncryption"])
```
macnev2013, line 3891: Why are we skipping to verify the path `ServerSideEncryption` ?

macnev2013 (approved):
Have one question about snapshot else LGTM.

macnev2013 (approved):
Have one question about snapshot else LGTM.

macnev2013 (approved):
Have one question about snapshot else LGTM.

tests/integration/s3/test_s3.py
bentsku: Because AWS now has server side encryption by default on all newly created buckets: https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-encryption-faq.html

This behaviour is not yet in moto or LocalStack, so we need to skip the return value for now. All newly created tests now have to have this skip until we fix it 😞

Issue #8211, opened 2023-04-27
bug: SQS deduplication cache is not reset when queue is purged


martin-walsh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`sqs receive-message` does not retrieve messages after a purge if messages have been pushed to the queue which match earlier deduplication id. An empty array is returned instead.

### Expected Behavior

Messages can be retrieved reliably after being sent to queue, after purge.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```

  localstack:
    image: localstack/localstack:2.0.0
    ports:
      - 4566:4566
    volumes:
      - ./localstack:/etc/localstack/init/ready.d
      - "/var/run/docker.sock:/var/run/docker.sock"
    environment:
      - LOCALSTACK_HOST=localstack
      - AWS_DEFAULT_REGION=us-west-2
      - SERVICES=s3,sqs
```

#### Setup script mounted inside ready.d

```bash
awslocal sqs create-queue --queue-name "support-queue.fifo" --attributes FifoQueue=true
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```
awslocal sqs send-message --queue-url http://localstack:4566/000000000000/support-queue.fifo --message-body "Testing 123" --message-group-id "ABC" --message-deduplication-id "123"

awslocal sqs receive-message --queue-url http://localstack:4566/000000000000/support-queue.fifo


awslocal sqs purge-queue --queue-url http://localstack:4566/000000000000/support-queue.fifo

awslocal sqs send-message --queue-url http://localstack:4566/000000000000/support-queue.fifo --message-body "Testing 123" --message-group-id "ABC" --message-deduplication-id "123"

awslocal sqs receive-message --queue-url http://localstack:4566/000000000000/support-queue.fifo
```

### Environment

```markdown
- OS: docker-compose running on macOS or Linux
- LocalStack: 2.0.0, 2.0.2. Worked reliably with earlier localstack.
```

thrau:
should be fixed by #8218, thanks to Martin @martin-walsh for submitting the fix :+1:

PR #8218, opened 2023-04-28
sqs: Clear depduplication cache when FifoQueue is cleared

martin-walsh:
Issue: https://github.com/localstack/localstack/issues/8211

PR #8293, opened 2023-05-10
clear deduplication cache in SQS fifo queues

thrau:
Small fix for https://github.com/localstack/localstack/issues/8211

We didn't clear the deduplication cache when purging a FIFO queue.


thrau:
Completely forgot about #8218 :facepalm:

PR #8297, opened 2023-05-11
Fix loading of CloudFormation deploy UI HTML code

whummer:
Fix loading of CloudFormation deploy UI HTML code, accessible under localhost:4566/_localstack/cloudformation/deploy . Currently we're getting this error with the Pro image:

```
...
localstack_main  |   File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/services/internal.py", line 145, in on_get
localstack_main  |     deploy_html = deploy_html.replace(f"<{key}>", value)
localstack_main  | AttributeError: 'NoneType' object has no attribute 'replace'
localstack_main  | 2023-05-11T07:29:37.101 DEBUG --- [   asgi_gw_0] l.aws.handlers.fallback    : setting internal failure response for 'NoneType' object has no attribute 'replace'
```

The issue was that the `index.html` file is not added to the `localstack-core` pip package, and the new layout of the `localstack/localstack-pro` image loads the dependencies into the `.venv` only:
```
$ docker run --entrypoint= --rm localstack/localstack-pro ls -la
total 16
drwxrwxrwx 1 root root 4096 May  8 18:50 .
drwxr-xr-x 1 root root 4096 May  8 18:47 ..
drwxr-xr-x 1 root root 4096 May  8 18:50 .venv
-rw-r--r-- 1 root root 3222 May  8 18:50 requirements-runtime.txt
```

In contrast, the `localstack/localstack` image has the source code added in CWD directly:
```
$ docker run --entrypoint= --rm localstack/localstack ls
Makefile
bin
localstack
localstack_core.egg-info
pyproject.toml
setup.cfg
setup.py
```

This PR embeds the HTML code in a Python source file directly instead of the `.html` file, which automatically adds it to the pip package.

/cc @alexrashed @MarcelStranak

### DIFF
--- a/localstack/services/cloudformation/deploy.html
+++ b/localstack/services/cloudformation/deploy.html
@@ -1,4 +1,3 @@
-DEPLOY_UI_HTML = """
 <!DOCTYPE html>
 <html lang="en">
   <head>
@@ -143,4 +142,3 @@ DEPLOY_UI_HTML = """
     </script>
   </body>
 </html>
-""".strip()
--- a/localstack/services/cloudformation/deploy_ui.py
+++ b/localstack/services/cloudformation/deploy_ui.py
@@ -1,3 +1,4 @@
+DEPLOY_UI_HTML = """
 <!DOCTYPE html>
 <html lang="en">
   <head>
@@ -142,3 +143,4 @@
     </script>
   </body>
 </html>
+""".strip()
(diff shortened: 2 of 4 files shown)


alexrashed (commented):
Great catch!
Your approach certainly works! But maybe it would be simpler (and also allow syntax highlighting / IDE support on the HTML file) by just adding it to the package data?
The appropriate section would be here: https://github.com/localstack/localstack/blob/815c3993a3d1c734d3b6416ec1525b7311dfe1f0/setup.cfg#L54-L60

And we've already seen an issue like this right with the release, which was fixed with this commit: https://github.com/localstack/localstack/commit/629feb19c0fae6b23b0bab6f8e3064dd22b260dc

whummer:
Good point, thanks for the pointer @alexrashed . Probably makes sense to include it in the package data, so potential future *.html files would also get added and there's no surprises in the difference between the pro and community image. 👍 Updated the PR..

alexrashed (approved):
Awesome, looks good to me! 🚀

PR #8302, opened 2023-05-11
Fix KeyId return value as ARN for "generate_data_key*" methods

dfangl:
## Motivation
#8253 introduced a regression for some methods the key arn in the `KeyId` field instead of the ID.

The API documentation for (at least one) of the operation clearly states it will return the ARN in that field: https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html#API_GenerateDataKey_ResponseSyntax

## Changes
* Add snapshot validated tests for the generate_data_key* methods
* Fix the return value

This is mainly done because the `aws_encryption_sdk` depends on the format being an ARN, which breaks our -ext pipeline.

## Future work
* Check more return values of `KeyId` to be the correct ones @viren-nadkarni

### DIFF
--- a/tests/integration/test_kms.py
+++ b/tests/integration/test_kms.py
@@ -20,11 +20,6 @@ from localstack.services.kms.utils import get_hash_algorithm
 from localstack.utils.strings import short_uid, to_str
 
 
-@pytest.fixture(autouse=True)
-def kms_api_snapshot_transformer(snapshot):
-    snapshot.add_transformer(snapshot.transform.kms_api())
-
-
 @pytest.fixture(scope="class")
 def kms_client_for_region(aws_client_factory):
     def _kms_client(
@@ -66,6 +61,10 @@ def _get_alias(kms_client, alias_name, key_id=None):
 
 
 class TestKMS:
+    @pytest.fixture(autouse=True)
+    def kms_api_snapshot_transformer(self, snapshot):
+        snapshot.add_transformer(snapshot.transform.kms_api())
+
     @pytest.fixture(scope="class")
     def user_arn(self, aws_client):
         return aws_client.sts.get_caller_identity()["Arn"]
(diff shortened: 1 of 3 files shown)


tests/integration/test_kms.py
```diff
@@ -1222,3 +1175,74 @@ def test_cross_accounts_access(self, aws_client, aws_client_factory, kms_create_
...
+    def generate_key_transformers(self, snapshot):
+        snapshot.add_transformer(snapshot.transform.resource_name())
+
+    @pytest.mark.aws_validated
+    def test_generate_data_key_pair_without_plaintext(self, kms_key, aws_client, snapshot):
+        snapshot.add_transformer(
+            snapshot.transform.key_value("PrivateKeyCiphertextBlob", reference_replacement=False)
+        )
+        snapshot.add_transformer(
+            snapshot.transform.key_value("PublicKey", reference_replacement=False)
```
sannya-singal, line 1191: Same as above

tests/integration/test_kms.py
```diff
@@ -1222,3 +1175,74 @@ def test_cross_accounts_access(self, aws_client, aws_client_factory, kms_create_
...
+
+class TestKMSGenerateKeys:
+    @pytest.fixture(autouse=True)
+    def generate_key_transformers(self, snapshot):
+        snapshot.add_transformer(snapshot.transform.resource_name())
+
+    @pytest.mark.aws_validated
+    def test_generate_data_key_pair_without_plaintext(self, kms_key, aws_client, snapshot):
+        snapshot.add_transformer(
+            snapshot.transform.key_value("PrivateKeyCiphertextBlob", reference_replacement=False)
```
sannya-singal, line 1188: Lets add it in the `kms_api()` transform utility as we are transforming this key for every operation in this service.

tests/integration/test_kms.py
```diff
@@ -1222,3 +1175,74 @@ def test_cross_accounts_access(self, aws_client, aws_client_factory, kms_create_
...
+        )
+        snapshot.match("generate-data-key-pair-without-plaintext", result)
+
+    @pytest.mark.aws_validated
+    def test_generate_data_key_pair(self, kms_key, aws_client, snapshot):
+        snapshot.add_transformer(
+            snapshot.transform.key_value("PrivateKeyCiphertextBlob", reference_replacement=False)
+        )
+        snapshot.add_transformer(
+            snapshot.transform.key_value("PrivateKeyPlaintext", reference_replacement=False)
```
sannya-singal, line 1206: Same as above

sannya-singal (changes requested):

sannya-singal:
Thanks for tacking this @dfangl! 🎉

LGTM! Only a minor nit with the transformer.

tests/integration/test_kms.py
sannya-singal: `kms_api` will transform the `key_id` in either case, the regex transformer is used to preserve the structure of the field. Thats why [here](https://github.com/localstack/localstack/pull/8302/files#diff-b64eb19535de434dd14ad306c91e150306f94aa9a23a127046463b1f43d7bb95R1432) the key id is key arn.

sannya-singal (changes requested):

tests/integration/test_kms.py
sannya-singal: kms_api will transform the key_id in either case, the regex transformer is used to preserve the structure of the field. Thats why [here](https://github.com/localstack/localstack/pull/8302/files#diff-b64eb19535de434dd14ad306c91e150306f94aa9a23a127046463b1f43d7bb95R1432) the key id is key arn.

dfangl: Discussed in slack, will put it here for everyone: The tests were moved to a new class to avoid using the autouse fixture with the kms_api transformer.  This was done to avoid a blind key_value replacement of the `KeyId` field, which would have made it impossible to make assertions on its structure.

PR #8306, opened 2023-05-12
Add encryption context to encrypt and decrypt operations in KMS

sannya-singal:
- Fix the return value of `Encrypt` operation 
- Add `EncryptionContext` as an additional authentication data during encryption and decryption 
- Add encryption tags 
- Change mode of encryption from `CBC` to `GCM`, since the previous encryption method lacked built-in authentication and tags. 
- Add snapshot tests for encrypt and decrypt operations

### DIFF
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -255,7 +255,7 @@ class KmsKey:
 
     # Encrypt is a method of KmsKey and not of KmsCryptoKey only because it requires KeyId, and KmsCryptoKeys do not
     # hold KeyIds. Maybe it would be possible to remodel this better.
-    def encrypt(self, plaintext: bytes, encryption_context: EncryptionContextType) -> bytes:
+    def encrypt(self, plaintext: bytes, encryption_context: EncryptionContextType = None) -> bytes:
         iv = os.urandom(IV_LEN)
         aad = _serialize_encryption_context(encryption_context=encryption_context)
         ciphertext, tag = encrypt(self.crypto_key.key_material, plaintext, iv, aad)
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -129,11 +129,14 @@ def deserialize_ciphertext_blob(ciphertext_blob: bytes) -> Ciphertext:
 
 
 def _serialize_encryption_context(encryption_context: EncryptionContextType) -> bytes:
-    aad = io.BytesIO()
-    for key, value in sorted(encryption_context.items(), key=lambda x: x[0]):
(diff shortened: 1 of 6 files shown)


localstack/services/kms/provider.py
```diff
@@ -776,7 +778,10 @@ def decrypt(
         self._validate_key_for_encryption_decryption(context, key)
         self._validate_key_state_not_pending_import(key)
 
-        plaintext = key.decrypt(ciphertext)
+        try:
+            plaintext = key.decrypt(ciphertext, encryption_context)
+        except Exception:
```
viren-nadkarni, line 783: This is a very broad exception block and not recommended. Can you be more specific with the exeception here?

localstack/utils/crypto.py
```diff
@@ -160,18 +161,22 @@ def unpad(s: bytes) -> bytes:
     return s[0 : -s[-1]]
 
 
-def encrypt(key: bytes, message: bytes, iv: bytes = None) -> bytes:
+def encrypt(key: bytes, message: bytes, iv: bytes = None, aad: bytes = None) -> bytes:
     iv = iv or b"0" * BLOCK_SIZE
-    cipher = Cipher(algorithms.AES(key), modes.CBC(iv))
+    cipher = Cipher(algorithms.AES(key), modes.GCM(iv), backend=default_backend())
```
viren-nadkarni, line 166: The PR description contains the **what** but misses out on **why** -- could you clarify why this is changed to GCM?

localstack/utils/crypto.py
```diff
@@ -160,18 +161,22 @@ def unpad(s: bytes) -> bytes:
     return s[0 : -s[-1]]
 
 
-def encrypt(key: bytes, message: bytes, iv: bytes = None) -> bytes:
+def encrypt(key: bytes, message: bytes, iv: bytes = None, aad: bytes = None) -> bytes:
```
viren-nadkarni, line 164: Please update the return type

localstack/services/kms/models.py
```diff
@@ -129,11 +129,14 @@ def deserialize_ciphertext_blob(ciphertext_blob: bytes) -> Ciphertext:
 
 
 def _serialize_encryption_context(encryption_context: EncryptionContextType) -> bytes:
```
viren-nadkarni, line 131:
```suggestion
def _serialize_encryption_context(encryption_context: Optional[EncryptionContextType]) -> bytes:
```

because the default arg is now set to `None` in `encrypt` method

viren-nadkarni (changes requested):

viren-nadkarni (approved):

PR #8307, opened 2023-05-12
remove install dependency on localstack_client

alexrashed:
- It moves the dependency on `localstack_client` (and its transitive dependency `boto3` from `install` to the `runtime` extra).
  - It also cleans up the dependencies a bit (moving `apispec` to `runtime`, removes an unnecessary version pin on `amazon_kclpy`).
  - This effectively reduces the CLI dependencies (because the install dependencies currently are the CLI dependencies).
- It replaces the legacy `service:port` mapping from the `localstack-client` with an dynamically detected list of services based on the loaded plugins.
  - This means that the `localstack-client` does not have to be changed and released when introducing a new service emulator in LocalStack.
  - The ports from the `localstack-client` would only be used with the environment variable `USE_LEGACY_PORTS`, but these wouldn't work with the new "one endpoint for everything" anyways.
  - In the long run, we should try to deprecate the rest of `localstack-client` as well (or maybe completely strip it down to the part patching `boto3`), however, this is out of scope for this PR. It's rather a first step towards that direction.
- Updates the list of "official localstack images" to the new images defined in v2 (this is used by the validation command in the CLI).

I am happy for any feedback!

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import platform
-import re
 import socket
 import subprocess
 import tempfile
@@ -15,7 +14,6 @@ from localstack.constants import (
     DEFAULT_BUCKET_MARKER_LOCAL,
     DEFAULT_DEVELOP_PORT,
     DEFAULT_LAMBDA_CONTAINER_REGISTRY,
-    DEFAULT_SERVICE_PORTS,
     DEFAULT_VOLUME_DIR,
     ENV_INTERNAL_TEST_COLLECT_METRIC,
     ENV_INTERNAL_TEST_RUN,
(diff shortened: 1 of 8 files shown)


localstack/utils/bootstrap.py
```diff
@@ -213,7 +214,13 @@ def get_enabled_apis() -> Set[str]:
 
     The result is cached, so it's safe to call. Clear the cache with get_enabled_apis.cache_clear().
     """
-    return resolve_apis(config.parse_service_ports().keys())
+    services = os.environ.get("SERVICES", "").strip()
```
whummer, line 217: I guess we still need to split the string here? (similar to what we had in `parse_service_ports()` before)

localstack/config.py
```diff
@@ -1016,6 +1014,44 @@ def get_gateway_listen(gateway_listen: str) -> List[HostAndPort]:
 # Whether to return and parse access key ids starting with an "A", like on AWS
 PARITY_AWS_ACCESS_KEY_ID = is_env_true("PARITY_AWS_ACCESS_KEY_ID")
 
+# List of services supported by LocalStack
+SUPPORTED_SERVICES = [
```
whummer, line 1018: Wondering if we should enumerate the list of services explicitly here - would be nice if we can somehow integrate this with the service plugin mechanism. (maybe @thrau has some good ideas here..?)

whummer (commented):
Nice set of changes, kudos for tackling this cleanup @alexrashed ! 🧹 Added two clarification questions at this point. 

Overall, I'm super happy that we're getting rid of `localstack-client` 🚀 and it looks good to me in general, but will let @thrau do the detailed review/signoff.. 👍

localstack/config.py
alexrashed: I thought about this, but the problem is that this is already used when populating the service specific configs (`<service>_PROVIDER_OVERRIDE` and others) and this list is actually used in controlling the services which should be loaded. That made it a bit too risky for me in the first iteration...
But I am happy for any suggestions and would love to get rid of this list all together!

thrau: Might this be a way forward? 5766efb103806d463807b136dd9661ffaa350f7f

* the `populate_config_env_var_names` is mostly used for choosing which env vars should be handed to the container. here it makes sense to not use the service plugin manager yet, since this is called in the `config.py`. however, i think it would be OK to be a bit more lenient, and just use a heuristic based on the key syntax. worst case the container sees some variables it doesn't need. i feel this is an acceptable trade off if it means we don't need to maintain a list of supported services.
* The `get_enabled_apis` control path is only used when `EAGER_SERVICE_LOADING` is set to 1, and we've already called `SERVICE_PLUGINS.list_available()` before, so this seems like a no-brainer

thrau (commented):
Overall the changes look good to me. Quickly tested with the legacy s3 and lambda providers and they still seem to work :+1: I agree with Waldemar that it would be better to use this opportunity to get rid of a static list of services. We don't really need it and if we merge it now it's just going to stick around.

I made a suggestion of how to proceed in an inline comment.

localstack/utils/bootstrap.py
alexrashed: Good point. We would not support the different service ports anyway, but I re-added the parsing (but the ports will just be discarded).

localstack/config.py
alexrashed: Great! Thanks a lot for the commit. I just directly cherry-picked it from your branch.

thrau (approved):
nice! thanks for going the extra mile here. let's :shipit:

PR #8308, opened 2023-05-12
add ClientRequestToken validation for SecretsManager

bentsku:
For `ClientRequestToken` in the following operations (CreateSecret, UpdateSecret, RotateSecret, PutSecretValue), there's a specific note in the documentation:
https://docs.aws.amazon.com/cli/latest/reference/secretsmanager/create-secret.html#options

> If you use the AWS CLI or one of the AWS SDKs to call this operation, then you can leave this parameter empty. The CLI or SDK generates a random UUID for you and includes it as the value for this parameter in the request. If you don't use the SDK and instead generate a raw HTTP request to the Secrets Manager service endpoint, then you must generate a ClientRequestToken yourself for the new version and include the value in the request.

As we tested with boto, we never tried to run a request without it against AWS (just mocked locally with `request`). It is now done with the use of the `aws_http_client_factory`, and I could validate the behaviour against AWS.

This PR also fixes and rewrites tests that had a bad assumption of having the `ClientRequestToken` being created server side like moto does, but as the AWS validated test shows, AWS actually reject any request not having this field. I wonder if those tests have actual value, as they are not AWS tested and I think most of them were testing the wrong behaviour.

### DIFF
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -133,7 +133,12 @@ class SecretsmanagerProvider(SecretsmanagerApi):
     def create_secret(
         self, context: RequestContext, request: CreateSecretRequest
     ) -> CreateSecretResponse:
+        if "ClientRequestToken" not in request:
+            raise InvalidRequestException(
+                "You must provide a ClientRequestToken value. We recommend a UUID-type value."
+            )
         self._raise_if_invalid_secret_id(request["Name"])
+
         return call_moto(context)
 
     @handler("DeleteResourcePolicy", expand=False)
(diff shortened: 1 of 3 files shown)


localstack/services/secretsmanager/provider.py
```diff
@@ -133,7 +133,12 @@ def cancel_rotate_secret(
     def create_secret(
         self, context: RequestContext, request: CreateSecretRequest
     ) -> CreateSecretResponse:
+        if "ClientRequestToken" not in request:
```
MEPalma, line 136: There is some duplication about this check, I think it'd be nicer to have a similar pattern to `_raise_if_invalid_secret_id`.

MEPalma (approved):
Thank you! I think it's ok 😄

localstack/services/secretsmanager/provider.py
bentsku: Done! ✅

PR #8318, opened 2023-05-15
Fix CFn model class & attribute lookup

dominikschubert:
Fixes a regression introduced in recent CFn refactorings. 

PRO overwrites get_resource_type to add support for custom resources, but we didn't use this in get_ref and other call sites, causing this to fail -ext tests.

### DIFF
--- a/localstack/services/cloudformation/models/ec2.py
+++ b/localstack/services/cloudformation/models/ec2.py
@@ -41,7 +41,7 @@ class EC2RouteTable(GenericBaseModel):
             },
             "delete": {
                 "function": "delete_route_table",
-                "parameters": {"RouteTableId": "PhysicalResourceId"},
+                "parameters": {"RouteTableId": "RouteTableId"},
             },
         }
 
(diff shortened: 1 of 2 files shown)


localstack/services/cloudformation/models/ec2.py
```diff
@@ -577,10 +595,6 @@ def _store_instance_id(result, resource_id, resources, resource_type):
             },
             "delete": {
                 "function": "terminate_instances",
-                "parameters": {
-                    "InstanceIds": lambda params, **kw: [
-                        kw["resources"][kw["resource_id"]]["PhysicalResourceId"]
-                    ]
-                },
+                "parameters": {"InstanceIds": ["InstanceId"]},
```
whummer, line 598: nit: If I'm not mistaken, this would create a parameters dict like 
```
{"InstanceIds": "<my-instance-id>"}
```
instead of
```
{"InstanceIds": ["<my-instance-id>"]}
```

Not very intuitive, but that's the current logic in `template_deployer.py`, if I'm reading it correctly: https://github.com/localstack/localstack/blob/master/localstack/services/cloudformation/engine/template_deployer.py#L920-L922

whummer (approved):
Kudos for greenifying the build @dominikschubert ! 🚀 

I have only a small comment/question, but imho we could also address this in a follow-up if we want to get the changes out quickly. 👍

localstack/services/cloudformation/models/ec2.py
dominikschubert: Will fix in a follow-up soon :+1:

PR #8319, opened 2023-05-15
Integrate CloudTrail helper into LocalStack test fixtures

simonrw:
This PR adds a fixture which performs the following:

1. Sets up an IAM role (with blanket permissions) for use with CloudFormation.
2. Yield to the test so that CloudFormation uses this role to deploy a stack.
3. Trigger a step function which waits 5 minutes
4. During step function execution, capture CloudTrail events emitted using the specified role, and save them to S3.

We use our existing fixtures to ensure proper teardown.
An example test has been created at `tests/integration/cloudformation/test_cloudtrail_trace.py`.

In addition, the `deploy_cfn_template` fixture has been updated to accept a `role_arn` argument.

The CDK stack to create the infrastructure resources can be found under `localstack/testing/pytest/cloudtrail_tracking`.

**Documentation for this feature will be available under a new ["Writing a Resource Provider" page in notion](https://www.notion.so/localstack/Writing-a-Resource-Provider-d83bbd2d17ec4bff90d193f232a4e945).**


localstack/services/cloudformation/cloudtrail_tracking/tests/test_cloudtrail_tracking_handler.py
```diff
@@ -0,0 +1,86 @@
+import json
```
dominikschubert, line 1: I'd suggest moving this to `localstack.testing.pytest`

dominikschubert: Nvm seems I misinterpreted the purpose of this file as this is pretty self contained as far as I can see. Not sure if we want to keep that in here at all then.

dominikschubert (approved):
Gave it a quick glance and didn't see any particular issues. Let's get this in and iterate over it afterwards :+1:

PR #8321, opened 2023-05-16
remove pin on botocore

alexrashed:
A recent community contribution was aiming at removing the upper version pin on `botocore` (https://github.com/localstack/localstack/pull/8210, thanks @matt-mercer!). Unfortunately, the latest release at the time contained an invalid spec which was reported in https://github.com/boto/botocore/issues/2925.

The latest release of `botocore` now contains the requested fixes to the kafka service specs: https://github.com/boto/botocore/commit/8bd1a4d486f28e3dc08991e181e06122f9b0d2bd
This is why we can finally _remove the pin_ again.

In the future we might introduce an automate approach (using dependabot or similar bots) to explicitly upgrade these core libraries on a fixed schedule (to move more towards a proactive approach instead of the currently rather reactive approach when new releases of these libraries - like werkzeug or botocore - break our builds).
With #8307 we also moved `botocore` and `boto3` from the install dependencies to the runtime extra (aka from a CLI dependency to a runtime dependency, which allows us to be more restrictive with the pins since the CLI dependencies have a wider impact compared to the runtime dependencies which are usually only used within our Docker images).


alexrashed:
Unfortunately, the upgrade introduces a new issue.
Only a single test fails:
```
test_request_with_response_header_location_fields
```
It seems that the issue is caused by `botocore==1.29.133` (the test is successful with `1.29.132` and reliably fails with `1.29.133`).
This release contains the following changes (unfortunately quite a big changeset due to formatting changes): https://github.com/boto/botocore/compare/1.29.132...1.29.133

The issue is a bit weird though. A single response parameter parsed by the moto response is now different.
This field of the moto response now not only contains the ID of the route53 hosted zone, but it also contains a URL path prefix (f.e. `/hostedzone/BZDOYRJ3BTBQ8OK` instead of `BZDOYRJ3BTBQ8OK`):
https://github.com/localstack/localstack/blob/d8a2f6525a7aabc4ed31e02d5118f0cac244cda8/tests/integration/test_moto.py#L340

thrau (approved):
That's a nice catch! LGTM!

PR #8322, opened 2023-05-16
Refactor handling of CFn stack parameters

dominikschubert:
# Changes
- Parameters are resolved only once when executing CREATE/UPDATE tasks. While the corresponding logic lives in `parameters.py` now, the responsibility is moved further up to the API provider to make this as transparent as possible.
- Unifies parameter resolving (previously this was done in multiple places such as in Stack methods, template_processing, transformations, template deployer)
- Update handling is now more robust and simpler.

# What hasn't changed

- Does not affect how parameters are used when resolving values/refs in the TemplateDeployer/Stack, i.e. they are still merged with resources and mappings/conditions in `Stack.resources` for now. This will be split up in a future PR.
- No new parameter types have been added

Also note I didn't add further tests yet since I didn't want to bloat this too much. I've started on more parity focused tests but I want to give them a bit more polish and build a proper parameter test suite out of it first.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -22,11 +22,7 @@ from localstack.services.cloudformation.deployment_utils import (
     log_not_available_message,
     remove_none_values,
 )
-from localstack.services.cloudformation.engine.entities import (
-    Stack,
-    StackChangeSet,
-    resolve_ssm_parameter_value,
-)
+from localstack.services.cloudformation.engine.entities import Stack, StackChangeSet
 from localstack.services.cloudformation.service_models import (
     KEY_RESOURCE_STATE,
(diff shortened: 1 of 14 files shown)


tests/unit/test_cloudformation.py
```diff
@@ -1,40 +1,8 @@
-import re
```
dominikschubert, line 1: I didn't see much of a value at this point for these resolve unit tests since I'd prefer to use proper integration tests here soon.

localstack/services/cloudformation/engine/template_deployer.py
simonrw, line 1: It's nice to not have to deal with parameters in the deployer itself!

simonrw (approved):
Excited to move this piece of the process to its proper place - before the template deployer gets created :+1:

PR #8327, opened 2023-05-17
CFn: Add AWS::CloudFormation::WaitConditionHandle

simonrw:
This PR implements a resource required by a support request.

https://localstack-community.slack.com/archives/C03F7EQM2M8/p1684235809932229

We don't support wait conditions in CloudFormation, but the resource was required by the deploy process or else the stack didn't deploy.

In addition, we change the resource used by `test_unsupported` to _another_ resource that's not supported by CloudFormation, and we rename the test/file to more directly indicate its use.


pinzon (approved):
LGTM :+1:

PR #8329, opened 2023-05-17
fix ListBuckets tag order

### LINKED ISSUE #8325: bug: S3 bucket owner's name is too long for the API to read. A null exception raised.

### Current Behavior

Bucket.getOwner() in the Java API is not able to parse the new size of the owner's name. A null pointer exception is raised. 

### Expected Behavior

You should see an object of the `Owner` in the Java in version 1.4.0 while a crash with a null pointer exception in localstack version 2.0.0

With a docker-compose file

### Steps To Reproduce

1. To replicate, use the following docker-compose file, twice, once with version 1.4.0, and the other with 2.0.0

Each time run the following command:
2. ` awslocal --endpoint="http://localhost:8010" s3api list-buckets`

In the first you should get
```
{
    "Buckets": [],
    "Owner": {
        "DisplayName": "webfile",
        "ID": "bcaf1ffd86f41161ca5fb16fd081034f"
    }
}
````

In the second you should get 
```
{
    "Buckets": [],
    "Owner": {
        "DisplayName": "webfile",
        "ID": "75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a"
    }
}
```

Notice the difference in the large string in the ID. 

3. create a new bucket:  `awslocal --endpoint="http://localhost:8010" s3 mb s3://mybucket`
4. Use java API to get the owner of the bucket which should map to the behavior [here](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/Bucket.html#getOwner--).

Docker file:
```
version: '3.0'
services:
  localstack:
    container_name: localstack
    image: localstack/localstack:2.0.0 # or switch to 1.4.0
    network_mode: bridge
    environment:
      - DEBUG=1
      - DEFAULT_REGION=us-east-1
      - LAMBDA_EXECUTOR=docker
      - LAMBDA_REMOTE_DOCKER=true
      - LAMBDA_REMOVE_CONTAINERS=true
      - LOCALSTACK_API_KEY=XXXXXX
    volumes:
      - ./localstack/bootstrap:/opt/bootstrap/
    ports:
      - '8010:4566'
```

### Environment

```markdown
- OS:Ubuntu 22.04.2 LTS
- LocalStack: The problem is found since 2.0.0
```

### Anything else?

_No response_

bentsku:
This PR fixes #8325 
This bug seemed really weird, as our response was exactly the same as AWS, format was the same.

However, after trying to debug the issue in Java, it seemed the `Owner` field was properly parsed by the SDK. 

After looking at the code of the SDK, it seemed the order of the XML tag was important for the way they populate each bucket with the owner. Basically, if the Owner tag is second, they first populate every bucket with an empty owner, even though the owner was properly parsed at the end. (I'll put the Java code snippet from the SDK at the end of the PR).

We keep the order from the specs when we serialize the response: we use and iterate over the shapes members.
In the specs, `Buckets` is first and `Owner` is second. The trick is to remove `Buckets` from the specs and add it back, to put it as the last member. Dirty, but it works. 

I could test the fix in the `localstack-java-utils` repo, by adding a test case in `S3FeaturesTest`. 

```java
@Test
public void testGetBucketOwner() {
	AmazonS3 s3 = TestUtils.getClientS3();

	String bucketName = UUID.randomUUID().toString();
	Bucket bucket = s3.createBucket(bucketName);

	List<Bucket> buckets = s3.listBuckets();
	System.out.println("buckets " + buckets);

}
```

I've also added an integration test that is AWS validated to keep the order.

SDK snippet:
```java
protected void doStartElement(String uri, String name, String qName, Attributes attrs) {
    if (this.in(new String[]{"ListAllMyBucketsResult"})) {
        if (name.equals("Owner")) {
            this.bucketsOwner = new Owner();
        }
    } else if (this.in(new String[]{"ListAllMyBucketsResult", "Buckets"}) && name.equals("Bucket")) {
        this.currentBucket = new Bucket();
        this.currentBucket.setOwner(this.bucketsOwner);
    }

}

protected void doEndElement(String uri, String name, String qName) {
    if (this.in(new String[]{"ListAllMyBucketsResult", "Owner"})) {
        if (name.equals("ID")) {
            this.bucketsOwner.setId(this.getText());
        } else if (name.equals("DisplayName")) {
            this.bucketsOwner.setDisplayName(this.getText());
        }
    } else if (this.in(new String[]{"ListAllMyBucketsResult", "Buckets"})) {
        if (name.equals("Bucket")) {
            this.buckets.add(this.currentBucket);
            this.currentBucket = null;
        }
    } else if (this.in(new String[]{"ListAllMyBucketsResult", "Buckets", "Bucket"})) {
        if (name.equals("Name")) {
            this.currentBucket.setName(this.getText());
        } else if (name.equals("CreationDate")) {
            Date creationDate = DateUtils.parseISO8601Date(this.getText());
            this.currentBucket.setCreationDate(creationDate);
        }
    }

}
```

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -2271,8 +2271,8 @@ class ListBucketMetricsConfigurationsRequest(ServiceRequest):
 
 
 class ListBucketsOutput(TypedDict, total=False):
-    Buckets: Optional[Buckets]
     Owner: Optional[Owner]
+    Buckets: Optional[Buckets]
 
 
 class MultipartUpload(TypedDict, total=False):
--- a/localstack/aws/spec-patches.json
+++ b/localstack/aws/spec-patches.json
@@ -773,6 +773,18 @@
         "documentation": "<p>The specified method is not allowed against this resource.</p>",
         "exception": true
       }
+    },
+    {
+      "op": "remove",
+      "path": "/shapes/ListBucketsOutput/members/Buckets"
+    },
+    {
+      "op": "add",
+      "path": "/shapes/ListBucketsOutput/members/Buckets",
+      "value": {
+        "shape":"Buckets",
+        "documentation":"<p>The list of buckets owned by the requester.</p>"
+      }
     }
   ]
 }
--- a/tests/integration/s3/test_s3.py
+++ b/tests/integration/s3/test_s3.py
@@ -3414,6 +3414,12 @@ class TestS3:
 
         resp_dict = xmltodict.parse(resp.content)
         assert "ListAllMyBucketsResult" in resp_dict
+        # validate that the Owner tag is first, before Buckets. This is because the Java SDK is counting on the order
+        # to properly set the Owner value to the buckets.
+        resp_dict["ListAllMyBucketsResult"].pop("@xmlns", None)
+        list_buckets_tags = list(resp_dict["ListAllMyBucketsResult"].keys())
+        assert list_buckets_tags[0] == "Owner"
+        assert list_buckets_tags[1] == "Buckets"
 
         # Lists all objects in a bucket
         bucket_url = _bucket_url(s3_bucket)


thrau (approved):
:scream: what a catch! :medal_sports: thanks @bentsku, the change looks good to me! glad we have the jsonpatch in place. some comments would be useful to understand why patches were made, but i know it's not supported in json. at least we have the blame history and a great PR description.

:100:

PR #8333, opened 2023-05-18
Fix extraction of container Image when using podman

whummer:
Fix extraction of container Image when using podman. We have some customers who are using podman (with Docker not being available), and one of the issues that has surfaced is that the `list_containers(..)` Docker SDK call fails when using podman.

image

The root cause is that certain versions of podman use a slightly different API response format for listing containers: https://github.com/containers/podman/issues/8329 . In particular, the `Image` attribute of containers is `"sha256:..."` for Docker, but it contains the actual image name/tag like `"myimage:mytag-123"` for podman.

A simple integration test is added which simulates the podman API response via a monkey patch. Without the fix in the PR, this test would fail, and the patch in `docker_sdk_client.py` makes the test pass (and should fix the customer issue, too).

We can consider contributing this fix to the upstream repo, too - but for now the goal is to get a fix out quickly to our users.

The PR also removes a few unused strings in `constants.py`. 🧹

### DIFF
--- a/localstack/utils/container_utils/docker_sdk_client.py
+++ b/localstack/utils/container_utils/docker_sdk_client.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import queue
+import re
 import socket
 import threading
 from time import sleep
@@ -788,3 +789,24 @@ class SdkDockerClient(ContainerClient):
             self.client().login(username, password=password, registry=registry, reauth=True)
         except APIError as e:
             raise ContainerException() from e
+
+
+# apply patches required for podman API compatibility
+
+
+@property
+def _container_image(self):
+    image_id = self.attrs.get("ImageID", self.attrs["Image"])
+    if image_id is None:
+        return None
+    image_ref = image_id
+    # Fix for podman API response: Docker returns "sha:..." for `Image`, podman returns "<image-name>:<tag>".
+    # See https://github.com/containers/podman/issues/8329 . Without this check, the Docker client would
+    # blindly strip off the suffix after the colon `:` (which is the `<tag>` in podman's case) which would
+    # then lead to "no such image" errors.
+    if re.match("sha256:[0-9a-f]{64}", image_id, flags=re.IGNORECASE):
+        image_ref = image_id.split(":")[1]
+    return self.client.images.get(image_ref)
+
+
+Container.image = _container_image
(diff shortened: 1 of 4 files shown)


localstack/utils/container_utils/docker_sdk_client.py
```diff
@@ -788,3 +789,24 @@ def login(self, username: str, password: str, registry: Optional[str] = None) ->
...
+def _container_image(self):
+    image_id = self.attrs.get("ImageID", self.attrs["Image"])
+    if image_id is None:
+        return None
+    image_ref = image_id
+    # Fix for podman API response: Docker returns "sha:..." for `Image`, podman returns "<image-name>:<tag>".
+    # See https://github.com/containers/podman/issues/8329 . Without this check, the Docker client would
+    # blindly strip off the suffix after the colon `:` (which is the `<tag>` in podman's case) which would
+    # then lead to "no such image" errors.
+    if re.match("sha256:[0-9a-f]{64}", image_id, flags=re.IGNORECASE):
```
simonrw, line 807: I'm glad this is a specific regex - the more specific we can be the better :+1:

tests/integration/docker_utils/test_docker.py
```diff
@@ -710,6 +711,40 @@ def test_list_containers_filter(self, docker_client: ContainerClient, create_con
...
+        cleanups.append(lambda: docker_client.remove_image(image_name))
+
+        def container_init(self, attrs=None, *args, **kwargs):
+            # Simulate podman API response, Docker returns "sha:..." for Image, podman returns "<image-name>:<tag>".
+            #  See https://github.com/containers/podman/issues/8329
+            attrs["Image"] = image_name
+            container_init_orig(self, attrs=attrs, *args, **kwargs)
+
+        # apply patch to simulate podman behavior
+        container_init_orig = Container.__init__
```
simonrw, line 729: This is probably my issue, but I find it really confusing to have a variable used (`container_init_orig`) _before_ it has been defined. I know Python's scoping rules allow this, but still it's weird. Not that you need to change it, just raising for discussion.

... Or you could move the function definition to in between lines 728 and 729?

simonrw (approved):
Looks good, I'm glad we have a test covering the behaviour, and if it fixes a discrepancy between docker and podman, then it's all great!

tests/integration/docker_utils/test_docker.py
whummer: Good point, I've gotten so used to Python's scoping that I don't notice the difference anymore (although you already pointed it out to me several times!). 😄 But makes sense, updated it! 👍

PR #8339, opened 2023-05-19
SNS: Implement cross-account access

viren-nadkarni:
This PR adds cross-account access support for SNS topics.

This is allowed for following operations:

- GetTopicAttributes
- SetTopicAttributes
- AddPermission
- RemovePermission
- Publish
- Subscribe
- ListSubscriptionsByTopic
- DeleteTopic

Depends on [getmoto/moto#6330](https://github.com/getmoto/moto/pull/6330)

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -573,7 +573,7 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
 
         # An endpoint may only be subscribed to a topic once. Subsequent
         # subscribe calls do nothing (subscribe is idempotent).
-        for existing_topic_subscription in store.topic_subscriptions[topic_arn]:
+        for existing_topic_subscription in store.topic_subscriptions.get(topic_arn, []):
             sub = store.subscriptions.get(existing_topic_subscription, {})
             if sub.get("Endpoint") == endpoint:
                 return SubscribeResponse(SubscriptionArn=sub["SubscriptionArn"])
@@ -596,6 +596,7 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
                 )
 
         store.subscriptions[subscription_arn] = subscription
+        store.topic_subscriptions[topic_arn] = store.topic_subscriptions.get(topic_arn) or []
         store.topic_subscriptions[topic_arn].append(subscription_arn)
 
         # store the token and subscription arn
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -5,6 +5,7 @@ from typing import Dict, List
 from botocore.utils import InvalidArnException
 from moto.core.utils import camelcase_to_pascal, underscores_to_camelcase
 from moto.sns import sns_backends
+from moto.sns.exceptions import SNSNotFoundError
 from moto.sns.models import MAXIMUM_MESSAGE_LENGTH, SNSBackend, Topic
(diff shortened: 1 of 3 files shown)


localstack/services/sns/provider.py
```diff
@@ -109,173 +96,32 @@ def on_after_init(self):
...
-        platform_application_arn: String,
-        attributes: MapStringToString,
-    ) -> None:
-        call_moto(context)
-
-    def set_sms_attributes(
-        self, context: RequestContext, attributes: MapStringToString
-    ) -> SetSMSAttributesResponse:
-        call_moto(context)
-        return SetSMSAttributesResponse()
```
viren-nadkarni, line 247: These definitions are redundant - the provider will fall back to Moto for unimplemented methods

localstack/services/sns/provider.py
```diff
@@ -753,6 +596,7 @@ def subscribe(
                 )
 
         store.subscriptions[subscription_arn] = subscription
+        store.topic_subscriptions[topic_arn] = store.topic_subscriptions.get(topic_arn) or []
         store.topic_subscriptions[topic_arn].append(subscription_arn)
```
bentsku, line 600:
```suggestion
        store.topic_subscriptions.setdefault(topic_arn, []).append(subscription_arn)
        # or maybe
        topic_subscriptions = store.topic_subscriptions.setdefault(topic_arn, [])
        topic_subscriptions.append(subscription_arn)
```
nit: maybe a personal feeling but I like using `setdefault` instead of `or` in a `defaultdict` fashion

localstack/services/sns/provider.py
```diff
@@ -109,173 +96,32 @@ def on_after_init(self):
...
-        self, context: RequestContext, endpoint_arn: String
-    ) -> GetEndpointAttributesResponse:
-        moto_response: GetEndpointAttributesResponse = call_moto(context)
-        return moto_response
-
-    def get_platform_application_attributes(
-        self, context: RequestContext, platform_application_arn: String
-    ) -> GetPlatformApplicationAttributesResponse:
-        moto_response = call_moto(context)
-        return select_from_typed_dict(GetPlatformApplicationAttributesResponse, moto_response)
```
bentsku, line 160: question: there seems to be a `select_from_typed_dict` in this one specific API operation, not sure if it was AWS validated, but we might leave it just in case?

bentsku (approved):
LGTM! Only one minor nit that is mostly personal preference and one question, but really awesome! Also, big thank you for cleaning up the provider by removing all the moto fallback operations. A breath of fresh air 😄

localstack/services/sns/provider.py
viren-nadkarni: The invocation is redundant. The [response structure returned by moto](https://github.com/localstack/moto/blob/3166d063c2e00cf0230b8c35da31b45e5a4bd568/moto/sns/responses.py#L452-L454) is identical to [botocore definition](https://github.com/localstack/localstack/blob/800c54e29231caf5fcd78d5be52a55cb1fe4f453/localstack/aws/api/sns/__init__.py#L408-L409).

localstack/services/sns/provider.py
viren-nadkarni: Nice, TIL about `dict.setdefault` :+1: 

Added in 321c4c32

PR #7936, opened 2023-03-23
Bind endpoints to access SMS messages sent via SNS

ckolek:
**Please refer to the contribution guidelines in the README when submitting PRs.**

fixes #7935

PR #8340, opened 2023-05-20
bind endpoints to access SNS SMS messages

### LINKED ISSUE #7935: feature request: expose endpoint to retrieve SMS messages delivered via SNS

### Feature description

SNS allows for the publishing of messages to an SMS destination with a phone number. Those messages are logged and stored in the `SnsStore`, but there is no way to retrieve them for verification. The request here is to expose a new endpoint which returns SMS messages from the store, similar to the one made available to retrieve email messages sent via SES.

### 🧑‍💻 Implementation

The `SNSServicePlatformEndpointMessagesApiResource` class makes endpoints available for accessing platform endpoint messages sent via SNS. A similar resource class could be implemented to provide new endpoints for accessing SMS messages.

Given that interaction platform endpoint messages is exposed at the path `/_aws/sns/platform-endpoint-messages`, SMS message interaction would presumably be exposed at `/_aws/sns/sms-messages`. A `GET` endpoint could accept an optional `phoneNumber` query parameter to filter messages based on the destination phone number.

```python
class SNSServiceSMSMessagesApiResource:
  @route(sns_constants.SMS_MSGS_ENDPOINT, methods=["GET"])
    def on_get(self, request: Request):
      account_id = #...
      region = #...
      filter_phone_number = request.args.get("phoneNumber")
      # get store; conditionally filter messages based on filter_phone_number
```

### Anything else?

_No response_

rutviklhase:
Fixes #7935 

This PR is based on ( #7936 ) with some appropriate changes to ensure compatibility with the code that has since undergone some change.

### DIFF
--- a/localstack/services/sns/constants.py
+++ b/localstack/services/sns/constants.py
@@ -31,3 +31,4 @@ GCM_URL = "https://fcm.googleapis.com/fcm/send"
 
 # Endpoint to access all the PlatformEndpoint sent Messages
 PLATFORM_ENDPOINT_MSGS_ENDPOINT = "/_aws/sns/platform-endpoint-messages"
+SMS_MSGS_ENDPOINT = "/_aws/sns/sms-messages"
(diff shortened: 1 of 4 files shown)


rutviklhase:

rutviklhase:
recheck

PR #8344, opened 2023-05-22
upgrade debian base image from buster to bullseye

alexrashed:
This PR changes the Python base image type from `slim-buster` to `slim-bullseye`, effectively upgrading the Docker base image's operating system from Debian Buster to Debian Bullseye.
This is basically an updated version of https://github.com/localstack/localstack/pull/7876 (Thanks to @Mins).


thrau (approved):
nice! LGTM :+1:

PR #8348, opened 2023-05-22
fix intrinsic functions Select and FindInMap

pinzon:
The FindInMap intrinsic function should allow the users to select a mapping using a template parameter. And the Select function should allow to define the index as a string.

Changes:

- [x] Template deployer support those specific situations
- [x] Tests modifications to validate the changes

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -573,7 +573,7 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str | 
             attribute = value[keys_list[0]][1]
             attr = resolve_refs_recursively(stack_name, resources, attribute)
             selected_map = value[keys_list[0]][0]
-            if isinstance(selected_map, dict):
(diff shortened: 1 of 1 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -570,8 +570,12 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str |
             return result
 
         if stripped_fn_lower == "findinmap":
-            attr = resolve_refs_recursively(stack_name, resources, value[keys_list[0]][1])
-            result = resolve_ref(stack_name, resources, value[keys_list[0]][0], attribute=attr)
+            attribute = value[keys_list[0]][1]
+            attr = resolve_refs_recursively(stack_name, resources, attribute)
+            selected_map = value[keys_list[0]][0]
+            if isinstance(selected_map, dict):
```
simonrw, line 576: Is it worth being more specific here, i.e
```suggestion
            if isinstance(selected_map, dict) and "Ref" in selected_map:
```

simonrw (approved):
LGTM!

PR #8349, opened 2023-05-22
Enhance support for Podman API in Docker client

whummer:
Enhance support for targeting Podman API with our Docker client. Running on Podman has been increasingly requested by some customers recently, and during testing we've identified some gaps and discrepancies between the Docker API and Podman API responses (see also previous PR #8333). 

## Design Notes

This PR aims to be a starting point to address Podman support more systematically, and also have it covered by tests (we can reuse our excellent Docker client test suite for that purpose). Some/most of the existing Docker client tests are already working against Podman, and a new test marker `@skip_for_podman` has been introduced to temporarily skip failing tests for now. The goal would be to make all test pass over time, and then remove the `@skip_for_podman` marker.

The initial idea was to create a derived `PodmanClient` and inherit the functionality from the Docker client, but it turns out that the two platforms are sufficiently similar to handle both targets in a single client implementation. The majority of compatibility switches (especially in the `CmdDockerClient` class) is to handle slightly different response payloads or error messages returned by the `docker` vs `podman` commands.

The PR introduces a new CI build workflow to run the Podman Docker client tests in a separate build, whenever any of the relevant files are changing.

## Patch vs minor release

Currently the changes are targeting a simple patch release and are not behind a feature flag (arguably shouldn't be breaking, as so far we didn't officially support Podman, and our Docker test coverage is pretty decent). I'm open to discussion, though, whether we want to add the compatibility mode behind a feature flag for now. 👍 

## Summary of changes

* adjust the logic in the Docker CLI and SDK clients to support both, Docker and Podman API responses
* add `strip_wellknown_repo_prefixes` utility to strip off well-known image repo names like `localhost/` or `docker.io/library/` which are added by the Podman API, but not by Docker
* introduce a new CI build workflow to run the Podman Docker client tests in a separate build
* ... more details following soon ...

### DIFF
--- a/.github/workflows/tests-podman.yml
+++ b/.github/workflows/tests-podman.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - ".github/workflows/tests-podman.yml"
       - "localstack/utils/container_utils"
+      - "tests/integration/docker_utils"
     branches:
       - master
       - release/*
@@ -13,6 +14,7 @@ on:
     paths:
       - ".github/workflows/tests-podman.yml"
       - "localstack/utils/container_utils"
+      - "tests/integration/docker_utils"
     branches:
       - master
       - release/*
@@ -54,7 +56,6 @@ jobs:
         run: |
           # determine path of podman socket
           podmanSocket=$(podman system info --format json | jq -r '.host.remoteSocket.path')
-          echo $podmanSocket
-          ls -la $podmanSocket || true
+          echo "Running tests against local podman socket $podmanSocket"
 
           DOCKER_CMD=podman DOCKER_HOST=$podmanSocket TEST_PATH=tests/integration/docker_utils DEBUG=1 make test
--- a/.github/workflows/tests-podman.yml
+++ b/.github/workflows/tests-podman.yml
@@ -5,16 +5,18 @@ on:
   pull_request:
     paths:
       - ".github/workflows/tests-podman.yml"
-      - "localstack/utils/container_utils"
-      - "tests/integration/docker_utils"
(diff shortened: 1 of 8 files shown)


thrau (approved):
Changes LGTM. Love to see the podman support being tested :+1: 

I'm OK merging it as is for now. Having said that, at this point, I would have preferred introdcing a `PodmanContainerClient` class implementation that inherits from the docker sdk client, but isolates the peculiarities of podman into the implementation, rather than putting something like `strip_wellknown_repo_prefixes` into the base abstraction.

Anyway, nice work!

alexrashed (approved):
LGTM! I agree that we should create a separate client class in the next iteration :)
Concerning the bump type: I would say it's a minor change. But this doesn't mean that it needs to be behind a feature flag right now. We currently only block breaking changes from being merged to `master`.

whummer:
@thrau @alexrashed +1 , totally agree - let's pull it out into a separate client in the next iteration. Was very much on the fence whether to extend the base interface here (especially the `strip_wellknown_repo_prefixes` is indeed a bit painful!). 

We'll need to extend the logic in [`create_docker_client()`](https://github.com/localstack/localstack/blob/master/localstack/utils/docker_utils.py#L38) a bit to detect whether to create a Docker or a Podman client (e.g., by inspecting the `system info`, and allowing an override via config variable). We'll also need to consider the fact that we [allow lazy initialization of the client](https://github.com/localstack/localstack/blob/master/localstack/utils/container_utils/docker_sdk_client.py#L59-L61) in case the API is intermittently down (this is where a config override could help). Good plan for the next iteration! 👌

PR #8351, opened 2023-05-23
Added support for s3 analytics configuration

macnev2013:
This PR adds support for S3 Analytics Configurations.

### Bug
The API was not throwing `Not Implemented` error but the feature was not implemented.

### Implemented APIs
- get_bucket_analytics_configuration
- put_bucket_analytics_configuration
- list_bucket_analytics_configurations
- delete_bucket_analytics_configuration

### Parity Improvements with TF
- TestAccS3BucketAnalyticsConfiguration_basic
- TestAccS3BucketAnalyticsConfiguration_removed
- TestAccS3BucketAnalyticsConfiguration_updateBasic
- TestAccS3BucketAnalyticsConfiguration_WithFilter_empty
- TestAccS3BucketAnalyticsConfiguration_WithStorageClassAnalysis_empty
- TestAccS3BucketAnalyticsConfiguration_WithStorageClassAnalysis_default
- TestAccS3BucketAnalyticsConfiguration_WithStorageClassAnalysis_full
- TestAccS3BucketAnalyticsConfiguration_WithFilter_singleTag
- TestAccS3BucketAnalyticsConfiguration_WithFilter_multipleTags
- TestAccS3BucketAnalyticsConfiguration_WithFilter_prefixAndTags
- TestAccS3BucketAnalyticsConfiguration_WithFilter_prefix
- TestAccS3BucketAnalyticsConfiguration_WithFilter_remove

### Notes
- Related closed PR: https://github.com/localstack/localstack/pull/8346

### DIFF
--- a/localstack/services/s3/models.py
+++ b/localstack/services/s3/models.py
@@ -6,7 +6,8 @@ from moto.s3.models import S3Backend as MotoS3Backend
 from localstack import config
 from localstack.aws.api import RequestContext
 from localstack.aws.api.s3 import (
-    AnalyticsConfigurationList,
+    AnalyticsConfiguration,
+    AnalyticsId,
     BucketLifecycleConfiguration,
     BucketName,
     CORSConfiguration,
@@ -50,7 +51,7 @@ class S3Store(BaseStore):
     )
 
     bucket_analytics_configuration: Dict[
-        BucketName, AnalyticsConfigurationList
+        BucketName, Dict[AnalyticsId, AnalyticsConfiguration]
     ] = CrossRegionAttribute(default=dict)
 
 
--- a/localstack/services/s3/models.py
+++ b/localstack/services/s3/models.py
@@ -6,6 +6,7 @@ from moto.s3.models import S3Backend as MotoS3Backend
 from localstack import config
 from localstack.aws.api import RequestContext
 from localstack.aws.api.s3 import (
+    AnalyticsConfigurationList,
     BucketLifecycleConfiguration,
     BucketName,
     CORSConfiguration,
@@ -47,6 +48,10 @@ class S3Store(BaseStore):
     bucket_website_configuration: Dict[BucketName, WebsiteConfiguration] = CrossRegionAttribute(
         default=dict
     )
+
+    bucket_analytics_configuration: Dict[
+        BucketName, AnalyticsConfigurationList
+    ] = CrossRegionAttribute(default=dict)
 
 
 class BucketCorsIndex:
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1172,16 +1172,15 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         moto_backend = get_moto_s3_backend(context)
         get_bucket_from_moto(moto_backend, bucket)
 
-        validate_analytics_configuration(analytics_configuration)
         store = self.get_store()
-        if bucket not in store.bucket_analytics_configuration:
-            store.bucket_analytics_configuration[bucket] = [analytics_configuration]
-            return
-        for analytics_config in store.bucket_analytics_configuration[bucket]:
-            if analytics_config["Id"] == id:
-                analytics_config.update(analytics_configuration)
-                return
-        store.bucket_analytics_configuration[bucket].append(analytics_configuration)
+
+        validate_bucket_analytics_configuration(
+            id=id, analytics_configuration=analytics_configuration
+        )
+
+        if bucket not in store.bucket_analytics_configuration.keys():
+            store.bucket_analytics_configuration[bucket] = {}
+        store.bucket_analytics_configuration[bucket][id] = analytics_configuration
 
     def get_bucket_analytics_configuration(
         self,
@@ -1194,15 +1193,13 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         get_bucket_from_moto(moto_backend, bucket)
 
         store = self.get_store()
-        analytics_configurations: AnalyticsConfigurationList = (
-            store.bucket_analytics_configuration.get(bucket, [])
-        )
-        for analytics_configuration in analytics_configurations:
-            if analytics_configuration["Id"] == id:
-                return GetBucketAnalyticsConfigurationOutput(
-                    AnalyticsConfiguration=analytics_configuration
-                )
-        raise NoSuchConfiguration("The specified configuration does not exist.")
+
+        analytics_configuration: AnalyticsConfiguration = store.bucket_analytics_configuration.get(
+            bucket, {}
+        ).get(id)
+        if not analytics_configuration:
+            raise NoSuchConfiguration("The specified configuration does not exist.")
+        return GetBucketAnalyticsConfigurationOutput(AnalyticsConfiguration=analytics_configuration)
 
     def list_bucket_analytics_configurations(
         self,
@@ -1215,8 +1212,12 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         get_bucket_from_moto(moto_backend, bucket)
 
         store = self.get_store()
-        analytics_configurations: AnalyticsConfigurationList = (
-            store.bucket_analytics_configuration.get(bucket, [])
+        analytics_configurations: Dict[
+            AnalyticsId, AnalyticsConfiguration
+        ] = store.bucket_analytics_configuration.get(bucket, {})
+        analytics_configurations = dict(sorted(analytics_configurations.items()))
+        analytics_configurations: AnalyticsConfigurationList = list(
+            analytics_configurations.values()
         )
         return ListBucketAnalyticsConfigurationsOutput(
             IsTruncated=False, AnalyticsConfigurationList=analytics_configurations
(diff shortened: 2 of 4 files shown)


tests/integration/s3/test_s3.py
```diff
@@ -25,7 +25,7 @@
 from botocore import UNSIGNED
 from botocore.auth import SigV4Auth
 from botocore.client import Config
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, ParamValidationError
```
bentsku, line 28: This kind of `ParamValidationError` are actually a client side validation made by boto. It does not make much sense to snapshot validate them, as they might be different for every SDK, and we only aim to support server side validation. If you want to validate the server side validation, you will need to disable the client side one.
Here are some link on how to do it: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html see `parameter_validation `. We also have an example in code, here: https://github.com/localstack/localstack/blob/master/localstack/aws/forwarder.py#L163-L167

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,109 @@ def get_object_attributes(
...
+            errors.append(
+                f"Unknown parameter in AnalyticsConfiguration: \"{key}\", must be one of: {', '.join(required_keys + optional_keys)}"
+            )
+    for key in ["Id", "StorageClassAnalysis"]:
+        # check if required key is present
+        if key not in analytics_configuration.keys():
+            errors.append(f'Missing required parameter in AnalyticsConfiguration:: "{key}"')
+    if len(errors) > 0:
+        error_message = "\n".join(errors)
+        raise _create_invalid_argument_exc(error_message)
```
bentsku, line 1265: this is a bit unsure for me now, because the exception raised is not the one that you snapshotted, as it was a boto client side validation. You might need to rewrite this once you have the proper server side exception from AWS.

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,109 @@ def get_object_attributes(
...
+def validate_analytics_configuration(analytics_configuration: AnalyticsConfiguration):
+    # TODO: add more concrete validation
+    # mapping of valid keys and whether they are required
+    required_keys = ["Id", "StorageClassAnalysis"]
+    optional_keys = ["Filter"]
+
+    errors = []
+    for key in analytics_configuration.keys():
+        # check if key is valid
+        if key not in required_keys + optional_keys:
```
bentsku, line 1255: this will merge the list on every iteration, maybe it should be done outside the loop once?

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,109 @@ def get_object_attributes(
...
+            errors.append(
+                f"Unknown parameter in AnalyticsConfiguration: \"{key}\", must be one of: {', '.join(required_keys + optional_keys)}"
+            )
+    for key in ["Id", "StorageClassAnalysis"]:
```
bentsku, line 1259:
```suggestion
    for key in required_keys:
```

localstack/services/s3/models.py
```diff
@@ -48,6 +49,10 @@ class S3Store(BaseStore):
         default=dict
     )
 
+    bucket_analytics_configuration: Dict[
+        BucketName, AnalyticsConfigurationList
+    ] = CrossRegionAttribute(default=dict)
+
```
bentsku, line 55: nit: we've seen this pattern emerge several times, and I personally prefer to use dictionaries when storing this kind of data. It makes it much easier to access a specific element and also delete them. When calling `List*` operations, you can simply uses `.values()`, and it keeps the order of insertion. I'd be in favour of changing this to `Dict[BucketName, Dict[AnalyticsId, AnalyticsConfiguration]]` instead.

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,109 @@ def get_object_attributes(
...
+        validate_analytics_configuration(analytics_configuration)
+        store = self.get_store()
+        if bucket not in store.bucket_analytics_configuration:
+            store.bucket_analytics_configuration[bucket] = [analytics_configuration]
+            return
+        for analytics_config in store.bucket_analytics_configuration[bucket]:
+            if analytics_config["Id"] == id:
+                analytics_config.update(analytics_configuration)
+                return
+        store.bucket_analytics_configuration[bucket].append(analytics_configuration)
```
bentsku, line 1184: nit: related to the model change, this could be a bit easier by using a dict to store the bucket's configurations

bentsku (changes requested):
Thanks Nevil for adding this new feature to S3! I have a few comments regarding the data model to store the analytics configurations for a bucket, where I think it could be a bit easier to use a `dict`. 
Also, it seems we're snapshotting client-side validation here, and this would need some changes. I know it's not easy to add server side validation when the SDK already does it, I'm really sorry about that. 
Thanks for also adding the new Terraform tests we're now passing! 🎉

tests/integration/s3/test_s3.py
macnev2013: Sorry, my bad ! I've updated the tests.

tests/integration/s3/test_s3.py
```diff
@@ -4406,6 +4406,153 @@ def test_s3_sse_default_kms_key(
...
+                Id=config_id,
+                AnalyticsConfiguration=analytics_configuration,
+            )
+            get_response = aws_client.s3.get_bucket_analytics_configuration(
+                Bucket=bucket,
+                Id=config_id,
+            )
+            snapshot.match(f"get_{snapshot_id}", get_response)
+
+        def _assert_list_length(expected_length: int):
```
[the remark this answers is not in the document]
macnev2013: The list was unsorted before so it was failing snapshot matching. But I've updated the code to have perfect parity with AWS.

localstack/services/s3/provider.py
macnev2013: Thanks for pointing out. I've removed this validations.

localstack/services/s3/provider.py
macnev2013: Removed this part

localstack/services/s3/provider.py
macnev2013: Removed this part

localstack/services/s3/models.py
macnev2013: Okay. Thanks for the suggestion. I used list earlier to keep it in sync with the inbuilt type that ASF provides. But i agree it will be a lot easier and clean to use dict.

localstack/services/s3/provider.py
macnev2013: updated to dict

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,93 @@ def get_object_attributes(
...
+        analytics_configurations = store.bucket_analytics_configuration.get(bucket, {})
+        if id not in analytics_configurations:
+            raise NoSuchConfiguration("The specified configuration does not exist.")
+        analytics_configurations.pop(id, None)
+
+
+def validate_bucket_analytics_configuration(
+    id: AnalyticsId, analytics_configuration: AnalyticsConfiguration
+) -> None:
+    if id != analytics_configuration["Id"]:
```
bentsku, line 1246: nit: what if `Id` is not the dict? we could be safe with
```suggestion
    if id != analytics_configuration.get("Id"):
```

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,93 @@ def get_object_attributes(
...
+        get_bucket_from_moto(moto_backend, bucket)
+
+        store = self.get_store()
+        analytics_configurations: Dict[
+            AnalyticsId, AnalyticsConfiguration
+        ] = store.bucket_analytics_configuration.get(bucket, {})
+        analytics_configurations = dict(sorted(analytics_configurations.items()))
+        analytics_configurations: AnalyticsConfigurationList = list(
+            analytics_configurations.values()
+        )
```
bentsku, line 1221: I'm not sure I understand what the issue is exactly here? we didn't return the exact same order as AWS? How does AWS sort the list?

We could simplify the code a bit by sorting only the resulting list if AWS sorts on the `Id` value:
```suggestion
        analytics_configurations: AnalyticsConfigurationList = sorted(analytics_configurations.values(), key=lambda x: x["Id"])
```

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,93 @@ def get_object_attributes(
...
+        expected_bucket_owner: AccountId = None,
+    ) -> None:
+        moto_backend = get_moto_s3_backend(context)
+        get_bucket_from_moto(moto_backend, bucket)
+
+        store = self.get_store()
+        analytics_configurations = store.bucket_analytics_configuration.get(bucket, {})
+        if id not in analytics_configurations:
+            raise NoSuchConfiguration("The specified configuration does not exist.")
+        analytics_configurations.pop(id, None)
```
bentsku, line 1240:
```suggestion
        if not analytics_configurations.pop(id, None):
            raise NoSuchConfiguration("The specified configuration does not exist.")
```

bentsku (commented):
Thanks @macnev2013 for updating the PR with the changes! This is looking really good, I have a few minor comments and questions left, and then it should be good to go! 🎉

localstack/services/s3/provider.py
```diff
@@ -1150,6 +1161,93 @@ def get_object_attributes(
...
+
+        store = self.get_store()
+
+        validate_bucket_analytics_configuration(
+            id=id, analytics_configuration=analytics_configuration
+        )
+
+        if bucket not in store.bucket_analytics_configuration.keys():
+            store.bucket_analytics_configuration[bucket] = {}
+        store.bucket_analytics_configuration[bucket][id] = analytics_configuration
```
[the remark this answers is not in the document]
macnev2013: Thanks for the suggestion. I've updated it.

localstack/services/s3/provider.py
macnev2013: Ideally, `AnalyticsConfiguration` is validated by the SDK. And Id and `StorageClassAnalysis` is required field.

But it makes more sense to use safe retrival `.get`.

localstack/services/s3/provider.py
macnev2013: I've updated it but didn't got a chance to push before this review.

Also, Now since the list operation returns same ordered dict as AWS, we are snapshot validating it.

localstack/services/s3/provider.py
macnev2013: Updated. Thanks.

tests/integration/s3/test_s3.py
```diff
@@ -4416,6 +4416,110 @@ def test_s3_sse_default_kms_key(
...
+            },
+        }
+        # id in storage analysis is different from the one in the request
+        with pytest.raises(ClientError) as err_put:
+            aws_client.s3.put_bucket_analytics_configuration(
+                Bucket=bucket,
+                Id="different-id",
+                AnalyticsConfiguration=storage_analysis,
+            )
+        snapshot.match("put_config_with_storage_analysis_err", err_put.value)
```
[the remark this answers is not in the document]
macnev2013: Got it. I've updated it.

bentsku (approved):
Awesome! Thank you for the quick changes! Thanks a lot again for adding this feature and nice snapshot tests 🙏

PR #8353, opened 2023-05-23
fix account resolution in SQS QueryAPI operations

thrau:
## Motivation

This PR fixes the account resolution for operations to the SQS QueryAPI. Previously, we were not parsing authorization headers to allow account access outside the default `000000000000` account. This meant that, if you tried to perform an operation on, say, `http://localhost:4566/123456789012/my-queue`, then after proxying the call to the SQS Provider, `context.account_id` would return `000000000000` instead.

Moreover, in the `delete_queue` implementation in the backend, we were not using the account/region from the parsed queue URL, but instead using that of the context. Technically cross-account delete is not allowed, so it would be fine, but it made sense to fix that and add a defensive piece of code to warn the user that they're potentially doing something that's not allowed.

The issue originally surfaced when a user tried to perform a `DeleteQueue` call using a queue in a different account than the default one.

## Changes

* Extract the credentials from the auth headers and set it as `aws_access_key_id` before proxying the call to the localstack backend
* Fix the `delete_queue` method to use region and account from the queue URL, and log a warning if the users it trying cross-account access
* Add the ability to add credentials to the HTTP client factory
* Add a localstack-only test for multi-account queue delete

### DIFF
--- a/localstack/services/sqs/query_api.py
+++ b/localstack/services/sqs/query_api.py
@@ -20,6 +20,7 @@ from localstack.http import Request, Response, Router, route
 from localstack.http.dispatcher import Handler
 from localstack.services.sqs.exceptions import MissingParameter
 from localstack.utils.aws import aws_stack
+from localstack.utils.aws.aws_stack import extract_access_key_id_from_auth_header
 from localstack.utils.aws.request_context import extract_region_from_headers
 from localstack.utils.strings import long_uid
 
@@ -169,7 +170,13 @@ def try_call_sqs(request: Request, region: str) -> Tuple[Dict, OperationModel]:
 
     # TODO: permissions encoded in URL as AUTHPARAMS cannot be accounted for in this method, which is not a big
     #  problem yet since we generally don't enforce permissions.
-    client = aws_stack.connect_to_service("sqs", region_name=region)
+    account_id = extract_access_key_id_from_auth_header(headers)
+    client = aws_stack.connect_to_service(
+        "sqs",
+        region_name=region,
+        aws_access_key_id=account_id,
+        aws_secret_access_key=account_id,
+    )
     try:
         # using the layer below boto3.client("sqs").<operation>(...) to make the call
         boto_response = client._make_api_call(operation.name, service_request)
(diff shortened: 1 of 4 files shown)


localstack/services/sqs/query_api.py
```diff
@@ -169,7 +170,13 @@ def try_call_sqs(request: Request, region: str) -> Tuple[Dict, OperationModel]:
 
     # TODO: permissions encoded in URL as AUTHPARAMS cannot be accounted for in this method, which is not a big
     #  problem yet since we generally don't enforce permissions.
-    client = aws_stack.connect_to_service("sqs", region_name=region)
+    account_id = extract_access_key_id_from_auth_header(headers)
+    client = aws_stack.connect_to_service(
+        "sqs",
+        region_name=region,
+        aws_access_key_id=account_id,
+        aws_secret_access_key=account_id,
```
baermat, line 178: Just making sure that this is what we want, using the account id as access key id AND as secret access key. Haven't looked into how we manage access here to be honest, it just looked like a possible copy paste problem.

baermat (approved):
Looks mostly good to me :+1:
I have one question which is just to double check something.

viren-nadkarni:
Funnily enough, I have a [draft PR](https://github.com/localstack/localstack/pull/8223/files#diff-5f951499e31a109029f2e44e6b22d1ce3ef7ec6b94ea38c4cd8ec76790735dd3L169) fixing this exact issue :stuck_out_tongue_closed_eyes:

viren-nadkarni (approved):
LGTM!

Fixed a tiny nit myself to get this in before the freeze. Will merge when green.

thrau:
> Funnily enough, I have a [draft PR](https://github.com/localstack/localstack/pull/8223/files#diff-5f951499e31a109029f2e44e6b22d1ce3ef7ec6b94ea38c4cd8ec76790735dd3L169) fixing this exact issue stuck_out_tongue_closed_eyes

oh sorry i completely missed that :( let's make sure you're co-author of this PR then :+1:

> Fixed a tiny nit myself to get this in before the freeze. Will merge when green.
looks like something went wrong :thinking:

PR #8357, opened 2023-05-23
extend conditional if att functionality

pinzon:
This PR allows the template deployer ignore a resource dependency in a ref function if such resource has a conditional deployment and it evaluates to false.

Explained in a different way. If a resource is not deployable due to a condition, and it is referenced ( in this case inside an if function) the ref evaluates to AWS::NoValue instead of preventing the resource to not be deployed.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -517,6 +517,10 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str | 
         # process special operators
         if keys_list == ["Ref"]:
             ref = resolve_ref(stack_name, resources, value["Ref"], attribute="Ref")
+            if ref is None and not evaluate_resource_condition(
+                stack_name, resources, resources[value["Ref"]]
+            ):
+                return PLACEHOLDER_AWS_NO_VALUE
             if ref is None:
                 msg = 'Unable to resolve Ref for resource "%s" (yet)' % value["Ref"]
                 LOG.debug("%s - %s", msg, resources.get(value["Ref"]) or set(resources.keys()))
@@ -1162,7 +1166,10 @@ class TemplateDeployer:
         result = {}
         for resource_id, resource in resources.items():
             if self.is_deployable_resource(resource):
-                if not self.is_deployed(resource):
+                conditional_deployable = evaluate_resource_condition(
+                    self.stack_name, self.resources, resource
+                )
+                if not self.is_deployed(resource) and conditional_deployable:
                     LOG.debug(
                         "Dependency for resource %s not yet deployed: %s %s",
                         depending_resource,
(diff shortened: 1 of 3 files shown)


tests/integration/templates/cfn_if_attribute_none.yml
```diff
@@ -0,0 +1,42 @@
...
+    Default: "Value1"
+  Value2:
+    Type: String
+    Default: "Value2"
+  CreateParameter:
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
```
simonrw, line 13: Just to check - these values are not booleans; does CFn translate boolean yaml values to strings if the parameter type is `String`?!

tests/integration/cloudformation/test_template_engine.py
```diff
@@ -245,6 +245,23 @@ def test_get_azs_function(self, deploy_cfn_template):
...
+
+        deployed = deploy_cfn_template(
+            template_path=template_path,
+            parameters={"CreateParameter": create_parameter},
+        )
+
+        if create_parameter == "false":
+            assert deployed.outputs["Result"] == "Value1"
+        else:
+            assert deployed.outputs["Result"] == "Value2"
```
simonrw, line 263: Suggestion: Maybe snapshot this result, making this test a little simpler?
```suggestion
        snapshot.match("output-value", deployed.outputs["Result"])
```

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1162,7 +1166,10 @@ def get_unsatisfied_dependencies_for_resources(
         result = {}
         for resource_id, resource in resources.items():
             if self.is_deployable_resource(resource):
-                if not self.is_deployed(resource):
+                conditional_deployable = evaluate_resource_condition(
```
simonrw, line 1169: As above

simonrw (changes requested):
I'm really not happy adding more calls to `evaluate_resource_condition`. If it's not a major blocking issue I propose we discuss this on Monday, and maybe you have a go at my suggestion. If it turns out to be insurmountable then we may have to consider this as an option, but I want to be sure we've explored other avenues first.

cc @dominikschubert

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -517,6 +517,10 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str |
         # process special operators
         if keys_list == ["Ref"]:
             ref = resolve_ref(stack_name, resources, value["Ref"], attribute="Ref")
+            if ref is None and not evaluate_resource_condition(
+                stack_name, resources, resources[value["Ref"]]
+            ):
+                return PLACEHOLDER_AWS_NO_VALUE
```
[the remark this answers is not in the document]
pinzon: Seems pretty reasonable. let's talk more about this on monday then :smile:

PR #8364, opened 2023-05-24
make test_fifo_queue_send_message_with_delay_on_queue_works more robust

thrau:
This is a minor change for a test that has been flaking. The reason is that potentially not enough time has passed between the delay and receiving the messages. Since we changed the behavior of `receive_message` in combination with `MaxNumberOfMessages` and `WaitTimeSeconds` there's a chance that fewer messages than 3 will be returned, although we are asking for 3.

So now, we do a hard sleep with an extra second to give the messages enough time to arrive in the queue. since this is a FIFO queue, we should receive all three messages in the `receive_message` call if they are in the queue.


baermat (approved):
LGTM :+1:

PR #8365, opened 2023-05-25
fix release helper script for requirements.txt files

alexrashed:
This PR contains a very small fix which allows using `set-dep-ver` in `bin/release-helper.sh` with `requirements.txt` files (instead of just `setup.cfg`). It just changes the regex such that the spaces at the beginning of the file are optional (`\s*`) instead of "at least once" (`\s+`).


thrau (approved):
:+1:

PR #8366, opened 2023-05-25
fix postgres installation in test pipeline

alexrashed:
This PR fixes our pipeline which tests our integration from Community with Pro.
Thanks @viren-nadkarni for the fix in our pro pipelines, I just migrated the change to this pipeline.

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -24,7 +24,7 @@ on:
         type: string
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


viren-nadkarni (approved):
:heavy_check_mark:

alexrashed:
Since the PR only contains changes to unblock the system setup in a single workflow, I'll not wait for the whole workflow to finish (since the workflows are already past that section and the fix is verified to be working).

PR #8371, opened 2023-05-25
wip: improvement with apigw rest api import

bentsku:
🚧 improving APIGW import from OpenAPI file following recently surfaced issues with the serverless shopping cart sample


tests/integration/apigateway/test_apigateway_basic.py
```diff
@@ -812,95 +809,6 @@ def test_api_gateway_authorizer_crud(self, aws_client):
                 restApiId=self.TEST_API_GATEWAY_ID, authorizerId=authorizer_id
             )
 
-    def test_apigateway_with_lambda_integration(self, create_rest_apigw, aws_client):
```
calvernaz, line 815: this was a complete removal or a move?

localstack/services/apigateway/helpers.py
```diff
@@ -658,41 +659,66 @@ def add_path_methods(rel_path: str, parts: List[str], parent_id=""):
             method_resource = create_method_resource(resource, field, field_schema)
             method_resource.request_parameters = method_integration.get("requestParameters")
             responses = field_schema.get("responses", {})
-            for status_code in responses:
+            for status_code, response in responses.items():
                 response_model = None
-                if model_schema := responses.get(status_code, {}).get("schema", {}):
+                if model_schema := response.get(
```
calvernaz, line 664: This clearly has improved on functionality but I find it hard to read. Let's take this example from swagger documentation.

```
{
  "200": {
    "description": "a pet to be returned",
    "schema": {
      "$ref": "#/definitions/Pet"
    }
  },
  "default": {
    "description": "Unexpected error",
    "schema": {
      "$ref": "#/definitions/ErrorModel"
    }
  }
}
```

The response object keys are **status codes** and the **default** key, so lookups were natural to read before, response-> status code -> schema. Now we destruct early (status_code, response) and that makes a lookup less intuitive and the status code variable is distantly far from where it's used again.

localstack/services/apigateway/provider.py
```diff
@@ -436,7 +444,7 @@ def put_method(
         # TODO: add missing validation? check order of validation as well
         moto_backend = apigw_models.apigateway_backends[context.account_id][context.region]
         moto_rest_api: MotoRestAPI = moto_backend.apis.get(rest_api_id)
-        if not moto_rest_api or not moto_rest_api.resources.get(resource_id):
+        if not moto_rest_api or not (moto_resource := moto_rest_api.resources.get(resource_id)):
```
calvernaz, line 447: probably not needed.

localstack/services/apigateway/patches.py
```diff
@@ -54,12 +54,10 @@ def apigateway_response_integrations(fn, self, request, *args, **kwargs):
 
         if self.method == "PUT":
             timeout_milliseconds = self._get_param("timeoutInMillis")
-            request_parameters = self._get_param("requestParameters") or {}
```
calvernaz, line 57: nice catch

localstack/services/apigateway/provider.py
```diff
@@ -508,6 +520,14 @@ def update_method(
         applicable_patch_operations = []
         modifying_auth_type = False
         modified_authorizer_id = False
+        had_req_params = (
+            isinstance(moto_method.request_parameters, dict)
+            and len(moto_method.request_parameters) > 0
+        )
+        had_req_models = (
+            isinstance(moto_method.request_models, dict) and len(moto_method.request_models) > 0
```
calvernaz, line 528: Non-empty collections are "truthy" I don't necessary think this makes it more readable.

localstack/services/apigateway/provider.py
```diff
@@ -248,7 +254,8 @@ def put_rest_api(self, context: RequestContext, request: PutRestApiRequest) -> R
         response = rest_api.to_dict()
         remove_empty_attributes_from_rest_api(response)
         store = get_apigateway_store(account_id=context.account_id, region=context.region)
-        store.rest_apis[request["restApiId"]].rest_api = response
+        rest_api_container = store.rest_apis[request["restApiId"]]
+        rest_api_container.rest_api = response
```
calvernaz, line 258: The "response" was already there, but I was scratching my head at this, we could rename response to rest_api_response? something more domain-specific.

calvernaz (commented):
Thanks @bentsku for tackling this I have a couple of comments on readability mostly but overall great job adding better support and a bunch of tests.

calvernaz:
One of the things I mentioned before and there is an issue to tackle that, is to create a better separation between versions of the specification we import. I think we have tackled the reference resolvers and we understand the AWS vendor nuances, but the import method logic has both specifications mixed, I think it's worth the effort

localstack/services/apigateway/provider.py
bentsku: Oh yeah, sorry. This could be removed, I suppose I was using the rest_api_container at some point but removed it, so the "split" stayed. I'll remove.

localstack/services/apigateway/provider.py
bentsku: I'm using the resource under by getting the `moto_http_method` (line 489)

tests/integration/apigateway/test_apigateway_basic.py
bentsku: This was move and migration to be AWS validated to `test_apigateway_lambda.py` file, under the name `test_lambda_aws_integration_with_request_template`

bentsku:
> One of the things I mentioned before and there is an issue to tackle that, is to create a better separation between versions of the specification we import. I think we have tackled the reference resolvers and we understand the AWS vendor nuances, but the import method logic has both specifications mixed, I think it's worth the effort

I agree! This PR was already pretty big, and having both logic in the same "importer" makes it harder to use, I agree. It would be worth tackling this at some point in a not too distant future!

localstack/services/apigateway/helpers.py
bentsku: I'm gonna look at this better tomorrow, but I was a bit confused by the `responses.get(status_code, {})` part, because the `status_code` would always be in `responses` as its part of its key? I'll look at the overall loop again better.

bentsku: I reworked this part of the importer, I've added more tests now and there were still missing fields and misplaced ones, especially `requestParameters` for `Method`.
I've also commented more the code, explained what we were creating and why, from which part of the imported file.

If you could review this part of the code again, that would be awesome! 🙏 

Also, the example you shared, I don't think we're supporting that at the moment, as we would try to create a `MethodResponse` with a `default` value for `response_code`. We will need to create an AWS validated test with this kind of spec.

bentsku:
Reconverting to draft for today, as I'm refactoring the tests to add them in their own file, and will fix the last failing tests in the meantime.

bentsku:
The PR is now ready for review, no more tests are failing. The snapshotting itself added 4k lines 😬

whummer (approved):
Fantastic set of changes @bentsku ! 🚀🚀 This is a massive change set and will be a huge boost in parity for API Gateway. 🥳 Big kudos for the huge coverage of new snapshot tests!

Added a few questions/suggestions, but nothing major that should block the merge.. 👍 

Regarding code structure, I feel like the `import_api_from_openapi_spec(..)` function is getting to a size where we could pull it out into a separate `api_imports.py` module (and potentially introduce some structure, separate functions, or even some class abstractions). Not for this PR, but to be considered for a future iteration 👍 

Also, opening the can of worms  🪱 here - but given that we're overwriting a large portion of the API methods already, this could soon become a candidate for decoupling entirely from moto.. (soon ™️ )

localstack/services/apigateway/helpers.py
```diff
@@ -551,13 +552,37 @@ def import_api_from_openapi_spec(
     store = get_apigateway_store(account_id=account_id, region=region)
     rest_api_container = store.rest_apis[rest_api.id]
 
+    def is_api_key_required(path_payload: dict) -> bool:
+        # TODO: consolidate and refactor with `create_authorizer`, duplicate logic for now
+        if "security" not in path_payload:
+            return False
+
+        security_schemes = path_payload.get("security")
```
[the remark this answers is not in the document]
bentsku: Good catch! I'll replace with `if security_schemes := path_payload.get("security"):` which will cover both cases!

localstack/services/apigateway/helpers.py
```diff
@@ -650,81 +678,214 @@ def add_path_methods(rel_path: str, parts: List[str], parent_id=""):
...
+                            "Ignoring unsupported requestParameters/requestModels location value for %s: %s",
+                            param_name,
+                            param_location,
+                        )
+                        continue
+                method_resource.request_models = request_models or None
+
+            # we check if there's a path parameter, AWS adds the requestParameter automatically
+            resource_name = parts[-1].strip("/")
+            if resource_name.startswith("{") and not resource_name.endswith("+}"):
```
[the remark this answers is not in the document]
bentsku: Wow, good catch, I was going to say I wasn't aware of those methods but I can see I'm the one who made them... thanks 👍

localstack/services/apigateway/helpers.py
```diff
@@ -650,81 +678,214 @@ def add_path_methods(rel_path: str, parts: List[str], parent_id=""):
...
+            for method_status_code, method_response in method_responses.items():
+                method_response_model = None
+                model_schema = None
+
+                # separating the two different versions, Swagger (2.0) and OpenAPI 3.0
+                if "schema" in method_response:  # this is Swagger
+                    model_schema = method_response.get("schema")
+                elif "content" in method_response:  # this is OpenAPI 3.0
+                    model_schema = (
+                        method_response["content"].get(APPLICATION_JSON, {}).get("schema")
```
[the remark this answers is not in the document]
bentsku: You're right, I'll do it when we create `Model` as well!

localstack/services/apigateway/helpers.py
```diff
@@ -650,81 +678,214 @@ def add_path_methods(rel_path: str, parts: List[str], parent_id=""):
...
-                name=name, description="", schema=model, content_type=APPLICATION_JSON
-            )
+    models = resolved_schema.get("definitions") or resolved_schema.get("components", {}).get(
+        "schemas", {}
+    )
+    schema_map = {}
+    for name, model in models.items():
+        schema = json.dumps(model)
+        # keep a map of the schema to retrieve the schema name for responseModels resolved $ref
+        schema_map[schema] = name
```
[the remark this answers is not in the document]
bentsku: I've finally made use of `canonical_json`, to do a lookup of `models` I would need to iterate over every key and check the found model is the same as the value. I basically need a reverse map, sadly, and the model is not hashable.

localstack/services/apigateway/provider.py
```diff
@@ -1500,22 +1629,42 @@ def remove_empty_attributes_from_method(method: Method) -> Method:
...
         method.pop("requestModels", None)
 
-    if not method.get("requestParameters"):
+    if method.get("requestParameters") is None:
         method.pop("requestParameters", None)
 
     return method
 
 
+def remove_empty_attributes_from_integration(integration: Integration):
```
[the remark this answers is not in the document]
bentsku: Right, very good idea! There as sometimes weird and subtle differences like the following:
```python

def remove_empty_attributes_from_integration(integration: Integration):
    if not integration:
        return integration

    if not integration.get("integrationResponses"):
        integration.pop("integrationResponses", None)

    if integration.get("requestParameters") is None:
        integration.pop("requestParameters", None)

    return integration
```

Note that if `integrationResponses` is not set, or is not truthy, I delete the field.
But for `requestParameters`, it needs to either not be in the dict, or explicitly be set to `None`...
Subtle difference, but annoying. 

But we could make use of such a method for other for sure!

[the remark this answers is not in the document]
whummer: Ah true! Sorry, misread this before - once we have a negative and once a positive check. Doesn't really pay off then - please ignore this comment for now.. 👍

calvernaz (approved):
Amazing job 🚀

PR #8382, opened 2023-05-27
remove thundra plugin

thrau:
This PR removes the Thundra plugin. Since, sadly, Thundra has ceased all operations with May 12th, we can safely remove the lambda plugin as part of a minor release.


joe4dev (approved):
Nice cleanup in the old Lambda provider 👍 
Agreed, even their website (https://www.thundra.io/) is offline now :(

The new Lambda provider aims to support 3rd party integrations through official [Lambda extensions](https://docs.aws.amazon.com/lambda/latest/dg/lambda-extensions.html).

whummer (approved):
Indeed a bit sad to see this go, but also good to reduce the surface area of the old Lambda provider. 👍  Thanks for cleaning this up.

PR #8384, opened 2023-05-28
Enhance bootstrap logic to allow CLI startup from within Docker

whummer:
This PR adds small enhancements to the bootstrap logic, to enable CLI startup (`localstack start`) from within Docker containers in different environments.

The starting point for this change is a Github Codespaces workspace that we're currently preparing for the AWS Builders Day workshop: https://github.com/localstack/localstack-workshop/blob/main/.devcontainer/devcontainer.json . The Codespaces environment is itself running in a Docker container (devcontainer), and with the current version of the CLI we're receiving this error:
```
$ localstack start
 ...
  File "/tmp/venv/lib/python3.10/site-packages/localstack/cli/localstack.py", line 168, in cmd_start
    bootstrap.start_infra_in_docker()
  File "/tmp/venv/lib/python3.10/site-packages/localstack/utils/bootstrap.py", line 591, in start_infra_in_docker
    container.truncate_log()
  File "/tmp/venv/lib/python3.10/site-packages/localstack/utils/bootstrap.py", line 458, in truncate_log
    with open(self.logfile, "wb") as fd:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/localstack/tmp/localstack_main_container.log'
```

Note that the bootstrap code is attempting to access `/var/lib/localstack`, which is the default directory for the LocalStack volume inside the container.

One of the key ingredients of the approach is to allow an override config `OVERRIDE_IN_DOCKER=0`, in addition to the already supported `OVERRIDE_IN_DOCKER=1`. It seems reasonable to not only allow overriding the Docker detection for false negatives, but also for false positives.  As far as I can see in the code (and our docs), this change should be backward-compatible (there is one reference in our [docs](https://docs.localstack.cloud/references/configuration/#miscellaneous) which needs to be updated).

(As a general note: currently the host/container switch is coupled to the "is in Docker" checks (which makes sense, as it covers 99% of use cases!) - but at some point we could consider splitting up the two, and make the distinction a bit more explicit.)

The second change in the PR is around mounting of the LocalStack volume dir. When executing in Docker, it can happen that the volume directory inside the container cannot be mounted into a new container, as it may not be available to the Docker engine. The following example is when executing our `test_start_cli_within_container(..)` test using Docker Desktop on macOS without passing the custom volume dir override:
```
...
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.42/containers/90af042d297b9254ad5cdf957937e434ef65c884030ba07e3bc9c5e0c64a133c/start: Internal Server Error ("b'Mounts denied: \nThe path /root/.cache/localstack/volume is not shared from the host and is not known to Docker.\nYou can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.\nSee https://docs.docker.com/desktop/mac for more info.'")
```

The change proposed in the PR adds a fallback logic which looks for this error message, and then. This is covered by an integration test - parametrizing the existing `test_start_cli_within_container(..)` test to start up with and without the custom volume dir configuration (it should ideally work out of the box without having to specify `LOCALSTACK_VOLUME_DIR` manually).

---
Summary of changes:
* add new `parse_boolean_env(..)` helper function to parse the value of boolean env variables (if specified)
* update the logic around `OVERRIDE_IN_DOCKER` to allow both, overriding with falsy and truthy values
* adjust the logic in `cli/localstack.py` to ensure `config.OVERRIDE_IN_DOCKER` and `config.is_in_docker` are both set to False, and then re-initialize the directories (somewhat similar to the `if host` logic switch further above in the same function)
* add a fallback logic for `Mounts denied` errors, to gracefully handle situations where the local volume directory (could be either on the host, or inside a countainer) cannot be mounted by Docker into the LocalStack container

With these changes, the container properly starts up in a Codespaces environment, after installing the CLI and doing a `localstack start` out of the box. 

/cc @giograno

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -162,6 +162,12 @@ def cmd_start(docker: bool, host: bool, no_banner: bool, detached: bool):
                 "If you would like to use --host, please reinstall localstack using `pip install localstack[runtime]`"
             )
     else:
+        # make sure to initialize the bootstrap environment and directories for the host (even if we're executing
+        # in Docker), to allow starting the container from within other containers (e.g., Github Codespaces).
+        config.OVERRIDE_IN_DOCKER = False
+        config.is_in_docker = False
+        config.dirs = config.init_directories()
+
         if detached:
             bootstrap.start_infra_in_docker_detached(console)
         else:
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -6,7 +6,7 @@ import subprocess
 import tempfile
 import time
 from dataclasses import dataclass
-from typing import Any, Dict, List, Mapping, Tuple, TypeVar
+from typing import Any, Dict, List, Mapping, Optional, Tuple, TypeVar, Union
 
 from localstack import constants
 from localstack.constants import (
@@ -197,18 +197,28 @@ class Directories:
         return str(self.__dict__)
 
 
-def eval_log_type(env_var_name):
-    """get the log type from environment variable"""
+def eval_log_type(env_var_name: str) -> Union[str, bool]:
+    """Get the log type from environment variable"""
     ls_log = os.environ.get(env_var_name, "").lower().strip()
     return ls_log if ls_log in LOG_LEVELS else False
 
 
-def is_env_true(env_var_name):
+def parse_boolean_env(env_var_name: str) -> Optional[bool]:
+    """Parse the value of the given env variable and return True/False, or None if it is not a boolean value."""
+    value = os.environ.get(env_var_name, "").lower().strip()
+    if value in TRUE_STRINGS:
+        return True
+    if value in FALSE_STRINGS:
+        return False
+    return None
+
+
+def is_env_true(env_var_name: str) -> bool:
     """Whether the given environment variable has a truthy value."""
     return os.environ.get(env_var_name, "").lower().strip() in TRUE_STRINGS
 
 
-def is_env_not_false(env_var_name):
+def is_env_not_false(env_var_name: str) -> bool:
     """Whether the given environment variable is empty or has a truthy value."""
     return os.environ.get(env_var_name, "").lower().strip() not in FALSE_STRINGS
 
@@ -233,7 +243,7 @@ def is_persistence_enabled() -> bool:
     return PERSISTENCE and dirs.data
 
 
-def is_linux():
+def is_linux() -> bool:
     return platform.system() == "Linux"
 
 
@@ -253,8 +263,8 @@ def in_docker():
     Returns True if running in a docker container, else False
     Ref. https://docs.docker.com/config/containers/runmetrics/#control-groups
     """
-    if OVERRIDE_IN_DOCKER:
-        return True
+    if OVERRIDE_IN_DOCKER is not None:
+        return OVERRIDE_IN_DOCKER
 
     # check some marker files that we create in our Dockerfiles
     for path in [
@@ -336,8 +346,8 @@ def in_docker():
     return False
 
 
-# whether the in_docker check should always return true
-OVERRIDE_IN_DOCKER = is_env_true("OVERRIDE_IN_DOCKER")
+# whether the in_docker check should always return True or False
+OVERRIDE_IN_DOCKER = parse_boolean_env("OVERRIDE_IN_DOCKER")
 
 is_in_docker = in_docker()
 is_in_linux = is_linux()
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -423,7 +423,7 @@ class LocalstackContainer:
         if isinstance(DOCKER_CLIENT, CmdDockerClient):
             DOCKER_CLIENT.default_run_outfile = self.logfile
 
-        def _run_container(_volume_mounts: List[SimpleVolumeBind]):
+        try:
             return DOCKER_CLIENT.run_container(
                 image_name=self.image_name,
                 stdin=self.stdin,
@@ -434,7 +434,7 @@ class LocalstackContainer:
                 tty=self.tty,
                 detach=self.detach,
                 command=self.command or None,
-                mount_volumes=_volume_mounts,
+                mount_volumes=self._get_mount_volumes(),
                 ports=self.ports,
                 env_vars=self.env_vars,
                 user=self.user,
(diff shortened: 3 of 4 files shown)


thrau (commented):
I can't quite seem to make it out from the PR description and I don't know much about codespaces. Could you clarify whether this is for either
1. starting LocalStack as docker container inside another docker container, so docker in docker? 
2. starting a LocalStack container on the host from within another container (presuming it can talk to the docker socket)
3. running LocalStack in host mode inside an already running docker container?

whummer:
Good point. 👍  Sorry, should have made that clearer. The PR is targeting a combination of points 1 and 2 above - whenever the CLI is run inside a container to spawn LS in a new container (either via DinD, or on the host via the Docker socket).

* For the change related to `OVERRIDE_IN_DOCKER=0` and updating `config.dirs` (identifying the LS volume dir for the environment in which the CLI process is running), this change would address both point 1 and point 2 above.

* For the change related to volume mounting (fallback attempting to spawn without volume mount in case of errors), this would be primarily for case 2 above (as this is the more problematic case, if the mount path is only local to the container and not known/exposed to the host).

More specifically, the Codespaces environment uses DinD (at least using the default `mcr.microsoft.com/devcontainers/universal` image), so this would target point 1 above. The issue around volume mount fallback (targeting point 2 above) surfaced during local testing. It can be replicated using this sequence of commands (as least on macOS):

```
$ docker run -v /var/run/docker.sock:/var/run/docker.sock -it python:3 bash
root@ca6c629167ed:/# pip install docker localstack
...
root@ca6c629167ed:/# DEBUG=1 localstack start
...
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.42/containers/ab2ec2a9f2856655babe4542d51b7fdd93a717faa399b8bfd850861c8e43455b/start: Internal Server Error ("b'Mounts denied: \nThe path /root/.cache/localstack/volume is not shared from the host and is not known to Docker.\nYou can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.\nSee https://docs.docker.com/desktop/mac for more info.'")
```

localstack/cli/localstack.py
```diff
@@ -162,6 +162,12 @@ def cmd_start(docker: bool, host: bool, no_banner: bool, detached: bool):
                 "If you would like to use --host, please reinstall localstack using `pip install localstack[runtime]`"
             )
     else:
+        # make sure to initialize the bootstrap environment and directories for the host (even if we're executing
+        # in Docker), to allow starting the container from within other containers (e.g., Github Codespaces).
+        config.OVERRIDE_IN_DOCKER = False
+        config.is_in_docker = False
+        config.dirs = config.init_directories()
```
thrau, line 169: i had to think about this change a bit but i think it's not problematic since it doesn't change the behavior of the CLI for users currently, other than that we're doing another call to `config.init_directories()`, but this is side-effect free so we should be good.

localstack/utils/bootstrap.py
```diff
@@ -445,6 +445,23 @@ def run(self):
...
+                if "Mounts denied" in err_str:
+                    # This can happen if Docker doesn't have access to the LocalStack volume directory.
+                    # We remove the volume dir mount from the list, and then attempt to run the container again.
+                    LOG.warning(
+                        "Error mounting volumes in Docker - attempting again without LocalStack volume mount"
+                    )
+                    volume_mounts = [vol for vol in volume_mounts if vol[1] != DEFAULT_VOLUME_DIR]
+                    DOCKER_CLIENT.remove_container(self.name, force=True)
+                    return _run_container(volume_mounts)
+                raise
```
thrau, line 464: would be great if we could get a second pair of :eyes: here from @dfangl

thrau (approved):
i had to mull it over a bit but i think it looks good.

the volume mounting fallback i'm not super confident about though. i'm generally not in favor of silent fallback logic of this kind. to me, missing permissions is something that the user should fix and not something that localstack should be ignoring. if people set up a localstack volume directory, and then localstack just starts without mounting the volume, then this misleads users into believing the volume was mounted correctly. i don't really have a good alternative solution for this in the docker-in-docker case, which is i guess what this is targeting, so :shrug: 

it would be great to get a final sign-off from @dfangl on that one though.

localstack/config.py
```diff
@@ -197,18 +197,28 @@ def __str__(self):
...
+    return None
+
+
+def is_env_true(env_var_name: str) -> bool:
     """Whether the given environment variable has a truthy value."""
     return os.environ.get(env_var_name, "").lower().strip() in TRUE_STRINGS
 
 
-def is_env_not_false(env_var_name):
+def is_env_not_false(env_var_name: str) -> bool:
```
dfangl, line 221: Love the type hints!

localstack/utils/bootstrap.py
dfangl: I am also not a fan of this, more or less silent, automatic fallback here. While we will see there is an error if the customers send us the logs due to the warning (the warning should definitely include the error itself here!), the fallback might not be clear to the user as the warning is easily missed. This can lead to increased support loads over a immediate failure with (ideally clear) error messages.

In my opinion, for such a case, we should print the complete error returned by the docker daemon (the whole "this directory is not shared with the VM" thing), and abort. As a user, I would like it better for the software to fail hard at the first time with a clear error message, than to get unexpected behavior later (no persistence working for example), after the warning is already burried in the logs.

tests/bootstrap/test_cli.py
```diff
@@ -149,7 +149,9 @@ def test_container_starts_non_root(self, runner, monkeypatch, container_client):
         output = container_client.exec_in_container(config.MAIN_CONTAINER_NAME, ["ps", "-fu", user])
         assert "localstack-supervisor" in to_str(output[0])
 
-    def test_start_cli_within_container(self, runner, container_client):
+    @pytest.mark.parametrize("custom_volume_dir", [True, False])
+    def test_start_cli_within_container(self, runner, container_client, custom_volume_dir):
+        env_vars = {"LOCALSTACK_VOLUME_DIR": "/tmp/ls-volume"} if custom_volume_dir else {}
```
dfangl, line 154: Could we use the `tmp_path` fixture here, instead of hardcoding the path?

dfangl (commented):
I am also no fan of the automatic fallback, as the warning is easily missed (and the actual error is missing from the warning and therefore not clear!).

I think an immediate failure would actually be more user friendly here, as it gives the chance to fix the issue and have no unexpected problems (like missing persistence) down the line.

localstack/utils/bootstrap.py
whummer: Fair points! Will revert these changes for now (the other part of the PR related to `config.dirs`/`OVERRIDE_IN_DOCKER` is the more critical one for now anyway), we can see over time if the volume mount permissions actually cause issues with our users. 

On second thought, agree with both of you that silently not mounting a LocalStack volume dir may be problematic. 👍 In the particular example above, it can be easily worked around by setting `LOCALSTACK_VOLUME_DIR`, e.g.:
```
$ docker run -v /var/run/docker.sock:/var/run/docker.sock -it python:3 bash
root@ca6c629167ed:/# pip install docker localstack
...
root@ca6c629167ed:/# LOCALSTACK_VOLUME_DIR=/tmp DEBUG=1 localstack start
```

PR #8389, opened 2023-05-29
[minor] Fix extraction of Arn in KMS Key CloudFormation resource

whummer:
Small change to fix extraction of `Arn` attribute in KMS Key CloudFormation resources.

The starting point for this change was an issue with the [`copilotlocal`](https://github.com/localstack/copilot-cli-local) CLI, which was raising seg faults:
```
$ copilotlocal app init
...
$ copilotlocal env init
...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x102ebd0ec]

goroutine 1 [running]:
github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack.ToAppRegionalResources(0x14000139438)
	github.com/aws/copilot-cli/internal/pkg/deploy/cloudformation/stack/app.go:237 +0xdc
...
```

The root cause for this that the stack outputs of the stack created by Copilot contains an output without `OutputValue` attribute (the value should be included in the output below):
```
$ awslocal cloudformation describe-stacks --stack-name ... 
...
            "Outputs": [
                {
                    "OutputKey": "KMSKeyARN",
                    "Description": "KMS Key used by CodePipeline for encrypting artifacts.",
                    "ExportName": "hosted-ArtifactKey"
                },
...
```

The PR extends one of the existing tests to include a `!GetAtt KMSKey.Arn` stack output, and also adds a snapshot test to verify we have the same behavior as AWS.

---

The PR also contains a small fix in `create_stack_instances(..)`, passing the region name to `find_stack(..)`. Turns out this is required - otherwise the `copilotlocal` deployment raises the following error, when deploying the stack against a non-default region:
```
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateStack operation: Stack named "sset-hosted-infrastructure-000000000000" already exists with status "CREATE_COMPLETE"
```

### DIFF
--- a/localstack/services/cloudformation/models/kms.py
+++ b/localstack/services/cloudformation/models/kms.py
@@ -13,7 +13,7 @@ class KMSKey(GenericBaseModel):
 
     def get_cfn_attribute(self, attribute_name):
         if attribute_name == "Arn":
-            return self.props.get("KeyMetadata", {}).get("Arn")
+            return arns.kms_key_arn(self.physical_resource_id)
         return super(KMSKey, self).get_cfn_attribute(attribute_name)
 
     def fetch_state(self, stack_name, resources):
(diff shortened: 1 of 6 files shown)


pinzon (approved):
LGTM :rocket:

Issue #8392, opened 2023-05-30
bug: Complete multipart upload for non-existing uploads does not produce the NoSuchUpload error code.


real-az:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The Complete Multipart Upload operation generates the **InternalError** error for non-existing uploads.

### Expected Behavior

The Complete Multipart Upload operation generates the **NoSuchUpload** error for non-existing uploads.

### How are you starting LocalStack?

With a `docker run` command

### Steps To Reproduce

    docker run -it --rm -p 4566:4566 localstack/localstack

    aws s3api create-bucket --bucket test --endpoint-url http://localhost:4566 --create-bucket-configuration LocationConstraint=us-west-2

    aws s3api complete-multipart-upload --bucket test --endpoint-url http://localhost:4566 --key test --upload-id test

### Environment

```markdown
- OS: Windows 11
- LocalStack: latest
```

### Anything else?

The Abort Multipart Upload operation works correctly in LocalStack.

The official documentation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html
Amazon S3 throws the correct **NoSuchUpload** error for both Abort and Complete operations.

PR #8396, opened 2023-05-30
fix s3 complete_multipart and upload_part internal error 

bentsku:
This issue has been reported in #8392, moto would raise an `InternalError` when trying to complete a multipart upload that wasn't initiated, because it was directly accessing the `multiparts` bucket dict field with the `upload_id` as the key, which would raise a `KeyError`. It had the same issue in `upload_part`, but this was correct in `abort_multipart_upload`.

Added AWS validated tests, and fixed it in the provider for a quick fix. Ultimately, this should be fixed upstream.

_fixes #8392_


thrau (approved):
LGTM!

PR #6314, opened 2022-06-21
Make hashable list/dict util classes immutable

whummer:
Make hashable list/dict utility classes immutable. This is a great point raised by @dfangl - we should ensure that the hashable collections represent a frozen/immutable copy and cannot be modified under the covers, to avoid unexpected side effects further down the line. 👍

Issue #5758, opened 2022-03-28
bug: DNS server inaccessible from host with LocalStack CLI


MarcelStranak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Docker flags for port 53 are not provided when running the `DEBUG=1  localstack start`

### Expected Behavior

Should be same as when started with `DEBUG=1   DOCKER_FLAGS='-p 127.0.0.1:53:53 -p 127.0.0.1:53:53/udp' localstack start`

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

not working 
`DEBUG=1  localstack start`

working
`DEBUG=1   DOCKER_FLAGS='-p 127.0.0.1:53:53 -p 127.0.0.1:53:53/udp' localstack start`

### Environment

```markdown
- OS: Ubuntu 
- LocalStack: latest
```

MarcelStranak:
This was already resolved.

PR #8402, opened 2023-05-31
fix multi-protocol port mapping

alexrashed:
This PR addresses https://github.com/localstack/localstack/issues/5758.
It fixes several issues concerning the protocol handling in the `PortMapping` class, which was introduced in #6314.
Previously, the existence checks when adding new mappings were not considering the protocol. The fixes in this PR now allow adding both, a UDP and TCP port mapping which has the same range. This in turn fixes an issue in the LocalStack Pro CLI where the DNS TCP port was added to the port mapping, but the UDP port was suppressed.

/cc @whummer

### DIFF
--- a/localstack/utils/container_utils/container_client.py
+++ b/localstack/utils/container_utils/container_client.py
@@ -134,6 +134,8 @@ class Ulimit:
 
 # defines the type for port mappings (source->target port range)
 PortRange = Union[List, HashableList]
+# defines the protocol for a port range ("tcp" or "udp")
+PortProtocol = str
 
 
 def isinstance_union(obj, class_or_tuple):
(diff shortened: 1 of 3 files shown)


localstack/utils/container_utils/container_client.py
```diff
@@ -184,92 +188,109 @@ def add(
...
             ]
 
         items = [item for k, v in self.mappings.items() for item in entry(k, v)]
         return dict(items)
 
-    def contains(self, port: int) -> bool:
-        for from_range, to_range in self.mappings.items():
-            if self.in_range(port, from_range):
-                return True
+    def contains(self, port: int, protocol: str = "tcp") -> bool:
```
dfangl, line 277:
```suggestion
    def contains(self, port: int, protocol: PortProtocol = "tcp") -> bool:
```
Shouldn't this also use the port protocol type?

dfangl (approved):
LGTM, thanks for fixing this!

PR #8407, opened 2023-05-31
show error on frozen attempt to update localstack-cli

alexrashed:
This PR fixes a confusing error message when using the `localstack update` command to update the CLI (either with `all` or `localstack-cli`) when using our LocalStack CLI frozen binaries built with PyInstaller ([localstack/localstack-cli](https://github.com/localstack/localstack-cli)):
```
$ localstack update localstack-cli
─────────── Updating LocalStack CLI ───────────
⠦ Updating LocalStack CLI...Usage: localstack [OPTIONS] COMMAND [ARGS]...
Try 'localstack --help' for help.

Error: No such option: -m
✖ LocalStack CLI update failed
```
With this PR we explicitly detect that we are running in a frozen environment, print a warning and exit with a non-zero exit code:
```
$ localstack update all
✖ The LocalStack CLI can only update itself if installed via PIP. Please follow the instructions on https://docs.localstack.cloud/ to update your CLI.
```

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -370,6 +370,16 @@ def cmd_update_all(ctx):
 @localstack_update.command(name="localstack-cli", help="Update LocalStack CLI tools")
 @publish_invocation
 def cmd_update_localstack_cli():
+    # if is_frozen_bundle():
+    if True:
+        # "update" can only be performed if running from source / in a non-frozen interpreter
+        console.print(
+            ":heavy_multiplication_x: The LocalStack CLI can only update itself if installed via PIP. "
+            "Please follow the instructions on https://docs.localstack.cloud/ to update your CLI.",
+            style="bold red",
+        )
+        sys.exit(1)
+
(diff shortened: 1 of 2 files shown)


localstack/cli/localstack.py
```diff
@@ -370,6 +370,16 @@ def cmd_update_all(ctx):
 @localstack_update.command(name="localstack-cli", help="Update LocalStack CLI tools")
 @publish_invocation
 def cmd_update_localstack_cli():
+    # if is_frozen_bundle():
+    if True:
```
thrau, line 374: I guess this is still leftover code from debugging locally? :thinking:

thrau (changes requested):
Nice addition! Looks like there's still an error in the condition to trigger the code

thrau (approved):
This will definitely help in making it more clear to users how to update the CLI when using the binary! Looking forward to introducing a self-updating mechanism in the future ;-)

PR #8412, opened 2023-06-01
Managed docker network

simonrw:
This PR allows the user to configure a "managed" docker network.

Putting containers on the same docker network means more straightforward configuration. The containers can make requests to each other using their container names. LocalStack currently adds configuration to allow the user to specify which docker network to add created containers to, for services such as Lambda or ECS. This requires:

1. the user to be aware of docker networking, and
2. for them to have set up a network in advance.

Instead, let's make LocalStack the curator of this docker network. On startup, create a named network (if it doesn't exist) and add all containers to it for the user. This means:

1. the user does not _have_ to know about docker networking, and
2. they don't have to perform any setup.

Users using docker compose will see the example compose file in the repo and replicate the behaviour in their own compose files.

## Implementation

We use the existing configuration variables to control the networks of created containers, and add new configuration to override their behaviour. When the CLI starts up, we create the network if it doesn't exist.

## Questions

* Should we delete the network after the container stops? Probably not, but this is untidy
* Should we make this behaviour the default in the compose files? I.e. uncomment the `USE_MANAGED_DOCKER_NETWORK` lines?

### DIFF
--- a/docker-compose-pro.yml
+++ b/docker-compose-pro.yml
@@ -15,6 +15,7 @@ services:
       - PERSISTENCE=${PERSISTENCE-}
       - LOCALSTACK_API_KEY=${LOCALSTACK_API_KEY-}  # required for Pro
       - DOCKER_HOST=unix:///var/run/docker.sock
+      # - USE_MANAGED_DOCKER_NETWORK=1  # allow LocalStack to manage networking
     volumes:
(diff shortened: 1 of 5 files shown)


simonrw:
Invalidated by https://github.com/localstack/localstack/pull/8772

PR #8417, opened 2023-06-01
update SQS QueryAPI to use new aws client factory

thrau:
Uses the new `connect_to` construct instead of `aws_stack` to connect to SQS in the SQS Query API. I'd like to use this later to try out #7396  /cc @viren-nadkarni @dfangl 

I also removed a couple of deprecated tests.


baermat (approved):
LGTM :+1:
should we specifically state the test-names in the PR description for traceability? No idea if we will ever need that though.

PR #8419, opened 2023-06-01
replace the method wait_until with boto3 waiters in cfn fixture

pinzon:
This PR replace the usage of the wait_until method for boto3 waiters due to  the better behavior on stacks that take too much time to deploy.


simonrw (approved):
Looks nice, glad we can use waiters in this fixture now!

PR #8425, opened 2023-06-02
add BodyS3Location property to AWS::ApiGateway::RestApi

bentsku:
Depends on #8371 

Implementing the `BodyS3Location` property to `AWS::ApiGateway::RestApi` as requested by a customer.

This property allows Cloudformation to directly fetch the "Body" property from an S3 bucket.

However, I am not sure how to do negative testing with Cloudformation, so for now any S3 exception is going to bubble up? 

Small note: I've also removed a line from the API Gateway importer because it was erasing tags after updating a RestAPI, which is not the default mode anymore. 

Resource:
https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-s3location.html

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -241,6 +241,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
 
     @handler("PutRestApi", expand=False)
     def put_rest_api(self, context: RequestContext, request: PutRestApiRequest) -> RestApi:
+        # TODO: take into account the mode: overwrite or merge
+        # the default is now `merge`, but we are removing everything
         body_data = request["body"].read()
         rest_api = get_moto_rest_api(context, request["restApiId"])
 
--- a/localstack/services/cloudformation/models/apigateway.py
+++ b/localstack/services/cloudformation/models/apigateway.py
@@ -1,6 +1,7 @@
 import json
 from urllib.parse import urlparse
 
+from localstack.aws.connect import connect_to
 from localstack.services.cloudformation.deployment_utils import (
     generate_default_name,
     lambda_keys_to_lower,
@@ -181,7 +182,7 @@ class GatewayRestAPI(GenericBaseModel):
                         get_obj_kwargs["VersionId"] = version_id
 
                     # what is the approach when client call fail? Do we bubble it up?
-                    s3_client = aws_stack.connect_to_service("s3")
+                    s3_client = connect_to().s3
                     get_obj_req = s3_client.get_object(
                         Bucket=s3_body_location.get("Bucket"),
                         Key=s3_body_location.get("Key"),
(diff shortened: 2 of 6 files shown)


localstack/services/cloudformation/models/apigateway.py
```diff
@@ -166,12 +167,45 @@ def _create(resource_id, resources, resource_type, func, stack_name):
...
+
+                if body:
+                    body = json.dumps(body) if isinstance(body, dict) else body
+                else:
+                    get_obj_kwargs = {}
+                    if version_id := s3_body_location.get("Version"):
+                        get_obj_kwargs["VersionId"] = version_id
+
+                    # what is the approach when client call fail? Do we bubble it up?
+                    s3_client = aws_stack.connect_to_service("s3")
```
dominikschubert, line 184: minor nit
```suggestion
                    s3_client = connect_to().s3
```

localstack/services/cloudformation/models/apigateway.py
```diff
@@ -166,12 +167,45 @@ def _create(resource_id, resources, resource_type, func, stack_name):
...
+                        Bucket=s3_body_location.get("Bucket"),
+                        Key=s3_body_location.get("Key"),
+                        **get_obj_kwargs,
+                    )
+                    if etag := s3_body_location.get("ETag"):
+                        if etag != get_obj_req["ETag"]:
+                            # TODO: validate the exception message (how?)
+                            raise Exception(
+                                "The ETag provided for the S3BodyLocation does not match the S3 Object"
+                            )
```
dominikschubert, line 195: It should lead to a `CREATE_FAILED` state and checking the stack events should indicate why the resource failed to deploy. Currently though this is not yet supported very well, so for now I'd say it's ok if this isn't covered specifically via a test :+1:

dominikschubert (approved):
LGTM! Thanks for extending this!

dominikschubert (approved):
LGTM! Thanks for extending this!

localstack/services/cloudformation/models/apigateway.py
bentsku: Thanks, perfect, it makes more sense!

localstack/services/cloudformation/models/apigateway.py
```diff
@@ -166,12 +168,45 @@ def _create(resource_id, resources, resource_type, func, stack_name):
             result = client.create_rest_api(**kwargs)
 
             body = props.get("Body")
-            if body:
+            s3_body_location = props.get("BodyS3Location", {})
```
calvernaz, line 171: Just because an empty dictionary is also _falsy_

```suggestion
            s3_body_location = props.get("BodyS3Location")
```

localstack/services/apigateway/provider.py
```diff
@@ -241,6 +241,8 @@ def update_rest_api(
 
     @handler("PutRestApi", expand=False)
     def put_rest_api(self, context: RequestContext, request: PutRestApiRequest) -> RestApi:
+        # TODO: take into account the mode: overwrite or merge
```
calvernaz, line 244: yep, no "mode" means "merge" - not necessarily a fix for now, but we could return a non-supported error if the mode is overwrite

calvernaz (approved):
LGTM! Thanks @bentsku

localstack/services/apigateway/provider.py
bentsku: Right, but right now we are not merging but always overwriting I believe?

PR #8426, opened 2023-06-02
fix package incompatibilities in pipeline

alexrashed:
It seems like we have issue with the integration tests against our pro extensions due to a conflict with packages being installed on the CI system before running the tests.
This seems to be related to the package repos on the GitHub runners (it does not happen on other, private runners).
Unfortunately, apt is not automatically resolving the conflict by downgrading postgres to a matching version without explicitly stating it.
I cancelled the circle pipeline runs, because this only touches the GitHub workflow.
I'm happy for any feedback, this apt call caused quite a lot of issues in the past.

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -148,7 +148,7 @@ jobs:
       - name: Install OS packages
         run: |
           sudo apt-get update
-          sudo apt-get install -y --allow-downgrades libsasl2-dev jq postgresql-14=14.8* postgresql-client postgresql-plpython3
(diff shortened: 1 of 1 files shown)


dfangl (approved):
Thanks for fixing this!

simonrw:
As you mentioned, this problem has come up before, and the git churn on that line is quite significant. Do we have a plan to work around this? Could we run our own apt repo just with our testing packages?

alexrashed:
That sounds like an overkill to be honest. I was more aiming towards a more robust set of installation commands which is more resilient when it comes to certain temporary conflicts...

PR #8427, opened 2023-06-02
SNS: Minor cleanup

viren-nadkarni:
This PR does some minor cleanup for SNS by removing redudant handler definitions and exception handling


localstack/services/sns/provider.py
```diff
@@ -113,12 +110,7 @@ def _get_topic(arn: str, context: RequestContext) -> Topic:
     def get_topic_attributes(
         self, context: RequestContext, topic_arn: topicARN
     ) -> GetTopicAttributesResponse:
-        try:
-            moto_response: GetTopicAttributesResponse = call_moto(context)
-        except CommonServiceException as exc:
-            if exc.code == "NotFound":
-                raise NotFoundException("Topic does not exist")
-            raise
+        moto_response: GetTopicAttributesResponse = call_moto(context)
```
bentsku, line 113: question: why are we removing a fix to then add a `skip_snapshot_verify` path? This was done to counter the specific issue in the test? ` # This undocumented field is currently not returned by Moto`

localstack/services/sns/provider.py
```diff
@@ -308,25 +300,6 @@ def list_tags_for_resource(
...
-    def create_platform_application(
-        self, context: RequestContext, name: String, platform: String, attributes: MapStringToString
-    ) -> CreatePlatformApplicationResponse:
-        # TODO: validate platform
-        # see https://docs.aws.amazon.com/cli/latest/reference/sns/create-platform-application.html
-        # list of possible values: ADM, Baidu, APNS, APNS_SANDBOX, GCM, MPNS, WNS
-        # each platform has a specific way to handle credentials
-        # this can also be used for dispatching message to the right platform
-        moto_response: CreatePlatformApplicationResponse = call_moto(context)
-        return moto_response
```
bentsku, line 328: I'd like to keep this operation or at the minimum the `TODO`, as it could a the next big work to do on SNS, as our emulation level there is very weak.

bentsku (commented):
Thanks for cleaning up! I'm not sure I understand the removal of a fix to then add a new path to the `skip_snapshot_verify`, was there a specific reason for this?

localstack/services/sns/provider.py
viren-nadkarni: Added back :+1:

localstack/services/sns/provider.py
viren-nadkarni: This is now implemented in Moto upstream, so the ignore has now been removed

https://github.com/getmoto/moto/pull/6465

bentsku: Awesome, thanks for fixing it upstream! that's better than my solution 😄

bentsku (approved):
LGTM! Thanks a lot for cleaning up, especially for the fix upstream! 🙏 
edit: test failure seems to be unrelated, I think it's good to merge 😄

PR #8429, opened 2023-06-02
replace events client in message publisher

pinzon:
this small PR updates the events client in utility message_publisher.py


dfangl (approved):
LGTM! Just rebased it.

PR #8431, opened 2023-06-03
Share ffmpeg package between services

simonrw:
This package will be used by `mediaconvert` in the ext repo, and so we should share the package location.

This change moves the ffmpeg package to `localstack.packages` and updates the `transcribe` service accordingly.

**Question**: the package is currently directly loaded by the service, but is exposed via the plugins system. Should we update the service to load ffmpeg as a plugin? No code runs at import time, and the plugin only defines two classes, so the runtime overhead of importing the module when `transcribe` is loaded is probably negligable, suggesting that it's fine as it is.


viren-nadkarni:
> Question: the package is currently directly loaded by the service...

Not sure I understand the internals of the plugin system, maybe @alexrashed can give his insights?

viren-nadkarni (approved):

alexrashed (approved):
The changes are looking perfectly fine. 🚀 
And to answer your question of how the service should access the package:
The services actually should _not_ load the packages via the plugin mechanism (they are quite tightly coupled to the actual package being installed, and there just was not a need yet). But `lpm` uses the plugin mechanism to discover all packages that can be installed:
https://github.com/localstack/localstack/blob/284b5e2f3c2a982534605972df2ddb76c9da1a0d/localstack/cli/lpm.py#L114-L135

simonrw:
Ok so the service will import as per usual, but we make a plugin for `lpm` to pick up :+1: That's great, thanks both

PR #8437, opened 2023-06-05
fix validating APIGW API key when apiKeySource is AUTHORIZER

bentsku:
When looking at an OpenAPI file, I've spotted an issue where we didn't validate what the `apiKeySource` was before validating the key (if the source is set to `AUTHORIZER`, as authorizers are only mocked in community, it would always fail before). 

I've migrated one integration test to be AWS validated, there still one left but it'll be migrated when we pushed the migration further for `UsagePlan` and `ApiKey`. 

Also fixed a parity issue where we wouldn't validate if the key was enabled.

### DIFF
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -6,6 +6,7 @@ from botocore.exceptions import ClientError
 from jsonschema import ValidationError, validate
 from requests.models import Response
 
+from localstack.aws.connect import connect_to
 from localstack.constants import APPLICATION_JSON
 from localstack.services.apigateway import helpers
 from localstack.services.apigateway.context import ApiInvocationContext
@@ -149,7 +150,10 @@ def authorize_invocation(invocation_context: ApiInvocationContext):
 def validate_api_key(api_key: str, invocation_context: ApiInvocationContext):
 
     usage_plan_ids = []
-    client = aws_stack.connect_to_service("apigateway", region_name=invocation_context.region_name)
+    client = connect_to(
+        aws_access_key_id=invocation_context.account_id, region_name=invocation_context.region_name
+    ).apigateway
+
     usage_plans = client.get_usage_plans()
     for item in usage_plans.get("items", []):
         api_stages = item.get("apiStages", [])
@@ -176,8 +180,11 @@ def is_api_key_valid(invocation_context: ApiInvocationContext) -> bool:
     client = aws_stack.connect_to_service("apigateway")
     rest_api = client.get_rest_api(restApiId=invocation_context.api_id)
 
-    # It means it's the authorizer job to return the API key, and as we're only mocking it, it can't
     if rest_api.get("apiKeySource") != "HEADER":
+        # When the apiKeySource is set to AUTHORIZER, the authorizer is supposed to return the API key as a field
+        # `usageIdentifierKey`
+        # see https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html
+        # Authorizers are only mocked, so we can't validate the key. Return True in that case
         return True
 
     api_key = invocation_context.headers.get("X-API-Key")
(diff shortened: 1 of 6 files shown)


localstack/services/apigateway/invocations.py
```diff
@@ -146,36 +146,45 @@ def authorize_invocation(invocation_context: ApiInvocationContext):
...
+                return api_key.get("enabled") in ("true", True)
 
     return False
 
 
-def is_api_key_valid(is_api_key_required: bool, headers: Dict[str, str], stage: str):
-    if not is_api_key_required:
+def is_api_key_valid(invocation_context: ApiInvocationContext) -> bool:
+    # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-key-source.html
+    client = aws_stack.connect_to_service("apigateway")
```
whummer, line 176: Nit: could be replaced with the new `connect_to().apigateway` over time :+1:

whummer (approved):
Great set of changes @bentsku ! :rocket: Nice to see the improved parity around API keys and authorizers

localstack/services/apigateway/invocations.py
```diff
@@ -234,7 +243,8 @@ def invoke_rest_api(invocation_context: ApiInvocationContext):
         return make_error_response("Invalid request body", 400)
 
     api_key_required = resource.get("resourceMethods", {}).get(method, {}).get("apiKeyRequired")
-    if not is_api_key_valid(api_key_required, headers, invocation_context.stage):
+    if api_key_required and not is_api_key_valid(invocation_context):
```
calvernaz, line 246: According to https://repost.aws/knowledge-center/api-gateway-troubleshoot-403-forbidden, the message would be "Invalid API Key identifier specified"

localstack/services/apigateway/invocations.py
```diff
@@ -146,36 +146,45 @@ def authorize_invocation(invocation_context: ApiInvocationContext):
...
 
 
-def is_api_key_valid(is_api_key_required: bool, headers: Dict[str, str], stage: str):
-    if not is_api_key_required:
+def is_api_key_valid(invocation_context: ApiInvocationContext) -> bool:
+    # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-key-source.html
+    client = aws_stack.connect_to_service("apigateway")
+    rest_api = client.get_rest_api(restApiId=invocation_context.api_id)
+
+    # It means it's the authorizer job to return the API key, and as we're only mocking it, it can't
```
calvernaz, line 179: can you clarify this comment? Is it related to the line above where we check the api key source or a generic comment about the method?

calvernaz (approved):
LGTM, thanks for doing this!

localstack/services/apigateway/invocations.py
bentsku: Yes, that is weird because from my tests in AWS, I always only got a json message with `{"message": "Forbidden"}` and nothing else

localstack/services/apigateway/invocations.py
bentsku: I'll clarify in the PR as well.
Basically, when the `apiKeySource` is set to `AUTHORIZER`, the authorizer is supposed to return the api key as a field (`usageIdentifierKey`): see https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html

Right now I don't have a setup to verify what the actual behaviour is in AWS, but I expected them to validate the key from the authorizer context only after the authorizer has ran if the key returned is valid.

However, as the authorizers are only mocked in community, there is now way to properly validate that, so we always return that the key is valid for simplicity. It's up to us to patch this method in -ext if we want to implement it.

localstack/services/apigateway/invocations.py
whummer: Good point! +1 that in-memory actually would make sense here. Indeed we could have the same issue here, as the region may not be correctly set in the request context, if I understand correctly.. :grimacing:

PR #8438, opened 2023-06-05
Update lambda init to v0.1.18-pre

dominikschubert:
Update lambda init version to https://github.com/localstack/lambda-runtime-init/releases/tag/v0.1.18-pre

This will remove some confusing exception logging where raised exceptions during a function invocation were also logged in the same way that an init error was logged.


dfangl (approved):

joe4dev (approved):
LGTM 👍

PR #8439, opened 2023-06-05
Prefix dynamodb streams shard iterator with stream arn

dfangl:
## Motivation
As seen here: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_GetShardIterator.html, the shard iterator for dynamodbstreams differs from the kinesis one in one important aspect, namely the prefix of the stream arn followed by `|<int>|` and then an iterator in a similar format like kinesis data streams.

## Changes
* Add tests specifically asserting on the sharditerator format
* Fix shard iterator format

## Future challenges
It is quite unclear to me, and I did not find any information on that, what the integer between the pipes `|<int>|` is supposed to represent. I got the values 1 and 2 back from AWS, not sure what they are exactly denoting. As long as this is unclear, and since it does not seem to make a difference for now, I hardcoded it to 1 with a comment for it.

### DIFF
--- a/localstack/services/dynamodbstreams/provider.py
+++ b/localstack/services/dynamodbstreams/provider.py
@@ -102,7 +102,7 @@ class DynamoDBStreamsProvider(DynamodbstreamsApi, ServiceLifecycleHook):
             raise ExpiredIteratorException("Shard iterator has expired")
         result = {
             "Records": [],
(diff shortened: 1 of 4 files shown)


viren-nadkarni (approved):
LGTM :+1:

PR #8440, opened 2023-06-05
CFn: simplify deploy function handling

simonrw:
The CloudFormation `GenericBaseModel` classes allowed use of an arbitrary function to deploy the resource. The function signature was:

```
def create(resource_id, resources, resource_type, func, stack_name):
    ...
```

When deploying a single resource:
* the list of resources is redundant
* the resource type is already known
* the `func` argument is not relevant

This PR changes the function signature to `def create(logical_resource_id: str, resource: dict, stack_name: str)` to only supply the relevant information.

To prevent `-ext` breakage, if attempting to call this function raises a type error (because the signature is still the old signature) then we fall back to using the old function signature.

## Motivation

A PR integrating the new style resource providers depends on this change, since I need to call `execute_resource_action` but don't have access to the complete list of resources in a new style provider. Updating `execute_resource_action` will happen once we've removed the fallback old behaviour from all code.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -73,12 +73,10 @@ ResourceDefinition = dict[str, ResourceProp]
 
 class FuncDetailsValue(TypedDict):
     # Callable here takes the arguments:
-    # - resource_id
-    # - resources
-    # - resource_type
-    # - func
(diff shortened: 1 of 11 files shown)


dominikschubert (approved):
LGTM! Nice first step in simplifying the signatures here :+1:

PR #8445, opened 2023-06-05
fix apigw import default authorizer, apiKeySource and ANY method

bentsku:
This PR is building on #8437, we were not properly setting the `apiKeySource` of the REST API when importing the file.

Also, we were not making use of default authorizer defined as a top level property. 

I also fixed passing `x-amazon-apigateway-any-method` as an HTTP method, but properly replaced it by  "ANY" like AWS does. I think we could remove any `x-amazon-apigateway-any-method` in the invocation code now, as it has been properly replaced.

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -72,6 +72,34 @@ TAG_KEY_CUSTOM_ID = "_custom_id_"
 EMPTY_MODEL = "Empty"
 ERROR_MODEL = "Error"
 
+
+# TODO: we could actually parse the schema to get TypedDicts with the proper schema/types for each properties
+class OpenAPIExt:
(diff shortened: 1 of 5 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -72,6 +72,34 @@
 EMPTY_MODEL = "Empty"
 ERROR_MODEL = "Error"
 
+
+# TODO: we could actually parse the schema to get TypedDicts with the proper schema/types for each properties
+class OpenAPIExt:
```
whummer, line 77: Nice constants, much cleaner!

whummer (approved):
Looks great! :100: Again, kudos for the additional snapshots. :tada:

PR #8446, opened 2023-06-06
fix apigw model resolving when importing and when validating

bentsku:
Changes made to avoid circular references in `Model` containing references to itself, which can happen easily.

Example: A Person contains a `children` property, which is a collection of `Person`. After resolving the models, it will fail to be serialised to JSON because it contains circular references. Or a container which contains an object with a reference to its container. 

AWS solves this issue by storing the template with its absolute reference[[1]](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri)[[2]](https://stackoverflow.com/questions/37823112/how-do-i-reference-one-model-from-another-model-using-aws-api-gateway). It might be stored fully resolved server-side, or might be resolved when actually validating the template. ~Still WIP, so I'll see how I do this, this is only the importer phase now, next step is to properly read those schema when validating.~

This PR also implements `RequestValidator` when importing ([see here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-request-validators.html)), and support for `requestBody` property in `Method` for OpenAPI 3.0 ([see here](https://swagger.io/specification/v3/#request-body-object))

### Example:
```json
{
  "Person": {
    "type": "object",
    "description": "Random person schema.",
    "properties": {
      "randomProperty": {
        "type": "string",
        "description": "Random property",
        "format": "byte"
      },
      "container": {
        "$ref": "#/components/schemas/House"
      }
    }
  },
  "House": {
    "type": "object",
    "description": "Where Person can live",
    "properties": {
      "randomProperty": {
        "type": "string",
        "description": "Random property",
        "format": "byte"
      },
      "contains": {
        "type": "array",
        "items": {
          "$ref": "#/components/schemas/Person"
        },
        "minItems": 1,
        "description": "The information of who is living in the house"
      }
    }
  }
}
```
Once resolved before, it would have lead to `House` having a reference to `Person` and `Person` having a reference to `House`. These 2 models could be part of a bigger model, thus being in the same dictionary and make it non-serializable.

From this example, we want our Model to be able to be used to validate incoming requests. First, it will be resolved like AWS, with absolute paths. We will replace the relative `$ref` part with an absolute link to it, `$ref: "http://apigateway.{localstackhost}/restapis/{apiId}/models/{ModelName}`. This allows us to safely JSON dump them, avoiding circular references.

Now, the real issue comes when using those Model. We need to be able to resolve them.
The key is to use some tricks from JSON Schema, namely recursive[[3]](https://json-schema.org/understanding-json-schema/structuring.html#recursion) and `$defs` [[4]](https://json-schema.org/understanding-json-schema/structuring.html#defs). 
This will allow use to get our Model schema, and all `$ref` inside will be resolved and put into a `$defs` field.
We will recursively resolve all models and put them into `$defs`, even transitive ones. If we spot a `$ref` to our Model, to avoid infinite looping, we will replace its `$ref` with `#`, signifying to JSON schema that it references its own model. 

That's basically the gist of it. The PR validates that, changes a bit our schema Resolver, and adds a new `ModelResolver` before validating a request, and that resolved model would be cached in our store (also with the benefit of not `json.loads` every time). 

### What it gives:
Starting with those 2 schemas:
```json
"schemas": {
  "Person": {
    "type": "object",
    "description": "Random person schema.",
    "properties": {
      "name": {
        "type": "string",
        "description": "Random property"
      },
      "b": {
        "type": "number"
      },
      "house": {
        "$ref": "#/components/schemas/House"
      }
    },
    "required": ["name", "b"]
  },
  "House": {
    "type": "object",
    "description": "Where a Person can live",
    "properties": {
      "randomProperty": {
        "type": "string",
        "description": "Random property",
        "format": "byte"
      },
      "contains": {
        "type": "array",
        "items": {
          "$ref": "#/components/schemas/Person"
        },
        "minItems": 1,
        "description": "The information of who is living in the house"
      }
    }
  }
}
```
We then resolve them and store them as `Model` with absolute paths (from the snapshot):
```json
[
  {
    "contentType": "application/json",
    "description": "Where a Person can live",
    "id": "<model-id:2>",
    "name": "House",
    "schema": {
      "type": "object",
      "properties": {
        "randomProperty": {
          "type": "string",
          "description": "Random property",
          "format": "byte"
        },
        "contains": {
          "minItems": 1,
          "type": "array",
          "description": "The information of who is living in the house",
          "items": {
            "$ref": "<model-base-url>/restapis/<rest-id:1>/models/Person"
          }
        }
      },
      "description": "Where a Person can live"
    }
  },
  {
    "contentType": "application/json",
    "description": "Random person schema.",
    "id": "<model-id:3>",
    "name": "Person",
    "schema": {
      "required": [
        "b",
        "name"
      ],
      "type": "object",
      "properties": {
        "name": {
          "type": "string",
          "description": "Random property"
        },
        "b": {
          "type": "number"
        },
        "house": {
          "$ref": "<model-base-url>/restapis/<rest-id:1>/models/House"
        }
      },
      "description": "Random person schema."
    }
  }
]
```

And once we want to use them to validate our request, this is the `Person` schema, using `House` for one of its field:
It's a valid schema, as we can attest [here](https://www.jsonschemavalidator.net/s/xCgryAyF)
```json
{
  "type": "object",
  "description": "Random person schema.",
  "properties": {
    "name": {
      "type": "string",
      "description": "Random property"
    },
    "b": {
      "type": "number"
    },
    "house": {
      "$ref": "#/$defs/House"
    }
  },
  "required": [
    "name",
    "b"
  ],
  "$defs": {
    "House": {
      "type": "object",
      "description": "Where a Person can live",
      "properties": {
        "randomProperty": {
          "type": "string",
          "description": "Random property",
          "format": "byte"
        },
        "contains": {
          "type": "array",
          "items": {
            "$ref": "#"
          },
          "minItems": 1,
          "description": "The information of who is living in the house"
        }
      }
    }
  }
}
```

### Notes
- [[1] https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri)](https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri) 
- [[2] https://stackoverflow.com/questions/37823112/how-do-i-reference-one-model-from-another-model-using-aws-api-gateway](https://stackoverflow.com/questions/37823112/how-do-i-reference-one-model-from-another-model-using-aws-api-gateway)
- [[3] https://json-schema.org/understanding-json-schema/structuring.html#recursion](https://json-schema.org/understanding-json-schema/structuring.html#recursion)
- [[4] https://json-schema.org/understanding-json-schema/structuring.html#defs](https://json-schema.org/understanding-json-schema/structuring.html#defs)

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -32,7 +32,11 @@ from localstack.constants import (
     PATH_USER_REQUEST,
 )
 from localstack.services.apigateway.context import ApiInvocationContext
-from localstack.services.apigateway.models import ApiGatewayStore, apigateway_stores
+from localstack.services.apigateway.models import (
+    ApiGatewayStore,
+    RestApiContainer,
+    apigateway_stores,
+)
 from localstack.utils import common
 from localstack.utils.aws import aws_stack, queries
 from localstack.utils.aws import resources as resource_utils
@@ -147,8 +151,6 @@ class Resolver:
         if refpath in self._refpaths and not self.allow_recursive:
             raise Exception("recursion detected with allow_recursive=False")
 
-        print(f"{refpath=}")
-        print(f"{self.current_path=}")
         # We don't resolve the Model definition, we will return a absolute reference to the model like AWS
         # When validating the schema, we will need to resolve the $ref there
         # Because if we resolved all $ref in schema, it can lead to circular references in complex schemas
@@ -159,10 +161,10 @@ class Resolver:
 
         # We should not resolve the Model either, because we need its name to set it to the Request/ResponseModels,
         # it just makes our job more difficult to retrieve the Model name
-        if self.current_path.endswith("schema"):
-            return {"$ref": refpath}
+        # We still need to verify that the ref exists
+        is_schema = self.current_path.endswith("schema")
 
-        if refpath in self._cache:
+        if refpath in self._cache and not is_schema:
             return self._cache.get(refpath)
 
         with self._pathctx(refpath):
@@ -175,10 +177,14 @@ class Resolver:
                 cur = cur.get(step)
 
             self._cache[self.current_path] = cur
+
+            if is_schema:
+                # If the $ref doesn't exist in our schema, return None, otherwise return the ref
+                return {"$ref": refpath} if cur else None
+
             return cur
 
     def _namespaced_resolution(self, namespace: str, data: Union[dict, list]) -> Union[dict, list]:
-        # print(namespace)
         with self._pathctx(namespace):
             return self._resolve_references(data)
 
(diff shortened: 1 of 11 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -108,12 +119,14 @@ def get_apigateway_store(account_id: str = None, region: str = None) -> ApiGatew
 
 
 class Resolver:
```
calvernaz, line 121: At the time, `Resolver` looked like a ok name, not anymore, we need a better name for this openapi spec reference resolver :)

localstack/services/apigateway/helpers.py
```diff
@@ -108,12 +119,14 @@ def get_apigateway_store(account_id: str = None, region: str = None) -> ApiGatew
...
-    def __init__(self, document: dict, allow_recursive=True):
+    def __init__(self, document: dict, rest_api_id: str, allow_recursive=True):
         self.document = document
         self.allow_recursive = allow_recursive
         # cache which maps known refs to part of the document
         self._cache = {}
         self._refpaths = ["#"]
+        host_definition = localstack_host(use_localhost_cloud=True)
```
calvernaz, line 128: never seen this before, what is this?

localstack/services/apigateway/helpers.py
```diff
@@ -138,7 +151,20 @@ def _resolve_refpath(self, refpath: str) -> dict:
         if refpath in self._refpaths and not self.allow_recursive:
             raise Exception("recursion detected with allow_recursive=False")
 
-        if refpath in self._cache:
+        # We don't resolve the Model definition, we will return a absolute reference to the model like AWS
```
calvernaz, line 154: it can be a bug, sure, but the cache should avoid all circular loops. Resolving all refs up front, cleans up a lot of the lookups after.

calvernaz (commented):
few comments that I hope we can discuss them

localstack/services/apigateway/helpers.py
bentsku: Right, I could rename it to `OpenAPISpecificationResolver` or `OASResolver`!

localstack/services/apigateway/helpers.py
bentsku: I found the PR introducing it: https://github.com/localstack/localstack/pull/7774

localstack/services/apigateway/helpers.py
bentsku: Agreed. However, AWS does not resolve the Model at all as shown by the snapshot, they replace it with an absolute reference to the Model. This is also how you set up reference to other `Model` with the API outside of importing. I'd say `Model` is a specific case here.

Also, not resolving Model makes sense, because we don't use the schema of the `Model` but only set its name to the `requestModels` and `responseModels`.

bentsku: Also, the JSON Schema spec is talking about recursion and circular references, and they say that it's disallowed.
Also, I don't think the cache will avoid circular loop in case of circular references in the model itself once it's resolved.

calvernaz (approved):
we discussed offline, my comments are not blocking - approved 🚀

localstack/services/apigateway/helpers.py
bentsku: I'll also refactor at some point, like Waldemar said here: https://github.com/localstack/localstack/pull/8371#pullrequestreview-1462384708

It would be nice to be able to separate the import/export code.

We've also added a bunch a new supported features to the importer, we should try to add them to the exporter too.

localstack/services/apigateway/helpers.py
calvernaz: Yea, it's disallowed because if done in a certain way, you get a loop. The cache works as a mechanism to avoid chasing the tail if the reference is already resolved.

bentsku: I'm not sure I follow. I guess the goal of our resolver is to allow us to map this OAS to an API Gateway API, not necessarily properly resolve the schema? We don't need a valid OAS schema after resolving? Technically, I think it's still valid, it follows the retrieval URI spec: https://json-schema.org/understanding-json-schema/structuring.html#retrieval-uri

If AWS maps the Model this way in their `requestModels` and `requestResponses`, we can take this shortcut to allow us to map it in an easier way to what we need to create the API?
Sorry, I might be missing something here, genuine question.

calvernaz: > AWS does not resolve the Model at all as shown by the snapshot
>      "house": {
>          "$ref": "<model-base-url>/restapis/<rest-id:1>/models/House"
>       }

Having a reference or the inline method is valid. So it doesn't matter what AWS does, in fact, what does it mean? Did you export  and that's how it came out?

bentsku: No, I'm talking to how it maps a schema in `components/schema` to a Model, and how it manages those references inside the Model, see https://stackoverflow.com/questions/37823112/how-do-i-reference-one-model-from-another-model-using-aws-api-gateway

bentsku: After thinking about this, we should try to import an API with no Model defined in `definitions/schema`, but instead directly set the schema. I wonder what comes out of this, because AWS would lack the schema name. I think all of this discussion can actually be solved by this test?

bentsku: To quickly come back on this, if you don't specify a $ref in `responses/<statusCode>/schema` field, AWS will automatically create the Model for you. 
Basically done the following (simplified):
```json
{
  "paths": "/pet",
  "post": {
    "responses": {
      "200": {
        "description": "Successful operation",
        "schema": {
          "type": "object",
          "properties": {
            "pet": {
              "$ref": "#/definitions/Pet"
            },
            "message": {
              "type": "string"
            }
          }
        },
        "headers": {
          "Access-Control-Allow-Origin": {
            "type": "string",
            "description": "URI that may access the resource"
          }
        }
      }
    }
  }
}
```

And AWS created the follow Model (normally json encoded Schema field): 
```json
{
  "name": "MODEL03987f",
  "contentType": "application/json",
  "schema": {
    "type": "object",
    "properties": {
      "pet": {
        "$ref": "https://apigateway.amazonaws.com/restapis/<api-id>/models/Pet"
      },
      "message": {
        "type": "string"
      }
    }
  }
}
```

Seems like we're still missing that feature, I'll note it and add it later on!

PR #8450, opened 2023-06-07
Use new clients in lambda event source mappings

dfangl:
## Motivation
To provide IAM enforcement, we have to use the new clients, with the execution role assumed, for the event source mapping operations.

## Changes
* Use Adapters to get the right clients for the operation
* Migrate to new client factory
* Consolidate `wait_for_dynamodb_stream_ready` fixtures
* Handle failure for get_records operation in the streams event source listener separately, to avoid crashing the whole thread

### DIFF
--- a/localstack/services/awslambda/event_source_listeners/adapters.py
+++ b/localstack/services/awslambda/event_source_listeners/adapters.py
@@ -1,14 +1,18 @@
+import abc
 import json
 import logging
 import threading
 from abc import ABC
 from concurrent.futures import Future
+from functools import lru_cache
 from typing import Callable, Optional
 
 from localstack import config
 from localstack.aws.api.lambda_ import InvocationType
+from localstack.aws.connect import ServiceLevelClientFactory, connect_to
 from localstack.aws.protocol.serializer import gen_amzn_requestid
 from localstack.services.awslambda import api_utils
+from localstack.services.awslambda.api_utils import function_locators_from_arn, qualifier_is_version
 from localstack.services.awslambda.invocation.lambda_models import InvocationError, InvocationResult
 from localstack.services.awslambda.invocation.lambda_service import LambdaService
 from localstack.services.awslambda.invocation.models import lambda_stores
@@ -16,6 +20,7 @@ from localstack.services.awslambda.lambda_executors import (
     InvocationResult as LegacyInvocationResult,  # TODO: extract
 )
 from localstack.services.awslambda.lambda_utils import event_source_arn_matches
+from localstack.utils.aws.client_types import ServicePrincipal
 from localstack.utils.json import BytesEncoder
 from localstack.utils.strings import to_bytes, to_str
 
@@ -54,6 +59,10 @@ class EventSourceAdapter(ABC):
         pass
 
     def get_event_sources(self, source_arn: str):
+        pass
+
+    @abc.abstractmethod
+    def get_client_factory(self, function_arn: str, region_name: str) -> ServiceLevelClientFactory:
         pass
 
 
@@ -118,6 +127,9 @@ class EventSourceLegacyAdapter(EventSourceAdapter):
         from localstack.services.awslambda.lambda_api import get_event_sources
 
         return get_event_sources(source_arn=source_arn)
+
+    def get_client_factory(self, function_arn: str, region_name: str) -> ServiceLevelClientFactory:
+        return connect_to(region_name=region_name)
 
 
 class EventSourceAsfAdapter(EventSourceAdapter):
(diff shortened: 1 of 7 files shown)


localstack/testing/pytest/fixtures.py
```diff
@@ -656,7 +656,7 @@ def is_stream_ready():
                 "UPDATING",
             ]
```
joe4dev, line 658: idea: The triple nesting can be confusing to use. A usage example would be helpful.

localstack/services/awslambda/event_source_listeners/adapters.py
```diff
@@ -258,3 +270,29 @@ def get_event_sources(self, source_arn: str):
...
+
+    def _get_role_for_function(self, function_arn: str) -> str:
+        function_name, qualifier, account, region = function_locators_from_arn(function_arn)
+        store = lambda_stores[account][region]
+        function = store.functions.get(function_name)
+        if qualifier and qualifier != "$LATEST":
+            if qualifier_is_version(qualifier):
+                version_number = qualifier
+            else:
+                # the role of the routing config version and the regular configured version has to be identical
```
joe4dev, line 288: do we have snapshot tests for this?

localstack/services/awslambda/event_source_listeners/kinesis_event_source_listener.py
```diff
@@ -64,9 +66,7 @@ def _create_lambda_event_payload(self, stream_arn, records, shard_id=None):
                     "eventSource": "aws:kinesis",
                     "eventVersion": "1.0",
                     "eventName": "aws:kinesis:record",
-                    "invokeIdentityArn": "arn:aws:iam::{0}:role/lambda-role".format(
-                        get_aws_account_id()
-                    ),
+                    "invokeIdentityArn": f"arn:aws:iam::{get_aws_account_id()}:role/lambda-role",
```
joe4dev, line 69: ✨ 👍

localstack/services/awslambda/event_source_listeners/stream_event_source_listener.py
```diff
@@ -198,63 +200,75 @@ def _listen_to_shard_and_invoke_lambda(self, params: Dict):
...
+                                        failure_destination,
+                                    )
+                            else:
+                                should_get_next_batch = False
+                    if should_get_next_batch:
+                        shard_iterator = records_response["NextShardIterator"]
+                        num_invocation_failures = 0
                 time.sleep(self._POLL_INTERVAL_SEC)
         except Exception as e:
+            #
```
joe4dev, line 271: comment missing or typo ?

localstack/services/awslambda/event_source_listeners/stream_event_source_listener.py
```diff
@@ -198,63 +200,75 @@ def _listen_to_shard_and_invoke_lambda(self, params: Dict):
...
+                try:
+                    records_response = stream_client.get_records(
+                        ShardIterator=shard_iterator, Limit=batch_size
                     )
-                    if is_invocation_successful:
-                        should_get_next_batch = True
+                except ClientError as e:
+                    if "AccessDeniedException" in str(e):
+                        LOG.warning(
+                            "Insufficient permissions to get records from stream %s: %s",
```
joe4dev, line 210: Are these types of IAM exceptions logged somewhere else on AWS?

localstack/testing/pytest/fixtures.py
dfangl: You are right but that is pretty unrelated to the PR. Would definitely be nice to add more docs to the test fixtures though!

localstack/testing/pytest/fixtures.py
```diff
@@ -656,7 +656,7 @@ def is_stream_ready():
...
 
-        poll_condition(is_stream_ready)
+        return poll_condition(is_stream_ready)
```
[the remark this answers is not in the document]
dfangl: The return is needed if you do something like:
```
assert wait_for_stream_ready()
```
So you can make sure that it not only waited for it, but the wait also ended successfully (not just a timeout). I merged another fixture with this one, which is why this is needed.
We should do it everywhere, but the user of the fixture also has to check the return value for this, which would make this a bigger adventure.

localstack/services/awslambda/event_source_listeners/stream_event_source_listener.py
dfangl: Whoopsie, that's not supposed to be in there. Will remove!

localstack/services/awslambda/event_source_listeners/stream_event_source_listener.py
dfangl: Might be in cloudtrail, but as it is a data operation, probably not even there. So no, it's very hard in AWS to detect if something like this happens.

This is why in AWS you have a test for the permissions before you create the event source mappings, but that is kind of lambda specific. We do not have that yet, but it is definitely planned to be added.

localstack/services/awslambda/event_source_listeners/adapters.py
dfangl: I do not think so, no.

PR #8453, opened 2023-06-07
Verify AssumeRolePolicyDocument is valid json

dfangl:
## Motivation
If the AssumeRolePolicyDocument is not valid json, we have problems later with IAM enforcement, as it breaks the interface contract of the STS IAM plugin.

## Changes
* Verify the AssumeRolePolicyDocument actually being valid json by trying to decode it
* AWS validated test for this behavior

### DIFF
--- a/tests/integration/test_iam.py
+++ b/tests/integration/test_iam.py
@@ -123,6 +123,28 @@ class TestIAMExtensions:
         get_user_reply = aws_client.iam.get_user(UserName=user_name)
         assert "PermissionsBoundary" not in get_user_reply["User"]
 
+    @pytest.mark.aws_validated
+    def test_create_role_with_malformed_assume_role_policy_document(self, aws_client, snapshot):
+        role_name = f"role-{short_uid()}"
+        # The error in this document is the trailing comma after `"Effect": "Allow"`
+        assume_role_policy_document = """
+        {
+          "Version": "2012-10-17",
+          "Statement": [
+            {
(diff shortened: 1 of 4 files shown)


tests/integration/test_iam.py
```diff
@@ -123,6 +123,28 @@ def test_create_user_add_permission_boundary_afterwards(
...
+        role_name = f"role-{short_uid()}"
+        # The error in this document is the trailing comma after `"Effect": "Allow"`
+        assume_role_policy_document = """
+        {
+          "Version": "2012-10-17",
+          "Statement": [
+            {
+              "Action": "sts:AssumeRole",
+              "Principal": "*",
+              "Effect": "Allow",
```
dominikschubert, line 137: I assume something like `"Effect": "invalid"` would also be invalid. Is that currently checked anywhere? I could imagine that this would also cause issues downstream if the document is otherwise somehow invalid.

dominikschubert (approved):
LGTM :+1:

tests/integration/test_iam.py
dfangl: The structure of the statements is not checked yet, that is correct.
There is a lot more validation to do on IAM policies, as we currently make barely any assertion. This particular case with malformed json was mostly important as it leads to different types returned by boto3 when querying the policy, if it's not valid json, it will just be returned as string instead of dict.

Definitely a good point, we need more validation on the policies.

PR #8457, opened 2023-06-07
CFn: Add scaffolding system

simonrw:
Add scaffolding system for resource providers.

This will form the basis for how new resource providers are generated going forward. The scaffolding currently generates:

* a resource provider skeleton file
* a test for helping to understand which properties are valid `GetAtt` targets
* a template for the test generated.


simonrw:
@dominikschubert and I discussed that since this doesn't touch CFn core code that we should merge this and refactor later.

PR #8466, opened 2023-06-08
Add support for ruby3.2, remove tests for deprecated runtimes

dfangl:
## Motivation
AWS released the ruby3.2 runtime yesterday: https://aws.amazon.com/blogs/compute/ruby-3-2-runtime-now-available-in-aws-lambda/

Also, nodejs12.x and dotnetcore3.1 are deprecated since end of April/beginning of may: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy

Since we cannot snapshot deprecated runtimes anymore, they need to be removed.

## Changes
* Add support for ruby3.2
* Add ruby3.2 to runtime tests
* Regenerate Function API snapshots
* Update supported runtimes message
* Remove nodejs12.x test execution
* Remove dotnetcore3.1 test execution + test code
* Remove some obsolete listing of images + hashes, as they are outdated anyway


dfangl:
Valid point. My reason for "patch" was the general notion that parity fixes with AWS are in general patch releases. I agree however, that minor might be a better fit.

joe4dev (approved):
Great to see quick adoption of new Lambda runtimes 🥳  👏👏👏  LGTM 👍 

➕ Adding new runtimes has become quite easy with the multi-runtime tests 👍 
In the future, we could consider picking up specification-based validation for the InvalidParameterValueException.

💡 Let's see whether it affects any customers that want to test the migration path. This is an unlikely case but a 'legacy' mode could be helpful to confirm that a deprecated runtime works and a newer runtime fails.

PR #8470, opened 2023-06-09
add special char s3 test + fix moto lookup with cleaned up key name

bentsku:
This fix the second issue reported in #8174

Moto will by default clean up the key name (I'm not sure to what end? I think this should be removed because it can cause conflict between keys, one url encoded and the other not will have the same key in moto). 

So I've added the key cleaning in our moto access, which at least allows us to use keys with `%XX` encoded in them.

However, as moto cleans up name, what returns from `ListObject` and `ListObjectsV2` is still wrong and is url decoded. However, we cannot for sure know if the key was clean or not, so we can't fix it in our provider. This would need to be fixed upstream.


macnev2013 (approved):
LGTM

PR #8476, opened 2023-06-10
upgrade moto to 4.1.11.post1

thrau:
Upgrade moto to of 4.1.11.post1

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -84,7 +84,7 @@ runtime =
     jsonpath-ng>=1.5.3
     jsonpath-rw>=1.4.0,<2.0.0
     localstack-client>=2.0
-    moto-ext[all]==4.1.10.post2
+    moto-ext[all]==4.1.11.post1
     opensearch-py==2.1.1
     pproxy>=2.7.0
     pymongo>=4.2.0


viren-nadkarni (approved):
Looks like a single CW test failing. Also make sure to run -ext pipeline against this branch.

Otherwise, LGTM!

steffyP:
Added xfail-marker for `test_put_metric_alarm`, as it seems to be flaky now (first time on this branch, it run successfully). 

It is the only test, where we expect several changes of the ALARM state. I debugged the test locally, and the state seems to change correctly, but we also expect a SQS message, that is triggered by the alarm (Cloudwatch publishes the SNS topic when the alarm state changes). 
The messages cannot be received in time anymore. I am not sure if this something new, or caused by the moto upgrade. Couldn't find anything suspicious so far. 
However, the test is AWS validated, and with the set timeouts the SQS messages can be retrieved on AWS. 

Not sure if we should further dig into this before merging. Do you have any ideas what could cause this behavior?

viren-nadkarni:
-ext pipeline is :green_circle:. Merging PR.

PR #8488, opened 2023-06-12
CFn: add error reasons and tidy error message on stack deploy failure

simonrw:
When a cloudformation stack fails to deploy, it's hard to know why. In a [previous PR](https://github.com/localstack/localstack/pull/8336) we added printing the events to the traceback, but we don't capture the reason why the stack failed. Also the output format was hard to read.

Now the output is nicer, and the resource deploy events are nicer.

*PICTURE TBD*

Note: we don't assert parity right now, just that the failure reason is _present_. It's too much API surface to try to ensure parity on failure events.


dominikschubert (approved):
LGTM!

PR #8489, opened 2023-06-12
Add a fetch state call after CFn resource deployment

dominikschubert:
Previously when deploying a resource in `determine_physical_resource_id` a fetch_state was executed and the result was then available in the resource props for later deletion.

With our rework of the physical resource ID, this will now not necessarily be the case anymore since the `PhysicalResourceId` is available immediately after the handler/boto call and therefore `determine_physical_resource_id` won't be called and the state missing from the props.

This sometimes now leads to issues when trying to delete resources because the attributes they depend on might not have been saved to the resource state if they haven't been added manually in the result handler to the resource properties.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -1233,6 +1233,18 @@ class TemplateDeployer:
             if physical_id:
                 resource["PhysicalResourceId"] = physical_id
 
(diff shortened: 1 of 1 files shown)


simonrw (approved):
This will be nice to have before our removal of physical resource ids, since I have seen a few resources that require additional data alongside the physical resource id.

PR #8182, opened 2023-04-21
wip: fix S3 CopyObject copying or returning Checksum

bentsku:
This partially addresses #8172

It appeared S3 was not returning a checksum when requested, or would not copy the source object's checksum if not provided. This PR fixes that.

That is not enough for #8172 as moto's check when copying objects are wrong. There's a test in this PR that is xfail demonstrating the behaviour. This is not fixable by patching, so I'll open a PR upstream with the change I'm already working on, and will open new PR then here de-xfailing the test and adding more cases.

edit: this is a temporary fix, as moto has been working on adding and saving checksum in `FakeKey`, which will allow us to remove our own fixes in the provider. Until then (and until we update moto too, I'm going to open the PR for the `CopyObject` overhaul upstream), our fix here will provide this functionality.

PR #8498, opened 2023-06-14
use checksum stored in moto key, implement checksum for CopyObject

bentsku:
This PR supersedes #8182

This is needed for https://github.com/localstack/moto/pull/67, so that we can isolate the changes in this PR (adding the behaviour that has been removed in moto) and already release a new moto version without merging the big S3 stream handling PR. 

Also, moto does not handle the checksum for CRC32C as it depends on `awscrt`. We are recalculating it for this specific algorithm. 

It also has benefits on its own, as it allows us to properly cache the checksum value in the S3 Object instead of calculating it every time it's requested.

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -127,6 +127,7 @@ from localstack.services.s3.presigned_url import (
 from localstack.services.s3.utils import (
     _create_invalid_argument_exc,
     capitalize_header_name_from_snake_case,
+    extract_bucket_key_version_id_from_copy_source,
     get_bucket_from_moto,
     get_header_name,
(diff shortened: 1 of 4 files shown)


macnev2013 (commented):
LGTM, Can we add a test around this?

bentsku:
> LGTM, Can we add a test around this?

@macnev2013  There are a massive amount of tests already for CopyObject, that I've added for the round of PR in moto. They are still passing, so this should be good to go.

macnev2013 (approved):

thrau (approved):
LGTM! thanks for preparing this.

PR #8499, opened 2023-06-14
Extend list of supported Lambda layer runtimes, update validation snapshots

whummer:
Add Python 3.10 to list of supported Lambda layer runtimes.

We're currently getting this error when creating a Layer with Python 3.10 listed in compatible runtimes:
```
$ awslocal lambda publish-layer-version --layer-name layer1 --zip-file fileb:///tmp/testlayer.zip --compatible-runtimes python3.10
An error occurred (ValidationException) when calling the PublishLayerVersion operation: 1 validation error detected: Value '[python3.10]' at 'compatibleRuntimes' failed to satisfy constraint: Member must satisfy enum value set: [nodejs12.x, java17, provided, nodejs16.x, nodejs14.x, ruby2.7, python3.10, java11, dotnet6, go1.x, nodejs18.x, provided.al2, java8, java8.al2, dotnetcore3.1, python3.7, python3.8, python3.9]
```

This surfaced when dogfooding LS for integration testing with our platform. Let me know if you think this should be covered by a test, and what would be a good place to add one.


joe4dev:
Tests for Lambda Layers in the new providers are here: `tests.integration.awslambda.test_lambda_api.TestLambdaLayer`

We currently have multiple implementations for these validations 😬 :
* `localstack.services.awslambda.api_utils.validate_layer_runtimes_and_architectures` 👈  applies here for publish_layer_version
* `localstack.services.awslambda.api_utils.validate_layer_runtime`
* `localstack.services.awslambda.api_utils.validate_layer_architecture`

Such validation could be handled through specification-based validation (as commented by @thrau  in this PR https://github.com/localstack/localstack/pull/7675#discussion_r1107777058). For example, supported runtimes are [here](https://github.com/boto/botocore/blob/1.29.153/botocore/data/lambda/2015-03-31/service-2.json#L5211) in the specification.
👍 This facilitates updates because we get them with new boto updates.
⚠️ We need to update snapshots covering such validation messages together with boto updates to get the pipeline green
⚠️ If the boto release is not synchronized with launching new features, we get a mismatch in the snapshot tests.

whummer:
Thanks @dfangl @joe4dev , great findings and insights. Re-generated the snapshots, and also added support for `ruby3.2` and `java17` (verified that both these work against AWS). 👍

tests/integration/awslambda/test_lambda_api.snapshot.json
```diff
@@ -8021,7 +8021,7 @@
       "publish_layer_version_exc_invalid_runtime_arch": {
         "Error": {
           "Code": "ValidationException",
-          "Message": "2 validation errors detected: Value '[invalidruntime]' at 'compatibleRuntimes' failed to satisfy constraint: Member must satisfy enum value set: [nodejs12.x, java17, provided, nodejs16.x, nodejs14.x, ruby2.7, python3.10, java11, dotnet6, go1.x, nodejs18.x, provided.al2, java8, java8.al2, dotnetcore3.1, python3.7, python3.8, python3.9]; Value '[invalidarch]' at 'compatibleArchitectures' failed to satisfy constraint: Member must satisfy constraint: [Member must satisfy enum value set: [x86_64, arm64]]"
```
whummer, line 8024: Based on this output, we should actually be able to remove the `nodejs12.x`/`dotnetcore3.1` runtimes as well at some point..

dfangl: Yes, they are not available anymore for new function creation. We already removed them from the multi runtime tests, as they are not verifiable against AWS anymore, so they are not tested anymore, but left the support in for now, in case people want to test their legacy applications for some reason.

We currently do not have any structure for removing the support, as far as I know.

whummer:
Btw, haven't fully gotten to the bottom of it yet, but it seems that the `get_function` API may be indeterministic when it comes to returning the `RuntimeVersionConfig` attribute.

In the previous snapshot the `RuntimeVersionConfig` attribute was removed:
https://github.com/localstack/localstack/pull/8499/commits/4dda790aa9c83695473f32e37f655f1231209ef1#diff-c54f64e276578b0f71ee8be09b748883f67646b23aee6e9d201dc17b86e3b6ecL8074-L8076

After I re-generating the snapshot, it was returned again:
https://github.com/localstack/localstack/pull/8499/commits/8d20af2bfba4cce754a13cd118066276d36deb55#diff-c54f64e276578b0f71ee8be09b748883f67646b23aee6e9d201dc17b86e3b6ecR8074-R8076

dominikschubert:
@whummer 

It might also be caused by changing boto versions and/or region used when running the tests, so that could be one cause of the `RuntimeVersionConfig` sometimes being there and sometimes not :thinking:

localstack/services/awslambda/invocation/lambda_models.py
```diff
@@ -86,7 +86,7 @@ class Invocation:
 
 class ArchiveCode(metaclass=ABCMeta):
     @abstractmethod
-    def generate_presigned_url(self, endpoint_url):
+    def generate_presigned_url(self, endpoint_url: str | None = None):
```
whummer, line 89: Small change here, to match the signature of the implementation subclasses (detected this as it was highlighted by the IDE..)

whummer:
@dominikschubert That would make more sense as an explanation. I cannot fully rule out that the two commits referenced above are coming from different boto versions, unfortunately (ran the tests in different configurations, from the CLI, as well as the IDE). So it's quite likely that this is actually not an indeterminism, but rather coming from different botocore versions. Great catch! 👌

localstack/services/awslambda/api_utils.py
```diff
@@ -596,7 +599,7 @@ def parse_layer_arn(layer_version_arn: str) -> Tuple[str, str, str, str]:
...
+        return f"Value '{compatible_runtime}' at 'compatibleRuntime' failed to satisfy constraint: Member must satisfy enum value set: [ruby2.6, dotnetcore1.0, python3.7, nodejs8.10, nasa, ruby2.7, python2.7-greengrass, dotnetcore2.0, python3.8, java21, dotnet6, dotnetcore2.1, python3.9, java11, nodejs6.10, provided, dotnetcore3.1, java17, nodejs, nodejs4.3, java8.al2, go1.x, nodejs20.x, go1.9, byol, nodejs10.x, provided.al2023, python3.10, java8, nodejs12.x, python3.11, nodejs8.x, python3.12, nodejs14.x, nodejs8.9, nodejs16.x, provided.al2, nodejs4.3-edge, nodejs18.x, ruby3.2, python3.4, ruby2.5, python3.6, python2.7]"
```
joe4dev, line 602: interesting hints for upcoming runtimes 👁️  (the python3.11 Dockerfile is already on [GitHub](https://github.com/aws/aws-lambda-base-images/branches/active))

joe4dev (approved):
Thanks for fixing the validations for new runtimes 👍

PR #8507, opened 2023-06-15
KMS: Fix segfault during persistence state restore

viren-nadkarni:
A customer reported segmentation faults during persistence state restoration involving KMS keys and aliases. It turned out that we were persisting the private key object as part of the KMS key abstraction. This was problematic because this object, created by the underlying SSL implementation, maintained in-memory references. These references broke when the KMS state was deserialised and restored, causing a segfault.

This PR removes the assignment of the private key object to `KmsCryptoKey` abstraction, preventing it from going into the state. Instead it is created dynamically from the DER encoded private key string.

I've taken this opportunity to also refactor the KMS store definition. The helper methods have been moved out of the model definitions.

### DIFF
--- a/localstack/services/kms/models.py
+++ b/localstack/services/kms/models.py
@@ -13,9 +13,11 @@ from dataclasses import dataclass
 from typing import Dict, List, Optional, Tuple
 
 from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, hmac
 from cryptography.hazmat.primitives import serialization as crypto_serialization
 from cryptography.hazmat.primitives.asymmetric import ec, padding, rsa, utils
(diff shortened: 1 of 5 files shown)


localstack/services/kms/models.py
```diff
@@ -180,10 +170,10 @@ def __init__(self, key_spec: str):
 
         if key_spec.startswith("RSA"):
             key_size = RSA_CRYPTO_KEY_LENGTHS.get(key_spec)
-            self.key = rsa.generate_private_key(public_exponent=65537, key_size=key_size)
+            key = rsa.generate_private_key(public_exponent=65537, key_size=key_size)
         elif key_spec.startswith("ECC"):
             curve = ECC_CURVES.get(key_spec)
-            self.key = ec.generate_private_key(curve)
```
viren-nadkarni, line 186: This here is the fix. Everything else is just refactoring to dissociate helper methods from the store definition.

sannya-singal (approved):

PR #8509, opened 2023-06-15
ApiGateway integration client rework

dfangl:
**Please refer to the contribution guidelines in the README when submitting PRs.**


localstack/utils/sync.py
```diff
@@ -51,7 +51,10 @@ def wait_until(
         return wait_until(fn, next_wait, max_retries, strategy, _retries + 1, _max_wait)
 
 
-def retry(function, retries=3, sleep=1.0, sleep_before=0, **kwargs):
+T = TypeVar("T")
+
+
+def retry(function: Callable[..., T], retries=3, sleep=1.0, sleep_before=0, **kwargs) -> T:
```
dfangl, line 57: I got really annoyed my the missing typing here, so I kind of had to do it. I had a more sophisticated version with ParamSpec, but it turns out you cannot use the paramspec kwargs within the same method definition, only within the body of the method.

localstack/services/apigateway/invocations.py
```diff
@@ -215,14 +217,28 @@ def apply_response_parameters(invocation_context: ApiInvocationContext):
...
     return_code = str(response.status_code)
-    if return_code not in entries:
-        if len(entries) > 1:
-            LOG.info("Found multiple integration response status codes: %s", entries)
-            return response
-        return_code = entries[0]
-    response_params = int_responses[return_code].get("responseParameters", {})
+    # Selecting the right integration response
+    # FIXME is this check valid? It should map to the default one or based on the selectionPattern
+    # Not sure if it gets mapped automatically to a method response if it matches the statusCode by chance
```
bentsku, line 231: nit: I think this comment can now be removed with the introduction of `select_integration_response`. Thanks a lot for fixing this and adding the correct behaviour, this is really great, really, really grateful!

localstack/services/apigateway/invocations.py
```diff
@@ -215,14 +217,28 @@ def apply_response_parameters(invocation_context: ApiInvocationContext):
...
+    LOG.debug(
+        "Found multiple integration response status codes: %s, choosing by selection pattern",
+        entries,
+    )
```
bentsku, line 235: question: there could be only one integration response here, right, we're not checking the length of `entries` anymore? Do we still need this logging statement, or the logging in `select_integration_response` sufficient?

localstack/services/apigateway/invocations.py
```diff
@@ -215,14 +217,28 @@ def apply_response_parameters(invocation_context: ApiInvocationContext):
...
+    selected_integration_response = select_integration_response(return_code, invocation_context)
+    # set status code of integration response
+    response.status_code = selected_integration_response["statusCode"]
+    response_params = int_responses[selected_integration_response["statusCode"]].get(
+        "responseParameters", {}
+    )
```
bentsku, line 241:
```suggestion
    response_params = selected_integration_response.get("responseParameters") or {}
```
I think we already have the integration response here, we can directly access it. I've changed the check because it might be that the `responseParameters` key is present in the `dict`, but set to `None` or something. With the `or` check, we're sure that we're using a dictionary when calling `.items()` right under.

localstack/services/apigateway/invocations.py
```diff
@@ -344,7 +369,7 @@ def invoke_rest_api_integration_backend(invocation_context: ApiInvocationContext
         if "s3:path/" in uri or "s3:action/" in uri:
             return S3Integration().invoke(invocation_context)
 
-        if method == "POST" and ":sqs:path" in uri:
+        if integration_method == "POST" and ":sqs:path" in uri:
```
bentsku, line 372: good catch 👍

tests/integration/apigateway/test_apigateway_integrations.py
```diff
@@ -53,6 +56,112 @@ def test_http_integration(create_rest_apigw, aws_client, echo_http_server):
...
+    def _echo(request: Request) -> Response:
+        result = {
+            "data": request.data or "{}",
+            "headers": dict(request.headers),
+            "url": request.url,
+            "method": request.method,
+        }
+        status_code = request.url.rpartition("/")[2]
+        response_body = json.dumps(result)
+        print(result)
```
bentsku, line 74: leftover debugging statement?

tests/integration/apigateway/test_apigateway_integrations.py
```diff
@@ -53,6 +56,112 @@ def test_http_integration(create_rest_apigw, aws_client, echo_http_server):
...
+        return Response(response_body, status=int(status_code))
+
+    httpserver.expect_request("").respond_with_handler(_echo)
+    http_endpoint = httpserver.url_for("/")
+    return http_endpoint
+
+
+@pytest.mark.aws_validated
+def test_http_integration_status_code_selection(
```
bentsku, line 83: awesome test, thanks a lot 🙏

localstack/services/apigateway/invocations.py
```diff
@@ -295,6 +311,14 @@ def invoke_rest_api_integration(invocation_context: ApiInvocationContext):
...
         response = apply_response_parameters(invocation_context)
         return response
+    except ApiGatewayIntegrationError as e:
+        LOG.warning(
+            "Error while invoking integration for ApiGateway ID %s: %s",
+            invocation_context.api_id,
+            e,
+            exc_info=LOG.isEnabledFor(logging.DEBUG),
+        )
+        return e.to_response()
```
bentsku, line 321: this is really cool, really like the way we can properly return exceptions and a proper response from it. awesome idea!

bentsku (approved):
Awesome! LGTM! Thanks a lot for tackling this, this is really great and I'm glad to see we're getting better and better with API Gateway, the selection pattern is something we totally missed, and I'm really grateful you took the time to implement it.

Just some minor nit regarding I think leftover statements, and one access to the `responseParameters`, but it's really great! 😄 I don't have a lot of experience with integrations yet, but this seems to me like it can only improve our coverage and remove bugs.

localstack/services/apigateway/invocations.py
```diff
@@ -215,14 +217,28 @@ def apply_response_parameters(invocation_context: ApiInvocationContext):
...
+
     entries = list(int_responses.keys())
     return_code = str(response.status_code)
-    if return_code not in entries:
-        if len(entries) > 1:
-            LOG.info("Found multiple integration response status codes: %s", entries)
-            return response
-        return_code = entries[0]
-    response_params = int_responses[return_code].get("responseParameters", {})
+    # Selecting the right integration response
```
calvernaz, line 229: What you did is great and will be included in ongoing work to wrap this logic inside `ResponseTemplates`. The class already exists but not applied on every integration.

This is an example of code that should go away because the template handling would have been done inside ResponseTemplates

localstack/services/apigateway/invocations.py
```diff
@@ -215,14 +217,28 @@ def apply_response_parameters(invocation_context: ApiInvocationContext):
     int_responses = integration.get("integrationResponses") or {}
     if not int_responses:
         return response
+
+    integration_type_orig = integration.get("type") or integration.get("integrationType") or ""
```
calvernaz, line 221: Interesting, because `_PROXY` integrations can't have integration responses, it should have returned on [line](https://github.com/localstack/localstack/pull/8509/files#diff-ae146813864a8ec6b0ac068c7ea1098520a01dd30eeb6063458e99c385428148R218)

calvernaz (approved):
Thanks, @dfangl for adding selection pattern to the mix. We only supported direct status code matches for response template handling.

localstack/services/apigateway/invocations.py
calvernaz: It can be removed the new function picks up if there is more than one and is also logging.

tests/integration/apigateway/test_apigateway_integrations.py
```diff
@@ -53,6 +56,112 @@ def test_http_integration(create_rest_apigw, aws_client, echo_http_server):
...
+    create_rest_apigw, aws_client, status_code_http_server
+):
+    api_id, _, root_id = create_rest_apigw(name="my_api", description="this is my api")
```
calvernaz, line 86: Are we adding the policy tests on another PR? I know this took you from what you initially intended to do, so just checking

dfangl: Yes, it is in the linked -ext PR.

localstack/services/apigateway/invocations.py
dfangl: You are right. However, it might be possible to still do this on LS, it is just an additional guard here.

localstack/services/apigateway/invocations.py
dfangl: Yes, I just plain up forgot about that!

localstack/services/apigateway/invocations.py
dfangl: Also a leftover, thanks!

Issue #8505, opened 2023-06-15
bug: KMS Decrypt fails when passing EncryptionContext


davecardwell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Several KMS APIs take an optional encryption context argument. For example, when using `GenerateDataKey` if you pass an encryption context option you must pass the exact same context when later calling `Decrypt` on the key.

With LocalStack up to `v2.0.2` I don’t think the context was used correctly, but at least it would not fail when passing a context to decrypt. As of `v2.1.0` and `latest` the decrypt command throws an error.

The previous behavior of ignoring the context at least allowed legitimate requests to succeed.

### Expected Behavior

If `GenerateDataKey` was called with an encryption context, the `Decrypt` command should fail if you don’t pass the same context, and succeed if you do pass the same context. As of `v2.1.0` the opposite happens.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    # https://github.com/localstack/localstack/blob/master/docker-compose-pro.yml
    docker compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```sh
awslocal kms create-key
# {
#     "KeyMetadata": {
#         "AWSAccountId": "000000000000",
#         "KeyId": "fc3a54e7-8d7b-467e-871b-17bd3ea565bb",
#         "Arn": "arn:aws:kms:us-east-1:000000000000:key/fc3a54e7-8d7b-467e-871b-17bd3ea565bb",
#         "CreationDate": 1686789632.373033,
#         "Enabled": true,
#         "Description": "",
#         "KeyUsage": "ENCRYPT_DECRYPT",
#         "KeyState": "Enabled",
#         "Origin": "AWS_KMS",
#         "KeyManager": "CUSTOMER",
#         "CustomerMasterKeySpec": "SYMMETRIC_DEFAULT",
#         "KeySpec": "SYMMETRIC_DEFAULT",
#         "EncryptionAlgorithms": [
#             "SYMMETRIC_DEFAULT"
#         ],
#         "MultiRegion": false
#     }
# }

# using the KeyId from above:
awslocal kms generate-data-key --key-id fc3a54e7-8d7b-467e-871b-17bd3ea565bb --encryption-context foo=bar
# {
[...]

Issue #8503, opened 2023-06-14
Should pass EncryptionContext down the call stack for KMS data key generation


nathanmcgarvey-modopayments:
This effectively makes data key generation broken:

https://github.com/localstack/localstack/blob/7e002033125f2cd8e758cb57166793809df5e34e/localstack/services/kms/provider.py#L594

There are also several other related TODOs in that file.

I'm not quite sure why those aren't implemented. It looks like the encrypt function/method and the API inputs both take encryption contexts and the data key generation just decides to drop it and not pass it down?

sannya-singal:
Hey @nathanmcgarvey-modopayments, the issue should be now resolved, please pull the `latest` docker image and give it a try. Also, thanks for your feedback on #8517. The issue mentioned regarding the `GenerateDataKeyPair` handler was already in pipeline shall be rolled out today. Thank you!

nathanmcgarvey-modopayments:
> Hey @nathanmcgarvey-modopayments, the issue should be now resolved, please pull the `latest` docker image and give it a try. Also, thanks for your feedback on #8517. The issue mentioned regarding the `GenerateDataKeyPair` handler was already in pipeline shall be rolled out today. Thank you!

Just tried the `:latest` tag with this change in it. Looks great and works both via CLI and the aws-go-sdk v2.

PR #8517, opened 2023-06-16
kms: Add encryption context to generate data key with and without plaintext 

sannya-singal:
- Add `EncryptionContext` parameter to `GenerateDataKeyWithoutPlaintext` and `GenerateDataKey'. 
- Write snapshot tests to check the expected behaviour against AWS. 

Related issues: https://github.com/localstack/localstack/issues/8505 and https://github.com/localstack/localstack/issues/8503


nathanmcgarvey-modopayments:
FWIW (and I *just* missed the merge): This should be applied to the GenerateDataKeyPair functions/handlers as well:

https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKeyPair.html

That's because, although it is *generating* an asymmetric keypair, the actual operation that gives you back ciphertext encrypts the blob with a symmetric key, and, as such, takes an encryption context as well.

PR #8529, opened 2023-06-19
CFn: simplify "result_handler" function signature

simonrw:
Update signature of the "result_handler" functions to remove unused parameters, and parameters that the provider should not have access to, particularly the full list of resource definitions.

This work is in preparation of integrating the new resource providers into the template deployer.


dominikschubert (approved):
LGTM :wink: 

Just want to wait until the companion PR is up and green before merging this to avoid breaking -ext somehow

Issue #8380, opened 2023-05-26
Need ability to silence production creds warning, or support TEST_AWS_ACCOUNT_ID


joebowbeer:
Please either support the (single account) TEST_AWS_ACCOUNT_ID overrride that was removed from v2, or provide a way to silence the production creds warning that was added in v2.

In v1, my localstack init/ready.d scripts created resources using the default 0 id, my client services used genuine test accounts, and my localstack v1 logs were clean.

In v2, my localstack v2 logs are full of production creds warnings.

I'm running several services in an e2e test and using localstack to stub only a _subset_ of the AWS services. The services under test are using genuine AWS test account creds in a test cluster.

In this scenario, I don't want the services under test to need to know if they're communicating with a real AWS service (e.g., SSM Parameter Store) or a Localstack emulation (e.g., S3). They shouldn't need to switch creds depending on where their request is sent. That would require the code-under-test to be test-aware, which would result in a different code path being taken during testing.

This mixed-service e2e test worked fine up until now, but now in v2, my localstack logs are spammed with warnings about ignoring production creds.

I tried the new PARITY_AWS_ACCESS_KEY_ID config (v2.1), but then localstack uses the real id instead of the default id, and this results in a mismatch between the resources created with the default id in the init script, and those referenced using a real id during testing. This result is worse than having the logs filled with warnings.

_Originally posted by @joebowbeer in https://github.com/localstack/localstack/issues/8225#issuecomment-1536429238_

joebowbeer:
Fixed by #8530

PR #8530, opened 2023-06-19
add config variable to explicitly ignore AWS credentials

thrau:
quick fix for #8225 and #8380

i'd like to get some input from @dfangl and @viren-nadkarni whether this makes sense and is something we can expand on.

### DIFF
--- a/localstack/aws/accounts.py
+++ b/localstack/aws/accounts.py
@@ -117,7 +117,7 @@ def get_account_id_from_access_key_id(access_key_id: str) -> str:
         if not config.PARITY_AWS_ACCESS_KEY_ID:
             # If AWS_ACCESS_KEY_ID has production AWS credentials, ignore them
             if access_key_id.startswith("ASIA") or access_key_id.startswith("AKIA"):
-                LOG.warning(
+                LOG.debug(
                     "Ignoring production AWS credentials provided to LocalStack. Falling back to default account ID."
                 )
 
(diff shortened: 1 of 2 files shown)


localstack/aws/accounts.py
```diff
@@ -103,9 +103,10 @@ def get_account_id_from_access_key_id(access_key_id: str) -> str:
         if not config.PARITY_AWS_ACCESS_KEY_ID:
             # If AWS_ACCESS_KEY_ID has production AWS credentials, ignore them
             if access_key_id.startswith("ASIA") or access_key_id.startswith("AKIA"):
-                LOG.warning(
```
viren-nadkarni, line 106: I would prefer changing the log level here to `debug` than adding yet another config option which simply supresses a single log message.

thrau: Also ok for me

dfangl: Hard to say, I think. People are using DEBUG=1 very often, so it might still spam them. Then, on the other hand, I think warning is the correct log level, as it is something unexpected happening, not just some site note to a expected flow.

I am okay with either way. I agree with this approach not scaling in the future but on the other hand it does not really fit the "debug" logs.

In any case, I think it only solves #8380, and not the other issue (we need to investigate where these "production" credentials come from there).

viren-nadkarni (approved):
LGTM!

We can revisit this if this is still an issue for the users.

PR #8531, opened 2023-06-20
CFn: Integrate new resource provider deploy system

simonrw:
This PR adds an adaptor for the `GenericBaseModel` resource providers, which is invoked when a new resource provider is not found. It creates two "new style" resource providers: SSM parameter and OpenSearch Domain.

This behaviour is hidden behind a new feature flag `CFN_RESOURCE_PROVIDERS_V2` which must be enabled to opt-in to the new behaviour.

## Enabling new providers

Providers are enabled by setting `CFN_RESOURCE_PROVIDERS_V2=1` and ensuring that you have re-build the entrypoints list with `make entrypoints`. Then new providers are used in preference to old ones.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -476,7 +476,6 @@ if is_trace_logging_enabled():
         "Initializing the configuration took %s ms", int((load_end_time - load_start_time) * 1000)
     )
 
-
 # expose services on a specific host externally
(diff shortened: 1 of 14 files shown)


dominikschubert (approved):
LGTM for its intended purpose now. Awesome that it's already working with the existing tests :+1:

PR #8541, opened 2023-06-21
Opensearch: Multi-accounts compatibility

viren-nadkarni:
This PR adds multi-accounts awareness to OpenSearch and ElasticSearch. 

Now, account IDs and regions are only taken either from request context or ARNs in scope. This means these provider operations won't break when using a non-default account ID.


alexrashed (approved):
LGTM! 🚀 
Nice to see that it's just a small change now to make a service multi account compatible!
Maybe we could migrate an existing OpenSearch test or create a new test to properly test the multi-accounts setup for ES/OS in the next iteration...

PR #8544, opened 2023-06-21
Fix container port check and host detection on podman

dfangl:
## Motivation
* Podman has issues with empty entrypoints. While they seem to be fixed in some cases, using the docker sdk seems to still trigger this behavior.
* Podman cannot resolve `host.docker.internal`, which is necessary for services like EKS. The default fallback to the bridge network gateway is invalid as well. It can however use `host.containers.internal`.

## Changes
* Start port detection container with `sh -c` instead of an empty entrypoint to circumvent the issue. We might seek a more fundamental fix at some point, but this will do for now.
* Add a fallback to `host.containers.internal` if `host.docker.internal` does not resolve. If this as well fails, the fallback behaves like it currently does.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -720,7 +720,10 @@ try:
             DOCKER_HOST_FROM_CONTAINER = "host.docker.internal"
     # update LOCALSTACK_HOSTNAME if host.docker.internal is available
     if is_in_docker:
-        DOCKER_HOST_FROM_CONTAINER = socket.gethostbyname("host.docker.internal")
+        try:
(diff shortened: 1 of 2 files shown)


dominikschubert (approved):
LGTM!

alexrashed (approved):
Awesome, thanks for jumping on this! The fallback seems good, maybe at some point Docker will also resolve `host.containers.internal`... 🚀

dfangl:
Podman tests seem to fail due to https://bugs.launchpad.net/ubuntu/+source/libpod/+bug/2024394 .

PR #8546, opened 2023-06-22
Refactor CFn conditions & mappings

dominikschubert:
Changes

- Move default resource property generation to LegacyResourceProvider
- Remove mappings from `Stack.resources` for a clearer data flow / future restructuring
- Explore/fix handling of intrinsic conditions functions & `AWS::NoValue` handling and add corresponding tests


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -450,33 +448,44 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str |
             item_to_sub[1].update(attr_refs)
 
             for key, val in item_to_sub[1].items():
-                val = resolve_refs_recursively(stack_name, resources, val)
+                val = resolve_refs_recursively(stack_name, resources, mappings, val)
```
whummer, line 451: General observation / nit: With the various refactorings going on right now, it seems a bit brittle to rely mostly on positional args. Technically it should be highlighted by the IDE if incorrect param types are passed to a function, but wondering if we could sprinkle in a few keyword args, to reduce the surface area of incorrect mappings. 🤷

whummer (approved):
LGTM, awesome to see the ongoing refactorings and enhancements in the template deployer. Love the depth of test coverage with the additional snapshot tests for different edge/error cases! 🚀

localstack/services/cloudformation/engine/template_deployer.py
dominikschubert: Yes! That's something I want to start doing (forcing kwargs usage instead of position arguments).

> Technically it should be highlighted by the IDE if incorrect param types are passed to a function
In practice that hasn't worked out well unfortunately, which is why I'm also of the opinion that kwargs are usually to be preferred where possible.

It's extremely hard to refactor with positional args only especially since most of the GenericBaseModel implementations were using the signatures very inconsistently

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -397,19 +389,21 @@ def resolve_refs_recursively(stack_name, resources, value):
...
+            ref = resolve_ref(stack_name, resources, mappings, value["Ref"], attribute="Ref")
             if ref is None:
                 msg = 'Unable to resolve Ref for resource "%s" (yet)' % value["Ref"]
                 LOG.debug("%s - %s", msg, resources.get(value["Ref"]) or set(resources.keys()))
                 raise DependencyNotYetSatisfied(resource_ids=value["Ref"], message=msg)
-            ref = resolve_refs_recursively(stack_name, resources, ref)
+            ref = resolve_refs_recursively(stack_name, resources, mappings, ref)
             return ref
 
         if stripped_fn_lower == "getatt":
```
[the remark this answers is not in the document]
dominikschubert: Yup, that will be a whole intrinsic function rework. Unfortunately we're not quite there yet were we can replace the intrinsic function resolving. The idea we had here was to build a sort of "hydrated" stack that contains the whole dependency graph and a somwhat "rich" API to navigate it & resolve refs, etc.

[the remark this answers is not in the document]
dominikschubert: The current approach is also a bit of a problem because we allow things that we shouldn't. A lot of sections / Intrinsic functions only allow a subset of other intrinsic functions to be contained in their fields, which we currently   can't easily model with this construct.

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -450,33 +448,44 @@ def _resolve_refs_recursively(stack_name, resources, value: dict | list | str |
...
-            if isinstance(selected_map, dict) and "Ref" in selected_map:
-                selected_map = resolve_ref(stack_name, resources, selected_map["Ref"], "Ref")
-            result = resolve_ref(stack_name, resources, selected_map, attribute=attr)
-            if not result:
+            # "Fn::FindInMap"
+            mapping_id = value[keys_list[0]][0]
+
+            if isinstance(mapping_id, dict) and "Ref" in mapping_id:
+                # TODO: ??
+                mapping_id = resolve_ref(stack_name, resources, mappings, mapping_id["Ref"], "Ref")
```
dominikschubert, line 464: a bit of context for the ?? here: I'm just not a fan of using resolve_ref here (because not all refs here are valid) in the long term but for now its fine.

PR #8548, opened 2023-06-22
Add scaffolded IAM resource provider with new scaffold test layout

dominikschubert:
Still a bit WIP but another iteration on which tests we want to auto-generate when scaffolding a new resource provider.

The PR also includes a few minor fixes to the scaffolding script itself.


simonrw (approved):
LGTM!

Issue #7426, opened 2023-01-04
bug: DynamoDB DescribeTable treats the queried table as a replica


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The DynamoDB DescribeTable operation treats the current queried table as a replica.

```
$ awslocal dynamodb describe-table --table-name global01 --query 'Table.Replicas' --region us-west-1
[
    {
        "RegionName": "ap-south-1",
        "ReplicaStatus": "ACTIVE"
    },
    {
        "RegionName": "eu-central-1",
        "ReplicaStatus": "ACTIVE"
    },
    {
        "RegionName": "us-west-1",  # <<< THIS MUST NOT BE RETURNED
        "ReplicaStatus": "ACTIVE"   # <<<
    }
]
```

### Expected Behavior

The correct response should be like so:

```
$ awslocal dynamodb describe-table --table-name global01 --query 'Table.Replicas' --region us-west-1
[
    {
        "RegionName": "ap-south-1",
        "ReplicaStatus": "ACTIVE"
    },
    {
        "RegionName": "eu-central-1",
        "ReplicaStatus": "ACTIVE"
    }
]
```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack: latest
```

PR #8549, opened 2023-06-22
DynamoDB: Fix replica responses

viren-nadkarni:
Closes: #7426

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -338,11 +338,14 @@ def modify_context_region(context: RequestContext, region: str):
         flags=re.IGNORECASE,
     )
 
-    yield context
-
-    # revert the original context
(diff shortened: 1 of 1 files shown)


localstack/services/dynamodb/provider.py
```diff
@@ -338,11 +338,14 @@ def modify_context_region(context: RequestContext, region: str):
...
-    context.region = original_region
-    context.request.headers["Authorization"] = original_authorization
+    try:
+        yield context
+    except Exception:
+        raise
+    finally:
+        # revert the original context
+        context.region = original_region
+        context.request.headers["Authorization"] = original_authorization
```
viren-nadkarni, line 348: This restores the original context even if there is an exception in the context block

giograno (approved):
Thanks for tackling and the docs update 👍

PR #8554, opened 2023-06-22
Let SQS AddPermission call actually influence policy attribute

dfangl:
## Motivation
Currently, SQS AddPermission and RemovePermission calls do only save the permissions in some internal datastructure - they are never returned with sqs.GetQueueAttributes. This is problematic, since this is a necessity for cross account evaluation.

You could just use SetQueueAttributes on the `Policy` attribute but we should match AWS behavior.

## Changes
* Rework SQS permission management, by adapting the stored policy
* Add AWS validated test for permission lifecycles.

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -7,7 +7,7 @@ import re
 import threading
 import time
 from queue import Empty, PriorityQueue, Queue
-from typing import Dict, NamedTuple, Optional, Set
+from typing import Dict, Optional, Set
 
 from localstack import config
 from localstack.aws.api import RequestContext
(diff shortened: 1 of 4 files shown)


dfangl:
Tests against pro will be fixed with #8557

localstack/services/sqs/models.py
```diff
@@ -163,13 +163,6 @@ def __repr__(self):
         return f"SqsMessage(id={self.message_id},group={self.message_group_id})"
 
 
-class Permission(NamedTuple):
-    # TODO: just a placeholder for real policies
```
thrau, line 167: nice to finally get that todo done :100:

localstack/services/sqs/models.py
```diff
@@ -496,6 +488,74 @@ def validate_queue_attributes(self, attributes):
             if k not in valid:
                 raise InvalidAttributeName(f"Unknown Attribute {k}.")
 
+    def add_permission(self, actions: list[str], account_ids: list[str], label: str) -> None:
```
thrau, line 491: super minor nit: i think having label as the first argument would be a more consistent API

thrau (approved):
excellent! :rocket:

PR #8564, opened 2023-06-23
unify CLI error messages, add global error message handler

alexrashed:
This PR aims at unifying the CLI error- and exitcode-handling.
- It explicitly catches some common exceptions and prints proper error messages (f.e. Docker is not installed, localstack_main is not running for status command,...). /cc @SimonWallner 
- It uses the `ClickException` for all error cases to use a unified error message as well as a non-zero exit code: https://click.palletsprojects.com/en/8.1.x/exceptions/#exception-handling

Changes postponed to later iterations:
- Introduce more fine grained Docker exceptions, clearly indicating that a container cannot be started due to port- or naming conflicts. (This was initially the plan for this PR, but this needs some changes in the Docker clients, which would bloat up this PR quite a lot.) /cc @dfangl

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import sys
+import traceback
 from typing import Dict, List, Optional
 
 from localstack import config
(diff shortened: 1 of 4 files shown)


thrau (approved):
this looks very nice! :100:

PR #8568, opened 2023-06-25
invoke test method and status code

calvernaz:
# Changes

- Adds log, latency and multiValueHeaders to the response for the "Test Invoke Request"
- Add snapshot test.

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -4,6 +4,7 @@ import json
 import logging
 import re
 import time
+from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict, Union
 from urllib import parse as urlparse
@@ -43,7 +44,7 @@ from localstack.utils.aws import resources as resource_utils
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.aws.aws_responses import requests_error_response_json, requests_response
 from localstack.utils.aws.request_context import MARKER_APIGW_REQUEST_REGION, THREAD_LOCAL
-from localstack.utils.strings import long_uid, short_uid
+from localstack.utils.strings import long_uid, short_uid, to_str
 from localstack.utils.time import TIMESTAMP_FORMAT_TZ, timestamp
 from localstack.utils.urls import localstack_host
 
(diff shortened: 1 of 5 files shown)


tests/integration/apigateway/test_apigateway_api.py
```diff
@@ -69,6 +77,145 @@ def _factory(*args, **kwargs):
...
+                pathWithQueryString=path_with_query_string,
+                body=body,
+            )
+            assert 200 == res.get("status")
+            return res
+
+        response = retry(
+            invoke_method, retries=10, sleep=5, path_with_query_string="/pets/123", method="GET"
+        )
+        snapshot.match("test-invoke-method-get", response)
```
whummer, line 198: Do we maybe want to add 1-2 assertions for the content of `log` in the response? (as we're excluding the log details from the snapshot, which makes sense)

whummer (commented):
LGTM overall, great to see `test_invoke_method` covered by a snapshot test. 🎉  

Added a few minor comments/suggestions.. Would be great if @bentsku could take a quick look as well. 👀

tests/integration/apigateway/test_apigateway_api.py
```diff
@@ -69,6 +77,145 @@ def _factory(*args, **kwargs):
...
+        # Create the response for method POST /pets/{petId}
+        create_rest_api_method_response(
+            aws_client.apigateway,
+            restApiId=api_id,
+            resourceId=resource_id,
+            httpMethod="POST",
+            statusCode="200",
+        )
+        # Create the integration to connect GET /pets/{petId} to a backend
+        integration_uri, _ = create_rest_api_integration(
```
bentsku, line 140: nit: do we need anything this method returns?

bentsku (approved):
LGTM, thanks Waldemar for the review! Overall, the same comments as Waldemar, as some very minor nits regarding using keyword arguments for methods with many arguments for readability, and a question regarding the usage of the fixtures but not really fixtures 😄

Thanks a lot for adding the snapshot test and improve the number of fields we now return!

localstack/services/apigateway/helpers.py
```diff
@@ -71,7 +72,18 @@
 PATH_REGEX_CLIENT_CERTS = r"/clientcertificates/?([^/]+)?$"
 PATH_REGEX_VPC_LINKS = r"/vpclinks/([^/]+)?(.*)"
 PATH_REGEX_TEST_INVOKE_API = r"^\/restapis\/([A-Za-z0-9_\-]+)\/resources\/([A-Za-z0-9_\-]+)\/methods\/([A-Za-z0-9_\-]+)/?(\?.*)?"
-
+INVOKE_TEST_LOG_TEMPLATE = """Execution log for request {}
```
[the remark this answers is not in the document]
calvernaz: you probably mean on the log function, but yeah agreed!

localstack/services/apigateway/helpers.py
```diff
@@ -1479,3 +1482,56 @@ def is_greedy_path(path_part: str) -> bool:
...
+    for key in elements:
+        if isinstance(key, (list, tuple)):
+            key, value = key
+        else:
+            value = elements[key]
+
+        key = to_str(key)
+        temp_mv_dict[key].append(value)
+
+    return dict((k, tuple(v)) for k, v in temp_mv_dict.items())
```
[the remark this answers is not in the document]
calvernaz: Yep, i just moved this code around but use of comprehension here makes sense

tests/integration/apigateway/test_apigateway_api.py
calvernaz: nice catch, removed

localstack/services/apigateway/provider.py
```diff
@@ -110,28 +114,48 @@ def on_after_init(self):
...
     ) -> TestInvokeMethodResponse:
-
         invocation_context = to_invocation_context(context.request)
         invocation_context.method = request["httpMethod"]
+        invocation_context.api_id = request["restApiId"]
+        invocation_context.path_with_query_string = request["pathWithQueryString"]
+
+        response = connect_to().apigateway.get_resource(
+            restApiId=invocation_context.api_id, resourceId=request["resourceId"]
+        )
```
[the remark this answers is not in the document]
calvernaz: It should return a 404 in case the resource doesn't exist, and given the api id is a required parameter it should fail on the client side (using aws clients). However for the rest api case we can make sure it's present and it's in our store.

Good point accessing moto or our store for the resource directly would be a better way to do it.

tests/integration/apigateway/test_apigateway_api.py
calvernaz: Yep will do!

Issue #8567, opened 2023-06-25
bug: Stage variables not evaluated for HTTP integration URI


agershman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

AWS API Gateway allowed for use of stage variables in an HTTP integration URI (ref: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-stages.html). E.g. `http://${stageVariables.foo}`. I've observed that localstack will not evaluate stage variable interpolation in this context, which in turn causes the upstream HTTP request to fail due to what is a nonsensical host.

```sh
requests.exceptions.ConnectionError: MyHTTPConnectionPool(host='$%7bstagevariables.foo%7d', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x4046249120>: Failed to establish a new connection: [Errno -2] Name or service not known'))
```

### Expected Behavior

I'd like to see localstack's emulation of AWS API Gateway follow the documented contract for stage variables in an HTTP integration URI. See the API Gateway stage variables reference / HTTP integration URIs section of https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-stages.html for documented behavior.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start

### Place the following JSON in a file named `api.json`

```json
{
   "openapi":"3.0.1",
   "info":{
      "title":"example",
      "version":"1.0"
   },
   "paths":{
      "/path1":{
         "get":{
            "x-amazon-apigateway-integration":{
               "httpMethod":"GET",
               "payloadFormatVersion":"1.0",
               "type":"HTTP_PROXY",
               "uri":"http://${stageVariables.foo}"
            }
         }
      }
   }
}
```

### Create a REST API which has an HTTP integration URI referencing a stage variable
```sh
awslocal apigateway import-rest-api --body 'fileb:///path/to/api.json'
```
 
### Create a deployment for our REST API
```sh
awslocal apigateway create-deployment --rest-api-id <rest-api-id>
```

### Create a dummy stage and deployment for our REST API. Note the stage variables
```sh
awslocal apigateway create-stage --rest-api-id <rest-api-id> --stage-name v1 --variables foo=docker.host.internal:5000 --deployment-id <deployment-id>
```

### Start up a dummy HTTP server to back our integration
```sh
[...]

PR #8569, opened 2023-06-25
Adds HTTP integration URIs

calvernaz:
# Changes

Adds support for HTTP integration URIs  https://docs.aws.amazon.com/apigateway/latest/developerguide/aws-api-gateway-stage-variables-reference.html

### DIFF
--- a/tests/integration/apigateway/test_apigateway_import.py
+++ b/tests/integration/apigateway/test_apigateway_import.py
@@ -43,7 +43,7 @@ OAS_30_CIRCULAR_REF = os.path.join(PARENT_DIR, "files", "openapi.spec.circular-r
 OAS_30_CIRCULAR_REF_WITH_REQUEST_BODY = os.path.join(
     PARENT_DIR, "files", "openapi.spec.circular-ref-with-request-body.json"
 )
-
+OAS_30_STAGE_VARIABLES = os.path.join(PARENT_DIR, "files", "openapi.spec.stage-variables.json")
 TEST_LAMBDA_PYTHON_ECHO = os.path.join(PARENT_DIR, "awslambda/functions/lambda_echo.py")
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/apigateway/test_apigateway_import.py
```diff
@@ -722,3 +722,23 @@ def test_import_with_circular_models_and_request_validation(
         request = requests.post(url, json=wrong_request_schema)
         assert request.status_code == 400
         assert request.json().get("message") == "Invalid request body"
+
+    def test_import_with_stage_variables(self, import_apigw, aws_client, echo_http_server):
```
bentsku, line 726: nit: maybe we could just run the test against AWS using the `echo_http_server_post` fixture which uses `httpbin.org` when running against AWS and the default pytest fixture against LocalStack? Just so that the test can marked as `aws_validated`? We would just need to change the method to `POST`.

bentsku (commented):
LGTM! This is awesome, thanks for adding this feature so quickly, sorry for the delay in reviewing.
It would be nice if we could validate it against AWS, even if it seems obvious this will work against it, just so that we have the marker and we have confidence it runs (also for maybe future automated tests against AWS). I've left a comment regarding a fixture that can be used to run against AWS as well.

Thanks again! Also, awesome use of the import feature that makes it so much simpler to set up simple resources instead of so many direct SDK calls. 🚀

tests/integration/apigateway/test_apigateway_import.py
calvernaz: Yep makes sense, let me fix that

whummer (approved):
LGTM 🚀 - kudos @calvernaz for adding the AWS validated test.

bentsku (approved):
Awesome, thanks for making it AWS validated, sorry for scheme issue. 🚀

ZaheerUdDeen:
HI team,

I am facing this issue, any support here will highly be appreciated.

https://github.com/localstack/localstack/issues/8567#issuecomment-2215188281

PR #8573, opened 2023-06-26
Support to import documentation parts

calvernaz:
REST API supports adding documentation to an API, using the "import documentation parts" API.

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-documenting-api-quick-start-import-export.html

This PR adds support to "import documentation parts" for REST API but only the "override" mode.

# Changes

- implements `import_documentation_parts` for "overwrite" mode.
- adds snapshot testing.


bentsku (approved):
LGTM! This is great, thanks for adding it! TIL about the `import_documentation_parts` API method, I had no idea it existed. Thanks a lot for the snapshot test, this is great!

PR #8574, opened 2023-06-26
Removing duplicate pickling logic for ACM

giograno:
We have already the same custom pickling approach implemented [here](https://github.com/localstack/localstack-plugin-persistence/blob/main/localstack_persistence/pickling/reducers/services/acm.py). This is redundant.


alexrashed (approved):
LGTM! 🚀

PR #8087, opened 2023-04-06
Update to Python 3.11

dfangl:
## Motivation
This PR contains all changes necessary to upgrade our Python version for the runtime from 3.10 to 3.11.

## Changes
- Update all CircleCI runners (which use Python 3.11 by default).
- Update all GitHub actions to install Python 3.11 instead of Python 3.10.
- Install `libsnappy-dev` instead of `libsasl2-dev` when running against Pro (see changes there for a detailed explanation).
- Update the dockerfile base image to the latest sha digest of Python 3.11.5.
- Update all references to virtual environments which contain the Python version (f.e. `/usr/lib/localstack/python-packages/lib/python3.11/site-packages`).
- Remove the SSL proxy (it's not used in Community anymore and the used library - `python-proxy` - is not maintained and not compatible with 3.11).
- Update dill to a Python 3.11 compatible version.
- Update the Lambda Runtime version used for local execution (in the runtime of LocalStack itself) from 3.10 to 3.11.

## Impact
This PR only updates the runtime version of LocalStack itself. The LocalStack CLI will still be compatible with Python 3.7+.
However, [Python Initialization Hooks](https://docs.localstack.cloud/references/init-hooks/) and [LocalStack Extensions](https://docs.localstack.cloud/references/localstack-extensions/) will then run in a Python 3.11 environment instead of Python 3.10.

## TODO:
- [x] Analyze Podman tests (unrelated)
- [x] Ensure compatibility with LocalStack Pro before merging this PR.

PR #8577, opened 2023-06-26
remove pproxy

alexrashed:
This PR removes the `proxy_server.start_ssl_proxy` function, as well as the dependency `pproxy`.
Once `pproxy` is removed, we should be able to move towards Python 3.11 for the runtime: https://github.com/localstack/localstack/pull/8087


alexrashed:
This PR is closed in favor of #8087. This PR will combine all efforts necessary to migrate to Python 3.11 (if possible).

PR #8578, opened 2023-06-26
add docker client commands to get docker system info

thrau:
Adds two methods `get_system_info` and `get_system_id` to the container client interface. On docker, this is populated using `docker info`, and can provide a stable machine ID even from within the container.

Future work: check whether this works on podman (ad: daniel checked, it works), and how that would work on k8s.

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -4,7 +4,6 @@ import logging
 import os
 import re
 import time
-import uuid
 from typing import NamedTuple
 
 import pytest
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import time
+import uuid
 from typing import NamedTuple
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -118,6 +118,16 @@ def _create_network(network_name: str):
...
 class TestDockerClient:
+    def test_get_system_info(self, docker_client: ContainerClient):
+        info = docker_client.get_system_info()
+        assert "ID" in info
+        assert "OperatingSystem" in info
+        assert "Architecture" in info
+
+    def test_get_system_id(self, docker_client: ContainerClient):
+        assert len(docker_client.get_system_id()) > 1
+        assert docker_client.get_system_id() == docker_client.get_system_id()
```
dfangl, line 129: This part will fail on podman, as the id is regenerated every time. Might be problematic for our podman tests, once ubuntu fixes it's podman networking packages :)

dfangl (approved):
LGTM for now! We can tackle the podman tests once they work again.

tests/integration/docker_utils/test_docker.py
thrau: that's very good to know!

PR #8581, opened 2023-06-27
cfn: add fix for multiple subnet ids in vpc config in lambda functions

sannya-singal:
Add comma separated values for providing multiple subnet ids in vpc config for lambda function cloudformation template deployer.


dominikschubert (changes requested):
LGTM in general, but it should be covered via at least one test in the lambda cloudformation resource tests (`tests.integration.cloudformation.resources.test_lambda`)

You could use the template here: https://github.com/localstack/localstack/blob/59cab7842922c014843b5fa3f8d194511da8743b/tests/integration/templates/cfn_lambda_vpc.yaml#L295-L302 as a starting point

PR #8582, opened 2023-06-27
Fix termination of generators when client connection is closed

dfangl:
## Current Issue
We are currently not closing down our response generators after the client closes the connection. This leads to resource leaks, as the generators keep running in the background, even though no one will ever read the response.

This snipped from hypercorn `http_stream.py` shows us, that all response `send` calls will return immediately once the connection is closed.

```python
    async def app_send(self, message: Optional[ASGISendEvent]) -> None:
        if self.closed:
            # Allow app to finish after close
            return
```

We now need to also stop the generators, once we detect the client connection is closed.

## Changes
* Access internals of `send` call to determine if the connection to the client is closed already and raise a `BrokenPipeError` in that case.
* If any `ConnectionError` is raised during sending data from the iterator, we stop reading from it and close the (async) generator
* Add test for this behavior in the asgi bridge, to avoid future regression
* We do not read all responses in our request logging, to enable chunked responses for requests not parsed by the serializer

### DIFF
--- a/localstack/aws/handlers/logging.py
+++ b/localstack/aws/handlers/logging.py
@@ -82,7 +82,9 @@ class ResponseLogger:
     def _log(self, context: RequestContext, response: Response):
         aws_logger = self.aws_logger
         http_logger = self.http_logger
-        is_internal_call = is_internal_call_context(context.request.headers)
+        is_internal_call = (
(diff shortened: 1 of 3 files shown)


thrau (approved):
great find! this was a fun session for me also :grin:

PR #8583, opened 2023-06-28
Add support for list-type stack parameters 

dominikschubert:
Adds very basic support for CloudFormation stack parameters that aren't just a single value but actually a comma-delimited list, e.g. `List<AWS::EC2::Subnet::Id>`

The tests needs to be replaced at some point, just churned this out as a quick fix for now. The behavior has been verified against AWS and can be verified using the test by setting the subnet and security group IDs manually.

/cc @sannya-singal


pinzon (approved):
LGTM :rocket:

PR #8585, opened 2023-06-28
Add small fix and test to assert prepare_host hook is called with correct dirs

whummer:
Add small fix and test to assert that `prepare_host` hooks are called with appropriate LocalStack directories layout. This is a follow-up from https://github.com/localstack/localstack/pull/8384 . 

The previous PR #8384 adjusted the logic to start up LocalStack using the CLI when running inside a Docker container, for example to run `localstack start` in a [Github Codespaces](https://github.com/codespaces/) environment. This works well for the Community version, but unfortunately there was an oversight for the Pro version, where we're performing operations on the host via `prepare_host` hooks. In particular, the hook used to cache the key locally was failing with a `PermissionError`, as it was trying to access the wrong directory (using the container dirs layout, not the host dirs layout).
```
  File "/usr/local/python/3.10.4/lib/python3.10/site-packages/localstack_ext/bootstrap/licensing.py", line 64, in fetch_key
    C=json.loads(to_str(B.content))[_C];cache_key_locally(A,C)
  File "/usr/local/python/3.10.4/lib/python3.10/site-packages/localstack_ext/bootstrap/licensing.py", line 75, in cache_key_locally
    A=to_str(base64.b64encode(to_bytes(C)));E=get_key_cache();E.update({_E:int(B),_G:md5(api_key),_C:A});E.save()
  File "/usr/local/python/3.10.4/lib/python3.10/site-packages/localstack/utils/json.py", line 84, in save
    os.makedirs(os.path.dirname(self.path), exist_ok=True)
  File "/usr/local/python/3.10.4/lib/python3.10/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/usr/local/python/3.10.4/lib/python3.10/os.py", line 225, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/var/lib/localstack'
```

This PR contains a small follow-up change in `cmd_start(..)` to ensure `bootstrap.prepare_host(..)` is called _after_ we're performing the `config.*` adjustments introduced in #8384 . The change has been tested in Github Codespaces, which now allows to spin up a Pro container via the CLI if the API key is configured. A unit test is added to cover the functionality, with assertions about the "is in Docker" status and the accessibility of the `dirs.cache` directory.

Note: The PR also contains a few minor cleanups/refactorings (add type hints, f-strings, etc) along the way.

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -375,12 +375,13 @@ def cmd_start(docker: bool, host: bool, no_banner: bool, detached: bool) -> None
         else:
             console.log("starting LocalStack in Docker mode :whale:")
 
-    bootstrap.prepare_host(console)
-
     if not no_banner and not detached:
         console.rule("LocalStack Runtime Log (press [bold][yellow]CTRL-C[/yellow][/bold] to quit)")
 
     if host:
+        # call hooks to prepare host
+        bootstrap.prepare_host(console)
+
         # from here we abandon the regular CLI control path and start treating the process like a localstack
         # runtime process
         os.environ["LOCALSTACK_CLI"] = "0"
@@ -402,6 +403,9 @@ def cmd_start(docker: bool, host: bool, no_banner: bool, detached: bool) -> None
         config.OVERRIDE_IN_DOCKER = False
         config.is_in_docker = False
         config.dirs = config.init_directories()
+
+        # call hooks to prepare host (note that this call should stay below the config overrides above)
+        bootstrap.prepare_host(console)
 
         if detached:
             bootstrap.start_infra_in_docker_detached(console)
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -244,7 +244,7 @@ def is_api_enabled(api: str) -> bool:
         return True
 
     for enabled_api in apis:
-        if api.startswith("%s:" % enabled_api):
+        if api.startswith(f"{enabled_api}:"):
             return True
 
     return False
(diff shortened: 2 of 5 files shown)


localstack/cli/localstack.py
```diff
@@ -403,6 +404,9 @@ def cmd_start(docker: bool, host: bool, no_banner: bool, detached: bool) -> None
         config.is_in_docker = False
         config.dirs = config.init_directories()
 
+        # call hooks to prepare host (note that this call should stay below the config overrides above)
```
whummer, line 407: This is effectively the main change in the PR - calling the hooks _after_ we've adjusted the `config` values above.

localstack/utils/bootstrap.py
```diff
@@ -295,12 +295,10 @@ def validate_localstack_config(name: str):
     image_name = ls_service_details.get("image", "")
     if image_name.split(":")[0] not in constants.OFFICIAL_IMAGES:
         warns.append(
-            'Using custom image "%s", we recommend using an official image: %s'
-            % (image_name, constants.OFFICIAL_IMAGES)
+            f'Using custom image "{image_name}", we recommend using an official image: {constants.OFFICIAL_IMAGES}'
         )
 
     # prepare config options
-    image_name = ls_service_details.get("image")
```
whummer, line 303: nit: removed unused variable (duplicate from line 295 above, and not used further below)

localstack/utils/run.py
```diff
@@ -153,7 +153,7 @@ def _worker(*_args):
 
 def is_command_available(cmd: str) -> bool:
     try:
-        run("which %s" % cmd, print_error=False)
+        run(["which", cmd], print_error=False)
```
whummer, line 156: minor: changing to the non-shell `subprocess` execution here, to prevent potential command injection.

thrau (approved):
nice catch. changes make sense to me!

PR #8589, opened 2023-06-28
use docker system info for client metadata if possible

thrau:
This PR uses the docker socket (if available) to determine a machine id and the system. This gives us information about the host rather than only about the container. The machine ids are now also prefixed so we can determine how they were generated. Note that we always hash any real machine IDs for privacy purposes.

* `gen_` generated by localstack
* `dkr_` docker daemon id
* `sys_` machine id from the system file `/etc/machine-id`

Moreover, the PR uses the `LOCALSTACK_AUTH_TOKEN` as `api_key` field if available in preparation for our new licensing scheme.

### DIFF
--- a/localstack/utils/analytics/metadata.py
+++ b/localstack/utils/analytics/metadata.py
@@ -2,6 +2,7 @@ import dataclasses
 import logging
 import os
 import platform
+from typing import Optional
 
 from localstack import config, constants
 from localstack.runtime import hooks
@@ -48,7 +49,7 @@ def read_client_metadata() -> ClientMetadata:
     return ClientMetadata(
         session_id=get_session_id(),
         machine_id=get_machine_id(),
-        api_key=read_api_key_safe() or "",  # api key should not be None
+        api_key=get_api_key_or_auth_token() or "",  # api key should not be None
         system=get_system(),
         version=get_version_string(),
         is_ci=os.getenv("CI") is not None,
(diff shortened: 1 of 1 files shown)


localstack/utils/analytics/metadata.py
```diff
@@ -101,32 +102,58 @@ def _generate_session_id() -> str:
 
 
 def _generate_machine_id() -> str:
+    try:
+        # try to get a robust ID from the docker socket
+        from localstack.utils.docker_utils import DOCKER_CLIENT
+
+        docker_id = DOCKER_CLIENT.get_system_id()
```
dfangl, line 109: We could do something like `if DOCKER_CLIENT.get_system_id() == DOCKER_CLIENT.get_system_id():` here, to check if we are in Docker (ID reliable) or in podman (generating an new one every time, then we could as well use one of the fallbacks).

localstack/utils/analytics/metadata.py
```diff
@@ -101,32 +102,58 @@ def _generate_session_id() -> str:
...
 
-    # this can potentially be useful when generated on the host using the CLI and then mounted into the container via
-    # machine.json
+    # this can potentially be useful when generated on the host using the CLI and then mounted into the
+    # container via machine.json
     try:
         if os.path.exists("/etc/machine-id"):
             with open("/etc/machine-id") as fd:
-                return md5(str(fd.read()))[:8]
+                return f"sys_{md5(str(fd.read()))[:12]}"
```
ackdav, line 123: nit: the ids will have different lengths - `short_uid()` has length 8, here we use 12. slightly neater if we unify them :shrug:

ackdav (approved):
Looks good to me - I'm not super deep into the crypto space, but I wanted to highlight that the [man pages of ubuntu](https://manpages.ubuntu.com/manpages/focal/man5/machine-id.5.html) suggest:
> If a stable unique identifier that is tied to the machine is needed for some application, the machine ID or any part of it must not be used directly. Instead the machine ID should be hashed with a cryptographic, keyed hash function, using a fixed, *application-specific key*.

if I understand it correctly, we're not using an application-specific key within `md5` yet. correct? what's your take?

localstack/utils/analytics/metadata.py
thrau: good point. i also made sure the prefixes are 3 characters long

localstack/utils/analytics/metadata.py
thrau: good idea, will add

PR #8592, opened 2023-06-29
fix ASF parsing for optional list header fields

alexrashed:
This PR fixes a small issue in the ASF parser which was caused by the latest change of `botocore`: https://github.com/boto/botocore/commit/5c337d96e3ff0833fe639be6007ff94bd330bd55
It introduces a new, optional, list-shaped header field `OptionalObjectAttributes` in `ListObjectsV2Request` (header field name `x-amz-optional-object-attributes` in the HTTP request).

Thanks a lot to @dominikschubert for the info on the issue (currently blocking the pipeline on master), a perfect writeup, and a suggestion for the fix!

### DIFF
--- a/localstack/aws/api/iam/__init__.py
+++ b/localstack/aws/api/iam/__init__.py
@@ -10,6 +10,8 @@ else:
 from localstack.aws.api import RequestContext, ServiceException, ServiceRequest, handler
 
 ActionNameType = str
+CertificationKeyType = str
+CertificationValueType = str
 ColumnNumber = int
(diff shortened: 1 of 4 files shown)


thrau (approved):
nice! thank you for the quick fix :rocket:

dominikschubert (approved):
:rocket:

alexrashed:
I'll merge this PR since one test failed (`test_get_metric_data`, only on the ARM tests), which is known to be flaky.

PR #8594, opened 2023-06-29
Improve CFn provider framework / scaffolding

dominikschubert:
Lots of minor changes in here:

- Implemented nested property type scaffolding. Previously this was only creating resources for top-level properties and didn't create generic list/dict types.
- Add versioning for generated files. This is intended to make it easier for us to detect changes when refactoring and updating various things across scaffolded resources.
- A lot of minor improvements for the template files
- Added a reference implementation for `AWS::IAM::User`
- Added a `SCHEMA` property to the resource provider. At some point we can probably remove this with a schema lookup on the registry but for now we can use this for physical resource id extraction and schema validation.


localstack/services/cloudformation/engine/quirks.py
```diff
@@ -0,0 +1,44 @@
...
+PHYSICAL_RESOURCE_ID_SPECIAL_CASES = {
+    # Example
+    # "AWS::ApiGateway::Resource": "/properties/ResourceId",
+}
+
+# You can usually find the available GetAtt targets in the official resource documentation:
+#   https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html
+# Use the scaffolded exploration test to verify against AWS which attributes you can access.
+# This mapping is not in use yet (!)
+VALID_GETATT_PROPERTIES = {
```
dominikschubert, line 22: FYI: this is not in use yet

pinzon (approved):
Amazing!!

localstack/services/cloudformation/engine/quirks.py
dominikschubert: I was debating internally if we shouldn't actually move this to the provider files anyway. It feels a bit weird to have it here since we'll have to add this for nearly every model :thinking: 

I'd probably prefer just removing this automatically from the provider implementation in case of packaging as an extension in the future then.

/cc @simonrw

dominikschubert:
@alexrashed I've added the `*.schema.json` files to the package since we need the schema information at runtime

PR #8602, opened 2023-07-03
Additional debugging for network connection failure

dfangl:
## Motivation
* We got the case where we could not find the container in the list of connected endpoints in the network right after start with this network. If this happens, we want all the information we can get.

## Changes
* Add additional debug logs in error case
* Proper defaults for different container configuration flags.
* Sneaked in using new clients for retrieving code archives

### DIFF
--- a/localstack/services/awslambda/invocation/lambda_models.py
+++ b/localstack/services/awslambda/invocation/lambda_models.py
@@ -7,7 +7,7 @@ import threading
 from abc import ABCMeta, abstractmethod
 from datetime import datetime
 from pathlib import Path
-from typing import IO, TYPE_CHECKING, Dict, Optional, TypedDict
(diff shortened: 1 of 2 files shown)


localstack/services/awslambda/invocation/lambda_models.py
```diff
@@ -168,7 +165,7 @@ def _download_archive_to_file(self, target_file: IO) -> None:
 
         :param target_file: File the code archive should be downloaded into (IO object)
         """
-        s3_client: "S3Client" = aws_stack.connect_to_service("s3", region_name="us-east-1")
+        s3_client = connect_to(region_name="us-east-1").s3
```
dominikschubert, line 168: sneaky

dominikschubert (approved):
LGTM, let's hope this will help shed some light on the recent docker issues

PR #8610, opened 2023-07-04
Update codeowners with Cris and Simon

simonrw:
As joint service owners, we should be aware of any cloudformation PRs. In particular, as part of the migration to resource providers, there is some configuration in the CloudFormation service directory. We should all be made aware when new resource providers are added to ensure that the centralised configuration is updated.


thrau (approved):
excellent! awesome to see the shared ownership :-) :rocket:

Issue #8089, opened 2023-04-06
bug: Transcribe does not store output in designated S3 bucket


kkocel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a transcription job is scheduled I specify `OutputBucketName` and `OutputKey` to store the transcription, when the transcription job is complete, I check if output transcription is put there. Unfortunately, it's not there.

### Expected Behavior

Output transcription should be stored on S3 at the given bucket and key.

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

Using Test Containers 

Reproduced in project:  https://github.com/kkocel/local-stack-transcribe-repro

### Environment

```markdown
- OS: macOS
- LocalStack: latest
```

viren-nadkarni:
Hi @kkocel, unfortunately my Kotlin skills are quite lacking. Instead I tried to reproduce this via CLI but couldn't do so:

```
$ awslocal s3 mb s3://foo
make_bucket: foo

$ awslocal s3 cp tests/integration/files/en-gb.wav s3://foo
upload: tests/integration/files/en-gb.wav to s3://foo/en-gb.wav                                                                                                                                                                              

$ awslocal transcribe start-transcription-job --transcription-job-name lorem --language-code en-GB --media MediaFileUri=s3://foo/en-gb.wav 

$ awslocal transcribe get-transcription-job --transcription-job-name lorem --query TranscriptionJob.Transcript.TranscriptFileUri
"s3://foo/4194af57.json"

$ awslocal s3 cp s3://foo/4194af57.json .
download: s3://foo/4194af57.json to ./4194af57.json             

$ cat ./4194af57.json
{"jobName": "lorem", "status": "COMPLETED", "results": {"transcripts": [{"transcript": "hello my name is stephen tried i was born in london
...
```

Furthermore this passes in our [test suite](https://github.com/localstack/localstack/blob/master/tests/integration/test_transcribe.py#L91-L93)

viren-nadkarni:
Please note that in GetTranscriptionJob, AWS returns the presigned S3 URL in `TranscriptionJob.Transcript.TranscriptFileUri`, while LocalStack returns the S3 URI.

kkocel:
@viren-nadkarni I specify `outputBucketName` and `outputKey` in `startTranscriptionJob`. This means that output transcription should be put in the output S3 bucket. I think this functionality may be missing from LocalStack.

viren-nadkarni:
[...]

PR #8612, opened 2023-07-04
transcribe: fix transcription file uri

sannya-singal:
This PR focuses to fix a special case of `TranscriptFileUri` where output key is not null and the file doesn't end with .json. 
The earlier implementation was as per the AWS transcribe docs, but the actual AWS behaviour was different. 

FYI: we are currently skipping the check for `TranscriptFileUri` in all the test cases of transcribe due to other parity issues in the service. 

This will fix a part of #8089.

### DIFF
--- a/localstack/services/transcribe/provider.py
+++ b/localstack/services/transcribe/provider.py
@@ -354,4 +354,3 @@ class TranscribeProvider(TranscribeApi, ServiceLifecycleHook):
             job["TranscriptionJobStatus"] = TranscriptionJobStatus.FAILED
 
             LOG.exception("Transcription job %s failed: %s", job_name, job["FailureReason"])
-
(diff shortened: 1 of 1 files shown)


viren-nadkarni:
@sannya-singal: Did we ever hear back from AWS on this discrepancy between their docs and implementation?

sannya-singal:
> @sannya-singal: Did we ever hear back from AWS on this discrepancy between their docs and implementation?

I still had to triage and confirm a couple of things earlier before shooting them an email. Did the required investigations today. Planning to send them a message tomorrow with proper data that we have now @viren-nadkarni.

sannya-singal:
> LGTM 👍
> 
> Also consider adding a note to our transcribe docs

About the skipped check for `TranscriptFileUri`? @viren-nadkarni

viren-nadkarni:
> > LGTM 👍
> > Also consider adding a note to our transcribe docs
> 
> About the skipped check for `TranscriptFileUri`? @viren-nadkarni

That LS follows actual AWS behaviour and not what AWS docs say.

Issue #8097, opened 2023-04-07
bug: Transcribe does not return a presigned S3 URL to transcript


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The `GetTranscriptionJob` response contains an S3 URI.

```
{
    "TranscriptionJob": {
        "TranscriptionJobName": "lorem",
        "TranscriptionJobStatus": "COMPLETED",
        "LanguageCode": "en-GB",
        "MediaSampleRateHertz": 22050,
        "MediaFormat": "wav",
        "Media": {
            "MediaFileUri": "s3://foo/en-gb.wav"
        },
        "Transcript": {
            "TranscriptFileUri": "s3://foo/4194af57.json"  <<< THIS IS WRONG
        },
        "StartTime": "2023-04-07T17:44:10.405007+05:30",
        "CreationTime": "2023-04-07T17:44:10.404148+05:30",
        "CompletionTime": "2023-04-07T17:44:12.194050+05:30"
    }
}
```

### Expected Behavior

The `GetTranscriptionJob` response contains a presigned S3 URL valid for 15 minutes.

```
{
    "TranscriptionJob": {
        "TranscriptionJobName": "my-first-transcription-job",
        "TranscriptionJobStatus": "COMPLETED",
        "LanguageCode": "en-US",
        "MediaSampleRateHertz": 48000,
        "MediaFormat": "flac",
        "Media": {
            "MediaFileUri": "s3://DOC-EXAMPLE-BUCKET/my-input-files/my-media-file.flac"
        },
        "Transcript": {
            "TranscriptFileUri": "https://s3.the-URI-where-your-job-is-located.json"  <<< IT SHOULD BE A PRESIGNED S3 URL
        },
        "StartTime": "2022-03-07T15:03:44.246000-08:00",
        "CreationTime": "2022-03-07T15:03:44.229000-08:00",
        "CompletionTime": "2022-03-07T15:04:01.158000-08:00",
        "Settings": {
            "ChannelIdentification": false,
            "ShowAlternatives": false
        }
    }
}
```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```
$ awslocal s3 mb s3://foo
make_bucket: foo

[...]

PR #8619, opened 2023-07-05
transcribe: add presign url to transcript file uri

sannya-singal:
The PR fixes `GetTranscriptionJob` response for transcribe service that contains a presigned S3 URL valid for 15min.

Related issue #8097.

FYI: The related developer hub [sample for transcribe](https://github.com/localstack/sample-serverless-transcribe), needs to be changed accordingly.

### DIFF
--- a/localstack/services/s3/utils.py
+++ b/localstack/services/s3/utils.py
@@ -200,7 +200,7 @@ def get_bucket_and_key_from_s3_uri(s3_uri: str) -> Tuple[str, Optional[str]]:
     return output_bucket, output_key
 
 
-def get_bucket_and_key_from_presign_url(presign_url: str) -> Tuple[str, Optional[str]]:
+def get_bucket_and_key_from_presign_url(presign_url: str) -> Tuple[str, str]:
     """
     Extracts the bucket name and key from s3 presign url
     """
--- a/localstack/services/s3/utils.py
+++ b/localstack/services/s3/utils.py
@@ -200,6 +200,16 @@ def get_bucket_and_key_from_s3_uri(s3_uri: str) -> Tuple[str, Optional[str]]:
     return output_bucket, output_key
 
 
+def get_bucket_and_key_from_presign_url(presign_url: str) -> Tuple[str, Optional[str]]:
+    """
+    Extracts the bucket name and key from s3 presign url
+    """
+    parsed_url = urlparser.urlparse(presign_url)
+    bucket = parsed_url.path.split("/")[1]
+    key = "/".join(parsed_url.path.split("/")[2:]).split("?")[0]
+    return bucket, key
+
+
 def _create_invalid_argument_exc(
     message: Union[str, None], name: str, value: str, host_id: str = None
 ) -> InvalidArgument:
--- a/localstack/services/transcribe/provider.py
+++ b/localstack/services/transcribe/provider.py
@@ -30,7 +30,10 @@ from localstack.aws.api.transcribe import (
 )
 from localstack.packages.ffmpeg import ffmpeg_package
 from localstack.services.plugins import ServiceLifecycleHook
-from localstack.services.s3.utils import get_bucket_and_key_from_s3_uri
+from localstack.services.s3.utils import (
+    get_bucket_and_key_from_presign_url,
+    get_bucket_and_key_from_s3_uri,
+)
 from localstack.services.transcribe.models import TranscribeStore, transcribe_stores
 from localstack.utils.aws import aws_stack
 from localstack.utils.files import new_tmp_file
@@ -136,7 +139,16 @@ class TranscribeProvider(TranscribeApi, ServiceLifecycleHook):
         if not output_key:
             output_key = f"{job_name}.json"
 
-        transcript = Transcript(TranscriptFileUri=f"s3://{output_bucket}/{output_key}")
+        s3_client = aws_stack.connect_to_service("s3")
+
+        # the presign url is valid for 15 minutes
+        presign_url = s3_client.generate_presigned_url(
+            "put_object",
+            Params={"Bucket": output_bucket, "Key": output_key},
+            ExpiresIn=60 * 15,
+        )
+
+        transcript = Transcript(TranscriptFileUri=presign_url)
 
         job = TranscriptionJob(
             TranscriptionJobName=job_name,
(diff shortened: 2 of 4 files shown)


localstack/services/transcribe/provider.py
```diff
@@ -136,7 +139,16 @@ def start_transcription_job(
...
 
-        transcript = Transcript(TranscriptFileUri=f"s3://{output_bucket}/{output_key}")
+        s3_client = aws_stack.connect_to_service("s3")
+
+        # the presign url is valid for 15 minutes
+        presign_url = s3_client.generate_presigned_url(
+            "put_object",
+            Params={"Bucket": output_bucket, "Key": output_key},
+            ExpiresIn=60 * 15,
+        )
```
bentsku, line 149: question: what is this presigned URL for? Is it for the client to then retrieve the result of the job? should it be `get_object` then?
Also, the related issue says it's when calling `GetTranscriptionJob`, is the pre-signed URL generated every time we call `GetTranscriptionJob`? Because now it would be valid for 15min when starting the job, and when calling `GetTranscriptionJob` 20min after the beginning of the job, the URL would be expired. Is that how it works?

bentsku (commented):
Thanks for fixing this! I have a question regarding the pre-signed URL and what it is used for, as well as the expiration, if it's every time you can `GetTranscriptionJob`. Do we have any tests covering this? I know it's pretty tricky with snapshots, but maybe we could add some manual asserts on an AWS validated test? Parsing the returned URL and validating our assumptions?

localstack/services/s3/utils.py
```diff
@@ -200,6 +200,16 @@ def get_bucket_and_key_from_s3_uri(s3_uri: str) -> Tuple[str, Optional[str]]:
     return output_bucket, output_key
 
 
+def get_bucket_and_key_from_presign_url(presign_url: str) -> Tuple[str, Optional[str]]:
```
viren-nadkarni, line 203: Please add some unit tests for this

localstack/services/s3/utils.py
```diff
@@ -200,6 +200,16 @@ def get_bucket_and_key_from_s3_uri(s3_uri: str) -> Tuple[str, Optional[str]]:
...
+    """
+    Extracts the bucket name and key from s3 presign url
+    """
+    parsed_url = urlparser.urlparse(presign_url)
+    bucket = parsed_url.path.split("/")[1]
+    key = "/".join(parsed_url.path.split("/")[2:]).split("?")[0]
+    return bucket, key
```
viren-nadkarni, line 210: Under what conditions would `key` be `None` (as indicated in the return hint)

localstack/services/transcribe/provider.py
```diff
@@ -136,7 +139,16 @@ def start_transcription_job(
         if not output_key:
             output_key = f"{job_name}.json"
 
-        transcript = Transcript(TranscriptFileUri=f"s3://{output_bucket}/{output_key}")
+        s3_client = aws_stack.connect_to_service("s3")
```
viren-nadkarni, line 142: Use the new client here

```suggestion
        s3_client = connect_to().s3
```

viren-nadkarni (changes requested):

localstack/services/s3/utils.py
sannya-singal: @viren-nadkarni eg. the presign url is: http://s3.localhost.localstack.cloud:4566/test-output-bucket-2?AWSAccessKeyId=000000000000&Signature=2Yc%2BvwhXx8UzmH8imzySfLOW6OI%3D&Expires=1688561914

localstack/services/transcribe/provider.py
sannya-singal: >  should it be `get_object` then?

Yes, the change was already in pipeline and hence the draft 😃, but thanks for pointing it out 🙏. 


> Also, the related issue says it's when calling GetTranscriptionJob, is the pre-signed URL generated every time we call GetTranscriptionJob? Because now it would be valid for 15min when starting the job, and when calling GetTranscriptionJob 20min after the beginning of the job, the URL would be expired. Is that how it works?

Yes @bentsku, its mentioned in one of the NOTE of the [AWS transcribe service docs](https://docs.aws.amazon.com/cli/latest/reference/transcribe/start-transcription-job.html).

localstack/services/s3/utils.py
sannya-singal: It was already added in the [latest commit](https://github.com/localstack/localstack/pull/8619/commits/2fb35cb63fa7cb7d672604f856bbb09cc872020c). Is something else also required? 🤔

tests/unit/test_s3.py
```diff
@@ -252,6 +252,30 @@ def test_s3_bucket_name(self):
...
+                "test-output-bucket-5",
+                "test-files/test-output.json",
+            ),
+            (
+                "http://s3.localhost.localstack.cloud:4566/test-output-bucket-2?AWSAccessKeyId=000000000000&Signature=2Yc%2BvwhXx8UzmH8imzySfLOW6OI%3D&Expires=1688561914",
+                "test-output-bucket-2",
+                "",
+            ),
+        ]
+        for url, expected_output_bucket, expected_output_key in presign_urls:
```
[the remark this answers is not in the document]
sannya-singal: I usually use parameterise, but for the entire `unit/test_s3.py` the above pattern was used. Let me make the changes.

localstack/services/transcribe/provider.py
bentsku: > Note
> Temporary URIs for service-managed Amazon S3 buckets are only valid for 15 minutes. If you get an AccesDenied > > error, you can get a new temporary URI by running a GetTranscriptionJob or ListTranscriptionJob request.

So it is what I was saying, the URL is generated when calling `GetTranscriptionJob` and not when starting the job I think. I think that's a big logical change and we should do it too in `GetTranscriptionJob`.
Also, we're returning the `Transcript` field from `StartTranscriptionJob` but it's not supposed to be there, right?

localstack/services/transcribe/provider.py
```diff
@@ -136,7 +140,16 @@ def start_transcription_job(
...
 
-        transcript = Transcript(TranscriptFileUri=f"s3://{output_bucket}/{output_key}")
+        s3_client = connect_to().s3
+
+        # the presign url is valid for 15 minutes
+        presign_url = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": output_bucket, "Key": output_key},
+            ExpiresIn=60 * 15,
+        )
```
bentsku, line 150: I think we should re-generate the `TranscriptFileUri` for `GetTranscriptionJob` too so that we extend the validity of the pre-signed URL

bentsku (changes requested):
Almost good to go! Just the logical change with the pre-signed URL that is generated every time you call `GetTranscriptionJob` and it's good for me 🚀

localstack/services/transcribe/provider.py
sannya-singal: Thanks @bentsku. Have made the required changes.

bentsku: Awesome, thanks!

bentsku (approved):
LGTM! Thanks for addressing the comments so fast! 🚀

viren-nadkarni (approved):

PR #8621, opened 2023-07-05
Allow multiple lambda docker networks configured

### LINKED ISSUE #7338: feature request: Pass all networks assigned from localstack to lambda container

### Feature description

Allow lambda container to have multiple networks not just one.
With the current implementation Lambda network only gets a single network, even though the Localstack container accepts multiple network values.
this is how it works currently:

> Gets the main network of the LocalStack container (if we run in one, bridge otherwise)
    If there are multiple networks connected to the LocalStack container, we choose the first as "main" network

### 🧑‍💻 Implementation

This is where the "main" network value is retrieved, may be it should be a List?
https://github.com/localstack/localstack/blob/a71e661fa6b0189e54b468713f710f0ae1b93791/localstack/utils/container_networking.py#L14

### Anything else?

_No response_

dfangl:
## Motivation
This PR will extend the functionality of `LAMBDA_DOCKER_NETWORK` to provide more than one network, as requested in #7338 .

Networks need to be provided comma separated, like this: `LAMBDA_DOCKER_NETWORK=network1,network2`.
As long as only one network is provided, the behavior does not change.

Fixes #7338.

## Changes
* Allow multiple docker networks to be respected (only new provider)
* Add test for LAMBDA_DOCKER_NETWORK with multiple networks.

## TODOs
* After this is merged, we should document this.

### DIFF
--- a/localstack/services/awslambda/invocation/docker_runtime_executor.py
+++ b/localstack/services/awslambda/invocation/docker_runtime_executor.py
@@ -21,7 +21,7 @@ from localstack.services.awslambda.invocation.runtime_executor import (
 )
 from localstack.services.awslambda.lambda_utils import (
     HINT_LOG,
-    get_container_network_for_lambda,
+    get_all_container_networks_for_lambda,
     get_main_endpoint_from_container,
 )
 from localstack.services.awslambda.packages import awslambda_runtime_package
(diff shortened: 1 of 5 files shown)


dominikschubert (approved):
Love the added test :+1:

PR #8622, opened 2023-07-05
Add status: in progress mapping to label -> project status sync

dfangl:
This is a first try, as the emoji part of the target is a bit tricky since it is a zero-width joined emoji, in contrast to the other once.

According to the github preview, however, it seems to render correctly, still not sure if it will create an issue with finding the correct category.


alexrashed (approved):
LGTM! 🚀

Issue #6468, opened 2022-07-18
All resources listed as ResourceType `AWS::CloudFormation::Stack` when deploying a stack


specious:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When deploying a stack with [samlocal](https://github.com/localstack/aws-sam-cli-local/), the changeset looks correct with all resources labelled as their correct resource type, but the stack creation comes back as ResourceType `AWS::CloudFormation::Stack` for each resource rather than its actual resource type:

```
$ samlocal deploy -g

...

Previewing CloudFormation changeset before deployment
======================================================
Deploy this changeset? [y/N]: y

2022-07-16 18:59:45 - Waiting for stack create/update to complete

CloudFormation events from stack operations (refresh every 0.5 seconds)
-------------------------------------------------------------------------------------------------------------------------------------
ResourceStatus                    ResourceType                      LogicalResourceId                 ResourceStatusReason            
-------------------------------------------------------------------------------------------------------------------------------------
CREATE_COMPLETE                   AWS::CloudFormation::Stack        ScaleImageFunctionRole            -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        AddLocationFunctionRole           -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        GetLocationFunctionRole           -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        InfoFunctionRole                  -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        LoginFunctionRole                 -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        AuthFunctionRole                  -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        GetUploadLinkFunctionRole         -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        DBAccessFunctionRole              -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        AllowDBAccessPolicy               -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        MainDB                            -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        UsersTable                        -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        SessionsTable                     -                               
CREATE_COMPLETE                   AWS::CloudFormation::Stack        EventsTable                       -                               
[...]

PR #8626, opened 2023-07-05
add CREATE_IN_PROGRESS stack events

pinzon:
While triaging #6468 which has been already fixed. I found out that our template deployer wasn't generating the CREATE_IN_PROGRESS events and also it was creating UPDATE_IN_PROGRESS for conditionals and parameters even in a create stack situation. 

This changes fix all that.


pinzon:
Close due to fix already submitted in another PR

PR #8628, opened 2023-07-05
fix OpenSearch proxy for GZIP encoded responses

### LINKED ISSUE #7742: OpenSearch is not gzipped when Accept-Encoding: gzip is defined

### Current Behavior

In version 1.4.0 OpenSearch is returning invalid responses when specifying gzip encoding. This was working in 1.3.1. The proxied response is not gzip encoded even though the content-encoding response header says its gzip. Reproduction steps show the differences between the output directly from OpenSearch vs the proxied result.

### Expected Behavior

Returning properly encoded responses.

With a docker-compose file

### Steps To Reproduce

```
version: "3.9"

services:
  opensearch:
    container_name: opensearch
    image: opensearchproject/opensearch:2.3.0
    environment:
      - node.name=opensearch
      - cluster.name=opensearch-docker-cluster
      - discovery.type=single-node
      - bootstrap.memory_lock=true
      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
      - "DISABLE_SECURITY_PLUGIN=true"
    ports:
      - "9200:9200"
    ulimits:
      memlock:
        soft: -1
        hard: -1
    volumes:
      - data01:/usr/share/opensearch/data

  localstack:
    image: localstack/localstack
    ports:
      - "4566:4566"
    depends_on:
      - opensearch
    environment:
      - OPENSEARCH_CUSTOM_BACKEND=http://opensearch:9200
      - DEBUG=${DEBUG- }
      - PERSISTENCE=${PERSISTENCE- }
    volumes:

volumes:
  data01:
    driver: local
```

1. Run docker compose
```sh
docker-compose up -d
```

2. Create the OpenSearch domain:
```
awslocal opensearch create-domain --domain-name my-domain
```

3. Test - Notice content length and output
```
curl -v -X GET my-domain.us-east-1.opensearch.localhost.localstack.cloud:4566/  -H 'Accept-Encoding: gzip' --output -
curl -v -X GET localhost:9200/ -H 'Accept-Encoding: gzip' --output -
```

### Environment

```markdown
- OS:
- LocalStack: latest 1.4.0
```

### Anything else?

_No response_

alexrashed:
This PR fixes a small issue where GZIP encoded responses from an OpenSearch backend were decompressed by the proxies before sending them to the requesting client.
This is fixed by using the `SimpleStreamingRequestsClient` introduced by @bentsku in #8148.
It also enables the GZIP feature on internal OpenSearch domains (which is only used if the client sends the appropriate `Accept-Encoding` header).
In the future we could think about removing this separate implementation and making this the default implementation.

Fixes #7742

### DIFF
--- a/localstack/services/opensearch/cluster.py
+++ b/localstack/services/opensearch/cluster.py
@@ -14,7 +14,7 @@ from localstack.aws.api.opensearch import (
     EngineType,
     ValidationException,
 )
-from localstack.http.client import SimpleRequestsClient
+from localstack.http.client import SimpleStreamingRequestsClient
 from localstack.http.proxy import ProxyHandler
 from localstack.services.edge import ROUTER
 from localstack.services.infra import DEFAULT_BACKEND_HOST
@@ -229,7 +229,7 @@ def register_cluster(
     forward_url = config.OPENSEARCH_CUSTOM_BACKEND or forward_url
 
     # if the opensearch security plugin is enabled, only TLS connections are allowed, but the cert cannot be verified
-    client = SimpleRequestsClient()
+    client = SimpleStreamingRequestsClient()
     client.session.verify = False
     endpoint = ProxyHandler(forward_url, client)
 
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! I'm glad to see my client being used for something else than S3, but for the same problem 😄.
I agree that it should become the default implementation, it was a first step to have a quick fix at the time, same as here, but we should try a CI run with this streaming one at some point to see if would break anything!

PR #8629, opened 2023-07-05
fix ApiGateway Model schema from CFn/CDK

### LINKED ISSUE #7060: bug: ApiGateway Model schema stored in Wrong JSON format

### Current Behavior

APiGateway Model that being created by CDK project stores the JSON schema with a single quotes instead of double quotes which leading to an error in parsing this json object. 

you can refere to this article to see the steps 

https://dev.to/dvddpl/reduce-lambda-invocations-and-code-boilerplate-by-adding-a-json-schema-validator-to-your-api-gateway-15af

and it works fine on AWS account 

also you can look on the attached image how localstack stores the JSON object with a single quotes, I tried to edit it manually through UI but it return it back with single quotes

Screen Shot 2022-10-18 at 4 28 08 PM

### Expected Behavior

store the json schema in correct JSON format

With the `localstack` script

    localstack start

    cdklocal deploy

### Environment

```markdown
- OS:OSX
- LocalStack: 1.0.4
```

### Anything else?

_No response_

bentsku:
This PR will fix #7060 and a ticket from support. 
It appeared we store the JSON schema from Model directly as is, in Python `dict` format to `str`, without JSON encoding it, which made every request fail when validating against the schema.

### Solution:
- Added a method to jsonify the model before storing it
- snapshot validated tests uncovered additional issues:
  - we didn't check for `ANY` method in `RequestValidator`
  - we were missing `requestParameters` empty value if it was an empty `dict`
  - we were missing `operationName` field when creating `Method` from Cfn
  - the `PhysicalResourceId` of a `Model` is its name, and not its `id` (that's what you are using when calling `GetModel`, added a AWS validated test to verify this behaviour, a CDK stack uncovered the issue)

### DIFF
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -66,7 +66,7 @@ class RequestValidator:
             return True
 
         # check if there is validator for the resource
-        resource = resource_methods.get(self.context.method, {}) or resource_methods.get("ANY", {})
+        resource = resource_methods.get(self.context.method, resource_methods.get("ANY", {}))
         if not (resource.get("requestValidatorId") or "").strip():
             return True
 
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -62,11 +62,11 @@ class RequestValidator:
             return True
 
         resource_methods = self.context.resource["resourceMethods"]
-        if self.context.method not in resource_methods:
+        if self.context.method not in resource_methods and "ANY" not in resource_methods:
             return True
 
         # check if there is validator for the resource
-        resource = resource_methods[self.context.method]
+        resource = resource_methods.get(self.context.method, {}) or resource_methods.get("ANY", {})
         if not (resource.get("requestValidatorId") or "").strip():
             return True
 
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -457,7 +457,26 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
     ) -> Method:
         response: Method = call_moto(context)
         remove_empty_attributes_from_method(response)
-        remove_empty_attributes_from_integration(response.get("methodIntegration"))
+        method_integration = response.get("methodIntegration")
+        if method_integration:
+            remove_empty_attributes_from_integration(method_integration)
+            # moto will not return `responseParameters` field if it's not truthy, but AWS will return an empty dict
+            # if it was set to an empty dict
+            if "responseParameters" not in method_integration:
+                moto_rest_api = get_moto_rest_api(context, rest_api_id)
+                moto_resource = moto_rest_api.resources[resource_id]
+                moto_method_integration = moto_resource.resource_methods[
+                    http_method
+                ].method_integration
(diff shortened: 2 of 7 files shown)


localstack/services/apigateway/invocations.py
```diff
@@ -62,11 +62,11 @@ def is_request_valid(self) -> bool:
             return True
 
         resource_methods = self.context.resource["resourceMethods"]
-        if self.context.method not in resource_methods:
+        if self.context.method not in resource_methods and "ANY" not in resource_methods:
             return True
 
         # check if there is validator for the resource
-        resource = resource_methods[self.context.method]
+        resource = resource_methods.get(self.context.method, {}) or resource_methods.get("ANY", {})
```
calvernaz, line 69: nit:
```suggestion
        resource = resource_methods.get(self.context.method, resource_methods.get("ANY", {}))
```

localstack/services/apigateway/provider.py
```diff
@@ -457,7 +457,27 @@ def get_method(
     ) -> Method:
         response: Method = call_moto(context)
         remove_empty_attributes_from_method(response)
-        remove_empty_attributes_from_integration(response.get("methodIntegration"))
+        method_integration = response.get("methodIntegration")
+        if method_integration:
```
calvernaz, line 461: nit:
```suggestion
        if method_integration := response.get("methodIntegration")
```

localstack/services/apigateway/provider.py
```diff
@@ -457,7 +457,27 @@ def get_method(
...
+            remove_empty_attributes_from_integration(method_integration)
+            # moto will not return `responseParameters` field if it's not truthy, but AWS will return an empty dict
+            # if it was set to an empty dict
+            if "responseParameters" not in method_integration:
+                moto_rest_api = get_moto_rest_api(context, rest_api_id)
+                moto_resource = moto_rest_api.resources[resource_id]
```
calvernaz, line 467: should we edge this against missing keys? since the inputs are user inputs I'd make sure we edge it against invalid values.

calvernaz (approved):
LGTM :rocket:, this is not the first time that "ANY" comes into consideration and is missing. We need to have that in mind.

localstack/services/apigateway/provider.py
bentsku: good catch, I was first aiming for 2 conditions and when I removed one I forgot the walrus :D Thanks!

localstack/services/apigateway/provider.py
bentsku: I think we're calling `call_moto` before that, so if the resources don't exist, we would raise an exception, we're directly accessing the resource that is in the response. I can always guard against it, but it felt a bit overkill. We never know with concurrent accesses I guess!

bentsku: I could add a `try...except` clause in case of concurrent accesses, but if any resources in the try block does not exist, it means it's been modified between `call_moto` and accessing those resources. We could pass this small parity hack in that case. But in a lot of provider methods, we are accessing the moto methods directly. If it becomes a problem because of concurrent access, we might do it, but I think before that it would complicate the code quite much.

Issue #5345, opened 2022-01-24
bug: Using double quotes around etag in `If-None-Match` changes `content-type` in response


p-himik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using double quotes around etag in `If-None-Match` changes `content-type` in response

### Expected Behavior

Double quotes around etag should not affect anything. In fact, they should be the only thing that works since double quotes around etag in `If-None-Match` are required.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```bash
awslocal s3 mb s3://etag-test
# Generating a 1 second silent mp3, but any mp3 will do
ffmpeg -f lavfi -i anullsrc=r=44100:cl=mono -t 1 -b:a 32k -acodec libmp3lame silence.mp3
awslocal s3 cp silence.mp3 s3://etag-test
curl -v http://localhost:4566/etag-test/silence.mp3 --output /dev/null # [1]
# Using etag from [1] above
curl -H 'If-None-Match: 5782c73c28075f7db7c89348ad2057a7' -v http://localhost:4566/etag-test/silence.mp3 --output /dev/null  # [2]
curl -H 'If-None-Match: "5782c73c28075f7db7c89348ad2057a7"' -v http://localhost:4566/etag-test/silence.mp3 --output /dev/null  # [3]
```
Here's a diff of headers from [2] and [3] above:
```diff
 > Host: localhost:4566
 > User-Agent: curl/7.74.0
 > Accept: */*
-> If-None-Match: 5782c73c28075f7db7c89348ad2057a7
+> If-None-Match: "5782c73c28075f7db7c89348ad2057a7"
 > 
 * Mark bundle as not supporting multiuse
 < HTTP/1.1 304 
-< content-type: audio/mpeg
+< Content-Type: text/html; charset=utf-8
 < Content-Length: 0
-< content-md5: V4LHPCgHX323yJNIrSBXpw==
-< ETag: "5782c73c28075f7db7c89348ad2057a7"
-< last-modified: Mon, 24 Jan 2022 17:07:55 GMT
-< x-amzn-requestid: YTSC0AMTYZ8HXL8Q67VW0IVYDWNFD9LSFCGQTOBTWEVLEKA5WD2D
[...]

PR #8630, opened 2023-07-05
added test case for if-match headers for s3

macnev2013:
## Test Case
- Added test cases for `if-match` and `if-none-match` headers

## Related Issue
https://github.com/localstack/localstack/issues/5345

### DIFF
--- a/tests/integration/s3/test_s3.py
+++ b/tests/integration/s3/test_s3.py
@@ -4752,6 +4752,38 @@ class TestS3:
         response = aws_client.s3.list_bucket_intelligent_tiering_configurations(Bucket=bucket)
         snapshot.match("list_bucket_intelligent_tiering_configurations_3", response)
 
+    @pytest.mark.aws_validated
+    def test_s3_get_object_headers(self, aws_client, s3_create_bucket, snapshot):
+        bucket = s3_create_bucket()
+        key = "en-gb.wav"
+        file_path = os.path.join(f"../files/{key}")
+        aws_client.s3.upload_file(file_path, bucket, key)
+        objects = aws_client.s3.list_objects(Bucket=bucket)
+        etag = objects["Contents"][0]["ETag"]
+
+        try:
+            aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch=etag)
+        except ClientError as e:
+            snapshot.match("if_none_match_err_1", e.response)
+        try:
+            aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch=etag.strip('"'))
+        except ClientError as e:
+            snapshot.match("if_none_match_err_2", e.response)
+
+        response = aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch="etag")
+        snapshot.match("if_none_match_1", response["ResponseMetadata"]["HTTPStatusCode"])
+
+        response = aws_client.s3.get_object(Bucket=bucket, Key=key, IfMatch=etag)
+        snapshot.match("if_match_1", response["ResponseMetadata"]["HTTPStatusCode"])
+
+        response = aws_client.s3.get_object(Bucket=bucket, Key=key, IfMatch=etag.strip('"'))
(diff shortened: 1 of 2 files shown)


tests/integration/s3/test_s3.py
```diff
@@ -4752,6 +4752,38 @@ def test_s3_intelligent_tier_config(self, aws_client, s3_create_bucket, snapshot
         response = aws_client.s3.list_bucket_intelligent_tiering_configurations(Bucket=bucket)
         snapshot.match("list_bucket_intelligent_tiering_configurations_3", response)
 
+    @pytest.mark.aws_validated
+    def test_s3_get_object_headers(self, aws_client, s3_create_bucket, snapshot):
+        bucket = s3_create_bucket()
+        key = "en-gb.wav"
+        file_path = os.path.join(f"../files/{key}")
```
bentsku, line 4759: nit: we can use this so that the test is runnable from anywhere
```suggestion
        file_path = os.path.join(
            os.path.dirname(__file__),
            f"../files/{key}",
        )
```

tests/integration/s3/test_s3.py
```diff
@@ -4752,6 +4752,38 @@ def test_s3_intelligent_tier_config(self, aws_client, s3_create_bucket, snapshot
...
+        aws_client.s3.upload_file(file_path, bucket, key)
+        objects = aws_client.s3.list_objects(Bucket=bucket)
+        etag = objects["Contents"][0]["ETag"]
+
+        try:
+            aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch=etag)
+        except ClientError as e:
+            snapshot.match("if_none_match_err_1", e.response)
```
bentsku, line 4767: nit: we can use `pytest.raises`, same for the one under
```suggestion
        with pytest.raises(ClientError) as e:
            aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch=etag)
        snapshot.match("if_none_match_err_1", e.value.response)
```

tests/integration/s3/test_s3.py
```diff
@@ -4752,6 +4752,38 @@ def test_s3_intelligent_tier_config(self, aws_client, s3_create_bucket, snapshot
...
+        try:
+            aws_client.s3.get_object(Bucket=bucket, Key=key, IfNoneMatch=etag.strip('"'))
+        except ClientError as e:
+            snapshot.match("if_none_match_err_2", e.response)
```
bentsku, line 4771: nit: what I was interested with those tests would be to see if AWS would return additional headers in the 304 error.
After a quick test, it seems it only returns the `etag` and `last-modified` headers from the object, nothing else except `x-amz-id-2` and `x-amz-request-id` which we already return. 
Returned value: 
```python
{'Error': {'Code': '304', 'Message': 'Not Modified'}, 'ResponseMetadata': {'RequestId': 'M1HKSF3MBSFDGW5X', 'HostId': 'Y30wm4vnx1fsQLgCjpW2xu0vsAzwal5ITs2GOnqRbmTmcekr2PTtgaSNlKTleiCZRzUY7T6OkMQ=', 'HTTPStatusCode': 304, 'HTTPHeaders': {'x-amz-id-2': 'Y30wm4vnx1fsQLgCjpW2xu0vsAzwal5ITs2GOnqRbmTmcekr2PTtgaSNlKTleiCZRzUY7T6OkMQ=', 'x-amz-request-id': 'M1HKSF3MBSFDGW5X', 'date': 'Wed, 05 Jul 2023 19:19:51 GMT', 'last-modified': 'Wed, 05 Jul 2023 19:19:48 GMT', 'etag': '"6ac5910e18a3566fe197c4380baaf7a0"', 'server': 'AmazonS3'}, 'RetryAttempts': 0}}
```
Maybe we could add a TODO or something for us to return those values at some point in the 304 exception.

bentsku (approved):
LGTM! just some minor nit to use `pytest.raises` instead of `try...except`. Thanks a lot for adding the test, we can see that we just miss 2 headers from the 304 error, which are `etag` and `last-modified`, so it's pretty minor. Really, thanks again for adding this!

PR #8631, opened 2023-07-05
xfail flaky ssm describe_parameters test

alexrashed:
This PR aims at unblocking the pipeline on master by marking the currently failing SSM test with xfail.
This is just a temprary measure until https://github.com/getmoto/moto/pull/6484 is merged.
Until then, SSM's `DescribeParameters` might break if `PutParameters` was used without a parameter type before (which is an invalid request, according to the service specifications).


bentsku (approved):
LGTM! Thanks so much for taking care of it, nice to read about the root cause. Thanks for the quick xfail before the real fix in moto.

PR #8632, opened 2023-07-05
Add basic invoke test for AWS::Lambda::Function with code package on S3

dominikschubert:
We didn't really have any tests that used `AWS::Lambda::Function` with the code uploaded to an s3 bucket. Instead the tests usually directly embedded the code in the template.


simonrw (approved):
LGTM, glad we have coverage of this feature now

PR #8635, opened 2023-07-06
Deleting custom Unpickler

giograno:
Removing our custom `Unpickler` from the current implementation, as briefly discussed with @thrau.
To the best of my understanding, `dispatch_tables` are specific to `Picklers` only ([docs](https://docs.python.org/3/library/pickle.html#pickle.Pickler.dispatch_table)). 

cc @whummer

### DIFF
--- a/localstack/state/pickle.py
+++ b/localstack/state/pickle.py
@@ -225,19 +225,6 @@ class Pickler(dill.Pickler):
         self.dispatch = dispatch
 
 
-class Unpickler(dill.Unpickler):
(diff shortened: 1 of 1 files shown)


giograno:
Failing test `TestLambdaOutgoingSdkCalls::test_lambda_send_message_to_sqs[python3.8]` does not concern this PR.

thrau (approved):
Thanks for digging into this. And sorry for not documenting it properly in the first place!

Issue #8598, opened 2023-06-30
bug: Regression: Tag value is optional, passing "" works on S3 but causes InternalError on localstack.


Greg-Myers-SB:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With the following code:
```ruby
client = bucket.client
bucket.object_versions({ prefix: id }).each do |version|
  client.put_object_tagging(
    bucket: bucket.name,
    key: id,
    version_id: version.id,
    tagging: { tag_set: [{key: "my-tag", value: ""}] }
  )
end
```
On localstack 0.12 this applies the tag successfully as per [Tag Basics](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-basics) about value being optional.
On localstack 1.4.0 and 2.1.1.dev this fails with `Aws::S3::Errors::InternalError`

```
Aws::S3::Errors::InternalError:
       exception while calling s3.PutObjectTagging: Traceback (most recent call last):
         File "/opt/code/localstack/localstack/aws/chain.py", line 90, in handle
           handler(self, self.context, response)
         File "/opt/code/localstack/localstack/aws/handlers/service.py", line 123, in __call__
           handler(chain, context, response)
         File "/opt/code/localstack/localstack/aws/handlers/service.py", line 93, in __call__
           skeleton_response = self.skeleton.invoke(context)
         File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
           return self.dispatch_request(context, instance)
         File "/opt/code/localstack/localstack/aws/skeleton.py", line 166, in dispatch_request
           result = handler(context, instance) or {}
         File "/opt/code/localstack/localstack/aws/forwarder.py", line 60, in _call
           return handler(context, req)
         File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
           return self.fn(*args, **kwargs)
         File "/opt/code/localstack/localstack/aws/api/core.py", line 167, in operation_marker
           return fn(*args, **kwargs)
         File "/opt/code/localstack/localstack/services/s3/provider.py", line 731, in put_object_tagging
           response: PutObjectTaggingOutput = call_moto(context)
         File "/opt/code/localstack/localstack/services/moto.py", line 46, in call_moto
           return dispatch_to_backend(context, dispatch_to_moto, include_response_metadata)
         File "/opt/code/localstack/localstack/aws/forwarder.py", line 120, in dispatch_to_backend
[...]

PR #8638, opened 2023-07-06
added test case for s3 object tagging

macnev2013:
S3 can now accept empty value in the tag.

## Fixes
- Added test case for S3 tagging with empty tag

## Issue
https://github.com/localstack/localstack/issues/8598


bentsku (approved):
LGTM! Nice fix!

PR #8640, opened 2023-07-06
Rework conditional lookup and dependency resolving

dominikschubert:
This rework was motivated by a recent support case where this issue surfaced. The "else-branch" in a conditional "Fn::If" statement was evaluated and tried to `ref` the corresponding resource even though the condition was `True`. 
See `cfn_if_conditional_reference.yaml` for a reconstruction of the basic issue.

During this another issue in our implementation surfaced unfortunately. This is a bit out of scope and has a workaround but we currently don't properly support type coercion the same way CloudFormation does. There is no proper official documentation on when CloudFormation converts types but generally it seems like nearly all values are converted to strings internally. An example of this can be seen again in the referenced test template. The parameter type is String but the default value and allowed values are technically YAML  booleans that are converted into strings. This behavior isn't specific to the `String` parameter type. Even numbers are converted to strings (See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html).

## Changes

- Make conditions more "static" during stack execution by saving resolved conditions in the stack object and explicitly passing them down to stateless utils. (This still needs to be integrated into the resolve_refs_recursively call, for now this only affects the dependency check on resources to limit the initial blast radius)
- Do a proper structure traversal when determining resource dependencies instead of string matching. This will probably be somewhat obsolete in the future with a richer Resource graph / Stack.
- Only consider affected `Fn::If` branch for dependencies, to avoid `ref`-ing resources that don't exist due to conditionals.
- Any fetch/update state on GenericBaseModel is skipped if the physical resource ID is not set yet to force a successful CREATE before any state fetching. This will also have the side effect of removing a lot of noise in the logs.


pinzon (approved):
:rocket:

Issue #8494, opened 2023-06-13
bug: s3 object not created during lambda get-function export


jdouliez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the command `aws[local] lambda get-function --function-name <function_name> --region <region>`, the function code zip archive is not created in the S3 bucket.

The logs say "200" but the file is not present (if the S3 bucket had been cleaned) or overwritten.

### Expected Behavior

After running the command above, the S3 object should be present : `s3://awslambda-${REGION]}-tasks/snapshots/000000000000/${FUNCTION_NAME}-${FUNCTION_ID}`.

### How are you starting LocalStack?

With a docker-compose file or standalone script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```bash
$> localstack start -d                                                                                                                                                                                          ─╯
     __                     _______ __             __
    / /   ____  _________ _/ / ___// /_____ ______/ /__
   / /   / __ \/ ___/ __ `/ /\__ \/ __/ __ `/ ___/ //_/
  / /___/ /_/ / /__/ /_/ / /___/ / /_/ /_/ / /__/ ,<
 /_____/\____/\___/\__,_/_//____/\__/\__,_/\___/_/|_|

 💻 LocalStack CLI 2.1.0

[16:01:56] starting LocalStack in Docker mode 🐳                                                                                                                                                  localstack.py:142
           preparing environment                                                                                                                                                                   bootstrap.py:630
           configuring container                                                                                                                                                                   bootstrap.py:638
           starting container                                                                                                                                                                      bootstrap.py:645
[16:01:57] detaching                                                                                                                                                                               bootstrap.py:649
```

This bug had been reproducd on the localstack standalone script and from a docker-compose env.

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```bash
╭─ /tmp ▓▒░·······························································································································································································░▒▓ ✔ ─╮
╰─ awslocal s3 ls s3://awslambda-cn-north-1-tasks                                                                                                                                                               ─╯

An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist

╭─ /tmp ▓▒░···························································································································································································░▒▓ 254 х ─╮
[...]

PR #8643, opened 2023-07-06
A try fixing Lambda persistence

dfangl:
## Motivation
With PR #8235 , the persistence tests broke.

It seems, that connecting to the bucket from different accounts, made some problems with remote fetched layers. It is not quite clear what causes this behavior (maybe some cross account problem in s3?), but it needs to be solved ASAP to unblock the pipeline.

## Changes
* All code archive buckets are moved to some random account. This is parity with AWS

@giograno

### DIFF
--- a/localstack/services/awslambda/invocation/lambda_models.py
+++ b/localstack/services/awslambda/invocation/lambda_models.py
@@ -64,6 +64,9 @@ IMAGE_MAPPING = {
 SNAP_START_SUPPORTED_RUNTIMES = [Runtime.java11, Runtime.java17]
 
 
+BUCKET_ACCOUNT = "949334387222"
+
+
 # TODO: maybe we should make this more "transient" by always initializing to Pending and *not* persisting it?
 @dataclasses.dataclass(frozen=True)
 class VersionState:
@@ -169,7 +172,7 @@ class S3Code(ArchiveCode):
         """
         s3_client = connect_to(
             region_name=AWS_REGION_US_EAST_1,
-            aws_access_key_id=self.account_id,
+            aws_access_key_id=BUCKET_ACCOUNT,
         ).s3
         extra_args = {"VersionId": self.s3_object_version} if self.s3_object_version else {}
         s3_client.download_fileobj(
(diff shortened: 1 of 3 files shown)


localstack/services/awslambda/invocation/lambda_models.py
```diff
@@ -64,6 +64,9 @@
 SNAP_START_SUPPORTED_RUNTIMES = [Runtime.java11, Runtime.java17]
 
 
+BUCKET_ACCOUNT = "949334387222"
```
joe4dev, line 67: It would be helpful to add a comment clarifying that this is just some random account

joe4dev, line 67: ✅ We don't get any conflicts in multi-account setups because the `account_id` is used as namespace

joe4dev, line 67: ⚠️ This change breaks all existing Lambda pods or persisted lambda functions. This is a breaking change and should be labeled at least as `semver:minor`

joe4dev, line 67: "Hiding" LocalStack-internal lambda buckets basically addresses this question https://github.com/localstack/localstack/issues/8494

joe4dev: The recent cross-account enhancements as well, so good to batch them: https://github.com/localstack/localstack/commit/28427b9deedb747fbe0c360a572adaa4c44312a5

joe4dev (approved):
Hiding the LocalStack-internal Lambda S3 buckets makes sense 👍

localstack/services/awslambda/invocation/lambda_models.py
dfangl: This is actually a fix for that PR, as it broke persistence for remote fetched layers. Not exactly sure why currently, but it blocks the -ext pipeline.

dfangl: Will add!

PR #8644, opened 2023-07-06
sanitize extra cors allowed origins

lukqw:
This PR adds a sanitization step to origins parsing from `EXTRA_CORS_ALLOWED_ORIGINS.

This is mainly to filter out `""` when  it's specified by accident, or intentionally.


dfangl (approved):
LGTM! But please add in the PR description why this change is necessary, e.g. the case how the excess spaces were provided.

PR #7658, opened 2023-02-10
add OpenSearch SecurityPlugin configuration

alexrashed:
This PR extends the OpenSearch and ElasticSearch service such that it supports enabling the security plugin for OpenSearch domains.

This first iteration supports the internal user database, there is no integration with IAM yet (might be implemented in later iterations, based on user feedback). Currently, only OpenSearch domains are supported. ElasticSearch domains are currently not supported (neither via the OpenSearch nor the ElasticSearch service).

A secured OpenSearch domain can be spawned with this example CLI input:
```json
{
    "DomainName": "secure-domain",
    "ClusterConfig": {
        "InstanceType": "r5.large.search",
        "InstanceCount": 1,
            "DedicatedMasterEnabled": false,
            "ZoneAwarenessEnabled": false,
            "WarmEnabled": false
        },
    "EBSOptions": {
        "EBSEnabled": true,
        "VolumeType": "gp2",
        "VolumeSize": 10
    },
    "EncryptionAtRestOptions": {
        "Enabled": true
    },
    "NodeToNodeEncryptionOptions": {
        "Enabled": true
    },
    "DomainEndpointOptions": {
        "EnforceHTTPS": true
    },
    "AdvancedSecurityOptions": {
        "Enabled": true,
        "InternalUserDatabaseEnabled": true,
        "MasterUserOptions": {
            "MasterUserName": "admin",
            "MasterUserPassword": "really-secure-passwordAa!1"
        }
    }
}
```
It can be provisioned with the following aws(local) cli command:
```bash
$ awslocal opensearch create-domain --cli-input-json file://./opensearch_domain.json
```
Once the domain is set up (`Processing: false`), the cluster can only be accessed with the given master user credentials (via HTTP basic auth):
```
$ curl -u "admin:really-secure-passwordAa!1" http://secure-domain.us-east-1.opensearch.localhost.localstack.cloud:4566/_cluster/health
{"cluster_name":"opensearch","status":"green",...}
```
Any unauthorized requests will result in an HTTP response status code 401 (Unauthorized).

This feature is implemented by:
- Setting up the TLS certificates in the config folder of the OpenSearch installations.
- Setting up initial roles, roles mappings, and internal users in the config folder of the OpenSearch installstions.
  - It defines a hidden, reserved user called `localstack-internal` which is used to manipulate the OpenSearch instance at runtime.
  - This user is used to create the initial master user once the cluster is up.
[...]

PR #8645, opened 2023-07-06
fix proxy client to respect disabled session.verify

alexrashed:
This PR fixes an issue in OpenSearch which was triggered by the following scenario:
- The OpenSearch security plugin is used (added with #7658).
- This enforces the usage of HTTPS when communicating with the OpenSearch backend.
  - However, the server cannot be verified, because it's running on `localhost` and is being addressed via `127.0.0.1`.
- A custom CA certificate bundle is set according to our docs by setting `REQUESTS_CA_BUNDLE`: https://docs.localstack.cloud/references/custom-tls-certificates/

This triggered a bug in the health check, where the requests to the backend - effectively checking if the cluster is already up - would fail due to a but in the requests library (https://github.com/psf/requests/issues/3829). This bug always sets `verify=True` on requests if `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` is set.

The issue is fixed by setting a custom default session in the proxy infrastructure which circumvents this issue.

### DIFF
--- a/localstack/http/client.py
+++ b/localstack/http/client.py
@@ -41,11 +41,27 @@ class HttpClient(abc.ABC):
         self.close()
 
 
+class _VerifyRespectingSession(requests.Session):
+    """
+    A class which wraps requests.Session to circumvent https://github.com/psf/requests/issues/3829.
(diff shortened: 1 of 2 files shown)


thrau (approved):
LGTM!

PR #8646, opened 2023-07-06
fix parameters resolving for samtranslator for API Gateway

bentsku:
This PR will fix an issue we've had with https://github.com/localstack/serverless-shopping-cart/pull/1, when calling a lambda function with an authorizer which would fail. Basically, we had the parameter default value / name instead of its value (a UserPoolArn) in this part of the template, once deployed. 

```yaml
Parameters:
  UserPoolArn:
    Type: 'AWS::SSM::Parameter::Value<String>'
    Default: '/serverless-shopping-cart-demo/auth/user-pool-arn'
Resources:
  CartApi:
    Type: AWS::Serverless::Api
    DependsOn:
      - ApiGWAccount
    Properties:
      StageName: Prod
      Auth:
        Authorizers:
          CognitoAuthorizer:
            UserPoolArn: !Ref UserPoolArn
            Identity: # OPTIONAL
              Header: Authorization # OPTIONAL; Default: 'Authorization'
```
We would end up with an authorizer looking like that:
```bash
awslocal apigateway get-authorizers --rest-api-id 62vbixi2me
{
    "items": [
        {
            "id": "75q1ly8he5",
            "name": "CognitoAuthorizer",
            "type": "COGNITO_USER_POOLS",
            "providerARNs": [
                "/serverless-shopping-cart-demo/auth/user-pool-arn"  # this is bad, it should be the ARN
            ],
            "authType": "cognito_user_pools",
            "authorizerResultTtlInSeconds": 300
        }
    ]
}
```
Even though we could see the parameter was properly resolved in the CloudFormation provider, change set and stack.

We can see that we're having an issue with the resolving of the SSM parameter for the `Auth` part of the API.

After investigating, we're using `samtranslator` to go from a `AWS::Serverless::Api` to `AWS::ApiGateway::RestApi`, which under the hood will create an API from anOpenAPI spec body, or collect the routes from the `AWS::Serverless::Function` resources and synthetize an OpenAPI body by itself, to then create a CloudFormation resource with the created body. That created body did not have the stack parameters, thus would import weird values (the default in the template).
We're now passing the stack parameters to the `samtranslator` call, so that it can also use those values and set them in the OpenAPI body. The proper fix is in `cloudformation/engine/template_preparer.py` \cc @dominikschubert 

I've also fixed the logic when importing authorizers, previously we would not create them if they were not used in a route, but AWS will create them anyway. \cc @calvernaz 

\cc @steffyP @HarshCasper


tests/integration/cloudformation/resources/test_apigateway.py
```diff
@@ -438,3 +442,111 @@ def test_api_gateway_with_policy_as_dict(deploy_cfn_template, snapshot, aws_clie
...
+        Properties:
+          Events:
+            ApiEvent:
+              Type: Api
+              Properties:
+                Path: /
+                Method: get
+                RestApiId:
+                  Ref: ApiGatewayApi
+          Runtime: python3.7
```
dominikschubert, line 501: nit: Would keep this at the latest version so we have more time until we need to update it :grin:  (if there's not another specific reason to use 3.7 here)

tests/integration/cloudformation/resources/test_apigateway.py
```diff
@@ -438,3 +442,111 @@ def test_api_gateway_with_policy_as_dict(deploy_cfn_template, snapshot, aws_clie
...
+        "$.get-resources.items..resourceMethods.OPTIONS",
+        "$..methodIntegration.cacheNamespace",
+        "$.get-authorizers.items..authorizerResultTtlInSeconds",
+    ]
+)
+def test_rest_api_serverless_ref_resolving(
+    deploy_cfn_template, snapshot, aws_client, create_parameter, create_lambda_function
+):
+    template = """
+    AWSTemplateFormatVersion: '2010-09-09'
```
dominikschubert, line 461: Any particular reason keeping the template here inline instead of a separate file?

dominikschubert (approved):
Didn't look too much into the apigateway service changes and focused on the CloudFormation & test part which LGTM :+1:

localstack/services/apigateway/helpers.py
```diff
@@ -847,61 +847,57 @@ def is_api_key_required(path_payload: dict) -> bool:
...
+                # Applicable for the authorizer of the request and jwt type only
+                authorizer["identitySource"] = identity_source
+            if identity_validation_expression := aws_apigateway_authorizer.get(
+                "identityValidationExpression"
+            ):
+                authorizer["identityValidationExpression"] = identity_validation_expression
+
+            rest_api_container.authorizers[authorizer["id"]] = authorizer
+
+            authorizers[security_scheme_name] = authorizer
```
calvernaz, line 892: this authorizers dict was initially used as a way to avoid duplicates, but then with the rest_api_container we don't need it anymore, essentially they are tracking the same thing. I believe we can remove line 892.

calvernaz (approved):
Thanks @bentsku for tackling this, it looks good! :rocket:

tests/integration/cloudformation/resources/test_apigateway.py
bentsku: Nope, just copied/pasted the sample from AWS docs. Changed 👍 thanks!

tests/integration/cloudformation/resources/test_apigateway.py
bentsku: The template was really small in the beginning, but now that it became bigger I've pulled it into a file, thanks!

localstack/services/apigateway/helpers.py
bentsku: makes sense, thanks!

bentsku: it's still practical to directly access a map of the authorizers names to the authorizers, as the `rest_api_container` maps it to their `id` which are randomly generated. I'm using it with `get_authorizer`, I guess it makes sense to keep it in the scope of the import?

PR #8648, opened 2023-07-07
Reset threadlocal storage before every request

dfangl:
## Motivation
In our handler chain, we set the account ID in the threadlocal storage only AFTER serving the router rules:
```
                handlers.serve_default_listeners,
          >     handlers.serve_edge_router_rules,
                # start aws handler chain
                handlers.inject_auth_header_if_missing,
          >     handlers.add_account_id,
                handlers.add_internal_request_params,
```
This leads to issues when services like apigatway or Appsync use the `get_aws_account_id()` while handling their route, as it will have persisted from the previous call.

Note that this is a bug within the services as well, but currently a lot of our services depend on this. For apigateway already in the works, we need to replace this behavior, as within the routes we have to find out the account id in a service specific manner, rather than relying on a wrong thread local variable.

Also, we should replace them with contextvars :)

This is basically a follow up to our previous issue with this, #8134.

## Changes
* Delete the account id threadlocal storage every time before a request (at least until we can switch to contextvars).


thrau (approved):
LGTM! i would just suggest to add it to the push_request_context so we have it contained.

the "correct" place to put it would be pop_request_context, but as discussed, the response handlers may not be executed if `chain.terminate()` is executed (for instance in CORS enforcement). in the future we can maybe add cleanup handlers that are unconditionally executed.

Issue #6654, opened 2022-08-11
bug: S3 Lifecycle not showing expiration header 


FelipeNathan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm configuring the S3 bucket to have a lifecycle, everything works fine, but when I get the object, the header `Expiration` is not showing

### Expected Behavior

According to the documentation, if the bucket have a `lifecycle configuration` enabled, the header `expiration` should appear in the response

Executing the command
```bash
$ awslocal s3api get-object help
```

We get the documentation of `get-object` command and in the `OUTPUT` section we have:
>Expiration -> (string)
>          If the object expiration is configured (see PUT  Bucket  lifecycle),
>          the  response  includes this header. It includes the expiry-date and
>          rule-id key-value pairs providing object expiration information. The
>          value of the rule-id is URL-encoded.

### How are you starting LocalStack?

With a docker-compose file

#### Created the bucket-lifecycle.json
```json
{
  "Rules": [
    {
      "ID": "My Rule ID",
      "Status": "Enabled",
      "Expiration": {
        "Days": 1
      },
      "NoncurrentVersionExpiration": {
        "NoncurrentDays": 1
      }
    }
  ]
}
```

####  Command for create the lifecycle configuration on a existing bucket
```bash
$ awslocal s3api put-bucket-lifecycle-configuration --bucket my-bucket --lifecycle-configuration file://bucket-lifecycle.json
```

#### Validate the configuration
```bash
$ awslocal s3api get-bucket-lifecycle-configuration --bucket my-bucket

output:
{
    "Rules": [
        {
            "Expiration": {
                "Days": 1
            },
            "ID": "My Rule ID",
            "Status": "Enabled",
            "NoncurrentVersionExpiration": {
                "NoncurrentDays": 1
            }
        }
    ]
}
```

#### Uploaded an object:
```bash
$ awslocal s3api put-object --bucket my-bucket --key teste.json --body bucket-lifecycle.json  --content-type application/json

output:
{
    "ETag": "\"eb702ecb4e41b09ca62934217d5f13af\""
}
```

#### Fetch the object
```bash
$ awslocal s3api get-object --bucket my-bucket --key teste.json bucket-lifecycle.json 

output:
{
    "AcceptRanges": "bytes",
    "LastModified": "2022-08-11T13:39:11+00:00",
    "ContentLength": 287,
    "ETag": "\"eb702ecb4e41b09ca62934217d5f13af\"",
[...]

PR #8651, opened 2023-07-07
add S3 Lifecycle logic and return Expiration header

bentsku:
This PR will implement the request from #6654.

### Features
We will now validate the `BucketLifecycleConfiguration` when we call `PutBucketLifecycleConfiguration`, but only for the parts we actually use (see limitations).
We will now properly return the `Expiration` header from `PutObject`, `GetObject` and `HeadObject` depending on the lifecycle configuration and its rules/filters. 

Lifecycle configuration is quite the feature, as we can see from the length of the documentation on AWS:
  - https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html
  - https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-expire-general-considerations.html
  - https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html
  - https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleRule.html
  - https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-transition-general-considerations.html#lifecycle-general-considerations-transition-sc

## Limitations
- We won't for now expire objects nor NonCurrentVersion (as AWS don't return `Expiration` headers when you specify `VersionId`). We will only return the `Expiration` header. As expiration is minimum in days, it's rare we would actually expire objects, except when a `Date` in the past is specified. We can iterate over this PR to add the feature, but the `Expiration` header is the most important of all. 
- We only validate the `BucketLifecycleConfiguration` for the parts that we actually use (`Expiration` and `Filter`), we don't validate `Transitions yet.

note: I could still add more unit tests, but I've validated a lot with the integration tests. I might do it during the week-end. 

_fix #6654_

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import os
+from collections import defaultdict
 from typing import IO, Dict, List, Optional
 from urllib.parse import parse_qs, quote, urlencode, urlparse, urlunparse
 from zoneinfo import ZoneInfo
@@ -18,6 +19,7 @@ from localstack.aws.api.s3 import (
     AnalyticsConfigurationList,
     AnalyticsId,
     Body,
+    BucketLifecycleConfiguration,
     BucketLoggingStatus,
     BucketName,
     BypassGovernanceRetention,
@@ -40,6 +42,7 @@ from localstack.aws.api.s3 import (
     DeleteObjectTaggingOutput,
     DeleteObjectTaggingRequest,
     ETag,
+    Expiration,
     GetBucketAclOutput,
     GetBucketAnalyticsConfigurationOutput,
     GetBucketCorsOutput,
@@ -71,6 +74,7 @@ from localstack.aws.api.s3 import (
     InvalidStorageClass,
     InvalidTargetBucketForLogging,
     LifecycleConfiguration,
+    LifecycleRules,
     ListBucketAnalyticsConfigurationsOutput,
     ListBucketIntelligentTieringConfigurationsOutput,
     ListMultipartUploadsOutput,
@@ -139,12 +143,13 @@ from localstack.services.s3.utils import (
     get_bucket_from_moto,
     get_header_name,
     get_key_from_moto_bucket,
-    get_lifecycle_expiration_from_object,
+    get_lifecycle_rule_from_object,
     get_object_checksum_for_algorithm,
     is_bucket_name_valid,
     is_canned_acl_bucket_valid,
     is_key_expired,
     is_valid_canonical_id,
+    serialize_expiration_header,
     validate_kms_key_id,
     verify_checksum,
 )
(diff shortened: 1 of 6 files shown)


localstack/services/s3/provider.py
```diff
@@ -1474,6 +1533,53 @@ def validate_acl_acp(acp: AccessControlPolicy) -> None:
             raise ex
 
 
+def validate_lifecycle_configuration(lifecycle_conf: BucketLifecycleConfiguration) -> None:
+    """
+    Validate the Lifecycle configuration following AWS docs
+    See https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html
+    https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html
+    :param lifecycle_conf:
+    :raises
```
thrau, line 1542:
```suggestion
    :raises MalformedXML: raised when ...
    :raises InvalidArgument: raised when ...
```

localstack/services/s3/provider.py
```diff
@@ -1558,6 +1664,29 @@ def is_object_expired(
     return is_key_expired(key_object=key_object)
 
 
+def get_expiration_header(lifecycle_rules: LifecycleRules, moto_object, object_tags):
```
thrau, line 1667:
```suggestion
def get_expiration_header(lifecycle_rules: LifecycleRules, moto_object, object_tags) -> dict:
```

localstack/services/s3/provider.py
```diff
@@ -157,6 +165,10 @@
 # max file size for S3 objects kept in memory (500 KB by default)
 S3_MAX_FILE_SIZE_BYTES = 512 * 1024
 
+# runtime cache of Lifecycle Expiration headers, as they need to be calculated everytime we fetch an object in case the
+# rules have changed
+EXPIRATION_CACHE: dict[BucketName, dict[ObjectKey, Expiration]] = defaultdict(dict)
```
thrau, line 170: if this is a transient cache, do we maybe want to make this part of the `S3Provider` instance instead of module-global? from what i can tell, only instance methods of `S3Provider`, as well as `get_expiration_header` use the cache, so we can just make `get_expiration_header` a method of the provider. would make a for a nicer encapsulation maybe?

thrau (approved):
nice job on nailing down this feature! does look quite involved, but the test coverage is fantastic as always :ok_hand: 

just have one optional comment regarding the module-global cache.

localstack/services/s3/provider.py
bentsku: Right, I didn't like the cache, thought about it in the store or in `models.py` file but didn't like either, it didn't really make sense. It being in the provider makes more sense! Thanks, will update! 👍

Issue #7555, opened 2023-01-25
Localstack S3 emulation does not check precondition headers for CopyObject operation


rlauer6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

CopyObject API does not check precondition headers such as `x-amz-copy-source-if-modified-since`. Instead it seems to always copy the object.

### Expected Behavior

The API should return a 412 response with content similar to:

```
<Error><Code>PreconditionFailed</Code><Message>At least one of the pre-conditions you specified did not hold</Message><Condition>x-amz-copy-source-If-Modified-Since</Condition><RequestId>14HR9PXJRG5B1GR5</RequestId><HostId>HztBs2jYea1fzzPWo1JWy33NQJqo1S1IsK62rkbnVEI04PnGNCi8sPVrichGQxH3FQwLA2RIyF4=</HostId></Error>
```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
aws s3  --profile localstack --endpoint-url http://s3.localhost.localstack.cloud:4566 \
  ls s3://net-amazon-s3-test/Makefile.am
2023-01-25 10:47:40        537 Makefile.am

aws s3api copy-object --copy-source net-amazon-s3-test/Makefile.am --key Makefile.am.2 \
  --bucket net-amazon-s3-test2 --profile localstack --endpoint-url http://s3.localhost.localstack.cloud:4566 \
  --copy-source-if-modified-since "Wed, 25 Jan 2023 16:16:27 GMT"
{
    "CopyObjectResult": {
        "ETag": "\"46ada9d1fe8f8a311ff4504c226e061c\"",
        "LastModified": "2023-01-25T16:18:43+00:00"
    }
}
```

### Environment

```markdown
- OS: penguin
- LocalStack: latest
```

thrau:
looks like moto doesn't have that feature, but we could add it by implementing pre-condition checking in our new provider here: https://github.com/localstack/localstack/blob/e8a7f54e012b513a1ac2b31f90ee01f86fd87731/localstack/services/s3/provider.py#L386-L396 

cc @bentsku @macnev2013 

also this actually looks like a good first issue :-)

shivanii1702:
[...]

PR #8653, opened 2023-07-07
implement Precondition checks for CopyObject

bentsku:
This PR fixes and implements #7555

## Features
This implements the different preconditions checks following the documentation here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html
It adds an AWS validated test, checking all wrong assumptions and a positive one.

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -3,6 +3,7 @@ import logging
 import os
 from typing import IO, Dict, List, Optional
 from urllib.parse import parse_qs, quote, urlencode, urlparse, urlunparse
+from zoneinfo import ZoneInfo
 
 import moto.s3.responses as moto_s3_responses
 
@@ -91,6 +92,7 @@ from localstack.aws.api.s3 import (
     ObjectLockToken,
     ObjectVersionId,
     PostResponse,
+    PreconditionFailed,
     PutBucketAclRequest,
     PutBucketLifecycleConfigurationRequest,
     PutBucketLifecycleRequest,
@@ -488,8 +490,6 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         if not config.S3_SKIP_KMS_KEY_VALIDATION and (sse_kms_key_id := request.get("SSEKMSKeyId")):
             validate_kms_key_id(sse_kms_key_id, dest_moto_bucket)
 
-        response: CopyObjectOutput = call_moto(context)
-
         src_bucket, src_key, src_version_id = extract_bucket_key_version_id_from_copy_source(
             request["CopySource"]
         )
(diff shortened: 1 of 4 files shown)


localstack/services/s3/provider.py
```diff
@@ -494,6 +494,41 @@ def copy_object(
             src_moto_bucket, key=src_key, version_id=src_version_id
         )
 
+        # see https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html
+        condition = None
+        source_object_last_modified = source_key_object.last_modified.replace(
+            tzinfo=ZoneInfo("GMT")
```
thrau, line 500: i wanted to suggest to use `datetime.timezone.utc` for this, as i think this is what we want. i can see we're using `ZoneInfo("GMT")` in other areas in the s3 codebase though. maybe we can take a note to unify this with the rest of the codebase, where we tend to use `datetime` or `dateutil`.

tests/integration/s3/test_s3.py
```diff
@@ -1816,6 +1816,101 @@ def test_s3_copy_object_with_checksum(self, s3_create_bucket, snapshot, aws_clie
...
+        put_object = aws_client.s3.put_object(
+            Bucket=s3_bucket,
+            Key=object_key,
+            Body=b"data",
+        )
+        head_obj = aws_client.s3.head_object(Bucket=s3_bucket, Key=object_key)
+        snapshot.match("head-object", head_obj)
+
+        # wait for the condition to pass
+        time.sleep(3)
```
thrau, line 1835: question: which condition exactly? is this some date expiry that is being tested here?

thrau (approved):
LGTM! couple of questions, nothing blocking

localstack/services/s3/provider.py
thrau: actually i can see that parts of S3 really operate on GTM. in that case: disregard :-)

tests/integration/s3/test_s3.py
bentsku: right yes, the `if-modified-since` needs a bit of time for the condition to pass, initially I had put 1 sec for LS and 3 for AWS but it failed in the GH actions, I don't like sleeping in test but for those case... S3 compares the `last-modified` field, but you can't set the value in the future otherwise it ignores it. So it needs to be now or less, but the object needs to be a bit earlier than that. I can update the comment while rebasing

localstack/services/s3/provider.py
bentsku: Yes, most headers from S3 really returns GMT, I think it's the format for these headers (RFC1123 maybe?). I like ZoneInfo as it's now baked it since 3.9, and I think it's the recommended way to get timezone info now?

bentsku: Also, I need to do this because moto is storing them as TZ unaware I believe, whereas the serializer/parser correctly parse the datetime, which is great

PR #8654, opened 2023-07-07
fix validation of PutBucketLogging

### LINKED ISSUE #6764: PutBucketLogging giving MalformedXML error for payload working with real AWS
Hi,

I am using below payload for PutBucketLogging api and getting MalformedXML error in response. I have checked it with real AWS and it is working perfectly fine. Am I missing something or this needs a fix in localstack?

`"LoggingEnabled": {
            "TargetBucket": "test_bucket_name",
            "TargetPrefix": "log",
            "TargetGrants": [
                {
                    "Grantee": {
                        "URI": "http://acs.amazonaws.com/groups/s3/LogDelivery",
                        "Type": "Group",
                    },
                    "Permission": "WRITE",
                },
                {
                    "Grantee": {
                        "URI": "http://acs.amazonaws.com/groups/s3/LogDelivery",
                        "Type": "Group",
                    },
                    "Permission": "READ_ACP",
                },
            ],
        }
 `

bentsku:
This PR fixes #6764.

## Features
Moto was a bit overzealous in their validation of the bucket logging configuration. We're keeping the same checks that are made in moto, without the problematic part. We could extend the validation then, as now with Object Ownership, a bucket created without config shouldn't be able to get a config with `TargetGrants`. This would be part of a bigger effort to use the same new defaults as AWS.

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -771,6 +771,20 @@ class MethodNotAllowed(ServiceException):
     ResourceType: Optional[ResourceType]
 
 
+class CrossLocationLoggingProhibitted(ServiceException):
+    code: str = "CrossLocationLoggingProhibitted"
+    sender_fault: bool = False
+    status_code: int = 403
+    TargetBucketLocation: Optional[BucketRegion]
+
+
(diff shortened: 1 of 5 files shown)


thrau (approved):
nicely done! patches and parity tests :heart:

PR #6778, opened 2022-08-29
upgrade hypercorn to 0.14.2

thrau:
This PR fixes an import error that occurred with a change in hypercorn https://github.com/pgjones/hypercorn/commit/c2608a0ab1133a11a2a8eeb659388c8dd13bc9b1.

I renamed the affected imports and also added an explicit dependency to the latest hypercorn version.

Side note: Hypercorn now supports serving WSGI apps. Looks like the ASGI/WSGI bridge isn't as mature as ours yet, but looks promising! Maybe we can get rid of some code soon.

**Update**: Looks like hypercorn is now also lowercasing the raw headers. there seems to be no way to restore the original header casings, which, however, some cases rely on, such as #4532.

**Update**: specifically, it seems h11 is no longer returning a `h11.Headers` object #6413, and it's now just a tuple of lower-case header name and value. https://github.com/localstack/localstack/blob/7c02200e449380b6d5198eb09f0da14968bb49b6/localstack/http/asgi.py#L78-L82

there seems to be no way to restore the original header casing, which means that, if AWS lambda expects headers to be cased a certain way, we'll have to first title-case and then run canonicalize_headers.

**Update**: now with hypercorn 0.14.2, the headers are correctly preserved

PR #8660, opened 2023-07-10
unpin hypercorn, set h11_pass_raw_headers

### LINKED ISSUE #6849: HTTP header casing when localstack proxies requests
In #6778 we found that AWS forwards HTTP headers to Lambdas from clients without modifying their casing. Since users are relying on this behavior, it effectively renders a web server that lowercases headers unusable as proxy for localstack. In general, I think there is a case to be made that web servers should not manipulate the casing of request headers, so that they can function as application-layer proxy to backends that make assumptions about header casing.

The issue of header casing for HTTP/1.1 in hypercorn was addressed with https://github.com/pgjones/hypercorn/issues/66, but later reverted with https://github.com/pgjones/hypercorn/issues/77. Apparently the relevant ASGI spec is being discussed here https://github.com/django/asgiref/issues/246 where I think we should make that case.

If this is a fundamental limitation of ASGI and we cannot get this into hypercorn, then we will be forced to either
- apologize to lambda users and tell them that they need to re-write their lambda code in a way that treats headers in a case-insensitive way
- fork hypercorn and maintain our own version
- use another webserver that does not manipulate HTTP headers, or at least allows access to the raw headers in some way

alexrashed:
With #6778, Hypercorn has been pinned to `0.14.2`, since hypercorn reverted a fix to preserve header casing with https://github.com/pgjones/hypercorn/commit/68e832af53b55bcd7798eac1e84c974eb5753827 / `0.14.3` (since the spec defines that they should not be preserved).
@thrau suggested a config to enable preserving the headers with https://github.com/pgjones/hypercorn/pull/117, which is contained in the latest release: `0.14.4`.

This PR unpins `hypercorn` in the `setup.cfg` and enables the header preserving in the configs.
#6778 introduced a regression test which ensures that the gateway properly preserves the HTTP header casing (which is why this PR does not contain any tests).

Fixes #6849.

### DIFF
--- a/localstack/aws/serving/hypercorn.py
+++ b/localstack/aws/serving/hypercorn.py
@@ -29,6 +29,7 @@ def serve(
     :param kwargs: any oder parameters that can be passed to the hypercorn.Config object
     """
     config = Config()
+    config.h11_pass_raw_headers = True
     config.bind = f"{host}:{port}"
     config.use_reloader = use_reloader
 
--- a/localstack/http/hypercorn.py
+++ b/localstack/http/hypercorn.py
@@ -93,6 +93,7 @@ class GatewayServer(HypercornServer):
         """
         # build server config
         config = Config()
+        config.h11_pass_raw_headers = True
         setup_hypercorn_logger(config)
 
         bind_address = ensure_list(bind_address)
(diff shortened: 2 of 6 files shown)


thrau (approved):
nice, thanks for jumping on this. dopamine! :grin:

Issue #8663, opened 2023-07-10
bug: S3 presigned url mismatch on key with equals sign in them


krisgeus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a test setup using localstack and delta-sharing docker containers where delta-sharing is generating pre-signed urls which are used to retrieve the files form s3.
When bumping to the latest localstack this started to fail with a signature mismatch error on specific keys 

I have created a small setup to test this. See attached files
[README.md](https://github.com/localstack/localstack/files/12001909/README.md)
[test-file-and-code.zip](https://github.com/localstack/localstack/files/12001924/test-file-and-code.zip)

### Expected Behavior

Localstack encoding the key before generating the signature like all other methods do

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

See attached files

### Environment

```markdown
- OS: MacOSX
- LocalStack: 2.1.0
```

bentsku:
Hello @krisgeus and thanks for your report.

Generating pre-signed URL is purely a client side operation, and LocalStack doesn't have anything to do with that. I'm not sure I follow the code that is attached.

This code only shows the generation of the pre-signed URL, is that right? And depending on the client used, they will differ?
Or is that to show that a specific part in our code does not work?

bentsku:
I'm currently writing tests for the specific use case and will come back to you with a fix soon! I can see the issue with the signature not matching now. Thanks again for the report!

krisgeus:
@bentsku Thx. Indeed the checking of the signature is what is reporting a wrong signature. The attached code was to demonstrate the different methods used by Minio, Awscli and localstack. I copied the relevant parts from each of the libraries to show the difference.
Thx for looking into it and creating a testcase (and possible fix). Let me know if you want me to verfiy or help out in any way.

bentsku:
Hi @krisgeus, thanks for trying out the PR already and confirming that it fixed your issue!

The image is now ready and available under the `localstack/localstack:latest` tag. We will have a new tagged release (`2.2`) next Thursday, which will include this fix. Thank you again for the sample to reproduce!

PR #8664, opened 2023-07-10
fix url encoding when validating pre-signed signature

bentsku:
As reported by #8663, we were not properly re-URL-encoding the request path before validating the signature of the pre-signed URL.

I've made the necessary modifications into both signature types, and simplified the HMACv1 (signature type `s3`) quite a bit by not recreating a `Request` object but instead directly recreating a `SplitResult` and the new `HTTPHeaders`. 

Added a test to validate the use case of the user and the fix.


macnev2013 (approved):
LGTM 🚀

PR #7086, opened 2022-10-25
fix internal botocore API usage

alexrashed:
[boto/botocore](https://github.com/boto/botocore) recently changed the endpoint detection: https://github.com/boto/botocore/commit/357d5a81311b9c63286423b64d54ff3d6b99534b
These changes (which are contained in [`botocore==1.28.0`](https://github.com/boto/botocore/releases/tag/1.28.0)) break some of the internal APIs we are using for testing and for our forwarding API (because the endpoint parameter became mandatory).

This PR aims at adjusting to these changes.

PR #8672, opened 2023-07-11
fix s3 op router following botocore 1.31.2

bentsku:
We're having issues with S3 following the 1.31.2 botocore update, containing a bug fix for generating pre-signed URL (https://github.com/boto/botocore/pull/2971)

Basically, this is a follow up from #7086, where there is already a good explanation of what happened with botocore stripping the bucket from the operation model, leading to issue routing the requests.

We had a workaround using the set `authPath` containing the original request URI. However, following a bug fix, this `authPath` field is now stripped of its query parameters. 

The fix consists of using the path part from the `authPath` and the query part from the `requestUri`, effectively reversing what botocore does (botocore keeps only the part part of the `requestUri` by calling `req_uri = req_uri.split('?')[0]`).


alexrashed (approved):
Thanks for jumping on this! LGTM! 🚀 🦸🏽

PR #8674, opened 2023-07-11
Remove Boto resource fixtures

viren-nadkarni:
This PR removes all use of [Boto resources](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html) and replaces it with equivalent functionality offered by our new internal client (which uses the Boto client underneath).

There are no plans to support Boto resources as part of our new internal AWS client library. Furthermore, the Boto team does not actively develop Boto resources.

### DIFF
--- a/localstack/testing/pytest/fixtures.py
+++ b/localstack/testing/pytest/fixtures.py
@@ -1731,8 +1731,12 @@ def cleanups(aws_client):
 
 @pytest.fixture(scope="session")
 def account_id(aws_client):
-    sts_client = aws_client.sts
-    return sts_client.get_caller_identity()["Account"]
+    return aws_client.sts.get_caller_identity()["Account"]
+
+
+@pytest.fixture(scope="session")
+def secondary_account_id(secondary_aws_client):
+    return secondary_aws_client.sts.get_caller_identity()["Account"]
 
 
 @pytest.hookimpl
--- a/localstack/utils/aws/aws_stack.py
+++ b/localstack/utils/aws/aws_stack.py
@@ -21,6 +21,7 @@ from localstack.constants import (
     APPLICATION_X_WWW_FORM_URLENCODED,
     ENV_DEV,
     HEADER_LOCALSTACK_ACCOUNT_ID,
+    INTERNAL_AWS_ACCESS_KEY_ID,
     LOCALHOST,
     MAX_POOL_CONNECTIONS,
     REGION_LOCAL,
@@ -194,10 +195,18 @@ def get_boto3_region() -> str:
     return boto3.session.Session().region_name
 
 
-def is_internal_call_context(headers):
+# TODO: remove this and use the `is_internal_call` property of RequestContext
+def is_internal_call_context(headers) -> bool:
     """Return whether we are executing in the context of an internal API call, i.e.,
     the case where one API uses a boto3 client to call another API internally."""
-    return HEADER_LOCALSTACK_ACCOUNT_ID in headers.keys()
+    if HEADER_LOCALSTACK_ACCOUNT_ID in headers.keys():
+        # TODO: Used by the old client, marked for removal
+        return True
+
+    if INTERNAL_AWS_ACCESS_KEY_ID in headers.get("Authorization", ""):
+        return True
+
+    return False
 
 
 def get_local_service_url(service_name_or_port: Union[str, int]) -> str:
--- a/tests/integration/s3/test_s3.py
+++ b/tests/integration/s3/test_s3.py
@@ -33,9 +33,7 @@ from localstack.config import LEGACY_S3_PROVIDER, STREAM_S3_PROVIDER
 from localstack.constants import (
     LOCALHOST_HOSTNAME,
     S3_VIRTUAL_HOSTNAME,
-    SECONDARY_TEST_AWS_ACCESS_KEY_ID,
     SECONDARY_TEST_AWS_REGION_NAME,
-    SECONDARY_TEST_AWS_SECRET_ACCESS_KEY,
     TEST_AWS_ACCESS_KEY_ID,
     TEST_AWS_SECRET_ACCESS_KEY,
 )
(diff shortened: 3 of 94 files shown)


localstack/testing/pytest/fixtures.py
```diff
@@ -278,6 +222,29 @@ def s3_bucket(s3_create_bucket, aws_client) -> str:
...
+
+        response = aws_client.s3.list_object_versions(Bucket=bucket_name)
+        object_versions = [
+            {"Key": obj["Key"], "VersionId": obj["VersionId"]}
+            for obj in response.get("Versions", [])
+        ]
+        if object_versions:
+            aws_client.s3.delete_objects(Bucket=bucket_name, Delete={"Objects": object_versions})
+
+    yield factory
```
viren-nadkarni, line 245: This new fixture replaces the functionality offered by the Boto S3 Resource

localstack/utils/aws/aws_stack.py
```diff
@@ -239,6 +244,11 @@ def connect_to_resource_external(
     """
     Generic method to obtain an AWS service resource using boto3, based on environment, region, or custom endpoint_url.
     """
+    warnings.warn(
+        "`connect_to_resource_external` is obsolete. Use `localstack.aws.connect`",
+        DeprecationWarning,
+    )
+
```
viren-nadkarni, line 251: With this PR, these utilities are no longer used in the codebase but I've added these warnings so that they're properly highlighted in the IDE.

Btw, Python 3.12 will have a decorator to do exactly this https://peps.python.org/pep-0702/

tests/integration/apigateway/test_apigateway_dynamodb.py
```diff
@@ -45,10 +44,9 @@ def test_rest_api_to_dynamodb_integration(
     table_name = table["TableName"]
 
     # insert items
-    dynamodb_table = dynamodb_resource.Table(table_name)
     item_ids = ("test", "test2", "test 3")
     for item_id in item_ids:
-        dynamodb_table.put_item(Item={"id": item_id})
+        aws_client.dynamodb.put_item(TableName=table_name, Item={"id": {"S": item_id}})
```
viren-nadkarni, line 49: The client needs the items to be typed. The Boto resource had an intrinsic [type conversion](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/dynamodb.html#valid-dynamodb-types) so it wasn't necessary earlier

tests/integration/s3/test_s3.py
```diff
@@ -6846,11 +6842,8 @@ def upload_file(size_in_kb: int):
         upload_file(9)
         upload_file(15)
 
-        objects = s3_resource.Bucket(s3_bucket).objects.all()
-        keys = []
-        for obj in objects:
-            keys.append(obj.key)
```
viren-nadkarni, line 6852: The `keys` var wasn't used, hence removed

tests/integration/test_kms.py
```diff
@@ -80,15 +80,15 @@ def test_create_alias(self, kms_create_alias, kms_create_key, snapshot):
         snapshot.match("create_alias", e.value.response)
 
     @pytest.mark.aws_validated
-    def test_create_key(self, kms_client_for_region, kms_create_key, snapshot, aws_client):
-        region = "us-east-1"
-        kms_client = kms_client_for_region(region)
-        account_id = aws_client.sts.get_caller_identity()["Account"]
```
viren-nadkarni, line 86: This call is replaced by the `account_id` fixture

localstack/testing/pytest/fixtures.py
dominikschubert: I'd probably prefer this to be a normal utility over a fixture but should be ok to leave as-is for now :+1:

tests/integration/apigateway/test_apigateway_dynamodb.snapshot.json
```diff
@@ -77,7 +85,7 @@
           "date": "date",
           "x-amz-apigw-id": "x-amz-apigw-id",
           "x-amzn-requestid": "<uuid:1>",
-          "x-amzn-trace-id": "Root=1-640e7ae9-63faf1aa71313dd3638b68fd"
+          "x-amzn-trace-id": "Root=1-64afea44-3087b2f50c440c1958c4d62e"
```
dominikschubert, line 88: This is actually an invalid snapshot (before and now) since it won't validate against AWS. The unique value here would need to be replaced via a transformer to make it valid. 

Since this hasn't been introduced in this PR, feel free to leave as-is though.

tests/integration/test_dynamodb.py
```diff
@@ -433,66 +439,70 @@ def test_more_than_20_global_secondary_indexes(
...
+        response = aws_client.dynamodb.put_item(TableName=table_name, Item=item1)
         snapshot.match("PutFirstItem", response)
 
-        response = table.put_item(Item=item2)
+        response = aws_client.dynamodb.put_item(TableName=table_name, Item=item2)
         snapshot.match("PutSecondItem", response)
 
         # clean up
-        table.delete()
+        aws_client.dynamodb.delete_table(TableName=table_name)
```
dominikschubert, line 505: nit: might be good while going through here anyway to do this in a safe manner (e.g. using the cleanups fixture instead)

dominikschubert (approved):
Small nits, otherwise LGTM


Thanks so much for tackling this. :sweat_smile: 
The resource abstraction definitely didn't bring enough value to warrant maintaining its support besides the normal clients.
Love that we now finally have a unified abstraction and are quite close to remove the rest of the `aws_stack` clients soon :rocket:

tests/integration/test_dynamodb.py
viren-nadkarni: I have a TODO note to convert the constants on top of this file into fixture which will take care of this :) I'll address these in follow up PRs

https://github.com/localstack/localstack/pull/8674/files/5701d33c5344d96114f3bdc6372c299ba8a7d1cd#diff-674a25994f44ac7d1dd2bdaff60e30cb9eb9c4b895c36d56b5c33b0a9bcfe7e7R27

viren-nadkarni: See https://github.com/localstack/localstack/pull/8747/

PR #8679, opened 2023-07-11
Update lambda runtime init to handle non-executable bootstrap files

### LINKED ISSUE #8459: bug: Lambda execution doesn't work with Golang binary lambdas: <title>

### Current Behavior

Running `awslocal lambda invoke --function-name "localStage-mediaStack-backfillGetFunction91FAA-b70d6394" "manual_ test_output.json"` even a single time causes localstack to continuously try and fail to run the lambda, the terminal hands whilst docker compose logs output:
```
localstack_main  | 2023-06-07T16:47:15.160  INFO --- [   asgi_gw_4] localstack.request.aws     : AWS sts.AssumeRole => 200
localstack_main  | 2023-06-07T16:47:16.000  INFO --- [   asgi_gw_0] localstack.request.http    : POST /_localstack_lambda/f1a13ad3d39e7c5f771f6cf85f78456d/status/f1a13ad3d39e7c5f771f6cf85f78456d/error => 202
localstack_main  | 2023-06-07T16:47:16.378  WARN --- [   asgi_gw_2] l.s.a.i.executor_endpoint  : Execution environment startup failed: {"errorMessage":"Error: fork/exec /var/task/bootstrap: no such file or directory","errorType":"Runtime.InvalidEntrypoint"}
```

The logs won't stop until I pull the container down and restart it (hence I know it's continuously retrying to no avail)

### Expected Behavior

Lambda should be executed just like it does on AWS, returning a sample return string like "Test" (the lambda is very simple)
Even if it does fail (I'm not sure why it'd fail here but not on AWS) then it should try a small finite number of times and then stop

With a docker-compose file (shown below)

### Steps To Reproduce

I don't think my code itself is the problem since it does work fine on AWS, but for a bit of context:
I use the cdk to compile the binaries and infrastructure which is deployed to LocalStack via:
    cdklocal deploy -a "cdk.out/assembly-localStage/" --all --require-approval "never"

My infra:
https://github.com/KamWithK/exSTATic-backend/tree/master/infrastructure

I can potentially create a small isolated example which illustrates the problem if needed

### Environment

```markdown
version: "3.8"

services:
  localstack:
    image: localstack/localstack:latest
    ports:
      - "127.0.0.1:4566:4566" # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559" # external services port range
    environment:
      - DEBUG=${DEBUG-}
      - PERSISTENCE=/tmp/localstack/data
      - AWS_DEFAULT_REGION=ap-southeast-2
    volumes:
    healthcheck:
      test: curl http://localhost:4566/_localstack/health
      interval: 1s
      timeout: 1s
      retries: 10
```

### Anything else?

Here's some function info:
```
{
            "FunctionName": "localStage-mediaStack-backfillGetFunction91FAA-b70d6394",
            "FunctionArn": "arn:aws:lambda:us-east-1:000000000000:function:localStage-mediaStack-backfillGetFunction91FAA-b70d6394",
            "Runtime": "provided.al2",
            "Role": "arn:aws:iam::000000000000:role/localStage-mediaStack-backfillGetFunctionServi-f0162982",
            "Handler": "bootstrap",
            "CodeSize": 14604,
            "Description": "",
            "Timeout": 3,
            "MemorySize": 128,
            "LastModified": "2023-06-07T15:58:33.766076+0000",
            "CodeSha256": "nqvy9NYt9j59ura5fEUE4QaeXOTuTiEsDNgEvsSQGdk=",
            "Version": "$LATEST",
            "TracingConfig": {
                "Mode": "PassThrough"
            },
            "RevisionId": "377a6465-e32b-4ed3-a9f1-396af7ee0d86",
            "PackageType": "Zip",
            "Architectures": [
                "x86_64"
            ],
            "EphemeralStorage": {
                "Size": 512
            },
            "SnapStart": {
                "ApplyOn": "None",
                "OptimizationStatus": "Off"
            }
        }
```

Test lambda:
```golang
package main

import (
	"github.com/aws/aws-lambda-go/lambda"
)

func HandleRequest() (string, error) {
	return "Test", nil
}

func main() {
	lambda.Start(HandleRequest)
}
```

I did find this issue which sounded similar at first, but in my case the code does run on AWS (so I think it's a different problem, put here for reference though):
https://github.com/localstack/localstack/issues/4216

Any help would be greatly appreciated!

dfangl:
## Motivation
The new lambda init version will automatically set the `bootstrap` file to executable if it is not yet.

This will fix #8459 .
Releases https://github.com/localstack/lambda-runtime-init/pull/21

It will do so unconditionally, even though AWS, in some cases, will refuse to allow a bootstrap file without executable permissions. It is yet to determine, under which circumstances this happens.

This will not work for hot reloading, as the privileges are already dropped when the reload takes place. If you want hot reloading, please make sure to set your bootstrap binary as executable after compilation.

## Changes
* Update lambda runtime init to get new bootstrap executable behavior.

### DIFF
--- a/localstack/services/awslambda/packages.py
+++ b/localstack/services/awslambda/packages.py
@@ -10,7 +10,7 @@ from localstack.utils.platform import get_arch
 
 LAMBDA_RUNTIME_INIT_URL = "https://github.com/localstack/lambda-runtime-init/releases/download/{version}/aws-lambda-rie-{arch}"
 
-LAMBDA_RUNTIME_DEFAULT_VERSION = "v0.1.18-pre"
+LAMBDA_RUNTIME_DEFAULT_VERSION = "v0.1.19-pre"
 LAMBDA_RUNTIME_VERSION = config.LAMBDA_INIT_RELEASE_VERSION or LAMBDA_RUNTIME_DEFAULT_VERSION
 
 # GO Lambda runtime


joe4dev (approved):
LGTM 👍 

We should document the limitation for hot-reloading somewhere after dropping privileges.
At least mention it when notifying https://github.com/localstack/localstack/issues/8459

As a workaround for hot-reloading, setting `LAMBDA_DOCKER_FLAGS=-e LOCALSTACK_USER=""` disables dropping privileges.

dfangl:
Sadly, with this approach, we are *only* changing the executable status on container status, once at startup. Even with more privileges, we do not try again currently.

joe4dev:
oh right, that makes sense because it is part of the bootstrapping. Hence, it does not happen within the sandbox for recurring invocations. That means we're also not getting log spamming when using hot-reloading in general. Hot reloading for non-executable bootstrap files just fails now (i.e., known limitation).

PR #8684, opened 2023-07-12
fix CmdDockerClient not raising DockerNotAvailable exception

alexrashed:
This PR performs a small fix to the `CmdDockerClient` (which is basically only used by the CLI), such that it raises a `DockerNotAvailable` exception in case the `DOCKER_CMD` (by default `docker`) is not available on the system.

Contrary to the `SdkDockerClient`, this is done only once (the result is cached) to avoid checking the command existence on every single usage of the client. We basically assume that, if the `DOCKER_CMD` is not available once, it won't become available during this python session lifetime.

This fixes the CLI error handling in this case (introduced in #8564).

### DIFF
--- a/tests/integration/docker_utils/conftest.py
+++ b/tests/integration/docker_utils/conftest.py
@@ -1,4 +1,5 @@
 import os
+from typing import Type
 
 import pytest
 
@@ -17,13 +18,15 @@ def _check_skip(client: ContainerClient):
 
 
 @pytest.fixture(
-    params=[CmdDockerClient, SdkDockerClient], ids=["CmdDockerClient", "SdkDockerClient"]
+    params=[CmdDockerClient, SdkDockerClient],
+    ids=["CmdDockerClient", "SdkDockerClient"],
+    scope="class",
 )
-def docker_client_class(request):
-    yield request.param
+def docker_client_class(request) -> Type[ContainerClient]:
+    return request.param
 
 
-@pytest.fixture
+@pytest.fixture(scope="class")
 def docker_client(docker_client_class):
     client = docker_client_class()
     _check_skip(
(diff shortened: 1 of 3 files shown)


tests/integration/docker_utils/conftest.py
```diff
@@ -17,10 +17,15 @@ def _check_skip(client: ContainerClient):
 
 
 @pytest.fixture(
-    params=[CmdDockerClient(), SdkDockerClient()], ids=["CmdDockerClient", "SdkDockerClient"]
+    params=[CmdDockerClient, SdkDockerClient], ids=["CmdDockerClient", "SdkDockerClient"]
 )
-def docker_client(request):
-    client = request.param
+def docker_client_class(request):
+    yield request.param
```
dfangl, line 23:
```suggestion
    return request.param
```
Is there a reason for yielding here?

tests/integration/docker_utils/conftest.py
```diff
@@ -17,10 +17,15 @@ def _check_skip(client: ContainerClient):
...
+
+
+@pytest.fixture
+def docker_client(docker_client_class):
+    client = docker_client_class()
```
dfangl, line 28: With the cache, creating a new docker client will yield in a bit of a memory leak, since the lru cache will hold a reference to "self", which leads to the client not removed. Before, we only had one client for all tests, now we have a new one for every test, and the CmdDockerClients will not be garbage collected. Again, not a huge leak, so probably not worth following up, but still. We could scope the docker client to class level, which would reduce that issue.

dfangl (approved):
LGTM! Just minor comments

tests/integration/docker_utils/conftest.py
alexrashed: Good catch! I added the scope :)

tests/integration/docker_utils/conftest.py
alexrashed: Good catch, I applied your suggestion in the latest commit.

dfangl (approved):

PR #8685, opened 2023-07-12
add PythonPackageInstaller for runtime package installation

alexrashed:
This PR adds a new type of `PackageInstaller` which allows the installation of Python packages at runtime for specific services.
The PoC was implemented based on our Transcribe emulator, removing the `vosk` dependency in the `setup.cfg` and instead installing it on demand.
VOSC is one of the bigger dependencies, with ~7MB compressed and ~26MB uncompressed. In my tests this reduced the uncompressed layer size of the `venv` copying layer from 426 MB to 399 MB:
https://github.com/localstack/localstack/blob/10630f07dd838c9b5a71dd087de8bbc044b417fe/Dockerfile#L144

- It introduces two new linked virtual environments, one for static (docker build time) and one for var libs (runtime, using a shared folder mounted from the host) in the `Dockerfile`.
  - This is implemented similarly to the extensions venv by adding a [`pth` file to extend the search path](https://docs.python.org/3/install/index.html#modifying-python-s-search-path) of the main virtual environment.
- It introduces a new `PackageInstaller` method which is executed once if the package is already installed (to perform specific installation instructions which are necessary to be done at runtime, even if the package has been installed in previous executions).
  - This is necessary for the `PythonPackageInstaller` to inject the sys path of the already existing virtual environment outside of a Docker image (i.e. in host mode).
- It removes `vosk` from the list of runtime extra dependencies.
- It adds a new `VoskPackage` to the `transcribe` module, which is used in the Transcribe provider to ensure that vosk is installed when needed.
  - The installation of `vosk` as well as `ffmpeg` has been moved from the .`ServiceLifecycleHook#on_before_start` to the async execution which actually performs the transcoding and transcription.
  - This should actually improve the UX, since the service starts instantly, while downloading the additional packages in an async thread when needed (instead of running into a client timeout on the first call).

This is part of our Hackathon, and this was on my backlog to test for quite a while (which is why it's a draft PR). I am happy for any feedback (also to discard the idea completely)!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -168,6 +168,12 @@ RUN --mount=type=cache,target=/root/.cache \
 RUN echo /var/lib/localstack/lib/extensions/python_venv/lib/python3.10/site-packages > localstack-extensions-venv.pth && \
     mv localstack-extensions-venv.pth .venv/lib/python*/site-packages/
 
+# link the python package installer virtual environments into the localstack venv
+RUN echo /var/lib/localstack/lib/python-packages/lib/python3.10/site-packages > localstack-var-python-packages-venv.pth && \
+    mv localstack-var-python-packages-venv.pth .venv/lib/python*/site-packages/
+RUN echo /usr/lib/localstack/python-packages/lib/python3.10/site-packages > localstack-static-python-packages-venv.pth && \
+    mv localstack-static-python-packages-venv.pth .venv/lib/python*/site-packages/
+
 # Install the latest version of the LocalStack Persistence Plugin
 RUN --mount=type=cache,target=/root/.cache \
     (. .venv/bin/activate && pip3 install --upgrade localstack-plugin-persistence)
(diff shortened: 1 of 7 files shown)


thrau:
i've noticed this is still in draft mode. @alexrashed i think this is an excellent PR and we could basically merge as is. we just need ext companion PR :+1:

thrau (approved):
i think this is an excellent PR and we could basically merge as is. we just need ext companion PR 👍 i also think the changes are non-controversial. this was exactly what the whole `VirtualEnvironment` class and linking of venvs was conceived for (next to extensions). let's :shipit:

alexrashed:
Thanks for the reassuring feedback! I just rebased the PR and put it in ready for review.
Maybe @viren-nadkarni and @sannya-singal could give me some feedback on the changes as well?

viren-nadkarni (approved):
LGTM :+1: 

Moving the Vosk download and installation bits to transcription thread is a big QoL improvement for users :raised_hands:

Issue #8556, opened 2023-06-22
bug: Terraform Lambda Function vpc_config doesn't get updated in terraform state


FarhanSajid1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Whenever an apply is ran, the security group ids and subnet ids get updated to the same value.
```
      + vpc_config {
          + security_group_ids = [
              + "sg-30710aba47f78b56c",
            ]
          + subnet_ids         = [
              + "subnet-99026414",
            ]
        }
        # (4 unchanged blocks hidden)
    }
```

Looking at the terraform state file
```
       "tracing_config": [
              {
                "mode": "Active"
              }
            ],
            "version": "3",
            "vpc_config": []
          },
```

the `vpc_config` values are empty.

### Expected Behavior

A no op should occur

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

localstack start
Create a lambda function with a vpc and security group
Reapply

### Environment

```markdown
- OS:
- LocalStack:
```

dominikschubert:
Hi @FarhanSajid1 ! This issue should be fixed in the latest version, please pull the newest image and give it a try.

PR #8686, opened 2023-07-12
Add basic CRUD support for Lambda VpcConfig

dominikschubert:
Addresses #8556

### DIFF
--- a/tests/integration/awslambda/test_lambda_api.py
+++ b/tests/integration/awslambda/test_lambda_api.py
@@ -581,7 +581,7 @@ class TestLambdaFunction:
         Test "VpcConfig" Property on the Lambda Function
 
         Note: on AWS this takes quite a while since creating a function with VPC usually takes at least 4 minutes
-
+        FIXME: Unfortunately the cleanup in this test doesn't work properly on AWS and the last subnet/security group + vpc are leaking.
         TODO: test a few more edge cases (e.g. multiple subnets / security groups, invalid vpc ids, etc.)
         """
 
--- a/tests/integration/awslambda/test_lambda_api.py
+++ b/tests/integration/awslambda/test_lambda_api.py
@@ -80,6 +80,7 @@ class TestLambdaFunction:
             MemorySize=256,
             Timeout=5,
         )
+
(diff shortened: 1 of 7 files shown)


tests/integration/awslambda/test_lambda_api.py
```diff
@@ -573,6 +574,110 @@ def test_list_functions(self, create_lambda_function, lambda_su_role, snapshot,
...
+        snapshot.add_transformer(
+            snapshot.transform.regex(security_group_id_1, "<security_group_id_1>")
+        )
+        snapshot.add_transformer(
+            snapshot.transform.regex(security_group_id_2, "<security_group_id_2>")
+        )
+
+        cleanups.append(
+            lambda: aws_client.awslambda.delete_function(FunctionName=function_name)
+        )  # needed because otherwise VPC is still linked to function and deletion is blocked
```
dominikschubert, line 622: Unfortunately this still doesn't work properly on AWS and the last subnet/security group + vpc are leaking. Not critical for this PR but we probably need to wait for the lambda to actually  disappear before this is detached an can be deleted :thinking:

tests/integration/templates/cfn_parameter_list_type.yaml
```diff
@@ -1,29 +1,14 @@
...
     Type: List<String>
-  SecurityGroupId:
-    Type: String
-  FunctionRole:
-    Type: String
 Resources:
-  Function76856677:
-    Type: AWS::Lambda::Function
+  MyParam:
+    Type: AWS::SSM::Parameter
```
joe4dev, line 8: nice test simplification 👍

tests/integration/awslambda/test_lambda_api.py
joe4dev: VPC testing is annoying 🐌 

nit: would be good to add the comment in the test such that it does not get lost in this PR discussion
💡 slow tests + improper cleanup are candidates for test labels

joe4dev (approved):
LGTM 👍 

Lambda model updates break CloudPods. Good to get this PR in now with a couple of recent Lambda model updates.

Issue #8675, opened 2023-07-11
bug: S3 SelectObjectContentCommand results in ProtocolSerializerError in LocalStack


yuvalue:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm using LocalStack to emulate AWS S3 and I'm trying to use the SelectObjectContentCommand from the AWS SDK to perform SQL-like queries on a CSV file stored in S3. However, I'm encountering a ProtocolSerializerError stating "Expected iterator for streaming event serialization".

I'm encountering the following error:
```
localstack.aws.protocol.serializer.ProtocolSerializerError: Expected iterator for streaming event serialization.
```

This issue occurs consistently every time I try to run the command. I'm not sure if it's a bug in LocalStack, an issue with the way my local environment is set up, or potentially some other unforeseen issue. Any assistance would be greatly appreciated.

### Expected Behavior

Payload async generator should properly and send events, more specifically `event.Records`

Payload does exist but I never get inside the for await loop
```
      const contentResponse = await this.s3Client.send(contentCommand);
      for await (const event of contentResponse.Payload) {
        debugger;
       }
```

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

        const localstackContainer = await new GenericContainer('localstack/localstack', 'latest')
        .withName('tests-localstack-s3')
        .withEnvironment({ SERVICES: 's3', DEBUG: 1 })
        .withExposedPorts(4566)
        .start();

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

1. I'm storing a CSV file in S3 with the following data:
```
i,d
0,aaaa
1,aaaa
2,aaaa
3,aaaa
4,aaaa
```

2. I'm running the following command on my s3 client: 
```
new SelectObjectContentCommand({
    Bucket: this.bucket,
    Key: this.key,
    Expression: 'SELECT s.d FROM S3Object s WHERE CAST(s.i AS int) >= 0 AND CAST(s.i AS int) <= 9',
    ExpressionType: 'SQL',
    InputSerialization: {
        CSV: {
            FileHeaderInfo: 'USE',
        },
    },
    OutputSerialization: {
        CSV: {},
    },
});
```

### Environment

```markdown
- OS: windows 10
- LocalStack: latest
```

### Anything else?

this is the entire error:

```
[...]

PR #8689, opened 2023-07-12
raise NotImplemented error for SelectObjectContent

bentsku:
Following the report from #8675, it seemed moto had introduced experimental support for `SelectObjectContent`. However, the format from the returned response would often end raising `ProtocolSerializerError`, as the serializer expect a stream of response (like for Kinesis, for example).

_fixes #8675_


thrau (approved):
LGTM! It's better not to pass through moto features that aren't up to our standards, especially if we have that functionality in pro :+1:

Issue #8444, opened 2023-06-05
enhancement request: support for s3:ObjectRestore:* bucket notifications


piedrahitapablo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Currently, I'm using Localstack to test locally a lambda function that takes s3`ObjectRestore:Completed` notifications as inputs and it would be really great to have support for these events.

I know that right now as a workaround I can invoke the lambda function manually using a payload with the same shape that s3 uses, but it's better to have the process run as close as it would run in AWS.

Thanks for creating and maintaining localstack, it's really great!

### 🧑‍💻 Implementation

Not sure, but happy to help if you can give me some pointers.

PR #8690, opened 2023-07-13
add ObjectRestore:Post and :Completed S3 Event Notification

bentsku:
As requested in #8444, this adds support for `ObjectRestore:Post` and `ObjectRestore:Completed`, as well as sneak in a bug fix for EventBridge where we use the time provided to the `Entry` instead of `utcnow()`. 

See: 
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html#AmazonS3-RestoreObject-request-GlacierJobParameters
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html#ev-events-object-restore-complete

_fixes #8444_

### DIFF
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -572,7 +572,10 @@ def events_handler_put_events(self):
                 event_time = datetime.datetime.utcfromtimestamp(event_timestamp)
             except ValueError:
                 # if we can't parse it, pass and keep using `utcnow`
-                pass
+                LOG.debug(
+                    "Could not parse the `Time` parameter, falling back to `utcnow` for the following Event: '%s'",
+                    event,
+                )
 
         # See https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html
         formatted_event = {
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -565,6 +565,15 @@ def events_handler_put_events(self):
         if not matching_rules:
             continue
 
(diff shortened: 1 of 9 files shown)


macnev2013 (approved):
LGTM

localstack/services/events/provider.py
```diff
@@ -565,14 +565,23 @@ def events_handler_put_events(self):
         if not matching_rules:
             continue
 
+        event_time = datetime.datetime.utcnow()
+        if event_timestamp := event.get("Time"):
+            try:
+                # if provided, use the time from event
+                event_time = datetime.datetime.utcfromtimestamp(event_timestamp)
+            except ValueError:
+                # if we can't parse it, pass and keep using `utcnow`
```
dominikschubert, line 574: 1. is this in parity with AWS or a simplification we do for LocalStack?
2. I'd recommend at least logging in this case, since users might wonder why their custom provided event time isn't working.

dominikschubert (approved):
Minor nit & question, otherwise LGTM :+1:

localstack/services/events/provider.py
bentsku: 1. good question, I think boto is validating this field so it's a bit tricky to validate with tests, and most SDK should
2. I'll add the logging, if it happens this is an issue, we can try to make a more complex test case. I didn't want to introduce a breaking change, so I fell back to the previous behaviour.

PR #8692, opened 2023-07-13
fix schedule cron conversion in events

ackdav:
This PR fixes a cron syntax error. Currently if you schedule an events rule hourly or daily, it will execute every minute. This is due to cron's syntax, which will match every minute due to the `*`.
The following resource helps to understand cron syntax: [https://crontab.guru/every-1-hour](https://crontab.guru/every-1-hour)

Thanks to @whummer for the help in finding this.


dominikschubert (approved):
Nice catch :+1:

Issue #8694, opened 2023-07-13
Enhancement Request: Support S3 Inventory Report Generation inside Localstack


c0dzilla:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

We are trying to simulate AWS S3 Inventory Generation for a S3 bucket in Localstack. https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory.html

We want to create an Inventory Configuration that will deliver the Inventory Report for the specified bucket inside a destination bucket. We then want to use the generated Inventory Report to test our product.

For this, we first create an Inventory Configuration as follows:

Request:
```
awslocal --endpoint-url=http://localhost:4566 s3api put-bucket-inventory-configuration --bucket bucket1 --id test-inventory --inventory-configuration='{"Destination":{"S3BucketDestination":{"AccountId":"test","Bucket":"inv-bucket","Format":"CSV"}},"IsEnabled":true,"Id":"test-inventory","IncludedObjectVersions":"All","Schedule":{"Frequency":"Daily"}}'
```

This returns 200 Success response. We expect this to have created Inventory Configuration for bucket 'bucket1' inside destination bucket 'inv-bucket' (both buckets already exist).

We then list the Inventory Configurations for bucket 'bucket1' as follows:

Request:
```
awslocal --endpoint-url=http://localhost:4566 s3api list-bucket-inventory-configurations --bucket bucket1
```

However, we get an empty response:
```
{
    "IsTruncated": false
}
```

We then try to perform a GET for the Inventory Configuration with id 'test-inventory' that we created earlier for bucket 'bucket1' as follows:

Request:
```
awslocal --endpoint-url=http://localhost:4566 s3api get-bucket-inventory-configuration --bucket bucket1 --id test-inventory
```

This also returns an empty response:
```
{
    "InventoryConfiguration": {}
}
```

This indicates that the following is not currently supported inside Localstack:
- Creating an Inventory Configuration
- Localstack generating Inventory Report and delivering it to the specified destination bucket

Are there plans to add this functionality? It would be useful to products that wish to use Localstack for testing S3 Inventory Reports (such as backup providers). 

### Anything else?

https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory.html
https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketInventoryConfiguration.html

bentsku:
Hello @c0dzilla and thanks for your report.

[...]

PR #8696, opened 2023-07-13
implement S3 BucketInventoryConfiguration mocking

bentsku:
This PR partly implements #8694, at least the mocking part. We were actually not supporting any CRUD operations, as it was not implemented in moto. 

Also made use of the `context` in every `get_store` call, as we should remove any use of the `threadlocal` and make use of the nicely provided context. 

## TODO

- [x] implement Get/Put/List/Delete, happy path
- [x] implement validation and negative testing

## Added operations:
- `PutBucketInventoryConfiguration`
- `ListBucketInventoryConfigurations`
- `GetBucketInventoryConfiguration`
- `DeleteBucketInventoryConfiguration`

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import os
 from collections import defaultdict
+from operator import itemgetter
 from typing import IO, Dict, List, Optional
 from urllib.parse import parse_qs, quote, urlencode, urlparse, urlunparse
 from zoneinfo import ZoneInfo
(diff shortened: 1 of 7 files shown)


localstack/services/s3/provider.py
```diff
@@ -1587,6 +1599,79 @@ def select_object_content(
...
+        moto_backend = get_moto_s3_backend(context)
+        get_bucket_from_moto(moto_backend, bucket)
+
+        store = self.get_store(context)
+        bucket_inventory_configurations = store.bucket_inventory_configurations.get(bucket, {})
+
+        # bucket_inventory_configurations = sorted(
+        #     bucket_inventory_configurations.values(), key=lambda x: x["Id"]
+        # )
+        # TODO: verify the sorting?
```
macnev2013, line 1653: nit: `list_bucket_xxxx_configurations` are usually sorted by `Id` that is provided on configuration creation. So what you have commented here should be correct.
Similar sorting is being used in `list_bucket_analytics_configurations`. For this maybe we can have a snapshot test to verify.

macnev2013 (approved):

localstack/services/s3/provider.py
bentsku: Good catch! I'll add a snapshot test for this, forgot to do it... 😅

macnev2013:
Awesome changes. LGTM 🚀

PR #8697, opened 2023-07-13
Extend Docker client to allow passing --dns user flags

whummer:
Extend Docker client to allow passing `--dns` user flags, e.g., via `DOCKER_FLAGS` or `ECS_DOCKER_FLAGS`.

Summary of changes:
* Turn the existing `dns` references from a `str` into a `List[str]`, as this is a multi-value attribute. Should be backward-compatible, and type hints were adjusted to `[Union[str, List[str]]]`
* Extend the `Util.parse_additional_flags(..)` utility to support parsing of `--dns ...` arguments
* Add a test to cover the new functionality
* Slightly restructure the tests and group the `test_run_with_additional_arguments_*` tests into a `TestRunWithAdditionalArgs` class

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -1301,6 +1301,20 @@ class TestDockerClient:
         )
         assert "TEST_CONTENT" == file_path.read().strip()
 
+    def test_get_container_ip_non_existing_container(self, docker_client: ContainerClient):
+        with pytest.raises(NoSuchContainer):
+            docker_client.get_container_ip(f"hopefully_non_existent_container_{short_uid()}")
+
+    # TODO: getting container IP not yet working against Podman
+    @skip_for_podman
+    def test_get_container_ip(self, docker_client: ContainerClient, dummy_container):
+        docker_client.start_container(dummy_container.container_id)
+        ip = docker_client.get_container_ip(dummy_container.container_id)
+        assert is_ipv4_address(ip)
+        assert "127.0.0.1" != ip
+
+
+class TestRunWithAdditionalArgs:
     def test_run_with_additional_arguments(self, docker_client: ContainerClient):
         env_variable = "TEST_FLAG=test_str"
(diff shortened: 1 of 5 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -1330,17 +1344,19 @@ def test_run_with_additional_arguments_add_host(self, docker_client: ContainerCl
...
-
-    # TODO: getting container IP not yet working against Podman
-    @skip_for_podman
-    def test_get_container_ip(self, docker_client: ContainerClient, dummy_container):
-        docker_client.start_container(dummy_container.container_id)
-        ip = docker_client.get_container_ip(dummy_container.container_id)
-        assert is_ipv4_address(ip)
-        assert "127.0.0.1" != ip
+    def test_run_with_additional_arguments_add_dns(self, docker_client: ContainerClient):
+        additional_flags = "--dns 1.2.3.4 --dns 5.6.7.8"
```
dfangl, line 1348: Would be nice to test with one dns flag in run_container and one in additional_flags as well!

tests/unit/test_dockerclient.py
```diff
@@ -200,6 +203,15 @@ def test_user(self):
...
         assert flags.user == "nobody"
 
+    def test_dns(self):
+        argument_string = r"--dns 1.2.3.4"
+        flags = Util.parse_additional_flags(argument_string)
+        assert flags.dns == ["1.2.3.4"]
+
+        argument_string = r"--dns 1.2.3.4 --dns 5.6.7.8"
+        flags = Util.parse_additional_flags(argument_string)
+        assert flags.dns == ["1.2.3.4", "5.6.7.8"]
```
dfangl, line 213: Please also add a test to ensure that if no dns is set, the flags.dns is really empty. We had some problems with other flags for that in the past.

dfangl (approved):
All in all, LGTM! Just some minor test suggestions

PR #8702, opened 2023-07-14
fix issues with SSM Patchbaseline

pinzon:
Work for some errors not catched when deploying an SSM PatchBaseline and MaintenanceWindowTasks.


dominikschubert (approved):
LGTM, good job on fixing this :+1:

PR #8699, opened 2023-07-13
fix pod save after s3 multipart upload

bentsku:
This PR fixes #8072 

Basically, the issue stems from moto accessing thus creating a BackendDict with an `account_id` set to `None`.

Thanks to @taimourfenris, he discovered that it was after a `MultipartUpload`, the issue would arise. It happens also in lambda because lambda makes use of `upload_fileobj`, which will automatically switch to a multipart upload if the file is big.

I tried to debug moto to see when it would access a backend with an null account id.

I've created a small sample to create a multipart:
```python
# https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3.html#multipart-transfers
chunk_size = 5_242_880 * 10  # around 52MB
with io.BytesIO() as data:
    data.write(os.urandom(chunk_size * 2))
    data.seek(0)
    s3_client.upload_fileobj(data, "test-bucket", "test-key")
```
And I could confirm after running this command that `localstack pod save file://test.zip` would fail. 

Added a print statement to the `__get_item__` of `BackendDict` to see when it would be accessed:
```python
 File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/services/s3/provider.py", line 804, in upload_part
    response: UploadPartOutput = call_moto(context)
  File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/services/moto.py", line 47, in call_moto
    return dispatch_to_backend(context, dispatch_to_moto, include_response_metadata)
  File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/aws/forwarder.py", line 120, in dispatch_to_backend
    http_response = http_request_dispatcher(context)
  File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/services/moto.py", line 115, in dispatch_to_moto
    response = dispatch(request, request.url, request.headers)
  File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/utils/patch.py", line 38, in proxy
    return new(target, *args, **kwargs)
  File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/services/s3/provider.py", line 2043, in _fix_key_response
    status_code, resp_headers, key_value = fn(self, *args, **kwargs)
  File "/Users/benjaminsimon/Projects/localstack/moto/moto/utilities/aws_headers.py", line 74, in _wrapper
    response = f(*args, **kwargs)
  File "/Users/benjaminsimon/Projects/localstack/moto/moto/s3/responses.py", line 1208, in key_response
[...]

Issue #6659, opened 2022-08-12
bug: Checksum mismatch when putting object using AWS Java SDK


JurgenR:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When performing a putObject using Java AWS SDK with a providing only the checksum algorithm localstack returns a 400 InvalidRequest with error: `Value for x-amz-checksum-sha256 header is invalid.`

localstack logs:

```
2022-08-11T11:59:45.411 DEBUG --- [   asgi_gw_0] l.aws.serving.wsgi         : PUT localstack:4566/integration-bucket/testObjectKey_5f1gUa
[...]

PR #8704, opened 2023-07-15
fix decoding aws-chunked request for UploadPart

### LINKED ISSUE #8703: bug: The last version throws the "Expected hash not equal to calculated hash" error for uploading a multi part.

### Current Behavior

I am using the latest tag for the CI/CD process and today it started to fail because unit tests for the S3 provider that cover multipart upload throw the "Expected hash not equal to calculated hash" error on uploading a part. It worked just fine with the old version. The same unit tests work correctly with the real S3 provider as well.

And there is another issue. You fixed the https://github.com/localstack/localstack/issues/8392 bug that I reported previously but there is no version on Docker Hub that has that fix but does not have the new bug I described above. It seems that you rewrite the latest tag in Docker Hub which makes it impossible to rollback to the previous version.

### Expected Behavior

1. The upload part works without throwing an exception.
2. The previous versions of Docker images are available on Docker Hub.

With a docker-compose file

### Steps To Reproduce

1. Initiate multipart upload with the correct MD5 hash.
2. Upload a part for which the hash was calculated.

### Environment

```markdown
- OS: Windows 11
- LocalStack: latest
```

### Anything else?

_No response_

bentsku:
This PR fixes #8703 (I got confirmation with a code sample), this is a follow up from #8699

We had to set the `body` directly to moto to avoid a persistence bug, but moto has a step that we didn't execute: `_handle_v4_chunk_signatures`, which will decode the `aws-chunked` format. 

I've refactored the code used in the streaming provider that already handles this case, and made use of it in the case of `aws-chunked` requests for `UploadPart`, as we're skipping this moto step (actually one of the only one before actually calling the backend like we're doing in our provider). 

As a example, that is what the Java/.NET SDK would send if you try to upload a part with the following content:

Content: `Hello Blob`
Becomes:
```
a;chunk-signature=b5311ac60a88890e740a41e74f3d3b03179fd058b1e24bb3ab224042377c4ec9\r\n
Hello Blob\r\n
0;chunk-signature=78fae1c533e34dbaf2b83ad64ff02e4b64b7bc681ea76b6acf84acf1c48a83cb\r\n
```
Here's the doc on the `aws-chunked` format: 
https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

So in the previous PR, we would directly set the encoded body as is, and the `ETag` would not correspond anymore.

### DIFF
--- a/localstack/services/s3/constants.py
+++ b/localstack/services/s3/constants.py
@@ -66,3 +66,5 @@ SIGNATURE_V4_PARAMS = [
     "X-Amz-SignedHeaders",
     "X-Amz-Signature",
 ]
+
+S3_CHUNK_SIZE = 1024 * 16 * 4
--- a/localstack/services/s3/constants.py
+++ b/localstack/services/s3/constants.py
@@ -67,4 +67,6 @@ SIGNATURE_V4_PARAMS = [
     "X-Amz-Signature",
 ]
 
-S3_CHUNK_SIZE = 1024 * 16 * 4
+# The chunk size to use when iterating over and writing to S3 streams.
+# chosen as middle ground between memory usage and amount of iterations over the S3 object body
+S3_CHUNK_SIZE = 65536
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1,4 +1,5 @@
 import datetime
+import io
 import logging
 import os
 from collections import defaultdict
@@ -155,6 +156,7 @@ from localstack.services.s3.presigned_url import (
 from localstack.services.s3.utils import (
     _create_invalid_argument_exc,
     capitalize_header_name_from_snake_case,
+    decode_aws_chunked_object,
     extract_bucket_key_version_id_from_copy_source,
     get_bucket_from_moto,
     get_header_name,
@@ -802,7 +804,28 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                 ArgumentValue=part_number,
             )
 
-        part = body.read() if (body := request.get("Body")) else b""
+        body = request.get("Body")
+        headers = context.request.headers
+        if body and (
+            (content_sha_256 := headers.get("x-amz-content-sha256") or "")
+            and content_sha_256.startswith("STREAMING-")
+        ):
+            # this is a chunked request, we need to properly decode it while setting the key value
+            decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
+            checksum_algorithm = request.get("ChecksumAlgorithm") or ""
+            buffer = io.BytesIO()
+            # this will decode the original stream and set it in `part`
+            decode_aws_chunked_object(
+                stream=body,
+                buffer=buffer,
+                content_length=decoded_content_length,
+                checksum_algorithm=checksum_algorithm,
+                checksum_value=request.get(f"Checksum{checksum_algorithm.upper()}"),
+            )
+            buffer.seek(0)
+            part = buffer.getvalue()
+        else:
+            part = body.read() if body else b""
 
         # we are directly using moto backend and not calling moto because to get the response, moto calls
         # key.response_dict, which in turns tries to access the tags of part, indirectly creating a BackendDict
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -806,23 +806,18 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         body = request.get("Body")
         headers = context.request.headers
-        if body and (
-            (content_sha_256 := headers.get("x-amz-content-sha256") or "")
-            and content_sha_256.startswith("STREAMING-")
-        ):
+        if body and "aws-chunked" in (headers.get("Content-Encoding") or "").lower():
             # this is a chunked request, we need to properly decode it while setting the key value
             decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
-            checksum_algorithm = request.get("ChecksumAlgorithm") or ""
             buffer = io.BytesIO()
-            # this will decode the original stream and set it in `part`
+            # this will decode the original stream into `buffer`
             decode_aws_chunked_object(
                 stream=body,
                 buffer=buffer,
                 content_length=decoded_content_length,
-                checksum_algorithm=checksum_algorithm,
-                checksum_value=request.get(f"Checksum{checksum_algorithm.upper()}"),
             )
             buffer.seek(0)
+            # read the buffer value now that's decoded
             part = buffer.getvalue()
         else:
             part = body.read() if body else b""
@@ -832,9 +827,6 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         # with an account_id set to None (because moto does not set an account_id to the FakeKey representing a Part)
         key = moto_backend.upload_part(bucket_name, upload_id, part_number, part)
         response = UploadPartOutput(ETag=key.etag)
-
-        if key.checksum_algorithm is not None:
-            response[f"Checksum{key.checksum_algorithm.upper()}"] = key.checksum_value
 
         if key.encryption is not None:
             response["ServerSideEncryption"] = key.encryption
--- a/localstack/services/s3/provider_stream.py
+++ b/localstack/services/s3/provider_stream.py
@@ -42,7 +42,7 @@ from localstack.services.moto import call_moto, call_moto_with_request
 from localstack.services.s3.models import get_moto_s3_backend
 from localstack.services.s3.provider import S3Provider
 from localstack.services.s3.utils import (
-    InvalidRequest,
+    decode_aws_chunked_object,
     extract_bucket_key_version_id_from_copy_source,
     get_bucket_from_moto,
     get_key_from_moto_bucket,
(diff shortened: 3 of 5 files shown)


localstack/services/s3/constants.py
```diff
@@ -66,3 +66,5 @@
     "X-Amz-SignedHeaders",
     "X-Amz-Signature",
 ]
+
+S3_CHUNK_SIZE = 1024 * 16 * 4
```
alexrashed, line 70: What exactly is this constant here? It's 64k, but why is that? Could you please add a comment?

localstack/services/s3/utils.py
```diff
@@ -151,6 +156,92 @@ def verify_checksum(checksum_algorithm: str, data: bytes, request: Dict):
...
+) -> tuple[str, str]:
+    """
+
+    :param stream: the original stream to read
+    :param buffer: the stream where we set the decoded data
+    :param content_length: the total length of the original stream
+    :param checksum_algorithm:
+    :param checksum_value:
+    :return:
+    """
```
alexrashed, line 174: Would be great to have a better docstring here. How does the return tuple look like? What's the algo and value params?

localstack/services/s3/utils.py
```diff
@@ -151,6 +156,92 @@ def verify_checksum(checksum_algorithm: str, data: bytes, request: Dict):
...
+            trailing_headers.append(next_line.strip())
+
+        # try for trailing headers after
+        while line := stream.readline():
+            trailing_header = line.strip()
+            if trailing_header:
+                trailing_headers.append(trailing_header)
+
+    # look for the checksum header in the trailing headers
+    # TODO: we could get the header key from x-amz-trailer as well
```
alexrashed, line 225: I think it would make sense to read the `x-amz-trailer` header, because the hard-coded `x-amz-checksum-*` seems a bit dangerous.

localstack/services/s3/utils.py
```diff
@@ -151,6 +156,92 @@ def verify_checksum(checksum_algorithm: str, data: bytes, request: Dict):
...
+    for trailing_header in trailing_headers:
+        try:
+            header_key, header_value = trailing_header.decode("utf-8").split(":", maxsplit=1)
+            if header_key.lower() == f"x-amz-checksum-{checksum_algorithm}".lower():
+                checksum_value = header_value
+        except (IndexError, ValueError, AttributeError):
+            continue
```
alexrashed, line 232: What are the typical cases for these errors being raised? Are they really all safe to be ignored here without a single log?

localstack/services/s3/provider.py
```diff
@@ -802,7 +804,28 @@ def upload_part(self, context: RequestContext, request: UploadPartRequest) -> Up
                 ArgumentValue=part_number,
             )
 
-        part = body.read() if (body := request.get("Body")) else b""
+        body = request.get("Body")
+        headers = context.request.headers
+        if body and (
+            (content_sha_256 := headers.get("x-amz-content-sha256") or "")
```
alexrashed, line 810: According to the docs you linked, the AWS Chunked Encoding also has to be declared in the `Content-Encoding` header. This might be a cleaner way to define that the payload is encoded with AWS' proprietary chunk encoding.

localstack/services/s3/provider_stream.py
```diff
@@ -125,20 +126,17 @@ def put_object(
...
         headers = context.request.headers
-        content_sha_256 = context.request.headers.get("x-amz-content-sha256") or ""
-        try:
-            if content_sha_256.startswith("STREAMING-"):
-                # this is a chunked request, we need to properly decode it while setting the key value
-                decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
-                key_object.set_value_from_chunked_payload(body, decoded_content_length)
+        if (
+            content_sha_256 := headers.get("x-amz-content-sha256") or ""
+        ) and content_sha_256.startswith("STREAMING-"):
```
alexrashed, line 131: Same as above, I think the distinction should be done based on the content encoding header if possible.

localstack/services/s3/provider.py
```diff
@@ -802,7 +804,28 @@ def upload_part(self, context: RequestContext, request: UploadPartRequest) -> Up
                 ArgumentValue=part_number,
             )
 
-        part = body.read() if (body := request.get("Body")) else b""
+        body = request.get("Body")
```
alexrashed, line 807: It feels like this decoding could actually be done automatically, similar to the GZIP decoding, right in the handler chain in a wrapping generator.
For the GZIP handling we explicitly exclude S3 because it actually needs the raw data because it directly stores it and (against the HTTP spec) the checksums are calculated for the gzip encoded content instead of the raw content.

However, here it seems that S3 always decodes the body for all operations, where it would make sense to have an explicit decoder in the handler chain. What do you think?

alexrashed (commented):
Thanks for jumping on this! Yet another AWS/S3 specific HTTP feature which causes some issues. 😛 
The PR is looking good, I only have some minor nitpicks and one more fundamental question (if this should be done transparently in the handler chain).
Afterwards we are good to go! 🚀

localstack/services/s3/provider.py
bentsku: I see what you mean, in the moto provider it makes sense because we read the whole payload in memory to then set it directly. However, in the new provider, it means we would iterate over the stream twice, once to decode it, then once to set it. Right now, in the stream provider, we're both decoding and validating checksum/etag at the same time, so we iterate over it only once.

I see the value of having the handler chain doing it, especially with the current implementation (it could parse the trailing headers and set it in the requests, etc), but it would also mean more work I think?

localstack/services/s3/utils.py
bentsku: I can log them, our current support for `aws-chunk` and checksum validation is experimental only with the stream provider, and with the lack of testing support for it, I didn't want to implement a new feature at the same time as the bug fix so ignored them for now. Currently in moto `UploadPart` does not verify any checksum.

localstack/services/s3/provider.py
bentsku: The separation of concern between the decoding and the checksum validation would be great as well. I really like this idea, it's too bad moto won't let use implement it, I might be able to patch it though, and remove their decoding.

localstack/services/s3/constants.py
bentsku: I'll add a comment, this was the result of quick tests with the S3 streaming to see what kind of value would have impact on the memory usage vs time spent iterating over the stream. Not very empirically chosen I have to say.

localstack/services/s3/provider.py
bentsku: I've just found the original issue we've had, and this was the sent request:

```
PUT /bucket1/data HTTP/1.1
Host: 127.0.0.1:14572
User-Agent: MinIO (linux; amd64) minio-go/v6.0.35
Content-Length: 185
Authorization: AWS4-HMAC-SHA256 Credential=foo/20190918/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length,Signature=9ff757b67570b0d555fed0e4944ce012ebcb3f963b40044296f3fb89cdb739d8
Content-Type: audio/mpeg
X-Amz-Content-Sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
X-Amz-Date: 20190918T051509Z
X-Amz-Decoded-Content-Length: 13

d;chunk-signature=af5e6c0a698b0192e9aa5d9083553d4d241d81f69ec62b184d05c509ad5166af
Hello  



0;chunk-signature=f2a50a8c0ad4d212b579c2489c6d122db88d8a0d0b987ea1f3e9d081074a5937
```

I'm worried the `Content-Encoding` header is not always respected...

localstack/services/s3/provider.py
alexrashed: If it would be problematic with the current moto integration, it is perfectly fine to move this to the content decoder in the handler chain in another iteration :)

bentsku: Perfect, but thanks a lot for the idea, it makes a lot of sense to separate it. Just need to find the right implementation now 😄

localstack/services/s3/utils.py
alexrashed: It's okay if these exceptions are all explicitly ignored, and it's documented why they are ignored.
So you can either add a debug log (if that makes sense), or add an informative comment on why you are explicitly ignoring these three errors and why it is okay to do so.

localstack/services/s3/provider.py
alexrashed: And this request would work against AWS? It seems like a test-request coming from a minio client? The spec quite explicitly says that you have to set the content encoding: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

bentsku: I'm not sure, because I would need to properly sign it and I can't do it with boto. I'll need a test setup with Java or .NET and manually change the headers to see if AWS will reject it. I guess it should.
From another issue (https://github.com/localstack/localstack/issues/6659), the `Content-Encoding` is properly set from the Java client. I think we can follow the spec like you indicated, and if issues arise because AWS clients do not follow the specs then we will fix it.

bentsku: Found the [minio implementation](https://github.com/minio/minio-go/blob/b75cfd43c196482756ff95460c779793e18d3ce4/pkg/signer/request-signature-v4.go#L292) now, it seems they are indeed setting the header, so we should be good, it was an old issue.

localstack/services/s3/utils.py
bentsku: I'll keep this for the streaming provider/new decoder! removed now from this PR

PR #8705, opened 2023-07-15
fix api gateway domains not being saved between versions

calvernaz:
Custom domains created on any version of API Gateway should be visible across APIs. For example, using aws cli with apigatewayv2 client to create a custom domain and query the same domain using aws cli with apigateway client after, should return the previous domain created.

## Changes

- Add a new field to the ApiGw model to hold the created domains.
- Implements the `create_domain_name,` `get_domain_name` and `get_domain_names` API calls to fetch the domains from the model. 
- Move away from moto store and use localstack apigw store to store created domains
- Almost all operations but update domains act on both APIs, for example deleting a domain happen in both stores, same as create and fetch.

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -1,5 +1,6 @@
 import contextlib
 import copy
+import hashlib
 import json
 import logging
 import re
@@ -47,7 +48,7 @@ from localstack.utils.aws import resources as resource_utils
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.aws.aws_responses import requests_error_response_json, requests_response
 from localstack.utils.aws.request_context import MARKER_APIGW_REQUEST_REGION, THREAD_LOCAL
-from localstack.utils.strings import long_uid, short_uid, to_str
+from localstack.utils.strings import long_uid, short_uid, to_bytes, to_str
 from localstack.utils.time import TIMESTAMP_FORMAT_TZ, timestamp
 from localstack.utils.urls import localstack_host
 
@@ -1554,3 +1555,21 @@ def log_template(
         response_headers=response_headers,
         status_code=status_code,
     )
+
+
+def get_domain_name_hash(domain_name: str) -> str:
+    """
+    Return a hash of the given domain name, which help construct regional domain names for APIs.
+    TODO: use this in the future to dispatch API Gateway API invocations made to the regional domain name
+    """
+    return hashlib.shake_128(to_bytes(domain_name)).hexdigest(4)
+
+
+def get_regional_domain_name(domain_name: str) -> str:
+    """
+    Return the regional domain name for the given domain name.
+    In real AWS, this would look something like: "d-oplm2qchq0.execute-api.us-east-1.amazonaws.com"
+    In LocalStack, we're returning this format: "d-<domain_hash>.execute-api.localhost.localstack.cloud"
+    """
+    domain_name_hash = get_domain_name_hash(domain_name)
+    return f"d-{domain_name_hash}.execute-api.localhost.localstack.cloud"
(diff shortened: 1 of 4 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -293,6 +301,74 @@ def put_rest_api(self, context: RequestContext, request: PutRestApiRequest) -> R
...
+        store: ApiGatewayStore = get_apigateway_store(context.account_id, context.region)
+        if store.domain_names.get(domain_name):
+            raise ConflictException(f"Domain name with ID {domain_name} already exists")
+
+        # find matching hosted zone
+        zone_id = None
+        route53 = connect_to().route53
+        hosted_zones = route53.list_hosted_zones().get("HostedZones", [])
+        hosted_zones = [hz for hz in hosted_zones if domain_name.endswith(hz["Name"].strip("."))]
+        zone_id = hosted_zones[0]["Id"].replace("/hostedzone/", "") if hosted_zones else zone_id
```
bentsku, line 334: question: I'm not too familiar with route53 and the integration with the domain names. Will it always match to a `HostedZone`? What if it doesn't?

bentsku (commented):
LGTM! That looks great 😄 I suppose this is impossible to test against AWS, right? Only one question related to `route53` as I don't know it that much, but otherwise good to go!

localstack/services/apigateway/models.py
```diff
@@ -68,6 +69,8 @@ class ApiGatewayStore(BaseStore):
     # maps resource ARN to tags
     TAGS: Dict[str, Dict[str, str]] = CrossRegionAttribute(default=dict)
 
+    domain_names: Dict[str, DomainName] = LocalAttribute(default=dict)
```
whummer, line 72: tiny nit: can we maybe group it visually to move the `LocalAttribute`s together, and have the `CrossRegionAttribute` below it? Would also be great to add a short comment (or docstring) to the variable, to be in line with the other fields in this class 👍

localstack/services/apigateway/helpers.py
```diff
@@ -1554,3 +1555,21 @@ def log_template(
...
+
+
+def get_regional_domain_name(domain_name: str) -> str:
+    """
+    Return the regional domain name for the given domain name.
+    In real AWS, this would look something like: "d-oplm2qchq0.execute-api.us-east-1.amazonaws.com"
+    In LocalStack, we're returning this format: "d-<domain_hash>.execute-api.localhost.localstack.cloud"
+    """
+    domain_name_hash = get_domain_name_hash(domain_name)
+    return f"d-{domain_name_hash}.execute-api.localhost.localstack.cloud"
```
whummer, line 1575: nit: would be good to use `constants.LOCALHOST_HOSTNAME` here
```suggestion
    return f"d-{domain_name_hash}.execute-api.{LOCALHOST_HOSTNAME}"
```

whummer (approved):
LGTM @calvernaz ! 👍 Just a small nitpick. In terms of testing, we could craft up a snapshot test, but Route53 may need some manual DNS verification steps. Happy to discuss this offline for a follow-up..

calvernaz:
> LGTM @calvernaz ! +1 Just a small nitpick. In terms of testing, we could craft up a snapshot test, but Route53 may need some manual DNS verification steps. Happy to discuss this offline for a follow-up..

Certificate issuing has a validation step, until is available, it's possible, but I rather do it in a follow-up, just with a snapshot test.

localstack/services/apigateway/provider.py
calvernaz: Not sure I understand the question, maybe something to explore offline. A zone can group many domains. But looking at this code, which was already present in other places I think, it could be done using `list-hosted-zones-by-name`. 

Let me look into it

PR #8711, opened 2023-07-17
Remove usage of GenericBaseModel outside of LegacyResourceProvider

dominikschubert:
## Changes

- Introduced `CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES` config variable to determine if we should fail if a resource type doesn't exist. E.g. setting `CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES=0` will cause a stack deployment to fail if the resource type doesn't have an associated provider.
- Removed any access to `RESOURCE_MODELS`, i.e. the GenericBaseModel classes in the template_deployer, *except* the current fallback for retrieving `Fn::GetAtt` targets that still needs to use them. If we refactor the models to remove all `get_cfn_attribute` calls, we can delete this completely.
- Removed `retrieve_resource_details`
- Removed `extract_resource_attribute`
- Removed a lot of unnecessary code from `resolve_ref`
- Removed `get_resource_model_instance`
- Removed `invoke_function` (unused)
- Removed `get_ref_from_model` => simple physical resource ID lookup now
- Removed `determine_resource_physical_id` => simple physical resource ID lookup now
- Removed `is_deployable_resource` and `is_updateable` => everything should be a "deployable resource" and `is_updateable` is now only used in the `LegacyResourceProvider`
- Removed `update_resource_details` => this was a bit of a headache unfortunately since it was responsible for populating the `_state_`, i.e. the result of the fetch_state call in the models. This initially broke UPDATEs on the old models. I've introduced a workaround for Add and Modify operations in the LegacyResourceProvider. It's a bit hacky but we also shouldn't let those currently 15 resources with (partial) UPDATE support live for too long as GenericBaseModels anyway.
- Removed `GenericBaseModel.get_ref`
- Removed `GenericBaseModel.fetch_state_if_missing`

### DIFF
--- a/localstack/services/cloudformation/resource_provider.py
+++ b/localstack/services/cloudformation/resource_provider.py
@@ -461,6 +461,12 @@ class LegacyResourceProvider(ResourceProvider):
         resource_provider.add_defaults(
             self.all_resources[request.logical_resource_id], request.stack_name
         )
+        # for some reason add_defaults doesn't even change the values in the resource provider...
+        # incredibly hacky again but should take care of the defaults
+        resource_provider.resource_json["Properties"] = self.all_resources[
+            request.logical_resource_id
+        ]["Properties"]
+        resource_provider.properties = self.all_resources[request.logical_resource_id]["Properties"]
 
         func_details = resource_provider.get_deploy_templates()
         # TODO: be less strict about the return value of func_details
--- a/localstack/services/cloudformation/resource_provider.py
+++ b/localstack/services/cloudformation/resource_provider.py
@@ -43,7 +43,7 @@ PUBLIC_REGISTRY: dict[str, Type[ResourceProvider]] = {}
 # by default we use the GenericBaseModel (the legacy model), unless the resource is listed below
 # add your new provider here when you want it to be the default
 PROVIDER_DEFAULTS = {
-    # "AWS::IAM::User": "GenericBaseModel",
+    "AWS::IAM::User": "ResourceProvider",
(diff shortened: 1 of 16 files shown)


tests/integration/cloudformation/api/test_changesets.py
```diff
@@ -328,7 +328,7 @@ def test_describe_change_set_nonexisting(snapshot, aws_client):
     snapshot.match("exception", ex.value)
 
 
-@pytest.mark.xfail(reason="fails because of the properties mutation in the result_handler")
+@pytest.mark.skip(reason="fails because of the properties mutation in the result_handler")
```
dominikschubert, line 331: don't see a point in waiting 5min for all the timeouts here, so this should speed up the tests a bit :sweat_smile:

localstack/services/cloudformation/resource_provider.py
```diff
@@ -597,43 +627,61 @@ def deploy_loop(
...
+
+            except NoResourceProvider:
+                log_not_available_message(
+                    raw_payload["resourceType"],
+                    f"No resource provider found for \"{raw_payload['resourceType']}\"",
+                )
+
+                if config.CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES:
+                    # TODO: figure out a better way to handle non-implemented here?
+                    return ProgressEvent(OperationStatus.SUCCESS, resource_model={})
```
dominikschubert, line 681: not sure about the returned model here

pinzon (approved):
Awesome!!!

PR #8712, opened 2023-07-17
fix detection of S3 aws-chunked encoding

### LINKED ISSUE #8703: bug: The last version throws the "Expected hash not equal to calculated hash" error for uploading a multi part.

bentsku:
Follow up from #8704, the detection of `aws-chunked` encoding was changed to follow the S3 specs, it appears however that the SDK are not following it. Reverse back to the previous detection using `x-amz-content-sha256`. 

Related comment: https://github.com/localstack/localstack/pull/8704#discussion_r1265118283

Fixes #8703

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -813,7 +813,10 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         body = request.get("Body")
         headers = context.request.headers
-        if body and "aws-chunked" in (headers.get("Content-Encoding") or "").lower():
+        # AWS specifies that the `Content-Encoding` should be `aws-chunked`, but some SDK don't set it.
+        # Rely on the `x-amz-content-sha256` which is a more reliable indicator that the request is streamed
+        content_sha_256 = (headers.get("x-amz-content-sha256") or "").upper()
+        if body and content_sha_256 and content_sha_256.startswith("STREAMING-"):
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
LGTM! 🚀 
It's a pity that so many client's are not compliant with the spec 😛

bentsku:
> LGTM! 🚀
> It's a pity that so many client's are not compliant with the spec 😛

We are starting to get used to it, aren't we? 😄

PR #8713, opened 2023-07-17
move legacy S3 provider into legacy directory

bentsku:
Doing some clean up, moving the old legacy S3 provider into its own legacy directory like SQS, so we know what depends on it and can start moving it around.


tests/unit/aws/protocol/test_parser.py
```diff
@@ -17,7 +17,7 @@
 )
 from localstack.aws.spec import load_service
 from localstack.http import Request as HttpRequest
-from localstack.services.s3 import s3_utils
+from localstack.services.s3.legacy import s3_utils
```
alexrashed, line 20: Maybe we could - instead of importing the legacy module here - just copy the code here to avoid the import.
This import is only used for the `get_bucket_hostname`, which in turn is only used in this test. It also is just one single, trivial line:
```
"%s.%s:%s" % (bucket_name, S3_VIRTUAL_HOSTNAME, config.EDGE_PORT)
```

I think it would be awesome if we could get rid of this import now instead of stumbling across it when we remove the legacy module.
This also is - as far as I can see - the only legacy import in a non-legacy codepath. This means that we could just simply remove the legacy module, the providers, and some legacy codepaths (like in `generic_proxy` or the cloudformation `api_utils`) later on without any complications. 🚀

localstack/services/s3/utils.py
alexrashed, line 1: To summarize the changes in this file:
- `uses_host_addressing` and `normalize_bucket_name`:
  - These functions have been moved from the (now legacy) `s3_utils`.
- `extract_bucket_key_version_id_from_s3_url`:
  - This is a generalized version of the previous `extract_bucket_key_version_id_from_copy_source`.
  - Why was this necessary in a PR that should just move code?
  - There is a TODO for missing unit tests. Maybe there's a chance in writing one or two?
- `extract_bucket_name_and_key_from_headers_and_path`:
  - Where does this function come from and why has it been created in a PR which "just moves things"?

alexrashed (changes requested):
This is a great change! It prepares our upcoming legacy removal really well and also makes sure that we don't use any legacy code in new implementations (because they are easily identifiable by the module path). 🚀 🧹 
I only have two comments on the new utils functions and a remaining legacy import in one of the tests. After that, we are good to go! 🔝

tests/unit/aws/protocol/test_parser.py
bentsku: This test only validates the behaviour of the legacy provider, so it should be removed as well when we delete the legacy module, that's why I explicitly left the import so that we could find it.
See the marker:
```python
@pytest.mark.skipif(
    not config.LEGACY_S3_PROVIDER, reason="ASF provider does not rely on virtual host parser"
)
```

localstack/services/s3/utils.py
bentsku: Yes:
- `uses_host_addressing` has been rewritten to be a bit more efficient, and `normalize_bucket_name` simply moved.
- `extract_bucket_key_version_id_from_s3_url` I'm actually hesitating and might revert the change, I've spotted a utility Sannya added that actually does the second part I've added, so we could keep `extract_bucket_key_version_id_from_copy_source` + the other util. I'll revert.
- `extract_bucket_name_and_key_from_headers_and_path` is used in CloudFormation, this is the consolidation of 2 methods that were used together in `localstack/services/cloudformation/api_utils.py` but made sense to be merged together as they were doing the same assertions twice (uses virtual host, etc). It consolidates `s3_utils.extract_bucket_name` and `s3_utils.extract_key_name`. It is also used in -ext, in S3 select and CloudTrail.

PR #8714, opened 2023-07-17
Pin Cython<3 to fix PyYaml build

dfangl:
## Motivation
Since the release of Cython 3.0.0 an hour ago: https://pypi.org/project/Cython/#history , the installation of our dependency PyYaml broke.

See https://github.com/yaml/pyyaml/issues/601

## Changes
* Pin cython<3 for now, to ensure pyyaml builds correctly. We should remove the pin once the incompatibility is fixed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,6 +33,8 @@ install_requires =
     psutil>=5.4.8,<6.0.0
     python-dotenv>=0.19.1
     pyyaml>=5.1
+    # remove once https://github.com/yaml/pyyaml/issues/601 is fixed
+    Cython<3
     rich>=12.3.0
     requests>=2.20.0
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks for jumping on this so fast to unblock the build! 🚀

makimkin:
It doesn't seem to be working 😥

dfangl:
You are right, it is not. I can reproduce this locally, and I am currently trying to fix it.

dfangl:
We will wait until a better solution is available. If we pin an earlier version, any install of awscli will automatically upgrade it again. Will wait for a better solution before disabling the build isolation before every build, as this will still influence the CLI experience heavily.

PR #8717, opened 2023-07-18
Remove get_cfn_attribute from legacy models

dominikschubert:
Removes all usage of `get_cfn_attribute` from the community resource models. 

It does not yet introduce a subset selection for the migrated resources to stay in parity with the previous behavior for now where we nearly always fall back to a direct property/state lookup.

A follow-up PR will soon remove the base method from the GenericBaseModel and unify the code paths between old and legacy providers.

### DIFF
--- a/localstack/services/cloudformation/models/cloudformation.py
+++ b/localstack/services/cloudformation/models/cloudformation.py
@@ -15,12 +15,12 @@ class CloudFormationStack(GenericBaseModel):
     def cloudformation_type():
         return "AWS::CloudFormation::Stack"
 
-    def fetch_state(self, stack_name, resources):
-        client = connect_to().cloudformation
-        child_stack_name = self.props["StackName"]
-        result = client.describe_stacks(StackName=child_stack_name)
-        result = (result.get("Stacks") or [None])[0]
-        return result
+    # def fetch_state(self, stack_name, resources):
+    #     client = connect_to().cloudformation
+    #     child_stack_name = self.props["StackName"]
+    #     result = client.describe_stacks(StackName=child_stack_name)
+    #     result = (result.get("Stacks") or [None])[0]
+    #     return result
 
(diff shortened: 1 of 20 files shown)


localstack/services/cloudformation/service_models.py
```diff
@@ -84,7 +84,16 @@ def add_defaults(resource, stack_name: str):
...
+        prop_candidate = self.props.get(attribute_name)
+        if prop_candidate:
+            return prop_candidate
+
+        if "." in attribute_name:
+            parts = attribute_name.split(".")
+            attribute = self.props
+            for part in parts:
+                attribute = attribute.get(part)
+            return attribute
```
dominikschubert, line 96: just temporary, since this whole method will soon be removed. this will enable resolving properties like `Outputs.NestedStackOutputA`

localstack/services/cloudformation/models/cloudformation.py
```diff
@@ -15,32 +15,12 @@ class CloudFormationStack(GenericBaseModel):
...
-            else:
-                raise Exception(f"Too many output values found for key {output_key=}")
-
-        return super(CloudFormationStack, self).get_cfn_attribute(attribute_name)
+    # def fetch_state(self, stack_name, resources):
+    #     client = connect_to().cloudformation
+    #     child_stack_name = self.props["StackName"]
+    #     result = client.describe_stacks(StackName=child_stack_name)
+    #     result = (result.get("Stacks") or [None])[0]
+    #     return result
```
dominikschubert, line 23: commented out for now, I'll probably churn through the models and check where we can delete fetch_state implementations after the two PRs are merged :smile:

simonrw (approved):
Looks good, glad to see we can get rid of the custom `get_cfn_attribute` handling and move towards the resource provider way of doing things

PR #8718, opened 2023-07-18
Initial Scenario testing and provisioning prototype

dominikschubert:
## Changes

- Introduced new `InfraProvisioner` to our testing library which can be used for more complex test setups such as CDK-based infrastructure definitions.
  - Set up tests directly via CDK (currently limited to single stacks, so you'll need to make sure the CDK app doesn't need any assets)
  - Manual setup and teardown can be registered if required
  - By default any S3 buckets created by CDK are cleaned before tearing down the stack again
  - Teardown can be skipped, to re-use the provisioned infrastructure between test runs. This is especially handy while writing test assertions and iterating over them. If you had to create/remove the stacks all the time, certain resources would make this impractical (e.g. clusters)
- Added initial scenario tests (most of them are still a kind of playground right now)


steffyP (approved):
Looking forward to our new scenario tests in practise 😄

PR #8719, opened 2023-07-18
Refs are updated on resource deployment completion

simonrw:
Update refs once a resource has finished deploying. This should reduce the number of calls to `resolve_refs_recursively`.


pinzon (approved):
LGTM

PR #8721, opened 2023-07-18
S3 AWS-chunked stream decoder

bentsku:
This PR implements a stream decoder which will decode on the fly the `aws-chunked` encoding, so that we only have one way to set the stream to the S3 objects value, which simplifies the code a lot, even for the new provider.

One of the main issue around this decoder is that we're getting data only after we read the whole stream, as the necessary headers for checksum calculation are appended at the end of the response.

We then need a way to access those headers for our logic to go forward. One way for this to be transparent is to pass the s3 object to the stream, so that once it has finish decoding, it will automatically set the trailing checksum to the object attribute.

Maybe one idea could also be to be able to register callback, but as it's always the same, I thought we don't need to make it extendable for now.

### DIFF
--- a/localstack/services/s3/codec.py
+++ b/localstack/services/s3/codec.py
@@ -96,4 +96,4 @@ class AwsChunkedDecoder(io.RawIOBase):
         while line := self._stream.readline():
             if trailing_header := line.strip():
                 header_key, header_value = trailing_header.decode("utf-8").split(":", maxsplit=1)
-                self._trailing_headers[header_key] = header_value.strip()
+                self._trailing_headers[header_key.lower()] = header_value.strip().lower()
--- a/localstack/services/s3/codec.py
+++ b/localstack/services/s3/codec.py
@@ -96,4 +96,4 @@ class AwsChunkedDecoder(io.RawIOBase):
         while line := self._stream.readline():
             if trailing_header := line.strip():
                 header_key, header_value = trailing_header.decode("utf-8").split(":", maxsplit=1)
-                self._trailing_headers[header_key.lower()] = header_value.strip().lower()
+                self._trailing_headers[header_key.lower()] = header_value.strip()
--- a/localstack/services/s3/codec.py
+++ b/localstack/services/s3/codec.py
@@ -1,0 +1,99 @@
+import io
+from typing import IO
+
+
+class AwsChunkedDecoder(io.RawIOBase):
+    """
+    This helper class takes a IO[bytes] stream, and decodes it on the fly, so that S3 can directly access the stream
+    without worrying about implementation details of `aws-chunked`.
+    It does need access to the trailing headers, which are going to be available once the stream is fully read.
+    See `aws-chunked` format here: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
+    """
+
+    def readable(self):
+        return True
+
+    def __init__(self, stream: IO[bytes], decoded_content_length: int):
+        self._stream = stream
+
+        self._decoded_length = decoded_content_length  # Length of the encoded object
+        self._new_chunk = True
+        self._end_chunk = False
+        self._chunk_size = 0
+        self._trailing_headers = {}
+
+    @property
+    def trailing_headers(self):
+        return self._trailing_headers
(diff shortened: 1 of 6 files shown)


localstack/services/s3/provider_stream.py
```diff
@@ -131,11 +132,21 @@ def put_object(
...
             decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
-            key_object.set_value_from_chunked_payload(body, decoded_content_length)
+            # we're passing the S3 Object, the decoder will set its `ChecksumValue` once it has finished reading
+            # the stream
+            body = AwsChunkedDecoder(body, decoded_content_length, key_object)
 
-        else:
-            # set the stream to be the value of the key
+        try:
             key_object.value = body
```
alexrashed, line 140: q: Hmm, why would setting the value here raise an exception?

localstack/services/s3/codec.py
```diff
@@ -0,0 +1,119 @@
...
+        self._decoded_length = decoded_content_length  # Length of the encoded object
+        self._new_chunk = True
+        self._end_chunk = False
+        self._chunk_size = 0
+        self._trailing_headers = {}
+        self.s3_object = s3_object
+
+    @property
+    def trailing_headers(self):
+        return self._trailing_headers
```
alexrashed, line 32: nit: You mentioned the callback above, but you could make it more explicit if you would raise an `ValueError` (or maybe there's something that's even more fitting) if the trailing headers are accessed even though they have not been read yet.
This would make it quite explicit and issues with missing trailing headers due to the stream not being fully read could be avoided.

alexrashed (approved):
This is just a great solution to a weird codec of AWS S3 😄
I only have some questions and a few nitpicks, nothing blocking. :shipit:

localstack/services/s3/codec.py
```diff
@@ -0,0 +1,119 @@
...
+from typing import IO, Any, Optional
+
+
+class AwsChunkedDecoder(io.RawIOBase):
+    """
+    This helper class takes a IO[bytes] stream, and decodes it on the fly, so that S3 can directly access the stream
+    without worrying about implementation details of `aws-chunked`.
+    It needs to expose the trailing headers, which will be available once the stream is fully read.
+    You can also directly pass the S3 Object, so the stream would set the checksum value once it's done.
+    TODO: should we register callback instead of passing the object?
```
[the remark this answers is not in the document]
bentsku: Will do, thanks! 👍

localstack/services/s3/codec.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def readable(self):
+        return True
+
+    def __init__(
+        self, stream: IO[bytes], decoded_content_length: int, s3_object: Optional[Any] = None
+    ):
+        self._stream = stream
+
+        self._decoded_length = decoded_content_length  # Length of the encoded object
```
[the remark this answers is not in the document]
bentsku: Yeah, we could use some heuristics to determine when the body is over (not having the content length encoded in the chunk), however AWS requires this header and it makes sense to use it to be sure we're properly retrieving the body... but agree it would be nice.

localstack/services/s3/provider_stream.py
bentsku: That's a good question 😄 basically, setting the value will trigger a property setter in moto, where it will verify the checksum. So it can trigger exception... in the new provider, I've decoupled the checksum logic, so there's a proper check after setting the value, but we're not there yet with this one.

localstack/services/s3/codec.py
bentsku: Agreed, that would be nice! I'm not directly accessing those trailing headers anymore, the previous commits were directly accessing them but it made a different code path between regular stream and aws-chunked one. But having this contract properly set would make things a bit more transparent I agree. I'm going to add this, and look for an exception to raise. Will change that 👍

localstack/services/s3/provider_stream.py
alexrashed: Okay, then maybe add a small comment explaining just that :)

PR #8722, opened 2023-07-19
Fix regression for Fn::Sub in condition section

dominikschubert:
Allows for `Fn::Sub` intrinsic function evaluation in the resolving of the conditions. Still very explicit, but this area of code will soon be reworked into a sort of visitor pattern allowing us to structure this a bit better.

This regression was introduced recently when splitting the resolve path for Conditions from resolve_refs_recursively


pinzon (approved):
Awesome!!

PR #8723, opened 2023-07-19
[SFN] Support for ItemSelector in Map states

MEPalma:
\* Extends https://github.com/localstack/localstack/pull/8720

Adds support for `ItemSelector` declarations in `Map` states, and relevant tests and fixes.


dominikschubert (approved):
:rocket:

PR #8725, opened 2023-07-19
8701 enhancement request custom lambda authorizer apikeysource is not validated

calvernaz:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-key-source.html

> When you associate a usage plan with an API and enable [API key](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-basic-concept.html#apigateway-definition-api-key)s on API methods, every incoming request to the API must contain an API key. API Gateway reads the key and compares it against the keys in the usage plan. If there is a match, API Gateway throttles the requests based on the plan's request limit and quota. Otherwise, it throws an InvalidKeyParameter exception. As a result, the caller receives a 403 Forbidden response.

The changes go a bit beyond the api key authorizer implementation by adding usage plans parity.
  
### Changes

- Override `create_api_key`, `create_usage_plan`, `update_usage_plan`, `get_usage_plan` and `get_usage_plans` and snapshoting.
- Implements and overrides request authorizer (api key)upstream and validates the api key against usage plans. 

### TF sample

https://github.com/localstack/localstack-terraform-samples/tree/master/apigateway-rest-lambda-api-key

### DIFF
--- a/localstack/services/apigateway/invocations.py
+++ b/localstack/services/apigateway/invocations.py
@@ -1,6 +1,5 @@
 import json
 import logging
-from typing import Dict
 
 from jsonschema import ValidationError, validate
 from requests.models import Response
@@ -35,7 +34,6 @@ from localstack.services.apigateway.integration import (
     StepFunctionIntegration,
 )
 from localstack.services.apigateway.models import ApiGatewayStore
-from localstack.utils.aws import aws_stack
 
 LOG = logging.getLogger(__name__)
 
@@ -142,21 +140,6 @@ class RequestValidator:
 # ------------
 
 
-def run_authorizer(invocation_context: ApiInvocationContext, authorizer: Dict):
-    # TODO implement authorizers
-    pass
-
-
-def authorize_invocation(invocation_context: ApiInvocationContext):
-    region_name = invocation_context.region_name or aws_stack.get_region()
-    client = connect_to(region_name=region_name).apigateway
-    authorizers = client.get_authorizers(restApiId=invocation_context.api_id, limit=100).get(
-        "items", []
-    )
-    for authorizer in authorizers:
-        run_authorizer(invocation_context, authorizer)
-
-
 def validate_api_key(api_key: str, invocation_context: ApiInvocationContext):
 
     usage_plan_ids = []
@@ -167,13 +150,14 @@ def validate_api_key(api_key: str, invocation_context: ApiInvocationContext):
     usage_plans = client.get_usage_plans()
     for item in usage_plans.get("items", []):
         api_stages = item.get("apiStages", [])
-        for api_stage in api_stages:
+        usage_plan_ids.extend(
+            item.get("id")
+            for api_stage in api_stages
             if (
                 api_stage.get("stage") == invocation_context.stage
                 and api_stage.get("apiId") == invocation_context.api_id
-            ):
-                usage_plan_ids.append(item.get("id"))
-
+            )
+        )
     for usage_plan_id in usage_plan_ids:
         usage_plan_keys = client.get_usage_plan_keys(usagePlanId=usage_plan_id)
         for key in usage_plan_keys.get("items", []):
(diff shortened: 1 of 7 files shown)


localstack/services/apigateway/invocations.py
```diff
@@ -192,18 +203,22 @@ def is_api_key_valid(invocation_context: ApiInvocationContext) -> bool:
...
+    match api_key_source:
+        case "HEADER":
+            api_key = invocation_context.headers.get("X-API-Key")
+            return validate_api_key(api_key, invocation_context) if api_key else False
+        case "AUTHORIZER":
+            # When the apiKeySource is set to AUTHORIZER, the authorizer is supposed to return the API key as a field
+            # `usageIdentifierKey`
+            # see https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html
+            # Authorizers are only mocked, so we can't validate the key. Return True in that case
+            return run_authorizer(invocation_context, rest_api.get("authorizer"))
```
bentsku, line 221: I think `run_authorizer` does not return `True` when not overridden, which means it will always fail for community

bentsku (commented):
I am not sure about the whole architecture of this implementation: I would not use `run_authorizer` to validate the result of the API key, validating the authorizer response again when it's already done once in -ext. In fact, I think we could remove this part of the code now that we have a proper authorizer flow. 

I think the whole key validation is already implemented properly in the -ext PR, and is part of the `authorize` call. Can't we just always return `True` if the case is `AUTHORIZER`, and the regular authorizer flow will take care of validating the key before arriving in this part of the code?

We could also maybe share the `validate_api_key` code, yours in -ext seems better now?

If I misunderstood something happy to be corrected, it's a bit hard to follow the flow between community and -ext 😅

localstack/services/apigateway/invocations.py
calvernaz: yep, after the discussion, I'm happy to not hook it here.

calvernaz:
> I am not sure about the whole architecture of this implementation: I would not use `run_authorizer` to validate the result of the API key, validating the authorizer response again when it's already done once in -ext. In fact, I think we could remove this part of the code now that we have a proper authorizer flow.
> 
> I think the whole key validation is already implemented properly in the -ext PR, and is part of the `authorize` call. Can't we just always return `True` if the case is `AUTHORIZER`, and the regular authorizer flow will take care of validating the key before arriving in this part of the code?
> 
> We could also maybe share the `validate_api_key` code, yours in -ext seems better now?
> 
> If I misunderstood something happy to be corrected, it's a bit hard to follow the flow between community and -ext sweat_smile

After discussing this, I'm happy to remove the hook, I think the current implementation doesn't run the multiple authorizers but it's beyond this PR. Having the validation on community makes sense, I was reusing the only place where the the usageIdentifier key is used, which is in the request authorizer, but it's passed on the response of the authorizer as well so it can be validated outside and reuse the same code to validate just the key and leave the authorizer intact.

localstack/services/apigateway/invocations.py
```diff
@@ -192,18 +176,19 @@ def is_api_key_valid(invocation_context: ApiInvocationContext) -> bool:
...
+    # - AUTHORIZER to read the API key from the UsageIdentifierKey from a custom authorizer.
 
-    return validate_api_key(api_key, invocation_context)
+    api_key_source = rest_api.get("apiKeySource")
+    match api_key_source:
+        case "HEADER":
+            api_key = invocation_context.headers.get("X-API-Key")
+            return validate_api_key(api_key, invocation_context) if api_key else False
+        case "AUTHORIZER":
+            api_key = invocation_context.auth_identity.get("apiKey")
```
calvernaz, line 190: this is populated on the authorizer logic upstream, if `usageIdentifierKey` is populated.

bentsku: I know this is annoying, but would there be a way to short circuit if the authorizer hasn't run? Some way to know that this is only community, so that we don't block people who don't have access to authorizers, and return True in that case? 😬 sorry for piling up

calvernaz: Well the way I had before was sensible to that (minus the return that could be easily changed), because either was implemented or not. Now moving the logic out of authorizers and `auth_identity` being the place where that is populated if the authorizer is executed, blocked if not. I don't think we have a way, proper way to check on features are enabled of disabled and definitely not want to hack around without proper checkpoints

bentsku: Okay, then let's see. People can also remove the authorizer from their setup if this is blocking.

calvernaz: patches bring their own issues with debugging and refactoring since moving code around implies changes on `-ext` that are invisible until that code is executed. There is definitely a balance I'm just on the opinion that the patch approach is not developer friendly

bentsku: Agreed, I'm not fond of them either. I think we'll go forward with that!

bentsku (approved):
LGTM! We had a a discussion around it where it was decided that most of the logic would remain in this part of the code, and just expose the API key returned by the lambda authorizer so that we could validate it. Awesome to see the parity increase for usage plans and api keys! And also the cleanup, that's a really great catch.

PR #8727, opened 2023-07-19
fix support for conditional refering to another conditional

pinzon:
This PR allows a Conditional ref another Conditional.

### DIFF
--- a/tests/integration/cloudformation/engine/test_conditions.py
+++ b/tests/integration/cloudformation/engine/test_conditions.py
@@ -378,20 +378,20 @@ class TestCloudFormationConditions:
         assert topic_arn_with_suffix.split(":")[-1] == f"{topic_prefix}-{region}-{suffix}"
 
     @pytest.mark.aws_validated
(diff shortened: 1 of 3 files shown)


tests/integration/cloudformation/engine/test_conditions.py
```diff
@@ -376,3 +376,22 @@ def test_sub_in_conditions(self, deploy_cfn_template, aws_client):
         topic_arn_with_suffix = stack.outputs["TopicWithSuffixRef"]
         aws_client.sns.get_topic_attributes(TopicArn=topic_arn_with_suffix)
         assert topic_arn_with_suffix.split(":")[-1] == f"{topic_prefix}-{region}-{suffix}"
+
+    @pytest.mark.aws_validated
+    @pytest.mark.parametrize("env", ["dev", "production"])
```
simonrw, line 381: minor: worth parametrizing the region as well? Just to check both sides of the conditional

simonrw (approved):
Looks good, thanks for adding this!

PR #8729, opened 2023-07-20
Introduce Markers class for all our custom pytest marks

dominikschubert:
- New class at `localstack.testing.pytest.marking.Markers` that should be used when assigning one of our custom pytest marks. Usage: `@Markers.snapshot.skip_snapshot_verify`
- Migrated `aws_validated`, `only_localstack`, `skip_offline` and `skip_snapshot_verify` marks in all integration tests to use the new `Markers` class.

This makes discovering our set of marks much easier and allows us to add proper typing support for our custom markers like so:

image


pinzon (approved):
Awesome 😎

localstack/testing/pytest/marking.py
```diff
@@ -0,0 +1,33 @@
...
+Custom pytest mark typings
+"""
+from typing import Callable, Optional
+
+import pytest
+
+
+class ParityMarkers:
+    aws_validated = pytest.mark.aws_validated
+    only_localstack = pytest.mark.aws_validated
```
steffyP, line 11: shouldn't that be `@pytest.mark.only_localstack`? 🤔

steffyP (approved):
I really like it, I think this will make marking the tests much easier 🚀 👍

simonrw (approved):
I love adding typing to our markers - you're right about getting parameters and names right. My nit-picky brain is slightly annoyed about the capital "Markers" which looks a little weird as a decorator, but I can live with it!

thrau (approved):
great idea. and agree with @simonrw, totally a nit, but maybe we can do module global instances of the classes that hold the different markers.

like `markers = Markers()`, you could even put that line in the `localstack/testing/pytest/__init__.py` file if you want to do `from localstack.testing.pytest import mark`
`@markers.skip_offline`. which looks more pythonic :-)

bentsku (approved):
LGTM! Awesome! This is great, I can say goodbye to my PyCharm snippets 😄 I agree with Simon and Thomas regarding the capital letter.

Issue #8313, opened 2023-05-13
bug: sts assume-role succeeds even if role doesn't exist


PikaBlue107:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running `sts assume-role`, the request may succeed even if the role at the specified ARN is not already defined.

### Expected Behavior

When running `sts assume-role`, the request should fail if the role at the specified ARN is not already defined.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal sts assume-role --role-arn arn:aws:iam::000000000000:role/this-role-doesnt-exist --role-session-name fear-me

### Environment

```markdown
- OS: macOS Monterey 12.6.5 (21G531)
- LocalStack: latest
```

### Anything else?

This actually makes things easier for me ([relevant xkcd](https://xkcd.com/1172/)), but still seems unintentional, or at least undocumented.

General references:
- LocalStack STS documentation: https://docs.localstack.cloud/references/coverage/coverage_sts/
- AWS CLI `sts assume-role` documentation: https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html

ccatterina:
Hi, I was looking at this issue because I would like to contribute to the project. But I don't think this is a bug because it is a documented and motivated behaviour of moto (http://docs.getmoto.org/en/latest/docs/services/sts.html) which I believe localstack uses to implement the sts service. 

> Assume an IAM Role. Note that the role does not need to exist. The ARN can point to another account, providing an opportunity to switch accounts.

PR #8730, opened 2023-07-20
fix(sts): check role existence

pandomic:
## Summary

Check role existence before assuming it.
AWS just throws `AccessDenied` all the time.

## Linked issues

https://github.com/localstack/localstack/issues/8313


dfangl (commented):
When using `ENFORCE_IAM=1`, this is automatically enforced - as no one has the permissions to assume a role which does not exist.

I think this is a very breaking change - it will fail a lot of services where we did not assert roles existence before - Lambda for example. Currently the requirement for lambda roles is "has to be a valid arn, but does not have to exist", to make it easier for users. With this, this will not be possible anymore, and any execution will fail.

pandomic:
@dfangl thanks for an update, I suspected this is something that should be already handled by the IAM enforcement.

However, maybe due to misconfiguration on my side or some missing peace the following results in a false-positive:

```bash
docker run \                                                                                                                                                                                                                                                  
  --rm -it \
  -p 4566:4566 \
  -p 4510-4559:4510-4559 \
  -e ENFORCE_IAM=1 \
  -e IAM_SOFT_MODE=0 \
  -e DEBUG=1 \
  -e LOCALSTACK_API_KEY=<...>\
  localstack/localstack-pro

awslocal sts assume-role \
  --role-arn arn:aws:iam::000000000000:role/random-role-name \
  --role-session-name random-session-name
```

PR #8734, opened 2023-07-21
Use `docker compose` as fallback if docker-compose is not available

### LINKED ISSUE #8732: bug: localstack config validate: `Errno 2 no such file or directory: 'docker-compose'`

### Current Behavior

- following the [installation steps](https://docs.localstack.cloud/getting-started/installation/) fails when validating localstack configuration

- error
image

- localstack is indeed running
image

### Expected Behavior

- localstack can consume the path I pass `--file`

With a docker-compose file

    docker compose up -d
    localstack config validate --file ./compose.yaml

### Environment

```markdown
OS: Ubuntu 20.04
    LocalStack: latest
```

### Anything else?

- you can reproduce [on this branch](https://github.com/noahehall/localstack/tree/rds) if it still exists
  - else switch to the develop branch

dfangl:
## Motivation
Since docker compose v2 is available, for many systems, `docker-compose` is not available any more (and not aliased to `docker compose`).

For these, we have to have a fallback to `docker compose`.

Fixes #8732

## Changes
* Use `docker compose` instead of `docker-compose` if the latter is not available.

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -25,7 +25,7 @@ from localstack.utils.container_utils.docker_cmd_client import CmdDockerClient
 from localstack.utils.docker_utils import DOCKER_CLIENT
 from localstack.utils.files import cache_dir, mkdir
 from localstack.utils.functions import call_safe
-from localstack.utils.run import run, to_str
+from localstack.utils.run import is_command_available, run, to_str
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Great catch! The fix is looking good, thanks for jumping on this! 🚀

PR #8746, opened 2023-07-26
fix cors for /_aws calls

lukqw:
This PR fixes missing cors headers on calls made to `/_aws` routes.


alexrashed (approved):
Great catch! The fix is simple and looks good! 🚀 
In the long run, it would be great to move to an approach where we map services to the routes and use them here in the service router.

burakkuzuu:
i didn't understand what am i haved to do...

PR #8749, opened 2023-07-26
Refactor `unittest` integration tests to use `pytest`

viren-nadkarni:
This PR refactors some of our older integration tests from `unittest` to `pytest`.

This will allow use of modern fixtures, `aws_client`, `aws_client_factory`, `secondary_aws_client` etc., that use the new internal client factory. Consequently, this will help improve test coverage for cross- and multi-account setups.


dfangl (approved):
LGTM! Thanks for churning through this!

PR #8751, opened 2023-07-26
Fix partition rewriting when body is urlencoded

dfangl:
## Motivation
For services which send their data with content type `application/x-www-form-urlencoded`, the partition reworking currently does not work properly, as it will match a urlencoded string with a not-urlencoded regex. 

Example: IAM

## Changes
* Check in content type header if content type is `application/x-www-form-urlencoded`, and use encoded logic for the replacement in these cases.

### DIFF
--- a/localstack/aws/handlers/partition_rewriter.py
+++ b/localstack/aws/handlers/partition_rewriter.py
@@ -101,8 +101,11 @@ class ArnPartitionRewriteHandler(Handler):
             get_full_raw_path(request), self.DEFAULT_INBOUND_PARTITION, encoded=True
         )
         parsed_forward_rewritten_path = urlparse(full_forward_rewritten_path)
+        body_is_encoded = "application/x-www-form-urlencoded" in request.headers.get(
+            "content-type", ""
+        )
         forward_rewritten_body = self._adjust_partition(
-            restore_payload(request), self.DEFAULT_INBOUND_PARTITION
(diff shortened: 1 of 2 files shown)


localstack/aws/handlers/partition_rewriter.py
```diff
@@ -101,8 +101,11 @@ def modify_request(self, request: Request) -> Request:
             get_full_raw_path(request), self.DEFAULT_INBOUND_PARTITION, encoded=True
         )
         parsed_forward_rewritten_path = urlparse(full_forward_rewritten_path)
+        body_is_encoded = "application/x-www-form-urlencoded" in request.headers.get(
+            "content-type", ""
+        )
```
thrau, line 106:
```suggestion
        body_is_encoded = request.mimetype == "application/x-www-form-urlencoded"
```

See:
```python
    @property
    def mimetype(self) -> str:
        """Like :attr:`content_type`, but without parameters (eg, without
        charset, type etc.) and always lowercase.  For example if the content
        type is ``text/HTML; charset=utf-8`` the mimetype would be
        ``'text/html'``.
        """
```

thrau (approved):
LGTM! just a suggestion to clean up the encoding check

localstack/aws/handlers/partition_rewriter.py
dfangl: Didn't know about that one, thanks!

PR #8753, opened 2023-07-26
CFn: fix nested dependency issue with Subs

simonrw:
The example template was mentioned to raise the error:

```
ResourceStatusReason: replace() argument 2 must be str, not dict
```

This is because the `Sub` intrinsic itself depended on a value that was itself a call to an intrinsic.

This PR catches this case and indicates that the second level dependency is not ready by raising a `DependencyNotYetSatisfied` exception.


tests/integration/cloudformation/test_template_engine.py
```diff
@@ -246,6 +246,14 @@ def test_get_azs_function(self, deploy_cfn_template):
         zone = "us-east-1a"  # TODO parametrize
         assert zone in deployed.outputs["Zones"]
 
+    def test_sub_not_ready(self, deploy_cfn_template):
```
dominikschubert, line 249:
```suggestion
    @markers.parity.aws_validated
    def test_sub_not_ready(self, deploy_cfn_template):
```

dominikschubert (approved):
LGTM :+1:

tests/integration/cloudformation/test_template_engine.py
simonrw: Thanks, nice use for the new marker! I've integrated this into a new commit

Issue #7268, opened 2022-11-30
feature request: Event Bridge Scheduler


thantos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Support Event Bridge Scheduler and CFN Resources for the Scheduler (Schedule, ScheduleGroup).

https://aws.amazon.com/eventbridge/scheduler/

### 🧑‍💻 Implementation

Extend the Scheduled Rules support to one time schedules and the new API/SDK. 

### Anything else?

https://aws.amazon.com/eventbridge/scheduler/

OAFCROB:
For anyone Googling this and wondering how to handle this in a Dockerised local environment. Then the approach we've taken is to encapsulate the scheduler SDK calls in a wrapper service. We then simply fake that service out for the local docker environment. Hope that helps, until the feature has been implemented.

Logic to determine which service to use. 
```
/**
 * Currently localstack does not support AWS's EventBridge Scheduler, meaning we have to fake our wrapper service out for local environment.
 * @see https://github.com/localstack/localstack/issues/7268
 */
const eventService = config.isLocal()
    ? getEventServiceFaker(logger)
    : getEventService(logger, config);
```

Local fake service
```
const service = (logger) => ({
    createScheduledEvent: async (event) => {
        logger.info(
            { event },
            'EventServiceFaker called on createScheduledEvent()'
        );
    },
    rescheduleEvent: async (event) => {
        logger.info({ event }, 'EventServiceFaker called on rescheduleEvent()');
    },
});

export default service;
```

Service example
```
import {
    SchedulerClient,
    CreateScheduleCommand,
} from '@aws-sdk/client-scheduler';

const service = (logger, config) => ({
    createScheduledEvent: async (event) => {
        logger.info({ event }, 'Attempting to schedule an event');

        try {
            const scheduler = new SchedulerClient({
                region: config.getAwsRegion(),
            });


            await scheduler.send(
                new CreateScheduleCommand({
                    /** 
                     * AWS create schedule command input object
                     * @see https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-scheduler/interfaces/createschedulecommandinput.html 
                     */
                }
            );

            logger.error({ event }, 'Successfully scheduled an event');
        } catch (e) {
            logger.error({ event, error: e }, 'Failed to schedule an event');
        }
    },
    rescheduleEvent: async (event) => {
        logger.info({ event }, 'Attempting to reschedule an event');

        /**
[...]

PR #8754, opened 2023-07-27
Introducing EventBridge Scheduler

viren-nadkarni:
This is one of the most requested AWS service.

EventBridge Scheduler has a relatively simple API with only about [a dozen operations](https://docs.aws.amazon.com/scheduler/latest/APIReference/API_Operations.html). For now, this PR creates a basic Moto dispatcher provider, but there's scope to add emulated functionality.

See https://discuss.localstack.cloud/t/support-for-eventbridge-scheduler/283

### Related

Closes https://github.com/localstack/localstack/issues/7268

### DIFF
--- a/localstack/aws/api/scheduler/__init__.py
+++ b/localstack/aws/api/scheduler/__init__.py
@@ -1,0 +1,844 @@
+import sys
+from datetime import datetime
+from typing import Dict, List, Optional
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+from localstack.aws.api import RequestContext, ServiceException, ServiceRequest, handler
+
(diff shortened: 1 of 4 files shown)


thrau:
@viren-nadkarni any idea what type of service integrations will work?

viren-nadkarni:
> @viren-nadkarni any idea what type of service integrations will work?

@thrau: None I'm afraid. Only CRUD/mock support for now

whummer (approved):
Kudos for adding this often requested feature @viren-nadkarni . 🚀  

The moto provider implementation is only ~270 LOC, so if and when we start getting more requests for specific features, it would make sense to bootstrap our own implementation.. I think we could also reuse a bit of functionality from our EventBridge implementation (EventBridge rules also have schedules, which have some commonalities..) 👍

PR #8758, opened 2023-07-27
implement read/write lock for s3 stream

bentsku:
We overlooked the read/write issue for S3, only locking for position in the stream but not modification of the underlying stream during iteration while returning responses.

We can make use of the already installed dependency `readerwriterlock` (https://github.com/elarivie/pyReaderWriterLock) with a nice explanation of the problem: https://wikipedia.org/wiki/Readers–writers_problem

Basically, when a thread is currently returning a S3 Object, it will read chunk by chunk from a stream. We want multiple threads to be able to read the same stream, so we can't fully lock around the iteration. But we want to lock writing at the time of the iteration, so that the underlying stream is not being modified while a "reader" is currently iterating and returning the object. 

Edit: I tried to remove the seek(0) when retrieving the value, which showed issue in CopyObject (that are fixable) and also one more glaring one in Website, where we directly access the moto objects... which is not great. I reverted so we keep the same as before.

### DIFF
--- a/localstack/services/s3/provider_stream.py
+++ b/localstack/services/s3/provider_stream.py
@@ -21,6 +21,7 @@ from moto.s3 import exceptions as s3_exceptions
 from moto.s3 import models as s3_models
 from moto.s3 import responses as s3_responses
 from moto.s3.utils import clean_key_name
+from readerwriterlock import rwlock
 from requests.structures import CaseInsensitiveDict
 
 from localstack import config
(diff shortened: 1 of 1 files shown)


localstack/services/s3/provider_stream.py
```diff
@@ -242,14 +242,16 @@ def copy_object(
...
+                dest_moto_bucket, key=request["Key"]
+            )
             dest_key_object.checksum_algorithm = checksum_algorithm
 
             if not source_key_object.checksum_value:
                 stream_value: SpooledTemporaryFile = dest_key_object.value
                 checksum = get_s3_checksum(checksum_algorithm)
 
-                with dest_key_object.lock:
+                with dest_key_object.lock, dest_key_object.read_lock:
```
thrau, line 254: Q: i think i may be missing something fundamental. why do we need both locks here and in all the other blocks?

thrau (approved):
code looks good to me. frankly i'm not familiar enough with the current S3 implementation to really say much about the approach. but i trust you completely :-)

just one question regarding the locks you could maybe clarify

Issue #8756, opened 2023-07-27
bug: error on multipart upload to S3


mattjones0111:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using localstack S3 with the Amazon AWSSDK.S3 NuGet v3.7.104.11 in a dotnet application.

Triggering a multipart upload against an instance of the localstack:latest image produces an error.

Full stack trace

[...]

PR #8760, opened 2023-07-27
fix S3 case where aws-chunked chunk is bigger than read size

bentsku:
This PR fix the newly arised issue #8756

Basically we were missing a reset to the signal for the end of a chunk. Quick fix, and added a test covering this case.


alexrashed (approved):
Great catch! 1 LoC in the logic, dozens of lines for a (really nice) test! 🚀

PR #8764, opened 2023-07-28
Events: Refactor tests to use fixture-based structure

viren-nadkarni:
This PR converts the existing helper `cleanup` into a fixture so that we can use the `aws_client` fixture and remove the `aws_stack.create_external_boto_client` utility


dominikschubert (approved):
LGTM :+1:

PR #8765, opened 2023-07-28
APIGW: Refactor tests to use the aws_client fixtures

viren-nadkarni:
This PR removes the use of `aws_stack.create_external_boto_client` in the API Gateway test suite.


dominikschubert (approved):
LGTM :rocket: 

Really hope we can remove then soon `aws_stack.create_external_boto_client` altogether :crossed_fingers:

PR #8767, opened 2023-07-28
New pytest parity markers & marker registration

dominikschubert:
Introduce new aws markers `@markers.aws.*`

```python
    # test has been successfully run against AWS, ideally multiple times
    validated = pytest.mark.aws_validated

    # implies aws_validated. test needs additional setup, configuration or some other steps not included in the test setup itself
    manual_setup_required = pytest.mark.aws_manual_setup_required

    # fails against AWS but should be made runnable against AWS in the future, basically a TODO
    needs_fixing = pytest.mark.aws_needs_fixing

    # only runnable against localstack by design
    only_localstack = pytest.mark.aws_only_localstack

    # it's unknown if the test works (reliably) against AWS or not
    unknown = pytest.mark.aws_unknown
```

### DIFF
--- a/localstack/testing/pytest/marking.py
+++ b/localstack/testing/pytest/marking.py
@@ -33,3 +33,5 @@ class Markers:
 
     skip_offline = pytest.mark.skip_offline
     only_on_amd64 = pytest.mark.only_on_amd64
+    resource_heavy = pytest.mark.resource_heavy
+    only_in_docker = pytest.mark.only_in_docker
--- a/localstack/testing/pytest/marking.py
+++ b/localstack/testing/pytest/marking.py
@@ -23,6 +23,11 @@ class SkipSnapshotVerifyMarker:
         ...
 
 
+class MultiRuntimeMarker:
+    def __call__(self, *, scenario: str, runtimes: Optional[list[str]] = None):
+        ...
+
+
 class SnapshotMarkers:
     skip_snapshot_verify: SkipSnapshotVerifyMarker = pytest.mark.skip_snapshot_verify
 
@@ -31,6 +36,9 @@ class Markers:
     parity = ParityMarkers
     snapshot = SnapshotMarkers
 
+    multiruntime: MultiRuntimeMarker = pytest.mark.multiruntime
+
+    # test selection
     skip_offline = pytest.mark.skip_offline
     only_on_amd64 = pytest.mark.only_on_amd64
     resource_heavy = pytest.mark.resource_heavy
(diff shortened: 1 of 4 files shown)


localstack/testing/pytest/marking.py
```diff
@@ -29,5 +36,10 @@ class Markers:
     parity = ParityMarkers
     snapshot = SnapshotMarkers
 
+    multiruntime: MultiRuntimeMarker = pytest.mark.multiruntime
```
joe4dev, line 39: The multiruntime tests are kind of a generalization of the runtime markers such as `parametrize_python_runtimes` in `tests/integration/awslambda/test_lambda_runtimes.py:63`, right?

tests/conftest.py
```diff
@@ -32,11 +32,37 @@ def pytest_configure(config):
...
         "aws_validated: mark the test as validated / verified against real AWS",
     )
 
+    config.addinivalue_line(
+        "markers",
+        "only_localstack: mark the test as inherently incompatible with AWS, e.g. when testing localstack-specific features",
+    )
+    config.addinivalue_line(
+        "markers",
+        "should_be_aws_validated: test fails against AWS but it shouldn't. Might need refactoring, additional permissions, etc.",
```
joe4dev, line 55: What about `todo_aws_validated`, `fix_aws_validated`, or `needs_to_be_aws_validated` ?

"Should" could be misunderstood as "it should work" but not being sure 🤷‍♂️

joe4dev (approved):
Love the extra IDE hints for markers 👍 

nits:

* `should_be_aws_validated` could potentially be mis-understood
* not sure whether I like the implicit `aws_validated` of manual setup but it makes it easier to filter ...

localstack/testing/pytest/marking.py
dominikschubert: They are actually a bit more. They're used in `test_lambda_common.py` and they go hand-in-hand with the injected `multiruntime_lambda` fixture which handles building the corresponding handler packages.

The code for the marker/fixture can be found in `tests/integration/awslambda/conftest.py`

dominikschubert: At some point we wanted to generalize this multiruntime lambda a bit to make it usable in other areas as well, but let's reevaluate this after we proceed with cdk-based test setups soon.

tests/conftest.py
joe4dev: good idea 👍

dfangl (approved):

steffyP (approved):
Looking forward to using the new markers 😄 👍 
nice work 🚀 🙏

PR #8768, opened 2023-07-28
Split out bootstrap tests for CircleCI

simonrw:
Move the bootstrap tests into their own location and run them in CircleCI

* the tests don't need to run against multiple python versions
* the tests are now run on CircleCI concurrently with the integration tests

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -499,7 +499,6 @@ class LocalstackContainerServer(Server):
                 'LocalStack container named "%s" is already running' % self.container.name
             )
 
-        config.dirs.mkdirs()
         return self.container.run()
 
     def do_shutdown(self):
--- a/tests/bootstrap/conftest.py
+++ b/tests/bootstrap/conftest.py
@@ -1,0 +1,8 @@
+import pytest
+
+from localstack import config
+
+
+@pytest.fixture(scope="session", autouse=True)
+def setup_host_config_dirs():
+    config.dirs.mkdirs()


localstack/utils/bootstrap.py
```diff
@@ -499,6 +499,7 @@ def do_run(self):
                 'LocalStack container named "%s" is already running' % self.container.name
             )
 
+        config.dirs.mkdirs()
```
alexrashed, line 502: I guess this line was added because some tests failed with the new bootstrap tests (because it seems that they were depending on the CLI tests). But imho this doesn't belong here, because the `LocalstackContainer` class does not define any volume mounts for the config dirs on its own.

The `mkdirs` creates the directories depending on where it's executed. If it's executed in Docker, it will create the directories for within the Docker container. If it's executed on the host, it will create the directories for the host system.

The creation of the directories at runtime - within the container - is invoked in `start_infra`:
https://github.com/localstack/localstack/blob/f48dbda6218901e0d41dc0e5ac8cb9c6d67df35f/localstack/services/infra.py#L373

In our case here, since it's executed before the container is started, it will create the _host directories_.
Since these mounts are not configured by the LocalstackContainer class, but by a using component, it is also in the responsibility of this using compontent to create the directories. 
For starting with the CLI, the directories are created in `prepare_docker_start`:
https://github.com/localstack/localstack/blob/71cfcfe651094214300b15ef2e73bb4f1b44a257/localstack/utils/bootstrap.py#L517
This is a utility function used in combination with `bootstrap.configure_container`, which configures the volumes.

TL/DR: Creating the dirs is in the responsibility of the users of this class. Please move this line to the caller(s).

alexrashed (changes requested):
We will extend the bootstrap tests quite a lot in the future (we have plenty of use-cases for these lifecycle tests), so this is nice ground-work preparing the tests to come.
I only found a small thing about the creation of the host directories, afterwards we're good to go! 🚀

localstack/utils/bootstrap.py
simonrw: That makes sense, thanks.

PR #8769, opened 2023-07-28
Add x-api-key to allowed CORS headers to enable access from Web app

whummer:
Add `x-api-key` to the list of allowed CORS headers, to enable passing this header from the LocalStack Web app.

This header is used, among others, for API Gateway and AppSync GraphQL APIs. For AppSync, we've recently extended our resource browser to make requests directly from the Web app. Currently this requires starting LS with `EXTRA_CORS_ALLOWED_HEADERS=x-api-key` configured, and with the changes in this PR it works out of the box.

This PR also contains a minor fix for a logging statement (missing parameter for placeholder in format string), which was detected during testing.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -645,11 +645,11 @@ class TemplateDeployer:
                 initialize=True,
                 action="CREATE",
             )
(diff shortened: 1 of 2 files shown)


pinzon (approved):
LGTM :+1:

PR #8771, opened 2023-07-30
use ContainerConfiguration object for LocalstackContainer

thrau:
## Motivation

There was a long standing TODO in the `LocalstackContainer` class: `# TODO merge with docker_utils.py:ContainerConfiguration`, which this PR fixes. This will be useful as we use the `LocalstackContainer` object as an abstraction for the bootstrap tests (cc @simonrw).
Also, I want to add some flags to the `localstack start` command (like `--network` to specify the container network the container should be started in), and this will make it a bit cleaner to do that.

## Changes
* Introduced `    def run_container_from_config(self, container_config: ContainerConfiguration) -> Tuple[bytes, bytes]:` to container client. adapted from the already existing `create_container_from_config`
* Refactored `LocalstackContainer` to use the `ContainerConfiguration` object internally to hold its configuration
* I was too lazy to refactor the config hooks, so i exposed a few of the config values through the `LocalstackContainer` as properties. This way it's also fully compatible with the ext container hooks and requires no changes there. The `additional_flags` thing is a bit of a mess.

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -1,3 +1,4 @@
+import copy
 import functools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


dfangl (approved):
LGTM! I agree, that we should migrate at least the additional flags to proper container configuration soon!

PR #8776, opened 2023-07-31
add S3 CORS test with referer header and regenerate CORS snaphots

bentsku:
Following a support case, added a test case validating that AWS does not return CORS headers if no `Origin` header is present, even if the `referer` one is there.

I took the opportunity to regenerate the snapshots, knowing there was now an issue regarding ACL, so I've added a new fixture to allow ACL on bucket. Something changed on AWS side, so I've fixed a small field for an Exception returned. 

Also refactored a bit the code to use the fix made in #6865, allowing us to add fields to the exception at instantiation as parameters and not by attaching them directly to the exception object like before. 

\cc @steffyP for the snapshot regeneration/fixture allowing AWS tests

### DIFF
--- a/localstack/services/s3/cors.py
+++ b/localstack/services/s3/cors.py
@@ -149,7 +149,10 @@ class S3CorsHandler(Handler):
 
                     context.operation = self._get_op_from_request(request)
                     raise AccessForbidden(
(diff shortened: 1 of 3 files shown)


steffyP (approved):
LGTM 👍 thanks for updating the snapshots @bentsku 🙏 🔝

PR #8781, opened 2023-07-31
Add bootstrap fixture for bootstrap tests

simonrw:
Add a fixture (and example tests) for black-box testing the LS container. Particularly: starting the container with different configurations and ensuring that it behaves as expected.

This fixture will be useful for:
* cli startup tests
* hypercorn configuration tests
* environment configuration tests
* DNS tests
* persistence tests
* startup script tests


localstack/utils/tail.py
```diff
@@ -43,6 +43,10 @@ def close(self):
         self.started.clear()
         self.thread = None
 
+    def truncate_log(self):
+        with open(self.file_path, "wb") as outfile:
+            outfile.write(b"")
```
thrau, line 48: nit: i think calling it `truncate_log` is incoherent with the level of abstraction that `FileListener` is on, which is much lower. could we just call it `truncate_file`?

localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -660,6 +660,11 @@ def start_container(
         LOG.debug("Start container with cmd: %s", cmd)
         return self._run_async_cmd(cmd, stdin, container_name_or_id)
 
+    def attach_to_container(self, container_name_or_id: str):
+        cmd = self._docker_cmd() + ["attach", container_name_or_id]
+        LOG.debug(f"Attaching to container {container_name_or_id}")
```
thrau, line 665: nit: don't use f-strings with loggers
```suggestion
        LOG.debug("Attaching to container %s", container_name_or_id)
```

thrau (changes requested):
Great set of refactorings! I have a few comments which are mostly minor. I wanted to just clarify where we stand on the issue of passing `ContainerConfiguration` to the hooks or rather stick with the `Container`. Should be reconciled with https://github.com/localstack/localstack/pull/8890

localstack/utils/container_utils/docker_cmd_client.py
thrau: 1) it's also a convention with `logging` we follow throughout the codebase :-)
2) the reason is that with `LOG.debug("Attaching to container %s", container_name_or_id)`, string creation is deferred until the log message is actually printed. if you make a `debug` call, but the log level is `info` or higher, the string will never actually be formatted. whereas with f-strings you're always creating the string even though you may not need it.

simonrw: Ok great, those are some good reasons. Thanks!

localstack/utils/bootstrap.py
```diff
@@ -434,74 +433,120 @@ def _ensure_container_network(self, network: str):
...
+                self.container.config.name
+            )
+
+        # block the current thread
+        self.container.attach()
+        return self.container
 
     def do_shutdown(self):
+        if not isinstance(self.container, RunningContainer):
+            raise ValueError(f"Container {self.container} not started")
```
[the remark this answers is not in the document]
simonrw: I've pulled the check outside the `do_shutdown` method so we don`t mess with the `threading.event` variables: https://github.com/localstack/localstack/pull/8781/commits/f341355878f3630ee03ebf1e4b2f078c284a978d - is that what you were getting at?

tests/bootstrap/conftest.py
```diff
@@ -1,8 +1,107 @@
...
 def setup_host_config_dirs():
     config.dirs.mkdirs()
+
+
+@pytest.fixture
+def wait_for_localstack_ready():
+    def _wait_for(container: RunningContainer, timeout: float | None = None):
+        container.wait_until_ready(timeout)
+
+        poll_condition(lambda: constants.READY_MARKER_OUTPUT in container.get_logs().splitlines())
```
thrau, line 105: should we add the timeout here as well just to be sure?

thrau (approved):
nice job @simonrw can't wait to start using this in #8994 :rocket:

tests/bootstrap/conftest.py
simonrw: Good spot, thanks will do :+1:

PR #8786, opened 2023-07-31
introduce new S3 native provider

bentsku:
First PR to implement the new S3 native provider not depending on moto. This implements all the basic abstractions, as well as the basic operations around buckets and objects. 

There are a lot of comments left, as it's still WIP, and this introduces only basic behaviour. We will need to implement all the bucket operations to be able to get the complete behaviour from basic operations. 

### List of implemented operations:
- CreateBucket
- DeleteBucket
- ListBuckets
- HeadBucket
- GetBucketLocation
- PutObject
- GetObject
- HeadObject
- DeleteObject
- DeleteObjects
- CopyObject
- ListObjects
- ListObjectsV2
- GetObjectAttributes
- RestoreObject

## Abstractions

This PR implements new abstractions around S3. We are splitting between "data", aka the buckets, objects and metadata around them, and the actual file object containing S3 Objects real content. This will allow us to properly split how we save these for persistence in the next iteration. 

We're introducing a new S3 `BaseStorageBackend` and `TemporaryStorageBackend`, which abstract away how those S3 file objects, which we will name "assets", are stored and retrieved. This will in turn allow us to either have a temporary filesystem, or the real filesystem in case of persistence, and give us greater flexibility.

The last core Operations and Models outside of Object and Bucket configuration are the multipart, coming in #8787

There a lot of bugs regarding the versioning code path, because it couldn't be tested before I could enable versioning on the bucket, which is now done and is fixed with #8799.

### DIFF
--- a/localstack/services/s3/constants.py
+++ b/localstack/services/s3/constants.py
@@ -1,7 +1,11 @@
 from localstack.aws.api.s3 import BucketCannedACL, ObjectCannedACL, Permission, StorageClass
 
 S3_VIRTUAL_HOST_FORWARDED_HEADER = "x-s3-vhost-forwarded-for"
+
 S3_UPLOAD_PART_MIN_SIZE = 5242880
+"""
+This is minimum size allowed by S3 when uploading more than one part for a Multipart Upload, except for the last part
+"""
 
 VALID_CANNED_ACLS_BUCKET = {
     # https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl
--- a/localstack/services/s3/constants.py
+++ b/localstack/services/s3/constants.py
@@ -1,6 +1,7 @@
 from localstack.aws.api.s3 import BucketCannedACL, ObjectCannedACL, Permission, StorageClass
 
 S3_VIRTUAL_HOST_FORWARDED_HEADER = "x-s3-vhost-forwarded-for"
+S3_UPLOAD_PART_MIN_SIZE = 5242880
 
 VALID_CANNED_ACLS_BUCKET = {
     # https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl
@@ -40,6 +41,13 @@ VALID_STORAGE_CLASSES = [
     StorageClass.DEEP_ARCHIVE,
 ]
 
+# TODO validate those?
+ARCHIVES_STORAGE_CLASSES = [
+    StorageClass.GLACIER,
+    StorageClass.GLACIER_IR,
+    StorageClass.DEEP_ARCHIVE,
+]
+
 # response header overrides the client may request
 ALLOWED_HEADER_OVERRIDES = {
     "ResponseContentType": "ContentType",
@@ -54,6 +62,16 @@ ALLOWED_HEADER_OVERRIDES = {
 # are creating bucket policies that enforce aws:SecureTransport, which makes the CDK deployment fail.
 # TODO: potentially look into making configurable
 ENABLE_MOTO_BUCKET_POLICY_ENFORCEMENT = False
+
+
+SYSTEM_METADATA_SETTABLE_HEADERS = [
+    "CacheControl",
+    "ContentDisposition",
+    "ContentEncoding",
+    "ContentLanguage",
+    "ContentMD5",
+    "ContentType",
+]
 
 # params are required in presigned url
 SIGNATURE_V2_PARAMS = ["Signature", "Expires", "AWSAccessKeyId"]
--- a/localstack/services/s3/exceptions.py
+++ b/localstack/services/s3/exceptions.py
@@ -1,0 +1,33 @@
+from localstack.aws.api import CommonServiceException
+
+
+class InvalidLocationConstraint(CommonServiceException):
+    def __init__(self, message=None):
+        super().__init__("InvalidLocationConstraint", status_code=400, message=message)
+
+
+class MalformedXML(CommonServiceException):
+    def __init__(self, message=None):
+        if not message:
+            message = "The XML you provided was not well-formed or did not validate against our published schema"
+        super().__init__("MalformedXML", status_code=400, message=message)
+
+
+class MalformedACLError(CommonServiceException):
+    def __init__(self, message=None):
+        super().__init__("MalformedACLError", status_code=400, message=message)
+
+
+class InvalidRequest(CommonServiceException):
+    def __init__(self, message=None):
+        super().__init__("InvalidRequest", status_code=400, message=message)
+
+
+class UnexpectedContent(CommonServiceException):
+    def __init__(self, message=None):
+        super().__init__("UnexpectedContent", status_code=400, message=message)
+
+
+class NoSuchConfiguration(CommonServiceException):
+    def __init__(self, message=None):
+        super().__init__("NoSuchConfiguration", status_code=404, message=message)
--- a/localstack/services/s3/models_native.py
+++ b/localstack/services/s3/models_native.py
@@ -1,0 +1,486 @@
+import logging
+from collections import defaultdict
+from datetime import datetime
+from io import RawIOBase
+from typing import IO, Any, Literal, Optional, Union
+
+from werkzeug.datastructures.headers import Headers
+
+from localstack import config
+from localstack.aws.api.s3 import (
+    AccountId,
+    AnalyticsConfiguration,
+    AnalyticsId,
+    BucketAccelerateStatus,
+    BucketName,
+    BucketRegion,
+    ChecksumAlgorithm,
+    CORSConfiguration,
+    ETag,
+    Expiration,
+    IntelligentTieringConfiguration,
+    IntelligentTieringId,
+    InvalidArgument,
+    LifecycleRules,
+    LoggingEnabled,
+    Metadata,
+    MethodNotAllowed,
+    MultipartUploadId,
+    NoSuchKey,
+    NoSuchVersion,
+    NotificationConfiguration,
+    ObjectKey,
+    ObjectLockConfiguration,
+    ObjectLockLegalHoldStatus,
+    ObjectLockMode,
+    ObjectOwnership,
+    ObjectStorageClass,
+    ObjectVersionId,
+    Owner,
+    Payer,
+    Policy,
+    PublicAccessBlockConfiguration,
+    ReplicationConfiguration,
+    Restore,
+    ServerSideEncryption,
+    ServerSideEncryptionRule,
+    Size,
+    SSEKMSKeyId,
+    StorageClass,
+    WebsiteConfiguration,
+    WebsiteRedirectLocation,
+)
+from localstack.services.s3.storage import LockedSpooledTemporaryFile
+from localstack.services.s3.utils import (
+    ParsedRange,
+    get_owner_for_account_id,
+    iso_8601_datetime_without_milliseconds_s3,
+    rfc_1123_datetime,
+)
+from localstack.services.stores import (
+    AccountRegionBundle,
+    BaseStore,
+    CrossAccountAttribute,
+    CrossRegionAttribute,
+)
+
+# TODO: beware of timestamp data, we need the snapshot to be more precise for S3, with the different types
+# moto had a lot of issue with it? not sure about our parser/serializer
+
+LOG = logging.getLogger(__name__)
+
+
+# note: not really a need to use a dataclass here, as it has a lot of fields, but only a few are set at creation
+class S3Bucket:
+    name: BucketName
+    bucket_account_id: AccountId
+    bucket_region: BucketRegion
+    creation_date: datetime
+    multiparts: dict[MultipartUploadId, Any]  # TODO
+    objects: Union["KeyStore", "VersionedKeyStore"]
+    versioning_status: Literal[None, "Enabled", "Disabled"]
+    lifecycle_rules: LifecycleRules
+    policy: Optional[Policy]
+    website_configuration: WebsiteConfiguration
+    acl: str  # TODO: change this
+    cors_rules: CORSConfiguration
+    logging: LoggingEnabled
+    notification_configuration: NotificationConfiguration
+    payer: Payer
+    encryption_rule: Optional[
+        ServerSideEncryptionRule
+    ]  # TODO validate if there can be more than one rule
+    public_access_block: PublicAccessBlockConfiguration
+    accelerate_status: BucketAccelerateStatus
+    object_ownership: ObjectOwnership
+    object_lock_configuration: Optional[ObjectLockConfiguration]
+    object_lock_enabled: bool
+    intelligent_tiering_configuration: dict[IntelligentTieringId, IntelligentTieringConfiguration]
+    analytics_configuration: dict[AnalyticsId, AnalyticsConfiguration]
+    replication: ReplicationConfiguration
+    owner: Owner
+
+    # set all buckets parameters here
+    def __init__(
+        self,
+        name: BucketName,
+        account_id: AccountId,
+        bucket_region: BucketRegion,
+        acl=None,  # TODO: validate ACL first, create utils for validating and consolidating
+        object_ownership: ObjectOwnership = None,
+        object_lock_enabled_for_bucket: bool = None,
+    ):
+        self.name = name
+        self.bucket_account_id = account_id
+        self.bucket_region = bucket_region
+        self.objects = KeyStore()
+        self.object_ownership = object_ownership
+        self.object_lock_enabled = object_lock_enabled_for_bucket
+        self.encryption_rule = None  # TODO
+        self.creation_date = datetime.utcnow()
+        self.multiparts = {}
+        self.versioning_status = None
+
+        # see https://docs.aws.amazon.com/AmazonS3/latest/API/API_Owner.html
+        self.owner = get_owner_for_account_id(account_id)
+
+    def get_object(
+        self,
+        key: ObjectKey,
+        version_id: ObjectVersionId = None,
+        http_method: Literal["GET", "PUT", "HEAD"] = "GET",  # TODO: better?
+    ) -> "S3Object":
+        """
+        :param key: the Object Key
+        :param version_id: optional, the versionId of the object
+        :param http_method: the HTTP method of the original call. This is necessary for the exception if the bucket is
+        versioned or suspended
+        see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeleteMarker.html
+        :return:
+        :raises NoSuchKey if the object key does not exist at all, or if the object is a DeleteMarker
+        :raises MethodNotAllowed if the object is a DeleteMarker and the operation is not allowed against it
+        """
+
+        if self.versioning_status is None:
+            if version_id and version_id != "null":
+                raise InvalidArgument(
+                    "Invalid version id specified",
+                    ArgumentName="versionId",
+                    ArgumentValue=version_id,
+                )
+
+            s3_object = self.objects.get(key)
+
+            if not s3_object:
+                raise NoSuchKey("The specified key does not exist.", Key=key)
+
+        else:
+            self.objects: VersionedKeyStore
+            if version_id:
+                s3_object_version = self.objects.get(key, version_id)
+                if not s3_object_version:
+                    raise NoSuchVersion(
+                        "The specified version does not exist.",
+                        Key=key,
+                        VersionId=version_id,
+                    )
+                elif isinstance(s3_object_version, S3DeleteMarker):
+                    raise MethodNotAllowed(
+                        "The specified method is not allowed against this resource.",
+                        Method=http_method,
+                        ResourceType="DeleteMarker",
+                        DeleteMarker=True,
+                        Allow="delete",
+                    )
+                return s3_object_version
+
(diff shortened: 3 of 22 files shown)


localstack/services/s3/provider_native.py
```diff
@@ -0,0 +1,1353 @@
...
+
+        if not version_id:
+            delete_marker_id = generate_version_id(s3_bucket.versioning_status)
+            delete_marker = S3DeleteMarker(key=key, version_id=delete_marker_id)
+            # TODO: verify with Suspended bucket? does it override last version or still append?? big question
+            #  I think it puts a delete marker with a `null` VersionId, which deletes the object under
+            s3_bucket.objects.set(key, delete_marker)
+            return DeleteObjectOutput(VersionId=delete_marker.version_id, DeleteMarker=True)
+
+        if key not in s3_bucket.objects:
```
bentsku, line 567: issue: KeyStore and VersionedKeyStore do not implement `__contains__` yet but this will be fixed in the next PR 😬 hard to fully test everything, this will be fixed as I can test more and more with more things implemented (implementing `PutBucketVersioning` allows me to test much more already and will contains all these small fixes).

localstack/services/s3/models_native.py
```diff
@@ -0,0 +1,486 @@
...
+        """
+        Set an S3 object, using its already set VersionId.
+        If the bucket versioning is `Enabled`, then we're just inserting a new Version.
+        If the bucket versioning is `Suspended`, the current object version will be set to `null`, so if setting a new
+        object at the same key, we will override it at the `null` versionId entry.
+        :param object_key: the key of the Object we are setting
+        :param s3_object: the S3 object or S3DeleteMarker to set
+        :return: None
+        """
+        self._store[object_key][s3_object.version_id] = S3Object
```
bentsku, line 374:
```suggestion
        self._store[object_key][s3_object.version_id] = s3_object
```
whoops, always the same issue with versioning and couldn't be tested. Will be part of the versioning PR 👍

localstack/services/s3/models_native.py
```diff
@@ -0,0 +1,486 @@
...
+
+        object_version = versions.pop(version_id, default)
+        if not versions:
+            self._store.pop(object_key)
+
+        return object_version
+
+    def values(self, with_versions: bool = False) -> list[S3Object | S3DeleteMarker]:
+        if with_versions:
+            return [object_version for values in self._store.values() for object_version in values]
```
bentsku, line 391:
```suggestion
            return [object_version for values in self._store.values() for object_version in values.values()]
```

localstack/services/s3/constants.py
```diff
@@ -55,6 +63,18 @@
...
+SYSTEM_METADATA_SETTABLE_HEADERS = [
+    "CacheControl",
+    "ContentDisposition",
+    "ContentEncoding",
+    "ContentLanguage",
+    "ContentMD5",
+    "ContentType",
+]
+
+STREAM_CHUNK_SIZE = 2048 * 2
```
thrau, line 76: is this something defined by S3 or something that we came up with? in that case, we could maybe clarify why it's that value. also, were you too lazy to calculate that it's 4096 ? :laughing:

localstack/services/s3/provider_native.py
```diff
@@ -0,0 +1,1353 @@
...
+        Botocore implements this hack for parsing the response in `botocore.handlers.py#parse_get_bucket_location`
+        """
+        store = self.get_store(context.account_id, context.region)
+        if not (s3_bucket := store.buckets.get(bucket)):
+            raise NoSuchBucket("The specified bucket does not exist", BucketName=bucket)
+
+        location_constraint = (
+            '<?xml version="1.0" encoding="UTF-8"?>\n'
+            '<LocationConstraint xmlns="http://s3.amazonaws.com/doc/2006-03-01/">{{location}}</LocationConstraint>'
+        )
```
thrau, line 243: s3 :smiling_face_with_tear:

localstack/services/s3/models_native.py
```diff
@@ -0,0 +1,486 @@
...
+        account_id: AccountId,
+        bucket_region: BucketRegion,
+        acl=None,  # TODO: validate ACL first, create utils for validating and consolidating
+        object_ownership: ObjectOwnership = None,
+        object_lock_enabled_for_bucket: bool = None,
+    ):
+        self.name = name
+        self.bucket_account_id = account_id
+        self.bucket_region = bucket_region
+        self.objects = KeyStore()
```
thrau, line 116: Q: I think haven't fully understood the `KeyStore`/`StorageBackend` separation and whether it is needed. I suppose one is for managing the metadata, and one for the actual object data?

localstack/services/s3/provider_native.py
```diff
@@ -0,0 +1,1353 @@
...
+        checksum_value = None
+        if checksum_algorithm := request.get("ChecksumAlgorithm"):
+            if is_aws_chunked:
+                checksum_value = body.trailing_headers.get(
+                    f"x-amz-checksum-{checksum_algorithm.lower()}"
+                )
+            else:
+                checksum_value = request.get(f"Checksum{checksum_algorithm.upper()}")
+
+        size, etag, calculated_checksum_value = readinto_fileobj(body, fileobj, checksum_algorithm)
```
thrau, line 321: Q: Shouldn't, conceptually, this entire logic be encapsulated in the storage backend? it seems to me that the decision when and how to write the object into a file object is the responsibility of the storage backend, not the provider.

as a thought experiment, i'm thinking about how to implement a "no-op" storage backend that just discards the files.

localstack/services/s3/storage.py
```diff
@@ -0,0 +1,282 @@
...
+        # TODO: might not need to hash it? just use the upload_id?
+        return str(hash(multipart_id))
+
+
+class BucketTemporaryFileSystem(TypedDict):
+    keys: dict[str, "LockedSpooledTemporaryFile"]
+    multiparts: dict[MultipartUploadId, dict[PartNumber, "LockedSpooledTemporaryFile"]]
+
+
+class TemporaryStorageBackend(BaseStorageBackend):
```
thrau, line 82: nit naming: why is it temporary?

localstack/services/s3/storage.py
```diff
@@ -0,0 +1,282 @@
...
+        raise NotImplementedError
+
+    def delete_bucket_directory(self, bucket_name: BucketName):
+        raise NotImplementedError
+
+    def create_upload_directory(self, bucket_name: BucketName, upload_id: MultipartUploadId):
+        raise NotImplementedError
+
+    def delete_upload_directory(self, bucket_name: BucketName, upload_id: MultipartUploadId):
+        raise NotImplementedError
```
thrau, line 39: Following up on the StorageBackend abstraction discussion: I'm also wondering whether the BaseStorageBackend shouldn't have fewer but more high-level methods. Like the creation of directories and such, shouldn't that be hidden from the S3Provider?

thrau (commented):
This already looks amazing @bentsku , well done! The separation into the StorageBackend is really good and the extra flexibility we gain will help a lot moving forward with issues around storage, performance, and persistence!

It's a big PR, but it already looks very good! Obviously lots of TODOs and blanks still in there, but that's fine at this stage. I'm also not going to nitpick much in the code. I don't see a lot of things that could really be improved, and I think it would be good to get a first version out ASAP.

Just some high level thoughts we could discuss:
The `KeyStore` / `StorageBackend` duality is interesting to me, and I'm noticing that the API needs to keep the two in sync, which could potentially be a source of bugs when not done correctly. I don't know whether there's any value in trying to combine the two concepts. In general, the separation of Metadata and Data management makes a lot of sense to me though.
It could make sense to add more higher-level constructs into the storage backend to push more functionality that the API currently does into the storage provider, basically hiding file objects all together.

localstack/services/s3/constants.py
bentsku: I think it's not even used 😅 will remove

localstack/services/s3/models_native.py
bentsku: Basically, the basic `KeyStore` could only be a `dict`, but it's there to hide away to complexity of the `VersionedKeyStore`. However, this can't really be mixed with the `StorageBackend` , as one is saved in our store, and the other one has no persistence. Basically, the `StorageBackend` is ephemeral, only there to hide how we store the file. Once we implement persistence, instead of storing the temporary files inside a dict, we will rely on the OS filesystem to retrieve the files. Right now, it seems like a lot of duplication as we need to store the ephemeral state inside `StorageBackend`

localstack/services/s3/provider_native.py
bentsku: I think we could implement `readinto_fileobj` inside the `StorageBackend`, but it would basically one operation for each different provider call, I think one method for each provider method: a different one for `CopyObject`, another one for `UploadPart`, `UploadCopyPart` and last for `PutObject`. Then we're coupling `StoragBackend` to the provider a lot. Not sure how much decoupling we really need

localstack/services/s3/storage.py
bentsku: I thought about it too, that it should be hidden, that was how I first wrote it. But then, if it's hidden from the provider, when do we create a directory? The first time we create a file object in the "bucket"? Then we need to check every time if the directory already exists. Following the "lifecycle" of the provider, which to me is the "orchestrator" of both Metadata (KeyStore, S3Buckets and such) and the StorageBackend. The issue as you pointed out is the synchronisation between both. We could need some kind of lock to be sure they're in sync, but already following the approach of always doing the updates in the metadata in the right order, but it need to be conscious. This is, in my opinion, the biggest flaw of the split between metadata/assets.

localstack/services/s3/v3/provider.py
```diff
@@ -0,0 +1,1231 @@
...
+        if range_header := request.get("Range"):
+            range_data = parse_range_header(range_header, s3_object.size)
+            s3_stored_object.seek(range_data.begin)
+            response["Body"] = LimitedIterableStream(
+                s3_stored_object, max_length=range_data.content_length
+            )
+            response["ContentRange"] = range_data.content_range
+            response[
+                "ContentLength"
+            ] = range_data.content_length  # TODO: should we set it for chunked encoding?
```
thrau, line 418:
```suggestion
            # TODO: should we set it for chunked encoding?
            response["ContentLength"] = range_data.content_length  
```

in werkzeug, the `Transfer-Encoding: chunked` header will be set automatically if the content-length is not set. if the content length is set, and you want it to be chunked, you need to also set the header explicitly.

i was told that #8841 is relevant here ;-)

localstack/services/s3/v3/provider.py
```diff
@@ -0,0 +1,1231 @@
...
+                raise InvalidArgument(
+                    "Invalid version id specified",
+                    ArgumentName="versionId",
+                    ArgumentValue=version_id,
+                )
+
+            found_object = s3_bucket.objects.pop(key, None)
+            # TODO: RequestCharged
+            if found_object:
+                self._storage_backend.remove(bucket, found_object)
```
thrau, line 526: q: are there any IO exceptions we should guard against? or maybe it's better to just pass them through as 500 internal server error

localstack/services/s3/v3/provider.py
```diff
@@ -0,0 +1,1231 @@
...
+                            Key=object_key,
+                            Message="The specified version does not exist.",
+                            VersionId=version_id,
+                        )
+                    )
+                    continue
+
+                found_object = s3_bucket.objects.pop(object_key, None)
+                if found_object:
+                    self._storage_backend.remove(bucket, found_object)
```
thrau, line 602: would it make sense to allow to pass a tuple/list of objects to remove? or add a method `remove_many` or something? that way the underlying storage system has an opportunity to optimize a batched remove.

localstack/services/s3/v3/provider.py
```diff
@@ -0,0 +1,1231 @@
...
+                if isinstance(found_object, S3DeleteMarker):
+                    deleted_object["DeleteMarker"] = True
+                    deleted_object["DeleteMarkerVersionId"] = found_object.version_id
+
+                deleted.append(deleted_object)
+
+            if isinstance(found_object, S3Object):
+                self._storage_backend.remove(bucket, found_object)
+
+        # TODO: request charged
```
thrau, line 650: ?

localstack/services/s3/v3/storage.py
thrau, line 1: nit: this file needs a bit more comments :-)

also, maybe once we have all PRs merged that build on this one, we could split the storage API (S3ObjectStore, S3StoredObject, ...) and everything in the default implementation (LockedSpooled..., etc) into different modules.
basically have `storage/core.py`, `storage/ephemeral.py` (or `storage/default.py`) etc. you can expose the api via `storage/__init__.py` to be able to `from localstack.services.s3.v3.storage import S3ObjectStore`

localstack/services/s3/v3/storage.py
```diff
@@ -0,0 +1,528 @@
...
+    @staticmethod
+    def _resolve_upload_directory(bucket_name: BucketName, upload_id: MultipartUploadId) -> str:
+        return f"{bucket_name}/{upload_id}"
+
+    def _create_bucket_directory(self, bucket_name: BucketName) -> str:
+        """
+        Create a temporary directory representing a bucket
+        :param bucket_name
+        """
+        tmp_dir = mkdtemp()
```
thrau, line 494: let's try and restructure the class so we can inject a `root_directory` path as dependency into the constructor, that is then used for all filesystem operations. this will help massively with persistence.

tests/unit/test_s3.py
```diff
@@ -1015,3 +1017,23 @@ def test_s3_aws_chunked_decoder_chunk_bigger_than_s3_chunk(self):
         assert stream.read(chunk_size + 1000) == total_body[:chunk_size]
         # assert that even if we read more, when accessing the rest, we're still at the same position
         assert stream.read(10) == total_body[chunk_size : chunk_size + 10]
+
+
+class TestS3TemporaryStorageBackend:
+    def test_get_fileobj_no_bucket(self):
+        temp_storage_backend = EphemeralS3ObjectStore()
```
thrau, line 1024: as suggested in the `EphemeralS3ObjectStore`, it would be great if we could use the `tmp_path` or `tmpdir` fixture and inject it into the class instead of using a global tmp directory.

localstack/services/s3/v3/storage.py
```diff
@@ -0,0 +1,528 @@
...
+
+        return stored_part
+
+
+class BucketTemporaryFileSystem(TypedDict):
+    keys: dict[str, LockedSpooledTemporaryFile]
+    multiparts: dict[MultipartUploadId, EphemeralS3StoredMultipart]
+
+
+class EphemeralS3ObjectStore(S3ObjectStore):
```
thrau, line 379: `BufferedFileSystemStore` ?

thrau (approved):
This iteration has made the code so much cleaner, thanks for bearing with me @bentsku and considering the changes!

The additional `S3ObjectStore` layer has made the Provider implementation much easier to understand, and isolation is much better :ok_hand:

My comments are mostly around adding more doc strings, and some minor tweaks. Some of the things can be tackled as follow-up PRs.

Really excellent work! :100: :rocket:

localstack/services/s3/exceptions.py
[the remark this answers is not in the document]
bentsku: Basically the idea was that if the exception needed something specific, like extra fields or specific default status code, we would generate it. Otherwise, we would follow the usual path of subclassing `CommonServiceException`, like many other services. I just pulled them out of the previous provider so that I can use in both of them, but in the future once we only have one again, we could pulled them again inside the provider file? Is that a good idea?

localstack/services/s3/utils.py
```diff
@@ -70,15 +79,16 @@
 RFC1123 = "%a, %d %b %Y %H:%M:%S GMT"
 
 
-class InvalidRequest(ServiceException):
-    code: str = "InvalidRequest"
-    sender_fault: bool = False
-    status_code: int = 400
+def get_owner_for_account_id(account_id: str):
+    return Owner(
```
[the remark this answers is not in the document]
bentsku: I will, it's basically a hardcoded string, not exactly sure yet basically, it was hardcoded in moto, but I'd rather have a method so that we extend it later on.

localstack/services/s3/utils.py
```diff
@@ -113,7 +131,7 @@ def get_s3_checksum(algorithm):
             raise InvalidRequest("The value specified in the x-amz-trailer header is not supported")
 
 
-class S3CRC32Checksum:
+class S3CRC32Checksum(ChecksumHash):
```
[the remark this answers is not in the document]
bentsku: Oops, right, thanks for the hint!

localstack/services/s3/v3/models.py
```diff
@@ -0,0 +1,607 @@
...
+)
+
+# TODO: beware of timestamp data, we need the snapshot to be more precise for S3, with the different types
+# moto had a lot of issue with it? not sure about our parser/serializer
+
+LOG = logging.getLogger(__name__)
+
+
+# note: not really a need to use a dataclass here, as it has a lot of fields, but only a few are set at creation
+class S3Bucket:
```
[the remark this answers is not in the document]
bentsku: no, this is more a sort of todo, when things are settled, to see what classes I can transform into dataclasses. S3Bucket is not one of them I would say, with a lot of default values

localstack/services/s3/v3/models.py
```diff
@@ -0,0 +1,607 @@
...
+        self.objects = KeyStore()
+        self.object_ownership = object_ownership
+        self.object_lock_enabled = object_lock_enabled_for_bucket
+        self.encryption_rule = None  # TODO
+        self.creation_date = datetime.utcnow()
+        self.multiparts = {}
+        self.versioning_status = None
+
+        # see https://docs.aws.amazon.com/AmazonS3/latest/API/API_Owner.html
+        self.owner = get_owner_for_account_id(account_id)
```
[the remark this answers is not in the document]
bentsku: Let's say it's not set in stone yet, like said above it was hardcoded in moto but I've pulled it as a method for future extendability. I'd say it's fine here for now and we can pull it out if needed?

localstack/services/s3/v3/models.py
```diff
@@ -0,0 +1,607 @@
...
+        self.lock_until = lock_until
+        self.acl = acl
+        self.expiration = expiration
+        self.website_redirect_location = website_redirect_location
+        self.is_current = True
+        self.last_modified = datetime.utcnow()
+        self.parts = []
+        self.restore = None
+
+    def get_system_metadata_fields(self):
```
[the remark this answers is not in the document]
bentsku: I think these should return a dict, because they are then merged into the TypedDict response like `**`. Thanks for pointing it out, I will correct that `Headers` thing 😬

localstack/services/s3/v3/models.py
```diff
@@ -0,0 +1,607 @@
...
+            object_etag.update(bytes.fromhex(s3_part.etag))
+            # TODO verify this, it seems wrong
+            self.object.parts.append((pos, s3_part.size))
+
+            multipart_etag = f"{object_etag.hexdigest()}-{len(parts)}"
+            self.object.etag = multipart_etag
+
+
+# TODO: use SynchronizedDefaultDict to prevent updates during iteration?
+class KeyStore:
```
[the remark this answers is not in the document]
bentsku: Would there be any upside? It was inheriting from dict in moto, but it was creating a bit more problem than solving? For example the `__getitem__` signature would be different between both?

localstack/services/s3/v3/provider.py
```diff
@@ -0,0 +1,1231 @@
...
+
+        body = request.get("Body")
+        # check if chunked request
+        headers = context.request.headers
+        is_aws_chunked = headers.get("x-amz-content-sha256", "").startswith("STREAMING-")
+        if is_aws_chunked:
+            decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
+            body = AwsChunkedDecoder(body, decoded_content_length)
+
+        # TODO check if key already exist, and if it is locked? so we don't override it
```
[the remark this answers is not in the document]
bentsku: Oh no sorry, this is an S3 internal, you can legally lock a key so that you can't override it or delete it. Sorry 😅

localstack/services/s3/v3/provider.py
bentsku: Yes, now I know we always need to return it 😄

localstack/services/s3/v3/provider.py
bentsku: good question, I have to say I'm not sure yet. I feel like 500 is fine for now

localstack/services/s3/v3/provider.py
bentsku: No yes it would make sense. Will add 👍

localstack/services/s3/v3/provider.py
bentsku: is this comment still about batching delete?

localstack/services/s3/v3/storage.py
bentsku: You mean in the base class? Because this one is not really supposed to be used with persistence? but for testing, sure, it makes a lot of sense

Issue #6797, opened 2022-09-01
bug: got an error while running localstack as a container in CI


okkez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I got the following error while running localstack as a container.

```
Error starting infrastructure: Could not get list of releases from https://api.github.com/repos/etspaceman/kinesis-mock/releases/tags/0.2.5: {"message":"API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
 Traceback (most recent call last):
  File "/opt/code/localstack/localstack/services/infra.py", line 419, in start_infra
    thread = do_start_infra(asynchronous, apis, is_in_docker)
  File "/opt/code/localstack/localstack/services/infra.py", line 520, in do_start_infra
    prepare_installation()
  File "/opt/code/localstack/localstack/utils/bootstrap.py", line 77, in wrapped
    return f(*args, **kwargs)
  File "/opt/code/localstack/localstack/services/infra.py", line 474, in prepare_installation
    install.install_components(apis)
  File "/opt/code/localstack/localstack/services/install.py", line 750, in install_components
    parallelize(install_component, names)
  File "/opt/code/localstack/localstack/utils/threads.py", line 135, in parallelize
    return pool.map(func, arr)
  File "/usr/local/lib/python3.10/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/local/lib/python3.10/multiprocessing/pool.py", line 771, in get
    raise self._value
  File "/usr/local/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/lib/python3.10/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/opt/code/localstack/localstack/services/install.py", line 746, in install_component
    installer()
  File "/opt/code/localstack/localstack/services/install.py", line 364, in install_kinesis
    install_kinesis_mock(bin_path)
  File "/opt/code/localstack/localstack/services/install.py", line 427, in install_kinesis_mock
    raise ValueError(
[...]

PR #8790, opened 2023-08-01
upgrade kinesis-mock to scalajs release, add node installer

### LINKED ISSUE #8680: enhancement request: Use Kinesis Mock JS Distribution

### Enhancement description

Kinesis Mock is preparing a switch to a Javascript executable offering, which will remove the old GraalVM executables. This should be much easier to work with from Localstack's perspective, as well as remove the java dependency for users. But I'm unsure as to how to approach this in Localstack's code base, and am looking for some guidance / direction.

The current plan is to have 2 GitHub release assets:
- main.js
- main.js.map

To start the service, you'd simply do `./main.js`, as it should be an executable. You can use `node main.js` as well. 

I also plan to publish this to NPM if that is an easier grab for localstack. 

References:

https://github.com/etspaceman/kinesis-mock/pull/531
https://github.com/etspaceman/kinesis-mock/blob/scalajs/docker/Dockerfile

### 🧑‍💻 Implementation

Update these lines to execute a nodejs command instead:

https://github.com/localstack/localstack/blob/master/localstack/services/kinesis/kinesis_mock_server.py#L101

Upgrade this to the newest version (planning for 0.4.0):

https://github.com/localstack/localstack/blob/master/localstack/services/kinesis/kinesis_mock_server.py#L101

Update this to get the `main.js` file for all platforms:

https://github.com/localstack/localstack/blob/master/localstack/services/kinesis/kinesis_mock_server.py#L101

### Anything else?

_No response_

alexrashed:
This PR upgrades `kinesis-mock` to the new version `0.4.1`, which migrates to ScalaJS (with https://github.com/etspaceman/kinesis-mock/pull/531).
It creates a new `NodePackageInstaller` which handles the installation of `npm` packages in LocalStack.

Thanks to @etspaceman for the heads-up, and the discussions prior to the release!
Fixes #8680

### DIFF
--- a/localstack/packages/core.py
+++ b/localstack/packages/core.py
@@ -2,14 +2,16 @@ import logging
 import os
 from abc import ABC
 from functools import lru_cache
+from typing import Optional
 
 import requests
 
 from localstack import config
 
 from ..utils.archives import download_and_extract
-from ..utils.files import chmod_r, mkdir, rm_rf
+from ..utils.files import chmod_r, chown_r, mkdir, rm_rf
 from ..utils.http import download
+from ..utils.run import is_root, run
 from .api import InstallTarget, PackageException, PackageInstaller
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


etspaceman:
Smart using NPM here. That should help users who suffer from https://github.com/localstack/localstack/issues/6797

localstack/packages/core.py
```diff
@@ -178,3 +180,52 @@ def _get_github_asset_name(self) -> str:
...
+            ]
+        )
+        # npm 9+ does _not_ set the ownership of files anymore if run as root
+        # - https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/
+        # - https://github.com/npm/cli/pull/5704
+        # - https://github.com/localstack/localstack/issues/7620
+        if is_root():
+            # if the package was installed as root, set the ownership manually
+            LOG.debug("Setting ownership root:root on %s", target_dir)
+            chown_r(target_dir, "root")
```
whummer, line 231: nit: presumably the `read` permission flag will be set on the files, but we can briefly do some tests (after the PR is merged) with starting in Docker as non-root user, to ensure this still works 👍

whummer (approved):
Looks awesome @alexrashed ! 🚀 Looking forward to using this new package, which also seems to be a bit more lightweight (>40MB binary vs ~12MB for the npm package).

ScalaJS looks pretty impressive! Technically this could even be run in the browser now :) . Thanks again to @etspaceman for adding this enhancement and bringing it to our attention. 🙌

localstack/packages/core.py
alexrashed: If `npm install` is executed with a user without superuser privileges, it will set the current user as owner (because everything else would enforce superuser priviledges). I also just tested the installation of the package locally with a non-root user and all files are owned by the current user.

PR #8792, opened 2023-08-01
Remove GenericBaseModel.get_cfn_attribute and improve stack failure reporting

dominikschubert:
## Changes
- Remove `GenericBaseModel.get_cfn_attribute` and unify code paths between legacy and new providers
- Improve stack failure reporting a tiny bit

Previously the ResourceStatusReason on the stack failure was empty, now it's set to the string representation of the caught exception. This is currently only enabled when CFN_VERBOSE_ERRORS is set though. We should probably make this the default though in the next minor release to increase parity. 

New stack status:
```
{"LogicalResourceId": "stack-9645eb0a", "ResourceType": "AWS::CloudFormation::Stack", "ResourceStatus": "CREATE_FAILED", "ResourceStatusReason": "Resource type AWS::SSM::Parameter does not support attribute {Invalid}"}
```

We might also want to by start failing stacks by default when outputs can't be resolved but since this differs quite a bit from the current behavior I think we should have a deprecation path for this.

### DIFF
--- a/localstack/services/cloudformation/engine/entities.py
+++ b/localstack/services/cloudformation/engine/entities.py
@@ -150,11 +150,13 @@ class Stack:
             result.setdefault(attr, [])
         return result
 
-    def set_stack_status(self, status):
+    def set_stack_status(self, status: str, status_reason: Optional[str] = None):
         self.metadata["StackStatus"] = status
         if "FAILED" in status:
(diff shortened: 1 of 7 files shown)


pinzon (approved):
Awesome :rocket:

PR #8796, opened 2023-08-01
Make scaffolding of resource provider tests & templates optional

dominikschubert:
Scaffolding of tests & templates in `tests/integration/...` will now require the additional `-t` flag. 
This is aimed at removing some barriers and relaxing our requirements for new resource providers.


thrau (approved):
nice! :100:  (sneaking in a quick review :P)

PR #8798, opened 2023-08-01
Implement resource provider for AWS::SQS::Queue

dominikschubert:
## Motivation

We've previously asked team members to give the new resource providers a try. I've revived and adapted the code from https://github.com/localstack/localstack/pull/8390 to create the first "production" resource provider `AWS::SQS::Queue`.

While this worked in isolation, tests kepts failing when performing updates on stacks that include an `AWS::SQS::Queue` even when not implementing updates at all because of broken state management.

This PR has seen a bit of feature creep but I think it makes sense to bundle at least one new provider with the changes in here.

## Changes
- Adds new `AWS::SQS::Queue` resource provider (adapted from from https://github.com/localstack/localstack/pull/8390) 
- Makes the new `AWS::SQS::Queue` resource provider enabled by default (i.e. it will supersede the old `GenericBaseModel` version)
- Adds two new fields on the "MetaResource" model to explicitly track certain property states
  - `SpecifiedProperties`
  - `_last_deployed_state`

These have a much clearer definition now

- `Properties` ... The currently specified & resolved properties form the template, i.e. without any previous state (=> `desiredState`)
- `SpecifiedProperties` ... The last specified properties. I.e. the `Properties` from the previous resource provider execution.  (=> `previousState`)
Example: 1. create() ... the returned properties will become (?)
- `_state_` ... The last result from a `fetch_state()` call
- `_last_deployed_state` ... snapshot of the merged Properties taken after the last successful resource provider execution. (What was returned in the `ProgressEvent`)

The naming should be changed before merging to reflect their purpose better. In the future we can remove certain parts (`_state_`) and refactor the whole "MetaResource` Model and its state tracking.

## Up next
- Now that we can replace models, I'll start with the old ones having update support (implementing `update_resource`) which will mean we can remove all GenericBaseModel-based updates.
- Create registry with schemas for all resource types so that we can make more schema-driven decisions when determining changes and especially replacement strategies.


localstack/services/cloudformation/resource_provider.py
```diff
@@ -675,11 +697,21 @@ def execute_action(
...
+                try:
+                    return resource_provider.update(request)
+                except NotImplementedError:
+                    LOG.warning(
+                        'Unable to update resource type "%s", id "%s"',
+                        request.resource_type,
+                        request.logical_resource_id,
+                    )
+                    return ProgressEvent(
+                        status=OperationStatus.SUCCESS, resource_model=request.desired_state
```
dominikschubert, line 709: should be previous state

tests/aws/services/cloudformation/api/test_stacks.py
```diff
@@ -228,6 +228,37 @@ def test_update_stack_with_same_template(self, deploy_cfn_template, aws_client):
...
+            max_wait=360,
+        )
+
+        queue_arn_2 = aws_client.sqs.get_queue_attributes(
+            QueueUrl=stack2.outputs["QueueUrl"], AttributeNames=["QueueArn"]
+        )["Attributes"]["QueueArn"]
+        assert queue_arn_2
+
+        assert queue_arn_1 != queue_arn_2
+        print("done")
```
dominikschubert, line 260: clean

simonrw (approved):
Awesome set of changes, excited to get the new providers in production, and I'm glad the updates are working well 🎉 🦖

PR #8799, opened 2023-08-01
implement S3 native bucket versioning

bentsku:
This PR implements Bucket Versioning in the new native S3 provider. This in turn allowed me to properly tests what I had already written but couldn't test because I couldn't actually enable the versioning, and fix what was wrong. 

This PR also contains a new feature in the serializer: we couldn't add headers to exceptions, when S3 actually does that. I'm not sure this is common to all XML protocol based services, so I contained it only in S3 for now. I tried to keep the same existing logic as much as I could, but applied to exceptions.

### DIFF
--- a/tests/aws/s3/test_s3.py
+++ b/tests/aws/s3/test_s3.py
@@ -5982,6 +5982,7 @@ class TestS3PresignedUrl:
         )
 
         # AWS seems to detected what kind of signature is missing from the policy fields
+        print(response.content)
         exception = xmltodict.parse(response.content)
         exception["StatusCode"] = response.status_code
         snapshot.match("exception-missing-signature", exception)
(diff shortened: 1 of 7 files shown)


tests/aws/s3/test_s3.py
```diff
@@ -5982,6 +5982,7 @@ def test_post_request_missing_signature(
         )
 
         # AWS seems to detected what kind of signature is missing from the policy fields
+        print(response.content)
```
bentsku, line 5985: will remove in the next PR if possible, as I wouldn't want to rebase everything. It could be merged right after!

thrau (approved):
LGTM! nice that it's so easy :D i'll let @alexrashed sign off on the serializer change

alexrashed (approved):
Awesome! The latest revision of the serializer changes are great and perfectly generalized! 🚀

PR #8802, opened 2023-08-02
implement S3 native notifications

bentsku:
This PR implement notifications for the new native S3 provider.

Only 2 types of notifications are missing, concerning ACL and Tagging, because those operations are not yet implemented in the new provider but this will come. The most important was to make it compatible with the new provider.


macnev2013 (approved):
LGTM 👍🏻

PR #8814, opened 2023-08-03
implement S3 native CORS

bentsku:
This PR implements CORS for the new S3 native provider. This is basically just to make it compatible between both provider.

### DIFF
--- a/localstack/services/s3/cors.py
+++ b/localstack/services/s3/cors.py
@@ -17,6 +17,7 @@ from localstack.constants import S3_VIRTUAL_HOSTNAME
 from localstack.http import Request, Response
 from localstack.services.s3.models import BucketCorsIndex
 from localstack.services.s3.utils import S3_VIRTUAL_HOSTNAME_REGEX
+from localstack.services.s3.v3.models import BucketCorsIndex as BucketCorsIndexV2
 
 # TODO: add more logging statements
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


localstack/services/s3/cors.py
```diff
@@ -31,10 +32,10 @@
 
 class S3CorsHandler(Handler):
 
-    bucket_cors_index: BucketCorsIndex
+    bucket_cors_index: BucketCorsIndex | BucketCorsIndexV2
 
-    def __init__(self):
-        self.bucket_cors_index = BucketCorsIndex()
+    def __init__(self, bucket_cors_index: BucketCorsIndex | BucketCorsIndexV2 = None):
+        self.bucket_cors_index = bucket_cors_index or BucketCorsIndex()
```
thrau, line 38: This is a minor thing, but since `S3CorsHandler` seems independent from the provider, maybe we could further remove the dependency here, instead of introducing a dependency to _both_ providers.
Suggestion (note: remove the call to `BucketCorsIndex()`):

introduce a Protocol here:
```python
class BucketCorsIndex(Protocol):

    @property
    def cors(self) -> dict[str, CORSConfiguration]:
        raise NotImplementedError

    @property
    def buckets(self) -> set[str]:
        raise NotImplementedError


class S3CorsHandler(Handler):
    bucket_cors_index: BucketCorsIndex

    def __init__(self, bucket_cors_index: BucketCorsIndex):
        self.bucket_cors_index = bucket_cors_index
        ...
```

then, make sure to instantiate the provider-specific BucketCorsIndex in the provier: https://github.com/localstack/localstack/blob/a7e97ea120cd37e90ce0722b241d1b7d82f34582/localstack/services/s3/provider.py#L249

localstack/services/s3/validation.py
thrau, line 1: nice :100:

thrau (approved):
Looks really good! I like the refactoring around the validators.

I have one minor suggestion but nothing blocking

PR #8815, opened 2023-08-03
fix s3_empty_bucket fixture to delete DeleteMarkers

bentsku:
Just realised I had 101 buckets in my AWS account after doing some S3 testing, and seeing `TooManyBuckets` rang a bell somewhere 😄🚨 

You need to even delete `DeleteMarkers`, otherwise your bucket is not considered empty. It's weird, but that's how it is 🤷‍♂️
I tried the fixture against my 101 buckets and it worked, cleaned them up and deleted them.


viren-nadkarni (approved):
Thanks for fixing this!

PR #8817, opened 2023-08-03
Small fixes to support CLI use with Python 3.8

whummer:
Tiny fixes to support CLI use with Python 3.8. Discovered this while working on a support case of a customer who attempted to install the AWS replicator/proxy extension.

Can be replicated when installing Python 3.8 via `pyenv`, and then:
```
$ DEBUG_PLUGINS=1 localstack -d extensions install localstack-extension-aws-replicator
DEBUG:stevedore._cache:reading /Users/whummer/Library/Caches/Python Entry Points/5e2a6eb1bdd93e1d7b62ee754448916f77349878c2468e3c113c4965f6f6e4d7
DEBUG:stevedore.extension:found extension EntryPoint(name='ext-all', value='localstack_ext.cli.localstack:ExtCliPlugin', group='localstack.plugins.cli')
ERROR:plugin.manager:error importing entrypoint EntryPoint(name='ext-all', value='localstack_ext.cli.localstack:ExtCliPlugin', group='localstack.plugins.cli'): 'type' object is not subscriptable
DEBUG:stevedore.extension:found extension EntryPoint(name='pro', value='localstack_ext.cli.localstack:ProLoggedInCliPlugin', group='localstack.plugins.cli')
ERROR:plugin.manager:error importing entrypoint EntryPoint(name='pro', value='localstack_ext.cli.localstack:ProLoggedInCliPlugin', group='localstack.plugins.cli'): 'type' object is not subscriptable
```

Added a simple test to assert that the module can get imported, which should be picked up by our `cli-tests` GH Action (which is parametrized for different Python versions). Over time we could consider adding an extensions CLI test for different Python versions (but probably not really high prio).

### DIFF
--- a/localstack/utils/venv.py
+++ b/localstack/utils/venv.py
@@ -74,9 +74,9 @@ class VirtualEnvironment:
         return matches[0]
 
     def inject_to_sys_path(self):
-        if path := str(self.site_dir):
-            if path not in sys.path:
-                sys.path.append(path)
+        path = str(self.site_dir)
+        if path and path not in sys.path:
+            sys.path.append(path)
 
     def add_pth(self, name, path: Union[str, os.PathLike, "VirtualEnvironment"]) -> None:
         """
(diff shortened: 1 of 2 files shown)


thrau (approved):
thanks for adding this :pray: and sorry, totally an oversight from me.

will add some extensions CLI tests soonTM

PR #8818, opened 2023-08-03
configure MAIN_DOCKER_NETWORK automatically if --network flag is used

thrau:
Something I forgot to add with #8772, was to also configure the `MAIN_DOCKER_NETWORK` variable automatically.


simonrw (approved):
Good catch!

PR #8819, opened 2023-08-03
fix S3 streaming usage of readwrite lock

bentsku:
There has been so CI failures for the streaming provider lately regarding releasing of un-acquired locks, but all of them were used with context managers, which means they had to be acquired at some point?

I think the[ documentation of the library](https://github.com/elarivie/pyReaderWriterLock) was a bit misleading, having to generate a lock every time you need one seems quite weird, but it seems it's how the library is supposed to be used and that's why it was failing, my mistake.


thrau (approved):
right, you need to call gen_wlock every time! totally missed that in the first review. LGTM!

PR #8820, opened 2023-08-03
fix credentials handling for capture-not-implemented

alexrashed:
This PR explicitly sets credentials in the capture-not-implemented script to avoid any issues if the credentials are not implicitly set via environment variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).
In CircleCI, our test credentials are set globally, but executed in different environments, this caused issues.


steffyP (approved):
thanks for jumping on this so quickly, @alexrashed 🙏

PR #8821, opened 2023-08-04
[SFN] Fix execution input, fix handling lambda payload i/o

MEPalma:
Removed redundant lambda payload cleanups that caused failures whenever payloads were not json dictionaries.
Adds relevant tests.
Fixes deprecated events initiation call in Execution objects.

### DIFF
--- a/tests/integration/stepfunctions/templates/services/lambdafunctions/return_bytes_str.py
+++ b/tests/integration/stepfunctions/templates/services/lambdafunctions/return_bytes_str.py
@@ -1,0 +1,2 @@
+def handler(event, context):
+    return b'"HelloWorld!"'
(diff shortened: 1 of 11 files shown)


tests/integration/stepfunctions/templates/services/lambdafunctions/return_bytes_str.py
```diff
@@ -0,0 +1,2 @@
+def handler(event, context):
+    return b'"HelloWorld!"'
```
dominikschubert, line 2: Maybe we should also test with different values here (list, boolean, integer, float, string, bytes, dict)

dominikschubert (approved):
:rocket: good catch

tests/integration/stepfunctions/templates/services/lambdafunctions/return_bytes_str.py
MEPalma: Absolutely, thank you:) Given this i added more fixes to this pr about default values for Lambda Task Payloads and start_execution inputs. I added these test scenarios without declaring new lambda functions as other json primitives can be passed to sfn as json values (not bytes).

PR #8822, opened 2023-08-04
Cfn: Multi-accounts compatibility

viren-nadkarni:
This PR adds account and region awareness to Cloudformation.

It will now be possible to deploy Cfn stack in any arbitrary account ID and any region.

Specifically, this PR makes the following changes:
- Take the account ID and region name from Request Context and explicitly pass it down the call stack
- Use these values when instantiating `connect_to()` clients and doing internal cross-service calls
- Remove use of `aws_stack.get_region()` and `get_account_id()`


dominikschubert (approved):
LGTM, thanks for adding this :+1:

PR #8823, opened 2023-08-04
Bundle lambda init into docker image

dominikschubert:
Certain users seem to have issues fetching the lambda init from GitHub which made us notice that we're not actually bundling it into the built Docker container. 

It's probably sensible to ship the lambda runtime init by default


dfangl (approved):

alexrashed (approved):
As discussed in https://github.com/localstack/localstack/pull/8194, we want to ship the packages for our top used services. Lambda definitely is one of them. Thanks for jumping on this! 🚀

PR #8826, opened 2023-08-04
[SFN] Parse fail on incorrect derivations, fix mulitple retry, fix language definitions

MEPalma:
- Raises InvalidDefinition errors when an incorrect/unsupported ASL derivation is provided on create_state_machine calls
- enables relevant test
- Adds fixes to the parser including to retry declaration for which multiple retry conditions could not be parsed, and other minors
- Adds multiple retry test to fix the parser's logic


dominikschubert (approved):
Great changes :+1:

PR #8829, opened 2023-08-04
[SFN] Enhanced support for empty fail states

MEPalma:
Enhances the support for empty fail states (where no error nor cause is defined), so that these can be handled by catch/retry statements. Adds relevant tests too.


dominikschubert (approved):
LGTM

tests/aws/stepfunctions/v2/scenarios/test_base_scenarios.py
```diff
@@ -177,6 +177,26 @@ def test_map_state_catch(
             exec_input,
         )
 
+    def test_map_state_catch_empty_fail(
```
dominikschubert, line 180: small nit: all tests in `tests/aws/*` will now need to have one of the `@markers.aws.*`

PR #8835, opened 2023-08-06
add WebSocket support for ASGI/WSGI bridge and Gateway

thrau:
## Motivation

This PR extends the http framework to let us serve and proxy websockets through the gateway.

Serving websockets through port `:4566` has been on the radar ever since we introduced the ASGI/WSGI bridge. This can replace our current pattern (for example in APIGWv2) where we need to spin up an additional server process and expose it on the external service port range. Moreover, the [need to proxy websockets](https://github.com/localstack/localstack-extensions/tree/main/mailhog#known-limitations) recently appeared when implementing our mailhog extension. @bentsku also pointed out that [ALB supports websocket listeners](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html) to proxy websockets to ECS tasks for instance.

I had some time on my hands on the rainy weekend, so I took a stab at it for fun.

## Changes

* Since there is no native WSGI concept for websockets or anything to map it onto, I had to add three interfaces to `localstack.http.asgi`. This layer is not exposed to the developer.
  - `ASGIWebSocket`: an interface which allows you to communicate with the underlying ASGI websocket IO objects in a synchronous way to deal with [ASGI websocket events](https://asgi.readthedocs.io/en/latest/specs/www.html#websocket). 
  - `WebSocketEnvironment`: a pseudo-WSGI environment with an additional key `asgi.websocket` that holds the `ASGIWebSocket` object (analogous to how the WSGIEnvironment holds the input stream)
  - `WebSocketListener`: handed to the bridge to call on incoming Websocket connections. It's a callable that accepts `accept(WebSocketEnvironment)`
* Introduced a pseudo Werkzeug Request object `WebSocketRequest` that inherits from a sansio Werkzeug request, wraps the lower level `ASGIWebsocket`, and provides a developer-friendly API to interact with websockets. I say pseudo because, while you can pass the request to *most* handlers in the handler chain, it technically doesn't quite match the contract of `werkzeug.Request`, since there is no meaningful way to resolve `request.data`. This is, however, [identical to how Quart](https://github.com/pallets/quart/blob/206d93f74846fd697650bf85ef1de36c86716d48/src/quart/wrappers/base.py#L16-L17) deals with this.
* To allow us to pass `WebSocketRequest`s to the handler chain, I had to make some slight changes that guard against `request.data` calls, without doing `isinstance` checks. To that end, I re-used Werkzeug's [`Request.shallow`](https://github.com/pallets/werkzeug/blob/41b74e7a92685bc18b6a472bd10524bba20cb4a2/src/werkzeug/wrappers/request.py#L48-L51) flag, and added checks for that in a couple of places. (side note: passing a shallow Request through the handler chain after each test run could help us identify handlers which mistakenly call `request.data`)

## Discussion

Probably the most controversial part of the PR is whether we should really pass `WebSocketRequest`s through the handler chain by default: https://github.com/localstack/localstack/blob/ddbe3712e2cbe23119aed0a27dcf1f3cdff0e39a/localstack/aws/gateway.py#L40-L43

Or whether it should be a completely separate control path.

The nice thing is that *most* handlers will actually work since they only operate on the URL and headers. Which is great because we can re-use the router, CORS enforcer, request logger, etc. However, if there's a handler in the request-preprocessor that accesses `request.data`, that will block any WebsocketRequests from working, so they would all have to do a `if request.shallow` check from now on. Also, the type checker will complain since a `WebsocketRequest` does not technically match the `werkzeug.Request` contract fully.

I did play around with having a separate "WebsocketHandlerChain" or "WebsocketRouter", which create a completely separate control flow for websockets to avoid the contract issue. It didn't seem very productive. I think it's better to look into maybe improving the handler, handlerchain, and request context to better handle the types and use `is request.scheme in ["ws", "wss"]` or similar checks at runtime.

### Routes:

There's a new method called `WEBSOCKET`, which can be used to register routes that deal with websocket requests:

```python
@route("/echo/<name>", methods=["WEBSOCKET"])
def echo_handler(request: WebSocketRequest, name: str):
    with request.accept() as websocket:
        websocket.send(f"thanks for connecting {name}")
        for line in websocket:
            websocket.send(f"echo: {line}")
            if line == "exit":
                websocket.send("ok bye!")
                return
```

This actually works when you drop it into the edge router :zany_face:

```python
@hooks.on_infra_start()
def add_websocket_demo_handler():
    from localstack.services.edge import ROUTER

    ROUTER.add(echo_handler)
```

### Proxy:

you can now write application layer websocket proxies like this (just a PoC, tested with our mailhog extension):

```python
def websocket_proxy(forward_base_url: str):
    def _proxy(request: WebSocketRequest, methods=["WEBSOCKET"], *args, **kwargs):
        url = f"{forward_base_url.rstrip('/')}{request.path}"

        backend = websocket.WebSocket()

        # TODO: forward headers, react to 403 from backend and such
        backend.connect(url)
        ws = request.accept()

        def _recv_thread():
            # receive from backend connection and forward to client
            try:
                for _data in backend:
                    ws.send(_data)
            except websocket.WebSocketConnectionClosedException:
                pass

        t = threading.Thread(target=_recv_thread)
        t.start()

        # receive from client connection and forward to backend
        for line in ws:
            backend.send(line)

        ws.close()
        backend.close()

    return _proxy
```

now just add the proxy to the router (needs to be added as separate route because `ProxyHandler` cannot deal with websockets - could look into that though so it handles websockets transparently).
```python
    ROUTER.add(
        "/api/v2/websocket",
        host=f"{hostname_prefix}<host>",
        endpoint=websocket_proxy("ws://localhost:8025/"),
    )
```

## Testing

The example is in the demo commit ddbe3712, so you can just checkout the branch, run `make entrypoints`, and start in host mode to try it out.

```console
python -m localstack.runtime.main
```

```console
npm install -g wscat
wscat -c localhost:4566/echo/localstack
```

**update**: i removed the demo commit now, and there are a bunch of unit and integration tests

## TODO

- [x] clarify whether we accept the slight contract mismatch between `WebsocketRequest` and the handler chain to get maximum reusability of our handlers
- [x] extend `WebsocketRequest` with useful features
- [x] tests
- [x] remove the demo commit ddbe3712


thrau:
@alexrashed: maybe you could comment on the overall approach first, and then i'll add tests and open the PR for review

cc @whummer @calvernaz @bentsku

localstack/aws/handlers/routes.py
```diff
@@ -29,4 +29,4 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
             chain.stop()
         except NotFound:
             if self.respond_not_found:
-                chain.respond(404)
+                chain.respond(404, "not found")
```
thrau, line 32: actually 404 requests for WebsocketRequests need to be dealt with separately, since the default implementation of `ROUTER` actually has `respond_not_found=False`, and falls back to the EmptyResponseHandler.

localstack/http/request.py
```diff
@@ -278,6 +278,9 @@ def restore_payload(request: Request) -> bytes:
       inevitably become the the basis of proxying werkzeug requests. The alternative is to build our own request object
       that memoizes the original payload before parsing.
     """
+    if request.shallow:
+        return b""
```
thrau, line 282: perhaps it would be better to raise an exception here instead of returning empty bytes, but this "just works" (it's mostly to fix the request logger)

alexrashed:
FYI: I just had a discussion on this topic yesterday with @thrau.
While it feels a bit weird to directly integrate the websocket handling into the Werkzeug handling, there are lots of reasons why this makes sense (the initial HTTP connection upgrade, the possibility to return HTTP responses before the connection is upgraded, the re-usability of existing handlers for these HTTP requests and responses - like CORS).
Anyway, `WebsocketRequest` will definitely inherit from the `Request` class. The other architectural decisions, most importantly the question if the handler chains should be separated, can pretty easily be changed afterwards.

This is a pretty cool change, that will enable quite some new features for services emulated in LocalStack, but also in LocalStack extensions! Also, it's really great to see that this change also only needed a few lines of code! 🚀

localstack/aws/gateway.py
```diff
@@ -34,3 +35,15 @@ def process(self, request: Request, response: Response):
...
+        response = Response(status=101)
+        self.process(request, response)
+
+        # only send the populated response if the websocket hasn't already done so before
+        if response.status_code != 101:
+            if request.is_upgraded():
+                return
+            if request.is_rejected():
+                return
+            request.reject(response)
```
alexrashed, line 49: nit: So this basically covers the case when a routed endpoint does not call accept on the request but sets a response, right?
Maybe a small test for this would be great as well, just to make sure that this control flow is covered.

localstack/http/websocket.py
```diff
@@ -0,0 +1,326 @@
...
+            self.socket.send(
+                {
+                    "type": "websocket.send",
+                    "bytes": text_or_bytes,
+                    "text": None,
+                },
+                timeout=timeout,
+            )
+
+    def receive(self) -> str:
```
alexrashed, line 89: q: The receive method here always returns a string. How would I implement a binary communication on the websocket (like some kind of compressed protocol)?

localstack/http/websocket.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        event = self.socket.receive()
+        if event["type"] == "websocket.receive":
+            text = event.get("text")
+            if text is not None:
+                return text
+
+            buf = event.get("bytes")
+            if buf is not None:
+                return buf.decode(self.encoding)
```
alexrashed, line 106: nit: I think this would break if the sent bytes are not UTF-8 encoded / aren't encoded text (but this function clearly expects a string being sent, so maybe this is on purpose?).

alexrashed (approved):
Awesome! A really great set of changes, well written, documented, and tested! 🚀 💯 
I only found one thing that might need a revision: Sending binary data (not only text).
Since this might have an impact on the `Websocket.receive` interface, you might want to tackle this in the first iteration already. But we could also tackle this along the way, once we have a use-case for it...

localstack/http/websocket.py
thrau: good point, i haven't really tested binary mode at all, and the API doesn't support it well as you pointed out. will test this explicitly and change the api to return either `str|bytes` depending on what mode the socket is in.

localstack/http/websocket.py
thrau: i can see that ASGI handles binary/text mode per-frame automatically based on a flag set in the frame. this is also the way client libraries do this. this means that `receive` can basically return `str | bytes`, which is what i implemented now. there's also a test for this

Issue #8739, opened 2023-07-23
bug: Several fields in ApiGateway Stage unsettable or unmodifiable


tor-gu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following fields in an api-gateway stage appear to be unsettable or unmodifialble:

- `description`: not modifiable
- `documentation_version`: not settable
- `variables`: not modifiable
- `tags`: not modifiable
- `access_log_settings.*`: not settable

For example, I cannot modify the `description` from `"description_foo"` to `"description_bar"`:
```bash
$ awslocal apigateway update-stage \
	--rest-api-id $(tflocal output -raw rest_api_id) \
	--stage-name example_stage \
	--patch-operations op=replace,path=/description,value=description_bar \
	| jq '.description'
"description_foo"
```

### Expected Behavior

```bash
$ awslocal apigateway update-stage \
	--rest-api-id $(tflocal output -raw rest_api_id) \
	--stage-name example_stage \
	--patch-operations op=replace,path=/description,value=description_bar \
	| jq '.description'
"description_bar"
```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

I'm using terraform, and, similar to #8642 , I noticed it when it appeared that `tflocal apply` was not idempotent.  

The full terraform file, `main.tf`, for this example is included at the end.

Applying `main.tf` repeatedly appears to show the `documentation_version` repeatedly being set:
```bash
$ tflocal init
$ tflocal apply --auto-approve
# ... output removed ...
$ tflocal apply --auto-approve
# ... some output removed ...
  # aws_api_gateway_stage.example will be updated in-place
  ~ resource "aws_api_gateway_stage" "example" {
      + documentation_version = "version_foo"
        id                    = "ags-16jh3gkepb-example_stage"
        tags                  = {
            "tag1" = "tagval_foo"
        }
        # (11 unchanged attributes hidden)
    }
```
Indeed, the `documentation_version` does not appear to have been set at all:
```bash
$ awslocal apigateway get-stage \
        --rest-api-id $(tflocal output -raw rest_api_id) \
        --stage-name example_stage 
{
    "deploymentId": "032tdew16k",
    "stageName": "example_stage",
    "description": "description_foo",
[...]

PR #8836, opened 2023-08-06
Enhance parity around API Gateway stage patch operations

whummer:
Enhance parity around API Gateway stage patch operations, add implementation for API GW documentation versions. Addresses https://github.com/localstack/localstack/issues/8739

The PR also adds some refactoring to the existing tests in `test_apigateway_basic.py` - grouping related tests into test classes, using proper fixtures for resource creation, getting rid of global constants to make tests more self-contained, etc. Still more work needs to be done to get them on par with our other integration tests - introduce more fixtures, add snapshots, etc.

Summary of changes:
* implement the `update_stage` method in the provider, and perform corresponding patch operations
* adds small changes to fix missing/invalid attributes in the responses for the `create_stage`, `get_stage`, `update_stage` methods
* add implementation for documentation version API methods
* add a snapshot test to validate the behavior of patch operations in stage updates
* add some negative tests for attributes (e.g., `documentation_version`, `tags`)
* refactor some of the tests in `test_apigateway_basic.py` - remove global static constants, add proper fixtures, etc
* left a `TODO` comment in the code to add `createdDate` / `lastUpdatedDate` in Stage operations (to be tackled in a future PR)

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -131,14 +131,14 @@ class OpenAPIExt:
 # TODO: make the CRUD operations in this file generic for the different model types (authorizes, validators, ...)
 
 
-def get_apigateway_store(
-    account_id: str = None, region: str = None, context: RequestContext = None
-) -> ApiGatewayStore:
-    if not account_id:
-        account_id = context.account_id if context else get_aws_account_id()
-    if not region:
-        region = context.region if context else aws_stack.get_region()
-    return apigateway_stores[account_id][region]
+def get_apigateway_store(context: RequestContext) -> ApiGatewayStore:
+    return apigateway_stores[context.account_id][context.region]
+
+
+def get_apigateway_store_for_invocation(context: ApiInvocationContext) -> ApiGatewayStore:
+    account_id = context.account_id or get_aws_account_id()
+    region_name = context.region_name or aws_stack.get_region()
+    return apigateway_stores[account_id][region_name]
 
 
 class ApiGatewayIntegrationError(Exception):
@@ -809,10 +809,11 @@ def add_documentation_parts(rest_api_container, documentation):
 
 
 def import_api_from_openapi_spec(
-    rest_api: RestAPI, body: Dict, query_params: Dict, account_id: str = None, region: str = None
+    rest_api: RestAPI, body: dict, context: RequestContext
 ) -> Optional[RestAPI]:
     """Import an API from an OpenAPI spec document"""
 
+    query_params: dict = context.request.values.to_dict()
     resolved_schema = resolve_references(copy.deepcopy(body), rest_api_id=rest_api.id)
 
     # TODO:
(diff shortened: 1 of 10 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -1985,12 +2130,18 @@ def create_custom_context(
...
-    if not isinstance(entity.__dict__, DelSafeDict):
-        entity.__dict__ = DelSafeDict(entity.__dict__)
+    patch_operations = patch_operations or []
+
+    if isinstance(entity, dict):
+        entity_dict = entity
+    else:
+        if not isinstance(entity.__dict__, DelSafeDict):
+            entity.__dict__ = DelSafeDict(entity.__dict__)
+        entity_dict = entity.__dict__
```
bentsku, line 2140: nice changes 👍

tests/integration/apigateway/conftest.py
```diff
@@ -168,3 +168,9 @@ def _import_apigateway_function(*args, **kwargs):
 
     for rest_api_id in rest_api_ids:
         delete_rest_api(aws_client.apigateway, restApiId=rest_api_id)
+
+
+@pytest.fixture
+def apigw_add_transformers(snapshot):
+    snapshot.add_transformer(snapshot.transform.key_value("id"))
```
bentsku, line 175: just a note, beware when using a straight `key_value` transformer with `id`, as multiple different kind of resources share this key, which can have weird effect in the snapshots... I've been bitten multiple times and I tend to use `json_path` now for those, to not mix up. When snapshotting only one kind of resource, it does make perfect sense though! 👍

bentsku (approved):
LGTM! Wow, that's a big push for API Gateway tests, and thanks for fixing Stages, and adding a new resource in the meantime! That's awesome really. `test_apigateway_basic.py` is quite the challenge, and migrating those to be aws validated will take some time. Thanks a lot for giving it structure! Also, thanks again for fixing all the store access.

localstack/services/apigateway/helpers.py
```diff
@@ -130,8 +131,14 @@ class OpenAPIExt:
 # TODO: make the CRUD operations in this file generic for the different model types (authorizes, validators, ...)
 
 
-def get_apigateway_store(account_id: str = None, region: str = None) -> ApiGatewayStore:
-    return apigateway_stores[account_id or get_aws_account_id()][region or aws_stack.get_region()]
+def get_apigateway_store(
```
calvernaz, line 134: another option would be to have two method overloading one with the existing interface and another with the context. That because if the only thing we care is about those two fields it's a bit redundant and we can eventually make the request context (RequestContext) present in the `ApigatewayContext` class so can be used everywhere and not only on the provider where it originates, and migrate towards that signature `get_apigateway_store(context: RequestContext)`

tests/integration/apigateway/conftest.py
calvernaz: Agreed, and sometimes it doesn't catches the eye especially on initial tests.

calvernaz (approved):
Thanks for tackling this, another bit covered :smiley_cat:

tests/integration/apigateway/conftest.py
whummer: Great catch, 👍  thanks for the pointer! Updated..

localstack/services/apigateway/helpers.py
whummer: Good point, agreed. Updated the PR and unified the store access a bit further. Turns out we only need two variants of get_store - one for `RequestContext`, one for `ApiInvocationContext`. I think this makes it more clear and allows us to converge towards cleaner signatures - thanks for the pointer! 👍

PR #8839, opened 2023-08-07
Move aws client fixtures to conftest.py

dominikschubert:
Fixes resolving of fixtures / autocompletion for PyCharm. When we moved them to `fixtures.py` PyCharm wasn't able to properly resolve the fixtures for some reason which has a simple workaround:  moving them to `conftest.py`.


viren-nadkarni (approved):
Thanks for taking this over!

PR #8844, opened 2023-08-07
fix issue with sending event from lambda to eventbus

steffyP:
## Motivation
Discovered a parity issue with events while working on a scenario test for loan broker:  the output of a lambda (`on-success`)  should be send to an EventBridge, and will be written to a `sqs` queue, which finally triggers another lambda. 

The `send_event_to_target` was triggered by lambda, but was not correctly forwarding the attributes, resulting in an empty body. 

## Changes
* Fixed the attribute forwarding in `send_event_to_target` for target `events`.
* Also fixed the json-dump for target `sqs` where the event/body should not contain any additional whitespace according to the parity test.

## Testing
Added snapshot test `test_invoke_lambda_eventbridge` for the mentioned scenario using new CDK setup.


tests/aws/awslambda/test_lambda_destinations.py
```diff
@@ -475,5 +493,135 @@ def _assert_event_count(count: int):
...
+    TRIGGERED_LAMBDA_CODE = """
+import json
+
+def handler(event, context):
+    print(json.dumps(event))
+    return {"invocation": True}
+"""
+
+    @pytest.fixture(scope="class", autouse=True)
+    def infrastructure(self, aws_client):
```
dominikschubert, line 520: yess :love_you_gesture:

tests/aws/awslambda/test_lambda_destinations.py
```diff
@@ -475,5 +493,135 @@ def _assert_event_count(count: int):
 #     ...  # TODO
 #
 #
-# class TestLambdaDestinationEventbridge:
-#     ...  # TODO
+class TestLambdaDestinationEventbridge:
+    EVENT_BRIDGE_STACK = "EventbridgeStack"
+    INPUT_FUNCTION_NAME = "InputFunc"
+    TRIGGERED_FUNCTION_NAME = "TriggeredFunc"
+    TEST_QUEUE_NAME = "TestQueueName"
```
dominikschubert, line 500: nit: I'd suggest using the autogenerated names here instead to avoid any conflicts

dominikschubert (approved):
LGTM! Thanks for tackling this. Love the usage of the new cdk-based provisioning test setup :+1:

tests/aws/awslambda/test_lambda_destinations.py
steffyP: @dominikschubert: you mean instead of the function-names defined below, eg. this part, right? 
```
  input_fn_name = f"input-fn-{short_uid()}"
  triggered_fn_name = f"triggered-fn-{short_uid()}"
```
The names defined here `TEST_QUEUE_NAME` etc are just used as identifier for the output variables.

joe4dev (approved):
LGTM 👍 

Thank you for adding the parity test 👏 🚀  Very helpful for our current async invocation loop rework 👍

PR #8845, opened 2023-08-07
fix VirtualHost proxy removing ContentLength for GetObject

bentsku:
## Motivation
It appears AWS does not support `Transfer-Encoding: chunked` responses for `GetObject`. We always set the `Content-Length` when returning from `GetObject`, but when passing the request through the Virtual Host proxy, we would remove it every time. 

## Changes
Added a check for the `Transfer-Encoding` header, so that we now conditionally remove it. Also added a unit test as well as an AWS validated integration tests for S3.

Also sneaked in a change, updating an already skipped path for `tests.aws.s3.test_s3_notifications_lambda.TestS3NotificationsToLambda.test_create_object_by_presigned_request_via_dynamodb` which seems to fail sporadically in CI. 

_fixes #8841_

### DIFF
--- a/tests/unit/http_/test_proxy.py
+++ b/tests/unit/http_/test_proxy.py
@@ -7,7 +7,7 @@ from pytest_httpserver import HTTPServer
 from werkzeug import Request as WerkzeugRequest
 
 from localstack.http import Request, Response, Router
-from localstack.http.client import SimpleRequestsClient
+from localstack.http.client import SimpleRequestsClient, SimpleStreamingRequestsClient
 from localstack.http.dispatcher import handler_dispatcher
 from localstack.http.hypercorn import HypercornServer
 from localstack.http.proxy import Proxy, ProxyHandler, forward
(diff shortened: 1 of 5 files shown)


tests/unit/http_/test_proxy.py
```diff
@@ -196,6 +196,37 @@ def test_proxy_with_custom_client(
...
+            request = Request(path="/", method="GET", headers={"Host": "127.0.0.1:80"})
+
+            response = proxy.request(request)
+
+            if chunked:
+                assert response.headers["Transfer-Encoding"] == "chunked"
+                assert "Content-Length" not in response.headers
+            else:
+                assert response.headers["Content-Length"] == str(len(body))
+                assert "Transfer-Encoding" not in response.headers
```
thrau, line 228: this is definitely the correct behavior :ok_hand:

thrau (approved):
this seems safe to do :ok_hand:

even though i'm not sure the http client is really the best place to do it. it seems we should rather correct this in the S3 serializer? anway, OK for now!

bentsku:
> even though i'm not sure the http client is really the best place to do it. it seems we should rather correct this in the S3 serializer? anway, OK for now!

I don't think I can do that in the serializer, that's the issue, we're inside a route here. The serializer is properly doing its job, but by passing through the proxy, I removed what the serializer had done. I think this is good until it won't 😅

PR #8846, opened 2023-08-07
Wait for open/elasticsearch domain completion

simonrw:
## Motivation

When performing `GetAtt` on an opensearch domain for a property that is not present until the domain is marked as "ready" caused a deploy error.

## Changes

* wait for opensearch/elasticsearch domain completion
* support returning `DomainEndpoint` from an opensearch/elasticsearch domain when complete without failing the stack
* correctly return the output from the `DomainStatus` field from `fetch_state` since this is the contents of the response to be merged with `properties`

### DIFF
--- a/localstack/services/cloudformation/models/elasticsearch.py
+++ b/localstack/services/cloudformation/models/elasticsearch.py
@@ -21,7 +21,13 @@ class ElasticsearchDomain(GenericBaseModel):
 
     def fetch_state(self, stack_name, resources):
         domain_name = self.props["DomainName"]
-        return connect_to().es.describe_elasticsearch_domain(DomainName=domain_name)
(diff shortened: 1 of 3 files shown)


tests/integration/templates/opensearch_domain.yml
```diff
@@ -13,3 +13,7 @@ Resources:
         InstanceType: 'm3.medium.search'
         # test using string types instead of int
         InstanceCount: "2"
+
+Outputs:
+  DomainEndpoint:
+    Value: !GetAtt Domain.DomainEndpoint
```
simonrw, line 19: Note: this block is not required in the equivalent elasticsearch test since that test works successfully _and_ references the domain endpoint, though I don't yet know why.

PR #8848, opened 2023-08-08
Use sane default for Docker stop timeout

viren-nadkarni:
## Motivation

When terminating Docker containers, the cleanup code (if any, running within the container) does not have enough time to finish running. Since the Docker abstraction uses a timeout of 0 seconds, a `SIGKILL` is issued immediately after SIGTERM.

## Changes

This PR changes the default timeout for the stop container operation to 10 seconds. This is same as the Docker library: https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.stop

Ext pipeline is green (excluding known failing tests) :heavy_check_mark: See Ext Integration Tests # 1746


dfangl (approved):
Looks good to me! I kind of miss the constant to change it in one place, but I guess we won't change it for a long time now.

PR #8850, opened 2023-08-08
migrate to new @markers.aws.* markers

dominikschubert:
## Motivation

We've recently introduced new pytest markers in https://github.com/localstack/localstack/pull/8767 and we want to now move towards using these new markers.

## Changes

Refactors all `aws_validated` and `only_localstack` markers to `@markers.aws.*`

## Future work

A follow-up PR will add `@markers.aws.unknown` markers to all unmarked tests.


tests/aws/test_network_configuration.py
```diff
@@ -20,7 +20,7 @@
 from localstack.utils.files import new_tmp_file, save_file
 from localstack.utils.strings import short_uid
 
-pytestmark = [markers.parity.only_localstack]
+pytestmark = [markers.aws.only_localstack]
```
steffyP, line 23: i think for the next iteration we should try to get rid of this module-level markers as well :)

tests/aws/test_ses.py
```diff
@@ -422,8 +422,8 @@ def test_ses_sns_topic_integration_send_email(
         messages.sort(key=sort_mail_sqs_messages)
         snapshot.match("messages", messages)
 
-    @markers.parity.only_localstack
-    @markers.parity.aws_validated
+    @markers.aws.only_localstack
```
steffyP, line 425: that's interesting: it seems to be both: `only_localstack` + `aws.validated` 😅 
as it's a snapshot test, i guess the `only_localstack` can be removed some time :)

steffyP (approved):
Nice, the first new markers in action 😄

tests/aws/test_ses.py
dominikschubert: yeah I'll soon be enforcing that a test can only ave one of the `@markers.aws.*` markers. Basically any test in `tests/aws/` has to have *exactly* **one** marker.

PR #8856, opened 2023-08-08
fix evaluation of not-implemented for stepfunction

steffyP:
## Motivation
A reported issue with `stepfunctions` revealed that our "not-implemented" evaluation for `stepfunctions` is not correct: 
Currently we assume that `500` is only returned when the operation is implemented. However, for stepfunctions we receive the `500` response directly from the stepfunctions-lib, together with the error code `UnsupportedOperation`. 
To reflect that this operation is not implemented, this PR adapts the scripts for the "capture-not-implemented" collection. 

## Changes
For `stepfunctions` set the status_code to `501` (our internal `Not implemented` code), when we get a return code `500 UnsupportedOperation`


dominikschubert (approved):
:rocket:

PR #8857, opened 2023-08-08
Allow separate specification of exposed ports in docker client

dfangl:
## Motivation
To avoid issues with the `--expose` flag usage, we need a proper field in the container client / client configuration to support this field.

## Changes
* Add `exposed_ports` parameter to docker client run/start methods
* Add `exposed_ports` field to ContainerConfiguration dataclass and use it
* Add test for exposed ports to make sure they do not actually publish anything.


alexrashed (approved):
Awesome! Thanks for jumping on this! 🚀

PR #8859, opened 2023-08-08
add validated marker to snapshotted tests

dominikschubert:
## Motivation

We currently assume any tests that use snapshots have been validated as a fairly simply heuristic to find validated tests that haven't been properly marked in the past.

## Changes

Any test with a `snapshot` in its arguments and a `@markers.aws.unknown` is now marked as `@markers.aws.validated`


steffyP (approved):
nice work 👍

PR #8861, opened 2023-08-08
Standardize lambda package structure

joe4dev:
## Motivation

Now that we moved many tests into `aws` in https://github.com/localstack/localstack/pull/8828 :boom: , it would be a good chance to standardize package naming within LocalStack: `awslambda` => `lambda_`

* We already use `localstack/aws/api/lambda_` in the API specs and should be consistent in other parts of the codebase
* Thomas mentioned that a trailing underscore is a Pythonic way to resolve collisions with reserved keywords. PEP 8 mentions that (in a similar context) here: https://peps.python.org/pep-0008/#designing-for-inheritance
* `lambda_` is easier to find in alphabetic ordering because we don't expect to look under awslambda. Anca also mentioned this in last week's demo.

## Why not?

- another test renaming (but better bundle them and do all in once).
- ~The release dance breaks the pipeline until the synchronized LS and ext changes are out~
  - It just blocks the Pro build until we the changes in there as well
- @giograno Does a change in the model package name break existing Lambda CloudPods?

## Changes

Consistently refactor `awslambda` into `lambda_`, especially:

* ❗ The internal boto client changes from `client.awslambda` to `client.lambda_`
* `localstack/services/awslambda` => `localstack/services/lambda_`
* `tests/aws/awslambda` => `tests/aws/lambda_`
* `tests/unit/services/awslambda` => `tests/unit/services/lambda_`

## What stays?

* The S3 storage bucket for Lambda code remains `awslambda-us-east-1-tasks` because of AWS parity.
* The legacy moto constructs called awslambda
* Anything that is related to AWS APIs (e.g., s3control `AwsLambdaTransformation` or CloudFormation resources `AWS::Lambda`)
* The Python AWS CDK called `aws_lambda` uses both: `lambda_` in their [documentation](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/README.html) and `aws_lambda` in their [samples](https://github.com/ksmin23/my-aws-cdk-examples/tree/main/lambda), so keeping that seems fine.
* @steffyP mentioned that no manual updates are required for updating the [Coverage docs](https://docs.localstack.cloud/references/coverage/coverage_lambda/) accordingly. Thank you for clarifying 🙏 

## TODO

- [x] Rename the AwsLambda packages


steffyP:
> We probably need to update the special case mapping in the [Coverage docs](https://docs.localstack.cloud/references/coverage/coverage_lambda/) accordingly.

I don't think there are any manual updates required. We have a weekly update anyhow, so the test-names will be updated accordingly. Everything else should not have an effect, as with the collection we mainly rely on the service name (not the package name), and nothing changed in that regard.

alexrashed (approved):
Awesome! I think the timing us just right, and we can add all these changes to [the git blame ignore revs](https://github.com/localstack/localstack/blob/master/.git-blame-ignore-revs) to avoid polluting the history / blame.
I don't really know what you mean by the release dance though. There is no cyclic dependency anymore. Pro just depends on Community. Once the changes here are done, Pro needs to be upgraded to the latest version of Community. You can also implement the changes right away with our "Companion Branch Detection" (which would be a good idea considering the amount of changes).

dominikschubert (approved):
Great work :partying_face:

dfangl (approved):
LGTM, thanks for this refactoring

joe4dev:
> Awesome! I think the timing us just right, and we can add all these changes to [the git blame ignore revs](https://github.com/localstack/localstack/blob/master/.git-blame-ignore-revs) to avoid polluting the history / blame.
Great. Thank you @alexrashed

@baermat will add them in one go after the upcoming global package restructuring in https://github.com/localstack/localstack/pull/8867. These are the values for the Lambda changes:

* localstack: https://github.com/localstack/localstack/blob/master/.git-blame-ignore-revs

  ```
  # standardize Lambda package structure
  9de01149af6584c8104bd6a02cd87687f6b85083
  ```
* localstack-ext: https://github.com/localstack/localstack-ext/blob/master/.git-blame-ignore-revs

  ```# standardize Lambda package structure
  849c3395a5abbee192ff00b9a797d282a7f34268
  ```

joe4dev:
> I don't really know what you mean by the release dance though. There is no cyclic dependency anymore. Pro just depends on Community. Once the changes here are done, Pro needs to be upgraded to the latest version of Community. You can also implement the changes right away with our "Companion Branch Detection" (which would be a good idea considering the amount of changes).

You are right @alexrashed , I updated the comment accordingly. We synced with Viren, merged together, and now the ext build is 🟢  again (apart from bigdata tests).

PR #8864, opened 2023-08-08
implement S3 native Website

bentsku:
## Motivation
This PR migrate the current S3 Website router to be compatible with both providers, as well as implementing the BucketWebsite API.

## Changes
The current S3 website was very tightly coupled with the moto backend, making it impossible to use with the new provider. We discussed it quickly with @alexrashed, and the internal boto client makes much more sense to use here as it totally abstracts away any difference there might be in the providers. Currently the implementation is the same as the previous one.

I've also changed some operations to be multi-account aware, so that they're able to access any buckets in any account when using the boto client. \cc @viren-nadkarni

### DIFF
--- a/localstack/services/s3/v3/models.py
+++ b/localstack/services/s3/v3/models.py
@@ -131,6 +131,7 @@ class S3Bucket:
         self.notification_configuration = {}
         self.cors_rules = None
         self.lifecycle_rules = None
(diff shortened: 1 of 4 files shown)


thrau (approved):
Looks great! Love the handler approach using the S3 client! :100:

PR #8866, opened 2023-08-09
Enforce exactly one AWS pytest marker on each test case

dominikschubert:
## Motivation

To make sure we don't forget to properly mark tests in `tests/aws/` a new guard in the pytest collection step is introduced which will fail in the collect step, i.e. before attempting to execute any tests.

## Changes

- Tests will immediately fail when they have no `@markers.aws.*` attached
- Tests will also immediately fail when they have more than one `@markers.aws.*` attached
- A new step has been added to the preflight job in circleci to fail as early as possible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,9 @@ jobs:
       - run:
           name: Linting
           command: make lint
+      - run:
+          name: Checking AWS compatibility markers
+          command: make check-aws-markers
 
   unit-tests:
     executor: ubuntu-machine-amd64
--- a/Makefile
+++ b/Makefile
@@ -206,6 +206,9 @@ lint:              		  ## Run code linter to check code style
 lint-modified:     		  ## Run code linter on modified files
 	($(VENV_RUN); python -m pflake8 --show-source `git diff --diff-filter=d --name-only HEAD | grep '\.py$$' | xargs` )
 
+check-aws-markers:     		  ## Lightweight check to ensure all AWS tests have proper compatibilty markers set
+	($(VENV_RUN); python -m pytest --co tests/aws/)
+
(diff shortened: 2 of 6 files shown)


alexrashed (commented):
Nice! The mandatory marks will make it a lot easier to reason over the tests and to decrease the amount of non-validated tests over time.
I only have two nitpicks:
- Is there any chance that this could automatically be integrated into the linting (because it somewhat is a linting rule I'd say)?
  - But if that's not easily possible, maybe add it to the precommit hook?
  - Otherwise I think it could be a bit annoying to realize that the marker is missing after the push / in the pipeline...
- Now that the markers are becoming mandatory, I think we should also mention that in the contribution guide in the docs: https://docs.localstack.cloud/contributing/integration-tests/

steffyP (approved):
Cool 😄 This approach will require more thinking about the kind of test we are writing, hopefully encouraging people to aws-validate the test immediately 🚀

dominikschubert:
> But if that's not easily possible, maybe add it to the precommit hook?
> Otherwise I think it could be a bit annoying to realize that the marker is missing after the push / in the pipeline..

@alexrashed 
Generally the test will already fail when trying to execute the test locally as well so as long as you actually locally executing the test you're writing, you'll see the error before pushing. It's also failing quite quickly now because its part of the preflight check as well.

For the local linting I think verifying via pytest is a bit too heavy weight but we can start looking into adding something like libcst for pre-commit hooks in the future to just check for the decorator.

alexrashed (approved):
Ah, right, thanks for addressing my comment! I haven't thought about the fact that you can't even run the test you are executing, which basically ensures that you are aware of the marker before pushing the commit. Thanks! 💯

PR #8868, opened 2023-08-09
Add schema file for "AWS::OpenSearchService::Domain" resource provider

dominikschubert:
## Motivation

Entrypoints generation was logging an error related to not being able to find the schema file for the `AWS::OpenSearchService::Domain` resource provider.

This provider is not in use yet, but to make the warning go away, I've updated it to the latest scaffolding.

## Changes

- Updates `AWS::OpenSearchService::Domain` to latest scaffolding. Doesn't change anything in terms of logic besides removing the old way of handling repeated invocations.


alexrashed (approved):
Thanks so much for jumping on this! 🥇 🚀

PR #8870, opened 2023-08-09
Add ability to specify custom IDs for KMS keys

whummer:
## Motivation

Some users have requested to specify custom IDs for KMS keys. The old `local-kms` KMS provider used to allow preseeding keys on LocalStack startup. This is now a bit more tricky, as with our new provider each key would get a new ID, making it harder for users to create test scripts that rely on static key IDs.

## Changes

* Add ability to specify custom IDs for KMS keys. We're using the `_custom_id_` tag key, which is already used in other places, e.g., for [API Gateway](https://docs.localstack.cloud/user-guide/aws/apigateway/#custom-ids-for-api-gateway-resources-via-tags)
* Add a small integration test (marked as `only_localstack`) to cover this functionality

## TODO

- [ ] Update the docs to cover this new feature

### DIFF
--- a/localstack/services/kms/provider.py
+++ b/localstack/services/kms/provider.py
@@ -142,7 +142,7 @@ VALID_OPERATIONS = [
     "GenerateDataKeyPairWithoutPlaintext",
 ]
 
-# special tag name to allow specifying a custom ID for new REST APIs
+# special tag name to allow specifying a custom ID for created keys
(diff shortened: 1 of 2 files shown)


sannya-singal (approved):

sannya-singal:
LGTM 🚀🎉

PR #8871, opened 2023-08-09
fix: S3 multipart upload part 10000

### LINKED ISSUE #8869: bug: S3 multipart upload erroneously rejects part number 10000

### Current Behavior

When performing a multipart upload, sending part number 10000, because [this check](https://github.com/localstack/localstack/blob/d23f0717cc66c0cc8a61b6b85e7531a6f4a03f96/localstack/services/s3/provider.py#L792C88-L792C88) in s3/provider.py is wrong.

### Expected Behavior

Part number 10000 is accepted

With a docker-compose file

    docker run localstack/localstack

I'm using AWS .NET SDK:

        using var inputStream = new MemoryStream(buffer, 0, readBytes, false);

        var uploadRequest = new UploadPartRequest
        {
            BucketName = bucketName,
            InputStream = inputStream,
            Key = key,
            PartNumber = partNumber,
            PartSize = readBytes,
            UploadId = uploadId,
        };

        await _client.UploadPartAsync(uploadRequest);

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack: 2.2
```

### Anything else?

As referenced above, the bug is in [this line](https://github.com/localstack/localstack/blob/d23f0717cc66c0cc8a61b6b85e7531a6f4a03f96/localstack/services/s3/provider.py#L792C88-L792C88). It should read

    elif (part_number := request.get("PartNumber", 0)) < 1 or part_number > 10000:

akojo:
Include part number 10000 in allowed part numbers when checking for PartNumber in UploadPart handler.

Fixes #8869

## Motivation

Fix a non-conformity with respect to AWS S3 API. Without this fix `localstack` doesn't allow part number 10000 in multipart upload.

## Changes

Fix part number validation in UploadPart handler

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -789,7 +789,7 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                 "The specified upload does not exist. The upload ID may be invalid, or the upload may have been aborted or completed.",
                 UploadId=upload_id,
             )
-        elif (part_number := request.get("PartNumber", 0)) < 1 or part_number >= 10000:
+        elif (part_number := request.get("PartNumber", 0)) < 1 or part_number > 10000:
             raise InvalidArgument(
                 "Part number must be an integer between 1 and 10000, inclusive",
                 ArgumentName="partNumber",


akojo:

bentsku:
Hello @akojo and thanks a lot for the fix! You're right, the check should have inclusive on the upper limit, as seen here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

We usually ask to provide an AWS validated test for that, but I can add it later on, this seems like a no-brainer. I'll merge!

bentsku (approved):

PR #8873, opened 2023-08-09
Further enhance parity of API Gateway patch operations

whummer:
## Motivation

Further enhance parity of API Gateway patch operations. Follow-up from  #8836. Addresses #8739 Unfortunately,  #8836 introduced a regression where certain fields of Stages cannot be updated now.

In the process of more snapshot testing, I've discovered a few more discrepancies in the moto logic, which are also fixed in this PR (and some obsolete moto patches removed). 

## Changes

* fix regression to allow updating `/*/*/throttling/burstLimit` stage attributes
* extend the snapshot test to cover this functionality
* remove obsolete moto patches

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -142,7 +142,9 @@ STAGE_UPDATE_PATHS = [
     "/*/*/metrics/enabled",
     "/*/*/logging/dataTrace",
     "/*/*/logging/loglevel",
-    "/*/*/throttling/burstLimit /*/*/throttling/rateLimit /*/*/caching/ttlInSeconds",
+    "/*/*/throttling/burstLimit",
+    "/*/*/throttling/rateLimit",
+    "/*/*/caching/ttlInSeconds",
     "/*/*/caching/enabled",
     "/*/*/caching/dataEncrypted",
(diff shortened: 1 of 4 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -880,19 +875,32 @@ def update_stage(
...
-            if not any(re.match(regex, patch_operation["path"]) for regex in path_regexes):
+            patch_path = patch_operation["path"]
+            path_valid = patch_path in STAGE_UPDATE_PATHS or any(
+                re.match(regex, patch_path) for regex in path_regexes
+            )
+            if not path_valid:
                 valid_paths = f"[{', '.join(STAGE_UPDATE_PATHS)}]"
+                # note: weird formatting in AWS - required for snapshot testing
+                valid_paths = valid_paths.replace("/burstLimit, /", "/burstLimit /")
+                valid_paths = valid_paths.replace("/rateLimit, /", "/rateLimit /")
```
bentsku, line 886: such a weird bug, good catch!

bentsku (approved):
LGTM! Awesome to see even more parity, especially those patches being removed... 🙏
Sorry I missed the missing commas in the previous review introducing the regression 😬 but good catch finding it so quickly!

PR #8882, opened 2023-08-10
write fix if attribute is None

sannya-singal:
## Motivation
The issue comes from a support case: [support_johnagray_u04ebfu5uhg](https://localstack-community.slack.com/archives/C05HSNXLQ2J/p1691500975036139?thread_ts=1691470794.419749&cid=C05HSNXLQ2J) where user encountered the following error: `('NoneType' object has no attribute 'get'): 'NoneType' object has no attribute 'get'`. 

## Changes
We add an if condition to check the attribute value exists and only then update the following values. 

## Testing
We can try running the sample provided by the user [here](https://localstack-community.slack.com/archives/C05HSNXLQ2J/p1691154107559409) and check if the resolution works.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -99,7 +99,8 @@ def get_attr_from_model_instance(
         parts = attribute_name.split(".")
         attribute = properties
         for part in parts:
-            attribute = attribute.get(part)
+            if attribute:
+                attribute = attribute.get(part)
         return attribute
 
     # If we couldn't find the attribute, this is actually an irrecoverable error.


simonrw (commented):
Thanks for making this change, but I think the true solution is to make sure the piece of data that's being requested is available in the resource properties. We should ensure that the information is present, and it should be an error case if it isn't.

Could you share the specific resource type and attribute that is being looked up that's causing the issue?

simonrw (approved):
Thanks for making this change. Since this change will not hinder any future refactorings to the template deployer dependency resolution, I'm happy to merge this. I would also like a test, but since it's fixing behaviour that

* is hard to recreate, and
* is better tested when we fix dependency resolution

then I'm happy to skip writing a test in this instance.

Awesome :tada:

PR #8885, opened 2023-08-10
implement S3 native RequestPayment, GetObject Part, Preconditions

bentsku:
## Motivation
This PR implements a small mocked API, BucketPaymentRequest, which is only mocked currently.
It was also an opportunity to add functionality to the "core" operations that were not done. Also, rounds of CI runs showed small issues in the current logic that were hard to spot earlier.

## Changes
This PR implements `BucketRequestPayment` API, which are only mocked and do not have actual effect on the global behaviour of S3 (like moto today).
It also implements preconditions checks for `HeadObject`, `GetObject` and `CopyObject`. 
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Unmodified-Since
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

It also implements a pretty unknown feature of S3, which allows you to query the individual parts after you created an object using a MultipartUpload. It uses the behaviour of a `Range` request, even returning the same kind of headers. It is done under the hood by keep the necessary range data when completing a multipart, and using it to construct a range request when a part is being requested. It seems it is done the same way in AWS, we did not have any tests but I had a hunch it was working that way. 

Small change on the storage engine as well, the ETag is now a property, allowing to be calculated later on if it wasn't done.

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -853,6 +853,14 @@ class ObjectLockConfigurationNotFoundError(ServiceException):
     BucketName: Optional[BucketName]
 
 
+class InvalidPartNumber(ServiceException):
+    code: str = "InvalidPartNumber"
+    sender_fault: bool = False
+    status_code: int = 416
+    PartNumberRequested: Optional[PartNumber]
+    ActualPartCount: Optional[PartNumber]
+
+
 AbortDate = datetime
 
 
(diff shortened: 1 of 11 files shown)


thrau (approved):
LGTM! nice changes all over, and :ok_hand: tests as always

PR #8886, opened 2023-08-10
implement S3 native PublicAccessBlock and ObjectOwnership

bentsku:
## Motivation
Implement Bucket API operations, those are currently only mocked and do not influence S3 behaviour. ObjectOwnership should have some impact, as it should prevent some ACLs to be set on Object, like in AWS. PublicAccessBlock cannot really be enforced, as we do not have the concept of public/anonymous access to S3 (credentials are always being set). It also has some logic tied to `s3control`, where the settings would be the most restrictive between the bucket PublicAccessBlock and the owner ones. As it's only mocked, we currently do not implemented this.

## Changes
Implement the following API operations:
- `GetBucketOwnershipControls`
- `PutBucketOwnershipControls`
- `DeleteBucketOwnershipControls`
- `GetPublicAccessBlock`
- `PutPublicAccessBlock`
- `DeletePublicAccessBlock`

And the validation snapshot tests that go with it.

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -861,6 +861,13 @@ class InvalidPartNumber(ServiceException):
     ActualPartCount: Optional[PartNumber]
 
 
+class OwnershipControlsNotFoundError(ServiceException):
+    code: str = "OwnershipControlsNotFoundError"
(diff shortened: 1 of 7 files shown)


thrau (approved):
LGTM!

Issue #8895, opened 2023-08-11
bug: KINESIS_INITIALIZE_STREAMS not set in latest docker image


etspaceman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Setting `KINESIS_INITIALIZE_STREAMS` no longer is taking in the newest localstack/localstack:latest image. You can see some logs here:

https://github.com/etspaceman/kinesis4cats/actions/runs/5835796510/job/15828063287

`initializeStreams` is null

```
[info] kinesis.mock.KinesisMockService$ 2023-08-11T18:33:21.836147Z contextId=1ce47d17-0b9a-4ad5-927d-4e9fed8b5829, cacheConfig={"awsAccountId":"000000000000","awsRegion":"us-east-1","createStreamDuration":{"length":0,"unit":"MILLISECONDS"},"deleteStreamDuration":{"length":0,"unit":"MILLISECONDS"},"deregisterStreamConsumerDuration":{"length":0,"unit":"MILLISECONDS"},"initializeStreams":null,"logLevel":"INFO","mergeShardsDuration":{"length":0,"unit":"MILLISECONDS"},"onDemandStreamCountLimit":10,"persistConfig":{"fileName":"000000000000.json","interval":{"length":5,"unit":"SECONDS"},"loadIfExists":true,"path":"../../../var/lib/localstack/tmp/state/kinesis","shouldPersist":true},"registerStreamConsumerDuration":{"length":0,"unit":"MILLISECONDS"},"shardLimit":100,"splitShardDuration":{"length":0,"unit":"MILLISECONDS"},"startStreamEncryptionDuration":{"length":0,"unit":"MILLISECONDS"},"stopStreamEncryptionDuration":{"length":0,"unit":"MILLISECONDS"},"updateShardCountDuration":{"length":0,"unit":"MILLISECONDS"}} Logging Cache Config
```

Reverting to 2.2.0 resolves the issue.

### Expected Behavior

`KINESIS_INITIALIZE_STREAMS` sets values

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack:latest

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal kinesis list-streams

### Environment

```markdown
- OS: ubuntu
- LocalStack: latest
```

viren-nadkarni:
[...]

PR #8896, opened 2023-08-11
Set INITIALIZE_STREAMS for Kinesis Mock

etspaceman:
## Motivation

Fix for #8895 

## Changes

Sets INITIALIZE_STREAMS for the kinesis-mock

## Testing

Description of how to test the changes


etspaceman:
Looks like I can't set labels so many of these checks are failing due to that.

viren-nadkarni (changes requested):
@etspaceman Could you fix the linting with `make format`?

viren-nadkarni:
Not sure why the secrets are not being passed to the jobs.

`docker-test-amd64` and `docker-test-arm64` pass, merging PR.

viren-nadkarni (approved):

alexrashed:
I'm sorry for the late answer, but just to answer your questions: Secrets are never able to be used by fork builds, because they could quite easily extract the secrets. We should modify our pipelines such that the use of secrets is optional (which they are currently not). Sorry for the inconvenience, and thanks for the contribution! 🚀

PR #8899, opened 2023-08-12
Apply response template to sqs int

calvernaz:


bentsku (approved):
LGTM! Awesome, thanks a lot for adding an AWS validated test! and great catch 👍

PR #8901, opened 2023-08-13
implement S3 native BucketLogging and BucketReplication

bentsku:
## Motivation
Implement S3 BucketLogging and BucketReplication mocking, only adding validation but not actually emulating the replication.

## Changes
Added a small fix regarding locking when accessing a Multipart, to not create it twice.
Implemented the following API operations:
- `PutBucketLogging`
- `GetBucketLogging`
- `PutBucketReplication`
- `GetBucketReplication`
- `DeleteBucketReplication`


macnev2013 (approved):
LGTM 🚀

PR #8907, opened 2023-08-14
implement S3 native pre-signed POST (PostObject)

bentsku:
## Motivation
Implementing the missing PostObject operation, which is needed for pre-signed POST requests. 

## Changes
`PostObject` is a bit of a weird operation in itself because everything is in a form, and the parser doesn't handle this kind of request right now, and maybe it shouldn't just for one operation. So most of the "parsing" is done directly in the provider method. I've added a bit more features than moto, as we now handle encryption parameters, tagging, and more response fields.


thrau (approved):
LGTM! the light-weight parsing approach by reading the form seems fine to me. a bit of conceptual duplication with the parser but not an issue IMO

PR #8918, opened 2023-08-15
implement S3 v3 pre-signed handler

bentsku:
## Motivation
Move the s3 pre-signed handler earlier in the handler chain, so that we can modify the request in the `RequestContext` before actually parsing it: this allows us to fully use the ASF parser. 
When pre-signing a request, the S3 client will move some of the request parameters into the query string, when those parameters would be in the headers. Now, the pre-signed handler will add back some headers that we know should be in there, and the parser will properly pick it up. 

## Changes
Create a general pre-signed handler in the handler chain, which will be able to manage more kind of pre-signed URL than S3. Only S3 is implemented now. Some logic had to be changed in the handler, because we were using the previously parsed `Bucket` from the request. We now need to parse it ourselves following the same logic as the RequestParser (especially a weird case where we had the sanitized path, and we needed the raw one).


thrau (approved):
LGTM! modifying the headers this way seems fine (we allowed for request header mutability for exactly these use cases). the impact on the handler chain is minimal since we're only executing this for specific services, which is good.
the encapsulation in the handler class is :ok_hand: 

great set of changes overall! :1st_place_medal:

PR #8919, opened 2023-08-16
Fix parsing of booleans in API Gateway Velocity templates

whummer:
## Motivation

We recently had a couple of changes in our `airspeed` for, for better AWS parity - see also recent PR #8914 .

One of the main changes is that booleans are now properly rendered as `true`/`false`, and not via the Python syntax `True`/`False`.

This leads to some potential issues in API Gateway (detected by upstream tests in -ext). In the API GW `ResponseTemplates` we were using `ast.literal_eval`, instead of proper `json.loads`, mainly to be able to deal with the booleans in Python syntax. This is fixed in this PR, `ast.literal_eval` is replaced by `json.loads`, and the logic is extended with a small unit test to cover this functionality.

## Changes

* fix parsing of booleans in API Gateway Velocity templates (replace `ast.literal_eval` with `json.loads`)
* add a small unit test to cover this functionality (also a negative test which shows that booleans like `False` are no longer in the templates)

### DIFF
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -1,4 +1,3 @@
-import ast
 import base64
 import copy
 import json
@@ -178,7 +177,7 @@ class Templates:
         return self.vtl.render_vtl(template, variables=variables)
 
     @staticmethod
-    def build_variables_mapping(api_context: ApiInvocationContext):
+    def build_variables_mapping(api_context: ApiInvocationContext) -> dict[str, Any]:
         # TODO: make this (dict) an object so usages of "render_vtl" variables are defined
         ctx = copy.deepcopy(api_context.context or {})
         # https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-override-request-response-parameters.html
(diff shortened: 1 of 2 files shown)


tests/unit/test_apigateway.py
```diff
@@ -560,6 +561,21 @@ def test_render_custom_template(self, template):
             "stageVariable2": "value2",
         }
 
+    def test_render_valid_booleans_in_json(self):
```
calvernaz, line 564: Similar to https://github.com/localstack/localstack/blob/c7d0336a3eb2f81b7b97af33a2ad135d74489bbe/tests/unit/test_templating.py#L254

localstack/services/apigateway/templates.py
```diff
@@ -283,9 +282,16 @@ def render(self, api_context: ApiInvocationContext, **kwargs) -> Union[bytes, st
...
-        rendered_tpl = self.render_vtl(template, variables=variables)
+        response._content = self._render_json_result(template, variables)
 
-        # TODO: we need to do a parity test for templates that generate invalid JSON
-        response._content = json.dumps(ast.literal_eval(rendered_tpl.strip()))
-        LOG.info("Endpoint response body after transformations:\n%s", response._content)
+        LOG.debug("Endpoint response body after transformations:\n%s", response._content)
         return response._content
+
+    def _render_json_result(self, template: str, variables: dict[str, Any]) -> str:
```
calvernaz, line 290: nit: we only support json templates atm, so this helper method is indeed a good idea I'd just consider a name like `_render_as_json` so we could apply this to others like `_render_as_text`

calvernaz (approved):
Awesome seeing this going :+1: LGTM!

localstack/services/apigateway/templates.py
whummer: Good point, updated 👍

PR #8920, opened 2023-08-16
Add max capabilities when creating nested stacks

dominikschubert:
## Motivation

Nested stacks can also contain transformations and are thus subject to capability checks.
This caused users to report issues regarding failed stack deployments to pop up in support which this PR aims to prevent.

## Changes

- When creating a nested stack (`AWS::CloudFormation::Stack`) it will now always created with all possible capabilities. This should be fixed when migrating to a resource provider.


simonrw (approved):
What does AWS do in this situation? Does it do the same, or does it propagate the original invocation capabilities?

dominikschubert:
> What does AWS do in this situation? Does it do the same, or does it propagate the original invocation capabilities?

I assume so, but not sure. That's why I left that TODO in there :man_shrugging:

We'll need to dig into it at some point but for now it seems safe enough to just allow all here

PR #8922, opened 2023-08-16
Gather analytics on account ID and region name use

viren-nadkarni:
## Motivation

We currently have no insight on how multi-accounts and multi-region namespacing is used in LocalStack. Moto gets a multi-account related issues reported now and then, but in LocalStack it seems quiet.

## Implementation

This PR collects two additional attributes in analytics events:

- Account ID: This field is populated from the Access Key ID. In its default config, LS ignores any keys that start with `ASIA` or `AKIA` and instead uses `000000000000` account ID. This means that in general, this field will be non-PII. However, a user could choose to disable this behaviour with `PARITY_AWS_ACCESS_KEY_ID`. If that is the case, account ID is potentially PII and this information will not collected.
- Region Name: LS forbids arbitrary region names by default but this can be overridden with `ALLOW_NONSTANDARD_REGIONS`. In either cases, this field should be non-PII. 

Internal calls are ignored as before.

### DIFF
--- a/localstack/aws/handlers/analytics.py
+++ b/localstack/aws/handlers/analytics.py
@@ -49,6 +49,8 @@ class ServiceRequestCounter:
                 service_name,
                 operation_name,
                 response.status_code,
+                context.account_id,
+                context.region,
                 err_type=err_type,
             )
         )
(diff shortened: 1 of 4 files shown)


viren-nadkarni:
@whummer @ackdav: What are your thoughts on this - is this info useful/acceptable to collect?

PR #8923, opened 2023-08-16
Add get_free_udp_port utility + test

dfangl:
## Motivation
In order to test the DNS server on arbitrary ports, we need a way to get random free udp ports, similar to the `get_free_tcp_port` utility for tcp.

## Changes
* Add `get_free_udp_port` utility
* Add a test for said utility
* use `shlex.join` instead of a normal string join for shell commandline in `run_module_as_sudo`

### DIFF
--- a/localstack/utils/net.py
+++ b/localstack/utils/net.py
@@ -174,6 +174,18 @@ def port_can_be_bound(port: IntOrPort) -> bool:
         return False
 
 
+def get_free_udp_port(blocklist: List[int] = None) -> int:
+    blocklist = blocklist or []
+    for i in range(10):
+        udp = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        udp.bind(("", 0))
+        addr, port = udp.getsockname()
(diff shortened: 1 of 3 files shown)


localstack/utils/net.py
```diff
@@ -170,10 +170,26 @@ def port_can_be_bound(port: IntOrPort) -> bool:
...
+def get_free_udp_port(blocklist: List[int] = None) -> int:
+    blocklist = blocklist or []
+    for i in range(10):
+        udp = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        udp.bind(("", 0))
+        addr, port = udp.getsockname()
+        udp.close()
+        if port not in blocklist:
+            return port
+    raise Exception(f"Unable to determine free TCP port with blocklist {blocklist}")
```
simonrw, line 190: TCP -> UDP

simonrw (approved):
Looks good, makes sense to have a UDP equivalent of `get_free_tcp_port`.

PR #8925, opened 2023-08-16
provide result handle to AWS::ApiGateway::GatewayResponse

calvernaz:
## Motivation

Support case https://localstack-community.slack.com/archives/C05HSNXLQ2J. The CF resource "AWS::ApiGateway::GatewayResponse" was missing a physical resource id value.

## Changes

Adds a handler function and generates a physical resource id.

Uses the new cdk based test setup, thanks @dominikschubert for the heads up

Also thanks to @sannya-singal for helping out with the discover and testing :+1:


dominikschubert (approved):
LGTM :+1: 

Thanks for giving the new CDK provisioning a spin as well :rocket:

PR #8926, opened 2023-08-16
fix close call propagation of response in asgi/wsgi bridge

thrau:
## Motivation

During the implementation of the new S3 provider, we noticed that `Response.close` is never called on the werkzeug response, and therefore not propagated to other underlying objects. The test I added worked when serving it directly through the werkzeug server, and I found it's calling close [here](https://github.com/pallets/werkzeug/blob/47c6bd568efd0870d413c27be316c2b4eee74486/src/werkzeug/serving.py#L360-L361), where `application_iter` is the `Response` object.

We actually had no analogous call in our ASGI/WSGI bridge, so this PR adds that.

## Changes

* Encapsulate the sync/async iterable wrapper into a class and propagate any calls to `aclose` to the underlying `close` method.

I was briefly worried that this may break [`test_chunked_transfer_encoding_client_timeout`](https://github.com/localstack/localstack/blob/555f8da5d3222ed191867f668d88cd54da274997/tests/unit/http_/test_asgi.py#L101-L103) but seems to still work, so although `aclose` is overwritten, it propagates the `GeneratorExit` exception correctly :+1:

### DIFF
--- a/localstack/http/asgi.py
+++ b/localstack/http/asgi.py
@@ -115,35 +115,41 @@ def populate_wsgi_environment(
     environ["asgi.headers"] = headers
 
 
-async def to_async_generator(
-    it: t.Iterator,
-    loop: t.Optional[AbstractEventLoop] = None,
-    executor: t.Optional[Executor] = None,
-) -> t.AsyncGenerator:
-    """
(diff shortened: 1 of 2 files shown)


dfangl (approved):
LGTM!

PR #8927, opened 2023-08-16
improve S3 v3 storage layer

bentsku:
## Motivation
Improving the current S3 storage layer, introduce 2 more operations to allow easier cleanup of resources with bucket creation and deletion. Also separate between `close` and `reset` the store, to have better semantics between these 2 operations (closing does not always mean deleting all underlying objects, especially with persistence enabled). 
Separate between the `close` call for the Ephemeral object, so that it doesn't close the underlying file object, which would delete it in Ephemeral mode, following the fix to the response with #8926. 

## Changes
Introduce said changes above, `create_bucket` and `delete_bucket` for the storage layer, and preparing support for persistence. 

Also added a quick fix regarding creating bucket in `us-east-1`, I wasn't returning early from the method and ended up overwriting the bucket in the store.

Make `close` a noop for `EphemeralS3StoredObject`, and implement a `delete` method for cleaning up.

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -443,6 +443,9 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                     "Your previous request to create the named bucket succeeded and you already own it.",
                     BucketName=bucket_name,
                 )
+            else:
+                # CreateBucket is idempotent in us-east-1
+                return CreateBucketOutput(Location=f"/{bucket_name}")
 
         if (
             object_ownership := request.get("ObjectOwnership")
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -279,6 +279,7 @@ from localstack.services.s3.validation import (
     validate_website_configuration,
 )
 from localstack.services.s3.website_hosting import register_website_hosting_routes
+from localstack.state import AssetDirectory, StateVisitor
 from localstack.utils.aws.arns import s3_bucket_name
 from localstack.utils.strings import short_uid, to_str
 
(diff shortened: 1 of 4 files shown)


localstack/services/s3/v3/provider.py
```diff
@@ -309,8 +309,16 @@ def on_after_init(self):
         modify_service_response.append(S3Provider.service, s3_presigned_url_response_handler)
         register_website_hosting_routes(router=ROUTER)
 
+    def accept_state_visitor(self, visitor: StateVisitor):
+        visitor.visit(s3_stores)
+        visitor.visit(AssetDirectory(self._storage_backend.root_directory))
```
bentsku, line 314: note: this won't work for community cloud pods, as there's a check that the path should start with `config.dirs.data`

thrau: acceptable limitation for now :+1:

localstack/services/s3/v3/storage/core.py
```diff
@@ -181,6 +183,17 @@ def get_multipart(self, bucket: BucketName, upload_id: MultipartUploadId) -> S3S
...
+        pass
+
+    def flush(self):
+        pass
+
     def close(self):
         pass
+
+    def reset(self):
+        pass
```
thrau, line 199: maybe some doc strings would be helpful to explain what the different behavior of flush,close,reset are

thrau (approved):
nice! glad we did the iterations and now have a clean distinction between reset/flush/close in the contract :+1:

localstack/services/s3/v3/provider.py
```diff
@@ -288,14 +289,13 @@
...
-# TODO: pre-signed URLS -> REMAP parameters from querystring to headers???
-#  create a handler which handle pre-signed and remap before parsing the request!
+DEFAULT_S3_TMP_DIR = "/tmp/localstack-s3-storage"
 
 
 class S3Provider(S3Api, ServiceLifecycleHook):
     def __init__(self) -> None:
         super().__init__()
-        self._storage_backend = EphemeralS3ObjectStore()
+        self._storage_backend = EphemeralS3ObjectStore(DEFAULT_S3_TMP_DIR)
```
thrau, line 298: i mentioned this before: maybe we could allow injecting a storage backend into the constructor

localstack/services/s3/v3/storage/core.py
bentsku: Will do 👍 funnily, I don't think we use flush anywhere at the moment, as persistence is always "on" for the filesystem storage backend. But we could keep it for some fun behaviour with community pods or something?

PR #8944, opened 2023-08-20
[SFN] Skip local iam role assume check

MEPalma:
## Motivation
Speeds up build by skipping local checks for StepFunction's role assume.

## Changes
Skip SFN can assume role check in create state machine iam role fixture.


dominikschubert (approved):
Makes sense :+1:

PR #8946, opened 2023-08-21
fix Response.update_from to also copy close handlers

thrau:
## Motivation

This is in some sense a follow up to #8926. I wanted to use the close callbacks of werkzeug to properly implement the restarting of localstack after extensions have been installed (see ext PR). I noticed this wasn't working because we actually often create new Response objects and then use `Response.update_from` in the handler chain, which didn't copy the internal on close callbacks correctly

## Changes

* Update our `Response.update_from` method to copy the internal `_on_close` handlers correctly
* Add an integration test to make sure the close callbacks work correctly


dfangl (approved):
LGTM!

Issue #8604, opened 2023-07-03
bug: Docker image fails to shutdown when not able to download SSL cert


toadzky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I try to run the localstack docker image (I'm using 2.0), it fails to start. The container doesn't error and shutdown like you would expect, it just sits there in a broken state refusing to shutdown. Settings the `SKIP_SSL_CERT_DOWNLOAD` property fixes the issue. I'm on Ubuntu 22.04. The problem exists on 2.0 and 2.1, but it didn't happen when I first started using the 2.x images.

### Expected Behavior

The container should download the SSL cert and start *or* show the error and terminate the container as a failed launch.

### How are you starting LocalStack?

With a docker command or using testcontainers in java.

### Steps To Reproduce

docker run -p 4566:4566 -e DEBUG=true localstack/localstack:2.1

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack: 2.0, 2.1
```

### Anything else?

```
LocalStack supervisor: starting
LocalStack supervisor: localstack process (PID 15) starting

LocalStack version: 2.1.0
LocalStack build date: 2023-05-25
LocalStack build git hash: 67b55096

2023-07-03T15:39:57.846 DEBUG --- [  MainThread] stevedore._cache           : reading /root/.cache/python-entrypoints/ccd88d0aa2b1947f243f2cbc674104fe0eeb286df48ae4b431534e29c134a33c
2023-07-03T15:39:57.857 DEBUG --- [  MainThread] stevedore._cache           : writing to /root/.cache/python-entrypoints/ccd88d0aa2b1947f243f2cbc674104fe0eeb286df48ae4b431534e29c134a33c
2023-07-03T15:39:57.857 DEBUG --- [  MainThread] stevedore.extension        : found extension EntryPoint(name='_patch_botocore_json_parser', value='localstack.aws.client:_patch_botocore_json_parser', group='localstack.hooks.on_infra_start')
2023-07-03T15:39:57.858 DEBUG --- [  MainThread] stevedore.extension        : found extension EntryPoint(name='_publish_config_as_analytics_event', value='localstack.runtime.analytics:_publish_config_as_analytics_event', group='localstack.hooks.on_infra_start')
[...]

PR #8948, opened 2023-08-21
Bind hypercorn to privileged ports if root

simonrw:
## Motivation

We currently split ports by privileged and unprivileged, and start TCP proxies for privileged ports (as root) since we don't want to run the server on the host as root. 
If we are _already_ root, then we do not need to make this split.

## Changes

* If we are running as root, bind the hypercorn server to privileged ports as well.

## Questions

* This means that if the privileged ports cannot be bound for some reason, then _no_ ports will be bound, and we won't be able to start LocalStack. Is this a concern in practice?

### DIFF
--- a/localstack/services/edge.py
+++ b/localstack/services/edge.py
@@ -488,6 +488,11 @@ def start_edge(listen_str: str, use_ssl: bool = True, asynchronous: bool = False
 
     # separate privileged and unprivileged addresses
     unprivileged, privileged = split_list_by(listen, lambda addr: addr.is_unprivileged() or False)
+    if is_root():
+        # we don't need to split since we are already root, so move all
+        # privileged ports to unprivileged and serve via hypercorn
+        unprivileged = unprivileged + privileged
+        privileged = []
 
     # check that we are actually started the gateway server
     if not unprivileged:


dfangl (approved):
Since docker version 20.03 (https://github.com/moby/moby/pull/41030) we do not need root privileges to bind to privileged ports anymore. 
So, the is_root check might miss cases where we are in docker and could still bind without a proxy, even if localstack is not running as root.

I am fine with this approach for now, but we should revisit this logic in the future to allow non-root users in docker to still bind the ports directly.

alexrashed (approved):
Nice! This change effectively disables the TCP proxies and instead directly binds hypercorn to all ports in our docker image (which is aweseome!).

Maybe we could rework the startup a bit in the next iteration such that:
- The startup supports using priviledged ports without root permissions with the right Docker version ([see @dfangl's comment](https://github.com/localstack/localstack/pull/8948#pullrequestreview-1594082695))
- The startup completely fails and the process / docker container stops in case the startup of hypercorn fails (either because we can't bind a port, because the certs are missing - see #8604, or due to any other issues), because the HTTP server is a completely essential component. It doesn't make any sense to continue the process at all.

simonrw:
> * The startup supports using priviledged ports without root permissions with the right Docker version ([see @dfangl's comment](https://github.com/localstack/localstack/pull/8948#pullrequestreview-1594082695))

Sounds like a good extension.

> * The startup completely fails and the process / docker container stops in case the startup of hypercorn fails (either because we can't bind a port, because the certs are missing - see [bug: Docker image fails to shutdown when not able to download SSL cert #8604](https://github.com/localstack/localstack/issues/8604), or due to any other issues), because the HTTP server is a completely essential component. It doesn't make any sense to continue the process at all.

This would be a serious quality of life improvement, at least provided we can explain clearly what has occurred and why it's a problem.

PR #8950, opened 2023-08-21
Add AWS::Kinesis::Stream resource provider

giograno:
A very minimal implementation to play out with the new CFN provider.


simonrw:
Thanks for adding this feature, and for investigating the resource provider framework. As you probably know we have taken this initiative and implemented the kinesis stream provider separately (in #9230). Thanks for trying out the new resource provider framework, but I'll close this PR as this work is complete.

PR #8953, opened 2023-08-22
add a note for insiders

HarshCasper:


SimonWallner (changes requested):
image
I almost missed it in the review, since I'm not looking for relevant content above the header. I immediately skip the header.

We could have such a 'news/announcements' block or callout after the <hr> and the next header. maybe also with a <hr> in between, something like this


[header card thing]
...
-----------------------------------------
:zap: join or insiders program...
-----------------------------------------

# Overview

SimonWallner (approved):
feel free to merge as is

PR #8954, opened 2023-08-22
[SFN] Support for DescribeStateMachineForExecution

MEPalma:
## Motivation
Adds support for missing api action `DescribeStateMachineForExecution`.

## Changes
Added support in the provider for v2.


tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -653,3 +653,35 @@ def test_create_update_state_machine_same_parameters(
             stateMachineArn=state_machine_arn
         )
         sfn_snapshot.match("describe_resp_t2", describe_resp_t2)
+
+    @markers.aws.unknown
```
dominikschubert, line 657: Why add an `unknown` marker here? Generally these markers should never really be added, they're just here for completeness sake. :thinking:

dominikschubert (approved):
LGTM :+1: 

Please adjust the markers before merging though. Since you've been collecting snapshots it would seem they should be marked as `validated`

PR #8955, opened 2023-08-22
Add back support for type hints in PyCharm for snapshot fixture

dominikschubert:
## Motivation
It seems we lost the type hinting in PyCharm for the snapshot fixture in our tests, which makes especially adding new tests much less intuitive.

## Changes

- Introduce new _snapshot_session fixture
- Add (technically redundant) snapshot fixture so that PyCharm picks up the correct typing.

## Coming up

-ext will receive a similar fixture that will expose the typed _snapshot_session and thus allow  us to have types again.


steffyP (approved):
Thanks for tackling this, @dominikschubert 🙏 
LGTM 👍

PR #8957, opened 2023-08-22
Fix missing implementation of CloudFormation EC2 UserData

joe4dev:
## Motivation

Support issue with CDK stack where `UserData` is not applied to freshly launched EC2 instances. `UserData` itself is implemented and works in EC2.

## Changes

Add the missing CloudFormation implementation for the `UserData` property. It requires a callback to decode the base64-encoded field.

Minor refactoring of verbose current folder implementation (initially started with EC2 test but I noticed that is better covered in a scenario test).

## Testing

Tested through custom CDK stack and support sample. An EC2 scenario test is partially drafted for localstack-ext (follow up in separate PR).

### DIFF
--- a/localstack/services/cloudformation/models/ec2.py
+++ b/localstack/services/cloudformation/models/ec2.py
@@ -1,3 +1,4 @@
+import base64
 import json
 
 from moto.ec2.utils import generate_route_id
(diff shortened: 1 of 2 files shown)


localstack/services/cloudformation/models/ec2.py
```diff
@@ -530,6 +531,13 @@ def add_defaults(resource, stack_name: str):
     @staticmethod
     def get_deploy_templates():
         # TODO: validate again
+
+        def get_user_data_decoded(
+            properties: dict, logical_resource_id: str, resource: dict, stack_name: str
```
viren-nadkarni, line 536: With #8822 merged, all callbacks get two additional args passed:

```suggestion
            account_id: str, region_name: str, properties: dict, logical_resource_id: str, resource: dict, stack_name: str
```

viren-nadkarni (changes requested):

PR #8958, opened 2023-08-22
increase WaiterConfig for setting up scenarios on AWS

steffyP:
## Motivation
The current configuration of the `WaiterConfig` used for the CDK based setup, does have a very limited waiting time configured. 
When running a complex scenario, e.g. setting up several database-cluster with instances, it will timeout on AWS before the deployment finishes, therefore failing the test, potentially starting the teardown. 

## Changes
* Increased the `WaiterConfig` for setup + teardown
* Only delete the CDK_BOOTSTRAP_PARAM for LocalStack runs

### DIFF
--- a/localstack/testing/scenario/provisioning.py
+++ b/localstack/testing/scenario/provisioning.py
@@ -16,7 +16,8 @@ from localstack.aws.connect import ServiceLevelClientFactory
 
 LOG = logging.getLogger(__name__)
 CDK_BOOTSTRAP_PARAM = "/cdk-bootstrap/hnb659fds/version"
-WAITER_CONFIG = {"Delay": 10, "MaxAttempts": 1000}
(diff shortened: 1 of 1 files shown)


localstack/testing/scenario/provisioning.py
```diff
@@ -14,6 +16,7 @@
 
 LOG = logging.getLogger(__name__)
 CDK_BOOTSTRAP_PARAM = "/cdk-bootstrap/hnb659fds/version"
+WAITER_CONFIG = {"Delay": 10, "MaxAttempts": 1000}
```
dominikschubert, line 19: I think we should have two configs here, one for AWS and one for localstack since a delay of 10s is probably unnecessary in our CI runs

dominikschubert (approved):
LGTM! Thanks also for addressing the comment already :+1:

PR #8960, opened 2023-08-22
route53 health check cfn implmentation

pinzon:
## Motivation
Cloudformation is unable to deploy a Route53 HealthCheck resource

## Changes
- Add resource model
- Add some missing properties for other models

## Testing
The test validates the deployment of the resource


tests/aws/services/cloudformation/resources/test_route53.py
```diff
@@ -44,3 +44,15 @@ def test_create_record_set_without_resource_record(deploy_cfn_template, route53_
...
+@markers.aws.validated
+def test_create_health_check(
+    deploy_cfn_template,
+):
+    deploy_cfn_template(
+        template_path=os.path.join(
+            os.path.dirname(__file__),
+            "../../../templates/route53_healthcheck.yml",
+        ),
+    )
```
dominikschubert, line 58: nit: would be great if you could add a service call here to verify deployment (& ideally a snapshot of it)

localstack/services/cloudformation/models/apigateway.py
```diff
@@ -877,6 +877,21 @@ def _handle_result(
             resource: dict,
         ):
             resource["PhysicalResourceId"] = result["domainName"]
+            hosted_zones = connect_to().route53.list_hosted_zones()
+            """
+            The hardcoded value is the only one that should be returned but due limitations it is not possible to
+            use it.
+            """
```
dominikschubert, line 884:
```suggestion
            # The hardcoded value is the only one that should be returned but due limitations it is not possible to use it.
```

dominikschubert: what limitations?

dominikschubert (approved):
LGTM with a few minor nits and questions

localstack/services/cloudformation/models/apigateway.py
pinzon: The CloudfrontDistributionZoneId that it's hardcoded it's the only value that should be return. But it requires that the cloudfront service always have that existing zone. which it doesn't. so the resource model returns the id of the first zone it finds before putting the correct value.

PR #8961, opened 2023-08-22
Fix community test against Pro runs

dfangl:
## Motivation
We are currently running out of memory in our test runs.
This is a combination of the recent test reordering, which showed us this issue, and insufficient test cleanup.

## Changes
* Cleanup serverless tests, even if the command fails
* Cleanup opensearch instances after starting them for endpoint strategy tests

### DIFF
--- a/tests/aws/test_network_configuration.py
+++ b/tests/aws/test_network_configuration.py
@@ -31,8 +31,10 @@ class TestOpenSearch:
         self, opensearch_create_domain, assert_host_customisation, aws_client
     ):
         domain_name = f"domain-{short_uid()}"
-        res = opensearch_create_domain(DomainName=domain_name)
-        endpoint = res["DomainStatus"]["Endpoint"]
+        opensearch_create_domain(DomainName=domain_name)
+        endpoint = aws_client.opensearch.describe_domain(DomainName=domain_name)["DomainStatus"][
+            "Endpoint"
+        ]
 
         assert_host_customisation(endpoint, use_localstack_cloud=True)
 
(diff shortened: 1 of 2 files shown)


tests/aws/test_network_configuration.py
```diff
@@ -28,23 +28,30 @@ class TestOpenSearch:
 
     @markers.aws.only_localstack
     def test_default_strategy(
-        self, opensearch_wait_for_cluster, assert_host_customisation, aws_client
+        self, opensearch_wait_for_cluster, assert_host_customisation, aws_client, cleanups
```
simonrw, line 31: There's a fixture for this that combines `create_domain` and `opensearch_wait_for_cluster`: `opensearch_create_domain`.

tests/aws/test_serverless.py
```diff
@@ -14,8 +15,11 @@ def get_base_dir() -> str:
     return os.path.join(os.path.dirname(__file__), "serverless")
 
 
+LOG = logging.getLogger(__name__)
+
+
 class TestServerless:
-    @pytest.fixture(scope="session")
+    @pytest.fixture(scope="class")
```
simonrw, line 22: nit: Why? There are no other tests in this file, and the fixture is only scoped to this class anyway.

simonrw (approved):
Looks good, hopefully this will bring the memory usage down a peg or two...

Also: me <- 🤦

tests/aws/test_serverless.py
dfangl: Because the fixture will not be cleaned up at the end of the class, even if it is only available (from a python standpoint) in there. A test run I had, did not perform the cleanup after the test class execution finished.

From the docs:
```
session: the fixture is destroyed at the end of the test session.
```

tests/aws/test_network_configuration.py
dfangl: You are right, I somehow thought you wanted to do the call yourself for a reason, I will replace it with the next commit

PR #8964, opened 2023-08-22
use LOCALSTACK_HOST var for external service urls

giograno:
## Motivation
We deprecated `HOSTNAME_EXTERNAL` since 2.0.0. and we replaced it with `LOCALSTACK_HOST`.
Some services rely on `external_service_url` for some use cases (e.g., Cognito uses it to generate the issuer URL generation in JTW tokens).

## Changes
This PR uses now `LOCALSTACK_HOST` as a default host when calling `external_service_url`.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1264,13 +1264,15 @@ def get_protocol() -> str:
     return "https" if USE_SSL else "http"
 
 
-def service_url(service_key: str, host: str | None = None, port: int | None = None) -> str:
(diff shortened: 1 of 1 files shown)


giograno:
As discussed with @simonrw, we are postponing this change due to the upcoming rework on these environment variables.
Until that, we advise our users to still keep using `HOSTNAME_EXTERNAL` cc: @MarcelStranak

PR #8965, opened 2023-08-22
add pods support for plugin discovery

bentsku:
## Motivation
While working on the S3 persistence workaround for mounted volumes, I would need to set up a specific pod logic when creating/injecting pods.
Basically, the data in S3 would live inside the container and not in the default `assets` folder (`config.dirs.data`). There's some custom logic needed, and I would need to be able to set up a plugin for it. This is the first step to be able to introduce it. 

## Changes
Creating a simple plugin to be able then declare and discover specific service related pod visitor. 

## TODO

What's left to do:

- [ ] run benchmarks with and without the S3 persistence workaround to see if this is really needed (still need to merge the last PRs to be able to run them)
- [ ] ...

### DIFF
--- a/localstack/state/pods.py
+++ b/localstack/state/pods.py
@@ -1,0 +1,28 @@
+import zipfile
+
+from plugin import Plugin
+
+from .core import Decoder, Encoder, StateVisitor
+
+
+class PodsPersistencePlugin(Plugin):
+    """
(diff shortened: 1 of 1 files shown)


bentsku:
Closing as we shouldn't need pod plugin 😄

PR #8967, opened 2023-08-23
Techdebt: Remove obsolete Lambda testutil

viren-nadkarni:
## Motivation

The test util `localstack.utils.testutil.delete_lambda_function` did not clean up the resources if they're created in non-default account ID or region. This PR removes this utility.

## Implementation

Based on how the Lambda function is created, the PR makes two changes:
- If the function is created using the fixture `create_lambda_function`, we remove the explicit delete operation since the fixture has proper cleanup stage built-in.
- If the function is created using the testutil `create_lambda_function`, we replace with a call to `DeleteFunction` operation using the `aws_client` fixture.


tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -1694,8 +1694,8 @@ def test_rest_api_multi_region(
 
     delete_rest_api(apigateway_client_eu, restApiId=api_eu_id)
     delete_rest_api(apigateway_client_us, restApiId=api_us_id)
-    testutil.delete_lambda_function(name=lambda_name, region_name="eu-west-1")
-    testutil.delete_lambda_function(name=lambda_name, region_name="us-west-1")
+    aws_client_factory(region_name="eu-west-1").lambda_.delete_function(FunctionName=lambda_name)
```
joe4dev, line 1697: Even better than manually deleting functions would be adopting the fixture `localstack.testing.pytest.fixtures.create_lambda_function`. It automatically cleans up dependencies (e.g., logs, role) properly. Further, it also waits until the function is active to simplify the test case.

joe4dev (approved):
nice cleanup 🧹 

👉 Adopting the `create_lambda_function` fixture would be even better simplifying the test case and properly cleaning up all resources.

tests/aws/services/apigateway/test_apigateway_basic.py
viren-nadkarni: `localstack.testing.pytest.fixtures.create_lambda_function` doesn't offer the flexibility to override the AWS region :frowning_face:

PR #8968, opened 2023-08-23
CloudWatch: Multi-accounts compatibility

viren-nadkarni:
## Motivation

When using the non-default account ID or region, CloudWatch must create the logs in the same account and region.

## Implementation

This PR makes use of appropriate accounts IDs and regions for each of the integrations that connect to CloudWatch or store logs there.


dfangl (approved):
LGTM!

PR #8969, opened 2023-08-23
add scenario tests for stepfunctions and lambda

steffyP:
## Motivation
Add another scenario test, this time for [loan-broker](https://github.com/localstack-samples/sample-loan-broker-stepfunctions-lambda). It's only a very basic step-function-sample; parametrized the test and snapshotted the results.

## Changes
* only included the simple stack (there is [another PR](https://github.com/localstack/localstack/pull/8803) where we also added the second stack for this sample; with a slightly different setup of the same scenario)
  * as the second stack doesn't output reliable results, and would need some restructuring, I would leave this out (at least for the beginning)
  * we can iterate or extend the sample later on (doing some tests with dynamdb for example). 
  * we discovered an issue with eventbus and lambda while working on the second stack, which was already fixed (and tested) in a [previous PR](https://github.com/localstack/localstack/pull/8844)

* added snapshots
* one test is failing on LocalStack, and is marked with `xfail`

* renamed the resource-directories (also for the note-taking-scenario)
  * we should decide soon on the naming of the scenarios, it probably makes more sense to highlight the used services :)


tests/aws/scenario/test_loan_broker.py
```diff
@@ -0,0 +1,259 @@
...
+            # score linked to this SSN will receive quotes
+            ({"SSN": "123-45-6789", "Amount": 5000, "Term": 30}, "SUCCEEDED"),
+            # score linked to this SSN will not receive quotes, but step function call succeeds
+            ({"SSN": "458-45-6789", "Amount": 5000, "Term": 30}, "SUCCEEDED"),
+            ({"SSN": "inv-45-6789", "Amount": 5000, "Term": 30}, "FAILED"),
+            ({"unexpected": "234-45-6789"}, "FAILED"),
+            pytest.param(
+                {"SSN": "234-45-6789"},
+                "FAILED",
+                marks=pytest.mark.xfail(reason="stays in RUNNING on LS, but should be FAILED"),
```
dominikschubert, line 106: whats the behavior of v2 here? :thinking:

tests/aws/scenario/test_loan_broker.py
```diff
@@ -0,0 +1,259 @@
...
+BANK_APP_JS = "./resources_loan_broker/bank_app.js"
+
+
+def _read_file_as_string(filename: str):
+    file_path = os.path.join(os.path.dirname(__file__), filename)
+
+    content = None
+    with open(file_path, "r") as file:
+        content = file.read()
+    return content
```
dominikschubert, line 40: nit: could also use our `load_file` util here

tests/aws/scenario/test_loan_broker.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@dataclass
+class Bank:
+    bank_id: str
+    base_rate: str
+    max_loan: str
+    min_credit_score: str
+
+    def get_env(self) -> str:
```
dominikschubert, line 50: the return value here doesn't seem correct :thinking:

tests/aws/scenario/test_loan_broker.py
```diff
@@ -0,0 +1,259 @@
...
+
+        # set skip_teardown=True to prevent the stack to be deleted
+        with infra.provisioner(skip_teardown=True) as prov:
+            if not infra.skipped_provisioning:
+                # here we could add some initial setup, e.g. pre-filling the app with data
+                bank_addresses = [{"S": bank_name} for bank_name in self.BANKS.keys()]
+                aws_client.dynamodb.put_item(
+                    TableName=LOAN_BROKER_TABLE,
+                    Item={"Type": {"S": "Home"}, "BankAddress": {"L": bank_addresses}},
+                )
```
dominikschubert, line 91: An alternative here would be to do this in a test itself, since you could also consider these operations a sort of scenario on the infrastructure :thinking: 
Let's discuss this offline later :+1:

dominikschubert (approved):
LGTM

tests/aws/scenario/test_loan_broker.py
steffyP: oopsi 🤦‍♀️

tests/aws/scenario/test_loan_broker.py
steffyP: @dominikschubert: moved this part, and included some additional testing with dynamodb (update/get-item/scan/delete-item)

PR #8972, opened 2023-08-23
Update dynamodb snapshots

steffyP:
## Motivation
Part of the workstream for AWS testing: updating old snapshots. 
This PR is for `dynamodb` service

## Changes
- updated snapshots
- removed skip parameters for a few parameters
- added skip parameter for `DeletionProtectionEnabled`
- fixed invalid catch in a test (did not work against AWS)

## TODO

What's left to do:

- [x] `test_data_encoding_consistency` is marked as localstack_only but has a snapshot


giograno (approved):
👍

PR #8973, opened 2023-08-23
Fix result handler signature AWS::ApiGateway::GatewayResponse result handler

dominikschubert:
## Motivation
The signature required for the result handlers has recently been  changed in https://github.com/localstack/localstack/pull/8822 but afterwards https://github.com/localstack/localstack/pull/8925 was merged which added a result handler with the old signature.

## Changes

- fix the signature of the result_handler introduced in https://github.com/localstack/localstack/pull/8925


calvernaz (approved):
Thanks @dominikschubert for tackling this mismatch.

PR #8975, opened 2023-08-23
Add tags support for CloudFormation DynamoDB create table

joe4dev:
## Motivation

Tags for creating DynamoDB table are not created. The issue was raised in a support case.

## Changes

* Add support for resource tags for the create table operation.
* Add `aws.validated` marker for `test_deploy_stack_with_dynamodb_table` and `test_globalindex_read_write_provisioned_throughput_dynamodb_table` because both tests ran successfully against AWS when executing all tests in the file.

## Testing

Extended an AWS-validated test case to cover tags.
Fixes the sample from the support case.


viren-nadkarni (approved):
LGTM :+1:

Issue #8816, opened 2023-08-03
bug: Subscription attributes mismatch not handled


artur-ma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When sending SubscribeCommand with already existing subscription but different attributes, it succeeds.

### Expected Behavior

It should fail same is real aws cloud fails withe error

```
InvalidParameterException: Invalid parameter: Attributes Reason: Subscription already exists with different attributes
```

```
  Error: {
    Type: 'Sender',
    Code: 'InvalidParameter',
    Message: 'Invalid parameter: Attributes Reason: Subscription already exists with different attributes',
    message: 'Invalid parameter: Attributes Reason: Subscription already exists with different attributes'
  },
```

### How are you starting LocalStack?

With a `docker run` command

### Steps To Reproduce

```ts
const sqs = new SQSClient()
  const sns = new SNSClient()

  const { QueueUrl: queueUrl } = await sqs.send(new CreateQueueCommand({
    QueueName: 'test-queue'
  }))

  const { Attributes: queueAttrs} = await sqs.send(new GetQueueAttributesCommand({
    AttributeNames: ['QueueArn'],
    QueueUrl: queueUrl
  }))

  const { TopicArn: topicArn } = await sns.send(new CreateTopicCommand({
    Name: 'test-topic'
  }))

  await sns.send(new SubscribeCommand({
    TopicArn: topicArn,
    Endpoint: queueAttrs?.QueueArn!,
    Protocol: 'sqs',
    Attributes: {
      RawMessageDelivery: 'false',
    }
  }))

  await sns.send(new SubscribeCommand({
    TopicArn: topicArn,
    Endpoint: queueAttrs?.QueueArn!,
    Protocol: 'sqs',
    Attributes: {
      RawMessageDelivery: 'true',
    }
  }))
```

### Environment

```markdown
- OS: osX
- LocalStack: 2.2.0
```

PR #8976, opened 2023-08-23
fix SNS subscribe idempotency with different attrs

bentsku:
## Motivation
This issue was raised in #8816, SNS would not raise an exception if the `Subscribe` call had different attributes when the subscription already exists.

## Changes
We now check if the found subscription if it exists would have the same attributes, and if it doesn't, raise an exception. Added a validated AWS test with it.


baermat (approved):
LGMT :+1: and sorry for the late reply!

PR #8981, opened 2023-08-24
add refactor PRs to ignore-revs

baermat:
## Motivation
This PR adds the two recent commits that improve standardization of our folder structure to ignore-revs, since these PRs touch a lot of files without changing functionality

## Changes
adds https://github.com/localstack/localstack/pull/8867 and https://github.com/localstack/localstack/pull/8861 to .git-blame-ignore-revs

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -2,3 +2,7 @@
 8c006f122803dd06811343782b34f067f0622baf
 # reorganize utility package
 a7602821ac8666c9ab3a253457eb99570908ac8e
+# standardize Lambda package structure
+9de01149af6584c8104bd6a02cd87687f6b85083
+# standardize tests folder structure
+1ec4e7ff5126f2d7a85b536e70b432502b5bad25


alexrashed:
Hey @baermat!
Thanks for setting up the ignore, which I asked for in https://github.com/localstack/localstack/pull/8861#pullrequestreview-1569392167. Actually, it seems like the changes aren't too big anyways (not comparable with the while black code style fixes and others) and all the file moves are correctly recognized. I'm sorry for the inconvenience, but I think we might not have to add them to the ignored revisions after all?
/cc @joe4dev @dominikschubert

baermat:
@alexrashed this is my first time working with ignore-revs and I do not know the finer implications of this. I was just executing the wishes relayed to me :) If you say this is not needed after all, I will follow your judgement. I just want to point out that this also includes the commit that refactors the other services' test structure. If this is again not causing really any issues, feel free to close this :)

alexrashed:
I'm sorry for the confusion and the inconvenience.

Just to clarify: The `.git-blame-ignore-revs` is used to list commit hashes for commits which should not be considered in the `git blame` view on [GitHub or in your local git client (if you configure it accordingly)](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).
This can be very useful if you have single commits which did a lot of refactoring (most of the time in an automated way), but this change doesn't really have any influence. Perfect examples are commits to introduce code style linting (basically reformatting the whole code base without really changing anything).

For this case, where we moved a lot of test code, I misjudged the situation. I thought it would be useful (because it might pollute git blame), but these file moves and import changes don't really have an impact on the blame view at all.
This is why I don't think it is useful to add these commits anymore (even though I explicitly asked for it before). Sorry for that.

baermat:
Closed because no longer necessary

PR #8982, opened 2023-08-24
Move lambda multiruntime test logic to testing module

dfangl:
## Motivation
We also want multiruntime tests in -ext, in this case for transparent endpoint injection tests.
This slight refactoring basically moves the logic to the testing/aws/lambda_utils module instead of the conftest.py, which enables us to use the logic with very little duplicated code in -ext.

Also, there were some logical issues recently introduced with changes in edge.py for `run_module_as_sudo` which are corrected with this pr /cc @simonrw 

## Changes
* Move the multiruntime test logic to testing module
* Fix `run_module_as_sudo` logic

### DIFF
--- a/localstack/services/edge.py
+++ b/localstack/services/edge.py
@@ -7,7 +7,7 @@ import shlex
 import subprocess
 import sys
 import threading
-from typing import Dict, List, Optional, TypeVar
+from typing import List, Optional, TypeVar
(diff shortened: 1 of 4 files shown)


dominikschubert (approved):
LGTM! Great that someone finally tackled this :sweat_smile: 

Thanks :+1:

PR #8983, opened 2023-08-24
refresh SNS snapshots

bentsku:
## Motivation
Refreshing SNS snapshots which were for most 1 year old

## Changes
Organised the tests in more classes, to try to separate a little the different cases. It's not very precise, but it allows a bit more fine grained organisation, and it's easier to run/refresh only a certain kind of tests only after a change instead of re-running the whole suite locally. 

Refreshing snapshots put into a light a change regarding FIFO topics: they now accept regular SQS queue as subscribers. Those are not guaranteed to be FIFO, but the deduplication is working (as it's a the topic level, not the subscriber queue level). This is not yet implemented in LocalStack, but there's a TODO to implement it.

Also added a dummy subscription principal to the SNS subscription, as the field was missing. Maybe we could do a call to STS to get the real IAM user doing the call.


steffyP (approved):
Nice work, @bentsku! 🚀
Changes LGTM 😄

PR #8987, opened 2023-08-25
Support checking a specific host for port bindability

simonrw:
## Motivation

When starting the DNS server, we check if we can publish 0.0.0.0:53, but on systems running systemd-resolved, 127.0.0.1:53 is bindable but 127.0.0.53 is not and the check returns falsy.

## Changes

We allow a custom host to be tested for whether a port can be bound or not.


simonrw:
This was implemented in #9049

PR #8988, opened 2023-08-25
[SFN] ApiGW url implicit endpoint conversion

MEPalma:
## Motivation
When evaluating `StepFunction`'s `ApiGateway` tasks, the provided `ApiEndpoint` is converted to a path style LocalStack endpoint, if a url one is provided.
Improves loading of json `ResponseBody` and removes out of scope test inputs.

## Changes
Added url transformation logic to the `ApiGateway`'s task resolver, as well as json request body parsing attempt.


dominikschubert (approved):
Makes sense to convert to path-based URLs for now :+1:

PR #8989, opened 2023-08-25
Add resource provider for AWS::SQS::QueuePolicy

dominikschubert:
## Motivation

We already enabled the AWS::SQS::Queue resource provider, so I thought it might be a good idea to "finish" the service by migrating the only other remaining CFn model for SQS

## Changes

- Add new resource provider for `AWS::SQS::QueuePolicy`


simonrw (approved):

localstack/services/sqs/resource_providers/aws_sqs_queuepolicy.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    def read(
+        self,
+        request: ResourceRequest[SQSQueuePolicyProperties],
+    ) -> ProgressEvent[SQSQueuePolicyProperties]:
+        """
+        Fetch resource information
+        """
+        raise NotImplementedError
```
thrau, line 74: Q: why leave out read here?

thrau (approved):
LGTM! just wondering why we don't have read implemented

localstack/services/sqs/resource_providers/aws_sqs_queuepolicy.py
dominikschubert: Not in use currently. Just aiming to migrate a few resources from their GenericBaseModel implementation to a "raw" ResourceProvider one.

PR #8992, opened 2023-08-25
refresh missing S3 snapshot

bentsku:
## Motivation
From the last report, it seemed some S3 snapshots were not refreshed. I think this was due to my rebasing, I must have failed some conflicts and kept the old date. 

## Changes
Refreshed the missing snapshots, only the dates are changed, it feels a bit silly but at least we're sure it works!


steffyP (approved):
Nice, thanks @bentsku 👍

PR #9001, opened 2023-08-28
fix global analytics bus to correctly disable tracking

thrau:
## Motivation

While working on the analytics backend, we noticed that the client actually isn't reacting to `track_events=False` responses from the server. Good catch @ackdav.

## Changes

* Properly disable the global analytics bus if the backend responds with `track_events=False`
* Add a gate for 403 errors, which we will introduce in the analytics backend to properly terminate older clients. In the older clients a 403 would trigger an exception in the `AnalyticsClient.start_session` method, also leading to disabling of tracking, albeit less gracefully.

### DIFF
--- a/localstack/utils/analytics/client.py
+++ b/localstack/utils/analytics/client.py
@@ -19,12 +19,17 @@ LOG = logging.getLogger(__name__)
 class SessionResponse:
 
     response: Dict[str, Any]
+    status: int
 
-    def __init__(self, response: Dict[str, Any]):
+    def __init__(self, response: Dict[str, Any], status: int = 200):
         self.response = response
+        self.status = status
 
     def track_events(self) -> bool:
         return self.response.get("track_events")
+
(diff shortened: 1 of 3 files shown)


localstack/utils/analytics/client.py
```diff
@@ -19,13 +19,18 @@
...
-    def __init__(self, response: Dict[str, Any]):
+    def __init__(self, response: Dict[str, Any], status: int = 200):
         self.response = response
+        self.status = status
 
     def track_events(self) -> bool:
         return self.response.get("track_events")
 
+    def __repr__(self):
+        return f"SessionResponse({self.status},{self.response!r})"
```
ackdav, line 32: :magic_wand: `!r` :smile:

ackdav (approved):
awesome! that makes the path more explicit and easier to follow :muscle: lgtm :+1:

PR #9005, opened 2023-08-29
Update outdated sqs snapshots/enable sse

baermat:
## Motivation
Some of our sqs snapshots are quite old by now. This PR updates all snapshots regarding SQS that are older than 9 months (as recorded in the snapshot age report).
While doing so, it became apparent that AWS is enabling SQS server side encryption by default [since 17/8/22
](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/step-create-queue.html). This PR addresses this in a mock fashion.

## Changes
- remove snapshot of "tests/aws/services/sqs/test_sqs.py::TestSqsProvider::test_create_queue_with_attributes", as this tests no longer exists
- several snaphots have their attribute order changed. This seems to be mostly ordered by key name now.
- update age on the outdated snapshots

- set `SqsManagedSseEnabled` to "true" by default

## TODO

- [x] Clarify if a mock is enough for sse
  - ~~[ ] if not, implement encryption~~

### DIFF
--- a/tests/aws/services/cloudformation/test_template_engine.snapshot.json
+++ b/tests/aws/services/cloudformation/test_template_engine.snapshot.json
@@ -1,6 +1,6 @@
 {
   "tests/aws/services/cloudformation/test_template_engine.py::TestTypes::test_implicit_type_conversion": {
-    "recorded-date": "12-08-2022, 16:14:04",
+    "recorded-date": "29-08-2023, 15:21:22",
     "recorded-content": {
       "queue": {
         "Attributes": {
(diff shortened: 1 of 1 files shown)


baermat:
@thrau As far as I can tell, we do not actually encrypt messages regardless of the state of `SqsManagedSseEnabled` right now. So there should not be any issue if we simply "pretend" it is encrypted by default. The only difference it makes for the user seems to be [anonymous send and receive](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-sqs-sse-queue.html), and this is discouraged by AWS anyway.

thrau (approved):
LGTM! since the tests all pass there doesn't seem to be an issue with other integrations with SQS.

let's keep an eye on issues that may be coming on related to this though. no idea how other client SDKs may react to this :-)

PR #9012, opened 2023-08-29
Revert hypercorn binding to privileged ports

simonrw:
## Motivation

When binding hypercorn on privileged ports, if one of the ports is not available the entire hypercorn server will fail to bind, leaving LocalStack in a terminal failure state. This state is not caught in a user friendly way (the background thread dies leaving a traceback).

## Changes

* This reverts commit 8246af9a3e3a8e1d17ead01932018c985c27aba2.
    * allow hypercorn to only bind on unprivileged ports, and revert to using TCP proxies for unprivileged ports, meaning that at least hypercorn should be able to bind to _something_ even if the proxies fail


alexrashed (approved):
Thanks for jumping on this! 🚀 
It's a pity we have to revert this, but we'll come back to it!

PR #9015, opened 2023-08-29
Add get_docker_host_from_container(..) util to remove DNS operations from config.py

whummer:
## Motivation

We should ideally not be performing I/O operations in the top-level import path, including `config.py`, but rather defer such operations until they are actually required. This has recently surfaced as a reproducible issue in the Testcontainers Go module (https://github.com/testcontainers/testcontainers-go/issues/1495), causing very slow startups under Linux in certain circumstances.

## Changes

* add a `get_docker_host_from_container(..)` utility function to remove DNS operations from `config.py`
* remove top-level constant `DOCKER_HOST_FROM_CONTAINER` from `config.py`
* use the new `get_docker_host_from_container(..)` function from `get_endpoint_for_network(..)` in `container_networking.py`
* add `get_addressable_container_host(..)` utility, which covers a fairly common use case to get the addressable endpoint for a service running inside a container (depending on execution context, running either inside or outside of Docker)

### DIFF
--- a/localstack/utils/net.py
+++ b/localstack/utils/net.py
@@ -8,8 +8,10 @@ from urllib.parse import urlparse
 
 import dns.resolver
 
+from .. import config, constants
 from .collections import CustomExpiryTTLCache
 from .numbers import is_number
+from .objects import singleton_factory
 from .sync import retry
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


localstack/utils/net.py
```diff
@@ -289,3 +291,48 @@ def _try_reserve_port(self, port: IntOrPort, duration: int) -> int:
...
+            if config.LOCALSTACK_HOSTNAME == constants.LOCALHOST:
+                result = "host.docker.internal"
+        # update LOCALSTACK_HOSTNAME if host.docker.internal is available
+        if config.is_in_docker:
+            try:
+                result = socket.gethostbyname("host.docker.internal")
+            except socket.error:
+                result = socket.gethostbyname("host.containers.internal")
+            # TODO still required? - remove
+            # if config.LOCALSTACK_HOSTNAME == config.DOCKER_BRIDGE_IP:
```
whummer, line 320: I _think_ this it is no longer needed and we can remove it, but would be good to also get a quick pair of eyes from @simonrw 👀

localstack/utils/net.py
```diff
@@ -289,3 +291,48 @@ def _try_reserve_port(self, port: IntOrPort, duration: int) -> int:
...
+def get_docker_host_from_container() -> str:
+    """
+    Get the hostname/IP to connect to the host from within a Docker container (e.g., Lambda function).
+    The logic is roughly as follows:
+      1. return `host.docker.internal` if we're running in host mode, in a non-Linux OS
+      2. return the IP address that `host.docker.internal` (or alternatively `host.containers.internal`)
+        resolves to, if we're inside Docker
+      3. return the Docker bridge IP (config.DOCKER_BRIDGE_IP) as a fallback, if option (2) fails
+    """
+    result = config.DOCKER_BRIDGE_IP
```
dfangl, line 306: With a recent ext PR, we added the `ip` util to the docker container. Once this is moved to community, we could also use it to determine the default gateway of the container, which under linux should point to the host. (for docker Desktop we still need the `host.<...>.internal` below, but using the default gateway would be preferable on docker running directly on linux.

dfangl (approved):
LGTM, good to get rid of network connections from config.py.

In the future, we might want to add an option to disable these checks if the internet connection is not available, since the timeouts are now shifted to later in the LS execution.

localstack/utils/net.py
whummer: Great point - looking forward to further iterating on this @dfangl ! 👌

PR #9020, opened 2023-08-30
fix lambda layer version compatibility with Python 3.11

alexrashed:
## Motivation
This PR fixes an issue where Lambda layer compatibility check failed for the `python3.11` runtime:
```
$ awslocal lambda publish-layer-version --layer-name layer1 --zip-file fileb://.serverless/MyLayer.zip --compatible-runtimes python3.11

An error occurred (ValidationException) when calling the PublishLayerVersion operation: 1 validation error detected: Value '[python3.11]' at 'compatibleRuntimes' failed to satisfy constraint: Member must satisfy enum value set: [java17, provided, nodejs16.x, nodejs14.x, ruby2.7, python3.10, java11, python3.11, dotnet6, go1.x, nodejs18.x, provided.al2, java8, java8.al2, ruby3.2, python3.7, python3.8, python3.9]
```

## Changes
This PR adds `python3.11` to the list of supported runtimes for Lambda layer validation.

## Testing
A snapshot verified test creating a Lambda layer with all possible validations.
However, this test does not save us from the same mistake happening again, because it uses the list of `RUNTIMES` which is used by the validation.
Initially, I wanted to use the keys of the `IMAGE_MAPPING` dict (which is described as the dict where newly supported runtimes should be added), but this mapping is contained in `lambda_models` (even though I think it's not a model).
We could move this mapping from there and directly use it for the runtime validation, then this issue could not happen in the future again. Happy for any feedback!

### DIFF
--- a/localstack/services/lambda_/api_utils.py
+++ b/localstack/services/lambda_/api_utils.py
@@ -90,6 +90,7 @@ URL_CHAR_SET = string.ascii_lowercase + string.digits
 # Date format as returned by the lambda service
 LAMBDA_DATE_FORMAT = "%Y-%m-%dT%H:%M:%S.%f+0000"
 
+# An unordered list of all Lambda runtimes supported by LocalStack.
 RUNTIMES = [
     Runtime.nodejs12_x,
     Runtime.nodejs14_x,
@@ -99,6 +100,7 @@ RUNTIMES = [
     Runtime.python3_8,
     Runtime.python3_9,
     Runtime.python3_10,
+    Runtime.python3_11,
     Runtime.ruby2_7,
     Runtime.ruby3_2,
     Runtime.java8,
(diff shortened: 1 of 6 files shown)


joe4dev (approved):
Nice fix, refactoring, and clarifying comments 👍 

nit: Adding a comment pointing to the `IMAGE_MAPPING` (and/or vice-versa) would help to catch this update chore the next time.

I agree that adding new runtimes is currently too much brittle error-prone manual work. Moving forward, the best solution would be to pick up specification-based validations again as suggested by Thomas a while back: https://github.com/localstack/localstack/pull/7675#discussion_r1107777058

Thank you for adding the validated test @alexrashed 🚀

alexrashed:
Thanks, @joe4dev for the review and the feedback! I added comments to both (`RUNTIMES` and `IMAGE_MAPPING`) and also added a simple unit test to ensure the set equality of the `IMAGE_MAPPING` keys and the `RUNTIMES` list.
Let me know what you think :)

joe4dev:
@alexrashed The `IMAGE_MAPPING` and everything around Lambda runtimes is now consolidated and much easier to maintain https://github.com/localstack/localstack/pull/9697

Thanks for the input 👍

PR #9023, opened 2023-08-30
EventBridge: Multi-accounts compatibility

viren-nadkarni:
## Motivation

When using values other than `000000000000` for account ID or `us-east-1` for region, EventBridge would still create the consequent resources in this accounts and region.

## Changes

- Use the correct account ID and region from the request context
- Remove uses of `get_account_id()` and `get_region()` helpers
- Make explicit use of test account IDs and region names in consturction of ARNs and queue URLs


localstack/utils/aws/resources.py
```diff
@@ -72,21 +73,21 @@ def _is_active():
...
     usage_plan_name=None,
-    region_name=None,
     auth_creator_func=None,  # function that receives an api_id and returns an authorizer_id
     client=None,
 ):
     if enabled_api_keys is None:
         enabled_api_keys = []
     if not client:
-        client = connect_to(region_name=region_name).apigateway
+        client = connect_to().apigateway
```
viren-nadkarni, line 90: Temporary change until the `client` arg is made mandatory

localstack/testing/pytest/fixtures.py
```diff
@@ -328,16 +330,6 @@ def sqs_queue(sqs_create_queue):
...
-@pytest.fixture
-def sqs_queue_arn(aws_client):
-    def _get_arn(queue_url: str) -> str:
-        return aws_client.sqs.get_queue_attributes(QueueUrl=queue_url, AttributeNames=["QueueArn"])[
-            "Attributes"
-        ]["QueueArn"]
-
-    return _get_arn
-
-
```
viren-nadkarni, line 340: There was a fixture and a utility (`localstack.utils.aws.arns.sqs_queue_arn`) with the same name that caused confusion and silent overwrite in some instances. The fixture is removed in favour of the utility.

tests/aws/services/events/test_events.py
```diff
@@ -1280,14 +1296,17 @@ def test_put_event_without_detail(self, aws_client_factory):
         assert response.get("Entries")
 
     @markers.aws.unknown
-    def test_trigger_event_on_ssm_change(self, aws_client, clean_up):
+    @pytest.mark.parametrize("strategy", ["domain", "path"])
+    def test_trigger_event_on_ssm_change(self, monkeypatch, aws_client, clean_up, strategy):
+        monkeypatch.setattr(config, "SQS_ENDPOINT_STRATEGY", strategy)
```
dominikschubert, line 1301: :+1:

dominikschubert (approved):
LGTM, though there"s a lot more than eventbridge in there :D

 thanks for tacking this

viren-nadkarni:
The fixture has been restored due to concerns from Daniel that it will hinder the ability to run all tests against AWS. It's been renamed to `sqs_get_queue_arn`

PR #9026, opened 2023-08-30
fix proxy Accept-Header handling, combine http clients

### LINKED ISSUE #8793: bug: OpenSearch response with compressed content

### Current Behavior

While running `curl` to reach OpenSearch health endpoint, I receive a binary response instead of plain JSON.

```shell
$ curl -vvv my-domain.us-east-1.opensearch.localhost.localstack.cloud:443/_cluster/health                                                                                                                                                                                                                               
*   Trying 127.0.0.1:443...
* Connected to my-domain.us-east-1.opensearch.localhost.localstack.cloud (127.0.0.1) port 443 (#0)
> GET /_cluster/health HTTP/1.1
> Host: my-domain.us-east-1.opensearch.localhost.localstack.cloud:443
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 200
< content-type: application/json; charset=UTF-8
< content-encoding: gzip
< Connection: close
< date: Tue, 01 Aug 2023 08:46:35 GMT
< server: hypercorn-h11
< Transfer-Encoding: chunked
<
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
* Failure writing output to destination
* Failed reading the chunked-encoded stream
* Closing connection 0
```

### Expected Behavior

I expect to receive a JSON response without the need to use `--compressed` option in the `curl` command.

 curl -vvv --compressed my-domain.us-east-1.opensearch.localhost.localstack.cloud:443/_cluster/health
*   Trying 127.0.0.1:443...
* Connected to my-domain.us-east-1.opensearch.localhost.localstack.cloud (127.0.0.1) port 443 (#0)
> GET /_cluster/health HTTP/1.1
> Host: my-domain.us-east-1.opensearch.localhost.localstack.cloud:443
> User-Agent: curl/7.81.0
> Accept: */*
> Accept-Encoding: deflate, gzip, br, zstd
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200
< content-type: application/json; charset=UTF-8
< content-encoding: gzip
< Connection: close
< date: Tue, 01 Aug 2023 08:47:06 GMT
< server: hypercorn-h11
< Transfer-Encoding: chunked
<
* Closing connection 0
{"cluster_name":"opensearch","status":"green","timed_out":false,"number_of_nodes":1,"number_of_data_nodes":1,"discovered_master":true,"discovered_cluster_manager":true,"active_primary_shards":0,"active_shards":0,"relocating_shards":0,"initializing_shards":0,"unassigned_shards":0,"delayed_unassigned_shards":0,"number_of_pending_tasks":0,"number_of_in_flight_fetch":0,"task_max_waiting_in_queue_millis":0,"active_shards_percent_as_number":100.0}

With a docker-compose file

### Steps To Reproduce

```yaml
version: '3.8'
services:
  localstack:
    image: localstack/localstack-pro:latest
    ports:
      - "127.0.0.1:53:53"
      - "127.0.0.1:53:53/udp"
      - "127.0.0.1:443:443"
      - "127.0.0.1:4510-4559:4510-4559"
      - "127.0.0.1:4566:4566"
    environment:
      - DEBUG=1
      - LS_LOG=trace
      - LOCALSTACK_API_KEY=${LOCALSTACK_API_KEY-} # only required for Pro
      - OPENSEARCH_ENDPOINT_STRATEGY=domain
    volumes:
      - "./volume:/var/lib/localstack" # mount Docker volume
    healthcheck:
      disable: true
```

```shell
awslocal opensearch create-domain --domain-name my-domain
curl -vvv my-domain.us-east-1.opensearch.localhost.localstack.cloud:443/_cluster/health          
```

### Environment

```markdown
- OS: Windows 10
- LocalStack: 2.2.1.dev20230731233910
```


### Anything else?

Following the guide from the documentation page: [OpenSearch](https://docs.localstack.cloud/user-guide/aws/opensearch/)

alexrashed:
## Motivation
This PR addresses #8793, which describes an issue with the handling of the `Accept-Encoding` header when communicating with OpenSearch clusters with enabled http compression (as enabled statically with #8628).
This bug is actually caused by our HTTP client used in the Proxy infrastructure used to proxy requests from the client via LocalStack to the started OpenSearch server. This HTTP client _always_ adds `Accept-Encoding: gzip, deflate` to the proxied HTTP request if no `Accept-Encoding` is set in the originating request.

With enabling the compression support for OpenSearch (i.e. OpenSearch will compress the responses using GZIP `Content-Encoding` in responses if the client sends `gzip` in the `Accept-Encoding` request header), this issue surfaced since every request to OpenSearch via LocalStack (which is the default) without the `Accept-Encoding` header (which is the default f.e. for a plain `curl` request).

It turns out that sending HTTP requests without any `Accept-Encoding` header is not possible with `requests` because it uses `urllib3` under the hood, which in turn uses Python's `http.client` standard lib:
- `urllib3` automatically adds `` if the `Accept-Encoding` header is not set at all. See [`urllib3.util.request.ACCEPT_ENCODING`](https://github.com/urllib3/urllib3/blob/23799f263998ac138d569ed92635df8b09278ed0/src/urllib3/util/request.py#L21).
  - This could be circumvented by setting `Accept-Encoding` explicitly to `None`
- BUT: One layer below - Python's `http.client` - sets the `Accept-Encoding: identity` in case it's not set or `None`.
  - However, as discussed in https://github.com/psf/requests/issues/2234, this behavior should be compliant to the specs / RFCs.

This means it's not really possible to identically relay a request which does not contain the `Accept-Encoding` header at all.

## Changes
- This PR explicitly sets the `Accept-Encoding` header to `identity` in case it's not set or `None`.
  - An alternative would have been just to circumvent the behavior of `urllib3` by setting the `Accept-Encoding` header to `None`, but I wanted to make this behavior as explicit as possible (because it was quite a pain to get to the root of this weird behavior).
- This PR also merges the `SimpleHttpClient` with the `SimpleStreamingHttpClient`. The latter was introduced with #8148 in addition to the `SimpleHttpClient` where it should actually just have been an extension of the default (but that was understandable at the time in order to prevent too wide-ranging changes).
  - This is why I classified this change as "minor".

## Testing
Testing this in Python is actually quite hard (since the test explicitly has to send a request without any `Accept-Encoding` header whatsoever). I actually had to directly use the `HTTPConnection` class of Python's std lib.

Fixes #8793.

### DIFF
--- a/localstack/http/client.py
+++ b/localstack/http/client.py
@@ -81,10 +81,6 @@ class SimpleRequestsClient(HttpClient):
         Very naive implementation to make the given HTTP request using the requests library, i.e., process the request
         as a client.
 
-        TODO: over time, this should become more sophisticated, specifically the use of restore_payload should only be
-         used when necessary (when the underlying stream has been consumed), and by default the stream should be
-         streamed to the destination.
-
         :param request: the request to perform
         :param server: the URL to send the request to, which defaults to the host component of the original Request.
         :return: the response.
@@ -92,43 +88,17 @@ class SimpleRequestsClient(HttpClient):
 
         url = self._get_destination_url(request, server)
 
-        response = self.session.request(
-            method=request.method,
-            # use raw base url to preserve path url encoding
-            url=url,
-            # request.args are only the url parameters
-            params=[(k, v) for k, v in request.args.items(multi=True)],
-            headers=dict(request.headers.items()),
-            data=restore_payload(request),
-        )
+        headers = dict(request.headers.items())
 
-        final_response = Response(
-            response=response.content,
-            status=response.status_code,
-            headers=Headers(dict(response.headers)),
-        )
-        if request.method == "HEAD":
-            # for HEAD  requests we have to keep the original content-length, but it will be re-calculated when creating
-            # the final_response object
-            final_response.content_length = response.headers.get("Content-Length", 0)
-        return final_response
-
-    def close(self):
-        self.session.close()
-
-
-class SimpleStreamingRequestsClient(SimpleRequestsClient):
-    def request(self, request: Request, server: str | None = None) -> Response:
-        """
-        Very naive implementation to make the given HTTP request using the requests library, i.e., process the request
-        as a client.
-
-        :param request: the request to perform
-        :param server: the URL to send the request to, which defaults to the host component of the original Request.
-        :return: the response.
-        """
-
-        url = self._get_destination_url(request, server)
+        # urllib3 (used by requests) will set an Accept-Encoding header ("gzip,deflate")
+        # - See urllib3.util.request.ACCEPT_ENCODING
+        # - This could be circumvented by setting Accept-Encoding explicitly to None
+        # BUT: Python's http client (used by urllib3) will always add an accept header if none is given
+        # - See https://github.com/psf/requests/issues/2234 and http.client.putrequest
+        # Explicitly set `Accept-Encoding: identity` here if no `Accept-Encoding` header is set in the originating
+        # request to avoid any unused manipulations by underlying libraries.
+        if not request.headers.get("accept-encoding"):
+            headers["accept-encoding"] = "identity"
 
         response = self.session.request(
             method=request.method,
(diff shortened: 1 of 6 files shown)


thrau (approved):
nice find! LGTM! and thanks for cleaning up :broom:

PR #9029, opened 2023-08-30
enable DomainEndpoint cfn output for Opensearch

steffyP:
## Motivation
Follow-up for https://github.com/localstack/localstack/pull/9027. 
The `DomainEndpoint` seems to be a commonly used [return value for Opensearch Domain](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opensearchservice-domain.html#aws-resource-opensearchservice-domain-return-values).

## Changes
Added support for output of `DomainEndpoint`.


alexrashed (approved):
Thanks for jumping on this! The changes are looking good, the test has been extended, and the addition will be very useful! 💯 🚀

PR #9033, opened 2023-08-31
Don't run in privileged mode

simonrw:
## Motivation

We should not be running the container in privileged mode if we don't need to.

## Changes

Do not run the LocalStack container in privileged mode (from the CLI at least).


dfangl (approved):
That was still in there?

alexrashed (approved):

PR #9034, opened 2023-08-31
remove extensions venv link from Dockerfile

thrau:
## Motivation

@simonrw noticed that when you first install an extensions using localstack pro, and then go back to using localstack community, there will be wonkyness because the extensions/python_venv contains a copy of localstack-ext.

We don't want to do this linking of venvs in community in the first place, so removing it here. This should be safe since we have this explicitly in the ext Dockerfile.

## Changes

* remove the extensions venv linking from the community Dockerfile


alexrashed (approved):
Good catch! 🚀 💯

PR #9035, opened 2023-08-31
add Cfn ResourceProvider for AWS::OpenSearchService::Domain

steffyP:
## Motivation
Working with the Cfn and `opensearch` revealed a few shortcomings the past days. Already fixed some issues in the old provider, but it might be time to move the new `ResourceProvider` now 🙂 

## Changes
* set the `opensearch` ResourceProvider as default for Cfn
* added logic to the `ResourceProvider` in a pairing session with @dominikschubert 
* enhanced test case a little (include `Tags`)

### DIFF
--- a/localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
+++ b/localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
@@ -202,8 +202,6 @@ class OpenSearchServiceDomainProvider(ResourceProvider[OpenSearchServiceDomainPr
 
         """
         model = request.desired_state
-        if tags := model.pop("Tags", []):
-            model["TagList"] = tags
         opensearch_client = request.aws_client_factory.opensearch
         if not request.custom_context.get(REPEATED_INVOCATION):
             # resource is not ready
@@ -224,7 +222,12 @@ class OpenSearchServiceDomainProvider(ResourceProvider[OpenSearchServiceDomainPr
                 # set defaults required for boto3 calls
                 cluster_config.setdefault("DedicatedMasterType", "m3.medium.search")
                 cluster_config.setdefault("WarmType", "ultrawarm1.medium.search")
-            opensearch_client.create_domain(**properties)
+
+            create_kwargs = {
+                **util.deselect_attributes(properties, ["Tags"]),
+                "TagList": properties.get("Tags"),
+            }
+            opensearch_client.create_domain(**create_kwargs)
             return ProgressEvent(
                 status=OperationStatus.IN_PROGRESS,
                 resource_model=model,
--- a/localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
+++ b/localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
@@ -180,6 +180,8 @@ class OpenSearchServiceDomainProvider(ResourceProvider[OpenSearchServiceDomainPr
         Primary identifier fields:
           - /properties/DomainName
 
+
+
         Create-only properties:
           - /properties/DomainName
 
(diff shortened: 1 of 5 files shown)


localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
```diff
@@ -200,41 +202,51 @@ def create(
 
         """
         model = request.desired_state
-
-        # Validations
-        assert model["DomainName"]
-
+        if tags := model.pop("Tags", []):
+            model["TagList"] = tags
```
dominikschubert, line 206:
```suggestion
        create_kwargs = {
            **util.deselect_attributes(model, ["Tags"]),
            "TagList": model.get('Tags')
        }
```

I'd also suggest moving this farther down to just before the `create_domain` call

In the future we'll be quite strict here and this would fail since it doesn't conform to the resource schema.
Only properties defined in the `OpenSearchServiceDomainProperties` typed dict should be set on the model.
Pycharm should've actually also complained about this :thinking:

localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
```diff
@@ -263,28 +275,9 @@ def delete(
...
-            return ProgressEvent(
-                status=OperationStatus.SUCCESS, resource_model=request.desired_state
-            )
-        except request.aws_client_factory.opensearch.exceptions.ResourceNotFoundException:
-            return ProgressEvent(
-                status=OperationStatus.SUCCESS, resource_model=request.desired_state
-            )
+        opensearch_client = request.aws_client_factory.opensearch
+        opensearch_client.delete_domain(DomainName=request.previous_state["DomainName"])
+        return ProgressEvent(status=OperationStatus.SUCCESS, resource_model={})
```
dominikschubert, line 280: Is the delete operation on opensearch synchronous? If not we need the same pattern as in create here.

dominikschubert (changes requested):

simonrw:
> Working with the Cfn and opensearch revealed a few shortcomings the past days. Already fixed some issues in the old provider, but it might be time to move the new ResourceProvider now 🙂

:tada:

tests/aws/services/cloudformation/resources/test_opensearch.py
```diff
@@ -35,7 +36,15 @@ def test_domain(deploy_cfn_template, aws_client, snapshot):
     result = deploy_cfn_template(template_path=template_path)
     domain_endpoint = result.outputs["SearchDomainEndpoint"]
     assert domain_endpoint
+    domain_arn = result.outputs["SearchDomainArn"]
+    assert domain_arn
```
simonrw, line 40: worth snapshotting the `result.outputs`?

simonrw (approved):
Looks great, thanks for doing this migration! I'm happy once you've implemented @dominikschubert's suggestions.

localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
steffyP: yes, seems to be synchronous.

localstack/services/opensearch/resource_providers/aws_opensearchservice_domain.py
```diff
@@ -200,41 +202,51 @@ def create(
...
+                cluster_config.setdefault("DedicatedMasterType", "m3.medium.search")
+                cluster_config.setdefault("WarmType", "ultrawarm1.medium.search")
+            opensearch_client.create_domain(**properties)
+            return ProgressEvent(
+                status=OperationStatus.IN_PROGRESS,
+                resource_model=model,
+                custom_context=request.custom_context,
+            )
+        opensearch_domain = opensearch_client.describe_domain(DomainName=model["DomainName"])
+        if opensearch_domain["DomainStatus"]["Processing"] is False:
```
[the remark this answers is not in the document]
steffyP: from the implementation it seems like it currently always returns `Created: true`.

tests/aws/services/cloudformation/resources/test_opensearch.py
steffyP: don't think so :) there would only be placeholders/replacements in the snapshot. 
We test the arn anyhow when retrieving the tags, and the domain-name for the describe. Those two responses are snapshotted.

Issue #9032, opened 2023-08-31
bug: UploadPart & PutObject Return Non-Empty Body


RobertLeahy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

UploadPart returns a non-empty body containing the following vacuous XML:

```
<?xml version='1.0' encoding='utf-8'?>
<UploadPartOutput />
```

PutObject has a similar issue:

```
<?xml version='1.0' encoding='utf-8'?>
<PutObjectOutput />
```

### Expected Behavior

UploadPart and PutObject should return an empty body as documented:

- https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run -d --rm --name=rleahy-localstack-2 localstack/localstack:2.2.0

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

Attempting to UploadPart via the libs3 C API

### Environment

- OS: Ubuntu 22.04
- LocalStack: 2.2.0

### Anything else?

The error in libs3 is raised here: https://github.com/bji/libs3/blob/287e4bee6fd430ffb52604049de80a27a77ff6b4/src/request.c#L279

libs3 succeeds when performing UploadPart and PutObject against Amazon S3 (and LocalStack 1.1.0).

Removing/modifying/avoiding the linked line (through modification to libs3) allows UploadPart and PutObject to succeed against LocalStack 2.

RobertLeahy:
This also seems to apply to PutObject, I have updated the title and initial issue description accordingly.

bentsku:
Hi @RobertLeahy and thank for your report! It seems the issue might be there for many API calls using the `PUT` method. Most of them should not return a body, but there are some exceptions like `CopyObject` for example. I'll be working on a fix and you should be updated soon on the progress. Thanks!

PR #9036, opened 2023-08-31
fix S3 PUT operations returning HTTP Body by default

bentsku:
## Motivation
As reported in #9032, some S3 operations (the one using the `PUT` http method) would return an empty XML body when the S3 docs specifies that the body should be empty. After looking at the specs, only `CopyObject` and `CopyObjectPart` would return a body. Luckily, those operations implement a field specifying that they return a body, the `payload` one.

## Changes
Removed the pre-signed handler deleting the response body, since this behaviour should have been for every `PUT` request except some, and it was actually fixing a symptom instead of the cause.

Implemented a check in the serializer for not serializing a body if the conditions were not set (if it's a PUT request without the `payload` field). 

Added 2 cases in a test checking the raw returned body without boto parsing, one for a known PUT operation which should return a body (`CopyObject`) and one for `UploadPart`, which should not return a body.

### DIFF
--- a/localstack/aws/protocol/serializer.py
+++ b/localstack/aws/protocol/serializer.py
@@ -1480,8 +1480,15 @@ class S3ResponseSerializer(RestXMLResponseSerializer):
         self._process_header_members(header_params, response, shape)
         # "HEAD" responses are basically "GET" responses without the actual body.
         # Do not process the body payload in this case (setting a body could also manipulate the headers)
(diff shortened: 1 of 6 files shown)


alexrashed (approved):
Nice! Really great to see how much indirections can be removed with just handling that one condition on a global level in the serializer! 🚀

PR #9039, opened 2023-08-31
fix APIGW UpdateStage with /tracingEnabled

bentsku:
## Motivation
A user reported a stuck Terraform deployment with an API GW REST API: Terraform would loop and never finish while trying to update a stage when `xray_tracing_enabled` is set. 

When setting `TF_LOG=trace`, we see there's an issue when unmarshalling the response, because we were returning a string `"true"` instead of a boolean `True`.
```
DEBUG: Unmarshal Response apigateway/UpdateStage failed, attempt 9/25, error SerializationError: failed decoding JSON RPC response
```

Add a case to a snapshot test to validate this behaviour as well. 

## Changes

Added a check if the path is `/tracingEnabled` to cast the value into a boolean.
Also used `copy.deepcopy` to the input, because we were modifying it in place, and the logging would show the modified input instead of what we really received. 

## Testing

I could reproduce the issue with the following stage resource:
```hcl
resource "aws_api_gateway_rest_api" "api" {
  name           = "rest"
  api_key_source = "HEADER"
}

resource "aws_api_gateway_deployment" "deployment" {
  rest_api_id = aws_api_gateway_rest_api.api.id

  lifecycle {
    create_before_destroy = true
  }
}

resource "aws_api_gateway_stage" "stage" {
  stage_name    = "dev"
  rest_api_id   = aws_api_gateway_rest_api.api.id
  deployment_id = aws_api_gateway_deployment.deployment.id
  depends_on    = [aws_cloudwatch_log_group.test]
  variables = {
    "TEST_ENV" = "local"
  }
  xray_tracing_enabled = "true"

  access_log_settings {
    destination_arn = aws_cloudwatch_log_group.test.arn
    format = "$context.identity.sourceIp $context.identity.caller $context.identity.user [$context.requestTime] \"$context.httpMethod $context.resourcePath $context.protocol\" $context.status $context.responseLength $context.requestId"
  }
}

resource "aws_cloudwatch_log_group" "test" {
  name = "Yada"

  tags = {
    Environment = "production"
    Application = "serviceA"
  }
}

```

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -873,7 +873,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         # construct list of path regexes for validation
         path_regexes = [re.sub("{[^}]+}", ".+", path) for path in STAGE_UPDATE_PATHS]
 
-        patch_operations = patch_operations or []
+        # copy the patch operations to not mutate them, so that we're logging the correct input
+        patch_operations = copy.deepcopy(patch_operations) or []
         for patch_operation in patch_operations:
             patch_path = patch_operation["path"]
             path_valid = patch_path in STAGE_UPDATE_PATHS or any(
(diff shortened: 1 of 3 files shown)


whummer (approved):
Great catch @bentsku , nice detective work with `TF_LOG=trace` ! 🔍

calvernaz (approved):
Thanks @bentsku! LGTM

PR #9043, opened 2023-09-01
replace CircleCI job for S3 stream to v3 and remove stream provider

bentsku:
## Motivation
Now that the S3 v3 provider is ready, I'd like to replace the CircleCI job for the S3 stream provider to the new native.
We also should remove the stream provider (it didn't work with persistence anyway, so no backward compatibility issue), we can make `PROVIDER_OVERRIDE_S3=stream` load the new provider as well. 

## Changes
Replacing the job and make it necessary to pass for the build to be fully green (it wasn't a necessary step for pushing before, to avoid flakiness for such a small subset). 

Removing the `stream` provider patching and make it point to the `v3` provider.

Also a small fix regarding a refactoring in the stream provider for the `AwsChunkedDecoder` which hadn't been ported to the v3 provider (and no tests to show it 😅)


alexrashed (approved):
Removing 890 LoC? :shipit: 😆 
No, really, great to see the S3 provider evolving at top speed! 🚀

PR #9044, opened 2023-09-01
Bump moto-ext to 4.2.0.post1

viren-nadkarni:
## Motivation

Bring in the latest Moto changes.

## Changes

Stop using the `clean_key_name` that was removed in https://github.com/getmoto/moto/pull/6602/

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -1867,7 +1867,7 @@ def apply_moto_patches():
         """
         return get_safe(self.querystring, "$.x-id.0") == "DeleteObjects" or fn(self)
 
-    @patch(moto_s3_responses.S3ResponseInstance.parse_bucket_name_from_url, pass_target=False)
+    @patch(moto_s3_responses.S3Response.parse_bucket_name_from_url, pass_target=False)
     def parse_bucket_name_from_url(self, request, url):
         """
         Requests going to moto will never be subdomain based, as they passed through the VirtualHost forwarder.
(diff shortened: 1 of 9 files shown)


localstack/services/s3/provider.py
```diff
@@ -1874,7 +1874,7 @@ def s3_response_is_delete_keys(fn, self):
         """
         return get_safe(self.querystring, "$.x-id.0") == "DeleteObjects" or fn(self)
 
-    @patch(moto_s3_responses.S3ResponseInstance.parse_bucket_name_from_url, pass_target=False)
+    @patch(moto_s3_responses.S3Response.parse_bucket_name_from_url, pass_target=False)
```
bentsku, line 1877: this should have been a sign from the big issue this bump is fixing, but well 🤷‍♂️ fixing the patch to properly work

localstack/services/secretsmanager/provider.py
```diff
@@ -314,23 +314,6 @@ def fake_secret_update(
         self.last_changed_date = time.time()
 
 
-class FakeSecretVersionStore(dict):
```
bentsku, line 317: I've ran some secretsmanager tests and it seems they run without the patch now, I think it has been fixed upstream with https://github.com/getmoto/moto/pull/6720

bentsku:
Seems like there's still an issue with `Logs`:
```
  File "/opt/code/localstack/localstack/services/logs/provider.py", line 318, in moto_put_log_events
    if events and self.destination_arn:
AttributeError: 'LogStream' object has no attribute 'destination_arn'
```
I'll take a look later tonight. I think a lot of these issue stems from this, because most of the failing test are test that have a logs integration (Lambda, SNS, SQS...)

bentsku:
Only `test_last_updated_date` failing because of difference in precision between timestamp in moto and our provider (moto round `time.time()` but secretsmanager should return 3 fraction digits). Adapted the test to check for relative equality instead of strict. The pipeline should now be green 🤞

bentsku:
Green -ext run on this branch: https://github.com/localstack/localstack-ext/actions/runs/6054306633

PR #9045, opened 2023-09-01
Set up acceptance test marker & CI job

dominikschubert:
## Motivation
In the future we want to experiment more with using only a subset of our integration tests as a sort of acceptance test suite.
Since we want to start early to get a feeling for how well this will work and what we need to look for in potential acceptance test candidates, we're now already starting with a very small subset of tests that are only executed in parallel to the current CI test jobs.

## Changes

- introduce a new pytest marker `acceptance_test`
  - initially this is exposed via our markers as `@markers.acceptance_tests_beta` to communicate more clearly that these tests aren't yet used as the only acceptance tests. (to avoid confusion)
- creates a job in CircleCI to execute all tests 
  - tests are currently only executed on the default architecture (amd64) 
  - as discussed previously, the tests form part of the gate for pushing the built images
  - the selected tests marked by `acceptance_test` are now being executed more than once during the total CI run since we don't deselect the marker yet in our "normal" test suite.
- mark one test as an acceptance test, so we have at least something already in there :sweat_smile: 
  - in the following week we'll constantly add more of these markers where we deem them to be appropriate
- currently only executes tests in `tests/aws/`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,7 +163,7 @@ jobs:
       - run:
           name: Acceptance tests (EXPERIMENTAL)
           environment:
-            TEST_PATH: "tests/aws/services/stepfunctions/v2/"
+            TEST_PATH: "tests/"
             COVERAGE_ARGS: "-p"
           command: |
             PYTEST_ARGS="${TINYBIRD_PYTEST_ARGS}--reruns 3 -m acceptance_test --junitxml=target/reports/acceptance_test.xml -o junit_suite_name='acceptance_test'" make test-coverage
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,6 +153,31 @@ jobs:
       - store_test_results:
           path: target/reports/
 
+  acceptance-tests:
+    executor: ubuntu-machine-amd64
+    working_directory: /tmp/workspace/repo
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - prepare-pytest-tinybird
+      - run:
+          name: Acceptance tests (EXPERIMENTAL)
+          environment:
+            TEST_PATH: "tests/aws/services/stepfunctions/v2/"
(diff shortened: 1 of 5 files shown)


steffyP:
Nice 🚀 
But why were the tests skipped in the run? 🤔 
```
tests/aws/scenario/test_ecs_scenario.py::TestEcsScenario::test_scenario_validate_infra SKIPPED
tests/aws/scenario/test_ecs_scenario.py::TestEcsScenario::test_scenario_call_service SKIPPED
```

dominikschubert:
> Nice 🚀 But why were the tests skipped in the run? 🤔
> 
> ```
> tests/aws/scenario/test_ecs_scenario.py::TestEcsScenario::test_scenario_validate_infra SKIPPED
> tests/aws/scenario/test_ecs_scenario.py::TestEcsScenario::test_scenario_call_service SKIPPED
> ```

the test case is marked with 
```
@pytest.mark.skip(
    reason="requires pro",
)
```

Was probably not the best one to start marking with the new marker, that's why I added another one later on :grimacing:

alexrashed (approved):
Awesome! Thanks for jumping on this!
By introducing this additional parallel step, we can nicely monitor the planned acceptance tests (their execution time, stability, and reliability) such that we can grow experience and confidence in them before deciding to switch! 
I only found 3 small things:
- As @steffyP mentioned, currently no test is executed (because the ones currently marked are skipped).
- The logs in the test execution contains an "address already bound" exception log right after printing the pytest result (could have been a regression on master at the time of creating the PR, but maybe there's something wrong with the fixtures - like multiple instances are tried to be started?).
- It would be great to check in a successful run that the coverage results are successfully archived. They will be quite valuable in the future. 🚀

steffyP (approved):
Nice work 🚀 😄 
Looking forward to working to our acceptance-test-suite 🙂

PR #9046, opened 2023-09-01
move container fixture into testing package for reusability

thrau:
## Motivation

To be able to re-use the container fixtures introduced in #8781 in localstack-ext, they need to be in the localstack source folder.
I totally missed this during the review, sorry!

## Changes

* move containerfactory and fixtures into `localstack/testing/pytest`
* import them as plugins in `tests/bootstrap/conftest.py`


simonrw (approved):
Thanks, I needed this in my feature branch shortly after merging the original PR!

dominikschubert (approved):

PR #9048, opened 2023-09-01
Container fixture improvements

simonrw:
## Motivation

The container fixtures created in ... were:

* only usable in the `bootstrap` subdir, but we need them in the `cli` subdir (amongst others)
* are too widely scoped (session scoped `ContainerFactory`)
* uses a confusing fixture order (cleanups)
* did not store the running container, so automatic cleanup was not possible

## Changes

Fix the items above

### DIFF
--- a/localstack/testing/pytest/container.py
+++ b/localstack/testing/pytest/container.py
@@ -1,7 +1,7 @@
 import logging
 import os
 import shlex
-from typing import Generator, List
+from typing import Generator, List, Optional
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


localstack/testing/pytest/container.py
```diff
@@ -86,7 +83,7 @@ def remove_all_containers(self):
                 )
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
```
simonrw, line 86: This or class scope?

thrau (approved):
awesome! thanks for fixing :100:

PR #9049, opened 2023-09-01
Migrate DNS server to community

simonrw:
## Motivation

We want to simplify network configuration. A point of confusion is that `localhost.localstack.cloud` is publically registered, and resolves to `127.0.0.1`. When communicating between docker containers, this is an issue - trying to connect to e.g. an opensearch domain with url `*.localhost.localstack.cloud` returned by LocalStack will attempt to connect to the "application" container, rather than LocalStack.

Recently we have upgraded the DNS server provided by LocalStack Pro to dynamically resolve the IP address of the LocalStack container based on request subnet, but this functionality was only available in Pro.

With this PR we bring the DNS server from Pro to LocalStack community. This change brings along with it increased test coverage. 

## Changes

* Move the DNS server to community
* Add bootstrap tests covering the DNS server behaviour
* Start the DNS server regardless of whether the host port can be bound
* If port 53 (or `DNS_PORT`) can be bound on the host, publish the docker port using the CLI
* Move the DNS server module from `localstack.services.dns_server` to `localstack.dns.server` to group core resources (e.g. `http`, `dns`, etc.) at the top of the tree
* Remove broken attempts to configure the DNS of the host

## Testing

* Start LocalStack: `DNS_ADDRESS=127.0.0.1 localstack start`
* If port 53/udp is available on the host, then run `dig @127.0.0.1 localhost.localstack.cloud` and a valid response is returned
* Start a container in the same docker network/subnet as the LocalStack container, then run from that container `dig @<ip of localstack container> localhost.localstack.cloud` and the IP address of the container should be returned
* Remove broken attempts to configure the host DNS

## TODO

What's left to do:

- [x] Get -ext PR ready to merge


simonrw:
podman tests have been broken for a while

localstack/config.py
```diff
@@ -1051,6 +1051,31 @@ def legacy_fallback(envar_name: str, default: T) -> T:
 # e.g. CFN_RESOURCE_PROVIDER_OVERRIDES='{"AWS::Lambda::Version": "GenericBaseModel","AWS::Lambda::Function": "ResourceProvider"}'
 CFN_RESOURCE_PROVIDER_OVERRIDES = os.environ.get("CFN_RESOURCE_PROVIDER_OVERRIDES", "{}")
 
+# bind address of local DNS server
+DNS_ADDRESS = os.environ.get("DNS_ADDRESS") or "0.0.0.0"
```
dfangl, line 1055: I wonder if we should not default to 127.0.0.1 here. So we do not expose it to the whole network without need.

localstack/utils/bootstrap.py
```diff
@@ -733,6 +756,25 @@ def __enter__(self):
...
+        Optionally specify the docker network
+        """
+        # TODO: podman may not have this information
+        inspect_result = self.container_client.inspect_container(self.id)
+        if docker_network is None:
+            return list(inspect_result["NetworkSettings"]["Networks"].values())[0]["IPAddress"]
+        else:
+            for network, details in inspect_result["NetworkSettings"]["Networks"].items():
+                if network == docker_network:
+                    return details["IPAddress"]
```
dfangl, line 772: In the container client, there is `get_container_ip` and `get_container_ip_for_network` already. We should use that utilities here instead

localstack/dns/server.py
```diff
@@ -0,0 +1,875 @@
...
+
+SKIP_PATTERNS = {
+    r".*(forums|console|docs|clientvpn|sso|boto3|(signin(\-reg)?))\.([^\.]+\.)?(aws\.amazon|amazonaws)\.com",
+    r".*captcha-prod\.s3\.amazonaws\.com",
+    r"^aws\.amazon\.com",
+    r"^github-production-release-.*\.s3\.amazonaws\.com",
+    r"^aws-glue-etl-artifacts\.s3\.amazonaws\.com",
+    rf"^{S3_ASSETS_BUCKET}\.s3\.amazonaws\.com",
+    r"^localstack-pods-.*\.s3\.amazonaws\.com",
+}
```
dfangl, line 256: I would argue if this should be here. Since we will add the transparent endpoint injection in -ext, this might fit better there as well. By default, we would not redirect these entries anyway.

localstack/dns/server.py
```diff
@@ -0,0 +1,875 @@
...
+    for name in NAME_PATTERNS_POINTING_TO_LOCALSTACK:
+        dns_server.add_host_pointing_to_localstack(name)
+    for skip_pattern in SKIP_PATTERNS:
+        dns_server.add_skip(skip_pattern)
+    if config.DNS_LOCAL_NAME_PATTERNS:
+        for skip_pattern in re.split(r"[,;\s]+", config.DNS_LOCAL_NAME_PATTERNS):
+            dns_server.add_skip(skip_pattern)
+    dns_server.start()
+    if not dns_server.wait_is_up(timeout=5):
+        LOG.warning("DNS server did not come up within 5 seconds.")
```
dfangl, line 807: In this case, we might want to do a `dns_server.shutdown()` to avoid leaking that server.

localstack/utils/docker_utils.py
```diff
@@ -132,13 +132,16 @@ def get_host_path_for_path_in_docker(path):
     return path
 
 
-def container_ports_can_be_bound(ports: Union[IntOrPort, List[IntOrPort]]) -> bool:
+def container_ports_can_be_bound(
+    ports: Union[IntOrPort, List[IntOrPort]],
+    address: Optional[str] = None,
+) -> bool:
```
dfangl, line 138: Good change!

dfangl (commented):
Already looks good, just some points for discussion

localstack/config.py
simonrw: Makes sense. We already default `GATEWAY_LISTEN` to `127.0.0.1` if not in docker and `0.0.0.0` otherwise (using `localstack.config.default_ip`). I can implement that.

dfangl (approved):
Thanks for adding the CODEOWNERS file, forgot to mention that. LGTM, great work!

localstack/utils/bootstrap.py
```diff
@@ -457,6 +457,28 @@ def _cfg(cfg: ContainerConfiguration):
...
+                docker_flags.extend(
+                    [
+                        "-p",
+                        f"{config.DNS_ADDRESS}:{port.port}:{port.port}/{port.protocol}",
+                    ]
+                )
+            if cfg.additional_flags is None:
+                cfg.additional_flags = " ".join(docker_flags)
+            else:
+                cfg.additional_flags += " ".join(docker_flags)
```
dfangl, line 480: When appending, we need to append a space in between as well!

dfangl (changes requested):

localstack/config.py
dfangl: Then we could also skip the override in dns_server.py, which would be really nice!

localstack/utils/bootstrap.py
simonrw: Yeah thanks, just saw that in the test failure

dfangl (approved):
🎉

PR #8994, opened 2023-08-26
add dev container run script

thrau:
## Motivation

This PR introduces a script for developers that aims to replace `bin/docker-run.sh` and ultimately all its derivatives.

We have several ways of starting what I call a "dev container", i.e., a localstack container with various mounted things like source folders or dependencies. Here are some:

* `bin/localstack-start-docker-dev.sh`
* `make docker-mount-run`
* `make test-docker-mount`
* `make docker-run-tests`
* `test-docker-mount-code`

In ext:
* `bin/docker-run.sh`
* `bin/test-in-docker.sh`
* `make docker-mount-run`
* ...

With the introduction of #8781 and #9050, there was an opportunity to unify some of the code necessary to start and configure such a dev container.

Moreover, with the growing amount of bootstrap tests (network configs, extensions, persistence, ...) there was a need to run these tests locally without having to build the image every time. So whatever configuration logic needed for a `docker-run.sh` replacement could be used there as well.

## Usage

This is what it looks like when you run a localstack container. The only main difference to `localstack start` is that it also mounts the source folders by default. The command also outputs the container config in a more human readable way:

Screenshot at 2023-09-04 00-32-55

Here's the help:

```console
 % python -m localstack.dev.run --help                     
Usage: python -m localstack.dev.run [OPTIONS] [COMMAND]...

  A tool for localstack developers to start localstack containers. Run this in
  your localstack or localstack-ext source tree to mount local source files or
  dependencies into the container. Here are some examples::

      python -m localstack.dev.run
      python -m localstack.dev.run -e DEBUG=1 -e LOCALSTACK_API_KEY=test
      python -m localstack.dev.run -- bash -c 'echo "hello"'

  Explanations and more examples:

  Start a normal container localstack container. If you run this from the
  localstack-ext repo, it will start localstack-pro::

      python -m localstack.dev.run

  If you start localstack-pro, you might also want to add the API KEY as
  environment variable::

      python -m localstack.dev.run -e DEBUG=1 -e LOCALSTACK_API_KEY=test

  If your local changes are making modifications to plux plugins (e.g., adding
  new providers or hooks), then you also want to mount the newly generated
  entry_point.txt files into the container::

      python -m localstack.dev.run --mount-entrypoints

  Start a new container with randomized gateway and service ports, and
  randomized container name::

      python -m localstack.dev.run --randomize

  You can also run custom commands:

      python -m localstack.dev.run bash -c 'echo "hello"'

  Or use custom entrypoints:

[...]

PR #9050, opened 2023-09-01
add container configurators for bootstrapping localstack

thrau:
## Motivation

With the new `Container` concept and the `ContainerFactory` fixture introduced in #8781, we now want an easy and re-usable way to configure containers. For instance, to test extensions, I always want to create a fresh localstack pro container with certain default configurations. The code doing the configuration should ideally be re-usable, have high-level concepts of the localstack container (such as "mount the localstack volume directory", "set gateway listen correctly"), while allowing fine-grained configuration of containers.

In #8994 I introduced a concept that is inspired by the [Functional Options Pattern](https://golang.cafe/blog/golang-functional-options-pattern.html) from go. I evolved this into a basic set of configurators that are typical for localstack.

This gives a very neat way of configuring the container together with the container factory, where you can see exactly what configurations are being added to the container:

```python
  container: Container = container_factory(
      configurators=[
          ContainerConfigurators.random_container_name,
          ContainerConfigurators.random_gateway_port,
          ContainerConfigurators.debug,
          ContainerConfigurators.mount_docker_socket,
          ContainerConfigurators.mount_localstack_volume("/tmp/localstack"),
      ]
  )
```

Furthermore, I wanted a way to stream container logs when running container tests, so I added a fixture that can be used as follows:

```python
def test(container_factory, stream_container_logs)
    container: Container = container_factory(...)
    with container.start() as running_container:
        stream_container_logs(container)
```
This creates a new background thread that will stream the logs of the container into a logger called `container.<container-name>`.

## Changes

* Add `ContainerConfigurator` concept
* Make `Container` aware of these configurators
* Add a default set of container configurators
* Refactor the `configure_container` method used by the CLI to use these configurators
* Add way to determine the gateway URL from a `Container`
* Add some tests and a fixture to stream logs
* Sneaked in a few changes to use `VolumeMappings.add` instead of `append` so we can maybe get rid of one of the functions

## TODO

- [x] Requires #9051


thrau:
testfailures are unrelated to the changes

localstack/config.py
```diff
@@ -625,6 +625,7 @@ def legacy_fallback(envar_name: str, default: T) -> T:
 
 
 # How to access LocalStack
+GATEWAY_LISTEN: List[HostAndPort]
```
simonrw, line 628: This should be covered by the types in the signature to `populate_legacy_edge_configuration`, no?

localstack/plugins.py
```diff
@@ -2,20 +2,10 @@
...
 
-@hooks.configure_localstack_container()
-def configure_edge_port(container: Container):
-    ports = [config.EDGE_PORT, config.EDGE_PORT_HTTP]
-    LOG.debug("configuring container with edge ports: %s", ports)
-    for port in ports:
-        if port:
-            container.config.ports.add(port)
-
-
```
simonrw, line 18: This is not needed because we use `GATEWAY_LISTEN` in `configure_container`, right? If so, thanks for picking up on this

localstack/testing/pytest/container.py
```diff
@@ -131,3 +206,32 @@ def docker_network(ensure_network):
     network_name = f"net-{short_uid()}"
     ensure_network(network_name)
     return network_name
+
+
+@pytest.fixture
+def stream_container_logs() -> Generator[LogStreamFactory, None, None]:
```
simonrw, line 212: Thank you for adding this. It is much better to see the container logs :+1:

simonrw (approved):
Looks like a great set of changes. Just some minor nits.

localstack/utils/bootstrap.py
```diff
@@ -376,6 +379,246 @@ def extract_port_flags(user_flags, port_mappings: PortMappings):
...
+        return _cfg
+
+    @staticmethod
+    def cli_params(params: Dict[str, Any]):
+        """
+        Parse docker CLI parameters and add them to the config. The currently known CLI params are:
+
+        :param params:
+        :return:
+        """
```
dfangl, line 553: It seems like something is missing here?

localstack/utils/bootstrap.py
```diff
@@ -376,6 +379,113 @@ def extract_port_flags(user_flags, port_mappings: PortMappings):
...
+
+    @staticmethod
+    def network(network: str):
+        def _cfg(cfg: ContainerConfiguration):
+            cfg.network = network
+
+        return _cfg
+
+
+def get_gateway_port(container: Container) -> int:
```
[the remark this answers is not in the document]
dfangl: While I agree with the sentiment - in this case this would only happen if we manually manufacture the container like this, right? All the configurators should prevent this from happening. The impact will be minimal, until we use a similar utility for a container not under our complete control.

I think it is perfectly fine to just raise a ValueError in this case, since the users would be us, not the enduser.

localstack/utils/bootstrap.py
```diff
@@ -376,6 +379,113 @@ def extract_port_flags(user_flags, port_mappings: PortMappings):
...
+        for value in gateway_listen.split(","):
+            host_and_port = config.HostAndPort.parse(
+                value,
+                default_host=constants.LOCALHOST_HOSTNAME,
+                default_port=constants.DEFAULT_PORT_EDGE,
+            )
+            edge_port = host_and_port.port
+            break
+
+    ports = container.config.ports.to_dict()
```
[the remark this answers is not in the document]
dfangl: It is set in `configure_container()`, but since it there is no default value for all containers, I agree.

localstack/utils/bootstrap.py
```diff
@@ -627,52 +907,20 @@ def configure_container(container: Container):
...
-    # pass through environment variables defined in config
-    for env_var in config.CONFIG_ENV_VARS:
-        value = os.environ.get(env_var, None)
-        if value is not None:
-            container.config.env_vars[env_var] = value
-    container.config.env_vars["DOCKER_HOST"] = f"unix://{config.DOCKER_SOCK}"
-
-    # TODO this is default now, remove once a considerate time is passed
-    # to activate proper signal handling
-    container.config.env_vars["SET_TERM_HANDLER"] = "1"
```
dfangl, line 648: Nice to get rid of that finally :)

dfangl (approved):
LGTM, love the new configuration options and used patterns!

localstack/utils/bootstrap.py
```diff
@@ -376,6 +379,113 @@ def extract_port_flags(user_flags, port_mappings: PortMappings):
...
+        def _cfg(cfg: ContainerConfiguration):
+            if cfg.volumes.find_target_mapping(constants.DEFAULT_VOLUME_DIR):
+                return
+            cfg.volumes.add(VolumeBind(str(host_path), constants.DEFAULT_VOLUME_DIR))
+
+        return _cfg
+
+    @staticmethod
+    def random_gateway_port(cfg: ContainerConfiguration):
+        cfg.ports.add(get_free_tcp_port(), constants.DEFAULT_PORT_EDGE)
```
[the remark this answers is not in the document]
thrau: true, this should be `gateway_listen(get_free_tcp_port())`

localstack/testing/pytest/container.py
```diff
@@ -29,6 +30,7 @@ def __call__(
         # convenience properties
         pro: bool = False,
         publish: list[int] | None = None,
```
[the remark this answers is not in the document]
thrau: good point, i think we should perhaps evolve the Container fixture in general to remove some of the localstack-specific assumptions, since they should be encapsulated in the configurators. in fact i was thinking that we could maybe, when we have time, establish a new container package, something like `localstack.containers`, much like we have our `localstack.http` framework, and start moving `Container`, `RunningContainer` etc there, to make things more re-usable across localstack.

localstack/utils/bootstrap.py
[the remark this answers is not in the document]
thrau: I added two default fields:
```
    volumes: VolumeMappings = dataclasses.field(default=VolumeMappings)
    ports: PortMappings = dataclasses.field(default=PortMappings)
```

localstack/plugins.py
thrau: yep, correct!

localstack/utils/bootstrap.py
thrau: My attention span when writing doc strings is very short

localstack/utils/bootstrap.py
thrau: i think enough time has passed :D

localstack/utils/bootstrap.py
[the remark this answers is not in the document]
thrau: thanks both for your comments. i quickly implemented simon's suggestion and added a test to document the behavior

PR #9051, opened 2023-09-02
add utility to find contiguous port range

thrau:
## Motivation

When creating multiple instances of localstack within tests, we need to randomize container ports. This includes the gateway port `4566`, but also the service port range. Specifically I need this in #9050 to make the `random_service_port_range` configurator work.

Until now, there was no good way of finding a random port range that is contiguous that could be mapped to the external service port range, this PR adds such a method to the `net` utils.

## Changes

* add `get_free_tcp_port_range` that finds a random contiguous port range of `num_ports`
* made `PortRange` inclusive, i.e., it now correctly considers the port `5002` in range if the range is `PortRange(5000,5002)`
* add global `dynamic_port_range` that marks the default dynamic port range 49152 to 65535. this is now used in `get_free_tcp_port` to also check+mark ports as reserved, making it a bit more robust
* did some refactoring along the way

### DIFF
--- a/localstack/utils/net.py
+++ b/localstack/utils/net.py
@@ -9,7 +9,8 @@ from urllib.parse import urlparse
 
 import dns.resolver
 
-from .. import config, constants
+from localstack import config, constants
+
 from .collections import CustomExpiryTTLCache
 from .numbers import is_number
 from .objects import singleton_factory
@@ -245,10 +246,7 @@ def get_free_tcp_port_range(num_ports: int, max_attempts: int = 50) -> "PortRang
                 return False
         return True
 
-    attempts = 0
-    while attempts <= max_attempts:
-        attempts += 1
-
+    for _ in range(max_attempts):
         # try to find a suitable starting point (leave enough space at the end)
         port_range_start = random.randint(
             dynamic_port_range.start, dynamic_port_range.end - num_ports - 1
@@ -304,6 +302,15 @@ class PortRange:
     """Manages a range of ports that can be reserved and requested."""
 
     def __init__(self, start: int, end: int):
+        """
+        Create a new port range. The port range is inclusive, meaning ``PortRange(5000,5005)`` is 6 ports
+        including both 5000 and 5005. This is different from ``range`` which is not inclusive, i.e.::
+
+            PortRange(5000, 5005).as_range() == range(5000, 5005 + 1)
+
+        :param start: the start port (inclusive)
+        :param end: the end of the range (inclusive).
+        """
         self.start = start
         self.end = end
 
@@ -315,6 +322,11 @@ class PortRange:
         self._ports_lock = threading.RLock()
 
     def as_range(self) -> range:
+        """
+        Returns a ``range(start, end+1)`` object representing this port range.
+
+        :return: a range
+        """
         return range(self.start, self.end + 1)
 
     def reserve_port(self, port: Optional[IntOrPort] = None, duration: Optional[int] = None) -> int:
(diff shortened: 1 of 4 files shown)


thrau:
Looks like podman tests have been failing for a while

localstack/utils/net.py
```diff
@@ -193,17 +201,74 @@ def get_free_udp_port(blocklist: List[int] = None) -> int:
...
+
+    def _is_port_range_free(_range: PortRange):
+        for _port in _range:
+            if dynamic_port_range.is_port_reserved(_port) or not port_can_be_bound(_port):
+                return False
+        return True
+
+    attempts = 0
+    while attempts <= max_attempts:
+        attempts += 1
```
simonrw, line 250: nit:
```suggestion
    for _ in range(max_attempts):
```

localstack/utils/net.py
```diff
@@ -193,17 +201,74 @@ def get_free_udp_port(blocklist: List[int] = None) -> int:
...
+
+        # try to find a suitable starting point (leave enough space at the end)
+        port_range_start = random.randint(
+            dynamic_port_range.start, dynamic_port_range.end - num_ports - 1
+        )
+        port_range = PortRange(port_range_start, port_range_start + num_ports - 1)
```
simonrw, line 256: I'm not sure starting at a random port and checking the range is a great approach. It does not take that long on my machine to scan the entire dynamic port range specified (~330ms) so could we check sequentially, perhaps with some intelligence that if a blocked port is found that we restart the scan from the port above that one?

I'm probably overthinking it though, the chances of there not being a valid range on the first few tries is pretty low.

simonrw (approved):
Nice, I like these QOL improvements!

localstack/utils/net.py
dfangl: That would change the semantics though, as we would now retry it `max_attempts + 1` times.
Since max attempts is defined as `    :param max_attempts: the number of times to retry if a contiguous range was not found`, I would argue it should the +1 behavior is correct, even though we might want to rename the "attempts" part of the parameter.

dfangl (approved):
Changes look very good from my side, had only minor comments.

localstack/utils/net.py
thrau: with `for _ in range(max_attempts)` there will be exactly 50 calls to `port_can_be_bound` (supposing each one fails). i think that's fine :-) added the check to the test

localstack/utils/net.py
thrau: this was the way i had it implemented initially, however went back to the completely random approach.
the problem i ran into was:
* either you start at `32768` every time and scan sequentially, then a subsequent call to `get_free_tcp_port_range(50)` would immediately fail (because max_attempts=50 and the first 50 calls would fail because you previously allocated from that range)
* or you start at a random port in which case you need to check  the boundary condition if you're at the end of the dynamic port range, in which case you need to select a new starting point and you might pick another random port again anyway.

you could implement some sort of exponential forward expansion when a port failed, but now it's getting really involved, and the random port approach should be pretty stable.

localstack/utils/net.py
```diff
@@ -21,6 +22,13 @@
     r"^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
 )
 
+# many linux kernels use 32768-60999, RFC 6335 is 49152-65535, so we use a mix here
```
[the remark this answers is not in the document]
thrau: That's a nice idea! I worry though that this will not work on windows so we need a fallback for that, and we're adding more complexity that way.

Issue #9054, opened 2023-09-04
sns.Subscribe fails with HTTP 500 in `latest` image


sseelmann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

HTTP 500 internal error when calling `sns.Subscribe` without attributes:

```
2023-09-04T05:55:45.886 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain
Traceback (most recent call last):
  File "/opt/code/localstack/localstack/aws/chain.py", line 90, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 123, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 93, in __call__
    skeleton_response = self.skeleton.invoke(context)
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(context, instance)
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 166, in dispatch_request
    result = handler(context, instance) or {}
  File "/opt/code/localstack/localstack/aws/forwarder.py", line 60, in _call
    return handler(context, req)
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
  File "/opt/code/localstack/localstack/services/sns/provider.py", line 582, in subscribe
    if sub.get(attr) != attributes.get(attr):
AttributeError: 'NoneType' object has no attribute 'get'
2023-09-04T05:55:45.888  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS sns.Subscribe => 500 (InternalError)
```

### Expected Behavior

sns.Subscribe works

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
$ cat docker-compose.yml 
version: "3.8"

services:
  localstack:
    container_name: localstack
    image: localstack/localstack
    ports:
      - "4566:4566"
    environment:
      - SERVICES=sqs,sns
      - EAGER_SERVICE_LOADING=1
      - DEBUG=1
      - DEFAULT_REGION=eu-west-1
      - DOCKER_HOST=unix:///var/run/docker.sock
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
[...]

PR #9055, opened 2023-09-04
Fix subscribe idempotency when no attributes are given

dominikschubert:
## Motivation
Addresses https://github.com/localstack/localstack/issues/9054

Specifying no attributes lead to the following exception 
```
E               if sub.get(attr) != attributes.get(attr):
E           AttributeError: 'NoneType' object has no attribute 'get'
```

## Changes

- Safety-check on the attributes before accessing keys.
- Adds a test case verifying this behavior

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -768,7 +768,7 @@ class TestSNSSubscriptionCrud:
         )
         snapshot.match("subscribe-idempotent", subscribe_resp)
 
-        # no attributes are working as well
+        # no attributes and empty attributes are working as well
         subscribe_resp = aws_client.sns.subscribe(
             TopicArn=topic_arn,
             Protocol="sqs",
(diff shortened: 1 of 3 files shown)


bentsku (approved):
🤦‍♂️ LGTM! Thanks a lot for that, good catch!

tests/aws/services/sns/test_sns.py
```diff
@@ -765,10 +768,23 @@ def test_subscribe_idempotency(
         )
         snapshot.match("subscribe-idempotent", subscribe_resp)
 
-        get_attrs_resp = aws_client.sns.get_subscription_attributes(
-            SubscriptionArn=subscribe_resp["SubscriptionArn"]
```
bentsku, line 769: q: why is this deleted now? I guess this check was there to verify that the subscription wasn't updated. We could do it also after subscribing with empty attributes, but I suppose the subscription stays the same

PR #9057, opened 2023-09-04
fix escaping backslash in comments

alexrashed:
## Motivation
The latest version of `botocore` contains a change in the spec (namely https://github.com/boto/botocore/commit/29dbcb9e1408c63ce9a3215447116789d6b4f528) which contains a backslash in the documentation of an operation of SESv2.
API code generated based on this botocore version fail the linting check, because the generated docstring contains the <code>\`\\`</code> (which violates [flake8's W605](https://www.flake8rules.com/rules/W605.html)).

## Changes
This PR just adds the backslash to the list of characters which need to be escaped in the docstring to fix the linting errors.

## Testing
The SESv2 API has been generated based on the change and it passes the linting.


thrau (approved):
great find, LGTM!

PR #9060, opened 2023-09-04
Add support for selection pattern for error responses

calvernaz:
## Motivation

When you want to override your backend response status codes, use API Gateway mapping templates or regular expressions to map the status codes. You can do this in proxy and non-proxy integrations with REST API.

This PR is mostly about "selection pattern" for error messages. Use a regular expression (regex) to customize an integration response based on the backend.

For a Lambda function, API Gateway matches the regex to the "errorMessage" header to return a status code. For example, to set a 400 response for any error that starts with "Malformed", set the method response status code to 400 and the Lambda error regex to Malformed.*.

For any unmatched errorMessage headers, API Gateway uses the default response.

## Changes

- Adds support for selection pattern for Lambdas based on the "errorMessage"
- Introduces ResponseParametersResolver and removes (most of) legacy "apply_response_parameters". This make the integrations self-contained and the response is not manipulated outside the integration code.
- Removes logic that checks if it's an AWS_PROXY integration for response templates.
- Migrate tests to "aws.verified" and snapshots


bentsku (approved):
LGTM! This is an awesome set of change, thank you so much! It's great to see more AWS validated tests, seeing `apply_response_parameters` gone and the new `ResponseParametersResolver`, awesome!

localstack/aws/api/lambda_/__init__.py
```diff
@@ -230,12 +231,11 @@ class Runtime(str):
     ruby2_5 = "ruby2.5"
     ruby2_7 = "ruby2.7"
     provided = "provided"
-    provided_al2 = "provided.al2"
```
joe4dev, line 233: same q as bentsku: what is this change? It seems unrelated. These API specs are auto-generated and should not be edited manually. It seems you want to order them alphabetically but I think their order should be rather consistent with the botocore specs: `.venv/lib/python3.10/site-packages/botocore/data/lambda/2015-03-31/service-2.json`

tests/aws/services/apigateway/test_apigateway_basic.snapshot.json
```diff
@@ -0,0 +1,17 @@
...
+      "start_execution_response": {
+        "executionArn": "<executionArn:1>",
+        "startDate": "<startDate>"
+      }
+    }
+  },
+  "tests/aws/services/apigateway/test_apigateway_basic.py::TestAPIGateway::test_apigateway_with_step_function_integration[DeleteStateMachine]": {
+    "recorded-date": "07-09-2023, 22:40:38",
+    "recorded-content": {
+      "delete_state_machine_response": {}
```
joe4dev, line 14: Is this snapshot intentionally empty?

joe4dev (commented):
Always love to see more aws-validated tests for the apigateway 🚀

doc/external_services_integration/kafka_self_managed_cluster/HOWTO.md
```diff
@@ -40,7 +40,7 @@ $ awslocal lambda create-function \
         "Mode": "PassThrough"
     },
     "RevisionId": "d85469d2-8558-4d75-bc0e-5926f373e12c",
-    "State": "Active",
+    "-State": "Active",
```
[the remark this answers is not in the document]
calvernaz: fat fingers it seems :)

localstack/aws/api/lambda_/__init__.py
calvernaz: good point, yea was trying to get it group together.

tests/aws/services/apigateway/test_apigateway_basic.snapshot.json
calvernaz: If the action is successful, the service sends back an HTTP 200 response with an empty HTTP body.

PR #9063, opened 2023-09-05
Implement APIGW UpdateGatewayResponse

bentsku:
## Motivation
We've had a feature request to implement UpdateGatewayResponse, which was missing from the moto implementation.
After a quick check to the moto implementation and some AWS validated tests, it showed that the current implementation was lacking all the default `GatewayResponse` from AWS. It was then quicker to implement it in our provider than to patch the whole moto implementation just to add an operation, especially since the moto implementation did not implement any validation. 

To note: even though this is implemented, we don't currently use `GatewayResponse` in our actual invocation part of the code. It also shows that there are different kinds of exception raised internally, and this could help us restructure the code around those exceptions (looking at the response type documentation page).

Documentation about `GatewayResponse`:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gatewayResponse-definition.html
https://docs.aws.amazon.com/apigateway/latest/developerguide/supported-gateway-response-types.html
https://docs.aws.amazon.com/apigateway/latest/api/API_UpdateGatewayResponse.html

## Changes
Migrated the following operations from moto:
- `PutGatewayResponse`
- `GetGatewayResponse`
- `GetGatewayResponses`
- `DeleteGatewayResponse`

Implemented the following:
- `UpdateGatewayResponse`

Created 3 new AWS validated tests validating the behaviour of our operations.

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -43,6 +43,9 @@ from localstack.aws.api.apigateway import (
     DomainNameStatus,
     EndpointConfiguration,
     ExportResponse,
+    GatewayResponse,
+    GatewayResponses,
+    GatewayResponseType,
     GetDocumentationPartsRequest,
     Integration,
     IntegrationResponse,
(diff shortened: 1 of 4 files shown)


calvernaz (approved):
this is not a feature I'm too familiar with but it seems from reading the docs on it, that it gives some flexibility on AWS API built-in errors, errors that do not reach the integration, like "missing authentication token" being a 403 can become a 404 and provide another body message.

The places where we have parity match the defaults, so potentially any built-in error can be modified and templated. 

Thanks for doing this, nice addition!

localstack/services/apigateway/provider.py
```diff
@@ -2034,6 +2037,170 @@ def get_usage_plans(
...
+        if response_type not in rest_api_container.gateway_responses:
+            # deep copy to avoid in place mutation of the default response when update using JSON patch
+            rest_api_container.gateway_responses[response_type] = copy.deepcopy(
+                DEFAULT_GATEWAY_RESPONSES[response_type]
+            )
+            rest_api_container.gateway_responses[response_type]["defaultResponse"] = False
+
+        patched_entity = rest_api_container.gateway_responses[response_type]
+
+        for index, operation in enumerate(patch_operations):
```
whummer, line 2168: Great parity in here! 👌

whummer (approved):
Awesome set of changes @bentsku , and amazing test coverage as usual! 🚀 

Great point on using `GatewayResponse` in our actual invocation logic, in some future iteration. This can surely help us restructure the logic around exception handling. 👌

Issue #8929, opened 2023-08-17
bug: PUT request on pre-signed URL with `Content-MD5` does not fail when the MD5 hash does not match the content


stekycz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Sending a `PUT` request to a pre-signed URL succeeds when the `Content-MD5` parameter/header value does not match the MD5 hash of the request content received.

It works as expected when using `PROVIDER_OVERRIDE_S3: legacy` though.

### Expected Behavior

The request should fail by MD5 hash of the content not matching the expected Content-MD5 hash.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

```js
const s3 = new AWS.S3({
  endpoint: 'http://localhost:4566',
  signatureVersion: 'v4', // Required for the presigned URL functionality with extra headers
  s3ForcePathStyle: true, // Do not use custom domains
});

const body = '123456';
const invalidBodyMd5AsBase64 = '123456'

const url = await s3.getSignedUrlPromise('putObject', {
  Bucket: 'test-bucket',
  Key: 'test-file',
  Expires: 3600,
  ContentMD5: invalidBodyMd5AsBase64 // Adds 'Content-MD5' parameter in query string
});

const result = await fetch(url, {
  method: 'PUT',
  body: body,
  headers: {
    'Content-Type': 'text/plain',
    'Accept': 'application/json',
    'Content-MD5': invalidBodyMd5AsBase64, // This is required, otherwise AWS refuses to accept the request
  },
});

assert(result.ok); // This should fail as the request should be denied because of miss-matching Content-MD5 hash
```

### Environment

```markdown
- OS: MacOS 13.5
- LocalStack: 2.1.0
```

idan-dagan-imagenai:
Any updates on this?

PR #9064, opened 2023-09-05
implement Content-MD5 check for PutObject

bentsku:
## Motivation
As requested and reported in #8929, we were missing `Content-MD5` validation for `PutObject`, which was implemented in the `legacy` provider.

`ChecksumAlgorithm` or `Content-MD5` allows to verify that the payload sent is properly received in the right shape.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#checking-object-integrity-md5

This will only concern `PutObject` for now, but many operation have this parameter (I think around 21, some like `PutBucketAcl` for example). 

As we had a look with @alexrashed a long, long time ago, the specs implement some kind of way to know which operations have to have some kind of required checksum.

The end goal would be to have a handler to manage the checksum verification for all operations, except the ones with a streaming body like `PutObject`, as those require special handling (reading the body only once). This is in the backlog and will be tackled in the future.

## Changes
Implemented different strategies depending on the provider.
For the default and streaming providers, I've done the same kind of logic as the checksum for the streaming provider, but using the `ETag` of the object, which represents the `hexdigest` of the md5 hash. The `Content-MD5` header is the base64 encoded digest of the hash, so it needs a bit of encoding/decoding to be able to compare them.
I did not do the same logic as the regular checksum in the default provider, because if the request is encoded using `aws-chunk`, the request data is not equal to the object body, and should be decoded twice (once in LocalStack and once in moto), so I'm doing after moto already calculated the ETag.

Same logic for the v3 provider, we first set the object to read the body only once, and we then verify the b64 encoded etag vs the provided `Content-MD5` value. 

Removed the xfail for the related test and improved it a bit.

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -232,8 +232,6 @@ class BucketLocationConstraint(str):
     us_gov_west_1 = "us-gov-west-1"
     us_west_1 = "us-west-1"
     us_west_2 = "us-west-2"
-    ap_south_2 = "ap-south-2"
-    eu_south_2 = "eu-south-2"
 
 
 class BucketLogsPermission(str):
@@ -370,8 +368,6 @@ class InventoryOptionalField(str):
     IntelligentTieringAccessTier = "IntelligentTieringAccessTier"
     BucketKeyStatus = "BucketKeyStatus"
     ChecksumAlgorithm = "ChecksumAlgorithm"
-    ObjectAccessControlList = "ObjectAccessControlList"
-    ObjectOwner = "ObjectOwner"
 
 
 class JSONType(str):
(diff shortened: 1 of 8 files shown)


thrau (approved):
LGTM! nicely done :+1:

localstack/aws/api/s3/__init__.py
```diff
@@ -232,8 +232,6 @@ class BucketLocationConstraint(str):
     us_gov_west_1 = "us-gov-west-1"
     us_west_1 = "us-west-1"
     us_west_2 = "us-west-2"
-    ap_south_2 = "ap-south-2"
```
bentsku, line 235: damn, seems like my botocore was not updated. 😭 will open a PR reintroducing them? or wait for the next automatic ASF update.

alexrashed: You can just manually trigger the action to trigger the update PR, which I did. The result is #9146. :)

PR #9065, opened 2023-09-05
upgrade to actions/checkout@v4, add actions to dependabot

alexrashed:
## Motivation
We are still seeing build issues related to https://github.com/actions/checkout/issues/1448, even though the incident should already be fixed (according to https://www.githubstatus.com/).
Some comments in this issue imply that an upgrade to the just newly released version 4 of [actions/checkout](https://github.com/actions/checkout) might fix the issue.
I used this opportunity toi also 

## Changes
- Upgrades all usages of [actions/checkout](https://github.com/actions/checkout) to version 4.
- Add GitHub actions to the dependabot config to file these upgrade PRs automatically in the future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,12 @@ updates:
     labels:
       - "area: dependencies"
       - "semver: patch"
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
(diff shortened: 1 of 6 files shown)


dfangl (approved):
LGTM!

PR #8970, opened 2023-08-23
Lambda invocation loop rework

joe4dev:
## Motivation

* Fixes the infinite Lambda re-spawn loop: Whenever a Lambda environment fails to start (e.g., even due to common user errors), the Lambda container enters an infinite retry loop instead of reporting an appropriate error message and stopping retries (depending on the error cause). This behavior is user-unfriendly and affects many GitHub issues and support tickets.
* Architecture mismatch between LocalStack and real AWS Lambda: The architectural mismatch causes parity issues, impedes debugging, and complicates adding new features.

## Goals

* Core loop: The default invocation path should be synchronous instead of asynchronous. This enables improved error handling and better AWS parity.
* Environment management: Allow for more explicit control over execution environments. This unblocks custom workers.
* Asynchronous event invocation: Separate asynchronous event invocations and handle retries, DLQs, and event destinations properly. This fixes bugs and increases AWS parity.
* Encapsulate concurrency logic: The logic related to concurrency management should be better encapsulated rather than scattered throughout an implicit and complex control flow through an asynchronous invocation loop. This enables better AWS parity and makes improving and adding features easier.

## Changes
This PR basically re-writes the data plane (i.e., everything related to environment assignment and invocation scheduling) of our new Lambda provider. The re-architected implementation is inspired based on the real architecture of AWS Lambda explained at [AWS re:Invent 2022 - A closer look at AWS Lambda (SVS404-R)](https://www.notion.so/localstack/Lambda-sync-async-invocation-model-d11375234d974fcbb0d1b70d737ccd80?pvs=4).

### Invocation model

image

Images from AWS: https://youtu.be/0_jfH6qijVY

### Synchronous invocation

image

### Asynchronous Invocation

image

More details to follow ...

### Internal Documentation

* [Collection of Meeting Notes (Notion)](https://www.notion.so/localstack/2023-07-05-until-2023-08-Lambda-Async-invoke-and-concurrency-2cd54c2281074e74b3530992c46bdac9?pvs=4)
* [Initial goals and brainstorming notes (Notion)](https://www.notion.so/localstack/2023-06-13-Lambda-Spawn-Loop-Rework-8f8c050054174f9d940a0d2e024863ce?pvs=4)

## Testing

This PR adds more AWS-validated tests and fixes some existing tests (e.g., fixes timing assumptions, and makes async tests faster). One example: We previously triggered the DLQ destination upon every failed invocation rather than only after the last retry.

## TODO

[...]

PR #9070, opened 2023-09-05
Make AWS region name configurable for tests

joe4dev:
## Motivation

We should have a configuration to change the test region. For cross-account tests, we should be able to configure the region in order to test cross-account tests within the same region.

Discovered during testing of the Lambda invocation loop rework https://github.com/localstack/localstack/pull/8970
/cc @dfangl @dominikschubert 

## Changes

Makes the AWS region name configurable in tests.


joe4dev:
We could also consider dropping the `_NAME` suffix to match the `AWS_REGION` config 🤔

viren-nadkarni (approved):
🚀

dfangl (approved):

joe4dev:
> We could also consider dropping the `_NAME` suffix to match the `AWS_REGION` config 🤔

Align environment variable naming with AWS CLI using `TEST_AWS_REGION` without the `_NAME` suffix.

Removing the suffix `_NAME` aligns the environment variable naming with the AWS CLI using `AWS_REGION` (which overwrites `AWS_DEFAULT_REGION`).
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

joe4dev:
Flaky arm64 test unrelated and reported in flaky test list (already know flake)

PR #9072, opened 2023-09-05
Add test for s3 bucket concurrency

dfangl:
## Motivation
We had some problem with simultaneous bucket operations. This PR adds a test for it. It should be fixed with #9044

## Changes
* Add list wrapper to avoid issue with concurrent modification of a dict while iterating
* Add test for 10 simultaneous bucket creations with the requirement of us-east-1 bucket creation idempotency.

### DIFF
--- a/tests/aws/services/s3/test_s3_concurrency.py
+++ b/tests/aws/services/s3/test_s3_concurrency.py
@@ -1,0 +1,40 @@
+import logging
+import threading
+
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+
+LOG = logging.getLogger(__name__)
+
+
+class TestParallelBucketCreation:
+    @markers.aws.unknown
+    def test_parallel_bucket_creation(self, aws_client_factory):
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Nice changes, and cool test! `Barrier` 😎

bentsku (changes requested):
Oops, I've missed something. We're not cleaning up the buckets we've created 😬

tests/aws/services/s3/test_s3_concurrency.py
```diff
@@ -0,0 +1,40 @@
...
+import threading
+
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+
+LOG = logging.getLogger(__name__)
+
+
+class TestParallelBucketCreation:
+    @markers.aws.unknown
```
bentsku, line 11:
```suggestion
    @markers.aws.only_localstack
```

bentsku (approved):

PR #9073, opened 2023-09-05
Allow multiple bind_hosts for port mappings

simonrw:
## Motivation

Currently the docker utilities only allow a single port mapping interface, i.e. all published ports must be bound to the same IP address. This means that e.g. the flags `-p 127.0.0.1:53:53 -p 4566:4566` is not currently allowed, without resorting to the additional flags concept.

## Changes

The docker utilities accept multiple `PortMappings` objects, which group port mappings by bind host. For backwards compatibility, the utilities support a single `PortMappings` object as well.


simonrw:
Closing this PR as it is very out of date, and we have not needed the functionality implemented here since.

PR #9074, opened 2023-09-05
add option to disable retries & disable boto retries for CI tests

dominikschubert:
## Motivation

These uncontrolled boto-client level retries are an inherent source of unpredictable behavior, so let's see if disabling them causes any unexpected issues.

Retries also cause unnecessary time waiting when we actually want to test a failure case.

## Changes

- disables retries for all clients created via our client factories in our test suite (CI)
- adds config variable `DISABLE_BOTO_RETRIES`. Set `DISABLE_BOTO_RETRIES=1` to opt-tin of the new behavior.

## Discussion

Might also want to go through cases where we use SDKs in lambda functions etc. to disable any retries.

### DIFF
--- a/localstack/aws/connect.py
+++ b/localstack/aws/connect.py
@@ -324,6 +324,12 @@ class ClientFactory(ABC):
         :return: Boto3 client.
         """
         with self._create_client_lock:
+            default_config = (
+                Config(retries={"max_attempts": 0})
+                if localstack_config.DISABLE_BOTO_RETRIES
+                else Config()
+            )
+
             client = self._session.client(
                 service_name=service_name,
                 region_name=region_name,
@@ -333,7 +339,7 @@ class ClientFactory(ABC):
                 aws_access_key_id=aws_access_key_id,
                 aws_secret_access_key=aws_secret_access_key,
                 aws_session_token=aws_session_token,
-                config=config.merge(Config(retries={"max_attempts": 0})),
+                config=config.merge(default_config),
             )
 
         return self._get_client_post_hook(client)
--- a/localstack/aws/connect.py
+++ b/localstack/aws/connect.py
@@ -333,7 +333,7 @@ class ClientFactory(ABC):
                 aws_access_key_id=aws_access_key_id,
                 aws_secret_access_key=aws_secret_access_key,
                 aws_session_token=aws_session_token,
-                config=config,
+                config=config.merge(Config(retries={"max_attempts": 0})),
             )
 
         return self._get_client_post_hook(client)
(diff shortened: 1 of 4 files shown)


dfangl (approved):
LGTM! Test runs on amd64 seem around 5 min shorter now. Of course it could just be random variance, but already looking good

localstack/aws/connect.py
```diff
@@ -333,7 +333,7 @@ def _get_client(
                 aws_access_key_id=aws_access_key_id,
                 aws_secret_access_key=aws_secret_access_key,
                 aws_session_token=aws_session_token,
-                config=config,
+                config=config.merge(Config(retries={"max_attempts": 0})),
```
viren-nadkarni, line 336: Would this be reverted? Are we trying to address an issue with the test suite? I would rather set this in the fixtures.

dominikschubert: No this is also intended to be set for internal communication, though we should observe this a bit. Generally I think it makes more sense to fail fast if there's an issue instead of retrying countless times.

joe4dev: We should not hide potential issues through retries for internal calls as long as we assume that LocalStack runs on the same machine (i.e., without a potentially problematic network connection).

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -2626,6 +2626,16 @@ def test_lambda_provisioned_lifecycle(self, create_lambda_function, snapshot, aw
             )
         snapshot.match("put_provisioned_on_alias_versionconflict", e.value.response)
 
+        def _wait_provisioned():
+            status = aws_client.lambda_.get_provisioned_concurrency_config(
+                FunctionName=function_name, Qualifier=function_version
+            )["Status"]
+            if status == "FAILED":
+                raise ShortCircuitWaitException("terminal fail state")
```
joe4dev, line 2634: nice pattern we should remember and use more often in tests 👍

joe4dev (approved):
LGTM 👍 

Did the rebase fix the other flaky tests or did you patch `b1a17823afc5cd2a147b5c5d83ab6a02aa660e9b` ?

dominikschubert:
> Did the rebase fix the other flaky tests or did you patch b1a17823afc5cd2a147b5c5d83ab6a02aa660e9b ?

just a normal re-run :man_shrugging:  but they're gone now on master anyway.

localstack/aws/connect.py
dominikschubert: To keep this a bit less controversion for now, I've switched the default back to normal retries & explicitly set `DISABLE_BOTO_RETRIES` in our CI tests. This means you'll need to set this in your environment locally though if you want to execute your tests without retries.

tests/aws/services/lambda_/test_lambda_api.py
joe4dev: FYI: also a better solution than the regular retry from the Lambda invocation loop 👍 (favored this one in rebase)

PR #9075, opened 2023-09-05
[SFN] Access to Context Object fields

MEPalma:
## Motivation
Fixes the access to Context Object fields through json path values.

## Changes
- Fixes lexing of object context fields which prevented the accessing of the entire Context Object value
- Fixes the values of several context object fields
- Removes empty Map values from the context object when not in a map state
- Adds relevant test for context object values accesses


tests/aws/services/stepfunctions/templates/base/statemachines/query_context_object_values.json5
```diff
@@ -0,0 +1,38 @@
+{
+  "Comment": "QUERY_CONTEXT_OBJECT_VALUES",
+  "StartAt": "QueryValues1",
+  "States": {
+    "QueryValues1": {
+      "Type": "Pass",
+      "Next": "QueryValues2",
+      "ResultPath": "$.query_values_1",
+      "Parameters": {
+        "context_object.$": "$$",
```
dominikschubert, line 10: good idea to check for this :+1:

dominikschubert (approved):
:rocket: 

That'll hopefully fix a lot of common use cases people are already running with v2

PR #9077, opened 2023-09-05
fix secretsmanager PutSecretValue on empty secret

bentsku:
## Motivation
A regression was introduced in moto and part of localstack after #9044 when calling `PutSecretValue` on a secret with no value.

## Changes
This adds a test for the fix in https://github.com/getmoto/moto/pull/6775
Edit: well, apparently this has been fixed with https://github.com/getmoto/moto/pull/6771
I don't know if we already want to do a moto bump, or we keep this fix inside LocalStack for now, this might be faster still. 

I've also put the fix in the patch for the moment until the next moto bump where we can remove it and use the one from upstream. This fix is kinda urgent, as the customer is downgrading moto in their LocalStack container in the meantime and it introduces incompatibilities, especially with `logs`.

### DIFF
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -496,7 +496,7 @@ def backend_update_secret_version_stage(
 @patch(FakeSecret.reset_default_version)
 def fake_secret_reset_default_version(fn, self, secret_version, version_id):
     # fn(self, secret_version, version_id)
-
+    # FIXME: remove for the next moto bump (above 4.2.post1) and uncomment line above
     # remove all old AWSPREVIOUS stages
     for old_version in self.versions.values():
         if "AWSPREVIOUS" in old_version["version_stages"]:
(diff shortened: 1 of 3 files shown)


localstack/services/secretsmanager/provider.py
```diff
@@ -495,7 +495,22 @@ def backend_update_secret_version_stage(
 
 @patch(FakeSecret.reset_default_version)
 def fake_secret_reset_default_version(fn, self, secret_version, version_id):
-    fn(self, secret_version, version_id)
+    # fn(self, secret_version, version_id)
+    # FIXME: remove for the next moto bump (above 4.2.post1) and uncomment line above
+    # remove all old AWSPREVIOUS stages
```
dominikschubert, line 500: Completely unrelated and task for futures us:
Would be great if we can find a way to unify this kind of thing across the code base a bit. In a lot of cases this is just forgotten and then causes issues down the road when upstream changes.

dominikschubert: I could also imagine adding a unit test that will fail as soon as we update moto, so that we are automatically reminded to fix this :thinking: 

again: not really something that needs any action right now

dominikschubert (approved):
LGTM

PR #9078, opened 2023-09-06
Patch default waiterconfig for boto waiters

dominikschubert:
## Motivation

We often spend way too long in wait calls caused by botocore waiters.

Botocore waiters are generated dynamically based on the spec in the waiters JSON files, e.g.:
```json
    "StackCreateComplete": {
      "delay": 30,
      "operation": "DescribeStacks",
      "maxAttempts": 120,
      "description": "Wait until stack status is CREATE_COMPLETE.",
      "acceptors": [
        {
          "argument": "Stacks[].StackStatus",
          "expected": "CREATE_COMPLETE",
          "matcher": "pathAll",
          "state": "success"
        },
        ...
      ]
    }
```

It should be immediately obvious why this is a problem. Without specifying a custom value for the waiters it will take the values in this spec (here 30s delays between retries). For our internal usage especially, this is way too long to wait between status checks and it's unreasonable to always specify a custom waiter config on each wait call.

## Changes

- Patches botocore.waiter.Waiter.wait to merge in lower default values that will have precedence over those loaded from the spec.
  - explicitly passed waiter configs will still have precedence over these defaults, e.g. `waiter.wait(StackName="something", WaiterConfig={...})`

## Example

`test_update_with_previous_parameter_value`

runtime with patch: ~4s
runtime without patch: ~32s

## Discussion

- We might want to fine-tune the default as 1s can become quite noisy
- Patching isn't as clean and we can try to somehow patch the json loader boto is using. I didn't see an immediate way to specify a custom loader for client creation, so I've gone with the patching of `Waiter.wait` for now.


localstack/aws/connect.py
```diff
@@ -27,10 +29,36 @@
 from localstack.utils.aws.aws_stack import get_local_service_url, get_s3_hostname
 from localstack.utils.aws.client_types import ServicePrincipal, TypedServiceClientFactory
 from localstack.utils.aws.request_context import get_region_from_request_context
+from localstack.utils.patch import patch
 from localstack.utils.strings import short_uid
 
 LOG = logging.getLogger(__name__)
 
+DEFAULT_DELAY = 5 if os.environ.get("TEST_TARGET") == "AWS_CLOUD" else 1
+DEFAULT_MAX_ATTEMPTS = 60 if os.environ.get("TEST_TARGET") == "AWS_CLOUD" else 120
```
dominikschubert, line 38: Might be better actually to use some monkeypatching here in the test setup instead. wdyt?

dominikschubert: Update: I've made it configurable now and only switch to aws defaults for tests in `tests/aws/`

tests/aws/conftest.py
```diff
@@ -36,7 +37,12 @@
 
 
 @pytest.hookimpl()
-def pytest_configure(config):
+def pytest_configure(config_):
```
dfangl, line 40: Seems like pytest does not like this renaming. We might want to rename the localstack config in the import instead.

dfangl (approved):
Looks good!

PR #9080, opened 2023-09-06
Disable terraform tests

dominikschubert:
## Motivation

The tests in `TestTerraform` have been a headache since forever.
They don't provide enough value given their flakiness, so we'll disable them and will look into some other strategy in the future. 

## Changes

- Disables terraform tests


alexrashed (approved):

bentsku (approved):

steffyP (approved):
thank you :)

PR #9081, opened 2023-09-06
cdk provisioning improvements for testing

steffyP:
## Motivation
Some improvements for the provisioning.

## Changes
* add optional parameter `delete_bucket` for the `cleanup_s3_bucket` helper
* don't assume stack-name exists for `get_stack_outputs`

### DIFF
--- a/localstack/testing/scenario/provisioning.py
+++ b/localstack/testing/scenario/provisioning.py
@@ -112,7 +112,7 @@ class InfraProvisioner:
                     )
 
     def get_stack_outputs(self, stack_name: str):
-        return self.cloudformation_stacks.get(stack_name, {}).get("Outputs")
+        return self.cloudformation_stacks.get(stack_name, {}).get("Outputs", {})
 
     def teardown(self):
         for fn in self.custom_cleanup_steps:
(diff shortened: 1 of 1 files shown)


localstack/testing/scenario/provisioning.py
```diff
@@ -110,7 +112,7 @@ def provision(self):
                     )
 
     def get_stack_outputs(self, stack_name: str):
-        return self.cloudformation_stacks[stack_name]["Outputs"]
+        return self.cloudformation_stacks.get(stack_name, {}).get("Outputs")
```
dominikschubert, line 115: Where/why would you call get_stack_outputs without a valid stack name? :thinking:

steffyP: Mainly thinking about custom cleanup-steps, where we try to access outputs. something like 
```
infra.add_custom_teardown(
            lambda: delete_db_instances_from_cluster(
                aws_client.rds,
                infra.get_stack_outputs(stack_name="NeptuneStack").get("NeptuneClusterId"),
            )
 )
 ```
In a scenario when the setup would have failed before: We still run the cleanup, accessing one stack's output might fail; only afterwards we would actually delete the stacks. Which would not happen if we already throw an exception before, right? 🤔 

We could argue that catching such errors should be part of the custom-cleanup, but I think it doesn't really hurt to add this change here?

Comment (dominikschubert) [reply localstack/testing/scenario/provisioning.py:115]: Guess its fine for now, but still not sure if I like it since it can hide the fact that stack name is actually just wrong. :sweat_smile: 

```suggestion
        return self.cloudformation_stacks.get(stack_name, {}).get("Outputs", {})
```

otherwise your sample above wouldn't work

dominikschubert (approved):
LGTM

Issue #9076, opened 2023-09-05
bug: S3 range requests where end < last return a 416, instead of returning the entire content.


sjperkins:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a "range: bytes=start-end" header is supplied where end < start, localstack's S3 implementation returns a 416 response.

https://github.com/localstack/localstack/blob/279aea173dc56ab58fcae7f25d5a0e515c3aff6d/localstack/services/s3/utils.py#L208-L222

By contrast AWS returns the entire content.

### Expected Behavior

AWS Behaviour returns the entire content if end < start:

```bash
$ echo "Hello World" > test.txt
$  aws s3 cp test.txt s3://ratt-public-data   # Permissions needed to upload, but I've left test.txt here
$ curl  -H "range: bytes=1-0" https://ratt-public-data.s3.af-south-1.amazonaws.com/test.txt
Hello World
$ curl  -H "range: bytes=15-1" https://ratt-public-data.s3.af-south-1.amazonaws.com/test.txt
Hello World
```

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

Localstack behaviour for a "range: bytes=1-0" header returns a 416 response:

```bash
$ echo "Hello World" > test.txt
$ localstack start -d
$ awslocal s3api create-bucket --bucket sample-bucket --region=us-east-1
{
    "Location": "/sample-bucket"
}
$ awslocal s3api put-object --bucket sample-bucket --key test.txt --body test.txt --region=us-east-1
{
    "ETag": "\"e59ff97941044f85df5297e1c302d260\""
}
$ curl -H "range: bytes=1-0" http://localhost:4566/sample-bucket/test.txt
<?xml version='1.0' encoding='utf-8'?>
<Error><Code>InvalidRange</Code><Message>The requested range is not satisfiable</Message><RequestId>aa8a2346-5dc0-4728-971c-a74839b3f1b0</RequestId><ActualObjectSize>12</ActualObjectSize><RangeRequested>bytes=1-0</RangeRequested></Error>
```

### Environment

```markdown
- OS: Ubuntu 20.04
- LocalStack: 2.2.0
```

bentsku:
[...]

PR #9082, opened 2023-09-06
fix S3 v3 Range handling

bentsku:
## Motivation
As reported in #9076, there is an issue in the way Moto and LocalStack handles HTTP Range requests for S3. AWS is apparently not following the RFC for range requests and is very, very forgiving, as shown by the added test.
> An [int-range](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range) is invalid if the [last-pos](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range) value is present and less than the [first-pos](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range).

As pointed out by @sjperkins who went through the RFC, the server can ignore the range header for some conditions:
> A server that supports range requests MAY ignore or reject a [Range](https://www.rfc-editor.org/rfc/rfc9110.html#field.range) header field that contains an invalid [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) ([Section 14.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#range.specifiers)), a [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) with more than two overlapping ranges, or a set of many small ranges that are not listed in ascending order,

Basically, if the range request is badly formatted or invalid, S3 will just treat the request as a non-range request and will return the entire object with a `200` status code.
Only certain conditions will trigger an exception and `416`, mainly if the beginning of the request is above the end of the object and the request is a valid `Range` request (meaning it can be parsed), or if the request is a "suffix" request and it ends with 0.

## Changes
This fixes the behaviour only for the v3 provider for now, I'll need to open a PR upstream in moto to fix the behaviour there to fix the default current provider. Edit: the PR upstream is already merged, so I'd like to have the v3 behaviour in line.

I might remove all the `skip` for the new provider tests and selectively xfail the tests for the default provider in a follow up PR, to see the extent of the issue with the current provider, and try to have some tests such as this one valid for both.

I've launched a run with this fix in the forced v3 S3 branch here, and it's green: https://app.circleci.com/pipelines/github/localstack/localstack/17924/workflows/3026ab82-afaa-4b31-9b45-fbd7afbab1bd

This also fix the behaviour for `UploadPartCopy` which has a very strict handling of the `CopySourceRange` parameter.

### DIFF
--- a/localstack/services/s3/utils.py
+++ b/localstack/services/s3/utils.py
@@ -191,7 +191,7 @@ class ObjectRange(NamedTuple):
     end: int  # the end of the end
 
 
-def parse_range_header(range_header: str, object_size: int) -> ObjectRange:
+def parse_range_header(range_header: str, object_size: int) -> ObjectRange | None:
     """
     Takes a Range header, and returns a dataclass containing the necessary information to return only a slice of an
     S3 object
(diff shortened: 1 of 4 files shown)


sjperkins:
> AWS is apparently not following the RFC for range requests and is very, very forgiving, as shown by the added test.

Just wanted to strengthen the above argument by drawing attention to https://github.com/localstack/localstack/issues/9076#issuecomment-1709846708. The [rfc](https://www.rfc-editor.org/rfc/rfc9110.html#name-range-requests) gives servers the latitude to ignore an invalid range request:

> A server that supports range requests MAY ignore or reject a [Range](https://www.rfc-editor.org/rfc/rfc9110.html#field.range) header field that contains an invalid [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) ([Section 14.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#range.specifiers)), a [ranges-specifier](https://www.rfc-editor.org/rfc/rfc9110.html#rule.ranges-specifier) with more than two overlapping ranges, or a set of many small ranges that are not listed in ascending order,

dfangl (approved):
LGTM!

PR #9085, opened 2023-09-07
update CODEOWNERS

alexrashed:
## Motivation
The `CODEOWNERS` file was a bit outdated.
This PR tries to update it as good as possible.
Please let me know if some of the assignments are not correct, or if anything you can think of is missing.
However, this change does not claim to be comprehensive.

## Changes
- Update `CODEOWNERS` file.
- Renames the elasticsearch cloudformation model to the correct name (`es` is the AWS service name for elasticsearch we use everywhere else). /cc @dominikschubert 

/cc mentioned code owners: @ackdav @alexrashed @baermat @calvernaz @dfangl @dominikschubert @giograno @macnev2013 @MEPalma @sannya-singal @silv-io @simonrw @steffyP @thrau @viren-nadkarni


dominikschubert (approved):
Thanks for updating this! :pray:

PR #9088, opened 2023-09-07
Implement resource provider for AWS::SNS::Topic

Morijarti:
## Motivation
CloudFormation is important part of AWS and extending its functionality with SNS topics will further improve its implementation in LocalStack

Initial implementation of SNS topic resource for use by CloudFormation
Added support for
- Creating topic with subscriptions
- Deleting topic

## Testing
`python -m pytest tests/aws/services/cloudformation/resources/test_sns.py`

## TODO

What's left to do:
- Idempotency tests (creating a topic with existing name)
- IAM permissions testing
- Testing if deleting topic using CF deletes subscriptions as well or if they need to be deleted as beforehand
- Implementing resource reading
- Implementing resource update
-->


localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,192 @@
...
+        model = request.desired_state
+        sns = request.aws_client_factory.sns
+        sns.delete_topic(TopicArn=model["TopicArn"])
+        """
+        Delete a resource
+
+        IAM permissions required:
+          - sns:DeleteTopic
+        """
+        return ProgressEvent(status=OperationStatus.SUCCESS, resource_model={})
```
dominikschubert, line 159: Would be great if you could implement the "deletion" (+ potential unsubscribe) here.

If you're feeling particularly motivated, you can of course also cover this via a test, because I think this is probably not properly covered right now :grimacing:

dominikschubert (commented):
Thanks for your contribution. :partying_face: 

I've added a few comments on what would need to be addressed for this to be merged :smile:

Morijarti:

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,192 @@
...
+          - sns:Subscribe
+          - sns:GetTopicAttributes
+          - sns:PutDataProtectionPolicy
+
+        """
+        model = request.desired_state
+        sns = request.aws_client_factory.sns
+        # TODO: validations and iam checks
+
+        if not request.custom_context.get(REPEATED_INVOCATION):
```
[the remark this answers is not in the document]
Morijarti: Removed

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,192 @@
...
+                model["TopicName"] = f"topic-{short_uid()}"
+
+            create_sns_response = sns.create_topic(Name=model["TopicName"],
+                                                   Attributes=attributes
+                                                   )
+            request.custom_context[REPEATED_INVOCATION] = True
+            model["TopicArn"] = create_sns_response["TopicArn"]
+
+            # now we add subscriptions if they exists
+            # TODO do we need to keep arns for unsubscribing before deleting topic
```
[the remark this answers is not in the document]
Morijarti: I was referring to subscription ARNs, but I checked AWS documentations and they are automatically deleted with topic

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,192 @@
...
+        model = request.desired_state
+        sns = request.aws_client_factory.sns
+        # TODO: validations and iam checks
+
+        if not request.custom_context.get(REPEATED_INVOCATION):
+            # this is the first time this callback is invoked
+            # TODO: defaults
+            # TODO: idempotency
+            attributes = {k: v for k, v in model.items()
+                          if v is not None}
```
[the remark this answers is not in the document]
Morijarti: Removed TopicName from attributes

localstack/services/sns/resource_providers/aws_sns_topic.py
Morijarti: Updated the code.

Test 
`tests.aws.services.cloudformation.resources.test_sns.test_deploy_stack_with_sns_topic`
checks if topic is properly deleted after stack is destroyed.

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,178 @@
...
+        subscriptions = []
+        if attributes.get("Subscription") is not None:
+            subscriptions = attributes["Subscription"]
+            del attributes["Subscription"]
+
+        # in case cloudformation didn't provide topic name
+        if model.get("TopicName") is None:
+            model["TopicName"] = f"topic-{short_uid()}"
+
+        create_sns_response = sns.create_topic(Name=model["TopicName"], Attributes=attributes)
```
dominikschubert, line 100: The attributes still potentially contain the raw `Tags` from the properties, which isn't valid as an attribute value unfortunately (see CI failure [here](https://app.circleci.com/pipelines/github/localstack/localstack/18238/workflows/1c229624-f507-48b6-95dd-11bb935fbe67/jobs/140322) )
```
...
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/botocore/validate.py", line 381, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter Attributes.Tags, value: [{'Key': 'param-value', 'Value': 'param ls-param-value-fdccf91f'}], type: <class 'list'>, valid types: <class 'str'>
```

dominikschubert (changes requested):

localstack/services/sns/resource_providers/aws_sns_topic.py
Morijarti: I can modify it so tags are extracted from attributes into a list. 
Then I would delete them from attributes and apply them separately, similarly to how I did it with subscriptions.

It's a bit odd, that same as for subscription, there is a valid string value to which we should be able to convert it (at least according to API), but everything I tried didn't produce any results.

Morijarti: Done

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,185 @@
...
+        request.custom_context[REPEATED_INVOCATION] = True
+        model["TopicArn"] = create_sns_response["TopicArn"]
+
+        # now we add subscriptions if they exists
+        # TODO do we need to keep arns for unsubscribing before deleting topic
+        for subscription in subscriptions:
+            sns.subscribe(
+                TopicArn=model["TopicArn"],
+                Protocol=subscription["Protocol"],
```
dominikschubert, line 114: Seems this is optional (latest CI run fails here with a `KeyError`) and might either need to be set to a default or dropped from the subscribe call if it doesn't exist. 

```
    event = self.execute_action(resource_provider, payload)
  File "/opt/code/localstack/localstack/services/cloudformation/resource_provider.py", line 712, in execute_action
    return resource_provider.create(request)
  File "/opt/code/localstack/localstack/services/sns/resource_providers/aws_sns_topic.py", line 114, in create
    Protocol=subscription["Protocol"],
KeyError: 'Protocol'
```

Morijarti: I wasn't sure what a default could be so, I've added the code that drops subscription in case there Protocol field is missing.
Maybe I could add a warning as well?

localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+
+        subscriptions = []
+        if attributes.get("Subscription") is not None:
+            subscriptions = attributes["Subscription"]
+            del attributes["Subscription"]
+
+        tags = []
+        if attributes.get("Tags") is not None:
+            subscriptions = attributes["Tags"]
```
dominikschubert, line 98:
```suggestion
            tags = attributes["Tags"]
```

I think this is the actual cause of the KeyError from earlier. 
I was wondering why it even entered the for loop further below and it seems the tags are set as the subscriptions here (probably unintentionally) :grimacing: 

Removing this should also remove the need for the `if "Protocol" in subscription:`,

Morijarti: good catch!

PR #9091, opened 2023-09-07
enable new s3 v3 tests for current provider

bentsku:
## Motivation
When writing the new S3 v3 provider, a lot of new tests have been written to either verify missing cases or tests things that were implemented in moto only.
This PR aims to enable those tests for the current provider, and xfail the ones that are not passing.

## Changes
Remove all the `skip` pytest markers and go individually over each test to verify if they pass. 

## TODO

What's left to do:

- [x] merge #9082 because this PR is based on it, as it added a new test that was the point to have those tests run for both providers


steffyP (approved):
Great work, @bentsku! 👏 🚀

PR #9092, opened 2023-09-07
Fix deletion of AWS::IAM::Policy 

dominikschubert:
## Motivation

When working on other tasks, I noticed the AWS::IAM::Policy usually fails to clean up leading to a lot of cleanup issues since other resources usually depend on this resource. It seems this was also implemented wrong in the original `GenericBaseModel` version.

## Changes

- Add proper delete for `AWS::IAM::Policy`
- Add test for a broken delete where resources need to be orphaned to "unblock" the delete.
- Fix transformers, so that it won't try to do a reference replacement with `""` values

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -22,6 +22,7 @@ from localstack.services.cloudformation.engine.template_utils import (
 )
 from localstack.services.cloudformation.resource_provider import (
     Credentials,
+    OperationStatus,
     ResourceProviderExecutor,
     ResourceProviderPayload,
     get_resource_type,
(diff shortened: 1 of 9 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1358,13 +1359,31 @@ def apply_change(self, change: ChangeConfig, stack: Stack) -> None:
 
         progress_event = executor.deploy_loop(resource_provider_payload)  # noqa
 
+        # TODO: clean up the surrounding loop (do_apply_changes_in_loop) so that the responsibilities are clearer
+        stack_action = get_action_name_for_resource_change(action)
+        match progress_event.status:
+            case OperationStatus.FAILED:
+                stack.set_resource_status(resource_id, f"{stack_action}_FAILED")
+                # TODO: remove exception raising here?
```
dominikschubert, line 1367: currently we still need to raise otherwise the enclosing loop in the template deployer won't properly set the stack/resource states

pinzon (approved):
LGTM

PR #9093, opened 2023-09-07
extend lambda alias model

pinzon:
## Motivation
This PR fixes the CFn crash that happens when deploying a AWS::Lambda::Alias resource that has a `ProvisionedConcurrencyConfig` 

## Changes
Model limits the correct parameters for the creation and then it check if the attribute exists.

## Testing
The testing just extends a previous tests and corrects the code of the lambda function so it can be deployed on AWS


pinzon:
closed in favor of #9100

PR #9095, opened 2023-09-07
Update deployment stage_name when creating API Gateway stages to fix parity issues

whummer:
## Motivation

Addresses a user issue reported in Slack. We are currently seeing parity issues in the linkage between API Gateway stages and deployments. In a nutshell, we need to establish a bidirectional link `stage<>deployment` for the `create_stage(..)` API. Moto was currently only adding the `stage->deployment`, but the reverse direction was still missing, leading to deployment errors, e.g., with Terraform.

## Changes

* update `create_stage(..)` to update the `stage_name` on the Deployment on creation of a Stage
* add small fix in `get_stages(..)` to transform the result entries
* add Makefile parametrization for `DISABLE_BOTO_RETRIES` (API Gateway snapshot tests require retries, due to high frequency of `TooManyRequestsException`)
* add a snapshot test to cover a sequence of different creations/deletions of stages&deployments

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -869,7 +869,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         response = call_moto(context)
         for stage in response["item"]:
             self._patch_stage_response(stage)
-            stage.pop("description", None)
+            if not stage.get("description"):
+                stage.pop("description", None)
         return Stages(**response)
 
     @handler("UpdateStage")
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -76,6 +76,7 @@ from localstack.aws.api.apigateway import (
     RestApis,
     SecurityPolicy,
     Stage,
+    Stages,
     StatusCode,
     String,
     Tags,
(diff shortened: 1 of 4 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -858,6 +863,15 @@ def get_stage(self, context: RequestContext, rest_api_id: String, stage_name: St
         self._patch_stage_response(response)
         return response
 
+    def get_stages(
+        self, context: RequestContext, rest_api_id: String, deployment_id: String = None
+    ) -> Stages:
+        response = call_moto(context)
+        for stage in response["item"]:
+            self._patch_stage_response(stage)
+            stage.pop("description", None)
```
calvernaz, line 872: I wonder if the description should only be discarded if not present, it's in the model, just a case of remove if not present perhaps? https://docs.aws.amazon.com/apigateway/latest/api/API_GetStages.html#API_GetStages_ResponseElements

calvernaz (commented):
Thanks for doing this, I do notice situations where I had to apply twice to get the expected result, maybe this will fix it.

I just left a comment regarding the removal of the description field.

localstack/services/apigateway/provider.py
whummer: Great catch, thanks for checking it! Updated the PR .. 👍

bentsku (approved):
LGTM! Great catch! Love to see all those nice parity upgrade for API Gateway, thanks a lot!

PR #9099, opened 2023-09-08
fix S3 s3_empty_bucket fixture and add parity

bentsku:
## Motivation
It seems I introduced an issue when fixing the `s3_empty_bucket` fixture to force deleting locked objects.
```python
2023-09-08T17:33:09.910 DEBUG --- [  MainThread] localstack.testing.pytest.fixtures : error cleaning up bucket test-bucket-5c896954: An error occurred (InvalidArgument) when calling the DeleteObjects operation: x-amz-bypass-governance-retention is only applicable to Object Lock enabled buckets.
```
It also showed that LocalStack didn't have parity when checking this, and allowed the parameter to be passed if the bucket doesn't have it enabled. Added the fix in LocalStack as well, but this was an issue only when running against AWS (lots of clean up to do 😭)

## Changes
- added a check in the fixture to check if the bucket can be locked.
- added a test for the fixture, so we're sure it's working.
- added parity in LocalStack to also raise the error if the parameter is used.

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -646,6 +646,16 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         request: DeleteObjectRequest,
     ) -> DeleteObjectOutput:
         # TODO: implement DeleteMarker response
+        if request.get("BypassGovernanceRetention") is not None:
+            bucket_name = request["Bucket"]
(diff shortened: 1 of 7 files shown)


viren-nadkarni (approved):
:100:

dominikschubert (approved):
great catch :rocket:

PR #9100, opened 2023-09-08
AWS::Lambda::Alias Resource Provider

pinzon:
## Motivation
The Generic Base model for the Lambda::Alias doesn't support the `ProvisionedConcurrencyConfig` attribute so I took the opportunity to code the resource provider for the model 
## Changes
- New Resource provider
- Test uses the attribute and is AWS validated

## Testing
The testing of the attribute requires the lambda to return a valid response due AWS running the function code before provisioning the concurrency config.


localstack/services/awslambda/resource_providers/aws_lambda_alias.py
```diff
@@ -0,0 +1,149 @@
...
+        request: ResourceRequest[LambdaAliasProperties],
+    ) -> ProgressEvent[LambdaAliasProperties]:
+        """
+        Delete a resource
+
+
+        """
+        model = request.desired_state
+        lambda_ = request.aws_client_factory.lambda_
+        lambda_.delete_alias(
```
dominikschubert, line 121: did you check if this also deletes the provisioned concurrency on AWS? :thinking:

localstack/services/awslambda/resource_providers/aws_lambda_alias.py
```diff
@@ -0,0 +1,149 @@
...
+
+        """
+        model = request.desired_state
+        lambda_ = request.aws_client_factory.lambda_
+
+        create_params = {
+            k: v
+            for k, v in model.items()
+            if k in ["FunctionName", "FunctionVersion", "Name", "Description", "RoutingConfig"]
+        }
```
dominikschubert, line 80: you can use the functions available in the imported `util` package which should have a helper for this (if you want)

localstack/services/awslambda/resource_providers/aws_lambda_alias.py
```diff
@@ -0,0 +1,149 @@
...
+                Qualifier=model["Id"].split(":")[-1],
+                ProvisionedConcurrentExecutions=model["ProvisionedConcurrencyConfig"][
+                    "ProvisionedConcurrentExecutions"
+                ],
+            )
+
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model=model,
+        )
```
dominikschubert, line 97: Seems this is an async operation that might have to wait for the provisioned concurrency to actually be available. It could also be the case that they just leave that async and immediately return CREATE_COMPLETE though :man_shrugging: . Did you test this? :thinking:

localstack/services/awslambda/resource_providers/aws_lambda_alias.py
```diff
@@ -0,0 +1,172 @@
...
+                    Qualifier=model["Id"].split(":")[-1],
+                    ProvisionedConcurrentExecutions=model["ProvisionedConcurrencyConfig"][
+                        "ProvisionedConcurrentExecutions"
+                    ],
+                )
+
+            return ProgressEvent(
+                status=OperationStatus.IN_PROGRESS,
+                resource_model=model,
+            )
```
dominikschubert, line 98: We'll need to check if the context on AWS is something you can mutate because otherwise this provider wouldn't work on AWS since you're not passing the context back with the progress event. :thinking: 

For now please also pass the context object on return.

dominikschubert (approved):
LGTM, minor nit :+1:

PR #9102, opened 2023-09-09
add support for xml to json responses

calvernaz:
## Motivation

Rework this PR to instead of forcing xml to json format exchange, it adds parity to AWS by defaulting SQS responses to JSON. It also works if the request parameter is set to `application/xml`, in that case, the response from the integration is not JSON but XML.

## Changes

- Applies the same defaults as AWS, the default response being JSON.
- It also adds a basic implementation of `update_integration` and `update_integration_response` just to comply with the tests. A follow-up PR will improve the implementation.
- Snapshot tests and exercise multiple response formats

## Testing

Adds a snapshot that tackles the customer issue

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -5,6 +5,7 @@ import json
 import logging
 import re
 import time
+import xml.etree.ElementTree as ET
 from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict, Union
@@ -551,6 +552,15 @@ def get_stage_variables(context: ApiInvocationContext) -> Optional[Dict[str, str
 # ---------------
 
 
+def is_valid_xml(xml_string):
+    try:
+        # Attempt to parse the XML string
+        ET.fromstring(xml_string.encode("utf-8"))
+        return True
+    except ET.ParseError:
+        return False
+
+
 def path_based_url(api_id: str, stage_name: str, path: str) -> str:
     """Return URL for inbound API gateway for given API ID, stage name, and path"""
     pattern = "%s/restapis/{api_id}/{stage_name}/%s{path}" % (
@@ -764,6 +774,10 @@ def apply_json_patch_safe(subject, patch_operations, in_place=True, return_list=
                     # if "path" is an attribute name pointing to an array in "subject", and we're running
                     # an "add" operation, then we should use the standard-compliant notation "/path/-"
                     operation["path"] = f"{path}/-"
+
+            if operation["op"] == "remove":
+                path = operation["path"]
+                common.assign_to_path(subject, path, value={}, delimiter="/")
 
             result = apply_patch(subject, [operation], in_place=in_place)
             if not in_place:
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -5,7 +5,6 @@ import json
 import logging
 import re
 import time
-import xml.etree.ElementTree as ET
 from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict, Union
@@ -550,15 +549,6 @@ def get_stage_variables(context: ApiInvocationContext) -> Optional[Dict[str, str
 # ---------------
 # UTIL FUNCTIONS
 # ---------------
-
-
-def is_valid_xml(xml_string):
-    try:
-        # Attempt to parse the XML string
-        ET.fromstring(xml_string.encode("utf-8"))
-        return True
-    except ET.ParseError:
-        return False
 
 
 def path_based_url(api_id: str, stage_name: str, path: str) -> str:
--- a/localstack/services/apigateway/patches.py
+++ b/localstack/services/apigateway/patches.py
@@ -70,8 +70,29 @@ def apply_patches():
                 integration.timeout_in_millis = int(integration.timeout_in_millis)
             if skip_verification := (integration.tls_config or {}).get("insecureSkipVerification"):
                 integration.tls_config["insecureSkipVerification"] = str_to_bool(skip_verification)
+            return 200, {}, json.dumps(integration.to_json())
 
         return result
+
+    @patch(APIGatewayResponse.integration_responses)
+    def apigateway_response_integration_responses(fn, self, request, *args, **kwargs):
+        result = fn(self, request, *args, **kwargs)
+        if self.method not in ["PATCH"]:
+            return result
+
+        url_path_parts = self.path.split("/")
+        function_id = url_path_parts[2]
+        resource_id = url_path_parts[4]
+        method_type = url_path_parts[6]
+
+        integration = self.backend.get_integration(function_id, resource_id, method_type)
+        if not integration:
+            return result
+
+        if self.method == "PATCH":
+            patch_operations = self._get_param("patchOperations")
+            apply_json_patch_safe(integration.to_json(), patch_operations, in_place=True)
+            return 200, {}, json.dumps(integration.to_json())
 
     @patch(APIGatewayResponse.usage_plan_individual)
     def apigateway_response_usage_plan_individual(
--- a/localstack/services/apigateway/patches.py
+++ b/localstack/services/apigateway/patches.py
@@ -74,26 +74,6 @@ def apply_patches():
 
         return result
 
-    @patch(APIGatewayResponse.integration_responses)
-    def apigateway_response_integration_responses(fn, self, request, *args, **kwargs):
-        result = fn(self, request, *args, **kwargs)
-        if self.method not in ["PATCH"]:
-            return result
-
-        url_path_parts = self.path.split("/")
-        function_id = url_path_parts[2]
-        resource_id = url_path_parts[4]
-        method_type = url_path_parts[6]
-
-        integration = self.backend.get_integration(function_id, resource_id, method_type)
-        if not integration:
-            return result
-
-        if self.method == "PATCH":
-            patch_operations = self._get_param("patchOperations")
-            apply_json_patch_safe(integration.to_json(), patch_operations, in_place=True)
-            return 200, {}, json.dumps(integration.to_json())
-
     @patch(APIGatewayResponse.usage_plan_individual)
     def apigateway_response_usage_plan_individual(
         fn, self, request, full_url, headers, *args, **kwargs
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -515,6 +515,46 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         parent_id = moto_resource.parent_id
         api_resources[parent_id].remove(resource_id)
 
+    def update_integration_response(
+        self,
+        context: RequestContext,
+        rest_api_id: String,
+        resource_id: String,
+        http_method: String,
+        status_code: StatusCode,
+        patch_operations: ListOfPatchOperation = None,
+    ) -> IntegrationResponse:
+        # XXX: THIS IS NOT A COMPLETE IMPLEMENTATION, just the minimum required to get tests going
+        # TODO: validate patch operations
+
+        moto_rest_api = get_moto_rest_api(context, rest_api_id)
+        moto_resource = moto_rest_api.resources.get(resource_id)
+        if not moto_resource:
+            raise NotFoundException("Invalid Resource identifier specified")
+
+        moto_method = moto_resource.resource_methods.get(http_method)
+        if not moto_method:
+            raise NotFoundException("Invalid Method identifier specified")
+
+        integration_response = moto_method.method_integration.integration_responses.get(status_code)
+        if not integration_response:
+            raise NotFoundException("Invalid Integration Response identifier specified")
+
+        for patch_operation in patch_operations:
+            op = patch_operation.get("op")
+            path = patch_operation.get("path")
+
+            # for path "/responseTemplates/application~1json"
+            if "/responseTemplates" in path:
+                value = patch_operation.get("value")
+                if not isinstance(value, str):
+                    raise BadRequestException(
+                        f"Invalid patch value  '{value}' specified for op '{op}'. Must be a string"
+                    )
+                param = path.removeprefix("/responseTemplates/")
+                param = param.replace("~1", "/")
+                integration_response.response_templates.pop(param)
+
     def update_resource(
         self,
         context: RequestContext,
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -187,6 +187,7 @@ class Templates:
             "path": {},
             "querystring": {},
         }
+
         return {
             "context": ctx,
             "stage_variables": api_context.stage_variables or {},
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -673,10 +673,23 @@ class SQSIntegration(BackendIntegration):
     def invoke(self, invocation_context: ApiInvocationContext):
         integration = invocation_context.integration
         uri = integration.get("uri") or integration.get("integrationUri") or ""
-
-        template = integration["requestTemplates"].get(APPLICATION_JSON)
         account_id, queue = uri.split("/")[-2:]
         region_name = uri.split(":")[3]
+
+        headers = get_internal_mocked_headers(
+            service_name="sqs",
+            region_name=region_name,
+            role_arn=invocation_context.integration.get("credentials"),
+            source_arn=get_source_arn(invocation_context),
+        )
+
+        # integration parameters can override headers
+        integration_parameters = self.request_params_resolver.resolve(context=invocation_context)
+        headers.update(integration_parameters.get("headers", {}))
+        if "Accept" not in headers:
+            headers["Accept"] = "application/json"
+
+        template = integration.get("requestTemplates").get(APPLICATION_JSON)
         if "GetQueueUrl" in template or "CreateQueue" in template:
             request_templates = RequestTemplates()
             payload = request_templates.render(invocation_context)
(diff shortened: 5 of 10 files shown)


bentsku (commented):
Thanks a lot for adding this!

I’m not sure what was the behaviour of AWS before SQS was able to return JSON, and if this is a result of it, or if they also convert the XML under the hood.  See https://github.com/localstack/localstack/issues/8267#issuecomment-1537933250 and https://github.com/localstack/localstack/pull/8268

Basically, I’m wondering if AWS uses JSON under the hood for the SQS response instead of parsing XML, or if their implementation of VTL is able to take XML as input. I suppose could test it with an HTTP integration and trying to do the same, and see if that works.  

I think it doesn’t hurt to be able to do it now, but I’m not sure if this should be only for SQS until we implement JSON, or if it should be part of every template rendering logic.

  After giving it a try in the AWS console, it seems AWS is unable to do so, it can’t parse XML as input from the integration response, I tried with http://httpbin.org/xml and the template was not being applied nor parsed. 

This was the API I tried it on with the nice test console.  
```yaml
openapi: "3.0.1"
info:
  title: "test-api-xml"
  version: "2023-09-09T15:26:38Z"
servers:
- url: "/{basePath}"
  variables:
    basePath:
      default: "test"
paths:
  /test:
    get:
      responses:
        "200":
          description: "200 response"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Empty"
      x-amazon-apigateway-integration:
        httpMethod: "GET"
        uri: "http://httpbin.org/xml"
        responses:
          default:
            statusCode: "200"
            responseTemplates:
              application/json: "#set($responseBody = $input.path('$.slideshow'))\n\
                #set($requestId = $input.path('$.slideshow.slide[0].title'))\n#set($messageId\
                \ = $slideshow.slide[1].title)\n{\n    \"title1\": \"$requestId\"\
                ,\n    \"title2\": \"$messageId\"\n}\n"
        passthroughBehavior: "when_no_match"
        type: "http"
components:
  schemas:
    Empty:
      title: "Empty Schema"
      type: "object"
```
The logs from the test console seemed to show it wasn't parsing the response, but if you'd like to give it a try too, feel free as I'm not 100% sure if I missed a step. I tried to look online and it seems people went to great length to be able to just only pass XML from a lambda. 

I also tried an SQS API, and you can see API Gateway is sending the `Accept=application/json` header, so SQS should respond in JSON too.  
Logs from the test console:
```
Sat Sep 09 15:57:46 UTC 2023 : Sending request to https://sqs.us-east-1.amazonaws.com/<account-id>/queue-518a623e
Sat Sep 09 15:57:46 UTC 2023 : Received response. Status: 200, Integration latency: 34 ms
Sat Sep 09 15:57:46 UTC 2023 : Endpoint response headers: {x-amzn-RequestId=315726e7-2463-5735-80bc-7b91bc59149e, X-Amzn-Trace-Id=Root=1-64fc95fa-630fac6d04ebca8e4d4f74ae, Date=Sat, 09 Sep 2023 15:57:46 GMT, Content-Type=application/json, Content-Length=312, connection=keep-alive}
Sat Sep 09 15:57:46 UTC 2023 : Endpoint response body before transformations: {"SendMessageResponse":{"ResponseMetadata":{"RequestId":"315726e7-2463-5735-80bc-7b91bc59149e"},"SendMessageResult":{"MD5OfMessageAttributes":null,"MD5OfMessageBody":"0c38abe3676e5cf7048b3e40018ea418","MD5OfMessageSystemAttributes":null,"MessageId":"a8e25ba3-2165-45d0-8e71-db475f6a4714","SequenceNumber":null}}}
```

So I’d be in favour to not have this as part of the global template rendering, but just in the SQS integration for now, until we can force the `Content-Type` to `application/json` to SQS so that it can return JSON natively so we could remove this workaround. What do you think? I don't know if other services using XML are possible to test?

localstack/services/apigateway/templates.py
```diff
@@ -282,6 +285,15 @@ def render(self, api_context: ApiInvocationContext, **kwargs) -> Union[bytes, st
...
         variables = self.build_variables_mapping(api_context)
+
+        # we need to convert xml (because airspeed doesn't support xml)
+        # to json before rendering the template
+        body = variables["input"]["body"]
+        if is_valid_xml(body):
+            xml_dict = xmltodict.parse(body)
+            xml_to_json_dict = json.dumps(xml_dict)
+            variables["input"]["body"] = xml_to_json_dict
+
```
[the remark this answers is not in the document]
calvernaz: I'll address the bigger picture in the main comment, but effectively what are we dealing with here, is a response template, thus the code should be in the response template.

calvernaz:
> Thanks a lot for adding this!

Starting from the bottom just for the sake of reproducibility.

> I also tried an SQS API, and you can see API Gateway is sending the `Accept=application/json` header, so SQS should respond in JSON too.   Logs from the test console:
> 
> ```
> Sat Sep 09 15:57:46 UTC 2023 : Sending request to https://sqs.us-east-1.amazonaws.com/<account-id>/queue-518a623e
> Sat Sep 09 15:57:46 UTC 2023 : Received response. Status: 200, Integration latency: 34 ms
> Sat Sep 09 15:57:46 UTC 2023 : Endpoint response headers: {x-amzn-RequestId=315726e7-2463-5735-80bc-7b91bc59149e, X-Amzn-Trace-Id=Root=1-64fc95fa-630fac6d04ebca8e4d4f74ae, Date=Sat, 09 Sep 2023 15:57:46 GMT, Content-Type=application/json, Content-Length=312, connection=keep-alive}
> Sat Sep 09 15:57:46 UTC 2023 : Endpoint response body before transformations: {"SendMessageResponse":{"ResponseMetadata":{"RequestId":"315726e7-2463-5735-80bc-7b91bc59149e"},"SendMessageResult":{"MD5OfMessageAttributes":null,"MD5OfMessageBody":"0c38abe3676e5cf7048b3e40018ea418","MD5OfMessageSystemAttributes":null,"MessageId":"a8e25ba3-2165-45d0-8e71-db475f6a4714","SequenceNumber":null}}}
> ```

You just said it there, ApiGW does send the "Accept" header as "application/json". In order to reproduce it, you would have to set the header to "application/xml" and you would get the following response:

request:
```
Sat Sep 09 18:28:24 UTC 2023 : Endpoint request headers: {Authorization=**********************************************************************************************************************************************************************************************************************************************************************77efbc, X-Amz-Date=20230909T182824Z, x-amzn-apigateway-api-id=g44647ybs7, Accept=application/xml, User-Agent=AmazonAPIGateway_g44647ybs7, X-Amz-Security-Token=IQoJb3JpZ2luX2VjEOv///////
```
response:
```
<?xml version="1.0"?><SendMessageResponse xmlns="http://queue.amazonaws.com/doc/2012-11-05/"><SendMessageResult><MessageId>0a772d86-0a15-45f8-872d-fb41018b411c</MessageId><MD5OfMessageBody>dcdf2bdd90cfccde643defa7df7c75a2</MD5OfMessageBody></SendMessageResult><ResponseMetadata><RequestId>8c6c3c94-673e-54d4-9db6-61e2ddb02bb4</RequestId></ResponseMetadata></SendMessageResponse>
```

> 
> I’m not sure what was the behaviour of AWS before SQS was able to return JSON, and if this is a result of it, or if they also convert the XML under the hood.  See [#8267 (comment)](https://github.com/localstack/localstack/issues/8267#issuecomment-1537933250) and #8268
> 
> Basically, I’m wondering if AWS uses JSON under the hood for the SQS response instead of parsing XML, or if their implementation of VTL is able to take XML as input. I suppose could test it with an HTTP integration and trying to do the same, and see if that works.  
> 
> I think it doesn’t hurt to be able to do it now, but I’m not sure if this should be only for SQS until we implement JSON, or if it should be part of every template rendering logic.
> 
>   After giving it a try in the AWS console, it seems AWS is unable to do so, it can’t parse XML as input from the integration response, I tried with http://httpbin.org/xml and the template was not being applied nor parsed.
> 

> I’m not sure what was the behaviour of AWS before SQS was able to return JSON, and if this is a result of it, or if they also convert the XML under the hood.  See [#8267 (comment)](https://github.com/localstack/localstack/issues/8267#issuecomment-1537933250) and #8268
> 
> Basically, I’m wondering if AWS uses JSON under the hood for the SQS response instead of parsing XML, or if their implementation of VTL is able to take XML as input. I suppose could test it with an HTTP integration and trying to do the same, and see if that works. 
 
So it seems this was a waste of time :) The problem stems from the issues you mentioned above and I can confirm AWS doesn't support xml transformations, so I wonder if this should be implemented it all.

bentsku (commented):
Interesting that we actually JSON from SQS! Looking at the PR of Alex I've linked earlier, it seems this point was marked as done:

> Implement a parser and serializer facade which can dynamically select the right parser/serializer based on the content type / accept headers of the request.

So content negotiation was in fact implemented! Thanks a lot for finding this out, this simplifies the PR a lot! 

However, I'm a bit worried about the `UpdateIntegration` and `UpdateIntegrationResponse` changes, could this break some user setup? What is the current state of those operation? Is it done in moto? Because how I see it, they would now throw exception for most of the calls? Where maybe before it would not have?

I think we have some patching going on for `Integration` (`apigateway_response_integrations` in `patches.py`), so I feel like the fix should either go there instead of implementing something not entirely which may break some Terraform setup, even though I know the current implementation is not great, or then be tested and moved to our provider. 

It seems like `UpdateIntegrationResponse` might not be implemented at all, so I'm not sure about this one: if they are no tests for it, I'd more in favour to not have it not raise specific exceptions if we're not certain they're being raised. So it would do what you want it do to, but if someone tries to do something that works against AWS but we raise exceptions, that could be a bit blocking. 

As an example, I think `UpdateIntegration` actually [supports more](https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html#UpdateIntegration-Patch): `/cacheKeyParameters`, `/requestParameters` and `/requestTemplates`. But now we would be raising exception if it's not `/requestParameters` only.

localstack/services/apigateway/provider.py
```diff
@@ -515,6 +515,100 @@ def _delete_children(resource_to_delete: str):
...
+            op = patch_operation.get("op")
+            path = patch_operation.get("path")
+            if "/requestParameters" not in path:
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op '{op}'. Must be one of: [/requestParameters]"
+                )
+            if op != "add":
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op '{op}'. Please choose supported operations"
+                )
```
bentsku, line 550: I think this is too strict, and would raise Exceptions wen AWS doesn't.

localstack/services/apigateway/provider.py
```diff
@@ -515,6 +515,100 @@ def _delete_children(resource_to_delete: str):
...
+            op = patch_operation.get("op")
+            path = patch_operation.get("path")
+            if "/responseTemplates" not in path:
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op '{op}'. Must be one of: [/requestParameters]"
+                )
+            if op != "remove":
+                raise BadRequestException(
+                    f"Invalid patch path  '{path}' specified for op '{op}'. Please choose supported operations"
+                )
```
bentsku, line 598: Same with this, this is too strict and could break user setup

localstack/services/apigateway/provider.py
```diff
@@ -515,6 +515,100 @@ def _delete_children(resource_to_delete: str):
...
+
+            # for path "/requestParameters/integration.request.header.Accept"
+
+            if "/requestParameters/integration.request.header" in path:
+                value = patch_operation.get("value")
+                if not isinstance(value, str):
+                    raise BadRequestException(
+                        f"Invalid patch value  '{value}' specified for op '{op}'. Must be a string"
+                    )
```
bentsku, line 559: does it need to have a `value` if it's a remove call?

localstack/services/apigateway/provider.py
calvernaz: yep it's strict because this is just a basic implementation to make to test pass. And does not implement any other operation.

localstack/services/apigateway/provider.py
calvernaz: good catch, it does not, my test add a value, that's why it does not break. will fix it!

localstack/services/apigateway/provider.py
calvernaz: too strict because because we don't implement other operations. I followed the same idea of `update_resource` you did given your experience in the patch zone.

calvernaz:
> Interesting that we actually JSON from SQS! Looking at the PR of Alex I've linked earlier, it seems this point was marked as done:
> 
> > Implement a parser and serializer facade which can dynamically select the right parser/serializer based on the content type / accept headers of the request.
> 
> So content negotiation was in fact implemented! Thanks a lot for finding this out, this simplifies the PR a lot!
> 
> However, I'm a bit worried about the `UpdateIntegration` and `UpdateIntegrationResponse` changes, could this break some user setup? What is the current state of those operation? Is it done in moto? Because how I see it, they would now throw exception for most of the calls? Where maybe before it would not have?
> 
> I think we have some patching going on for `Integration` (`apigateway_response_integrations` in `patches.py`), so I feel like the fix should either go there instead of implementing something not entirely which may break some Terraform setup, even though I know the current implementation is not great, or then be tested and moved to our provider.
> 
> It seems like `UpdateIntegrationResponse` might not be implemented at all, so I'm not sure about this one: if they are no tests for it, I'd more in favour to not have it not raise specific exceptions if we're not certain they're being raised. So it would do what you want it do to, but if someone tries to do something that works against AWS but we raise exceptions, that could be a bit blocking.
> 
> As an example, I think `UpdateIntegration` actually [supports more](https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html#UpdateIntegration-Patch): `/cacheKeyParameters`, `/requestParameters` and `/requestTemplates`. But now we would be raising exception if it's not `/requestParameters` only.

It does support more, but we don't I notice this by actually applying this in the test and figure there was no implementation. 

> Interesting that we actually JSON from SQS! Looking at the PR of Alex I've linked earlier, it seems this point was marked as done:
> 
> > Implement a parser and serializer facade which can dynamically select the right parser/serializer based on the content type / accept headers of the request.
> 
> So content negotiation was in fact implemented! Thanks a lot for finding this out, this simplifies the PR a lot!
> 
> However, I'm a bit worried about the `UpdateIntegration` and `UpdateIntegrationResponse` changes, could this break some user setup? What is the current state of those operation? Is it done in moto? Because how I see it, they would now throw exception for most of the calls? Where maybe before it would not have?
> 
> I think we have some patching going on for `Integration` (`apigateway_response_integrations` in `patches.py`), so I feel like the fix should either go there instead of implementing something not entirely which may break some Terraform setup, even though I know the current implementation is not great, or then be tested and moved to our provider.
> 
> It seems like `UpdateIntegrationResponse` might not be implemented at all, so I'm not sure about this one: if they are no tests for it, I'd more in favour to not have it not raise specific exceptions if we're not certain they're being raised. So it would do what you want it do to, but if someone tries to do something that works against AWS but we raise exceptions, that could be a bit blocking.
> 
> As an example, I think `UpdateIntegration` actually [supports more](https://docs.aws.amazon.com/apigateway/latest/api/patch-operations.html#UpdateIntegration-Patch): `/cacheKeyParameters`, `/requestParameters` and `/requestTemplates`. But now we would be raising exception if it's not `/requestParameters` only.

It does support more, no questions about that, but we don't. I figure this my implementing a test that was updating the integration and integration response https://github.com/localstack/localstack/pull/9102/files#diff-357303ef2a04cdb9ff3ebd56f3a8293cd23b26e77a5e8ef0fd72095547076d78R243-R256

bentsku:
> Overall, we want to move away from these patches as much as possible. Hard to follow and it's just builds on top of a model we are splitting way anyway

Totally agree, was suggesting the patch route as there was already an implementation, might not be working but I figured that'd be faster for you. 

Overall I just don't want to raise exception if we don't support it at the moment and we did not raise exception before.
I know it'd be better to fail if we don't support it, especially for `UpdateIntegration`, but as we already have something there that does not raise exception, I'm worried of breaking existing user setup (that might not be working perfectly I agree) and blocking them when deploying. It would be ok to just do nothing in that case if we don't support it.

localstack/services/apigateway/provider.py
bentsku: The implementation is good, it's just that the exceptions raised in `update_resource` are raised by AWS too, and I'm worried it's not the case here.

localstack/services/apigateway/patches.py
```diff
@@ -70,9 +70,30 @@ def apigateway_response_integrations(fn, self, request, *args, **kwargs):
...
+
+        integration = self.backend.get_integration(function_id, resource_id, method_type)
+        if not integration:
+            return result
+
+        if self.method == "PATCH":
+            patch_operations = self._get_param("patchOperations")
+            apply_json_patch_safe(integration.to_json(), patch_operations, in_place=True)
+            return 200, {}, json.dumps(integration.to_json())
+
```
bentsku, line 96: q: do we need this? or is it implemented with `update_integration_response` in the provider? I'm not sure I follow as it seems we're patching the whole `integration` here and not the `integration_response`, and we're patching only the result of `to_json()` and not the entity in itself. 
It seems from moto this method would be more like: 
```python
integration_response = self.backend.get_integration_response(
    function_id, resource_id, method_type, status_code
)
```
I think this might not be called as we would use the provider operation instead.

bentsku (commented):
Thanks a lot for removing the exceptions raised! Just a minor comment regarding the new patch to `integration_responses`, I think this might not be needed as we're using the provider for it, but other than that we're good!

bentsku (approved):
LGTM! Thanks for addressing all the comments! 🚀

whummer (approved):
Looks great, glad we were able to get to the bottom of this and come up with a solid fix, after a few iterations. 🚀 Kudos for keeping track of all the conversation in detail, which made it easy to follow the process 🙌

localstack/services/apigateway/helpers.py
```diff
@@ -551,6 +552,15 @@ def get_stage_variables(context: ApiInvocationContext) -> Optional[Dict[str, str
 # ---------------
 
 
+def is_valid_xml(xml_string):
```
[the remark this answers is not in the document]
calvernaz: will do!

PR #9104, opened 2023-09-11
fix APIGW DeleteDeployment to verify if stage is assigned to it

bentsku:
## Motivation
Follow up from #9095

The scenario described by the user was the following:
1. Create api gateway
2. Create api gateway stage
3. Create api gateway stage deployment
4. Create another api gateway stage deployment (which updates the stage itself)

## Changes
The issue was in the moto implementation: 
```python
        if deployment.stage_name and deployment.stage_name in self.stages:
            # Stage is still active
            raise StageStillActive()
```
But it didn't check if the actual stage which had the same name was attached to the deployment we were deleting. 
I've created a patch as we don't have any `deployment` operation in our provider at the moment and this is a quick fix, the issue was long running and hindering their Terraform deployments.
Added an AWS test which was reproducing the issue.

### DIFF
--- a/localstack/services/apigateway/patches.py
+++ b/localstack/services/apigateway/patches.py
@@ -3,8 +3,10 @@ import logging
 
 from moto.apigateway import models as apigateway_models
 from moto.apigateway.exceptions import (
+    DeploymentNotFoundException,
     NoIntegrationDefined,
     RestAPINotFound,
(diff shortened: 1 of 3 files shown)


calvernaz (approved):
a patch it is :( but LGTM

PR #9107, opened 2023-09-11
Adopt lambda dns setting, fix dns in host mode

dfangl:
## Motivation
We want to enable our DNS integration for all lambda images. In community we only forward all subdomains of `localhost.localstack.cloud` to LS, which is already helpful when using lambdas to connect to returned endpoints (e.g. apigateway endpoints).

Also, the host mode version is currently broken, so this has to be fixed.

## Changes
* Fix host mode DNS by creating a new server class for dns servers running in a separate (sudo) process. It is not possible to interact with it programatically yet, but it now it does not raise errors when calling DNSServerProtocol methods
* Set the DNS in lambda, if no other LAMBDA_DOCKER_DNS is set.

### DIFF
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -835,6 +835,7 @@ def start_server(upstream_dns: str, host: str, port: int = config.DNS_PORT):
         dns_server.shutdown()
         return
     DNS_SERVER = dns_server
+    LOG.debug("DNS server startup finished!")
 
 
 def stop_servers():
@@ -858,6 +859,7 @@ def start_dns_server_as_sudo(port: int):
         return
 
     DNS_SERVER = dns_server
+    LOG.debug("DNS server startup finished (as sudo)!")
 
 
 def start_dns_server(port: int, asynchronous: bool = False, standalone: bool = False):
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -698,7 +698,7 @@ class DnsServer(Server, DnsServerProtocol):
 class SeparateProcessDNSServer(Server, DnsServerProtocol):
     def __init__(
         self,
-        port: int,
+        port: int = 53,
         host: str = "0.0.0.0",
     ) -> None:
         super().__init__(port, host)
@@ -835,7 +835,7 @@ def start_server(upstream_dns: str, host: str, port: int = config.DNS_PORT):
         dns_server.shutdown()
         return
     DNS_SERVER = dns_server
-    LOG.debug("DNS server startup finished!")
+    LOG.debug("DNS server startup finished.")
 
 
 def stop_servers():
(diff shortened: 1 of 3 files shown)


joe4dev (commented):
The changes here in community look good to me.

Added some comments/questions and more comments in ext ...

localstack/dns/server.py
```diff
@@ -695,6 +695,48 @@ def clear(self):
...
+    def do_start_thread(self):
+        # For host mode
+        env_vars = {}
+        for env_var in config.CONFIG_ENV_VARS:
+            if env_var.startswith("DNS_"):
+                value = os.environ.get(env_var, None)
+                if value is not None:
+                    env_vars[env_var] = value
+
+        # note: running in a separate process breaks integration with Route53 (to be fixed for local dev mode!)
```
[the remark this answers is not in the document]
simonrw: It is more that Route53 interacts with the DNS server in memory to configure the DNS server after startup. This requires the DNS server to be running in the same process as Route53, but this is not the case in host mode.

localstack/dns/server.py
```diff
@@ -695,6 +695,48 @@ def clear(self):
         self.resolver.clear()
 
 
+class SeparateProcessDNSServer(Server, DnsServerProtocol):
+    def __init__(
+        self,
+        port: int,
```
simonrw, line 701: minor: default to 53?

localstack/dns/server.py
```diff
@@ -695,6 +695,48 @@ def clear(self):
...
+        host: str = "0.0.0.0",
+    ) -> None:
+        super().__init__(port, host)
+
+    @property
+    def protocol(self):
+        return "udp"
```
simonrw, line 708: Strictly not true, right? We accept TCP and UDP. Not that it matters, since the protocol is only used for the url which we don't use anyway.

localstack/dns/server.py
```diff
@@ -793,15 +835,40 @@ def start_server(upstream_dns: str, host: str, port: int = config.DNS_PORT):
         dns_server.shutdown()
         return
     DNS_SERVER = dns_server
+    LOG.debug("DNS server startup finished!")
```
simonrw, line 838: nit: I feel the exclamation mark is not inline with our current messages
```suggestion
    LOG.debug("DNS server startup finished")
```

simonrw (approved):
Looks great! 

A quick question I cannot add as a line comment: on `localstack/dns/server.py` new line 886 we call `port_can_be_bound` which checks `0.0.0.0`, but should probably check the value of `host`. This can be done by changing the call to 

```python
port_can_be_bound(Port(port, "udp"), address=host)
```

dominikschubert:
Shouldn't we now also add a port mapping for the default DNS port into the community `docker-compose.yaml`? :thinking: 

https://github.com/localstack/localstack/blob/e80de4c55d952185c3152f39177aec7f11d0cf8f/docker-compose.yml#L1-L15

dfangl:
For this PR, it is not strictly necessary. However, good point. We should discuss whether that makes sense. In general, for community, the DNS server is mostly relevant for `localhost.localstack.cloud`, which should resolve to 127.0.0.1 anyway. For Pro it is more relevant, if people want to use transparent endoint injection from the host.

PR #9109, opened 2023-09-11
add container port and volume flags to localstack start

thrau:
## Motivation

We want to make it easier to add customizations when running `localstack start`. For instance, to generalize something like `DNS_ADDRESS`, we could just provide a port mapping for `localstack start`, that allows people to map, say, `127.0.0.1:1053:53/udp`. Similarly, there's now easy way to mount custom volumes into the localstack container other than using `DOCKER_FLAGS`.

This PR adds a `-p` and `-v` flag to `localstack start`, which are the same as the flags used in `docker run`. This is a continuation of #8772

## Changes

* Add `-p/--publish` and `-v/--volume` flags to `localstack start`
* Re-use and refactor the necessary code from dev configurators (was already there) into the bootstrap `ContainerConfigurators`
* Fixed an issue where volume parsing wasn't working correctly if `:ro` suffix was provided
* Fixed an issue in ports parsing that would prevent a `/<protocol>` suffix or port range `<start>-<end>` to be parsed correctly
* Added a unit test for the configurator

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -396,9 +396,30 @@ def _print_service_table(services: Dict[str, str]) -> None:
     multiple=True,
     required=False,
 )
+@click.option(
+    "--publish",
+    "-p",
+    help="Additional port mappings that are passed to the LocalStack container",
+    multiple=True,
+    required=False,
+)
+@click.option(
+    "--volume",
+    "-v",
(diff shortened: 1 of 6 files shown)


alexrashed (approved):
Awesome! Great to see the the `ContainerConfigurators` being adopted, and a new bootstrap test! 🚀 
Should there maybe also be a new or enhanced CLI test?

PR #9114, opened 2023-09-12
Improve caching for Community Integration Tests against Pro

dominikschubert:
## Motivation

Currently the `Community Integration Tests against Pro` are the slowest part of our community test suite on PRs.
In this PR we attempt to reduce the runtime of these tests to get them to a point where we don't have to wait an additional 20+ minutes until the pro tests against community are done for a green pipeline.

The initial commit will currently never find a cache hit since we don't execute the action on master. Additionally the cache key seems to be calculated incorrectly by considering too many files for caching and never actually restored the venv/lib cache, even in a single PR with multiple runs.

Successful caching of the built lambda common packages should already save ~8 minutes.

## Changes

- stricter matching/hash calculation
- Add a restore key for the venv/lib caching to match cache from master
- execute action on master again to populate cache for PRs to use
- set `DISABLE_BOTO_RETRIES` for test execution

What's left to do:

- [ ] run scheduled on master instead of push-based
- [ ] we could actually even skip the tests on master again and just run the installers manually

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -38,6 +38,20 @@ on:
       - master
       - 'v[0-9]+'
       - release/*
+  push:
+    paths:
+      - ".github/workflows/tests-pro-integration.yml"
+      - "localstack/**"
+      - "tests/**"
+      - "setup.py"
+      - "pyproject.toml"
+      - "setup.cfg"
+      - "Dockerfile"
+      - "Dockerfile.rh"
+      - "docker-compose.yml"
+      - "bin/**"
+    branches:
+      - master
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


localstack/aws/connect.py
```diff
@@ -49,7 +49,6 @@ def my_patch(fn, self, **kwargs):
         patched_kwargs = {
             **kwargs,
             "WaiterConfig": {
-                # TODO: make these configurable
```
alexrashed, line 52: This change doesn't seem to be related, but if it's on purpose, it's okay to sneak this one in ;)

.github/workflows/tests-pro-integration.yml
```diff
@@ -169,11 +181,15 @@ jobs:
...
           # include the matrix group (to re-use the var-libs used in the specific test group)
-          key: community-it-${{ runner.os }}-python-${{ steps.setup-python.outputs.python-version }}-venv-${{ hashFiles('localstack-ext/setup.cfg', 'localstack-ext/pyproject.toml', 'localstack/setup.cfg', '**/packages.py', '**/packages/*') }}-${{steps.determine-companion-ref.outputs.result}}-group-${{ matrix.group }}
+          key: community-it-${{ runner.os }}-python-${{ steps.setup-python.outputs.python-version }}-venv-${{ hashFiles('localstack-ext/setup.cfg', 'localstack-ext/pyproject.toml','localstack-ext/**/packages.py','localstack-ext/packages/*.py', 'localstack/setup.cfg', 'localstack/**/packages.py',  'localstack/packages/*.py') }}-${{steps.determine-companion-ref.outputs.result}}-group-${{ matrix.group }}
+          restore-keys: |
+            community-it-${{ runner.os }}-python-${{ steps.setup-python.outputs.python-version }}-venv-${{ hashFiles('localstack-ext/setup.cfg', 'localstack-ext/pyproject.toml','localstack-ext/**/packages.py','localstack-ext/packages/*.py', 'localstack/setup.cfg', 'localstack/**/packages.py',  'localstack/packages/*.py') }}-master-
```
alexrashed, line 191: I think we should either implement the TODO you added (split the caches, one for the venv - used by both groups, and one for the package installers - used for each group invidually), or we have to add the group to the restore key as well.
Otherwise they would overwrite each other.
Also, I think the `${{steps.determine-companion-ref.outputs.result}}` would evaluate to `refs/heads/master` (and not `master`).

So I think the `restore-keys` should be changed to:

```suggestion
          restore-keys: |
            community-it-${{ runner.os }}-python-${{ steps.setup-python.outputs.python-version }}-venv-${{ hashFiles('localstack-ext/setup.cfg', 'localstack-ext/pyproject.toml','localstack-ext/**/packages.py','localstack-ext/packages/*.py', 'localstack/setup.cfg', 'localstack/**/packages.py',  'localstack/packages/*.py') }}-refs/heads/master-group-${{ matrix.group }}
```

alexrashed (changes requested):
Thanks a lot for jumping on this! I think we have to adjust the restore key a bit, but then this should greatly speed up the community tests in our PRs! 🚀 ⌛

alexrashed (approved):

PR #9117, opened 2023-09-12
Validate and refactor cloudwatch tests

dominikschubert:
## Motivation

These tests were quite flaky in CI, so I tried to make them a bit more robust now :sweat_smile: 

## Changes

- Remove all unknown markers and replace with proper markers.
- Any tests that would've been `needs_fixing` have been adapted/fixed to be compatible with AWS.
- Skip `test_put_metric_alarm` instead of xfailing because it always fails with localstack and the test is usually running for 3 minutes.

## TODO

What's left to do:

- [x] might need to adjust snapshots due to transformer change

### DIFF
--- a/localstack/testing/snapshots/transformer_utility.py
+++ b/localstack/testing/snapshots/transformer_utility.py
@@ -483,6 +483,8 @@ class TransformerUtility:
         """
         return [
             TransformerUtility.key_value("AlarmName"),
+            TransformerUtility.key_value("MetricName"),
+            TransformerUtility.key_value("Namespace"),
             KeyValueBasedTransformer(_resource_name_transformer, "SubscriptionArn"),
             TransformerUtility.key_value("Region", "region-name-full"),
         ]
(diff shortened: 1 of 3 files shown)


localstack/testing/snapshots/transformer_utility.py
```diff
@@ -483,6 +483,8 @@ def cloudwatch_api():
         """
         return [
             TransformerUtility.key_value("AlarmName"),
+            TransformerUtility.key_value("MetricName"),
```
steffyP, line 486: NIT: i think using a random `Namespace` should be sufficient. The `MetricName` could stay the same for most test cases.

steffyP (commented):
Thanks for fixing + improving the tests! 🙏 🙇‍♀️ 
I am quite certain some tests will fail because of the `Namespace` and `MetricName` replacement.

localstack/testing/snapshots/transformer_utility.py
dominikschubert: Yeah was thinking the same but only after I finished it :grimacing: 
Will adjust

Issue #8205, opened 2023-04-26
bug: Stepfunctions creates resources in the fallback account


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Stepfunctions v2 provider creates resources in the fallback account `000000000000` and not the account made by the original requester.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Start with `PROVIDER_OVERRIDE_STEPFUNCTIONS=v2`

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```sh
$ AWS_ACCESS_KEY_ID=111111111111 awslocal iam create-role \
   --role-name example-role \
   --assume-role-policy-document '{"Version": "2012-10-17", "Statement": {"Action": "sts:AssumeRole", "Effect": "Allow", "Principal": {"Service": "states.amazonaws.com"}}}'

$ AWS_ACCESS_KEY_ID=111111111111 awslocal iam create-policy \
   --policy-name example-policy \
   --policy-document '{"Version": "2012-10-17", "Statement": {"Action": "sns:createTopic", "Effect": "Allow", "Resource": "*"}}'

$ AWS_ACCESS_KEY_ID=111111111111 awslocal iam attach-role-policy \
   --role-name example-role \
   --policy-arn arn:aws:iam::111111111111:policy/example-policy

$ AWS_ACCESS_KEY_ID=111111111111 awslocal stepfunctions create-state-machine \
   --name example \
   --definition '{"StartAt": "CreateTopicTask", "States": {"CreateTopicTask": {"End": true, "Type": "Task", "Resource": "arn:aws:states:::aws-sdk:sns:createTopic", "Parameters": {"Name.$": "$.Name"}}}}' \
   --role-arn arn:aws:iam::111111111111:role/example-role

$ AWS_ACCESS_KEY_ID=111111111111 awslocal stepfunctions start-execution \
   --state-machine-arn arn:aws:states:us-east-1:111111111111:stateMachine:example \
   --input '{"Name": "example-topic"}'

# When execution finishes, it would be expected that the topic would be created in the the original account

$ AWS_ACCESS_KEY_ID=111111111111 awslocal sns list-topics                                                                                                                                  
{
    "Topics": []
}

# but in fact it is created in the fallback account

[...]

PR #9119, opened 2023-09-12
StepFunctions: Multi-accounts compatibility

viren-nadkarni:
## Motivation

This PR introduces multi-account support to the legacy StepFunctions provider. It also makes multi-account friendly fixes to the work-in-progress StepFunctions v2 provider.

cc: @MEPalma 

## Implementation

The legacy provider uses [StepFunctions Local](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-jar.html) which itself does not support namespacing. So we take an approach similar to Kinesis. Each account+region combo will run its own instance of SFLocal. Of course this is not optimal but should fulfill simple usecases. 

However, SFLocal does not respect the `--account-id` argument and overwrites ARNs like Lambda Functions during invocations. So I guess the way forward is to only support true multi-accounts in the new SF provider and keep this limitation in the legacy one. The multi-region parameter is maintained.

Criteria for merging: all stepfunctions tests (legacy as well as the **ci/circleci: itest-sfn-v2-provider** job)  pass when `AWS_ACCESS_KEY_ID` is set to a non-default account ID or when `AWS_REGION_NAME` is set to non-`us-east-1` region.

**79ec46e61e52940f2be5d2377b4933a77e5c2b61 must be reverted before merge**

## Related

Closes https://github.com/localstack/localstack/issues/8205


localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_dynamodb.py
```diff
@@ -124,8 +123,7 @@ def _from_error(self, env: Environment, ex: Exception) -> FailureEvent:
 
     def _eval_service_task(self, env: Environment, parameters: dict) -> None:
         api_action = camel_to_snake_case(self.resource.api_action)
-
-        dynamodb_client = connect_to(config=Config(parameter_validation=False)).dynamodb
+        dynamodb_client = get_boto_client(env, "dynamodb")
```
MEPalma, line 126: Is this still ensuring the parameters validation is off?

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/resource.py
```diff
@@ -85,10 +84,12 @@ def __init__(self, resource_arn: ResourceARN):
         self.account = resource_arn.account
 
     @staticmethod
-    def from_resource_arn(arn: str) -> Resource:
+    def from_resource_arn(account_id: str, region_name: str, arn: str) -> Resource:
```
MEPalma, line 87: Runtime values are ideally evaluated during execution. So resource should become an EvalComponent, evaluate its region and account in `_eval_body` using the environment, and push to the stack a resource object that includes all the values.

localstack/services/stepfunctions/asl/eval/environment.py
```diff
@@ -26,8 +26,14 @@
 
 
 class Environment:
-    def __init__(self, context_object_init: ContextObjectInitData):
+    def __init__(
+        self, account_id: str, region_name: str, context_object_init: ContextObjectInitData
+    ):
         super(Environment, self).__init__()
+
```
MEPalma, line 33: I think we should decorate these in an object where the environment can expose runtime values like these 2.

localstack/services/stepfunctions/asl/parse/preprocessor.py
```diff
@@ -191,6 +191,10 @@
 
 
 class Preprocessor(ASLParserVisitor):
+    def __init__(self, account_id: str, region_name: str):
```
MEPalma, line 194: Adding the values to the environment is a decent solution, and evaluation components can produce runtime values like region and account during evaluation. So I don't think we should pollute the Preprocessor with runtime values, as conceptually it just builds the evaluation tree.

MEPalma (changes requested):
Thank you for the changes! I would like the separation between building the evaluation tree and evaluating runtime values to remain consistent. We achieve this with Component and EvalComponent classes. Adding these runtime values to the environment is valid, but not in the preprocessor. Also a question about parameter validation of the new clients.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_dynamodb.py
viren-nadkarni: Yes! Please make sure you use this helper to get the Boto client and not `connect_externally_to`

https://github.com/localstack/localstack/blob/aad9b80492fdc9f7b22f48af0184cb5febf6dd23/localstack/services/stepfunctions/backend/utils.py#L7-L13

localstack/services/stepfunctions/asl/parse/preprocessor.py
viren-nadkarni: When the `Preprocessor` constructs the `Resource` object from ARN, the ARN does not necessarily contain the region and the account ID. So it is necessary that the correct account ID and region from the request context are passed down.

https://github.com/localstack/localstack/blob/aad9b80492fdc9f7b22f48af0184cb5febf6dd23/localstack/services/stepfunctions/asl/parse/preprocessor.py#L238

Using `aws_stack.get_region()` is not an option as it doesn't always have the proper region. Moreover, I'll be deprecating that method in the future for this exact reason.

MEPalma (approved):
Resources are now evaluation components for `ResourceRuntimePart` which includes the region and account. This means that the request's account and region are used if none is declared in the resource Arn. All other staticResource properties are still accessible from the declaration objects themselves.

PR #9120, opened 2023-09-12
Deprecate aws_stack.generate_presigned_url

viren-nadkarni:
This PR marks the `aws_stack.generate_presigned_url` helper as obsolete.

The recommended way forward is to use the `generate_presigned_url` method of the `S3` Boto client.


localstack/utils/aws/aws_stack.py
```diff
@@ -418,6 +418,11 @@ def get_s3_hostname():
 
 
 def generate_presigned_url(*args, **kwargs):
+    warnings.warn(
+        "`generate_presigned_url` is obsolete. Use the Boto client `generate_presigned_url` method",
```
dominikschubert, line 422:
```suggestion
        "`aws_stack.generate_presigned_url` is obsolete. Use the Boto client `generate_presigned_url` method",
```

Might be a bit clearer, but if that clashes with any other existing deprecation warnings, feel free to ignore.

dominikschubert (approved):
:+1:

localstack/utils/aws/aws_stack.py
viren-nadkarni: Good idea :+1:

PR #9121, opened 2023-09-12
implement S3 cross account across all operations

bentsku:
## Motivation
Issues with S3 Website were found from #9017, which showed a deeper issue in our S3 cross/multi account implementation.
`GetBucketWebsite` would return `NoSuchBucket` if a request from account A was done to a bucket in account B, even though you could get an object from account B.

We need to implement proper cross-account access for S3.

#### For the current provider:
Before, if we tried to access a bucket existing in account A from account B, we would get some weird issue, as we got the bucket from moto but we would then access the store from account B, which wouldn't have the bucket properties. We now use the found moto bucket account id and region to fetch the store.

#### For the v3 provider:
We now use the cross account access for every S3 operation, which will allow us to have some implementation of ACL if not using bucket policies, instead of returning `NoSuchBucket` if the bucket didn't exist in the caller account. This will need multi account setup in our test suite at some point? 

#### Misc
Added a quick cross region test to verify it, and we didn't have one earlier for bucket operations. 

~Also sneaked in a little change for the request, I would get a big logging error that the account_id and region were not passed to the logging format string, following #9084 @whummer. This might just return `None/None`, not sure if that would still be helping or if we should refactor the format string?~ deleted as it's now part of another PR.

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -258,9 +258,8 @@ class S3Provider(S3Api, ServiceLifecycleHook):
             s3_notif_ctx = S3EventNotificationContext.from_request_context(
                 context, key_name=key_name
             )
-        if notification_config := self.get_store(
-            context.account_id, context.region
-        ).bucket_notification_configs.get(s3_notif_ctx.bucket_name):
+        store = self.get_store(s3_notif_ctx.bucket_account_id, s3_notif_ctx.bucket_location)
+        if notification_config := store.bucket_notification_configs.get(s3_notif_ctx.bucket_name):
             self._notification_dispatcher.send_notifications(s3_notif_ctx, notification_config)
 
     def _verify_notification_configuration(
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -11,7 +11,6 @@ from zoneinfo import ZoneInfo
 import moto.s3.responses as moto_s3_responses
 
 from localstack import config
-from localstack.aws.accounts import get_aws_account_id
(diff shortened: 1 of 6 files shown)


localstack/services/s3/provider.py
```diff
@@ -212,14 +212,11 @@ def get_full_default_bucket_location(bucket_name):
 
 class S3Provider(S3Api, ServiceLifecycleHook):
     @staticmethod
-    def get_store(context: Optional[RequestContext] = None) -> S3Store:
-        if not context:
-            return s3_stores[get_aws_account_id()][aws_stack.get_region()]
+    def get_store(account_id: Optional[str] = None, region: Optional[str] = None) -> S3Store:
+        return s3_stores[account_id or DEFAULT_AWS_ACCOUNT_ID][region or AWS_REGION_US_EAST_1]
```
bentsku, line 216: note: not sure this is needed, but as it was optional before and I didn't check for all usages, better be safe than sorry for now. The new provider doesn't allow optional values.

viren-nadkarni:
We will have some conflicts with https://github.com/localstack/localstack/pull/9023 where I also introduce the account ID in `S3EventNotificationContext`.

Could we wait until that one is merged? I'm happy to take this PR over the line after that.

bentsku:
@viren-nadkarni sure, no problem. I'd like to keep the difference in name between `caller_account_id` and the bucket account id though. I can rebase once your PR is merged, no problem. 😄

viren-nadkarni (approved):

viren-nadkarni:
@bentsku: my PR will take a bit more work before it is merged. You can go ahead and merge this one, I'll fix any conflicts on my end :+1:

PR #9125, opened 2023-09-12
add service_name attribute to AssetDirectory

thrau:
## Motivation

In the `StateVisitor` implementations that handle cloud pod state, there is currently no way to know which service the `AssetDirectory` being visited belongs to, unlike `AccountRegionBundle` and `BackendDict` which both have a `service_name` attribute.
Knowing the service name for an `AssetDirectory` is required for technical reasons for s3 v3 persistence.

## Changes

* Make `StateContainer` a protocol and make `service_name` a mandatory attribute
* Refactor all instances of `AssetDirectory` to correctly use the service name


bentsku (approved):
LGTM! Awesome, thank you so much for adding this. I'm sure it will allow us to have more flexibility regarding persistence, especially cloud pods when not having persistence enabled!

PR #9129, opened 2023-09-13
fix gateway shutdown procedure

thrau:
## Motivation

With providers now depending on other services for their internal functionality, it becomes important that we allow services to continue to use other services during the infra shutdown procedure so they can properly shut down themselves. This means we need to shutdown services *before* the gateway so services can still make internal requests to properly shut down This was previously not possible because of how shutdown handlers were implemented.

This PR adds a band-aid for this issue applied until we complete re-write the `start_infra` procedure.

Moreover, if we shut down the gateway last, then we should be rejecting incoming requests while we're shutting down the infra. This PR adds a handler that implements this logic. It accepts internal requests so internal components can continue working during the shutdown procedure.

## Changes

* Move service plugin shutdown into a `infra_shutdown` hook with a well-known priority of `-10`
* Add a new `ShutdownHandler` instance that is executed with a lower priority so we have a way to execute shutdown handlers after services were shut down
* Add new handler into the handler chain to return 503 while the system is shutting down
* Move `handlers.add_internal_request_params` almost to the top of the handler chain, since knowing whether it's an internal request is needed in the handler
* Add shutdown handlers to list of plugin managers that can be queried via `/_localstack/plugins`

### DIFF
--- a/localstack/aws/app.py
+++ b/localstack/aws/app.py
@@ -27,6 +27,8 @@ class LocalstackAwsGateway(Gateway):
         self.request_handlers.extend(
             [
                 handlers.push_request_context,
+                handlers.add_internal_request_params,
+                handlers.handle_runtime_shutdown,
                 metric_collector.create_metric_handler_item,
                 handlers.preprocess_request,
                 handlers.parse_service_name,  # enforce_cors and content_decoder depend on the service name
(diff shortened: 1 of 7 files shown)


alexrashed (approved):
Thanks for jumping on this! The implementation with the prioritized shutdown handlers looks really nice and clean. 💯

Issue #9126, opened 2023-09-12
bug: Caching of boto3 clients in forwarding mechanism to broken causing major memory leak.


emopti-jrufer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`localstack.aws.connect.ClientFactory._get_client` is annotated with `@cache` which relies on the hash values of the arguments as the cache key. After debugging I discovered that the hash for the `config` argument from `localstack.aws.connect.InternalClientFactory.get_client` is always changing resulting in cache miss. This causes a new client to be created for every forwarding request. Every new client get added to the cache which has no eviction causing the memory leak. Initially noticed this using Kinesis streams due to the high volume of requests consumed about .5gb per hour. This was most likely introduced by #7984.

It may also be the cause of the following issues.
#9108
#9025
#8837

### Expected Behavior

The boto3 clients should be reused and memory consumption should not continue to grow.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)
Initially docker run localstack/localstack then debugged running inside of PyCharm.

#### Client commands

The following script reproduces the issue.

```shell
#!/usr/bin/env bash

stream_name=streamabc
aws --region us-west-2 --endpoint-url http://localhost:4566 kinesis create-stream --stream-name ${stream_name} --shard-count 1

shard_iterator=`aws --region us-west-2 --endpoint-url http://localhost:4566 kinesis get-shard-iterator --stream-name ${stream_name} --shard-id shardId-000000000000 --shard-iterator-type LATEST --query 'ShardIterator' --output text`
echo ${shard_iterator}

for i in {1..10000} ; do
    shard_iterator=`aws --region us-west-2 --endpoint-url http://localhost:4566 kinesis get-records --stream-arn arn:aws:kinesis:us-west-2:000000000000:stream/${stream_name} --shard-iterator ${shard_iterator} --query 'NextShardIterator' --output text`
    echo ${shard_iterator}
done
```

### Environment

```markdown
- OS: macOS
- LocalStack: latest
```

dfangl:
Thank you for your comprehensive bug report!
We can confirm this issue, and are working on a fix. The main problem is botocore.Config objects not being hashable, and therefore missing the cache every time. 
We are working on a fix for this!

emopti-jrufer:
@dfangl I pulled the latest and have verified the memory consumption has declined and stabilized.

[...]

PR #9130, opened 2023-09-13
Fix cache misses for Boto Client

viren-nadkarni:
## Motivation

The caching mechanism of `ClientFactory._get_client()` relies on the hash value of all its args. This includes a Boto `Config` object.

The use of `Config.merge()` which returned a new `Config` object everytime, causing cache misses in all invocations of `ClientFactory._get_client()`.

## Implementation

This PR introduces a dedicated merge method for `Config` objects that must be used for all Boto `Config` merges outside of `ClientFactory._get_client()`.

@dfangl: Feel free to change this if you think of a better way :)

## Related

https://github.com/localstack/localstack/issues/9126

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -44,6 +44,9 @@
 /localstack/services/stores.py @viren-nadkarni
 /tests/unit/test_stores.py @viren-nadkarni
 
+# Internal Boto client lib
+/localstack/aws/connect.py @viren-nadkarni @dfangl
+
 # Dockerfile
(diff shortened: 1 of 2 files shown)


viren-nadkarni:
Post this fix, there are no misses reported: 
```
>>> self._get_client.cache_info()
CacheInfo(hits=1209, misses=1, maxsize=None, currsize=1)
```
However I still observe a gradual memory increase with a simple looped Describe operation

PR #9134, opened 2023-09-13
Make botocore config object hashable

dfangl:
## Motivation
Alternative solution to #9130 , without the json serialization and without a separate cache.
It is not yet clear which approach proves to be the solution we go with, this PR is only for test runs for now.

## Changes
* Patch botocore config object to be hashable and comparable for equality
* Add tests for caching behavior
* Limit client cache to 256 clients


thrau (approved):
very nice! good catch also with the lru_cache :ok_hand:

Issue #4350, opened 2021-07-23
When creating DynamoDB to Kinesis streams with CF, stream name is ignored and replaced by a generated name


jhberthemet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using CF like this:
```
{
    "Resources": {
        "TableTest": {
            "Type": "AWS::DynamoDB::Table",
            "Properties": {
                "TableName": "TestTable",
                "KeySchema": [
                    {
                        "AttributeName": "id",
                        "KeyType": "HASH"
                    }
                ],
                "AttributeDefinitions": [
                    {
                        "AttributeName": "id",
                        "AttributeType": "S"
                    }
                ],
                "ProvisionedThroughput": {
                  "ReadCapacityUnits" : 1,
                  "WriteCapacityUnits" : 1
                },
                "StreamSpecification": {
                    "StreamViewType": "NEW_AND_OLD_IMAGES"
                },
                "KinesisStreamSpecification": {
                    "StreamArn": {
                        "Fn::GetAtt": [
                            "TestStream",
                            "Arn"
                        ]
                    }
                }
            }
        },
        "TestStream": {
            "Type": "AWS::Kinesis::Stream",
            "Properties": {
                "Name": "test-stream",
                "ShardCount": 1
            }
        }
    }
}
```

Two streams will be created, first one "test-stream" and another one like:
```
    "StreamARN": "arn:aws:kinesis:us-east-1:000000000000:stream/__ddb_stream_TestTable",
    "StreamName": "__ddb_stream_TestTable",
```

Dynamo events actually goes to __ddb_stream_TestTable instead of test-stream

### Expected Behavior

I would expect that only "test-stream" stream is created and that dynamo events goes it it.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

`awslocal cloudformation create-stack --stack-name myteststack --template-body file://mytemplate.json`

### Environment

```markdown
- OS: MacOS 
- LocalStack: latest
- Docker: 3.5.2
```

giograno:
[...]

PR #9138, opened 2023-09-13
Add AWS::DynamoDB::Table resource provider

pinzon:
## Motivation
While triaging a tricky issue(#4350) I took the opportunity to migrate the DynamoDB Table generic model into a resource provider.

## Changes
- New resource provider for `AWS::DynamoDB::Table`
- ~~The `docker-mount-run` command now also mounts the entrypoints~~ @dominikschubert: removing this from the PR since it's a bit of scope creep. feel free to open a new PR with those changes though
- New aws validated test to validate the dynamodb-kinesis integration 

## Testing
All the CFn deployments of a Table resource should still work


dominikschubert:
@pinzon short summary of the changes I made:

- Only use `StreamSpecification` for the create call if it is actually set in the properties (and only then add the enabled: True entry)
- Delete was not properly waiting for a potential asynchronous deletion. Fixed by using the context with proper IN_PROGRESS state.
- Adapted the resource provider for https://github.com/localstack/localstack/pull/9267

localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
```diff
@@ -0,0 +1,414 @@
...
+                "TableName",
+                "AttributeDefinitions",
+                "KeySchema",
+                "BillingMode",
+                "ProvisionedThroughput",
+                "LocalSecondaryIndexes",
+                "GlobalSecondaryIndexes",
+                "Tags",
+            ]
+            create_params = util.select_attributes(model, properties)
```
simonrw, line 219: non-blocking: why use this util function rather than building the dictionary yourself? It would be nice to break away from these utils at some point, since they have a weird API and are trying to be too general

tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -800,7 +800,9 @@ def test_dynamodb_stream_stream_view_type(self, aws_client):
         dynamodb.delete_table(TableName=table_name)
 
     @markers.aws.only_localstack
-    def test_dynamodb_with_kinesis_stream(self, aws_client, secondary_aws_client):
+    def test_dynamodb_with_kinesis_stream(
+        self, aws_client, secondary_aws_client, wait_for_stream_ready
```
simonrw, line 804: minor: `wait_for_stream_ready` is not used in this test. It should probably replace new lines 813-814

simonrw (approved):
Looks like some great changes. Just some minor comments, nothing blocking

localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
dominikschubert: It's probably still easier to keep the actual code more consistent using a few of these utils. In general they are only a subset now of the old ones an without their weird signatures.

tests/aws/services/dynamodb/test_dynamodb.py
dominikschubert: @pinzon can you give some context here? Do we know why this is marked `only_localstack`?

pinzon: I guess it's related with the secondary account necessary to test the cross-account integration.

PR #9139, opened 2023-09-13
refactor integration test startup into pytest plugin

thrau:
## Motivation

Through the separation of `tests/integration` and `tests/aws`, we duplicated the `conftest.py` file that starts localstack in the pytest interpreter process. This is a problem when you run both `tests/integration` and `tests/aws` in one pytest run, because it means that we attempt to start localstack twice, or start it a second time with polluted global state. specifically this manifested when @dfangl was testing the DNS component.

One way to fix this would be to run tests/aws and tests/integration in separate steps. This will surely be the longer-term solution. For now, we de-duplicated the code to start localstack into a pytest plugin. That way the localstack session is shared between test modules correctly. 

It was important to make sure the plugin has no global side effects so it can be safely imported in other tests that don't want to start localstack in memory. This makes the whole plugin look a bit strange. We initially tried a session-scoped fixture, but that had all sorts of problems associated with it in terms of startup/teardown of other session fixtures that make assumptions about localstack being available. So we rely solely on pytest hooks.

Co-authored this with @dfangl in a pairing session

## Changes

* Move all relevant code into `localstack/testing/pytest/in_memory_localstack.py`
* Greatly simplified the original code
* Move async initialization of service dependencies (like downloading opensearch) into a tests/aws/confitest.py hook


tests/aws/conftest.py
```diff
@@ -78,110 +47,5 @@ def pytest_runtestloop(session):
...
-    if localstack_started.is_set():
-        # called by all tests after the startup has completed and the initial tests are unblocked
-        yield
-        return
-
-    startup_monitor_event.set()
-    localstack_started.wait()
-    yield
-    return
+        fn()
```
dfangl, line 51: Maybe we should not have stripped the try: except: here, but if it cannot install the test dependencies, it might be good to fail immediately.

dfangl (approved):
I think this greatly improves the clarity of the LocalStack test startup!
And LocalStack only starts once :tada: 

Looking forward to upgrading pytest and using Stash instead of the options though :)

PR #9140, opened 2023-09-13
add resourcegroups validated tests

bentsku:
## Motivation
We did not have validated tests for the [ResourceGroups service](https://docs.aws.amazon.com/ARG/latest/userguide/resource-groups.html).

We would like to have more validated tests, especially covering the core operations and functionality (being able to query certain resources which have certain tags, or resources from a CloudFormation stack).

Resources:
- Support resources types: https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html#services-sqs
- Type of groups: https://docs.aws.amazon.com/ARG/latest/userguide/gettingstarted-query.html
- Integrations with other services: https://docs.aws.amazon.com/ARG/latest/userguide/integrated-services-list.html

## Changes
Added AWS validated tests for `ListGroupResources` for tag based queries and cloudformation based queries. One test also covering `SearchResources`, which offers the same functionality without created a group.

### DIFF
--- a/tests/aws/services/resource_groups/test_resource_groups.py
+++ b/tests/aws/services/resource_groups/test_resource_groups.py
@@ -124,7 +124,9 @@ class TestResourceGroups:
         assert 0 == len(response["Groups"])
 
     @markers.aws.validated
-    @pytest.mark.xfail(reason="Not implemented in moto (ListGroupResources)")
+    @pytest.mark.skipif(
+        condition=not is_aws_cloud(), reason="Not implemented in moto (ListGroupResources)"
+    )
     def test_resource_groups_tag_query(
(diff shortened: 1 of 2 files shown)


dominikschubert (approved):
LGTM!

Pity that we don't yet support the actually useful methods in that service :disappointed:

bentsku:
> Pity that we don't yet support the actually useful methods in that service 😞

Yeah, agree. I do think it will require substantial changes in our architecture to support though. For the tag based queries, we would need some kind of global tag database (could even use sqlite or something?) for every single resource supported by `resourcegroups`. A start would be to use the same instance of the `TaggingService` for all services instead of one per service, and also the reconciliation between moto resources (which have some kind of tagger service, but it might be per service as well...) and ours might be tricky. But I would really like something like that!

For CloudFormation queries, it might be a bit more straightforward?

PR #9142, opened 2023-09-13
Pin jsonpath_ng to 1.5.3 to unblock CI

dominikschubert:
## Motivation

CI started experiencing a failing unit test for snapshots with the recent `1.6.0` update of `jsonpath_ng`
## Changes

- pins `jsonpath_ng` to version `1.5.3`


thrau (approved):
thanks for the quick fix! let's make it compatible soon-ish!

Issue #9112, opened 2023-09-11
bug: Special Character, particularly '%' makes S3 gives error


Chitranshagarwal2000:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When trying to put object in S3 with special characters (tested with %) makes S3 give `NoSuchKey` error.
Tried multiple times with % sign and got the error, but after changing the % to _ everything works

### Expected Behavior

Special characters should not create a problem 

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    I start localstack via docker-compose.yml file, below is the snippet for localstack
    
`version: '3.8'
services:
  localstack:
    image: localstack/localstack
    ports:
      - "4566:4566"
      - "8080:8080"
    environment:
      - SERVICES=s3,sqs,secretsmanager,ssm
      - DOCKER_HOST=unix:///var/run/docker.sock
      - AWS_ACCESS_KEY_ID=test
      - AWS_SECRET_ACCESS_KEY=test
      - LS_LOG=trace
    volumes:
      - "./tmp/localstack:/tmp/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"`
      
I am using Spring Boot version 3.1.1 and localstack version 2.2.0. I use software.amazon.awssdk.services.s3.S3Client to put requests in the bucket

`@Bean
    public S3Client s3Client() {
        return S3Client.builder()
                       .region(Region.of(region))
                       .forcePathStyle(true)
                       .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create("test", "test")))
                       .endpointOverride(URI.create("http://localhost:4566"))
                       .build();}`
    
    This is my S3Client bean which I use to put Object

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    `PutObjectRequest request = PutObjectRequest.builder().
                    bucket("my-bucket").key("A12345%John%Smith%%01%XYZ-ABC.pdf").build();
            s3Client.putObject(request, RequestBody.fromBytes(stream));`
            
Here stream is a byte array
            
From the logs I can see `localstack.request.aws     : AWS s3.PutObject => 404 (NoSuchKey)`
            
Also if I try to list the files in a folder I get the below error
            
`exception during call chain: Unable to parse response (not well-formed (invalid token): line 1, column 433), invalid XML received. Further retries may succeed:
[...]

PR #9143, opened 2023-09-13
fix S3 handling of special character and trailing slash

bentsku:
## Motivation
This is a follow up from the Werkzeug upgrade, and the moto bump adapting to those changes, following the reported issue #9112

Moto now has a specific handling of special character to maintain compatibility between mocked botocore requests and werkzeuga requests. We've had issues when accessing directly moto's backend to get the key, with the proper key from the request.

I first suspected moto and tried working on a fix there and wrote some tests, but those were passing there, but not in LocalStack.

I've tracked it down to passing the request with `call_moto`, we would use the `request.url`, which is the decoded URL, which moto would then use to create its keys internally. 

Some examples of the issues:
#### Passing decoded request URL to moto:
Raw URL: `http://s3.localhost.localstack.cloud:4566/test-bucket-0ff73618/test%2540key/`
request.url passed to moto: `http://s3.localhost.localstack.cloud:4566/test-bucket-0ff73618/test@key/`
Key name in moto: `test@key/` when it should be `test%40key/`

#### Trailing slashes (once the issue above was done)
Raw path: `/test-bucket-18ba44f6/test%2540key/`
Parsed key in the parser: `test%40key` (missing the trailing slash, but was properly parsed in moto)
base_url: `http://s3.localhost.localstack.cloud:4566/test-bucket-18ba44f6/test@key/`
The check for `base_url.endswith(key)` does not work here, so the trailing slash isn't added.

## Changes

Moto needs the raw URL, so we now use `get_full_raw_path` to get the original request URL, so moto can properly use its logic to decode the URL and create the proper key name out of it. 

I've also thought of another scenario where this could be problematic, with special characters and handling slashes, which was already an issue before (this has been changed often, every time werkzeug changes its handling of URL), but sadly there's isn't a catch-all way to check if the request has a trailing slash that needs to be kept. 

Took the opportunity to fix the key returned by the multiple `List*` in the new provider as well, with some AWS validated tests, and removed some TODOs.

### DIFF
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -70,6 +70,7 @@ from abc import ABC
 from email.utils import parsedate_to_datetime
 from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 from typing.io import IO
+from urllib.parse import unquote
 from xml.etree import ElementTree as ETree
 
 import cbor2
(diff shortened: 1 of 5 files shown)


alexrashed (approved):
Awesome, great catch! 🚀

PR #9149, opened 2023-09-14
Fix scenario tests failing with non-default region

viren-nadkarni:
## Summary

This PR fixes the following scenario tests that fail in non-default account/region:

- `tests.aws.scenario.bookstore.test_bookstore.TestBookstoreApplication.test_setup`
- `tests.aws.scenario.bookstore.test_bookstore.TestBookstoreApplication.test_lambda_dynamodb`
- `tests.aws.scenario.bookstore.test_bookstore.TestBookstoreApplication.test_search_books`

## Changes

Earlier, the credentials and region were hardcoded within the Lambda function code. Now, a proper IAM role is passed as a [Lambda execution role](https://docs.aws.amazon.com/lambda/latest/dg/lambda-intro-execution-role.html). The function derives the proper credentials from this role and uses them for internal Boto instantiation.

## Tests

See `ci/circleci: acceptance-tests` job for 1b03412676039e77caad81adc839445053840e0a :green_circle: - this CI run used non-default test credentials, this demonstrates that the tests pass in such a situation.

The affected tests have also been re-validated against AWS.


steffyP (approved):
Thanks for jumping on this by improving and fixing the code 🚀 🙏

PR #9153, opened 2023-09-14
Set aws marker for all iam service tests

dfangl:
## Motivation
Fixes according to #9147.

## Changes
* Fix easy fixable tests to run against AWS
* Mark tests correctly
* Mark `test_simulate_principle_policy` as needs fixing, as it is wrong
* Allow create_role and create_user fixture to skip users which have already been deleted.


dominikschubert (approved):
LGTM! Thanks for going through them! :rocket:

PR #9157, opened 2023-09-14
Support for response overrides 

calvernaz:
## Motivation

This PR tackles a bunch of discrepancies detected in this [issue](https://github.com/klymenkoserhii/localstack-wisetack-samples/tree/main).

**1. Request Mapping Template Discrepancies.**

- [x] $util.escapeJavaScript($path.get($paramName)) - LocalStack version returns string surrounded by double quotes, while AWS version - without double quotes. ✅ 
- [x] $input.params().get('header') - LocalStack version returns data in different format - generate array of (key, value) tuples, while AWS version returns dictionary. ✅ 

**2. Response Mapping Template Issue - response headers and status codes override doesn't work.**

- [x]  it looks like **$util.parseJson()** mapping template function not implemented in LocalStack. ✅ 
- [x] It looks like mapping template $context object does not contain responseOverride property which is used to override an API's response parameters and status codes. ✅ 
- [x] A status code can’t be passed directly from the Lambda function in a non-proxy custom integration. AWS recommends to use mapping templates or regular expressions to map the status codes ✅ 

## Changes

- Add responseOverrides to the context to enable rendering of VTL templates
- Also tweaks the regular expression to detect dictionary access style on set assignment.
- Introduces `parseJson` and AttributeDict to enable access using dot notation
- Finally, improves openapi import function to enable multiple method responses and integration responses.


localstack/services/apigateway/helpers.py
```diff
@@ -1172,26 +1172,25 @@ def add_path_methods(rel_path: str, parts: List[str], parent_id=""):
...
-                    "default", {}
-                )
-                integration_response_templates = default_method_integration_response.get(
-                    "responseTemplates"
-                )
-                integration_response_parameters = default_method_integration_response.get(
-                    "responseParameters"
-                )
+                for pattern, integration_responses in method_integration_responses.items():
+                    method_integration_response = method_integration_responses.get(pattern, {})
```
whummer, line 1176: nit: no need to get the value from the `integration_responses` dict here, as we already extracted it via the iterator in the line above (`integration_responses` should be equivalent to `method_integration_response`)

localstack/services/apigateway/templates.py
```diff
@@ -104,6 +115,9 @@ def escapeJavaScript(self, obj: Any) -> str:
             return str(obj).lower()
         return str(obj)
 
+    def parseJson(self, s):
+        return AttributeDict(json.loads(s))
```
whummer, line 119: I guess the result of `json.loads` could also be a `list` or other type. Let's better check if the result is a `dict`, and only then wrap it in an `AttributeDict`.. 👍

localstack/utils/aws/templating.py
```diff
@@ -135,9 +135,7 @@ class ExtAssignment(airspeed.operators.Assignment):
     Extends the airspeed Assignment class to support names with dashes, e.g., "X-Amz-Target"
     """
 
-    START = re.compile(
-        r"\s*\(\s*\$([a-z_][a-z0-9_]*(?:\.[a-z_][a-z0-9_-]*)*)\s*=\s*(.*)$", re.S + re.I
-    )
+    START = re.compile(r"\s*\(\s*\$(\w*(?:\.[\w-]+|\[\"\$\w+\"\]*)*)\s*=\s*(.*)$", re.S + re.I)
```
whummer, line 138: Would be great to get these patches added to our airspeed fork directly at some point.

Question - seems like the semantics of the first part of the regex has slightly changed, should we have `\w+` instead of `\w*`?
```suggestion
    START = re.compile(r"\s*\(\s*\$(\w+(?:\.[\w-]+|\[\"\$\w+\"\]*)*)\s*=\s*(.*)$", re.S + re.I)
```

whummer (commented):
Great set of changes @calvernaz , nice to see the additional AWS validated test! 🚀 I've added a few minor questions and suggestions. Other than those, the changes LGTM!

tests/aws/services/lambda_/functions/lambda_error_responses.py
```diff
@@ -0,0 +1,7 @@
+def handler(event, context):
+    body = event.get("body")
+    status_code = body.get("httpStatus")
+    if int(status_code) >= 400:
+        return {"statusCode": 200, "body": "customerror"}
```
joe4dev, line 5: From a Lambda perspective, these are not really "errors" but rather different outputs.

FYI: We added dedicated error testing in `tests.aws.services.lambda_.test_lambda.TestLambdaErrors` with scenarios covering both handler errors and runtime startup errors.

bentsku (commented):
I have nothing more to add than Waldemar, thanks a lot for adding all these fixes! It's always great to see! 🚀

localstack/services/apigateway/templates.py
calvernaz: Yep, that's right! 

For the following string:

`{"errorMessage":"{\"key1\":\"var1\",\"key2\":{\"arr\":[1,2,3]}}"}`

and the VTL template
```
#set ($errorMessageObj = $util.parseJson($input.path('$.errorMessage')))
{
   "errorMessageObjKey2ArrVal" : $errorMessageObj.key2.arr[0]
}
```

should yield,

```
{
   "errorMessageObjKey2ArrVal" : 1
}
```

localstack/utils/aws/templating.py
calvernaz: Yep, the goal is to migrate from here to airspeed, it makes sense.

For correctness, we need after a `$` one or more characters as a variables, hence the `+`. But I think this is harmless so will keep it as before.

tests/aws/services/lambda_/functions/lambda_error_responses.py
calvernaz: Yep, the result of the call as in the HTTP status code is always 200. This example is just an example of mapping responses after the lambda call to "error" HTTP codes. 

Your comment is because of the lambda module name? Nevertheless this is not a lambda concern but API interpretation of a lambda response for expected errors, for example the input doesn't match an expectation therefore the lambda response contains a signal to be mapped by the gateway as a error (400 status code).

localstack/services/apigateway/helpers.py
calvernaz: Nice catch 😺 !!

tests/aws/services/lambda_/functions/lambda_error_responses.py
joe4dev: Agree, that makes sense. (just a minor naming nit but feel free to ignore)

Testing actual Lambda errors is another scenario (maybe worth covering in another test).

localstack/services/apigateway/templates.py
calvernaz: I reworked this a bit to apply to all keys.

whummer (approved):
Thanks for addressing the comments. LGTM! 🚀

Issue #9145, opened 2023-09-14
bug: Api-Gateway Get-Export Fails if given an api with resource with no methods defined


Fulgar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If I create a REST API in API GW with any resources that do not contain any methods, and try to export the JSON schema, it fails due to:
```python
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/localstack/services/apigateway/helpers.py", line 1468, in _add_paths
    for method, method_config in item.get("resourceMethods").items():
AttributeError: 'NoneType' object has no attribute 'items'
```

Resource Structure:
/ - no methods defined
/{somethingId} - GET method defined

### Expected Behavior

I should be able to export an API that contains resources that do not contain any methods, just like in AWS.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

I start localstack:
`localstack start`

I create some lambda that prints hello world:

lambda.js
```javascript
'use strict'

const apiHandler = (payload, context, callback) => {
    callback(null, {
        statusCode: 200,
        body: JSON.stringify({
            message: 'Hello from Lambda'
        }),
    }); 
}
    
module.exports = {
    apiHandler,
}

```

```bash
zip function.zip lambda.js
awslocal lambda create-function \
  --function-name apigw-lambda \
  --runtime nodejs16.x \
  --handler lambda.apiHandler \
  --memory-size 128 \
  --zip-file fileb://function.zip \
  --role arn:aws:iam::111111111111:role/apigw
```

I create the REST api:
```bash
awslocal apigateway create-rest-api --name 'API Gateway Lambda integration' --tags '{"_custom_id_": "test-api"}'
```

I get the root resource ID:
```bash
awslocal apigateway get-resources --rest-api-id test-api
```

Create a child resource:
```bash
awslocal apigateway create-resource \
  --rest-api-id test-api \
  --parent-id 4hvslaey04 \
  --path-part "{somethingId}"
```

Create method (ONLY FOR THE CHILD NODE):
```bash
awslocal apigateway put-method \
  --rest-api-id test-api \
  --resource-id oxykf16mty \
  --http-method GET \
  --request-parameters "method.request.path.somethingId=true" \
  --authorization-type "NONE"
```

Create Integration (ONLY FOR THE CHILD NODE):
```bash
awslocal apigateway put-integration \
  --rest-api-id test-api \
  --resource-id oxykf16mty \
  --http-method GET \
[...]

PR #9159, opened 2023-09-15
fix apigw import root and extend OpenAPIExporter

bentsku:
## Motivation
What started as a quick fix for #9145 got a bit out of hand with the introduction of snapshot test.

This PR fixes the bug mentioned, and extend the behaviour of the OpenAPI Exporter. It introduces more parity and separate clearly both types of exporter (I'm not exactly sure of the split yet and if I like it, but this could be a good intro for the Importer). 

Also introduces one apigateway extension, but this will need to be extended to support almost all of them, same way as the importer. 

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-export-api.html

## Changes
Fix the bug above by adding safety around getting attributes.

Refactored the importer for better separation of the two types of Exporter, those are pretty close but still quite different in their format, and it's hard to merge some behaviour. 

Added 2 snapshot validated tests around exporting, which helped a lot!

### DIFF
--- a/localstack/services/apigateway/exporter.py
+++ b/localstack/services/apigateway/exporter.py
@@ -8,6 +8,8 @@ from localstack.aws.api.apigateway import ListOfModel
 from localstack.aws.connect import connect_to
 from localstack.utils.time import TIMESTAMP_FORMAT_TZ, timestamp
 
+from .helpers import OpenAPIExt
+
 # TODO:
 # - handle more extensions
 #   see the list in OpenAPIExt
@@ -39,7 +41,7 @@ class _BaseOpenApiExporter(abc.ABC):
                 self._resolve_refs(value, base_path)
 
     @staticmethod
-    def _get_integration(method_integration) -> dict:
+    def _get_integration(method_integration: dict) -> dict:
         fields = {
             "type",
             "passthroughBehavior",
@@ -56,12 +58,23 @@ class _BaseOpenApiExporter(abc.ABC):
         return integration
 
     @abc.abstractmethod
-    def export(self, api_id: str, stage: str, export_format: str, with_extension: bool) -> str:
+    def export(
+        self, api_id: str, stage: str, export_format: str, with_extension: bool
+    ) -> str | dict:
         ...
 
-    @staticmethod
     @abc.abstractmethod
-    def _add_paths(spec, resources, with_extension):
+    def _add_paths(self, spec: APISpec, resources: dict, with_extension: bool):
+        """
+        This method iterates over the different REST resources and its methods to add the APISpec paths using the
+        `apispec` module.
+        The path format is different between Swagger (OpenAPI 2.0) and OpenAPI 3.0
+        :param spec: an APISpec object representing the exported API Gateway REST API
+        :param resources: the API Gateway REST API resources (methods, methods integrations, responses...)
+        :param with_extension: flag to add the custom OpenAPI extension `apigateway`, allowing to properly import
+        integrations for example, or authorizers. (all the `x-amazon` fields contained in `OpenAPIExt`).
+        :return: None
+        """
         ...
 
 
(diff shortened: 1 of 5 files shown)


localstack/services/apigateway/exporter.py
```diff
@@ -0,0 +1,285 @@
...
+class _OpenApiSwaggerExporter(_BaseOpenApiExporter):
+    VERSION = "2.0"
+
+    def _add_paths(self, spec, resources, with_extension):
+        for item in resources.get("items"):
+            path = item.get("path")
+            for method, method_config in item.get("resourceMethods", {}).items():
+                method = method.lower()
+
+                # TODO: this is extension, integration is extensions
```
bentsku, line 77: disregard, this is a leftover 😅

localstack/services/apigateway/exporter.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def _resolve_refs(self, schema: dict, base_path: str):
+        if "$ref" in schema:
+            schema["$ref"] = f'{base_path}/{schema["$ref"].rsplit("/", maxsplit=1)[-1]}'
+        for value in schema.values():
+            if isinstance(value, dict):
+                self._resolve_refs(value, base_path)
+
+    @staticmethod
+    def _get_integration(method_integration) -> dict:
```
whummer, line 42:
```suggestion
    def _get_integration(method_integration: dict) -> dict:
```

localstack/services/apigateway/exporter.py
```diff
@@ -0,0 +1,285 @@
...
+            integration["responses"] = {"default": responses.get("200")}
+        return integration
+
+    @abc.abstractmethod
+    def export(self, api_id: str, stage: str, export_format: str, with_extension: bool) -> str:
+        ...
+
+    @staticmethod
+    @abc.abstractmethod
+    def _add_paths(spec, resources, with_extension):
```
whummer, line 64: This method definition seems slightly different than in the subclasses - should it be something like:
```suggestion
    @abc.abstractmethod
    def _add_paths(spec: APISpec, resources: dict, with_extension: bool):
```

localstack/services/apigateway/exporter.py
```diff
@@ -0,0 +1,285 @@
...
+                if operation_name := method_config.get("operationName"):
+                    method_operations["operationId"] = operation_name
+                if with_extension and method_integration:
+                    method_operations["x-amazon-apigateway-integration"] = self._get_integration(
+                        method_integration
+                    )
+
+                spec.path(path=path, operations={method: method_operations})
+
+    def export(self, api_id: str, stage: str, export_format: str, with_extension: bool) -> str:
```
whummer, line 237: nit: Seems like the return type would be `str | dict`, depending on which `export_format` is specified

localstack/services/apigateway/exporter.py
```diff
@@ -0,0 +1,285 @@
...
+        return integration
+
+    @abc.abstractmethod
+    def export(self, api_id: str, stage: str, export_format: str, with_extension: bool) -> str:
+        ...
+
+    @staticmethod
+    @abc.abstractmethod
+    def _add_paths(spec, resources, with_extension):
+        ...
```
whummer, line 65: Maybe we could add a short docstring with a high-level description of the functionality here, since there's quite a bit of logic going on in the two implementations. 👍

whummer (approved):
Great set of changes @bentsku ! 🚀  Makes perfect sense to split out the logic for Swagger/OAS into separate subclasses, as the logic for the two is sufficiently distinct. Great to see the additional AWS-validated tests 🥳 💪

localstack/services/apigateway/exporter.py
bentsku: Good catch! Implementation has changed and I forgot to update the signature! Thanks a lot

localstack/services/apigateway/exporter.py
bentsku: Yes, definitely. I guess we should move the logic of the JSON dumps inside the exporter instead of the provider, this could be done in a follow up. Updated the signature in the meantime, thanks a lot for the thorough review.

Issue #7198, opened 2022-11-17
feature request: AWS Resource Access Manager


korenyoni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

It would be great if localstack supported AWS Resource Access Manager. This would allow for testing multi-account AWS Transit Gateway setups, for example.

### Anything else?

Related: https://github.com/localstack/localstack/issues/7041

PR #9161, opened 2023-09-15
Introducing Resource Access Manager (RAM)

viren-nadkarni:
## Motivation

This PR introduces the Resource Access Manager (RAM) service. RAM is an alternate way to grant access to principals for AWS resources.

LocalStack allows unrestricted cross-account and cross-region access (unless IAM enforcement is turned on) so this service is redundant. The primary use-case is to let users run and test their IaC workflows with minimal changes.

cc: @cabeaulac 

## Implementation

This is a very basic provider configured to fall back on Moto.

Currently following operations will be supported:

- CreateResourceShare
- UpdateResourceShare
- DeleteResourceShare
- GetResourceShares
- EnableSharingWithAwsOrganization

## Tests

Includes a basic sanity test.

## Related

Closes: https://github.com/localstack/localstack/issues/7198

## To-do

- Update aws-cli local to support this endpoint
- Update docs

### DIFF
--- a/localstack/services/ram/provider.py
+++ b/localstack/services/ram/provider.py
@@ -1,10 +1,9 @@
 import logging
 
 from localstack.aws.api.ram import RamApi
-from localstack.services.plugins import ServiceLifecycleHook
 
 LOG = logging.getLogger(__name__)
 
 
-class RamProvider(RamApi, ServiceLifecycleHook):
+class RamProvider(RamApi):
     pass
--- a/localstack/services/ram/provider.py
+++ b/localstack/services/ram/provider.py
@@ -1,0 +1,10 @@
+import logging
+
+from localstack.aws.api.ram import RamApi
+from localstack.services.plugins import ServiceLifecycleHook
+
+LOG = logging.getLogger(__name__)
+
+
+class RamProvider(RamApi, ServiceLifecycleHook):
+    pass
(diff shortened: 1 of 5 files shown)


localstack/services/ram/provider.py
```diff
@@ -0,0 +1,10 @@
+import logging
+
+from localstack.aws.api.ram import RamApi
+from localstack.services.plugins import ServiceLifecycleHook
+
+LOG = logging.getLogger(__name__)
+
+
+class RamProvider(RamApi, ServiceLifecycleHook):
```
alexrashed, line 9: nit: This provider should not implement a `ServiceLifecycleHook` here, it's not hooking into the lifecycle anywhere.

localstack/services/ram/models.py
alexrashed, line 1: q: I don't think this file is necessary, why did you create it?

alexrashed (approved):
Nice starting point for a new service! 🚀 
I only found two things which should not be necessary, but nothing blocking the merge.

localstack/services/ram/models.py
viren-nadkarni: It's the standard provider module structure, but it can also be created when needed. Removed in 0d2dcd536

localstack/services/ram/provider.py
viren-nadkarni: Removed in 0d2dcd536 :+1:

PR #9162, opened 2023-09-15
remove forward annotations from bootstrap.py

thrau:
## Motivation

There was an issue in the latest dev version of the CLI, where I would get these error messages when loading plugins from ext:

```
ERROR:plugin.manager:error importing entrypoint EntryPoint(name='activate_pro_key_on_host', value='localstack_ext.plugins:activate_pro_key_on_host', group='localstack.hooks.prepare_host'): cannot import name 'Container' from 'localstack.utils.bootstrap' (/home/thomas/.pyenv/versions/3.10.9/lib/python3.10/site-packages/localstack/utils/bootstrap.py)
```

I narrowed it down to the issue of forward annotations and `from __future__ import annotations`. When I added the import to `localstack_ext/plugins.py` it wold work correctly. It also worked if I removed the line from `bootstrap.py`. I didn't get to the bottom of whether this is something unique to the plugin loading, or something else, but it seemed best to remove the need for forward annotations completely for now until we better understand the issue.

## Changes

* Remove `from __future__ import annotations` from bootstrap.py and all instances of forward annotations

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import copy
 import functools
 import logging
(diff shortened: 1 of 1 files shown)


simonrw (approved):
Sad to see `annotations` future import go since it makes the code nicer to read, but fair enough.

There are more typing issues that were accepted with the `__future__` import.

thrau:
OK then i'll do it the other way around, to import `from __future__ import annotations` in ext!

PR #9163, opened 2023-09-15
fix(param_validation): regex didn't include nested dot notation parameters

eXamadeus:


eXamadeus:
This is replaced by #9165

Issue #9164, opened 2023-09-15
bug: boto parameters are not fixed when nested


eXamadeus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When trying to deploy through localstack using aws-cdk-local, I ran into an issue where the parameter validation logic was failing. The error I was seeing was:

```
 ❌ Deployment failed: Error: The stack named TestStack failed to deploy: CREATE_FAILED (Parameter validation failed:
Invalid type for parameter LaunchTemplate.Version, value: 1, type: <class 'int'>, valid types: <class 'str'>): Parameter validation failed:
Invalid type for parameter LaunchTemplate.Version, value: 1, type: <class 'int'>, valid types: <class 'str'>
    at FullCloudFormationDeployment.monitorDeployment (<projDir>/node_modules/aws-cdk/lib/index.js:443:10232)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.deployStack2 [as deployStack] (<projDir>/node_modules/aws-cdk/lib/index.js:446:153546)
    at async <projDir>/node_modules/aws-cdk/lib/index.js:446:136809
```

Basically, the parameters being sent are of the wrong type. This isn't an issue when using aws-cdk directly, as the actual deployment to AWS works fine. It is only happening in localstack.

By turning on `DEBUG` mode in localstack, I was able to see this issue in the logs:

```
localstack_main  | 2023-09-15T14:24:32.925 DEBUG --- [functhread86] l.s.c.resource_provider    : Converting parameters to allowed types
localstack_main  | 2023-09-15T14:24:32.925 DEBUG --- [functhread86] l.s.c.resource_provider    : Original parameters:  {'LaunchTemplate': {'LaunchTemplateId': 'lt-68a5be63ac0438a79', 'Version': 1}, 'MaxSize': 2, 'MinSize': 1, 'Tags': [{'Key': 'Name', 'PropagateAtLaunch': True, 'Value': 'TestStack/Cluster/ASGCapacity'}], 'VPCZoneIdentifier': 'subnet-0fb5a52e,subnet-18b570e2', 'AutoScalingGroupName': 'ClusterASGCapacityASGB1FEE0DD-c703f313', 'DefaultCooldown': 300}
[...]

PR #9165, opened 2023-09-15
fix(boto params): regex doesn't account for nested dot notation parameters

eXamadeus:
## Motivation

See #9164

## Changes

I updated the regex to match `.`-delineated nested parameter values, like `LaunchTemplate.Version`. Interestingly, the logic of the rest of the function worked as expected. I wonder if the error message format shifted from it's original?

## Testing

I wrote a unit test, which should cover the functionality. If you want to replicate it locally, use `awk-cdk-local` and the following CDK snippet (it's probably more than necessary, but it's how I first noticed the issue):

```ts
#!/usr/bin/env node
import 'source-map-support/register'
import * as cdk from 'aws-cdk-lib'
import { Stack, StackProps } from 'aws-cdk-lib'
import { Construct } from 'constructs'
import { InstanceClass, InstanceSize, InstanceType, Vpc } from 'aws-cdk-lib/aws-ec2'
import { Cluster } from 'aws-cdk-lib/aws-ecs'

const app = new cdk.App()
class TestStack extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props)

    // Create a simple VPC & security group
    const vpc = new Vpc(this, 'Vpc', { maxAzs: 2 })

    // Create an ECS Cluster
    const cluster = new Cluster(this, 'Cluster', { vpc })

    // Attach the capacity provider to the cluster
    cluster.addCapacity('ASGCapacity', {
      instanceType: InstanceType.of(InstanceClass.T2, InstanceSize.MICRO),
      minCapacity: 1,
      maxCapacity: 2,
    })
  }
}

new TestStack(app, 'TestStack', {})
```

I wrote a unit test directly for the function, as it felt like overkill to build out a deployment that generated a nested parameter. I also felt like a test based around deployment could one day change the nesting of parameters, which would make that type of test brittle, so I just wrote a unit test directly for the function.


localstack/services/cloudformation/provider_utils.py
```diff
@@ -53,7 +53,7 @@ def remove_none_values(obj):
 
 # FIXME: this shouldn't be necessary in the future
 param_validation = re.compile(
-    r"Invalid type for parameter (?P<param>\w+), value: (?P<value>\w+), type: <class '(?P<wrong_class>\w+)'>, valid types: <class '(?P<valid_class>\w+)'>"
+    r"Invalid type for parameter (?P<param>[\w.]+), value: (?P<value>\w+), type: <class '(?P<wrong_class>\w+)'>, valid types: <class '(?P<valid_class>\w+)'>"
```
eXamadeus, line 56: I'm not entirely sure what this separate `provider_utils.py` is for, but I assume the logic should be copied over.

eXamadeus:
I can't seem to set a label, but this would be `semver: patch`

localstack/services/cloudformation/resource_provider.py
```diff
@@ -239,6 +239,7 @@ def invoke_function(
                 raise
 
             LOG.debug("Converting parameters to allowed types")
+            LOG.debug(f"Report: {report}")
```
eXamadeus, line 242: This is useful in detecting any future deviations of the report message.

dominikschubert, line 242:
```suggestion
            LOG.debug("Report: %s", report)
```
To avoid unnecessary string interpolation if debug isn't enabled.

localstack/services/cloudformation/provider_utils.py
dominikschubert: We'll probably be able to remove this from this specific file in the future, but in the meantime thanks for keeping it in sync!

Some context: The idea of the `provider_utils.py` file is actually to be a simplified utils package with no dependencies to the rest of the localstack code base (or any other external dependencies) so that we can package this together with a resource provider when uploading to AWS for testing.

tests/unit/services/cloudformation/test_deployment_utils.py
```diff
@@ -0,0 +1,27 @@
+from localstack.services.cloudformation.deployment_utils import fix_boto_parameters_based_on_report
+from localstack.testing.pytest import markers
+
+
+class TestFixBotoParametersBasedOnReport:
+    @markers.aws.only_localstack
```
dominikschubert, line 6:
```suggestion
```
these are only required for tests that can potentially be executed against AWS (anything in `tests/aws/`)

tests/unit/services/cloudformation/test_deployment_utils.py
```diff
@@ -0,0 +1,27 @@
...
+            "Invalid type for parameter LaunchTemplate.Version, "
+            "value: 1, type: <class 'int'>, valid types: <class 'str'>"
+        )
+
+        fixed_params = fix_boto_parameters_based_on_report(params, message)
+        value = fixed_params["LaunchTemplate"]["Version"]
+        assert value == "1"
+        assert type(value) == str
+
+    @markers.aws.only_localstack
```
dominikschubert, line 19:
```suggestion
```

dominikschubert (changes requested):
Great catch and thanks for your contribution! :tada: 

I've noted a few small nits. As soon as they're addressed there's nothing blocking a merge :+1:

localstack/services/cloudformation/resource_provider.py
eXamadeus: Ahh good call. I'll add that.

tests/unit/services/cloudformation/test_deployment_utils.py
eXamadeus: Ohh, I think I tried writing a test there and saw the "required message" so just mindlessly copied the markers over. Good to know. I'll remove them.

dominikschubert (approved):
LGTM, thanks for addressing the previous issues and also in general thanks for your contribution! :rocket:

Issue #9154, opened 2023-09-14
bug: error when re-subscribing an SQS queue to an SNS topic with RawMessageDelivery=true


dbouclier:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When re-subscribing an SQS queue to an SNS topic with the attribute "RawMessageDelivery=true", localstack throws an error.
It's occuring both on my AWS Java Client and with CLI awslocal
it's working well without this attribute

### Expected Behavior

Since the queue is subscribed with the attribute "RawMessageDelivery=true" to the topic already, it should be a no-op and simply return the same subscription arn.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
awslocal sns create-topic --name localstack-topic
awslocal sqs create-queue --queue-name localstack-queue

awslocal sns subscribe   --topic-arn arn:aws:sns:eu-central-1:000000000000:localstack-topic    --protocol sqs    --notification-endpoint arn:aws:sqs:eu-central-1:000000000000:localstack-queue    --return-subscription-arn  --attributes RawMessageDelivery=True

awslocal sns subscribe   --topic-arn arn:aws:sns:eu-central-1:000000000000:localstack-topic    --protocol sqs    --notification-endpoint arn:aws:sqs:eu-central-1:000000000000:localstack-queue    --return-subscription-arn  --attributes RawMessageDelivery=True
```

### Environment

```markdown
aws-cli/1.22.34 Python/3.10.12 Linux/6.2.0-32-generic botocore/1.31.47
```

### Anything else?

subscription created with "RawMessageDelivery=true"

![image](https://github.com/localstack/localstack/assets/98832464/faac719d-b5fe-4920-a756-28abc17fd0e7)

but if the subscription is created without attributes it's working well 

![image](https://github.com/localstack/localstack/assets/98832464/44ed7afd-a3b2-4a19-aff4-cba7b75ad52c)

maybe it's related to https://github.com/localstack/localstack/issues/9058

PR #9167, opened 2023-09-15
fix SNS subscribe idempotency

bentsku:
## Motivation
This is finally fixing SNS Subscribe idempotency and Subscription Attributes validation, reported again with #9154, which was introduced in #8976 and already fixed with #9055. 

This time, I think I've got the logic right, basically SNS will only validate present attributes in the `Subscribe` request. If an attribute is not provided, it won't be replaced by its default value but simply ignored.

This kinda shows that validation is good, but sometimes can be a bit tricky to write tests for, my bad 😬 

## Changes
This PR introduces the above logic and should fix all subsequent issues. We now check if the attributes is provided, then check if individual validated attributes are present in the request before validating equality with the subscription's.


thrau (approved):
LGTM! had the same issue in SQS with the queue attributes. strange behavior, but smells like old services ;-)

PR #9168, opened 2023-09-15
Fix handling of exceptions without arguments in legacy provider with local executor

jairov4:
## Motivation
I am fixing the issue in Step Functions when a Lambda function raise an exception without arguments, like: `raise ValueError()`.
Currently, the code fails in this case trying to get the error message to propagate and make controller upstream time out.

Address #9169

## Changes
Patch the behavior of retrieving the error message from the exception during errors in the lambda invoked.

## Testing

Just create a step function which lambda perform `raise ValueError()`
Now the SFN execution wont time out.

## TODO

What's left to do: Nothing

### DIFF
--- a/localstack/services/lambda_/lambda_executors.py
+++ b/localstack/services/lambda_/lambda_executors.py
@@ -1434,7 +1434,7 @@ class LambdaExecutorLocal(LambdaExecutor):
                     # need to translate to dict here, as custom errors from handlers cannot be pickled
                     execute_error = {
                         "errorType": e.__class__.__name__,
-                        "errorMessage": e.args[0],
+                        "errorMessage": e.args[0] if e.args else execute_result,
                         "stackTrace": traceback.format_tb(e.__traceback__),
                     }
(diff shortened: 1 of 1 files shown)


jairov4:
recheck

jairov4:

dominikschubert (approved):
Thanks for you contribution! :tada: 

Falling back to the stringified exception in this case makes sense. :+1: 

Normally we'd ask you also add tests here, but in this case the behavior is quite clear and won't actually be subject to any further change since we won't be actively working on this part of the code (the legacy (`v1`) lambda provider will be removed later this year).

Furthermore this exception mapping is something that, in the new provider, is handled by the Lambda runtime interface client.

dominikschubert:
Some additional context:

This should only occur when using the deprecated legacy provider for the Lambda service, i.e. when `LAMBDA_EXECUTOR="local"` and `PROVIDER_OVERRIDE_LAMBDA="v1"` (or "legacy") is set.

PR #9170, opened 2023-09-15
Fix lambda invoke exception with invalid payload

joe4dev:
Addresses https://github.com/localstack/docs/issues/785
Related to https://github.com/localstack/localstack-pro-samples/issues/224

## Motivation

The AWS CLI v1 and v2 differ in their Lambda payload encoding and using the wrong command or CLI version causes LocalStack to return a confusing `Internal error while executing lambda` rather than raising an `InvalidRequestContentException`.

## Changes

This PR fixes the exception type and improves AWS parity of the error message. The detailed parsing message still differs at AWS but that's cumbersome to fix (probably not worth it).

Add an AWS-validated test with two different encoding errors.

### DIFF
--- a/localstack/services/lambda_/invocation/lambda_service.py
+++ b/localstack/services/lambda_/invocation/lambda_service.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING, Optional
 from localstack import config
 from localstack.aws.api.lambda_ import (
     InvalidParameterValueException,
+    InvalidRequestContentException,
     InvocationType,
     LastUpdateStatus,
     ResourceConflictException,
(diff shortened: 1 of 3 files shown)


localstack/services/lambda_/invocation/lambda_service.py
```diff
@@ -291,6 +292,16 @@ def invoke(
         # empty payloads have to work as well
         if payload is None:
             payload = b"{}"
+        else:
+            # detect invalid payloads early before creating an execution environment
+            try:
+                to_str(payload)
```
joe4dev, line 298: I did not test whether AWS also detects this parsing error early but guess that makes sense. I guess double `to_str` conversion should not be too problematic given that there are payload limits.

Limits: https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html

```
6 MB each for request and response (synchronous)

20 MB for each [streamed response](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html) (Synchronous. The payload size for streamed responses can be increased from default values. Contact AWS Support to inquire further.)

256 KB (asynchronous)
```

joe4dev:
> The error message on AWS seems to also come directly from their json parser, so I agree, it's probably not worth to try and get this 100% right here.

Agree, I had the same thought. Cross-language parser messages are beyond reasonable parity.

(the double prefix is still funny `Could not parse request body into json: Could not parse payload into json:` 😆 )

Issue #9173, opened 2023-09-18
bug: SNS issue when publishing to a disabled Mobile Push Endpoint


bentsku:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Issue reported on the community Slack channel:

> I’m trying to write some integration tests using SNS and localstack. I stumbled upon https://github.com/getmoto/moto/issues/2333 which prevents us from testing some flows. But this is not what I want to ask about here. (Actually I don’t know if this is issue is the same one in localstack and not sure if localstack uses moto under the hood, but the symptoms are the same)
> The issue I want to ask about is this. I have an SNS endpoint that I disable by setting its “Enabled” attribute to “false”. I verify that the attribute is set correctly using the aws cli. Now, when I try to send messages to this endpoint I expect the call to AmazonSNS#publish to throw [EndpointDisabledException](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sns/classes/endpointdisabledexception.html) but it doesn’t. Instead, the message is, unexpectedly, published and I can see it using the aws cli.
> I looked around in the github project issues but found nothing related to this. I’m hopeful that I get some help here.

### Expected Behavior

SNS would raise the proper exception, and we could properly check idempotency of `CreatePlatformEndpoint` (there's an skipped test for it).

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```python
response = sns_client.create_platform_application(
    Name=f"test-{short_uid()}",
    Platform="GCM",
    Attributes={"PlatformCredential": "123"},
)
platform_arn = response["PlatformApplicationArn"]
response = sns_client.create_platform_endpoint(
    PlatformApplicationArn=platform_arn, Token="test1",
)
endpoint_arn = response["EndpointArn"]

sns_client.set_endpoint_attributes(EndpointArn=endpoint_arn, Attributes={"Enabled": "false"})

# this should fail
sns_client.publish(TargetArn=endpoint_arn, MessageStructure="json", Message=json.dumps({"GCM": '{ "notification": {"title": "Title of notification", "body": "It works" } }'}))
```

### Environment

```markdown
- OS:macOS Ventura 13.5
- LocalStack: `latest`
```

dogruis:
[...]

PR #9174, opened 2023-09-18
fix CreatePlatformEndpoint idempotency and Publish to mobile endpoint when disabled

bentsku:
## Motivation
These issues were reported in our community slack channel and converted into an issue: #9173 

We wouldn't check the endpoint status when publishing, and we had an incomplete fix for the `CreatePlatformEndpoint` idempotency, which led to an `skip` SNS test. 

## Changes
- Properly return the `EndpointArn` when calling `CreatePlatformEndpoint` multiple times with the same `CustomUserData` or no user data, and return the proper exception when the `CustomUserData` is different ("InvalidParameter") 
- Properly validate the endpoint status before publishing and raise an exception if the endpoint is disabled. 

As always, we cannot AWS validate those because we need real credentials to a register mobile app in GCM (Firebase) or Apple for example to set up a test case. This is always in the backlog, and would be awesome to have at some point, but we'll trust user reports and AWS documentation.

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -478,11 +478,11 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
             store = self.get_store(account_id=parsed_arn["account"], region_name=context.region)
             moto_sns_backend = self.get_moto_backend(parsed_arn["account"], context.region)
             if is_endpoint_publish:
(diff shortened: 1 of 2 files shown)


thrau (approved):
LGTM! will be interesting to see how we can get to a setup to run those tests against AWS at some point :-)

PR #9175, opened 2023-09-18
added support for overriding the log level of Kinesis Mock Server

adinballew:
...when it is not the same as the desired log level set in LS_LOG

## Motivation
Kinesis Mock Server Log-Level Logs when consistent with the other logs show when LS_LOG is set to debug is too verbose and so we would like a way to control those logs separately.

## Changes
Added additional log level control to Kinesis Mock Server.

## Testing
Tested Locally using the docker image built from source as the base image in our localstack project.

Description of how to test the changes
Enable a Localstack DynamoDB service that utilizes the KINESIS_MOCK_LOG_LEVEL env var and LS_LOG. Set them to be separate values (e.g LS_LOG=debug and KINESIS_MOCK_LOG_LEVEL=error).


adinballew:

thrau (approved):
Great addition! LGTM, tested locally and works great :+1:

PR #9176, opened 2023-09-19
wip: rework S3Integration to support path/action

bentsku:
## Motivation
We're currently only supporting 2 cases of the APIGW AWS S3 integration, `GetObject` for `path` and `action` URI.

Following this scenario should be successful against LocalStack: https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html
https://github.com/aws-samples/serverless-patterns/blob/main/apigw-s3-cdk/cdk/lib/apigw-s3-cdk-stack.ts

## Changes
Rework the whole S3 integration to support both `path` and `action` with different strategies to forward the request to S3. TBD

## TODO

What's left to do:

- [ ] AWS validate the existing test (seems weird with the proxy+ integration)
- [ ] Import the swagger file and run all operations tests on it (`path` based URI)
- [ ] try to find some sample/write tests for `action` based URI. 
- [x] add support for `requestOverride.path` from VTL template
- [ ] add test for `requestOverride.path`

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -635,15 +635,21 @@ class S3Integration(BackendIntegration):
 
         integration_parameters = self.request_params_resolver.resolve(context=invocation_context)
 
+        accepts = integration_parameters.get("headers", {}).get("Accept")
+        request_templates = integration.get("requestTemplates", {})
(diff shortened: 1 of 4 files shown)


bentsku:
Closing in favor of the NextGen implementation already available in Community

Issue #8931, opened 2023-08-17
bug: SNS with multiple Lambda Subscribers Crashes


CShelton11:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I've been having issue with local stack. When I have an sns topic that has a single lambda subscriber the docker container for lambda gets created appropriately and it runs. When I have an sns topic that has more than one lambda subscriber local stack creates the associated lambda containers but then crashes and doesn't run anything. I'm then forced to restart docker to get everything to work again. Is this a known issue or is there something that I need to configure to get this to work properly.  Interestingly if I warm up each of the lambdas individually and then I invoke the sns topic everything works appropriately.

### Expected Behavior

When the sns topic with multiple lamda subscribers receives a message The lambda containers will both be created and executed as expected.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

 I am starting this with a docker compose file.  The steps to reproduce this are one sns topic with two lambda functions that both subscribe to that SNS topic.  

### Environment

```markdown
- OS: windows
- LocalStack: latest
```

CShelton11:
Side note.  In my case, the issue happens when the python code makes an http call into API being debugged in visual studio.  Again, if I warm up the containers everything works fine.  If the sns message triggers the creation of the lambda containers, it locks up.

LordAndreasOtten:
im experiencing the same issue from macos with localstack 2.2.0 and nodejs 18.15 (docker) images subscribing to the sns topic.

LordAndreasOtten:
For now resolved by going back top v1.4.0

joe4dev:
Thank you for reaching out @CShelton11 and @LordAndreasOtten 

I drafted a test case trying to reproduce the scenario where one SNS topic triggers two cold Lambda functions in [test_multiple_lambda_subscribe_sns_topic](https://github.com/localstack/localstack/blob/test-multiple-lamabda-subscribe-sns-topic/tests/aws/services/sns/test_sns.py#L158C35-L158C35). I could not reproduce a scenario where LocalStack or any container crashes. However, I discovered that sending multiple messages in quick succession spawns more Lambda containers than necessary upon cold start.

What are the differences between your scenario compared to the test scenario outlined below?

PS: If you have any reproduction sample at hand, feel free to share it here or through our Pro support to facilitate debugging.

## Test scenario

[...]

PR #9178, opened 2023-09-19
Resolve LOCALSTACK_HOST to the LocalStack container

simonrw:
## Motivation

The user may specify a custom value of `LOCALSTACK_HOST` which is the domain name of the LocalStack container from the outside. This will not work however in created compute environments, or external docker containers since we only resolve `localhost.localstack.cloud` to the LocalStack container.

## Changes

- Bind LOCALSTACK_HOST to LocalStack IP address
- Update bootstrap tests to use `stream_container_logs` and `ContainerConfigurators`

### DIFF
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -247,6 +247,7 @@ class NonLoggingHandler(DNSHandler):
 
 NAME_PATTERNS_POINTING_TO_LOCALSTACK = [
     f".*{LOCALHOST_HOSTNAME}",
+    f".*{config.LOCALSTACK_HOST.host}",
 ]
 
 
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -247,7 +247,6 @@ class NonLoggingHandler(DNSHandler):
 
 NAME_PATTERNS_POINTING_TO_LOCALSTACK = [
     f".*{LOCALHOST_HOSTNAME}",
-    f".*{config.LOCALSTACK_HOST.host}",
 ]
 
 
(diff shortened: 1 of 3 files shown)


localstack/dns/server.py
```diff
@@ -247,6 +247,7 @@ def handle(self, *args, **kwargs):
 
 NAME_PATTERNS_POINTING_TO_LOCALSTACK = [
     f".*{LOCALHOST_HOSTNAME}",
+    f".*{config.LOCALSTACK_HOST.host}",
```
simonrw, line 250: Question: if `LOCALSTACK_HOST` is not overridden, its default value is `localhost.localstack.cloud` so we add the same name twice. I don't think this is a _problem_, but it is not very clean. Perhaps `Resolver.zones` should be a `dict[str, set[NameRecord]]`?

dfangl: It should not be a problem, but we might return multiple records for `localhost.localstack.cloud` then - not really clean.

Can we add a simple guard to avoid adding it if it is `LOCALHOST_HOSTNAME`?

simonrw: Sure but a set is more elegant! That sounds like a reasonable first step anyway

dfangl: My issue with the set is the following: We expect on delete a value error. If two different users of the DNS server add a similar entry, and both delete it afterwards, we get one value error - although there were two creates and two deletes.
These cases would need to be documented clearly, so users could ignore the value error on delete.

joe4dev:
@simonrw Does this resolve the following networking issue with SNS/Lambda https://github.com/localstack/localstack/issues/8931?
It looks like a networking issue with external Docker containers trying to access LocalStack 🤔

LordAndreasOtten:
What seems to be happening withing sns is -> subscribe request will be send correct (interal and remote) but it gets. a response url back with 127.0.0.1:4556 instead of host:4566 and because of this the external/other container cant subscribe :)

PR #9179, opened 2023-09-19
Add test for Serverlesspresso AWS sample application

dominikschubert:
## Motivation

As part of our scenario testing exploration, we also wanted to look into migrating bigger samples to our CDK-based setup and add multiple test scenarios on that infrastructure.

## Changes

- Adds a scenario test using a slightly modified version of [Serverlesspresso](https://workshop.serverlesscoffee.com/), a pretty sophisticated event-based sample application from AWS.
- Implements UPDATE support for stacks in infra provisioner for faster iteration cycles while writing the test scenarios 

## Testing

Note that the test currently only works against AWS.

## TODO

- [ ] extract re-usable utils
- [ ] only update stack when running on AWS
- [ ] add a simple test that at least validates successful deployment on LS
- [ ] add timeout tests
- [ ] add full e2e test with cognito/apigw


tests/aws/scenario/serverlesspresso/test_serverlesspresso.py
```diff
@@ -0,0 +1,430 @@
...
+    def _filter_log_events_for_execution_id(
+        self, aws_client, log_group_name: str, execution_arn: str, detail_type: str
+    ):
+        log_events = aws_client.logs.filter_log_events(
+            logGroupName=log_group_name,
+            filterPattern=f'{{ ($.detail-type = "{detail_type}") && (($.resources[1] = "{execution_arn}") || ($.resources[0] = "{execution_arn}"))}}',  # TODO: is it guaranteed that the second one is the execution_arn?
+        )["events"]
+        return log_events
+
+    @markers.aws.unknown
```
steffyP, line 143: as we do have a recorded snapshot, I guess it should be `aws.validated`?

tests/aws/scenario/test_ecs_scenario.py
```diff
@@ -50,3 +52,56 @@ def test_scenario_call_service(self, aws_client, infrastructure):
         # TODO: add a test here to call the deployed NGINX service
         # outputs = infrastructure.get_stack_outputs(stack_name="ClusterStack")
         requests.get("...")  # not working yet
+
+
+
+class TestEcsScenario2:
+    @markers.aws.unknown
+    def test_scenario_ecs(self, aws_client):
+        vpc = aws_client.ec2.create_vpc(CidrBlock="0.0.0.0/16")
```
steffyP, line 61: if this is only a leftover, it should probably be removed :) else we are definitely missing cleanups :)

steffyP (commented):
🤯 this sample is truly amazing 👏 
looking forward to having it in the test suite, and really curious about the coverage of this one too 😄 

Really like the structure, with having the constructs separated - it makes it way more readable 🚀 

As this is still a wip, I only added a few remarks so far. 
The snapshots are missing some more transformers 🙂

PR #9180, opened 2023-09-19
Fix flaky lambda concurrency test

joe4dev:
## Motivation

A flaky Lambda test occasionally blocks the pipeline. 

Thank you @bentsku for raising this https://localstack-cloud.slack.com/archives/C02U7DN1B4M/p1695076802171769

## Changes

* Increase reserved concurrency in test case
* Document race condition
* Add extra debug logging for concurrency exceptions

## Discussion

a) The safest fix is to increase ReservedConcurrentExecutions=3 but then we only occasionally test the border case.
b) We could slow down the test and use sleeps to have a high likelyhood of succeeding
c) We separate the mixed async/sync invoke into a separate test case (or after asserting the logs).

### DIFF
--- a/localstack/services/lambda_/invocation/counting_service.py
+++ b/localstack/services/lambda_/invocation/counting_service.py
@@ -157,6 +157,15 @@ class CountingService:
                         on_demand_tracker.increment(unqualified_function_arn)
                         lease_type = "on-demand"
                     else:
+                        extras = {
+                            "available_reserved_concurrency": available_reserved_concurrency,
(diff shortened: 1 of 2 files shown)


dominikschubert (approved):
> a) The safest fix is to increase ReservedConcurrentExecutions=3 but then we only occasionally test the border case.
> b) We could slow down the test and use sleeps to have a high likelyhood of succeeding
> c) We separate the mixed async/sync invoke into a separate test case (or after asserting the logs).

Let's do a) for now with c) in the long run :+1:

PR #9187, opened 2023-09-20
Ensure GATEWAY_LISTEN is taken from the configuration

simonrw:
## Motivation

GATEWAY_LISTEN is propagated to the container in its default form (127.0.0.1:4566) if not specified on the command line. This overrides any customisations that the Pro container configuration has performed (in particular, adding port 443 to GATEWAY_LISTEN).

## Changes

* Update the configurators to ensure that GATEWAY_LISTEN is taken from the configuration


thrau (approved):
oh that's a great catch! sorry for missing that one. LGTM!

PR #9189, opened 2023-09-20
CirecleCI: add report for acceptance and diff coverage

steffyP:
## Motivation
With the introduction of the acceptance test suite in #9045 we started collecting metrics as well. 
This PR aims to provide the first prototype to quickly glance through the collected metrics.

Currently we have two kind of metrics:
- code coverage
- api coverage

In the future those metrics are intended to help service owners to choose tests that should be part of the acceptance test suite. 

### Code Coverage
The code coverage is collected using the pytest plugin `coverage`, and results in a sqlite database added as artifact in the `report` job. 
Code coverage for the "normal" integration test is already reported to coveralls.

In order to get an overview about the acceptance coverage and the diff, we create two html reports, using [pycobertura](https://github.com/aconrad/pycobertura):
* **coverage-acceptance.html**
  * shows the acceptance code coverage
  * currently it only considers the files in `localstack/services` (which IMO are only relevant ones for the acceptance test suite, but we can discuss this)
  * it also excludes all `__init__.py` files to prevent blowing up the report with (probably) irrelevant information. 
* **coverage-diff.html**
  * same restrictions as above, e.g. only considering `localstack/services` and ignoring `__init__.py`
  * gives an overview about code lines that are missed by the acceptance test suite

### API Coverage
The metric collection for our test suite include data about each API call. 
To keep things simple, the first prototype considers only service-operation-responsecode combinations. 
We could further enhance this to include specific exceptions, or parameters used. 
The report **api-coverage/report_metric_coverage.html** is added as an artifact to the `report` job as well.

#### Details
The PR adds a small script that can be used to create simple HTML report outlining some basic metrics:
* It will show a very naive calculation of service coverage, by counting the amount of operations for each service that have been called for the full integration test suite (which is then treated as 100%); vs the operation called by the acceptance test suite (not considering the specific response codes)
Screenshot 2023-09-21 at 17 47 34

* Additionally, the report contains a list of all service-operation-responsecode combinations and a hint if those have been covered by the acceptance tests.
Screenshot 2023-09-21 at 17 47 48
 

## Changes
* added two steps to the `report` job to create the html reports

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -372,8 +372,8 @@ jobs:
       # store (uncombined) coverage from acceptance tests
       - run:
           name: fetch isolated acceptance coverage
-          command: |
           # might need to combine coverage files in future
+          command: |
             cp target/coverage/.coverage.acceptance.* .coverage.acceptance
       - store_artifacts:
           path: .coverage.acceptance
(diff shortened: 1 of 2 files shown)


thrau (approved):
looks very cool for a first iteration! can't wait to see how we can operationalize this and get a continuous view on how the coverage changes over time

.circleci/config.yml
```diff
@@ -400,13 +406,35 @@ jobs:
...
+          # but we still want cirecleci to continue with the tasks, so we return 0 always
           command: |
-            mkdir acceptance_parity_metrics
-            mv target/metric_reports/metric-report*acceptance* acceptance_parity_metrics/
+            source .venv/bin/activate
+            pip install pycobertura
+            coverage xml --data-file=.coverage -o all.coverage.report.xml --include="localstack/services/*/**" --omit="*/**/__init__.py"
+            coverage xml --data-file=.coverage.acceptance -o acceptance.coverage.report.xml --include="localstack/services/*/**"  --omit="*/**/__init__.py"
+            pycobertura show --format html -s localstack/ acceptance.coverage.report.xml -o coverage-acceptance.html
+            bash -c "pycobertura diff --format html -s1 localstack/ -s2 localstack/ all.coverage.report.xml acceptance.coverage.report.xml -o coverage-diff.html; if [[ \$? -eq 0 ]] ; then exit 1 ; else exit 0 ; fi"
```
alexrashed, line 419: q: Here you return 1 in case 0, and 1 in case of anything else, so the exit code 0 seems to be a problem. Why is that? If you basically just want to completely ignore errors on this line, you could just pipe it to true:
https://support.circleci.com/hc/en-us/articles/360057590591-How-to-ignore-a-failure-in-a-step-

alexrashed (approved):
Awesome! This is a really great change which will help us gaining confidence with the acceptance test suite! 🚀

.circleci/config.yml
steffyP: I added a comment about it in line 411 above the `command`.
Basically for the current set we most likely will see exit code 2. 
From the docs:
```
Upon exit, the diff command may return various exit codes:

    0: all changes are covered, no new uncovered statements have been introduced
    1: some exception occurred (likely due to inappropriate usage or a bug in pycobertura)
    2: the changes worsened the overall coverage
    3: the changes introduced uncovered statements but the overall coverage is still better than before
```
Basically I was looking for a way to ignore that expected return code, and this was the one suggested way of doing it :) 
It might make sense to return code 1 though `1: some exception occurred` 🤔 

So I guess we could change it to: `if [[ \$? -eq 1 ]] ; then exit 1 ; else exit 0 ; fi`? WDYT?

alexrashed: Oh, in this case I would say yes, we need to fail the step in case of exit code == 1, all other cases seem to be fine.

dominikschubert (approved):
> currently it only considers the files in localstack/services (which IMO are only relevant ones for the acceptance test suite, but we can discuss this)

I think the assumption is fine for what we want to test. There are certain utils that live outside of these though that still cover service logic. I'd probably prefer moving them to their proper service directory though :+1:

PR #9194, opened 2023-09-21
introduce request finalizers into handler chain

thrau:
## Motivation

In the persistence plugin (and perhaps elsewhere), we are using a pattern where we are acquiring locks in request handlers, and freeing them in response handlers, to lock certain things across a request invocation. This can lead to locks never being freed, when handlers (like the CORS enforcer) call `chain.terminate()`, which prevents response handlers from being called.

This PR introduces the concept of finalizers into the handler chain, which are *always* called, even if the chain is terminated. Obviously these should be used sparsely, and only run essential finalization code.

## Changes

* Introduce the `finalizer: List[Handler]` attribute
* Re-formatted a bunch of docs to match the column limit of our black config

### DIFF
--- a/localstack/aws/chain.py
+++ b/localstack/aws/chain.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, List, Optional
 
 from werkzeug import Response
 
+from ..utils.functions import call_safe
 from .api import RequestContext
 
 LOG = logging.getLogger(__name__)
@@ -76,6 +77,7 @@ class HandlerChain:
 
         self.stopped = False
         self.terminated = False
+        self.finalized = False
         self.error = None
         self.response = None
         self.context = None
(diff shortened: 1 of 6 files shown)


localstack/aws/chain.py
```diff
@@ -61,15 +66,18 @@ def __init__(
...
     ) -> None:
         super().__init__()
         self.request_handlers = request_handlers or list()
         self.response_handlers = response_handlers or list()
         self.exception_handlers = exception_handlers or list()
+        self.finalizers = finalizers or list()
 
         self.stopped = False
         self.terminated = False
+        self.finalized = False
```
bentsku, line 80: q: when is this set to `True`?  or how can you set it to `True`?

bentsku (approved):
LGTM! This is nice! Lucky for us for now, we didn't run into that kind of issues yet, but it's great to have this. Great catch! Only one small question regarding the boolean `finalized`, but otherwise all good!

localstack/aws/chain.py
thrau: A handler could set it. It's just a way to control things for tests oder other systems that may not want/need familzers

bentsku: Ah right yes thanks! The `Handler` has access to the `HandlerChain`. Thanks for answering 😄

PR #9200, opened 2023-09-21
Fix service health endpoint with custom GATEWAY_LISTEN

simonrw:
## Motivation

When using `GATEWAY_LISTEN` to override the hypercorn port(s) the internal service health checks are not configured to use this gateway port. This means that the internal health checks will fail.

## Changes

- Add tests that creates a service and checks the internal health check log line, configuring `GATEWAY_LISTEN` and `LOCALSTACK_HOST` (based on existing behaviour).
- Proxy listeners adhere to `GATEWAY_LISTEN`

### DIFF
--- a/localstack/services/infra.py
+++ b/localstack/services/infra.py
@@ -8,7 +8,6 @@ import traceback
 from typing import Dict, List, Union
 
 import boto3
-from localstack_client.config import get_service_port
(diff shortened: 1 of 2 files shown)


simonrw:
Closing in favour of #9202

PR #9202, opened 2023-09-21
disable health check for default service providers

thrau:
## Motivation

For legacy reasons, we used the `local_api_checker` as default health check. This made sense when we had backends running on different ports, but now almost all providers run in memory and not as backends, and those that require backends (like dynamodb), aren't considered anymore through the default `local_api_checker` health check (since we basically just call `:4566`).
Basically, doing health checks in this way makes no sense any more, and we can evaluate whether the concept as such makes sense and if so what we need to do to perform meaningful health checks.

I've seen errors related to these health checks in logs recently, and I think this is a good opportunity to get rid of them.

## Changes

* When using `Service.for_provider` (which is done in most cases), disable the health check

### DIFF
--- a/localstack/services/plugins.py
+++ b/localstack/services/plugins.py
@@ -174,6 +174,7 @@ class Service:
             name=provider.service,
             listener=AwsApiListener(provider.service, delegate=delegate),
             lifecycle_hook=service_lifecycle_hook,
+            check=None,
         )
         service._provider = provider
 


simonrw:
Does this PR invalidate https://github.com/localstack/localstack/pull/9200?

thrau:
> Does this PR invalidate #9200?

thanks for making me aware of that one. actually i suppose it would. your effort to fix this is much more admirable than what i'm doing :smile: but i think we should just get rid of them.

alexrashed (approved):
Awesome! This will help decreasing the coupling with localstack-python even further. I think we can remove this legacy code with the next major release. 🚀

PR #6724, opened 2022-08-22
Switch to new Lambda provider implementation

dfangl:
## Summary
This PR should remain minimal in terms of code changes, and is mostly used for keeping track of the PRs regarding the new lambda provider going directly into master and for tracking test failure rates against the new provider.

## PRs
The following PRs have been created to fully implement the new Lambda provider:
* #5306

## Tasks
- [ ] Initial core runtime implementation of the new provider
- [ ] Refactor package imports
- [ ] replace run_lambda calls by botocore invocations / common interface
- [ ] extend integration tests for full API coverage
- [ ] Version & Alias CRUD
- [ ] Full API CRUD
- [ ] Alias routing
- [ ] Lambda URL port

PR #9203, opened 2023-09-21
Add removal warning for legacy lambda provider

joe4dev:
## Motivation

* With 1.3, the new Lambda `v2` provider became available as opt-in.
* With 1.4 (ish), the legacy `v1` provider was deprecated and we encouraged users to opt-in to our new Lambda `v2` provider.
* With 2.0, the new lambda provider `v2` became the default implementation https://github.com/localstack/localstack/pull/6724.
* With 2.3, we add removal warnings in the old `v1` provider (this PR). We also ship the Lambda invocation loop rework https://github.com/localstack/localstack/pull/8970 (re-write of data plane).
* With 3.0 the old (v1) provider will be removed completely.

## Changes

Add a log warning when the legacy Lambda v1 provider is used. Only logged during initialization (`on_after_init`) for people using the legacy Lambda service.

## Discussion

Pro-active warnings at LocalStack startup:
* Do we want to prominently warn about deprecated environment values such as `PROVIDER_OVERRIDE_LAMBDA=v1` at the startup of LocalStack? We currently do not support this with `EnvVarDeprecation`. I drafted an `EnvVarValueDeprecation` in `a26c77375` but reverted it because it needs further updates in the logging message and telemetry if we even need this.
* I guess we would mainly use it for (important) provider deprecation warnings (e.g., S3, Lambda).
* It would help customers to see the warning right at the startup rather than potentially miss it somewhere in the logs when the provider gets loaded lazily.

Consistency: Trying to align the language for the deprecations and removals of s3, lambda, stepfunctions /cc
* @dominikschubert StepFunctions: https://github.com/localstack/localstack/pull/9198
* @bentsku S3: https://github.com/localstack/localstack/pull/9204

### DIFF
--- a/localstack/deprecations.py
+++ b/localstack/deprecations.py
@@ -303,6 +303,22 @@ def log_deprecation_warnings(deprecations: Optional[List[EnvVarDeprecation]] = N
     affected_deprecations = collect_affected_deprecations(deprecations)
     log_env_warning(affected_deprecations)
 
+    provider_override_lambda = os.environ.get("PROVIDER_OVERRIDE_LAMBDA")
+    if provider_override_lambda and provider_override_lambda in ["v1", "legacy"]:
+        env_var_value = f"PROVIDER_OVERRIDE_LAMBDA={provider_override_lambda}"
+        deprecation_version = "2.0.0"
+        deprecation_path = (
+            f"Remove {env_var_value} to use the new Lambda 'v2' provider (current default). "
+            "For more details, refer to our Lambda migration guide "
+            "https://docs.localstack.cloud/user-guide/aws/lambda/#migrating-to-lambda-v2"
+        )
+        LOG.warning(
+            "%s is deprecated (since %s) and will be removed in upcoming releases of LocalStack! %s",
+            env_var_value,
+            deprecation_version,
+            deprecation_path,
+        )
+
 
 def deprecated_endpoint(
     endpoint: Callable, previous_path: str, deprecation_version: str, new_path: str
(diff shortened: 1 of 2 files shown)


localstack/deprecations.py
```diff
@@ -303,6 +303,22 @@ def log_deprecation_warnings(deprecations: Optional[List[EnvVarDeprecation]] = N
     affected_deprecations = collect_affected_deprecations(deprecations)
     log_env_warning(affected_deprecations)
 
+    provider_override_lambda = os.environ.get("PROVIDER_OVERRIDE_LAMBDA")
+    if provider_override_lambda and provider_override_lambda in ["v1", "legacy"]:
```
joe4dev, line 307: Do we want/need this prominently at startup?

Screenshot 2023-09-21 at 20 43 08

joe4dev: Yes. The Lambda team discussed this makes sense for Lambda given the substantial difference. We want to highlight this and push users to adopt v2 or raise any blocking issues.

joe4dev: Notice that the language differs from the lazily loaded message. This message is aligned to the regular LS-startup deprecation warnings.

dfangl (approved):
LGTM! But why do we have the warning twice?

joe4dev:
> LGTM! But why do we have the warning twice?

The warning upon LS startup is much more present and appears right next to other important deprecation warnings.
The warning upon lazy provider loading often gets lost in the logs. I guess we could consider skipping this one then 🤔 At least we are extra clear about it and get it covered for every old provider initialization.

PR #8785, opened 2023-07-31
wip: new s3 native provider

bentsku:
WIP, this is a long living PR which is going to force the new native S3 provider so that we can run the whole CI suite with it. Description to be completed.

## Contained PRs:
- Nothing yet
### Incoming PRs:
- Coming soon

PR #9204, opened 2023-09-21
Add S3 deprecation warnings

joe4dev:
## Motivation

* With 1.?, the new S3 `v2` provider became available as opt-in.
* With 1.4, the legacy `v1` provider was deprecated and we encouraged users to opt-in to our new S3 `v2` provider.
* With 2.0, the current S3 provider `v2` became the default implementation https://github.com/localstack/localstack/pull/6724.
* With 2.3, we add removal warnings in the old `v1` provider (this PR) and the new native S3 provider `v3` becomes available https://github.com/localstack/localstack/pull/8785.
* With 3.0 the old (v1) provider will be removed completely and the new native S3 provider `v3` becomes the default.
* With 3.1/3.2? (TBD), the current S3 provider `v2` provider will be deprecated
* With 4.0 the current S3 provider `v2` will be removed completely.

## Changes

Add a log warning when the legacy S3 v1 provider is used. Only logged during initialization (`on_after_init`) for people using the legacy S3 service.

⚠️ This PR needs to be merged before the ext PR: https://github.com/localstack/localstack-ext/pull/2115

## Discussion

Consistency: Trying to align the language for the deprecations and removals of s3, lambda, stepfunctions /cc
* @dominikschubert StepFunctions: https://github.com/localstack/localstack/pull/9198
* @joe4dev Lambda: https://github.com/localstack/localstack/pull/9203

### DIFF
--- a/localstack/services/providers.py
+++ b/localstack/services/providers.py
@@ -242,7 +242,11 @@ def s3_legacy():
     from localstack.services.s3.legacy import s3_listener, s3_starter
 
     return Service(
-        "s3", listener=s3_listener.UPDATE_S3, start=s3_starter.start_s3, check=s3_starter.check_s3
+        "s3",
+        listener=s3_listener.UPDATE_S3,
+        start=s3_starter.start_s3,
+        check=s3_starter.check_s3,
+        lifecycle_hook=s3_starter.S3LifecycleHook(),
     )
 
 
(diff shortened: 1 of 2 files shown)


bentsku (approved):
Awesome, thank you so much for doing this! I think you resumed things perfectly, I don't know when we will start deprecating `v2`, maybe around 3.1/3.2? I don't think we will deprecate it straight away, it's exactly as you wrote it. Perfect!

Also, nice use of the `LifecycleHook`! I didn't even know you could use it with the old ProxyListeners.

PR #9206, opened 2023-09-21
make config.dirs.tmp be inside the container filesystem, add mounted_tmp

bentsku:
## Motivation
When working on S3 performance, it was shown that using mounted volumes on macOS induces a drop in filesystem operation performance from between 25 to 50%. This can be acceptable if using persistence, as there wouldn't be any other way to do it. However, when not using persistence, for many services, we are using the `config.dirs.data` folder (which points to tmp), which is inside the mounted volume, to create assets. This creates a drop in performance that might not be needed, if we can use the container own filesystem where it wouldn't be mounted.

## Changes
Use the `tempfile.gettempdir` method to create a temp dir which wouldn't clash inside the container for the `config.dirs.tmp`, which has the advantage of already having the logic for when persistence is disabled, data would direct to `config.dirs.tmp / "state"`. 

Create a new directory for the services that expected a temporary mounted directory, be it for tests or actual service implementation like RDS using it for sharing data between containers. 

### Solved PR discussions
Edit: seems like some Lambda tests are failing because they expect `tmp` to be a mounted dir. I guess we could change that? @joe4dev not sure if we need to change the directory used for the test, or if this is a real problem.
- tests/aws/services/lambda_/test_lambda.py::TestLayerHotReloading::test_layer_only_hot_reloading
- tests/aws/services/lambda_/test_lambda.py::TestLayerHotReloading::test_layer_and_function_hot_reloading

The issue is also in other services:
-  tests/aws/services/batch/test_batch.py::TestBatch::test_create_with_additional_config
- tests/aws/services/ec2/test_ec2.py::TestEC2::test_create_instance_with_ebs_create_fs
- tests/aws/services/ecs/test_ecs.py::TestTaskExecution::test_task_mount_host_volume
- tests/aws/services/eks/test_eks.py::TestK3SCluster::test_volume_mount

Basically, it seems we're using this mounted tmp as the directory to test mounted volume things. Should we just mount an arbitrary volume, something only for tests? That we could define? Or define another directory in `Directories` just for this? As this seems overkill to define a mounted `tmp` in the image we distribute just for testing purposes. 

Basically linked to this code snippet and this method `get_host_path_for_path_in_docker`:
```python
# note: directory must be mountable from the host, cannot use "tmpdir" fixture from pytest here...
local_tmpdir = DockerUtil.mountable_tmp_file()
host_tmpdir = get_host_path_for_path_in_docker(local_tmpdir)
```
@dfangl any insights into `DockerUtil.mountable_tmp_file`, does it really need to be mountable or can it be a regular file? Do we need another directory just for this, something like the cache one?

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -173,7 +173,7 @@ class Directories:
             var_libs=None,
             cache=str(cache_dir),  # used by analytics metadata
             tmp=tmp_dir,
-            mounted_tmp=None,
+            mounted_tmp=tmp_dir,
             functions=None,
             data=os.path.join(tmp_dir, "state"),  # used by localstack_ext config TODO: remove
             logs=os.path.join(tmp_dir, "logs"),  # used for container logs
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -40,7 +40,8 @@ class Directories:
         static_libs: container only; binaries and libraries statically packaged with the image
         var_libs:    shared; binaries and libraries+data computed at runtime: lazy-loaded binaries, ssl cert, ...
         cache:       shared; ephemeral data that has to persist across localstack runs and reboots
-        tmp:         shared; ephemeral data that has to persist across localstack runs but not reboots
+        tmp:         container only; ephemeral data that has to persist across localstack runs but not reboots
+        mounted_tmp: shared; same as above, but shared for persistence across different containers, tests, ...
         functions:   shared; volume to communicate between host<->lambda containers
         data:        shared; holds localstack state, pods, ...
         config:      host only; pre-defined configuration values, cached credentials, machine id, ...
@@ -52,6 +53,7 @@ class Directories:
     var_libs: str
     cache: str
     tmp: str
+    mounted_tmp: str
     functions: str
     data: str
     config: str
@@ -64,6 +66,7 @@ class Directories:
         var_libs: str,
         cache: str,
         tmp: str,
+        mounted_tmp: str,
         functions: str,
         data: str,
         config: str,
@@ -75,6 +78,7 @@ class Directories:
         self.var_libs = var_libs
         self.cache = cache
         self.tmp = tmp
+        self.mounted_tmp = mounted_tmp
         self.functions = functions
         self.data = data
         self.config = config
@@ -88,7 +92,8 @@ class Directories:
             static_libs="/usr/lib/localstack",
             var_libs=f"{DEFAULT_VOLUME_DIR}/lib",
             cache=f"{DEFAULT_VOLUME_DIR}/cache",
-            tmp=f"{DEFAULT_VOLUME_DIR}/tmp",
+            tmp=os.path.join(tempfile.gettempdir(), "localstack"),
+            mounted_tmp=f"{DEFAULT_VOLUME_DIR}/tmp",
             functions=f"{DEFAULT_VOLUME_DIR}/tmp",  # FIXME: remove - this was misconceived
             data=f"{DEFAULT_VOLUME_DIR}/state",
             logs=f"{DEFAULT_VOLUME_DIR}/logs",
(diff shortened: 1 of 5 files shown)


joe4dev:
> Edit: seems like some Lambda tests are failing because they expect `tmp` to be a mounted dir. I guess we could change that? @joe4dev not sure if we need to change the directory used for the test, or if this is a real problem.
> 
>     * tests/aws/services/lambda_/test_lambda.py::TestLayerHotReloading::test_layer_only_hot_reloading
> 
>     * tests/aws/services/lambda_/test_lambda.py::TestLayerHotReloading::test_layer_and_function_hot_reloading

I guess code mounting for hot reloading does not work if the `hot_reloading_dir_path` only exists in the container and is not mounted leading to the following errors:

```log
2023-09-21T22:09:17.2939962Z 2023-09-21T22:09:17.292 [33mWARNING[0m --- [  MainThread] localstack.utils.docker_utils : Error while performing automatic host path replacement for path '/tmp/localstack/hot-layer-16b58a31' to source '/tmp/localstack'
2023-09-21T22:09:17.3070495Z 2023-09-21T22:09:17.306 [32m INFO[0m --- [  asgi_gw_11] localstack.request.aws     : AWS lambda.PublishLayerVersion => 201
2023-09-21T22:09:17.3114320Z 2023-09-21T22:09:17.310 [33mWARNING[0m --- [  MainThread] localstack.utils.docker_utils : Error while performing automatic host path replacement for path '/tmp/localstack/hot-function-05ba236f' to source '/tmp/localstack'
2023-09-21T22:09:17.3409065Z 2023-09-21T22:09:17.339 [32m INFO[0m --- [   asgi_gw_4] localstack.request.aws     : AWS lambda.CreateFunction => 201
```

I guess the tmp directory is just an arbitrary path assumed to be mounted on the host:

```
tmp_path = config.dirs.tmp
hot_reloading_dir_path = os.path.join(tmp_path, f"hot-reload-{short_uid()}")
```

@dfangl Replacing `/tmp` with anything else that is mounted should work 🤔 ?

The advantage of `tmp` is that we assume automatic cleanup. We might need to add an explicit cleanup if using another directory.

dfangl:
I just want to clarify that this is a test-thing. The folder specified as hot reloading path has not be mounted into the container at all.
But for tests, this has to be the case, as we are running pytest inside the container, and therefore need a way to write to the host.

If we want to stop mounting config.dirs.tmp, we need some replacement for testing all the code-mounting stuff. We could use `DOCKER_CLIENT.mountable_tmp_file()` as a way to centralize all that, but that needs to be mounted to the host in some way. 

We could reuse and rename the "functions" dir for this, or introduce something new. But we definitely need some way to get a tmp dir mounted on the host for this.

bentsku:
Should we revisit this before `v3`? This seems like a considerable gain when mounting volumes and not using persistence. I'd be for adding some kind of new folder into `config.Directories` for testing purposes only, I think we need to separate test necessities and runtime `tmp`. I'll try to get something running and modify the test use of `config.dirs.tmp`.

localstack/config.py
```diff
@@ -100,19 +105,21 @@ def defaults() -> "Directories":
...
         """
         defaults = Directories.defaults()
+        tmp_dir = os.path.join(tempfile.gettempdir(), "localstack")
 
         return Directories(
             static_libs=defaults.static_libs,
             var_libs=defaults.var_libs,
             cache=defaults.cache,
-            tmp=defaults.tmp,
+            tmp=tmp_dir,
```
bentsku, line 119: note: use default here

bentsku:
Well, I have things a try again by looking for every access and call to `get_host_path_for_path_in_docker` to make it use a new `mounted_tmp` (naming to be decided, I was just validating the idea) and it now seems to be working. There are special use cases in -ext as well, but replacing them with `config.dirs.mounted_tmp` is seamless with the previous behaviour. This could be a solution.

alexrashed (approved):
Thanks for digging into this. The separation between temporary folders which should survive container recreations / are mounted, and those which shouldn't make sense.
And the performance impact could be quite big for non-hypervisor docker runs (where the file mounts have a terrible performance) and certain services which have lots of disk writes. 🚀 🚤

dfangl (approved):
Good work, looks good! I like the new separation!

Issue #9207, opened 2023-09-21
bug: LS Pro API GW {proxy+} route does not get matched properly based on http method


paulrobello:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have routes:
POST /zap
OPTIONS /{proxy+}  

The {proxy+} does not get matched with a request method of OPTIONS  to  /zap
the {proxy+} route does fire on OPTIONS if no other paths match

### Expected Behavior

the OPTIONS {proxy+} route should be matched since the request is using http method OPTIONS and the only other route is for method POST

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

starting via docker compose file

using terraform hcl to provision simple api-gw

### Environment

```markdown
- OS: Windows 11 x64
- LocalStack: localstack/localstack-pro:latest  (version reported from health route "2.2.1.dev20230920045530")
```

PR #9208, opened 2023-09-21
fix APIGW matching with {proxy+} and methods

bentsku:
## Motivation
As reported in #9207, we did not properly take into account the request method to retrieve the associated APIGW resource. We would take into account the path only, and then retrieve the method. 

## Changes
We now take the method into account, and don't match a path if it doesn't contain the request method or `ANY`. Added tests validating this behaviour against AWS, and adapted the unit tests too.
I suppose a bigger refactor would be nice around this, but this seemed like a rather urgent thing to fix. I'll take this into the APIGW backlog together with the more global refactor around the invocation. 

Sneaked a quick fix regarding fixture usage and cleanup for some export tests importing API.

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -718,12 +718,10 @@ def get_resource_for_path(
         for match in matches:
             match_methods = list(match[1].get("resourceMethods", {}).keys())
             # only look for path matches if the request method is in the resource
-            if method in match_methods or "ANY" in match_methods:
+            if method.upper() in match_methods or "ANY" in match_methods:
                 # check if we have an exact match (exact matches take precedence) if the method is the same
-                if match[0] == path:
-                    return match
-                elif path_matches_pattern(path, match[0]):
-                    # not an exact match but parameters can fit in
+                if match[0] == path or path_matches_pattern(path, match[0]):
+                    # either an exact match or not an exact match but parameters can fit in
                     return match
 
                 filtered_matches.append(match)
@@ -1360,12 +1358,7 @@ def get_target_resource_method(invocation_context: ApiInvocationContext) -> Opti
     if not resource:
         return None
     methods = resource.get("resourceMethods") or {}
-    method_name = invocation_context.method.upper()
-    return (
-        methods.get(method_name)
-        or methods.get("ANY")
-        or methods.get("X-AMAZON-APIGATEWAY-ANY-METHOD")
-    )
+    return methods.get(invocation_context.method.upper()) or methods.get("ANY")
 
 
 def get_event_request_context(invocation_context: ApiInvocationContext):
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -694,7 +694,9 @@ def get_rest_api_paths(account_id: str, region_name: str, rest_api_id: str):
 #  https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings
 #  -method-request.html
 #  https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html
-def get_resource_for_path(path: str, path_map: Dict[str, Dict]) -> Optional[Tuple[str, dict]]:
+def get_resource_for_path(
+    path: str, method: str, path_map: Dict[str, Dict]
+) -> tuple[Optional[str], Optional[dict]]:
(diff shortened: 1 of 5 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -706,25 +708,34 @@ def get_resource_for_path(path: str, path_map: Dict[str, Dict]) -> Optional[Tupl
...
-                return match
-
-        # at this stage, we have more than one match but we have an eager example like
-        # /{proxy+} or /api/{proxy+}, so we pick the best match by sorting by length
-        sorted_matches = sorted(matches, key=lambda x: len(x[0]), reverse=True)
+            match_methods = list(match[1].get("resourceMethods", {}).keys())
+            # only look for path matches if the request method is in the resource
+            if method in match_methods or "ANY" in match_methods:
+                # check if we have an exact match (exact matches take precedence) if the method is the same
+                if match[0] == path:
```
calvernaz, line 723: nit: given we are not doing significant work on the different switches it can be written like this
```
if (match[0] == path or match[0] != path and path_matches_pattern(path, match[0])):
    return match
```

localstack/services/apigateway/helpers.py
```diff
@@ -706,25 +708,34 @@ def get_resource_for_path(path: str, path_map: Dict[str, Dict]) -> Optional[Tupl
...
     # /foo/{param1}/baz and /foo/{param1}/{param2} for inputs like /for/bar/baz
     if len(matches) > 1:
-        # check if we have an exact match (exact matches take precedence)
-        for match in matches:
-            if match[0] == path:
-                return match
-
-        # not an exact match but parameters can fit in
+        filtered_matches = []
         for match in matches:
```
calvernaz, line 718: Ok I see what happened here, we used to do this (https://github.com/localstack/localstack/blob/0c93517c4fa5372f0236b70d5f4f7b5c55ad3164/localstack/services/apigateway/helpers.py#L1340) which since has not being used, but if we are moving this logic `get_resource_for_path` we should account for the uppercase mismatches, and the `X-AMAZON-APIGATEWAY-ANY-METHOD` situation.


Also remove the `get_target_resource_method` since is not used anymore and probably the cause of this gap.

```
def get_target_resource_method(invocation_context: ApiInvocationContext) -> Optional[Dict]:
    """Look up and return the API GW resource method for the given invocation context."""
    _, resource = get_target_resource_details(invocation_context)
    if not resource:
        return None
    methods = resource.get("resourceMethods") or {}
    method_name = invocation_context.method.upper()
    return (
        methods.get(method_name)
        or methods.get("ANY")
        or methods.get("X-AMAZON-APIGATEWAY-ANY-METHOD")
    )
```

calvernaz (commented):
Thanks for tackling this, I have two observations, but one is the most concerning one that I'd like to see tackled.

localstack/services/apigateway/helpers.py
bentsku: I don't think this is the same issue. 
`get_target_resource_method ` is still used in `get_target_resource_method_or_route` so we shouldn't remove it.

It does handle the method match afterwards, and we can still use it that way. 

It's just that if there is an exact match, we would return the resource, then try to get the method and there wouldn't be one. So we would just raise 404. But we need to be conscious of what methods a resource implements on the matching already. I don't think we need to change the logic yet, it works properly like that even if there is some duplication. 

Now we match the route, then get the method for it. Before we would match only on the path, then try to get the method even if it wasn't implemented. 

Also, we don't need to take care of `X-AMAZON-APIGATEWAY-ANY-METHOD` anymore since this has been fixed in the OpenAPI importer, it's now properly converted to `ANY` (that's what AWS stores internally as well). I can change the logic to take care of the uppercase mismatch but I don't think it's an issue either, because the `InvocationContext` method comes from the routing framework and this is always uppercase.

bentsku: After checking, it seems `find_best_candidate` has the proper logic there, but it's only used for `find_matching_http_route`

calvernaz: Sounds good, the only hit comes from `ext` and that mislead me. The logic to take in account the method also makes sense.

Yes the `find_best_candidate` was part of a refactor but I think the rules of precedence are a bit different.

calvernaz (approved):
Banger, thanks for handling this! LGTM 🚀

PR #9210, opened 2023-09-22
introduce CLI command categories to improve help display

thrau:
## Motivation

As the number of CLI commands increases, it becomes harder to separate basic from advanced functionality that is part of our paid offering. This PR introduces the notion of command categories into our `localstack` command group, that allow us to visually separate the commands.

image

The current implementation is very rudimentary, with grouping commands manually with a list of commands. I know this is not great, but it's the easiest way to get started without having to introduce additional concepts in ext for now.

## Changes

* Rename the `ExceptionCmdHandler` to a more generic `LocalStackCliGroup`
* Overwrite the `format_commands` method of `click.Group` with our own implementation that adds categories

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -17,17 +17,33 @@ from .console import BANNER, console
 from .plugin import LocalstackCli, load_cli_plugins
 
 
-class ExceptionCmdHandler(click.Group):
+class LocalStackCliGroup(click.Group):
     """
-    A Click group implementation which globally handles exceptions by:
+    A Click group used for the top-level ``localstack`` command group. It implements global exception handling
+    by:
+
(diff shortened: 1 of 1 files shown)


SimonWallner (approved):
LGTM! :rocket: 

I'm leaning a bit towards a 'running localstack' section with [start, stop, logs, ssh, wait] but we are facing diminishing returns quickly if we have too many sections as already discussed. so happy with it how it is rn.

This change is a good starting point and we should revisit it whenever we add/remove commands. people aint gonna interact with the help every day, so changing it more often, whenever the need arises should not be a issue.

alexrashed (approved):
I think this is a nice first iteration. I agree that it would be nice to have a way to mark a command as "advanced", but we can look into this once this format has settled. 🚀

PR #9211, opened 2023-09-22
use run_interactive for localstack ssh to fix tty issues

thrau:
## Motivation

While working on #9210, I noticed that `localstack ssh` had some wonky terminal behavior.
I couldn't use CTRL+A, CTRL+E, CTRL+R and other control sequences correctly. New lines were also added twice.

Peek 2023-09-22 18-03

It looks like the `tty=True` option of `run` doesn't really work correctly. <strike>For #8994 I added a `run_interactive` method that is a much simpler implementation of running an interactive command in a subprocess.</strike> Using `os.execlp` (based on @dfangl's suggestion), is much simpler. This works much better now:

Peek 2023-09-22 18-03-update

## Changes

* Use `run_interactive` for `localstack ssh`

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -12,7 +12,6 @@ from localstack.cli.exceptions import CLIError
 from localstack.utils.analytics.cli import publish_invocation
 from localstack.utils.bootstrap import get_container_default_logfile_location
 from localstack.utils.json import CustomEncoder
-from localstack.utils.run import run_interactive
 
 from .console import BANNER, console
 from .plugin import LocalstackCli, load_cli_plugins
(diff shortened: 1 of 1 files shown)


localstack/cli/localstack.py
```diff
@@ -617,15 +618,13 @@ def cmd_ssh() -> None:
...
-    from localstack.utils.run import run
 
     if not DOCKER_CLIENT.is_container_running(config.MAIN_CONTAINER_NAME):
         raise CLIError(
             f'Expected a running LocalStack container named "{config.MAIN_CONTAINER_NAME}", but found none'
         )
     try:
-        process = run("docker exec -it %s bash" % config.MAIN_CONTAINER_NAME, tty=True)
-        process.wait()
+        run_interactive(["docker", "exec", "-it", config.MAIN_CONTAINER_NAME, "bash"])
```
dfangl, line 627: I would argue that we do not need run_interactive here. It would only be necessary when we want to do something afterwards.

In this case, I would suggest using `os.execv*` here, as replacing the current process with the new one makes sure every signal is handled as intended.

thrau: fair point. i don't really understand `os.exec` well and how it works on other platforms, but works fine on my machine this way :+1:

dfangl: It replaces the current process image with a new process image. Instead of running two processes, we only run one, and all signals will therefore get directly to docker, without any logic on our side :)

dfangl (approved):

localstack/cli/localstack.py
dfangl: Should be available on unix and windows :)

PR #9212, opened 2023-09-22
[SFN] Parameters filtering for lambda resource tasks

MEPalma:
## Motivation
Resolves issue for which lambda resource task would forward extra incorrect arguments if these were specified in Parameters declaration. 

## Changes
The issue is resolved by filtering by the supported parameters of the lambda invoke api action.


dominikschubert (approved):
:+1:  Good catch

PR #9213, opened 2023-09-22
fix apigw pathParameters context resolving before authorizer run

bentsku:
## Motivation
This issue was reported in a support case, the `pathParams` field was not available in the authorizer context. After checking, it seems we only resolved and added to the invocation context at the invocation step, which is executed after the authorizer.

~Beware, this PR is on top of #9208 because the new tests are using this routing to test different scenarios with the path.~

## Changes
Updated a util to set the context as soon as we have the resource details (so that we're able to resolve the path parameters), and removed the resolving in the integration step.


whummer (approved):
Great catch @bentsku ! 👍 LGTM (building on #9208)

PR #9218, opened 2023-09-25
upgrade flake8 and black

alexrashed:
## Motivation
The latest update of pyproject-flake (6.1.0) recently broke the build.
We pinned the version with https://github.com/localstack/localstack/pull/9215 to unblock the pipeline.
This PR removes the pin and upgrades black, flake8, and all related packages, to their latest versions.

## Changes
- Upgrade black, flake8, and all related packages, to their latest versions. 
- Fix new linting errors (flake8 upgrade).
- Reformat code with new black version (black upgrade).


thrau:
@alexrashed wouldn't it be easier to just exclude the offending rules (like `E721`) from our linter?

alexrashed:
@thrau: Why do you think this would be a better way? There were only a few violations, and I for myself think that E721 is quite useful. Explicit type checks are rarely useful, and they could be excluded if you really think.

thrau:
i was going to say that i generally would prefer a stable linter/formatting config rather than being subject to updates. but after thinking about it for a few minutes, i think it's actually the other way round. standards are better :dagger:

thrau (approved):
LGTM, let's move forward

alexrashed:
Superseded by https://github.com/localstack/localstack/pull/9399

PR #9228, opened 2023-09-26
Implementation of Route53 RecordSet resource provider for CloudFormation

Morijarti:
## Motivation
Implementation of "AWS::Route53::RecordSet" resource provider for CloudFormation

## TODO
In future we should add more tests and improve code to better handle invalid inputs and missing values


localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,181 @@
...
+          - /properties/Id
+
+
+
+        """
+        model = request.desired_state
+        route53 = request.aws_client_factory.route53
+        if not model.get("HostedZoneId"):
+            # if only name was provided for hosted zone
+            zone = route53.list_hosted_zones_by_name(DNSName=model["HostedZoneName"])[
```
simonrw, line 91: What if the `HostedZoneName` was not specified? We have some code that handles that case in the [`GenericBaseModel` implementation](https://github.com/localstack/localstack/blob/49298070be47907dd0ebfcb5d2d5e13b404ba0fa/localstack/services/cloudformation/models/route53.py#L75-L76) you can copy.

localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,181 @@
...
+            ][0]
+            model["HostedZoneId"] = zone["Id"]
+
+        resource_records = [{"Value": record} for record in model["ResourceRecords"]]
+        route53.change_resource_record_sets(
+            HostedZoneId=model["HostedZoneId"],
+            ChangeBatch={
+                "Changes": [
+                    {
+                        "Action": "UPSERT",
```
simonrw, line 102: I like the use of the upsert here, it means we don't have to worry about existing values.

localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,181 @@
...
+                        "ResourceRecordSet": {
+                            "Name": model["Name"],
+                            "Type": model["Type"],
+                            "TTL": int(model["TTL"]),
+                            "ResourceRecords": resource_records,
```
simonrw, line 107: This is only a subset of what can be set on a `ResourceRecordSet` entry, right? The [previous implementation](https://github.com/localstack/localstack/blob/49298070be47907dd0ebfcb5d2d5e13b404ba0fa/localstack/services/cloudformation/models/route53.py#L33-L44) supported more fields than this. Could we select the relevant attributes like the `GenericBaseModel`?

The previous model also handled things like the TTL being the wrong type, and defaulting the `EvaluateTargetHealth` property.

simonrw (changes requested):
Looks good in general, a few nits but the main takeaway is: we have some functionality in the previous `GenericBaseModel` that is missing from this implementation. Can you add it back in?

tests/aws/services/cloudformation/resources/test_route53.py
```diff
@@ -18,6 +18,22 @@ def test_create_record_set_via_id(route53_hosted_zone, deploy_cfn_template):
...
+    stack = deploy_cfn_template(
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../../templates/route53_hostedzoneid_template.yaml"
+        ),
+        parameters=parameters,
+        max_wait=300,
+    )
+    stack.destroy()
+
+
```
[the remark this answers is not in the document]
Morijarti: Agreed. I'll remove it.

localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,181 @@
...
+        Create-only properties:
+          - /properties/HostedZoneName
+          - /properties/Name
+          - /properties/HostedZoneId
+
+        Read-only properties:
+          - /properties/Id
+
+
+
```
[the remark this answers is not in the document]
Morijarti: Done

localstack/services/route53/resource_providers/aws_route53_recordset.py
Morijarti: I've added functionality that was present in previous implementation.

localstack/services/route53/resource_providers/aws_route53_recordset.py
simonrw: For the new resource providers, we should return a progress event with `OperationStatus.FAILED` but it looks like that event [goes nowhere...](https://github.com/localstack/localstack/blob/49298070be47907dd0ebfcb5d2d5e13b404ba0fa/localstack/services/cloudformation/engine/template_deployer.py#L916)

For now I think raising the `Exception` is fine.

simonrw (approved):
Great, thanks for making those changes!

pinzon (approved):
LGTM

localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,210 @@
+# LocalStack Resource Provider Scaffolding v2
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Optional, Type, TypedDict
+
+from mypy_boto3_route53 import Route53Client
```
simonrw, line 7:
```suggestion
if TYPE_CHECKING:
    from mypy_boto3_route53 import Route53Client
```

localstack/services/route53/resource_providers/aws_route53_recordset.py
```diff
@@ -0,0 +1,210 @@
...
+                    },
+                ]
+            },
+        )
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model=model,
+        )
+
+    def get_hosted_zone_id_from_name(self, hosted_zone_name: str, client: Route53Client):
```
simonrw, line 136:
```suggestion
    def get_hosted_zone_id_from_name(self, hosted_zone_name: str, client: "Route53Client"):
```

simonrw (commented):
A couple of suggestions for fixing the pipeline

The pipeline is failing (with an unhelpful message) because the resource provider cannot be loaded. It cannot be loaded because of an import of a package we do not install as part of the runtime, only during development.

PR #9229, opened 2023-09-26
Track DNS_ADDRESS usage in events

simonrw:
## Motivation

With the new DNS server, we want to know how many people are
* disabling the server with `DNS_ADDRESS=0`,
* or customising the value because of conflicting programs or other reasons.

## Changes

* Track the value of `DNS_ADDRESS`

## Motivation for tracking the value

* We care if the value is `0` or an IP address
* The IP address value is not private, and 99% of the time will be either `127.0.0.1` or `0.0.0.0`.


ackdav (approved):
cool, lgtm :+1: nice to see this being used :)

PR #9230, opened 2023-09-26
Add Kinesis Stream resource provider for CloudFormation

Morijarti:
## Changes
Add Kinesis Stream resource provider for CloudFormation

## TODO

What's left to do:
Current implementation ignores settings for Tags, StreamEncryption and RetentionPeriodHours


localstack/services/kinesis/resource_providers/aws_kinesis_stream.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        model["Arn"] = kinesis.describe_stream(StreamName=model["Name"])["StreamDescription"][
+            "StreamARN"
+        ]
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model=model,
+            custom_context=request.custom_context,
+        )
```
pinzon, line 102: If I rembeber correctly the creation of a Kinesis Stream is an async process and Cloudformation won't return a success event until the Stream has an status active.

pinzon (changes requested):

localstack/services/kinesis/resource_providers/aws_kinesis_stream.py
Morijarti: Thanks for information. I'll make necessary changes.

Morijarti: Done

PR #9232, opened 2023-09-26
Let hypercorn listen directly to privileged ports if run as root

dfangl:
## Motivation
This is basically a continuation of #8948 and #9012.

We basically reintroduce hypercorn binding directly to the privileged ports, if run as root.

Additionally, we check the `sys.net.ipv4.ip_unprivileged_port_start` kernel parameter, to check the port start of unprivileged ports.

We will need to provide proper paths of error handling, to avoid #9012 again.

## Changes
* Bind hypercorn to privileged ports if root
* Check the actual unprivileged port start instead of assuming 1024.

### DIFF
--- a/localstack/services/edge.py
+++ b/localstack/services/edge.py
@@ -489,6 +489,11 @@ def start_edge(listen_str: str, use_ssl: bool = True, asynchronous: bool = False
     # separate privileged and unprivileged addresses
     unprivileged, privileged = split_list_by(listen, lambda addr: addr.is_unprivileged() or False)
 
+    # if we are root, we can directly bind to privileged ports as well
+    if is_root():
+        unprivileged = unprivileged + privileged
+        privileged = []
+
     # check that we are actually started the gateway server
     if not unprivileged:
         unprivileged = parse_gateway_listen(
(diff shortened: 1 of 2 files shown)


localstack/services/edge.py
```diff
@@ -489,6 +489,11 @@ def start_edge(listen_str: str, use_ssl: bool = True, asynchronous: bool = False
     # separate privileged and unprivileged addresses
     unprivileged, privileged = split_list_by(listen, lambda addr: addr.is_unprivileged() or False)
 
+    # if we are root, we can directly bind to privileged ports as well
+    if is_root():
+        unprivileged = unprivileged + privileged
+        privileged = []
```
simonrw, line 495: Last time I tried making this change, it broke the -ext pipeline. Can you make sure you do a manual run against this branch.

dfangl: https://github.com/localstack/localstack-ext/pull/2131 running here. Let's hope the double start issue was the culprit last time :)

dfangl: Resolved the issue as discussed, the cause was a double-adding of the port 443 in ext tests.

PR #9233, opened 2023-09-26
fix S3 PutObject with MD5 and enabled ARN Partition Rewriting

alexrashed:
## Motivation
#9064 introduced the MD5 checks for S3's PutObject operation.
Until now, the ARN Partition Rewriting did not update the MD5 hashes in requests.
This now causes issues with the MD5 check in S3, since the checksum is verified, but the content is maybe modified due to the rewriting.

## Changes
- Replaces the `Content-MD5` header with a new value in case it is set.

### DIFF
--- a/localstack/aws/handlers/partition_rewriter.py
+++ b/localstack/aws/handlers/partition_rewriter.py
@@ -1,3 +1,5 @@
+import base64
+import hashlib
 import logging
 import re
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Thanks for quickly explaining the background on this! Solution looks good 👍

PR #9234, opened 2023-09-26
fix opensearch-knn plugin by setting LD_LIBRARY_PATH

### LINKED ISSUE #9193: Bug: [EDITED] java.lang.UnsatisfiedLinkError: no opensearchknn_faiss in java.library.path

### Current Behavior

I have a terraform-managed localstack setup that uses s3 and opensearch. 

In our docker-compose for localstack, we set the OPENSEARCH_ENDPOINT_STRATEGY to path, and when we use the aws cli for opensearch we can see the endpoint follows the expected path strategy. 

However, when we curl that endpoint, we get the following error:

```
$ curl http://localhost:4566/opensearch/us-east-1/example/
<?xml version='1.0' encoding='utf-8'?>
<Error><Code>InternalError</Code><Message>exception while calling s3 with unknown operation: MyHTTPConnectionPool(host='127.0.0.1', port=60769): Max retries exceeded with url: / (Caused by NewConnectionError('&lt;urllib3.connection.HTTPConnection object at 0x7efc3c205840&gt;: Failed to establish a new connection: [Errno 111] Connection refused'))</Message><RequestId>09b749b2-78a0-4a90-b175-9f672f14babc</RequestId></Error>
```
looks like localstack is routing to s3 instead of opensearch. Interestingly, this issue only started after I created an opensearch index and added a document to it -- before that, I was getting the expected opensearch curl results.

### Expected Behavior

I would expect the curl to hit the opensearch endpoint instead of s3

With a docker-compose file

### Steps To Reproduce

I'm not fully certain what the minimal repro is. We're starting localstack with a docker compose, which I can include:

```
version: "3"
services:
  localstack-mongodb:
    image: mongo:4.4.21
    container_name: localstack-mongodb
    ports:
      - "27017:27017"
    command: mongod --replSet rs0

    # Enable interactive mode
    stdin_open: true

    # Allocate a pseudo-TTY
    tty: true

  localstack-redis:
    image: redis:7.0.11
    container_name: localstack-redis
    ports:
      - "6379:6379"

  localstack:
    image: localstack/localstack
    container_name: localstack
    environment:
     - OPENSEARCH_ENDPOINT_STRATEGY=path
    ports:
      - "4566:4566"
      - "4510-4559:4510-4559"
```

followed by:
opensearch.js 
```
opensearchClient.create({
  index: 'foo'.
  body: {
    settings: { 'index.knn': true },
    mappings: {
      embedding: {
        type: 'knn_vector',
        dimension: 512,
        method: { engine: 'faiss', name: 'nsw' }
      }
    }
  } 
})

client.index({ index: 'foo', body: { embedding: [ ... 512 numbers ] });
```

### Environment

```markdown
- OS: 20.04
- LocalStack: latest
```

### Anything else?

NA

alexrashed:
## Motivation
The OpenSearch KNN plugin is currently not working in LocalStack, because we do not properly add the KNN libraries to the `LD_LIBRARY_PATH` when starting OpenSearch.
See the awesome bug report by @theahura in https://github.com/localstack/localstack/issues/9193 for more details.

## Changes
- When starting OpenSearch, we check if the specific installation contains a directory at `./plugins/opensearch-knn/lib`.
- If so, we add this path to an existing `LD_LIBRARY_PATH`.

## Testing
- An existing test was extended (to not waste additional resources in the CI run) which creates - in addition to some plain indices - an index where the KNN plugin is enabled.
- Afterwards a document is added and retrieved.
- The actual error (return status code 500), in case the plugin is not working properly, actually only occurs when retrieving the document!

Fixes #9193

### DIFF
--- a/localstack/services/opensearch/cluster.py
+++ b/localstack/services/opensearch/cluster.py
@@ -461,10 +461,18 @@ class OpensearchCluster(Server):
         return cmd
 
     def _create_env_vars(self, directories: Directories) -> Dict:
-        return {
+        env_vars = {
             "OPENSEARCH_JAVA_OPTS": os.environ.get("OPENSEARCH_JAVA_OPTS", "-Xms200m -Xmx600m"),
             "OPENSEARCH_TMPDIR": directories.tmp,
         }
+
+        # if the "opensearch-knn" plugin exists and has a "lib" directory, add it to the LD_LIBRARY_PATH
+        # see https://forum.opensearch.org/t/issue-with-opensearch-knn/12633
+        knn_lib_dir = os.path.join(directories.install, "plugins", "opensearch-knn", "lib")
+        if os.path.isdir(knn_lib_dir):
+            env_vars["LD_LIBRARY_PATH"] = f"{knn_lib_dir}:{os.environ.get('LD_LIBRARY_PATH', '')}"
+
+        return env_vars
 
     def _log_listener(self, line, **_kwargs):
         # logging the port before each line to be able to connect logs to specific instances
--- a/tests/aws/services/opensearch/test_opensearch.py
+++ b/tests/aws/services/opensearch/test_opensearch.py
@@ -563,6 +563,33 @@ class TestOpensearchProvider:
             assert req_result[0]["health"] in ["green", "yellow"]
             assert req_result[0]["index"] in indices
 
+        # create a knn index to make sure the knn plugin works
+        index_path = f"{opensearch_endpoint}/knn"
+        body = {
+            "settings": {"index": {"knn": True}},
+            "mappings": {
+                "properties": {
+                    "embedding": {
+                        "type": "knn_vector",
+                        "dimension": 2,
+                    }
+                }
+            },
+        }
+        put = requests.put(index_path, headers=COMMON_HEADERS, json=body)
+        assert put.status_code == 200
+        get = requests.get(f"{opensearch_endpoint}/_cat/indices/knn?format=json&pretty")
+        assert get.status_code == 200
+
+        # add a document
+        document_path = f"{opensearch_endpoint}/knn/_doc/test_document"
+        body = {"embedding": [1, 2]}
+        put = requests.put(document_path, headers=COMMON_HEADERS, json=body)
+        assert put.status_code == 201
+
+        get = requests.get(document_path)
+        assert get.status_code == 200
+
     @markers.aws.unknown
     def test_get_document(self, opensearch_document_path):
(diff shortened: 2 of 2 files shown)


theahura:
@alexrashed it's worth taking a peek at this additional issue on the opensearch knn library: https://github.com/opensearch-project/k-NN/issues/1150

they maintain a script, `opensearch-tar-install.sh`, that should set all the appropriate env variables

alexrashed:
Thanks @theahura!
I've seen the referenced issue and the `opensearch-tar-install.sh`, but I would like to first fix your issue.
This way, we get the fix in the upcoming release (tomorrow).
In a later iteration, I would like to consider this script and maybe use it in our implementation. However, this would cause a considerable refactoring of the OpenSearch server management (which we don't want to do so close to the release).

localstack/services/opensearch/cluster.py
```diff
@@ -461,11 +461,19 @@ def _create_run_command(
...
+        env_vars = {
             "OPENSEARCH_JAVA_OPTS": os.environ.get("OPENSEARCH_JAVA_OPTS", "-Xms200m -Xmx600m"),
             "OPENSEARCH_TMPDIR": directories.tmp,
         }
 
+        # if the "opensearch-knn" plugin exists and has a "lib" directory, add it to the LD_LIBRARY_PATH
+        # see https://forum.opensearch.org/t/issue-with-opensearch-knn/12633
+        knn_lib_dir = os.path.join(directories.install, "plugins", "opensearch-knn", "lib")
+        if os.path.isdir(knn_lib_dir):
+            env_vars["LD_LIBRARY_PATH"] = f"{knn_lib_dir}:{os.environ.get('LD_LIBRARY_PATH', '')}"
```
dfangl, line 473: Is there a reason for this order? This way, knn_lib_dir would override any potential shared libraries provided by LD_LIBRARY_PATH already.
This might make sense, or could lead to (unlikely) problems if someone wants to use LD_LIBRARY_PATH to load some other shared libraries, but knn_lib_dir has one with the same name.

It is very unlikely that this case will arise, I just want to make sure we make a conscious decision in one way or another here.

dfangl (approved):
LGTM in general! Just one comment about the order in `LD_LIBRARY_PATH`

localstack/services/opensearch/cluster.py
alexrashed: Good catch! I changed the order and ensured that an empty path cannot be added (in case the LD_LIBRARY_PATH is not set on the machine, an empty path would mean the current working directory).

PR #9238, opened 2023-09-27
add label sync, migrate workflows to shared workflows

alexrashed:
## Motivation
We are currently working on unifying processes across multiple repos in the GitHub organization.
For this purpose, we created an internal repo with some shared workflows.
This PR onboards `localstack/localstack` to these shared workflows.

## Changes
- Add the new "Sync Labels" workflow.
  - This means that labels defined in our shared repo are automatically added in this repo.
  - This allows using a specific set of labels for other automations (like the project status sync) in multiple repos.
  - The workflow allows defining multiple "categories", which is a way in defining different sets of labels for different kinds of repos.
- "Sync Project Cards" basically just migrates the existing code to the shared workflow.
  - However, the workflow has been extended a bit such that issues are also automatically added to the projects.
  - This means we do not have to rely on the super-low limit of GitHub project workflow automations (we don't have to have an automation for projects using this workflow).
- "Triage Stale Issues" also just migrates the existing code to the new shared workflow.
  - However, the configuration has been slightly adjusted:
    - The time before an issue becomes marked as "stale" from 60 days to 14 days.
    - The time before a stale issue gets closed from 14 days to 7 days.

## Testing
These workflows have been tested in a separate private repo.

### DIFF
--- a/.github/workflows/stale-bot.yml
+++ b/.github/workflows/stale-bot.yml
@@ -6,44 +6,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  issue-cleanup:
-    runs-on: ubuntu-latest
-    name: Triage Stale Issues
-    steps:
-    - uses: aws-actions/stale-issue-cleanup@v6
-      with:
-        issue-types: issues
-        ancient-issue-message: Hello 👋! It looks like this issue hasn’t been active in longer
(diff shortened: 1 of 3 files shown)


thrau (approved):
very nice! didn't even know you could reference workflows from other repos like this. really cool!

looking forward to seeing this in action

Issue #8533, opened 2023-06-20
bug: lambda docker executor fails when using auto-generated port


LaurentGoderre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using the Remote Debugging config from the docs for Lambda when debugging multiple lambda and trying to use random port, the containers fail to start.

docker-compose.yml

```yaml
localstack:
    image: localstack/localstack
    ports:
      - "4566:4566"
      - "4510-4559:4510-4559"
    environment:
      - LAMBDA_DOCKER_FLAGS=-e NODE_OPTIONS=--inspect=0.0.0.0:9229 -p 9229
      - LAMBDA_REMOTE_DOCKER=false
      - LAMBDA_EXECUTOR=docker-reuse
      - DOCKER_HOST=unix:///var/run/docker.sock
    volumes:
      - "${TMP_DIR:-./.localstack}:${TMP_DIR:-/tmp/localstack}"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

Then

```bash
serverless deploy --stage local
serverless invoke --stage local --function hello
```

Localstack logs

```
fable-app-aws-1   | 2023-03-31T17:59:01.339  INFO --- [ncthread4223] localstack.utils.threads   : Thread run method <function LambdaExecutor.execute.<locals>.do_execute at 0x7f1eafddcca0>(None) failed: ('Invalid port string provided: %s', '9229') Traceback (most recent call last):
fable-app-aws-1   |   File "/opt/code/localstack/localstack/utils/threads.py", line 58, in run
fable-app-aws-1   |     result = self.func(self.params, **kwargs)
fable-app-aws-1   |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 484, in do_execute
fable-app-aws-1   |     return _run(func_arn=func_arn)
fable-app-aws-1   |   File "/opt/code/localstack/localstack/utils/cloudwatch/cloudwatch_util.py", line 183, in wrapped
fable-app-aws-1   |     raise e
fable-app-aws-1   |   File "/opt/code/localstack/localstack/utils/cloudwatch/cloudwatch_util.py", line 179, in wrapped
fable-app-aws-1   |     result = func(*args, **kwargs)
fable-app-aws-1   |   File "/opt/code/localstack/localstack/services/awslambda/lambda_executors.py", line 471, in _run
fable-app-aws-1   |     raise e
[...]

PR #9241, opened 2023-09-27
allow single container port to auto-assigned host port mapping

kuvandjiev:
## Motivation
#8533

## Changes
Will allow adding 
`LAMBDA_DOCKER_FLAGS=-e NODE_OPTIONS=--inspect-brk=127.0.0.1:9229 -p 9229`
as environmental variable to Localstack in docker-compose. This will allow new containers to be started for lambda functions with an auto-assigned host port mapped to the container port.

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -365,7 +365,7 @@ def get_docker_image_to_start():
 
 
 def extract_port_flags(user_flags, port_mappings: PortMappings):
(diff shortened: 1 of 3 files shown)


kuvandjiev:

dfangl:
A similar approach was merged in #9420 . It should be available when specifying the host port with `0` like this:  `-p 0:<port>`.
Still, thank you for your contribution!

PR #9246, opened 2023-09-28
Fix partition rewriting for CloudFormation by guarding internal calls

dominikschubert:
## Motivation

The ARN/partition rewriting feature isn't widely used but it was designed as a low-effort way to add support for non-default (i.e. not `aws`) partitions. It works by rewriting incoming requests and outgoing responses based on the detected region in the ARN. 

Since we had the contract to only  rewrite external requests and ignore internal ones, this lead to issues in CloudFormation for example.

Another contributing factor was that legacy models used internal client credentials which work with the legacy detection of internal clients, while the newer `ResourceProvider`-based resources had `test`/`test` credentials.
They were internal clients according to `context.is_internal_call`, but *not* according to `is_internal_call_context(request.headers)`.
Since the rewriter only detected based on `is_internal_call_context`, it therefore didn't consider clients for newer models as internal.

With this PR we fix a few of these issues while also introducing a guard for internal calls which will rewrite both request and response to only include the `DEFAULT_INBOUND_PARTITION` (`aws`). This basically enforces our contract that internally there should only ever be the `DEFAULT_INBOUND_PARTITION` and protects it from any dynamic "injection" caused by internal clients. 

## Changes

This PR extends the functionality of the ARN rewriter a bit to make the rewriting more configurable.
### new modes
There are multiple operation modes you can choose for the rewriting:
- `request`: only the requests gets rewritten (to DEFAULT_INBOUND_PARTITION)
- `response`: only the response gets rewritten (to original partition based on the region)
- `bidirectional`: both request and response are rewritten as described above
- `internal-guard`: both request and response are rewritten, but the response is also rewritten to `DEFAULT_INBOUND_PARTITION` (!)
Default behavior for external clients is `bidirectional`.
Default behavior for internal clients is `internal-guard`.
Behavior can be overwritten by setting the "LS-INTERNAL-REWRITE-MODE" header

### other changes
- make internal client detection more robust
- unify client creation in CloudFormation

## TODO

- [ ] Should extend the unit tests to cover the new functionality
- [ ] Some legacy resources still use `connect_to` in the model definitions themselves. This isn't really clean since they'll have the wrong credentials, but for now it's fine because of the improved detection of internal clients.


alexrashed (approved):
Awesome! Thanks for digging into this! The new modes seem fine, and I think we can overlook the absence of tests for now (the whole rewriter should be deleted in favor of a long overdue refactoring for full partition support). 🚀 
I only found a nit about a comment, nothing blocking a merge.

PR #9251, opened 2023-09-28
cleanup and refactor scenario tests

steffyP:
## Motivation
Restructuring the scenario tests; and adding validation where we missed it before.

## Changes
* Package restructuring
   * moved apigateway test to `services` 
* added snapshots


dominikschubert (approved):
LGTM, thanks for churning through this and unifying it :rocket:

PR #9256, opened 2023-09-29
fix service router test by excluding bedrock

alexrashed:
## Motivation
This PR fixes the service router unit tests by excluding a newly added AWS Bedrock (which was added to `botocore` with https://github.com/boto/botocore/commit/730bfe40b1431f229d68f0d525756aaf67742a80)

## Changes
Exclude bedrock for now from the service router testing.


dominikschubert (approved):

alexrashed:
Since this commit really only touches a single unit test (which is now green again), the current pipeline run is failing due to a flake, and due to the fact that this is blocking the pipeline right now, I'll move forward and merge this one.

PR #9257, opened 2023-09-29
Stabilize first version of CDK-based InfraProvisioner

dominikschubert:
## Motivation

We've previously identified a few issues when working with the new CDK-based test setup which this PR aims to address.

While I was also experimenting with full asset support, this is probably a bit too far out for now, and therefore this PR aims to stabilize the current API a bit.

## Changes

- Replaced create_stack calls with more flexible change set generation
- Adds support for stack updates, allowing iterating over parts of the stack while developing the infra setup
- Persists templates (see point below)
- InfraProvisioner now requires `namespace` to be set to a repo-unique value. This is used for namespacing the generated templates in `tests/aws/cdk_templates/{namespace}/`.
- Upload CloudFormation template to s3 and reference it via TemplateURL instead of direct usage in TemplateBody if size exceeds CloudFormation upload limits (51_200 characters)
- Stack outputs are fetched lazily via the API instead of fetching them immediately after deployment in all cases
- 2 ways to force regeneration of stack templates
  - set `TEST_INFRA_UPDATE_TEMPLATE=1` (name subject to change)
  - add `force_template_update=True` when instantiating an `InfraProvisioner`
- deprecate `add_custom_setup_provisioning_step` in favor of `add_custom_setup` to unify naming with teardown (`add_custom_teardown`)

General setup looks like this now:

```python
    @pytest.fixture(scope="class", autouse=True)
    def infrastructure(self, infrastructure_setup):
        # note the use of the fixture here. we use it for access to the local cdk_templates path for  -ext compatibility as well as automatic client access
        infra = infrastructure_setup(
            namespace="SomeNamespace", force_template_update=True
        )
        stack1 = cdk.Stack(infra.cdk_app, "Stack11")
        stack2 = cdk.Stack(infra.cdk_app, "Stack22")
        topic1 = cdk.aws_sns.Topic(stack1, "topic")
        cdk.aws_ssm.StringParameter(stack1, "param", string_value="test")
        topic2 = cdk.aws_sns.Topic(
            stack2, "topic", topic_name=cdk.Fn.join("-", [topic1.topic_name, "suffix"])
        )
        cdk.CfnOutput(stack1, "TopicName", value=topic1.topic_name)
        cdk.CfnOutput(stack2, "TopicName", value=topic2.topic_name)

        with infra.provisioner() as prov:
            yield prov
```

## TODO

- [x] Add docstrings for user facing API
- [x] Make template persistence compatible with -ext (or disable for now)
- [x] Refactor existing tests
- [x] fix s3 URL when deploying against localstack


localstack/testing/scenario/provisioning.py
```diff
@@ -16,85 +23,168 @@
...
+                    Bucket=template_bucket_name, Key=key, Body=stack["Template"]
+                )
+                url = self.aws_client.s3.generate_presigned_url(
+                    ClientMethod="get_object",
+                    Params={"Bucket": template_bucket_name, "Key": key},
+                    ExpiresIn=1600,
+                )
+                # domain = "amazonaws.com" if is_aws_cloud() else f"localhost:{config.EDGE_PORT}"
+                # url = f"https://{template_bucket_name}.s3.{domain}/{key}"
+                # f"https://s3.us-east-1.amazonaws.com/{template_bucket_name}/{stack_name}.yaml"
```
bentsku, line 135: what is the issue with S3 in that case? shouldn't the pre-signed URL use the endpoint configured in the client? what is the problematic behaviour?

localstack/testing/scenario/provisioning.py
```diff
@@ -16,85 +23,168 @@
...
+    "Delay": 6,
+    "MaxAttempts": 600,
+}  # total timeout ~1 hour (6 * 600 = 3_600 seconds)
+WAITER_CONFIG_LS = {"Delay": 1, "MaxAttempts": 600}  # total timeout ~10 minutes
+TEMPLATE_DIR_NAME = "cdk_templates"
 
 
-def cleanup_s3_bucket(s3_client: "S3Client", bucket_name: str, delete_bucket: bool = False):
-    LOG.debug(f"Cleaning provisioned S3 Bucket {bucket_name}")
+def _cleanup_s3_bucket(s3_client: "S3Client", bucket_name: str, delete_bucket: bool = False):
```
steffyP, line 34: this function is currently also used in some of the scenario tests to cleanup buckets that have been created manual e.g the corresponding tear down for add_custom_setup_provisioning_step.

I think it makes more sense to move this function to a cdk-utility?

bentsku: It also has a lot in common with the S3 cleaning fixture. I had an issue with the above, as the bucket created with CDK was versioned so this didn't clean up properly. Maybe the code inside the fixture could be extracted as an util in S3, the used in both the fixture and this cdk util?

localstack/testing/scenario/provisioning.py
dominikschubert: nothing actually, was an older issue

localstack/testing/scenario/provisioning.py
dominikschubert: Left it in there for now but fixed the naming for the usage in tests again. We can tackle this in a follow-up :+1:

dominikschubert:
@steffyP  FYI the single test failure right now is a flake (docker sdk tests)

localstack/testing/scenario/provisioning.py
```diff
@@ -108,71 +234,135 @@ def provision(self):
...
+        warnings.warn(
+            "`add_custom_setup_provisioning_step` is deprecated. Use `add_custom_setup`",
+            DeprecationWarning,
+        )
+        self.add_custom_setup(setup_task)
 
-    def bootstrap_cdk(self):
+    def _bootstrap(self):
         # TODO: add proper bootstrap template to deploy here if there's no parameter yet
+        self._create_bucket_if_not_exists(self.get_asset_bucket())
```
steffyP, line 358: 🎉

steffyP (approved):
Amazing set of changes 🚀 👏 

The persistence of the templates will be speeding up things, and the update of stacks will also save time 😄 I also like the new concept of having an asset-bucket where we can upload lambda zips.
 
Thanks for adding so many features 🚀

steffyP:
@dominikschubert: I think the `infra_provisioner` fixture from the code sample in the description needs to be renamed to `infrastructure_setup`?  🤔

steffyP (approved):
🚀

PR #9259, opened 2023-09-29
Implementation of CloudFormation support for AWS::KinesisFirehose

Morijarti:
## Changes
Implementation of CloudFormation support for AWS::KinesisFirehose::DeliveryStream


pinzon (approved):
LGTM

localstack/services/cloudformation/resource_provider.py
```diff
@@ -60,7 +60,8 @@
     "AWS::Route53::HealthCheck": "ResourceProvider",
     "AWS::Route53::RecordSet": "ResourceProvider",
     "AWS::SNS::Topic": "ResourceProvider",
-    "AWS::Kinesis::Stream": "ResourceProvider"
+    "AWS::Kinesis::Stream": "ResourceProvider",
+    "AWS::KinesisFirehose::DeliveryStream": "ResourceProvider"
```
simonrw, line 64: super-minor: it might be worth adding a trailing comma to the last item in the list so there is only a 1-line change per PR

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
```diff
@@ -0,0 +1,492 @@
...
+          - firehose:DescribeDeliveryStream
+          - iam:GetRole
+          - iam:PassRole
+          - kms:CreateGrant
+          - kms:DescribeKey
+
+        """
+        model = request.desired_state
+        firehose = request.aws_client_factory.firehose
+        parameters = [
```
simonrw, line 386: The method `add_defaults` in the `FirehoseDeliveryStream` adds a default value for `DeliveryStreamName` if it's not supplied. Could you add this?

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
```diff
@@ -0,0 +1,492 @@
...
+            "Tags",
+        ]
+        attrs = util.select_attributes(model, params=parameters)
+
+        if not request.custom_context.get(REPEATED_INVOCATION):
+            response = firehose.create_delivery_stream(**attrs)
+            # TODO: defaults
+            # TODO: idempotency
+            model["Arn"] = response["DeliveryStreamARN"]
+            request.custom_context[REPEATED_INVOCATION] = True
```
simonrw, line 407: Nice use of the `REPEATED_INVOCATION`

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
```diff
@@ -0,0 +1,492 @@
...
+            "HttpEndpointDestinationConfiguration",
+            "KinesisStreamSourceConfiguration",
+            "RedshiftDestinationConfiguration",
+            "SplunkDestinationConfiguration",
+            "Tags",
+        ]
+        attrs = util.select_attributes(model, params=parameters)
+
+        if not request.custom_context.get(REPEATED_INVOCATION):
+            response = firehose.create_delivery_stream(**attrs)
```
simonrw, line 403: What if the delivery stream already exists? We should check that it does not first, or else raise an error (assuming that's what AWS does)

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
```diff
@@ -0,0 +1,492 @@
...
+
+        IAM permissions required:
+          - firehose:DeleteDeliveryStream
+          - firehose:DescribeDeliveryStream
+          - kms:RevokeGrant
+          - kms:DescribeKey
+        """
+        model = request.desired_state
+        firehose = request.aws_client_factory.firehose
+        firehose.delete_delivery_stream(DeliveryStreamName=model["DeliveryStreamName"])
```
simonrw, line 456: Question: does the delivery stream on AWS take a while to delete? Should we use the `REPEATED_INVOCATION` pattern here as well?

simonrw (changes requested):
Looks good but a few things need changing

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
Morijarti: Sure thing. I'll add it.

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
Morijarti: It does take some time to delete and while in DELETING state it might still accept info. I'd have to measure it to see how long it take in AWS.

From boto3 documentation
```While the delivery stream is in the DELETING state, the service might continue to accept records, but it doesn’t make any guarantees with respect to delivering the data. Therefore, as a best practice, first stop any applications that are sending records before you delete a delivery stream.```
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/firehose/client/delete_delivery_stream.html

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
Morijarti: I'm assuming boto3 should cast exception here, causing the error if delivery stream exists.
I can test with actual CF on AWS and see how it handles usage of non-unique name

localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py
Morijarti: Done

simonrw (approved):
Thanks for making these changes, looks great!

PR #9263, opened 2023-09-29
Ensure unique addresses for GATEWAY_LISTEN

simonrw:
## Motivation

If `LOCALSTACK_HOST` is set to include port 443, and `GATEWAY_LISTEN` is not set, in -ext we by default add port 443 to the list, so `GATEWAY_LISTEN` becomes:

```
GATEWAY_HOST = [
    HostAndPort("0.0.0.0", 443),
    HostAndPort("0.0.0.0", 443),
]
```

Hypercorn then tries to bind to the port twice and crashes.

## Changes

* Introduce the concept of a `UniqueHostAndPortList` - see the docstring for guarantees
* Configure the dev container to modify the edge port config

### DIFF
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -234,6 +234,15 @@ def run(
         network=network,
     )
 
+    # replicate pro startup
+    if pro:
+        try:
+            from localstack_ext.plugins import modify_edge_port_config
+
+            modify_edge_port_config(config)
+        except ImportError:
+            pass
+
     # setup configurators
     configurators = [
         ImageConfigurator(pro, image),
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -571,7 +571,7 @@ class HostAndPort:
         return f"HostAndPort(host={self.host}, port={self.port})"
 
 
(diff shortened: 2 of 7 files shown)


localstack/dev/run/__main__.py
```diff
@@ -234,6 +234,15 @@ def run(
...
 
+    # replicate pro startup
+    if pro:
+        try:
+            from localstack_ext.plugins import modify_edge_port_config
+
+            modify_edge_port_config(config)
+        except ImportError:
+            pass
+
```
dfangl, line 245: Why is this necessary? Shouldn't this take place in the container then anyway?

dfangl (approved):
Although I do not like that a "wrong" configuration like `GATEWAY_LISTEN=:4566,:4566` works now, it seems more user friendly this way :)

localstack/dev/run/__main__.py
simonrw: It is in a `configure_localstack_container` hook which is not run in this dev container configuration cc @thrau

dfangl: Ah, you are right, it is usually called as configure hook as well. thanks for the clarification!

Issue #9261, opened 2023-09-29
bug: message passes through without attribute if there is a filterPolicy in place for null value


somatczk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

It seems that it is possible to receive sns message in an sqs listener if there is a filterpolicy in place for null value in a message attribute, and a message is being sent without the message attribute.

### Expected Behavior

Don't recieve these type of messages

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

`sns subscribe --topic-arn "demoTopicArn"  --protocol sqs --notification-endpoint "demoSqsEndpoint" --attributes '{"RawMessageDelivery": "true"}'`
`sns set-subscription-attributes --subscription-arn demoSubsArn --attribute-name FilterPolicy --attribute-value '{"testForNull": [null,{"anything-but": ["notgood"]}]}'`
`sns publish --topic-arn demoTopicArn --message "testforrecieval"`
`sqs receive-message --queue-url  "demoUrl"`

### Environment

```markdown
- OS: macOS 13.6
- LocalStack: 2.2.0
```

PR #9264, opened 2023-09-29
fix SNS FilterPolicy with null value

bentsku:
## Motivation
As reported with #9261, we would not properly filter for the `null` value in a FilterPolicy. This is because we set the value to `None` if it didn't exist. We luckily have a field to indicated the presence of the field or not `field_exists`. 

## Changes
Add the check for the `field_exists` before comparing the value with the condition.
Added a snapshot tests for validating this use-case. 

_closes #9261_


thrau (approved):
:+1: LGTM

PR #9267, opened 2023-09-29
Separate resource provider plugis

simonrw:
## Motivation

The plugins should be separated out from the provider files. The provider files may import a lot of code and reduce startup time for the CloudFormation service.

## Changes

* Move plugins to their own file
* Update the scaffolding to render the new plugin file


pinzon (approved):
LGTM

PR #9268, opened 2023-09-29
add error handling for Events.PutRule

thrau:
## Motivation

While using event bus, I noticed that specifying invalid rule expressions did not create the correct error response.

There was also a TODO related to us allowing `rate(1 minutes)` (value 1 with plural unit) which is for some reason not allowed in AWS. This may have some implications on user code.

## Changes

* Added proper handling and an AWS validated test
* Added [weird plural/singular handling of rate units](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rate-expressions.html)
  * Fixed all occurrences of invalid rates
* drive-by validated a few other tests


thrau:
going to sit tight on merging this until 2.3.2 is out

PR #9269, opened 2023-09-29
add test for AWS Mythical Mysfits sample app

bentsku:
## Motivation
Based on https://github.com/aws-samples/aws-modern-application-workshop/blob/python-cdk

This is only Module 5 (https://github.com/aws-samples/aws-modern-application-workshop/blob/python-cdk/module-5) with the DynamoDB table from Module 3, which is normally accessed through a Fargate micro service. 

architecture-module-5

This sample/scenario is adapted from TypeScript CDK to Python CDK, still seems a bit weird, lots of manual setup with IAM/Roles which I believe could use "CDK magic" more.

There's only one Cfn construct used, which could maybe be replace with an alpha L2 (was using the cfn one in the sample) (Kinesis Firehose Delivery Stream). 

## Changes
Added the scenario and tests, this currently only works against AWS.


tests/aws/scenario/mythical_mysfits/test_mythical_misfits.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+
+    @pytest.fixture(scope="class", autouse=True)
+    def infrastructure(self, aws_client, infrastructure_setup):
+        # set up the bucket name so that we can clean it up
+        bucket_name = f"test-mythical-mysfits-{short_uid()}"
+        infra = infrastructure_setup(namespace="MythicalMisfits")
+        MythicalMysfitsCoreStack(infra.cdk_app, STACK_NAME, bucket_name=bucket_name)
+
+        infra.add_custom_teardown(lambda: cleanup_s3_bucket(aws_client.s3, bucket_name=bucket_name))
```
steffyP, line 42: this should actually not be required, as the bucket is already created within the stack. We added a feature in the provisioner that automatically cleans up buckets.

tests/aws/scenario/mythical_mysfits/test_mythical_misfits.snapshot.json
```diff
@@ -0,0 +1,683 @@
...
+    "recorded-date": "12-10-2023, 16:56:10",
+    "recorded-content": {
+      "outputs": {
+        "ClicksBucketDestinationName": "<clicks-bucket-destination-name:1>",
+        "DeliveryStreamArn": "arn:aws:firehose:<region>:111111111111:deliverystream/<delivery-stream-name:1>",
+        "DeliveryStreamName": "<delivery-stream-name:1>",
+        "<physical-resource-id:1>Name": "<physical-resource-id:1>",
+        "PopulateDbFunctionName": "<populate-db-function-name:1>",
+        "StreamProcessorFunctionName": "<stream-processor-function-name:1>",
+        "UserClicksServiceAPIEndpoint": "https://<user-clicks-service-a-p-i-id:1>.execute-api.<region>.amazonaws.com/<physical-resource-id:7>/",
```
steffyP, line 12: nit: once we run the test against LS, we might need to re-work the pattern here

steffyP (approved):
Nice work, @bentsku 👏 🚀 

hope we can find some time to fix the revealed parity issues in LS soon, so that the test can actually run 😄 

I added a few comments/NITs for future improvements, nothing urgent.

tests/aws/scenario/mythical_mysfits/artefacts/functions/populate_db.py
```diff
@@ -0,0 +1,260 @@
+# Source adapted from: https://github.com/aws-samples/aws-modern-application-workshop/blob/python-cdk
+from __future__ import print_function
+
+import os
+
+import boto3
+
+client = boto3.client("dynamodb")
+
+init_data = [
```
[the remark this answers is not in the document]
bentsku: Ah, true! Didn't think of it!

tests/aws/scenario/mythical_mysfits/test_mythical_misfits.py
bentsku: Ah right, that's better. I think this was a leftover from when the bucket was versioned, but the utility didn't clean it up anyway. Will remove 👍

tests/aws/scenario/mythical_mysfits/stacks/mysfits_core_stack.py
```diff
@@ -0,0 +1,103 @@
+import os
+
+import aws_cdk as cdk
+import constructs
+
+from localstack.utils.files import load_file
+from tests.aws.scenario.mythical_mysfits.constructs.user_clicks_service import UserClicksService
+
+
+class MythicalMysfitsCoreStack(cdk.Stack):
```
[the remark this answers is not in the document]
bentsku: shameless copy from the serverlesspresso :D

tests/aws/scenario/mythical_mysfits/test_mythical_misfits.snapshot.json
steffyP: yes, transformer would be nice. but we can iterate over this as well. There might be other cases we haven't considered yet, we will see once we make it work with LS

>  I think CFn might return the right endpoint for LocalStack 

i think so too :)

PR #9270, opened 2023-10-01
avoid installation of python3.9 in docker image

whummer:
## Motivation

In the process of resolving some CVEs, we detected that the Docker image contains `python3.9`, which gets pulled in transitively from the `nodejs` package:
```
$ docker run -it --rm --entrypoint= localstack/localstack which python3.9
/usr/bin/python3.9
```

We should try to eliminate this 3.9 Python installation, to resolve CVEs, and also to slightly reduce the image size.

## Changes

Update Dockerfile to skip installation of `python3.9` packages. We achieve this by downloading the `*.deb` package for nodejs, then extracting the `control.tar.xz` file from it, and updating the dependencies in the `control` file to remove the `python3` dependency. 

Below is the content of the `control` file, prior to replacing its content via `sed` (`python3` is then removed from the `Depends` list): 
```
Package: nodejs
Version: 18.18.0-1nodesource1
...
Depends: libc6 (>= 2.28), python3, ca-certificates. # <- this line gets updated
Conflicts: nodejs-dev, nodejs-doc, nodejs-legacy, npm
```

This is a pretty crude approach - not sure if there's an easier way to achieve this 🤷, but seems that most other approaches (incl. any `--ignore*` or `--exclude*` options for `apt` and `dpkg` ) result in a broken package state. (I've also tried things like `dpkg -i --ignore-depends=python3 nodejs*.deb`, but that was leading to issues further down the line)

The PR also contains minor refactoring to rename `apt-get` to the slightly shorter `apt` command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,9 @@ FROM python:3.11.5-slim-buster@sha256:9f35f3a6420693c209c11bba63dcf103d88e47ebe0
 ARG TARGETARCH
 
 # install OpenJDK 11
-RUN apt-get update && \
-        apt-get install -y openjdk-11-jdk-headless && \
-        apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN apt update && \
+        apt install -y openjdk-11-jdk-headless && \
+        apt clean && rm -rf /var/lib/apt/lists/*
 
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-${TARGETARCH}
 
@@ -42,9 +42,9 @@ ARG TARGETARCH
 
 # Install runtime OS package dependencies
 RUN --mount=type=cache,target=/var/cache/apt \
-    apt-get update && \
+    apt update && \
         # Install dependencies to add additional repos
-        apt-get install -y --no-install-recommends \
+        apt install -y --no-install-recommends \
             # Runtime packages (groff-base is necessary for AWS CLI help)
             ca-certificates curl gnupg git make openssl tar pixz zip unzip groff-base iputils-ping nss-passwords procps iproute2
 
@@ -53,9 +53,15 @@ RUN --mount=type=cache,target=/var/cache/apt \
     mkdir -p /etc/apt/keyrings && \
     curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
     echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main" > /etc/apt/sources.list.d/nodesource.list && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends \
-        nodejs
+    apt update && \
+    cd /tmp && \
+    apt download nodejs && \
+    dpkg -i --ignore-depends=python3.9 --ignore-depends=python3 nodejs*.deb && \
+    rm -rf /tmp/*.deb
+
+# assert that `node` is installed, but not `python3.9` (which is a default dependency for the nodejs package)
+RUN which node
+RUN test ! $(which python3.9)
 
 SHELL [ "/bin/bash", "-c" ]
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -42,20 +42,31 @@ ARG TARGETARCH
...
+    apt update && apt install -y binutils xz-utils && \
+    mkdir -p /tmp/nodejs.deb && cd /tmp/nodejs.deb && \
+    apt download nodejs && \
+    ar -xf nodejs*.deb && tar -xf control.tar.xz && \
+    sed -i 's/python3,//g' control && \
+    tar cfJ control.tar.xz control && ar r nodejs*.deb control.tar.xz && \
+    dpkg -i nodejs*.deb && \
+    rm -rf /tmp/*.deb && apt remove --purge -y binutils xz-utils
+
+# assert that `node` is installed, but not `python3.9` (which is a default dependency for the nodejs package)
```
alexrashed, line 67: Maybe we could add these two checks to the previous step / layer?

Dockerfile
```diff
@@ -42,20 +42,31 @@ ARG TARGETARCH
 
 # Install runtime OS package dependencies
 RUN --mount=type=cache,target=/var/cache/apt \
-    apt-get update && \
+    apt update && \
```
alexrashed, line 45: Actually, apt does not have a stable API, which is why we used `apt-get` here.

Dockerfile
```diff
@@ -42,20 +42,31 @@ ARG TARGETARCH
...
-    apt-get install -y --no-install-recommends \
-        nodejs
+    apt update && apt install -y binutils xz-utils && \
+    mkdir -p /tmp/nodejs.deb && cd /tmp/nodejs.deb && \
+    apt download nodejs && \
+    ar -xf nodejs*.deb && tar -xf control.tar.xz && \
+    sed -i 's/python3,//g' control && \
+    tar cfJ control.tar.xz control && ar r nodejs*.deb control.tar.xz && \
+    dpkg -i nodejs*.deb && \
+    rm -rf /tmp/*.deb && apt remove --purge -y binutils xz-utils
```
alexrashed, line 65: This is a pretty sophisticated solution. But maybe it would make sense to directly jump to a installation where we don't automatically install all dependencies? Maybe we could find an archived distribution which we can just extract and link?

Dockerfile
```diff
@@ -119,9 +130,9 @@ ARG TARGETARCH
 
 # Install build dependencies to base
 RUN --mount=type=cache,target=/var/cache/apt \
-    apt-get update && \
+    apt update && \
         # Install dependencies to add additional repos
-        apt-get install -y gcc python-dev
+        apt install -y gcc
```
alexrashed, line 135: Good catch!

alexrashed (commented):
Wow, really nice drill down to get rid of the unnecessary Python 3.9 in the image!
I added some comments, your solution is already seems pretty good! I'm just thinking that it might make sense to check if there's a simpler solution (like a archived distribution which can be extracted and linked).

Dockerfile
alexrashed: We could for example just jump to the binary archive installation method: https://github.com/nodejs/help/wiki/Installation
Something like this:
```
# get the latest 18.x version
NODE_ARCH=$([[ "${TARGETARCH}" == "amd64" ]] && echo "x64" || echo "${TARGETARCH}");  \
LATEST_VERSION_FILENAME=`curl -s https://nodejs.org/dist/latest-v18.x/SHASUMS256.txt | grep -o "node-v.*-linux-${NODE_ARCH}"`; \
mkdir -p /usr/local/lib/nodejs; \
curl -s -L https://nodejs.org/dist/latest-v18.x/${LATEST_VERSION_FILENAME}.tar.gz | tar -xz -C /usr/local/lib/nodejs; \
...
```
But I guess this is really a question of taste?

Dockerfile
whummer: Gotcha, thought that `apt` was the standard these days. Will revert :+1:

Dockerfile
whummer: Also looks good to me, thanks for digging more into it. Probably more stable to fetch the dist binary directly :+1:

dfangl (approved):

PR #9274, opened 2023-10-02
Add develop mode to dev script

simonrw:
## Motivation

It is sometimes nice to attach a remote debugger to a running container. For this we can use the `DEVELOP=1` flag with the CLI, which
* installs `debugpy` into the container before infrastructure startup, and
* publishes port 5678 (the default [DAP](https://microsoft.github.io/debug-adapter-protocol/) port) to the host.

These two commands _can_ be added to the debug script, however it is marginally annoying. And of course we know that developers hate to be _mildy_ annoyed... 😂

## Changes

* Add a `--develop/--no-develop` flag to the CLI which performs the two above steps

## Testing

```
python -m localstack.dev.run --develop`
```

should allow for remote debugging via the DAP protocol. Specifically it should:
* install debugpy into the container, and
* publish port 5678

### DIFF
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -51,6 +51,12 @@ from .paths import HostPaths
     help="Whether to start localstack pro or community. If not set, it will guess from the current directory",
 )
 @click.option(
+    "--develop/--no-develop",
+    is_flag=True,
+    default=False,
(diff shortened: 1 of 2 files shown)


thrau (approved):
Cool! I would argue for the dev start script this should be the default

PR #9272, opened 2023-10-01
exclude werkzeug 3.0

thrau:
## Motivation

PR builds started failing yesterday. Example: https://app.circleci.com/pipelines/github/localstack/localstack/18636/workflows/bf96fc58-330d-4526-8720-fac54a29c45e

I had a hunch it may be Werkzeug again, since they released 3.0 the other day: https://github.com/pallets/werkzeug/releases/tag/3.0.0. I haven't had time to investigate yet, but assume it has something to do with deprecated code they removed.

## Changes

* Constrain werkzeug version to <3.0

PR #9276, opened 2023-10-02
upgrade werkzeug to version 3

alexrashed:
## Motivation
This PR upgrades Werkzeug to version 3 and fixes some issues caused by the removal of deprecations (or other breaking changes with version 3).
This PR removes the pin (avoiding the upgrade) added with https://github.com/localstack/localstack/pull/9272.

## Changes
- Updates minimum version in `setup.cfg`.
- Fixes some issues which came up with the new major release:
  - Decode headers when populating the WSGI request headers from the ASGI requests and when forwarding requests.
    - Werkzeug's `Headers` object raises a `BadRequestKeyError` if a key is not of type `str`.
    - See https://github.com/pallets/werkzeug/commit/5ff0a573f4b78d9724f1f063fb058fd6bc76b24d.
  - Removes the usage of `charset` and `encoding_errors` on the `Map` (they have been removed).
    - Also see https://github.com/pallets/werkzeug/commit/5ff0a573f4b78d9724f1f063fb058fd6bc76b24d (even though it's not mentioned in the commit message).

## Testing
As soon as the pipeline is green again (and the upgrade is tested with LocalStack Pro), this upgrade is safe to be merged.

## TODO
- [x] Make sure all tests in the pipeline are 💚 
  - I executed a full run (amd64, arm64, community tests) of our Pro pipeline. All 💚.

### DIFF
--- a/localstack/aws/forwarder.py
+++ b/localstack/aws/forwarder.py
@@ -221,7 +221,7 @@ def create_http_request(aws_request: AWSPreparedRequest) -> Request:
     # prepare the RequestContext
     headers = Headers()
     for k, v in aws_request.headers.items():
-        headers[k] = v
+        headers[k] = to_str(v, "latin-1")
 
     return Request(
         method=aws_request.method,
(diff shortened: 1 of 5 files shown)


dominikschubert (approved):
:heavy_check_mark:

thrau (approved):
LGTM!

just thinking that updating the constraints of core dependencies like this may affect extensions in unexpected ways. not sure what to do about it right now or how to deal with it moving forward, but something we should consider

PR #9286, opened 2023-10-04
Fix layer version handling for external Lambda layers

joe4dev:
## Motivation

Addresses a support request.

Fixes a bug where fetching external layers broke when using multiple versions of the same layer.

## Changes

Handle multiple versions of the same layer when fetching external layers.

## Testing

Added in companion ext PR: https://github.com/localstack/localstack-ext/pull/2163

## Discussion

The logic starts to get quite complex here. Some thoughts for the future:
* a) Using an internal service call with added cross-account support could simplify this logic here quite a bit. This requires IAM support, which was not present at the moment we implemented this initial version.
* b) The layer fetcher could return a layer version. However, this would break up the logic between community and ext.

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -588,10 +588,13 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
 
         visited_layers = dict()
         for layer_version_arn in new_layers:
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
LGTM, thanks for jumping on this

PR #9288, opened 2023-10-04
upgrade debian to bookworm

alexrashed:
## Motivation
This PR upgrades the OS version of the Docker image from Debian Bullseye to Debian Bookworm.
The last upgrade was done with the release of version `2.1.0` in https://github.com/localstack/localstack/pull/8344 / https://github.com/localstack/localstack/pull/7876.
Since the Debian repositories for bookworm do not contain `openjdk-11` anymore, I also switched the java-builder to use the `eclipse-temurin:11` image.

## Changes
- The base image's OS has been upgraded from Debian Bullseye (oldstable) to Debian Bookworm (stable).
- The `java-builder` stage has been replaced to use [Eclipse Temurin](https://hub.docker.com/_/eclipse-temurin) image (because Debian Bookworm does not have OpenJDK in the repos anymore).
  - We are already using Eclipse Temurin in the Java installer if we need other Java versions (for BigData services).

## Testing
- Tested by building the image locally.
- If all tests are green, we can be quite confident that everything works correctly.

## Merge?
This PR contains some important changes:
- Updates the base image from Debian Bullseye to Debian Bookworm.
  - This could have an implication for users which install any kind of packages in the Docker image.
- Change the Java JDK from OpenJDK in the package repo to the Temurin build.
  - I think this should actually be a safe change.
  - Eclipse Temurin is based OpenJDK.

We should decide on when to merge this change. The last time (with the Debian upgrade from Buster to Bullseye), we didn't really see any issues among our user-base.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,13 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    reviewers:
+      - "alexrashed"
     ignore:
       - dependency-name: "python"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
+      - dependency-name: "eclipse-temurin"
+        update-types: ["version-update:semver-major"]
     labels:
       - "area: dependencies"
       - "semver: patch"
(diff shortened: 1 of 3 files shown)


dfangl (approved):
Looks good!

PR #9289, opened 2023-10-04
Allow --mount-dependencies for python modules which are not in a separate directory

dfangl:
## Motivation
Currently, in the dev script, we only mount python modules which are in there own subfolder.
However, the assumption that all dependencies are located in a subfolder does not hold:
For example, `nest-asyncio` is only one file, and is not located in a separate folder in `site-packages`. (This is a dependency for serverless tests in -ext, and without it, test collection fails).

We should therefore also mount python files in `site-packages`, unless they start with `__` (sometimes linked editable installs have a python finder file like this: `__editable___localstack_core_2_2_1_dev0_finder.py`).

## Changes
* Mount files from `site-packages` as dependency as well, unless they start with a dunder

### DIFF
--- a/localstack/dev/run/configurators.py
+++ b/localstack/dev/run/configurators.py
@@ -294,8 +294,8 @@ class DependencyMountConfigurator:
 
         # find dependencies from the host
         for dep_path in self.host_paths.venv_dir.glob("lib/python3.*/site-packages/*"):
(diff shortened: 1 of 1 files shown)


thrau (approved):
nice find :ok_hand:

PR #9290, opened 2023-10-04
add /_localstack/info endpoint for basic infos about the instance

thrau:
## Motivation

Several other localstack systems need access to basic information about the running instance beyond the health status. Much like `docker info`, this new endpoint aims to provide this information.

Example:

```json
{
  "version": "2.3.3.dev:c77b830",
  "edition": "pro",
  "is_license_activated": true,
  "session_id": "1830110c-6e4d-4256-9bc0-9008b8c5b6b9",
  "machine_id": "4256a311839b",
  "system": "linux",
  "is_docker": true,
  "server_time_utc": "2023-10-04T14:59:40",
  "uptime": 120
}
```

## Changes

* Add a `/_localstack/info` endpoint
* Add two new metadata methods: getting the edition (from files we create in the Dockerfiles), and checking whether license activation has been successful
* Add the `edition` key also to the health endpoint, since that has long since been requested
* Add the output of info also to the diagnose endpoint

## Testing

* You can use the dev run script and call `curl -s https://localhost.localstack.cloud:4566/_localstack/info | jq .`

### DIFF
--- a/localstack/utils/analytics/metadata.py
+++ b/localstack/utils/analytics/metadata.py
@@ -118,10 +118,16 @@ def get_localstack_edition() -> Optional[Literal["enterprise", "pro", "community
 
 def is_license_activated() -> bool:
     try:
+        from localstack_ext import config  # noqa
+    except ImportError:
+        return False
+
+    try:
         from localstack_ext.bootstrap import licensingv2
 
         return licensingv2.get_licensed_environment().activated
-    except ImportError:
+    except Exception:
+        LOG.exception("Could not determine license activation status")
         return False
 
 
--- a/localstack/utils/analytics/metadata.py
+++ b/localstack/utils/analytics/metadata.py
@@ -2,11 +2,12 @@ import dataclasses
 import logging
 import os
 import platform
-from typing import Optional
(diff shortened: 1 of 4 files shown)


localstack/utils/analytics/metadata.py
```diff
@@ -94,6 +105,26 @@ def get_machine_id() -> str:
...
+    elif os.path.exists("/usr/lib/localstack/.pro-version"):
+        return "pro"
+    elif os.path.exists("/usr/lib/localstack/.community-version"):
+        return "community"
+
+    return None
+
+
+def is_license_activated() -> bool:
+    try:
```
alexrashed, line 120: nit: Maybe it would make sense to separate the imports here:
- First try to import anything super-stable from ext.
  - return false if it's not possible
- Then try to import the specific method.
  - list a prominent error in this case, because that would mean that this code stopped working because of a refactoring, not because the package is missing

alexrashed (approved):
Nice! This is going to be super-useful when trying to help our users! 🚀

lukqw:
This is great! Thanks for the addition 🙌

PR #9295, opened 2023-10-05
Implementation of AWS::EC2:* resource providers for CloudFormation

Morijarti:
## Changes
Moved all Ec2 resources from GenericBaseModel to resource provider for CloudFormation
Resources migrated are:
 - `AWS::EC2::RouteTable`
 - `AWS::EC2::Route`
 - `AWS::EC2::InternetGateway`
 - `AWS::EC2::SubnetRouteTableAssociation`
 - `AWS::EC2::VPCGatewayAttachment`
 - `AWS::EC2::SecurityGroup`
 - `AWS::EC2::Subnet`
- `AWS::EC2::VPC`
- `AWS::EC2::NatGateway`
 - `AWS::EC2::Instance`


pinzon (approved):
Awesome!

localstack/services/ec2/resource_providers/aws_ec2_route.py
```diff
@@ -0,0 +1,132 @@
...
+            DestinationCidrBlock=cidr_block,
+            DestinationIpv6CidrBlock=ipv6_cidr_block,
+            RouteTableId=model["RouteTableId"],
+        )
+
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model=model,
+            custom_context=request.custom_context,
+        )
```
dominikschubert, line 80: A `model["Id"] = ...` is missing here

Morijarti: Fixed

localstack/services/ec2/resource_providers/aws_ec2_instance.py
```diff
@@ -0,0 +1,304 @@
...
+        """
+        Delete a resource
+
+
+        """
+        model = request.desired_state
+        ec2 = request.aws_client_factory.ec2
+        ec2.terminate_instances(InstanceIds=[model["Id"]])
+        # TODO add checking of ec2 instance state
+        return ProgressEvent(
```
simonrw, line 289: Normally this would be an async operation (i.e. requires use of the `REPEATED_INVOCATION` key) but in community this operation is in-memory only and does not create a VM, so it's fine.

localstack/services/ec2/resource_providers/aws_ec2_internetgateway.py
```diff
@@ -0,0 +1,119 @@
...
+        Delete a resource
+
+        IAM permissions required:
+          - ec2:DeleteInternetGateway
+        """
+        model = request.desired_state
+        ec2 = request.aws_client_factory.ec2
+
+        # detach it first before deleting it
+        response = ec2.describe_internet_gateways(InternetGatewayIds=[model["InternetGatewayId"]])
```
simonrw, line 92: Interestingly the handler is only given permission to `ec2:DeleteInternetGateway` so it cannot perform [`describe_internet_gateways`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInternetGateways.html)!

localstack/services/ec2/resource_providers/aws_ec2_internetgateway.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if len(response.get("InternetGateways", [])) > 0:
+            gateway = response["InternetGateways"][0]
+
+            for attachment in gateway.get("Attachments", []):
+                ec2.detach_internet_gateway(
+                    InternetGatewayId=model["InternetGatewayId"], VpcId=attachment["VpcId"]
+                )
```
simonrw, line 100: Not sure I follow but is it a 0 or 1 situation? Can there be more internet gateways? If so:

```suggestion
        for gateway in response.get("InternetGateways", []):
            for attachment in gateway.get("Attachments", []):
                ec2.detach_internet_gateway(
                    InternetGatewayId=model["InternetGatewayId"], VpcId=attachment["VpcId"]
                )
```

localstack/services/ec2/resource_providers/aws_ec2_natgateway.py
```diff
@@ -0,0 +1,167 @@
...
+
+        IAM permissions required:
+          - ec2:DeleteNatGateway
+          - ec2:DescribeNatGateways
+        """
+        model = request.desired_state
+        ec2 = request.aws_client_factory.ec2
+        """
+        boto3 documentation only has method for deleting single nat gateway
+        instead of the one used bellow
```
simonrw, line 140:
```suggestion
        instead of the one used below
```

localstack/services/ec2/resource_providers/aws_ec2_natgateway.py
```diff
@@ -0,0 +1,167 @@
...
+        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/delete_nat_gateway.html
+        """
+        ec2.describe_nat_gateways(NatGatewayIds=[model["NatGatewayId"]])
```
simonrw, line 143: incomplete? The gateway is not actually deleted

localstack/services/ec2/resource_providers/aws_ec2_securitygroup.py
```diff
@@ -0,0 +1,149 @@
...
+        if not params.get("GroupName"):
+            params["GroupName"] = util.generate_default_name(
+                request.stack_name, request.logical_resource_id
+            )
+        response = ec2.create_security_group(
+            VpcId=params["VpcID"],
+            GroupName=params["GroupName"],
+            Description=params.get("GroupDescription", ""),
+        )
+        model["GroupId"] = response["GroupId"]
```
simonrw, line 103: The primary identifier is `Id` so this needs to be added as well

simonrw (approved):
Looks really good. Just a point about deleting the nat gateway, the rest of my comments are fairly minor

localstack/services/ec2/resource_providers/aws_ec2_securitygroup.py
Morijarti: Fixed

localstack/services/ec2/resource_providers/aws_ec2_natgateway.py
Morijarti: fixed

localstack/services/ec2/resource_providers/aws_ec2_natgateway.py
Morijarti: Removed comment, as I rewrote function for deleting nat gateway

localstack/services/ec2/resource_providers/aws_ec2_internetgateway.py
Morijarti: Good catch!

Morijarti: I think we can rewrite this, after we make sure that CF properly detects dependencies.
Right now InternetGateway might be deleted first, so in order to not break things we should detach it from VPCs.

localstack/services/cloudformation/resource_provider.py
```diff
@@ -69,6 +69,18 @@
     # "AWS::ECR::Repository": "ResourceProvider",  # FIXME: add full -ext provider & override logic for -ext
     "AWS::KMS::Key": "ResourceProvider",
     "AWS::KMS::Alias": "ResourceProvider",
+    "AWS::EC2::RouteTable": "ResourceProvider",
+    "AWS::EC2::Route": "ResourceProvider",
+    "AWS::EC2::InternetGateway": "ResourceProvider",
+    "AWS::EC2::SubnetRouteTableAssociation": "ResourceProvider",
```
simonrw, line 75: Can't enable until we have sorted out shared models

simonrw (changes requested):
Blocking until we resolve shared community/pro models, but otherwise looks great!

PR #9297, opened 2023-10-05
fix S3 v3 HeadObject with checksum

bentsku:
## Motivation
As reported in https://github.com/localstack/localstack/issues/6659#issuecomment-1749013552, calling `HeadObject` on a an object with `ChecksumMode=ENABLED` will raise an exception. I totally missed it when copy/pasting the implementation from `GetObject`, because of a `#noqa` put there because of typing (dynamic key value inside the `*Response` object). 

## Changes
Fixed the line to use the object checksum instead of a non-existent variable.
I've also removed all the `noqa` from the provider because some type warnings are better than fully failing that this.


thrau (approved):
thanks for the super speedy fix @bentsku , and thanks for reporting @bdurrani!

PR #9298, opened 2023-10-05
rework eventbus scheduled rule execution

thrau:
- add default event for scheduled eventbridge rules
- rework eventbus scheduled rule execution

## Motivation

TODO

## Changes

* Rework JobScheduler using the Scheduler utility
* ... TODO


alexrashed:
@thrau This one seems a bit outdated, do you think we can close this one?

thrau:
yep, i was hoping this would eventually make it into the events code, but i don't see that happen right now, so will close!

PR #9309, opened 2023-10-06
Helper changes to support `eksctl` and CDK L2 EKS construct

simonrw:
## Motivation

We want to support `eksctl` in the -ext repo, but this requires some minor changes in utilities imported from community. Also there are some bugs that need fixing.

## Changes

- Enable `AWS::EKS::NodeGroup` resource provider
- Fix logging bug in CFn
- `deploy_cfn_template` deletes stacks in reverse order of creation
- Add `keys_to_upper` function

## TODO

- [ ] delete nodegroup service method for proper cleanup
- [ ] other nodegroup methods


dfangl (approved):
LGTM!

Issue #9306, opened 2023-10-06
bug: "exception while calling sns.GetTopicAttributes: not enough values to unpack" when trying to get attributes of topic that does not exist


kibertoad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If I try to send 

```ts
  const command = new GetTopicAttributesCommand({
    TopicArn: topicArn,
  })
```

where topicArn is for topic that does not exist, the following error is thrown:

```ts
{
  "name": "InternalErrorException",
  "$fault": "server",
  "$metadata": {
    "httpStatusCode": 500,
    "attempts": 3,
    "totalRetryDelay": 58
  },
  "Error": {
    "Code": "InternalError",
    "Message": "exception while calling sns.GetTopicAttributes: not enough values to unpack (expected 6, got 1)",
    "message": "exception while calling sns.GetTopicAttributes: not enough values to unpack (expected 6, got 1)"
  },
  "RequestId": "f94e94f3-f687-427a-b19b-8c32c07e7f17",
  "xmlns": "http://sns.amazonaws.com/doc/2010-03-31/"
}
```

### Expected Behavior

In LocalStack 2.1.0 following error was thrown:

```ts
{
  Code: AWS.SimpleQueueService.NonExistentQueue
}
```

This still doesn't look correct, as it should be about topic, not queue, but something close to that should be returned.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

docker-compose.yml -> see environment

Code:

```ts
export async function getTopicAttributes(
  snsClient: SNSClient,
  topicArn: string,
): Promise<Either<'not_found', QueueAttributesResult>> {
  const command = new GetTopicAttributesCommand({
    TopicArn: topicArn,
  })

  try {
    const response = await snsClient.send(command)
    return {
      result: {
        attributes: response.Attributes,
      },
    }
  } catch (err) {
    // @ts-ignore
    if (err.Code === 'AWS.SimpleQueueService.NonExistentQueue') {
      return {
        // @ts-ignore
        error: 'not_found',
      }
    }
    throw err
  }
}
```

### Environment

```markdown
- OS: Windows 11

- LocalStack:


services:
  localstack:
    image: localstack/localstack:2.3.2
    network_mode: bridge
    hostname: localstack
    ports:
      - '127.0.0.1:4566:4566' # LocalStack Gateway
      - '127.0.0.1:4510-4559:4510-4559' # external services port range
    environment:
[...]

PR #9310, opened 2023-10-06
fix GetTopicAttributes exception when arn is malformed

bentsku:
## Motivation
As reported in #9306, we would raise a 500 when the topic is not a properly formed ARN. This exception comes from moto, and should be handled there as well. In the meantime, we can fix it with our utils. 

## Changes
Moved up our method to get the topic, which will raise the proper exception if there's an issue before calling moto. Added a test with a malformed arn to validated the behaviour.


thrau (approved):
LGTM

Issue #9300, opened 2023-10-06
bug: CloudFormation Conditions not working with !Select 


Chad-Bailer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Cloudformation templates are failing in Conditions that use `!Select` function within the `!Equals`. 

```
Error: Failed to create changeset for the stack: select-test, An error occurred (InternalError) when calling the CreateChangeSet operation (reached max retries: 4): exception while calling cloudformation.CreateChangeSet: Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/chain.py", line 90, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 123, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 93, in __call__
    skeleton_response = self.skeleton.invoke(context)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(context, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 166, in dispatch_request
    result = handler(context, instance) or {}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/api/core.py", line 160, in operation_marker
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/cloudformation/provider.py", line 602, in create_change_set
    resolved_stack_conditions = resolve_stack_conditions(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #9311, opened 2023-10-06
Add support for fn::select in conditional

pinzon:
## Motivation
As shown in the issue #9300 CFn should be able to support using a `!Select` function in a Conditional. 

## Changes
The `resolve_condition` function now also supports a case for `Fn::Select`. According to the[ function documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-select.html), the select can also resolve references. 

## Testing
Added a simplet AWS validated test to confirm the expected behaviour.

### DIFF
--- a/localstack/services/cloudformation/engine/template_utils.py
+++ b/localstack/services/cloudformation/engine/template_utils.py
@@ -262,6 +262,21 @@ def resolve_condition(
                         ]
                     )
                     return result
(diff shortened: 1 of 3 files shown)


dominikschubert (approved):
LGTM :+1:

PR #9315, opened 2023-10-09
Disable AWS::ECR::Repository mock resource provider

dominikschubert:
## Motivation

The current code isn't set up to handle this case where -ext overwrites a resource provider, especially with the -ext provider being a GenericBaseModel. We'll handle this later but for now we should unblock the pipeline again.

## Changes

- Disables the activation of the `AWS::ECR::Repository` resource provider, which should unblock -ext


alexrashed (approved):
Thanks for jumping on this to make sure the pipeline is unblocked again! 💯

simonrw (approved):

PR #9318, opened 2023-10-09
Use region-agnostic bucket create for Transcribe tests

viren-nadkarni:
## Motivation

This PR fixes test execution error when a non-default `TEST_AWS_REGION_NAME` is used.

This happens because S3 CreateBucket requires `LocationConstraint` information when any region other than `us-east-1` is used. Our `s3_create_bucket` fixture takes care of such situations.

## Implementation

Just use the `s3_create_bucket` fixture to create the S3 bucket instead of the direct S3 CreateBucket.


ackdav (approved):
:+1: makes sense! thx!

PR #9319, opened 2023-10-09
fix SNS MessageBody filtering with `.`

bentsku:
## Motivation
This issue has been reported in our Community Slack.
The following SNS `FilterPolicy` with the `FilterPolicyScope` set `MessageBody` was not working like AWS:
`"FilterPolicy": "{\"detail.eventType\": [\"ride.requested\"]}"`
Note the `.` in the field name, which is very important.
The event that should pass and work this policy looks like this:
```json
{
  "version": "0",
  "id": "6a7e8feb-b491-4cf7-a9f1-bf3703467718",
  "detail-type": "ride.requested",
  "source": "lambda",
  "account": "111122223333",
  "time": "2017-12-22T18:43:48Z",
  "region": "us-west-2",
  "detail": {
    "eventType": "ride.requested",
    "data": "{}"
  }
}
```

After checking with AWS, it seems that indeed, that event should work, even if it seems a bit counter-intuitive.
My first idea was to use JSONPath, but what AWS is doing is every simpler.
With this policy: `{"object.nested.test": ["string.value"]}`
All these messages are working:
```json
{"object": {"nested": {"test": "string.value"}}}
{"object.nested.test": "string.value"}
{"object.nested": {"test": "string.value"}}
{"object": {"nested.test": "string.value"}}
```
JSONPath would not work here, because `.` indicates a child node. But for `{"object.nested.test": "string.value"}`, there are no child nodes, and JSONPath fails. 

So what AWS seems to be doing is flattening the filter policy to a single level, with depths separated with `.`.
They do the same to the payload, and then simply compare key per key, like they do with `MessageAttributes` scope.

This behaviour is not documented anywhere, and we can suspect it could change in the future. In the meantime, I've added this feature which also allows us to remove the previous matching function which was a bit complicated. 

## Changes

Added a new test testing this behaviour, I've done more manual testing with all kind of scenarios but the test was really long, so I've kept it to a minimal which validates the behaviour.

Updated the filtering method to flatten the policy and payload and then compare them on a single level of depth, which is way simpler.
Maybe for performance reason we could cache the result of the `FilterPolicy` flattening, but it doesn't seem to expensive for now.

### DIFF
--- a/localstack/services/sns/publisher.py
+++ b/localstack/services/sns/publisher.py
@@ -934,55 +934,29 @@ class SubscriptionFilter:
 
     def _evaluate_nested_filter_policy_on_dict(self, filter_policy, payload: dict) -> bool:
         """
-        This method evaluate the filter policy recursively, while still being able to validate the `exists` condition.
+        This method evaluates the filter policy against the JSON decoded payload.
+        Although it's not documented anywhere, AWS allows `.` in the fields name in the filter policy and the payload,
+        and will evaluate them. However, it's not JSONPath compatible:
         Example:
-        nested_filter_policy = {
-            "object": {
-                "key": [{"prefix": "auto-"}],
-                "nested_key": [{"exists": False}],
-            },
-            "test": [{"exists": False}],
-        }
-        payload = {
-            "object": {
-                "key": "auto-test",
-            }
-        }
-        This function then iterates on top level keys of the filter policy: ("object", "test")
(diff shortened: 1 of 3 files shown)


thrau (approved):
"Although it's not documented anywhere, AWS allows `.` ..." classic :jack_o_lantern: 

nice find, and great tests, as always! :+1:

PR #9323, opened 2023-10-09
fix SNS headers when publishing to HTTP endpoints

bentsku:
## Motivation
This issue was reported in our Community Slack channel.

When trying to connect our SNS implementation to [Opsgenie](https://www.atlassian.com/software/opsgenie), a user encountered an 422 error.

After looking online, I fell on this post: https://community.atlassian.com/t5/Opsgenie-questions/Mimicking-CloudWatch-SNS-notification-in-Opsgenie/qaq-p/1886688, which indicates it should be possible.
It seems the headers are important for them to process our requests, so I've updated on our test to properly check that the header format was right.
We can also read about them here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-and-json-formats.html#http-header
The most important for this specific integration seems to be:
`Content-Type: text/plain; charset=UTF-8`
And we had `Content-Type: text/plain` only. 

## Changes
Update one test to verify the changes with snapshots.
Update the headers sent to be more in parity with AWS.
Also spotted a bug with the `RawDelivery` header which was in a never-reachable `elif` block, changed the indentation for it to work again, and this is now validated by the test as well.

### DIFF
--- a/localstack/services/sns/publisher.py
+++ b/localstack/services/sns/publisher.py
@@ -403,23 +403,24 @@ class HttpTopicPublisher(TopicPublisher):
         message_body = self.prepare_message(message_context, subscriber)
         try:
             message_headers = {
-                "Content-Type": "text/plain",
+                "Content-Type": "text/plain; charset=UTF-8",
+                "Accept-Encoding": "gzip,deflate",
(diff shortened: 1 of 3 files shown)


thrau (approved):
LGTM! nice catch

PR #9312, opened 2023-10-08
Minor: Enhance parity around CloudFormation template URLs

whummer:
## Motivation

During testing some sample apps, I've noticed that we're returning slightly different responses/errors when using external template URLs for CFn stacks.

This PR adds a snapshot test and asserts the correct error messages for different values of `TemplateURL`.

## Changes

* add new snapshot test, with parametrization for different values of `TemplateURL`
* adjust logic in `api_utils.py` to return the proper error message
* slightly refactor tests:
  * move `test_get_template_summary.py` to more general name `test_templates.py`
  * move stack policy related tests from `test_stacks.py` into more specialized `test_stack_policies.py` file

PR #9329, opened 2023-10-10
follow up on CloudFormation template URLs S3 URI

bentsku:
## Motivation
Follow up from #9312 and following up a Community Slack issue, it seems the new validation added was missing one exception handling when the S3 key would be containing slashes.

For example:
`awslocal cloudformation create-stack --stack-name vpc-us-east-1 --template-url s3://cloudformation-templates/aws-account-cfn/vpc.json`

Is raising:
```python
File "/opt/code/localstack/localstack/services/cloudformation/api_utils.py", line 55, in get_template_body
    result = client.get_object(Bucket=parts[0], Key=parts[2])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/botocore/client.py", line 535, in _api_call
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/botocore/client.py", line 980, in _make_api_call
botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the GetObject operation: The specified bucket does not exist
```

Because the split occurring in `convert_s3_to_local_url` was returning 3 parts so it worked.

If I understood well, `s3://`-type URIs are not accepted in `--template-url` parameter, is that right? So I believe if the scheme is `s3`, we could trigger the validation error straight away.

Weirdly, this is an accepted answer by AWS employees... https://repost.aws/questions/QUvDk684FkRW2wvFLybXoSuA/templateurl-must-be-a-supported-url
But testing shows that it's not supported?

Adding a link showing how the url should look:
https://aws.amazon.com/blogs/infrastructure-and-automation/best-practices-for-using-amazon-s3-endpoints-in-aws-cloudformation-templates/

\cc @whummer 

## Changes
Added a check for the scheme, and raises the exception straight if the scheme is `s3`. 
Added the test to have a long key with `/` in it to trigger the issue if not fixed.

### DIFF
--- a/localstack/services/cloudformation/api_utils.py
+++ b/localstack/services/cloudformation/api_utils.py
@@ -86,14 +86,14 @@ def convert_s3_to_local_url(url: str) -> str:
     headers = {"host": url_parsed.netloc}
     bucket_name, key_name = extract_bucket_name_and_key_from_headers_and_path(headers, path)
 
-    if not bucket_name or not key_name:
-        if url_parsed.scheme != "s3" and not (
-            url_parsed.netloc.startswith("s3.") or ".s3." in url_parsed.netloc
-        ):
-            raise ValidationError("TemplateURL must be a supported URL.")
+    if url_parsed.scheme == "s3":
         raise ValidationError(
             f"S3 error: Domain name specified in {url_parsed.netloc} is not a valid S3 domain"
         )
+
+    if not bucket_name or not key_name:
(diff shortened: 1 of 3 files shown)


simonrw (approved):
Looks like a worthwhile change, thanks for adding!

PR #9332, opened 2023-10-11
Improve CDK test tooling with better error reporting and typing

dominikschubert:
## Motivation

In preparation for the CDK testing workshop we noticed a few issues with the tooling that this PR now addresses.

## Changes

- AWS seems to have changes the `StatusReason` when creating a change set without changes recently. Since this seems extremely flaky, I've decided to remove this check and always assume on update that a failure can be ignored.
- Re-use stack event outputs for failed stack deployments from deploy_cfn_template fixture
- Fix typing of InfraProvisioner.provisioner to be compatible with pycharm => intellisense should work now

### DIFF
--- a/localstack/testing/scenario/provisioning.py
+++ b/localstack/testing/scenario/provisioning.py
@@ -10,6 +10,7 @@ from botocore.exceptions import WaiterError
 
 from localstack.config import is_env_true
 from localstack.testing.aws.util import is_aws_cloud
+from localstack.testing.pytest.fixtures import StackDeployError
 from localstack.utils.aws.resources import create_s3_bucket
 from localstack.utils.files import load_file
 from localstack.utils.functions import call_safe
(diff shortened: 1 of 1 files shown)


localstack/testing/scenario/provisioning.py
```diff
@@ -204,28 +205,32 @@ def provision(self, skip_deployment: Optional[bool] = False):
...
+                try:
+                    self.aws_client.cloudformation.get_waiter(
+                        "stack_update_complete" if is_update else "stack_create_complete"
+                    ).wait(
                         StackName=stack_id,
                         WaiterConfig=WAITER_CONFIG_AWS if is_aws_cloud() else WAITER_CONFIG_LS,
                     )
+
+                except WaiterError as e:
+                    raise StackDeployError(
```
steffyP, line 225: 🚀

steffyP (approved):
Nice 🚀 thanks for the improvements 🙏

PR #9333, opened 2023-10-11
CFn: allow for pro resource provider overrides

simonrw:
## Motivation

We disabled AWS::ECR::Repository in #9315 because we did not have an override system for resource providers in community vs pro. In particular, we had:

* a community `GenericBaseModel` subclass
* a (new) community `ResourceProvider`
* a pro `GenericBaseModel` subclass

In this case, the community resource provider was being chosen, rather than the pro `GenericBaseModel` subclass.

## Changes

Most of the changes to allow us to revert #9315 have been made in the pro repository. In addition, we are trying to handle the case where we have pro and community resource providers.

* Prioritise pro resource providers if present
* Update scaffolding to support scaffolding into the pro repo (partially, only the plugin for now)
* Fix scaffolding issues with unused imports

## TODO

- [ ] allow for scaffolding into the pro repo

### DIFF
--- a/localstack/services/cloudformation/resource_provider.py
+++ b/localstack/services/cloudformation/resource_provider.py
@@ -29,7 +29,7 @@ from localstack.services.cloudformation.service_models import KEY_RESOURCE_STATE
 
 PRO_RESOURCE_PROVIDERS = False
 try:
-    from localstack_ext.services.cloudformation.cloudformation_extended import (
+    from localstack_ext.services.cloudformation.resource_provider import (
         CloudFormationResourceProviderPluginExt,
     )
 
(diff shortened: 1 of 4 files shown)


pinzon (approved):
Awesome!!

localstack/services/cloudformation/resource_provider.py
```diff
@@ -754,6 +764,20 @@ def load_resource_provider(self, resource_type: str) -> ResourceProvider:
...
+            try:
+                plugin = pro_plugin_manager.load(resource_type)
+                return plugin.factory()
+            except Exception:
+                LOG.debug(
+                    (
+                        "Failed to load resource type %s from pro as a ResourceProvider. "
+                        "Maybe this resource is implemented in community."
+                    ),
+                    resource_type,
```
dominikschubert, line 778: I feel like this is a bit too noisy for community usage.

dominikschubert (approved):
I'm not 100% sure I like the approach with two different plugin managers, since it feels like we shouldn't need it. 

An alternative could be appending a certain suffix the the plugin name when coming from ext and then handling that logic in the community plugin manager directly.


For now I don't really want to block this too much, especially since we can re-evaluate this setup at a later stage without any user-facing changes. So feel free to merge as is!

PR #9336, opened 2023-10-11
add SNS internal endpoint to fetch SubscriptionTokens

bentsku:
## Motivation
This PR will add a new internal endpoint to be able to fetch subscription tokens used to confirm a subscription.

In case of email and most importantly HTTP(S) subscriptions, a special message is sent to the subscriber with a link to confirm the subscription so that it will receive the messages afterwards.

However, when working with external integrations (like Opsgenie: https://support.atlassian.com/opsgenie/docs/integrate-opsgenie-with-incoming-amazon-sns/), the link sent will most probably be a link to your own localhost (you could always use `HOSTNAME_EXTERNAL` and some way to link it to your own computer, but it's not a given the external integration will go on a link which is not AWS related).

Instead, we could expose the sent token with the internal endpoint, so that you can manually confirm the subscription with [`ConfirmSubscription`](https://docs.aws.amazon.com/sns/latest/api/API_ConfirmSubscription.html).

You could fetch the token with the subscriptionArn and manually confirm it this way:
```bash
$ curl "http://localhost:4566/_aws/sns/subscription-tokens?subscriptionArn=arn:aws:sns:us-east-1:000000000000:test-opsgenie:654c8659-2919-4073-9dfd-3b766dbd532d"
{"subscription_tokens": {"arn:aws:sns:us-east-1:000000000000:test-opsgenie:654c8659-2919-4073-9dfd-3b766dbd532d": "75732d656173742d312f345a7e4e345a7e4e345a7e4e345a7e4e345a7e4e345a"}, "region": "us-east-1"}%

$ awslocal sns confirm-subscription --topic-arn arn:aws:sns:us-east-1:000000000000:test-opsgenie --token 75732d656173742d312f345a7e4e345a7e4e345a7e4e345a7e4e345a7e4e345a
{
    "SubscriptionArn": "arn:aws:sns:us-east-1:000000000000:test-opsgenie:654c8659-2919-4073-9dfd-3b766dbd532d"
}
```

This would then enable the subscription and allow you to use the external integration. 

I've done the test with Opsgenie which of course refuse the `SubscriptionConfirmation` message as it points to `localhost:4566`. 
I manually confirm the subscription, and after sending a message I can see the alert created. 

```python
sns_client: SNSClient = session.client("sns", endpoint_url="http://localhost:4566")
sns_client.publish(
    TopicArn="arn:aws:sns:us-east-1:000000000000:test-opsgenie",
    Message="This is the description of Sns Integration test alert",
    Subject="Sns Integration Test alert",
    MessageAttributes={"eventType": {"DataType": "String", "StringValue": "create"}},
)
```
Which gives:

Screenshot 2023-10-12 at 01 35 18

Which is pretty cool. 

## Changes
This adds an internal resource exposing the tokens. Those tokens are never deleted because they can be reused later on, so the filter is nicely useful.
A doc entry will need to be made in a follow up (will add the SMS endpoint documentation as well).

Also update a test to validate the headers sent with the `SubscriptionConfirmation` message just to be sure.

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -1051,35 +1051,43 @@ class SNSServiceSubscriptionTokenApiResource:
 
     This is registered as a LocalStack internal HTTP resource.
 
-    This endpoint accepts:
-    - GET param `accountId`: selector for AWS account. If not specified, return fallback `000000000000` test ID
-    - GET param `region`: selector for AWS `region`. If not specified, return default "us-east-1"
-    - GET param `subscriptionArn`: filter for `subscriptionArn` resource in SNS
+    This endpoint has the following parameter:
+    - GET `subscription_arn`: `subscriptionArn`resource in SNS for which you want the SubscriptionToken
     """
 
-    @route(sns_constants.SUBSCRIPTION_TOKENS_ENDPOINT, methods=["GET"])
-    def on_get(self, request: Request):
-        account_id = request.args.get("accountId", DEFAULT_AWS_ACCOUNT_ID)
-        region = request.args.get("region", AWS_REGION_US_EAST_1)
-        subscription_arn = request.args.get("subscriptionArn")
-        store: SnsStore = sns_stores[account_id][region]
-        if subscription_arn:
-            subscription_tokens = {}
-            for token, sub_arn in store.subscription_tokens.items():
-                if sub_arn == subscription_arn:
-                    subscription_tokens[subscription_arn] = token
-                    break
+    @route(f"{sns_constants.SUBSCRIPTION_TOKENS_ENDPOINT}/<path:subscription_arn>", methods=["GET"])
+    def on_get(self, _request: Request, subscription_arn: str):
+        try:
+            parsed_arn = parse_arn(subscription_arn)
+        except InvalidArnException:
+            response = Response("", 400)
+            response.set_json(
+                {
+                    "error": "The provided SubscriptionARN is invalid",
+                    "subscription_arn": subscription_arn,
+                }
+            )
+            return response
 
-            return {
-                "subscription_tokens": subscription_tokens,
-                "region": region,
-            }
+        store: SnsStore = sns_stores[parsed_arn["account"]][parsed_arn["region"]]
 
-        subscription_tokens = {
-            subscription_arn: token for token, subscription_arn in store.subscription_tokens.items()
-        }
+        found_token = None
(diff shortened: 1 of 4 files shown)


localstack/services/sns/provider.py
```diff
@@ -1041,3 +1042,44 @@ def on_delete(self, request: Request) -> Response:
...
+        if subscription_arn:
+            subscription_tokens = {}
+            for token, sub_arn in store.subscription_tokens.items():
+                if sub_arn == subscription_arn:
+                    subscription_tokens[subscription_arn] = token
+                    break
+
+            return {
+                "subscription_tokens": subscription_tokens,
+                "region": region,
```
thrau, line 1075: why return the region but not the account?

thrau (approved):
Looks like a cool new addition! Just a couple of minor comments. Let's also add a TODO for some docs :-)

localstack/services/sns/provider.py
```diff
@@ -1041,3 +1042,44 @@ def on_delete(self, request: Request) -> Response:
...
+    - GET param `region`: selector for AWS `region`. If not specified, return default "us-east-1"
+    - GET param `subscriptionArn`: filter for `subscriptionArn` resource in SNS
+    """
+
+    @route(sns_constants.SUBSCRIPTION_TOKENS_ENDPOINT, methods=["GET"])
+    def on_get(self, request: Request):
+        account_id = request.args.get("accountId", DEFAULT_AWS_ACCOUNT_ID)
+        region = request.args.get("region", AWS_REGION_US_EAST_1)
+        subscription_arn = request.args.get("subscriptionArn")
+        store: SnsStore = sns_stores[account_id][region]
```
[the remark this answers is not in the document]
bentsku: You're right, I followed the same logic as the previous one, because the `subscriptionArn` is optional, but I'm not so sure it should be. Maybe we could make it part of the route and not have those as parameters at all.
You just wouldn't fetch all the possible tokens. I think that'd be simpler, just a bit different from the previous ones but it makes sense. Will do that actually, if that's ok?

localstack/services/sns/provider.py
bentsku: I think it's the same as above, just remnant of the previous ones, because we must have issues with multi region and we were not too big into multi account yet. Will refactor this, thanks a lot 😄

thrau (approved):
nice! much simpler and cleaner!

PR #9339, opened 2023-10-12
Dev script can load extensions

simonrw:
## Motivation

When using LocalStack extensions, the CLI can be used to automatically mount the code in the correct place in the container to run dev extensions. This removes the possibility of making changes to LocalStack internals (e.g. to prototype any changes/monkeypatches that are required).

## Changes

* Add a `--dev-extensions` flag which performs the same steps as the CLI when
    * `EXTENSION_DEV_MODE=1` is supplied
    * pro is enabled

### DIFF
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -7,11 +7,10 @@ from rich.rule import Rule
 
 from localstack import config
 from localstack.cli import console
-from localstack.utils.bootstrap import ContainerConfigurators
+from localstack.utils.bootstrap import Container, ContainerConfigurators
 from localstack.utils.container_utils.container_client import (
     ContainerConfiguration,
     PortMappings,
-    VolumeBind,
     VolumeMappings,
 )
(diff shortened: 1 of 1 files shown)


localstack/dev/run/__main__.py
```diff
@@ -281,6 +285,25 @@ def run(
...
+                from localstack_ext.extensions.bootstrap import (
+                    run_on_configure_host_hook,
+                    run_on_configure_localstack_container_hook,
+                )
+
+                # collect dev extensions
+                run_on_configure_host_hook()
+
+                # create stub container with configuration to apply
+                c = Container(container_config=container_config)
```
simonrw, line 300: I`m not pleased about wrapping the configuration in an empty `Container` but it re-uses the code which is likely more important.

thrau: no i think we could refactor the code to use take the config instead of the container :+1: but i'll leave it up to you if you want to do it or not :-)

thrau (approved):
Sweet! Thanks for adding this

PR #9342, opened 2023-10-12
[SFN] Add support for previousEventId

MEPalma:
## Motivation
When dispatching `GetExecutionHistory` calls about some State Machine Execution, the current implementation would compile `previousEventId` values according to a global incremental counter. This logic means that in information about the true event source of any event is often incorrect, especially for distributed workflows such as Map and Parallel states. This PR aims at ensuring that for each event, the true source event id is provided.

## Changes
Introduces a `EventHistoryContext` in `Environment` objects. These allow the interpreter to cache the last event id published. As each context object operates in its own evaluation scope, this is information is ensured to not be tampered by the global execution of the program. Values in the context can then be resynchronised with the parent scope during frame closing operations, or skipped through frame deletion operations (see MapRun workflows). 
Evaluation logic for map iteration workers is modified to fit this new design, and resolved a number of concurrency issues when closing pool jobs or evaluation stops.
Related changes extend to other relevant state resolvers such as `state_map` and `state_parallel`.

## Testing
- removes all `skip_snapshot_verify` for `previous_event_id`
- enables `test_sfn_scenarios` for v2


dominikschubert (approved):
LGTM, thanks for jumping on this. :rocket:

PR #9344, opened 2023-10-13
improve CORS handler: no CORS headers if no Origin

bentsku:
## Motivation
By default, we always add and send the CORS headers to LocalStack responses.
However, those headers should be sent only when the request has an `Origin` header set, indicating that this is a CORS request. 
We can read more about this here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#the_http_request_headers
https://jakearchibald.com/2021/cors/

> Note that in any access control request, the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) header is always sent.

When debugging and checking logs, we always have the CORS headers being sent, and they pollute the logs quite a lot for no benefits.

## Changes

Added a check for the presence of the `Origin` header, and add the headers only if it's there.
Added the `Vary` header as we conditionally apply CORS headers depending on it.
Added the credentials header.

### DIFF
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 
 from flask_cors.core import (
     ACL_ALLOW_HEADERS,
+    ACL_CREDENTIALS,
     ACL_EXPOSE_HEADERS,
     ACL_METHODS,
     ACL_ORIGIN,
@@ -181,6 +182,7 @@ class CorsEnforcer(Handler):
         elif context.request.method == "OPTIONS" and not config.DISABLE_PREFLIGHT_PROCESSING:
             # we want to return immediately here, but we do not want to omit our response chain for cors headers
             response.status_code = 204
+            print("stopping chain here?")
             chain.stop()
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


localstack/aws/handlers/cors.py
```diff
@@ -258,3 +266,6 @@ def add_cors_headers(request_headers: Headers, response_headers: Headers):
             and ACL_ALLOW_PRIVATE_NETWORK not in response_headers
         ):
             response_headers[ACL_ALLOW_PRIVATE_NETWORK] = "true"
+
+        # we conditionally apply CORS headers depending on the Origin, so add it to `Vary`
+        response_headers["Vary"] = "Origin"
```
dfangl, line 271: Good catch!

localstack/aws/handlers/cors.py
```diff
@@ -229,22 +230,29 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
...
         if ACL_ORIGIN not in response_headers:
             response_headers[ACL_ORIGIN] = (
-                request_headers["origin"]
-                if request_headers.get("origin") and not config.DISABLE_CORS_CHECKS
+                request_headers["Origin"]
+                if request_headers.get("Origin") and not config.DISABLE_CORS_CHECKS
                 else "*"
             )
+        if "*" not in response_headers.get(ACL_ORIGIN, ""):
+            response_headers[ACL_CREDENTIALS] = "true"
```
dfangl, line 255: Did we have any problems with that or is this just for completeness?

dfangl (approved):
This looks good!

localstack/aws/handlers/cors.py
bentsku: I think I've seen the issue pops sometimes in Chrome or Safari where it would say that the request was not allowed because of this header missing. I believe it can be good to add it, as the browser could block some requests in the frontend if we don't include it.

> When a request's credentials mode ([Request.credentials](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)) is include, browsers will only expose the response to the frontend JavaScript code if the Access-Control-Allow-Credentials value is true.

> If the CORS request includes credentials, the response must include the Access-Control-Allow-Credentials: true header, and the value of Access-Control-Allow-Origin must reflect the request's Origin header (* isn't an acceptable value if the request has credentials).

As we control the list of allowed domains, I believe this can be useful, maybe for cognito login page for example? 
We also could add a `CORS_ALLOW_CREDENTIALS` feature flag in the same way as `flask_cors`, to add this header only if it's set?

bentsku: @dfangl do you think we should do something like above, or just merge this PR?

PR #9357, opened 2023-10-16
add ability to peek delayed and invisible sqs messages

thrau:
## Motivation

A user [asked](https://discuss.localstack.cloud/t/a-way-to-see-invisible-sqs-messages/565) whether it's possible to see invisible messages through the SQS developer endpoint, and I thought this would be a really nice addition. I added delayed messages as well since it made sense.

If you now add the new query args to your request like this:
```python
response = requests.get(
    "http://localhost:4566/_aws/sqs/messages",
    params={"QueueUrl": queue_url, "ShowInvisible": True, "ShowDelayed": True},
    headers={"Accept": "application/json"},
)
```

The endpoint now returns messages like this:
```json
[
    {
        "MessageId": "1c4187cc-f2c9-4f1c-9702-4a3bfaaa4817",
        "MD5OfBody": "a06498de7fb4bd539c8895748f03175d",
        "Body": "message-3",
        "Attribute": [
            {"Name": "SenderId", "Value": "000000000000"},
            {"Name": "SentTimestamp", "Value": "1697494407799"},
            {"Name": "ApproximateReceiveCount", "Value": "0"},
            {"Name": "ApproximateFirstReceiveTimestamp", "Value": "0"},
            {"Name": "IsVisible", "Value": "true"},   <--
            {"Name": "IsDelayed", "Value": "false"},  <--
        ],
        "ReceiptHandle": "SQS/BACKDOOR/ACCESS",
    },
  ...
]
```

Using this will feel a bit clumsy, since you have to go through all the attributes of a message every time you want to find out whether they're visible or not, but I think it's a good compromise to get a very simple implementation.

## Changes

* The dev endpoint now allows two additional non-standard query arguments `ShowInvisible=true` and `ShowDelayed=true`
* If set, we also add messages from `queue.inflight` and `queue.delayed` into the list of messages to return
* If set, we add a non-standard message attribute `IsVisible` or `IsDelayed`, respectively

### DIFF
--- a/tests/aws/services/sqs/test_sqs_backdoor.py
+++ b/tests/aws/services/sqs/test_sqs_backdoor.py
@@ -21,6 +21,10 @@ def _parse_message_attributes(xml) -> list[dict]:
     ]
 
 
+def _parse_attribute_map(json_message: dict) -> dict[str, str]:
+    return {attr["Name"]: attr["Value"] for attr in json_message["Attribute"]}
+
+
 class TestSqsDeveloperEdpoints:
     @markers.aws.only_localstack
     def test_list_messages_has_no_side_effects(self, sqs_create_queue, aws_client):
@@ -41,7 +45,6 @@ class TestSqsDeveloperEdpoints:
         response = aws_client.sqs.receive_message(
             QueueUrl=queue_url, VisibilityTimeout=0, MaxNumberOfMessages=1, AttributeNames=["All"]
         )
-        print(response)
         assert response["Messages"][0]["Body"] == "message-1"
         assert response["Messages"][0]["Attributes"]["ApproximateReceiveCount"] == "1"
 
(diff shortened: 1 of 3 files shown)


tests/aws/services/sqs/test_sqs_backdoor.py
```diff
@@ -154,6 +157,78 @@ def test_list_messages_as_json(self, sqs_create_queue, aws_client):
...
+        response = requests.get(
+            "http://localhost:4566/_aws/sqs/messages",
+            params={"QueueUrl": queue_url, "ShowInvisible": False},
+            headers={"Accept": "application/json"},
+        )
+        doc = response.json()
+        messages = doc["ReceiveMessageResponse"]["ReceiveMessageResult"]["Message"]
+        assert len(messages) == 2
+        assert messages[0]["Body"] == "message-2"
+        assert messages[1]["Body"] == "message-3"
```
baermat, line 180: nit: Since we are not using a fifo queue, relying on the order of the messages could be dangerous. However, since this a synchronous @only_localstack test, we shouldn't really have any problems.
This holds for all code in the test that uses indices when accessing messages, but as I said, probably not really worth fretting over it.

baermat (approved):
Looks good to me, and makes sense to me! Small nits we can probably ignore since this is a ls-only feature

tests/aws/services/sqs/test_sqs_backdoor.py
thrau: great catch! as you said i think it's fine now, probably not worth running the suite again :-)

PR #9358, opened 2023-10-17
sns: add multi-account capability to failing tests

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `sns` tests should still create the consequent resources in this accounts and region.

## Changes
This PR fixes all the failing tests in `tests.aws.services.sns.test_sns.py` by:
- removing the use of `get_aws_account_id()` helper.
- use moto internals in `unsubscribe` handler. 
- use parsed arn instead of context arn 

See #8204

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -352,8 +352,24 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         return result
 
     def unsubscribe(self, context: RequestContext, subscription_arn: subscriptionARN) -> None:
-        call_moto(context)
-        store = self.get_store(account_id=context.account_id, region_name=context.region)
+        try:
+            parsed_arn = parse_arn(subscription_arn)
+        except InvalidArnException:
+            raise InvalidParameterException("Invalid parameter: SubscriptionId")
+
+        account_id = parsed_arn["account"]
+        region_name = parsed_arn["region"]
+
+        moto_sns_backend = self.get_moto_backend(parsed_arn["account"], parsed_arn["region"])
+        moto_sns_backend.unsubscribe(subscription_arn)
+
+        response = {
+            "UnsubscribeResponse": {
+                "ResponseMetadata": {"RequestId": context.request_id},
+            }
+        }
+
+        store = self.get_store(account_id=account_id, region_name=region_name)
 
         # pop the subscription at the end, to avoid race condition by iterating over the topic subscriptions
         subscription = store.subscriptions.get(subscription_arn)
@@ -383,6 +399,8 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         store.topic_subscriptions[subscription["TopicArn"]].remove(subscription_arn)
         store.subscription_filter_policy.pop(subscription_arn, None)
         store.subscriptions.pop(subscription_arn, None)
+
+        return response
 
     def get_subscription_attributes(
         self, context: RequestContext, subscription_arn: subscriptionARN
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -356,9 +356,11 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         try:
             parsed_arn = parse_arn(subscription_arn)
         except InvalidArnException:
-            # TODO: check for invalid SubscriptionId and SubscriptionGUID
+            # TODO: check for invalid SubscriptionGUID
             if count < 6:
-                raise InvalidParameterException(f"Invalid parameter: SubscriptionArn Reason: An ARN must have at least 6 elements, not {count}")
+                raise InvalidParameterException(
+                    f"Invalid parameter: SubscriptionArn Reason: An ARN must have at least 6 elements, not {count}"
+                )
 
         account_id = parsed_arn["account"]
         region_name = parsed_arn["region"]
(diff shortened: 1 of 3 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -3810,7 +3811,14 @@ def _clean_headers(response_headers: dict):
...
             snapshot.match("http-message-headers", _clean_headers(notification_request.headers))
 
-        unsub_request = requests.get(expected_unsubscribe_url)
+        auth_header = aws_stack.mock_aws_request_headers(
+            "sns", region_name=TEST_AWS_REGION_NAME, access_key=TEST_AWS_ACCESS_KEY_ID
+        )["Authorization"]
+
+        unsub_request = requests.get(
+            url=expected_unsubscribe_url, headers={"Authorization": auth_header}
+        )
```
bentsku, line 3820: I think this is wrong, and that the issue is in SNS `Unsubscribe`, because when working/validating the test in AWS this will fail because the credentials are wrong.

It seems `Unsubscribe` is an operation that can be done unauthenticated.
https://docs.aws.amazon.com/sns/latest/api/API_Unsubscribe.html

However, we have an issue here, is that requests always fall back to some credentials, so we won't be able to fully implement the behaviour for now that the documentation is explaining (namely the fact that SQS unsubscribe should always be authenticated, because for us, they always are so we can't reject an unauthenticated one).

But we can fix Unsubscribe, as it works the same way as `ConfirmSubscription`: 
```python
        # this request can come from any http client, it might not be signed (we would need to implement
        # `authenticate_on_unsubscribe` to force a signing client to do this request.
        # so, the region and account_id might not be in the request. Use the ones from the topic_arn
        try:
            parsed_arn = parse_arn(topic_arn)
        except InvalidArnException:
            raise InvalidParameterException("Invalid parameter: Topic")

        store = self.get_store(account_id=parsed_arn["account"], region_name=parsed_arn["region"])
```
This above should be also be in `Unsubscribe`, to correctly get the account/region for the store to then actually do the work.

bentsku (changes requested):
Thanks a lot for tackling this, this put a bug into light with `Unsubscribe`!
I explained the issue that needs fixing in the comment regarding modifying the test, I'm really sorry that it's a tricky test to validate. I don't think we'll need to change it anyway, the fix will live in `Unsubscribe`.

tests/aws/services/sns/test_sns.py
```diff
@@ -3810,7 +3811,14 @@ def _clean_headers(response_headers: dict):
...
+
```
viren-nadkarni, line 3821: For posterity, authentication params are required when calling API via HTTP <https://docs.aws.amazon.com/sns/latest/api/API_Unsubscribe.html#API_Unsubscribe_Examples>. `AUTHPARAMS` is a SigV4 signature that can be sent either via headers (as is done here) or as query params <https://docs.aws.amazon.com/IAM/latest/UserGuide/aws-signing-authentication-methods.html>

@sannya-singal could you please verify that use of `mock_aws_request_headers` does not break the test when running against AWS?

tests/aws/services/sns/test_sns.py
sannya-singal: Thanks for pointing that out @bentsku.

localstack/services/sns/provider.py
```diff
@@ -384,6 +400,8 @@ def unsubscribe(self, context: RequestContext, subscription_arn: subscriptionARN
         store.subscription_filter_policy.pop(subscription_arn, None)
         store.subscriptions.pop(subscription_arn, None)
 
+        return response
```
viren-nadkarni, line 403: There should be no return value as per the function signature.

viren-nadkarni (changes requested):

localstack/services/sns/provider.py
```diff
@@ -352,8 +352,18 @@ def create_platform_endpoint(
         return result
 
     def unsubscribe(self, context: RequestContext, subscription_arn: subscriptionARN) -> None:
-        call_moto(context)
-        store = self.get_store(account_id=context.account_id, region_name=context.region)
+        try:
+            parsed_arn = parse_arn(subscription_arn)
+        except InvalidArnException:
+            raise InvalidParameterException("Invalid parameter: SubscriptionId")
```
bentsku, line 358: can we have an AWS validated test for this exception? I don't think we have one. Something like `test_publish_wrong_arn_format`would be nice.

localstack/services/sns/provider.py
```diff
@@ -352,8 +352,18 @@ def create_platform_endpoint(
...
+
+        account_id = parsed_arn["account"]
+        region_name = parsed_arn["region"]
+
+        moto_sns_backend = self.get_moto_backend(account_id, region_name)
+        moto_sns_backend.unsubscribe(subscription_arn)
```
bentsku, line 364: It's nice that moto does not raise any kind of exception here, so it's pretty safe to use! Perfect 😄

bentsku (changes requested):
Thanks for changing and fixing `Unsubscribe`! It would be nice if we could have an AWS validated test for the exception regarding the ARN, I know we didn't before and we relied on moto, but maybe the exception wasn't right. Would be good to check now that it's our responsibility 😄 after that it'll be good for me!

bentsku (approved):
LGTM! Awesome, thanks a lot, sorry for the requested changes, but this looks awesome now, thanks a lot for fixing an SNS bug 🐛🚀!

PR #9361, opened 2023-10-17
Use account and region aware test assertions

viren-nadkarni:
## Motivation

LocalStack test suite uses the `TEST_AWS_ACCOUNT_ID` and `TEST_AWS_REGION_NAME` to control the namespace of the test execution. Some of our tests do not run the correct assertions when these config options are set to non-default values.

## Implementation

This PR makes some of the test assertions account/region aware. Specifically, ARN constructions now make use of the proper account/region that was used to create the original testing resource. Plus, there is correct use of client fixtures in places.

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -8,6 +8,7 @@ import pytest
 from boto3.dynamodb.types import STRING
 
 from localstack.aws.api.dynamodb import PointInTimeRecoverySpecification
+from localstack.constants import AWS_REGION_US_EAST_1, TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME
 from localstack.services.dynamodbstreams.dynamodbstreams_api import get_kinesis_stream_name
 from localstack.testing.aws.lambda_utils import _await_dynamodb_table_active
 from localstack.testing.aws.util import is_aws_cloud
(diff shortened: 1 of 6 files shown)


tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -1579,16 +1584,20 @@ def test_batch_write_not_existing_table(self, aws_client):
...
+        Test for AWS NoSQL Workbench, which sends "localhost" as region in header.
+        LocalStack must assume "us-east-1" region in such cases.
+        """
         table_name = f"t-{short_uid()}"
-        dynamodb_create_table(table_name=table_name, partition_key=PARTITION_KEY)
-        # describe table for default region
-        table = aws_client.dynamodb.describe_table(TableName=table_name)
+
+        # Create a table in the `us-east-1` region
+        client = aws_client_factory(region_name=AWS_REGION_US_EAST_1).dynamodb
```
viren-nadkarni, line 1595: `aws_client` fixture isn't used, instead a client with static region set to `us-east-1` is used since the test is critical to that. `aws_client` fixtures uses the value of env config `TEST_AWS_REGION_NAME`

dominikschubert (approved):
LGTM

Small nit would be that we probably shouldn't be using the ARN utils, but I guess this is rather part of the apigw test rework.

PR #9362, opened 2023-10-17
Lambda: Fix VPC resolution in non-default accounts

viren-nadkarni:
## Motivation

Lambda has trouble resolving VPCs during VPC config construction when requests are made from a non-default account ID.

## Implementation

This PR ensures that Lambda looks for VPCs in the same account/region as the requestor when building VPC config.


dfangl (approved):
Looks good, good catch!

PR #9363, opened 2023-10-17
EC2: Fix IndexError in ModifyLaunchTemplate

viren-nadkarni:
## Motivation

EC2 ModifyLaunchTemplate threw an error saying template not found for a template which could be listed. This only happened when a non-default account ID was used.

The cause was the use of a `connect_to()` field which did not take the current request account ID into consideration and instead looked for the template in the default `000000000000` account.

## Implementation

This PR removes the `connect_to()` call and constructs the response in the provider handler.


steffyP (approved):
The change LGTM 👍 Thanks for fixing!

PR #9370, opened 2023-10-17
fix awsRegion field in S3 notification

bentsku:
## Motivation
It had been shown in the multi-account tests that we've had an issue where the `awsRegion` was not correct. It was using the client request region, when it should have been the bucket region.

Also, per the documentation for `events`: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html, the `region` field for the event should be the bucket region, which means the `events` client needs to be created with the bucket region and not the client region, as the `region` field in the event is from the internal client.

Also validated the behaviour with AWS, that the region was of the bucket and not the client.

## Changes
Fix the event payload for SQS/SNS/Lambda to use the region of the bucket for the `awsRegion` field.
Fix the region for the `events` client used in the notifier.

Add 2 tests to validate AWS behaviour (one created a bucket in another region than the default client for SQS, and one creating a client in a secondary region and the bucket in the default one).

### DIFF
--- a/localstack/services/s3/notifications.py
+++ b/localstack/services/s3/notifications.py
@@ -381,7 +381,7 @@ class BaseNotifier:
         record = EventRecord(
             eventVersion="2.1",
             eventSource="aws:s3",
-            awsRegion=ctx.region,
+            awsRegion=ctx.bucket_location,
             eventTime=timestamp_millis(ctx.event_time),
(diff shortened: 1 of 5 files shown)


macnev2013 (approved):
Thanks for the PR 🚀 LGTM

PR #9373, opened 2023-10-17
add support for mounting postgresql-proxy

bentsku:
## Motivation
While working on the `postgresql-proxy` library, I've checked before if it was being mounted when using the dev script and saw it wasn't.

## Changes
Added `postgresql-proxy` to the list of host paths to check when trying to mount sources, could verify it works locally by working with it 😄


thrau (approved):
nice!

PR #9374, opened 2023-10-18
Remove deprecated config USE_SINGLE_REGION

viren-nadkarni:
## Motivation

The `USE_SINGLE_REGION` environment config variable was used in early years of LocalStack when multi-region was unsupported. It has since been deprecated.

This PR removes this config variable for the upcoming major release.


localstack/runtime/analytics.py
```diff
@@ -37,7 +37,6 @@
     "REQUIRE_PRO",
     "SKIP_INFRA_DOWNLOADS",
     "SQS_ENDPOINT_STRATEGY",
-    "USE_SINGLE_REGION",
```
alexrashed, line 40: I think we should still keep track of the usage of the old variables (this way we can reason if it's still trying to be used, even though it does not have any effects anymore).

localstack/deprecations.py
```diff
@@ -47,12 +47,6 @@ def is_affected(self) -> bool:
         "0.12.8",
         "PORT_WEB_UI has been removed, and is not available anymore. Please remove this environment variable.",
     ),
-    # Since 0.12.7 - Full multi-region support
-    EnvVarDeprecation(
-        "USE_SINGLE_REGION",
-        "0.12.7",
-        "LocalStack now has full multi-region support. Please remove this environment variable.",
-    ),
```
alexrashed, line 55: I think we should keep this deprecation warning, but just slightly change the text to something like "LocalStack now has full multi-region support.This variable has been removed and does not have any effect anymore. Please remove it form your configuration."

localstack/config.py
```diff
@@ -1286,7 +1283,6 @@ def use_custom_dns():
     "SYNCHRONOUS_SNS_EVENTS",
     "TEST_AWS_ACCOUNT_ID",
     "TF_COMPAT_MODE",
-    "USE_SINGLE_REGION",
```
alexrashed, line 1289: I think we should still forward the variable in order to make sure that we show the deprecation warnings / inform the user that it's not having any effect anymore.

alexrashed (commented):
Great to see that we are finally getting rid of the old deprecated variables. 🧹 
However, I think we should still show warnings to users who are still using it...

localstack/config.py
viren-nadkarni: Good idea, fixed in 46efde0a7

localstack/runtime/analytics.py
viren-nadkarni: Restored in 46efde0a7

alexrashed (approved):
Aweseome! Thanks for addressing the comments and bringing the removal of the `USE_SINGLE_REGION` over the line! :rocket:

PR #9375, opened 2023-10-18
Remove deprecated config DEFAULT_REGION

viren-nadkarni:
## Motivation

The `DEFAULT_REGION` environment config variable was used in early years of LocalStack when multi-region was unsupported. It has since been deprecated.

This PR removes this config variable for the upcoming major release.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -9,7 +9,6 @@ from typing import Any, Dict, List, Mapping, Optional, Tuple, TypeVar, Union
 
 from localstack import constants
 from localstack.constants import (
-    AWS_REGION_US_EAST_1,
     DEFAULT_BUCKET_MARKER_LOCAL,
     DEFAULT_DEVELOP_PORT,
     DEFAULT_LAMBDA_CONTAINER_REGISTRY,
@@ -367,11 +366,6 @@ try:
 except ImportError:
     # dotenv may not be available in lambdas or other environments where config is loaded
     LOADED_PROFILE = None
-
-# default AWS region (DEPRECATED!)
-DEFAULT_REGION = (
-    os.environ.get("DEFAULT_REGION") or os.environ.get("AWS_DEFAULT_REGION") or AWS_REGION_US_EAST_1
-)
 
 # directory for persisting data (TODO: deprecated, simply use PERSISTENCE=1)
 DATA_DIR = os.environ.get("DATA_DIR", "").strip()
@@ -1169,7 +1163,6 @@ CONFIG_ENV_VARS = [
     "CUSTOM_SSL_CERT_PATH",
     "DEBUG",
     "DEBUG_HANDLER_CHAIN",
-    "DEFAULT_REGION",
     "DEVELOP",
     "DEVELOP_PORT",
     "DISABLE_BOTO_RETRIES",
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1163,6 +1163,7 @@ CONFIG_ENV_VARS = [
     "CUSTOM_SSL_CERT_PATH",
     "DEBUG",
     "DEBUG_HANDLER_CHAIN",
+    "DEFAULT_REGION",  # Not functional; deprecated in 0.12.7, removed in 3.0.0
     "DEVELOP",
     "DEVELOP_PORT",
     "DISABLE_BOTO_RETRIES",
--- a/localstack/deprecations.py
+++ b/localstack/deprecations.py
@@ -53,6 +53,7 @@ DEPRECATIONS = [
         "0.12.7",
         "LocalStack now has full multi-region support. Please remove this environment variable.",
     ),
+    # Deprecated in 0.12.7, removed in 3.0.0
     EnvVarDeprecation(
         "DEFAULT_REGION",
         "0.12.7",
--- a/localstack/deprecations.py
+++ b/localstack/deprecations.py
@@ -53,6 +53,11 @@ DEPRECATIONS = [
         "0.12.7",
         "LocalStack now has full multi-region support. Please remove this environment variable.",
     ),
+    EnvVarDeprecation(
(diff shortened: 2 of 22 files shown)


localstack/config.py
```diff
@@ -1168,7 +1162,6 @@ def use_custom_dns():
     "CUSTOM_SSL_CERT_PATH",
     "DEBUG",
     "DEBUG_HANDLER_CHAIN",
-    "DEFAULT_REGION",
```
alexrashed, line 1171: I think we should still forward it to the container such that we can inform the user that it is not used anymore and should be removed from their configuration.

localstack/deprecations.py
```diff
@@ -53,11 +53,6 @@ def is_affected(self) -> bool:
         "0.12.7",
         "LocalStack now has full multi-region support. Please remove this environment variable.",
     ),
-    EnvVarDeprecation(
-        "DEFAULT_REGION",
-        "0.12.7",
-        "LocalStack now has full multi-region support. Please remove this environment variable.",
-    ),
```
alexrashed, line 60: I think this deprecation should be kept, but rephrased such that it says that it does not have any effect anymore.

localstack/runtime/analytics.py
```diff
@@ -9,7 +9,6 @@
 
 TRACKED_ENV_VAR = [
     "DEBUG",
-    "DEFAULT_REGION",
```
alexrashed, line 12: We should still keep the env here.

localstack/utils/aws/aws_stack.py
```diff
@@ -149,11 +150,12 @@ def get_partition(region_name: str = None):
     return boto3.session.Session().get_partition_for_region(region_name)
 
 
+# TODO: Deprecate and remove this
 def get_local_region():
     global LOCAL_REGION
     if LOCAL_REGION is None:
         LOCAL_REGION = get_boto3_region() or ""
-    return config.DEFAULT_REGION or LOCAL_REGION
+    return AWS_REGION_US_EAST_1 or LOCAL_REGION
```
alexrashed, line 158: This can never fall back to `LOCAL_REGION`, can it? What is this function used for, and why should it be removed?

tests/unit/test_partition_rewriter.py
```diff
@@ -272,53 +272,6 @@ def test_arn_partition_rewriting_in_response_with_request_region(encoding):
...
-            status=200,
-            headers={"some-header-with-arn": "arn:aws-us-gov:iam::123456789012:ArnInHeader"},
-        )
-        rewrite_handler.modify_response_revert(response=response, request_region=None)
-
-        assert response.status_code == 200
-        assert response.headers["some-header-with-arn"] == "arn:aws:iam::123456789012:ArnInHeader"
-        assert response.data == to_bytes(
-            json.dumps({"some-data-with-arn": "arn:aws:iam::123456789012:ArnInData"})
-        )
```
alexrashed, line 294: You cannot remove this test, it needs to be ensured that the ARN partition rewriting works without the default region being set.

tests/unit/test_partition_rewriter.py
```diff
@@ -272,53 +272,6 @@ def test_arn_partition_rewriting_in_response_with_request_region(encoding):
...
-        rewrite_handler.modify_response_revert(response, request_region=None)
-
-        assert response.status_code == response.status_code
-        assert (
-            response.headers["some-header-with-arn"]
-            == "arn:aws-us-gov:iam::123456789012:ArnInHeader"
-        )
-        assert response.data == to_bytes(
-            json.dumps({"some-data-with-arn": "arn:aws-us-gov:iam::123456789012:ArnInData"})
-        )
```
alexrashed, line 319: Unfortunately, I think this could be a remaining use case for the default region: The ARN Partition Rewriter needs to be able to determine the default partition in case the request does not contain a region (like for IAM requests).

alexrashed (changes requested):
I am glad we are finally trying to get rid of the `DEFAULT_REGION`.
Unfortunately, it seems that we have to find an alternative for the ARN partition rewriting...
Besides that, I only found some minor things (like that we should still warn the user that they are using a variable that doesn't have an effect anymore)...

localstack/config.py
viren-nadkarni: Fixed, added a comment too

localstack/runtime/analytics.py
viren-nadkarni: :+1:

localstack/utils/aws/aws_stack.py
viren-nadkarni: This is function is related to the `Environment` abstractions defined in this file. It is currently used in Kinesis provider and DynamoDB provider. Ongoing work is focusing on use of `RequestContext.region` instead of these helpers, hence it will be removed together with `Environment` stuff

viren-nadkarni: This change mimics the current behaviour where `config.DEFAULT_REGION` will always be set

https://github.com/localstack/localstack/pull/9375/files#diff-8ee48b2bcad465473db1a1b202781ca2c7c23db7bd1670e79d45668fb2bd718fL371-L374

tests/unit/test_partition_rewriter.py
viren-nadkarni: Added a workaround with https://github.com/localstack/localstack/pull/9482

tests/unit/test_partition_rewriter.py
viren-nadkarni: We now have a new config option to address this usecase https://github.com/localstack/localstack/pull/9482

alexrashed (approved):
Thanks for addressing the comments, and cleaning up all the usages! This is a great step forward with the multi-region support, and to make sure that the region is properly handled for all requests. 🚀

PR #9377, opened 2023-10-18
Remove aws_stack.get_region() from test helpers

viren-nadkarni:
## Motivation

This PR removes the use of `aws_stack.get_region()` from `create_client_with_keys` helper and propagates this change across the codebase.

Use of `aws_stack.get_region()` is associated with several bugs when using multi-region and is currently being removed from the codebase.


tests/aws/services/iam/test_iam.py
```diff
@@ -34,8 +35,10 @@ def test_get_user_without_username_as_user(self, create_user, aws_client):
         )
         account_id = aws_client.sts.get_caller_identity()["Account"]
         keys = aws_client.iam.create_access_key(UserName=user_name)["AccessKey"]
-        wait_for_user(keys)
-        iam_client_as_user = create_client_with_keys("iam", keys=keys)
+        wait_for_user(keys, TEST_AWS_REGION_NAME)
```
dfangl, line 38: Not a fan of using the constants here. We could just use `aws_client.iam.meta.region_name` or something, so it always matches the region of the client.

dfangl (changes requested):
Basically the same comment for most of the changes: We can just dynamically use `aws_client.service.meta.region_name` to get the current region, without creating a dependency on setting a variable.

tests/aws/services/iam/test_iam.py
viren-nadkarni: Replaced with the `region` fixture

alexrashed (approved):
@dfangl's comment was addressed, and the other changes are looking good! 💯 🧹 🚀

PR #9381, opened 2023-10-18
Kinesis: Remove aws_stack.get_region() and get_aws_account_id()

viren-nadkarni:
## Motivation

## Implementation

This PR removes the use of the above utilities, instead passing the account ID and region down the call stack. The changes are made for Kinesis, Kinesis Analytics and Firehose


alexrashed (approved):
LGTM! 🧹 🚀

Issue #9387, opened 2023-10-18
S3 V3 issue with aws-s3 docker storage driver


bentsku:
I have this script which is using a package to communicate with s3, and it doesn't work with v3 provider.

```
package main

import (
	"bytes"
	"context"
	"fmt"
	"errors"

	_ "github.com/docker/distribution/registry/storage/driver/s3-aws"
	"github.com/docker/distribution/registry/storage/driver/factory"
)

func main() {
	params := map[string]interface{}{
		"name":          "s3",
		"rootdirectory": "/root",
		"regionendpoint": "localhost:4566",
		"region":        "us-east-2",
		"bucket":        "store",
		"secure":        true,
		"skipverify":    true,
	}

	store, err := factory.Create("s3", params)
	if err != nil {
		panic(err)
	}

	filepath := "/test/test"

	fileWriter, err := store.Writer(context.TODO(), filepath, false)
	if err != nil {
		panic(err)
	}

	if _, err = fileWriter.Write([]byte("first")); err != nil {
		panic(err)
	}

	fileWriter.Close()

	fileWriterAppend, err := store.Writer(context.TODO(), filepath, true)
	if err != nil {
		panic(err)
	}

	if _, err = fileWriterAppend.Write([]byte("second")); err != nil {
		panic(err)
	}

	fileWriterAppend.Commit()
	fileWriterAppend.Close()

	fileContent, err := store.GetContent(context.Background(), filepath)
	if err != nil {
		panic(err)
	}

	fmt.Println(string(fileContent))

	if string(fileContent) != "firstsecond" {
		panic(errors.New("not equal"))
	}

	fileReader, err := store.Reader(context.Background(), filepath, 0)
	if err != nil {
		panic(err)
	}

	buf := new(bytes.Buffer)
	buf.ReadFrom(fileReader)
	content := buf.String()

	if content != "firstsecond" {
		panic(errors.New("not equal"))
	}

	fmt.Println(string(content))
}
```
This is the output:
```
firstfirsts
panic: not equal

goroutine 1 [running]:
main.main()
        /home/peusebiu/localstack_test/main.go:62 +0x689
```

With v2 provider, script works as expected:
```
firstsecond
firstsecond
```

localstack v3 logs:
```
LocalStack version: 2.3.2
LocalStack build date: 2023-10-03
LocalStack build git hash: 1bccf790

[...]

PR #9388, opened 2023-10-18
fix s3 v3 multipart overwrite when key exists

bentsku:
## Motivation
Reported in #9387: when creating a multipart upload on an existing key, we would not truncate the key before overwrite it like we do when calling `PutObject`.

This was due to our optimization with `append` on the StoredS3Object, which was stateless, and bad oversight on my side.

Flow for the sample in the linked issue is:
- CreateMultipart on key `test`
- UploadPart with value "first"
- CompletePart on key `test`
- GetObject on key `test` -> value is "first"
- CreateMultipart again on key `test`
- UploadPart with the result of previous GetObject ("first") + "second" -> value = "firstsecond"
- CompleteMultipart on key `test`
- GetObject on key `test`, result should be "firstsecond"

Issue was appending on top of existing key, resulting in "firstfirstsecond".

## Changes
Added a new `truncate` method to the base `S3StoredObject` to be able to reset the file from the StoredMultipart.
Add `s3_object.seek(0) + s3_object.truncate()` in `complete_multipart` so that the existing file object representing the key is properly truncated before we write and we do not append the multipart on top of the existing key. 

Added an AWS validated test validating this behaviour.

### DIFF
--- a/localstack/services/s3/v3/storage/core.py
+++ b/localstack/services/s3/v3/storage/core.py
@@ -92,6 +92,9 @@ class S3StoredObject(abc.ABC, Iterable[bytes]):
     def seek(self, offset: int, whence: int = 0) -> int:
         pass
 
+    def truncate(self, size: int = None) -> int:
+        pass
+
     @property
     @abc.abstractmethod
     def checksum(self) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


thrau (approved):
nice catch, easy to miss!

PR #9389, opened 2023-10-18
Migrate Lambda to gateway listen

joe4dev:
## Motivation

Adopt the new gateway-listen network configuration instead of `config.EDGE_PORT` (deprecated, at least as a configurable variable)

## Changes

Change `config.EDGE_PORT` to `GATEWAY_LISTEN[0].port`

## Discussion

* The K8 executor still uses `config.get_edge_port_http()` but that will be migrated with the deprecation of `EDGE_PORT_HTTP` upon 3.0.
* The old provider still uses `EDGE_PORT` but will be removed anyways upon 3.0

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -1823,7 +1823,8 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         url_id = api_utils.generate_random_url_id()
 
         host_definition = localstack_host(
-            use_localhost_cloud=True, custom_port=config.EDGE_PORT_HTTP or config.EDGE_PORT
+            use_localhost_cloud=True,
+            custom_port=config.EDGE_PORT_HTTP or config.GATEWAY_LISTEN[0].port,
         )
         fn.function_url_configs[normalized_qualifier] = FunctionUrlConfig(
             function_arn=function_arn,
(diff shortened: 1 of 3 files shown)


localstack/services/lambda_/provider.py
```diff
@@ -1823,7 +1823,8 @@ def create_function_url_config(
         url_id = api_utils.generate_random_url_id()
 
         host_definition = localstack_host(
-            use_localhost_cloud=True, custom_port=config.EDGE_PORT_HTTP or config.EDGE_PORT
+            use_localhost_cloud=True,
+            custom_port=config.EDGE_PORT_HTTP or config.GATEWAY_LISTEN[0].port,
```
dfangl, line 1827: I would remove the EDGE_PORT_HTTP here, makes no sense I think, as those values are likely the same.

dfangl (approved):
Nice, LGTM! Only a minor nit.

localstack/services/lambda_/provider.py
joe4dev: I thought about batching breaking changes in a separate PR that removes `EDGE_PORT_HTTP`, `EDGE_BIND_HOST`, and write-access of `EDGE_PORT` localstack-wide.
I can also sneak it in here if you think it doesn't affect anyone 🤷

PR #9398, opened 2023-10-19
Make ARN builders require account ID and region args

viren-nadkarni:
## Motivation

ARNs in LocalStack were incorrect in certain situations because they used account ID and region from the thread context via the use of `get_aws_account_id()` and `aws_stack.get_region()`.

## Implementation

This PR is a complete refactoring of our ARN builder utilities that make the `account_id` and `region_name` field mandatory. The correct values for these are passed down the from the request context.

## Related

Depends on:
- https://github.com/localstack/localstack/pull/9400
- https://github.com/localstack/localstack/pull/9401
- https://github.com/localstack/localstack/pull/9418


localstack/services/apigateway/models.py
```diff
@@ -82,7 +83,9 @@ def __init__(self):
 
         self.account.update(
             {
-                "cloudwatchRoleArn": arns.role_arn("api-gw-cw-role"),
+                "cloudwatchRoleArn": arns.role_arn(
+                    "api-gw-cw-role", DEFAULT_AWS_ACCOUNT_ID
+                ),  # FIXME: account ID must be of the current store
```
viren-nadkarni, line 88: This mimics the current behaviour where the account ID of the first APIGW request when the provider starts up. The effort involved to refactor APIGW codebase to fix this is substantial and out of the scope of this PR and will be addressed in the future.

localstack/services/cloudformation/models/cloudformation.py
```diff
@@ -150,7 +150,9 @@ def generate_waitcondition_url(account_id: str, region_name: str, stack_name: st
     region = client.meta.region_name
 
     bucket = f"cloudformation-waitcondition-{region}"
-    key = arns.cloudformation_stack_arn(stack_name=stack_name)
+    key = arns.cloudformation_stack_arn(
+        stack_name=stack_name, stack_id="id-123", account_id=account_id, region_name=region_name
```
viren-nadkarni, line 154: The value for stack ID `id-123` is picked up from the existing default argument.

localstack/services/ecr/resource_providers/aws_ecr_repository.py
```diff
@@ -96,7 +97,9 @@ def create(
         )
         model.update(
             {
-                "Arn": arns.get_ecr_repository_arn(model["RepositoryName"]),
+                "Arn": arns.get_ecr_repository_arn(
+                    model["RepositoryName"], DEFAULT_AWS_ACCOUNT_ID, AWS_REGION_US_EAST_1
+                ),
```
viren-nadkarni, line 102: This is a mock repo so it's okay to use the default values.

localstack/services/lambda_/lambda_api.py
```diff
@@ -2338,7 +2338,7 @@ def create_code_signing_config():
     signing_profile_version_arns = data.get("AllowedPublishers").get("SigningProfileVersionArns")
 
     code_signing_id = "csc-%s" % long_uid().replace("-", "")[0:17]
-    arn = arns.code_signing_arn(code_signing_id)
+    arn = arns.code_signing_arn(code_signing_id, DEFAULT_AWS_ACCOUNT_ID, AWS_REGION_US_EAST_1)
```
viren-nadkarni, line 2341: For the legacy Lambda provider, again the effort involved to ensure ARNs are correct is substantial hence we resort to default values.

localstack/services/lambda_/lambda_api.py
```diff
@@ -2338,7 +2338,8 @@ def create_code_signing_config():
     signing_profile_version_arns = data.get("AllowedPublishers").get("SigningProfileVersionArns")
 
     code_signing_id = "csc-%s" % long_uid().replace("-", "")[0:17]
-    arn = arns.code_signing_arn(code_signing_id)
+    # See note above on the use of `aws_stack.get_region()` and `get_aws_account_id()`
+    arn = arns.code_signing_arn(code_signing_id, get_aws_account_id(), aws_stack.get_region())
```
viren-nadkarni, line 2342: For the legacy Lambda provider, the effort involved to refactor the codebase is substantial so we make an exception and use the `get_region()` and `get_aws_account_id()` helpers.

tests/aws/services/stepfunctions/legacy/test_stepfunctions_legacy.py
```diff
@@ -303,7 +303,7 @@ class TestStateMachine:
     @markers.aws.needs_fixing
     def test_create_choice_state_machine(self, custom_client):
         state_machines_before = custom_client.stepfunctions.list_state_machines()["stateMachines"]
-        role_arn = arns.role_arn("sfn_role", SF_TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME)
```
viren-nadkarni, line 306: The third argument of this function was actually called `env=` and was unused within the function.

This is now removed.

localstack/services/apigateway/models.py
bentsku: I see the problem here, the `_account_id` field of the store is not passed to the `__init__` called but manually attached later on to the store, so we can't access it here. Was there a specific reason it was done this way?

I don't think this mimics the current behaviour though:
```bash
$ awslocal apigateway get-account
{
    "cloudwatchRoleArn": "arn:aws:iam::000000000000:role/api-gw-cw-role",
    ...
}
$ AWS_ACCESS_KEY_ID=000000000002 awslocal apigateway get-account
{
    "cloudwatchRoleArn": "arn:aws:iam::000000000002:role/api-gw-cw-role",
    ...
}
```
This might create some issues, but I don't think we currently have the CloudWatch integration with APIGW, so it shouldn't be an IAM issue I hope.

Do you see how this could be achieved with the current architecture? Even a big a refactor in APIGW might not solve this if `ApiGatewayStore.account` is called in different places, we would need to manually update the account after the store instantiation. How can we address this issue?

localstack/services/dynamodb/provider.py
```diff
@@ -1732,14 +1753,10 @@ def is_index_query_valid(account_id: str, region_name: str, query_data: dict) ->
...
-def has_streams_enabled(table_name: str):
+def has_streams_enabled(account_id: str, region_name: str, table_name: str):
     if not table_name:
         return
-    table_arn = arns.dynamodb_table_arn(table_name)
-    # TODO: change this, it's using `_resource_arn` which calls get_aws_account_id() and get_region() which we
-    #  are extracting from the ARN after?
-    account_id = extract_account_id_from_arn(table_arn)
-    region_name = extract_region_from_arn(table_arn)
+    table_arn = arns.dynamodb_table_arn(table_name, account_id=account_id, region_name=region_name)
```
bentsku, line 1759: 🎉

tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -1978,7 +1978,9 @@ def test_api_gateway_sqs_integration_with_event_source(
         # create event source for sqs lambda processor
         event_source_data = {
             "FunctionName": integration_lambda,
-            "EventSourceArn": arns.sqs_queue_arn(sqs_queue),
+            "EventSourceArn": arns.sqs_queue_arn(
+                sqs_queue, account_id=TEST_AWS_ACCOUNT_ID, region_name=TEST_AWS_REGION_NAME
+            ),
```
bentsku, line 1983: nit: I think we could use the fixture here for the sqs queue ARN.

bentsku (commented):
Wow, quite an extensive set of changes, that's awesome! This is really a great!
I just have a question concerning the ApiGatewayStore instantiation with default values. Will we have a solution in the future for this use case? I'd like to know if we can have an idea of what to do to fix this issue before changing it that way.

localstack/services/apigateway/models.py
viren-nadkarni: > Was there a specific reason it was done this way?

IIRC this was done to avoid large scale code refactoring, you see `RegionBackend`s, the precursor to `Store`, defined `__init__` functions. We wanted to introduce stores in a backward compatible manner.

> Do you see how this could be achieved with the current architecture?

Basically this ties back to the previous question, the most straightforward way is not to define business logic in the store definition. Initialiser functions and other member methods with business logic must not be defined as part of the store definition.

tests/aws/services/apigateway/test_apigateway_basic.py
viren-nadkarni: This is resolved on the merge from `master`

localstack/services/cloudformation/models/cloudformation.py
steffyP: sorry, just realized that the definition changed as well :) but maybe it makes sense then to use a short-uid here instead? I am not sure why it was hard-coded in the first place though 🤔 Do you have any background information maybe?

steffyP (approved):
Thanks for cleaning up the usage of region and account-ids 🚀 👍 
The changes LGTM.

Just added a question regarding the usage of the default stack ID `id-123` for the cloudformation stack arn.

localstack/services/cloudformation/models/cloudformation.py
viren-nadkarni: > Do you have any background information maybe?

I don't know I'm afraid! I was hesitant to change it to `short_uid()` because I grepped it and saw it being used in some snapshots.

steffyP: I see. Maybe we can consult with @dominikschubert when he is back. 
Anyhow, this should not be blocking the PR.

localstack/services/apigateway/models.py
bentsku: I understand. I suppose we could get a quick fix once this is merged in `get_account` or something similar, to initialize the account value in the store if it doesn't exist instead. Maybe we could get something quick that wouldn't need a big refactor, I'll try to make something once this is merged. Thanks for the precisions!

bentsku (approved):
LGTM! I'll try to get something up to fix the APIGateway issue in the coming days. Thanks again for tackling this!

PR #9399, opened 2023-10-19
switch to ruff and upgrade test tooling

alexrashed:
## Motivation
This PR aims at upgrading our dev and test python infrastructure to the latest versions / tools:
- Use `ruff` instead of `flake8`, `isort`, `autoflake`
- Upgrade `pre-commit`
- Upgrade a bunch of other test / dev dependencies (most notably `pytest` and `pre-commit`)

This PR supersedes https://github.com/localstack/localstack/pull/9218, which is outdated (it has merge conflicts by now, and does not contain the switch to ruff).

## Changes
The changes are separated in 3 commits:
1. Introduce and configure `ruff` instead of `flake8`, `isort`, `autoflake` and update other dependencies.
  - This introduces a quite long list of rule exclusions which caused linting errors in ruff.
  - The plan is to fix most of these issues and reduce the list of exclusions with an upcoming PR (in order to get this PR over the line fast to avoid conflicts).
3. Apply auto-fixes by `ruff`,
4. Apply new formatting with black (due to the major version upgrade).

## Testing
If we're 💚, we are good to go. The applied fixes are all safe and cannot modify any behavior.


pyproject.toml
```diff
@@ -8,17 +8,60 @@ line_length = 100
...
-max-line-length = 110
-ignore = 'E203,E266,E501,W503,F403'
-select = 'B,C,E,F,I,W,T4,B9'
-exclude = 'node_modules,.venv*,venv*,dist,build,target,*.egg-info,fixes,localstack/infra,localstack/node_modules,.filesystem,.git,localstack/services/stepfunctions/asl/antlr/runtime'
+[tool.ruff]
+# Always generate Python 3.8-compatible code.
+target-version = "py38"
+line-length = 110
+ignore = [
+    "B005", # TODO Using `.strip()` with multi-character strings is misleading
```
dominikschubert, line 16: nice TODO list for us to work through :+1:

dominikschubert (approved):
LGTM! Didn't see anything controversial in the new formatting either.


Ruff is looking good so far :rocket:

PR #9400, opened 2023-10-19
Lambda: Use proper account ID and region during ARN construction

viren-nadkarni:
## Motivation

Consequently, these utilities must not be used in tests because tests run in a different thread, this causes the functions to return the wrong values.

The ARN contruction utility `lambda_function_arn` makes use of these helpers which is wrong. This leads to bugs when using multi-region and multi-accounts in LocalStack.

## Implementation

This PR removes the use of explicit constants for account ID and region for ARN construction in tests and everywhere else. 

In the long term, we want to get rid of `get_region` and `get_aws_account_id`. Certain legacy modules like the v1 Lambda provider are designed in a way to be reliant on these functions. As a consequence, these utils cannot be removed from these legacy modules without significant refactoring. Therefore when these provider is removed, these utils will also be safely removed.


localstack/utils/aws/arns.py
```diff
@@ -232,9 +234,12 @@ def fix_arn(arn):
     """Function that attempts to "canonicalize" the given ARN. This includes converting
     resource names to ARNs, replacing incorrect regions, account IDs, etc."""
     if arn.startswith("arn:aws:lambda"):
-        parts = arn.split(":")
-        region = parts[3] if parts[3] in get_valid_regions() else get_region()
-        return lambda_function_arn(lambda_function_name(arn), region_name=region)
+        arn_data = parse_arn(arn)
```
joe4dev, line 237: great to use the more robust `parse_arn` utils here instead of the custom split by `:`  👍

localstack/utils/testutil.py
```diff
@@ -340,10 +340,9 @@ def create_lambda_api_gateway_integration(
 
     # create Lambda
     zip_file = create_lambda_archive(handler_file, get_content=True, runtime=runtime)
-    create_lambda_function(
+    func_arn = create_lambda_function(
         func_name=func_name, zip_file=zip_file, runtime=runtime, client=lambda_client
-    )
-    func_arn = arns.lambda_function_arn(func_name)
+    )["CreateFunctionResponse"]["FunctionArn"]
```
joe4dev, line 345: nice catch 👍

tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -653,16 +651,13 @@ def test_malformed_response_apigw_invocation(self, create_lambda_function, aws_c
         lambda_resource = "/api/v1/{proxy+}"
         lambda_path = "/api/v1/hello/world"
 
-        create_lambda_function(
+        lambda_uri = create_lambda_function(
```
joe4dev, line 654: nit: should be arn instead of URI (but not worth running the pipeline again because of this)

tests/unit/test_lambda.py
```diff
@@ -8,6 +8,7 @@
 
 from localstack import config
 from localstack.aws.accounts import get_aws_account_id
+from localstack.constants import TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME
```
joe4dev, line 11: Did @dfangl refer to these constants here: https://github.com/localstack/localstack/pull/9377#discussion_r1368390630 ?

joe4dev (approved):
LGTM 👍 

Thanks for tackling this @viren-nadkarni 
Looking forward to removing the legacy Lambda provider altogether 🚀

PR #9401, opened 2023-10-19
APIGW: Use proper account ID and region during Invocation construction

viren-nadkarni:
## Motivation

Consequently, these utilities must not be used in tests because tests run in a different thread, this causes the functions to return the wrong values.

The ARN contruction utility `apigateway_invocations_arn` makes use of these helpers which is wrong. This leads to bugs when using multi-region and multi-accounts in LocalStack.

## Implementation

This PR removes the use of the above utilties from the tests and makes use of proper constants for ARN construction.

### DIFF
--- a/localstack/utils/aws/arns.py
+++ b/localstack/utils/aws/arns.py
@@ -346,9 +346,9 @@ def kinesis_stream_name(kinesis_arn):
     return kinesis_arn.split(":stream/")[-1]
 
 
-def apigateway_invocations_arn(lambda_uri, region_name: str = None):
+def apigateway_invocations_arn(lambda_uri: str, region_name: str) -> str:
     return "arn:aws:apigateway:%s:lambda:path/2015-03-31/functions/%s/invocations" % (
-        region_name or get_region(),
+        region_name,
         lambda_uri,
     )
 
--- a/localstack/utils/testutil.py
+++ b/localstack/utils/testutil.py
@@ -25,7 +25,12 @@ import requests
 
 from localstack import config
 from localstack.aws.accounts import get_aws_account_id
-from localstack.constants import LOCALHOST_HOSTNAME, LOCALSTACK_ROOT_FOLDER, LOCALSTACK_VENV_FOLDER
+from localstack.constants import (
+    LOCALHOST_HOSTNAME,
+    LOCALSTACK_ROOT_FOLDER,
+    LOCALSTACK_VENV_FOLDER,
+    TEST_AWS_REGION_NAME,
+)
 from localstack.services.lambda_.lambda_api import LAMBDA_TEST_ROLE
 from localstack.services.lambda_.lambda_utils import (
     LAMBDA_DEFAULT_HANDLER,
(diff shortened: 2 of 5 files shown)


localstack/utils/aws/arns.py
```diff
@@ -346,9 +346,9 @@ def kinesis_stream_name(kinesis_arn):
     return kinesis_arn.split(":stream/")[-1]
 
 
-def apigateway_invocations_arn(lambda_uri, region_name: str = None):
+def apigateway_invocations_arn(lambda_uri: str, region_name: str) -> str:
     return "arn:aws:apigateway:%s:lambda:path/2015-03-31/functions/%s/invocations" % (
-        region_name or get_region(),
+        region_name,
         lambda_uri,
     )
```
bentsku, line 353: q: should this use the region of the original lambda ARN instead? what is the purpose of overriding it?

bentsku (commented):
Thanks for tackling this. I believe the original `apigateway_invocations_arn` region overriding might be wrong through? Shouldn't we use the region from the lambda arn?

localstack/utils/aws/arns.py
viren-nadkarni: My knowledge of APIGW is non-existent, but it could be possible that the APIGW is invoking a Lambda function in another region?

bentsku: I think the invocation path should always the extended lambda ARN, basically if you create a lambda in `us-west-1` but modify the region to be `us-east-1` then the resource wouldn't exist, I believe APIGW will use the region of the invocation ARN to then invoke the lambda. It should be the region of the original lambda ARN.

bentsku: Yes, exactly, you're right! https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-cross-account-lambda-integrations.html

This is here. So yeah, we will need the region name for the APIGW region, this is right! Sorry for the misleading comment 😬

bentsku: This is the code which get the lambda ARN from the invocation ARN:
```python
# localstack.services.apigateway.integration.LambdaIntegration._lambda_integration_uri
uri = uri.split(":lambda:path")[1].split("functions/")[1].split("/invocations")[0]
```
So it does not take into account the APIGW region,

bentsku (approved):
LGTM! Sorry for the misleading comment, this is right! When using the helper method with multi region scenario in APIGW, we should note to be careful that we should use the same region as the client creating the API I believe. Thanks for tackling this!

localstack/utils/aws/arns.py
bentsku: I'll add to this: it seems we can indeed construct the ARN, but then AWS will reject it, it's not possible to do so. https://github.com/aws/aws-cdk/issues/17458

We should properly construct the wrong ARN anyway using the API region basically. But we should then reject the request if the invocation ARN passed to `create_integration` has 2 different regions in it, which is a different issue all together. Thanks for putting this into light, multi region is always showing nice behaviours we were not aware!

PR #9402, opened 2023-10-19
add config variable DYNAMODB_LOCAL_PORT to control ddb local port

thrau:
## Motivation

We are investigating performance issues with dynamodb, and there it will be helpful to control which port dynamodb is created on. Moreover, this will enable workarounds where people may need to access ddb local directly. Tested locally.

cc @bentsku 

## Changes

* add `DYNAMODB_LOCAL_PORT` port config, that is used if not None or > 0
* also sneaked in a changed that avoids a call to `random` when using `put_item`

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -852,6 +852,9 @@ DYNAMODB_HEAP_SIZE = os.environ.get("DYNAMODB_HEAP_SIZE", "").strip() or "256m"
 # single DB instance across multiple credentials are regions
 DYNAMODB_SHARE_DB = int(os.environ.get("DYNAMODB_SHARE_DB") or 0)
 
+# the port on which to expose dynamodblocal
+DYNAMODB_LOCAL_PORT = int(os.environ.get("DYNAMODB_LOCAL_PORT") or 0)
+
 # Used to toggle PurgeInProgress exceptions when calling purge within 60 seconds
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Thanks a lot for adding this, this was really helpful in testing performance and what we can achieve. However, beware as there is a major limitation when using the port to directly connect to DynamoDB.
We do headers manipulation, so when we create the table in LS, we can’t access it with the same credentials directly through the port: you need to use the same logic for the credentials as internally used by LS:

Example:
```python
port_to_use = 4566
port_to_use_ddb = 4510

account_id = "000000000000"
secret_key = "test"
region = "us-east-1"
ddb_account_id = f"{account_id}{region}".replace("-", "")

# this client is used to communicated to DDB via LocalStack
client = boto3.client(
    'dynamodb',
    endpoint_url=f'http://localhost:{port_to_use}',
    aws_access_key_id=account_id,
    aws_secret_access_key=secret_key,
    region_name="us-east-1",
)
# this client is used to directly communicate via the exposed port 
client_ddb = boto3.client(
    'dynamodb',
    endpoint_url=f'http://localhost:{port_to_use_ddb}',
    aws_access_key_id=ddb_account_id,
    aws_secret_access_key=secret_key,
    region_name="us-east-1",
)
```

thrau:
that's really helpful @bentsku thank you!

PR #9404, opened 2023-10-19
Implementation of AWS::S3::* Resource Provider support for CloudFormation

Morijarti:
## Changes
Implementation of support for:
- "AWS::S3::BucketPolicy"
- "AWS::S3::Bucket"


localstack/services/s3/resource_providers/aws_s3_bucket.py
```diff
@@ -0,0 +1,722 @@
...
+
+        IAM permissions required:
+          - s3:DeleteBucket
+        """
+        model = request.desired_state
+        s3_client = request.aws_client_factory.s3
+
+        try:
+            s3_client.delete_bucket_policy(Bucket=model["BucketName"])
+        except s3_client.exceptions.ClientError:
```
Morijarti, line 666: In original code of BaseLegacyProvider we were catching Exception. Instead of it, I went with more specific exception here.
Should I revert it to Exception?

pinzon: @bentsku Probably would know if the NotFound exception is considered a ClientError.

bentsku: why do we need to manually delete the bucket policy in the first place? Shouldn't it be deleted by deleting the bucket?

Morijarti: I've removed explicit deletion of bucket policies and replaced exception with more general `ClientError`
Let me know If there is anything else I can do to improve the code.

pinzon (approved):
LGTM

PR #9405, opened 2023-10-19
Migrate deprecated edge variables

joe4dev:
Depends on `LOCALSTACK_HOST` migration: https://github.com/localstack/localstack/pull/9390

## Motivation

`EDGE_PORT`, `EDGE_PORT_HTTP` and `EDGE_BIND_HOST` are replaced by `GATEWAY_LISTEN`, and have been deprecated since version 2.0. These variables should be removed in a major version, and can be done in version 3.0.

## Changes

* Make `EDGE_PORT` read-only. Internal use is widespread and therefore we consider it as a shortcut for now with the semantics `EDGE_PORT == GATEWAY_LISTEN[0].port`. We also keep exposing `EDGE_PORT` in LocalStack compute services because we have not advertised a migration path yet.
* Remove `EDGE_PORT_HTTP`
* Remove `EDGE_BIND_HOST`

## Background

* Announced networking changes upon 2.0: https://discuss.localstack.cloud/t/upcoming-changes-for-localstack-v2/239#networking-7
* GATEWAY_LISTEN behavior regarding protocols `http` vs `https`:
    
    > When we bind to a ip/port, we accept both http and https requests, so the protocol distinction is not important for the purposes of our hypercorn server. For example you can send HTTPS requests to both 4566 and 443 in your example above. I think that means that we allow all combinations of port and protocol
    >

* More details in internal decision document: https://www.notion.so/localstack/Migration-path-for-LOCALSTACK_HOSTNAME-and-EDGE_PORT-9cf800595f854add96ce2788a6d93367?pvs=4

## Discussion

* CORS is now more permissive (see `localstack/aws/handlers/cors.py`) and the tests required updates (`tests/unit/test_cors.py`). Is this ok?
* Monkeypatching of `GATEWAY_LISTEN` needs to use typed `[HostAndPort]` rather than the user-facing string-based API. Does this work properly without side effects? Do we need to provide any guards when used with strings?
* StepFunctions V1: Does anything depend on the environment variable `EDGE_PORT_HTTP` (generated in localstack.services.stepfunctions.stepfunctions_starter.StepFunctionsServer.generate_env_vars) used when starting the Java StepFunctionsLocal.jar ? (I would leave it for now as this StepFunctions implementation will be deprecated anyway).

## TODO

- [x] Rebase upon the dependent `LOCALSTACK_HOST` migration https://github.com/localstack/localstack/pull/9390
- [x] Fix `tests.unit.test_config.TestEdgeVariablesDerivedCorrectly.test_localstack_host_overrides_edge_variables` depending on the migration to `LOCALSTACK_HOST` https://github.com/localstack/localstack/pull/9390/files#diff-8ee48b2bcad465473db1a1b202781ca2c7c23db7bd1670e79d45668fb2bd718fR616
- [x] Remove duplicated commit from Viren from this PR (upon next rebase)
- [x] Synchronize merge with Lambda removal PRs (might have to resolve a few conflicts) https://github.com/localstack/localstack/pull/9543

## Follow-up

- [ ] Remove `get_edge_port_http()` (this refactoring can also be done post v3 to limit the impact; used 13x in LS + 13x in ext)

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1209,6 +1209,9 @@ CONFIG_ENV_VARS = [
     "WAIT_FOR_DEBUGGER",
     "WINDOWS_DOCKER_MOUNT_PREFIX",
     # Removed in 3.0.0
+    "EDGE_BIND_HOST",  # deprecated since 2.0.0
+    "EDGE_PORT",  # deprecated since 2.0.0
+    "EDGE_PORT_HTTP",  # deprecated since 2.0.0
     "LAMBDA_XRAY_INIT",  # deprecated since 2.0.0
     "LAMBDA_CODE_EXTRACT_TIME",  # deprecated since 2.0.0
     "LAMBDA_CONTAINER_REGISTRY",  # deprecated since 2.0.0
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -596,10 +596,6 @@ class UniqueHostAndPortList(List[HostAndPort]):
         super().append(value)
 
 
-def default_gateway_listen() -> List[HostAndPort]:
-    return [HostAndPort(host=default_ip, port=constants.DEFAULT_PORT_EDGE)]
-
-
 def populate_legacy_edge_configuration(
     environment: Mapping[str, str]
 ) -> Tuple[HostAndPort, UniqueHostAndPortList, int]:
@@ -637,7 +633,7 @@ def populate_legacy_edge_configuration(
             )
     else:
         # use default if gateway listen is not defined
-        gateway_listen = default_gateway_listen()
+        gateway_listen = [HostAndPort(host=default_ip, port=localstack_host_port)]
 
     # the actual value of the LOCALSTACK_HOST port now depends on what gateway listen actually listens to.
     if localstack_host_raw is None:
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -596,9 +596,13 @@ class UniqueHostAndPortList(List[HostAndPort]):
         super().append(value)
 
 
+def default_gateway_listen() -> List[HostAndPort]:
+    return [HostAndPort(host=default_ip, port=constants.DEFAULT_PORT_EDGE)]
+
+
 def populate_legacy_edge_configuration(
     environment: Mapping[str, str]
-) -> Tuple[HostAndPort, UniqueHostAndPortList, str, int, int]:
+) -> Tuple[HostAndPort, UniqueHostAndPortList, int]:
     localstack_host_raw = environment.get("LOCALSTACK_HOST")
     gateway_listen_raw = environment.get("GATEWAY_LISTEN")
 
@@ -632,11 +636,8 @@ def populate_legacy_edge_configuration(
                 )
             )
     else:
-        edge_port = int(environment.get("EDGE_PORT", localstack_host_port))
-        edge_port_http = int(environment.get("EDGE_PORT_HTTP", 0))
-        gateway_listen = [HostAndPort(host=default_ip, port=edge_port)]
-        if edge_port_http:
-            gateway_listen.append(HostAndPort(host=default_ip, port=edge_port_http))
+        # use default if gateway listen is not defined
+        gateway_listen = default_gateway_listen()
 
     # the actual value of the LOCALSTACK_HOST port now depends on what gateway listen actually listens to.
     if localstack_host_raw is None:
@@ -654,20 +655,13 @@ def populate_legacy_edge_configuration(
     assert gateway_listen is not None
     assert localstack_host is not None
 
-    # derive legacy variables from GATEWAY_LISTEN unless GATEWAY_LISTEN is not given and
-    # legacy variables are
-    edge_bind_host = legacy_fallback("EDGE_BIND_HOST", gateway_listen[0].host)
+    # derive legacy variables from GATEWAY_LISTEN
     edge_port = int(legacy_fallback("EDGE_PORT", gateway_listen[0].port))
-    edge_port_http = int(
-        legacy_fallback("EDGE_PORT_HTTP", 0),
-    )
 
     return (
         localstack_host,
         UniqueHostAndPortList(gateway_listen),
-        edge_bind_host,
         edge_port,
-        edge_port_http,
     )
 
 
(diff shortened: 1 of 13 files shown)


tests/unit/test_config.py
```diff
@@ -232,30 +198,6 @@ def test_gateway_listen_multiple_addresses(self):
         ]
         # take the first value
         assert edge_port == 9999
-        assert edge_port_http == 0
-        assert edge_bind_host == "0.0.0.0"
-
-    def test_legacy_variables_override_if_given(self, default_ip):
```
simonrw, line 238: Could we reinstate this test but invert the logic, i.e. that if `GATEWAY_LISTEN` is _not_ specified, that setting `EDGE_PORT` does not overwrite it?

tests/unit/test_cors.py
```diff
@@ -2,22 +2,31 @@
...
 from localstack import config
 from localstack.aws.handlers import cors
+from localstack.config import HostAndPort
+
+# The default host depends on whether running in Docker (see config.py::default_ip) but that's good enough for testing:
+default_gateway_listen = [HostAndPort(host="0.0.0.0", port=4566)]
+default_gateway_listen_ext = [
+    HostAndPort(host="0.0.0.0", port=4566),
+    HostAndPort(host="0.0.0.0", port=443),
+]
```
simonrw, line 12: It feels a bit weird testing the -ext configuration here, but :shrug: it was here before... :joy:

simonrw (changes requested):
Excited to see these changes landing. It's been a while since 2.0 and the logic is much cleaner!



A few comments that I'd like to resolve before we merge, but nothing significant.

simonrw (commented):
Also a change I missed when updating GATEWAY_LISTEN: on [`config.py` line `671`](https://github.com/localstack/localstack/pull/9405/files#diff-8ee48b2bcad465473db1a1b202781ca2c7c23db7bd1670e79d45668fb2bd718fR671) the type signature of `GATEWAY_LISTEN` should be `UniqueHostAndPortList`, not `List`. Is it ok to update this? (Or remove the dedicated type signature, since it should come from the signature of `populate_legacy_edge_configuration`.

localstack/config.py
```diff
@@ -660,20 +657,13 @@ def legacy_fallback(envar_name: str, default: T) -> T:
     assert gateway_listen is not None
     assert localstack_host is not None
 
-    # derive legacy variables from GATEWAY_LISTEN unless GATEWAY_LISTEN is not given and
-    # legacy variables are
-    edge_bind_host = legacy_fallback("EDGE_BIND_HOST", gateway_listen[0].host)
+    # derive legacy variables from GATEWAY_LISTEN
     edge_port = int(legacy_fallback("EDGE_PORT", gateway_listen[0].port))
```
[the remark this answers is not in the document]
joe4dev: oh yes, very important catch. Thank you 🙏 
That might have gotten lost during a rebase 😱 😢 
Glad we caught this. Fixed.

localstack/config.py
```diff
@@ -604,7 +604,7 @@ def append(self, value: HostAndPort):
 
 def populate_legacy_edge_configuration(
```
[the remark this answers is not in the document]
joe4dev: fully agree 👍 I removed the `legacy` part.
(I also stumbled upon this but delayed the change after the rebase and missed it)

localstack/config.py
```diff
@@ -687,9 +677,7 @@ def legacy_fallback(envar_name: str, default: T) -> T:
     # <ip_address>:<port>(,<ip_address>:port>)*
     GATEWAY_LISTEN,
     # -- Legacy variables
-    EDGE_BIND_HOST,
     EDGE_PORT,
```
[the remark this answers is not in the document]
joe4dev: I guess so. It is used more often in testing than not and `EDGE_PORT` is still widely used in our codebase. It might be better to tackle that post-v3.

tests/unit/test_config.py
joe4dev: nice idea 💡 
Done

tests/unit/test_cors.py
joe4dev: yeah, valid point. I would try to avoid localstack<=>localstack-ext refactorings within the scope of the release branch ...

joe4dev:
> Also a change I missed when updating GATEWAY_LISTEN: on [`config.py` line `671`](https://github.com/localstack/localstack/pull/9405/files#diff-8ee48b2bcad465473db1a1b202781ca2c7c23db7bd1670e79d45668fb2bd718fR671) the type signature of `GATEWAY_LISTEN` should be `UniqueHostAndPortList`, not `List`. Is it ok to update this? (Or remove the dedicated type signature, since it should come from the signature of `populate_legacy_edge_configuration`.

I removed the typing. It seems clearer to rely on the correct signature from the function, especially given the tuple. The single typing appears rather confusing.

simonrw (approved):
Awesome stuff, nice work! I'm super happy to be migrating fully over to `GATEWAY_LISTEN` and it's more straightforward to have one variable than the three we used to have!

joe4dev:
FYI: I rebased to resolve conflicts (mainly in `config.py`)

joe4dev:
New conflicts streaming in 🏃 => new rebase

localstack/config.py
```diff
@@ -1332,12 +1312,7 @@ def get_edge_url(localstack_hostname=None, protocol=None):
 
 
 def edge_ports_info():
-    if EDGE_PORT_HTTP:
-        result = "ports %s/%s" % (EDGE_PORT, EDGE_PORT_HTTP)
-    else:
-        result = "port %s" % EDGE_PORT
-    result = "%s %s" % (get_protocol(), result)
-    return result
+    return f"{get_protocol()} port {GATEWAY_LISTEN[0].port}"
```
dfangl, line 1315: Not sure what this method exactly is for, but shouldn't we print all ports of GATEWAY_LISTEN here, not only the first?

dfangl (approved):
Love it, finally getting rid of the confusing EDGE_PORT_HTTP :D

localstack/config.py
joe4dev: It seems used for log messages, hence adding all ports makes sense.
Done 👍

PR #9406, opened 2023-10-19
add DHCPOptions resource provider

pinzon:
## Motivation
The EC2 service supports the resource but there is no support in CFn.  

## Changes
New resource provider with test

## Testing
The test validates the simple creation of the resource through CFn


simonrw (approved):
Nice work, nice clean implementation and I like the test. Using the `Ref` intrinsic ensures that the physical resource id was set correctly, and the use of `describe_dhcp_options` nicely tests the created resource, very cool!

PR #9410, opened 2023-10-20
enhancement: dynamodb performance optimization

bentsku:
## Motivation
Ongoing work to optimize DynamoDB performance.

## Changes
- remove the check for Lambda ESM:
  - this is actually not needed, because lambda can only subscribe to a dynamodb stream, [which is what is passed to `CreateEventSourceMapping`](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.Tutorial.html#Streams.Lambda.Tutorial.CreateTrigger). Our check was passing the table ARN, when it should have been the stream ARN, so it never actually returned a result. We have a better in-memory internal way to check if the table has a stream attached, so we should use this (there's actually no public API for DynamoDBStreams).
- optimize the `BatchWriteItem` operation. We would do operations that would have needed to be done only if a stream was attached all the time. Changed those operations to be done only if needed. Refactored the code. (We should still optimize it later on with some batch read, but the priority is performance when no streams are attached).

## Testing

Benchmarks were done and show that an empty provider implementation with only forwarding is as fast as this PR, which provides:
- 1.85x increase in throughput for `PutItem`
- 8.25x increase for `BatchWrite` with batch size = `10`
- 12x increase for `BatchWrite` with batch size = `25`

Small results from the benchmark with LocalStack running in Docker on macOS:
- put 500 items with PutItem = 10.55s -> ~5.5 +-0.1
- put 5000 items with BatchWriteItem = 62s -> 7.6s (to put it in perspective, reducing the time to seed 600k records from 2hrs to 15min)
- put 12500 items with BatchWriteItem = 120s -> 10.2s (which further reduces it from 1h36 to 8min)

## TODO

What's left to do:

- [x] write AWS validated tests to validate the DDB Streams forwarding changes
- [x] write more test around BatchItemWrite to validate behaviour changes

_related issue #1205_

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -44,6 +44,7 @@ from localstack.aws.api.dynamodb import (
     DescribeKinesisStreamingDestinationOutput,
     DescribeTableOutput,
     DescribeTimeToLiveOutput,
+    DestinationStatus,
     DynamodbApi,
     ExecuteStatementInput,
     ExecuteStatementOutput,
@@ -917,8 +918,7 @@ class DynamoDBProvider(DynamodbApi, ServiceLifecycleHook):
                         unprocessed_items_for_table.append(request)
 
                     elif stream_enabled:
-                        # TODO: we should at minimum BatchRead those values instead of individually checking them
-                        #  also, BatchWrite does not update but overwrite, so verify if we should check the existence
+                        # TODO: we should BatchRead those values instead of individually checking them
                         item = ItemFinder.find_existing_item(
                             inner_request, table_name, context.account_id, context.region
                         )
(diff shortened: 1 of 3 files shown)


localstack/services/dynamodb/provider.py
```diff
@@ -1739,8 +1722,10 @@ def get_updated_records(table_name: str, existing_items: List) -> List:
           into the PartiQL query execution inside DynamoDB Local and directly extract the list of updated items.
     """
     result = []
+    # TODO: remove get_region(), but the logic is all around dynamodb
+    region = aws_stack.get_region()
     stream_spec = dynamodb_get_table_stream_specification(
-        account_id=get_aws_account_id(), region_name=aws_stack.get_region(), table_name=table_name
+        account_id=get_aws_account_id(), region_name=region, table_name=table_name
     )
```
viren-nadkarni, line 1729: With https://github.com/localstack/localstack/pull/9382 merged, you should use `account_id` and `region_name` available as function args

viren-nadkarni (approved):
The performance improvements are fantastic, nice one!

localstack/services/dynamodb/provider.py
bentsku: of course yes, this was about the `get_updated_records` method, but I've updated the signature and passed the context account and region instead, as it's used in only one place so not a lot of refactoring. Good luck for dynamodb, it is really everywhere there 😬

bentsku: It's now rebased so it properly uses the context! Quite the rebasing but it does feel better 😄

PR #9411, opened 2023-10-20
fix inter services IAM for S3 Notifications

bentsku:
## Motivation
Follow up from  #9370, to add proper region support to S3 notifications, I've added the region to the client's creation, but also added the account id as done in #9023 for sending notifications. I'll trigger an -ext run on this PR and link it here. Sorry!

https://github.com/localstack/localstack-ext/actions/runs/6582528350

https://github.com/localstack/localstack-ext/runs/17885245621
S3 notifications are fixed, only appsync remains

## Changes
Remove the account id from the client creation to trigger test event.

Also removed the account id from the client's creation to send to SQS/Lambda/SNS. But this might create issue again with multi account? \cc @viren-nadkarni why was this change made in the first place?

I believe that SNS/Lambda/SQS will retrieve the account id from the ARN passed and not from the credentials of the request to then retrieve the resource in their store, is that right?

### DIFF
--- a/localstack/services/s3/notifications.py
+++ b/localstack/services/s3/notifications.py
@@ -492,9 +492,7 @@ class SqsNotifier(BaseNotifier):
         queue_arn = config["QueueArn"]
 
         parsed_arn = parse_arn(queue_arn)
-        sqs_client = connect_to(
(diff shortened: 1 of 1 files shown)


dfangl (approved):
LGTM! Thanks for fixing it :)

PR #9414, opened 2023-10-20
fixed support for multi-region for cloudfomation

macnev2013:
## Motivation
This PR fixes for multi account support for CloudFormation. It fixed all the failing test cases of CloudFormation.

## Changes
CloudFormation was using `INTERNAL_AWS_ACCESS_KEY_ID` for resource context - which was causing the providers to create the resources in the default account which is `000000000000`. With this change, CloudFormation now uses the account id passed from the parent function.

## Testing
- In order to test this PR I've update the `test_account` and `test_access_key_id`  to `000000000001` and changed the default test region to `us-west-1`.


pinzon (approved):
Great!

PR #9416, opened 2023-10-20
add multi account support to failing tests in apigateway-sqs integration

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, sns tests should still create the consequent resources in this accounts and region.

## Changes
This PR fixes `tests.aws.services.apigateway.test_apigateway_sqs` test cases when using non-default region and account id by:
- fixing `create_sqs_queue`, adding `region_name` and `aws_access_key_id` to the `connect_to()` client. 
- use queue arn for various sqs operations, this helps us fetching correct region and account id from arn instead of falling back to the default values. 
- remove `create_sqs_queue` in `localstack/utils/aws/resources.py`.

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_sqs.py
+++ b/tests/aws/services/apigateway/test_apigateway_sqs.py
@@ -7,7 +7,7 @@ import requests
 from localstack.constants import TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME
 from localstack.services.apigateway.helpers import connect_api_gateway_to_sqs, path_based_url
 from localstack.testing.pytest import markers
-from localstack.utils.aws import queries
+from localstack.utils.aws import arns, queries
 from localstack.utils.aws import resources as resource_util
 from localstack.utils.strings import short_uid, to_str
 from localstack.utils.sync import retry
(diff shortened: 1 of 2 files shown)


tests/aws/services/apigateway/test_apigateway_sqs.py
```diff
@@ -19,15 +19,20 @@
...
-    # create target SQS stream
+    # create target SQS stream'
     queue_name = f"queue-{short_uid()}"
-    resource_util.create_sqs_queue(queue_name)
+    queue_arn = arns.sqs_queue_arn(
+        queue_name, region_name=TEST_AWS_REGION_NAME, account_id=TEST_AWS_ACCOUNT_ID
+    )
+    resource_util.create_sqs_queue(
+        queue_name, account_id=TEST_AWS_ACCOUNT_ID, region_name=TEST_AWS_REGION_NAME
+    )
```
bentsku, line 29: this should use a fixture instead (`sqs_create_queue`), so we actually could remove the util in `aws/resources`.

tests/aws/services/apigateway/test_apigateway_sqs.py
```diff
@@ -19,15 +19,20 @@
 
 @markers.aws.unknown
 def test_api_gateway_sqs_integration(aws_client):
-    # create target SQS stream
+    # create target SQS stream'
     queue_name = f"queue-{short_uid()}"
-    resource_util.create_sqs_queue(queue_name)
+    queue_arn = arns.sqs_queue_arn(
+        queue_name, region_name=TEST_AWS_REGION_NAME, account_id=TEST_AWS_ACCOUNT_ID
+    )
```
bentsku, line 26: there's a fixture to get the SQS queue ARN: `sqs_get_queue_arn`

bentsku (changes requested):
Thanks for tackling this! I think this issue could be fixed by using the pytest fixtures instead of LocalStack utils! Then one day we might be able to validate the test 😄

bentsku (approved):
LGTM! Awesome, thanks for fixing it!

PR #9418, opened 2023-10-20
Cfn: Add account ID and region awareness to Stacks and StackChangeSets

viren-nadkarni:
This PR ensures that CloudFormation Stacks and StackChangeSets have the correct ARNs when used in multi-account and multi-region setups.


dominikschubert (approved):
LGTM, looks innocent enough :+1:

Issue #9393, opened 2023-10-18
enhancement request: Add support for AWS::EC2::TransitGatewayAttachment


pinzon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

The support for this resource is missing

### 🧑‍💻 Implementation

The implementation of a new resource provider

pinzon:
The solution has been merged.

PR #9424, opened 2023-10-20
add resource providers for EC2::TransitGateway and EC2::TransitGatewayAttachment

pinzon:
Motivation
The EC2 service supports the resources but there is no support in CFn. Solves #9393 

Changes
New resource providers and test

Testing
The test validates the simple creation of the resources through CFn and validates the creation in the EC2 service


localstack/services/ec2/resource_providers/aws_ec2_transitgateway.py
```diff
@@ -0,0 +1,145 @@
...
+
+        IAM permissions required:
+          - ec2:DeleteTransitGateway
+          - ec2:DeleteTags
+        """
+        model = request.desired_state
+        request.aws_client_factory.ec2.delete_transit_gateway(TransitGatewayId=model["Id"])
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model={},
```
simonrw, line 129: I don't understand how this does not fail any tests. We look at the properties of the `resource_model` on execution of the function and extract fields from it. Can you double check that there are no messages in the logs about "failed to delete..." or similar?

pinzon: I extended the test to assert the deletion. It seems that for the TransitGateway, AWS and moto act different but no error was thrown during the deletion.

dominikschubert (approved):
LGTM! thanks for adding them :pray:

Issue #1205, opened 2019-03-18
dynamodb performance (dynalite is waaaaay faster)


mvayngrib:
guys, thanks for building this wonderful project, we've been using it almost since its inception!

that said, we've struggled with performance for a while now, and recently I tried to see which service out of the ones we use (s3,sns,apigateway,iam,sts) was responsible. It was dynamodb. A put() takes up to half a second. 

After swapping in an image of dynalite, dynamodb calls became 3-10x times faster (!). What was the original reason for switching to dynamodb local?

my setup:

localstack: 0.8.7 and 0.9.0 (dockerized)
os: macOS 10.14.2
docker for mac: 2.0.0.3

dynalite image built from the one in this [PR](https://github.com/mhart/dynalite/pull/104) to dynalite

┆Issue is synchronized with this [Jira Bug](https://localstack.atlassian.net/browse/LOC-91) by [Unito](https://www.unito.io/learn-more)

KyleBS:
I'm struggling with similar performance problems with LocalStack + DynamoDB. However, I'm not convinced that the problem is with DynamoDB Local alone. 

When I ran tests locally going through LocalStack wrapping DynamoDB Local, I saw per-API high variances in latency, sometimes exceeding 1s. When I took LocalStack out of the equation and used DyanmoDB Local directly these latencies were consistently in the low tens of millis.

mvayngrib:
@KyleBS good find! 
@whummer do you know what might be the cause?

whummer:
Thanks for reporting @mvayngrib @KyleBS . Has this been fixed in the meantime, or is DynamoDB performance still an issue? Do you think you could provide a small script to run some tests and make the performance numbers reproducible? Thanks!

awilkins:
Not op but : it's very noticeable.

e.g. we have a script that calls the command line client to set up the schema. This completes almost immediately when directly using the DynamoDB local docker image. It takes around 10s via localstack.

dwikle:
I have also noticed significant performance differences between DynamoDB in localstack and the native DynamoDB Local offering from AWS.

I don't have any hard numbers but from the experiments I ran I observed the following:
- running local jar from AWS: fast
- running `amazon/dynamodb-local` docker image: fast
- running via localstack: slow
- running via localstack but with larger `MAX_HEAP_SIZE` in `dynamodb_starter.py`: slow

[...]

PR #9425, opened 2023-10-21
implement AwsRequestProxy that re-uses HTTP client connections

thrau:
## Motivation

This PR implements one of the optimizations listed in [this comment](https://github.com/localstack/localstack/issues/1205#issuecomment-1223999205) for improved dynamodb performance.
Previously we were creating a new `Proxy` instance for every `forward_request` call, and therefore never re-using HTTP client connections, which is bad for throughput.

This PR introduces the `AwsRequestProxy` which is a higher-level proxy that satisfies the `ServiceRequestHandler` protocol, and builds on a `Proxy` instance.

This also addresses some comments originally raised in #5796 such as [this one](https://github.com/localstack/localstack/pull/5796#discussion_r842067952).

## Changes

* Introduce `AwsRequestProxy` as high-level proxy for ASF requests to AWS backends
* Change DynamoDB to use one proxy instance per `DynamoDBServer` instance, rather than instantiating one on every request.

## Some preliminary results

I ran a normal `PutItem` loop from [this benchmark](https://github.com/localstack/localstack/issues/1205#issuecomment-1223999205) with 200 calls. I ran each experiment several time and took the median result. I also tested the changes with #9410.

| description | result |
| --- | --- |
| without connection re-use | 200 requests took 3.4139s = 17.0693 ms/op = 58.58 req/sec |
| with connection re-use | 200 requests took 3.0002s = 15.0009 ms/op = 66.66 req/sec |
| **With https://github.com/localstack/localstack/pull/9410** | | 
| without connection re-use | 200 requests took 2.3314s = 11.6570 ms/op = 85.79 req/sec |
| with connection re-use | 200 requests took 2.1307s = 10.6533 ms/op = 93.87 req/sec |

So it adds another 10% performance improvement on top of Ben's optimization, at least for the simple `PutItem` case.


giograno (approved):
Nice optimization 🏅

bentsku (approved):
LGTM! This is actually awesome, love the idea and thanks a lot for tackling this, this is the kind of high level construct we were lacking and I didn't have the vision for 😄 We can see it sheds 1ms per request, not sure why 2ms before the optimization though. 

Anyway, this is a great change, I love all the constructs we have now for different use cases. Thank you!

viren-nadkarni (approved):
Nice one, we could now consider looking into connection reuse for other services like Kinesis :rocket:

Issue #9427, opened 2023-10-21
bug: ListObjectsV2 Returns EncodingType When Not Specified


tustvold:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
$ docker run -d -p 4566:4566 -e PROVIDER_OVERRIDE_S3=v3 localstack/localstack:2.3

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 mb s3://tustvold

$ curl -X PUT --data-binary 'test' -H "Content-Type: text/plain" 'http://localhost:4566/tustvold/a%252Fb.txt'

$ curl 'http://localhost:4566/tustvold'
<?xml version='1.0' encoding='utf-8'?>
<ListBucketResult><IsTruncated>false</IsTruncated><Marker /><Name>tustvold</Name><Prefix /><MaxKeys>1000</MaxKeys><EncodingType>url</EncodingType><Contents><Key>a%252Fb.txt</Key><ETag>"d41d8cd98f00b204e9800998ecf8427e"</ETag><Owner><DisplayName>webfile</DisplayName><ID>75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a</ID></Owner><Size>0</Size><LastModified>2023-10-21T18:45:37.298607Z</LastModified><StorageClass>STANDARD</StorageClass></Contents></ListBucketResult>
```

Using `PROVIDER_OVERRIDE_S3=v3` does not appear to change this behaviour, nor does it differ between ListObjects and ListObjectsV2

### Expected Behavior

```
$ docker run -d -p 4566:4566 localstack/localstack:2.0

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 mb s3://tustvold

$ curl -X PUT --data-binary 'test' -H "Content-Type: text/plain" 'http://localhost:4566/tustvold/a%252Fb.txt'

$ curl http://localhost:4566/tustvold/a%252Fb.txt
test

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 ls s3://tustvold
2023-10-21 19:43:04          4 a%2Fb.txt

$ curl 'http://localhost:4566/tustvold'
<?xml version='1.0' encoding='utf-8'?>
[...]

PR #9430, opened 2023-10-22
fix s3 v3 ListObjectsV2 and ListObjectVersions pagination

bentsku:
## Motivation
As reported in https://github.com/localstack/localstack/issues/9427#issuecomment-1773934333
> one thing I did notice was ListObjectsV2 startAfter appeared to be including the startAfter key, when it should be exclusive

After validating against AWS and writing some more tests which we didn't cover, we were not sorting keys properly (not leaving at the right moment in the loop).

This is based on #9429 as both are modifying the behaviour and I didn't want the merge conflict. 

## Changes
Fixed the behaviour to properly match AWS and wrote a good amount of tests to validate this. Took the opportunity to validate and introduce the behaviour for `ListObjectVersions` pagination, which is not done in moto yet but is work in progress.

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -1393,8 +1393,11 @@ class S3Provider(S3Api, ServiceLifecycleHook):
     ) -> ListObjectsV2Output:
         store, s3_bucket = self._get_cross_account_bucket(context, bucket)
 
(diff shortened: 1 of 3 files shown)


macnev2013 (approved):
Thanks for the PR. LGTM 👍🏻

Issue #9428, opened 2023-10-21
bug: DeleteObjects Percent Decoding Input


tustvold:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
$ docker run -d -p 4566:4566 localstack/localstack:2.3

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 mb s3://tustvold
make_bucket: tustvold
```

If we then create a path with a percent encoded object path, deleting it works correctly
 
```
$ curl -X PUT --data-binary 'test' -H "Content-Type: text/plain" 'http://localhost:4566/tustvold/a%252Fb.txt'

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 ls s3://tustvold
2023-10-21 21:33:50          4 a%2Fb.txt

$ aws s3api delete-objects --no-sign-request --endpoint http://localhost:4566 --bucket tustvold --delete 'Objects=[{Key=a%2Fb.txt}]'
{
    "Deleted": [
        {
            "Key": "a%2Fb.txt"
        }
    ]
}
```

However, if the object key is a percent encoded emoji, we get very peculiar behaviour, where the DeleteObject appears to be trying to delete the percent decoded version of the object

```
$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 ls s3://tustvold
$ curl -X PUT --data-binary 'test' -H "Content-Type: text/plain" 'http://localhost:4566/tustvold/a/%25F0%259F%2598%2580.file'

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 ls s3://tustvold --recursive
2023-10-21 21:36:07          4 a/%F0%9F%98%80.file

$ aws s3api delete-objects --no-sign-request --endpoint http://localhost:4566 --bucket tustvold --delete 'Objects=[{Key=a/%F0%9F%98%80.file}]'
{
    "Deleted": [
        {
            "Key": "a/😀.file"
        }
    ]
}
$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 ls s3://tustvold --recursive
2023-10-21 21:36:07          4 a/%F0%9F%98%80.file
```

### Expected Behavior

In version 2.0 this worked correctly

```
$ docker run -d -p 4566:4566 localstack/localstack:2.0

$ aws --no-sign-request --endpoint-url=http://localhost:4566 s3 mb s3://tustvold

[...]

PR #9431, opened 2023-10-22
add test for DeleteObjects encoding

bentsku:
## Motivation
As reported in #9428, there's an issue with the URL encoding of S3 object keys. This is an issue in moto, and it's working properly in our `v3` provider.

I have a PR open here: https://github.com/getmoto/moto/pull/6933 but there's still some issue. This PR would only be green once the upstream fix is part of moto-ext, so I will `xfail` it for `v2` until the PR is merged as it's a big change (it requires moto to use `RAW_URI` as well and change how to handle raw path with the update of werkzeug, they are manually re-encoding the path and it has some drawbacks for some edge cases like emoji), knowing that `v3` is going to become default soon so I'd like to have it tested already now.

## Changes
Only introducing tests to validate the proper behaviour.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -3288,6 +3288,7 @@ class TestS3:
                 ],
             },
         )
+        response["Deleted"].sort(key=itemgetter("Key"))
         snapshot.match("deleted-resp", response)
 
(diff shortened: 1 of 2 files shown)


macnev2013 (approved):
LGTM 💯

PR #9435, opened 2023-10-23
apigateway: add multi-account and region capability

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, tests should still create the consequent resources in this accounts and region.

## Changes
This PR fixes all the failing tests in `tests/aws/services/apigateway/test_apigateway_extended.py` by adding region and account id to the helper functions.

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -1836,8 +1836,8 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
             export_type=export_type,
             export_format=accepts,
             with_extension=has_extension,
-            region_name=context.region,
             account_id=context.account_id,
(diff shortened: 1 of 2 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -1836,6 +1836,8 @@ def get_export(
             export_type=export_type,
             export_format=accepts,
             with_extension=has_extension,
+            region_name=context.region,
+            account_id=context.account_id,
```
viren-nadkarni, line 1840: Nitpick: to keep the codebase consistent, prefer having the `account_id` arg before the `region_name`.

bentsku (approved):
LGTM! Thanks for taking care of this 🚀

PR #9436, opened 2023-10-23
fix ASF update action if there are no changes

alexrashed:
## Motivation
With https://github.com/localstack/localstack-ext/pull/2217, we changed to `ruff` which replaces a bunch of tools (`flake8`, `isort`, `autoflake`). Previously, these tools were directly called in the ASF scaffold and the result of the code generation was already perfectly formatted.
`ruff` cannot be called from Python code on in-memory strings, which is why we switched to our `format-modified` `make` target.
Turns out, the target does not work in case there are no changes (since then black does not get any input).
This PR slightly modifies the ASF update action such that this issue is prevented.

## Changes
- Changes the ASF Update action such that the `format-modified` target is only executed if there are changes in the repo.

## Testing
- [x] We will execute the action from this branch. Since there currently no changes, this change is safe if the action run is successful.
  - https://github.com/localstack/localstack/actions/runs/6610132434/job/17951451836 💚

### DIFF
--- a/.github/workflows/asf-updates.yml
+++ b/.github/workflows/asf-updates.yml
@@ -43,7 +43,6 @@ jobs:
         run: |
           source .venv/bin/activate
           python3 -m localstack.aws.scaffold upgrade
-          make format-modified
 
       - name: Check for changes
         id: check-for-changes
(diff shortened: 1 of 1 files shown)


dfangl (approved):
Good catch!

PR #9441, opened 2023-10-23
Rebase release branches once a day instead of on every push

dfangl:
## Motivation
The automatic rebase of the release branches is useful, although can get very noisy if we rebase on every commit on master.

With limiting the rebase to once a day (on weekdays), we can limit unnecessary CI spending, while still keeping the release branch reasonable up to date. If someone needs it more current, I also added a workflow dispatch trigger.

Once the change is discussed here, I will do the same in ext.

## Changes
* Remove push event to trigger workflow
* Add workflow_dispatch and a schedule which runs the workflow Monday-Friday at 5am UTC

### DIFF
--- a/.github/workflows/rebase-release-prs.yml
+++ b/.github/workflows/rebase-release-prs.yml
@@ -1,8 +1,8 @@
 name: Rebase Release PRs
 on:
-  push:
-    branches:
-    - master
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 * * MON-FRI
 jobs:
   find-release-branches:
     runs-on: ubuntu-latest


.github/workflows/rebase-release-prs.yml
```diff
@@ -1,8 +1,8 @@
 name: Rebase Release PRs
 on:
-  push:
-    branches:
-    - master
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 * * MON-FRI
```
alexrashed, line 5: Hmm, according to the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule), the schedule can be defined using a [POSIX cron schedule](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07).
I think this standard does not contain weekday names, but only the index (0-6, where 0 is Sunday).
So I think you would have to change this to:
```suggestion
    - cron: 0 5 * * 1-5
```

dfangl: Further down in the document it gives an example using SUN-SAT. Our ASF workflow also uses MON. Happy to change it if you prefer it, but it would surprise me if it was an issue.

alexrashed: Oh, no, not at all! If it works, this is the cleaner / more readable way of defining the schedule! 👍🏽

PR #9442, opened 2023-10-23
Feat/cloudwatch add provider switch

baermat:
## Motivation
Add a switch for the new CloudWatch provider v2

## Changes
switch for the new CloudWatch provider

## TODO

What's left to do:

- [ ] add suite for tests


steffyP (approved):
Nice 👍 this will enable testing the v2 provider. 
IMO we can merge this on our feature branch already, and add the test suite afterwards. 
We will probably add names for `v1`, `asf` etc later on, to follow the same pattern other services supporting different providers currently have.

Anyhow, we need to rebase the `feat-cloudwatch-v2` after merging this PR, because I think in the meantime we switched to ruff.

PR #9444, opened 2023-10-23
Fix LS container DNS configuration

simonrw:
## Motivation

When on a support call with a customer, we noticed that the DNS server run by LocalStack was functioning and configured correctly, but running `curl -v s3.amazonaws.com` returned a public IP address, not `127.0.0.1` - expected because of transparent endpoint injection. When I investigated further, their `/etc/resolv.conf` contained `nameserver 127.0.0.11`.

We perform a check in `localstack.dns.server.add_resolv_entry` that if the `/etc/resolv.conf` contains `nameserver 127.0.0.1` then we don't update the file. `nameserver 127.0.0.11` _does_ contain `nameserver 127.0.0.1` and so I believe this is the source of the error.

## Changes

* We only call the `add_resolv_entry` in one place, and only in the docker container, and so we can rework the function to be less conditional. We _always_ update the file.
* We ensure that we are being executed from within docker, and not on the host.

## Testing

* start localstack with `LOCALSTACK_HOST=example.com localstack start`
* `docker exec -it localstack_main curl -v example.com` should show the IP address of `127.0.0.1`

### DIFF
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -793,8 +793,13 @@ def add_resolv_entry(file_path: Path | str = Path("/etc/resolv.conf")):
     """
     )
     file_path = Path(file_path)
-    with file_path.open("w") as outfile:
-        print(content, file=outfile)
+    try:
+        with file_path.open("w") as outfile:
+            print(content, file=outfile)
+    except Exception:
+        LOG.warning(
+            "Could not update container DNS settings", exc_info=LOG.isEnabledFor(logging.DEBUG)
+        )
 
 
 def setup_network_configuration():
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -3,10 +3,12 @@ import copy
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


localstack/dns/server.py
```diff
@@ -779,29 +779,22 @@ def get_available_dns_server():
...
+    LOG.debug("Overwriting container DNS server to point to localhost")
+    content = textwrap.dedent(
+        """
+    # The following line is required by LocalStack
+    nameserver 127.0.0.1
+    """
+    )
+    file_path = Path(file_path)
+    with file_path.open("w") as outfile:
+        print(content, file=outfile)
```
dfangl, line 797: What if we are not running as root in the container? Probably sudo would not work then as well, but we should consider it, at least to handle the error correctly.

simonrw: Sure, that makes sense, thanks. I'll add an exception handler to at least catch any error that occurs.

PR #9446, opened 2023-10-23
Simplify localstack init user config

joe4dev:
Depends on https://github.com/localstack/lambda-runtime-init/pull/26

## Motivation

This PR simplifies the user's configuration in the Lambda init Go binary. We drop root privileges in the Lambda init Go binary by default and offer a configuration variable to skip this parity feature. However, it is cumbersome to configure using a misleading empty string `""`. We skip dropping privileges when `LAMBDA_INIT_USER` is set to `root` (`""` is still supported but discouraged).
This is much easier to configure and leads to less errors around escaping quotes etc.

## Changes

* Bump Lambda init version skipping privileges drop with `root`
* Fix the LocalStack configuration logic for `""`

### DIFF
--- a/localstack/services/lambda_/invocation/execution_environment.py
+++ b/localstack/services/lambda_/invocation/execution_environment.py
@@ -155,8 +155,9 @@ class ExecutionEnvironment:
             env_vars.update(self.function_version.config.environment)
         if config.LAMBDA_INIT_DEBUG:
             # Disable dropping privileges because it breaks debugging
-            env_vars["LOCALSTACK_USER"] = ""
-        if config.LAMBDA_INIT_USER:
+            env_vars["LOCALSTACK_USER"] = "root"
+        # Forcefully overwrite the user might break debugging!
+        if config.LAMBDA_INIT_USER is not None:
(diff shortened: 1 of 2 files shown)


dfangl (approved):

localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -155,8 +155,9 @@ def get_environment_variables(self) -> Dict[str, str]:
             env_vars.update(self.function_version.config.environment)
         if config.LAMBDA_INIT_DEBUG:
             # Disable dropping privileges because it breaks debugging
-            env_vars["LOCALSTACK_USER"] = ""
-        if config.LAMBDA_INIT_USER:
+            env_vars["LOCALSTACK_USER"] = "root"
+        # Forcefully overwrite the user might break debugging!
+        if config.LAMBDA_INIT_USER is not None:
```
joe4dev, line 160: @LS devs: Just be careful when using `LAMBDA_INIT_USER` in combination with init debugging.

We could also consider flipping this 🤷

PR #9448, opened 2023-10-23
fix s3 v3 preconditions

bentsku:
## Motivation
Using the `object_store` rust crate test suite, I've spotted an issue in our handling of preconditions for v3.
Basically, the `LastModified` field on the S3 object is stored with millisecond precision for internal use like sorting when calling `ListObjects*`. 

But when comparing against `ModifiedSince`/`UnmodifiedSince` preconditions headers, those are sent using second precision, and we need to check for strict equality/superiority/inferiority. So something like the following should not be true if the precision is second:
`last_modified=datetime.datetime(2023, 10, 23, 16, 19, 41, 798495, tzinfo=zoneinfo.ZoneInfo(key='GMT')) > if_unmodified_since=datetime.datetime(2023, 10, 23, 16, 19, 41, tzinfo=datetime.timezone.utc)`

## Changes
Adapted the check by rounding last modified before comparison, and created 2 AWS tests.

### DIFF
--- a/localstack/services/s3/utils.py
+++ b/localstack/services/s3/utils.py
@@ -936,6 +936,8 @@ def validate_failed_precondition(
     :raises NotModified, 304 with an empty body
     """
     precondition_failed = None
+    # last_modified needs to be rounded to a second so that strict equality can be enforced from a RFC1123 header
+    last_modified = last_modified.replace(microsecond=0)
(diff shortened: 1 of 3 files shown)


macnev2013 (approved):
LGTM 👍🏻

Issue #9394, opened 2023-10-18
bug: waiter for nested stack is too short


pinzon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

if a nested stack takes too much time to deploy the parent stack fails to deploy and it never reaches a CREATE_COMPLETE status.

### Expected Behavior

The template deployer should have a better tolerance with nested stacks

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

Use a nested stack that deploys several lambdas or custom resources 

### Environment

```markdown
- OS:Manjaro
- LocalStack: 2.3.2
```

pinzon:
The solution to this issue has been merged. Thank you for reporting it.

PR #9449, opened 2023-10-23
add AWS::CloudFormation::Stack resource provider

pinzon:
## Motivation
The generic base model of  this type implements a waiter to finish the deployment of the resource. This Waiter by default has a timer of 1 hour (30 seconds delay with 120 max attempts). For CFn this not correct due to the API indicating that if the stack has no timeout it shouldn't failed because of that. Recently a customer encountered a situation where the nested stack exceeded this limit and just extending the limit is not the correct solution. 

Solves: #9394
  

## Changes
New resource provider

## Testing
It's a migrations so it's fine as long the previous tests pass.


localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
```diff
@@ -0,0 +1,198 @@
...
+            )
+
+        descriptions = request.aws_client_factory.cloudformation.describe_stacks(
+            StackName=model["Id"]
+        )
+        if not descriptions.get("Stacks"):
+            raise Exception("Stack not found")
+
+        stack = descriptions["Stacks"][0]
+        if stack["StackStatus"] == "CREATE_COMPLETE":
```
dominikschubert, line 112: nit: would be a bit clearer here using a match statement

There are technically also stack states that aren't explicitly "CREATE_FAILED" that should still return a failed operation status, e.g. rollbacks. Would be great if we could change this part to be complete in regards to all possible returned states.

localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
```diff
@@ -0,0 +1,198 @@
...
+                status=OperationStatus.IN_PROGRESS,
+                resource_model=model,
+                custom_context=request.custom_context,
+            )
+
+        descriptions = request.aws_client_factory.cloudformation.describe_stacks(
+            StackName=model["Id"]
+        )
+        if not descriptions.get("Stacks"):
+            raise Exception("Stack not found")
```
dominikschubert, line 109: The DescribeStacks API actually returns an exception here and doesn't just return an empty array. At least it should according to parity with AWS.

localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
```diff
@@ -0,0 +1,198 @@
...
+                resource_model=model,
+                custom_context=request.custom_context,
+            )
+
+        descriptions = request.aws_client_factory.cloudformation.describe_stacks(
+            StackName=model["Id"]
+        )
+
+        if not descriptions.get("Stacks"):
+            raise Exception("Stack not found")
```
dominikschubert, line 168: same as above.

The delete operation should also be idempotent, i.e. a not found exception should still lead to a success.

localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
```diff
@@ -0,0 +1,198 @@
...
+        if stack["StackStatus"] == "DELETE_FAILED":
+            return ProgressEvent(
+                status=OperationStatus.FAILED,
+                resource_model=model,
+                custom_context=request.custom_context,
+            )
+        return ProgressEvent(
+            status=OperationStatus.IN_PROGRESS,
+            resource_model=model,
+            custom_context=request.custom_context,
```
dominikschubert, line 186: nit: same as above, I'd suggest refactoring this to a match  and defaulting to failed instead of in progress

dominikschubert (changes requested):
A few nits, then good to go :+1:

dominikschubert (approved):
LGTM! Thanks for addressing the nits :rocket:

PR #9454, opened 2023-10-24
SQS: Introduce 'standard' queue URL strategy

viren-nadkarni:
## Motivation

By default, LocalStack uses the `off` endpoint strategy for SQS. This results in SQS queue URLs which do not contain region in them (e.g. `localhost:4566/<account_id>/<queue_name>`). This hinders out-of-the box multi-region experience for LocalStack users. If a user creates a queue in any region other than `us-east-1` and tries the SQS Query API, it results in QueueNotFound errors.

The `domain` strategy solves this problem but really it generates [legacy style endpoints](https://docs.aws.amazon.com/general/latest/gr/sqs-service.html#sqs_region).

## Implementation

This PR aims to solve the above problems. It introduces a new endpoint strategy `standard`. The new strategy produces SQS queue URLs that resemble AWS as closely as possible.

The default strategy will be changed in a https://github.com/localstack/localstack/labels/semver%3A%20major PR: https://github.com/localstack/localstack/pull/9468

AWS 'modern' queue URLs: 
`https://sqs.us-east-1.amazonaws.com/123456789012/my-queue`

LocalStack `standard` strategy queue URLs: 
`http://sqs.eu-central-1.localhost.localstack.cloud:443/000000000000/my-queue`

## Tests

Updates existing parametrised tests to use the new strategy. Also adds a new focused test.

## Related

https://github.com/localstack/localstack-artifacts/pull/20
https://github.com/localstack/docs/pull/877


alexrashed (commented):
Nice set of changes! This now enables full region support for SQS, and increases the parity with the domains.
The code is looking great, but I think this PR depends on https://github.com/localstack/localstack-artifacts, right?

viren-nadkarni:
> The code is looking great, but I think this PR depends on https://github.com/localstack/localstack-artifacts, right?

@alexrashed: Correct! Would you be kind to review that PR too?

There will be a follow up PR to update the user docs too.

PR #9458, opened 2023-10-24
Update route matchers to accept any host

simonrw:
## Motivation

While configuring LocalStack, the `LOCALSTACK_HOST` variable can be used to configure domains that are returned from responses. We also have a few route matchers that match the host header of an HTTP request, which in turn dispatch to handlers depending on (usually) the request path. If the user customises the `LOCALSTACK_HOSTNAME` we should still accept their request, even if they do not use `*.localhost.localstack.cloud`.

## Changes

* Update s3 website hosting to accept any host

### DIFF
--- a/localstack/services/s3/website_hosting.py
+++ b/localstack/services/s3/website_hosting.py
@@ -18,13 +18,12 @@ from localstack.aws.api.s3 import (
 )
 from localstack.aws.connect import connect_to
 from localstack.aws.protocol.serializer import gen_amzn_requestid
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM! Nice to see we can accept more now! This is a bit hard to document, but good to know we support it.

Note: I've found 2 usage of the `S3_STATIC_WEBSITE_HOSTNAME` string when returning the URL of the S3 Website in CloudFormation that we shouldn't forget to migrate to `localstack_host` then.

PR #9459, opened 2023-10-24
[SFN] Improvements for state update during retry-catch workflows

MEPalma:
## Motivation
Currently the language interpreter for StepFunctions v2 is unable to update the memory state during retry-catch sequences. This pr addresses such limitations.

## Changes
This pr adds relevant tests to capture the desired behaviour and updates the evaluation logic for retry-catch sequences, as well as the input computation for lambda native calls.


dominikschubert (approved):
LGTM :+1:

PR #9461, opened 2023-10-24
remove s3 legacy code and usage

bentsku:
## Motivation
Following the deprecation for the S3 legacy provider, we can now remove it to allow the legacy edge proxy code to be removed as well.

## Changes
Removed all S3 legacy code and its usage, especially in conditional around tests.
I believe we're going to gain quite a lot of coverage once again 😄


localstack/services/cloudformation/models/s3.py
```diff
@@ -368,3 +363,8 @@ def fetch_state(self, stack_name, resources):
             return None
 
         return response
+
+
+# test_multipart_and_list_parts
+# test_put_object_checksum
+#
```
bentsku, line 370: comment: will remove, leftover from tests

localstack/services/s3/utils.py
alexrashed, line 1: Why are these changes necessary? They seem a bit unrelated?

alexrashed (approved):
I only found a small code change which doesn't seem related (but also doesn't seem to be harmful?). Everything else is looking great! Nice to see thousands of lines being deleted!
The coverage will go up by more than 1%! 🚀 

However, I am just a bit concerned about the number of changed files. A lot are just removals, but specifically the removal of the `LEGACY_S3_PROVIDER` causes a lot of changes in other files which could become problematic with rebases in the next few weeks. But I guess we'll see how the rebasing works... 😛

bentsku:
@alexrashed I think you're right, I was thinking about this as well.
How would you feel about removing all the conditional around the legacy provider in tests in master first? That would reduce the numbers of files by a good amount, and we don't run the tests suite with it anymore so the tests would just break when ran against it, which is no big deal.

localstack/services/s3/utils.py
bentsku: Yes, will remove, it was due to the tests refactoring/migration to keep coverage. Will move this to master instead.

alexrashed: I don't understand. If this is a change that makes sense in this context, then it can also stay here?

alexrashed:
I think the tests themselves are not really the problem, but the actual service implementations and utils code which uses the `LEGACY_S3_PROVIDER` variable. A potential way around this would be to just set it to `False` (instead of reading the env var) and basically leaving everything else. This would essentially deactivate the whole feature, with a single line of code.
Then we could remove it silently in the background in upcoming changes (after the release branch has been merged into master). However, I think it's nice to have this together in a single PR, and collect the dopamine of the removals.
So let's move forward with the changes as they are in here 🚀 🧹

localstack/services/s3/utils.py
bentsku: It was something that wasn't tested before and added now, it would also makes sense to have it in master. It's basically to have the method follow the suggested typing, but it's not a big functionality change (it was returning an empty string instead of `None`). However the change is so small and insignificant that I figured I wouldn't open a PR for it, but it's a bit outside of the scope of this PR. If you're fine with it, we can leave it here, it was let's say some house cleaning. 🧹

bentsku:
I see. Yes, deleting the entry in the `providers.py` and setting it to `False` would have made the same, but the cleaning would feel different! If it becomes too difficult to rebase until the release, we can always revert and keep it for after the release like you explained. I'll remove the leftovers comments and merge then.

Issue #9111, opened 2023-09-11
bug: Localstack does not support import a key material when key usage is SIGN_VERIFY


thiagosantosleite:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Localstack does not support import a key material which key usage is SIGN_VERIFY.
```

aws kms create-key --origin EXTERNAL --key-usage=SIGN_VERIFY --key-spec=ECC_NIST_P256

export key_id=1d0e24a5-1bfe-4099-9c81-41c3155c506b

aws kms get-parameters-for-import \
   --key-id $key_id \
   --wrapping-algorithm RSAES_OAEP_SHA_256 \
   --wrapping-key-spec RSA_4096

Error:
An error occurred (InvalidKeyUsageException) when calling the GetParametersForImport operation: arn:aws:kms:us-east-1:000000000000:key/1d0e24a5-1bfe-4099-9c81-41c3155c506b key usage is SIGN_VERIFY which is not valid for GetParametersForImport.
```

### Expected Behavior

Import SIGN_VERIFY keys import works.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

```
aws kms create-key --origin EXTERNAL --key-usage=SIGN_VERIFY --key-spec=ECC_NIST_P256

export key_id=1d0e24a5-1bfe-4099-9c81-41c3155c506b

aws kms get-parameters-for-import \
   --key-id $key_id \
   --wrapping-algorithm RSAES_OAEP_SHA_256 \
   --wrapping-key-spec RSA_4096

```

### Environment

```markdown
- OS:
- LocalStack:
```

PR #9463, opened 2023-10-25
kms: fix get_parameters_for_import

sannya-singal:
## Motivation
While importing parameters for import when key usage is `SIGN_VERIFY`, we got an `InvalidKeyUsageException` error. 

## Changes
This PR:
- adds correct error response if key origin is not `EXTERNAL` which is mandatory for this operation. 
- removes a validation for checking key usage as `ENCRYPT_DECRYPT`.
- add an aws validated test case to check the above conditions.  

Related Issue: #9111


viren-nadkarni (approved):
:rocket:

PR #9474, opened 2023-10-25
corrects status for instance profile resource provider

pinzon:
## Motivation
This resource provider is not returning the correct status after the deletion which causes depending resources to fail during the deletion of a stack. cc/ @dominikschubert 

## Changes
Returns correct event


dominikschubert (approved):
Good catch, LGTM

PR #9475, opened 2023-10-25
add sqlite for metric storage cloudwatch v2

steffyP:
## Motivation
wip, adding sqlite for storing metrics, 

started with:
* adding sqlite tables
* added `cloudwatch_database_helper` class where we will run all the queries
   * started with `add_metric_data` -> only for single value metrics
   * started `get_metric_data_stat` -> several TODOs left (see TODO section)

## TODO

What's left to do:

- [x] `put_metrics`
   - [x]  validation of input
   - [x] implement `Values`
   - [x] implement `StatisticValues`
- [x] `get_metric_data`
   - [x] include data from aggregated-metrics
   - [x] aggregate + list by period (check with moto implementation)
   - [x] handle null values
   - [x] add tests
   - [x] filter by dimension
   - [x] Paginate results

- [x] `get_metric_statistics` 
- [x] `list_metrics`
  - [x] single metrics 
  - [x] aggregated-metrics table 
  - [x] filter by dimension
  - [x] filter by dimension value

Follow-up Tasks + TODOs
- [ ] verify that the default timestamp added to the metrics is UTC time (might be local timestamp?)
- [ ] `get_raw_metrics` endpoint
- [ ] persistence -> add directory, check if data is restored
- [ ] handle different units in the metrics
- [ ] add a test where we try to insert single + static values at the same time (in one put-metric-data statement (probably will throw an error, but we need to test)
- [ ] get-metric-data:
     - [ ] null values, check if dimensions must be null though if missing
     - [ ] test default order (asc/desc timestamp for get-metric-data)
     - [ ] handle empty filters for dimensions
     - [ ] handle filters for unit
     - [ ] check if we can simplify the logic for min/max/sum/avg the data query, by selecting everything in one query something like: `SELECT from (SELECT from simpletable ... UNION select from aggregatedtable...`)
 - [ ] shutdown/clear table: placeholders currently, we should delete the table when the `reset` endpoint is called
 - [ ] add pagination tests 
 - [ ] add snapshots to the existing tests
 - [ ] remove the `time.sleep` from the tests and replace with `retry` or other utility

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -1,0 +1,91 @@
+import logging
+import os
+import sqlite3
+
+from localstack import config
+from localstack.utils.files import mkdir
+
+LOG = logging.getLogger(__name__)
+
+
+class CloudwatchDatabase:
+    DB_NAME = "metrics.db"
+    CLOUDWATCH_DATA_ROOT: str = os.path.join(config.dirs.data, "cloudwatch")
+    METRICS_DB: str = os.path.join(CLOUDWATCH_DATA_ROOT, DB_NAME)
+    TABLE_SINGLE_METRICS = "SINGLE_METRICS"
+    TABLE_AGGREGATED_METRICS = "AGGREGATED_METRICS"
+
+    def __init__(self):
+        if os.path.exists(self.METRICS_DB):
+            LOG.debug(f"database for metrics already exists ({self.METRICS_DB})")
+            return
+
+        mkdir(self.CLOUDWATCH_DATA_ROOT)
+        conn = None
+        try:
+            # TODO check if the SQLITE_THREADSAFE setting is 2 (multi-threaded) or 1 (serialized)
+            conn = sqlite3.connect(
+                self.METRICS_DB,
+                check_same_thread=False,
+            )
+            common_columns = """
+                "id"	                INTEGER,
+                "account_id"	        TEXT,
+                "region"	            TEXT,
+                "metric_name"	        TEXT,
+                "namespace" 	        TEXT,
+                "timestamp"	            NUMERIC,
+                "dimensions"	        TEXT,
+                "unit"	                TEXT,
+                "storage_resolution"	INTEGER
+            """
+            conn.execute(
+                f"""
+            CREATE TABLE "{self.TABLE_SINGLE_METRICS}" (
+                {common_columns},
+                "value"	                NUMERIC,
+                PRIMARY KEY("id")
+            );
+            """
+            )
+
+            conn.execute(
+                f"""
+            CREATE TABLE "{self.TABLE_AGGREGATED_METRICS}" (
+                {common_columns},
+                "sample_count"          NUMERIC,
+                "sum"	                NUMERIC,
+                "min"	                NUMERIC,
+                "max"	                NUMERIC,
+                PRIMARY KEY("id")
+            );
+            """
+            )
+            # create indexes
+            conn.executescript(
+                """
+            CREATE INDEX idx_single_metrics_comp ON SINGLE_METRICS (metric_name, namespace);
+            CREATE INDEX idx_aggregated_metrics_comp ON AGGREGATED_METRICS (metric_name, namespace);
+            """
+            )
+
+        except Exception as e:
+            LOG.error("Could not create sqlite database", e)
+        finally:
+            if conn:
+                conn.close()
+
+    def add_metric_data(self):
+        pass
+
+    def get_metric_data(self):
+        pass
+
+    def clear_tables(self):
+        # TODO clear tables
+        pass
+
+    def shutdown(self):
+        # TODO delete tmpdir/database if we do not have persistence enabled?
+        # anything else we should consider?
+        ...
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -1,8 +1,11 @@
 import logging
 import os
 import sqlite3
+from datetime import datetime, timezone
+from typing import Dict, List, Optional
 
 from localstack import config
+from localstack.aws.api.cloudwatch import MetricData, MetricStat, ScanBy
 from localstack.utils.files import mkdir
 
 LOG = logging.getLogger(__name__)
@@ -21,13 +24,8 @@ class CloudwatchDatabase:
             return
 
         mkdir(self.CLOUDWATCH_DATA_ROOT)
-        conn = None
-        try:
-            # TODO check if the SQLITE_THREADSAFE setting is 2 (multi-threaded) or 1 (serialized)
-            conn = sqlite3.connect(
-                self.METRICS_DB,
-                check_same_thread=False,
-            )
+        with sqlite3.connect(self.METRICS_DB, isolation_level="EXCLUSIVE") as conn:
+            cur = conn.cursor()
             common_columns = """
                 "id"	                INTEGER,
                 "account_id"	        TEXT,
@@ -39,7 +37,7 @@ class CloudwatchDatabase:
                 "unit"	                TEXT,
                 "storage_resolution"	INTEGER
             """
-            conn.execute(
+            cur.execute(
                 f"""
             CREATE TABLE "{self.TABLE_SINGLE_METRICS}" (
                 {common_columns},
@@ -49,7 +47,7 @@ class CloudwatchDatabase:
             """
             )
 
-            conn.execute(
+            cur.execute(
                 f"""
             CREATE TABLE "{self.TABLE_AGGREGATED_METRICS}" (
                 {common_columns},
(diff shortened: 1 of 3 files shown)


localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,406 @@
...
+    def __init__(self):
+        if os.path.exists(self.METRICS_DB):
+            LOG.debug(f"database for metrics already exists ({self.METRICS_DB})")
+            return
+
+        mkdir(self.CLOUDWATCH_DATA_ROOT)
+        with sqlite3.connect(self.METRICS_DB, isolation_level="EXCLUSIVE") as conn:
+            cur = conn.cursor()
+            common_columns = """
+                "id"	                INTEGER,
```
baermat, line 46: just making sure we want an integer as ID, since you usually don't perform arithmetic operations on them

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,406 @@
...
+        # anything else we should consider?
+        ...
+
+    def _get_ordered_dimensions_with_separator(self, dims: Optional[List[Dict]]):
+        if not dims:
+            return None
+        dims.sort(key=lambda d: d["Name"])
+        dimensions = ""
+        for d in dims:
+            dimensions += f"{d['Name']}={d['Value']}\t"  # aws does not allow ascii control characters, we can use it a sa separator
```
baermat, line 377: nit: typo in comment

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -214,3 +329,88 @@ def tag_resource(
...
+
+        metrics = [
+            {
+                "Namespace": metric.get("namespace"),
+                "MetricName": metric.get("metric_name"),
+                "Dimensions": metric.get("dimensions"),
+            }
+            for metric in result.get("metrics", [])
+        ]
+        return ListMetricsOutput(Metrics=metrics, NextToken=None)
```
baermat, line 360: nit: this is missing the pagination while get_metric_data_output has it

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -100,6 +104,54 @@ def test_put_metric_data_gzip(self, aws_client):
...
+                ],
+            )
+        err = ex.value.response["Error"]
+        assert err["Code"] == "InvalidParameterCombination"
+        assert (
+            err["Message"]
+            == "The parameters MetricData.member.1.Value and MetricData.member.1.Values are mutually exclusive and you have specified both."
+        )
+
+        # test invalid due to data can not have and values mismatched_counts
```
baermat, line 134: nit: "values and mismatched_counts" instead

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -194,6 +246,231 @@ def _get_metric_data_sum():
...
+                }
+            ],
+        )
+
+        aws_client.cloudwatch.put_metric_data(
+            Namespace=namespace,
+            MetricData=[
+                {
+                    "MetricName": "metric1",
+                    "Value": 11,
```
baermat, line 423: nit: it would probably be good if this value were distinct from the previous one to have better error detection

baermat (commented):
Wow that looks great! I am sadly not that deep into the topic of metric data to have full insight and give feedback accordingly :sweat_smile: There is a couple of small things that popped up but nothing severe as far as I can see

localstack/services/cloudwatch/cloudwatch_database_helper.py
pinzon: I think  Id as INTEGER is common practice in Relational DBs. But @steffyP can talk more about this.

steffyP: I think using the auto-increment feature of sqlite for the [INTEGER PRIMARY KEY](https://www.sqlite.org/datatypes.html) makes our life lot easier:

> INTEGER PRIMARY KEY columns can be used to implement the equivalent of AUTOINCREMENT. If you try to insert a NULL into an INTEGER PRIMARY KEY column, the column will actually be filled with an integer that is one greater than the largest key already in the table.

No, we don't operate on the ID, but we also don't really have a unique identifier by default. The timestamps, namespace, and dimensions can be freely chosen, there is no restriction. Duplicates would be valid AFAIK.

Still it would be nice to have an internal unique identifier for each metric we insert. Even though we currently don't need it, it may be helpful in the future :)

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+                                self._get_ordered_dimensions_with_separator(
+                                    metric.get("Dimensions")
+                                ),
+                                metric.get("Unit"),
+                                metric.get("StorageResolution"),
+                                insert.get("Value"),
+                            ),
+                        )
+
+                if statistic_values := metric.get("StatisticValues"):
```
steffyP, line 137: It would be great if we could also add a test where we try to insert single + static values at the same time (in one `put-metric-data` statement. I think it will throw an error.

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -109,6 +158,72 @@ def delete_alarms(self, context: RequestContext, alarm_names: AlarmNames) -> Non
...
+            if query_result.get("messages"):
+                messages.extend(query_result.get("messages"))
+
+            # Paginate
+            aliases_list = PaginatedList(query_result.get("datapoints", {}).items())
+            page, nxt = aliases_list.get_page(
+                lambda metric_result: metric_result.get("Id"),
+                next_token=next_token,
+                page_size=limit,
+            )
```
steffyP, line 203: are we sure that the the pagination is for the individual datapoints, and not for the entire "formatted_results"? 🤔

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+        region: str,
+        query: MetricDataQuery,
+        start_time: datetime,
+        end_time: datetime,
+        scan_by: str,
+    ):
+        # TODO exclude null values, check if dimensions must be null though if missing
+        # TODO test default order
+        # TODO add filters for dimensions
+        # TODO add filters for unit
```
steffyP, line 170: please add these TODOs also as remaining tasks for the PR 🙏 we can tackle those issues in upcoming PRs. 
but being able to handle optional Namespace, and dimensions would be important 🙂

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+from datetime import datetime, timezone
+from typing import Dict, List, Optional
+
+from localstack import config
+from localstack.aws.api.cloudwatch import MetricData, MetricDataQuery, ScanBy
+from localstack.utils.files import mkdir
+
+LOG = logging.getLogger(__name__)
+
+STAT_TO_SQLITE_FUNC = {
```
steffyP, line 13: nice! 🎉

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+                            count += 1
+                    cleaned_datapoints[timestamp] = total_sum / count
+
+            return {
+                "id": query.get("Id"),
+                "datapoints": cleaned_datapoints,
+                "label": f"{metric.get('MetricName')} {stat}",
+            }
+
+    def list_metrics(self, account_id, region, namespace, metric_name, dimensions) -> dict:
```
steffyP, line 297: please add types to the parameters :)

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+            return {"metrics": (single_metrics_result + aggregated_metrics_result)}
+
+    def clear_tables(self):
+        # TODO clear tables for reset calls on cloudwatch
+        pass
+
+    def shutdown(self):
+        # TODO delete tmpdir/database if we do not have persistence enabled?
+        # anything else we should consider?
+        ...
```
steffyP, line 369: please also add the open TODOs to the PR description :)

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+                data,
+            )
+            aggregated_metrics_result = [
+                {
+                    "metric_name": r[0],
+                    "namespace": r[1],
+                    "dimensions": self._restore_dimensions_from_string(r[2]),
+                }
+                for r in cur.fetchall()
+            ]
```
steffyP, line 358: alternatively we could use just one combined query, simplified (need to add `WHERE` clauses still):

```
SELECT DISTINCT metric_name, namespace, dimensions
FROM (
    SELECT metric_name, namespace, dimensions
    FROM SINGLE_METRICS
    UNION
    SELECT metric_name, namespace, dimensions
    FROM AGGREGATED_METRICS
) AS combined
```

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+
+                # calculate statistic
+                if stat == "Sum" or stat == "SampleCount":
+                    cleaned_datapoints[timestamp] = sum(cleaned_values)
+
+                elif stat == "Minimum":
+                    cleaned_datapoints[timestamp] = min(cleaned_values)
+
+                elif stat == "Maximum":
+                    cleaned_datapoints[timestamp] = max(cleaned_values)
```
steffyP, line 277: for sum/min/max we could also use a combined SQL. 
I am still thinking about if we can also make it work for avg 🤔

steffyP (commented):
Great work @pinzon! These are the fundamentals for our new provider 🥳 
Let's sync together on what we still want to improve in this PR, and what will go into a follow-up PR!
It might also be better for the future if you create the PR, as I cannot officially do a review this way 😅 
Anyhow, this one was special, because you took over the work I started 🙇‍♀️ 👏

identifier.sqlite
alexrashed, line 1: Should this file maybe be added to the gitignore? :)

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -0,0 +1,407 @@
...
+
+            def _get_current_unix_timestamp_utc():
+                now = datetime.utcnow().replace(tzinfo=timezone.utc)
+                return int(now.timestamp())
+
+            for metric in metric_data:
+                unix_timestamp = (
+                    self._convert_timestamp_to_unix(metric.get("Timestamp"))
+                    if metric.get("Timestamp")
+                    else _get_current_unix_timestamp_utc()
```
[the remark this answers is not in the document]
steffyP: adding to follow-up tasks ✔️

localstack/services/cloudwatch/cloudwatch_database_helper.py
steffyP: adding to follow-up task

localstack/services/cloudwatch/cloudwatch_database_helper.py
steffyP: added to follow-up tasks

localstack/services/cloudwatch/cloudwatch_database_helper.py
steffyP: adding to follow-up task

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -100,6 +104,54 @@ def test_put_metric_data_gzip(self, aws_client):
         assert 1 == len(rs["Metrics"])
         assert namespace == rs["Metrics"][0]["Namespace"]
 
+    @markers.aws.validated
```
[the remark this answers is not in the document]
steffyP: adding to follow-up tasks

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1223,6 +1612,95 @@ def test_describe_minimal_metric_alarm(self, snapshot, aws_client, cleanups):
...
+        for i in range(10):
+            aws_client.cloudwatch.put_metric_data(
+                Namespace=namespace,
+                MetricData=[
+                    dict(MetricName="metric", Value=i, Timestamp=utc_now + timedelta(seconds=1))
+                ],
+            )
+
+        if is_aws_cloud():
+            time.sleep(2)
```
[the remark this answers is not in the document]
steffyP: added to follow-up task

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -317,16 +324,25 @@ def list_metrics(self, account_id, region, namespace, metric_name, dimensions) -
...
+                    FROM SINGLE_METRICS
+                    UNION
+                    SELECT metric_name, namespace, dimensions, account_id, region, timestamp
+                    FROM AGGREGATED_METRICS
+                ) AS combined
+                WHERE account_id = ? AND region = ?
+                {namespace_filter}
+                {metric_name_filter}
+                {dimension_filter}
                 ORDER BY timestamp DESC
```
steffyP, line 338: nice 🎉 
Let's see how we do with the performance in the future. We may be able to improve the query, if we add the `WHERE` clause to the two inner selects already :)

PR #9478, opened 2023-10-26
Fix CFn schema json name resolution for -ext

dominikschubert:
## Motivation

Since we have resource providers in -ext now, we need to fix the schema json lookup. Since -ext code is encrypted, files are renamed and have an `.py.enc` file suffix instead of the `.py` suffix in community.

## Changes

- tries to remove both `.py.enc` as well as `.py` when searching for the schema json file.


simonrw (approved):
Good catch, thanks for that

PR #9479, opened 2023-10-26
Fix updates of event source mapping targeting new function versions

dominikschubert:
## Motivation

- Initiated by a support case, we saw that we never updated the stored function ARN in the event source mapping, which lead to the event source mapping still calling the previous version instead.

Example scenario (see accompanying test case):
```
  Testing an update to an event source mapping that changes the targeted lambda function version
  Resources used:
  - Lambda function
  - 2 published versions of that lambda function
  - 1 event source mapping
  First the event source mapping points towards the qualified ARN of the first version.
  A message is sent to the SQS queue, triggering the function version with ID 1.
  The lambda function is updated with a different value for the environment variable and a new version published.
  Then we update the event source mapping and make the qualified ARN of the function version with ID 2 the new target.
  A message is sent to the SQS queue, triggering the function with version ID 2.
  We should have one log entry for each of the invocations.
```

## Changes

- Properly overwrite the new lambda function ARN when updating the event source mapping to point towards the new function version.
- Adds a snapshot test verifying this previously broken behavior.

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -1678,7 +1678,12 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         context: RequestContext,
         request: UpdateEventSourceMappingRequest,
     ) -> EventSourceMappingConfiguration:
-        state = lambda_stores[context.account_id][context.region]
+        account_id, region = api_utils.get_account_and_region(request["FunctionName"], context)
+        function_name, qualifier = api_utils.get_name_and_qualifier(
+            request["FunctionName"], None, context
+        )
+
+        state = lambda_stores[account_id][region]
         request_data = {**request}
         uuid = request_data.pop("UUID", None)
         if not uuid:
@@ -1690,7 +1695,16 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
             raise ResourceNotFoundException(
                 "The resource you requested does not exist.", Type="User"
             )  # TODO: test?
-        event_source_mapping = {**old_event_source_mapping, **request_data}
+
+        # normalize values to overwrite
+
+        event_source_mapping = {
+            **old_event_source_mapping,
+            **request_data,
+            "FunctionArn": api_utils.qualified_lambda_arn(
+                function_name, qualifier, account_id, region
+            ),
+        }
 
         temp_params = (
             {}
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -1678,12 +1678,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         context: RequestContext,
         request: UpdateEventSourceMappingRequest,
     ) -> EventSourceMappingConfiguration:
-        account_id, region = api_utils.get_account_and_region(request["FunctionName"], context)
-        function_name, qualifier = api_utils.get_name_and_qualifier(
-            request["FunctionName"], None, context
-        )
-
-        state = lambda_stores[account_id][region]
+        state = lambda_stores[context.account_id][context.region]
         request_data = {**request}
         uuid = request_data.pop("UUID", None)
         if not uuid:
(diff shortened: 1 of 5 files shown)


localstack/services/lambda_/provider.py
```diff
@@ -1678,7 +1678,12 @@ def update_event_source_mapping(
         context: RequestContext,
         request: UpdateEventSourceMappingRequest,
     ) -> EventSourceMappingConfiguration:
-        state = lambda_stores[context.account_id][context.region]
+        account_id, region = api_utils.get_account_and_region(request["FunctionName"], context)
```
joe4dev, line 1681: 💡 Is that the new pattern we should prefer over getting the `account_id` and `region` directly from the context?

tests/aws/services/lambda_/test_lambda_integration_sqs.py
```diff
@@ -1168,5 +1172,119 @@ def test_sqs_invalid_event_filter(
...
+        )
+        mapping_uuid = create_event_source_mapping_response["UUID"]
+        cleanups.append(lambda: aws_client.lambda_.delete_event_source_mapping(UUID=mapping_uuid))
+        snapshot.match("create-event-source-mapping-response", create_event_source_mapping_response)
+        _await_event_source_mapping_enabled(aws_client.lambda_, mapping_uuid)
+
+        aws_client.sqs.send_message(QueueUrl=queue_url_1, MessageBody=json.dumps({"foo": "bar"}))
+
+        events = retry(
+            check_expected_lambda_log_events_length,
```
joe4dev, line 1239: convenient util 👍

tests/aws/services/lambda_/test_lambda_integration_sqs.py
```diff
@@ -1168,5 +1172,119 @@ def test_sqs_invalid_event_filter(
...
+            FunctionName=publish_v2["FunctionArn"]
+        )
+        # we're now pointing the existing event source mapping towards the new version.
+        # only v2 should now be called
+        updated_esm = aws_client.lambda_.update_event_source_mapping(
+            UUID=mapping_uuid, FunctionName=publish_v2["FunctionArn"]
+        )
+        assert mapping_uuid == updated_esm["UUID"]
+        assert publish_v2["FunctionArn"] == updated_esm["FunctionArn"]
+        # snapshot.match("updated_esm", updated_esm)
```
joe4dev, line 1267: Is this commented out intentionally?

localstack/services/lambda_/provider.py
joe4dev: It looks like `FunctionName` does not always exist (3 failing tests with KeyError):

```
E             File "/opt/code/localstack/localstack/services/lambda_/provider.py", line 1681, in update_event_source_mapping
E               account_id, region = api_utils.get_account_and_region(request["FunctionName"], context)
E                                                                     ~~~~~~~^^^^^^^^^^^^^^^^
E           KeyError: 'FunctionName'
```

tests/aws/services/lambda_/test_lambda_integration_sqs.py
bentsku: I didn't know, so I un-commented and re-ran the test to create the snapshot and it seems to be working.

tests/aws/services/lambda_/test_lambda_integration_sqs.snapshot.json
```diff
@@ -1464,8 +1464,8 @@
                   "ApproximateFirstReceiveTimestamp": "timestamp"
                 },
                 "messageAttributes": {},
-                "md5OfBody": "<md5-of-body:1>",
```
joe4dev: (interesting flip 😆 )

localstack/services/lambda_/provider.py
```diff
@@ -1696,16 +1691,25 @@ def update_event_source_mapping(
                 "The resource you requested does not exist.", Type="User"
             )  # TODO: test?
 
-        # normalize values to overwrite
+        # remove the FunctionName field
+        function_name_or_arn = request_data.pop("FunctionName", None)
```
joe4dev, line 1695: great catch 🕵️

localstack/services/lambda_/provider.py
```diff
@@ -1696,16 +1691,25 @@ def update_event_source_mapping(
...
         event_source_mapping = {
             **old_event_source_mapping,
             **request_data,
-            "FunctionArn": api_utils.qualified_lambda_arn(
-                function_name, qualifier, account_id, region
-            ),
         }
 
+        if function_name_or_arn:
+            # if the FunctionName field was present, update the FunctionArn of the EventSourceMapping
```
joe4dev, line 1704: I like the clarifying comment 👏

joe4dev (approved):
Thank you very much for taking this over @bentsku and fixing the logic around the conditional `FunctionName` and state retrieval 💯 👏👏👏 

It looks good now 👍

Issue #9439, opened 2023-10-23
bug: terraform destroy fails to delete step function with ExecutionDoesNotExist error


ikalyvas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am using a simple terraform file to create a state machine:

```

# Create an S3 Bucket

data "aws_iam_policy_document" "assume_role_sf" {
  statement {
    effect = "Allow"

    principals {
      type        = "Service"
      identifiers = ["states.amazonaws.com"]
    }

    actions = ["sts:AssumeRole"]
  }
}


# Create an IAM role for the Step Function
resource "aws_iam_role" "step_function_role" {
  name = "step_function_role"
  assume_role_policy = data.aws_iam_policy_document.assume_role_sf.json

}

resource "aws_sfn_state_machine" "dynamodb_updater_workflow" {
  name     = "dynamodb_updater_workflow"
  definition = jsonencode({
    Comment = "A Step Function that writes to DynamoDB",
    StartAt = "Upload",
    States = {
      Upload = {
        Type     = "Task",
        Resource = "arn:aws:states:::dynamodb:putItem",
        Parameters = {
          "TableName": "Files",
          "Item": {
            "FileName": { "S.$": "$.fileName" },
          }
        },
        End     = true,
      }
    }
  })
  role_arn = aws_iam_role.step_function_role.arn
  timeouts {
    create = "5m"
    delete = "5m"
  }

}

```

and using docker-compose to bring up the localstack with PROVIDER_OVERRIDE_STEPFUNCTIONS=v2.

When I do terraform apply the state machine is getting created successfully.
However, when I do terraform destroy I get the following error:

```
│ Error: waiting for Step Functions State Machine (arn:aws:states:eu-central-1:000000000000:stateMachine:dynamodb_updater_workflow) delete: ExecutionDoesNotExist: 
```

I see from the debug logs in the localstack the following:
```
localstack  | 2023-10-23T10:33:00.478  INFO --- [   asgi_gw_1] localstack.request.aws     : AWS stepfunctions.DeleteStateMachine => 200
localstack  | 2023-10-23T10:33:00.486  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS stepfunctions.DescribeStateMachine => 400 (ExecutionDoesNotExist)
```

So it deletes the state machine, but then it tries to do  a list (in the already deleted SM), which causes terraform destroy to fail.

### Expected Behavior

[...]

PR #9481, opened 2023-10-26
fixed describe sfn exception

macnev2013:
## Motivation
This PR fixes the exception raised when calling describe api on non-existing state machine. Which was causing the terraform to fail.

## Changes
Replaced the `ExecutionDoesNotExists` exception to `StateMachineDoesNot` exists

## Testing
Added AWS validated test cases for the updated exception

## Related Issue
- https://github.com/localstack/localstack/issues/9439

### DIFF
--- a/localstack/services/stepfunctions/provider_v2.py
+++ b/localstack/services/stepfunctions/provider_v2.py
@@ -253,7 +253,7 @@ class StepFunctionsProvider(StepfunctionsApi):
         # TODO: add arn validation.
         state_machine = self.get_store(context).state_machines.get(state_machine_arn)
         if state_machine is None:
-            raise ExecutionDoesNotExist()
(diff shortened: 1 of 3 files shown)


MEPalma (approved):
Thank you! LGTM

ikalyvas:
Hello, one question about this fix.
Will this make terraform destroy to succeed ? Because how is changing the exception class going to make terraform continue to the end ? The exception is still raised right ?

macnev2013:
Hi @ikalyvas , Apologies for the delayed response. Yes, terraform destroy works now. It was failing initially because terraform was expecting `StateMachineDoesNot` exception whereas we were responding with `ExecutionDoesNotExists` which was causing terraform to break. Please give it a try and feel free to report if the issue still persists.

PR #9482, opened 2023-10-26
Add fallback partition config option for ARN partition rewriter

viren-nadkarni:
## Motivation

The ARN partition rewriter depends on the deprecated environment config variable `DEFAULT_REGION`. It uses this to determine which partition to use as a fallback.

This is preventing us from removing the `DEFAULT_REGION` config option with the upcoming major release.

## Implementation

This PR introduces a new config option `ARN_PARTITION_FALLBACK` which can be used to set which partition to use as fallback. It defaults to `aws`, which is identical to old behaviour where it would be determined from the `us-east-1` region.

The `DEFAULT_REGION` config option will no longer have any effect on the rewriter.

Acceptance of this PR will entail an update to enterprise image build steps.

## Tests

Slightly tweaks existing tests to assert the new behaviour.

### DIFF
--- a/localstack/aws/handlers/partition_rewriter.py
+++ b/localstack/aws/handlers/partition_rewriter.py
@@ -9,7 +9,6 @@ from urllib.parse import urlparse
 from localstack import config
 from localstack.aws.api import RequestContext
 from localstack.aws.chain import Handler, HandlerChain
-from localstack.constants import AWS_REGION_US_EAST_1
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Thanks! The code is nice and clean! Thanks for addressing this in order to finally be able to remove the `DEFAULT_REGION`. 🚀 🧹

PR #9483, opened 2023-10-26
fix s3 v3 CopyObject in place + general validations for S3

bentsku:
## Motivation
I was looking at the S3 v3 `CopyObject` code and spotted an issue in the code: we wouldn't check the bucket equality when validating the in-place checks.

The v3 provider has parity with the v2, but it could be nice to push things a bit further for `ListMultipartUploads` and `ListParts`, as well as adding more validated tests of `UploadPartCopy`. This is to be done in a follow up PR.

## Changes
The PR was so small I also took a stab at the leftovers TODO in the v3 provider and added tests when removing some, and fixed the behaviour in both `v2` and `v3` for the following operations:
- `UploadPart`
- `CompleteMultipartUpload`
- `AbortMultipartUpload` and `ListParts` only for v3
- `CopyObject` additional validations for source object

### DIFF
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -613,8 +613,17 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         )
         src_moto_bucket = get_bucket_from_moto(moto_backend, bucket=src_bucket)
         source_key_object = get_key_from_moto_bucket(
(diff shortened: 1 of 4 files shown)


macnev2013 (approved):
LGTM

PR #9320, opened 2023-10-09
Add support for AWS::ApiGateway::* resources

Morijarti:
## Changes
Add support for

- "AWS::ApiGateway::GatewayResponse"
- "AWS::ApiGateway::RequestValidator"
- "AWS::ApiGateway::RestApi"
- "AWS::ApiGateway::Deployment"
- "AWS::ApiGateway::Resource"
- "AWS::ApiGateway::Method"
- "AWS::ApiGateway::Stage"
- "AWS::ApiGateway::UsagePlan"
- "AWS::ApiGateway::ApiKey"
- "AWS::ApiGateway::UsagePlanKey"
- "AWS::ApiGateway::DomainName"
- "AWS::ApiGateway::BasePathMapping"
- "AWS::ApiGateway::Model"
- "AWS::ApiGateway::Account"

PR #9493, opened 2023-10-27
fix linting targets and reformat files

alexrashed:
## Motivation
With #9399 we switched from flake8 / isort / autoflake to ruff.
We have now seen some weird behavior that `make lint` does not complain, while `make format` would reformat some files.
This PR addresses these issues. /cc @pinzon @steffyP @joe4dev 

## Changes
- Adds `black` to the linting commands to make sure that the linting commands are failing if code is not properly formatted (this was previously done by a flake8-black integration).
- Reformats the code which slipped through the linting (with https://github.com/localstack/localstack/pull/9320 and https://github.com/localstack/localstack/pull/9404).


Makefile
```diff
@@ -202,19 +202,19 @@ test-docker-mount-code:
...
 check-aws-markers:     		  ## Lightweight check to ensure all AWS tests have proper compatibilty markers set
 	($(VENV_RUN); python -m pytest --co tests/aws/)
 
-format:            		  ## Run black and ruff code formatter
+format:            		  ## Run ruff and black to format the whole codebase
 	($(VENV_RUN); python -m ruff check --show-source --fix .; python -m black .)
 
-format-modified:   		  ## Run black and ruff code formatter on modified files
+format-modified:          ## Run ruff and black to format only modified code
 	($(VENV_RUN); python -m ruff check --show-source --fix `git diff --diff-filter=d --name-only HEAD | grep '\.py$$' | xargs`; python -m black `git diff --diff-filter=d --name-only HEAD | grep '\.py$$' | xargs` )
```
joe4dev, line 218: Is the `;` between `ruff` and `lint` intentional compared to `&&` for `lint-modified`?
i.e., What's the reason for not failing the make target if `ruff` fails?

Makefile
```diff
@@ -202,19 +202,19 @@ test-docker-mount-code:
 	PACKAGES_DIR=$$(echo $$(pwd)/.venv/lib/python*/site-packages | awk '{print $$NF}'); \
 		DOCKER_FLAGS="$(DOCKER_FLAGS) --entrypoint= -v `pwd`/localstack/config.py:/opt/code/localstack/localstack/config.py -v `pwd`/localstack/constants.py:/opt/code/localstack/localstack/constants.py -v `pwd`/localstack/utils:/opt/code/localstack/localstack/utils -v `pwd`/localstack/services:/opt/code/localstack/localstack/services -v `pwd`/localstack/aws:/opt/code/localstack/localstack/aws -v `pwd`/Makefile:/opt/code/localstack/Makefile -v $$PACKAGES_DIR/moto:/opt/code/localstack/.venv/lib/python3.11/site-packages/moto/ -e TEST_PATH=\\'$(TEST_PATH)\\' -e LAMBDA_JAVA_OPTS=$(LAMBDA_JAVA_OPTS) $(ENTRYPOINT)" CMD="make test" make docker-run
 
-lint:              		  ## Run code linter to check code style
-	($(VENV_RUN); python -m ruff check --show-source .)
+lint:              		  ## Run code linter to check code style and check if formatter would make changes
+	($(VENV_RUN); python -m ruff check --show-source . && python -m black --check .)
```
joe4dev, line 206: Assuming we have compatible `line-length` settings across `ruff` and `black`, this should be compatible: https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black

joe4dev: If or once we decide to replace `black` as well, this won't be necessary anymore.

joe4dev (approved):
LGTM 👍 

Thank you for tackling all Makefile targets.

Makefile
alexrashed: Yeah, it is compatible, but ruff does not do all the formatting (for now, as you mentioned). So it's necessary to also run `black --check` until we switch to `ruff format`.

Makefile
alexrashed: That's a good point. It's basically just how it was previously (with isort).
When you execute `format`, you would like to run it on as much as possible (even if subcommands fail, f.e. because some code has syntax errors). The linting has to be strict, every single command has to trigger a non-zero exit code because it's the quality gate.

PR #9494, opened 2023-10-27
add restricted services config

bentsku:
## Motivation
Disclaimer: targeting `master` for now to validate all tests, and we will set the `STRICT_SERVICE_LOADING` variable to default to `True` in the release branch afterwards. This variable won't be advertised and is used to decouple eager service loading from "strict" service loading, and might be removed after the release. 

When some services are problematic, or we want to use LocalStack for only one or more services, we should be able to use the `SERVICES` variable coupled with `STRICT_SERVICE_LOADING=1` in order to be able to only enable the listed services.

The `EAGER_LOADING_SERVICE` functionality is untouched for now, and is applied on top: meaning if you don't have `STRICT_SERVICE_LOADING` set then you'll have all available services, and will only eager load the services in `SERVICES`.

If `STRICT_SERVICE_LOADING=1` is set, then the services in `SERVICES` will be the only ones available. You can additionally eager load them by specifying `EAGER_LOADING_SERVICE=1`.

Example: you only want S3 and SQS loaded:
`SERVICES=sqs,s3` + `STRICT_SERVICE_LOADING=1`

You want kinesis eager loaded at startup, but still lazy load other services:
`SERVICES=kinesis` + `EAGER_LOADING_SERVICE=1`

You want to only use kinesis, dynamodb, s3 and sqs, but with eager loading:
`SERVICES=kinesis,dynamodb,sqs,s3` + `STRICT_SERVICE_LOADING=1` + `EAGER_LOADING_SERVICE=1`

We should also do a new pass on the `API_DEPENDENCIES` constant to validate the inter-service dependencies, and be sure we will load the necessary services.

\cc @steffyP 

## Changes
Change the logic for `get_enabled_apis`, which is now linked to `STRICT_SERVICE_LOADING`.
Create new logic for eager loading, linked to the `get_preloaded_services` method and `EAGER_LOADING_SERVICE`. 
Add a test to validate the new behaviour of disabling services.


localstack/aws/handlers/service_plugin.py
```diff
@@ -40,7 +41,7 @@ def require_service(self, _: HandlerChain, context: RequestContext, response: Re
             return
 
         service_name: str = context.service.service_name
-        if not self.service_manager.exists(service_name):
+        if not self.service_manager.exists(service_name) or not is_api_enabled(service_name):
             raise NotImplementedError
```
alexrashed, line 45: This could be problematic in combination with the handling of the `NotImplementedError` in `skeleton.py`:
https://github.com/localstack/localstack/blob/d97a2e8b4183d7b978b61bd8583f811e5ccd392a/localstack/aws/skeleton.py#L195-L222

It would be great if the error message could be different if a service is disabled (in comparison to the error message when the service operation is just not implemented / not found at all).
We could maybe handle this by creating a specific subclass of `NotImplementedError` (maybe something like `ServiceNotEnabledError`?), and in the skeleton we could then differentiate and use a different message (other than the `MESSAGE_TEMPLATE` in `coverage_docs.py`.
Another, possibly simpler, solution would be to raise a `NotimplementedError` with the specific error message right here (which will then be used, since the message template is just a fallback).

localstack/utils/bootstrap.py
```diff
@@ -249,9 +281,8 @@ def get_enabled_apis() -> Set[str]:
...
 
 
-# DEPRECATED, lazy loading should be assumed
-def is_api_enabled(api: str) -> bool:
-    apis = get_enabled_apis()
+def should_eager_load_api(api: str) -> bool:
+    apis = get_preloaded_services()
 
     if api in apis:
         return True
```
alexrashed, line 288: nit:
```suggestion
def should_eager_load_api(api: str) -> bool:
    return api in get_preloaded_services()
```

localstack/utils/bootstrap.py
```diff
@@ -221,17 +221,49 @@ def resolve_apis(services: Iterable[str]) -> Set[str]:
...
-    elif services_env:
+
+    if services_env and is_env_true("STRICT_SERVICE_LOADING"):
+        from localstack.services.plugins import SERVICE_PLUGINS
+
+        # SERVICES and STRICT_SERVICE_LOADING are set
+        # we check with SERVICE_PLUGINS.list_available() to cross the user-provided list with the available ones
+        available_services = SERVICE_PLUGINS.list_available()
+        services = [service for service in services_env.split(",") if service in available_services]
+        # TODO: log a message if a service was not supported? see with pro loading
```
alexrashed, line 240: nit: I think this would be a good idea :)

localstack/utils/bootstrap.py
```diff
@@ -221,17 +221,49 @@ def resolve_apis(services: Iterable[str]) -> Set[str]:
...
+
+    if not services:
+        from localstack.services.plugins import SERVICE_PLUGINS
+
+        services = SERVICE_PLUGINS.list_available()
```
alexrashed, line 245: Would this be the same as removing this section and initially setting `services = []` instead?

alexrashed (commented):
Nice! Really cool implementation of the strict service loading. Awesome that it stays pretty simple and clean. I only had a few comments, most of them are nitpicks :)

localstack/services/infra.py
```diff
@@ -445,15 +450,15 @@ def preload_services():
...
         # listing the available service plugins will cause resolution of the entry points
-        available_services = SERVICE_PLUGINS.list_available()
+        available_services = get_enabled_apis()
 
         # lazy is the default beginning with version 0.13.0
         if not config.EAGER_SERVICE_LOADING:
             return
 
         for api in available_services:
             # this should be the only call to is_api_enabled left
```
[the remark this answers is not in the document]
bentsku: good catch!

localstack/utils/bootstrap.py
bentsku: I agree, but I think this might create some issues, I'll need to test it properly, as it might be called once with community, then the cache is cleared and once again in Pro. If the services are not there yet, we might log some messages twice. It will need some local testing but I'll keep that in mind for a following PR.

localstack/utils/bootstrap.py
bentsku: I don't think so, in this case we will load all discovered services. If we set `services = []`, then we would load none. 

We could however simplify the logic, setting `services = SERVICE_PLUGINS.list_available()` above then filter it if the var is set. That would make sense! Will update

localstack/utils/bootstrap.py
bentsku: Then we will remove the check for the `port` part that we had previously. I'm okay with removing it, but we had specific tests in place to remove the port part like `SERVICES=sqs:123` However, those would not work with `STRICT_SERVICE_LOADING`, so I will remove.

localstack/aws/handlers/service_plugin.py
bentsku: Right, good catch. Thomas said the services should raise `NotImplemented` if not enabled. But a different message would be great, definitely, and will help users understand what's wrong.

localstack/utils/bootstrap.py
alexrashed: I'm a bit confused, this change request is semantically equivalent, right?

bentsku: It's missing a part:
```python
def should_eager_load_api(api: str) -> bool:
    apis = get_preloaded_services()
    if api in apis:
        return True
    for enabled_api in apis:
        if api.startswith(f"{enabled_api}:"):
            return True
    return False
```
But it's good we're removing this, this is very old and will be good to be removed

alexrashed: Oh, that was just due to the review view in GitHub. I haven't seen the lines below. Yours to judge if it's time to remove this. Otherwise, please ignore the comment!

localstack/utils/bootstrap.py
```diff
@@ -282,16 +274,7 @@ def get_preloaded_services() -> Set[str]:
...
-
-    if api in apis:
-        return True
-
-    for enabled_api in apis:
-        if api.startswith(f"{enabled_api}:"):
-            return True
-
-    return False
+    return api in get_preloaded_services()
```
alexrashed, line 277: Unfortunately, I haven't seen the remaining part of the function when I did the initial PR review (suggesting the simplification of this function). I think it might be a bit hasty, based on my mistaken comment, to break this behavior. I have just recently seen a `SERVICES` config which still contains port mappings, and I think there's no harm in continuing to just ignore them.
Sorry for the inconvenience and the confusion with the initial PR comment...

alexrashed (commented):
Sorry about the confusion with my (mistaken) PR comment in the last review.
Besides the new change concerning the port mapping, the code is looking great! 🚀

alexrashed (approved):

PR #9502, opened 2023-10-30
sns: get region and account from parsed `endpointArn`

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `sns` tests should still create the consequent resources in this accounts and region.

## Changes
This PR fixes the failing tests in `tests.aws.services.sns.test_sns.py` when using non default region and account by:
- using region and account from parsed `endpointArn` in case of platform endpoint messages. 
- using `TEST_AWS_REGION_NAME` instead of static `us-east-1` value. 
- remove use of region and account id in case `endpointArn` is present.

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -60,7 +60,12 @@ from localstack.services.sns.publisher import (
     SnsBatchPublishContext,
     SnsPublishContext,
 )
-from localstack.utils.aws.arns import ArnData, parse_arn
+from localstack.utils.aws.arns import (
+    ArnData,
+    extract_account_id_from_arn,
+    extract_region_from_arn,
+    parse_arn,
+)
 from localstack.utils.strings import short_uid
 
 # set up logger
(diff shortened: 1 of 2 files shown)


localstack/services/sns/provider.py
```diff
@@ -495,8 +500,10 @@ def publish(
         if not phone_number:
             # use the account to get the store from the TopicArn (you can only publish in the same region as the topic)
             parsed_arn = parse_and_validate_topic_arn(topic_or_target_arn)
-            store = self.get_store(account_id=parsed_arn["account"], region_name=context.region)
-            moto_sns_backend = self.get_moto_backend(parsed_arn["account"], context.region)
+            store = self.get_store(
+                account_id=parsed_arn["account"], region_name=parsed_arn["region"]
+            )
+            moto_sns_backend = self.get_moto_backend(parsed_arn["account"], parsed_arn["region"])
```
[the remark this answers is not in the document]
sannya-singal: @bentsku, here I was getting DEFAULT values of region in `context.region` and `context.account_id` due to which the test was failing for multi accounts and regions. So I used parsed region for `region_name` as already being used for the `account_id`.

tests/aws/services/sns/test_sns.py
```diff
@@ -4638,7 +4638,10 @@ def test_subscription_tokens_can_retrospect(
             timeout=2,
         )
 
-        wrong_sub_arn = subscription_arn.replace(TEST_AWS_REGION_NAME, "us-west-1")
+        wrong_sub_arn = subscription_arn.replace(
+            TEST_AWS_REGION_NAME,
+            "us-west-1" if TEST_AWS_REGION_NAME != "us-west-1" else "us-east-1",
+        )
```
viren-nadkarni, line 4644: Maybe use an uncommon region instead?

```suggestion
        wrong_sub_arn = subscription_arn.replace(TEST_AWS_REGION_NAME, "me-south-1")
```

viren-nadkarni (approved):
Clever change with the ARN fallback. Do rename the PR before merge so that it reflects the change :)

bentsku (approved):
LGTM, thanks for removing the provider part, I believe it was fixed with the changes in #9506. Like Viren said, if you could change the PR title to reflect the change, that would be awesome 🚀

tests/aws/services/sns/test_sns.py
sannya-singal: maybe this `"il-central-1" if TEST_AWS_REGION_NAME != "il-central-1" else "me-south-1"` would be better @viren-nadkarni? The check seems necessary.

PR #9503, opened 2023-10-30
fix s3 notification for lambda

macnev2013:
## Motivation
Fix multi account stabilty. This PR fixes s3 notification for lambda function invocation.

## Changes

Modified the `lambda_client.invoke` function parameter to use the **Lambda Function ARN** (Amazon Resource Name) instead of the **Lambda function name**.

## Testing

This PR fixes the test case mentioned below, when ran using non-default account id.

Required envs for the test
```
TEST_AWS_ACCOUNT_ID=111111111111;TEST_AWS_ACCESS_KEY_ID=111111111111;TEST_AWS_REGION=us-west-1
```

Test Case
```
tests/aws/services/s3/test_s3_notifications_lambda.py
```

### DIFF
--- a/localstack/services/s3/notifications.py
+++ b/localstack/services/s3/notifications.py
@@ -635,10 +635,10 @@ class LambdaNotifier(BaseNotifier):
         lambda_client = connect_to(region_name=arn_data["region"]).lambda_.request_metadata(
             source_arn=s3_bucket_arn(ctx.bucket_name), service_principal=ServicePrincipal.s3
         )
-        lambda_function_config = arns.lambda_function_name(lambda_arn)
+
         try:
             lambda_client.invoke(
-                FunctionName=lambda_function_config,
+                FunctionName=lambda_arn,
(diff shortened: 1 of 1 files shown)


bentsku (changes requested):
This is not the proper fix, this has already been reverted with #9411
This will make IAM tests fail in -ext, because AWS does not use the account id to trigger the notification.

I don't know which test is failing and why, and what this PR fixes exactly, but I don't think the root cause is this fix, and it might be somewhere else.

macnev2013:
Hi @bentsku, Thanks for the feedback. I've updated the PR with proper fix.

bentsku (approved):
LGTM! Awesome, this seems like it should do the trick 👍
I ran the -ext suite against it to validate we don't break any IAM thing, but this should be good: https://github.com/localstack/localstack-ext/actions/runs/6704809720

Let's see if it's all green, and if it is it's all good to me! 🚀

PR #9504, opened 2023-10-30
fix ASF update service detection

alexrashed:
## Motivation
Recent ASF update PRs wrongly state that all services have changed (f.e. see https://github.com/localstack/localstack/pull/9500).
This is because in https://github.com/localstack/localstack/pull/9436 we moved the code formatting to after the "change" (and "changed service") detection.
Since we now generate all APIs without formatting them, all APIs seem to change.

## Changes
- Move the formatting step in the GitHub action a bit such that it is executed before the changes are detected.

## Testing
This test run shows that the formatting now works correctly (which is what #9500 was aiming at), and the change detection works correctly (which this PR is aiming at): https://github.com/localstack/localstack/actions/runs/6691512704

### DIFF
--- a/.github/workflows/asf-updates.yml
+++ b/.github/workflows/asf-updates.yml
@@ -44,6 +44,9 @@ jobs:
           source .venv/bin/activate
           python3 -m localstack.aws.scaffold upgrade
 
+      - name: Format code
+        run: make format-modified
+
       - name: Check for changes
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM! thanks for the test run validating the fix and the explanation, makes a lot of sense 👍

PR #9505, opened 2023-10-30
fix test cdk multi account

macnev2013:
## Motivation
Fix multi account stabilty. This PR fixes request headers in the test case to create resources in the correct region and account.

## Changes
Added access key and the region in header of the request in the test case.

## Testing

Fixes the test `test_cdk` when using non-default account id and region.


pinzon (approved):
LGTM

dominikschubert (approved):

dominikschubert:
Seems this was implemented by https://github.com/localstack/localstack/pull/9509 in the meantime

PR #9506, opened 2023-10-30
Ensure backdoor endpoint tests respect the test account ID

viren-nadkarni:
## Motivation

Certains tests related to the backdoor/developer endpoints did not account for the multi-account nature of LocalStack.

This PR ensures that the this is tested appropriately.

## Implementation

With SNS, the assertions against the endpoints are performed in the account where the SNS resource was created.

With SQS, the tests are parametrised so that they run with endpoint schemes that support multi-region.

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -10,6 +10,7 @@ from operator import itemgetter
 import pytest
 import requests
 import xmltodict
+from botocore.auth import SigV4Auth
 from botocore.exceptions import ClientError
 from pytest_httpserver import HTTPServer
 from werkzeug import Response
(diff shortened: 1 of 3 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -265,7 +268,13 @@ def test_publish_by_path_parameters(
...
         sns_create_sqs_subscription(topic_arn=topic_arn, queue_url=queue_url)
 
-        client = aws_http_client_factory("sns", region="us-east-1")
+        client = aws_http_client_factory(
+            "sns",
+            signer_factory=SigV4Auth,
+            region=TEST_AWS_REGION_NAME,
+            aws_access_key_id=TEST_AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=TEST_AWS_SECRET_ACCESS_KEY,
+        )
```
viren-nadkarni, line 277: This ensures that the credentials are correctly sent when using the API in a raw manner.

bentsku: q: can this work against AWS? the test is AWS validated

bentsku (commented):
Nice changes! Just have a question regarding the AWS validated test change.
For the backdoor tests, it seems #9502 is doing some changes there too, seems like they might clash.

For the SQS changes, those looks good to me 👍

tests/aws/services/sns/test_sns.py
viren-nadkarni: Yes. I also sneaked a small change in 534348be9 to ensure multi-region

bentsku: Okay great, I wasn't sure, thanks for confirming! 😄

bentsku (approved):

PR #9464, opened 2023-10-25
Bump moto-ext to 4.2.6.post2

viren-nadkarni:
Brings in:
- https://github.com/getmoto/moto/pull/6933 @bentsku 
- https://github.com/getmoto/moto/pull/6930

Ext Integration Tests passing :heavy_check_mark:  # 2912

PR #9507, opened 2023-10-30
remove s3 encoding tets xfail marker and xfail flaky events scheduler test

bentsku:
## Motivation
Now that #9464 has been merged (thanks @viren-nadkarni!), the fix for S3 object key encoding is contained in LocalStack.
We need to remove the xfail marker to validate the fix and close the related the issue.

Also, `test_scheduled_rule_logs` has been really flaky, it might be a race condition but to not block the pipeline and force re-run, I've xfailed the test for now.

## Changes
Remove the xfail marker for `test_delete_objects_encoding`. 
xfail `tests.aws.services.events.scheduled_rules.test_events_scheduled_rules_logs.test_scheduled_rule_logs`

_fixes #9428_


alexrashed (approved):
Awesome! Thanks for verifying the fix and for xfailing the flaky test! 🚀 🧹

PR #9509, opened 2023-10-30
Refactor aws_stack.mock_aws_request_headers()

viren-nadkarni:
## Motivation

The `aws_stack.mock_aws_request_headers()` utility is used in our codebase to generate headers for AWS requests. Among these headers include `Content-Type`, `Accept-Encoding`, date and especially the `Authorization` header.

The issue this PR is addressing is the optional nature of the access key and region name arguments for this function, in which the `get_aws_access_key_id()` and `aws_stack.get_region()` are used. We already know that these are problematic functions that rely on thread context storage and are not guaranteed to return the correct values. Furthermore they must not be used in tests, indirectly which is exactly what is happening.

## Implementation

This PR makes the access key ID and region name field mandatory.

### DIFF
--- a/localstack/utils/aws/request_context.py
+++ b/localstack/utils/aws/request_context.py
@@ -9,7 +9,13 @@ from requests.models import Request
 from requests.structures import CaseInsensitiveDict
 
 from localstack import config
-from localstack.constants import APPLICATION_JSON, APPLICATION_XML, HEADER_CONTENT_TYPE
+from localstack.constants import (
+    APPLICATION_JSON,
+    APPLICATION_XML,
+    AWS_REGION_US_EAST_1,
+    DEFAULT_AWS_ACCOUNT_ID,
+    HEADER_CONTENT_TYPE,
+)
 from localstack.utils.aws import aws_stack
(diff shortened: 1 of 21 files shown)


localstack/utils/aws/request_context.py
```diff
@@ -136,7 +143,12 @@ def configure_region_for_current_request(region_name: str, service_name: str):
 
     headers = request_context.headers
     auth_header = headers.get("Authorization")
-    auth_header = auth_header or aws_stack.mock_aws_request_headers(service_name)["Authorization"]
+    auth_header = (
+        auth_header
+        or aws_stack.generate_aws_request_headers(
+            service_name, aws_access_key_id=DEFAULT_AWS_ACCOUNT_ID, region_name=AWS_REGION_US_EAST_1
+        )["Authorization"]
+    )
```
viren-nadkarni, line 151: This is dead code but updated anyway

alexrashed (approved):
LGTM! Thanks for churning over all these refactorings to properly use the account ID and regions everywhere! 🧹 🚀

Issue #9258, opened 2023-09-29
bug: step function with terraform throws ExecutionDoesNotExist


timoly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using `PROVIDER_OVERRIDE_STEPFUNCTIONS=v2` Step function defined using terraform throws the following error during the `tflocal plan` stage.

```
Error: reading Step Functions State Machine (arn:aws:states:eu-west-1:000000000000:stateMachine:***): ExecutionDoesNotExist: 
   with module.step-function-jobs.module.***_step_function_9606D46A.aws_sfn_state_machine.this[0],
   on .terraform/modules/step-function-jobs.***_step_function_9606D46A/main.tf line 15, in resource "aws_sfn_state_machine" "this":
   15: resource "aws_sfn_state_machine" "this" {
```

I can see in logs the following entry `AWS stepfunctions.DescribeStateMachine => 400 (ExecutionDoesNotExist)`

### Expected Behavior

tflocal plan works with any errors.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

tflocal plan

### Environment

```markdown
- OS: mac 13.6
- LocalStack: 2.3.0
```

kevle1:
Can confirm I'm also having this issue, using `docker run`.

Happens if the Terraform project has already been initialized and applied at least once. 

If you delete the local state files and re-apply it will work: 
```
.terraform
terraform.lock.hcl
terraform.tfstate
terraform.tfstate.backup
```

But it's pretty annoying nevertheless. Looks like if theres an existing state file, Terraform queries for executions and throws when Localstack returns that 400.

whummer:
Hi @timoly , thanks for reporting this issue with persistence support for the StepFunctions v2 provider. This should be resolved in the meantime - when you get a chance, can you please pull the `localstack/localstack-pro:latest` Docker image and give it another try? Please keep us posted on how it goes.. Thanks

kevle1:
I can confirm on my end that re-applying Terraform that has an existing state with a step function is now successful. Thanks for that!

PR #9515, opened 2023-10-31
Add state visitor handler for StepFunctions v2 persistence support

whummer:
## Motivation

Persistence support has been requested by a customer using the v2 StepFunctions provider (running LS with `PROVIDER_OVERRIDE_STEPFUNCTIONS=v2 PERSISTENCE=1`). Addresses https://github.com/localstack/localstack/issues/9258

## Changes

* Add state visitor handler (`accept_state_visitor(..)`) to the StepFunctions v2 provider, to enable persistence support and persist the `sfn_stores` object
* Pull out the nested classes in the store models as top-level classes - nested classes can be problematic from a persistence/pickle perspective, in fact, I did a quick check with the nested classes, and ran into a `EOFError: Ran out of input` error when attempting to load the pickled state from disk)

As a next step, tests will be added in the upstream repo.

### DIFF
--- a/localstack/services/stepfunctions/backend/execution.py
+++ b/localstack/services/stepfunctions/backend/execution.py
@@ -79,7 +79,6 @@ class BaseExecutionWorkerComm(ExecutionWorkerComm):
 
 
 class Execution:
-
     name: Final[str]
     role_arn: Final[Arn]
     exec_arn: Final[Arn]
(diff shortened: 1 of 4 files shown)


MEPalma (approved):
Thank you for jumping on this issue! 

In later stages we should also focus on enforcing StateMachine terminations before extracting the state. If not, the state would be polluted with running StateMachine data that however cannot be resumed, as the evaluation tree will be lost and cannot  be trivially reconstructed.

PR #9516, opened 2023-10-31
add utility function to check comma-delimited strings

giograno:
## Motivation
In [2263](https://github.com/localstack/localstack-ext/pull/2263) we moved quite some CLI-related code around.
During the core review, it turned out that `localstack.utils.collections` is a better place for this utility function.

## Changes
- Moving the function;
- Adding a small unit test.


alexrashed (approved):
💯 🧹

PR #9517, opened 2023-10-31
harden s3 v3 concurrency

bentsku:
## Motivation
Following a failure when trying to set the v3 provider as default, I had this PR shelved since a while as I was expecting it to happen.
https://app.circleci.com/pipelines/github/localstack/localstack/19625/workflows/c064060a-3962-4b34-96a0-6c0bfe0efc00/jobs/153568/tests

This PR hardens S3 v3 concurrency, by shallow copying `dict` that will be accessed concurrently before calling `.values`, `.keys` or `.items` on them, which could trigger exception as they would be modified during iteration.

\cc @dfangl since we worked on this together for the v2 provider

## Changes
Shallow-copy some dictionaries so that they wouldn't be modified during iteration from concurrent requests. 
Also added a test to check for iteration during `List*` calls.

### DIFF
--- a/localstack/services/s3/v3/models.py
+++ b/localstack/services/s3/v3/models.py
@@ -543,7 +543,8 @@ class KeyStore:
         return self._store.pop(object_key, default)
 
     def values(self, *_, **__) -> list[S3Object | S3DeleteMarker]:
-        return [value for value in self._store.values()]
+        # we create a shallow copy with dict to avoid size hanged during iteration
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Thanks for proactively tackling this issue! 🚀 
For some context: This is the change from @dfangl which implemented the same measures to harden the bucket creation - https://github.com/localstack/localstack/pull/9072/.

PR #9520, opened 2023-10-31
Implemented AWS::CDK::Metadata resource provider

Morijarti:
## Changes
Implements resource provider for CDKMetadataProperties


simonrw (approved):
Great, a small but important resource!

localstack/services/cdk/resource_providers/cdk_metadata.schema.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "definitions": {
+  },
+  "required": [
+  ],
+  "createOnlyProperties": [
+  ],
+  "primaryIdentifier": [
+    "/properties/Id"
+  ],
```
dominikschubert, line 18: I'm actually curious if a Ref on that resource type would even work :D

PR #9521, opened 2023-10-31
raise exception for undefined dependency in CFn template

pinzon:
## Motivation
 The Logs are pretty hard to understand and are not helpful when a CloudFormation template resource uses a dependency or refers to a resource that is not defined in the template.

## Changes
Now the TemplateDeployer raises an exception indicating that it doesn't find in the template the definition for a referenced resource.

## Testing
Simple 'only_localstack' tests to validated the failed deployed with a readable reason for it

## Note:
The alternative to raising an error is to just print a LOG and continue with the rest of resources but this can lead to unexpected results and it's an error that should make the template deployment fail.

AWS validates the template and dependencies before deployment, but that's a big issue to resolve in LocalStack.

cc/ @alexrashed

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -964,6 +964,10 @@ class TemplateDeployer:
     ):
         result = {}
         for resource_id, resource in resources.items():
+            if not resource:
+                raise Exception(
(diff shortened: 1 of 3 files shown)


simonrw (approved):
Nice, better to fail loud and early

alexrashed (approved):
Nice! Thanks a lot for jumping on this! It's great to fail gracefully here with a clear description of the issue. 🚀

PR #9522, opened 2023-10-31
[SFN] Fix TimestampPath Value Access

MEPalma:
## Motivation
In the current implementation of StepFunctions v2, the interpreter is unable to access the string value for `TimestampPath` definitions. This PR adds validated snapshot tests for `Wait` StateMachines with both `Timestamp` and `TimestampPath` invocations.

## Changes
Updated json path access for `TimestampPath` declarations to the interpreter's standard logic.


alexrashed (approved):
Awesome! Thanks for jumping on this issue! 🚀

PR #9527, opened 2023-11-01
kinesis-lambda: Remove aws_stack.get_region() and get_aws_account_id()

sannya-singal:
## Motivation

This resolution also fixes failing `tests.aws.services.lambda_.test_lambda_integration_kinesis.TestKinesisSource.test_create_kinesis_event_source_mapping` for non-default account and region. 

## Changes
This PR removes the use of the above helper functions and instead use region and account id from parsed ARN. 

## Testing
When using values other than `000000000000` for account ID or `us-east-1` for region, the tests should still create the consequent resources in this accounts and region.


joe4dev (approved):
LGTM 👍 

Thank you for tackling this `aws_stack` clean up @sannya-singal 👏

PR #9528, opened 2023-11-01
Remove fallback account ID and region for core ARN builder

viren-nadkarni:
## Motivation

Over the past few weeks, ARN builders in LocalStack were refactored to use request context account IDs and regions: https://github.com/localstack/localstack/pull/9398

This PR wraps up these refactorings.

All the ARN builders use a common function `_resource_arn()` which falls back to the problematic `get_aws_account_id()` and `aws_stack.get_region()` utilities. This PR removes these fallbacks.

## Tests

Ext Integration Tests # 3441 :green_circle:


localstack/services/s3/utils.py
```diff
@@ -593,7 +593,7 @@ def capitalize_header_name_from_snake_case(header_name: str) -> str:
     return "-".join([part.capitalize() for part in header_name.split("-")])
 
 
-def get_kms_key_arn(kms_key: str, account_id: str, bucket_region: str = None) -> Optional[str]:
+def get_kms_key_arn(kms_key: str, account_id: str, bucket_region: str) -> Optional[str]:
```
bentsku, line 596: great catch, thanks for fixing this! 🙏

alexrashed (approved):
Nice! I'm glad to see that these efforts are paying off and everything is coming together! 🧹 🚀

PR #9536, opened 2023-11-02
KMS: Remove legacy 'local-kms' based provider

viren-nadkarni:
## Motivation

The old 'local-kms' based KMS provider was deprecated in v1.4. It will be removed in v3.

Use of the legacy option `KMS_PROVIDER` will now print a warning.


alexrashed (approved):
Nice and clean removal! 🧹 🔝

PR #9537, opened 2023-11-02
Remove aws_stack.get_region() and get_aws_account_id() from tests

viren-nadkarni:
## Motivation

The functions `aws_stack.get_region()` and `get_aws_account_id()` use thread local storage to retrieve their values. The correct values are set in LocalStack request handler chain. If these values are not set, these functions return the fallback values.

Consequently these functions must not be used in tests. This is because the account ID and regions used by the test fixture `aws_client` do not necessarily match the fallback values.

## Implementation

This PR removes all use of these functions from the community tests.


alexrashed (approved):
LGTM! 🚀 🧹

PR #9541, opened 2023-11-02
Fix broken multi-account/multi-region functionality

viren-nadkarni:
## Motivation

This PR fixes multi-account/multi-region region functionality that misbehave in non-default region and accounts.

In effect, fixes following tests:
- [ ]  tests.aws.services.sqs.test_sqs.TestSqsQueryApi
    - [ ]  test_get_queue_attributes_works_without_authparams
    - [ ]  test_get_queue_url_works_for_same_queue[standard]
    - [ ]  test_get_queue_url_works_for_same_queue[domain]
    - [ ]  test_get_queue_url_works_for_same_queue[path]
    - [ ]  test_get_queue_url_work_for_different_queue[standard]
    - [ ]  test_get_queue_url_work_for_different_queue[domain]
    - [ ]  test_get_queue_url_work_for_different_queue[path]
- [ ]  tests.aws.services.sqs.test_sqs.TestSQSMultiAccounts
    - [ ]  test_cross_account_get_queue_url[standard]
    - [ ]  test_cross_account_get_queue_url[domain]
    - [ ]  test_cross_account_get_queue_url[path]
- [ ]  tests.aws.services.stepfunctions.legacy.test_stepfunctions_legacy
    - [ ]  test_default_logging_configuration
- [ ]  tests.aws.services.dynamodb.test_dynamodb.TestDynamoDB
    - [ ]  test_stream_spec_and_region_replacement
- [ ]  tests.aws.services.logs.test_logs.TestCloudWatchLogs
    - [ ]  test_put_subscription_filter_lambda
    - [ ]  test_put_subscription_filter_firehose
    - [ ]  test_put_subscription_filter_kinesis
- [ ]  tests.aws.services.sqs.test_sqs_backdoor.TestSqsDeveloperEndpoints
    - [ ]  test_list_messages_with_invisible_messages[standard]
    - [ ]  test_list_messages_with_invisible_messages[domain]
    - [ ]  test_list_messages_with_invisible_messages[path]
    - [ ]  test_list_messages_with_delayed_messages[standard]
    - [ ]  test_list_messages_with_delayed_messages[domain]
    - [ ]  test_list_messages_with_delayed_messages[path]
- [ ]  tests.aws.services.s3.test_s3.TestS3MultiAccounts
    - [ ]  test_shared_bucket_namespace
    - [ ]  test_cross_account_access


tests/aws/test_integration.py
```diff
@@ -102,7 +102,7 @@ def test_firehose_s3(self, firehose_create_delivery_stream, s3_create_bucket, aw
         assert TEST_TAGS == tags["Tags"]
 
         # create target S3 bucket
-        aws_client.s3.create_bucket(Bucket=bucket_name)
+        s3_create_bucket(Bucket=bucket_name)
```
viren-nadkarni, line 105: Using this fixture ensures the the appropriate `LocationConstraint` is added automatically.

tests/aws/services/logs/test_logs.py
```diff
@@ -289,9 +287,9 @@ def test_put_subscription_filter_lambda(
         result = aws_client.lambda_.add_permission(
             FunctionName=test_lambda_name,
             StatementId=test_lambda_name,
-            Principal=f"logs.{config.AWS_REGION_US_EAST_1}.amazonaws.com",
+            Principal=f"logs.{TEST_AWS_REGION_NAME}.amazonaws.com",
             Action="lambda:InvokeFunction",
-            SourceArn=f"arn:aws:logs:{config.AWS_REGION_US_EAST_1}:{account_id}:log-group:{logs_log_group}:*",
+            SourceArn=f"arn:aws:logs:{TEST_AWS_REGION_NAME}:{account_id}:log-group:{logs_log_group}:*",
```
viren-nadkarni, line 292: Assertions expected the `us-east-1` region even when the resources were created in a different region (by setting `TEST_AWS_REGION_NAME`)

localstack/services/logs/provider.py
```diff
@@ -341,20 +348,30 @@ def moto_put_log_events(self: "MotoLogStream", log_events):
...
                     StreamName=stream_name,
                     Data=payload_gz_encoded,
                     PartitionKey=self.log_group.name,
                 )
+
             if ":firehose:" in destination_arn:
-                client = connect_to().firehose
+                client = connect_to(
+                    aws_access_key_id=arn_data["account"], region_name=arn_data["region"]
+                ).firehose
```
viren-nadkarni, line 374: Ensures logs are sent to correct account and region.

Identified in following tests:
- test_put_subscription_filter_lambda
- test_put_subscription_filter_firehose
- test_put_subscription_filter_kinesis

tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -218,7 +218,7 @@ def test_stream_spec_and_region_replacement(self, aws_client):
         table = aws_client.dynamodb.describe_table(TableName=table_name)["Table"]
 
         # assert ARN formats
-        expected_arn_prefix = "arn:aws:dynamodb:" + aws_stack.get_local_region()
+        expected_arn_prefix = "arn:aws:dynamodb:" + TEST_AWS_REGION_NAME
```
viren-nadkarni, line 221: Ensures expected value is correctly constructed.

Fixes this test: test_stream_spec_and_region_replacement

tests/aws/services/stepfunctions/legacy/test_stepfunctions_legacy.py
```diff
@@ -721,6 +723,7 @@ def test_default_logging_configuration(create_state_machine, custom_client):
         assert result["ResponseMetadata"]["HTTPStatusCode"] == 200
         assert result["loggingConfiguration"] == {"level": "OFF", "includeExecutionData": False}
     finally:
+        custom_client.stepfunctions.delete_state_machine(stateMachineArn=result["stateMachineArn"])
```
viren-nadkarni, line 726: The test did not use `custom_client` to create the statemachine and then failed during retrieval using this client.

Affected test: test_default_logging_configuration

tests/aws/services/sqs/test_sqs.py
```diff
@@ -4149,7 +4155,7 @@ def test_cross_account_get_queue_url(
         queue_name = f"test-queue-cross-account-{short_uid()}"
         queue_url = sqs_create_queue(QueueName=queue_name)
         account_id, region_name, queue_name_from_url = parse_queue_url(queue_url)
-        assert account_id == DEFAULT_AWS_ACCOUNT_ID
+        assert account_id == TEST_AWS_ACCOUNT_ID
```
viren-nadkarni, line 4158: Assertion was done against the wrong account

localstack/services/logs/provider.py
```diff
@@ -250,8 +250,12 @@ def moto_put_subscription_filter(fn, self, *args, **kwargs):
     if not log_group:
         raise ResourceNotFoundException("The specified log group does not exist.")
 
+    arn_data = parse_arn(destination_arn)
+
     if ":lambda:" in destination_arn:
-        client = connect_to(region_name=extract_region_from_arn(destination_arn)).lambda_
+        client = connect_to(
+            aws_access_key_id=arn_data["account"], region_name=arn_data["region"]
+        ).lambda_
```
bentsku, line 258: I believe this might be the same issue we had in S3 notifications. If `logs` also cross-account/cross-region destinations, I don't believe we need to add the account here to the client call, as lambda is already cross-account/cross-region if you use the lambda function ARN. 
You should use the ARN here in `client.get_function(FunctionName=destination_arn)` instead.
This will help later on with IAM enforcement I think. We can then remove the `lambda_name = arns.lambda_function_name(destination_arn)` line. 
This is how lambda uses the ARN:
```python
    def get_function(
        self,
        context: RequestContext,
        function_name: NamespacedFunctionName,
        qualifier: Qualifier = None,
    ) -> GetFunctionResponse:
        account_id, region = api_utils.get_account_and_region(function_name, context)
        function_name, qualifier = api_utils.get_name_and_qualifier(
            function_name, qualifier, context
        )
```

localstack/services/logs/provider.py
```diff
@@ -341,20 +348,30 @@ def moto_put_log_events(self: "MotoLogStream", log_events):
...
 
+            arn_data = parse_arn(destination_arn)
+
             if ":lambda:" in destination_arn:
-                client = connect_to(region_name=extract_region_from_arn(destination_arn)).lambda_
+                client = connect_to(
+                    aws_access_key_id=arn_data["account"], region_name=arn_data["region"]
+                ).lambda_
                 lambda_name = arns.lambda_function_name(destination_arn)
                 client.invoke(FunctionName=lambda_name, Payload=json.dumps(event))
```
bentsku, line 358: Same as commented above, I believe we shouldn't extract the lambda function name but use the ARN to invoke the function:
```python
client = connect_to(region_name=extract_region_from_arn(destination_arn)).lambda_
client.invoke(FunctionName=destination_arn, Payload=json.dumps(event))
```

localstack/services/logs/provider.py
```diff
@@ -341,20 +348,30 @@ def moto_put_log_events(self: "MotoLogStream", log_events):
...
-                client = connect_to().kinesis
+                client = connect_to(
+                    aws_access_key_id=arn_data["account"], region_name=arn_data["region"]
+                ).kinesis
                 stream_name = arns.kinesis_stream_name(destination_arn)
                 client.put_record(
                     StreamName=stream_name,
                     Data=payload_gz_encoded,
                     PartitionKey=self.log_group.name,
                 )
```
bentsku, line 369: For `PutRecord`, I believe it takes a `StreamARN` parameter, but I don't think it works with `kinesis-mock` and the one server for each account and the call to `get_aws_account_id`. Something to think about for the future and IAM maybe at some point.

bentsku (commented):
Nice changes! I have some comments regarding the lambda changes, I'd also like the opinion of @dfangl (for IAM, maybe future?) or maybe a lambda service owner, but I believe we should use the ARN there. Good catch for the S3 buckets as well, more cleanups! 🧹 

What about the default region changes? I believe some of the test logic was always expecting `us-east-1` to be the default, especially for S3, and snapshots might need to be regenerated.

dfangl:
Thanks for the tag @bentsku ! You are right, for cross account/region requests which are done on behalf of the service, we should definitely work on having the proper account just in the request body as the target resource, and use the request metadata to set the service name, instead of setting the account id as access key id. If we do the latter, IAM for example would not detect it as a cross-account request, as the originating account is the same as the targets. Ideally, all services extract the account id from the arn if provided, to access the correct account.

localstack/services/logs/provider.py
```diff
@@ -247,46 +246,61 @@ def moto_put_subscription_filter(fn, self, *args, **kwargs):
...
         raise ResourceNotFoundException("The specified log group does not exist.")
 
+    arn_data = arns.parse_arn(destination_arn)
+
+    if role_arn:
+        factory = connect_to.with_assumed_role(
+            role_arn=role_arn, service_principal=ServicePrincipal.logs
+        )
+    else:
+        factory = connect_to(aws_access_key_id=arn_data["account"], region_name=arn_data["region"])
```
bentsku, line 261: q: is this a fallback? this whole `with_assumed_role` addition is looking really nice 😄 I lack some knowledge in the IAM part, so this kind of things I don't feel qualified enough to review.

tests/aws/services/s3/test_s3.py
```diff
@@ -5379,14 +5380,13 @@ def secondary_client(self, secondary_aws_client):
 
     @markers.aws.unknown
     def test_shared_bucket_namespace(self, primary_client, secondary_client):
+        bucket_name = short_uid()
+
         # Ensure that the bucket name space is shared by all accounts and regions
-        primary_client.create_bucket(Bucket="foo")
+        create_s3_bucket(bucket_name=bucket_name, s3_client=primary_client)
```
bentsku, line 5386: this does not clean up I believe?

tests/aws/services/s3/test_s3.py
```diff
@@ -5396,13 +5396,13 @@ def test_cross_account_access(self, primary_client, secondary_client):
...
 
-        bucket_name = "foo"
+        bucket_name = short_uid()
         key_name = "lorem/ipsum"
         body1 = b"zaphod beeblebrox"
         body2 = b"42"
 
         # First user creates a bucket and puts an object
-        primary_client.create_bucket(Bucket=bucket_name)
+        create_s3_bucket(bucket_name=bucket_name, s3_client=primary_client)
```
bentsku, line 5405: same here, I believe this does not clean up

bentsku (approved):
LGTM! This is looking great! thanks for addressing the comments! There's just this part with the role and `with_assumed_role` and the `else` part which I'm not sure I understand the intricacies, but the rest is all good to me!

There's just 2 S3 tests that are not cleaning up due to the use of `localstack.utils.aws.resources.create_s3_bucket` now. Maybe we could add its cleanup to the `cleanups` fixture. 

Thanks a lot again for all of these fixes! 🚀

tests/aws/services/s3/test_s3.py
viren-nadkarni: Good catch, fixed in 4cf4e20cd

localstack/services/logs/provider.py
viren-nadkarni: You're correct, if a role ARN is available, IAM enforcement can be performed. Else there is a fallback.

This usage is similar to elsewhere, eg:

https://github.com/localstack/localstack/blob/05832fc4e4b0e23db05a023564986acb2d1d8c8a/localstack/services/sns/publisher.py#L646-L652

or

https://github.com/localstack/localstack/blob/05832fc4e4b0e23db05a023564986acb2d1d8c8a/localstack/services/firehose/provider.py#L734-L739

dfangl: IAM enforcement can be performed in both cases. However, _only_ if the aws_access_key_id is set to the `INTERNAL_AWS_ACCESS_KEY_ID` value. There should be no case, other than resource creation, to set the account id as access key id for internal clients, in my opinion. The first snippet by viren for example leads to no IAM enforcement in the else case, which is neither parity nor should happen.

PR #9545, opened 2023-11-03
apigateway: remove get_aws_account_id() from integration tests

sannya-singal:
## Motivation
The function `get_aws_account_id()` relies on the thread local storage. The LocalStack handler chain sets the right values for these functions on a per request basis.

This resolution fixes failing test `tests.aws.services.apigateway.test_apigateway_integrations` for non-default account and region.

## Changes
This PR removes the use of the `get_aws_account_id()` function and instead use `TEST_AWS_ACCOUNT_ID`

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_integrations.py
+++ b/tests/aws/services/apigateway/test_apigateway_integrations.py
@@ -11,7 +11,7 @@ from werkzeug import Request, Response
 
 from localstack import config
 from localstack.aws.accounts import get_aws_account_id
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM, this is a bandaid fix until we validate the test and create a proper role I believe.

sannya-singal:
closing this as there is a merged conflicting [PR](https://github.com/localstack/localstack/pull/9537) including same changes by @viren-nadkarni.

PR #9547, opened 2023-11-03
stabilized multi account for apigw test_invoke_method

macnev2013:
## Motivation
This PR improves the multi-account functionality for the API Gateway `test_invoke_method`. Previously, the `account_id` within the `invocation_context` was not being set and defaulted to "none". To rectify this, the `account_id` and `region` are now set from the `context` variable, with reference to the `invocation_context.api_id`.

## Changes
The `invocation_context` object's `account_id` and `region` parameters have been updated to reflect accurate information extracted from the API ID.

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111, TEST_AWS_ACCESS_KEY_ID=111111111111, TEST_AWS_REGION=us-west-1`). The affected tests are:
- `tests.aws.services.apigateway.test_apigateway_api.TestApiGatewayApi.test_invoke_test_method`

## TODO

- [ ] Validate the changes by running the -ext test suite

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -97,6 +97,7 @@ from localstack.services.apigateway.helpers import (
     ERROR_MODEL,
     OpenAPIExt,
     apply_json_patch_safe,
+    get_api_account_id_and_region,
     get_apigateway_store,
     get_regional_domain_name,
     import_api_from_openapi_spec,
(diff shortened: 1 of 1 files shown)


bentsku (commented):
Nice find! I however don't think that's how we should tackle the issue.
`TestInvokeMethod` is not multi-account/multi-region aware, it is dependent on the caller.
So we could simply use the context account and region here to set in the `invocation_context`, no need to call `get_api_account_id_and_region`.

I've done some tests:
```bash
$ aws --profile ls-sandbox --region us-east-1 apigateway test-invoke-method --rest-api-id aljmlzddv1 --resource-id j1apuk --http-method GET     
{
    "status": 200,
    "body": "{\"petId\": \"{petId}\"}",
    [... cutting short, long response]
}
$ aws --profile ls-sandbox --region us-east-2 apigateway test-invoke-method --rest-api-id aljmlzddv1 --resource-id j1apuk --http-method GET

An error occurred (NotFoundException) when calling the TestInvokeMethod operation: Invalid API identifier specified {accountid}:aljmlzddv1
```

bentsku (approved):
LGTM! Nice, thanks for addressing the comment. Could you update the PR description to reflect the change? Thanks!

macnev2013:
Thanks @bentsku  for the review and suggestions.

PR #9548, opened 2023-11-03
Allow setting numeric config variables in config update endpoint

whummer:
## Motivation

One of the Chaos Engineering features we want to enable in the Web app with @webdev51  is setting the `KINESIS_ERROR_PROBABILITY`/`DYNAMODB_ERROR_PROBABILITY` config variables, to simulate rate limiting for Kinesis/DynamoDB.

This can be enabled by asking the user to enable `ENABLE_CONFIG_UPDATES=1` on LS startup (with the caveats / security implications that come with it, and which we have also documented in our docs), and then using the config update endpoint.

However, currently updating the value to `ENABLE_CONFIG_UPDATES=1` is not fully working, as the service providers expect those values to be numeric, whereas the config update listener sets them as string. It seems reasonable to check if config values are numeric, and then convert them to float (the other approach would be to fix the logic in the provider, to be able to handle both `float` and `str` values).

## Changes

* add `is_number(..)` check to config update, to automatically convert numeric strings to numbers

### DIFF
--- a/tests/integration/test_config_endpoint.py
+++ b/tests/integration/test_config_endpoint.py
@@ -39,7 +39,7 @@ def test_config_endpoint(config_endpoint):
     body = {"variable": "FOO", "value": "BAR"}
     url = f"{config.get_edge_url()}{CONFIG_UPDATE_PATH}"
     response = requests.post(url, json=body)
-    assert 200 == response.status_code
+    assert response.ok
     response_body = response.json()
     assert body == response_body
     assert body["value"] == config.FOO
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
The change looks good!
I just feel like this could be somewhat of a rabbit hole without a standardized way of handling things (is the config responsible of the parsing or the providers?). However, the config update is a bit brittle, but somewhat safe if we - as the ones who know its limitations - are using it. 😄

tests/integration/test_config_endpoint.py
```diff
@@ -48,15 +48,22 @@ def custom_listener(config_key, config_value):
...
 
+    # test numeric value update
+    body = {"variable": "FOO", "value": 0.9}
+    response = requests.post(url, json=body)
+    assert response.ok
+    assert config.FOO == 0.9
+    assert isinstance(config.FOO, float)
+
     del config.FOO
+    config_listener.CONFIG_LISTENERS.remove(custom_listener)
```
bentsku, line 69: good catch!

bentsku (approved):
thanks a lot for adding the test, LGTM! 👍

PR #9552, opened 2023-11-03
[SFN] Fix reentrant wait function 

MEPalma:
## Motivation
Currently the reentrant wait function used by the interpreter for StepFunctions v2 incorrectly keeps track of the time elapsed between wait intervals. This PR addresses such issues.

## Changes
Simplified reentrant wait function, changed time tracking to use local epochs instead of date seconds.


alexrashed (approved):
Awesome! Thanks for jumping in this! 🚀

PR #9559, opened 2023-11-06
add services + strict_service_loading to analytics logging

steffyP:
## Motivation
With https://github.com/localstack/localstack/pull/9494 we introduced a new setting for strictly loading services. We want to keep track of how people are using this flag, and which services are chosen.

## Changes
add `SERVICES` and `STRICT_SERVICE_LOADING` to the tracked env vars.

/cc @bentsku


ackdav (approved):
lgtm :rocket: :+1:

PR #9560, opened 2023-11-06
s3: Fix broken multi account and region functionality

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, S3 tests should still create the consequent resources in this accounts and region.

## Changes
The PR fixes failing tests in `tests.aws.services.s3.test_s3.TestS3`


tests/aws/services/s3/test_s3.py
```diff
@@ -4251,16 +4257,21 @@ def test_create_bucket_head_bucket(self, snapshot, aws_client):
                 snapshot.transform.regex(r"s3\.amazonaws\.com", "<host>"),
                 snapshot.transform.regex(r"s3\.localhost\.localstack\.cloud:4566", "<host>"),
                 snapshot.transform.regex(r"s3\.localhost\.localstack\.cloud:443", "<host>"),
+                snapshot.transform.key_value("Location", reference_replacement=False),
```
bentsku, line 4260: This was a very specific snapshotting to check the format of the response, and now it's totally overridden by this new line. This test is testing the behaviour of `us-east-1` vs all other regions, so it should stay that way.

We need to have it hardcoded to `us-east-1` for the first bucket, and whatever region for the second, but it is really important.

The snapshot would fail anyway when run into multi-account because of the `if` check and the snapshot hard matching against `us-east-2`.

tests/aws/services/s3/test_s3.py
```diff
@@ -4251,16 +4257,21 @@ def test_create_bucket_head_bucket(self, snapshot, aws_client):
...
             ]
         )
 
         try:
-            response = aws_client.s3.create_bucket(Bucket=bucket_1)
+            response = create_s3_bucket(bucket_name=bucket_1, s3_client=aws_client.s3)
```
bentsku, line 4265: not sure we should use the `aws.resources.create_s3_bucket` here, maybe we could use a fixture instead of the big `try...finally` block. Not blocking though, this is okay for now but a consideration for the future.

bentsku (changes requested):

bentsku (approved):
LGTM! thanks for addressing the comments 🚀

tests/aws/services/s3/test_s3.py
```diff
@@ -4251,20 +4256,26 @@ def test_create_bucket_head_bucket(self, snapshot, aws_client):
...
             snapshot.match("create_bucket", response)
 
             response = aws_client.s3.create_bucket(
                 Bucket=bucket_2,
-                CreateBucketConfiguration={"LocationConstraint": "us-west-1"},
+                CreateBucketConfiguration={
+                    "LocationConstraint": "us-east-2"
+                    if TEST_AWS_REGION_NAME != "us-east-2"
+                    else "us-west-2"
+                },
```
viren-nadkarni, line 4274: I'm a bit confused by this bit: is the intention to use a different region than `TEST_AWS_REGION_NAME`? could we use `SECONDARY_TEST_AWS_REGION_NAME` since it will never be same as `TEST_AWS_REGION_NAME`?

bentsku: I suppose that's good as long as the second region is not us-east-1

sannya-singal: @viren-nadkarni I have made the required changes.

sannya-singal: @viren-nadkarni thanks for pointing this out, I have made the required changes.

viren-nadkarni: @sannya-singal would this be applicable at other places like [here](https://github.com/localstack/localstack/pull/9560/files#diff-9e996d8db48c7a9aab34a4fb63fdfc3e2e5ddc9da017632edc42407e121ac592R4813-R4815) or [here](https://github.com/localstack/localstack/pull/9560/files#diff-9e996d8db48c7a9aab34a4fb63fdfc3e2e5ddc9da017632edc42407e121ac592R4881-R4882)?

sannya-singal: This test is specifically comparing `Arn` here. We need to either write a transformer for reference replacement of the region or create region specific clients, the above implementation is just a workaround for now.

sannya-singal: @viren-nadkarni I have made the required changes here- https://github.com/localstack/localstack/pull/9570

tests/aws/services/s3/test_s3.py
```diff
@@ -4799,10 +4808,15 @@ def test_s3_sse_validate_kms_key(
...
         bucket_name = f"bucket-test-kms-{short_uid()}"
         s3_create_bucket(
-            Bucket=bucket_name, CreateBucketConfiguration={"LocationConstraint": "us-west-2"}
+            Bucket=bucket_name,
+            CreateBucketConfiguration={
+                "LocationConstraint": "us-west-2"
+                if TEST_AWS_REGION_NAME != "us-west-2"
+                else "me-south-1"
+            },
         )
```
bentsku, line 4817: it seems all of these changes conflict with #9570? Should they be removed?

bentsku (commented):
LGTM, but it seems there are still changes in `test_s3_sse_validate_kms_key` which would conflict with #9570?

tests/aws/services/s3/test_s3.py
sannya-singal: yes @bentsku, I will rebase this PR once [#9570](https://github.com/localstack/localstack/pull/9570) is green.

PR #9561, opened 2023-11-06
make strict service loading default for v3

bentsku:
## Motivation
Following the change merged in `master` with #9494, we now want to make it enabled by default with `v3.0`. 

## Changes

Make use of `is_env_not_false` to validate if the feature is not disabled (which makes it enabled by default). 
Added a test for the default enabling and disabling of the feature as well.


alexrashed (approved):
Nice and clean change of the defaults. I would classify this as a major change though. Maybe you could change the label?
However, this isn't critical, since it is targeting the release branch anyways :)

PR #9563, opened 2023-11-06
show available/disabled services in health endpoint + fix sqs strict loading

steffyP:
## Motivation
With `STRICT_SERVICE_LOADING` added in https://github.com/localstack/localstack/pull/9494, we can now start selected services only.
In this PR we fix the loading of `sqs` only (disable init of `cloudwatch` workers). 
Additionally, the `_localstack/health` endpoint returns the `available` and `disabled` services.

## Changes
* `SQS` provider: skip init of cloudwatch workers if the service `cloudwatch` is not enabled
*  return the correct `ServiceState` if the service is not available


baermat (approved):
🎉

bentsku (approved):
LGTM! Great test, it really shows nicely the feature now! Awesome really!! 🚀

PR #9566, opened 2023-11-06
Correct `DomainEndpointOptions` in opensearch/elasticsearch responses and persisted state

### LINKED ISSUE #9556: bug: opensearch/elasticsearch custom endpoint not saved or included in responses

### Current Behavior

When creating an opensearch/elasticsearch domain, a custom endpoint URL may be set on the domain via the `CustomEndpoint`/`CustomEndpointEnabled` request params. localstack respects these params when initially setting up the domain and correctly routes requests made to the custom endpoint, but it doesn't "remember" this config, which has two effects:

1. The response from the `CreateElasticsearchDomain` (and any subsequent `DescribeElasticsearchDomain`) request incorrectly has `CustomEndpoint` omitted and `CustomEndpointEnabled` set to `false`
2. When restoring an opensearch/elasticsearch domain from persisted state, any custom endpoint config is lost/ignored, and making a request to the custom endpoint returns a 404 (of course this is also broken by #8092, but this bug would still happen even if that one is fixed!)

With a docker-compose file

### Steps To Reproduce

```
awslocal es create-elasticsearch-domain \
  --domain-name example \
  --domain-endpoint-options CustomEndpointEnabled=true,CustomEndpoint=example.com:1234/foo
```
This gives the following response (irrelevant properties omitted):
```jsonc
{
    "DomainStatus": {
        "Endpoint": "example.com:1234/foo",
        "DomainEndpointOptions": {
            "EnforceHTTPS": false,
            "TLSSecurityPolicy": "Policy-Min-TLS-1-0-2019-07",
            "CustomEndpointEnabled": false
        },
        ...
    }
}
```

Note that `CustomEndpointEnabled` is false when it should be true, and `CustomEndpoint` is omitted when it should be `example.com:1234/foo` (the same as the `Endpoint` property). 

The same response is given when subsequently running `awslocal es describe-elasticsearch-domain --domain-name example`

### Environment

```markdown
- OS: docker desktop (on a windows 11 host)
- LocalStack: latest (2.3.3.dev)
```

### Anything else?

_No response_

GREsau:
## Motivation
Fixes #9556

This PR makes `CreateDomain`/`DescribeDomain`/`DescribeDomainConfig` responses include the `DomainEndpointOptions` from the request that created the domain, instead of returning a hard-coded response with `"CustomEndpointEnabled": false`. This also makes custom endpoints work for domains that were loaded from persisted state.

## Changes

I also brought the `CreateDomain` request validation more into line with real AWS, in that it now requires `CustomEndpoint` to be set if and only if `CustomEndpointEnabled` is true. The previous behaviour was to silently ignore the request's custom endpoint config when only one of these params was given/enabled. This means that, while the new validation is more similar to AWS, it is technically a breaking change.

AWS actually also requires `CustomEndpointCertificateArn` to be set, but setting a custom endpoint without a certificate is a legitimate use-case in localstack ([and is even shown as working in the docs](https://docs.localstack.cloud/user-guide/aws/opensearch/#custom-endpoints)), and localstack doesn't actually respect the `CustomEndpointCertificateArn` param anyway - so it didn't seem sensible to require that param too.

You may notice I deleted this code, since it looked like a mistake to me:
```python
if stored_status.get("Endpoint"):
    new_status["Endpoint"] = new_status.get("Endpoint")
```
Maybe it was intended to be `new_status["Endpoint"] = stored_status.get("Endpoint")`? But even that wouldn't make much sense because that's effectively the initial value of `new_status["Endpoint"]`!

## Testing
Existing tests were extended to validate the domain endpoint options in the `DescribeDomain` response, and new tests were added to cover the new request validation.

To manually verify the change, run:
```
awslocal es create-elasticsearch-domain \
  --domain-name example \
  --domain-endpoint-options CustomEndpointEnabled=true,CustomEndpoint=example.com:1234/foo \
  --query DomainStatus.DomainEndpointOptions
```

That command now returns:

```json
{
    "EnforceHTTPS": false,
    "TLSSecurityPolicy": "Policy-Min-TLS-1-0-2019-07",
    "CustomEndpointEnabled": true,
    "CustomEndpoint": "example.com:1234/foo"
}
```

Previously, it returned
```json
{
    "EnforceHTTPS": false,
    "TLSSecurityPolicy": "Policy-Min-TLS-1-0-2019-07",
    "CustomEndpointEnabled": false
}
```

### DIFF
--- a/localstack/services/opensearch/provider.py
+++ b/localstack/services/opensearch/provider.py
@@ -117,6 +117,12 @@ DEFAULT_OPENSEARCH_CLUSTER_CONFIG = ClusterConfig(
     DedicatedMasterCount=1,
 )
 
+DEFAULT_OPENSEARCH_DOMAIN_ENDPOINT_OPTIONS = DomainEndpointOptions(
+    EnforceHTTPS=False,
+    TLSSecurityPolicy=TLSSecurityPolicy.Policy_Min_TLS_1_0_2019_07,
+    CustomEndpointEnabled=False,
+)
+
 
 def cluster_manager() -> ClusterManager:
     global __CLUSTER_MANAGER
@@ -355,18 +361,13 @@ def get_domain_status(domain_key: DomainKey, deleted=False) -> DomainStatus:
             AutomatedUpdateDate=datetime.fromtimestamp(0, tz=timezone.utc),
             OptionalDeployment=True,
         ),
-        DomainEndpointOptions=DomainEndpointOptions(
-            EnforceHTTPS=False,
-            TLSSecurityPolicy=TLSSecurityPolicy.Policy_Min_TLS_1_0_2019_07,
-            CustomEndpointEnabled=False,
-        ),
+        DomainEndpointOptions=stored_status.get("DomainEndpointOptions")
+        or DEFAULT_OPENSEARCH_DOMAIN_ENDPOINT_OPTIONS,
         AdvancedSecurityOptions=AdvancedSecurityOptions(
             Enabled=False, InternalUserDatabaseEnabled=False
         ),
         AutoTuneOptions=AutoTuneOptionsOutput(State=AutoTuneState.ENABLE_IN_PROGRESS),
     )
-    if stored_status.get("Endpoint"):
-        new_status["Endpoint"] = new_status.get("Endpoint")
     return new_status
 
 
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Great contribution! Thanks for jumping on this! The provider changes and the test adoptions are looking good, I'll merge this PR right away. 💯

PR #9568, opened 2023-11-07
make SNS Delivery Logs opt-in 

bentsku:
## Motivation
We had a TODO in the code because SNS Delivery logs were always-on for LocalStack when it's opt-in in AWS.
Also it has been shown they are making performance quite a bit worse.

Here are some numbers:

  | Base | Disable SQS metrics | Disable SNS Delivery logs | Both
-- | -- | -- | -- | --
Throughput for 1kb publish (req/s) | 20.17 | 28.72 | 31.87 | 68.78

- Base: `1000 requests took 49.5744 s = 49.5744 ms/op = 20.17 req/sec`
- Disable SQS metrics: `1000 requests took 34.8189 s = 34.8189 ms/op = 28.72 req/sec`
- Disable SNS Delivery logs: `1000 requests took 31.3768 s = 31.3768 ms/op = 31.87 req/sec`
- Both: `1000 requests took 14.5392 s = 14.5392 ms/op = 68.78 req/sec`

From a simple benchmark with one topic, one SQS queue subscribed to it, pushing 1000 messages/

I suspect the fact that the publisher is pushing a lot of SQS requests concurrently, plus the CloudWatch Events one, are making the gateway struggle to serve them concurrently and the latency grows quite significantly.

To push further, we will revisit the in-memory client dispatching which might provide a boost in performance, and we now have a good simple benchmark for it.

Here's all resources:
- https://docs.aws.amazon.com/sns/latest/dg/sns-topic-attributes.html#msg-status-sdk
- https://docs.aws.amazon.com/sns/latest/dg/sns-msg-status.html
- https://docs.aws.amazon.com/sns/latest/dg/sms_stats_cloudwatch.html

## Changes
Make publishing the delivery logs opt-in like AWS, by setting a role as a Topic Attribute (`<ENDPOINT>SuccessFeedbackRoleArn` and `<ENDPOINT>FailureFeedbackRoleArn`).
We do not support yet `<ENDPOINT>SuccessFeedbackSampleRate` and will publish every message without sampling. 

For SMS messages, I left the behaviour as always-on, as the documentation only mention the console, and it's a niche use-case.

We have an AWS validated test `tests.aws.services.sns.test_sns.TestSNSPublishDelivery.test_delivery_lambda` testing this behaviour. We should add test to this class for SQS, HTTP and Firehose deliveries as well. 

Also sneaked a small change regarding unsubscribe + an AWS test for it.

### DIFF
--- a/localstack/services/sns/provider.py
+++ b/localstack/services/sns/provider.py
@@ -66,6 +66,7 @@ from localstack.utils.aws.arns import (
     extract_region_from_arn,
     parse_arn,
 )
+from localstack.utils.bootstrap import log_duration
 from localstack.utils.strings import short_uid
 
 # set up logger
@@ -129,7 +130,7 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         # see the attributes we need: https://docs.aws.amazon.com/sns/latest/dg/sns-topic-attributes.html
         # would need more work to have the proper format out of moto, maybe extract the model to our store
         for attr in vars(moto_topic_model):
-            if "success_feedback" in attr:
+            if "_feedback" in attr:
                 key = camelcase_to_pascal(underscores_to_camelcase(attr))
                 moto_response["Attributes"][key] = getattr(moto_topic_model, attr)
             elif attr == "signature_version":
@@ -221,10 +222,12 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
                 success["SequenceNumber"] = msg_ctx.sequencer_number
             response["Successful"].append(success)
 
+        moto_topic = self._get_topic(topic_arn, context)
         publish_ctx = SnsBatchPublishContext(
             messages=message_contexts,
             store=store,
             request_headers=context.request.headers,
+            topic_attributes=vars(moto_topic),
         )
         self._publisher.publish_batch_to_topic(publish_ctx, topic_arn)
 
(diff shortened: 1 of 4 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -833,6 +816,39 @@ def subscribe_queue_to_topic(attributes: dict = None) -> dict:
...
+        queue_name = f"queue-{short_uid()}"
+        topic_arn = sns_create_topic(Name=topic_name)["TopicArn"]
+        queue_url = sqs_create_queue(QueueName=queue_name)
+        subscription = sns_create_sqs_subscription(topic_arn=topic_arn, queue_url=queue_url)
+        sub_arn = subscription["SubscriptionArn"]
+
+        unsubscribe_1 = aws_client.sns.unsubscribe(SubscriptionArn=sub_arn)
+        snapshot.match("unsubscribe-1", unsubscribe_1)
+        unsubscribe_2 = aws_client.sns.unsubscribe(SubscriptionArn=sub_arn)
+        snapshot.match("unsubscribe-2", unsubscribe_2)
```
alexrashed, line 833: This test here seems to be valuable, but it I'm missing a test which actually tests the changes in the provider.
I can see that the positive case (the subscriptions are enabled and working) are tested in multiple tests here, but I think it would be great to also have a test checking that the logs are not written in case the logs are not explicitly enabled?

alexrashed (approved):
The changes are looking good, but I think it would be great if you could add a test for disabled logs. However, this shouldn't block a merge, we could also add this in a later iteration. 
Nice to see this awesome speedup! 🚀

tests/aws/services/sns/test_sns.py
bentsku: True yes, that's a very good point! I just need to decide a fair value to estimate that the logs will never arrive and that we can assert the test is successful without adding a 30s test to the pipeline. But that would be very valuable to have!

localstack/services/sns/provider.py
```diff
@@ -358,21 +360,23 @@ def create_platform_endpoint(
...
             parsed_arn = parse_arn(subscription_arn)
         except InvalidArnException:
             # TODO: check for invalid SubscriptionGUID
-            if count < 6:
-                raise InvalidParameterException(
-                    f"Invalid parameter: SubscriptionArn Reason: An ARN must have at least 6 elements, not {count}"
-                )
+            raise InvalidParameterException(
+                f"Invalid parameter: SubscriptionArn Reason: An ARN must have at least 6 elements, not {count}"
+            )
```
baermat, line 373: It seems a bit weird to have this code twice. Is there a particular reason why we check the count ourselves and then let moto do the conversion with the same check again?

baermat (approved):
LGTM 👍 There seem to be some duplication of code, but nothing blocking

localstack/services/sns/provider.py
bentsku: Right yes just checked the botocore code, and they also validate the `6 elements` part. Will rewrite and update the test like Alex said at the same time. Thanks a lot! 👍

tests/aws/services/sns/test_sns.py
bentsku: I added the test now, with some snapshot asserting that the first message wasn't logged and that the second message was properly logged, matching `messageId` to the events. Thanks for the review!

PR #9569, opened 2023-11-07
add in-memory ASF request dispatching for boto clients

bentsku:
## Motivation
Taken from #7396 from @thrau:

> This PR adds code to make calls from boto clients directly to a Gateway without an HTTP server roundtrip. This can make internal calls much faster because they don't require IO.
> 
> The mechanism used to achieve this is botocore's `before-send` event, which can be used to short-circuit the request logic by returning a response from the event handler, before the HTTP request is even made.
> 
> I did some basic refactoring to make this easier, including:
> * ~created `localstack.runtime.components` as a place for runtime-level singleton components~
> * added a new method to `Gateway` with the signature `def handle(self, context: RequestContext, response: Response)`, which now makes the Gateway a drop-in replacement for a handler chain (and easier to call if you want to invoke it with a pre-existing RequestContext).

I believe we could merge this as is, as it's only opt-in, and we could improve in follow-up iterations. We'd at least get a preview version of it, and we can give it more tries. 

Currently, the feature can be enabled with `IN_MEMORY_CLIENT=1`.

## Benchmarks:
Results:

  | Base | In-memory client | Everything (#9568) + in-memory
-- | -- | -- | --
Throughput for 1kb publish (req/s) | 20.17 | 28.98 | 102.77

In-memory only with cloudwatch SQS metrics + SNS delivery logs: `1000 requests took 34.5011 s = 34.5011 ms/op = 28.98 req/sec`
All fixes: `1000 requests took 9.7307 s = 9.7307 ms/op = 102.77 req/sec`

## Testing
Ext Integration Tests 6506 ✅ 
Successful run with the flag enabled: https://app.circleci.com/pipelines/github/localstack/localstack/25062/workflows/401a985e-e287-483e-82f3-229daf251654

### TODO:
- [X] basic PoC implementation
- [X] integration with the client factory and `connect_to`
- [X] more benchmarks to see whether/where this is really useful
- [ ] documentation


thrau:
@bentsku i rebased the PR to use the new runtime framework

thrau:
i've added @dfangl as reviewer, as i partly wrote the PR, so ben and I are already aligned. i'd like to just know whether it's integrated into the client factory appropriately

localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
```diff
@@ -604,7 +604,8 @@ def _create_bucket_if_does_not_exist(self, model, region_name, s3_client):
         try:
             s3_client.head_bucket(Bucket=model["BucketName"])
         except ClientError as e:
-            if e.response["Error"]["Message"] != "Not Found":
+            # TODO: check why the ClientError is not the same, something with Boto
+            if e.response["ResponseMetadata"]["HTTPStatusCode"] != 404:
```
dfangl, line 608: Did you find out why this is necessary?

localstack-core/localstack/aws/gateway.py
```diff
@@ -25,3 +27,13 @@ def __init__(
...
+
+    def handle(self, context: RequestContext, response: Response) -> None:
+        """Exposes the same interface as ``HandlerChain.handle``."""
+        return self.new_chain().handle(context, response)
+
+    def process(self, request: Request, response: Response):
+        # TODO: this overrides might not be necessary
+        context = self.context_class(request)
+
+        self.handle(context, response)
```
dfangl, line 39: This looks identical to upstream?

localstack-core/localstack/config.py
```diff
@@ -1100,6 +1100,9 @@ def use_custom_dns():
 
 DISTRIBUTED_MODE = is_env_true("DISTRIBUTED_MODE")
 
+# This flag enables `connect_to` to be in-memory only and not do networking calls
+IN_MEMORY_CLIENT = is_env_true("IN_MEMORY_CLIENT")
```
dfangl, line 1104: We should add this to `CONFIG_ENV_VARS` as well!

localstack-core/localstack/aws/client.py
```diff
@@ -221,3 +272,74 @@ def raise_service_exception(response: Response, parsed_response: Dict) -> None:
...
+        if operation.service_model.service_name not in {
+            "sqs-query",
+            "docdb",
+            "neptune",
+            "timestream-write",
+        }:
+            context.service = operation.service_model
+
+        context.operation = operation
+        context.service_request = params["body"]
```
dfangl, line 327: So the request parameters are always stored in "body"? Interesting :) If I undestand this correctly, this should let us avoid service detection and service request parsing, right?

dfangl (approved):
I think this looks good! I would be a tad happier if the `_create_and_enrich_request` patch would be behind the feature flag as well, but I get that that will clutter the whole thing a tad more (since we probably need to use a hook then?).

Looks nicely, had only minor nits, which I am happy with not fixing as well for now :)

localstack-core/localstack/aws/gateway.py
bentsku: Yes, I think this was done earlier and was left in, but should be removed now. Good catch!

localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
bentsku: Not exactly, but I believe `HeadBucket` is a bit problematic, I remember having some issue in an S3 test with it not really raising the right error too. I could try to look more in detail again why this happened 🤔

localstack-core/localstack/aws/client.py
bentsku: So it seems 😄 it exactly does, yes! It skips the service detection, but as the hacky way above demonstrate, it can lead to a few issues so I don't set it there on purpose for those services.

bentsku (commented):
Thanks @dfangl for the review!

> I would be a tad happier if the _create_and_enrich_request patch would be behind the feature flag as well, but I get that that will clutter the whole thing a tad more (since we probably need to use a hook then?).

I think we might not need a hook, it might look a bit messy, but maybe just adding a switch to the config flag should do it, no?

something like
```python
if localstack_config.IN_MEMORY_CLIENT:
    @patch(Endpoint.create_request)
    def _create_and_enrich_request(
        fn, self: Endpoint, params: dict, operation_model: OperationModel = None
    ):
```

or 

```python
def _create_and_enrich_request(
    fn, self: Endpoint, params: dict, operation_model: OperationModel = None
):

if localstack_config.IN_MEMORY_CLIENT:
     patch = Patch.function(Endpoint.create_request, _create_and_enrich_request))
     patch.apply()
```

I can add that to the PR, I guess it could avoid some issues we might not have expected? 

I'll also address the other comments in the meantime.

bentsku (commented):
Sorry @dfangl for asking for a quick review again, if you could review the patching commits, then I think we could merge 😄 thank you!

dfangl (approved):
Thanks for addressing my comments, and the conditional patching! LGTM!

thrau (approved):
can't believe this is ready to ship now :heart_eyes: thanks ben for taking up the initial PR and getting it over the line. it's great that we have it hidden behind a feature flag for now. it'll be great to start rolling this out to people who are known to have performance issues :+1:

PR #9498, opened 2023-10-27
Fix key error in snapshot matching of parametrized test

joe4dev:
## Motivation

The Lambda test `test_assume_role[2]` was flaky quite a bit in the past days. Recent example failing CI run on master:
* https://app.circleci.com/pipelines/github/localstack/localstack/19457/workflows/06e37713-6771-4e70-bed4-35d06ae5d57e/jobs/151885/tests
* https://app.circleci.com/pipelines/github/localstack/localstack/19382/workflows/da518a79-1a89-4a0d-9036-1defc99066e8/jobs/151207/tests

The error `KeyError: 'invoke-result-assumed-role-arn'` might indicate some race condition in the snapshot testing framework (or maybe a user error 😄 ) when using parametrized tests and snapshotting results that are expected to be the same across parameters.

## Changes

Parametrize the snapshot key with the parameter of the test and re-generate snapshot.

## Background

**EDIT**: As indicated with *might*, I don't know yet what really causes this issue. I cannot reproduce it locally even after 60+ retries of the test using this script:

```bash
COUNTER=0

true
while [ $? -eq 0 ]
do
	COUNTER=$(( COUNTER + 1 ))
	echo "Attempt $COUNTER"
	LAMBDA_DEV_PORT_EXPOSE=1 TEST_PATH="tests/aws/services/lambda_/test_lambda.py::TestLambdaBaseFeatures::test_assume_role" make test
done
```

Creating a minimal snapshot test also doesn't reproduce it:

```python
    @markers.aws.validated
    @pytest.mark.parametrize("execution_number", range(1000))
    def test_snapshot_params(self, execution_number, snapshot):
        snapshot.match("my-key", "myvalue")
```

Feel free to comment if you have any ideas 🤔 ?

PR #9574, opened 2023-11-07
Fix snapshot reference replacement replacing parts of the snapshot keys

dfangl:
Supersedes #9498 

## Motivation
We regularly have the issue that the snapshot reference replacements also act upon the keys of the snapshot. This is especially annoying if it introduces test flakyness due to randomized variable names - if the name per chance matches up with a part of the key name, it will be replaced.
This is especially true for short resource names, which are likely to be present in the snapshot key name.

## Changes
The reference replacement never replaces the top level snapshot key - only everything in the actual snapshot values.


tests/aws/services/lambda_/test_lambda.snapshot.json
```diff
@@ -3456,15 +3456,15 @@
     }
   },
   "tests/aws/services/lambda_/test_lambda.py::TestLambdaBaseFeatures::test_assume_role[1]": {
-    "recorded-date": "10-10-2023, 14:55:53",
+    "recorded-date": "07-11-2023, 17:06:55",
     "recorded-content": {
-      "invoke-result-assumed-role-arn": "arn:aws:sts::111111111111:assumed-role/lambda-autogenerated-<lambda-autogenerated-role-prefix>/<resource:1>"
+      "invoke-result-assumed-role-arn": "arn:aws:sts::869636330914:assumed-role/lambda-autogenerated-<lambda-autogenerated-role-prefix>/<function-name>@lambda_function"
```
joe4dev, line 3461: I guess we still need a transformer for the account here after clearing the default transformers 🤔

dominikschubert, line 3461:
```suggestion
      "invoke-result-assumed-role-arn": "arn:aws:sts::111111111111:assumed-role/lambda-autogenerated-<lambda-autogenerated-role-prefix>/<function-name>@lambda_function"
```

the clear operation also removed the account id replacement

dfangl: Yes, totally missed that one. Corrected!

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -276,8 +278,14 @@ def test_assume_role(self, create_lambda_function, aws_client, snapshot, functio
...
         """
         lambda_autogen_role = "(?<=lambda-autogenerated-)([0-9a-f]{8})"
+        # avoid any other transformers possible registering reference replacements (especially resource transformer)
+        snapshot.transformers.clear()
         snapshot.add_transformer(
-            snapshot.transform.regex(lambda_autogen_role, "<lambda-autogenerated-role-prefix>")
+            [
+                snapshot.transform.regex(account_id, "1" * 12),
+                snapshot.transform.regex(lambda_autogen_role, "<lambda-autogenerated-role-prefix>"),
+                snapshot.transform.regex(r'(?<=/)[a-zA-Z]{1,2}(?="|@)', "<function-name>"),
```
joe4dev, line 287: much more elegant than the hacky conditional assertions 🥇 👏👏👏 

We can get rid of the `if function_name_length == 1:` conditional at the end of the test now.

steffyP (approved):
Thanks for the fix! 🙏

PR #9576, opened 2023-11-07
update api-dependencies and check cloudwatch/logs enabled

steffyP:
## Motivation
Previous PR https://github.com/localstack/localstack/pull/9494 enabled the loading of specific services only. 
In this PR we update the `API_DEPENDENCIES` for community, and added checks for before calling non-required services.

## Changes
* **Lambda**
    * prevent `EventSourceListener` to start if the event-source service is not loaded
    * skip the start of `LogHandler` thread, if `logs` is not enabled
    
* **CloudWatch-Util**
   * added checks before calling service (logs + cloudwatch)
   * the  `publish_lambda_metric` is also used by `lambda`
 
* **Logs**
  * skip metric-filter if `cloudwatch` is disabled

* updated `API_DEPENDENCIES`


localstack/services/lambda_/invocation/logs.py
```diff
@@ -62,10 +63,15 @@ def run_log_loop(self, *args, **kwargs) -> None:
                 )
 
     def start_subscriber(self) -> None:
+        if not is_api_enabled("logs"):
```
steffyP, line 66: just preventing the thread start here. we can discuss if there are better approaches to handle this when `logs` are disabled.

localstack/utils/bootstrap.py
```diff
@@ -44,10 +44,9 @@
 # maps from API names to list of other API names that they depend on
 API_DEPENDENCIES = {
     "dynamodb": ["dynamodbstreams"],
-    "dynamodbstreams": ["kinesis"],
+    "dynamodbstreams": ["kinesis", "dynamodb"],
     "es": ["opensearch"],
-    "lambda": ["logs", "cloudwatch", "s3", "sqs"],
-    "kinesis": ["dynamodb"],
+    "lambda": ["s3", "sqs", "sts"],
```
bentsku, line 49: good catch! 👍

localstack/utils/bootstrap.py
```diff
@@ -44,10 +44,9 @@
 # maps from API names to list of other API names that they depend on
 API_DEPENDENCIES = {
     "dynamodb": ["dynamodbstreams"],
-    "dynamodbstreams": ["kinesis"],
+    "dynamodbstreams": ["kinesis", "dynamodb"],
```
bentsku, line 47: I suppose the dependency is the other way around, if somebody wants to use `dynamodbstreams` then they have to enable `dynamodb`, I don't think we should enable it for them.

bentsku (approved):
LGTM! This is awesome! Also, great catch for `sts`. 🚀

localstack/utils/bootstrap.py
bentsku: Also, I don't think this is blocking or anything, we can update the list later on!

steffyP: that makes sense. i will remove it again :)

joe4dev (approved):
This PR covers the basic Lambda scenarios 👍
It's gonna be tough to maintain these conditionals for many service combinations.

I tested the provided testing scenario with `SERVICES=lambda` and this PR fixes Lambda create-function, invoke, and (dummy) event source mapping (already create-function would fail without this PR) 👏👏👏 .

⚠️ I assume DLQs and success/failure destinations would fail with an exception in the `event_manager.py` 🤔 Would it provide some helpful logging output?

I guess one of the relevant scenarios for 3.0 is related to users who have set `SERVICES` since a long time before it got re-purposed. Obviously, any user code depending on other services within Lambda could fail somewhat implicitly but this likely falls under misconfiguration.


Out of the scope for this PR but I just wanted to comment on integrations to consider for ext:
* X-Ray
* and potentially ECR for pulling Lambda
* IAM

localstack/services/lambda_/event_source_listeners/event_source_listener.py
```diff
@@ -40,6 +44,12 @@ def start_listeners_for_asf(event_source_mapping: Dict, lambda_service: LambdaSe
             )
             if self_managed_endpoints.get("KAFKA_BOOTSTRAP_SERVERS"):
                 service_type = "kafka"
+        elif not is_api_enabled(service_type):
+            LOG.info(
```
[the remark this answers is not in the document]
steffyP: IMO `info` should be fine, because it's impossible to pass a valid/existing source-arn here (otherwise the service would be available).

PR #9578, opened 2023-11-07
Update remaining usages of LOCALHOST_HOSTNAME

simonrw:
## Motivation

Some usages of `LOCALHOST_HOSTNAME` were missed during the migration to `LOCALSTACK_HOST`. This PR addresses that.

Note: usages of _returning_ `LOCALHOST_HOSTNAME` are changed, not internal usages of `LOCALHOST_HOSTNAME` to return to `localhost`. This is important, because `LOCALSTACK_HOST` is a "cosmetic" variable, and as such should not be used as a value to reach `localhost`.

## Changes

* Update external usages of `LOCALHOST_HOSTNAME` to use `LOCALSTACK_HOST`.
* Move Route53 hosted zone fixture to a common location for use in the network configuration tests in LocalStack Pro
* `LOCALSTACK_HOST` no longer configures the edge port if `GATEWAY_LISTEN` is _not_ specified
    * this was weird behaviour, and a confusing mix of cosmetic and functional variables

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -603,16 +603,6 @@ def populate_edge_configuration(
     localstack_host_raw = environment.get("LOCALSTACK_HOST")
     gateway_listen_raw = environment.get("GATEWAY_LISTEN")
 
-    # new for v2
-    # get the potentially set port from LOCALSTACK_HOST first to use for gateway listen
-    localstack_host_port = constants.DEFAULT_PORT_EDGE
-    if localstack_host_raw is not None:
-        localstack_host_port = HostAndPort.parse(
-            localstack_host_raw,
-            default_host=constants.LOCALHOST_HOSTNAME,
-            default_port=constants.DEFAULT_PORT_EDGE,
-        ).port
-
     # parse gateway listen from multiple components
     if gateway_listen_raw is not None:
         gateway_listen = []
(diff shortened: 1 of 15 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -619,7 +618,8 @@ def host_based_url(rest_api_id: str, path: str, stage_name: str = None):
 
 def get_execute_api_endpoint(api_id: str, protocol: str = "") -> str:
     port = config.get_edge_port_http()
```
dfangl, line 620: Can we change this to localstack_host.host_and_port instead of using get_edge_port_http?

dfangl (approved):
LGTM!

localstack/services/apigateway/helpers.py
joe4dev: mhm, maybe that's a sign to remove `get_edge_port_http()` anyways with 3.0 and force extensions to update 🤔 
It's a bit short notice and uncool that all extensions need to be re-released but it might be nicer to break it with a major change ...
I could tackle it in a follow up if we wanna go this router 🎢

joe4dev (approved):
LGTM 👍

localstack/services/apigateway/helpers.py
simonrw: I agree we should use `localstack_host().host_and_port()` in this case 👍

tests/unit/test_config.py
```diff
@@ -75,114 +82,68 @@ class TestEdgeVariablesDerivedCorrectly:
...
-    def test_defaults(self, default_ip):
+    # This parameterised test forms a table of scenarios we need to cover. Each
+    # input variable (gateway_listen, localstack_host) has four unique
+    # combinations of inputs:
+    # * default
+    # * host only
+    # * ip only
+    # * host and ip
+    # and there are two variables so 16 total tests
+    @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
joe4dev: It would also simplify removing `EDGE_PORT` from `populate_edge_configuration` if we wish to do so later 👍

localstack/config.py
```diff
@@ -603,16 +603,6 @@ def populate_edge_configuration(
     localstack_host_raw = environment.get("LOCALSTACK_HOST")
     gateway_listen_raw = environment.get("GATEWAY_LISTEN")
 
-    # new for v2
```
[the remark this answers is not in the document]
joe4dev: It would be interesting to know why we added it in the first place but I guess that remains unknown 🪄

joe4dev (approved):
More explicit, clearer, and easier 💯  LGTM 🚀 

(the major change is a bit sneaky in the usage update PR ;)

MEPalma (approved):

simonrw:
> (the major change is a bit sneaky in the usage update PR ;)

I've put in the PR description, so it's fine right?! 🤷

dfangl (approved):
Love it! Test cases look good as well, i think it's quite clear what's going on.

PR #9580, opened 2023-11-08
s3: add regex transformer for different regions

sannya-singal:
## Motivation
When using non-default values for account and region, S3 tests should still create the consequent resources in this accounts and region. If the region is set as `us-west-2` in `TEST_AWS_REGION_NAME` and we create a `KMS` client for the same region I get kms key ARN as `arn:aws:kms:<region>:111111111111:key/<uuid:1>` i.e. the value for `us-west-2` is replaced by the default region regex transformer `<region>`. But now if we create a KMS client say for some other region other than the `TEST_AWS_REGION_NAME`  assume `us-west-1` the value of the ARN is `arn:aws:kms:us-west-1:111111111111:key/<uuid:1>`. 

## Changes
This PR adds a `RegexTransformer` to both the hardcoded regions and updates the corresponding snapshot for `tests.aws.services.s3.test_s3.TestS3.test_s3_sse_validate_kms_key`.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -48,6 +48,7 @@ from localstack.services.s3.utils import (
 )
 from localstack.testing.aws.util import is_aws_cloud
 from localstack.testing.pytest import markers
+from localstack.testing.snapshots.transformer import RegexTransformer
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM, this is a limitation of the snapshot framework, but good to see it addressed in this test! 👍

PR #9584, opened 2023-11-08
Remove SQS_PORT_EXTERNAL and unify networking helpers

joe4dev:
## Motivation

* `<SERVICE>_PORT_EXTERNAL` has been deprecated since 2022-07-13 in [this docs PR](https://github.com/localstack/docs/pull/201/files#diff-0502d352f33bf9d3db34ba16e95a02f45a111e7f9029f0577c2f37749c148d9fR328). However, it has never made it into `deprecations.py`. The migrations of `LOCALSTACK_HOST` offers a consistent way throughout LocalStack to achieve the same goal and should be preferred over service-specific solutions such as `SQS_PORT_EXTERNAL`.
* The migrations of `LOCALSTACK_HOST` (https://github.com/localstack/localstack/pull/9390) and removal of the legacy edge variables (https://github.com/localstack/localstack/pull/9405) has left some networking helpers in `config.py` in an inconsistent state.
  * Some combine elements of internal and external usage (e.g., `get_edge_url()`). This breaks the proper `LOCALSTACK_HOST` configuration.
  * Internal and external usage was previously not separated. Therefore, `LOCALSTACK_HOST` is not yet fully supported and we need to inspect all usages of these deprecated helpers to fix this.
* `SQS_PORT_EXTERNAL` is used in one of these helpers `service_port()` and therefore these changes are related (at least partially).

## Changes

* Remove `SQS_PORT_EXTERNAL`
* Unify the networking helpers in `config.py` by adding a deprecation path and attempting to fix inconsistencies following the migrations of `LOCALSTACK_HOST` and edge variables.

## Testing

* Adjusted tests for `SQS_PORT_EXTERNAL` to use `LOCALSTACK_HOST` instead for achieving the same behavior.
* Our tests have insufficient coverage of configuration scenarios for `LOCALSTACK_HOST` usage: internal vs. external ⚠️ 

## Follow up

We need to go over all usages of these deprecated helpers and migrate them to either `external_service_url` or `internal_service_url`.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -821,9 +821,6 @@ SQS_DELAY_PURGE_RETRY = is_env_true("SQS_DELAY_PURGE_RETRY")
 # Used to toggle QueueDeletedRecently errors when re-creating a queue within 60 seconds of deleting it
 SQS_DELAY_RECENTLY_DELETED = is_env_true("SQS_DELAY_RECENTLY_DELETED")
 
-# expose SQS on a specific port externally
-SQS_PORT_EXTERNAL = int(os.environ.get("SQS_PORT_EXTERNAL") or 0)
-
 # Strategy used when creating SQS queue urls. can be "off", "standard" (default), "domain", or "path"
 SQS_ENDPOINT_STRATEGY = os.environ.get("SQS_ENDPOINT_STRATEGY", "") or "standard"
 
@@ -1178,7 +1175,6 @@ CONFIG_ENV_VARS = [
     "SQS_DELAY_PURGE_RETRY",
     "SQS_DELAY_RECENTLY_DELETED",
     "SQS_ENDPOINT_STRATEGY",
-    "SQS_PORT_EXTERNAL",
     "SQS_DISABLE_CLOUDWATCH_METRICS",
     "SQS_CLOUDWATCH_METRICS_REPORT_INTERVAL",
     "STEPFUNCTIONS_LAMBDA_ENDPOINT",
@@ -1201,6 +1197,7 @@ CONFIG_ENV_VARS = [
     "LAMBDA_FORWARD_URL",  # deprecated since 2.0.0
     "LAMBDA_REMOTE_DOCKER",  # deprecated since 2.0.0
     "LAMBDA_STAY_OPEN_MODE",  # deprecated since 2.0.0
+    "SQS_PORT_EXTERNAL",  # deprecated in docs since 2022-07-13
     "SYNCHRONOUS_KINESIS_EVENTS",  # deprecated since 1.3.0
     "SYNCHRONOUS_SNS_EVENTS",  # deprecated since 1.3.0
     "SYNCHRONOUS_DYNAMODB_EVENTS",  # deprecated since 1.3.0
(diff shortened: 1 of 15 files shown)


localstack/deprecations.py
```diff
@@ -248,6 +248,11 @@ def is_affected(self) -> bool:
         "1.4.0",
         "This option has no effect anymore. Please use the AWS client and init hooks instead.",
     ),
+    EnvVarDeprecation(
+        "SQS_PORT_EXTERNAL",
```
joe4dev, line 252: better late than never 😬

tests/aws/services/sqs/test_sqs.py
```diff
@@ -1085,19 +1084,18 @@ def test_external_hostname_via_host_header(self, monkeypatch, sqs_create_queue):
         queue_name = f"queue-{short_uid()}"
         sqs_create_queue(QueueName=queue_name)
 
-        edge_url = config.get_edge_url()
```
joe4dev, line 1088: ⚠️ Are we correct assuming this is an external URL here?
External `localhost.localstack.cloud` vs. internal `localhost`

localstack/config.py
```diff
@@ -1268,43 +1265,63 @@ def populate_config_env_var_names():
...
+# TODO: Go over all usages and decide whether it's an internal or external usage
 def get_edge_url(localstack_hostname=None, protocol=None):
-    port = get_edge_port_http()
-    protocol = protocol or get_protocol()
-    localstack_hostname = localstack_hostname or LOCALSTACK_HOST.host
-    return "%s://%s:%s" % (protocol, localstack_hostname, port)
+    """@deprecated: Use `internal_service_url()` instead.
+    We assume that most usages are internal but really need to check and update each usage accordingly.
+    """
+    return internal_service_url(host=localstack_hostname, protocol=protocol)
```
joe4dev, line 1321: I have the impression that `external_service_url` might be more suitable here (we had 2 few failing tests indicating usages of external_service_url) but it might be mixed as well 🤷 
* `tests/aws/test_network_configuration.py::[1mTestAppsync::test_default_strategy`
* `tests/aws/services/apigateway/test_http_apis.py::[1mTestHttpApis::test_apigw_v2_http_jwt_authorizer`

localstack/services/sqs/models.py
simonrw, line 1: minor: the docstring for `SqsQueue.url` still mentions `SQS_PORT_EXTERNAL`, and in fact fetching the URL from the host header, which we also removed.

Should `scheme` come from `get_protocol()` rather than the request scheme? I'm not sure about this point.

tests/aws/services/sqs/test_sqs.py
simonrw: I think so, since it is asserting what the create queue call returns (which is returned to the user), and is in fact the same as the implementation in `sqs/models.py`.

tests/aws/test_network_configuration.py
```diff
@@ -166,7 +166,11 @@ def test_off_strategy_with_external_port(
     ):
         external_port = 12345
         monkeypatch.setattr(config, "SQS_ENDPOINT_STRATEGY", "off")
-        monkeypatch.setattr(config, "SQS_PORT_EXTERNAL", external_port)
+        monkeypatch.setattr(
+            config,
+            "LOCALSTACK_HOST",
+            config.HostAndPort(host=localstack_host().host, port=external_port),
+        )
```
simonrw, line 173: I think using a custom port as well as custom hostname would make a good addition to the `assert_host_customisation` fixture, though not in this PR.

simonrw (approved):
Thanks for catching these somewhat widely used helper functions in `config.py`.

joe4dev:
Thank you for the valuable comments @simonrw 
Merging this as discussed and then follow up the comments because we need to go over the usages anyways.

PR #9585, opened 2023-11-08
[minor] skip mounting binary pip dependencies in dev container

whummer:
## Motivation

This PR skips mounting a few pip dependencies into the dev container, to improve compatibility for non-Linux environments (running on Macbook Pro M1 here).

Discovered this recently when running some integration tests that are spawning external dev containers, e.g.,  `tests/pod_remotes/test_remotes_save_load.py::TestSmokePodPlatform`

```
...
ls-dev-29ad46fe:   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/cryptography/hazmat/primitives/ciphers/base.py", line 10, in <module>
ls-dev-29ad46fe:     from cryptography.exceptions import (
ls-dev-29ad46fe:   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/cryptography/exceptions.py", line 9, in <module>
ls-dev-29ad46fe:     from cryptography.hazmat.bindings._rust import exceptions as rust_exceptions
ls-dev-29ad46fe: ImportError: /opt/code/localstack/.venv/lib/python3.11/site-packages/cryptography/hazmat/bindings/_rust.abi3.so: invalid ELF header
```

## Changes

Add list of `skipped_dependencies` and skip mounting those into the dev container for testing.

### DIFF
--- a/localstack/dev/run/configurators.py
+++ b/localstack/dev/run/configurators.py
@@ -274,6 +274,9 @@ class DependencyMountConfigurator:
 
     dependency_glob = "/opt/code/localstack/.venv/lib/python3.*/site-packages/*"
 
+    # skip mounting dependencies with incompatible binaries (e.g., on MacOS)
+    skipped_dependencies = ["cryptography", "psutil", "rpds"]
+
     def __init__(
         self,
         *,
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
I guess the current approach is a bit problematic for all dependencies which rely on certain libs on the system, but this exclusion seems to be a fair compromise for now. 💯

PR #9590, opened 2023-11-09
remove duplicate on_extension_load call

thrau:
## Motivation

I noticed that extensions' `on_extension_load` methods are called twice. This is because the extension manager in localstack_ext.extension.manager already calls `on_extension_load` via the `ExtensionLifecycleListener` (see ext code [here](https://github.com/localstack/localstack-ext/blob/master/localstack_ext/extensions/manager.py#L24-L28)).

## Changes

* Remove the call to `self.on_extension_load()` in the `load` method of the extension.


alexrashed (approved):
Nice find! 💯

PR #9592, opened 2023-11-09
Add internal endpoint to return config

webdev51:
## Motivation
We needed config values in our web app and currently there's no endpoint which returns the config values so I added one.

## Changes
Added an endpoint in  `localstack/services/internal.py` file


webdev51:

dominikschubert (approved):

PR #9595, opened 2023-11-10
apigateway-kinesis: add region and account id params

sannya-singal:
## Motivation
The purpose of this PR is to enhance the stability of apigateway's integration with kinesis when dealing with multi-accounts and multi-regions. The main issue addressed is that when connecting api gateway to kinesis, the `apigateway` client was being instantiated without an `aws_access_key_id`. This omission caused the `kinesis` to be associated with an incorrect account.

## Changes
This PR introduces the use of `region` and `account_id` as mandatory fields in order to connect with the correct `apigateway` client `connect_to` method. The remaining modifications involve propagating the values to the relevant functions.

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_basic.py
+++ b/tests/aws/services/apigateway/test_apigateway_basic.py
@@ -13,8 +13,8 @@ from jsonpatch import apply_patch
 from requests.structures import CaseInsensitiveDict
 
 from localstack import config
-from localstack.aws.connect import connect_to
 from localstack.aws.api.lambda_ import Runtime
+from localstack.aws.connect import connect_to
 from localstack.aws.handlers import cors
 from localstack.config import get_edge_url
 from localstack.constants import (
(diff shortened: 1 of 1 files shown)


bentsku (changes requested):
I don't think we should ever call `connect_to` in our tests? Isn't it what our fixtures are for? I think we could use the `aws_client_factory` instead.

sannya-singal:
> edit: I see the test is using `resource_util`, maybe that's where the `connect_to` call comes from, as it's what used in it? Still, I believe in tests we should use our fixture, would make the migration for the test to be validated easier.

The `connect_to` call is being used by a helper function `connect_api_gateway_to_kinesis` which is called by the `test_api_gateway_kinesis_integration`. I see similar helper functions eg: `connect_api_gateway_to_sqs` using  `connect_to` call for client, @bentsku 🤔.

bentsku:
I understand, but I still don't think we should explicitly call `connect_to` in a test? These used in the `helpers` should be only used for internal calls, not during test I would say. We will need to refactor the `connect_api_gateway_to_sqs` as well to be able to validate them against AWS. None of those SQS are validated either. Is there a specific reason to use `connect_to`, doesn't the fixture work?

viren-nadkarni:
@bentsku is correct, `connect_to()` must not be used in tests. For exceptional uses, we have `connect_externally_to()` which may be used in tests.

@sannya-singal: The proper way will be to pass the `aws_client.apigateway` client instance to `connect_api_gateway_to_kinesis()`

sannya-singal:
Thanks for pointing this out @bentsku. I have made the required changes using `aws_client_factory` for apigateway client for a specific region and account. Please let me know if there is anything else. 

Also will refactor `connect_api_gateway_to_sqs` in the later PR's. 👍

bentsku (approved):
Awesome, thanks for fixing it! 🚀

sannya-singal:
LocalStack-ext test suite run: https://github.com/localstack/localstack-ext/actions/runs/6826365794

PR #9600, opened 2023-11-10
Remove deprecated networking config helpers

simonrw:
## Motivation

With the introduction of `GATEWAY_LISTEN` and `LOCALSTACK_HOST`, we introduced the split between cosmetic and functional configuration variables. Cosmetic variables were not used internally to define internal routes for e.g. health checks or service-to-service calls. They were only used when returning URLs outside of LocalStack. The motivation was that some users may wish to customise the domain name of LocalStack, for example when using the docker container name to resolve to the LocalStack container.

Previous to #9584 the helper functions in `localstack.config` used a mix of "cosmetic" and "functional" variables. For example, `get_edge_url` took the host from `LOCALSTACK_HOST` but the port from `GATEWAY_LISTEN`. This mixing of variable types led to strange situations. #9584 introduced a new API which was much clearer to use, and this PR updates all internal call sites of the old API to use the new API.

‼️ While migrating the functions, I found that OpenSearch adds a route handler for the specific URL that it returns, i.e. if it returns `mydomain.us-east-1.opensearch.foo.bar`, it only matches requests with that header. This makes testing annoying since it would be good to make a request to `mydomain.us-east-1.opensearch.localhost.localstack.cloud` (i.e. `127.0.0.1`) instead. I have had to work around this by setting the `Host` header manually, which in fact is the same thing done in the [`EdgeProxiedOpensearchCluster`](https://github.com/localstack/localstack/blob/80a6e272b1fd26be515e4bca991c9edbf28b6268/localstack/services/opensearch/cluster.py#L602-L609).

https://github.com/localstack/localstack/blob/80a6e272b1fd26be515e4bca991c9edbf28b6268/tests/bootstrap/test_cosmetic_configuration.py#L253-L258

It might be nice to accept any domain suffix. This might be good to do in a follow up PR as it does not remove functionality.

## Changes

* Remove all usages of `get_service_url`, `service_port` and `get_edge_url` from the LocalStack codebase
* Add a scenario test covering internal communication between services that have their URLs customised
* Add more obvious deprecation warnings for deprecated functions
* Address comments in #9584

### DIFF
--- a/localstack/testing/pytest/container.py
+++ b/localstack/testing/pytest/container.py
@@ -165,7 +165,7 @@ def container_factory() -> Generator[ContainerFactory, None, None]:
     factory.remove_all_containers()
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def wait_for_localstack_ready():
     def _wait_for(container: RunningContainer, timeout: Optional[float] = None):
         container.wait_until_ready(timeout)
--- a/localstack/utils/aws/aws_stack.py
+++ b/localstack/utils/aws/aws_stack.py
@@ -164,14 +164,10 @@ def get_boto3_region() -> str:
 
 def get_local_service_url(service_name_or_port: Union[str, int]) -> str:
     """Return the local service URL for the given service name or port."""
+    # TODO(srw): we don't need to differentiate on service name any more, so remove the argument
     if isinstance(service_name_or_port, int):
         return f"{config.get_protocol()}://{LOCALHOST}:{service_name_or_port}"
-    service_name = service_name_or_port
-    if service_name == "s3api":
-        service_name = "s3"
-    elif service_name == "runtime.sagemaker":
-        service_name = "sagemaker-runtime"
-    return config.service_url(service_name)
+    return config.internal_service_url()
 
 
 def get_s3_hostname():
--- a/localstack/utils/kinesis/kinesis_connector.py
+++ b/localstack/utils/kinesis/kinesis_connector.py
@@ -376,8 +376,8 @@ def start_kcl_client_process(
     kwargs = {"metricsLevel": "NONE", "initialPositionInStream": "LATEST"}
     # set parameters for local connection
     if aws_stack.is_local_env(env):
-        kwargs["kinesisEndpoint"] = config.get_edge_url(protocol="http")
-        kwargs["dynamoDBEndpoint"] = config.get_edge_url(protocol="http")
+        kwargs["kinesisEndpoint"] = config.internal_service_url(protocol="http")
+        kwargs["dynamoDBEndpoint"] = config.internal_service_url(protocol="http")
         kwargs["disableCertChecking"] = "true"
     kwargs.update(configs)
     # create config file
--- a/localstack/utils/kinesis/kinesis_connector.py
+++ b/localstack/utils/kinesis/kinesis_connector.py
@@ -23,6 +23,7 @@ from localstack.utils.strings import short_uid, to_str
 from localstack.utils.sync import retry
 from localstack.utils.threads import TMP_THREADS
 from localstack.utils.time import now
+from localstack.utils.urls import localstack_host
 
 EVENTS_FILE_PATTERN = os.path.join(tempfile.gettempdir(), "kclipy.*.fifo")
 LOG_FILE_PATTERN = os.path.join(tempfile.gettempdir(), "kclipy.*.log")
@@ -305,7 +306,7 @@ def get_stream_info(
     if aws_stack.is_local_env(env):
         stream_info["conn_kwargs"] = {
             "host": LOCALHOST,
-            "port": config.service_port("kinesis"),
+            "port": localstack_host().port,
             "is_secure": bool(config.USE_SSL),
         }
     if endpoint_url:
--- a/localstack/utils/kinesis/kinesis_connector.py
+++ b/localstack/utils/kinesis/kinesis_connector.py
@@ -23,7 +23,6 @@ from localstack.utils.strings import short_uid, to_str
 from localstack.utils.sync import retry
 from localstack.utils.threads import TMP_THREADS
 from localstack.utils.time import now
-from localstack.utils.urls import localstack_host
 
 EVENTS_FILE_PATTERN = os.path.join(tempfile.gettempdir(), "kclipy.*.fifo")
 LOG_FILE_PATTERN = os.path.join(tempfile.gettempdir(), "kclipy.*.log")
@@ -306,7 +305,7 @@ def get_stream_info(
     if aws_stack.is_local_env(env):
         stream_info["conn_kwargs"] = {
             "host": LOCALHOST,
-            "port": localstack_host().port,
+            "port": config.GATEWAY_LISTEN[0].port,
             "is_secure": bool(config.USE_SSL),
         }
     if endpoint_url:
--- a/tests/bootstrap/cdk_templates/LocalStackHostBootstrap/ClusterStack.json
+++ b/tests/bootstrap/cdk_templates/LocalStackHostBootstrap/ClusterStack.json
@@ -18,7 +18,7 @@
           "EnforceHTTPS": false,
           "TLSSecurityPolicy": "Policy-Min-TLS-1-0-2019-07"
         },
-        "DomainName": "domain-65913c6b",
+        "DomainName": "domain-cf686810",
         "EBSOptions": {
           "EBSEnabled": true,
           "VolumeSize": 10,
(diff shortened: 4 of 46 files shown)


tests/bootstrap/cdk_templates/LocalStackHostBootstrap/ClusterStack.json
```diff
@@ -0,0 +1,556 @@
...
+          "DedicatedMasterEnabled": false,
+          "InstanceCount": 1,
+          "InstanceType": "r5.large.search",
+          "ZoneAwarenessEnabled": false
+        },
+        "DomainEndpointOptions": {
+          "EnforceHTTPS": false,
+          "TLSSecurityPolicy": "Policy-Min-TLS-1-0-2019-07"
+        },
+        "DomainName": "domain-cf686810",
```
joe4dev, line 21: I guess we don't need to transform these because the stack serves as one generated output 🤔 (just double-check)

tests/bootstrap/resources/event_handler.py
```diff
@@ -0,0 +1,44 @@
...
+            print(f"Error fetching cluster health status: {e!r}")
+
+        assert custom_localstack_hostname in body["UnsubscribeURL"]
+
+        # write the result to s3
+        client.put_object(
+            Bucket=results_bucket, Key=results_key, Body=message["message"].encode("utf8")
+        )
+
+        # just take the first record for now
```
joe4dev, line 43: unrelated comment?

tests/bootstrap/test_cosmetic_configuration.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+@pytest.fixture(scope="class")
+def chosen_localstack_host() -> str:
+    """
+    Choose a domain name that is guaranteed never to resolve, except by the LocalStack DNS server
+
+    https://www.rfc-editor.org/rfc/rfc6761.html#section-6.4
+    """
+    return "foo.invalid"
```
joe4dev, line 44: `.invalid` 😮  Love the comment 👍

tests/integration/aws/test_app.py
```diff
@@ -175,7 +175,10 @@ def _echo_websocket_handler(request: WebSocketRequest, param: str):
         rule = ROUTER.add(_echo_websocket_handler)
         cleanups.append(lambda: ROUTER.remove(rule))
 
-        url = get_edge_url("localhost.localstack.cloud", protocol="wss") + "/_ws/world"
+        url = (
+            config.internal_service_url(host="localhost.localstack.cloud", protocol="wss")
```
joe4dev, line 179: nice non-http(s) example 👍

tests/unit/cli/test_cli.py
```diff
@@ -111,8 +111,15 @@ def mock_call(*args, **kwargs):
...
 def test_status_services(runner, httpserver, monkeypatch):
+    # configure LOCALSTACK_HOST because the services endpoint makes a request against the
+    # external URL of LocalStack, which may be different to the edge port
     monkeypatch.setattr(
-        config, "GATEWAY_LISTEN", [HostAndPort(host="0.0.0.0", port=httpserver.port)]
+        config,
+        "LOCALSTACK_HOST",
+        HostAndPort(
+            host="localhost.localstack.cloud",
+            port=httpserver.port,
```
joe4dev, line 121: nice catch 👍

localstack/testing/pytest/container.py
```diff
@@ -165,7 +165,7 @@ def container_factory() -> Generator[ContainerFactory, None, None]:
     factory.remove_all_containers()
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
dfangl, line 168: good catch!

localstack/utils/aws/aws_stack.py
```diff
@@ -179,14 +179,10 @@ def is_internal_call_context(headers) -> bool:
 
 def get_local_service_url(service_name_or_port: Union[str, int]) -> str:
     """Return the local service URL for the given service name or port."""
+    # TODO(srw): we don't need to differentiate on service name any more, so remove the argument
```
dfangl, line 182: Is this something that will be done in a follow up or should this be part of this PR?

localstack/utils/kinesis/kinesis_connector.py
```diff
@@ -305,7 +306,7 @@ def get_stream_info(
     if aws_stack.is_local_env(env):
         stream_info["conn_kwargs"] = {
             "host": LOCALHOST,
-            "port": config.service_port("kinesis"),
+            "port": localstack_host().port,
```
dfangl, line 309: Since this happens locally, shouldn't we use the gateway listen port here?

joe4dev (approved):
LGTM 👍 Awesome work churning over so many usages 👏👏👏 Looking forward to more consistent networking configurations across LocalStack 🚀 

nothing blocking, mainly minor polishing comments

joe4dev:
💡  It is pretty hard to cover so many usages with meaningful tests. In the future, we could consider setting up a periodic or on-demand CI scenario that runs LocalStack behind a proxy and then we can simply run the entire test suite with a customized `LOCALSTACK_HOST` configuration.

tests/bootstrap/test_cosmetic_configuration.py
```diff
@@ -0,0 +1,297 @@
...
+
+        with infra.provisioner() as prov:
+            yield prov
+
+    def test_scenario(self, port, infrastructure, aws_client_factory, chosen_localstack_host):
+        """
+        Scenario:
+            * API Gateway handles web request
+            * Broadcasts message onto SNS topic
+            * Lambda subscribes via SQS and inserts message into OpenSearch
```
dfangl, line 240: It does not do that, right? It just checks if it's ready and puts the message into an s3 bucket.

dfangl (approved):
I would in general argue that the tests should use `external_service_url` and not the internal one, as they simulate and external caller (and we should not set our tests to use an invalid LOCALSTACK_HOST for the most part). However, it's fine for me to leave it like that for now.

Good work regardless, only have some minor nits

tests/bootstrap/resources/event_handler.py
simonrw: No, the loop exits early with the `return` statement below it. The number of records should always be 1 but this is here just in case

tests/bootstrap/cdk_templates/LocalStackHostBootstrap/ClusterStack.json
simonrw: Yeah sort of. This is just a cache that prevents needless `cdk synth`. The domain name can be fixed for this.

localstack/utils/aws/aws_stack.py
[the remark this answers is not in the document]
simonrw: I'm ok with leaving this for now as it's an internal refactoring. As you say we should remove the `aws_stack` stuff in the future anyway.

tests/bootstrap/test_cosmetic_configuration.py
simonrw: Absolutely correct, it was planned, but was more complex than I really wanted to get into.

localstack/utils/kinesis/kinesis_connector.py
simonrw: Yeah that explains the hard coding of `LOCALHOST` as well :+1:

localstack/utils/aws/aws_stack.py
simonrw: Follow up

localstack/utils/kinesis/kinesis_connector.py
joe4dev: There are lots of hardcoded `LOCALHOST` and `LOCALHOST_HOSTNAME` examples for returning URLs that might need a follow-up as well 😬 ...

simonrw: Do you have any examples? I don't immediately see anything that I missed

joe4dev: For example this one mixing `localhost.localstack.cloud` (external) with `EDGE_PORT` (internal):
https://github.com/localstack/localstack/blob/ccb95521b867a35576e9ad3d61e620fbd761f015/localstack/utils/testutil.py#L689

The API gateway tests have similar occurrences, for example here:
https://github.com/localstack/localstack/blob/ccb95521b867a35576e9ad3d61e620fbd761f015/tests/aws/services/apigateway/test_apigateway_basic.py#L1223


I rebased the edge port migration PRs upon these PRs fixing these issues and other `EDGE_PORT` occurrences I spotted while reviewing this PR. We should not delay this any further.

joe4dev: https://github.com/localstack/localstack/pull/9633

PR #9603, opened 2023-11-10
fix S3 parser with empty integer query string parameters

bentsku:
## Motivation
When scrolling over my GitHub homepage, I stumbled upon this (edit: apparently dead) library: https://github.com/beyondstorage/go-storage

(Story: https://xuanwo.io/en-us/2023/01-beyond-storage-why-we-failed/, all commits are from dependabot now 😅 and it seems [OpenDAL](https://opendal.apache.org) ([repo](https://github.com/apache/incubator-opendal)) is superseding the project, I've ran their Rust test suite (106 tests!) and it's all green 🎉)

It seems it's has been sponsored by Vercel, and I saw they had coverage for S3, and an automated in CI integration test suite.
I've forked the repo and ran the suite locally with LocalStack 😄 I saw they had support for minio too.

I've ran it locally and found only one test failure:
```go
  When ListMultiPart 
      ListMultipart error should be nil ✔✔
      Next error should be nil ✘
      The part number and size should be match 🔥
```
With the following trace:
```python
    File "/Users/benjaminsimon/Projects/localstack/localstack/localstack/aws/protocol/parser.py", line 266, in _parse_shape
      raise ProtocolParserError(
  localstack.aws.protocol.parser.ProtocolParserError: Invalid type when parsing PartNumberMarker: '' cannot be parsed to integer.
```

I've checked the request we received:
```
GET s3.localhost.localstack.cloud:4566/test/3452e591-f8a8-4bc6-ac1c-7f98e4a045c9/e2e0d2d4-ecc1-40f1-a721-0545dee865c8?max-parts=200&part-number-marker=&uploadId=lEfzr1V8eelmpBbNK_jZfPZps2I3fQjspBA_WzlIISL8_6nY8N026B16MYT0kRB7r3G4nxTWCraDKtbHE6vM5ZXCbzFx6D-wnikrY96R3fHebKq662uvnukbdcmf4JXx&x-id=ListParts
# note the ?max-parts=200&part-number-marker=&...`
```

So it seems S3 accepts empty query string parameters that are of `integer` type in the specs.

After fixing the issue, here's the test result (87 successful tests!):
<details>
<summary>Test Results</summary>

`$ make integration_test`

```go
go test -count=1 -race -covermode=atomic -v ./tests
=== RUN   TestStorage
    utils_test.go:15: Setup test for s3
=== RUN   TestStorage/TestDelete
=== RUN   TestStorage/TestDelete/TestDelete
=== RUN   TestStorage/TestList
[...]

PR #9604, opened 2023-11-11
fix S3 v3 VirtualHost removing trailing slash

bentsku:
## Motivation
Following #9603, I've ran the OpenDAL integration test with `OPENDAL_S3_ENABLE_VIRTUAL_HOST_STYLE=on` to see if there would be any difference, and 2 tests failed. After investigation, it was because calling `ListObjectsV2` would return the keys without the trailing slash, making the test assertions fail. After investigating the difference between the calls, I could see the parsed request was different between the 2 modes:
Virtual-Host before the fix:
`PUT opendal-testing.s3.localhost.localstack.cloud:4566/CI/16b716e8-4cd2-4f6e-9d82-ee338b5e48b1/580cc505-82cf-424e-8448-ca4bf9dfd1ac/`
`-> PutObjectRequest({'ACL': None, 'Bucket': 'opendal-testing', ..., 'Key': 'CI/16b716e8-4cd2-4f6e-9d82-ee338b5e48b1/580cc505-82cf-424e-8448-ca4bf9dfd1ac', 'Metadata': {}...) `

Path style:
`PUT s3.localhost.localstack.cloud:4566/opendal-testing/CI/cc66a8ac-a924-42b2-950a-b5a9b6bca060/23a5e248-3249-436f-881f-cc1b8dcbffbe/`
`-> PutObjectRequest({'ACL': None, 'Bucket': 'opendal-testing', ..., 'Key': 'CI/cc66a8ac-a924-42b2-950a-b5a9b6bca060/23a5e248-3249-436f-881f-cc1b8dcbffbe/', 'Metadata': {}...)`

As we can see, the Virtual Host mangled the trailing slash, thus creating an object with the wrong key. This was due to a previous assertion in the VirtualHostRewriter which was written to actually mangle it. I've changed it to be closer to our previous proxy, and simply prefix the path with the bucket. 

Those test suites are really building confidence in our new S3 implementation, most of them are passing on the first try and some real edge cases are popping up. I'll continue to scour the internet to find S3 test suites to test LocalStack against. Maybe another Apache project which will use LocalStack 😄

Just for info, here are the 2 previously failing tests:
- [test_list_nested_dir](https://github.com/apache/incubator-opendal/blob/c5410f25322ab9daf8dc270bfed6cb70511a2463/core/tests/behavior/list.rs#L263C26-L263C26)
- [test_list_sub_dir](https://github.com/apache/incubator-opendal/blob/c5410f25322ab9daf8dc270bfed6cb70511a2463/core/tests/behavior/list.rs#L240) 

## Changes
- Updated an existing test verifying this behavior for the regular path style (we already have a special case for this in the parser, see `localstack.aws.protocol.parser.S3RequestParser._parse_shape`) to also test for virtual host mode.
- Modified the VirtualHostRewriter to not mangle the trailing slash
- Also spotted a different issue in the V2 legacy provider, where the v-host proxy was not proxying the request with the raw path, so the test would fail.

### DIFF
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -991,18 +991,12 @@ class S3RequestParser(RestXMLRequestParser):
                 # remove the bucket name from the host part of the request
                 new_host = self.old_host.removeprefix(f"{bucket_name}.")
 
-                # split the url and put the bucket name at the front
-                path_parts = self.old_path.split("/")
-                path_parts = [bucket_name] + path_parts
-                path_parts = [part for part in path_parts if part]
-                new_path = "/" + "/".join(path_parts) or "/"
+                # put the bucket name at the front
+                new_path = "/" + bucket_name + self.old_path or "/"
 
                 # create a new RAW_URI for the WSGI environment, this is necessary because of our `get_raw_path` utility
                 if self.old_raw_uri:
-                    path_parts = self.old_raw_uri.split("/")
-                    path_parts = [bucket_name] + path_parts
-                    path_parts = [part for part in path_parts if part]
-                    new_raw_uri = "/" + "/".join(path_parts) or "/"
+                    new_raw_uri = "/" + bucket_name + self.old_raw_uri or "/"
                     if qs := self.request.query_string:
                         new_raw_uri += "?" + qs.decode("utf-8")
                 else:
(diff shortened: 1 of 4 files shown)


alexrashed (approved):
Nice and clean change! Thanks for jumping on this, and providing the additional snapshot validated test! 🚀 💯

Issue #9408, opened 2023-10-19
enhancement request: OpenSearch 2.9 support


theahura:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

OpenSearch 2.9 supports a lot of useful features that AWS opensearch also supports, such as being able to use filters in on FAISS searches in the KNN plugin. It would be great to have these in localstack!

### 🧑‍💻 Implementation

this may be as trivial as a version bump? As far as I can tell, localstack is just pulling the opensearch docker image

theahura:
hey @ackdav thanks for accepting this feature request. Curious what the timeline is for looking into this!

alexrashed:
Hi @theahura!
Thanks for your initial contribution and @viren-nadkarni for bringing it over the line! I just merged the PR, and it will be available in `localstack/localstack` with the next build on `master`, and in `localstack/localstack-pro` in the next few hours.
Thanks again!

theahura:
Fantastic. Any idea when the next build on master would be?

PR #9606, opened 2023-11-11
Bump supported versions in opensearch to 2.9

theahura:
## Motivation
Resolves https://github.com/localstack/localstack/issues/9408

As far as I can tell, localstack pulls the latest OpenSearch docker images and more or less lets the cluster run as is. 2.9 has significant improvements compared to 2.5 -- most relevant to me, being able to run OpenSearch KNN queries with metadata filters. As far as I can tell, 2.9 is also backwards compatible with 2.5, so I figured a very straightforward version bump would be sufficient to get it running as a first pass.

## Changes
Just adds OpenSearch 2.9 as an accepted version.

I am basing this PR off the following:
https://github.com/localstack/localstack/commit/5f591c7b5569fc8bbcf31c379bda90013bcb4fc7 (which adds OS 2.5)
https://github.com/localstack/localstack/commit/e91ea756a7ee34cc4eb46d6c5d7c6105ea226924 (which adds OS 2.7)

Pulled what seemed like the relevant pieces in to bump version, but may have missed a test or an integration point. The default OpenSearch version is still set to 2.5.

## Testing

Tested this locally, `aws opensearch create-domain --domain-name example --engine-version OpenSearch_2.9 --endpoint-url http://localhost:4566` works as expected, as does `describe-domain` on the same endpoint.

## TODO

What's left to do:

- [ ] Update tests if there are any relevant tests
- [ ] Add other compatible versions in the same file
- [ ] ???

### DIFF
--- a/localstack/services/opensearch/versions.py
+++ b/localstack/services/opensearch/versions.py
@@ -15,6 +15,7 @@ from localstack.utils.common import get_arch
 
 # Internal representation of the OpenSearch versions (without the "OpenSearch_" prefix)
 _opensearch_install_versions = {
+    "2.9": "2.9.0",
     "2.7": "2.7.0",
     "2.5": "2.5.0",
     "2.3": "2.3.0",
--- a/localstack/services/opensearch/versions.py
+++ b/localstack/services/opensearch/versions.py
@@ -221,6 +221,7 @@ compatible_versions = [
         TargetVersions=["OpenSearch_2.5"],
     ),
     CompatibleVersionsMap(SourceVersion="OpenSearch_2.5", TargetVersions=["OpenSearch_2.7"]),
+    CompatibleVersionsMap(SourceVersion="OpenSearch_2.7", TargetVersions=["OpenSearch_2.9"]),
 ]
 
 
--- a/tests/aws/services/es/test_es.py
+++ b/tests/aws/services/es/test_es.py
@@ -14,7 +14,10 @@ from localstack.utils.common import short_uid, start_worker_thread
 LOG = logging.getLogger(__name__)
 
 # Common headers used when sending requests to OpenSearch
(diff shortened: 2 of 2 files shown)


theahura:

localstack/services/opensearch/versions.py
```diff
@@ -220,6 +221,7 @@
         TargetVersions=["OpenSearch_2.5"],
     ),
     CompatibleVersionsMap(SourceVersion="OpenSearch_2.5", TargetVersions=["OpenSearch_2.7"]),
+    CompatibleVersionsMap(SourceVersion="OpenSearch_2.7", TargetVersions=["OpenSearch_2.9"]),
```
viren-nadkarni, line 224: This is incomplete, you will also need to update the `OpenSearch_2.X` version maps since that offers a valid upgrade path to 2.9.

Refer to AWS's `GetCompatibleVersions` operation for full map. You will also need to update the `test_get_compatible_versions` test.

viren-nadkarni (changes requested):
Thanks for the PR @theahura!

We want to make LocalStack use the latest engine version like AWS. You will need to bump the `OPENSEARCH_DEFAULT_VERSION` var in `localstack.constants`.

As per the [AWS migration guide](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/version-migration.html) there are no breaking changes with this bump so we're good on that front.

You can ignore the **Community Integration Tests against Pro** CI jobs as they don't work with contributor branches. Greenifying the **docker-test-XXX** jobs are good enough.

viren-nadkarni:
After conferring with my colleage @alexrashed, we're taking over this PR. LocalStack v3.0 release is imminent and we think this would be a great change to include this in.

See https://github.com/localstack/localstack/pull/9626

PR #9607, opened 2023-11-11
fix SQS JSON leftovers

bentsku:
## Motivation
To quickly merge the changes, the PR comments of #8268 had not been addressed yet. This PR does the second iteration.

## Changes
- remove a debug print statement and fix a typo
- add a comment regarding the conflict resolution between `sqs` and `sqs-query` service names, might need to create a specific step in the service name parser resolution if more services start to migrate to the `json` protocol, based on `Content-Type` header?
- add `sqs-query` as a dependency of SQS in the API dependencies for `SERVICES`, so that we enable both in case `sqs` is enabled. 

## TODO

What's left to do:

- [ ] keep an eye on https://github.com/boto/botocore/commit/50861b96f9d51632d6ddd8bafac8d50f4e8b027a, which will be in release  >1.31.84 (not yet released then), which explains this change in the previous PR:
  ```diff
  -        with pytest.raises(aws_client.sqs.exceptions.QueueNameExists):
  +        with pytest.raises(ClientError) as e:
  ```
  (I don't think it would change anything in snapshot but we can verify then)

### DIFF
--- a/localstack/aws/protocol/serializer.py
+++ b/localstack/aws/protocol/serializer.py
@@ -1633,7 +1633,7 @@ class SqsResponseSerializer(JSONResponseSerializer):
     ) -> None:
         """
         Overrides _serialize_error as SQS has a special header for query API legacy reason: 'x-amzn-query-error',
-        which contatained the exception code as well as a Sender field.
+        which contained the exception code as well as a Sender field.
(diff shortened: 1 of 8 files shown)


alexrashed (approved):
🧹

Issue #9605, opened 2023-11-11
bug: listing objects with delimiter, max-keys and continuation-token returns same CommonPrefixes multiple times


foodprocessor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

These objects are put in the bucket:
```
folder/aSubfolder/subFile1
folder/aSubfolder/subFile2
folder/file1
folder/file2
```
And then we call [ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) with prefix=`folder`, delimiter="/" and max-keys=1.

ListObjectsV2 returns no Contents and `aSubfolder` in CommonPrefixes from multiple calls (twice, in this case) when listing the `folder` prefix.

### Expected Behavior

`aSubfolder` should only be returned by one call to ListObjectsV2. After that it should move past `aSubfolder` and return `file1`.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://test

Here's the Go program I used to repro this (~150 lines)
```
package main

import (
	"bytes"
	"context"
	"fmt"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/credentials"
	"github.com/aws/aws-sdk-go-v2/service/s3"
)

func main() {

	fmt.Println("Connecting...")
	awsS3Client, err := connect()
	if err != nil {
		fmt.Println("Failed to connect. Exiting...")
		return
	}

	fmt.Println("Connected. Writing test objects...")

	// put test objects
	err = putObject(awsS3Client, "test", "outer/inner/innerFile1")
	if err != nil {
		return
	}
	err = putObject(awsS3Client, "test", "outer/inner/innerFile2")
	if err != nil {
		return
	}
	err = putObject(awsS3Client, "test", "outer/outerFile")
	if err != nil {
		return
	}

	fmt.Println("Listing objects one by one...")

	err = listObjectsOneByOne(awsS3Client, "test", "outer/", 1)
	if err != nil {
		return
	}

	fmt.Println("Done.")
}

func connect() (*s3.Client, error) {
	// endpoint
[...]

PR #9608, opened 2023-11-12
fix S3 List* pagination CommonPrefixes

bentsku:
## Motivation
As reported in #9605, we were not handling properly the `Max*` argument when combined with `Delimiter` and `Prefix`, which returns `CommonPrefixes`. 

Here's the description of the `Delimiter` parameter, which helps understand the behavior:
> The delimiter grouping the included keys. A delimiter is a character that you specify to group keys. All keys that contain the same string between the prefix and the first occurrence of the delimiter are grouped under a single result element in CommonPrefixes. These groups are counted as one result against the max-keys limitation. These keys are not returned elsewhere in the response.

The behavior is most `List*` operations is pretty close, but has subtle changes between them. 

## Changes
- Split all tests related to `List*`, especially around pagination, as those are quite verbose and the `test_s3` file is becoming extremely huge (>10k lines) into its own file.
- Create new tests validating this behavior for every List* operation except `ListParts` which doesn't have the `CommonPrefix` behavior.  
  - [`ListObjects`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html)
  - [`ListObjectsV2`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html)
  - [`ListObjectVersions `](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html)
  - [`ListMultipartUploads `](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListMultipartUploads.html)

EDIT: it was green before rebasing, pipeline is struggling (kclpy): https://app.circleci.com/pipelines/github/localstack/localstack/20034/workflows/a3805d65-c1b9-4708-a79b-f5ceed202830

## Testing notes

I've also re-ran external test suites with this change and it's all green there as well.


macnev2013 (approved):
LGTM 👍🏻

PR #9609, opened 2023-11-12
fix S3 pre-signed logic for IGNORED_SIGV4_HEADERS

bentsku:
## Motivation
Following up on #9603, I've quickly investigated the issue for the pre-signed URL message showing a non valid signature for a very simple request and quickly found the issue related to signed headers being in the ignore list.

## Changes
- fix the logic when ignoring SigV4 headers, we should still include them if they're in the `SignedHeaders`, and only ignore if they're missing
- add a test validating this behavior and copying the Go SDK behavior with the help of registering handlers. This was extremely tricky to force boto to behave this way, as the AWS CRT library does not allow you to do so. The Go SDK seems a bit weird, but it allowed us to verify the behavior of previously thought "ignored" headers.

### DIFF
--- a/localstack/services/s3/presigned_url.py
+++ b/localstack/services/s3/presigned_url.py
@@ -565,11 +565,10 @@ class S3SigV4SignatureContext:
         not_signed_headers = []
         for header, value in headers.items():
             header_low = header.lower()
-            if header_low.startswith("x-amz-"):
-                if header_low not in signed_headers.lower():
-                    if header_low in IGNORED_SIGV4_HEADERS:
-                        continue
-                    not_signed_headers.append(header_low)
+            if header_low.startswith("x-amz-") and header_low not in signed_headers.lower():
+                if header_low in IGNORED_SIGV4_HEADERS:
+                    continue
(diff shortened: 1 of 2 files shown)


localstack/services/s3/presigned_url.py
```diff
@@ -64,20 +64,7 @@
...
-# query params overrides for multipart upload and node sdk
-# TODO: this will depends on query/post v2/v4. Manage independently
-ALLOWED_QUERY_PARAMS = [
-    "x-id",
-    "x-amz-user-agent",
-    "x-amz-content-sha256",
-    "versionid",
-    "uploadid",
-    "partnumber",
-]
```
bentsku, line 76: this was not used anywhere

steffyP (approved):
Great work, @bentsku! 🙇‍♀️ 🚀 

Love that the test is even aws validated ❤️ and very smart test setup with the meta-events 😄

This is already a very specific case for S3, the new provider is truely amazing 🚀 👏 

Just added a few NITs, but nothing blocking the merge!

Issue #9612, opened 2023-11-13
enhancement request: stable docker image


sha1n:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Our team uses the `localstack/localstask:latest` image for testing a large number of services. Over the past few weeks we had several issues caused by what appears to a bug in the latest image which caused CI failures around the dynamodb service in this case. I looked around for a different tag which is not a fixed version number with `stable` semantics and it doesn't seem to exist.

In order to help teams stay up to date without the need to constantly update image versions, it would be very helpful if you could provide a `stable` tag we can count on to be functional, or make sure the `latest` tag is stable.

### 🧑‍💻 Implementation

You could add a small step to the image release process, which adds a `stable` tag to the released image and pushes it, similar to `latest`.

alexrashed:
Hi there!
Thanks for the input and the contribution!
I am pretty confident that we can bring this over the line today, with `3.0` being the first release with a `stable` tag.

However, I just wanted to mention this here, since it's pretty related: With https://github.com/localstack/localstack/issues/9488 and https://github.com/localstack/localstack/pull/9490, we will also be pushing major version tags starting with `3.0`.
This tag can be useful if you want to explicitly and manually upgrade major versions.

sha1n:
Thank you very much @alexrashed! I appreciate the quick response 🚀

PR #9613, opened 2023-11-13
 Publish stable version on non-dev versioning

mormamn:
## Motivation
Motivation is mentioned here: #9612

## Changes
Added a new `stable` tag to Dockerhub publishing, this will allow users to be on the most latest code, without compromising stability - The new `stable` tag should only be published on non-dev code

## Testing
`make docker-push-master` with `.dev` in `localstack/__init__.py`, and then without.
Validate that when we have `.dev` version - `stable` tag is not pushed, only `latest` is pushed
When we do not have `.dev` version - along side with the `IMAGE_TAG`, the `stable` is also pushed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,9 @@ docker-create-push-manifests:	## Create and push manifests for a docker image (d
 			(docker manifest create $(MANIFEST_IMAGE_NAME):$(IMAGE_TAG) \
 			--amend $(MANIFEST_IMAGE_NAME):$(IMAGE_TAG)-amd64 \
 			--amend $(MANIFEST_IMAGE_NAME):$(IMAGE_TAG)-arm64 && \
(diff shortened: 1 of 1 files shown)


mormamn:

mormamn:
semver: patch

alexrashed:
Hi @mormamn!
Thanks a lot for your contribution! Since we are currently preparing a release (v3.0 will be released tomorrow), I just took this one over to bring this over the line before the release.
Your changes only created the platform specific tags, I pushed another commit which picks them up and creates the multi-arch manifest tags which you are most likely using.
From my perspective this is looking good, and I also tested it with a test-image in my personal account on DockerHub. Everything works just as expected now.
I'll wait for another pair of eyes, and then we can move forward with the merge. `3.0` will be the first tagged release with a `stable` tag! 💯

dfangl (approved):

PR #9343, opened 2023-10-13
Remove the deprecated internal boto client library

viren-nadkarni:
## Motivation

We introduced a new internal Boto client library `localstack.aws.connect_to`. Subsequently the codebase was refactored to use this new client library and the old library in `localstack.utils.aws.aws_stack` was marked obsolete.

This PR removes these deprecated methods as they are no longer used in either codebases.

Ext Integration Tests passing :heavy_check_mark: # 2735 

## Related

- https://github.com/localstack/localstack/pull/8636
- https://github.com/localstack/localstack/pull/8748
- https://github.com/localstack/localstack/pull/8858
- https://github.com/localstack/localstack/pull/9120

PR #9616, opened 2023-11-13
Remove legacy internal call detection helper

viren-nadkarni:
## Motivation

LocalStack now uses the `RequestContext.is_internal_call` attribute to indicate internal cross-service calls.

## Implementation

This PR removes the use of legacy helper which relies on `HEADER_LOCALSTACK_ACCOUNT_ID` header which itself is a vestige of the old `aws_stack.connect_to_service()` utility. 

The old client was removed in https://github.com/localstack/localstack/pull/9343, and this is not a breaking change.


alexrashed (approved):
🧹 🚀

PR #9619, opened 2023-11-13
Adding AWS:Amplify:*  Resources

Morijarti:
## Changes

Added following resources
- AWS:Amplify:App


dominikschubert:
Won't be necessary anymore with https://github.com/localstack/localstack/pull/10242

Util will be extracted into a more fitting PR

PR #9621, opened 2023-11-13
add services to API_DEPENDENCIES

bentsku:
## Motivation
Following service owners, we're adding more dependencies to the list of services.

## Changes
- add `cloudformation` to the list
- add `transcribe`
- remove `tests/integration/test_services_enabled.py` as we now have a way better bootstrap test which is more solid


steffyP (approved):
Thanks so much for updating, @bentsku! 🙏
LGTM!

PR #9624, opened 2023-11-14
Bump moto-ext to 4.2.9.post2

viren-nadkarni:
Brings in:
- https://github.com/getmoto/moto/pull/7009

cc: @dfangl 

Ext Integration Tests # 3632 :green_circle:


localstack/services/s3/notifications.py
```diff
@@ -136,7 +136,7 @@ def from_request_context(
             )
         except NoSuchKey as ex:
             if allow_non_existing_key:
-                key: FakeKey = FakeKey(key_name, "")
+                key: FakeKey = FakeKey(key_name, "", request_context.account_id)
```
bentsku, line 139: they finally fixed the key not having an account id in multipart parts, so that the S3Backend would be initialized with no account id, which led to an issue with persistence. Better late than never I guess!

bentsku (approved):
LGTM for the S3 part!

alexrashed (commented):
The changes are looking good to me! But please make sure to get a green ext it before merging, even if it was a flake (please rerun, or make sure this is fixed first).

alexrashed:
It would be great if we could get the fix for using new botocore versions into the next moto upgrade: https://github.com/getmoto/moto/pull/7030
This would allow us to unpin botocore again, which we had to do with 815d6c77245e3ebfb9d298a70fe16bd12df032f7 :)

steffyP:
FYI: [this change in moto](https://github.com/localstack/moto/commit/447710c6a68e7d5ea7ad6d7df93c663de32ac7f1) will break instance creation for docdb + neptune

localstack/services/s3/notifications.py
bblommers: > Better late than never I guess!

This was changed in Moto to improve the type annotations, @bentsku . It wasn't fixed as a bug, because no-one has ever raised an issue/PR to indicate there was a problem with it.

bentsku: Ah, sorry: my tone was not great re-reading myself.
This bug was mostly on our side with the persistence mechanism, the multipart `Part` object not having an account id was not really causing a bug in Moto itself, so we fixed our handling of it. It is true that I could have opened a bug report about it, but it felt very specific and not adding anything to Moto. Reading my comment, I can see how rude it sounds. My apologies.

bblommers: All good, no harm done. Appreciate the explanation!

PR #9627, opened 2023-11-14
refresh sqs snapshots and fix exceptions in serializer

bentsku:
## Motivation
Following botocore update regarding exception handling of SQS (introduced in https://github.com/boto/botocore/pull/3054, released with 1.31.85), we needed to refresh snapshots for the newly added field.
It put into light some changes regarding error codes, so I've adapted the exceptions raised and their mapping in the serializer. 
It also showed that the error message is different for `QueueDoesNotExist` depending on `query` or `json` protocol, put into light with a new snapshot test.

## Changes
- refresh all SQS snapshots
- modify the exception classes to match their new names
- add those new exceptions to the map for their old names (different between `Error` and `QueryErrorCode`)
- added some snapshot tests to validate behavior
- added a quick fix in the provider to have a different message depending on the service name. Might need to think of a better handling of this, but historically error message were the responsibility of providers, not serializer. 
- reverted the change in the CloudFormation models regarding using the boto client exceptions.
- edit: I forgot, but I also fixed the marker detection for `only_localstack` so that we don't run those tests against AWS. I believe one nice step also would be to not create a snapshot entry in the json file if the test does not have the snapshot fixture, I'm sure we could detect that? @dominikschubert 
- update the call to `connect_to` in the query API to use `sqs_query` client


alexrashed (approved):
Awesome! Thanks for tackling this! Hopefully this is the last iteration related to https://github.com/localstack/localstack/pull/8268 and https://github.com/localstack/localstack/issues/8267, and associated with the recent changes in `botocore`! 🦸🏽 🧹

bentsku:
-ext run: https://github.com/localstack/localstack-ext/actions/runs/6863466683

dominikschubert:
> I believe one nice step also would be to not create a snapshot entry in the json file if the test does not have the snapshot fixture, I'm sure we could detect that?

@bentsku 
Sure but it's a bit hacky and potentially confusing since technically those tests do actually use the snapshot fixture (because of the autouse). What you could do is split them into classes and only autoimport in the non-"only_localstack" classes. :thinking:

bentsku:
@dominikschubert good point, I didn't think about the auto-use, it makes a lot of sense! Or maybe something at the "save" time, if there are no keys and the dict is empty, we could maybe not create an entry in the file? That might be simpler. Let's see after v3, this is extremely minor 😄 thanks for the response!

PR #9630, opened 2023-11-14
Properly parse error message for container host port open detection

dfangl:
## Motivation
Currently we are getting unwarranted `Unexpected error when attempting to determine container port status:` errors on some systems, depending on the docker daemon API answer.

## Changes
* Add additional possible log message to suppress faulty warning


simonrw (approved):
I think there may be a few messages that we might get in the exception string, since different platforms print different messages, but for now I think this captures the common cases I've seen. Great stuff!

dominikschubert (approved):

PR #9634, opened 2023-11-14
fix SQS json requests sent to query route

bentsku:
## Motivation
As reported in https://github.com/localstack/localstack/issues/8267#issuecomment-1810805915, it seems the [PHP SDK](https://github.com/aws/aws-sdk-php) still struggles to communicate with LocalStack.

After investigation, it showed that the PHP SDK is still sending the requests to the queue URL even though the specs are specifying that it should be against `/`. Our router caught the request and assumed it was a Query API. 

## Changes
- if the `Content-Type` header is `application/x-amz-json-1.0`, raise `NotFound` in the route so that the handler chain can take of the request
- add a test testing this very weird behavior by creating a client with an endpoint set to the queue URL (tested against AWS, it weirdly works...)

Tested with a PHP sample, it solves the issue now

## Testing

To test the changes with the PHP SDK:
- `docker pull php`
- `docker run --rm -it php bash`
- `curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer`
- `apt update`
- `apt install zip`
- `cd /usr/src` & `mkdir app` & `cd app`
- `composer require aws/aws-sdk-php`
- set this in the file `test.php`: 
```php
<?php
// Require the Composer autoloader.
require 'vendor/autoload.php';

use Aws\Sqs\SqsClient;

$sqs = new SqsClient([
    'version' => 'latest',
    'region'  => 'us-west-2'
]);

$params = [
    'DelaySeconds' => 10,
    'MessageAttributes' => [
        "Title" => [
            'DataType' => "String",
            'StringValue' => "The Hitchhiker's Guide to the Galaxy"
        ]
    ],
    'MessageBody' => "Information about current NY Times fiction bestseller for week of 12/11/2016.",
    'QueueUrl' => 'http://host.docker.internal:4566/queue/us-east-1/000000000000/test'
];

$result = $sqs->sendMessage($params);
var_dump($result);
```
- `export AWS_ACCESS_KEY_ID=test`
- `export AWS_SECRET_ACCESS_KEY=test`
- `php test.php`
- run, it should now properly show the result and not an exception

### DIFF
--- a/localstack/services/sqs/query_api.py
+++ b/localstack/services/sqs/query_api.py
@@ -138,7 +138,7 @@ def handle_request(request: Request, region: str) -> Response:
     # some SDK (PHP) still send requests to the Queue URL even though the JSON spec does not allow it in the
     # documentation. If the request is `json`, raise `NotFound` so that we continue the handler chain and the provider
     # can handle the request
-    if request.headers.get("Content-Type").lower() == "application/x-amz-json-1.0":
+    if request.headers.get("Content-Type", "").lower() == "application/x-amz-json-1.0":
         raise NotFound
 
     request_id = long_uid()
--- a/localstack/services/sqs/query_api.py
+++ b/localstack/services/sqs/query_api.py
@@ -9,6 +9,7 @@ from urllib.parse import urlencode
 from botocore.exceptions import ClientError
 from botocore.model import OperationModel
 from werkzeug.datastructures import Headers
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Perfect, thanks so much for investigating on this weird behavior, fixing it, and writing a validated test for it! 🦸🏽

PR #9638, opened 2023-11-15
SQS: Rework is_sqs_queue_url() utility

viren-nadkarni:
## Motivation

The `is_sqs_queue_url()` utility relied on the `get_aws_account_id()` function, which is about to be deprecated and removed.

This PR reworks it to use dedicated regexes for each endpoint scheme.

### DIFF
--- a/localstack/services/sqs/constants.py
+++ b/localstack/services/sqs/constants.py
@@ -32,9 +32,9 @@ INTERNAL_QUEUE_ATTRIBUTES = [
 ]
 
 # URL regexes for various endpoint strategies
-STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
-DOMAIN_STRATEGY_URL_REGEX = r"((?P<region_name>[a-z0-9-]{1,}).)?queue.localhost.localstack.cloud:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
-PATH_STRATEGY_URL_REGEX = r"localhost:\d{4,5}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,})\.[^:]+:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+DOMAIN_STRATEGY_URL_REGEX = r"((?P<region_name>[a-z0-9-]{1,})\.)?queue\.[^:]+:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+PATH_STRATEGY_URL_REGEX = r"[^:]+:\d{4,5}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
 LEGACY_STRATEGY_URL_REGEX = (
-    r"localhost:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+    r"[^:]+:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
 )
--- a/localstack/services/sqs/constants.py
+++ b/localstack/services/sqs/constants.py
@@ -32,9 +32,9 @@ INTERNAL_QUEUE_ATTRIBUTES = [
 ]
 
 # URL regexes for various endpoint strategies
-STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
-DOMAIN_STRATEGY_URL_REGEX = r"(?P<region_name>[a-z0-9-]{1,}).queue.localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
-PATH_STRATEGY_URL_REGEX = r"localhost:\d{1,}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{1,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+DOMAIN_STRATEGY_URL_REGEX = r"(?P<region_name>[a-z0-9-]{1,}).queue.localhost.localstack.cloud:\d{1,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+PATH_STRATEGY_URL_REGEX = r"localhost:\d{1,}\/queue\/(?P<region_name>[a-z0-9-]{1,5})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
 LEGACY_STRATEGY_URL_REGEX = (
-    r"localhost:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+    r"localhost:\d{1,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
 )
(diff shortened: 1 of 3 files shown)


localstack/services/sqs/constants.py
```diff
@@ -30,3 +30,11 @@
     QueueAttributeName.LastModifiedTimestamp,
     QueueAttributeName.QueueArn,
 ]
+
+# URL regexes for various endpoint strategies
+STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+DOMAIN_STRATEGY_URL_REGEX = r"(?P<region_name>[a-z0-9-]{1,}).queue.localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+PATH_STRATEGY_URL_REGEX = r"localhost:\d{1,}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+LEGACY_STRATEGY_URL_REGEX = (
+    r"localhost:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
```
baermat, line 39: nit: the ports can be restricted to \d{1,5}

localstack/services/sqs/constants.py
```diff
@@ -30,3 +30,11 @@
     QueueAttributeName.LastModifiedTimestamp,
     QueueAttributeName.QueueArn,
 ]
+
+# URL regexes for various endpoint strategies
+STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+DOMAIN_STRATEGY_URL_REGEX = r"(?P<region_name>[a-z0-9-]{1,}).queue.localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
```
baermat, line 36: It seems that the Domain strategy regex does not allow the special case of the omitted us-east-1 region. This small correction should do the trick.
`((?P<region_name>[a-z0-9-]{1,}).)?queue.localhost.localstack.cloud:\d{1,}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$`
Also it might probably be good to add a test case in `test_is_sqs_queue_url` for this.

baermat (changes requested):
This looks great and so much more well defined :tada: I think there is one small mistake in the regex and a nit from my side, once that is tackled it is  a go from me.

bentsku (changes requested):
I think those regex should be able to match on externally provided hostname too, I think @simonrw recently removed all of these kind of regexes around the codebase, and he might be able to provide better insight here, but I don't think we should restrict to our own domain. 

Edit: it seems we are locking on our own domain for SQS, is there a reason why we didn't change it too Simon? (in `localstack/services/sqs/query_api.py`)

For example, for the path one, could we only match on the path and not the full URL?

Is there a performance cost of running 4 regexes on every request? 

edit: this might not need change, I would wait for Simon input on this, but if he's okay then it's good for me!

localstack/services/sqs/constants.py
viren-nadkarni: Good catch, fixed :+1:

simonrw:
> I think those regex should be able to match on externally provided hostname too, I think @simonrw recently removed all of these kind of regexes around the codebase, and he might be able to provide better insight here, but I don't think we should restrict to our own domain.
> 
> Edit: it seems we are locking on our own domain for SQS, is there a reason why we didn't change it too Simon? (in `localstack/services/sqs/query_api.py`)
> 
> For example, for the path one, could we only match on the path and not the full URL?
> 
> Is there a performance cost of running 4 regexes on every request?
> 
> edit: this might not need change, I would wait for Simon input on this, but if he's okay then it's good for me!

You're absolutely right, these regexes (and the handlers in `query_api.py`) should be able to handle arbitrary domains. Sorry this was missed as part of the `LOCALSTACK_HOST` migration. Are you happy to implement this, or would you like me to add a commit?

simonrw:
> Are you happy to implement this, or would you like me to add a commit?

Sorry I should have specified: @viren-nadkarni are you ok to replace `localhost.localstack.cloud` in your regexes for wildcards? Also it might be worth replacing the "."s in your regexes for "\." literal characters if you don't want a regex "any character".

localstack/services/sqs/constants.py
```diff
@@ -30,3 +30,11 @@
...
     QueueAttributeName.QueueArn,
 ]
+
+# URL regexes for various endpoint strategies
+STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,}).localhost.localstack.cloud:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+DOMAIN_STRATEGY_URL_REGEX = r"((?P<region_name>[a-z0-9-]{1,}).)?queue.localhost.localstack.cloud:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+PATH_STRATEGY_URL_REGEX = r"localhost:\d{4,5}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+LEGACY_STRATEGY_URL_REGEX = (
+    r"localhost:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
+)
```
simonrw, line 40: Perhaps something like this will work to capture (but ignore) any arbitrary domain?

```suggestion
STANDARD_STRATEGY_URL_REGEX = r"sqs.(?P<region_name>[a-z0-9-]{1,})\.[^:]+:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
DOMAIN_STRATEGY_URL_REGEX = r"((?P<region_name>[a-z0-9-]{1,})\.)?queue\.[^:]+:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
PATH_STRATEGY_URL_REGEX = r"localhost:\d{4,5}\/queue\/(?P<region_name>[a-z0-9-]{1,})\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
LEGACY_STRATEGY_URL_REGEX = (
    r"localhost:\d{4,5}\/(?P<account_id>\d{12})\/(?P<queue_name>[a-zA-Z0-9_-]+(.fifo)?)$"
)
```

tests/unit/test_sqs.py
```diff
@@ -140,3 +144,41 @@ def test_parse_queue_url_invalid():
...
+    assert (
+        is_sqs_queue_url("http://us-east-1.queue.localhost.localstack.cloud:4566/111111111111/foo1")
+        is True
+    )
+    assert is_sqs_queue_url("http://localhost:4566/queue/ap-south-1/222222222222/my-queue") is True
+    assert is_sqs_queue_url("http://localhost:4566/111111111111/bar") is True
+
+    # Path strategy uses `localhost`
+    assert (
+        is_sqs_queue_url("local.localstack.cloud:4566/queue/ap-south-1/222222222222/bar") is False
```
simonrw, line 181:
```suggestion
        is_sqs_queue_url("localhost.localstack.cloud:4566/queue/ap-south-1/222222222222/bar") is False
```

tests/unit/test_sqs.py
```diff
@@ -140,3 +144,41 @@ def test_parse_queue_url_invalid():
...
+    )
+    # Domain strategy may omit region
+    assert is_sqs_queue_url("http://queue.localhost.localstack.cloud:4566/111111111111/foo") is True
```
simonrw, line 184:
```suggestion
    assert is_sqs_queue_url("http://queue.localhost.localstack.cloud:4566/111111111111/foo") is True
    
    # Custom domain name
    assert is_sqs_queue_url("http://foo.bar:4566/queue/us-east-1/111111111111/foo") is True
    assert is_sqs_queue_url("http://us-east-1.queue.foo.bar:4566/111111111111/foo") is True
    assert is_sqs_queue_url("http://queue.foo.bar:4566/111111111111/foo") is True
    assert is_sqs_queue_url("http://sqs.us-east-1.foo.bar:4566/111111111111/foo") is True
    
```

tests/unit/test_sqs.py
simonrw, line 181:
```suggestion
    # Path strategy uses any domain name
    assert (
        is_sqs_queue_url("foo.bar:4566/queue/ap-south-1/222222222222/bar") is True
```

viren-nadkarni:
@simonrw: Thanks for the pointers! Agreed that omitting the hostname parts is much more elegant. Also thanks for updating the routes in https://github.com/localstack/localstack/pull/9642

bentsku (commented):
LGTM! It's a bit annoying that we need to pre-parse the URL in the service name parser before the router actually catches it, but I suppose it will go away at some point once the router will be before in the handler chain, we should be able to remove this utility then.

I'm actually not even sure this is still needed.. actually I've just ran the whole SQS test suite and nothing failed... do we even need this? Is this a remnant from before the router time and now that SQS legacy has been removed we don't need it?

Edit: I'm fairly certain we don't need this utility anymore, maybe it would be good to delete it instead, as we would not run 4 regexes for nothing?

viren-nadkarni:
> I'm fairly certain we don't need this utility anymore, maybe it would be good to delete it instead, as we would not run 4 regexes for nothing?

I think it is best to defer its removal to the service owners (cc: @thrau @baermat). I've changed the regex to make them compiled which should help in the performance aspect.

bentsku (approved):
LGTM for me then, I suppose it helps to assign a service to the request in the meantime. Thanks for tackling this!

baermat (approved):

PR #9639, opened 2023-11-15
apigateway: add IAM role to apigateway integration with services

sannya-singal:
## Motivation
The purpose of this PR is to enhance the stability of apigateway's integration with kinesis with multi-accounts and multi-regions. The main issue addressed is to introduce a proper IAM role required for the integration of these services. 

## Changes
This PR introduces the use of role arn when integrating kinesis with apigateway and add `credentials` field to the api gateway `put_integrations` api.

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`). The affected tests are:

`tests.aws.services.apigateway.test_apigateway_basic.TestIntegrations.test_api_gateway_kinesis_integration`

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -137,7 +137,6 @@ def get_service_factory(region_name: str, role_arn: str):
 @lru_cache(maxsize=64)
 def get_internal_mocked_headers(
     service_name: str,
-    account_id: str,
     region_name: str,
     source_arn: str,
     role_arn: str | None,
@@ -151,7 +150,7 @@ def get_internal_mocked_headers(
             ]["AccessKeyId"]
         )
     else:
-        access_key_id = account_id
+        access_key_id = None
     headers = aws_stack.mock_aws_request_headers(
         service=service_name, aws_access_key_id=access_key_id, region_name=region_name
     )
@@ -484,7 +483,6 @@ class KinesisIntegration(BackendIntegration):
         # forward records to target kinesis stream
         headers = get_internal_mocked_headers(
             service_name="kinesis",
-            account_id=invocation_context.account_id,
             region_name=invocation_context.region_name,
             role_arn=invocation_context.integration.get("credentials"),
             source_arn=get_source_arn(invocation_context),
@@ -674,7 +672,6 @@ class SQSIntegration(BackendIntegration):
 
         headers = get_internal_mocked_headers(
             service_name="sqs",
-            account_id=invocation_context.account_id,
             region_name=region_name,
             role_arn=invocation_context.integration.get("credentials"),
             source_arn=get_source_arn(invocation_context),
@@ -926,7 +923,6 @@ class EventBridgeIntegration(BackendIntegration):
         region_name = uri.split(":")[3]
         headers = get_internal_mocked_headers(
             service_name="events",
-            account_id=invocation_context.account_id,
             region_name=region_name,
             role_arn=invocation_context.integration.get("credentials"),
             source_arn=get_source_arn(invocation_context),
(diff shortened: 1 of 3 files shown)


bentsku:
This is @dfangl code for IAM, so I would prefer to have him review this. I believe the test might be failing because we're not properly providing the role ARN, and maybe we shouldn't use the account id of the invocation context for this.

bentsku:
Also, it would be good to get an -ext run against this to validate it, we need -ext to be green 😄

sannya-singal:
> Also, it would be good to get an -ext run against this to validate it, we need -ext to be green 😄

The `localstack-ext` integration test pipeline is already running here- https://github.com/localstack/localstack-ext/actions/runs/6876194122

localstack/services/apigateway/integration.py
```diff
@@ -147,7 +151,7 @@ def get_internal_mocked_headers(
             ]["AccessKeyId"]
         )
     else:
-        access_key_id = None
+        access_key_id = account_id
```
dfangl, line 154: Not sure about this one. It will probably break the IAM enforcement if there is no role specified, as in -ext, we test for the access key id being the `INTERNAL_AWS_ACCESS_KEY_ID` for IAM to see that the source principal is a service.

sannya-singal: @dfangl do we have the null `role_arn` just in case of -ext tests?

dfangl: There a basically two options to allow an apigateway integration for a service to work.
1) Use a role, which apigateway assumes, and performs the requests assumed as this role -> the permissions for the integrations are attached to the role
2) Use resource based policies, and attach the permissions to the target resource, and allow apigateway as service principal to access the resource.

This is basically  this if/else condition. In the second case, we currently assume that the INTERNAL_AWS_ACCESS_KEY_ID is set.

sannya-singal: @dfangl do we have something in specific that we can check to validate this fix for IAM enforcement in case no role is specified? I had already run the [-ext pipeline](https://github.com/localstack/localstack-ext/actions/runs/6876194122) and its green.

bentsku (changes requested):
Circling back on this PR: 

I believe we should not change code for a non-aws validated test if it's not obvious it is a bug on our side. From what I've understood from @dfangl , we could fix the test by using a role for Kinesis.

In this particular PR, I believe we have 2 possibilities:
-  make the test AWS validated before making actual changes in the code
- or introduce some kind of markers to skip the test in multi account usage, as it's not validated but should not introduce changes in the code if we are not sure it's the right one \cc @viren-nadkarni 

I would say making it validated and fixing it would be better, and would make the behavior aligned with AWS, but this might not be a simple task.
There might be middle fix by using a role like Daniel already indicated, so that `get_internal_mocked_headers` would be able to pick up the right account id.

localstack/services/apigateway/integration.py
sannya-singal: @dfangl I have made the required changes, please review whenever possible.

dfangl (approved):
I think providing the role is a good way of fixing the test itself. However, it also has to work without a role, we should write a aws_validated test for both cases, ideally with cross account and proper policies (maybe even a negative case for IAM), to check how much cross-account is supported without a role.

For this particular test fix this is fine from the IAM side, but I think we have a lot of more to test for the integrations in the future, which will uncover issues with our integration implementation.

bentsku (approved):
LGTM! Thanks for fixing the test by adding a role and following Daniel recommendations 🚀 would be nice to AWS validate it then

PR #9640, opened 2023-11-15
Pass context attributes to forward URL getter

viren-nadkarni:


alexrashed (commented):
Thanks for jumping on this, that's another piece to get rid of the deprecated `get_aws_account_id` and `get_region`.
But it might be better to just forward the complete context here instead of just these two specific context members. This way we wouldn't have to extend it in case other use cases would come up (f.e. by using the request ID, or parts of the request?).
What's your opinion?

localstack/services/dynamodb/provider.py
```diff
@@ -455,7 +459,9 @@ def handle_shell_ui_request(self, request: werkzeug.Request, req_path: str) -> R
         #  -> keeping this for now, to allow configuring custom installs; should consider removing it in the future
         # https://repost.aws/questions/QUHyIzoEDqQ3iOKlUEp1LPWQ#ANdBm9Nz9TRf6VqR3jZtcA1g
         req_path = f"/{req_path}" if not req_path.startswith("/") else req_path
-        url = f"{self.get_forward_url()}/shell{req_path}"
+        account_id = extract_account_id_from_headers(request.headers)
+        region_name = extract_region_from_headers(request.headers)
+        url = f"{self.get_forward_url(account_id, region_name)}/shell{req_path}"
```
viren-nadkarni, line 464: > it might be better to just forward the complete context

@alexrashed : I considered that, but we can avoid constructing a `RequestContext` object for uses like this

alexrashed: Ah, that makes sense, thanks!

alexrashed (approved):
Thanks for the clarification! Lookin' good! 🚀

PR #9641, opened 2023-11-15
Remove last uses of get_aws_account_id() and aws_stack.get_region()

viren-nadkarni:
## Motivation

This PR removes last notable remaining uses of `get_aws_account_id()` and `aws_stack.get_region()`.

These functions will be marked as deprecated or removed altogether in a subsequent PR.


localstack/services/apigateway/helpers.py
```diff
@@ -134,8 +135,8 @@ def get_apigateway_store(context: RequestContext) -> ApiGatewayStore:
 
 
 def get_apigateway_store_for_invocation(context: ApiInvocationContext) -> ApiGatewayStore:
-    account_id = context.account_id or get_aws_account_id()
-    region_name = context.region_name or aws_stack.get_region()
+    account_id = context.account_id or DEFAULT_AWS_ACCOUNT_ID
+    region_name = context.region_name or AWS_REGION_US_EAST_1
```
viren-nadkarni, line 139: This mimics the existing behaviour. At the time of executing this function, the account ID and region enricher handlers are not executed in the handler chain:

https://github.com/localstack/localstack/blob/f7eff1f5320c4ef35d6a0b88c519ed11431618b5/localstack/aws/app.py#L38

Thus, the `get_aws_account_id()` and `aws_stack.get_region()` return the fallback values.

localstack/utils/aws/aws_stack.py
```diff
@@ -102,20 +103,23 @@ def get_s3_hostname():
     return LOCALHOST
 
 
-def fix_account_id_in_arns(response, colon_delimiter=":", existing=None, replace=None):
+def fix_account_id_in_arns(
+    response, replacement: str, colon_delimiter: str = ":", existing: str | list[str] = None
+):
     """Fix the account ID in the ARNs returned in the given Flask response or string"""
-    existing = existing or ["123456789", "1234567890", "123456789012", get_aws_account_id()]
+    existing = existing or ["123456789", "1234567890", DEFAULT_ACCOUNT_ID]
```
viren-nadkarni, line 110: See https://github.com/getmoto/moto/blob/e409e17960c9baf2247bead2379456f25623e995/moto/core/models.py#L37

localstack/services/apigateway/helpers.py
bentsku: Thanks for the explanation! I hope this doesn't fallback often, I guess it falls back before actually failing but at least we have values. This makes sense and will be revisited in an upcoming APIGW refactor! thanks a lot

localstack/services/apigateway/integration.py
```diff
@@ -692,7 +691,7 @@ def invoke(self, invocation_context: ApiInvocationContext):
             queue_url = f"{config.internal_service_url()}/{account_id}/{queue}"
             new_request = f"{payload}&QueueUrl={queue_url}"
 
-        url = urljoin(config.internal_service_url(), f"{get_aws_account_id()}/{queue}")
+        url = urljoin(config.internal_service_url(), f"{account_id}/{queue}")
```
bentsku, line 694: Oh, damn, this was actually a bad bug 😬 shouldn't this URL depends on the SQS URL strategy actually?

bentsku (approved):
LGTM! This is great again, nice cleaning up! 🧹 that is quite the effort that lead to this final removal. Awesome! 
Also, nice spotting the bug in the APIGW SQS integration. It might need a follow up on our side to make the URL construction depend on the SQS URL strategy.

localstack/services/apigateway/integration.py
viren-nadkarni: Not necessarily, URL strategy controls how queues will be formatted in outputs. SQS will still recognise other queue URL formats if presented. Moreover the use of `config.internal_service_url()` prevents us from using the standard and domain strategies.

But you're right in another way, the region is currently omitted in the queue URL. I've added it with 1b4fe6a04. Great catch, thanks!

PR #9642, opened 2023-11-15
Update remaining host matching for localhost.localstack.cloud

simonrw:
## Motivation

There were a few usages of the constant `localhost.localstack.cloud` that remain after #9578 and #9600. This PR updates these. Ideally all routes should match on a wildcard domain suffix, e.g. `*.opensearch.*` or `*.execute-api.*` rather than e.g. `*.localhost.localstack.cloud` as we currently have. If this is not possible, we should at least match on `*.LOCALSTACK_HOST` since this has been customised.

Thanks @bentsku for [pointing out that I missed some](https://github.com/localstack/localstack/pull/9638#pullrequestreview-1731644001).

## Changes

* Update remaining functional uses of `localhost.localstack.cloud`. Note: the value is still ok in tests since they target localhost anyway.


simonrw:
The first test run succeeded but the "Community Integration Tests against Pro" job timed out.

dfangl (approved):
Looks good from my side, if all tests are green. We might want to test this against ext as well.

Issue #9637, opened 2023-11-14
bug: V3 S3 Provider .. object LastModified timestamps too precise?


matt-mercer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when querying the s3api  against latest localstack LastModified timestamps are being displayed to microsecond precision ..  I THINK this may be causing problems for us when using s3 sync ...

previously we could do 
`awslocal  s3 sync s3://local-resources/ca-refresh/ /tmp/certs --exact-timestamps --exclude="*" --include=crl.pem --no-progress` and if the file was unchanged ... it would not be re-downloaded, since pulling latest, now running 
`awslocal  s3 sync s3://local-resources/ca-refresh/ /tmp/certs --exact-timestamps --exclude="*" --include=crl.pem --no-progress` will download the file every time .. regardless even if the source file has not been modified 

I suspec this may be down to the LastModified timestamp format?

this is a s3api list-objects-v2 response from localstack
awslocal s3api list-objects-v2 --bucket local.crl.store
```{
    "Contents": [
        {
            "Key": ".well-known/jwks.json",
            "LastModified": "2023-11-14T22:22:51.188952+00:00",
            "ETag": "\"7c6fae6e0f26d2ca3d30b05c771af40d\"",
            "Size": 1632,```

this is one from aws 

aws s3api list-objects-v2 --bucket=crl.nhs.uk
```{
    "Contents": [
        {
            "Key": "PTL/",
            "LastModified": "2022-08-04T15:49:35+00:00",
```

### Expected Behavior

repeatedly running `aw s3 sync` with `--exact-timestamps` should not redownload a file if the source s3 object has not been modified ...

should be able to run a command like this against localstack and only download file once 
`awslocal  s3 sync s3://local-resources/ca-refresh/ /tmp/certs --exact-timestamps --exclude="*" --include=crl.pem --no-progress`

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

  docker-compose.yml

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```shell
echo 'hello' > test.txt
awslocal s3 cp ./test.txt s3://local-resources/
mkdir -p sync
awslocal  s3 sync s3://local-resources/ ./sync --exact-timestamps --exclude="*" --include=test.txt --no-progress
# download: s3://local-resources/test.txt to sync/test.txt

[...]

PR #9643, opened 2023-11-15
fix S3 timestamp precision

bentsku:
## Motivation
As reported in #9637, we return too precise timestamps. AWS returns ISO6081 milliseconds timestamps for S3 List operations, but with the milliseconds set to `000` (`%Y-%m-%dT%H:%M:%S.000Z`). This is handled in the serializer, as the declared type of the fields of the `TypedDict` is `datetime`

On a related note, AWS also returns RFC1123 timestamps for `LastModified` headers, as they should, but I've validated the behavior as well with AWS validated tests. 

## Changes
- updated the S3 serializer to return `ISO6081` datetime with this specific format. 
- wrote a whole bunch of test to validate that this does not alter behavior everywhere where we return timestamp in S3. I've checked most of the specs and the stubs, there might be some edge cases but I doubt they are using the base datetime parsing. 

## TODO

What's left to do:

Check all date times returned by S3, the specific ones are already covered by test, this is the rest.

- [X] write test for ListObjects and ListObjectsV2 
- [x] write test for `CopyObject`, `HeadObject`, `GetObject`, `GetObjectAttributes` (this can all be done in one test) -> `LastModified` with RFC1123 format, should be okay already
- [x] write test for `ListParts`, `ListObjectVersions` -> `LastModified` with ISO8601 format, should be same as ListObjects and ListObjectsV2
- [x] write test for `ListBuckets` (`CreationDate`), should be ISO8601 with 0 milliseconds too, to be verified
- [x] write test for `ListMultipartUploads` -> `Initiated`, should be ISO8601 with 0ms, to be verified

### DIFF
--- a/tests/aws/services/s3/test_s3_list_operations.py
+++ b/tests/aws/services/s3/test_s3_list_operations.py
@@ -2,6 +2,7 @@
 This file is to test specific behaviour of List* operations of S3, especially pagination, which is pretty specific to
 each implementation. They all have subtle differences which make it difficult to test.
 """
+import datetime
 import os
 from io import BytesIO
 
@@ -99,7 +100,10 @@ class TestS3ListObjects:
         snapshot.match("list-objects-marker-empty", resp)
 
     @markers.aws.validated
-    @markers.snapshot.skip_snapshot_verify(paths=["$..Prefix"])
+    @markers.snapshot.skip_snapshot_verify(
+        condition=is_v2_provider,
+        paths=["$..Prefix"],
+    )
     def test_s3_list_objects_empty_marker(self, s3_bucket, snapshot, aws_client):
         snapshot.add_transformer(snapshot.transform.s3_api())
         resp = aws_client.s3.list_objects(Bucket=s3_bucket, Marker="")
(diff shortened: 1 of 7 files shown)


tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -39,6 +40,18 @@ def _endpoint_url(region: str = "", localstack_host: str = None) -> str:
...
+def assert_timestamp_is_iso8061_s3_format(timestamp: str):
+    # the timestamp should be looking like the following
+    # 2023-11-15T12:02:40.000Z
+    assert timestamp.endswith(".000Z")
+    assert len(timestamp) == 24
+    # assert that it follows the right format and it does not raise an exception during parsing
+    parsed_ts = datetime.datetime.strptime(timestamp, "%Y-%m-%dT%H:%M:%S.%fZ")
+    assert parsed_ts.microsecond == 0
+
+    return True
```
alexrashed, line 52: nit: It's not really necessary to `return True` here and only use it with `assert` statements. If any of the asserts here is failing, then an exception is thrown anyway. If not, the return value is always `True`.

alexrashed (approved):
Nice, clean, and general solution in the serializer! 🚀 
I only found a small nitpick in the test assertions, but nothing blocking a merge! 💯

tests/aws/services/s3/test_s3_list_operations.py
bentsku: yes I agree, but I used `assert` under just for the look of validating something in the test 😬 it felt weird to validate the test with a simple method call. But it might make it a bit more... verbose and weird, I agree.

Maybe it's the same debate as `snapshot.match` 😅

bentsku: The return value would be `None` no?

alexrashed: Yes, the return value would be `None`, but I would just go with the assert method call without an assert statement in the test itself. The method is clearly named, the additional sugar is confusing and can never be triggered anyways.

bentsku: Will update, I'm convinced 😉

PR #9644, opened 2023-11-15
Refactor kinesis connector

dfangl:
## Motivation
Our `kinesis_connector.py` is currently used at several locations in our code, and the usage of it is in principal valid.
However, it has some problems with threads not shutting down, processes kept alive indefinitely, and a lot of unused configuration options cluttering the code.

To continue using it, we want:
* A clear shutdown of all processes and threads, with a single stop
* A reduction in created processes and threads
* Simpler code, by refactoring and removing unused configuration options

## Changes
* Enable new (-ext) Resource Provider for `AWS::KinesisAnalytics::Application`
* Refactor kinesis connector, get rid of output file reader threads, proper shutdown, more class encapsulation, less configuration variable
* Remove "environment" logic from aws stack
* Add try/finally around docker container startup logic to prevent stuck threads waiting for a event forever.


tests/aws/test_integration.py
dfangl, line 1: The changes here are only removal of the shard_id parameter of process records, and migration to the cleanups fixture instead of try: finally:. The changed lines within that block are just indentation changes.

dfangl:
Discussed with alex, and decided to go with patch for now. Up for discussion though, if a reviewer thinks it should be minor :)

simonrw (approved):
I have no strong opinions on the kinesis connector refactoring, but I think this is a step in the right direction. Anything to properly manage the lifetimes of the kinesis connector resources is a good thing!

PR #9633, opened 2023-11-14
Fix edge port usages

joe4dev:
Depends on https://github.com/localstack/localstack/pull/9600

## Motivation

We previously removed the write-usage of `EDGE_PORT` but kept read-access (e.g., `config.EDGE_PORT`) in our codebase. Given the many occurrences, it was easy to miss a few write-usages, and read-access gets mixed up with internal vs. external access.
I just noticed some of these issues while reviewing https://github.com/localstack/localstack/pull/9600 

## Changes

This PR fixes these inconsistencies by migrating all usages.

* Remove a few `EDGE_PORT` write usages we previously missed
* Fix some mixed-up `EDGE_PORT` usages migrating them to the new networking API
* Migrate usages of `LOCALSTACK_HOSTNAME` and `EDGE_PORT` in Lambda functions to `AWS_ENDPOINT_URL`. They are functionally equivalent.

## Testing

I have already run the full pipeline with these changes on top of the commits from https://github.com/localstack/localstack/pull/9600 All tests were green except for one flake in community `test_firehose_kinesis_to_s3`. Subsequently, I rebased upon master for a clean PR view.

PR #9646, opened 2023-11-15
Add subdomains to service url helpers

joe4dev:
## Motivation

While migrating the networking configuration, @simonrw [commented](https://github.com/localstack/localstack/pull/9633#discussion_r1393271424) that subdomains are commonly used but insufficiently supported by the current helpers.

It would be nice to use:
```
subdomains = f"{api_id}.execute-api"
url = f"{config.external_service_url(subdomains=subdomains)}/{stage}/{path}"
```
rather than misusing the hostname and duplicating default values.

**Why 3.0?**
We anticipate that this API will be used from LocalStack Extensions quite a bit and therefore compatibility with LocalStack >=3.0 is important.

## Changes

* Add optional `subdomains` parameter to `_service_url` helpers. This change is compatible with the existing code and primarily relevant internally within LocalStack as well as for LocalStack Extensions.
* Add parametized tests covering common scenarios
* Cleanup unused legacy variables from parametrized config tests

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1269,8 +1269,8 @@ def external_service_url(
     machine than LocalStack. The configurations LOCALSTACK_HOST and USE_SSL can customize these returned URLs.
     The optional parameters can be used to customize the defaults.
     Examples with default configuration:
-    * external_service_url() == http:localhost.localstack.cloud:4566
-    * external_service_url(subdomains="s3") == http:s3.localhost.localstack.cloud:4566
+    * external_service_url() == http://localhost.localstack.cloud:4566
+    * external_service_url(subdomains="s3") == http://s3.localhost.localstack.cloud:4566
     """
     protocol = protocol or get_protocol()
     subdomains = f"{subdomains}." if subdomains else ""
@@ -1289,8 +1289,8 @@ def internal_service_url(
     The configuration USE_SSL can customize these returned URLs but LOCALSTACK_HOST has no effect.
     The optional parameters can be used to customize the defaults.
     Examples with default configuration:
-    * internal_service_url() == http:localhost:4566
-    * internal_service_url(port=8080) == http:localhost:8080
+    * internal_service_url() == http://localhost:4566
+    * internal_service_url(port=8080) == http://localhost:8080
     """
     protocol = protocol or get_protocol()
     subdomains = f"{subdomains}." if subdomains else ""
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1260,28 +1260,43 @@ def get_protocol() -> str:
 
 
 def external_service_url(
-    host: Optional[str] = None, port: Optional[int] = None, protocol: Optional[str] = None
+    host: Optional[str] = None,
+    port: Optional[int] = None,
+    protocol: Optional[str] = None,
+    subdomains: Optional[str] = None,
 ) -> str:
-    """Returns a service URL to an external client used outside where LocalStack runs.
-    The configurations LOCALSTACK_HOST and USE_SSL can customize these returned URLs.
-    `host` can be used to overwrite the default for subdomains.
+    """Returns a service URL (e.g., SQS queue URL) to an external client (e.g., boto3) potentially running on another
+    machine than LocalStack. The configurations LOCALSTACK_HOST and USE_SSL can customize these returned URLs.
+    The optional parameters can be used to customize the defaults.
+    Examples with default configuration:
+    * external_service_url() == http:localhost.localstack.cloud:4566
+    * external_service_url(subdomains="s3") == http:s3.localhost.localstack.cloud:4566
     """
     protocol = protocol or get_protocol()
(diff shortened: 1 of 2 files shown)


localstack/config.py
```diff
@@ -1260,28 +1260,43 @@ def get_protocol() -> str:
...
 ) -> str:
-    """Returns a service URL to an external client used outside where LocalStack runs.
-    The configurations LOCALSTACK_HOST and USE_SSL can customize these returned URLs.
-    `host` can be used to overwrite the default for subdomains.
+    """Returns a service URL (e.g., SQS queue URL) to an external client (e.g., boto3) potentially running on another
+    machine than LocalStack. The configurations LOCALSTACK_HOST and USE_SSL can customize these returned URLs.
+    The optional parameters can be used to customize the defaults.
+    Examples with default configuration:
+    * external_service_url() == http:localhost.localstack.cloud:4566
+    * external_service_url(subdomains="s3") == http:s3.localhost.localstack.cloud:4566
```
simonrw, line 1273:
```suggestion
    * external_service_url() == http://localhost.localstack.cloud:4566
    * external_service_url(subdomains="s3") == http://s3.localhost.localstack.cloud:4566
```

tests/unit/test_config.py
```diff
@@ -276,3 +266,53 @@ def test_port_out_of_range(self, port):
...
+        assert url == expected_service_url
+
+    def test_internal_service_url(self):
+        # defaults
+        assert internal_service_url() == "http://localhost:4566"
+        # subdomains
+        assert (
+            internal_service_url(subdomains="abc.execute-api")
+            == "http://abc.execute-api.localhost:4566"
+        )
```
simonrw, line 318: TIL that `localhost` [is a TLD](https://datatracker.ietf.org/doc/html/rfc2606#section-2)!

localstack/config.py
```diff
@@ -1260,28 +1260,43 @@ def get_protocol() -> str:
...
+    subdomains: Optional[str] = None,
 ) -> str:
-    """Returns a service URL for internal use within where LocalStack runs. Cannot be customized
-    through LOCALSTACK_HOST because we assume LocalStack runs on the same host (i.e., localhost).
+    """Returns a service URL for internal use within LocalStack (i.e., same host).
+    The configuration USE_SSL can customize these returned URLs but LOCALSTACK_HOST has no effect.
+    The optional parameters can be used to customize the defaults.
+    Examples with default configuration:
+    * internal_service_url() == http:localhost:4566
+    * internal_service_url(port=8080) == http:localhost:8080
```
simonrw, line 1293:
```suggestion
    * internal_service_url() == http://localhost:4566
    * internal_service_url(port=8080) == http://localhost:8080
```

localstack/config.py
```diff
@@ -1260,28 +1260,43 @@ def get_protocol() -> str:
...
     """
     protocol = protocol or get_protocol()
+    subdomains = f"{subdomains}." if subdomains else ""
     host = host or LOCALHOST
     port = port or GATEWAY_LISTEN[0].port
-    return f"{protocol}://{host}:{port}"
+    return f"{protocol}://{subdomains}{host}:{port}"
```
simonrw, line 1299: minor: do we need to guard against trying to apply subdomains if the caller specifies `host="127.0.0.1"` for example? We may want to use the IP address in some circumstances, e.g. if ipv6 causes us problems again

tests/unit/test_config.py
```diff
@@ -67,15 +67,7 @@ def ip() -> str:
...
-    """
-    Post-v2 we are deriving
-
-    * EDGE_PORT
-
-    from GATEWAY_LISTEN. We are also ensuring the configuration behaves
-    well with LOCALSTACK_HOST, i.e. if LOCALSTACK_HOST is supplied and
-    GATEWAY_LISTEN is not, then we should propagate LOCALSTACK_HOST configuration
-    into GATEWAY_LISTEN.
+    """We are deriving GATEWAY_LISTEN and LOCALSTACK_HOST from provided environment variables.
```
simonrw, line 70: Thanks for updating this test :thumbsup:

simonrw (approved):
Looks like a great and worthwhile addition, thanks!

localstack/config.py
joe4dev: ups, fixed in a pending commit. Thanks.

localstack/config.py
joe4dev: ups, fixed in a pending commit. Thanks.

localstack/config.py
joe4dev: mhm, good point. Not sure whether we should be as smart as we think or wait for a use case and keep it simple for now...
We would need a reliable is_ip check (ideally for IPv4, IPv5). What do you think?

joe4dev:
ok, merging now because the last commit only updated docstrings (i.e., no code changes)

PR #9652, opened 2023-11-15
Issue 9602 use salt lengths in parity with real kms

georgecodes:
Modified KMS to use specific salt lengths to PSS signatures in line with AWS KMS

## Motivation

I discovered a disparity between AWS KMS and localstack KMS during signing of some mtls 
messages and raised this issue

https://github.com/localstack/localstack/issues/9602

When constructing the parameters for a PSS signature, an appropriate salt length is selected, rather than allowing a default which proves to be non-deterministic.

## Testing

Signing a message with a probabilistic scheme, and then verifying the signature offline with, for example, OpenSSL. The verification will fail. Contrast with AWS KMS, where it will not.

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -8,6 +8,8 @@ from random import getrandbits
 import pytest
 from botocore.config import Config
 from botocore.exceptions import ClientError
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import ec, padding, utils
 from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15
 from cryptography.hazmat.primitives.serialization import load_der_public_key
 
@@ -60,6 +62,37 @@ def _get_alias(kms_client, alias_name, key_id=None):
             break
         next_token = response["nextToken"]
     return None
+
+
+def enforce_salt_length(signing_algorithm, message_type):
+    kwargs = {}
+
+    if "SHA_256" in signing_algorithm:
+        hasher = hashes.SHA256()
+        salt = 32
+    elif "SHA_384" in signing_algorithm:
+        hasher = hashes.SHA384()
+        salt = 48
+    elif "SHA_512" in signing_algorithm:
+        hasher = hashes.SHA512()
+        salt = 64
+
+    if message_type == "DIGEST":
+        kwargs["algorithm"] = utils.Prehashed(hasher)
+    else:
+        kwargs["algorithm"] = hasher
+
+    if signing_algorithm.startswith("ECDSA"):
+        kwargs["signature_algorithm"] = ec.ECDSA(algorithm=kwargs.pop("algorithm", None))
+        return kwargs
+
+    if signing_algorithm.startswith("RSA"):
+        if "PKCS" in signing_algorithm:
+            kwargs["padding"] = padding.PKCS1v15()
+        elif "PSS" in signing_algorithm:
+            print("Adding padding")
+            kwargs["padding"] = padding.PSS(mgf=padding.MGF1(hasher), salt_length=salt)
+    return kwargs
 
 
 class TestKMS:
@@ -435,8 +468,17 @@ class TestKMS:
         verification = aws_client.kms.verify(
             MessageType="RAW", Signature=signature["Signature"], Message=plaintext, **kwargs
         )
+
         snapshot.match("verification", verification)
         assert verification["SignatureValid"]
+
+        response = aws_client.kms.get_public_key(KeyId=key_id)
+        public_key_bytes = response.get("PublicKey")
+        key = load_der_public_key(public_key_bytes)
+
+        raw_vargs = enforce_salt_length(sign_algo, "RAW")
+
+        key.verify(signature=signature["Signature"], data=plaintext, **raw_vargs)
 
         # Ensure pre-hashed messages can be signed and verified
         signature = aws_client.kms.sign(MessageType="DIGEST", Message=digest, **kwargs)
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -472,6 +472,7 @@ class TestKMS:
         snapshot.match("verification", verification)
         assert verification["SignatureValid"]
 
+        # Ensure signatures can be verified using the public key and specific params
         response = aws_client.kms.get_public_key(KeyId=key_id)
         public_key_bytes = response.get("PublicKey")
         key = load_der_public_key(public_key_bytes)
@@ -486,6 +487,10 @@ class TestKMS:
             MessageType="DIGEST", Signature=signature["Signature"], Message=digest, **kwargs
         )
         assert verification["SignatureValid"]
+
+        digest_vargs = enforce_salt_length(sign_algo, "DIGEST")
+
+        key.verify(signature=signature["Signature"], data=digest, **digest_vargs)
 
         # Ensure bad digest raises during signing
         with pytest.raises(ClientError) as exc:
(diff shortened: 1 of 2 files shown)


georgecodes:
Unable to add a semver label as none are available to me. It's a patch though.

sannya-singal:
Hi @georgecodes, thanks for contributing to localstack 🚀. Kindly add AWS validated tests to the changes similar to the ones present [here](https://github.com/localstack/localstack/blob/master/tests/aws/services/kms/test_kms.py).

georgecodes:
Yep no problemOn 4 Jan 2024, at 07:38, Sannya Singal ***@***.***> wrote:﻿
Hi @georgecodes, thanks for contributing to localstack 🚀. Kindly add AWS validated tests to the changes similar to the ones present here.

tests/aws/services/kms/test_kms.py
```diff
@@ -62,6 +64,37 @@ def _get_alias(kms_client, alias_name, key_id=None):
...
+
+    if signing_algorithm.startswith("ECDSA"):
+        kwargs["signature_algorithm"] = ec.ECDSA(algorithm=kwargs.pop("algorithm", None))
+        return kwargs
+
+    if signing_algorithm.startswith("RSA"):
+        if "PKCS" in signing_algorithm:
+            kwargs["padding"] = padding.PKCS1v15()
+        elif "PSS" in signing_algorithm:
+            print("Adding padding")
```
sannya-singal, line 93: kindly remove the print statements being used.

tests/aws/services/kms/test_kms.py
```diff
@@ -62,6 +64,37 @@ def _get_alias(kms_client, alias_name, key_id=None):
...
+        kwargs["algorithm"] = hasher
+
+    if signing_algorithm.startswith("ECDSA"):
+        kwargs["signature_algorithm"] = ec.ECDSA(algorithm=kwargs.pop("algorithm", None))
+        return kwargs
+
+    if signing_algorithm.startswith("RSA"):
+        if "PKCS" in signing_algorithm:
+            kwargs["padding"] = padding.PKCS1v15()
+        elif "PSS" in signing_algorithm:
```
sannya-singal, line 92: we need to add the logic in `kms/models.py`

tests/aws/services/kms/test_kms.py
```diff
@@ -435,16 +468,30 @@ def test_sign_verify(self, kms_create_key, snapshot, key_spec, sign_algo, aws_cl
...
 
+        # Ensure signatures can be verified using the public key and specific params
+        response = aws_client.kms.get_public_key(KeyId=key_id)
+        public_key_bytes = response.get("PublicKey")
+        key = load_der_public_key(public_key_bytes)
+
+        raw_vargs = enforce_salt_length(sign_algo, "RAW")
+
+        key.verify(signature=signature["Signature"], data=plaintext, **raw_vargs)
+
```
sannya-singal, line 483: need to add snapshot validation for these changes.

sannya-singal (changes requested):

tests/aws/services/kms/test_kms.py
sannya-singal: we usually write such logics in `models.py` since this piece of code can be used by other backend parts we shouldn't write it in test files. 

Eg: When we `decrypt` a kms key, we add the logic part of [here](https://github.com/localstack/localstack/blob/e393a7fdeeab4f178540d0be9daf60e438dd68af/localstack/services/kms/models.py#L273-L279) in the `kms/models.py`, and [test](https://github.com/localstack/localstack/blob/e393a7fdeeab4f178540d0be9daf60e438dd68af/tests/aws/services/kms/test_kms.py#L515-L517) just calls the handler which further uses the decrypt logic added in models.

sannya-singal:
Hi @georgecodes! I have added some more context to the comments. Kindly refer to our [contribution guidelines](https://docs.localstack.cloud/contributing/) and let us know if you need some further assistance. 

Also please make sure that all the checks in the pipeline are green! 
Happy to help if you come across any further issues 🙂

tests/aws/services/kms/test_kms.py
FlorinAsavoaie: In fact I think the author of this PR is right in testing this way. Testing by doing the verification offline using the Public Key will clearly show why the current behaviour of LocalStack is NOT actually compatible with AWS, even though it claims so. RSAPSS signing in LocalStack is 100% not compatible with AWS KMS.

tests/aws/services/kms/test_kms.py
FlorinAsavoaie: This logic is definitely only required in testing. Why would testing-only code go into an application file?

FlorinAsavoaie:
This is a significant bug that causes severe incompatibility with AWS KMS when using RSAPSS signing. Basically the code is 100% NOT compatible if you don't do the verification in KMS (which clearly cannot be done in lots of apps - as verification is usually done by other peers).

I am interested to improve this PR as needed until it is merged if the original author cannot continue to work on it.

sannya-singal:
Hey @FlorinAsavoaie thanks for your interest in resolving this bug 🙌 🎉 Kindly go ahead and work on this fix: if you are planning to re-use a part of code by @georgecodes by cherry picking the commits, don't forget to give him the credits 😉 Thanks 🚀

georgecodes:
> This is a significant bug that causes severe incompatibility with AWS KMS when using RSAPSS signing. Basically the code is 100% NOT compatible if you don't do the verification in KMS (which clearly cannot be done in lots of apps - as verification is usually done by other peers).
> 
> I am interested to improve this PR as needed until it is merged if the original author cannot continue to work on it.

Hi, I would love to spend some time on this but I am busy. I'm happy to revisit and address the comments later, but for now I have solved the problem by re-implementing KMS in pure Java, as I needed mocked KMS, and only KMS, for the specific work I was doing at the time I raised the issue.

But I'm also happy for you to pick this up if you have the time, naturally.

FlorinAsavoaie:
@georgecodes would you mind trying to rebase the PR and removing that printed comment?

Other than that, the PR should be good to go probably. If not, I can definitely cherry-pick your code but it sounds faster and more accurate from a recognition perspective if you do this, doesn't look like more than a couple minutes thing.

tests/aws/services/kms/test_kms.py
sannya-singal: This should already be solved here: https://github.com/localstack/localstack/blob/f6c21e30cd112832e4b56cd6831df7e5b947c666/localstack/services/kms/models.py#L394 and https://github.com/localstack/localstack/blob/f6c21e30cd112832e4b56cd6831df7e5b947c666/localstack/services/kms/models.py#L416

tests/aws/services/kms/test_kms.py
sannya-singal: We can rather use the existing code as mentioned in the above comments.

georgecodes:
@FlorinAsavoaie I will do so, yes

sannya-singal:
Hey @georgecodes @FlorinAsavoaie 👋 Wanted to follow up on this PR, are you actively working on this? And if you would like some more inputs on further steps?

FlorinAsavoaie:
This is not a priority for me at the moment, but the bug itself should probably be fixed as the behavior is significantly different comparing to AWS KMS.

sannya-singal:
Thanks for letting me know @FlorinAsavoaie, this is already a part of our backlog. I will wait for @georgecodes response. Thanks 🙌

sannya-singal:
Thank you for your contributions! I’m closing this PR for now, as no one is currently working on it. This issue shall be addressed in a future follow-up PR. However, if you’d like to continue working on it, please don’t hesitate to open another PR—I’d be happy to assist you with it.

PR #9653, opened 2023-11-15
add cw tests for lambda Invocations and Errors metrics

steffyP:
## Motivation
While testing the pro-samples we found that lambda is not reporting the `Errors` metrics anymore. 
The metric should be reported when a lambda invocation fails.

We didn't have integration tests for lambda metrics yet, so I added those now.
/cc @joe4dev 

## Changes
* added aws-validated tests
* skipping `test_lambda_invoke_error` for LS as it is currently not working -> this should be fixed 🙂

### DIFF
--- a/tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
@@ -1,0 +1,128 @@
+from datetime import datetime, timedelta
+from typing import TYPE_CHECKING
+
+import pytest
+
+from localstack.testing.pytest import markers
+from localstack.testing.pytest.snapshot import is_aws
+from localstack.utils.strings import short_uid
+
+if TYPE_CHECKING:
+    from mypy_boto3_cloudwatch import CloudWatchClient
+from localstack.utils.sync import retry
+
+TEST_SUCCESSFUL_LAMBDA = """
+def handler(event, context):
+    return {"success": "ok"}
+"""
+
+TEST_FAILING_LAMBDA = """
+def handler(event, context):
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Nice tests, clean and fast 🚀 thanks a lot for adding those, will make it easier to implement the fix 👍

tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        # wait for metrics
+        invocation_res = retry(
+            lambda: self._wait_for_lambda_metric(
+                aws_client.cloudwatch,
+                fn_name=fn_name,
+                metric_name="Invocations",
+                expected_return=[1.0],
+            ),
+            retries=200 if is_aws() else 20,
+            sleep=10 if is_aws() else 1,
```
bentsku, line 84: and oof, that's... a long wait 😭

steffyP: you never know how long it takes on AWS 😅 but in general it's long for the metrics to show up. I didn't want to spam the AWS endpoint, so I just went for a longer period. 
you think it could become an issue? maybe once we really automated the snapshot updates 😄 ⏳

bentsku: oh no sorry, it was a general comment, I know for firehose sometimes the wait are really long, I guess that's where we shine? 😄
I guess those kind of tests is why it's hard to automate, but we can always introduce some kind of markers at that time 😄

PR #9656, opened 2023-11-15
fix test_s3_timestamp_precision

bentsku:
## Motivation
The test was failing if there was more than one bucket in the `ListBuckets` call because of bad test cleanup

## Changes
- fix the test to really handle the case where they are multiple buckets (my xml parsing guarding was wrong)


alexrashed (approved):
🦸

PR #9662, opened 2023-11-16
fix: supported multi account for test kinesis forward chain

macnev2013:
## Motivation

This PR adds support for multi account to `test_kinesis_lambda_forward_chain`.

## Changes
Earlier we were using `create_bucket` api (sdk) for s3 which requires `LocationConstraint` for bucket if created in region not equal to `us-east-1`. We have updated `create_bucket` api (sdk) to fixture which takes care of adding `LocationConstraint`.

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111, TEST_AWS_ACCESS_KEY_ID=111111111111, TEST_AWS_REGION=us-west-1`). The affected tests are:
- `tests.aws.test_integration.test_kinesis_lambda_forward_chain`


viren-nadkarni (approved):
LGTM :heavy_check_mark:

PR #9672, opened 2023-11-17
Fix int exception for API GW

lakkeger:
## Motivation
Botocore exception for integer casting on empty strings when changing API gateway's minimum compression size from 0 to disabled.

## Changes
- not allowing minimum compression size to be an empty string
- add some get api calls after update operations during tests

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -828,11 +828,6 @@ def apply_json_patch_safe(subject, patch_operations, in_place=True, return_list=
                 path = operation["path"]
                 common.assign_to_path(subject, path, value={}, delimiter="/")
 
-            # minimum_compression_size is a unique path as it's a nullable integer,
-            # it would throw an error if it's an empty string
-            if operation["path"] == "/minimum_compression_size" and operation.get("value") == "":
-                operation["value"] = None
-
             result = apply_patch(subject, [operation], in_place=in_place)
             if not in_place:
                 subject = result
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -828,6 +828,11 @@ def apply_json_patch_safe(subject, patch_operations, in_place=True, return_list=
                 path = operation["path"]
                 common.assign_to_path(subject, path, value={}, delimiter="/")
 
+            # minimum_compression_size is a unique path as it's a nullable integer,
+            # it would throw an error if it's an empty string
+            if operation["path"] == "/minimum_compression_size" and operation.get("value") == "":
+                operation["value"] = None
+
             result = apply_patch(subject, [operation], in_place=in_place)
             if not in_place:
                 subject = result
--- a/tests/aws/services/apigateway/test_apigateway_api.py
+++ b/tests/aws/services/apigateway/test_apigateway_api.py
@@ -417,10 +417,8 @@ class TestApiGatewayApi:
         snapshot.match("enable-compression", response)
 
         # check that listing is not exploding after update, null -> 10
-        try:
-            aws_client.apigateway.get_rest_api(restApiId=api_id)
-        except ClientError:
-            assert False
+        response = aws_client.apigateway.get_rest_api(restApiId=api_id)
+        assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
 
         # from the docs: to disable compression, apply a replace operation with the value property set to null or
         # omit the value property.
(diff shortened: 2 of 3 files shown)


lakkeger:

lakkeger:
localstack/chalice-local#11

.envrc
bentsku, line 1: did this slip through?

localstack/services/apigateway/helpers.py
```diff
@@ -828,6 +828,11 @@ def apply_json_patch_safe(subject, patch_operations, in_place=True, return_list=
                 path = operation["path"]
                 common.assign_to_path(subject, path, value={}, delimiter="/")
 
+            # minimum_compression_size is a unique path as it's a nullable integer,
+            # it would throw an error if it's an empty string
+            if operation["path"] == "/minimum_compression_size" and operation.get("value") == "":
+                operation["value"] = None
+
```
bentsku, line 835: I'm not sure I would add a specific case here in this helper. Wouldn't it be possible to fix the issue after the patch has been applied? So we don't have "subject" specific fixes in the global util? 

Something like the fix for `vpcEndpointIds` done after the patches in `update_rest_api`:
```python
        _patch_api_gateway_entity(rest_api, fixed_patch_ops)

        # fix data types after patches have been applied
        endpoint_configs = rest_api.endpoint_configuration or {}
        if isinstance(endpoint_configs.get("vpcEndpointIds"), str):
            endpoint_configs["vpcEndpointIds"] = [endpoint_configs["vpcEndpointIds"]]
        if rest_api.minimum_compression_size == "":
            rest_api.minimum_compression_size = None
```

I think this would be easier and would contain the fix where the issue lies, while keeping this util free of case specific logic.

tests/aws/services/apigateway/test_apigateway_api.py
```diff
@@ -433,6 +433,7 @@ def test_update_rest_api_compression(self, apigw_create_rest_api, snapshot, aws_
         response = aws_client.apigateway.update_rest_api(
             restApiId=api_id, patchOperations=patch_operations
         )
+        # aws_client.apigateway.get_rest_api(restApiId=api_id)
```
bentsku: is this left on purpose? Was there a way to reproduce the issue with the boto client?

tests/bootstrap/conftest.py
bentsku, line 1: what are those changes? is this a leftover?

tests/conftest.py
bentsku, line 1: same comment here, is it a leftover?

bentsku (changes requested):
Thanks a lot for tackling this issue, this is a great find! 🚀  It seems some files were changed with no link to the PR, not sure if it slipped through?

I would apply the fix somewhere more specific and not introduce an operation-specific fix in a "global" util if possible. It seems there already a "post patch" logic in `update_rest_api`, so we could add the fix there maybe?

bentsku (approved):
Awesome, LGTM! So it wasn't possible to reproduce with boto? What a weird issue, maybe boto was fixing it itself under the hood before sending the request?

But this fix looks good, awesome!

lakkeger:
It was possible to reproduce it with boto and the pytests were throwing the exception too if you not create the api with the empty string but update it from a valid integer (ie 10 -> "") value. It's a really dormant error as the update itself is not showing any sign of an exception until you want to list/describe the APIs.

The origin is the default handling of the `replace` operations with `None` value in the `helpers.py` file [here](https://github.com/localstack/localstack/blob/c3d368d503185e1bd21eb0ddbbdfa70814d401fe/localstack/services/apigateway/helpers.py#L804) since this is a unique case with a nullable integer value.

bentsku:
Ah, I see! Maybe we should have a test that fails without the fix then, to avoid any regression?

tests/aws/services/apigateway/test_apigateway_api.py
```diff
@@ -416,6 +416,12 @@ def test_update_rest_api_compression(self, apigw_create_rest_api, snapshot, aws_
         )
         snapshot.match("enable-compression", response)
 
+        # check that listing is not exploding after update, null -> 10
+        try:
+            aws_client.apigateway.get_rest_api(restApiId=api_id)
+        except ClientError:
+            assert False
```
bentsku, line 423: You don't need the `try...except` clause here, you can leave the exception be raised and it will make the test fail. 
```suggestion

        aws_client.apigateway.get_rest_api(restApiId=api_id)
```

bentsku (changes requested):
Almost good to go! You can simplify the test by leaving the exception be raised, as now if the test failed we wouldn't really know why because of the `assert False`. Sorry to make you update it again 😭

lakkeger:
How about if we keep the current way but giving a custom message with the
assert that contains the exception too, so we know exactly which update has
issues from the beginning 🤔

bentsku:
I think the error log from pytest will give you the line so you can know where it fails I'd say? If you really want to assert, you can add the line we talked about: 
```python
response = aws_client.apigateway.get_rest_api(restApiId=api_id)
```

I think that's a clear way of expressing we want this call to succeed after an update call.

bentsku (approved):
LGTM! Awesome, sorry for the multiple back and forth!

Issue #9666, opened 2023-11-17
bug: S3 Object Metadata not being persisted on localstack:3.0+ with presigned urls


deathran:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, when using localstack S3 to generate a presigned upload that has metadata written into the object, the metadata is not persisted to the object. Resulting in no metadata being returned when `aws s3api head-object` is executed.

### Expected Behavior

As a localstack user
I want localstack S3 to correctly persist the metadata into the s3 object when using a presigned upload
So that I can test if my presign upload feature is working as expected

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
# docker-compose.yml

    localstack:
        image: ${DEP_PROXY_PREFIX}/localstack/localstack:2.3
        hostname: localstack
        ports:
            - 4566:4566
        environment:
            Services: 's3,sqs,sns'
            AWS_ACCESS_KEY_ID: localstack
            AWS_SECRET_ACCESS_KEY: localstack
            AWS_DEFAULT_REGION: eu-central-1
            EAGER_SERVICE_LOADING: 1
        volumes:
            - localstack:/var/lib/localstack
            - '/var/run/docker.sock:/var/run/docker.sock
```

```bash
docker compose up
```
#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```typescript
// test.ts
import { HeadObjectCommand, PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
import { uploadFileViaHttp } from './tests/integration/scripts';
import { randomBytes } from 'crypto';

async function runError() {
    const client = new S3Client({
        region: 'eu-central-1',
        endpoint: 'http://localhost:4566',
        forcePathStyle: true
    });
    const buffer = randomBytes(100);

    const initialMetaData = {
        firstid: 'b8066141-3c1a-4c10-96a7-0664aeebe2aa',
        secondid: '800001',
        thirdid: 'b0372e50-f4f3-4ae2-9a44-7b8c70604cf4'
    };
    const putObjectCommand = new PutObjectCommand({
        Bucket: 'my-test-bucket',
        Key: 'test-key.txt',
        ContentLength: buffer.length,
        Metadata: initialMetaData
    });

[...]

Issue #9664, opened 2023-11-17
Getting 403 AccessDenied on upload with presigned url, localstack 3.0.0


antal-kajzer-epam:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

hi guys,

I'm stuggling to test our S3 logic locally. The service that calls localstack s3 is getting the following error:
`2023-11-13 15:45:57 com.amazonaws.services.s3.model.AmazonS3Exception: There were headers present in the request which were not signed (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: ea7d6d50-1e86-452c-a1c6-00fca1a237a5; S3 Extended Request ID: 9Gjjt1m+cjU4OPvX9O9/8RuvnG41MRb/18Oux2o5H5MY7ISNTlXN+Dz9IG62/ILVxhAGI0qyPfg=; Proxy: null)`

The localstack logs are the following:

> LocalStack version: 3.0.0
LocalStack build date: 2023-11-16
LocalStack build git hash: 3cd32364
2023-11-17T08:46:39.471  INFO --- [-functhread6] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
2023-11-17T08:46:39.471  INFO --- [-functhread6] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
2023-11-17T08:46:39.723  INFO --- [  MainThread] localstack.utils.bootstrap : Execution of "start_runtime_components" took 903.21ms
Ready.
2023-11-17T08:47:33.276  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.GetBucketAcl => 404 (NoSuchBucket)
2023-11-17T08:47:33.381  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.CreateBucket => 200
2023-11-17T08:47:33.406  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.GetBucketAcl => 404 (NoSuchBucket)
2023-11-17T08:47:33.413  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.CreateBucket => 200
2023-11-17T08:47:33.425  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.GetBucketAcl => 404 (NoSuchBucket)
[...]

PR #9676, opened 2023-11-17
fix s3 skip presign for missing SignedHeaders and fix query string param to headers

bentsku:
## Motivation
Following #9664 and #9666, 2 different reports about pre-signed URL behavior.
- first issue is about raising an exception when the request is not valid even with `S3_SKIP_SIGNATURE_VALIDATION` not specified to False (default behavior is to skip), this is due to some exception handling not being caught for the specific case of Signed Headers not being given.
- second issue is about specific Signature V3 that is different from boto. Basically some SDK will put what would normally be a header inside the query string (JS SDK V3), Boto will add it to the signed headers and you will have to add it manually, etc... and because it was not in the signed headers, we would put it back as a header for the `ServiceRequestParser` to properly parse all request parameters.

## Changes
- add proper exception handling for missing `SignedHeaders`, saving them in the context to be able to raise `AccessDenied` for the SigV4 validation later on in the validation process
- add SigV4 behavior to add query string parameters starting with `x-amz-` to the list of parameters to put back as headers for the service request parser 
- updated some lambdas for the tests, as I need to test it with the AWS JS SDK v2 and v3

_fixes #9664_
_fixes  #9666_


steffyP (approved):
Great work, @bentsku! 
Thanks for always jumping on the s3 issues so quickly 🚀 🙇‍♀️ and always making sure about AWS parity! 

The changes LGTM!

PR #9679, opened 2023-11-17
fix SNS publish across region exception

bentsku:
## Motivation
Some users had issues with region in another issue, but their trace logs showed another issue badly handled by SNS when the topic didn't exist and wasn't properly checked. 

## Changes
Added better check and a AWS validated test when trying to publish from another region.


viren-nadkarni (approved):
LGTM :heavy_check_mark:

baermat (approved):

Issue #9678, opened 2023-11-17
bug: SQS FIFO: The maximum length of MessageDeduplicationId must be 128 characters


mbermas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Now it is 127 characters:
https://github.com/localstack/localstack/blob/32a7946f878ace1a8bdd098bb3c9d53898424a78/localstack/services/sqs/provider.py#L435-L438

### Expected Behavior

The maximum length of MessageDeduplicationId is 128 characters. AWS docs:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html#API_SendMessage_RequestSyntax

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS:
- LocalStack:
```

hanshal101:
Hello @mbermas , I checked the AWS Documentation, and reviewed the code the max MessageDeduplicationId should be 128 Characters. 
Correct me if I'm wrong , here it should be :
```len(fifo_id) <= 128```

PR #9680, opened 2023-11-17
Fix max fifo_id len

mbermas:
## Motivation
Fixes https://github.com/localstack/localstack/issues/9678
AWS docs: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html#SQS-SendMessage-request-MessageDeduplicationId

## Changes
Change max `fifo_id` len from 127 to 128 characters

### DIFF
--- a/localstack/services/lambda_/invocation/metrics.py
+++ b/localstack/services/lambda_/invocation/metrics.py
@@ -5,30 +5,33 @@ from localstack.utils.cloudwatch.cloudwatch_util import publish_lambda_metric
 LOG = logging.getLogger(__name__)
 
 
-def record_cw_metric_invocation(function_name: str, region_name: str):
+def record_cw_metric_invocation(function_name: str, account_id: str, region_name: str):
     try:
         publish_lambda_metric(
             "Invocations",
             1,
             {"func_name": function_name},
             region_name=region_name,
+            account_id=account_id,
         )
     except Exception as e:
         LOG.debug("Failed to send CloudWatch metric for Lambda invocation: %s", e)
 
 
-def record_cw_metric_error(function_name: str, region_name: str):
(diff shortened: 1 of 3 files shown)


mbermas:

thrau (changes requested):
Thanks for the contribution @mbermas.

The change of `>= 128` to `> 128` looks correct to me. The exception message we had before was wrong, but the new one also doesn't seem correct. Here's the error I get:

```json
        "Error": {
          "Code": "InvalidParameterValue",
          "Message": "Value 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111 for parameter MessageGroupId is invalid. Reason: MessageGroupId can only include alphanumeric and punctuation characters. 1 to 128 in length.",
          "QueryErrorCode": "InvalidParameterValueException",
          "Type": "Sender"
        },
```

it would be great if we could add a proper snapshot test to the PR and make it work. here's the test i used:

```python 
    @markers.aws.validated
    def test_message_group_id_length_for_fifo_queue(self, sqs_create_queue, snapshot, aws_client):
        queue_url = sqs_create_queue(
            QueueName=f"queue-{short_uid()}.fifo",
            Attributes={
                "FifoQueue": "true",
                "ContentBasedDeduplication": "false",
            },
        )

        message_group_id = "1" * 129

        with pytest.raises(ClientError) as e:
            aws_client.sqs.send_message(
                QueueUrl=queue_url,
                MessageBody="test",
                MessageGroupId=message_group_id,
                MessageDeduplicationId="1",
            )

        snapshot.match("error-response", e.value.response)
```

See https://docs.localstack.cloud/contributing/parity-testing/ on how to run parity tests

PR #9684, opened 2023-11-17
fix get_object_tagging should work on versioned buckets without providing an explicit VersionId

matt-mercer:
## Motivation

previously  `s3.get_object_tagging` worked on versioned buckets, without specifying an explicit `VersionId`

## Changes
* localstack/services/s3/v3/provider.py::put_object_tagging sets tags version to `s3_object.version_id` if present
* localstack/services/s3/v3/provider.py::get_object_tagging uses `s3_object.version_id` if pesent
* small addtion to `tests/aws/services/s3/test_s3_api.py::TestS3BucketObjectTagging::test_object_tagging_versioned` to also exercise `put_object` with `Tagging` set on a versioned bucket

## Testing

covered by `tests/aws/services/s3/test_s3_api.py::TestS3BucketObjectTagging::test_object_tagging_versioned`

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -2518,7 +2518,7 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         validate_tag_set(tagging["TagSet"], type_set="object")
 
-        key_id = get_unique_key_id(bucket, key, version_id)
+        key_id = get_unique_key_id(bucket, key, s3_object.version_id)
         # remove the previous tags before setting the new ones, it overwrites the whole TagSet
         store.TAGS.tags.pop(key_id, None)
         store.TAGS.tag_resource(key_id, tags=tagging["TagSet"])
@@ -2553,9 +2553,9 @@ class S3Provider(S3Api, ServiceLifecycleHook):
             e.Key = f"{bucket}/{key}"
             raise e
 
-        tag_set = store.TAGS.list_tags_for_resource(get_unique_key_id(bucket, key, version_id))[
-            "Tags"
-        ]
+        tag_set = store.TAGS.list_tags_for_resource(
+            get_unique_key_id(bucket, key, s3_object.version_id)
+        )["Tags"]
         response = GetObjectTaggingOutput(TagSet=tag_set)
         if s3_object.version_id:
             response["VersionId"] = s3_object.version_id
--- a/tests/aws/services/s3/test_s3_api.py
+++ b/tests/aws/services/s3/test_s3_api.py
@@ -1,5 +1,6 @@
 import json
 from operator import itemgetter
+from urllib.parse import urlencode
 
 import pytest
 from botocore.exceptions import ClientError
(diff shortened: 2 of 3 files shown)


bentsku:
Hello @matt-mercer and thanks again for your contribution.

However, I believe this functionality is already covered by quite many tests in `tests.aws.services.s3.test_s3_api.TestS3BucketObjectTagging`, and your test is actually green on this PR. 

I think the tagging should work as expected, however, would you have by chance encountered this issue with pre-signed URL? We have a fix in the works that should be merged on Monday (#9676) that, if your issue was with pre-signed URL, would fix. 

Basically, every parameter that would have been in headers but passed as query string parameter and not in SignedHeaders would be missing from the object when using pre-signed. Could that have been your issue?

matt-mercer:
> Hello @matt-mercer and thanks again for your contribution.
> 
> However, I believe this functionality is already covered by quite many tests in `tests.aws.services.s3.test_s3_api.TestS3BucketObjectTagging`, and your test is actually green on this PR (except for the legacy provider in the specific job for it, because of default bucket encryption, nothing major).
> 
> I think the tagging should work as expected, however, would you have by chance encountered this issue with pre-signed URL? We have a fix in the works that should be merged on Monday (#9676) that, if your issue was with pre-signed URL, would fix.
> 
> Basically, every parameter that would have been in headers in a regular request but passed as query string parameter and not in SignedHeaders would be missing from the object when using pre-signed (see #9666). Could that have been your issue?
> 
> However, if it was a different issue, maybe it's not present with the Python SDK but with another one and we could have missed it when implementing tags? Could you please share how you had the issue?

Hi @bentsku  ...  it is indeed, I had to go through a couple of iterations before I found the underlying issue, have updated the PR accordingly .. but the problem I was seeing was specific to buckets with versioning enabled.

matt-mercer:
no idea what I'm doing to upset the PR labels :joy:

tests/aws/services/s3/test_s3.py
```diff
@@ -904,6 +904,40 @@ def test_put_object_tagging_empty_list(self, s3_bucket, snapshot, aws_client):
         object_tags = aws_client.s3.get_object_tagging(Bucket=s3_bucket, Key=key)
         snapshot.match("deleted-object-tags", object_tags)
 
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(
+        condition=is_v2_provider,
+        paths=["$..ServerSideEncryption"],
+    )
+    def test_get_object_tagging_with_versioned_bucket(self, s3_bucket, snapshot, aws_client):
```
bentsku, line 912: there is already a test here: `tests.aws.services.s3.test_s3_api.TestS3BucketObjectTagging.test_object_tagging_versioned` which is doing almost the same thing, maybe we could add the assertion there to not have duplicated tests?

I can see why the test was not failing, it's because the exact same issue you found in `get_object_tagging` is also in `put_object_tagging`, could you please fix it there too? 

You could update the test to set tags in `put_object` too so we could merge your test and the existing one I think.

localstack/services/s3/v3/provider.py
```diff
@@ -2553,6 +2553,8 @@ def get_object_tagging(
             e.Key = f"{bucket}/{key}"
             raise e
 
+        version_id = version_id or s3_object.version_id
+
         tag_set = store.TAGS.list_tags_for_resource(get_unique_key_id(bucket, key, version_id))[
```
bentsku, line 2558: Would this work?
```suggestion
        tag_set = store.TAGS.list_tags_for_resource(get_unique_key_id(bucket, key, s3_object.version_id))[
```

bentsku (commented):
Thanks a lot for your fix! This is awesome 😄 I was pretty puzzled because I had a test testing this exact behavior... however, `put_object_tagging` has the exact same issue, so that's why I didn't spot the bug you've found.

If you could merge both tests by adding your case, and also add the fix in `put_object_tagging`, that would be awesome! Thanks a lot again!

localstack/services/s3/v3/provider.py
matt-mercer: yes .. better

bentsku:
Would you like me to merge both tests, fix `put_object_tagging` and push the changes?

matt-mercer:
> We are currently planning a bug fix release at the end of the week, so it would be nice to get this merged for it. If you don't have the time to fix the comments, that's alright, no pressure!

@bentsku  .. updated as suggested, should be good to go

bentsku (approved):
Awesome! Thanks a lot for addressing the comments and finding this tricky bug. Thank you so much again for your contribution.

PR #9692, opened 2023-11-20
Introduce DNS_SKIP_LOCAL_RESOLUTION config

simonrw:
## Motivation

Sometimes we may wish to skip resolving a DNS name with the LocalStack DNS server. For example, when using [transparent endpoint injection](https://docs.localstack.cloud/user-guide/tools/transparent-endpoint-injection/) some AWS URLs need to resolve to addresses on the public internet in order to e.g. fetch something from an S3 bucket.

We currently support this feature with the configuration variable `DNS_LOCAL_NAME_PATTERNS`, but this name sounds like it does the opposite -> "local name patterns" sounds like it will either:

* add a name to be resolved locally, or
* resolve a specific name to the local address.

This is not the case.

## Changes

We cannot remove this configuration variable since it is a breaking change, and we have not introduced a deprecation phase. This PR starts the deprecation phase of `DNS_LOCAL_NAME_PATTERNS`, and instead lets the user configure `DNS_SKIP_LOCAL_RESOLUTION`. Currently during this deprecation phase, we accept both variables, however in the future `DNS_LOCAL_NAME_PATTERNS` will not be used to configure DNS resolution.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1015,7 +1015,8 @@ DNS_PORT = int(os.environ.get("DNS_PORT", "53"))
 # Comma-separated list of regex patterns for DNS names to resolve locally.
 # Any DNS name not matched against any of the patterns on this whitelist
 # will resolve to the real DNS entry, rather than the local one.
-DNS_LOCAL_NAME_PATTERNS = (os.environ.get("DNS_LOCAL_NAME_PATTERNS") or "").strip()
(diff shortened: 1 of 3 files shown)


dfangl (approved):
I think this is a good idea, and the new name is clearer. Not entirely sure about it, but I cannot name a better alternative for now :)

Issue #9695, opened 2023-11-20
bug: SFN - `$$.Execution.StartTime` is giving `"21:09:14.150266"` instead of `"2019-03-26T20:14:13.192Z"` format


BrianMcBrayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, when passing `$$.Execution.StartTime` as a parameter to a SFN lambda, it is coming in in the format `HH:MM:SS.###`.

### Expected Behavior

`$$.Execution.StartTime` should be an ISO 8601 datetime string.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

This is run in a docker compose file.

Step function definition is

```json
{
  "Comment": "Sample SFN",
  "StartAt": "SampleLambda",
  "States": {    
    "SampleLambda": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "End": true,
      "Parameters": {
        "FunctionName": "Sample-Function",
        "Payload": {
          "data": {
            "startDateTime.$": "$$.Execution.StartTime"
          }
        }
      }
    }
  }
}
```

### Environment

```markdown
- OS: Mac OS Ventura 13.6.1
- LocalStack:
LocalStack version: 3.0.1.dev
LocalStack build date: 2023-11-20
LocalStack build git hash: 3e32b438
```

### Anything else?

Thank you so much for your hard work.

PR #9702, opened 2023-11-21
Fix StepFunctions execution date formatting

dominikschubert:
## Motivation

Resolves https://github.com/localstack/localstack/issues/9695

The returned string needs to look like this: "2023-11-21T06:27:31.545Z" which is not the default isoformat which we've been using previously. It also seems like AWS always returns UTC timestamps as far as I can see.

## Changes

- Changes format of serialized timestamp for execution start and end date.
- Explicitly use UTC everywhere when creating timestamps to hopefully avoid any future issues.

## TODO

What's left to do:

- [ ] cover other (de-)serializations, e.g. strptime

### DIFF
--- a/localstack/services/stepfunctions/asl/component/state/state.py
+++ b/localstack/services/stepfunctions/asl/component/state/state.py
@@ -117,7 +117,7 @@ class CommonStateField(EvalComponent, ABC):
         )
 
         env.context_object_manager.context_object["State"] = State(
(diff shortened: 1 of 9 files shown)


BrianMcBrayer (commented):
Looks great to me

BrianMcBrayer:
@dominikschubert @MEPalma any chance on this getting merged? I am sorry to bother you and I appreciate all your work.

MEPalma (approved):
Thank you @dominikschubert!

PR #9709, opened 2023-11-22
remove sqs-query from health endpoint

alexrashed:
## Motivation
This PR removes the service alias for SQS to support the old `query` protocol from the health endpoint (because it's not its own AWS service, just an alias for an existing service used to distinguish between the routing and allow loading a separate spec).

## Changes
Explicitly removes `sqs-query` from the health endpoint as a temporary workaround (until the "service alias" has been removed again).

/cc @Pive01

### DIFF
--- a/localstack/services/internal.py
+++ b/localstack/services/internal.py
@@ -87,7 +87,10 @@ class HealthResource:
         if reload:
             self.service_manager.check_all()
         services = {
-            service: state.value for service, state in self.service_manager.get_states().items()
+            service: state.value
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! thanks a lot for fixing this, which could be a bit weird for users. That will be a big search for all the "sqs-query" when removing the alias 😄

alexrashed:
@bentsku Oh yes, looking forward to cleaning that up. Actually I'm already on it: https://github.com/localstack/localstack/pull/9710

PR #9710, opened 2023-11-22
fix sqs multi-protocol handling and upgrade botocore

alexrashed:
## Motivation
This PR aims at cleaning up after the quick fixes with the protocol switch in SQS of `botocore` from `query` to `json` (and now back again).

These changes of the protocol are highly problematic, because this means that clients are suddenly changing the way how they talk to AWS (and in turn also with LocalStack). There's nothing we can do on our side besides implementing the new protocol (in addition to the old protocol to remain downwards-compatible with other or older clients).

Here's a timeline of what happened:
- **2023-05-04** - `botocore` switched the protocol for SQS from `query` to `json` in version [`1.29.127`](https://github.com/boto/botocore/releases/tag/1.29.127) (specifically this commit: https://github.com/boto/botocore/commit/2395012bd369f89b0bcfbbf830718866fc6ae166).
  - The changes were quickly hitting LocalStack users: https://github.com/localstack/localstack/issues/8267
- **2023-05-05** - `botocore` reverts the change with https://github.com/boto/botocore/pull/2931 in version [`1.29.128`](https://github.com/boto/botocore/releases/tag/1.29.128).
- **2023-11-09** - `botocore` again performs the switch in version [`1.31.81`](https://github.com/boto/botocore/releases/tag/1.31.81) (specifically this commit: https://github.com/boto/botocore/commit/84c784761d4cde0c0ce3b2b32382f970d23b3cc8).
- **2023-11-11** - Two days after the switch, we were ready with a somewhat hacky solution: https://github.com/localstack/localstack/pull/8268
  - https://github.com/localstack/localstack/pull/9607 - Performs a quick cleanup after the initial merge of the new protocol support.
  - https://github.com/localstack/localstack/pull/9634 - Fixes an issue in the routing
  - https://github.com/localstack/localstack/pull/9611 - Adjusts the ASF specs and the provider implementation accordingly
  - https://github.com/localstack/localstack/pull/9709 - Removes the artificial service from the list of services in the health endpoint
- **2023-11-11** - `botocore` changes the exception handling for SQS in version [`1.31.85`](https://github.com/boto/botocore/releases/tag/1.31.85) in https://github.com/boto/botocore/pull/3054.
  - These changes have been addressed with https://github.com/localstack/localstack/pull/9627.
- **2023-11-15** - `botocore` switches to gzipped specs which is not compatible with moto at the time.
  - https://github.com/localstack/localstack/commit/815d6c77245e3ebfb9d298a70fe16bd12df032f7 pins `botocore` as a first response.
- **2023-11-22** - https://github.com/localstack/localstack/pull/9624 upgrades `moto` such that it contains https://github.com/getmoto/moto/pull/7030.
- **2023-11-22** - https://github.com/localstack/localstack/pull/9667 unpins `botocore` again
- **2023-11-22** - `botocore` switches the protocol for SQS back from `json` to `query` in version [`1.32.6`](https://github.com/boto/botocore/releases/tag/1.32.6) (specifically this PR: https://github.com/boto/botocore/pull/3071)
  - https://github.com/localstack/localstack/pull/9714 pins `botocore` as a first response.

This is where we are at. With this PR we want to:
- [x] Get rid of the artificial "sqs-query" service. The term `sqs-query` should only be used in very specific sections explicitly handling the `query` spec.
- [x] Remove the pin on `botocore`, effectively switching back to `query` as the default protocol, but still support `json`.
- [x] Remove the pin of AWS CLI in the `Dockerfile` introduced with https://github.com/localstack/localstack/pull/9767.

## Changes
- The first commit cleans up after the previous efforts to adjust to the chaotic changes around the protocol changes in SQS.
  - This change centers around switching from differentiating the requests on protocol level, rather than on service level.
  - It removes the `sqs-query` service "alias", and removes all special cases handling this service alias.
  - It adjusts the `serializer` and `parser` such that they differentiate based on the protocol of the given service model (instead of only based on the service).
  - It adjusts the `service_router` such that it delivers a service model instead of only the service name (because that would be ambiguous, there are two service models for SQS).
- The second commit upgrades the ASF API stubs based on the latest `botocore` version (since they had to be hold back, see https://github.com/localstack/localstack/pull/9735).
- The third commit actually performs the upgrade / unpinning of `botocore`.
  - This means another switch from `json` back to `query` as default protocol for SQS.
  - The internalized spec for SQS is switched from `query` to `json` (the "non-default" protocol).
  - Changes the `serializer` adjustments which are necessary due to the duality of the protocols, their divergence, and the fact that we only generate a single ASF stub based on one of the specs.
  - Adjusts the SQS provider to some recent changes (empty lists / dicts are not contained anymore in certain cases, like messages or tags).
  - Fixes / updates quite a lot of tests and snapshots.
    - Also parameterizes most of the SQS tests such that they are snapshot tested against both protocols.
- The fourth commit contains some test fixes related to the new changes in the S3 specs / generated code. /cc @bentsku

### DIFF
--- a/localstack/aws/handlers/service.py
+++ b/localstack/aws/handlers/service.py
@@ -34,14 +34,12 @@ class ServiceNameParser(Handler):
         if context.service:
             return
 
-        service = determine_aws_service_model(context.request)
+        service_model = determine_aws_service_model(context.request)
 
-        if not service:
+        if not service_model:
             return
 
-        context.service = service
-        headers = context.request.headers
-        headers["x-localstack-tgt-api"] = service  # TODO: probably no longer needed
+        context.service = service_model
 
 
 class ServiceRequestParser(Handler):
--- a/localstack/aws/handlers/service.py
+++ b/localstack/aws/handlers/service.py
@@ -2,7 +2,6 @@
 import logging
 import traceback
 from collections import defaultdict
-from functools import lru_cache
 from typing import Any, Dict, Union
 
 from botocore.model import OperationModel, ServiceModel
@@ -17,9 +16,8 @@ from ..chain import CompositeResponseHandler, ExceptionHandler, Handler, Handler
 from ..client import parse_response, parse_service_exception
 from ..protocol.parser import RequestParser, create_parser
 from ..protocol.serializer import create_serializer
-from ..protocol.service_router import determine_aws_service_name
+from ..protocol.service_router import determine_aws_service_model
 from ..skeleton import Skeleton, create_skeleton
-from ..spec import load_service
 
 LOG = logging.getLogger(__name__)
 
@@ -36,18 +34,14 @@ class ServiceNameParser(Handler):
         if context.service:
             return
 
-        service = determine_aws_service_name(context.request)
+        service = determine_aws_service_model(context.request)
 
         if not service:
             return
 
-        context.service = self.get_service_model(service)
+        context.service = service
         headers = context.request.headers
         headers["x-localstack-tgt-api"] = service  # TODO: probably no longer needed
-
-    @lru_cache()
-    def get_service_model(self, service: str) -> ServiceModel:
-        return load_service(service)
 
 
 class ServiceRequestParser(Handler):
@@ -69,17 +63,8 @@ class ServiceRequestParser(Handler):
 
         return self.parse_and_enrich(context)
 
-    def get_parser(self, service: ServiceModel):
-        name = service.service_name
-
-        if name in self.parsers:
-            return self.parsers[name]
-
-        self.parsers[name] = create_parser(service)
-        return self.parsers[name]
-
     def parse_and_enrich(self, context: RequestContext):
-        parser = self.get_parser(context.service)
+        parser = create_parser(context.service)
         operation, instance = parser.parse(context.request)
 
         # enrich context
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -1120,6 +1120,7 @@ class SQSQueryRequestParser(QueryRequestParser):
         return primary_name
 
 
+@functools.cache
 def create_parser(service: ServiceModel) -> RequestParser:
     """
     Creates the right parser for the given service model.
@@ -1133,8 +1134,8 @@ def create_parser(service: ServiceModel) -> RequestParser:
     # within the parser implementations, the service-specific parser implementations (basically the implicit /
     # informally more specific protocol implementation) has precedence over the more general protocol-specific parsers.
     service_specific_parsers = {
-        "s3": S3RequestParser,
-        "sqs-query": SQSQueryRequestParser,
+        "s3": {"rest-xml": S3RequestParser},
+        "sqs": {"query": SQSQueryRequestParser},
     }
     protocol_specific_parsers = {
         "query": QueryRequestParser,
@@ -1144,9 +1145,12 @@ def create_parser(service: ServiceModel) -> RequestParser:
         "ec2": EC2RequestParser,
     }
 
-    # Try to select a service-specific parser implementation
-    if service.service_name in service_specific_parsers:
-        return service_specific_parsers[service.service_name](service)
+    # Try to select a service- and protocol-specific parser implementation
+    if (
+        service.service_name in service_specific_parsers
+        and service.protocol in service_specific_parsers[service.service_name]
+    ):
+        return service_specific_parsers[service.service_name][service.protocol](service)
     else:
         # Otherwise, pick the protocol-specific parser for the protocol of the service
         return protocol_specific_parsers[service.protocol](service)
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -9,7 +9,12 @@ from werkzeug.http import parse_dict_header
 
 import localstack
 from localstack import config
-from localstack.aws.spec import ServiceCatalog, build_service_index_cache, load_service_index_cache
+from localstack.aws.spec import (
+    ServiceCatalog,
+    ServiceModelIdentifier,
+    build_service_index_cache,
+    load_service_index_cache,
+)
 from localstack.constants import LOCALHOST_HOSTNAME, PATH_USER_REQUEST
 from localstack.http import Request
 from localstack.services.s3.utils import uses_host_addressing
@@ -73,43 +78,43 @@ def _extract_service_indicators(request: Request) -> _ServiceIndicators:
 signing_name_path_prefix_rules = {
     # custom rules based on URI path prefixes that are not easily generalizable
     "apigateway": {
-        "/v2": "apigatewayv2",
+        "/v2": ServiceModelIdentifier("apigatewayv2"),
     },
     "appconfig": {
-        "/configuration": "appconfigdata",
+        "/configuration": ServiceModelIdentifier("appconfigdata"),
     },
     "execute-api": {
-        "/@connections": "apigatewaymanagementapi",
-        "/participant": "connectparticipant",
-        "*": "iot",
+        "/@connections": ServiceModelIdentifier("apigatewaymanagementapi"),
+        "/participant": ServiceModelIdentifier("connectparticipant"),
+        "*": ServiceModelIdentifier("iot"),
     },
     "ses": {
-        "/v2": "sesv2",
-        "/v1": "pinpoint-email",
+        "/v2": ServiceModelIdentifier("sesv2"),
+        "/v1": ServiceModelIdentifier("pinpoint-email"),
     },
     "greengrass": {
-        "/greengrass/v2/": "greengrassv2",
+        "/greengrass/v2/": ServiceModelIdentifier("greengrassv2"),
     },
     "cloudsearch": {
-        "/2013-01-01": "cloudsearchdomain",
+        "/2013-01-01": ServiceModelIdentifier("cloudsearchdomain"),
     },
-    "s3": {"/v20180820": "s3control"},
+    "s3": {"/v20180820": ServiceModelIdentifier("s3control")},
     "iot1click": {
-        "/projects": "iot1click-projects",
-        "/devices": "iot1click-devices",
+        "/projects": ServiceModelIdentifier("iot1click-projects"),
+        "/devices": ServiceModelIdentifier("iot1click-devices"),
     },
     "es": {
-        "/2015-01-01": "es",
-        "/2021-01-01": "opensearch",
+        "/2015-01-01": ServiceModelIdentifier("es"),
+        "/2021-01-01": ServiceModelIdentifier("opensearch"),
     },
     "sagemaker": {
-        "/endpoints": "sagemaker-runtime",
-        "/human-loops": "sagemaker-a2i-runtime",
+        "/endpoints": ServiceModelIdentifier("sagemaker-runtime"),
+        "/human-loops": ServiceModelIdentifier("sagemaker-a2i-runtime"),
     },
 }
 
 
-def custom_signing_name_rules(signing_name: str, path: str) -> Optional[str]:
+def custom_signing_name_rules(signing_name: str, path: str) -> Optional[ServiceModelIdentifier]:
     """
     Rules which are based on the signing name (in the auth header) and the request path.
     """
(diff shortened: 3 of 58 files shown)


tests/aws/services/sqs/test_sqs.py
```diff
@@ -72,18 +75,23 @@ def sqs_snapshot_transformer(snapshot):
     snapshot.add_transformer(snapshot.transform.sqs_api())
 
 
+@pytest.fixture(params=["sqs", "sqs_json"])
+def aws_sqs_client(aws_client, request: str) -> "SQSClient":
+    yield getattr(aws_client, request.param)
```
alexrashed, line 80: FYI: This fixture basically parameterizes every test which uses it to test both protocols, `query` and `json`. The change from the default `aws_client` fixture to this one in most of the tests adds 122 new test executions. This causes most of the changed lines in this test file.

tests/aws/services/sqs/test_sqs.py
```diff
@@ -1724,89 +1749,87 @@ def test_fifo_set_content_based_deduplication_strategy(
...
 
-        aws_client.sqs.delete_message(
+        aws_sqs_client.delete_message(
             QueueUrl=queue_url, ReceiptHandle=result_recv["Messages"][0]["ReceiptHandle"]
         )
-        result_recv = aws_client.sqs.receive_message(QueueUrl=queue_url)
-        assert "Messages" in result_recv
-        assert result_recv["Messages"] == []
+        result_recv = aws_sqs_client.receive_message(QueueUrl=queue_url)
+        assert "Messages" not in result_recv or result_recv["Messages"] == []
```
baermat, line 1798: as before

tests/aws/services/sqs/test_sqs.py
```diff
@@ -2996,35 +3022,34 @@ def test_invalid_string_attributes_cause_invalid_parameter_value_error(
...
-        result_recv_2 = aws_client.sqs.receive_message(QueueUrl=queue_url)
+        result_recv_1 = aws_sqs_client.receive_message(QueueUrl=queue_url)
+        result_recv_2 = aws_sqs_client.receive_message(QueueUrl=queue_url)
         assert (
             result_recv_1.get("Messages")[0]["MessageId"]
             == result_receive.get("Messages")[0]["MessageId"]
         )
-        assert "Messages" in result_recv_2
-        assert result_recv_2["Messages"] == []
+        assert "Messages" not in result_recv_2 or result_recv_2["Messages"] == []
```
baermat, line 3047: as before

tests/aws/services/sqs/test_sqs.py
```diff
@@ -3055,172 +3080,165 @@ def test_dead_letter_queue_execution_lambda_mapping_preserves_id(
...
         assert int(approx_nr_of_messages["Attributes"]["ApproximateNumberOfMessages"]) > 1
 
-        aws_client.sqs.purge_queue(QueueUrl=queue_url)
+        aws_sqs_client.purge_queue(QueueUrl=queue_url)
 
-        receive_result = aws_client.sqs.receive_message(QueueUrl=queue_url)
-        assert "Messages" in receive_result
-        assert receive_result["Messages"] == []
+        receive_result = aws_sqs_client.receive_message(QueueUrl=queue_url)
+        assert "Messages" not in receive_result or receive_result["Messages"] == []
```
baermat, line 3127: as before

tests/aws/services/sqs/test_sqs.py
```diff
@@ -3055,172 +3080,165 @@ def test_dead_letter_queue_execution_lambda_mapping_preserves_id(
...
+        aws_sqs_client.purge_queue(QueueUrl=queue_url)
 
         # wait for visibility timeout to expire
         time.sleep(3)
 
-        receive_result = aws_client.sqs.receive_message(QueueUrl=queue_url, WaitTimeSeconds=1)
-        assert "Messages" in receive_result
-        assert receive_result["Messages"] == []
+        receive_result = aws_sqs_client.receive_message(QueueUrl=queue_url, WaitTimeSeconds=1)
+        assert "Messages" not in receive_result or receive_result["Messages"] == []
```
baermat, line 3162: as before

tests/aws/services/sqs/test_sqs.py
```diff
@@ -3055,172 +3080,165 @@ def test_dead_letter_queue_execution_lambda_mapping_preserves_id(
...
+        aws_sqs_client.purge_queue(QueueUrl=queue_url)
 
         # wait for delay to expire
         time.sleep(2)
 
-        receive_result = aws_client.sqs.receive_message(QueueUrl=queue_url, WaitTimeSeconds=1)
-        assert "Messages" in receive_result
-        assert receive_result["Messages"] == []
+        receive_result = aws_sqs_client.receive_message(QueueUrl=queue_url, WaitTimeSeconds=1)
+        assert "Messages" not in receive_result or receive_result["Messages"] == []
```
baermat, line 3179: as before

tests/aws/services/sqs/test_sqs.py
```diff
@@ -251,34 +262,33 @@ def test_send_receive_message_multiple_queues(self, sqs_create_queue, aws_client
         aws_client.sqs.send_message(QueueUrl=queue0, MessageBody="message")
 
         result = aws_client.sqs.receive_message(QueueUrl=queue1)
-        assert "Messages" in result
-        assert result["Messages"] == []
+        assert "Messages" not in result or result["Messages"] == []
```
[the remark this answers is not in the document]
joe4dev: For empty lists, AWS often omits the field entirely, independent of the protocol. I noticed this behavior recently while parity testing different SDK versions (which use different protocols) in the Lambda multiruntime tests.

I moved my Notion testing page to the SQS service to demonstrate the behavior for the `listQueues` operation (see [here](https://www.notion.so/localstack/SQS-SDK-parity-testing-164bfb9dbdd541928fd2b2abc275bac7?pvs=4)). It looks like we have an SQS parity issue when listing empty queues (unrelated to this PR).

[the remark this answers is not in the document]
alexrashed: Thanks for the comment! Since this is tested in dozens of the snapshot tests (specifically for each individual protocol due to the new fixture being used in the tests), I wanted to make these assertions independent of the protocol. Also because the next switch is around the corner (I'm quite sure the `botocore` team will try another run on the switch somewhat soon).
Also, it seems like they are currently actually changing the behavior server-side (the `json` service is _not_ returning empty lists anymore). But a gain, I think this is something which should belong in the snapshot verification. Here we actually semantically want to check if there are no messages left (and not explicitly want to test the protocol).

I'll close the other comments concerning this, but I'm happy to discuss this. If we want to change this, then obviously I'll apply it everywhere.

bentsku (approved):
Wow, this is an extensive set of changes. Biggest takeaway the clean separation between the protocol and the service name, made possible with the nice changes in the service router!
We've also had a sync looking at a lot of test changes, and it looks great! I like the parametrization making sure we properly test both protocols. 

There's a lot of clean up in here, and it seems #9828 has been merged, which would maybe need additional snapshots and parametrization as well. 

All in all, LGTM! Thank you so much for addressing this and making a nice clean fix 🚀

localstack/aws/handlers/service.py
```diff
@@ -63,17 +57,8 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
...
-
-        if name in self.parsers:
-            return self.parsers[name]
-
-        self.parsers[name] = create_parser(service)
-        return self.parsers[name]
-
     def parse_and_enrich(self, context: RequestContext):
-        parser = self.get_parser(context.service)
+        parser = create_parser(context.service)
```
thrau, line 61: i guess it should be fine creating the parser every time! very little overhead since it caries no state.

localstack/aws/protocol/parser.py
```diff
@@ -1132,8 +1133,8 @@ def create_parser(service: ServiceModel) -> RequestParser:
     # within the parser implementations, the service-specific parser implementations (basically the implicit /
     # informally more specific protocol implementation) has precedence over the more general protocol-specific parsers.
     service_specific_parsers = {
-        "s3": S3RequestParser,
-        "sqs-query": SQSQueryRequestParser,
+        "s3": {"rest-xml": S3RequestParser},
+        "sqs": {"query": SQSQueryRequestParser},
```
thrau, line 1137: very nice! :100:

localstack/aws/spec.py
```diff
@@ -70,6 +70,10 @@ def load_service(service: ServiceName, version: str = None, model_type="service-
     For example: load_service("sqs", "2012-11-05")
     """
     service_description = loader.load_service_model(service, model_type, version)
+    # if the service name is sqs-json, we just loaded our internalized SQS query spec,
+    # the service name needs to be set to standard sqs
+    if service == "sqs-json":
+        service = "sqs"
     return ServiceModel(service_description, service)
```
thrau, line 77: It would be really good if we could get `protocol: str` as an optional parameter into our `load_service` implementation instead of relying on the fake service names. i understand that we currently need them under the hood to know which spec file to load, basically our api should understand protocols as first-class citizen.

thrau (commented):
Overall this is already a huge improvement over what we had before! I do have a few open questions/comments that I'd like to understand before approving, but overall really great work!

localstack/aws/protocol/service_router.py
```diff
@@ -275,28 +276,32 @@ def get_service_catalog() -> ServiceCatalog:
...
+        # default sqs protocol (`query`).
         content_type = request.headers.get("Content-Type")
-        return "sqs" if content_type == "application/x-amz-json-1.0" else "sqs-query"
+        return "sqs-json" if content_type == "application/x-amz-json-1.0" else "sqs"
 
 
-def determine_aws_service_name(request: Request, services: ServiceCatalog = None) -> Optional[str]:
+def determine_aws_service_model(
+    request: Request, services: ServiceCatalog = None
+) -> Optional[ServiceModel]:
```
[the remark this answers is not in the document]
alexrashed: I see your point. In this case the function is not used anywhere else in the whole organization, which is why I would like to avoid using a deprecation path.

tests/unit/aws/protocol/test_serializer.py
```diff
@@ -476,7 +476,7 @@ def test_query_serializer_sqs_none_value_in_map():
 def test_query_protocol_error_serialization():
     exception = InvalidMessageContents("Exception message!")
     _botocore_error_serializer_integration_test(
-        "sqs-query", "SendMessage", exception, "InvalidMessageContents", 400, "Exception message!"
+        "sqs", "SendMessage", exception, "InvalidMessageContents", 400, "Exception message!"
     )
```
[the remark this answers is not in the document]
alexrashed: SQS is used a lot across the parser tests. In fact, the parser and serializer tests should be refactored and structured in the future, but I would like to avoid adding this to this PR, if it's okay for you.

localstack/aws/spec.py
```diff
@@ -70,6 +70,10 @@ def load_service(service: ServiceName, version: str = None, model_type="service-
     For example: load_service("sqs", "2012-11-05")
     """
     service_description = loader.load_service_model(service, model_type, version)
+    # if the service name is sqs-json, we just loaded our internalized SQS query spec,
```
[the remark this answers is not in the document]
alexrashed: Good catch!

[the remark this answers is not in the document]
alexrashed: This comment is fixed in the latest revision.

localstack/aws/protocol/service_router.py
```diff
@@ -275,28 +276,32 @@ def get_service_catalog() -> ServiceCatalog:
         return ServiceCatalog()
 
 
-def resolve_conflicts(candidates: Set[str], request: Request):
+def resolve_conflicts(candidates: Set[ServiceModel], request: Request):
```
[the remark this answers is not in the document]
alexrashed: It was hashable, yes, but it's a good point. The service router has now been restructured such that the different stages are working based on service model identifiers, and the main function loads and returns the model.
So in the latest revision, this specific parameter has been changed to a set of `ServiceIdentifier` (a named tuple containing the service name and the protocol).

localstack/aws/protocol/service_router.py
```diff
@@ -275,28 +276,32 @@ def get_service_catalog() -> ServiceCatalog:
...
-        # default in botocore starting 1.31.81.
+        # implementation. `sqs-json` represents the `json` protocol spec, and `sqs` the `query` protocol
+        # (default again in botocore starting with 1.32.6).
         # The `application/x-amz-json-1.0` header is mandatory for requests targeting SQS with the `json` protocol. We
-        # can safely route them to the `sqs` JSON parser/serializer. If not present, route the request to `query`.
+        # can safely route them to the `sqs-json` JSON parser/serializer. If not present, route the request to the
+        # default sqs protocol (`query`).
         content_type = request.headers.get("Content-Type")
-        return "sqs" if content_type == "application/x-amz-json-1.0" else "sqs-query"
+        return "sqs-json" if content_type == "application/x-amz-json-1.0" else "sqs"
```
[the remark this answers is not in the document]
alexrashed: Thanks for you input, this is how it's working now (we have a named tuple uniquely identifying a loadable service model), which is the return type of the conflict resolution / heuristic stage functions.

localstack/aws/spec.py
```diff
@@ -108,41 +112,48 @@ def service_names(self) -> List[ServiceName]:
         return list(self._services.keys())
 
     @cached_property
-    def target_prefix_index(self) -> Dict[str, List[ServiceName]]:
+    def target_prefix_index(self) -> Dict[str, List[ServiceModel]]:
```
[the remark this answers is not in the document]
alexrashed: This is a very good point, the model caching totally slipped my mind.
I changed the index such that it now only uses `ServiceModelIdentifier` (a `NamedTuple`).
This obviously increases the size of the service catalog pickle, but just by ~225kb instead of multiple megabytes: 383,9kb (old) compared to 608,6kb (new)
This results in the following building and loading times of the service catalog:

| Stage | Previous Index | New Index |
| -------- | -------------------- | -------------- |
|Building| 2.0 - 2.1s | 1.8 - 2.2s |
|Loading| 14 - 15ms | 27 - 40ms |

However, the variance was quite high for my tests (on my notebook).
But we can see that the building doesn't really take considerably longer, the loading takes double the time, but is only executed once per LocalStack run and it's an impact of max. 25ms.

localstack/aws/spec.py
alexrashed: This is a good point. The new solution does not use the `sqs-json` service anymore in the code.
However, using a custom service name has a big benefits: This way it integrates really nicely with `botocore` and our internal AWS client factory. We can just use `aws_client.sqs_json` to load a client which reads the spec we ship in the localstack module. This can be really helpful in the tests, and might also come in handy for other use cases. So we would like to have two ways to load a service with a specific protocol:
- Add the suffix to the service name (f.e. `sqs-json`), which enables the use case mentioned above.
- Add the protocol to the service loading function, to have a clean and clear way to define a specific protocol of a service to load.

I also tried to avoid patching `botocore`, but since the `spec#load_service` is the only function using the custom loader, which is why this custom logic (implementing the protocol-specific loading) is contained in that function.

thrau (approved):
fantastic job @alexrashed! thanks a lot for your patience and persistence on this long standing issue! i think we found a good solution for now, and it's exciting to see how we're really pushing the boundaries of botocore and what we need from it.

well done!

thrau:
merging the PR to get some data over the weekend. will keep an eye on it!

PR #9712, opened 2023-11-22
create s3 image and pipeline

bentsku:
## Motivation
Now that the new S3 provider is out and does not have any dependencies on Moto, we wanted to get an S3-only LocalStack image, which would try to be slimmed down as much as we could without major changes to the code base.

## Changes
- create a new `Dockerfile.s3` to build a slim LocalStack image with almost no dependencies
- create a new GH Actions pipeline to test the image when changes in the core framework or S3 are made, and to manually publish it via a workflow dispatch.
- Check for Moto import and try to scope them inside patch functions
- modify providers to import the `MotoFallbackDispatcher` inside the plugin, as some other providers were doing, and not a global import.
- remove an SQS old util not used anymore, which imported Moto in `sqs/utils` which is used in the service router, imported globally
- split the S3 utils between regular and Moto ones for the legacy provider
- do conditional imports for legacy S3 needs in `s3/notifications.py`
- move a fixture importing Moto into the specific test file needing them `test_inspect.py`
- add a workaround for `account_id`/`region` fixture (needed for snapshot testing) when `sts` is not enabled and in LocalStack. \cc @dominikschubert I would like to have your input on this
- add `skipif` for S3 tests for tests depending on other services

For the PR to be ready to push to the regular LocalStack repo with the `localstack/localstack:latest-s3` image name and tag, we just need to change the value of `DOCKERHUB_IMAGE_NAME` in `.github/workflows/tests-s3-image.yml`, and change pack the push to Dockerhub to be only when done manually.

## Limitations
The S3 image does not contain:
- the global awslocal installation inside the container
- the DNS server to gain some size, but this could be added back if there's a need (it's started with `DNS_ADDRESS=false`)

## Testing
This was a test run with pushing enabled to a test repository:
https://github.com/localstack/localstack/actions/runs/6972644488
The image is available with `docker pull localstack/localstack-s3-test:latest-s3`, build 23/11/23 around 18:30
And last test run:
https://github.com/localstack/localstack/actions/runs/6981817553?pr=9712
Image is available with `docker pull localstack/localstack:s3-latest`. 
https://hub.docker.com/r/localstack/localstack/tags

The -ext pipeline was green except for some flake on ARM. 
## TODO

What's left to do:

- [x] find a way to have node inside the S3 image during testing for the InfraProvisioner fixture, or a way to not load it
- [x] should we create a new docker-save-image make target with gzip?


localstack/services/s3/notifications.py
```diff
@@ -47,6 +42,22 @@
...
+        get_key_from_moto_bucket,
+    )
+
+try:
+    from localstack.aws.api.events import PutEventsRequestEntry
+    from localstack.aws.api.lambda_ import InvocationType
+except ImportError:
+    # if we can't import it, we are in the S3-only image, where we don't need those, so we can safely skip
+    pass
+
```
bentsku, line 60: todo: we now keep all API, so this is safe to remove

setup.cfg
```diff
@@ -87,11 +87,29 @@ runtime =
     Quart>=0.19.2
     readerwriterlock>=1.0.7
     requests-aws4auth>=1.0
-    # explicitly set urllib3 to force its usage / ensure compatibility
-    urllib3>=2.0.7
```
bentsku, line 91: oops, this is a mistake from rebasing, will add back

tests/unit/test_sqs.py
```diff
@@ -4,8 +4,8 @@
 import localstack.services.sqs.models
 from localstack.services.sqs import provider
 from localstack.services.sqs.constants import DEFAULT_MAXIMUM_MESSAGE_SIZE
+from localstack.services.sqs.provider import _create_message_attribute_hash
 from localstack.services.sqs.utils import (
-    get_message_attributes_md5,
```
bentsku, line 8: old legacy util, was not used anymore

localstack/services/s3/provider.py
```diff
@@ -189,6 +185,13 @@
 from localstack.utils.time import parse_timestamp
 from localstack.utils.urls import localstack_host
 
+if config.LEGACY_V2_S3_PROVIDER:
+    from localstack.services.s3.utils_moto import (
+        get_bucket_from_moto,
+        get_key_from_moto_bucket,
+        is_moto_key_expired,
+    )
+
```
bentsku, line 194: this might not be needed, as it would only be loaded if `config.LEGACY_V2_S3_PROVIDER` is True.

Dockerfile.s3
```diff
@@ -0,0 +1,154 @@
...
+RUN --mount=type=cache,target=/var/cache/apt \
+    apt-get update && \
+        # Install dependencies to add additional repos
+        apt-get install -y --no-install-recommends \
+            # Runtime packages (groff-base is necessary for AWS CLI help)
+             ca-certificates curl make openssl
+# do we need the dns server?
+#             ca-certificates curl make openssl iputils-ping iproute2
+# default under
+#             ca-certificates curl gnupg make openssl tar pixz zip unzip groff-base iputils-ping nss-passwords procps iproute2 xz-utils
```
bentsku, line 19: I wasn't sure if we should ship the DNS server, but maybe we should to enable virtual host. It shouldn't be too much work to enable it, but it might route everything to LS instead of just S3.

Dockerfile.s3
```diff
@@ -0,0 +1,154 @@
...
+
+# expose default environment
+# Set edge bind host so localstack can be reached by other containers
+# set library path and default LocalStack hostname
+ENV USER=localstack
+ENV PYTHONUNBUFFERED=1
+
+# Install the latest version of awslocal globally
+# RUN --mount=type=cache,target=/root/.cache \
+#     pip3 install --upgrade awscli awscli-local requests
```
bentsku, line 51: I commented out the globally installed awscli inside the container, as it's really heavy. Now that they compress the specs, maybe we could add it but maybe it could be opt-it?

Dockerfile.s3
```diff
@@ -0,0 +1,154 @@
...
+# delete the API specs stubs for other services
+# delete the providers implementations for other services
+# TODO: remove the SQS ones, it depends on 'is_sqs_queue_url' in service_router.py
+# TODO: remove lambda one, testutil.py depends on it for fixtures
+# TODO: we also need dynamodb, don't remove the API for now, until we find a way for tests...
+# and now we need node for CDK... what a mess omg
+# Maybe we need a bit of an heavier image for tests?
+# RUN find localstack/aws/api/ -mindepth 1 -maxdepth 1 -type d -not -name s3 -not -name sqs -not -name lambda_ -exec rm -rf '{}' \;
+    # \ && \
+    # find localstack/services/ -mindepth 1 -maxdepth 1 -type d -not -name s3 -exec rm -rf '{}' \;
```
bentsku, line 107: this was a big back and forth for saving some MB, deleting the generated specs, but the tests need them, and it only adds some MB to the image. For now, we can leave it in.

Dockerfile.s3
```diff
@@ -0,0 +1,154 @@
...
+
+# Generate the plugin entrypoints
+RUN make entrypoints
+
+# Remove dependecies
+# RUN apt-get remove -y make
+# TODO: we need make to run tests, maybe we could have 2 steps: one for tests and one for push?
+
+# remove the other services from the entrypoints / TODO: not needed anymore with SERVICES
+# RUN sed -i '/\[localstack\.aws\.provider\]/,/^$/ { /^\[localstack\.aws\.provider\]/ b; /^$/ b; /localstack.services.providers:s3/!d; }' ./localstack_core.egg-info/entry_points.txt
```
bentsku, line 117: this is not needed anymore thanks to the `SERVICES` variable, I'll remove

Dockerfile.s3
```diff
@@ -0,0 +1,129 @@
...
+    mkdir -p /var/lib/localstack && \
+    chmod -R 777 /var/lib/localstack && \
+    mkdir -p /usr/lib/localstack && \
+    mkdir /tmp/localstack && \
+    chmod -R 777 /tmp/localstack && \
+    touch /tmp/localstack/.marker
+
+# install basic (global) tools to final image
+RUN --mount=type=cache,target=/root/.cache \
+    pip install --no-cache-dir --upgrade virtualenv
```
alexrashed, line 32: Is this necessary in the image? This could be neglected, or moved down to the installation of the dependencies.

Dockerfile.s3
```diff
@@ -0,0 +1,129 @@
...
+ADD bin/localstack bin/localstack.bat bin/localstack-supervisor bin/
+
+# install dependencies to run the LocalStack Pro runtime and save which ones were installed
+RUN --mount=type=cache,target=/root/.cache \
+    make install-s3
+RUN . .venv/bin/activate && pip3 freeze -l > requirements-runtime.txt
+
+# delete the botocore specs for other services (>80mb)
+# TODO: well now it's compressed and it's much lighter: 20mb maybe not worth it
+RUN find .venv/lib/python3.11/site-packages/botocore/data/ -mindepth 1 -maxdepth 1 -type d -not -name s3 -exec rm -rf '{}' \;
```
alexrashed, line 75: I guess this is not great, but not terrible. If it doesn't cause any issues, we could leave this optimization in. But if it turns out that this could be problematic, I would vote for removing this (since it seems a bit hacky).

bentsku: I agree it's not great, but it saves a lot of space. It didn't create any issues yet, so I felt safe doing it, and I'm not seeing another way of doing this. Because if we don't do this, without the `SERVICES` variables, then this is basically a LocalStack image for all services fully implemented in LocalStack like SQS. Let's see then!

Dockerfile.s3
bentsku: Oh, good question. This might have been kept from the base `Dockerfile`, I will check. But quick question, why would this not be necessary in this image but in the default `localstack` image?

.github/workflows/tests-s3-image.yml
```diff
@@ -0,0 +1,201 @@
...
+
+      - name: Save the S3 image
+        run: docker save ${{ env.IMAGE_NAME }} | gzip > ${{ env.IMAGE_NAME }}.tar.gz
+
+      - name: Store Docker image as artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.IMAGE_NAME }}
+          path: ${{ env.IMAGE_NAME }}.tar.gz
+          retention-days: 1
```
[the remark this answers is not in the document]
bentsku: Okay, yes, I wasn't sure, I felt like usually we kept the image to be able to download it from the artifacts if we couldn't find why there was an issue. However, the S3 image is rather easy to build locally so might not be needed?

Makefile
```diff
@@ -171,10 +174,19 @@ docker-create-push-manifests:	## Create and push manifests for a docker image (d
 docker-run-tests:		  ## Initializes the test environment and runs the tests in a docker container
 	# Remove argparse and dataclasses to fix https://github.com/pytest-dev/pytest/issues/5594
 	# Note: running "install-test-only" below, to avoid pulling in [runtime] extras from transitive dependencies
-	docker run -e LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC=1 --entrypoint= -v `pwd`/tests/:/opt/code/localstack/tests/ -v `pwd`/target/:/opt/code/localstack/target/ -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/localstack:/var/lib/localstack \
+	docker run -e LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC=1 --entrypoint= -v `pwd`/tests/:/opt/code/localstack/tests/ -v `pwd`/localstack/testing/:/opt/code/localstack/localstack/testing/ -v `pwd`/target/:/opt/code/localstack/target/ -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/localstack:/var/lib/localstack \
```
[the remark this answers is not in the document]
bentsku: This might be a remnant from when I was running tests locally and might not have rebuilt the image... but I can't remember for sure. I will remove it again because I can't think of a reason it would still be needed.

localstack/services/s3/notifications.py
```diff
@@ -679,7 +683,7 @@ class EventBridgeNotifier(BaseNotifier):
     @staticmethod
     def _get_event_payload(
         ctx: S3EventNotificationContext, config_id: NotificationId = None
-    ) -> PutEventsRequestEntry:
+    ) -> "PutEventsRequestEntry":
```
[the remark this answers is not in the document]
bentsku: Good catch, it's a remnant of a conditional import I removed. thanks!

tests/aws/conftest.py
```diff
@@ -30,21 +27,27 @@ def pytest_runtestloop(session):
...
-    for test_class in test_classes:
-        # set flag that terraform will be used
-        if TestTerraform is test_class:
-            test_init_functions.add(TestTerraform.init_async)
-            continue
+    # set flag that terraform will be used
+    if "TestTerraform" in test_classes:
+        from tests.aws.test_terraform import TestTerraform
+
+        test_init_functions.add(TestTerraform.init_async)
```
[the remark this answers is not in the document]
bentsku: I'll give it a try again if this is really necessary. I might have been overzealous, all these changes might actually not be needed anymore, this seems like a remnant from deleting all services except S3.

setup.cfg
```diff
@@ -92,6 +92,26 @@ runtime =
...
+    hypercorn>=0.14.4
+    localstack-client>=2.0
+    pyopenssl>=23.0.0
+    readerwriterlock>=1.0.7
+    requests-aws4auth>=1.0
+    Werkzeug>=3.0.0
+    xmltodict>=0.13.0
+    # TODO: remove those 2 dependencies
+    flask>=3.0.0
+    Quart>=0.19.2
```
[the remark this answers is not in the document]
bentsku: only thing is that we're going to re-introduce the hard dependency on `Flask`, because in the default `runtime` we were counting on `Moto` to install it for us. But it's planned to remove, we can do it explicitly then?

tests/aws/conftest.py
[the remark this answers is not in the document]
bentsku: I reverted the `TestTerraform` changes, but left the conditional import as trying to import OpenSearch is failing due to missing python dependencies (`opensearch-py`)

Dockerfile.s3
alexrashed: To be honest, I think the installation of `virtualenv` is a remnant from older times. I don't think we use it anywhere anymore (I think we use the builtin `venv` module of Python 3 everywhere).

alexrashed (commented):
Great iteration! Once we have @dominikschubert signing off the PR, and clarified the last open questions, we should be good to go! 🚀

setup.cfg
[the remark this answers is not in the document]
alexrashed: I like the new naming (`base-runtime`), so this conversation is resolved.

setup.cfg
```diff
@@ -57,40 +57,47 @@ localstack =
...
+    pyopenssl>=23.0.0
+    readerwriterlock>=1.0.7
+    requests-aws4auth>=1.0
+    # explicitly set urllib3 to force its usage / ensure compatibility
+    urllib3>=2.0.7
+    Werkzeug>=3.0.0
+    xmltodict>=0.13.0
+    # TODO: remove those 2 dependencies
+    flask>=3.0.0
+    Quart>=0.19.2
```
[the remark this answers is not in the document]
alexrashed: FYI: I created a PR to cleanup the last remaining usages of Quart (https://github.com/localstack/localstack/pull/9754).

Dockerfile.s3
bentsku: Oh okay, great, I'll remove 👍

localstack/services/s3/notifications.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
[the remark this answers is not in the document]
bentsku: Because of the conditional imports of Moto classes, with it the types evaluation are postponed and will not fail. It's not great, but the other solution would be to split out the notifications into Moto/no Moto, or import inside the context creation, which is maybe better but incurs a little performance cost.

dominikschubert (approved):
Looked primarily over the test fixture related changes and they LGTM. We've recently discussed some alternatives like incorporating STS as a localstack-native service with s3.

For now I'm also in favor of the more pragmatic solution in this PR though :rocket:

alexrashed (approved):
This is awesome! Thanks for addressing the comments and bringing this over the line! 🚀 📦 💯

bentsku:
Thanks a lot @alexrashed for the thorough review(s) and @dominikschubert for the validation of the fixtures. I'll merge now. Thanks again!

PR #9713, opened 2023-11-23
s3: fix test_put_bucket_logging_wrong_target for multi-account and region 

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `test_put_bucket_logging_wrong_target` test should still create the consequent resources in this accounts and region.

## Changes
This PR:
- uses region specific `s3` clients for performing `PutBucketLogging` and `CreateBucket` operations. 
- adds regions transformers for both target and source origins. 
- updates snapshot.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -9214,12 +9214,29 @@ class TestS3BucketLogging:
         snapshot.match("get-bucket-logging", resp)
 
     @markers.aws.validated
-    def test_put_bucket_logging_wrong_target(self, aws_client, s3_create_bucket, snapshot):
+    def test_put_bucket_logging_wrong_target(
+        self,
+        aws_client,
+        aws_client_factory,
+        s3_create_bucket,
+        s3_create_bucket_with_client,
+        snapshot,
+    ):
         snapshot.add_transformer(snapshot.transform.key_value("TargetBucket"))
-        bucket_name = s3_create_bucket()
-        target_bucket = s3_create_bucket(
-            CreateBucketConfiguration={"LocationConstraint": "us-west-2"}
+
+        region_1 = "us-east-1"
+        region_2 = "us-west-2"
+
+        snapshot.add_transformer(RegexTransformer(region_1, "<region_1>"))
+        snapshot.add_transformer(RegexTransformer(region_2, "<region_2>"))
+
+        bucket_name = f"bucket-{short_uid()}"
+        client = aws_client_factory(region_name=region_1).s3
+        s3_create_bucket_with_client(
+            client,
+            Bucket=bucket_name,
         )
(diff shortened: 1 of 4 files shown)


tests/aws/services/s3/test_s3.py
```diff
@@ -9197,12 +9197,29 @@ def test_put_bucket_logging_accept_wrong_grants(self, aws_client, s3_create_buck
...
+        region_2 = "us-west-2"
+
+        snapshot.add_transformer(RegexTransformer(region_1, "<region_1>"))
+        snapshot.add_transformer(RegexTransformer(region_2, "<region_2>"))
+
+        bucket_name = f"bucket-{short_uid()}"
+        client = aws_client_factory(region_name=region_1).s3
+        s3_create_bucket_with_client(
+            client,
+            Bucket=bucket_name,
```
bentsku, line 9220: I believe we could greatly simplify the test by not using using `us-east-1` here, then we wouldn't need multiple clients (do we really need them?). Then we can use the default `aws_client.s3` everywhere in the tests. 
```suggestion
        region_1 = "us-east-2"
        region_2 = "us-west-2"

        snapshot.add_transformer(RegexTransformer(region_1, "<region_1>"))
        snapshot.add_transformer(RegexTransformer(region_2, "<region_2>"))

        bucket_name = s3_create_bucket(CreateBucketConfiguration={"LocationConstraint": region_1})
        target_bucket = s3_create_bucket(CreateBucketConfiguration={"LocationConstraint": region_2})
```

bentsku (commented):
Thanks for fixing the test! Region assumptions are always annoying in S3. I believe we could simplify the test while not using `aws_client_factory` and `s3_create_bucket_with_client` here, by not using `us-east-1` at all. Do you think that's reasonable?

tests/aws/services/s3/test_s3.py
sannya-singal: Yes @bentsku, we can do what you have suggested, I made the `us-east-1` client here because some tests are specifically written to test the behaviour against the default region.

bentsku: Yes I think this one is just logging across regions, no `us-east-1` specific behavior, so we could simplify it

bentsku: As validated with #9737, `us-east-1` is needed for the snapshot tests to work, so I'll approve the PR as is! Great investigation, thanks a lot @sannya-singal!

bentsku (approved):
After reviewing #9737, it shows that `us-east-1` is needed for the snapshot to work, so awesome work, this can be merged as is! LGTM

sannya-singal:
Ext integration tests pipeline- https://github.com/localstack/localstack-ext/actions/runs/7015036312

PR #9714, opened 2023-11-23
avoid botocore 1.32.6

alexrashed:
## Motivation
Yesterday we unpinned `botocore` again (with https://github.com/localstack/localstack/pull/9667).
This PR avoids the latest release of `botocore`, which - again - reverts the SQS protocol back to `query` (with https://github.com/boto/botocore/pull/3071).

## Changes
- Avoids the latest version of botocore again which switches the SQS protocol (once again) back to query.

## Next Steps
It feels like this is a safety measure for now, but the next try will be there soon.
We already have a solution at hand which is pretty hacky, but would get better with https://github.com/localstack/localstack/pull/9710.
The next steps therefore should be:
- Rebase this PR, finish the cleanup, and remove the pin.
- Switch the special handling from query to json and internalize the json SQS spec instead of the query spec.
  - This would allow us to use the latest version of botocore, but also provide both protocols.
- As soon as `botocore` switches this again, we have to switch again.

In the meantime we will start efforts to pin our dependencies and only do coordinated updates of `botocore` (coming in the next few weeks), such that these breaking changes can be handled in a coordinated manner.


bentsku (approved):
LGTM! Hope this stabilizes soon...

PR #9716, opened 2023-11-23
fix transfer.ListTagsForResource collect not implemented call

dfangl:
## Motivation
With a recent botocore update, we broke our collect-not-implemented operation in -ext.
Especially, the `transfer.ListTagsForResource` operation had issues, due to an arn pattern in the spec of `arn:\\S+`, which is not usable for generation.

## Changes
* Add pattern `arn:\\S+` to sanitation list.
* Remove empty leftover files of lambda cleanup (sneaked those in here)

### DIFF
--- a/localstack/aws/mocking.py
+++ b/localstack/aws/mocking.py
@@ -162,6 +162,7 @@ def sanitize_arn_pattern(pattern: str) -> str:
         "arn:aws.*:*",
         "^arn:aws.*",
         "^arn:.*",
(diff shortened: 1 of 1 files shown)


steffyP (approved):
Thanks so much for jumping onto this @dfangl 🙏 
Once the pipeline is green, it's ready to merge 🚀

alexrashed (approved):
Thanks for jumping on this! 🚀 🧹

PR #9717, opened 2023-11-23
fix: support for non-homogeneous lists

macnev2013:
## Motivation
This PR addresses an issue in the snapshot tests mechanism, specifically when dealing with dictionaries containing lists with non-homogeneous elements. An example scenario is shown below:

```python
d = {"key1": [{"key2": "value1"}, "value2", 3]}
snapshot.map("test_generic", d)
```

## Changes
The proposed changes address the issue by altering the approach to parsing dictionaries containing lists. Instead of assuming the data type based on the first element of the list, this update ensures that the data type of each element in the list is checked and handled appropriately.

## Testing

- added test case for testing out the supported element


steffyP (approved):
LGTM! thanks for the fix! 🚀

PR #9719, opened 2023-11-23
CW v2 Tests and small improvements 

pinzon:
## Changes
- [x] **Missing Tests + Test improvements**:
  - [x] add pagination tests 
  - [x] add snapshots to the existing tests
  - [x] remove the `time.sleep` from the tests and replace with `retry` or other utility
  
- [x] **put-metric-data** improvments (follow up #9475):
  - [x] verify that the default timestamp added to the metrics is UTC time (might be local timestamp?)
  - [x] Use query str instead of function call for more readable code
  - [x] add a test where we try to insert single + static values at the same time (in one put-metric-data statement (probably will throw an error, but we need to test)

- [x] **get-metric-data** improvements (follow up #9475):
     - [x] test default order (asc/desc timestamp for get-metric-data)
     - [x] handle empty filters for dimensions
     - [x] handle filters for unit, handle different units in the metrics
     - [x] ~check if we can~ simplify the logic for min/max/sum/avg the data query, by selecting everything in one query something like: `SELECT from (SELECT from simpletable ... UNION select from aggregatedtable...`)

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -295,7 +295,6 @@ class CloudwatchDatabase:
             return {
                 "id": query.get("Id"),
                 "datapoints": cleaned_datapoints,
-                "label": f"{metric.get('MetricName')} {stat}",
             }
 
     def list_metrics(
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -27,10 +27,10 @@ STAT_TO_SQLITE_COLUMNS = {
 }
 
 STAT_TO_SQLITE_FUNC_2 = {
-    "Sum": "SUM(value) ",
-    "Average": "SUM(value), SUM(count) ",
-    "Minimum": "MIN(value) ",
-    "Maximum": "MAX(value) ",
+    "Sum": "SUM(value)",
+    "Average": "SUM(value)",
+    "Minimum": "MIN(value)",
+    "Maximum": "MAX(value)",
     "SampleCount": "Sum(count)",
 }
 
@@ -207,20 +207,20 @@ class CloudwatchDatabase:
                 metric.get("Namespace"),
                 metric.get("MetricName"),
             )
-            #
-            dimension_filter = ""
-            for dimension in dimensions:
-                dimension_filter += "AND dimensions LIKE ? "
-                data = data + (f"%{dimension.get('Name')}={dimension.get('Value')}%",)
-
             unit_filter = ""
             if unit:
                 unit_filter = "AND unit LIKE ? "
-                data = data + (unit,)
+                data += (unit,)
+
+            dimension_filter = ""
+            for dimension in dimensions:
+                dimension_filter += "AND dimensions LIKE ? "
+                data = data + (f"%{dimension.get('Name')}={dimension.get('Value','')}%",)
 
             sql_query = f"""
             SELECT
-            {STAT_TO_SQLITE_FUNC_2[stat]}
+                {STAT_TO_SQLITE_FUNC_2[stat]},
+                SUM(count)
             FROM (
                 SELECT
                 value, 1 as count,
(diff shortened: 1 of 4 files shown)


steffyP (changes requested):
Nice work, thanks for the follow-up work and improving the tests 🚀 😄 

There are still a few test failures, also some tests that worked before are now failing for the v2 provider.

* please make sure that you rebase with the `feat/cloudwatch-v2` branch - it seems like you have two additional commits from the original branch, and one commit is missing (last one from last week, that fixed the pipeline)
* some of the new tests are failing for the old providers: make sure to add the skip-flag if needed
* some skips should probably only be added for the v2 provider (added comments in the code)
* I added several inline-comments, suggestions and questions, we can go through it in a 1:1 session if you like

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -163,10 +183,8 @@ def get_metric_data_stat(
         start_time: datetime,
         end_time: datetime,
         scan_by: str,
-    ):
+    ) -> dict:
```
silv-io, line 186: We could use a generic `Dict` type here instead of dict, since we know the typing of the keys and the values.

silv-io (commented):
LGTM,Steffy's input is probably more valuable :) 

Just have a small suggestion

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -215,35 +220,48 @@ def get_metric_data(
             if query_result.get("messages"):
                 messages.extend(query_result.get("messages"))
 
+            # TODO: Fix this. AWS sometimes returns the metric name as label, sometimes is metric with stat
```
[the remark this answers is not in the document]
steffyP: -> @baermat will work on that: adding a test and providing a fix

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -14,21 +14,26 @@
...
+    return os.environ.get("PROVIDER_OVERRIDE_CLOUDWATCH") != "v2" and not is_aws_cloud()
+
+
+def is_new_provider():
+    return os.environ.get("PROVIDER_OVERRIDE_CLOUDWATCH") == "v2"
 
 
 class TestCloudwatch:
     @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..Datapoints..Unit"], condition=is_new_provider)
```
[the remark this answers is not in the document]
steffyP: -> @baermat will add a test here :)

steffyP:
@pinzon: 
The failing unittest + integration tests relating to websocket import error should be fixed on master already (we don't have this commit yet on our feature branches). 
We can ignore this for now, merge this PR first, and then rebase the `feat/cloudwatch-v2` target branch. 

However, there is still one test failing for the v2 provider: `test_put_metric_uses_utc` - it seems like you used your local-time for testing? ;) Maybe you can try setting an explicit timezone?

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1659,49 +1717,39 @@ def test_get_metric_data_with_zero_and_labels(self, aws_client):
...
+            # keep one assert to avoid storing incorrect values
+            avg = [res for res in response["MetricDataResults"] if res["Id"] == "result_Average"][0]
+            assert [int(val) for val in avg["Values"]] == [19]
+            snapshot.match("get_metric_data_with_zero_and_labels", response)
+
+        retry(_match_results, retries=10, sleep=1.0)
 
     @markers.aws.validated
-    def test_get_metric_statistics(self, aws_client):
+    @markers.snapshot.skip_snapshot_verify(paths=["$..Datapoints..Unit"], condition=is_new_provider)
```
steffyP, line 1750: my bad: this test is failing with the old provider as well, we need the skip for both providers.

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1713,26 +1761,310 @@ def test_get_metric_statistics(self, aws_client):
...
+            )
+
+        def assert_metrics_count():
+            response = aws_client.cloudwatch.list_metrics(Namespace=namespace)
+            assert len(response["Metrics"]) == max_metrics and response.get("NextToken") is not None
+
+        retry(assert_metrics_count, retries=10, sleep=1.0, sleep_before=1.0)
+
+    @markers.aws.validated
+    def test_get_metric_data_pagination(self, aws_client):
```
steffyP, line 1804: this test is still with the old provider: please add a skip marker

steffyP (commented):
Thanks for all the updates and rework, @pinzon 🥳 
I still see two tests failing with the old provider, where we need to adapt/add markers for the test. 
Once we have this, it's good to merge!

PR #9721, opened 2023-11-24
[SFN] Enhancements to V2 Interpreter

MEPalma:
## Motivation
This pull request addresses some minor grammatical limitations in the current V2 interpreter implementation. The existing codebase lacked support for empty collection definitions in retry and catch statements, and it did not handle composite choice expressions with singleton values. Additionally, there were constraints on using States.Runtime as error names. This update rectifies these limitations, providing support for empty collections, enabling the use of composite choice expressions, and allowing for the inclusion of States.Runtime in error names, albeit without full handling capabilities (as expected). Furthermore, the exception handling has been refined to ensure consistent and informative parsing error reports.

## Changes
- Updated ASL grammar about catch_decl, retry_decl, choice_operator, and error names
- Removed bail parse strategy to allow dumping of all syntax error messages before raising the exception
- added relevant tests


dominikschubert (approved):
LGTM!

Really moving towards a new level of parity now.
Catching and fixing all those weird edge cases will probably keep us busy for some time :sweat_smile:

PR #9723, opened 2023-11-24
Fix for double-slash uri request

calvernaz:
## Motivation

The RAW_URI can contain a path with a double slash for example, for the request ` https://28e9f096.execute-api.localhost.localstack.cloud:4566//hello`, RAW_URI will be `//hello`. 

This will cause an issue parsing the url using the function `urlparse`. With this fix, we replace `//` with `/` to proceed with the parsing.

## Changes

Normalizes the RAW_URI by replacing the `//` with `/`. It does it in two places, on the `http` package and on the apigateway `router_asf`.

## Testing

Unit tests for the specific parsing function.

### DIFF
--- a/localstack/services/apigateway/context.py
+++ b/localstack/services/apigateway/context.py
@@ -27,6 +27,9 @@ class ApiInvocationContext:
     data: InvocationPayload
     headers: Dict[str, str]
 
+    # raw URI (including query string) retired from werkzeug "RAW_URI" environment variable
+    raw_uri: str
+
     # invocation context
     context: Dict[str, Any]
     # authentication info for this invocation
--- a/localstack/services/apigateway/router_asf.py
+++ b/localstack/services/apigateway/router_asf.py
@@ -13,6 +13,7 @@ from localstack.services.apigateway.context import ApiInvocationContext
 from localstack.services.apigateway.helpers import get_api_account_id_and_region
 from localstack.services.apigateway.invocations import invoke_rest_api_from_request
 from localstack.utils.aws.aws_responses import LambdaResponse
+from localstack.utils.strings import remove_leading_extra_slashes
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


localstack/http/request.py
```diff
@@ -196,7 +196,12 @@ def get_raw_path(request) -> str:
     if hasattr(request, "environ"):
         # werkzeug/flask request (already a string, and contains the query part)
         # we need to parse it, because the RAW_URI can contain a full URL if it is specified in the HTTP request
-        return urlparse(request.environ.get("RAW_URI", request.path)).path
+        request_uri = request.environ.get("RAW_URI")
+        if request_uri:
+            return urlparse(request_uri.replace("//", "/")).path
+        return request.path
```
thrau, line 202: @alexrashed could you sanity check this?

bentsku: I'd like to test that with S3 as well, it makes heavy use of `RAW_URI`

bentsku: (I think we have `test_resource_object_with_slashes_in_key` 
Which does requests like this which works:
`GET s3.localhost.localstack.cloud:4566/test-bucket-fa62efc6//bar/foo/`

Could update it to also run on virtual host to validate this format:
`GET test-bucket.s3.localhost.localstack.cloud:4566//bar/foo/`

But I think this PR is gonna make S3 tests fail and would break behavior.

bentsku: I have something up that I could thoroughly test with S3, should I open a PR? is it okay @calvernaz?

calvernaz: Yep, let's do that! Will remove the changes on the `http` package

bentsku: The other PR has been merged, should be good to go to rebase!

bentsku (approved):
LGTM! This is great because it does not change current behavior, but allows us to use the new `raw_uri` field. This is a really good solution! 

Just a small question in comment, but nothing major really. Thanks a lot for fixing this!

PR #9724, opened 2023-11-24
Remove HEADER_LOCALSTACK_ACCOUNT_ID

viren-nadkarni:
## Motivation

This PR removes the long deprecated `HEADER_LOCALSTACK_ACCOUNT_ID`. It was used by the old `aws_stack.connect_to_service()` client to pass the Account ID of origin caller. 

We have since moved to the new client `connect_to()` which is more intelligent with cross-account calls by passing the appropriate IAM credentials in the process.

## Changes

- Remove all use of `HEADER_LOCALSTACK_ACCOUNT_ID`
- Make the `mock_aws_request_headers()` use the current method of detecting internal calls
- Move the `mock_aws_request_headers()` to `localstack.utils.aws.request_context` to avoid a circular import. It also makes sense to group the AWS request related functions together - and this module already has `mock_request_for_region()`

### DIFF
--- a/localstack/utils/aws/aws_stack.py
+++ b/localstack/utils/aws/aws_stack.py
@@ -14,7 +14,6 @@ from localstack.constants import (
     APPLICATION_AMZ_JSON_1_1,
     APPLICATION_X_WWW_FORM_URLENCODED,
     AWS_REGION_US_EAST_1,
-    HEADER_LOCALSTACK_ACCOUNT_ID,
     LOCALHOST,
 )
 from localstack.utils.strings import is_string_or_bytes, to_str
@@ -157,9 +156,8 @@ def extract_access_key_id_from_auth_header(headers: Dict[str, str]) -> Optional[
             return access_id[0]
 
 
-# TODO remove the `internal` arg
 def mock_aws_request_headers(
-    service: str, aws_access_key_id: str, region_name: str, internal: bool = False
+    service: str, aws_access_key_id: str, region_name: str
 ) -> Dict[str, str]:
     """
     Returns a mock set of headers that resemble SigV4 signing method.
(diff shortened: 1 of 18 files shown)


localstack/utils/aws/aws_stack.py
```diff
@@ -155,36 +151,3 @@ def extract_access_key_id_from_auth_header(headers: Dict[str, str]) -> Optional[
...
-            "AWS4-HMAC-SHA256 "
-            + f"Credential={aws_access_key_id}/20160623/{region_name}/{service}/aws4_request, "
-            + "SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=1234"
-        ),
-    }
-    if internal:
-        # TODO: This method of detecting internal calls is no longer valid
-        # We now use the `INTERNAL_REQUEST_PARAMS_HEADER` header which is set to the DTO
-        headers[HEADER_LOCALSTACK_ACCOUNT_ID] = get_aws_account_id()
-    return headers
```
viren-nadkarni, line 190: This function is moved to a new module `localstack.utils.aws.headers`. This is to avoid a circular import but also to improve code organisation

localstack/utils/aws/request_context.py
```diff
@@ -213,3 +219,42 @@ def thread_run(fn, self, *args, **kwargs):
...
+        "Authorization": (
+            "AWS4-HMAC-SHA256 "
+            + f"Credential={aws_access_key_id}/20160623/{region_name}/{service}/aws4_request, "
+            + "SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=1234"
+        ),
+    }
+
+    if internal:
+        dto = InternalRequestParameters()
+        headers[INTERNAL_REQUEST_PARAMS_HEADER] = dump_dto(dto)
```
viren-nadkarni, line 258: This is changed to use the current method of constructing internal calls

bentsku (approved):
Changes look good to me, and it seems it's mostly used in tests, and everything is green so I believe this is good to go! 

Thanks a lot for addressing this, this is great to remove!

alexrashed (approved):
Thanks for the cleanup! 🧹

PR #9725, opened 2023-11-24
fix get_raw_path with double slash prefix path for legacy v-host S3

bentsku:
## Motivation
As raised by #9723, when supplying a path only starting with double slashes, `get_raw_path` does not parse it properly, it returns an empty `.path` property. 
If `RAW_URI` contains only a path, `get_raw_path` does need to call `urlparse`, we can manually strip the query string and return the path from it. 

## Changes
- manually return the path from `RAW_URI` if it starts with a double slashes
- add S3 tests in virtual host mode which validates the fix for the legacy provider (calling `get_raw_path` when rewriting the URL) 
- add a unit test


alexrashed (approved):
Clean fix, with great tests 👍🏽 💯

PR #9728, opened 2023-11-24
CFn: render ext providers with correct plugin namespace

simonrw:
## Motivation

While activating the pro resource providers in https://github.com/localstack/localstack/pull/9601 I neglected to read the `pro` flag from the command line arguments when rendering the plugin file. This means that we will get conflicting plugin namespaces which has an undefined load order. To get around this, we chose to add a new namespace for the pro plugins, and load the plugins in priority order (pro first). This feature has not worked correctly!

In the companion PR in the pro repo, I migrate the current plugins to this new namespace.

cc @Morijarti 

## Changes

* Read the `--pro` flag from the environment when rendering the plugin file


dominikschubert (approved):
LGTM, good catch

PR #9739, opened 2023-11-27
fix cloudwatch resource providers

pinzon:
## Motivation
The cloudwatch resource providers were found to be unusable. After researching it seems that the pipeline never failed for the migration PR due to the lack of tests.

## Changes
- Resource Providers for AWS::Cloudwatch::MetricAlarm and AWS::Cloudwatch::CompositeAlarm now considers more parameters and also considers that not all parameters are mandatory.

## Testing
- snapshoted test that validates the creation and  the deletion of  the resources
- snaphoted test for an alarm with no statistic

CC: @thrau @Morijarti


dominikschubert (approved):
LGTM, thanks also for adding the additional test :+1:

Issue #9534, opened 2023-11-02
enhancement request: Add support for Fn::GetAZs CloudFormation intrinsic function


lukasz-mazur2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Hello,

Localstack Pro user here.
We are in the process of adapting Localstack to our workflow/pipelines to reduce the load on our AWS accounts. In out organization we are heavily using CloudFormation and unified custom modules for IaC components. Some of the modules rely on the Fn::GetAZs CloudFormation intrinsic function. It would be a great help if it was supported in Localstack as well.

pinzon:
The feature has been merged. Please pull the latest image.

PR #9740, opened 2023-11-27
Enable test for Fn::GetAZs and extend feature

pinzon:
## Motivation
This CFn function was requested in #9534 but it seems that @viren-nadkarni  already developed the feature while working on the Multi-Region support.

## Changes
I extended the amount of AZs the function returns to match the amount AWS returns for `us-east-1`

## Testing
I un-skipped the preexisting test of this feature and added a snapshot

### DIFF
--- a/tests/aws/services/cloudformation/test_template_engine.py
+++ b/tests/aws/services/cloudformation/test_template_engine.py
@@ -235,20 +235,23 @@ class TestIntrinsicFunctions:
         assert deployed.outputs["Address"] == "10.0.0.0/24"
 
     @markers.aws.validated
-    @pytest.mark.skip(reason="function not currently supported")
-    def test_get_azs_function(self, deploy_cfn_template):
+    def test_get_azs_function(self, deploy_cfn_template, snapshot):
+        """
+        TODO parametrize this test.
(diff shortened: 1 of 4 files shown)


tests/aws/services/cloudformation/test_template_engine.py
```diff
@@ -235,20 +235,23 @@ def test_cidr_function(self, deploy_cfn_template):
...
 
     @markers.aws.validated
-    @pytest.mark.skip(reason="function not currently supported")
-    def test_get_azs_function(self, deploy_cfn_template):
+    def test_get_azs_function(self, deploy_cfn_template, snapshot):
+        """
+        TODO parametrize this test.
+        For that we need to be able to parametrize the client region. The docs show the we should be
+        able to put any region in the parameters but it doesn't work. It only accepts the same region from the client config
+        if you put anything else it just returns an empty list.
```
dominikschubert, line 243: :+1: 

I think it's fine for now since we usually assume us-east-1 anyway, i.e. we don't currently handle region specific idiosyncrasies and treat everything as if it were us-east-1, which has the broadest support out of all regions.

dominikschubert (approved):
LGTM

Issue #7699, opened 2023-02-16
bug: cdklocal destroy doesn't remove stack completely


vincenthongzy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Running `cdklocal destroy` to destroy a stack and re-running `cdklocal deploy` causes no changes in deployment. Checking with `awslocal cloudformation list-stacks --region ap-southeast-1` leads to:

```
        {
            "StackId": "arn:aws:cloudformation:ap-southeast-1:000000000000:stack/ExampleStack/d098b2f8",
            "StackName": "ExampleStack",
            "CreationTime": "2023-02-16T08:02:10.181000Z",
            "LastUpdatedTime": "2023-02-16T08:02:10.181000Z",
            "DeletionTime": "2023-02-16T08:02:44.247000Z",
            "StackStatus": "DELETE_COMPLETE",
            "DriftInformation": {
                "StackDriftStatus": "NOT_CHECKED"
            }
        },
```

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS: MacOS 13.2 
- LocalStack: latest
```

vincenthongzy:
Hi Dominik, 

After running `cdk destroy ExampleStack` , I was expecting `cdk deploy ExampleStack` to redeploy the stack. But instead from the logs I get:
```
✅  ExampleStack (no changes)
```
And the stack is not being deployed

PR #9748, opened 2023-11-28
Support creation of CloudFormation stacks with the same name after deletion

dominikschubert:
## Motivation

Tackling the issue where we couldn't  cycle between deleting and creating stacks with the same name. E.g. doing a CDK deployment, deleting the stack and then doing another deployment which will typically use the same name again.

Resolves https://github.com/localstack/localstack/issues/7699

## Changes

- Improved handling of stack ARN for `StackName` parameter in `CreateStack` and `CreateChangeSet`
- Creating multiple stacks with a single name are allowed now, as long as only one is an active state (i.e. not deleted).
  - this should work for both `CreateStack` and `CreateChangeSet` now
- Better handling of Stack in `REVIEW_IN_PROGRESS`
- Added tests covering a few scenarios involving deleted
- Fixed `test_stack_lifecycle` which was implicitly depending on a stack being in `DELETE_IN_PROGRESS` and when waiting for the deletion to be completed the test failed against AWS as well.

## TODO (not in this PR)

Some things still need to be done in the near future to improve detection of deleted stacks and in general the provider parity here.

- find a good abstraction for fetching the stack by name/arn/status after testing and fixing a few more operations
- we should cover more stack life cycle scenarios via tests but this isn't super urgent since only a few states are realistically relevant for use with LocalStack

### DIFF
--- a/localstack/services/cloudformation/provider.py
+++ b/localstack/services/cloudformation/provider.py
@@ -157,9 +157,6 @@ class InternalFailure(CommonServiceException):
 
 
 class CloudformationProvider(CloudformationApi):
-    def _get_stack_by_name_or_arn(self, context: RequestContext, stack_name_or_arn: str):
-        ...
-
     def _stack_status_is_active(self, stack_status: str) -> bool:
         return stack_status not in [StackStatus.DELETE_COMPLETE]
 
@@ -173,7 +170,7 @@ class CloudformationProvider(CloudformationApi):
         # get stacks by name
         active_stack_candidates = [
             s
-            for stack_arn, s in state.stacks.items()
+            for s in state.stacks.values()
             if s.stack_name == stack_name and self._stack_status_is_active(s.status)
         ]
 
--- a/localstack/services/cloudformation/provider.py
+++ b/localstack/services/cloudformation/provider.py
@@ -6,9 +6,11 @@ from copy import deepcopy
 
 from localstack.aws.api import CommonServiceException, RequestContext, handler
 from localstack.aws.api.cloudformation import (
+    AlreadyExistsException,
     CallAs,
     ChangeSetNameOrId,
     ChangeSetNotFoundException,
+    ChangeSetType,
     ClientRequestToken,
     CloudformationApi,
     CreateChangeSetInput,
@@ -61,6 +63,7 @@ from localstack.aws.api.cloudformation import (
     StackName,
     StackNameOrId,
     StackSetName,
+    StackStatus,
     StackStatusFilter,
     TemplateParameter,
     TemplateStage,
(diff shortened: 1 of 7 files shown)


pinzon (approved):
Pretty impressive :rocket:

tests/aws/services/cloudformation/api/test_stacks.py
```diff
@@ -51,11 +54,17 @@ def test_stack_lifecycle(self, is_stack_updated, deploy_cfn_template, snapshot,
...
+        aws_client.cloudformation.get_waiter("stack_delete_complete").wait(StackName=stack_name)
+
+        with pytest.raises(aws_client.cloudformation.exceptions.ClientError) as e:
+            aws_client.cloudformation.describe_stacks(StackName=stack_name)
+        snapshot.match("describe_deleted_by_name_exc", e.value.response)
+
+        deleted = aws_client.cloudformation.describe_stacks(StackName=deployed.stack_id)["Stacks"][
+            0
+        ]
+        assert "DeletionTime" in deleted
```
simonrw, line 66: nit: presumably this is redundant?

tests/aws/services/cloudformation/api/test_stacks.py
```diff
@@ -642,3 +651,94 @@ def test_blocked_stack_deletion(aws_client, cleanups, snapshot):
...
+MINIMAL_TEMPLATE_UPDATED = """
+Resources:
+    SimpleParam:
+        Type: AWS::SSM::Parameter
+        Properties:
+            Value: test2
+            Type: String
+"""
+
+
```
simonrw, line 674: nit: this template is not used

tests/aws/services/cloudformation/api/test_stacks.py
```diff
@@ -642,3 +651,94 @@ def test_blocked_stack_deletion(aws_client, cleanups, snapshot):
...
+@markers.snapshot.skip_snapshot_verify(
+    paths=["$..EnableTerminationProtection", "$..LastUpdatedTime"]
+)
+@markers.aws.validated
+def test_name_conflicts(aws_client, snapshot, cleanups):
```
simonrw, line 679: Nice, a good clear test outlining the functionality. Good to explore what the AWS functionality is here as well!

tests/aws/services/cloudformation/api/test_changesets.py
```diff
@@ -744,3 +747,251 @@ def test_autoexpand_capability_requirement(cleanups, aws_client):
...
+            ChangeSetName="revived-stack-changeset",
+            ChangeSetType="CREATE",
+            TemplateBody=MINIMAL_TEMPLATE,
+        )
+    snapshot.match("recreate_deleted_with_id_exception", e.value.response)
+
+
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        # gotta skip quite a lot unfortunately
```
simonrw, line 894: :(

localstack/services/cloudformation/provider.py
```diff
@@ -154,10 +157,36 @@ def __init__(self, message=None):
...
+        # TODO: fix/implement this code path
+        #   this needs more investigation how Cloudformation handles it (e.g. normal stack create or does it create a separate changeset?)
+        # REVIEW_IN_PROGRESS is another special status
+        # in this case existing changesets are set to obsolete and the stack is created
+        # review_stack_candidates = [s for s in stack_candidates if s.status == StackStatus.REVIEW_IN_PROGRESS]
+        # if review_stack_candidates:
+        # set changesets to obsolete
+        # for cs in review_stack_candidates[0].change_sets:
+        #     cs.execution_status = ExecutionStatus.OBSOLETE
+
```
simonrw, line 186: I think it's fine to leave this as a TODO for now. We already have much improved parity with this PR

simonrw (approved):
Nice work, I'm glad we have better behaviour around this as it's a nice UX improvement to be able to destroy the stack and recreate it rather than restarting LS every time. I shall remember this next time I work with a heavy CFn project!

localstack/services/cloudformation/provider.py
dominikschubert: Yeah  I think it's better to first remove the inheritance and rework the change set entity a bit before tackling this.

tests/aws/services/cloudformation/api/test_stacks.py
dominikschubert: Yep  good catch. Was going through a few iterations of the test cases, one also had an update with a different template after a successful create. I decided against this to keep the already kinda long test from bloating too much.

Issue #9742, opened 2023-11-27
bug: SFN - `States.StringSplit` errors when the same value in AWS succeeds


BrianMcBrayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the `States.StringSplit` intrinsic function for AWS Step Functions, I get the following error:

```
│ Error: creating Step Functions State Machine (repro): InvalidDefinition: Error=ValueError Args=["Expected 2 arguments for function type '<class 'localstack.services.stepfunctions.asl.component.intrinsic.function.statesfunction.string_operations.string_split.StringSplit'>', but got: '(FunctionArgumentList| {'arg_list': [], 'size': 0}'."] in definition '{
│   "Comment": "States.StringSplit error",
│   "StartAt": "WillStatesStringSplitError",
│   "States": {
│     "WillStatesStringSplitError": {
│       "Type": "Task",
│       "Resource": "arn:aws:states:::lambda:invoke",
│       "End": true,
│       "Parameters": {
│         "FunctionName": "SomeFunction",
│         "Payload": {
│           "inputArray.$": "States.StringSplit($$.Execution.Input.csv, ',')"
│         }
│       }
│     }
│   }
│ }
│ '.
│ 
│   with module.step_function.aws_sfn_state_machine.this[0],
│   on .terraform/modules/step_function/main.tf line 15, in resource "aws_sfn_state_machine" "this":
│   15: resource "aws_sfn_state_machine" "this" {
```

See repro for state machine definition that produces this error. The same repro works in AWS.

### Expected Behavior

I expect the step function to behave as intended and to split the CSV into an array.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

I'm using docker-compose with the following step function definition:

```json
{
  "Comment": "States.StringSplit error",
  "StartAt": "WillStatesStringSplitError",
  "States": {
    "WillStatesStringSplitError": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "End": true,
      "Parameters": {
        "FunctionName": "SomeFunction",
        "Payload": {
          "inputArray.$": "States.StringSplit($$.Execution.Input.csv, ',')"
        }
      }
    }
  }
}
```

### Environment

```markdown
- OS: MacOS Ventura
- LocalStack: latest
```

### Anything else?

Thank you for all your hard work!

MEPalma:
[...]

PR #9749, opened 2023-11-28
[SFN] Enhancements for Intrinsic Function StringSplit

MEPalma:
## Motivation
For the current implementation of the SFN interpreter, the Intrinsic Functions States.StringSplit does not filter for non-empty strings after computing the parts. This PR addresses such behaviour by filtering for empty string results.
Closes: #9742 

## Changes
Added empty string filtering after computing the split parts in the implementation for States.StringSplit, and a number of relevant test cases

### DIFF
--- a/localstack/services/stepfunctions/asl/component/intrinsic/function/statesfunction/string_operations/string_split.py
+++ b/localstack/services/stepfunctions/asl/component/intrinsic/function/statesfunction/string_operations/string_split.py
@@ -67,5 +67,5 @@ class StringSplit(StatesFunction):
         pattern = "|".join(patterns)
 
         parts = re.split(pattern, string)
-        parts_clean = list(filter(lambda sub: bool(sub), parts))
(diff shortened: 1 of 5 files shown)


localstack/services/stepfunctions/asl/component/intrinsic/function/statesfunction/string_operations/string_split.py
```diff
@@ -67,4 +67,5 @@ def _eval_body(self, env: Environment) -> None:
         pattern = "|".join(patterns)
 
         parts = re.split(pattern, string)
-        env.stack.append(parts)
+        parts_clean = list(filter(lambda sub: bool(sub), parts))
```
dominikschubert, line 70:
```suggestion
        parts_clean = list(filter(bool, parts))
```

small nit

dominikschubert (approved):
Small nit, otherwise LGTM!

:bowling:  Next one down

PR #9750, opened 2023-11-28
Remove thread-local region, access key and account ID

viren-nadkarni:
## Motivation

This PR is the culmination of months of refactoring efforts!

We are finally deprecating and removing `aws_stack.get_region()` and `get_aws_account_id()` (and few other functions that rely on the thread-local storage).

Thanks for patiently reviewing all the PRs folks! cc: @alexrashed @baermat @bentsku @dfangl @dominikschubert @giograno @joe4dev @MEPalma @steffyP @thrau @whummer

### DIFF
--- a/localstack/aws/accounts.py
+++ b/localstack/aws/accounts.py
@@ -3,17 +3,11 @@ import base64
 import binascii
 import logging
 import re
-import threading
-import warnings
 
 from localstack import config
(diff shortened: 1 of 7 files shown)


alexrashed (approved):
Awesome! Thanks so much, great to have this part finally being removed (which was a constant source of pain)! 🚀 🧹 💯

PR #9751, opened 2023-11-28
fix kms key creation for s3 encryption

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `test_s3_bucket_encryption_sse_kms_aws_managed_key` test should still create the consequent resources in the corresponding accounts and region.

## Changes
This PR adds `account_id` when creating kms key for s3 bucket encryption in `create_s3_kms_managed_key_for_region` and the remaining modifications involve propagating the `account_id` value to the relevant functions.

### DIFF
--- a/localstack/constants.py
+++ b/localstack/constants.py
@@ -150,14 +150,10 @@ DEFAULT_AWS_ACCOUNT_ID = "000000000000"
 # Credentials used in the test suite
 # These can be overridden if the tests are being run against AWS
 # If a structured access key ID is used, it must correspond to the account ID
-# TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
-# TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
-# TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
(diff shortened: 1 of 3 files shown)


bentsku:
This is a great fix, I don't think it needs additional tests. I totally overlooked the account part! I believe this could be merged like this, as this would prove very tricky to test. As soon as you mark it ready, I'll approve it. Thanks a lot for fixing this bug! 

> If you don't specify a customer managed key, Amazon S3 automatically creates an AWS managed key in your AWS account the first time that you add an object encrypted with SSE-KMS to a bucket. By default, Amazon S3 uses this KMS key for SSE-KMS.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#aws-managed-customer-managed-keys

And very hard to test the cross-account part:
> Objects encrypted using SSE-KMS with [AWS managed keys](http://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#aws-managed-cmk) can't be shared cross-account. If you need to share SSE-KMS data cross-account, you must use a [customer managed key](http://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#customer-cmk) from AWS KMS.

sannya-singal:
>This is a great fix, I don't think it needs additional tests. I totally overlooked the account part! I believe this could be merged like this, as this would prove very tricky to test. As soon as you mark it ready, I'll approve it. Thanks a lot for fixing this bug!

Thanks @bentsku 🚀 🎉

> From this, it looks like the KMS key will live in the bucket owner account, like you did. We might need to add a check in the future to prevent cross-account access of the managed KMS key

Agreed! Thanks for pointing this out 🙏

macnev2013 (approved):
LGTM 🚀

PR #9754, opened 2023-11-28
remove remaining (transitive) usages of Quart

alexrashed:
## Motivation
During the review of https://github.com/localstack/localstack/pull/9712, it was made explicit that we still have a dependency on `Quart` (even though we switched to the underlying `hypercorn` as ASGI server quite a while back).

This PR removes the last remaining references to `Quart` and removes the dependency from the `setup.cfg`.

## Changes
- Removes all remaining references to Quart.
- Migrates some tests using Quart as a mock server to use the `httpserver` fixture.
- Remove the old function providing a `Quart` server (which was only used in tests lately).
- Cleanup and make the HTTP server patching explicit (even though it's still not too obvious, but now it's at least explicitly called instead of implicitly when loading the module _somewhere_).

## Testing
- All tests have been migrated to the `httpserver` fixture, so this PR basically just cleans up some remnants of the big v3 cleanup.
  - If the tests are 💚, this change should be safe for `localstack/localstack` and `localstack/localstack-pro`.
- However, there is a dependency on [`run_server` (which is removed in the companion PR) in `auth_proxy.py` in the AWS Replicator Extension](https://github.com/localstack/localstack-extensions/blob/main/aws-replicator/aws_replicator/client/auth_proxy.py).
  - This is the only usage of `run_server` (at the time of writing this) in any other repo of the organization on GitHub.

## TODO
- [x] Make sure the code is not used in the replicator extension anymore before merging this PR.
  - Tracked in https://github.com/localstack/localstack-extensions/issues/45.
  - The code has actually been moved to the extension with https://github.com/localstack/localstack-extensions/pull/64.


thrau:
I rebased the PR, had to do a few changes, but should be good to go now! @alexrashed

alexrashed:
Thanks a lot, @thrau for moving the legacy code to the extension with https://github.com/localstack/localstack-extensions/pull/64 and rebasing this PR! 🦸🏽 
From my point of view it's looking good, however I can't review my own PR 😛 
Happy for any reviews / approvals, afterwards - and after https://github.com/localstack/localstack-extensions/pull/64 is merged - we can finally merge this! 🥳

thrau (approved):
thanks so much for driving to clean this up! happy we're giving way to the modern frameworks we have :partying_face:

PR #9760, opened 2023-11-28
Update Lambda runtime test skip condition

simonrw:
## Motivation

While attempting to run the Lambda runtime tests using the kubernetes executor against a k8s cluster (https://github.com/localstack/ls-on-k8s-testing/pull/1) we found that the Lambda runtime tests were being skipped. The skip condition was checking for a system architecture of `x86_64` but the helper method used (`localstack.utils.platform.get_arch`) standardises the name to return `amd64` meaning that even on a supported test architecture, the tests are being skipped.

## Changes

* Update the skip condition to allow for test execution on x86_64

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_common.py
+++ b/tests/aws/services/lambda_/test_lambda_common.py
@@ -47,7 +47,7 @@ def snapshot_transformers(snapshot):
 
 
 @pytest.mark.skipif(
-    condition=get_arch() != "x86_64", reason="build process doesn't support arm64 right now"
+    condition=get_arch() != "amd64", reason="build process doesn't support arm64 right now"
 )
 @markers.lambda_runtime_update
 class TestLambdaRuntimesCommon:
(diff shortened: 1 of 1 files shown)


tests/aws/services/lambda_/test_lambda_common.py
```diff
@@ -47,7 +47,7 @@ def snapshot_transformers(snapshot):
 
 
 @pytest.mark.skipif(
-    condition=get_arch() != "x86_64", reason="build process doesn't support arm64 right now"
+    condition=get_arch() != "amd64", reason="build process doesn't support arm64 right now"
```
joe4dev, line 50: Can we fix the additional occurrence at line `251` as well?

joe4dev, line 50: nit: Should we consider the typed `Arch.amd64`?

joe4dev (approved):
Great catch 🙀🙀🙀 🥲 💯 

Maybe fix the other occurrence as well, otherwise LGTM 👍

tests/aws/services/lambda_/test_lambda_common.py
joe4dev, line 50: I recently talked to @dfangl that we should finally fix the platform detection properly. I'm not sure whether that skip was/is related to the CI infrastructure because I am able to build and run these tests on ARM (after uncommenting that annoying skip ;)

PR #9761, opened 2023-11-28
Add image sha to startup information and more details to bug report template

dominikschubert:
## Motivation

Going through issues it's often quite hard to reproduce issues, especially when a bug report simply specifies that it was run with `localstack/localstack:latest`. Even with the date of the bug report this is not reliable information since the issue reporter might not have pulled a new version right before the issue was opened.
This PR intends to improve the bug report template a tiny bit by giving us better information about the version of localstack used.

## Changes

- Add another line to the startup logs of LocalStack, e.g. `LocalStack Docker image sha: sha256:f7c3a25c88a7888fafe5029069fc7d093a888abd0bdd9c4303ead58e90d8bc8a`
- Remove `latest` from the bug report template and instead refer to this startup output.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -60,10 +60,11 @@ body:
         - **LocalStack**:
           You can find this information in the logs when starting localstack
 
-          LocalStack version: 2.3.2
-          LocalStack Docker image sha: sha256:f7c3a25c88a7888fafe5029069fc7d093a888abd0bdd9c4303ead58e90d8bc8a
-          LocalStack build date: 2023-10-03
-          LocalStack build git hash: 1bccf790
+          LocalStack version: 3.4.1.dev
+          LocalStack Docker image sha: sha256:f02ab8ef73f66b0ab26bb3d24a165e1066a714355f79a42bf8aa1a336d5722e7
+          LocalStack build date: 2024-05-14
+          LocalStack build git hash: ecd7dc879
+
     value: |
         - OS:
         - LocalStack:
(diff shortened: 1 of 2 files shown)


dfangl (approved):
I agree with logging the error in case of a container exception, but other than that LGTM!

dominikschubert:
@alexrashed  I assume this should be covered by now with the env vars we have on images like this example of `docker image inspect localstack/localstack:3.4.0` or do you think we should still add those as labels as well? :thinking: 


```json
{
...
                "LOCALSTACK_BUILD_DATE=2024-04-25",
                "LOCALSTACK_BUILD_GIT_HASH=6f971ac81",
                "LOCALSTACK_BUILD_VERSION=3.4.0"
...
}
```

alexrashed:
I think there might be a misconception. The hash here is the hash of the git commit of the OSS version. I think there would still be value in adding the _image_ hash, since this would allow us to directly pull the exact same image as the user.
Or am I missing something? 🤔

alexrashed (approved):
As mentioned in my comment above, I think this would be super-useful. :)
The code LGTM! 💯

dominikschubert:
> I think there might be a misconception. The hash here is the hash of the git commit of the OSS version. I think there would still be value in adding the _image_ hash, since this would allow us to directly pull the exact same image as the user. Or am I missing something? 🤔

@alexrashed  the image hash is already included and yes that was the main motivation behind the whole PR :grin:  Or maybe I'm now missing something :grimacing: 

```
            image_sha = inspect_result["Image"]
            print("LocalStack Docker image sha: %s" % image_sha)
```


I was mostly asking about your previous comment regarding the image **labels** since I'm not sure about the whole context around that anymore

alexrashed:
Ah, right, yeah, the label's shoudn't be necessary if we can easily extract the set environment variables from the image without creating a container from it! Sorry for the confusion! 🚀

Issue #9658, opened 2023-11-16
bug: Provisioned Lambda instances do not get used - a new Lambda instance is created instead.


RowlandBanks:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have deployed a Lambda in LocalStack 2.3.2 and configured Provisioned Concurrency to have 2 instances. 

LocalStack correctly starts up 2 instances.

I invoke the Lambda and LocalStack spawns a new Lambda instance and uses the new instance for processing.

### Expected Behavior

I expect LocalStack to use one of the existing provisioned instances, rather than spawning a new instance.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Create the following files in a single folder, then run the commands below.

Dockerfile:
```dockerfile
# Build a lambda function in C# that will write to the console.
FROM mcr.microsoft.com/dotnet/sdk:6.0 as dotnet

RUN apt update
RUN apt install -y zip

WORKDIR /app
COPY Lambda.csproj Lambda.csproj
COPY Program.cs Program.cs
RUN dotnet publish Lambda.csproj -c Release --output publish

RUN cd publish && \
    zip -r /app/publish/deployment-archive.zip . && \
    cd ..

# Now copy the deployment archive containing the C# application to a
# new image based on LocalStack.
FROM localstack/localstack:2.3.2 AS localstack
RUN mkdir /app
COPY --from=dotnet /app/publish/deployment-archive.zip /app/deployment-archive.zip
```

docker-compose.yml
```yml
version: "3.9"
services:
  aws:
    build:
      context: .
      dockerfile: Dockerfile
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
    expose: # Expose ports to other containers running in the same compose instance.
      - "4566"
    networks:
      - docker_network
    environment:
      - DEBUG=1
      - LS_LOG=trace
      - LAMBDA_DOCKER_NETWORK=my_docker_network
      - DOCKER_HOST=unix:///var/run/docker.sock
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./docker/output/volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
      - .:/app/code

networks:
  docker_network:
    name: my_docker_network
    # We have to name this network, to permit the `LAMBDA_DOCKER_NETWORK` env var to be set.
[...]

PR #9762, opened 2023-11-28
Test lambda provisioned concurrency scheduling

joe4dev:
## Motivation

Lambda provisioned concurrency misses a test for synchronous invokes with spillover from provisioned concurrency into on-demand instances. Related to the issue report https://github.com/localstack/localstack/issues/9658 (which is not a bug but highlights some testing deficit).

Why not:

* The synchronous test `test_provisioned_concurrency` already covers the prefer-provisioned-concurrency scenario but misses spillover
* The asynchronous test `test_reserved_concurrency_async_queue` covers a very similar scenario using async events

## Changes

* Add a dedicated test for synchronous invokes with spillover from provisioned-concurrency into on-demand concurrency.
* Enable a previously skipped concurrency type assertion (now implemented)
* Fix reserved concurrency update exception when a reserved concurrency config already exists for the same function
* Add concurrency quota checks for all relevant concurrency tests

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -2535,8 +2535,8 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
             "UnreservedConcurrentExecutions"
         ]
         if (
-            provisioned_concurrent_executions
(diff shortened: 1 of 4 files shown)


dominikschubert (approved):
LGTM, the level or parity in the lambda provider these days is really cool to see :smile:

PR #9763, opened 2023-11-28
add APIGW PutIntegration IntegrationType validation

bentsku:
## Motivation
We had a support case where the user was having issue with the `HTTPS_PROXY` integration type. After some investigation, it looked like it didn't exist in AWS either. We need validation early so that it would fail when setting up infrastructure and not when actually calling the `execute-api` endpoint.

## Changes
Add validation for `IntegrationType` in `PutIntegration`, and write a small snapshot test.

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -159,6 +159,14 @@ STAGE_UPDATE_PATHS = [
     "/tracingEnabled",
 ]
 
+VALID_INTEGRATION_TYPES = {
+    IntegrationType.AWS,
+    IntegrationType.AWS_PROXY,
+    IntegrationType.HTTP,
+    IntegrationType.HTTP_PROXY,
+    IntegrationType.MOCK,
+}
+
 
 class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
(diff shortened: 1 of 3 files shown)


localstack/services/apigateway/provider.py
```diff
@@ -1751,7 +1759,15 @@ def get_integration(
     def put_integration(
         self, context: RequestContext, request: PutIntegrationRequest
     ) -> Integration:
-        if request.get("type") == IntegrationType.AWS_PROXY:
+        if (integration_type := request.get("type")) not in VALID_INTEGRATION_TYPES:
```
calvernaz, line 1762: interesting...we don't have many negative tests around this - I wonder if some aws sdks validate this before it actually makes the request.

calvernaz (approved):
LGTM, thanks for tackling this 👍

localstack/services/apigateway/provider.py
bentsku: Yes, for example boto validates and does not let you send the request. But I suppose some SDK are more lenient. We've often relied on SDK validation to do the work for us, but now can see we even need to disable boto validation to be able to test real AWS behavior...

PR #9764, opened 2023-11-28
Spike: track resource management in tests

simonrw:
## Motivation

The stability of the CI test pipeline is always a priority for the LocalStack project. It is important to have a reliable pipeline to ensure product quality. Unfortunately we have variability in the pipeline that makes it hard to rely on.

We have had a few cases where leaky resources in CI runs have caused test timeouts. It would be good to have a way to detect these.

## Changes

This PR introduces a method of capturing API calls during test runs. The pytest plugin captures information about the test, and marks the start and end points of the test. The handler itself records any API calls made into LocalStack.

We define a mapping of `(service, operation)` combinations that define _create_ and _delete_ operations. At the end of each test we compute the sum of these operations (assigning `+1` to create calls and `-1` to negative calls). Any tests that indicate a positive score is presented at the end of the test run. At the moment, any negative score is also presented, even though this in principle is not a problem: too may deletes (except CloudFormation where multiple deletes are common).


dominikschubert (commented):
Great to see another push towards more stability in our pipeline. We've been experimenting with this in the past a lot but those things never really left their prototype stage. Given that we're now also again prioritizing pipeline stability and as a result of that also test quality, it makes a lot of sense to spend more effort on this! :+1: 


I have a few thoughts about the concrete proposal though:

- The `METHOD_PAIRS` dict seems to be a sort of smaller/simpler version of the previously prototyped "automatic client-based resource cleanup", which makes me think it might be better to spend the effort there since it will have other immediate benefits as well.

- Trying to capture any leaks makes sense even if we had automated cleanup, but I don't think this should be based on create + delete calls. It should probably at least actually be based on a sort of List or Get with some form of validation to make sure the resource is actually gone. When focusing only on LocalStack it might make more sense to simply use our Stores for this and check for entities in there. 

- This being part of the handler chain and not part of the client instrumentation, means it will have zero effect against AWS which reduces its usefulness a bit (but would still be useful for the full CI run ofc).

PR #9765, opened 2023-11-29
fix type hints incompatible with python < 3.10 in aws_stack

viren-nadkarni:
## Motivation

Use of subscriptable types broke CLI tests in for older Python version, specifically the **cli-tests (3.8)** job.

This PR makes use of backward-compatible type hints.

### DIFF
--- a/localstack/utils/aws/aws_stack.py
+++ b/localstack/utils/aws/aws_stack.py
@@ -98,7 +98,7 @@ def get_s3_hostname():
 
 
 def fix_account_id_in_arns(
-    response, replacement: str, colon_delimiter: str = ":", existing: str | List[str] = None
+    response, replacement: str, colon_delimiter: str = ":", existing: Union[str, List[str]] = None
 ):
     """Fix the account ID in the ARNs returned in the given Flask response or string"""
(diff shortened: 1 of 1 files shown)


localstack/utils/aws/aws_stack.py
```diff
@@ -98,7 +98,7 @@ def get_s3_hostname():
 
 
 def fix_account_id_in_arns(
-    response, replacement: str, colon_delimiter: str = ":", existing: str | list[str] = None
+    response, replacement: str, colon_delimiter: str = ":", existing: str | List[str] = None
```
alexrashed, line 101: The type operator for union types was introduced with 3.10, so this won't fix the issue I'm afraid...

alexrashed (changes requested):

localstack/utils/aws/aws_stack.py
viren-nadkarni: The CLI Tests / cli-tests (3.8) job seems to be passing but I've gone ahead and changed this in 5f755f268

alexrashed (approved):
Thanks for jumping on this! 🚀 🧹 💚

PR #9767, opened 2023-11-29
pin global awscli install in Dockerfile

bentsku:
## Motivation
We should make sure the global awscli install is the same version as the one we have in our virtual environment. 

## Changes
Pin the `awscli` version to be the same as the one is installed. We should make sure to update it with the botocore update.


alexrashed (approved):
Great catch! We should really try to avoid having two environments in the image, but unfortunately, that would be a breaking change.
We just need to make sure that we remove this pin with https://github.com/localstack/localstack/pull/9710.

dominikschubert (approved):

PR #9772, opened 2023-11-29
Add formatting-aware timestamp replacement via new TimestampTransformer

dominikschubert:
## Motivation
In the past we've had issues with missing parity of timestamp formats due to the default transformer replacing all timestamps with the same generic tokens. This is an attempt to improve on this by encoding the date format. It's intentionally not trying to be too smart by doing too much magic(e.g. trying to parse & infer formatting) and needs to be explicitly extended by regex +  replacement.

I really like how Golang does date formatting, so in that style, the replacements are using a reference timestamp: `2022-07-13T13:48:01Z` which is the UTC timestamp of the v1.0.0 commit of localstack/localstack: https://github.com/localstack/localstack/commit/cf26bd9199354a9a55e0b65e312ceee4c407f6c0

## Changes

- New TimestampTransformer
  - replacement with `<datetime>` for anything that boto deserializes as a `datetime.datetime` (typically epoch numbers in the API)
  - replacement with `<timestamp:...>` for serialized timestamps, e.g. `"2022-07-13T13:48:01.000000+00:00"` => `"<timestamp:2022-07-13T13:48:01.000000+00:00>"`
- Added an exemption list that matches based on substrings of the executed test's file path. Anything that matches this list will keep using the old transformer, while anything else is now using the new TimestampTransformer by default.
  - This means that it's fairly easy to incrementally migrate/regenerate snapshots and avoids a big-bang merge with a lot of merge conflicts. 

## Examples

Example from a currently failing run:
```
_____________ TestS3PresignedUrl.test_s3_presigned_url_expired[s3] _____________
>> match key: expired-exception
	(~) /Error/Expires 'date' → '<timestamp:2022-07-13T13:48:01.000Z>' ... (expected → actual)
	(~) /Error/ServerTime 'date' → '<timestamp:2022-07-13T13:48:01.000Z>' ... (expected → actual)

```

## Follow-Ups

What's left to do:

- [ ] add follow-up PR that removes the exemptions (stays open until both community & -ext are migrated)
- [ ] adjust snapshots in -ext
- [ ] adjust snapshots in community


localstack/testing/pytest/snapshot.py
```diff
@@ -122,7 +125,41 @@ def _snapshot_session(request: SubRequest, account_id, region):
...
+        "tests/aws/services/sns",
+        "tests/aws/services/stepfunctions",
+        "tests/aws/services/sqs",
+        "tests/aws/services/transcribe",
+        "tests/aws/scenario/bookstore",
+        "tests/aws/scenario/note_taking",
+        "tests/aws/scenario/lambda_destination",
+        "tests/aws/scenario/loan_broker",
+        "localstack_ext",
+        "localstack-ext",
```
steffyP, line 157: for localstack-ext would need to further break this down so we don't have to do a big-bang approach?

steffyP (approved):
Great work 🚀 👍 Love that we can now check the actual expected date format 👏 

> the replacements are using a reference timestamp: 2022-07-13T13:48:01Z which is the UTC timestamp of the v1.0.0 commit of localstack/localstack

nice, I remember we once talked about this approach 😄 

> This means that it's fairly easy to incrementally migrate/regenerate snapshots and avoids a big-bang merge with a lot of merge conflicts.

👍 🙇‍♀️

localstack/testing/pytest/snapshot.py
dominikschubert: yup, can be done in another PR though. Let's first try to get community migrated :+1:

PR #9775, opened 2023-11-29
Add option to capture raw snapshots when running tests

dominikschubert:
## Motivation

A useful util for debugging snapshots. RAW snapshots are the RAW untransformed recorded content . In contrast to normal snapshots these should not be committed (see gitignore) and should only be used for debugging & exploration purposes.

## Changes

- Introduces **RAW snapshot mode**: Set `SNAPSHOT_RAW=1` or use the CLI argument `--snapshot-raw` to enable this mode. When enabled it will save a separate JSON file next to the normal snapshots with the "raw", i.e. untransformed recorded contents.

## Testing

Set `SNAPSHOT_RAW=1` and execute a snapshot test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ gen/
 
 # hypothesis pytest plugin
 .hypothesis
+
+# RAW snapshots
+*.raw.snapshot.json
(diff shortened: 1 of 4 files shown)


bentsku (commented):
This is great! I think this is a nice addition that can be really helpful when debugging snapshots and trying to understand why some parts are different, or why some transformers are not behaving like we expect them too.

It still marked as POC, but I'd up for merging this as is, as it's only opt-in and a nice tool. 

It's also very nice that it keeps the `ResponseMetadata`, as sometimes some headers are not parsed/in the specs so we can quickly see if boto is missing something.

I now see what you meant with saving raw LocalStack snapshots too, to be able to full compare them. Maybe this would be a good addition 🤔

bentsku (approved):

Issue #9727, opened 2023-11-24
bug: startup freeze when running localstack in Docker mode on MacOS


maslokeeper01:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running `localstack start -d`, program freezes on "configuring container" phase and won't pass through that phase.

<img width="1050" alt="image" src="https://github.com/localstack/localstack/assets/61380431/0c9b8646-269c-4bda-a0bd-6d6dc54ff54d">

### Expected Behavior

System should start without freezing.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

Run localstack in Docker mode with command `localstack start -d` on the MacOS platform.

### Environment

```markdown
- OS: MacOS Monterey v12.6.8
- LocalStack: latest
```

simonrw:
Hi @maslokeeper01, thanks for reaching out to us. Could you try running LocalStack setting the environment variable `DEBUG=1` and paste the output please? This will help us understand what the problem is.

PR #9778, opened 2023-11-30
Fix missing "Pulling container image" message from CLI startup

simonrw:
## Motivation

When starting LocalStack in Docker mode with the CLI, if the Docker image is not present we used to print "Pulling container image" to indicate that the Docker image was being pulled.

Since #9049, we instead start the DNS server first. To do this, we check if we can publish port 53 to the host, implemented via `container_ports_can_be_bound`. This in turn tries to run a docker image with the publish flag `-p 53`:

https://github.com/localstack/localstack/blob/c3348f842a875a9f7ba025ecaebb5f4bbdaac896/localstack/utils/docker_utils.py#L150-L156

We use the LocalStack docker image to check for port publishing, so we end up pulling the Docker image (silently) first before calling `ensure_container_image` which prints the "Pulling container image" message.

Possibly related: https://github.com/localstack/localstack/issues/9727

## Changes

* Ensure we have the container image present before testing for port bindability
* Add a CLI test to capture the behaviour

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -1165,10 +1165,10 @@ def start_infra_in_docker(console, cli_params: Dict[str, Any] = None):
     # create and prepare container
     container_config = ContainerConfiguration(get_docker_image_to_start())
     container = Container(container_config)
+    ensure_container_image(console, container)
 
     configure_container(container)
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Thanks for jumping on this! This will also improve the user experience, because it uses the same image for the port check as it's using later to start! 🚀 ⌛

PR #9784, opened 2023-12-01
events: add account_id while creating client in sending events to target

sannya-singal:
## Motivation
When using values other than `000000000000` for account id or `us-east-1` for region, `events` tests should still create the consequent resources in the corresponding accounts and region.

## Changes
This PR introduces the use of `account_id` field fetched from `target_arn` in order to connect with the correct client in `connect_to()` method while sending events to the target.

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`). The affected tests are:

```
tests.aws.services.events.test_events.TestEvents.test_put_events_into_event_bus[standard]
tests.aws.services.events.test_events.TestEvents.test_put_events_into_event_bus[domain]
tests.aws.services.events.test_events.TestEvents.test_put_events_into_event_bus[path]
tests.aws.services.events.test_events.TestEvents.test_put_events_with_target_firehose
tests.aws.services.events.test_events.TestEvents.test_put_events_with_target_kinesis
tests.aws.services.cloudformation.resources.test_events.test_event_rule_to_logs
```

### DIFF
--- a/localstack/utils/aws/message_forwarding.py
+++ b/localstack/utils/aws/message_forwarding.py
@@ -47,7 +47,7 @@ def send_event_to_target(
             role_arn=role, service_principal=source_service, region_name=region
         )
     else:
-        clients = connect_to(region_name=region)
+        clients = connect_to(aws_access_key_id=event.get("account"), region_name=region)
 
     if ":lambda:" in target_arn:
         lambda_client = clients.lambda_.request_metadata(
--- a/localstack/utils/aws/message_forwarding.py
+++ b/localstack/utils/aws/message_forwarding.py
@@ -39,6 +39,7 @@ def send_event_to_target(
     source_service: str = None,
 ):
     region = extract_region_from_arn(target_arn)
+    account_id = extract_account_id_from_arn(target_arn)
 
     if target is None:
         target = {}
@@ -47,7 +48,7 @@ def send_event_to_target(
             role_arn=role, service_principal=source_service, region_name=region
         )
     else:
-        clients = connect_to(aws_access_key_id=event.get("account"), region_name=region)
+        clients = connect_to(aws_access_key_id=account_id, region_name=region)
 
     if ":lambda:" in target_arn:
         lambda_client = clients.lambda_.request_metadata(
--- a/localstack/utils/aws/message_forwarding.py
+++ b/localstack/utils/aws/message_forwarding.py
@@ -38,7 +38,7 @@ def send_event_to_target(
     source_arn: str = None,
     source_service: str = None,
 ):
-    region = extract_region_from_arn(source_arn)
+    region = extract_region_from_arn(target_arn)
     account_id = extract_account_id_from_arn(source_arn)
 
     if target is None:
--- a/localstack/utils/aws/message_forwarding.py
+++ b/localstack/utils/aws/message_forwarding.py
@@ -48,14 +48,16 @@ def send_event_to_target(
             role_arn=role, service_principal=source_service, region_name=region
         )
     else:
-        client1 = connect_to(aws_access_key_id=account_id, region_name=region)
-        client2 = connect_to(region_name=region)
         try:
             from localstack_ext.config import ENFORCE_IAM
 
-            clients = client2 if ENFORCE_IAM else client1
+            clients = (
+                connect_to(region_name=region)
+                if ENFORCE_IAM
+                else connect_to(aws_access_key_id=account_id, region_name=region)
+            )
         except ImportError:
-            clients = client1
+            clients = connect_to(aws_access_key_id=account_id, region_name=region)
 
     if ":lambda:" in target_arn:
         lambda_client = clients.lambda_.request_metadata(
--- a/localstack/utils/aws/message_forwarding.py
+++ b/localstack/utils/aws/message_forwarding.py
@@ -5,10 +5,10 @@ import re
 import uuid
 from typing import Dict, Optional
 
-from localstack_ext.config import ENFORCE_IAM
 from moto.events.models import events_backends
 
 from localstack.aws.connect import connect_to
+from localstack.config import is_env_true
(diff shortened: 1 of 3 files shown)


viren-nadkarni (dismissed):
Similar to the previous PRs, the proper fix here is to make sure IAM setup is performed in the tests.

In this case, you must create and pass an appropriate role in the `PutRule` or `PutTarget` calls in these tests.

https://github.com/localstack/localstack/blob/f4e9632f4ff2bff6514f8fd42c631e0db510e204/tests/aws/services/events/test_events.py#L530-L534

Refer to https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutTargets.html

sannya-singal:
> this case, you must create and pass an appropriate role in the `PutRule` call in these tests.

In case of there is no IAM enforcement the setup should still work, right? 🤔 @viren-nadkarni?

sannya-singal:
According to my understanding in case of the [previous PR](https://github.com/localstack/localstack/pull/9639#discussion_r1394128323) we were explicitly setting the `access_key_id` as `None` but in this case we are already using `connect_to()` with `region` which will lead it to connect with the correct region and incorrect `account_id` even when the role is not specified.

viren-nadkarni:
@sannya-singal Please also make sure that the tests you're trying to address pass in the CI with a commit like: b21b9d86a1fc6a886fb04ac8f8f9c3f3807ad561 which you can revert before merge

sannya-singal:
> The test is not AWS validated, so we can't make that assumption. Even if IAM enforcement is disabled, role assumption is still a proper way to access resources.

@viren-nadkarni 

Maybe lets wait for review from @dfangl because its a bit confusing due to a comment [here](https://github.com/localstack/localstack/pull/9639#pullrequestreview-1744449202) which says it has to work without a role also, additionally we can add some more test cases for IAM enforcement to increase the robustness of multi-account and region functionality. 

Will also run a CI pipeline for non-default `account_id` and `region`.

dfangl (approved):
Like discussed, we need to probably look at it in a follow up!

sannya-singal:
Failing ext integration tests pipeline- https://github.com/localstack/localstack-ext/actions/runs/7738980761/job/21144883663

Failing tests: 

```
FAILED tests/aws/services/iam/test_inter_service_enforcement.py::TestInterServiceCommunicationEnforcement::test_events_sqs - Failed: DID NOT RAISE <class 'AssertionError'>
FAILED tests/aws/services/iam/test_inter_service_enforcement.py::TestInterServiceCommunicationEnforcement::test_events_lambda - Failed: DID NOT RAISE <class 'botocore.exceptions.ClientError'>
FAILED tests/aws/services/iam/test_inter_service_enforcement.py::TestInterServiceCommunicationEnforcement::test_events_sns - Failed: DID NOT RAISE <class 'AssertionError'>
FAILED tests/aws/services/iam/test_inter_service_enforcement.py::TestInterServiceCommunicationEnforcement::test_events_logs - Failed: DID NOT RAISE <class 'AssertionError'>
```

localstack/utils/aws/message_forwarding.py
```diff
@@ -47,7 +48,16 @@ def send_event_to_target(
...
+        try:
+            from localstack_ext.config import ENFORCE_IAM
+
+            clients = (
+                connect_to(region_name=region)
+                if ENFORCE_IAM
+                else connect_to(aws_access_key_id=account_id, region_name=region)
+            )
+        except ImportError:
+            clients = connect_to(aws_access_key_id=account_id, region_name=region)
```
viren-nadkarni, line 60: Are you sure about introducing this circular dependency? For happy path, shouldn't it be ideal to pass the role?

sannya-singal: Passing the role param here was already discussed and approved earlier. The new fix of introducing the `ENFORCE_IAM` attribute accounts to handle the failing ext test cases as mentioned [here](https://github.com/localstack/localstack/pull/9784#issuecomment-1938155138). Waiting for @dfangl review on this.

dfangl, line 60: I am not a fan of this at all. We should never make client creation dependent on the enforce IAM flag set - it should always be configured identically, with error handling if it fails due to a missing IAM permission. We should also not import pro code from community.

We have to ask us why this is necessary here. For all services which allow resource based policies (like lambda, sqs) it should not be necessary - they should parse the arn from the target themselves.

For those where it is necessary, it should always be conceptually the source account, not the target one (if cross account is possible).
We can allow this in the IAM engine itself - I can open a PR for it. Currently, it will recognize a call as being made from a service principal only if the access key id is `__internal__` - if this is too limiting, we can accept numeric account ids there as well.

localstack/utils/aws/message_forwarding.py
```diff
@@ -39,6 +39,7 @@ def send_event_to_target(
     source_service: str = None,
 ):
     region = extract_region_from_arn(target_arn)
+    account_id = extract_account_id_from_arn(target_arn)
```
dfangl, line 42: We should use the source arn here conceptually - not sure if cross account is allowed here, but if it is, the source arn would be the way to go. Services that do not allow cross account requests (like firehose) would not work, but that would be fine.

dfangl (changes requested):
I am not a fan of this ext import here at all - the client creation should always be identical, and always contain all information necessary for IAM enforcement.

I detailed it in the comments, and will open up a PR in -ext today.

dfangl (approved):
If all the tests pass, the changes look good :)

sannya-singal:
Ext Integration Tests: https://github.com/localstack/localstack-ext/actions/runs/7927487686

PR #9786, opened 2023-12-01
Add list to dictionary transformation function in provider_utils For CFn

Morijarti:
## Motivation
Often CFn will provide tags in a form of list of dictionaries instead of a single dictionary.
GenericModel solved by returning a function for transformation, instead of actually calling it.

Do to changes in resource providers, we are calling function directly and as such this method is required to reduce amount of duplicate code

## Changes
Implemented transform_list_to_dict method


simonrw (approved):
Looks great, thanks this will make adding `Tags` way easier.

PR #9789, opened 2023-12-01
CFn: fix AWS::Events::Rule ref/id/name attributes

simonrw:
## Motivation

When trying out the [AWS IATK](https://awslabs.github.io/aws-iatk/) I found that the outputs from creating an Events::Rule were incorrect. Specifically:

* we did not set the `Id` attribute, which goes against the spec
* we used the `Name` property as the physical resource id, but `Name` is a create-only property and is not accessible after the resource has been created
* the `Id` property (physical resource id) is built up from `<event-bus-name>|<rule-name>` (where `Name` is not required, and can be autogenerated if not specified)

## Changes

* Add an AWS validated test for AWS::Event::Rule resource capturing all (both) available properties - `Arn` and `Id` (via `!Ref`)
* Update the provider to _not_ set the `Name` property, and to correctly generate the `Id` property
    * note: this means the resource is no longer included in the `quirks.py` file 🎉


simonrw:
arm tests are unrelated lambda failures

localstack/services/cloudformation/engine/quirks.py
```diff
@@ -14,7 +14,6 @@
     # Example
     # "AWS::ApiGateway::Resource": "/properties/ResourceId",
     "AWS::Events::EventBus": "/properties/Name",
-    "AWS::Events::Rule": "/properties/Name",
```
dominikschubert, line 17: :partying_face:

dominikschubert (approved):
LGTM! Great catch

PR #9792, opened 2023-12-04
Add option to disable AWS_ENDPOINT_URL

joe4dev:
## Motivation

The environment variable `AWS_ENDPOINT_URL` automatically re-configures the endpoint URL of new AWS SDK clients (see supported list [here](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html)). We expose this variable by default in Lambda and should offer a way to disable this behavior if someone wants to connect to real AWS (as discussed in the Lambda team).

This option is useful if you prefer [Transparent Endpoint Injection](https://docs.localstack.cloud/user-guide/tools/transparent-endpoint-injection/) or custom certificate injection over automatic SDK configuration via `AWS_ENDPOINT_URL`. This option is currently only relevant when bundling newer AWS SDKs with your Lambda function but will become more important once newer runtimes ship with SDK versions that natively support `AWS_ENDPOINT_URL`.

## Changes

* Add configuration option to disable the automatic injection of `AWS_ENDPOINT_URL`

## Alternatives

`AWS_IGNORE_CONFIGURED_ENDPOINT_URLS` can be "used to ignore all custom endpoints configurations."
Source: https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html

However, this disables all possibilities to configure any AWS tool or SDK.

## Discussion

Do we generally aim for configuration options to have the default `0`?

What should be the default?
* a) `LAMBDA_DISABLE_AWS_ENDPOINT_URL=0`
* b) `LAMBDA_AWS_ENDPOINT_URL=1`

Double negations are more confusing to understand.

## Follow Up

- [ ] Document option in docs: https://github.com/localstack/docs/pull/947


localstack/config.py
```diff
@@ -834,6 +834,10 @@ def populate_edge_configuration(
     os.environ.get("BUCKET_MARKER_LOCAL", "").strip() or DEFAULT_BUCKET_MARKER_LOCAL
 )
 
+# PUBLIC: Opt-out to inject the environment variable AWS_ENDPOINT_URL for automatic configuration of AWS SDKs:
+# https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html
+LAMBDA_DISABLE_AWS_ENDPOINT_URL = is_env_true("LAMBDA_DISABLE_AWS_ENDPOINT_URL")
```
dfangl, line 839: We should add this to the environment variable list farther down as well.

thrau (approved):
interesting! LGTM

PR #9796, opened 2023-12-04
Update boto3 stubs and types for our client factory

Morijarti:
## Changes
Added:
- types for `pipes` and `application_autoscaling`
- updated setup.cfg

Edit @dominikschubert : Hijacking this PR a bit since it already included the application autoscaling client and I'v added `pipes` and updated dependencies + setup.cfg


pinzon (approved):
Great!

dominikschubert:
@alexrashed @silv-io 
Would like your input on a minor thing regarding pinned dependency updates: Do you have any concerns with running `make upgrade-pinned-dependencies` and only committing a single file or should we always commit other unrelated requirements file changes as well? :thinking:

silv-io:
> @alexrashed @silv-io Would like your input on a minor thing regarding pinned dependency updates: Do you have any concerns with running `make upgrade-pinned-dependencies` and only committing a single file or should we always commit other unrelated requirements file changes as well? 🤔
> 
> Example diff for this case with `requirements-runtime.txt`:
> 
> ```diff
> diff --git a/requirements-runtime.txt b/requirements-runtime.txt
> index cef8e34ba..7bb99fa03 100644
> --- a/requirements-runtime.txt
> +++ b/requirements-runtime.txt
> @@ -283,7 +283,7 @@ requests==2.31.0
>      #   rolo
>  requests-aws4auth==1.2.3
>      # via localstack-core (pyproject.toml)
> -responses==0.24.1
> +responses==0.25.0
>      # via moto-ext
>  rfc3339-validator==0.1.4
>      # via openapi-schema-validator
> @@ -291,7 +291,7 @@ rich==13.7.0
>      # via localstack-core (pyproject.toml)
>  rolo==0.3.0
>      # via localstack-core (pyproject.toml)
> -rpds-py==0.17.1
> +rpds-py==0.18.0
>      # via
>      #   jsonschema
>      #   referencing
> ```

I'd say if the pipeline remains green with changes that affect test and all the extras it depends on (runtime), there is no grief in just doing those upgrades. If it leads to a red pipeline, those issues should be fixed in their own PRs, and the old pins should remain.

alexrashed:
> Would like your input on a minor thing regarding pinned dependency updates: Do you have any concerns with running make upgrade-pinned-dependencies and only committing a single file or should we always commit other unrelated requirements file changes as well? 🤔

I'm not 100% sure I understand, but basically: The different requirement files should stay in sync. This is why the make target explicitly updates all of the requirements. Otherwise this could cause issues, f.e. by installing runtime requirements, but then installing different version with the test dependencies (as you already identified in https://github.com/localstack/localstack/pull/10164).
So if there are other changed requirement files than the one you might be particularly interested in the specific PR, please still also commit them.

pinzon (approved):

PR #9798, opened 2023-12-04
Improve lambda multiruntime tests

joe4dev:
## Motivation

The Lambda multiruntime tests were skipped due to a bug in the skipif condition ("x86_64" instead of "amd64" Arch.amd64). The skip was fixed in https://github.com/localstack/localstack/pull/9760 but the endpoint injection tests were still broken. Further, the Lambda multiruntime tests have several problems related to test selection, ARM compatibility, unnecessary dependency downloads, AWS parity, community-ext consistency, and unnecessary configurations.

## Changes

- Fix test selection for ARM and x86 test execution (we still need to find a way how to detect compatibility)
- Fix parity against AWS, with conditional for LocalStack for client creation
- Download less unused dependencies
- Prepare for ARM-compatible CI builds using Docker image tags instead of SHAs
- Unify Makefiles (still duplicated for now). They now have a `clean` target and parametrized `IMAGE`.
- Unify endpointinjection tests between community and ext
- Minimal SDK re-configuration instead of hardcoding region or credentials if not necessary (not necessary anymore with new Lambda provider)

## Follow Ups

- [ ] Run all Lambda tests in ARM runner (when all al1-based runtimes are deprecated by the end of 2023)
- [ ] Extend Lambda endpoint injection tests (e.g., test AWS_ENDPOINT_URL configuration)
- [ ] Document AWS SDK configuration for Lambda
- [ ] Remove legacy localstack-java-utils Java package


tests/aws/services/lambda_/test_lambda_common.py
```diff
@@ -246,30 +268,57 @@ def test_runtime_wrapper_invoke(self, multiruntime_lambda, snapshot, tmp_path, a
         assert invocation_result_payload["environment"]["WRAPPER_VAR"] == test_value
 
 
-# TODO: Split this and move to PRO
-@pytest.mark.skip(reason="this test is broken and is being worked on")
+# TODO: remove this once all non-arm compatible runtimes are deprecated by the end of 2023
+x86_runtimes = [
```
dfangl, line 272: Can we rename those? It was quite unclear for me that these are only relevant for the test_manual_endpoint_injection test.

tests/aws/services/lambda_/test_lambda_common.py
```diff
@@ -201,12 +220,15 @@ def test_uncaught_exception_invoke(self, multiruntime_lambda, snapshot, aws_clie
...
-    # this does only work on al2 lambdas, except provided.al2.
+    # Only works for >=al2 runtimes, except for any provided runtimes
     # Source: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html#runtime-wrapper
     @markers.multiruntime(
         scenario="introspection",
-        runtimes=["nodejs"],
-        # runtimes=["nodejs", "python3.8", "python3.9", "java8.al2", "java11", "dotnet", "ruby"],
+        runtimes=list(
+            set(TESTED_RUNTIMES)
+            - set(RUNTIMES_SKIP_ARM)
```
dfangl, line 229: Why do we substract RUNTIMES_SKIP_ARM here? I get it for the al1 runtimes, but currently the list also includes al2, which should work?

dfangl (approved):
Looks good, great iteration! We can tackle the build issues on arm for some runtimes in a later PR.

tests/aws/services/lambda_/test_lambda_common.py
joe4dev: You are right 🕵️  Newer runtimes should not be subtracted (e.g., >=`java8.al2`, at least not for x86 for now)
That difference came from the added runtimes due to the broken CI builds. We should be able to replace this with `runtimes` by the end of the year.

in any case better than the hardcoded `nodejs` runtimes from before 😅

Issue #9706, opened 2023-11-22
bug: SFN The function 'States.Format('Run for period {} to {}', $.fromDate, $.toDate)' had the following error: The JsonPath argument for the field '$.fromDate'


paul-barilla-1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```json
    "StepTwo": {
      "Type": "Task",
      "Resource": "arn:aws:states:::sqs:sendMessage.waitForTaskToken",
      "Parameters": {
        "MessageGroupId.$": "$$.Execution.Name",
        "MessageDeduplicationId.$": "States.Format('create-pay-review-{}', $$.Execution.Name)",
        "MessageBody": {
          "clientId": "#####",
          "runId.$": "$$.Execution.Name",
          "timestamp.$": "$$.State.EnteredTime",
          "name.$": "$.resources.transformTask.result.FileName",
          "description.$": "States.Format('AOC run for period {} to {}', $.fromDate, $.toDate)",
          "fileName.$": "$.resources.transformTask.result.FileName",
          "zipKey.$": "$.resources.transformTask.result.SharedBucketKey",
          "stepFunctionTaskToken.$": "$$.Task.Token"
        },
        "MessageAttributes": {
          "eventType": {
            "DataType": "String",
            "StringValue": "#####"
          }
        },
        "QueueUrl": "${aoc-events-queue}"
      },
      "Catch": [
        {
          "ErrorEquals": [
            "States.ALL"
          ],
          "Next": "Error Handler",
          "ResultPath": "$.error"
        }
      ],
      "TimeoutSeconds": 7200,
      "ResultPath": "$.resources.createTask",
      "ResultSelector": {
        "result.$": "$"
      },
      "Next": "Start 'GeneratingReports'"
    },
```

**Error at runtime**
```json
{
    "Type": "ExecutionFailed",
    "PreviousEventId": 75,
    "ExecutionFailedEventDetails": {
        "Error": "States.Runtime",
        "Cause": "An error occurred while executing the state 'StepTwo' (entered at the event id #75). The function 'States.Format('AOC run for period {} to {}', $.fromDate, $.toDate)' had the following error: 
        The JsonPath argument for the field '$.fromDate' could not be found in the input" 
        {
            "Execution": {
                "Id": "arn:aws:states:ap-southeast-2: 000000000000:execution:Workflow:a5dd0131-9e17-4c34-bade-8061d9925634",
                "Input": {
[...]

PR #9801, opened 2023-12-04
[SFN] Enhancements to States I/O, Support for null InputPath and OutputPath

MEPalma:
## Motivation
Currently the SFN v2 interpreters is sometimes unable to propagate the correct states input and output values, especially if InputPath declarations are used to modify it. The interpreter also does not support null values for InputPath and OutputPath declarations. This PR add support for such declarations, and reworks the environment's stack usage to ensure the updated input value is sampled for the remainder of the program state.

Closes: #9706 
Closes: other internal bug reports

## Changes
- Added grammar support and logic for null input and output path operators
- Reworked several components to source the input from the stack, instead of the state's input, these include: the payload json path fields and json path intrinsic function arguments.
- Reworked the use of the stack in several cases, such as intrinsic functions and map components, to ensure the correct values are pushed and consumed from the stack (now more critical as we keep at the first position the states' input)
- Added relevant test
- other minor supporting changes


dominikschubert (approved):
LGTM, another weird edge case :upside_down_face:

PR #9805, opened 2023-12-05
fix: handle BucketAlreadyOwnedByYou exception when uploading lambda in scenario test cases 

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, scenario tests should still create the consequent resources in corresponding accounts and region. When we try to upload lambas using non-default region value in scenario tests it throws `BucketAlreadyOwnedByYou` exception. 

## Changes
This PR: 
- handles this exception and doesn't try to recreate the bucket if the existing bucket has the same owner for non-default region in `_upload_to_s3` helper function. 
- replaces static region value `us-east-1` from environment variables. 

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`). This PR fixes: 

`tests.aws.scenario.note_taking.test_note_taking.TestNoteTakingScenario.test_validate_infra_setup`
`tests.aws.scenario.note_taking.test_note_taking.TestNoteTakingScenario.test_notes_rest_api`

### DIFF
--- a/localstack/constants.py
+++ b/localstack/constants.py
@@ -150,14 +150,10 @@ DEFAULT_AWS_ACCOUNT_ID = "000000000000"
 # Credentials used in the test suite
 # These can be overridden if the tests are being run against AWS
 # If a structured access key ID is used, it must correspond to the account ID
-# TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
-# TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
-# TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
-# TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-east-1"
-TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or "000000000001"
-TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "000000000001"
-TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test1"
-TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-west-1"
+TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
+TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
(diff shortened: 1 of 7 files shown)


localstack/constants.py
```diff
@@ -150,10 +150,14 @@
...
-TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
-TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-east-1"
+# TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
+# TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
+# TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
+# TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-east-1"
+TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or "000000000001"
+TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "000000000001"
+TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test1"
+TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-west-1"
```
sannya-singal, line 160: These changes are to showcase that the test tried to be address passes in the CI, will revert the this before merge.

localstack/testing/scenario/cdk_lambda_helper.py
```diff
@@ -141,9 +143,9 @@ def _zip_lambda_resources(
...
-    options = {"Bucket": bucket_name}
-    region_name = s3_client.meta.region_name
-    if region_name != AWS_REGION_US_EAST_1:
-        options["CreateBucketConfiguration"] = {"LocationConstraint": region_name}
-    s3_client.create_bucket(**options)
+    try:
+        create_s3_bucket(bucket_name, s3_client)
+    except ClientError as exc:
+        if exc.response["Error"]["Code"] != "BucketAlreadyOwnedByYou":
+            raise exc
```
dominikschubert, line 150:
```suggestion
    except ClientError as exc:
        # when creating an already existing bucket, regions differ in their behavior:
        # us-east-1 will silently pass (idempotent)
        # any other region will return a `BucketAlreadyOwnedByYou` exception.
        if exc.response["Error"]["Code"] != "BucketAlreadyOwnedByYou":
            raise exc
```

dominikschubert (approved):
LGTM. Thanks for fixing the tests here!

Good to merge when the suggestion is addressed, the changes in `constants.py` reverted and the pipeline afterwards is green :+1:

sannya-singal:
Ext Integration Tests Pipeline- https://github.com/localstack/localstack-ext/actions/runs/7104710813

PR #9806, opened 2023-12-05
add cross-account for CopyObject

bentsku:
## Motivation
As reported here: https://discuss.localstack.cloud/t/nosuchbucket-error-when-copying-file/640
The user had an issue where the bucket was created by another account, and he would then try to call `CopyObject` on it. 
I don't really have the setup for testing multi-account against AWS, but it feels like as S3 buckets are global, this should be possible and only a matter of IAM permission / ACL. 

https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

## Changes
- make `CopyObject` use cross-account access while getting the S3 buckets
- add a multi account test for it


macnev2013 (approved):
Thanks for the PR. LGTM 🚀

Issue #9804, opened 2023-12-05
bug: API Gateway + OpenAPI with Integration doesn't respect integration httpMethod


zollf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using `x-amazon-apigateway-integration` and setting `httpMethod`, it won't use `httpMethod` that is set, instead it looks like it is using the incoming request method.

Here is an example:

```terraform
resource "aws_api_gateway_rest_api" "main" {
  name = "TestApiGateway"

  body = jsonencode({
    openapi = "3.0.3"
    #...other properties
    paths = {
      # ... other methods
      "/" = {
        delete = {
          summary = "DELETE Test"
          description = "DELETE Test"
          "x-amazon-apigateway-integration" = {
            httpMethod           = "POST"
            type                 = "aws"
            passthroughBehaviour = "WHEN_NO_TEMPLATES"
            uri                  = "arn:aws:apigateway:${local.region}:events:action/PutEvents"
            requestTemplates     = {
              "application/json" = "..."
            }
          }
          responses = {
            "200" = {
              description = "Success"
              content = {}
            }
          }
        }
      }
    }
  })
}
```

In `x-amazon-apigateway-integration` I have `httpMethod` "POST", however the current behaviour is using "DELETE" when integrating with AWS services. This setup works on AWS

This breaks `arn:aws:apigateway:${local.region}:events:action/PutEvents` as it expects "POST" but receives "DELETE"

This is the output when sending a DELETE request to that endpoint

```bash
egration for API Gateway ID 'uhy3xhv50s': API Gateway integration type "AWS", method "DELETE", URI "arn:aws:apigateway:ap-southeast-2:events:action/PutEvents" not yet implemented
```

### Expected Behavior

Expected behaviour is for `httpMethod` within `x-amazon-apigateway-integration` to control the method used within for aws integrations. 

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

I've created a replication repository using terraform and docker-compose: https://github.com/zollf/localstack-apigateway-integration-bug

### Environment

```markdown
- OS: macOS Sonoma Version 14.0
- LocalStack: latest
```

mmarkus-twilio:
Hello!
We have stumbled onto the same issue. This issue is closed but it doesn't look like this has been included in the latest release. Can you confirm?

bentsku:
Hello @mmarkus-twilio!

[...]

PR #9810, opened 2023-12-05
fix APIGW import integration method for AWS integration type

bentsku:
## Motivation
As reported in #9804, we would not take into account the `httpMethod` when importing integration of `AWS` type.

## Changes
- create a test to validate the use case
- update the helper to take the `httpMethod` field, and ignore it for different type of integration

Note: test failure is unrelated


bentsku:
Test failure is unrelated

calvernaz (approved):
Thanks for tackling this. I'll go ahead and merge it

Issue #9722, opened 2023-11-24
bug: stepfunctions time data does not match format of ISO date with milliseconds


attila:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

State machines do not recognise ISO dates with milliseconds as `TimeStampPath`.

Example state machine:

```json
{
  "Comment": "Wait with TimestampPath demo",
  "StartAt": "Wait",
  "States": {
    "Wait": {
      "Type": "Wait",
      "TimestampPath": "$.start_at",
      "End": true
    }
  }
}
```

When started with `{ "start_at": "2023-11-24T08:38:00.000Z" }`

Execution fails with

```text
localstack_main  | 2023-11-24T08:49:07.437 ERROR --- [ad-63 (eval)] l.s.s.a.c.eval_component   : Exception=ValueError, Details=["time data '2023-11-24T09:00:00.000Z' does not match format '%Y-%m-%dT%H:%M:%SZ'"] at '(TimestampPath| {'path': '$.start_at'}'
localstack_main  | 2023-11-24T08:49:07.437 ERROR --- [ad-63 (eval)] l.s.s.a.c.eval_component   : Exception=ValueError, Details=["time data '2023-11-24T09:00:00.000Z' does not match format '%Y-%m-%dT%H:%M:%SZ'"] at '(StateWait| {'comment': None, 'input_path': (InputPath| {'input_path_src': '$'}, 'output_path': (OutputPath| {'output_path': '$'}, 'state_entered_event_type': 'WaitStateEntered', 'state_exited_event_type': 'WaitStateExited', 'name': 'Wait', 'state_type': <StateType.Wait: 20>, 'continue_with': <localstack.services.stepfunctions.asl.component.state.state_continue_with.ContinueWithEnd object at 0x7f1ba0fec5d0>, 'wait_function': (TimestampPath| {'path': '$.start_at'}}'
localstack_main  | 2023-11-24T08:49:07.438 ERROR --- [ad-63 (eval)] l.s.s.a.c.program.program  : Stepfunctions computation ended with exception '<class 'ValueError'>(time data '2023-11-24T09:00:00.000Z' does not match format '%Y-%m-%dT%H:%M:%SZ')'.
```

### Expected Behavior

It should work like in the AWS service.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

docker-compose.yml

```yml
[...]

PR #9813, opened 2023-12-06
Fix parsing of StepFunction timestamps 

dominikschubert:
## Motivation

Resolves https://github.com/localstack/localstack/issues/9722

We've been assuming that all timestamps should only include up to second precision, but StepFunctions on AWS also allows specifying milliseconds or microseconds (though they seem to be ignored anyway)

## Changes

- Added tests for multiple different timestamp formats in `TimestampPath` (both positive & negative)  

## TODO

What's left to do:

- [x] Implement `An error occurred while executing the state 'State_1' (entered at the event id #2). ` error wrapping
- [x] Add test for older timestamp
- [x] Add test for timestamp too far in future
- [x] Fix timestamp parsing for milli & microseconds (dropping them)


MEPalma (approved):
LGTM!

PR #9815, opened 2023-12-06
Fix: lookup more specific DNS names directly

simonrw:
## Motivation

When resolving DNS names, if someone adds a more specific entry to an existing entry with a wildcard, the more specific name is not matched. For example:

```
ADD foo.example.com 1.2.3.4
ADD *.example.com 5.6.7.8

dig foo.example.com => 5.6.7.8
dig bar.example.com => 5.6.7.8
```

A request to `foo.example.com` should resolve to `1.2.3.4`

## Changes

Look up the DNS name by string, which is what we store in the `add_host` method

### DIFF
--- a/localstack/dns/server.py
+++ b/localstack/dns/server.py
@@ -93,6 +93,15 @@ ClientAddress = Tuple[str, int]
 psutil_cache = TTLCache(maxsize=100, ttl=10)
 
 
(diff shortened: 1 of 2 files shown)


dfangl (approved):
Good catch! The trailing dot inconsistency were quite annoying!

PR #9820, opened 2023-12-07
Fix serialization of content in raw snapshots

dominikschubert:
## Motivation

Since we're not using transformers with raw snapshots, we need to make sure datetime objects are handled properly, otherwise `json.dumps` with the default JSON encoder will fail.

## Changes

- Use `CustomEncoder` from our json utils to encode the response.


bentsku (approved):
Good catch, LGTM!

PR #9825, opened 2023-12-08
Fix lambda function name validation for CreateFunction

joe4dev:
## Motivation

The `FunctionName` in the Lambda `CreateFunction` operation is currently not validated and cannot handle function ARNs.
The [AWS API reference](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-FunctionName) describes that the operation can accept the following options for `FunctionName`:

* Function name – my-function.
* Function ARN – arn:aws:lambda:us-west-2:123456789012:function:my-function.
* Partial ARN – 123456789012:function:my-function.

This issue came up in a support case where a function ARN crashed in the `event_manager.py` because it tried to create an SQS queue with an invalid name (containing `:`).

## Changes

* Add AWS-validated tests for exceptions and valid function ARNs
* Fix validation of `FunctionName` for the Lambda `CreateFunction` operation

## Testing

* `tests.aws.services.lambda_.test_lambda_api.TestLambdaFunction.test_function_arns`

## Notes

* Other operations in the Lambda API also do not cover all validation implemented for `CreateFunction` now. However, they often use `api_utils.get_name_and_qualifier`, which supports function ARNs and handles some validations (i.e., qualified and region). If we want to improve server-side validations consistently, we need better boto-spec-based validation abstractions as suggested by @thrau [here](https://github.com/localstack/localstack/pull/7675#discussion_r1107777058).
* The `FunctionName` cannot be updated, so we don't need to consider renaming.

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -404,6 +404,113 @@ class TestLambdaFunction:
             == get_function_response_updated["Configuration"]["CodeSize"]
         )
 
+    # TODO: fix type of AccessDeniedException yielding null
+    @markers.snapshot.skip_snapshot_verify(paths=["function_arn_other_account_exc..Error.Message"])
+    @markers.aws.validated
+    def test_function_arns(
+        self, create_lambda_function, region, account_id, aws_client, lambda_su_role, snapshot
+    ):
+        # create_function
+        function_name_1 = f"test-function-arn-{short_uid()}"
+        function_arn = f"arn:aws:lambda:{region}:{account_id}:function:{function_name_1}"
+        function_arn_response = create_lambda_function(
+            handler_file=TEST_LAMBDA_PYTHON_ECHO,
+            func_name=function_arn,
+            runtime=Runtime.python3_12,
+        )
+        snapshot.match("create-function-arn-response", function_arn_response)
+
+        function_name_2 = f"test-partial-arn-{short_uid()}"
+        partial_arn = f"{account_id}:function:{function_name_2}"
+        function_partial_arn_response = create_lambda_function(
+            handler_file=TEST_LAMBDA_PYTHON_ECHO,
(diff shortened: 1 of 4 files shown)


dfangl (approved):
Looks good! Do you think it makes sense to put that logic in a separate function, just to preserve readability in the CreateFunction body?

joe4dev:
> Looks good! Do you think it makes sense to put that logic in a separate function, just to preserve readability in the CreateFunction body?

Possibly. It is rather minor as long as we cannot reuse it. However, if the validation creep continues, we could consider reviving the idea of server-side "boto-spec-based validation abstractions as suggested by Thomas https://github.com/localstack/localstack/pull/7675#discussion_r1107777058.":

I did a quick try but it didn't work so far (doesn't produce validation errors):
```
from localstack.aws.protocol.validate import ParamValidator

service = load_service("lambda")
operation = service.operation_model("CreateFunction")
shape = operation.input_shape.members["FunctionName"]
params = "invalid:fun"
validator = ParamValidator()
validation_errors = validator.validate(params, shape)

print(str(validation_errors))
```

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -404,6 +404,113 @@ def test_lambda_code_location_s3(
             == get_function_response_updated["Configuration"]["CodeSize"]
         )
 
+    # TODO: fix type of AccessDeniedException yielding null
+    @markers.snapshot.skip_snapshot_verify(paths=["function_arn_other_account_exc..Error.Message"])
```
joe4dev, line 408: We would need some other type of `AccessDeniedException` to fix this tiny snapshot diff because AWS returns `None` but the exception translates the None message into an empty string:

`/Error/Message None (type: <class 'NoneType'>) → '' (type: <class 'str'>)... (expected → actual)`

PR #9828, opened 2023-12-08
Fix AWS parity for SQS list queues with empty response

joe4dev:
## Motivation

Related discussion in https://github.com/localstack/localstack/pull/9710#discussion_r1419321271

> For empty lists, AWS often omits the field entirely, independent of the protocol. I noticed this behavior recently while parity testing different SDK versions (which use different protocols) in the Lambda multiruntime tests.
> I moved my Notion testing page to the SQS service to demonstrate the behavior for the listQueues operation (see [here](https://www.notion.so/localstack/SQS-SDK-parity-testing-164bfb9dbdd541928fd2b2abc275bac7?pvs=4)).

This behavior is independent of the protocol (both query + json).

## Changes

* Add AWS-validated test scenario with queue prefix matching that returns an empty result
* Omit the `QueueUrls` in the API response for empty results.

### DIFF
--- a/localstack/services/sqs/provider.py
+++ b/localstack/services/sqs/provider.py
@@ -771,6 +771,9 @@ class SqsProvider(SqsApi, ServiceLifecycleHook):
             #  NextToken response element directly outside of the AWS CLI.
             urls = urls[:max_results]
 
+        if len(urls) == 0:
+            return ListQueuesResult()
+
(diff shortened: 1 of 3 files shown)


thrau (approved):
nice find! LGTM

null/empty field handling always gets me

Issue #9826, opened 2023-12-08
bug: SQS queues cause deployment failures when using IaC (AWS CDK)


rudolfsberzins:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After the initial `cdklocal deploy` run, after any change has happened, Localstack will not be able to locate the created SQS queue and will delete it.
![Screenshot 2023-12-08 at 10 16 15](https://github.com/localstack/localstack/assets/16351444/f7a29aa3-43f9-41b2-bba1-c71b1fbc2d8b)

### Expected Behavior

No error when adding changes after having provisioned SQS queue

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

How to replicate:
1. In an empty folder initialise a new CDK project using `cdklocal init -l python`
2. Substitute the `..._stack.py` file with.
```
from aws_cdk import Duration, Stack, aws_sqs as sqs, aws_s3 as s3
from constructs import Construct


class LocalstackTestStack(Stack):
    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
        super().__init__(scope, construct_id, **kwargs)

        # The code that defines your stack goes here

        # example resource
        queue = sqs.Queue(
            self,
            "LocalstackTestQueue",
            visibility_timeout=Duration.seconds(300),
        )
```
3. Run `cdklocal bootstrap`and `cdklocal deploy`.
4. Make any "significant" change in ..._stack.py file. For example add a bucket.
```
from aws_cdk import Duration, Stack, aws_sqs as sqs, aws_s3 as s3
from constructs import Construct


class LocalstackTestStack(Stack):
    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
        super().__init__(scope, construct_id, **kwargs)

        # The code that defines your stack goes here

        # example resource
        queue = sqs.Queue(
            self,
            "LocalstackTestQueue",
            visibility_timeout=Duration.seconds(300),
        )

        self.input_bucket = s3.Bucket(
            self,
            "input_bucket",
            block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
        )
```
5. Run `cdklocal deploy`. This will throw the error.

### Environment

```markdown
- OS: macOS Sonoma Version 14.1.1
- LocalStack:

> cdklocal --version
2.114.1 (build 02bbb1d)

>localstack --version
3.0.2
```
```


### Anything else?

[...]

PR #9831, opened 2023-12-08
Fix SQS CFn updates with no change

simonrw:
## Motivation

If a CFn stack containing an SQS queue is updated, but the SQS queue is not changed, we trigger invalid behaviour where we assume information in the incoming request that is not present. Specifically, we try to get the queue name from the incoming state, and if this was not specified in the first place (i.e. we rely on auto-generating the queue name) the stack fails to update.

While investigating this issue I found another issue when adding a completely new resource. The "update" method is still invoked, but because we raise a `NotImplementedError` we try to handle this in the executor, but this assumes there is a previous state. For a completely new resource, this is not the case.

Closes https://github.com/localstack/localstack/issues/9826

## Changes

- Add test for updating SQS queues
- Fix updating sqs queues
- Fix situations where an update is incorrectly detected for a completely new resource


localstack/services/sqs/resource_providers/aws_sqs_queue.py
```diff
@@ -181,18 +183,35 @@ def update(
...
 
+        assert request.previous_state is not None
+
+        should_replace = (
+            request.desired_state.get("QueueName", request.previous_state["QueueName"])
+            != request.previous_state["QueueName"]
+        ) or (
+            request.desired_state.get("FifoQueue", request.previous_state["FifoQueue"])
+            != request.previous_state["FifoQueue"]
+        )
```
simonrw, line 195: @dominikschubert there must be a way to automate this sort of detection from the spec... :thinking:

localstack/services/cloudformation/resource_provider.py
```diff
@@ -809,6 +809,11 @@ def execute_action(
                         request.resource_type,
                         request.logical_resource_id,
                     )
+                    if request.previous_state is None:
+                        # this is an issue with our update detection. We should never be in this state.
+                        request.action = "Add"
```
thrau, line 814: so is this a TODO for the framework? or what do we do about the comment?

thrau (approved):
LGTM! thanks for adding the test

bentsku (approved):
LGTM! thanks for making this cover more use cases and upgrading my very naive fix!

localstack/services/cloudformation/resource_provider.py
simonrw: This is an issue with the CFn engine in LocalStack. We should never try to perform an update/modify on a resource that has no previous state, since this means the resource has not been created yet.

PR #9841, opened 2023-12-11
Install websocket-client as it is no longer included with docker client

simonrw:
## Motivation

In `docker-py` version 7.0.0 the automatic installation of `websocket-client` was removed. We require `websocket-client` in our tests as we test our websocket support there, however automatically updating docker to 7.0.0 (from 6.1.3) has removed this dependency for us.

## Changes

Explicitly install `websocket-client` in our `tests` extra


alexrashed (approved):
Thanks so much for jumping on this and unblocking the pipeline! 💚 🦸🏽

PR #9856, opened 2023-12-12
fix s3 key handling with trailing slash

### LINKED ISSUE #9837: bug: S3 listObjects and folder names containing a space

### Current Behavior

When folder names contain a space then `listObjects` returns names without `/` at the end. Without space it works fine.

### Expected Behavior

Folders containing space in the name are returned with `/` at the end.

With a `docker run` command

### Steps To Reproduce

awslocal s3api put-object --bucket bucket-name --key "folder name/"

awslocal s3api list-objects --bucket bucket-name
```
{
    "Contents": [
        {
            "Key": "folder name",
            "LastModified": "2023-12-11T10:15:20.000Z",
            "Size": 0,
            "StorageClass": "STANDARD",
            "Owner": {
                "DisplayName": "webfile",
                "ID": "75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a"
            }
        }
    ],
    "RequestCharged": null
}
```

### Environment

```markdown
- OS: Linux Mint 21.2
- LocalStack: localstack/localstack:3.0
```

### Anything else?

_No response_

bentsku:
## Motivation
This would fix #9837. We've had multiple iterations of this, we were still unquoting the key to compare with the `base_url` but there was no need for it, I think this was a remnant of pre-Werkzeug 3.0

## Changes
- update the logic to be more resilient in the parser (proper splitting of the path, multiple trailing slashes...) 
- add more test cases

### DIFF
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -1060,10 +1060,8 @@ class S3RequestParser(RestXMLRequestParser):
     ) -> Any:
         """
         Special handling of parsing the shape for s3 object-names (=key):
-        trailing '/' are valid and need to be preserved, however, the url-matcher removes it from the key
-        we check the request.url to verify the name.
-        Trailing slashes are stripped from the path, so we need special logic to compare the parsed Key parameter
-        against the path and add the missing slashes
+        Trailing '/' are valid and need to be preserved, however, the url-matcher removes it from the key.
+        We need special logic to compare the parsed Key parameter against the path and add back the missing slashes
         """
         if (
             shape is not None
--- a/localstack/aws/protocol/parser.py
+++ b/localstack/aws/protocol/parser.py
@@ -1062,8 +1062,8 @@ class S3RequestParser(RestXMLRequestParser):
         Special handling of parsing the shape for s3 object-names (=key):
         trailing '/' are valid and need to be preserved, however, the url-matcher removes it from the key
         we check the request.url to verify the name.
-        We encode the key for the comparison with `base_url` in case of special characters with the same logic as
-        Werkzeug, with safe characters.
+        Trailing slashes are stripped from the path, so we need special logic to compare the parsed Key parameter
+        against the path and add the missing slashes
         """
         if (
             shape is not None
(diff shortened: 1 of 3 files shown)


localstack/aws/protocol/parser.py
```diff
@@ -1063,17 +1062,21 @@ def _parse_shape(
         Special handling of parsing the shape for s3 object-names (=key):
         trailing '/' are valid and need to be preserved, however, the url-matcher removes it from the key
         we check the request.url to verify the name.
-        We decode the key for the comparison with `base_url` in case of special characters.
+        Trailing slashes are stripped from the path, so we need special logic to compare the parsed Key parameter
+        against the path and add the missing slashes
```
alexrashed, line 1066: nit: This comment is a bit confusing now, because it describes the removal of the trailing slashes twice.

localstack/aws/protocol/parser.py
```diff
@@ -1063,17 +1062,21 @@ def _parse_shape(
...
         if (
             shape is not None
             and uri_params is not None
             and shape.serialization.get("location") == "uri"
             and shape.serialization.get("name") == "Key"
-            and request.base_url.endswith(f"{unquote(uri_params['Key'])}/")
+            and (
+                (trailing_slashes := request.path.partition(uri_params["Key"])[2])
+                and all(char == "/" for char in trailing_slashes)
+            )
```
alexrashed, line 1076: If I see this correctly, this now takes everything after the key, and then checks if there are only slashes afterwards. Afterwards we add (the arbitrary amount of slashes) back to the parsed key. Is that correct?
So you solution now works for 0 to arbitrary trailing slashes, is that correct?
How exactly does this fix issues with spaces in the key name?

alexrashed (commented):
Nice catch! Quite a tricky case! I only have two comments, when these are clarified we should be good to go! :) 🦸🏽

localstack/aws/protocol/parser.py
bentsku: Oops, yes, will rewrite everything!

localstack/aws/protocol/parser.py
bentsku: The issue with space was the call to `unquote` of the key, which was compared a "quoted" path. So this was one part of the fix, and I've also added the logic to handle multiple trailing slashes.

alexrashed: Ah, right, I missed the removal of unquote in the changeset. That makes sense, thanks!

alexrashed (approved):

PR #9858, opened 2023-12-12
Feat/cloudwatch v2 put composite alarm

baermat:
## Motivation
Get the new cloudwatch v2 provider more on par with the old one with composite alarms

## Changes
- Adds put_composite_alarm to the new provider in the same mock fashion as in the v1 provider
- changes test name (since the test didn't do what the name implied


steffyP (approved):
Nice work 👍 clean and precise!

Issue #9857, opened 2023-12-12
bug: s3.GetObject throws error with applied lifecycle rule


febluthn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

An unexpected error (python) occurred when calling the GetObject operation

```
An error occurred (InternalError) when calling the GetObject operation (reached max retries: 2): exception while calling s3.GetObject: Traceback (most recent call last):
  File "/opt/code/localstack/localstack/aws/chain.py", line 90, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 123, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 93, in __call__
    skeleton_response = self.skeleton.invoke(context)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(context, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 166, in dispatch_request
    result = handler(context, instance) or {}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/api/core.py", line 160, in operation_marker
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/v3/provider.py", line 784, in get_object
    if expiration_header := self._get_expiration_header(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/v3/provider.py", line 376, in _get_expiration_header
    if lifecycle_rule := get_lifecycle_rule_from_object(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/utils.py", line 676, in get_lifecycle_rule_from_object
    if any(
       ^^^^
  File "/opt/code/localstack/localstack/services/s3/utils.py", line 677, in <genexpr>
[...]

PR #9859, opened 2023-12-12
fix S3 lifecycle tags filter

bentsku:
## Motivation
As reported in #9857, we have an issue with S3 lifecycle configuration and tag filters when the object does not have tags.

## Changes
- add a proper check when applying lifecycle configuration if the object tags are not `None` before trying to get attributes from it


macnev2013 (approved):
Thanks for fixing it. LGTM 🚀

PR #9862, opened 2023-12-12
Fix secret rotation in secretsmanager

lakkeger:
## Motivation
Secret rotation seemed to be completely broken.

## Changes
- added patch for response so immediate rotation is optional
- added logic to set/enable rotation but do not run immediately
- Lambda invokation happens via LS instead of moto using the user's previously created Lambda
- fixed AWSCURRENT and AWSPENDING tag on same secret version, which is pretty common (we were failing on it)
- calculating next rotation date now
- skipping other lambda steps in rotation if lambda returns an error 
- in progress rotation exception is delayed after lambda trigger so we can give a chance to the lambda function to fix the state, in AWS the runs are scheduled until tags are fixed
- smoke test

### DIFF
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -565,12 +565,8 @@ def backend_rotate_secret(
 
     try:
         lm_client = connect_to(region_name=self.region_name).lambda_
-        get_func_res = lm_client.get_function(FunctionName=rotation_lambda_arn)
+        lm_client.get_function(FunctionName=rotation_lambda_arn)
     except Exception:
-        # Fall through to ResourceNotFoundException.
-        pass
-    #
-    if not get_func_res:
         raise ResourceNotFoundException("Lambda does not exist or could not be accessed")
 
     secret = self.secrets[secret_id]
@@ -651,8 +647,8 @@ def backend_rotate_secret(
         except Exception:
             if pending_version:
                 raise pending_version.pop()
-            # Fall through so we'll "stuck" with a secret version in AWSPENDING state.
-            pass
+            # Fall through if there is no previously pending version so we'll "stuck" with a new
+            # secret version in AWSPENDING state.
 
     return secret.to_short_dict(version_id=new_version_id)
 
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -586,6 +586,7 @@ def backend_rotate_secret(
     # version as AWSCURRENT then any later invocation of RotateSecret assumes
     # that a previous rotation request is still in progress and returns an error.
     try:
+        pending_version = None
         version = next(
             version
             for version in secret.versions.values()
@@ -593,7 +594,8 @@ def backend_rotate_secret(
         )
         if AWSCURRENT not in version["version_stages"]:
             msg = "Previous rotation request is still in progress."
-            raise InvalidRequestException(msg)
+            # Delay exception, so we can trigger lambda again
+            pending_version = list(InvalidRequestException(msg), version)
 
     except StopIteration:
         # Pending is not present in any version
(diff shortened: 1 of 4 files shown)


localstack/services/secretsmanager/provider.py
```diff
@@ -515,6 +514,22 @@ def fake_secret_remove_version_stages_from_old_versions(fn, self, version_stages
         del self.versions[version_no_stages]
 
 
+@patch(SecretsManagerResponse.rotate_secret, pass_target=False)
+def rotate_secret(self) -> str:
```
dominikschubert, line 518: I'd suggest adding a comment as to why we patch this. Otherwise this knowledge gets lost pretty quickly and we can't easily reassess if we still need that patch

localstack/services/secretsmanager/provider.py
```diff
@@ -574,61 +582,75 @@ def backend_rotate_secret(
...
-        tags=secret.tags,
-        version_id=new_version_id,
-        version_stages=[AWSPENDING],
-    )
     secret.rotation_lambda_arn = rotation_lambda_arn
-    if rotation_rules:
-        secret.auto_rotate_after_days = rotation_rules.get(rotation_days, 0)
+    secret.auto_rotate_after_days = rotation_rules.get(rotation_days, 0)
     if secret.auto_rotate_after_days > 0:
+        secret.next_rotation_date = int(time.time()) + (int(rotation_period) * 86400)
```
dominikschubert, line 603:
```suggestion
        wait_interval_s = int(rotation_period) * 86400
        secret.next_rotation_date = int(time.time()) + wait_interval_s
```

nit: splitting this makes the behavior clearer to the reader

localstack/services/secretsmanager/provider.py
```diff
@@ -574,61 +582,75 @@ def backend_rotate_secret(
...
+        if not pending_version:
+            # Begin the rotation process for the given secret by invoking the lambda function.
+            #
+            # We add the new secret version as "pending". The previous version remains
+            # as "current" for now. Once we've passed the new secret through the lambda
+            # rotation function (if provided) we can then update the status to "current".
+            new_version_id = self._from_client_request_token(client_request_token)
+
+            # An initial dummy secret value is necessary otherwise moto is not adding the new
+            # secret version.
```
dominikschubert, line 617: Again, great comment. Especially since this is might be something we can drop at some point if moto changes.

localstack/services/secretsmanager/provider.py
```diff
@@ -574,61 +582,75 @@ def backend_rotate_secret(
...
+            new_version_id = pending_version.pop()["version_id"]
+
+        try:
+            for step in ["create", "set", "test", "finish"]:
+                resp = lm_client.invoke(
+                    FunctionName=rotation_lambda_arn,
+                    Payload=json.dumps(
+                        {
+                            "Step": step + "Secret",
+                            "SecretId": secret.name,
```
dominikschubert, line 640: Not sure if this is 100% correct but probably not as important right now

localstack/services/secretsmanager/provider.py
```diff
@@ -574,61 +582,75 @@ def backend_rotate_secret(
...
+                            "ClientRequestToken": new_version_id,
+                        }
+                    ),
+                )
+                if resp.get("FunctionError"):
+                    raise Exception
```
dominikschubert, line 646: I'd suggest at least giving this Exception a proper message and debug log it.

dominikschubert (approved):
Great that this has finally been reworked :+1: 
Also thanks for the basic happy path test! I guess at some point we should also add a test for a failure case (e.g. lambda invocation fails), but it's already a lot better than before obviously :grin: 

Since this is a fairly complex operation, I'd say there is still room for more context/documentation but I really like the nature of the comments you've added since they give additional context.

Left a few nits and suggestions. Feel free to merge once they have been addressed.

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -387,7 +432,8 @@ def test_rotate_secret_with_lambda_2(
         sig_rnfe_1 = lambda_rotate_secret.secret_signal_resource_not_found_exception_on_create(
             version_id_1
         )
-        get_sig_rnfe_1 = aws_client.secretsmanager.get_secret_value(SecretId=sig_rnfe_1)
+        with pytest.raises(Exception):
```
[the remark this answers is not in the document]
lakkeger: This is a legacy test, which marked skipped.

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -98,6 +97,23 @@ def _is_secret_deleted():
...
+                if key == secret_version:
+                    secret_stage_tags = tags
+                    break
+            return "AWSCURRENT" in secret_stage_tags
+
+        success = poll_condition(condition=_is_secret_rotated, timeout=120, interval=5)
+        if not success:
+            LOG.warning(
+                f"Timed out whilst awaiting for secret '{secret_id}' to be rotated to new version."
+            )
```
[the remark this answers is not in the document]
lakkeger: Wanted to keep this universal as there might be a test in the future where we want a failing rotation first and then a successful one.

PR #9864, opened 2023-12-13
pin docker dependency to <7.0.0

dominikschubert:
## Motivation
It doesn't seem to break anything here yet but it technically could lead to the same kind of issues for users of community.

Recently the docker PyPI package was updated to 7.0.0 which started breaking integration tests in -ext with the following kind of exception:
```
docker.errors.DockerException: invalid tag 'localhost.localstack.cloud:4514/r-a9dfa6f4:latest': invalid reference format` anymore.
```

## Context

Change that was included in their `7.0.0` release: https://github.com/docker/docker-py/pull/3191

## Changes

- pin `docker` dependency to `<7.0.0` to avoid the breaking change for now until we know how to circumvent this

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,7 +65,7 @@ base-runtime =
     cbor2>=5.2.0
     dnspython>=1.16.0
     # TODO tag incompatibility introduced in 7.0.0 with https://github.com/docker/docker-py/pull/3191
-    docker>=6.1.1<7.0.0
+    docker>=6.1.1,<7.0.0
     jsonpatch>=1.24,<2.0
     hypercorn>=0.14.4
     pyopenssl>=23.0.0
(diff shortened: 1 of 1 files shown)


dfangl (approved):
Thanks for the quick fix!

dfangl:
Already reported upstream: https://github.com/docker/docker-py/issues/3195
Should be fixed by: https://github.com/docker/docker-py/pull/3196

dominikschubert:
Failure of `test_kinesis_firehose_opensearch_s3_backup` seems like a flake

consideRatio:
Reported this as a bug in https://github.com/docker/docker-py/issues/3240

alexrashed:
Hi @consideRatio!
It seems like there might be a misunderstanding. The bug was already reported and fixed in docker-py (see the comment from @dfangl above). There just isn't a release containing the fix yet.
However, we are pinning the version of `docker-py` until the release is out so you shouldn't see this issue in LocalStack.

consideRatio:
Oh, thank you for helping me get that!

PR #9865, opened 2023-12-13
Updated Cloudformationn Schema to latest version

Morijarti:
## Motivation
Schema file which we are using for autogeneration of resource providers is out-dates, causing it to generate providers with missing resources

## Changes
Updated schema with latest version, using us-east-1
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html


simonrw (approved):
Great, thanks for updating this.

We will look for automating this in the future

PR #9866, opened 2023-12-13
Remove old basic snapshot transformer and exemptions

dominikschubert:
## Motivation

Follow-up to https://github.com/localstack/localstack/pull/9772 which introduced an exemption list with paths that are still using the old transformer. This PR can't be merged yet since it will have a lot of CI failures. It should stay open until other PRs have regenerated/fixed the issues in the specific services. Until then it can be rebased and re-run to get an overview of all currently detected issues.

## Changes

- Remove old basis snapshot transformer list & exemption paths

### DIFF
--- a/localstack/testing/pytest/snapshot.py
+++ b/localstack/testing/pytest/snapshot.py
@@ -17,7 +17,6 @@ from localstack.testing.snapshots.report import render_report
 from localstack.testing.snapshots.transformer import RegexTransformer
 from localstack.testing.snapshots.transformer_utility import (
     SNAPSHOT_BASIC_TRANSFORMER,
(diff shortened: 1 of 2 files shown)


localstack/testing/pytest/snapshot.py
```diff
@@ -17,7 +17,6 @@
 from localstack.testing.snapshots.transformer import RegexTransformer
 from localstack.testing.snapshots.transformer_utility import (
```
meteharun: Looks good to me.

alexrashed:
@dominikschubert This PR seems a bit outdated, do you think we can close it?

dominikschubert:
I actually recently updated it and it's overall still relevant as the migration hasn't progressed, but willing to recreate it at a later time :+1:

PR #9871, opened 2023-12-14
Add flag to disable kinesis-mock persistence

dfangl:
## Motivation
With recent kinesis performance tests, we have seen a significant performance degradation up to an out-of-memory error of kinesis-mock, especially when pushing over 50 entries with the maximum payload of 1024KB to a stream.
This behavior is way worse when the persistence is enabled when starting kinesis-mock. For now, to fix it for users not needing persistence, we need a flag to disable it.

In the future, we need to spend some work on increasing the performance and memory usage of the persistence of our kinesis implementation.

## Changes
- Add flag to disable the persistence mechanism for kinesis, which is currently active regardless of cloud pods or persistence being used on the localstack level.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -769,6 +769,7 @@ LAMBDA_JAVA_OPTS = os.environ.get("LAMBDA_JAVA_OPTS", "").strip()
 
 # limit in which to kinesis-mock will start throwing exceptions
 KINESIS_SHARD_LIMIT = os.environ.get("KINESIS_SHARD_LIMIT", "").strip() or "100"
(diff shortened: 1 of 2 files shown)


thrau (commented):
that's a great catch! why not just use `if config.PERSISTENCE` though?

dfangl:
Because then cloud pods will not work anymore for kinesis, I think. For usual payload sizes currently persistence is not a problem, but for bigger ones it becomes one, so I felt like a separate variable to disable it made more sense.

giograno:
Correct, pods won't work if we don't start kinesis already with the correct env variables for persistence (there won't be any assets to persist). It would be very handy to have a sort of dump command to be called ad-hoc in `kinesis mock` 🤔

PR #9873, opened 2023-12-14
Allow merging of config profiles

dfangl:
## Motivation
We are often utilizing config profiles for our testing setups. However, those need quite some duplication for common variables.
This PR will introduce the ability to specify multiple config profiles, which are all loaded.

## Usage
Use a comma separated list of config profiles for the `CONFIG_PROFILE` environment variable.

```
CONFIG_PROFILE=profile1,profile2
```

Identically named environment variables specified in earlier profiles will be overridden by those in later profiles.
Configuration profiles will never override environment variables already set.

If profile1.env looks like this:

```
VAR1=test1
VAR2=test2
VAR3=test3
```

and profile2.env looks like this:

```
VAR3=override3
VAR4=test4
```

the finally set variables will look like this:

```
VAR1=test1
VAR2=test2
VAR3=override3
VAR4=test4
```

## Changes
* Add option to add multiple config profiles
* Add tests

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -249,7 +249,7 @@ def load_environment(profiles: str = None, env=os.environ) -> List[str]:
         if not os.path.exists(path):
             print(f"Profile '{profile}' specified, but file {path} not found.")
             continue
-        environment |= dotenv.dotenv_values(path)
+        environment.update(dotenv.dotenv_values(path))
 
     for k, v in environment.items():
         # we do not want to override the environment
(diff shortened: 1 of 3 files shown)


localstack/config.py
```diff
@@ -230,22 +230,33 @@ def is_env_not_false(env_var_name: str) -> bool:
...
 
-    dotenv.load_dotenv(path, override=False)
-    return profile
+    for profile in profiles:
+        profile = profile.strip()
+        path = os.path.join(CONFIG_DIR, f"{profile}.env")
+        if not os.path.exists(path):
+            print(f"Profile '{profile}' specified, but file {path} not found.")
+            continue
+        environment |= dotenv.dotenv_values(path)
```
thrau, line 252:
```suggestion
        environment.update(dotenv.dotenv_values(path))
```

much easier to read :-)

thrau (approved):
nice feature! :100: :shipit:

PR #9874, opened 2023-12-14
fix S3 service name parsing consuming the stream

bentsku:
## Motivation
As reported in #9867, we have an issue when using S3 pre-signed URL from a browser encoding the request as `application/x-www-form-urlencoded` and targeting a default LocalStack endpoint and not the S3 specific one. The `ServiceNameParser` will try to determine the AWS service, and as S3 is the "fallback" service, it will go through the whole service detection, which includes accessing `request.values`. This will:
- consume the stream
- if you send binary data, as the request is form encoded, Werkzeug will try to parse the form and fail decoding it as `utf-8`. 
```python
  File "/Users/benjaminsimon/Projects/localstack/localstack-ext/.venv/lib/python3.11/site-packages/werkzeug/formparser.py", line 284, in _parse_urlencoded
    stream.read().decode(),
    ^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 0: invalid start byte
```

## Changes
When targeting S3, we have an early handler trying to determine if the request is targeting S3 by checking the bucket name and verifies that it exists. This will allow to actually skip the service name parsing, as we already set the service in the context there.

We also need to have the form parsing around a `try...exception` block in case the bucket does not exist, because the S3 CORS handler will not mark it as an S3 request, and the decoding of the form will fail. We will just swallow the exception and pass so that the service name parser tries the next step. This does not matter that it consumes the stream, because it will lead in an exception anyway.

### DIFF
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -401,7 +401,12 @@ def determine_aws_service_name(request: Request, services: ServiceCatalog = None
             candidates.update(query_candidates)
 
     except RequestEntityTooLarge:
-        LOG.debug("Failed to parse the form while determining AWS service name")
+        # Some requests can be form-urlencoded but also contain binary data, which will fail the form parsing (S3 can
+        # do this). In that case, skip this step and continue to try to determine the service name.
+        LOG.debug(
+            "Failed to parse the form while determining AWS service name",
+            exc_info=LOG.isEnabledFor(logging.DEBUG),
+        )
         pass
 
     # 6. resolve service spec conflicts
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -402,12 +402,12 @@ def determine_aws_service_name(request: Request, services: ServiceCatalog = None
 
     except RequestEntityTooLarge:
         # Some requests can be form-urlencoded but also contain binary data, which will fail the form parsing (S3 can
-        # do this). In that case, skip this step and continue to try to determine the service name.
+        # do this). In that case, skip this step and continue to try to determine the service name. The exception is
+        # RequestEntityTooLarge even if the error is due to failed decoding.
         LOG.debug(
-            "Failed to parse the form while determining AWS service name",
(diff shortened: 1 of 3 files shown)


localstack/aws/protocol/service_router.py
```diff
@@ -374,29 +375,39 @@ def determine_aws_service_name(request: Request, services: ServiceCatalog = None
...
+            candidates.update(query_candidates)
+
+    except RequestEntityTooLarge:
+        # Some requests can be form-urlencoded but also contain binary data, which will fail the form parsing (S3 can
+        # do this). In that case, skip this step and continue to try to determine the service name.
+        LOG.debug(
+            "Failed to parse the form while determining AWS service name",
+            exc_info=LOG.isEnabledFor(logging.DEBUG),
+        )
+        pass
```
thrau, line 410:
```suggestion
        LOG.debug(
            "Failed to determine AWS service from request body because payload was too large",
            exc_info=LOG.isEnabledFor(logging.DEBUG),
        )
```

thrau (approved):
thanks for fixing this! as discussed offline, i think this solution is the best we can do for now!

maybe two things we can add in the future (as discussed also, not needed for this PR)
* get service indicators not only from headers but also from query args to cover pre-signed URLs
* instead of using `request.values` which triggers form parsing, first use `request.args` and then check if the body can be parsed as form (which would perhaps cover some rare pre-signed URL cases, but we should probably first find specific cases)

just have tiny suggestion for a better logging message

localstack/aws/protocol/service_router.py
bentsku: Hmm, I'm not too sure about this because the error was not that the payload was too large, but that it contained binary data. Werkzeug will however raise a 413 exception which indicate the payload was too large, but it was a red herring. Which is why I wanted to log a form parsing failure instead of the size error. Does it make sense?

bentsku: @thrau could `Failed to determine AWS service from request body because the form could not be parsed` be a middle ground?

PR #9875, opened 2023-12-14
add S3 object key validation

bentsku:
## Motivation
Add validation for the S3 key length

## Changes
- add validation in the provider and patch the spec to create the missing exception

## Resources
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html


thrau (approved):
LGTM! :100:

PR #9877, opened 2023-12-15
test get-metric-data with no unit

baermat:
## Motivation
This PR aims at checking what happens when multiple units are present in the cloudwatch database, then no unit is specified in the get-metrics-data call.
The result is that this seems to be not as described [in the docs](https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/get-metric-data.html):
```
In a Get operation, if you omit Unit then all data that was collected with any unit is returned,
along with the corresponding units that were specified when the data was reported to CloudWatch.
If you specify a unit, the operation returns only data that was collected with that unit specified.
If you specify a unit that does not match the data collected, the results of the operation are null.
CloudWatch does not perform unit conversions.
```
AWS however acts differently with a given statistic: When multiple Units are present, AWS seems to simply pick the one that comes "first" in the list of allowed Units, then exclusively operates on values of that unit for calculating the statistic. This list seems to be the following (this was returned when specifying a unit not present here)
```
[Megabits, Terabits, Gigabits, Count, Bytes, Gigabytes, Gigabytes / Second, Kilobytes, Kilobits / Second,
Terabytes, Terabits/Second, Bytes/Second, Percent, Megabytes, Megabits/Second, Milliseconds, Microseconds,
Kilobytes/Second, Gigabits/Second, Megabytes/Second, Bits, Bits/Second, Count/Second, Seconds, Kilobits,
Terabytes/Second, None ].
``` 
A message is then added to the response stating that multiple units were found.

## Changes
Adds test for:
- 2 units, 1 value each
- 2 units, 2 values each
- as before, but with a new unit that comes "prior" to the others in the list

## TODO

What's left to do:

- [ ] Clarify how to proceed. This requires some adjustment in the query code. /cc @steffyP

### DIFF
--- a/tests/aws/services/cloudwatch/test_cloudwatch.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch.py
@@ -2056,6 +2056,91 @@ class TestCloudwatch:
         sleep_before = 2 if is_aws_cloud() else 0.0
         retry(assert_results, retries=retries, sleep=1.0, sleep_before=sleep_before)
 
+    base_metric_data = [
+        {
+            "MetricName": "<metric_name>",
+            "Timestamp": "<now>",
+            "Value": 60000,
+            "Unit": "Milliseconds",
+        },
+        {
+            "MetricName": "<metric_name>",
+            "Timestamp": "<now>",
+            "Value": 60,
+            "Unit": "Seconds",
+        },
+    ]
+    count_metric = {
+        "MetricName": "<metric_name>",
+        "Timestamp": "<now>",
+        "Value": 5,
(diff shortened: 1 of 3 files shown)


steffyP (commented):
Thanks for the extensive testing of this behavior @baermat! 🚀 

This is indeed a very weird behavior, and not what we expected when checking the docs. 
As we discussed offline: This seems to be a very specific edge case, requiring some dirty hacking in our provider. We will tackle this, if it is requested.

As the test was already added a while ago, it seems the validation-file was not yet updated: could you please check 🙏

steffyP (approved):

PR #9880, opened 2023-12-15
Feat/cloudwatch v2 fix labels

baermat:
## Motivation
This PR fixes the label generation in the new cloudwatch provider. It seems to be the case that for certain parameters, whenever we have multiple queries in a `get_metric_data` call that have different values for these parameters (like "sum" and "average" for Stat or "30" and "60" for Period) the values are added to the label.

## Changes
- fixes the label generation in the provider
- adds 7way parametrized test to make sure label generation behaves as expect


steffyP (approved):
Great work @baermat! 🎉 
Smart way of changing the label dynamically! 🚀 
Thank you for the detailed testing.

Just waiting for the pipeline (added a skip as the newly added test case doesn't work with the old provider).

PR #9884, opened 2023-12-15
Allow kinesis get_records shard iterator to have surrounding quotes

dfangl:
## Motivation
Using the AWS CLI, it is easily accomplished to have surrounding double quotes around the Shard Iterator of the `get_records` request.

Even though this is not technically correct, both the CLI and AWS allow this.

This PR will add a test for the behavior on LS side

Kinesis-mock issue: etspaceman/kinesis-mock#682
Upstream PR: etspaceman/kinesis-mock#689

## Changes
* Add test for double quotes within shard iterator

### DIFF
--- a/tests/aws/services/kinesis/test_kinesis.py
+++ b/tests/aws/services/kinesis/test_kinesis.py
@@ -413,6 +413,24 @@ class TestKinesis:
         assert shard_iterator != get_records_response["NextShardIterator"]
         assert new_shard_iterator != get_records_response["NextShardIterator"]
 
+    @markers.aws.validated
+    def test_get_records_shard_iterator_with_surrounding_quotes(
+        self, kinesis_create_stream, wait_for_stream_ready, aws_client
+    ):
+        stream_name = kinesis_create_stream(ShardCount=1)
+        wait_for_stream_ready(stream_name)
+
+        first_stream_shard_data = aws_client.kinesis.describe_stream(StreamName=stream_name)[
+            "StreamDescription"
+        ]["Shards"][0]
+        shard_id = first_stream_shard_data["ShardId"]
+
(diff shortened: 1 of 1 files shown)


tests/aws/services/kinesis/test_kinesis.py
```diff
@@ -413,6 +413,24 @@ def test_get_records_next_shard_iterator(
...
+        first_stream_shard_data = aws_client.kinesis.describe_stream(StreamName=stream_name)[
+            "StreamDescription"
+        ]["Shards"][0]
+        shard_id = first_stream_shard_data["ShardId"]
+
+        shard_iterator = aws_client.kinesis.get_shard_iterator(
+            StreamName=stream_name, ShardIteratorType="LATEST", ShardId=shard_id
+        )["ShardIterator"]
+
+        aws_client.kinesis.get_records(ShardIterator=f'"{shard_iterator}"')
```
simonrw, line 432: Is there a scenario where there might be single quotes? Do we need to cover both cases?

simonrw, line 432: Do we need to validate that we actually get records?

simonrw (approved):
Thanks for adding this regression test, nice and clear. Only really a question about receiving any records, but otherwise :tada:

tests/aws/services/kinesis/test_kinesis.py
dfangl: I did not check whether single quotes would be allowed, but the main issue of the customer was that the aws cli returns it wrapped in double quotes, so this is the most important case for now :)

dfangl: Not really necessary for preventing regression, still a good idea to double check. Added it!

Issue #9486, opened 2023-10-26
bug: The response for CreateHostedZone in route53 misses the DelegationSet


stof:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When calling CreateHostedZone, the response does not contain DelegationSet, even though the API definition files of AWS SDKs describe it as a required field of the response shape.

### Expected Behavior

The response should have the DelegationSet to match the AWS behavior.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run -d -p 4576:4566 -e SERVICES=route53 -v /var/run/docker.sock:/var/run/docker.sock --name async_aws_localstack-route53 localstack/localstack:2.3.2

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

https://github.com/async-aws/aws/blob/516667a7e77905dcf5c2ca470e7ab141c857f466/src/Service/Route53/tests/Integration/Route53ClientTest.php#L81-L100

This is the test that fails when updating localstack from 0.14.2 to 2.3.2 because the response parsing expects the required field to be present.

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack:2.3.2
```

### Anything else?

For reference, here is the response returned when using localstack 2.3.2:

```xml
<?xml version='1.0' encoding='utf-8'?>
<CreateHostedZoneResponse><HostedZone><Id>/hostedzone/KVRMC4FVNEG7536</Id><Name>test-domain.com.</Name><Config><Comment>foo</Comment><PrivateZone>true</PrivateZone></Config><ResourceRecordSetCount>2</ResourceRecordSetCount></HostedZone><ChangeInfo><Id>/change/C1PA6795UKMFR9</Id><Status>INSYNC</Status><SubmittedAt>2017-03-15T01:36:41.958000Z</SubmittedAt></ChangeInfo><VPC><VPCRegion /><VPCId /></VPC></CreateHostedZoneResponse>
```

And here is the response using localstack 0.14.2

```xml
<CreateHostedZoneResponse xmlns="https://route53.amazonaws.com/doc/2012-12-12/">
   <HostedZone>
      <Id>/hostedzone/HPWU1OMIC5LRCWS</Id>
      <Name>test-domain.com.</Name>
      <ResourceRecordSetCount>0</ResourceRecordSetCount>
      <Config>
        
[...]

PR #9885, opened 2023-12-15
Fix: Route53 populating VPC fields

simonrw:
## Motivation

While investigating https://github.com/localstack/localstack/issues/9486 I discovered that `moto` does not populate the VPC information for private hosted zones. This PR both addresses this missing information and adds an AWS validated test for private hosted zones.

*Note*: the existing test `test_associate_vpc_with_hosted_zone` will not work on AWS since creating a private zone without specifying VPC parameters fails with a client error from `boto3`. Rather than updating that test, which tests additional things, I created a new test that focuses purely on the create response.

## Changes

* Update `route53::create_hosted_zone` response to populate the VPC parameters if missing and when creating a private zone
* Add AWS validated test to capture responses
* Update `test_associate_vpc_with_hosted_zone` to use hosted zone fixture to ensure proper teardown

### DIFF
--- a/localstack/services/route53/provider.py
+++ b/localstack/services/route53/provider.py
@@ -39,7 +39,7 @@ class Route53Provider(Route53Api, ServiceLifecycleHook):
         response = call_moto(context)
 
         # moto does not populate the VPC struct of the response if creating a private hosted zone
-        if hosted_zone_config.get("PrivateZone", False):
+        if hosted_zone_config and hosted_zone_config.get("PrivateZone", False):
             response["VPC"]["VPCId"] = response["VPC"]["VPCId"] or vpc["VPCId"]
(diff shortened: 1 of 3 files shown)


giograno (approved):
It looks good, thanks for tackling this!

I have only one general question here: when it comes to these parity fixes do we have a general rule of thumb to address them in LS or in moto?

simonrw:
> It looks good, thanks for tackling this!
> 
> I have only one general question here: when it comes to these parity fixes do we have a general rule of thumb to address them in LS or in moto?

Ideally I think these would be tackled in `moto`, but hopefully I've made the fix backwards compatible so that when `moto` is updated, we can remove this method from our provider. I wanted to try to fix the issue for us first, proving that it is correct before upstreaming it.

PR #9886, opened 2023-12-15
use batches to report SQS metrics for approximate query metrics

steffyP:
## Motivation
Improving the sqs-metrics reporting for the "Approximate*" metrics, by collecting metric-data in batches before sending the data to cloudwatch. 
As a reminder: those are the metrics that are reported regularly (by default once every minute) for all queues.

## Changes
* Collect the data of all queues in slices, before sending the batch to cloudwatch (previously we would call the cloudwatch-api three times for each queue)
* Added a new helper in `cloudwatch_util.py` to handle sqs-batched data

### DIFF
--- a/tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
@@ -123,3 +123,37 @@ class TestCloudWatchLambdaMetrics:
         )
         assert res["MetricDataResults"][0]["Values"] == expected_return
         return res
+
+
+class TestSqsApproximateMetrics:
+    @markers.aws.validated
+    def test_sqs_approximate_metrics(self, aws_client, sqs_create_queue):
+        queue_names = []
+        for _ in range(0, 10):
+            q_name = f"my-test-queue-{short_uid()}"
+            sqs_create_queue(QueueName=q_name)
+            queue_names.append(q_name)
+
+        for queue in queue_names:
+            retry(
+                lambda: self._assert_approximate_metrics_for_queue(
+                    aws_client.cloudwatch, queue_name=queue
+                ),
+                retries=70,  # should be reported every 60 seconds
+                sleep=10 if is_aws() else 1,
+            )
+
+    def _assert_approximate_metrics_for_queue(
+        self,
+        cloudwatch_client: "CloudWatchClient",
+        queue_name: str,
+    ):
+        namespace = "AWS/SQS"
+        dimension = [{"Name": "QueueName", "Value": queue_name}]
+
+        res = cloudwatch_client.list_metrics(Namespace=namespace, Dimensions=dimension)
+        metric_names = [m["MetricName"] for m in res["Metrics"]]
+        assert "ApproximateNumberOfMessagesVisible" in metric_names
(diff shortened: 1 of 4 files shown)


tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
```diff
@@ -123,3 +123,37 @@ def _wait_for_lambda_metric(
...
+                ),
+                retries=70,  # should be reported every 60 seconds on LS
+                sleep=10 if is_aws() else 1,
+            )
+
+    def _assert_approximate_metrics_for_queue(
+        self,
+        cloudwatch_client: "CloudWatchClient",
+        queue_name: str,
+    ):
```
baermat, line 150: stupid question, I have never seen this before: what does the "CloudWatchClient" type do? I couldn't really find anything useful online, and when I enter it like that, pycharm tells me that the type could be Any.

tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
```diff
@@ -123,3 +123,37 @@ def _wait_for_lambda_metric(
...
+            sqs_create_queue(QueueName=q_name)
+            queue_names.append(q_name)
+
+        for queue in queue_names:
+            retry(
+                lambda: self._assert_approximate_metrics_for_queue(
+                    aws_client.cloudwatch, queue_name=queue
+                ),
+                retries=70,  # should be reported every 60 seconds on LS
+                sleep=10 if is_aws() else 1,
```
baermat, line 143: nit: Attempting this 70 times seems odd, especially if LS only reports this after 60s (which seems to be the case). Couldn't we `sleep_before=60` before we start checking this, then have 7 retries? Or does AWS need those 70 10s intervals?

baermat (approved):
Looks mostly good to me! I have two questions that are not blocking (unless my stupid question turns out to be a non-stupid one :smile: )

tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
steffyP: It's a type hint, meaning that the `cloudwatch_client` should be of type `CloudWatchClient`. 

We have a conditional import for the type:
```
if TYPE_CHECKING:
    from mypy_boto3_cloudwatch import CloudWatchClient
```
[Here are some docs](https://docs.python.org/3.10/library/typing.html#typing.TYPE_CHECKING).

tests/aws/services/cloudwatch/test_cloudwatch_metrics.py
steffyP: two things:
1. we don't really know when the scheduler started (it executes the metric-reporting every 60 seconds, but starts when the SQS provider was first called), and how it long it takes to finish the reporting. 70 repeats seemed like a safe choice. but when we are lucky the metrics could already be available within a few seconds.
2. the retry is running in a loop, for each queue_name, so adding a `sleep_before` doesn't make sense either

AWS doesn't really provide any information on the interval they use, when I run tests yesterday it also seemed very random (no relation between metric report time and creation/age of the queue).

PR #9887, opened 2023-12-15
use mounts instead of volumes docker api

thrau:
## Motivation

I tried to mount a volume bind with a colon in the name, which fails both for the CLI and SDK docker client. The reason is explained here https://github.com/docker/docker-py/issues/2041

This can be fixed by using the more flexible `mounts` api parameter. 

> mounts is more modern and a better choice 99% of the time.
-- https://github.com/docker/docker-py/issues/2041#issuecomment-393359696

> In general, --mount is more explicit and verbose. The biggest difference is that the -v syntax combines all the options together in one field, while the --mount syntax separates them
-- https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag

This PR changes both the docker SDK and CMD client to use `mounts` instead of `volumes`/`bind` for mounting volumes.
There is a significant impact though: the `-v` flag would create a source directory on the host if it didn't exist, whereas `--mount` would fail. To make sure we don't break code, we explicitly create the directories now, however they will be created as the user that owns the python process (typically the user), rather than the docker process (typically root).

Open to discuss the implications and how we could mitigate them.

## Changes

* Rework the `VolumeBind` mapping to map to `mounts` instead of `volumes`
* If a volume source didn't exist on the host before spawning a container, the owner of the created directory is now the user running localstack and no longer the user running docker (which would typically be `root`)

### DIFF
--- a/localstack/utils/container_utils/container_client.py
+++ b/localstack/utils/container_utils/container_client.py
@@ -11,7 +11,7 @@ import tempfile
 from abc import ABCMeta, abstractmethod
 from enum import Enum, unique
 from pathlib import Path
-from typing import Dict, List, Literal, NamedTuple, Optional, Protocol, Tuple, Union, get_args
+from typing import Any, Dict, List, Literal, NamedTuple, Optional, Protocol, Tuple, Union, get_args
 
 from localstack import config
 from localstack.utils.collections import HashableList, ensure_list
@@ -682,7 +682,7 @@ class ContainerClient(metaclass=ABCMeta):
         """Returns a blocking generator you can iterate over to retrieve log output as it happens."""
 
     @abstractmethod
-    def inspect_container(self, container_name_or_id: str) -> Dict[str, Union[Dict, str]]:
+    def inspect_container(self, container_name_or_id: str) -> Dict[str, Any]:
         """Get detailed attributes of a container.
 
         :return: Dict containing docker attributes as returned by the daemon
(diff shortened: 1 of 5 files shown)


alexrashed:
Awesome! It would be great to properly assign permissions to the localstack volume directory! 🦸🏽 
Concerning the impact of this: I haven't dug into the specifics of `mount`, but would this approach work with already exising volumes owned and writable only by `root`?
Otherwise, imho this would be a breaking change. This can be easily fixed by removing the directory, but I guess this should either be a migration instruction in a major release, or it should be done automatically...

dfangl:
I rebase this to get a current test run.

My biggest concern with this is the changed behavior of creating non-existent directories on the host. They are captured in the CLI tests I think. 
Another instance that might be problematic is persistence, where we mount a subfolder of the /var/lib/localstack/state directly into a container. We have to make absolutely sure that that will exist before.

I took a quick look into it, and there might be some logical errors there, preventing this change.

dfangl:
I converted this to draft, as we discussed that there are too many open questions to go forward with it right now.

thrau:
closing this PR since as @dfangl said there are too many issues with it potentially breaking environments.
i would keep this in mind though when developing container abstractions /cc @simonrw. 

perhaps we can also refactor some of the existing code docker client code to more clearly separate "volume" and "mount" terminology in method parameters etc, so we can start introducing mounts as an alternative when creating containers.

PR #9900, opened 2023-12-18
Add link-local address and init args to Docker client API

viren-nadkarni:
## Motivation

This PR adds the possibility to specify certain startup and network related flags when using the internal Docker API:

- When starting a container, the `init` option: https://docs.docker.com/engine/reference/commandline/run/#options
- For connecting a container to a network, the link local IP addresses: https://docs.docker.com/engine/reference/commandline/network_connect/#options


dfangl (approved):
LGTM!

Issue #9851, opened 2023-12-12
Second call to S3.UploadPart returns invalid response if first call is cancelled


adam-fowler:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If a call to `S3.UploadPart` is cancelled halfway through and then called again the `Etag` header returned in the second call's response is set to "None".

### Expected Behavior

I would expect the `Etag` header to have a valid Etag in it.

### How are you starting LocalStack?

With a docker run command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

I found this while testing Soto the unofficial Swift SDK for AWS. It is hard to replicate using the CLI as each run of the CLI creates a new multipart upload. You can find the test that fails here https://github.com/soto-project/soto/blob/92623a6943cbbed9e57af80468d318506abce13c/Tests/SotoTests/Services/S3/S3Tests%2Basync.swift#L623. It starts a multipart upload, and cancels it half way through and then calls a resume function which will retry uploading the parts that haven't fully uploaded.

### Environment

```markdown
- OS: MacOS 14.1.2
- LocalStack: 3.0.3.dev
```

bentsku:
Hello @adam-fowler and thanks for your report!

Nice to see LocalStack being used here 😄
I'll try running the tests myself, and fix this issue. Thanks a lot again for reporting!

bentsku:
Hello @adam-fowler and thanks again for the report!
The issue should be fixed in the `latest` image. Don't hesitate to report any issues with LocalStack. I'll give a try to the tests being skipped in your test suite 😄 thanks again!

PR #9901, opened 2023-12-18
fix S3 UploadPart logic when cancelling/error a request

bentsku:
## Motivation
As reported with #9851, we would face an issue when cancelling/error an `UploadPart` request and retrying on the same PartNumber. 
This was due because of our ephemeral MultipartUpload object is not fully stateless, and keeps the `EphemeralS3StoredObject` in a dict to allow easier "fusion" of all underlying parts when completing (it does not have to recreate the `EphemeralS3StoredObject` on top of the `LockedSpooledTemporaryFile`). 

However, this created an issue when the request would error/cancel, because the `EphemeralS3StoredObject` kept the reference to the `S3Part` that errored (created in the provider operation which raised an Exception). When creating a new S3Part in a non-failing provider operation, the asynchronous calculation of the `ETag` due to the `aws-chunked` encoding would not set the `ETag` to the new `S3Part` but to the previously created one, hence the returned `ETag` set to `None`. 

## Changes
The simple fix is to simply remove the part entry for the underlying storage if the writing of that part fails. It could maybe be wrapped in a context manager which would clean up only if the operation fails, but this is simple and clear what it does.

I've also added a test to verify the fix, ran the external `soto` test suite and validated that it now works.

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -1880,7 +1880,11 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         stored_multipart = self._storage_backend.get_multipart(bucket_name, s3_multipart)
         stored_s3_part = stored_multipart.open(s3_part)
-        stored_s3_part.write(body)
+        try:
+            stored_s3_part.write(body)
+        except Exception:
+            stored_multipart.remove_part(s3_part)
+            raise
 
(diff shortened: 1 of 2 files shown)


steffyP (approved):
LGTM 👍 😄 
Thanks for the detailed explanation and adding a new test, which will help detect any regression in the future 🚀

PR #9902, opened 2023-12-18
add reset feature to CW v2

pinzon:
## Motivation
This adds the ability to reset the Cloudwatch v2 state cleaning the database  metrics tables.  
Related to localstack/localstack#9347

## Changes
 - removed unused functions in the cloudwatch_database_helper.py and their calls
 - the shutdown function in the helper implements the query to reset their metrics tables and the index Autoincrement.

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -317,13 +317,13 @@ class CloudwatchDatabase:
             return {"metrics": metrics_result}
 
     def clear_tables(self):
-        # TODO clear tables for reset calls on cloudwatch
-        pass
-
-    def shutdown(self):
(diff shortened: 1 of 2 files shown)


bentsku (commented):
This looks good, awesome to have it implemented. 

Quick question however, what if you call `reset` twice in a row? Are the table recreated before? I think calling `DELETE FROM` on a table which does not exist might raise an exception.

pinzon:
The database is initialized (if it doesn't exist) every time the database helper class is initialized. For now there's no situation where the database exists but the tables doesn't so we could call reset state several times and it should be fine.

bentsku (approved):
> The database is initialized (if it doesn't exist) every time the database helper class is initialized. For now there's no situation where the database exists but the tables doesn't so we could call reset state several times and it should be fine.

Ah right I see it now! thanks, sorry for not seeing it before. Thanks for answering!

PR #9321, opened 2023-10-09
Implement resource provider for AWS::Lambda::Function

dominikschubert:
## Motivation

Part of the ongoing effort to migrate to `ResourceProvider` based resources.

Normally we'd add the whole service in a single PR but since the lambda function is pretty big in itself, I thought it makes more sense to introduce this separately.

## Changes

- Introduces a new resource provider for `AWS::Lambda::Function`. (Partially based on work from @joe4dev  in https://github.com/localstack/localstack/pull/8406)
- Adds support for cfnresponse module in python functions (used for simple custom resource communication). See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html
- Improve support for cfn-response module in NodeJS functions (proper locations for `cfn-response.js` files and optional detection)
- Support for idempotent delete on a function

PR #9903, opened 2023-12-18
Added missing type conversion for CFn update of Lambda::Function MemorySize to int

Morijarti:
## Motivation
User has reported that they are receiving following error during UPDATE operation in CFn

`Invalid type for parameter MemorySize, value: 256, type: <class 'str'>, valid types: <class 'int'>`

## Changes
Added type conversion of MemorySize to int in LegacyModel of CFn providers and provided comments to https://github.com/localstack/localstack/pull/9321 to avoid bug recurring when switching to new providers.


dominikschubert (approved):
Thanks for jumping on it. 
Please note that this generally will need to be covered via a test in the future but for now, I think it's fine (given it's an UPDATE operation).

PR #9904, opened 2023-12-18
add cli option to restart localstack container

HarshCasper:
## Motivation

This PR exposes a CLI option to restart the LocalStack runtime, which is often required for development/testing/running demos.

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -57,7 +57,9 @@ class LocalStackCliGroup(click.Group):
             if ctx and ctx.params.get("debug"):
                 click.echo(traceback.format_exc())
             from localstack.utils.container_utils.container_client import (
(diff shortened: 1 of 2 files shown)


thrau (commented):
This is a great addition! Instead of restarting the whole container by using the DOCKER_CLIENT, we could use the health endpoint, which has a special restart function that restart localstack within the container. This might be nice because it doesn't require a container restart!

You can invoke it like this:

```python

from localstack import config


def main():
    response = requests.post(
        f"{config.internal_service_url()}/_localstack/health",
        json={"action": "restart"},
    )
```

it would have a similar structure like `cmd_status_services` which also calls the health endpoint.

WDYT?

thrau (approved):
very useful addition! in the tests it would have been good to assert that localstack was actually restarted correctly :-) but it's fine for now

PR #9906, opened 2023-12-18
Bump Kinesis Mock to 0.4.5

etspaceman:
https://github.com/etspaceman/kinesis-mock/releases/tag/v0.4.5

## Motivation

Solves https://github.com/localstack/localstack/pull/9884

## Changes

Bumps kinesis-mock to 0.4.5

### DIFF
--- a/localstack/services/kinesis/packages.py
+++ b/localstack/services/kinesis/packages.py
@@ -5,7 +5,7 @@ from typing import List
 from localstack.packages import Package, PackageInstaller
 from localstack.packages.core import NodePackageInstaller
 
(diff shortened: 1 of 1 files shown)


alexrashed (commented):
Thanks as always for pro-actively updating the version after implementing the fix!
The failing tests are caused by:
- The fork not having the permissions to execute secrets (to execute the pro tests).
- An issue with the upgrade of dynamodblocal, which is addressed in https://github.com/localstack/localstack/pull/9910.

From my point of view this PR is looking good and could be merged, but I have limited availability today to take a closer look. So I'll leave it to @dfangl to review / approve / merge. Thanks!

PR #9909, opened 2023-12-19
Upgrade and pin DynamoDB Local version to 2.2.0 and update snapshot

dominikschubert:
## Motivation

We've started seeing repeatable failures in CI due to a snapshot mismatch for `tests/aws/services/cloudformation/resources/test_lambda.py::TestCfnLambdaIntegrations::test_cfn_lambda_dynamodb_source`.

```
__________ TestCfnLambdaIntegrations.test_cfn_lambda_dynamodb_source ___________
>> match key: describe_table_result
	(+)  /Table/DeletionProtectionEnabled ( False )

```

My investigation pointed towards a new update of DynamoDB local as the culprit ([AWS recently released a 2.2.0 version with support for Table delete protection](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocalHistory.html))

## Changes

- Download DynamoDB Local artifact from maven instead of AWS so that we can pin the downloaded version to avoid PRs like this in the future and can upgrade in more controlled matter. (e.g. driven by issues detected by snapshot updates against AWS).
- Add `2.1.0`, `2.2.0` and a `latest` (previous default) to the package
- Set default version to the new `2.2.0` version
- Updated snapshot of affected test

## TODO

- [ ] dependencies are not bundled in the jar file, so we'd need to bundle them separately if we want to continue with this solution  in the future

### DIFF
--- a/localstack/services/dynamodb/packages.py
+++ b/localstack/services/dynamodb/packages.py
@@ -35,6 +35,9 @@ class DynamoDBLocalPackage(Package):
         return DynamoDBLocalPackageInstaller(version)
 
     def get_versions(self) -> List[str]:
+        """
+        https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocalHistory.html
+        """
         return ["2.1.0", "2.2.0"]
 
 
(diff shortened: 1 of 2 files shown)


viren-nadkarni:
Superseded by #11064

PR #9910, opened 2023-12-19
Update dynamodb snapshot to unblock pipeline

dominikschubert:
## Motivation

We've started seeing repeatable failures in CI due to a snapshot mismatch for `tests/aws/services/cloudformation/resources/test_lambda.py::TestCfnLambdaIntegrations::test_cfn_lambda_dynamodb_source`.

```
__________ TestCfnLambdaIntegrations.test_cfn_lambda_dynamodb_source ___________
>> match key: describe_table_result
	(+)  /Table/DeletionProtectionEnabled ( False )

```

My investigation pointed towards a new update of DynamoDB local as the culprit ([AWS recently released a 2.2.0 version with support for Table delete protection](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocalHistory.html))

Was running into some issues with the initial PR in https://github.com/localstack/localstack/pull/9909

I still want to tackle the problem with a more long-term solution to avoid these unpredictable dependency update but for now this should unblock the pipeline.

## Changes

- Update snapshot for `tests/aws/services/cloudformation/resources/test_lambda.py::TestCfnLambdaIntegrations::test_cfn_lambda_dynamodb_source`

### DIFF
--- a/localstack/services/dynamodb/server.py
+++ b/localstack/services/dynamodb/server.py
@@ -159,6 +159,7 @@ class DynamodbServer(Server):
             log_listener=_log_listener,
             auto_restart=True,
             name="dynamodb-local",
+            env_vars={"DDB_LOCAL_TELEMETRY": "0"},
         )
         TMP_THREADS.append(t)
         t.start()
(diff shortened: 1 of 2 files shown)


simonrw (approved):
I think this is a fair change for now. I appreciate you trying to resolve the problem in a more complete way, but with this change we can implement the "proper" solution in our own time.

localstack/services/dynamodb/server.py
```diff
@@ -159,6 +159,7 @@ def do_start_thread(self) -> FuncThread:
             log_listener=_log_listener,
             auto_restart=True,
             name="dynamodb-local",
+            env_vars={"DDB_LOCAL_TELEMETRY": "0"},
```
viren-nadkarni, line 162: Good one :+1:

viren-nadkarni (approved):

PR #9913, opened 2023-12-19
Remove error message if environment does not exist

dfangl:
## Motivation
Originally, with #9873 , I added an error message if a config profile cannot be found. However, this interfered with the usages of the config in CLI tools where the output might get parsed, like lpm, since it is always the case due to the default profile.

Since this error message has quite significant impact, it is best for now to remove it, and find out a better way to debug profile loading should the need arise.

I also thought about only suppressing it in the default case, but the potential impact would still outweigh the benefits in my opinion.

## Changes
* Remove error output printed to stdout in case profile does not exist.


alexrashed (approved):
Thanks for jumping on this! 🦸‍♀️💯

PR #9915, opened 2023-12-19
fix CFN RequestValidator parameters being optional

bentsku:
## Motivation
We've seen a regression from #9320 where optional parameters were accessed directly from the model in CloudFormation. Those parameters are marked as optional, and will default to `False`.
See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-requestvalidator.html

## Changes
Access those parameters with `.get` instead of direct attribute access, and make default to `False`.
Update a test to not specify a `False` parameter to check that it still properly create the resource and default to `False`.

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_apigateway.py
+++ b/tests/aws/services/cloudformation/resources/test_apigateway.py
@@ -212,7 +212,13 @@ def test_cfn_with_apigateway_resources(deploy_cfn_template, aws_client, snapshot
 
 @markers.aws.validated
 @markers.snapshot.skip_snapshot_verify(
-    paths=["$.get-resources.items..resourceMethods.ANY"]  # TODO: empty in AWS
+    paths=[
+        "$.get-resources.items..resourceMethods.ANY",  # TODO: empty in AWS
+        "$..requestParameters",  # FIXME: it seems AWS does not return empty dicts anymore, will need to fix
+        "$..responseTemplates",  # FIXME: it seems AWS does not return empty dicts anymore, will need to fix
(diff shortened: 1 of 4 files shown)


tests/aws/services/cloudformation/resources/test_apigateway.py
```diff
@@ -212,7 +212,13 @@ def test_cfn_with_apigateway_resources(deploy_cfn_template, aws_client, snapshot
 
 @markers.aws.validated
 @markers.snapshot.skip_snapshot_verify(
-    paths=["$.get-resources.items..resourceMethods.ANY"]  # TODO: empty in AWS
+    paths=[
+        "$.get-resources.items..resourceMethods.ANY",  # TODO: empty in AWS
+        "$..requestParameters",  # FIXME: it seems AWS does not return empty dicts anymore, will need to fix
+        "$..responseTemplates",  # FIXME: it seems AWS does not return empty dicts anymore, will need to fix
```
dominikschubert, line 218: :face_exhaling:  of course. guess that will affect a lot of apigw snapshots

dominikschubert (approved):
Good catch! Thanks for jumping on it and covering it via the test adaptation.

tests/aws/services/cloudformation/resources/test_apigateway.py
bentsku: Yes, I'll take some time next week to try to regenerate most of them and adapt the code, mostly moto patches, oh joy 😄

PR #9917, opened 2023-12-19
Test lambda performance and mitigate async invoke timeout error (#9917)

joe4dev:
## Motivation

Customers reported connection errors and read timeouts within Lambda pollers such as:

> l.s.l.i.event_manager      : Error while polling lambda events for function arn:aws:lambda:us-east-1:000000000000:function:MyLambdaFunction:$LATEST: Read timeout on endpoint URL: "http://localhost:4566/"

We have also noticed similar errors in some CI runs but could not reproduce them. An extended Lambda performance test suite helps to reproduce this issue and improve the reliability and scalability of event invokes.

### Testing

* Extend the lambda performance testing suite, adding 5 scenarios
  * `test_number_of_function_versions`: Test how many function versions LocalStack can support
  * `test_number_of_functions`: Test how many active functions LocalStack can support
  * `test_lambda_event_invoke`: Test concurrent Lambda event invokes and validate the number of Lambda invocations using CloudWatch and S3
  * `test_lambda_event_source_mapping_sqs`: Test SQS => Lambda event source mapping with concurrent event invokes and validates the number of invocations.
  * `test_sns_subscription_lambda`: Test SNS => Lambda subscription with concurrent event invokes and validates the number of invocations.

### Main changes

* Optimize the internal botocore client configuration for Lambda async pollers. Increased timeouts (+retries) should fix connection errors because they were too short before (initially to avoid delaying the LocalStack shutdown but the improved shutdown order omits this restriction).
* Switch from long polling to short polling in the event manager to reduce the number of open connections

### Auxillary changes

* [minor] Increase the default `LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT` from 10 to `20` because Lambda startup often needs more time when deploying functions with larger or many files and slow I/O. This also mitigates flaky tests due to slow I/O.
* Implement batching in the event manager reduces the number of poller invocations to schedule async invokes
* Improve Lambda logging:
  * Additional DEBUG logs with details including the function name and request id help to track asynchronous event invokes
  * Prefixed loggers with extra context help to track concurrent invocation logs (instead of just `>`)
* Cache the lambda-runtime-init check and perform it earlier (i.e., upon version preparation) instead of for every invoke. This also reduces unnecessary log spam upon every invoke

## Testing

* Ext-CI run against this branch: https://github.com/localstack/localstack-ext/actions/runs/7299151568

Prerequisite: a [LocalStack development environment](https://docs.localstack.cloud/contributing/development-environment-setup/) with Python 3.11.

Run the tests against a running LocalStack instance (by default running on http://localhost:4566)

1. Clone LocalStack and checkout `test-lambda-performance`
2. `make install`
3. `TEST_SKIP_LOCALSTACK_START=1 TEST_PERFORMANCE=1 TEST_PATH="tests/aws/services/lambda_/test_lambda_performance.py::test_lambda_event_source_mapping_sqs" make test`

## Limitations

* The current CloudWatch implementation has concurrency issues and the `Invocations` metric is very unreliable for detecting the number of actual Lambda invocations. We are working on a new CloudWatch implementation that should fix this issue.
* The internal SQS queues in the service account emit CloudWatch metrics, which break AWS parity and use unnecessary resources. Consider skipping metrics for internal service resources and/or internal requests.
* The thread management within asynchronous Lambda invocations needs consolidation to avoid spawning excessive threads. Spawning too many threads can use up connection pools and lead to unnecessary resource contention at scale. This should be tackled as part of the Lambda event source mappings rework.
* The current test suite version uses potentially flaky periods of sleep to avoid putting extra load on LocalStack during test execution using polling. However, poll-based checks are available in the tests and should be re-integrated, potentially in combination with periods of sleep to improve the reliability of the tests.

## TODO

What's left to do:

- [ ] Update docs to reflect updated `LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT` => 20 is the new default

(see [this Notion](https://www.notion.so/localstack/Test-lambda-parallelism-and-batch-processing-scenarios-b99e85aaf06f4575aeadb55dc2fef00e?pvs=4) task for further internal details)

### DIFF
--- a/localstack/services/lambda_/invocation/docker_runtime_executor.py
+++ b/localstack/services/lambda_/invocation/docker_runtime_executor.py
@@ -409,6 +409,8 @@ class DockerRuntimeExecutor(RuntimeExecutor):
     @classmethod
     def prepare_version(cls, function_version: FunctionVersion) -> None:
         lambda_hooks.prepare_docker_executor.run(function_version)
+        # Trigger the installation of the Lambda runtime-init binary before invocation and
+        # cache the result to save time upon every invocation.
         get_runtime_client_path()
         if function_version.config.code:
             function_version.config.code.prepare_for_execution()
--- a/localstack/services/lambda_/invocation/docker_runtime_executor.py
+++ b/localstack/services/lambda_/invocation/docker_runtime_executor.py
@@ -2,6 +2,7 @@ import dataclasses
 import json
 import logging
 import shutil
+from functools import cache
 from pathlib import Path
 from typing import Callable, Dict, Literal, Optional
 
@@ -164,6 +165,9 @@ class RuntimeImageResolver:
 resolver = RuntimeImageResolver()
 
 
+# TODO: handle architecture-specific installer and caching because we currently assume that the lambda-runtime-init
+#   Golang binary is cross-architecture compatible.
+@cache
 def get_runtime_client_path() -> Path:
     installer = lambda_runtime_package.get_installer()
     installer.install()
@@ -405,6 +409,7 @@ class DockerRuntimeExecutor(RuntimeExecutor):
     @classmethod
     def prepare_version(cls, function_version: FunctionVersion) -> None:
         lambda_hooks.prepare_docker_executor.run(function_version)
+        get_runtime_client_path()
         if function_version.config.code:
             function_version.config.code.prepare_for_execution()
             image_name = resolver.get_image_for_runtime(function_version.config.runtime)
--- a/tests/aws/services/lambda_/test_lambda_performance.py
+++ b/tests/aws/services/lambda_/test_lambda_performance.py
@@ -98,7 +98,98 @@ class ThreadSafeCounter:
 
 
 @markers.aws.unknown
+def test_max_number_of_function_versions(create_lambda_function, s3_bucket, aws_client):
+    num_function_versions = 30000
+
+    function_name = f"echo-func-{short_uid()}"
+    create_lambda_function(
+        handler_file=TEST_LAMBDA_PYTHON_S3_INTEGRATION,
+        func_name=function_name,
+        runtime=Runtime.python3_12,
+        Environment={"Variables": {"S3_BUCKET_NAME": s3_bucket}},
+    )
+
+    for _ in range(num_function_versions):
+        aws_client.lambda_.publish_version(FunctionName=function_name)
+
+    # idle for a while to see if the pollers can manage the load and any errors occur
+    time.sleep(90)
+
+    result_1 = aws_client.lambda_.invoke(
+        FunctionName=function_name,
+        InvocationType=InvocationType.RequestResponse,
+    )
+    assert "FunctionError" not in result_1
+    request_id_1 = result_1["ResponseMetadata"]["RequestId"]
+
+    result_2 = aws_client.lambda_.invoke(
+        FunctionName=function_name,
+        InvocationType=InvocationType.Event,
+    )
+    assert "FunctionError" not in result_2
+    request_id_2 = result_2["ResponseMetadata"]["RequestId"]
+
+    # wait to complete the event invoke without causing extra load through polling
+    time.sleep(10)
+    s3_request_ids = get_s3_keys(aws_client, s3_bucket)
+    assert set(s3_request_ids) == {request_id_1, request_id_2}
+
+
+@markers.aws.unknown
+def test_max_number_of_functions(create_lambda_function, s3_bucket, aws_client, aws_client_factory):
+    num_functions = 150
+    uuid = short_uid()
+
+    for num in range(num_functions):
+        function_name = f"echo-func-{uuid}-{num}"
+        create_lambda_function(
+            handler_file=TEST_LAMBDA_PYTHON_S3_INTEGRATION,
+            func_name=function_name,
+            runtime=Runtime.python3_12,
+            Environment={"Variables": {"S3_BUCKET_NAME": s3_bucket}},
+        )
+
+    # idle for a while to see if the pollers can manage the load and any errors occur
+    time.sleep(30)
+
+    request_ids = []
+    for num in range(num_functions):
+        function_name = f"echo-func-{uuid}-{num}"
+        result_1 = aws_client.lambda_.invoke(
+            FunctionName=function_name,
+            InvocationType=InvocationType.RequestResponse,
+        )
+        assert "FunctionError" not in result_1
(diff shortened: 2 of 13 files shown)


tests/aws/services/lambda_/test_lambda_performance.py
```diff
@@ -81,6 +103,520 @@ def invoke():
...
+    # Validate S3 object creation
+    def assert_s3_objects():
+        s3_keys_output = get_s3_keys(aws_client, s3_bucket)
+        # assert len(s3_keys_output) == num_invocations
+        return len(s3_keys_output)
+
+    s3_count = assert_s3_objects()
+
+    # TODO: the CloudWatch metric count for `Invocations` seems very unreliable => double-check API and implementation
+    assert [metric_count, log_count, s3_count] == [
```
joe4dev, line 460: @pinzon @steffyP For 2000 async invokes, I'm getting metrics like `[10.0, 1934, 1942]` 😬 

Hence, only `10.0` of 2k concurrent `Invocations` are tracked in CloudWatch v1. Do you have any idea why (user error/wrong query, concurrency issue)?

joe4dev: FYI: concurrency is an issue (but I assume not that bad)

```
2023-12-15T17:26:53.250  INFO --- [sqs-metrics-cloudwatch-dispatcher_1] localstack.utils.cloudwatch.cloudwatch_util : Unable to put metric data for metric "NumberOfEmptyReceives" to CloudWatch: An error occurred (InternalError) when calling the PutMetricData operation (reached max retries: 0): exception while calling cloudwatch with unknown operation: Traceback (most recent call last):
  File "/Users/joe/Projects/LocalStack/localstack/localstack/aws/chain.py", line 90, in handle
  File "/Users/joe/Projects/LocalStack/localstack/localstack/aws/handlers/routes.py", line 27, in __call__
    router_response = self.router.dispatch(context.request)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joe/Projects/LocalStack/localstack/localstack/http/router.py", line 440, in dispatch
    handler, args = matcher.match(get_raw_path(request), method=request.method)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joe/Projects/LocalStack/localstack-ext/.venv/lib/python3.11/site-packages/werkzeug/routing/map.py", line 582, in match
    self.map.update()
  File "/Users/joe/Projects/LocalStack/localstack-ext/.venv/lib/python3.11/site-packages/werkzeug/routing/map.py", line 365, in update
    for rules in self._rules_by_endpoint.values():
RuntimeError: dictionary changed size during iteration
```

tests/aws/services/lambda_/test_lambda_performance.py
```diff
@@ -81,6 +103,520 @@ def invoke():
...
+    LOG.info(f"N={num_invocations} took {diff.total_seconds()} seconds")
+    assert error_counter.counter == 0
+
+    # Sleeping here is a bit hacky, but we want to avoid polling for now because polling affects the results.
+    sleep_seconds = 200
+    LOG.info(f"Sleeping for {sleep_seconds} ...")
+    time.sleep(sleep_seconds)
+
+    # Validate CloudWatch invocation metric
+    def assert_cloudwatch_metric():
```
joe4dev, line 281: @steffyP @pinzon Is this calculation of the CloudWatch metric `Invocations` correct?
I'm getting numbers that are way off (see assert below)

What's the status of v2? Does it support such a metric and query?

localstack/services/lambda_/invocation/docker_runtime_executor.py
```diff
@@ -405,6 +409,7 @@ def get_logs(self) -> str:
     @classmethod
     def prepare_version(cls, function_version: FunctionVersion) -> None:
         lambda_hooks.prepare_docker_executor.run(function_version)
+        get_runtime_client_path()
```
dfangl, line 412: We might want to add a comment there explaining why we do this here

dfangl (approved):
I think the actual provider changes are a good idea.
More logging, short-polling to reduce connection count, higher timeouts.

We might be able to improve the tests in cases, but for now let's get out the actual provider changes :)

localstack/services/lambda_/invocation/docker_runtime_executor.py
joe4dev: agree, done

joe4dev: sidenote: as long as we respect the function states and do not perform an invoke while the function is still pending, we should be protected from race conditions where two downloads are triggered simultaneously. This is currently a non-issue because the runtime-init is baked into the image.

joe4dev:
Logging of internal resource requests pollutes the logs quite heavily. Every 2 seconds, we get:

```
2023-12-22T17:48:30.587 DEBUG --- [ asgi_gw_190] localstack.aws.serving.wsgi : POST localhost:4566/
2023-12-22T17:48:30.587 DEBUG --- [ asgi_gw_190] localstack.aws.protocol.serializer : No accept header given. Using request's Content-Type (application/x-amz-json-1.0) as preferred response Content-Type.
2023-12-22T17:48:30.588  INFO --- [ asgi_gw_190] localstack.request.internal.aws : AWS sqs.ReceiveMessage => 200
```

This makes the logs barely usable and we should fix that quickly.

tests/aws/services/lambda_/test_lambda_performance.py
steffyP: yes, concurrency is a known issue with the current moto fallback implementation of cloudwatch.

tests/aws/services/lambda_/test_lambda_performance.py
steffyP: simple queries like this one should work :) but actually they should also work the current implementation aka v1.

joe4dev: Thanks @steffyP 

The main issue was a too-optimistic waiting time and `end_time`. [This code change](https://github.com/localstack/localstack/pull/9917/commits/42d2dc9cea1896e52d1a3c64e7a055a041ea4991#diff-c626c358a191e4cc10999a6ed5f301e2f5a7eca9d379e91731268810803e3a8cR291) made the metric more reliable.

PR #9919, opened 2023-12-19
add more tests for cw v2

pinzon:
## Motivation
This PR addresses an essential aspect by introducing tests that specifically target unique and critical scenarios within Cloudwatch. The motivation behind this effort is to enhance the robustness and reliability of our codebase, ensuring comprehensive test coverage for cases that are both distinctive and significant.

## Changes / Testing
- Implemented a test case for scenarios where the query lacks dimensions, but the data contains them.
- Adjusted the database helper to ensure the success of the first test.
- Added a test to cover scenarios where the query results in no data.

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -247,6 +247,9 @@ class CloudwatchDatabase:
                 dimension_filter += "AND dimensions LIKE ? "
                 data = data + (f"%{dimension.get('Name')}={dimension.get('Value','')}%",)
 
+            if not dimensions:
+                dimension_filter = "AND dimensions = '' "
+
             unit_filter = ""
             if unit:
                 if unit == "NULL_VALUE":
@@ -254,6 +257,7 @@ class CloudwatchDatabase:
                 else:
                     unit_filter = "AND unit = ? "
                     data += (unit,)
+
 
             sql_query = f"""
             SELECT
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -258,7 +258,6 @@ class CloudwatchDatabase:
                     unit_filter = "AND unit = ? "
                     data += (unit,)
 
-
             sql_query = f"""
             SELECT
                 {STAT_TO_SQLITE_AGGREGATION_FUNC[stat]},
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -248,7 +248,7 @@ class CloudwatchDatabase:
                 data = data + (f"%{dimension.get('Name')}={dimension.get('Value','')}%",)
 
             if not dimensions:
-                dimension_filter = "AND dimensions = '' "
+                dimension_filter = "AND dimensions is null "
 
             unit_filter = ""
             if unit:
--- a/tests/aws/services/cloudwatch/test_cloudwatch.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch.py
@@ -2189,6 +2189,51 @@ def _get_lambda_logs(logs_client: "CloudWatchLogsClient", fn_name: str):
     filtered_logs.sort(key=lambda e: e["timestamp"], reverse=True)
     return filtered_logs[0]["message"]
 
+    @markers.aws.validated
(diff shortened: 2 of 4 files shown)


localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -247,6 +247,9 @@ def get_metric_data_stat(
                 dimension_filter += "AND dimensions LIKE ? "
                 data = data + (f"%{dimension.get('Name')}={dimension.get('Value','')}%",)
 
+            if not dimensions:
+                dimension_filter = "AND dimensions is null "
+
```
baermat, line 252: It would be interesting to see if you can actually filter for data without dimensions this way (does something like that even exist?), and if AWS handles this case as implied by this change. Then again, I guess we have to finish our fringe case testing at some point. So I will leave this to @steffyP if we need another test case here or not (or if it is easy we could just write one)

baermat (commented):
Nice, more coverage! Mostly LGTM, but there is one change which might need another test? Unsure, see comment (hence no approval yet sorry 😬 )

tests/aws/services/cloudwatch/test_cloudwatch.validation.json
```diff
@@ -20,6 +20,9 @@
   "tests/aws/services/cloudwatch/test_cloudwatch.py::TestCloudwatch::test_enable_disable_alarm_actions": {
     "last_validated_date": "2023-09-12T10:00:45+00:00"
   },
+  "tests/aws/services/cloudwatch/test_cloudwatch.py::TestCloudwatch::test_get_metric_with_null_dimensions": {
+    "last_validated_date": "2024-01-09T20:13:11+00:00"
+  },
```
steffyP, line 25: for some reason the validation date for `test_get_metric_with_no_results` is missing - did you forget to add it to the commit?

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -2211,6 +2227,51 @@ def log_group_exists():
...
+                        },
+                    }
+                ],
+                StartTime=utc_now - timedelta(seconds=60),
+                EndTime=utc_now + timedelta(seconds=60),
+            )
+            snapshot.add_transformer(snapshot.transform.key_value("Label"))
+            snapshot.match("result", data)
+
+        retry(validate, sleep_before=(2 if is_aws_cloud() else 0))
```
steffyP, line 2273: This test is a bit tricky, because we try to verify the negative test case, where no data is expected to be returned. For this kind of tests it's always hard to know how long we should wait (as we expect the data to be unchanged, e.g. always return empty).

With the current implementation, the `retry` doesn't do anything because we don't have a validation in there. 

I would suggest improving this test by also adding a `get-metric-data` query where the metric would be returned (verify this first), and then we can check that the other `get-metric-data` call with the dimensions does not return the result.

steffyP (changes requested):
Thanks so much for putting more work into dimensions test, @pinzon! 🚀 

For the `test_get_metric_with_no_results`, please check my comments for improving the test :) 

There are also 4 commits in your PR that don't belong there (leftovers from rebase I assume). Please make sure to clean up the history 🙂

tests/aws/services/cloudwatch/test_cloudwatch.validation.json
baermat: My guess for that was that the snapshot was simply created before the new validation system was on master. Which seemed fine to me, since we have a bunch of snapshots like that right now. The other test is in the validation file here because due to a hiccup yesterday it had to be re-generated

steffyP: @baermat: Actually I thought we should have validation files for all existing snapshots, as this was added with https://github.com/localstack/localstack/pull/9928. Maybe some slipped through because they were merged in PRs in between.

baermat: ah my bad, then disregard please, and yes that should be re-generated then

tests/aws/services/cloudwatch/test_cloudwatch.py
pinzon: Thanks for the comment. I went with `list-metrics` op to check when the test metrics appear and then do the snapshot.

steffyP (approved):
Nice! Thanks for jumping on the requested changes so quickly 🙇‍♀️ 🔝 
Let's merge after the tests are green!

Issue #9711, opened 2023-11-22
bug: EventBridge `anything-but` event pattern does not match null field


aaronrenner:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the following `anything-but` event pattern on localstack, I'm getting some inconsistencies with the actual eventbridge behavior.

**Event Rule**
```json
{
  "detail": {
    "foo": [{
      "anything-but": ["baz"]
    }]
  }
}
```
**Event that matches on AWS and LocalStack**
```json
{
  "id": "42c61f75-5cf5-4aa8-b430-6fd3d0d8a404",
  "account": "000000000000",
  "source": "myapp",
  "time": "2023-11-22T20:27:11Z",
  "region": "us-east-1",
  "resources": [],
  "detail-type": "MyEvent",
  "detail": { "foo": "a" }
}
```
**Event that matches in AWS, but NOT Localstack**
```json
{
  "id": "42c61f75-5cf5-4aa8-b430-6fd3d0d8a404",
  "account": "000000000000",
  "source": "myapp",
  "time": "2023-11-22T20:27:11Z",
  "region": "us-east-1",
  "resources": [],
  "detail-type": "MyEvent",
  "detail": { "foo": null }
}
```

### Expected Behavior

Both event examples should match in Localstack like they do in AWS.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Docker-compose.yml

```yaml
version: "3.9"

services:
  localstack:
    image: "localstack/localstack"
    ports:
      - "4566:4566"
```
   
#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

Setup commands

    awslocal events create-event-bus \
        --name detail-bus-1
    
    awslocal sqs create-queue \
        --queue-name "detail-sqs-1"
    
    awslocal events put-rule \
        --name "detail-rule-1" \
        --event-bus-name "detail-bus-1" \
        --event-pattern '{ "detail": { "foo": [ {"anything-but": ["baz"]} ] } }'
    
    awslocal events put-targets \
        --rule "detail-rule-1" \
        --event-bus-name "detail-bus-1" \
        --targets "Id"="detail-target-rule-1","Arn"="arn:aws:sqs:us-east-1:000000000000:detail-sqs-1"

Send event that matches

[...]

PR #9925, opened 2023-12-21
Fix: EventBus: match null value for anything-but rule

maxhoheiser:
## Motivation
As reported in [#9711](https://github.com/localstack/localstack/issues/9711) the current behavior does not fully match the aws cli behavior regarding the EventBridge pattern `anything-but`  and a `null` value. 

This rule:
```
{
	"source": ["core.update-account-command"],
	"detail-type": ["core.update-account-command"],
	"detail": {"command": [{"anything-but": ["no-message"]}]},
}
``` 

In combination with this message:
```
{
	"version": "0",
	"id": "<uuid:2>",
	"detail-type": "core.update-account-command",
	"source": "core.update-account-command",
	"account": "111111111111",
	"time": "date",
	"region": "<region>",
	"resources": [],
	"detail": {
	    "command": null,
	    "payload": "baz"
	}
}
```

does not match on LocalStack, but does correctly match on AWS.

The issue arises from using the default fallback of the getter while matching event patterns, that also resolves to `None` and thus fields with `None` values are also excluded.

## Changes
Replace the default fallback while matching event patterns with a custom string.

Potential pitfalls: in the rare circumstance, that the custom string is also used as a value in a message, the pattern matching will not have the desired outcome.  

## Testing

To extend snapshot testing for event messages, the custom fixture that handles setting up the perquisites, posting the message, retrieving the message and comparing it was extended to return the message content.
Although comparing expected and actual massage values is still handled inside the fixture, additional snapshot matching is implemented to ensure AWS parity.

### DIFF
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -516,9 +516,9 @@ def filter_event_based_on_event_format(
 ):
     def filter_event(event_pattern_filter: Dict[str, Any], event: Dict[str, Any]):
         for key, value in event_pattern_filter.items():
-            # match keys in the event in a case-agnostic way
-            event_value = event.get(key.lower(), event.get(key))
-            if event_value is None:
+            fallback = "NOT_FOUND_IN_DICT"
+            event_value = event.get(key.lower(), event.get(key, fallback))
+            if event_value is fallback:
                 return False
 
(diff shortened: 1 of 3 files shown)


dfangl (approved):
Looking good to me, including the test, nice work!

PR #9926, opened 2023-12-21
Cfn: Fix nested stack name generation

viren-nadkarni:
## Summary

This PR fixes a bug where the internal name generation used request ARN instead of name which caused invalid Cfn stack names to be generated leading to CDK deployment errors which were caught by Lambda provider input validation:

```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the CreateFunction operation: 1 validation error detected: Value 'arn:aws:cloudformation:us-eas-LogRetentionaae0aa3c5b4d-25486f2c' at 'functionName' failed to satisfy constraint: Member must satisfy regular expression pattern: (arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{2}((-gov)|(-iso(b?)))?-[a-z]+-\\d{1}:)?(\\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\\$LATEST|[a-zA-Z0-9-_]+))?
```

## Tests

Amended an existing test to assert that stack name does not contain invalid chars (`:` which is found in ARNs). Revalidated test against AWS.

## Related

https://github.com/localstack/localstack/pull/9449

### DIFF
--- a/localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
+++ b/localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py
@@ -62,7 +62,7 @@ class CloudFormationStackProvider(ResourceProvider[CloudFormationStackProperties
         if not request.custom_context.get(REPEATED_INVOCATION):
             if not model.get("StackName"):
                 model["StackName"] = util.generate_default_name(
-                    request.stack_id, request.logical_resource_id
+                    request.stack_name, request.logical_resource_id
                 )
 
             create_params = util.select_attributes(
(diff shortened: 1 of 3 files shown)


dominikschubert (changes requested):
Since this seems to have caused an issue already, I'd like this to be covered with a test. @viren-nadkarni  do you have  a template to reproduce the issue?

viren-nadkarni:
@dominikschubert We only have customer provided CDK files, see `support_mjanda_u034ce7sxs4`. My knowledge of CDK is non-existent so it's unreadable to me :sweat_smile:

dominikschubert (approved):
LGTM, thanks for adding the test as well :raised_hands:

PR #9927, opened 2023-12-21
Track date of last passed validation run of tests targeting AWS

dominikschubert:
## Motivation

When creating tests we validate that they are passing successfully against AWS. This can change over time and so we need a way to have a sensible way to periodically re-evaluate this validation status.  In the future we can do this e.g. on a schedule where we re-validate any tests that haven't been validated in the last x (e.g. 12) months.

Earlier this year we started a similar effort on keeping snapshots up to date to avoid recurring issues where changes on AWS' side cause undetected issues. We can probably re-target the tools used there to instead track the validation date, since the snapshot update date can actually just stay the same until it really needs to be changed.

In the past a few of our team expressed interest in having this data available, so there you go :)

## Changes

- Add new pytest plugin which creates a new entry in a `*.validation.json` companion file that is updated when a test is ran against AWS and executes successfully.
- Add filtering option to select tests that have a validation date older than a user-provided date.
- Added a  small (commented-out) util that I'll use in a follow-up PR to set the validation date for existing snapshot tests to the snapshot's updated date.

## Testing

```
pytest ... --validation-date-limit-days=10
pytest ... --validation-date-limit-timestamp="2023-12-01T00:00:00"
```

## Future work
- Linting that makes sure if something is marked `@markers.aws.validated` that is also has a validation file/date (and the other way around as well.
- A follow-up PR will populate an initial set of validation files by assuming the snapshot date as the last validated date.

### DIFF
--- a/localstack/testing/pytest/validation_tracking.py
+++ b/localstack/testing/pytest/validation_tracking.py
@@ -1,0 +1,74 @@
+"""
+When a test (in tests/aws) is executed against AWS, we want to track the date of the last successful run.
+
+Keeping a record of how long ago a test was validated last,
+we can periodically re-validate ALL AWS-targeting tests (and therefore not only just snapshot-using tests).
+"""
+
+import datetime
+import json
+import os
+from typing import Optional
+
+import pluggy
+import pytest
+
+from localstack.testing.aws.util import is_aws_cloud
+
+
+def find_snapshot_for_item(item: pytest.Item) -> Optional[dict]:
+    base_path = os.path.join(item.fspath.dirname, item.fspath.purebasename)
+    snapshot_path = f"{base_path}.snapshot.json"
+
+    if not os.path.exists(snapshot_path):
+        return None
+
+    with open(snapshot_path, "r") as fd:
+        file_content = json.load(fd)
+        return file_content.get(item.nodeid)
+
+
+def record_passed_validation(item: pytest.Item, timestamp: Optional[datetime.datetime] = None):
+    base_path = os.path.join(item.fspath.dirname, item.fspath.purebasename)
+    with open(f"{base_path}.validation.json", "w+") as fd:
+        # read existing state from file
+        try:
+            content = json.load(fd)
+        except json.JSONDecodeError:  # expected on first try (empty file)
+            content = {}
+
+        # update for this pytest node
+        if not timestamp:
+            timestamp = datetime.datetime.now(tz=datetime.timezone.utc)
+        content[item.nodeid] = {"last_validated_date": timestamp.isoformat(timespec="seconds")}
+
+        # save updates
+        fd.seek(0)
+        json.dump(content, fd, indent=2)
+
+
+# TODO: we should skip if we're updating snapshots
+# make sure this is *AFTER* snapshot comparison => tryfirst=True
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
+def pytest_runtest_call(item: pytest.Item):
+    outcome: pluggy.Result = yield
+
+    # we only want to track passed runs against AWS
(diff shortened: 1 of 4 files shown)


localstack/testing/pytest/validation_tracking.py
```diff
@@ -0,0 +1,138 @@
...
+# @pytest.hookimpl(trylast=True)
+# def pytest_collection_modifyitems(session, config, items: list[pytest.Item]):
+#     for item in items:
+#         snapshot_entry = find_snapshot_for_item(item)
+#         if not snapshot_entry:
+#             continue
+#
+#         snapshot_update_timestamp = datetime.datetime.strptime(snapshot_entry["recorded-date"], "%d-%m-%Y, %H:%M:%S").astimezone(tz=datetime.timezone.utc)
+#
+#         record_passed_validation(item, snapshot_update_timestamp)
```
bentsku, line 86: Just a minor thing about the collection and loading the file for each item, but it feels like a way too early optimization: would there be a way to group the items by `base_path` and load the file only once? I feel like it would over-complicate things quite a bit for maybe not so much gain.
But looking at the S3 snapshot file, it has 218 entries, and loading this 340kb file 218 times feels like a lot. 

Maybe using `itertools.groupby` could help a bit somehow, I've given it a quick try and I could get the time to collect all snapshots from 850ms to 30ms. 

```python

def get_snapshot_file(base_path: str) -> Optional[dict]:
    snapshot_path = f"{base_path}.snapshot.json"

    if not os.path.exists(snapshot_path):
        return None

    with open(snapshot_path, "r") as fd:
        file_content = json.load(fd)
        return file_content


def find_snapshot_for_item(item: pytest.Item, snapshot_data: dict = None) -> Optional[dict]:
    if not snapshot_data:
        base_path = os.path.join(item.fspath.dirname, item.fspath.purebasename)
        snapshot_path = f"{base_path}.snapshot.json"

        if not os.path.exists(snapshot_path):
            return None

        with open(snapshot_path, "r") as fd:
            snapshot_data = json.load(fd)

    return snapshot_data.get(item.nodeid)


@pytest.hookimpl(trylast=True)
def pytest_collection_modifyitems(session, config, items: list[pytest.Item]):
    def sort_key(x: pytest.Item):
        return os.path.join(x.fspath.dirname, x.fspath.purebasename)

    for key, g_items in groupby(sorted(items, key=sort_key), key=sort_key):
        snapshot_data = get_snapshot_file(base_path=key)
        if not snapshot_data:
            continue
        for item in g_items:
            snapshot_entry = find_snapshot_for_item(item, snapshot_data)
            if not snapshot_entry:
                continue
```

It's just a thing I've thought about while reviewing, it's not so much linked to this PR but more to the collection, but the gain is so small that it feels quite overkill for no reason. I just gave it a try, so I thought it was worth sharing anyway 😅

bentsku (approved):
LGTM! This is pretty cool! 
I like the idea of recording the last validation, it is definitely true that it was a missing piece that is going to be really useful. API Gateway already changed some things during re:invent and we will need to regenerate again.

I've gave it a try and it worked perfectly! Thanks a lot 🚀

localstack/testing/pytest/validation_tracking.py
dominikschubert: You absolutely could, but you also correctly identified that it's basically pointless because of the scale, in which case I prefer keeping this as straight-forward as possible for now. These things are basically only executed on-demand on your local machine for now. The commented out area especially is literally something that we only do once for community and once for pro now :sweat_smile: 

Still a good exercise to think about, so feel free to go ahead and add a follow-up PR in the future. I'd be glad to share some ownership over the validation process :wink:

PR #9928, opened 2023-12-21
Add initial validation files for recorded snapshots

dominikschubert:
## Motivation

Follow-up to #9927 to bootstrap the first validation files with known dates

## Changes

- For every recorded snapshot this script creates a validation file with the same timestamp
- Multiple test cases in a single validation file are now sorted by their nodeid
- Saving multiple validations in a single file has been fixed (this was always overwriting previous ones)


localstack/testing/pytest/validation_tracking.py
```diff
@@ -42,7 +42,7 @@ def find_validation_data_for_item(item: pytest.Item) -> Optional[dict]:
 
 def record_passed_validation(item: pytest.Item, timestamp: Optional[datetime.datetime] = None):
     base_path = os.path.join(item.fspath.dirname, item.fspath.purebasename)
-    with open(f"{base_path}.validation.json", "w+") as fd:
+    with open(f"{base_path}.validation.json", "r+") as fd:
```
bentsku, line 45: oof, good catch... I only tried it with one test and totally missed that one. Sorry 😞

bentsku (approved):
LGTM! Good catch for the overwriting, sorry I totally missed it. The sorting is nice as well, looks clean 👍 

Funny to see which services have snapshots, around 22, I'm sure we could do better 😄 

Thanks a lot for tackling this!

localstack/testing/pytest/validation_tracking.py
```diff
@@ -42,7 +43,9 @@ def find_validation_data_for_item(item: pytest.Item) -> Optional[dict]:
 
 def record_passed_validation(item: pytest.Item, timestamp: Optional[datetime.datetime] = None):
     base_path = os.path.join(item.fspath.dirname, item.fspath.purebasename)
-    with open(f"{base_path}.validation.json", "w+") as fd:
+    file_path = f"{base_path}.validation.json"
+    Path(file_path).touch()
+    with open(file_path, "r+") as fd:
```
bentsku, line 48: And I missed that again when I encountered the same issue with S3 😅 
I did it in a try/except there because that's what `touch` does under the hood, but this is shorter! 
```suggestion
    file_path = Path(f"{base_path}.validation.json")
    file_path.touch()
    with file_path.open(mode="r+") as fd:
```

bentsku (commented):
Just added a small comment with using `Path`, sorry I missed _again_

localstack/testing/pytest/validation_tracking.py
dominikschubert: yeah I noticed it when creating the -ext files :grin:

PR #9932, opened 2023-12-22
Add events method to docker clients

simonrw:
## Motivation

For an upcoming change to the way containers are handled, we will want to be able to capture the docker events in order to properly handle notifications of container termination or shutdown.

## Changes

This PR adds an `events` method, which returns a generator of the events from the docker daemon.


localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -408,6 +409,39 @@ def stream_container_logs(self, container_name_or_id: str) -> CancellableStream:
...
+        process: subprocess.Popen = run(
+            cmd, asynchronous=True, outfile=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+
+        for msg in CancellableProcessStream(process):
+            try:
+                yield json.loads(msg)
+            except json.JSONDecodeError as e:
+                LOG.warning("Error decoding docker event %s: %s", msg, e)
+
```
dfangl, line 444: Not sure about this. The stream is not cancellable with this construct. I think for these method we should always use a CancellableStream, so we can cancel it if necessary.

dfangl (changes requested):
I think this can help us significantly, but we need to make the streams cancellable, to avoid resource leaks.

dfangl (approved):
Thanks for making it cancellable!

Issue #9937, opened 2023-12-24
bug: StepFunction with JMESPath yields "Unexpected character"


phildunphytoo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running a step function that includes a ResultSelector with this json:
```
{
  "FailedTasks.$": "$[?(@.Status != '201')]"
}
```

Localstack fails with error:
```
Exception=JsonPathLexerError, Details=['Error on line 1, col 2: Unexpected character: ? '] ...
```

### Expected Behavior

Expected behaviour as in AWS
(Specifically here: the following should check if any of the results contains a status that's not '201')

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Run localstack with docker-compose
Create the step function with a map and ResultSelector as described
Run the step function and see it fail :(

### Environment

```markdown
- OS:Rhel 8.6
- LocalStack: 3.0
```

phildunphytoo:
I found this: https://github.com/h2non/jsonpath-ng/issues/8
It could be related

phildunphytoo:
The same in 3.0.3

phildunphytoo:
Thanks! Waiting for the release :)

MEPalma:
The StepFunctions interpreter now correctly delegates to the jsonpath library, which resolves several issues related to parsing and accessing JSON path definitions. However, it's important to note that the jsonpath_ng library doesn't always adhere completely to the spec (as discussed in [h2non/jsonpath-ng#8](https://github.com/h2non/jsonpath-ng/issues/8)).

Considering this, I am staling this issue for now. Going forward, I will keep an eye on the progress of [h2non/jsonpath-ng#8](https://github.com/h2non/jsonpath-ng/issues/8) to ensure alignment with the specifications. Additionally, the tests added in this PR will remain enabled for filtering JSON path definitions.

Feel free to reopen this issue or raise any further concerns if needed. Thank you again for your contribution!

phildunphytoo:
Hi, I'm testing this in localstack 3.3 but it seems it doesn't work right.
I have a condition: "$[?(@.Status != '200')]" and it catches outputs that have Status = 200.. Any ideas?

PR #9942, opened 2023-12-26
[SFN] Support for JsonPath Regexes

MEPalma:
## Motivation
Currently the SFN v2 provider incorrectly parses json path strings especially when filtering options. Furthermore, the non extended version of `jsonpath_ng` parser is used. The PR simplified the parsing of json path definitions and delegates them all as a complete string to `jsonpath_ng`. The PR also resolves some issue with returning single objects or list of objects in some json path accesses.
However, currently, the `jsonpath_ng` is not always able to adhere to spec https://github.com/h2non/jsonpath-ng/issues/8. This issue should be followed in the future, and the tests added in this PR enabled for filtering json path definitions. Addresses and stales: #9937 

## Changes
- updated the intrinsic lexer and parser to bind any json path definition to a single token
- updated the json path access logic to use `jsonpath_ng` `ext` parser
- updated the json path access logic to more consistently return single objects or list of objects depending on the json path definition
- adds relevant tests


phildunphytoo:
I would love to see this merged :)

dominikschubert (approved):
LGTM :rocket: 

I'd be curious to see at some point what subset of jsonpath stepfunctions on AWS supports, but I guess for now just depending on jsonpath_ng and its support should be fine for most use cases. :+1:

PR #9945, opened 2023-12-27
cloudwatch: fix functionality of metrics for multi-accounts and region

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `cloudwatch` tests should still create the consequent resources in these accounts and region.

## Changes
This PR adds `account_id` parameter to the `cloudwatch` client when publishing the lambda metric on function invoke. The remaining changes involves propagating its value to the corresponding functions.

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`). This PR fixes: 

```
tests.aws.services.cloudwatch.test_cloudwatch_metrics.TestCloudWatchLambdaMetrics.test_lambda_invoke_successful
tests.aws.services.cloudwatch.test_cloudwatch_metrics.TestCloudWatchLambdaMetrics.test_lambda_invoke_error
```

### DIFF
--- a/localstack/services/lambda_/invocation/metrics.py
+++ b/localstack/services/lambda_/invocation/metrics.py
@@ -5,30 +5,33 @@ from localstack.utils.cloudwatch.cloudwatch_util import publish_lambda_metric
 LOG = logging.getLogger(__name__)
 
 
-def record_cw_metric_invocation(function_name: str, region_name: str, function_arn: str):
+def record_cw_metric_invocation(function_name: str, account_id: str, region_name: str):
     try:
         publish_lambda_metric(
             "Invocations",
             1,
-            {"func_name": function_name, "func_arn": function_arn},
+            {"func_name": function_name},
             region_name=region_name,
+            account_id=account_id,
         )
     except Exception as e:
         LOG.debug("Failed to send CloudWatch metric for Lambda invocation: %s", e)
 
 
-def record_cw_metric_error(function_name: str, region_name: str, function_arn: str):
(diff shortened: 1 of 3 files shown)


joe4dev (changes requested):
Thank you for fixing this @sannya-singal ✨ 

I can confirm that using a different account failed before the fix and now works properly 👏👏👏: `TEST_AWS_ACCOUNT_ID=111111111111 LAMBDA_DEV_PORT_EXPOSE=1 TEST_PATH="tests/aws/services/cloudwatch/test_cloudwatch_metrics.py::TestCloudWatchLambdaMetrics::test_lambda_invoke_successful" make test` 

The region should already work properly.


Parsing the `account_id` from the function ARN requires regex matching for every CloudWatch metric submission. We can speed things up by passing the account_id explicitly instead of parsing it. Hence, I suggest passing something like `account_id=self.function_version.id.account,` in `version_manager.py`.

sannya-singal:
Ext Integration Tests- https://github.com/localstack/localstack-ext/actions/runs/7721773635

joe4dev (approved):
LGTM 👍 

Thank you @sannya-singal  for fixing multi-accounts for Lambda CloudWatch metrics 🥳

sannya-singal:
Thanks @joe4dev 🚀 🎉

Issue #9947, opened 2023-12-27
bug: Stepfunction ASLParserException for Version


kelhelawy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When my StepFunction has a `Version` field, I get the below error:
```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: ASLParserException line 13:2, at "Version", mismatched input '"Version"' expecting {'"Comment"', '"States"', '"StartAt"', '"TimeoutSeconds"'}
```

Sample function:
```
{
  "Comment": "Jira C2C Pre flight check execution",
  "StartAt": "MarkImportSucceed",
  "States": {
    "MarkImportSucceed": {
      "Type": "Task",
      "Resource": "arn:aws:states:::dynamodb:updateItem",
      "ResultPath": "$.response",
      "End": true,
      "Retry": []
    }
  },
  "Version": "1.0"
}
```

### Expected Behavior

It should work as Version is an optional field in the State machine structure
https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-state-machine-structure.html

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

### Environment

```markdown
- OS: Mac
- LocalStack: 3.0.2
```

PR #9948, opened 2023-12-27
[SFN] Support for Version Declarations

MEPalma:
## Motivation
The current stepfunctions interpreter lacks support for 'Version' declarations. This pull request introduces support for such declarations.
Closes: #9947 

## Changes
Added grammar support for `Version` declaration in the ASL Lexer and Parser. Added new parsing component `Version` to encapsulate the values of such declarations. Added relevant test, but not negative test as we don't currently support linters on creation. Currently we fail if the version given is different to 1.0, as done by AWS.


dominikschubert (approved):
LGTM, Kinda hope we never need to properly support this semantically :grin:

Issue #1645, opened 2019-10-12
Docker image not acting on SIGINT/SIGTERM


jishi:
When running the docker image isolated with something like:

`docker run --rm -p 4584:4584 -e SERVICES=secretsmanager localstack/localstack`

it doesn't stop on Ctrl + C, and if you send a `docker stop [container]` it will timeout and docker will finally kill it since it doesn't act on neither SIGINT nor SIGTERM.

maybe it's possible to `trap` SIGINT in the docker-entrypoint.sh and do a graceful shutdown? My bash skills are a bit weak to actually know :)

whummer:
Thanks for reporting @jishi - have you tried running with the `-it` flags enabled?
```
docker run -it --rm -p 4584:4584 -e SERVICES=secretsmanager localstack/localstack
```
In general, the recommended way to start LocalStack is using the pip package `localstack`. It sets all the required Docker flags for you. Example output on Mac OS:
```
$ localstack start
docker run -it -e DEFAULT_REGION="us-east-1" -e LOCALSTACK_HOSTNAME="localhost" -p 0.0.0.0:53:53 -p 0.0.0.0:53:53/udp -p 4510-4530:4510-4530 --rm --privileged --name localstack_main -p 8080:8080 -p 8081:8081  -p 443:443 -p 4567-4607:4567-4607  -v "/private/var/folders/n0/xq8l4cb94278hvzj00wzn4c40000gp/T/localstack:/tmp/localstack" -v "/var/run/docker.sock:/var/run/docker.sock" -e DOCKER_HOST="unix:///var/run/docker.sock" -e HOST_TMP_FOLDER="/private/var/folders/n0/xq8l4cb94278hvzj00wzn4c40000gp/T/localstack" "localstack/localstack"
...
```

jishi:
Thank you for the prompt reply! Ah, my bad, it does actually catch SIGINT correctly when using `-it` flag. 

However, it still doesn't work with external signals, like

`docker kill -s INT [container]` or `docker stop [container]`

Docker stop sends SIGTERM by default (IIRC) and will send a SIGKILL if the process doesn't respond within 10 seconds, which seems to always be the case. 

I would find this useful for daemonizing the container in a CI environment when running my tests, although at least supporting Ctrl +C was my main problem to begin with. 

If clean shutdown from a `docker stop` (and I guess this affects `docker-compose stop` as well) is irrelevant, you may close this issue.

whummer:
[...]

PR #9949, opened 2023-12-27
Added SIGINT trap, improved docker workflow

kirill-dev-pro:
Now docker container will catch SIGINT from `docker stop` and similar commands and will exit right away instead of waiting docker timeout of 10s and getting SIGKILL

Haven't tested it

linking https://github.com/localstack/localstack/issues/1645

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Function to perform cleanup before exiting
+cleanup() {
+    echo "Caught SIGINT, exiting..."
+    exit 1
+}
+
+# Trap the SIGINT signal and call the cleanup function
+trap cleanup SIGINT
+
 set -eo pipefail
 shopt -s nullglob
 


thrau:
since we're yielding control to `localstack-supervisor` here, i would think signal handling happens there? https://github.com/localstack/localstack/blob/0f1b97d84260d46140d34dbf6f3c02fc69ffd807/bin/docker-entrypoint.sh#L39-L40

in fact we already do explicit signal handling there: https://github.com/localstack/localstack/blob/0f1b97d84260d46140d34dbf6f3c02fc69ffd807/bin/localstack-supervisor#L144-L147

SIGINT has behaved weird in the past (as also stated in the nebulous TODO). perhaps unrelated though.

not sure what to make of the change proposed in the PR exactly given this additional context. adding @dfangl who's been working on signal handling in localstack.

thrau (commented):
thanks for taking a stab at this @kirill-dev-pro!

since we're yielding control to `localstack-supervisor` here, i would think signal handling happens there? https://github.com/localstack/localstack/blob/0f1b97d84260d46140d34dbf6f3c02fc69ffd807/bin/docker-entrypoint.sh#L39-L40

in fact we already do explicit signal handling there: https://github.com/localstack/localstack/blob/0f1b97d84260d46140d34dbf6f3c02fc69ffd807/bin/localstack-supervisor#L144-L147

SIGINT has behaved weird in the past (as also stated in the nebulous TODO). perhaps unrelated though.

not sure what to make of the change proposed in the PR exactly given this additional context. adding @dfangl who's been working on signal handling in localstack.

alexrashed:
Hi there! 
Sorry for chiming in, but out of curiosity I just tested a `SIGTERM` and a `SIGINT` with the `latest` image.
- Started the container:
  ```
  $ docker run --rm --name localstack-main localstack/localstack
  
  LocalStack version: 3.0.3.dev
  LocalStack build date: 2023-12-29
  LocalStack build git hash: 563edf7a
  
  2024-01-02T09:20:27.963  INFO --- [-functhread6] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
  2024-01-02T09:20:27.963  INFO --- [-functhread6] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
  2024-01-02T09:20:28.150  INFO --- [  MainThread] localstack.utils.bootstrap : Execution of "start_runtime_components" took 602.10ms
  Ready.
  ```
- Testing `SIGTERM`: `docker stop localstack-main`
- Testing `SIGINT`: `docker kill --signal SIGINT localstack-main`

For both signals, localstack terminates correctly in time with the log indicating that the supervisor takes over and stops all services. This takes ~3 seconds for me.
```
2024-01-02T09:20:42.358  INFO --- [  MainThread] l.runtime.shutdown         : [shutdown] Stopping all services
```

dfangl:
I also agree that the supervisor is responsible for signal handling, the bash script is not executed at that point anymore.

@kirill-dev-pro could you maybe create an issue with the exact error you are seeing, so we can discuss possible solutions for it? 

The signal trap in here will only be relevant during the sourcing of the venv and the execution of the boot init hooks. If those are the problem, we need to find a way to also forward the signals correctly, if they are not sent to the whole process group.

kirill-dev-pro:
There is no particular error, just docker container doesn't receive SIGINT and docker send SIGKILL after 10 second timeout.

alexrashed:
Could you maybe post a set of commands in order to reliably reproduce this? As I tried to show in my comment above, I'm having a hard time reproducing your issue...

alexrashed:
@kirill-dev-pro: I'll close this PR for now. Thanks a lot for your proactive approach, and trying to fix your issue. I would be really happy to pick this up again if you can give us a proper example how to reproduce this. As mentioned in my comment above, I did a simple test trying to reproduce your issues, but I was not successful in doing so. 
However, it could definitely be that there's some edge case where we need a more elaborate test to reproduce your issue. Please feel free to report back (by creating a new issue or a PR with a test) and we're happy to take this one up again!

PR #9960, opened 2023-12-29
refactor the status cli command

HarshCasper:
This PR refactors the `localstack status` command to display details from `/info` endpoint.

```bash
$ python3 -m localstack.cli.main status
┌──────────────────────┬──────────────────────────────────────┐
│ version              │ 3.0.3.dev20231227173904:c249976      │
│ edition              │ pro                                  │
│ is_license_activated │ True                                 │
│ session_id           │ 554ab315-b019-4486-92bb-190093111eaf │
│ machine_id           │ 0c49752c                             │
│ system               │ linux                                │
│ is_docker            │ True                                 │
│ server_time_utc      │ 2023-12-29T12:26:24                  │
│ uptime               │ 6                                    │
└──────────────────────┴──────────────────────────────────────┘
```

### DIFF
--- a/localstack/cli/localstack.py
+++ b/localstack/cli/localstack.py
@@ -7,6 +7,7 @@ from typing import Dict, List, Optional, Tuple, TypedDict
 
 import click
 import requests
+from rich.table import Table
 
 from localstack import __version__, config
 from localstack.cli.exceptions import CLIError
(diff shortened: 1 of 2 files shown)


thrau (changes requested):
thanks for taking a stab at improving the status command @HarshCasper! i would generally not just replace what we previously had with the output of the info endpoint (which is really for internal systems to use).
what i could see as an improvement though is additionally querying the info endpoint if the localstack instance is running, and extracting the "edition" and "license activation" field, and adding that to the output we already had.

PR #9962, opened 2023-12-29
Refactor: EventBridge: move clean_up fixture to localstack/testing/fixtures be able to use in localstack-ext

maxhoheiser:
## Motivation
Currently the fixture that is used by the EventBrdige tests to clean up the created items, is not accessible from localstack-ext. 

## Changes
To avoid code duplication, the fixture is moved to `localstack/testing/fixtures`


dfangl:
As discussed in slack, the cleanup utility is highly deprecated. I would prefer using the currently used techniques for test cleanup, instead of porting this utility.

PR #9971, opened 2024-01-02
Add performance logging for tests

joe4dev:
## Motivation

After merging https://github.com/localstack/localstack/pull/9917 to mitigate async Lambda reliability issues upon high load, the pipeline got blocked by 🔴 build timeouts because test time massively increased (from ~30min to >45-60min for each runner). We need better performance logging to track performance regressions that make many tests a little bit slower but are hard to pinpoint in a single test (especially without dedicated performance tests).

[Pytest duration profiling](https://stackoverflow.com/a/55095253/6875981) is not applicable for runners that time out. Therefore, we add print the timediff directly in the live logs. Example to print all durations: `pytest --durations=0 --durations-min=0.0 -vv`

## Changes

* Add very basic timediff logging in seconds for each test

## Limitations

Ad-hoc approach: the timediff for the first test is inaccurate

### DIFF
--- a/tests/aws/conftest.py
+++ b/tests/aws/conftest.py
@@ -1,4 +1,6 @@
+import json
 import os
+import time
 from typing import Optional
 
 import pytest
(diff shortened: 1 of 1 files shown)


joe4dev:
Closing because we identified and fixed the issue in https://github.com/localstack/localstack/pull/9972

PR #9944, opened 2023-12-27
Revert recent changes to async lambda invoke (#9917)

dominikschubert:
## Motivation

Starting from that merge on, we experienced timeouts in our builds. To unblock other efforts, we can revert it and investigate re-enabling it separately

## Changes

- This PR reverts commit ce5fe90919f92d60a141793afa8de7d14f418d58 from #9917

PR #9972, opened 2024-01-02
Test lambda performance and mitigate async invoke timeout error v2

joe4dev:
## Motivation

Resume https://github.com/localstack/localstack/pull/9917, which we had to revert in https://github.com/localstack/localstack/pull/9944 due to timeouts in the CI.

We found the issue with the short polling, which was responsible for the pipeline timeouts, and reverted only this part, while leaving other changes intact.

## Changes

* Re-apply the changes from https://github.com/localstack/localstack/pull/9917

While reintroducing the changes, we made the following changes:

* smaller defaults for the client configuration
* improved error handling which prints errors, but continues listener operation
* without the short polling, which was responsible for our pipeline performance issues.

### DIFF
--- a/localstack/services/lambda_/invocation/event_manager.py
+++ b/localstack/services/lambda_/invocation/event_manager.py
@@ -134,24 +134,22 @@ class Poller:
             )
             function_timeout = self.version_manager.function_version.config.timeout
             while not self._shutdown_event.is_set():
-                # TODO: Fix proper shutdown causing EndpointConnectionError
-                # https://app.circleci.com/pipelines/github/localstack/localstack/17428/workflows/391fc320-0cec-4dd1-9e3b-d7511de61d12/jobs/132663/parallel-runs/2
-                # Test case (happens not every time!):
-                # tests.aws.services.cloudformation.resources.test_legacy.TestCloudFormation.test_updating_stack_with_iam_role
-                messages = sqs_client.receive_message(
+                response = sqs_client.receive_message(
                     QueueUrl=self.event_queue_url,
                     WaitTimeSeconds=2,
-                    # TODO: MAYBE: increase number of messages if single thread schedules invocations
-                    MaxNumberOfMessages=1,
+                    # Related: SQS event source mapping batches up to 10 messages:
+                    # https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html
+                    MaxNumberOfMessages=10,
                     VisibilityTimeout=function_timeout + 60,
                 )
-                if not messages.get("Messages"):
+                if not response.get("Messages"):
                     continue
-                message = messages["Messages"][0]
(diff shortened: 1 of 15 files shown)


localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -104,9 +105,10 @@ def has_enough_time_for_retry(
     )
 
 
+# TODO: optimize this client configuration. Do we need to consider client caching here?
 CLIENT_CONFIG = Config(
-    connect_timeout=1,
-    read_timeout=5,
+    connect_timeout=5,
```
joe4dev, line 110: this is the key change of the mitigation

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -128,41 +130,48 @@ def __init__(self, version_manager: LambdaVersionManager, event_queue_url: str):
...
-            if isinstance(e, ConnectionRefusedError) and self._shutdown_event.is_set():
-                return
-            # TODO: investigate what causes ReadTimeoutError (fixed with increasing read timeout?)
-            LOG.error(
-                "Error while polling lambda events for function %s: %s",
-                self.version_manager.function_version.qualified_arn,
-                e,
-                exc_info=LOG.isEnabledFor(logging.DEBUG),
-            )
+            except Exception as e:
```
joe4dev, line 159: the 2nd key change is to catch errors in the poller and retry instead of terminating the poller upon the first error

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -128,41 +130,48 @@ def __init__(self, version_manager: LambdaVersionManager, event_queue_url: str):
...
+            self.version_manager.function_version, client_config=CLIENT_CONFIG
+        )
+        function_timeout = self.version_manager.function_version.config.timeout
+        while not self._shutdown_event.is_set():
+            try:
+                response = sqs_client.receive_message(
                     QueueUrl=self.event_queue_url,
+                    # TODO: consider replacing with short polling instead of long polling to prevent keeping connections open
+                    # however, we had some serious performance issues when tried out, so those have to be investigated first
                     WaitTimeSeconds=2,
```
dfangl, line 143: Leaving the long polling here instead of switching to short polling is the main fix for the whole of the changes.

dfangl (approved):
Looks good, but that does not say too much, since I also wrote it :D

localstack/services/lambda_/invocation/event_manager.py
joe4dev: yeah, that's the main fix of the fix :)
For unknown reasons, short-polling caused accumulated slowdowns of 20-35min across our entire test suite 🤷

Issue #9957, opened 2023-12-29
bug: `AllowedHeaders` in S3 CORS configurations should not be case-sensitive


bigen1925:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`AllowedHeaders` in S3 CORS configurations seems to be case-sensitive.

For example, 
```
{
  "CORSRules": [
    {
      "AllowedHeaders": ["Content-Type"],
      ...
    }
  ]
}
```
this settings reject pre-flight requests with
```
Access-Control-Request-Headers: content-type
```

### Expected Behavior

Access-Control-Request-Headers of preflight requests is accepted in case-insensitive as real S3 does so.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose up

```yaml
  localstack:
    image: localstack/localstack:3.0.2
    volumes:
      - ./localstack/ready.d:/etc/localstack/init/ready.d
      - /var/run/docker.sock:/var/run/docker.sock
    ports:
      - '4566:4566'
    environment:
      AWS_ACCESS_KEY_ID: test
      AWS_SECRET_ACCESS_KEY: test
      REGION: ap-northeast-1
      S3_BUCKET: bucket
      S3_SKIP_SIGNATURE_VALIDATION: 0
      S3_CORS_ALLOWED_ORIGIN: http://localhost:3000 // frontend application port
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
# make bucket
awslocal s3 mb s3://${S3_BUCKET}

# cors settings
awslocal s3api put-bucket-cors --bucket ${S3_BUCKET} --cors-configuration '
{
  "CORSRules": [
    {
      "AllowedHeaders": ["Content-Type"],
      "AllowedMethods": ["GET", "PUT"],
      "AllowedOrigins": ["'http://localhost:3000'"]
    }
  ]
}
'
```

#### And then

1. generate signed url of putObject
2. putObject from browser application
```
fetch(signedUrl, { method: 'PUT', body: file })
```

fetch API sends pre-flight requests with 
```
Access-Control-Request-Headers: content-type
```
in default (Note that `content-type` is with all lower cases thought CORS settings is with Upper cases), and it is rejected.

### Environment

```markdown
- OS: Mac OS 14.2.1
- LocalStack: 3.0.2
```

### Anything else?

[...]

PR #9973, opened 2024-01-02
fix S3 CORS headers casing matching

bentsku:
## Motivation
As reported in #9957, the header matching was case sensitive for the `AllowedHeaders` field for `OPTIONS` requests with the `Access-Control-Request-Headers` header.

## Changes
Wrote a AWS validated test to verify the behavior and fixed the matching to now be case insensitive.


steffyP (approved):
LGTM! Thanks for providing a fix 🚀

Issue #8724, opened 2023-07-19
bug: Incorrect behavior of receive message in sqs fifo queue with same group ID


shawn-ann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Prerequisites: Using Spring JMS and AWS sdk 1.x to receive messages

1. Create a fifo queue
```bash
awslocal sqs create-queue --queue-name example.fifo --attributes '{"VisibilityTimeout":"60","FifoQueue":"true","ContentBasedDeduplication":"true","DeduplicationScope":"messageGroup","FifoThroughputLimit":"perMessageGroupId"}'

```
2. Send a few messages to a fifo queue with a same group ID
3. Using 5 threads to receive and process messages through Spring JMS, it takes 50 seconds to process a message (Thread.sleep).
```java 
public static void main(String[] args) throws Exception {
        AmazonSQSAsync sqsClient = AmazonSQSAsyncClientBuilder.standard()
                .withCredentials(new DefaultAWSCredentialsProviderChain())
                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(SQSConfig.serviceEndpoint, "us-east-1"))
                .build();
        SQSConnectionFactory connectionFactory = new SQSConnectionFactory(new ProviderConfiguration(), sqsClient);

        MessageListener messageListener = new ReceiverCallback();
        DefaultMessageListenerContainer listenerContainer = new DefaultMessageListenerContainer();
        listenerContainer.setConnectionFactory(connectionFactory);
        listenerContainer.setDestinationName(SQSConfig.queueName);

        listenerContainer.setSessionTransacted(false);
        listenerContainer.setSessionAcknowledgeMode(Session.CLIENT_ACKNOWLEDGE);

        listenerContainer.setMessageListener(messageListener);
        listenerContainer.setConcurrentConsumers(5); 
        listenerContainer.setAutoStartup(true);
        listenerContainer.initialize();
        listenerContainer.start();
    }

    public static class ReceiverCallback implements MessageListener {
        // Used to listen for message silence
        private final Map<String, List<String>> messageGroups = new ConcurrentHashMap<>();

        @Override
        public void onMessage(Message message) {
            try {
                String groupId = message.getStringProperty("JMSXGroupID");
                System.out.println(String.format("%s %s receive,Group ID: %s, start..",
                        Thread.currentThread().getName(),
                        System.currentTimeMillis(), groupId));
                Thread.sleep(50000);
[...]

PR #7037, opened 2022-10-17
Add integration test for SQS FIFO event source order/parallelism

dominikschubert:
Adds an integration test covering  #7036

Issue #7036, opened 2022-10-17
bug: Lambda should correctly process FIFO queue event sources


dominikschubert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Multiple lambda environments are spawned in parallel

### Expected Behavior

A single lambda environment should handle the messages sequentially.

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

Scenario:

1. Create a lambda with a set timeout (e.g. 30s) + fifo queue + disabled event source mapping
2. Send 5 messages to the queue with the same message group id ( => guarantees that they should be processed one by one in-order)
3. enable the event source mapping so that the Lambda service can start processing the messages
4. Wait for some time ( e.g. 90s) and check how many log streams have been created

### Environment

```markdown
-
```

thrau:
confirmed to work with latest using the test you provided!

PR #9975, opened 2024-01-03
fix sqs message group visibility when sending messages

thrau:
## Motivation

Even after the initial implementation of FIFO message groups in #8238, users have reported odd behavior when using SQS FIFO queues related to message group visibility specifically that consumers are triggered although the message group is expected to be invisible #8724. This also affected lambdas triggered with fifo queues #7036.

This PR fixes the mechanism we were using to reset message group visibility, which was too general. It assumed that every message put into the message group reset its visibility, when really it should just be messages that were previously invisible, not new ones.

## Changes

* message group visibility is no longer reset when new messages are sent to the queue

Fixes
* #8724
* #7036 (tested with #7037)

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -938,7 +938,9 @@ class FifoQueue(SqsQueue):
             # put the message into the group
             message_group.push(message)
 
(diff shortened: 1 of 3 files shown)


baermat (approved):
LGTM :+1:

Issue #8934, opened 2023-08-18
bug: cloudwatch alarm fails on getting metric data


MikhailStrokov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When cloudwatch alarm should trigger alarm action sending message to sns topic the following error messages occurs:

`2023-08-18T08:01:21.489 ERROR --- [   asgi_gw_1] l.aws.handlers.logging     : exception during call chain: could not convert string to float: 'None'
2023-08-18T08:01:21.918 ERROR --- [   asgi_gw_5] l.aws.handlers.logging     : exception during call chain: could not convert string to float: 'None'
2023-08-18T08:01:23.526 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: could not convert string to float: 'None'
2023-08-18T08:01:26.861 ERROR --- [   asgi_gw_3] l.aws.handlers.logging     : exception during call chain: could not convert string to float: 'None'
2023-08-18T08:01:32.779 ERROR --- [   asgi_gw_2] l.aws.handlers.logging     : exception during call chain: could not convert string to float: 'None'
2023-08-18T08:01:32.782 ERROR --- [ch-scheduler] l.s.c.alarm_scheduler      : Error executing scheduled alarm
Traceback (most recent call last):
  File "/opt/code/localstack/localstack/utils/scheduler.py", line 64, in run
    self.task(*self.args, **self.kwargs)
  File "/opt/code/localstack/localstack/services/cloudwatch/alarm_scheduler.py", line 294, in calculate_alarm_state
    metric_values = collect_metric_data(alarm_details, client)
  File "/opt/code/localstack/localstack/services/cloudwatch/alarm_scheduler.py", line 243, in collect_metric_data
    metric_data = client.get_metric_data(
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/botocore/client.py", line 534, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/opt/code/localstack/.venv/lib/python3.10/site-packages/botocore/client.py", line 976, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InternalError) when calling the GetMetricData operation (reached max retries: 4): exception while calling cloudwatch.GetMetricData: could not convert string to float: 'None'
`

### Expected Behavior

Alarm should send message to sns topic

### How are you starting LocalStack?

With a `docker run` command

[...]

PR #9978, opened 2024-01-03
fix cloudwatch alarm evaluation for SampleCount

steffyP:
## Motivation
Fix for https://github.com/localstack/localstack/issues/8934: the alarm did not trigger for `SampleCount` statistics, and was throwing errors. 
The issue was caused by casing-mismatch - the metric query was send to moto using the statistic `Samplecount` but it's expecting an exact match. 

## Changes
* fixed the casing issue
* added a snapshot test for the reported scenario (it relies on metrics automatically reported by SQS when a message is send)
* some cleanup and fixes revealed by snapshot test

### DIFF
--- a/localstack/services/cloudwatch/provider.py
+++ b/localstack/services/cloudwatch/provider.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 from xml.sax.saxutils import escape
 
 from moto.cloudwatch import cloudwatch_backends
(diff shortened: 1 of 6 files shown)


localstack/services/cloudwatch/provider.py
```diff
@@ -142,6 +143,12 @@ def put_metric_alarm(
     )
 
 
+def convert_camel_to_snake_upper_case(input: str) -> str:
+    # currently we only need to handle "SampleCount" camel case
+    input = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", input)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", input).upper()
```
silv-io, line 149: This seems generic enough to be in utils, where there are already functions with very similar purpose:
`localstack.utils.strings.snake_to_camel_case`.

steffyP: ah, yes - thanks for the hint! there is already a `camel_to_snake_case` utility 🙂

PR #9988, opened 2024-01-04
add SQS message move task feature

### LINKED ISSUE #8471: Enhancement request: Support AWS SQS's newly announced "SQS Dead-Letter Queue Redrive via AWS SDK or CLI"

### Enhancement description

On June 8, 2023, Amazon just announced `support for dead-letter queue redrive via AWS SDK or CLI`

The SDK introduces three new 'tasks' that can be performed against an SQS dead letter queue:
* `StartMessageMoveTask`
* `CancelMessageMoveTask`
* `ListMessageMoveTasks`

LocalStack's SQS implementation should be updated to support these new DLQ redrive operations.

-------------------------

Not surprisingly when I attempting to do this the day after announcement, LocalStack doesn't yet support these operations. Using the latest LocalStack docker container tag of `2.1.0` and the latest Java AWS SDK version `2.20.82`, I attempted to perform one of these new `StartMessageMoveTask` against LocalStack using the following:
```java
StartMessageMoveTaskResponse moveTaskResponse =
      sqs.startMessageMoveTask(builder -> builder.sourceArn(dlqArn).destinationArn(qArn));
```
The container produced the following error message:
```
2023-06-09T17:06:50.098 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain: Operation detection failed.Operation StartMessageMoveTask could not be found for service ServiceModel(sqs).
2023-06-09T17:06:50.099347340Z 2023-06-09T17:06:50.099  INFO --- [   asgi_gw_0] localstack.request.http    : POST / => 500
```

### 🧑‍💻 Implementation

_No response_

### Anything else?

Relevant AWS announcements and API specs:
* https://aws.amazon.com/blogs/aws/a-new-set-of-apis-for-amazon-sqs-dead-letter-queue-redrive/
* https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-sqs-dead-letter-queue-redrive-aws-sdk-cli/
* https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-dead-letter-queue-redrive.html

thrau:
## Motivation

A user requested in #8471 the Amazon SQS API for "[dead-letter queue redrive](https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-sqs-dead-letter-queue-redrive-aws-sdk-cli/)" which was launched on June 8 2023. The feature consists of three operations:

1) [StartMessageMoveTask](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_StartMessageMoveTask.html), to start a new message movement task from the dead-letter queue;
2) [CancelMessageMoveTask](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CancelMessageMoveTask.html), to cancel the message movement task;
3) [ListMessageMoveTasks](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ListMessageMoveTasks.html), to get 10 most recent message movement tasks for a specified source queue.

This PR adds this feature on top of our current SQS implementation, and a number of snapshot validated tests. The implementation introduces a new entity into the store, the `MessageMoveTask` which stores the basic attributes. The `MessageMoveTaskManager` implements an algorithm to execute a `MessageMoveTask` asynchronously, that uses our internal high-level queue API.

## Changes

* SQS now supports emulation for `StartMessageMoveTask`, `CancelMessageMoveTask`, and `ListMessageMoveTasks`
* Closes #8471

## Limitations

* Persistence for move tasks does currently not work (looks like a super fringe use case though)
* Some behavior is untested/unvalidated for the sake of getting a first version out quickly
  * What happens when messages are added to the DLQ while a move task is active? does the ApproximateMessagesToMove counter increase?
  * Multi-account access for move tasks is not clear (can you move message between queues in different accounts? if so, in which account do you have to create the move task in?)
* Error message serialization of failed move tasks are not fully correct because of the query/json protocol duality

### DIFF
--- a/localstack/services/sqs/provider.py
+++ b/localstack/services/sqs/provider.py
@@ -9,6 +9,7 @@ from concurrent.futures.thread import ThreadPoolExecutor
 from itertools import islice
 from typing import Dict, List, Optional, Tuple
 
+from botocore.utils import InvalidArnException
 from moto.sqs.models import BINARY_TYPE_FIELD_INDEX, STRING_TYPE_FIELD_INDEX
 from moto.sqs.models import Message as MotoMessage
 
@@ -22,6 +23,7 @@ from localstack.aws.api.sqs import (
     BatchRequestTooLong,
     BatchResultErrorEntry,
     BoxedInteger,
+    CancelMessageMoveTaskResult,
     ChangeMessageVisibilityBatchRequestEntryList,
     ChangeMessageVisibilityBatchResult,
     CreateQueueResult,
@@ -35,6 +37,8 @@ from localstack.aws.api.sqs import (
     InvalidBatchEntryId,
     InvalidMessageContents,
     ListDeadLetterSourceQueuesResult,
+    ListMessageMoveTasksResult,
+    ListMessageMoveTasksResultEntry,
     ListQueuesResult,
     ListQueueTagsResult,
     Message,
@@ -50,11 +54,13 @@ from localstack.aws.api.sqs import (
     QueueDoesNotExist,
     QueueNameExists,
     ReceiveMessageResult,
+    ResourceNotFoundException,
     SendMessageBatchRequestEntryList,
     SendMessageBatchResult,
     SendMessageBatchResultEntry,
     SendMessageResult,
     SqsApi,
+    StartMessageMoveTaskResult,
     String,
     TagKeyList,
     TagMap,
@@ -71,6 +77,8 @@ from localstack.services.sqs import constants as sqs_constants
 from localstack.services.sqs.exceptions import InvalidParameterValueException
 from localstack.services.sqs.models import (
     FifoQueue,
+    MessageMoveTask,
+    MessageMoveTaskStatus,
     SqsMessage,
     SqsQueue,
     SqsStore,
@@ -78,6 +86,7 @@ from localstack.services.sqs.models import (
     sqs_stores,
 )
 from localstack.services.sqs.utils import (
+    decode_move_task_handle,
     generate_message_id,
     is_fifo_queue,
     is_message_deduplication_id_required,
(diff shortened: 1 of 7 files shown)


tests/aws/services/sqs/test_sqs_move_task.py
```diff
@@ -0,0 +1,452 @@
...
+    snapshot.match("error", e.value.response)
+
+
+@markers.aws.validated
+def test_cancel_with_invalid_source_arn_in_task_handle(aws_client, snapshot):
+    source_arn = "arn:aws:sqs:us-east-1:878966065785:test-queue-doesnt-exist-123456"
+    task_handle = encode_move_task_handle("10f57157-fc38-4da9-a113-4de7e12d05dd", source_arn)
+
+    with pytest.raises(ClientError) as e:
+        aws_client.sqs.cancel_message_move_task(TaskHandle=task_handle)
```
baermat, line 119: nit: we are relying on the fact here that the queue gets checked before the task. To be 100% precise, we would have to create a task that generates a valid handle, then test with a non-existent queue and said handle. But I think we can accept this here.

tests/aws/services/sqs/test_sqs_move_task.py
```diff
@@ -0,0 +1,452 @@
...
+
+@pytest.fixture(autouse=True)
+def sqs_snapshot_transformer(snapshot):
+    snapshot.add_transformer(snapshot.transform.sqs_api())
+
+
+@pytest.fixture()
+def sqs_create_dlq_pipe(sqs_create_queue):
+    def _factory(max_receive_count: int = 1) -> tuple[QueueUrl, QueueUrl]:
+        dl_queue_url = sqs_create_queue()
```
baermat, line 82: Very Nice!

tests/aws/services/sqs/test_sqs_move_task.py
```diff
@@ -0,0 +1,452 @@
...
+    started = time.time()
+    messages = sqs_collect_messages(sqs, destination_queue, n, 60)
+    assert {message["Body"] for message in messages} == {
+        "message-0",
+        "message-1",
+        "message-2",
+        "message-3",
+    }
+
+    assert time.time() - started >= 3
```
baermat, line 282: I don't quite understand the time calculation/assertion here. That is mostly due to me reading through too much code right now :smile: but maybe a small comment would help (or I need a screen break)

tests/aws/services/sqs/test_sqs_move_task.py
```diff
@@ -0,0 +1,452 @@
...
+
+def sqs_collect_messages(
+    sqs_client, queue_url: str, expected: int, timeout: int, delete: bool = True
+):
+    collected = []
+
+    def _receive():
+        response = sqs_client.receive_message(
+            QueueUrl=queue_url,
+            WaitTimeSeconds=min(max(1, int(timeout)), 5),
```
baermat, line 25: I don't quite understand WaitTimeSeconds.
a) nit: it's an int regarding to the type hint, do we really need the cast?
b) There are multiple invocations with timeout=60 in the tests if I see this correctly. However, `min(max(1, timeout),5) will always return a number between 1 and 5, so why do we repeatedly invoke with 60?

tests/aws/services/sqs/test_sqs_move_task.py
```diff
@@ -0,0 +1,452 @@
...
+    snapshot.match("list-while", response)
+
+    response = sqs.cancel_message_move_task(TaskHandle=task_handle)
+    snapshot.match("cancel", response)
+
+    # check move task completion (in AWS, approximate number of messages may take a while to update)
+    def _wait_for_task_cancellation():
+        _response = aws_client.sqs.list_message_move_tasks(SourceArn=source_arn)
+        assert _response["Results"][0]["Status"] == "CANCELLED"
+        return _response
```
baermat, line 342: Here we check for the status, but not the approx num which is mentioned in the comments (which we checked in the in the test commented on before).

baermat (changes requested):
Very cool addition! I have a couple of nits and 1 potential error regarding the default max_number in the list_task operation. I apologize if some of the (later) remarks seem not well thought through, but the review started to get long at some point :smile:

localstack/services/sqs/provider.py
```diff
@@ -404,6 +413,105 @@ def stop(self):
...
+    def close(self):
+        with self.mutex:
+            for move_task in self.move_tasks.values():
+                move_task.cancel_event.set()
+
+            self.executor.shutdown(wait=False)
+
+    def _run(self, move_task: MessageMoveTask):
+        move_task.status = MessageMoveTaskStatus.RUNNING
+
```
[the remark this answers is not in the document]
thrau: i was thinking about persistence here, so we can re-start existing move tasks. but since we don't support persistence at all, might as well remove this one here :-) i'll add another try/catch around `start` so it doesn't stay in "RUNNING" if an error occurs before actually starting the task

localstack/services/sqs/provider.py
```diff
@@ -404,6 +413,105 @@ def stop(self):
...
+        try:
+            LOG.info(
+                "Move task started %s (%s -> %s)",
+                move_task.task_id,
```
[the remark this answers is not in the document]
thrau: fair point. i think i'm going to leave the logging here though to make sure we can debug whether the asynchronous task has really started or not! that way we can catch issues with the executor

localstack/services/sqs/provider.py
```diff
@@ -404,6 +413,105 @@ def stop(self):
...
+                source_queue.remove(receipt_handle)
+                move_task.approximate_number_of_messages_moved += 1
+
+                if rate := move_task.max_number_of_messages_per_second:
+                    move_task.cancel_event.wait(timeout=(1 / rate))
+
+        except Exception as e:
+            LOG.info(
+                "Move task failed %s: %s",
+                move_task.task_id,
```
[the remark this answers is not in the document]
thrau: good point. users can debug failed tasks by calling `ListMessageMoveTasks` and checking how many messages were moved.

localstack/services/sqs/provider.py
```diff
@@ -1213,6 +1320,193 @@ def set_queue_attributes(
...
+           (DLQs) <https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html>`__
+           only. In this context, the source queue is the dead-letter queue
+           (DLQ), while the destination queue can be the original source queue
+           (from which the messages were driven to the dead-letter-queue), or a
+           custom destination queue.
+
+        -  Currently, only standard queues are supported.
+
+        -  Only one active message movement task is supported per queue at any
+           given time.
```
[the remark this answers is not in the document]
thrau: thanks for the hint! this is actually the generated API docs i moved here while developing! i'm going to remove them all together ;-)

localstack/services/sqs/provider.py
```diff
@@ -1213,6 +1320,193 @@ def set_queue_attributes(
...
+        tasks = [
+            move_task
+            for move_task in store.move_tasks.values()
+            if move_task.source_arn == source_arn
+            and move_task.status != MessageMoveTaskStatus.CREATED
+        ]
+        tasks.sort(key=lambda t: t.started_timestamp, reverse=True)
+
+        # convert to result list
+        n = max_results or 10
```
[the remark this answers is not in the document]
thrau: well spotted!

tests/aws/services/sqs/test_sqs_move_task.py
thrau: arguably, this is even better since it tests parity of the order of AWS validations :smile:

tests/aws/services/sqs/test_sqs_move_task.py
thrau: will add a comment! max number of messages is set to 1, so assuming there is a 1 second sleep between each message, and there are four messages, that makes at least 3 seconds :-)

tests/aws/services/sqs/test_sqs_move_task.py
thrau: a) good point!
b) yea it's maybe a bit confusing what i did here. just tried to find a trade off of not polling too often, and not waiting too long on each poll

tests/aws/services/sqs/test_sqs_move_task.py
thrau: i just wanted at least one test that covers that `ApproximateNumberOfMessagesMoved` is set correctly eventually some point. as noted in a comment, in AWS, approximate number of messages may take a while to update, even after the "Status" field has been set to COMPLETED. :shrug:

baermat (approved):

Issue #9219, opened 2023-09-25
bug: Duplicate Event Source Mappings Allowed for SQS and Lambda


Blind-Striker:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

LocalStack allows the creation of multiple identical event source mappings between an SQS queue and a Lambda function without raising any errors. This behavior contrasts with AWS's approach. For instance, in LocalStack, if I repeatedly create a mapping between a specific SQS queue and a Lambda function, each attempt is successful and results in a new, yet identical, event source mapping.

### Expected Behavior

When trying to establish an event source mapping that already exists, LocalStack should raise a `ResourceConflictException`, similar to AWS's behavior. A unique mapping should be defined by its SQS ARN and the associated Lambda function. Duplicate mappings should not be permitted.

**Example:**
```bash
An error occurred (ResourceConflictException) when calling the CreateEventSourceMapping operation: An event source mapping with SQS arn (" arn:aws:sqs:eu-central-1:377140207735:profile-service-demo-queue ") and function (" message-handler-demo ") already exists. Please update or delete the existing mapping with UUID 836dd66d-f542-4844-b020-918b3ac1662d
```

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

- Create a Lambda function in LocalStack.
- Create an SQS queue in LocalStack.
- Set up an event source mapping between the SQS queue and the Lambda function.
- Attempt to set up the exact event source mapping again. This will succeed without errors, leading to multiple identical mappings.

```yml
version: "3.0"

services:
  localstack:
    image: localstack/localstack:latest
    ports:
      - "4566:4566"
      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"

    environment:
      - DEBUG="1"
      - DOCKER_HOST=unix:///var/run/docker.sock
      - LAMBDA_DOCKER_NETWORK=development

    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock"

    networks:
      - development

networks:
  development:
    name: development
    driver: bridge
```

### Environment

```markdown
- OS: Windows
- LocalStack: 2.2.1.dev
```

#### Observations:

- AWS verifies the uniqueness of an event source mapping based on the SQS ARN and the Lambda function.
- In LocalStack, it's currently possible to create an unlimited number of identical event source mappings using the same SQS ARN - and Lambda functions.

PR #9996, opened 2024-01-04
fix lambda create event source mapping duplicates

thrau:
## Motivation

Fixes https://github.com/localstack/localstack/issues/9219, where a user noted that duplicate event source mappings are not handled correctly. Of course exception messages differ across services, because why make things consistent :wink: 

Tested for sqs, kinesis, and dynamodbstreams.

## Changes

* lambda `create_event_source_mapping` now raises a `ResourceConflictException` if a mapping with the same event source arn and function arn already exists.
* added a `create_event_source_mapping` fixture that does cleanup

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -133,6 +133,7 @@ from localstack.aws.api.lambda_ import (
 from localstack.aws.api.lambda_ import FunctionVersion as FunctionVersionApi
 from localstack.aws.api.lambda_ import ServiceException as LambdaServiceException
 from localstack.aws.connect import connect_to
+from localstack.aws.spec import load_service
 from localstack.services.edge import ROUTER
 from localstack.services.lambda_ import api_utils
 from localstack.services.lambda_ import hooks as lambda_hooks
@@ -1657,7 +1658,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         if "EventSourceArn" not in request:
             raise InvalidParameterValueException("Unrecognized event source.", Type="User")
 
-        service = extract_service_from_arn(request.get("EventSourceArn"))
+        service = extract_service_from_arn(request["EventSourceArn"])
         if service in ["dynamodb", "kinesis", "kafka"] and "StartingPosition" not in request:
             raise InvalidParameterValueException(
                 "1 validation error detected: Value null at 'startingPosition' failed to satisfy constraint: Member must not be null.",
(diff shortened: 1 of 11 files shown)


dfangl (approved):
Of course it is different per event source :)

Looks good, good catch!

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
```diff
@@ -140,6 +141,77 @@ def _send_and_receive_events():
...
+        snapshot.match("error", response)
+
+        # this should work without problem since it's a new function
+        create_lambda_function(
+            handler_file=TEST_LAMBDA_PYTHON_ECHO,
+            func_name=function_name_2,
+            runtime=Runtime.python3_9,
+            role=lambda_su_role,
+        )
+        aws_client.lambda_.create_event_source_mapping(
```
dfangl, line 209: We need to remove the event source mappings as well. Just saw the test errors, and they seem related to the missing cleanup!

dfangl (changes requested):
Sorry for the rash review, I did not quite catch that before looking at the tests. it seems the event source mappings are not removed, but we should do that. It seems especially kinesis does not like it otherwise at all.

localstack/services/lambda_/provider.py
```diff
@@ -1696,6 +1697,31 @@ def create_event_source_mapping(
...
+        # check for event source duplicates
+        # TODO: currently validated for sqs, kinesis, and dynamodb
+        service_id = load_service(service).service_id
+        for uuid, mapping in state.event_source_mappings.items():
+            if (
+                mapping["FunctionArn"] == fn_arn
+                and mapping["EventSourceArn"] == request["EventSourceArn"]
+            ):
+                if service == "sqs":
+                    # *shakes fist at SQS*
```
joe4dev, line 1709: 😢

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
thrau: good point! added a create_event_source_mapping fixture that does cleanup

localstack/testing/pytest/fixtures.py
```diff
@@ -1291,6 +1291,24 @@ def _create_function():
...
+    def _create_event_source_mapping(*args, **kwargs):
+        response = aws_client.lambda_.create_event_source_mapping(*args, **kwargs)
+        uuids.append(response["UUID"])
+        return response
+
+    yield _create_event_source_mapping
+
+    for uuid in uuids:
+        try:
+            aws_client.lambda_.delete_event_source_mapping(UUID=uuid)
```
joe4dev, line 1307: 🙏 thanks for fixing this resource leak 👍

Adarsha-1994:
Hi Team 
I got a ResourceConflictException error when creating second event-source-mapping for MSK trigger with different topic and consumer group on the same lambda function, 
why can't we create second event-source-mapping for the same FunctionArn and EventSourceArn with different topic and consumer group?
Screenshot 2024-11-21 at 6 05 53 PM

PR #9997, opened 2024-01-04
add SNS signing of messages

bentsku:
## Motivation
SNS supports signing of the message with asymmetric signing:
https://docs.aws.amazon.com/sns/latest/dg/sns-verify-signature-of-message.html
The best data about it is in this release note PDF: http://sns-public-resources.s3.amazonaws.com/SNS_Message_Signing_Release_Note_Jan_25_2011.pdf

It does so by signing the message by creating a canonical string with some fields of the message, then sign it with the private key of a cert. 
The link to the cert is available in the message as one of the message field. You can then download the cert and validate your message signature with the certificate.

This feature has been requested and upvoted by several users.

## Changes
This PR implements the message signing with a certificate that is generated when loading the SNS provider.
- The certificate is available with a public endpoint exposed with a router handler (at `/_aws/sns/SimpleNotificationService-00000000000000000000000000000000.pem`).
- SNS will now properly sign messages, then the client can verify the signature by downloading the certificate. 
- Created 2 AWS validated tests for SQS and Lambda, for Signature v1 and v2. 
- added unit tests for the new utils

_fixes #5454_
## Limitations
When validating the signature of messages kept in queue after restoring persistence, the certificate won't be same anymore, and the signature validation will fail. This scenario won't be too frequent and it's a limitation we can accept for now as a first round for SNS Signature. We can tackle persistence in a second round.

## Resources:
- Padding used by SNS in C# SDK util, used to know which padding type to use to verify/sign the message: https://github.com/aws/aws-sdk-net/blob/98a48c7b735a9863700d2aea00bcd936981f6daf/sdk/src/Services/SimpleNotificationService/Custom/Util/Message.cs#L309
  - taken from this SO question: https://stackoverflow.com/questions/77649510/how-can-i-verify-sns-messages-in-a-server-side-swift-https-endpoint
- AWS Documentation:
  - https://aws.amazon.com/blogs/security/sign-amazon-sns-messages-with-sha256-hashing-for-http-subscriptions/
  - http://sns-public-resources.s3.amazonaws.com/SNS_Message_Signing_Release_Note_Jan_25_2011.pdf
- `cryptography` library documentation for signing using x509 cert with RSA private/public keys: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#signing


thrau (commented):
This looks really cool, congrats!

I have two questions (without having looked into it too much):
* How will creating a new sns server cert every time on startup affect persistence? will restored messages have invalid signatures?
* I don't fully understand the reasoning behind the endpoint `/_aws/sns/SimpleNotificationService-00000000000000000000000000000000.pem`. Could you explain? why not just `/_aws/sns/SimpleNotificationService.pem` since it's static? if it's not static, why not parameterize the ID?

bentsku:
Thanks! 

> * How will creating a new sns server cert every time on startup affect persistence? will restored messages have invalid signatures?

True, good point, I forgot to talk about this limitation in the PR description but was aware of it. Exactly, yes, when trying to validate the signature, they would be wrong. I thought this was an acceptable limitation for now, knowing we didn't have signature validation at all before, and I thought that `cryptography` objects were often tricky to pickle. 
However, I could spike some persistence for the cert by storing its string representation I think. I'll spike something out today. 

> * I don't fully understand the reasoning behind the endpoint `/_aws/sns/SimpleNotificationService-00000000000000000000000000000000.pem`. Could you explain? why not just `/_aws/sns/SimpleNotificationService.pem` since it's static? if it's not static, why not parameterize the ID?

This was to follow the same structure as AWS. It does look a bit weird tbh (some static hex char would look better already). If I implement persistence, I could parametrize the ID, if a cert becomes too old (well, it's supposed to last a year), I could have 2 certs active like in AWS and separate them by ID. Will work on that with persistence

thrau:
@bentsku i think it's OK to merge without persistence if it's documented as limitation

bentsku:
@thrau okay, thanks, I'll start to draft a PR in the docs about the feature and the limitations. I'll update the cert ID to have some hex characters in there to respect the format but look less weird. 😄

thrau (approved):
nice! well done :100:

Issue #9832, opened 2023-12-08
bug: Localstack doesn't work properly with SQS client from AWS SDK v2 version 2.21.18 and above


mihalyr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

My issue is described in this issue on AWS SDK v2 https://github.com/aws/aws-sdk-java-v2/issues/4759

I thought it is an SDK bug as it appeared once I updated the SDK to 2.21.18 or above. But it seems that SDK works with SQS service and it is only breaking functionality when used with Localstack. I think there was a protocol change in 2.21.18 on SQS side that might not be implemented on Localstack. 

This problem is now causing failures all over the SQS integration tests using Localstack. 

### Expected Behavior

* SQS SDK `SendMessageBatchResponse.hasFailed()` should return `false` when all messages were sent successfully (no failed items) -> this is returning `true` on SDK 2.21.18+ with Localstack
* SQS SDK `ReceiveMessageResponse.hasMessages()` should return `false` when there were no messages received -> this is returning `true` on SDK 2.21.18+ with Localstack

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Using TestContainers

```java
  static final LocalStackContainer localStack = new LocalStackContainer(localStackImage)
      .withEnv("DYNAMODB_IN_MEMORY", "1")
      .withEnv("SQS_DISABLE_CLOUDWATCH_METRICS", "1")
      // Set LOCALSTACK_HOST explicitly to avoid issues running docker-in-docker in CI, where TestContainers
      // might automatically try to map the dind container's hostname to it
      .withEnv("LOCALSTACK_HOST", "localhost")
      // Bind init scripts to localstack container
      // These will ensure all necessary stacks are created before the tests run
      .withCopyFileToContainer(MountableFile.forHostPath(LOCALSTACK_INIT_SCRIPT), LOCALSTACK_INIT_SCRIPT_MOUNT)
      .withCopyFileToContainer(MountableFile.forHostPath(JSON_STORE_STACK_TEMPLATE), JSON_STORE_STACK_TEMPLATE_MOUNT)
      .withCopyFileToContainer(MountableFile.forHostPath(ID_QUEUE_STACK_TEMPLATE), ID_QUEUE_STACK_TEMPLATE_MOUNT)
[...]

PR #9998, opened 2024-01-04
fix sqs SendMessageBatch result for SDK compatibility

thrau:
## Motivation

In #9832, a user noted that the Java SDK v2 method `hasFailed()` of the `SendMessageBatchResponse` would always return `true`, even though there are no failures in the batch.
This is the way they're doing it:

```java
    public final boolean hasFailed() {
        return this.failed != null && !(this.failed instanceof SdkAutoConstructList);
    }
```

Basically, the SDK expects a null value, but we're sending empty lists. I opportunistically updated an existing test to assert this accordingly, and re-validated it.

## Changes

* `SendMesageBatchResult` no longer returns empty lists in `Failed` or `Successful` fields

### DIFF
--- a/localstack/services/sqs/provider.py
+++ b/localstack/services/sqs/provider.py
@@ -992,9 +992,10 @@ class SqsProvider(SqsApi, ServiceLifecycleHook):
                         )
                     )
 
+        # some SDKs expect ``Failed`` to not be present at all
         return SendMessageBatchResult(
-            Successful=successful,
-            Failed=failed,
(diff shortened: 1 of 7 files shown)


thrau:
ben pointed out that this may be part of the query/rest protocol duality. any idea how to solve this in a generalized way @alexrashed ? will #9710 affect this?

thrau:
Actually looks like this fix is already part of #9710, so closing it :-)

PR #9999, opened 2024-01-04
improve S3 aws-chunked detection

bentsku:
## Motivation
This might be a fix to #9989, I could not reproduce the issue yet, but I suspect this could be one of the reason: we did not detect the payload was of `aws-chunked` format. We once set the detection to only be about the `Content-Encoding`, but it introduced a regression for some client SDKs, I think it was .NET. See #8712
So we can combine both detection to be sure. 

## Changes
Added a test with a similar format as the reported issue, with no signing of the chunked payload. 
Improved the detection to check both the content encoding of the request and the special header AWS uses to indicate it's an `aws-chunked` payload even when the content-encoding is not set.

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -630,7 +630,9 @@ class S3Provider(S3Api, ServiceLifecycleHook):
         body = request.get("Body")
         # check if chunked request
         headers = context.request.headers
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Even though it's a blind fix, the fact that this solution now combines multiple heuristics on determining `aws-chunked` payloads makes them more robust (and the test is looking good)! 🚀 💯

PR #10004, opened 2024-01-05
add raw-metrics endpoint for v2

steffyP:
## Motivation
Adds the raw-metrics endpoint for the cloudwatch v2 provider, which was introduced in #3535 (customer request). 

## Changes
* in v1 the endpoint currently requires aws headers, which seems a bit odd
  * for v2 removed this requirement, instead we also return the account-id and region for each datapoint
  * it currently returns all the single-metric data (like it did before)  

## TODO

In a future iteration we should decide if we want to keep this endpoint, and/or collect data about how many people are actually using it. 
Considering that a AWS header is currently required to access the endpoint, I don't think it is actually used (it defeats the actual purpose). 

In case we keep the endpoint:
- [ ] add paging (there may be a huge amount of data)
- [ ] add aggregated metric data (for v1/v2 currently only single metric data is returned)

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -351,3 +351,34 @@ class CloudwatchDatabase:
         self, timestamp: datetime
     ):  # TODO verify if this is the standard format, might need to convert
         return int(timestamp.timestamp())
+
+    def get_all_metric_data(self):
+        with sqlite3.connect(self.METRICS_DB) as conn:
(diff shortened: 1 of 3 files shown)


steffyP:
The two failing tests are due fixes/features on master, that haven't been added to our branch yet. Not related to this PR (we create a new one to add patches).
- TestCloudwatch::test_alarm_lambda_target
- TestSQSMetrics::test_alarm_number_of_messages_sent

silv-io (approved):

pinzon (approved):
LGTM

PR #10005, opened 2024-01-05
fix unit-handling for get-metric-statistics

steffyP:
## Motivation
This PR fixes two issues revealed by tests that concern the returned unit for `get-metric-statistics`:
 1) the unit is returned, even if was not part of the query
 2) if no unit is provided for the query, but there are several units that match, the result is aggregated on `unit`

## Changes
* add a query-helper to retrieve the unique units that will match a query
* run the statistic for each unit, aggregate the returned result
* enhanced the test `get_metric_statistics_with_different_units` by adding also metric-data that does not have a `unit`

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -159,6 +159,58 @@ class CloudwatchDatabase:
 
             conn.commit()
 
+    def get_units_for_metric_data_stat(
+        self,
+        account_id: str,
+        region: str,
+        start_time: datetime,
+        end_time: datetime,
+        metric_name: str,
+        namespace: str,
+    ):
+        with sqlite3.connect(self.METRICS_DB) as conn:
+            cur = conn.cursor()
+
+            # prepare SQL query
+            start_time_unix = self._convert_timestamp_to_unix(start_time)
+            end_time_unix = self._convert_timestamp_to_unix(end_time)
+
+            data = (
+                account_id,
+                region,
+                namespace,
+                metric_name,
+                start_time_unix,
+                end_time_unix,
+            )
+
+            sql_query = f"""
+            SELECT GROUP_CONCAT(unit) AS unit_values
+            FROM(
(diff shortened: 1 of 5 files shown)


steffyP:
the two failing tests will be fixed in another PR (related to fixes/feature on master, that are not yet in v2)
- TestCloudwatch::test_alarm_lambda_target 
- TestSQSMetrics::test_alarm_number_of_messages_sent

pinzon (approved):
LGTM

localstack/services/cloudwatch/cloudwatch_database_helper.py
```diff
@@ -159,6 +159,58 @@ def _get_current_unix_timestamp_utc():
...
+                AND timestamp >= ? AND timestamp < ?
+            ) AS subquery
+            """
+
+            cur.execute(
+                sql_query,
+                data,
+            )
+            result_row = cur.fetchone()
+            return result_row[0].split(",") if result_row[0] else ["NULL_VALUE"]
```
silv-io, line 212: could this theoretically result in an error if the result is not subscriptable?

sqlite3 docs:

> If [row_factory](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.row_factory) is None, return the next row query result set as a [tuple](https://docs.python.org/3/library/stdtypes.html#tuple). Else, pass it to the row factory and return its result. Return None if no more data is available.

None[0] would result in an error I think

silv-io (approved):
LGTM, just one small question/nitpick :)

localstack/services/cloudwatch/cloudwatch_database_helper.py
steffyP: @silv-io: thanks for the pointer :) i checked it once more:
The query will return `NULL` in case there are no data, meaning that the `cur.fetchone()` will return a tuple `(None,)`. The call should be safe therefore. 
Subsequent calls using `cur.fetchone()` would return `None` however (as there are no more rows to fetch).

PR #10008, opened 2024-01-06
fix support calling sqs from api ws gw on disconnect endpoint

calvernaz:
## Motivation

The user issue was related to a call to the "$disconnect" route and not triggering the SQS backend integration. The problem is we didn't support WS+SQS integration until now. This PR fixes that.

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -119,6 +119,23 @@ class BackendIntegration(ABC):
                 response.headers[header_name] = value.strip("'")
         return response
 
+    @classmethod
+    def render_template_selection_expression(cls, invocation_context: ApiInvocationContext):
+        integration = invocation_context.integration
+        template_selection_expression = integration.get("templateSelectionExpression")
+        if not template_selection_expression:
+            return "$default"
+        variables = {
+            "request": {
+                "header": invocation_context.headers,
+                "querystring": invocation_context.query_params(),
+                "body": invocation_context.data_as_string(),
+                "context": invocation_context.context or {},
+                "stage_variables": invocation_context.stage_variables or {},
+            }
+        }
+        return VtlTemplate().render_vtl(template_selection_expression, variables)
+
 
 @lru_cache(maxsize=64)
 def get_service_factory(region_name: str, role_arn: str):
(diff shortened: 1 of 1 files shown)


localstack/services/apigateway/integration.py
```diff
@@ -682,12 +698,17 @@ def invoke(self, invocation_context: ApiInvocationContext):
         if "Accept" not in headers:
             headers["Accept"] = "application/json"
 
-        template = integration.get("requestTemplates").get(APPLICATION_JSON)
-        if "GetQueueUrl" in template or "CreateQueue" in template:
+        if invocation_context.is_websocket_request():
+            template_key = self.render_template_selection_expression(invocation_context)
+            payload = self.request_templates.render(invocation_context, template_key)
```
bentsku, line 703: I think this should be the following, from what I've seen in our call yesterday:
```suggestion
            payload = self.render_template_selection_expression(invocation_context)
```

localstack/services/apigateway/integration.py
```diff
@@ -466,7 +483,6 @@ def invoke(self, invocation_context: ApiInvocationContext):
             # integration type "AWS" is only supported for WebSocket APIs and REST
             # API (v1), but the template selection expression is only supported for
             # Websockets
-            template_key = None
             if invocation_context.is_websocket_request():
                 template_key = invocation_context.integration.get(
                     "TemplateSelectionExpression", "$default"
```
bentsku, line 488: q: should this also be replace with `render_template_selection_expression`?

localstack/services/apigateway/integration.py
```diff
@@ -119,6 +119,23 @@ def apply_response_parameters(
                 response.headers[header_name] = value.strip("'")
         return response
 
+    @classmethod
+    def render_template_selection_expression(cls, invocation_context: ApiInvocationContext):
```
bentsku, line 123: nit: I'm not too fond of mixing websockets logic in community as we've talked offline, but I believe for now, it's the quickest way, as it's already done for kinesis anyway. We will refactor this at some point!

bentsku (commented):
Thanks a lot for tackling this! I believe there's a typo in the render logic, we render it to pass as the `template_key`. See the comments. Thanks!

localstack/services/apigateway/integration.py
calvernaz: no this is correct, the template_key is rendered and becomes the request template

localstack/services/apigateway/integration.py
calvernaz: yes, can be done on this PR as well.

localstack/services/apigateway/integration.py
bentsku: Ah my bad! I see it now! Thanks a lot 👍

bentsku (approved):
LGTM! Sorry for the wrong comment.

PR #10011, opened 2024-01-07
refactor http router for better werkzeug compatibility

thrau:
## Motivation

Werkzeug has clever utilities such as `Submount` which allows you to streamline the creation of routers:

```python
url_map = Map([
    Rule('/', endpoint='index'),
    Submount('/blog', [
        Rule('/', endpoint='blog/index'),
        Rule('/entry/<entry_slug>', endpoint='blog/show')
    ])
])
```

Our `@route`, `@resource`, or endpoint concepts have not been compatible with these, because the werkzeug `Rule` objects are created within `Router.add`, instead of an encapsulating `RuleFactory`, which is werkzeug's mechanism of building higher-level routing utilities.

This PR moves all the code that generates `Rule` objects out of `Router.add`, and instead encapsulates them into `RuleFactory` classes. This makes them naturally compatible with werkzeug.

For instance, now you can do something like:

```python
@route("/v1/hello", methods=["GET"])
def hello(request: Request, _args):
    return "hello"

@route("/v1/world", methods=["GET"])
def world(request: Request, _args):
    return "world"

def main():
    router = Router()

    api = RuleAdapter([
        hello,
        world,
    ])
    
    router.add([
        WithHost("myextension.localhost.localstack.cloud:4566", [api]),
        Submount("/_ext/myextension", [api]),
    ])
```

Where you'll be able to call:
* `localhost:4566/_ext/myextension/v1/hello`
* `myextension.localhost.localstack.cloud:4566/v1/hello`

This is going to be very useful to implement higher-level patterns like in extensions. For example, for the openai mock extension https://github.com/localstack/localstack-extensions/pull/36, there are several routes that all have hardcoded routes prefixed with `/v1/...`. We wouldn't want to expose this via `localhost.localstack.cloud:4566/v1/...` because of potential collisions, but would be ok through `openai.localhost...` or `localhost.localstack.cloud:4566/openai-mock/v1/...`.
The utilities of this PR allows you to easily do that with a combination of `WithHost` and `Submount`, as in the example above.

## Changes

* the logic to create werkzeug `Rule` objects from endpoints and other localstack concepts is now in a `RuleAdapter` class that is compatible with werkzeug's `RuleFactory`
* Removed deprecated methods from `Router` and refactored existing code accordingly

### DIFF
--- a/localstack/http/router.py
+++ b/localstack/http/router.py
@@ -509,6 +509,14 @@ class RuleGroup(RuleFactory):
 
 
 class _EndpointRule(RuleFactory, Generic[E]):
+    """
+    Generates default werkzeug ``Rule`` object with the given attributes. Additionally, it makes sure that
+    the generated rule always has a default host value, if the map has host matching enabled. Specifically,
+    it adds the well-known placeholder ``<__host__>``, which is later stripped out of the request arguments
+    when dispatching to the endpoint. This ensures compatibility of rule definitions across routers that
+    have host matching enabled or not.
+    """
+
     def __init__(
         self,
         path: str,
@@ -540,6 +548,19 @@ class _EndpointRule(RuleFactory, Generic[E]):
 
 
 class _EndpointFunction(RuleFactory):
+    """
+    Internal rule factory that generates router Rules from ``@route`` annotated functions, or anything else
+    that can be interpreted as a ``_RouteEndpoint``. It extracts the rule attributes from the
+    ``_RuleAttributes`` attribute defined by ``_RouteEndpoint``. Example::
+
+        @route("/my_api", methods=["GET"])
+        def do_get(request: Request, _args):
+            # should be inherited
+            return Response(f"{request.path}/do-get")
+
+        router.add(do_get)  # <- will use an _EndpointFunction RuleFactory.
+    """
+
     def __init__(self, fn: _RouteEndpoint):
         self.fn = fn
 
--- a/localstack/http/router.py
+++ b/localstack/http/router.py
@@ -1,6 +1,7 @@
 import functools
 import inspect
 import threading
+import typing as t
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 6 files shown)


localstack/http/router.py
```diff
@@ -462,23 +373,223 @@ def route(
...
+
+        # the typing for endpoint is a str, but the doc states it can be any value,
+        # however then the redirection URL building will not work
+        rule = Rule(
+            self.path, endpoint=self.endpoint, methods=self.methods, host=host, **self.kwargs
+        )
+        yield rule
+
+
+class _EndpointFunction(RuleFactory):
```
alexrashed, line 542: Maybe you could add a small docstring explaining that this class is used for callables which have been decorated with the route decorator (which will take care of setting the expected `rule_attributes` as a member of the callable object)?

localstack/http/router.py
```diff
@@ -462,23 +373,223 @@ def route(
...
+
+    def get_rules(self, map: Map) -> t.Iterable[Rule]:
+        for rule in self.rules:
+            yield from rule.get_rules(map)
+
 
-    def remove_rule(self, rule: Rule):
-        """DEPRECATED: use ``remove`` instead."""
-        self._remove_rules([rule])
+class _EndpointRule(RuleFactory, Generic[E]):
```
alexrashed, line 511: This is a nice encapsulation of the creation of rules based on the rule attributes (which was previously in the `_add_endpoint` function). Maybe you could also add a small docstring here?

localstack/http/router.py
```diff
@@ -462,23 +373,223 @@ def route(
...
+        elif "fn" in kwargs or callable(args[0]):
+            self.factory = _EndpointFunction(*args, **kwargs)
+        elif "rule_factory" in kwargs:
+            self.factory = kwargs["rule_factory"]
+        elif isinstance(args[0], RuleFactory):
+            self.factory = args[0]
+        elif isinstance(args[0], list):
+            self.factory = RuleGroup([RuleAdapter(rule) for rule in args[0]])
+        else:
+            self.factory = _EndpointsObject(*args, **kwargs)
```
alexrashed, line 483: This is a nice encapsulation of the different possibilities of defining rules directly using Werkzeug classes! 💯

alexrashed (approved):
Wow, really nice refactoring! This is a really great enhancement, unlocking quite some possibilities to use more features from Werkzeug directly! This will definitely simplify the rule definition and increase their readibility! 💯

PR #10016, opened 2024-01-08
Add config option to control maximum gateway worker thread count

dfangl:
## Motivation
On very high (burst) load scenarios, we can get connection refused errors quite quickly, if there are no worker threads available.

We should make it configurable, to see if we can bypass some of them.

## Changes
* Add config option `GATEWAY_WORKER_COUNT` to control maximum number of worker threads

## TODO

What's left to do:

- [ ] Test real life impact of this for certain high connection count scenarios
- [x] Discuss naming of the variable
- [x] Discuss where to set the variable.. We might want to opt to set it in the GatewayServer (in http/hypercorn.py) instead, or even in the serve_gateway variable, to limit the variables use to the "main gateway".

### DIFF
--- a/localstack/aws/serving/asgi.py
+++ b/localstack/aws/serving/asgi.py
@@ -3,7 +3,6 @@ import concurrent.futures.thread
 from asyncio import AbstractEventLoop
 from typing import Optional
 
-from localstack import config
 from localstack.aws.gateway import Gateway
 from localstack.aws.serving.wsgi import WsgiGateway
 from localstack.http.asgi import ASGIAdapter, ASGILifespanListener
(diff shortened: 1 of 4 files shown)


localstack/aws/serving/asgi.py
```diff
@@ -39,14 +39,14 @@ def __init__(
...
-        threads: int = 1000,
+        threads: int = None,
         lifespan_listener: Optional[ASGILifespanListener] = None,
         websocket_listener=None,
     ) -> None:
         self.gateway = gateway
 
         self.event_loop = event_loop or asyncio.get_event_loop()
-        self.executor = _ThreadPool(threads, thread_name_prefix="asgi_gw")
+        self.executor = _ThreadPool(threads or 1000, thread_name_prefix="asgi_gw")
```
thrau, line 49: much better!

localstack/config.py
```diff
@@ -655,6 +655,8 @@ def populate_edge_configuration(
     GATEWAY_LISTEN,
 ) = populate_edge_configuration(os.environ)
 
+GATEWAY_WORKER_THREAD_COUNT = int(os.environ.get("GATEWAY_WORKER_THREAD_COUNT") or 1000)
```
thrau, line 658: naming seems fine to me :+1: could argue `GATEWAY_WORKER_COUNT` to make it more generic but threads is more accurate :+1:

thrau (approved):
nice! great addition :-)

Issue #9854, opened 2023-12-12
EC2.describeInstanceTypes outputs invalid data for EnaSupport


adam-fowler:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`EC2.describeInstanceTypes` outputs the value `False` for `InstanceTypes[].NetworkInfo.EnaSupport` in the NetworkInfo type

### Expected Behavior

It should return one of "required", "supported", "unsupported". See AWS reference documentation https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInfo.html

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal ec2 describe-instance-types  

### Environment

```markdown
- OS: macOS 14.1.1
- LocalStack: 3.0.2.dev
```

PR #10017, opened 2024-01-08
Bump moto-ext to 5.0.0alpha1.post1

viren-nadkarni:
This is an incremental update in Moto with no breaking changes despite the major version bump.

Brings in:
- https://github.com/getmoto/moto/pull/7134

Closes https://github.com/localstack/localstack/issues/9854

Ext Integration Tests # 4246 :green_circle:


alexrashed (approved):
Nice! Thanks for keeping `moto` up to date! 💯 🏅 
I think the failed test step is unrelated, but it would be great to see if we could get a green build before merging...

PR #10019, opened 2024-01-08
add extension mechanism for exceptions

thrau:
## Motivation

The need has come up in extensions to be able to hook in custom exception handlers into the handler chain. Specifically, @ackdav needs this for the snowflake extension.

This PR extends the handler chain with a composite exception handler that can be used in the same way as other handlers, and adds an `update_exception_handlers` method to the `Extension` interface.

## Changes

* the handler chain now has a composite handler for custom exception handlers
* the extension api now supports adding exception handlers into this composite


thrau:
great points @bentsku! let's come back to this discussion soon

PR #9985, opened 2024-01-03
add lambda alarm action for cloudwatch

steffyP:
## Motivation
AWS recently [added support for AWS Lambda as an alarm state action](https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-cloudwatch-alarms-lambda-change-action/). 
This PR enables Lambda targets for alarm actions on LocalStack. 

## Changes
* added snapshot test, that triggers a lambda on state ALARM
* implemented the lambda invocation for LS

PR #9878, opened 2023-12-15
CloudWatch: Use SNS client with same region as topic

viren-nadkarni:
## Motivation

SNS requires the Boto client to be in the same region as the topic (See test introduced in https://github.com/localstack/localstack/pull/9679)

This PR fixes CloudWatch <> SNS integration which was broken and detected in https://github.com/localstack/localstack/pull/8204.

Detected by:

- [ ]  tests.aws.services.cloudwatch.test_cloudwatch.TestCloudwatch
    - [ ]  test_set_alarm
    - [ ]  test_breaching_alarm_actions
    - [ ]  test_enable_disable_alarm_actions

PR #9799, opened 2023-12-04
fix: fix failing dashboard lifecycle test for multi-account and region

sannya-singal:
## Motivation
When using values other than `000000000000` for account ID or `us-east-1` for region, `cloudwatch` tests should still create the consequent resources in corresponding accounts and region.

## Changes
This PR fixes all the failing tests `tests.aws.services.cloudwatch.test_cloudwatch.TestCloudwatch.test_dashboard_lifecycle` by replacing the static value of region `us-east-1` with `TEST_AWS_REGION_NAME`. 

## Testing

The tests were failing previously when executed with a non-default account ID, set through environment variables (`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`).

PR #10026, opened 2024-01-09
Integrate recent v1 CW changes into new v2 provider

baermat:
## Motivation
This PR integrates recent v1 CW changes into the new v2 Provider

## Changes
This PR ensures that the following v1 changes are reflected in the new v2 provider
- https://github.com/localstack/localstack/pull/9641/files#diff-23bd8190ec59c94fb372da30e2453223f10581711317648b7792fd76ba6c8cb0
- https://github.com/localstack/localstack/pull/9799
- https://github.com/localstack/localstack/pull/9878
- https://github.com/localstack/localstack/pull/9985
- https://github.com/localstack/localstack/pull/9978

### DIFF
--- a/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack/services/cloudwatch/provider_v2.py
@@ -80,6 +80,7 @@ from localstack.utils.aws import arns
 from localstack.utils.aws.arns import extract_account_id_from_arn, lambda_function_name
 from localstack.utils.collections import PaginatedList
 from localstack.utils.json import CustomEncoder as JSONEncoder
+from localstack.utils.strings import camel_to_snake_case
 from localstack.utils.sync import poll_condition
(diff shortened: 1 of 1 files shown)


steffyP (approved):
Thanks for taking care that all the relevant changes from v1 that happened on master in the meantime are now ported to v2, @baermat! 🚀

Also kudos for spotting a weird behavior, with the lamda-payload, and making me aware of it 👍 

For document purpose, as we discussed this in person: 
* the payload sets `dimensions` by default to a dict, when it should be a list. 
   ```
    "dimensions": alarm.alarm.get("Dimensions") or {},
    ```  
* We do have an AWS validated test for this, but we should check what is returned when we actually have a dimensions-filter set for the alarm. 
* It maybe a weird AWS issue as well, considering this feature is very new.  
* We will follow-up on this in a separate PR (it also concerns v1).

PR #10028, opened 2024-01-09
add a call_safe_with_result function

giograno:
## Motivation
`call_safe` is a utility function that invokes a provided `Callable` (`func`) and returns its return value or logs an exception, if raised.
While exceptions get logged, it would be handy to programmatically know if one is actually raised (e.g., [here](https://github.com/localstack/localstack-ext/blob/master/localstack_ext/persistence/manager.py)). 
A simple solution would have been to check for a `None` return from `call_safe`, which would imply an exception.
However, this assumption does not hold if `func` returns `None` even in successful cases.

## Changes
- implemented a function `call_safe_with_result` function that returns a `Result` object, that can either return the value or the exception for the called `func`;
- `call_safe` now uses the `call_safe_with_result` function;
- adding a few small unit tests.

### DIFF
--- a/localstack/utils/functions.py
+++ b/localstack/utils/functions.py
@@ -2,7 +2,7 @@
 import functools
 import inspect
 import logging
-from typing import Any, Callable, Dict, Optional, Tuple
+from typing import Any, Callable, Dict, NamedTuple, Optional, Tuple
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


thrau (commented):
just trying to understand the value

if you're doing something like
```python
result = call_safe_with_result(fn, arg1, arg2)
if result.error:
    ...
```

wouldn't it be more pythonic to do
```python
try:
    fn(arg1, arg2)
except Exception:
    ...
```

can you show an example where this is more useful than the standard block?

PR #10012, opened 2024-01-07
fix flaky sns unit test

thrau:
## Motivation

A flake showed up in the sns untit test `test_canonical_string_calculation`: https://app.circleci.com/pipelines/github/localstack/localstack/21448/workflows/caff7afc-1a8b-4ec3-87b9-45e83aa06e90/jobs/171800/tests

```
AssertionError: assert 'Message\ntes...otification\n' == 'Message\ntes...otification\n'
    Message
    test content
    MessageId
    abdcdef
    Subject
    random
    Timestamp
  - 2024-01-07T20:43:26.204Z
  ?                       ^
  + 2024-01-07T20:43:26.205Z
  ?                       ^
    TopicArn
    arn
    Type
    Notification
```

looks like tiny variations between timestamps

## Changes

* use the created timestamp throughout the test

PR #10036, opened 2024-01-10
implement apigw request validation for parameters (header/path/qs)

bentsku:
## Motivation
As requested in #10012, we did not support request validation for API Gateway yet for parameters, aka headers/path/query string. 

## Changes
This PR implements the validation, as well as updates a test to properly assert the new feature.

The original test did not run anymore against AWS, most probably because it needed more time between every operation, and things needed more retries. I tested another test related to request validation and updated the snapshot at the same time.

As a side note, I've started implementing different exception for this use case, as those are specified for `GatewayResponses`: https://docs.aws.amazon.com/apigateway/latest/developerguide/supported-gateway-response-types.html

Basically, we should start properly instantiating the kind of exception defined there, so that we could catch them globally and return a GatewayResponse instead. This also allows to pass more context through the different layer of callers we have.


localstack/services/apigateway/invocations.py
```diff
@@ -60,45 +83,50 @@ def __init__(self, context: ApiInvocationContext, store: ApiGatewayStore = None)
...
-            if not is_valid_parameters:
-                return is_valid_parameters
+        if self.should_validate_body(validator) and not self.validate_body(resource):
+            raise BadRequestBody(message="Invalid request body")
 
-        return True
+        return
 
-    def validate_body(self, resource):
+    def validate_body(self, resource) -> bool:
```
calvernaz, line 121: should follow the same command name pattern, either use `is_body_valid` and returns a bool or `validate_body` and does return anything.

localstack/services/apigateway/invocations.py
```diff
@@ -127,9 +155,32 @@ def validate_body(self, resource):
             LOG.warning("failed to validate request body, request data is not valid JSON %s", e)
             return False
 
-    # TODO implement parameters and headers
-    def validate_parameters_and_headers(self, resource):
-        return True
+    def validate_parameters_and_headers(self, resource) -> list[str]:
```
calvernaz, line 158: maybe `show` or `list` missing params

localstack/services/apigateway/invocations.py
```diff
@@ -226,8 +277,10 @@ def invoke_rest_api(invocation_context: ApiInvocationContext):
 
     # validate request
     validator = RequestValidator(invocation_context)
-    if not validator.is_request_valid():
-        return make_error_response("Invalid request body", 400)
+    try:
+        validator.validate_request()
```
calvernaz, line 281: nice!

calvernaz (approved):
Thanks for tackling this, LGTM! 

Small nits but not blockers!

localstack/services/apigateway/invocations.py
bentsku: yes you're right! I've renamed into `_is_body_valid`, but for the parameters and headers, as the logic needed is different, I'll rename it into `_get_missing_required_parameters`. Not fan of the naming but didn't get more inspiration.

Issue #10015, opened 2024-01-08
bug: API WS Gateway does not provide sourceIp address as part of lambda payload


kalitsinskyi81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

API WS Gateway does not provide sourceIp address as part of lambda payload, just empty string
```
{
    "requestContext": {
        "identity": {
            "sourceIp": ""
        },
    }
}
```

But AWS has
<img width="362" alt="image" src="https://github.com/localstack/localstack/assets/108662648/26771f38-a964-4754-92eb-04205e1e2837">

### Expected Behavior

be able get real ip address in ipv4 format from this path `request.requestContext.identity.sourceIp`

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

```
  localstack:
    profiles: # explicit start only
      - localstack
    container_name: "${LOCALSTACK_DOCKER_NAME-localstack_main_local}"
    image: localstack/localstack-pro:3.0.1
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4550:4510-4550"  # external services port range
      - "127.0.0.1:53:53"                # DNS config (required for Pro)
      #- "127.0.0.1:53:53/udp"           # do not work after Docker update, DNS config (required for Pro)
      - "127.0.0.1:443:443"              # LocalStack HTTPS Gateway (required for Pro)
    environment:
      - SERVICES=apigateway,lambda,dynamodb,sqs,sns,s3,iam,sts,logs,ec2,ssm
      - DOCKER_HOST=unix:///var/run/docker.sock
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN}
      - DEBUG=1
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./localstack-volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

no opportunity to provide exact list of cli commands, it uses terraform/terragrunt for deploying, but the main flow is

1. create lambda which logs all the input
2. create API Websocket Gateway (v2)
3. assign created lambda on API GW `$connect` endpoint
4. call API WS GW as a client with any tool like `wscat`
5. observe absence in lambda logs empty value in path `request.requestContext.identity.sourceIp`

### Environment

```markdown
- OS: MacOS 14.2
- LocalStack: 3.0.1
```

thrau:
[...]

PR #10038, opened 2024-01-10
fix source ip value for the lambda event request

calvernaz:
## Motivation
PR for issue: https://github.com/localstack/localstack/issues/10015
These are needed change to expose properly the source IP address to the context.

## Changes
Expose the source IP address and insert it into the auth context.


bentsku (approved):
LGTM! This should enable using the source IP down the chain. Thanks!

Issue #10022, opened 2024-01-08
enhancement request: Add DynamoDB TTL to evict stale KV pairs


robertlcx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Lots of teams rely on TTL on their DynamoDB to reduce their significant AWS cloud spend by having stale KV pairs evicted from the store. Naturally, most teams want this on Localstack as well to maintain parity with what they have on AWS.

More on this here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html

### Anything else?

This is a user-requested feature.

giograno:
Initial implementation provided in #10041

PR #10041, opened 2024-01-10
feat: DynamoDB TTL

giograno:
## Motivation
This PR implements the Time to Live feature in DynamoDB, as reported in #10022.

## Changes
- We added a `ExpiredItemsWorker` class in the DDB provider that schedules a thread removing the stale keys across all the tables that define an enabled TTL specification. It uses a `Scheduler` under the hood and runs every hour. We could make the period interval parameterized.
- We added the `DELETE /_aws/dynamodb/expired` endpoint that programmatically triggers the stale items' deletion.

### Misc
- to implement the deletion, we scan every table with the `Scan` operation, collecting the stale items.
- we use `BatchItemWrite` to delete the collected stale items from the DBs. The call should consume write throughput, while TTL does not.
- an alternative would be to mark the expired element, instead of deleting them, and exclude them from all the relevant operations.

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -4,12 +4,14 @@ import logging
 import os
 import random
 import re
+import threading
 import time
 import traceback
 from binascii import crc32
 from contextlib import contextmanager
+from datetime import datetime
 from operator import itemgetter
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 import requests
 import werkzeug
@@ -105,7 +107,7 @@ from localstack.constants import (
     AWS_REGION_US_EAST_1,
     TEST_AWS_SECRET_ACCESS_KEY,
 )
-from localstack.http import Response
+from localstack.http import Request, Response, route
 from localstack.services.dynamodb.models import DynamoDBStore, dynamodb_stores
 from localstack.services.dynamodb.server import DynamodbServer
 from localstack.services.dynamodb.utils import (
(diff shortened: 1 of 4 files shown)


localstack/services/dynamodb/provider.py
```diff
@@ -356,15 +359,123 @@ def modify_context_region(context: RequestContext, region: str):
...
+
+def delete_expired_items() -> int:
+    """
+    This utility function iterates over all stores, looks for tables with TTL enabled,
+    scan such tables and delete expired items.
+    """
+    no_expired_items = 0
+
+    for account_id, account_bundle in dynamodb_stores.items():
+        for region_name, state in account_bundle.items():
```
thrau, line 383:
```suggestion
    for account_id, region_name, state in dynamodb_stores.iter_stores():
```

localstack/services/dynamodb/provider.py
```diff
@@ -356,15 +359,123 @@ def modify_context_region(context: RequestContext, region: str):
...
+            ttl_specs = state.ttl_specifications
+            for table_name, ttl_spec in ttl_specs.items():
+                if ttl_spec.get("Enabled", False):
+                    attribute_name = ttl_spec.get("AttributeName")
+                    client = connect_to(region_name=region_name).dynamodb
```
thrau, line 388: can we move the client connection to the outer loop?

thrau (approved):
LGTM already! maybe we can hide this behind a feature flag for now, at least the automatic scheduler.

i'll also let @viren-nadkarni make the final approval!

viren-nadkarni (approved):
LGTM :+1:

localstack/services/dynamodb/provider.py
```diff
@@ -356,15 +359,120 @@ def modify_context_region(context: RequestContext, region: str):
...
+
+def delete_expired_items() -> int:
+    """
+    This utility function iterates over all stores, looks for tables with TTL enabled,
+    scan such tables and delete expired items.
+    """
+    no_expired_items = 0
+    for account_id, region_name, state in dynamodb_stores.iter_stores():
+        ttl_specs = state.ttl_specifications
+        client = connect_to(region_name=region_name).dynamodb
```
viren-nadkarni, line 383:
```suggestion
        client = connect_to(aws_access_key_id=account_id, region_name=region_name).dynamodb
```

Otherwise the expired items will never be cleared in tables belonging to non-default accounts.

viren-nadkarni (changes requested):

PR #10043, opened 2024-01-10
refactor Poller.handle_message and extract process_event_retry

thrau:
## Motivation

Tiny refactoring to make it easier to instrument the poller with logging. This is for a quick instrumentation exception we are pushing out to a user.

## Changes

* the block to retry a lambda without increasing the counter is now in its own method

### DIFF
--- a/localstack/services/lambda_/invocation/event_manager.py
+++ b/localstack/services/lambda_/invocation/event_manager.py
@@ -208,36 +208,7 @@ class Poller:
                     self.process_dead_letter_queue(sqs_invocation, invocation_result)
                     return
                 # 3) Otherwise, retry without increasing counter
-
-                # If the function doesn't have enough concurrency available to process all events, additional
-                # requests are throttled. For throttling errors (429) and system errors (500-series), Lambda returns
-                # the event to the queue and attempts to run the function again for up to 6 hours. The retry interval
-                # increases exponentially from 1 second after the first attempt to a maximum of 5 minutes. If the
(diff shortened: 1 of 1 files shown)


localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -299,6 +270,36 @@ def handle_message(self, message: dict) -> None:
                 "Error handling lambda invoke %s", e, exc_info=LOG.isEnabledFor(logging.DEBUG)
             )
 
+    def process_event_retry(self, sqs_invocation: SQSInvocation, error: Exception):
```
joe4dev, line 273: nit: The `event_retry` suffix could be a bit misleading because the extracted code only covers "Throttles (429) and system errors (5xx)". `process_throttles_and_system_errors` would be more accurate here, but longer.

"Function errors (returned from code or runtime, such as timeouts)" are the real "event retries" but are handled in the code below. Retry handling differs depending on the error: https://aws.amazon.com/blogs/compute/introducing-new-asynchronous-invocation-metrics-for-aws-lambda/

joe4dev (approved):
LGTM (pure refactoring) 👍 
nit: The naming could be a little more accurate

Are we going to have some implicit dependency on these method names in the future from an Extension?

thrau:
> Are we going to have some implicit dependency on these method names in the future from an Extension?

well for now yes, but this is obviously not a good solution, so having a better tracing interface will definitely be a good project moving forward!

PR #10045, opened 2024-01-10
Add new internalized sqs queue for lambda events

dfangl:
## Motivation
We currently have a lot of issues with an increasing load on the gateway, and sometimes timeouts, mostly due to the many requests by our lambda event queue.

This leads to timeouts in scenarios with many lambdas, many events, or overall high pressure on the gateway.

We should find a way to reduce the number of internal requests, by using the SqsQueue construct directly (and therefore using the features available) instead of invoking it through the sqs service.

## Changes
* Add new implementation of the sqs queueing with usage of the internal SqsQueue construct
* Add fake client to have a drop in replacement so switching is easily possible, and no code changes are necessary in the usage
* Add feature flag `LAMBDA_EVENTS_INTERNAL_SQS` to turn off the feature for now

## Performance Testing
@joe4dev 

### Test environment

Using the same setup as in previous testing in https://github.com/localstack/localstack/pull/9972

- LocalStack docker-mode with DEBUG logs using `python -m localstack.dev.run -e DEBUG=1`
- LocalStack version: `fc599c1f0` (use correct queue update worker) from this PR
- Running against the refined performance tests in https://github.com/localstack/localstack/pull/10032/commits/f6205d387f8889782dc8f7c940c06ec88c1f69bc
- MacBook Pro M1 10/16 (non-binned ARM) 16 GB RAM
- macOS 14.2 (23C64)

Disclaimer: We expect much better performance under Linux with native Docker rather running within the Docker Desktop VM on macOS and running Lambda using Rosetta translation.

## Results

`test_lambda_event_invoke`

- N=1500 ✅ 57s
- N=1800 ✅ 59s
- N=2000 ✅ 1min 6s
- N=2500 ✅ 1min 23s
- N=3000 ✅ 1min 43s
- N=4000 ✅ 2min 15s
- N=4094 ✅ 2min 12s
- N=4095 ⚠️ hitting a Python thread limit in the current test setup `RuntimeError: can't start new thread`

Without this PR, we see timeout errors upon higher asgi gateway load. The results match previous testing in https://github.com/localstack/localstack/pull/9972:

- N=1500 ✅ 1min 22s
- N=1600 ❌ with 1510 failed invokes and lots of timeout errors after 3min 15s
- N=1700 ❌ with 1616 failed invokes and lots of timeout errors after 7min 15s
- N=1800 ❌ with 1664 failed invokes and lots of timeout errors
- N=2000 ❌ with 1833 failed invokes and lots of timeout errors

`test_number_of_functions`

- N=100 ✅ 4min 7s
- N=120 ✅ 5min 17s
- N=150 ❌ 6min 33s with many `Connection refused` errors (likely due to gateway limit)
- N=200 ❌ 7min 28s

`test_number_of_function_versions_async`

- N=1000 ✅ 58s

## Conclusion

This PR improves the scalability of Lambda in LocalStack when using `LAMBDA_EVENTS_INTERNAL_SQS=1`:

- Event invokes are more resilient, faster, and less resource-intensive because we dispatch them to a new internal lightweight SQS implementation for Lambda instead of going through the full HTTP roundtrip via the LocalStack ASGI gateway. We tested up to 4096 concurrent asynchronous invocations ingested at the same time. Event invokes are used by the following integrations:
    - Direct invokes with `InvocationType="Event"`
    - SNS ⇒ Lambda
    - S3 ⇒ Lambda
    - API gateway (if async configured) ⇒ Lambda
    - EventBridge ⇒ Lambda
- LocalStack supports more simultaneously deployed Lambda functions (>100) and function versions (>1000) than before (numbers are initial tests).


localstack/config.py
```diff
@@ -931,6 +931,8 @@ def populate_edge_configuration(
     os.environ.get("LAMBDA_LIMITS_MAX_FUNCTION_ENVVAR_SIZE_BYTES", 4 * 1024)
 )
 
+LAMBDA_EVENTS_INTERNAL_SQS = is_env_not_false("LAMBDA_EVENTS_INTERNAL_SQS")
```
dfangl, line 934: This will obviously be changed to is_env_true before shipping

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,277 @@
...
+            self.queue_update_worker.remove_queue(queue)
+
+
+if config.LAMBDA_EVENTS_INTERNAL_SQS:
+    QUEUE_MANAGER = QueueManager()
+else:
+    QUEUE_MANAGER = None
+
+
+class FakeSqsClient:
```
thrau, line 166: can we remove this and instead inject the queue manager as dependency into the fake client?

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,277 @@
...
+        return {
+            "MessageId": message["MessageId"],
+            "MD5OfMessageBody": message["MD5OfBody"],
+            "MD5OfMessageAttributes": message.get("MD5OfMessageAttributes"),
+            "SequenceNumber": queue_item.sequence_number,
+            "MD5OfMessageSystemAttributes": _create_message_attribute_hash(MessageSystemAttributes),
+        }
+
+
+CLIENT = FakeSqsClient()
```
thrau, line 268: maybe we can use the factory pattern here instead to defer the creation
```suggestion
@singleton_factory
def get_fake_sqs_client():
    return FakeSqsClient(QueueManager())
```

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,277 @@
...
+
+
+def get_sqs_client(function_version: FunctionVersion, client_config=None):
+    if config.LAMBDA_EVENTS_INTERNAL_SQS:
+        return CLIENT
```
thrau, line 273:
```suggestion
        return get_fake_sqs_client()
```

localstack/services/sqs/models.py
```diff
@@ -635,6 +636,31 @@ def remove_permission(self, label: str) -> None:
         else:
             del self.attributes[QueueAttributeName.Policy]
 
+    def get_queue_attributes(self, attribute_names: AttributeNameList = None) -> dict[str, str]:
```
thrau, line 639: nice!

localstack/config.py
thrau, line 934: also add it to the config vars for passing it to the container.

thrau (approved):
nice! very neat implementation and can give us a quick win to improve lambda performance. perhaps over time this can become the default :+1:

localstack/services/lambda_/invocation/event_manager.py
dfangl: Yeah, that was a quick hack, good catch, would have missed it!

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,238 @@
...
+                account_id=config.INTERNAL_RESOURCE_ACCOUNT,
+            )
+            self.queues[queue_name] = queue
+            self.queue_update_worker.add_queue(queue)
+        return queue
+
+    def delete_queue(self, queue_name: str) -> None:
+        with self.queue_lock:
+            if queue_name not in self.queues:
+                raise ValueError("Queue not available")
```
dominikschubert, line 116: nit: could add the queue name in the message

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,238 @@
...
+
+            queue = self.queues.pop(queue_name)
+            self.queue_update_worker.remove_queue(queue)
+
+
+class FakeSqsClient:
```
dominikschubert, line 122: Really like the idea!

One thing Id' suggest is moving this and accompanying logic to a separate file though. The event manager ideally shouldn't care about this implementation and is already fairly complex, so we should maybe try to keep the file from bloating too much.

dominikschubert (approved):
LGTM, left a few minor remarks. 

Nice approach! I'm curious in how far the in-memory client would also help us here. Long-polling would probably still be an issue but otherwise it would probably behave fairly similar right? :thinking:

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,238 @@
...
+    def remove_queue(self, queue: SqsQueue):
+        self.queues.remove(queue)
+
+    def iter_queues(self) -> Iterable[SqsQueue]:
+        return iter(self.queues)
+
+
+class QueueManager:
+    queues: dict[str, StandardQueue]
+    queue_lock: threading.RLock
```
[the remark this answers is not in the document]
dfangl: Will do!

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,238 @@
...
+        """
+        Creates a queue.
+        :param queue_name: Queue name, has to be unique
+        :return: Queue Object
+        """
+        with self.queue_lock:
+            if queue_name in self.queues:
+                return self.queues[queue_name]
+
+            # Truncate function name to ensure queue name limit of max 80 characters
```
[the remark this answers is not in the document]
dfangl: Woops, leftover, good catch!

localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -7,34 +7,238 @@
...
+        return {"Attributes": result}
+
+    def purge_queue(self, QueueUrl: String) -> None:
+        queue = self.queue_manager.get_queue(queue_name=QueueUrl)
+        queue.clear()
+
+    def receive_message(
+        self,
+        QueueUrl: String,
+        attribute_names: AttributeNameList = None,
```
joe4dev, line 149: Is the mixed naming style of CamelCase and snake_case intentional here?

localstack/services/lambda_/invocation/event_manager.py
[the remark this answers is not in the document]
joe4dev: agree, that makes it easier to silence the exceptions upon LS shutdown:

```
During handling of the above exception, another exception occurred:

  File "/opt/code/localstack/localstack/services/lambda_/invocation/event_manager.py", line 419, in handle_message
    sqs_client.delete_message(
  File "/opt/code/localstack/localstack/services/lambda_/invocation/event_manager.py", line 173, in delete_message
    queue = self.queue_manager.get_queue(queue_name=QueueUrl)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/lambda_/invocation/event_manager.py", line 90, in get_queue
    raise ValueError("Queue not available")
ValueError: Queue not available
```

joe4dev (approved):
Amazing improvements @dfangl 🥇  💯  👏👏👏 
Maybe, you're going to forget to flip the `is_env_not_false` by accident ;)
Looking forward to having that as default in the future 🚀 

That's mostly an SQS PR and Thomas and Dominik have covered most points.

The numbers from performance testing look really promising 👍 
I'll add them to the PR description. The most relevant test `test_lambda_event_invoke` now succeeds with N=4094 ✅ under 2min 12s 🥳 (before hitting a Python thread limit in the test setup ;)

In comparison, without this fix, above N=1500 starts to cause timeout errors and everything is slower (CloudWatch v1 also causes quite some delay, which we don't use anymore with the internal SQS implementation).

localstack/services/lambda_/invocation/event_manager.py
dfangl: No, should all be camel case for API compatibility, good catch! Missed one

dfangl:
Regarding the in memory client: It will have similar improvements, but here we still have some advantages:
1. Complete independence of the running SQS service and state
2. No cloudwatch events published
3. No serialization/deserialization necessary, which further reduces latency

So all in all, I think it is for now beneficial to keep the solution, even if we get the in memory clients at some point :)

joe4dev:
> Regarding the in memory client: It will have similar improvements, but here we still have some advantages:
> 
>     1. Complete independence of the running SQS service and state
> 
>     2. No cloudwatch events published
> 
>     3. No serialization/deserialization necessary, which further reduces latency
> 
> 
> So all in all, I think it is for now beneficial to keep the solution, even if we get the in memory clients at some point :)

Thank you for adding this comment @dfangl . That is valuable information to remember.

Your point that the in-memory client is not yet ready was also an important factor. It looks like this is the latest PR https://github.com/localstack/localstack/pull/9569

(we have a total of 3 PRs WIP related to in-memory clients https://github.com/localstack/localstack/pulls?q=is%3Apr+is%3Aopen+in-memory+client)

PR #10046, opened 2024-01-10
improve + test thread safety for metric access

steffyP:
## Motivation
Some improvements for thread handling when accessing the sqlite database.

Some observations for sqlite database:
- the sqlite database is compiled once per system, so when you run in host-mode you may have different configuration then when running in docker
- observations on my personal computer: 
  - host mode runs a sqlite with different [threadsafety](https://docs.python.org/3.11/library/sqlite3.html#sqlite3.threadsafety): `THREADSAFE=2`, while in docker runs `THREADSAFE=1` (see also https://sqlite.org/threadsafe.html)
    - 2 means Multithreaded: Threads may share the module, but not connections 
    - 1 means Serialized: Threads may share the module, connections and cursors, should be save to access from multiple threads

General implementation:
* only `put_metric_data` requires write access
* all other operations only require read access, using `sqlite3.connect(file:db_name?mode=ro, uri=True)` 

Given this information my initial assumptions was:
* we only need a Lock for the `add_metrics` (which is called by `put_metric_data`), and when creating the database, or resetting the provider (which clears the table)

However, testing this with a huge amount of threads that run in parallel (see `test_parallel_put_metric_data_list_metrics`), we do need locks for every access (apparently the READ locks may lock the entire database under some circumstances (wasn't able to further pin this down). 

Additionally, when running in host-mode (assuming you only have `THREADSAFE=2`), each operation needs locks as well. 

Some run-time results from the test `test_parallel_put_metric_data_list_metrics` that I run on my computer:

started with `python -m localstack.dev.run` (starting as pro-image)
image-details
```
LocalStack version: 3.0.3.dev
LocalStack Docker container id: 676537bd7304
LocalStack build date: 2024-01-05
LocalStack build git hash: ba992b1
```

**Using v1 or running the tests with no-locks or only locks for `add_metrics` always failed in some iterations.**

I also ran the test `test_lambda_performance#test_lambda_event_invoke` several times successful with that configuration.

## Changes
* fix paging for `get-metrics`
* fix `delete-alarms` + added validated test
* refactor database-helper
  * make sure to create the connection after data has been prepared
  * add Locks for database access
* added `_STORE_LOCK` in the provider, for when we modify the alarm
  * reasoning: the `set-alarm-state` can be triggered any time by the alarm-scheduler, changing the alarm
  * this makes concurrent access way more likely
  * @baermat I saw in SQS a similar approach, maybe you can share some insights + thoughts on this 🙏  
* added a basic concurrency test
* added performance tests (which don't run with the pipeline)

## TODO
- [ ] add performance tests for sqs (will do this is another PR)
- [ ] optional follow-up: do benchmarks for using the `RWLockWrite` that @bentsku [suggested](https://github.com/localstack/localstack/commit/87ceac3595d17916b9edd9a9ae55380a0b3c0e76)
    -  Using the `python -m localstack.dev.run` I tested the approach for `test_parallel_put_metric_data_list_metrics`, and it works for the first run, starting from the second run (not restarting container in between), I get connection errors. 
    - Wild guess: Maybe the write-locks are preferred over reads, and the reads timeout?

### DIFF
--- a/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -44,10 +44,10 @@ class CloudwatchDatabase:
             return
 
         mkdir(self.CLOUDWATCH_DATA_ROOT)
-        with self.DATABASE_LOCK:
-            with sqlite3.connect(self.METRICS_DB) as conn:
-                cur = conn.cursor()
-                common_columns = """
+        # with self.DATABASE_LOCK:
+        with sqlite3.connect(self.METRICS_DB) as conn:
+            cur = conn.cursor()
+            common_columns = """
                     "id"	                INTEGER,
                     "account_id"	        TEXT,
                     "region"	            TEXT,
@@ -58,18 +58,18 @@ class CloudwatchDatabase:
                     "unit"	                TEXT,
                     "storage_resolution"	INTEGER
                 """
-                cur.execute(
-                    f"""
+            cur.execute(
+                f"""
                 CREATE TABLE "{self.TABLE_SINGLE_METRICS}" (
                     {common_columns},
                     "value"	                NUMERIC,
                     PRIMARY KEY("id")
                 );
                 """
-                )
+            )
 
-                cur.execute(
-                    f"""
+            cur.execute(
+                f"""
                 CREATE TABLE "{self.TABLE_AGGREGATED_METRICS}" (
                     {common_columns},
                     "sample_count"          NUMERIC,
@@ -79,15 +79,15 @@ class CloudwatchDatabase:
                     PRIMARY KEY("id")
                 );
                 """
-                )
-                # create indexes
-                cur.executescript(
-                    """
+            )
+            # create indexes
+            cur.executescript(
+                """
                 CREATE INDEX idx_single_metrics_comp ON SINGLE_METRICS (metric_name, namespace);
                 CREATE INDEX idx_aggregated_metrics_comp ON AGGREGATED_METRICS (metric_name, namespace);
                 """
-                )
-                conn.commit()
+            )
+            conn.commit()
 
     def add_metric_data(
         self, account_id: str, region: str, namespace: str, metric_data: MetricData
@@ -134,16 +134,16 @@ class CloudwatchDatabase:
 
                         cur.execute(
                             f"""INSERT INTO {self.TABLE_SINGLE_METRICS}
-                    ("account_id", "region", "metric_name", "namespace", "timestamp", "dimensions", "unit", "storage_resolution", "value")
-                    VALUES {prepared_placeholder}""",
+                        ("account_id", "region", "metric_name", "namespace", "timestamp", "dimensions", "unit", "storage_resolution", "value")
+                        VALUES {prepared_placeholder}""",
                             data,
                         )
 
                     if statistic_values := metric.get("StatisticValues"):
                         cur.execute(
                             f"""INSERT INTO {self.TABLE_AGGREGATED_METRICS}
-                        ("account_id", "region", "metric_name", "namespace", "timestamp", "dimensions", "unit", "storage_resolution", "sample_count", "sum", "min", "max")
-                        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""",
+                            ("account_id", "region", "metric_name", "namespace", "timestamp", "dimensions", "unit", "storage_resolution", "sample_count", "sum", "min", "max")
+                            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""",
                             (
                                 account_id,
                                 region,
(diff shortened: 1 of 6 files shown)


localstack/services/cloudwatch/cloudwatch_database_helper.py
bentsku, line 1: I think this is good! 👌 I believe the actual database operation is really fast compared to the rest of the handler chain, so even if we're locking for every read/write access, the actual locking time is pretty small compared to what we can serve. This should be more than enough for now! Sorry for bringing up the read/write lock, I now think it would be overkill for now.

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -183,12 +185,14 @@ def delete_alarms(self, context: RequestContext, alarm_names: AlarmNames) -> Non
         """
         Delete alarms.
         """
-
-        for alarm_name in alarm_names:
-            alarm_arn = arns.cloudwatch_alarm_arn(
-                alarm_name, account_id=context.account_id, region_name=context.region
-            )  # obtain alarm ARN from alarm name
-            self.alarm_scheduler.delete_scheduler_for_alarm(alarm_arn)
+        with _STORE_LOCK:
```
bentsku, line 188: q: can I ask what was the issue that necessitated to lock the whole store for the alarms? Is it the size changed during iteration error? or something else?

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -183,12 +185,14 @@ def delete_alarms(self, context: RequestContext, alarm_names: AlarmNames) -> Non
...
+            for alarm_name in alarm_names:
+                alarm_arn = arns.cloudwatch_alarm_arn(
+                    alarm_name, account_id=context.account_id, region_name=context.region
+                )  # obtain alarm ARN from alarm name
+                self.alarm_scheduler.delete_scheduler_for_alarm(alarm_arn)
+                store = self.get_store(context.account_id, context.region)
+                store.Alarms.pop(alarm_arn, None)
```
bentsku, line 195: nit: unrelated to this PR, but the `store` attributes `Alarms` with a capital A does not look too pythonic... would there be a way to change it before the provider goes live and that it would break persistence? I believe it should be `alarms`. Was there a specific reason for it being Pascal cased?

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -288,60 +292,60 @@ def set_alarm_state(
...
-
-        store = self.get_store(context.account_id, context.region)
-        alarm = store.Alarms.get(
-            arns.cloudwatch_alarm_arn(
-                alarm_name, account_id=context.account_id, region_name=context.region
+        with _STORE_LOCK:
+            store = self.get_store(context.account_id, context.region)
+            alarm = store.Alarms.get(
+                arns.cloudwatch_alarm_arn(
+                    alarm_name, account_id=context.account_id, region_name=context.region
```
bentsku, line 299: q: depending on the response to the reason of the locking, if we need to keep the lock, do we need to lock the whole iteration + actually calling the actions, or could we just lock the store access? It doesn't seem the store is accessed anymore after those lines, and I guess we wouldn't need to lock the whole store for it.

bentsku (commented):
This looks really good! Nice work 👌 I just have a question around the `_STORE_LOCK` and a small nit around the `Alarms` attribute of the store, but other than this, it looks really great!

localstack/services/cloudwatch/provider_v2.py
steffyP: basically the reason what I wrote here: 
> added _STORE_LOCK in the provider, for when we modify the alarm
    reasoning: the set-alarm-state can be triggered any time by the alarm-scheduler, changing the alarm
    this makes concurrent access way more likely
    @baermat I saw in SQS a similar approach, maybe you can share some insights + thoughts on this 🙏

so i didn't encounter any issues during running the tests, but I think it's fair to assume this could become an issue.

localstack/services/cloudwatch/provider_v2.py
steffyP: we do access the store in line 312 when calling 
```
self_update_state(...)
```
Also my intention to avoid a situation where we send wrong information about the `alarm` to sns/lambda. As we access the content of `alarm`, it could in theory be modified by another store-access before calling the actions, leading to inconsistent alarm-details. 
Maybe I am overcautions though - what are you thoughts on that?

localstack/services/cloudwatch/provider_v2.py
steffyP: https://github.com/localstack/localstack/pull/10087

baermat (commented):
Regarding the STORE_LOCK, you pretty much hit the mark. The SQS provider has a similar setup, a worker (two actually since the latest addition of move message tasks) that can manipulate a queue at any given moment in theory, so we need to guard ourselves against this possible conflict. Whether we have encountered any specific issues in the past is unknown to me, but the setup makes sense, and since the interval of these worker operations is within seconds, not milliseconds, the impact is probably acceptable.

localstack/services/cloudwatch/provider_v2.py
baermat: My take on this is that as long as we are not experiencing heavy performance degradation when locking down the store for access we should play it safe. Especially since the more fine-grained options proved more fickle than anticipated.

bentsku: Agreed, thanks for answering my questions! 😄

bentsku (approved):
Thanks for answering my questions! LGTM, thanks for tackling this!

localstack/services/cloudwatch/provider_v2.py
steffyP: as discussed with @bentsku offline: we need to see how things evolve. this approach of locking may slow things down in case we have a high number of alarm-destinations. In that case make a copy of the alarm object could be preferred. For now we leave it that way.

PR #10049, opened 2024-01-10
use request parameter resolvers

calvernaz:
## Motivation

Improves community PR by using the `RequestParametersResolver` that can handle, headers, path and query strings and also static values.

Fixes the previous test - static values are configured using single quotes, "'Fixed-Header-Value'", but transformed into normal header values when passed to the integration.


bentsku (approved):
LGTM! Awesome, thanks for tackling this. I've ran the -ext pipeline and all tests are green as well there, so this has no chance of breaking things there either. Nice to use the new resolver 👍

PR #10051, opened 2024-01-11
Make lambda sqs event source mapping interval configurable

dfangl:
## Motivation
Our current lambda sqs event source mapping implementation is fairly limited, but we should introduce an option for customers to change the interval, if they need better performance.

## Changes
* Add configuration variable `LAMBDA_SQS_EVENT_SOURCE_MAPPING_INTERVAL`


joe4dev (approved):
LGTM as a quick workaround 🩹 

nit: Do we intentionally omit the unit `SEC`?


I would consider this config experimental or temporary. We should discuss ways to manage the configuration lifecycle better.

dfangl:
Ah, no not intentionally, just forgot it. I would also prefer not to document it, as it should go away with the event source mapping rework, and we really do not want people in general messing with it.

PR #10052, opened 2024-01-11
patch HTTPServer fixture to spawn daemon thread

bentsku:
## Motivation
We've had issue with the pipeline timing out very frequently for the community tests against Pro.
It seems all test runs timing out had a leftover thread in common, spawned by the `httpserver` fixture. 
This thread was not spawned as `daemon`, meaning it would prevent the interpreter from shutting down.

## Changes
Temporary fix: patch the fixture to spawn the thread as daemon, so that even if it's still running, we can force its removal.

## Next up
- Investigate why this thread is still there, prevent it and remove the patch

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -26,18 +26,15 @@ pytest_plugins = [
 
 
 # FIXME: remove this, quick hack to prevent the HTTPServer fixture to spawn non-daemon threads
-def patch_http_server_fixture():
+def pytest_sessionstart(session):
     import threading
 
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
LGTM, seems to be a sensible solution for now to at least not block the CI action after test execution finished. We'll still need to get to the bottom of why this is leaking but It's not extremely  urgent, but let's open a follow-up PR that reverts this immediately after merge so we don't forget about it

PR #10053, opened 2024-01-11
refactor EventSourceAsfAdapter for easier instrumentation

thrau:
## Motivation

Similar refactoring as in #10043, making it easier to instrument the lifecycle of an event source mapping call.

## Changes

* sync and async invocations are now separated out into separate functions so we can instrument them

### DIFF
--- a/localstack/services/lambda_/event_source_listeners/adapters.py
+++ b/localstack/services/lambda_/event_source_listeners/adapters.py
@@ -1,6 +1,7 @@
 import abc
 import json
 import logging
+import threading
 from abc import ABC
 from functools import lru_cache
 from typing import Callable, Optional
@@ -19,7 +20,6 @@ from localstack.services.lambda_.invocation.models import lambda_stores
 from localstack.utils.aws.client_types import ServicePrincipal
 from localstack.utils.json import BytesEncoder
 from localstack.utils.strings import to_bytes, to_str
-from localstack.utils.threads import FuncThread
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/aws/services/lambda_/test_lambda_performance.py
```diff
@@ -352,7 +352,7 @@ def test_lambda_event_source_mapping_sqs(
 ):
     """Test SQS => Lambda event source mapping with concurrent event invokes and validate the number of invocations."""
     # TODO: define IAM permissions
-    num_invocations = 2000
+    num_invocations = 4
```
joe4dev, line 355: accidental commit?
The latest test are still in the this PR: https://github.com/localstack/localstack/pull/10032

localstack/services/lambda_/event_source_listeners/adapters.py
```diff
@@ -74,52 +74,79 @@ def __init__(self, lambda_service: LambdaService):
         self.lambda_service = lambda_service
 
     def invoke(self, function_arn, context, payload, invocation_type, callback=None):
+        request_id = gen_amzn_requestid()
+        self._invoke_async(request_id, function_arn, context, payload, invocation_type, callback)
+
+    def _invoke_async(
```
joe4dev, line 80: We should not count this as "async invoke" because that is purely our internal legacy construct of dispatching invokes. Conceptually, it is just a single synchronous invoke that happens here!

joe4dev (approved):
ok as pure refactoring 🙈 

(Looking forward to getting rid of this soonish 🧹 )

localstack/services/lambda_/event_source_listeners/adapters.py
thrau: understand, thanks!

PR #10058, opened 2024-01-12
Refactor: Kinesis: Clean-up test fixtures

maxhoheiser:
## Motivation
I was working on implementing new tests for kinesis and kinesis firehose and wanted to use the opportunity to clean up the existing tests. I believe it is much cleaner to separate test fixtures and actual tests since this generally keeps test files shorter and easier to read.

## Changes
Moved existing fixtures to `conftest.py` and added a new fixture for creating a kinesis stream consumer that handles cleanup after the fact.

### DIFF
--- a/tests/aws/services/kinesis/conftest.py
+++ b/tests/aws/services/kinesis/conftest.py
@@ -1,6 +1,33 @@
+import logging
+
 import pytest
 
 from localstack.utils.common import poll_condition
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def register_kinesis_consumer(aws_client):
+    kinesis = aws_client.kinesis
+    consumer_arns = []
+
+    def _register_kinesis_consumer(stream_arn: str, consumer_name: str):
+        response = kinesis.register_stream_consumer(
+            StreamARN=stream_arn, ConsumerName=consumer_name
+        )
+        consumer_arn = response["Consumer"]["ConsumerARN"]
+        consumer_arns.append(consumer_arn)
+
+        return consumer_arn
+
+    yield _register_kinesis_consumer
+
+    for consumer_arn in consumer_arns:
+        try:
+            kinesis.deregister_stream_consumer(ConsumerARN=consumer_arn)
+        except Exception:
+            LOG.info("Failed to deregister stream consumer %s", consumer_arn)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


tests/aws/services/kinesis/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    consumer_arns = []
+
+    def _register_kinesis_consumer(stream_arn: str, consumer_name: str):
+        response = kinesis.register_stream_consumer(
+            StreamARN=stream_arn, ConsumerName=consumer_name
+        )
+        consumer_arn = response["Consumer"]["ConsumerARN"]
+        consumer_arns.append(consumer_arn)
+
+        return consumer_arn
```
simonrw, line 22: Suggestion: it's usually more useful to return the response from the API call rather than a piece of data from within it. That way if there are other properties that are needed, they are available. See for example https://github.com/localstack/localstack/blob/0b060540fcd5a9826fcdf7a17aace861f6ea76d9/localstack/testing/pytest/fixtures.py#L372-L389 vs https://github.com/localstack/localstack/blob/0b060540fcd5a9826fcdf7a17aace861f6ea76d9/localstack/testing/pytest/fixtures.py#L614-L631

tests/aws/services/kinesis/conftest.py
simonrw, line 1: Minor: it might be worth moving `register_kinesis_consumer` and `wait_for_consumer_ready` to `localstack/testing/pytest/fixtures.py` as this makes them available for any test in both community and -ext. If you do this, I would suggest renaming `wait_for_consumer_ready` to `wait_for_kinesis_consumer_ready` or something to indicate that it is a kinesis fixture.

simonrw (approved):
Generally looks good, just a couple of suggestions. I am excited to see `kinesis` getting some attention. Hopefully this fixture centralising means more tests to come!

tests/aws/services/kinesis/conftest.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/commit/7580ee14132c9696bf976feace0df8bc60523fcb

tests/aws/services/kinesis/conftest.py
maxhoheiser: I think for now I will not move it to global fixtures, since it is much more readable in a separate config file directly at the service tests, but I aligned the naming of the fixtures to be immediately identifiable with the service for global use in: https://github.com/localstack/localstack/commit/347e40d0ad4c35972e194a8e0295af005dec5000

PR #10061, opened 2024-01-12
move http framework code to rolo and only keep references

thrau:
## Motivation

For a while now we have discussed moving certain stable parts of LocalStack into standalone projects. The HTTP framework is a good candidate as it is very cohesive and re-usable for future projects.

We've released https://github.com/localstack/rolo that now contains most of the HTTP framework. The next steps will be extracting the handler chain and gateway into rolo. 
At some point we can also remove the framework tests and migrate to the rolo test utilities.

## Changes

* Most HTTP framework code now lives in https://github.com/localstack/rolo
* For backwards compatibility, we maintain imports of rolo to the old imports of localstack (`localstack.http`), this way, no dependent code needs to change for now


alexrashed (approved):
Awesome! 🚀 
This is a really nice and clean extraction of a stable and concise part of our core codebase! It basically "graduates" and becomes its own library.
The "deprecation path" is pretty clean, I couldn't find any previous declarations which are not exported using `__all__`. 💯 

I think moving forward it would be interesting to discuss how we want to migrate the imports of `localstack.http` to `rolo`. Should we:
- Keep the imports of `localstack.http`, basically hiding `rolo` between an internally maintained API surface. This would potentially allow us to extend certain things defined in rolo without any changes to the codebase.
  - For example, we extend the `RoloResponse` (because we define a custom encoder), but we don't extend `RoloRequest` yet. If we change the existing imports to directly import from `rolo`, it would be a bit more effort to introduce this abstraction layer again.
- Directly import from `rolo` instead of maintaining the internal API layer.

Another question is how we want to proceed with the tests. Currently, a lot of the unit tests are duplicated (they are in `rolo`, but also in this repo). In the next iteration, we could maybe cleanup the tests, such that only unit tests which cover localstack specifics remain in this repository.

All in all: Exciting, nice, and clean changeset! 🥇 🚀

PR #10062, opened 2024-01-12
Make the internal sqs implementation of lambda events the default

dfangl:
## Motivation
With #10045 , we introduced a new internalized mechanism based on sqs for lambda event queues.

This leads to a huge performance bonus, as well as less read timeouts in the logs.

We should make it default with the next minor release. We are still keeping in the option to switch back for now, so I would argue this as a minor change (no external API changes, but some behavioral corrections in terms of e.g. cloudwatch metrics shown for internal queues).

## Changes
* Set `LAMBDA_EVENTS_INERNAL_SQS` to true by default.


thrau (approved):
i think this is a good change! just making sure we're aligned with @joe4dev and @dominikschubert

joe4dev (approved):
This makes a lot of sense 👍 

The numbers in https://github.com/localstack/localstack/pull/10045 speak for themselves 🏎️ 🚀 

+1 definitely "minor"

Issue #9912, opened 2023-12-19
bug(CloudFormation): ObjectLockConfiguration is not created 


nikpivkin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The bucket is created without object lock configuration

### Expected Behavior

The bucket is created with object lock configuration

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

Deploy stack:
```bash
awslocal cloudformation deploy --stack-name test --template-file test2.yaml

Waiting for changeset to be created..
Waiting for stack create/update to complete
Successfully created/updated stack - test

awslocal s3api list-buckets --no-cli-pager
{
    "Buckets": [
        {
            "Name": "test-bucket",
            "CreationDate": "2023-12-19T10:03:10+00:00"
        }
    ],
    "Owner": {
        "DisplayName": "webfile",
        "ID": "75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a"
    }
}
```

Check the Object Lock configuration:
```bash
awslocal s3api get-object-lock-configuration --bucket test-bucket

An error occurred (ObjectLockConfigurationNotFoundError) when calling the GetObjectLockConfiguration operation: Object Lock configuration does not exist for this bucket
```

Logs:
```bash
2023-12-19 16:43:20 
2023-12-19 16:43:20 LocalStack version: 3.0.3.dev
2023-12-19 16:43:20 LocalStack Docker container id: 20c4fda0118d
2023-12-19 16:43:20 LocalStack build date: 2023-12-05
2023-12-19 16:43:20 LocalStack build git hash: c1dcbc50
2023-12-19 16:43:20 
2023-12-19 16:43:21 2023-12-19T09:43:21.334  INFO --- [-functhread4] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
2023-12-19 16:43:21 2023-12-19T09:43:21.334  INFO --- [-functhread4] hypercorn.error            : Running on https://0.0.0.0:4566 (CTRL + C to quit)
2023-12-19 16:43:21 Ready.
2023-12-19 17:02:52 2023-12-19T10:02:52.302  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS s3.ListBuckets => 200
[...]

PR #10070, opened 2024-01-15
add support for ObjectLock for CFN S3 Bucket

bentsku:
## Motivation
As reported by #9912, we did not support enabling ObjectLock for S3 Bucket, nor setting the object lock configuration to the bucket.

- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-objectlockconfiguration.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-s3-bucket.html#cfn-s3-bucket-objectlockenabled
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-s3-bucket.html#cfn-s3-bucket-objectlockconfiguration

## Changes
- update the CloudFormation model to support it
- created a new test checking the functionality and the split between enabling the bucket to be locked and setting up the configuration
- renamed some variables in the regular S3 tests because PyCharm now checks for fixture usage, and the name clashed (it was saying the fixture wasn't imported when we did not need it)

_fixes #9912_

### DIFF
--- a/localstack/services/s3/resource_providers/aws_s3_bucket.py
+++ b/localstack/services/s3/resource_providers/aws_s3_bucket.py
@@ -461,6 +461,12 @@ class S3BucketProvider(ResourceProvider[S3BucketProperties]):
                 CORSConfiguration=cors_configuration,
             )
 
+        if object_lock_configuration := model.get("ObjectLockConfiguration"):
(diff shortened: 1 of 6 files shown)


pinzon (approved):
LGTM

PR #10071, opened 2024-01-15
supports enabling S3 Object Lock on existing buckets

bentsku:
## Motivation
AWS released this statement regarding enabling S3 Object Lock on existing buckets. It wasn't possible before, and the bucket needed to be created with a specific configuration flag. 

See https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-s3-enabling-object-lock-buckets/

## Changes
Update the test and the logic to now allow enabling S3 object lock on existing buckets, if they are versioned.


steffyP (approved):
Nice addition 🚀 
The changes LGTM 👍 

Thanks for adding the aws validated tests 😄 
It's interesting to see how AWS changes the messages for invalid configuration so quickly.

PR #10087, opened 2024-01-19
cleanup store + remove moto fallback

steffyP:
## Motivation
Removing moto fallback for cloudwatch v2 provider, and cleaning up some variables in `models.py`

PR #10089, opened 2024-01-19
Add describe_alarms_for_metric and describe_alarm_history

steffyP:
## Motivation
When checking the tests, I realized there is still a flaky, and currently skipped test `test_put_metric_alarm`. 
It turns out that the current test setup was the reason for the flakiness, so I refactored it.
Also two operations where still missing from the provider v2 that are used in this test.

## Changes
- added missing implementation for `describe_alarms_for_metric` and `describe_alarm_history`
- parity fixes, test refactoring

For the test `test_put_metric_alarm`:
- initially, it would treat missing data as not breaching. depending on when the alarm is evaluated the first state would either change to OK or to ALARM (because the metrics we insert after breaches the condition)
- changed the alarm evaluation to "ignore", and removed the second part, where we expected the alarm to go back to state OK

### DIFF
--- a/localstack/services/cloudwatch/alarm_scheduler.py
+++ b/localstack/services/cloudwatch/alarm_scheduler.py
@@ -47,7 +47,7 @@ class AlarmScheduler:
         Shutsdown the scheduler, must be called before application stops
         """
         self.scheduler.close()
-        self.thread.join(5)
+        self.thread.join(10)
 
     def schedule_metric_alarm(self, alarm_arn: str) -> None:
         """(Re-)schedules the alarm, if the alarm is re-scheduled, the running alarm scheduler will be cancelled before
--- a/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack/services/cloudwatch/provider_v2.py
@@ -21,9 +21,12 @@ from localstack.aws.api.cloudwatch import (
     DashboardNames,
     Datapoint,
     DeleteDashboardsOutput,
+    DescribeAlarmHistoryOutput,
+    DescribeAlarmsForMetricOutput,
     DescribeAlarmsOutput,
     DimensionFilters,
     Dimensions,
+    ExtendedStatistic,
     ExtendedStatistics,
     GetDashboardOutput,
     GetMetricDataMaxDatapoints,
@@ -58,6 +61,7 @@ from localstack.aws.api.cloudwatch import (
     StateReason,
     StateReasonData,
     StateValue,
+    Statistic,
     Statistics,
     TagKeyList,
     TagList,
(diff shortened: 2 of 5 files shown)


localstack/services/cloudwatch/provider_v2.py
```diff
@@ -648,25 +697,39 @@ def _update_state(
...
+        state_reason_data: dict = None,
     ):
         old_state = alarm.alarm["StateValue"]
+        old_state_reason = alarm.alarm["StateReason"]
         store = self.get_store(context.account_id, context.region)
         current_time = datetime.datetime.now()
+        if state_reason_data:
+            state_reason_data["version"] = "1.0"
+        history_data = {
+            "version": "1.0",
```
baermat, line 709: small nit: even though it's hardcoded, maybe add `version = 1.0` before so once we actually start properly using these versions we are already on the right track

localstack/services/cloudwatch/alarm_scheduler.py
```diff
@@ -47,7 +47,7 @@ def shutdown_scheduler(self) -> None:
         Shutsdown the scheduler, must be called before application stops
         """
         self.scheduler.close()
-        self.thread.join(5)
+        self.thread.join(10)
```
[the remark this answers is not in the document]
steffyP: When I ran the tests locally, I sometimes ran into issues, where some threads did not properly shutdown. [`thread.join(x)`](https://docs.python.org/2/library/threading.html#threading.Thread.join) should wait up to x seconds to wait until the thread is closed. It will not always wait for 10 sec. I did not experience issues with the local execution when increasing the value. But I am not sure if this is just an issue when running tests in host mode.

localstack/services/cloudwatch/provider_v2.py
```diff
@@ -457,6 +476,36 @@ def describe_alarms(
...
+        extended_statistic: ExtendedStatistic = None,
+        dimensions: Dimensions = None,
+        period: Period = None,
+        unit: StandardUnit = None,
+    ) -> DescribeAlarmsForMetricOutput:
+        store = self.get_store(context.account_id, context.region)
+        alarms = [
+            a.alarm
+            for a in store.Alarms.values()
+            if isinstance(a, LocalStackMetricAlarm)
```
bentsku, line 494: q: can it be anything else than `LocalStackMetricAlarm`? I think not yet, is it for the future?

bentsku (approved):
LGTM! Nice catch, and nice seeing more operations being added! 
I was looking at the filtering, done with multiple list comprehensions, this is nicely readable. Nice work 😄

localstack/services/cloudwatch/provider_v2.py
steffyP: yes, you are right, currently we only have `LocalStackMetricAlarm` in use. 
We prepared `LocalStackCompositeAlarm` as well in the model, put currently we don't use it. 
I would rather leave this here though, because we will likely forget about this, once we add the composite alarms 😅

bentsku: Of course yes! I wasn't asking to remove it 😄

pinzon (approved):
Looks good. Just worth mentioning that you could probably encounter some conflicts when mergin due to the renaming of the stores attributes in this [PR](https://github.com/localstack/localstack/pull/10087)

Issue #9986, opened 2024-01-03
bug: Trying to rotation secret in SECRETS MANAGER


lucasmmo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I was trying to rotate the secret within the SecretsManager

- I'm using the lambda function [aws template](https://github.com/aws-samples/aws-secrets-manager-rotation-lambdas/tree/master/SecretsManagerRotationTemplate/lambda_function.py)
- I gave the lambda permission to access the secret manager using add-permissions

input cli:

```sh
awslocal secretsmanager rotate-secret \
    --secret-id test-secret \
    --rotation-lambda-arn arn:aws:lambda:us-east-1:000000000000:function:my-rotation-function \
    --rotation-rules "{\"ScheduleExpression\": \"cron(0 16 1,15 * ? *)\", \"Duration\": \"2h\"}
```

output cli:
```sh
An error occurred (InternalError) when calling the RotateSecret operation (reached max retries: 4): exception while calling secretsmanager.RotateSecret: cannot access local variable 'new_version_id' where it is not associated with a value
```

i did anything wrong?

I saw a bug that was merged into master [9862](https://github.com/localstack/localstack/pull/9862) that seems resolved...

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

this is my docker-compose

```docker-compose.yaml
version: "3.8"

services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack:latest
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    environment:
      - DEBUG=${DEBUG:-0}
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
 ```

create lambda function [template](https://github.com/aws-samples/aws-secrets-manager-rotation-lambdas/tree/master/SecretsManagerRotationTemplate/lambda_function.py)

```sh 
$ awslocal lambda create-function \      
    --function-name my-rotation-function \
    --runtime python3.9 \                                                                      
    --zip-file fileb://my-function.zip \                                                        
    --handler lambda_handler \              
    --role arn:aws:iam::000000000000:role/service-role/rotation-lambda-role
``` 

adding permission 

```sh
[...]

PR #10091, opened 2024-01-21
Fix secretsmanager secret rotation failing because RotateImmediately is unset

cowlingj:
## Motivation

Issue: https://github.com/localstack/localstack/issues/9986

Secret rotation currently doesn't work when the `RotateImmediately` parameter is unset. It looks like this is because `backend_rotate_secret` receives the argument `rotate_immediately` with the value `None` when the parameter `RotateImmediately` is unset (if the argument is unset it defaults to `True`, but in this case it's being explicitly set to `None`), this causes the `new_version_id` to not be set resulting in the error described in the issue `cannot access local variable 'new_version_id' where it is not associated with a value`.

## Changes

Check if rotate_immediately is `None` and set it to `True` in that case

## Testing

I'm unable to run any of the tests locally, so that'll be worth paying extra attention to when reviewing.

### DIFF
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -337,10 +337,11 @@ class TestSecretsManager:
             SecretId=secret_name, ForceDeleteWithoutRecovery=True
         )
 
+    @pytest.mark.parametrize("rotate_immediately", [True, None])
     @markers.snapshot.skip_snapshot_verify(paths=["$..Versions..KmsKeyIds"])
     @markers.aws.validated
     def test_rotate_secret_with_lambda_success(
-        self, sm_snapshot, secret_name, create_secret, create_lambda_function, aws_client
+        self, sm_snapshot, secret_name, create_secret, create_lambda_function, aws_client, rotate_immediately
     ):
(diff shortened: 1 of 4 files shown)


cowlingj:

dominikschubert:
Rebased on master & generated updated snapshots for the parametrized test.

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -367,13 +378,17 @@ def test_rotate_secret_with_lambda_success(
...
+            rotation_kwargs["RotateImmediately"] = rotate_immediately
         rot_res = aws_client.secretsmanager.rotate_secret(
             SecretId=secret_name,
             RotationLambdaARN=function_arn,
             RotationRules={
                 "AutomaticallyAfterDays": 1,
             },
-            RotateImmediately=True,
+            **rotation_kwargs
+            # RotateImmediately=False,
```
dominikschubert, line 391:
```suggestion
```
small leftover from my test edit

dominikschubert (approved):
LGTM, thanks for fixing the default behavior for the RotateImmediately field!

There's a few failing tests right now but those are caused by  an opensearch flake.

PR #10092, opened 2024-01-21
move gateway and handler chain framework to rolo

thrau:


tests/unit/services/s3/test_virtual_host.py
```diff
@@ -21,6 +21,9 @@ def request(self, request: Request, server: str | None = None) -> Response:
         self.requests.put((request, server))
         return Response()
 
+    def close(self):
+        pass
+
```
bentsku, line 26: I believe this has been superseded by #10099, but sadly I didn't put it at the same spot 😬 we might need to rebase before merging. Sorry!

bentsku (approved):
LGTM! Nice refactoring, it's nice to see this part of the codebase being pulled out as it became quite solid and a good abstraction to work with. 

I think we will need to remove [c79fcd7](https://github.com/localstack/localstack/pull/10092/commits/c79fcd7f1b7008e3c9277372862932adf4ccfdf0) to avoid conflict with #10999 but other than this, this is great!

Love the fact that the chain and gateway are now configurable. This is gonna be of a great help soon!

PR #10099, opened 2024-01-22
fix S3 v-host unit tests after rolo migration

bentsku:
## Motivation
We were missing an abstract class method definition following the rolo migration

## Changes
Added the `close` method to the test class for the S3 vhost tests.


thrau (approved):
thanks for fixing this!

i had to add the `@abstractmethod` decorator in rolo to shut up ruff, which then caused the client implementations to fail as i released 0.2.0. should have maybe added the same ruff exceptions that we have in localstack.

bentsku:
Ah, thanks for the explanation, I see 😄 it's gonna be good to also maybe make the ruff rules a bit stricter in LocalStack, I believe this is planned!

Issue #9939, opened 2023-12-25
bug: cdk to create ec2 instance failed with MaxCount/MinCount required


leecarney:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using CDK to create an ec2 instance fails with:
```
6:48:30 PM | CREATE_FAILED        | AWS::EC2::Instance                    | Ins
tanceLocalstack99C06A68
Parameter validation failed:
Missing required parameter in input: "MaxCount"
Missing required parameter in input: "MinCount"
```

These appear to be parameters for Instances inside an ASG rather than single ec2 instance.

### Expected Behavior

The ec2 instance should be created without needing to add min and max values as the ec2 instance module doesn't support them

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose using localstack recommended config and pro licence

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```
    import * as cdk from 'aws-cdk-lib'
import {
  Instance,
  Vpc,
  MachineImage,
  InstanceType,
  CfnKeyPair,
} from 'aws-cdk-lib/aws-ec2'
import { Construct } from 'constructs'

export class LocalstackCdkStack extends cdk.Stack {
  constructor(scope: Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props)

    const vpc = new Vpc(this, 'VpcLocalstack', {
      maxAzs: 1
    })

    const keyPair = new CfnKeyPair(this, 'KeyPairLocalstack', {
      keyName: 'localstack-ec2-key'
    })

    const instance = new Instance(this, 'InstanceLocalstack', {
      vpc,
      instanceType: new InstanceType('t4g.small'),
      machineImage: MachineImage.genericLinux({
        'us-west-2': 'ami-ff0fea8310f3'
      }),
      keyName: keyPair.keyName
      // TODO: new keypair format


    })
  }
}
```
```
cdklocal bootstrap
cdklocal deploy
```
### Environment

```markdown
- OS: Mac Ventura 13.6.2
- LocalStack: pro  2.3.3.dev20231111104446
```

MarcelStranak:
Hi @leecarney,

Could you please confirm if this issue has been resolved in the `latest` release of the LocalStack Docker image? If that is the case, kindly mark the matter as resolved. Thank you.

leecarney:
[...]

PR #10100, opened 2024-01-22
Add AWS::EC2::KeyPair

pinzon:
## Motivation

During the investigation of issue #9939, it became evident that the AWS::EC2::KeyPair resource provider was missing from the implementation. Consequently, this PR addresses that gap by introducing the necessary implementation.

## Changes

- Introduces a new resource provider for AWS::EC2::KeyPair.
- Adds default values for MaxCount and MinCount to enhance the EC2 instance resource provider.

Testing
- added a simple deployment test


localstack/services/ec2/resource_providers/aws_ec2_keypair.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Read-only properties:
+          - /properties/KeyPairId
+          - /properties/KeyFingerprint
+
+        IAM permissions required:
+          - ec2:CreateKeyPair
+          - ec2:ImportKeyPair
+          - ec2:CreateTags
+          - ssm:PutParameter
```
dominikschubert, line 66: kinda curious about the put parameter here.

From the AWS docs:

> When you create a new key pair, the private key is saved to AWS Systems Manager Parameter Store, using a parameter with the following name: /ec2/keypair/{key_pair_id}. For more information about retrieving private key, and the required permissions, see [Create a key pair using AWS CloudFormation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/create-key-pairs.html#create-key-pair-cloudformation) in the Amazon EC2 User Guide.

Not super urgent but seems this would be missing in terms of parity with the resource on AWS. Basically we need to create a parameter here and delete it again in the delete call.

tests/aws/templates/ec2_instance_keypair.yml
```diff
@@ -0,0 +1,17 @@
+Resources:
+  NewKeyPair:
+    Type: 'AWS::EC2::KeyPair'
+    Properties:
+      KeyName: !Sub 'keypair-${AWS::StackName}'
+  Ec2Instance:
+    Type: 'AWS::EC2::Instance'
+    Properties:
+      ImageId: "ami-0e9107ed11be76fde"
+      KeyName: !Ref NewKeyPair
```
dominikschubert, line 10: Any particular reason why the instance is included in the template here?

localstack/services/ec2/resource_providers/aws_ec2_keypair.py
pinzon: Thanks I just added that behavior to resource provider

tests/aws/templates/ec2_instance_keypair.yml
pinzon: I wrongly thought there were no deployment test for ec2 instances. so I rewrote the test a little just to assert the deployment and deletion of the keypair

dominikschubert (approved):
LGTM, thanks for also jumping on the identified SSM parameter parity gap and fixing it!

PR #10108, opened 2024-01-23
correctly execute host and container configuration hooks in dev.run

thrau:
## Motivation

In #9339, we added a way to explicitly load the extension dev mode configuration hooks. This was highly specific to this one use case, and a better way would be to run the hooks that are also run during bootstrapping when running `localstack start`.

This PR adds both the `prepare_host` and `configure_container` hooks to the dev script, to more closely match what `localstack start` is doing, which also includes extension dev mode when available.

The PR also passes through the config environment variables to the container as `localstack start` does

## Changes

* `prepare_host` and `configure_container` hooks are now run when running `localstack.dev.run`
* as a consequence, a license check is performed on the host before starting the container (it's in a `prepare_host` hook in ext)
  * this would require us to set `LOCALSTACK_API_KEY=test python -m localstack.dev.run -e LOCALSTACK_API_KEY`, which is a bit annoying as it breaks existing scripts, so i extracted the licensing credentials from the CLI arguments, so you can continue to just use `python -m localstack.dev.run -e LOCALSTACK_API_KEY=test`.
  * at the same time, i added the `config_env_vars` configurator, so you can also just `LOCALSTACK_API_KEY=test python -m localstack.dev.run`

## Testing

* check out the branch and activate the venv in localstack
* navigate to localstack-ext and try out `EXTENSION_DEV_MODE=1 LOCALSTACK_API_KEY=test python -m localstack.dev.run`

### DIFF
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -268,6 +268,7 @@ def run(
         ConfigEnvironmentConfigurator(pro),
         ContainerConfigurators.mount_localstack_volume(host_paths.volume_dir),
         CoverageRunScriptConfigurator(host_paths=host_paths),
+        ContainerConfigurators.config_env_vars,
     ]
 
     # create stub container with configuration to apply
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -1,12 +1,13 @@
 import dataclasses
 import os
-from typing import Tuple
+from typing import Iterable, Tuple
 
 import click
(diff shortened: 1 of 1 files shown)


simonrw (approved):
Thanks for generalising this approach to do the more correct thing. The closer we make the startup procedure to that of the CLI the better.

With the `CONFIG_PROFILE` system, it's even easier to get a dev workflow with this, setting `LOCALSTACK_AUTH_TOKEN=test` in the config profile and selecting it at runtime.

PR #10111, opened 2024-01-24
Dev mode: return correctly from raw mode

simonrw:
## Motivation

When using `localstack.dev.run`, the terminal would not return from raw mode after the container completed its execution. This would leave the terminal in a raw state where the enter key needed to be pressed multiple times to return to the shell.

## Changes

Use `console.status` as a context manager to ensure proper shutdown of `rich`'s terminal handling, returning to ["cooked mode"](https://en.wikipedia.org/wiki/Terminal_mode) on completion.

*Note: the diff is confusing, but I changed the call site of `console.status` to be a context manager and indented the rest of the `run` function until the `status.stop` was called.*

### DIFF
--- a/localstack/dev/run/__main__.py
+++ b/localstack/dev/run/__main__.py
@@ -212,113 +212,108 @@ def run(
         │   └── ...
 
     """
-    status = console.status("Configuring")
-    status.start()
+    with console.status("Configuring") as status:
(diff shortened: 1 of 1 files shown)


thrau (approved):
brilliant! nice catch :100:

PR #10112, opened 2024-01-24
Bump moto-ext to 5.0.0.post1

alexrashed:
## Motivation

[getmoto/moto](https://github.com/getmoto/moto) just had a new major release.

All changes are released with [moto-ext 5.0.0.post1](https://pypi.org/project/moto-ext/5.0.0.post1/)

This bump brings in the following fixes to LocalStack:
- https://github.com/getmoto/moto/pull/7204
- https://github.com/getmoto/moto/pull/7240 by @bentsku 
- https://github.com/getmoto/moto/pull/7251 by @bentsku 

## Changes

- Upgrades moto-ext to v5

## Related

Closes https://github.com/localstack/localstack/issues/8642

Issue #10009, opened 2024-01-06
bug: Creating SNS subscription returns 500 on empty FilterPolicy


Moulick:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When creating a SNS Subscription and setting the attribute `FilterPolicy` to `""` causes an 500 internal server error

### Expected Behavior

SNS Subscription is created without a `FilterPolicy`. This is the behaviour in AWS and is the way to unset the `FilterPolicy`

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

`docker run --rm -p 4566:4566 -p 4510-4559:4510-4559 --env DEBUG=1 --env LS_LOG=trace --name localstack localstack/localstack:latest`

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
    
```bash
export AWS_ENDPOINT_URL=http://localhost:4566 AWS_ACCESS_KEY_ID=fake AWS_SECRET_ACCESS_KEY=fake
aws sqs create-queue --queue-name queue.fifo --attributes FifoQueue=true
aws sns create-topic --name topic.fifo --attributes FifoTopic=true
aws sns subscribe --topic-arn arn:aws:sns:eu-central-1:000000000000:topic.fifo --protocol "sqs" --notification-endpoint arn:aws:sqs:eu-central-1:000000000000:queue.fifo --attributes FilterPolicy=""
```

### Environment

```markdown
- OS: MacOS Sonoma
- LocalStack version: 3.0.3.dev
```

### Anything else?

Setting `FilterPolicy` to `""` is the way to remove it from AWS

Logs from the container:

```python
[...]

PR #10125, opened 2024-01-26
add tests for fix to SNS empty filter policy

bentsku:
## Motivation
As reported with #10009, we've had an issue when setting empty filter policy. This fix is based on https://github.com/getmoto/moto/pull/7251, now part of LocalStack with #10112, and has additional logic in place where we manually work with the filter policy.

## Changes
Update existing tests to validate the fix, and update the logic in `Subscribe` too. 

_fixes #10009_


thrau (approved):
LGTM!

Moulick:
Thanks @bentsku

PR #10127, opened 2024-01-26
Introduce dependency pinning

silv-io:
## Motivation

We continuously run into issues with the pipeline that are caused by an updated dependency. Dependencies currently can just update on the fly in an uncontrolled manner and we don't know when a dependency receives an upgrade. 
This can obscure other issues with the pipeline or otherwise reduce the development velocity.

## Changes

This PR and its companion in the Ext repository introduce dependency pinning in LocalStack Ext and Core. 

To this end, new make targets are added which allow for the creation of lock files with the naming scheme `requirements-<extra>.txt`. These make targets can either be executed in CI or locally by devs:

- In the case of CI, this PR introduces a new workflow `upgrade-python-dependencies` which periodically (Mondays at 5AM - same as the ASF updates) and in a controlled manner update all the dependencies in the lock file to latest allowable versions defined in setup.cfg

- In the case of devs, this PR introduces a new pre-commit hook, referenced from the https://github.com/localstack/pre-commit-hooks repository, which checks whether a change in `setup.cfg` would make the lock files incompatible with the `setup.cfg`. If that is the case, the hook would fail and the dev will be prompted to run the make target for upgrading all dependency lock files. As an extra precaution, the run of this pre-commit hook was additionally added to the CI pipeline in the linting step.

## Testing

To test out how the pinning works you can execute the `upgrade-pinned-dependencies` make target.
You can add any package you wish to the `setup.cfg` and stage it and try to commit it to check if the pre-commit hook works.
You can also manually run the pre-commit hook with the staged changes in the `setup.cfg`:

```bash
pre-commit run check-pinned-deps-for-needed-upgrade --files $(git diff --name-only --cached)
```

## TODO

What's left to do:

- [x] Make sure all CI pipelines are green in the PR
- [x] Make sure the release process is solid
- [x] Decide on what to do with the eager updating in the Docker build
- [x] Run the dependency upgrade before merge to get off to a good start ;)


Makefile
```diff
@@ -21,7 +21,7 @@ endif
 VENV_RUN = . $(VENV_ACTIVATE)
 
 usage:                    ## Show this help
-	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/:.*##\s*/##/g' | awk -F'##' '{ printf "%-25s %s\n", $$1, $$2 }'
+	@grep -Fh "##" $(MAKEFILE_LIST) | grep -Fv fgrep | sed -e 's/:.*##\s*/##/g' | awk -F'##' '{ printf "%-25s %s\n", $$1, $$2 }'
```
alexrashed, line 24: nit: What's this change about?

Makefile
```diff
@@ -33,26 +33,52 @@ venv: $(VENV_ACTIVATE)    ## Create a new (empty) virtual environment
...
+	$(VENV_RUN); pip-compile --upgrade --extra test -o requirements-test.txt pyproject.toml
+
+upgrade-dev-reqs: pip-tools
+	$(VENV_RUN); pip-compile --upgrade --extra dev,test,runtime -o requirements-dev.txt pyproject.toml
+
+upgrade-dev-types-reqs: pip-tools
+	$(VENV_RUN); pip-compile --upgrade --extra typehint -o requirements-typehint.txt pyproject.toml
+
+upgrade-s3-reqs: pip-tools
+	$(VENV_RUN); pip-compile --upgrade --extra base-runtime -o requirements-base-runtime.txt pyproject.toml
```
alexrashed, line 55: Why would I need these individual targets? I always have to update all of them, right?

Makefile
```diff
@@ -172,19 +198,15 @@ docker-create-push-manifests:	## Create and push manifests for a docker image (d
...
+	    bash -c "make install-test && apt-get install -y --no-install-recommends gnupg && mkdir -p /etc/apt/keyrings && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && echo \"deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main\" > /etc/apt/sources.list.d/nodesource.list && apt-get update && apt-get install -y --no-install-recommends nodejs && DEBUG=$(DEBUG) PYTEST_LOGLEVEL=debug PYTEST_ARGS='$(PYTEST_ARGS)' TEST_PATH='$(TEST_PATH)' TINYBIRD_PYTEST_ARGS='$(TINYBIRD_PYTEST_ARGS)' TINYBIRD_DATASOURCE='$(TINYBIRD_DATASOURCE)' TINYBIRD_TOKEN='$(TINYBIRD_TOKEN)' TINYBIRD_URL='$(TINYBIRD_URL)' CI_COMMIT_BRANCH='$(CI_COMMIT_BRANCH)' CI_COMMIT_SHA='$(CI_COMMIT_SHA)' CI_JOB_URL='$(CI_JOB_URL)' CI_JOB_NAME='$(CI_JOB_NAME)' CI_JOB_ID='$(CI_JOB_ID)' make test"
```
alexrashed, line 209: q: What's the removal of `pip uninstall -y argparse dataclasse` about?

Makefile
```diff
@@ -228,9 +250,12 @@ test-docker-mount-code:
 
 lint:              		  ## Run code linter to check code style and check if formatter would make changes
 	($(VENV_RUN); python -m ruff check --show-source . && python -m black --check .)
+	$(VENV_RUN); pre-commit run upgrade-deps-if-changed --files $(git diff master --name-only)
```
alexrashed, line 253: The precommit doesn't really upgrade the requirements, right? And if so, I think it should be named the same way as in the make targets (where it's "reqs").

alexrashed, line 253: This make target should not test on changes, but perform the whole check. Is this the case here? Why is it the same as in `lint-modified`?

setup.cfg
```diff
@@ -94,19 +94,15 @@ runtime =
...
-    jsonschema<=4.19.0
     moto-ext[all]==5.0.0.post1
     opensearch-py>=2.4.1
     pymongo>=4.2.0
     pyopenssl>=23.0.0
 
-# @deprecated - use extra 'runtime' instead.
-full =
-    %(runtime)s
-
```
alexrashed, line 107: Are you sure this extra is not used anywhere anymore? Because this would basically break all these usages.

pyproject.toml
```diff
@@ -88,3 +88,6 @@ log_cli = true
 log_level = "DEBUG"
 log_cli_format = "%(asctime)s.%(msecs)03d %(levelname)5s --- [%(threadName)12s] %(name)-26s : %(message)s"
 log_cli_date_format = "%Y-%m-%dT%H:%M:%S"
+
+[tool.pip-tools]
+unsafe-package = ["pip", "setuptools", "distribute"] # packages that should not be pinned
```
alexrashed, line 93: q: What is the `distribute` package and why is it unsafe?

alexrashed (changes requested):
This PR will solve so many issues in our build pipelines! 💯 🦸🏽 😄 
It already looks great, I only have a few questions and nitpicks (some confusion about naming and some somwhat unrelated changes).
We also need to make sure that the pipeline cache keys are adjusted accordingly.

Afterwards we should be good to go! 🚀

Makefile
silv-io: The comment above the make target mentioned that this was a temporary fix to resolve some issue. I checked if removing that to keep the target more simple would now also work. Since it does, I would say we can keep that out now.

alexrashed: Okay, thanks for cleaning up!

Makefile
silv-io: The pre-commit hook is written in such a way that it needs to receive a list of files which it runs against. I can also just hardcode "setup.cfg" in the files option here instead.
What it does here currently is only run the check if the setup.cfg changed compared to the master branch.

setup.cfg
silv-io: I searched for any usage of `make full` in both community and ext with ripgrep and did not find any usage of it there. Is there any other place I could search?

alexrashed: I prefer to use the GitHub search for tasks like this: https://github.com/search?q=org%3Alocalstack+%22%5Bfull%5D%22&type=code

Looking good!

dfangl (approved):
Looks good from my side! Great work!

Issue #10118, opened 2024-01-25
bug: StepFunctions: Doesnt support ErrorPath


utibeabasi6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When i try to deploy a Step functions workflow on local stack with AWS CDK, i get the following error
```
[...]

PR #10130, opened 2024-01-26
[SFN]: Support for CausePath and ErrorPath

MEPalma:
## Motivation
On September 7, 2023, AWS extended the language support for StepFunctions to include `CausePath` and `ErrorPath` in `Fail` states. These source the error and cause strings from the state, or pass the result of some intrinsic function. This PR adds support for such language extensions.
Closes: #10118 

## Changes
- Added lexer and grammar support for such language extension
- Slightly reworked abstraction of Error and Cause declarations, to allow the preprocessor to recognise ErrorPath and CausePath declarations and Error and Cause declarations
- Added evaluation components for ErrorPath with JsonPath and IntrinsicFunction bindings
- Added evaluation components for CausePath with JsonPath and IntrinsicFunction bindings
- Added relevant tests


localstack/services/stepfunctions/asl/component/state/state_fail/error_path.py
```diff
@@ -0,0 +1,51 @@
...
+            StatesFunctionNameType.Format,
+            StatesFunctionNameType.JsonToString,
+            StatesFunctionNameType.ArrayGetItem,
+            StatesFunctionNameType.Base64Decode,
+            StatesFunctionNameType.Base64Encode,
+            StatesFunctionNameType.Hash,
+            StatesFunctionNameType.UUID,
+        ],
+    )
+)
```
dominikschubert, line 25: nit: a more idiomatic version would be a comprehension like `_STRING_RETURN_FUNCTIONS: Final[set[str]] = { typ.name() for typ in [....] }`

dominikschubert (approved):
Minor nit regarding more idiomatic map/set, otherwise LGTM!

PR #10133, opened 2024-01-29
Rework internal lambda routes to avoid router modifications for every spawned environment

dfangl:
## Motivation
With #10129, adding routes became way slower than before, due to fixes in regards of thread safety.
In lambda, we will add 5 routes for every spawned execution environment, and delete them after it shuts down.
This introduces unnecessary changes in the router, which can be solved faster and easier with an internal construct.

Core target of this PR is to reduce adding/removing of routes to the router.

## Changes
* Add a `ExecutorRouter` (name to be decided) which will pass the request on to the right executor endpoint. It has a routing rule which parameterizes the executor id, which leads to only 5 routes for all executors.
* `ExecutorEndpoint`s will add / remove themselves from this "router".
* I tried to self contain the changes as much as possible
* I removed unnecessary log entries and the method `_build_executor_endpoint`, as it just pollutes the log, without any real world use.

### DIFF
--- a/localstack/services/lambda_/invocation/executor_endpoint.py
+++ b/localstack/services/lambda_/invocation/executor_endpoint.py
@@ -1,3 +1,4 @@
+import abc
 import logging
 from concurrent.futures import CancelledError, Future
 from http import HTTPStatus
@@ -5,11 +6,11 @@ from typing import Dict, Optional
 
 import requests
 from werkzeug import Request
-from werkzeug.routing import Rule
 
-from localstack.http import Response, Router
+from localstack.http import Response, route
 from localstack.services.edge import ROUTER
 from localstack.services.lambda_.invocation.lambda_models import InvocationResult
+from localstack.utils.objects import singleton_factory
 from localstack.utils.strings import to_str
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


joe4dev (approved):
LGTM 👍 

It adds another layer of indirection but it makes sense to consolidate Lambda-internal routing given the many route modifications upon version startup and shutdown.

I didn't notice a clear difference using the sync invoke part of `test_number_of_functions` because the Docker overhead and variability are too big.

localstack/services/lambda_/invocation/executor_endpoint.py
```diff
@@ -36,97 +37,128 @@ def __init__(self, message):
...
+    def status_error(self, request: Request, env_id: str, executor_id: str) -> Response:
+        endpoint = self.endpoints[executor_id]
+        return endpoint.status_error(request, executor_id)
+
+
+@singleton_factory
+def executor_router():
+    router = ExecutorRouter()
+    ROUTER.add(router)
+    return router
```
thrau, line 104: not the best solution to have this global, but :shrug: probably better than to add extra layers right now

thrau (approved):
nice! i think we could have gotten away with just leaving it the way it is and rely on the rolo fix. but glad we got it over the line!

PR #10135, opened 2024-01-29
add dynamic SQS endpoint strategy to build SQS queues from called host

### LINKED ISSUE #9753: Bring back SQS queue "dynamic" format

### Enhancement description

Before 3.x, SQS would use the hostname and port of the incoming request to determine the URL to return, so the URL would work both if it was queried from the host or from inside the docker (compose) network. This was a critical and incredibly useful feature for us. While we have tried to not depend on it by providing environment variables that the users of our tool should use, some teams depend on third party libraries that query the URL and won't have the capacity to replace them. 

Additionally, we have the requirement of running with published random ports and there are use cases that get the URL both from the host and the docker network. While we will be looking into workarounds and try to find a way to still upgrade (since the [AWS SDK compatibility fix](https://github.com/localstack/localstack/issues/8267) is also critical) I don't think it's possible to have something that works in both use cases (host and network).

Honestly, I think it was awesome that LocalStack did this and I would have kept it as the default, so I hope you consider and decide bringing it back, maybe as a `SQS_ENDPOINT_STRATEGY` option.

### 🧑‍💻 Implementation

_No response_

### Anything else?

https://localstack-community.slack.com/archives/CMAFN2KSP/p1700665707589479

thrau:
## Motivation

When there is no way to control the port creation of the localstack gateway port (say you're get a random port mapping `:34567->:4566/tcp` from docker), there's currently no good way of configuring localstack s.t. endpoints are accessible from both the host and inside the container network. The ideal solution would be to gain control of the port creation, and if you need random ports, make sure the gateway is bound to the same port by setting `GATEWAY_LISTEN=:<random-port>`.

In #9753, users mentioned that they work in an environment where this is not possible, and they rely on random ports. In this scenario, SQS queue URLs won't work either from the host or from within the container network. Users have asked us to put in place the old "dynamic" endpoint strategy which allowed queue URL generation based on the incoming host header. This way you can run `ListQueues` or `GetQueueUrl` from both the host or the container, and will receive URLs that you can always call from where you resolved them. This PR adds it back to unblock the users, albeit as an undocumented flag.

This is not a problem unique to SQS, which is why we've been reluctant to add this back. Services like API Gateway, AppSync, ELB, or any other service that generates routes/endpoints exposed through the gateway, would have to have the same functionality, which is significant effort. Moving forward, instead we should address the issue of diverging gateway ports (different port on the host than inside the container) in some generalized way.

## Changes

* add `SQS_ENDPOINT_STRATEGY=dynamic` which generates queue URLs dynamically based on the host. the pattern follows the `path` strategy, but where the host url part is populated dynamically from the incoming request. e.g., `http://localhost:9999/queue/us-east-1/000000000000/foobar` 
* fixes #9753

## Testing

tested this manually using
* `python -m localstack.dev.run -p 9999:4566 -e SQS_ENDPOINT_STRATEGY=dynamic` and then things like
* `aws --endpoint-url=http://localhost:9999 sqs create-queue --queue-name foobar`, ...
* `curl "http://localhost:9999/queue/us-east-1/000000000000/foobar?Action=ReceiveMessage"`
* but also `curl "http://localhost:4566/queue/us-east-1/000000000000/foobar?Action=ReceiveMessage"`

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -346,6 +346,12 @@ class SqsQueue:
         elif config.SQS_ENDPOINT_STRATEGY == "path":
             # https?://localhost:4566/queue/us-east-1/00000000000/my-queue (us-east-1)
             host_url = f"{scheme}://{host_definition.host_and_port()}/queue/{self.region}"
+        elif config.SQS_ENDPOINT_STRATEGY == "dynamic":
+            # undocumented strategy to help unblock folks who depended on the feature previously,
+            # especially when running localstack in the container on a different port than the one exposed
+            # by docker (let's say :<random-port>->:4566/tcp.). however, this should be handled in a more
+            # fundamental way in localstack.
+            host_url = f"{context.request.host_url.rstrip('/')}/queue/{self.region}"
         else:
             host_url = f"{scheme}://{host_definition.host_and_port()}"
 


baermat (approved):
LGTM as quickfix, but I agree, we need to find a good way to do this moving forward

PR #10136, opened 2024-01-29
Remove the upper bound on the botocore pin

simonrw:
## Motivation

During the release of v3.1.0, AWS released an incompatible update of `botocore`. This disrupted our release process, so we put an upper limit to a known supported version. We no longer need the upper limit.

## Changes

Reverts a4645ad6cfec


alexrashed (approved):
Thanks for taking care of removing the pin again! 🧹💯

simonrw:
Obsoleted by #10112

PR #10142, opened 2024-01-30
Enhance implementation of is_comma_delimited_list to fix --services in pods CLI

whummer:
## Motivation

Enhance implementation of `is_comma_delimited_list(..)` to fix --services in pods CLI.

We are currently not accepting dashes in items of comma-separated lists, which is required, e.g., for `cognito-idp` in the following command:
```
$ localstack -d pod save --services s3,cognito-idp test-123
...
❌ Error: Input the services as a comma-delimited list
```

## Changes

* adjust the logic in `is_comma_delimited_list(..)` to allow passing an `item_regex`
* define a default `DEFAULT_REGEX_LIST_ITEM` regex that allows alphanumeric characters as well as dashes
* addjust and extend the unit tests

As far as I can see, the util function is currently only used for the pods client logic, so it should not break existing logic in other places.

### DIFF
--- a/localstack/utils/collections.py
+++ b/localstack/utils/collections.py
@@ -29,6 +29,9 @@ import cachetools
 
 LOG = logging.getLogger(__name__)
 
+# default regex to match an item in a comma-separated list string
+DEFAULT_REGEX_LIST_ITEM = r"[\w-]+"
+
 
 class AccessTrackingDict(dict):
     """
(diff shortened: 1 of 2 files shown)


giograno (approved):
Thanks for the quick fix @whummer. Was this related to your recent errors when saving a pod?

whummer:
> Was this related to your recent errors when saving a pod?

@giograno Partially, at least this one was reproducible. Will keep digging more and will keep you posted if I discover more issues 👍

PR #10143, opened 2024-01-30
Enhance parity for DynamoDB to Kinesis stream integration

whummer:
## Motivation

Following a customer request, it seems that our integration between DynamoDB tables and Kinesis streams is not fully working for certain DDB operations, in particular when using the `transact_write_items` operation. (Apparently it worked for older versions prior to 3.0, but I haven't confirmed).

🚧  At this point, the PR only adds a snapshot test to illustrate the behavior - proper logic still needs to be implemented. (may need some help from service owners here, if possible 🙌 ) 

Wondering if this TODO comment could be related (as I know @bentsku has been working on some pretty significant performance enhancements recently 🚀 ): https://github.com/localstack/localstack/blob/45d39edbdc27c00d1f762c40e3c050d5b1b36e39/localstack/services/dynamodb/provider.py#L1112

**Update**: After discussing with @bentsku , the TODO above is not related - the fix was essentially related to converting a table ARN to the table name.

## Changes

* enhance logic in `has_streams_enabled(..)` to extract the table name from incoming table ARN
* fix attributes in forwarded `record`s, for better parity
* add a `de_dynamize_record` util function and refactor `dynamize_value` to use boto `TypeSerializer` utils
* add a `test_enable_kinesis_streaming_destination` snapshot test to cover the functionality

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -239,55 +239,6 @@ class TestDynamoDB:
         aws_client.dynamodb.delete_table(TableName=table_name)
 
     @markers.aws.only_localstack
-    def test_stream_spec_and_region_replacement(self, aws_client):
-        ddbstreams = aws_client.dynamodbstreams
-        kinesis = aws_client.kinesis
-        table_name = f"ddb-{short_uid()}"
-        resources.create_dynamodb_table(
-            table_name,
-            partition_key=PARTITION_KEY,
-            stream_view_type="NEW_AND_OLD_IMAGES",
-            client=aws_client.dynamodb,
-        )
-
-        table = aws_client.dynamodb.describe_table(TableName=table_name)["Table"]
-
-        # assert ARN formats
-        expected_arn_prefix = "arn:aws:dynamodb:" + TEST_AWS_REGION_NAME
-        assert table["TableArn"].startswith(expected_arn_prefix)
-        assert table["LatestStreamArn"].startswith(expected_arn_prefix)
-
-        # test list_streams filtering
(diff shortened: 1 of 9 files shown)


tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -238,55 +238,6 @@ def test_list_tags_of_resource(self, aws_client):
 
         aws_client.dynamodb.delete_table(TableName=table_name)
 
-    @markers.aws.only_localstack
```
whummer, line 241: Moving this test to a new `test_dynamodb_streams.py` file (only refactoring, no functional changes)..

bentsku:
I can add a comment related to the comment: for `BatchWriteItem`, we always prepped the records as if they would be sent down to a stream, which took time. I've refactored `BatchWriteItem` as it was the affected operation at the time, and introduced logic to not prep records if no streams were attached to the targeted tables.

The comment was the only change in `TransactWriteItems`, so I'm not aware of anything that would have broken it. I think the only changes around 3.0 were multi accounts and the performance work. I hope I did not break anything with unrelated changes, but as the other operations still "stream", I suppose it could be a minor issue.

Refactoring `TransactWriteItems` to follow the same kind of model as `BatchWriteItems` would be good I think, and a first step in trying to fix the issue.

whummer:
@bentsku Kudos for fixing the tests and pushing the PR over the line! 🚀 I cannot approve my own PR, but LGTM! :)

giograno (approved):
🚀

bentsku (approved):
I'll also approve my own work 😄 the -ext pipeline is also all green for this PR!

PR #10149, opened 2024-01-31
CFn: improve error message for invalid ref

simonrw:
## Motivation

When we cannot resolve a `Ref` - specifically if it is targeting a non-existent resource or parameter, we currently print "Should be detected earlier." which is not very useful. Instead, since this is a programming error from the creator of the template, we should be more helpful and print out the ref that cannot be resolved.

## Changes

- Increase clarity of error message when not finding a ref
- Add test to ensure the ref target is in the error message
- Add more clarity to additional error messages

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -40,7 +40,7 @@ from localstack.utils.objects import get_all_subclasses
 from localstack.utils.strings import to_bytes, to_str
 from localstack.utils.threads import start_worker_thread
 
-from localstack.services.cloudformation.models import *  # noqa: F401, isort:skip
+from localstack.services.cloudformation.models import *  # noqa: F401, F403, isort:skip
 from localstack.utils.urls import localstack_host
 
 ACTION_CREATE = "create"
@@ -169,7 +169,9 @@ def resolve_ref(
     # resource
     resource = resources.get(ref)
     if not resource:
-        raise Exception("Should be detected earlier.")
+        raise Exception(
+            f"Resource target for `!Ref {ref}` could not be found. Is there a resource with name {ref} in your stack?"
+        )
 
     return resources[ref].get("PhysicalResourceId")
 
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -327,7 +327,10 @@ def _resolve_refs_recursively(
             )
             # TODO: we should check the deployment state and not try to GetAtt from a resource that is still IN_PROGRESS or hasn't started yet.
             if resolved_getatt is None:
-                raise DependencyNotYetSatisfied(resource_ids=resource_logical_id, message="")
+                raise DependencyNotYetSatisfied(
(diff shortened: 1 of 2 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -169,7 +169,9 @@ def resolve_ref(
     # resource
     resource = resources.get(ref)
     if not resource:
-        raise Exception("Should be detected earlier.")
+        raise Exception(
+            f"Resource target for `!Ref {ref}` could not be found. Is there a resource with name {ref} in your stack?"
```
dominikschubert, line 173: *with the logical resource id

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -325,7 +327,10 @@ def _resolve_refs_recursively(
             )
             # TODO: we should check the deployment state and not try to GetAtt from a resource that is still IN_PROGRESS or hasn't started yet.
             if resolved_getatt is None:
-                raise DependencyNotYetSatisfied(resource_ids=resource_logical_id, message="")
+                raise DependencyNotYetSatisfied(
+                    resource_ids=resource_logical_id,
+                    message=f"Could not resolve attribute {attribute_name} on resource {resource_logical_id}",
```
dominikschubert, line 332: nit: enclosing replaced values with quotation marks might make this a bit more readable

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -361,7 +366,7 @@ def _resolve_refs_recursively(
             none_values = [v for v in join_values if v is None]
             if none_values:
                 raise Exception(
-                    "Cannot resolve CF fn::Join %s due to null values: %s" % (value, join_values)
+                    f"Cannot resolve CF Fn::Join {value} due to null values: {join_values}"
```
dominikschubert, line 369: Here we're using the full canonical version while using !Ref above. Would suggest keeping this consistent. (Personal preference would be slightly towards canonical)

tests/aws/services/cloudformation/engine/test_references.py
```diff
@@ -53,3 +55,42 @@ def test_fn_sub_cases(self, deploy_cfn_template, aws_client, snapshot):
         )
 
         snapshot.match("outputs", deployment.outputs)
+
+
+@markers.aws.only_localstack
+def test_useful_error_when_invalid_ref(deploy_cfn_template):
```
dominikschubert, line 61: Good idea to add a test here tp avoid us messing it up when refactoring on the future. We might want to pull out tests like this from the file in the future though.

dominikschubert (approved):
LGTM thanks for jumping on this. Minor change but will be super useful for anyone debugging their Stacks!

localstack/services/cloudformation/engine/template_deployer.py
simonrw: I was trying to not be too technical in this message. I think referring to "name" is more actionable, rather than "logical resource id" which is technically correct.

PR #10152, opened 2024-01-31
Fix bootstrap tests failing in non-default region

viren-nadkarni:
## Summary

This PR fixes the following bootstrap test when run in non-default account ID/region:

- `tests.bootstrap.test_cosmetic_configuration.TestLocalStackHost.test_scenario`

## Changes

The S3 bucket had the account ID and region hardcoded in the name within the CDK template. When the tests were run in a different account ID or region, the bucket name would be different causing the test to fail.

## Tests

See `ci/circleci: bootstrap-tests` job for commit 707ebf99770af066be5eb6f220e0f14e2b1b6a1a 🟢 - this run used non-default test credentials to demonstrate that the tests pass

### DIFF
--- a/tests/bootstrap/test_cosmetic_configuration.py
+++ b/tests/bootstrap/test_cosmetic_configuration.py
@@ -11,6 +11,7 @@ from botocore.exceptions import ClientError
 from localstack import constants
 from localstack.aws.connect import ServiceLevelClientFactory
 from localstack.config import in_docker
+from localstack.constants import TEST_AWS_REGION_NAME
 from localstack.testing.pytest import markers
 from localstack.testing.pytest.container import ContainerFactory, LogStreamFactory
 from localstack.testing.scenario.cdk_lambda_helper import load_python_lambda_to_s3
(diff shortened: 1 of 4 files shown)


tests/bootstrap/test_cosmetic_configuration.py
```diff
@@ -164,7 +170,7 @@ def create_lambda_function(
             infra.add_custom_setup(
                 lambda: load_python_lambda_to_s3(
                     s3_client=aws_client.s3,
-                    bucket_name=infra.get_asset_bucket(),
```
viren-nadkarni, line 167: https://github.com/localstack/localstack/blob/b3686f173f6d7b5be2f7e4ffd3098c103c1cbce9/localstack/testing/scenario/provisioning.py#L102-L105

simonrw: I think this makes sense in this case, but we probably want to allow `infrastructure_setup` to take account/region. Then this helper would be OK I think? cc @dominikschubert 

See

https://github.com/localstack/localstack/blob/896ba4b3ff1f5ca6431dd5d9fb52c2b7f5a89d77/tests/bootstrap/conftest.py#L22-L37

We build a new client factory here without accepting region/account configuration.

simonrw (approved):
Thanks for tackling this issue. I agree this should be fixed, but I am unsure if the change you have made is correct. I have approved since it works, but this might need to be upstreamed into the CDK setup code to cover more cases. What do you think @dominikschubert?

tests/bootstrap/test_cosmetic_configuration.py
dominikschubert: I'll make a note of it, since I'll soon try to rework the provisioner a bit for asset support as well :+1:  For now the solution seems fine, the bucket naming was mostly done for compatibility with live-AWS which isn't really relevant for bootstrap tests here

dominikschubert (approved):
LGTM

PR #10161, opened 2024-02-01
Stop DNS server on shutdown

simonrw:
## Motivation

While looking through a hanging test I saw the DNS server was alive waiting on a `.join` call. This _may_ have contributed to incorrect test shutdown. Regardless, we do not shut the server down properly.

## Changes

Add a shutdown hook for the DNS server for clean termination.

Note: the priority of this hook must come after any hooks that use the gateway, which itself is shutdown with a priority of -20:

https://github.com/localstack/localstack/blob/bda6479827bf3bcf92a54d469588698187af6186/localstack/aws/serving/edge.py#L31-L35

in case DNS resolution is required for any requests using the gateway. The configuration of these priorities is brittle however. To prevent possible import problems I have not made `DNS_SHUTDOWN_PRIORITY` relative to other priorities, which would be more future proof. Instead, I have noted that the priority must be lower than the other priorities, but this may break in the future.

### DIFF
--- a/localstack/dns/plugins.py
+++ b/localstack/dns/plugins.py
@@ -26,3 +26,13 @@ def setup_dns_configuration_on_host():
             server.setup_network_configuration()
     except Exception as e:
         LOG.warning("error setting up dns server: %s", e)
+
+
+@hooks.on_infra_shutdown()
+def stop_server():
+    try:
+        from localstack.dns import server
+
+        server.stop_servers()
+    except Exception as e:
+        LOG.warning("Unable to stop DNS servers: %s", e)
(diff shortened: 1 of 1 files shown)


localstack/dns/plugins.py
```diff
@@ -26,3 +26,13 @@ def setup_dns_configuration_on_host():
             server.setup_network_configuration()
     except Exception as e:
         LOG.warning("error setting up dns server: %s", e)
+
+
+@hooks.on_infra_shutdown()
+def stop_server():
```
dfangl, line 32: The only issue I have with it, is that we should make sure that we do not do any request after the infra shutdown. All external http requests will probably fail without DNS.

Extensions maybe, among others.  Could we maybe decrease the priority here, so we can make sure we stop the dns after these hooks are executed?

dfangl (approved):
Just one minor comment regarding shutdown order :)

PR #10163, opened 2024-02-01
Update lambda init binary

dfangl:
## Motivation
We pulled upstream changes into our lambda init binary. (see https://github.com/localstack/lambda-runtime-init/pull/30).

We should now update the binary, and set the account id in the environment, so it can properly returned when the `Lambda-Extension-Accept-Feature` header is set to `accountId` during the extension register call (https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#extensions-registration-api-a)

## Changes
* Update lambda init binary
* Set account id of function into execution environment


dfangl:
https://github.com/localstack/localstack-ext/actions/runs/7799613007/job/21271112203 ext run fails with non-related tests only, so this should be good to go :)

joe4dev (approved):
I am happy to get these Lambda init update changes integrated 🚀 (at least a first iteration)

In the future, it would be great to support the Telemetry API to enable more Lambda extensions, and ultimately true X-ray tracing ✨

PR #10166, opened 2024-02-02
Feature: Lambda: Add aws validated test for two lambdas subscribing to same kinensis event stream

maxhoheiser:
## Motivation
This snapshot test is added to remove questions about supported behavior differences between AWS and LocalStack, as requested by a user. Two lambda functions can be triggered by the same kinesis event stream, if the `StartingPosition="TRIM_HORIZON"` is correctly configured. 
`StartingPosition="LATEST"` does not reliably pick up every event from the subscribed stream. 
If a new subscribed lambda should not be triggered backward for all still existing events in the stream, `StartingPosition="AT_TIMESTAMP"` in combination with a valid timestamp is recommended.

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_integration_kinesis.py
+++ b/tests/aws/services/lambda_/test_lambda_integration_kinesis.py
@@ -124,6 +124,71 @@ class TestKinesisSource:
         assert int(math.log10(timestamp)) == 9
 
     @markers.aws.validated
+    def test_create_kinesis_event_source_mapping_multiple_lambdas_single_kinesis_event_stream(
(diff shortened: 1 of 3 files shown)


alexrashed (changes requested):
This PR still contains the hacky pipeline step to debug the disk space issues from #10186.
Please rebase on `master` again instead.

alexrashed (approved):
Hmm, seems like I can't just "comment" after I requested changes once. 😄 
The snapshot test is looking good. Great to see a snapshot validated test for this specific advanced case. 💯

PR #10168, opened 2024-02-02
[SFN] Support for TestState Api Action

MEPalma:
## Motivation
Currently the StepFunctions V2 provider does not support the newly introduced AWS Api Action <a href="https://docs.aws.amazon.com/step-functions/latest/apireference/API_TestState.html" target="_blank">TestState</a>. This PR adds base support of such action to both the provider and interpreter.

## Changes
- Introduced minimal StaticAnalyser workflow to walk the program's AST and raise errors: used to detect unsupported language features for TestState derivations
- Added a new Preprocessor to map the ASL's AST to the TestState evaluations components
- Added a new TestStateProgram for evaluating single state
- Added a new Environment type TestStateEnvironment to allow evaluation components to log debug/trace data during execution
- Added relevant aspects to evaluation components to use the TestStateEnvironment to log their outputs
- Added various Execution components to allow the provider to dispatch TestState executions
- Other related changes
- Added tests suite for all tracing scenarios for each supported state type 

\*\* The conversion of TestState Static Analysers' failures into StepFunctions InvalidDefinition errors will be introduced in a follow-up PR.


localstack/services/stepfunctions/asl/parse/asl_parser.py
```diff
@@ -48,11 +48,11 @@ def __repr__(self):
 
 class AmazonStateLanguageParser(abc.ABC):
     @staticmethod
-    def parse(src: str) -> Program:
+    def parse(derivation: str) -> tuple[EvalComponent, ParserRuleContext]:
```
joe4dev, line 51: I guess changing this signature is required to get more flexibility for integrating static analysis

joe4dev, line 51: derivation vs definition vs src? (see terminology comment below)

joe4dev (changes requested):
Awesome to see the static analyzer coming 🎉 

The testing endpoint can be useful for isolating issues and maybe even for us being able to do parity testing with fewer dependencies 👍 


Overall, the changes look good. My main concern is regarding code duplication in the tests. I think parameterized tests could help a lot to make them more maintainable. I also added some other comments (e.g., Lambda runtime updates) that would be good to fix.

Great work overall 👏

localstack/services/stepfunctions/backend/execution.py
```diff
@@ -218,46 +219,60 @@ def _to_serialized_date(timestamp: datetime.datetime) -> str:
...
             exec_worker.stop(stop_date=stop_date, cause=cause, error=error)
 
-    def _publish_execution_status_change_event(self):
+    def publish_execution_status_change_event(self):
         input_value = (
             dict() if not self.input_data else to_json_str(self.input_data, separators=(",", ":"))
         )
-        output_value = self.output
+        output_value = (
+            None if self.output is None else to_json_str(self.output, separators=(",", ":"))
```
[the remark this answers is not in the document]
MEPalma: I can confirm we do want to isolate the None case, we have some tests that assert this behaviour.

localstack/services/stepfunctions/provider.py
```diff
@@ -252,7 +260,7 @@ def _validate_definition(definition: str):
             exception_args = list(exception.args)
             invalid_definition = InvalidDefinition()
             invalid_definition.message = (
-                f"Error={exception_name} Args={exception_args} in definition '{definition}'."
+                f"Error={exception_name} Args={exception_args} in derivation '{derivation}'."
```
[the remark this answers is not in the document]
MEPalma: Thank you for this, aws calls SFN language derivations "definitions", so this is better if reverted to definition

tests/aws/services/stepfunctions/v2/test_state/test_test_state_scenarios.py
```diff
@@ -0,0 +1,864 @@
...
+    paths=[
+        "$..loggingConfiguration",
+        "$..tracingConfiguration",
+        "$..SdkHttpMetadata",
+        "$..SdkResponseMetadata",
+    ]
+)
+class TestStateCaseScenarios:
+    @staticmethod
+    def _send_test_state_request(aws_client_factory, **kwargs):
```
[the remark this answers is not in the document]
MEPalma: For TestState calls, boto will prepend "sync-" to the endpoint string. As we operate on localhost, this function creates a new client with that functionality disabled, and forwards the test_state call. 
As for now this is limited to TestState calls, my initial idea was to make this issue explicit. Creating a step functions client that always disables this feature I though could hide this issue in other/future instances. Maybe this may result in more efficient customer feedback.
I propose adding a fixture for SFN client but include TestState in the name, to improve readability of the test suite but force future occurrences to handle this issue explicitly.

localstack/services/stepfunctions/asl/parse/asl_parser.py
MEPalma: Correct, I will change the "derivation" naming here too

localstack/services/stepfunctions/asl/parse/test_state/asl_parser.py
```diff
@@ -0,0 +1,38 @@
...
+        parser.removeErrorListeners()
+        parser.addErrorListener(syntax_error_listener)
+        tree = parser.state_decl_body()
+
+        errors = syntax_error_listener.errors
+        if errors:
+            raise ASLParserException(errors=errors)
+
+        # Attempt to preprocess the AST into evaluation components.
+        preprocessor = TestStatePreprocessor()
```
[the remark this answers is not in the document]
MEPalma: Indeed, however it was my intention to wait to abstract this parsing process until a pilot static analyser for the program, in order to see if support objects my be necessary. I am not confident we have a clear enough idea of all the different components and call sequences needed here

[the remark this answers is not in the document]
joe4dev: makes a lot of sense, thanks for the clarification 👍

localstack/services/stepfunctions/asl/parse/test_state/preprocessor.py
```diff
@@ -48,20 +48,20 @@ def _decorated_updates_inspection_data(method, inspection_data_key: InspectionDa
...
 
     return wrapper
 
 
 def _decorate_state_field(state_field: CommonStateField) -> None:
     if isinstance(state_field, ExecutionState):
         state_field._eval_execution = _decorated_updates_inspection_data(
-            method=state_field._eval_execution,
-            inspection_data_key=InspectionDataKey.RESULT,  # noqa
+            method=state_field._eval_execution,  # noqa: as part of the decoration we access this protexted member.
```
joe4dev, line 59: super nit: protected

FYI: you can disable individual [rules](https://docs.astral.sh/ruff/rules/) using `# noqa: F401` (example for unused imports)

tests/aws/services/stepfunctions/conftest.py
```diff
@@ -146,6 +147,11 @@ def transform(self, input_data: dict, *, ctx: TransformContext) -> dict:
         return input_data
 
 
+@pytest.fixture
+def stepfunctions_client_test_state(aws_client_factory):
```
joe4dev, line 151: nit: adding a comment on why this is necessary would be nice-to-have (e.g., using your explanation in the PR)

tests/aws/services/stepfunctions/v2/test_state/test_test_state_scenarios.py
```diff
@@ -0,0 +1,246 @@
...
+
+from localstack.aws.api.lambda_ import Runtime
+from localstack.aws.api.stepfunctions import InspectionLevel
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+from tests.aws.services.stepfunctions.templates.services.services_templates import (
+    ServicesTemplates as ST,
+)
+from tests.aws.services.stepfunctions.templates.test_state.test_state_templates import (
+    TestStateTemplate as TCT,
```
joe4dev, line 14: (What does the C stand for?)

tests/aws/services/stepfunctions/v2/test_state/test_test_state_scenarios.py
```diff
@@ -0,0 +1,246 @@
...
+        sfn_role_arn = create_iam_role_for_sfn()
+
+        template = TCT.load_sfn_template(tct_template)
+        definition = json.dumps(template)
+
+        test_case_response = stepfunctions_client_test_state.test_state(
+            definition=definition,
+            roleArn=sfn_role_arn,
+            input=execution_input,
+            inspectionLevel=InspectionLevel.INFO,
```
joe4dev, line 79: The inspection level could be extra parameter and pytest automatically covers all combinations (e.g., https://stackoverflow.com/a/72215438/6875981).
Downside: the snapshot skips would apply to everything (I don't think we support conditional snapshot skips based on params); it would be pretty easy in this case (INFO ✅ , others ❌ )

joe4dev (approved):
LGTM ✅ 🚀 

The tests are much clearer and maintainable now using parametrization 👍 Thanks for fixing the test IDs as well 👏 Very thorough refinements 💯

tests/aws/services/stepfunctions/v2/test_state/test_test_state_scenarios.py
MEPalma: I agree, the snapshot skips required do represent a downside, but I am also wondering if expanding the parameter size and combinations would yield returning benefits at this point. I say this as test ids will also need expanding (probably appending the inspection level) and may make it inconvenient, although feasible, to rerun individual tests in the future.
I will add a comment to investigate aggregating all `test_base_` into a single parametrised function

joe4dev: definitely a very valid point 👍 good pragmatic solution 👍 👏

tests/aws/services/stepfunctions/v2/test_state/test_test_state_scenarios.py
```diff
@@ -11,7 +11,7 @@
     ServicesTemplates as ST,
 )
 from tests.aws.services.stepfunctions.templates.test_state.test_state_templates import (
-    TestStateTemplate as TCT,
+    TestStateTemplate as TST,
```
joe4dev, line 14: nice extra catch ✨ 💯 
(I got confused by this as well)

MEPalma: Thank you!

PR #10172, opened 2024-02-04
fix dynamodb router rule removal

thrau:
## Motivation

shutdown hook of dynamodb was using an old API to remove rules from the router, leading to exceptions:

```
  File "/opt/code/localstack/localstack/utils/functions.py", line 41, in call_safe
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/dynamodb/provider.py", line 481, in on_before_stop
    ROUTER.remove_rule(rule)
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'Router' object has no attribute 'remove_rule'
```

## Changes

* dynamodb routes are now removed using `ROUTER.remove`

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -477,8 +477,7 @@ class DynamoDBProvider(DynamodbApi, ServiceLifecycleHook):
 
     def on_before_stop(self):
         self._expired_items_worker.stop()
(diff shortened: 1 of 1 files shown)


thrau:
thanks @viren-nadkarni ! totally missed your PR

Issue #9872, opened 2023-12-14
bug: Step Functions aws-sdk:sfn:startExecution requires manual JSON conversion


guillaume-macquat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm executing a state machine using the following definition:

_(It's not a valid definition on AWS because the Parameters keys in the state StartTarget start with a lowercase for compatibility with LocalStack, see [9853](https://github.com/localstack/localstack/issues/9853))_

```json
{
  "StartAt": "SetupVariables",
  "States": {
    "SetupVariables": {
      "Type": "Pass",
      "Next": "StartTarget",
      "Parameters": {
        "input": {
          "key": "value"
        }
      }
    },
    "StartTarget": {
      "Type": "Task",
      "Resource": "arn:aws:states:::aws-sdk:sfn:startExecution",
      "Parameters": {
        "stateMachineArn": "arn:aws:states:eu-west-3:000000000000:stateMachine:my-state-machine",
        "input.$": "$.input"
      },
      "End": true
    }
  }
}
```

It fails with the following error:
```json
{
  "error": "States.Runtime",
[...]

PR #10174, opened 2024-02-05
[SFN] Enhance Automatic Serialisation of AWS-SDK Task Parameters

MEPalma:
## Motivation
When incorporating task states that invoke AWS services directly through the AWS SDK integrations (such as Lambda, StepFunctions, DynamoDB, S3, etc.), these typically require the definition of task parameters in the Amazon States Language (ASL) within the state machine definition. This enables AWS Step Functions to transmit these parameters directly to the AWS service, obviating the need for manual conversion of dictionaries into JSON strings. However, the interpreter currently does not perform such json serialisation automatically. This PR adds such behaviour.
Closes: #9872  

## Changes
- Added automatic json serialisation within the workflow that reduces Step Functions Parameters to boto3 parameters.
- Relevant aws-sdk test that induces such behaviour and reproduces related issue


tests/aws/services/stepfunctions/templates/services/services_templates.py
```diff
@@ -23,6 +23,9 @@ class ServicesTemplates(TemplateLoader):
     AWS_SDK_SFN_START_EXECUTION: Final[str] = os.path.join(
         _THIS_FOLDER, "statemachines/aws_sdk_sfn_start_execution.json5"
     )
+    AWS_SDK_SFN_START_EXECUTION_2: Final[str] = os.path.join(
+        _THIS_FOLDER, "statemachines/aws_sdk_sfn_start_execution_2.json5"
```
dominikschubert, line 27: nit: `_2` could be replaced with something a bit more descriptive like `_implicit_json_conversion` or something similar

dominikschubert (approved):
minor nit otherwise LGTM :+1:

PR #10176, opened 2024-02-05
add APIGW RequestParametersResolver support for context var

bentsku:
## Motivation
There was a #TODO for the RequestParametersResolver to properly resolve `context` variables to integration request parameters. 
This was also lacking for a customer use case with AppSync. 

https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html#mapping-request-parameters

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference

## Changes
- added the context variable manually into the dict used for resolving integration request parameters (this can be revisited  later with the `context.context` variable and more logic into it)
- added an AWS validated test to check we're properly passing integration request parameters and resolving them


calvernaz (approved):
Thanks for adding this, we will find the right objects to use on the templates, and potentially just pass the right "context" shape - unfortunately is not consistent across all APIs on AWS

PR #10183, opened 2024-02-06
pin botocore with ASF API updates, add ASF API forward compatibility

alexrashed:
## Motivation
This PR addresses multiple issues around the ASF API updates.
In the past we had issues several times because a `botocore` update slipped in (`botocore` was not pinned), while the ASF APIs have not been updated (they only run every Monday morning).
This can then cause issues in the fallback forwarding due to our strict APIs:
- The operation calls are dispatched using keyword arguments.
- Our parser is based on the specs contained in the `botocore` package.
  - This means that our parser can parse new parameters from an incoming request if the botocore package is updated, but the operation's method signature has not been updated yet (adding this new parameter to the list of keyword arguments).
- When dispatching the parsed request (including a new parameter), we can a `TypeError`.

The last instance of this issue occurred during our 3.1 release: [CircleCI Job # 175308](https://app.circleci.com/pipelines/github/localstack/localstack/21796/workflows/21384f16-5972-45fd-9f36-c4baeb090787/jobs/175308/tests)
```
botocore.exceptions.ClientError: An error occurred (InternalError) when calling the CreateRouteTable operation (reached max retries: 0): exception while calling ec2.CreateRouteTable: Traceback (most recent call last):
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 112, in __call__
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 82, in __call__
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 168, in dispatch_request
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/forwarder.py", line 135, in _call
    return handler(context, req)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/api/core.py", line 161, in operation_marker
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
TypeError: Ec2Api.create_route_table() got an unexpected keyword argument 'client_token'
```

This PR aims at solving this issue with two different approaches at once:
- We pin `botocore` and only update it together with the generated APIs.
- We implement forward-compatibility in the ASF APIs by adding `**kwargs` to all ASF API definitions and implementations.
  - This is necessary, because we have other projects which are also generating APIs based on `botocore`. In these projects we cannot pin `botocore`, because it's transitively pinned though `localstack-core`.

This PR is basically an extension of https://github.com/localstack/localstack/pull/10127 (now that we have proper dependency pinning, explicitly ping botocore and tie it to the ASF updates).

## Changes
The changes are clearly separated in different commits:
1. updates the scaffold to add `**kwargs` to all generated API operations.
2. pins the version of `botocore` in the `setup.cfg` to the version which was last used to update the ASF APIs (https://github.com/localstack/localstack/pull/10173).
3. changes the ASF Update workflow to also update the pin on `botocore`.
4. updates the generated APIs to reflect the latest changes in the scaffold
5. updates the service provider implementations after the API (superclasses) updates

## Testing
- We do have a nice unit test which checks that the provider signatures do not differ from the signatures of the generated APIs: [test_asf_providers.py](https://github.com/localstack/localstack/blob/8a9845db38bdbdbf7a3da3c088564b697e655f95/tests/unit/aws/api/test_asf_providers.py)
- I verified manually that the catching all keyword args in the generated APIs fixes the issue outlined above.
- We have lots of integration tests testing the providers ;)
- I [manually ran the ASF update action](https://github.com/localstack/localstack/actions/runs/7803716373) which resulted in the following PR: https://github.com/localstack/localstack/pull/10190

### DIFF
--- a/localstack/services/apigateway/provider.py
+++ b/localstack/services/apigateway/provider.py
@@ -267,6 +267,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         stage_keys: ListOfStageKeys = None,
         customer_id: String = None,
         tags: MapOfStringToString = None,
+        **kwargs,
     ) -> ApiKey:
         api_key = call_moto(context)
 
@@ -277,7 +278,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
 
         return api_key
 
-    def get_rest_api(self, context: RequestContext, rest_api_id: String) -> RestApi:
+    def get_rest_api(self, context: RequestContext, rest_api_id: String, **kwargs) -> RestApi:
         rest_api: RestApi = call_moto(context)
         remove_empty_attributes_from_rest_api(rest_api)
         return rest_api
@@ -287,6 +288,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         context: RequestContext,
         rest_api_id: String,
         patch_operations: ListOfPatchOperation = None,
+        **kwargs,
     ) -> RestApi:
         rest_api = get_moto_rest_api(context, rest_api_id)
 
@@ -396,6 +398,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         security_policy: SecurityPolicy = None,
         mutual_tls_authentication: MutualTlsAuthenticationInput = None,
         ownership_verification_certificate_arn: String = None,
+        **kwargs,
     ) -> DomainName:
         if not domain_name:
             raise BadRequestException("No Domain Name specified")
@@ -427,7 +430,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         return domain
 
     @handler("GetDomainName")
-    def get_domain_name(self, context: RequestContext, domain_name: String) -> DomainName:
+    def get_domain_name(self, context: RequestContext, domain_name: String, **kwargs) -> DomainName:
         store: ApiGatewayStore = get_apigateway_store(context=context)
         if domain := store.domain_names.get(domain_name):
             return domain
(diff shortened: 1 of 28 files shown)


alexrashed:
The last step - updating the service provider implementations after the API superclasses were updated - would have meant to iterate over hundreds of operations manually. In order to avoid this, I wrote a small script which uses `inspect` and `libcst` to perform the modification automatically:

```python
import inspect
from collections import defaultdict
from pathlib import Path
from types import FunctionType

import libcst

from localstack.testing.aws.asf_utils import collect_implemented_provider_operations

# collect all provider operations
provider_ops = collect_implemented_provider_operations()

# iterate over all operations and group all operations which need to be modified per provider file
operations_per_provider_file: dict[str, list[str]] = defaultdict(list)
for provider_class, base_class, provider_function in provider_ops:
    try:
        sub_function = getattr(provider_class, provider_function)
    except AttributeError:
        raise AttributeError(
            f"Given method name ('{provider_function}') is not a method of the sub class ('{provider_class.__name__}')."
        )

    if not isinstance(sub_function, FunctionType):
        raise AttributeError(
            f"Given method name ('{provider_function}') is not a method of the sub class ('{provider_class.__name__}')."
        )

    if not getattr(sub_function, "expand_parameters", True):
        # if the operation on the subclass has the "expand_parameters" attribute (it has a handler decorator) set to False, we don't care
        continue

    if wrapped := getattr(sub_function, "__wrapped__", False):
        # if the operation on the subclass has a decorator, unwrap it
        sub_function = wrapped

    operations_per_provider_file[inspect.getfile(provider_class)].append(sub_function.__name__)


class ProviderMethodTransformer(libcst.CSTTransformer):
    def __init__(self, operations: list[str]):
        self.operations = operations
        super().__init__()

    def leave_FunctionDef(
        self, original_node: libcst.FunctionDef, updated_node: libcst.FunctionDef
    ) -> libcst.CSTNode:
        if original_node.name.value in self.operations and not original_node.params.star_kwarg:
            print(f"  -> {original_node.name.value}")
            new_params = updated_node.params.with_changes(
                star_kwarg=libcst.Param(name=libcst.Name("kwargs"))
            )
            return updated_node.with_changes(params=new_params)

        return updated_node


# iterate over all providers and tranform the found operations
for provider_file, ops in operations_per_provider_file.items():
    print(f"- {provider_file}")
    module = libcst.parse_module(Path(provider_file).read_bytes())
    updated_module = module.visit(ProviderMethodTransformer(ops))
    Path(provider_file).write_text(updated_module.code)
```

thrau (commented):
Changes look great! Just before we go this route: have you thought about solving this on the dispatcher level, i.e., filtering the kwargs in the Skeleton based on the handler's signature, so superfluous arguments don't even make it into the method? That would limit the necessary code changes.

I'm worried that we will become lazy with updating providers, now that we also no longer get warnings in the IDE that signatures between API stubs and providers no longer match, and drift will become worse over time.

alexrashed:
Good question, and I thought about this a lot actually. I would argue that:
- The check would be an unnecessary runtime overhead.
- This will still be enforced by the unit test mentioned above.
  - I was the only one who proactively added new parameters every Monday with the ASF PRs. There wasn't a single contribution from any service owners (also because distributing that would take longer than fixing it). I will continue doing that for now.
- Service owners would still get the reviewer notification through the `CODEOWNERS` config.
- We could actually use this in the future in some providers (if we decide to weaken the unit test enforcing the equality of the functions).
  - f.e. in this function (as discussed with @bentsku): https://github.com/localstack/localstack/blob/8e182da604f43b1999dcb61ca8c18097677490ef/localstack/services/s3/provider.py#L1260
- In the future we could even implement this adjustment to be done automatically, with a modification of the script I posted above.

thrau (approved):
well clearly you thought about it so let's go with this solution :-) other than that i don't like the look of the trailing `**kwargs`attribute in every method, i don't have a strong argument against it right now. the runtime overhead of filtering a dict would be negligible though.

anyway, great work on the overall changes!

PR #10186, opened 2024-02-06
fix disk space issues in GitHub workflow

alexrashed:
## Motivation
Recently, the Community Tests against Pro are failing somewhat regularly.
@maxhoheiser found out that they fail with an error message indicating that they are running out of disk space.
After a bit of analysis, it shows that there's a lot OpenSearch versions downloaded and cached. In order to mitigate these issues, I changed the tests a bit to avoid some OpenSearch versions to be downloaded.
If we hit the limit again, there's lot of potential in maximizing the disk space (f.e. with [easimon/maximize-build-space](https://github.com/easimon/maximize-build-space)).

## Changes
- Reduces the disk space being used by decreasing the number of OpenSearch versions being used in the tests.

## Testing
- The pipeline in question is now running stable again in this PR.
- I analyzed the cache storage logs of the following runs:
  - [Before the OpenSearch optimization: ~5553 MB](https://github.com/localstack/localstack/actions/runs/7802506071/job/21280116439)
  - [Afterwards: ~2401 MB](https://github.com/localstack/localstack/actions/runs/7805006599/job/21288220013)

## TODO
- [x] Optimize the workflow to get a proper run.
- [x] Analyze the run and see if we can optimize the tests instead of the environment.
- [x] Fix it!

### DIFF
--- a/tests/aws/services/opensearch/test_opensearch.py
+++ b/tests/aws/services/opensearch/test_opensearch.py
@@ -14,6 +14,7 @@ from opensearchpy.exceptions import AuthorizationException
 from localstack import config
 from localstack.aws.api.opensearch import AdvancedSecurityOptionsInput, MasterUserOptions
 from localstack.constants import (
+    ELASTICSEARCH_DEFAULT_VERSION,
     OPENSEARCH_DEFAULT_VERSION,
     OPENSEARCH_PLUGIN_LIST,
     TEST_AWS_ACCOUNT_ID,
@@ -60,9 +61,6 @@ def install_async():
             LOG.info("installing opensearch default version")
             opensearch_package.install()
             LOG.info("done installing opensearch default version")
-            LOG.info("installing opensearch 1.0")
-            opensearch_package.install(version="OpenSearch_1.0")
-            LOG.info("done installing opensearch 1.0")
             installed.set()
 
     start_worker_thread(run_install)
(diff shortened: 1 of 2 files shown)


dfangl (approved):
Less downloads and less disk space are always welcome :) Great job hunting down the issue @alexrashed and @maxhoheiser

tests/aws/services/opensearch/test_opensearch.py
```diff
@@ -360,13 +358,8 @@ def test_create_domain(self, opensearch_wait_for_cluster, aws_client):
             aws_client.opensearch.delete_domain(DomainName=domain_name)
 
     @markers.aws.only_localstack
-    @pytest.mark.parametrize(
-        "engine_version",
-        # Test once per major version
-        ["OpenSearch_1.3", "OpenSearch_2.5"],
```
maxhoheiser, line 366: I guess it is good enough to test with the latest version?

alexrashed: Yep, that is the intention. It is by far the version which is used most often, and it turned out that the test actually failed with the latest version (it was an issue in the test though, since the security plugin got more strict on the passwords).

maxhoheiser (commented):
Looks good, thanks @alexrashed  for unblocking the pipelines :)

PR #10193, opened 2024-02-07
[SFN] Makefile Targets for Parsers Generation

MEPalma:
## Motivation
The current development process for the ANTLR4 projects within the StepFunctions v2 Interpreter and Intrinsic Functions Interpreter presents significant hurdles for contributors. These hurdles include the necessity for developers to manually generate parsers from project grammar files, which requires the downloading of specific tools and versions, as well as the correct application of arguments. This manual process not only introduces potential for human error but also creates a barrier to entry for new developers, potentially stifling innovation and collaboration.

To address these challenges, this PR proposes the integration of makefile targets designed to automate the aspects of the development workflow: downloading the tool, building the parsers, cleaning the build, and formatting the grammar files.

## Changes
- Added default environment variables and make targets to: install the parser generator `install`, build the parsers `build`,  clear the build `clean`, and format the grammar files `format`.
- Removed generated files
- Added related .gitignore definitions


Makefile
```diff
@@ -268,4 +292,8 @@ clean-dist:				  ## Clean up python distribution directories
 	rm -rf dist/ build/
 	rm -rf *.egg-info
 
+clean-dev-antlr4:        ## Clean up the ANTLR4 project directory.
+	rm -rf $(ANTLR4_TARGET_DIR)
+	rm -rf $(ANTLR4_DIR)
+
 .PHONY: usage freeze install-basic install-runtime install-test install-dev install entrypoints dist publish coveralls start docker-save-image docker-build docker-build-multiarch docker-push-master docker-create-push-manifests docker-run-tests docker-run docker-mount-run docker-cp-coverage test test-coverage test-docker test-docker-mount test-docker-mount-code lint lint-modified format format-modified init-precommit clean clean-dist pip-tools upgrade-pinned-dependencies
```
joe4dev, line 299: All three targets should be `.PHONY`

joe4dev (approved):
That's a great initiative with a well-described PR 👏 Makefile are super useful for this scenario 👍 

✅  I tested the make targets on macOS and can confirm everything works as expected. The generated runtime files match perfectly ✨ ✨✨.

👉 My main concern is that these make targets are very specific to StepFunctions and clutter the top-level Makefile, which is already long. Would it be more cohesive to move the `Makefile` into the StepFunctions directory `localstack/services/stepfunctions/asl`?
This would also make it easier to add a short `README.md` next to it with some basic instructions on how to use it and setup requirements, such as Java version X required.

PR #10194, opened 2024-02-07
fine-tune workflow triggers

alexrashed:
## Motivation
We have quite a lot of pipelines, where some of them are even triggered if the code change does not influence the pipeline at all (f.e. `CODEOWNERS` updates). Other pipeline triggers are not respecting recent changes to the dependency management (#10127).

This PR aims at fine-tuning and updating the pipeline triggers.

## Changes
- Fine tune the pipeline triggers.
- Fixes the check that prevents the execution of the `test-pro-integration` workflow (which needs secrets to run) for forks (which don't have and should not have access to secrets).

## TODO
- [x] Check if we can easily configure CircleCI to only trigger a workflow execution if the commit touches certain files.
  - Seems like this can only be done with a fairly complex feature called [dynamic configuration](https://circleci.com/docs/using-dynamic-configuration/).
  - This is considered too complicated for now.
- [x] Configure pipelines using secrets such that they are either skipped or work without secrets if they are not available.
  - This is necessary to fix the pipeline for community contributions. PRs coming from forks do not have access to any secrets other than the GitHub token assigned to the PR.


.github/workflows/tests-pro-integration.yml
```diff
@@ -24,16 +24,21 @@ on:
...
-      - "setup.cfg"
-      - "Dockerfile"
-      - "Dockerfile.rh"
-      - "docker-compose.yml"
-      - "bin/**"
+      - '**'
+      - '!.github/**'
+      - '.github/workflows/tests-pro-integration.yml'
+      - '!doc/**'
+      - '!scripts/**'
```
dfangl, line 31: The pro workflow uses a script in `scripts/`. While it does not usually get changed, it makes sense to include it. It's `./scripts/build_common_test_functions.sh`.

dfangl (approved):
LGTM! Just one minor comment regarding an exclusion.

.github/workflows/tests-pro-integration.yml
alexrashed: Nice catch, thanks! I explicitly included this script in both triggers (`push` and `pull_request`).

PR #10195, opened 2024-02-07
remove pins in setup.cfg, add .python-version

alexrashed:
## Motivation
#10127 introduced dependency pinning.
This is a good opportunity to revisit current pins in the `setup.cfg` and try to upgrade as much as possible, as well as to remove some safety pins which have just defined to avoid breaking changes in the future (which is not necessary, because this only happens in a dedicated PR now).

## Changes
- Removes as many version pins in the `setup.cfg` as possible and updates the pins.
  - Currently the only remaining pins are:
    - `moto-ext` - explicitly updated when we create a release of our `moto` fork
    - `dill` - iirc this should be pinned to avoid a breaking change? /cc @giograno
    - `botocore` - explicitly updated with ASF updates, see https://github.com/localstack/localstack/pull/10183
    - `docker` - waiting for a release of https://github.com/docker/docker-py/issues/3195 / https://github.com/docker/docker-py/pull/3196
    - `black` - limited to the major version 23 to avoid lots of reformatting, should be upgraded separately or replaced with `ruff format`
    - `jsonpath-ng` - has been pinned with https://github.com/localstack/localstack/pull/9142, breaks skip verification handling in snapshot library (`test_dot_in_skip_verification_path`) /cc @dominikschubert 
    - `antlr4-python3-runtime` - the antlr4 runtime is tightly coupled to the generated code. The pin will remain for now, and once the generation is automated, we can automate version upgrades. See https://github.com/localstack/localstack/pull/10195#discussion_r1482942730.
- Reintroduces a `.python-version` file.
  - This is getting more important after the introduction of the dependency pinning because the dependencies can differ between python versions and the python version used to generate the lock files is recorded in the lock file itself.

## TODO
- [x] Check failures in the test pipeline (also make sure to execute downstream pipelines).
  - There are 4 remaining test failures in the pipeline testing the pro integration. These are related to the upgrade of the test naming to pytest 8: https://github.com/localstack/localstack/pull/10195/commits/6105fedfa7042f9b23dcef229b19268d40a28b40
  - This means that this PR will cause integration issues which will be resolved by upgrading to pytest 8 there as well.
  - Another option would be to duplicate the snapshot data for both naming styles, but I don't think this is necessary.
- [x] Maybe circle back on pins which are still necessary (and inform service owners + add comments).
  - After several iterations, we identified the minimum needed pins (as listed above).


setup.cfg
```diff
@@ -120,14 +120,14 @@ test =
...
+    black>=23.10.0
     coveralls>=3.3.1
     Cython
     networkx>=2.8.4
     pandoc
-    pre-commit==3.5.0
+    pre-commit>=3.5.0
     pypandoc
-    ruff==0.1.0
+    ruff>=0.1.0
```
silv-io, line 130: Looks like some of the main issues are with the linting being different now. I'd say we keep those pins until we decide on doing the switch to `ruff format`. Then we can remove `black`, upgrade `ruff` and do the formatting in one step. 

There will still be the question however what happens with the formats in the future. If there is some big change in `ruff` which would affect a lot of files, the dependency pin upgrade PR would contain a lot of these in order to be green. We would then potentially need to add a lot of those upgrade commits to the `.git-blame-ignore-revs` list.

alexrashed: Generally speaking yes, if an upgrade needs a proper "fix", it should be done outside of the update PR (where we would avoid this version in the `setup.cfg` in the update PR, and remove the pin in the "fix PR"). If it's just a few lines, an "adjustment" to the new version, I would do it in the upgrade PR itself.
We actually already do it like this with the ASF updates. Small adjustments (adjusting the provider method signatures to the changed ASF API stubs) are done in the ASF update PRs directly, while bigger fixes related to upgrades (like the sqs protocol handling) are done separately.

silv-io: So to clarify with the `black` update:
The further procedure if we continue with `black` and want to update it would be:
- Pin the `black` version to the latest one that worked and merge it like that.
- Create a PR where only the `black` pin is removed again
- In that PR regenerate the pins with the newest `black` version
- Fix the issues in the new PR (in this case just the reformat)

Correct?

silv-io: Another thing that comes to mind here is: How can we ensure that pins that get introduced because of issues like that will get resolved in a timely manner? Only setting a TODO might get lost over time, like some of the pins we already had in there. Maybe we should have it as a rule that immediately after merging an upgrade PR with a pin, we should create a PR that removes the pin again and keep it around until the proper fix happens either upstream or by us doing the fix in the PR.

tests/aws/services/stepfunctions/v2/services/test_lambda_task.snapshot.json
```diff
@@ -549,7 +549,7 @@
       }
     }
   },
-  "tests/aws/services/stepfunctions/v2/services/test_lambda_task.py::TestTaskLambda::test_invoke_json_values[00]": {
```
silv-io, line 552: Do these changes need to remain here or is the extra PR sufficient?

silv-io (approved):
LGTM! Just a small thought and question

setup.cfg
```diff
@@ -85,15 +85,16 @@ runtime =
     %(base-runtime)s
     airspeed-ext>=0.6.3
     amazon_kclpy>=2.0.6,!=2.1.0
-    antlr4-python3-runtime==4.13.1
+    antlr4-python3-runtime>=4.13.1
```
MEPalma, line 88: I would prefer having this pinned at a precise version for now, as it is preferable to have the runtime library version match the version that generated the parsers in the first place. If this is not true, warning may also be triggered, but I have also had to adjust the interpreter recently to match some changes in the latest version (4.13.1). Unfortunately, the parser generator library is run locally and needs to be manually installed; I recently tried to [formalise the build](https://github.com/localstack/localstack/pull/10193) to help resolve this gap. We'll probably need to discuss this further as another option may be to generate these grammars on install.

alexrashed: But wouldn't the test cover this case? This change is just removing the exact pin in the `setup.cfg`.
The dependency is still pinned in the requirements files (see https://github.com/localstack/localstack/pull/10127).
This change basically allows updating the pin on a weekly basis (with an automated workflow creating a PR). On this PR all tests will be triggered, and we can see if the version changes are breaking anything. If not, we'll merge the new pins. If it does, we would try to address them.

But that's your call. If you explicitly want to pin the version, I can just re-add the exact pin.

tests/aws/services/stepfunctions/v2/services/test_lambda_task.snapshot.json
alexrashed: I don't know what you mean with an extra PR?
The change here is necessary because pytest 8 changed the way paramtrized tests are named.
This change is definitely small enough to have in here and avoid the additional overhead of creating an explicit PR handling this upgrade, if this is what you mean.

setup.cfg
```diff
@@ -75,7 +75,7 @@ base-runtime =
     urllib3>=2.0.7
     Werkzeug>=3.0.0
     xmltodict>=0.13.0
-    rolo>=0.3,<0.4
```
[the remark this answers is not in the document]
alexrashed: > Currently, we can rely on the pins to keep us diligent on updating the lower bound of the version as well.
I don't think this is true, tbh.

> How will we handle increasing lower bounds of versions in the future?
We didn't really do that in the past, and there also isn't a need for this for runtime dependencies in my opinion.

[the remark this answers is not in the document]
alexrashed: > Currently, we can rely on the pins to keep us diligent on updating the lower bound of the version as well.

Actually, in this specific case as an exception, this is true. And until recently it was useful to define upper version boundary (f.e. when trying to build an older version in the future). But now we preserve the actually tested version in the requirement lock files anyways, so I think we should just handle this the same way we handle all other dependencies:
- Avoid any upper version limits if possible.
- Update the pins regularly.
- If we explicitly upgrade to a new version (a new release of `rolo` was released and we want to upgrade to it immediately, instead of waiting for the update PR). We will do so in an explicit PR. If we explicitly upgrade our code to be compatible with a new version, we can increase the minimal version to reflect the technical minimal version.

setup.cfg
MEPalma: I believe this is a fantastic idea, as it would not only keep SFN up to date with the latest ANTLR4 updates but also streamline our process. For optimal integration with SFN, it would be best if the generator library were updated simultaneously, allowing parsers to be automatically generated upon installation. If we only update the runtime library, we would have to manually regenerate the parsers to address at least the programmatic warnings that will appear. 
Therefore, I propose we enable automatic updates once we have automated the generator workflow within the installation process, ensuring both the generator and parsers are automatically updated.

alexrashed: Perfect, then I will re-introduce the pin, and you can remove it once you automated the generation :)

MEPalma: Thank you, I hope to see this soon!!

PR #10206, opened 2024-02-08
APIGW: add more debug log when lambda proxy response format is wrong

bentsku:
## Motivation
When a AWS_PROXY integration type with a lambda is used in API Gateway, the lambda must follow a specific format when responding. It's often a source of confusion when AWS Gateway returns `502` error. 
We would log the following line:
`2024-02-08T07:44:44.336  WARN --- [ asgi_gw_458] l.s.apigateway.integration : Lambda output should follow the next JSON format: { "isBase64Encoded": true|false, "statusCode": httpStatusCode, "headers": { "headerName": "headerValue", ... },"body": "..."}`

But in this specific user case, it looked like I could not find a response anywhere in the logs and would like to see what we actually received.

This is the AWS documentation page specifying the format:
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

## Changes
I would like to add to the logging line the actual payload returned, to understand why exactly the payload did not conform.
The lambda logs did not help a lot, and I'm not sure if in that case the issue is in lambda or APIGW.

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -219,7 +219,8 @@ class LambdaProxyIntegration(BackendIntegration):
         keys = parsed_result.keys()
         if "statusCode" not in keys or "body" not in keys:
             LOG.warning(
-                'Lambda output should follow the next JSON format: { "isBase64Encoded": true|false, "statusCode": httpStatusCode, "headers": { "headerName": "headerValue", ... },"body": "..."}'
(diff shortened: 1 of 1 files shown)


thrau (approved):
that's a great addition! will make it much easier to troubleshoot :100: 

time for improved logging @dominikschubert :grin:

PR #10207, opened 2024-02-09
Use fixtures for test credentials and region name

viren-nadkarni:
## Background

This PR makes some code improvements for test credentials and region.

## Changes

- ~~Remove redundant `is_aws()` helper and use `is_aws_cloud()` everywhere.~~ Added by https://github.com/localstack/localstack/pull/10164
- Across our tests, instead of directly importing constants for test account ID and test region name, fixtures are used. New fixtures for secondary test account ID and secondary region name are also introduced.


localstack/testing/pytest/snapshot.py
```diff
@@ -103,16 +98,8 @@ def pytest_runtest_call(item: Item) -> None:
         sm._assert_all(verify, paths)
 
 
-@pytest.fixture(name="region", scope="session")
-def fixture_region(aws_client):
-    if is_aws() or is_api_enabled("sts"):
-        return aws_client.sts.meta.region_name
-    else:
-        return TEST_AWS_REGION_NAME
```
viren-nadkarni, line 111: This is moved to `localstack.testing.pytest.fixtures` alongside its companion fixtures

localstack/testing/pytest/fixtures.py
```diff
@@ -1800,33 +1794,38 @@ def cleanups():
...
     if is_aws_cloud() or is_api_enabled("sts"):
         return secondary_aws_client.sts.get_caller_identity()["Account"]
     else:
-        return SECONDARY_TEST_AWS_ACCOUNT_ID
+        return constants.SECONDARY_TEST_AWS_ACCOUNT_ID
+
+
+@pytest.fixture(scope="session")
+def secondary_region_name():
+    return constants.SECONDARY_TEST_AWS_REGION_NAME
```
dominikschubert, line 1822: Not a particular fan of using the constants for the fixtures. One of the longer-running efforts will be moving to a state where we won't have access to localstack internals from the test code (e.g. because we're running LS in a container and test against that). I also think `constants` should generally not be a place where we put values used in testing.

How about we use environment variables to configure these and falling back to a static default in the fixture?

viren-nadkarni: Actually, this value already gives precedence to the env variable

https://github.com/localstack/localstack/blob/unify-fixtures/localstack/constants.py#L159-L163

I admit that its place in `localstack.constants` is misleading.

Would it be okay if I add a todo note and make the changes in a follow up PR, given how large this PR is already?

localstack/services/s3/presigned_url.py
```diff
@@ -134,7 +133,9 @@ def create_signature_does_not_match_sig_v2(
     ex = SignatureDoesNotMatch(
         "The request signature we calculated does not match the signature you provided. Check your key and signing method."
     )
-    ex.AWSAccessKeyId = TEST_AWS_ACCESS_KEY_ID  # todo maybe return access key used for the request
+    ex.AWSAccessKeyId = (
+        constants.TEST_AWS_ACCESS_KEY_ID
+    )  # todo maybe return access key used for the request
```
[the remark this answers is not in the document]
bentsku: For the presigned URLs, we will need a bit more logic especially when creating the Credentials to validate the URL and remove those constants. Right now, we explicitly say that we need the credentials to be `test` / `test` in order to validate them (so technically the test should fail when using other credentials, which is why I believe we're using `TEST_*` and not `DEFAULT_*`). 

The secret access key is not passed in the signature, so we will need a network request to fetch it like IAM does, in order to properly validate it. But I agree that implementing this TODO would be really great!

Let me know if there's anything I can do there.

tests/unit/aws/test_connect.py
```diff
@@ -92,7 +92,7 @@ def test_external_client_credentials_origin(self, mock, monkeypatch):
         )
         mock.assert_called_once_with(
             service_name="def",
-            region_name="us-east-1",
+            region_name=region_name,
```
dominikschubert, line 95: I'm not so sure we want to do this with unit tests. The behavior of the underlying region lookup seems to be not as well defined here :thinking:

tests/aws/services/s3/test_s3.py
```diff
@@ -2689,15 +2687,17 @@ def test_get_range_object_headers(self, s3_bucket, aws_client):
...
         # Boto still does not support chunk encoding, which means we can't test with the client nor
         # aws_http_client_factory. See open issue: https://github.com/boto/boto3/issues/751
         # Test for https://github.com/localstack/localstack/issues/1571
         object_key = "data"
         body = "Hello\r\n\r\n\r\n\r\n"
         headers = {
             "Authorization": mock_aws_request_headers(
-                "s3", aws_access_key_id=TEST_AWS_ACCESS_KEY_ID, region_name=TEST_AWS_REGION_NAME
+                "s3",
+                aws_access_key_id=TEST_AWS_ACCESS_KEY_ID,
```
dominikschubert, line 2699: we should be able to get this via STS as well,so might be worth to make this a fixture too

dominikschubert (approved):
LGTM, didn't spot any obvious issues.

I have a bit of a concern in terms of using this fixture for unit tests, though. I think keeping those static is fine

tests/aws/services/s3/test_s3.py
viren-nadkarni: Do you mean [`STS:GetSessionToken`](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetSessionToken.html)?

tests/unit/aws/test_connect.py
viren-nadkarni: You're right, I picked this from https://github.com/localstack/localstack/pull/8204 but I'm unable to reproduce the original reason for the change. Reverted for now, I'll revisit this if it reoccurs.

Issue #10196, opened 2024-02-07
bug: ASLParserException on CreateStateMachine operation if 'Retry' block contains 'Comment' field


Relenna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hi, 
I've encountered issue with creating state machine if 'Comment' field is present in 'Retry' block of step definition. 

`An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: ASLParserException line 76:10, at "Comment", mismatched input '"Comment"' expecting {'"ErrorEquals"', '"IntervalSeconds"', '"MaxAttempts"', '"BackoffRate"'}`

This error doesn't occur when using option `PROVIDER_OVERRIDE_STEPFUNCTIONS=legacy` 

### Expected Behavior

I expect that state machine should be created correctly, as this field is supported by AWS
For example, this is AWS state machine designer with 'hello-world' template: 
![aws-hello-world-retrier](https://github.com/localstack/localstack/assets/159269912/7cb55bbd-a835-4689-b9d6-08e76cccd6f0)

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal stepfunctions create-state-machine --name "IssueExample" --definition file://hello_world_with_retry_comment.json --role-arn "arn:aws:iam::000000000000:role/stepfunctions-role"

I've used exported definition from AWS 'hello-world' template. The only change I've made was to add retries with 'comment' field. 
[hello_world_with_retry_comment.json](https://github.com/localstack/localstack/files/14197014/hello_world_with_retry_comment.json)

### Environment

```markdown
- OS: Ubuntu 20.04.6 LTS
- LocalStack (Pro): 3.1.1.dev20240205090040
```

PR #10211, opened 2024-02-09
[SFN] Enhanced Support for Comments Declarations

MEPalma:
## Motivation
Currently the SFN v2 interpreter does not support Comments declared in Retry and Catch, as well as in choice rules. This PR addresses such issues and adds relevant snapshot tests for comments declarations.
Closes: #10196 

## Changes
- Fixed ASL grammar to allow comments in Retry, Catch, and choice rules
- Fixed evaluation components and preprocessor to propagate such values (we don't want to skip them to enhance debugging)
- Added comments positive tests (as per docs) and one negative (comments in Parameters definitions)
- Other related minor changes


dominikschubert (approved):
LGTM, was a bit confused about the intention behind the use of comments in the parameters section at first, but makes sense to include it as a sort of negative test :+1:

PR #10218, opened 2024-02-12
First attempt to externalize CFn resource providers

dominikschubert:
## Motivation

To facilitate external contributions, we'd like to externalize the CloudFormation resource providers into their own repository/lib. The plugin-based providers make this quite easy already, so this is a first attempt at externalized at least a few resource providers.

## Changes

* Adds dependency on extracted lib: https://github.com/localstack/cloudformation-resource-providers
* Removes internal IAM resource providers
* Alias `OperationStatus` to the extracted one for the equality check to work

## TODO

Lots of things still need to happen for this to be productive, e.g.
- Split resource_providers.py into part that is required for the lib and part that we only need in localstack
- Better utils for CFn resource providers
- Extract testing of individual providers at least into providers repo
- more, TBD


dominikschubert:
Closing this as we currently don't plan to continue with this for the time being and there's not much here that would be hard to recreate (adding a dependency and aliasing the OperationStatus).

PR #10221, opened 2024-02-12
Restore DNS settings in container after dns server shuts down

dfangl:
## Motivation
We are currently seeing failures due to DNS resolution issues on pytest session finish hooks in LS.
This is likely due to the shutdown of the DNS server after LS shuts down (which can happen before the last session finish hooks are executed).

After we stop the DNS server, we should restore the `resolv.conf` config to avoid all subsequent requests failing.

## Changes
* Cache and restore `resolv.conf` on DNS server start/stop.

### DIFF
--- a/localstack/dns/plugins.py
+++ b/localstack/dns/plugins.py
@@ -39,6 +39,7 @@ def stop_server():
     try:
         from localstack.dns import server
 
+        server.revert_network_configuration()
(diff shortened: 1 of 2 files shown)


simonrw (approved):
Looks a reasonable change to fix the pipeline. Thanks

PR #10223, opened 2024-02-12
Sqs fix message groups visibility

### LINKED ISSUE #10107: bug: can't receive message from fifo queue with the same MessageGroupId

### Current Behavior

with python boto3 client

- send a message to fifo queue with MessageGroupId='1'
- receive the message
- delete the message
- send another message to the queue with MessageGroupId='1'
- try to receive the message, get none
- check message availability with get_queue_attributes, get `{'ApproximateNumberOfMessages': '1', 'ApproximateNumberOfMessagesNotVisible': '0', 'ApproximateNumberOfMessagesDelayed': '0', ...`

### Expected Behavior

expecting to receive message from the queue

With a docker-compose file

### Steps To Reproduce

docker-compose up

### Environment

```markdown
- OS:MacOS Sonoma 14.2.1
- LocalStack: latest
```

### Anything else?

I tried to run with Rosetta, no result

baermat:
## Motivation
Users reported an issue with message groups becoming blocked when polling fifo queues. It turned out that a [recent change](https://github.com/localstack/localstack/pull/9975) that was generally correct had a side effect on empty message groups which we didn't consider.
The current implementation relies on removing message groups from the group queue when receiving until the queue is empty. We cannot iterate over the group queue, and it seems to be impractical to add the groups back when new messages arrive (since we then have to distinguish between empty groups due to all messages processed and empty groups due to invisible messages). Therefore this PR collects all empty groups on receive and adds them back to the group queue once the receive call is finished.

## Changes
- add  empty message groups back to the queue on receive.
- fixes #10107

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -922,11 +922,7 @@ class FifoQueue(SqsQueue):
         """
         with self.mutex:
             if message_group_id not in self.message_groups:
-                # a newly created message group is added to the queue immediately
-                message_group = self.message_groups[message_group_id] = MessageGroup(
-                    message_group_id
-                )
-                self.message_group_queue.put_nowait(message_group)
+                self.message_groups[message_group_id] = MessageGroup(message_group_id)
 
             return self.message_groups.get(message_group_id)
 
@@ -1018,14 +1014,19 @@ class FifoQueue(SqsQueue):
         message_group = self.get_message_group(message.message_group_id)
 
         with self.mutex:
+            previously_empty = message_group.empty()
             # put the message into the group
             message_group.push(message)
 
-            # if an older message becomes visible again in the queue, that message's group becomes visible also.
-            if message.receive_count < 1:
+            # new messages should not make groups visible that are currently inflight
+            if message.receive_count < 1 and message_group in self.inflight_groups:
                 return
+            # if an older message becomes visible again in the queue, that message's group becomes visible also.
             if message_group in self.inflight_groups:
                 self.inflight_groups.remove(message_group)
+                self.message_group_queue.put_nowait(message_group)
+            # if the group was previously empty, it was not yet added back to the queue
+            elif previously_empty:
                 self.message_group_queue.put_nowait(message_group)
 
     def remove_expired_messages(self):
@@ -1072,7 +1073,6 @@ class FifoQueue(SqsQueue):
         start = time.time()
 
         received_groups: Set[MessageGroup] = set()
-        empty_groups: Set[MessageGroup] = set()
 
         # collect messages over potentially multiple groups
         while True:
@@ -1088,7 +1088,6 @@ class FifoQueue(SqsQueue):
                 #  want to block the group
                 # TODO: check behavior in case it happens if all messages were removed from a group due to message
                 #  retention period.
-                empty_groups.add(group)
                 timeout -= time.time() - start
                 if timeout < 0:
                     timeout = 0
@@ -1157,11 +1156,6 @@ class FifoQueue(SqsQueue):
                 else:
                     self.inflight.add(message)
 
-        # finally, add all empty groups back to the queue since there are potentially no inflight messages
-        # to put them back
-        for group in empty_groups:
-            self.message_group_queue.put_nowait(group)
-
         return result
 
     def _on_remove_message(self, message: SqsMessage):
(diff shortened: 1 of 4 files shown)


localstack/services/sqs/models.py
```diff
@@ -1083,7 +1086,8 @@ def receive(
             self.inflight_groups.add(group)
 
             if group.empty():
-                # this can be the case if all messages in the group are still invisible
+                # this can be the case if all messages in the group are still invisible or
+                # if all messages of a group have been processed.
```
baermat, line 1090: While investigating I saw that the timeout code in this block is not used later on, should we remove it?

localstack/services/sqs/models.py
```diff
@@ -1154,6 +1157,11 @@ def receive(
                 else:
                     self.inflight.add(message)
 
+        # finally, add all empty groups back to the queue since there are potentially no inflight messages
+        # to put them back
+        for group in empty_groups:
+            self.message_group_queue.put_nowait(group)
```
baermat, line 1163: RIght now this is not part of the mutex block since the initial pop from the queue isn't either

thrau, line 1163: Should we check if the group is empty here? If in the meantime messages were added, would the group be in the queue twice?

thrau (commented):
Thanks a lot for digging into this, the fix looks good to me! I just have one question to clarify.

Overall (outside the scope of this PR - since it's just fixing existing behavior), I still need to make sense of:
* Are message groups ever removed correctly?
* What happens when group 1 is empty, we block to wait on the next group, and in the meantime messages appear in group 1? Is it possible we run into scenarios where we're not receiving messages that we should?

localstack/services/sqs/models.py
baermat: You are right, that can happen at the moment. Then we should probably make that empty check + requeue atomic. It wouldn't do any harm I think on a semantic level (the queues are "emptied" sequentially, so a subsequent duplicate message group "should" be empty) but it's probably begging for trouble in the long run, also from a performance perspective

baermat:
> Are message groups ever removed correctly? What happens when group 1 is empty, we block to wait on the next group, and in the meantime messages appear in group 1? Is it possible we run into scenarios where we're not receiving messages that we should?

While writing this wall of text I had an idea which I will commit as new solution to this PR. The new solution will add back groups on delete only if they are non-empty; empty groups will be added back a) when messages become visible again, making the group visible too b) when new messages arrive for non-inflight empty groups

A good question, it jogged my understanding of the blocking, which I didn't really consider when writing this PR, and which poses a problem:
Right now, an empty group being de-queued while waiting for messages means that we won't receive messages for this group while waiting.
I didn't consider that the blocking for `waitTimeSeconds` effectively relies on the group queue being empty/empty groups not being queued until they get new message, **supposedly**.
But it doesn't hold even with the current implementation I think. A fully processed group will be in the queue but without messages (this is addressed in the new solution, we won't do that anymore), will then be dequeued, and not be eligible for receiving messages while we wait, UNLESS we re-queue the message group when a new message comes in. The initial solution for this PR did exactly that, however it comes down to the problem of test `test_fifo_message_group_visibility`: a message group that is inflight should not be made available by new messages.

localstack/services/sqs/models.py
```diff
@@ -1018,15 +1014,20 @@ def _put_message(self, message: SqsMessage):
...
+            # new messages should not make groups visible that are currently inflight
+            if message.receive_count < 1 and message_group in self.inflight_groups:
                 return
+            # if an older message becomes visible again in the queue, that message's group becomes visible also.
             if message_group in self.inflight_groups:
                 self.inflight_groups.remove(message_group)
                 self.message_group_queue.put_nowait(message_group)
+            # if the group was previously empty, it was not yet added back to the queue
+            elif previously_empty:
+                self.message_group_queue.put_nowait(message_group)
```
thrau, line 1030: nit: Should we consolidate some of the branches, and make the unhandled case explicit with an exception, or at least a log message?

```python

            if message_group in self.inflight_groups:
                # new messages should not make groups visible that are currently inflight
                if message.receive_count < 1:
                    return

                # if an older message becomes visible again in the queue, that message's group becomes
                # visible also.
                self.inflight_groups.remove(message_group)
                self.message_group_queue.put_nowait(message_group)
            # if the group was previously empty, it was not yet added back to the queue
            elif previously_empty:
                self.message_group_queue.put_nowait(message_group)
            else:
                raise ValueError(
                    f"Message group {message_group.message_group_id} is in an unknown state"
                ) # or LOG.warning(...)
```

thrau (approved):
thank you bernhard for all the clarifying comments. this solution is really much cleaner! especially removing the automatic adding of message groups into the queue makes it much easier to understand.

also looks good from a performance perspective :+1: great job! :clap:

PR #10227, opened 2024-02-13
Fix Lambda function alias name checks

dominikschubert:
## Motivation

A user reported an issue where invoking a lambda with an alias like `2024-01-01` would fail.

## Changes

- Introduce test covering both a valid case like `2024-01-01` and invalid one like `2024`
- Fix the alias regex we're using to check if a qualifier is an alias or not
- Add a check to the `create_alias` method. In the reported case, the alias was created but we only checked the regex on invoke which then failed to resolve the proper alias.

### DIFF
--- a/localstack/services/lambda_/api_utils.py
+++ b/localstack/services/lambda_/api_utils.py
@@ -83,7 +83,7 @@ VERSION_REGEX = re.compile(r"^[0-9]+$")
 # Pattern for an alias qualifier
 # Rules: https://docs.aws.amazon.com/lambda/latest/dg/API_CreateAlias.html#SSS-CreateAlias-request-Name
 # The original regex from AWS misses ^ and $ in the second regex, which allowed for partial substring matches
-ALIAS_REGEX = re.compile(r"(?!^[0-9]+)(^[a-zA-Z0-9-_]+$)")
+ALIAS_REGEX = re.compile(r"(?!^[0-9]+$)(^[a-zA-Z0-9-_]+$)")
 # Permission statement id
 STATEMENT_ID_REGEX = re.compile(r"^[a-zA-Z0-9-_]+$")
 
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -1636,6 +1636,54 @@ class TestLambdaAlias:
             )
         snapshot.match("alias_does_not_exist_esc", e.value.response)
 
+    @markers.aws.unknown
+    def test_alias_naming(self, aws_client, snapshot, create_lambda_function_aws, lambda_su_role):
+        """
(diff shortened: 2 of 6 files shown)


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -1636,6 +1636,55 @@ def test_notfound_and_invalid_routingconfigs(
             )
         snapshot.match("alias_does_not_exist_esc", e.value.response)
 
+    @markers.snapshot.skip_snapshot_verify(paths=["$..LoggingConfig"])
```
dominikschubert, line 1639: Another reason to tackle this soonish

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -1636,6 +1636,55 @@ def test_notfound_and_invalid_routingconfigs(
...
+            FunctionName=function_name,
+            Handler="index.handler",
+            Code={
+                "ZipFile": create_lambda_archive(
+                    load_file(TEST_LAMBDA_PYTHON_ECHO), get_content=True
+                )
+            },
+            PackageType="Zip",
+            Role=lambda_su_role,
+            Runtime=Runtime.python3_9,
```
dfangl, line 1656: Why python 3.9?

localstack/services/lambda_/api_utils.py
```diff
@@ -83,7 +83,7 @@
 # Pattern for an alias qualifier
 # Rules: https://docs.aws.amazon.com/lambda/latest/dg/API_CreateAlias.html#SSS-CreateAlias-request-Name
 # The original regex from AWS misses ^ and $ in the second regex, which allowed for partial substring matches
-ALIAS_REGEX = re.compile(r"(?!^[0-9]+)(^[a-zA-Z0-9-_]+$)")
+ALIAS_REGEX = re.compile(r"(?!^[0-9]+$)(^[a-zA-Z0-9-_]+$)")
```
dfangl, line 86: I am pretty sure the last `$` in the regex is not necessary if used with match and not search, but it does not hurt either :)

dfangl (approved):
LGTM!

tests/aws/services/lambda_/test_lambda_api.py
dominikschubert: copypasta

dominikschubert: Let's update them all when we do the LoggingConfig where we need to regenerate all the snapshots anyway

joe4dev: we might be able to save a lot of time on image pulling (at least in some runners not testing the multi-runtime tests) if we can make our entire test suite use the new al2023 images (so much smaller: 3.11=>3.12 from 781.95 MB =>557.9 MB).

PR #10230, opened 2024-02-13
Handler chain must not use test variables

viren-nadkarni:
## Background

Certain functions in the request handler chain used test variables as fallback.

Test variables (`localstack.constants.TEST_*`) must not be used outside of tests.

## Changes

- Fallback to the default account ID i.e. `000000000000` if cannot be derived from access key ID

Some code reorganisation:
- Move `get_account_id_from_request()` to `localstack.utils.aws.request_context`, and remove the now empty `localstack.aws.proxy` module.
- Remove redundant function `extract_region_from_auth_header()` from `localstack.utils.aws.aws_stack` and use the one defined in `localstack.utils.aws.request_context`
- Move `extract_access_key_id_from_auth_header()` from `localstack.utils.aws.aws_stack` to `localstack.utils.aws.request_context`


thrau (approved):
LGTM! thanks for cleaning this up :broom: 

hoping we can get rid of the flask request object soon :-)

PR #10238, opened 2024-02-13
Implement dynamodb TimeToLive in cloudformation

giograno:
## Motivation
With https://github.com/localstack/localstack/pull/10041 we recently introduced the TTL in DynamoDB.
While working on a platform test today @lukqw noticed that this feature was not working when deploying a DDB table with cdk. To enable TTL in DDB you would first create the table with the `CreateTable` operation and then call the `UpdateTimeToLive` operation to enable the feature and specify the attribute name.
The `AWS::DynamoDB::Table` template has a `TimeToLiveSpecification` attribute that carries the same payload we would pass to `UpdateTimeToLive`. However, we were not reading this attribute and performing the additional call in our `DynamoDBTableProvider`.

## Changes
- We implemented the missing logic in the `DynamoDBTableProvider` cloudformation provider. We read the eventual `TimeToLiveSpecification` and eventually make a `UpdateTimeToLive` call.
- We added a verified cdk-based test with the same configuration as the platform test that let surface the issue.

### DIFF
--- a/localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
+++ b/localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
@@ -231,12 +231,6 @@ class DynamoDBTableProvider(ResourceProvider[DynamoDBTableProperties]):
                     **self.get_ddb_kinesis_stream_specification(model)
                 )
 
-            if model.get("TimeToLiveSpecification"):
-                request.aws_client_factory.dynamodb.update_time_to_live(
-                    TableName=model["TableName"],
-                    TimeToLiveSpecification=model["TimeToLiveSpecification"],
-                )
-
             return ProgressEvent(
                 status=OperationStatus.IN_PROGRESS,
                 resource_model=model,
@@ -252,6 +246,12 @@ class DynamoDBTableProvider(ResourceProvider[DynamoDBTableProperties]):
                 status=OperationStatus.IN_PROGRESS,
                 resource_model=model,
                 custom_context=request.custom_context,
+            )
+
+        if model.get("TimeToLiveSpecification"):
+            request.aws_client_factory.dynamodb.update_time_to_live(
+                TableName=model["TableName"],
+                TimeToLiveSpecification=model["TimeToLiveSpecification"],
             )
 
         if description["Table"].get("LatestStreamArn"):
--- a/localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
+++ b/localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
@@ -231,6 +231,12 @@ class DynamoDBTableProvider(ResourceProvider[DynamoDBTableProperties]):
                     **self.get_ddb_kinesis_stream_specification(model)
                 )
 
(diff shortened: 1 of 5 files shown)


lukqw (approved):
Thanks for taking care of this issue!

localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
```diff
@@ -231,6 +231,12 @@ def create(
                     **self.get_ddb_kinesis_stream_specification(model)
                 )
 
+            if model.get("TimeToLiveSpecification"):
+                request.aws_client_factory.dynamodb.update_time_to_live(
+                    TableName=model["TableName"],
+                    TimeToLiveSpecification=model["TimeToLiveSpecification"],
+                )
+
```
simonrw, line 239: Just to check: I assume this operation can be done on a table that is still being created? Otherwise we should delay this call until e.g. (new) line 257

tests/aws/services/cloudformation/resources/test_dynamodb.py
```diff
@@ -137,3 +140,26 @@ def test_global_table(deploy_cfn_template, snapshot, aws_client):
...
+        time_to_live_attribute="expire_at",
+    )
+
+    cdk.CfnOutput(stack, "TableName", value=table.table_name)
+
+    with infra.provisioner() as prov:
+        outputs = prov.get_stack_outputs(stack_name="DDBStackTTL")
+        table_name = outputs["TableName"]
+        table = aws_client.dynamodb.describe_time_to_live(TableName=table_name)
+        snapshot.match("table", table)
```
simonrw, line 165: question: do we have coverage of the TTL feature itself? I realise
* this test covers the CFn part, and the TTL behaviour should be tested with the service tests, and
* we rely on a third party package for the implementation so we should defer to the tests of that implementation,
but I still think it would be good to. have.

tests/aws/services/cloudformation/resources/test_dynamodb.py
```diff
@@ -137,3 +140,26 @@ def test_global_table(deploy_cfn_template, snapshot, aws_client):
 
     error_code = ex.value.response["Error"]["Code"]
     assert "ResourceNotFoundException" == error_code
+
+
+@markers.aws.validated
+def test_ttl_cdk(aws_client, snapshot, infrastructure_setup):
```
simonrw, line 146: 🎉 for writing a CDK test!

simonrw (approved):
Looks good from a CFn perspective, just a couple of questions

localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
giograno: Good point. I moved the code down below. AWS triggers a `ResourceNotFoundException` if the table is not yet there.

tests/aws/services/cloudformation/resources/test_dynamodb.py
giograno: Yup, we do have tests that cover TTL itself in the DDB suite.

PR #10240, opened 2024-02-14
Lambda: Fix test_cross_account_access flakes

viren-nadkarni:
## Background

The test `TestLambdaMultiAccounts:test_cross_account_access` would fail if `short_uid()` returned a value consisting entirely of digits. This failed Boto input validation for `AliasName`

## Changes

- Make sure `AliasName` contains alphabets
- Adds a note that prevented me from AWS validating the test. Is this a bug in AWS? Did this go unnoticed for three years? Not sure


dfangl (approved):
Thanks for the comments regarding the permission, and great catch!

PR #10241, opened 2024-02-14
enhance download progress logging

alexrashed:
## Motivation
In LocalStack, we do download some packages / backends lazily when they are needed for the first time.
For example, when creating an OpenSearch domain with a specific version for the first time, the package will be downloaded and cached in the LocalStack volume directory.
Currently, the logs are quite verbose and log on every MB downloaded. For big packages, this can pollute the logs quite a lot.
This PR aims at reducing the amount of logs if possible: If the total size is known (e.g. the `Content-Length` header is set on the HTTP response), we log at most one line for every 10% of download progress.

## Changes
- If the total content length is known, only log every 10% of the progress, instead of every MB.
- Fixes a small bug where the totally downloaded size was miscalculated for GZIP encoded responses (due to a transparent decompression in `urllib3` / `requests`).

## Testing
- Added a unit test explicitly testing the log messages for streamed downloads with and without the content length set.

### DIFF
--- a/localstack/utils/http.py
+++ b/localstack/utils/http.py
@@ -205,7 +205,7 @@ def download(
         if r.headers.get("Content-Length"):
             total_size = int(r.headers.get("Content-Length"))
 
-        total_written = 0
+        total_downloaded = 0
         if not os.path.exists(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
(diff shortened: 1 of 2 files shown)


viren-nadkarni (approved):
:+1:

PR #10242, opened 2024-02-14
Load new resource providers by default if available and fall back to legacy models 

dominikschubert:
## Motivation

Adding new resource types is quite annoying right now, especially in our -ext repo, since every new provider needs to be accompanied by a companion PR in community to actually enable that provider.

We never really switched back to a legacy provider and I also don't really see us doing that in the near future. We're also nearly done with the provider migration and if there are any issues coming up will prioritze fixing them in the resource providers over the legacy models.

## Changes

- New resource lookup order: 1. PRO provider (if PRO enabled), 2. Community provider, 3. legacy model
  - This means as long as the entrypoints are generated, any scaffolded provider in -ext will be picked up immediately without adding anything in `resource_provider.py` anymore.
- Remove our manual default override list with all new resource providers, since they are now prioritized over legacy models anyway.

### DIFF
--- a/localstack/services/cloudformation/resource_provider.py
+++ b/localstack/services/cloudformation/resource_provider.py
@@ -50,146 +50,6 @@ Properties = TypeVar("Properties")
 
 PUBLIC_REGISTRY: dict[str, Type[ResourceProvider]] = {}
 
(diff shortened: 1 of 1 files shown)


simonrw (approved):
Looks like a good change, thanks for making this the default

PR #10243, opened 2024-02-14
Feature: Scenario Tests: Firehose central scenario test

maxhoheiser:
## Motivation
To further close gaps between LocalStack and AWS more scenario tests focusing on combining multiple services are required, with these two requirements:

1. close gaps in existing integration tests for streaming services: 🎯goal: focus on integration tests on services that are a. frequently used & b. not well covered currently
2. implement standard scenarios of service workflows used by our customers & on AWS in general: 🎯goal: represent the most common behavior LocalStack is used for to guarantee that this works for code changes

Scenarios meeting both requirements should be ranked in order to focus on the most relevant ones first. These 4 ranking criteria are proposed: 

1. how much are the involved services used by customers
2. how many integration tests exist covering the scenario
    1. how many integration tests can be covered with a single scenario test
    2. how long does the covered integration test run
3. are bot services supporting CDK deployment
4. how “common” is the scenario (common = used in real-production scenarios)

This PR and consecutive PRs building on it are focused on streaming services, and will consecutively implement this scenario:

Firehose Scenario Test

## Changes
- Add CDK stack for Kinesis, Firehose, and S3
- Add basic test for messages flowing through the pipeline

### DIFF
--- a/tests/aws/scenario/kinesis_firehose/conftest.py
+++ b/tests/aws/scenario/kinesis_firehose/conftest.py
@@ -38,7 +38,7 @@ def get_all_messages_from_s3(read_s3_data):
         timeout: int = 300,
         sleep: int = 5,
         retries: int = 3,
-        assert_message_count: int = 4,
+        expected_message_count: int = 4,
     ) -> list[str]:
         # poll file from s3 buckets
         def get_all_messages():
@@ -48,7 +48,7 @@ def get_all_messages_from_s3(read_s3_data):
                 json_array_string = "[" + input_string.replace("}{", "},{") + "]"
                 message = json.loads(json_array_string)
                 messages.extend(message)
-            if len(messages) != assert_message_count:
+            if len(messages) != expected_message_count:
                 raise Exception(f"Failed to receive all sent messages: {messages}")
             else:
                 return messages
--- a/tests/aws/scenario/kinesis_firehose/conftest.py
+++ b/tests/aws/scenario/kinesis_firehose/conftest.py
@@ -3,45 +3,39 @@ import json
 from localstack.utils.sync import retry
 
 
-def read_s3_data(aws_client, bucket_name: str, timeout: int = 10) -> dict[str, str]:
-    def _get_data():
-        response = aws_client.s3.list_objects(Bucket=bucket_name)
-        if response.get("Contents") is None:
-            raise Exception("No data in bucket yet")
+def read_s3_data(aws_client, bucket_name: str) -> dict[str, str]:
+    response = aws_client.s3.list_objects(Bucket=bucket_name)
+    if response.get("Contents") is None:
+        raise Exception("No data in bucket yet")
 
-        keys = [obj.get("Key") for obj in response.get("Contents")]
+    keys = [obj.get("Key") for obj in response.get("Contents")]
 
-        bucket_data = dict()
-        for key in keys:
-            response = aws_client.s3.get_object(Bucket=bucket_name, Key=key)
-            data = response["Body"].read().decode("utf-8")
-            bucket_data[key] = data
-        return bucket_data
-
-    bucket_data = retry(_get_data, sleep=1, retries=timeout)
+    bucket_data = dict()
+    for key in keys:
+        response = aws_client.s3.get_object(Bucket=bucket_name, Key=key)
+        data = response["Body"].read().decode("utf-8")
+        bucket_data[key] = data
     return bucket_data
 
 
-def get_all_messages_from_s3(
+def get_all_expected_messages_from_s3(
     aws_client,
     bucket_name: str,
-    timeout: int = 300,
     sleep: int = 5,
     retries: int = 3,
-    expected_message_count: int = 4,
+    expected_message_count: int | None = None,
 ) -> list[str]:
-    # poll file from s3 buckets
     def get_all_messages():
-        s3_data_bucket = read_s3_data(aws_client, bucket_name, timeout=timeout)
+        bucket_data = read_s3_data(aws_client, bucket_name)
         messages = []
-        for input_string in s3_data_bucket.values():
+        for input_string in bucket_data.values():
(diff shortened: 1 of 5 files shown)


tests/aws/scenario/kinesis_firehose/test_kinesis_firehose.py
```diff
@@ -0,0 +1,192 @@
...
+                ),
+                PartitionKey="1",
+            )
+        # delete messages from bucket after read
+        cleanups.append(s3_empty_bucket(bucket_name))
+
+        bucket_data = get_all_messages_from_s3(
+            bucket_name,
+            timeout=300,
+            assert_message_count=message_count,
```
simonrw, line 190: This argument is redundant with the snapshot matching

tests/aws/scenario/kinesis_firehose/test_kinesis_firehose.py
```diff
@@ -0,0 +1,192 @@
...
+                iam.PolicyStatement(
+                    effect=iam.Effect.ALLOW,
+                    actions=["logs:PutLogEvents", "logs:CreateLogStream"],
+                    resources=[firehose_s3_log_group.log_group_arn],
+                ),
+            ],
+        )
+        role_firehose_s3.attach_inline_policy(policy_firehose_s3)
+
+        extended_s3_destination_configuration = firehose.CfnDeliveryStream.ExtendedS3DestinationConfigurationProperty(
```
simonrw, line 118: Idea: when this scenario is fleshed out, it would be cool to build a `Construct` for this resource so we don't have to deal with manual roles/policies and some of this destination configuration can be streamlined

tests/aws/scenario/kinesis_firehose/conftest.py
simonrw, line 1: I think the fixtures in this file can be streamlined. There is only really one fixture that is used in a test: `get_all_messages_from_s3` and there are no teardown steps, so perhaps this pair of fixtures would be better as a helper function?

I think `read_s3_data` could definitely be a helper function rather than `fixture -> function -> function`.

tests/aws/scenario/kinesis_firehose/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+            keys = [obj.get("Key") for obj in response.get("Contents")]
+
+            bucket_data = dict()
+            for key in keys:
+                response = s3.get_object(Bucket=bucket_name, Key=key)
+                data = response["Body"].read().decode("utf-8")
+                bucket_data[key] = data
+            return bucket_data
+
+        bucket_data = retry(_get_data, sleep=1, retries=timeout)
```
simonrw, line 27: What if new objects are created after the first successful retry iteration? Doesn't this mean that some objects will be missed?

tests/aws/scenario/kinesis_firehose/test_kinesis_firehose.py
simonrw: Ah I see it's a way to say keep trying until `assert_message_count` messages have been received. I think there are cleaner ways of expressing this than raising an exception and using the `retry` function. Perhaps a loop would be easier to understand?

```suggestion
            expected_message_count=message_count,
```

simonrw (commented):
Excited to have this initial scenario with coverage for these services. I think this is a good start, but perhaps could do with some cleaning up. 

I am glad that sensible timeouts are considered when waiting for these messages, but I think there is a simpler approach - I imagine only one use of the `retry` function should be needed.

How reliable is the delivery here? Is it a robust test? Have you experienced any flakes? It would be worth exploring running this test and trying to get missed messages.

tests/aws/scenario/kinesis_firehose/conftest.py
maxhoheiser: this simple get_data is only used if you want to fetch the first (or only) data from a bucket, if you want to fetch all or more data, you would need the get_all_messages

tests/aws/scenario/kinesis_firehose/conftest.py
maxhoheiser: refactored in: [`f19741b` (#10243)](https://github.com/localstack/localstack/pull/10243/commits/f19741b669c64eba312b9c72c79a30514c61daa4)

tests/aws/scenario/kinesis_firehose/test_kinesis_firehose.py
maxhoheiser: resolved in : [`7dfcb04` (#10243)](https://github.com/localstack/localstack/pull/10243/commits/7dfcb046774ce67a9c313af2e641fcb1367410de)

tests/aws/scenario/kinesis_firehose/conftest.py
```diff
@@ -1,47 +1,45 @@
...
+        bucket_data = read_s3_data(aws_client, bucket_name)
         messages = []
-        for input_string in s3_data_bucket.values():
+        for input_string in bucket_data.values():
             json_array_string = "[" + input_string.replace("}{", "},{") + "]"
             message = json.loads(json_array_string)
+            LOG.debug(f"Received messages: {message}")
             messages.extend(message)
-        if len(messages) != expected_message_count:
+        if not expected_message_count and len(messages) != expected_message_count:
```
simonrw, line 39:
```suggestion
        if expected_message_count is not None and len(messages) != expected_message_count:
```

?

simonrw (approved):
Looks good, excited to have service coverage in this scenario.

Just a minor suggestion about the `get_all_expected_messagages_from_s3` to correct the behaviour, and make it more useful in the future outside this scenario, but otherwise :tada:

tests/aws/scenario/kinesis_firehose/conftest.py
maxhoheiser: fixed in: [216d0b8](https://github.com/localstack/localstack/pull/10243/commits/216d0b8bad510fe8e515e89a6233facc3fbe0243)

PR #10244, opened 2024-02-14
fix S3 image rate limiting with docker login

bentsku:
## Motivation
As encountered here: https://github.com/localstack/localstack/actions/runs/7903228515/job/21570653837
We were getting rate limited by Docker Hub because we would not log in to pull the images.

Thanks to @alexrashed for the pointer!

## Changes
Add a step to login to Docker Hub in order to not be rate limited

### DIFF
--- a/.github/workflows/tests-s3-image.yml
+++ b/.github/workflows/tests-s3-image.yml
@@ -83,6 +83,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Login to Docker Hub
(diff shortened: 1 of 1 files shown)


alexrashed (approved):

dominikschubert:
Just chiming in with some additional context: This is because we're using the buildjet runner in that workflow. GitHub has its own contract with Docker and their hosted runners are **authenticated by default**, thus not usually experiencing those rate limits.

PR #10246, opened 2024-02-14
Fix Lambda handler update for UpdateFunctionConfiguration

joe4dev:
## Motivation

The operation [UpdateFunctionConfiguration](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionConfiguration.html) does not support updating the function [Handler](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionConfiguration.html#lambda-UpdateFunctionConfiguration-request-Handler) in LocalStack. This issue was reported in a support case.

## Changes

* Add a test `test_lambda_handler_update` to test updating a Lambda function handler
* Add API support for updating the function handler in the Lambda provider
* Introduce a new `id` in the version manager that can be used as a globally unique key within the assignment service to avoid key clashes for updating `$LATEST`

## Notes

In addition to the missing API operation (easy fix), I discovered a race condition in our version manager we should fix. 
Updating the function `handler` of `$LATEST` does not work properly. Invokes somehow use the old version.
The test passes when using `LAMBDA_KEEPALIVE_MS=0`, publishing a new version explicitly, or adding a sleep before the invoke (current test workaround).

The race condition also causes invalid status exceptions such as:

> localstack.services.lambda_.invocation.assignment : InvalidStatusException: Execution environment 99823878c6fbe0dbed4451c5d175b349 can only be set to status ready while running. Current status: RuntimeStatus.STOPPED

👉 The root cause is in the assignment service. It uses the `function_version.qualified_arn` as a key for managing global function versions but this ARN remains the same for updates of `$LATEST`. Therefore, new published versions work but updates for `$LATEST` don't.

## TODO

What's left to do:

- [x] Fix the race condition in `lambda_service.update_version_state` that causes invokes to be sent to an old Lambda version until the old version is stopped **asynchronously** in version_manager.stop via assignment_service.stop_environments_for_version
- [x] Remove the `time.sleep(1)` that make the test pass

### DIFF
--- a/localstack/services/lambda_/invocation/lambda_service.py
+++ b/localstack/services/lambda_/invocation/lambda_service.py
@@ -91,7 +91,7 @@ class LambdaService:
         shutdown_futures = []
         for event_manager in self.event_managers.values():
             shutdown_futures.append(self.task_executor.submit(event_manager.stop))
-        # TODO: switch shutdown order?
+        # TODO: switch shutdown order? yes, shutdown starting versions before the running versions would make more sense
         for version_manager in self.lambda_running_versions.values():
             shutdown_futures.append(self.task_executor.submit(version_manager.stop))
         for version_manager in self.lambda_starting_versions.values():
--- a/localstack/services/lambda_/invocation/assignment.py
+++ b/localstack/services/lambda_/invocation/assignment.py
@@ -28,7 +28,7 @@ class AssignmentService(OtherServiceEndpoint):
     scope: LocalStack global
     """
 
-    # function_version (fully qualified function ARN) => runtime_environment_id => runtime_environment
+    # function_version manager id => runtime_environment_id => runtime_environment
     environments: dict[str, dict[str, ExecutionEnvironment]]
 
     # Global pool for spawning and killing provisioned Lambda runtime environments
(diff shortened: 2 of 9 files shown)


localstack/services/lambda_/invocation/lambda_service.py
```diff
@@ -91,7 +91,7 @@ def stop(self) -> None:
         shutdown_futures = []
         for event_manager in self.event_managers.values():
             shutdown_futures.append(self.task_executor.submit(event_manager.stop))
-        # TODO: switch shutdown order?
+        # TODO: switch shutdown order? yes, shutdown starting versions before the running versions would make more sense
         for version_manager in self.lambda_running_versions.values():
             shutdown_futures.append(self.task_executor.submit(version_manager.stop))
         for version_manager in self.lambda_starting_versions.values():
```
joe4dev, line 97: We learned a lesson that the `qualified_arn` is not unique when dealing with updates of `$LATEST`.

Albeit unlikely, this could potentially be a problem as well in the `lambda_service` where we use the `qualified_arn` to manager starting and running versions as well as event managers. I guess concurrent updates on latest are prohibited (?), which would limit the issue.

dfangl (commented):
Since I wrote a lot of this, I would like to let @dominikschubert approve, but I like the test and the handler update changes look good to me :)

dominikschubert (approved):
:heart:  Awesome detective work and important fix! Thanks so much for tackling this :rocket: 

And yes I don't think we should support concurrent updates to $LATEST :sweat_smile:

PR #10253, opened 2024-02-15
Remove hardcoded credentials and region for unit tests

viren-nadkarni:
## Summary

Unit tests require the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_DEFAULT_REGION` environment variables set. These previously used hardcoded values. This PR changes them to use the standard env test config values. This will make things more predictable when debugging the tests.


tests/unit/conftest.py
```diff
@@ -1,11 +1,13 @@
...
 
 @pytest.fixture(autouse=True)
 def set_boto_test_credentials_and_region(monkeypatch):
     """
     Automatically sets the default credentials and region for all unit tests.
     """
-    monkeypatch.setenv("AWS_ACCESS_KEY_ID", "test")
-    monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "test")
-    monkeypatch.setenv("AWS_DEFAULT_REGION", "us-east-1")
+    monkeypatch.setenv("AWS_ACCESS_KEY_ID", constants.TEST_AWS_ACCESS_KEY_ID)
```
viren-nadkarni, line 11: To be replaced with fixtures

PR #10260, opened 2024-02-16
Implement ThreadPoolExecutor publisher pattern for EventBridge (WIP)

josefdlange:
## Motivation
The `PutEvents` method on the Event Bridge implementation blocks on waiting for targets to process events. This can make for long chains of dependent event emissions that take forever to resolve the initial `PutEvents` or even involve retries upstream, creating a sort of thundering herd issue.

## Changes
This implements a Publisher class which maintains a `ThreadPoolExecutor`. This class is used both by the `Provider` for Event Bridge and passed into a callable `PutEventsHandler` class which takes the place of the previous handler function -- both share the publisher instance owned by the `Provider` class which controls thread pool shutdown behavior via its lifecycle methods.

## Testing

Testing incoming. I wanted to get something up to get some other eyeballs on it as I am not terribly familiar with the LocalStack codebase and style.

## TODO

What's left to do:

- [ ] Test, in the general hand-wavy way to ensure the change even accomplishes my goal
- [ ] Test, in the sense that the new code has ample test coverage and follows through on the ideal that all PRs should _increase_ test coverage.

## Other notes
It's plausible that this publisher wrapper for `send_event_to_target` can be leveraged in other places where `send_event_to_target` is used, but I want to keep this tidy for the moment, plus I was told in Slack that with the incoming implementation of EB Pipes stuff that this code change may be made obsolete sooner than later.

### DIFF
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -41,6 +41,7 @@ from localstack.constants import APPLICATION_AMZ_JSON_1_1
 from localstack.http import route
 from localstack.services.edge import ROUTER
 from localstack.services.events.models import EventsStore, events_stores
+from localstack.services.events.publisher import EventTargetPublisher
 from localstack.services.events.scheduler import JobScheduler
 from localstack.services.moto import call_moto
 from localstack.services.plugins import ServiceLifecycleHook
@@ -71,7 +72,8 @@ class ValidationException(ServiceException):
 
 class EventsProvider(EventsApi, ServiceLifecycleHook):
     def __init__(self):
-        apply_patches()
+        self._publisher = EventTargetPublisher()
+        apply_patches(self._publisher)
 
     def on_after_init(self):
         ROUTER.add(self.trigger_scheduled_rule)
@@ -81,6 +83,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
 
     def on_before_stop(self):
         JobScheduler.shutdown()
+        self._publisher.shutdown()
 
     @route("/_aws/events/rules/<path:rule_arn>/trigger")
     def trigger_scheduled_rule(self, request: Request, rule_arn: str):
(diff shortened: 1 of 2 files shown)


josefdlange:

localstack/services/events/provider.py
```diff
@@ -622,16 +616,27 @@ def process_event_with_input_transformer(input_transformer: Dict, event: Dict) -
     return templated_event
 
 
-def process_events(event: Dict, targets: list[Dict]):
```
josefdlange, line 625: The diff, especially around here, is a little messy-looking but it's mainly creating class `PutEventsHandler` that is callable, the `__call__` method being the same implement as `events_handler_put_events`. This class also absorbs some functions into internal methods, mainly because they're only used and needed for the handler's sake. Implementation details should remain relatively the same, save for the executor-publisher being involved instead of directly calling `send_event_to_target`

josefdlange:
Can I get a little assistance with the failing tests here? The bootstrap fail I cannot reproduce locally.

dfangl:
Hi @josefdlange !
You are patching a method with an object. An object which implements `__call__`, but still an object. Moto inspects the responses methods before invoking them, and it has to be a function.

The lines making the check can be found here: https://github.com/getmoto/moto/blob/f9add957c9f8311f7e1aa927de7a83f2cb5adce2/moto/core/responses.py#L570-L573
The method used can be found here: https://github.com/getmoto/moto/blob/f9add957c9f8311f7e1aa927de7a83f2cb5adce2/moto/core/utils.py#L66-L68

Even though callable, your object does not pass the `inspect.isfunction` test. Therefore, `put_events` is not available as a method, and fails on every call.

I would recommend patching it with a function, and not an object.

dfangl:
Hi @josefdlange !
There are still some failing tests in our CI runs. They seem to not be aws validated, so the test failure might very well be an issue with the test more than with the implementation.

Still, we need to fix them beforehand. Could you take a stab into validating the test against AWS? You can find a guide on how to do it here: https://docs.localstack.cloud/contributing/integration-tests/#running-integration-tests-against-aws

If you do not have an AWS account available, please just tell me, and I can fix those tests for you :)

josefdlange:
@dfangl I've been on vacation the last couple weeks and otherwise a little busy 😓 I'll look into this sometime soon!

josefdlange:
I'm frankly not going to get to this. We've moved on from EB for solving our use case and so needing to improve this behavior has moved out-of-scope for me. I'm going to close this PR but anyone is welcome to recover it and work on it, of course. I just want to keep my GitHub dashboard tidy.

PR #10261, opened 2024-02-16
SNS: Fix test assertions to query to correct account and region

viren-nadkarni:
## Background

SNS provides developer endpoints for introspection and debugging. These endpoints accept the account ID and region for filtering purpose.

## Changes

This PR makes a slight fix to the test so that assertions are made in the correct account ID and region ie. the test account ID and region. Earlier running the test in non-default account ID or region would cause failures because the resources would be created in the test region but assertion would happen in the default region ie. 'us-east-1'.

## Related

Cherry picked from #8204


bentsku (approved):
LGTM! Good catch, sorry for not thinking in a multi-account manner 😬 Will make efforts on that way!

PR #10263, opened 2024-02-16
Fix architecture update for Lambda function

joe4dev:
## Motivation

[UpdateFunctionCode](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionCode.html) does not support updating the [Architectures](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionCode.html#lambda-UpdateFunctionCode-request-Architectures) (e.g., x64 => ARM).
This issue has been reported and reproduced by @cabeaulac based on customer input.

## Changes

* Adjust the mixed architecture test `test_mixed_architecture` to use `UpdateFunctionCode`
* Implement Architecture update in `UpdateFunctionCode`

## Testing

Our CI pipeline currently does not support cross-architecture emulation. Therefore, this change can only be tested locally on a machine with cross-architecture support (e.g., ARM MacBook):

1. Unskip `tests.aws.services.lambda_.test_lambda.TestLambdaBehavior.test_mixed_architecture`
2. Run `tests.aws.services.lambda_.test_lambda.TestLambdaBehavior.test_mixed_architecture`

### DIFF
--- a/localstack/services/lambda_/provider.py
+++ b/localstack/services/lambda_/provider.py
@@ -728,8 +728,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                 f"Request must be smaller than {config.LAMBDA_LIMITS_CREATE_FUNCTION_REQUEST_SIZE} bytes for the CreateFunction operation"
             )
 
-        architectures = request.get("Architectures")
-        if architectures:
+        if architectures := request.get("Architectures"):
             if len(architectures) != 1:
                 raise ValidationException(
                     f"1 validation error detected: Value '[{', '.join(architectures)}]' at 'architectures' failed to "
@@ -1136,6 +1135,21 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
 
         old_function_version = function.versions.get("$LATEST")
         replace_kwargs = {"code": code} if code else {"image": image}
+
+        if architectures := request.get("Architectures"):
+            if len(architectures) != 1:
+                raise ValidationException(
+                    f"1 validation error detected: Value '[{', '.join(architectures)}]' at 'architectures' failed to "
+                    f"satisfy constraint: Member must have length less than or equal to 1",
+                )
+            if architectures[0] not in ARCHITECTURES:
+                raise ValidationException(
+                    f"1 validation error detected: Value '[{', '.join(architectures)}]' at 'architectures' failed to "
+                    f"satisfy constraint: Member must satisfy constraint: [Member must satisfy enum value set: "
+                    f"[x86_64, arm64], Member must not be null]",
+                )
+            replace_kwargs["architectures"] = architectures
+
         config = dataclasses.replace(
             old_function_version.config,
             internal_revision=short_uid(),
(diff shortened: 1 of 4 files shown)


localstack/services/lambda_/provider.py
```diff
@@ -1136,6 +1135,21 @@ def update_function_code(
 
         old_function_version = function.versions.get("$LATEST")
         replace_kwargs = {"code": code} if code else {"image": image}
+
+        if architectures := request.get("Architectures"):
+            if len(architectures) != 1:
+                raise ValidationException(
+                    f"1 validation error detected: Value '[{', '.join(architectures)}]' at 'architectures' failed to "
+                    f"satisfy constraint: Member must have length less than or equal to 1",
+                )
```
dominikschubert, line 1144:
```suggestion
            if len(architectures) > 1:
                raise ValidationException(
                    f"1 validation error detected: Value '[{', '.join(architectures)}]' at 'architectures' failed to "
                    f"satisfy constraint: Member must have length less than or equal to 1",
                )
```

Shouldn't be an empty architectures list be fine according to the error message? (would also affect the `architectores[0]` below since then it won't be safe anymore.

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -523,65 +523,48 @@ def test_ignore_architecture(self, create_lambda_function, monkeypatch, aws_clie
...
+            "$..RuntimeVersionConfig.RuntimeVersionArn",
+        ]
+    )
     @markers.aws.validated
-    def test_mixed_architecture(self, create_lambda_function, aws_client):
-        """Test emulation and interaction of lambda functions with different architectures.
-        Limitation: only works on ARM hosts that support x86 emulation.
+    def test_mixed_architecture(self, create_lambda_function, aws_client, snapshot):
+        """Test emulation of a lambda function changing architectures.
+        Limitation: only works on hosts that support both ARM and AMD64 architectures.
```
dominikschubert, line 536: Interesting limitation. Think we should introduce a marker for this

localstack/services/lambda_/provider.py
dominikschubert: If that's not the case (bc AWS treats it as null and thus returns the error msg below), please add a comment :+1:

dominikschubert (approved):
:raised_hands: were just talking about this yesterday, great job

localstack/services/lambda_/provider.py
joe4dev: Added a comment:

```py
# An empty list of architectures is also forbidden. Further exceptions are tested here for create_function:
# tests.aws.services.lambda_.test_lambda_api.TestLambdaFunction.test_create_lambda_exceptions
```

tests/aws/services/lambda_/test_lambda.py
joe4dev: Agree that adding marker is better than the plain skip. Maybe adding an arm marker:

```
only_on_amd64 = pytest.mark.only_on_amd64
only_on_arm64 = pytest.mark.only_on_arm64
```

This special test would then have both markers.

joe4dev: It's just visually a bit tricky to distinguish:

`amd64`
`arm64`

dominikschubert: True but fortunately markers are mostly for automation and filtering purposes, so I'd say that's ok :+1:

PR #10265, opened 2024-02-17
update S3 pre-signed credentials logic

bentsku:
## Motivation
Currently in LocalStack, in order to use pre-signed URLs while validating the signature (with `S3_SKIP_SIGNATURE_VALIDATION` set to false), the user needs to set hardcoded credentials: `test`/`test`. 

This is due to the fact that we did not retrieve the Secret Access Key linked to the sent Access Key Id, which is needed to take the incoming request server side, sign it as well and compare with the received signature. 

We also used the `TEST_AWS*` credentials to generate the signature server side, which was a way for our test to always work as those are hardcoded. 

## Changes
- Implement a way to retrieve the Secret Access Key of the user by directly accessing the moto backend, as there are no public API to retrieve the secret access key for obvious security reasons
- however, as we don't create IAM users by default in LocalStack, we still need to rely on the hardcoded value of the secret access key to be `test`, so that we can properly validate pre-signed URL even the user does not provide credentials that actually exists in LocalStack. 
- update the tests to use the hardcoded S3 constants instead of the test credentials

Note: as we can use random Access Keys in LocalStack without having them being created in IAM, most of the time we won't be able to retrieve the secret key, and will need to fallback to the `test` secret access key.

However, we now support passing proper existing credentials and having the handler properly validate them. 

\cc @dfangl as I'm directly accessing IAM internals, not sure what is your opinion on this, and if it can be done any other way?

### DIFF
--- a/localstack/services/s3/presigned_url.py
+++ b/localstack/services/s3/presigned_url.py
@@ -253,7 +253,6 @@ def get_secret_access_key_from_access_key_id(access_key_id: str) -> Optional[str
     For now, we need to access Moto internals, as they are no public APIs to retrieve it for obvious reasons.
     If the AccessKey is not registered, use the default `test` value that was historically used for pre-signed URLs, in
     order to support default use cases
-    :param account_id: the account id derived from the AccessKey
     :param access_key_id: the provided AccessKeyID in the Credentials parameter
     :return: the linked secret_access_key to the access_key
     """
--- a/localstack/services/s3/presigned_url.py
+++ b/localstack/services/s3/presigned_url.py
@@ -233,20 +233,17 @@ def get_credentials_from_parameters(parameters: dict) -> PreSignedCredentials:
         if len(credential_value):
             access_key_id = credential_value[0]
 
-    secret_access_key = get_secret_access_key_from_access_key_id(access_key_id)
-
-    if not access_key_id or not secret_access_key:
-        # if we could not retrieve the secret access key, it means the access key was not registered in LocalStack,
-        # so passing the right access key id but the wrong secret access key will always end up in a failure
-        # fallback to hardcoded necessary test/test credentials then
-        secret_access_key = DEFAULT_PRE_SIGNED_SECRET_ACCESS_KEY
+    if not access_key_id:
+        # fallback to the hardcoded value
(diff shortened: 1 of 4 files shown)


localstack/services/s3/presigned_url.py
```diff
@@ -206,6 +217,55 @@ def _get_op_from_request(self, request: Request):
...
+    We need to retrieve the internal secret access key in order to also sign the request on our side to validate it
+    For now, we need to access Moto internals, as they are no public APIs to retrieve it for obvious reasons.
+    If the AccessKey is not registered, use the default `test` value that was historically used for pre-signed URLs, in
+    order to support default use cases
+    :param access_key_id: the provided AccessKeyID in the Credentials parameter
+    :return: the linked secret_access_key to the access_key
+    """
+    from moto.iam.models import AccessKey, iam_backends
+
+    account_id = get_account_id_from_access_key_id(access_key_id)
```
viren-nadkarni, line 261: Q: Is it possible to use `STS:GetSessionToken` at this point and obtain the access key ID + secret access key programatically?

bentsku: I don't think so, I think `GetSessionToken` returns a set of temporary credentials which will be then be different than what was used to signed the request on the client side.

I think you would also need to create the client for STS with the Access Key ID and the Secret Access Key of the client?

bentsku: This made me think of @dfangl hackathon, where he must had to implement a similar logic to validate the signature of every request:
https://github.com/localstack/localstack/commit/f7583642a0c78215dd742880f290411db93836f6

https://github.com/localstack/localstack/compare/master...hackathon-sig

And I guess it was the same issue, about having credentials that are set in LocalStack.

So I'm just not sure about the reverse of the access key id <-> account id to access the store, or I need to iterate in the same way.

dfangl: There seems to be no API for getting the secret key, so I think this is the best way for now :)

dfangl (approved):
Only really looked at the IAM/STS stuff, but looks good! Sadly there is no better way as far as I know :)

PR #10235, opened 2024-02-13
Update to Python 3.12

alexrashed:
## Motivation
This PR contains all changes necessary to upgrade our Python version for the runtime from 3.11 to 3.12.
Similar to previous updates:
- https://github.com/localstack/localstack/pull/8087
- https://github.com/localstack/localstack/pull/5632

## Changes
- Update all CircleCI runners [(which use Python 3.12.1 by default)](https://discuss.circleci.com/t/ubuntu-linux-vm-machine-images-2022-april-q2-update/43749).
- Update all GitHub actions to install Python 3.12 instead of Python 3.11.
- Update the dockerfile base image to the latest sha digest of Python 3.12.2.
- Update all references to virtual environments which contain the Python version (f.e. `/usr/lib/localstack/python-packages/lib/python3.12/site-packages`).
- Update the Lambda Runtime version used for local execution (in the runtime of LocalStack itself) from 3.11 to 3.12.
Impact
- Update the pinned dependencies (they need to be generated with the right version).
- Update the `.python-version` file for `pyenv`

This PR only updates the runtime version of LocalStack itself. The LocalStack CLI will still be compatible with Python 3.8+.
However, [Python Initialization Hooks](https://docs.localstack.cloud/references/init-hooks/) and [LocalStack Extensions](https://docs.localstack.cloud/references/localstack-extensions/) will then run in a Python 3.12 environment instead of Python 3.11.

## TODO
- [x] js2py is currently not compatible with Python 3.12.
  - https://github.com/PiotrDabkowski/Js2Py/issues/317
  - https://github.com/PiotrDabkowski/Js2Py/pull/327
- [ ] Verify that the tests are all green / all libraries and our code are compatible.
- [ ] Verify that this change is compatible with our downstream projects.

PR #10266, opened 2024-02-18
upgrade plux to 1.7

thrau:
## Motivation

This PR upgrade plux to 1.7 which provides python 3.12 support (see #10235). The PR also migrates `plugin` imports to the new way to do things (`from plux import ...`). Since 1.7 is backwards compatible, we can make these changes for ext separately.
Moreover, plux now has a way of regenerating entrypoints without invoking setup.py (which is deprecated), therefore unblocking us from moving away from a `setup.py` to `python -m build`.

We could also look into removing `localstack.cli.plugins`, and move some of the functionality to plux (if needed).

## Changes

* localstack now uses plux>=1.7 (also adds a constraint for <2 where restructuring is expected)
* `make entrypoints` now uses the plux frontend `python -m plux entrypoints` instead of `setup.py plugins egg_info`
* localstack now favors `from plux` for imports instead of `from plugin`

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -1116,7 +1116,7 @@ def configure_container(container: Container):
     else:
         container.config.additional_flags = f"{container.config.additional_flags} {user_flags}"
 
-    # get additional parameters from plugins
+    # get additional parameters from pluxs
     hooks.configure_localstack_container.run(container)
 
     if config.DEVELOP:
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -1116,7 +1116,7 @@ def configure_container(container: Container):
     else:
         container.config.additional_flags = f"{container.config.additional_flags} {user_flags}"
 
-    # get additional parameters from pluxs
+    # get additional parameters from plux
     hooks.configure_localstack_container.run(container)
 
     if config.DEVELOP:
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ install_requires =
     dill==0.3.6
     dnslib>=0.9.10
     dnspython>=1.16.0
-    plux>=1.7,<2.0
+    plux>=1.7
     psutil>=5.4.8
     python-dotenv>=0.19.1
     pyyaml>=5.1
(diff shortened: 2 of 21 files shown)


localstack/utils/bootstrap.py
```diff
@@ -1116,7 +1116,7 @@ def configure_container(container: Container):
     else:
         container.config.additional_flags = f"{container.config.additional_flags} {user_flags}"
 
-    # get additional parameters from plugins
+    # get additional parameters from pluxs
```
alexrashed, line 1119: Looks like a search-replace issue here 😄 
```suggestion
    # get additional parameters from plux
```

setup.cfg
```diff
@@ -31,7 +31,7 @@ install_requires =
     dill==0.3.6
     dnslib>=0.9.10
     dnspython>=1.16.0
-    plux>=1.3.1
+    plux>=1.7,<2.0
```
alexrashed, line 34: I think this is a good opportunity to discuss upper version limits.
We explicitly removed pins like this in https://github.com/localstack/localstack/pull/10195.
We did this because there's no need for the pins for as long as the currently released version is compatible.
New releases will either be picked up proactively by us (in PRs like this, explicitly upgrading a single dependency), or they are picked up by the weekly upgrades where the breaking changes are identified and (a) fixed in the PR by adding a small fix, or (b) pinned and explicitly upgraded in a follow up PR (where the min version is updated and the upper version limit removed).
In this case, the library is maintained by us, meaning that we already explicitly know that `2.0` will contain breaking changes. However, if the person who creates the release does not proactively update it here, we most likely won't find out (because the automated updates won't dare to touch it 😛)...
This is why I am leaning towards not having these pins, but again, this is a great opportunity to discuss them. 😄

alexrashed (commented):
The changes are looking good! The release implements a nice deprecation path, and can seamlessly be upgraded. 🔝
I only found a small nitpick, and I think it would be a good opportunity to discuss upper version limits in the `setup.cfg`.

alexrashed (approved):
Sorry for the initial "comment" instead of the approval, that was a mistake.
I am happy with merging this as it is, but I would be happy to discuss the comment concerning the upper version limit. 😄

alexrashed:
FYI: Due to the nice deprecation path / backwards compatibility of the changes in `plux==1.7.0`, the requirements lock files have already been updated with https://github.com/localstack/localstack/pull/10270. This PR will basically just contain the code changes and the change in the `setup.cfg`.

Issue #10233, opened 2024-02-13
bug: AWS_REGION overwritten with us-east-1 when using docker compose


carole-lavillonniere:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I pass `AWS_REGION=eu-central-1` to the localstack container in the environment section but it is overwritten with `us-east-1`.

### Expected Behavior

`AWS_REGION` should be `eu-central-1`

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

This is my localstack container in the docker compose file:

```
  mock-aws:
    image: localstack/localstack-pro:3.1
    ports:
      - "4566:4566"
    volumes:
      - type: bind
        source: ../integration-tests/aws
        target: /etc/localstack/init/ready.d
        read_only: true
      # Path to local Docker UNIX domain socket
      - "/var/run/docker.sock:/var/run/docker.sock"
    environment:
      - LS_LOG=debug
      - AWS_ACCESS_KEY_ID=test
      - AWS_SECRET_ACCESS_KEY=test
      - AWS_REGION=eu-central-1
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN:- }
      - EMAIL_SENDER_ADDRESS=noreply@host.com

```

I mount a folder which contains bash script to be run to initialize the localstack services.
Here is for example the content of `integration-tests/aws/ses.sh`:

```
#!/bin/bash

awslocal ses verify-email-identity \
  --email-address ${EMAIL_SENDER_ADDRESS} \
  --region ${AWS_REGION}

awslocal ses list-identities \
  --region ${AWS_REGION}

```

The script is executed but `AWS_REGION` seems to be overwritten and now has the value `us-east-1`.
Please note that the other variable called `EMAIL_SENDER_ADDRESS` has the proper value.

See output:
```
local-dev-mock-aws-1  | + awslocal ses verify-email-identity --email-address noreply@hosy.com --region us-east-1
local-dev-mock-aws-1  | 2024-02-13T14:17:04.827  INFO --- [   asgi_gw_1] localstack.request.aws     : AWS ses.VerifyEmailIdentity => 200
local-dev-mock-aws-1  | + awslocal ses list-identities --region us-east-1
local-dev-mock-aws-1  | 2024-02-13T14:17:06.026  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS ses.ListIdentities => 200
[...]

PR #10272, opened 2024-02-19
Remove implicit credentials for Boto to fall back on

viren-nadkarni:
## Background

LocalStack currently sets (overrides) the Boto credentials and region during startup. The original intention was to handle cross-account usecases but since then, cross-account/region has been reworked to use explicit credentials obtained via request context.

Technically, the fallback from the environment is no longer required. Furthermore, by removing this fallback, we force future contributions use request context credentials programmatically. Same for external components that may make use of these env vars (such as QLDB). 

## Changes

- Removes access key ID and secret access key being set at startup.
- Removes some dead code.
- Pass proper test credentials to `test_terraform.py`. This went undetected because `terraform` would use the env values set by LocalStack startup.
- Maintain backward compatibility with init hooks. Upcoming major release will make init hook scripts require explicit credentials

## Related

Closes https://github.com/localstack/localstack/issues/10233


localstack/services/infra.py
```diff
@@ -308,10 +293,7 @@ def prepare_environment():
 
         # set environment
         os.environ["AWS_REGION"] = AWS_REGION_US_EAST_1
```
viren-nadkarni, line 295: Remove this too

localstack/constants.py
```diff
@@ -5,12 +5,6 @@
 # LocalStack version
 VERSION = localstack.__version__
 
-# constant to represent the "local" region, i.e., local machine
-REGION_LOCAL = "local"
-
-# dev environment
-ENV_DEV = "dev"
-
```
viren-nadkarni, line 13: The Environment logic was removed in https://github.com/localstack/localstack/pull/9644

localstack/constants.py
```diff
@@ -5,12 +5,6 @@
 # LocalStack version
 VERSION = localstack.__version__
 
-# constant to represent the "local" region, i.e., local machine
-REGION_LOCAL = "local"
-
```
viren-nadkarni, line 10: 'REGION_LOCAL' logic was removed in https://github.com/localstack/localstack/pull/9750

thrau (commented):
> Technically, the fallback from the environment is no longer required.

does this imply that there are some other situations where it may be required for UX purposes? i don't think i understand the implications well enough to give a meaningful review.

maybe @dfangl can also take a peek

dfangl:
I agree that we could avoid setting it now. If the tests pass, it is (manually) specified everywhere. However, this affects child processes which inherit the parents (LocalStacks) environment. I think for this change we should execute -ext tests before merging them, as it could have an unexpected repercussions.

Another thing this might influence are init scripts - for shell scripts it inherits the environment, and for python code it is executed in-process anyway. E.g. using the aws cli will probably break by default. For those it could make sense to continue setting them - otherwise it is probably a breaking change.

viren-nadkarni:
@dfangl ext is tested working, see localstack/localstack-ext#2639

dfangl:
Ah yes, somehow missed it!

alexrashed (commented):
I agree with @dfangl. I had a quick look and to the best of my knowledge, this change is looking good for as we can make sure that we don't break any init scripts.
@viren-nadkarni Can you check the implication of this change on the different types of [init hooks](https://docs.localstack.cloud/references/init-hooks/) and how we might be able to mitigate them?
The code is basically all here: https://github.com/localstack/localstack/blob/master/localstack/runtime/init.py

viren-nadkarni:
Thanks for the pointers!

With 04c3c589faa54100c4791a5b382c38fd6ef927f7, the init scripts will have access to the config vars as now. I think it makes sense to make the credentials a requirement within the init scripts in the next major release. I've marked this as so.

alexrashed (approved):
Thanks for addressing the comments! The change looks good to me now, let's make sure to remove the vars for the init hooks with the next major release. 🧹

Issue #10169, opened 2024-02-04
bug: Unable to update stack "RemovePermission operation: No policy is associated with the given resource."


pigfrown:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Updating a CDK Stack containing API Gateway + Lambda results in :

`An error occurred (ResourceNotFoundException) when calling the RemovePermission operation: No policy is associated
with the given resource.`

(this happens even when the stack has not changed)

### Expected Behavior

should be able to update the stack after initial deployment

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

CDK Stack: https://github.com/pigfrown/cdklocal-bug/blob/main/lib/cdklocal-bug-stack.ts

Recreate with:
```
git clone https://github.com/pigfrown/cdklocal-bug
cd cdklocal-bug
npm i

localstack start -d

npx cdklocal bootstrap
npx cdklocal deploy # This deployment works/curling the endpoint works
npx cdklocal deploy # This deployment fails and breaks the deployed stack
```

Error message is: 
```
npx cdklocal deploy

✨  Synthesis time: 2.54s

CdklocalBugStack: deploying... [1/1]
CdklocalBugStack: creating CloudFormation changeset...
11:17:50 | UPDATE_FAILED        | AWS::Lambda::Permission     | MyApihelloGETApiPe...C2GEThelloC0249EEE
An error occurred (ResourceNotFoundException) when calling the RemovePermission operation: No policy is associated
with the given resource.


 ❌  CdklocalBugStack failed: Error: The stack named CdklocalBugStack failed to deploy: UPDATE_FAILED (An error occurred (ResourceNotFoundException) when calling the RemovePermission operation: No policy is associated with the given resource.): An error occurred (ResourceNotFoundException) when calling the RemovePermission operation: No policy is associated with the given resource.
    at FullCloudFormationDeployment.monitorDeployment (/tmp/cdklocal-bug/node_modules/aws-cdk/lib/index.js:428:10615)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.deployStack2 [as deployStack] (/tmp/cdklocal-bug/node_modules/aws-cdk/lib/index.js:431:182305)
    at async /tmp/cdklocal-bug/node_modules/aws-cdk/lib/index.js:431:164665

[...]

Issue #10042, opened 2024-01-10
bug: Unable to update API gateway proxy lambda using CDK


jamesphilipps:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using CDK (cdklocal command) to create a simple API with a proxy integration to a lambda handler, the lambda is successfully created on the first run of CDK. On the second run, the update fails with the following message:

> failed: Error: The stack named example-stack failed to deploy: UPDATE_FAILED (An error occurred (BadRequestException) when calling the PutIntegration operation: Integrations of type 'AWS_PROXY' currently only supports Lambda function and Firehose stream invocations.): An error occurred (BadRequestException) when calling the PutIntegration operation: Integrations of type 'AWS_PROXY' currently only supports Lambda function and Firehose stream invocations.

> Deployment failed: Error: The stack named example-stack failed to deploy: UPDATE_FAILED (An error occurred (BadRequestException) when calling the PutIntegration operation: Integrations of type 'AWS_PROXY' currently only supports Lambda function and Firehose stream invocations.): An error occurred (BadRequestException) when calling the PutIntegration operation: Integrations of type 'AWS_PROXY' currently only supports Lambda function and Firehose stream invocations.

### Expected Behavior

The lambda is updated and the deployment completes without error

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

A full reproduction of the issue can be found at: https://github.com/jamesphilipps/localstack-cdk-error-example

### Environment

```markdown
- OS:WSL Ubuntu Linux with Windows Docker integration
- LocalStack Pro

ENV LOCALSTACK_BUILD_DATE=2024-01-09
ENV LOCALSTACK_BUILD_GIT_HASH=2028820
ENV LOCALSTACK_BUILD_VERSION=3.0.3.dev
```

### Anything else?

A full reproduction of the issue can be found at: https://github.com/jamesphilipps/localstack-cdk-error-example

Yehonal:
I'm facing the same issue, when I change my lambda code and re-deploy

Yehonal:
I'm wondering how someone could use Localstack when having an API with query parameters is one of the most common situations when you work on web applications.

Am I doing something wrong? Is there an alternative to it?

Yehonal:
As a workaround for now I'm doing a double deployment locally, the first deployment without the API integration and the second deployment with the API integration. In this way, it removes the integration first (if any) and re-add it later.
It works but it takes double time to be deployed obviously

Issue #9895, opened 2023-12-17
bug: error when re-deploying a lambda function


gabrandalisse:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I start localstack for the first time and I do a` sls deploy --local` and everything goes ok. But when I do a change in my lambda function and I do `sls deploy --local` again it throw the following error:

```
Error:
DELETE_FAILED: ApiGatewayDeployment1702839928281 (AWS::ApiGateway::Deployment)
An error occurred (BadRequestException) when calling the DeleteDeployment operation: Active stages pointing to this deployment must be moved or deleted
```

### Expected Behavior

To re deploy the lambda function correctly. 

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

1. Run localstack start 
2. Do a sls deploy --local
3. Do a change in a lambda function
4. Do a sls deploy --local again
5. Should see the error

### Environment

```markdown
- OS: macOS (Sonoma 14.1.2)
- LocalStack: 3.0.2
```

mdeeter:
following....

valeryan:
Seeing this same issue:
```
Error:
DELETE_FAILED: ApiGatewayDeployment1705955950755 (AWS::ApiGateway::Deployment)
An error occurred (BadRequestException) when calling the DeleteDeployment operation: Active stages pointing to this deployment must be moved or deleted
```

Here is what is in the docker log:
```
2024-01-22 14:41:16   File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1177, in _run
2024-01-22 14:41:16     self.do_apply_changes_in_loop(changes, stack)
2024-01-22 14:41:16   File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1251, in do_apply_changes_in_loop
2024-01-22 14:41:16     self.apply_change(change, stack=stack)
2024-01-22 14:41:16   File "/opt/code/localstack/localstack/services/cloudformation/engine/template_deployer.py", line 1366, in apply_change
2024-01-22 14:41:16     progress_event = executor.deploy_loop(resource_provider_payload)  # noqa
2024-01-22 14:41:16                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-01-22 14:41:16   File "/opt/code/localstack/localstack/services/cloudformation/resource_provider.py", line 719, in deploy_loop
[...]

PR #10273, opened 2024-02-19
Fix issues in CFn resources and filter notfound exceptions for delete operations

dominikschubert:
## Motivation

Users reported simple updates with serverless/SAM/CDK failing, usually due to issues in individual resource providers, especially for delete operations caused by updates.

Fixes https://github.com/localstack/localstack/issues/9895
Fixes https://github.com/localstack/localstack/issues/10042
Fixes https://github.com/localstack/localstack/issues/10169

## Changes

- `"AWS::ApiGateway::Method"`: Fixed a few issues related to parameter transformation and passing
- `"AWS::ApiGateway::Deployment"`: Delete any stages that are still linked to this deployment before deleting the deployment itself
- `AWS::Lambda::Function`: Added a waiter after updating the function code. This was leading to intermittent 500 exceptions and retrying the handler. Didn't usually affect stack deployments but lead to some weird errors in the logs.
- Added filter for any "notfound" exceptions when trying to call the delete handler of a resource
- Added a regression test for this that simulates what the serverless deploy steps shown below would do

## Testing

```bash
# a serverless deploy is actually a CreateStack AND an UpdateStack operation
serverless deploy --stage=local
# change lambda handler name for example
serverless deploy --stage=local
```


tests/aws/services/cloudformation/resources/test_apigateway.validation.json
```diff
@@ -1,4 +1,13 @@
 {
+  "tests/aws/services/cloudformation/resources/test_apigateway.py::TestServerlessApigwLambda::test_serverless_like_deployment_with_update": {
+    "last_validated_date": "2024-02-19T08:55:12+00:00"
+  },
+  "tests/aws/services/cloudformation/resources/test_apigateway.py::TestServerlessApigwLambda::test_stuff": {
+    "last_validated_date": "2024-02-16T14:51:15+00:00"
+  },
+  "tests/aws/services/cloudformation/resources/test_apigateway.py::TestServerlessApigwLambda::test_update_trigger": {
+    "last_validated_date": "2024-02-16T22:21:12+00:00"
+  },
```
dominikschubert, line 10:
```suggestion
```

localstack/services/apigateway/resource_providers/aws_apigateway_method.py
```diff
@@ -104,7 +105,7 @@ def create(
         # key_to_lower makes in-place changes which will cause crash
         # when we try to use model outside of this function
         # for example generating composite physical id
-        params = keys_to_lower(model.copy())
+        params = keys_to_lower(deepcopy(model))
```
dominikschubert, line 108: Long term we should fix this by using the provider utils and not keys_to_lower.

dominikschubert:
Failures seem to be caused by the broken deletes due to us basically ignoring dependencies for deletes/replacements. Will need to look into it. We should probably try to ignore any deletes during stack updates for now :man_shrugging:

pinzon (approved):

Yehonal:
Thank you! We can't wait to test it 
Do you know when the localstack-pro image will be updated on the Docker Hub?
I've seen the `localstack` one has been already updated

PR #10274, opened 2024-02-19
Sqs fifo same dedup different message groups

baermat:
## Motivation
While investigating a support case, I stumbled upon this behaviour which is not yet implemented. If a message uses the same message deduplication ID while the fifo queue is set to "[high throughput mode](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/high-throughput-fifo.html)", the deduplication id seems to be evaluated at a group level and not at the queue level. The effects on ordering still need to be investigated once we implement this feature (it is unclear to me why I would use this high throughput fifo queue over a regular queue).

## Changes
- adds 1 test each for:
  - Same deduplication id with different message groups, normal mode
  - Same deduplication id with different message groups, high throughput mode

### DIFF
--- a/tests/aws/services/sqs/test_sqs.py
+++ b/tests/aws/services/sqs/test_sqs.py
@@ -3686,6 +3686,35 @@ class TestSqsProvider:
         )[0].get("MD5OfBody")
 
     @markers.aws.validated
+    def test_sqs_fifo_same_dedup_id_different_message_groups(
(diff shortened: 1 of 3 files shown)


thrau (approved):
that is excellent, really nice find @baermat! thanks for adding the test, always great to have keep an xfailed test around for a TODO :100:

PR #10283, opened 2024-02-20
Save session tags when assuming a role

dfangl:
## Motivation
There is no API for retrieving session attributes once the session is created. Since we need the session tags for IAM evaluation, we need to store them.

We store them by access key id instead of session ARN, to avoid collisions with identical SessionNames (those sessions still preserve the right session tags).

## Changes
* Allow `create_role` fixture users to pass additional variables
* Save session tags based on access key id.

### DIFF
--- a/localstack/services/sts/models.py
+++ b/localstack/services/sts/models.py
@@ -1,0 +1,9 @@
+from localstack.services.stores import AccountRegionBundle, BaseStore, CrossRegionAttribute
+
+
+class STSStore(BaseStore):
+    # maps access key ids to tags for the session they belong to
+    session_tags: dict[str, dict[str, str]] = CrossRegionAttribute(default=dict)
+
+
+sts_stores = AccountRegionBundle("sts", STSStore)
--- a/localstack/services/sts/provider.py
+++ b/localstack/services/sts/provider.py
@@ -21,6 +21,7 @@ from localstack.aws.api.sts import (
 from localstack.services.moto import call_moto
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.services.sts.models import sts_stores
+from localstack.utils.aws.arns import extract_account_id_from_arn
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


localstack/services/sts/models.py
```diff
@@ -0,0 +1,9 @@
+from localstack.services.stores import AccountRegionBundle, BaseStore, CrossRegionAttribute
+
+
+class STSStore(BaseStore):
+    # maps access key ids to tags for the session they belong to
+    session_tags: dict[str, dict[str, str]] = CrossRegionAttribute(default=dict)
```
dominikschubert, line 6: Creating our own store is probably a good hint that we might want to internalize STS

localstack/services/sts/provider.py
```diff
@@ -14,3 +31,29 @@ def get_caller_identity(self, context: RequestContext, **kwargs) -> GetCallerIde
...
+        token_code: tokenCodeType = None,
+        source_identity: sourceIdentityType = None,
+        provided_contexts: ProvidedContextsListType = None,
+        **kwargs,
+    ) -> AssumeRoleResponse:
+        response: AssumeRoleResponse = call_moto(context)
+
+        if tags:
+            transformed_tags = {tag["Key"]: tag["Value"] for tag in tags}
+            store = sts_stores[context.account_id][context.region]
```
dominikschubert, line 56: If I understand this correctly, it means this feature is not cross-account compatible for now, since we should save in the target account/region (based on the ARN of the role).

Assume_role is probably one of the most used operations in the context of cross-account usage :thinking:

localstack/services/sts/models.py
dfangl: Yeah, would love that level of control to it, currently a lot of the operations are mocked.

dfangl: This will at some point be extended anyway, as we need to include session policies as well.

localstack/services/sts/provider.py
```diff
@@ -14,3 +32,31 @@ def get_caller_identity(self, context: RequestContext, **kwargs) -> GetCallerIde
...
+        provided_contexts: ProvidedContextsListType = None,
+        **kwargs,
+    ) -> AssumeRoleResponse:
+        response: AssumeRoleResponse = call_moto(context)
+
+        if tags:
+            transformed_tags = {tag["Key"]: tag["Value"] for tag in tags}
+            # we should save it in the store of the role account, not the requester
+            account_id = extract_account_id_from_arn(role_arn)
+            store = sts_stores[account_id]["us-east-1"]
```
dominikschubert, line 59: nit: Might be good to add comment explaining why this is statically set to `us-east-1`. Can be confusing, especially given the effort we're putting into being multi-region everywhere :grin:

dominikschubert (approved):
LGTM, thanks for fixing the account id :+1:

PR #10288, opened 2024-02-20
HTTP IntegrationSubtype support for Kinesis-PutRecord 

calvernaz:
## Motivation

Adds support to Kinesis-PutRecord.

## Changes

Adds Kinesis backend integration support for integration subtype operation "Kinesis-PutRecord"

## Testing

Snapshot parity testing

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -469,7 +469,7 @@ class KinesisIntegration(BackendIntegration):
         integration_type = integration_type_orig.upper()
         uri = integration.get("uri") or integration.get("integrationUri") or ""
 
-        if uri.endswith("kinesis:action/PutRecord"):
+        if uri.endswith("kinesis:action/PutRecord") or integration.get('integrationSubtype') == "Kinesis-PutRecord":
             target = "Kinesis_20131202.PutRecord"
         elif uri.endswith("kinesis:action/PutRecords"):
             target = "Kinesis_20131202.PutRecords"
@@ -483,7 +483,8 @@ class KinesisIntegration(BackendIntegration):
 
         try:
             # xXx this "event" request context is used in multiple places, we probably
-            # want to refactor this into a model class
+            # want to refactor this into a model class.
+            # I'd argue
             invocation_context.context = helpers.get_event_request_context(invocation_context)
             invocation_context.stage_variables = helpers.get_stage_variables(invocation_context)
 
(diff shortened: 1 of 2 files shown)


localstack/services/apigateway/integration.py
```diff
@@ -518,6 +526,69 @@ def invoke(self, invocation_context: ApiInvocationContext):
...
+                request_parameters.get("ExplicitHashKey"), variables
+            )
+        if "SequenceNumberForOrdering" in request_parameters:
+            payload["SequenceNumberForOrdering"] = VtlTemplate().render_vtl(
+                request_parameters.get("SequenceNumberForOrdering"), variables
+            )
+        if "Region" in request_parameters:
+            payload["Region"] = VtlTemplate().render_vtl(
+                request_parameters.get("Region"), variables
+            )
```
bentsku, line 589: I think we should not pass the `Region` to the Kinesis payload because it's not a supported parameter for `PutRecord` itself, and the Kinesis service will raise an exception.

I know we still need to render it though, so maybe we need to pop it out of the dict after, to properly pass it to the headers creation? Or we don't treat it for now and leaves it as a TODO and a log message indicating it's not supported yet, as creating a multi-region test for it might a bit tricky as this PR is ready otherwise. 

In any case, we should not pass the `Region` parameter directly to the HTTP call.

bentsku (commented):
Awesome, thanks for the quick turnaround on this one! Nice to see that the KinesisIntegration did not need too many changes. 

I just have one comment around the `Region` parameter that will I believe trigger an exception in Kinesis if passed directly in the payload. Once this is resolved, this is good to go for me! Thanks a lot!

localstack/services/apigateway/integration.py
calvernaz: Localstack doesn't support it, you mean? Because is in the docs https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-aws-services-reference.html#Kinesis-PutRecord

bentsku: LocalStack do support it, but it's not a parameter of PutRecord: https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html#API_PutRecord_RequestParameters

It must mean that API Gateway maps this parameter differently, it does not pass it directly as a parameter to PutRecord but direct it differently. It's also a bit weird as the region should be part of the StreamARN. I think it's fine to leave it as a TODO for now, we can tackle it if we get a report about it. 

Sorry for comment, but at least we avoid the issue if it happens. Thanks for addressing it!

bentsku (approved):
LGTM! Thanks for removing the region parameter. Thanks again for the quick turnaround!

localstack/services/apigateway/integration.py
calvernaz: Because you can use `StreamName` and not `StreamArn` maybe the region plays a role on that. I don't know.

PR #10297, opened 2024-02-21
Add support for CFn dot attributes that conflict with schema

dominikschubert:
## Motivation
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-cachecluster.html#aws-resource-elasticache-cachecluster-return-values

For -ext we need to support a case where the schema has a conflict between the specified readOnlyProperties, i.e. normally the attributes and the actual property spec. 

Example:  a `!GetAtt Cluster.RedisEndpoint.Address` will look up `RedisEndpointAddress` on the resource

## Changes

- Try to access attribute/property by removing `.` from the attribute name

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -97,9 +97,16 @@ def get_attr_from_model_instance(
 
     attribute_candidate = properties.get(attribute_name)
     if "." in attribute_name:
+        # in case we explicitly add a property with a dot, e.g. resource["Properties"]["Endpoint.Port"]
+        # this is only really used for legacy cases since this isn't supported in the propertiers part of the schema
         if attribute_candidate:
-            # in case we explicitly add a property with a dot, e.g. resource["Properties"]["Endpoint.Port"]
             return attribute_candidate
+
+        # some resources (e.g. ElastiCache) have their readOnly attributes defined as Aa.Bb but the property is named AaBb
+        if attribute_candidate := properties.get(attribute_name.replace(".", "")):
+            return attribute_candidate
+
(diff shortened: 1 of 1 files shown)


localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -97,9 +97,16 @@ def get_attr_from_model_instance(
...
+        # in case we explicitly add a property with a dot, e.g. resource["Properties"]["Endpoint.Port"]
+        # this is only really used for legacy cases since this isn't supported in the propertiers part of the schema
         if attribute_candidate:
-            # in case we explicitly add a property with a dot, e.g. resource["Properties"]["Endpoint.Port"]
             return attribute_candidate
+
+        # some resources (e.g. ElastiCache) have their readOnly attributes defined as Aa.Bb but the property is named AaBb
+        if attribute_candidate := properties.get(attribute_name.replace(".", "")):
+            return attribute_candidate
+
```
simonrw, line 108: Is this something better fixed in the provider? I can see centralising the logic here means that provider writers do not have to be aware of CFn internals, but this might become difficult to maintain in the future.

simonrw (approved):
Seems like a reasonable change for now, but I have a comment about the location for this fix.

Morijarti (approved):

localstack/services/cloudformation/engine/template_deployer.py
dominikschubert: In general, like the quirks, I agree we should move  them to the provider in the future where they would just be ignored on AWS. For now handling schema inconsistencies is done in the Service to stay consistent with the rest of the quirks.

PR #10298, opened 2024-02-22
Change schedule for dependency pin upgrade

silv-io:
## Motivation

Upgrading the botocore dependencies and the regular dependencies might run into conflicts at some point. Therefore we should only upgrade the other dependencies after the botocore dependency upgrades are already merged.

## Changes

This PR changes the schedule so that the upgrade of the pinned dependencies happens on Tuesday morning instead of Monday morning, so that we have time to merge the botocore upgrades before the PR for the other dependencies is created.

### DIFF
--- a/.github/workflows/upgrade-python-dependencies.yml
+++ b/.github/workflows/upgrade-python-dependencies.yml
@@ -2,7 +2,7 @@ name: Upgrade Pinned Python Dependencies
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Nice change to have a smooth schedule, updating the dependencies after the ASF update handled the upgrade of the `botocore` pin. 🧹 ⌚

PR #10304, opened 2024-02-22
Remove legacy CFn models

dominikschubert:
## Motivation

It should be safe to delete the legacy CFn models, now that we've finished migrating all the resources to newer resource providers.
We've been driving the migration in a way where we actually always fixed the newer providers and never communicated any fallback. 

## Changes
- Remove legacy GenericBaseModel variants
- Move EC2 utils to vpc resource provider

## Follow-up
- [ ] Remove unused config to switch providers


Morijarti (approved):
Really great to see all models gone and even better pipelines green with them removed.

I was a bit worried that we missed something

PR #10306, opened 2024-02-23
Allow extension hooks to modify non-Boto env vars

viren-nadkarni:
## Background

Follow up to https://github.com/localstack/localstack/pull/10272#discussion_r1500576710

## Changes

After an init hook has run, refrain from clearing the environment and doing a full restore. Instead only restore the Boto credentials and region if it has been overridden.

### DIFF
--- a/localstack/runtime/init.py
+++ b/localstack/runtime/init.py
@@ -152,8 +152,12 @@ class InitScriptManager:
                 else:
                     script.state = State.SUCCESSFUL
                 finally:
-                    os.environ.clear()
-                    os.environ.update(env_original)
+                    # Restore original state of Boto credentials.
+                    for env_var in ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_REGION"):
+                        if env_var in env_original:
+                            os.environ[env_var] = env_original[env_var]
+                        else:
+                            os.environ.pop(env_var, None)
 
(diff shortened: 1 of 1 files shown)


localstack/runtime/init.py
```diff
@@ -152,8 +152,12 @@ def run_stage(self, stage: Stage) -> List[Script]:
                 else:
                     script.state = State.SUCCESSFUL
                 finally:
-                    os.environ.clear()
-                    os.environ.update(env_original)
+                    # Restore original state of Boto credentials.
+                    for env_var in ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_REGION"):
+                        if env_var in env_original:
+                            os.environ[env_var] = env_original[env_var]
```
viren-nadkarni, line 158: I'm wondering whether a modification to the value made by the init hook should be reset or retained?

dfangl (approved):
Thanks for addressing it so quickly! I like this version way better - we might be overriding those variables still if set in a python script - but I honestly think this is fine, users should not expect to set this internal configuration like this.

LGTM, great work getting rid of all the legacy stuff!

localstack/runtime/init.py
dfangl: I think reset is honestly fine, as you might just use them for the operations inside the init hook script.

Issue #10310, opened 2024-02-23
bug: API Gateway does not allow lambda integrations to return JSON without a body


Garethp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If I attempt to execute a RestAPI Integration whose lambda returns a response with no body, LocalStack returns a `{"message": "Internal server error"}` with the following error in the docker-compose logs:

```
localstack-bugs | 2024-02-23T17:17:54.343 DEBUG --- [   asgi_gw_0] l.s.l.i.version_manager    : [gateway-stack-lambdahandler909F9205-bc4b1fd1-8bd32ae7-88c8-48f2-a494-0d7410e2244c] END RequestId: 8bd32ae7-88c8-48f2-a494-0d7410e2244c
localstack-bugs | 2024-02-23T17:17:54.343 DEBUG --- [   asgi_gw_0] l.s.l.i.version_manager    : [gateway-stack-lambdahandler909F9205-bc4b1fd1-8bd32ae7-88c8-48f2-a494-0d7410e2244c] REPORT RequestId: 8bd32ae7-88c8-48f2-a494-0d7410e2244c      Duration: 1.79 ms       Billed Duration: 2 ms   Memory Size: 128 MB Max Memory Used: 128 MB 
localstack-bugs | 2024-02-23T17:17:54.343 DEBUG --- [   asgi_gw_0] l.s.lambda_.provider       : Lambda invocation duration: 921.40ms
localstack-bugs | 2024-02-23T17:17:54.345  WARN --- [   asgi_gw_1] l.s.apigateway.integration : Lambda output should follow the next JSON format: { "isBase64Encoded": true|false, "statusCode": httpStatusCode, "headers": { "headerName": "headerValue", ... },"body": "..."}
localstack-bugs | 2024-02-23T17:17:54.345  INFO --- [   asgi_gw_1] localstack.request.http    : GET /prod/ => 502
```

### Expected Behavior

In AWS, doing the same will result in a successful response with no body data instead of returning an error

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```ts
export class GatewayStack extends Stack {
  constructor(scope: Construct, id: string, props: StackProps) {
    super(scope, id, props);

[...]

PR #10312, opened 2024-02-23
update validation of Lambda AWS Proxy format

bentsku:
## Motivation
As reported in #10310, it seems we were overzealous over the validation of the output of the AWS Proxy lambda integration. 
We need to validate that no wrong fields are returned, but no field is mandatory. 

The documentation was not clear on that:
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

## Changes
- added a test to validate different formats of return of the lambda function
- updated the logic, not sure why it's duplicated in 2 spots but made it in both for now to be sure. We can refactor this in the future


calvernaz (approved):
Interesting :)

Issue #10309, opened 2024-02-23
bug: S3 Provider V3 fails on Presigned Uploads


Garethp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With the new S3 Provider that Localstack V3 uses, when I try to upload a file in node using presigned S3 URLs, I receive the following error in the docker logs, but the request itself succeeds. No file is found in the bucket afterwards:

```
localstack-bugs | 2024-02-23T15:50:21.627 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain
localstack-bugs | Traceback (most recent call last):
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
localstack-bugs |     handler(self, self.context, response)
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 112, in __call__
localstack-bugs |     handler(chain, context, response)
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 82, in __call__
localstack-bugs |     skeleton_response = self.skeleton.invoke(context)
localstack-bugs |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 154, in invoke
localstack-bugs |     return self.dispatch_request(serializer, context, instance)
localstack-bugs |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 168, in dispatch_request
localstack-bugs |     result = handler(context, instance) or {}
localstack-bugs |              ^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack-bugs |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 118, in __call__
localstack-bugs |     return self.fn(*args, **kwargs)
localstack-bugs |            ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10314, opened 2024-02-24
fix pre-signed POST when file is passed as regular form field

bentsku:
## Motivation
As reported in #10309, we were supporting a multipart form that AWS supported. Thanks to the user's reproducer, I could inspect the request we received and see the differences:

This were the two different lines:
```
# from node
Content-Disposition: form-data; name="file"
# from python
Content-Disposition: form-data; name="file" filename="file"
```

Which made the access to `context.request.files["file"]` fail because it could not really parse `file` as a file because it had no filename. 

## Changes
- added a case to directly fetch the file from the form is it is present there (if it is not, it is in the form) and updated the logic around that
- added a test reproducing this exact case
- also added a test case for when we don't set the multipart/form header, to prevent future issues

Also found something around the `${filename}` variable that can actually be used in all form fields, this will be tackled in a follow up PR.

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -4,6 +4,7 @@ import datetime
 import json
 import logging
 from collections import defaultdict
+from io import BytesIO
 from operator import itemgetter
 from secrets import token_urlsafe
 from typing import IO, Optional, Union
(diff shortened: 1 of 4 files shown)


macnev2013 (approved):
Looks good to me 👍

The CI status hasn't refreshed yet, but I confirmed on CircleCI that the tests are successful.

Issue #10311, opened 2024-02-23
bug: When defining APIGateway Resources with Path Parameters, the order of definition in CDK causes behavior change


Garethp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When defining path parameter resources, for example:

 * `/documents/{documentId+}`
 * `/documents/search`

In CDK, the order of definition changes what LocalStack returns. If you define the path parameter `/documents/{documentId+}` before `/documents/search` and then try to hit `/documents/search`, LocalStack instead invokes the integration for `/documents/{documentId+}`, rather than the intended one.

However, if you define `/documents/search` before `/documents/{documentId+}` in CDK, LocalStack invokes the correct integration.

### Expected Behavior

In AWS, the order does not matter, and no matter which order they're defined in trying to hit `/document/search` will result in the integration for that resource being invokes.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```ts
export class GatewayStack extends Stack {
  constructor(scope: Construct, id: string, props: StackProps) {
    super(scope, id, props);

    const api = new RestApi(this, "test-api", {
      restApiName: "test-api",
    });

    const getByIdLambda = new Function(this, "get-by-id-handler", {
      runtime: Runtime.NODEJS_20_X,
      code: Code.fromInline(
        "exports.handler = async () => ({ statusCode: 200, body: 'Get By Id' });",
      ),
      handler: "index.handler",
    });

    const searchLambda = new Function(this, "search-lambda-handler", {
      runtime: Runtime.NODEJS_20_X,
      code: Code.fromInline(
        "exports.handler = async () => ({ statusCode: 200, body: 'Search Results' });",
      ),
      handler: "index.handler",
    });

    const documents = api.root.addResource("document");
    const getById = documents.addResource("{documentId+}");
    const search = documents.addResource("search");

    getById.addMethod("GET", new LambdaIntegration(getByIdLambda));
    search.addMethod("GET", new LambdaIntegration(searchLambda));
  }
}
```

```sh
GATEWAY_ID=$(awslocal apigateway get-rest-apis --query "items[0].id" --output text)
curl https://$GATEWAY_ID.execute-api.localhost.localstack.cloud:4566/prod/document/search
```

### Environment

```markdown
[...]

PR #10317, opened 2024-02-25
fix API Gateway route matching order

bentsku:
## Motivation
As reported in #10311, the order of resource creation mattered in our route matching logic, because we put at the same level "exact" matching and "parameter" matching. 
This is a follow up from #9208 in a way, which fixed the logic around method matching, but kept a flaw in the logic. 

## Changes
- only do an early return if we have an exact match. If we have regular parameter matching, return the resource with the least parameters in the path, which means it has more "hardcoded" values. Keep the proxy logic the same
- add cases in the unit tests, which validate the order of the resources
- also add a validated integration test with the user use-case

I've also tested the use reproducer given here: https://github.com/Garethp/localstack-bugs/tree/order-dependant-path-parts, with instructions on how to reproduce, and I've confirmed it fixed the issue (https://github.com/localstack/localstack/issues/10311#issuecomment-1962763343).

### DIFF
--- a/localstack/services/apigateway/helpers.py
+++ b/localstack/services/apigateway/helpers.py
@@ -744,27 +744,38 @@ def get_resource_for_path(
     # /{proxy+} and /api/{proxy+} for inputs like /api/foo/bar
     # /foo/{param1}/baz and /foo/{param1}/{param2} for inputs like /for/bar/baz
     if len(matches) > 1:
-        filtered_matches = []
+        proxy_matches = []
+        param_matches = []
         for match in matches:
             match_methods = list(match[1].get("resourceMethods", {}).keys())
             # only look for path matches if the request method is in the resource
             if method.upper() in match_methods or "ANY" in match_methods:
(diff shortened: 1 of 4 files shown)


localstack/services/apigateway/helpers.py
```diff
@@ -744,27 +744,38 @@ def get_resource_for_path(
...
+                elif path_matches_pattern(path, match[0]):
+                    # parameters can fit in
+                    param_matches.append(match)
+                    continue
+
+                proxy_matches.append(match)
+
+        if param_matches:
+            # count the amount of parameters, return the one with the least which is the most precise
+            sorted_matches = sorted(param_matches, key=lambda x: x[0].count("{"))
```
calvernaz, line 766: nit: we could have debug logs on the returning points since we have a couple.

calvernaz (approved):
Thanks for improve this 🚀 

Wonder if the new test could fit in one of the existing tests, like `test_create_proxy_resource_validation` but either way looks good

Issue #10216, opened 2024-02-10
bug: Step functions does not support ECS service type


utibeabasi6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
6:48:04 AM | CREATE_FAILED        | AWS::StepFunctions::StateMachine | BuildSourceCodeWor...ateMachine67076E41
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: Error=NotImplementedError Args=["Unsupported service: 'ecs'."] in definition '{"StartAt":"Update Deployment: Status=BUILD_IN_PROGRESS","States"
```

I get the above error when trying to deploy my step functions workflow with CDKLocal

### Expected Behavior

The stack should deploy successfully

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS: Linux
- LocalStack: latest
```

utibeabasi6:
would appreciate any help @MEPalma

MEPalma:
@utibeabasi6 thanks for compiling this. It's planned in development and should be available in the coming weeks. I'll update this issue when support will be added.

utibeabasi6:
Thanks

PR #10321, opened 2024-02-26
[SFN] Base Support for ECS Optimised Service Integration

MEPalma:
## Motivation
This PR enhances the SFN v2 interpreter by introducing foundational support for optimised ECS tasks. Moving forward, the focus will shift towards improving failure decoration, with these changes enabling the mapping of ECS failures to runtime errors.
Closes: #10216 

## Changes
- Added SFN - ECS optimised integration snapshot tests
- Added ECS service state including support for .sync operations
- Added botocore arguments normalisation for list types


tests/aws/services/stepfunctions/v2/services/test_ecs_task_service.py
```diff
@@ -0,0 +1,516 @@
...
+            "$..StartedAt",
+            "$..StartedBy",
+            "$..StopCode",
+            "$..StoppedAt",
+            "$..StoppedReason",
+            "$..StoppingAt",
+            "$..TaskDefinitionArn",
+            "$..Version",
+            "$..parameters.Cluster",
+        ]
```
dominikschubert, line 248: nit: not required for this PR but would maybe make sense to re-use this list across the tests in the files

tests/aws/services/stepfunctions/v2/services/test_ecs_task_service.py
```diff
@@ -0,0 +1,516 @@
...
+from localstack_snapshot.snapshots.transformer import RegexTransformer
+
+from localstack.testing.pytest import markers
+from localstack.utils.analytics.metadata import is_license_activated
+from tests.aws.services.stepfunctions.utils import launch_and_record_execution
+
+
+# TODO: figure out a better way, maybe via marker? e.g. @markers.localstack.ext
+@pytest.mark.skipif(condition=not is_license_activated(), reason="integration test with pro")
+class TestTaskServiceECS:
```
dominikschubert, line 17: The file shows an interesting problem emerging: handling multiple infra configurations. Just highlighting it since this is something I'd like to investigate further in the future. Might make sense to use this test class as a sample.

tests/aws/services/stepfunctions/v2/services/test_ecs_task_service.py
```diff
@@ -0,0 +1,516 @@
...
+import pytest
+from localstack_snapshot.snapshots.transformer import RegexTransformer
+
+from localstack.testing.pytest import markers
+from localstack.utils.analytics.metadata import is_license_activated
+from tests.aws.services.stepfunctions.utils import launch_and_record_execution
+
+
+# TODO: figure out a better way, maybe via marker? e.g. @markers.localstack.ext
+@pytest.mark.skipif(condition=not is_license_activated(), reason="integration test with pro")
```
dominikschubert, line 16: should be ok for now but I'll add a marker for this as a follow-up soon :tm:

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
```diff
@@ -0,0 +1,127 @@
...
+                return
+
+        env.stack.append(response)
+
+    def _sync_to_run_task(
+        self,
+        env: Environment,
+        resource_runtime_part: ResourceRuntimePart,
+    ) -> None:
+        ...
```
dominikschubert, line 85:
```suggestion
```

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
```diff
@@ -0,0 +1,127 @@
...
+                self._normalise_response(
+                    response=describe_tasks_output, service_action_name="describe_tasks"
+                )
+                return describe_tasks_output["Tasks"][0]  # noqa
+
+            return None
+
+        termination_output: Optional[dict] = None
+        while env.is_running() and termination_output is None:
+            termination_output: Optional[dict] = _has_terminated()
```
dominikschubert, line 109: does the env.is_running cause a wait interval between the describe_tasks checks or is this a kind of busy waiting scheme?

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
```diff
@@ -0,0 +1,127 @@
...
+                    cluster=cluster_arn, tasks=[task_arn]
+                )
+                describe_tasks_output.pop("ResponseMetadata", None)
+                self._normalise_response(
+                    response=describe_tasks_output, service_action_name="describe_tasks"
+                )
+                env.stack.append(describe_tasks_output)
+                return
+
+        env.stack.append(response)
```
dominikschubert, line 78: do we even want to have this as a fallback, given that this shouldn't be valid?

dominikschubert (changes requested):
Mostly LGTM, just a few minor things that need to be addressed/answered before merge. No need to change anything in the test setup for the PR.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
MEPalma: This logic simply terminates the evaluation if the program execution received a stop signal whilst synching. As sync may be executed until the state timeout, this just clears resources sooner by exiting the sync now rather than at timeout.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
MEPalma: My thought was it could ease bug reports if we see the invalid valued being passed on, but yes technically it can be removed.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
dominikschubert: I might not have expressed my question correctly. What is preventing this from doing basically unlimited describe_tasks calls in the while loop, is there any throttling/minimum interval involved?

MEPalma: Unlimited requests are prevented by stop signals or the state's timeout, however yes absolutely, good point, we don't have throttling and we should add one.

MEPalma: I have added a temporary solution to this by invoking a throttling function in each iteration. We should probably abstract this logic in the callback class to enforce all sync operations to implement an sync iteration pattern which is automatically throttled and interrupted.

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_callback.py
```diff
@@ -108,6 +109,10 @@ def _sync2(
             f"Unsupported .sync:2 callback procedure in resource {self.resource.resource_arn}"
         )
 
+    @staticmethod
+    def _throttle_sync_iteration():
+        time.sleep(0.5)
```
dominikschubert, line 114: might make sense to make this configurable. not a blocker though

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_ecs.py
dominikschubert: My argument against this would be that this hides an issue that the user needs to fix in his state machine and won't deploy like this against AWS as well, so I don't see a reason to hide it. It will also make potential bug reports harder to debug.

dominikschubert (approved):
LGTM, minor nits/concerns but nothing blocking :+1:

PR #10322, opened 2024-02-26
skip validating S3 pre-signed POST length for CI flake

bentsku:
## Motivation
Follow up from #10314, I missed skipping the length in the `ListObjectsV2` call, and this will most probably lead in CI failure as sometimes in CI the POST body is empty. We're still not quite sure what's happening here and it's not reproducible locally, but in the meantime we should avoid any CI blocks because of this. 

## Changes
- add the `Size` field in the snapshot ignore

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -10377,6 +10377,7 @@ class TestS3PresignedPost:
     @markers.snapshot.skip_snapshot_verify(
         paths=[
             "$..ContentLength",
+            "$..Size",
             "$..ETag",
             "$..HostId",
         ],  # FIXME: in CI, it fails sporadically and the form is empty


macnev2013 (approved):
Just curious, are we only snapshot testing `size` parameter in this test case, or are there other tests where this is snapshot tested and works correctly?

It looks good to merge from my end!

bentsku:
I think I'm going to wait a bit before merging until we actually hit the flake, maybe this is fixed now... 🤔 at least it's ready to merge at the first flake we see!

The issue seemed to happen only for S3 POST with form requests, they always work locally, but sometimes in CI, the file part of the form would just be empty, so the ETag and the size field would be different, but it would still be successful otherwise (not sure we can consider it a success though 🙃). Never could reproduce locally. 

So let's see if we hit a flake and then I'll merge! Thanks for the review!

PR #10323, opened 2024-02-26
Validation lambda function name pattern on invoke

dominikschubert:
## Motivation

This surfaced when debugging a customer issue where a lambda was invoked via stepfunctions but the specified ARN was incorrect. This lead to a misleading error message which should be fixed with this PR

## Changes

- Extract regex from exception message and raise ValidationException when a  non-matching function name is provided on invoke(). 

## Discussion
We might want to add this to other places as well but that will require more extensive test adaptations.


dfangl (approved):
LGTM!

PR #10324, opened 2024-02-26
Add SQS high-throughput deduplication behaviour

baermat:
## Motivation
Follow-up of #10274 
This PR implements "FIFO high-throughput" mode. More precisely, since we do not enforce any rate limits for SQS, this PR adds the necessary changes in regards to the deduplication behaviour.
Usually, messages in a queue are treated as duplicates simply if they have the same deduplication ID. Message groups are ignored. In HT mode deduplication happens at MessageGroup level.
On a side-note: it seems that you cannot change a "regular" FIFO queue to a HT one after creation. You can change the attributes, but AWS does not seem to change the deduplication behaviour of the queue. This makes it necessary to track HT FIFO queues somehow. For now this is simply a flag because a new queue type seemed like overkill.

## Changes
- adds deduplication behaviour for high-throughput mode FIFO queues
- add flag to mark HT FIFO queues.
- fixes small discrepancy in how we return bool values of queue attributes

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -663,6 +663,8 @@ class SqsQueue:
                 value = func()
                 if value is not None:
                     result[attr] = value
+            elif value == "False" or value == "True":
+                result[attr] = str(value).lower()
             elif value is not None:
                 result[attr] = value
         return result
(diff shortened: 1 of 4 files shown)


thrau (approved):
very cool addition, and great job on analyzing the AWS behavior! changes LGTM. well done! :clap:

PR #10337, opened 2024-02-27
Make keys_pascalcase_to_lower_camelcase recursive

dominikschubert:
## Motivation
Extended `keys_pascalcase_to_lower_camelcase` cloudformation resource provider util since a few resources need this functionality to be recursive.

## Changes

- make `keys_pascalcase_to_lower_camelcase` recursive


simonrw (approved):
Nice clean implementation. Are there any cases where the recursive behaviour would not be the correct behaviour? I guess the tests will uncover that.

Morijarti (approved):
I've tested it with "AWS::IoTAnalytics::Dataset", which requires recursion and also lower-casing dictionaries inside of a list.

Works perfectly

Issue #10328, opened 2024-02-26
bug: S3 CopyObject returns 500 - Internal error


anna-mazhar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

- Localstack fails with HTTP 500, error code "InternalError", message "not enough values to unpack (expected 2, got 1)" when `CopySource` is an invalid string for `copy_object` API.

### Expected Behavior

- For the same API `copy_object`, the cloud fails with HTTP 400, error code "InvalidArgument".
- Localstack should handle invalid `CopySource` i.e., it is a dictionary with one value or an invalid string not adhering to "{bucket}/{key}" format.

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    botoclient.copy_object(CopySource="Invalid_random_string", Bucket="test_bucket", Key=key) 

 test_bucket is a valid bucket.  

### Environment

```markdown
- OS: Ubuntu 22.04.2 LTS
- LocalStack: 2.1.0
- boto3: 1.28.6
```

PR #10338, opened 2024-02-27
fix S3 copy-source format validation

bentsku:
## Motivation
As reported in #10328, we would not properly validate the `CopySource` parameter before parsing it.

## Changes
Added a check for the presence of `/` before splitting, and a test for the case. 

_fixes #10328_


macnev2013 (approved):
Thanks for fixing this. LGTM 🚀

PR #10341, opened 2024-02-27
Fix attributes, create and delete for AWS::Redshift::Cluster

dominikschubert:
## Motivation

@maxhoheiser found some issues with the current provider which this PR aims to address

## Changes

- Made create wait for the actual cluster state
- Made delete wait for the resource not existing anymore
- Added missing read-only properties (`Endpoint.Address`, `Endpoint.Port`)
- Extended test with GetAtt


simonrw (approved):
Looks good thanks for spotting

PR #10346, opened 2024-02-28
Fix CircleCI workflow tracking

silv-io:
## Motivation

The tracking of workflows does not happen currently when CircleCI tests are failing. Also, the tracked job ID in circle-ci does not match the semantics of the job id in the GitHub actions. In CircleCI it is a specific singular job run in a workflow. In GitHub it means the classifier of a job inside all the workflows.

## Changes

- Use a [workaround](https://discuss.circleci.com/t/workaround-run-jobs-sequentially-regardless-of-the-outcome-of-the-required-jobs/40807) to send circle ci workflow data also on fail (something which CircleCI can't do out of the box)
- Set the job id in the circle-ci job to the classifier of the job.
- Set the GH action version to a major tag so that it does get minor release updates automatically

## Testing

For the CircleCI fix, see these two (simplified) workflow runs:

1. In the [first run](https://app.circleci.com/pipelines/github/localstack/localstack/22737/workflows/1ae80f24-3d73-45a2-b5aa-d17da494a4d5), we have a failing job (`exit 1`), but sending the stats still worked, with the outcome "failure".
2. In the [second run](https://app.circleci.com/pipelines/github/localstack/localstack/22739/workflows/31ee30dc-3519-4ccb-81c8-d8388b550ff6), the failing job was "fixed" (`exit 0`), and the outcome is "success".

For the GitHub Action version fix, see if the push-to-tinybird job works properly with the new version indicator.

## TODO

- [x] Add back the entire CircleCI pipeline, and the guards

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,16 +505,31 @@ jobs:
             paths:
                 - repo/stats/
   waiter:
-    docker:
-      - image: circleci/node
+    executor: ubuntu-machine-amd64
+    working_directory: /tmp/workspace/repo
     steps:
       - run: |
-          while [[ $(curl --location --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job"| jq -r '.items[]|select(.name != "waiter" and .name != "push")|.status' | grep -c "running") -gt 0 ]]
-            do
-              sleep 5
+          while [[ $(curl --location --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job"| jq -r '.items[]|select(.name != "waiter" and .name != "push")|.status' | grep -c "running") -gt 0 ]]; do
+              sleep 10
             done
-          curl --location --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CCI_PROJECT_TOKEN"
-      - run: echo "All required jobs have now completed"
+          FAILED_COUNT=$(curl --location --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" | jq -r '.items[]|.status' | grep -c "failed") || true
+          echo "failed count" $FAILED_COUNT
+          mkdir stats
+          if [[ $FAILED_COUNT -eq 0 ]]; then
+              echo "success" > stats/outcome.txt
+          else
+              echo "failure" > stats/outcome.txt
+          fi
+          cat stats/outcome.txt
+      - persist_to_workspace:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -536,6 +548,15 @@ workflows:
   main:
     jobs:
       - save-start-time
+      - waiter:
+          requires:
+            - save-start-time
```
alexrashed, line 553: Would it also make sense to also only run this on `master`?

.circleci/config.yml
```diff
@@ -504,29 +504,41 @@ jobs:
             root: /tmp/workspace
             paths:
                 - repo/stats/
+  waiter:
```
alexrashed, line 507: Really terrible how hard it is in CircleCI to run a command as a "finally" block... 🤦🏽 
Great work with this workaround though!

.github/workflows/tests-cli.yml
```diff
@@ -89,14 +89,15 @@ jobs:
...
     steps:
       - name: Push to Tinybird
-        uses: localstack/tinybird-workflow-push@v3.0.0
+        uses: localstack/tinybird-workflow-push@v3
         with:
           workflow_id: "tests_cli"
           tinybird_token: ${{ secrets.TINYBIRD_CI_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          tinybird_datasource: "ci_workflows"
+          tinybird_datasource: "ci_workflows_test"
```
alexrashed, line 103: Looks like you are using a test datasource here still?

alexrashed (commented):
Nice, we're nearly there! Really weird that CircleCI does not allow running a step easily if another failed...

.github/workflows/tests-cli.yml
silv-io: Just added this quickly right now to test out if the major version tag really works. Will remove shortly

.circleci/config.yml
```diff
@@ -535,10 +547,17 @@ jobs:
...
+          filters:
+            branches:
+              only: master
+      - waiter:
           requires:
             - save-start-time
+      - send-stats:
+          requires:
+            - waiter
+      - install
```
alexrashed, line 560: Shouldn't install still depend on save-start-time? Otherwise this could be problematic in case there are not enough runners available.

alexrashed (approved):
Only found one small nitpick, afterwards we are good to go! 🚀

.circleci/config.yml
silv-io: When I make install depend on save-start-time, then it (and the whole pipeline) won't always run because save-start-time only runs in master

PR #10353, opened 2024-02-29
Disable legacy CFn model execution

dominikschubert:
## Motivation

First part of the cleanup, now that we have basically finished the migration of the resource providers

## Changes
- Remove adapter for legacy providers
- Remove special cases that were only in place for legacy usage
- Stricter handling of the returned OperationStatus from a resource provider
- Improved exception message for `TimeoutError` (now includes logical resource id / type)

## TODO
- [x] Test -ext against this to avoid any surprises

## Follow-up
- Delete old models

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -56,6 +56,7 @@ LOG = logging.getLogger(__name__)
 # list of static attribute references to be replaced in {'Fn::Sub': '...'} strings
 STATIC_REFS = ["AWS::Region", "AWS::Partition", "AWS::StackName", "AWS::AccountId"]
 
+# FIXME: remove
 # maps resource type string to model class
 RESOURCE_MODELS: dict[str, Type[GenericBaseModel]] = {
     model.cloudformation_type(): model for model in get_all_subclasses(GenericBaseModel)
@@ -77,14 +78,6 @@ def get_attr_from_model_instance(
     resource: dict, attribute_name: str, resource_type: str, resource_id: str
 ) -> str:
     properties = resource.get("Properties", {})
-
-    # TODO: fix this somewhere else
-    if legacy_state := resource.get("_state_"):
-        properties = {**properties, **legacy_state}
-
-        if legacy_deployed_state := resource.get("_last_deployed_state"):
-            properties = {**properties, **legacy_deployed_state}
-
     # if there's no entry in VALID_GETATT_PROPERTIES for the resource type we still default to "open" and accept anything
     valid_atts = VALID_GETATT_PROPERTIES.get(resource_type)
     if valid_atts is not None and attribute_name not in valid_atts:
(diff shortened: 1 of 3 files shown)


localstack/services/cloudformation/resource_provider.py
```diff
@@ -745,27 +519,12 @@ def execute_action(
...
-            return self._load_legacy_resource_provider(resource_type)
-
         # 1. try to load pro resource provider
         # prioritise pro resource providers
         if PRO_RESOURCE_PROVIDERS:
-            # temporary patch until this has equivalent resource providers in -ext
-            if resource_type in {
-                "AWS::ECR::Repository",
-                "AWS::SecretsManager::SecretTargetAttachment",
-                "AWS::EC2::SubnetRouteTableAssociation",
```
dominikschubert, line 766: this especially might cause some new issues with -ext since we didn't test them yet because of this exemption list!

tests/aws/services/stepfunctions/v2/timeouts/test_timeouts.py
```diff
@@ -83,6 +83,8 @@ def test_fixed_timeout_service_lambda(
             exec_input,
         )
 
+    # FIXME: https://app.circleci.com/pipelines/github/localstack/localstack/22941/workflows/ed5c7ca1-f354-4e6a-b4c2-c85007d2cceb/jobs/186457?invite=true#step-104-3159
+    @pytest.mark.skip(reason="flaky")
```
dominikschubert, line 87: sneaked that in :eyes:

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1413,9 +1406,6 @@ def create_resource_provider_executor(self) -> ResourceProviderExecutor:
             stack_name=self.stack.stack_name,
             stack_id=self.stack.stack_id,
             provider_config=self.provider_config,
-            # FIXME: ugly
-            resources=self.resources,
-            legacy_base_models=RESOURCE_MODELS,
```
simonrw, line 1418: :tada:

localstack/services/cloudformation/resource_provider.py
```diff
@@ -686,7 +465,9 @@ def deploy_loop(
                     raise  # re-raise here if explicitly enabled
 
         else:
-            raise TimeoutError("Could not perform deploy loop action")
+            raise TimeoutError(
+                f"Resource deployment for resource {raw_payload['requestData']['logicalResourceId']} (type {raw_payload['resourceType']}) timed out."
+            )
```
simonrw, line 470: Thank you for including this. This will be a nice QOL improvement

simonrw (approved):
My only concern is lack of test coverage and missed resource providers (now that we are strictly using the new versions). How confident are we that we have migrated them all?!

dominikschubert:
> How confident are we that we have migrated them all?!
Fairly confident, since I've checked that via script at least (comparing old/new resources). Yes, some resources might not be covered via tests but those should be fairly few now. At least when I was going through resources and doing reviews I've added a lot of missing tests as well. 

Everything else can be done reactively, I'd say :+1:

PR #10355, opened 2024-02-29
Fix UserData regression from provider migration in AWS::EC2::Instance

dominikschubert:
## Motivation

Initially introduced into the legacy model with https://github.com/localstack/localstack/pull/8957

Users reported this was broken again, so I investigated and found that there was a small bug in the provider migration that caused this setting to never be passed to the API call.

## Changes

- Fix param selection for API call


Morijarti (approved):
Good catch. LGTM

PR #10358, opened 2024-02-29
Fix bin/localstack venv detection

simonrw:
## Motivation

While using [functiontrace](https://functiontrace.com/) to profile LocalStack performance, I tried to use the `bin/localstack` script to run the code in host mode. `functiontrace` requires a script to run, so this seemed a script to run for profiling. Unfortunately it did not pick up my virtual environment correctly (I do not use `.venv`), so this PR aims to detect the venv path correctly (or at least _better_).

## Changes

Try to read the venv path from the standard `VIRTUAL_ENV` environment variable, and fall back to `.venv` if not present.

### DIFF
--- a/bin/localstack
+++ b/bin/localstack
@@ -4,8 +4,9 @@ import glob
 import os
 import sys
 
+VIRTUAL_ENV_PATH = os.getenv("VIRTUAL_ENV")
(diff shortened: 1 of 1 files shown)


simonrw:
Closing as this script is not worth maintaining. There are easier ways to run LocalStack from a script.

PR #10361, opened 2024-02-29
add timestream to the list of composite apis

giograno:
## Motivation
With this PR we add `timestream` to the list of composite APIs.


thrau (approved):
LGTM!

PR #10363, opened 2024-02-29
Update kinesis-mock to 0.4.6

etspaceman:
Just a dependency updates release. See https://github.com/etspaceman/kinesis-mock/releases/tag/v0.4.6


alexrashed (approved):
Thanks a lot for the proactive update! 🦸🏽 🚀 
I'm looking into why the failing pipelines are executed (they should actually be skipped, see #10366), the failures are completely unrelated.

PR #10365, opened 2024-03-01
Adding support for partial wildcards for S3 CORS

cloutierMat:
## Motivation

Fixing #10090 . AWS S3 CORS allows for partial wildcard `"http://*.example.com"`. The current changes will allow `OPTIONS` requests to go through instead of returning a 403 error

## Changes

- Replace the logic of looking for an exact match from Origin header to configured allowed origins by a regex that will match anything in lieu of the `*`.

## Testing

As mentioned on the issues description running the following command targeting an S3 bucket. With CORS configured as follow.

```
$ aws s3api get-bucket-cors --bucket foo --endpoint-url=http://localhost:4566                                                          
{
    "CORSRules": [
        {
            "AllowedHeaders": [
                "*"
            ],
            "AllowedMethods": [
                "GET",
                "PUT"
            ],
            "AllowedOrigins": [
                "http://*.example.com",
                "https://app.localstack.cloud",
                "http://app.localstack.cloud"
            ]
        }
    ]
}
```

```
$ curl -v --request OPTIONS 'http://127.0.0.1:4566/foo/' -H 'Origin: http://subd.example.com' -H 'Access-Control-Request-Method: GET'
```

PR #10366, opened 2024-03-01
fix workflow execution for PRs coming from forks

alexrashed:
## Motivation
Unfortunately, it turns out that the fix to avoid the execution of test pipelines which rely on secrets implemented in https://github.com/localstack/localstack/pull/9383 is not correctly working, as we saw with the workflows executed on the following PRs:
- https://github.com/localstack/localstack/pull/10363
- https://github.com/localstack/localstack/pull/10365

It turns out that `github.repository` evaluates to the PR target repo (which in this case is `localstack/localstack`.
However, we want to make sure that the source and the target of the PR are not the same repo.
This PR tries to fix this condition by checking the pull request head target as described in https://github.com/orgs/community/discussions/26829.

## Changes
- Prevent the execution of the "Community Tests against Pro" (which inherently rely on the usage of secrets which cannot and shouldn't be provided for runs on untrusted code) by fixing the condition.
- Fixing the S3 image tests by only using the secrets if they are available.

## Testing
Unfortunately, this cannot really be tested properly as long as it's in a PR. Happy for any feedback / tips on how to test this though!

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -1228,9 +1228,6 @@ class TestDockerClient:
         docker_client.restart_container(name)
         assert docker_client.is_container_running(name)
         docker_client.stop_container(name)
-        import time
-
-        time.sleep(5)
         assert not docker_client.is_container_running(name)
 
     @markers.skip_offline
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -1228,6 +1228,9 @@ class TestDockerClient:
         docker_client.restart_container(name)
         assert docker_client.is_container_running(name)
         docker_client.stop_container(name)
+        import time
+
+        time.sleep(5)
         assert not docker_client.is_container_running(name)
 
     @markers.skip_offline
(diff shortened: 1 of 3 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -1228,6 +1228,9 @@ def test_is_container_running(self, docker_client: ContainerClient, dummy_contai
         docker_client.restart_container(name)
         assert docker_client.is_container_running(name)
         docker_client.stop_container(name)
+        import time
+
+        time.sleep(5)
         assert not docker_client.is_container_running(name)
```
dominikschubert, line 1234: nit: wouldn't a retry be a bit more stable for this? Or is this just some leftover from debugging?

dominikschubert (approved):
LGTM, seems sensible. Probably not worth it to test it extensively before merging. We can just open a fork PR after the fact and iterate over it if needed.

tests/integration/docker_utils/test_docker.py
alexrashed: Whoops, sorry! That's a leftover which smuggled itself into the stage... Thanks for catching that!

PR #10367, opened 2024-03-01
scheduled run for non-default credentials

sannya-singal:
## Motivation
Establishing a nightly scheduled pipeline to execute circleci tests using randomized non-default credentials across multiple accounts and regions in the community. The purpose is to guarantee cross-account compatibility within the community and to identify any failing test cases.

## Changes
This PR sets a nightly scheduled workflow based on the value for `randomize-aws-credentials` for non-default random variables and pushes this to `community_tests_circleci_ma_mr` tinybird  datasource. 

## TODO

What's left to do:

- [x] change `cron` value to 1:00am UTC which 2:00am CET(open for suggestions) 
- [x] change filter branch to `master`
- [x] discuss how to work on community contributors PR for non-default creds with @alexrashed 
- [x] check if non default values are getting passed to acceptance, bootstrap, integration, etc. tests 
- [x] discuss if we need to send stats and records for non-default creds workflow to tinybird 
- [x] remove printing non-default values in `generate-random-aws-credentials` job
- [x] push this data to a new tinybird datasource being used for cross-accounts run and notify @silv-io about it


.circleci/config.yml
```diff
@@ -406,6 +448,36 @@ jobs:
...
+          command: |
+            # Generate a random 12-digit non-default account number
+            while true; do
+              # Generate a random 12-digit number for TEST_AWS_ACCOUNT_ID
+              export TEST_AWS_ACCOUNT_ID=$(LC_ALL=C tr -dc '0-9' < /dev/urandom | fold -w 12 | head -n 1)
+              # Check if the generated number is not 000000000000
+              if [ "$TEST_AWS_ACCOUNT_ID" != "000000000000" ]; then
+                break
+              fi
+            done
```
alexrashed, line 468: nit: I'm not sure if it's really necessary to check that the randomly generated 12-digit number is not all-zeros (the default).  The chance is one in a trillion, right (assuming it's perfectly random)?

.circleci/config.yml
```diff
@@ -599,6 +671,7 @@ jobs:
 
 
 workflows:
+  version: 2
```
alexrashed, line 674: q: interesting! I didn't know you could version CircleCI workflows. What would be the effect?

.circleci/config.yml
```diff
@@ -10,6 +10,9 @@ parameters:
   skip_test_selection:
     type: boolean
     default: false
+  schedule-non-default-creds:
+    type: string
+    default: "0 2 * * *"
```
sannya-singal, line 15: the `cron` value is set to 1:00am UTC which 2:00am CET(open for suggestions)

.circleci/config.yml
```diff
@@ -697,3 +770,119 @@ workflows:
             - docker-test-amd64
             - docker-test-arm64
             - unit-tests
+  scheduled-workflow:
+    triggers:
+      - schedule:
+          cron: << pipeline.parameters.schedule-non-default-creds >>
+          filters:
+            branches:
+              only: multi_acc_schedule_run
```
sannya-singal, line 779: TOOD: change filter branch to `master` before merging this PR

.circleci/config.yml
```diff
@@ -697,3 +770,119 @@ workflows:
...
+            branches:
+              only: master
+          requires:
+            - itest-sfn-legacy-provider
+            - itest-s3-v2-legacy-provider
+            - itest-cloudwatch-v2-provider
+            - acceptance-tests
+            - docker-test-amd64
+            - docker-test-arm64
+            - unit-tests
```
alexrashed, line 888: I think it would be much simpler to:
- Introduce a boolean parameter which enables or disables the account ID and region randomization.
  - This parameter is `false` by default, but can be enabled when triggered manually or in a schedule trigger.
- The schedule trigger then just sets the parameter to `true`.

This would have the following benefits:
- The whole pipeline definition (which is duplicated here) could be reused.
- A user of the pipeline would be able to also manually trigger the pipeline run with enabled account ID  and region randomization.

What's your opinion? Is there a reason you explicitly defined a second workflow here?
I'm happy to assist if you need any help in migrating the current approach :)

.circleci/config.yml
```diff
@@ -34,7 +37,11 @@ commands:
             if [[ $CIRCLE_BRANCH == "master" ]] ; then
               echo "export TINYBIRD_PYTEST_ARGS='--report-to-tinybird '" >> $BASH_ENV
             fi
-            echo "export TINYBIRD_DATASOURCE=community_tests_circleci" >> $BASH_ENV
+            if [ -f aws_credentials.env ]; then
+              echo "export TINYBIRD_DATASOURCE=scheduled_tests_circleci" >> $BASH_ENV
```
alexrashed, line 41: The name here should imply that it's not just a scheduled run, but a multiaccounts / multiregion test run.
Maybe something like `community_tests_circleci_ma_mr`?

.circleci/config.yml
```diff
@@ -124,6 +131,11 @@ jobs:
             TEST_PATH: "tests/unit"
             COVERAGE_ARGS: "-p"
           command: |
+            if [ -f aws_credentials.env ]; then
+              source aws_credentials.env
+            else
+              echo "aws_credentials.env file not found. Skipping setup of non-default aws credentials."
+            fi
```
alexrashed, line 138: nit: These sections are a bit repetitive. I think it would be great if we could have this in a command, similar to the `prepare-pytest-tinybird` command. You can also implement the conditional quite nicely.
Maybe something like `prepare-account-region-randomization`?

.circleci/config.yml
alexrashed: If you migrate this workflow to be parameterized instead of duplicated you would also be able to use the parameter here.
This would have the benefit that it does not rely on the order of execution. This code here currently only works if `prepare-pytest-tinybird` is executed _after_ `generate-random-aws-credentials`.

.circleci/config.yml
alexrashed: If you run this in a command, you could maybe just put the whole `generate-random-aws-credentials` in there and not have a specific command alltogether.
This would simplify the code, have all the account-id and region randomization code together, and it could be "plugged" into a job by just adding the command. However, this would mean that each job would have a different set of randomly generated IDs.

alexrashed (changes requested):
Nice! It will be really cool to have this in our main pipeline, effectively circumventing a drift concerning the multi accounts and multi region support!
The general principle is great! I just found some potential to cleanup the code and maybe make it a bit more accessible (allow users to manually trigger this).
Let me know if you have any questions or if you need any help! 🚀

.circleci/config.yml
sannya-singal: Yes I wanted to avoid different set of randomly generated IDs for every job, likewise we have it for default credentials in community tests right now. But yes we could do this as well.

.circleci/config.yml
sannya-singal: As discussed we will configure this through circlici project settings

.circleci/config.yml
sannya-singal: It was used to issue warnings for deprecation or breaking changes. Although as per docs for v2.1 config we do not require this anymore.

sannya-singal:
For non default creds i.e when `randomize-aws-credentials:true` we can refer: https://app.circleci.com/pipelines/github/localstack/localstack/23369/workflows/bfd9c1d5-00f8-487f-9849-60e2c2c21641/jobs/190644 build. 

Why introduced sourcing the environment file in unit and integration tests we can refer: https://app.circleci.com/pipelines/github/localstack/localstack/23356/workflows/64dbd1b7-3862-4312-8b9a-ce8f7c146436

For default credentials i.e. for `randomize-aws-credentials:false` we can refer: https://app.circleci.com/pipelines/github/localstack/localstack/23371/workflows/1e5323db-54b6-4d5e-b8e1-39f9cfbe3d4d

alexrashed (approved):
Thanks a lot for addressing all the comments! The result is really nice, concise, and clean! 🚀
As soon as this PR's checks are 💚 I can directly configure the schedule to run this new pipeline configuration once a day at night. 🚀 🤖

PR #10371, opened 2024-03-01
Improve delete reliability for kinesis streams

dominikschubert:
## Motivation

Deleting kinesis streams sometimes fails due to still existing consumers and/or because it is trying to be deleted while not in `ACTIVE` state yet.

## Changes

- Add flag to delete so that we can delete even though consumers are still present
- Add flag also to the deletion of streams for dynamodb streams and remove exception suppression


dfangl (approved):
Looks good! And getting rid of suppressing exceptions without log :tada:

PR #10375, opened 2024-03-01
Feat/firehose add redshift destination

maxhoheiser:
## Motivation
Using Firehose to distribute data to Redshift via a COPY statement from an intermediary S3 Bucket is one of the main targeted uses of Firehose, and thus should be also supported by LocalStack. We have two separate implementations of Redshift, in community we use the default Moto method as a fallback, but in ext, we use a Postgres database to enable persistence of Redshift data. Using the native boto `redshift_data` client for writing data from Firehose to Redshift, we can ensure, that we stay compatible with both community and ext. 

Configuration definitions of Redshift as target for Firehose: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_kinesisfirehose/CfnDeliveryStream.html#aws_cdk.aws_kinesisfirehose.CfnDeliveryStream.S3DestinationConfigurationProperty

## Changes
- adding the functionality to send data from Firehose to the defined S3 bucket, this is required to be consistent with AWS, since Redshift loads data from this intermediary bucket via a COPY statement
- adding the functionality to directly put data into the Redshift database

## Testing
To configure public connection to a Redshift cluster running on AWS, it needs to be created in ist own subnet, this is only possible if the subnet_id of an existing subnet is submitted during creation of the cluster, this functionality is only supported in -ext, therefore the test is not validated.

Testing Redshift directly without validating it against AWS is also only possible in -ext, since the `redshift_data` API is required to query data from Redshift, which in turn is a pro feature.

There is a follow-up PR that creates a scenario test incorporating Firehose to Redshift, in -ext, that will be validated against AWS.

### DIFF
--- a/localstack/aws/api/firehose/__init__.py
+++ b/localstack/aws/api/firehose/__init__.py
@@ -915,6 +915,7 @@ class RedshiftDestinationDescription(TypedDict, total=False):
     ClusterJDBCURL: ClusterJDBCURL
     CopyCommand: CopyCommand
     Username: Username
+    Password: Password
     RetryOptions: Optional[RedshiftRetryOptions]
     S3DestinationDescription: S3DestinationDescription
     ProcessingConfiguration: Optional[ProcessingConfiguration]
--- a/localstack/aws/api/firehose/__init__.py
+++ b/localstack/aws/api/firehose/__init__.py
@@ -915,7 +915,6 @@ class RedshiftDestinationDescription(TypedDict, total=False):
     ClusterJDBCURL: ClusterJDBCURL
     CopyCommand: CopyCommand
     Username: Username
-    Password: Password
     RetryOptions: Optional[RedshiftRetryOptions]
     S3DestinationDescription: S3DestinationDescription
     ProcessingConfiguration: Optional[ProcessingConfiguration]
--- a/localstack/services/firehose/provider.py
+++ b/localstack/services/firehose/provider.py
@@ -810,9 +810,6 @@ class FirehoseProvider(FirehoseApi):
         rows_to_insert = [self._prepare_records_for_redshift(record) for record in records]
         columns_placeholder_str = self._extract_columns(records[0])
         sql_insert_statement = f"INSERT INTO {table_name} VALUES ({columns_placeholder_str})"
-        for row in rows_to_insert:
-            sql_insert_statement += f"({row}),"
-        sql_insert_statement = sql_insert_statement.rstrip(",") + ";"
 
         execute_statement = {
             "Sql": sql_insert_statement,
@@ -827,14 +824,15 @@ class FirehoseProvider(FirehoseApi):
             aws_access_key_id=account_id, region_name=region_name
         ).redshift_data
 
-        try:
-            LOG.debug(
-                "Publishing to Redshift destination: %s. Data: %s", jdbcurl, sql_insert_statement
-            )
-            redshift_data.execute_statement(**execute_statement)
-        except Exception as e:
-            LOG.exception(f"Unable to put records {records} to redshift cluster.")
-            raise e
+        for row_to_insert in rows_to_insert:  # redsift_data only allows single row inserts
+            try:
+                LOG.debug(
+                    "Publishing to Redshift destination: %s. Data: %s", jdbcurl, row_to_insert
+                )
+                redshift_data.execute_statement(Parameters=row_to_insert, **execute_statement)
+            except Exception as e:
+                LOG.exception(f"Unable to put records {row_to_insert} to redshift cluster.")
+                raise e
 
     def _get_cluster_id_from_jdbc_url(self, jdbc_url: str) -> str:
         pattern = r"://(.*?)\."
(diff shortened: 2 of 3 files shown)


localstack/services/firehose/provider.py
```diff
@@ -779,3 +795,76 @@ def _get_s3_object_path(self, stream_name, prefix):
...
+            "Database": db_name,
+            "ClusterIdentifier": cluster_id,  # cluster_identifier in cluster create
+        }
+
+        if role_arn := redshift_destination_description.get("RoleARN"):
+            factory = connect_to.with_assumed_role(
+                role_arn=role_arn, service_principal=ServicePrincipal.firehose
+            )
+        else:
+            factory = connect_to()
```
simonrw, line 851: In this situation we need to get the account id and secret from the request context. See other usages of `connect_to`, e.g.

https://github.com/localstack/localstack/blob/76292e1d01d3141cbedfc0aea9f7b36095d6a23a/localstack/services/kinesis/provider.py#L92-L94

localstack/services/firehose/provider.py
```diff
@@ -779,3 +795,76 @@ def _get_s3_object_path(self, stream_name, prefix):
...
+                    value = value.replace("\t", " ")
+                    value = f"'{value}'"
+                elif value is None:
+                    value = "NULL"
+                else:
+                    value = str(value)
+                values.append(value)
+
+            value_str = ",".join(values)
+            rows_to_insert.append(value_str)
```
simonrw, line 833: Suggestion: this feels like it is a concept that could be named. Perhaps extract into a helper function? Then:

```suggestion
        columns_str = None
        rows_to_insert = [_transform_record_for_redshift(record) for record in records]
```

localstack/services/firehose/provider.py
```diff
@@ -779,3 +795,76 @@ def _get_s3_object_path(self, stream_name, prefix):
...
+
+        sql_insert_statement = f"INSERT INTO {table_name} ({columns_str}) VALUES "
+        for row in rows_to_insert:
+            sql_insert_statement += f"({row}),"
+        sql_insert_statement = sql_insert_statement.rstrip(",") + ";"
```
simonrw, line 838: We should pass the values in via the `Parameters` field rather than this SQL-injection-waiting-to-happen

simonrw (changes requested):
Hi, in general these changes look good, but I have a concern about the SQL statement we are using to inject the data. We should follow SQL best practice and pass the parameters in rather than building up a string (without any sort of escaping).

localstack/aws/api/firehose/__init__.py
```diff
@@ -915,6 +915,7 @@ class RedshiftDestinationDescription(TypedDict, total=False):
     ClusterJDBCURL: ClusterJDBCURL
     CopyCommand: CopyCommand
     Username: Username
+    Password: Password
```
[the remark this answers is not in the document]
maxhoheiser: Okay will remove it, but as far as I can see, `Password` is a required input field: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/firehose/client/create_delivery_stream.html
Removed in: https://github.com/localstack/localstack/commit/84ed09e542b491cd9ceef21cbc4ac5b0f8bc9c6a

localstack/services/firehose/provider.py
```diff
@@ -779,3 +796,71 @@ def _get_s3_object_path(self, stream_name, prefix):
...
+    def _put_to_redshift(
+        self,
+        stream_name: str,
+        records: List[Dict],
+        redshift_destination_description: RedshiftDestinationDescription,
+    ):
+        jdbcurl = redshift_destination_description.get("ClusterJDBCURL")
+        # pattern = r"jdbc:redshift://([^:/]+):(\d+)/([^/]+)(/[^/]+)?"
+        # match = re.search(pattern, jdbcurl)
+        # address, port, db_name, db_name = match.groups()
```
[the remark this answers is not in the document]
maxhoheiser: Is also already removed in the latest state

localstack/services/firehose/provider.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/commit/3b33957d536cc34f5244be94ced08fb9556c79ce

localstack/services/firehose/provider.py
maxhoheiser: extracted into separate functions in: https://github.com/localstack/localstack/commit/051c6630851564b916b556a226a043cfe125b20f

localstack/services/firehose/provider.py
simonrw: Since you have the two implementations, could you run some tests to work out how much slower the performance gets, and at what point it becomes prohibitive to use the `Parameters`?

maxhoheiser: As mentioned this will only take effect if the firehose `_put_records` API is used to put multiple records to Firehose, this is not an AWS supported method to interact with Firehose directly. If Firehose receives messages from other services, each message is put individually to Firehose, therefore only a single row will be written each time. 

Considering this, `Parameters` are the better choice in the end.

simonrw (approved):
Great, thanks for making my suggestions. I think this is an exciting addition to our service integrations!

PR #10379, opened 2024-03-04
Support custom key material when creating KMS keys

### LINKED ISSUE #8910: enhancement request: Support preseeding the new KMS provider

### Enhancement description

It would be great if the new KMS provider could be pre-seeded with actual key material.

The former `local-kms` provider support pre-seeding, but also lacked several endpoints. While new provider recently [added support for custom key IDs](https://github.com/localstack/localstack/commit/82bc739e2286c2a29bfcdf9c4e6ad3856adb9694), there are scenarios where this is not sufficient.

In particular, when developing against web3 smart contracts, having static key material for testing can be necessary to establish trust between local services and the blockchain.

Unfortunately, using `local-kms` isn't an option due to the lack of HMAC support.

### 🧑‍💻 Implementation

Ideally the key material could be loaded from a local file mounted into the localstack container, similar to how `local-kms` worked. Another option could be an API that can be called via an initialization hook.

### Anything else?

_No response_

jalaziz:
## Motivation

Add support for creating KMS keys with custom key material using custom tags. This is similar to an existing feature that allows the KMS Key ID to be set when creating KMS keys. This is necessary when working with external systems and decryption or HMACs must work across localstack reboots.

Fixes #8910

## Changes

* Introduces a new special tag `_custom_key_material_` that can be used when creating a KMS key to set the key material of the key.
* Refactors the existing `_custom_id_` so that it handled in the `KmsKey` model.

## Testing

I've introduced two new unit tests to ensure the custom key material is working for symmetric and HMAC KMS keys.

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -108,6 +108,51 @@ class TestKMS:
         assert result["KeyId"] == key_id
         assert result["Arn"].endswith(f":key/{key_id}")
 
+    @markers.aws.only_localstack
+    def test_create_key_custom_key_material_hmac(self, kms_create_key, aws_client):
+        custom_key_material = base64.b64encode(b"custom test key material").decode("utf-8")
+        key_spec = "HMAC_256"
+        mac_algo = "HMAC_SHA_256"
+        expected_mac = b"}\xa2\x95\xb8\xac\xc9*$\xe7!G\xdb\xf2\xc7:8y\xb1\xcf\xdf\xcc[5\xdcA\xd0e\x98\x13d\x88\x8e"
+
+        key_id = kms_create_key(
+            KeySpec=key_spec,
+            KeyUsage="GENERATE_VERIFY_MAC",
+            Tags=[{"TagKey": "_custom_key_material_", "TagValue": custom_key_material}],
+        )["KeyId"]
+
+        mac = aws_client.kms.generate_mac(
+            KeyId=key_id,
+            Message="some important message",
+            MacAlgorithm=mac_algo,
+        )["Mac"]
+        assert mac == expected_mac
+
+        verify_mac_response = aws_client.kms.verify_mac(
+            KeyId=key_id,
+            Message="some important message",
+            MacAlgorithm=mac_algo,
+            Mac=expected_mac,
+        )
+        assert verify_mac_response["MacValid"]
+
+    @markers.aws.only_localstack
+    def test_create_key_custom_key_material_symmetric_decrypt(self, kms_create_key, aws_client):
+        custom_key_material = base64.b64encode(b"custom test key material").decode("utf-8")
+        algo = "SYMMETRIC_DEFAULT"
+        message = b"test message 123 !%$@ 1234567890"
+        encrypted_message = b"m\xdc\x8e\xb5\xc4\xa2\\\x9f\x02\xa3\xfd[+L2N\x06`\xe8F\x87\xa0\xebh\xac\xa0qR \xec\x8eh?\n\x84l\x8d6\x0e\xd8d75\xa313H\x1e\x16\xf9\x14\x83,onX\xc5\x9d\x82\x84\x1d\xe91\x9c:\xea\x08\xc6\xdbc\x1f\xdcB\xf4\x83R\x83\x885\xc7"
+
+        key_id = kms_create_key(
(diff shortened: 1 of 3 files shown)


jalaziz:

jalaziz:
recheck

jalaziz:
recheck

tests/aws/services/kms/test_kms.py
```diff
@@ -107,6 +107,51 @@ def test_create_key_custom_id(self, kms_create_key, aws_client):
...
+            Mac=expected_mac,
+        )
+        assert verify_mac_response["MacValid"]
+
+    @markers.aws.only_localstack
+    def test_create_key_custom_key_material_symmetric_decrypt(self, kms_create_key, aws_client):
+        custom_key_material = base64.b64encode(b"custom test key material").decode("utf-8")
+        algo = "SYMMETRIC_DEFAULT"
+        message = b"test message 123 !%$@ 1234567890"
+        encrypted_message = b"m\xdc\x8e\xb5\xc4\xa2\\\x9f\x02\xa3\xfd[+L2N\x06`\xe8F\x87\xa0\xebh\xac\xa0qR \xec\x8eh?\n\x84l\x8d6\x0e\xd8d75\xa313H\x1e\x16\xf9\x14\x83,onX\xc5\x9d\x82\x84\x1d\xe91\x9c:\xea\x08\xc6\xdbc\x1f\xdcB\xf4\x83R\x83\x885\xc7"
```
sannya-singal, line 143: I would suggest to use some encryption method to generate these bytes string rather than using a static value, it primarily gives a broader context to the test being added.

tests/aws/services/kms/test_kms.py
```diff
@@ -107,6 +107,51 @@ def test_create_key_custom_id(self, kms_create_key, aws_client):
         assert result["KeyId"] == key_id
         assert result["Arn"].endswith(f":key/{key_id}")
 
+    @markers.aws.only_localstack
+    def test_create_key_custom_key_material_hmac(self, kms_create_key, aws_client):
+        custom_key_material = base64.b64encode(b"custom test key material").decode("utf-8")
+        key_spec = "HMAC_256"
+        mac_algo = "HMAC_SHA_256"
+        expected_mac = b"}\xa2\x95\xb8\xac\xc9*$\xe7!G\xdb\xf2\xc7:8y\xb1\xcf\xdf\xcc[5\xdcA\xd0e\x98\x13d\x88\x8e"
```
sannya-singal, line 115: Can we give add some more context here, like rather than using a static value we can actually use some encoding method to generate these bytes which can be used by the below section of code.

sannya-singal (changes requested):

sannya-singal:
Hi @jalaziz, thanks for contributing to localstack 🚀. These are amazing set of changes 🙂👏

I have added few comments, rest LGTM!

tests/aws/services/kms/test_kms.py
jalaziz: Thanks for the feedback. I purposefully didn't want to use the kms encrypt calls because it wouldn't really test the static encryption key (the test could pass if the custom key was ignored).

That being said, I could generate the encrypted text directly instead to provide more context.

sannya-singal: I am not sure what you mean by generating encrypted text directly, can you explain this a bit more?

jalaziz: Sorry, what I mean is calling the cryptography library directly in the test.

The purpose of this change and the tests is to ensure that secrets encrypted with the custom secret can be decrypted by the same secret across localstack restarts.

To ensure this, ideally the encrypted cipher text is static or generated in a way that is very stable.

sannya-singal: Do you mean using something like:

`from cryptography.hazmat.primitives import hashes` 

to encrypt?

jalaziz: Yup exactly. I've just pushed a commit with the test changes.

sannya-singal:
@jalaziz thanks for fixing reviews! LGTM! I would request you to rebase your branch and then I can merge it! 🎉

jalaziz:
@sannya-singal done!

sannya-singal:
Thanks for fixing this @jalaziz 👏 🚀. Can you kindly request a re-review:
image
as this is blocking the merge. Thanks!

sannya-singal (approved):

Issue #10380, opened 2024-03-04
enhancement request: Support XML templating in API Gateway


shucoshuco:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Currently, API Gateway templating only supports JSON templates. There is a TODO comment in the code related to this enhancement.
Because they are the most common media types, I'd say it's an interesting option add also support for XML templating. 

### 🧑‍💻 Implementation

Two changes are required:
* Read the Accept header in order to decide which template we must apply to a response
* Add validation for XML response, the same as already exists for JSON responses

### Anything else?

Link to the TODO comment: https://github.com/localstack/localstack/blob/76292e1d01d3141cbedfc0aea9f7b36095d6a23a/localstack/services/apigateway/templates.py#L325

bentsku:
Closed via #10381, thanks again for your contribution!

PR #10381, opened 2024-03-04
Support XML responses in api gateway templating

shucoshuco:
## Motivation
https://github.com/localstack/localstack/issues/10380

## Changes
Read the Accept header of the HTTP request and use the corresponding template and validate the response (only XML support added)

## Testing
Define an API gateway with an XML template as response_template, and make an HTTP request with Accept: application/xml header. The response must be formatted as a valid XML instead of JSON.

## TODO

What's left to do:

- Add support to other media types
- **I can't label this PR as semver: patch**

### DIFF
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -371,8 +371,8 @@ class ResponseTemplates(Templates):
     @staticmethod
     def _validate_xml(content: str):
         """
-        Checks that the content received is a valid JSON.
-        :raise JSONDecodeError: if content is not valid JSON
+        Checks that the content received is a valid XML.
+        :raise xml.parsers.expat.ExpatError: if content is not valid XML
         """
         try:
             xmltodict.parse(content)
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -329,10 +329,9 @@ class ResponseTemplates(Templates):
         accept = api_context.headers.get("accept", APPLICATION_JSON)
         supported_types = [APPLICATION_JSON, APPLICATION_XML]
         media_type = accept if accept in supported_types else APPLICATION_JSON
-        if media_type not in response_templates:
+        if not (template := response_templates.get(media_type, {})):
             return response._content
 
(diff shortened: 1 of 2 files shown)


localstack/services/apigateway/templates.py
```diff
@@ -320,35 +322,61 @@ def render(self, api_context: ApiInvocationContext, **kwargs) -> Union[bytes, st
...
         # see https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html#selecting-mapping-templates
-        template = response_templates.get(APPLICATION_JSON, {})
-        if not template:
+        accept = api_context.headers.get("accept", APPLICATION_JSON)
+        supported_types = [APPLICATION_JSON, APPLICATION_XML]
+        media_type = accept if accept in supported_types else APPLICATION_JSON
+        if media_type not in response_templates:
             return response._content
 
+        template = response_templates.get(media_type, {})
```
bentsku, line 335: nit: we could have kept the previous logic
```suggestion
        if not (template := response_templates.get(media_type, {})):
            return response._content
```

bentsku (approved):
LGTM! This looks pretty good to me. 

As a note, we usually ask for AWS validated integration tests: https://docs.localstack.cloud/contributing/integration-tests/

However, I realize that in this case, we don't really a good model in place that you could take inspiration from and extend to validate the XML change.
The unit tests look really good, so I'll take it on myself to do a follow-up PR to add more tests for this. 
I know our behavior with request/response templates is not fully in parity, especially around select the first template, and we will need to work on that:
> When the Accept header is not specified in the request, API Gateway assumes that its default value is `application/json`. In this case, API Gateway selects an existing mapping template for `application/json` to map the response payload. If no template is defined for `application/json`, API Gateway selects the first existing template and uses it as the default to map the response payload. 

Anyway, this is a great contribution, thanks a lot! Just a minor comment but nothing blocking, we can merge as is.

localstack/services/apigateway/templates.py
shucoshuco: Fixed in 4219f062

Issue #10003, opened 2024-01-05
S3: Too much data for declared Content-Length


tustvold:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We are seeing intermittent failures in CI with a rather opaque error message - https://github.com/apache/arrow-rs/issues/5283

```
Traceback (most recent call last):
  File "/opt/code/localstack/localstack/http/asgi.py", line 548, in handle_http
    await response.write(packet)
  File "/opt/code/localstack/localstack/http/asgi.py", line 302, in write
    await self.send({"type": "http.response.body", "body": data, "more_body": True})
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/hypercorn/protocol/http_stream.py", line 179, in app_send
    await self.send(
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/hypercorn/protocol/h11.py", line 136, in stream_send
    await self._send_h11_event(h11.Data(data=event.data))
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/hypercorn/protocol/h11.py", line 240, in _send_h11_event
    data = self.connection.send(event)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/h11/_connection.py", line 512, in send
    data_list = self.send_with_data_passthrough(event)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/h11/_connection.py", line 545, in send_with_data_passthrough
    writer(event, data_list.append)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/h11/_writers.py", line 65, in __call__
    self.send_data(event.data, write)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/h11/_writers.py", line 91, in send_data
    raise LocalProtocolError("Too much data for declared Content-Length")
h11._util.LocalProtocolError: Too much data for declared Content-Length
```

I have never been able to reproduce this locally, so it could be something funny with the way Github Actions handles proxying for containers, but it is intermittent and doesn't correspond with any change to our code that I am aware of.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
[...]

PR #10384, opened 2024-03-04
add rolo to dev script

bentsku:
## Motivation
When debugging a nasty bug (#10003), I would like to sprinkle some print statement in rolo, and logging the actual data returned from the exception, to see if it's an S3 issue or not, and which one is failing exactly.

## Changes
Add rolo to the list of local project to mount


thrau (approved):
thanks for adding this!

Issue #10369, opened 2024-03-01
enhancement request: Addnew  common string functions to vtl renderer


shucoshuco:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

During the vtl template rendering, only three common string functions are imlemented: trim, toLowerCase and toUpperCase.

I'm trying to render templates with other functions (toString, contains and replaceAll) and rendering is failing.

### 🧑‍💻 Implementation

This can be achieved adding new methods to class ExtendedString into VtlTemplate. I'll create a PR to add this methods.

bentsku:
Closing, thanks again for implementing this! 🙏

PR #10391, opened 2024-03-05
Support new common string functions in velocity templating

shucoshuco:
## Motivation
Allow using of more string common functions in vtl: https://github.com/localstack/localstack/issues/10369

## Changes
Adds more common string methods to class ExtendedString

## Testing
Into ApiGateway service, we now can use functions toString(), contains() and replaceAll() in the velocity templates

## TODO

What's left to do:

- Many others string methods are still missing.
- **This PR should be labeled with semver: patch**

### DIFF
--- a/localstack/utils/aws/templating.py
+++ b/localstack/utils/aws/templating.py
@@ -66,6 +66,9 @@ class VtlTemplate:
         # add extensions for common string functions below
 
         class ExtendedString(str):
+            def toString(self, *_, **__):
+                return self
+
             def trim(self, *args, **kwargs):
                 return ExtendedString(self.strip(*args, **kwargs))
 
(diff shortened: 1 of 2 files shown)


bentsku (commented):
LGTM! This is pretty nice, thanks a lot for your contribution and the associated test!

  Maybe @whummer or @calvernaz could give the final sign off, as I'm pretty unfamiliar with airspeed and how we usually extend its functionality, as it seems `contains` and `toString` should have been already done with https://github.com/localstack/airspeed/pull/13. 

Also, it seems AWS can diverge from official VTL in some way, so we usually validate those fix with AWS as well. If any of you can guide us to validate this and maybe where to work on integrating more functions? Thanks!

whummer (approved):
LGTM, kudos for adding this contribution @shucoshuco ! 🚀 

> airspeed and how we usually extend its functionality, as it seems contains and toString should have been already done with https://github.com/localstack/airspeed/pull/13

@bentsku This is a bit of a tricky case, which is caused by the duality of our patches for airspeed's `__additional_methods__` and our `ExtendedString`.

The problem is that airspeed looks up the additional methods via an exact type lookup, i.e., the additional methods are only added for `str` but not for `ExtendedString`: https://github.com/localstack/airspeed/blob/7aced144607cf7094dcbbe36fd51bc8faa3b37e3/airspeed/operators.py#L637

Currently we need to keep both in sync, which is not ideal. My general feeling is that we should try to move away from `ExtendedString`, and migrate all our patches to `__additional_methods__` directly.

In any case, this is something we can tackle in a follow-up iteration, along with adding more snapshot tests to our airspeed repo. This PR looks good to merge for me!

Issue #10393, opened 2024-03-05
bug:  Unable to get-object from S3 Glacier Instant Retrieval


ostavnaas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Files stored in S3 GLACIER_IR cannot be retrieved. 
```
$ aws s3api head-object --bucket bucket --key Dockerfile 
{
    "AcceptRanges": "bytes",
    "LastModified": "Tue, 05 Mar 2024 13:05:49 GMT",
    "ContentLength": 192,
    "ETag": "\"3ed0a9b7a98391730c8d22f553306d7d\"",
    "ContentType": "binary/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {},
    "StorageClass": "GLACIER_IR"
}
$  aws s3api get-object --bucket bucket --key Dockerfile

An error occurred (InvalidObjectState) when calling the GetObject operation: The operation is not valid for the object's storage class
``` 

### Expected Behavior

aws s3api get-object --bucket bucket --key Dockerfile  Download
{
    "AcceptRanges": "bytes",
    "LastModified": "Tue, 05 Mar 2024 13:08:21 GMT",
    "ContentLength": 192,
    "ETag": "\"3ed0a9b7a98391730c8d22f553306d7d\"",
    "ContentType": "binary/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://bucket
    touch file
    awslocal s3api put-object --bucket bucket --key file --body file --storage-class GLACIER_IR
    awslocal s3api head-object --bucket bucket --key file
    awslocal s3api get-object --bucket bucket --key file  Download

### Environment

```markdown
- OS: ubuntu 22.04
- LocalStack: 3.1.1.dev
```

ostavnaas:
I believe `StorageClass.GLACIER_IR` does not belong in `ARCHIVES_STORAGE_CLASSES`  in [constants.py](https://github.com/localstack/localstack/blob/master/localstack/services/s3/constants.py#L55)

PR #10395, opened 2024-03-05
fix S3 GLACIER_IR not being an archive type

bentsku:
## Motivation
As reported in #10393, the `GLACIER_IR` is not an archive type and can be retrieved instantly.

## Changes
- update an existing test to validate the instant retrieval of storage classes or not
- remove `GLACIER_IR` from the archive types


macnev2013 (approved):
LGTM

PR #10399, opened 2024-03-05
Enhance parity for API Gateway deletion of wildcard method settings

whummer:
## Motivation

Enhance parity for API Gateway deletion of wildcard method settings. This issue was recently reported by a customer - Terraform is running this particular type of API requests when deleting API GW resources and stages, which is currently leading to errors (see screenshot below):

image

## Changes

* handle special logic to the provider to handle updates (for `op=remove`) for wildcard method settings (`*/*`)
* fix the default value for `throttlingBurstLimit`, which moto currently returns as `2000`, but AWS returns `5000`
* add a new snapshot test to cover the functionality
* refactor existing test and pull the creation of API/stage out into a `_create_api_with_stage` fixture, for better re-usability

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_common.py
+++ b/tests/aws/services/apigateway/test_apigateway_common.py
@@ -583,39 +583,53 @@ class TestDocumentations:
 
 
 class TestStages:
-    @markers.aws.validated
-    @markers.snapshot.skip_snapshot_verify(paths=["$..createdDate", "$..lastUpdatedDate"])
-    def test_create_update_stages(
+    @pytest.fixture
+    def _create_api_with_stage(
         self, aws_client, create_rest_apigw, apigw_add_transformers, snapshot
     ):
         client = aws_client.apigateway
 
-        # create API, method, integration, deployment
-        api_id, api_name, root_id = create_rest_apigw()
-        client.put_method(
-            restApiId=api_id, resourceId=root_id, httpMethod="GET", authorizationType="NONE"
-        )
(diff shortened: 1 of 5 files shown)


tests/aws/services/apigateway/test_apigateway_common.py
```diff
@@ -669,6 +683,48 @@ def test_create_update_stages(
...
+                patchOperations=[
+                    {"op": "remove", "path": "/*/*"},
+                ],
+            )
+            snapshot.match("update-stage-reset", response)
+
+        # attempt to delete wildcard method settings (should initially fail)
+        with pytest.raises(ClientError) as exc:
+            _delete_wildcard()
+        snapshot.match("delete-error", exc.value)
```
bentsku, line 709:
```suggestion
        snapshot.match("delete-error", exc.value.response)
```

not blocking, but this capture the proper exception structure

bentsku (approved):
LGTM! Nice catch, and very quick fix 👍 nice and clean fix, thanks a lot for jumping on this! 

Just a small comment about the exception matching but honestly not blocking, we should merge as is, everything is green! 🚀

tests/aws/services/apigateway/test_apigateway_common.py
whummer: Nice catch, 💯  thanks!

bentsku: might have to regenerate the snapshot though 😬

PR #10400, opened 2024-03-05
SFN: Allow ProgramError without details

bblommers:
## Motivation
Bugfix in how SFN handles an ErrorState without any details.

## Testing

Example template that I've used to verify this works against AWS:
```
{
  "StartAt": "ChoiceStateX",
  "States": {
    "ChoiceStateX": {
      "Type": "Choice",
      "Choices": [
        {
          "And": [{"Variable": "$.type", "StringEquals": "Public"}],
          "Next": "Public"
        }
      ],
      "Default": "DefaultState"
    },
    "Public": {"Type": "Pass", "End": true},
    "DefaultState": {
      "Type": "Fail"
    }
  }
}
```
Note the `DefaultState` at the end; it does not have an error or cause, and AWS will accept this (and simply return `'cause': None, 'error': None`)

## StackTrace

This is what it looks like without these fixes:
```
    File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
      self.run()
    File "/usr/lib/python3.10/threading.py", line 953, in run
      self._target(*self._args, **self._kwargs)
    File "/.../stepfunctions/backend/execution_worker.py", line 76, in _execution_logic
      self._exec_comm.terminated()
    File "/.../stepfunctions/backend/execution.py", line 68, in terminated
      self.execution.error = exit_program_state.error["error"]
  KeyError: 'error'
  
    warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

### DIFF
--- a/localstack/services/stepfunctions/backend/execution.py
+++ b/localstack/services/stepfunctions/backend/execution.py
@@ -67,8 +67,8 @@ class BaseExecutionWorkerComm(ExecutionWorkerComm):
             self.execution.exec_status = ExecutionStatus.ABORTED
         elif isinstance(exit_program_state, ProgramError):
             self.execution.exec_status = ExecutionStatus.FAILED
-            self.execution.error = exit_program_state.error["error"]
-            self.execution.cause = exit_program_state.error["cause"]
+            self.execution.error = exit_program_state.error.get("error")
+            self.execution.cause = exit_program_state.error.get("cause")
         elif isinstance(exit_program_state, ProgramTimedOut):
             self.execution.exec_status = ExecutionStatus.TIMED_OUT
         else:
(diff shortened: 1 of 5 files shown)


bblommers:

MEPalma:
Thank you @bblommers! Could we please have a snapshot test for this scenario

bblommers:
Hi @MEPalma! There is a test for this exact scenario, actually, so I decided to just extend the test to also capture the `describe_execution` call, as that is the operation that showcases the problem.

Let me know if you're OK with this approach - this change seemed the least invasive to me.

MEPalma (approved):
Thanks again!

PR #10402, opened 2024-03-06
Remove unused CFN_RESOURCE_PROVIDER_OVERRIDES option

dominikschubert:
## Motivation

While this technically removes an option, it was one that has not been communicated and was only used internally for configuring generated tests. Even that hasn't been relevant since we've switched to prioritizing newer resource providers over the legacy ones by default.

## Changes

- Remove config `CFN_RESOURCE_PROVIDER_OVERRIDES` and all usage


thrau (approved):
LGTM! if it wasn't documented as part of the public config API, no need to really treat it as a major change :+1:

PR #10408, opened 2024-03-06
add fix for cfn-response code for nodejs custome resource

pinzon:
## Motivation
Custom resource models backed by nodejs lambda are experiencing deployment issues as a result of the nodejs cfn-response file containing an additional line ending character within the container, and the request is directing traffic to port 443.

## Changes
- convert standard string to raw string to avoid extra \n character in the container code
- the request for the insertion of the resource deployment into the bucket uses url port not directly 443

## Testing
The custom resource are not able to be deployed on this project so there is no need for a test

### DIFF
--- a/localstack/services/lambda_/resource_providers/aws_lambda_function.py
+++ b/localstack/services/lambda_/resource_providers/aws_lambda_function.py
@@ -161,7 +161,7 @@ def send(event, context, responseStatus, responseData, physicalResourceId=None, 
 """
 
 # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html
-NODEJS_CFN_RESPONSE_CONTENT = """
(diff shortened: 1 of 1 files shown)


localstack/services/lambda_/resource_providers/aws_lambda_function.py
```diff
@@ -189,7 +189,7 @@ def send(event, context, responseStatus, responseData, physicalResourceId=None,
     var parsedUrl = url.parse(event.ResponseURL);
     var options = {
         hostname: parsedUrl.hostname,
-        port: 443,
+        port: parsedUrl.port,
```
dominikschubert, line 192: Careful with that. Please at least add a comment explaining that this is a change that we made (and why), otherwise it's going to be hard to track any changes we've made to the official cfn-response code.

dominikschubert (approved):
LGTM, in terms of testing this is actually covered in -ext. I think currently we only cover the python one, but we should extend this to have both a python-based as well as a NodeJS based custom resource test :+1:

PR #10410, opened 2024-03-06
fix entrypoint discovery for Pro tests

bentsku:
## Motivation
The current Pro workflow actually runs with Pro not activated, as the entrypoints were not discovered (so just running twice our community tests)

## Changes
TBD


bentsku:
Closing as the fix will most probably not live here.

Issue #10392, opened 2024-03-05
StartMessageMoveTask fails when optional destination ARN is not supplied


iamvinov-atlassian:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently performing `StartMessageMoveTask` fails with `500 Internal Server Error` when destination queue ARN is not supplied. According to AWS [docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_StartMessageMoveTask.html), `destinationArn` is not required. 

Logs:
```
localstack.request.aws     : AWS sqs.StartMessageMoveTask => 500 (InternalError)
2024-03-04T16:12:38.862 DEBUG --- [   asgi_gw_0] rolo.gateway.wsgi          : POST 127.0.0.1:4566/
2024-03-04T16:12:38.863 ERROR --- [   asgi_gw_0] l.aws.handlers.logging     : exception during call chain
Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 112, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack/aws/handlers/service.py", line 82, in __call__
    skeleton_response = self.skeleton.invoke(context)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(serializer, context, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 168, in dispatch_request
    result = handler(context, instance) or {}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/sqs/provider.py", line 1402, in start_message_move_task
    self._require_queue_by_arn(context, destination_arn)
  File "/opt/code/localstack/localstack/services/sqs/provider.py", line 792, in _require_queue_by_arn
[...]

PR #10414, opened 2024-03-07
implement SQS DLQ redrive to original source queue

thrau:
## Motivation

A user correctly noted in #10392 that the `StartMessageMoveTask` doesn't need a destination arn. [The doc says](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_StartMessageMoveTask.html):

> If this field is left blank, the messages will be redriven back to their respective original source queues.

This PR adds this behavior.

The basic case was easy enough, but the plural in that sentence ("source **queues**") turns out to be a bit harder to implement. Messages are split up and returned to their original source queues, which means there is some sort of tracking where the messages originate from. We have no such tracking, and I didn't have enough time to implement this in this iteration. Left a TODO and a test case.

## Changes

* `DestinationArn` in SQS `StartMessageMoveTask` operation is now optional
* There is a skipped test case for the multi source queue case which will be a bit trickier to implement

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -205,7 +205,9 @@ class MessageMoveTask:
 
     # configurable fields
     source_arn: str
-    destination_arn: str
+    """The arn of the DLQ the messages are currently in."""
+    destination_arn: str | None = None
+    """If the DestinationArn is not specified, the original source arn will be used as target."""
     max_number_of_messages_per_second: int | None = None
 
     # dynamic fields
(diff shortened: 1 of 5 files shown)


baermat (approved):
LGTM :+1:
We can merge this for now, my personal naive implementation of the multi source problem would be this:
This is only supported for DLQs, and we know where we move to DLQ -> add an internal field to sqs messages `dlq_origin` or something similar. Set this when moved to DLQ. Then retrieve the source from the message itself.
Of course there are still things to check (e.g. can we move from one DLQ to another? What's the source then? Is the source deleted when moved? etc.), but we could deal with the edge cases when they come in.
Or this is completely unfeasible due to some unconsidered side-effects. Simply food for thought 🙂

PR #10416, opened 2024-03-08
make SQS dynamic endpoint strategy guess the endpoint strategy used

thrau:
## Motivation

After a lengthy follow-up discussion of #9753, we decided to extend the "dynamic" SQS endpoint strategy to also guess path, standard, or domain strategy of the given host URL.

When starting localstack with `SQS_ENDPOINT_STRATEGY=dynamic`, it will now behave as follows:

* path (fallback)
  ```
   % awslocal sqs list-queues                     
  {
      "QueueUrls": [
          "http://localhost:4566/queue/us-east-1/000000000000/foobar"
      ]
  }
  ```
* standard (when using `sqs.<region>.<host>`)
  ```
   % aws --endpoint-url=https://sqs.us-east-1.localhost.localstack.cloud:4566 sqs list-queues
  {
      "QueueUrls": [
          "https://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/foobar"
      ]
  }
  ```
  :warning: note that, using `--endpoint-url=https://sqs.localhost.localstack.cloud:4566` may not behave as expected, because we cannot for certain say that `localhost` is not a region, we must interpret as such, and therefore the host part will be `localstack.cloud`.
* domain
  ```
   % aws --endpoint-url=https://queue.localhost.localstack.cloud:4566 sqs list-queues
  {
      "QueueUrls": [
          "https://queue.localhost.localstack.cloud:4566/000000000000/foobar"
      ]
  }
  ```
  (note how us-east-1 is the default region, and therefore removed from the URL)
  ```
   % aws --endpoint-url=http://us-east-1.queue.localhost.localstack.cloud:4566 sqs list-queues 
  {
      "QueueUrls": [
          "http://queue.localhost.localstack.cloud:4566/000000000000/foobar"
      ]
  }
  ```

## Changes

* `SQS_ENDPOINT_STRATEGY=dynamic` now dynamically and heuristically determines the endpoint strategy based on the incoming host and renders queue URLs accordingly

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -32,6 +32,7 @@ from localstack.services.sqs.utils import (
     encode_move_task_handle,
     encode_receipt_handle,
     global_message_sequence,
+    guess_endpoint_strategy_and_host,
     is_message_deduplication_id_required,
 )
 from localstack.services.stores import AccountRegionBundle, BaseStore, LocalAttribute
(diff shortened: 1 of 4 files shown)


localstack/services/sqs/utils.py
```diff
@@ -34,6 +34,37 @@ def is_sqs_queue_url(url: str) -> bool:
...
+    components = host.split(".")
+
+    if host.startswith("sqs."):
+        return "standard", ".".join(components[2:])
+
+    if host.startswith("queue."):
+        return "domain", ".".join(components[1:])
+
+    if len(components) > 2 and components[1] == "queue":
+        return "domain", ".".join(components[2:])
```
baermat, line 63: I am probably forgetting something, but why len > 2? The examples listed above would suggest at least len > 3 (`<region>.queue.(amazonaws | localhost.localstack).com`). Or is it to avoid a IndexError? but then > 1 is enough :)

tests/unit/test_sqs.py
```diff
@@ -175,3 +176,20 @@ def test_is_sqs_queue_url():
     assert is_sqs_queue_url("http://us-east-1.queue.foo.bar:4566/111111111111/foo") is True
     assert is_sqs_queue_url("http://queue.foo.bar:4566/111111111111/foo") is True
     assert is_sqs_queue_url("http://sqs.us-east-1.foo.bar:4566/111111111111/foo") is True
+
```
baermat, line 179: It would probably good if we could have integration tests for this as well. Not blocking for now, but having tests that set the endpoint like our customers might become necessary at some point

baermat (commented):
I was curious how this would be tackled :)  A couple of comments that are non-blocking. After that, and confirming that the failing tests are unrelated, it's a go from me.
- This would benefit very much from an accompanging docs update :) 
  - The guessing excludes the "off" option of the strategies. We should mention that in the docs as well.
- Considering QoL for our customers, will `awslocal --endpoint=<new_endpoint>` also work? Or will awslocal overwrite the endpoint in this case?

localstack/services/sqs/utils.py
thrau: perhaps to allow `us-west-1.queue.localhost` but not `us-west-1.queue`. although it's not clear to me why anyone would name their host `queue`, but not unthinkable :D

thrau:
thanks for the review @baermat! all great points. i'll wait for the user's final confirmation and then go ahead and document it

localstack/services/sqs/models.py
```diff
@@ -326,34 +327,34 @@ def url(self, context: RequestContext) -> str:
...
-        if config.SQS_ENDPOINT_STRATEGY == "standard":
+        endpoint_strategy = config.SQS_ENDPOINT_STRATEGY
+
+        if endpoint_strategy == "dynamic":
+            scheme = context.request.scheme
+            endpoint_strategy, host_and_port = guess_endpoint_strategy_and_host(
+                context.request.host
+            )
+
+        if endpoint_strategy == "standard":
```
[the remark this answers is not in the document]
thrau: actually the previous block re-configures the endpoint strategy, and then re-uses the blocks below to build the correct url. i left a comment

PR #10417, opened 2024-03-08
update s3 lambdas to node20

bentsku:
## Motivation
~It seems the pipeline just failed with some missing dependency for some lambdas used in S3: `fast-xml-parser`. Not sure what produced this, maybe it's not a default dependency anymore.~

Edit: ah, well, they updated it. https://github.com/aws/aws-sdk-js-v3/pull/5869
It seems they released a version with an issue, so this is not blocking anymore. I'll remove the added dependency but updating the lambda can't be bad. 

## Changes
~Added the dependency to the list of install.~
I've also spotted that Node14 is not supported anymore in lambda, so I've updated the lambda used to Node 20 + to use ES6.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -3612,7 +3612,7 @@ class TestS3:
         )
         temp_folder = create_tmp_folder_lambda(
             handler_file,
-            run_command="npm i @aws-sdk/util-endpoints @aws-sdk/client-s3 @aws-sdk/s3-request-presigner @aws-sdk/middleware-endpoint fast-xml-parser",
+            run_command="npm i @aws-sdk/util-endpoints @aws-sdk/client-s3 @aws-sdk/s3-request-presigner @aws-sdk/middleware-endpoint",
         )
 
         function_name = f"func-integration-{short_uid()}"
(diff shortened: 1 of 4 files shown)


tests/aws/services/s3/test_s3.py
```diff
@@ -3608,18 +3608,18 @@ def test_s3_lambda_integration(
     ):
         snapshot.add_transformer(snapshot.transform.s3_api())
         handler_file = os.path.join(
-            os.path.dirname(__file__), "../lambda_/functions/lambda_s3_integration.js"
+            os.path.dirname(__file__), "../lambda_/functions/lambda_s3_integration.mjs"
         )
         temp_folder = create_tmp_folder_lambda(
             handler_file,
-            run_command="npm i @aws-sdk/util-endpoints @aws-sdk/client-s3 @aws-sdk/s3-request-presigner @aws-sdk/middleware-endpoint",
+            run_command="npm i @aws-sdk/util-endpoints @aws-sdk/client-s3 @aws-sdk/s3-request-presigner @aws-sdk/middleware-endpoint fast-xml-parser",
```
dfangl, line 3615: Should we merge this change if they correct it upstream anyway?

dfangl (approved):
Thanks for upgrading the lambdas, we should definitely refrain from using deprecated runtimes in our tests.
Only one question about adding `fast-xml-parser`, since it should be fixed upstream?

Issue #10325, opened 2024-02-26
bug: Lambda Event Source Mapping Filter Criteria not respecting comparison operator exists: false


Chad-Bailer-Regal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

While using Lambda event source mappings with Dynamodb Streams there appears to be a bug surrounding filter criteria. It appears that the comparison operator exists doesn't work as intended when it's false. It seems to work as intended when set to true, but when false and the property doesn't exist in the item it doesn't seem to trigger. 

There's another bug as well with event source mappings where it doesn't appear that SAM deploy will update the filter criteria in the event source mapping when there's a change. We can easily test this by making a change to the FilterCriteria in the SAM template and then redeploying. When we view the event source mappings it'll still have the old value. Interestingly `awslocal lambda update-event-source-mapping` does seem to update it while SAM will not.

### Expected Behavior

* When filter criteria includes `"exists": false` and the item added in does not include that property it should trigger the lambda.

* After updating the filter criteria and running a sam deploy it should modify the filter criteria which should be visible when running `list-event-source-mappings` 

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up 
    
#### template.yaml
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  localstack-dynamostream-helloworld

  Sample SAM Template for localstack-dynamostream-helloworld

# More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
Globals:
  Function:
    Timeout: 3
    MemorySize: 128

Resources:
  StreamsSampleDDBTable:
    Type: AWS::DynamoDB::Table
    Properties:
      TableName: localstack-dynamostream-helloworld
      AttributeDefinitions:
        - AttributeName: "PK"
          AttributeType: "S"
        - AttributeName: "SK"
          AttributeType: "S"
      KeySchema:
        - AttributeName: "PK"
          KeyType: "HASH"
        - AttributeName: "SK"
          KeyType: "RANGE"
      StreamSpecification:
        StreamViewType: "NEW_AND_OLD_IMAGES"
      ProvisionedThroughput:
        ReadCapacityUnits: 5
        WriteCapacityUnits: 5

  LambdaExecutionRole:
    Type: AWS::IAM::Role
    Properties:
[...]

PR #10418, opened 2024-03-08
fix lambda event source mapping FilterCriteria

sannya-singal:
## Motivation
While using Lambda event source mappings with Dynamodb Streams there was a bug reported in #10325 surrounding `FilterCriteria`. It appears that the comparison operator `exists` doesn't work as intended when it's false. When `exists` is set to false and the property doesn't exist in the item the lambda doesn't trigger.

For instance in the [AWS tutorial to process events with Lambda and DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.Tutorial2.html), when we use `[{"homemade": {"S": [{"exists": false}]}}]` the lambda doesn't trigger when the property is missing in dynamodb `PutItem` call. 

## Changes
This PR:
- Adds a condition in while filtering the stream record to handle special case of exists when the value is a `dict`. 
- Adds a test case to validate the fix. 

This closes first part of #10325 bug.

### DIFF
--- a/localstack/services/lambda_/event_source_listeners/utils.py
+++ b/localstack/services/lambda_/event_source_listeners/utils.py
@@ -45,28 +45,28 @@ def filter_stream_record(filter_rule: Dict[str, any], record: Dict[str, any]) ->
                 append_record = filter_stream_record(value, record_value)
         else:
             # special case 'exists'
-            def filter_rule_value_list(val):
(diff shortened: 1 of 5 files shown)


pinzon (approved):
LGTM :+1:

PR #10425, opened 2024-03-08
fix SQS DLQ message attributes and redrive into multiple source queues

thrau:
## Motivation

As discussed in #10414, SQS DLQ redrive returns the messages back to their original source when no destination ARN is specified. To implement that, "some sort of tracking" is needed, and it turns out that AWS already has an attribute for that, namely the `DeadLetterQueueSourceArn` (which of course is not really documented anywhere...).

Basically what @baermat suggested here:
>  This is only supported for DLQs, and we know where we move to DLQ -> add an internal field to sqs messages dlq_origin or something similar. Set this when moved to DLQ. Then retrieve the source from the message itself.

which is exactly how it's implemented now.

A necessary fix was to make sure that the `message["Attributes"]` dict is retained correctly when it's being passed around through the queues. I was also reminded of [one of the limitations ](https://docs.localstack.cloud/user-guide/aws/sqs/#limitations), that `ApproximateReceiveCount` will be reset to 0 when a message moves to a DLQ. The problem was very similar so I fixed that along the way.

I briefly tested what happens when you delete one of the original source queues while the move task is running, and it simply moves into a `FAILED` state with a QueueNotFound error. So it's no different from the `test_move_task_delete_destination_queue_while_running` test case we already have.

At some point we should clean up the message system attribute handling, and also re-think whether `SqsQueue.put(Message, ...)` is the correct signature, or whether it should be `SqsQueue.put(SqsMessage)`.

## Changes

* `StartMessageMoveTask` will now correctly route messages back to their original source queues
* `ApproximateReceiveCount` is now correctly counted when the messages moves through DLQs
* Other message attributes are now also retained when messages move through DLQs (including the MessageID)

## Other notes

* The TODO in `test_dead_letter_queue_execution_lambda_mapping_preserves_id` is now obsolete, but something else is wrong with the test. there have been so many refactorings of it but it was never run because it's being skipped.

### DIFF
--- a/tests/aws/services/sqs/test_sqs.py
+++ b/tests/aws/services/sqs/test_sqs.py
@@ -2830,8 +2830,21 @@ class TestSqsProvider:
         snapshot.match("dlq-arn", dl_queue_arn)
         snapshot.match("sourcen-arn", sqs_get_queue_arn(queue_url))
 
+        # check that attributes are retained
+        msg_attrs = {"MyAttribute": {"StringValue": "foobar", "DataType": "String"}}
+        msg_system_attrs = {
+            "AWSTraceHeader": {
+                "StringValue": "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1",
+                "DataType": "String",
+            }
+        }
         # send a messages
-        sqs.send_message(QueueUrl=queue_url, MessageBody="message-1")
+        sqs.send_message(
+            QueueUrl=queue_url,
+            MessageBody="message-1",
+            MessageSystemAttributes=msg_system_attrs,
+            MessageAttributes=msg_attrs,
+        )
         sqs.send_message(QueueUrl=queue_url, MessageBody="message-2")
 
         # receive each message two times to move them into the dlq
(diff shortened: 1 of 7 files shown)


tests/aws/services/sqs/test_sqs.py
```diff
@@ -2802,6 +2804,80 @@ def test_dead_letter_queue_max_receive_count(
...
+
+        # check that attributes are retained
+        msg_attrs = {"MyAttribute": {"StringValue": "foobar", "DataType": "String"}}
+        msg_system_attrs = {
+            "AWSTraceHeader": {
+                "StringValue": "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1",
+                "DataType": "String",
+            }
+        }
+        # send a messages
```
baermat, line 2841: nit: typo

baermat (approved):
Looks good to me! I have nothing to add besides a pedantic typo :smile: Also not having to iterate over the store for the destination resolving is very nice :tada:

tests/aws/services/sqs/test_sqs.py
thrau: i want to do a spring cleaning of the SQS code base (including comment formatting, doc strings, maybe some light refactoring), and will clean up the comments as well :-)

PR #10427, opened 2024-03-10
[SFN] Reduce Flakiness of Timeout and Heartbeat Tests

MEPalma:
## Motivation
We've noticed some inconsistency in the StepFunctions tests, particularly related to timeout exceptions. These issues arise due to the strict 1-second timeout value. Consequently, there are instances where the interpreter doesn't have sufficient time to log the events for execution start and scheduling. While this behavior aligns with expectations, our focus lies in validating the state machine's sequence of events. We're particularly interested in testing the scenario where the state returns early.

This pull request aims to address these inconsistencies by mitigating the flakiness of these tests. The proposed solution, for now, involves increasing the timeout or heartbeat time allocated to the states.

## Changes
- Increased Timeout/Heartbeat value
- Recomputed relevant snapshots


dominikschubert (approved):
LGTM, should make the test less flaky for now. Let's investigate why the 1s timeout isn't enough in certain cases when we tackle the stepfunctions performance analysis in the next few months.

PR #10430, opened 2024-03-11
fix SNS Publish exception when using empty topic

bentsku:
## Motivation
Found out this small issue when launching a benchmark without the initial setup. When calling `SNS.Publish(Message="test", TopicArn="")`, boto accepts it but it fails on LocalStack side. 

## Changes
- add a test for the issue
- add a small check to use an empty string if the value is falsy


thrau (approved):
LGTM!

PR #10432, opened 2024-03-11
Move to only using `pyproject.toml`

silv-io:
## Motivation

Using only `pyproject.toml` to define project metadata and dependencies is the new preferred way of doing things. Future distributions will only use `pyproject.toml` instead of a combination of `pyproject.toml`, `setup.py` and `setup.cfg`. Having all those things homogenized is helpful for introducing a multi-distribution project.

## Changes

- Remove `setup.cfg` and `setup.py` (and all references thereof) in favor of storing all this information in `pyproject.toml` 
- To define relationships between optional dependencies, we now don't use the `%(extra)s` syntax from the `setup.cfg` but [recursively reference the package itself](https://discuss.python.org/t/pyproject-toml-optional-dependencies-redundancy-aka-dry-extras/8428/9?u=flying-sheep) (so if extra `test` depends on extra `runtime`, extra `test` needs to contain `"localstack-ext[runtime]"`). This means that the package itself needs to be marked as unsafe for the pinning of dependencies.
- Upgrade all pinned dependencies, because the `build` package needs to be added, and the unsafe packages need to be adapted.
- Use the `build` package for building distributions instead of `setup.py`
- Use `plux` for creating the entrypoints instead of `setup.py`
- Update pre-commit hooks (adding support for `pyproject.toml` in pinned dependency checker and align black version with pinned version)
- _**Note**_: Beforehand the `setup.cfg` defined the minimum Python version. This got transferred to the `pyproject.toml`. `black` beforehand did not properly deduce from the `setup.cfg` that it should only produce 3.8 compatible code. With this information now being in the `pyproject.toml` it does. For this reason some of the existing code needs to be reformatted (every single generated AWS API)

## Testing

The pipeline should be green and the produced artifacts should remain equivalent

## TODO

What's left to do:

- [x] Green pipeline


pyproject.toml
```diff
@@ -3,6 +3,156 @@
...
+
+[project.optional-dependencies]
+# minimal required to actually run localstack on the host for services natively implemented in python
+base-runtime = [
+    "awscrt>=0.13.14",
+    "boto3>=1.26.121",
+    "botocore==1.34.54",
+    "cbor2>=5.2.0",
+    "dnspython>=1.16.0",
+    # TODO tag incompatibility introduced in 7.0.0 with https://github.com/docker/docker-py/pull/3191
```
bentsku, line 55: it seems `localstack-twisted` vanished from `setup.cfg` and `pyproject.toml`? 

https://github.com/localstack/localstack/blob/0b9c6b1de4be4c8d24f52d6ba863f7085e0d0741/setup.cfg#L71

silv-io: Should be resolved now after rebasing

bin/release-helper.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 VERSION_FILE=${VERSION_FILE-*/__init__.py}
-DEPENDENCY_FILE=${DEPENDENCY_FILE-setup.cfg}
+DEPENDENCY_FILE=${DEPENDENCY_FILE-pyproject.toml}
```
alexrashed, line 6: q: Have you tested that the usages of this variable also work for the `pyproject.toml` (and not only for `*.cfg` files)?

pyproject.toml
```diff
@@ -90,3 +248,6 @@ log_cli = true
 log_level = "DEBUG"
 log_cli_format = "%(asctime)s.%(msecs)03d %(levelname)5s --- [%(threadName)12s] %(name)-26s : %(message)s"
 log_cli_date_format = "%Y-%m-%dT%H:%M:%S"
+
+[tool.pip-tools]
+unsafe-package = ["localstack-core", "pip", "setuptools", "distribute"] # packages that should not be pinned
```
alexrashed, line 253: A comment here would be great, since this looks a bit weird:
- Why is there pip, setuptools, distribute (because it's the default)
- Why is there localstack-core (because of the extra stacking)

pyproject.toml
```diff
@@ -3,6 +3,164 @@
 requires = ['setuptools', 'wheel', 'plux>=1.7']
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "localstack-core"
+authors = [
+        { name = "LocalStack Contributors", email = "info@localstack.cloud" }
+]
+description = "The core library and runtime of LocalStack"
+requires-python = ">=3.8"
```
alexrashed, line 12: This change is the one which makes the second commit (adding all these trailing commas) necessary:
- `black` did not read this from the `setup.cfg`: https://github.com/psf/black/issues/3859#issuecomment-1699370363
- `black` adds a trailing comma for `kwargs` for 3.6+: https://test-black.readthedocs.io/en/style-guide/style_guide/trailing_commas.html

alexrashed (approved):
Awesome! This is quite a step forward towards up-to-date best-practices with python project config / packaging! 🚀 
I only had a few questions (basically just for comments), afterwards we are good to go! 🚀 💯 🧹

PR #10433, opened 2024-03-11
Fix DynamoDB Table and Global Table Creation with SSESpecification and TTL

Morijarti:
## Motivation
When table is created with following attributes
```
SSESpecification:
    SSEEnabled: True
    SSEType: AES256
TimeToLiveSpecification:
    AttributeName: expire_at
    Enabled: true
```

SSESpecification will cause crash, while TTL is ignored. This PR fixes this

## Changes
 Updated both Table and GlobalTable CFn resource provider to properly pass parameters from CFn Template

## Testing
TODO: Add tests

### DIFF
--- a/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
+++ b/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
@@ -268,8 +268,7 @@ class DynamoDBGlobalTableProvider(ResourceProvider[DynamoDBGlobalTableProperties
             # add TTL config
             if ttl_config := model.get("TimeToLiveSpecification"):
                 request.aws_client_factory.dynamodb.update_time_to_live(
-                    TableName=model["TableName"],
-                    TimeToLiveSpecification=ttl_config
+                    TableName=model["TableName"], TimeToLiveSpecification=ttl_config
                 )
 
             return ProgressEvent(
--- a/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
+++ b/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
@@ -229,12 +229,16 @@ class DynamoDBGlobalTableProvider(ResourceProvider[DynamoDBGlobalTableProperties
                     "SSESpecification",
                     "StreamSpecification",
                     "TableName",
-                    "TimeToLiveSpecification",
                     "WriteProvisionedThroughputSettings",
                 ],
             )
 
             replicas = create_params.pop("Replicas", [])
+
+            if sse_specification := create_params.get("SSESpecification"):
+                # rename bool attribute to fit boto call
+                sse_specification["Enabled"] = sse_specification["SSEEnabled"]
+                del sse_specification["SSEEnabled"]
 
             creation_response = request.aws_client_factory.dynamodb.create_table(**create_params)
             model["Arn"] = creation_response["TableDescription"]["TableArn"]
@@ -259,6 +263,13 @@ class DynamoDBGlobalTableProvider(ResourceProvider[DynamoDBGlobalTableProperties
 
                 request.aws_client_factory.dynamodb.update_table(
                     ReplicaUpdates=replicas_to_create, TableName=model["TableName"]
+                )
+
+            # add TTL config
+            if ttl_config := model.get("TimeToLiveSpecification"):
+                request.aws_client_factory.dynamodb.update_time_to_live(
+                    TableName=model["TableName"],
+                    TimeToLiveSpecification=ttl_config
                 )
 
             return ProgressEvent(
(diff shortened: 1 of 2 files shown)


Morijarti:
I did notice an interesting behaviour in CFn, if template for `GlobalTable` contains
```
      ProvisionedThroughput:
        ReadCapacityUnits: 1
        WriteCapacityUnits: 1
```

Which are Table only attributes, it will fail on AWS. Should we also add check if template contains extra parameters for resource? This would also require us to keep schema always up-to-date.

For now I didn't address this behavior, just made sure templates are valid and working on AWS and LS

localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
```diff
@@ -229,13 +229,17 @@ def create(
...
                     "WriteProvisionedThroughputSettings",
                 ],
             )
 
             replicas = create_params.pop("Replicas", [])
 
+            if sse_specification := create_params.get("SSESpecification"):
+                # rename bool attribute to fit boto call
+                sse_specification["Enabled"] = sse_specification["SSEEnabled"]
+                del sse_specification["SSEEnabled"]
```
dominikschubert, line 241:
```suggestion
                sse_specification["Enabled"] = sse_specification.pop("SSEEnabled")
```

localstack/services/dynamodb/resource_providers/aws_dynamodb_table.py
```diff
@@ -214,9 +214,15 @@ def create(
...
                 "GlobalSecondaryIndexes",
                 "Tags",
+                "SSESpecification",
             ]
             create_params = util.select_attributes(model, properties)
 
+            if sse_specification := create_params.get("SSESpecification"):
+                # rename bool attribute to fit boto call
+                sse_specification["Enabled"] = sse_specification["SSEEnabled"]
+                del sse_specification["SSEEnabled"]
```
dominikschubert, line 224:
```suggestion
                sse_specification["Enabled"] = sse_specification.pop("SSEEnabled")
```

same as above

tests/aws/services/cloudformation/resources/test_dynamodb.py
```diff
@@ -163,3 +163,45 @@ def test_ttl_cdk(aws_client, snapshot, infrastructure_setup):
...
         table = aws_client.dynamodb.describe_time_to_live(TableName=table_name)
         snapshot.match("table", table)
+
+
+@markers.aws.validated
+@markers.snapshot.skip_snapshot_verify(
+    [
+        "$..Table.ProvisionedThroughput.LastDecreaseDateTime",
+        "$..Table.ProvisionedThroughput.LastIncreaseDateTime",
+        "$..Table.Replicas",
```
dominikschubert, line 173: I'd encourage adding a comment why these are skipped. Is it because LocalStack returns those keys but AWS doesn't?

tests/aws/services/cloudformation/resources/test_dynamodb.py
```diff
@@ -163,3 +163,45 @@ def test_ttl_cdk(aws_client, snapshot, infrastructure_setup):
...
+        "$..Table.SSEDescription.KMSMasterKeyArn",
```
dominikschubert, line 174: This should be a transformer, rather than a skip here. If I execute this in my account, this test will not pass, since anything in  `skip_snapshot_verify` is disabled when testing against AWS.

tests/aws/services/cloudformation/resources/test_dynamodb.py
```diff
@@ -163,3 +163,45 @@ def test_ttl_cdk(aws_client, snapshot, infrastructure_setup):
...
+    snapshot.match("table_description", response)
+
+
+@markers.aws.validated
+@markers.snapshot.skip_snapshot_verify(
+    [
+        "$..Table.ProvisionedThroughput.LastDecreaseDateTime",
+        "$..Table.ProvisionedThroughput.LastIncreaseDateTime",
+        "$..Table.Replicas",
+        "$..Table.SSEDescription.KMSMasterKeyArn",
```
dominikschubert, line 195: same arguments as above apply here :+1:

dominikschubert (changes requested):
Mostly LGTM!

I'm curious if some of the table operations we're doing on create wouldn't need to also wait for the active state when executed on AWS. For LS this probably isn't an issue right now but it's something we might need to change in the future.

Some minor nits in code, but one required change in the tests (`KMSMasterKeyArn`)

Morijarti:
> I'm curious if some of the table operations we're doing on create wouldn't need to also wait for the active state when executed on AWS. For LS this probably isn't an issue right now but it's something we might need to change in the future.

I think you are right, my intuition tells me that we should wait, but we can test it in future and see how thing behave.

tests/aws/services/cloudformation/resources/test_dynamodb.py
Morijarti: We return them, but AWS doesn't. I'll add the comment in the code

tests/aws/services/cloudformation/resources/test_dynamodb.py
Morijarti: Fixed

tests/aws/services/cloudformation/resources/test_dynamodb.py
Morijarti: Fixed

dominikschubert (approved):
LGTM, thanks for addressing all the mentioned issues :pray:

Issue #7466, opened 2023-01-09
feature request: ability to specific Lambda hot reload path glob


davecardwell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

I would like the ability to specify one or more “patterns” that limit which file changes trigger a hot reload with LocalStack Lambdas. For example, this could be implemented to accept a comma-separated list of Python standard library [glob](https://docs.python.org/3/library/glob.html)s:
```
LAMBDA_HOT_RELOAD_PATTERNS=**/dist/**/*.js
```

Using this setting, only changes to `.js` files under a `dist` folder would cause a hot reload.

---

I have a Node.js monorepo which includes workspaces which depend on each other. pnpm creates symlinks under one workspace’s `node_modules` folder to another, for example `packages/a/node_modules/@myapp/b` is a symlink to `../../../../packages/b`.

If I want to launch a lambda with code under `packages/a` I have to set the s3 key in my `hot-reload` bucket to the project root, otherwise the symlink to `../../../../packages/b` won’t resolve as that file won’t exist under the lambda container `/var/task` directory.

Setting the S3 key to the project root works fine, but my project root contains lots of other files which change often and have no impact on my lambda (eg. files in other projects, log files under the localstack dir, etc.).

Being able to configure a pattern to narrow which files the hot reload file watcher cares about would prevent unnecessary reloads and speed up my development loop.

The globs would be relative to the `/var/task` folder, so I could instead have specified `packages/*/dist/**/*.js` if I wanted to be more specific, or `**/dist/**` if I wanted to be even less specific (in case I had other non-JS files in my build dir that I want to trigger hot reloads).

dfangl:
I feel this is a valid and useful suggestion, so I put it on accepted! Can't promise you an ETA, though!

davecardwell:
@dfangl Thank you!

The new Lambda provider and hot-reload mechanism feels like a big improvement. Even with the unnecessary hot reloads it still feels responsive enough - I’m seeing ~5s for the initial request which feels a little slow, but then 0.25–0.35s for subsequent requests which is fast enough for our purposes.

joe4dev:
> I’m seeing ~5s for the initial request which feels a little slow

Thank you very much for your valuable feedback @davecardwell  🙏🙏🙏 

[...]

PR #10438, opened 2024-03-12
Update runtime init, add dynamic configuration for folders to run chmod on in Lambda

dfangl:
## Motivation
We currently always run a `chmod` operation on certain task related folders when starting up a Lambda environment.
This however can take a lot of time if lambda prebuild mode is on, since it will do a copy on write operation on all changed files (which can be a lot). This introduces unnecessary IO load and can lead to extremely long startup times.

We should avoid running `chmod` on everything but `/tmp` in prebuild mode, and configure which folders to `chmod` on the LocalStack side, to avoid relying on potentially unreliable heuristics.

## Changes
* Update lambda init to include localstack/lambda-runtime-init#34
* Configure lambda init to run chmod operations:
  - On `/tmp`: always
  - On `/var/task` and `/opt` if there are layers configured (no matter if pro or not)
* Configure prebuild mode to run chmod on the code directory before building if layers are present (chmod on layers will be in pro, due to the design of the hook)

### DIFF
--- a/localstack/services/lambda_/invocation/docker_runtime_executor.py
+++ b/localstack/services/lambda_/invocation/docker_runtime_executor.py
@@ -17,6 +17,7 @@ from localstack.services.lambda_.invocation.executor_endpoint import (
 )
 from localstack.services.lambda_.invocation.lambda_models import FunctionVersion
 from localstack.services.lambda_.invocation.runtime_executor import (
+    ChmodPath,
     LambdaPrebuildContext,
     LambdaRuntimeException,
     RuntimeExecutor,
@@ -40,7 +41,7 @@ from localstack.utils.container_utils.container_client import (
     VolumeMappings,
 )
 from localstack.utils.docker_utils import DOCKER_CLIENT as CONTAINER_CLIENT
-from localstack.utils.files import rm_rf
+from localstack.utils.files import chmod_r, rm_rf
 from localstack.utils.net import get_free_tcp_port
 from localstack.utils.strings import short_uid, truncate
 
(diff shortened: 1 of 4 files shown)


localstack/services/lambda_/invocation/docker_runtime_executor.py
```diff
@@ -349,6 +357,19 @@ def start(self, env_vars: dict[str, str]) -> None:
...
             container_config.ports.add(config.LAMBDA_INIT_DELVE_PORT, config.LAMBDA_INIT_DELVE_PORT)
 
+        if (
+            self.function_version.config.layers
+            and not config.LAMBDA_PREBUILD_IMAGES
+            and self.function_version.config.package_type == PackageType.Zip
+        ):
+            # avoid chmod on mounted code paths
+            hot_reloading_env = container_config.env_vars.get(HOT_RELOADING_ENV_VARIABLE, "")
+            if "/opt" not in hot_reloading_env:
```
joe4dev, line 367: sidenote: if we are going to support advanced globbing in the future (see https://github.com/localstack/localstack/issues/7466), we should keep in mind that this check might not catch it. This would lead to chmod being applied on the host (or failure for read-only mounts).

localstack/services/lambda_/invocation/docker_runtime_executor.py
```diff
@@ -349,6 +357,19 @@ def start(self, env_vars: dict[str, str]) -> None:
                 container_config.ports = PortMappings()
             container_config.ports.add(config.LAMBDA_INIT_DELVE_PORT, config.LAMBDA_INIT_DELVE_PORT)
 
+        if (
```
joe4dev, line 360: This LocalStack-side chmod control is much more reliable than the previous heuristic 👍

localstack/services/lambda_/invocation/runtime_executor.py
```diff
@@ -128,6 +128,11 @@ class LambdaPrebuildContext:
     function_version: FunctionVersion
 
 
+class ChmodPath(TypedDict):
```
joe4dev, line 131: nice typings 💯

joe4dev (approved):
LGTM 👍 

Tested on macOS as part of the init review https://github.com/localstack/lambda-runtime-init/pull/34#pullrequestreview-1931066111

localstack/services/lambda_/invocation/docker_runtime_executor.py
dfangl: Good point! Not sure how we will implement globbing in the end, but definitely something to consider!

PR #10441, opened 2024-03-12
Update lambda runtime deprecations and fix Lambda ARM builds

joe4dev:
## Motivation

Some deprecated Lambda runtimes are fully phased out at AWS, such that we cannot create snapshots anymore. We need to update our Lambda runtimes lifecycle accordingly.

## Changes

* Update deprecated Lambda runtimes (matching the snapshot behavior): Deprecate 6 runtimes and remove them from test execution
* Update Lambda runtime snapshots
* Add Makefile to conveniently clean and build all Lambda common tests
* Fix all ARM builds such that the Lambda multiruntime tests are now fully ARM compatible.
* Re-purpose the Golang tests in `test_lambda_runtimes.py` while re-using the @multiruntime infrastructure

## Testing

* Community integration tests against Pro (not really necessary; already ran here and no ARM version): https://github.com/localstack/localstack/actions/runs/8339573132

### Keep supported runtimes in sync with resource browser

❓  Do we need to start offering a dynamic endpoint that lists the supported (and deprecated) runtimes?

Currently, the supported runtimes seem hardcoded. This only works with the latest LocalStack version. Users with older LocalStack versions might face issues.

### AWS behavior does not match docs

❓ What do we use as a reference?

> [!NOTE]
> This was mainly an issue right during the API updates but now everything looks in sync again.

Snapshot testing reveals that deprecated Lambda runtimes still remain supported much longer than described in the documentation. Therefore, our deprecation warnings are not in sync with AWS (mainly enforced by the unit test `tests.unit.services.lambda_.test_api_utils.TestApiUtils.test_check_runtime` that enforces that our tested runtimes match the supported runtimes). We could remove them from our tests if we value deprecation warning parity higher than current behavior parity (i.e., still being able to create deprecated runtimes).


tests/aws/services/lambda_/functions/common/README.md
```diff
@@ -0,0 +1,29 @@
...
+A runtime can be an aggregated runtime defined in `runtimes.py` (e.g., "python") or
+a specific runtime (e.g., "python3.12") if customizations are required.
+
+Each runtime directory defines a `Makefile` that
+* MUST define a `build` target that:
+  * a) creates a `build` directory containing all Lambda sources ready for packaging
+  * b) creates a `handler.zip` file with a Lambda deployment package
+* SHOULD define an `ARCHITECTURE` parameter to overwrite the target architecture (i.e., `x86_64` or `arm64`)
+  if architecture-specific builds are required (e.g., Dotnet, Golang, Rust).
+  * By default, the Makefile should build a deployment package with the same architecture as the host.
```
joe4dev, line 16: This default behavior is debatable:
* It currently aims for compatibility. For example, the build on a Linux ARM machine should not fail.
* It differs from the default Lambda behavior using `x86_64` by default.

Hence, on multi-architecture platforms (e.g., Apple Silicon), x86_64 builds require a parameter `make ARCHITECTURE=x86_64`.

tests/aws/services/lambda_/functions/common/echo/dotnet6/Makefile
```diff
@@ -1,9 +1,20 @@
+UNAME := $(shell uname -m)
+ifeq ($(UNAME),x86_64)
+	ARCHITECTURE ?= x86_64
+else
+	ARCHITECTURE ?= arm64
+endif
+DOCKER_PLATFORM ?= linux/$(ARCHITECTURE)
+# The packaging function architecture is x86_64 by default and needs to be set explicitly for arm64
+# https://github.com/aws/aws-extensions-for-dotnet-cli/blob/cdd490450e0407139d49248d94a4a899367e84df/src/Amazon.Lambda.Tools/LambdaDefinedCommandOptions.cs#L111
+FUNCTION_ARCHITECTURE ?= $(ARCHITECTURE)
```
joe4dev, line 10: Dotnet is a good example why an explicit `ARCHITECTURE` flag is needed:
* In general, the `--platform=$(DOCKER_PLATFORM)` flag is needed for cross-architecture on multi-architecture platforms
* The Dotnet build specifically requires the `--function-architecture` flag for ARM build. It does not detect the right architecture automatically (unlike Golang builds). Therefore, being consistently verbose clarifies the build options.

localstack/services/lambda_/runtimes.py
```diff
@@ -25,50 +31,58 @@
...
+    Runtime.python3_7,  # deprecated Dec 4, 2023 => Jan 9, 2024 => Feb 8, 2024
+    Runtime.dotnetcore3_1,  # deprecated Apr 3, 2023 => Apr 3, 2023 => May 3, 2023
+    Runtime.nodejs12_x,  # deprecated Mar 31, 2023 => Mar 31, 2023 => Apr 30, 2023
 ]
 # An unordered list of all AWS-supported runtimes.
 SUPPORTED_RUNTIMES: list[Runtime] = list(set(IMAGE_MAPPING.keys()) - set(DEPRECATED_RUNTIMES))
 
+# A temporary list of missing runtimes not yet supported in LocalStack. Used for modular updates.
+# TODO: add Dotnet8 runtime
+MISSING_RUNTIMES = [Runtime.dotnet8]
```
joe4dev, line 84: This should be removed with the follow-up addition in the Dotnet8 PR

localstack/testing/aws/lambda_utils.py
```diff
@@ -109,7 +108,19 @@ def package_for_lang(scenario: str, runtime: str, root_folder: Path) -> str:
         return package_path
 
     # packaging
-    result = subprocess.run(["make", "build"], cwd=runtime_dir)
```
joe4dev, line 112: [L90 above](https://github.com/localstack/localstack/pull/10441/files#diff-51f9f58239b7e55074f6f55c5c019aeec4f16cffede7c7afb8bf7b91f636dcf5R90): We could consider renaming `common` to multiruntime to make the reuse a bit more explicit.

tests/aws/services/lambda_/functions/common/echo/dotnet6/Makefile
```diff
@@ -1,9 +1,20 @@
+UNAME := $(shell uname -m)
+ifeq ($(UNAME),x86_64)
```
joe4dev, line 2: I wished this could be shorter but Ubuntu (in CI) yields `aarch64` and macOS yields `arm64`, so we need the conditional to standardize the `UNAME` architecture  :(

joe4dev:
just noticed that some of the ARM test skips were still there 😬 
Fixed and gonna see what fails in the CI now. Will see whether the Java tests need the `ARCHITECTURE` conditional for the Docker build as well ...

localstack/testing/aws/lambda_utils.py
```diff
@@ -35,21 +37,18 @@
     **dict.fromkeys(RUNTIMES_AGGREGATED.get("nodejs"), "index.handler"),
     **dict.fromkeys(RUNTIMES_AGGREGATED.get("ruby"), "function.handler"),
     **dict.fromkeys(RUNTIMES_AGGREGATED.get("java"), "echo.Handler"),
-    **dict.fromkeys(RUNTIMES_AGGREGATED.get("custom"), "function.handler"),
-    **dict.fromkeys(RUNTIMES_AGGREGATED.get("go"), "main"),
-    "dotnetcore3.1": "dotnetcore31::dotnetcore31.Function::FunctionHandler",  # TODO lets see if we can accumulate those
-    "dotnet6": "dotnet6::dotnet6.Function::FunctionHandler",
+    # TODO: check if "function.handler" is relevant; or just main or bootstrap?!
```
dfangl, line 40: It generally is not, unless the custom runtime reads it in some way. However, some value has to be put in there :)

tests/aws/services/lambda_/functions/common/Makefile
```diff
@@ -0,0 +1,10 @@
+# Top-level Makefile to invoke all make targets in sub-directories
+
+# Based on https://stackoverflow.com/a/72209214/6875981
+SUBDIRS := $(patsubst %/,%,$(wildcard */))
```
dfangl, line 4: Nice!

dfangl (approved):
Great changes, with nice comments to explain the reasoning :tada:

localstack/testing/aws/lambda_utils.py
joe4dev: Good clarification 👍  replaced TODO with a comment

joe4dev:
The Makefiles were not part of the caching checksum 😬, so the specific Rust build fix was never applied 🙈
Finally, everything is green 🟢 now that Makefile updates invalidate the cache as well 🚀

joe4dev:
Sidenote about the Rust ARM builds:

* List of Rust build targets in the docs: https://doc.rust-lang.org/nightly/rustc/platform-support.html
* This issue mentioned that "Aarch64 stack probes are tested in CI" and everything should work: https://github.com/rust-lang/rust/issues/77071
* The fix was done in this PR and released with Rust `1.76.0`: https://github.com/rust-lang/rust/pull/118491
* The `-musl` suffix was required to fix a GLIBC not found error with the Lambda runtime `provided.al2`

PR #10444, opened 2024-03-13
Set package data key to `localstack`

silv-io:
## Motivation

The key for the non-python package data is set to `localstack_ext` which is wrong and leaves out those files in the distribution.

## Changes

Change the key to the correct `localstack` so that those files are included


alexrashed (approved):
Awesome! Thanks for jumping on this! 🚀 🧹

PR #10445, opened 2024-03-13
Fix skip_teardown and force_synth usage in CDK tests

dominikschubert:
## Motivation

While working on another sample I've noticed some cases where we forgot to disable the `skip_teardown` and `force_synth` flags.

## Changes

- Removes any `force_synth=True` and `skip_teardown=True` cases.


steffyP (approved):
Thanks for fixing!

PR #10452, opened 2024-03-13
FIx: Referencing AWS::StackID is not consistent between AWS and Localstack

Morijarti:
## Motivation
Following template won't work on LS, but will pass on AWS
```
Resources:
  MyResource:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: unimportant
Outputs:
  SelectedOutput:
    Value: !Select [2, !Split ["/", !Ref "AWS::StackId"]]
  StackID:
    Value: !Ref "AWS::StackId"
```

This because on AWS `AWS::StackId` is ARN and in Localstack it is something like `stack-619106c6`

## Changes
- Still in exploration phase

## Testing

[test_select_split_stack_id](https://github.com/localstack/localstack/blob/b858e72bf0252e09a041dd009633680f509aa55b/tests/aws/services/cloudformation/engine/test_select_split.py#L7)

### DIFF
--- a/tests/aws/services/cloudformation/engine/test_select_split.py
+++ b/tests/aws/services/cloudformation/engine/test_select_split.py
@@ -1,0 +1,13 @@
+import os
+
+from localstack.testing.pytest import markers
+
+
+@markers.aws.validated
+def test_select_split_stack_id(aws_client, deploy_cfn_template, snapshot):
+    stack = deploy_cfn_template(
+                template_path=os.path.join(
+                    os.path.dirname(__file__), "../../../templates/engine/cfn_select_split.yml"
+                )
+            )
+    snapshot.match("outputs", stack.outputs)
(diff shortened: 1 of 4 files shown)


Morijarti:
We need to change

https://github.com/localstack/localstack/blob/b858e72bf0252e09a041dd009633680f509aa55b/localstack/services/cloudformation/engine/template_utils.py#L130C6-L130C12

```
case "AWS::StackId":
            # TODO return proper stack id!
            return stack_name
```

I'll take more in-depth look tommorow to see if we can modify signature to include stack_id, or maybe use 
`cloudformation.describe_stacks` call, but I'm unsure how good of idea is to create a client inside this function.

Morijarti:
Pull request closed as these changes will be addressed during CFn refactoring

PR #10453, opened 2024-03-13
Add aws-sam-cli-local to Docker image

eddumelendez:
## Motivation
Use samlocal within the Docker image.

## Changes
Install sam local in the Docker Image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,7 +133,7 @@ ENV PYTHONUNBUFFERED=1
 
 # Install the latest version of awslocal globally
 RUN --mount=type=cache,target=/root/.cache \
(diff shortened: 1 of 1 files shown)


alexrashed:
Hi @eddumelendez!
Thanks for your contribution! We are currently trying to avoid adding additional tools to the Docker image if it's possible, especially to the global Python environment.
Could you explain a bit your use-case? Would it suffice to use an [initialization hook](https://docs.localstack.cloud/references/init-hooks/) to install additional tools?

alexrashed:
I will close this PR for now, because it would add quite a few MB to the image. Please feel free to re-open the PR if you think this is something we really should add to the image.
Thanks!

PR #10454, opened 2024-03-13
Fix filter for provider detection in legacy StepFunctions tests

dominikschubert:
## Motivation

Noticed during my work on test selection, that the stepfunctions tests are executed particularly quickly. Looking at it in more detail I noticed that all tests are being skipped due to an issue in the conditional for the `skipif` marker.

```
======================= 14 skipped, 1 warning in 11.10s ========================
```

## Changes

- Fixes the skipif condition to actually evaluate instead of using the function itself as a truthy value... Python...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,6 @@ jobs:
           at: /tmp/workspace
       - prepare-testselection
       - prepare-pytest-tinybird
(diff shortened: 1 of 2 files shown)


alexrashed:
@MEPalma It looks like the filter now works correctly, but the tests are failing. Could you take a look at this?

PR #10457, opened 2024-03-14
added support for resource based policies in secretsmanager

macnev2013:
## Motivation
This Pull Request introduces support for resource-based policies in `secretsmanager`.

**Supported APIS**
- `DescribeSecret`
- `GetResourcePolicy`
- `ListSecretVersionIds`
- `PutResourcePolicy`
- `TagResource`
- `UntagResource`

## Changes
We now return data for resources that are in other account based on the `ARN` provided.
- NOTE: IAM enforcement is supported in LocalStack PRO - which means we are not testing against unauthorized resource access in Community version.

## Testing
Added aws validated test cases to access secretsmanager from cross-account.

### DIFF
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -232,7 +232,8 @@ class SecretsmanagerProvider(SecretsmanagerApi):
         secret_id = request["SecretId"]
         self._raise_if_invalid_secret_id(secret_id)
         backend = SecretsmanagerProvider.get_moto_backend_for_resource(secret_id, context)
-        backend.put_resource_policy(request["SecretId"], request["ResourcePolicy"])
+        arn, name = backend.put_resource_policy(secret_id, request["ResourcePolicy"])
+        return PutResourcePolicyResponse(ARN=arn, Name=name)
 
     @handler("PutSecretValue", expand=False)
     def put_secret_value(
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -6,13 +6,14 @@ import re
 import time
 from typing import Final, Optional, Union
 
+from botocore.utils import InvalidArnException
 from moto.iam.policy_validation import IAMPolicyDocumentValidator
 from moto.secretsmanager import secretsmanager_backends
 from moto.secretsmanager import utils as secretsmanager_utils
 from moto.secretsmanager.exceptions import SecretNotFoundException as MotoSecretNotFoundException
 from moto.secretsmanager.models import FakeSecret, SecretsManagerBackend
 from moto.secretsmanager.responses import SecretsManagerResponse
-from botocore.utils import InvalidArnException
+
 from localstack.aws.api import CommonServiceException, RequestContext, handler
 from localstack.aws.api.secretsmanager import (
     CancelRotateSecretRequest,
--- a/localstack/services/secretsmanager/provider.py
+++ b/localstack/services/secretsmanager/provider.py
@@ -7,6 +7,7 @@ import time
 from typing import Final, Optional, Union
 
 from moto.iam.policy_validation import IAMPolicyDocumentValidator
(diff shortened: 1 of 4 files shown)


tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -1947,3 +1947,91 @@ def test_create_secret_version_from_empty_secret(self, aws_client, snapshot, cle
...
+
+        # get tags from the primary account
+        response = aws_client.secretsmanager.describe_secret(SecretId=secret_arn)
+        assert response["Tags"] == [{"Key": "tag1", "Value": "value1"}]
+
+        # set tags from the secondary account
+        secondary_aws_client.secretsmanager.untag_resource(SecretId=secret_arn, TagKeys=["tag1"])
+
+        # get tags from the primary account
+        # Note: when removing tags, the response will be empty list in case of AWS,
```
viren-nadkarni, line 2021: Is this filed as a bug?

macnev2013: I couldn't find any bug describing such issue.

More information for reference:
- On AWS
  - if we describe the secret without adding any tags - It does not contain any tags
  - if we add tags > removes all the tags > describe secret - it contains `[]` (empty tag list)

macnev2013: I will create a short reproducible example and create a github issue on for it.

localstack/services/secretsmanager/provider.py
```diff
@@ -104,6 +105,21 @@ def __init__(self):
...
 
+    @staticmethod
+    def _check_is_arn(name_or_arn: str):
+        return ":" in name_or_arn
+
+    @staticmethod
+    def get_moto_backend_for_resource(
+        name_or_arn: str, context: RequestContext
+    ) -> SecretsManagerBackend:
+        if SecretsmanagerProvider._check_is_arn(name_or_arn):
```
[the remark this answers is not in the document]
macnev2013: Thanks for the suggestion. I've added this in the recent commit.

tests/aws/services/secretsmanager/test_secretsmanager.py
macnev2013: Reference to the Bug: https://github.com/localstack/localstack/issues/10477

localstack/services/secretsmanager/provider.py
```diff
@@ -173,15 +186,24 @@ def delete_secret(
...
     def get_resource_policy(
         self, context: RequestContext, request: GetResourcePolicyRequest
     ) -> GetResourcePolicyResponse:
-        self._raise_if_invalid_secret_id(request["SecretId"])
-        return call_moto(context, request)
+        secret_id = request["SecretId"]
+        self._raise_if_invalid_secret_id(secret_id)
+        backend = SecretsmanagerProvider.get_moto_backend_for_resource(secret_id, context)
+        policy = backend.get_resource_policy(secret_id)
+        return GetResourcePolicyResponse(**json.loads(policy))
```
viren-nadkarni, line 206: Since we are modifying the response generation, have you confirmed that these operations are snapshot tested? I only looked at this op and it appears that the only test using this operation `test_resource_policy` does not use snapshots. Earlier this would be tested in Moto but now we've moved that logic to the provider. I would strongly recommend adding this to make sure we're not introducing any deviations.

viren-nadkarni (changes requested):
Thanks for the fixes Nevil. I have a concern around ensuring the responses. Sorry for another cycle, I should've caught this earlier.

localstack/services/secretsmanager/provider.py
macnev2013: I have checked all the operations are snapshot validated other than resource policy. I've added snapshot validation for the same in the latest changes.

viren-nadkarni (approved):

PR #10458, opened 2024-03-14
add opt. service deps, opt-in services to test selection

alexrashed:
## Motivation
Now that we have the selective test execution for opt-in services introduced in https://github.com/localstack/localstack/pull/10301, we can onboard additional services to the test selection.
In this PR I added the services I own to the list (😻), but it might make sense to collect the first iteration of opt ins here and merge them together.
So please feel free to push to this PR if you want the tests of your service being selectively executed in PRs here.
In order to to so, just:
1. Make sure the mandatory and optional service dependencies in `localstack/utils/bootstrap.py` are correct.
  - They should list your service (if it's not completely self-contained).
  - `API_DEPENDENCIES`
    - This mapping states that a service requires one or more other services to provide it's basic functionality.
    - f.e. the following example means that `firehose` depends on `kinesis` to work: 
      ```python
      "firehose": ["kinesis"],
      ```
 - `API_DEPENDENCIES_OPTIONAL`
   - This mapping states that a service needs one or more other services for certain non-basic parts, f.e. with optional integrations.
   - On the example of `firehose` that would be at least `opensearch` and `elasticsearch`, because these are _potential_ targets for `firehose` streams.
2. Add your service provider code and the test code to the `OPT_IN`. The list is alphabetically sorted, I expect it to grow quite fast (and then become obsolete).

## Changes
- Introduces the notion of optional service dependencies, in contrast to "normal" / mandatory service dependencies.
- Adds the following services to the Opt-In of the selective test execution:
  - ACM
  - ElasticSearch
  - OpenSearch
  - SNS

## TODO
- [x] Discuss if we should split the optional from the mandatory service dependency definition.
  - `API_DEPENDENCIES` is used for strict service loading, where it is used as a mandatory service definition.
    - f.e. `lambda` cannot work without `s3` and `sts`
  - For the test selection, we are using this mapping for all dependencies (including optional dependencies).
    - f.e. `firehose` can work without `opensearch` for as long as you don't use it as a target. However, it cannot work without `kinesis` at all.
  - /cc @bentsku @steffyP @thrau

### DIFF
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -53,7 +53,7 @@ API_DEPENDENCIES = {
     # es forwards all requests to opensearch (basically an API deprecation path in AWS)
     "es": ["opensearch"],
     "cloudformation": ["s3", "sts"],
-    "lambda": ["s3", "sqs", "sts"],
+    "lambda": ["s3", "sts"],
     # firehose currently only supports kinesis as source, this could become optional when more sources are supported
     "firehose": ["kinesis"],
     "transcribe": ["s3"],
@@ -68,7 +68,7 @@ API_DEPENDENCIES = {
 API_DEPENDENCIES_OPTIONAL = {
     # firehose's optional dependencies are supported delivery stream destinations
     "firehose": ["es", "opensearch", "s3", "redshift"],
-    "lambda": ["cloudwatch", "dynamodbstreams", "logs", "kafka", "kinesis", "msk"],
+    "lambda": ["cloudwatch", "dynamodbstreams", "logs", "kafka", "kinesis", "msk", "sqs"],
     "ses": ["sns"],
     "sns": ["sqs", "lambda", "firehose", "ses", "logs"],
     "sqs": ["cloudwatch"],
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -62,7 +62,14 @@ API_DEPENDENCIES = {
 #   but which is needed for certain features (f.e. for one of multiple integrations)
 # - this mapping is used f.e. used for the selective test execution (localstack.testing.testselection)
 # - only add optional dependencies of services here, use API_DEPENDENCIES for mandatory dependencies
-API_DEPENDENCIES_OPTIONAL = {"firehose": ["opensearch", "es"]}
+API_DEPENDENCIES_OPTIONAL = {
+    "firehose": ["opensearch", "es", "s3"],
+    "lambda": ["logs", "cloudwatch"],
+    "ses": ["sns"],
+    "sns": ["sqs", "lambda", "firehose", "ses", "logs"],
+    "sqs": ["cloudwatch"],
+    "logs": ["lambda", "kinesis", "firehose"],
+}
 
 # composites define an abstract name like "serverless" that maps to a set of services
 API_COMPOSITES = {
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -41,16 +41,29 @@ from localstack.utils.sync import poll_condition
 
 LOG = logging.getLogger(__name__)
 
-# maps from API names to list of other API names that they depend on
+# Mandatory dependencies of services on other services
+# - maps from API names to list of other API names that they _explicitly_ depend on: <service>:<dependent-services>
+# - an explicit service dependency is a service without which another service's basic functionality breaks
+# - this mapping is used when enabling strict service loading (use SERVICES env var to allow-list services)
+# - do not add "optional" dependencies of services here, use API_DEPENDENCIES_OPTIONAL instead
 API_DEPENDENCIES = {
     "dynamodb": ["dynamodbstreams"],
     "dynamodbstreams": ["kinesis"],
     "es": ["opensearch"],
     "cloudformation": ["s3", "sts"],
     "lambda": ["s3", "sqs", "sts"],
-    "firehose": ["kinesis", "opensearch", "es"],
+    "firehose": ["kinesis"],
     "transcribe": ["s3"],
 }
+
+# Optional dependencies of services on other services
+# - maps from API names to list of other API names that they _optionally_ depend on: <service>:<dependent-services>
+# - an optional service dependency is a service without which a service's basic functionality breaks,
+#   but which is needed for certain features (f.e. for one of multiple integrations)
+# - this mapping is used f.e. used for the selective test execution (localstack.testing.testselection)
+# - only add optional dependencies of services here, use API_DEPENDENCIES for mandatory dependencies
+API_DEPENDENCIES_OPTIONAL = {"firehose": ["opensearch", "es"]}
+
 # composites define an abstract name like "serverless" that maps to a set of services
 API_COMPOSITES = {
     "serverless": [
--- a/localstack/utils/bootstrap.py
+++ b/localstack/utils/bootstrap.py
@@ -48,22 +48,26 @@ LOG = logging.getLogger(__name__)
 # - do not add "optional" dependencies of services here, use API_DEPENDENCIES_OPTIONAL instead
 API_DEPENDENCIES = {
     "dynamodb": ["dynamodbstreams"],
+    # dynamodbsteams uses kinesis under the hood
     "dynamodbstreams": ["kinesis"],
+    # es forwards all requests to opensearch (basically an API deprecation path in AWS)
     "es": ["opensearch"],
     "cloudformation": ["s3", "sts"],
     "lambda": ["s3", "sqs", "sts"],
+    # firehose currently only supports kinesis as source, this could become optional when more sources are supported
     "firehose": ["kinesis"],
     "transcribe": ["s3"],
 }
 
 # Optional dependencies of services on other services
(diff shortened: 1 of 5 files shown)


localstack/utils/bootstrap.py
```diff
@@ -48,7 +48,7 @@
     "es": ["opensearch"],
     "cloudformation": ["s3", "sts"],
     "lambda": ["s3", "sqs", "sts"],
-    "firehose": ["kinesis"],
+    "firehose": ["kinesis", "opensearch", "es"],
```
bentsku, line 51: I can try to articulate better my point around mixing those dependencies up, with examples. 

Right now, this variable means: this service is _built_ on top on those. 
And this PR will change the meaning of this: it will then mean this service is either _built_ on top or _interact_ with this service, what we could also refer as "required" vs "optional", but I maybe see it in a different way that those words.

For the test selection, this would not work for 2 of my services: S3 and SNS. 

We would like to keep the strict loading for both of them, as they are compatible with it.

S3 has S3 notifications integration and a KMS integration: this would interact with the following services: SQS, Lambda (who needs `sts`), SNS, EventBridge, KMS. SNS would then need SES and Firehose, which then needs Opensearch and Kinesis. 
And the S3 image uses the strict service loading, but now it has 10 (or 11 with `es`) services more. 

Same for SNS, it is compatible with strict service loading, but when setting it up for the tests, it would now need: SQS, SES, Lambda, Firehose etc... same as above.

So merging those 2 variables in one has pretty huge repercussion in my opinion. 
Either if we need to add a level to the dependencies, "required" or "optional", or have 2 different variables, this I don't know. 

\cc @thrau @steffyP

steffyP: I agree with @bentsku. We are mixing up two concepts when we just "re-use" this `API_DEPENDENCIES`. 

For the testing approach there should be a separate list IMO.

alexrashed: Definitely, I'm happy to implement this in here directly.
@thrau suggested `API_DEPENDENCIES_OPTIONAL` just next to `API_DEPENDENCIES`.
We could just use a merged version of the two in the test selection for now.
It should also be a bit restructured in upcoming iterations (@dominikschubert already mentioned that in a TODO), but I think that could be a good next step.
Would that suffice for now, @bentsku @steffyP?

bentsku: Once the change is there, I'll add SNS in the opt-in! S3 is everywhere, so I don't mind doing a full run 😄

alexrashed: I added `API_DEPENDENCIES_OPTIONAL` in the last commit. Let me know what you think 😄

bentsku: Nice, that's neat 👍 I like the `defaultdict` with `set` 😉 awesome, thank you so much! I'll push SNS soon then 😄

bentsku (approved):
LGTM! Nice, that looks really neat now, thank you so much for updating it. Can I still push SNS to the list of opt-in, even though it's ready for review? 
Also, thanks for updating the PR description, it's really clear 👍

alexrashed:
> LGTM! Nice, that looks really neat now, thank you so much for updating it. Can I still push SNS to the list of opt-in, even though it's ready for review? Also, thanks for updating the PR description, it's really clear 👍

Of course, please feel free to add your opt-ins, I'll merge this PR early next week with all the added opt-ins. :)
This obviously also counts for all other service owner's opt ins, but we can also just add more in a follow-up PR or individually! 🚀

bentsku:
This is not exactly related to this PR, but to the previous which added the test selection. I think we have a recursion problem when expending the list, I've tried to add SNS and related services, and we can't resolve the list anymore. 

If we imagine we have the following (real) optional dependencies:

```python
API_DEPENDENCIES_OPTIONAL = {
    "firehose": ["opensearch", "es", "s3"],
    "lambda": ["logs", "cloudwatch"],
    "ses": ["sns"],
    "sns": ["sqs", "lambda", "firehose", "ses", "logs"],
    "sqs": ["cloudwatch"],
    "logs": ["lambda", "kinesis", "firehose"]
}
```

We will forever be trying to resolve the `logs` service. Lambda will publish to `logs`, so if a change to `logs` is done, we would want to execute `lambda` tests (and `sns` with this list). But `logs` can also publish to a `lambda` destination, and now we have a recursion loop. 

But even then: I believe we might not need to actually do a recursive call, because we would trigger more tests that would not be needed. 

Let's take the same list as above. Let's say I'm doing a change in `ses`. `sns` has some dependency/tests with `ses`, so we want to execute them. But I don't think that means we want to execute all dependencies of `sns`, as we did not touch this service. `sns` will take care itself of testing the changes. If we're testing multiple integrations like like let's say `sns` into `lambda` into `logs` into `kinesis` into whatever next, I think a scenario test could do it, but I don't think we need to run the whole test suite. 

Maybe I am missing something, but do you think this is necessary? Should we just resolve the `API_DEPENDENCIES_OPTIONAL` once? I believe the list will be clear once we add all services there, and would not need recursion.

\cc @dominikschubert @alexrashed

dominikschubert:
I agree with @bentsku. We should for now avoid testing the full transitive relation here and just consider the first level of dependencies. Doing this fully recursively with all the potential service integrations is going to kinda defeat the whole purpose here.

I'd also suggest we separate the opt-in gathering and the introduction of the optional dependencies into a separate PR, but I'm also not opposed to just bundle it here if that's ok with @alexrashed.

alexrashed:
Please just go ahead and push it directly in this PR, @bentsku :)

bentsku:
> Please just go ahead and push it directly in this PR, @bentsku :)

I've added SNS and some known dependencies linked to it with https://github.com/localstack/localstack/pull/10458/commits/555ce88f23fe3d49d3dae1f3f53d3a003dffd324

I've also updated the recursion and added a test for services depending on each other with https://github.com/localstack/localstack/pull/10458/commits/cb6be456be2ed9d880fd9c0ce4da4bf730db8bcb

localstack/utils/bootstrap.py
```diff
@@ -51,6 +55,22 @@
     "firehose": ["kinesis"],
     "transcribe": ["s3"],
 }
+
+# Optional dependencies of services on other services
+# - maps from API names to list of other API names that they _optionally_ depend on: <service>:<dependent-services>
+# - an optional service dependency is a service without which a service's basic functionality breaks,
```
dominikschubert, line 61:
```suggestion
# - an optional service dependency is a service without which a service's basic functionality doesn't break,
```

dominikschubert: but I thought the point is that the basic functionality of a service does not break if the service it optionally  depends on is disabled?

localstack/utils/bootstrap.py
```diff
@@ -51,6 +55,25 @@
...
+#   but which is needed for certain features (f.e. for one of multiple integrations)
+# - this mapping is used f.e. used for the selective test execution (localstack.testing.testselection)
+# - only add optional dependencies of services here, use API_DEPENDENCIES for mandatory dependencies
+API_DEPENDENCIES_OPTIONAL = {
+    "firehose": ["opensearch", "es", "s3"],
+    "lambda": ["cloudwatch", "dynamodbstreams", "logs", "kafka", "kinesis", "msk"],
```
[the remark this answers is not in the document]
bentsku: I agree, I also thought of it. It's kinda related to "upstream" vs "downstream". With event source mapping, it's a bit tricky... not many services are like this I believe? 

Most of the current dependencies on these list are "targets" or external dependency like "global" services for telemetry / observability / events (`events`, `logs`, `cloudwatch`) I believe. Except for event source mapping 😅

localstack/utils/bootstrap.py
alexrashed: You are totally right 😄

localstack/utils/bootstrap.py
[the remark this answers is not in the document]
dominikschubert: Generally I'd treat these dependencies just like we do for testing. If it's configured via the Lambda API, it's a dependency of Lambda, no matter if it's a "source" or a "sink"/"destination".

localstack/utils/bootstrap.py
```diff
@@ -41,16 +41,43 @@
...
+# - do not add "optional" dependencies of services here, use API_DEPENDENCIES_OPTIONAL instead
 API_DEPENDENCIES = {
     "dynamodb": ["dynamodbstreams"],
+    # dynamodbsteams uses kinesis under the hood
     "dynamodbstreams": ["kinesis"],
+    # es forwards all requests to opensearch (basically an API deprecation path in AWS)
     "es": ["opensearch"],
     "cloudformation": ["s3", "sts"],
-    "lambda": ["s3", "sqs", "sts"],
+    "lambda": ["s3", "sts"],
```
dfangl, line 56: I moved the "sqs" dependency to optional here with 7f7f143832afed217261a57d86f5ce72aa231c34, since #10062 it is not required anymore for the internal behavior of lambda per default.

However due to event source mapping / destination / dead letter queues, it is still an optional dependency.

alexrashed: No worries, thanks for changing the deps! 🚀

PR #10116, opened 2024-01-24
KMS: Fixup ImportKeyMaterial for non-symmetric keys

uubk:
## Motivation
KMS allows you to import key material, that is, create KMS keys with "private keys" you already have. In the context of Localstack this is fairly important now that the old "local-kms" backend was removed since the new backend cannot be preseeded at the moment and controlling key material can be very important in testing.

## Changes
This PR drops the restriction on encrypt/decrypt for the KeyUsage in ImportKeyMaterial (the restriction for GetParametersForImport was already dropped as part of the fix for #9111) and should therefore fix #10115. It then fixes the issues that crop up when trying to use this with existing code - notably the hardcoded assumption that all imported key material is a symmetric key. Finally, it adds a new test to track this usecase and fixes a few other small issues you bump into when you update the KMS test snapshots.

PR #10459, opened 2024-03-14
fix test selection for community contributions

alexrashed:
## Motivation
https://github.com/localstack/localstack/pull/10301 introduced a dynamic test selection based on the touched service code.
Unfortunately, this just broke the pipeline of https://github.com/localstack/localstack/pull/10116 (a community contribution) because it expects a `GITHUB_API_TOKEN` which is a secret not being set for PRs coming from forks.

I tested the call to retrieve the PR data locally:
```
curl -v https://api.github.com/repos/localstack/localstack/pulls/10116
```
This call is successful without an auth token.
However, this could be flaky if these runners and their IP are used for a lot of unauthenticated GitHub API calls.

## Changes
- Make it optional to use the auth token, just send an unauthenticated request in case the token is not available.

## Testing
- Test if the execution works without the GitHub token with the last commit: 1a5a10394eb8ca0e04e9296b3927ac3f1d21a1eb

## TODO
- [x] Revert the last commit used for testing.

### DIFF
--- a/localstack/testing/testselection/github.py
+++ b/localstack/testing/testselection/github.py
@@ -6,7 +6,9 @@ def get_pr_details(repo_name: str, pr_number: str, token: str) -> (str, str):
     Fetch the base commit SHA, and the head commit SHA of a given pull request number from a GitHub repository.
     """
     url = f"https://api.github.com/repos/{repo_name}/pulls/{pr_number}"
-    headers = {"Authorization": f"token {token}", "Accept": "application/vnd.github.v3+json"}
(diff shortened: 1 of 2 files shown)


sannya-singal (approved):
Thanks for jumping on this! LGTM 🚀

alexrashed:
FYI: The CI already passed the step where this code is touched (both for the test commit, and for the invocation with the reverted test commit). I will move forward and merge this PR to make the fix available to the community contribution.

PR #10462, opened 2024-03-14
Add experimental option to distribute services across multiple localstack instances

dfangl:
## Motivation
In high load situations, the single threaded model of python seems to get into our way when trying to achieve better performance for LocalStack.

With this PR, you can use DNS to route requests between multiple LocalStack instances, which all run as one coherent unit.

This mode is, for now, dubbed "distributed mode".

It effectively disables all endpoint overrides, to let DNS handle the routing between instances. Only certain services keep a localhost usage by default, for example to assure dynamodbstreams on the same instance as dynamodb, and avoiding more load on the kinesis instance.

As long as the (highly experimental) flag is not set, there should not be any behavioral changes.

## Changes
* Add `DISTRIBUTED_MODE` config flag to disable all endpoint url overrides in internal calls
* Let dynamodbstreams use local clients instead of the default ones, which might point to other instances (to be debated)
* Adapt endpoint of the `kinesis_connector` to be redirectable with DNS

### DIFF
--- a/localstack/aws/connect.py
+++ b/localstack/aws/connect.py
@@ -486,7 +486,7 @@ class InternalClientFactory(ClientFactory):
             config = self._config.merge(config)
 
         endpoint_url = endpoint_url or get_service_endpoint()
-        if service_name == "s3":
+        if service_name == "s3" and endpoint_url:
             if re.match(r"https?://localhost(:[0-9]+)?", endpoint_url):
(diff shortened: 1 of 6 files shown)


thrau (approved):
nice! can't wait to try this. glad the changes needed are this minimal :ok_hand: 

since it's hidden behind a flag, we can still revisit the decisions you made about endpoints for dynamodb and kinesis

PR #10463, opened 2024-03-15
use ThreadPoolExecutor to forward DDB Streams events

bentsku:
## Motivation
Follow up from #10415

We previously spawned a thread for each incoming request, which resulted in a somewhat unbounded amount of threads being spawn when running benchmarks.

We could see this kind of error in logs:
```python
During handling of the above exception, another exception occurred:

    [...]
    raise ConnectionClosedError(
botocore.exceptions.ConnectionClosedError: Connection was closed before we received a valid response from endpoint URL: "http://localhost:4566/".

2024-03-15T01:11:04.444  INFO --- [.reactor-120] localstack.request.aws     : AWS dynamodb.BatchWriteItem => 200
2024-03-15T01:11:16.509  WARN --- [uncthread379] urllib3.connectionpool     : Connection pool is full, discarding connection: localhost. Connection pool size: 150
2024-03-15T01:11:16.610  WARN --- [uncthread380] urllib3.connectionpool     : Connection pool is full, discarding connection: localhost. Connection pool size: 150
2024-03-15T01:11:16.710  WARN --- [uncthread381] urllib3.connectionpool     : Connection pool is full, discarding connection: localhost. Connection pool size: 150
2024-03-15T01:11:16.809  WARN --- [uncthread382] urllib3.connectionpool     : Connection pool is full, discarding connection: localhost. Connection pool size: 150
2024-03-15T01:11:16.916  WARN --- [uncthread383] urllib3.connectionpool     : Connection pool is full, discarding connection: localhost. Connection pool size: 150
```

## Changes
Make use of a `ThreadPoolExecutor`, with a small number of thread, regulating the amount of outbound requests to Kinesis. The "cold start" of a thread in the executor is rather slow, around 0.5ms for me. Once all threads have been spawned, it takes less than 0.03ms, improving the overall performance (only slightly for small amount of items, but by a good amount of big amount of items, due to the onslaught of concurrent requests with the threads).
This might actually slow down the delivery of Kinesis records, but as of now, those records wouldn't even be delivered. 

## Benchmarks

It is the same as Scenario 2 from #10415, but with 1000 requests each. 

| DDB Streams | `latest` with Twisted | Fix & Twisted
| --- | --- | --- |
| **Throughput** (higher is better) |  |  | 
| `PutItem` throughput item/s | 224.66 | 231.41
| `BatchWriteItem (25)` throughput item/s | 2589.46 | 3701.11
| **Total time** (lower is better) |  |  | 
| `PutItem` total seconds (1000 items) | 4.4511 | 4.3214
| `BatchWriteItem (25)` total seconds (25k items) | 9.6545 | 6.7547

| Kinesis | `latest` with Twisted | Fix & Twisted
| --- | --- | --- |
| **Throughput** (higher is better) |  |  | 
| `PutItem` throughput item/s | 227.77 | 232.61
| `BatchWriteItem (25)` throughput item/s | 3102.02 | 4238.63
| **Total time** (lower is better) |  |  | 
| `PutItem` total seconds (1000 items) | 4.3904 | 4.2990
| `BatchWriteItem (25)` total seconds (25k items) | 8.0593 | 5.8981

We can see an improvement of around 30 to 50% for `BatchWriteItem`, and very minimal for `PutItem`, around 5%.

### DIFF
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -190,7 +190,7 @@ def dynamodb_table_exists(table_name: str, client=None):
 
 
 class EventForwarder:
-    def __init__(self, num_thread: int = 20):
+    def __init__(self, num_thread: int = 10):
         self.executor = ThreadPoolExecutor(num_thread, thread_name_prefix="ddb_stream_fwd")
 
     def shutdown(self):
--- a/localstack/services/dynamodb/provider.py
+++ b/localstack/services/dynamodb/provider.py
@@ -9,6 +9,7 @@ import time
 import traceback
 from binascii import crc32
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from datetime import datetime
 from operator import itemgetter
@@ -149,7 +150,7 @@ from localstack.utils.common import short_uid, to_bytes
 from localstack.utils.json import BytesEncoder, canonical_json
 from localstack.utils.scheduler import Scheduler
 from localstack.utils.strings import long_uid, md5, to_str
-from localstack.utils.threads import FuncThread, start_thread, start_worker_thread
+from localstack.utils.threads import FuncThread, start_thread
 
 # set up logger
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


thrau (approved):
Nice! The queue+worker approach seems to work nicely. I think there are some lessons here die other systems like firehose and eventbridge (cc @maxhoheiser @joe4dev )

Would be interesting to know what number of threads maximize throughout

bentsku:
> Nice! The queue+worker approach seems to work nicely. I think there are some lessons here for other systems like firehose and eventbridge (cc @maxhoheiser @joe4dev ). Perhaps also time to think about a higher level abstraction or shared thread pool across all these dispatching services.

I agree, I think we often do not need the `Future`-based result from the `ThreadPoolExecutor`, so I think we could have both: a shared pool, plus a simpler implementation of it which would be fire and forget. I'm sure we could shave a bit more time. 

> Would be interesting to know what number of threads maximize throughput
I can answer this, and to maximize simply DDB throughput, the optimal amount is... `1` 😄 I guess it puts no pressure on the gateway, and there are no "cold start" spawning the thread. 
But on the Kinesis level throughput, I think this will depend on the hardware + other usage of LocalStack. Ideally I suppose, if the input throughput is constant, the workers should be able to dish as much, I just hope we don't introduce too much pressure. But it's still better than being unbounded and absolutely destroying the gateway (or even having 300 inflight requests).

joe4dev:
Nice catch @bentsku 👏👏 Love the supporting benchmark results 💯 

We have several places where unbound thread spawning is inefficient and can hurt us (e.g., CloudWatch metric submissions). That's a nice pattern to adopt. We use it in Lambda at several places:
* `event_manager.py` to scale but limit async invokes
* `assignment.py` to provision and de-provision execution environments

It can have limitations when many events start to queue up in the ThreadPoolExecutor with limited visibility and we assume it is processing or already done.

> Perhaps also time to think about a higher level abstraction or shared thread pool across all these dispatching services.

Fully agree 👍 . A shared pool abstraction (within some scope) could indeed be very useful in LocalStack to avoid resource bottlenecks. @dfangl started an interesting tech spike of an AutoScaler where we can have more control over the scaling (needed to implement certain AWS scaling policies) in this PR https://github.com/localstack/localstack/pull/10032

---

nit: `max_workers=10` might be a little more explicit

PR #10468, opened 2024-03-16
refactor DDB streaming, event creation and forwarding

bentsku:
## Motivation
When working with DDB Streams and Kinesis Destinations, we would execute a lot of logic for queries with a lot of updated/inserted/deleted records. 
I thought we could improve the throughput by streamlining the process of forwarding records, and making a bit easier to follow by introducing some typing and structure. 

## Changes

- refactoring around the logic for DDB streams, trying to set up some typing and a notion of "RecordsMap" akin to a context to be passed to be forwarded. 
- improving performance by
  - reducing creations of new objects
  - no calls to `copy.deepcopy` which pickles/unpickles
  - reduce the number of calls to get the table specification, and passing around the data needed instead
  - conditionally set the `OldImage` and `NewImage` to the record, to not always have to `pop` it from the record if we don't need it

Also fixed an issue where the detection if a table had streaming enabled was still `True` if the `StreamEnabled` was set to `False`, which then executed all the logic for streams when not needed.

We should think about separating DynamoDB only tests from the one concerned with streams (DDB Streams or Kinesis).

## Benchmarks

Early benchmarks are showing a ~20% increase of throughput for `BatchWrite25`, the gain will be for better for bigger insert (possible with `TransactWriteItem` for example).

| With DDB Streams | latest | with fix |
| --- | --- | --- |
| **Throughput** (higher is better) |  |  |
| `PutItem` throughput item/s | 228.36 | 233.97 |
| `TransactWriteItems` (10) throughput item/s | 1706.06 | 1744.41 |
| `BatchWrite25` (25) throughput item/s | 3766.90 | 4574.19 |
| **Total time** (lower is better) |  |  |
| `PutItem` total seconds (1000 items) | 4.3791 | 4.2741 |
| `TransactWriteItems` (10) total seconds (5k items) | 5.8615 | 5.7326 |
| `BatchWrite25` (25) total seconds (25k items) | 6.6367 | 5.4655 |

| With Kinesis Streams | latest | with fix |
| --- | --- | --- |
| **Throughput** (higher is better) |  |  |
| `PutItem` throughput item/s | 231.92 | 237.03 |
| `TransactWriteItems` (10) throughput item/s | 1904.16 | 1947.37 |
| `BatchWrite25`(25) throughput item/s | 4242.45 | 5116.76 |
| **Total time** (lower is better) |  |  |
| `PutItem` total seconds (1000 items) | 4.3119 | 2.9776 |
| `TransactWriteItems` (10) total seconds (5k items) | 5.2517 | 5.1351 |
| `BatchWrite25` (25) total seconds (25k items) | 5.8928 | 4.8859 |

### DIFF
--- a/localstack/services/dynamodb/models.py
+++ b/localstack/services/dynamodb/models.py
@@ -1,5 +1,5 @@
 import dataclasses
-from typing import Optional, TypedDict
+from typing import TypedDict
 
 from localstack.aws.api.dynamodb import (
     AttributeMap,
@@ -30,7 +30,7 @@ class TableStreamType:
     is_kinesis: equivalent to NEW_AND_OLD_IMAGES, can be set at the same time as StreamViewType
     """
 
-    stream_view_type: Optional[StreamViewType]
+    stream_view_type: StreamViewType | None
     is_kinesis: bool
 
     @property
@@ -52,15 +52,16 @@ class DynamoDbStreamRecord(TypedDict, total=False):
     ApproximateCreationDateTime: int
     SizeBytes: int
     Keys: Key
-    StreamViewType: Optional[StreamViewType]
-    OldImage: Optional[AttributeMap]
-    NewImage: Optional[AttributeMap]
-    SequenceNumber: Optional[int]
+    StreamViewType: StreamViewType | None
+    OldImage: AttributeMap | None
+    NewImage: AttributeMap | None
+    SequenceNumber: int | None
 
 
 class StreamRecord(TypedDict, total=False):
     """
-    Global record type, this can contain both a KinesisRecord and a DynamoDBStreams record?
+    Related to DynamoDB Streams and Kinesis Destinations
+    This class contains data necessary for both KinesisRecord and DynamoDBStreams record
     """
 
     eventName: str
@@ -75,10 +76,18 @@ StreamRecords = list[StreamRecord]
 
 
 class TableRecords(TypedDict):
+    """
+    Container class used to forward events from DynamoDB to DDB Streams and Kinesis destinations.
+    It contains the records to be forwarded and data about the streams to be forwarded to.
+    """
+
     table_stream_type: TableStreamType
     records: StreamRecords
 
 
+# the RecordsMap maps the TableName to TableRecords, allowing forwarding to the destinations
+# some DynamoDB calls can modify several tables at once, which is why we need to group those events per table, as each
+# table can have different destinations
 RecordsMap = dict[TableName, TableRecords]
 
 
--- a/localstack/services/dynamodb/models.py
+++ b/localstack/services/dynamodb/models.py
@@ -1,6 +1,12 @@
+import dataclasses
+from typing import Optional, TypedDict
+
 from localstack.aws.api.dynamodb import (
+    AttributeMap,
+    Key,
     RegionName,
     ReplicaDescription,
+    StreamViewType,
     TableName,
     TimeToLiveSpecification,
 )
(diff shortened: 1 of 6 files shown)


localstack/services/dynamodb/provider.py
```diff
@@ -179,15 +180,46 @@
...
+
+
+class TableRecords(TypedDict):
+    table_stream_type: TableStreamType
+    records: StreamRecords
+
+
+RecordsMap = dict[TableName, TableRecords]
+
+
```
viren-nadkarni, line 216: I think these should live in `.models`

localstack/services/dynamodb/provider.py
```diff
@@ -198,104 +230,88 @@ def shutdown(self):
...
+            LOG.exception("Error while publishing to DynamoDB streams")
 
-        # forward to ddb_streams
-        forward_records = self.prepare_records_to_forward_to_ddb_stream(records)
-        records_to_ddb = copy.deepcopy(forward_records)
-        self.forward_to_ddb_stream(account_id, region_name, records_to_ddb)
+        try:
+            self.forward_to_ddb_stream(account_id, region_name, records_map)
+        except Exception:
+            LOG.exception("Error while publishing to Kinesis streams")
```
viren-nadkarni, line 257: The error messages seem exchanged between this and the previous try-except blocks

localstack/services/dynamodb/provider.py
bentsku: Makes sense, there's the small dependency to the `utils` one, but I've moved it to `models` too. Should make more sense!

localstack/services/dynamodb/models.py
```diff
@@ -12,6 +18,70 @@
...
+    eventVersion: str
+    dynamodb: DynamoDbStreamRecord
+    awsRegion: str
+    eventSource: str
+
+
+StreamRecords = list[StreamRecord]
+
+
+class TableRecords(TypedDict):
```
giograno, line 77: nit: I'd add a short docstring-style commend to briefly explain what's a `TableRecords` and a `RecordMap`.

giograno (approved):
Kudos for the cleanup and the performance improvements 💯 
Left a few nits, but nothing really blocking. 

+1 for the idea of separating the test suites.

Question: since we are looking at performances and you already have some benchmarks, would it make sense to start thinking about introducing proper performance testing?

localstack/services/dynamodb/provider.py
```diff
@@ -1945,31 +1916,30 @@ def find_existing_item_for_keys_values(item_keys: dict) -> AttributeMap | None:
...
-        return records
+            records_map[table_name] = TableRecords(
+                records=table_records, table_stream_type=stream_type
+            )
+
+        return records_map
 
     def forward_stream_records(
-        self, account_id: str, region_name: str, records: List[Dict], table_name: str = None
+        self, account_id: str, region_name: str, records_map: RecordsMap, table_name: str = None
```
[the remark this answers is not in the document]
bentsku: good catch, thanks!

bentsku:
Thanks for the review! 

> Question: since we are looking at performances and you already have some benchmarks, would it make sense to start thinking about introducing proper performance testing?

I think that's a good point, I'm not too sure how we would do that. I could share some of my benchmarks to be put in tests which wouldn't be executed in CI, but that people could look at locally when doing big changes. I'll look into that!

I'm not sure how we would check the performance, I think the CI runners are pretty bad/slow, and not consistent in how they would handle load. But yes, introducing general performance testing to avoid big regressions would be pretty neat!

bentsku:
Side note: just added a small optimization with https://github.com/localstack/localstack/pull/10468/commits/b989b17f243f6f96c18cc0b0242e08cf401e9832

From the Python documentation: https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects

> Keys views are set-like since their entries are unique and [hashable](https://docs.python.org/3/glossary.html#term-hashable). Items views also have set-like operations since the (key, value) pairs are unique and the keys are hashable.
[...]
> For set-like views, all of the operations defined for the abstract base class [collections.abc.Set](https://docs.python.org/3/library/collections.abc.html#collections.abc.Set) are available (for example, ==, <, or ^). While using set operators, set-like views accept any iterable as the other operand, unlike sets which only accept sets as the input.

So I figured we could use set operations and that would be faster than `any`. 

Small benchmark with `timeit`:

<details><summary>Details</summary>
<p>

```python
items = [
    {"id": "bla1", "key": "foo", "key2": "bar", "key4": "test"},
    {"id": "bla2", "key": "foo2", "key2": "bar", "key4": "test"},
    {"id": "bla3", "key": "foo", "key2": "bar", "key4": "test"},
    {"id": "bla4", "key": "foo4", "key2": "bar", "key4": "test"},
    {"id": "bla5", "key": "foo", "key2": "bar", "key4": "test"},
    {"id": "bla6", "key": "foo6", "key2": "bar", "key4": "test"},
    {"id": "bla7", "key": "foo", "key2": "bar", "key4": "test"},
    {"id": "bla8", "key": "foo7", "key2": "bar", "key4": "test"},
    {"id": "bla", "key": "foo", "key2": "bar", "key4": "test"},
]
item_keys = {"id": "bla", "key": "foo"}


def test():
    for item in items:
        if all(item.get(k) == v for k, v in item_keys.items()):
            return item


def test2():
    for item in items:
        if item_keys.items() <= item.items():
            return item


def test3():
    keys = item_keys.items()
    for item in items:
        if keys <= item.items():
            return item


if __name__ == '__main__':
    import timeit

    result = timeit.timeit(test)
    print(result)

    result = timeit.timeit(test2)
    print(result)

    result = timeit.timeit(test3)
    print(result)
```

This gives the following:
```
2.4056403339782264
0.6815093750192318
0.5881134170049336
```

So quite a good improvement, even though it might be something like 0.01ms 😅 

</p>
</details>

viren-nadkarni (approved):

PR #10478, opened 2024-03-18
fix: standardise the TEST_AWS_REGION_NAME usage

sannya-singal:
## Motivation
Standardise the usage of `TEST_AWS_REGION_NAME` and `SECONDARY_TEST_AWS_REGION_NAME` in our codebase. 

Currently CircleCi uses `TEST_AWS_REGION_NAME` #10367 to set a random credential but the LocalStack codebase picks region from the `TEST_AWS_REGION` env variable. This caused the tests to not pick up the random region and instead fall back to the default.

## Changes
Eliminate `TEST_AWS_REGION`, `SECONDARY_TEST_AWS_REGION` environment variable and use `TEST_AWS_REGION_NAME`, `SECONDARY_TEST_AWS_REGION_NAME` respectively.


sannya-singal:
Currently 37 tests are failing https://app.circleci.com/pipelines/github/localstack/localstack/23423/workflows/28f8a13e-66aa-44e6-be14-de9d351a9710 for multi-accounts and multi-regions.

localstack/constants.py
```diff
@@ -148,7 +148,7 @@
 TEST_AWS_ACCOUNT_ID = os.getenv("TEST_AWS_ACCOUNT_ID") or DEFAULT_AWS_ACCOUNT_ID
 TEST_AWS_ACCESS_KEY_ID = os.getenv("TEST_AWS_ACCESS_KEY_ID") or "test"
 TEST_AWS_SECRET_ACCESS_KEY = os.getenv("TEST_AWS_SECRET_ACCESS_KEY") or "test"
-TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION") or "us-east-1"
+TEST_AWS_REGION_NAME = os.getenv("TEST_AWS_REGION_NAME") or "us-east-1"
```
alexrashed, line 151: These kind of changes can be very dangerous, because this is technically a breaking change. Is this env variable really only used internally and only used for testing? I guess so, but just to make sure... 😛

localstack/constants.py
```diff
@@ -148,7 +148,7 @@
...
 
 # Additional credentials used in the test suite (when running cross-account tests)
 SECONDARY_TEST_AWS_ACCOUNT_ID = os.getenv("SECONDARY_TEST_AWS_ACCOUNT_ID") or "000000000002"
```
alexrashed, line 154: I can't comment on lines which have not been touched, but line 157 is the following:
```
SECONDARY_TEST_AWS_REGION_NAME = os.getenv("SECONDARY_TEST_AWS_REGION") or "ap-southeast-1"
```
This one should then maybe also be changed to `SECONDARY_TEST_AWS_REGION_NAME`?

alexrashed (changes requested):
Thanks for jumping on this! If I understand this correctly, this is used to fix the CircleCI config.
If so, it would be great to link the PR where you introduced these changes to the workflow (as context for reviewers) and give a bit more context in the description of the PR.

localstack/constants.py
sannya-singal: Yes, also I had a discussion around this with @viren-nadkarni who helped me pointing this out.

alexrashed (approved):

viren-nadkarni:
> Currently 37 tests are failing for the this set of random creds: https://app.circleci.com/pipelines/github/localstack/localstack/23423/workflows/679f1237-6253-495d-947e-45f19e2fc8e9/jobs/191264/tests.

Interesting that these failures have now appeared. Do you have any insights in whether they are due to the choice of region (i.e. AWS services not being available in that region) or because of incompatible changes introduced in the last couple of weeks?

sannya-singal:
> > Currently 37 tests are failing for the this set of random creds: https://app.circleci.com/pipelines/github/localstack/localstack/23423/workflows/679f1237-6253-495d-947e-45f19e2fc8e9/jobs/191264/tests.
> 
> Interesting that these failures have now appeared. Do you have any insights in whether they are due to the choice of region (i.e. AWS services not being available in that region) or because of incompatible changes introduced in the last couple of weeks?

Yes, majorly because of AWS services not available in a particular region, for instance, in the linked workflow we will see
`'eu-south-1' is not a valid AWS region name for transcribe`.

PR #10480, opened 2024-03-18
Fixed issue where TEST_DISABLE_RETRIES_AND_TIMEOUTS would get always evaluated to True

Morijarti:
## Motivation
In current implementation `TEST_DISABLE_RETRIES_AND_TIMEOUTS`  would get evaluated to True regardless if it was set to 0 or 1. Only way to get it to be false is to remove it from env variables

## Changes
Replaced `os.environ.get` with `is_env_true`


alexrashed (approved):
Good catch! Thanks for the fix! LGTM! 🐞

Issue #10451, opened 2024-03-13
bug: api gateway v2 (pro) - entire query string not proxied on POST requests if :// is in any values in the query string


joshbalfour:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When proxying requests with API gateway, the query string doesn't make it through to the destination when in localstack, but on AWS it does

### Expected Behavior

The query string makes it to the destination

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)
localstack:
    container_name: "localstack-main"
    image: localstack/localstack-pro
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    environment:
      # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
      - DEBUG=${DEBUG:-0}
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN:?}  # required for Pro
      - DISABLE_CORS_CHECKS=1
      - DISABLE_CUSTOM_CORS_APIGATEWAY=1
      - EXTRA_CORS_ALLOWED_HEADERS=x-edgeapi-session-id
      - DISABLE_TRANSPARENT_ENDPOINT_INJECTION=1
    volumes:
      - "./volume:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"

### Environment

```markdown
- OS: macOS 14.1.1 (23B81)
- LocalStack: latest
```

bentsku:
Hello @joshbalfour and thanks for your report!

Could you maybe give a little more information about the setup? What kind of integration are you using and with what parameters? Maybe a minimal reproducible sample just showing the behavior would be great.
Thanks!

joshbalfour:
I've managed to reproduce it with this cloudformation:

```json
{
 "Resources": {
  "apiC8550315": {
   "Type": "AWS::ApiGatewayV2::Api",
   "Properties": {
    "Name": "repro",
    "ProtocolType": "HTTP"
   },
   "Metadata": {
    "aws:cdk:path": "repro/api/Resource"
   }
  },
  "apiDefaultStage04B80AC9": {
   "Type": "AWS::ApiGatewayV2::Stage",
   "Properties": {
    "ApiId": {
     "Ref": "apiC8550315"
    },
[...]

PR #10483, opened 2024-03-18
fix query string parser with non encoded http url in parameters

bentsku:
## Motivation
As reported in #10451, we would have an issue when trying to parse query string parameters containing a URL. 
I suspected a split somewhere with `http://`. It seems it was a somewhat related issue.

## Changes
Fix the `parse_query_string` util (actually only used in APIGW, would maybe be good to move it around). 
It was checking for the presence of `http://` in the string (to check if it was a full URL, a path or a query string), but it could be container as a query string parameter also. 
I've replaced it with a regex containing lower case characters between 2 and 5 (for `ws://` to `https://`) now at the beginning of the string. I've added more cases to the test and verified that it fixes the issue. (More tests in the companion PR).

### DIFF
--- a/localstack/utils/aws/aws_responses.py
+++ b/localstack/utils/aws/aws_responses.py
@@ -19,6 +19,8 @@ from localstack.utils.strings import short_uid, str_startswith_ignore_case, to_b
 
 REGEX_FLAGS = re.MULTILINE | re.DOTALL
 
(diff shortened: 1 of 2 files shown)


whummer (approved):
Great catch @bentsku , kudos for fixing this! 🚀

PR #10492, opened 2024-03-19
fix CircleCI Tinybird reporting on reruns

alexrashed:
## Motivation
#10346 introduced reporting of our CircleCI workflow metadata to Tinybird for our internal CI analytics.
Unfortunately, we are seeing that the reporting is not properly updated in case a failed workflow is "rerun workflow from failed":
image
This is because this really only executes jobs that failed as well as their dependents.
This PR fixes this by also failing the waiter (running alongside the workflow waiting for the pipeline to finish and to report the result) in case any of the steps failed. This results in the waiter being rerun as well with the "Rerun workflow from failed" feature.

## Changes
- Simplify the waiter jobs to a single job.
- Fail the waiter job in case the outcome of the main workflow is not `success`.

## Testing
- The last commit always fails and removes the filter to only run the waiter on `master`.
- This triggered a run which failed, which I restarted using "rerun workflow from failed": https://app.circleci.com/pipelines/github/localstack/localstack/23482/workflows/26c43878-ad39-4724-9951-8bd9530b1bf0
- This rerun also triggered the waiter to rerun.

## TODO
- [x] Revert the last testing commit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -581,70 +581,51 @@ jobs:
             source .venv/bin/activate
             bin/release-helper.sh set-ver $(bin/release-helper.sh next-dev-ver)
             make publish
-  save-start-time:
-    executor: ubuntu-machine-amd64
-    working_directory: /tmp/workspace/repo
-    steps:
-      - run:
-          name: Save start time to file
-          command: mkdir stats && echo $(date +%s) > stats/start-time.txt
-      - persist_to_workspace:
-            root: /tmp/workspace
-            paths:
-                - repo/stats/
   waiter:
     executor: ubuntu-machine-amd64
     working_directory: /tmp/workspace/repo
     steps:
-      - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -581,70 +581,51 @@ jobs:
...
+            curl -X POST "https://api.tinybird.co/v0/events?name=circle_ci_workflows" -H "Authorization: Bearer $TINYBIRD_CI_TOKEN" -d @stats.json
 
+            # Fail this step depending on the success to trigger a rerun of this step together with others in case of a "rerun failed"
+            [[ $OUTCOME = "success" ]] && exit 0 || exit 1
 
 workflows:
   main:
     jobs:
-      - save-start-time:
+      - waiter:
```
silv-io, line 625: Maybe we can rename this into something more descriptive now.

Maybe `push-to-tinybird` like in the GitHub actions so that it's more easily discoverable. 

Need to add this name to the filter in the waiter however though

silv-io (approved):
LGTM! Just a thought about naming.

Adding the job url also shouldn't be much of an issue I think.

PR #10495, opened 2024-03-19
fix S3 copy in place in versioned bucket

bentsku:
## Motivation
As reported in the Community Slack, a user encountered an error after copying in place an object in a versioned bucket. The first reported what about a previously fixed issue, which took a while to understand what was the final bug. 

It turns out that I've had a big gap in the logic of the Storage system of S3 when copying in place. To avoid unnecessary copying of data, if the copy operation was detected as "in place", we would return the same underlying file object. However, for versioned bucket, those are not really "in place", they create a new version. 

## Changes
- fix the logic to not return the same underlying stored object if any of the objects have a version id. Only minor issue is if both objects have their version id set to `null`, which would mean that the bucket has versioning disabled, we could still apply the optimization, but this is fine. Better avoid nasty bugs 😬

### DIFF
--- a/localstack/services/s3/v3/storage/core.py
+++ b/localstack/services/s3/v3/storage/core.py
@@ -233,3 +233,31 @@ class S3ObjectStore(abc.ABC):
         Resetting the `S3ObjectStore` will delete all the contained resources.
         """
         pass
+
+
+def should_copy_in_place(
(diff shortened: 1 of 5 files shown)


thrau (approved):
LGTM! kudos for finding this so quickly! :100:

PR #10497, opened 2024-03-20
Move version single source outside of `__init__.py`

silv-io:
## Motivation

To introduce namespace packages, the modern way is to remove all `__init__.py` files in the namespace package, our case `localstack_ext`. Currently this is not possible because our init file contains the version attribute used in various scripts. 

## Changes

- Move the `__version__` out of the `__init__.py` into a `VERSION` file residing in the top-level of the project.
- Make `pyproject.toml` take the version out of the `VERSION` file
- Adapt all the scripts and Makefile to do their changes based on and in the `VERSION` file
- To still allow for run-time access of the version, add a `setup.py` which - on install/build - takes the value from the `VERSION` file and produces a `version.py` file in `localstack`.
- For backwards compatibility with older images, switch all the usages of the version to the pre-existing `VERSION` in `constants.py` and take the value from `VERSION` out of `localstack.version.__version__` (`version.py` might not yet exist in older images which can lead to issues in the CI right now)

## Testing

- Pipeline should be green
- Distribution should look the same (maybe we can introduce some regression tests here in the future, so that we can more easily check if we botched the distribution by some change)

## TODO

What's left to do:

- [ ] After merging, merge the corresponding ext PR (https://github.com/localstack/localstack-ext/pull/2744) as soon as the changes are in the published package
- [ ] Try to understand why we have an AppVeyor job in CI and why it fails

> [!NOTE]
> This PR was recreated with a new branch name to trigger the companion branch detection


thrau (commented):
thanks for taking a stab at this @silv-io! could you clarify for posterity why we need a `VERSION` file in this iteration? wouldn't introducing the `version` module and `localstack.version.__version__` be enough?

silv-io:
> thanks for taking a stab at this @silv-io! could you clarify for posterity why we need a `VERSION` file in this iteration? wouldn't introducing the `version` module and `localstack.version.__version__` be enough?

@thrau, there are multiple benefits to that.

As you can see in the changes, many of the scripts just get way easier to handle when having the version source separated into a simple text file. This is also corroborated in [the setuptools guide](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/) in (4.):
> An advantage with this technique is that it’s not specific to Python. Any tool can read the version.

So no need to use `grep`, `awk` and `sed` anymore here. While extracting the version with these tools is fairly easy, any change we might make at some point might be error-prone because of the heavy usage of regex patterns. A simple `cat VERSION` won't lead to many errors when trying to change things :) 

Another point is that in a multi distribution setup it is also much easier to send a plain version file to subpackages that they can reference in the `pyproject.toml`, than to write a script which would put a python version file in exactly the right spot for all the different distributions.

bin/release-dev.sh
```diff
@@ -3,11 +3,10 @@
 set -e
 
 # use UTC timestamp as version
-ver=$(date -u +%Y%m%d%H%M%S)
+timestamp=$(date -u +%Y%m%d%H%M%S)
 
-sed -i -r "s/^__version__ = \"(.*\.dev)\"/__version__ = \"\1${ver}\"/" localstack/__init__.py
+echo "$timestamp" >> VERSION
```
thrau, line 8: if i run the script several times we would keep appending the timestamp, right? not that it's intended to be used that way, but perhaps making it more robust this way?

thrau (approved):
Overall looks really good! The arguments make sense to me and I can see that the version file is a good way forward!
Just one comment related to updating the version number. I think I may be missing something though since I saw that the content is overwritten somewhere in the release helper.

bin/release-dev.sh
silv-io: You're right. Will fix that shortly!

Issue #10437, opened 2024-03-12
bug: Headers aren't being transformed into Title Case the way that they are in AWS


Garethp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you send a request to API Gateway in LocalStack with the header "authorization" in lower-case, it's not being transformed into the title case "Authorization".

### Expected Behavior

Accord to the [AWS Documentation for API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html) headers are case-sensitive, not insensitive. Furthermore, if you try to ping an API Gateway behind a Cloudfront with a lower-case header for something like "Authorization" then it gets transformed before reaching the handler Lambda. That means that in AWS, if you send a request with the header "authorization" then it gets recieved as "Authorization"

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```ts
export class GatewayStack extends Stack {
  constructor(scope: Construct, id: string, props: StackProps) {
    super(scope, id, props);

    const api = new RestApi(this, "test-api", {
      restApiName: "test-api",
    });

    new Distribution(this, "ApiCloudfront", {
      defaultBehavior: {
        origin: new RestApiOrigin(api),
        allowedMethods: AllowedMethods.ALLOW_ALL,
        cachePolicy: CachePolicy.CACHING_DISABLED,
        originRequestPolicy: OriginRequestPolicy.ALL_VIEWER,
        viewerProtocolPolicy: ViewerProtocolPolicy.HTTPS_ONLY,
      },
    });

    const lambda = new Function(this, "lambda-handler", {
      runtime: Runtime.NODEJS_20_X,
      handler: "index.handler",
      code: Code.fromInline(
        "exports.handler = async (event) => ({statusCode: 200, body: JSON.stringify(event.headers, null, 4) })",
      ),
    });

    api.root.addMethod("GET", new LambdaIntegration(lambda));
  }
}
```

```sh
curl -H "authorization: Bearer test" "https://$(awslocal apigateway get-rest-apis | jq .items[0].id -r).execute-api.localhost.localstack.cloud:4566/prod/"
```

### Environment

```markdown
- OS: Ubuntu 20.04
- LocalStack: 3.2
```

### Anything else?

[...]

PR #10500, opened 2024-03-20
fix capitalization of Authorization header for AWS_PROXY integration

bentsku:
## Motivation
As reported in #10437, we were missing the capitalization of Authorization header in API Gateway AWS_PROXY integration. 
This is a weird fix, because it seems that AWS_PROXY integrations are behind CloudFront in some way, because it returns CloudFront headers even if not configured. The user does not have any say in the configuration of said "default" cloudfront distribution. But it seems to have the same logic around the Authorization header.
See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/add-origin-custom-headers.html#add-origin-custom-headers-forward-authorization

## Changes
- add a fix to capitalize the authorization header if present in any casing
- modify a currently existing test to verify the assumption

\cc @giograno @Morijarti

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -276,11 +276,20 @@ class LambdaProxyIntegration(BackendIntegration):
         single_value_query_string_params = {
             k: v[-1] if isinstance(v, list) else v for k, v in query_string_params.items()
         }
+        # Some headers get capitalized like in CloudFront, see
+        # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/add-origin-custom-headers.html#add-origin-custom-headers-forward-authorization
+        # It seems AWS_PROXY lambda integrations are behind cloudfront, as seen by the returned headers in AWS
+        to_capitalize: list[str] = ["authorization"]  # some headers get capitalized
+        headers = {
+            k.capitalize() if k.lower() in to_capitalize else k: v for k, v in headers.items()
+        }
+
         # AWS canonical header names, converting them to lower-case
         headers = canonicalize_headers(headers)
+
         return {
             "path": path,
-            "headers": dict(headers),
+            "headers": headers,
             "multiValueHeaders": multi_value_dict_for_list(headers),
             "body": data,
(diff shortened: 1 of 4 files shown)


Morijarti (approved):
LGTM.

Thanks for implementing changes so quickly.

localstack/services/apigateway/integration.py
```diff
@@ -276,11 +276,20 @@ def construct_invocation_event(
         single_value_query_string_params = {
             k: v[-1] if isinstance(v, list) else v for k, v in query_string_params.items()
         }
+        # Some headers get capitalized like in CloudFront, see
+        # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/add-origin-custom-headers.html#add-origin-custom-headers-forward-authorization
+        # It seems AWS_PROXY lambda integrations are behind cloudfront, as seen by the returned headers in AWS
```
whummer, line 281: nit: Out of curiosity - presumably that's only relevant for `AWS_PROXY` integrations, not for regular `AWS` integrations, right? Not something we need to cover here, was just curious if you had some insights into that.

whummer (approved):
Great insights @bentsku , interesting behavior by API GW, again! 🤷 :)

Also kudos for updating/enhancing the snapshot transformers along the way.. 👌

localstack/services/apigateway/integration.py
bentsku: You're right! I had the same question, it should not be relevant for other `AWS` integrations because we selectively pass headers to the integration via request parameters.
However, I wonder if it's relevant to `HTTP_PROXY` integrations. However, we do not have good validated tests in community yet. But I keep a note for the next refactor for this kind of thing 😄 good to keep in mind! 👌

bentsku: @cloutierMat had a very good insight about this: this is only the case for Edge optimized API endpoints: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-endpoint-types.html

We should probably fix this in a next iteration, as this is quite the edge case, but this really good to know.

PR #10508, opened 2024-03-21
fix: add transformer to region attribute in failing s3 for 'eu-west-1'

sannya-singal:
## Motivation
We ran a pipeline for randomised region and account: [here](https://app.circleci.com/pipelines/github/localstack/localstack/23423/workflows/679f1237-6253-495d-947e-45f19e2fc8e9/jobs/191264/tests) and found that `test_region_header_exists` fails with the following error: 

> >> match key: head_bucket
> 	#x1B[33m(~)#x1B[0m /BucketRegion 'eu-west-1' → region... (expected → actual)
> 
> 	Ignore list (please keep in mind list indices might not work and should be replaced):
> 	["$..BucketRegion"]

So, basically when we set the `TEST_AWS_REGION_NAME` to `eu-west-1` we get this error. 

## Changes
This PR adds a transformer for region and re-generates the snapshot for `tests.aws.services.s3.test_s3.TestS3.test_region_header_exists`.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -340,14 +340,16 @@ class TestS3:
     @markers.snapshot.skip_snapshot_verify(paths=["$..AccessPointAlias"])
     def test_region_header_exists(self, s3_create_bucket, snapshot, aws_client):
         snapshot.add_transformer(snapshot.transform.s3_api())
+        region = "eu-west-1"
+        snapshot.add_transformer(RegexTransformer(region, "<region>"))
         bucket_name = s3_create_bucket(
-            CreateBucketConfiguration={"LocationConstraint": "eu-west-1"},
+            CreateBucketConfiguration={"LocationConstraint": region},
         )
         response = aws_client.s3.head_bucket(Bucket=bucket_name)
-        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == "eu-west-1"
+        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region
         snapshot.match("head_bucket", response)
         response = aws_client.s3.list_objects_v2(Bucket=bucket_name)
-        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == "eu-west-1"
+        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region
         snapshot.match("list_objects_v2", response)
 
     @markers.aws.validated
--- a/tests/aws/services/s3/test_s3.snapshot.json
+++ b/tests/aws/services/s3/test_s3.snapshot.json
@@ -1,10 +1,10 @@
 {
   "tests/aws/services/s3/test_s3.py::TestS3::test_region_header_exists": {
-    "recorded-date": "06-12-2023, 16:47:40",
+    "recorded-date": "21-03-2024, 08:05:13",
     "recorded-content": {
       "head_bucket": {
         "AccessPointAlias": false,
-        "BucketRegion": "eu-west-1",
+        "BucketRegion": "<region>",
         "ResponseMetadata": {
           "HTTPHeaders": {},
           "HTTPStatusCode": 200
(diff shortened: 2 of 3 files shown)


tests/aws/services/s3/test_s3.py
```diff
@@ -340,14 +340,16 @@ def test_copy_object_kms(self, s3_bucket, kms_create_key, snapshot, aws_client):
...
+            CreateBucketConfiguration={"LocationConstraint": region},
         )
         response = aws_client.s3.head_bucket(Bucket=bucket_name)
-        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == "eu-west-1"
+        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region
         snapshot.match("head_bucket", response)
         response = aws_client.s3.list_objects_v2(Bucket=bucket_name)
-        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == "eu-west-1"
+        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region
         snapshot.match("list_objects_v2", response)
```
bentsku, line 353: I would replace the test with the following:
```python
    @markers.aws.validated
    def test_region_header_exists(self, s3_create_bucket_with_client, aws_client_factory):
        region_default = "us-east-1"
        region_non_default = "eu-west-1"
        client_default_region = aws_client_factory(region_name=region_default).s3
        client_non_default_region = aws_client_factory(region_name=region_non_default).s3

        bucket_name = f"test-bucket-{short_uid()}"
        s3_create_bucket_with_client(
            client_default_region,
            Bucket= bucket_name,
            CreateBucketConfiguration={"LocationConstraint": region_non_default},
        )
        response = client_default_region.head_bucket(Bucket= bucket_name)
        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region_non_default

        response = client_default_region.list_objects_v2(Bucket= bucket_name)
        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region_non_default

        response = client_non_default_region.head_bucket(Bucket= bucket_name)
        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region_non_default

        response = client_non_default_region.list_objects_v2(Bucket= bucket_name)
        assert response["ResponseMetadata"]["HTTPHeaders"]["x-amz-bucket-region"] == region_non_default
```

It would not be impacted anymore by the multi account test changes that way.

bentsku (changes requested):
~This does not work: this does not fix the issue. I've just gave it a try on your branch and the region header is still empty with `TEST_AWS_REGION_NAME=eu-west-1`, which makes the snapshot fail.
Also, the failure in the test would not be resolved with this change, as it seems we were getting an empty string? Very weird.~

I think this is an issue with the patching of the specs, by default Boto does not set the `BucketRegion` and this is something we patched. That might be related? When I try to regenerate snapshots, I don't even get the field anymore. 

I think it's safe to ignore snapshots here for the true intent of the test. 

Edit: the description of the PR is wrong, which led me into the wrong assumptions: 
This is the failing message: `#x1B[33m(~)#x1B[0m /BucketRegion 'eu-west-1' → '<region>' ... (expected → actual)` which like you said, shows the region of the client being transformed. This makes more sense now, and the fix would work. Still weird that I can't reproduce snapshots locally, this is a bit worrying.

I would still remove the snapshots all together in these kind of hardcoded regions, to avoid this issue. What do you think?

bentsku (commented):
I had requested changes, but it was from the wrong assumption of the header being empty and not replaced with `<region>`. I'll just comment here, tell me what you think about replacing snapshots in that case. This would work, but it would safer to remove: what do you think?

bentsku (approved):
Sorry for the spam here. In any case, the fix should work, so we can merge as is. Please beware of copying the right message for the failing test, it led into a bit of a rabbit hole trying to understand why we would return an empty string.

sannya-singal:
Sorry for the confusion @bentsku! I tried to look at the error message but I found it being correctly copied:
image

Due to special tags it was showing us an empty string,I have removed it now, thanks! 🙌

bentsku:
@sannya-singal that's really unlucky 😅 no worries, we can merge as is, thanks for fixing this! 🚀

PR #10509, opened 2024-03-21
switch from black to ruff format

alexrashed:
## Motivation
After https://github.com/localstack/localstack/pull/10502 was filed by Dependabot due to a security vulnerability in `black`, I thought it might be time to also switch to `ruff` for our code formatting (after switching the import sorting and linting with #9399).

This PR removes `black` as a dependency and uses `ruff format` instead.

## Changes
- Remove `black` from the dev dependencies.
- Use `ruff format` everywhere where `black` was used.
- Remove the black execution from cloudformation scaffolding.
  - This is a similar issue we had with the ASF scaffolding (see changes to the ASF update action in #9399).
  - I think formatting can easily be handled outside of the code generation, but I am happy for any suggestions! /cc @simonrw 
- Runs the formatting once, there are slight changes in the formatting.

## TODO
- [x] Update the pinned dependencies.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -12,6 +12,7 @@
   <a href="#sponsors"><img alt="Sponsors on Open Collective" src="https://opencollective.com/localstack/sponsors/badge.svg"></a>
   <a href="https://img.shields.io/pypi/l/localstack.svg"><img alt="PyPI License" src="https://img.shields.io/pypi/l/localstack.svg"></a>
   <a href="https://github.com/psf/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
+  <a href="https://github.com/astral-sh/ruff"><img alt="Ruff" src="https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json"></a>
   <a href="https://twitter.com/localstack"><img alt="Twitter" src="https://img.shields.io/twitter/url/http/shields.io.svg?style=social"></a>
 </p>
 
@@ -82,7 +83,7 @@ Create a queue using SQS with LocalStack's [`awslocal`](https://docs.localstack.
 
 ```
 $ awslocal sqs create-queue --queue-name test-queue
-$ awslocal sqs list-queues 
+$ awslocal sqs list-queues
 ```
 
 **Notes**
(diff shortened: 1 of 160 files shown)


DOCKER.md
```diff
@@ -12,6 +12,7 @@
   <a href="#sponsors"><img alt="Sponsors on Open Collective" src="https://opencollective.com/localstack/sponsors/badge.svg"></a>
   <a href="https://img.shields.io/pypi/l/localstack.svg"><img alt="PyPI License" src="https://img.shields.io/pypi/l/localstack.svg"></a>
   <a href="https://github.com/psf/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
```
silv-io, line 14: I think we can remove this now

Makefile
```diff
@@ -248,11 +248,11 @@ lint-modified:     		  ## Run code linter to check code style, check if formatte
 check-aws-markers:     		  ## Lightweight check to ensure all AWS tests have proper compatibilty markers set
 	($(VENV_RUN); python -m pytest --co tests/aws/)
 
-format:            		  ## Run ruff and black to format the whole codebase
-	($(VENV_RUN); python -m ruff check --output-format=full --fix .; python -m black .)
+format:            		  ## Run ruff to format the whole codebase
+	($(VENV_RUN); python -m ruff format .; python -m ruff check --output-format=full --fix .)
```
silv-io, line 252: If we only want to sort imports here with the linter, we can do only:

`ruff check --select I --fix`

README.md
```diff
@@ -16,6 +16,7 @@
   <a href="#sponsors"><img alt="Sponsors on Open Collective" src="https://opencollective.com/localstack/sponsors/badge.svg"></a>
   <a href="https://img.shields.io/pypi/l/localstack.svg"><img alt="PyPI License" src="https://img.shields.io/pypi/l/localstack.svg"></a>
   <a href="https://github.com/psf/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
```
silv-io, line 18: Here we could also remove the black shield

silv-io (approved):
Nice to consolidate our tooling :)

Just some small nits

README.md
alexrashed: That was actually a conscious decision to keep `black`. We still use the codestyle of `black` and it is still a great project.

Makefile
alexrashed: Actually, this is the case for a while now (and not really in the scope of this PR)... I guess I would also say that I expect the formatter changing the code by applying linting fixes? But that's definitely arguable...

DOCKER.md
alexrashed: Same as below, we are still using the `black` codestyle (and I would like to keep honoring black as a project)...

Makefile
silv-io: The formatter itself sadly can't sort imports:

`https://docs.astral.sh/ruff/formatter/#sorting-imports` 

I'd say linting goes deeper than formatting, so the formatter shouldn't apply those fixes.

alexrashed: This change here only replaces the usage of `black` with `ruff format`.
I agree that it might be arguable if the `format` make target should do that, but honestly I don't think this is in the scope of this PR.
Happy to discuss this in a follow-up PR from you as soon as this is through ;)

Issue #10473, opened 2024-03-17
bug: API Gateway works with GET HTTP integration method for Lambda Proxy, when it should only be POST


tinyg210:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

API Gateway will proxy requests to a Lambda function using the GET integration HTTP method, which does not work under normal circumstances on AWS. Trying to call the Lambda through the dedicated endpoint will result in an `Internal server error` message on AWS.

### Expected Behavior

Only the POST integration HTTP Method should be permitted for proxying requests to a Lambda.
The reason why this is important is that this is an essential configuration, and users who rely on LocalStack to test their infrastructure provisioning tools rely on this.
The [documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html) is complex and ambiguous, but here's the important part:
<img width="706" alt="Screenshot 2024-03-17 at 23 30 54" src="https://github.com/localstack/localstack/assets/49899593/dad20533-9215-49bb-96fa-8cee5e255c11">

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

1. Clone the repo: https://github.com/localstack-samples/cloud-pods-collaboration-demo
2. Run `docker compose up`
3. Navigate to `terraform/main.tf`
4. Find the resource `"aws_iam_policy" "lambda_dynamodb_policy"` and uncomment the action  `"dynamodb:GetItem"`.
5. Find the resource `"aws_api_gateway_integration" "get_product_integration"` and modify the `integration_http_method` from POST to GET 
[The previous 2 steps exist because this repo demos a different feature]
6. Run the Terraform config file using `terraform-local`. There's a script in the same folder that covers all steps (run-tflocal.sh)
7. In the root folder, run the `invoke.sh` script. It does 3 curl calls: POST, POST, GET -- the last one should fail
8. The output of the script is:
`Product added/updated successfully. Product added/updated successfully. {"price":"29.99","name":"EcoFriendly Water Bottle","description":"A durable, eco-friendly water bottle designed to keep your drinks cold for up to 24 hours and hot for up to 12 hours. Made from high-quality, food-grade stainless steel, it's perfect for your daily hydration needs.","id":"34534"}⏎  `
And it should be:
`Product added/updated successfully. Product added/updated successfully. Internal server error⏎`
Because the last method should not work.

### Environment

```markdown
- OS: macOS Sonoma 14.4
[...]

Issue #8273, opened 2023-05-05
bug: API Gateway works with the wrong integration http method for Lambda Proxy


evbo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

API Gateway will work without a hitch proxying requests to a Lambda function using an `ANY` integration HTTP Method set, which should never work normally.

### Expected Behavior

[as documented](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-integrations.html) only a `POST` integration HTTP Method should be acceptable for proxying requests to a Lambda.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS:ARM64/Ubuntu22.04
- LocalStack:2.0.2-arm64
```

### Anything else?

You should clarify free version doesn't enforce "Resource Policies" in addition to IAM Role Policies. I don't think I saw that documented, but that is another issue with API Gateway Proxies I didn't list here because I presumed only the Pro version would support that.

PR #10512, opened 2024-03-21
Api gateway integration proxy only accepts post method

cloutierMat:
## Motivation

Apigateway's integration with AWS_PROXY can only use POST as integration method. See [AWS Documentation](api-gateway-integration-proxy-only-accepts-POST-method) and documented on issue #10473 . Currently, we were accepting and forwarding any calls with consideration for this expected error to be returned.

## Changes

Aws does allow for the Proxy Integration to be created even with the wrong method. And returning a Internal Server Error when a call is attempted. This fix will mimic this behavior.

## Testing

One `aws.validated` test was added to confirm the error message. Many other `aws.unknown` tests were failing as they were creating integration with `integrationHttpMethod` not equal to `POST`. One of these test was removed but in view of the upcoming rework of the service, the other tests simply got patched to use the POST method and will be left to properly refactor in `aws.validated` tests.

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_basic.py
+++ b/tests/aws/services/apigateway/test_apigateway_basic.py
@@ -1799,7 +1799,7 @@ def test_rest_api_multi_region(
         resourceId=resource_eu_id,
         httpMethod=method,
         type="AWS_PROXY",
-        integrationHttpMethod=method,
+        integrationHttpMethod="POST",
         uri=uri_eu,
     )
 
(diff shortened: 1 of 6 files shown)


tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -221,57 +220,6 @@ def test_update_rest_api_deployment(self, create_rest_apigw, aws_client):
         )
         assert deployment["description"] == "new-description"
 
-    @markers.aws.unknown
-    def test_api_gateway_lambda_integration(
```
bentsku, line 225: Nice cleanup! 🧹 awesome to see them either deleted or rewritten!

bentsku (approved):
LGTM! This is great, really awesome to see this finally fixed. There was actually another GitHub issue related to see that I found: #8273. 

The fix is really clean, I like the separation into a new method. I really have nothing to change on this PR, which is.. well, really awesome. The speed at which you're picking up the tools is insane 🚀

Issue #10481, opened 2024-03-18
bug: Cloudformation doesn't support Fn::GetAtt dot.notation


lukestoward:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm not sure whether this is technically valid syntax, however it is possible to use `Fn::GetAtt SQSDeadLetterQueue.Arn` in a CloudFormation template when deploying to AWS directly.

However, when using LocalStack via the latest Docker image the above function results in the following error:

```
WARN --- [functhread12] l.s.c.engine.entities      : CFn resource failed to deploy: SQSQueue (Resource 'S' not found in stack example-sqs-stack)
```

Using `Fn::GetAtt [SQSDeadLetterQueue, Arn]` and `!GetAtt SQSDeadLetterQueue.Arn` works as expected. Raising this to highlight the disparity between LocalStack and remote AWS CloudFormation behaviour.

### Expected Behavior

I would expect the following syntax `Fn::GetAtt SQSDeadLetterQueue.Arn` to work in LocalStack.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Using the following files:

`template-example.yaml`

```yaml
AWSTemplateFormatVersion: 2010-09-09

Resources:
  SQSDeadLetterQueue:
    Type: AWS::SQS::Queue
    Properties:
      QueueName:
        Fn::Sub: example-dlq

  SQSQueue:
    Type: AWS::SQS::Queue
    Properties:
      QueueName:
        Fn::Sub: example-queue
      RedrivePolicy:
        deadLetterTargetArn:
          Fn::GetAtt: SQSDeadLetterQueue.Arn
        maxReceiveCount: 3
```

`init-localstack.sh`
```bash
#!/bin/bash

set -e

awslocal cloudformation deploy \
    --template-file /template-example.yaml \
    --stack-name example-sqs-stack \
    --region=eu-west-1
```

`docker-compose.yml`
```yaml
version: '3.9'
services:
  localstack:
    image: localstack/localstack:latest
    environment:
      SERVICES: cloudformation,sqs
      LS_LOG: info
    volumes:
      - "./init-localstack.sh:/etc/localstack/init/ready.d/init-localstack.sh"
      - "./template-example.yaml:/template-example.yaml"
```

And running `docker compose up localstack` you should see the error mentioned above.

### Environment

```markdown
- OS: macOS Ventura 13.6.4
[...]

PR #10514, opened 2024-03-21
add support for getAtt string parameter in dependency

pinzon:
## Motivation
There is limitation in Localstack's cloudformation implementation (#10481) where it is unable to successfully deploy a resource that utilizes an attribute of another resource when the GetAtt function is used with a string containing a dot instead of an array of parameters.

## Changes
- fix the function that list the dependencies of a resource

## Testing
- snapshoted test to validate the behavior


Morijarti (approved):
LGTM

PR #10515, opened 2024-03-22
add limited number of regions to random creds

sannya-singal:
## Motivation
As per our recent discussions about how to handle the randomised region for cross-accounts workflow considering the fact that not all services are available in all regions we need to limit the number of regions in our scheduled workflow. 

## Changes
This PR limits the number of regions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ commands:
                   export TEST_AWS_ACCOUNT_ID=$(LC_ALL=C tr -dc '0-9' < /dev/urandom | fold -w 12 | head -n 1)
                   export TEST_AWS_ACCESS_KEY_ID=$TEST_AWS_ACCOUNT_ID
                   # Set TEST_AWS_REGION_NAME to a random AWS region other than us-east-1
-                  export AWS_REGIONS=("me-south-1" "me-central-1" "sa-east-1" "us-east-2" "us-west-1" "us-west-2" "eu-west-1" "eu-west-2" "eu-west-3" "ap-east-1" "ap-south-2" "ap-south-1" "ap-southeast-1" "ap-southeast-2" "ap-northeast-1" "ap-northeast-2" "ap-northeast-3" "sa-east-1" "af-south-1" "ca-central-1" "ca-west-1" "eu-central-1" "eu-south-1" "eu-south-2"	"eu-north-1" "eu-central-2" "il-central-1" )
+                  export AWS_REGIONS=("us-east-2", "us-west-1", "us-west-2")
                   export TEST_AWS_REGION_NAME=${AWS_REGIONS[$RANDOM % ${#AWS_REGIONS[@]}]}
                   echo "export TEST_AWS_REGION_NAME=${TEST_AWS_REGION_NAME}" >> $BASH_ENV
                   echo "export TEST_AWS_ACCESS_KEY_ID=${TEST_AWS_ACCESS_KEY_ID}" >> $BASH_ENV
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ commands:
                   export TEST_AWS_ACCOUNT_ID=$(LC_ALL=C tr -dc '0-9' < /dev/urandom | fold -w 12 | head -n 1)
                   export TEST_AWS_ACCESS_KEY_ID=$TEST_AWS_ACCOUNT_ID
                   # Set TEST_AWS_REGION_NAME to a random AWS region other than us-east-1
-                  export AWS_REGIONS=("us-east-2" "us-west-1" "us-west-2")
+                  export AWS_REGIONS=("us-east-2" "us-west-1" "us-west-2" "ap-southeast-1" "ap-southeast-2" "ap-northeast-1" "eu-central-1" "eu-west-1")
                   export TEST_AWS_REGION_NAME=${AWS_REGIONS[$RANDOM % ${#AWS_REGIONS[@]}]}
                   echo "export TEST_AWS_REGION_NAME=${TEST_AWS_REGION_NAME}" >> $BASH_ENV
                   echo "export TEST_AWS_ACCESS_KEY_ID=${TEST_AWS_ACCESS_KEY_ID}" >> $BASH_ENV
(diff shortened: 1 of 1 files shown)


alexrashed (commented):
Limiting the regions is fair considering that not all services are supported in all regions, and we should not allow requests to these regions as well (for parity), but it would be great to have the biggest spread possible while limiting the regions.
Are there other combinations which still support all the services we implement, but cover more cases?

sannya-singal:
> Limiting the regions is fair considering that not all services are supported in all regions, and we should not allow requests to these regions as well (for parity), but it would be great to have the biggest spread possible while limiting the regions. Are there other combinations which still support all the services we implement, but cover more cases?

Updated the list which now covers regions from EU and Asia as well.

alexrashed (commented):
This set is looking good now, but do you know that all these regions should be working with all services we implement?
Did you make sure that all services implemented in LocalStack are available in all these regions? Do you maybe have a script for that? Would be great to preserve the strategy how you define these regions here as well.

alexrashed (approved):
I wanted to jump on the exercise and wrote a quick script to double check if the regions you selected are really supported by all services implemented in LocalStack:
```python
import boto3

# define regions you want to use in the multi-region tests
regions_to_test = {"us-east-2", "us-west-1", "us-west-2", "ap-southeast-1", "ap-southeast-2", "ap-northeast-1", "eu-central-1", "eu-west-1"}

# Community health endpoint data: localhost:4566/_localstack/health
health_endpoint_data = {"services": {"acm": "available", "apigateway": "available", "cloudformation": "available", "cloudwatch": "available", "config": "available", "dynamodb": "available", "dynamodbstreams": "available", "ec2": "available", "es": "available", "events": "available", "firehose": "available", "iam": "available", "kinesis": "available", "kms": "available", "lambda": "available", "logs": "available", "opensearch": "available", "redshift": "available", "resource-groups": "available", "resourcegroupstaggingapi": "available", "route53": "available", "route53resolver": "available", "s3": "available", "s3control": "available", "scheduler": "available", "secretsmanager": "available", "ses": "available", "sns": "available", "sqs": "available", "ssm": "available", "stepfunctions": "available", "sts": "available", "support": "available", "swf": "available", "transcribe": "available"}, "edition": "community", "version": "3.2.1.dev"}

# extract the list of services
services = list(health_endpoint_data["services"].keys())
# get the regions for all services (do not allow non-regional to filter them in the next step)
regions_per_service = {service: boto3.Session().get_available_regions(service, allow_non_regional=False) for service in services}
# remove services without any regions (including non-regional services)
regions_per_service = {service: regions for service, regions in regions_per_service.items() if len(regions) != 0}

first_service, first_regions = next(iter(regions_per_service.items()))
print(f"Starting with regions for service {first_service}: {first_regions}")
regions_with_all_services = set(first_regions)

for service, regions in regions_per_service.items():
    print(f"Service: {service} / Regions: {regions}")
    regions_with_all_services.intersection_update(regions)
    print(f"Remaining regions: {regions_with_all_services}")
    print("---")

# check if regions used for tests are in all regions
illegal_test_regions = regions_to_test - regions_with_all_services

if len(illegal_test_regions) == 0:
    print("Test regions are supported in all services!")
else:
    print(f"The following regions are not supported in at least one supported service: {illegal_test_regions}")
```

And the result is... 🥁
```
Test regions are supported in all services!
```

sannya-singal:
Even I wrote a script but I used SSM instead since its more reliable and not the boto3 session cc: @alexrashed.

alexrashed:
Perfect! Would be great to drop the script here to preserve it in case we want to re-evaluate the regions at a later point in time :)

sannya-singal:
```
import boto3
import os

ParamPath = "/aws/service/global-infrastructure/regions/{}/services"

def fetch_enabled_services_in_region(region, session):
    services = {}
    ssm_client = session.client('ssm')
    NextToken = ''
    while True:
        req = ssm_client.get_parameters_by_path(
            Path=ParamPath.format(region),
            NextToken=NextToken
        )
        resp = req['Parameters']
        if resp:
            for param in resp:
                if 'Value' in param and param['Value'] is not None:
                    srv = format_service_name(param['Value'])
                    services[srv] = True
        NextToken = req.get('NextToken', '')
        if not NextToken:
            break
    return services

def format_service_name(s):
    name = s.strip()
    name = ''.join(c if c.isalnum() else ' ' for c in name)
    name = name.lower()
    name = name.replace(" ", "")
    return name

def get_services():
    # this list was taken programmatically from our user-guide docs
    return ['kinesis', 'pinpoint', 'organizations', 'ssm', 'secretsmanager', 'mwaa', 'kms', 'elastictranscoder', 'cloudwatch', 'glue', 's3', 'emr', 'sns', 'resource_groups', 'fis', 'config', 'cognito', 'memorydb', 'cloudfront', 'opensearch', 'glacier', 'serverlessrepo', 'elasticbeanstalk', 'neptune', 'transfer', 'qldb', 'ecr', 'dynamodb', 'applicationautoscaling', 'ecs', 'elb', 'sqs', 'timestream', 'iot', 'amplify', 'redshift', 'cloudtrail', 'pipes', 'appconfig', 'iam', 'mediastore', 'logs', 'kinesisanalytics', 'identitystore', 'textract', 'eks', 'support', 'msk', 'batch', 'appsync', 'lambda', 'managedblockchain', 'servicediscovery', 'waf', 'mq', 'es', 'docdb', 'firehose', 'ses', 'account', 'events', 'backup', 'cloudformation', 'elasticache', 'ce', 'rds', 'swf', 'transcribe', 'stepfunctions', 'codecommit', 'xray', 'apigateway', 'ram', 'efs', 'sts', 'sagemaker', 'acm', 'athena', 'route53', 'ec2']


def find_uncommon_elements(list1, list2):
    set1 = set(list1)
    set2 = set(list2)
    return set1 - set2


if __name__ == "__main__":
    services_localstack = get_services()
    session = boto3.Session()
    service_region = {}
    for region in ["us-west-1", "us-east-2", "us-west-2", "ap-southeast-1", "ap-southeast-2", "ap-northeast-1", "eu-central-1", "eu-west-1"]:
        services = fetch_enabled_services_in_region(region, session)
        service_region[region] = services
    
    for region, services in service_region.items():
        services_list = list(services.keys())
        if(set(services_localstack).issubset(set(services_list))):
            print("Services are same!")
        else: 
            print(f'Missing services for {region}: ', find_uncommon_elements(services_localstack, services_list))
```

Please note that the above shared data is including the pro services.

alexrashed (approved):

PR #10524, opened 2024-03-23
Adding missing parameter for authorization scopes

giograno:
## Motivation
The `AWS::ApiGateway::Method` resource was not using the `authorizedScopes` parameter.
This led to a bug in APIGW Cognito Authorizer as described in the [companion PR](https://github.com/localstack/localstack-ext/pull/2776).


bentsku (approved):
LGTM! Thanks for tackling this 🚀

PR #10530, opened 2024-03-25
Reduce the range of dependencies to speed up pinning

silv-io:
## Motivation

In `localstack-ext`, the dependency pinning can take a very long time now. This is connected to the switch to `pyproject.toml` as well as to backtracking a lot inside the dependencies of `localstack-core`. 

## Changes

This PR reduces the admissible range for the following packages, so that the backtracking is reduced:
- `awscli>=1.32.69`
- `boto3-stubs>=1.34.69`

## Testing

Locally, this achieves a reduction in the pinning of `localstack-ext` from 45min to 6min30 on my machine. 

To check it locally, you can replace the dependencies to `localstack-core` in `localstack-ext` to reference this instead:

`localstack-core @ file://localhost/{absolute-path-to-package}#egg=localstack_core`

However, effectively checking the performance is not possible this way, as using such a specific version of `localstack-core` here will lead to performance improvements without changes as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ runtime = [
     "antlr4-python3-runtime==4.13.1",
     "apispec>=5.1.1",
     "aws-sam-translator>=1.15.1",
-    "awscli>=1.22.90",
+    "awscli>=1.32.69",
     "crontab>=0.22.6",
     "cryptography>=41.0.5",
     "json5>=0.9.11",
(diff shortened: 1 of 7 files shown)


alexrashed (approved):
Awesome! This will speed up the dependency pinning a lot! In fact, we should try to also move to exact pins for the boto3 types and the aws cli (since they are directly bound to `botocore` - which is fully pinned - anyways).
But this will ease the pain with the make target taking ages quite a lot! 🚀 🧹 🚤

PR #10534, opened 2024-03-26
fix `test_different_location_constraint` for `us-east-2`

sannya-singal:
## Motivation
Test `test_different_location_constraint` fails for `us-east-2` and throws the following error: 

```
__________________ TestS3.test_different_location_constraint ___________________
>> match key: get_bucket_location_bucket_3
        (~) /LocationConstraint 'us-east-2' → '<region>' ... (expected → actual)

        ["$..LocationConstraint"]
>> match key: get_bucket_location_bucket_2
        (~) /LocationConstraint 'us-east-2' → '<region>' ... (expected → actual)

        ["$..LocationConstraint"]
```
when the scheduled flow was triggered: https://app.circleci.com/pipelines/github/localstack/localstack/23638/workflows/ee54cc00-9770-42db-af01-6fed44c5085f/jobs/193748 for `TEST_AWS_REGION_NAME=us-east-2`

## Changes
This PR adds regex transformer for to region and regenerates the snapshots.

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -2545,6 +2545,7 @@ class TestS3:
         s3_create_bucket,
         aws_client_factory,
         s3_create_bucket_with_client,
+        region_name,
         snapshot,
         aws_client,
     ):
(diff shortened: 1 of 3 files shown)


bentsku (approved):
LGTM! We will need to hunt every single hardcoded region in our tests. I'll try to keep it in mind when writing new tests. Thanks for tackling this!

sannya-singal:
Thanks for the review @bentsku!

PR #10537, opened 2024-03-26
extend pinning in ASF update action

alexrashed:
## Motivation
We recently had issues with the newly introduced dependency pin updates.
Due to an unknown reason pip started backtracking _a lot_.
This is why we updated some of our dependency constraints to limit the search space in https://github.com/localstack/localstack/pull/10530.
This PR goes one step further and increases the amount of packages explicitly pinned to a single version by the ASF update action.
The ASF update action directly pins `botocore` since the generated ASF APIs are directly coupled to the specs which have been used to generated them (which are contained in `botocore`).
But: `aws-cli`, `boto3`, and `boto3-stubs` are in turn directly coupled to `botocore`.
This PR extends the pinning performed in the ASF update action to these dependencies.

## Changes
- Directly pin `aws-cli`, `boto3`, and `boto3-stubs`.
- Update these pins together with the `botocore` pin in the ASF update action.
- Fixes the release helper script regex to allow updating dependencies with extras which contain commas and dashes (`boto3-stubs`).
- Migrates to `--strip-extras` for all `pip-compile` calls, as it makes sense and will be the default with the next major release anways.

## Testing
- Tested the whole part of the workflow ("Update botocore and transitive pins") locally.

## TODO
- [x] Test the local runtime of pinning the dependencies with these exact pins.
  - This revealed a weird error which seems to be reproducible on the branch now:
    ```
    × python setup.py egg_info did not run successfully.
    │ exit code: 1
    ╰─> [10 lines of output]
          File "<string>", line 2, in <module>
          File "<pip-setuptools-caller>", line 34, in <module>
          File "/tmp/pip-resolve-k6a5y0km/awscli_05b2471b169b416d899f78a12a76bdd5/setup.py", line 47, in <module>
            data_files=build_doclist(),
                       ^^^^^^^^^^^^^^^
          File "/tmp/pip-resolve-k6a5y0km/awscli_05b2471b169b416d899f78a12a76bdd5/setup.py", line 24, in build_doclist
            os.path.walk('doc/text', visitfn, doclist)
            ^^^^^^^^^^^^
        AttributeError: module 'posixpath' has no attribute 'walk'
        [end of output]
    ```
   - This issue was caused by the wrong usage of `--upgrade-package` in `pip-compile`. Unfortunately it seems to ignore the constraints in input source files.


bin/release-helper.sh
```diff
@@ -158,8 +158,8 @@ function cmd-set-dep-ver() {
     dep=$1
     ver=$2
 
-    grep -Eh "^(\s*\")${dep}(\[[a-zA-Z0-9]+\])?(>|=|<)(.*\",)" ${DEPENDENCY_FILE} || { echo "dependency ${dep} not found in ${DEPENDENCY_FILE}"; return 1; }
-    sed -i -r "s/^(\s*\")(${dep})(\[[a-zA-Z0-9,]+\])?(>|=|<)(.*\",)/\1\2\3${ver}\",/g" ${DEPENDENCY_FILE}
+    grep -Eh "^(\s*\")${dep}(\[[a-zA-Z0-9,\-]+\])?(>|=|<)(.*\",)" ${DEPENDENCY_FILE} || { echo "dependency ${dep} not found in ${DEPENDENCY_FILE}"; return 1; }
```
silv-io, line 161: in addition to alphanumeric characters and `-`, also `_` and `.` are admissible in dependency names afaik

pyproject.toml
```diff
@@ -47,9 +47,11 @@ Issues = "https://github.com/localstack/localstack/issues"
 [project.optional-dependencies]
 # minimal required to actually run localstack on the host for services natively implemented in python
 base-runtime = [
-    "awscrt>=0.13.14",
-    "boto3>=1.26.121",
+    # pinned / updated by ASF update action
```
silv-io, line 50: Good addition!

silv-io (approved):
LGTM! Just one further suggestion for the release-helper function :)

bin/release-helper.sh
alexrashed: Thanks! I'll keep it for now though, I don't want to sneak too much in this change here for now... The regex is already complex enough 😄

alexrashed:
I'll move this PR back into draft due to some weird errors which occurred during testing (See PR description).

requirements-typehint.txt
```diff
@@ -60,7 +60,7 @@ boto3==1.34.74
     #   aws-sam-translator
     #   localstack-core
     #   moto-ext
-boto3-stubs[acm,acm-pca,amplify,apigateway,apigatewayv2,appconfig,appconfigdata,application-autoscaling,appsync,athena,autoscaling,backup,batch,ce,cloudcontrol,cloudformation,cloudfront,cloudtrail,cloudwatch,codecommit,cognito-identity,cognito-idp,dms,docdb,dynamodb,dynamodbstreams,ec2,ecr,ecs,efs,eks,elasticache,elasticbeanstalk,elbv2,emr,emr-serverless,es,events,firehose,fis,glacier,glue,iam,identitystore,iot,iot-data,iotanalytics,iotwireless,kafka,kinesis,kinesisanalytics,kinesisanalyticsv2,kms,lakeformation,lambda,logs,managedblockchain,mediaconvert,mediastore,mq,mwaa,neptune,opensearch,organizations,pi,pipes,qldb,qldb-session,rds,rds-data,redshift,redshift-data,resource-groups,resourcegroupstaggingapi,route53,route53resolver,s3,s3control,sagemaker,sagemaker-runtime,secretsmanager,serverlessrepo,servicediscovery,ses,sesv2,sns,sqs,ssm,sso-admin,stepfunctions,sts,timestream-query,timestream-write,transcribe,wafv2,xray]==1.34.75
+boto3-stubs==1.34.74
```
silv-io, line 63: very satisfying :)

silv-io (approved):
LGTM! Stripping the extras would also be nice for a possible future switch to `uv` as it strips the extras by default as well.

PR #10538, opened 2024-03-26
fix apigw tests for `eu-west-1`

sannya-singal:
## Motivation
There were failing api gateway tests for `eu-west-1` in the workflow: https://app.circleci.com/pipelines/github/localstack/localstack/23655/workflows/49291732-7851-49ea-917a-2a293dde3288/jobs/193896/tests

## Changes
This PR:
- removes the static `uri` value hardcoded for `eu-west-1`
- regenerates the snapshots

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_extended.snapshot.json
+++ b/tests/aws/services/apigateway/test_apigateway_extended.snapshot.json
@@ -1,6 +1,6 @@
 {
   "tests/aws/services/apigateway/test_apigateway_extended.py::test_export_swagger_openapi[TEST_IMPORT_PETSTORE_SWAGGER]": {
-    "recorded-date": "26-03-2024, 09:53:07",
+    "recorded-date": "26-03-2024, 11:32:19",
     "recorded-content": {
       "import-api": {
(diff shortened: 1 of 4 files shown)


tests/aws/services/apigateway/test_apigateway_extended.snapshot.json
```diff
@@ -650,6 +651,7 @@
         },
         "id": "<api-id:1>",
         "name": "Simple PetStore (Swagger)",
+        "rootResourceId": "m7b1ab5lr9",
```
bentsku, line 654: nit: could you add a transformer for this value, so that it doesn't change every time we run test? We will still need to have it in the `skip_snapshot_verify` list because we don't support it yet, but at least it wouldn't change when regenerating.
`snapshot.add_transformer(snapshot.transform.key_value("rootResourceId"))`
I think it will be needed in both tests. Thanks a lot!

bentsku (commented):
LGTM, one comment about some new resource that is not transformed, but once this is addressed we can merge. Thanks for fixing this!

bentsku (approved):

PR #10539, opened 2024-03-26
fix s3 and sns tests for `ap-southeast-1`

sannya-singal:
## Motivation
There were failing sns and s3 tests for `ap-southeast-1` in the workflow: https://app.circleci.com/pipelines/github/localstack/localstack/23655/workflows/49291732-7851-49ea-917a-2a293dde3288/jobs/193896/tests 

The issue is caused when secondary and primary regions are same. We have set the [TEST_AWS_SECONDARY_REGION_NAME](https://github.com/localstack/localstack/blob/b7f34c67c49c8f8d4aa6e0b08584b248501a64bb/localstack/constants.py#L158) to `ap-southeast-1`.

## Changes
This PR:
- adds fallback if primary and secondary regions are same

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -409,7 +409,9 @@ class TestSNSPublishCrud:
         # create a client in another region
         sns_client_region_2 = aws_client_factory.get_client(
             service_name="sns",
-            region_name=secondary_region_name if region_name != "ap-southeast-1" else "us-east-2",
+            region_name=secondary_region_name
+            if region_name != secondary_region_name
+            else "us-east-2",
         )
 
         message = "This is a test message"
@@ -4746,7 +4748,7 @@ class TestSNSRetrospectionEndpoints:
             api_platform_endpoints_msgs[endpoint_arn][0]["MessageAttributes"] == message_attributes
         )
 
-        region_2 = secondary_region_name if region_name != "ap-southeast-1" else "us-east-2"
+        region_2 = secondary_region_name if region_name != secondary_region_name else "us-east-2"
 
         # Ensure you can select the region
         msg_with_region = requests.get(
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -395,7 +395,13 @@ class TestSNSPublishCrud:
 
     @markers.aws.validated
     def test_topic_publish_another_region(
-        self, sns_create_topic, snapshot, aws_client, aws_client_factory, secondary_region_name
+        self,
+        sns_create_topic,
+        snapshot,
+        aws_client,
+        aws_client_factory,
+        region_name,
+        secondary_region_name,
     ):
         # create the topic in the default region, so that it's easier to clean up with the fixture
         topic_arn = sns_create_topic()["TopicArn"]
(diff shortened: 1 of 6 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -1105,7 +1111,7 @@ def test_redrive_policy_lambda_subscription(
         snapshot.match("messages", response)
 
     @markers.aws.validated
-    @pytest.mark.parametrize("signature_version", ["1", "2"])
+    @pytest.mark.parametrize("signature_version", ["2"])
```
bentsku, line 1114: is there a specific reason why this was changed?

tests/aws/services/sns/test_sns.py
```diff
@@ -395,15 +395,21 @@ def test_unknown_topic_publish(self, sns_create_topic, snapshot, aws_client):
...
+        secondary_region_name,
     ):
         # create the topic in the default region, so that it's easier to clean up with the fixture
         topic_arn = sns_create_topic()["TopicArn"]
 
         # create a client in another region
         sns_client_region_2 = aws_client_factory.get_client(
             service_name="sns",
-            region_name=secondary_region_name,
+            region_name=secondary_region_name if region_name != "ap-southeast-1" else "us-east-2",
```
bentsku, line 412: I don't understand this change. Why `ap-southeast-1`?

bentsku (changes requested):
The changes with random regions is nice, but for tests specifically verifying cross-region behavior, it seems it gets tricky.

I'm not sure I fully understood: can you have both the primary and secondary region the same in your randomized tests? Doesn't that defeat the whole purpose? Aren't those made to test cross-region behavior? I would make sure those can't be the same. It seems that tests will for sure fail if we want to test cross-region and we have the same region..

It seems the `validation` file have not been updated, did the tests run against AWS?
The PR description also mention updating the snapshots and the transformers but there are none for this PR. 

Also, it seems a parametrize test have been changed but there are no comments about that.

tests/aws/services/sns/test_sns.py
sannya-singal: My bad, this was an accidental commit. I have updated it now, sorry for the confusion!

tests/aws/services/sns/test_sns.py
sannya-singal: For now we just have randomised `TEST_AWS REGION_NAME` which is the primary region getting randomly generated and not the `SECONDARY_TEST_AWS_REGION_NAME`.

The default value we set for our tests of SECONDARY_TEST_AWS_REGION_NAME is `ap-southeast-1 `, so currently it is possible in the circleci workflow to have these regions same. Its already planned for the next iteration to add randomly generated secondary region in the workflow.

bentsku: Should we maybe update your change to check against `SECONDARY_TEST_AWS_REGION_NAME` then and not hardcoded `ap-southeast-1`? So it will also work if we change `SECONDARY_TEST_AWS_REGION_NAME` somehow?

bentsku (commented):
Looking at the CircleCI workflow, I believe removing `ap-southeast-1` from the primary region selection would be simpler to fix this issue.

I believe having `region_name` and `secondary_region_name` be 2 different regions is reasonable expectation, and breaking this goes against the meaning of those fixtures. 

Having a temporary fix in the workflow by not allowing `ap-southeast-1` be a primary region because the current secondary region is hardcoded to it. Once we can have a secondary random selection, then we could add it back. However, we should always make sure primary and secondary are not the same region, to not break expectations.

sannya-singal:
https://github.com/localstack/localstack/pull/10540

PR #10542, opened 2024-03-26
Add version.py file to list of added git files in release helper

silv-io:
## Motivation

The release helper currently does not commit the generated `version.py` file. (Which is created everywhere consistently with the change in the release workflow)

## Changes

It adds the `version.py` file to the commits for releasing and incrementing development iteration

## Testing

Check the changes in the release workflow for guides on how to test this

### DIFF
--- a/bin/release-helper.sh
+++ b/bin/release-helper.sh
@@ -3,7 +3,7 @@
 set -e
 
 VERSION_FILE=${VERSION_FILE-VERSION}
-VERSION_PACKAGE=${VERSION_PACKAGE-*/version.py}
+VERSION_PY=${VERSION_PY-*/version.py}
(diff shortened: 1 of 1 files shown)


bin/release-helper.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 
 VERSION_FILE=${VERSION_FILE-VERSION}
+VERSION_PACKAGE=${VERSION_PACKAGE-*/version.py}
```
alexrashed, line 6: nit: I think `VERSION_PACKAGE` is a bit of a misleading name. Maybe just `VERSION_PY`?

alexrashed (approved):
Great catch! Thanks for Awesome! Thanks for jumping on this and making sure that our release on Thursday works smoothly! 🎖️ 🚀 🧹

PR #10545, opened 2024-03-26
add update operations for AWS::SNS::Subscription and AWS::SQS::QueuePolicy. 

pinzon:
## Motivation
This pull request (PR) addresses a feature request from a customer. It introduces the update operation functionality for two AWS resource types: AWS::SNS::Subscription and AWS::SQS::QueuePolicy. 

## Changes
- update operation for AWS::SNS::Subscription
- update operation for AWS::SQS::QueuePolicy

## Testing
- tests that create the resource in a CFn stack and update them.

### DIFF
--- a/localstack/services/sns/resource_providers/aws_sns_subscription.py
+++ b/localstack/services/sns/resource_providers/aws_sns_subscription.py
@@ -66,9 +66,6 @@ class SNSSubscriptionProvider(ResourceProvider[SNSSubscriptionProperties]):
 
         params = util.select_attributes(model=model, params=["TopicArn", "Protocol", "Endpoint"])
 
-        def attr_val(val):
-            return json.dumps(val) if isinstance(val, (dict, list)) else str(val)
-
         attrs = [
             "DeliveryPolicy",
             "FilterPolicy",
@@ -76,7 +73,7 @@ class SNSSubscriptionProvider(ResourceProvider[SNSSubscriptionProperties]):
             "RawMessageDelivery",
             "RedrivePolicy",
         ]
-        attributes = {a: attr_val(model[a]) for a in attrs if a in model}
+        attributes = {a: self.attr_val(model[a]) for a in attrs if a in model}
 
         if attributes:
             params["Attributes"] = attributes
@@ -130,6 +127,7 @@ class SNSSubscriptionProvider(ResourceProvider[SNSSubscriptionProperties]):
 
         """
         model = request.desired_state
+        model["Id"] = request.previous_state["Id"]
         sns = request.aws_client_factory.sns
 
         attrs = [
@@ -144,11 +142,14 @@ class SNSSubscriptionProvider(ResourceProvider[SNSSubscriptionProperties]):
                 sns.set_subscription_attributes(
                     SubscriptionArn=model["Id"],
                     AttributeName=a,
-                    AttributeValue=json.dumps(model[a]),
+                    AttributeValue=self.attr_val(model[a]),
                 )
-
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
             resource_model=model,
             custom_context=request.custom_context,
         )
+
+    @staticmethod
+    def attr_val(val):
+        return json.dumps(val) if isinstance(val, dict) else str(val)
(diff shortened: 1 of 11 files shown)


tests/aws/services/cloudformation/resources/test_sqs.py
```diff
@@ -120,3 +122,34 @@ def test_update_queue_no_change(deploy_cfn_template, aws_client, snapshot):
...
+
+    updated_stack = deploy_cfn_template(
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../../templates/sqs_with_queuepolicy_updated.yaml"
+        ),
+        is_update=True,
+        stack_name=stack.stack_name,
+    )
+
+    if is_aws_cloud():
```
Morijarti, line 148: Could we perhaps use waiter or some other way to check if changes have been deployed?

Morijarti (approved):
Always good to see more support for updates being added :)
Good work.

LGTM

localstack/services/sns/resource_providers/aws_sns_subscription.py
```diff
@@ -128,6 +125,31 @@ def update(
...
+                )
+        return ProgressEvent(
+            status=OperationStatus.SUCCESS,
+            resource_model=model,
+            custom_context=request.custom_context,
+        )
+
+    @staticmethod
+    def attr_val(val):
+        return json.dumps(val) if isinstance(val, dict) else str(val)
```
bentsku, line 155: it might be useful to also properly cast boolean values into lower cased string here with `localstack.utils.strings.canonicalize_bool_to_str`if the val is `bool`? 
Something like the following?
```suggestion
        if isinstance(val, dict):
            return json.dumps(val)
        elif isinstance(val, bool):
            return canonicalize_bool_to_str(val)
        else:
            return str(val)
```

bentsku (commented):
Nice seeing new update operations! Thanks for tackling this! Just one comment regarding a skip snapshot verify that seems to be the only thing the test is checking (apart that the operation is supported). Maybe it's the casing, added a comment about it. Once this is addressed I would this is good to go!

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -100,3 +100,35 @@ def test_deploy_stack_with_sns_topic(deploy_cfn_template, aws_client):
     rs = aws_client.sns.list_topics()
     topics = [tp for tp in rs["Topics"] if tp["TopicArn"] == topic_arn]
     assert not topics
+
+
+@markers.aws.validated
+@markers.snapshot.skip_snapshot_verify(paths=["$..Attributes.RawMessageDelivery"])
```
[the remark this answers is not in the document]
pinzon: The thing is that the AWS behavior is to return the boolean capitalized, True or False but the moto behavior is to return 'true' or 'false'

localstack/services/sns/resource_providers/aws_sns_subscription.py
pinzon: in the resource providers we avoid importing outside of cloudformation since in the future the resource providers will be allocated in a new repo

tests/aws/services/cloudformation/resources/test_sns.py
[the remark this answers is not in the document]
pinzon: update: used a  json path transformer to avoid skipping it

localstack/services/sns/resource_providers/aws_sns_subscription.py
bentsku: I see! We should use the same logic as the helper though in the resource provider, because it's the right behavior to lower case boolean values I think for SNS

tests/aws/services/cloudformation/resources/test_sns.py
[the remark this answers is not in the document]
bentsku: Understood, I think it's missing parity in SNS, even if we receive an upper case value, the provider should save it lower case. I'll fix it in SNS, we can remove the skip then

bentsku (approved):
LGTM! We will be able to remove the snapshot skip/snapshot transformer once the parity issue is fixed in the SNS provider, I'll open a quick PR about it! thanks for making this visible 😄 

PR here: #10575

PR #10549, opened 2024-03-27
fix `test_function_arns` for `eu-central-1`

sannya-singal:
## Motivation
The `lambda` test was failing when the [workflow](https://app.circleci.com/pipelines/github/localstack/localstack/23708/workflows/e17be93c-ffcb-446f-a8f0-2f355da30539/jobs/194487/tests) was run against `eu-central-1` with the following error: 

```
>> match key: long_function_arn_exc
	#x1B[33m(~)#x1B[0m /Error/Message "1 validation error detected: Value 'arn:aws:lambda:<region>:111111111111:function:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' at 'functionName' failed to satisfy constraint: Member must have length less than or equal to 140" → "1 validation error detected: Value 'arn:aws:lambda:<region>:111111111111:function:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' at 'functionName' failed to satisfy constraint: Member must have length less than or equal to 140" ... (expected → actual)

	Ignore list (please keep in mind list indices might not work and should be replaced):
	["$..Error.Message"]
```

The issue is caused due to variable lengths of suffix length: 

`suffix_length = max_function_arn_length - len(function_arn_prefix) + 1` can be seen [here](https://github.com/localstack/localstack/blob/18a8863d1ca7c6791dfc43246150d92b3b8bae46/tests/aws/services/lambda_/test_lambda_api.py#L466-L467) in the code. 

## Changes
This PR adds a transformer to the long function name generated through the suffix length from function arn and regenerates the snapshots.

PR #10550, opened 2024-03-27
fix `test_dashboard_lifecycle` for `eu-central-1`

sannya-singal:
## Motivation
The `cloudwatch` test was failing when the [workflow](https://app.circleci.com/pipelines/github/localstack/localstack/23708/workflows/e17be93c-ffcb-446f-a8f0-2f355da30539/jobs/194487/tests) was run against `eu-central-1` with the following error: 

```
>> match key: list_dashboards_prefix
	#x1B[33m(~)#x1B[0m /DashboardEntries/[0]/Size 225 → 228 ... (expected → actual)

	["$..DashboardEntries..Size"]
>> match key: list_dashboards
	#x1B[33m(~)#x1B[0m /DashboardEntries/[0]/Size 225 → 228 ... (expected → actual)

	["$..DashboardEntries..Size"]
```

The issue is caused due to variable lengths regions causing difference in `size` parameter of list dashboard entries. 

## Changes
This PR skips verifying the `Size` of `DashboardEntries`.

### DIFF
--- a/tests/aws/services/cloudwatch/test_cloudwatch.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch.py
@@ -1391,7 +1391,7 @@ class TestCloudwatch:
 
     @markers.aws.validated
     @markers.snapshot.skip_snapshot_verify(
-        paths=["$..DashboardArn"], condition=is_old_provider
+        paths=["$..DashboardArn", "$..DashboardEntries..Size"], condition=is_old_provider
     )  # ARN has a typo in moto
     def test_dashboard_lifecycle(self, aws_client, region_name, snapshot):
         dashboard_name = f"test-{short_uid()}"


steffyP (approved):
Thanks for tackling @sannya-singal! :)  

I think the fix should be fine for now, but this test will fail again once we change the default the default provider (as snapshot verify is currently only skipped conditional for `is_old_provider`). 

If we want to make it more stable, we should remove the condition here.

I wonder if there are other tests where the region would change size or hash values for certain attributes in the snapshot 🤔

sannya-singal:
Thanks for the review @steffyP 🙌 

> If we want to make it more stable, we should remove the condition here.
Agreed. 

> I wonder if there are other tests where the region would change size or hash values for certain attributes in the snapshot 
Yes, there are some others, eg: https://github.com/localstack/localstack/pull/10549

Issue #10327, opened 2024-02-26
bug: Previously deleted stacks throw ChangeSetNotFound when trying to redeploy the stack


Chad-Bailer-Regal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We are unable to redeploy previously deleted stacks with SAM. If we deploy a stack, delete it, and attempt to redeploy it it will throw a `ChangeSetNotFound` error. 

First Deploy:
```
$ samlocal deploy --stack-name  localstack-sam-destroy

        Deploying with following values
        ===============================
        Stack name                   : localstack-sam-destroy
        Region                       : us-east-1
        Confirm changeset            : True
        Disable rollback             : False
        Deployment s3 bucket         : aws-sam-cli-managed-test
        Capabilities                 : ["CAPABILITY_IAM"]
        Parameter overrides          : {}
        Signing Profiles             : {}

Initiating deployment
=====================

File with same data already exists at e812028d2e06af6ed56b4147f5fddb92.template, skipping upload


Waiting for changeset to be created..

CloudFormation stack changeset
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Operation                                                    LogicalResourceId                                            ResourceType                                                 Replacement
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ Add                                                        LogGroupDestroy                                              AWS::Logs::LogGroup                                          N/A
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Changeset created successfully. arn:aws:cloudformation:us-east-1:000000000000:changeSet/samcli-deploy1708981432/f08fd3b8


Previewing CloudFormation changeset before deployment
======================================================
Deploy this changeset? [y/N]: y

2024-02-26 16:03:53 - Waiting for stack create/update to complete

CloudFormation events from stack operations (refresh every 10.0 seconds)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ResourceStatus                                               ResourceType                                                 LogicalResourceId                                            ResourceStatusReason
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
CREATE_IN_PROGRESS                                           AWS::CloudFormation::Stack                                   localstack-sam-destroy                                     -
[...]

PR #10551, opened 2024-03-27
Fix handling reuse of stack names for describe stack events / delete stack operations

dfangl:
## Motivation
As reported in #10327 , we still have problems while redeploying AWS SAM stacks with the same stack name against LocalStack.
The root cause of this issue is `describe_stack_events` not only returning the events for a single stack, but for all stacks with that name. The expected behavior is for that operation to only work by name if the stack is active (i.e. not deleted), and by id for every stack, but only return the events for one in any case (since there can only be one active with a name, and stack ids are unique anyway).
Also, the `delete_stack` operation by name did not affect the active stack with the name, but the first stack with it (which might already be deleted), leading to deletes not working.

## Changes
* Add new `find_active_stack_by_name_or_id` function, for operations which work on active stacks primarily
* Add new `find_stack_by_id` function, to find any stack, by its id only
* Add test for "not found" or "null" errors for `describe_stack_events`
* Fix behavior for `describe_stack_events` and `delete_stack` when multiple stacks with the same name exist
* Extend `test_name_conflicts` test to test the new behavior

### DIFF
--- a/localstack/services/cloudformation/stores.py
+++ b/localstack/services/cloudformation/stores.py
@@ -61,6 +61,47 @@ def find_stack(account_id: str, region_name: str, stack_name: str) -> Stack | No
     )[0]
 
 
+def find_stack_by_id(account_id: str, region_name: str, stack_id: str) -> Stack | None:
+    """
+    Find the stack by id.
+
+    :param account_id: account of the stack
+    :param region_name: region of the stack
+    :param stack_id: stack id
+    :return: Stack if it is found, None otherwise
+    """
+    state = get_cloudformation_store(account_id, region_name)
+    for stack in state.stacks.values():
+        # there can only be one stack with an id
+        if stack_id == stack.stack_id:
+            return stack
+    return None
+
+
+def find_active_stack_by_name_or_id(
+    account_id: str, region_name: str, stack_name_or_id: str
+) -> Stack | None:
+    """
+    Find the active stack by name. Some cloudformation operations only allow referencing by slack name if the stack is
+    "active", which we currently interpret as not DELETE_COMPLETE.
+
+    :param account_id: account of the stack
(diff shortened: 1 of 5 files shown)


localstack/services/cloudformation/stores.py
```diff
@@ -61,6 +61,47 @@ def find_stack(account_id: str, region_name: str, stack_name: str) -> Stack | No
...
+            return stack
+    return None
+
+
+def find_active_stack_by_name_or_id(
+    account_id: str, region_name: str, stack_name_or_id: str
+) -> Stack | None:
+    """
+    Find the active stack by name. Some cloudformation operations only allow referencing by slack name if the stack is
+    "active", which we currently interpret as not DELETE_COMPLETE.
```
simonrw, line 86: > which we currently interpret as not DELETE_COMPLETE

Did you validate this with AWS? I _think_ this would be the correct approach, but I am not sure.

simonrw (approved):
Great set of changes, nice that we have some helper functions for picking the correct stack. As a follow up we should replace all usages of `find_stack` e.g. in `describe_stack_resource`:

https://github.com/localstack/localstack/pull/10551/files#diff-21252aafb0a3b77f8f2944c7405ef6321e37825908431516c86e19ffdddc51e7R904

to use these new helpers, and phase out the old `find_stack`. And of course do the same work with `find_changeset`.

localstack/services/cloudformation/stores.py
dfangl: The AWS documentation talks about "running stacks" and "deleted stacks": https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackEvents.html
I did not yet verify if, for example, `DELETE_IN_PROGRESS` counts as running, but we assume it is.

PR #10552, opened 2024-03-27
Feature: Eventbridge v2: Scaffold new provider

maxhoheiser:
## Motivation
This PR layers the groundwork for the implementation of the new EventBridge native provider. 

## Changes
This PR implements the basic scaffold of the new native EventBridge provider. 
It also updates the API stubs to add missing docstrings.
Furthermore, a flag to select the new provider is introduced, with `PROVIDER_OVERRIDE_EVENTS=v2`.
Test selection is introduced for all eventbridge tests that are not yet supported by the new provider, as the implementation improves these skips tests will be continuously removed, to gain full test coverage also for the new provider.

## Testing

To test the new provider manually, the env variable `PROVIDER_OVERRIDE_EVENTS=v2` needs to be set. 
All except the test `test_events.TestEventsEventBus.test_create_custom_event_bus` are marked as skipped, since the mocked new provider does not support them yet.

### DIFF
--- a/localstack/services/events/provider_v2.py
+++ b/localstack/services/events/provider_v2.py
@@ -1,0 +1,75 @@
+import logging
+
+from localstack.aws.api import RequestContext, handler
+from localstack.aws.api.events import (
+    CreateEventBusResponse,
+    EventBusName,
+    EventBusNameOrArn,
+    EventPattern,
+    EventsApi,
+    EventSourceName,
+    PutRuleResponse,
+    RoleArn,
+    RuleDescription,
+    RuleName,
+    RuleState,
+    ScheduleExpression,
+    TagList,
+)
+from localstack.services.plugins import ServiceLifecycleHook
+
+LOG = logging.getLogger(__name__)
+
+
+class EventsProvider(EventsApi, ServiceLifecycleHook):
+    def __init__(self):
+        self._rules = {}
+        self._event_buses = {}
+
+    @handler("CreateEventBus")
+    def create_event_bus(
+        self,
+        context: RequestContext,
+        name: EventBusName,
+        event_source_name: EventSourceName = None,
+        tags: TagList = None,
+        **kwargs,
+    ) -> CreateEventBusResponse:
+        event_bus_arn = f"arn:aws:events:{context.region}:{context.account_id}:event-bus/{name}"
+        event_bus = {"Name": name, "Arn": event_bus_arn}
+        self._event_buses[name] = event_bus
+
+        response = CreateEventBusResponse(
+            EventBusArn=event_bus_arn,
+        )
+        return response
+
+    @handler("PutRule")
+    def put_rule(
+        self,
+        context: RequestContext,
+        name: RuleName,
+        schedule_expression: ScheduleExpression = None,
+        event_pattern: EventPattern = None,
+        state: RuleState = None,
+        description: RuleDescription = None,
+        role_arn: RoleArn = None,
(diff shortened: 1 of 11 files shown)


localstack/services/events/provider_v2.py
```diff
@@ -0,0 +1,75 @@
+import logging
```
joe4dev, line 1: ℹ️  At least once we make the new v2 officially available, I think it makes sense to move all the v1-related functionality into a `legacy` (or v1) sub-directory (which might require ext changes!). This makes it easier to distinguish old from new.

We should synchronize on that one. It's probably better to defer that change until my event filtering refactorings and fixes are merged.

tests/aws/services/events/test_events.py
```diff
@@ -329,12 +338,18 @@ def received(q_urls):
             assert json.loads(msg_received["Body"]) == event
 
         # clean up
-        target_ids = [topic_target_id, sm_target_id, queue_target_id, fifo_queue_target_id]
+        target_ids = [
```
joe4dev, line 341: just wonder whether these are ruff-related changes or manually triggered 🤔

tests/aws/services/events/helper_functions.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+from localstack.testing.aws.util import is_aws_cloud
+
+
+def is_v2_provider():
```
joe4dev, line 6: This should go into some testing utility and not in the core codebase.

joe4dev (changes requested):
Looking forward to seeing EventBridge native v2 taking off 🚀 

Two things to change before merging:
1. The documentation should be removed from the API specs
2. The testing helper function should not be shipped with the core codebase

Given we don't have a CI pipeline, it would be helpful to add a "Testing" section to the PR description to describe:
* How did you manually test it?
* How can a reviewer test it? (often optional, but helps to get better feedback)

For example here, this could include mentioning a test case (or operation) that works when configuring the new provider.

localstack/aws/api/events/__init__.py
```diff
@@ -214,66 +214,99 @@ class RuleState(str):
 
 
 class ConcurrentModificationException(ServiceException):
+    """There is concurrent modification on a rule, target, archive, or replay."""
```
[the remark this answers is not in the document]
maxhoheiser: Removed it, I thought that this is the new API stubs should include docstrings since they do for a few services already.

tests/aws/services/events/test_events.py
maxhoheiser: Ruff is not playing nicely with VS Code fixed in.

tests/aws/services/events/helper_functions.py
joe4dev: Please ignore. Sorry, I missed the `tests` prefix. All good 👍 

We usually call it utils but helper_functions works as well 👍

localstack/aws/api/events/__init__.py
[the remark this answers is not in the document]
joe4dev: Interesting, for which services?

I checked a few new services such as DMS and didn't see docs. The `asf-updates.yml` workflow uses `python3 -m localstack.aws.scaffold upgrade` without argument, which should default to `--no-doc`. We should be compatible with it.

joe4dev (approved):
LGTM 👍 

PS: The failing CloudWatch test looks like a flake

PR #10540, opened 2024-03-26
remove `ap-southeast-1` from random regions

sannya-singal:
## Motivation
There are failing tests when primary and secondary regions are same. We have set the [TEST_AWS_SECONDARY_REGION_NAME](https://github.com/localstack/localstack/blob/b7f34c67c49c8f8d4aa6e0b08584b248501a64bb/localstack/constants.py#L158) to `ap-southeast-1`.

Example workflow: https://app.circleci.com/pipelines/github/localstack/localstack/23655/workflows/49291732-7851-49ea-917a-2a293dde3288/jobs/193896/tests

## Changes
This PR removes `ap-southeast-1` from the list of random regions.

PR #10553, opened 2024-03-27
Feat: Events v2: add ci step for eventbridge v2 provider

maxhoheiser:
## Motivation
https://github.com/localstack/localstack/pull/10552 introduces a new v2 provider for EventBridge, that can be selected via the env variable `PROVIDER_OVERRIDE_EVENTS=v2`. We need a new step in the CI run, that sets this env variable and runs tests against the new provider.

## Changes
Adds new step for CircleCi to test new EventBridge provider

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -730,6 +730,7 @@ workflows:
             - itest-sfn-legacy-provider
             - itest-s3-v2-legacy-provider
             - itest-cloudwatch-v2-provider
+            - itest-events-v2-provider
             - acceptance-tests
             - docker-test-amd64
             - docker-test-arm64
@@ -743,6 +744,7 @@ workflows:
             - itest-sfn-legacy-provider
             - itest-s3-v2-legacy-provider
             - itest-cloudwatch-v2-provider
+            - itest-events-v2-provider
             - acceptance-tests
             - docker-test-amd64
             - docker-test-arm64
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -269,6 +269,32 @@ jobs:
       - store_test_results:
           path: target/reports/
 
+  itest-events-v2-provider:
+    executor: ubuntu-machine-amd64
+    working_directory: /tmp/workspace/repo
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - prepare-pytest-tinybird
+      - prepare-account-region-randomization
+      - run:
+          name: Test EventBridge v2 provider
+          environment:
+            PROVIDER_OVERRIDE_EVENTS: "v2"
+            TEST_PATH: "tests/aws/services/events/"
+            COVERAGE_ARGS: "-p"
+          command: |
+            COVERAGE_FILE="target/coverage/.coverage.eventsV2.${CIRCLE_NODE_INDEX}" \
+            PYTEST_ARGS="${TINYBIRD_PYTEST_ARGS}${TESTSELECTION_PYTEST_ARGS}--reruns 3 --junitxml=target/reports/events_v2.xml -o junit_suite_name='events_v2'" \
+            make test-coverage
+      - persist_to_workspace:
+          root:
+            /tmp/workspace
+          paths:
+            - repo/target/coverage/
+      - store_test_results:
+          path: target/reports/
+
   docker-build:
     parameters:
       platform:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ commands:
                   export TEST_AWS_ACCOUNT_ID=$(LC_ALL=C tr -dc '0-9' < /dev/urandom | fold -w 12 | head -n 1)
                   export TEST_AWS_ACCESS_KEY_ID=$TEST_AWS_ACCOUNT_ID
                   # Set TEST_AWS_REGION_NAME to a random AWS region other than us-east-1
-                  export AWS_REGIONS=("us-east-2" "us-west-1" "us-west-2" "ap-southeast-2" "ap-northeast-1" "eu-central-1" "eu-west-1")
+                  export AWS_REGIONS=("us-east-2" "us-west-1" "us-west-2" "ap-southeast-1" "ap-southeast-2" "ap-northeast-1" "eu-central-1" "eu-west-1")
```
bentsku, line 66: just in case this slips back in, this is an older version of the current file, and this should not get merged
(@sannya-singal is working hard to greenify the pipeline and this would undo her change with #10540)

maxhoheiser: Thanks - rebased :)

joe4dev: @maxhoheiser Why do we need to change the regions in this PR?

This PR is still undoing the change from @sannya-singal here: https://github.com/localstack/localstack/pull/10540

.circleci/config.yml
```diff
@@ -269,6 +269,32 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - prepare-pytest-tinybird
+      - prepare-account-region-randomization
+      - run:
+          name: Test EventBridge v2 provider
+          environment:
+            PROVIDER_OVERRIDE_EVENTS: "v2"
+            TEST_PATH: "tests/aws/services/events/"
```
joe4dev, line 284: Do we have a single test that could pass, maybe from https://github.com/localstack/localstack/pull/10552?
I guess all would be skipped, for now, right?

.circleci/config.yml
```diff
@@ -269,6 +269,32 @@ jobs:
...
+            COVERAGE_ARGS: "-p"
+          command: |
+            COVERAGE_FILE="target/coverage/.coverage.eventsV2.${CIRCLE_NODE_INDEX}" \
```
joe4dev, line 287: Follows the same pattern as for CloudWatch v2 👍

joe4dev (changes requested):
The extra runner looks good following the same pattern as for CloudWatch v2 👍 

It seems the rebasing didn't work properly 😬 The region issue should get fixed.

This PR also depends on https://github.com/localstack/localstack/pull/10552, so I recommend rebasing after the merge and getting a green CI run before merging this.

.circleci/config.yml
maxhoheiser: I will rebase once https://github.com/localstack/localstack/pull/10552 is merged, to include the single passing test `test_events.TestEventsEventBus.test_create_custom_event_bus`

.circleci/config.yml
```diff
@@ -269,6 +269,32 @@ jobs:
       - store_test_results:
           path: target/reports/
 
+  itest-events-v2-provider:
```
joe4dev, line 272: The job still does not run in the pipeline. It misses some dependency declarations at the bottom of the job (see workflows > main jobs).
👉 We should manually check that 1) `itest-cloudwatch-v2-provider` actually runs and 2) the test `test_events.TestEventsEventBus.test_create_custom_event_bus` passes 👀

joe4dev (changes requested):
The dependency declarations are the bottom are missing ...

.circleci/config.yml
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10553/commits/c584c7e79f4e5f86ddd466533fc586e7581ddee4 and https://github.com/localstack/localstack/pull/10553/commits/3cc18d4259434f57771751f235ba412a22dcadca

joe4dev: great 👏 

Thanks for the clear traceability 💯

joe4dev (approved):
LGTM now 👍 Everything ready to 🚀

PR #10556, opened 2024-03-27
Add update template_body

lakkeger:
## Motivation
CFn get-template after change-set execution is not showing the right template.

## Changes
- Update template-body when change-set is applied
- Added aws validated test cases.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -182,7 +182,14 @@ def resolve_refs_recursively(
     value,
 ):
     result = _resolve_refs_recursively(
-        account_id, region_name, stack_name, resources, mappings, conditions, parameters, value
+        account_id,
+        region_name,
+        stack_name,
+        resources,
+        mappings,
+        conditions,
+        parameters,
+        value,
     )
 
     # localstack specific patches
@@ -314,7 +321,10 @@ def _resolve_refs_recursively(
             resource = resources.get(resource_logical_id)
 
             resolved_getatt = get_attr_from_model_instance(
-                resource, attribute_name, get_resource_type(resource), resource_logical_id
+                resource,
+                attribute_name,
+                get_resource_type(resource),
+                resource_logical_id,
             )
             # TODO: we should check the deployment state and not try to GetAtt from a resource that is still IN_PROGRESS or hasn't started yet.
             if resolved_getatt is None:
@@ -385,7 +395,8 @@ def _resolve_refs_recursively(
                     # We don't have access to the resource that's a dependency in this case,
                     # so do the best we can with the resource ids
                     raise DependencyNotYetSatisfied(
-                        resource_ids=key, message=f"Could not resolve {val} to terminal value type"
+                        resource_ids=key,
+                        message=f"Could not resolve {val} to terminal value type",
                     )
                 result = result.replace("${%s}" % key, resolved_val)
 
@@ -409,7 +420,12 @@ def _resolve_refs_recursively(
             if isinstance(mapping_id, dict) and "Ref" in mapping_id:
                 # TODO: ??
                 mapping_id = resolve_ref(
-                    account_id, region_name, stack_name, resources, parameters, mapping_id["Ref"]
+                    account_id,
+                    region_name,
+                    stack_name,
+                    resources,
+                    parameters,
+                    mapping_id["Ref"],
                 )
 
             selected_map = mappings.get(mapping_id)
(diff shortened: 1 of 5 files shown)


tests/aws/services/cloudformation/api/test_update_stack.py
```diff
@@ -407,3 +408,43 @@ def test_update_with_rollback_configuration(deploy_cfn_template, aws_client):
...
+    )
+    aws_client.cloudformation.get_waiter("stack_update_complete").wait(StackName=stack.stack_name)
+    get_template_response_1 = aws_client.cloudformation.get_template(StackName=stack.stack_name)
+    snapshot.match("get-template-response-1", get_template_response_1)
+
+    with pytest.raises(botocore.exceptions.ClientError):
+        aws_client.cloudformation.update_stack(
+            StackName=stack.stack_name,
+            TemplateBody=template_2,
+        )
```
simonrw, line 448: It would be good to capture the error response, e.g.

```suggestion
    with pytest.raises(botocore.exceptions.ClientError) as exc_info:
        aws_client.cloudformation.update_stack(
            StackName=stack.stack_name,
            TemplateBody=template_2,
        )
    snapshot.match("update-2-error", exc_info.value.response)
```

tests/aws/services/cloudformation/api/test_update_stack.py
simonrw, line 1: I found this test a little confusing. I found it weird that the template after the _first_ deploy was not snapshotted. I also think the second snapshot is redundant since the `update_stack` call is guaranteed to fail (by the nature of the `pytest.raises`).

Perhaps snapshot the template after stack creation, and remove the final snapshot?

tests/aws/services/cloudformation/api/test_update_stack.py
```diff
@@ -407,3 +408,43 @@ def test_update_with_rollback_configuration(deploy_cfn_template, aws_client):
...
+                Value: test
+                Type: String
+    """)
+    template_2 = textwrap.dedent("""
+    Resources:
+        SimpleParam1:
+            Type: AWS::SSM::Parameter
+            Properties:
+                Value: test1
+                Type: String
```
simonrw, line 429: I found the change from `test` to `test1` a little easy to miss on first read. Perhaps make the two `Value` values more significantly different, e.g. `before-stack-update` and `after-stack-update`?

localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -182,7 +182,14 @@ def resolve_refs_recursively(
...
     result = _resolve_refs_recursively(
-        account_id, region_name, stack_name, resources, mappings, conditions, parameters, value
+        account_id,
+        region_name,
+        stack_name,
+        resources,
+        mappings,
+        conditions,
+        parameters,
+        value,
```
simonrw, line 192: nit: these formatting changes seem to dominate the PR diff. Is that because of our switching to `ruff`? If not can we remove these formatting changes to clean up the PR?

simonrw (changes requested):
I would like a few updates to the tests, but in general this looks great, thanks for fixing this bug!

tests/aws/services/cloudformation/api/test_update_stack.py
macnev2013: Yes. Agreed. I've removed final snapshot. And added deployment status checks as final steps.

tests/aws/services/cloudformation/api/test_update_stack.py
macnev2013: Error from localstack was not in parity with AWS. I've updated the parity issue.

tests/aws/services/cloudformation/api/test_update_stack.py
```diff
@@ -407,3 +408,47 @@ def test_update_with_rollback_configuration(deploy_cfn_template, aws_client):
...
+
+    with pytest.raises(botocore.exceptions.ClientError) as exc_info:
+        aws_client.cloudformation.update_stack(
+            StackName=stack.stack_name,
+            TemplateBody=template_2,
+        )
+    snapshot.match("update-error", exc_info.value.response)
+
+    describe_stack_response = aws_client.cloudformation.describe_stacks(StackName=stack.stack_name)
+    assert describe_stack_response["Stacks"][0]["StackStatus"] == "UPDATE_COMPLETE"
```
simonrw, line 454: nit: this is a little confusing. The stack status is set to `UPDATE_COMPLETE` after the first stack update, not after the failed attempt to update the stack. Perhaps move this block to before the `pytest.raises` block?

simonrw (approved):
Looks good, just a minor suggestion about code understandability. Glad we have better parity on stack updates!

tests/aws/services/cloudformation/api/test_update_stack.py
macnev2013: In LocalStack, we [manually set the stack status](https://github.com/localstack/localstack/pull/10556/files#diff-21252aafb0a3b77f8f2944c7405ef6321e37825908431516c86e19ffdddc51e7R379-R381) when no stack updates are necessary. So I wanted to verify that the status is indeed updated.

simonrw: Sure, but the stack was _already_ in `UPDATE_COMPLETE`

Issue #8026, opened 2023-03-31
bug: lastRotatedDate & nextRotationDate not set by secret rotation lambdas


davidsloan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using 
* secret manager
* lambda function to rotate the secret

after the secret is rotated successfully and I describe the secret, then two fields in the DescribeSecretResponse:
* lastRotatedDate
* nextRotationDate
are both null.

### Expected Behavior

`lastRotatedDate` and `nextRotationDate` should be set according to the secret's rotation rules

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Using `testcontainers`, which uses Docker.
A test project has been pushed to 
https://bitbucket.org/lenses-test/localstack-secret-rotation/
to demonstrate the problem.

Just run `sbt test`

There’s a test here (src/test/scala/io/lenses/connect/secrets/integration/AwsSecretProviderIT.scala) that reproduces the issue as described above.  I’ve tried to reduce the noise to the minimum.  

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
Please see above project for complete example runnable in scala. Here are the main parts:
```
// create role
        val newRole = iam.createRole(CreateRoleRequest.builder().roleName("iamLambdaRole").build())
// create secret
        
        val secret = CreateSecretRequest
          .builder()
          .name(awsSecretPath)
          .secretString(secretJson)
          .build()

        val sec = client
          .createSecret(secret)
        sec.sdkHttpResponse().statusCode().toString should startWith("2")

// create lambda
        val waiter       = lamb.waiter()
        val filePath     = getClass.getResourceAsStream("/functionpy/main.zip")
        val fileToUpload = SdkBytes.fromInputStream(filePath)

        val code = FunctionCode.builder()
          .zipFile(fileToUpload)
          .build()

        val createFnReq = CreateFunctionRequest
          .builder()
          .code(code)
          .functionName(functionName)
          .role(roleArn)
          .handler(" lambda_function.lambda_handler")
          .runtime(Runtime.PYTHON3_9)
          .environment(Environment.builder().variables(
            Map[String, String]("SECRETS_MANAGER_ENDPOINT" -> networkVaultAddress).asJava,
          ).build())
          .build()

        val lambdaFn = lamb.createFunction(createFnReq)

[...]

PR #10564, opened 2024-03-28
fix lastRotatedDate updation on secret rotation

macnev2013:
## Motivation
This pull request addresses the issue highlighted in https://github.com/localstack/localstack/issues/8026, where `lastRotatedDate` was not being updated upon secret rotation. This fix ensures that `lastRotatedDate` is correctly updated when secrets are rotated.

## Changes
- `lastRotatedDate` is updated during secret rotation.
- Added `Versions`, `VersionId` to snapshot matching: It was causing test flakiness.

## Testing
- Added AWS validated test case.


viren-nadkarni (approved):
A clean fix, LGTM :+1:

PR #10566, opened 2024-03-28
Fix non-default-region compatibility for recent lambda invalid invoke test

dfangl:
## Motivation
With #10323 we introduced a validation of lambda arns when invoking. However, the test hardcoded us-east-1 in the arn, which was by default replaced by the region snapshot transformer.
However, with different regions, it does not get replaced, making the test incompatible with multiple regions (the functionality itself was correct, though).

Also, there seems to have been a recent update to the error message and validation regex for iso regions, which I adapted in LocalStack as well.

## Changes
* Use current region for invalid arn to allow region snapshot transformer to work properly @sannya-singal 
* Adjust regex and error message to recent AWS changes

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -818,7 +818,8 @@ class TestLambdaFunction:
     def test_invalid_invoke(self, aws_client, snapshot):
         with pytest.raises(aws_client.lambda_.exceptions.ClientError) as e:
             aws_client.lambda_.invoke(
-                FunctionName="arn:aws:lambda:us-east-1:123400000000@function:myfn", Payload=b"{}"
+                FunctionName=f"arn:aws:lambda:{aws_client.lambda_.meta.region_name}:123400000000@function:myfn",
+                Payload=b"{}",
             )
         snapshot.match("invoke_function_name_pattern_exc", e.value.response)
 
--- a/localstack/services/lambda_/api_utils.py
+++ b/localstack/services/lambda_/api_utils.py
@@ -59,7 +59,7 @@ DESTINATION_ARN_PATTERN = re.compile(
 )
 
 AWS_FUNCTION_NAME_REGEX = re.compile(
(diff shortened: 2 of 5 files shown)


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -818,7 +818,8 @@ def test_vpc_config(
     def test_invalid_invoke(self, aws_client, snapshot):
         with pytest.raises(aws_client.lambda_.exceptions.ClientError) as e:
             aws_client.lambda_.invoke(
-                FunctionName="arn:aws:lambda:us-east-1:123400000000@function:myfn", Payload=b"{}"
+                FunctionName=f"arn:aws:lambda:{aws_client.lambda_.meta.region_name}:123400000000@function:myfn",
```
sannya-singal, line 821: nit: we can use the `region_name` fixture here.

sannya-singal: `FunctionName=f"arn:aws:lambda:{region_name}:{account_id}@function:myfn",`

dfangl: The account id should not be an actually used one, it's just some random number, so I do not need the fixture here. For the region, we use this pattern all over lambda tests, so I did not want to change that yet :)

sannya-singal: Aah got it, although we can case once in a trillion where the `account_id` could be `123400000000` but that should be fine.

sannya-singal: That should be fine but we lately have a bunch of places in lambda where we use the fixtures: https://github.com/localstack/localstack/blob/3b8099b864d7cdf684fc630a2c6d09f8d3abab35/tests/aws/services/lambda_/test_lambda_api.py#L424

joe4dev (approved):
nice catch ✨ 

These regexes are living entities 😅

sannya-singal (approved):

tests/aws/services/lambda_/test_lambda_api.py
dfangl: Ah yes, you are absolutely right! We should definitely bring all that into line at some point. If you do not have a strong opinion on this (which I would be fine with as well) I would leave it as is for now, and migrate all of the remaining tests to the new fixture for easier readability at a later point :)

sannya-singal: Yes that should be fine for now! 🙌

PR #10568, opened 2024-03-28
raise exception if required ENV is missing

steffyP:
## Motivation
The helper script to upload tinybird coverage metrics was silently failing on ext, as we only had a return statement if any ENV was missing.

## Changes
* raise Exception if required ENV is missing to fail the workflow


silv-io (approved):
LGTM! Thanks for fixing this :)

PR #10571, opened 2024-03-29
remove the conditional provider in `test_dashboard_lifecycle`

sannya-singal:
## Motivation
As per previous discussion in #10550 PR [here](https://github.com/localstack/localstack/pull/10550#pullrequestreview-1963137286), we should remove the conditional provider in `test_dashboard_lifecycle`  to make the test more robust for old and `v2` both cloudwatch providers.  

## Changes
This PR removes the conditional old provider in skip snapshot verifification. 

## Testing
The following test has been verified against new(`PROVIDER_OVERRIDE_CLOUDWATCH=v2`) and old providers.

### DIFF
--- a/tests/aws/services/cloudwatch/test_cloudwatch.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch.py
@@ -1393,7 +1393,7 @@ class TestCloudwatch:
     @markers.snapshot.skip_snapshot_verify(
         condition=is_old_provider,
         paths=[
-            "$..DashboardArn",
+            "$..DashboardArn",  # ARN has a typo in moto
             "$..DashboardEntries..Size",
         ],
     )
--- a/tests/aws/services/cloudwatch/test_cloudwatch.py
+++ b/tests/aws/services/cloudwatch/test_cloudwatch.py
@@ -1391,7 +1391,7 @@ class TestCloudwatch:
 
     @markers.aws.validated
     @markers.snapshot.skip_snapshot_verify(
-        paths=["$..DashboardArn", "$..DashboardEntries..Size"], condition=is_old_provider
+        paths=["$..DashboardArn", "$..DashboardEntries..Size"]
     )  # ARN has a typo in moto
     def test_dashboard_lifecycle(self, aws_client, region_name, snapshot):
         dashboard_name = f"test-{short_uid()}"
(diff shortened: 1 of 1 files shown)


tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1391,7 +1391,7 @@ def contains_receive_delete_metrics() -> int:
 
     @markers.aws.validated
     @markers.snapshot.skip_snapshot_verify(
```
steffyP, line 1393: Actually i think it would be better if we use a different approach here. Because the `DashboardArn` should not be skipped for the new provider.

Also adding a comment would be nice, so that we know why this change was necessary in the first place.
When we re-run snapshots, we sometimes test if the skips are still relevant, and remove them if the test works fine. The `DasboardEntries..Size` is one parameter where I think this could happen, so having a comment to make it clear why it is there would be great 🙏 

I would suggest 
```
        condition=is_old_provider,
        paths=[
            "$..DashboardArn",
        ],
        )
        paths=[
            "$..DashboardEntries..Size", # need to be skipped because size changes if the region name length is longer
        ]
    )
```

steffyP (changes requested):
Thanks for tackling, Sannya 🙏 
I added a comment, after giving it a thought, I proposed a different solution.

tests/aws/services/cloudwatch/test_cloudwatch.py
sannya-singal: Agreed. Thanks for the review @steffyP!

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1391,8 +1391,17 @@ def contains_receive_delete_metrics() -> int:
...
-        paths=["$..DashboardArn", "$..DashboardEntries..Size"], condition=is_old_provider
-    )  # ARN has a typo in moto
+        condition=is_old_provider,
+        paths=[
+            "$..DashboardArn",  # ARN has a typo in moto
+            "$..DashboardEntries..Size",
```
steffyP, line 1397: NIT: we shouldn't need the `"$..DashboardEntries..Size"` here, as the other skip is unconditional and therefore includes the skip already. But nothing urgent.

steffyP (approved):
Thanks for tackling this so fast 🚀 
added a NIT but it's fine for now.

tests/aws/services/cloudwatch/test_cloudwatch.py
sannya-singal: nice catch 🙌 have fixed this

PR #10575, opened 2024-03-29
fix SNS RawMessageDelivery casing

bentsku:
## Motivation
As shown in #10545, we had an issue with the SNS provider not lower casing incoming value, even though it works with AWS.

## Changes
- fix the casing of `RawMessageDelivery` for `SetSubscriptionAttributes` and `Subscribe`
- add positive and negative tests for both `SetSubscriptionAttributes` and `Subscribe`
- remove a doc string for `call_moto_with_request` that I had added a while back and forgot to remove when the PR was updated

\cc @pinzon

### DIFF
--- a/localstack/services/moto.py
+++ b/localstack/services/moto.py
@@ -59,7 +59,6 @@ def call_moto_with_request(
 
     :param context: the original request context
     :param service_request: the dictionary containing the service request parameters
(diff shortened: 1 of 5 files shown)


pinzon (approved):
Awesome!

Issue #10461, opened 2024-03-14
bug: API Gateway Proxy does not forward the backend response code (always returns 200)


dmitriyklyuzov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The LocalStack-based API Gateway does not forward the HTTP code from the backend and instead always returns a 200 response, regardless of what the backend response code was:
![Screenshot 2024-03-14 at 10 20 17 AM](https://github.com/localstack/localstack/assets/5950326/ae2f4df9-9537-42b4-9416-74394e2fcd1d)

### Expected Behavior

According to the [AWS Documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-response.html), _For proxy integrations, API Gateway automatically passes the backend response to the method response._ In other words, if the backend is returning an HTTP 401, the API Gateway should also return a 401 (not 200), as confirmed when deploying the same infrastructure to AWS:
![Screenshot 2024-03-14 at 10 20 21 AM](https://github.com/localstack/localstack/assets/5950326/bf6ae810-27d4-4a9c-840c-43a5e36666f5)

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Terraform code
```
resource "aws_api_gateway_rest_api" "root" {
  name        = "test-api-gw"
  description = "Test API Gateway"
}

resource "aws_api_gateway_resource" "icat_resource" {
  rest_api_id = aws_api_gateway_rest_api.root.id
  parent_id   = aws_api_gateway_rest_api.root.root_resource_id
  path_part   = "icat"
}

resource "aws_api_gateway_resource" "icat_proxy" {
  rest_api_id = aws_api_gateway_rest_api.root.id
  parent_id   = aws_api_gateway_resource.icat_resource.id
  path_part   = "{proxy+}"
}

resource "aws_api_gateway_method" "icat_api_method" {
  rest_api_id   = aws_api_gateway_rest_api.root.id
  resource_id   = aws_api_gateway_resource.icat_proxy.id
  http_method   = "ANY"
  authorization = "NONE"

  request_parameters = {
    "method.request.path.proxy" = true
  }
}

resource "aws_api_gateway_integration" "icat_integration" {
  rest_api_id             = aws_api_gateway_rest_api.root.id
[...]

PR #10583, opened 2024-04-01
Fix apigw http proxy response passthrough

cloutierMat:
## Motivation

As mentioned in issue #10461, we were matching the response status code to the integration method for `HTTP_PROXY` integrations. As documented [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-response.html), `With a proxy integration, API Gateway passes the backend response through to the method response automatically.`

## Changes

Skipping entirely the response response template rendering for `HTTP_PROXY` integrations. As neither status code nor body transformation should be processed for this type of integration.

### DIFF
--- a/localstack/services/apigateway/integration.py
+++ b/localstack/services/apigateway/integration.py
@@ -802,9 +802,10 @@ class HTTPIntegration(BackendIntegration):
                 uri,
                 result.status_code,
             )
-        # apply custom response template
+        # apply custom response template for non-proxy integration
         invocation_context.response = result
-        self.response_templates.render(invocation_context)
+        if integration["type"] != "HTTP_PROXY":
+            self.response_templates.render(invocation_context)
         return invocation_context.response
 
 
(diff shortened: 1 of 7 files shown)


localstack/services/apigateway/integration.py
```diff
@@ -802,9 +802,10 @@ def invoke(self, invocation_context: ApiInvocationContext):
                 uri,
                 result.status_code,
             )
-        # apply custom response template
+        # apply custom response template for non-proxy integration
         invocation_context.response = result
-        self.response_templates.render(invocation_context)
+        if integration["type"] != "HTTP_PROXY":
+            self.response_templates.render(invocation_context)
```
bentsku, line 808: nice catch!! love that the fix is always some little line changes but the implication and set up around it is big 😄

bentsku (approved):
LGTM! Very nice, clean fix, very well tested, great investigation 🕵️‍♂️

dmitriyklyuzov:
Thanks for fixing this! Any ETA on when this will be released?

PR #10563, opened 2024-03-28
fix clouformation ec2 tests for `ap-northeast-1` and validate against AWS

sannya-singal:
## Motivation
Test `test_vpc_creates_default_sg` and `test_transit_gateway_attachment`fail when run in the [workflow](https://app.circleci.com/pipelines/github/localstack/localstack/23755/workflows/5d782a03-6bc1-4bd2-bcad-c6f1170e5381/jobs/194931/tests) for region: `ap-northeast-1` and throws a `StackDeployError`. 

## Changes
This PR validates test: `test_vpc_creates_default_sg` against AWS and deploys both the tests specifically in `us-east-1`.

PR #10586, opened 2024-04-02
fix cfn templates deployed in invalid AZs

sannya-singal:
## Motivation
While working on greenifying cross accounts pipeline #10563, we discovered the issue with creating cfn templates in invalid availability zones.  

## Changes
This PR: 
- updates the way we fetch the AZs: rather than traversing over `"a", "b", "c", "d", "e", "f"`, we call `connect_to().ec2.describe_availability_zones(...)`
- reverts the changes in #10563 for deploying the templates specifically in `us-east-1`.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -610,9 +610,11 @@ def _resolve_refs_recursively(
                 )
                 or region_name
             )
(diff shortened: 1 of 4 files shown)


Morijarti (approved):
LGTM :)

Thank you for making the changes and making CFn's handling of AZs more robust

PR #10587, opened 2024-04-02
update pre-commit hook versions on dep updates

alexrashed:
## Motivation
We now automatically update our pinned dependencies in the `requirements*.txt` files every week using a scheduled GitHub action. However, this currently does not update the - separately managed - `pre-commit` hook versions. This could lead to problems because we enforce the linting rules based on the latest `ruff` version, while the `pre-commit` hook is outdated.
This PR fixes this issue by updating the pre-commit hooks together with the pinned dependencies.

## Changes
- Automatically executes `pre-commit autoupdate` when updating the dependencies such that they are updated together with the dev tools in the venv.

## Testing
- Tested locally, the resulting update of `.pre-commit-config.yaml` is the result of `make upgrade-pinned-dependencies`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.3.3
+    rev: v0.3.5
     hooks:
       - id: ruff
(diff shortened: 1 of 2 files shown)


silv-io (approved):
Awesome! Didn't expect it to be such a small fix :)

Issue #10588, opened 2024-04-02
bug: Dynamodb 500 error after upgrading to 3.3


magnusjtvisma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Dynamodb started crashing after upgrading to 3.3:

File "/opt/code/localstack/localstack/services/dynamodb/provider.py", line 1263, in transact_write_items
     records_map = self.prepare_transact_write_item_records(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/opt/code/localstack/localstack/services/dynamodb/provider.py", line 1775, in prepare_transact_write_item_records
     stream_type = tables_stream_type[table_name]
                   ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
 KeyError: MyTable

This is on a table without streams, but I do have other tables with streams enabled.

### Expected Behavior

_No response_

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up


### Environment

```markdown
- OS: Windows
- LocalStack: 3.3
```


### Anything else?

_No response_

bentsku:
Thank you for the report again, sorry it slipped through during our performance improvements work 😅 the fix should be available in the next `latest` Docker image. Thanks again!

PR #10593, opened 2024-04-02
fix DDB issue when TransactWriteItems targets tables with and without streams

bentsku:
## Motivation
As reported by #10588, we had an issue when calling `TransactWriteItems` with requests targeting different tables, when some of them had stream enabled and some didn't. 

## Changes
- add a test checking the scenario
- add a check to skip the requests where the table has no stream enabled when iterating over the transact items

_fixes #10588_


viren-nadkarni (approved):
Thanks for fixing this!

Issue #10567, opened 2024-03-28
bug: Array properties aren't handled correctly when Filter Policy Scope is set to MessageBody for SNS subscription


ABAG603:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Messages are not not being routed to the queues based on the Filter Policy (see steps to reproduce for more details)

### Expected Behavior

Messages are being correctly routed. This is the behaviour when tested against "real" SQS/SNS.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

Setup
- SNS topic: topic1
- SQS queues: queue1, queue2

Both queues are being subscribed to the topic1 with **Filter Policy Scope** set to **MessageBody**

Filter policy for **queue1** topic subscription:
```json
{
    "headers": {
      "route-to": ["queue1"]  
    }
}
```
Filter policy for **queue2** topic subscription:
```json
{
    "headers": {
      "route-to": ["queue2"]  
    }
}
``` 

**Scenario 1**
Send message below to the **topic1**
```json
{
    "headers": {
      "route-to": ["queue1"]  
    }
}
```
Expected: message to appear in the **queue1** (this is exactly what is happening in AWS SNS\SQS)
Actual: no messages in the **queue1**

**Scenario 2**
Send message below to the **topic1**
```json
{
    "headers": {
      "route-to": ["queue2"]  
    }
}
```
Expected: message to appear in the **queue2** (this is exactly what is happening in AWS SNS\SQS)
Actual: no messages in the **queue2**

**Scenario 3**
Send message below to the **topic1**
```json
{
    "headers": {
      "route-to": ["queue1", "queue2"]  
    }
}
```
Expected: message to appear in both **queue1** and **queue2** (this is exactly what is happening in AWS SNS\SQS)
Actual: message is neither in **queue1** nor **queue2**

### Environment

```markdown
- OS: Mac OS Sonoma 14.4.1
- LocalStack: 3.2.0
```

ABAG603:
Hi @bentsku . It looks like this fix introduced a regression. Now filter policy works only when header is an array. So,

this works - message will be delivered to queue1
```
{
    "headers": {
      "route-to": ["queue1"]  
    }
}
```

this doesn't
```
{
    "headers": {
      "route-to": "queue1"
    }
}
```

I'm testing with LocalStack 3.4.0. I've also tested in AWS and it works in both cases above.

bentsku:
[...]

PR #10594, opened 2024-04-02
fix SNS MessageBody filtering when value is a list

bentsku:
## Motivation
As reported by #10567, JSON array properties were not handled well when the scope of the filter was `MessageBody`. We would handle it correctly for `MessagesAttributes` but it was only tested in unit tests. 

## Changes
- add 2 AWS validated tests: one for `MessageAttributes` scope with `String.Array` filter and one for array property in a JSON message body
- add the fix to recursively match the values of an array if the value is an array in an `OR` fashion (at least one value must match)

_fixes #10567_


baermat (approved):
LGTM :+1:

PR #10597, opened 2024-04-03
Fix release helper for cli release

silv-io:
## Motivation

Switching the release helper to work with `pyproject.toml` broke the CLI release since it did not work with `requirements.txt` files anymore. Also, changing the version single sourcing broke the release commit function, since it does not have a version.py file.

## Changes

1. The regular expressions in `grep` and `sed` were changed, so that they can capture both `requirements.txt` as well as `pyproject.toml` files.
2. The `git add` for the release and increment commits was changed, so that only files are added that truly exist. 
3. Reintroduce the usage of `egrep` which was previously removed because it is deprecated in modern versions of `grep`. However, on macOS, `egrep` is still the way to go.

## Testing

In `localstack/localstack-cli` you can execute the following:

```
DEPENDENCY_FILE="requirements.txt" ../localstack/bin/release-helper.sh set-dep-ver "localstack" "==3.4.0"
DEPENDENCY_FILE="requirements.txt" VERSION_FILE="" ../localstack/bin/release-helper.sh git-commit-release 3.4.0
```

And check if there is a new commit which changes the version of localstack correctly.

You can also check setting an arbitrary dependency in `localstack/localstack` and see if it still works.

Additionally you can see in this screenshot that all relevant options are matched:

Screenshot 2024-04-03 at 11 15 15

## TODO

- [x] Check whether we need to make the release script default Mac compatible. I had to install the GNU versions `grep` and `sed` via `brew` and adapt the `$PATH` for it to work. Can we expect LocalStack team members with Macs to do the same?

### DIFF
--- a/bin/release-helper.sh
+++ b/bin/release-helper.sh
@@ -159,7 +159,7 @@ function cmd-set-dep-ver() {
     dep=$1
     ver=$2
 
-    grep -Eh "^(\s*\"?)(${dep})(\[[a-zA-Z0-9,]+\])?(>=|==|<=)([^\"]*)(\")?(,)?$" ${DEPENDENCY_FILE} || { echo "dependency ${dep} not found in ${DEPENDENCY_FILE}"; return 1; }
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks for jumping on this issue that appeared during the last release! 🐞 💯 
I quickly tested the change and it seems that it now works just fine for the `requirements.txt` in `localstack/localstack-cli`.

PR #10599, opened 2024-04-03
Add event matching test suite

joe4dev:
## Motivation

Event filtering is insufficiently tested and implemented in too many places in LocalStack 😢 
Before we can unify all existing implementations, we need a robust AWS-validated test suite.

Existing implementations in LocalStack:
1. [localstack] [test_event_pattern](https://github.com/localstack/localstack/blob/acd38e462a6bb398d8585b0e18338e5c42ae2e59/localstack/services/events/provider.py#L118) in the EventBridge provider implements a simplistic check whether a key is present in ~1 LOC.
2. [localstack] [filter_event](https://github.com/localstack/localstack/blob/acd38e462a6bb398d8585b0e18338e5c42ae2e59/localstack/services/events/provider.py#L534) (inner method!) in the EventBridge provider implements the [Amazon EventBridge event patterns](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html) in ~200 LOC. The implementation is confusing (e.g., operators are implemented multiple times), hard to maintain, and has several refactoring TODOs.
3. [moto-ext] [_does_event_match](https://github.com/localstack/moto/blob/e7cf92bf3d09bab1117ac55579bc74eee3407245/moto/events/models.py#L885) within the `EventPattern` models of moto-ext implements the same [Amazon EventBridge event patterns](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html) in ~50 LOC. The implementation is short and concise (e.g., mapping operators to functions `<` => `lt`) but incomplete. LocalStack patches this implementation in the EventBridge provider. Hence, it is not used within LocalStack.
4. [localstack-ext] [filter_event](https://github.com/localstack/localstack-ext/blob/6cabb831044c87481f61818ef95e6590a33af683/localstack_ext/services/pipes/pipe_utils.py#L93) in the Pipes utils is a copy/pasted implementation from EventBridge. It was temporarily copied because the EventBridge implementation was not re-usable but will be unified within this PR. It implements [Amazon EventBridge Pipes filtering](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-filtering.html), which appears to be the same implementation as for EventBridge.
5. [localstack] [filter_stream_record](https://github.com/localstack/localstack/blob/acd38e462a6bb398d8585b0e18338e5c42ae2e59/localstack/services/lambda_/event_source_listeners/utils.py#L23) in the Lambda Event Source Mapping (ESM) utils is ~110 LOC. It implements the [Lambda event filtering](https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html), which appears to be the same implementation as for Pipes and EventBridge. We know from other sources that ESM and Pipes use the same infrastructure at AWS. 
6. [localstack] [_evaluate_nested_filter_policy_on_dict](https://github.com/localstack/localstack/blob/acd38e462a6bb398d8585b0e18338e5c42ae2e59/localstack/services/sns/publisher.py#L1129) within the "SubscriptionFilter" of SNS is ~130 LOC. It implements [Amazon SNS subscription filter policies](https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html), which appear very similar to the EventBridge implementation. However, this implementation is not identical (e.g., does not support the wildcard operator).
7. [moto-ext] [_compute_body_checks](https://github.com/localstack/moto/blob/60fb7edc3f39ec3f21ddc08d61bd9d8394b4b846/moto/sns/utils.py#L94) within the SNS utils in moto-ext implements [Amazon SNS subscription filter policies](https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html) is ~340 LOC. In addition to the matching functionality, moto-ext also implements the filter validation using a complexity calculation in [_validate_filter_policy](https://github.com/localstack/moto/blob/60fb7edc3f39ec3f21ddc08d61bd9d8394b4b846/moto/sns/models.py#L782). These extra ~170 LOC also cover several exception cases. According to @bentsku, our moto-ext implementation also got backported to upstream moto.

AWS [recently](https://aws.amazon.com/about-aws/whats-new/2024/02/amazon-eventbridge-event-bus-matching-event-ruler/) (2024-02-01) open-sourced https://github.com/aws/event-ruler. This rule engine implements the above filtering patterns. It is written in Java in ~6k LOC. Their [blog post](https://aws.amazon.com/blogs/opensource/open-sourcing-event-ruler/) mentions that the internals are built on top of finite state machines, which can be updated dynamically with new rules.

## Changes

* Make `test_event_pattern` in the event provider use our EventBridge filtering implementation
* Add a suite of ~70 aws-validated tests based on the documentation of the filtering patterns. The tests mainly based on the original [Amazon EventBridge event patterns](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html) but also inspired from the open source event ruler and discovered edge cases. The test suite has a good initial operator coverage but misses further edge cases and exception cases.
* Extract the implementation from the EventBridge provider into a separate utils file
* Refactor (e.g., `filter_event` => `matches_event`) and fix several issues in the EventBridge `filter_events` method and its helpers (e.g., improved handling of numeric operators and improved list intersections handling)
* Refactor (e.g., `filter_stream_record` => `does_match_event`) and fix several issues (e.g., operator casing, numeric conditions, raising exceptions)

## Testing

Within [localstack/services/events/provider.py](https://github.com/localstack/localstack/compare/add-event-filtering-test-suite?expand=1#diff-e93fb0ca12285e3a1b71692d6dcae8cf7333857a6019bbe5af6115615541d088), `test_event_pattern` can be used to invoke different implementation within LocalStack.

## Discussion

Raising exceptions instead of silently ignoring the error case and returning True/False is a behavior change that can help us to make issues more visible. Background processes should handle such exceptions better (e.g., The SQS ESM implementation just drops the entire batch upon unhandled exceptions). At AWS, more exceptions are probably caught earlier because the rule engine can validate rules before actually doing a matching.


tests/aws/services/events/test_event_patterns.py
```diff
@@ -0,0 +1,160 @@
...
+    "or-exists-parent",
+]
+
+
+# TODO: extend these test cases based on the open source docs + tests: https://github.com/aws/event-ruler
+#  For example, "JSON Array Matching", "And and Or Relationship among fields with Ruler", rule validation,
+#  and exception handling.
+@pytest.mark.skipif(is_v2_provider(), reason="V2 provider does not support this feature yet")
+@pytest.mark.parametrize(
+    "request_template,label", request_template_tuples, ids=[t[1] for t in request_template_tuples]
```
joe4dev, line 84: probably debatable whether it's a good idea to parameterize tests based on directory contents. It makes it very easy to add new test cases but add a small delay to test collection because it needs to list the files in the directory 🤷

tests/aws/services/events/test_events_rules.py
```diff
@@ -261,7 +263,8 @@ def test_put_event_with_content_base_rule_in_pattern(aws_client, clean_up):
             "test2": [{"anything-but": "test2"}],
             "test3": [{"anything-but": ["test3", "test33"]}],
             "test4": [{"anything-but": {"prefix": "test4"}}],
-            "ip": [{"cidr": "10.102.1.0/24"}],
```
joe4dev, line 264: That IP check never worked but it got silently ignored 😬  (just double-checked by running the test suite against the old implementations before the refactorings and fixes)

tests/aws/services/events/test_event_patterns.py
dfangl: I think this is fine for tests like this :)

tests/aws/services/events/test_event_patterns.py
```diff
@@ -0,0 +1,160 @@
...
+    b) NO MATCH (_NEG suffix): The EventPattern does NOT match the Event yielding false as result.
+    c) EXCEPTION (_EXC suffix): The EventPattern is invalid and raises an exception.
+    """
+    if label in SKIP_LABELS and not is_aws_cloud():
+        pytest.skip("Not yet implemented")
+
+    event = request_template["Event"]
+    event_pattern = request_template["EventPattern"]
+
+    if label.endswith("_EXC"):
```
dfangl, line 99: What was your reasoning here for encoding this in the file name, instead of having another field in the json5 file?

tests/aws/services/events/test_events_rules.py
```diff
@@ -247,11 +247,13 @@ def test_put_event_with_content_base_rule_in_pattern(aws_client, clean_up):
     queue_url = aws_client.sqs.create_queue(QueueName=queue_name)["QueueUrl"]
     queue_arn = arns.sqs_queue_arn(queue_name, TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME)
 
+    # EventBridge apparently converts some fields, for example: Source=>source, DetailType=>detail-type
+    # but the actual pattern matching is case-sensitive by key!
```
dfangl, line 251: That's fun

dfangl (approved):
Good test and a nice refacotring, looking forward to this being unified all over the place :)

tests/aws/services/events/test_event_patterns.py
joe4dev: Good question. It kinda evolved because I wanted to have an explicit way to express the intent of the test. The surprise effect helps to write better test cases. It could as well be a field in the the json5 file though.

Not really a strong reason, some thoughts:

* The intention of the test is easy to see in the file explorer.
* It currently matches the [TestEventPattern](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_TestEventPattern.html) API. Maybe, it would be easier to copy/paste and test in the AWS console.

tests/aws/services/events/test_event_patterns.py
```diff
@@ -0,0 +1,160 @@
...
+        exception_info = {"exception_type": type(e.value), "exception_message": e.value.response}
+        snapshot.match(label, exception_info)
+    else:
+        response = aws_client.events.test_event_pattern(
+            Event=json.dumps(event),
+            EventPattern=json.dumps(event_pattern),
+        )
+
+        # Validate the test intention: The _NEG suffix indicates negative tests (i.e., a pattern not matching the event)
+        if label.endswith("_NEG"):
```
joe4dev, line 114: It might be useful in the future to snapshot the result together with a hash of the request because currently, it's possible to modify the test case after being AWS-validated; kinda proper testing police ;)

tests/aws/services/events/test_event_patterns.py
joe4dev: I initially built a small templating engine supporting variables but noticed we can simplify and don't care about any dynamic values 🤦  So the entire request "templating" part comes from there 😅

PR #10604, opened 2024-04-04
fix s3 create bucket constraint

bentsku:
## Motivation
We had some issues with our WebApp regarding Bucket Location Constraint, and one proposed fixed was to pass a `undefined` location constraint when the bucket region was `us-east-1`. This fix worked with LocalStack, but this was missing parity. AWS will not accepted an empty constraint either for `us-east-1`. 

## Changes
- add a check if the constraint is actually set or not. Some quirks of XML parsing makes it that a full empty body and the location constraint set to `None` will end up as an empty dict for both. We have to check if the body is really empty, in that case no constraint are set. 
- extend the test to validate a constraint for `us-east-1` and validate a `None` constraint

\cc @webdev51 @lukqw

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -925,6 +925,13 @@ class KeyTooLongError(ServiceException):
     Size: Optional[KeyLength]
 
 
+class InvalidLocationConstraint(ServiceException):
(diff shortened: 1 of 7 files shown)


thrau (approved):
location constraint strike again! :grin: glad we have the patching. 

LGTM!

PR #10520, opened 2024-03-22
added aws validate test case for ListSecrets filtering

macnev2013:
## Motivation
This PR introduces AWS-validated test cases to verify the filtering functionality of the `secretsmanager` list API.

The issue was originally reported [here](https://github.com/localstack/localstack/issues/4820). The fix has been implemented in [Moto](https://github.com/getmoto/moto/pull/7511).

⚠️ NOTE: This PR should only be merged after the dependent PR in [Moto](https://github.com/getmoto/moto/pull/7511) has been merged and moto version in LocalStack is bumped to support this fix.

## Changes
- Implemented an AWS-validated test case.

cc. @viren-nadkarni

PR #10607, opened 2024-04-05
Revert "added aws validate test case for ListSecrets filtering"

macnev2013:
Reverts localstack/localstack#10520


alexrashed (approved):
Thanks for jumping on this, unblocking the pipeline!

PR #10608, opened 2024-04-05
added aws validate test case for ListSecrets filtering

macnev2013:
## Motivation
This PR addresses issues introduced by a recent addition to the test (PR https://github.com/localstack/localstack/pull/10520) (which was reverted  in PR https://github.com/localstack/localstack/pull/10607), that resulted in a failing test case due to leftover resources from other tests.

This PR builds upon the changes proposed in https://github.com/localstack/localstack/pull/10520 by refining the affected test case to ensure it operates independently from the outcomes of other tests.

## Changes
- Implemented an AWS-validated test case to enhance test's reliability.

### DIFF
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -283,7 +283,6 @@ class TestSecretsManager:
                 assert secret["Name"] in include_secrets
                 assert secret["Name"] not in exclude_secrets
 
-        # name based filtering
         response = aws_client.secretsmanager.list_secrets(
             Filters=[{"Key": "name", "Values": ["/"]}]
         )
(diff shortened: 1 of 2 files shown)


tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -261,6 +261,125 @@ def test_call_lists_secrets_multiple_times_snapshots(
         )
         sm_snapshot.match("delete_secret_res_1", delete_secret_res_1)
 
+    @markers.aws.validated
+    def test_list_secrets_filtering(self, aws_client, create_secret):
```
alexrashed, line 265: Using static names / identifiers can be problematic when this test is executed multiple times (retries?).
Also, humans are remarkably un-creative when it comes to naming (the next person writing a test will most likely create a similar identifier).
This is why the best practice would be to use random names for every test execution.
You could f.e. use `from localstack.utils.strings import short_uid`...

Do you think it would be feasible to create randomized identifiers in this test case without blowing up the complexity of the filter tests?

alexrashed (commented):
Thanks for re-enabling the test! It looks really good, I would just like to make sure that is really is resilient against the influence of other tests by avoiding static names...

alexrashed (approved):
Awesome! Thanks for re-adding the test and making it more resilient! 🥳

PR #10615, opened 2024-04-08
Add experimental event ruler

joe4dev:
## Motivation

We have several parity gaps and bugs in multiple different implementations for [event pattern matching](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-content-based-filtering.html) (see PR https://github.com/localstack/localstack/pull/10599). We can offer high AWS-parity quickly using the [Amazon EventBridge Event Bus event matching now runs on Open Source Event Ruler v1.5.0](https://aws.amazon.com/about-aws/whats-new/2024/02/amazon-eventbridge-event-bus-matching-event-ruler/) released `event-ruler` Java library from AWS described in [this blog post](https://aws.amazon.com/blogs/opensource/open-sourcing-event-ruler/).

## Changes

* Add an experimental opt-in `EVENT_RULE_ENGINE=java` flag using the Java [event-ruler](https://github.com/aws/event-ruler) library from AWS through [JPype](https://github.com/jpype-project/jpype).
* Add a `MavenPackageInstaller` utility ~~(could probably move to `localstack/packages/core.py` DONE)~~ 

## Testing

1. Set `EVENT_RULE_ENGINE=java`
2. Ensure that `JAVA_HOME` points to a valid JVM installation if you are testing in host mode
3. Uncomment all `SKIP_LABELS` for the test [test_test_event_pattern](https://github.com/localstack/localstack/blob/389d942ac65caf2f824d2ed88fbbe628d1a28b83/tests/aws/services/events/test_event_patterns.py#L87)
4. Run the test `test_test_event_pattern`

### LocalStack Shutdown

@dfangl and @joe4dev tested the LocalStack shutdown as follows:

1. Build LS image using `make docker-build`
2. Start LocalStack using `EVENT_RULE_ENGINE=java DEBUG=1 localstack start`
3. Run a test case for event pattern matching (e.g., `tests.aws.services.events.test_event_patterns.test_test_event_pattern_with_escape_characters) against the running LS instance
4. `docker stop $CONTAINER_ID` should stop the container immediately and not after a 10s delay

=> worked without delay

## Limitations

* The cross-language call from Python to Java is fast but could add a few milliseconds of delay.
* The tight coupling between Python and Java sharing the same process could crash each others language runtime upon unexpected errors.
* We had to disable the JVM shutdown because LocalStack has too many daemon threads running and JPype only triggers the shutdown after all daemon threads terminated to prevent shutting down the JVM while potentially still in use.
* The arm build failed with `ERROR: Could not build wheels for jpype1` due to `gcc: fatal error: cannot execute ‘cc1plus’: execvp: No such file or directory`. Supposedly, JPype [only supports up to Python 3.10](https://stackoverflow.com/a/77674837/6875981) 🤔, which would be a blocker 🔴  (they recently merged a fix for Python 3.12 though). We are using Python 3.11 in the image and jpype1 builds fine on my ARM MacBook using Python 3.11.7.
  * This build error only happened for ARM builds on Linux and required the extra dependency `apt-get install -y g++`.
  * For S3, we had to add this installation step into the make target `docker-run-tests-s3-only` to keep it out of the S3 image.

### Kinesisanalytics

Kinesisanalytics seems incompatible with `EVENT_RULE_ENGINE=java` indicated by the following [failing tests](https://github.com/localstack/localstack-ext/actions/runs/8734151459/job/23964542786):
- `TestKinesisAnalyticsProvider.test_tag_list_tag_untag_resource`
- `TestKinesisAnalyticsProvider.test_input_processing_configuration`
- `TestKinesisAnalyticsProvider.test_application_output`
- `TestKinesisAnalytics.test_run_siddhi_query`

⇒ all these tests keep failing in the AMD64 pipeline. They are not ARM compatible.
⇒ They fail with the assertion: `AssertionError: assert 'STARTING' in ['READY', 'RUNNING']`
⇒ The siddhi query test also fails with a name error `NameError: name 't_siddhi_api_core' is not defined` in the Siddhi loader `PySiddhi/SiddhiLoader.py`

LS uses some badly outdated fork of the unmaintained (5y; LS 3y last commit) PySiddhi project: https://github.com/localstack/PySiddhi. They are using some kind of custom Python⇒Java bridge based on JNI, which could cause interference.

ℹ️ We decided to ignore this limitation for now. If JPype proves successful, it might be a better alternative than PySiddhi ...

## TODO

- [x] Switch default `EVENT_RULE_ENGINE` back from `"java"` (used for CI testing) to `""`
- [x] https://github.com/localstack/docs/pull/1182

## Follow Up

* Fix LS Python-native implementation for `tests.aws.services.lambda_.test_lambda_integration_dynamodbstreams.TestDynamoDBEventSourceMapping.test_dynamodb_event_filter`
  * `numeric_filter` fails due to missing `anything-but` support
  * `exists_false_filter` fails due to broken implementation
  * => It is likely better to adopt Ben's SNS implementation and unify them all because the current status is hardly maintainable.


localstack/config.py
```diff
@@ -762,6 +762,11 @@ def populate_edge_configuration(
 # get-function call.
 INTERNAL_RESOURCE_ACCOUNT = os.environ.get("INTERNAL_RESOURCE_ACCOUNT") or "949334387222"
 
+# Determine which implementation to use for the event rule / event filtering engine used by multiple services:
+# EventBridge, EventBridge Pipes, Lambda Event Source Mapping, SNS
+# Options: provider (default) | java
+EVENT_RULE_ENGINE = os.environ.get("EVENT_RULE_ENGINE", "").strip()
```
joe4dev, line 768: ✅ CI run with JPype enabled using `java` here in [017bae5](https://github.com/localstack/localstack/pull/10615/commits/017bae519040155366f4e43feecddd790eb19907)

localstack/services/events/event_ruler.py
```diff
@@ -0,0 +1,57 @@
...
+if config.EVENT_RULE_ENGINE == "java":
+    import jpype
+    import jpype.imports
+    from jpype import config as jpype_config
+    from jpype import java
+
+    # Workaround to unblock LocalStack shutdown. By default, JPype wait until all daemon threads are terminated,
+    # which blocks the LocalStack shutdown upon any leaked thread (know LS issue).
+    # See https://github.com/MPh-py/MPh/issues/15#issuecomment-778486669
+    jpype_config.destroy_jvm = False
```
joe4dev, line 23: That's unfortunately needed. Could this cause any side-effects?

pyproject.toml
```diff
@@ -67,6 +67,8 @@ base-runtime = [
     "Werkzeug>=3.0.0",
     "xmltodict>=0.13.0",
     "rolo>=0.4",
+    # allow Python programs full access to Java class libraries. Used for opt-in event ruler.
+    "JPype1>=1.5.0",
```
joe4dev, line 71: 431KB without any extra dependencies (see their [pyproject.toml](https://github.com/jpype-project/jpype/blob/master/pyproject.toml#L31)), just `packaging` and `typing_extensions`, which we have as well

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
```diff
@@ -411,67 +411,110 @@ def verify_failure_received():
...
+                {"dynamodb": {"NewImage": {"presentKey": {"S": [{"exists": True}]}}}},
+                1,
+                id="exists_filter_type",
             ),
+            # TODO: Fix native LocalStack implementation for exists
+            # pytest.param(
+            #     {"id": {"S": "id_value_1"}},
+            #     {"id": {"S": "id_value_2"}, "presentKey": {"S": "presentValue"}},
+            #     {"dynamodb": {"NewImage": {"presentKey": [{"exists": False}]}}},
+            #     2,
```
joe4dev, line 472: This false negative took me forever to spot. A combination of bad test design and inappropriate configuration led to wrong assumptions!
It was configured to 1 but both events match against AWS. However, the second event never gets waited. Negative testing for the last event is unreliable!

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
```diff
@@ -486,14 +529,22 @@ def test_dynamodb_event_filter(
...
+        """
         function_name = f"lambda_func-{short_uid()}"
         table_name = f"test-table-{short_uid()}"
         max_retries = 50
 
         create_lambda_function(
             handler_file=TEST_LAMBDA_PYTHON_ECHO,
             func_name=function_name,
-            runtime=Runtime.python3_9,
+            runtime=Runtime.python3_12,
```
joe4dev, line 547: /cc @Morijarti I'm bumping the runtime here together with significant test changes, so can can simply take over my changes upon rebasing

tests/unit/services/lambda_/test_lambda_utils.py
```diff
@@ -136,7 +136,7 @@ def test_no_match_partial(self):
         filters = [
             {
                 "Filters": [
-                    {"Pattern": json.dumps({"partitionKey": "2", "data": {"City": ["Seattle"]}})}
+                    {"Pattern": json.dumps({"partitionKey": ["2"], "data": {"City": ["Seattle"]}})}
```
joe4dev, line 139: another example why non-validated unit-tests are tricky to get right 😢

Dockerfile.s3
```diff
@@ -48,7 +48,9 @@ ARG TARGETARCH
 RUN --mount=type=cache,target=/var/cache/apt \
     apt-get update && \
         # Install dependencies to add additional repos
-        apt-get install -y gcc
+        apt-get install -y gcc && \
+        # Workaround to fix JPype1 compile error on ARM Linux "gcc: fatal error: cannot execute ‘cc1plus’"
+        apt-get install -y g++
```
joe4dev, line 53: I wished this was not necessary, especially not needed for S3 😿

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
Morijarti: Awesome, thank you

Dockerfile.s3
alexrashed: I guess this would be a bit faster if it would be in a single `apt-get install` call? 🤔

alexrashed: And maybe this should be added somewhere in the contribution guide? Because it seems Mac users will struggle to run `make install` without `g++` (and Java when they use this feature) on their machines?

alexrashed (commented):
Wow, this a pretty sophisticated change which drastically improve parity with the event pattern matching! 💯 
I hope the new dependency doesn't cause too many issues with https://github.com/localstack/localstack/pull/10235 (but as you stated, they seem to work on compatibility above 3.10).
Besides that I only had some thoughts concerning the internal developer experience (Mac users need to be aware that they have to install `g++`) and some smaller nitpicks ;)
Besides that, I'll leave the final signoff to the service owners :)

Dockerfile.s3
joe4dev: +1 for a single apt-get call, updated ✅ 

The `g++` dependency might be specific to ARM Linux because it worked out of the box for me on an ARM MacBook.

I agree that the Java installation and setting `JAVA_HOME` is a bit cumbersome in host mode 😿
@alexrashed Where is the right place to document service-specific requirements for host mode? (we have Notion, the dev guide on the website, getting started guide)

localstack/services/events/packages.py
```diff
@@ -0,0 +1,86 @@
...
+from localstack.packages import DownloadInstaller, Package, PackageInstaller
+from localstack.packages.api import InstallTarget, MultiPackageInstaller
+
+# https://central.sonatype.com/artifact/software.amazon.event.ruler/event-ruler
+EVENT_RULER_VERSION = "1.7.3"
+# The dependent jackson.version is defined in the Maven POM File of event-ruler
+JACKSON_VERSION = "2.16.2"
+
+
+class MavenPackageInstaller(DownloadInstaller):
```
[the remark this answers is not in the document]
joe4dev: I had the same two thoughts 😆 
it might be harder to discover at the current place and it's quite likely we can reuse it :)

localstack/services/events/packages.py
```diff
@@ -0,0 +1,86 @@
...
+    def _get_installer(self, version: str) -> PackageInstaller:
+        return EventRulerPackageInstaller()
+
+
+class EventRulerPackageInstaller(MultiPackageInstaller):
+    def __init__(self):
+        event_ruler = MavenPackageInstaller(
+            f"pkg:maven/software.amazon.event.ruler/event-ruler@{EVENT_RULER_VERSION}"
+        )
+        event_ruler_dir = event_ruler._get_install_dir(InstallTarget.VAR_LIBS)
```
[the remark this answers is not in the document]
joe4dev: @alexrashed How can we solve this alternatively?

The EventRuler package installer aims to bundle all dependent jars for a given event ruler version like this rather than having each jar in its own directory, which would require copying them afterward:
Screenshot 2024-04-12 at 11 32 24

localstack/services/events/packages.py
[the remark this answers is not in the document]
joe4dev: Moved ✅

localstack/services/events/packages.py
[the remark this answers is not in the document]
alexrashed: Hmm, I feel like the usage of the `MultiPackageInstaller` is counter productive. It breaks with the definition of a package, where a package is an independently used piece of software with a name and a version.
I think this is the typical use case when you download Jar files from Maven: You could implement the MavenPackageInstaller in a way that it accepts a name and a version, but also a list of group/artifact IDs with their versions which should be downloaded.

This would mean that the installer has a name and a version (event ruler version X), which consists of multiple maven artifacts to be downloaded.

Does that make sense?

Dockerfile.s3
alexrashed: I think this should be part of the contribution guidelines (because it affects all open source contributors), which is currently located in our docs.

localstack/services/events/packages.py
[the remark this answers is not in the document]
joe4dev: Yeah, I see. So basically like this:

```py
        return MavenPackageInstaller(
            f"pkg:maven/software.amazon.event.ruler/event-ruler@{EVENT_RULER_VERSION}",
            f"pkg:maven/com.fasterxml.jackson.core/jackson-annotations@{JACKSON_VERSION}",
            f"pkg:maven/com.fasterxml.jackson.core/jackson-core@{JACKSON_VERSION}",
            f"pkg:maven/com.fasterxml.jackson.core/jackson-databind@{JACKSON_VERSION}",
        )
```

I tried to re-use the modular constructs such as the [DownloadInstaller](https://github.com/localstack/localstack/blob/7c7cd158212dfabc37c9e1a39d0a03e60a124e9a/localstack/packages/core.py#L300) for better consistency but that does not seem possible using this approach.

alexrashed (commented):
Awesome! 🚀 Thanks a lot for addressing all the comments! From my side all these changes are looking good! As mentioned in the initial comment, I'll leave the final sign-off to the code owners 🥳

localstack/services/events/provider.py
```diff
@@ -110,28 +115,44 @@ def test_event_pattern(
...
-
-        # EventBridge implementation:
-        result = matches_event(event_pattern_dict, event_dict)
-
-        # EventSourceMapping implementation:
+        if config.EVENT_RULE_ENGINE == "java":
+            try:
+                result = matches_rule(event, event_pattern)
+            except InternalInvalidEventPatternException as e:
+                raise InvalidEventPatternException(e.message)
```
dfangl, line 122:
```suggestion
                raise InvalidEventPatternException(e.message) from e
```

localstack/services/events/event_ruler.py
dfangl: The container shutdown will probably not work, as we have an orphan process which did not get terminated by its parent on shutdown. The supervisor will probably wait forever, until docker kills the container.

localstack/services/lambda_/event_source_listeners/utils.py
```diff
@@ -21,8 +23,14 @@ def filter_stream_records(records, filters: list[FilterCriteria]):
...
-                filter_pattern: dict[str, any] = json.loads(rule["Pattern"])
-                if does_match_event(filter_pattern, record):
+                if config.EVENT_RULE_ENGINE == "java":
+                    event_str = json.dumps(record)
+                    event_pattern_str = rule["Pattern"]
+                    match_result = matches_rule(event_str, event_pattern_str)
+                else:
+                    filter_pattern: dict[str, any] = json.loads(rule["Pattern"])
+                    match_result = does_match_event(filter_pattern, record)
+                if match_result:
```
dfangl, line 33: Could we unify the interfaces, so we do not have to have a config switch before we invoke the method, but within the method? I see this `if config.EVENT_RULE_ENGINE` change a lot, but it really should be hidden from the users of the method, what do you think? Similar to the method in the ext call, we should have a unified interface for both.

localstack/services/events/event_ruler.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+
+        try:
+            # "Static rule matching" is the easiest implementation to get started.
+            # "Matching with a machine" using a compiled machine is faster and enables rule validation before matching.
+            # https://github.com/aws/event-ruler?tab=readme-ov-file#matching-with-a-machine
+            return Ruler.matchesRule(event, rule)
+        except java.lang.Exception as e:
+            reason = e.args[0]
+            raise InvalidEventPatternException(reason=reason)
```
dfangl, line 52:
```suggestion
            raise InvalidEventPatternException(reason=reason) from e
```

localstack/services/events/event_ruler.py
```diff
@@ -0,0 +1,57 @@
...
+THIS_FOLDER = os.path.dirname(os.path.realpath(__file__))
+
+LOG = logging.getLogger(__name__)
+
+if config.EVENT_RULE_ENGINE == "java":
+    import jpype
+    import jpype.imports
+    from jpype import config as jpype_config
+    from jpype import java
+
```
dfangl, line 19: Is it necessary to do this on module level on import? including the startup? Could it be done on first match as well?

tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py
dfangl: Will this test be reintroduced at a later point?

localstack/services/lambda_/event_source_listeners/dynamodb_event_source_listener.py
```diff
@@ -76,3 +73,14 @@ def _get_first_and_last_arrival_time(self, first_record, last_record):
...
+    def _transform_records(self, raw_records: list[dict]) -> list[dict]:
+        """Convert dynamodb.ApproximateCreationDateTime datetime to float"""
+        records_new = []
+        for record in raw_records:
+            record_new = record.copy()
+            if creation_time := record.get("dynamodb", {}).get("ApproximateCreationDateTime"):
+                # convert datetime object to float timestamp
+                record_new["dynamodb"]["ApproximateCreationDateTime"] = creation_time.timestamp()
+            records_new.append(record_new)
+        return records_new
```
dfangl, line 86: How is this change related, and why do we need to copy the records instead of modifying in place?

joe4dev: It's a parity issue captured by the improved parity tests and is related because it caused parsing issues in the event-ruler.

That's why it's needed: `json.dumps` is not happy with this one ;)
> Convert dynamodb.ApproximateCreationDateTime datetime to float

localstack/services/events/event_ruler.py
joe4dev: I guess we can go full dynamic import upon first event matching:
* it could hide potential issues
* JVM startup delay is probably not an issue though.
 
Do you think a lazy import is better here?

localstack/services/lambda_/event_source_listeners/utils.py
joe4dev: Yes, I think unifying should be the goal. Unfortunately, our implementations (see https://github.com/localstack/localstack/pull/10599) behave so differently that we first need to fix cross-implementation regressions before we can have a single unified method. Therefore, I just added an opt-in event ruler for each existing implementation (i.e., the else depends on the context) and pushed unification into a follow-up. Or do you think we should add such a helper method anyway for each implementation?

localstack/services/events/provider.py
joe4dev: Done 👍

joe4dev:
Trying to reply to @dfangl comment: https://github.com/localstack/localstack/pull/10615#discussion_r1565783339
> The container shutdown will probably not work, as we have an orphan process which did not get terminated by its parent on shutdown. The supervisor will probably wait forever, until docker kills the container.

This a problem ⚠️ 😢 

How should we test this? Any suggestion on how to fix this?

localstack/services/lambda_/event_source_listeners/utils.py
joe4dev: As discussed with Daniel, let's skip this until we can unify the different implementations.

localstack/services/events/event_ruler.py
joe4dev: Resolved using a dynamic import now (as discussed in sync).

joe4dev:
Using `EVENT_RULE_ENGINE=java`, all tests are 🟢 in community. However, in ext, we have some Kinesis analytics test still failing (e.g., `TestKinesisAnalyticsProvider.test_application_output`).

The failing Kinesisanalytics tests are likely caused by an interaction with our Kinesisanalytics implementation using https://github.com/localstack/PySiddhi, an unmaintained project that uses some kind of custom Python=>Java bridge. This [C-based](https://github.com/localstack/PySiddhi/blob/master/__PySiddhiProxy/threadfix_c_code/io_siddhi_pythonapi_threadfix_PyThreadFix.c) proxy uses JNI (like JPype) and could potentially cause an unintended interaction.


Failing [ext test run](https://github.com/localstack/localstack-ext/actions/runs/8703976069/job/23882376476):

- `TestKinesisAnalyticsProvider.test_tag_list_tag_untag_resource`
- `TestKinesisAnalyticsProvider.test_input_processing_configuration`
- `TestKinesisAnalyticsProvider.test_application_output`
- `TestKinesisAnalytics.test_run_siddhi_query`

joe4dev:
Without using the new opt-in Java-based event-ruler, all tests are green 🟢 , both in community and ext 🚀

dfangl (approved):
Looks good now, thank you for addressing my comments and avoiding start of the jvm on module import.
Looking forward for more unification of all the event matching implementations :)

andy-g:
Could this also be supported when running localstack in a docker container? And I assume this will resolve the current issues with what appears to be unsupported syntax such as:  
* [$or](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-content-based-filtering.html#eb-filtering-complex-example-or) matching
* [exists](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns-content-based-filtering.html#eb-filtering-exists-matching) matching

There are current matching rules that don't appear to be working for me, I'm currently running localstack 3.3 in docker.

joe4dev:
Thanks for the comment @andy-g 

> Could this also be supported when running localstack in a docker container?

Yes, it will work in Docker.

> And I assume this will resolve the current issues with what appears to be unsupported syntax such as:
Correct, I assume `EVENT_RULE_ENGINE=java` should fix all issues related to event pattern matching for EventBridge, EventBridge Pipes, and Lambda Event Source Mapping. It should reach almost 100% AWS parity. We currently do not yet implement the same complexity calculation, so very complex pattern might work in LocalStack but throw an exception in AWS.
We are also looking into how to adopt the same rule engine for SNS (which already implements some complexity calculations). //cc @bentsku

All builds are 🟢 , so I expect this to be merged anytime soon 🚀

andy-g:
> All builds are 🟢 , so I expect this to be merged anytime soon 🚀

Brilliant, thanks @joe4dev, I'll keep an eye on this PR then, and thanks so much for the quick response.

pyproject.toml
bentsku: could we move it to `runtime` instead? `base-runtime` is the base one used by the S3 image, which has no dependency basically. Or was there a specific reason to put into `base-runtime`?

Dockerfile.s3
bentsku: Why do we need it for S3? Because we put it into `base-runtime`? Could we just move the dependency into `runtime` instead? I'm not sure I entirely follow here.

bentsku (commented):
I've added 2 comments: https://github.com/localstack/localstack/pull/10615#discussion_r1569587259 and https://github.com/localstack/localstack/pull/10615#discussion_r1569588431, which I believe should be addressed before merging if possible, as this PR would then have no impact on the S3 image or base runtime. There might be something I missed, feel free to correct me if it's really needed to go into `base-runtime`. Thanks!

pyproject.toml
joe4dev: That makes a lot of sense @bentsku 💯 Thank you for this valuable input 👀  👏 
I fixed everything (including the follow-up S3 tests because the test deps include the runtime deps 😬 ; that's a bit of an ugly hack though)

I didn't understand the full implications of `base-runtime` vs `runtime` dependencies.

With the fixed solution, the S3 image does not get larger 👌

Dockerfile.s3
joe4dev: We don't need it for S3. Thanks for suggesting the switch to `runtime` deps 👍 
Fixed as commented in https://github.com/localstack/localstack/pull/10615#discussion_r1570279927

bentsku (approved):
thanks a lot for removing the changes for the S3 image, sorry it has been a bit complicated with the tests dependencies and had to go further down the path of quick hacks in the make target 😬 but very grateful you've addressed it. Thank you so much!

And sorry again for missing this and coming in late requesting changes. 😞 

Great work, can't wait for us to have a unified event ruler! 🚀

andy-g:
@joe4dev I see this was merged, and I see that new docker images have been built, am I right in assuming these changes should be in latest?

I'm running:
```
docker pull localstack/localstack:latest
latest: Pulling from localstack/localstack
Digest: sha256:e9b46037d6af0b947aad43895edf371b9de96b17cacb52383f5c6c7a2467817a
Status: Image is up to date for localstack/localstack:latest
docker.io/localstack/localstack:latest
```

Using this rule my lambda works (as it has previously):
```json
{
  "detail-type": ["PaymentReceived","PaymentCompleted"],
  "source": ["producer-1"]
}
```

...but changing my rule to either of these result in the lambda not being invoked (I was hoping this PR would change that):  
```json
{
  "detail-type": ["PaymentReceived"],
  "detail.payload.paymentDetails": {
    "url": [{"exists": true}],
    "type": ["Multi"]
  },
  "source": ["producer-1"]
}
```
```
{
  "$or": [
    {
      "detail-type": ["PaymentCompleted"],
      "detail.payload.url": [{"exists": true}]
    },
    {
      "detail-type": ["PaymentReceived"],
      "detail.payload.paymentDetails": {
        "url": [{"exists": true}],
        "type": ["Multi"]
      }
    }
  ],
  "source": ["producer-1"]
}
```

These are working fine on AWS itself.

Am I missing something here?

andy-g:
Sorry, after a closer look I realised I needed to set the `EVENT_RULE_ENGINE` environment variable to `java`. I added that to my docker-compose environment array and it's working like a charm, thanks so much 🍻 .

joe4dev:
> I see this was merged, and I see that new docker images have been built, am I right in assuming these changes should be in latest?

Yes, the changes are released for `localstack/localstack` but NOT yet for `localstack/localstack-pro` (I just merged a fix for a hidden ARM build issue in our CI, hence the pro image should be released in the next few hours 🤞 ).

@andy-g Awesome to hear everything works as expected using `EVENT_RULE_ENGINE=java` 🚀 🙌
We wanted to learn more how the Python=>Java bridge works before rolling out that feature by default. That's why we opted for a feature flag. I'm looking forward to hearing your feedback if you notice anything (e.g., performance, stability etc) 👂

joe4dev:
@andy-g I can confirm the `localstack/localstack-pro` image is now successfully released and the new opt-in event-ruler is working 🚀

Issue #10351, opened 2024-02-28
enhancement request: Emulate more of AWS S3 errors (specifically "EntityTooLarge" and "Policy Condition failed")


caitlinshk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Hi there!

My project uses LocalStack to emulate file uploads to AWS S3 using a `<form>` element and it overall works great. But, I've noticed that there's some errors that are only thrown in real AWS, not in LocalStack. This means that I have to wait until we deploy to our staging environment to know when AWS throws errors and what those errors are. It would be nicer if LocalStack could throw the same errors in the same situations so I can (1) not need to deploy to staging to see these errors; and (2) can fix the errors locally and know that my fix works.

The two specific errors I've seen that aren't in LocalStack are:

### EntityTooLarge

We enforce that files uploaded to S3 can have a maximum size by [setting the `"content-length-range"` on the AWS upload request](https://github.com/civiform/civiform/blob/0e9425f85de22f2c15f904c2a70eec8c79aef62b/server/app/services/cloud/aws/SignedS3UploadRequest.java#L330-L346). If a user uploads a file larger than that see, AWS sends this error:
```
<Error>
  <Code>EntityTooLarge</Code>
  <Message>Your proposed upload exceeds the maximum allowed size</Message>
</Error>
```

but LocalStack will let the file upload proceed like normal. See https://github.com/civiform/civiform/issues/5675 and https://github.com/civiform/civiform/issues/6658.

### Policy condition failed
AWS enforces that the value in the `<input name="success_action_redirect">` element matches the value in the `<input name="Policy">` element. If they don't match, AWS sends this error:

```
<Error>
  <Code>AccessDenied</Code>
  <Message>Invalid according to Policy: Policy Condition failed: ["eq", "$success_action_redirect", "https://staging-aws.civiform.dev/programs/92148/blocks/7/updateFile/true/NEXT_BLOCK"]</Message>
</Error>
```
but LocalStack also let the file upload proceed like normal even when they didn't match. See https://github.com/civiform/civiform/issues/6737.

--------------

It'd be amazing if LocalStack could all throw these errors in the right circumstances so I notice these issues immediately instead of only noticing when we deploy to our staging environment with real AWS.

Thanks!
Caitlin

bentsku:
Hello @caitlinshk and thanks a lot for the very detailed comment!

[...]

PR #10616, opened 2024-04-08
add s3 POST policy validation

bentsku:
## Motivation
As requested with #10351, we did not implement specific policy validation for pre-signed POST. 
We were lacking validating the conditions specified in the policy. This PR will take a stab at implementing the requested ones with a general implementation. 

See: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html
Some are still left to do: 
- AWS will reject a policy that has extra fields passed to it that are not defined in the `Conditions` field (for ex. if you set `success_action_redirect` without defining it as a condition, the request should fail). This is not implemented yet as we need to know and test exactly which fields are considered "extra", with help from the documentation but will need quite some tests
- some fields only support one kind of matching (usually "exact matching" and some support both exact and "starts-with" matching). Right now we do not make the distinction, we will need some kind of map to validate that. Can be done at the same time as the other missing feature listed right above

## Changes
Added validation of the conditions using structural pattern matching, which I hope make it more readable, otherwise we can fall back to some de-structuring + some if/else statements. 

Added tests around the defined use cases of the feature requests + some negative testing. 

_fixes #10351_

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -932,6 +932,15 @@ class InvalidLocationConstraint(ServiceException):
     LocationConstraint: Optional[BucketRegion]
 
 
+class EntityTooLarge(ServiceException):
+    code: str = "EntityTooLarge"
+    sender_fault: bool = False
+    status_code: int = 400
+    MaxSizeAllowed: Optional[KeyLength]
+    HostId: Optional[HostId]
+    ProposedSize: Optional[ProposedSize]
+
+
 AbortDate = datetime
 
 
--- a/localstack/services/s3/v3/storage/core.py
+++ b/localstack/services/s3/v3/storage/core.py
@@ -2,7 +2,7 @@ import abc
 from io import RawIOBase
 from typing import IO, Iterable, Iterator, Literal, Optional
 
-from localstack.aws.api.s3 import BucketName, MultipartUploadId, PartNumber
+from localstack.aws.api.s3 import BucketName, PartNumber
 from localstack.services.s3.utils import ObjectRange
 from localstack.services.s3.v3.models import S3Multipart, S3Object, S3Part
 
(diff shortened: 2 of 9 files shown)


localstack/services/s3/v3/storage/core.py
```diff
@@ -202,7 +202,7 @@ def copy(
         pass
 
     @abc.abstractmethod
-    def get_multipart(self, bucket: BucketName, upload_id: MultipartUploadId) -> S3StoredMultipart:
+    def get_multipart(self, bucket: BucketName, upload_id: S3Multipart) -> S3StoredMultipart:
```
bentsku, line 205: sneaked in a tiny typing issue on the base class of the storage abstraction

localstack/aws/api/s3/__init__.py
```diff
@@ -932,6 +932,15 @@ class InvalidLocationConstraint(ServiceException):
     LocationConstraint: Optional[BucketRegion]
 
 
+class EntityTooLarge(ServiceException):
```
cloutierMat, line 935: I looked at the `EntityTooSmall` class. It does have different keys, but the snapshot appears the be identical. Is it just some more of the S3 patch weirdness or an outdated scaffold?

cloutierMat (commented):
Clean implementation with lots of relevant tests! :+1: 

nit: There are no tests that uses `"starts-with"` with a value other than `""`.

Question: Does capitalisation matter for those policies (key or value). Would it be relevant to have a test for it?

localstack/aws/api/s3/__init__.py
bentsku: Nice check! This is because I've already added `EntityTooSmall` earlier as this exception is also raised by `CompleteMultipartUpload` when a part which is not the last part is under 5Mb... so I could add the specific one here, but reused it. The shape is similar. Good catch 😄

bentsku:
> nit: There are no tests that uses "starts-with" with a value other than "".
Question: Does capitalisation matter for those policies (key or value). Would it be relevant to have a test for it?

Good catch! I think this is a good point, I'll try to simplify the tests also as there are a lot of repetitive parts. I'll also check the 2 missing ones! 👍

bentsku:
Thanks for review, I've re-requested it for you to validate the changes 😄 there was indeed some weird things with the condition key casing, which is now fixed. Thanks a lot for catching this!

cloutierMat (approved):
Looks terrific. So many tests, I love it.

I wrote one last nitpick, feel free to ignore and merge if you don't fell it relevant.

PR #10624, opened 2024-04-09
Add container labels to container configuration and container run methods

dfangl:
## Motivation
We are currently not supporting container labels the `run_container` method and the `ContainerConfiguration` class.
Since we want to make use of labels more often, to allow grouping containers and adding additional information, we should allow setting labels for these options to create containers.

## Changes
* Add `labels` parameter to `run_container` methods
* Add `labels` as a new field in the `ContainerConfiguration` class
* Add small test assuring `labels` to be correctly set during the `run_container` call

## Testing

We currently have no explicit tests for the `ContainerConfiguration` class, but test it rather implicitly. The `labels` field of it will be tested while testing new `ecs` changes.


simonrw (approved):
LGTM!

PR #10627, opened 2024-04-10
fix secretsmanager test

alexrashed:
## Motivation
- https://github.com/localstack/localstack/pull/10520 introduced a test for filtering secrets in secretsmanager.
  - Unfortunately, the assertions were way too strict, any leftover secret from another test would let this test fail.
- https://github.com/localstack/localstack/pull/10607 reverted the test then.
- https://github.com/localstack/localstack/pull/10608 tried to re-introduce the test.

Unfortunately, it turns out that it still fails on a single leftover from another test (similar to https://github.com/localstack/localstack/pull/10520).
This PR aims at properly fixing the test.

## Changes
- Cleanup `test_list_secrets_filtering`, fix assertions, make it resilient to other changes.

## TODO
- [ ] Fix all the assertions, it seems like a lot of them are wrong.
- [ ] Re-validate against AWS.

## Testing
- [ ] Run the ARM test suite where it seems that the test ordering also executes a stepfunctions test on the same runner (leaving a secret behind).

### DIFF
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import random
 import uuid
 from datetime import datetime
 from math import isclose
@@ -263,12 +264,13 @@ class TestSecretsManager:
 
     @markers.aws.validated
     def test_list_secrets_filtering(self, aws_client, create_secret):
-        unique_id = short_uid()
-        secret_name_1 = f"testing1/one-{unique_id}"
-        secret_name_2 = f"/testing2/two-{unique_id}"
-        secret_name_3 = f"testing3/three-{unique_id}"
-        secret_name_4 = f"/testing4/four-{unique_id}"
+        suffix = random.randint(10000, 99999)
+        secret_name_1 = f"testing1/one-{suffix}"
+        secret_name_2 = f"/testing2/two-{suffix}"
+        secret_name_3 = f"testing3/three-{suffix}"
+        secret_name_4 = f"/testing4/four-{suffix}"
 
+        create_secret(Name="MyTestDatabaseSecret", SecretString="secret", Description="a secret")
         create_secret(Name=secret_name_1, SecretString="secret", Description="a secret")
         create_secret(Name=secret_name_2, SecretString="secret", Description="an secret")
         create_secret(Name=secret_name_3, SecretString="secret", Description="asecret")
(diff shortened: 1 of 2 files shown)


tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -279,51 +280,55 @@ def test_list_secrets_filtering(self, aws_client, create_secret):
...
-                assert secret["Name"] in include_secrets
-                assert secret["Name"] not in exclude_secrets
+        def assert_secret_names(res: dict, include_secrets: set[str], exclude_secrets: set[str]):
+            secret_names = {secret["Name"] for secret in res["SecretList"]}
+            assert (
+                include_secrets - secret_names
+            ) == set(), "At least one secret which should be included is not."
+            assert (
+                exclude_secrets - secret_names
+            ) == exclude_secrets, "At least one secret which should not be included is."
```
silv-io, line 290: nice fix using set semantics here. But is there a reason why we don't just do a snapshot test here?

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -263,11 +264,11 @@ def test_call_lists_secrets_multiple_times_snapshots(
...
-        unique_id = short_uid()
-        secret_name_1 = f"testing1/one-{unique_id}"
-        secret_name_2 = f"/testing2/two-{unique_id}"
-        secret_name_3 = f"testing3/three-{unique_id}"
-        secret_name_4 = f"/testing4/four-{unique_id}"
+        suffix = random.randint(10000, 99999)
+        secret_name_1 = f"testing1/one-{suffix}"
+        secret_name_2 = f"/testing2/two-{suffix}"
+        secret_name_3 = f"testing3/three-{suffix}"
+        secret_name_4 = f"/testing4/four-{suffix}"
```
silv-io, line 271: Does `short_uid()` produce characters that are not allowed in secret names?

tests/aws/services/stepfunctions/v2/test_stepfunctions_v2.py
```diff
@@ -831,3 +831,4 @@ def check_invocations():
 
     # clean up
     cleanup(sm_arn, state_machines_before, aws_client.stepfunctions)
+    # TODO also clean up other resources (like secrets)
```
silv-io, line 834: Since this is something that's definitely broken, I'd use `FIXME` here instead of `TODO`. But I don't know if we have a proper decision on when to use which here.

silv-io (approved):
LGTM! just some minor questions :)

tests/aws/services/secretsmanager/test_secretsmanager.py
alexrashed: The problem is that the filter tests are based on characters and a random "a" in the suffix could break the tests. ;)

tests/aws/services/secretsmanager/test_secretsmanager.py
alexrashed: The problem is that we really want to make this test resilient against any other secrets being left over from other tests (because that's what broke the first two tries). The filters are very generic and a leftover secret could leak into a snapshot / response. This is why this assert really just checks if certain secrets are in the response and others are not (while leaving the possibility of other secrets being in the response).

tests/aws/services/stepfunctions/v2/test_stepfunctions_v2.py
alexrashed: Yeah, that's a subtle difference where we don't have a clear structure yet. I do agree that a FIXME would maybe fit better, but for the sake of unblocking master asap, I would keep it like this for now.

PR #10630, opened 2024-04-10
Ensure container clients return the same format for container labels on list call

dfangl:
## Motivation
Currently, the `docker_cmd_client` returns labels on the `list_container` operations in this format: `key1=value1,key2=value2`
The SDK based client however returns a dict: `{"key1": "label1", "key2": "label2"}`

As they should behave identical, we need to change the CMD client to use the same dict format (as it is easier to use).

## Changes
* Change the CMD docker client to return a dict of labels, instead of a comma separated list of key-value pairs
* Add a test asserting that format

### DIFF
--- a/localstack/utils/container_utils/docker_cmd_client.py
+++ b/localstack/utils/container_utils/docker_cmd_client.py
@@ -833,7 +833,7 @@ class CmdDockerClient(ContainerClient):
         if any(msg.lower() in process_stdout_lower for msg in error_messages):
             raise NoSuchContainer(container_name_or_id, stdout=error.stdout, stderr=error.stderr)
 
-    def _transform_container_labels(self, labels: str) -> dict[str, str]:
+    def _transform_container_labels(self, labels: str) -> Dict[str, str]:
         """
         Transforms the container labels returned by the docker command from the key-value pair format to a dict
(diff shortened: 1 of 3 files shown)


localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -831,3 +832,13 @@ def _check_and_raise_no_such_container_error(
         process_stdout_lower = to_str(error.stdout).lower()
         if any(msg.lower() in process_stdout_lower for msg in error_messages):
             raise NoSuchContainer(container_name_or_id, stdout=error.stdout, stderr=error.stderr)
+
+    def _transform_container_labels(self, labels: str) -> Dict[str, str]:
```
simonrw, line 836: How stable is this format do you think? Is there any chance the Docker client might add any whitespace, for example?

simonrw (approved):
Thanks for making this correction. Good to have consistency across the implementations with labels

localstack/utils/container_utils/docker_cmd_client.py
dfangl: This is only used on the return values of `docker ps` with the format set to `json`. I think it should be rather stable, as a change there could break a lot of automated tools.

PR #10633, opened 2024-04-10
fix docker client ulimits

cloutierMat:
## Motivation

Trying to setup ulimits for a container using the `ContainerClient.run_container()` method leads to invalid configuration. 

The sdk client tried to transform the Ulimit object twice. Once in `run_container` and then again in `start_container`. Removed it from `run_container` as it calls the `start_container`, which should keep it's interface.

The cmd client had a typo in ulimit.

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -25,6 +25,7 @@ from localstack.utils.container_utils.container_client import (
     NoSuchNetwork,
     PortMappings,
     RegistryConnectionError,
+    Ulimit,
     Util,
     VolumeInfo,
 )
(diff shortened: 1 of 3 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -1432,6 +1433,28 @@ def test_run_with_additional_arguments_random_port(
...
+    def test_run_with_ulimit(self, docker_client: ContainerClient):
+        # test baseline without ulimit
+        container_name = f"c-{short_uid()}"
+        stdout, _ = docker_client.run_container(
+            "alpine",
+            name=container_name,
+            remove=True,
+            command=["sh", "-c", "ulimit -n"],
+        )
+        assert stdout.decode(config.DEFAULT_ENCODING).strip() == "1073741816"
```
dfangl, line 1445: This assertion will be highly system specific. I would rather avoid constraints which make the test fail depending on the system limit configuration.

dfangl (changes requested):
In principal the changes look good, I just have one concern regarding the test for default limits.

tests/integration/docker_utils/test_docker.py
cloutierMat: Is it? I thought those limit were set by the image. I will remove that section.

dfangl: Usually, those are inherited from the docker daemon, and are whatever that process is launched with :)

dfangl (approved):
Looks good, thanks for the fix!

PR #10634, opened 2024-04-10
Check for VPC existence for hostedzone

simonrw:
## Motivation

If a hosted zone is to be created in a VPC, `moto` does not check for existence for that VPC whereas AWS does.

## Changes

* If the VPC details are specified by the caller of `create_hosted_zone`, look up the VPC and raise an error if it does not exist.
* Update `test_create_private_hosted_zone` to call `list_hosted_zones_by_vpc` and snapshot the result


giograno (approved):
💯

PR #10639, opened 2024-04-11
More consistent CFn logging for individual resources

simonrw:
## Motivation

When adding new resources, CFn logs a message like:

```
Handling "Add" for resource "V1MysqlClusterSecretAttachment649A3B90" (1/1) type "AWS::SecretsManager::SecretTargetAttachment" in loop iteration 2
```

Which is useful context.

When deleting resources, CFn does not print anything 😥

## Changes

- Add log message for handling removes during stack deletes
- Add logging for _resource_ removal


Morijarti (approved):
I love this.
More logging == easier debugging

PR #10640, opened 2024-04-11
fix lambda function URLs when they don't exist or the service is not loaded

thrau:
## Motivation

I noticed an error when calling non-existing lambda function URLs. The code we had was apparently faulty, and I got this error:

```
2024-04-11T17:20:40.966 ERROR --- [et.reactor-0] l.aws.handlers.logging     : exception during call chain
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/handlers.py", line 27, in __call__
    router_response = self.router.dispatch(context.request)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/router.py", line 378, in dispatch
    return self.dispatcher(request, handler, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/dispatcher.py", line 71, in _dispatch
    result = endpoint(request, **args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/lambda_/urlrouter.py", line 79, in handle_lambda_url_invocation
    match = FULL_FN_ARN_PATTERN.search(lambda_url_config.function_arn).groupdict()
```

After writing a test, I further ran into a problem that we've had in the past, which is that, services that have a data plane are not properly loaded based on calling URLs of the data plane, even though we could.

Previously localstack would behave like this when calling a lambda function URL without lambda being loaded first (e.g., by calling `awslocal lambda list-functions`.

```
 % curl http://a38wes0ubqxh7ipckjdl8ilptt40uiha.lambda-url.us-east-1.localhost.localstack.cloud:4566
{"__type": "InternalError", "message": "exception while calling lambda with unknown operation: Traceback (most recent call last):\n  File \"/opt/code/localstack/localstack/aws/protocol/parser.py\", line 557, in parse\n    operation, uri_params = self._operation_router.match(request)\n                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/code/localstack/localstack/aws/protocol/op_router.py\", line 321, in match\n    rule, args = matcher.match(path, method=method, return_rule=True)\n                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/code/localstack/.venv/lib/python3.11/site-packages/werkzeug/routing/map.py\", line 631, in match\n    raise NotFound() from None\nwerkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.\n\nThe above exception was the direct cause of the following exception:\n\nTraceback (most recent call last):\n  File \"/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py\", line 166, in handle\n    handler(self, self.context, response)\n  File \"/opt/code/localstack/localstack/aws/handlers/service.py\", line 63, in __call__\n    return self.parse_and_enrich(context)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/code/localstack/localstack/aws/handlers/service.py\", line 67, in parse_and_enrich\n    operation, instance = parser.parse(context.request)\n                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/code/localstack/localstack/aws/protocol/parser.py\", line 172, in wrapper\n    return func(*args, **kwargs)\n           ^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/code/localstack/localstack/aws/protocol/parser.py\", line 559, in parse\n    raise OperationNotFoundParserError(\nlocalstack.aws.protocol.parser.OperationNotFoundParserError: Unable to find operation for request to service lambda: GET /\n"}%  
```

In this situation:
* the service is identified as "lambda" because of the hostname matching rule (`.lambda-url.`)
* the service hasn't been loaded yet, so the edge routes for lambda (i.e., the data plane) haven't been added yet either
* therefore, the request moves on to the service request parser, which fails because no operation is associated with the request.

bad and ugly!

instead, localstack now correctly loads the lambda service:

```
curl http://a38wes0ubqxh7ipckjdl8ilptt40uiha.lambda-url.us-east-1.localhost.localstack.cloud:4566
{"Message": null}
```

I had to
* adapt the service loader so it correctly loads services even when there is no api operation set (for the life of me i cannot remember why i thought it would be useful to check individual operations rather than just whole services).
* add a special type of service loader that only checks the hostname rules

## Changes

* A selected set of services are now properly loaded when calling data plane URLs. Specfically, URLs of APIGW (`.execute-api.`), S3 website (`.s3-website.`) (cc @bentsku) and lambda (`.lambda-url.`)
* Calling non-existing lambda function URLs now returns the correct HTTP response

### DIFF
--- a/localstack/aws/app.py
+++ b/localstack/aws/app.py
@@ -3,7 +3,7 @@ from localstack.aws import handlers
 from localstack.aws.api import RequestContext
 from localstack.aws.chain import HandlerChain
 from localstack.aws.handlers.metric_handler import MetricHandler
-from localstack.aws.handlers.service_plugin import ServiceLoader
+from localstack.aws.handlers.service_plugin import ServiceLoader, ServiceLoaderForDataPlane
 from localstack.aws.trace import TracingHandlerChain
 from localstack.services.plugins import SERVICE_PLUGINS, ServiceManager, ServicePluginManager
 from localstack.utils.ssl import create_ssl_cert, install_predefined_cert_if_available
@@ -22,6 +22,7 @@ class LocalstackAwsGateway(Gateway):
         self.service_request_router = ServiceRequestRouter()
         # lazy-loads services into the router
         load_service = ServiceLoader(self.service_manager, self.service_request_router)
+        load_service_for_data_plane = ServiceLoaderForDataPlane(load_service)
 
         metric_collector = MetricHandler()
         # the main request handler chain
@@ -31,6 +32,7 @@ class LocalstackAwsGateway(Gateway):
                 handlers.add_internal_request_params,
                 handlers.handle_runtime_shutdown,
                 metric_collector.create_metric_handler_item,
+                load_service_for_data_plane,
                 handlers.preprocess_request,
                 handlers.parse_service_name,  # enforce_cors and content_decoder depend on the service name
                 handlers.enforce_cors,
--- a/localstack/aws/handlers/service_plugin.py
+++ b/localstack/aws/handlers/service_plugin.py
@@ -40,15 +40,15 @@ class ServiceLoader(Handler):
             return
 
         service_name: str = context.service.service_name
+        if service_name in self.loaded_services:
+            return
+
         if not self.service_manager.exists(service_name):
             raise NotImplementedError
         elif not is_api_enabled(service_name):
             raise NotImplementedError(
                 f"Service '{service_name}' is not enabled. Please check your 'SERVICES' configuration variable."
             )
-
-        if service_name in self.loaded_services:
-            return
 
         request_router = self.service_request_router
 
--- a/localstack/aws/handlers/service_plugin.py
+++ b/localstack/aws/handlers/service_plugin.py
@@ -2,7 +2,6 @@
 
 import logging
 import threading
-from typing import Optional
 
 from localstack.http import Response
 from localstack.services.plugins import Service, ServiceManager
@@ -10,8 +9,8 @@ from localstack.utils.sync import SynchronizedDefaultDict
 
 from ...utils.bootstrap import is_api_enabled
 from ..api import RequestContext
-from ..api.core import ServiceOperation
 from ..chain import Handler, HandlerChain
+from ..protocol.service_router import determine_aws_service_model_for_data_plane
 from .service import ServiceRequestRouter
 
 LOG = logging.getLogger(__name__)
@@ -31,6 +30,7 @@ class ServiceLoader(Handler):
         self.service_manager = service_manager
         self.service_request_router = service_request_router
         self.service_locks = SynchronizedDefaultDict(threading.RLock)
+        self.loaded_services = set()
 
     def __call__(self, chain: HandlerChain, context: RequestContext, response: Response):
         return self.require_service(chain, context, response)
@@ -47,20 +47,19 @@ class ServiceLoader(Handler):
                 f"Service '{service_name}' is not enabled. Please check your 'SERVICES' configuration variable."
             )
 
-        service_operation: Optional[ServiceOperation] = context.service_operation
+        if service_name in self.loaded_services:
+            return
+
         request_router = self.service_request_router
 
         # Ensure the Service is loaded and set to ServiceState.RUNNING if not in an erroneous state.
         service_plugin: Service = self.service_manager.require(service_name)
 
-        # Continue adding service skelethon and handlers to the router if these are missing.
-        if service_operation in request_router.handlers:
-            return
-
         with self.service_locks[context.service.service_name]:
             # try again to avoid race conditions
-            if service_operation in request_router.handlers:
+            if service_name in self.loaded_services:
                 return
+            self.loaded_services.add(service_name)
             if isinstance(service_plugin, Service):
                 request_router.add_skeleton(service_plugin.skeleton)
             else:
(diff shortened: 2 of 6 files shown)


alexrashed (commented):
Thanks for jumping on this! Really great to see the dynamic route based service loading finally being tackled! I think this solution is already great and tackles the vast majority of routes registered for the different services.

I am just a bit concerned because it depends on the service router's host detection capabilities which need to be in sync with the routes registered by the services.
They are in sync most of the time, but this approach would for example not work if a router is service specific but not based on the host (f.e. for path-based API GW invocations).

Maybe it would be better to handle this outside of the handler chain, by decorating / wrapping the endpoints registered in the router?
For example, we could create a subclass of our router which by default wraps the endpoints with a service loader. Whenever the route is invoked, first the service who registered the route could then be ensured to be loaded.

What do you think?

thrau:
Thanks for the feedback Alex! I think this premise here is where I have a different perspective:

> I am just a bit concerned because it depends on the service router's host detection capabilities which need to be in sync with the routes registered by the services.

Conceptually I would argue that anything that's part of the spec should go into the service router. After learning more about Azure, it's clear that they did a much better job at adding the "data plane" aspect into their spec than AWS did, but nonetheless I believe we should be treating the data plane (i.e., `.lambda-url.`, `.s3-website.`, etc.) as first-class citizens of the spec, rather than something optional added by services. Therefore we shouldn't depend on service implementations to provide these routes, just as we don't depend on service implements to provide which API operations are known to the router.

> Maybe it would be better to handle this outside of the handler chain, by decorating / wrapping the endpoints registered in the router?

This would suffer from the bootstrapping problem though right? We're trying load a service based on the routes Routes it provides, but they aren't provided until the service loads. So we'd need an orthogonal plugin to the service plugin that registers these routes. Doable, but not sure that's really better.

That said, I think we should provide your suggestion as an additional feature when services add localstack-specific data plane endpoints (like `http://localhost:4566/restapis/<apiId>/<stageId>/_user_request_`, which is not known to the spec).

So to summarize, I would distinguish for now between things we know a priori as an implicit part of the spec (`.lambda-url.`, ...), which this PR addresses, and those that are unknown and added by localstack like the example above, where we need a solution you suggested.

localstack/aws/handlers/service_plugin.py
```diff
@@ -47,24 +47,42 @@ def require_service(self, _: HandlerChain, context: RequestContext, response: Re
                 f"Service '{service_name}' is not enabled. Please check your 'SERVICES' configuration variable."
             )
 
-        service_operation: Optional[ServiceOperation] = context.service_operation
+        if service_name in self.loaded_services:
+            return
```
alexrashed, line 51: nit: I see that we can save a few operations if we keep the state of loaded services here, but I feel like this should be in the responsibility of the `service _manager`. But also this code is executed on every single request, so it might make sense to avoid the plugin handling logic if possible.

alexrashed (approved):
You are right, we should differentiate between these two use cases. And honestly, I was more focused on dynamically created routes (like the mentioned API GW).
But when thinking about it (and as you pointed out), this doesn't really make sense, because they can only be registered by (a) an operation call which means the service is loaded or (b) a persistence restore which is also done by the service loaded by the persistence framework.
So the use case I was focused on is actually already partially covered, and your solution perfectly fixes the other part. 💯 
The implementation is looking good, I just added a nitpicky comment if it really is worth keeping the state of the loaded services in the handler in addition to the service manager. 🤔

localstack/aws/protocol/service_router.py
```diff
@@ -138,14 +138,19 @@ def custom_signing_name_rules(signing_name: str, path: str) -> Optional[ServiceM
 
 def custom_host_addressing_rules(host: str) -> Optional[ServiceModelIdentifier]:
```
joe4dev, line 139: Probably a super edge case to ignore but I guess if someone uses a custom domain name containing one of these strings, we might route it to the wrong service 💥

localstack/aws/app.py
```diff
@@ -31,6 +32,7 @@ def __init__(self, service_manager: ServiceManager = None) -> None:
                 handlers.add_internal_request_params,
                 handlers.handle_runtime_shutdown,
                 metric_collector.create_metric_handler_item,
+                load_service_for_data_plane,
```
joe4dev, line 35: more work for the handler chain 🥲

joe4dev, line 35: @dfangl handling that case would likely need custom logic for the distributed LocalStack scenario because service detection happens in the proxy, right?

joe4dev (approved):
Great catch 👏  LGTM from the Lambda side 👍

I tried to think about edge cases but only came up with very exotic scenarios (we can likely ignore them).

localstack/aws/protocol/service_router.py
thrau: agree. this is actually a problem we have in lots of places where we do fuzzy url matching. :shrug: so far we haven't had anyone running into that issue though

localstack/aws/app.py
thrau: this is a very cheap operation and it's not doing duplicate work. if this already determines the service, then the service loader does no extra work.

thrau: that's a great point actually! curious to get daniel's perspective

dfangl: Sorry for the late reply, I actually had one here, but forgot to submit the review, it still shows as pending :(
You can route based on the `lambda-url` in the domain, so it should work without a problem.

PR #10646, opened 2024-04-12
Fix dev run container exiting from raw mode

simonrw:
## Motivation

Running a dev container with `python -m localstack.dev.run` does not return the terminal to "cooked" mode correctly. This means the user has to press Enter multiple times to get their prompt back.

## Changes

Swap out `run_interactive` for `subprocess.check_call` since I think all functionality implemented in `run_interactive` is performed by `check_call`. TBC

### DIFF
--- a/localstack/utils/run.py
+++ b/localstack/utils/run.py
@@ -158,42 +158,7 @@ def run_interactive(command: List[str]):
 
     :param command: the command to pass to subprocess.Popen
     """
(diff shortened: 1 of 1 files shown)


thrau (approved):
i'm surprised to see this working so well. cool thank you! no idea why i thought we couldn't use check_output at the time.

:+1: for simple solutions

PR #10649, opened 2024-04-12
Validate existing test acm

cloutierMat:
## Motivation

Part of the effort to improve parity with aws. This first step sorts all `aws.unknown` tags and replaces them with `aws.validated` or `aws.needs_fixing`. 

## Changes

Removed all `unknown` markers from acm.

There is no comment on the `test_boto_wait_for_certificate_validation` as the TODO is already well captured by the comment preceding the `acm_request_certificate` fixtures

### DIFF
--- a/tests/aws/services/acm/test_acm.py
+++ b/tests/aws/services/acm/test_acm.py
@@ -61,12 +61,22 @@ class TestACM:
 
         snapshot.match("describe-certificate-response", describe_res)
 
-    @markers.aws.unknown
-    def test_domain_validation(self, acm_request_certificate, aws_client):
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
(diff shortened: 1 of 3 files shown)


steffyP (commented):
Thanks for addressing the first couple of aws-markers 🎉 
Nice to see that we could even add a snapshot for one of the tests 🚀 

The changes LGTM, I only wonder what is causing the `collection_failure` for a couple of tests in the CI run. 🤔

alexrashed:
I tried looking into the issue with the pipeline and started with a rebase, where the commits disappeared.
It seems that the changes here are already contained in https://github.com/localstack/localstack/pull/10653 and this PR is obsolete?

cloutierMat:
Good catch Alex. I must have branched off the wrong branch.

PR #10652, opened 2024-04-12
Adapt source distribution glob to new PEP 625 canonicalized name

dfangl:
## Motivation
With the recently released setuptools version `69.3.0` (https://setuptools.pypa.io/en/latest/history.html#v69-3-0), setuptools normalizes the source distribution name to conform with [PEP 625](https://peps.python.org/pep-0625/).

In summary, source distribution names should be named as followed: `{distribution}-{version}.tar.gz`.
Only a single hypen is now allowed:
> Conforming sdist files can be recognised by the presence of the .tar.gz suffix and a single hyphen in the filename.

Since our package was named `localstack-core-{version}.tar.gz`, it is now normalized to the above format with `localstack_core-{version}.tar.gz`.

In our Makefile we look for the source distribution when checking the source distribution for valid entrypoints using a glob, which has to be adapted.

## Changes
* Adapt glob for checking the to-be-uploaded the localstack-core source distribution to new normalized name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ dist: entrypoints        ## Build source and built (wheel) distributions of the 
 
 publish: clean-dist dist  ## Publish the library to the central PyPi repository
 	# make sure the dist archive contains a non-empty entry_points.txt file before uploading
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks so much for jumping on this to unblock `master` after the setuptools upgrade! 🦸🏽

PR #10653, opened 2024-04-12
Validate existing test apigw

cloutierMat:
## Motivation

Part of the effort to improve parity with aws. This first step sorts all `aws.unknown` tags and replaces them with `aws.validated` or `aws.needs_fixing`.

## Changes

Started to go through apigw

### DIFF
--- a/localstack/testing/pytest/fixtures.py
+++ b/localstack/testing/pytest/fixtures.py
@@ -1940,7 +1940,15 @@ def create_rest_apigw(aws_client_factory):
 
     def _create_apigateway_function(**kwargs):
         region_name = kwargs.pop("region_name", None)
-        apigateway_client = aws_client_factory(region_name=region_name).apigateway
+        client_config = None
+        if is_aws_cloud():
+            client_config = botocore.config.Config(
+                # Api gateway can throttle requests pretty heavily. Leading to potentially undeleted apis
+                retries={"max_attempts": 10, "mode": "adaptive"}
+            )
+        apigateway_client = aws_client_factory(
+            region_name=region_name, config=client_config
+        ).apigateway
         kwargs.setdefault("name", f"api-{short_uid()}")
 
         response = apigateway_client.create_rest_api(**kwargs)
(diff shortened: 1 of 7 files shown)


tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -157,7 +157,8 @@ def test_delete_rest_api_with_invalid_id(self, aws_client):
         assert "foobar" in e.value.response["Error"]["Message"]
 
     @pytest.mark.parametrize("url_function", [path_based_url, host_based_url])
-    @markers.aws.unknown
+    @markers.aws.needs_fixing
+    # This is not a possible feature on aws. Is there a reason for us to add that support?
```
bentsku, line 161: This is a specific feature so that people can always have the same API ID when recreating their state. I think it avoids the whole custom domain. I think it has some use case, if you're developing something quickly and restart LocalStack to redeploy some code, you don't need to redeploy your own code with the new API ID as well. 
```suggestion
    @markers.aws.only_localstack
    # This is not a possible feature on aws.
```

localstack/testing/pytest/fixtures.py
```diff
@@ -1940,7 +1940,15 @@ def create_rest_apigw(aws_client_factory):
...
-        apigateway_client = aws_client_factory(region_name=region_name).apigateway
+        client_config = None
+        if is_aws_cloud():
+            client_config = botocore.config.Config(
+                # Api gateway can throttle requests pretty heavily. Leading to potentially undeleted apis
+                retries={"max_attempts": 10, "mode": "adaptive"}
+            )
+        apigateway_client = aws_client_factory(
+            region_name=region_name, config=client_config
+        ).apigateway
```
bentsku, line 1951: Ah, this is a great way to fix it. We also have some other fixtures which just try deleting forever or almost if getting a 429, the throttling of `DeleteRestApi` is really insane to me, you'd need to add 15+ seconds sleep in between calls 😭 neat 👌

bentsku (approved):
LGTM! Nice work, and thanks for the comments, really helpful in understanding why there's an issue. 🙇‍♂️

tests/aws/services/apigateway/test_apigateway_basic.py
cloutierMat: oops! I had a last commit on my machine changing this.

cloutierMat:
@bentsku I added a lost commit forgotten on my machine! It contained the update you mentioned for `only_localstack` and a quick swap of markers from one test to the other. (I had gotten confused in the navigation of the fixture calling fixture of that class)

PR #10656, opened 2024-04-13
Use single character wildcard when matching the source distribution to avoid breaking with yanked setuptools

dfangl:
## Motivation
The setuptools upgrade which necessitated #10652 was yanked due to issues with the name normalization, but it is expected that the change will be re-released soon.
To avoid waiting for that release, and to avoid reverting now and reintroducing the changes when it happens, we should make the pattern more resilient.

## Changes
* Use a `?` (single character wildcard) instead of `-` or `_` in the publish operation, to always match the name, no matter the setuptools version.


alexrashed (approved):
You're a real 🦸🏽!
Thanks for dealing with this `setuptools` weirdness! This looks like the most resilient solution when thinking about future reverts of the revert! 😜🚀

Issue #10648, opened 2024-04-12
bug: Step Functions: heartbeat tokens become invalid when several executions run concurrently


mbaynton:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If I
 * start several (I'm usually running about 6 at once) executions of the same state machine
 * and that state machine has a step that sends to localstack SQS with `waitForTaskToken` and also utilizes `HeartbeatSeconds`
 * and my SQS consumers start sending heartbeat callbacks with their corresponding step function tokens

Then at first, all consumers' `SendTaskHeartbeat` calls work, but eventually at least one of the running `SendTaskHeartbeat` calls starts failing with this stringified `error`:
> {error 26 0  operation error SFN: SendTaskHeartbeat, https response error StatusCode: 400, RequestID: c95077cf-4251-464f-aa24-abdf9b7753d6, InvalidToken: }

Subsequent SendTaskHeartbeat calls continue to fail in the same manner until the state machine step times out on account of not having received any hearbeats for the defined `HeartbeatSeconds` period.

Some of the other concurrent SFN executions complete and others start during the time the longer one starts experiencing this issue, not sure if that's the cause.

### Expected Behavior

`SendTaskHeartbeat` calls continue to be accepted until I call back to SFN with a success or failure.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

By `docker compose run`ning a service with a defined dependency on `localstack`. docker-compose for the localstack service below.

```
version: '3.0'
services:
  localstack:
    image: localstack/localstack:3.3.0
    environment:
      - SERVICES=dynamodb,s3,sqs,stepfunctions
      - DATA_DIR=/var/lib/localstack/data
    ports:
      - 4566:4566
      - 4571:4571
    volumes:
      - ./scripts/localstack-startup.sh:/docker-entrypoint-initaws.d/localstack-startup.sh:cached
      - ./tmp/localstack:/var/lib/localstack
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

Probably requires more than a few commands, sorry.
We're experiencing this in a private codebase, I would have to write a reproduce from scratch.

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack: 3.3.0
```

### Anything else?

Fixed by using `PROVIDER_OVERRIDE_STEPFUNCTIONS=legacy`

[...]

PR #10663, opened 2024-04-15
[SFN] Fix Heartbeat Callback Locking

MEPalma:
## Motivation
Currently the SFN interpreter is unable to always correctly synchronise with heartbeats notifications as the workflow clears the heartbeat event before waiting for it to be set. This may especially be an issue if before starting a new heartbeat wait period, a success of failure event was submitted to the task. This pr addresses such issue by adding a mutex lock for event set and check. It also adds two related snapshot tests. `test_multiple_heartbeat_notifications` is a base case in which multiple heartbeat notifications are sent before a task success notification. `test_multiple_executions_and_heartbeat_notifications` extends the previous test case by launching multiple parallel executions. Addresses #10648.

## Changes
- added locking mechanism in the Heartbeat Endpoint class
- added relevant snapshot tests

### DIFF
--- a/tests/aws/services/stepfunctions/v2/error_handling/test_task_service_sqs.py
+++ b/tests/aws/services/stepfunctions/v2/error_handling/test_task_service_sqs.py
@@ -126,10 +126,10 @@ class TestTaskServiceSqs:
         create_state_machine,
         sqs_create_queue,
         sqs_send_task_failure_state_machine,
-        snapshot,
+        sfn_snapshot,
     ):
-        snapshot.add_transformer(snapshot.transform.sqs_api())
-        snapshot.add_transformer(
+        sfn_snapshot.add_transformer(sfn_snapshot.transform.sfn_sqs_integration())
+        sfn_snapshot.add_transformer(
             JsonpathTransformer(
                 jsonpath="$..TaskToken",
                 replacement="task_token",
@@ -139,8 +139,8 @@ class TestTaskServiceSqs:
 
         queue_name = f"queue-{short_uid()}"
         queue_url = sqs_create_queue(QueueName=queue_name)
-        snapshot.add_transformer(RegexTransformer(queue_url, "<sqs_queue_url>"))
-        snapshot.add_transformer(RegexTransformer(queue_name, "<sqs_queue_name>"))
+        sfn_snapshot.add_transformer(RegexTransformer(queue_url, "<sqs_queue_url>"))
+        sfn_snapshot.add_transformer(RegexTransformer(queue_name, "<sqs_queue_name>"))
 
         sqs_send_task_failure_state_machine(queue_url)
 
@@ -154,7 +154,7 @@ class TestTaskServiceSqs:
             aws_client.stepfunctions,
             create_iam_role_for_sfn,
             create_state_machine,
-            snapshot,
+            sfn_snapshot,
             definition,
             exec_input,
         )


tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -471,3 +511,141 @@ def test_start_execution_sync_delegate_timeout(
...
+            )
+            sfn_snapshot.match(
+                f"task_token_inner_handler_{inner_handler_id}_responses",
+                task_token_handler_responses,
+            )
+            task_token_handler_latch.count_down()
+
+        for i in range(execution_count):
+            inner_handler_thread = threading.Thread(target=_sqs_task_token_handler, args=(i,))
+            inner_handler_thread.start()
```
joe4dev, line 639: good to know 💡 : if you ever need a scenario where multiple threads should be ready and then start at the same time, a barrier could be useful. Example in Lambda perftests: https://github.com/localstack/localstack/blob/0f4b52eefee5b9eca65c2050cd69ec423e3eea9f/tests/aws/services/lambda_/test_lambda_performance.py#L302

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -471,3 +511,141 @@ def test_start_execution_sync_delegate_timeout(
             definition,
             exec_input,
         )
+
+    @markers.snapshot.skip_snapshot_verify(paths=["$..MD5OfMessageBody"])
+    @markers.aws.needs_fixing
```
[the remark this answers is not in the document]
joe4dev: I tried against AWS and validation failed. Does this mean the test case is not deterministic?

```python
tests/aws/services/stepfunctions/v2/callback/test_callback.py:515 (TestCallback.test_multiple_heartbeat_notifications)
>> match key: get_execution_history
	(+) /events/[6]/executionFailedEventDetails ( {'error': 'States.Timeout'} )
	(~) /events/[4]/taskSubmittedEventDetails/output/MD5OfMessageBody '77b6f94944d0d189f49858e0c286b1f9' → '77e38bc5c824da0660ea6d2146cefd92' ... (expected → actual)
	(+) /events/[5]/taskTimedOutEventDetails ( {'error': 'States.Timeout', 'resource': 'sendMessage.waitForTaskToken', 'resourceType': 'sqs'} )
	(~) /events/[5]/type 'TaskSucceeded' → 'TaskTimedOut' ... (expected → actual)
	(-) /events/[6]/stateExitedEventDetails ( {'name': 'SendMessageWithWait', 'output': {'Message': 'txt', 'TaskToken': '<task_token:1>'}, 'outputDetails': {'truncated': False}} )
	(-) /events/[5]/taskSucceededEventDetails ( {'output': {'Message': 'txt', 'TaskToken': '<task_token:1>'}, 'outputDetails': {'truncated': False}, 'resource': 'sendMessage.waitForTaskToken', 'resourceType': 'sqs'} )
	(~) /events/[6]/type 'TaskStateExited' → 'ExecutionFailed' ... (expected → actual)
	(-) /events/[7] ( {'executionSucceededEventDetails': {'output': {'Message': 'txt', 'TaskToken': '<task_token:1>'}, 'outputDetails': {'truncated': False}}, 'id': 8, 'previousEventId': 7, 'timestamp': 'timestamp', 'type': 'ExecutionSucceeded'} )
	(+) /events/[7] ( not present )

	["$..events", "$..events..executionFailedEventDetails", "$..events..stateExitedEventDetails", "$..events..taskSubmittedEventDetails.output.MD5OfMessageBody", "$..events..taskSucceededEventDetails", "$..events..taskTimedOutEventDetails", "$..events..type"]
/Users/joe/Projects/LocalStack/localstack-ext/.venv/lib/python3.11/site-packages/_pytest/threadexception.py:77: PytestUnhandledThreadExceptionWarning: Exception in thread Thread-1 (_handle_sqs_task_token)

  File "/Users/joe/.pyenv/versions/3.11.7/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    self.run()
  File "/Users/joe/.pyenv/versions/3.11.7/lib/python3.11/threading.py", line 982, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/joe/Projects/LocalStack/localstack/tests/aws/services/stepfunctions/v2/callback/test_callback.py", line 540, in _handle_sqs_task_token
    task_token_handler_responses = _handle_sqs_task_token_with_heartbeats_and_success(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joe/Projects/LocalStack/localstack/tests/aws/services/stepfunctions/v2/callback/test_callback.py", line 46, in _handle_sqs_task_token_with_heartbeats_and_success
    send_task_heartbeat_response = aws_client.stepfunctions.send_task_heartbeat(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joe/Projects/LocalStack/localstack-ext/.venv/lib/python3.11/site-packages/botocore/client.py", line 565, in _api_call
  File "/Users/joe/Projects/LocalStack/localstack-ext/.venv/lib/python3.11/site-packages/botocore/client.py", line 1021, in _make_api_call
botocore.errorfactory.TaskTimedOut: An error occurred (TaskTimedOut) when calling the SendTaskHeartbeat operation: Task Timed Out: 'Provided task does not exist anymore'

  warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

tests/aws/services/stepfunctions/v2/callback/test_callback.py
MEPalma: Thank you! -- In this case we don't need for all threads to wait on a barrier as they interface with separate state machines

tests/aws/services/stepfunctions/v2/callback/test_callback.snapshot.json
```diff
@@ -1507,5 +1507,1674 @@
...
+            },
+            "timestamp": "timestamp",
+            "type": "TaskStarted"
+          },
+          {
+            "id": 5,
+            "previousEventId": 4,
+            "taskSubmittedEventDetails": {
+              "output": {
+                "MD5OfMessageBody": "77b6f94944d0d189f49858e0c286b1f9",
```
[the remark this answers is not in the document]
MEPalma: That's correct, at the moment the sqs message body outputs a different MD5 for LocalStack

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -471,3 +511,141 @@ def test_start_execution_sync_delegate_timeout(
...
+            create_state_machine,
+            sfn_snapshot,
+            definition,
+            exec_input,
+        )
+
+        task_token_consumer_thread.join(timeout=180)
+
+    @markers.snapshot.skip_snapshot_verify(paths=["$..MD5OfMessageBody"])
+    @markers.aws.needs_fixing
```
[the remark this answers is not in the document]
MEPalma: Yes, sorry, this was fixed, missed updating the marker

tests/aws/services/stepfunctions/v2/callback/test_callback.py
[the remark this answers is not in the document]
MEPalma: Yes I missed updating the marker to validated. I am surprised however the validation did not succeed for you, as I cannot replicate this. I will increase the maximum timeouts to reduce the likelihood of these errors.

tests/aws/services/stepfunctions/v2/callback/test_callback.snapshot.json
```diff
@@ -1507,5 +1507,1674 @@
...
+            "HTTPStatusCode": 200
+          }
+        },
+        "send_task_heartbeat_response_8": {
+          "ResponseMetadata": {
+            "HTTPHeaders": {},
+            "HTTPStatusCode": 200
+          }
+        },
+        "send_task_heartbeat_response_9": {
```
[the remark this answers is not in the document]
MEPalma: I am removing these calls as if they succeed they are empty by spec, otherwise they raise an exception and terminate the thread

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -513,7 +501,7 @@ def test_start_execution_sync_delegate_timeout(
         )
 
     @markers.snapshot.skip_snapshot_verify(paths=["$..MD5OfMessageBody"])
```
joe4dev, line 503: Validation against AWS works now ✅ 

However, I am getting different checksums when validating against AWS (the same for the other test):
Screenshot 2024-04-17 at 12 31 31

Are you sure that the checksum is deterministic on AWS? Is it the checksum of a static body or does it include some dynamic fields (e.g., datetime)?

If the checksum is not deterministic against AWS, we should change it into a transformer.

joe4dev (approved):
LGTM 👍 Thank you for refining the tests. 👏👏👏 

Please (re-)verify the checksum behavior and potentially transform the skip into a transformer before merging.

tests/aws/services/stepfunctions/v2/callback/test_callback.py
MEPalma: You have noticed this discrepancy in snapshot tests for time now, however have never made it up the priority to be addressed yet. Here stepFunctions is merely forwarding the SQS reply to the rest of the program. I believe more focus should go in verifying the behaviour of the SQS provider, and secondly, like you say, verify if the message body should be recomputed by stepfunctions and account for non deterministic behaviours. 

I am adding an item in the backlog to investigate this aspect

joe4dev: Maybe it's a misunderstanding. My comment is about the **test case** and not about LocalStack.
If the checksum changes every time we re-create the snapshot, it should be transformed (e.g., that's the case if the checksum depends on some dynamic field such as datetime). If AWS always produces the same deterministic snapshot for the checksum (e.g., the checksum depends on some static field and NEVER changes), then the current approach is fine.

Hence, if we create the snapshot (`SNAPSHOT_UPDATE=1`) and re-run the test against AWS without updating the snapshot (`SNAPSHOT_UPDATE=0`), there should be no difference. I'm getting a different checksum for multiple runs against AWS. Can you reproduce this?

MEPalma: Thank you Joel, I understand your point now (btw sorry there was a typo in my previous message). I will definitely look into this!

MEPalma: After some investigations, it appears as though in StepFunctions, the MD5OfMessageBody is not a direct representation of the message body only. So this changes for every run.
I have added an sqs transformer utility for stepfunctions which treats MD5OfMessageBody, and I have updated all of the snapshots affected.

joe4dev: great 👏 Thank you for addressing this 👍 


Now, we have a clear separation between 🚀 :
* snapshot_skip => LocalStack parity issue
* transformer => dynamic behavior, also in AWS

PR #10664, opened 2024-04-15
Feat: Eventbridge v2:  Add pattern matching

maxhoheiser:
## Motivation
This PR adds pattern matching to distribute incoming events based on event bus specified in the event, rules in the matching event bus and matches the patterns in the rules. 
It uses the existing rule matching algorithm from v1, which is limited, e.g. patterns with `anything but` or `exists = True`. 

## Changes
Add parsing for event pattern to a dictionary, and add validating of pattern.
Add correct formatting of events and add additional keys. 
Add API method to test event pattern using same matching algorithm as for distributing events to targets.
Remove skip markers from now passing tests.

## Testing
Set the env variable `PROVIDER_OVERRIDE_EVENTS=v2` and run tests in events folder. All unsupported tests have the appropriate skip markers.

### DIFF
--- a/localstack/services/events/event_ruler.py
+++ b/localstack/services/events/event_ruler.py
@@ -3,7 +3,6 @@ import os
 from functools import cache
 from pathlib import Path
 
-from localstack import config
 from localstack.services.events.packages import event_ruler_package
 from localstack.services.events.utils import InvalidEventPatternException
 from localstack.utils.objects import singleton_factory
@@ -43,8 +42,6 @@ def matches_rule(event: str, rule: str) -> bool:
     There is a single static boolean method Ruler.matchesRule(event, rule) -
     both arguments are provided as JSON strings.
     """
-    if config.EVENT_RULE_ENGINE != "java":
-        raise NotImplementedError("Set EVENT_RULE_ENGINE=java to enable the Java Event Ruler.")
 
     start_jvm()
     import jpype.imports  # noqa F401: required for importing Java modules
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -94,11 +94,3 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
-
-
-class InternalInvalidEventPatternException(Exception):
-    reason: str
-
-    def __init__(self, reason=None, message=None) -> None:
-        self.reason = reason
-        self.message = message or f"Event pattern is not valid. Reason: {reason}"
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Any, Optional
 
 from localstack.aws.api.core import ServiceException
 from localstack.aws.api.events import (
@@ -58,6 +58,8 @@ class Rule:
 
 
 RuleDict = dict[RuleName, Rule]
+
+EventPatternDict = dict[str, Any]
 
 
 @dataclass
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Any, Optional
+from typing import Optional
 
 from localstack.aws.api.core import ServiceException
 from localstack.aws.api.events import (
@@ -58,8 +58,6 @@ class Rule:
 
 
 RuleDict = dict[RuleName, Rule]
-
-EventPatternDict = dict[str, Any]
 
 
 @dataclass
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -94,3 +94,9 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
+
+
+class InvalidEventPatternException(ServiceException):
+    code: str = "InvalidEventPatternException"
+    sender_fault: bool = True
+    status_code: int = 400
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -96,3 +96,11 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
+
+
+class InternalInvalidEventPatternException(Exception):
+    reason: str
+
+    def __init__(self, reason=None, message=None) -> None:
+        self.reason = reason
+        self.message = message or f"Event pattern is not valid. Reason: {reason}"
--- a/localstack/services/events/models_v2.py
+++ b/localstack/services/events/models_v2.py
@@ -96,9 +96,3 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
-
-
-class InvalidEventPatternException(ServiceException):
-    code: str = "InvalidEventPatternException"
-    sender_fault: bool = True
-    status_code: int = 400
--- a/localstack/services/events/provider_v2.py
+++ b/localstack/services/events/provider_v2.py
@@ -4,7 +4,6 @@ import logging
 from datetime import datetime, timezone
 from typing import Optional
 
-from localstack import config
 from localstack.aws.api import RequestContext, handler
 from localstack.aws.api.events import (
     Arn,
@@ -29,6 +28,8 @@ from localstack.aws.api.events import (
     PutEventsRequestEntry,
     PutEventsRequestEntryList,
     PutEventsResponse,
+    PutEventsResultEntry,
+    PutEventsResultEntryList,
     PutPartnerEventsRequestEntryList,
     PutPartnerEventsResponse,
     PutRuleResponse,
@@ -67,7 +68,6 @@ from localstack.services.events.models_v2 import (
 )
 from localstack.services.events.rule import RuleService, RuleServiceDict
 from localstack.services.events.target import TargetSender, TargetSenderDict, TargetSenderFactory
-from localstack.services.events.utils import matches_event
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.strings import long_uid
 
@@ -103,6 +103,24 @@ def get_event_time(event: PutEventsRequestEntry) -> str:
             )
     formatted_time_string = event_time.strftime("%Y-%m-%dT%H:%M:%SZ")
     return formatted_time_string
+
+
+def validate_event(event: PutEventsRequestEntry) -> None | PutEventsResultEntry:
+    if not event.get("Source"):
+        return {
+            "ErrorCode": "InvalidArgument",
+            "ErrorMessage": "Parameter Source is not valid. Reason: Source is a required argument.",
+        }
+    elif not event.get("DetailType"):
+        return {
+            "ErrorCode": "InvalidArgument",
+            "ErrorMessage": "Parameter DetailType is not valid. Reason: DetailType is a required argument.",
+        }
+    elif not event.get("Detail"):
+        return {
+            "ErrorCode": "InvalidArgument",
+            "ErrorMessage": "Parameter Detail is not valid. Reason: Detail is a required argument.",
+        }
 
 
 def format_event(event: PutEventsRequestEntry, region: str, account_id: str) -> dict:
@@ -441,14 +459,12 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         endpoint_id: EndpointId = None,
         **kwargs,
     ) -> PutEventsResponse:
-        processed_entries_uuids, failed_entries = self._process_entries(context, entries)
+        entries, failed_entry_count = self._process_entries(context, entries)
 
         response = PutEventsResponse(
-            Entries=[{"EventId": id} for id in processed_entries_uuids],
-            FailedEntryCount=len(failed_entries),
+            Entries=entries,
+            FailedEntryCount=failed_entry_count,
         )
-        if failed_entries:
-            response["FailedEntries"] = failed_entries
         return response
 
     @handler("PutPartnerEvents")
@@ -642,18 +658,22 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
 
     def _process_entries(
         self, context: RequestContext, entries: PutEventsRequestEntryList
-    ) -> tuple[list, list]:
-        processed_entries_uuids = []
-        failed_entries = []
+    ) -> tuple[PutEventsResultEntryList, int]:
+        processed_entries = []
+        failed_entry_count = 0
         for event in entries:
             event_bus_name = event.get("EventBusName", "default")
+            if event_failed_validation := validate_event(event):
+                processed_entries.append(event_failed_validation)
+                failed_entry_count += 1
+                continue
             event = format_event(event, context.region, context.account_id)
-            processed_entries_uuids.append(event["id"])
             store = self.get_store(context)
             try:
                 event_bus = self.get_event_bus(event_bus_name, store)
             except ResourceNotFoundException:
-                # ignore events for non-existing event buses
+                # ignore events for non-existing event buses but add processed event
+                processed_entries.append({"EventId": event["id"]})
                 continue
             matching_rules = [rule for rule in event_bus.rules.values()]
             for rule in matching_rules:
@@ -664,12 +684,13 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
                         target_sender = self._target_sender_store[target["Arn"]]
                         try:
                             target_sender.send_event(event)
+                            processed_entries.append({"EventId": event["id"]})
                         except Exception as error:
-                            failed_entries.append(
+                            processed_entries.append(
                                 {
-                                    "Entry": event,
                                     "ErrorCode": "InternalException",
                                     "ErrorMessage": str(error),
                                 }
                             )
-        return processed_entries_uuids, failed_entries
+                            failed_entry_count += 1
+        return processed_entries, failed_entry_count
--- a/localstack/services/events/provider_v2.py
+++ b/localstack/services/events/provider_v2.py
@@ -4,6 +4,7 @@ import logging
 from datetime import datetime, timezone
 from typing import Optional
 
+from localstack import config
 from localstack.aws.api import RequestContext, handler
 from localstack.aws.api.events import (
     Arn,
@@ -18,6 +19,7 @@ from localstack.aws.api.events import (
     EventPattern,
     EventsApi,
     EventSourceName,
+    InvalidEventPatternException,
     LimitMax100,
     ListEventBusesResponse,
     ListRuleNamesByTargetResponse,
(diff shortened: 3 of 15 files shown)


localstack/services/events/provider_v2.py
```diff
@@ -58,18 +319,358 @@ def put_rule(
...
+            ids = [ids]
+        for target_id in ids:
+            if target := rule.targets.get(target_id):
+                target_arn = target["Arn"]
+                try:
+                    del self._target_workers[target_arn]
+                except KeyError:
+                    LOG.error(f"Error deleting target worker {target_arn}.")
+
+    def _put_entries(
```
maxhoheiser, line 644: Rename to a more meaningful function name, also include docstring

localstack/services/events/target.py
```diff
@@ -17,12 +21,43 @@
...
+        LOG.error("%s key does not exist in input_transformer.", e)
+        raise e
+    for key, path in input_paths.items():
+        value = extract_jsonpath(event, path)
+        if not value:
+            value = ""
+        input_template = input_template.replace(f"<{key}>", value)
+    event = re.sub('"', "", input_template)
+    return event
+
```
maxhoheiser, line 59: input path and input transformers are not purely in scope, but helpful to get certain tests passed that validate event pattern matching, due to the small change I felt it is valid to add it in this PR

localstack/services/events/provider_v2.py
```diff
@@ -578,25 +635,36 @@ def _delete_target_sender(self, ids: TargetIdList, rule) -> None:
...
         for event in entries:
             event_bus_name = event.get("EventBusName", "default")
+            event = format_event(event, context.region, context.account_id)
+            entries_uuids.append(event["id"])
             store = self.get_store(context)
-            event_bus = self.get_event_bus(event_bus_name, store)
-            # TODO add pattern matching
+            try:
+                event_bus = self.get_event_bus(event_bus_name, store)
+            except ResourceNotFoundException:
```
maxhoheiser, line 650: this is validated AWS behavior - see tests `test_put_events_nonexistent_event_bus`

localstack/services/events/rule.py
```diff
@@ -155,6 +163,19 @@ def validate_targets_input(self, targets: TargetList) -> PutTargetsResultEntryLi
 
         return validation_errors
 
+    def load_event_pattern(self, raw_pattern: Optional[str]) -> EventPatternDict:
+        """Loads and validates an event pattern from a JSON string."""
+        if raw_pattern is None:
```
maxhoheiser, line 168: necessary if schedule expression is used and event pattern is empty

maxhoheiser, line 168: necessary for schedule expression if event pattern is empty

localstack/services/events/provider_v2.py
```diff
@@ -79,6 +85,39 @@ def get_filtered_dict(name_prefix: str, input_dict: dict) -> dict:
     return {name: value for name, value in input_dict.items() if name.startswith(name_prefix)}
 
 
+def get_event_time(event: PutEventsRequestEntry) -> str:
```
joe4dev, line 88: style: the `event` is not needed for a time conversion. Passing minimal knowledge is considered best practice. Logging could still happen at a different place.

localstack/services/events/provider_v2.py
joe4dev: maybe dispatch_events, deliver_events, ...?

The API is called put_entries but that's from a user's perspective. Within the EventBridge service, using active terminology such as dispatch makes more sense. Just reading over the AWS docs and drafting a docstring might help decide upon a better name.

joe4dev (changes requested):
Looks good overall 👏👏

There are a couple of things we should address before merging:
* Validate the AWS tests with the `unknown` marker
* Adopt or at least add opt-in for the new Event Ruler https://github.com/localstack/localstack/pull/10615
* Add some test coverage for validating event patterns to avoid rejecting valid patterns. If not needed for v1 parity, that would be a candidate to implement in the next stage focusing on AWS parity.

See comments for other minor suggestions.


Thank you very much for splitting this PR into more manageable chunks 🙏 👍👍

localstack/services/events/provider_v2.py
maxhoheiser: renamed in: https://github.com/localstack/localstack/pull/10664/commits/da0f12a0405588b61d29283965634af5c3d8a66b

localstack/services/events/models_v2.py
```diff
@@ -94,3 +96,9 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
+
+
+class InvalidEventPatternException(ServiceException):
```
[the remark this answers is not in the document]
maxhoheiser: it is not, I did overlook that there is an existing definition in the API stubs, fixed in: https://github.com/localstack/localstack/pull/10664/commits/cb628bc59754a96ddc57498ae0455f4f76fdb903

localstack/services/events/provider_v2.py
```diff
@@ -303,6 +342,19 @@ def put_rule(
...
+    @handler("TestEventPattern")
+    def test_event_pattern(
+        self, context: RequestContext, event_pattern: EventPattern, event: str, **kwargs
+    ) -> TestEventPatternResponse:
+        """Test event pattern uses EventBridge event pattern matching:
+        https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
+        """
+        event_pattern_dict = json.loads(event_pattern)
+        event_dict = json.loads(event)
+        result = matches_event(event_pattern_dict, event_dict)
```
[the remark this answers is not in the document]
maxhoheiser: Is it decided that using the Java event rule is the only implementation in the future? I have mixed feelings of adding Java code. I added the opt-in to be flexible to use native Python matching in the future and build on that or fully switch to the java implementation. Added in: https://github.com/localstack/localstack/pull/10664/commits/d643828de8a3f659d5cfc2ff901b0e908ea4388b

localstack/services/events/provider_v2.py
```diff
@@ -384,9 +436,14 @@ def put_events(
...
     ) -> PutEventsResponse:
-        failed_entries = self._put_entries(context, entries)
+        entries_uuids, failed_entries = self._put_entries(context, entries)
 
-        response = PutEventsResponse(FailedEntryCount=len(failed_entries), Entries=failed_entries)
+        response = PutEventsResponse(
+            Entries=[{"EventId": id} for id in entries_uuids], FailedEntryCount=0
+        )
+        if failed_entries:
+            response["FailedEntryCount"] = len(failed_entries)
```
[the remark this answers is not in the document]
maxhoheiser: Fixed in: https://github.com/localstack/localstack/pull/10664/commits/f9a493f10f379b1119a3751015fafe5c9acd396f

localstack/services/events/provider_v2.py
```diff
@@ -79,6 +85,39 @@ def get_filtered_dict(name_prefix: str, input_dict: dict) -> dict:
...
+def get_event_time(event: PutEventsRequestEntry) -> str:
+    event_time = datetime.now(timezone.utc)
+    if event_timestamp := event.get("Time"):
+        try:
+            # use time from event if provided
+            event_time = event_timestamp.replace(tzinfo=timezone.utc)
+        except ValueError:
+            # use current time if event time is invalid
+            LOG.debug(
+                "Could not parse the `Time` parameter, falling back to current time for the following Event: '%s'",
```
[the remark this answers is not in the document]
maxhoheiser: this is from this fix in this pr: https://github.com/localstack/localstack/pull/8690
I cannot reproduce the bug since boto validates input and throws an error if it is not a string that can be parsed to a valid datetime object. I would keep it in for now, since it is mentioned that it has been an issue.

localstack/services/events/rule.py
```diff
@@ -182,5 +203,20 @@ def _check_target_limit_reached(self) -> bool:
             return True
         return False
 
+    def _validate_event_pattern(self, pattern):
```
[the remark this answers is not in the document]
maxhoheiser: These checks are used in our v1 implementation since they are currently used by moto, I would keep them in since our goal is parity with v1.
As mentioned by you I would add a more sophisticated checking engine in a later stage.

tests/aws/services/events/test_events_rules.py
```diff
@@ -339,8 +335,9 @@ def test_put_event_with_content_base_rule_in_pattern(aws_client, clean_up):
 
 
 @markers.aws.validated
-@pytest.mark.parametrize("schedule_expression", ["rate(1 minute)", "rate(1 day)", "rate(1 hour)"])
+# TODO move to test_events_schedules.py
```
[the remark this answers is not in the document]
maxhoheiser: It does not make sense here, since the hole structure of tests (file names, classes) needs to be reworked.

localstack/services/events/provider_v2.py
maxhoheiser: true, but in this case the function not only serves to convert time but also conditionally set the time based if the time filed is set in events or not.

tests/aws/services/events/test_events.py
```diff
@@ -139,7 +139,6 @@ def test_list_tags_for_resource(self, aws_client, clean_up):
         clean_up(rule_name=rule_name)
 
     @markers.aws.unknown
```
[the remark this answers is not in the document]
maxhoheiser: added validation to enabled tests and refactored tests

localstack/services/events/rule.py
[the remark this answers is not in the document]
maxhoheiser: Since we decided to only use the JAVA rule matching engine, there is no parsing of the input pattern required anymore, therefore there is also no validation of the pattern required.

localstack/services/events/provider_v2.py
```diff
@@ -578,25 +635,36 @@ def _delete_target_sender(self, ids: TargetIdList, rule) -> None:
...
-                            {
-                                "Entry": event,
-                                "ErrorCode": "InternalException",
-                                "ErrorMessage": str(error),
-                            }
-                        )
-        return failed_entries
+                rule_service = self.get_rule_service(context, rule.name, event_bus_name)
+                event_pattern = rule_service.event_pattern
+                if matches_event(event_pattern, event):
```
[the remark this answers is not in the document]
maxhoheiser: Since we decided to only use the JAVA rule/pattern matching engine, opt-in is removed

localstack/services/events/provider_v2.py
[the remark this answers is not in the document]
joe4dev: Botocore validation can be disabled using `parameter_validation=False` (see [botocore docs](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html#botocore.config.Config)). If it's a known issue, it might be worth adding a test case in a follow-up.

localstack/services/events/provider_v2.py
joe4dev: ok, then the naming `get_event_time` is a bit misleading here. Probably worth testing/refactoring in a follow-up.

localstack/services/events/models_v2.py
```diff
@@ -94,3 +94,11 @@ class ValidationException(ServiceException):
     code: str = "ValidationException"
     sender_fault: bool = True
     status_code: int = 400
+
+
+class InternalInvalidEventPatternException(Exception):
```
joe4dev, line 99: Why do we need to duplicate this exception?
Could we import the `InvalidEventPatternException` from the event_ruler as `InternalInvalidEventPatternException`?

localstack/services/events/event_ruler.py
```diff
@@ -43,8 +42,6 @@ def matches_rule(event: str, rule: str) -> bool:
     There is a single static boolean method Ruler.matchesRule(event, rule) -
     both arguments are provided as JSON strings.
     """
-    if config.EVENT_RULE_ENGINE != "java":
```
joe4dev, line 46: ok to remove this defensive guard to enable using the Java-based event ruler without setting the env variable 👍

localstack/services/events/provider_v2.py
```diff
@@ -79,6 +89,57 @@ def get_filtered_dict(name_prefix: str, input_dict: dict) -> dict:
...
+    formatted_event = {
+        "version": "0",
+        "id": str(long_uid()),
+        "detail-type": event.get("DetailType"),
+        "source": event.get("Source"),
+        "account": account_id,
+        "time": get_event_time(event),
+        "region": region,
+        "resources": event.get("Resources", []),
+        "detail": json.loads(event.get("Detail", "{}")),
```
joe4dev, line 137: According to the [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html#ev-events-detail), `detail` should always be a JSON object. Hence, I guess we can safely assume that other strings or base64-encoded content is not expected here, right?

> detail — A JSON object that contains information about the event. For more information about what can be included in this field, see [Event message detail field](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html#ev-events-detail).

localstack/services/events/provider_v2.py
```diff
@@ -578,25 +656,41 @@ def _delete_target_sender(self, ids: TargetIdList, rule) -> None:
...
-                    except Exception as error:
-                        failed_entries.append(
-                            {
-                                "Entry": event,
-                                "ErrorCode": "InternalException",
-                                "ErrorMessage": str(error),
-                            }
-                        )
-        return failed_entries
+                event_pattern = rule.event_pattern
```
joe4dev, line 680: super nit: `event_pattern_str` or str typing could clarify the type here (especially given we have both versions in our codebase)

tests/aws/services/events/test_events.py
```diff
@@ -76,6 +81,81 @@
...
+        entries = [
+            {
+                "DetailType": TEST_EVENT_PATTERN_NO_SOURCE["detail-type"][0],
+                "Detail": json.dumps(EVENT_DETAIL),
+            },
+        ]
+        response = aws_client.events.put_events(Entries=entries)
+        snapshot.match("put-events", response)
+
+    @markers.aws.unknown
```
joe4dev, line 99: Is this test not yet aws-validated?

tests/aws/services/events/test_events_integrations.py
```diff
@@ -483,7 +485,6 @@ def check_invocation():
 
 
 @markers.aws.unknown
```
joe4dev, line 487: not yet validated?

tests/aws/services/events/test_events_integrations.py
```diff
@@ -175,7 +180,6 @@ def test_put_events_with_target_sns(
...
```
joe4dev, line 182: not yet validated?

tests/aws/services/events/test_events_integrations.py
```diff
@@ -106,7 +112,6 @@ def test_put_events_with_target_sqs_event_detail_match(put_events_with_filter_to
...
```
joe4dev, line 113: not yet validated?

tests/aws/services/events/test_events_integrations.py
```diff
@@ -77,7 +84,6 @@ def test_put_events_with_target_sqs_new_region(aws_client_factory):
 
 
 @markers.aws.unknown
```
joe4dev, line 86: not yet validated?

tests/aws/services/events/test_events.py
```diff
@@ -817,7 +654,6 @@ def test_list_event_buses_with_limit(self, create_event_bus, aws_client, snapsho
     @markers.aws.unknown
```
joe4dev, line 654: not yet validated?

joe4dev (approved):
LGTM 👍 

Can we avoid duplicating the `InternalInvalidEventPatternException` before merging?
There are also a couple of enabled tests, which still have the `unknown` marker (see comments).

tests/aws/services/events/test_events.py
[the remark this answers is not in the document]
joe4dev: it seems there are still some `unknown` markers around for enabled tests

localstack/services/events/provider_v2.py
maxhoheiser: Now for the v2 provider, since we use only the Java matching engine we only ever have event pattern as a string, we don't parse it into a JSON object. In addition, the API string type is called `EventPattern`, therefor I would keep it without the str suffix. 
Also, I think the naming convention of match_rule is misleading since we don't match the rule but the rule_pattern and only pass in the pattern. We should think of updating this in the `event_ruler`.

localstack/services/events/models_v2.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10664/commits/81d63c0bcf34bcf70377b1858bb736165682c209

tests/aws/services/events/test_events.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10664/commits/d2b4586aebd3486090d03bc97244f06873ed1eed

tests/aws/services/events/test_events_integrations.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10664/commits/d2b4586aebd3486090d03bc97244f06873ed1eed

tests/aws/services/events/test_events_integrations.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10664/commits/d2b4586aebd3486090d03bc97244f06873ed1eed

tests/aws/services/events/test_events.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10664/commits/d2b4586aebd3486090d03bc97244f06873ed1eed

localstack/services/events/provider_v2.py
joe4dev: > Now for the v2 provider, since we use only the Java matching engine we only ever have event pattern as a string, we don't parse it into a JSON object. In addition, the API string type is called EventPattern, therefor I would keep it without the str suffix.

It's a good thing we don't need JSON parsing for the event pattern anymore in the new provider, but right one line below we have the counter-example of transforming an `event` into an `event_str` 😬 . Looking over the many implementations in LocalStack, this `str` vs `json` duality was very confusing, and being explicit can help to clarify the API usage here.

---

How do you think does `rule` differ from `rule_pattern`?
> the naming convention of match_rule is misleading since we don't match the rule but the rule_pattern and only pass in the pattern.

## Background

The terminology `Ruler.matchesRule(event, rule)`  in the `event_ruler.py` follows the official AWS [event-ruler](https://github.com/aws/event-ruler?tab=readme-ov-file#static-rule-matching) and aims to be more generic than just for EventBridge. Notice that different services use slightly different terminology:
* [Amazon EventBridge event patterns](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html)
* [Amazon EventBridge Pipes filtering](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-filtering.html), also called `FilterCriteria`
* Event Source Mapping [Lambda event filtering](https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html), also called `FilterCriteria`
* [Amazon SNS message filtering](https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html), also called `filter policy`

maxhoheiser: json vs strg makes sense - the old fallacy if you build it you know it and think its obvious to everyone else

PR #10671, opened 2024-04-16
fix missing href inside README.md

gabriel-vasile:
Previously: `install`, `installation`, and `installing` we're all used. Clicking on `Install` link inside README did nothing.
Currently: use `install` everywhere.

## Changes
- fix a link that is not working in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 <p align="center">
   <a href="#overview">Overview</a> •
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks a lot for proactively fixing the link with this PR! 💯 
Sorry for the delay, I'll merge this right away 😉

PR #10673, opened 2024-04-16
CFn: handle `OperationStatus.PENDING` events from resource providers

simonrw:
## Motivation

CloudFormation resource providers are able to return the `PENDING` status from operations such as `create` or `delete`, indicating that the operation is in progress. We can use this status in our CloudFormation engine to signal to the `do_apply_changes_in_loop` loop (and equivalent in `delete_stack`) that the operation is _knowingly_ not going to work at the moment, and we should come back to the operation and try again after continuing to attempt to operate on the other resources in the stack.

Specific example:

* Resource `A` depends on resources `B` and `C` (e.g. `A` = `AWS::RDS::DBCluster` and `B` and `C` are `AWS::RDS::DBInstance`s)
* CFn tries to delete `A` before deleting `B` and `C`, which raises an exception from the service

We can use the `PENDING` status to tell CloudFormation to give up trying to delete `A` until `B` and `C` have been deleted.

## Changes

* Handle `PENDING` status in `do_apply_changes_in_loop`
* Handle `PENDING` status in `delete_stack`

No specific tests have been added at this point, as it is part of the core engine and testing would be artificial.

### DIFF
--- a/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack/services/cloudformation/engine/template_deployer.py
@@ -415,7 +415,8 @@ def _resolve_refs_recursively(
             selected_map = mappings.get(mapping_id)
             if not selected_map:
                 raise Exception(
-                    f"Cannot find Mapping with ID {mapping_id} for Fn::FindInMap: {value[keys_list[0]]} {list(resources.keys())}"  # TODO: verify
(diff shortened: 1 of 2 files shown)


Morijarti (approved):
Thank you for adding these changes.
Nice to see improvements in core CFn engine

PR #10680, opened 2024-04-17
Core: Use EmulatorBaseStore

bblommers:
## Motivation
As part of the push towards LS becoming a multi-emulator platform, one of the key features is to have a generic data + persistence layer that can be used across all emulators.
This PR is a first step in that process, using a generic data store in LocalStack (AWS).

## Changes
 - Introduces the EmulatorBaseStore dependency
 - Overrides the BaseStore to keep the AWS-specific validation that was already in place

## Testing

Running the regular test suite is enough, as there are no changes in functionality.

## TODO

What's left to do:

- [ ] Review the [EmulatorBaseStore project](https://github.com/localstack/emulator-base-store/blob/main/emulator_base_store/base_store.py) (private for now)
- [ ] Decide on the naming
- [ ] Release the EmulatorBaseStore
- [ ] Change the dependency here to use the actual release

### DIFF
--- a/localstack/services/stores.py
+++ b/localstack/services/stores.py
@@ -36,9 +36,9 @@ from typing import Type
 from localstack import config
 from localstack.utils.aws.aws_stack import get_valid_regions_for_service
 
-from emulator_base_store.base_store import CrossAccountAttribute, CrossRegionAttribute, LocalAttribute  # noqa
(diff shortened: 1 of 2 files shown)


bblommers:
No longer necessary now that we have a monorepo

PR #10681, opened 2024-04-17
CFn: Parametrize Fn::GetAZs test across all regions

simonrw:
## Motivation

We cannot currently run the `test_get_azs_function` test against any other region but us-east-1 due to snapshot capture, and the `deploy_cfn_template` fixture being locked to the current profile region. We do not get coverage of other regions with this test.

Note: the `Fn::GetAZs` intrinsic function _must_ be called with the stack region as its argument, otherwise it returns nothing.

## Changes

- Test cleanup
- Support custom AWS client for `deploy_cfn_template` including standardising the stack deletion process between the automatic fixture cleanup, and the `DeployResult.destroy` method, ensuring the correct region is used
- Parametrize `test_get_azs_function` to evaluate all regions. This basically ignores the chosen region for the cross-region tests, but in a sense we gain _more_ coverage as we are testing **all** regions each time.
- As a side-effect of the removal of a 2 second sleep after manual destroy in a kinesis test, the kinesis stream resource provider has been updated to asynchronously delete the resource

**This PR will fail until the list of AZs per region as been corrected in `moto`.**

/cc @sannya-singal

### DIFF
--- a/tests/aws/services/cloudformation/test_template_engine.py
+++ b/tests/aws/services/cloudformation/test_template_engine.py
@@ -1,6 +1,7 @@
 import base64
 import json
 import os
+import re
 from copy import deepcopy
 
 import botocore.exceptions
@@ -247,15 +248,13 @@ class TestIntrinsicFunctions:
         ],
     )
     @markers.aws.validated
-    def test_get_azs_function(self, deploy_cfn_template, region, aws_client_factory, snapshot):
+    def test_get_azs_function(self, deploy_cfn_template, region, aws_client_factory):
         """
         TODO parametrize this test.
         For that we need to be able to parametrize the client region. The docs show the we should be
         able to put any region in the parameters but it doesn't work. It only accepts the same region from the client config
         if you put anything else it just returns an empty list.
         """
-        snapshot.add_transformer(snapshot.transform.regex(region, "<region>"))
-
         template_path = os.path.join(
             os.path.dirname(__file__), "../../templates/functions_get_azs.yml"
         )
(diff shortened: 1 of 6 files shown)


localstack/testing/pytest/fixtures.py
```diff
@@ -1005,17 +1003,18 @@ def _deploy(
         if role_arn is not None:
             kwargs["RoleARN"] = role_arn
 
-        response = aws_client.cloudformation.create_change_set(**kwargs)
+        cfn_aws_client = custom_aws_client if custom_aws_client is not None else aws_client
```
sannya-singal, line 1006: we can add a comment here stating why was this custom aws client required

sannya-singal (approved):
Thanks @simonrw for creating this PR and having a fun exploration debugging session 🙌 🚀 The fixes LGTM!

simonrw:
As found by @sannya-singal, there was an attempt to [update moto a little while ago](https://github.com/getmoto/moto/issues/5494) which was reverted.

localstack/services/kinesis/resource_providers/aws_kinesis_stream.py
```diff
@@ -138,14 +138,24 @@ def delete(
...
+        model = request.previous_state
+        client = request.aws_client_factory.kinesis
+
+        if not request.custom_context.get(REPEATED_INVOCATION):
+            client.delete_stream(StreamARN=model["Arn"], EnforceConsumerDeletion=True)
+            request.custom_context[REPEATED_INVOCATION] = True
+
+        try:
+            client.describe_stream(StreamARN=model["Arn"])
+            return ProgressEvent(
```
Morijarti, line 150: All good with this changes for Kinesis.
We should probably check in future, if we are missing clean up for any other service

tests/aws/services/cloudformation/test_template_engine.py
```diff
@@ -234,24 +233,40 @@ def test_cidr_function(self, deploy_cfn_template):
...
         able to put any region in the parameters but it doesn't work. It only accepts the same region from the client config
         if you put anything else it just returns an empty list.
         """
+        snapshot.add_transformer(snapshot.transform.regex(region, "<region>"))
 
         template_path = os.path.join(
             os.path.dirname(__file__), "../../templates/functions_get_azs.yml"
         )
 
+        aws_client = aws_client_factory(region_name=region)
```
Morijarti, line 263: This logic looks good to me.
We are allowing CFn to use custom client for deployment.

PR #10682, opened 2024-04-17
fix apigw data plane service matching

thrau:
## Motivation

After merging https://github.com/localstack/localstack/pull/10640, we noticed an issue in ext that a request to apigwv2 websockets would mistakenly go into the service parser, which would fail because the request was detected as "apigateway" when it really should have been "apigatewaymanagementapi". It turns out setting `apigateway` as a service simply based on the condition `if `.execute-api. in request.host` is much too strong, because it's ambiguous.

however, we were relying on this (wrong) behavior in the CORS handler, to trigger the custom apigw cors check, so we had to weaken the condition for that control path as well, now that we're no longer setting the service name in the request to `apigateway`.

Moving forward, we have the option to handle data plane service loading without setting `context.service` (which at the end of the day is just an optimization, not a necessity for loading the service), and we can load more than one service in the `if `.execute-api. in request.host` case (namely apigw v1, v2 and management API). the decision whether we want to do that i've left to @bentsku who will give it some thought.

until then we accept the limitation that pre-loading the service from data-plane URLs will not work for api gateway.

## Changes

* apigateway is no longer set automatically as service when `.execute-api. in request.host`
* custom cors check for APIGW is now done independently on whether the service is set or not (TODO: we could probably optimize and skip the check completely if the service is already set to something that's not APIGW)

## TODO

What's left to do:

- [x] Run the ext pipeline on this change

### DIFF
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -145,19 +145,19 @@ def should_enforce_self_managed_service(context: RequestContext) -> bool:
                     the targeting service
     :return: True if the CORS rules should be enforced in here.
     """
-    if config.DISABLE_CUSTOM_CORS_S3 and config.DISABLE_CUSTOM_CORS_APIGATEWAY:
-        return True
     # allow only certain api calls without checking origin
-    if context.service:
-        service_name = context.service.service_name
-        if not config.DISABLE_CUSTOM_CORS_S3 and service_name == "s3":
+    if not config.DISABLE_CUSTOM_CORS_S3:
+        if context.service and context.service.service_name == "s3":
             return False
-        if not config.DISABLE_CUSTOM_CORS_APIGATEWAY and service_name == "apigateway":
-            is_user_request = (
-                PATH_USER_REQUEST in context.request.path or ".execute-api." in context.request.host
-            )
-            if is_user_request:
-                return False
+
+    if not config.DISABLE_CUSTOM_CORS_APIGATEWAY:
+        # we don't check for service_name == "apigw" here because ``.execute-api.`` can be either apigw v1 or v2
+        is_user_request = (
+            ".execute-api." in context.request.host or PATH_USER_REQUEST in context.request.path
+        )
+        if is_user_request:
+            return False
+
     return True
 
 
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -140,10 +140,13 @@ def custom_host_addressing_rules(host: str) -> Optional[ServiceModelIdentifier]:
     """
     Rules based on the host header of the request, which is typically the data plane of a service.
 
-    # TODO: ELB, AppSync, CloudFront, ...
+    Some services are added through a patch in ext.
     """
-    if ".execute-api." in host:
-        return ServiceModelIdentifier("apigateway")
+
+    # a note on ``.execute-api.`` and why it shouldn't be added as a check here: ``.execute-api.`` was previously
+    # mapped distinctly to ``apigateway``, but this assumption is too strong, since the URL can be apigw v1, v2,
+    # or apigw management api. so in short, simply based on the host header, it's not possibly to unambiguously
+    # associate a specific apigw service to the request.
 
     if ".lambda-url." in host:
         return ServiceModelIdentifier("lambda")
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -145,7 +145,7 @@ def custom_host_addressing_rules(host: str) -> Optional[ServiceModelIdentifier]:
 
     # a note on ``.execute-api.`` and why it shouldn't be added as a check here: ``.execute-api.`` was previously
     # mapped distinctly to ``apigateway``, but this assumption is too strong, since the URL can be apigw v1, v2,
-    # or apigw management api. so in short, simply based on the host header, it's not possibly to unambiguously
+    # or apigw management api. so in short, simply based on the host header, it's not possible to unambiguously
     # associate a specific apigw service to the request.
 
     if ".lambda-url." in host:


localstack/aws/protocol/service_router.py
```diff
@@ -140,10 +140,13 @@ def custom_host_addressing_rules(host: str) -> Optional[ServiceModelIdentifier]:
...
 
-    # TODO: ELB, AppSync, CloudFront, ...
+    Some services are added through a patch in ext.
     """
-    if ".execute-api." in host:
-        return ServiceModelIdentifier("apigateway")
+
+    # a note on ``.execute-api.`` and why it shouldn't be added as a check here: ``.execute-api.`` was previously
+    # mapped distinctly to ``apigateway``, but this assumption is too strong, since the URL can be apigw v1, v2,
+    # or apigw management api. so in short, simply based on the host header, it's not possibly to unambiguously
```
alexrashed, line 148: nit: Small typo, which made me take embarrassingly long to read this sentence 😛 (please don't add a co-authorship for this though 😅).
```suggestion
    # or apigw management api. so in short, simply based on the host header, it's not possible to unambiguously
```

bentsku (approved):
LGTM! Seems like everything is green for AMD64, so should be safe to merge.

Things look good to me, I believe the CORS handler will still do its job. Pretty weird that anything similar with setting the service to `apigateway`. 

We can keep thinking about returning list of services for a certain data plane, I'd like to see if we have other services in this situation also, which would make it worth generalizing this over. 

Also, this issue was due to a very weird set of condition, mainly `PostToConnection` needing to have the API endpoint and the stage set as its `endpoint-url`, but it being a provider operation. 

All in all, LGTM, and let's tackle loading API GW in a follow up sometimes soon! 🚀
Thanks a lot for jumping on this 🙏

localstack/aws/handlers/cors.py
```diff
@@ -145,19 +145,19 @@ def should_enforce_self_managed_service(context: RequestContext) -> bool:
...
-            is_user_request = (
-                PATH_USER_REQUEST in context.request.path or ".execute-api." in context.request.host
-            )
-            if is_user_request:
-                return False
+
+    if not config.DISABLE_CUSTOM_CORS_APIGATEWAY:
+        # we don't check for service_name == "apigw" here because ``.execute-api.`` can be either apigw v1 or v2
+        is_user_request = (
+            ".execute-api." in context.request.host or PATH_USER_REQUEST in context.request.path
```
alexrashed, line 156: I think there is a very small edge case that could actually slip through this here, which would mean that a request which should potentially be blocked is not checked.
Ultimately, I think this if has to be equal to the check in the service router:
https://github.com/localstack/localstack/blob/9af4246ef854ab27a2f420799ce45caedf8d2f26/localstack/aws/protocol/service_router.py#L181-L183

Actually, we have multiple rules in the service router for the `.execute-api.` in the host, so I think the only difference would be the surrounding slashes in the path for `PATH_USER_REQUEST`:
https://github.com/localstack/localstack/blob/451968fabb223e189323df02a8d08fac110fddec/localstack/aws/protocol/service_router.py#L145-L147

Which means the following check should be equivalent to the service router rules, ultimately closing this small gap:
```suggestion
            ".execute-api." in context.request.host or f"/{PATH_USER_REQUEST}/" in context.request.path
```

What do you think?

alexrashed (commented):
The changes are looking good, and they seem to resolve our failing test! Thanks a lot for jumping on this! 🚀 🥳 
I only found a small nitpick and one potential small improvement, let me know what you think.

localstack/aws/handlers/cors.py
bentsku: Do you think that's blocking? To me the check in the service router is only stricter. Which means the CORS check would not let something pass with this fix that it did not let before.

bentsku: Also, I think this legacy condition: `host.endswith(LOCALHOST_HOSTNAME) and "execute-api" in host` is not really good anymore with the changes from Simon what's after the prefix is not matched anymore. I think this code "should" be dead code.

To resume, we do not need to have the service name set for the router, and the APIGW routes will match with or without the service set in the context.

The issue created by the previous PR and now fixed with this PR is that we would set the service early for a request that what sent with the SDK/AWS-CLI (so it signed the request, the service name parser would return very early with the signing name), but the `execute-api` endpoint. See https://docs.aws.amazon.com/cli/latest/reference/apigatewaymanagementapi/post-to-connection.html#examples
And because the service name parser returns early if the service is already set, we would skip and keep `apigateway` (v1) set for that request. 

So in the end, the fact that we removed it from the CORS check means we should be able to remove this legacy condition in the router too.

thrau:
Merging to get the pipeline unblocked

localstack/aws/handlers/cors.py
bentsku: I see what you mean, thanks for clarifying.
I think both previous and current approach are very approximative, and do not really guarantee a correct detection of APIGW. Even if `PATH_USER_REQUEST`  is between `/`, it does not mean it's going to match with a route, the first one linked here does not have a mandatory slash on the right. Also, the `PATH_USER_REQUEST` between slashes could be anywhere in the path, and the route might not match. So it's already a pretty flawed approach, and the change makes it a bit weaker only. 

Also, I think with our current approach of both CORS and CRSF in the same check, we're not really weakening CORS, because we will just not return CORS headers when we should, which in the end doesn't allow the browser to read the response and block it. But the same check is used for CRSF, which means we're not blocking it either.

We could maybe create stricter conditions, being that the current request matches against APIGW registered routes? I can have a quick look into it, to make the condition stricter (regex matching on the host or the path), but that's a bit expensive.

bentsku: I'm currently working on a PR and I will add tests for it. 
I think we will need to revisit anyway, because at some point it would be good to be able to execute those routes from the Web App somehow, but currently you would have to setup CORS manually in your API for that. I'll keep that in mind for the future. 
CORS handling is a bit trickier in APIGW than S3 because there are no explicit APIs to manage it, it's either MOCK responses for `OPTIONS` or your lambda returning it... not an easy one.

This is the PR: https://github.com/localstack/localstack/pull/10688

Thanks a lot for raising this!

Issue #10679, opened 2024-04-17
bug: DynamoDB TransactWriteItems Update Produces `MODIFY` CDC event for new items


rajveerappan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using TrasactWriteItems with Update action produces a `MODIFY` event on the DynamoDB stream even for new items

### Expected Behavior

Using TrasactWriteItems with Update action produces an `INSERT` event on the DynamoDB stream for new items

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose.yml

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

1. Create a DynamoDB table and configure it to produce `NEW_AND_OLD_IMAGES` on the DynamoDB stream
2. Use aws cli or another method to TransactWriteItem a list of one or more **new** items (i.e. brand new keys) and specify `Update` as the action for the items
3. Log the event that is produced on the DynamoDB stream

The event on the DynamoDB stream will say `'eventName': 'MODIFY'` and will only contain the `NewImage`. It should say `'eventName': 'INSERT'` instead

### Environment

```markdown
- OS: macOS 14.1.1
- LocalStack: 3.3.0 (reproducible in 3.2.0 also)
```

### Anything else?

[DynamoDB docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html#transaction-apis-txwriteitems:~:text=Update%C2%A0%E2%80%94%C2%A0Initiates%20an%20UpdateItem%20operation%20to%20edit%20an%20existing%20item%27s%20attributes%20or%20add%20a%20new%20item%20to%20the%20table%20if%20it%20does%20not%20already%20exist.%20Use%20this%20action%20to%20add%2C%20delete%2C%20or%20update%20attributes%20on%20an%20existing%20item%20conditionally%20or%20without%20a%20condition.) indicating that `Update` will insert items if they don't already exist:
> Update — Initiates an UpdateItem operation to edit an existing item's attributes **or add a new item to the table if it does not already exist**. Use this action to add, delete, or update attributes on an existing item conditionally or without a condition.

Docs for [CDC event schema](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_Record.html#:~:text=eventName,the%20DynamoDB%20table) for DynamoDB:

[...]

PR #10685, opened 2024-04-17
fix DDB eventName for TransactWriteItems Update inserting data

bentsku:
## Motivation
As reported with #10679, we would return `MODIFY` when the call to `Update` would insert data that didn't exist.
A condition was wrong and was checking for the wrong value (`updated_item` instead of `existing_item`). 

## Changes
- added a new call in the `test_transact_write_items_streaming` test to test for an `Update` call which inserts data
- fix the condition check 

_fixes #10679_


thrau (approved):
nice! thanks for fixing this so quickly ben

PR #10688, opened 2024-04-18
improve user request route detection for APIGW in CORS

bentsku:
## Motivation
Follow up from #10682, Alex has raised valid concerns that we were weakening the CORS/CRSF checks for APIGW by removing the service check.

## Changes
- I've added a bit stricter check than before, and added tests for APIGW CORS handling.

### DIFF
--- a/localstack/aws/handlers/cors.py
+++ b/localstack/aws/handlers/cors.py
@@ -154,7 +154,7 @@ def should_enforce_self_managed_service(context: RequestContext) -> bool:
         # we don't check for service_name == "apigw" here because ``.execute-api.`` can be either apigw v1 or v2
         path = context.request.path
         is_user_request = ".execute-api." in context.request.host or (
-            path.startswith("/restapis") and f"/{PATH_USER_REQUEST}" in context.request.path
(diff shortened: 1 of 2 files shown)


localstack/aws/handlers/cors.py
```diff
@@ -152,8 +152,9 @@ def should_enforce_self_managed_service(context: RequestContext) -> bool:
 
     if not config.DISABLE_CUSTOM_CORS_APIGATEWAY:
         # we don't check for service_name == "apigw" here because ``.execute-api.`` can be either apigw v1 or v2
-        is_user_request = (
-            ".execute-api." in context.request.host or PATH_USER_REQUEST in context.request.path
+        path = context.request.path
+        is_user_request = ".execute-api." in context.request.host or (
+            path.startswith("/restapis") and f"/{PATH_USER_REQUEST}" in context.request.path
```
alexrashed, line 157: This check seems good to me, we could just add the trailing slash to `restapis` here as well I guess:
```suggestion
            path.startswith("/restapis/") and f"/{PATH_USER_REQUEST}" in context.request.path
```

alexrashed (approved):
Thanks for jumping on this and writing a test to avoid any regression on this topic! 🚀

PR #10689, opened 2024-04-18
Fix missing config in list for CLI

joe4dev:
## Motivation

In https://github.com/localstack/localstack/pull/10615, we missed to add the new config flag `EVENT_RULE_ENGINE` into the list. Hence, the CLI does not pass the flag.

## Changes

Add missing config to config list for CLI.


dfangl (approved):
Also missed that, thanks for adding it!

PR #10690, opened 2024-04-18
Avoid failing host_path_for_path_in_docker if no docker daemon is available

dfangl:
## Motivation
The `get_host_path_for_path_in_docker` function should not fail if we have no docker daemon available.
If docker is required later, we can still fail there. If not, because there are other options available, we should not fail.

This is a bandaid fix, we need to have a bigger refactoring for this method in the future.

## Changes
* Check if we have access to the docker socket before executing operations for it
* Check `has_docker` call for the sdk client as well, similar to the cmd client.
* Properly pass ulimit configuration from `ContainerConfiguration` to `create_container`.


simonrw (approved):
Great, glad we have this cleanup ready.

Issue #9738, opened 2023-11-27
enhancement request: SNS FilterPolicy extras support


robinduckett:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

AWS has added extra features to their SNS filter policy system and the LocalStack SNS does not support them:

The features as noted by @bram-tv:

[$or operator](https://docs.aws.amazon.com/sns/latest/dg/and-or-logic.html#or-operator)
[suffix matching](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-suffix-matching)
[case insensitive matching ("equals-ignore-case")](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-equals-ignore)

### Anything else?

More info:
* https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-sns-message-additional-filtering-operators/
* https://docs.aws.amazon.com/sns/latest/dg/and-or-logic.html#or-operator

robinduckett:
@bram-tv  that’s fair, not sure how I misread that - any suggestions for a better title? Other than omitting the nonexistent operator

PR #10691, opened 2024-04-19
implement SNS Filter/operators $or, suffix, equals-ignore-case, anything-but

bentsku:
## Motivation
As reported in #9738, AWS launched new operators for SNS filtering. `suffix` and `equals-ignore-case` are pretty simple to implement, but `$or` necessitated a quite consequent rewrite, as the logic can get quite convoluted pretty fast. It quickly creates a lot of different forks, where one rule can have multiple branches and a payload can have multiple branches too if it contains array. 

Also included better support for `anything-but`, which can accept any regular "plain" values like string, numbers and list of string & number. It can also accept the `prefix` operator, see https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-anything-but-matching-prefix

Also, we still had an issue with some ordering of complex payload with arrays, see https://github.com/localstack/localstack/issues/10567#issuecomment-2096128740

## Changes
- switched the model for the policy, making it a list where each item of the list is an OR condition. This comes from the documentation of AWS: https://docs.aws.amazon.com/sns/latest/dg/and-or-logic.html
  - this is done via the improved `flatten_policy`, thanks a lot @cloutierMat for the great pairing session we've had, and you providing the final solution fixing all our issues! 🥳 
- added support for `suffix` and `equals-ignore-case`
- improved support for `anything-but`
- added a lot of unit tests for filtering
- added a few AWS tests to validate the new behavior
- remove the `skip` for the AWS validated `$or` test
- fix the recursive function `flatten_payload`

_fixes #10567, fixes #9738_

### DIFF
--- a/tests/aws/services/sns/test_sns_filter_policy.py
+++ b/tests/aws/services/sns/test_sns_filter_policy.py
@@ -1067,7 +1067,6 @@ class TestSNSFilterPolicyBody:
         snapshot.match("messages", {"Messages": received_messages})
 
     @markers.aws.validated
-    @pytest.mark.skip("Not yet supported by LocalStack")
     def test_filter_policy_on_message_body_or_attribute(
         self,
         sqs_create_queue,
@@ -1237,7 +1236,7 @@ class TestSNSFilterPolicyConditions:
         topic_arn = sns_create_topic()["TopicArn"]
 
         def _subscribe(policy: dict):
-            sns_subscription(
+            return sns_subscription(
                 TopicArn=topic_arn,
                 Protocol="sms",
                 Endpoint=phone_number,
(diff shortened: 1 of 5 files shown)


tests/aws/services/sns/test_sns_filter_policy.py
```diff
@@ -1413,6 +1424,72 @@ def _subscribe(policy: dict):
...
+            filter_policy = {"key": [{"anything-but": {"wrong-operator": None}}]}
+            _subscribe(filter_policy)
+        self._add_normalized_field_to_snapshot(e.value.response)
+        snapshot.match("error-condition-wrong-operator", e.value.response)
+
+        with pytest.raises(ClientError) as e:
+            filter_policy = {"key": [{"anything-but": {"suffix": "test"}}]}
+            _subscribe(filter_policy)
+        self._add_normalized_field_to_snapshot(e.value.response)
+        snapshot.match("error-condition-anything-but-suffix", e.value.response)
```
cloutierMat, line 1457: Nice! I kept wondering if this was a thing we needed to add support for! :zap:

cloutierMat (approved):
Lots of great work! :rocket: That is a pretty big step forward! 

I like your simplification of the recursive function for the policy.

tests/aws/services/sns/test_sns_filter_policy.py
bentsku: yeah me too, I thought the documentation was lying... but no, they really do not support anything else than `prefix` 😆

bentsku:
> I like your simplification of the recursive function for the policy.

Thanks @cloutierMat, it's actually heavily inspired by yours 🚀 you definitely fixed all the recursion of SNS! Thank you so much 🙏

Issue #7734, opened 2023-02-22
bug: SAM deploy doesn't seem to be creating the EventSourceMapping, causing the stack to fail deployment


JoeWHoward:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running samlocal deploy with my cfn.yaml, containing a Lambda and an SQS queue, the EventSourceMapping does not get created from the "[Events](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-events)" object specified on the Lambda.  Since this resource doesn't get created, the stack creation fails.

### Expected Behavior

Deploying a SAM template lambda with Events: specified to create the appropriate EventSourceMapping for the given SQS Queue.

### How are you starting LocalStack?

localstack start

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

In a Python 3.10 environment:  `localstack start` (with Docker daemon running in background)

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

`samlocal deploy --stack-name testStack --template-file cfn.yaml --capabilities CAPABILITY_IAM --resolve-s3`
    
**cfn.yaml:**
```yaml
AWSTemplateFormatVersion: 2010-09-09
Transform: AWS::Serverless-2016-10-31

Description: |
  Example of SAM EventSourceMapping failure

Globals:
  Function:
    Timeout: 3

Resources:
  #########################
  #        Lambda         #
  #########################
  Lambda:
    Type: AWS::Serverless::Function
    Properties:
      AutoPublishAlias: live
      CodeUri: build/
      Handler: bundle.handle
      Runtime: nodejs16.x
      Events:
        SourceSQSEvent:
          Type: SQS
          Properties:
            Queue: !GetAtt SQSQueue.Arn
            BatchSize: 10
            MaximumBatchingWindowInSeconds: 10
      Architectures:
        - arm64

  SQSQueue:
    Type: AWS::SQS::Queue
    Properties:
      MessageRetentionPeriod: 3600
      VisibilityTimeout: 60
```

### Environment

```markdown
- OS: macOS (Apple Silicon/M1)
- LocalStack: latest
```

### Anything else?

![samlocal deploy log](https://user-images.githubusercontent.com/9206287/220769014-e69f9feb-e79e-44ec-a75d-9180cdfc2d67.png)

[...]

PR #10700, opened 2024-04-19
Capture example lambda triggered by SQS event

simonrw:
## Motivation

Based on https://github.com/localstack/localstack/issues/7734, capture a lambda function being invoked by an SQS message via a SAM template.

## Changes

* Include new test for SAM template
* Validate test against AWS

Closes https://github.com/localstack/localstack/issues/7734


pinzon (approved):
LGTM

PR #10701, opened 2024-04-21
create pattern construct for a webapp extension

thrau:
## Motivation

Webapp extensions such as:
* https://github.com/localstack/localstack-ngrok-tunnels
* https://github.com/thrau/localstack-aws-log-viewer
* https://github.com/localstack/localstack-extension-event-studio

Follow a common pattern. They
* expose routes through subdomains and submounts (`http://my-extension.localhost.localstack.cloud:4566`, and `http://localhost.localstack.cloud:4566/_extension/my-extension`)
* Enable logging for their root module `my_extension`
* Serve jinja2 templates
* Serve static files

Their file layout will typically look like this:
```
my_extension
├── extension.py
├── __init__.py
├── static              <-- make sure static resources get packaged!
│   ├── __init__.py
│   ├── favicon.ico
│   └── style.css
└── templates            <-- jinja2 templates
   └── index.html
```

This PR introduces this pattern as a class that can be extended to quickly develop extensions like this.

Examples of how existing extensions have been simplified can be found in:
* https://github.com/thrau/localstack-aws-log-viewer/pull/1
* https://github.com/localstack/localstack-ngrok-tunnels/commit/1838953260e44ca31046928f9576c7902abc17d1
* https://github.com/localstack/localstack-extension-event-studio

Requires [Rolo v0.5.0](https://github.com/localstack/rolo/releases/tag/v0.5.0)

## Changes

* Introduce `localstack.extension.patterns`
* Add a high-level pattern `WebAppExtension`

## Testing

If you want to test this you can:

* download and checkout https://github.com/thrau/localstack-aws-log-viewer/pull/1 (or just build your own extension if you want)
  * run `make install`
  * run `localstack extensions dev enable .`
* checkout this branch
* in localstack-ext, run `EXTENSION_DEV_MODE=1 python -m localstack.dev.run -e LOCALSTACK_API_KEY=test -e EXTENSION_DEV_MODE=1`

## TODO

What's left to do:

- [x] Wait for rolo dependency to be updated v0.5.0

### DIFF
--- a/localstack/extensions/patterns/webapp.py
+++ b/localstack/extensions/patterns/webapp.py
@@ -13,7 +13,7 @@ from localstack.aws.api import RequestContext
 from localstack.extensions.api import Extension, http
 
 if t.TYPE_CHECKING:
-    # make sure jinja2 stays optional
+    # although jinja2 is included transitively via moto, let's make sure jinja2 stays optional
     import jinja2
 
 LOG = logging.getLogger(__name__)
@@ -25,7 +25,8 @@ class WebAppExtension(Extension):
     """
     EXPERIMENTAL! This class is experimental and the API may change without notice.
 
-    A webapp extension serves routes, templates, and static files via a submount and a subdomain through localstack.
+    A webapp extension serves routes, templates, and static files via a submount and a subdomain through
+    localstack.
 
     It assumes you have the following directory layout::
 
@@ -59,6 +60,13 @@ class WebAppExtension(Extension):
     * Submount: https://localhost.localstack.cloud:4566/_extension/my-extension
     * Subdomain: https://my-extension.localhost.localstack.cloud:4566/
 
+    Both are created for full flexibility:
+
+    * Subdomains: create a domain namespace that can be helpful for some extensions, especially when
+      running on the local machine
+    * Submounts: for some environments, like in ephemeral instances where subdomains are harder to control,
+      submounts are more convenient
+
     Any routes added by the extension will be served relative to these URLs.
     """
 
@@ -72,17 +80,20 @@ class WebAppExtension(Extension):
         static_url_path: str = None,
     ):
         """
-        Overwrite to customize your extension. For example, you can disable certain behavior by calling ``super(
-        ).__init__(subdomain=None, static_package_path=None)``, which will disable serving through a subdomain,
-        and disable static file serving.
+        Overwrite to customize your extension. For example, you can disable certain behavior by calling
+        ``super( ).__init__(subdomain=None, static_package_path=None)``, which will disable serving through
+        a subdomain, and disable static file serving.
 
-        :param mount: the "mount point" which will be used as default value for the submount and subdirectory, i.e.,
-            ``<mount>.localhost.localstack.cloud`` and ``localhost.localstack.cloud/_extension/<mount>``. Defaults to
-            the extension name.
-        :param submount: the submount path, needs to start with a trailing slash (default ``/_extension/<mount>``)
+        :param mount: the "mount point" which will be used as default value for the submount and
+            subdirectory, i.e., ``<mount>.localhost.localstack.cloud`` and
+            ``localhost.localstack.cloud/_extension/<mount>``. Defaults to the extension name.  Note that,
+            in case the mount name clashes with another extension, extensions may overwrite each other's
+            routes.
+        :param submount: the submount path, needs to start with a trailing slash (default
+            ``/_extension/<mount>``)
         :param subdomain: the subdomain (defaults to the value of ``mount``)
-        :param template_package_path: the path to the templates within the module. defaults to ``templates`` which
-            expands to ``<extension-module>.templates``)
+        :param template_package_path: the path to the templates within the module. defaults to
+            ``templates`` which expands to ``<extension-module>.templates``)
         :param static_package_path: the package serving static files. defaults to ``static``, which expands to
             ``<extension-module>.static``.
         :param static_url_path: the URL path to serve static files from (defaults to `/static`)
@@ -104,8 +115,8 @@ class WebAppExtension(Extension):
 
     def collect_routes(self, routes: list[t.Any]):
         """
-        This method can be overwritten to add more routes to the controller. Everything in ``routes`` will be added
-        to a ``RuleAdapter`` and subsequently mounted into the gateway router.
+        This method can be overwritten to add more routes to the controller. Everything in ``routes`` will
+        be added to a ``RuleAdapter`` and subsequently mounted into the gateway router.
 
         Here are some examples::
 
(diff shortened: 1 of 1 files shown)


localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+        template_package_path: str | None = _default,
+        static_package_path: str | None = _default,
+        static_url_path: str = None,
+    ):
+        """
+        Overwrite to customize your extension. For example, you can disable certain behavior by calling ``super(
+        ).__init__(subdomain=None, static_package_path=None)``, which will disable serving through a subdomain,
+        and disable static file serving.
+
+        :param mount: the "mount point" which will be used as default value for the submount and subdirectory, i.e.,
```
joe4dev, line 79: What happens upon naming clashes? the last registered route wins 🤔

localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+from rolo.router import RuleAdapter, WithHost
+from werkzeug.routing import Submount
+
+from localstack import config
+from localstack.aws.api import RequestContext
+from localstack.extensions.api import Extension, http
+
+if t.TYPE_CHECKING:
+    # make sure jinja2 stays optional
+    import jinja2
```
joe4dev, line 17: might be worth commenting that `jinja2` is a peer dependency that needs to be installed through the extension for templating

localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+        Factory method to create the jinja2 template environment.
+        :return: a new jinja2 environment
+        """
+        import jinja2
+
+        return jinja2.Environment(
+            loader=jinja2.PackageLoader(
+                self.get_extension_module_root(), self.template_package_path
+            ),
+            autoescape=jinja2.select_autoescape(),
```
joe4dev, line 180: Let's hope this jinja2 API never changes because the peer dependency in the extension needs to package a compatible version 🤞

localstack/extensions/patterns/container.py
joe4dev, line 1: Are these empty files intentional?

localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        Default pre-processor, which implements a default behavior to add a trailing slash to the path if the
+        submount is used directly. For instance ``/_extension/my-extension``, then it forwards to
+        ``/_extension/my-extension/``. This is so you can reference relative paths like ``<link
+        href="static/style.css">`` in your HTML safely, and it will work with both subdomain and submount.
+        """
+        path = context.request.path
+
+        if path == self.submount.rstrip("/"):
+            chain.respond(301, headers={"Location": context.request.url + "/"})
```
joe4dev, line 235: nice catch 💯 
tested ✔️

joe4dev (commented):
Looking at the extension examples, it makes the API so much easier to use. Auto-CORS config is a nice extra as well. 🙌 

I tested these changes with the https://github.com/localstack/localstack-extension-event-studio and it works like a charm. 👍 

I'm not so familiar with our routing internals and would suggest someone else signing it off, for example @alexrashed or @dfangl.

localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+                # reference `static/style.css` to serve the static file from your package
+                return self.render_template_response("index.html")
+
+            @route("/hello")
+            def hello(request: Request):
+                return {"message": "Hello World!"}
+
+    This will create an extension that localstack serves via:
+
+    * Submount: https://localhost.localstack.cloud:4566/_extension/my-extension
```
alexrashed, line 59: nit: I guess this already evolved and is used for existing extensions, which would make it hard to change, but I would suggest to use the plural for API paths that contain lists:
```suggestion
    * Submount: https://localhost.localstack.cloud:4566/_extensions/my-extension
```

And I guess it was already discussed in the past if the `_extension` endpoint should be a top-level entry / first class citizen in LocalStack? For example, CloudPods live in `/_localstack/pods`. But extensions can also be totally unrelated to LocalStack itself (like `mailhog`), where it might even be confusing...

localstack/extensions/patterns/webapp.py
```diff
@@ -0,0 +1,321 @@
...
+    * Subdomain: https://my-extension.localhost.localstack.cloud:4566/
```
alexrashed, line 60: nit / q: If there are a _lot_ of extensions this might become problematic because this could overlap with other domain-based routes. Would it make sense to have them combined in a defined subcomain like `my-extension.extensions.localhost.localstack.cloud`?

localstack/extensions/patterns/container.py
alexrashed: +1 for removing this file before merging. :)

alexrashed (approved):
Really nice abstraction, which saves a lot of repetitive code among the different extension implementations, but also helps tackling edge cases in a generic way (f.e. the 301 redirect for relative imports)! 🚀 
I think this can be merged as is (after removing the `container.py`). 🥳

localstack/extensions/patterns/webapp.py
thrau: True. nothing we can do about that right now unfortunately. Added a comment though

localstack/extensions/patterns/webapp.py
thrau: for now it should be OK since we have a pin on jinja through moto. in the worst case people can override the method in their extension.

but the issue of dependencies in extensions (and potential conflicts) is something we definitely need to tackle at some point.

localstack/extensions/patterns/container.py
thrau: i gave away the other patterns i'm playing with :D

localstack/extensions/patterns/webapp.py
thrau: well it hasn't really been discussed directly, i'm making a "product owner" decision here to introduce it. i believe having a well-known namespace for extensions is important to have since it allows us to reason over the requests. moreover, it should be distinct from `/_localstack` for the reason you correctly identified. we should port our existing extensions to that convention.

i prefer singular for the following (completely arbitrary and personal taste reason which you are welcome to disagree with :D). we have `/_localstack/extensions` which indicates that you can _manage_ one or more extensions. whereas the singular `/_extension/<something>` indicates that you're talking to an individual and specific extension.

alexrashed: I feel like it's a best practice when it comes to HTTP / RESTful API design to use the plural for path segments that in turn contains a collection of sub-resources (`/_extensions` is a collection resource whereas the specific extension route would expose a single resource). I don't see a reason to use the `/_extension(s)` path in the future to show metadata on the _collection_ of extensions registered in this "namespace", for example a `GET /_extensions` listing the different sub-routes for `WebAppExtensions`...
But again, just a nitpick 😛

thrau: All fair points. This is not a REST API though :)

localstack/extensions/patterns/webapp.py
thrau: definitely a good point that we can explore! for now i'll keep this so we don't need to create a new wildcard certificate. for `*.extension....`

thrau:
marking this as patch since it is purely additive and affects no existing localstack control paths

PR #10703, opened 2024-04-22
switch default gateway server from hypercorn to twisted

alexrashed:
## Motivation
@thrau added twisted as supported gateway server with #9834. Since then this was tested by selected users, and it turned out to be a great server. In reduced complexity and improved performance.

## Changes
- Switches the default for `GATEWAY_SERVER` from `hypercorn` to `twisted`.

## Testing
- Since the gateway is the foundation for every single request served in LocalStack, this change has an insane coverage. If all our tests are 💚, we should be good to go!
- We also need to execute tests for our downstream projects, to make sure other projects don't break.

## TODO
- [x] Execute tests in downstream projects.
- [x] Fix issues with streaming responses in our downstream usage.

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -658,7 +658,7 @@ def populate_edge_configuration(
 GATEWAY_WORKER_COUNT = int(os.environ.get("GATEWAY_WORKER_COUNT") or 1000)
 
 # the gateway server that should be used (supported: hypercorn, twisted dev: werkzeug)
-GATEWAY_SERVER = os.environ.get("GATEWAY_SERVER", "").strip() or "hypercorn"
+GATEWAY_SERVER = os.environ.get("GATEWAY_SERVER", "").strip() or "twisted"
 
 # IP of the docker bridge used to enable access between containers
 DOCKER_BRIDGE_IP = os.environ.get("DOCKER_BRIDGE_IP", "").strip()


alexrashed:
FYI: Changed the PR to draft and changed the milestone to `3.5`, because we found some issues with downstream projects.

thrau:
Following up on our discussion @dfangl, I found that there are at least two checks in the stack trace whether the data buffer is empty or not:

* https://github.com/twisted/twisted/blob/02a2b658cd1ade5d7f41f97d898913686313e615/src/twisted/web/http.py#L1251-L1255
* https://github.com/twisted/twisted/blob/02a2b658cd1ade5d7f41f97d898913686313e615/src/twisted/internet/abstract.py#L359-L363

The switch makes perfect sense to me. Why write anything when there's no data? While we could forego these conditions, I found no way to trigger a write buffer flush. I found [this SO post](https://stackoverflow.com/questions/28899304/flushing-twisteds-write-buffer) on the topic, where one of the maintainers also suggest you shouldn't be talking to the reactor in this way in the first place.

So in summary I would say this is not a problem with the server, but rather in the particular downstream implementation that relies on the assumption that a buffer will be flushed when you write empty bytes to it.

My recommendation is therefore to fix the particular downstream implementation (or even just the test) to make it more compatible with different IO frameworks.

alexrashed:
@thrau @bentsku The issues in our downstream projects have been addressed (in a backwards compatible manner) and there have been multiple runs validating that the fix works properly.
In my opinion this PR is ready to be merged! 🥳 🚀

bentsku (approved):
LGTM! Nice to see everything addressed, thanks a lot for fixing it! 🚀 

Can't wait to see all the nice performance increases showed in #9834!

PR #10707, opened 2024-04-22
[SFN] Fix TaskToken Creation Logic

MEPalma:
## Motivation
Currently the SFN v2 interpreter runs into States.Runtime errors whenever a `waitForTaskToken` task does not include a `$$.Task.Token` field. This is due to the TaskToken not being created. Furthermore, the TaskToken is created whenever a parameter specifies a `$$.Task.Token` value, regardless of the task resource specifying a `waitForTaskToken` callback mechanism. This PR addresses such issues by moving the `TaskToken` creation logic to the state object instead of the Parameters object.

## Changes
- Move `TaskToken` creation logic to the callback state object
- Add negative tests for this issue, and multiple `TaskToken` invocations


localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_callback.py
```diff
@@ -140,6 +140,17 @@ def _from_error(self, env: Environment, ex: Exception) -> FailureEvent:
...
+    def _eval_body(self, env: Environment) -> None:
+        # Generate a TaskToken uuid within the context object, if this task resources has a waitForTaskToken condition.
+        if self._is_condition() and self.resource.condition == ResourceCondition.WaitForTaskToken:
+            task_token = env.context_object_manager.update_task_token()
+            env.callback_pool_manager.add(task_token)
+
+        super()._eval_body(env=env)
+
+        # Ensure the TaskToken field is reset, as this is only available during waitForTaskToken task evaluations.
+        env.context_object_manager.context_object["Task"] = None
```
joe4dev, line 152: Does setting it to `None` have the same effect as deleting it from the dict? (just wondering)

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_callback.py
```diff
@@ -140,6 +140,17 @@ def _from_error(self, env: Environment, ex: Exception) -> FailureEvent:
             return self._get_callback_outcome_failure_event(env=env, ex=ex)
         return super()._from_error(env=env, ex=ex)
 
+    def _eval_body(self, env: Environment) -> None:
+        # Generate a TaskToken uuid within the context object, if this task resources has a waitForTaskToken condition.
```
joe4dev, line 144: nit: it might be helpful to include more references to AWS docs for better traceability. For example: https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html#connect-wait-token

joe4dev (approved):
LGTM 👍

localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_callback.py
MEPalma: It has the same affect, however I agree it would be better to have the field missing to remove ambiguity. I have changed the annotation from Optional to NotRequired, and pop this field at this depth. I think it's a lot clearer now, and would like to expand this pattern in other areas of the interpreter

PR #10710, opened 2024-04-23
Add custom Lambda runtimes endpoint

joe4dev:
## Motivation

The LocalStack Webapp has no way of determining which Lambda runtimes are supported by a given LocalStack instance and currently hardcodes them in the user interface. A metadata endpoint could allow for dynamic querying (with a fallback). See discussion in https://github.com/localstack/localstack-web/pull/1309

## Changes

* Add a `/_aws/lambda/runtimes` endpoint to query the supported Lambda runtimes. The routes are added and removed through the Lambda provider.
* This metadata API should be available before the Lambda provider is loaded and therefore requires custom routing.

## Testing

```sh
curl 'http://localhost:4566/_aws/lambda/runtimes'
```
> {"Runtimes": ["provided.al2023", "java17", "dotnet6", "nodejs18.x", "nodejs20.x", "java21", "nodejs16.x", "python3.9", "python3.8", "python3.10", "python3.12", "ruby3.3", "java11", "dotnet8", "python3.11", "ruby3.2", "provided.al2", "java8.al2"]}

```sh
curl http://localhost.localstack.cloud:4566/_aws/lambda/runtimes?filter=all
```
> {"Runtimes": ["java17", "nodejs16.x", "python3.7", "ruby3.3", "java11", "java8", "nodejs12.x", "java21", "ruby3.2", "dotnet6", "nodejs20.x", "python3.9", "python3.8", "python3.10", "provided.al2", "dotnetcore3.1", "nodejs14.x", "python3.12", "go1.x", "provided", "dotnet8", "java8.al2", "provided.al2023", "nodejs18.x", "python3.11", "ruby2.7"]}

```sh
curl http://localhost.localstack.cloud:4566/_aws/lambda/runtimes?filter=supported\&filter=deprecated
```
> {"Runtimes": ["provided.al2023", "ruby2.7", "provided.al2", "python3.10", "nodejs14.x", "dotnet6", "nodejs12.x", "nodejs20.x", "dotnet8", "java21", "python3.8", "java8", "java8.al2", "provided", "ruby3.3", "ruby3.2", "python3.9", "nodejs16.x", "nodejs18.x", "dotnetcore3.1", "java17", "java11", "go1.x", "python3.11", "python3.12", "python3.7"]}

sidenote: multiple comma-separated filters such as `?filter=supported,deprecated` are currently not handled

## Discussion

~~Is the hardcoded `/_aws/lambda/` path a good idea in the data plane service detector introduced in https://github.com/localstack/localstack/pull/10682 by @thrau ?~~
-> Not necessary anymore after Alex's suggestion

### DIFF
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -325,6 +325,11 @@ def determine_aws_service_model_for_data_plane(
         services = services or get_service_catalog()
         return services.get(*custom_host_match)
 
+    # Custom service-specific endpoints that should be available before the service is loaded
+    if request.path.startswith("/_aws/lambda/"):
+        services = services or get_service_catalog()
+        return services.get("lambda")
+
 
 def determine_aws_service_model(
     request: Request, services: ServiceCatalog = None
--- a/localstack/aws/protocol/service_router.py
+++ b/localstack/aws/protocol/service_router.py
@@ -325,11 +325,6 @@ def determine_aws_service_model_for_data_plane(
         services = services or get_service_catalog()
         return services.get(*custom_host_match)
 
-    # Custom service-specific endpoints that should be available before the service is loaded
-    if request.path.startswith("/_aws/lambda/"):
-        services = services or get_service_catalog()
-        return services.get("lambda")
-
 
 def determine_aws_service_model(
     request: Request, services: ServiceCatalog = None
--- a/localstack/services/lambda_/custom_endpoints.py
+++ b/localstack/services/lambda_/custom_endpoints.py
@@ -1,0 +1,31 @@
+import urllib.parse
+from typing import List, TypedDict
+
+from rolo import Request, route
+
+from localstack.aws.api.lambda_ import Runtime
+from localstack.constants import TRUE_STRINGS
+from localstack.services.lambda_.runtimes import (
+    ALL_RUNTIMES,
+    DEPRECATED_RUNTIMES,
+    SUPPORTED_RUNTIMES,
+)
+
+
+class LambdaRuntimesResponse(TypedDict, total=False):
+    Runtimes: List[Runtime]
+
+
+class LambdaCustomEndpoints:
+    @route("/_aws/lambda/runtimes")
+    def runtimes(self, request: Request) -> LambdaRuntimesResponse:
+        """This metadata endpoint needs to be loaded before the Lambda provider.
+        It can be used by the Webapp to query supported Lambda runtimes of an unknown LocalStack version."""
+        # Query parameter values are all lists. Example: { "all": ["true"] }
+        query_params = urllib.parse.parse_qs(request.environ["QUERY_STRING"])
+        if query_params.get("all", ["false"])[0] in TRUE_STRINGS:
+            return LambdaRuntimesResponse(Runtimes=ALL_RUNTIMES)
+        elif query_params.get("deprecated", ["false"])[0] in TRUE_STRINGS:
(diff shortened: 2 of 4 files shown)


localstack/aws/protocol/service_router.py
```diff
@@ -325,6 +325,11 @@ def determine_aws_service_model_for_data_plane(
         services = services or get_service_catalog()
         return services.get(*custom_host_match)
 
+    # Custom service-specific endpoints that should be available before the service is loaded
+    if request.path.startswith("/_aws/lambda/"):
```
joe4dev, line 329: This method gets executed for every single request against LocalStack!

localstack/services/lambda_/provider.py
```diff
@@ -308,7 +313,11 @@ def on_after_init(self):
         self.router.register_routes()
         get_runtime_executor().validate_environment()
 
+    def on_before_start(self):
+        self.custom_router_rules = ROUTER.add(LambdaCustomEndpoints())
```
alexrashed, line 317: Since this endpoint does not need a full service loading, this could - instead of adding these endpoints when loading the service - be done in an init hook (similar to `localstack.services.lambda.plugins.validate_configuration`).
This would help you avoid the data plane service router changes, because there's no need for a service detection.

localstack/services/lambda_/custom_endpoints.py
```diff
@@ -0,0 +1,31 @@
...
+    @route("/_aws/lambda/runtimes")
+    def runtimes(self, request: Request) -> LambdaRuntimesResponse:
+        """This metadata endpoint needs to be loaded before the Lambda provider.
+        It can be used by the Webapp to query supported Lambda runtimes of an unknown LocalStack version."""
+        # Query parameter values are all lists. Example: { "all": ["true"] }
+        query_params = urllib.parse.parse_qs(request.environ["QUERY_STRING"])
+        if query_params.get("all", ["false"])[0] in TRUE_STRINGS:
+            return LambdaRuntimesResponse(Runtimes=ALL_RUNTIMES)
+        elif query_params.get("deprecated", ["false"])[0] in TRUE_STRINGS:
+            return LambdaRuntimesResponse(Runtimes=DEPRECATED_RUNTIMES)
```
alexrashed, line 29: These query parameters are a bit misleading for me, what happens if both are set to `true`?
I would suggest using a single query parameter `filter` with different values.

alexrashed (commented):
Nice to see service specific APIs exposing metadata which can be used by other consumers, and also our webapp!
I only added some comments on the query params and the registration of the routes :)

localstack/aws/protocol/service_router.py
joe4dev: Not needed anymore 🧹

localstack/services/lambda_/provider.py
joe4dev: Awesome 💯  That's what I was looking for. Thanks for the suggestion @alexrashed 🙏 
Migrated to hooks now 👍

localstack/services/lambda_/custom_endpoints.py
joe4dev: Great input 👍 

The new API is now also much more flexible 🙌

joe4dev: nit: `filter` might be a bit misleading with `all` because it adds rather than filters but we win the convenience that the default endpoint returns what's most useful.

alexrashed (approved):
Thanks for addressing the comments! Looks good to me! This will allow the platform (after they detected that a compatible version is running) fetch the compatible runtimes directly! 🚀

Issue #10712, opened 2024-04-24
bug: DELETE s3 returns empty response with content-type: application/XML breaking the client


vdayanand:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Similar issue: https://github.com/localstack/localstack/issues/4324

Delete request:
```
DELETE /bucket.123-venk/Project.toml HTTP/1.1
User-Agent: AWS.jl/1.0.0
Host: localhost
x-amz-date: xxx
x-amz-content-sha256: xxx
Content-MD5: xxx
x-amz-security-token: xxx
Accept: */*
Content-Length: 0
Accept-Encoding: gzip
```

Delete response:
```
HTTP/1.1 204 
Content-Type: application/xml
x-amz-request-id: xxx
x-amz-id-2: cas
Connection: close
date: Wed, 24 Apr 2024 06:20:18 GMT
server: hypercorn-h11
content-length: 0
```
Note the response is empty given content-type is `application/xml`. Empty response is not a valid XML and this breaks the AWS client.

### Expected Behavior

This is what I would expect (simllar to Amazon s3)
Delete request:
```
DELETE /bucket.123-venk/Project.toml HTTP/1.1
User-Agent: AWS.jl/1.0.0
Host: s3.us-east-1.amazonaws.com␍
x-amz-date: xxx
x-amz-content-sha256: xxx
Content-MD5: xxx
x-amz-security-token: xxx
Accept: */*
Content-Length: 0
Accept-Encoding: gzip

```
Delete response:
```
HTTP/1.1 204 No Content
x-amz-id-2: xxx
x-amz-request-id: xxx
Date: Wed, 24 Apr 2024 06:19:17 GMT
Server: AmazonS3
content-length: 0
```

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

### Environment

```markdown
- OS: macOS (arm64-apple-darwin22.4.0) 
- LocalStack: 3.3.0
```

bentsku:
Hello @vdayanand and thanks for your report!

Having the `Content-Type` set for a payload with a content length of 0 should technically still be in line with HTTP specs, see this StackOverflow post: https://stackoverflow.com/questions/29784398/should-content-type-header-be-present-when-the-message-body-is-empty

I believe this is an issue with the client, it should not try to decode an empty message body.

[...]

PR #10717, opened 2024-04-24
fix headers for empty S3 responses

bentsku:
## Motivation
As reported by #10712, LocalStack would return some `Content-Type` headers when AWS would not. Even though this was still in line with HTTP specs, some client were relying on this behavior.

We could not put this in the serializer, as the serializer handler calls `Response.update_from`, which is only additive. We cannot remove headers from this, and a default `Content-Type` is set at the beginning of the handler chain. 

## Changes
- create a new simple handler for S3 responses, which will remove some headers to be in line with AWS
- modify some tests to validate the new changes

_fixes #10712_

### DIFF
--- a/localstack/aws/protocol/serializer.py
+++ b/localstack/aws/protocol/serializer.py
@@ -1501,7 +1501,6 @@ class S3ResponseSerializer(RestXMLResponseSerializer):
                 request_id,
             )
         self._serialize_content_type(response, shape, shape_members, mime_type)
-        self._prepare_additional_traits_in_response(response, operation_model, request_id)
 
     def _serialize_error(
         self,
--- a/localstack/services/s3/provider.py
+++ b/localstack/services/s3/provider.py
@@ -134,7 +134,11 @@ from localstack.aws.api.s3 import (
     WebsiteConfiguration,
 )
 from localstack.aws.forwarder import NotImplementedAvoidFallbackError
-from localstack.aws.handlers import preprocess_request, serve_custom_service_request_handlers
+from localstack.aws.handlers import (
+    modify_service_response,
+    preprocess_request,
+    serve_custom_service_request_handlers,
+)
 from localstack.constants import AWS_REGION_US_EAST_1, DEFAULT_AWS_ACCOUNT_ID
 from localstack.services.edge import ROUTER
 from localstack.services.moto import call_moto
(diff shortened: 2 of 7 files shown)


localstack/aws/protocol/serializer.py
```diff
@@ -1501,7 +1501,6 @@ def _serialize_response(
                 request_id,
             )
         self._serialize_content_type(response, shape, shape_members, mime_type)
-        self._prepare_additional_traits_in_response(response, operation_model, request_id)
```
bentsku, line 1504: I removed this from the S3ResponseSerializer, because `_prepare_additional_traits_in_response` is already called once in `localstack.aws.protocol.serializer.ResponseSerializer.serialize_to_response`: it first calls `_serialize_response` then `_prepare_additional_traits_in_response`.

thrau (commented):
Nicely done containing the necessary logic! It's clear why this must be in a response finalizer. Even though the header handling is bad, at least it's very isolated now to s3!

One thing I'm worried about and like to get clarity on is what happens to requests that are mistakenly detected as s3 requests. Currently, basically all requests to extensions for instance are detected as s3 requests because the service name parser sits before the edge route serving. Would those requests not be augmented in the same way?

cloutierMat (commented):
Clean use of the chain handlers. Thanks for tagging me in here! That should prove useful soon! :+1:

bentsku:
> Even though the header handling is bad, at least it's very isolated now to s3!

Not sure what you mean by that, but technically this handler still follow HTTP specs, it shouldn't break anything downstream. A 204 should not have a body (might happen somehow still...) and should not have a content-type (SHOULD in the HTTP specs... so can still happen, but this is following the specs).

Same for a 200 with `PUT`, maybe I could make the check stricter and set it to only `Content-Lenght == 0` and not `None`, this should still work.

> One thing I'm worried about and like to get clarity on is what happens to requests that are mistakenly detected as s3 requests. Currently, basically all requests to extensions for instance are detected as s3 requests because the service name parser sits before the edge route serving. Would those requests not be augmented in the same way?

Like I said, it should still pretty standard behavior, and I think our handler chain adding `Content-Type: text/plain` by default is kinda non-standard when there is no body. 
For the fact that extensions are detected as S3 requests, well, that's a tricky one? I can always add a check if the service has an operation set, so that it's only truly S3 requests? I think the operation parser is after the edge routes, right?

thrau (approved):
thanks, this makes it much more contained to S3 api operations which makes it easier to understand the impact :+1:

PR #10721, opened 2024-04-25
CFN_PER_RESOURCE_TIMEOUT env variable

lakkeger:
## Motivation
LocalStack's CloudFormation implementation has a hard limit of 2.5 mins to deploy a single resource which is not always enough for bigger resource deployments like nested stacks.

## Changes
- Add CFN_PER_RESOURCE_TIMEOUT env variable
- Increase default value to 300s

### DIFF
--- a/localstack/config.py
+++ b/localstack/config.py
@@ -1041,6 +1041,9 @@ PARITY_AWS_ACCESS_KEY_ID = is_env_true("PARITY_AWS_ACCESS_KEY_ID")
 # Show exceptions for CloudFormation deploy errors
 CFN_VERBOSE_ERRORS = is_env_true("CFN_VERBOSE_ERRORS")
 
+# Set the timeout for CloudFormation resources
+CFN_PER_RESOURCE_TIMEOUT = int(os.environ.get("CFN_PER_RESOURCE_TIMEOUT") or 150)
+
 # How localstack will react to encountering unsupported resource types.
 # By default unsupported resource types will be ignored.
 # EXPERIMENTAL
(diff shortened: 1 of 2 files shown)


localstack/config.py
```diff
@@ -1041,6 +1041,9 @@ def populate_edge_configuration(
 # Show exceptions for CloudFormation deploy errors
 CFN_VERBOSE_ERRORS = is_env_true("CFN_VERBOSE_ERRORS")
 
+# Set the timeout for CloudFormation resources
+CFN_PER_RESOURCE_TIMEOUT = int(os.environ.get("CFN_PER_RESOURCE_TIMEOUT") or 150)
```
simonrw, line 1045: We should also [document this](https://github.com/localstack/docs).

localstack/config.py
```diff
@@ -1041,6 +1041,9 @@ def populate_edge_configuration(
 # Show exceptions for CloudFormation deploy errors
 CFN_VERBOSE_ERRORS = is_env_true("CFN_VERBOSE_ERRORS")
 
+# Set the timeout for CloudFormation resources
```
simonrw, line 1044:
```suggestion
# Set the timeout to deploy each individual CloudFormation resource
```

localstack/config.py
simonrw, line 1045: We should probably increase the default as well since we know that nested stacks are already a problem. Perhaps `300` (5 minutes)?

localstack/services/cloudformation/resource_provider.py
```diff
@@ -397,11 +398,13 @@ def deploy_loop(
         self,
         resource: dict,
         raw_payload: ResourceProviderPayload,
-        max_iterations: int = 30,
+        max_timeout: int = config.CFN_PER_RESOURCE_TIMEOUT,
         sleep_time: float = 5,
     ) -> ProgressEvent[Properties]:
         payload = copy.deepcopy(raw_payload)
 
+        max_iterations = ceil(max_timeout/sleep_time)
```
simonrw, line 406: We may want to ensure at least 2 iterations...

```suggestion
        max_iterations = max(ceil(max_timeout/sleep_time), 2)
```

simonrw (approved):
Awesome nice work! It is good to have this configurable for large and deeply nested stacks, thanks!

PR #10728, opened 2024-04-26
Fix SubtypesInstanceManager not loading subtypes lazily

viren-nadkarni:
## Background

`SubtypesInstanceManager` is supposed to lazily create a singleton instance of subtypes and return it. However a bug caused all subtypes to be instantiated eagerly if any subtype was requested.

This was identified when the Libvirt EC2 VM manager was being instantiated in the CI (which would fail since CI runners do not have a Libvirt socket) despite not being enabled explicitly in code or configuration.

## Changes

Fix `SubtypesInstanceManager` so that only the requested subtype is instantiated.


whummer (approved):
Great catch, thanks for fixing this @viren-nadkarni ! 🙌

PR #10730, opened 2024-04-26
Refactor: Events v2: Move existing provider code to v1 folder

maxhoheiser:
## Motivation
To improve the readability of events service folder by separating v1 and v2 providers not via file suffixes but dedicated folders. 
There are also dependencies in ext that rely on v1 and v2 provider files, these are updated in a companion PR in ext: https://github.com/localstack/localstack-ext/pull/2907

## Changes
Move existing (v1) provider and related files to v1 folder.
Remove suffix for v2 provider and related files.


localstack/services/events/v1/models.py
```diff
@@ -0,0 +1,11 @@
+from typing import Dict
+
+from localstack.services.stores import AccountRegionBundle, BaseStore, LocalAttribute
+
+
+class EventsStore(BaseStore):
```
joe4dev, line 6: 💡 Please check whether persistence still works before merging after moving the model to a new location.

maxhoheiser: This should be done by the persistency test in the companion branch in pro: https://github.com/localstack/localstack-ext/pull/2907
Since all the tests are green including the persistency tests, it should still work after moving.

localstack/services/events/rule.py
maxhoheiser, line 1: Although it is used with the v2 implementation I descided to move also the ruler into the v1 folder and only move the refactored version out again into the main folder.

PR #10732, opened 2024-04-26
fix serialization for sqs http calls

baermat:
## Motivation
When performing API calls for sqs like "ReceiveMessage" directly via HTTP outside the CLI, the response was malformed. It was malformed in general, and had an additional error for json messages. This PR addresses these issues.

## Changes
- fixes the json parsing by taking into account the escaping of a potential json dump

## TODO

What's left to do:

- [x] clarify if the asymmetry between the marker conversions has a reason. It seems odd that the code goes to great lengths explaining the reason for all the conversion, only to have these two ways differ -> not doing this results in strange conversion errors for \r on non-json messages.
- [ ] while tackling this, I found another bug: the response for ReceiveMessage is potentially further malformed. A usual cli call returns ```{Messages:[{<msg_1>}, {<msg_2>}, ...]}``` but we return ```{Message:{<msg_1>}}``` if there is one message, and ```{Message:[{<msg_1>}, {<msg_2>}, ...]}``` if there are more than one (note the missing plural s of Message). AWS' response is also different with a "direct" http call, but it is only the case that is different ("Messages" vs "messages"). How to proceed here? It seems to be an error in the automatic response creation of LS

### DIFF
--- a/localstack/aws/protocol/serializer.py
+++ b/localstack/aws/protocol/serializer.py
@@ -1654,12 +1654,13 @@ class SqsQueryResponseSerializer(QueryResponseSerializer):
         # Undo the second escaping of the &
         # Undo the second escaping of the carriage return (\r)
         if mime_type == APPLICATION_JSON:
+            # At this point the json was already dumped and escaped, so we replace directly.
             generated_string = generated_string.replace(r"__marker__\"__marker__", r"\"").replace(
                 "__marker__-r__marker__", r"\r"
             )
         else:
-            generated_string = generated_string.replace('__marker__"__marker__', '"').replace(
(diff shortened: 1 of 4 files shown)


alexrashed (approved):
Whoof, that must have been terrible to pin down! Really great catch! Thanks a lot for tackling this! 🦸🏽 
The implementation looks good (as good as it can be on my initial hacky solution), the AWS validated test is great to avoid a regression in the future, and to have a kick-start when tackling the other issue you detected. 🚀

thrau (approved):
Tested and works great, also via the dev endpoint. Thanks @baermat, nice job tracking this down! :100:

Issue #10606, opened 2024-04-05
enhancement request: Add AWS:EC2::PrefixList and AWS::EC2::VPCEndpoint to CloudFormation


Relenna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

Hi, I would like to request adding possibility to create 'AWS:EC2::PrefixList' and 'AWS::EC2::VPCEndpoint' resources via CloudFormation

pinzon:
The solution for this issue has been merged. Please make sure to pull the latest image.

PR #10735, opened 2024-04-26
add resource providers for vpc endpoint and prefix list

pinzon:
## Motivation
Addresses #10606. This PR adds new resource providers for the EC2 service

## Changes
- New resource provider for AWS::EC2::PrefixList
- New resource provider for AWS::EC2::VpcEndpoint

## Testing
- new deployment test for each resource


localstack/services/ec2/resource_providers/aws_ec2_prefixlist.py
```diff
@@ -0,0 +1,135 @@
...
+        if "Tags" in create_params:
+            create_params["TagSpecifications"] = [
+                {"ResourceType": "prefix-list", "Tags": create_params.pop("Tags")}
+            ]
+
+        response = request.aws_client_factory.ec2.create_managed_prefix_list(**create_params)
+        model["Arn"] = response["PrefixList"]["PrefixListId"]
+        model["OwnerId"] = response["PrefixList"]["OwnerId"]
+        model["PrefixListId"] = response["PrefixList"]["PrefixListId"]
+        model["Version"] = response["PrefixList"]["Version"]
```
simonrw, line 88: In the future I think we should _always_ use the asynchronous form of this method (i.e. the default from the scaffolding) to make sure the resource provider ensures that resource has created correctly. It's not just about time taken to create the resource, but using a resource provider in CFn should be atomic i.e. when asked to create a resource, either the resource is created successfully or it fails. In resource providers we should make sure the resource has been created by using a `describe_*` or `get_*` method.

Not a blocker, but something to consider for the future.

cc @dominikschubert

localstack/services/ec2/resource_providers/aws_ec2_prefixlist.py
```diff
@@ -0,0 +1,135 @@
...
+
+        IAM permissions required:
+          - EC2:CreateManagedPrefixList
+          - EC2:DescribeManagedPrefixLists
+          - EC2:CreateTags
+
+        """
+        model = request.desired_state
+        create_params = util.select_attributes(
+            model, ["PrefixListName", "Entrie", "MaxEntries", "AddressFamily", "Tags"]
```
simonrw, line 76: typo

```suggestion
            model, ["PrefixListName", "Entries", "MaxEntries", "AddressFamily", "Tags"]
```

localstack/services/ec2/resource_providers/aws_ec2_prefixlist.py
```diff
@@ -0,0 +1,135 @@
...
+        Delete a resource
+
+        IAM permissions required:
+          - EC2:DeleteManagedPrefixList
+          - EC2:DescribeManagedPrefixLists
+        """
+        request.aws_client_factory.ec2.delete_managed_prefix_list(
+            PrefixListId=request.desired_state["PrefixListId"]
+        )
+        return ProgressEvent(status=OperationStatus.SUCCESS, resource_model=None)
```
simonrw, line 119: Similar to above, we should ensure the resource has been deleted with a call to `describe_*` or `get_*`.

simonrw (approved):
Provider looks good for now. I mentioned we should probably use the async form of the provider methods _always_ since it upholds the CFn resource provider contract of atomicity. Not a blocker, we might want to do a pass over "non-conforming" providers in the future once we have more accurate deployment order

pinzon:
@simonrw thank you for the comments. I addressed them all and made the improvements to the resource providers. I'll merge once is green again.

PR #10750, opened 2024-04-30
[SFN] Improve Handling of Empty SendTaskFailure Calls

MEPalma:
## Motivation
Currently the StepFunctions v2 interpreter is unable of handling empty SentTaskFailure api calls. These are SendTaskFailure messages for waitForTaskToken tasks which do not provide an error string. Such calls would result in internal States.Runtime errors, and misleading error messages such as `Custom Error Names MUST NOT begin with the prefix 'States.', got 'State.Runtime'`.  This PR adds snapshot tests replicating this occurrence and add support for CustomErrorNames to have nullable error strings. It also identifies need for future work on StateTaskAbort events logging.

## Changes
- Nullable error strings in CustomErrorName objects
- Related handling of null error strings
- Snapshot tests

### DIFF
--- a/localstack/services/stepfunctions/asl/component/common/catch/catcher_decl.py
+++ b/localstack/services/stepfunctions/asl/component/common/catch/catcher_decl.py
@@ -75,8 +75,8 @@ class CatcherDecl(EvalComponent):
             )
         spec_event_details: dict = list(failure_event.event_details.values())[0]
         # If no cause or error fields are given, AWS binds an empty string; otherwise it attaches the value.
-        error = str() if "error" not in spec_event_details else spec_event_details["error"]
-        cause = str() if "cause" not in spec_event_details else spec_event_details["cause"]
+        error = spec_event_details.get("error", "")
+        cause = spec_event_details.get("cause", "")
         # Stepfunctions renames these fields to capital in this scenario.
         return {
             "Error": error,
--- a/localstack/services/stepfunctions/asl/component/common/catch/catcher_decl.py
+++ b/localstack/services/stepfunctions/asl/component/common/catch/catcher_decl.py
@@ -74,8 +74,9 @@ class CatcherDecl(EvalComponent):
                 f"Internal Error: invalid event details declaration in FailureEvent: '{failure_event}'."
             )
         spec_event_details: dict = list(failure_event.event_details.values())[0]
-        error = spec_event_details["error"]
-        cause = spec_event_details.get("cause") or ""
+        # If no cause or error fields are given, AWS binds an empty string; otherwise it attaches the value.
+        error = str() if "error" not in spec_event_details else spec_event_details["error"]
+        cause = str() if "cause" not in spec_event_details else spec_event_details["cause"]
         # Stepfunctions renames these fields to capital in this scenario.
         return {
             "Error": error,
--- a/tests/aws/services/stepfunctions/v2/callback/test_callback.py
+++ b/tests/aws/services/stepfunctions/v2/callback/test_callback.py
@@ -750,7 +750,13 @@ class TestCallback:
             task_token = message_body["TaskToken"]
             aws_client.stepfunctions.send_task_failure(taskToken=task_token)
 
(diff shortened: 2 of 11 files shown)


localstack/services/stepfunctions/asl/component/common/catch/catcher_decl.py
```diff
@@ -74,8 +74,9 @@ def _extract_error_cause(failure_event: FailureEvent) -> dict:
                 f"Internal Error: invalid event details declaration in FailureEvent: '{failure_event}'."
             )
         spec_event_details: dict = list(failure_event.event_details.values())[0]
-        error = spec_event_details["error"]
-        cause = spec_event_details.get("cause") or ""
+        # If no cause or error fields are given, AWS binds an empty string; otherwise it attaches the value.
+        error = str() if "error" not in spec_event_details else spec_event_details["error"]
+        cause = str() if "cause" not in spec_event_details else spec_event_details["cause"]
```
dominikschubert, line 79: Is there a particular reason you've used `str()` over a simple empty string literal `""`? If not I'd argue the latter is a bit more clearer to the reader and should also avoid an unnecessary function call

dominikschubert: An alternative would be for both just doing a `.get("...", "")` which is a bit more "pythonic" and therefore clearer in purpose

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -697,3 +698,69 @@ def test_sqs_wait_for_task_token_no_token_parameter(
...
+        request,
+    ):
+        if (
+            request.node.name
+            == "test_sqs_failure_in_wait_for_task_tok_no_error_field[SQS_PARALLEL_WAIT_FOR_TASK_TOKEN]"
+        ):
+            # TODO: The conditions in which TaskStateAborted error events are logged requires further investigations.
+            #  These appear to be logged for Task state workers but only within Parallel states. The behaviour with
+            #  other 'Abort' errors should also be investigated.
+            pytest.skip("Investigate occurrence logic of 'TaskStateAborted' errors")
```
dominikschubert, line 725: Is this also not working against AWS or just localstack?

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -697,3 +698,69 @@ def test_sqs_wait_for_task_token_no_token_parameter(
...
+                    QueueUrl=queue_url, MaxNumberOfMessages=1
+                )
+                return receive_message_response["Messages"][0]["Body"]
+
+            message_body_str = retry(_get_message_body, retries=60, sleep=1)
+            message_body = json.loads(message_body_str)
+            task_token = message_body["TaskToken"]
+            aws_client.stepfunctions.send_task_failure(taskToken=task_token)
+
+        threading.Thread(target=_empty_send_task_failure_on_sqs_message, args=()).start()
```
dominikschubert, line 753: Please make sure any created thread is a) named and b) ideally set as a daemon thread to avoid blocking the shutdown and improve debugging.

dominikschubert (changes requested):
Mostly LGTM! Definitely another interesting edge case.

Changes requested is mostly referring to the thread handling which I'd like to verify before a final approval.

tests/aws/services/stepfunctions/v2/callback/test_callback.py
MEPalma: It works against AWS. For LocalStack I first want to collect more insight into when it is triggered before updating the interpreter

dominikschubert: nit: You could in this case extend the if case by a check against `is_aws_cloud()` as well to make sure we only skip it when running against LS

dominikschubert (approved):
LGTM, thanks for addressing the previous points. Minor nit regarding the test skip, but generally good to go from my side :)

PR #10757, opened 2024-05-02
Parametrize test selection scripts for usage in dependent repositories

silv-io:
## Motivation

Using the test selection scripts in repositories with slightly different requirements is difficult because of a lack of parametrization.

## Changes

This PR adds parametrization in regards to repository name, opt_in dictionary and matching rules for test selection.

/cc @alexrashed 

## Testing

The test selection should still work as before:

1. Create a dummy change commit where you only change a simple thing in one service
2. Run
`python -m localstack/testing/testselection/scripts/generate_test_selection_from_commits.py . <HEAD~> <HEAD> test-selection`

Output should be only the service tests (if it's opted in)

## TODO

What's left to do:

- [x] parametrize generic testing rule

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -7,11 +7,11 @@ on:
         required: false
         type: boolean
         default: false
-      disableTestSelection:
-        description: 'Disable Test Selection'
+      enableTestSelection:
+        description: 'Enable Test Selection'
         required: false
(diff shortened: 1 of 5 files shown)


alexrashed (approved):
LGTM! 👍🏽

dominikschubert (approved):
Nice further layer of parametrization :+1: 
And thanks for extending the type hints :rocket:

PR #10758, opened 2024-05-02
Fix: Firehose: Drop keys in destinations description not in respective return types

maxhoheiser:
## Motivation
The operation `list_delivery_streams` returns unwanted keys for multiple destination descriptions.  
Destination descriptions are loaded from the store and backend, and can contain keys that are not defined by the respective return types and thus not expected to be returned, e.g. `RedshiftDestinationDescription` contains a Password parameter.

As observed in different CI runs, this can lead to unnecessary log messages:
```
Response object RedshiftDestinationDescription contains a member which is not specified: Password
AWS firehose.DescribeDeliveryStream => 200
POST localhost:4566/
```

## Changes
 This pr drops parameters in all supported destinations that are not defined by the respective return type definitions.


pinzon (approved):
LGTM

PR #10761, opened 2024-05-03
decrease test logging verbosity

alexrashed:
## Motivation
One of the most-often heard feedback when it comes to the usability of our CI pipelines is that the logs are just overwhelming.
This PR tries to find the right balance of logging when executing our tests such that:
- We still get the information we need when digging into test failures.
- We are not completely overwhelmed with the amount of logs.

This is why this parameter tries to tweak the settings such that - by default - the only logs which are of the level `warn` and above will be shown in the logs as long as the tests are successful.
If a test fails, the debug logs for this failing test are shown.

## Changes
- Disables the thread leakage detector. /cc @dominikschubert @dfangl 
- Disables `log_cli` / `-s` for `pytest` to avoid direct verbose logging to stdout.
- Enables colors in all GitHub actions (see https://github.com/pytest-dev/pytest/pull/7462).
- Unifies the test command invocation use the `test` target everywhere.
- Adds parameters to the pipelines to allow manually setting the log level.
  - For GitHub Actions this is a choice-type workflow dispatch input (i.e. can be selected when manually triggering the workflow).
  - For CircleCI the parameter `PYTEST_LOGLEVEL` can be set when triggering the pipeline run manually.
  - This parameter is _not_ used for the CLI tests (GitHub Actions) and for the unit tests (in CircleCI) due to https://github.com/pallets/click/issues/824.

## Testing
- Analyze triggered pipeline runs if the output is more usable than it is right now.

## TODO
- [x] Find the right default settings.
- [x] Add a flag such that verbose logging can be enabled again in case we really need detailed logs (f.e. when pinning down memory leaks or other issues that span across multiple tests and might not directly result in test failures).
  - https://github.com/localstack/localstack/actions/runs/8968244829/job/24627268104
  - https://github.com/localstack/localstack/actions/runs/8968248363/job/24627276805
  - https://app.circleci.com/pipelines/github/localstack/localstack/24595/workflows/f51b95ee-1a52-4dd6-9a98-c63f99bd27ae


.github/workflows/tests-cli.yml
```diff
@@ -79,14 +94,15 @@ jobs:
...
+          source .venv/bin/activate
           pip install -e .
           pip install pytest pytest-tinybird
       - name: Run CLI tests
         env:
           PYTEST_ADDOPTS: "${{ env.TINYBIRD_PYTEST_ARGS }}-p no:localstack.testing.pytest.fixtures -p no:localstack_snapshot.pytest.snapshot -p no:localstack.testing.pytest.filters -p no:localstack.testing.pytest.fixture_conflicts -p no:localstack.testing.pytest.validation_tracking -p no:localstack.testing.pytest.path_filter -p no:tests.fixtures -s"
-        run: |
-          python -m pytest tests/cli/
+          TEST_PATH: "tests/cli/"
+        run: make test
```
dominikschubert, line 105: nice unification :+1:

.github/workflows/tests-pro-integration.yml
```diff
@@ -316,9 +330,9 @@ jobs:
           AWS_SECRET_ACCESS_KEY: "test"
           AWS_ACCESS_KEY_ID: "test"
           AWS_DEFAULT_REGION: "us-east-1"
-          PYTEST_LOGLEVEL: debug
           TEST_PATH: "../localstack/tests/aws/"  # TODO: also include tests/integration/
-          PYTEST_ARGS: "${{ env.TINYBIRD_PYTEST_ARGS }}${{ env.TESTSELECTION_PYTEST_ARGS }}--splits 2 --group ${{ matrix.group }} --durations-path ../localstack/.test_durations --capture=no --reruns 2 --junitxml=pytest-junit-community-${{ matrix.group }}.xml"
+          JUNIT_REPORTS_FILE: "pytest-junit-community-${{ matrix.group }}.xml"
```
dominikschubert, line 334: also nice refactoring, reducing the amount of CLI args :+1:

tests/conftest.py
```diff
@@ -11,7 +11,6 @@
     "localstack_snapshot.pytest.snapshot",
     "localstack.testing.pytest.filters",
     "localstack.testing.pytest.fixture_conflicts",
-    "localstack.testing.pytest.detect_thread_leakage",
```
dominikschubert, line 14: LGTM, I think IIRC I had a local branch once where I was disabling this too since its just a bit too noisy and not always relevant. Would love to get the opinion of @dfangl  as well, but this wouldn't block the merge anyway since it can be easily re-enabled if needed.

pyproject.toml
```diff
@@ -225,7 +225,7 @@ exclude_lines = [
 ]
 
 [tool.pytest.ini_options]
-log_cli = true
+log_cli = false
```
dominikschubert, line 228: :+1:  this should avoid at least a few weirdly interleaved outputs as well

Makefile
```diff
@@ -211,15 +212,11 @@ docker-cp-coverage:
...
-		DEBUG=$(DEBUG) \
-		DISABLE_BOTO_RETRIES=$(DISABLE_BOTO_RETRIES) \
-		LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC=1 \
-		python -m coverage run $(COVERAGE_ARGS) -m \
-		pytest --durations=10 --log-cli-level=$(PYTEST_LOGLEVEL) -s $(PYTEST_ARGS) $(TEST_PATH))
+	($(VENV_RUN); $(TEST_EXEC) pytest --durations=10 --log-cli-level=$(PYTEST_LOGLEVEL) $(PYTEST_ARGS) $(TEST_PATH))
+
+test-coverage: LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC = 1
+test-coverage: TEST_EXEC = python -m coverage run $(COVERAGE_ARGS) -m
+test-coverage: test	  ## Run automated tests and create coverage report
```
dominikschubert, line 219: TIL
super clean :sparkles:

dominikschubert (approved):
Looks like a huge step towards more readable logs in CI :+1: :rocket: 

There's a few other areas we can tackle in the future to reduce noise and better highlight root causes, etc. but that'll probably be more suited for the general logging effort. :sweat_smile:

tests/conftest.py
dfangl: I agree with it being pretty verbose. I'm fine with disabling it, but we might also selectively enable it if the pytest log level is set to DEBUG?
In any case, I'm fine with it being disabled for now.

pyproject.toml
joe4dev: That affects the log level of pytest in Pycharm as well 😬 . We can overwrite this configuration using `--log-cli-level=DEBUG -s` in the pytest test configuration (template). The `-s` disables all capturing and should help with interleaved outputs.

Issue #10662, opened 2024-04-15
bug: Stepfunction stops working when starting distributed map state second time


jhKone:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm trying to test decently simple step function which can run multiple loops of getting part of data, processing it in distributed state map and doing this until all data has been processed.

Simplied graph of the step function

<img width="428" alt="Screenshot 2024-04-15 at 13 19 38" src="https://github.com/localstack/localstack/assets/128794921/4666a3e7-6c2e-4ce0-a95e-8b391f1f4aaf">

What ends up happening is that the first time this step function goes through the map state it is fine and if it doesn't need to get more data, it ends with SUCCEEDED. 

But if it goes back to the data fetching first lambda, and then starts another distrubuted map state, everything stops after that and nothing happens until the step function time outs. If I change the map state mode to INLINE for localstack, everything works correctly and multiple loops can be run with no problems.

Here first execution is when running map state only once, second what happens  when running it the second time using distributed mode.

<img width="994" alt="Screenshot 2024-04-15 at 13 22 43" src="https://github.com/localstack/localstack/assets/128794921/23c5f9a7-75a9-48d2-a14f-34b000b63377">

Last event I can see in the execution history is MapRunStarted, but then nothing.

<img width="378" alt="Screenshot 2024-04-15 at 13 25 38" src="https://github.com/localstack/localstack/assets/128794921/e7ffef5c-7e2f-4627-9c5e-dd11344460cd">

In docker logs this is the last message I can see, then nothing:

l.s.s.a.c.eval_component   : [ASL] [comp] [StateMap]: '(StateMap| {'comment': None, 'input_path': (InputPath| {'input_path_src': '$'}, 'output_path': (OutputPath| {'output_path': '$'}, 'state_entered_event_type': 'MapStateEntered', 'state_exited_event_type': 'MapStateExited', 'result_path': (ResultPath| {'result_path_src': None}, 'result_selector': None,

...

[...]

PR #10763, opened 2024-05-03
[SFN] Execution of Reentrant Distributed Map States

MEPalma:
## Motivation
Currently the StepFunctions v2 interpreter is unable to execute reentrant distributed map states. This is due to a lack of end of evaluation cleanups, which means the workers activate logic does not start any new workers. This PR addresses such issue and adds relevant snapshot tests. Addresses: #10662

## Changes
- Added member cleanups at the end of distributed map evaluation
- Added relevant snapshot tests

### DIFF
--- a/localstack/services/stepfunctions/asl/component/state/state_execution/state_map/iteration/distributed_iteration_component.py
+++ b/localstack/services/stepfunctions/asl/component/state/state_execution/state_map/iteration/distributed_iteration_component.py
@@ -202,6 +202,8 @@ class DistributedIterationComponent(InlineIterationComponent, abc.ABC):
             raise ex
         finally:
             env.event_history = execution_event_history
+            self._eval_input = None
+            self._workers.clear()
 
         # TODO: review workflow of program stops and maprunstops
         # program_state = env.program_state()
--- a/tests/aws/services/stepfunctions/templates/scenarios/statemachines/map_state_config_distributed_reentrant.json5
+++ b/tests/aws/services/stepfunctions/templates/scenarios/statemachines/map_state_config_distributed_reentrant.json5
@@ -1,0 +1,61 @@
+{
+  "Comment": "MAP_STATE_CONFIG_DISTRIBUTED_REENTRANT",
+  "StartAt": "StartState",
+  "States": {
+    // Populate memory with two fields: number of iterations left, and input values.
+    "StartState": {
+      "Type": "Pass",
+      "Parameters": {
+        "Iterations": 3,
+        "Values.$": "States.ArrayRange(0, 3, 1)"
+      },
+      "Next": "BeforeIteration"
+    },
+    // Prepare the iteration by updating the iterations count.
+    "BeforeIteration": {
+      "Type": "Pass",
+      "Parameters": {
+        "Iterations.$": "States.MathAdd($.Iterations, -1)",
+        "Values.$": "$.Values"
+      },
+      "Next": "IterationBody"
+    },
+    // Run a distributed map state on the values field.
+    "IterationBody": {
+      "Type": "Map",
+      "MaxConcurrency": 1,
+      "ItemsPath": "$.Values",
+      "ItemProcessor": {
+        "ProcessorConfig": {
+          "Mode": "DISTRIBUTED",
+          "ExecutionType": "STANDARD"
+        },
+        "StartAt": "ProcessValue",
+        "States": {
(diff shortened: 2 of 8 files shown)


tests/aws/services/stepfunctions/templates/scenarios/statemachines/map_state_config_distributed_reentrant.json5
```diff
@@ -0,0 +1,61 @@
...
+        "Values.$": "$.Values"
+      },
+      "Next": "IterationBody"
+    },
+    // Run a distributed map state on the values field.
+    "IterationBody": {
+      "Type": "Map",
+      "MaxConcurrency": 1,
+      "ItemsPath": "$.Values",
+      "ItemProcessor": {
```
joe4dev, line 28: This `ItemProcessor` is the only change compared to the legacy version. Maybe a comment could help to point out this difference because I didn't know the difference between legacy and this one without doing a diff.

tests/aws/services/stepfunctions/templates/scenarios/statemachines/map_state_config_distributed_reentrant_lambda.json5
```diff
@@ -0,0 +1,64 @@
...
+      "ItemProcessor": {
+        "ProcessorConfig": {
+          "Mode": "DISTRIBUTED",
+          "ExecutionType": "STANDARD"
+        },
+        "StartAt": "ProcessValue",
+        "States": {
+          "ProcessValue": {
+            "Type": "Task",
+            "Resource": "_tbd_",
```
joe4dev, line 39: That seems to be the main difference here, having the Lambda resource invocation. Consider highlighting the important pieces through comments such that it is easier to spot them within lengthy state machine definitions.

tests/aws/services/stepfunctions/v2/scenarios/test_base_scenarios.py
```diff
@@ -579,6 +579,94 @@ def test_map_state_config_distributed_item_selector(
...
+        aws_client,
+        create_iam_role_for_sfn,
+        create_state_machine,
+        create_lambda_function,
+        sfn_snapshot,
+    ):
+        # Replace MapRunArns with fixed values to circumvent random ordering issues.
+        sfn_snapshot.add_transformer(
+            JsonpathTransformer(
+                jsonpath="$..mapRunArn", replacement="map_run_arn", replace_reference=False
```
joe4dev, line 643: Did you manually check (e.g., using `SNAPSHOT_RAW=1`) that we return the correct ARN in the right format here?

localstack/services/stepfunctions/asl/component/state/state_execution/state_map/iteration/distributed_iteration_component.py
```diff
@@ -202,6 +202,8 @@ def _eval_body(self, env: Environment) -> None:
             raise ex
         finally:
             env.event_history = execution_event_history
+            self._eval_input = None
+            self._workers.clear()
```
joe4dev, line 206: I'm assuming that `self._job_pool.await_jobs()` within `_map_run` ensures that all workers finished execution here, so we can safely clear the `_workers` list.

joe4dev (approved):
LGTM 👍 

I added a small suggestion for a comment to clarify the intent of the test scenarios.

tests/aws/services/stepfunctions/v2/scenarios/test_base_scenarios.py
MEPalma: I can confirm we do produce the correct ARN format, and we do have tests that validate this aspect of MapRuns. However, here I decided to skip this check to avoid crowding the test state with ordering reordering logic. We can however still validate that for each map input the correct output (and events) are computed.

localstack/services/stepfunctions/asl/component/state/state_execution/state_map/iteration/distributed_iteration_component.py
MEPalma: Indeed!

tests/aws/services/stepfunctions/v2/scenarios/test_base_scenarios.py
joe4dev: Reasonable. No need to double-check if we have other tests covering this part 💯

PR #10770, opened 2024-05-06
fixed physical resource id for AWS::Glue::SchemaVersionMetadata

macnev2013:
## Motivation
This PR addresses the issue with the `PhysicalResourceId` format for `AWS::Glue::SchemaVersionMetadata`. Previously, it was generated as `<uuid>-<key>-<value>`, whereas in AWS it should be `<uuid>|<key>|<value>`.

## Changes
Modified `PhysicalResourceId` to align with AWS.

## Testing
- This changes are validated (snapshot tested) in https://github.com/localstack/localstack-ext/pull/584.


dominikschubert (approved):
LGTM. Thanks for adding the quirk. :+1:

PR #10774, opened 2024-05-06
Remove fix to relative persistence path for Kinesis

giograno:
## Motivation
I randomly came across this `FIXME` comment here. The use of the relative path should not be needed, as the upstream [issue](https://github.com/etspaceman/kinesis-mock/issues/554) has been resolved.
I also run the persistence tests in a [companion branch](https://github.com/localstack/localstack-ext/actions/runs/8929367616/job/24527310273?pr=2937).

## Changes
- Removed the use of `os.path.relpath` for the `PERSIST_PATH` env variable.


alexrashed (approved):
LGTM! 🧹 Thanks for cleaning up!

PR #10784, opened 2024-05-07
enable sqs test selection

baermat:
## Motivation
This PR enables test selection introduced in #10301 for SQS. Functionality was tested in https://github.com/localstack/localstack/pull/10785. Thanks Alex for explaining where the missing tests went :smile: /cc @thrau 

## Changes
- adds SQS to the OPT_IN

## TODO

What's left to do:

- [x] most important dependencies on SQS are already covered, but need to check if there are other services that depend on SQS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,7 @@ commands:
                 command: |
                   if [[ -n "$CI_PULL_REQUEST" ]] ; then
                     echo "export TESTSELECTION_PYTEST_ARGS='--path-filter=target/testselection/test-selection.txt '" >> $BASH_ENV
+                    cat target/testselection/test-selection.txt
                   fi
       - run:
           name: Setup Environment Variables
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,6 @@ commands:
                 command: |
                   if [[ -n "$CI_PULL_REQUEST" ]] ; then
                     echo "export TESTSELECTION_PYTEST_ARGS='--path-filter=target/testselection/test-selection.txt '" >> $BASH_ENV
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Awesome! Thanks for onboarding SQS to the test selection! This will make future iterations way faster when working on the service!
In addition to the opt-in file matchers, it's important that the API dependencies (mandatory and optional) are  up-to-date to ensure that tests for dependent services are properly executed (more info on that in https://github.com/localstack/localstack/pull/10458).

Do you know if these dependencies are already up to date?
If not, I think they should be updated in t his PR by just updating the dependency mapping here:
https://github.com/localstack/localstack/blob/84f4a31a12299a391b5b14fd538c46751ecfddff/localstack/utils/bootstrap.py#L45-L80

baermat:
> Awesome! Thanks for onboarding SQS to the test selection! This will make future iterations way faster when working on the service! In addition to the opt-in file matchers, it's important that the API dependencies (mandatory and optional) are up-to-date to ensure that tests for dependent services are properly executed (more info on that in #10458).
> 
> Do you know if these dependencies are already up to date? If not, I think they should be updated in t his PR by just updating the dependency mapping here:
> 
> https://github.com/localstack/localstack/blob/84f4a31a12299a391b5b14fd538c46751ecfddff/localstack/utils/bootstrap.py#L45-L80

Yes I went over them, and as far as I can tell they are up to date. I probed the services for the common ways of interconnecting to sqs, and everything that came up was already covered.

thrau (approved):
Awesome! Glad we got it over the line :-) Looks like we cut the test runtime from ~33 minutes down to ~18 minutes. Cool!

PR #10786, opened 2024-05-07
fix CircleCI test error reporting by disabling colored reporting

alexrashed:
## Motivation
With #10761 we tweaked the logging config for our CI pipelines. With that, we also enabled the PyTest colored log output wherever possible.
This is quite useful and is properly displayed in the CircleCI logs, however, the CircleCI [test reporting doesn't properly strip away the ANSI color coding](https://app.circleci.com/pipelines/github/localstack/localstack/24659/workflows/b7bf4278-b715-4d88-996b-3e0149b30732/jobs/204905/tests):
image

This PR disables the pytest coloring for the CircleCI workflow again. 😢 

## Changes
- Removes `PY_COLORS=1` from the test command in the Makefile (which is used by CircleCI).

## Testing
- [x] Take a look at the triggered CircleCI workflow run to see if the coloring is disabled again.


dfangl (approved):
Sad, but necessary for now. Would be nice to disable it only for the xml reports though :)

PR #10790, opened 2024-05-08
Separate test selection env setup from tinybird setup

dominikschubert:
## Motivation

CircleCI Pipeline started failing on the steps that are not actively using the test selection but were using the tinybird pytest setup command since it assumed the existence of the generated test selection file without having an actual dependency on it defined via the workflow.

See: https://github.com/localstack/localstack/pull/10785
And especially the run here: https://app.circleci.com/pipelines/github/localstack/localstack/24674/workflows/f7383c5f-7a94-48c3-8734-eb4a6d4387b7/jobs/205055

## Changes

- Separates the env setup command for the testselection from the tinybird setup so it will only be executed on jobs that actually make use of it and depend on it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ executors:
       image: << pipeline.parameters.ubuntu-amd64-machine-image >>
 
 commands:
-  prepare-pytest-tinybird:
+  prepare-testselection:
     steps:
       - unless:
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Good catch! Thanks for jumping on this! 🦸🏽

PR #10793, opened 2024-05-08
Update PULL_REQUEST_TEMPLATE.md

thrau:
## Motivation

I noticed that the Changes section in PR descriptions often contain a list of things that the developer did to the code base, similar to a commit history. "refactored X", "add new function Y". This is not surprising given the section is called "Changes", and how the comment above the section was phrased ("What notable changes does this PR make"). What is more relevant to PR description is what _behavioral_ changes were made. How does localstack behave differently now? "Service X now raises an exception when", "Call to X now correctly lists ...", "CFn resource X is now supported", ...

So I made a minor change to the comment to evoke this type of response, rather than simply creating what is basically a list of commit messages.

The Testing and TODO sections were added into a single comment block, which made it a bit annoying to "enable" the sections. Now you can uncomment individual blocks.

I also noticed the link to the contribution guide is now outdated since we moved the contribution docs.

## Changes

* The template now correctly links to the contribution guidelines
* The phrasing for the Changes section now better describes what we should be
* The optional sections can now be commented out individually

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,25 +1,24 @@
-<!-- Please refer to the contribution guidelines before raising a PR: https://github.com/localstack/localstack/blob/master/CONTRIBUTING.md -->
+<!-- Please refer to the contribution guidelines before raising a PR: https://github.com/localstack/localstack/blob/master/docs/CONTRIBUTING.md -->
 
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
LGTM! 💯
Changing the phrasing can make quite a difference, I'm really curious how the PR descriptions will change after this subtle adjustment of the comments. 😃

PR #10795, opened 2024-05-08
CFn: add schema validation for resource providers

simonrw:
## Motivation

We use the CloudFormation resource provider schema to generate our scaffolded resources, but we don't make any assertions that what is returned by a resource provider conforms to the schema. Now that we have migrated all resources to the new resource provider API we should be stricter about the responses coming back from the resource providers. This will help us ensure parity with AWS CFn.

## Changes

* Implement a schema validation method in the `ResourceProviderExecutor`
    * at the moment, this only validates that all required fields are present, but will provide a place to put further validations in the future
    * the `PhysicalResourceId` is validated implicitly with the call to `extract_physical_resource_id_from_model_with_schema`, which raises an exception (in `resolve_json_pointer`) if the value is not present

## Testing

The green CI pipeline run shows that all community resource providers already conform to the schema (:tada:). To check that the validation works, pick a resource provider and comment out setting one of the `required` fields.

### DIFF
--- a/localstack/services/cloudformation/resource_provider.py
+++ b/localstack/services/cloudformation/resource_provider.py
@@ -425,6 +425,11 @@ class ResourceProviderExecutor:
                                 "A ResourceProvider should always have a SCHEMA property defined."
                             )
                         resource_type_schema = resource_provider.SCHEMA
+                        self.validate_model_properties(
+                            logical_resource_id=raw_payload["requestData"]["logicalResourceId"],
+                            schema=resource_type_schema,
+                            model=event.resource_model,
+                        )
                         physical_resource_id = (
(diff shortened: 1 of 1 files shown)


pinzon (approved):
Awesome!

simonrw (commented):
Notes from review pairing with @dominikschubert and @Morijarti :
* validation of `required` fields should be done _before_ resource provider execution as this represents input values supplied by the user
* we could validate no _extra_ properties are assigned by the resource provider on `SUCCESS`
* notes for the future:
    * we could typecheck the fields, possibly depending on how we want to handle implicit typecasting

simonrw:
Closing as we think this effort is not worth the trouble.

PR #10800, opened 2024-05-09
move source code into localstack-core

thrau:
## Motivation

This is the first transitional step into a multi-distribution + namespace package localstack setup. Before we introduce multiple distributions and a namespace package, we can move the localstack source code into `localstack-core`, and use the [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) to keep everything backwards compatible, while at the same time committing one of the biggest involved changes, which is `mkdir localstack-core && mv localstack localstack-core`.

Once we introduce multiple distribution, each distribution will have a flat layout again, but that requires restructuring of the entire project setup.

TODO: pycharm/vscode config update

## Changes

* the localstack source code (`localstack` package) now lives in `./localstack-core` rather than `./`. In the container, it now moved from `/opt/code/localstack/localstack` to `/opt/code/localstack/localstack-core/localstack`

## Merging the PR

I would like to manually merge this PR because i'd like the file move and config updates separate from each other. this will make it easier to understand in the history, potentially make rebasing easier, and also allow us to clearly ignore the renames (although they should be by default by git).

## TODO

What's left to do:

- [X] Fix plux https://github.com/localstack/plux/pull/20
- [x] Fix build pipeline
  - [x] Fix test selection
  - [x] Fix s3 image pipeline
  - [x] Fix pro test pipeline
  - [x] Fix not-implemented collection
  - [x] Fix cobertura report
- [x] Release and update to plux 1.10 https://github.com/localstack/localstack/pull/10830
- [x] Update `localstack.dev.run`
- [x] Create documentation how the editor (pycharm/vscode) setup now changes
- [x] Write an issue description that explains the change (and while it may break some people's localstack scripts, the source code location is not part of the public API)
- [x] Communicate reasons and implications to the team
- [x] Prepare PR rebasing instructions
- [x] Clean up the PR commit history for manual merge
- [ ] Manually merge the PR


.circleci/config.yml
```diff
@@ -134,7 +134,7 @@ jobs:
                     circleci-agent step halt
                   else
                     source .venv/bin/activate
-                    python -m localstack.testing.testselection.scripts.generate_test_selection_from_pr /tmp/workspace/repo $CI_PULL_REQUEST target/testselection/test-selection.txt
+                    PYTHONPATH=localstack-core python -m localstack.testing.testselection.scripts.generate_test_selection_from_pr /tmp/workspace/repo $CI_PULL_REQUEST target/testselection/test-selection.txt
```
dominikschubert, line 137: This should be removed after https://github.com/localstack/localstack/pull/10830 has been merged, right?

dominikschubert:
@thrau 

collecting a few of my findings below, will update with more when I encounter them

- ASF generation is broken `FileNotFoundError: [Errno 2] No such file or directory: 'localstack/aws/api/stepfunctions'
- CloudFormation resource provider generation script is broken and needs to be adjusted since it assumes a few things around the localstack/localstack-ext directory setup

.circleci/config.yml
thrau: no this has nothing to do with the plugin loading, but just the src layout.

without installing localstack with `pip install -e .` into the activated venv, the python binary cannot find the source code when a src layout is used.
another option is to cd into `localstack-core`, or move test selection scripts out of ls all together

Issue #10073, opened 2024-01-15
StepFunctions: add optimised service integration for AWS Glue.


MEPalma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

The current provider for AWS Step Functions does not support AWS Glue optimised service integrations https://docs.aws.amazon.com/step-functions/latest/dg/connect-glue.html

PR #10802, opened 2024-05-10
[SFN] Support for Glue Optimised Service Integration

MEPalma:
## Motivation
Currently the SFN v2 provider does not support [Glue optimised integrations](https://docs.aws.amazon.com/step-functions/latest/dg/connect-glue.html). This PR adds base support for Glue optimised integration.
Closes: #10073 (I'm author of this issue)

## Changes
- Added new handler for the optimised service integration

## Testing
Glue is a Pro feature, tests are added in the PR in tandem with this one.

### DIFF
--- a/localstack/testing/pytest/stepfunctions/fixtures.py
+++ b/localstack/testing/pytest/stepfunctions/fixtures.py
@@ -12,9 +12,6 @@ from localstack_snapshot.snapshots.transformer import (
 from localstack.testing.aws.util import is_aws_cloud
 from localstack.testing.pytest.stepfunctions.utils import await_execution_success
 from localstack.utils.strings import short_uid
-from tests.aws.services.stepfunctions.templates.callbacks.callback_templates import (
-    CallbackTemplates,
-)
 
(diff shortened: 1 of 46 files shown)


joe4dev (approved):
LGTM. Nevertheless, I would prefer having a green test run before merging. Can we use companion branches to run the full test suite, in particular the new tests for this code in localstack-ext?
Re-basing both PRs (it seems the localstack community PR might be outdated causing issues in the ext pipeline) and pushing them simultaneously should work.

Just commented some (tiny) nits.

localstack/testing/pytest/stepfunctions/fixtures.py
```diff
@@ -427,3 +571,74 @@ def _setup(event_pattern):
         return queue_url
 
     return _setup
+
+
+@pytest.fixture
+def sfn_events_to_sqs_queue(events_to_sqs_queue):
```
joe4dev, line 577: cc @maxhoheiser We now have SFN fixtures in localstack, which can be useful for integrations in EventBridge and EventBridge Pipes 🥳

PR #10808, opened 2024-05-13
Introduce LOGGING_OVERRIDE config var

simonrw:
## Motivation

TBD

## Changes

TBD


localstack/config.py
```diff
@@ -418,6 +418,9 @@ def in_docker():
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
+# allow setting custom log levels for individual loggers
+LOGGING_OVERRIDE = os.environ.get("LOGGING_OVERRIDE", "{}")
```
dfangl, line 422: Not sure about the name, maybe `LOG_LEVEL_CONFIG` or `LOGGING_CONFIG`? Since this is for now an undocumented change, it might not be that important, but those changes sometimes stay :)

localstack/logging/setup.py
```diff
@@ -78,6 +79,23 @@ def setup_logging_from_config():
         for name, level in trace_internal_log_levels.items():
             logging.getLogger(name).setLevel(level)
 
+    if raw_value := config.LOGGING_OVERRIDE:
+        try:
+            logging_overrides = json.loads(raw_value)
```
dfangl, line 84: WDYT json vs key-value pairs? I think the latter might be easier to write, and we do not need nested structures. I do not have a strong opinion though.

localstack/logging/setup.py
```diff
@@ -78,6 +79,23 @@ def setup_logging_from_config():
         for name, level in trace_internal_log_levels.items():
             logging.getLogger(name).setLevel(level)
 
+    if raw_value := config.LOGGING_OVERRIDE:
```
dfangl, line 82: Not sure if we can use this syntax here - this file is imported by the cli if the debug flag is set. Our tests pass, but that flag might break.

dfangl (commented):
Thanks for implementing this, configuring specific components to log on a higher level will hopefully be useful in support cases.

localstack/config.py
simonrw: I think this name is misleading, since it sounds like it updates the _entire_ config, rather than just allowing overrides. WDYT?

localstack/logging/setup.py
simonrw: Good point, thanks!

localstack/logging/setup.py
simonrw: I think that's a good usability improvement :+1:

simonrw: Implemented in db163efc557d2a75ca5798875142e0534a1aa4c1

localstack/config.py
dfangl: You are right, let's keep it to override for now :)

alexrashed (commented):
Thanks for challenging the logging config and testing new approaches.
I only added 2 questions to the code, but I am not 100% sure about the reasoning behind it.
We already have the distinction between `default_log_levels`, `trace_log_levels`, and `trace_internal_log_levels`. Are the logs really so noisy that they are even too verbose for `trace_internal_log_levels`?

localstack/config.py
```diff
@@ -418,6 +418,9 @@ def in_docker():
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
+# allow setting custom log levels for individual loggers
```
dominikschubert, line 421:
```suggestion
# EXPERIMENTAL
# allow setting custom log levels for individual loggers
```

Just to make this very explicit

dominikschubert: I'd also recommend adding a sample value here

localstack/utils/collections.py
```diff
@@ -533,3 +533,18 @@ def is_comma_delimited_list(string: str, item_regex: Optional[str] = None) -> bo
     if pattern.match(string) is None:
         return False
     return True
+
+
+def parse_key_value_pairs(raw_text: str) -> dict:
```
dominikschubert, line 538: nit: could add a docstring here with for example the values from one of the test cases below to clearly demonstrate its purpose.

localstack/logging/setup.py
```diff
@@ -78,6 +79,24 @@ def setup_logging_from_config():
...
+                level = getattr(logging, level_name, None)
+                if not level:
+                    raise RuntimeError(
+                        f"Failed to configure logging overrides ({raw_logging_override}): '{level_name}' is not a valid log level"
+                    )
+                logging.getLogger(logger).setLevel(level)
+        except RuntimeError:
+            raise
+        except Exception as e:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
simonrw: I am not sure, so since we "raise ... from e" I am happy to remove the "malformed input" part of the error message as it only printed `e` anyway.

localstack/utils/collections.py
simonrw: Thanks for the suggestion!

localstack/utils/collections.py
```diff
@@ -533,3 +533,18 @@ def is_comma_delimited_list(string: str, item_regex: Optional[str] = None) -> bo
...
+    result = {}
+    for pair in raw_text.split(","):
+        items = pair.split("=", maxsplit=2)
+        if len(items) != 2:
+            raise ValueError(f"invalid key/value pair: '{pair}'")
```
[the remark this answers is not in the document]
simonrw: You are right, thanks!

localstack/logging/setup.py
```diff
@@ -78,6 +79,24 @@ def setup_logging_from_config():
...
+    raw_logging_override = config.LOGGING_OVERRIDE
+    if raw_logging_override:
+        try:
+            logging_overrides = parse_key_value_pairs(raw_logging_override)
+            for logger, level_name in logging_overrides.items():
+                level = getattr(logging, level_name, None)
+                if not level:
+                    raise RuntimeError(
+                        f"Failed to configure logging overrides ({raw_logging_override}): '{level_name}' is not a valid log level"
+                    )
```
thrau, line 91: is there a specific reason to raise `RuntimeError` and not `ValueError`?

thrau (approved):
really nice addition! thanks for going through the review iterations, i think the reviews improved the already great PR a lot!

My only nit would be that I'm also not 100% sold on the name. If we're overriding log levels, then i believe those three words should appear in the config variable, `LOG_LEVEL_OVERRIDES` for instance?
* LOG_LEVEL to indicate what it is you're configuring
* OVERRIDES (plural) to indicate that you're overriding multiple log level configs

alexrashed (approved):
Nice cleanup, thanks for tackling all the comments, the result is super small and clean! 💯

simonrw:
Marking as `minor` since it changes the "public" API (i.e. configuration) of LocalStack.

Issue #10796, opened 2024-05-08
bug: ListSubscriptionsByTopic no longer paginates results


matt-phylum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Regardless of how many subscribers you have, ListSubscriptionsByTopic returns all of them in a single page.

### Expected Behavior

ListSubscriptionsByTopic should paginate results, similar to how AWS does it and previous versions of LocalStack did it, so that tests of what happens when there are multiple pages work correctly.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Localstack is started as part of a custom test environment. 101 e-mails are registered via Subscribe, and then the subscriptions are retrieved using ListSubscriptionsByTopic.

### Environment

```markdown
- OS: localstack/localstack container
- LocalStack: localstack/localstack@latest (sha256:f9310db87d9d76c449918a849562303b1c53f77c3598fdcb9c6cd7843d0f8cb6?)
```

### Anything else?

This was working yesterday (2024-05-07). It looks like #10788 replaced the existing implementation that supported pagination with a new one that doesn't. ListSubscriptions is probably similarly affected.

bentsku:
Hello @matt-phylum and thanks for the report!

Yes, this bug fix for cross-account subscriptions introduced the small change in behavior, I'll re-introduce the pagination shortly. Thanks again for the report.

PR #10810, opened 2024-05-13
fix pagination for SNS ListSubscriptions and ListSubscriptionsByTopic

bentsku:
## Motivation
As reported with #10796, I've introduced a small regression with #10788 by fixing cross-account and pulling the listing operations from moto. 

## Changes
- re-introduced the pagination and an AWS validated test for the pagination. 

_fixes #10796_

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -841,6 +841,7 @@ class TestSNSSubscriptionCrud:
             sns_subscription(TopicArn=topic_arn, Protocol="sms", Endpoint=phone_number)
 
         response = aws_client.sns.list_subscriptions_by_topic(TopicArn=topic_arn)
+        page_1_subs = {sub["SubscriptionArn"] for sub in response["Subscriptions"]}
         # not snapshotting the results, it contains 100 entries
         assert "NextToken" in response
         # seems to be b64 encoded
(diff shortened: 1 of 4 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -822,6 +822,45 @@ def test_list_subscriptions(
...
+        )
+
+        base_phone_number = "+12312312"
+        topic_arn = sns_create_topic()["TopicArn"]
+        for phone_suffix in range(101):
+            phone_number = f"{base_phone_number}{phone_suffix}"
+            sns_subscription(TopicArn=topic_arn, Protocol="sms", Endpoint=phone_number)
+
+        response = aws_client.sns.list_subscriptions_by_topic(TopicArn=topic_arn)
+        page_1_subs = {sub["SubscriptionArn"] for sub in response["Subscriptions"]}
```
cloutierMat, line 844: :rocket: Nice use of sets!

tests/aws/services/sns/test_sns.py
```diff
@@ -822,6 +822,45 @@ def test_list_subscriptions(
...
+        # not snapshotting the results, it contains 100 entries
+        assert "NextToken" in response
+        # seems to be b64 encoded
+        assert response["NextToken"].endswith("==")
+        assert len(response["Subscriptions"]) == 100
```
cloutierMat, line 849: Wouldn't you want the `len(page_1_subs)`? As I assumed you made it a set to catch dedups? And you are asserting only one should come on page two.

cloutierMat (approved):
Looks good to me!

I only left one questions regarding length assertion.

tests/aws/services/sns/test_sns.py
bentsku: I think it's similar, I was using a set above not for deduplication but just because the `in` call would be a tiny bit faster 😅 but yes, calling `len` would be the same on both. 
The set was made is just to verify that the one coming on page 2 is not already present in page 1. But I'll update it to be a bit clearer! Thanks!

cloutierMat: haaa! Now I understand. Just my assumption that was wrong :stuck_out_tongue_closed_eyes:

PR #10814, opened 2024-05-14
Fix entrypoint script configurator encoding and line endings in windows

dfangl:
## Motivation
When using the CLI in extension dev mode under windows, the custom container entrypoint file gets generated using windows file endings (`\r\n`).
This file is then mounted into the linux container, and fails to execute due to the wrong line endings.

To avoid this, we now force linux file endings for container entrypoints, and in addition utf-8 encoding, since some entrypoints (for example the extension dev mode one) contains unicode characters.

## Changes
- LocalStack now starts in extension dev mode under windows without additional intervention


thrau (approved):
nice catch! lgtm :100:

PR #10815, opened 2024-05-14
fix GitHub error reporting by disabling colors

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/10761 we tweaked the logging config for our CI pipelines. With that, we also enabled the PyTest colored log output wherever possible.
This is quite useful and is properly displayed in the error reporting.
This was fixed for CircleCI with https://github.com/localstack/localstack/pull/10786, but apparently this is also an issue with the GitHub error reporting.
This PR disables the colors everywhere to fix the reports.

## Changes
- Removes `PY_COLORS` everywhere to fix the error reporting.


silv-io (approved):
Sad that we have to remove it again. Hopefully in the future GitHub will support the color tags in the summary

Issue #10252, opened 2024-02-15
bug: Unable to invoke StepFunction from API Gateway. 


sn0wm1ku:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

1) Insert these code to default cdk project

```typescript

import * as apigw from "aws-cdk-lib/aws-apigateway";
import * as stepfunctions from "aws-cdk-lib/aws-stepfunctions";
import { Construct } from "constructs";
export class ApiGWStepFns extends Construct {
  constructor(scope: Construct, id: string) {
    super(scope, id);
    const sfn = this.defineSfn(`${id}Sfn`);
    this.defineApiGw(sfn);
  }
  defineSfn = (stateMachineName: string) => {
    const step = new stepfunctions.Pass(this, `${stateMachineName}Pass`, {
      result: { value: "Hello.World" },
    });
    const workflow = stepfunctions.Chain.start(step);
    const sfn = new stepfunctions.StateMachine(this, stateMachineName, {
      stateMachineName,
      stateMachineType: stepfunctions.StateMachineType.EXPRESS,
      definitionBody: stepfunctions.DefinitionBody.fromChainable(workflow),
    });
    return sfn;
  };
  defineApiGw = (sfn: stepfunctions.StateMachine) => {
    return new apigw.StepFunctionsRestApi(this, "api", {
      stateMachine: sfn,
    });
  };
}

```

2) cdklocal deploy
3) Run
```shell
APIID=$(aws --endpoint-url http://127.0.0.1:4566 apigateway get-rest-apis --query "items[*]".id --output text)
curl -X POST \
http://$APIID.execute-api.localhost.localstack.cloud:4566/prod/
```

4) Error
{"Type": "User", "message": "Error invoking integration for API Gateway ID '1h0vyqn2i0': An error occurred (StateMachineDoesNotExist) when calling the StartExecution operation: State Machine Does Not Exist: 'arn:aws:states:ap-northeast-1:000000000000:stateMachine:dbWriteSfn'", "__type": "InvalidRequest"}%                                                                        

### Expected Behavior

Show result of step function?

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose up -d

docker-compose.yml
```yml

version: "3.8"

services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack
[...]

PR #10816, opened 2024-05-14
APIGW: Default to an empty dict when the provided body for an API Gateway step function is an empty string

marcciosilva:
## Motivation

This attempts to solve issue #10252 .

I created a CDK project following what the reporter of the issue added to its description, and when running `curl -X POST https://$APIID.execute-api.us-east-1.amazonaws.com/prod/`, I got the expected output:
```
"Hello.World"
```

When doing the same through cdklocal against a local instance of Localstack, I got the following output:
```
{"Type": "User", "message": "Error invoking integration for API Gateway ID 'fsbb50s4si': An error occurred (InvalidExecutionInput) when calling the StartExecution operation: Expecting value: line 1 column 10 (char 9)", "__type": "InvalidRequest"}%
```

This isn't exactly the same as what the user reported (which was probably partially fixed since February), but something was definitely wrong - so upon inspection, I found that when a request body was not provided, we ended up generating malformed JSON, which was in turn sent to the step function. With this fix, the output is now the expected one (added a small unit test for the templating logic, trying to mimic an actual velocity template with body forwarding).

## Changes

We now default to an empty dict when an empty body comes when executing a step function - this allows us to create a correct JSON.

### DIFF
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -190,7 +190,7 @@ class ApiGatewayVtlTemplate(VtlTemplate):
             namespace["stageVariables"] = stage_var
         input_var = variables.get("input") or {}
         variables = {
-            "input": VelocityInput(input_var.get("body"), input_var.get("params")),
+            "input": VelocityInput(input_var.get("body") or {}, input_var.get("params")),
             "util": VelocityUtilApiGateway(),
         }
         namespace.update(variables)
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! I've looked around in the codebase and I could see instances where we use the `ApiGatewayVtlTemplate` class, and manually do something like:
```python
message = message or {}
variables = {"request": {"body": message}}
route = ApiGatewayVtlTemplate().render_vtl(
    route_selection_expression, variables=variables
)
```
before rendering it. So this seems like a good fix. We usually try to add an AWS validated test for the particular fix, but in that case, that would probably means a lot of setup and pretty hard to validate directly. 

Thanks a lot for addressing this! 🚀

PR #10818, opened 2024-05-14
add small quirks for new resource providers

pinzon:
## Motivation
This PR adds changes to allow the correct deployment of WAFv2 resource providers. These resources physical ids are composed like: Name|Id|Scope. 

Also, trying to use the scaffolding tool to generate the files for WAFv2::WebACL throws a RecursionError. 

## Changes
- CFn quirks added for WAFv2 resources
- Scaffolding changes for correct WebACL scaffolding


simonrw (approved):
Thanks for updating the quirks here.

Regarding the scaffolding: I'm happy to include this fix for now as the resulting type is valid, though not very helpful. I think we should fix the `self.resolve_type_of_property` though to be complete.

simonrw:
update: it looks like the spec is incomplete. There is no `#/definitions/Rule` definition in the file!

pinzon:
Alright, I'll go ahead with the merge. We can fix or add a better handling for that later.

PR #10819, opened 2024-05-14
fix SNS cross-account issues with Subscriptions

bentsku:
## Motivation
Follow up from #10788

We now had issues around accessing subscriptions made by another account for `SetSubscriptionAttributes`, which has extensive validations of the `FilterPolicy` field. 
The issue was the same as before, Moto was not saving the subscription in the right backend for the topic account.

We were currently blocked for #10691 because the validation of filter policies containing `$or` was incorrect, so it was the right opportunity to move the validation inside LocalStack, and remove usage of moto around SNS Subscriptions. We were already saving and using the `Subscription` data from our store only, and used moto for additional validation. 

As of today, `Topic` is still stored in moto and extended in LocalStack, and mocked operations around A2P (SMS, and Platform Applications/Endpoint, related to mobile notifications) are still done by moto. 

## Changes
- migrated all the tests related to `FilterPolicy` in its own file as it was becoming consequent, and we added a few more. 
  - the `TestSNSFilterPolicyConditions` class in `test_sns_filter_policy.py` contains the new tests
- move the code related to Filter Policy to its own file, to group the actual filter policy matcher and the validator
- remove usage of `call_moto` and `call_moto_with_request` in the SNS provider
- sneaked a small parity fix for `HTTP` subscriptions (returning `pending confirmation` instead of the ARN if the subscription is still pending confirmation, except if `ReturnSubscriptionArn` is set to `true`).
  - added `test_http_subscription_response` in `test_sns.py`


localstack/services/sns/filter.py
```diff
@@ -0,0 +1,429 @@
...
+            return field_exists and value == condition
+        elif (must_exist := condition.get("exists")) is not None:
+            # if must_exists is True then field_exists must be True
+            # if must_exists is False then fields_exists must be False
+            return must_exist == field_exists
+        elif value is None:
+            # the remaining conditions require the value to not be None
+            return False
+        elif anything_but := condition.get("anything-but"):
+            return value not in anything_but
```
cloutierMat, line 97: "anything-but" can also be a "prefix" https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-anything-but-matching-prefix

localstack/services/sns/filter.py
```diff
@@ -0,0 +1,429 @@
...
+
+        return _inner(filter_policy)
+
+    def _validate_rule(self, rule: t.Any) -> None:
+        match rule:
+            case None | str() | bool():
+                return
+
+            case int() | float():
+                # TODO: AWS says they support onyl from -10^9 to 10^9 but seems to accept it, so we just return
```
cloutierMat, line 325: typo: onyl => only

localstack/services/sns/filter.py
```diff
@@ -0,0 +1,429 @@
...
+            raise InvalidParameterException(
+                f"{self.error_prefix}FilterPolicy: Value of {exc_operator} must be numeric"
+            )
+
+        if not num_values:
+            return
+
+        if operator not in (">", ">="):
+            raise InvalidParameterException(
+                f"{self.error_prefix}FilterPolicy: Too many elements in numeric expression"
```
cloutierMat, line 401: Really? That is the error that aws will return if you place the `<` before `>`? That sounds silly! :rofl:

cloutierMat (changes requested):
Lots of work in there, congrats for keeping all that easily readable and adding a lot of meaningful tests.

I added most of my comments before realising that most of the `SubscriptionFilter` class was a refactor... Some of them might not be as relevant to your current task then.

The only missing part I see is the lack of support for `"event":[{"anything-but": {"prefix": "order-"}}]`. Let me know if you feel this is better suited for a follow up work.

localstack/services/sns/filter.py
```diff
@@ -0,0 +1,429 @@
...
+
+from localstack.aws.api.sns import InvalidParameterException
+
+
+class SubscriptionFilter:
+    def check_filter_policy_on_message_attributes(
+        self, filter_policy: dict, message_attributes: dict
+    ):
+        for criteria, conditions in filter_policy.items():
+            if not self._evaluate_filter_policy_conditions_on_attribute(
```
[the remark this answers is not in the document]
bentsku: Hmm, the biggest difference is that `MessageAttributes` can only be on one level, they can't have nested dict. So the logic is much simpler there. I think we could maybe join the 2 implementations, but not sure if worth it. You technically can have objects inside the message for `MessageAttributes`, but the behavior is undefined from the documentation. We could have a look in the follow up, as you said, this was just moved around and will be updated with the next PR #10691

localstack/services/sns/filter.py
bentsku: Yes! This is not supported yet, and we also do not support `suffix` and `equals-ignore-case`, which will be supported by #10691. I could try to sneak the `anything-but` into `prefix` in there as well.

cloutierMat (approved):

PR #10822, opened 2024-05-15
WIP: Introduce experimental JSON logging

dominikschubert:
## Motivation

We've had a few requests in the past for outputting our logs in a structured format. We've been experimenting in the past but would like to push forward with the idea now. This will be in an experimental stage for some time now and will mostly be used internally before going into public preview and ideally GA release with 4.0 :crossed_fingers: 

## Changes

- Logs can now also be streamed to a file in JSON format. This is done quite in a quite naive way right now, we simply take the current log records, and translate it into a JSON dict which then ends up being streamed as newline-delimited JSON where each line corresponds to one log entry.
- The current CLI output is generally  not affected by this (except by the addition of the raw request logger)

What's left to do:

- [ ] Clean up the code from the initial spike
- [ ] Make this opt-in
- [ ] Make log sink configurable
- [ ] Map keys extracted from the LogRecord to more suitable names
- [ ] Investigate the DEBUG-related code we had to comment out for the aws request/response logging

There will also be some follow ups such as:
- Add an internal endpoint to stream logs from (e.g. via websocket)
- Introduce a proper structured logging strategy and rework log levels
- ... lots more to come in the next few months

### DIFF
--- a/localstack/aws/app.py
+++ b/localstack/aws/app.py
@@ -1,4 +1,3 @@
-
 from localstack import config
 from localstack.aws import handlers
 from localstack.aws.api import RequestContext
@@ -29,6 +28,7 @@ class LocalstackAwsGateway(Gateway):
         # the main request handler chain
         self.request_handlers.extend(
             [
+                handlers.log_request,
                 handlers.push_request_context,
                 handlers.add_internal_request_params,
                 handlers.handle_runtime_shutdown,
--- a/localstack/aws/app.py
+++ b/localstack/aws/app.py
@@ -1,3 +1,4 @@
+
 from localstack import config
 from localstack.aws import handlers
 from localstack.aws.api import RequestContext
@@ -45,6 +46,7 @@ class LocalstackAwsGateway(Gateway):
                 handlers.add_region_from_header,
                 handlers.add_account_id,
                 handlers.parse_service_request,
+                # TODO: add logger that initializes a request "trace"
                 metric_collector.record_parsed_request,
                 handlers.serve_custom_service_request_handlers,
                 load_service,  # once we have the service request we can make sure we load the service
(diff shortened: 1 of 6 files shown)


localstack/services/lambda_/provider.py
```diff
@@ -765,6 +765,7 @@ def create_function(
                 f"Value {request.get('Runtime')} at 'runtime' failed to satisfy constraint: Member must satisfy enum value set: {VALID_RUNTIMES} or be a valid ARN",
                 Type="User",
             )
+
```
dominikschubert, line 768:
```suggestion
```

localstack/aws/app.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, service_manager: ServiceManager = None) -> None:
                 handlers.add_region_from_header,
                 handlers.add_account_id,
                 handlers.parse_service_request,
+                # TODO: add logger that initializes a request "trace"
```
dominikschubert, line 49:
```suggestion
```

localstack/aws/handlers/logging.py
```diff
@@ -95,6 +95,10 @@ def _log(self, context: RequestContext, response: Response):
                     response.status_code,
                     context.service_exception.code,
                     extra={
+                        "request_id": context.request_id,
```
joe4dev, line 98: `request_id` propagation is gonna be crucial for proper grouping and is going to require quite some manual work within service-specific loggers. Other fields can then be inferred based on the `request_id` because some other fields won't be available at different stages (e.g., service, region, and operation won't be available before parsing an HTTP request into an AWS request; response-related fields such as status code obviously won't be available before handling the request)

dominikschubert:
Closing, as this is unlikely to be picked up in the near future.

Issue #10801, opened 2024-05-09
bug: S3 integration returns nonconformant `ListVersionsResult` object as per provided AmazonS3.xsd


Dragas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Returned `ListVersionsResult` element for bucket versions request does not conform to the expected structure provided by amazon s3.

### Expected Behavior

`ListVersionsResult` element matches that provided by actual s3.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Following is fragment from docker-compose

```yml
      localstack:
         image: localstack/localstack:3.4
         ports:
          - "4566:4566"
          - "4510-4559:4510-4559"
         environment:
          - SERVICES=s3
          - DOCKER_HOST=unix:///var/run/docker.sock
          - AWS_DEFAULT_REGION=us-east-1
         volumes:
          - /var/run/docker.sock:/var/run/docker.sock
```
#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

Setup the created bucket somewhat close to the awsdocs bucket provided by amazon.

```sh
        touch arbitraryfile.txt
        awslocal s3 mb s3://awsdocs
        awslocal s3api put-object --bucket awsdocs --key "AMP/2.0/configuring-connector-2-0.pdf" --body arbitraryfile.txt
        awslocal s3api put-object --bucket awsdocs --key "AMP/2.0/deploying-connector-2-0.pdf" --body arbitraryfile.txt
        awslocal s3api put-object --bucket awsdocs --key "AMP/2.0/managing-connector-2-0.pdf" --body arbitraryfile.txt
        awslocal s3api put-object --bucket awsdocs --key "AWSCloudFormation/2010-05-15/cfn-api-2010-05-15.pdf" --body arbitraryfile.txt
        awslocal s3api put-object --bucket awsdocs --key "AWSCloudFormation/2010-05-15/cfn-apiref-2010-05-15.pdf" --body arbitraryfile.txt
        awslocal s3api put-object --bucket awsdocs --key "AWSCloudFormation/2010-05-15/cfn-gsg-2010-05-15.pdf" --body arbitraryfile.txt\
        awslocal s3api put-object --bucket awsdocs --key "AWSCloudFormation/2010-05-15/cfn-ug-2010-05-15.pdf" --body arbitraryfile.txt
[...]

PR #10829, opened 2024-05-15
fix ListObjectVersions field order

bentsku:
## Motivation
Similar to #8329, we had a report with #10801 where the AWS Java SDK v1 could not properly parse the Bucket name out of the field. The reporter added very useful information about `AmazonS3.xsd`, we quickly communicated with @alexrashed about it, especially the fact that the S3 fields are reported as `<xsd:sequence>`, meaning the order is required/important. We will need to think about how we could add such specs, but also keep them in sync. 

## Changes
- similar to #8329, removing the `Versions` field from the spec and re-adding it, counting of Python dictionary ordering to make it last
- added a test to verify the fix

_fixes #10801_

### DIFF
--- a/localstack/aws/api/s3/__init__.py
+++ b/localstack/aws/api/s3/__init__.py
@@ -2588,7 +2588,6 @@ class ListObjectVersionsOutput(TypedDict, total=False):
     VersionIdMarker: Optional[VersionIdMarker]
     NextKeyMarker: Optional[NextKeyMarker]
     NextVersionIdMarker: Optional[NextVersionIdMarker]
-    Versions: Optional[ObjectVersionList]
     DeleteMarkers: Optional[DeleteMarkers]
     Name: Optional[BucketName]
     Prefix: Optional[Prefix]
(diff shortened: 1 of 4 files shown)


thrau (approved):
XML strikes again. glad you found a quick fix!

unfortunately it's very implicit, since it makes assumptions about how the serializer works, and the patch (first removing and then adding the same object) may look very odd to someone looking at the patches. it shows we have cases where more control of the serializer would be helpful (like making the order of certain things explicit, or mapping of tag names). then again, it seems that it only ever comes up with s3 :smile: anyway, nothing actionable, just an observation.

great find!

bentsku:
@thrau yes, definitely agree with everything you said here, it's all very, very implicit. I wonder if we could augment the S3 serializer with the XML specs `.xsd` for example. We'd need to update them, and verify that the XSD file at https://doc.s3.amazonaws.com/2006-03-01/AmazonS3.xsd is updated at the same time as Boto specs. 

This is a case where the dynamic nature of Python, which botocore takes advantage of, is playing against us, as the botocore specs can be less "tight" than for other languages. But definitely on the radar. Not actionable right now, but it's there! Thanks for the review

bentsku:
Thanks! I think the "biggest" issue is that we updates our generated code every week, from the specs contained in the Python client `boto3` and `botocore`.

We would need to be sure this `.xsd` file is updated at the same rhythm, and adapt our code generation for it. I'm not too worried about how to generate the code from the specifications, but a bit more about how to get those. It is good to know about `pyxb`! Thanks a lot! 🙏

PR #10830, opened 2024-05-15
update plux to 1.10.0

thrau:
## Motivation

Plux 1.10.0 now properly supports entry point resolution for editable wheels, the lack of which has caused headaches in the past, and also blocks #10800.

## Changes

* localstack now uses plux 1.10.0 and no longer needs stevedore

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 # LocalStack project configuration
 [build-system]
-requires = ['setuptools', 'wheel', 'plux>=1.7']
+requires = ['setuptools', 'wheel', 'plux>=1.10']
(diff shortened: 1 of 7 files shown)


thrau:
the environment keeps flaking on me :sob: now it's taking over an hour to run the amd64 tests

alexrashed (approved):
Awesome! Great to see transitive dependencies go, and I can't wait for the new project setup this upgrade enables! 🥳 🚀
The pipeline issues are weird, the runs take more than 15 minutes longer than usual. I restarted a run, this could have been caused by some kind of slow infra of stuff we download during the tests...

Issue #10280, opened 2024-02-20
bug: Lambda logs "Queue not available" when Lambda function is re-created


exoego:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In Docker logs, Localstack continuously says, `Error while polling lambda events for function arn:aws:lambda:us-east-1:000000000000:function:my-lambda:$LATEST: Queue not available`, when the Lambda function is re-created.

### Expected Behavior

Lambda's internal queue should be available and no error log should be shown.

### How are you starting LocalStack?

With a docker-compose file

### 1. Clone the repro repo

```bash
git clone https://github.com/exoego/localstack-lambda-queue-repro

cd localstack-lambda-queue-repro
```

### 2. Run docker compose in a terminal with the following command:

```bash
docker-compose up
```

### 3. Open another terminal and run the following command:

```bash
npm ci

npm run test
npm run test
```

### 4. Go to the tab where the docker is running and you will see the following error in Docker logs:
    
```bash
localstack-1  | 2024-02-20T04:22:34.170 ERROR --- [functhread94] l.s.l.i.event_manager      : Error while polling lambda events for function arn:aws:lambda:us-east-1:000000000000:function:my-lambda:$LATEST: Queue not available
localstack-1  | 2024-02-20T04:22:35.171 ERROR --- [functhread94] l.s.l.i.event_manager      : Error while polling lambda events for function arn:aws:lambda:us-east-1:000000000000:function:my-lambda:$LATEST: Queue not available
localstack-1  | 2024-02-20T04:22:36.173 ERROR --- [functhread94] l.s.l.i.event_manager      : Error while polling lambda events for function arn:aws:lambda:us-east-1:000000000000:function:my-lambda:$LATEST: Queue not available
```

### Environment

```markdown
- OS: macOS 14.2.1
- LocalStack: 3.1.0
```

### Anything else?

Repro: https://github.com/exoego/localstack-lambda-queue-repro

MarcelStranak:
Hello @exoego, the Repro: https://github.com/exoego/localstack-lambda-queue-repro, is inaccessible. Please set the proper permissions to access the repository.

exoego:
@MarcelStranak Ah, sorry. Made it public

joe4dev:
Thank you very much @exoego for sharing the clear reproducer 👏👏

[...]

PR #10831, opened 2024-05-15
Fix Lambda async invocation queue namespace

joe4dev:
## Motivation

Addresses https://github.com/localstack/localstack/issues/10280

Re-creating a Lambda function immediately after deleting it, triggers a race condition and deletes the internal async invoke queue. This causes continuous error logs and breaks async invokes.
The namespace clash also causes further issues because queued invokes can get deleted.

## Changes

* Introduce a volatile (i.e., non-persisted) function instance id unique to a running LocalStack installation that can be used for the async queue namespace
* Add test `test_recreate_function` to reproduce the reported function recreate issue
* Add test `test_function_update_during_invoke` to validate the AWS behavior of function updates during a running invocation => Currently broken on LocalStack ⚠️ 
* Add test `test_async_invoke_queue_upon_function_update` to validate the queuing behavior of async invokes when a function update happens => kinda works but the first invoke fails and gets retried due function updates aborting all running invokes (even for $LATEST)

Unrelated changes sneaked in:
* Fix config default for event rule engine (align with docs)
* Remove unnecessary `lambda_service` backlink in Lambda service manager (missing declaration cleanup from [this PR](https://github.com/localstack/localstack/pull/10614/files#diff-056f8f1dc9188bf87729941f6be9805c0caf16189dafa882f16fb259676f79daL66))
* Fix an outdated comment in the Docker runtime executor
* Fix misleading name in test and snapshot

## Testing

State pickling testing whether the `instance_id` attribute is persisted or not:

1. Debug `tests.aws.services.lambda_.test_lambda.TestLambdaCleanup.test_recreate_function` with a breakpoint at the end of the tests
2. Export the state using `localstack-pro state export --format json lambda-state`
3. Load the pickled file in a scratch file:

```python
import pickle
import pprint

with open("/Users/joe/Downloads/lambda-state/api_states/000000000000/lambda/us-east-1/store.state", "rb") as f:
    lambda_store = pickle.load(f)

    pprint.pp(lambda_store)
    print("attr_functions".center(80, "="))
    pprint.pp(lambda_store.attr_functions)
    print("functions".center(80, "="))
    pprint.pp(lambda_store.functions)

```

## Discussion

* I'm not really happy with the volatile variable hack in the lambda model. Maybe, we can find some better solution here 🤔 
* We should not abort running executions for updates of $LATEST. That seems like a common scenario.

### DIFF
--- a/localstack/services/lambda_/invocation/docker_runtime_executor.py
+++ b/localstack/services/lambda_/invocation/docker_runtime_executor.py
@@ -363,7 +363,8 @@ class DockerRuntimeExecutor(RuntimeExecutor):
             container_config.dns = config.LAMBDA_DOCKER_DNS
         else:
             if dns_server.is_server_running():
-                # Enable transparent endpoint injection by setting DNS to the embedded DNS re-writer in the Go init.
+                # Set the container DNS to LocalStack to resolve localhost.localstack.cloud and
+                # enable transparent endpoint injection (Pro image only).
                 container_config.dns = self.get_endpoint_from_executor()
 
         lambda_hooks.start_docker_executor.run(container_config, self.function_version)
--- a/localstack/services/lambda_/invocation/lambda_models.py
+++ b/localstack/services/lambda_/invocation/lambda_models.py
@@ -41,7 +41,7 @@ from localstack.aws.connect import connect_to
 from localstack.constants import AWS_REGION_US_EAST_1
 from localstack.services.lambda_.api_utils import qualified_lambda_arn, unqualified_lambda_arn
 from localstack.utils.archives import unzip
-from localstack.utils.strings import long_uid
+from localstack.utils.strings import long_uid, short_uid
 
 LOG = logging.getLogger(__name__)
 
@@ -598,6 +598,13 @@ class Function:
 
     def latest(self) -> FunctionVersion:
         return self.versions["$LATEST"]
+
+    # HACK to model a volatile variable that should be ignored for persistence
+    def __post_init__(self):
+        # Identifier unique to this function and LocalStack instance.
+        # A LocalStack restart or persistence load should create a new instance id.
+        # Used for retaining invoke queues across version updates for $LATEST, but separate unrelated instances.
+        self.instance_id = short_uid()
 
 
 class ValidationException(CommonServiceException):
--- a/localstack/services/lambda_/invocation/lambda_models.py
+++ b/localstack/services/lambda_/invocation/lambda_models.py
@@ -606,6 +606,18 @@ class Function:
         # Used for retaining invoke queues across version updates for $LATEST, but separate unrelated instances.
         self.instance_id = short_uid()
 
+    def __getstate__(self):
+        """Ignore certain volatile fields for pickling.
+        # https://docs.python.org/3/library/pickle.html#handling-stateful-objects
+        """
+        # Copy the object's state from self.__dict__ which contains
+        # all our instance attributes. Always use the dict.copy()
+        # method to avoid modifying the original state.
+        state = self.__dict__.copy()
+        # Remove the volatile entries.
+        del state["instance_id"]
+        return state
+
 
 class ValidationException(CommonServiceException):
     def __init__(self, message: str):
--- a/localstack/services/lambda_/invocation/version_manager.py
+++ b/localstack/services/lambda_/invocation/version_manager.py
@@ -3,7 +3,6 @@ import logging
 import threading
 import time
 from concurrent.futures import Future
-from typing import TYPE_CHECKING
 
 from localstack import config
 from localstack.aws.api.lambda_ import (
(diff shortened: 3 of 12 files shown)


thrau:
removing myself as reviewer as i was added only because of the config.py change

localstack/services/lambda_/invocation/docker_runtime_executor.py
```diff
@@ -363,7 +363,8 @@ def start(self, env_vars: dict[str, str]) -> None:
             container_config.dns = config.LAMBDA_DOCKER_DNS
         else:
             if dns_server.is_server_running():
-                # Enable transparent endpoint injection by setting DNS to the embedded DNS re-writer in the Go init.
```
dfangl, line 366: Great catch!

localstack/services/lambda_/invocation/lambda_models.py
```diff
@@ -599,6 +599,25 @@ class Function:
     def latest(self) -> FunctionVersion:
         return self.versions["$LATEST"]
 
+    # HACK to model a volatile variable that should be ignored for persistence
```
dfangl, line 602: I am also not 100% satisfied with this hack, but I think for now it is the best solution. Let's hope we can find a way to get by without this hack in the future :)

localstack/services/lambda_/invocation/version_manager.py
```diff
@@ -53,8 +49,6 @@ class LambdaVersionManager:
     state: VersionState | None
     provisioned_state: ProvisionedConcurrencyState | None  # TODO: remove?
     log_handler: LogHandler
-    # TODO not sure about this backlink, maybe a callback is better?
-    lambda_service: "LambdaService"
```
dfangl, line 57: :tada:

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -2601,6 +2646,156 @@ def test_lambda_handler_update(self, aws_client, create_lambda_function, snapsho
...
+        )
+
+        errored = False
+
+        def _update_function():
+            nonlocal errored
+            try:
+                # Make it very likely that the invocation is being processed
+                # In LocalStack, Lambda image pulling could potentially affect the exact timing,
+                # but at least the invocation runtime startup should be in progress.
```
dfangl, line 2672: I think this comment is misleading. Image pulling should have already taken place during the transition to Active. This is maybe just for the container creation?

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -2601,6 +2646,156 @@ def test_lambda_handler_update(self, aws_client, create_lambda_function, snapsho
...
+        assert reserved_concurrency_response["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+        payload = {"request_prefix": f"{1:02}-sleep", "sleep_seconds": 22}
+        aws_client.lambda_.invoke(
+            FunctionName=function_name,
+            InvocationType="Event",
+            Payload=json.dumps(payload),
+        )
+        # Make it very likely that the first request is being processed before sending further requests without sleeps.
+        # We could use a side effect (e.g., via S3) to signal a status update if Lambda image pulling causes flakiness,
```
dfangl, line 2754: Same question here regarding pulling vs container creation

dfangl (approved):
I think this is a good solution for now.

We should definitely tackle the lambda function deletion locks and the aborted invocations soon!

I just had some minor questions regarding some comments, otherwise approved from my side :)

tests/aws/services/lambda_/test_lambda.py
joe4dev: good catch. I updated the comment.
Acquiring the lock is actually pretty fast for invokes 🚀 Container creation happens later and does not matter because we always send invokes to environments we spawn on demand (i.e., no optimization step here).

tests/aws/services/lambda_/test_lambda.py
joe4dev: updated as well 👍

localstack/services/lambda_/invocation/lambda_models.py
joe4dev: yeah, pretty ugly hack 😢 
We currently don't have a Python object such as the VersionManager at function level. Alternative hacks:
* Use a dict to track them in the lambda service, but passing it down into the event managers is gonna be awkward
* Add it to the version manager of a specific version (e.g., v1, $LATEST is annoying because then it needs to migrate)

localstack/services/lambda_/invocation/lambda_models.py
```diff
@@ -41,7 +41,7 @@
 from localstack.constants import AWS_REGION_US_EAST_1
 from localstack.services.lambda_.api_utils import qualified_lambda_arn, unqualified_lambda_arn
 from localstack.utils.archives import unzip
-from localstack.utils.strings import long_uid
+from localstack.utils.strings import long_uid, short_uid
```
joe4dev, line 44: @giograno Does this hack invalidate all Lambda CloudPods due to a changed import 🤔

PR #10833, opened 2024-05-16
update test durations

alexrashed:
## Motivation
We use `pytest-split` to balance the tests across our parallel test runners in our GitHub workflows (see https://github.com/localstack/localstack/pull/9003).
This test splitting is based on test duration data. Since we constantly add tests, refctor / move them and change service implementations, this test data gets outdated.
This PR updates them again to re-balance the parallel test execution times.

## Changes
- Updates the `.test_durations` with the latest test data.


.test_durations
```diff
@@ -968,16 +968,3389 @@
     "tests/aws/services/lambda_/test_lambda_developer_tools.py::TestHotReloading::test_hot_reloading[nodejs18.x]": 3.06303496099963,
     "tests/aws/services/lambda_/test_lambda_developer_tools.py::TestHotReloading::test_hot_reloading[python3.9]": 2.900148914000056,
     "tests/aws/services/lambda_/test_lambda_developer_tools.py::TestHotReloading::test_hot_reloading_publish_version": 0.17520181300005788,
-    "tests/aws/services/lambda_/test_lambda_integration_dynamodbstreams.py::TestDynamoDBEventSourceMapping::test_deletion_event_source_mapping_with_dynamodb": 6.989548434999961,
```
silv-io, line 971: This one seems suspicious 🤔 Did it get more testing added to it?

silv-io (approved):
LGTM!

On a side thought: Considering that we've got that many new tests. Would it maybe make sense to start increasing the number of "shards" to our split?

.test_durations
alexrashed: Could you add a bit more context please? Why is it suspicious? What do you mean by "did it get more testing added to it"?

alexrashed:
> On a side thought: Considering that we've got that many new tests. Would it maybe make sense to start increasing the number of "shards" to our split?

That depends on the runtime of the tests and if they are a limiting factor. Here in our OSS project, this pipeline ise only an indicator if that changes here might break stuff down the line. So I don't really see this being a pain point right now.
For our downstream projects: Sure, there it might make sense, but this is already being evaluated / being worked on.

.test_durations
silv-io: It went from roughly 7 seconds to 65 seconds. That's quite the increase. I was wondering if we have some kind of performance degradation here, or if the test was extended so that it takes 10 times longer now.

alexrashed: Well, that's hard to say honestly. We'll see the results in the upcoming pipeline runs. If the data is too far off, we would see that in the balancing (that they are still not well balanced). Then we can dig deeper.

PR #10835, opened 2024-05-16
Revert "APIGW: Default to an empty dict when the provided body for an API Gateway step function is an empty string"

dfangl:
Reverts localstack/localstack#10816

The reverted PR causes an -ext test to fail, which uses the following request template: `Action=SendMessage&MessageBody=$util.urlEncode($input.body)`.

An empty body is defaulted to a dict, which cannot be urlencoded. We either need special handling, a different default value, or a completely different fix for the original issue the PR tried to address.

cc @marcciosilva

### DIFF
--- a/localstack/services/apigateway/templates.py
+++ b/localstack/services/apigateway/templates.py
@@ -190,7 +190,7 @@ class ApiGatewayVtlTemplate(VtlTemplate):
             namespace["stageVariables"] = stage_var
         input_var = variables.get("input") or {}
         variables = {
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Sorry for not catching it, it seems changing things in the Velocity logic is quite tricky, or maybe the fix was in the step functions integration directly to handle the case. Not too sure, but we'll investigate! Sorry this wasn't caught, and thanks for investigating!

PR #10742, opened 2024-04-29
Bump moto-ext to 5.0.6.post2

viren-nadkarni:
Brings in:
- https://github.com/getmoto/moto/pull/7639
- https://github.com/getmoto/moto/pull/7620
- https://github.com/getmoto/moto/pull/7679
- https://github.com/getmoto/moto/pull/7680

Closes:
- https://github.com/localstack/localstack/issues/10315

CI:
- Random credentials https://app.circleci.com/pipelines/github/localstack/localstack/24577/workflows/3f4a3096-5e17-464d-a79f-1e63446aba6b :green_circle: Passing
- Ext Integration Tests # 6349 :yellow_circle:  Running

PR #10837, opened 2024-05-16
mitigate CI timeouts by increasing parallelism

alexrashed:
## Motivation
The run time of our pipeline increased dramatically somewhat around the time of our latest bump of moto (with https://github.com/localstack/localstack/pull/10742).
This is causing some timeouts in the pipeline due to the hard limit of 60 minutes for a job in CircleCI.
While we are still investigating the issue, this PR tries to unblock the pipeline by adding another parallel runner in the meantime.

## Changes
- Adds another runner to the CircleCI integration tests for until we identified the issue.

/cc @viren-nadkarni

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -366,7 +366,7 @@ jobs:
       image: << parameters.machine_image >>
     resource_class: << parameters.resource_class >>
     working_directory: /tmp/workspace/repo
-    parallelism: 4
+    parallelism: 5
     environment:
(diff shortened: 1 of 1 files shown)


silv-io (approved):
LGTM! 

Running independent tests should be embarrassingly parallel so increasing the runners is a good shout :)

dominikschubert (approved):
LGTM, good call unblocking us for now

PR #10789, opened 2024-05-08
Feature/eventbridge v2 add input transformer

maxhoheiser:
## Motivation
This PR adds input transformers https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html to the v2 provider. Input transformers use either predefined placeholders or custom placeholders defined in input template or a combination of both, that are used to re-write the returned event. 
Predefined placeholders can return either a string or a dict ( dicts are only allowed in json input maps)

The event that is returned is the input map that gets populated with the placeholders described before with the variables from the original event and rule. The input map can be either a plain string or a json string that results in a return dict. 

## Changes
Add input transformer to transform incoming event based on input path and input map.
Add predefined placeholders.
Add tests for input path string and json.
Add test for predefined placeholders.
Add test for combining input path and input transfomrer.

PR #10840, opened 2024-05-16
Feature: Eventbridge v2: Add tagging

maxhoheiser:
## Motivation
Event bridge allows for adding tags to event buses and rules: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-tagging.html
This PR introduces the ability to create event buses and rules with tags, add and remove tags to existing buses or rules and list tags for buses and rules.

## Changes
Use most widely used pattern of handling tags from multiple other services by extend store with TagDict mapping resource arn to the dictionary of tags.
Exten test suite to cover all different options of tagging.

### DIFF
--- a/localstack/services/events/models.py
+++ b/localstack/services/events/models.py
@@ -16,7 +16,6 @@ from localstack.aws.api.events import (
     RuleName,
     RuleState,
     ScheduleExpression,
-    Tag,
     TagList,
     Target,
     TargetId,
@@ -27,6 +26,7 @@ from localstack.services.stores import (
     CrossRegionAttribute,
     LocalAttribute,
 )
+from localstack.utils.tagging import TaggingService
 
 TargetDict = dict[TargetId, Target]
 
@@ -86,15 +86,13 @@ class EventBus:
 
 EventBusDict = dict[EventBusName, EventBus]
 
-TagDict = dict[Arn, Tag]
-
 
 class EventsStore(BaseStore):
     # Map of eventbus names to eventbus objects. The name MUST be unique per account and region (works with AccountRegionBundle)
     event_buses: EventBusDict = LocalAttribute(default=dict)
 
     # Maps resource ARN to tags
-    TAGS: TagDict = CrossRegionAttribute(default=dict)
+    TAGS: TaggingService = CrossRegionAttribute(default=TaggingService)
 
 
 events_store = AccountRegionBundle("events", EventsStore)
--- a/localstack/services/events/models.py
+++ b/localstack/services/events/models.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass, field
+from enum import Enum
 from typing import Optional, TypeAlias, TypedDict
 
 from localstack.aws.api.core import ServiceException
@@ -15,6 +16,7 @@ from localstack.aws.api.events import (
     RuleName,
     RuleState,
     ScheduleExpression,
+    Tag,
     TagList,
     Target,
     TargetId,
@@ -22,6 +24,7 @@ from localstack.aws.api.events import (
 from localstack.services.stores import (
     AccountRegionBundle,
     BaseStore,
+    CrossRegionAttribute,
     LocalAttribute,
 )
 
@@ -83,10 +86,15 @@ class EventBus:
 
 EventBusDict = dict[EventBusName, EventBus]
 
+TagDict = dict[Arn, Tag]
+
 
 class EventsStore(BaseStore):
     # Map of eventbus names to eventbus objects. The name MUST be unique per account and region (works with AccountRegionBundle)
     event_buses: EventBusDict = LocalAttribute(default=dict)
+
+    # Maps resource ARN to tags
+    TAGS: TagDict = CrossRegionAttribute(default=dict)
 
 
 events_store = AccountRegionBundle("events", EventsStore)
@@ -119,3 +127,8 @@ class FormattedEvent(TypedDict):
 
 
 TransformedEvent: TypeAlias = FormattedEvent | dict | str
+
+
+class ResourceType(Enum):
+    EVENT_BUS = "event_bus"
+    RULE = "rule"
--- a/localstack/services/events/provider.py
+++ b/localstack/services/events/provider.py
@@ -541,12 +541,8 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         if not resource_type:
             pass  # TODO handle error for tagging not rule or event_bus
         self._check_resource_exists(resource_arn, resource_type, store)
-        try:
-            tags = store.TAGS.get(resource_arn, {})
-            tags_list = [{"Key": key, "Value": value} for key, value in tags.items()]
-        except KeyError:
-            raise ResourceNotFoundException(f"Resource not found: '{resource_arn}'")
-        return ListTagsForResourceResponse(Tags=tags_list)
+        tags = store.TAGS.list_tags_for_resource(resource_arn)
+        return ListTagsForResourceResponse(tags)
 
     @handler("TagResource")
     def tag_resource(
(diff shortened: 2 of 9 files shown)


localstack/services/events/models.py
```diff
@@ -83,11 +86,16 @@ def __post_init__(self):
...
 
+TagDict = dict[Arn, Tag]
+
 
 class EventsStore(BaseStore):
     # Map of eventbus names to eventbus objects. The name MUST be unique per account and region (works with AccountRegionBundle)
     event_buses: EventBusDict = LocalAttribute(default=dict)
 
+    # Maps resource ARN to tags
+    TAGS: TagDict = CrossRegionAttribute(default=dict)
```
joe4dev, line 97: We have a `TaggingService` that implements tagging transparently (noticed by looking into DMS models as one of our new implementations). Is that applicable here or does different handling based on the resource type require a special case?

localstack/services/events/provider.py
```diff
@@ -143,6 +156,21 @@ def format_event(event: PutEventsRequestEntry, region: str, account_id: str) ->
     return formatted_event
 
 
+def get_bus_or_rule(arn: Arn) -> ResourceType:
+    parsed_arn = parse_arn(arn)
+    service = parsed_arn["resource"].split("/", 1)[0]
```
joe4dev, line 161: `service` could be misleading here given that `parsed_arn["service"]` exists. `resource_type` is more accurate here following the ARN definition: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#arns-syntax

localstack/services/events/provider.py
```diff
@@ -143,6 +156,21 @@ def format_event(event: PutEventsRequestEntry, region: str, account_id: str) ->
     return formatted_event
 
 
+def get_bus_or_rule(arn: Arn) -> ResourceType:
```
joe4dev, line 159: maybe `get_resource_type()`

localstack/services/events/provider.py
```diff
@@ -477,6 +514,61 @@ def put_partner_events(
...
+    @handler("TagResource")
+    def tag_resource(
+        self, context: RequestContext, resource_arn: Arn, tags: TagList, **kwargs
+    ) -> TagResourceResponse:
+        # each tag key must be unique
+        # https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html#tag-best-practices
+        store = self.get_store(context)
+        resource_type = get_bus_or_rule(resource_arn)
+        if not resource_type:
+            pass  # TODO handle error for tagging not rule or event_bus
```
joe4dev, line 546: any particular reason why we don't handle that? It looks like an easy case

localstack/services/events/provider.py
```diff
@@ -477,6 +514,61 @@ def put_partner_events(
...
+        self, context: RequestContext, resource_arn: Arn, tag_keys: TagKeyList, **kwargs
+    ) -> UntagResourceResponse:
+        store = self.get_store(context)
+        resource_type = get_bus_or_rule(resource_arn)
+        if not resource_type:
+            pass  # TODO handle error for tagging not rule or event_bus
+        self._check_resource_exists(resource_arn, resource_type, store)
+        try:
+            resource_tags = store.TAGS.get(resource_arn, {})
+        except KeyError:
```
joe4dev, line 564: When can a dict.get() raise a KeyError? This seems untested

localstack/services/events/provider.py
```diff
@@ -477,6 +514,61 @@ def put_partner_events(
...
+            raise ResourceNotFoundException(f"Resource not found: '{resource_arn}'")
+        for tag_key in tag_keys:
+            try:
```
joe4dev, line 567: style: `resource_tags.pop(tag_key, None)` would be more pythonic and less verbose

localstack/services/events/provider.py
```diff
@@ -596,12 +688,21 @@ def _get_limited_dict_and_next_token(
...
+        self, resource_arn_or_name: EventBusNameOrArn | RuleArn | None
     ) -> EventBusName:
         """Return the event bus name. Input can be either an event bus name or ARN."""
-        if not event_bus_name_or_arn:
+        if not resource_arn_or_name:
+            return "default"
+        if "arn:aws:events" not in resource_arn_or_name:
+            return resource_arn_or_name
+        resource_type = get_bus_or_rule(resource_arn_or_name)
+        # TODO how to deal with / in event bus name or rule name
```
joe4dev, line 699: This seems like a recurring theme (see https://github.com/localstack/localstack/pull/10789#discussion_r1600181172). Can we get that fixed before we introduce this bug at too many different places?

localstack/services/events/provider.py
```diff
@@ -596,12 +688,21 @@ def _get_limited_dict_and_next_token(
         return limited_dict, next_token
 
     def _extract_event_bus_name(
```
joe4dev, line 690: This method is getting increasingly complex. Could be a candidate for a unit test to cover edge cases easily (as long as we can ensure AWS parity).

tests/aws/services/events/test_events_tags.py
```diff
@@ -0,0 +1,289 @@
...
+import pytest
+
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+from tests.aws.services.events.helper_functions import is_v2_provider
+from tests.aws.services.events.test_events import TEST_EVENT_PATTERN
+
+
+@markers.aws.validated
+@pytest.mark.parametrize("event_bus_name", ["event_bus_default", "event_bus_custom"])
```
joe4dev, line 12: nice coverage 💯

tests/aws/services/events/test_events_tags.py
```diff
@@ -0,0 +1,289 @@
...
+    with pytest.raises(aws_client.events.exceptions.ResourceNotFoundException) as error:
+        aws_client.events.untag_resource(
+            ResourceARN=resource_arn,
+            TagKeys=[tag_key_1],
+        )
+    snapshot.match("untag_not_existing_resource_error", error)
+
+
+@markers.aws.validated
+@pytest.mark.parametrize("event_bus_name", ["event_bus_default", "event_bus_custom"])
```
joe4dev, line 128: test design note to consider for the future: There's always a trade-off between test complexity and code duplication. That looks like an example to consider whether parametrization is worth the extra complexity given the conditionals cover most of the test. Sometimes duplication or a helper for asserts can more appropriate.

joe4dev (approved):
The tests are quite comprehensive. Nice coverage 👍 

My main question is whether it would be feasible to reuse the tagging service (maybe not because the resource type differentiation is too specific)?
I added some suggestions to simplify the code and resolve potential smaller bugs.

localstack/services/events/models.py
maxhoheiser: used TaggingService in: https://github.com/localstack/localstack/pull/10840/commits/df9b76c5dadd4ba6b2496aacfc21bc1b8bc0fe85

The only issue I see is, that the pattern other services use is to store the TagginService objects in the store, which violates a bit of what we discussed a couple of weeks ago that data and functions should be strictly separated. This comes back to the issue that the AcountRegionBundle needs additional logic that handles this separation which would make a lot of providers cleaner.

localstack/services/events/provider.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10840/commits/2ccecbc2d1b61126a3528c7296862365c04b0126

localstack/services/events/provider.py
maxhoheiser: fixed with using the TaggingService

localstack/services/events/provider.py
maxhoheiser: not needed anymore due to using TaggingService

localstack/services/events/provider.py
maxhoheiser: Not really, I do not have any approach to fixing this, and we seam to also assume in many other providers that names will not contain /

localstack/services/events/provider.py
maxhoheiser: refactored in: https://github.com/localstack/localstack/pull/10840/commits/ab40f18492de93569b5c43202f5aac1cccfd089c

localstack/services/events/provider.py
maxhoheiser: fixed in: https://github.com/localstack/localstack/pull/10840/commits/bd3f9db9848e4296f9c672be1f388f252c0eb56b

localstack/services/events/models.py
joe4dev: Very neat ✨
IMHO, that's a compromise worth taking vs. re-implementing the same thing for every service. The likelihood of a major change in tagging is very low.

PR #10842, opened 2024-05-16
Cleanup code mounting dev scripts

simonrw:
## Motivation

We have some make targets and helper scripts that rely on mounting the local code into the right place into the Docker container. We are planning to restructure the code base to change the code layout, and as such these helpers need to be updated.

However, we handle a lot of determining which paths should be mounted into the container with our helper script `python -m localstack.dev.run`. We should centralise any knowledge of where code is located into a single place, this module, and remove any other commands.

## Changes

- Remove make targets that mount code

## Migration guide

- `Makefile#docker-run`:
    - run `python -m localstack start`
- `Makefile#docker-mount-run`:
    - run `python -m localstack.dev.run`
- `Makefile#test-docker*`:
    - run `python -m localstack.dev.run --mount-entrypoints --mount-dependencies -v $PWD/tests:/opt/code/localstack/tests -- .venv/bin/python -m pytest ...`
- `bin/localstack-start-docker-dev.sh`:
    - run `python -m localstack.dev.run`


thrau (approved):
i like this PR a lot

alexrashed (approved):
🧹

PR #10849, opened 2024-05-20
CFn: DAG based deploy order

simonrw:
## Motivation

When deploying stacks, our CloudFormation engine currently performs a number of retries to deploy a stack. This is required because the implementation does not take dependency order into account when deploying resources. Resources are strictly taken from the template in the order they are defined. We retry up to 30 times for deployments and 10 times for deletions.

This means that a stack with 31 resources defined in the reverse order of their dependencies will fail to deploy.

This is a particularly bad example, yes, but it is a fragile and brittle implementation that (especially on deletes) leads to potential breakages.

If we had a way of ordering dependencies so that we could iterate through them in order, then we would not need a retry process, and the behaviour would conform to the AWS behaviour. This would also be a first step towards concurrent deployment in our CFn engine.

## Changes

* Introduce `order_resources` and `order_changes` which order resources and changes based on the dependencies of each resource (or associated change).
* Extract V2 template deployer which
    * specialises two methods, `apply_changes_in_loop` and `delete_stack`
    * new methods use the dependency sorting, and remove the outer loop retrying the deployment process a number of times
* Add `CFN_LEGACY_TEMPLATE_DEPLOYER` flag to revert back to old behaviour
* Perform initial validation in `create_changeset` where references to non-existent resources leads to a validation error during the call to `create_changeset` as per AWS, where before it would fail the stack deploy.
* Fix `resolve_dependencies` for `Fn::Join` use case
* Fix resources which return `None` in a progress event from the `delete` handlers:
    * `CDK::Metadata`
    * `AWS::IAM::User`
    * `AWS::Lambda::Alias`
* Add test that permutes a 3-chain of resources to cover all deployment orders (this test fails with the old implementation). This test also checks the order of resource _deletions_ as well.


localstack/services/cloudformation/engine/template_utils.py
```diff
@@ -74,6 +74,9 @@ def resolve_dependencies(d: dict, evaluated_conditions: dict[str, bool]) -> set[
                     items = items.union(resolve_dependencies(item, evaluated_conditions))
             else:
                 pass
+    elif isinstance(d, list):
```
simonrw, line 77: Added to support e.g.

```yaml
Fn::Join:
  - ""
  - - https://
    - Ref: LsApi42D61DD0
    - .execute-api.
    - Ref: AWS::Region
    - "."
    - Ref: AWS::URLSuffix
    - /
    - Ref: LsApiDeploymentStageprod82D8C5E7
    - /

```

localstack-core/localstack/services/iam/resource_providers/aws_iam_user.py
```diff
@@ -133,7 +133,7 @@ def delete(
         """
         iam_client = request.aws_client_factory.iam
         iam_client.delete_user(UserName=request.desired_state["Id"])
-        return ProgressEvent(status=OperationStatus.SUCCESS, resource_model=None)
+        return ProgressEvent(status=OperationStatus.SUCCESS, resource_model=request.previous_state)
```
pinzon, line 136: So, this is the appropriate value from now on right?

tests/aws/services/cloudformation/api/test_stacks.py
```diff
@@ -788,3 +790,90 @@ def test_describe_stack_events_errors(aws_client, snapshot):
...
+        # TODO
+        "$..PhysicalResourceId",
+        # TODO
+        "$..ResourceProperties",
+    ]
+)
+@pytest.mark.parametrize(
+    "deploy_order", TEMPLATE_ORDER_CASES, ids=["-".join(vals) for vals in TEMPLATE_ORDER_CASES]
+)
+def test_stack_deploy_order(deploy_cfn_template, aws_client, snapshot, deploy_order: tuple[str]):
```
pinzon, line 811: Great test 👌

tests/aws/services/cloudformation/engine/test_references.py
```diff
@@ -57,8 +57,8 @@ def test_fn_sub_cases(self, deploy_cfn_template, aws_client, snapshot):
         snapshot.match("outputs", deployment.outputs)
 
 
-@markers.aws.only_localstack
-def test_useful_error_when_invalid_ref(deploy_cfn_template):
+@markers.aws.validated
```
pinzon, line 60: 🎊

pinzon (approved):
LGTM👌

thrau (commented):
nice! this is great, a DAG is definitely the better concept for the deployer :raised_hands: 

just curious: have you looked into using cfnlint? i played around with this over a year ago so it's been a while, but IIRC it can create a graph from a cfn template that you can then run a depth first search, which i assume then gives the correct the deployment order?

simonrw:
> nice! this is great, a DAG is definitely the better concept for the deployer 🙌
> 
> just curious: have you looked into using cfnlint? i played around with this over a year ago so it's been a while, but IIRC it can create a graph from a cfn template that you can then run a depth first search, which i assume then gives the correct the deployment order?

I think it would be a definite improvement to use something like `networkx` to handle the graph-like nature. We already implement determining resource dependencies, but there could be something good to learn from with `cfnlint`. They do a lot of parsing of the template that we could learn from as well.

After investigating, I found that we don't actually use `cfnlint`, but it's a transitive dependency from `moto`. I don't think we should rely on `cfnlint` as it's a large package, and it might be nice to remove it in the future.

I want to get a first working version in place (even if hacky) which we can then refactor in a follow up PR.

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -791,7 +802,102 @@ class ChangeConfig(TypedDict):
...
+            if change["ResourceChange"]["LogicalResourceId"] == logical_resource_id:
+                sorted_changes.append(change)
+                break
+    assert len(sorted_changes) > 0
+    if reverse:
+        sorted_changes = sorted_changes[::-1]
+    return sorted_changes
+
+
+class TemplateDeployerBase(ABC):
```
dominikschubert, line 900: While I appreciate it, I also really hope we can remove this soon*ish :grimacing:

dominikschubert (approved):
Great change! Let's get it over the line and see if we encounter any issues with it. 

I haven't checked every line of the copied TemplateDeployerBase subclasses and focused mostly on the other parts where I didn't find any obvious issues :partying_face:  We also already did a walkthrough of the changes recently, so from my side it's a go :+1:

localstack-core/localstack/services/cdk/resource_providers/cdk_metadata.py
```diff
@@ -70,7 +70,7 @@ def delete(
 
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
-            resource_model=None,
+            resource_model=request.previous_state,
```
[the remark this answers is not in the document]
simonrw: I was fixing warnings with teardowns, and found some warnings to do with the `resource_model` of the resulting `ProgressEvent` from the delete method be `None` when trying to access the physical resource id. I think it is within the scope to reduce the stack delete warnings, as this is evidence that the deploy order is correct.

Issue #10844, opened 2024-05-20
bug: Presigned URL failed on GET for downloading


cchentyro:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

URL:
```
127.0.0.1:32848/<bucket-name>/presignTestFile?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20240517T090340Z&X-Amz-SignedHeaders=host&X-Amz-Credential=accessKeyId%252F20240517%252Fap-southeast-2%252Fs3%252Faws4_request&X-Amz-Expires=604800&X-Amz-Signature=ce80156d648142563ef1e3be3f41a3b4b901ac14c3add2db8f1adc99b50ab173

```

Error:
```
2024-05-17T09:08:44.719 ERROR --- [et.reactor-0] l.aws.handlers.logging     : exception during call chain
Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack/aws/handlers/presigned_url.py", line 23, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 201, in __call__
    validate_presigned_url_s3v4(context)
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 425, in validate_presigned_url_s3v4
    sigv4_context, exception = _find_valid_signature_through_ports(context)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 487, in _find_valid_signature_through_ports
    sigv4_context = S3SigV4SignatureContext(context=context)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 543, in __init__
    region = self._query_parameters["X-Amz-Credential"].split("/")[2]
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### Expected Behavior

It should parse the region from the parameter.

### How are you starting LocalStack?

Custom (please describe below)

[...]

PR #10856, opened 2024-05-21
add S3 pre-signed credentials validation

bentsku:
## Motivation
Following #10844 where the request sent was malformed but the raised exception was an internal one:
```python
2024-05-17T09:08:44.719 ERROR --- [et.reactor-0] l.aws.handlers.logging     : exception during call chain
  File "/opt/code/localstack/localstack/aws/handlers/presigned_url.py", line 23, in __call__
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 201, in __call__
    validate_presigned_url_s3v4(context)
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 425, in validate_presigned_url_s3v4
    sigv4_context, exception = _find_valid_signature_through_ports(context)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 487, in _find_valid_signature_through_ports
    sigv4_context = S3SigV4SignatureContext(context=context)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack/services/s3/presigned_url.py", line 543, in __init__
    region = self._query_parameters["X-Amz-Credential"].split("/")[2]
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

I validated that AWS does not accept double URL encoded `X-Amz-Credential` and saw the proper exception raised, so wanted the error to be more descriptive and add parity to LocalStack. 

## Changes
- add validation before splitting and getting data from `X-Amz-Credential` query string parameter in pre-signed URL. This was done in a non intrusive way, only done where we would directly access the data. We could maybe do a little parity push on pre-signed URL all around and do better validation of incoming mandatory parameters.
- added a test to check the exception on AWS

### DIFF
--- a/localstack/services/s3/presigned_url.py
+++ b/localstack/services/s3/presigned_url.py
@@ -232,7 +232,7 @@ def get_credentials_from_parameters(parameters: dict) -> PreSignedCredentials:
         credential_value = parameters.get(
             "X-Amz-Credential", parameters.get("x-amz-credential", "")
         ).split("/")
-        if len(credential_value):
+        if credential_value:
             access_key_id = credential_value[0]
 
     if not access_key_id:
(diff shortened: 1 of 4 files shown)


steffyP (approved):
Thanks for tackling, this is a very neat fix!

Issue #9974, opened 2024-01-02
bug: CloudPod of a Lambda function with hot-reloading contain absolute path


joe4dev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating a CloudPod of a Lambda function with [hot-reloading](https://docs.localstack.cloud/user-guide/tools/lambda-tools/hot-reloading/) encodes an absolute path. Hence, the CloudPod is not cross-machine compatible and loading the CloudPod on another machine leads to such errors:

Function invocation:

```log
2023-11-29 16:14:43 2023-11-29T15:14:43.796 ERROR --- [   asgi_gw_8] l.s.lambda_.provider       : Error while invoking lambda
2023-11-29 16:14:43 Traceback (most recent call last):
2023-11-29 16:14:43   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/docker/api/client.py", line 268, in _raise_for_status
2023-11-29 16:14:43     response.raise_for_status()
2023-11-29 16:14:43   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
2023-11-29 16:14:43     raise HTTPError(http_error_msg, response=self)
2023-11-29 16:14:43 requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http+docker://localhost/v1.41/containers/5f77f01ced1e56cdd2b5fbc9863b73390501f00034f0d7c0debf70a71a564383/start
```

Docker mount denied error:

> 2023-11-29 16:14:43 docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.41/containers/5f77f01ced1e56cdd2b5fbc9863b73390501f00034f0d7c0debf70a71a564383/start: Internal Server Error ("b'Mounts denied: \nThe path /home/runner/work/localstack-platform/localstack-platform is not shared from the host and is not known to Docker.\nYou can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.\nSee https://docs.docker.com/desktop/mac for more info.'")

### Expected Behavior

Options:

a) Convert any hot-reloaded Lambda function (i.e., `HotReloadingCode(ArchiveCode)`) into a regular S3-backed Lambda function (i.e., `S3Code(ArchiveCode)`) upon persisting.
[...]

PR #10857, opened 2024-05-21
Add support for environment variable placeholders in hot-reloading paths

dfangl:
## Motivation
When specifying a path for hot-reloading, we require an absolute path on the host, as it is expected for docker bind mounts.
However, this has an important limitation: Templates with a path defined or Cloudpods cannot be easily shared between systems where the code location differs.

For example, a checked out repository with the code at `/home/user1/shared-repository/src` could be checked out at `/home/user2/repositories/shared-repository/src` at another users machine.

To enable sharing of templates and cloudpods with hot-reloading paths between different machines without modification, this PR will add the option to use environment variables expansion within the hot-reloading paths. A hot-reloading path like `$HOST_DIR/src` can now be used in the templates or cloud pods, if an environment variable is set in each LocalStack instance using that path, with the concrete absolute path part as its value.

This variable could be defined relative to the docker compose project, for example, using a template like this:

```yaml
version: "3.8"

services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    environment:
      # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
      - DEBUG=${DEBUG:-0}
      - HOST_DIR=${PWD}
    volumes:
```

## Changes
* The S3Key for a hot-reloading path now accepts placeholders for environment variables using `$KEY` or `${KEY}`

### DIFF
--- a/localstack/services/lambda_/invocation/lambda_service.py
+++ b/localstack/services/lambda_/invocation/lambda_service.py
@@ -3,6 +3,7 @@ import concurrent.futures
 import dataclasses
 import io
 import logging
+import os.path
 import random
 import uuid
 from concurrent.futures import Executor, Future, ThreadPoolExecutor
@@ -549,12 +550,26 @@ def store_lambda_archive(
     )
 
 
-def create_hot_reloading_code(path: str) -> HotReloadingCode:
-    # TODO extract into other function
-    if not PurePosixPath(path).is_absolute() and not PureWindowsPath(path).is_absolute():
+def assert_hot_reloading_path_absolute(path: str) -> None:
+    """
+    Check whether a given path is an absolute path, either posix or windows.
+    Will expand variables in the path before the check.
+
+    :param path: Posix or windows path, potentially containing environment variable placeholders.
+    """
+    # expand variables in path before checking for an absolute path
+    expanded_path = os.path.expandvars(path)
+    if (
+        not PurePosixPath(expanded_path).is_absolute()
+        and not PureWindowsPath(expanded_path).is_absolute()
+    ):
         raise InvalidParameterValueException(
             f"When using hot reloading, the archive key has to be an absolute path! Your archive key: {path}",
         )
+
+
+def create_hot_reloading_code(path: str) -> HotReloadingCode:
+    assert_hot_reloading_path_absolute(path)
     return HotReloadingCode(host_path=path)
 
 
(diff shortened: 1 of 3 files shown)


localstack/services/lambda_/invocation/lambda_service.py
```diff
@@ -549,12 +550,26 @@ def store_lambda_archive(
     )
 
 
-def create_hot_reloading_code(path: str) -> HotReloadingCode:
-    # TODO extract into other function
-    if not PurePosixPath(path).is_absolute() and not PureWindowsPath(path).is_absolute():
+def assert_hot_reloading_path_absolute(path: str) -> None:
+    """
+    Check whether a given path is an absolute path, either posix or windows.
```
joe4dev, line 555: nit: it checks whether it can be expanded into an absolute path, but does not need to be an absolute path itself 💡

joe4dev: Maybe, `assert_hot_reloading_path_absolute_resolves_absolute` ?

joe4dev (approved):
Great flexible idea giving full control to the user @dfangl 💡 

Can we add an example to the [Hot Reloading docs](https://docs.localstack.cloud/user-guide/lambda-tools/hot-reloading/)?

This solution enables a workaround for the CloudPod/Persistence bug https://github.com/localstack/localstack/issues/9974
@giograno Do you think this solution is sufficient to resolve https://github.com/localstack/localstack/issues/9974?

localstack/services/lambda_/invocation/lambda_service.py
joe4dev: fair point, let's leave the method. I just got confused by "Check whether a given path is an absolute path"
because `$HEAD_DIR/my/relative/suffix/path` is an invalid path by itself but is valid if it resolved to an absolute path

dfangl: Hmm, I get it now. Will adjust the docstring to make that clearer.

giograno:
> This solution enables a workaround for the CloudPod/Persistence bug #9974 @giograno Do you think this solution is sufficient to resolve #9974?

That's a great solution 💯 To me, it solves https://github.com/localstack/localstack/issues/9974. I am updating the docs!

joe4dev:
> > This solution enables a workaround for the CloudPod/Persistence bug #9974 @giograno Do you think this solution is sufficient to resolve #9974?
> 
> That's a great solution 💯 To me, it solves #9974. I am updating the docs!

Great, thank you @giograno. Can you link to the updated docs and close #9974?

PR #10858, opened 2024-05-21
Replace large bytes payloads in http request logging to avoid cluttering the logs

dfangl:
## Motivation
When we started to route ECR requests over our gateway, customers noticed that the entire image blob payload would be logged into the LocalStack logs if `LS_LOG=trace` or `trace-internal`.
This is caused by unconditional logging of the input and output payload in the http request logger.

Since the ECR blobs could get serveral hundred MBs up to multiple GBs big, this leads to severe performance degradation and massive issues in log file size.

We should avoid logging big binary payloads, as we do with the AWS request logger already.

## Changes
* LocalStack will not log binary payloads of HTTP requests if they are bigger than 512 bytes, the current threshold for logging aws bytes payloads as well.

Example log entry in trace logging after the change (for ECR):
```
2024-05-21T15:42:13.190  INFO --- [et.reactor-0] localstack.request.http    : PATCH /v2/repo-1263d7d4/blobs/uploads/dc472cde-4885-469d-a893-e2d3a6864608 => 202; Request(Bytes(29.126MB), headers={'Host': '000000000000.dkr.ecr.us-east-1.localhost.localstack.cloud:4566', 'User-Agent': 'docker/26.1.2 go/go1.22.3 git-commit/ef1912d8b6 kernel/6.8.9-arch1-2 os/linux arch/amd64 UpstreamClient(Docker-Client/26.1.2 \\(linux\\))', 'Transfer-Encoding': 'chunked', 'Content-Type': 'application/octet-stream', 'Accept-Encoding': 'gzip', 'Connection': 'close'}); Response(None, headers={'Content-Type': 'text/plain; charset=utf-8', 'Content-Length': '0', 'Docker-Distribution-Api-Version': 'registry/2.0', 'Docker-Upload-Uuid': 'dc472cde-4885-469d-a893-e2d3a6864608', 'Location': 'http://000000000000.dkr.ecr.us-east-1.localhost.localstack.cloud:4566/v2/repo-1263d7d4/blobs/uploads/dc472cde-4885-469d-a893-e2d3a6864608?_state=yCFl7lYcEJ4qsVBLHlBs4Dvs4V_b5OJ95nREhw8W1Dx7Ik5hbWUiOiJyZXBvLTEyNjNkN2Q0IiwiVVVJRCI6ImRjNDcyY2RlLTQ4ODUtNDY5ZC1hODkzLWUyZDNhNjg2NDYwOCIsIk9mZnNldCI6MjkxMjU5NjMsIlN0YXJ0ZWRBdCI6IjIwMjQtMDUtMjFUMTM6NDI6MTBaIn0%3D', 'Range': '0-29125962', 'X-Content-Type-Options': 'nosniff', 'Date': 'Tue, 21 May 2024 13:42:13 GMT', 'Connection': 'close'})
```

### DIFF
--- a/localstack/logging/format.py
+++ b/localstack/logging/format.py
@@ -2,7 +2,7 @@
 
 import logging
 from functools import lru_cache
-from typing import Dict
+from typing import Any, Dict
 
 from localstack.utils.numbers import format_bytes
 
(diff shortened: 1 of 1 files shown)


thrau (approved):
LGTM as a quick fix. Fundamentally though we should make sure to not keep large payloads in memory at all and use streams/tempspoolfiles instead.

PR #10861, opened 2024-05-21
add cdk utility to import docker images to ecs

pinzon:
## Motivation
This utility helps CDK test to load Docker images into ECR helpful when using services that can use ECR images but no Docker.

## Changes
- new utility


steffyP (commented):
thanks for adding more helper functions 🚀 
just one NIT, the rest LGTM (but I am also not an ECR expert)

steffyP (approved):

PR #10862, opened 2024-05-21
fix S3 Pre-signed POST policy condition casing

bentsku:
## Motivation
As reported in [our Slack community](https://localstack-community.slack.com/archives/CMAFN2KSP/p1716214732559039), we had an issue validating a S3 Pre-signed POST request containing the following condition in the policy: `{"Content-Type": "text/plain"}`. This was due to an issue in the condition matching where we would lower all the form keys, but tried to get the original casing condition key in the form. 

## Changes
- use the lowered condition key to fetch the value from the form
- add a test validating the use case for this condition format (using `dict` format instead of list with `eq`).

### DIFF
--- a/localstack/services/s3/presigned_url.py
+++ b/localstack/services/s3/presigned_url.py
@@ -812,7 +812,7 @@ def _verify_condition(condition: list | dict, form: dict, additional_policy_meta
                 if k == "bucket":
                     return additional_policy_metadata.get("bucket") == val
                 else:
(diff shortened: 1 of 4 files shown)


cloutierMat (approved):
Nice catch and clean fix!

And a short test to to it off! :rocket:

steffyP (approved):
Nice catch! as always: great to have more tests 🚀

PR #10867, opened 2024-05-22
Opt in services to the test selection

silv-io:
## Motivation

The test selection allows future PRs containing strictly service changes to only execute tests corresponding to those services and services that depend on them. However, to enable this, services need to opt in. 

## Changes

This PR offers the opportunity to opt in services to the test selection. These services are opting in with this PR:

- IAM
- STS
- KMS
- Transcribe
- Secretsmanager
- Route53 Resolver
- <add your service as needed>

Furthermore, paths are adjusted to the new structure, prefixed with `localstack-core/`


alexrashed (approved):
Awesome! So many new services opting into the test selection! The amount of API dependencies seems to be a bit low to be honest, but we can grow this over time :)
And thanks for adjusting the paths after https://github.com/localstack/localstack/pull/10800. 🦸🏽

PR #10869, opened 2024-05-22
Using ARN_PARTITION_REWRITING flag causes s3api head-object to provide invalid content-length

Morijarti:
## Motivation
Client using ARN_PARTITION_REWRITING flag has reported that content-length provided by head-object method in s3api is always invalid.

## Changes
Since in WerkzeugResponse class, setting data also causes header content-length to be invalid, I've added skip for partition adjustment of response data if it is empty.

Since we are working on new solution for partitions and this is a temporary fix until then,  I've only did manual tests.

### DIFF
--- a/localstack/aws/handlers/partition_rewriter.py
+++ b/localstack/aws/handlers/partition_rewriter.py
@@ -167,7 +167,10 @@ class ArnPartitionRewriteHandler(Handler):
         response.headers = self._adjust_partition(
             dict(response.headers), request_region=request_region
         )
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM! That's a good catch, effectively, setting `response.data` will be recalculating the content length, because `WerkzeugResponse.automatically_set_content_length` default to `True`. 

The fix looks sound to me, I'll let a partition rewriter expert maybe double approve, but this is great! Great investigation 🕵️

PR #10870, opened 2024-05-22
limit github token permissions, fix cli action for fork runs

ceddy4395:
Hey! 🙂
I've made the following changes to the workflow:

- Use permissions whenever using Github Token
  - Permissions should be used when using the GitHub token as this token full access to the whole repository otherwise meaning that malicious code can tamper with the whole repository.
- Run tests on multiple OS's
  - Running the tests on multiple OS's provides a better test suite as all the functionality will be tested on the supported OS's by the platform.
- Define permissions for workflows with external actions
  - Permissions should be used when running actions written by other developers because there may be security leaks exposed through these actions.
- Avoid uploading artefacts forks
  - Uploading artefacts from forks is usually unwanted behaviour, especially to places outside GitHub. Forks don't have the correct credentials to do this and will therefore fail whilst uploading.

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

### DIFF
--- a/.github/workflows/tests-cli.yml
+++ b/.github/workflows/tests-cli.yml
@@ -70,6 +70,9 @@ env:
   # report to tinybird if executed on master
   TINYBIRD_PYTEST_ARGS: "${{ github.ref == 'refs/heads/master' && '--report-to-tinybird ' || '' }}"
 
+permissions: 
+  contents: read # checkout the repository
+
 jobs:
   cli-tests:
     runs-on: ubuntu-latest
@@ -103,9 +106,11 @@ jobs:
         run: make test
 
   push-to-tinybird:
-    if: always() && github.ref == 'refs/heads/master'
+    if: always() && github.ref == 'refs/heads/master' && github.repository == 'localstack/localstack'
     runs-on: ubuntu-latest
     needs: cli-tests
+    permissions: 
+      actions: read
     steps:
       - name: Push to Tinybird
         uses: localstack/tinybird-workflow-push@v3


ceddy4395:

.github/workflows/tests-cli.yml
```diff
@@ -70,18 +70,22 @@ env:
...
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
+        os: [ubuntu-latest, windows-latest]
     timeout-minutes: 10
     env:
       # Set job-specific environment variables for pytest-tinybird
-      CI_JOB_NAME: ${{ github.job }}-${{ matrix.python-version }}
-      CI_JOB_ID: ${{ github.job }}-${{ matrix.python-version }}
+      CI_JOB_NAME: ${{ github.job }}-${{ matrix.python-version }}-${{matrix.os}}
+      CI_JOB_ID: ${{ github.job }}-${{ matrix.python-version }}-${{matrix.os}}
```
alexrashed, line 88: We do test the compatibility with Windows in `localstack/localstack-cli`. This pipeline is explicitly meant to test the compatibility with the different supported python versions.

.github/workflows/tests-cli.yml
```diff
@@ -103,9 +108,11 @@ jobs:
         run: make test
 
   push-to-tinybird:
-    if: always() && github.ref == 'refs/heads/master'
+    if: always() && github.ref == 'refs/heads/master' && github.repository == 'localstack/localstack'
```
alexrashed, line 111: A fork does not have the secret necessary to push to Tinybird...

alexrashed (changes requested):
Thanks for your contribution, but I don't think these (semi-automated?) changes are useful for us (see comments). For very few of the pipelines, maybe the GitHub token permission restrictions would make sense...

.github/workflows/tests-cli.yml
ceddy4395: Yes that was my reasoning for adding this change. This would prevent the job from failing is it will not run at all.

alexrashed (approved):
Thanks for your contribution, and acting on the comments! The change not limits the scope of the token, but also increases the developer experience for other contributors! 👍🏽

PR #10877, opened 2024-05-23
[SFN] Add Generic Resolver for Unsupported Optimised Service Integrations

MEPalma:
## Motivation
Currently the SFN v2 interpreter lacks support for a subset of the (stepfunctions optimised service integrations)[https://docs.aws.amazon.com/step-functions/latest/dg/connect-supported-services.html]. This results in state machine that define unsupported service integrations to fail on creation. However, this prevents the execution of state machines that may never execute such logic, or prevents the execution in circumstances in which a simple forward to the target service with support for `.waitForTaskToken` may be required. This PR addresses such behaviour by logging the incident and attempting to relay the call to the target service, and also expose support for `.waitForTaskToken` workflows.

## Changes
- Added generic resolver for unsupported service integrations.

### DIFF
--- a/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_unsupported.py
+++ b/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_unsupported.py
@@ -34,12 +34,12 @@ class StateTaskServiceUnsupported(StateTaskServiceCallback):
         # and relays the call to the target service with the computed parameters.
         self._log_unsupported_warning()
         service_name = self._get_boto_service_name()
-        api_action = self._get_boto_service_action()
-        sqs_client = boto_client_for(
+        boto_action = self._get_boto_service_action()
+        boto_client = boto_client_for(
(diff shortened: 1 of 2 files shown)


localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_unsupported.py
```diff
@@ -0,0 +1,45 @@
...
+        env: Environment,
+        resource_runtime_part: ResourceRuntimePart,
+        normalised_parameters: dict,
+    ):
+        # Logs that the evaluation of this optimised service integration is not supported
+        # and relays the call to the target service with the computed parameters.
+        self._log_unsupported_warning()
+        service_name = self._get_boto_service_name()
+        api_action = self._get_boto_service_action()
+        sqs_client = boto_client_for(
```
joe4dev, line 38: nit: copy/paste sqs_client

joe4dev (approved):
Nice fallback option 👍

PR #10878, opened 2024-05-23
Formalize Testing Rules

silv-io:
## Motivation

We recurrently experience stability issues in our CI pipeline. Many of the issues we encounter are avoidable and similar to each other. However, for those issues to be avoided we need common rules to adhere to.

## Changes

This PR slightly restructures the testing part of the contribution documentation and adds on a ruleset for stable testing pipelines. The ruleset is inspired by recent issues with pipeline stability and how those issues were resolved.

Further input is greatly appreciated! :)


docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Rules for stable tests
+
+Through experience, we encountered some guiding principles and rules when it comes to testing LocalStack.
+These aim to ensure a stable pipeline and keep flakes minimal. Any newly added test and feature should keep these in mind!
+
+| **ID** | **Rule** |
+|-------|-----------|
+| R01   | Inform code owners and/or test authors when a test is marked as flaky so that it can be fixed ASAP. |
+| R02   | Do not assume external dependencies (AWS resources, files, packages, images, licenses) are indefinitely available. |
+| R02a  | Where possible, add automatic creation of dependencies if they no longer exist (e.g., S3 buckets, roles). |
```
alexrashed, line 17: R02 are common issues with our test pipeline, but we don't really have a clear solution yet.
And I am not really sure what R02a means exactly to be honest...

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Rules for stable tests
+
+Through experience, we encountered some guiding principles and rules when it comes to testing LocalStack.
+These aim to ensure a stable pipeline and keep flakes minimal. Any newly added test and feature should keep these in mind!
+
+| **ID** | **Rule** |
+|-------|-----------|
+| R01   | Inform code owners and/or test authors when a test is marked as flaky so that it can be fixed ASAP. |
+| R02   | Do not assume external dependencies (AWS resources, files, packages, images, licenses) are indefinitely available. |
```
thrau, line 16: this one is not clear to me. could you elaborate/rephrase? why would i assume indefinitely available resources in a test? is there an example?

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+- [Terraform Tests](terraform-tests/README.md)
+
+## Rules for stable tests
+
+Through experience, we encountered some guiding principles and rules when it comes to testing LocalStack.
+These aim to ensure a stable pipeline and keep flakes minimal. Any newly added test and feature should keep these in mind!
+
+| **ID** | **Rule** |
+|-------|-----------|
+| R01   | Inform code owners and/or test authors when a test is marked as flaky so that it can be fixed ASAP. |
```
alexrashed, line 15: Maybe we should even directly file a PR marking it as skip? This would automatically inform code owners that a test is flaky 😛

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+| R02   | Do not assume external dependencies (AWS resources, files, packages, images, licenses) are indefinitely available. |
+| R02a  | Where possible, add automatic creation of dependencies if they no longer exist (e.g., S3 buckets, roles). |
+| R02b  | If on-demand creation is not possible, make access failures clearly visible for further investigation. |
```
thrau, line 18: does this just refer to error reporting?

alexrashed (commented):
This is a great spike for clearer and cleaner testing guidelines and rules for both authors and reviewers! 🚀 
I added a few comments, but I think this should really be a PR where we have an initial spike used to kickstart a discussion, which then converges to the best possible outcome.
Which is why I think this PR should also see some attention from others (especially @dominikschubert, @HarshCasper, @thrau) in order to converge to a state where we have a clear set of guidelines which _help_ both authors of tests and reviewers of PRs containing these tests. Any input is highly appreciated!

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+| R02c  | Add mechanisms to avoid access failures caused by rate limiting. |
+| R03   | Instead of waiting for a set amount of time, wait only as long as needed (with a reasonable cap). |
```
thrau, line 20: perhaps we should rephrase this to not wait a _set_ amount of time in general, but rather use event-based notification mechanisms (like using SQS queues to wait for lambda results) where possible, and otherwise use polling with reasonable timeouts?
perhaps also adding that this is specifically for asynchronous operations?

thrau (commented):
I think this is a very nice idea, and a great first iteration! I like the use of examples, and most of the rules are clear. I added comments where I think clarification is needed.

:100:

docs/testing/integration-tests/README.md
```diff
@@ -4,7 +4,7 @@ LocalStack has an extensive set of [integration tests](https://github.com/locals
 
 ## Writing integration tests
 
-The following guiding principles apply to writing integration tests:
+The following guiding principles apply to writing integration tests in addition to the [general rules](../README.md):
```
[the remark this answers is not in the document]
silv-io: I'm not sure I understand. Do you mean the "general rules" contain things that are more akin to "guiding principles" or the other way around, or both ways?

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+- [Parity Testing](parity-testing/README.md)
+- [Multi-account and Multi-region Testing](multi-account-region-testing/README.md)
+- [Terraform Tests](terraform-tests/README.md)
+
+## Rules for stable tests
+
+Through experience, we encountered some guiding principles and rules when it comes to testing LocalStack.
+These aim to ensure a stable pipeline and keep flakes minimal. Any newly added test and feature should keep these in mind!
+
+| **ID** | **Rule** |
```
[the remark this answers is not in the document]
silv-io: I'm not sure every rule would fit into this scheme, but for those which do we can add this for sure. Makes them easier to be memorized as well

docs/testing/README.md
silv-io: In a failing test involving pods, it failed because some S3 bucket that got accessed was deleted. I'm not sure if it was a write or a read access.
But generally, if there is some test where something should be written to a bucket `x`, you can add a check beforehand if `x` exists, and if it doesn't just create it before using it.

silv-io: Also, I think manifesting mindfulness of these issues in the rules will lead to development of best practices for these problems in the future

docs/testing/README.md
silv-io: That is what I was thinking about concerning the "Inform" part. I will make this more explicit!

docs/testing/README.md
silv-io: We have tests which access remote resources. Some access our own resources, like tests surrounding pods, and some access some files (like libraries, container images) stored on servers which are not under our control.

docs/testing/README.md
silv-io: Yes, access errors should be clearly marked as such, and as Alex mentioned should lead to fast failures.
This helps avoiding needless debugging of our own code when the issue lies outside of it.

In the cases where this was relevant in the past, we actually did well here. But making it a general rule will help us keep it like this in the future. If this is too trivial, we can also remove it again, though :)

docs/testing/README.md
silv-io: This is what I was trying to say with that but couldn't properly express it concisely enough

docs/testing/README.md
dominikschubert: Still a bit confused about the wording around `access` here. Since we're for the most part in the context of AWS this feels like you're talking about IAM permissions. Maybe rewording this might clear things up? E.g. retrieving/using vs. accessing?

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+|-------|-----------|
+| R01   | Inform code owners and/or test authors when a test is marked as flaky so that it can be fixed ASAP. |
+| R02   | Do not assume external dependencies (AWS resources, files, packages, images, licenses) are indefinitely available. |
+| R02a  | Where possible, add automatic creation of dependencies if they no longer exist (e.g., S3 buckets, roles). |
+| R02b  | If on-demand creation is not possible, make access failures clearly visible for further investigation. |
+| R02c  | Add mechanisms to avoid access failures caused by rate limiting. |
```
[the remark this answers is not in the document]
dominikschubert: So I assume this are things like: authenticated API access, retry logic, ... ? Anything else that would fit under here?

docs/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+| R03   | Instead of waiting for a set amount of time, wait only as long as needed (with a reasonable cap). |
+| R04   | For tests with multiple steps, handle waits separately and start each wait in the correct state. |
+| R05   | Ensure features interacting with account numbers work with arbitrary account numbers and multiple accounts simultaneously. (see [here](multi-account-region-testing/README.md) for further documentation) |
+| R06   | Ensure deterministic responses for anything that reaches an assertion or a snapshot match (e.g., by using proper transformations). (see [here](parity-testing/README.md) for further documentation) |
+| R07   | When a feature or its tests require a Python dependency, track updates to the dependency for potential stability impacts. |
+| R08   | Ensure all dependencies are available and functional on both AMD64 and ARM64 architectures. If a dependency is exclusive to one architecture, mark the corresponding test accordingly. |
```
dominikschubert, line 25: Maybe generally adding the list of test markers + explanations here in the README might be quite nice as well!

alexrashed (commented):
This is already looking really good! 🚀 
I think we can move forward after this round of reviews is processed and merge the PR 🥳

docs/testing/README.md
```diff
@@ -0,0 +1,428 @@
...
+snapshot = {"key": "<key:1>"}  # representing snapshot as a dict for presentation purposes
+
+
+def test_something(snapshot):
+    snapshot.add_transformer(...)  # add appropriate transformers
+    key = f"key-{short_uid()}"
+    snapshot.match(snapshot, {"key": key})
+```
+
+[documentation](parity-testing/README.md)
```
alexrashed, line 292: This seems to be a leftover.

alexrashed (approved):
Awesome! The cross-linking and the additional description is super helpful! Thanks for distilling these rules / guidelines!
As soon as the leftover (see comment) is removed, I think this should be merged! 🥳 
Let's make sure that we come back to these rules, explicitly reference them in reviews, and most importantly evolve them over time! 🚀

PR #10879, opened 2024-05-23
Resolved issue where using CDK wouldn't generate SecretString value

Morijarti:
## Motivation
Client has reported issue where using CDK would generate SecretString value on AWS but not on LS

## Changes
In CFn if you omit both GenerateSecretString and SecretString, you create an empty secret.

CDK wil generate empty dictionary, which means that CFn will receive value for GenrateSecretString (empty dictionary). 
In which case we also need to generate SecretString.

I've modified CFn provider for Secret, to also work with empty dictionary and added tests.

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_secretsmanager.py
+++ b/tests/aws/services/cloudformation/resources/test_secretsmanager.py
@@ -1,6 +1,7 @@
 import json
 import re
 
+import aws_cdk as cdk
 import pytest
 
 from localstack.testing.pytest import markers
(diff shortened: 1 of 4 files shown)


simonrw (changes requested):
When I run this test locally, I get an intermittent test failure. I think it might have to do with the contents of the secret that's generated, and the snapshot library, e.g. a secret value of


```
xbu^_A>@g,+O%lrN3^p$>\\*w<*1n}yG2
```
fails

_edit: I've just done a test and it fails 52% of the time, which is not great. Attaching error log_

[results.log](https://github.com/localstack/localstack/files/15419025/results.log)

tests/aws/services/cloudformation/resources/test_secretsmanager.py
```diff
@@ -149,3 +150,32 @@ def test_cfn_secret_policy(deploy_cfn_template, block_public_policy, aws_client,
...
+
+    with infra.provisioner() as prov:
+        outputs = prov.get_stack_outputs(stack_name=stack_name)
+
+        secret_name = outputs["SecretName"]
+        secret_arn = outputs["SecretARN"]
+
+        response = aws_client.secretsmanager.get_secret_value(SecretId=secret_name)
+
+        snapshot.add_transformer(snapshot.transform.key_value("SecretString"))
```
Morijarti, line 177: As @simonrw  noticed, it looks like `key_value` transformer fails with certain characters in value of key
I also managed to reproduce it when value of "SecretString" is  `zCB=.WGwc?!'\\<P5w2xI@\\;kz$'-3-zb`

bentsku: Would setting `reference_replacement=False` help here? 
`snapshot.add_transformer(snapshot.transform.key_value("SecretString", reference_replacement=False))`

Haven't tried it, just floating the idea

Morijarti: YES!
I've run test 10 times and it passed everytime.
Thank you @bentsku

bentsku: I believe we do regex matching on the whole string when replacing references, and there must be characters in the string that escapes the regex. With no replacement, no issue 😄 glad I could help

macnev2013 (approved):
Thanks for the PR. 🚀 LGTM.

PR #10880, opened 2024-05-23
Add support for the --env-file flag in docker flags

dfangl:
## Motivation
We currently do not support `--env-file` specification in the `DOCKER_FLAGS` variables.
It has, however, been requested by a customer to support this flag.

Please note that in the case of `--env-file`, the path to the env file has to be inside the LocalStack docker container.
This means, that the file has to be copied or mounted into the container before its usage, in essence before an affected container is spawned.

I avoided the usage of `docker-dotenv` (even though already installed) and wrote a simple manual parser of the environment files here, since the docker environment file parsing is quite limited. It for example does not respect comments at the end of a line (only when the line starts with a #) and also does not support referencing existing variables.
The usage of `docker-dotenv` would not allow us to be in parity with this restrictions of docker, even though it would be simpler and more powerful.

As context, we need to parse all the docker flags manually, as we do not simply forward them to a command line client, but use the docker sdk, which only accepts already parsed arguments and no command line flags.

## Changes
* `DOCKER_FLAGS` variables, like `LAMBDA_DOCKER_FLAGS` now support the docker `--env-file` parameter with all its limitations.

### DIFF
--- a/localstack-core/localstack/utils/container_utils/container_client.py
+++ b/localstack-core/localstack/utils/container_utils/container_client.py
@@ -1142,6 +1142,12 @@ class Util:
             "--env", "-e", help="Set environment variables", dest="envs", action="append"
         )
         parser.add_argument(
+            "--env-file",
+            help="Set environment variables via a file",
+            dest="env_files",
+            action="append",
+        )
+        parser.add_argument(
             "--label", "-l", help="Add container meta data", dest="labels", action="append"
         )
         parser.add_argument("--network", help="Connect a container to a network")
@@ -1182,6 +1188,20 @@ class Util:
                 extra_hosts = extra_hosts if extra_hosts is not None else {}
                 hosts_split = add_host.split(":")
                 extra_hosts[hosts_split[0]] = hosts_split[1]
+
+        # set env file values before env values, as the latter override the earlier
+        if args.env_files:
+            env_vars = env_vars if env_vars is not None else {}
+            for env_file in args.env_files:
+                with open(env_file, mode="rt") as f:
+                    env_file_lines = f.readlines()
+                for idx, line in enumerate(env_file_lines):
+                    line = line.strip()
+                    if not line or line.startswith("#") or "=" not in line:
+                        # skip comments or empty lines
+                        continue
+                    lhs, _, rhs = line.partition("=")
+                    env_vars[lhs] = rhs
 
         if args.envs:
             env_vars = env_vars if env_vars is not None else {}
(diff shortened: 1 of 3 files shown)


tests/unit/test_dockerclient.py
```diff
@@ -237,6 +238,38 @@ def test_random_ports(self):
         assert ports.to_str() == "-p 0:80"
         assert ports.to_dict() == {"80/tcp": None}
 
+    def test_env_files(self, tmp_path):
```
dominikschubert, line 241: nice thoughtful test!

tests/integration/docker_utils/test_docker.py
```diff
@@ -1444,6 +1445,49 @@ def test_run_with_ulimit(self, docker_client: ContainerClient):
...
+            command=["env"],
+            additional_flags=f"-e {env_variable} --env-file {env_file}",
+        )
+        env_output = stdout.decode(config.DEFAULT_ENCODING)
+        # behavior differs here from more advanced env file parsers
+        assert env_variable in env_output
+        assert "TEST1=VAL1" in env_output
+        assert "TEST2=VAL2" in env_output
+        assert "TEST3=${TEST2}" in env_output
+        assert "TEST4=VAL # end comment" in env_output
```
dominikschubert, line 1474: I assume those validations specifically would be working properly with other parsers right?

localstack-core/localstack/utils/container_utils/container_client.py
```diff
@@ -1183,6 +1189,20 @@ def parse_additional_flags(
                 hosts_split = add_host.split(":")
                 extra_hosts[hosts_split[0]] = hosts_split[1]
 
+        # set env file values before env values, as the latter override the earlier
+        if args.env_files:
+            env_vars = env_vars if env_vars is not None else {}
+            for env_file in args.env_files:
+                with open(env_file, mode="rt") as f:
+                    env_file_lines = f.readlines()
```
dominikschubert, line 1197: Adding a guard here to check if the file actually exists and logging a warning with some context/instructions (e.g. that the file needs to be in the container) would be helpful here!

dominikschubert (approved):
LGTM with some minor nits! Would be good to extend the docs with this new functionality as well :pray:

tests/integration/docker_utils/test_docker.py
dfangl: With a proper env file parser, the end of line comment would be stripped, and the variable would be substituted. However, docker does not do it, so neither should we I guess 🤷

dominikschubert: Love those parity situations where we need to lower our feature set so we aren't too powerful :grin:

localstack-core/localstack/utils/container_utils/container_client.py
dfangl: Added a log error for FileNotFoundError and OSError with some hints. I still reraised the exception, I think the container creation should probably fail if the `DOCKER_FLAGS` are incorrect.

Issue #10881, opened 2024-05-23
bug: S3 – CopyObject hangs on bucket with VersioningConfiguration explicitly set to `Status: Suspended`


eliellis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

LocalStack hangs when copying an S3 object to itself when the `VersioningConfiguration` is explicitly set to `Status=Suspended`, but *does not* hang when performing the same action if `VersioningConfiguration` is never specified

### Expected Behavior

LocalStack does not hang, and the `CopyObject` request either fails or succeeds

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

`docker run --rm -p 4566:4566 -v /var/run/docker.sock:/var/run/docker.sock localstack/localstack:latest`

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

[See here for a tiny repro](https://github.com/eliellis/localstack-s3-copy-object-error-repro)

### Environment

```markdown
- OS:
  ProductName: macOS
  ProductVersion: 14.4.1
  BuildVersion: 23E224
- LocalStack:
  LocalStack version: 3.4.1.dev
  LocalStack Docker container name: elegant_wing
  LocalStack Docker container id: 3f64fd392a69
  LocalStack Docker image sha: sha256:88710686251bcec4dfe66648e55ec176ec2c1025b7886aad70a7fa92ca67c111
  LocalStack build date: 2024-05-23
  LocalStack build git hash: bd1a23ada
```

bentsku:
Thanks a lot @eliellis for the report! I'll look into this right away 👀

bentsku:
Super interesting bug 😂 going from `Status=Enabled` to `Status=Suspended` works, but going straight to nothing to `Suspended` doesn't. I'm sure some other scenarios were failing. It's not the first time I'm getting bitten by this small optimization I've made for in-place copies. I'm already working on a fix in the linked PR, but thanks a lot for this report and the small sample! 🚀

eliellis:
> Super interesting bug 😂 going from `Status=Enabled` to `Status=Suspended` works, but going straight to nothing to `Suspended` doesn't. I'm sure some other scenarios were failing. It's not the first time I'm getting bitten by this small optimization I've made for in-place copies. I'm already working on a fix in the linked PR (#10882), but thanks a lot for this report and the small sample! 🚀

[...]

PR #10882, opened 2024-05-23
fix S3 CopyObject in place deadlock for Suspended buckets

bentsku:
## Motivation
As reported in #10881, we hit a deadlock in S3 when copying in place from a never versioned bucket into a `Suspended` bucket. This was due to a missing check in the copy in-place detection. We only ever checked if the object did not have a `version_id`, but `version_id = None` give the same result as `version_id = "null"`. 

This part of the code was then responsible for the deadlock:
```python
if should_copy_in_place(src_bucket, src_object, dest_bucket, dest_object):
    return self.open(dest_bucket, dest_object, mode="r")

with self.open(src_bucket, src_object, mode="r") as src_stored_object:
    dest_stored_object = self.open(dest_bucket, dest_object, mode="w")
    dest_stored_object.write(src_stored_object)
```

As the copy was not flagged as "in-place" when it should, we would not return early and try to open the same underlying object both in `write` and `read` mode 😅 
(We had an S3 object with `version_id = None` and the other was `version_id = "null"`, effectively being the same thing).

## Changes
- added more tests around copy in place behavior in AWS regarding `Suspended` buckets
- fixed the condition by adding a small helper method to verify that the object is non-versioned

_fixes #10881_

### DIFF
--- a/localstack/services/s3/v3/provider.py
+++ b/localstack/services/s3/v3/provider.py
@@ -1313,7 +1313,11 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         add_encryption_to_response(response, s3_object=s3_object)
 
-        if src_s3_bucket.versioning_status and src_s3_object.version_id:
+        if (
+            src_s3_bucket.versioning_status
+            and src_s3_object.version_id
+            and src_s3_object.version_id != "null"
+        ):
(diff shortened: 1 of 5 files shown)


steffyP (approved):
Great catch, thanks for fixing 🚀 🙌

Issue #10787, opened 2024-05-07
bug: Step Functions - `States.StringToJson` error when used in `ResultSelector` for S3 `getObject` action


bilki:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Whenever I use the `StringToJson` intrinsic function (https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html#asl-intrsc-func-json-manipulate) in the `ResultSelector` of a task calling the S3 `getObject` action (see steps to reproduce for the definition), the step function fails to execute that function with the messages:

```
2024-05-07T12:41:51.176 ERROR --- [ad-77 (eval)] l.s.s.a.c.eval_component   : Exception=AttributeError, Details=["'StreamingBody' object has no attribute 'strip'"] at '(StringToJson| {'name': (StatesFunctionName| {'name': 'tringToJso', 'function_type': <StatesFunctionNameType.StringToJson: 12>}, 'arg_list': (FunctionArgumentList| {'arg_list': [(FunctionArgumentJsonPath| {'_value': <botocore.response.StreamingBody object at 0xffff23890730>, '_json_path': '$.Body'}], 'size': 1}}'
2024-05-07T12:41:51.176 ERROR --- [ad-77 (eval)] l.s.s.a.c.eval_component   : Exception=AttributeError, Details=["'StreamingBody' object has no attribute 'strip'"] at '(PayloadBindingIntrinsicFunc| {'field': 'latest', 'src': 'States.StringToJson($.Body)', 'function': (StringToJson| {'name': (StatesFunctionName| {'name': 'tringToJso', 'function_type': <StatesFunctionNameType.StringToJson: 12>}, 'arg_list': (FunctionArgumentList| {'arg_list': [(FunctionArgumentJsonPath| {'_value': <botocore.response.StreamingBody object at 0xffff23890730>, '_json_path': '$.Body'}], 'size': 1}}}'
[...]

PR #10883, opened 2024-05-23
[SFN] Add Normalisation of Service Response Values

MEPalma:
## Motivation
Currently the StepFunctions interpreter does not perform normalisation steps to the output values of service responses. This is observable whenever the response type is not json inferable. For example responses of `s3` `GetObject` that include StreamingBody data types. This PR addresses such issue by adding a normalisation step for the response values, and uses this pattern to normalise StreamingBody data types. Addresses: #10787 

## Changes
- Add a response value normalisation step to the response normalisation workflow
- Add normalisation of StreamingBody data types
- Add related tests using `aws-sdk:s3:getObject`


MEPalma:
Thanks for the input Joel! I'll add this in the backlog

PR #10887, opened 2024-05-24
Fix deduplication scope for non-high-throughput FIFO messages

### LINKED ISSUE #10460: bug: ContentBasedDeduplication with DeduplicationScope in FIFO queue not working as expected

### Current Behavior

I created an SQS FIFO queue with ContentBasedDeduplication set to true and DeduplicationScope as messageGroup.

When i try to send 2 messages with same message body but with different message group id , only one message is present in the queue

The MessageId remains same for each send-message response

### Expected Behavior

with ContentBasedDeduplication set to true and DeduplicationScope as messageGroup, sending 2 messages with same message body but with different message group id should result in two messages being in the queue

The MessageId changes for each send-message response

With a docker-compose file

    docker-compose up -d localstack

1. First create SQS FIFO queue with the following command
```
awslocal sqs create-queue --region eu-central-1 --queue-name test-local-task-queue.fifo --attributes \
'{"FifoQueue":"true","ContentBasedDeduplication":"true","FifoThroughputLimit":"perQueue","DeduplicationScope":"messageGroup","VisibilityTimeout":"30","MessageRetentionPeriod":"345600"}'
```

2. Send a message to the queue with message group id G1

```
awslocal sqs send-message --queue-url http://localhost:4566/000000000000/test-local-task-queue.fifo --message-body A --message-group-id G1
```

4. Send a message to the queue with message group id G2

```
awslocal sqs send-message --queue-url http://localhost:4566/000000000000/test-local-task-queue.fifo --message-body A --message-group-id G2
```

5. Receive messages from the queue 

```
awslocal sqs receive-message --queue-url http://localhost:4566/queue/eu-central-1/000000000000/test-local-task-queue.fifo
```

### Environment

```markdown
- OS: MacOs
- LocalStack: latest
```

### Anything else?

When the same queue is created in AWS and if two messages with same body for different groups are sent, then you can see two messages in the queue 

image

baermat:
## Motivation
Apparently the deduplication scope for high-throughput queues is not restricted to said high-throughput queues, but can be used for any FIFO queue really. This PR addresses this issue.
Fixes #10460 

## Changes
- Change deduplication scope check to only consider said scope, and not if the queue is in HT mode
- Change the flag for tracking this to match the scope (I still haven't figured out why this cannot be changed on AWS after queue creation even though they specifically mention that you can do just that)

### DIFF
--- a/localstack/services/sqs/models.py
+++ b/localstack/services/sqs/models.py
@@ -934,7 +934,7 @@ class FifoQueue(SqsQueue):
     message_groups: dict[str, MessageGroup]
     inflight_groups: set[MessageGroup]
     message_group_queue: Queue
-    high_throughput: bool
+    deduplication_scope: str
 
     def __init__(self, name: str, region: str, account_id: str, attributes=None, tags=None) -> None:
         super().__init__(name, region, account_id, attributes, tags)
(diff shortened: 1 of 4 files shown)


thrau (approved):
very nice catch there! :ok_hand:  changes LGTM!

PR #10889, opened 2024-05-24
fix pre-signed POST str type for content-length-range

bentsku:
## Motivation
As a follow up from #10862, the same user encounter an issue when the policy condition were the following:
`["content-length-range", "0", "104857600000"]]`, as being generated by https://docs.fineuploader.com/endpoint_handlers/amazon-s3.html

This was a hole in our validation, we didn't check that the content length range could be a string. 

## Changes
- added test cases for content length range with strings (valid and invalid)
- fixed the logic to properly handle strings

### DIFF
--- a/localstack-core/localstack/services/s3/presigned_url.py
+++ b/localstack-core/localstack/services/s3/presigned_url.py
@@ -790,8 +790,9 @@ def validate_post_policy(
     form_dict = {k.lower(): v for k, v in request_form.items()}
     for condition in conditions:
         if not _verify_condition(condition, form_dict, additional_policy_metadata):
+            str_condition = str(condition).replace("'", '"')
             raise AccessDenied(
-                f"Invalid according to Policy: Policy Condition failed: {condition}",
+                f"Invalid according to Policy: Policy Condition failed: {str_condition}",
                 HostId=FAKE_HOST_ID,
             )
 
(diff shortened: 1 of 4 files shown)


localstack/services/s3/presigned_url.py
```diff
@@ -790,8 +790,9 @@ def validate_post_policy(
     form_dict = {k.lower(): v for k, v in request_form.items()}
     for condition in conditions:
         if not _verify_condition(condition, form_dict, additional_policy_metadata):
+            str_condition = str(condition).replace("'", '"')
```
steffyP, line 793: Q: do you have an example of when this would happen? I think it's not covered by the newly added test?
If it's already casted to str, why do we still need to replace the single quote? Wouldn't we end up with double quotes inside the string?

steffyP (approved):
Thanks for fixing 🚀

localstack/services/s3/presigned_url.py
bentsku: https://github.com/localstack/localstack/pull/10889/commits/bafa871166b7da2fd96088b7ed3494d53a8fa2de should be a bit clearer, sorry for not mentioning it in the PR description! I realized after that I left the util in the previous tests, now it's cleaner 😄

Issue #10893, opened 2024-05-24
bug: localstack/localstack:lastest fails to startup 


robo-corg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have been using localstack in a github workflow for CI and it seems to fail to come up now:

```
  services:
      localstack:
        image: localstack/localstack
        env:
          SERVICES: s3
          DEFAULT_REGION: us-east-1
          DATA_DIR: /tmp/localstack/data
        ports:
        - 4566:4566
 ```
 
 Seems to fail from connecting to postgresql 
 ```
 Service container localstack failed.
  /usr/bin/docker logs --details 8dc603486e76bc37f7762a235feb82e480926166c0b4cbebc904bca9e971b5ef
   
   LocalStack version: 3.4.1.dev
   LocalStack Docker container name: localstack-main
   Failed to inspect docker container: Docker not available Traceback (most recent call last):
     File "/opt/code/localstack/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 793, in urlopen
       response = self._make_request(
                  ^^^^^^^^^^^^^^^^^^^
     File "/opt/code/localstack/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 496, in _make_request
       conn.request(
     File "/opt/code/localstack/.venv/lib/python3.11/site-packages/urllib3/connection.py", line 400, in request
       self.endheaders()
     File "/usr/local/lib/python3.11/http/client.py", line 1298, in endheaders
       self._send_output(message_body, encode_chunked=encode_chunked)
     File "/usr/local/lib/python3.11/http/client.py", line 1058, in _send_output
       self.send(msg)
     File "/usr/local/lib/python3.11/http/client.py", line 996, in send
       self.connect()
     File "/opt/code/localstack/.venv/lib/python3.11/site-packages/docker/transport/unixconn.py", line 26, in connect
       sock.connect(self.unix_socket)
   FileNotFoundError: [Errno 2] No such file or directory
   ```

### Expected Behavior

localstack service should be ready and not timeout typically it is ready in <10s

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

Found this using localstack as a github action as described above but I suspect a simple docker run might trigger the bug

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

### Environment

```markdown
- OS:
- LocalStack:
[...]

PR #10894, opened 2024-05-24
fix healthcheck in dockerfile

thrau:
## Motivation

As a user pointed out in https://github.com/localstack/localstack/issues/10893#issuecomment-2130190699, the healthcheck command was faulty after #10800:

```
thomas@ninox /tmp/foo % docker exec -it foo-localstack-1 bash
root@cb3bded7f1f7:/opt/code/localstack# ./bin/localstack status services 
  File "/opt/code/localstack/./bin/localstack", line 23, in <module>
    main()
  File "/opt/code/localstack/./bin/localstack", line 17, in main
    from localstack.cli import main
ModuleNotFoundError: No module named 'localstack'
```

I'm deliberately not fixing `bin/localstack` because I believe we should not be using it directly in this way.

I also noticed that rolo is logging HTTP health check requests again, so I updated the default log levels to what we had in `localstack.aws.serving`.

## Changes

* `HEALTHCHECK` in `Dockerfile` now uses `localstack` from the `.venv` instead the plain script, which uses the correct pythonpath
* default log levels for rolo is now `WARNING` and `DEBUG` only for tracing

### DIFF
--- a/localstack-core/localstack/logging/setup.py
+++ b/localstack-core/localstack/logging/setup.py
@@ -22,6 +22,7 @@ default_log_levels = {
     "s3transfer": logging.INFO,
     "urllib3": logging.WARNING,
     "werkzeug": logging.WARNING,
+    "rolo": logging.WARNING,
     "localstack.aws.accounts": logging.INFO,
     "localstack.aws.protocol.serializer": logging.INFO,
     "localstack.aws.serving.wsgi": logging.WARNING,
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
LGTM, tested locally as well to make sure it works. Was confused at first why that should work but looking at the contents of the script it makes sense :upside_down_face:

PR #10901, opened 2024-05-27
Remove CloudFormation moto backend from persisted state

dominikschubert:
## Motivation

For some reason the moto backend was probably unintentionally added last year to the persisted CFn state even though we don't actually use moto in the service.

Saw this by chance when trying to debug persistence. It shouldn't have any actual effect on persistence tests since we're not using moto in the service but still good to clean this up.

## Changes

- Only our internal CloudFormation store will be persisted now


simonrw (approved):
Thanks for cleaning this up.

PR #10902, opened 2024-05-27
Fix lambda persistence restore

dfangl:
## Motivation
With #10831 we introduced an function instance id, which should not be persisted.
However, on persistence restore we do not set the instance_id again, so the restored state does not contain the attribute, which leads to errors down the road.

We need to explicitly set the instance id after restore, `__post_init__` is not called by default on unpickling.

## Changes
* LocalStack Lambda persistence should be fixed, no more `AttributeError` thrown after persistence restore

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/lambda_models.py
+++ b/localstack-core/localstack/services/lambda_/invocation/lambda_models.py
@@ -620,6 +620,12 @@ class Function:
         del state["instance_id"]
         return state
 
+    def __setstate__(self, state):
+        # Inject persistent state
+        self.__dict__.update(state)
+        # Create new instance id
+        self.__post_init__()
+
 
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
LGTM but let's leave the final :heavy_check_mark:  to @joe4dev since he authored the original instance id PR.

joe4dev (approved):
Thank you for the hotfix related to https://github.com/localstack/localstack/pull/10831 @dfangl  👏👏👏

I wasn't aware that post_init doesn't apply upon state restore 🧠 
It is also quite unfortunate that we only discover such issues in the CI after the image has been released 😢 

As discussed, we are [both not happy](https://github.com/localstack/localstack/pull/10831#discussion_r1612069382) with this hack, but we didn't find a better solution for now.

PR #10276, opened 2024-02-19
[wip] Add util script for building/pushing Docker images

whummer:
## Motivation

A common use case we're seeing is to build multi-arch Docker images with semantic versioning (build tags like `:3`, `3.1`, `latest`, etc), for which we currently have targets in the Makefile (`docker-build`, `docker-push-master`, `docker-create-push-manifests`, etc). Maintaining this non-trivial logic in the Makefile is cumbersome, and we could also benefit from making it reusable across different repos. This PR is a first step towards making these targets reusable across different repos, including -ext, the upcoming Snowflake image, etc.

## Changes

Add util script for building/pushing Docker images. The script follows a similar structure as the existing `bin/release-helper.sh` script we already have in place.

🚧 more details following soon.. /cc @alexrashed 

TODOs:
- [x] add logic to extract correct major/minor/patch version numbers from `__version__` in `__init__.py` files

## Testing

The script has been tested in an upstream pipeline (snowflake repo). Still need to coordinate whether and how to integrate the script into the build process of this repo.

PR #10903, opened 2024-05-27
fix docker helper usage typo and push without force flag

alexrashed:
## Motivation
#10872 migrated the make targets to push the Docker images to a separate helper script.
Unfortunately a few small issues sneaked into the script and its usage, which are fixed with this PR.

## Changes
- Fixes the usage of the docker-helper in the CircleCI workflow (small typo, `push-manifests` instead of `push-manifest`)
- Fix the push targets in case `FORCE_VERSION_TAG_PUSH` is not set.
- Fix the detection of a version bump.
  - There was still the old version detection from #10276, but this has changed by @silv-io with #10497 in the meantime.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -651,9 +651,9 @@ jobs:
           name: Create multi-platform manifests
           command: |
             # Push to Docker Hub
-            ./bin/docker-helper.sh push-manifest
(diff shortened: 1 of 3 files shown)


thrau (approved):
LGTM!

PR #10904, opened 2024-05-27
Partitions: ARN's now contain correct partition

bblommers:
## Motivation
Users should be able to use LocalStack in any region. That includes the regular AWS regions, but also regions in the China, GovCloud and ISO-partitions. 

One of the major implications of  using a region in a non-standard partition is the fact that ARN's should reflect the correct partitions. For example:
 - `"arn:aws:apigateway:us-east-1:s3:path/"` for the regular `aws` partition
 - `"arn:aws-cn:apigateway:cn-north-1:s3:path/"` for the China partition
 - etc.

The current implementation uses a regex to rewrite any mention of `arn:aws` to `arn:aws-cn` in the raw response. This is a fragile hack, breaks frequently, doesn't consider inter-service comunication scenarios, and has a host of other issues.

## Changes

This PR changes all ARN's to include the correct partition on creation.

## Tests

Tests were added to verify the following:
 - ARN's generated by LocalStack have the correct partition
 - ARN's generated by Moto have the correct partition
 - ARN's are correctly validated, in case of inter-service communication

## TODO

What's left to do before merging:

- [x] Wait for the latest Moto-bump, as that includes native partition-support in Moto
- [x] Rebase to include the multi-distribution refactoring

What's left to to  after merging:
- [ ] Make the same changes in LocalstackPro
- [ ]  Remove the current ArnPartitionRewriter

### DIFF
--- a/tests/aws/services/s3/test_s3_partitions.py
+++ b/tests/aws/services/s3/test_s3_partitions.py
@@ -1,0 +1,38 @@
+import pytest
+
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+
+
+class TestS3Partitions:
+    # We only have access to the AWS partition, not CHINA/US-GOV/etc
+    @markers.aws.manual_setup_required
+    @pytest.mark.parametrize("region,partition", [("us-east-2", "aws"), ("cn-north-1", "aws-cn")])
+    def test_sns_notifications_in_different_partitions(
+        self, account_id, aws_client_factory, region, partition
+    ):
+        s3 = aws_client_factory(region_name=region).s3
+        sns = aws_client_factory(region_name=region).sns
+
+        topic_name = f"topic-{short_uid()}"
+        topic_arn = sns.create_topic(Name=topic_name)["TopicArn"]
+
+        bucket_name = f"test-bucket-{short_uid()}"
+        s3.create_bucket(
+            Bucket=bucket_name, CreateBucketConfiguration={"LocationConstraint": region}
+        )
+
+        s3.put_bucket_notification_configuration(
+            Bucket=bucket_name,
+            NotificationConfiguration={
+                "TopicConfigurations": [
+                    {
+                        "TopicArn": topic_arn,
+                        "Events": ["s3:ObjectCreated:*"],
+                    }
+                ]
+            },
+        )
+
+        resp = s3.get_bucket_notification_configuration(Bucket=bucket_name)
+        assert resp["TopicConfigurations"][0]["TopicArn"] == topic_arn
--- a/tests/aws/services/s3/test_s3_partitions.py
+++ b/tests/aws/services/s3/test_s3_partitions.py
@@ -2,8 +2,10 @@ import pytest
 
 from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
+from tests.aws.services.s3.conftest import TEST_S3_IMAGE
 
 
+@pytest.mark.skipif(condition=TEST_S3_IMAGE, reason="SNS not enabled in S3 image")
 class TestS3Partitions:
     # We only have access to the AWS partition, not CHINA/US-GOV/etc
     @markers.aws.manual_setup_required
--- a/tests/aws/services/s3/test_s3_partitions.py
+++ b/tests/aws/services/s3/test_s3_partitions.py
@@ -1,11 +1,16 @@
 import pytest
 
+from localstack.config import LEGACY_V2_S3_PROVIDER
 from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
 from tests.aws.services.s3.conftest import TEST_S3_IMAGE
 
 
 @pytest.mark.skipif(condition=TEST_S3_IMAGE, reason="SNS not enabled in S3 image")
+@pytest.mark.skipif(
+    condition=LEGACY_V2_S3_PROVIDER,
+    reason="Will be fixed in a future Moto release (https://github.com/getmoto/moto/pull/7731)",
+)
 class TestS3Partitions:
     # We only have access to the AWS partition, not CHINA/US-GOV/etc
     @markers.aws.manual_setup_required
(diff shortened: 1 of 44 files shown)


localstack/utils/aws/arns.py
```diff
@@ -88,24 +89,19 @@ def _resource_arn(name: str, pattern: str, account_id: str, region_name: str) ->
...
+def iam_role_arn(role_name: str, account_id: str, region_name: str) -> str:
     if not role_name:
         return role_name
-    if role_name.startswith("arn:aws:iam::"):
+    if re.match(f"{ARN_PARTITION_REGEX}:iam::", role_name):
         return role_name
-    return "arn:aws:iam::%s:role/%s" % (account_id, role_name)
-
-
-def iam_policy_arn(policy_name: str, account_id: str) -> str:
```
bblommers, line 99: This method wasn't used anywhere, as far as I could tell, so I just removed it

localstack/utils/aws/arns.py
```diff
@@ -443,20 +448,6 @@ def route53_resolver_query_log_config_arn(id: str, account_id: str, region_name:
 #
 
 
-def fix_arn(arn: str):
```
bblommers, line 446: This method wasn't used anywhere either

tests/aws/test_multi_accounts.py
```diff
@@ -1,25 +1,10 @@
-import pytest
-
 from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
 
 
-@pytest.fixture
```
bblommers, line 7: This was moved to `tests/conftest.py`, so that I could reuse it for all the new multi-region tests.

localstack/utils/aws/arns.py
```diff
@@ -88,24 +89,19 @@ def _resource_arn(name: str, pattern: str, account_id: str, region_name: str) ->
...
-    return "arn:aws:iam::{}:policy/{}".format(account_id, policy_name)
+    return "arn:%s:iam::%s:role/%s" % (get_partition(region_name), account_id, role_name)
 
 
 def iam_resource_arn(resource: str, account_id: str, role: str = None) -> str:
     if not role:
         role = f"role-{resource}"
-    return iam_role_arn(role_name=role, account_id=account_id)
+    # Only used in tests, so we can hardcode the region for now
+    return iam_role_arn(role_name=role, account_id=account_id, region_name="us-east-1")
```
dfangl, line 104: That might become an issue when we do multi-partition tests! I can fix it as part of the test work though.

localstack/utils/aws/aws_stack.py
```diff
@@ -20,6 +20,29 @@
...
+
+def get_partition(region: Optional[str]) -> str:
+    if not region:
+        return DEFAULT_PARTITION
+    if region in PARTITION_NAMES:
+        return region
+    for prefix in REGION_PREFIX_TO_PARTITION:
+        if region.startswith(prefix):
+            return REGION_PREFIX_TO_PARTITION[prefix]
+    return DEFAULT_PARTITION
```
dfangl, line 43: `aws_stack.py` is currently mostly legacy code, we try to replace everything in it step by step. I would like to avoid putting new code in here, but rather move it to either `arns.py` or perhaps some new module in the `util` package, `partition.py` for example.

tests/aws/services/sns/test_sns_regions.py
```diff
@@ -0,0 +1,100 @@
...
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+from localstack.utils.sync import retry
+
+
+class TestSnsPartitions:
+    # We only have access to the AWS partition, not CHINA/US-GOV/etc
+    @markers.aws.manual_setup_required
+    @pytest.mark.parametrize("region,partition", [("us-east-2", "aws"), ("cn-north-1", "aws-cn")])
+    def test_topic_in_different_partitions(self, account_id, client_factory, region, partition):
```
dfangl, line 14: I think the `client_factory` was renamed `regional_client_factory`? In any case, see my other comment, it is generally better to use `aws_client_factory` directly, as it is callable with a similar interface, but provides typing. The only reason to go the other way is if the client should be dynamically determined by a string with the service name.

tests/aws/test_multi_accounts.py
dfangl: In general, you can also use the fixture `aws_client_factory` directly, like this:
```python
aws_client_factory(region_name=region_name, aws_access_key_id=access_key_id).sqs
```
This properly preserves the typing. I know the multi account tests do it differently, but generally the direct usage way would be preferred.

dfangl (commented):
I think this is already really great, only some suggestions regarding the test fixtures used!

tests/aws/test_multi_accounts.py
bblommers: Done - all tests now use `aws_client_factory`. Thanks for the hint.

I've reverted the change to the multi account tests as well.

localstack/utils/aws/aws_stack.py
bblommers: Done - all these methods are now in `arns.py`

dfangl (approved):
This is good to go from my side, the changes look good! Could you rebase the PR to master, so we can see if everything will be green?

dfangl (changes requested):
Sorry, the approval was a tad premature, I just noticed the tests are still red, and it seems related to your changes. Some methods seem to be missing the adaption to the new arn utility parameter, and a rebase is still missing. Will do my final review once this is resolved :)

tests/aws/services/s3/test_s3_partitions.py
```diff
@@ -0,0 +1,38 @@
...
+
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+
+
+class TestS3Partitions:
+    # We only have access to the AWS partition, not CHINA/US-GOV/etc
+    @markers.aws.manual_setup_required
+    @pytest.mark.parametrize("region,partition", [("us-east-2", "aws"), ("cn-north-1", "aws-cn")])
+    def test_sns_notifications_in_different_partitions(
```
dfangl, line 11: We should skip this test in the s3 image run, as sns will not be available, like other notifications test, using this skip marker: `@pytest.mark.skipif(condition=TEST_S3_IMAGE, reason="SQS not enabled in S3 image")`

dfangl:
Superceded/changes included in #10912

PR #10912, opened 2024-05-28
Add rewriter for invalid regions in the handler chain

dfangl:
## Motivation
This PR includes changes to rewrite regions which are not in the default partition to `us-east-1`, to avoid inconsistent behavior when specifying regions in different partitions, as those are currently not supported here.

It also provides the initial work to generate the partition part of ARNs dynamically, and to alter our test cases to use a placeholder for the partitions, to be able to test non-default partitions as well, if they are supported.

## Changes

* Add region rewriter to rewrite invalid regions to `us-east-1`, instead of failing later on store creation.
* This PR changes all ARN's to include the correct partition for the region they are created with
* Rewrite snapshots to include partition placeholders
* Add snapshot transformers for the partition

# Important commits
* Initial support for different partitions in ARN generation: bfebe00ba3314571549aef1060fd06ba6076266e
* Snapshot + transformer changes: 83e73a7002b23cbb02902aa41bb02e78163360a9
    Snapshots were transformed using the following command:
    ```bash
    fd -e "snapshot.json" --exec sed -i 's/arn:aws:/arn:<partition>:/g' {}
    ```
* Rewriter for not supported regions: 326768c629ab13f8d2df7ef0f115855c46a6ac54

## Testing

* Try making a request with an invalid region to LocalStack
* A warning should appear, and the region should be rewritten to `us-east-1`.

## TODO

What's left to do before merging:

- [x] Wait for the latest Moto-bump, as that includes native partition-support in Moto
- [x] Rebase to include the multi-distribution refactoring

What's left to do after merging:
- [ ]  Remove the current ArnPartitionRewriter

## Superceded PRs

This PR supercedes, and includes, #10904 by @bblommers . To make merging and testing easier, we will merge this PR, with both our works combined.


bblommers (approved):

localstack-core/localstack/services/apigateway/integration.py
```diff
@@ -188,7 +188,7 @@ def get_internal_mocked_headers(
 
 
 def get_source_arn(invocation_context: ApiInvocationContext):
-    return f"arn:aws:execute-api:{invocation_context.region_name}:{invocation_context.account_id}:{invocation_context.api_id}/{invocation_context.stage}/{invocation_context.method}{invocation_context.path}"
+    return f"arn:{get_partition(invocation_context.region_name)}:execute-api:{invocation_context.region_name}:{invocation_context.account_id}:{invocation_context.api_id}/{invocation_context.stage}/{invocation_context.method}{invocation_context.path}"
```
bentsku, line 191: q: we have the same util in `next_gen`, is it alright if we do a follow PR to implement this change as well? in `localstack.services.apigateway.next_gen.execute_api.helpers.get_source_arn`. If not important we can change it in a follow PR for us too. 😄

dfangl: Very good catch! I would do it in an immediate follow up, I would like to get this one in as soon as possible to avoid merge conflicts (rebasing this one is a real pain), and fix not-critical issues in follow ups. If there is any other change necessary anyway, I will make it here!

localstack-core/localstack/aws/handlers/region.py
```diff
@@ -15,9 +21,83 @@ class RegionContextEnricher(Handler):
...
+    def available_regions(self) -> list[str]:
+        """
+        Returns a list of supported regions.
+        :return: List of regions in the `aws` partition.
+        """
+        # We cannot cache the session here, as it is not thread safe. As the entire method is cached, this should not
+        # have a significant impact.
+        # using s3 as "everywhere available" service, as it usually is supported in all regions
+        # the S3 image also deletes other botocore specifications, so it is the easiest possibility
+        return Session().get_available_regions("s3", "aws")
```
dominikschubert, line 89: nit: IAM or STS might be an even better candidate, but I don't think it will make any practical difference here

tests/aws/conftest.py
```diff
@@ -111,6 +112,9 @@ def snapshot(request, _snapshot_session: SnapshotSession, account_id, region_nam
 
     _snapshot_session.add_transformer(RegexTransformer(account_id, "1" * 12), priority=2)
     _snapshot_session.add_transformer(RegexTransformer(region_name, "<region>"), priority=2)
+    _snapshot_session.add_transformer(
+        RegexTransformer(f"arn:{get_partition(region_name)}:", "arn:<partition>:"), priority=2
```
dominikschubert, line 116: nice :raised_hands:

dominikschubert (approved):
LGTM, Awesome work @bblommers  and @dfangl.

This should now really help with issues around mismatched/unexpected ARNs when used with non-standard regions! Great PR description, really made it a breeze to efficiently review the changes.

localstack-core/localstack/aws/handlers/region.py
dfangl: Yes, I initially had `ec2` here, but changed it for `s3`. The main issue is the `s3` image, where we delete all botocore specs except those of s3, so the regions will be empty for all other services. To avoid removing this deletion, I used s3, as it is also in all regions. (I hope this is clear from the comments)

alexrashed (approved):
Wow! 😲 What a massive churn over the whole codebase! Really great piece of work, @dfangl and @bblommers! 💯 🏅 
Nice implementation, and great and thorough approach in testing and churning over the codebase!

Issue #10898, opened 2024-05-26
S3 HeadObject does not return checksum for multi part uploaded objects


leslie-wang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

As solved in https://github.com/localstack/localstack/issues/8676 and https://github.com/localstack/localstack/issues/6659, checksum will be returned for regular put-object call, but not for the object put with multi-part methods. `head-object` will not return checksum even if passing `--checksum-mode Enabled` parameter

### Expected Behavior

checksume will be returned when head-object is called for multi-part uploaded objects

### How are you starting LocalStack?

With a `docker run` command

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS:Linux
- LocalStack:
  LocalStack version: 1.32.91
  LocalStack Docker image sha: ed0d88483d5bb02d754df8ed738b3d7c0be6be4aa3608b431c2af8d0895fb3f6
  LocalStack build date:
  LocalStack build git hash:
```

bentsku:
Hello @leslie-wang and thanks for your report!

This was a long standing missing feature of our S3 implementation. This should now be implemented and available as part of our `latest` image. Could you please pull `localstack/localstack:latest` and see if that works for you? Feel free if there's still anything. Thank you!

leslie-wang:
Thanks for quick fix! I'll let you know if I need any more help!

BTW, this is great project!

PR #10917, opened 2024-05-28
implement S3 checksum for Multipart uploads

bentsku:
## Motivation
As reported in #10898, multipart uploads still did not have checksum implemented. This was left as a TODO during the migration of S3, and I've now implemented it. 

See https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#large-object-checksums

## Changes
- enable and regenerate tests for this feature
- implement logic in the models for checksum logic, properly checking if the checksum provided are there, and if we're providing checksum or not depending on what was done during `CreateMultipartUpload`
- additional fixes around `PartsCount` for `HeadObject`

_fixes #10898_

### DIFF
--- a/localstack-core/localstack/services/s3/utils.py
+++ b/localstack-core/localstack/services/s3/utils.py
@@ -196,7 +196,7 @@ def get_s3_checksum(algorithm) -> ChecksumHash:
 
 
 class S3CRC32Checksum:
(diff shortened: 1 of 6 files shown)


steffyP (approved):
great work 🚀 there can't be a lot of unimplemented features left in s3 😅

PR #10919, opened 2024-05-29
Increase performance for first store creation per account/service

dfangl:
## Motivation
We currently create the boto3 session multiple times for checking available regions in the RegionBundle.
This code is called on the first store access per account for every service.

The changes in this PR should cache the result, and reduce the initial store instantiation time from around 200ms to around 50ms.
Due to the caching, the first request in additional accounts will be instantenous.

## Changes
* LocalStack should respond ~150ms (depending on the system) faster on the first store-accessing call for each service for a new account.


viren-nadkarni (approved):
Nice one :+1:

PR #10922, opened 2024-05-29
Move testing config from constants.py to its own module

dfangl:
## Motivation
We currently define a lot of test config variables in constants.py.

Since they are variable config options, they should not be in constants.py in the first place. In addition to that, the import order prevents config profiles to work for those variables, as they are initialized before the profiles are loaded (in config.py, as config.py depends on constants.py).

This PR introduces a new testing/config.py module, which contains testing specific configuration.

## Changes
* Move testing configuration into its own module
* Testing using config profiles is now more efficient, as they can be used to define the test config variables


dfangl:
Sorry for alerting everyone, I will remove the reviewers where only an import change took place (feel free to take a look anyway of course :))

localstack-core/localstack/testing/test_config.py
```diff
@@ -0,0 +1,18 @@
+import os
```
viren-nadkarni, line 1: Calling this `test_config` is a bit of a misnomer as it looks like a test file. 

Consider simply calling it `localstack.testing.config`

dfangl: Ah, did not notice that, good point! Will change it!

viren-nadkarni (approved):
Great stuff :100:

PR #9141, opened 2023-09-13
Triage aws_unknown markers across AWS test suite

dominikschubert:
## Motivation

To unblock another task (opening a GH issue for all aws_unknown/aws_needs_fixing markers, I need to reduce them first, otherwise we're running into issues with the maximum character limit for GH issues.

## Changes

- Validate a lot of `@markers.aws.unknown`-marked integration tests against AWS and assigned `@markers.aws.validated`/`@markers.aws.needs_fixing`/`@markers.aws.only_localstack` where applicable.

PR #10923, opened 2024-05-29
Validate remaining unknown sqs tests

baermat:
## Motivation
Support the ongoing initiative of #9141 and remove remaining unknown markers for sqs related tests.
Side note: the tests in test_notifications.py should probably be moved/deleted, they not seem special or complex enough to warrant its own test class

## Changes
- remove unknown markers and replace them with validated/needs_fixing
- fix all tests so that they at least work against AWS


thrau (approved):
thanks @baermat! have you determined the reasons the tests need fixing? what's failing exactly? would be great to have some context for the person who will fix the test :-)

baermat:
The reasons were mostly bad handling handling of API calls to AWS. Missing cleanups, wrong clients, no retries, etc.

PR #6005, opened 2022-05-03
Add filter for cognito identity fields

dfangl:
This PR adds a hacky filter, to the cognito_identity field of the lambda context. This will only allow correct values to be passed into it.

Once a real fix is available (by rethinking and redoing the `localstack.services.apigateway.apigateway_listener.invoke_rest_api_integration_backend` and `localstack.services.apigateway.helpers.get_event_request_context` methods @whummer @calvernaz ) so that this part:
```python
        "identity": {
            "accountId": account_id,
            "sourceIp": source_ip,
            "userAgent": headers.get("User-Agent"),
        },
```
does not show up in the cognito_identity part of the lambda context, this PR should be reverted.

Until then, it will only allow `"cognitoIdentityId"` and `"cognitoIdentityPoolId"` as directory keys for cognito_identity.

This should help fixing #5783 .

Test for apigateway - rust lambda integration added to prevent further regressions.

PR #10926, opened 2024-05-30
apigw validate test

cloutierMat:
## Motivation

validate apigw tests

### DIFF
--- a/tests/aws/services/apigateway/apigateway_fixtures.py
+++ b/tests/aws/services/apigateway/apigateway_fixtures.py
@@ -198,11 +198,17 @@ class UrlType(Enum):
 
 
 def api_invoke_url(
-    api_id: str, stage: str = "", path: str = "/", url_type: UrlType = UrlType.HOST_BASED
+    api_id: str,
+    stage: str = "",
+    path: str = "/",
+    url_type: UrlType = UrlType.HOST_BASED,
+    region: str = "",
 ):
     if is_aws_cloud():
+        if not region:
+            region = aws_stack.get_boto3_region()
         stage = f"/{stage}" if stage else ""
-        return f"https://{api_id}.execute-api.{aws_stack.get_boto3_region()}.amazonaws.com{stage}{path}"
+        return f"https://{api_id}.execute-api.{region}.amazonaws.com{stage}{path}"
     if url_type == UrlType.HOST_BASED:
         return host_based_url(api_id, stage_name=stage, path=path)
     return path_based_url(api_id, stage_name=stage, path=path)
--- a/tests/aws/services/apigateway/test_apigateway_basic.py
+++ b/tests/aws/services/apigateway/test_apigateway_basic.py
@@ -68,7 +68,6 @@ from tests.aws.services.apigateway.conftest import (
 from tests.aws.services.lambda_.test_lambda import (
     TEST_LAMBDA_HTTP_RUST,
     TEST_LAMBDA_NODEJS,
-    TEST_LAMBDA_NODEJS_APIGW_502,
     TEST_LAMBDA_NODEJS_APIGW_INTEGRATION,
     TEST_LAMBDA_PYTHON,
     TEST_LAMBDA_PYTHON_ECHO,
@@ -657,36 +656,6 @@ class TestAPIGateway:
             aws_client.apigateway.get_authorizer(
                 restApiId=get_api_gateway_id, authorizerId=authorizer_id
             )
-
-    @markers.aws.unknown
-    def test_malformed_response_apigw_invocation(
-        self, create_lambda_function, aws_client, region_name
-    ):
-        lambda_name = f"test_lambda_{short_uid()}"
-        lambda_resource = "/api/v1/{proxy+}"
-        lambda_path = "/api/v1/hello/world"
-
-        lambda_uri = create_lambda_function(
-            func_name=lambda_name,
-            zip_file=testutil.create_zip_file(TEST_LAMBDA_NODEJS_APIGW_502, get_content=True),
-            runtime=Runtime.nodejs16_x,
-            handler="apigw_502.handler",
-        )["CreateFunctionResponse"]["FunctionArn"]
-
-        target_uri = f"arn:aws:apigateway:{region_name}:lambda:path/2015-03-31/functions/{lambda_uri}/invocations"
-        result = testutil.connect_api_gateway_to_http_with_lambda_proxy(
-            "test_gateway",
-            target_uri,
-            path=lambda_resource,
-            stage_name="testing",
-        )
-        api_id = result["id"]
-        url = path_based_url(api_id=api_id, stage_name="testing", path=lambda_path)
-        result = requests.get(url)
-
-        assert result.status_code == 502
-        assert result.headers.get("Content-Type") == "application/json"
-        assert json.loads(result.content)["message"] == "Internal server error"
 
     # Missing certificate creation to create a domain
     # this might end up being a bigger issue to fix until we have a validated certificate we can use
--- a/tests/aws/services/apigateway/test_apigateway_basic.py
+++ b/tests/aws/services/apigateway/test_apigateway_basic.py
@@ -594,17 +594,19 @@ class TestAPIGateway:
         result = retry(invoke, sleep=2, retries=10, url=url)
         assert result.content == b""
 
-    @markers.aws.unknown
-    def test_api_gateway_mock_integration(self, create_rest_apigw, aws_client):
+    @markers.aws.validated
+    def test_api_gateway_mock_integration(self, create_rest_apigw, aws_client, snapshot):
         rest_api_name = f"apigw-{short_uid()}"
+        stage_name = "test"
         rest_api_id, _, _ = create_rest_apigw(name=rest_api_name)
 
         spec_file = load_file(TEST_IMPORT_MOCK_INTEGRATION)
         aws_client.apigateway.put_rest_api(restApiId=rest_api_id, body=spec_file, mode="overwrite")
+        aws_client.apigateway.create_deployment(restApiId=rest_api_id, stageName=stage_name)
 
-        url = path_based_url(api_id=rest_api_id, stage_name="latest", path="/echo/foobar")
+        url = api_invoke_url(api_id=rest_api_id, stage=stage_name, path="/echo/foobar")
         response = requests.get(url)
-        assert response._content == b'{"echo": "foobar", "response": "mocked"}'
+        snapshot.match("mocked-response", response.json())
 
     @pytest.mark.xfail(reason="Behaviour is not AWS compliant, need to recreate this test")
     @markers.aws.unknown
--- a/tests/aws/services/apigateway/test_apigateway_basic.py
+++ b/tests/aws/services/apigateway/test_apigateway_basic.py
@@ -66,7 +66,6 @@ from tests.aws.services.apigateway.conftest import (
     STEPFUNCTIONS_ASSUME_ROLE_POLICY,
 )
 from tests.aws.services.lambda_.test_lambda import (
-    TEST_LAMBDA_HTTP_RUST,
     TEST_LAMBDA_NODEJS,
     TEST_LAMBDA_NODEJS_APIGW_INTEGRATION,
     TEST_LAMBDA_PYTHON,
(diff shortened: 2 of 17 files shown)


tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -1756,13 +1669,34 @@ def test_apigw_call_api_with_aws_endpoint_url(aws_client, region_name):
 
 
 @pytest.mark.parametrize("method", ["GET", "ANY"])
-@pytest.mark.parametrize("url_function", [path_based_url, host_based_url])
-@markers.aws.unknown
+@pytest.mark.parametrize("url_type", [path_based_url, UrlType.HOST_BASED])
+@markers.aws.validated
+# TODO clean up the client instances. We might not need 4.
+#  We might also be ok not parametrizing the method, as it doesn't seem to be what we are testing here
 def test_rest_api_multi_region(
```
bentsku, line 1676: it seems we have snapshots created for this test, but I can't see the `snapshot` fixture. Should we delete them from the snapshot file?

tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -1836,24 +1771,32 @@ def test_rest_api_multi_region(
...
+        api_us_id, stage=stage_name, path="/demo", region="us-west-1", url_type=url_type
+    )
+    retry(_invoke_url, retries=20, sleep=2, url=endpoint)
 
     delete_rest_api(apigateway_client_eu, restApiId=api_eu_id)
     delete_rest_api(apigateway_client_us, restApiId=api_us_id)
-    aws_client_factory(region_name="eu-west-1").lambda_.delete_function(FunctionName=lambda_name)
-    aws_client_factory(region_name="us-west-1").lambda_.delete_function(FunctionName=lambda_name)
+    lambda_eu_west_1_client.delete_function(FunctionName=lambda_name)
+    lambda_us_west_1_client.delete_function(FunctionName=lambda_name)
```
bentsku, line 1794: As we're using fixtures, I believe those are not necessary anymore!

tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -1836,24 +1771,32 @@ def test_rest_api_multi_region(
...
 
 
 class TestIntegrations:
-    @markers.aws.unknown
+    @markers.aws.needs_fixing
+    # Attempts to create invalid integration with proxy
```
bentsku, line 1799: oh yes, this is a bad one 😭 we have some reports about the S3 integration, I had started a PR a long time ago and never finished it 😅

tests/aws/services/apigateway/test_apigateway_lambda.py
```diff
@@ -832,3 +836,64 @@ def invoke_api(url):
         elif lambda_format_type == "wrong-format":
             assert response.status_code == 502
             assert response.json() == {"message": "Internal server error"}
+
+
+@markers.aws.validated
+@markers.only_on_amd64
+def test_lambda_rust_proxy_integration(
```
bentsku, line 843: awesome, I was worried about deleting the test above but didn't see it was migrated! 🥳

tests/aws/services/apigateway/test_apigateway_basic.py
```diff
@@ -2012,108 +1965,6 @@ def test_api_gateway_http_integrations(
         assert expected == content["data"]
         assert ctype == headers["content-type"]
 
-    @markers.aws.unknown
-    def test_api_gateway_kinesis_integration(
```
bentsku, line 2016: this was already tested in `test_apigateway_kinesis.py`, right?

tests/aws/services/apigateway/apigateway_fixtures.py
```diff
@@ -198,11 +198,17 @@ class UrlType(Enum):
 
 
 def api_invoke_url(
```
bentsku, line 200: at some point, we should think about migrating this as a real fixture, so we could use the `region_name` fixture as well. Could be useful! it's used everywhere 😬

tests/aws/services/apigateway/test_apigateway_basic.snapshot.json
```diff
@@ -116,5 +116,138 @@
...
+    "recorded-content": {}
+  },
+  "tests/aws/services/apigateway/test_apigateway_basic.py::test_rest_api_multi_region[UrlType.HOST_BASED-GET]": {
+    "recorded-date": "29-05-2024, 23:24:41",
+    "recorded-content": {}
+  },
+  "tests/aws/services/apigateway/test_apigateway_basic.py::test_rest_api_multi_region[UrlType.HOST_BASED-ANY]": {
+    "recorded-date": "29-05-2024, 23:25:19",
+    "recorded-content": {}
+  },
```
bentsku, line 144: this looks like leftovers

bentsku (approved):
Wow, massive amount of work here! That is really great, well done!

Minor comment about some leftovers snapshots it seems, but really, validating these was really needed, thank you so much for doing all of this. 

Thank you so much! 🙏

tests/aws/services/apigateway/test_apigateway_basic.snapshot.json
cloutierMat: Good catch I had intended to use a snapshot, when I realised that asserting 200 was enough for what this test is testing. I will remove.

tests/aws/services/apigateway/test_apigateway_basic.py
cloutierMat: You are right, I didn't realise the lambda fixture was storing the client along with the function name.

tests/aws/services/apigateway/test_apigateway_basic.py
cloutierMat: Yes there are 5 tests that were already validated, I captured it with the commit message as I commited 1 test per commit.
- removed `test_malformed_response_apigw_invocation` tested by `test_apigateway.test_apigateway_lambda.py::test_lambda_aws_proxy_response_format`
- `test_api_gateway_update_resource_path_part` removed as covered in `test_apigateway.TestApiGatewayApiResource.test_resource_lifecycle`
- `test_api_gateway_kinesis_integration` removed already validated with `test_apigateway_kinesis.py::test_apigateway_to_kinesis`
- `test_api_gateway_sqs_integration_with_event_source` removed as validated by `test_apigateway_sqs.py`
- `test_http_integration` removed as testing same functionality as `test_apigateway_http.test_http_integration_with_lambda`

tests/aws/services/apigateway/test_apigateway_lambda.py
joe4dev: @bentsku Does Rust do something weirdly specific here that is important for the API Gateway integration?

I'm primarily wondering about ARM compatibility. The API Gateway also has these two parametrized tests left in ext: https://github.com/localstack/localstack-ext/pull/2899
- tests/aws/services/apigateway/test_rest_apis.py::TestRestAPIs::test_custom_authorizer[NONE-GET]
- tests/aws/services/apigateway/test_rest_apis.py::TestRestAPIs::test_custom_authorizer[NONE-POST]
- tests/aws/services/apigateway/test_rest_apis.py::TestRestAPIs::test_custom_authorizer[CUSTOM-GET]
- tests/aws/services/apigateway/test_rest_apis.py::TestRestAPIs::test_custom_authorizer[CUSTOM-POST]
- tests/aws/services/apigateway/test_websocket_apis.py::TestWebSockets::test_websocket_api_with_http_integration[HTTP-False-True]
- tests/aws/services/apigateway/test_websocket_apis.py::TestWebSockets::test_websocket_api_with_http_integration[HTTP-False-False]

If we have a strong reason for using Rust rather here, we could consider adding it to an ARM-compatible cached build under the `lambda_/functions/common` as demonstrated here: `tests/aws/services/lambda_/functions/common/uncaughtexception/provided/Makefile`

Secondly, I'm not a big fan of packaged zip functions because they are black boxes: hard to understand and maintain.

----

Awesome work @cloutierMat 👏👏👏 🚀🚀🚀

bentsku: I believe it does, yes, most probably it's about the API Gateway context serialization/deserialization which must be pretty strict in Rust. Here's the PR introducing it. 

We could add it, yes! Not too sure what's the value of testing ARM here, as I believe it will behave strictly the same. 
Maybe @dfangl can chime in too, he was the author of the PR introducing it, but I think having Rust tests for API Gateway kinda makes sense.

tests/aws/services/apigateway/test_apigateway_basic.py
bentsku: Ah right, sorry, haven't really read the commit messages 😅

tests/aws/services/apigateway/test_apigateway_basic.py
bentsku: Ah, I actually didn't realize... well, good catch for verifying 😅

tests/aws/services/apigateway/test_apigateway_lambda.py
cloutierMat: #6005 was the original PR

joe4dev: > I believe it does, yes, most probably it's about the API Gateway context serialization/deserialization which must be pretty strict in Rust. Here's the PR introducing it.

Thanks for the clarification. I'm mainly curious about what makes this specific to Rust (beyond being a good regression example). Wouldn't snapshotting the context achieve the same?

> We could add it, yes! Not too sure what's the value of testing ARM here, as I believe it will behave strictly the same.

Valid assumption 👍  I think it's valuable to separate what fails on ARM via `@only_on_amd64` from "CI optimization skips" because we need a clearer picture of our ARM gaps (i.e., LS broken vs. test broken).

bentsku: > Thanks for the clarification. I'm mainly curious about what makes this specific to Rust (beyond being a good regression example). Wouldn't snapshotting the context achieve the same?

True, yes! I think Rust might be stricter than our snapshot regarding types, etc? However the PR was done before the snapshots were really into use. I think having at least one APIGW <-> lambda test with a strongly typed language shouldn't hurt just as a sanity check, but I agree with the second point about separating the markers. Any idea how we could go about that?

joe4dev: Alright, two suggestions then:
1. **Snapshots vs. Rust**: Use snapshot testing wherever possible and add a comment to clarify the need for a Rust Lambda. Snapshots also gives more actionable errors compared to some random Rust serialization message.
2. **ARM skips**: Only use the existing Markers (e.g., `@only_on_amd`) for flagging broken LS functionality. We can temporarily use it for "broken" tests with an appropriate comment. Adding a `reason` field could be helpful. I would ignore pre-mature CI optimization for now.

cloutierMat: Thanks for the input @joe4dev. Based on it and given the scope of this ticket, I will make the following changes.

- Change the assert for a snapshot in the current Rust test
- Add comment to explain it's existence
- Remove the `only_on_amd` tag

There are already tests making snapshots of the lambda event context ie: `tests.aws.services.apigateway.test_apigateway_lambda.test_lambda_aws_proxy_integration` and adding new tests falls out of the current work scope: Remove `aws.unknown` tags.

dfangl: Chiming in here, as this is my test: it predates the snapshotting, so I am in principle fine with replacing it. However, I think it still has it's validity, as there are still some shortcomings in the snapshots (like types after snapshot transforms, date precision if opted out), where a test in one or multiple statically typed languages makes sense.

I would still advocate for having at least one test per payload Type in some statically typed languages, with the specific payload Type.

bentsku: Wouldn't removing the `only_amd_64` flag make the test fail now, as the function must be compiled for AMD64, hence the flag? 
This is the build command:
`cargo build --release --target x86_64-unknown-linux-musl`

I think we could leave it for now, as it's not exactly the scope of this PR, but leave a TODO on the test to address this in a future iteration to make it compatible for both AMD64 and ARM?

cloutierMat: Good catch! Will do!

PR #10929, opened 2024-05-30
remove unknown-markers in test_kms

sannya-singal:
## Motivation
As an initiative to remove all `markers.aws.unknown` markers from the tests, this PR aims to deal with the ones present in kms tests.

## Changes
This PR: 
- changes the marker to `validated` if it runs against AWS
- changed the marker to `needs_fixing` if it didn't runs against AWS

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -540,7 +540,7 @@ class TestKMS:
             )
         assert exc.match("ValidationException")
 
-    @markers.aws.unknown
+    @markers.aws.validated
     def test_invalid_key_usage(self, kms_create_key, aws_client):
         key_id = kms_create_key(KeyUsage="ENCRYPT_DECRYPT", KeySpec="RSA_4096")["KeyId"]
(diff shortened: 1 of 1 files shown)


tests/aws/services/kms/test_kms.py
```diff
@@ -540,7 +540,7 @@ def test_sign_verify(self, kms_create_key, snapshot, key_spec, sign_algo, aws_cl
             )
         assert exc.match("ValidationException")
 
-    @markers.aws.unknown
+    @markers.aws.validated
```
cloutierMat, line 543: Strange, the tests already had a validated timestamp, but were not flagged as validated.

Easy wins, then! :smile:

cloutierMat (approved):
:rocket: One more service down!

tests/aws/services/kms/test_kms.py
sannya-singal: Yeah, It was strange to me as well in the first place and so I ran them twice to confirm if I am missing on something 😄

PR #10931, opened 2024-05-30
Fix k8s dev script code locations

simonrw:
## Motivation

The kubernetes dev script generates incorrect templates. Specifically:
* paths have not been updated to account for the code restructuring from #10800, and
* the `hostPath` syntax is incorrect

## Changes

- Update root path location after code relocation
- Mount community entry points
- Correct schema of hostPath mounts


pinzon (approved):
Thanks for fixing this

PR #10934, opened 2024-05-30
remove unknown markers from S3 tests 

bentsku:
## Motivation
This PR is a part of removing `unknown` test markers.

## Changes
The only `unknown` markers were on the cross account tests. As those are simplified tests with no IAM permissions, verifying that we can access cross-account resources, those might be marked `only_localstack` for now.
It could also be considered `needs_fixing`, but I think we could move those tests as part of the IAM tests for cross buckets access? 
Also, I believe you would need a set of 2 AWS accounts to validate it, making use of the secondary account fixture, which not many people have? 

I've tried to look around what already existed for other services, and some are `only_localstack`, some `validated` and some `unknown`. Maybe @viren-nadkarni can chime in for how those tests should be considered? 

I've also taken the opportunity to validate the CloudFormation tests related to S3:
- `tests.aws.services.cloudformation.resources.test_s3.test_bucketpolicy`
- `tests.aws.services.cloudformation.resources.test_s3.test_bucket_versioning`

### DIFF
--- a/localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
+++ b/localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
@@ -448,10 +448,11 @@ class S3BucketProvider(ResourceProvider[S3BucketProperties]):
             s3_client.put_bucket_notification_configuration(**put_config)
 
         if version_conf := model.get("VersioningConfiguration"):
+            # from the documentation, it seems we `Status` is a required parameter
             s3_client.put_bucket_versioning(
                 Bucket=model["BucketName"],
                 VersioningConfiguration={
-                    "Status": version_conf.get("Status", "Disabled"),
+                    "Status": version_conf.get("Status", "Suspended"),
                 },
             )
 
--- a/localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
+++ b/localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
@@ -448,7 +448,7 @@ class S3BucketProvider(ResourceProvider[S3BucketProperties]):
             s3_client.put_bucket_notification_configuration(**put_config)
 
         if version_conf := model.get("VersioningConfiguration"):
-            # from the documentation, it seems we `Status` is a required parameter
+            # from the documentation, it seems that `Status` is a required parameter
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
```diff
@@ -448,10 +448,11 @@ def create(
             s3_client.put_bucket_notification_configuration(**put_config)
 
         if version_conf := model.get("VersioningConfiguration"):
+            # from the documentation, it seems we `Status` is a required parameter
```
cloutierMat, line 451: typo?
```suggestion
            # from the documentation, it seems that `Status` is a required parameter
```

tests/aws/services/cloudformation/resources/test_s3.py
```diff
@@ -28,10 +31,11 @@ def test_bucketpolicy(deploy_cfn_template, aws_client):
...
         template_mapping={"include_policy": False},
     )
-    with pytest.raises(Exception) as err:
-        aws_client.s3.get_bucket_policy(Bucket=bucket_name).get("Policy")
+    with pytest.raises(ClientError) as err:
+        aws_client.s3.get_bucket_policy(Bucket=bucket_name)
+    snapshot.match("no-policy", err.value.response)
 
-    assert err.value.response["Error"]["Code"] == "NoSuchBucketPolicy"
+    # assert err.value.response["Error"]["Code"] == "NoSuchBucketPolicy"
```
cloutierMat, line 38: Should we remove that line entirely instead of commenting it out as it is part of the snapshot.

cloutierMat (approved):
Looks good to me! :rocket: 

I left a couple small comment.

localstack-core/localstack/services/s3/resource_providers/aws_s3_bucket.py
bentsku: Oops 😇 thanks!

tests/aws/services/cloudformation/resources/test_s3.py
bentsku: Yes, definitely! good catch, was a bit too fast here 😅

Issue #10925, opened 2024-05-29
bug: get-object-attributes breaks when `X-Amz-Object-Attributes` has spaces in it.


docwhat:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

hen making a call to `get_object_attributes()` in Ruby using the `aws-sdk-s3` Gem I get only partial responses.

For example:

```rb
s3_client.get_object_attributes(
  bucket: s3_bucket,
  key: input_s3_key,
  object_attributes: %w[ETag Checksum ObjectParts StorageClass ObjectSize]
)
```

Will return something like this:
```
#<struct Aws::S3::Types::GetObjectAttributesOutput delete_marker=nil, last_modified=2024-05-29 21:46:59 +0000, version_id=nil, request_charged=nil, etag="eca36cd6c16e750db2e256cee9c6daec", checksum=nil, object_parts=nil, storage_class=nil, object_size=nil>
```

The exact attribute(s) returned don't seem to be consistent.

### Expected Behavior

This can be reproduced with `curl`. It seems the new Gem puts spaces in the `X-Amz-Object-Attributes` which gives Localstack fits.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

Note the spaces in the `X-Amz-Object-Attributes` header.

```console
❯ curl --url 'http://localhost:4566/test-temp/input-eca36cd6c16e750db2e256cee9c6daec-551837.pdf?attributes' -H 'X-Amz-Object-Attributes: ETag, Checksum, ObjectParts, StorageClass, ObjectSize'
<?xml version='1.0' encoding='utf-8'?>
<GetObjectAttributesOutput xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><ETag>eca36cd6c16e750db2e256cee9c6daec</ETag></GetObjectAttributesOutput>
```

### Environment

```markdown
- OS: macOS
- LocalStack:
  LocalStack version: stable, 3.4.0
  LocalStack Docker image sha: `sha256:8569dadc82ddf0627569775dde517297351691ddbe28e60edd1f515b46f7ef2e`
  LocalStack Docker image digest: `sha256:54fcf172f6ff70909e1e26652c3bb4587282890aff0d02c20aa7695469476ac0`
[...]

PR #10935, opened 2024-05-31
fix header list parsing in ASF

bentsku:
## Motivation
As reported in #10925, we've had an issue parsing header list for the `GetObjectAttributes` S3 operation, following some updates in the Ruby SDK. The reporter also opened an issue in the ruby SDK repo, where they linked the RFC regarding header list and that optional whitespace is indeed accepted. 
RFC: https://www.rfc-editor.org/rfc/rfc9110.html#name-lists-rule-abnf-extension

## Changes
- create an AWS validated test to validate the change
- add test for the parser
- add the fix in the parser: stripping the whitespace of header list values

_fixes #10925_


thrau (approved):
:+1: nice catch, great explanation (even with a link to a spec), simple fix :100:

PR #10937, opened 2024-05-31
validate unknown dynamodb streams test and update behavior

bentsku:
## Motivation
`test_dynamodb_stream_records_with_update_item` still had the `unknown` marker. 
While AWS validating it, it showed that our implementation was not right: we would send an event if the update operation was not actually updating the record. So I quickly fixed it and add some validations in the test for this use-case.

## Changes
- validated `test_dynamodb_stream_records_with_update_item`
- changed the logic in `update_item`, `put_item` and `transact_write_items` to not send events if the update operation did not actually update the record
- updated two test for `put_item` and `transact_write_items` validation of behavior

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -1418,26 +1418,37 @@ class TestDynamoDB:
         retry(lambda: _get_records_amount(4), sleep=1, retries=3)
         snapshot.match("get-records", {"Records": records})
 
-    @pytest.mark.xfail(reason="this test flakes regularly in CI")
-    @markers.aws.unknown
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            "$..SizeBytes",
+            "$..ProvisionedThroughput.NumberOfDecreasesToday",
+            "$..StreamDescription.CreationRequestDateTime",
+        ]
+    )
     def test_dynamodb_stream_records_with_update_item(
-        self, dynamodb_create_table, wait_for_stream_ready, aws_client
+        self,
+        aws_client,
+        dynamodb_create_table_with_parameters,
+        wait_for_dynamodb_stream_ready,
+        snapshot,
+        dynamodbstreams_snapshot_transformers,
     ):
         table_name = f"test-ddb-table-{short_uid()}"
 
-        dynamodb_create_table(
-            table_name=table_name,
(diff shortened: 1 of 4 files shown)


tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -1464,38 +1480,57 @@ def test_dynamodb_stream_records_with_update_item(
...
+            assert insert_seq_number >= starting_sequence_number
+        assert isinstance(
+            event_update["dynamodb"]["ApproximateCreationDateTime"],
+            datetime,
+        )
+        assert event_update["dynamodb"]["ApproximateCreationDateTime"].microsecond == 0
+        assert event_update["eventVersion"] == "1.1"
+        assert event_update["eventName"] == "MODIFY"
+        assert "OldImage" in event_update["dynamodb"]
+        assert int(event_update["dynamodb"]["SequenceNumber"]) > starting_sequence_number
```
cloutierMat, line 1533: Most of those assertions seem redundant with the snapshot, unless I am missing something...

tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -1446,15 +1460,17 @@ def test_dynamodb_stream_records_with_update_item(
...
         )
+        snapshot.match("get-shard-iterator", response)
         assert response["ResponseMetadata"]["HTTPStatusCode"] == 200
         assert "ShardIterator" in response
-        iterator_id = response["ShardIterator"]
+        shard_iterator = response["ShardIterator"]
 
-        item_id = short_uid()
+        item_id = "my-item-id"
+        # assert that when we insert/update the record with the same value, no event is sent
```
cloutierMat, line 1473: :rocket:

cloutierMat (approved):
LGTM :rocket: 

Are all of the assertion made because snapshot transform? It doesn't seem like there are that many even in the `dynamodbstreams_snapshot_transformers`. If they serve a purpose, I have no problem with them. I just feel like the snapshot should help prevent so many asserts :smile:

tests/aws/services/dynamodb/test_dynamodb.py
bentsku: Good point yes, I was supposed to remove a lot of them. I believe the microsecond one, datetime type and the sequence number could be kept, but the rest can go! 🧹

viren-nadkarni (approved):
LGTM, thanks for fixing this :+1:

PR #10938, opened 2024-05-31
CFn: add simple template validations

simonrw:
## Motivation

We currently allow templates that are not allowed on AWS. For example, extra top level fields for resources, or outputs without a `Value` field.

Users deploying these get a false sense of security when deploying CFn stacks. To make LocalStack more useful to users of CloudFormation, we should repliate AWS behaviour and perform template validations.

## Changes

* Add new `validations` submodule to the CloudFormation service
* Define a "static" template validation type, which is performed on the template only
* Implement two checks
    * that the resources blocks are well formed
    * that the outputs all have a `Value`, and that it is not statically known to be `null`

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/validations.py
+++ b/localstack-core/localstack/services/cloudformation/engine/validations.py
@@ -39,6 +39,42 @@ def outputs_have_values(template: dict):
             raise ValidationError(f"[{key}] 'null' values are not allowed in templates")
 
 
+def resources_top_level_keys(template: dict):
+    """
+    Validate that each resource
+    - there is a resources key
+    - includes the `Properties` key
+    - does not include any other keys that should not be there
+    """
+    resources = template.get("Resources")
+    if resources is None:
+        raise ValidationError(
+            "Template format error: At least one Resources member must be defined."
+        )
+
+    allowed_keys = {
+        "Type",
+        "Properties",
+        "DependsOn",
+        "CreationPolicy",
+        "DeletionPolicy",
+        "Metadata",
+        "UpdatePolicy",
+        "UpdateReplacePolicy",
+    }
+    for resource_id, resource in resources.items():
+        if "Type" not in resource:
+            raise ValidationError(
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/cloudformation/engine/validations.py
```diff
@@ -0,0 +1,81 @@
...
+    Base class for static analysis of the template
+    """
+
+    def __call__(self, template: dict):
+        """
+        Execute a specific validation on the template
+        """
+
+
+def outputs_have_values(template: dict):
```
dominikschubert, line 28: Thanks :grin:

localstack-core/localstack/services/cloudformation/engine/validations.py
```diff
@@ -0,0 +1,81 @@
+"""
```
dominikschubert, line 1: Super useful. Guess we could look into making use of jsonschema for this or did you encounter any particular validations where this wouldn't work?

tests/aws/services/cloudformation/api/test_validations.py
```diff
@@ -0,0 +1,74 @@
+import json
+
+import pytest
+
+from localstack.testing.pytest import markers
+
+
+@markers.aws.validated
```
dominikschubert, line 8: pretty nice that those validations are even in parity now :rocket:

dominikschubert (approved):
LGTM, great initial step to help avoid some common/easy-to-make mistakes when handling CFn templates!
Would be quite interested in also checking if we can just provide a jsonschema and validating a few more required fields that way

localstack-core/localstack/services/cloudformation/engine/validations.py
simonrw: That's a nice suggestion. I didn't look into it, but my concern is that we would lose the granularity of the error messages presented to the user. Perhaps we could somehow instrument the validation so we could customise the error messages, but with the framework I set out for validations I hope we could extend them further, leading on to full `jsonschema` validation in the future 👍

PR #10941, opened 2024-06-03
Fixed scaffolding generation for CFN resources

Morijarti:
## Motivation
Fixed broken CFn resource generation introduced with new project structure changes

## Changes
- Minor changes in paths that take into account locastack-core


simonrw (approved):
Thank you for updating these paths, yes they got left in the code layout migration :+1:

PR #10945, opened 2024-06-03
move cloudformation deploy ui to cloudformation provider

thrau:
## Motivation

In an effort to reduce coupling between generic and aws-specific runtime components #10946, I'm moving the cloudformation deploy UI `/_localstack/cloudformation/deploy` to the cloudformation provider package, and registering it into the internal resources with a hook.

## Changes

* The `CloudFormationUi` now lives in the cloudformation provider package, and is registered with a runtime hook instead of a static import


pinzon (approved):
Great :+1:

PR #10947, opened 2024-06-03
remove unused localstack.services.messages

thrau:
## Motivation

Removing some dead code. `localstack.services.messages` used to be part of the generic proxy, which has now been removed for a while. I did not find any usages of the module.

## Changes

* remove `localstack/services/messages.py`


alexrashed (approved):
🧹

PR #10951, opened 2024-06-04
Fix detection of docker image digest for diagnostic purposes

dominikschubert:
## Motivation

I was asked about the different meaning of localstack version/commit/image SHAs and coincidentally noticed that the digests we were showing were actually the local image ids and not the digest on docker hub that we can actually pull for reproducing the issue.

 
## Changes

- The diagnose endpoint now also outputs the image sha that can be used to pull from docker hub via `docker pull localstack/localstack@sha256:....`
- On startup the value of `LocalStack Docker image sha` is now the actual digest that can be pulled (see above). This was a bit unfortunate since we now also refer to this in our GitHub issue template. If there's no manifest yet, it returns the fallback value `Unavailable`

## Discussion

@dfangl I noticed that fortunately the docker client works correctly when called with `image_details = DOCKER_CLIENT.inspect_image(image_name='sha256:6dd80f78fdaa60bcf9dcc3a34a072d3be3f577350e9e46b91f538f54531eef81')
` even though the naming of the parameter in `inspect_image` might not suggest so. The `inspect_container` method calls the parameter `container_name_or_id`, maybe we should do t he same for `inspect_image`  as well? wdyt?

### DIFF
--- a/localstack-core/localstack/services/infra.py
+++ b/localstack-core/localstack/services/infra.py
@@ -209,7 +209,8 @@ def print_runtime_information(in_docker: bool = False):
             container_id = inspect_result["Id"]
             print("LocalStack Docker container id: %s" % container_id[:12])
             image_details = DOCKER_CLIENT.inspect_image(inspect_result["Image"])
-            print("LocalStack Docker image sha: %s" % image_details["RepoDigests"][0])
+            digests = image_details.get("RepoDigests") or ["Unavailable"]
+            print("LocalStack Docker image sha: %s" % digests[0])
         except ContainerException:
             print(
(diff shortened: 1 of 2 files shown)


dominikschubert:
Unfortunately it seems for CI this doesn't work and needs a fallback since the image doesn't have a manifest yet.

dfangl (approved):
Looks good! I think renaming the parameter is a good idea, I honestly also was not aware that it was possible like this, we should definitely document it!

dominikschubert:
> Looks good! I think renaming the parameter is a good idea, I honestly also was not aware that it was possible like this, we should definitely document it!

Gonna do that in a follow-up though so we can get this into the release still

PR #10952, opened 2024-06-04
Refine optional Lambda dependencies

joe4dev:
Introduce test selection: https://github.com/localstack/localstack/pull/10301

## Motivation

As discussed in our Lambda service owner sync (2024-06-03), we missed some optional dependencies and should split the definitions between community + Pro since merging is now supported.

## Changes

* Add missing optional dependencies for Lambda destinations
* Add comments to clarify the integration type
* Move Pro services into ext


silv-io (approved):
thanks for fixing this and also for the nice explanations for each dependency!

alexrashed (approved):

Issue #10914, opened 2024-05-28
bug: S3 checksum is not checked


faltimiras:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

S3 checksum validation in S3 (localstack) do not fail if checksum is invalid.

This code should throw a S3Exception because the pre cacalculated checksum ("thisiswrong") do not match with what is calculated in S3 (localstack), this has been tested in real S3 and it fails as expected

```
try (S3Client s3 = S3Client
                .builder()
                .endpointOverride(localstack.getEndpoint())
                .credentialsProvider(
                        StaticCredentialsProvider.create(
                                AwsBasicCredentials.create(localstack.getAccessKey(), localstack.getSecretKey())
                        )
                )
                .region(Region.of(localstack.getRegion()))
                .build()) {
            s3.createBucket(CreateBucketRequest.builder().bucket(BUCKET).build());

            PutObjectRequest putOb = PutObjectRequest.builder()
                    .bucket(BUCKET)
                    .key("somekey")
                    .checksumSHA256("thisiswrong")
                    .build();

            s3.putObject(putOb, RequestBody.fromString("lololo"));

            var con = s3.getObject(GetObjectRequest.builder().key("somekey").bucket(BUCKET).build()).readAllBytes();
            System.out.println(new String(con));
        }
```

### Expected Behavior

Localstack should throw a S3Exception explaining that checksum provided with checksum calculated do not match

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Testing it with testContainers with
```
 private static final DockerImageName localstackImage = DockerImageName.parse("localstack/localstack:s3-latest");
    @Container
    public static final LocalStackContainer localstack = new LocalStackContainer(localstackImage)
            .withServices(S3);
```

Execute the code in "current behaviour"

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

faltimiras:
Cool! that was fast, thank you!

is this automatically released when it is merged to master? or when i should expect to have the fix released? thanks!

bentsku:
Thank you for the report! It leads to a nice improvement, so that was great! AWS documentation wasn't too explicit on that point.

[...]

PR #10954, opened 2024-06-04
fix Checksum validation and assumptions for S3

bentsku:
## Motivation
As reported by #10914, we had an issue with Checksum validation for S3. We would rely on the `x-amz-sdk-checksum-algorithm` header, but this one is not considered for `PutObject` and `UploadPart`, AWS does not use it. It is only internally used by the SDK to know it needs to automatically calculate the right `x-amz-checksum-{checksum_algorithm}` value to send along. 

This is only valid for `UploadPart` and `PutObject`. Some other requests like `CreateMultipartUpload` uses only the `ChecksumAlgorithm` parameter with no value (same headers but different purposes). 

The logic was then a bit flawed, because when the client would send a checksum we would not properly consider it if the SDK `x-amz-sdk-checksum-algorithm` was not sent. 

## Changes
- add multiple tests related to the change:
  - add a test with no SDK interaction, raw requests sent to understand what is the relation between the 2 headers (result: none, AWS does not consider it at all)
  - add a test with the SDK, but no automatic calculation, just manual, to check the different raised errors
  - add a small check in an existing test regarding `ChecksumAlgorithm` validation for `CreateMultipartUpload`
  - add a small check regarding `UploadPart` to see if it follows the same logic as `PutObject`
  - modify the logic around checksum validation and introduce new helper method to retrieve the right algorithm name from the 4 possibles parameters/headers
  - remove an AWS bug behavior that was introduced in LocalStack as well following snapshots for parity. I've tracked it down to 2 botocore / global AWS SDK issues with empty `ContentEncoding` issue, so we just snapshot skip for now
  - also fix checksum check, taking the checksum algorithm from the trailing headers, to know what algorithm is used and set it to the object so that the `AwsChunkedDecoder` correctly picks the value

_fixes #10914_

### DIFF
--- a/localstack-core/localstack/services/s3/constants.py
+++ b/localstack-core/localstack/services/s3/constants.py
@@ -1,4 +1,5 @@
 from localstack.aws.api.s3 import (
+    ChecksumAlgorithm,
     Grantee,
     Permission,
     PublicAccessBlockConfiguration,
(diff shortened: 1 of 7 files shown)


steffyP (approved):
wow 🤯 this seems like a huge effort to reproduce and test all the different cases. great work, @bentsku  👏

PR #10955, opened 2024-06-04
SQS: Add MessageSystemAttributeNames for receive-message

### LINKED ISSUE #10908: bug: Latest localstack doesn't support message-system-attribute-names in sqs

### Current Behavior

Currently, passing **message-system-attribute-names** to localstack when receiving messages from sqs, does not work. This is a change that AWS introduced recently and is really important, because the previous argument (attribute-names) has been deprecated. 
More information: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sqs/send-message.html 

### Expected Behavior

Message-system-attribute-names parameter works when receving messages in sqs.

Custom (please describe below)

### Steps To Reproduce

aws --endpoint-url=http://localhost:4566 sqs receive-message ....

### Environment

```markdown
- OS: Ubuntu 20.04.6 LTS
- LocalStack:
  LocalStack version:2.4.0 (also tried latest)
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

### Anything else?

_No response_

baermat:
## Motivation
AWS deprecated the `attribute-names` parameter for `receive-message` operation and changed it to `message-system-attribute-names`. This PR addresses this change.
fixes #10908 

## Changes
- adds the necessary alias handling in LocalStack
- add tests
- add custom entry to query spec. Our sqs-query spec is not updated automatically. Kudos to @bentsku for going through the query spec code with me! For now this PR adds this change manually, if the solution is too hacky we have to reconsider. /cc @alexrashed

### DIFF
--- a/localstack-core/localstack/aws/data/sqs-query/2012-11-05/service-2.json
+++ b/localstack-core/localstack/aws/data/sqs-query/2012-11-05/service-2.json
@@ -1166,6 +1166,9 @@
         "AttributeNames":{
           "shape":"AttributeNameList"
         },
+        "MessageSystemAttributeNames":{
+          "shape":"AttributeNameList"
+        },
         "MessageAttributeNames":{
           "shape":"MessageAttributeNameList"
         },
--- a/localstack-core/localstack/aws/data/sqs-query/2012-11-05/README.md
+++ b/localstack-core/localstack/aws/data/sqs-query/2012-11-05/README.md
@@ -6,3 +6,5 @@ This switch removed a lot of spec data which is necessary for the proper parsing
 - The file is licensed with Apache License 2.0.
 - Modifications:
   - Removal of documentation strings with the following regex: `(,)?\n\s+"documentation":".*"`
+  - Added `MessageSystemAttributeNames` to `ReceiveMessageRequest.members` with AWS deprecating `AttributeNames`.
+  The patches in `spec-patches.json` are not present in the boto client for our sqs-query tests right now because the custom loading is not fully integrated at the moment, so it is changed directly in the spec.
--- a/localstack-core/localstack/services/sqs/provider.py
+++ b/localstack-core/localstack/services/sqs/provider.py
@@ -1239,8 +1239,11 @@ class SqsProvider(SqsApi, ServiceLifecycleHook):
 
         # prepare result
         messages = []
+        message_system_attribute_names = message_system_attribute_names or attribute_names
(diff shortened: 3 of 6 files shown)


palarnab (commented):
Thanks. This should work

localstack-core/localstack/aws/data/sqs-query/2012-11-05/service-2.json
```diff
@@ -1166,6 +1166,9 @@
         "AttributeNames":{
           "shape":"AttributeNameList"
         },
+        "MessageSystemAttributeNames":{
+          "shape":"AttributeNameList"
+        },
```
alexrashed, line 1171: I am afraid we might loose the overview of how we modified the original query spec. There would still be our version control, but in fact we already have a very well working way of maintaining these patches:
https://github.com/localstack/localstack/blob/930aac67c6516e58428b59dd3520c9d79ae9f2f1/localstack-core/localstack/aws/spec-patches.json

This file contains JSON Patches which are applied to the services when being loaded in our spec loader:
https://github.com/localstack/localstack/blob/930aac67c6516e58428b59dd3520c9d79ae9f2f1/localstack-core/localstack/aws/spec.py#L64-L70

Your changes here would be equivalent to the following addition to the `spec-patches.json`:
```json
  "sqs-query/2012-11-05/service-2": [
    {
      "op": "add",
      "path": "/shapes/ReceiveMessageRequest/members/MessageSystemAttributeNames",
      "value": {
        "shape": "AttributeNameList"
      }
    }
  ],
```

For the sake of transparency / keeping track of the changes, I would prefer this way.
Another option would be to add a note to the list of modifications on the internalized specs:
https://github.com/localstack/localstack/blob/930aac67c6516e58428b59dd3520c9d79ae9f2f1/localstack-core/localstack/aws/data/sqs-query/2012-11-05/README.md?plain=1#L7-L8

alexrashed (commented):
Awesome! Thanks a lot for jumping on this issue! 🦸🏽 
The implementation and the tests are looking great and are executed for both supported protocols!
My only request would be to either migrate the spec changes to our `spec-patches.json` (my preferred option), or to mention the modification in the README next to the spec (see my comment).
Afterwards we're good to go! 🚀

alexrashed (approved):
Awesome! Thanks for adding the note and doing a final round of validation! This is looking great and ready to be merged! 🚀

HenrikPublic:
Will this be added to 3.5?

PR #10956, opened 2024-06-04
add codeowners validation to gh actions

lukqw:
## Motivation

Codeowners can get outdated quickly when refactoring code, or shuffling around modules.
This PR aims to introduce a safety net which warns a user about changes in PRs that would introduce such issues to the Codeowners file 

## Changes

This PR adds a small workflow to our suite of actions, which on each PR raised will verify whether the codeowners file is still valid.

## Testing
In the checks of this PR you should already see the workflow with the name `LocalStack - Validate Codeowners` show up.
It's currently red, as the codeowners file has some issues.

## TODO

What's left to do:

- [x] clean up the CODEOWNERS file, for which I'd need input on from the localstack team

```
    [err] line 83: "/tests/unit/utils/testing/" does not match any files in repository
```

Should we just remove this line?

### DIFF
--- a/.github/workflows/validate-codeowners.yml
+++ b/.github/workflows/validate-codeowners.yml
@@ -10,7 +10,12 @@ on:
 
 jobs:
   validate-codeowners:
-    uses: localstack/meta/.github/workflows/validate-codeowners.yml@add-validate-codeowners
-    with:
-      codeowners-file: './CODEOWNERS'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
 
+      - name: Validate codeowners
+        uses: mszostok/codeowners-validator@v0.7.4
+        with:
+          checks: "files"
--- a/.github/workflows/validate-codeowners.yml
+++ b/.github/workflows/validate-codeowners.yml
@@ -18,4 +18,5 @@ jobs:
       - name: Validate codeowners
         uses: mszostok/codeowners-validator@v0.7.4
         with:
-          checks: "files"
+          checks: "files,duppatterns,syntax"
+          experimental_checks: "avoid-shadowing"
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-codeowners.yml
```diff
@@ -0,0 +1,21 @@
...
+  validate-codeowners:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Validate codeowners
+        uses: mszostok/codeowners-validator@v0.7.4
+        with:
+          checks: "files"
```
alexrashed, line 21: I think it would be nice to also enable the `duppatterns`, the `syntax`, and the (experimental, but well working) `avoid-shadowing` check. The `owners` check requires a github token, but I think it's not important enough.
```suggestion
        with:
          checks: "files,duppatterns,syntax"
          experimental_checks: "avoid-shadowing"
```

alexrashed (approved):
Thanks a lot for jumping on this! This is a great tool to make sure that the `CODEOWNERS` file stays up-to-date and valid, and could be a nice tool in case we want to introduce more rules (for example every submodule of x has to have an owner). 🧹 I would even say we should enable additional checkers (see comment)! 🥳 

Concerning your question in the PR description:
> Should we just remove this line?

I had a quick look, and yes you can remove this line.
This line was matching this module: https://github.com/localstack/localstack/tree/v2.3.0/tests/unit/utils/testing
However, it was moved to [localstack/localstack-snapshot](https://github.com/localstack/localstack-snapshot) with https://github.com/localstack/localstack/pull/10164.

.github/workflows/validate-codeowners.yml
lukqw: Thanks for raising this comment, I wasn't aware of these features!
I have adjusted the workflow to include the addition checks, seems like we're good to go as everything has passed 🚀

Issue #10946, opened 2024-06-03
Reduce coupling between generic and AWS specific code


thrau:
As part of our efforts to better modularize the localstack codebase, we are pushing to reduce the coupling between generic runtime components, and those that are aws specific.

Ultimately, all aws-specific code should live in `localstack.aws`, but even individual code files often mix generic and aws concepts.

This issue serves as a way to track all PR related to this refactoring effort.

HarshCasper:
Due to the forthcoming changes in our distribution and development model ([read more](https://blog.localstack.cloud/the-road-ahead-for-localstack/)), this issue tracker is no longer being maintained, and this issue is now being closed. If you'd like us to take a look or need further support, please reach out through one of our support channels:

-   💬 [GitHub Discussions](https://github.com/orgs/localstack/discussions)
-   📖 [Help & Support](https://docs.localstack.cloud/aws/getting-started/help-support/)

PR #10958, opened 2024-06-04
move aws-specific patches into localstack.aws.patches

thrau:
## Motivation

As part of the effort to reduce coupling of runtime and AWS code #10946, I'm moving some patches that are specific to moto AWS into their own package, and applying them using a hook rather than statically in the infra code. This will also help in #10942.

I left some parts of the `patch_moto_request_handling` because I'd like to separately investigate whether we can simplify or remove it.

## Changes

* `patch_instance_tracker_meta` is now applied using a hook
* patching the moto config to set `load_aws_managed_policies` is now done in its own patch

### DIFF
--- a/localstack-core/localstack/aws/patches.py
+++ b/localstack-core/localstack/aws/patches.py
@@ -1,0 +1,58 @@
+from importlib.util import find_spec
+
+from localstack.runtime import hooks
+from localstack.utils.patch import patch
+
+
+def patch_moto_instance_tracker_meta():
+    """
+    Avoid instance collection for moto dashboard. Introduced in
+    https://github.com/localstack/localstack/pull/3250.
+    """
+    from moto.core.base_backend import InstanceTrackerMeta
+    from moto.core.common_models import BaseModel
+
+    if hasattr(InstanceTrackerMeta, "_ls_patch_applied"):
+        return  # ensure we're not applying the patch multiple times
+
+    @patch(InstanceTrackerMeta.__new__, pass_target=False)
+    def new_instance(meta, name, bases, dct):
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/aws/patches.py
```diff
@@ -0,0 +1,58 @@
...
+    Enable loading AWS IAM managed policies in moto by default.Introduced in
+    https://github.com/localstack/localstack/pull/10112.
+    """
+    from moto.core.config import default_user_config
+
+    default_user_config["iam"]["load_aws_managed_policies"] = True
+
+
+# TODO: this could be improved by introducing a hook specifically for applying global patches that is run
+#  before any other code is imported.
```
dominikschubert, line 46: good idea, that would make it a lot more explicit

localstack-core/localstack/utils/aws/request_context.py
```diff
@@ -166,11 +166,6 @@ def patch_moto_request_handling():
 
     from moto.core import utils as moto_utils
 
-    # enable loading AWS IAM managed policies
-    from moto.core.config import default_user_config
-
-    default_user_config["iam"]["load_aws_managed_policies"] = True
-
     # make sure we properly handle/propagate "not implemented" errors
     @patch(moto_utils.convert_to_flask_response.__call__)
```
dominikschubert, line 170: what about moving those patches as well?

dominikschubert (approved):
LGTM! :broom:

localstack-core/localstack/utils/aws/request_context.py
dominikschubert: nevermind, forgot that this was mentioned in the description of the PR :+1:

PR #10959, opened 2024-06-04
add DaemonAwareThreadPool for thread pools to block shutdown

thrau:
## Motivation

It's know that ThreadPoolsExecutors don't care about daemon threads. They add all their threads to a global list which is then iterated and joined on `atexit`. We have a patch in pro that adds this functionality, and we also added it a while back for hypercorn, which now [lives in rolo](https://github.com/localstack/rolo/blob/b1c24d7bea94012fece1323dc6fd616e53f44485/rolo/gateway/asgi.py#L13-L20).

This PR adds the functionality as a re-usable class, and uses it to implement the `AdaptiveThreadPool`, which is the thing being patched by pro. This allows us to get rid of the patch, which will help #10942.

## Changes

* `THREAD_POOL` will no longer join its threads on exit if the threads are daemon threads

### DIFF
--- a/localstack-core/localstack/utils/asyncio.py
+++ b/localstack-core/localstack/utils/asyncio.py
@@ -1,5 +1,5 @@
 import asyncio
-import concurrent.futures
+import concurrent.futures.thread
 import functools
 import time
 from contextvars import copy_context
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
Nice :+1: Makes a lot more sense to have this here

PR #10960, opened 2024-06-05
fix S3 image publishing

bentsku:
## Motivation
As seen here with this manual run to publish the S3 image, we did not checkout the repo in the publish job or download the docker helper script, so we did not have access to it.
https://github.com/localstack/localstack/actions/runs/9382410974/job/25834104662

## Changes
- ~download the docker helper script in order to be able to execute it~
- checkout the repo in order to have the script available and the git hash

## Testing

Currently running with the `publish` flag set to `true` in order to test the fix, as I need a new image out to contain all the latest changes.

### DIFF
--- a/.github/workflows/tests-s3-image.yml
+++ b/.github/workflows/tests-s3-image.yml
@@ -43,7 +43,7 @@ on:
         description: 'Publish S3-only images on Dockerhub'
         required: false
         type: boolean
-        default: false
+        default: true
       PYTEST_LOGLEVEL:
         type: choice
         description: Loglevel for PyTest
--- a/.github/workflows/tests-s3-image.yml
+++ b/.github/workflows/tests-s3-image.yml
@@ -194,6 +194,11 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-s3-image.yml
```diff
@@ -43,7 +43,7 @@ on:
         description: 'Publish S3-only images on Dockerhub'
         required: false
         type: boolean
-        default: false
+        default: true
```
alexrashed, line 46: Quick reminder to remove this before merging the PR :)

alexrashed (approved):
Thanks a lot for jumping on this! This slipped through with https://github.com/localstack/localstack/pull/10872. Sorry about that!
The changes are looking good! Just added a comment as a reminder to reset the default after testing :)

.github/workflows/tests-s3-image.yml
bentsku: It actually did not work 😭 Maybe I had 2 conditions, I think there was something up with default values. I'll try something else

bentsku: @alexrashed it seems I can't try it on a branch anymore because the helper script forces it to be on the master branch. I suppose I'll merge this iteration and we can see then? 😅

PR #10962, opened 2024-06-05
validate STS tests

bentsku:
## Motivation
TBD

## Changes
TBD

### DIFF
--- a/tests/aws/services/sts/test_sts.py
+++ b/tests/aws/services/sts/test_sts.py
@@ -12,6 +12,7 @@ from localstack.testing.pytest import markers
 from localstack.utils.aws.request_context import mock_aws_request_headers
 from localstack.utils.numbers import is_number
 from localstack.utils.strings import short_uid, to_str
+from localstack.utils.sync import retry
 
 TEST_SAML_ASSERTION = """
 <?xml version="1.0"?>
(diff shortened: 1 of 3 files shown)


tests/aws/services/sts/test_sts.py
```diff
@@ -92,8 +93,53 @@
...
+                RoleArn=created_role["Role"]["Arn"], RoleSessionName=test_role_session_name
+            )
+            return assume_role_resp
+
+        response = retry(assume_role, sleep=5, retries=4)
+        snapshot.match("assume-role", response)
+
+        if response["AssumedRoleUser"]["AssumedRoleId"]:
+            assume_role_id_parts = response["AssumedRoleUser"]["AssumedRoleId"].split(":")
+            assert assume_role_id_parts[1] == test_role_session_name
```
cloutierMat, line 139: Isn't that a redundant check, since we already have a snapshot for the response? Aws seems to use the `RoleSessionName` in the ARN, so it is validating that both part are equal already?

```json
        "AssumedRoleUser": {
          "Arn": "arn:aws:sts::111111111111:assumed-role/<resource:1>/<resource:2>",
          "AssumedRoleId": "<role-id:1>:<resource:2>"
        },
```

cloutierMat (approved):
Good work the test split there! Very valuable :rocket:

tests/aws/services/sts/test_sts.py
bentsku: Yes, I think the snapshot validates that the second part of `AssumeRoleId` is the same as the last part of the `Arn` resource.

But from the snapshot, nothing tells us that the last part of the `AssumedRoleUser` is the `test_role_session_name` I would say. It's kinda implicit looking at it? I could also instead of this check, just create a snapshot entry `snapshot.match("role-session-name", {"RoleSessionName": test_role_session_name})` and then create a transformer on it, that might clearer than `"<resource:2>"`?

bentsku: Updated with https://github.com/localstack/localstack/pull/10962/commits/475a6f600e2c5ece235dd609cbbaac4128009a65, tell me if that's clearer 😄

cloutierMat: Love it! :heart:

PR #10964, opened 2024-06-05
fix S3 aws-chunked when checksum is provided

bentsku:
## Motivation
Follow up from #10954, a case has not been thought of in this PR, when the checksum would be manually provided as a header and not part of a trailing header with the `aws-chunked` encoding.

This PR fixes this case. 

## Changes
- update the logic so that if the checksum is provided, do not override it by trying to get the trailing headers that might not be present
- add a test for this use case

_fixes #10914_ (again)


cloutierMat (approved):
Good catch Quick Fix! :rocket:

PR #6091, opened 2022-05-18
sqs tag map parsing

thrau:
**WIP**

I've added a test for #4391

PR #10966, opened 2024-06-06
Fix sqs urlencoded parsing

### LINKED ISSUE #10949: bug: MessageAttributes not being passed in APIGW SQS integration

### Current Behavior

When sending a message to SQS using APIGW integration the message attributes are not sent to SQS (or not read by SQS)

### Expected Behavior

The SQS message should have message attributes added to it.

With a docker-compose file

### Steps To Reproduce

When creating an integration between APIGW and SQS, the message attributes are not passed to SQS
### Integration
```shell
awslocal apigateway put-integration \
    --type AWS \
    --rest-api-id $API_ID \
    --resource-id $RESOURCE_ID \
    --http-method POST \
    --integration-http-method POST \
    --request-parameters method.request.header.Content-Type='application/x-www-form-urlencoded' \
    --request-templates application/json='Action=SendMessage&MessageBody=$util.urlEncode($util.base64Encode($input.body))&MessageAttributes.1.Name=source&MessageAttributes.1.Value.DataType=String&MessageAttributes.1.Value.StringValue=source-name' \
    --uri arn:aws:apigateway:us-east-1:sqs:path/000000000000/myQueue \
    --passthrough-behavior NEVER
```

 When I’m invoking this endpoint the message body is transformed correctly but the message attributes are not added
```shell
awslocal sqs receive-message 
    --queue-url http://localhost.localstack.cloud:4566/queue/us-east-1/000000000000/myQueue \
    --max-number-of-messages 10 \
    --message-attribute-names All
```
### Output
```json
{
    "Messages": [
        {
            "MessageId": "96950d76-e2c4-425f-973d-7fdeb4b750ec",
            "ReceiptHandle": "NjhhZGZiODItMTM2Zi00DU1",
            "MD5OfBody": "3685b3d1513e1f432e6dbdb44aaff98d",
            "Body": "eyJhZSI6IDEwfQ==",
            "Attributes": {
                "SenderId": "000000000000",
                "SentTimestamp": "1717313365742",
                "ApproximateReceiveCount": "2",
                "ApproximateFirstReceiveTimestamp": "1717313434634"
            }
        }
    ]
}
````

### Environment

```markdown
- OS: MacOS14.4.1 
- LocalStack:
  LocalStack version: localstack/localstack:3.3
  LocalStack Docker image sha: 16d96c2200fe
```

### Anything else?

_No response_

cloutierMat:
## Motivation

Fixes #10949 

It appears to be a similar issue to #6091 where sqs accepts both `MessageAttribute.1.Name` and `MessageAttributes.1.Name` when resolving url encoded forms.

## Changes

In addition to verifying for the presence of `.1.Key` we also need to look for `.1.Name` when deciding if we use `MessageAttribute` or `MessageAttributes` to build the object.

### DIFF
--- a/localstack-core/localstack/aws/protocol/parser.py
+++ b/localstack-core/localstack/aws/protocol/parser.py
@@ -1096,6 +1096,8 @@ class SQSQueryRequestParser(QueryRequestParser):
         For example, both works for the TagQueue operation:
         - Using the proper serialized name "Tag": Tag.1.Key=key&Tag.1.Value=value
         - Using the member name "Tag" in the parent structure: Tags.1.Key=key&Tags.1.Value=value
+        - Using "Name" to represent the Key for a nested dict: MessageAttributes.1.Name=key&MessageAttributes.1.Value.StringValue=value
+            resulting in {MessageAttributes: {key : {StringValue: value}}}
         The Java SDK implements the second variant: https://github.com/aws/aws-sdk-java-v2/issues/2524
         This has been approved to be a bug and against the spec, but since the client has a lot of users, and AWS SQS
         supports both, we need to handle it here.
@@ -1103,21 +1105,23 @@ class SQSQueryRequestParser(QueryRequestParser):
         # ask the super implementation for the proper serialized name
         primary_name = super()._get_serialized_name(shape, default_name, node)
 
-        # determine a potential suffix for the name of the member in the node
-        suffix = ""
+        # determine potential suffixes for the name of the member in the node
+        suffixes = []
         if shape.type_name == "map":
             if not shape.serialization.get("flattened"):
-                suffix = ".entry.1.Key"
+                suffixes = [".entry.1.Key", ".entry.1.Name"]
             else:
-                suffix = ".1.Key"
+                suffixes = [".1.Key", ".1.Name"]
         if shape.type_name == "list":
             if not shape.serialization.get("flattened"):
-                suffix = ".member.1"
+                suffixes = [".member.1"]
             else:
-                suffix = ".1"
+                suffixes = [".1"]
 
         # if the primary name is _not_ available in the node, but the default name is, we use the default name
-        if f"{primary_name}{suffix}" not in node and f"{default_name}{suffix}" in node:
+        if not any(f"{primary_name}{suffix}" in node for suffix in suffixes) and any(
+            f"{default_name}{suffix}" in node for suffix in suffixes
+        ):
             return default_name
         # otherwise we use the primary name
         return primary_name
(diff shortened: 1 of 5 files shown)


tests/aws/services/apigateway/test_apigateway_sqs.py
alexrashed, line 1: This test is great! But it is only testing API Gateway's SQS integration, while the fix is in the SQS parser.
It would be great if you could also add a small unit test in `test_parser.py`:
https://github.com/localstack/localstack/blob/b3a7651ec72417895d7ebfecba2cc2b7479c99d0/tests/unit/aws/protocol/test_parser.py

Having all these tests in the form of simple unit tests allows us to easily refactor / change certain parts of this complex component while being fairly sure that we do not introduce any regressions. :)

alexrashed (commented):
Wow, great find! These parser issues are pretty hard to pin down, and the parser is a pretty complex component! 🦸🏽 
The code is looking great, and the integration test for API Gateway is AWS validated and snapshot verified! However, it would be great to add one more simple unit test to the parser-specific tests, then we're good to go! 🚀

tests/aws/services/apigateway/test_apigateway_sqs.py
cloutierMat: Oups, I had a last commit on my machine, I didn't push last night! :rofl:

Pushed now, the new test will be there!

alexrashed (approved):
Perfect! I really appreciate that you dug into this complex component! 💯 🏅

localstack-core/localstack/aws/protocol/parser.py
```diff
@@ -1096,28 +1096,32 @@ def _get_serialized_name(self, shape: Shape, default_name: str, node: dict) -> s
         For example, both works for the TagQueue operation:
         - Using the proper serialized name "Tag": Tag.1.Key=key&Tag.1.Value=value
         - Using the member name "Tag" in the parent structure: Tags.1.Key=key&Tags.1.Value=value
+        - Using "Name" to represent the Key for a nested dict: MessageAttributes.1.Name=key&MessageAttributes.1.Value.StringValue=value
+            resulting in {MessageAttributes: {key : {StringValue: value}}}
         The Java SDK implements the second variant: https://github.com/aws/aws-sdk-java-v2/issues/2524
         This has been approved to be a bug and against the spec, but since the client has a lot of users, and AWS SQS
         supports both, we need to handle it here.
```
cloutierMat, line 1103: @thrau The comment you left here was absolute gold, and made understanding the current issue and fix so much easier! :bomb: :fire: :rocket:

thrau: i could spot alex' handwriting, but had to double check as well :smile:

PR #10967, opened 2024-06-06
Refactor all pytest xfail markers to skip instead

dominikschubert:
## Motivation

I've noticed we have a lot of tests that are xfailed vs. skipped. I'd argue we actually want to skip all of those, since they can still lead to issues for either the pipeline (e.g. by extending the total duration) or other tests. The whole reason they're xfailed is because there's some issue that we usually don't immediately want to deal with. Often that's issues that lead to timeouts for example where an xfail won't help at all. I'm also not aware that we would've ever actually looked at any XPASS results to "un-xfail" tests.

In favor of simplicity and improving stability of the pipeline, let's fully skip the tests that we know to be flaky/broken.

## Changes

- All previously xfailed tests are now being skipped and thus not executed at all in our tests. This should provide a minor speed up of the pipeline and improve stability a bit.

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_api.py
+++ b/tests/aws/services/apigateway/test_apigateway_api.py
@@ -123,7 +123,7 @@ class TestApiGatewayApiRestApi:
         snapshot.match("get-rest-api-after-delete", response)
 
     @markers.aws.validated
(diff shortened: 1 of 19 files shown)


maxhoheiser (approved):
Makes sense, it also should speed things up in the pipeline :)

PR #10969, opened 2024-06-06
Fix `Dockerfile.s3` healthcheck

krishanbhasin-px:
## Motivation
The latest-pushed `latest-s3` image fails to boot properly for me:

```
 localstack-s3 Pulled 
 Container localstack-s3-1  Creating
 Container localstack-s3-1  Created
 Container localstack-s3-1  Starting
 Container localstack-s3-1  Started
 Container localstack-s3-1  Waiting
<after 60s timeout passes>
container localstack-s3-1 is unhealthy
```

Comparing the logs between the old image that worked for us and the new one that doesn't, I can't see any healthcheck calls.

Shelling into the image, manually trying to run the heathcheck command `./bin/localstack status services --format=json` fails with `ModuleNotFoundError: No module named 'localstack'`

## Changes
Copied over the healthcheck changes in https://github.com/localstack/localstack/pull/10894/ to the second dockerfile

I manually tested this by running the latest image and running `.venv/bin/localstack status services --format=json`, and got a successful response back

## Request:

It looks like only `latest-*` images are being published to dockerhub; would it be possible to return to publishing tagged/pinned versions?

### DIFF
--- a/Dockerfile.s3
+++ b/Dockerfile.s3
@@ -97,7 +97,7 @@ RUN echo /usr/lib/localstack/python-packages/lib/python3.11/site-packages > loca
 # expose edge service and debugpy
 EXPOSE 4566 5678
 
-HEALTHCHECK --interval=10s --start-period=15s --retries=5 --timeout=5s CMD ./bin/localstack status services --format=json
+HEALTHCHECK --interval=10s --start-period=15s --retries=5 --timeout=5s CMD .venv/bin/localstack status services --format=json
 
 # default volume directory
 VOLUME /var/lib/localstack


bentsku (approved):
The change looks good to me! Sorry for the inconvenience on this, as you found out, the issue stems from the same issue as #10894 and is fixed the same way.

Regarding the `latest-s3` tag being published, for now the S3-only image is not aligned to the regular LocalStack tagged version. However, we can think about it and maybe start releasing tagged version for this kind of image, as it starts to get more traction, in order to give it more stability.

bentsku:
@krishanbhasin-px, additionally, could you please accept the CLA by following the instructions in this comment in order for us to be able to merge this PR? https://github.com/localstack/localstack/pull/10969#issuecomment-2151848940

Thank you!

krishanbhasin-px:
@bentsku aaah, I need to get signoff from my employer; I expect it should be straightforward but the relevant people aren't available until next week.

bentsku:
@krishanbhasin-px ah, I see 😬 Would you mind if I quickly open a PR myself to do the fix in order to unblock the S3 image? We can always merge this PR anyway next week in order for your contribution to get in. Would that be alright?

krishanbhasin-px:
@bentsku yes absolutely!

bentsku:
Closing in favor of #10970, thank you again!

PR #10970, opened 2024-06-06
fix S3 image healthcheck

bentsku:
## Motivation
As reported and fixed in #10969, the S3 image health check had not been migrated following the refactoring/move into `localstack-core` (#10800). The fix is the same as #10894, and is now ported to the S3 image as well.

The original PR could not merge quickly due to the CLA, but in order to unblock the S3 image, we can merge this quicker.
Thanks again to @krishanbhasin-px for the fix.

## Changes
- pick the health check from the regular localstack Dockerfile

### DIFF
--- a/Dockerfile.s3
+++ b/Dockerfile.s3
@@ -97,7 +97,7 @@ RUN echo /usr/lib/localstack/python-packages/lib/python3.11/site-packages > loca
 # expose edge service and debugpy
 EXPOSE 4566 5678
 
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks for this fix, and thanks to @krishanbhasin-px for pointing us to it. Since this issue might block a lot of other people, I agree that we should move somewhat fast on this.
It's a pity that we can't give the credit to you (by merging the original PR / getting your code in directly), but maybe you can get the approval and find something else to fix. 😛

thrau (approved):
:heavy_check_mark: Please add @krishanbhasin-px as co-author to the PR (Use the Co-Author syntax in the commit message)

PR #10971, opened 2024-06-06
fix flaky APIGW test

bentsku:
## Motivation
It showed in the CI pipeline that a newly validated test of APIGW was flaky around region snapshot transforms, and took longer than before. 
Also showed up in CircleCI: https://app.circleci.com/insights/github/localstack/localstack/workflows/main/tests?branch=master

## Changes
- updated the test to no longer need an echo server
- replace the hard coded region and account id with fixtures, as it was not important for the test
- modify the snapshot of exception to validate the whole exception value from boto


alexrashed (approved):
Thanks a lot for jumping on this flaky test proactively! It showed up quite often recently, really great that it was fixed before it was skipped 😛 🦸🏽

bentsku:
@alexrashed thanks! Worst (or best??) thing is that it didn't even show in one of my PR, I just went to check the CI health dashboard and showed up big there so I was like: oops 😬
So, it was extremely useful! 🙏

PR #10972, opened 2024-06-06
Add support for placeholder values in env-file flag for docker flags

dfangl:
## Motivation
While revisting the features of the `--env-file` parameter for the docker CLI, I noticed we did not include the support for placeholders (variable names without an `=`) in #10880.

To achieve parity with the docker CLI client, I implemented the missing feature + tests.

## Changes
* An env file like this:

```
TEST
```

with the `TEST` environment variable set inside the container, should correctly be passed (with that value) into the Lambda container (for `LAMBDA_DOCKER_FLAGS`).


dominikschubert (approved):
LGTM, nice catch

PR #10973, opened 2024-06-06
fix flaky new DDB test

bentsku:
## Motivation
As seen [in this run](https://app.circleci.com/pipelines/github/localstack/localstack/25500/workflows/dd484d9d-b92e-4fca-92f1-40298a91e645/jobs/214277), the ARM test failed for `test_dynamodb_stream_records_with_update_item`.
This looks like it can be due to parallel request being processed and not leaving the time to DynamoDB to properly ingest the first item (so it does not return it when looking if the object already exists when "updating" the item with the second call). 

This is a race condition, DynamoDB has a somewhat eventual consistency, so maybe this can happen in AWS, or this could be treated as a race condition/bug in LocalStack, but we do not have a real way of preventing it for now. At least the test is more robust now. 

## Changes
- update the test to make sure the first request has been properly processed, before sending the same request again

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -1471,17 +1471,39 @@ class TestDynamoDB:
 
         item_id = "my-item-id"
         # assert that when we insert/update the record with the same value, no event is sent
-        for _ in range(2):
-            aws_client.dynamodb.update_item(
-                TableName=table_name,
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Thanks for another proactive fix of a flaky test! 🦸🏽 
This is really highly appreciated and everyone else benefits from less distractions due to a more stable pipeline! 🚀

PR #10977, opened 2024-06-07
validate unknown tests - ec2

cloutierMat:
## Motivation

Part of the push to get rid of `unknown` tests.

## Changes

I did not create fixtures for the validated tests and used cleanups instead. Vpc resources are really order dependent for deletion and it allowed for a better control.

* test_create_route_table_association => `validated`
* test_create_vpc_endpoint => `needs_fixing`, multiple assumptions are wrong and some moto issues with default values
* test_reserved_instance_api => `only_localstack` I don't want to start $100 bills
* test_vcp_peering_difference_regions => `validated`
* test_describe_vpn_gateways_filter_by_vpc => `validated`
* test_describe_vpc_endpoints_with_filter => `needs_fixing` AWS returns a lot of data. We might want to adapt this test to be more efficient and robust against aws 
* test_pickle_ec2_backend => `only_localstack`

### DIFF
--- a/tests/aws/services/ec2/test_ec2.py
+++ b/tests/aws/services/ec2/test_ec2.py
@@ -81,8 +81,11 @@ class TestEc2Integrations:
         ]:
             assert route_tables["Associations"] == []
 
-    @markers.aws.unknown
-    def test_create_vpc_end_point(self, cleanups, aws_client):
+    @markers.aws.needs_fixing
+    # TODO LocalStack fails to delete endpoints
+    #  LocalStack does not properly initiate Endpoints with no VpcEndpointType fix probably needed in moto
+    #  AWS does not allow for lowercase VpcEndpointType: gateway => Gateway
+    def test_create_vpc_endpoint(self, cleanups, aws_client):
         vpc = aws_client.ec2.create_vpc(CidrBlock="10.0.0.0/16")
         cleanups.append(lambda: aws_client.ec2.delete_vpc(VpcId=vpc["Vpc"]["VpcId"]))
         subnet = aws_client.ec2.create_subnet(VpcId=vpc["Vpc"]["VpcId"], CidrBlock="10.0.0.0/24")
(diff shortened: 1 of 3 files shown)


steffyP (approved):
Great work in adding snapshots and validating more of the ec2 tests 🚀 
Thanks also for adding comments to the needs-fixing ones, which gives a lot of additional context!

tests/aws/services/ec2/test_ec2.py
```diff
@@ -170,110 +190,147 @@ def test_reserved_instance_api(self, aws_client):
...
+                snapshot.transform.key_value("vpc1-id"),
+                snapshot.transform.key_value("vpc2-id"),
+                snapshot.transform.key_value("peering-connection-id"),
+            ]
+        )
         region1 = region_name
-        region2 = region_name  # When cross-region peering is supported, change to SECONDARY_TEST_AWS_REGION_NAME
-
-        # Note: different regions currently not supported due to set_default_region_in_headers(..) in edge.py
+        region2 = SECONDARY_TEST_AWS_REGION_NAME
```
bentsku, line 220: nit: I think you can avoid importing this config variable by using the `secondary_region_name` fixture, from the [multi account document](https://github.com/localstack/localstack/tree/master/docs/testing/multi-account-region-testing):
> Importing and using localstack.constants.TEST_ values is not advised.

I've launched a multi account run just to be sure 😄 as we're meddling in region territory here.

https://app.circleci.com/pipelines/github/localstack/localstack/25538/workflows/4147ddde-7bf2-41d7-ad04-0b45a28e5eb2

bentsku (approved):
LGTM! Nice fixes, thanks for picking EC2 😇

Just a comment about the region import, I've launched a multi account/multi region run just in case, as we're having cross-region here. The link is in the comment 😄 

Thank you again, really nice work! 🙏

tests/aws/services/ec2/test_ec2.py
cloutierMat: Thanks for the pointer. I will switch to the appropriate fixture.

cloutierMat: I have updated the code to use the fixture and updated the snapshot to remove the reliance on the hard-coded `ap-southeast-1` . I am not rerunning the pipeline with `randomize-aws-credentials` as it did pass the first time and being that this is a cross region and not a cross-account test, the credentials shouldn't have an effect.

viren-nadkarni (approved):
Thanks for updating these :+1:

PR #10978, opened 2024-06-07
Fix CloudFormation Fn::Transform replacing unintended values

dominikschubert:
## Motivation

Based on a reported issue from a user where they loaded a stepfunction statemachine definition through the `AWS::Include` macro from s3 which lead to the RoleArn missing. 
The issue was that the transformation replaced the whole level of the current object in the iteration and not just the one particular key it's at.

## Changes

- The `Fn::Transform` can now be used with additional keys at the same object level as the intrinsic function.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/transformers.py
+++ b/localstack-core/localstack/services/cloudformation/engine/transformers.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


dominikschubert:
CI had one failing test but that was a flake (install of opensearch failed)

Morijarti (approved):
LGTM

Issue #10307, opened 2024-02-23
bug: .Net lambda not deleting SQS message


barry-roomex:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am experiencing the same as closed issue #8283 but have some more information.

A .Net 6 lambda with SQS event source is executing successfully but the message is not deleted.

Running localstack with trace logging shows the following:

```
2024-02-23 13:29:02 2024-02-23T13:29:02.264  INFO --- [   asgi_gw_1] localstack.request.http    : POST /_localstack_lambda/451df979d84ca72131aa2d5248f3398a/invocations/98e7e200-7e83-4cb7-8e60-b787fcef6c59/response => 202; Request(b'', headers={'Host': '172.17.0.3:4566', 'User-Agent': 'Go-http-client/1.1', 'Content-Length': '0', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip'}); Response(b'', headers={'Content-Type': 'text/plain; charset=utf-8'})
...
2024-02-23 13:29:02 2024-02-23T13:29:02.269 DEBUG --- [b787fcef6c59] l.s.l.e.adapters           : Encountered an exception while handling callback
2024-02-23 13:29:02 Traceback (most recent call last):
2024-02-23 13:29:02   File "/opt/code/localstack/localstack/services/lambda_/event_source_listeners/adapters.py", line 133, in _invoke_sync
2024-02-23 13:29:02     result=to_str(json.loads(result.payload)),
2024-02-23 13:29:02                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-23 13:29:02   File "/usr/local/lib/python3.11/json/__init__.py", line 346, in loads
2024-02-23 13:29:02     return _default_decoder.decode(s)
2024-02-23 13:29:02            ^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-23 13:29:02   File "/usr/local/lib/python3.11/json/decoder.py", line 337, in decode
2024-02-23 13:29:02     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
2024-02-23 13:29:02                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-23 13:29:02   File "/usr/local/lib/python3.11/json/decoder.py", line 355, in raw_decode
2024-02-23 13:29:02     raise JSONDecodeError("Expecting value", s, err.value) from None
[...]

PR #10980, opened 2024-06-07
Fix handling None payload in Lambda Event Source Mapping

barry-roomex:
Fix for successfully invoked lambdas not deleting source SQS message if the lambda does not have a return value

## Motivation
To resolve the issue I have been having using LocalStack with .net lambdas https://github.com/localstack/localstack/issues/10307

## Changes
It checks if the lambda returned a value before trying to parse it as json

## Testing
Please excuse the lack of tests included with this, I am not a python developer.

To manually test:
* Create a new .net AWS lambda using the `AWS Lambda Project (.NET Core - C#)` template from the Amazon.Lambda.Templates. This will create a function handler without any return value.
* Package and deploy the sample lambda
* Associate with SQS/SNS
* Send a test message

Before this change the message would not be deleted

See the steps to reproduce in https://github.com/localstack/localstack/issues/10307

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_listeners/adapters.py
+++ b/localstack-core/localstack/services/lambda_/event_source_listeners/adapters.py
@@ -128,9 +128,7 @@ class EventSourceAsfAdapter(EventSourceAdapter):
                 error = None
                 if result.is_error:
                     error = "?"
-                result_payload = (
-                    to_str(json.loads(result.payload)) if result.payload else ""
-                )
+                result_payload = to_str(json.loads(result.payload)) if result.payload else ""
                 callback(
                     result=LegacyInvocationResult(
                         result=result_payload,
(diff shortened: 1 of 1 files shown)


barry-roomex:

barry-roomex:
I do not have access to apply a semver label to this PR

joe4dev:
Hi @barry-roomex 

Thank you for raising the issue and contributing this PR 👏👏👏 .

Ultimately, we should add an [AWS Parity Test](https://blog.localstack.cloud/2022-08-04-parity-explained/) to validate the behavior with AWS. We are working on improving Lambda Event Source Mapping due to several issues, so you can expect further improvements in this area 🚀 
Therefore, I am considering merging this to unblock you and follow up as part of our project.

joe4dev (approved):
LGTM 👍 Thank you for the fix and refactoring.

I made a note in our backlog to add a regression test as part of our ongoing rework.

PR #10982, opened 2024-06-07
introduce handler chain (next gen) in apigateway v1 invocation

bentsku:
## Motivation
This is the first implementation of the handler chain inside an edge router handler for API Gateway REST APIs invocations.

Currently, APIGW registers an edge router handler which then invokes a function which passes a context around, where the separation of concerns is not very clear (it does a few things globally, like checking for UsagePlans API Key, or body/parameters validation), but then defers most of it to the `BackendIntegration` subclasses, which do things quite differently between themselves.
However, when looking at API Gateway and how they trigger/dispatch requests, a lot should be done outside the integrations themselves. 

Using the new LocalStack constructs, we will be able to have clear separations of concerns between different steps of the invocation. This PR already introduces a few handlers, but more will come (API Key validation as a separate step). We are also making use of `CompositeHandler` in order for the code to be extendable from outside. 

As a next step, we will also make use of `Plugin` and `PluginManager` to register the different AWS integrations transparently, and the `IntegrationHandler` will contain the logic necessary to load them and invoke them. 

Additionally, a parity issue is present in LocalStack: once we create a REST API, it is available to be invoked under the `execute-api` route. However, this is not the case in AWS. You will need to create a `Deployment`, which is a snapshot (also referred as an "executable" of your API by AWS) of your current REST API, which then can be served by linking that deployment to a `Stage`. Currently, in LocalStack, invoking your API will always hit your "current" REST API, meaning if you do some updates, they will automatically be used while invoking the API. While this might be useful, this is very far from parity (we might re-introduce this mode with a feature flag, something akin to "hot reloading" of your API, but this would be opt-in) and it prevents things like testing canary releases or having different stages pointing to different versions of your API. 
This new implementation, freezing the REST API and populating the context with it, will simplify greatly how to fetch the data in order to do the work during the invocation steps. 

Currently, in order to work in an iterative way, we have introduced a few handlers outlining the future work to split the current code into a few different handlers, but one "temporary" or "legacy" handler holds all the invocation. This PR simply introduces the concept of a handler chain for the invocation. 

## Changes
- create a new opt-in provider called `next_gen` in order to separate the new code and registrations of routes from the current implementation
- create a new edge router handler, registered on the global APIGW routes. This handler instantiates a subclass or the `rolo.Gateway` in order to pass the incoming invocation request through a `HandlerChain`. This handler can be subclassed.
- the `RestApiGateway` contains a first iteration of the handler list to be executed while invoking a REST API
- introduce a new subclass of the `RequestContext` which will be extended by future PRs, to contain all the necessary data for a successful REST API invocation. 
- some new fields in the APIGW store in order to store the deployments and active deployments for the edge router handler to properly understand what API is being invoked and populate the context
- a few fixes to existing `only_localstack` or `needs_fixing` tests that did not deploy the REST API before invoking it. 

## TODO

What's left to do:

- [ ] remove the forced usage of this new prototype in `config.py` before merging


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
```diff
@@ -0,0 +1,19 @@
...
+
+from localstack.http import Response
+
+from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
+from ..context import RestApiInvocationContext
+
+LOG = logging.getLogger(__name__)
+
+
+class InvocationRequestParser(RestApiGatewayHandler):
```
bentsku, line 11: naming might be a bit misleading here: we probably think this handler will populate the context from the linked `deployment`, by "routing" the request depending on its path to the right APIGW `Resource` & `Method`. Naming is not set in stone and will be updated in follow up PRs.

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -0,0 +1,114 @@
...
+            context = RestApiInvocationContext(request)
+            self.populate_rest_api_invocation_context(context, api_id, stage)
+
+            self.rest_gateway.process_with_context(context, response)
+
+            return response
+
+        else:
+            # TODO: return right response
+            print("here?")
```
bentsku, line 42: oopsie, already removed in a not-yet-pushed commit

localstack-core/localstack/config.py
```diff
@@ -1032,6 +1032,13 @@ def populate_edge_configuration(
 # Whether to really publish to GCM while using SNS Platform Application (needs credentials)
 LEGACY_SNS_GCM_PUBLISHING = is_env_true("LEGACY_SNS_GCM_PUBLISHING")
 
+# Whether to enable the Next Gen APIGW invocation logic (handler chain)
+# TODO: remove this before merge!
+if not os.environ.get("PROVIDER_OVERRIDE_APIGATEWAY"):
+    os.environ["PROVIDER_OVERRIDE_APIGATEWAY"] = "next_gen"
```
alexrashed, line 1038: TODO remove before merge (as you already mentioned in the PR description, just another reminder 😛)

localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self):
+        super().__init__(context_class=RestApiInvocationContext)
+        self.request_handlers.extend(
+            [
+                handlers.preprocess_request,
+                handlers.method_request_handler,
+                handlers.integration_request_handler,
+                handlers.integration_handler,
+                # temporary handler which executes everything for now
+                handlers.global_temporary_handler,
```
alexrashed, line 18: nit: Previously we used to use the term "legacy" for this kind of backwards compatibility / migration logic.

localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway.py
```diff
@@ -0,0 +1,33 @@
+from rolo import Response
+from rolo.gateway import Gateway
+
+from . import handlers
+from .context import RestApiInvocationContext
+
+
+class RestApiGateway(Gateway):
```
alexrashed, line 8: nit: Would be great to add a bit of documentation here: This is the main part controlling the API GW invocation handling now. Maybe we can add a few sentences (can also be in upcoming iterations) on what this is and how to extend / plug into this.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_response.py
```diff
@@ -0,0 +1,23 @@
...
+from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
+from ..context import RestApiInvocationContext
+
+LOG = logging.getLogger(__name__)
+
+
+class MethodResponseHandler(RestApiGatewayHandler):
+    """
+    This class might not too much, we still need to investigate but from a first look, does not have much impact
+    on the HTTP response
```
alexrashed, line 14: nit: This comment is a bit confusing 😄

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
alexrashed: Interesting, this is exactly what I thought this class would do 😄 
But what is it intended to do then?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/tempory_global.py
alexrashed, line 1: nit: as mentioned above, might be more clear if this module is called `legacy` and the handler is called `LegacyHandler`?

localstack-core/localstack/config.py
```diff
@@ -1032,6 +1032,13 @@ def populate_edge_configuration(
...
+
+APIGW_NEXT_GEN_PROVIDER = os.environ.get("PROVIDER_OVERRIDE_APIGATEWAY", "") == "next_gen"
```
alexrashed, line 1040: nit: I think the term `next_gen` is not very future proof (next one would be "next next gen"? 😛), but honestly I'm struggling a bit to find a better name as well (as `v2` would be highly confusing 😛).

localstack-core/localstack/services/apigateway/next_gen/execute_api/api.py
alexrashed, line 1: Beautiful 😍

tests/integration/test_security.py
```diff
@@ -178,7 +178,7 @@ def test_cors_apigw_not_applied(self, aws_client):
                 "Origin": "https://app.localstack.cloud",
             },
         )
-        assert response.status_code == 404
+        assert response.status_code in (403, 404)
```
alexrashed, line 181: nit: Maybe we could make this parameterized to already take the proper return code in case of the new provider (same below)?

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -0,0 +1,114 @@
...
+from localstack.services.edge import ROUTER
+
+from .context import RestApiInvocationContext
+from .gateway import RestApiGateway
+
+LOG = logging.getLogger(__name__)
+
+
+# TODO: subclass from -ext, and this handler will do the dispatching between v1 and v2 Gateways
+class ApiGatewayHandler:
```
alexrashed, line 19: nit: As discussed, we could maybe rename this to `Endpoint` or `EndpointHandler` to differentiate between `HandlerChain` handlers here (since this is what's invoked by the router on a URL match).

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -0,0 +1,114 @@
...
+    def __init__(self, rest_gateway: RestApiGateway = None, store: ApiGatewayStore = None):
+        self.rest_gateway = rest_gateway or RestApiGateway()
+        # we only access CrossAccount attributes in the handler, so we use a global store in default account and region
+        self._global_store = (
+            store or apigateway_stores[DEFAULT_AWS_ACCOUNT_ID][AWS_REGION_US_EAST_1]
+        )
+
+    def __call__(self, request: Request, **kwargs) -> Response:
+        # api_id can be cased because of custom-tag id
+        api_id, stage = kwargs.get("api_id", "").lower(), kwargs.get("stage")
```
alexrashed, line 29: Awesome! This means we are finally using `werkzeug`'s request parsing directly from the route matching! 🚀 🤩

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -0,0 +1,114 @@
...
+        if self.is_rest_api(api_id, stage):
+            LOG.info("Next-gen handler for APIGW v1 called")
```
alexrashed, line 31: nit: Maybe we shouldn't have the term "next gen" leak into the logs...

alexrashed (approved):
Awesome! This is a great starting point for the upcoming work on API Gateway and you can already see that this will bring way more structure, way less patching, way more extensibility, and way less boilerplate code (f.e. for the request parsing)! 🧹 💯 
The (currently dummy) handlers in the chain already show how well structured this is going to look like afterwards 🤩
I only found some nitpicks, nothing blocking a merge. The changes contained (as soon as the provider override switch is removed again) should be completely safe to merge! 🥳

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_response.py
bentsku: Yes, sorry, this was mostly internal 😅 we realized with @cloutierMat that the `MethodResponse` resource in APIGW might just be for "decorative purposes" and used to create SDKs from the REST API, but we're not 100% sure this handler does nothing. I'll update the comment!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
bentsku: I think this class might still be needed under this name, to parse for example the `RAW_URI` from the request and such things.
And we can also have a "router" handler that will properly match the incoming request to the right APIGW resource. Once we start splitting things, I think naming will become clearer. I still think we might need a parser somewhere 😄

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
bentsku: True, yes, APIGW v1 Endpoint or something similar will be better

tests/integration/test_security.py
bentsku: Thing is, I'm not too sure yet what is the appropriate status code, I believe it is 403. 

How would we parametrize this?
Shouldn't we instead use the config variable to do the assert? `if config.APIGW_NEXT_GEN_PROVIDER: assert 403 else assert 404`?

localstack-core/localstack/config.py
bentsku: Yeah, @cloutierMat had the exact (I mean, exactly exact) comment about the "next gen" value. I'd like to find something better too, maybe something related to the handler chain? but then also "leak" implementation detail through the feature flag.  I've been lazy and took the work stream name 😅

bentsku (commented):
Thanks a lot for the review 🙏 will address the comments shortly!
When this PR is green, I will remove the flag and merge 🚀

localstack-core/localstack/config.py
cloutierMat: Could we instead not name this one and already name the current one "LEGACY"? and turn that on by default. then remove the "LEGACY" default when we are ready?

bentsku: That might a bit tricky, because we would need to already override the default, also change it in -ext to override the default and then we'd lose the ability to fall back to the current implementation for -ext.

For now it's still not set in stone so we might be able to still find a better name! 😄

cloutierMat: I figured it might not be ideal... Let's keep the gear rolling! :gear:

alexrashed: Yeah, I was also lacking some creativity for a better name :D This was really just a nitpick ;)

tests/integration/test_security.py
alexrashed: Well, if we don't know it currently then no.
I had the impression that the old code (404) is not in parity with AWS and the new provider will (403).
If this is the case, then we should make sure that the assert for the old provider is properly removed.
However, if we don't know this yet, then it's fine for now. ;)

PR #10983, opened 2024-06-07
Update dev script to mount community code correctly

simonrw:
## Motivation

When running the `localstack.dev.run` script, the LocalStack community source code is not mounted. This is due to #10800 as it moved the source code locations around. Looks like this particular instance was missed.

## Changes

* Mount the community code from `localstack/localstack-core/localstack` rather than `localstack/localstack`.


thrau (approved):
Thanks Simon! Seems this one slipped

Issue #10976, opened 2024-06-06
bug: StepFunctions state machine ARN and execution ARN should support the period (.) character (supported in AWS)


northvankiwiguy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When provided with a state machine ARN that contains a period (.) character, LocalStack returns an InvalidArn exception. However, in AWS it accepts these characters. The same is true for state machine execution ARNs.

### Expected Behavior

It should accept state machine (execution) ARNs that contain a period.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

`    localstack start
`
#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
   $ awslocal stepfunctions create-state-machine --name "my.name" --definition '{"StartAt": "Pass", "States": { "Pass": { "Type": "Pass", "End": true }}}' --role-arn "arn:aws:iam::role/Admin"
   $ awslocal stepfunctions describe-state-machine --state-machine-arn "arn:aws:states:us-east-1:000000000000:stateMachine:my.name"
```

### Environment

```markdown
- OS: MacOS Sonoma and Amazon Linux
- LocalStack:
  LocalStack version: 3.4.1.dev
  LocalStack Docker image sha: sha256:b78aa1229dbdc49ba482d48a9506cd0e8a9bd76390873d1a1cef546baf188241
  LocalStack build date: 2024-05-31
  LocalStack build git hash: 4473f8967
```

### Anything else?

This is an easy fix. I'm going to submit a PR for this myself.

northvankiwiguy:
This is now resolved in V3.5.0.

PR #10985, opened 2024-06-07
Step Functions: Allow describing ARNs that contain periods

northvankiwiguy:
## Motivation

For Step Functions, we can create a state machine or state machine execution that contains a period (.) in the name, but
trying to "describe" that state machine (or execution) incorrectly gives an`InvalidArn` exception. See #10976.

## Changes

Validation code for state machine ARNs and state machine execution ARNs now allows for periods (.) in the name. This matches AWS behaviour.

## Testing

Unit tests were added.

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -136,11 +136,11 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
         visitor.visit(sfn_stores)
 
     _STATE_MACHINE_ARN_REGEX: Final[re.Pattern] = re.compile(
-        r"^arn:aws:states:[a-z0-9-]+:[0-9]{12}:stateMachine:[a-zA-Z0-9-_]+(:\d+)?$"
+        r"^arn:aws:states:[a-z0-9-]+:[0-9]{12}:stateMachine:[a-zA-Z0-9-_.]+(:\d+)?$"
     )
 
     _STATE_MACHINE_EXECUTION_ARN_REGEX: Final[re.Pattern] = re.compile(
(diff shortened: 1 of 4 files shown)


MEPalma (commented):
Thank you for addressing this issue. It is true that currently, the provider does not correctly handle state machine names that contain periods.
We should replace the unit tests added in this PR with [Parity/Snapshot Tests](
https://github.com/localstack/localstack/blob/master/docs/testing/parity-testing/README.md)
Since these tests address the base crud operations, a new test case may be added in `aws.services.stepfunctions.v2.test_sfn_api.TestSnfApi`. The test should at least cover the behaviour on [CreateStateMachine](https://docs.aws.amazon.com/step-functions/latest/apireference/API_CreateStateMachine.html) and [DescribeExecution](https://docs.aws.amazon.com/step-functions/latest/apireference/API_DescribeExecution.html)
Should you prefer, I would be happy to add these changes to your branch. Feel free to let me know. 
Thank you again for taking the time to contribute to the project, and I look forward to your feedback!

northvankiwiguy:
@MEPalma I've made your recommended changes to use snapshot testing. One issue I found was that AWS returns `Redrive*` fields for `DescribeExecution`, but LocalStack doesn't have those fields. I took the approach of removing those fields from the snapshot file. I'm not sure if there's a preferred approach I should have followed.

MEPalma:
@northvankiwiguy Thank you for these changes! 
Regarding the `redrive` fields, instructing the test suite not to validate for these would be the desired approach. To enable reproducibility, snapshots are never edited manually (other snapshot in that test suite may not include redrive fields as the may predate this AWS feature). 
We can do this by simply adding an annotation of this type to the test functions `@markers.snapshot.skip_snapshot_verify(paths=["$..redriveCount", "$..redriveStatus", "$..redriveStatusReason"])`. 
I think this should be the last change needed here. Thanks again!

northvankiwiguy:
@MEPalma Thanks, I've added the annotation to ignore the redrive fields. Great suggestion.

MEPalma (approved):
Thank you!

PR #10986, opened 2024-06-10
add support for downloading python dependancies using urls

macnev2013:
## Motivation
This PR aims to add support to download python packages using urls like `git+https://github.com/localstack/aws-glue-libs.git@glue-0.9#egg=aws-glue-libs`

## Changes
- Adds support for package url for python installer

## Testing
- Tested manually for installing on of the libraries in spark.

### DIFF
--- a/localstack-core/localstack/packages/core.py
+++ b/localstack-core/localstack/packages/core.py
@@ -247,9 +247,9 @@ class PythonPackageInstaller(PackageInstaller):
     normalized_name: str
     """Normalized package name according to PEP440."""
 
-    def __init__(self, name: str, version: str, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


alexrashed (commented):
I think it would make sense to implement this change together with a usage (in order to see the motivation for the change, as well as having a test for it). Another option would be to add (unit) tests for this installer (maybe by mocking the run command) which shows and tests the new feature.

alexrashed:
@macnev2013 I think this one might have already been superseded by other changes? What do you think? Can we close this one?

macnev2013:
Thanks for keeping an eye on this. Yes, we can close this.

PR #10989, opened 2024-06-10
fix asf update action by avoiding aws cli pin

alexrashed:
## Motivation
https://github.com/localstack/localstack/pull/10537 extended the pinning of dependencies performed in the `asf-updates.yaml` action - from just `botocore` to also include the transitive pins `boto3`, `boto3-stubs`, as well as `awscli` - in order to reduce the time it takes to perform the backtracking.
However, for the AWS CLI the versioning is not aligned with `botocore`, which is why we introduced a hacky way to determine the AWS CLI version. This hacky approach now broke due to a minor release, as we can see [in the latest scheduled run of the action](https://github.com/localstack/localstack/actions/runs/9442735748):
```
  ERROR: Cannot install awscli==1.32.122 because these package versions have conflicting dependencies.
  File "/home/runner/work/localstack/localstack/.venv/lib/python3.11/site-packages/pip/_vendor/resolvelib/resolvers.py", line 397, in resolve
    self._add_to_criteria(self.state.criteria, r, parent=None)
  File "/home/runner/work/localstack/localstack/.venv/lib/python3.11/site-packages/pip/_vendor/resolvelib/resolvers.py", line 174, in _add_to_criteria
    raise RequirementsConflicted(criterion)
pip._vendor.resolvelib.resolvers.RequirementsConflicted: Requirements conflict: SpecifierRequirement('awscli==1.32.122')
```

This PR addresses this fix by just removing this hack. It is currently not necessary because:
- The backtracking currently does not take too long.
- We can now switch to `uv` to address performance issues with the backtracking.

## Changes
- Removes the exact pin of `awscli` (and instead just introduces a lower limit) in the `pyproject.toml`.
- Removes the faulty `awscli` pinning from the ASF update action.

#### 💚 Manually:
I tested the commands executed in the ASF action manually and they are successful and the requirements files are updated correctly:
```bash
bin/release-helper.sh set-dep-ver botocore "==1.34.122"
bin/release-helper.sh set-dep-ver boto3 "==1.34.122"
bin/release-helper.sh set-dep-ver boto3-stubs "==1.34.122"
pip-compile --strip-extras --upgrade-package "botocore==1.34.122" --upgrade-package "boto3==1.34.122" --extra base-runtime -o requirements-base-runtime.txt pyproject.toml
pip-compile --strip-extras --upgrade-package "botocore==1.34.122" --upgrade-package "boto3==1.34.122" --upgrade-package "awscli" --extra runtime -o requirements-runtime.txt pyproject.toml
pip-compile --strip-extras --upgrade-package "botocore==1.34.122" --upgrade-package "boto3==1.34.122" --upgrade-package "awscli" --extra test -o requirements-test.txt pyproject.toml
pip-compile --strip-extras --upgrade-package "botocore==1.34.122" --upgrade-package "boto3==1.34.122" --upgrade-package "awscli" --extra dev -o requirements-dev.txt pyproject.toml
pip-compile --strip-extras --upgrade-package "botocore==1.34.122" --upgrade-package "boto3==1.34.122" --upgrade-package "awscli" --upgrade-package "boto3-stubs==1.34.122" --extra typehint -o requirements-typehint.txt pyproject.toml
```
#### 💚 Action execution:
I triggered an execution of the affected action from this branch: https://github.com/localstack/localstack/actions/runs/9444259692
Here is the resulting PR: https://github.com/localstack/localstack/pull/10990 (which has the wrong base - namely the branch of this PR - and will therefore be closed again).

### DIFF
--- a/.github/workflows/asf-updates.yml
+++ b/.github/workflows/asf-updates.yml
@@ -78,18 +78,13 @@ jobs:
           bin/release-helper.sh set-dep-ver boto3 "==$BOTOCORE_VERSION"
           bin/release-helper.sh set-dep-ver boto3-stubs "==$BOTOCORE_VERSION"
 
-          # aws-cli is two minor versions behind botocore
-          AWSCLI_VERSION=$(python -c 'import botocore; from packaging.version import Version; version = Version(botocore.__version__); print(f"{version.major}.{version.minor - 2}.{version.micro}")')
-          echo "Pinning aws-cli to version $AWSCLI_VERSION"
-          bin/release-helper.sh set-dep-ver awscli "==$AWSCLI_VERSION"
-
           # upgrade the requirements files only for the botocore package
           pip install pip-tools
           pip-compile --strip-extras --upgrade-package "botocore==$BOTOCORE_VERSION" --upgrade-package "boto3==$BOTOCORE_VERSION" --extra base-runtime -o requirements-base-runtime.txt pyproject.toml
-          pip-compile --strip-extras --upgrade-package "botocore==$BOTOCORE_VERSION" --upgrade-package "boto3==$BOTOCORE_VERSION" --upgrade-package "awscli==$AWSCLI_VERSION" --extra runtime -o requirements-runtime.txt pyproject.toml
-          pip-compile --strip-extras --upgrade-package "botocore==$BOTOCORE_VERSION" --upgrade-package "boto3==$BOTOCORE_VERSION" --upgrade-package "awscli==$AWSCLI_VERSION" --extra test -o requirements-test.txt pyproject.toml
-          pip-compile --strip-extras --upgrade-package "botocore==$BOTOCORE_VERSION" --upgrade-package "boto3==$BOTOCORE_VERSION" --upgrade-package "awscli==$AWSCLI_VERSION" --extra dev -o requirements-dev.txt pyproject.toml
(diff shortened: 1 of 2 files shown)


.github/workflows/asf-updates.yml
```diff
@@ -78,18 +78,13 @@ jobs:
           bin/release-helper.sh set-dep-ver boto3 "==$BOTOCORE_VERSION"
           bin/release-helper.sh set-dep-ver boto3-stubs "==$BOTOCORE_VERSION"
 
-          # aws-cli is two minor versions behind botocore
-          AWSCLI_VERSION=$(python -c 'import botocore; from packaging.version import Version; version = Version(botocore.__version__); print(f"{version.major}.{version.minor - 2}.{version.micro}")')
-          echo "Pinning aws-cli to version $AWSCLI_VERSION"
-          bin/release-helper.sh set-dep-ver awscli "==$AWSCLI_VERSION"
-
```
silv-io, line 85: We could theoretically treat the awscli separately, instead of basing its version on botocore.

So upgrade installing it in L40 alongside botocore. (Assuming that they won't release incompatible versions alongside each other)

And then we can pull out the selected version for awscli like we do for botocore. 

Don't know if it is worth the overhead however, as it probably won't cause any major problems being upgraded later on with all the rest.

silv-io (approved):
LGTM!

Just a minor thought :)

.github/workflows/asf-updates.yml
alexrashed: `awscli` is updated together with `botocore` because it has an explicit pin on a single version of `botocore`. It always has to be updated directly with `botocore`, which is why it's done here.

Issue #2997, opened 2020-09-17
Using localstack kinesis with AWS SDK


abhishekjoy1:
# Type of request: This is a ...

[x] bug report
[ ] feature request

# Detailed description

I am using AWS-SDK-JAVA with Localstack Kinesis. I have written the following code snippet:

     ```
      AmazonKinesisClientBuilder clientBuilder = AmazonKinesisClientBuilder.standard();
      clientBuilder.setCredentials(new EnvironmentVariableCredentialsProvider());
      clientBuilder.setEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(
          "http://localhost:4568",
          Regions.US_WEST_2.getName()));

      AmazonKinesis kinesisClient = clientBuilder.build();

      PutRecordRequest putRecordRequest = new PutRecordRequest();
      putRecordRequest.setStreamName("test");
      putRecordRequest.setData(ByteBuffer.wrap(String.format("testData-%d", 1).getBytes()));
      putRecordRequest.setPartitionKey(String.format("partitionKey-%d", 1));
      kinesisClient.putRecord(putRecordRequest);

   ```

In this context I have also done the following changes:

Added `export AWS_CBOR_DISABLE=true` in .bash_profile

Added following line in ~/.aws/config:

`cborEnabled = false`

## Expected behavior

This should not throw any error, I think.

## Actual behavior

I keep getting the following error in the Java side:

    `Unable to unmarshal exception response with the unmarshallers provided (Service: AmazonKinesis; Status Code: 502; Error Code: null; Request ID: null)`
  ```

And in localstack, I am getting following error:

 ```
2020-09-17T14:44:50:ERROR:localstack.services.generic_proxy: Error forwarding request: 'utf-8' codec can't decode byte 0xbf in position 0: invalid start byte Traceback (most recent call last):
  File "/opt/code/localstack/localstack/services/generic_proxy.py", line 247, in forward
    listener_result = listener.forward_request(method=method,
  File "/opt/code/localstack/localstack/services/kinesis/kinesis_listener.py", line 28, in forward_request
    data = json.loads(to_str(data or '{}'))
  File "/opt/code/localstack/localstack/utils/common.py", line 735, in to_str
    return obj.decode(encoding, errors) if isinstance(obj, six.binary_type) else obj
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbf in position 0: invalid start byte
```


# Steps to reproduce

## Command used to start LocalStack

`localstack start`

[...]

PR #10994, opened 2024-06-10
replace unknown markers for kinesis-tests

steffyP:
## Motivation
Following up https://github.com/localstack/localstack/issues/9147 to remove all `unknown` markers, this PR tackles some kinesis-tests.

## Changes
* added `validated` to `test_create_stream_without_stream_name_raises` which works with AWS, however when attempting to add a snapshot it shows that the error message is still returned in a different format.
* added `needs_fixing` to two tests that use the LocalStack-internal route to retrieve CBOR encoded records. The test likely relates to https://github.com/localstack/localstack/issues/2997 which reported issues with java AWS SDK.
  *  `test_get_records`
  *  `test_get_records_empty_stream`
* added `localstack-only` to `test_run_kcl` which runs some LocalStack specific implementation.

### DIFF
--- a/tests/aws/services/kinesis/test_kinesis.py
+++ b/tests/aws/services/kinesis/test_kinesis.py
@@ -36,13 +36,14 @@ def get_shard_iterator(stream_name, kinesis_client):
 
 
 class TestKinesis:
-    @markers.aws.unknown
+    @markers.aws.validated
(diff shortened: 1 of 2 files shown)


sannya-singal (approved):
LGTM 🚀

PR #10996, opened 2024-06-10
Unify test selection scripts

silv-io:
## Motivation

The test selection scripts for using PR URLs and commits are almost identical. Making changes to them in future iterations will therefore introduce an overhead. Also, the PR URL based script now how an extended use case of not giving any information (and instead letting the script find the correct base and head commits automatically). Adding such extended use cases makes it more difficult to find the correct location to change should something go wrong. Furthermore, it requires more involved processes to parametrize the test selection process.

## Changes

This PR merges the two scripts for generating test selection files. To separate the use cases, the user can now give optional parameters:

- `--pr-url`: Supports the use case previously supported by `generate_test_selection_by_pr.py`
- `base-commit-sha` & `head-commit-sha`: Supports the use case previously supported by `generate_test_selection_by_commit.py`
- leaving out the optional parameters will lead to the script automatically finding the base and head commit SHAs based on the current branch.

## Testing

Test Selection works as intended here:

https://github.com/localstack/localstack/actions/runs/9450399895 -> test selection correctly grabs the commit information from the PR associated with this branch

and in the PR workflow (https://github.com/localstack/localstack/actions/runs/9450201550/job/26028226361?pr=10996)
-> correctly uses the information from the PR event.

and in the CircleCI workflow (https://app.circleci.com/pipelines/github/localstack/localstack/25604/workflows/14656e45-2d92-486b-9b64-043c417602df/jobs/215410)
--> correctly uses the information received via the PR URL


alexrashed (commented):
Awesome! Thanks for the cleanup! This really makes it easier to understand and work with the script! 🚀 🧹
My only concern is that one use case has not been tested (or at least was not explained in the PR description): What if there is no PR for the commit on the branch, most importantly, what if we run this with enabled test selection on the `master` branch?

alexrashed (approved):
Ah right, sorry for the confusion, thanks for the clarification! We actually already discussed this in https://github.com/localstack/localstack/pull/10913. 🙈

PR #11002, opened 2024-06-11
move legacy_v2 S3 into legacy folder

bentsku:
## Motivation
Just cleaning up, now that the `v3` implementation of S3 is the default, we can move the previous provider into a `legacy` folder to clearly indicate that this is the legacy implementation, and make the `v3` code the "main" code. 

## Changes
- move files around to set the legacy code into a `legacy` package, and make `v3`default


localstack-core/localstack/services/s3/legacy/provider.py
alexrashed, line 1: As we can see with this change (new file, 2032 newly added lines), this is a create instead of a move + edit operation in git. This would mean that you loose (or at least mix up) the complete history of the S3 providers.
It would be great if we could change this to two different commits (one commit for the move command, one for adjusting the imports / changing the code). This would not only greatly simplify the review (which is basically impossible right now), and preserve the history.
Let me know if I can help!

alexrashed (changes requested):
The changes themselves are great! This is a nice cleanup, and will simplify the final removal of the old provider(s) with the next major release! 🚀 
However, it would be a shame to loose the history for the prestigious S3 providers.
Let me know if I can help to untangle that. ;)

localstack-core/localstack/services/s3/legacy/provider.py
bentsku: Good point! I was a bit confused about it as well! I'll try it and if I don't succeed, will ask for help! Thanks!

bentsku: @alexrashed hmm, this is weird. I've separated the commits, individually they are "move" then "edit", but when looking at the diff all together it seems it's not really working...

alexrashed (approved):
Awesome! Thanks for restructuring the commits! 🧹 
The diff view with all commits still can't properly visualize the changes, but it works when looking at the individual commits.
I think this should work just as expected and preserve the history correctly for as long as we do not squash merge these changes. And let's wait for the 3.5 release before merging this.

alexrashed:
FYI: This PR can be merged now. The workflow to warn on minor merges being forbidden has already been disabled, but the "x" doesn't go away apparently. -> :shipit:

PR #11005, opened 2024-06-11
check AWS Support validation status

bentsku:
## Motivation
As part of the effort to validate tests, I've tried to run the AWS Support test against AWS.
It seems you need a Premium account to use the Support API. I had rewritten the test to be easily snapshot-able. 

## Changes
- mark the test a `needs_fixing` as you cannot run them against AWS without a premium plan. Debatable if this is `needs_fixing` or `only_localstack` though, technically it's not by design, but also needs something specific.

### DIFF
--- a/tests/aws/services/support/test_support.py
+++ b/tests/aws/services/support/test_support.py
@@ -4,7 +4,7 @@ from localstack.constants import AWS_REGION_US_EAST_1
 from localstack.testing.pytest import markers
 
 TEST_SUPPORT_CASE = {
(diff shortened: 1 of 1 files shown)


cloutierMat (approved):
You might say it the easy one, but it is a great resolution with great comments added to it!

I personally think that needs_fixing is fine for now. It could be useful though to be able to create a snapshot for them in the future. But that would make those tests harder to maintain...

sannya-singal (approved):
I agree to Mathieu, thanks for fixing this, changes LGTM 🚀

PR #11006, opened 2024-06-11
validate scheduler test and check SWF validation

bentsku:
## Motivation
As those 2 services were small, I've done PR for 2:
- Scheduler test was easy to validate as it does basically nothing
- SWF is more complex, I've added comments to it but the test does not cleanup, does not use fixtures and will most probably need rework

## Changes
- add markers


cloutierMat (approved):
LGTM! Thanks for the added comments!

sannya-singal (approved):
LGTM 🚀

PR #11008, opened 2024-06-12
Feature: Introduce env variable for making sure we only allow supported lambda runtimes

Morijarti:
## Motivation
At the moment, we allow creation of lambda even with runtimes that AWS has deprecated. This PR introduces flag `LAMBDA_RUNTIME_VALIDATION` that allows using only supported runtimes.

## Changes
Current default is set to `0` but in next major change we should be able to set to `1` by default

## Note

When we decide to make this behavior mandatory, we should add this env variable to analytics, to check how many users are depending on running deprecated runtime.

## TODO
- [x] Add log message in case turning the flag off would resolve the error on lambda side (part of preparations for when we move to this turn on by default)
- [x] Implement support for this in `function create`


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -77,6 +82,48 @@ def environment_length_bytes(e: dict) -> int:
...
+    @markers.aws.validated
+    @pytest.mark.parametrize("runtime", DEPRECATED_RUNTIMES)
+    def test_create_deprecated_function_runtime_with_validation_enabled(
+        self, runtime, create_lambda_function, lambda_su_role, aws_client, monkeypatch, snapshot
+    ):
+        monkeypatch.setattr(config, "LAMBDA_RUNTIME_VALIDATION", 1)
+        function_name = f"fn-{short_uid()}"
+
+        with pytest.raises(aws_client.lambda_.exceptions.InvalidParameterValueException) as e:
+            create_lambda_function(
```
Morijarti, line 113: Rather than using fixture, I should switch to caling lambda client directly, in order to avoid retry and speed up tests

dfangl: That would be nice indeed!

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -971,6 +972,51 @@ def create_function(
...
+            case Runtime.dotnetcore3_1:
+                latest_runtime = RUNTIMES_AGGREGATED["dotnet"][0]
+            case _:
+                match = re.match(r"^[a-zA-Z]+", deprecated_runtime)
+                runtime_type = match.group(0)
+
+                if runtime_type in RUNTIMES_AGGREGATED:
+                    latest_runtime = RUNTIMES_AGGREGATED[runtime_type][0]
+                else:
+                    latest_runtime = None
```
dfangl, line 1012: Not a big fan of making this dependent on the order of items in `RUNTIMES_AGGREGATED`, since it is not always latest to newest ordered, and we make an assumption that the suggestions might change with the release of new runtimes.

Maybe introduce a mapping of `deprecated-runtime -> suggestion`, similar to our `RUNTIMES_AGGREGATED`, in runtimes.py? This way we could have a simple lookup in a dict, and have every suggestion on one place. We should regenerate the snapshots of the suggestion exception after a new runtime as well, to see if it changes. Might be worth adding those tests to the runtime marker, WDYT @joe4dev ?

dfangl (changes requested):
I'd love some changes to the detection of runtime replacements, since I really would like to avoid depending on the order in `RUNTIMES_AGGREGATED`.

localstack-core/localstack/services/lambda_/provider.py
Morijarti: I love these suggestions.
I'll add them :)

Thank you.

tests/aws/services/lambda_/test_lambda_api.py
Morijarti: Sounds good. I'll add it :)

localstack-core/localstack/services/lambda_/provider.py
Morijarti: I've added a marker and `DEPRECATED_RUNTIMES_UPGRADES` dictionary

tests/aws/services/lambda_/test_lambda_api.py
Morijarti: Done

dfangl (approved):
Looks good, thanks for addressing my comments!

PR #11009, opened 2024-06-12
Add Nix and Flox to installation methods

bryanhonof:
## Motivation

LocalStack is available in Nix and Flox, and it seems to be working well with those 2 packages managers.

## Changes

Just README changes, no functionality has changed.

### Nix

1. Install the [Nix package manager](https://nixos.org/download/) and follow the steps in the README.
2a. For Nix installations that have the `nix-command` and `flakes` experimental features enabled, one can use `nix profile install`.
2b. In the case those experimental features aren't enabled, use either `nix-shell` or `nix-env`.

### Flox

1. Install the [Flox CLI](https://flox.dev/docs/install-flox/).
2. Create a Flox environment in the current working directory with `flox init`
3. Install `localstack` with `flox install localstack`
4. Activate the environment with `flox activate`

After that, localstack will be available to the user as long as the environment is activated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,34 @@ The `localstack-cli` installation enables you to run the Docker image containing
 
 > **Important**: Do not use `sudo` or run as `root` user. LocalStack must be installed and started entirely under a local non-root user. If you have problems with permissions in macOS High Sierra, install with `pip install --user localstack`
 
+### Nix (MacOS, Linux)
+
+LocalStack can be installed using the Nix package manager, using `nix profile`.
+
+```bash
+nix profile install nixpkgs#localstack
+```
+
+Or `nix-env`.
+
+```bash
+nix-env --install localstack
+```
+
+It's also possible to get a temporary shell with localstack installed in it.
+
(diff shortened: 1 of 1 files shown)


bryanhonof:

alexrashed (changes requested):
Thanks a lot for your contribution, @bryanhonof!
However, the nix package for LocalStack is not maintained by us but by the nix community and it is quite outdated (we are about to release 3.5 tomorrow, [the nix package points towards 3.0.0](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/localstack/default.nix)).
While we really appreciate the community maintaining all these different packaging formats, I don't think we should really actively advertise them for as long as we cannot guarantee that they will be kept up to date.

bryanhonof:
@alexrashed That's completely understandable.
I'll go a head and close this PR for now, then.
Maybe we can revisit it at a later date. 😁

alexrashed:
Thanks for your understanding!

PR #11010, opened 2024-06-12
Allow secondary client fixtures to take credentials from an aws profile

dfangl:
## Motivation
Currently, we only allow setting credentials for the secondary account test fixtures using the `SECONDARY_TEST_AWS_ACCESS_KEY_ID` variables and so forth.

This can be quite cumbersome, especially when you want to take advantage of temporary credentials using the AWS IAM Identity Center.

To mitigate this, I refactored our client fixtures a tad, so we can now use `SECONDARY_TEST_AWS_PROFILE` to set an aws profile, if the tests are run against AWS. If it runs against LS, nothing changes to the current behavior, only that credentials are now set on the session, and not on the clients anymore.

Also, the secondary clients now use a different session, but the impact of that should be negligable.

## Changes
* You can now use `SECONDARY_TEST_AWS_PROFILE` to set the profile used for your secondary aws account for cross region tests. If set, the other `SECONDARY_TEST_*` variables will be ignored.

### DIFF
--- a/localstack-core/localstack/testing/aws/util.py
+++ b/localstack-core/localstack/testing/aws/util.py
@@ -28,6 +28,7 @@ from localstack.testing.config import (
     SECONDARY_TEST_AWS_SECRET_ACCESS_KEY,
     SECONDARY_TEST_AWS_SESSION_TOKEN,
     TEST_AWS_ACCESS_KEY_ID,
+    TEST_AWS_REGION_NAME,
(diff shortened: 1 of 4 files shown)


maxhoheiser:
Works for me. 😀

viren-nadkarni (approved):
Tests assume that the secondary client uses the same region as primary client, but you've fixed it already :+1:

LGTM!

PR #11012, opened 2024-06-12
Fix TableV2 construct with streaming enabled

giograno:
## Motivation
LocalStack is not able to deploy a [TableV2 construct](https://aws.amazon.com/blogs/devops/a-new-and-improved-aws-cdk-construct-for-amazon-dynamodb-tables/) with `StreamSpecification` enabled.
A sample was provided by a user and is reflected in the implemented test.
Specifying the stream view type in the new construct should be enough to enable the stream (while LocalStack was failing with a missing required field).

## Changes
- Improve provider for `AWS::DynamoDB::GlobalTable` to account for `StreamSpecification`;
- Write a validated CDK test.

### DIFF
--- a/localstack-core/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
+++ b/localstack-core/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
@@ -242,7 +242,7 @@ class DynamoDBGlobalTableProvider(ResourceProvider[DynamoDBGlobalTableProperties
             if stream_spec := model.get("StreamSpecification"):
                 create_params["StreamSpecification"] = {
                     "StreamEnabled": True,
-                    **(stream_spec or {}),
+                    **stream_spec,
                 }
 
             creation_response = request.aws_client_factory.dynamodb.create_table(**create_params)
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
```diff
@@ -239,6 +239,12 @@ def create(
                 # rename bool attribute to fit boto call
                 sse_specification["Enabled"] = sse_specification.pop("SSEEnabled")
 
+            if stream_spec := model.get("StreamSpecification"):
+                create_params["StreamSpecification"] = {
+                    "StreamEnabled": True,
+                    **(stream_spec or {}),
```
viren-nadkarni, line 245: Nit: which values for `stream_spec` will cause the `or` case to be invoked?

viren-nadkarni (approved):

localstack-core/localstack/services/dynamodb/resource_providers/aws_dynamodb_globaltable.py
giograno: Nice nit. It shouldn't be possible to hit the `or` case (given the assignment expression). `{}` also won't be a good default value anyway.

PR #11013, opened 2024-06-13
Disable CloudFormation template validation

dominikschubert:
## Motivation

A regression was reported on [Discuss ](https://discuss.localstack.cloud/t/validation-error-when-creating-stack-with-cloudformation/943) where not having a `Type` field in the Resource section leads to the following validation error now:
```
An error occurred (ValidationError) when calling the CreateChangeSet operation: Template format error: [/Resources/NotificationsTable] Every Resources object must contain a Type member.
```

This is generally correct behavior but not in combination with transformations since we don't know which fields will be available when processed before doing the actual processing/transformation.

Changing the order to validate only after processing happened still had issues since we currently modify the template by e.g. adding a `LogicalResourceId` field which then fails the validations again. Therefore I decided to just disable them for now and introduce them again when the template handling is refactored.

## Changes

 - Template validation is now disabled (no active default rules) until further notice. Will re-evaluate this with @simonrw when he  is back. Left in the regression test and new position for the validations though so we can pick it up from there

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/validations.py
+++ b/localstack-core/localstack/services/cloudformation/engine/validations.py
@@ -77,6 +77,10 @@ def resources_top_level_keys(template: dict):
 
 
 DEFAULT_TEMPLATE_VALIDATIONS: list[TemplateValidationStep] = [
-    outputs_have_values,
-    resources_top_level_keys,
+    # FIXME: disabled for now due to the template validation not fitting well with the template that we use here.
+    #  We don't have access to a "raw" processed template here and it's questionable if we should have it at all,
+    #  since later transformations can again introduce issues.
+    #   => Reevaluate this when reworking how we mutate the template dict in the provider
+    # outputs_have_values,
+    # resources_top_level_keys,
 ]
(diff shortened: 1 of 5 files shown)


tests/aws/services/cloudformation/api/test_transformers.py
```diff
@@ -112,3 +112,32 @@ def test_transformer_property_level(deploy_cfn_template, s3_bucket, aws_client,
...
+    template = textwrap.dedent("""
+        Parameters:
+          BucketName:
+            Type: String
+        Resources:
+          MyResource:
+            "Fn::Transform":
+                Name: "AWS::Include"
+                Parameters:
+                  Location: !Sub "s3://${BucketName}/data.yaml"
```
dominikschubert, line 134: this part is basically the regression test for this case now

localstack-core/localstack/services/cloudformation/engine/validations.py
```diff
@@ -76,6 +77,10 @@ def resources_top_level_keys(template: dict):
 
 
 DEFAULT_TEMPLATE_VALIDATIONS: list[TemplateValidationStep] = [
-    outputs_have_values,
-    resources_top_level_keys,
+    # FIXME: disabled for now due to the template validation not fitting well with the template that we use here.
+    #  We don't have access to a "raw" processed template here and it's questionable if we should have it at all,
+    #  since later transformations can again introduce issues.
+    #   => Reevaluate this when reworking how we mutate the template dict in the provider
```
Morijarti, line 83: I completely forgot that raw template isn't the one that get to resource provider.
Good find

Morijarti (approved):
Nice fix and I love that now we have regression test for this.

PR #11016, opened 2024-06-13
validate route53 integration tests against AWS

sannya-singal:
## Motivation
As an initiative to validate tests and remove all `markers.aws.unknown` markers, this PR handles the ones concerning the route53 integration tests. 

## Changes
This PR adds changes for the tests to work against AWS and thereby update the marker to `validated` for: 
- `tests/aws/services/route53/test_route53.py::TestRoute53::test_crud_health_check`
- `tests/aws/services/route53/test_route53.py::TestRoute53::test_associate_vpc_with_hosted_zone`
- `tests/aws/services/route53/test_route53.py::TestRoute53::test_reusable_delegation_sets`

### DIFF
--- a/tests/aws/services/route53/test_route53.py
+++ b/tests/aws/services/route53/test_route53.py
@@ -1,3 +1,5 @@
+from urllib.parse import urlparse
+
 import pytest
 
 from localstack.testing.pytest import markers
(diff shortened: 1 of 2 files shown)


giograno (approved):
It looks good, thank for validating these tests 💯 
Probably it's out of the scope here, but we could also think about introducing snapshotting for these tests as well at some point.

PR #11021, opened 2024-06-13
extend bootstrap lifecycle test to include restarts

alexrashed:
## Motivation
With the latest refactoring of the runtime (https://github.com/localstack/localstack/pull/10942) we had some issues with the restart handling (which can be triggered via the health endpoint).
This issue was fixed in https://github.com/localstack/localstack/pull/11018.
This PR aims at extending an existing bootstrap lifecycle test to also include the restart.

## Changes
- Extends `tests.bootstrap.test_localstack_container_server.TestLocalstackContainerServer.test_lifecycle` to send a restart command and check if the restart in the container is working correctly.


thrau (approved):
Great test! Thanks for adding

PR #11022, opened 2024-06-13
update rolo and fix lambda url form parsing

bentsku:
## Motivation
This PR updates to rolo to include https://github.com/localstack/rolo/pull/13

While writing a test for APIGW upstream, I've realized Lambda URL were parsing the form and just forwarding an empty body instead of the raw body containing the form data, so I've fixed it as well, as it feels logical to fix in the same PR where we update rolo to include a fix for form parsing

## Changes
- update rolo to `0.5.1`
- add a fix for lambda URL, calling `restore_payload` to restore the form/multipart part of the request
- add an AWS validate test for it


dfangl (approved):
Looks good for me, thanks for tackling this!

Issue #3508, opened 2021-01-25
Cloudwatch: Add validation API


jakobbraun:
Hi,

we would like to use localstack to test our CloudWatch connector.
For that, it is required to check if the data was reported correctly (no duplicates, no points missing).

Using the official CloudWatch API we can, however not test this, since it only gives eventual consistency and aggregates past data.
Since the aim of localstack is to simulate this API probably it will also not be possible to test it using localstack.

What would really help here would be a non-AWS standardized API that for example gives access to all reported metrics.
Is such or a similar approach for validating the behavior planned or possible?

whummer:
Thanks for reporting @jakobbraun . A backdoor API to receive the raw CW metrics has been added in #3535. Is that what you had in mind?

Example:
```
$ curl http://localhost:4566/cloudwatch/metrics/raw
{'metrics': [{'ns': 'some/thing', 'n': 'someMetric', 'v': 23.0, 't': '2021-01-30T18:35:30.132Z', 'd': []}, {'ns': 'some/thing', 'n': 'someMetric', 'v': 18.0, 't': '2021-01-30T18:35:30.160Z', 'd': []}, {'ns': 'ug/thing', 'n': 'ug', 'v': 23.0, 't': '2021-01-30T18:35:30.186Z', 'd': []}]}
```
Hope that helps - please let us know if anything else is missing.

jakobbraun:
Thanks a lot! That was exactly what I had in mind. Can you give an estimation of when you will make the next release, and by that, I can use this feature?

whummer:
Thanks @jakobbraun , the change is available in the `latest` Docker image, and we are preparing a new release `0.12.6` which will be available in the course of the day (today).

jakobbraun:
cool! Thanks a lot for your quick replies!

PR #11025, opened 2024-06-14
Add OpenAPI schema for public REST API

viren-nadkarni:
This PR introduces the OpenAPI schema for the internal endpoints in LocalStack.

Contributions:
- A YAML file with (in `localstack-core/localstack/openapi`) that contains the entire spec for the endpoints in community. The specs have been handcrafted by @viren-nadkarni 🙌 .
- We added a step that validates the spec in the `lint` target. The same check is added also to the pre-commit hooks.
- We added a validator in the handler that validates the requests to the internal endpoints according to the OpenAPI spec. This step can be disabled with the `OPENAPI_VALIDATE_REQUEST` variable. `OPENAPI_VALIDATE_RESPONSE` activates the same check for the responses. It's currently activated by default in our CI pipeline.
- A few unit tests.


Makefile
```diff
@@ -119,7 +119,7 @@ test-coverage: test	  ## Run automated tests and create coverage report
 lint:              		  ## Run code linter to check code style, check if formatter would make changes and check if dependency pins need to be updated
 	($(VENV_RUN); python -m ruff check --output-format=full . && python -m ruff format --check .)
 	$(VENV_RUN); pre-commit run check-pinned-deps-for-needed-upgrade --files pyproject.toml # run pre-commit hook manually here to ensure that this check runs in CI as well
-
+	$(VENV_RUN); python -c 'import localstack.spec, json; print(json.dumps(localstack.spec.OPENAPI))' | openapi-spec-validator -
```
alexrashed, line 122: Great to see that the validity of the spec is checked on every commit. In order to make sure that a developer gets early feedback when breaking the spec, I think it would make sense to also add the precommit hook (by adding a config to `.pre-commit-config`).
There actually is an official pre-commit hook for the `openapi-spec-validator`: https://github.com/python-openapi/openapi-spec-validator/blob/master/.pre-commit-hooks.yaml

requirements-runtime.txt
alexrashed, line 1: There are a few dependencies being added here to the runtime (which is where we need to be cautious not to add too much as it directly increases the Docker image size). Do you have an idea on how big these additional dependencies are?

localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -106,7 +106,7 @@ def get_machine_id() -> str:
     return doc["machine_id"]
 
 
-def get_localstack_edition() -> Optional[str]:
```
alexrashed, line 109: q: Why is this change here necessary? This isn't only a type hint change, but actually modifies what we return (which could have impact on clients using the API).

localstack-core/localstack/spec.py
alexrashed, line 1: Wow! This is really a great push to create a specification for all the LocalStack endpoints!
But I don't really see a reason to have this in a `spec.py` module. The spec is currently really only used in the `openapi.py` module, which is why I would propose to:
- Have the spec in a proper JSON or YAML file.
  - This allows IDEs to properly highlight the syntax and there are lots of IDE plugins around assisting with writing OpenAPI specs. [[1]](https://plugins.jetbrains.com/plugin/14837-openapi-swagger-editor) [[2]](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi)
  - This would also allow you to the `openapi-spec-validator` linting "hack".
  - I personally prefer the YAML format because it allows comments, but that's subjective and not really important.
- Remove the `spec.py` module and load the file where it's needed (in `openapi.py`). If this takes too long, then pre-load it in a runtime hook.

localstack-core/localstack/services/cloudwatch/provider.py
```diff
@@ -364,6 +364,8 @@ def get_raw_metrics(self, request: Request):
                 "v": r.value,
                 "t": r.timestamp,
                 "d": [{"n": d.name, "v": d.value} for d in r.dimensions],
+                "account": account_id,
+                "region": region,
```
alexrashed, line 368: q: How is this change related to the OpenAPI specs?

localstack-core/localstack/aws/handlers/openapi.py
```diff
@@ -0,0 +1,71 @@
...
+from localstack.aws.chain import Handler, HandlerChain
+from localstack.constants import INTERNAL_RESOURCE_PATH
+from localstack.http import Response
+
+LOG = logging.getLogger(__name__)
+
+
+class OpenAPIRequestValidator(Handler):
+    """
+    Validates the internal requests with the OpenAPI spec.
```
alexrashed, line 25: It would be great if we could expand this comment a little bit, as it's pretty short and not 100% clear:
- "internal request" seems to be a bit of an overloaded term. I would think of a request from LocalStack to LocalStack as an internal request (see `InternalRequestParamsEnricher`).
- This module here really should explain why it is necessary.
  - We can globally validate incoming requests, simplifying the implementations of the endpoints.
  - We can use this in testing in order to reconcile that the spec is in sync with the endpoint implementations (I miss static strong typing sometimes 😅).
  - ???

localstack-core/localstack/aws/handlers/openapi.py
```diff
@@ -0,0 +1,71 @@
...
+                # Note: we only check request body validation errors, like invalid body requests or missing required
+                #   parameters. Other things we do currently explicitly check are ServerNotFound, OperationNotFound,
+                #   and PathNotFound. They are most likely intercepted in the handler chain.
+                match e:
+                    case RequestValidationError():
+                        response.status_code = 400
+                        response.set_json({"error": "Bad Request", "message": str(e)})
+                        chain.stop()
+                    case _:
+                        LOG.debug("Uncaught exception: (%s): %s", e.__class__.__name__, str(e))
```
alexrashed, line 50: This seems a bit unusual. I think you should rather use explicit except statements rather than a match case on the exception.

alexrashed, line 50: What other kind of exceptions can be raised here? Should we really only log them on debug here and completely suppress them?

localstack-core/localstack/aws/handlers/openapi.py
```diff
@@ -0,0 +1,71 @@
...
+
+        path = context.request.path
+
+        if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):
+            try:
+                self.openapi.validate_request(WerkzeugOpenAPIRequest(context.request))
+            except OpenAPIError as e:
+                # Note: we only check request body validation errors, like invalid body requests or missing required
+                #   parameters. Other things we do currently explicitly check are ServerNotFound, OperationNotFound,
+                #   and PathNotFound. They are most likely intercepted in the handler chain.
```
alexrashed, line 43: I am a bit confused by this comment:

> "Other things we do currently explicitly check are ServerNotFound, OperationNotFound, and PathNotFound."

I guess that should mean "Other things that we currently explicitly do _not_ check"?

> They are most likely intercepted in the handler chain.

I am a bit confused by this sentence. What does it mean / tell us?

localstack-core/localstack/aws/handlers/openapi.py
```diff
@@ -0,0 +1,71 @@
...
+        path = context.request.path
+
+        if path.startswith(INTERNAL_RESOURCE_PATH) or path.startswith("/_aws/"):
+            try:
+                self.openapi.validate_response(
+                    WerkzeugOpenAPIRequest(context.request),
+                    WerkzeugOpenAPIResponse(response),
+                )
+            except OpenAPIError as exc:
+                LOG.debug(exc)
```
alexrashed, line 68: I don't think this log should be on debug but on error (you are actually going to reject the response here), and should be a bit more concrete.

localstack-core/localstack/aws/handlers/openapi.py
```diff
@@ -0,0 +1,71 @@
...
+                response.status_code = 400
```
alexrashed, line 69: A status code 400 would mean that this is a client-side error, which for the responses it definitely is not, right? Imho, this would be a server-side error / `5xx`...

localstack-core/localstack/spec.py
alexrashed, line 1: Just some food for thought: In the future, it will also be interesting how we are going to deal with our different namespace packages!
- We will have multiple packages in this repo (`localstack.core`, `localstack.aws`, `locasltack.http`,...).
- A lot of different packages could expose LocalStack endpoints (either additional functionality for services, or LocalStack specific features).
- In the best case, the spec is as close as possible to the code (in the absolute best-case it's generated from the docstring or even the type hints).

localstack-core/localstack/config.py
```diff
@@ -742,6 +742,11 @@ def populate_edge_configuration(
 # Whether to enable the partition adjustment listener (in order to support other partitions that the default)
 ARN_PARTITION_REWRITING = is_env_true("ARN_PARTITION_REWRITING")
 
+# Feature flag to enable validation of internal endpoint responses in the handler chain. For test use only.
+OPENAPI_VALIDATE_RESPONSE = is_env_true("OPENAPI_VALIDATE_RESPONSE")
+# Flag to enable the validation of the requests made to the LocalStack internal endpoints. Active by default.
+OPENAPI_VALIDATE_REQUEST = is_env_not_false("OPENAPI_VALIDATE_REQUEST")
```
alexrashed, line 748: q: Should this config maybe be enabled by default in our test pipeline?

tests/unit/aws/handlers/openapi.py
```diff
@@ -0,0 +1,88 @@
...
+
+class TestOpenAPIRequestValidator:
+    def test_valid_request(self):
+        chain = HandlerChain([OpenAPIRequestValidator()])
+        context = RequestContext(
+            Request(
+                path="/_localstack/diagnose",
+                method="GET",
+                scheme="http",
+                headers={"Host": "localhost.localstack.cloud:4566"},
```
alexrashed, line 19: Every successful test has a clear host header here (always using `localhost.localstack.cloud:4566`).
Could you maybe add a test verifying that this is not a necessity? The requests need to work just as fine if the `Host` header contains `localhost:4566` or any other value.

tests/unit/aws/handlers/openapi.py
alexrashed, line 1: Great to see some tests for the handler itself! But in order to ensure that the spec is in sync with the actual implementation, we would basically have to have tests for each and every endpoint covered in the spec.
Do we already have tests like this? How could we bootstrap / push these kinds of tests?

alexrashed (commented):
This is a great first iteration, finally introducing a machine-readable spec for all endpoints which are not clearly defined by the AWS APIs! 🚀 🤖 
I added quite a few comments, lots of them are just questions (for this and later iterations), but also a few suggestions for changes. Let me know if I can help anywhere!

localstack-core/localstack/services/cloudwatch/provider.py
viren-nadkarni: This makes the older CloudWatch provider return the same API response as the v2. I agree this can go in a separate PR. I will raise is separately.

localstack-core/localstack/utils/analytics/metadata.py
viren-nadkarni: This was another inconsistency discovered when drafting the spec. This function is used by the `/health`, `/info` and `/diagnostics` to return the LS edition. I changed this to consistently return a string value to avoid a special `null` case. True, there is a possibility of this affecting the client in which case it should be reverted.

localstack-core/localstack/services/cloudwatch/provider.py
steffyP: Honestly, I am not sure this is worth the effort. The endpoint is currently not even documented, it was introduced a long time ago as a [customer requested it](https://github.com/localstack/localstack/issues/3508). But back then this endpoint could be used by simply using the browser. 

With the requirement change of having a valid request-header where we could extract the account and region, IMO this endpoint become unusable (I think this was introduced with this [PR](https://github.com/localstack/localstack/pull/5478)). 

For v2 we removed this requirement, e.g. it's possible to just browse the page without any request headers. This is now feasible as we store all the data in sqlite and simply dump the entire database (also returning the account and region fields).

tests/unit/aws/handlers/openapi.py
giograno: We do not explicitly test the endpoints. However, they are tested serendipitously in multiple integration tests.
I agree that over time we should be more systematic (involving the service owners at least for the `_aws` endpoints).

localstack-core/localstack/config.py
giograno: Shouldn't be the case already with `is_env_not_false`?

alexrashed: Sorry, that's true, jumped a few lines here. I meant `OPENAPI_VALIDATE_RESPONSE`.

requirements-runtime.txt
giograno: The biggest by far is `openapi-core`, that's about 1.5M

localstack-core/localstack/aws/handlers/openapi.py
giograno: In this first step, I only considered request validation errors (e.g., a missing required in the body).
Other errors that are commonly raised here are `ServerNotFound` (which we shouldn't consider) or `PathNotFound`. I extended the comments a bit to make it more clear.

giograno:
Thanks for the comments @alexrashed 🙏 
I tried to address them one by one. Most notably:
- I extracted the spec into a YAML file and used `setuptools` [data files support](https://setuptools.pypa.io/en/latest/userguide/datafiles.html) to distribute the file with the package;
- I added a pre-commit hook for the YAML spec;
- Simplified exception handling; added comments and clarifications.

requirements-runtime.txt
alexrashed: Okay, but how big are all additional dependencies together?

localstack-core/localstack/spec/openapi.yaml
alexrashed, line 1: nit: I think for now we could just put it in `localstack-core/localstack`, as this is a bit of a generic pattern across all future packages, right?

localstack-core/localstack/utils/analytics/metadata.py
alexrashed: Well, I think since this can't really happen for anyone starting localstack with the official containers, we can keep it as it is now, unifying the return values for all endpoints and simplifying the spec. Thanks for the follow-up, @viren-nadkarni!

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -0,0 +1,74 @@
...
+                response.set_json({"error": "Bad Request", "message": str(e)})
+                chain.stop()
+            except OpenAPIError as e:
+                # Other errors can be raised when validating a request against the OpenAPI specification.
+                #   The most common are: ServerNotFound, OperationNotFound, or PathNotFound.
+                #   We explicitly do not check any other error but RequestValidationError ones.
+                LOG.debug("OpenAPI validation exception: (%s): %s", e.__class__.__name__, str(e))
+
+
+class OpenAPIResponseValidator(OpenAPIRequestValidator):
```
alexrashed, line 56: nit: This inheritance structure breaks the Liskov substitution principle (the L in SOLID) as a `ResponseValidator` is not a `RequestValidator`. If you want to share code between the two, please create a common super-class (something like `OpenAPIValidator`).

alexrashed (commented):
Awesome! This is really taking great shape, and it's nearly ready to be merged, I would just want to quickly discuss the open questions / comments (including the dependency size).
Maybe someone else also wants to take a quick look? :)

localstack-core/localstack/spec/openapi.yaml
giograno: This is what I wanted to do as well, but I was finding issues with `importlib`, namespace packages, and editable installs (this is why I went for a `spec` package).
This [issue is tracked](https://github.com/python/importlib_resources/issues/311) and should be solved soon. In the meantime, I just used a relative import.

requirements-runtime.txt
alexrashed: @giograno Any update on this one? I guess the new dependencies aren't too big, but I would still want to avoid any surprises :P

giograno: @alexrashed yes I built the image locally to have a comparison. Overall, the new image is ~10MB bigger, while a comparison after `docker save | gzip` shows that the new compressed image is <1MB larger.
I would say agree saying that the new dependencies are not too big :)

alexrashed: Perfect, thanks! 💯

alexrashed (approved):
Thanks for addressing the comments. The iteration is lookin' great! We now have an opt-out request validator, an opt-in response validator (which is enabled in the tests), and most importantly a package-root openapi spec for all the already existing non-AWS-operation endpoints! 💯 
Thanks a lot, @viren-nadkarni and @giograno! This is a great step towards a well defined API used by all kinds of SDKs / tools!

PR #11031, opened 2024-06-17
Upgraded Lambda Runtimes to use al2023 Base Image

Morijarti:
## Motivation
Update all Lambda runtimes (where possible) to profit from the slimmer al2023 image and keep our test suite updated (existing tests are often used as samples, so outdated runtimes in tests spread). This requires updating the associated snapshots.

Latest Lambda runtimes: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

## Changes
- Update older runtimes used in code or in default values to latest version that is using al2023 as operating system.
- Update and regenerate snapshot tests where required

## TODO
- [ ] `tests/unit/services/lambda_/test_image_resolver.py` update image mapping
- [ ] `tests/unit/services/lambda_/test_lambda_utils.py`
- [ ] Check for string usage, instead of just for Runtime attribute usage
- [x] Remove `tests/aws/services/lambda_/functions/lambda_s3_integration_sdk_v2.js` once `tests.aws.services.s3.test_s3.TestS3PresignedUrl.test_presigned_url_v4_signed_headers_in_qs` has been fully migrated


tests/aws/services/s3/test_s3.py
```diff
@@ -7603,7 +7603,7 @@ def test_presigned_url_v4_signed_headers_in_qs(
         create_lambda_function(
             func_name=function_name,
             zip_file=testutil.create_zip_file(temp_folder, get_content=True),
-            runtime=Runtime.nodejs16_x,
+            runtime=Runtime.nodejs20_x,
```
Morijarti, line 7606: As mentioned in comments
`we need the AWS SDK v2, and Node 16 still has it by default`
However since AWS has deprecated Node16 we should make this work on Node20

tests/aws/services/lambda_/functions/lambda_s3_integration_sdk_v3.mjs
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+        s3 = new S3Client({ signatureVersion: 'v4' })
+    }
+
+    const url = await getSignedUrl(
+        s3,
+        new PutObjectCommand({
+            Bucket: BUCKET_NAME,
+            Key: 'key-for-signed-headers-in-qs',
+            ServerSideEncryption: 'AES256', // Adds 'X-Amz-Server-Side-Encryption'
```
Morijarti, line 30: For some reason it refuses to accept setting of `ServerSideEncryption` 
Documentation
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/PutObjectCommand/

tests/aws/services/s3/test_s3.py
```diff
@@ -7569,6 +7569,7 @@ def test_presigned_url_v4_x_amz_in_qs(
         head_object = aws_client.s3.head_object(Bucket=function_name, Key=object_key)
         snapshot.match("head-object", head_object)
 
+    @pytest.mark.skip(reason="Current test needs to be rewritten to work with AWS SDK v3")
```
Morijarti, line 7572: Temporary added skip so I can more easily focus on other failing tests in pipeline.
This skip needs to be removed before merge.

Morijarti: Hi @bentsku 
Could you take a look at this test?
We need to update runtime to nodejs20 as node16 is getting deprecated and nodejs20 supports aws sdk v3.
I tried updating it, but sadly my JavaScript skills are lacking.

I think this should be of help
https://github.com/aws/aws-sdk-js-v3/issues/1576

Thank you

bentsku: Hmm, this is a bit tricky, because I believe this test is actively testing against the AWS JS SDK v2, if we use v3 then the test is not useful anymore. 

We would need to install the AWS JS SDK v2 inside the lambda in order for the test to stay relevant. Worst case, you can leave it skipped and I'll tackle it in a follow up. But the lambda using the AWS JS SDK v2 should stay as is for now.

Morijarti: Thanks for giving it a look.


We are introducing a flag (off by default until next major version, I think). 
Switching it on would prevent starting lambda with this runtime and crash the test, but in theory default CI runs should run without issues until then.

I'm not sure if it would be better to leave test as it is, or mark it as skip?
What do you think?

joe4dev (approved):
Looking good  ✨ Thank you for turbocharging our Lambda tests 🚀  

There's just one remaining test failure in the community integration tests related to a resource that already exists:

```
ERROR ../localstack/tests/aws/scenario/bookstore/test_bookstore.py::TestBookstoreApplication::test_setup - localstack.testing.pytest.fixtures.StackDeployError:
...
BookstoreStack (AWS::CloudFormation::Stack) -> CREATE_FAILED (An error occurred (ResourceAlreadyExistsException) when calling the CreateDomain operation: domain bookstorestack-domain66ac69e already exists in region us-east-1)
```

tests/aws/services/s3/test_s3.py
```diff
@@ -7614,6 +7614,7 @@ def test_presigned_url_v4_signed_headers_in_qs(
 
         function_name = f"func-integration-{short_uid()}"
         # we need the AWS SDK v2, and Node 16 still has it by default
+        # TODO since Node 16 is getting depricated we should consider ugrading to Node 20
```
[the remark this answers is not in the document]
bentsku: Yes, thanks, we had a discussion about it here: https://github.com/localstack/localstack/pull/11031#discussion_r1646594495. I'll update the test to run Node20 but still with the manually AWS v2 SDK, even though deprecated, we will still need to support it for a long while I would say. 😄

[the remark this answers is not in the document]
joe4dev: Makes sense. Thanks for the extra context @bentsku 👍 
We can use the caching infrastructure used by the multiruntime tests for Lambda functions under `functions/common/*` once we need a dedicated build step for bundling an older SDK.

PR #11039, opened 2024-06-17
Add *_DOCKER_FLAGS option to support docker compose-style environment files

dfangl:
## Motivation
Per request, we want to support environment files as supported in docker-compose as well as env-files for the docker cli.

They have a quite different feature set, as docker compose allows stripping quotes, end of line comments, and support referencing other variables in a bash-expansion style syntax.

This PR will add, at least to the docker SDK client used within the container, a flag called `--compose-env-file`, which will support these environment files.
It uses the python `dotenv` library for this usecase.

## Changes
* Support for `--compose-env-file`, which will parse the env file similar to docker compose (which uses `godotenv`)

### DIFF
--- a/localstack-core/localstack/utils/container_utils/container_client.py
+++ b/localstack-core/localstack/utils/container_utils/container_client.py
@@ -13,6 +13,8 @@ from enum import Enum, unique
 from pathlib import Path
 from typing import Dict, List, Literal, NamedTuple, Optional, Protocol, Tuple, Union, get_args
 
+import dotenv
+
 from localstack import config
 from localstack.utils.collections import HashableList, ensure_list
 from localstack.utils.files import TMP_FILES, rm_rf, save_file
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/container_utils/container_client.py
```diff
@@ -1193,35 +1241,12 @@ def parse_additional_flags(
...
-                    else:
-                        # No "=" in the line, only the name => lookup in local env
-                        if env_value := os.environ.get(lhs):
-                            env_vars[lhs] = env_value
+                env_vars.update(Util._read_docker_cli_env_file(env_file))
+
+        if args.compose_env_files:
+            env_vars = env_vars if env_vars is not None else {}
+            for env_file in args.compose_env_files:
+                env_vars.update(dotenv.dotenv_values(env_file))
```
dominikschubert, line 1249: We should maybe document the precedence order somewhere. Basically that the --compose-env-file arguments are evaluated after the --env-file arguments and might override them

dominikschubert (approved):
LGTM

localstack-core/localstack/utils/container_utils/container_client.py
dfangl: Agreed! I will try to make it part of the docs for the lambda docker flags!

Issue #11019, opened 2024-06-13
bug: Invalid type for parameter stageDescription


lukasz-mazur2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Localstack complains about stageDescription parameter and is expecting a string value instead of a dictionary. 

Parameter validation failed:
Invalid type for parameter stageDescription, value: {'TracingEnabled': True, 'LoggingLevel': 'ERROR', 'DataTraceEnabled': True, 'MetricsEnabled': True, 'AccessLogSetting': {'DestinationArn': 'arn:aws:logs:eu-west-1:000000000000:log-group:/aws/apigateway/ikvj3dtbni/test:*', 'Format': '{ "xrayTraceId":"$context.xrayTraceId", "requestId":"$context.requestId",  "ip":"$context.identity.sourceIp", "caller":"$context.identity.caller",  "user":"$context.identity.user",  "userCognitoIdentityId":"$context.identity.cognitoIdentityId",  "requestTime":"$context.requestTime", "httpMethod":"$context.httpMethod",  "httpPath":"$context.path", "resourcePath":"$context.resourcePath",  "status":"$context.status", "protocol":"$context.protocol",  "responseLength":"$context.responseLength", "error":"$context.error.message",  "errorWaf":"$context.waf.error", "errorAuthorize":"$context.authorize.error",  "errorAuthorizer":"$context.authorizer.error",  "errorAuthenticate":"$context.authenticate.error",  "errorIntegration":"$context.integration.error" }'}, 'ThrottlingBurstLimit': -1, 'ThrottlingRateLimit': -1}, type: <class 'dict'>, valid types: <class 'str'> Traceback (most recent call last):

Full log:

```
LocalStack version: 3.4.1.dev20240605144018
LocalStack Docker container name: localstack-main
LocalStack Docker container id: da8acb69df1f
LocalStack Docker image sha: sha256:aaff8d61122880e2aa3fc65f53ac2c1945318dbf489000ce0c196c0319ac60fd
LocalStack build date: 2024-06-06
LocalStack build git hash: aa923d1

2024-06-12T10:15:46.021 DEBUG --- [  MainThread] plux.runtime.manager       : instantiating plugin PluginSpec(localstack.hooks.on_infra_start._patch_botocore_json_parser = <function _patch_botocore_json_parser at 0x7fcf6c8d36a0>)
[...]

PR #11040, opened 2024-06-17
fix for ApiGW resource provider

pinzon:
## Motivation
The resource provider for AWS::ApiGateway::Deployment fails to deploy when the StageDescription attribute is define due to a missing conversion. Evidence shown in in #11019

## Changes
- Add conversion from dict to a json string of the attribute in the Create method for the resource.

## Testing
- extended existing deployment test

### DIFF
--- a/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_deployment.py
+++ b/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_deployment.py
@@ -1,6 +1,7 @@
 # LocalStack Resource Provider Scaffolding v2
 from __future__ import annotations
 
+import json
 from pathlib import Path
 from typing import Optional, TypedDict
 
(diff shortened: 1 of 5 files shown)


cloutierMat (commented):
Nice catch :fishing_pole_and_fish:  Thank you @pinzon for solving this issue with apigw! :rocket: 

I appreciate the update to the test to prevent regression, but I feel it would be better to have it in a `aws.validated` test. The template you updated is used by `test_url_output` which isn't validated and makes assertion on the outptut Cfn.

I believe we have other validated tests that are more related to deployment and stages that could be used to increase transparency and include a snapshot.

pinzon:
@cloutierMat  Thanks for the comment. I just commited the suggested change.

Morijarti (approved):
Looks good.
Thank you for updating test as well.

PR #11043, opened 2024-06-18
apigw create python exceptions for gateway response

cloutierMat:
## Motivation

Base python exception to handle gateway integration error.

## Changes

Each exception is subclassed from one of the Default response. It will allow us to quickly check if `error.default_type` is a configured `GatewayResponse` by the user when building the response in the exception handler


bentsku (approved):
LGTM! This looks pretty neat 😄 can't wait to see it in action in the exception handler! this is gonna be neat 👌

PR #11047, opened 2024-06-18
Feature: EventBridge v2: Add test put events limit to 10 messages

maxhoheiser:
## Motivation
`put_events` is the main method to directly put events to an event bus, it takes as input argument a list of events containing `source` `detail-type` `detail` and `bus-name` but it is limited to a maximum of 10 events at a time.

In addition, AWS introduced some changes to default parameters for event buses, in particular creation time and update time, these changes are also incorporated here to get updated snapshot tests passing.

## Changes
Add error handling for `put_events` number of entries exceeding 10
Pars entries and return correctly parsed events

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -577,6 +577,12 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         endpoint_id: EndpointId = None,
         **kwargs,
     ) -> PutEventsResponse:
+        if len(entries) > 10:
+            formatted_entries = [self._event_to_error_type_event(entry) for entry in entries]
+            formatted_entries = f"[{', '.join(formatted_entries)}]"
+            raise ValidationException(
+                f"1 validation error detected: Value '{formatted_entries}' at 'entries' failed to satisfy constraint: Member must have length less than or equal to 10"
+            )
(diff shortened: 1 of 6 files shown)


Morijarti (approved):
Looks really straightforward.
Good work

localstack-core/localstack/services/events/provider.py
```diff
@@ -754,6 +760,26 @@ def _event_bus_dict_to_api_type_event_bus(self, event_bus: EventBus) -> ApiTypeE
 
         return event_bus_api_type
 
+    def _event_to_error_type_event(self, entry: PutEventsRequestEntry) -> str:
+        detail = (
+            json.dumps(json.loads(entry["Detail"]), separators=(", ", ": "))
```
joe4dev, line 765: nit: assigning the loaded json to a variable could improve readability and makes debugging easier (e.g., if parsing fails)

joe4dev (approved):
LGTM, nice catch 👍

localstack-core/localstack/services/events/provider.py
maxhoheiser: I would keep it as a list comprehension since this just adds a space to the separators if detail exists since aws has a strange response for errors in this particular case

PR #11051, opened 2024-06-18
add parsing and routing to APIGW next gen

bentsku:
## Motivation
This PR introduces the parsing of the request via a handler to populate a context with what I think is necessary right now. 

It also introduces the routing handler, which will take the incoming request and match it against the deployment, to get the right `Resource` and `Method` to attach it to the context. Those will be used by all the next handlers, it will allow us to know what to do against the incoming request (method request parameters, integration request, integration itself). 

## Changes
- update the parsing handler, with what I think is necessary, but we can add as we go
- add the routing handler, copy-pasting the current routing code (it is quite complex) and adapt it to our use

## TODO

What's left to do:

- [ ] add tests
- [ ] ...

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -1,9 +1,36 @@
-from typing import Optional
+from http import HTTPMethod
+from typing import Optional, TypedDict
 
 from rolo import Request
 from rolo.gateway import RequestContext
+from werkzeug.datastructures import Headers
 
+from localstack.aws.api.apigateway import Method, Resource
 from localstack.services.apigateway.models import RestApiDeployment
+
+
+class InvocationRequest(TypedDict, total=False):
+    http_method: Optional[HTTPMethod]
+    """HTTP Method of the incoming request"""
+    raw_uri_path: Optional[str]
+    """RAW URI path of the incoming request, needed to keep double forward slashes"""
+    raw_path: Optional[str]
+    """Raw path of the request with no URL decoding"""
+    path_parameters: Optional[dict[str, str]]
+    """Path parameters of the request"""
+    query_string_parameters: Optional[dict[str, str]]
+    """Query string parameters of the request"""
+    # TODO: need to check if we need the raw headers (as it's practical for casing reasons)
+    raw_headers: Optional[Headers]
+    """Raw headers using the Headers datastructure which allows access with no regards to casing"""
+    headers: Optional[dict[str, str]]
+    """Headers of the request"""
+    multi_value_query_string_parameters: Optional[dict[str, list[str]]]
+    """Multi value query string parameters of the request"""
+    multi_value_headers: Optional[dict[str, list[str]]]
+    """Multi value headers of the request"""
+    body: Optional[bytes]
+    """Body content of the request"""
 
 
 class RestApiInvocationContext(RequestContext):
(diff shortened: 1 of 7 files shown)


tests/unit/services/apigateway/test_handler_request.py
```diff
@@ -0,0 +1,279 @@
...
+class TestRoutingHandler:
+    @pytest.fixture
+    def deployment_with_routes(self, dummy_deployment):
+        """
+        This can be represented by the following routes:
+        - /
+        - GET /test
+        - PUT /test/{param}
+        - DELETE /{proxy+}
+        """
```
cloutierMat, line 136: Nice and clean! The added comment helps a lot

localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
```diff
@@ -1,22 +1,67 @@
...
+    # TODO: need to check if we need the raw headers (as it's practical for casing reasons)
+    raw_headers: Optional[Headers]
+    """Raw headers using the Headers datastructure which allows access with no regards to casing"""
+    headers: Optional[dict[str, str]]
+    """Headers of the request"""
+    multi_value_query_string_parameters: Optional[dict[str, list[str]]]
+    """Multi value query string parameters of the request"""
+    multi_value_headers: Optional[dict[str, list[str]]]
+    """Multi value headers of the request"""
+    body: Optional[bytes]
```
cloutierMat, line 32: Would it be useful to have `raw_body` and `decoded_body`, or similar?

cloutierMat (commented):
:rocket: 
 The amount of work you put into this really shows!

I have left one question regarding the body in the `InvocationRequest`

localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
bentsku: Not sure I understand, how do you mean?

Currently, we use `restore_payload` for the body, because the request parsing of Werkzeug might consume the data to create a parsed form, and then we would not be able to forward/read the raw body. 

It splits in 3 parts: `data`, `form` and `files`. My belief is that APIGW does not manipulate different parts of the body directly? But I'm not sure. 

It's also possible I'm not familiar with what is possible to do with APIGW 😅

cloutierMat: :thinking: I think the only thing we can do on the body is through the request templates. I haven't looked fully into the integration, so maybe this is the right way to go. I was just concerned about decoding it in with LS parser to restore it, to decode it again to apply the templates.

bentsku: Ah, I see. 
I think the issue here is what `restore_payload` does. We don't really "decode", we "re-encode" it to be like what we actually received from the client. 

For example, the usual way is to call `request.data` to get the `bytes` content of the body. 

But let's imagine your request is a `Content-Type: application/x-www-form-urlencoded`.
You get an HTTP body looking like this: `field1=value1&field2=value2`

When accessing `request.data`, you'll get... `b""`, an empty `bytes` object because Werkzeug will have parsed the request and consumed the stream, and set `request.form` to be equal to an `ImmutableMultiDict` looking like `{"field1": "value1", "field2": "value2"}`.

One of the way here is to directly consume the `request.stream` once to get the raw data out. But anything earlier in the handler chain called `request.data`, the stream will have be consumed and you have no way to get it back (yet! with the new `twisted` web server, we actually have a way now 🥳 which means we can have a short circuit `restore_payload`), which is why we re-encode the body 😕 

So in itself I'd say we have the proper `raw_body` now, the one the request template should act upon?

cloutierMat: I guess it makes sense then since only one handler should need to perform operations on the body, that step can ensure the body returns to bytes once it has completed and and shouldn't need to keep a dict representation of it anywhere. :+1:

cloutierMat (approved):

alexrashed (approved):
This is really a great push forward for the Next Gen API GW, providing a solid foundation re-using our internal frameworks 😍 
Especially in combination with #11057 (where basically all the parsing / routing is then replaced to directly done by Werkzeug! 🥳

> If needed, I can already merge the 2 PRs in one if that makes it easier 😄

Having two separate PRs helped to really review the different iterations (where each are quite big). Now, after the individual reviews, merging one into the other should work perfectly fine. :)

Issue #11037, opened 2024-06-17
bug: CopyObject implementation breaks object when clearing out tags.


mathieuk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After performing a S3 CopyObject on an object with a tag and `TaggingDirective=REPLACE` and `Tagging=`, subsequent calls to S3 GetObjectTagging for that object fail with an internal error:

```
2024-06-17 16:09:30 2024-06-17T14:09:30.761 ERROR --- [   asgi_gw_2] l.aws.handlers.logging     : exception during call chain
2024-06-17 16:09:30 Traceback (most recent call last):
2024-06-17 16:09:30   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
2024-06-17 16:09:30     handler(self, self.context, response)
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/aws/handlers/service.py", line 113, in __call__
2024-06-17 16:09:30     handler(chain, context, response)
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/aws/handlers/service.py", line 83, in __call__
2024-06-17 16:09:30     skeleton_response = self.skeleton.invoke(context)
2024-06-17 16:09:30                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
2024-06-17 16:09:30     return self.dispatch_request(serializer, context, instance)
2024-06-17 16:09:30            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/aws/skeleton.py", line 168, in dispatch_request
2024-06-17 16:09:30     result = handler(context, instance) or {}
2024-06-17 16:09:30              ^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
2024-06-17 16:09:30     return self.fn(*args, **kwargs)
2024-06-17 16:09:30            ^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-17 16:09:30   File "/opt/code/localstack/localstack/services/s3/v3/provider.py", line 2643, in get_object_tagging
[...]

PR #11056, opened 2024-06-19
fix TaggingDirective REPLACE for CopyObject

bentsku:
## Motivation
As reported in #11037, we've had an issue with empty `Tagging` when using the `TaggingDirective` `REPLACE` with `CopyObject`. This was because we would set the result of `if tagging := request.get("Tagging"):` to the tags, which would be `None`. The fix would be to conditionally set the data, but in that case we would need to pop the previous value, or add `tagging or {}` which will set a dict if the value is `None`. 

## Changes
- add a regression test for this particular use case
- add the fix in the provider

_fixes #11037_


cloutierMat (approved):
:rocket: LGTM

And a nice simple test to boot!

Issue #10981, opened 2024-06-07
Problem resolving Env aws variables when deploying lambda CloudFormation


danielferayte:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello, i would like to mention a problem that i found when trying to deploy a lamba by Cloudformation:
i do have this template:
```
LambdaFunctionProcesoEnvio:
    Type: 'AWS::Lambda::Function'
    Properties:
      Timeout: 900
      Code:
        S3Bucket: mybucket
        S3Key: code.zip  
      FunctionName: proccesEnvio
      Runtime: nodejs20.x
      Handler: "src/claims/app.handler"
      Role: !GetAtt LambdaExecutionRole.Arn
      Description: Lambda claims
      Environment:
        Variables:
          DATABASE_SECRET_NAME: "co-claim-integration-layer-oracle-db-local"
          AUTHENTICATION_API_PATH: 'https://hlqwwkrg4g.execute-api.us-east-1.amazonaws.com/desa/api/autenticar'
      MemorySize: 128
```
The problem is when i do deploy this stack in localstack with aws --endpoint-url=http://localhost:4566 cloudformation deploy --stack-name myStack--template-file "./template.yml” and i see the lamba configuration with aws lambda get-function-configuration --function-name proccesEnvio --endpoint-url=http://localhost:4566
i do get this env values:

```
"Environment": {
    "Variables": {
      "DATABASE_SECRET_NAME": "co-claim-integration-layer-oracle-db-local"  
      "AUTHENTICATION_API_PATH": "https://hlqwwkrg4g.execu.amazonaws.com:4566/desa/api/autenticar",
  }
```
For some reason is changing the env value of AUTHENTICATION_API_PATH to the local, but that is not the idea since i need to use the external api. To resolve this i do tried to add the url to the DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM in order to ignore this dns but it does not work, i have tried different patterns but the result is the same
```
localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack
    ports:
      - "4566:4566"            # LocalStack Gateway
      - "4510-4559:4510-4559"  # External services port range
    environment:
      - DEBUG=${DEBUG:-0}
      - MAIN_DOCKER_NETWORK=salvatagio
[...]

PR #11060, opened 2024-06-19
Add CFN_STRING_REPLACEMENT_DENY_LIST config to allow the user to prevent Localstack from changing AWS URLs to point to itself

marcciosilva:
## Motivation

This aims to resolve issue #10981 by giving users the possibility of specifying which AWS URLs they don't want Localstack to change when processing CFN templates.
This situation can be entirely avoided by the user by using non-AWS custom domains for URLs they don't want replaced / pointed to Localstack, so this might be useful but not 100% necessary.

## Changes

It adds a configuration environment variable so that the user can provide a comma separated list of AWS URLs they don't want changed when their CFN templates get processed.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
@@ -9,7 +9,6 @@ from typing import Optional
 
 from localstack import config
 from localstack.aws.connect import connect_to
-from localstack.config import CFN_STRING_REPLACEMENT_DENY_LIST
 from localstack.constants import INTERNAL_AWS_SECRET_ACCESS_KEY
 from localstack.services.cloudformation.deployment_utils import (
     PLACEHOLDER_AWS_NO_VALUE,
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -9,6 +9,7 @@
 
 from localstack import config
 from localstack.aws.connect import connect_to
+from localstack.config import CFN_STRING_REPLACEMENT_DENY_LIST
```
dfangl, line 12: If you import it directly, the monkeypatch in the test cases will not work anymore, as it is now a module attribute of the template deployer module.

To support the monkeypatching, you need to use it with `config.CFN_STRING_REPLACEMENT_DENY_LIST`, this way it will always reference the current `config` value. That is why the test fails, but presumably did not locally, as the `template_deployer` module was likely already imported in the CI run :)

dfangl: Yes, you are basically patching the module level variable in config - if you go the indirection via config.VAR, it will always resolve to the current value in that module, which was changed with monkey patch. You could also assign directly, but the monkey patch fixture takes care of restoring the value after the test, which is convenient.

pinzon (approved):
LGTM

PR #11061, opened 2024-06-19
generalize tracing for gateway and handler chain

thrau:
## Motivation

More efforts towards #10946. I wanted to debug something in a different emulator using `DEBUG_HANDLER_CHAIN`, but couldn't re-use it because of the coupling to the AWS request context.

This PR generalizes the `TracingHandlerChain` so it can be re-used outside AWS. The biggest change was just to create a loop over `context.__dict__` rather than hard-coding each attribute.
I also removed the `Handler` and `ExceptionHandler` from the list of implemented classes, since in rolo they are Protocols, and you shouldn't inherit from protocol classes unless you are creating other protocols.

## Changes

* The `TracingHandlerChain` is now generalized and free of AWS concepts
* The `TracingHandlerChain` now lives in `localstack.http` instead of `localstack.aws`

## Testing

* check out the PR and run ls with `DEBUG_HANDLER_CHAIN=1`. observe it still works when making aws calls

### DIFF
--- a/localstack-core/localstack/http/trace.py
+++ b/localstack-core/localstack/http/trace.py
@@ -4,13 +4,11 @@ import logging
 import time
 from typing import Any, Callable
 
+from rolo import Response
+from rolo.gateway import ExceptionHandler, Handler, HandlerChain, RequestContext
 from werkzeug.datastructures import Headers
 
-from localstack.http import Response
 from localstack.utils.patch import Patch, Patches
-
-from .api import RequestContext
-from .chain import ExceptionHandler, Handler, HandlerChain
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/http/trace.py
```diff
@@ -181,19 +179,18 @@ def do_trace_call(
...
-                actions.append(SetAttributeAction("service_response"))
+            # detect when attributes are set in the request contex
+            context_args = dict(context.__dict__)
+            context_args.pop("request", None)  # request is handled separately
+
+            for k, v in context_args.items():
+                if not v:
+                    continue
+                if prev_context.get(k):
+                    # TODO: we could introduce "ModifyAttributeAction(k,v)" with an additional check
```
bentsku, line 190: this would be extremely neat actually

bentsku (approved):
Wow, really cool!! LGTM! This is really neat 😄 and looking at the APIGW Next-gen project 👀 this could see nice usage already, to debug what's wrong and what we could be missing. Super neat, and nice trick to decouple from AWS... another step towards it! 🚀 

I've tried it locally and it still works 💯

PR #11065, opened 2024-06-20
split routing utilities not related to service operation routing out of op_router.py

bentsku:
## Motivation
Follow up from #11057 and this comment: https://github.com/localstack/localstack/pull/11057#pullrequestreview-2130154194

I've pulled everything not directly related to service-operation-routing out of `op_router`, in order to be able to use them in API Gateway resource routing as well. 

## Changes
- pull some utilities outside of `op_router.py` into `routing.py` and update the imports


alexrashed (approved):
Nice cleanup! 🧹
Nice to see that these few utility functions for custom routing are also useful for the new APIGW routing!

PR #11068, opened 2024-06-21
improve CFN APIGW Method parity

bentsku:
## Motivation
We got a report in our Slack channel regarding a type cast error for the `selectionPattern` field for `APIGW::Method` resource. 

After taking a look and updating one of our test to validate the fix, I've realized we've had a few parity issues and snapshot skips. I've also fixed them in the meantime 😄 

## Changes
- add type casting of `selectionPattern`
- remove default value of `{}` for multiple fields, which should not be sent at all if they are not set. AWS can accept empty dicts from them, so we have to check against `None` value directly

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_apigateway.py
+++ b/tests/aws/services/cloudformation/resources/test_apigateway.py
@@ -51,7 +51,8 @@ Resources:
 """
 
 
-@markers.aws.unknown
+# this is an `only_localstack` test because it makes use of _custom_id_ tag
+@markers.aws.only_localstack
 def test_cfn_apigateway_aws_integration(deploy_cfn_template, aws_client):
     api_name = f"rest-api-{short_uid()}"
     custom_id = short_uid()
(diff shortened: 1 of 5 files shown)


Morijarti (approved):
Thank for improving provider and tests as well.

Just checked the boto docs and you were right, I was sending optional parameters with empty dictionary value. Good catch.

tests/aws/services/cloudformation/resources/test_apigateway.py
```diff
@@ -267,17 +264,6 @@ def test_cfn_deploy_apigateway_models(deploy_cfn_template, snapshot, aws_client)
...
-    paths=[
-        "$..methodIntegration.integrationResponses",
-        "$..methodIntegration.requestParameters",  # missing {}
-        "$..methodIntegration.requestTemplates",  # missing {}
-        "$..methodResponses",  # missing {}
-        "$..requestModels",  # missing {}
-        "$..requestParameters",  # missing {}
-        "$..rootResourceId",  # shouldn't exist
-    ]
-)
```
cloutierMat, line 280: :partying_face:

PR #11069, opened 2024-06-21
remove THREAD_LOCAL usage

bentsku:
## Motivation
We've been meaning too remove the `THREAD_LOCAL` usage since a long time, and we could finally do so while working on the refactor of APIGW. 

This did not have any use anymore, and it relied on setting additional data in a threadlocal, but thanks to the awesome work of @viren-nadkarni, nothing uses `aws_stack.get_region()` anymore, so we can be sure nothing is accessing/reading those values anymore. 

I really want to emphasis that this is only possible thanks to the relentless work of Viren 😄 

## Changes
- remove `THREAD_LOCAL` and its usages across the codebase

### DIFF
--- a/localstack-core/localstack/services/apigateway/router_asf.py
+++ b/localstack-core/localstack/services/apigateway/router_asf.py
@@ -147,7 +147,7 @@ class ApigatewayRouter:
             strict_slashes=True,
         )
 
-    def invoke_rest_api(self, request: Request, **url_params: Dict[str, str]) -> Response:
+    def invoke_rest_api(self, request: Request, **url_params: str) -> Response:
         account_id, region_name = get_api_account_id_and_region(url_params["api_id"])
         if not region_name:
             return Response(status=404)
--- a/localstack-core/localstack/utils/aws/request_context.py
+++ b/localstack-core/localstack/utils/aws/request_context.py
@@ -4,13 +4,9 @@ This module has utilities relating to creating/parsing AWS requests.
 
 import logging
 import re
-import threading
 from typing import Dict, Optional
 from urllib.parse import urlparse
 
-from flask import request
-from requests.models import Request
-from requests.structures import CaseInsensitiveDict
 from rolo import Request as RoloRequest
 
 from localstack.aws.accounts import get_account_id_from_access_key_id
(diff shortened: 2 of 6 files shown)


thrau (approved):
fantastic cleanup! thanks @bentsku and @viren-nadkarni :100:

localstack-core/localstack/services/apigateway/router_asf.py
```diff
@@ -147,7 +147,7 @@ def register_routes(self) -> None:
             strict_slashes=True,
         )
 
-    def invoke_rest_api(self, request: Request, **url_params: Dict[str, str]) -> Response:
+    def invoke_rest_api(self, request: Request, **url_params: str) -> Response:
```
alexrashed, line 150: Is this really related to this changeset? Looking at the access pattern to `url_params` this can't be a string (and I would expect the params to be a dict).

localstack-core/localstack/utils/aws/request_context.py
```diff
@@ -4,13 +4,9 @@
 
 import logging
 import re
-import threading
 from typing import Dict, Optional
 from urllib.parse import urlparse
 
-from flask import request
```
alexrashed, line 11: 🥳 
This is great! Would this allow us now to remove the dependency of flask altogether?
With a quick look I can only find two more usages:
- `aws_responses` where I think that the usage can directly be replaced with `werkzeug`'s `Response` (the superclass of `flask`'s `Response` class)
- Down below in this module in the `patch_moto_request_handling` function, which is maybe not necessary anymore since we don't use the moto server anywhere anymore?

But this should not scope-creep this PR, just an idea for a follow up... 😉

alexrashed (approved):
Awesome! It's really great to see this wonky thread local handling being gone (and it's great that it was basically just removed -> well prepared by  @viren-nadkarni)! 🚀 🦸🏽

localstack-core/localstack/services/apigateway/router_asf.py
bentsku: Not exactly related, but the type warning by PyCharm was driving me a bit crazy, and I've looked into it and it seems for `*args` and `**kwargs` you only type the values and not the variable in itself: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

I feel like I had found a better PEP for this but cannot find it anymore...

alexrashed: Today I Learned! Thanks! Make sense, but isn't really obvious to be honest 😛

bentsku: It really isn't 😅 I've been looking at this warning for more than a year and only looked it now so 😬 it didn't make sense to me either

PR #11070, opened 2024-06-21
fix NG APIGW unit tests usage of RestApiDeployment

bentsku:
## Motivation
Following #11062 and #11051, we've had hidden merge conflicts, the signature of one of the class changed, weirdly the CI run was green even though the other PR had been merge quite much earlier. 

Not exactly sure why it happened, ~maybe some caching regarding unit tests~? 

edit: ah, I think this is because CircleCI checks out the repo on the branch, whereas GitHub Actions adds your PR on top of master, which would have caught this. 

## Changes
- fix the usage of `RestApiDeployment` to match the signature


cloutierMat (approved):
Good catch, I didn't rebase first and the changes from #11051 introduced new required arguments to `RestApiDeployment`! :+1: Thanks for the fix

PR #11076, opened 2024-06-24
Mark acceptance test marker as GA

silv-io:
## Motivation

The acceptance tests are used as proper acceptance tests now in the main pipeline (#11049). However, their marker still indicates a beta status. 

## Changes

This PR removes the `_beta` suffix from the acceptance test marker.


alexrashed (approved):
Awesome! Thanks for jumping on this! 🧹 💯

PR #11078, opened 2024-06-24
parity: kinesis event payload

giograno:
## Motivation
This PR fixes a minor parity issue for AWS Lambda Kinesis Data Streams event source mappings.
As per [docs](https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html), the event payload should have a `kinesisSchemaVersion` key.
Few libraries (e.g., AWS Java SDK) seem to expect this value.

## Changes
- adding `kinesisSchemaVersion` to the event payload;
- removing snapshot skip from our tests;


joe4dev (approved):
Nice catch 👍 
So easy 🙈

PR #11083, opened 2024-06-25
Switch test selection from opt-in to opt-out

silv-io:
## Motivation

After implementing the test selection, using it without any issues for a while, and with the merge of the acceptance tests only (#11049), we cat switch to an Opt-Out approach for the test selection (instead of the current opt-in).

## Changes

Exchanges the current opt-in of the test selection to opt out.


dominikschubert (approved):
LGTM!

Makes sense to keep an explicit opt-out. Might be communicating the purpose clearer than simply adding a rule for each opt-out directly.

silv-io:
Great! I'll merge this once #11049 is done as that marks the final step of having the new setup up and running.

silv-io:
Since, doing the tests for #11049 will take some time due to resource limits, I'll merge this now already.

PR #11085, opened 2024-06-25
refactor Moto out of APIGW next-gen invocation logic

bentsku:
## Motivation
During a pair programmation session with @cloutierMat, he emphasized the need to get decoupled from Moto very early, so that we would not be polluted nor have a lot of tech debt across the new invocation logic. 

We then removed the `MotoRestAPI` usage in our deployment/invocation, and only return the APIGW `Resources` needed for the invocation.

If we realize we need more things from Moto, we will need to pull them out of it in  `reeze_rest_api` and add fields to the `MergedRestApi` class.

Once we tackle the removal and split brain situation of Moto from APIGW, we can remove the `MergedRestApi` class and only use the container from our store. 

This will also help us during testing, as we won't need to setup Moto resources like currently done in `tests/unit/services/apigateway/test_handler_request.py`, which is currently tech debt: it is easier to set up lots of resources for testing routing by using the moto utils than to write a lot of nested dicts. 

## Changes
- refactor the Moto REST API out of the Deployment.
- simplify the signature of the deployment, to have all the resources in only one container
- update our tests to make use of the new signatures

### DIFF
--- a/localstack-core/localstack/services/apigateway/models.py
+++ b/localstack-core/localstack/services/apigateway/models.py
@@ -1,6 +1,5 @@
 from typing import Any, Dict, List
 
-from moto.apigateway.models import RestAPI as MotoRestAPI
 from requests.structures import CaseInsensitiveDict
 
 from localstack.aws.api.apigateway import (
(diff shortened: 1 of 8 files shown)


cloutierMat (approved):
It was a great pair coding session! :rocket: 
I am really happy that we could decouple moto from the handlers. That will make things simpler for now and lighten the upcoming tech debt! :man_dancing: 

I just left one suggestion to fix the failing test. But I believe this is good to go otherwise.

Issue #10860, opened 2024-05-21
Internal Refactoring: Towards a Multi-Distribution Setup


dominikschubert:
Over the next few weeks and months we’re refactoring the code in this repository to move toward a **multi-distribution setup**.

For now this only affects active contributors as well as any developers that depend on code existing in the published container under the path `/opt/code/localstack/localstack`. Most users should not be affected by this change.

## Motivation
This will enable us to define clearer boundaries and allows for easier re-use of individual components.
Some of the previously internal code has already been moved into external open repositories such as [localstack/rolo](https://github.com/localstack/rolo).
Other parts of the codebase will keep living in this repository but under its own distribution.
How we will map this to PyPI items is still being discussed and should become clearer over the next weeks during the initial refactorings.

The code layout is not part of any official API or semver guarantees, nevertheless we still want to use this chance to give you a heads up and some guidance in how to make your existing code compatible with the new structure.

### 1. Moving everything into `localstack-core`

As a first step, the entirety of the localstack module is moved into a `localstack-core` directory with https://github.com/localstack/localstack/pull/10800, which will make up one of the multiple distributions.

In this initial step on our way to a multi-distribution system, only the additional root level of `localstack-core` is introduced and the rest of the directory structure is unchanged.

- If you have an open PR, you can rebase onto master after  https://github.com/localstack/localstack/pull/10800 has been merged.
- If you are an active contributor and you're using the PyCharm IDE, you need to adapt your project structure by marking the new `localstack-core` module as a source folder. Otherwise you will encounter errors where it will complain about not being able to find the `localstack` module.
  - ![Untitled (1)](https://github.com/localstack/localstack/assets/620817/e7d9694f-1e2f-49dd-a272-f1e79df7eefb)

- If you want to call code from the `localstack` module, you now need to perform an installation of the project (e.g. with `pip install -e .`). Previously, since `localstack` was a root-level module, python automatically included it in its import path. With a source directory layout there is a more strict boundary now which also helps avoiding unintentional imports. See [here](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) for more information on the differences between a flat and a src layout.

- The location of test files is unchanged.
- Locally the code moves from `.../localstack/localstack/...`  => `.../localstack/localstack-core/localstack/...`
- In the published container the code moves from `/opt/code/localstack/localstack` => `/opt/code/localstack/localstack-core/localstack`

### ?. Next steps
[...]

PR #11088, opened 2024-06-25
update pro source code location

cloutierMat:
## Motivation

Following source code refactor in -ext initiated by #10860, the command `python -m localstack.dev.run --mount-source-code` was failing to mount -pro source code as the location change. This is a very useful command allowing for running localstack in docker mode with local code mounted. 

## Changes

Point the pro source code to the new code directory.


thrau (approved):
Great catch there! Thanks for fixing 👍

PR #11089, opened 2024-06-25
EC2SecurityGroup: add TagSpecifications parameter when Tags present in CloudFormation template

### LINKED ISSUE #10768: bug: Security Groups created via Cloudformation template do not have the tags specified in the template

### Current Behavior

Deploying a Cloudformation template which specifies a security group with one or more tags successfully creates that security group, but the tags are missing.  Adding the tags to the security group manually after the template is deployed works as expected.

### Expected Behavior

Tags on all resources in a Cloudformation template should be created on those resources automatically as part of the deployment of that template.  I deployed this same template on an actual AWS account and the tags were present as expected.

With the `localstack` script

    localstack --debug start

    # Download the example CFN that reproduces this behavior
    curl https://gist.githubusercontent.com/anelson/4e5af3f511e8429a3b06b1c9d4294f71/raw/eae7bca475642608b694b756e7c5c08244803537/cfn.yaml -o cfn.yaml

    # Deploy the CFN to localstack
    awslocal cloudformation create-stack --stack-name farcast --template-body "$(cat cfn.yaml)"

    # Wait a few seconds for the deploy to complete.

    # Describe the security groups; note that the security group `FarcastDefaultComputeB5C4A6CA` has no tags
    awslocal ec2 describe-security-groups

### Environment

```markdown
- OS: Fedora 39
- LocalStack: 3.4.0
```

### Anything else?

I made a gist with a CFN that reproduces this behavior:

https://gist.github.com/anelson/4e5af3f511e8429a3b06b1c9d4294f71

jcollado:
## Motivation

Add tags to EC2 Security Group resource when specified in CloudFormation template.

Fixes #10768 

## Changes

Pass the `TagSpecifications` parameter to the `create_security_group` method when the `Tags` have been passed to the CloudFormation template.

## Testing

Using this CloudFormation template:
```
Resources:
  SecurityGroup:
    Type: AWS::EC2::SecurityGroup
    Properties:
      GroupDescription: Security Group
      Tags:
        - Key: key1
          Value: value1
        - Key: key2
          Value: value2
```

run:
```
awslocal cloudformation create-stack --stack-name sg --template-body "$(cat cfn.yaml)"
awslocal ec2 describe-security-groups
```

Output should contain `Tags` in the newly created EC2 security group as in the following example:
```
        {
            "Description": "Security Group",
            "GroupName": "sg-SecurityGroup-4e42e960",
            "IpPermissions": [],
            "OwnerId": "000000000000",
            "GroupId": "sg-16639ba90ec152e26",
            "IpPermissionsEgress": [
                {
                    "IpProtocol": "-1",
                    "IpRanges": [
                        {
                            "CidrIp": "0.0.0.0/0"
                        }
                    ],
                    "Ipv6Ranges": [],
                    "PrefixListIds": [],
                    "UserIdGroupPairs": []
                }
            ],
            "Tags": [
                {
                    "Key": "key1",
                    "Value": "value1"
                },
                {
                    "Key": "key2",
                    "Value": "value2"
                }
            ],
            "VpcId": "vpc-e1f7c70b"
        }

```

### DIFF
--- a/docs/testing/integration-tests/README.md
+++ b/docs/testing/integration-tests/README.md
@@ -72,7 +72,7 @@ def test_something_on_multiple_buckets(s3_create_bucket):
   # both buckets will be deleted after the test returns
 ```
 
-You can find the list of available fixtures in the [fixtures.py](https://github.com/localstack/localstack/blob/master/localstack/testing/pytest/fixtures.py).
+You can find the list of available fixtures in the [fixtures.py](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/testing/pytest/fixtures.py) file.
 
 
 ## Running the test suite
@@ -161,8 +161,8 @@ Once you're confident your test is reliably working against AWS you can add the 
 
 #### Create a snapshot test
 
-Once you verified that your test is running against AWS, you can record snapshots for the test run. A snapshot records the response from AWS and can be later on used to compare the response of LocalStack. 
+Once you verified that your test is running against AWS, you can record snapshots for the test run. A snapshot records the response from AWS and can be later on used to compare the response of LocalStack.
 
-Snapshot tests helps to increase the parity with AWS and to raise the confidence in the service implementations. Therefore, snapshot tests are preferred over normal integrations tests. 
+Snapshot tests helps to increase the parity with AWS and to raise the confidence in the service implementations. Therefore, snapshot tests are preferred over normal integrations tests.
 
 Please check our subsequent guide on [Parity Testing](parity-testing.md) for a detailed explanation on how to write AWS validated snapshot tests.
--- a/docs/testing/integration-tests/README.md
+++ b/docs/testing/integration-tests/README.md
@@ -165,4 +165,4 @@ Once you verified that your test is running against AWS, you can record snapshot
 
 Snapshot tests helps to increase the parity with AWS and to raise the confidence in the service implementations. Therefore, snapshot tests are preferred over normal integrations tests.
 
-Please check our subsequent guide on [Parity Testing](parity-testing.md) for a detailed explanation on how to write AWS validated snapshot tests.
+Please check our subsequent guide on [Parity Testing](../parity-testing/README.md) for a detailed explanation on how to write AWS validated snapshot tests.
--- a/tests/aws/services/cloudformation/resource_providers/ec2/test_ec2.py
+++ b/tests/aws/services/cloudformation/resource_providers/ec2/test_ec2.py
@@ -57,6 +57,22 @@ def test_deploy_prefix_list(deploy_cfn_template, aws_client, snapshot):
 
 
 @markers.aws.validated
+def test_deploy_security_group_with_tags(deploy_cfn_template, aws_client):
+    """Create security group in default VPC with tags."""
+    stack = deploy_cfn_template(
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../../../templates/ec2_security_group_with_tags.yml"
+        )
+    )
+
+    response = aws_client.ec2.describe_security_groups(GroupIds=[stack.outputs["SecurityGroupId"]])
+    assert response["SecurityGroups"][0]["Tags"] == [
+        {"Key": "key1", "Value": "value1"},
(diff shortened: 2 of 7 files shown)


jcollado:

pinzon (changes requested):
Great fix. However, to ensure that this change does not cause a regression, we should add an integration test or extend an existing one. I would recommend adding a test in the following file: `tests/aws/services/cloudformation/resource_providers/ec2/test_ec2.py`.

jcollado:
@pinzon Thanks for the pointer to the proper test file. That has been useful.

I've added a new test there that uses the same CloudFormation template that I used to reproduce the error and that I used to verify the same behavior in a real AWS account.

Despite it is not part of the issue, I'd like to highlight that AWS also adds the following tags to the resource:
- `aws:cloudformation:logical-id`
- `aws:cloudformation:stack-id`
- `aws:cloudformation:stack-name`

pinzon (changes requested):
Great job on the test! It just needs a few tweaks. At LocalStack, we're constantly working to improve compatibility with AWS. That's why we use Snapshots for validation – you can find more details about it [here](https://github.com/localstack/localstack/blob/master/docs/testing/parity-testing/README.md).

jcollado:
@pinzon Thanks for the information regarding snapshots. I've updated the test case to use a snapshot.

A few comments:

- There were some broken links in the documentation. I've updated them in this PR. If you prefer me to move those changes to a different PR, let me know.

- In the CloudFormation template I had to use `GetAtt` instead of `Ref` because AWS doesn't behave in the same way as LocalStack. From the documentation:

  > When you pass the logical ID of this resource to the intrinsic Ref function, Ref returns the ID of the security group if you specified the VpcId property. Otherwise, it returns the name of the security group.

  However, LocalStack always returns the ID of the security group. Should this behavior mismatch be reported in an issue?

- As commented above, CloudFormation adds additional tags to resources. For the test case to work, I had to filter them out from the response using custom code. Is there any filter transformer that can do the same job or is the custom code acceptable in this case?

docs/testing/integration-tests/README.md
pinzon, line 1: Thanks!

tests/aws/services/cloudformation/resource_providers/ec2/test_ec2.py
```diff
@@ -56,6 +56,29 @@ def test_deploy_prefix_list(deploy_cfn_template, aws_client, snapshot):
...
+    snapshot.add_transformer(snapshot.transform.key_value("GroupId"))
+    snapshot.add_transformer(snapshot.transform.key_value("GroupName"))
+    snapshot.add_transformer(snapshot.transform.key_value("VpcId"))
+    snapshot.add_transformer(SortingTransformer("Tags", lambda tag: tag["Key"]))
+    response = aws_client.ec2.describe_security_groups(GroupIds=[stack.outputs["SecurityGroupId"]])
+    security_group = response["SecurityGroups"][0]
+
+    security_group["Tags"] = [
+        tag for tag in security_group["Tags"] if not tag["Key"].startswith("aws:cloudformation")
+    ]
```
pinzon, line 77: Nice use of transformers.

The marker `@markers.snapshot.skip_verify` enables the skipping of certain attributes from validation using json paths, but it may not be applicable in this case. It's great that you filtered the tags that the resource provider, Another option could be for the resource provider to add them (even though it may be beyond the current scope of the issue).

tests/aws/templates/ec2_security_group_with_tags.yml
```diff
@@ -0,0 +1,14 @@
...
+      GroupDescription: Security Group
+      Tags:
+        - Key: key1
+          Value: value1
+        - Key: key2
+          Value: value2
+
+Outputs:
+  SecurityGroupId:
+    Value: !GetAtt SecurityGroup.GroupId
```
pinzon, line 14: We should address in a future PR the issue you mention about the Security Group Ref.

pinzon (approved):
Thank you for addressing my comments. I am happy to approve the PR. Let's proceed with merging it once all the necessary checks have passed.

PR #11090, opened 2024-06-26
remove flask

alexrashed:
## Motivation
We don't use the moto server in a long time, and we also don't use flask to serve anything in LocalStack itself anymore for quite a while.
Now that we [finally got rid of the thread local](https://github.com/localstack/localstack/pull/11069) (thanks @bentsku!) access to the flask request context, we can just remove the dependency completely.

## Changes
- Removes any remaining references to `flask`:
  - A weird indirection of creating a `requests` response via a `flask` response (might just have been a leftover from a previous cleanup?)
  - The patching of the moto server (since we don't use it anymore this patch should not be necessary anyways?)
- Updates the dependency pins after removing the dependency definition from the `pyproject.toml`

## Testing
- If all our pipelines are green, we should be good to go 😄 
- I also ran a test in our downstream repo against this branch (Ext IT #9677195344) which was also successful.

### DIFF
--- a/localstack-core/localstack/aws/patches.py
+++ b/localstack-core/localstack/aws/patches.py
@@ -51,8 +51,5 @@ def apply_aws_runtime_patches():
     """
     if find_spec("moto"):
         # only load patches when moto is importable
-        from localstack.utils.aws.request_context import patch_moto_request_handling
-
-        patch_moto_request_handling()
         patch_moto_iam_config()
         patch_moto_instance_tracker_meta()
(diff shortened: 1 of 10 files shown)


bentsku (approved):
LGTM! I can't believe we're getting there! Thank you so much for tackling this, one less dependency! Look how far we've come... 😄

This is all possible with all your and the team hard work with ASF 🚀 thank you so much for making our developer experience so great 💯

PR #11092, opened 2024-06-26
Adapt localstack.dev.kubernetes script to new src layout

dominikschubert:
## Motivation

Recent changes when migrating to a src layout in pro have broken the `localstack.dev.kubernetes` CLI. This PR fixes this by adapting the path to the new nesting structure.

## Changes

- Fixed mounting of pro code into k8s cluster via `localstack.dev.kubernetes` script.
- Fixed mounting of community entrypoints

## Testing
- `python -m localstack.dev.kubernetes --pro` (and check the volume mount paths in the output)

### DIFF
--- a/localstack-core/localstack/dev/kubernetes/__main__.py
+++ b/localstack-core/localstack/dev/kubernetes/__main__.py
@@ -26,7 +26,7 @@ def generate_k8s_cluster_config(pro: bool = False, mount_moto: bool = False, por
     )
     if pro:
         ext_path = os.path.join(root_path, "..", "localstack-ext")
-        ext_code_path = os.path.join(ext_path, "localstack_ext")
+        ext_code_path = os.path.join(ext_path, "localstack-pro-core", "localstack_ext")
         volumes.append(
             {
                 "volume": f"{os.path.normpath(ext_code_path)}:/code/localstack_ext",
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/dev/kubernetes/__main__.py
```diff
@@ -17,7 +17,9 @@ def generate_k8s_cluster_config(pro: bool = False, mount_moto: bool = False, por
         }
     )
 
-    egg_path = os.path.join(root_path, "localstack_core.egg-info/entry_points.txt")
+    egg_path = os.path.join(
+        root_path, "localstack-core", "localstack_core.egg-info/entry_points.txt"
+    )
```
alexrashed, line 22: q: Interesting, this seems to be a leftover from #10931. Which means the entrypoint mount wasn't working?

alexrashed (approved):
Awesome, thanks for adjusting the paths here after the adjustment of the source path initiated by #10800. 🦸🏽

alexrashed:
The test failures in the pipeline are unrelated, I think we can just move on and merge this PR! 🥳

localstack-core/localstack/dev/kubernetes/__main__.py
dominikschubert: Yep, noticed this coincidentally when checking the generated paths

PR #11093, opened 2024-06-26
remove dead code related to S3 v1 (pre-ASF)

bentsku:
## Motivation
Looking at #11090, I saw some dead code related to the legacy S3 provider(true legacy, the legacy v1 and not v2 😄)

This PR removes it the dead code, and move a bit of code around related to APIGW as it is only used there. 

## Changes
- remove dead code related to S3 select
- move one function to APIGW helpers
- remove a comment related to the dead code


alexrashed (approved):
LGTM! 🧹

cloutierMat (approved):
:rocket: Thanks for taking the time to do some maintenance :broom:

PR #11094, opened 2024-06-26
harden test_dynamodb_stream_records_with_update_item

bentsku:
## Motivation
Following up another test failure in https://app.circleci.com/pipelines/github/localstack/localstack/26140/workflows/8cc6ef1b-70fd-4e35-bea3-77ff3637a2db/jobs/221443/tests

And follow up from #10973 which was unsuccessful at fixing the test as it didn't tackle the proper origin of the problem. 

I've tried to harden the test to be more resilient. It seems the flake came from the fact that we would try to fetch the records, and if there weren't enough in the kinesis stream, we would fetch again and extend the already existing list of records with newly fetched records, leading to duplicate items (we can see the two`INSERT` events have the same `SequenceNumber`

## Changes
- update the logic to fetch the records from DynamoDB Streams

\cc @alexrashed

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -1451,7 +1451,7 @@ class TestDynamoDB:
         response = aws_client.dynamodbstreams.get_shard_iterator(
             StreamArn=stream_arn,
             ShardId=shard_id,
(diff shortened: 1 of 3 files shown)


giograno (approved):
Thanks for jumping on this 💯  The fix makes sense to me! Hopefully, it will stabilize the test.

PR #11095, opened 2024-06-26
Add support for docker container log configs

dfangl:
## Motivation
We currently have no support for docker logs configs, to set different docker log drivers, in our docker client implementation.

In order to facilitate some future functionality, we need to be able to use different log drivers.

This PR adds this functionality, and an example test which streams the logs to fluent-bit.

## Changes
* Support docker log configuration in the docker client


dominikschubert (approved):
LGTM! Nice addition! Curious if we might want integrate this via a custom log driver into our own logging system at some point as well, so it can be better correlated with individual resources, shown in the app, etc.

PR #11096, opened 2024-06-26
Apigw NG create method request handler

cloutierMat:
## Motivation

This PR ports the functionality from `apigateway/invocations.py:RequestValidator` to a method request handler.

The old invocation process wasn't routing the request before validation so some "passthrough" were required when no matching method were found. Cleaning up some of that code will make it cleaner and more obvious as to what is happening.

We can also start raising error from the gateway responses to be prepared for it's future implmentation.

## Changes

- updated the method request handler to validate the request parameters and the body.
- removed passthrough when no methods exists
- raise proper gateway exceptions
- added tests porting from the previous unit tests of the `RequestValidator` to ensure the functionality hasn't regressed.
- added a debug statement when the `ModelResolver` fails to find it's ref. There might be more work to do here, but that is the statement that helped me understand why my model wasn't resolving! :)

### Out of scope

While there could be an argument to include validation of api key as part of this handler as they are configured together,  they are separate enough in implementation that we can simplify readability/maintainability of the code by further breaking it down into it's own handler. All of the api key validation is done before request validation so a new handler seems like the best solution.

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
@@ -123,9 +123,7 @@ class MethodRequestHandler(RestApiGatewayHandler):
                     is_missing = param_value not in request.get("headers", [])
                 case "path":
                     path = request.get("path", "")
-                    print(param_value in path)
                     is_missing = param_value not in path
-                    print(f"{is_missing=}")
                 case "querystring":
                     is_missing = param_value not in request.get("query_string_parameters", [])
                 case _:
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
@@ -38,8 +38,8 @@ class MethodRequestHandler(RestApiGatewayHandler):
         self, method: Method, rest_api: RestApiContainer, request: InvocationRequest
     ) -> None:
         """
-        :raises BadRequestParameters if the request has required parameters which are not present
-        :raises BadRequestBody if the request has required body validation with a model and it does not respect it
+        :raises BadRequestParametersError if the request has required parameters which are not present
+        :raises BadRequestBodyError if the request has required body validation with a model and it does not respect it
         :return: None
         """
 
@@ -50,6 +50,7 @@ class MethodRequestHandler(RestApiGatewayHandler):
         # check if there is a validator for this request
         if not (validator := rest_api.validators.get(request_validator_id)):
             # TODO Should we raise an exception instead?
+            LOG.exception("No validator were found with matching id: '%s'", request_validator_id)
             return
 
         if self.should_validate_request(validator) and (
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
```diff
@@ -20,4 +28,120 @@ def __call__(
...
+            if not required:
+                continue
+
+            param_type, param_value = request_parameter.removeprefix("method.request.").split(".")
+            match param_type:
+                case "header":
+                    is_missing = param_value not in request.get("headers", [])
+                case "path":
+                    path = request.get("path", "")
+                    print(param_value in path)
```
bentsku, line 126: nit: looks like a leftover print statement

tests/unit/services/apigateway/test_handler_method_request.py
```diff
@@ -0,0 +1,288 @@
...
+from localstack.testing.config import TEST_AWS_ACCOUNT_ID, TEST_AWS_REGION_NAME
+
+
+@pytest.fixture
+def method_handler_chain() -> RestApiGatewayHandlerChain:
+    """Returns a dummy chain for testing."""
+    method_handler_chain = RestApiGatewayHandlerChain(request_handlers=[MethodRequestHandler()])
+    # Sets raise_on_error to allow capturing the Exceptions raised by the validator
+    method_handler_chain.raise_on_error = True
+    return method_handler_chain
```
bentsku, line 27: nit: should we use the handler chain with only handler, or should we directly instantiate the handler in the tests, so that we would not need to set `method_handler_chain.error = None` after every invocation?

In our Gateway, we instantiate a new handler chain for every incoming request.
We could maybe make the intent clearer by only testing the handler itself?

bentsku (approved):
LGTM! Nice refactor of the code 🚀 Just a minor comment regarding the testing, still not sure what our strategy should be regarding testing of handlers... I'd lean towards testing handlers in isolation, and should we need it, only instantiate a chain if we need different handlers executed one after another? What do you think?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_request.py
cloutierMat: oups! thanks :pray:

tests/unit/services/apigateway/test_handler_method_request.py
cloutierMat: I think you are right, I will see how I can make it cleaner by calling the handler itself! :smile:

cloutierMat:
I think you are right for the chain creation. 

My first thoughts were to also use the gateway response handlers and make assertion on the responses. But decided against it as it will be better as integration and snapshot tests later.

I updated the tests to use only the handler, I think it is much cleaner that way! Thanks for the comment :rocket:

tests/unit/services/apigateway/test_handler_method_request.py
bentsku: Very neat update 💯 I like that, I think we should follow the same pattern for new tests, with the invoker!

PR #11097, opened 2024-06-26
Add test for lifecycle of an IAM role with a path

dfangl:
## Motivation
I was doing some work on IAM role issues with paths, and wrote this test during trying to reproduce the issue.

The issue was elsewere, but the tests are still useful.

## Changes
* Add tests for the lifecycle of an IAM role with a path

### DIFF
--- a/tests/aws/services/iam/test_iam.py
+++ b/tests/aws/services/iam/test_iam.py
@@ -151,6 +151,36 @@ class TestIAMExtensions:
             )
         snapshot.match("invalid-json", e.value.response)
 
+    @markers.aws.validated
+    def test_role_with_path_lifecycle(self, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.iam_api())
+        role_name = f"role-{short_uid()}"
+        path = f"/path{short_uid()}/"
+        snapshot.add_transformer(snapshot.transform.regex(path, "<path>"))
+        assume_policy_document = {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Action": "sts:AssumeRole",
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -412,6 +412,10 @@ jobs:
               TEST_FILES=$(circleci tests glob "tests/aws/**/test_*.py" "tests/integration/**/test_*.py" | PYTHONPATH=localstack-core python -m localstack.testing.testselection.scripts.filter_by_test_selection target/testselection/test-selection.txt | circleci tests split --verbose --split-by=timings | tr '\n' ' ')
             fi
             echo $TEST_FILES
+            if [[ -z "$TEST_FILES" ]] ; then
```
alexrashed, line 415: Did this one here maybe slip into the commit? :)

alexrashed (approved):
The test is looking great, just looks like something slipped in from one of your recent PRs on the pipeline :)

.circleci/config.yml
dfangl: Actually, this is something I want to test out. In the test run linked (since a couple seconds) in the description, we can see the runners fail if not a single test is both selected and assigned to the runner, which leads to a red pipeline.

This is a preliminary, to be discussed change, to prevent that. Would love your input on it! We can also separate it into a separate PR.

alexrashed: Oh right, sorry, you mentioned that in the description, sorry! With the increased scope of the test selection this will definitely happen more often. Imho makes sense!
/cc @silv-io @dominikschubert

dfangl: #11105 is the replacement for this!

PR #11098, opened 2024-06-26
add plugin namespace localstack.init.runner to make init script runners pluggable

thrau:
## Motivation

People have asked whether they can use CDK or terraform in init hooks. To implement something like this, you'd currently need to statically code a runner into the init hook system, however I wanted to encapsulate terraform file support for init hooks into an extension.

To that end, I introduced a new plugin namespace `localstack.init.runner`, made the `ScriptRunner` a plugin of that namespace, and updated the `InitScriptManager` to use a `PluginManager` instead. After introducing this Inversion of Control, I noticed that it's better to have the condition, whether a script runner should be used or not, also encapsulated in the `ScriptRunner`. So now the `ScriptRunner` decides whether it is used for a script or not. The default implementation still uses file suffixes to make that determination, so nothing changed there.

## Changes

* The `ScriptRunner` class that executes init scripts is now a Plugin and loaded as such
* The `ScriptRunner` itself now determines whether it should be used to execute a script

## Testing

I've written an extension here https://github.com/thrau/localstack-terraform-init that builds on this change.

* Check out this PR
* In `localstack-ext` (extensions require pro), run
  ```sh
  python -m localstack.dev.run \
    -e EXTENSION_AUTO_INSTALL="git+https://github.com/thrau/localstack-terraform-init/#egg=localstack-terraform-init" \
    -e LOCALSTACK_API_KEY=test \
    -e DEBUG=1 \
    --mount-entrypoints \
    -v ./main.tf:/etc/localstack/init/ready.d/main.tf
  ```

In the logs you should see:

```
2024-06-26T20:36:19.946  INFO --- [ady_monitor)] l.extension                : Applying terraform project from file /etc/localstack/init/ready.d/main.tf
2024-06-26T20:36:19.946 DEBUG --- [ady_monitor)] localstack.utils.run       : Executing command: ['tflocal', '-chdir=/etc/localstack/init/ready.d', 'init', '-input=false']
2024-06-26T20:36:26.864 DEBUG --- [ady_monitor)] localstack.utils.run       : Executing command: ['tflocal', '-chdir=/etc/localstack/init/ready.d', 'apply', '-auto-approve']
```

### DIFF
--- a/localstack-core/localstack/runtime/init.py
+++ b/localstack-core/localstack/runtime/init.py
@@ -7,6 +7,8 @@ from enum import Enum
 from functools import cached_property
 from typing import Dict, List, Optional
 
+from plux import Plugin, PluginManager
+
 from localstack import constants
 from localstack.runtime import hooks
 from localstack.utils.objects import singleton_factory
@@ -47,10 +49,13 @@ class Script:
     state: State = State.UNKNOWN
 
 
-class ScriptRunner:
+class ScriptRunner(Plugin):
     """
     Interface for running scripts.
     """
+
+    namespace = "localstack.init.runner"
+    suffixes = []
 
     def run(self, path: str) -> None:
         """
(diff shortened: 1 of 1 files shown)


dfangl (approved):
Looks good!

PR #11099, opened 2024-06-27
add empty integrations for NG APIGW

bentsku:
## Motivation
This is first iteration to introduce the different REST API integrations.
The first idea was to use Plugins for each different `IntegrationType`, which allows easy discovery and to extend the list of integration in a transparent way.

However, while working on the `AWS` integration type, it became apparent that we could pull off all AWS services with a single integration, with no need to subclass like the current code does (we have `KinesisIntegration`, `S3Integration`...). But it happens that the `AWS` integration has the same logic for every service, and we will work to make it possible in the new iteration. 

## Changes
- create the different Integration types with their base class and some documentation


cloutierMat (approved):
:rocket: LGTM!

PR #11100, opened 2024-06-27
Fix docker image digest in startup runtime information for new runtime

dominikschubert:
## Motivation

Follow-up to a minor regression from https://github.com/localstack/localstack/pull/10942

Equivalent to https://github.com/localstack/localstack/pull/10951 but for the code path of the new runtime startup now.

## Changes

- On startup the value of LocalStack Docker image sha is now again the actual digest that can be pulled from the registry.


dominikschubert:
flake is unrelated (bookstore scenario)

PR #11106, opened 2024-06-27
Add pagination for SFN ListStateMachines, ListStateMachineVersions, and ListExecutions

gregfurman:
## Motivation
Improve parity of StepFunctions provider by adding pagination for [ListStateMachines](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ListStateMachines.html), [ListStateMachineVersions](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ListStateMachineVersions.html), and [ListExecutions](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ListExecutions.html).

## Changes
- Add pagination functionality and parameter validation checks to `list_state_machines`, `list_state_machine_versions`, and `list_executions` using `PaginatedList`.
 - Fix in `list_state_machines`: `ListStateMachine` response now returns `stateMachine` items in alphanumeric order as opposed to by datetime.
 - Fixed typo in `list_state_machine_versions` when raising a `RuntimeError` exception.

## Testing
In order to test pagination functionality, several new integration tests have been added with a `_pagination` suffix:
- Test `list_state_machines` with `test_list_sms_pagination`
- Test `list_state_machine_versions` with `test_list_state_machine_versions_pagination`)
- Test `list_executions` with `test_list_executions_pagination` and `test_list_executions_versions_pagination`

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -646,8 +646,9 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
         max_results: PageSize = None,
         **kwargs,
     ) -> ListStateMachineVersionsOutput:
-        # TODO: add paging support.
         self._validate_state_machine_arn(state_machine_arn)
+        assert_pagination_parameters_valid(max_results, next_token)
+        max_results = normalise_max_results(max_results)
 
         state_machines = self.get_store(context).state_machines
         state_machine_revision = state_machines.get(state_machine_arn)
@@ -661,11 +662,23 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
                 state_machine_version_items.append(state_machine_version.itemise())
             else:
                 raise RuntimeError(
-                    f"Expected {version_arn} to be a StateMachine Version, but gott '{type(state_machine_version)}'."
+                    f"Expected {version_arn} to be a StateMachine Version, but got '{type(state_machine_version)}'."
                 )
 
         state_machine_version_items.sort(key=lambda item: item["creationDate"], reverse=True)
-        return ListStateMachineVersionsOutput(stateMachineVersions=state_machine_version_items)
+
+        paginated_state_machine_versions = PaginatedList(state_machine_version_items)
+        page, token_for_next_page = paginated_state_machine_versions.get_page(
+            token_generator=lambda item: get_next_page_token_from_arn(
+                item.get("stateMachineVersionArn")
+            ),
+            page_size=max_results,
+            next_token=next_token,
+        )
+
+        return ListStateMachineVersionsOutput(
+            stateMachineVersions=page, nextToken=token_for_next_page
+        )
 
     def get_execution_history(
         self,
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -120,12 +120,18 @@ from localstack.services.stepfunctions.backend.state_machine import (
 )
 from localstack.services.stepfunctions.backend.store import SFNStore, sfn_stores
 from localstack.services.stepfunctions.backend.test_state.execution import TestStateExecution
+from localstack.services.stepfunctions.stepfunctions_utils import (
+    assert_pagination_parameters_valid,
+    get_next_page_token_from_arn,
+    normalise_max_results,
+)
 from localstack.state import StateVisitor
 from localstack.utils.aws.arns import (
     stepfunctions_activity_arn,
     stepfunctions_execution_state_machine_arn,
     stepfunctions_state_machine_arn,
 )
+from localstack.utils.collections import PaginatedList
 from localstack.utils.strings import long_uid, short_uid
 
 LOG = logging.getLogger(__name__)
@@ -613,14 +619,24 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
         next_token: PageToken = None,
         **kwargs,
     ) -> ListStateMachinesOutput:
-        # TODO: add paging support.
+        assert_pagination_parameters_valid(max_results, next_token)
+        max_results = normalise_max_results(max_results)
+
         state_machines: StateMachineList = [
             sm.itemise()
             for sm in self.get_store(context).state_machines.values()
             if isinstance(sm, StateMachineRevision)
         ]
-        state_machines.sort(key=lambda item: item["creationDate"])
-        return ListStateMachinesOutput(stateMachines=state_machines)
+        state_machines.sort(key=lambda item: item["name"])
+
+        paginated_state_machines = PaginatedList(state_machines)
+        page, token_for_next_page = paginated_state_machines.get_page(
+            token_generator=lambda item: get_next_page_token_from_arn(item.get("stateMachineArn")),
+            page_size=max_results,
+            next_token=next_token,
+        )
+
+        return ListStateMachinesOutput(stateMachines=page, nextToken=token_for_next_page)
 
     def list_state_machine_versions(
         self,
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -572,12 +572,16 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
         **kwargs,
     ) -> ListExecutionsOutput:
         self._validate_state_machine_arn(state_machine_arn)
+        assert_pagination_parameters_valid(
+            max_results=max_results,
+            next_token=next_token,
+            next_token_length_limit=3096,
+        )
+        max_results = normalise_max_results(max_results)
 
         state_machine = self.get_store(context).state_machines.get(state_machine_arn)
         if state_machine is None:
             self._raise_state_machine_does_not_exist(state_machine_arn)
-
-        # TODO: add support for paging
 
         allowed_execution_status = [
             ExecutionStatus.SUCCEEDED,
@@ -610,7 +614,17 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
                 execution, state_machine_arn=state_machine_arn, status_filter=status_filter
             )
         ]
-        return ListExecutionsOutput(executions=executions)
+
+        executions.sort(key=lambda item: item["startDate"], reverse=True)
+
+        paginated_executions = PaginatedList(executions)
+        page, token_for_next_page = paginated_executions.get_page(
+            token_generator=lambda item: get_next_page_token_from_arn(item.get("executionArn")),
+            page_size=max_results,
+            next_token=next_token,
+        )
+
+        return ListExecutionsOutput(executions=page, nextToken=token_for_next_page)
 
     def list_state_machines(
         self,
--- a/localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
+++ b/localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
@@ -1,9 +1,12 @@
+import base64
 import logging
 from typing import Dict
 
+from localstack.aws.api.stepfunctions import ValidationException
 from localstack.aws.connect import connect_to
 from localstack.utils.aws.arns import parse_arn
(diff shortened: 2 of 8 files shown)


gregfurman:

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,6 +317,102 @@ def test_list_sms(
...
+
+            sfn_snapshot.add_transformer(sfn_snapshot.transform.sfn_sm_create_arn(creation_resp, i))
+
+            state_machine_arn: str = creation_resp["stateMachineArn"]
+            state_machine_arns.append(state_machine_arn)
+
+        def _list_state_machines(expected_results_count: int, **kwargs):
+            """Returns a filtered list of relevant State Machines"""
+            state_machines = aws_client.stepfunctions.list_state_machines(**kwargs)
+            filtered_sms = [sm for sm in state_machines["stateMachines"] if sm["name"] in sm_names]
```
joe4dev, line 349: This filtering mitigates flaky tests if other state machines are present. Well done 👏
We often use unique prefixes, but this full match approach works equally well 💡

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,6 +317,102 @@ def test_list_sms(
...
+                    sm_name_set = {sm.get("name") for sm in filtered_state_machines}
+                    # assert that none of the State Machines being added are already present
+                    assert not any(sm.get("name") in sm_name_set for sm in filtered_page)
+
+                    filtered_state_machines.extend(filtered_page)
+
+            assert len(filtered_state_machines) == len(sm_names)
+            return filtered_state_machines
+
+        # Since ListStateMachines is eventually consistent, we should re-attempt pagination
```
joe4dev, line 377: 👍

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,6 +317,102 @@ def test_list_sms(
...
+        listed_state_machines = retry(_verify_paginate_results, retries=20, sleep=1)
+        sfn_snapshot.match("list-state-machines-page-2", listed_state_machines[-3:])
```
joe4dev, line 379: Is pagination fully reproducible on AWS? (i.e., guaranteed ordering by name according to our implementation)

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,6 +317,102 @@ def test_list_sms(
...
+        retry(
+            lambda: _list_state_machines(expected_results_count=13, maxResults=0),
+            retries=20,
+            sleep=1,
+        )
+
+        for state_machine_arn in state_machine_arns:
+            aws_client.stepfunctions.delete_state_machine(stateMachineArn=state_machine_arn)
+
+        # expect no state machines to be leftover after deletion
```
joe4dev, line 413: nit: more precisely, no filtered state machines that were created in this test :)

joe4dev (approved):
Congratulations to your first LocalStack PR @gregfurman 🥳 🚀 
Thorough test coverage and clean implementation 👏👏👏 

It looks good to me, so I just commented a few nits. I will leave the final review to our StepFunctions expert @MEPalma.

localstack-core/localstack/services/stepfunctions/provider.py
```diff
@@ -604,7 +614,17 @@ def list_executions(
                 execution, state_machine_arn=state_machine_arn, status_filter=status_filter
             )
         ]
-        return ListExecutionsOutput(executions=executions)
+
+        executions.sort(key=lambda item: item["startDate"], reverse=True)
+
+        paginated_executions = PaginatedList(executions)
+        page, nxt = paginated_executions.get_page(
```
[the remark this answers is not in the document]
gregfurman: Agree the naming I used here is a bit too abbreviated. Will change

tests/aws/services/stepfunctions/v2/test_sfn_api.py
gregfurman: I believe so. Parity tests were initially failing when the provider was sorting state machines by `creationDate`. Here, page 2 consisted of  `state_machine_{10:12}` (in that order). 

Equivalent page 2 when run against AWS returned `state_machine_{7:9}` (due to alphanumeric ordering). I attached the snippet below:

https://github.com/localstack/localstack/blob/e813ac5b77c72af79b00c2a02d8e6dfd8fc7a86d/tests/aws/services/stepfunctions/v2/test_sfn_api.snapshot.json#L1775-L1793

localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
```diff
@@ -23,3 +26,42 @@ def _get_result():
...
+                f"Value '{next_token}' at 'nextToken' failed to satisfy constraint: "
+                f"Member must have length less than or equal to {next_token_length_limit}"
+            )
+
+    if validation_errors:
+        errors_message = "; ".join(validation_errors)
+        message = f"{len(validation_errors)} validation {'errors' if len(validation_errors) > 1 else 'error'} detected: {errors_message}"
+        raise ValidationException(message)
+
+    return max_results, next_token
```
MEPalma, line 67: Why do we return `next_token` here? I can't see  this being used

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -383,6 +481,219 @@ def test_list_execution_invalid_arn(self, sfn_snapshot, aws_client):
...
+                stepfunctions_client=aws_client.stepfunctions, execution_arn=execution_arn
+            )
+
+        paginator = aws_client.stepfunctions.get_paginator("list_executions")
+
+        page_1_executions = paginator.paginate(
+            stateMachineArn=state_machine_arn, PaginationConfig={"MaxItems": 10}
+        ).build_full_result()
+
+        assert len(page_1_executions["executions"]) == 10
```
MEPalma, line 533: It seems snapshotting the results here would achieve the same check, and ensure the results are well formed also when using pagination

tests/aws/services/stepfunctions/v2/test_sfn_api_versioning.py
```diff
@@ -131,6 +131,126 @@ def test_create_publish_describe_with_version_description(
...
+            await_state_machine_version_listed(
+                aws_client.stepfunctions, state_machine_arn, update_resp_1["stateMachineVersionArn"]
+            )
+
+        page_1_state_machine_versions = aws_client.stepfunctions.list_state_machine_versions(
+            stateMachineArn=state_machine_arn,
+            maxResults=10,
+        )
+
+        assert len(page_1_state_machine_versions["stateMachineVersions"]) == 10
```
MEPalma, line 179: Same comment as above about recording the responses instead.

MEPalma (commented):
Thank you for your work on this! It looks good overall.
I'm curious why we aren't testing scenarios where the items in the pagination change. For instance, what if a new execution starts after the first listing request? Testing this could help solidify the behaviour of the various listing strategies for each field now and in the future.

tests/aws/services/stepfunctions/v2/test_sfn_api.py
gregfurman: Sure that makes sense. This assertion here serves more as a sanity check. Initially was creating 103 state machines (100 items on page 1; 3 items on page 2) so snapshotting that many results seemed excessive.

I left this in since if the first 10 items (page 1) were incorrectly paginated, I'd expect the last 3 items (page 2) to be incorrect -- hence just checking page 2 seemed less expensive.

If you think this assumption is incorrect or if an extra 10 elements being included will add minimal overhead, I'm happy to change this to also snapshot the first 10 items too.

localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
gregfurman: You're correct that it's not currently being used. Probably a bit premature but I wanted the function to be flexible enough to potentially return a mutated/edited version of the `next_token` (similar to how the default value of the `max_results` gets set).

tests/aws/services/stepfunctions/v2/test_sfn_api.py
MEPalma: On one hand, I agree with you that we should aim to keep the snapshots small. However, these are baseline tests for a specific feature, and I think it is important to capture this behaviour. Recording ten responses should be acceptable. Future tests that build on this feature can potentially skip these detailed responses.

As a side note, we could pass an empty dictionary as execution input to each machine. There is no real need to pass the numbered binding you're using now. The ordering should still be deducible from the execution ARN encoding.

gregfurman: > Future tests that build on this feature can potentially skip these detailed responses.

Considering this is testing the core pagination functionality, I agree there is good reason to be more explicit in testing. Will change.

> As a side note, we could pass an empty dictionary as execution input to each machine.

Seems reasonable 👍 The test doesn't actually care about this input so adding that enumerated JSON seems like an unnecessary step.

localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
MEPalma: I understand your point, and conceptually, it does seem appealing. However, I don't think there's enough context provided to this function to produce another next_token.

Additionally, the function is quite crowded. It might be beneficial to separate its responsibilities between normalising and validating. Currently, the function handles validation and returns the max results value if one isn't set. In the future, the next token could be adjusted during the normalisation step instead. What do you think?

gregfurman: I coupled this normalisation and validation functionality together since:
- If `max_results` has been normalised, we don't need to check whether it's valid.
- The normalisation step was so trivial (default to `100` if `None` or `0`, I thought including it in the validation step would yield less overhead.

Not sure if this small check necessitates a new function given it would probably just look like:

```python
def normalise_max_results(max_results: int = 100) -> int:
   if not max_results:
      return _DEFAULT_SFN_MAX_RESULTS
   return max_results
```

Ideally, are you suggesting a normalisation function (like above) with

```python
max_results = normalise_max_results(max_result)
assert_pagination_parameters_valid(max_results, next_token) # will raise an Exception if invalid
```

I agree, however, about returning `next_token` being unnecessary, so happy to remove that.

tests/aws/services/stepfunctions/v2/test_sfn_api.py
gregfurman: I also ended up removing the unnecessary paginator objects in `test_list_executions_pagination`, test_list_executions_versions_pagination`, and `test_list_state_machine_versions_pagination`.

gregfurman:
> I'm curious why we aren't testing scenarios where the items in the pagination change. For instance, what if a new execution starts after the first listing request? Testing this could help solidify the behaviour of the various listing strategies for each field now and in the future.

@MEPalma What are your thoughts on these "reliability" tests residing in separate functions? The current `_pagination` tests are looking quite long-as-is.

localstack-core/localstack/services/stepfunctions/stepfunctions_utils.py
MEPalma: I agree the normalisation function would be quite small, however it means the validation function would only validate, like for all the other validation functions in this provider. I believe a separation like this, however small, would work best (also in light of your considerations about having the next_token change in the future).

MEPalma:
These tests are best placed in new test cases as they would validate a specific scenario; and absolutely we do not want these tests to grow in complexity!

tests/aws/services/stepfunctions/v2/test_sfn_api.py
MEPalma: Thank you!

MEPalma (approved):
Great work on this @gregfurman, and thank you for contributing to LocalStack and StepFunctions! 

Changes looked good from the start, but thank you for addressing the comments too, which mainly focused on using your momentum on this feature to cover some edge cases. 

I look forward to using this feature in the future! 🎉

PR #11107, opened 2024-06-27
add APIGW NG context variables

cloutierMat:
## Motivation

Start of the process to populate the context variables for template rendering.


bentsku (approved):
LGTM! Thanks for pro-actively merging this so that we don't have conflict 👍

PR #11111, opened 2024-06-28
add context variables and stage variables to API NG context

bentsku:
## Motivation
We need a separate context, representing the `$context` variables used to render VTL templates, and which is also passed in the payload to `AWS_PROXY` integrations. 

More informations about `$context` here. 
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference

It is a bit of duplicated information from the original `RestApiInvocationContext` object and the `InvocationRequest`, but they serve different purposes.

If we need to update different contexts at the same time, then maybe we could simplify. But we learned that creating those context when needed, collecting informations from different sources, can be pretty hard. 

The `InvocationRequest` goal is simply to represent the incoming request "as is", with no transformation. 

The idea would be to go from `InvocationRequest`, get the `Resource`. Use the original `InvocationRequest`, the `Resource`,the `ContextVariables` and `StageVariables` to construct the `IntegrationRequest` and send it down.
The `InvocationRequest` will not be mutated. The `ContextVariables` will be mutated and populated along the way of the different handlers. 

## Changes
- add the `ContextVariables` and populate it
- add the not yet used `LoggingContextVariables` which can add more data to the logging
- add `IntegrationRequest` for future work in integration request handler
- add the stage variables and fetch them from the stage
- add small tests to validate we're populate the contexts properly

## TODO

What's left to do:

- [x] add unit tests

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -65,6 +65,8 @@ class RestApiInvocationContext(RequestContext):
     """The REST API identifier of the invoked API"""
     stage: Optional[str]
     """The REST API stage linked to this invocation"""
+    deployment_id: Optional[str]
+    """The REST API deployment linked to this invocation"""
     region: Optional[str]
     """The region the REST API is living in."""
     account_id: Optional[str]
@@ -87,6 +89,7 @@ class RestApiInvocationContext(RequestContext):
         self.deployment = None
         self.api_id = None
         self.stage = None
+        self.deployment_id = None
         self.account_id = None
         self.region = None
         self.invocation_request = None
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -8,6 +8,8 @@ from werkzeug.datastructures import Headers
 from localstack.aws.api.apigateway import Method, Resource
 from localstack.services.apigateway.models import RestApiDeployment
 
+from .variables import ContextVariables, LoggingContextVariables
+
 
 class InvocationRequest(TypedDict, total=False):
     http_method: Optional[HTTPMethod]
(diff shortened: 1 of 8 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
```diff
@@ -33,55 +35,21 @@ class InvocationRequest(TypedDict, total=False):
...
+    query_string_parameters: Optional[dict[str, str]]
+    """Query string parameters of the request"""
+    headers: Optional[dict[str, str]]
+    """Headers of the request"""
+    multi_value_query_string_parameters: Optional[dict[str, list[str]]]
+    """Multi value query string parameters of the request"""
+    multi_value_headers: Optional[dict[str, list[str]]]
+    """Multi value headers of the request"""
+    body: Optional[bytes]
+    """Body content of the request"""
```
cloutierMat, line 52: Having that clear separation from the invocation request will be key in maintaining transparency over the requests.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
```diff
@@ -99,3 +111,44 @@ def _get_single_and_multi_values_from_headers(
...
+            path=invocation_request[
+                "path"
+            ],  # TODO: check if we need the raw path? with forward slashes
+            protocol="HTTP/1.1",
+            requestId=short_uid(),  # TODO: use snapshot tests to verify format
+            requestTime=timestamp(time=now, format=REQUEST_TIME_DATE_FORMAT),
+            requestTimeEpoch=int(now.timestamp() * 1000),
+            stage=context.stage,
+        )
+        return context_variables
```
cloutierMat, line 140: Should we log this dict immediately?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/resource_router.py
```diff
@@ -94,6 +96,16 @@ def route_and_enrich(self, context: RestApiInvocationContext):
         )
         context.resource_method = method
 
+        self.update_context_variables_with_resource(context.context_variables, resource)
+
+    @staticmethod
+    def update_context_variables_with_resource(
+        context_variables: ContextVariables, resource: Resource
+    ):
+        # TODO: log updating the context_variables?
```
cloutierMat, line 105: It is fine to leave as a TODO, but I think it would really useful to log them in debug mode.

localstack-core/localstack/services/apigateway/next_gen/execute_api/variables.py
```diff
@@ -0,0 +1,181 @@
+from typing import Optional, TypedDict
+
+
+class ContextVarsAuthorizer(TypedDict, total=False):
+    # this is merged with the Context returned by the Authorizer, which can attach any property to this dict in string
+    # format
```
cloutierMat, line 6: Good catch!

cloutierMat (approved):
Awesome! I can see so much extra code and repeated functionality we will be able to remove by having the context variables built and maintained together like that!

The split in integration request already makes it very clear as to what is meant to happen and this clear display of the logic's order will help maintaining the service :rocket: 

Nice additions to the tests too :wink:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
bentsku: Done! Good catch! We can revisit the overall logging later on when we have a clear picture, to make it a consistent experience, but for now simply logging it is really good. I believe we will also realize later on some values needs default values, maybe `$context.identity` for example?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/resource_router.py
bentsku: Done 👍

bentsku (commented):
Thanks a lot for the review! 🙏 good comments about the logging, I've looked into #11114 and took the same format as you, it was very good 😄

PR #11114, opened 2024-06-28
APIGW NG implement api key validation handler

cloutierMat:
## Motivation

Implementation of a new api key validation handler. There is a mix of moto helpers in the implementation as not all relevant information live in the deployment. 

#### Important configuration in the frozen deployment:
- The resource method describes `apiKeyRequired`, which determines if validation is required or skipped.
- The rest api describes `apiKeySource`, which determines if the key is to be retrieved from the headers or from an identity context populated by an Lambda Authorizer

#### Important configuration in apigateway store.
Those items, when created, modified or deleted, will affect a deployment without the need to redeploy. see https://docs.aws.amazon.com/apigateway/latest/developerguide/updating-api.html
- Api keys and their properties are stored in moto (contradicts the documentation, but manually tested that updating `enabled` from the key will affect without redeploying)
- Usage plans are stored in moto
- Usage plan key (association of a key to a usage plan) are also stored in moto

## Changes

- Finds and validates api key based on deployment configuration and keys/usage plans stored on the backend
- Updates the `$context` variables with relevant data.
- Added `moto_helpers` to keep moto backend isolated and ease technical debt.
- Added tests to validate functionality

## TODO

What's left to do:

- [x] This PR currently targets #11111, as it requires changes to the imports. When merged, we should rebase and target master

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/api_key_validation.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/api_key_validation.py
@@ -1,0 +1,112 @@
+import logging
+from typing import Optional
+
+from localstack.aws.api.apigateway import ApiKey, ApiKeySourceType, RestApi
+from localstack.http import Response
+
+from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
+from ..context import IdentityContext, InvocationRequest, RestApiInvocationContext
+from ..gateway_response import InvalidAPIKeyError
+from ..moto_helpers import get_api_key, get_usage_plan_keys, get_usage_plans
+
+LOG = logging.getLogger(__name__)
+
+
+class ApiKeyValidationHandler(RestApiGatewayHandler):
+    """
+    Handles Api key validation.
+    If an api key is required, we will validate that a usage plan associated with that stage
+    has a usage plan key with the corresponding value.
+    """
+
+    # TODO We currently do not support rate limiting or quota limit. As such we are not raising any related Exception
+
+    def __call__(
+        self,
+        chain: RestApiGatewayHandlerChain,
+        context: RestApiInvocationContext,
+        response: Response,
+    ):
+        method = context.resource_method
+
+        # If api key is not required by the method, we can exit the handler
+        if not method.get("apiKeyRequired"):
+            return
+
+        identity = context.context_variables["identity"]
+        request = context.invocation_request
+        rest_api = context.deployment.rest_api.rest_api
+
+        # Look for the api key value in the request. If it is not found, raise an exception
+        if not (api_key_value := self.get_request_api_key(rest_api, request, identity)):
+            LOG.debug("API Key is empty")
+            raise InvalidAPIKeyError("Forbidden")
+
+        # Get the validated key, if no key is found, raise an exception
+        if not (validated_key := self.get_validated_key(api_key_value, context)):
+            LOG.debug("Provided API Key is not valid")
+            raise InvalidAPIKeyError("Forbidden")
+
+        # Update context's identity with the key value and Id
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/api_key_validation.py
```diff
@@ -0,0 +1,115 @@
...
+        request = context.invocation_request
+        rest_api = context.deployment.rest_api.rest_api
+
+        # If api key is not required by the method, we can exit the handler
+        if not method.get("apiKeyRequired"):
+            return
+
+        # If the Identity context was not created yet, instantiate it and attach it to the context variables
+        if not (identity := context.context_variables.get("identity", ContextVarsIdentity())):
+            context.context_variables["identity"] = identity
```
bentsku, line 41: You can simplify with `setdefault(key, default_value)` which does the same thing under the hood:
```suggestion
        identity = context.context_variables.setdefault("identity", ContextVarsIdentity())
```

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/api_key_validation.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Update the context's identity with the key value and id
+        if not identity.get("apiKey"):
+            LOG.debug("Updating $context.identity.apiKey='%s'", validated_key["value"])
+            identity["apiKey"] = validated_key["value"]
+
+        LOG.debug("Updating $context.identity.apiKeyId='%s'", validated_key["id"])
+        identity["apiKeyId"] = validated_key["id"]
+
+    def validate_api_key(
```
bentsku, line 61: I believe once we migrate `UsagePlan` and `ApiKey` to our store, seeing this usage, we should keep a map of API Key values (+ maybe stage?) directly in order to quickly validate them? We could maybe then simplify this to not have to iterate over all usage plans and usage plan keys.

bentsku (approved):
LGTM! 🚀 extremely well covered PR, the tests are really great, love the abstractions and usage of fixtures. 

The handler is looking extremely good, well documented, nicely encapsulated functions and helpers, this is awesome 💯

localstack-core/localstack/services/apigateway/next_gen/execute_api/moto_helpers.py
```diff
@@ -38,3 +45,34 @@ def get_stage_variables(
...
+        UsagePlan(**usage_plan.to_json()) for usage_plan in apigateway_backend.usage_plans.values()
+    ]
+
+
+def get_api_key(api_key_id: str, account_id: str, region_name: str) -> ApiKey:
+    """
+    Will return an api key from the moto store.
+    """
+    apigateway_backend: APIGatewayBackend = apigateway_backends[account_id][region_name]
+    return ApiKey(**apigateway_backend.keys[api_key_id].to_json())
```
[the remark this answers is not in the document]
cloutierMat: Actually realised I don't even need to go that far and and with simply `return apigateway_backend.keys[api_key_id].to_json()` all the typing works fine. Which I can also use to simplify the code of `get_usage_plans` and `get_usage_plan_keys` and remove the destructing while keeping typing!
:+1:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/api_key_validation.py
cloutierMat: Yes, I believe we can keep the item in the store in a way that will be more favourable to this specific access pattern. I don't think we will be able to avoid all looping, without creating some data duplication but we should be able to reduce the amount of work being done here.

Issue #11104, opened 2024-06-27
feature request: Implement the idempotent nature of start-execution in Standard Step Functions


robbiewareham:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

When requesting the start of an execution of a standard Step Function, there is a level of idempotency supported by AWS if a duplicate execution name is used:

Taken from https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartExecution.html:

**_StartExecution** is idempotent for STANDARD workflows. For a STANDARD workflow, if you call **StartExecution** with the same name and input as a running execution, the call succeeds and return the same response as the original request. If the execution is closed or if the input is different, it returns a **400 ExecutionAlreadyExists** error. You can reuse names after 90 days._

However, this is not implemented in Localstack.  Instead, any attempt to reuse an existing execution name results in:

```
Error while processing message c41df0f9-5704-4a48-9df3-d2554e4afdb0: The service returned an error with Error Code InvalidName and HTTP Body: {"__type": "InvalidName", "message": ""}, Inner Exception: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
```

gregfurman:
Hey! Can I be assigned this?

gregfurman:
Hi @robbiewareham! This should be solved from `+v3.7`. Can you confirm all works?

PR #11118, opened 2024-07-01
Add idempotency when calling StartExecution in Standard Step Functions

gregfurman:
## Motivation
- Calling `StartExecution` with the same `name` and `input` payload as a currently running execution should be idempotent for `STANDARD` workflows . 
- That is, the  `StartExecution` request should succeed and return the same response as the original `StartExecution` request (from #11104)

From [AWS StartExecution docs](https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartExecution.html):
> StartExecution is idempotent for STANDARD workflows. For a STANDARD workflow, if you call StartExecution with the same name and input as a running execution, the call succeeds and return the same response as the original request. If the execution is closed or if the input is different, it returns a 400 ExecutionAlreadyExists error. You can reuse names after 90 days.
>
> StartExecution isn't idempotent for EXPRESS workflows.

## Changes
- Added `_idempotent_start_execution` function that returns a `RUNNING` execution matching `input` and `name` parameters for workflows of type `STANDARD`;  and associated tests in `test_start_execution_idempotent`
- Added a shorter wait-and-pass BaseTemplate that waits 10 seconds before passing.
- Fixed a typo when naming snapshot in `test_describe_execution`  (`ddescribe -> describe`).
- Fixed `test_start_execution` test by skipping `redriveCount` check.

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -247,6 +247,38 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
                 return state_machine
         return None
 
+    def _idempotent_start_execution(
+        self,
+        context: RequestContext,
+        exec_arn: Arn,
+        state_machine: StateMachineInstance,
+        name: Name,
+        input_data: SensitiveData,
+    ) -> Optional[Execution]:
+        # StartExecution is idempotent for STANDARD workflows. For a STANDARD workflow,
+        # if you call StartExecution with the same name and input as a running execution,
+        # the call succeeds and return the same response as the original request.
+        # If the execution is closed or if the input is different,
+        # it returns a 400 ExecutionAlreadyExists error. You can reuse names after 90 days.
+
+        execution = self.get_store(context).executions.get(exec_arn)
+        if not execution:
+            return None
+
+        match (name, input_data, execution.exec_status, state_machine.sm_type):
+            case (
+                execution.name,
+                execution.input_data,
+                ExecutionStatus.RUNNING,
+                StateMachineType.STANDARD,
+            ):
+                return execution
+
+        raise CommonServiceException(
+            code="ExecutionAlreadyExists",
+            message=f"Execution Already Exists: '{exec_arn}'",
+        )
+
     def _revision_by_name(
         self, context: RequestContext, name: str
     ) -> Optional[StateMachineInstance]:
@@ -508,7 +540,17 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
             normalised_state_machine_arn, exec_name
         )
         if exec_arn in self.get_store(context).executions:
-            raise InvalidName()  # TODO
+            # Return already running execution if name and input match
+            existing_execution = self._idempotent_start_execution(
+                context=context,
+                exec_arn=exec_arn,
+                state_machine=state_machine_clone,
+                name=name,
+                input_data=input_data,
+            )
+
+            if existing_execution:
+                return existing_execution.to_start_output()
 
         # Create the execution logging session, if logging is configured.
         cloud_watch_logging_session = None
@@ -531,6 +573,7 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
             trace_header=trace_header,
             activity_store=self.get_store(context).activities,
         )
+
         self.get_store(context).executions[exec_arn] = execution
 
         execution.start()
--- a/tests/aws/services/stepfunctions/templates/base/statemachines/wait_10_sec_and_pass.json5
+++ b/tests/aws/services/stepfunctions/templates/base/statemachines/wait_10_sec_and_pass.json5
@@ -1,0 +1,18 @@
+{
+  "Comment": "WAIT_10_SEC",
+  "StartAt": "State_1",
+  "States": {
+    "State_1": {
+      "Type": "Wait",
+      "Seconds": 10,
+      "Next": "State_2"
+    },
+    "State_2": {
+      "Type": "Pass",
+      "Result": {
+        "Arg1": "argument1",
+      },
+      "End": true,
+    },
+  },
+}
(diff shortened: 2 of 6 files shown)


tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,7 +315,58 @@ def test_list_sms(
...
+        # Should fail because the execution has the same 'name' as another but a different 'input'.
+        with pytest.raises(Exception) as err:
+            aws_client.stepfunctions.start_execution(
+                stateMachineArn=state_machine_arn,
+                input='{"body" : "different-data"}',
+                name=execution_name,
+            )
+        sfn_snapshot.match("start_exec_already_exists", err.value.response)
+
+        await_execution_success(
```
MEPalma, line 362: This test last 10 seconds as it waits for the wait state in the state machine to exit. You could use a [`waitForTaskToken`](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html#connect-wait-token) on an `sqs` queue `SendMessage` to have the running state machine wait until you can safely shut it down with a `SendTaskSuccess`. We have many tests for this service, you can take inspiration from those that use the `sqs_send_task_success_state_machine` fixture. This would also remove any flakiness should the second start command happen after the first execution terminated.

tests/aws/services/stepfunctions/templates/base/statemachines/wait_10_sec_and_pass.json5
```diff
@@ -0,0 +1,18 @@
+{
```
MEPalma, line 1: This is probably not needed as it is a duplicate of `WAIT_1_MIN`, which can easily be reconfigured as it's done in tests such as `aws.services.stepfunctions.v2.base.test_base.TestSnfBase.test_event_bridge_events_base`. Should save us from having too many templates.

MEPalma (changes requested):
Thank you for looking into this issue!
I just have a concern about the behaviour of these changes (and AWS's) whenever the execution name is a duplicate, but not the input data. Adding a test on this scenario may prove useful. I would also like for the test case added to be revised, as it could be sped up and have it's flakiness reduced/zeroed.

tests/aws/services/stepfunctions/v2/test_sfn_api.py
```diff
@@ -315,7 +315,58 @@ def test_list_sms(
         lst_resp_filter = [sm for sm in lst_resp["stateMachines"] if sm["name"] in sm_names]
         sfn_snapshot.match("lst_resp_del_filter", lst_resp_filter)
 
-    @markers.aws.needs_fixing
+    @markers.aws.validated
```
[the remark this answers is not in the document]
gregfurman: This is already in the test. See how the below retains the same ARN but changes the input data:

https://github.com/localstack/localstack/blob/54ad23a14312c3a2d13b871bc536823ea655c893/tests/aws/services/stepfunctions/v2/test_sfn_api.py#L354-L360

localstack-core/localstack/services/stepfunctions/provider.py
```diff
@@ -508,7 +540,17 @@ def start_execution(
...
+            # Return already running execution if name and input match
+            existing_execution = self._idempotent_start_execution(
+                context=context,
+                exec_arn=exec_arn,
+                state_machine=state_machine_clone,
+                name=name,
+                input_data=input_data,
+            )
+
+            if existing_execution:
```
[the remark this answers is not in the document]
MEPalma: I see now, this function returns None after sampling the store for the execution arn! A similar check is run by the provider before invoking `_idempotent_start_execution`. Maybe we can access the store only once and revise the flow slightly here to avoid confusions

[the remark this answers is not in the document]
gregfurman: I added the store check to the `_idempotent_start_execution` function to make it better suited to running in isolation. Agree that the double check is unneccesary so happy to remove it.

tests/aws/services/stepfunctions/v2/test_sfn_api.py
gregfurman: I gave the `sqs_send_task_success_state_machine` fixture a look and I'm not sure it's too compatible with how these state machine and execution steps are run.

How about we increase the running time to 60 seconds (instead of 10) and terminate the execution after getting the execution already exists error (the final test case). Then we can block until the execution has terminated/finished. 

That way, the test will have an upper bound of 60 seconds (minimising flakiness) while also _technically_ being allowed to take less than 10 seconds -- making this faster in the average case.

tests/aws/services/stepfunctions/v2/test_sfn_api.snapshot.json
```diff
@@ -2260,5 +2261,52 @@
...
+      "exec_resp": {
+        "executionArn": "arn:<partition>:states:<region>:111111111111:execution:<resource:1>:<resource:2>",
+        "startDate": "datetime",
+        "ResponseMetadata": {
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 200
+        }
+      },
+      "exec_resp_idempotent": {
+        "executionArn": "arn:<partition>:states:<region>:111111111111:execution:<resource:1>:<resource:2>",
```
MEPalma, line 2285: Idempotency check is indeed captured

tests/aws/services/stepfunctions/v2/test_sfn_api.snapshot.json
```diff
@@ -1534,7 +1535,7 @@
     }
   },
   "tests/aws/services/stepfunctions/v2/test_sfn_api.py::TestSnfApi::test_describe_execution": {
-    "recorded-date": "08-03-2024, 11:52:48",
+    "recorded-date": "30-06-2024, 14:51:30",
```
MEPalma, line 1538: We should try to generate this again, as the typo was not updated here

MEPalma (commented):
Thank you @gregfurman! 
I just see the need for a small rearrangement for the shutting down of the primary state machine, and a simple update of a snapshot test where the typo was fixed.

MEPalma (approved):
Thank you @gregfurman!

PR #11121, opened 2024-07-01
APIGW NG implement request data mapping for integration request

bentsku:
## Motivation

This PR implements Parameter Mappings and the base util class. It is also integrated into the Integration Request handler, but that part is still untested because we will wait for the handler to be complete before testing its behavior. 

More informations about Parameters Mappings in this documentation page:
https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html

It allows you to map values from your incoming request (path parameters, query string and headers), as well as the body and even parts of the body with JSON Path. You can also map values from the `$context` variable and stage variables, as well as static data. 

## Changes
- implement a small change to Gateway Response, as it seems AWS is raising `DEFAULT_4XX` exceptions directly and that wasn't expected. We can now specify a status code directly at runtime when raising this kind of exception
- implement the `ParametersMapper` class which contains the method to get the needed values for your integration request: header, path parameters and query string. 
- implement unit tests to validate the behavior. Most of these unit tests have been written by checking with the AWS Console, but will need to be ported to a few integration tests to confirm behavior.

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -52,8 +52,14 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         # TODO: create helper method to render the different URI with stageVariables and parameters
         rendered_integration_uri = integration["uri"]  # use request_data_mapping["path"]
 
+        # TODO: verify the assumptions about the method with an AWS validated test
+        # if the integration method is defined and is not ANY, we can use it for the integration
+        if not (integration_method := integration["httpMethod"]) or integration_method == "ANY":
+            # otherwise, fallback to the request's method
+            integration_method = context.invocation_request["method"]
+
         integration_request = IntegrationRequest(
-            http_method=integration["httpMethod"],
+            http_method=integration_method,
             uri=rendered_integration_uri,
             query_string_parameters=request_data_mapping["querystring"],
             headers=request_data_mapping["header"],
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -30,7 +30,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
 
         if integration_type in (IntegrationType.AWS_PROXY, IntegrationType.HTTP_PROXY):
             # `PROXY` types cannot use integration mapping templates
-            # TODO: check if PROXY can still parameters mapping and substitution in URI for example?
+            # TODO: check if PROXY can still parameters mapping and substitution in URI for example? normally not
             # See
             return
 
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -38,7 +38,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
 
         # TODO: this handler might not be tested yet until `TemplateMappings` are implemented, as overall behavior will
         #  change
-        integration_request_parameters = integration["requestParameters"]
+        integration_request_parameters = integration["requestParameters"] or {}
         request_data_mapping = self.get_integration_request_data(
             context, integration_request_parameters
         )
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -4,7 +4,8 @@ from localstack.aws.api.apigateway import Integration, IntegrationType
 from localstack.http import Response
 
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
-from ..context import RestApiInvocationContext
+from ..context import IntegrationRequest, RestApiInvocationContext
+from ..parameters_mapping import ParametersMapper, RequestDataMapping
 
 LOG = logging.getLogger(__name__)
 
@@ -14,6 +15,9 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
     This class will take care of the Integration Request part, which is mostly linked to template mapping
     See https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-integration-settings-integration-request.html
     """
+
+    def __init__(self):
+        self._param_mapper = ParametersMapper()
 
     def __call__(
         self,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -26,15 +30,47 @@ def __call__(
...
+        #  it might populate the context requestOverride too, maybe deepcopy the ContextVariables because it will
+        #  get mutated, or pop it directly? might just be better!
+
+        # TODO: extract the code under into its own method
+
+        # TODO: create helper method to render the different URI with stageVariables and parameters
+        rendered_integration_uri = integration["uri"]  # use request_data_mapping["path"]
+
+        integration_request = IntegrationRequest(
+            http_method=integration["httpMethod"],
```
cloutierMat, line 56: Question: Can this be `ANY`?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -26,15 +30,47 @@ def __call__(
 
         if integration_type in (IntegrationType.AWS_PROXY, IntegrationType.HTTP_PROXY):
             # `PROXY` types cannot use integration mapping templates
-            # TODO: check if PROXY can still parameters mapping and substitution in URI for example?
+            # TODO: check if PROXY can still parameters mapping and substitution in URI for example? normally not
             # See
             return
```
cloutierMat, line 35: Question: Would there be an advantage to returning an `IntegrationRequest` already populated with the `InvocationRequest` value? Could that streamline the work in the `_proxy` handlers?

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -0,0 +1,160 @@
...
+                value = self._retrieve_parameter_from_stage_variables(
+                    stage_var_name, stage_variables
+                )
+
+            elif request_mapping.startswith("'") and request_mapping.endswith("'"):
+                value = request_mapping.strip("'")
+
+            else:
+                # TODO: verify this
+                value = None
```
cloutierMat, line 71: Should we raise or log? Could help some strange debugging in the future?

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -0,0 +1,160 @@
...
+        pass
+
+    def _retrieve_parameter_from_invocation_request(
+        self, expr: str, invocation_request: InvocationRequest
+    ) -> str | list[str]:
+        if expr.startswith("body"):
+            body = invocation_request["body"] or b"{}"
+            body = body.strip()
+            # AWS only tries to JSON decode the body if it starts with some leading characters ({, [, ", ')
+            # otherwise, it ignores it
```
cloutierMat, line 88: nit: This comment is duplicate from the `_json_load` method. Should it just be removed from here?

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -0,0 +1,160 @@
...
+                decoded_body = self._json_load(body)
+            except ValueError:
+                raise Default4xxError(message="Invalid JSON in request body")
+
+            if expr == "body":
+                return to_str(body)
+
+            elif expr.startswith("body."):
+                json_path = expr.removeprefix("body.")
+                return self._get_json_path_from_dict(decoded_body, json_path)
```
cloutierMat, line 99: nit: In case there is a malformed expression, we should probably make sure that we don't continue matching. Maybe we should add a else and LOG something and return `None`?

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -0,0 +1,160 @@
...
+
+        elif param_type == "header":
+            multi_headers = invocation_request["multi_value_headers"].get(param_name)
+            if multi_headers:
+                return multi_headers[-1]
+
+        elif param_type == "multivalueheader":
+            multi_headers = invocation_request["multi_value_headers"].get(param_name, [])
+            return ",".join(multi_headers)
+
```
cloutierMat, line 128: Same here, we could log if no match exists, that could simplify debugging.

tests/unit/services/apigateway/test_parameters_mapping.py
```diff
@@ -0,0 +1,399 @@
...
+
+        mapping = mapper.map_integration_request(
+            request_parameters=request_parameters,
+            invocation_request=default_invocation_request,
+            context_variables=default_context_variables,
+            stage_variables={},
+        )
+        # this was validated against AWS
+        # it does not forward the body even with passthrough, but the content of `method.request.body` is `{}`
+        # if the body is empty
```
cloutierMat, line 165: :rofl::rofl::rofl:

cloutierMat (commented):
Looks really good! :rocket: I am excited for this handler to start taking shape like this!

Nice investigation on the json body decoding issue! That is quite counter intuitive what is happening there, I am glad those behaviors are now well documented in the code and tests.

The added tests already cover a good range of edge cases!

I left one comment concerning `ANY`. Just making sure it doesn't come back to bite us again!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
bentsku: I'm not sure, and I also see this parameter as not required from the documentation, we'd need to write tests for this: https://docs.aws.amazon.com/apigateway/latest/api/API_PutIntegration.html#API_PutIntegration_RequestSyntax

It might fall back to the default `resourceMethod`? I think we need custom logic here that we need to investigate. I'll update with something temporary with a TODO to come back to this.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
bentsku: Yeah, I wonder that as well. Should the `_proxy` just directly use the `InvocationRequest`, or should we just manually basically copy `InvocationRequest` into `IntegrationRequest`, where we could for example remove the `Host` header. I believe this could be useful and also make sure all Integrations uses use `IntegrationRequest`

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
bentsku: We always could, but this should be caught by the provider CRUD level. Also, setting the value to `None` makes the value not go anywhere. I can add a logging statement I think!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -52,8 +52,14 @@ def __call__(
         # TODO: create helper method to render the different URI with stageVariables and parameters
         rendered_integration_uri = integration["uri"]  # use request_data_mapping["path"]
 
+        # TODO: verify the assumptions about the method with an AWS validated test
+        # if the integration method is defined and is not ANY, we can use it for the integration
+        if not (integration_method := integration["httpMethod"]) or integration_method == "ANY":
+            # otherwise, fallback to the request's method
+            integration_method = context.invocation_request["method"]
```
cloutierMat, line 59: That seems more accurate. But yes some AWS validated tests will help confirm later on. Good to have the TODO.

cloutierMat (approved):
Awesome! Thanks for addressing the comments. I like the direction we are going! and the added logs should help us and user to debug much quicker!

PR #11125, opened 2024-07-02
Increase stability of Bookstore test

silv-io:


alexrashed:
Thanks for digging into this, I think I found the root cause of the instabilities (rate limiting issues due to disabled caching): https://github.com/localstack/localstack/pull/11168
I'll close this PR for now.

PR #11129, opened 2024-07-03
APIGW NG fix routing with `ANY` request and remove stage from path

bentsku:
## Motivation
As @cloutierMat reported, when actually using the next gen APIGW provider, we would not properly route requests, because we had the full path containing the `stage` which didn't map properly to any created Werkzeug Rule. This is how APIGW represents their `path`, this is only what is after the `stage`. 

Only the `$context.path` contains the `stage` and `path`, so we can get weird snapshot with one field named `path` containing `/your-path` and one named `requestContext.path` containing `/stage-name/your-path`. 🤷‍♂️ 

We also had the issue with the `ANY` method, we would try to pass that when creating a Werkzeug Rule. 
We now create a special case where `ANY` gets mapped to `DELETE`, `GET`, `HEAD`, `OPTIONS`, `PATCH`, `POST`, and `PUT` (the supported HTTP methods of APIGW).

I was worried about having a rule order issue in Werkzeug, but this proved to not be a problem because we only return the `Resource` from the match, and we can manually get the right `resourceMethod` afterwards. 

## Changes
- create a new Werkzeug Rule to take into account the `ANY` method
- remove the stage from the path in the parsing step
- add new tests for the `ANY` method with different order of creation
- update tests to have the stage in the `Request` path

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/resource_router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/resource_router.py
@@ -1,14 +1,14 @@
 import logging
 from functools import cache
+from http import HTTPMethod
 from typing import Iterable
 
 from werkzeug.exceptions import MethodNotAllowed, NotFound
-from werkzeug.routing import Map, MapAdapter
+from werkzeug.routing import Map, MapAdapter, Rule
 
 from localstack.aws.api.apigateway import Resource
 from localstack.aws.protocol.routing import (
     GreedyPathConverter,
-    StrictMethodRule,
     path_param_regex,
     post_process_arg_name,
     transform_path_params_to_rule_vars,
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/resource_router.py
```diff
@@ -23,6 +23,35 @@
...
+                HTTPMethod.HEAD,
+                HTTPMethod.OPTIONS,
+                HTTPMethod.PATCH,
+                HTTPMethod.POST,
+                HTTPMethod.PUT,
+            }
+        else:
+            # Make sure Werkzeug's Rule does not add any other methods
+            # (f.e. the HEAD method even though the rule should only match GET)
+            self.methods = {method.upper()}
```
cloutierMat, line 52: Nice Catch!

cloutierMat (approved):
Nice fix! Thanks for addressing this. Seems like Werkzeug is still making our life easier! :partying_face:

PR #11131, opened 2024-07-03
Add documentation for different types of tests

silv-io:
## Motivation

With #11049 we switched to only execute acceptance tests for pushes onto the `master` branch. Currently, however, the suite of acceptance tests is not very extensive. 
To facilitate (correct) growth here, we need guidance on what acceptance tests exactly are and what we expect of them.

## Changes

This PR adds a new section to the contribution documentation, documenting the different types of tests we execute in our CI pipeline.
The main focus is documentation about acceptance tests, when they are executed, what is expected of them, and how new ones can be developed.

### DIFF
--- a/docs/testing/test-types/README.md
+++ b/docs/testing/test-types/README.md
@@ -1,0 +1,46 @@
+# Test Types
+
+In the LocalStack codebase we differentiate between the following test types:
+
+- Unit tests
+- Acceptance Tests
+- Integration Tests
+
+Depending on the workflow and its trigger not all of those tests are executed at once.
+For ordinary pushes to `master` we only want to execute the Unit and Acceptance tests.
+On a regular schedule, however, we want to execute all tests to have as big of a coverage of our logic as possible.
+This differentiation also educates what we expect from the different types of tests.
+
+## Unit tests
+
+As the name suggests, these tests are performed on smaller units of logic to check if they're sound and perform the operations they claim to.
+This small unit can most often be a kind of helper function inside of a larger procedure.
+These tests should be able to complete their execution very quickly, so they never contain any interaction with some kind of infrastructure.
+If you need some kind of waiting mechanism in your unit test, it is most likely that you are not writing a unit test.
+
+
+## Acceptance tests
+
+We use acceptance tests to gain a quick understanding of whether the recently pushed commit to `master` fulfils minimally viable quality criteria.
+This means that these tests do not aim at maximum coverage but instead should test that the most important functionality works.
+This in general is the entire serving infrastructure and the main features of the most used services.
+
+As these tests are executed very often we need them to be as stable, fast and relevant as possible.
+We ensure this by the following criteria:
+
+- Should perfectly conform to all the testing rules laid out [here](../README.md)
+- Does not contain wait times (e.g., for resources to spin up) averaging longer than 20 seconds
+- Should be able to run in parallel. If tests need to run together (e.g., in a scenario), they need to be added to the same test class.
+- Displays some kind of real-world usage. This can be a scenario with multiple services. 
+    - When composing these scenarios, the services should not overlap too much with already existing acceptance tests. We want to avoid redundancy
+    - Existing samples (from [our samples organization](https://github.com/localstack-samples)) can serve as a good starting point for constructing such a scenario
+- The test should have a proven track record of being stable over a month of being in the regular integration test suite.
+    - Should an acceptance test turn flaky, it will be skipped until it is fixed ([as the rules already state](../README.md))
+
+## Integration tests
+
+These tests are quite similar to the acceptance tests, but are less restrictive.
+Any acceptance test can be demoted to an integration test should it not satisfy the needs of the acceptance test suite anymore.
+However, this does not mean that integration tests do not have any quality requirements in their own right.
+Flaky integration tests can (and will) still be skipped until their flake is resolved.
+Also, they still should all conform to the testing rules.
--- a/docs/testing/test-types/README.md
+++ b/docs/testing/test-types/README.md
@@ -18,6 +18,8 @@ This small unit can most often be a kind of helper function inside of a larger p
 These tests should be able to complete their execution very quickly, so they never contain any interaction with some kind of infrastructure.
 If you need some kind of waiting mechanism in your unit test, it is most likely that you are not writing a unit test.
 
+A good example for a unit test is `tests.unit.testing.testselection.test_matching.test_service_dependency_resolving_with_dependencies`.
+It tests whether an algorithm implemented inside of a bigger implementation performs as it is expected of it. 
 
 ## Acceptance tests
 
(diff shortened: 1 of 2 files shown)


docs/testing/test-types/README.md
```diff
@@ -0,0 +1,46 @@
...
+We ensure this by the following criteria:
+
+- Should perfectly conform to all the testing rules laid out [here](../README.md)
+- Does not contain wait times (e.g., for resources to spin up) averaging longer than 20 seconds
+- Should be able to run in parallel. If tests need to run together (e.g., in a scenario), they need to be added to the same test class.
+- Displays some kind of real-world usage. This can be a scenario with multiple services. 
+    - When composing these scenarios, the services should not overlap too much with already existing acceptance tests. We want to avoid redundancy
+    - Existing samples (from [our samples organization](https://github.com/localstack-samples)) can serve as a good starting point for constructing such a scenario
+- The test should have a proven track record of being stable over a month of being in the regular integration test suite.
+    - Should an acceptance test turn flaky, it will be skipped until it is fixed ([as the rules already state](../README.md))
```
alexrashed, line 38: I added a few suggested changes to this section:
- I think the 20 seconds were a bit arbitrary, it's hard to define an exact number here. I would just state that they need to be fast.
- At the same time I think that the wording should be strict. For example, the acceptance tests should not be conform to the rules, but they have to.
- I don't think we should introduce a "graduation procedure" for acceptance tests (needs to be stable for x), but we will definitely be super-strict when it comes to skipping flaky ones.

```suggestion
- It is perfectly conform to all the testing rules laid out [here](../README.md)
- It does not contain long wait times (e.g., for resources to spin up). The acceptance tests need to be fast.
- It is fully parallelizable. If certain acceptance tests need to run together (e.g., in a scenario), they need to be added to the same test class.
- It shows some kind of real-world usage. This is usually a scenario with multiple services.
    - When composing these scenarios, the services should not overlap too much with already existing acceptance tests. We want to avoid redundancy where possible. At the same time we want to have our primary services and typical use-cases being covered.
    - Existing samples (from [our samples organization](https://github.com/localstack-samples)) can serve as a good starting point for constructing such a scenario.
- The test needs to be extremely stable.
    - Should an acceptance test turn flaky, it will be skipped until it is fixed ([as we already state in our testing rules](../README.md)).
```

alexrashed (commented):
Awesome! This is a great push towards a shared understanding of the different test types, most importantly the somewhat new acceptance tests! 💯 
I only added a comment on the definition of the acceptance tests with some reasoning on the different changes I would suggest.

I would also love to get some input from others! Is there anything this definition is missing? Is it too strict? Should it be stricter?
/cc @dominikschubert @simonrw @steffyP @joe4dev @viren-nadkarni @thrau

docs/testing/test-types/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+As these tests are executed very often we need them to be as stable, fast and relevant as possible.
+We ensure this by the following criteria:
+
+- Should perfectly conform to all the testing rules laid out [here](../README.md)
+- Does not contain wait times (e.g., for resources to spin up) averaging longer than 20 seconds
+- Should be able to run in parallel. If tests need to run together (e.g., in a scenario), they need to be added to the same test class.
+- Displays some kind of real-world usage. This can be a scenario with multiple services. 
+    - When composing these scenarios, the services should not overlap too much with already existing acceptance tests. We want to avoid redundancy
+    - Existing samples (from [our samples organization](https://github.com/localstack-samples)) can serve as a good starting point for constructing such a scenario
```
steffyP, line 36: Is there a specific reason that we highlight the localstack-samples here? Because from our initial efforts on adding scenario tests, we found that most samples are too small and use to little different resources to make great acceptance tests.

docs/testing/test-types/README.md
```diff
@@ -0,0 +1,46 @@
...
+This means that these tests do not aim at maximum coverage but instead should test that the most important functionality works.
+This in general is the entire serving infrastructure and the main features of the most used services.
+
+As these tests are executed very often we need them to be as stable, fast and relevant as possible.
+We ensure this by the following criteria:
+
+- Should perfectly conform to all the testing rules laid out [here](../README.md)
+- Does not contain wait times (e.g., for resources to spin up) averaging longer than 20 seconds
+- Should be able to run in parallel. If tests need to run together (e.g., in a scenario), they need to be added to the same test class.
+- Displays some kind of real-world usage. This can be a scenario with multiple services.
```
steffyP, line 34: I wonder if we should mention cdk-based testing somewhere as well - I think currently we don't have any official documentation for cdk-based testing yet?

steffyP, line 34: I would move the `Displays some kind of real-world usage...` to the top of the list.
```suggestion
- Displays some kind of real-world usage. This can be a scenario with multiple services involved and tested. 
```

docs/testing/test-types/README.md
silv-io: - I wanted to be very specific with the execution time. It being "very fast" or waiting times being "long" can mean something else for everyone. The 20 seconds was just a starting point for further discussion. However, we can also decide those things on a case-by-case basis.

- I agree, stricter wording is better here

- I had the concern that adding a lot of acceptance tests of which we don't know the long-term stability yet will defeat the purpose of having a stable set of acceptance tests to rely on. The track record of one month was, like the 20 seconds before, chosen as a starting point for discussion. However, if we accept that there might be some churn in the beginning of adding and skipping tests, we can also remove this requirement.

alexrashed: That's totally understandable, but I think we kind of have to decide the execution time requirements on a case-by-case basis because everything else would be a weird guess-game right now. We can distill some educated numbers once we have gained some experience with it.

Concerning the "graduation": I think we should move relatively fast with adding acceptance tests, but be rigorously strict in removing any tests that cause any kind of friction. 😄

steffyP (commented):
Great work on putting the test types into writing 🙌 
This a though undertaking and I am sure over time we will get more experience and may adapt the definition over time :)

**Here are my remarks:**
Personally, I would love to see some more concrete examples for each of the test types to make it more relatable. Especially for integration vs acceptance tests we could outline one best-practise sample (I think outlining the test intention would be enough, no need for code; or we link to existing tests in the code base).

Some questions that came to my mind - basically can we check these objectively or will this be a highly subjective observation from the reviewer:
* How to check that the test can run in parallel?
* How do we make sure that there is no (or not much) overlap with existing acceptance tests?
* Is there a hard requirement of using at least x resources? Would an acceptance test that only tests one service be ok?
* Are the "main features of the most used services" outlined somewhere? Especially for community contributors it might be hard to know.
* Are we confident that the current acceptance-tests follow all these definitions? E.g. we have DMS acceptance tests, but I don't think its one of the most-used services yet 😄

docs/testing/test-types/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Acceptance tests
+
+We use acceptance tests to gain a quick understanding of whether the recently pushed commit to `master` fulfils minimally viable quality criteria.
+This means that these tests do not aim at maximum coverage but instead should test that the most important functionality works.
+This in general is the entire serving infrastructure and the main features of the most used services.
+
+As these tests are executed very often we need them to be as stable, fast and relevant as possible.
+We ensure this by the following criteria:
```
steffyP, line 29: One thing we could still add: The test needs to be aws-validated.

docs/testing/test-types/README.md
silv-io: I think that even though they're quite small, they can still serve as some kind of starting point. Also, ideally we would also like to have samples with some kind of architectural complexity so I think the goals of acceptance tests and samples are kind of aligned here. 

However, I'll add a remark that further complexity might be needed for a sample to become an acceptance test.

docs/testing/test-types/README.md
silv-io: I think @dominikschubert might be able to chime in where we could add such documentation best so that it can be referenced here as an option of how to construct such scenarios.

docs/testing/test-types/README.md
silv-io: Depending on what kind of functionality we're going to test here it might be that we have some non-validated tests here as well. E.g., some value-add functionality. But I can add that criteria with that caveat as well.

alexrashed: You can just add that they either have to be AWS validated or LocalStack Only. We already enforce that all integration tests have to have a marker anyways, so this would basically mean that it must not be `unknown` or `needs fixing`, which is perfectly reasonable :D

docs/testing/test-types/README.md
alexrashed: I think this could maybe be mentioned here as soon as there is some guidance / documentation around the CDK testing framework it self, but for as long as this isn't there, I would not mention that here right now...

docs/testing/test-types/README.md
steffyP: I think @silv-io suggestion would be sufficient for now :) we also looked at official AWS samples, but here we also have the issue that AWS doesn't keep them up-to-date and they are not guaranteed to still work with AWS, so there might be a lot of additional work required to make them work. But they could at least be used as inspiration for use cases.

steffyP (approved):
Thanks for iterating over the suggested changes 🙌 

IMO the definition for the first iteration 🚀 
But I would to follow-up on my previous PR comments (with the questions I posted [here](https://github.com/localstack/localstack/pull/11131#pullrequestreview-2171714109)) and maybe get some more input from other people on their perception as well 🙂

alexrashed (approved):
Great push towards a unified understanding of what the different tests are! 🏅 
From my perspective, this can be merged right away.
As mentioned in the documentation itself, this is the first definition which is a great basis to iterate on as soon as we have more experience with this testing methodology. 🚀

PR #11134, opened 2024-07-03
make /_localstack/plugin endpoint extendable

thrau:
## Motivation

It's currently not easily possible for ext or other modules to add plugin managers to the output list of `/_localstack/plugins`.

This internal refactoring creates a class-level list of plugin managers that can be extended using hooks or other discovery mechanisms. Specifically this is used by ext to add the `localstack.platform.plugin` namespace.

## Changes

* the output `/_localstack/plugin` can now be extended by other components
* the init `ScriptRunners` introduced in #11098 are now listed correctly
* the `requires_license` output is now shown in the list of plugins, if the plugin has been initialized and the attribute is present
* PluginManagers that were in `/_localstack/plugins` are now imported a bit earlier (when the `PluginsResource` is created, not when it's called for the first time - this should be OK though, since all the code has been imported already through other modules anyway)

## Testing

* Run `python -m localstack.dev.run`
* Observe that `curl -s localhost.localstack.cloud:4566/_localstack/plugins` still works ;-)

### DIFF
--- a/localstack-core/localstack/services/internal.py
+++ b/localstack-core/localstack/services/internal.py
@@ -6,8 +6,8 @@ import re
 import time
 from collections import defaultdict
 from datetime import datetime
-from typing import List
 
+from plux import PluginManager
 from werkzeug.exceptions import NotFound
 
 from localstack import config, constants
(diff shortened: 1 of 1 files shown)


dfangl (approved):
Looks good! I think having a static list of plugin managers might not be super elegant, but it should do here :)

thrau:
> Looks good! I think having a static list of plugin managers might not be super elegant, but it should do here :)

Agreed. I looked into an alternative, but the solution would have required significant refactoring for ext to get access to the `PluginsResource` instance

PR #11135, opened 2024-07-03
APIGW NG add Integration Response selection and parameter mapping

bentsku:
## Motivation

This PR starts implementing the Integration Response handler, who's responsible for selecting the right `IntegrationResponse` based on the response status code (or an `errorMessage` field very specific to the `AWS` integration with Lambda, which we will tackle later). 

The handler should then do the response parameters mapping, then the VTL rendering of the body. 
We do not test the handler yet, same as the Integration Request handler, as the behavior is incomplete yet.

Instead, we can focus on the integration response selection, and the response parameters mapping. All of the unit tests have been manually tested against AWS, with the console and a SSH tunnel with localhost.run to my own small flask server to be able to return arbitrary payload from my HTTP payload. 

More informations about the different subjects of the PR:

- https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html#mapping-response-parameters
- https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-integration-settings-integration-response.html
- https://docs.aws.amazon.com/apigateway/latest/developerguide/handle-errors-in-lambda-integration.html?icmpid=apigateway_console_help

## Changes
- implement the base of the Integration Response handler, still very WIP, with comments outlining what I currently think is happening
- ported the current `selectionPattern` util to select the right integration response (thanks @dfangl for implementing it!) and unit test it
- refactor the parameter mapping a bit to share code between request and response parameter mappings (code is still pretty similar, but has a few distinct difference, we could always refactor later on)
- add the response parameter mapping logic 
- add unit tests for response parameter mapping

## TODO

What's left to do:

- [x] add unit tests for Response Parameters by validating with AWS
- [x] add unit tests for selecting the integration response

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
@@ -275,7 +275,7 @@ class ParametersMapper:
             return None
 
     @staticmethod
-    def _json_load(body: bytes | str) -> dict | list:
+    def _json_load(body: bytes) -> dict | list:
         """
         AWS only tries to JSON decode the body if it starts with some leading characters ({, [, ", ')
         otherwise, it ignores it
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
@@ -10,6 +10,7 @@ import json
 import logging
 from typing import TypedDict
 
+from localstack.http import Response
 from localstack.utils.json import extract_jsonpath
 from localstack.utils.strings import to_str
 
@@ -28,6 +29,12 @@ class RequestDataMapping(TypedDict):
     querystring: dict[str, str | list[str]]
 
 
+class ResponseDataMapping(TypedDict):
+    # Method response header parameters can be mapped from any integration response header or integration response body,
+    # $context variables, or static values.
+    header: dict[str, str]
+
+
 class ParametersMapper:
     def map_integration_request(
         self,
@@ -41,7 +48,6 @@ class ParametersMapper:
             path={},
             querystring={},
         )
-        # TODO: maybe extract functionality and re-use in `map_integration_response`
 
         for integration_mapping, request_mapping in request_parameters.items():
             integration_param_location, param_name = integration_mapping.removeprefix(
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -54,39 +60,145 @@ def map_integration_request(
...
                 )
 
-            elif request_mapping.startswith("'") and request_mapping.endswith("'"):
-                value = request_mapping.strip("'")
+            if value:
+                response_data_mapping["header"][header_name] = value
+
+        return response_data_mapping
+
+    def _retrieve_parameter_from_variables_and_static(
```
cloutierMat, line 107: Nice! That will be ready for reuse in the gateway responses as well when we get there!

localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
```diff
@@ -54,39 +60,145 @@ def map_integration_request(
...
+        :param integration_response: the Response to map parameters from
+        :return: the value to map in the ResponseDataMapping
+        """
+        if expr.startswith("body"):
+            body = integration_response.get_data() or b"{}"
+            body = body.strip()
+            try:
+                decoded_body = self._json_load(body)
+            except ValueError:
+                # x-amzn-errortype: InternalFailureException
```
cloutierMat, line 153: :pray:

tests/unit/services/apigateway/test_parameters_mapping.py
```diff
@@ -397,3 +399,228 @@ def test_default_request_mapping_casing(
...
+        mapping = mapper.map_integration_response(
+            response_parameters=response_parameters,
+            integration_response=integration_response,
+            context_variables=default_context_variables,
+            stage_variables={},
+        )
+
+        # this was validated against AWS
+        assert mapping == {
+            "header": {"body_value": "{}"},
```
cloutierMat, line 475: It's back :ghost:

cloutierMat (approved):
:rocket: Awesome PR. Nice structure to the handler.

I still have some doubt/questions around the `AWS` lambda integration and the differences in how it handles success/error, but I guess we will discover more as we complete it's handler. For now, it seems logic to proceed this way, as the error end up in the body for the mapping, so we need it here.

Really good and clear tests. Congrats!

bentsku:
> I still have some doubt/questions around the AWS lambda integration and the differences in how it handles success/error, but I guess we will discover more as we complete it's handler. For now, it seems logic to proceed this way, as the error end up in the body for the mapping, so we need it here.

@cloutierMat yes, it's a bit specific, but it seems it's well documented in the documentation, and it seems it really happens at the Integration Response level, as they talk about the fact that the Integration will always return `200`, success or failure. It seems like a small hacky solution on AWS side too! But we can see better later, yes. It might be that the `AWS` lambda integration might have some specificity. Surprise later 😄

PR #11138, opened 2024-07-03
Apigw ng request template mapping

cloutierMat:
## Motivation

Adding VTL Template Mapping to our Integration request handler and port `ApiGatewayVtlTemplate` class to NG provider.

The request template allows to create a request body based on the provided VTL template. You can also use it to `#set` requests override the path parameters, headers and query strings. See: https://docs.aws.amazon.com/apigateway/latest/developerguide/models-mappings.html

When no templates are found based on content-type, an `UnsupportedMediaTypeError` can be raised depending on the set passthrough behaviour.

## Changes

- Implements `ApiGatewayVtlTemplate`, which is ported and simplified from legacy `templates.py`. 
- adds logic for and exception for passthrough behavior
- augments request template with the request overrides

## TODO

What's left to do:

- [x] Implement unit tests for `ApiGatewayVtlTemplate`. Port from legacy to prevent regressions as we move forward
- [x] Implement unit tests for `IntegrationRequestHandler`. There might be more tests needed as parameters mapping was not tested yet

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -1,5 +1,4 @@
 import logging
-from http import HTTPMethod
 
 from localstack.aws.api.apigateway import Integration, IntegrationType
 from localstack.http import Response
@@ -79,13 +78,6 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         if not (integration_method := integration["httpMethod"]) or integration_method == "ANY":
             # otherwise, fallback to the request's method
             integration_method = context.invocation_request["method"]
-        # AWS doesn't send a body for these methods. Mock needs the body to get the status code
-        # TODO this might need to be done in the integrations?
-        if (
-            integration_method in [HTTPMethod.GET, HTTPMethod.HEAD, HTTPMethod.OPTIONS]
-            and not integration_type == IntegrationType.MOCK
-        ):
-            body = b""
 
         integration_request = IntegrationRequest(
             http_method=integration_method,
@@ -116,7 +108,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         template: str,
     ) -> tuple[bytes, ContextVarsRequestOverride]:
         request: InvocationRequest = context.invocation_request
-        body = request.get("body")
+        body = request["body"]
 
         if not template:
             return body, {}
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -1,5 +1,4 @@
 import logging
-from enum import Enum
 
 from localstack.aws.api.apigateway import Integration, IntegrationType
 from localstack.http import Response
@@ -20,7 +19,7 @@ from ..variables import (
 LOG = logging.getLogger(__name__)
 
 
-class PassthroughBehavior(Enum):
+class PassthroughBehavior(str):
     # TODO maybe this class should be moved where it can also be used for validation in
     #  the provider when we switch out of moto
     WHEN_NO_MATCH = "WHEN_NO_MATCH"
@@ -66,7 +65,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         body, request_override = self.render_request_template_mapping(
             context=context, template=request_template
         )
-        merge_recursive(request_override, request_data_mapping)
+        merge_recursive(request_override, request_data_mapping, overwrite=True)
 
         # TODO: extract the code under into its own method
 
@@ -108,32 +107,33 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         template: str,
     ) -> tuple[str, ContextVarsRequestOverride]:
         request: InvocationRequest = context.invocation_request
-        body = str(request.get("body", ""))
+        body = request.get("body", b"").decode()
 
         if not template:
             return body, {}
 
-        variables = MappingTemplateVariables(
-            context=context.context_variables,
-            stageVariables=context.stage_variables or {},
-            input=MappingTemplateInput(
-                body=body,
-                params=MappingTemplateParams(
-                    path=request.get("path_parameters"),
-                    querystring=request.get("query_string_parameters", {}),
-                    header=request.get("headers", {}),
+        return self._vtl_template.render_request(
+            template=template.strip(),
+            variables=MappingTemplateVariables(
+                context=context.context_variables,
+                stageVariables=context.stage_variables or {},
+                input=MappingTemplateInput(
+                    body=body,
+                    params=MappingTemplateParams(
+                        path=request.get("path_parameters"),
+                        querystring=request.get("query_string_parameters", {}),
+                        header=request.get("headers", {}),
+                    ),
                 ),
             ),
         )
-        result = self._vtl_template.render_vtl(template=template.strip(), variables=variables)
-        return result, variables.get("requestOverride")
 
     def get_request_template(self, integration: Integration, request: InvocationRequest) -> str:
         """
         Attempts to return the request template.
         Will raise UnsupportedMediaTypeError if there are no match according to passthrough behavior.
         """
-        request_templates = integration.get("requestTemplates")
+        request_templates = integration.get("requestTemplates", {})
         passthrough_behavior = integration.get("passthroughBehavior")
         # If content-type is not provided aws assumes application/json
         content_type = request["raw_headers"].get("Content-Type", "application/json")
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -1,13 +1,30 @@
 import logging
+from enum import Enum
 
 from localstack.aws.api.apigateway import Integration, IntegrationType
 from localstack.http import Response
 
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
-from ..context import IntegrationRequest, RestApiInvocationContext
+from ..context import IntegrationRequest, InvocationRequest, RestApiInvocationContext
+from ..gateway_response import UnsupportedMediaTypeError
 from ..parameters_mapping import ParametersMapper, RequestDataMapping
+from ..template_mapping import ApiGatewayVtlTemplate
+from ..variables import (
+    ContextVarsRequestOverride,
+    MappingTemplateInput,
+    MappingTemplateParams,
+    MappingTemplateVariables,
+)
 
 LOG = logging.getLogger(__name__)
+
+
+class PassthroughBehavior(Enum):
+    # TODO maybe this class should be moved where it can also be used for validation in
+    #  the provider when we switch out of moto
+    WHEN_NO_MATCH = "WHEN_NO_MATCH"
+    WHEN_NO_TEMPLATES = "WHEN_NO_TEMPLATES"
+    NEVER = "NEVER"
 
 
 class IntegrationRequestHandler(RestApiGatewayHandler):
@@ -18,6 +35,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
 
     def __init__(self):
         self._param_mapper = ParametersMapper()
+        self._vtl_template = ApiGatewayVtlTemplate()
 
     def __call__(
         self,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -34,18 +52,20 @@ def __call__(
...
             context, integration_request_parameters
         )
 
-        # TODO: work on TemplateMappings with VTL to render the body of the request
-        #  it might populate the context requestOverride too, maybe deepcopy the ContextVariables because it will
-        #  get mutated, or pop it directly? might just be better!
+        body, request_override = self.render_request_template_mapping(
+            context=context, template=request_template
+        )
+        merge_recursive(request_override, request_data_mapping, overwrite=True)
```
bentsku, line 68: nice find for this util!

cloutierMat:
@bentsku We might need more tests in the handler to make sure that both step work good together.

I didn't want to test all scenarios for now as I feel most of the edge cases should be tested either in the unit tests for `ParametersMapper` or `ApiGatewayVtlTemplate` or in AWS validated tests later on, and focus the handler testa on asserting a correct order of operations.

Happy to collaborate and align our testing efforts! :metal:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -80,3 +100,64 @@ def get_integration_request_data(
...
             stage_variables=context.stage_variables,
         )
+
+    def render_request_template_mapping(
+        self,
+        context: RestApiInvocationContext,
+        template: str,
+    ) -> tuple[str, ContextVarsRequestOverride]:
+        request: InvocationRequest = context.invocation_request
+        body = request.get("body", b"").decode()
```
bentsku, line 110: nit: I don't think we should decode the body before checking if there's a template, because then we need to re-encode it. 

Maybe we could do the decoding later, and always return the body as `bytes`, so we could return the body as is. Then, after the rendering in this function you can encode it.

Then we don't need to encode it back in the caller method?

localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
```diff
@@ -0,0 +1,218 @@
...
+        result = self.render_vtl(template=template.strip(), variables=vars)
+        return result, vars["context"]["requestOverride"]
+
+    def render_response(
+        self, template: str, variables: MappingTemplateVariables
+    ) -> tuple[str, ContextVarsResponseOverride]:
+        pass
+
+    # TODO Maybe we don't need those methods and they should belong on the integration response handler?
+    #  And we should raise the appropriate exception from the gateway responses.
```
bentsku, line 187: Agreed here, I think we will update this, I wonder when it is actually executed. Interesting find!

localstack-core/localstack/services/apigateway/next_gen/execute_api/variables.py
```diff
@@ -179,3 +185,20 @@ class LoggingContextVariables(TypedDict, total=False):
...
+
+class MappingTemplateInput(TypedDict):
+    body: str
+    params: MappingTemplateParams
+
+
+class MappingTemplateVariables(TypedDict, total=False):
+    context: ContextVariables
+    input: MappingTemplateInput
+    stageVariables: dict[str, str]
```
bentsku, line 204: Not too sure this should be here, maybe the file itself is too restrictively named, but I would maybe put this in the `template_mapping.py`, as it's only related to it?

tests/unit/services/apigateway/test_handler_integration_request.py
```diff
@@ -0,0 +1,265 @@
...
+    headers = Headers(header_dict)
+    return InvocationRequest(
+        http_method=HTTPMethod.POST,
+        raw_path="/test/test-path-value",
+        path="/test/test-path-value",
+        path_parameters={"path": "test-path-value"},
+        query_string_parameters={"qs": "test-qs-value"},
+        raw_headers=headers,
+        headers=dict(headers),
+        multi_value_query_string_parameters={"qs": ["ignored-value", "test-qs-value"]},
```
bentsku, line 94: I believe in AWS, in their context, if you have `multi_value_query_string_parameters={"qs": ["val1", "val2"]}` then `query_string_parameters={"qs": "val1"}`. It's really weird, but it is how it is. 

I wonder if should try to use the parse part of the handler here for a request, that might avoid duplication of the headers and such? We could try refactoring previous tests with it and see how that goes. 
We could manually invoke `InvocationRequestParser().create_invocation_request`.

tests/unit/services/apigateway/test_handler_integration_request.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def test_request_override(
+        self, integration_request_handler, default_invocation_request, create_context
+    ):
+        context = create_context(request=default_invocation_request)
+        context.resource_method["methodIntegration"]["requestParameters"] = {
+            "integration.request.path.path": "method.request.path.path",
+            "integration.request.querystring.qs": "method.request.querystring.qs",
+            "integration.request.header.header": "method.request.header.header",
+        }
```
bentsku, line 245: what happens if you were to use the following:
```python
"integration.request.querystring.qs": "method.request.multivaluequerystring.qs",
```
and then overrides? does it just erase the list and sets only one value after the override? I'd say so, that would be pretty logic!

localstack-core/localstack/services/apigateway/next_gen/execute_api/variables.py
```diff
@@ -8,7 +8,7 @@ class ContextVarsAuthorizer(TypedDict, total=False):
     # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html
     claims: Optional[dict[str, str]]
     """Claims returned from the Amazon Cognito user pool after the method caller is successfully authenticated"""
-    principal_id: Optional[str]
+    principalId: Optional[str]
```
bentsku, line 11: nice catch!

bentsku (commented):
Wow! Impressive set of change! Agreed about the handler, to me it makes lots of sense to do this way. 👍 

Just have a comment about the decoding/re-encoding of the body, I believe we could avoid the little dance by always returning `bytes` from `render_request_template_mapping`. 

Great work once again, nice set of tests, this is really taking shape now! I'll hurry to complete the `IntegrationResponse` now so that we can continue from there with the VTL templating too. 

Also a small comment about the `InvocationRequest` in the tests, I wonder if we should just use raw `Request` object and parse them directly to simplify our work. 🤔

localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
cloutierMat: They were part on the ResponseTemplate. It was rendering and then trying to load the Json. But I feel like it shouldn't be a responsibility of the template, but instead of the integration response.

bentsku: Do you think there's a similar thing for IntegrationRequest, and that it checks if the rendered body is valid or not before sending it down?

tests/unit/services/apigateway/test_handler_integration_request.py
cloutierMat: Good question.. I should play with that a bit!

localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
cloutierMat: I will have to double check with AWS, but our current implementation was raising an error

localstack-core/localstack/services/apigateway/next_gen/execute_api/variables.py
cloutierMat: Agreed, I wasn't too happy to put it there either. `template_mapping.py` seems better! :+1:

tests/unit/services/apigateway/test_handler_integration_request.py
cloutierMat: That is a godd idea! Let me see what I can do

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
bentsku: Oh, also, forgot to add: `to_str` and `to_bytes` utils take care of the default encoding, and are similar to calling `.decode()` and `.encode()`.

cloutierMat: Thanks for the tip.
Updated

localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
cloutierMat: I double checked for the requestTemplate only.
Everything is sent as is. except for GET, OPTIONS, HEAD, an empty string is sent instead.

I updated the handler to reflect

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -77,13 +79,20 @@ def __call__(
         if not (integration_method := integration["httpMethod"]) or integration_method == "ANY":
             # otherwise, fallback to the request's method
             integration_method = context.invocation_request["method"]
+        # AWS doesn't send a body for these methods. Mock needs the body to get the status code
+        # TODO this might need to be done in the integrations?
```
bentsku, line 83: I agree with the comment, I think it can/should be up to the integration to use the body or not 👍especially since those have strong control about how they create their request, and we can see `MOCK` receives a body somehow.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -105,20 +114,20 @@ def render_request_template_mapping(
         self,
         context: RestApiInvocationContext,
         template: str,
-    ) -> tuple[str, ContextVarsRequestOverride]:
+    ) -> tuple[bytes, ContextVarsRequestOverride]:
         request: InvocationRequest = context.invocation_request
-        body = request.get("body", b"").decode()
+        body = request.get("body")
```
bentsku, line 119: was the removal of the fallback value on purpose? I guess `body` maybe can't really empty? should we access it directly it without `.get()` to be sure we never have `None`?

tests/unit/services/apigateway/test_handler_integration_request.py
```diff
@@ -81,20 +83,17 @@ def _create_context(
...
-        raw_path="/test/test-path-value",
-        path="/test/test-path-value",
-        path_parameters={"path": "test-path-value"},
-        query_string_parameters={"qs": "test-qs-value"},
-        raw_headers=headers,
-        headers=dict(headers),
-        multi_value_query_string_parameters={"qs": ["ignored-value", "test-qs-value"]},
-        multi_value_headers=header_dict,
-        body=b"",
+    request = InvocationRequestParser().create_invocation_request(
```
bentsku, line 86: nice, very neat 👌

tests/unit/services/apigateway/test_handler_integration_request.py
cloutierMat: I have looked and yes the override always overrides. I was not able to create a `join` scenario. I have also added a test for it

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -77,13 +79,20 @@ def __call__(
         if not (integration_method := integration["httpMethod"]) or integration_method == "ANY":
             # otherwise, fallback to the request's method
             integration_method = context.invocation_request["method"]
+        # AWS doesn't send a body for these methods. Mock needs the body to get the status code
+        # TODO this might need to be done in the integrations?
+        if (
+            integration_method in [HTTPMethod.GET, HTTPMethod.HEAD, HTTPMethod.OPTIONS]
+            and not integration_type == IntegrationType.MOCK
+        ):
+            body = b""
```
cloutierMat, line 88: @bentsku I added this section after validating it's behavior against aws.

bentsku: Ah, I answered here too 😄 https://github.com/localstack/localstack/pull/11138#discussion_r1664991116

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
bentsku: It could also lead to a bit of code duplication if they all need to check for the HTTP request before sending however. If we see that's the case, maybe we could migrate it back here? We can also leave it as is for now, I'm not too opinionated about this!

bentsku (approved):
Thanks for addressing the comments! 🚀 very nice changes, I like the parsing of the request in the test, makes it a bit simpler to not juggle with the headers kind, we might be able to simplify at some point... maybe the `headers` field of the context is only used in `AWS_PROXY`, so we might be able to remove it.

Really nice PR, really neat and well organized! And a lot of added tests! 🚀

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
cloutierMat: I guess it can't be, I wasn't totally sure but I guess it can't

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
cloutierMat: I will remove that and the tests with it then!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
cloutierMat: The fallback was originally to protect the `decode` but since using `to_str` the fallback was no longer required

bentsku: Ah, I see! I wasn't sure if the VTL template could take `None` for the `input.body` or not, but I guess it won't be none and just `b""`!

PR #11139, opened 2024-07-04
Use Java 21

viren-nadkarni:


Dockerfile
```diff
@@ -186,12 +149,23 @@ RUN make entrypoints
...
     chown -R localstack:localstack /usr/lib/localstack && \
     chmod -R 777 /usr/lib/localstack
 
+# Set up Java
+ENV JAVA_HOME /usr/lib/localstack/java/11
+RUN { \
+        echo '#!/bin/sh'; echo 'set -e'; echo; \
+        echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
+    } > /usr/local/bin/docker-java-home \
+    && chmod +x /usr/local/bin/docker-java-home
```
alexrashed, line 165: Is this code here still necessary? I can't really see where `docker-java-home` is being used...

Dockerfile
```diff
@@ -186,12 +149,23 @@ RUN make entrypoints
...
+RUN ln -s $JAVA_HOME/bin/java /usr/bin/java
+ENV PATH="${PATH}:${JAVA_HOME}/bin"
```
alexrashed, line 167: q: Is there a chance to completely avoid setting Java home globally? We want to be able to switch between versions, and this could be a pitfall in such a case...
This would mean that, when using Java, you would have to get `JAVA_HOME` from the package first and set it in the environment when calling the `java` command (which would make it explicit that it can vary and there are actually only very few usages of Java).

Dockerfile
```diff
@@ -186,12 +149,23 @@ RUN make entrypoints
 RUN --mount=type=cache,target=/root/.cache \
     --mount=type=cache,target=/var/lib/localstack/cache \
     source .venv/bin/activate && \
+    python -m localstack.cli.lpm install java --version 11 && \
```
alexrashed, line 152: q / nit: If we can manage to avoid setting the JAVA_HOME, I think we should avoid setting the version here. We should just install the default version (which is 11 now anyways, but we would dynamically move to new defaults in the future).

Dockerfile
```diff
@@ -1,34 +1,6 @@
...
-java.security.jgss,jdk.security.auth,\
-# Elasticsearch 7+ crashes without Thai Segmentation support
-jdk.localedata --include-locales en,th \
-    --compress 2 --strip-debug --no-header-files --no-man-pages --output /usr/lib/jvm/java-11 && \
-  cp ${JAVA_HOME}/bin/javac /usr/lib/jvm/java-11/bin/javac && \
-  cp -r ${JAVA_HOME}/include /usr/lib/jvm/java-11/include && \
-  mv /usr/lib/jvm/java-11/lib/modules /usr/lib/jvm/java-11/lib/modules.bk; \
-  cp -r ${JAVA_HOME}/lib/* /usr/lib/jvm/java-11/lib/; \
-  mv /usr/lib/jvm/java-11/lib/modules.bk /usr/lib/jvm/java-11/lib/modules; \
-  rm -rf /usr/bin/java ${JAVA_HOME} && ln -s /usr/lib/jvm/java-11/bin/java /usr/bin/java
```
alexrashed, line 28: This optimization saves ~55 MB (as you mentioned in the PR description), so this change would increase the image size by ~14%. Is there a chance we could implement this optimization in the installer? Or are there any other ways to optimize the size of the Java package?

localstack-core/localstack/packages/java.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Default version if not specified
+DEFAULT_JAVA_VERSION = "11"
+
+# Supported Java LTS versions mapped with Eclipse Temurin build semvers
+JAVA_VERSIONS = {
+    "8": "8u422-b05",
+    "11": "11.0.24+8",
+    "17": "17.0.12+7",
+    "21": "21.0.4+7",
```
alexrashed, line 19: With the exact versions, we are loosing the automated updates (by dependabot for the Dockerfile updates, and by Debian repo updates for other versions).
Is there a chance to use the `latest` release link for a specific version to always pick the latest revision / release from temurin?

localstack-core/localstack/packages/java.py
```diff
@@ -0,0 +1,70 @@
...
+}
+
+JRE_DISTRIB_URL = "https://github.com/adoptium/temurin{version}-binaries/releases/download/{tag_slug}/OpenJDK{version}U-jre_{arch}_{os}_hotspot_{semver_safe}.tar.gz"
+
+
+class JavaPackageInstaller(ArchiveDownloadAndExtractInstaller):
```
alexrashed, line 25: nit: Since this package installer is introduced here, and might be removed from the `Dockerfile` in the future, it might make sense to call this installer in the `_prepare_installation` steps of installers which depend on Java (but we can also do that in a later iteration when we actually want to get rid of the JRE).

alexrashed (commented):
Wow, this is a great step forward to upgrading Java, and maybe in the future really just installing it on demand for the few services that depend on it!
I think we need to find a solution for the pretty big size increase of the image due to the change, and I added some comments to other things we could maybe optimize from the beginning (for example using latest patch versions).

Dockerfile
viren-nadkarni: I wasn't sure whether or not to remove it as it could be used in non-LS or bigdata images.

For now, removed it in c108cb5c2, we could revert if required

Dockerfile
viren-nadkarni: With a325dfc79, the installer uses [jlink](https://docs.oracle.com/javase/9/tools/jlink.htm) to optimise the JRE as before. The install footprint post-optimisation is as follows:

```
 8  197M
11   62M
17   66M
21   75M
```
Java 8 is not optimised because it does not include jlink, which was introduced in Java 9.

Dockerfile
viren-nadkarni: Yes, this should be possible. But first we need to update all the Java components with explicit installer dependencies and make them set the relevant `JAVA_HOME`. This is planned in a follow up iteration.

Dockerfile
viren-nadkarni: As mentioned in another comment, the ideal situation is where all LPM packages have explicit Java dependencies. This will let us get rid of the above default installation of Java 11 altogether. This is planned for later iterations.

localstack-core/localstack/packages/java.py
viren-nadkarni: GitHub [allows this](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release):


> To link directly to a download of your latest release asset that was manually uploaded, the suffix is `/releases/latest/download/asset-name.zip`.

But unfortunately the `asset-name.zip` as published by Adoptium contains a semver. We can't remove the above hardcoded semvers without adding additional logic to query and filter the assets.

Adoptium publishes Java builds as GitHub release entries (eg. https://github.com/adoptium/temurin8-binaries). They don't publish a 'latest stable' release unfortunately. Things are mixed up with beta versions. Their [download page](https://adoptium.net/en-GB/temurin/releases/?version=8) uses an [API endpoint](https://api.adoptium.net/v3/assets/latest/8/hotspot?os=linux&architecture=x64).

In the earlier approach, we had a hardcoded Docker image. I believe this keeps the same behaviour, but I admit automatic bumps would be great

localstack-core/localstack/packages/java.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _download_url_latest_release(self) -> str:
+        """
+        Return the download URL for latest stable JDK build.
+        """
+        endpoint = (
+            f"https://api.adoptium.net/v3/assets/latest/{self.version}/hotspot?"
+            f"os=linux&architecture={self.arch}&image_type=jdk"
+        )
+        response = requests.get(endpoint, headers={"user-agent": "example/0.0.0"}).json()
```
alexrashed, line 103: q: The user agent setting here is a bit unorthodox. Is there a reason why you explicitly set it here? I think - if possible - you should either not set the user agent at all (fall back to `requests`), or set the LocalStack version as user agent.

alexrashed (approved):
Thanks a lot for addressing all the comments and for cleaning up along the way (f.e. by removing the `docker-java-home`)!
The changes are looking great! Especially the automatic latest-version lookup! 💯 
I only had one minor nitpick / question concerning the user-agent, but nothing blocking a release.

localstack-core/localstack/packages/java.py
viren-nadkarni: Yes, this could've been more elegant. For whatever reason, the Adoptium API denies service to the requests client

```
$ curl 'https://api.adoptium.net/v3/assets/latest/11/hotspot?os=linux&architecture=x64&image_type=jdk' -H user-agent:python-requests/2.32.3
<html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
<hr><center>Microsoft-Azure-Application-Gateway/v2</center>
</body>
</html>

$ curl 'https://api.adoptium.net/v3/assets/latest/11/hotspot?os=linux&architecture=x64&image_type=jdk' -H user-agent:
[
    {
        "binary": {
            "architecture": "x64",
...
```

In 680a58a65, I define and use a standard LS user-agent string.

PR #11140, opened 2024-07-04
APIGW NG implement URI rendering

bentsku:
## Motivation
Throughout APIGW invocation, there are several places where you can use stageVariables inside some URI and ARN. 
We also need to interpolate path parameters inside the Integration URI

This PR implements 2 utils, to render stage variables following the VTL syntax, and another one to render path parameters, only in curly braces.
The Integration Request handler combines those 2 utils to render the Integration URI. 

More informations about stage variables here:
https://docs.aws.amazon.com/apigateway/latest/developerguide/aws-api-gateway-stage-variables-reference.html#stage-variables-in-integration-HTTP-uris

We will also use the utils inside the Integration for the credentials, and in the cognito authorizer to render the UserPool. 

## Changes
- implement stage variables rendering util
- implement path parameters rendering util
- add unit tests for the URI rendering
- renamed a variable inside the template mappings due to a PyCharm linting, as we were overriding the builtin `vars`

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -72,6 +72,9 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
 
         # looks like the stageVariables rendering part is done in the Integration part in AWS
         # but we can avoid duplication by doing it here for now
+        # TODO: if the integration if of AWS Lambda type and the Lambda is in another account, we cannot render
+        #  stageVariables. Work on that special case later (we can add a quick check for the URI region and set the
+        #  stage variables to an empty dict)
         rendered_integration_uri = self.render_integration_uri(
             uri=integration["uri"],
             path_parameters=request_data_mapping["path"],
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -8,6 +8,7 @@ from localstack.utils.strings import to_bytes, to_str
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
 from ..context import IntegrationRequest, InvocationRequest, RestApiInvocationContext
 from ..gateway_response import UnsupportedMediaTypeError
+from ..helpers import render_uri_with_path_parameters, render_uri_with_stage_variables
 from ..parameters_mapping import ParametersMapper, RequestDataMapping
 from ..template_mapping import (
     ApiGatewayVtlTemplate,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -66,12 +67,16 @@ def __call__(
         body, request_override = self.render_request_template_mapping(
             context=context, template=request_template
         )
+        # TODO: log every override that happens afterwards (in a loop on `request_override`)
```
cloutierMat, line 70: Good point. I missed that one!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -66,12 +67,16 @@ def __call__(
...
-
-        # TODO: create helper method to render the different URI with stageVariables and parameters
-        rendered_integration_uri = integration["uri"]  # use request_data_mapping["path"]
+        # looks like the stageVariables rendering part is done in the Integration part in AWS
+        # but we can avoid duplication by doing it here for now
+        rendered_integration_uri = self.render_integration_uri(
+            uri=integration["uri"],
+            path_parameters=request_data_mapping["path"],
+            stage_variables=context.stage_variables,
+        )
```
cloutierMat, line 79: I guess we are ignoring the cross-account thing for now! :wink:

tests/unit/services/apigateway/test_handler_integration_request.py
```diff
@@ -266,6 +263,33 @@ def test_multivalue_mapping(
...
+    def test_integration_uri_path_params_undefined(
+        self, integration_request_handler, default_invocation_request, create_context
+    ):
+        context = create_context(request=default_invocation_request)
+        context.resource_method["methodIntegration"]["requestParameters"] = {
+            "integration.request.path.path": "method.request.path.wrongvalue",
+        }
+        context.resource_method["methodIntegration"]["uri"] = "https://example.com/{path}"
+        integration_request_handler(context)
+        assert context.integration_request["uri"] == "https://example.com/{path}"
```
cloutierMat, line 275: Has this behavior been tested against AWS? Should we raise an error or let the integration fail?

tests/unit/services/apigateway/test_uri_interpolation.py
```diff
@@ -0,0 +1,128 @@
+import pytest
+
+from localstack.services.apigateway.next_gen.execute_api.helpers import (
+    render_uri_with_path_parameters,
+    render_uri_with_stage_variables,
+)
+
+
+class TestUriInterpolationStageVariables:
+    @pytest.mark.parametrize(
```
cloutierMat, line 10: Nice use of parametrize, really quick and easy to see what is going on!

tests/unit/services/apigateway/test_uri_interpolation.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        assert "{path}" in rendered
+
+    def test_uri_interpolating_with_no_variable(self):
+        uri_with_only_braces = "https://test.domain.example/root/${path}"
+        stage_variables = {}
+        rendered = render_uri_with_stage_variables(
+            uri=uri_with_only_braces, stage_variables=stage_variables
+        )
+        assert rendered == "https://test.domain.example/root/"
```
cloutierMat, line 90: Nice I was wondering if that was the result we would get! :+1: 
I assume the same happens if you use `${stageVariables.path}`?

cloutierMat (commented):
Nice PR! Well tested!

I just left a question about how to fail when a path still contains `{path}`, but nothing blocking. We can always get back to it when we do the integration.

cloutierMat (approved):

tests/unit/services/apigateway/test_handler_integration_request.py
bentsku: It has been tested yes! It is the integration itself which is failing, it says the URI contains invalid characters or something similar, I guess it's up to the HTTP client then to validate.

tests/unit/services/apigateway/test_uri_interpolation.py
bentsku: Ah! Good catch! This is actually wrong, this should be the result when using `${stageVariables.path}`, I believe if we were to use `${path}` it might just try to interpolate it as a path param, so if no path param named `path` it will leave it to `${path}`, otherwise might render `$<path-param>`, I need to try it... oops 😬

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
bentsku: Added for a comment/TODO for it 😄

PR #11142, opened 2024-07-04
Apigw ng add x header to gateway responses

cloutierMat:
## Motivation

This PR adds `x-amzn-errortype` header to all gateway responses.

These are not documented and the value of the token isn't consistent from error to error. The `DEFAULT_XXX` errors appears to be raised with different header value depending on the cause of the error.

## Changes

- Add `x_header` to `BaseGatewayException`
- Add the header to the response in the Gateway Exception Handler
- Populate the value for all errors. Only some were validated using the console or curl. The others have been marked with a TODO so as to avoid confusion in the future. As we get more snapshot tests going, we should be able to populate more of them
- Some unit tests were moved out of `/unit/apigateway.py` to help separating the legacy tests with the NG tests.

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway_response.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway_response.py
@@ -58,25 +58,25 @@ class InternalServerError(Default5xxError):
 
 class AccessDeniedError(BaseGatewayException):
     type = GatewayResponseType.ACCESS_DENIED
-    # Header Untested
+    # TODO validate this header with aws validated tests
     x_header = "AccessDeniedException"
 
 
 class ApiConfigurationError(BaseGatewayException):
     type = GatewayResponseType.API_CONFIGURATION_ERROR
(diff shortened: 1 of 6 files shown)


bentsku (approved):
LGTM! Nice feature, this is great to see the exception handler being extended! 

I just have a comment regarding the naming of the new field.
In the serializer, I saw that some services also return the `X-Amzn-Errortype` (JSON services it seems). In `localstack.aws.protocol.serializer.JSONResponseSerializer._serialize_error`, for example, it seems we're calling the field used to populate that header `exception.code`, which seems to be something global across AWS for exception message. Do you think that'd make sense to rename it the same?

cloutierMat:
Agreed! Nice find. `code` will work perfect.

I will also fix the capitalization, I copied it from curl... low caps :smile:  apigw seems to have it as `x-amzn-ErrorType` instead of `X-Amzn-Errortype`. Probably won't matter and it seems all over the place in the code as well, so I will just copy from the aws console at the moment!

PR #11143, opened 2024-07-05
APIGW NG add HTTP and HTTP_PROXY integration

bentsku:
## Motivation
This PR implements the `HTTP` and `HTTP_PROXY` integrations. The previous work we have done make those extremely simple to implement and understand, and it shows how much the current work is going to simplify APIGW. 

The PR also contains a few changes across the board, because we can now test full end-to-end integration tests: the `tests/aws/services/apigateway/test_apigateway_http.py` test file is fully green with the new PR, with improved parity. 

## Changes
- implemented the `HTTP` and `HTTP_PROXY` integrations
- change some typing around some existing contexts, as those were not really `Optional` and it helps with making sure we're not lacking data once we're in the integration
- remove the `legacy` handler as we can now enable full end-to-end tests
- remove the full clear of headers when receiving a response from an integration that is not of `_PROXY` type, but instead have a mechanism to keep some
- return the integration response/method response status code for the response
- fix the `path` value, as we would replace the `/<stage>` multiple times through, and if you had a path looking `/test/test` where `test` was your stage and `test` just a regular resource, we're replace everything
- add default values for `Accept` and `Content-Type`, behavior checked with AWS and add test for it
- remove some snapshot skips when running the HTTP integrations integration tests with the new provider
- update some unit tests to be a bit simpler and avoid mistakes

## Testing
Set `PROVIDER_OVERRIDE_APIGATEWAY=next_gen` and run all tests in `tests/aws/services/apigateway/test_apigateway_http.py`, they now all run! 🥳 

## TODO

What's left to do:

- [x] PR description

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
@@ -54,7 +54,7 @@ class IntegrationResponseHandler(RestApiGatewayHandler):
         )
 
         # we then need to apply Integration Response parameters mapping, to only return select headers
-        response_parameters = integration_response["responseParameters"] or {}
+        response_parameters = integration_response.get("responseParameters") or {}
         response_data_mapping = self.get_method_response_data(
             context=context,
             response=response,
@@ -68,11 +68,15 @@ class IntegrationResponseHandler(RestApiGatewayHandler):
         # here we update the `Response`. We basically need to remove all headers and replace them with the mapping, then
         # override them if there are overrides.
         # for the body, it will maybe depend on configuration?
+        for key in list(response.headers.keys(lower=True)):
+            if key not in ("connection", "content-type"):
+                del response.headers[key]
 
-        response.headers.clear()
         # there must be some default headers set, snapshot those?
         if mapped_headers := response_data_mapping.get("header"):
             response.headers.update(mapped_headers)
+
+        response.status_code = int(integration_response["statusCode"])
 
     def get_method_response_data(
         self,
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
@@ -44,6 +44,10 @@ class BaseRestApiHttpIntegration(RestApiIntegration):
             "user-agent": f"AmazonAPIGateway_{context.api_id}",
         }
 
+    @staticmethod
+    def _get_integration_timeout(integration: Integration) -> float:
+        return int(integration.get("timeoutInMillis", 29000)) / 1000
+
 
 class RestApiHttpIntegration(BaseRestApiHttpIntegration):
     """
@@ -70,8 +74,9 @@ class RestApiHttpIntegration(BaseRestApiHttpIntegration):
         if method not in NO_BODY_METHODS:
             request_parameters["data"] = integration_req["body"]
 
-        # TODO: configurable timeout (29 by default)
-        # request_parameters["timeout"] = 29
+        # TODO: configurable timeout (29 by default) (check type and default value in provider)
+        # integration: Integration = context.resource_method["methodIntegration"]
+        # request_parameters["timeout"] = self._get_integration_timeout(integration)
         # TODO: check for redirects
         # request_parameters["allow_redirects"] = False
 
@@ -127,6 +132,9 @@ class RestApiHttpProxyIntegration(BaseRestApiHttpIntegration):
         # TODO: validate this for HTTP_PROXY
         if method not in NO_BODY_METHODS:
             request_parameters["data"] = invocation_req["body"]
+
+        # TODO: configurable timeout (29 by default) (check type and default value in provider)
+        # request_parameters["timeout"] = self._get_integration_timeout(integration)
 
         request_response = requests.request(**request_parameters)
 
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
@@ -1,4 +1,19 @@
+from http import HTTPMethod
+
+import requests
+from werkzeug.datastructures import Headers
+
+from localstack.http import Response
+
+from ..context import IntegrationRequest, InvocationRequest, RestApiInvocationContext
+from ..helpers import render_uri_with_path_parameters
 from .core import RestApiIntegration
+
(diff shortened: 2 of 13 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
```diff
@@ -1,20 +1,137 @@
...
     """
     This is a simplified REST API integration responsible to send a request to another HTTP API by proxying it almost
     directly.
     https://docs.aws.amazon.com/apigateway/latest/developerguide/setup-http-integrations.html#api-gateway-set-up-http-proxy-integration-on-proxy-resource
     """
 
     name = "HTTP_PROXY"
+
+    def invoke(self, context: RestApiInvocationContext) -> Response:
+        invocation_req: InvocationRequest = context.invocation_request
```
bentsku, line 97: I think this might be wrong, we need to use the integration method, with maybe the same fallback mechanism in case of `ANY`. Should create a new test, and extract a helper method we can also use in the Integration Request handler?

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
```diff
@@ -1,20 +1,143 @@
...
+            "url": integration_req["uri"],
+            "params": integration_req["query_string_parameters"],
+            "headers": default_apigw_headers,
+        }
+
+        if method not in NO_BODY_METHODS:
+            request_parameters["data"] = integration_req["body"]
+
+        # TODO: configurable timeout (29 by default)
+        # request_parameters["timeout"] = 29
```
cloutierMat, line 74: Question: Do you know if the timeout also affects proxy integration? Is it something we should manage at the Integration handler level?

tests/aws/services/apigateway/test_apigateway_http.py
```diff
@@ -42,23 +43,36 @@ def add_http_integration_transformers(snapshot):
...
+        # TODO: for HTTP integration only: requests (urllib3) automatically adds `Accept-Encoding` when sending the
+        #  request, seems like we cannot remove it
+        "$..headers.accept-encoding",
+        "$..headers.x-amz-apigw-id",
+        # TODO: only missing for HTTP_PROXY
+        "$..headers.x-amzn-remapped-x-amzn-requestid",
+        "$..headers.x-amzn-trace-id",
+        # TODO: only missing for HTTP
+        "$..headers.x-amzn-requestid",
+        "$..headers.x-amzn-trace-id",
```
cloutierMat, line 60: nit: seems like `"$..headers.x-amzn-trace-id"` is missing for both.

tests/unit/services/apigateway/test_handler_integration_request.py
```diff
@@ -220,6 +224,20 @@ def test_request_override(self, integration_request_handler, default_context):
             "multivalue": ["1header", "2header"],
         }
 
+    def test_request_override_casing(self, integration_request_handler, default_context):
```
cloutierMat, line 227: Good! Another strange behavior is mostly supported out of the box! :smile:

cloutierMat (approved):
Fantastic work! So much to unpack in a small pr! :rocket: 

Great addition to the tests and super excited to see the tests starting to go green! and already some improvements on parity! :man_dancing: 

I have left a question about the way we delete the headers, but nothing blocking!

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/http.py
bentsku: I believe this should be the value `timeoutInMillis` from the `Integration`: https://docs.aws.amazon.com/apigateway/latest/api/API_Integration.html

So it should be managed at the integration level, I also had it for `PROXY` I think but removed it when refactoring. I'll update 👍 

Thinking of which, there's also caching which is currently not supported at all, we might want to think of that later for the features not currently implemented!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -68,12 +68,16 @@ def __call__(
         # here we update the `Response`. We basically need to remove all headers and replace them with the mapping, then
         # override them if there are overrides.
         # for the body, it will maybe depend on configuration?
+        for key in list(response.headers.keys(lower=True)):
+            if key not in ("connection", "content-type"):
+                del response.headers[key]
```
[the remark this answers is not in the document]
bentsku: I'm not too sure about the whole "reconstruct" the whole Response object from the integration response handler, it almost defeat the purpose a little bit (we could just create a dict with the integration "response", and then only serialize it in the next "request handler", instead of having the integration response handler recreating the response completely. But still the big philosophical question I don't have a proper answer for 😄 

But yes, sorry, we could basically override the whole `Response.headers` directly. One way would be initialize an empty `Headers` object and add the few values that should be kept, but this is a bit annoying because we need to be mindful of the fact that some headers might be "multi-values" and we need to call individually `getlist` on every key, and then update this with the mapping. 

Something like:
```python
response_headers = Headers()
for header in ("connection", "content-type"):
    if values := response.headers.getlist(header):
        for value in values:
            response_headers.add(header, value)
```
or maybe a simplified version:
```python
response_headers = Headers()
for header in ("connection", "content-type"):
    if values := response.headers.getlist(header):
        response_headers.setlist(header, values)
```

I don't really have a preference, it also look a bit big, and same level of indentation!

tests/aws/services/apigateway/test_apigateway_http.py
bentsku: good catch, thanks!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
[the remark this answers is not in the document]
bentsku: @cloutierMat I think I'll merge the PR now, and we can update this in #11144 if you think any of the one above is better! I think we will need to rebase anyway. Is that ok?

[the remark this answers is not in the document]
cloutierMat: Yes, totally. Sorry for the delay, I don't think it is a deal breaker at all. Let's have that conversation and adjust in a future PR.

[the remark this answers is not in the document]
bentsku: No worries, no delay at all here 😄 Just didn't want to block future merge with conflicts 😅 sorry for that!

PR #11144, opened 2024-07-05
implement response template

cloutierMat:
## TODO

What's left to do:

- [ ] write unit tests


tests/unit/services/apigateway/test_handler_integration_response.py
```diff
@@ -93,3 +105,202 @@ def select_int_response(selection_value: str) -> IntegrationResponse:
...
+        self, default_context, create_default_response, integration_response_handler
+    ):
+        integration_response = IntegrationResponse(
+            statusCode="300",
+            selectionPattern="",
+            responseParameters=None,
+            responseTemplates=None,
+        )
+        default_context.resource_method["methodIntegration"]["integrationResponses"] = {
+            "default": integration_response
```
bentsku, line 210: nit: I got a bit confused by the `default` naming here 😅 is it supposed to be status code of the integration response? not very important as we don't rely on it

bentsku (commented):
Nice work! 🚀 lots of validation went into this, it's quite apparent 😄

I mostly have a comment regarding the code paths in the handler if we have a response template or not. Not quite sure what is the right solution, and I'd like your opinion on it too! 

Thank you so much for adding this, I can't wait to try to run the pipeline with all the open PRs merged!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -62,17 +73,27 @@ def __call__(
...
-        if mapped_headers := response_data_mapping.get("header"):
-            response.headers.update(mapped_headers)
+        if headers := merge_dicts(response_data_mapping["header"], response_override["header"]):
+            response.headers.update(headers)
+
+        # We set the status code from the integration response unless there was a response override
+        response.status_code = response_override["status"] or integration_response["statusCode"]
+
+        # The body will be taken from template mapping.
+        response.data = body
```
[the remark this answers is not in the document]
cloutierMat: That's true that `.data` is doing a lot of extra work. I will see if I can come up with along the lines of you suggestion.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -128,3 +149,59 @@ def select_integration_response(
...
+        if not (response_templates := integration_response["responseTemplates"]):
+            return ""
+
+        # Aws seems to ignore the integration request headers and uses the invocation request header
+        accept = request["headers"].get("accept", APPLICATION_JSON)
+        if template := response_templates.get(accept):
+            return template
+        # TODO aws seemed to favor application/json as default when unmatched regardless of "first"
+        if template := response_templates.get(APPLICATION_JSON):
+            return template
```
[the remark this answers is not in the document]
cloutierMat: Yeah, it is a bit redundant in the case where it isn't set and `application/json` does not exists. It might also be simpler to follow the logic without that redundancy, Good call

tests/unit/services/apigateway/test_handler_integration_response.py
```diff
@@ -93,3 +105,202 @@ def select_int_response(selection_value: str) -> IntegrationResponse:
...
+    Create a context populated with what we would expect to receive from the chain at runtime.
+    We assume that the parser and other handler have successfully populated the context to this point.
+    """
+
+    context = RestApiInvocationContext(
+        Request(
+            # method=HTTPMethod.POST,
+            # headers={"header": ["header1", "header2"]},
+            # path=f"{TEST_API_STAGE}/resource/path",
+            # query_string="qs=qs1&qs=qs2",
```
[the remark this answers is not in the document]
cloutierMat: oups thanks

tests/unit/services/apigateway/test_handler_integration_response.py
cloutierMat: I actually meant to put `$default`, as to test the behaviour different with integration request, which would prioritize it.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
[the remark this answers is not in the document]
cloutierMat: Actually I realised that neither mine or yours  are accurate. There are certain headers that pass through, as you noticed with `("connection", "content-type")`, but some do get remapped when attempting to override though either request template or mapping.

I have updated the code to reflect. It is a little bit more verbose but hopefully is well separated enough to be easy to understand. I am not married to it either and look forward to your feedback!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -61,22 +73,57 @@ def __call__(
...
+        for header in remapped_headers:
+            if value := headers.get(header):
+                headers[f"x-amzn-remapped-{header}"] = value
+                headers.remove(header)
 
-        # there must be some default headers set, snapshot those?
-        if mapped_headers := response_data_mapping.get("header"):
-            response.headers.update(mapped_headers)
+        # Those headers are passed through from the response headers, there might be more?
+        passthrough_headers = ("connection", "content-type", "content-length")
```
bentsku, line 115: maybe `Content-Type` might not come from the response but is set here, depending on the template? We'll need to try with an integration that returns XML or HTML but the response template is JSON or there is no response template 🤔 or maybe you've done it already 😅 

> API Gateway includes a Content-Type header for all integration responses. By default, the content type is application/json.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -132,3 +179,59 @@ def select_integration_response(
...
+        If there are no templates defined, the body will pass through.
+        Apigateway looks at the integration request `Accept` header and defaults to `application/json`.
+        If no template is matched, Apigateway will use the "first" existing template and use it as default.
+        https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html#transforming-request-response-body
+        """
+        if not (response_templates := integration_response["responseTemplates"]):
+            return ""
+
+        # The invocation request header is used to find the right response templated
+        accept = request["headers"].get("accept")
```
bentsku, line 197: I also found this in the documentation: 
> When a request contains multiple media types in its Accept header, API Gateway only honors the first Accept media type.

I guess it depends it's a proper header list, in that case `get` might get the first, or if we need to parse the value and split on `,`. Maybe we could write a quick test for it?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -61,22 +73,57 @@ def __call__(
...
+            "connection",
+            "content-length",
+            "date",
+            "x-amzn-requestid",
+            "content-type",
+        )
+        for header in remapped_headers:
+            if value := headers.get(header):
+                headers[f"x-amzn-remapped-{header}"] = value
+                headers.remove(header)
```
bentsku, line 112: This is a really great find! It's also very similar to what is done in the `HTTP_PROXY` integration.
[The big table in the documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis) shares more insights, and it seems it can be different from each integration... this is painful. 

I think we should tackle this in a follow up, I wonder how we can generalize this. Maybe a util? Because I was thinking of a handler, but in the integration response step, this is very specific, it happens after mapping/overriding but before passing through, unlike the proxy where it just happens on the response directly. This is going to be fun to test 😅 and we also need to do it for requests, it seems a few are dropped, nothing remapped.

tests/unit/services/apigateway/test_handler_integration_response.py
```diff
@@ -93,3 +105,212 @@ def select_int_response(selection_value: str) -> IntegrationResponse:
...
+        integration_response_handler(default_context, response)
+        assert response.data == b"json"
+
+        # Aws favors json when not math
+        response = create_default_response()
+        response.headers["content-type"] = "text/html"
+        integration_response_handler(default_context, response)
+        assert response.data == b"json"
+
+    def test_remapped_headers(
```
bentsku, line 291: nice new test, made it very clear 💯

bentsku (approved):
LGTM! Great job! And great investigation, we're pushing far into parity here 🚀 I believe most of the comments could be addressed in follow-ups. 

The remapping of headers is super interesting! 

I've actually looked the remapped headers now, and they are, for once, documented:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis

I found some gems here, some we hit our heads against:
> The test invocation of a method uses the default content type of application/json and ignores specifications of any other content types.

So yes, we have a bit of work regarding the remapping, but I'm sure we can nicely generalize it and make it usable both in the integration response here and directly in the `_PROXY` integration as well. It might make the integration response handler a bit smaller as well, removing the headers list from it (we might need to declare them somewhere else, TBD 😄) as there are multiple dimensions to it, dropped/remapped/passthrough. 

Great job, thanks for pushing through and finding where we were wrong!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
cloutierMat: There will be a follow up pr with it that will include a fix for the headers

PR #11148, opened 2024-07-05
Apigw ng implement mock integration

cloutierMat:
## Motivation

Implements Mock integration. https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-mock-integration.html

Upon testing with the console, I confirmed that any integer, positive or negative will work. It will then get matched to the response as usual.

I didn't test much of the validated tests yet as #11143 and #11144 both bring fixes required to enable end to end integrations.

## Changes

- Mock integration
- Unit tests


bentsku (approved):
LGTM! This is great! This is going to enable so many AWS-validated tests, thank you so much 💯 clean implementation 👌 
Just a minor question around the `json.loads`, but then it should be good to merge!

PR #11141, opened 2024-07-04
fix opensearch and kinesis test markers

alexrashed:
## Motivation
While working on #11133, I saw that the test module contains a few tests with "needs fixing".
Also, @bentsku made me aware of the fact that the opensearch test module contains a lot of `markers.aws.unkown` which we try to get rid of with lots of PRs in the last few months. Just adding a few cross-links to connect these efforts a bit:
- https://github.com/localstack/localstack/pull/10994
- https://github.com/localstack/localstack/pull/10991
- https://github.com/localstack/localstack/pull/10987
- https://github.com/localstack/localstack/pull/10929
- https://github.com/localstack/localstack/pull/10920
- https://github.com/localstack/localstack/pull/10892
- https://github.com/localstack/localstack/pull/10653
- https://github.com/localstack/localstack/pull/9151
- https://github.com/localstack/localstack/pull/9141

## Changes
- Fixes all but one non-AWS validated kinesis test (the last one fails with an Internal Server Error / 500 by AWS).
- Basically marks most of the opensearch tests as `needs_fixing` (the domain creation fixture is not working against AWS yet), and some as `only_localstack` (there are a lot of tests for the different endpoint strategies).

PR #11154, opened 2024-07-08
fix aws_http_client_factory fixture to use TEST_AWS_REGION_NAME

alexrashed:
## Motivation
In our [latest scheduled run testing multi-account / multi-region capabilities of LocalStack](https://app.circleci.com/pipelines/github/localstack/localstack/26458/workflows/852a2bc9-8c02-497f-a314-9b4c048af228/jobs/225179/parallel-runs/1/steps/1-111), the Kinesis tests adjusted with https://github.com/localstack/localstack/pull/11133 and https://github.com/localstack/localstack/pull/11141 failed:
```
FAILED tests/aws/services/kinesis/test_kinesis.py::TestKinesis::test_subscribe_to_shard_cbor_at_timestamp - assert 200 == 500
 +  where 500 = <Response [500]>.status_code
FAILED tests/aws/services/kinesis/test_kinesis.py::TestKinesis::test_get_records - assert 200 == 400
 +  where 400 = <Response [400]>.status_code
FAILED tests/aws/services/kinesis/test_kinesis.py::TestKinesis::test_get_records_empty_stream - assert 200 == 400
 +  where 400 = <Response [400]>.status_code
```

It turns out that our `aws_http_client_factory` fixture did not yet take the `TEST_AWS_REGION_NAME` config into account.

## Changes
- Adjusts the `aws_http_client_factory` fixture to also set the region from `TEST_AWS_REGION_NAME`.

## Testing
- Set the following env vars when running the failing tests: `TEST_AWS_ACCOUNT_ID=123456789012;TEST_AWS_REGION_NAME=ap-southeast-2;TEST_AWS_ACCESS_KEY_ID=123456789012` 
- [Triggered Multi-Accounts / Multi-Region test run](https://app.circleci.com/pipelines/github/localstack/localstack/26468/workflows/f4663d57-5d04-4e76-9ae7-b4fb7a4012a7)

### DIFF
--- a/localstack-core/localstack/testing/aws/util.py
+++ b/localstack-core/localstack/testing/aws/util.py
@@ -190,6 +190,7 @@ def base_aws_session() -> boto3.Session:
     session = boto3.Session(
         aws_access_key_id=TEST_AWS_ACCESS_KEY_ID,
         aws_secret_access_key=TEST_AWS_SECRET_ACCESS_KEY,
+        region_name=TEST_AWS_REGION_NAME,
     )
     # make sure we consider our custom data paths for legacy specs (like SQS query protocol)
     session._loader.search_paths.append(LOCALSTACK_BUILTIN_DATA_PATH)
--- a/localstack-core/localstack/testing/aws/util.py
+++ b/localstack-core/localstack/testing/aws/util.py
@@ -190,7 +190,6 @@ def base_aws_session() -> boto3.Session:
     session = boto3.Session(
         aws_access_key_id=TEST_AWS_ACCESS_KEY_ID,
         aws_secret_access_key=TEST_AWS_SECRET_ACCESS_KEY,
-        region_name=TEST_AWS_REGION_NAME,
     )
     # make sure we consider our custom data paths for legacy specs (like SQS query protocol)
     session._loader.search_paths.append(LOCALSTACK_BUILTIN_DATA_PATH)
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/testing/aws/util.py
```diff
@@ -190,6 +190,7 @@ def base_aws_session() -> boto3.Session:
     session = boto3.Session(
         aws_access_key_id=TEST_AWS_ACCESS_KEY_ID,
         aws_secret_access_key=TEST_AWS_SECRET_ACCESS_KEY,
+        region_name=TEST_AWS_REGION_NAME,
```
viren-nadkarni, line 193: It is by design that these fixtures only consider the credentials and not the region. One must use `aws_client_factory` fixture to construct a client associated with a region.

alexrashed: Ah, interesting, why is that necessary? Why shouldn't it default to the region we have configured in the config?
The problem is that this fixture is also used by `aws_http_client_factory`, where we then do not take the `TEST_AWS_REGION_NAME` into account otherwise.

viren-nadkarni: It may be a good idea to keep sessions not bound to a region. If the `base_aws_session()` uses `TEST_AWS_REGION_NAME`, one might assume `secondary_aws_session()` uses the `SECONDARY_TEST_AWS_REGION_NAME`.

Consider updating `aws_http_client_factory` to take into account the `TEST_AWS_REGION_NAME` as a fallback, in line with how `base_aws_client_factory` does it

alexrashed: Thanks, I moved setting the region to the `aws_http_client_factory`. Let me know what you think :)

viren-nadkarni (approved):
LGTM :+1:

PR #11160, opened 2024-07-09
Feature/eventbridge v2 use arn util for arns

maxhoheiser:
## Motivation
Using the arn util to define arns via and access them via a property of the dataclasses defined for each top-level entity of events (bus, rule, archive, replay) is preferable to having the logic of creating arns located in the model files of event bridge.

### DIFF
--- a/localstack-core/localstack/services/events/models.py
+++ b/localstack-core/localstack/services/events/models.py
@@ -38,7 +38,12 @@ from localstack.services.stores import (
     CrossRegionAttribute,
     LocalAttribute,
 )
-from localstack.utils.aws.arns import events_archive_arn, events_replay_arn, events_rule_arn
+from localstack.utils.aws.arns import (
+    event_bus_arn,
+    events_archive_arn,
+    events_replay_arn,
+    events_rule_arn,
+)
 from localstack.utils.tagging import TaggingService
 
 TargetDict = dict[TargetId, Target]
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/events/models.py
```diff
@@ -38,7 +38,12 @@
     CrossRegionAttribute,
     LocalAttribute,
 )
-from localstack.utils.aws.arns import events_archive_arn, events_replay_arn
+from localstack.utils.aws.arns import (
+    event_bus_arn,
```
joe4dev, line 42: nit: intentionally `event_` vs `events_` prefix

localstack-core/localstack/services/events/models.py
```diff
@@ -136,6 +136,10 @@ def __post_init__(self):
         if self.state is None:
             self.state = RuleState.ENABLED
 
+    @property
+    def arn(self) -> Arn:
```
joe4dev, line 140: nice that we don't need to save the redundant field 👍

joe4dev (approved):
neat refactoring 👍

localstack-core/localstack/services/events/models.py
maxhoheiser: Yes, internally everything is called event_ .... but externally anything related to event bridge uses the shorthand events_ ... here particular only related to defining the respective ARNs for event_bridge

PR #11161, opened 2024-07-09
Add utility to prepend a string with a slash

viren-nadkarni:
## Changes

This PR adds a small utility to prepend `/` to a given string.

This is being moved from localstack-ext MWAA provider.


alexrashed (approved):
LGTM, but I am not really sure if it's necessary to add this to this repo when it's only used in a downstream dependency. We could also just introduce a utility module in the downstream dependency itself for as long as it's not being used here... I'll leave that up to you though, no strong opinion.

PR #11163, opened 2024-07-09
Feature/eventbridge v2 custom json encoder exporter

maxhoheiser:
## Motivation
Using a custom util function to dump json to strings utilizing the custom json encoder that handles exporting of time stamp to the correct string format removes the potential pitfall of forgetting to include the custom json class when using `json.dumps`.

### DIFF
--- a/localstack-core/localstack/services/events/utils.py
+++ b/localstack-core/localstack/services/events/utils.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 from datetime import datetime, timezone
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from localstack.aws.api.events import (
     ArchiveName,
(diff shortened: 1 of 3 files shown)


joe4dev (approved):

localstack-core/localstack/services/events/utils.py
```diff
@@ -38,6 +38,10 @@ def default(self, obj):
         return super().default(obj)
 
 
+def to_json(obj: Any, separators: tuple[str] = None) -> str:
+    return json.dumps(obj, separators, cls=EventJSONEncoder)
```
joe4dev, line 42: Separators are a positional argument: https://docs.python.org/3/library/json.html#json.JSONEncoder

This fix should resolve the failing EventBridge v2 provider tests

joe4dev: You can check out this sample from Marco in SFN: `localstack.services.stepfunctions.asl.utils.encoding.to_json_str`

maxhoheiser: Done thanks for the hint :)

PR #11167, opened 2024-07-09
APIGW NG add AWS_PROXY integration

bentsku:
## Motivation
This PR implements the `AWS_PROXY` API Gateway integration. 
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html

## Changes
- populate the `identity` field of the context with the default values for a method with no authorizer
- port a few utils needed for the lambda integration
- implement the `AWS_PROXY` integration, adding the utils to create the input payload and parsing the response from lambda and passing it down the chain

## Testing
All the tests related to `AWS_PROXY` in `tests.aws.services.apigateway.test_apigateway_lambda` are passing when forcing the provider, with way less snapshot ignore than before 🥳 
(With `test_lambda_aws_proxy_integration_non_post_method` depending #11183 to be merged due to the gateway response default value issue).

## TODO

What's left to do:

- [ ] write maybe a bit of unit test
- [ ] manage all exceptions, do validation in console too
- [ ] ?

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
@@ -1,11 +1,13 @@
 import copy
 import logging
 import re
+from typing import Type, TypedDict
 
 from moto.apigateway.models import RestAPI as MotoRestAPI
 
 from localstack.services.apigateway.models import MergedRestApi, RestApiContainer, RestApiDeployment
 
+from .context import RestApiInvocationContext
 from .moto_helpers import get_resources_from_moto_rest_api
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
```diff
@@ -53,3 +55,36 @@ def render_uri_with_path_parameters(uri: str, path_parameters: dict[str, str]) -
...
+    )
+
+
+def get_lambda_function_arn_from_invocation_uri(uri: str) -> str:
+    """
+    "arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:123456789012:function:SimpleLambda4ProxyResource/invocations",
+    :param uri: the integration URI value for a lambda function
+    :return: the lambda function ARN
+    """
+    return uri.split(":lambda:path")[1].split("functions/")[1].removesuffix("/invocations")
```
cloutierMat, line 78: It seems like you are doing an extra split? Wouldn't splitting once at `"functions/"` be enough?

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
```diff
@@ -37,3 +87,173 @@ class RestApiAwsProxyIntegration(RestApiIntegration):
...
+            Payload=event,
+            InvocationType="Event" if asynchronous else "RequestResponse",
+        )
+        if payload := inv_result.get("Payload"):
+            return payload.read()
+        return b""
+
+    def parse_lambda_response(self, payload: bytes) -> LambdaProxyResponse:
+        try:
+            lambda_response = json.loads(payload)
```
cloutierMat, line 175: Is there still a weird edge case where a lambda returning `"[]"` would lead to a `TypeError` down the line?

tests/aws/services/apigateway/test_apigateway_lambda.py
```diff
@@ -109,7 +89,15 @@ def test_lambda_aws_proxy_integration(
...
-    snapshot.add_transformer(snapshot.transform.key_value("deploymentId"))
+    # TODO: update global transformers, but we will need to regenerate all snapshots at once
+    snapshot.add_transformers_list(
+        [
+            snapshot.transform.key_value("deploymentId"),
+            snapshot.transform.jsonpath("$..headers.Host", value_replacement="host"),
+            snapshot.transform.jsonpath("$..multiValueHeaders.Host[0]", value_replacement="host"),
+        ],
+        priority=-1,
+    )
```
cloutierMat, line 100: That already cleaned it up a lot!
:rocket:

tests/aws/services/apigateway/test_apigateway_lambda.py
```diff
@@ -300,8 +288,15 @@ def test_lambda_aws_proxy_integration_non_post_method(
...
     )
-    aws_client.apigateway.create_deployment(restApiId=api_id, stageName=stage_name)
 
+    # TODO: comment
+    aws_client.apigateway.put_gateway_response(
+        restApiId=api_id,
+        responseType="INTEGRATION_FAILURE",
+        responseParameters={},
+    )
+
```
cloutierMat, line 298: That seems like a left over from our session? Do you want to test some gateway responses here?

cloutierMat (commented):
:rocket:  Super clean work and a nice set of utils we should be able to reuse!

Appart from a couple little nitpick there is the weird issue where the response could be an array, which would decode to json but fail down the line. It would probably be good to understand AWS behaviour here. I would assume It should raise the `InternalServerError` but it is hard to predict sometimes :rofl:

cloutierMat (approved):
Wow, it actually behave as expected when passing a list? :rofl: 
Great work :rocket:

PR #11168, opened 2024-07-09
fix caching in tests-pro-integration after source move

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/10800 the main module in localstack/localstack has been moved from the root of the repository to the `localstack-core` folder.
Unfortunately, this silently killed the caching in the workflow which runs the Community tests against the LocalStack Pro code (since the `.filesystem` folder now is in `localstack/localstack-core/.filesystem`).
This PR addresses this issue which should hopefully lead to an increase in the pipeline stability.

## Changes
- Adjusts the path for the var lib cache in `tests-pro-integration.yml`

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -249,7 +249,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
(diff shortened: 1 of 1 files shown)


dfangl (approved):
nice catch

PR #11170, opened 2024-07-10
Clarify docker client volume API

dfangl:
## Motivation
Our current usage of bind mounts in docker utilizes the docker `volumes` API. However, this API has been superceded by the `mounts` API.
There was also one, so far unsuccessful, try to use `mounts` for everything in LocalStack, with #9887.

There are some drawbacks when using the mounts API for bind mounts, for example the fact that the host path is not automatically created.
However, it would also achieve many advantages, like being able to mount subpaths of named volumes into containers.

To prepare for the future use of mounts, for now in addition to volumes, this PR renames all current occurrences of `mounts` or `mount_volumes` to `volumes`, to make the differenciation clearer.
We can then start to implement the mounts API in the docker client, and step by step migrate usages of `volumes` to `mounts` - where possible, as well as enable new usecases, like using a named volume as mount point for `/var/lib/localstack`, or adding `tmpfs` file systems to compute containers.

## Changes
* No userfacing changes
* Rename all occurrences of `mounts` or `mount_volumes` when using the docker client to `volumes`

### DIFF
--- a/localstack-core/localstack/utils/container_utils/container_client.py
+++ b/localstack-core/localstack/utils/container_utils/container_client.py
@@ -499,7 +499,7 @@ class DockerRunFlags:
     env_vars: Optional[Dict[str, str]]
     extra_hosts: Optional[Dict[str, str]]
     labels: Optional[Dict[str, str]]
-    mounts: Optional[List[SimpleVolumeBind]]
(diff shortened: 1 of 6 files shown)


alexrashed (approved):
LGTM! Nice prep work for upcoming changes! 🚀

PR #11171, opened 2024-07-10
fix coverage drops by skipping report step for `acceptance-only-run`

alexrashed:
## Motivation
Since #11049 by default we only run the acceptance tests for commits on `master`.
While this PR skips pushing test results to our internal test analytics, it still pushes the coverage data, [which leads to the coverage metrics jumping between 47% (acceptance tests only) and 84% (full test suite)](https://coveralls.io/github/localstack/localstack).
image

## Changes
- Avoids the report step completely for the `acceptance-only-run` workflow in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -643,9 +643,7 @@ jobs:
       - run:
           name: Upload test metrics and implemented coverage data to tinybird
           command: |
-            if [[ $ONLY_ACCEPTANCE_TESTS == "true" ]]; then
-              echo "Skipping parity reporting to tinybird because only acceptance tests ran"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -877,11 +867,6 @@ workflows:
           resource_class: medium
           requires:
             - docker-build-amd64
-      - report:
-          requires:
-            - acceptance-tests-amd64
-            - acceptance-tests-arm64
-            - unit-tests
```
alexrashed, line 884: FYI: This is the important part here, we completely remove the report step from the workflow definition of the `acceptance-only-run` workflow.

dfangl (approved):

PR #11173, opened 2024-07-10
update coverage config, remove obsolete dev run configurator

alexrashed:
## Motivation
This PR serves two purposes. It (a) adjusts the coverage config a bit such that the package path and the source config are properly separated, and (b) removes an obsolete coverage-related dev container config which is not used anymore.

## Changes
- Removes the `CoverageRunScriptConfigurator` since the `run-coverage.py` is not used anymore.
- Slightly changes the `pyproject.toml` to properly separate between the configuration of the sources and the package lookup config.

## TODO
- [ ] Wait for the removal of the usage of `run-coverage.py` in our downstream dependency.


silv-io (approved):
LGTM!

PR #11176, opened 2024-07-10
Use `StrEnums` instead of `str` for generated constant classes in the ASF APIs

dfangl:
## Motivation
Currently, the scaffolding currently translates the `StringShape`s from the botocore specs to enum string classes, for example:

```python
class Architecture(str):
    x86_64 = "x86_64"
    arm64 = "arm64"
```

The issue here, from a type checking perspective, is that the values, for example Architecture.x86_64, are not of type Architecture, they are simple strings:

```python
from localstack.aws.api.lambda_ import Architecture

print(isinstance(Architecture.x86_64, Architecture))
print(isinstance(Architecture.x86_64, str))
```
for example, returns:

```
False
True
```

This is not really an issue for production - they are strings, and the exact type does not really matter after it is serialized anyway. However, when developing, we are hit with fun IDE warnings and mypy also reports it as an error:

```python
def test(param: Architecture):
    pass


test(Architecture.arm64)
```

Mypy error:

```bash
mypy test.py
test.py:14: error: Argument 1 to "test" has incompatible type "str"; expected "Architecture"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

My proposed solution would be to switch those types to StrEnum, a new type of enum introduced with python3.11 (https://docs.python.org/3/library/enum.html#enum.StrEnum). It is only applicable for code not used by the CLI, but our API scaffolds should not be imported from the CLI.

A StrEnum is both an instance of str and enum, and more importantly, the members are also a type of the respective class.
Also, the dunder __str__ and __repr__ methods are those of string, therefore it should be interchangable (we will see how the testing of this PR works out).

## Changes
* Use `StrEnum` instead of `str` for generated API `StringShape`s, which will lead to correct types for the enum members.

### DIFF
--- a/localstack-core/localstack/services/s3/utils.py
+++ b/localstack-core/localstack/services/s3/utils.py
@@ -5,7 +5,8 @@ import hashlib
 import logging
 import re
 import zlib
-from typing import IO, Any, Dict, Literal, NamedTuple, Optional, Protocol, Tuple, Type, Union
+from enum import StrEnum
+from typing import IO, Any, Dict, Literal, NamedTuple, Optional, Protocol, Tuple, Union
 from urllib import parse as urlparser
 
 import xmltodict
@@ -516,8 +517,8 @@ def uses_host_addressing(headers: Dict[str, str]) -> str | None:
         return bucket_name
 
 
-def get_class_attrs_from_spec_class(spec_class: Type[str]) -> set[str]:
-    return {getattr(spec_class, attr) for attr in vars(spec_class) if not attr.startswith("__")}
+def get_class_attrs_from_spec_class(spec_class: StrEnum) -> set[str]:
+    return {str(spec) for spec in spec_class}
 
 
 def get_system_metadata_from_request(request: dict) -> Metadata:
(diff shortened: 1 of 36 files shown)


localstack-core/localstack/services/s3/utils.py
```diff
@@ -516,8 +517,8 @@ def uses_host_addressing(headers: Dict[str, str]) -> str | None:
         return bucket_name
 
 
-def get_class_attrs_from_spec_class(spec_class: Type[str]) -> set[str]:
-    return {getattr(spec_class, attr) for attr in vars(spec_class) if not attr.startswith("__")}
+def get_class_attrs_from_spec_class(spec_class: type[StrEnum]) -> set[str]:
+    return {str(spec) for spec in spec_class}
```
dfangl, line 521: This is the only actual code change necessary, which iterates over the classes, now an enum.

alexrashed (approved):
Nice change and really well explained! The stronger the typing the better! 💯 
The changes are looking good, they update the scaffold, update all generated APIs, as well as a single issue with the usage of the generated code. 🏅

PR #11177, opened 2024-07-10
add ability for shutdown hooks to call internal APIs

thrau:
## Motivation

@HarshCasper noted that this sample didn't work with the new runtime https://github.com/localstack-samples/localstack-init-hooks/tree/main/example-docker-compose

This is because we were running shutdown hooks after the runtime server was shut down. I changed the runtime to correct the order to what we were doing previously.

I also noticed that we're returning 503 errors when calling `/_localstack` APIs when shutting down, which I find stifles the use of shutdown hooks. I added an exception for those internal APIs.

Moreover I added a bootstrap test to validate the shutdown hook behaviors

## Changes

* Gateway is now shut down _after_ shutdown hooks are run
* `/_localstack` endpoints can now be called while localstack is shutting down
* added a bootstrap test for testing shutdown hooks

### DIFF
--- a/tests/bootstrap/test_init.py
+++ b/tests/bootstrap/test_init.py
@@ -1,0 +1,109 @@
+import json
+
+import pytest
+import requests
+
+from localstack.config import in_docker
+from localstack.testing.pytest.container import ContainerFactory
+from localstack.utils.bootstrap import ContainerConfigurators
+from localstack.utils.container_utils.container_client import VolumeBind
+
+pytestmarks = pytest.mark.skipif(
+    condition=in_docker(), reason="cannot run bootstrap tests in docker"
+)
+
+
+class TestInitHooks:
+    def test_shutdown_hooks(
+        self,
+        container_factory: ContainerFactory,
+        stream_container_logs,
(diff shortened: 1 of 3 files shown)


tests/bootstrap/test_init.py
```diff
@@ -0,0 +1,105 @@
...
+                ContainerConfigurators.volume(
+                    VolumeBind(str(shutdown_hooks), "/etc/localstack/init/shutdown.d")
+                ),
+            ]
+        )
+        running_container = container.start(attach=False)
+        stream_container_logs(container)
+        wait_for_localstack_ready(running_container)
+
+        # check the ports listening on 0.0.0.0
```
dfangl, line 53: Is this a copy-paste leftover?

dfangl (approved):
Looks good! So this effectively now restores the behavior previously done by the `ON_AFTER_SERVICE_SHUTDOWN_HANDLERS`, right?

thrau:
> So this effectively now restores the behavior previously done by the ON_AFTER_SERVICE_SHUTDOWN_HANDLERS, right?

correct!

PR #11178, opened 2024-07-10
Add generic botocore-model-based conversion util for CFn resource providers

dominikschubert:
## Motivation

Recently we had a report for another casing mismatch between the definition of the CloudFormation resource properties (as defined in the corresponding resource schema) and the downstream service API. This is quite tedious to hunt down on a case-by-case basis, so I thought I'd introduce a util that takes care of this recursively for the whole request kwargs.

The reported issue specifically affected `AWS::Events::Rule` and  the downstream `Events.PutTargets` API call when using the `AwsVpcConfiguration` key. On the API side this is actually `awsvpcConfiguration` which lead to an exception being raised on the `client.put_targets` call.

## Changes

- Resource providers now have another util at their disposal. `convert_request_kwargs` can be used to recursively fix issues mapping from the resource properties to the botocore client call.
  - Casing issues are fixed
  - Types are converted automatically between int/str/boolean
  - Unknown keys are dropped
- `AWS::Events::Rule` resources can now successfully be deployed with ecs parameters that include the `AwsVpcConfiguration` property.

## Discussion

This approach can be extended in various ways like:
- ~Drop elements if the input shape doesn't expect them~ :heavy_check_mark: 
- ~Convert passed types automatically~ :heavy_check_mark: 
- Instrument the boto client factory that we pass into the resource provider to do this automatically via boto hooks

### DIFF
--- a/localstack-core/localstack/services/cloudformation/provider_utils.py
+++ b/localstack-core/localstack/services/cloudformation/provider_utils.py
@@ -13,6 +13,8 @@ from copy import deepcopy
 from pathlib import Path
 from typing import Callable, List, Optional
 
+from botocore.model import Shape, StructureShape
+
 
 def generate_default_name(stack_name: str, logical_resource_id: str):
     random_id_part = str(uuid.uuid4())[0:8]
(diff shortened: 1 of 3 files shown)


pinzon (approved):
Great!!

PR #11181, opened 2024-07-11
use new runtime procedure for starting localstack in host mode

thrau:
## Motivation

With #10942 we wanted to deprecate `localstack start --host`, which is why I didn't update it. However, @alexrashed argued (correctly) that people are still using it and we should keep backwards compatibility.

This PR does a minimal change to simply make `localstack start --host` basically equivalent to ``python -m localstack.runtime.main``.

## Changes

* `localstack start --host` now uses the new runtime to start localstack

## Testing

* check out the PR, activate venv
* run `python -m localstack.cli.main start --host`, observe it still works

### DIFF
--- a/localstack-core/localstack/utils/bootstrap.py
+++ b/localstack-core/localstack/utils/bootstrap.py
@@ -360,9 +360,9 @@ def should_eager_load_api(api: str) -> bool:
 
 
 def start_infra_locally():
-    from localstack.services import infra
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Awesome, after this PR has been merged we can basically remove all the old runtime code! Kudos! 🥳 🚀

PR #11182, opened 2024-07-11
Refactor/eventbridge v2 clean tests archive and replay

maxhoheiser:
## Changes
Move functionality to create custom event bus and get event bus arn for default or custom event bus into fixture.

### DIFF
--- a/tests/aws/services/events/conftest.py
+++ b/tests/aws/services/events/conftest.py
@@ -67,6 +67,21 @@ def events_create_event_bus(aws_client, region_name, account_id):
 
 
 @pytest.fixture
+def events_create_default_or_custom_event_bus(events_create_event_bus, region_name, account_id):
+    def _create_default_or_custom_event_bus(event_bus_type: str = "default"):
+        if event_bus_type == "default":
+            event_bus_name = "default"
+            event_bus_arn = f"arn:aws:events:{region_name}:{account_id}:event-bus/default"
+        if event_bus_type == "custom":
+            event_bus_name = f"test-bus-{short_uid()}"
+            response = events_create_event_bus(Name=event_bus_name)
+            event_bus_arn = response["EventBusArn"]
+        return event_bus_name, event_bus_arn
+
+    return _create_default_or_custom_event_bus
+
+
+@pytest.fixture
 def create_role_event_bus_source_to_bus_target(create_iam_role_with_policy):
     def _create_role_event_bus_to_bus():
         assume_role_policy_document_bus_source_to_bus_target = {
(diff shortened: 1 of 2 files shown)


tests/aws/services/events/test_archive_and_replay.py
```diff
@@ -24,25 +24,17 @@ class TestArchive:
...
-            event_bus_name = f"test-bus-{short_uid()}"
-            response = events_create_event_bus(Name=event_bus_name)
-            event_source_arn = response["EventBusArn"]
+        event_bus_name, event_bus_arn = events_create_default_or_custom_event_bus(event_bus_type)
 
         archive_name = f"test-archive-{short_uid()}"
         response_create_archive = events_create_archive(
             ArchiveName=archive_name,
-            EventSourceArn=event_source_arn,  # ARN of the source event bus
+            EventSourceArn=event_bus_arn,  # ARN of the source event bus
```
joe4dev, line 37: Helpful clarification comment 👍

tests/aws/services/events/conftest.py
```diff
@@ -66,6 +66,21 @@ def _create_event_bus(**kwargs):
             LOG.warning(f"Failed to delete event bus {event_bus_name}: {e}")
 
 
+@pytest.fixture
+def events_create_default_or_custom_event_bus(events_create_event_bus, region_name, account_id):
+    def _create_default_or_custom_event_bus(event_bus_type: str = "default"):
+        if event_bus_type == "default":
+            event_bus_name = "default"
+            event_bus_arn = f"arn:aws:events:{region_name}:{account_id}:event-bus/default"
+        if event_bus_type == "custom":
```
joe4dev, line 75: elif

tests/aws/services/events/conftest.py
```diff
@@ -66,6 +66,21 @@ def _create_event_bus(**kwargs):
...
+            event_bus_name = f"test-bus-{short_uid()}"
+            response = events_create_event_bus(Name=event_bus_name)
+            event_bus_arn = response["EventBusArn"]
+        return event_bus_name, event_bus_arn
```
joe4dev, line 79: What happens in the else case?

It's probably better to fail fast here and raise for `else` rather than ignore silently

joe4dev: The IDE also warns about this:
Screenshot 2024-07-12 at 14 20 46

joe4dev (approved):
LGTM after fixing the missing else case handling 👍

tests/aws/services/events/conftest.py
maxhoheiser: per default, it will be the "default" one since this is the default value set if no value is provided for event_bus_type, this should also avoid that no value is set, so  the IDE warning can be disregarded

tests/aws/services/events/conftest.py
maxhoheiser: I used else since any bus type that is not default should create a custom event bus

tests/aws/services/events/conftest.py
joe4dev: 👍  `else` fixes the warning and using `cutsom` (typo) ;)

PR #11183, opened 2024-07-11
APIGW NG implement AWS bug for INTEGRATION_FAILURE GatewayResponse

bentsku:
## Motivation
While implementing the `AWS_PROXY` integration with #11167, I've stumbled onto a weird issue: AWS would raise an `INTEGRATION_FAILURE` error, which the default status code should be 504 (Gateway Timeout, that's already suspicious), but instead returned `500`. 

After a lot of investigation with @cloutierMat, we came to the conclusion that the CRUD later (`PutGatewayResponse`, `GetGatewayResponse` etc) is not in line with the actual behavior, and that the actual default status code is `500`, but the CRUD calls still returns 504. 

This PR implements the logic to enable the difference between the CRUD layer and the actual invocation layer by attaching a default status code to the exception, while still using the previous default for other exceptions, allowing us to override the behavior. 

This behavior is validated by `tests.aws.services.apigateway.test_apigateway_lambda.test_lambda_aws_proxy_integration_non_post_method`. 

## Changes
- use the Python exception status code if provided
- add a fix in the provider to deal with the inconsistency between what the AWS CRUD operations returns and what AWS actually does under the hood

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway_response.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway_response.py
@@ -8,6 +8,30 @@ from localstack.aws.api.apigateway import (
 )
 
 
+class GatewayResponseCode(StatusCode, Enum):
+    REQUEST_TOO_LARGE = "413"
+    RESOURCE_NOT_FOUND = "404"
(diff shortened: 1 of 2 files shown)


cloutierMat (approved):
:rocket: 
I am glad we had a couple brains that day to figure this one out! :rofl:

Issue #11180, opened 2024-07-11
bug: Different behaviour from AWS for getObjectTagging on a delete marker


bjpirt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If you add tags to an object in a versioned bucket and then delete it (so it creates a delete marker). When you get the object tagging for the key it still returns an empty list of tags

### Expected Behavior

Delete markers in AWS can't hold any data, including tagging. When you call `get-object-tagging` on a delete marker in AWS you get back an error message:

```An error occurred (MethodNotAllowed) when calling the GetObjectTagging operation: The specified method is not allowed against this resource.```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    echo "test data" > test-file.txt
    awslocal s3 mb s3://test-bucket
    awslocal s3api put-bucket-versioning --bucket test-bucket --versioning-configuration Status=Enabled
    awslocal s3 cp test-file.txt s3://test-bucket/test-file.txt
    awslocal s3api put-object-tagging --bucket test-bucket --key test-file.txt --tagging '{"TagSet": [{ "Key": "my-tag", "Value": "my-value" }]}'
    awslocal s3api get-object-tagging --bucket test-bucket --key test-file.txt
    ## lists the tags

    awslocal s3 rm s3://test-bucket/test-file.txt
    awslocal s3api list-object-versions --bucket test-bucket
    # Shows the delete marker

    awslocal s3api get-object-tagging --bucket test-bucket --key test-file.txt
    ## returns an empty list of tags

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

PR #11185, opened 2024-07-11
fix ObjectTagging with DeleteMarker

bentsku:
## Motivation
As reported in #11180, we've had the wrong behavior for DeleteMarker with `ObjectTagging` operations. 

However, my previous assumptions were wrong due to a mistake in a test, which lead to a design decision in the S3 models (the `S3Bucket` `get_object` method signature). 

We will need to update it to support properly this behavior, however this might break persistence as the previous loaded objects might have the previous signature, and fails when used in the provider. We will need to wait for the next major release to remove the previous parameter `raise_for_delete_marker` and add a new one permitting use to separate behavior between calls from `GetObject` vs `GetObjectTagging` (one raise `NoSuchKey` and the other `MethodNotAllowed`). 

## Changes
- add a hacky way to support the behavior reported, and add a TODO for the future to fix the behavior without that hack

_fixes #11180_


cloutierMat (approved):
Good thinking for the store!

Nice set of extra test to validate the behaviour!

PR #11189, opened 2024-07-12
Refactor/eventbridge v2 clean tests event pattern

maxhoheiser:
## Changes
- move tests for events patterns to `test_events_patterns` to align naming of test files
- separate pattern tests into classes - testing `test_pattner` API and testing rule with pattern
- move tests for pattern in rule from `test_events` to `tests_events_patterns` since the pattern not the rule API is tested


joe4dev (changes requested):
Please separate refactoring renames from edits (e.g., test validation) to follow our "Git(Hub) Best Practices" (on internal Notion) – see "… scope PRs to preserve the git file history".

Sidenote:
`test_event_patterns.py` was intentional focusing on the "event_patterns" being a shared functionality used beyond the scope of EventBridge (Pipes, Lambda ESM, potentially SNS). I can see that using the `test_events_*` for consistency seems ok, but would lead to the weird naming `test_events_event_pattern` (which I wanted to avoid). I think just `test_event_patterns` would also be fine for the test name because `test_events_event_patterns` should be only one API test while `test_event_patterns` focuses on the actual functionality of event pattern matching (i.e., event ruler)

joe4dev (approved):
Grouping all event pattern tests makes sense 👍 

The git practices are still not ideal (git mv renames should ideally be separated from any edits such as introducing classes or reordering; partial moves should delete the source within the same PR), but let's move on given the reorderings and edits mix up everything at the end anyway 🏃

PR #11190, opened 2024-07-12
Make localstack an implicit namespace package

dominikschubert:
## Motivation

To allow us to introduce namespace packages downstream like `localstack.pro.core` or `localstack.pro.aws`, we need to first make `localstack` itself a namespace package.

## Changes

- With the removal of `localstack/__init__.py`, `localstack` will become an [implicit namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages)

## TODO

What's left to do:

- [x] Check if anything breaks => caching the venv was breaking the pro integration tests

### DIFF
--- a/localstack-core/localstack/dns/server.py
+++ b/localstack-core/localstack/dns/server.py
@@ -12,12 +12,8 @@ from pathlib import Path
 from socket import AddressFamily
 from typing import Iterable, Literal, Tuple
 
-import dns.flags
-import dns.message
-import dns.query
 import psutil
 from cachetools import TTLCache, cached
-from dns.exception import Timeout
 from dnslib import (
     AAAA,
     CNAME,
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/dns/server.py
alexrashed, line 1: nit: The changes in this file might not even be necessary, at least on first sight they don't seem to have an impact?

alexrashed (approved):
Awesome! Kudos to you and to @thrau for the great preparation leading to this super-small change which at the same time will have quite an impact on how we structure and build LocalStack in the future! 💯🦸🤩

localstack-core/localstack/dns/server.py
dominikschubert: Unfortunately ruff started complaining about the ordering here both in CLI and locally, so :man_shrugging:

PR #11194, opened 2024-07-12
Fix deletion of lambda function versions

dfangl:
## Motivation
We are currently experiencing a bug when deleting a Lambda function version without deleting the entire function, due to a misuse of the `qualified_arn` property.

We had not test deleting only the function version before, which is why I added one as well.

## Changes
* Lambda function versions can now be deleted without an error

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -1314,7 +1314,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
             # delete a version of the function
             version = function.versions.pop(qualifier, None)
             if version:
-                self.lambda_service.stop_version(version.qualified_arn())
+                self.lambda_service.stop_version(version.id.qualified_arn())
                 destroy_code_if_not_used(code=version.config.code, function=function)
         else:
             # delete the whole function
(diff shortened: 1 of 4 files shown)


joe4dev (approved):
Thanks for fixing this @dfangl 🩹 

That's the one that caused `TypeError: 'str' object is not callable` (one of the most common exceptions 😬 )

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1314,7 +1314,7 @@ def delete_function(
             # delete a version of the function
             version = function.versions.pop(qualifier, None)
             if version:
-                self.lambda_service.stop_version(version.qualified_arn())
+                self.lambda_service.stop_version(version.id.qualified_arn())
```
dominikschubert, line 1317: maybe I'm missing something but the version is defined as
```
@dataclasses.dataclass(frozen=True)
class FunctionVersion:
    id: VersionIdentifier
    config: VersionFunctionConfiguration

    @property
    def qualified_arn(self) -> str:
        return self.id.qualified_arn()
```

so why would this change the behavior :thinking:

dominikschubert: ah ok, just saw Joel's comment, makes sense. Kinda ignored the additional `()` there!

PR #11195, opened 2024-07-12
Refactor/eventbridge_v2: Rename test event pattern file

maxhoheiser:
## Motivation
test_event_patterns.py was intentional focusing on the "event_patterns" being a shared functionality used beyond the scope of EventBridge (Pipes, Lambda ESM, potentially SNS). Using the naming test_events_* for consistency here, although fully following it would lead to the weird naming test_events_event_pattern . 
This test file should contain all the tests related to event pattern, this means testing the pattern matching engine on its own, but also testing the API `test_event_pattern` as well as rules with event pattern.


joe4dev (approved):
LGTM after converting the validations to a rename as well (doesn't git detect that automatically? maybe checkout the hints such as `git config --global log.follow true` in our Git best practices guide).

Screenshot 2024-07-12 at 17 48 27

PR #11196, opened 2024-07-12
APIGW NG implement AWS integration

bentsku:
## Motivation
This PR implements the last missing integration: `AWS`. This integration allows API Gateway to interact with other AWS services, by constructing mostly manually a request following the service specs (APIGW will take of the signing and endpoint mostly, and in some case it helps directing the request to the right operation).

There are 2 kind of AWS integrations: `path` and `action`. The documentation is very sparse on those. 
Path allows you to override the path where the request is sent, this can be used for services like S3 for example. 
`action` allows you to target an AWS service action directly, and you might or might not need to give very specific data like the `X-Amz-Target` header to your request. 

Before, we would specify a sub type of integration for each service, but after looking deeply into it with @cloutierMat, we realized it is only one integration in the end, with maybe a few differences between services but that are generalizable between services.
If things get out of control with specific behavior for the `action` type, we could register a "before send" and "after response" callback to fix the request/response before sending it for specific services. It's worth to say that there is literally zero documentation about the `action` type and which/how services are supported. 

Which means we're now support way more services out of the box when using `path` URI! 🚀 

By running all the AWS integrations test (which now all pass ✅), I've spotted a few bugs that have been fixed in the PR as well. 

As a side note, this PR put into light the fact that AWS still supports the Query Protocol for the SSM service, and that API Gateway still sends request with that protocol by default to SSM... even though the botocore specs have always been the JSON specs. Fun 😬 \cc @alexrashed you might like that one... 

## Changes
- implement a generalized way to support `AWS` integration
  - support the `path` type which is very simple
  - for the `action` type, it is a bit more complicated and will still need a bit of effort to support new services, since it seems they fall into different categories. 
- validate existing tests with AWS integrations (DynamoDB, SQS, SFn, Kinesis, EventBridge, Lambda, SSM and S3) 
  - moved the SFn test into its own file
- change how we mapped default values for `Content-Type` and `Accept` for integration request (not passthrough but provided by AWS)
- implement rendering of URI, still need to implement more tests in the future
- skip `needs_fixing` tests for the new provider, as they often don't setup APIGW correctly
- last fix (last in list): fix the access to the REST API to be case insensitive when fetching the stage variables

## Testing
Now with the new provider enabled, all APIGW tests only are passing locally! We will soon enable a new pipeline for the APIGW provider only. 

Still 8 failing tests in the pipeline, fixed with #11198:
```
FAILED tests/aws/scenario/note_taking/test_note_taking.py::TestNoteTakingScenario::test_notes_rest_api
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_base
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_with_body_post[None]
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_with_body_post[]
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_with_body_post[HelloWorld]
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_with_body_post[request_body3]
FAILED tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py::TestTaskApiGateway::test_invoke_with_query_parameters
FAILED tests/aws/test_multiregion.py::TestMultiRegion::test_multi_region_api_gateway
```

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
@@ -36,13 +36,15 @@ def freeze_rest_api(
     )
 
 
-def render_uri_with_stage_variables(uri: str, stage_variables: dict[str, str]):
+def render_uri_with_stage_variables(uri: str | None, stage_variables: dict[str, str]) -> str | None:
     """
     https://docs.aws.amazon.com/apigateway/latest/developerguide/aws-api-gateway-stage-variables-reference.html#stage-variables-in-integration-HTTP-uris
     URI=https://${stageVariables.<variable_name>}
     This format is the same as VTL, but we're using a simplified version to only replace `${stageVariables.<param>}`
     values, as AWS will ignore `${path}` for example
     """
+    if not uri:
+        return uri
 
     def replace_match(match_obj: re.Match) -> str:
         return stage_variables.get(match_obj.group("varName"), "")
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
@@ -59,6 +59,28 @@ def render_uri_with_path_parameters(uri: str, path_parameters: dict[str, str]) -
     return uri
 
 
+def render_integration_uri(
+    uri: str, path_parameters: dict[str, str], stage_variables: dict[str, str]
+) -> str:
+    """
+    A URI can contain different value to interpolate / render
+    It will have path parameters substitutions with this shape (can also add a querystring).
+    URI=http://myhost.test/rootpath/{path}
+
+    It can also have another format, for stage variables, documented here:
+    https://docs.aws.amazon.com/apigateway/latest/developerguide/aws-api-gateway-stage-variables-reference.html#stage-variables-in-integration-HTTP-uris
+    URI=https://${stageVariables.<variable_name>}
+    This format is the same as VTL.
+
+    :param uri: the integration URI
+    :param path_parameters: the list of path parameters, coming from the parameters mapping and override
+    :param stage_variables: -
+    :return: the rendered URI
+    """
+    uri_with_path = render_uri_with_path_parameters(uri, path_parameters)
+    return render_uri_with_stage_variables(uri_with_path, stage_variables)
+
+
 def get_source_arn(context: RestApiInvocationContext):
     method = context.resource_method["httpMethod"]
     path = context.resource["path"]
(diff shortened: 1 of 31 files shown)


tests/aws/services/apigateway/test_apigateway_sqs.py
```diff
@@ -250,7 +250,7 @@ def invoke_api(url, is_valid_xml=None):
         patchOperations=[{"op": "replace", "path": "/deploymentId", "value": deployment["id"]}],
     )
 
-    response = retry(invoke_api, sleep=2, retries=10, url=invocation_url, is_valid_xml=is_valid_xml)
+    response = retry(invoke_api, sleep=2, retries=10, url=invocation_url, validate_xml=True)
```
cloutierMat, line 253: :confused: How did that work before, was it just referencing a function, so it would evaluate as `True`? Nice catch!

cloutierMat (approved):
Impressive amount and quality of changes!

It is really awesome that we can have, for now, some clear path for AWS integrations. It will be good to explore how the other services behave!

localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
```diff
@@ -36,13 +36,15 @@ def freeze_rest_api(
...
-def render_uri_with_stage_variables(uri: str, stage_variables: dict[str, str]):
+def render_uri_with_stage_variables(uri: str | None, stage_variables: dict[str, str]) -> str | None:
     """
     https://docs.aws.amazon.com/apigateway/latest/developerguide/aws-api-gateway-stage-variables-reference.html#stage-variables-in-integration-HTTP-uris
     URI=https://${stageVariables.<variable_name>}
     This format is the same as VTL, but we're using a simplified version to only replace `${stageVariables.<param>}`
     values, as AWS will ignore `${path}` for example
     """
+    if not uri:
+        return uri
```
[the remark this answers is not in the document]
bentsku: Yes, for `MOCK` implementation 😅 I got bitten by it, and it was easier to manage in one point

tests/aws/services/apigateway/test_apigateway_sqs.py
bentsku: I honestly have no idea how it worked 😅 I think it was overriding and kinda working somehow, but it's nicer like this 😬

PR #11199, opened 2024-07-13
APIGW NG refactor handlers

bentsku:
## Motivation
This PR refactors our usage of the handler chain: we would make use of response handlers to mutate the response in place, and that defeated the extensibility. 
Also, exception handlers were affected by the response handlers and it did not make a lot of sense regarding the architecture of APIGW. 

Now, all handlers are request handlers, passing data around in the context. Only the last handler is responsible for serializing the response.

We could arguably already serialize the response in the `_PROXY` integrations, but we will most probably add additional serialization logic in the Method Response handler regarding remapped headers.

This will also enable us to create response handlers which will extend the behavior, for example read all data from the context (which is mostly added and not mutated except for variables) and log it in a response handler. 

## Changes
- refactor Integration Response and Method Response handlers to be request handlers. 
- create 2 new attributes for the context, to return what is called the "endpoint response" in AWS (the integration response, but the naming is confusing), and the actual "integration response" after mapping when the integration is not of type proxy
- refactor Integration handler to not serialize the response anymore but attach it to the context
- modify unit tests to the new logic

## Testing
successful full-run with the flag forced: https://app.circleci.com/pipelines/github/localstack/localstack/26694/workflows/4b8d8b8c-a66a-4c36-9ea3-51a1dbada036

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -69,7 +69,7 @@ class EndpointResponse(BaseResponse):
 
 
 class InvocationResponse(BaseResponse):
-    """Represents the response coming after being serialized in an integration response in AWS"""
+    """Represents the response coming after being serialized in an Integration Response in AWS"""
 
     pass
 
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -51,7 +51,7 @@ class IntegrationRequest(TypedDict, total=False):
     """Body content of the request"""
 
 
-class BaseResponse(TypedDict, total=False):
+class BaseResponse(TypedDict):
     """Base class for Response objects in the context"""
 
     status_code: int
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -51,6 +51,29 @@ class IntegrationRequest(TypedDict, total=False):
     """Body content of the request"""
 
 
+class BaseResponse(TypedDict, total=False):
+    """Base class for Response objects in the context"""
+
+    status_code: int
+    """Status code of the response"""
+    headers: Headers
+    """Headers of the response"""
+    body: bytes
+    """Body content of the response"""
+
+
+class EndpointResponse(BaseResponse):
+    """Represents the response coming from an integration, called Endpoint Response in AWS"""
+
+    pass
+
+
+class InvocationResponse(BaseResponse):
+    """Represents the response coming after being serialized in an integration response in AWS"""
+
+    pass
+
+
 class RestApiInvocationContext(RequestContext):
     """
     This context is going to be used to pass relevant information across an API Gateway invocation.
@@ -82,6 +105,10 @@ class RestApiInvocationContext(RequestContext):
     """Additional $context variables available only for access logging, not yet implemented"""
     integration_request: Optional[IntegrationRequest]
     """Contains the data needed to construct an HTTP request to an Integration"""
+    endpoint_response: Optional[EndpointResponse]
+    """Contains the data returned by an Integration"""
+    invocation_response: Optional[InvocationResponse]
+    """Contains the data serialized and to be returned by an invocation"""
 
     def __init__(self, request: Request):
         super().__init__(request)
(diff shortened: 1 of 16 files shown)


tests/unit/services/apigateway/test_handler_integration_response.py
```diff
@@ -136,161 +140,133 @@ def default_context():
...
+# def create_default_response():
+#     def _response():
+#         return Response(
+#             status=200,
+#             headers={"content-type": "application/json", "header": ["multi", "header"]},
+#             response=b'{"foo":"bar"}',
+#         )
+#
+#     return _response
+#
```
bentsku, line 162: oops, will remove when rebasing

localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
```diff
@@ -51,6 +51,29 @@ class IntegrationRequest(TypedDict, total=False):
...
+
+
+class EndpointResponse(BaseResponse):
+    """Represents the response coming from an integration, called Endpoint Response in AWS"""
+
+    pass
+
+
+class InvocationResponse(BaseResponse):
+    """Represents the response coming after being serialized in an integration response in AWS"""
```
cloutierMat, line 72: Just keeping the syntax style similar :wink: 
```suggestion
    """Represents the response coming after being serialized in an Integration Response in AWS"""
```

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/gateway_exception.py
```diff
@@ -28,9 +29,6 @@ def __call__(
         context: RestApiInvocationContext,
         response: Response,
     ):
-        # Need to terminate the chain here to prevent the response handler to modify this response
-        chain.terminate()
-
```
cloutierMat, line 33: I am glad we can now get rid of that!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_response.py
```diff
@@ -77,56 +85,56 @@ def __call__(
...
-        response.headers = headers
-
-        # Body is updated last to make sure the content-length is set
         LOG.debug("Method response body after transformations: %s", body)
-        response.data = body
+        context.invocation_response = InvocationResponse(
+            body=body,
+            headers=response_headers,
+            status_code=response_status_code,
+        )
```
cloutierMat, line 132: I like how it removes the need for overwriting the previous headers! :rocket:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/method_response.py
```diff
@@ -3,15 +3,14 @@
...
+from ..context import InvocationResponse, RestApiInvocationContext
 
 LOG = logging.getLogger(__name__)
 
 
 class MethodResponseHandler(RestApiGatewayHandler):
     """
-    Currently, it seems the `MethodResponse` APIGW resource might not add logic/data to the returning
-    IntegrationResponse. We will need to investigate if this handler is needed.
+    Last handler of the chain, responsible for serializing the Response object
```
cloutierMat, line 13: :rofl: Finally found a use for it! :partying_face:

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
```diff
@@ -388,14 +393,12 @@ def invoke(self, context: RestApiInvocationContext) -> Response:
         headers.update(response_headers)
         # TODO: look into remapped headers (like HTTP_PROXY and integration response handlers)
 
-        response = Response(
+        return EndpointResponse(
             headers=headers,
-            response=lambda_response.get("body") or "",
-            status=lambda_response.get("statusCode") or 200,
+            body=to_bytes(lambda_response.get("body") or ""),
+            status_code=lambda_response.get("statusCode") or 200,
```
cloutierMat, line 399: Question: Can the status code of a lambda response be a string?

cloutierMat (approved):
All in all, this refactor seemed pretty straight forward! Good job!

It will also leave the context so much cleaner and easier to inspect!

I appreciate the continuous effort to update `"application/json"` to `APPLICATION_JSON`.

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
bentsku: Good question, I'll try to quickly check with a test. If it cannot, this should probably be added to the validation of the lambda response as well. Will update 👍

bentsku: It can yes, and it's allowed:
```
Endpoint response body before transformations: {"statusCode": "201"}
```

Updating the code now!

PR #11206, opened 2024-07-15
Add AWS validation information for es service

silv-io:
## Motivation
In an effort to have complete information on which tests are validated to be in parity with AWS the es service is still missing.

## Changes

This PR adds that missing information, with one test working on AWS, one test testing specific localstack functionality and the rest not conforming to AWS (because of a discrepancy in the create_domain endpoint).

Note that the entire test suite is still marked as flaky.


alexrashed (approved):
Awesome! The markers are looking good, thanks for checking! 🚀 
Cross-Linking https://github.com/localstack/localstack/pull/11141 here to associate to the other marker changes, and pinging @steffyP for visibility. :)

PR #11207, opened 2024-07-15
Upgrade Bookstore test to OpenSearch 2.11 and revalidate

silv-io:
## Motivation

The Bookstore tests show flaky behavior, which was supposed to be mitigated by ...
That PR fixed the usage of the cache for OpenSearch installation.
However, it can still lead to issues, because the version used for OpenSearch in the Bookstore test is not the default version and thus needs to be downloaded extra. This might again lead to rate limiting issues.

## Changes

- This PR upgrades the Bookstore test to use the current default version of OpenSearch (2.11). 
- To facilitate that, it also fixes the responses given by the compatible version check in the new version and skips some of the new keys that are in the response
- For the fix of the version output, two of the opensearch tests were converted to snapshot tests and validated against AWS. This will make changes like these easier in the future.

## Testing

- The Bookstore test should be more stable now

### DIFF
--- a/localstack-core/localstack/services/opensearch/versions.py
+++ b/localstack-core/localstack/services/opensearch/versions.py
@@ -15,8 +15,8 @@ from localstack.utils.common import get_arch
 
 # Internal representation of the OpenSearch versions (without the "OpenSearch_" prefix)
 _opensearch_install_versions = {
+    "2.13": "2.13.0",
     "2.11": "2.11.1",
-    "2.10": "2.10.0",
     "2.9": "2.9.0",
     "2.7": "2.7.0",
     "2.5": "2.5.0",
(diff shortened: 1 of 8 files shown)


localstack-core/localstack/services/opensearch/versions.py
```diff
@@ -15,8 +15,8 @@
 
 # Internal representation of the OpenSearch versions (without the "OpenSearch_" prefix)
 _opensearch_install_versions = {
+    "2.13": "2.13.0",
     "2.11": "2.11.1",
-    "2.10": "2.10.0",
```
silv-io, line 19: this version is not in the AWS response so I removed it

alexrashed (approved):
Wow, this is really an awesome set of changes!
It verifies a previously non-verified test, removes a lot of test code in favor of a proper snapshot validation, increases parity with the opensearch service, and all in order to make our acceptance test more resilient! 🦸🏽 Really kudos for the thorough fix!
:shipit: 🥳

PR #11208, opened 2024-07-15
Fix/eventbridge_v2: IAM policy enfocement for cross account

maxhoheiser:
## Motivation
Two operations are allowed to take place across the account border, one the one hand sending events from one event bridge bus to another one in a different account, on the other using `put_events` to put events directly to an event bridge bus in another account - if the permissions are set correctly.
To enforce IAM policies localstack relies internally on target arns - and the request must originate with the source account id.

## Changes
- Use the event bus target arn for internal `put_events` with TargetSenders
- Use the event bus target arn if provided to extract the region and account id to load the store for the target account id and region from the account region bundle

## Testing
Add a validated test for `put_events` directly into another account.

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -115,9 +115,14 @@ from localstack.services.events.models import (
 from localstack.services.events.replay import ReplayService, ReplayServiceDict
 from localstack.services.events.rule import RuleService, RuleServiceDict
 from localstack.services.events.scheduler import JobScheduler
-from localstack.services.events.target import TargetSender, TargetSenderDict, TargetSenderFactory
+from localstack.services.events.target import (
+    TargetSender,
+    TargetSenderDict,
+    TargetSenderFactory,
+)
 from localstack.services.events.utils import (
     extract_event_bus_name,
+    extract_region_and_account_id,
     format_event,
     get_resource_type,
     is_archive_arn,
@@ -207,7 +212,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
     ) -> CreateEventBusResponse:
         region = context.region
         account_id = context.account_id
-        store = self.get_store(context)
+        store = self.get_store(region, account_id)
         if name in store.event_buses.keys():
             raise ResourceAlreadyExistsException(f"Event bus {name} already exists.")
         event_bus_service = self.create_event_bus_service(
@@ -227,7 +232,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
     def delete_event_bus(self, context: RequestContext, name: EventBusName, **kwargs) -> None:
         if name == "default":
             raise ValidationException("Cannot delete event bus default.")
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         try:
             if event_bus := self.get_event_bus(name, store):
                 del self._event_bus_services_store[event_bus.arn]
@@ -243,7 +250,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         self, context: RequestContext, name: EventBusNameOrArn = None, **kwargs
     ) -> DescribeEventBusResponse:
         name = extract_event_bus_name(name)
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         event_bus = self.get_event_bus(name, store)
 
         response = self._event_bus_to_api_type_event_bus(event_bus)
@@ -258,7 +267,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         limit: LimitMax100 = None,
         **kwargs,
     ) -> ListEventBusesResponse:
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         event_buses = (
             get_filtered_dict(name_prefix, store.event_buses) if name_prefix else store.event_buses
         )
@@ -285,7 +296,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         policy: String = None,
         **kwargs,
     ) -> None:
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         event_bus = self.get_event_bus(event_bus_name, store)
         event_bus_service = self._event_bus_services_store[event_bus.arn]
         event_bus_service.put_permission(action, principal, statement_id, condition, policy)
@@ -299,7 +312,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         event_bus_name: NonPartnerEventBusName = None,
         **kwargs,
     ) -> None:
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         event_bus = self.get_event_bus(event_bus_name, store)
         event_bus_service = self._event_bus_services_store[event_bus.arn]
         if remove_all_permissions:
@@ -320,7 +335,9 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         event_bus_name: EventBusNameOrArn = None,
         **kwargs,
     ) -> None:
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
         event_bus_name = extract_event_bus_name(event_bus_name)
         event_bus = self.get_event_bus(event_bus_name, store)
         rule = self.get_rule(name, event_bus)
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/events/provider.py
```diff
@@ -243,7 +250,9 @@ def describe_event_bus(
         self, context: RequestContext, name: EventBusNameOrArn = None, **kwargs
     ) -> DescribeEventBusResponse:
         name = extract_event_bus_name(name)
-        store = self.get_store(context)
+        region = context.region
```
Morijarti, line 253: In case we get ARN, I believe we should use it for extracting region and account_id

localstack-core/localstack/services/events/target.py
```diff
@@ -102,14 +103,18 @@ def __init__(
...
         self.target = target
         self.rule_arn = rule_arn
         self.rule_name = rule_name
         self.service = service
+        self.region = region
+        self.account_id = account_id
 
-        self.region = extract_region_from_arn(self.target["Arn"])
-        self.account_id = extract_account_id_from_arn(self.target["Arn"])
+        self.target_region = extract_region_from_arn(self.target["Arn"])
```
joe4dev, line 116: Consider using a `@cached_property` for `target_*` https://docs.python.org/3/library/functools.html#functools.cached_property
Makes these properties immutable and seems more intention-revealing for a dependent value rather than a separate field

localstack-core/localstack/services/events/utils.py
```diff
@@ -42,6 +49,23 @@ def to_json_str(obj: Any, separators: Optional[tuple[str, str]] = (",", ":")) ->
...
+    """Returns the region and account id from the arn,
+    or falls back on the region and account id of the context"""
+    account_id = None
+    region = None
+    if ArnParser.is_arn(name_or_arn):
+        parsed_arn = parse_arn(name_or_arn)
+        region = parsed_arn.get("region")
+        account_id = parsed_arn.get("account")
+    if not account_id or not region:
+        region = context.get("region")
```
joe4dev, line 64: Can region and account_id in context be None? (typing `str` or `str | None`)

joe4dev (changes requested):
I agree with @Morijarti and think we need to check and potentially parse the `region` and `account_id` from the ARN for more operations (especially where `EventBusNameOrArn` is accepted).

Can we add regression tests to cover the scenarios described in this PR (two scenarios mentioned in the beginning)?
> Add a validated test for put_events directly into another account.
Are the tests in another PR?

localstack-core/localstack/services/events/provider.py
maxhoheiser: I manually checked and so far I had no luck using describe for an event bus in a different region / account against AWS. I believe that this behavior is only possible for `put_events` but I will add a specific test for each API endpoint that allows for event bus ARN. 
This is also why I only pars the region and account_id for `put_events` since cross region / account is only valid for this operation.

maxhoheiser: AWS has the following behavior that I could confirm against AWS for `put_events`:
- cross-account is possible with the correct permissions for custom event busses
- cross-region is not allowed for custom event busses with a client-side error
- cross-account and cross-region does not work for default event busses but there is no error message

Since this is a much larger endeavor to test all API endpoints for cross-region and cross-account, as well as add error messages for not allowing cross-region for custom event buses and not allowing region as well as cross-account for default bus but without raising an error to be fully compatible with AWS behavior - I will deal with this in a followup pr.
This PR is purely concerned with fixing the cross-account and region event bus to event bus communication for allowing IAM policy enforcement.

localstack-core/localstack/services/events/utils.py
maxhoheiser: yes if it is not ARN arn that is provided but a name

localstack-core/localstack/services/events/provider.py
```diff
@@ -356,7 +375,9 @@ def describe_rule(
         event_bus_name: EventBusNameOrArn = None,
         **kwargs,
     ) -> DescribeRuleResponse:
-        store = self.get_store(context)
+        region = context.region
+        account_id = context.account_id
+        store = self.get_store(region, account_id)
```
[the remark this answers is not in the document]
maxhoheiser: As explained above, the full implementation for cross-region / cross-account with the according tests for custom and default event busses will be implemented in a follow-up PR, non the less a first test for `put_events` is added here: https://github.com/localstack/localstack/pull/11208/commits/9170f75922b47a7e15120424e0dd84cb7a9c9de8

tests/aws/services/events/conftest.py
```diff
@@ -416,3 +416,41 @@ def _add_resource_policy_logs_events_access(log_group_arn: str):
...
+    secondary_aws_client_factory,
+    region_name,
+    secondary_region_name,
+    account_id,
+    secondary_account_id,
+):
+    def _get_primary_secondary_clients(cross_scenario: str):
+        secondary_region = secondary_region_name
+        secondary_account = secondary_account_id
+        if cross_scenario not in ["region", "account", "region_account"]:
```
joe4dev, line 433: Having that as `else` below would be more robust

localstack-core/localstack/services/events/utils.py
```diff
@@ -183,3 +208,31 @@ def re_format_event(event: FormattedEvent, event_bus_name: EventBusName) -> PutE
...
+
+
+def get_trace_header_encoded_region_account(
+    event: PutEventsRequestEntry | FormattedEvent | TransformedEvent,
+    source_region: str,
+    source_account_id: str,
+    target_region: str,
+    target_account_id: str,
+) -> str | None:
+    """Encode the original region and account_id for cross-region and cross-account
```
joe4dev, line 220: That trace header encoding is a hack we likely need to revisit once we add X-Ray support. It might be worth clarifying that it's a hack and not expected behavior (currently unclear from the comment).

tests/aws/services/events/test_events_cross_account_region.py
```diff
@@ -320,3 +318,138 @@ def test_event_bus_to_event_bus_cross_account_region(
...
+        snapshot,
+    ):
+        # Create aws clients
+        response = get_primary_secondary_client(cross_scenario)
+        aws_client = response["primary_aws_client"]
+        secondary_aws_client = response["secondary_aws_client"]
+        secondary_region_name = response["secondary_region_name"]
+        secondary_account_id = response["secondary_account_id"]
+
+        # overwriting randomized region https://docs.localstack.cloud/contributing/multi-account-region-testing/
```
joe4dev, line 346: helpful clarification 👍

tests/aws/services/events/test_events_cross_account_region.py
```diff
@@ -320,3 +318,138 @@ def test_event_bus_to_event_bus_cross_account_region(
...
+        # Create rule in secondary region / account
+        rule_name = f"test-rule-sqs-{short_uid()}"
+        secondary_aws_client.events.put_rule(
+            Name=rule_name,
+            EventPattern=json.dumps({"source": [SOURCE_PRIMARY]}),
+            EventBusName=event_bus_name,
+        )
+
+        # Create target in secondary region / account
+        target_id_sqs_primary = f"test-target-primary-sqs-{short_uid()}"
```
joe4dev, line 406: The variable name is misleading because it contradicts the comment: should it be secondary here?

tests/aws/services/events/test_events_cross_account_region.py
```diff
@@ -320,3 +318,138 @@ def test_event_bus_to_event_bus_cross_account_region(
...
+        secondary_aws_client.events.put_permission(
+            StatementId=f"TargetEventBusAccessPermission{short_uid()}",
+            EventBusName=event_bus_name,
+            Action="events:PutEvents",
+            Principal="*",
+        )
+
+        # Create SQS queue in secondary region / account
+        queue_name = f"test-queue-{short_uid()}"
+        queue_url = secondary_aws_client.sqs.create_queue(QueueName=queue_name)["QueueUrl"]
```
joe4dev, line 376: Alternatively, the `localstack.testing.pytest.fixtures.sqs_create_queue` fixture could be used by configuring it with the secondary_aws_client instead of manually creating and cleaning up the SQS queue.
SO post about indirect parametrization: https://stackoverflow.com/a/33879151

tests/aws/services/events/test_events_cross_account_region.py
```diff
@@ -320,3 +318,138 @@ def test_event_bus_to_event_bus_cross_account_region(
         PolicyName=policy_name_bus_primary_to_bus_secondary,
     )
     aws_client.iam.delete_role(RoleName=role_name_bus_primary_to_bus_secondary)
+
+
+class TestEventsCrossAccountRegion:
+    @markers.aws.validated
+    @pytest.mark.skipif(is_old_provider(), reason="Not supported in v1 provider")
+    @pytest.mark.parametrize("cross_scenario", ["account"])
```
joe4dev, line 326: I guess the plan is to add other scenarios here as well, right?
A short list in the PR description could clarify the scope of this vs. the follow-up PR (we discussed them via PM)

joe4dev (approved):
Thanks for adding the nice test scenario 👍 

Nothing blocking a merge, though some small clarify improvements would be good to fix (e.g., contradicting variable name).

I can confirm that the test fails against an older LocalStack version without these fixes 👍 
For some reason, the test cleanup blocks when I run it locally. Does that happen to you as well? It might be some JPype shutdown issue 🤔

tests/aws/services/events/test_events_cross_account_region.py
joe4dev: more of an idea than suggestion to implement, feel free to skip

Issue #11128, opened 2024-07-02
bug: GetAtt not extracting values properly from Stack Outputs when JSON form is used


lukasz-mazur2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm not sure if this is limited to SSM Parameter only, but this is what I've observed in our use cases.

We have some complex templates that utilize PyPlate macros and json formatted parameters for nested stacks. 
I've simplified the scenario to the following:

Main stack - parent.yaml:
```AWSTemplateFormatVersion: '2010-09-09'
Resources:
  
  Param:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://standard-common-modules.s3.localhost.localstack.cloud/test-child2.yaml
      Parameters:
        Value: |-
            { 
              "Fn::GetAtt": "ChildParam.Outputs.Value"
            }
```
 
Child stacks:
- test-child2:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform:
  - PyPlate
Parameters:
  Value:
    Type: String
Resources:
  ChildParam:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://standard-common-modules.s3.localhost.localstack.cloud/test-child.yaml
  Param:
    DependsOn: ChildParam
    Type: AWS::SSM::Parameter
    Properties:
      Name: test-param
      Type: String
      Value: |
        #!PyPlate
        import json
        output = json.loads(params['Value'])
Outputs:
  Name:
    Value: !Ref Param
  Value:
    Value: !GetAtt Param.Value
```

- test-child:
```
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  Param:
    Type: AWS::SSM::Parameter
    Properties:
      Name: child-param-name
      Type: String
      Value: child-2-param-name
Outputs:
  Name:
    Value: !Ref Param
  Value:
    Value: !GetAtt Param.Value
```

Everything seem to deploy properly from cloudformation perspective, but the actual value that GetAtt retrieved and put into the parameter value are all the available stack outputs

```$ awslocal ssm get-parameter --name test-param
{
    "Parameter": {
        "Name": "test-param",
        "Type": "String",
        "Value": "{'Name': 'child-param-name', 'Value': 'child-2-param-name'}",
        "Version": 1,
        "LastModifiedDate": 1719925660.265,
        "ARN": "arn:aws:ssm:eu-west-1:000000000000:parameter/test-param",
[...]

PR #11212, opened 2024-07-16
cloudformation: resolve attribute sub names from model instance

sannya-singal:
## Motivation
When resolving the references in the cloudformation template for `Fn::GetAtt` we do not consider the attribute sub name. For example: 
In `"Fn::GetAtt": "ChildParam.Outputs.Value"`, where attribute reference is `ChildParam.Outputs.Value` the:
- resource logical id is `ChildParam`
- attribute name is `Outputs`
 
we do not further consider to fetch the `Value` attribute from the resource properties. 

## Changes
This PR:
- adds an additional attribute subname if it exists and gets it value from the resource properties of the model instance. 
- adds an AWS validated test to verify the results. 

Closes https://github.com/localstack/localstack/issues/11128


pinzon (changes requested):
Hi @sannya-singal  Thank you for tackling this. The modifications to the template deployer look pretty good. 

This test is a bit complicated to understand (even for me and I was involved in this) so I'm requesting a couple of changes that would improve it:

1. Adding a detailed doctstring to understand what the purpose of the test and why it was necessary (Like you PR comment). For any future maintenance that the test would receive.
2. Reducing the structure from 3 stacks to just 2, a parent and a child. This would make it even more easier to understand.

pinzon (approved):
Thanks for making the changes!

PR #11214, opened 2024-07-16
Switch default provider for cloudwatch

steffyP:
## Motivation
We introduced the cloudwatch v2 provider a few months ago. 
There hasn't been a lot adoption yet, but the new provider was able to resolve some reported issues already. 

Moving forward, the new provider should be the default, so we also get more feedback and can fix potential oversights and issues only in the new provider.

## Changes
* switch the `default` provider to use v2
* adapt the `old_provider` check for the test-skips
* adapt the circleci config to run the test suite also for v1
* fix some parameters for the composite alarm (revealed by a cloudformation test)
* re-create snapshot for cloudformation test and adapt transformers
* add the state-visitor pattern to v2 provider to fix persistence issue with cloudpods

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_cloudwatch.py
+++ b/tests/aws/services/cloudformation/resources/test_cloudwatch.py
@@ -1,7 +1,12 @@
 import json
 import os
+import re
+from typing import List
+
+from localstack_snapshot.snapshots.transformer import KeyValueBasedTransformer
 
 from localstack.testing.pytest import markers
+from localstack.testing.snapshots.transformer_utility import PATTERN_ARN
 from localstack.utils.strings import short_uid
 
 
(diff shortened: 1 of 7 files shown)


steffyP:
companion test-run to verify persistence for cloudpods is working now as well: https://github.com/localstack/localstack-ext/actions/runs/9969809252/job/27547788677

tests/aws/services/cloudformation/resources/test_cloudwatch.py
```diff
@@ -43,24 +48,34 @@ def test_alarm_creation(deploy_cfn_template, snapshot):
...
+
     stack = deploy_cfn_template(
         template_path=os.path.join(
             os.path.dirname(__file__), "../../../templates/cfn_cw_composite_alarm.yml"
         ),
     )
     composite_alarm_name = stack.outputs["CompositeAlarmName"]
+
+    def alarm_action_name_transformer(key: str, val: str):
+        if key == "AlarmActions" and isinstance(val, List) and len(val) == 1:
```
bentsku, line 65: I believe we should use the `list` type here and not from `typing`? 
```suggestion
        if key == "AlarmActions" and isinstance(val, list) and len(val) == 1:
```

bentsku (approved):
LGTM! Nice, great to see it become default! 

The updated CI run passed and is properly using the v1 provider, so all is good! 🚀 

https://app.circleci.com/pipelines/github/localstack/localstack/26783/workflows/95dc3fbb-2ebd-4c04-8526-a52404d07b94/jobs/228260

Just a minor comment regarding the use of `isinstance` for the transformer which should maybe use the proper type from Python. Other than that, good to go! 💯

pinzon (approved):
Happy to see this!

Issue #11175, opened 2024-07-10
bug: AWS::ResourceGroups::Group parameter validation fails with valid configuration


AliSajid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am trying to build a small SAM-based app and using localstack to test it. One of the resources I creating is an AWS::ResourceGroups::Group. When I run `samlocal deploy`, I see the following error when it comes to the specific resource:

```
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ResourceStatus                              ResourceType                                LogicalResourceId                           ResourceStatusReason                      
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
CREATE_IN_PROGRESS                          AWS::CloudFormation::Stack                  cloudcrier-app                              -                                         
UPDATE_IN_PROGRESS                          AWS::SQS::Queue                             WeatherQueue                                -                                         
UPDATE_IN_PROGRESS                          AWS::ResourceGroups::Group                  ApplicationResourceGroup                    -                                         
UPDATE_IN_PROGRESS                          AWS::ApplicationInsights::Application       ApplicationInsightsMonitoring               -                                         
UPDATE_IN_PROGRESS                          AWS::Lambda::Function                       QueingFunction                              -                                         
UPDATE_IN_PROGRESS                          AWS::IAM::Role                              QueingFunctionRole                          -                                         
UPDATE_IN_PROGRESS                          AWS::Lambda::Function                       NWSApiFunction                              -                                         
UPDATE_IN_PROGRESS                          AWS::IAM::Role                              NWSApiFunctionRole                          -                                         
UPDATE_IN_PROGRESS                          AWS::Lambda::Function                       OpenWeatherApiFunction                      -                                         
UPDATE_IN_PROGRESS                          AWS::IAM::Role                              OpenWeatherApiFunctionRole                  -                                         
UPDATE_IN_PROGRESS                          AWS::DynamoDB::Table                        TransactionTable                            -                                         
UPDATE_COMPLETE                             AWS::SQS::Queue                             WeatherQueue                                -                                         
UPDATE_FAILED                               AWS::ResourceGroups::Group                  ApplicationResourceGroup                    Parameter validation failed: Missing      
                                                                                                                                    required parameter in ResourceQuery:      
                                                                                                                                    "Query"                                   
CREATE_FAILED                               AWS::CloudFormation::Stack                  cloudcrier-app                              Parameter validation failed: Missing      
[...]

PR #11217, opened 2024-07-16
fix defaults for ResourceGroups::Group provider

pinzon:
## Motivation
As mentioned in the issue #11175 the resource provider for AWS::ResourceGroups::Group is supposed to handle a Group definition without a Query if the Type is CLOUDFORMATION_STACK_1_0. 

## Changes
- Replace auto generation of the Name attribute as it's a required attribute.
- Add default query

## Testing
- New CFn test that replicates the issue


Morijarti (approved):
LGTM

Issue #11048, opened 2024-06-18
bug: get-parameter and get-parameters on SSM does not work with ARNs (Localstack 3.5.0)


wbeardall:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Queries to Localstack SSM endpoints with the `get-parameter` or `get-parameters` commands do not work if parameter names are provided as ARNs. This appears to be due to the internal LocalStack parameter validation disallowing forward slashes in SSM parameter names. We observe the following:

```
$ awslocal ssm get-parameter --name arn:aws:service:us-east-1:0000000000:parameter/myparam
An error occurred (ValidationException) when calling the GetParameter operation: Parameter name: can't be prefixed with "ssm" (case-insensitive). If formed as a path, it can consist of sub-paths divided by slash symbol; each sub-path can be formed as a mix of letters, numbers and the following 3 symbols .-_
```

Removing the forward slash passes the input validation, but obviously fails to fetch a parameter:

```
$ awslocal ssm get-parameter --name arn:aws:service:us-east-1:0000000000:parametermyparam
An error occurred (ParameterNotFound) when calling the GetParameter operation: Parameter arn:aws:service:us-east-1:0000000000:parametermyparam not found.
```

### Expected Behavior

`get-parameter` and `get-parameters` should allow ARNs in names, following the [official docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ssm/get-parameters.html#options)

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal ssm get-parameter --name arn:aws:service:us-east-1:0000000000:parameter/myparam

### Environment

```markdown
- OS: Sonoma 14.5
- LocalStack:
  LocalStack version: 3.5.1.dev20240618022512
  LocalStack Docker image sha: sha256:5cd0557de2fdfac98d8d26d2f861b8266dcfc07ed09dbdacad7dc21ee2560310
  LocalStack build date: 2024-06-18
  LocalStack build git hash: 666e239
```

mikeshultz:
The root cause of this may also be preventing the use of SSM secret string params in ECS task definitions.

[...]

PR #11218, opened 2024-07-16
allow SSM::GetParameter action to use ARNs instead of only names

pinzon:
## Motivation
As shown in the issue #11048. The SSM service allows the usage of ARN's as Names in the GetParameter action. 

## Changes
- Extend the name validation function to detect if it's an ARN and extract the parameter name if it is.

## Testing
- New AWS validated test that confirms the behavior.


dominikschubert (approved):
LGTM, let's get it into 3.6 :rocket:

PR #11219, opened 2024-07-16
force all APIGW providers to use NextGen if main one is set

bentsku:
## Motivation
We're planning to advertise the new implementation of API Gateway "Next Gen" in the next release.

However, we will need all our implementation of API Gateway (with upstream) to be on the same page, and to not try to register the same routes with different implementations. 

This PR makes sure we override the different versions of APIGW to use the new implementation if the flag is set.

Also sneaked a small refactor we've talked with @cloutierMat in order to be easier to override behavior. 

## Changes
- update `apigatewayv2` and `apigatewaymanagementapi` to also use `next_gen` if set. 
- small refactor for the endpoint

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -1038,6 +1038,15 @@ LEGACY_SNS_GCM_PUBLISHING = is_env_true("LEGACY_SNS_GCM_PUBLISHING")
 
 # Whether the Next Gen APIGW invocation logic is enabled (handler chain)
 APIGW_NEXT_GEN_PROVIDER = os.environ.get("PROVIDER_OVERRIDE_APIGATEWAY", "") == "next_gen"
+if APIGW_NEXT_GEN_PROVIDER:
+    # in order to not have conflicts with different implementation registering their own router, we need to have all of
+    # them use the same new implementation
+    if not os.environ.get("PROVIDER_OVERRIDE_APIGATEWAYV2"):
+        os.environ["PROVIDER_OVERRIDE_APIGATEWAYV2"] = "next_gen"
+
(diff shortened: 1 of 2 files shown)


cloutierMat (approved):
That refactor already seems to work wonder! :wink:

localstack-core/localstack/config.py
```diff
@@ -1038,6 +1038,15 @@ def populate_edge_configuration(
 
 # Whether the Next Gen APIGW invocation logic is enabled (handler chain)
 APIGW_NEXT_GEN_PROVIDER = os.environ.get("PROVIDER_OVERRIDE_APIGATEWAY", "") == "next_gen"
```
alexrashed, line 1040: nit: This wouldn't work if a user would use `PROVIDER_OVERRIDE_APIGATEWAYV2=next_gen`, right?
But I don't think this would be necessary, it's just a matter of communication with the users in our release notes / announcements.

alexrashed (approved):
LGTM! I only had a comment on the potential usages of other provider overrides, but as mentioned in the comment, it's not really necessary.

localstack-core/localstack/config.py
bentsku: Exactly, yes, I think we will communicate only about this one override, to not confuse users, as this can get quite complicated quickly. The other idea would be to have a specific flag like `ENABLE_APIGW_NEXT_GEN` which would overwrite the 3 `PROVIDER_OVERRIDE_*`, not sure what is cleaner?

alexrashed: No, I think it's totally fine to stick to how it is. This is just an intermediary solution, and no users will just jump on this without reading about this in the release notes or an announcement, where we need to make sure that it's clear with env var to set. 👍🏽

PR #11220, opened 2024-07-16
fix: kinesis CBOR blob handling

simonrw:
## Motivation

A report from a customer showed that a round trip of data through a Kinesis stream from LocalStack did not preserve the data content.

For example:

* `PutRecord` in Kinesis stream, `"test data"`
* `GetRecords` from kinesis stream, convert 1st record to `String`, which turns out different from "test data": `��-u�Z`

They were using the Java SDK. When testing on AWS, a `GetRecords` request returns the message body (decoded as a String) of `"test data"` correctly.

Since this is exercised via the Java SDK, it is the handling of the [CBOR](https://cbor.io/) format that it as fault. It turned out that we did not parse `CBOR` `blob`s correctly.

* #11133 made some changes to the CBOR parsing to do with timestamps milliseconds vs seconds.
* As part of this change, the [`BaseJSONRequestParser._parse_blob`](https://github.com/localstack/localstack/blob/a4095638a6c92a70e15aebb07b9520e8907c3e50/localstack-core/localstack/aws/protocol/parser.py#L894-L901) method was removed.
* This meant that blob data types (like plain strings) were parsed using the [`RequestParser._parse_blob`](https://github.com/localstack/localstack/blob/a87cba7f0308262aa518368236700a20855a411e/localstack-core/localstack/aws/protocol/parser.py#L297-L299) method, which incorrectly base64-decoded the value.
* This PR also updated `tests/unit/aws/protocol/test_parser.py::test_json_cbor_blob_parsing` to test this incorrect behaviour.

## Changes

This PR reverts the removal of `BaseJSONRequestParser._parse_blob` and updates the relevant test accordingly.

### DIFF
--- a/localstack-core/localstack/aws/protocol/parser.py
+++ b/localstack-core/localstack/aws/protocol/parser.py
@@ -898,6 +898,15 @@ class BaseJSONRequestParser(RequestParser, ABC):
             return self._convert_str_to_timestamp(node, self.CBOR_TIMESTAMP_FORMAT)
         return super()._parse_timestamp(request, shape, node, uri_params)
 
+    def _parse_blob(
+        self, request: Request, shape: Shape, node: bool, uri_params: Mapping[str, Any] = None
+    ) -> bytes:
+        if isinstance(node, bytes) and request.mimetype.startswith("application/x-amz-cbor"):
+            # CBOR does not base64 encode binary data
+            return bytes(node)
+        else:
+            return super()._parse_blob(request, shape, node, uri_params)
+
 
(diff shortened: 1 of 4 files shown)


alexrashed (approved):
Thanks a lot for digging into this issue (which is quite tedious due to the lack of CBOR support in botocore)!
And I'm sorry I introduced this regression with the last iteration of the CBOR support... In retrospect I can't really tell what lead me to the conclusion that the CBOR content needs to be parsed as base64 too... 😅
The changeset looks good, and your analysis was great and thorough! 💯

Issue #11224, opened 2024-07-17
bug: AttributeError: 'NoneType' object has no attribute 'to_bytes'


devinrsmith:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After upgrading from localstack/localstack:3.4 to localstack/localstack:3.5, I'm no longer able to put. Here's a the log with `DEBUG=1`:

```
LocalStack supervisor: starting                                                                                                                                                                                                                                                                                                                     
LocalStack supervisor: localstack process (PID 18) starting                                                                                                                                                                                                                                                                                         
2024-07-17T21:03:50.424 DEBUG --- [  MainThread] l.utils.docker_utils       : Using SdkDockerClient. LEGACY_DOCKER_CLIENT: False, SDK installed: True                                                                                                                                                                                               
2024-07-17T21:03:50.557  WARN --- [  MainThread] l.services.internal        : Enabling diagnose endpoint, please be aware that this can expose sensitive information via your network.                                                                                                                                                              
2024-07-17T21:03:50.565 DEBUG --- [  MainThread] plux.runtime.manager       : instantiating plugin PluginSpec(localstack.runtime.components.aws = <class 'localstack.aws.components.AwsComponents'>)                                                                                                                                                
2024-07-17T21:03:50.565 DEBUG --- [  MainThread] plux.runtime.manager       : loading plugin localstack.runtime.components:aws                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                    
LocalStack version: 3.5.0                                                                                                                                                                                                                                                                                                                           
LocalStack build date: 2024-06-13                                                                                                                                                                                                                                                                                                                   
LocalStack build git hash: 54edb407d                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                    
2024-07-17T21:03:50.565 DEBUG --- [  MainThread] localstack.utils.run       : Executing command: rm -rf "/tmp/localstack"
...
2024-07-17T21:03:55.042 DEBUG --- [et.reactor-0] plux.runtime.manager       : loading plugin localstack.aws.provider:s3:default                                                                                                                                                                                                                     
2024-07-17T21:03:55.277  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.CreateBucket => 200                                                                  
2024-07-17T21:03:55.325 ERROR --- [et.reactor-1] l.aws.handlers.logging     : exception during call chain                                                                 
[...]

PR #11225, opened 2024-07-17
fix S3 CRC32 checksum with empty body logic

bentsku:
## Motivation
As reported by #11224, we had an issue when the body of an S3 object was empty with a CRC32 checksum. 
This was due to a flawed logic always considering the `Checksum.update()` was called, and not properly initializing the checksum object.

## Changes
- fix the logic in the CRC32 util
- add a test with an empty body to assert the fix
- update a previous test which did not declare the checksum trailing header, which made the test not validate the checksum

_fixes #11224_


cloutierMat (approved):
LGTM! :rocket: 

Good catch!

PR #11229, opened 2024-07-18
fix SOURCES.txt by avoiding plux entrypoints command

alexrashed:
## Motivation
We recently switched to namespace packages in this repo (with https://github.com/localstack/localstack/pull/11190) as well as in downstream repos, and we are currently trying to upgrade the CLI according to the new import paths with https://github.com/localstack/localstack-cli/pull/27.
During this upgrade, we stumbled upon an issue with the generated `SOURCES.txt` containing an absolute path (which kills the Windows PyInstaller build of the CLI), and could potentially impact Windows users installing the python package with the CLI install.
It turns out that this issue is somehow caused by `plux entrypoints`. More info can be found in https://github.com/localstack/plux/issues/23.

## Changes
- Temporarily replaces `plux entrypoints` with `setup plugins egg_info` to avoid the absolute path in `SOURCES.txt`.

## Testing
- See https://github.com/localstack/plux/issues/23 for more infos on how to reproduce the issue / test the fix.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -172,7 +172,7 @@ FROM base
 COPY --from=builder /opt/code/localstack/.venv /opt/code/localstack/.venv
 
 # add project files necessary to install all dependencies
-ADD Makefile pyproject.toml VERSION ./
+ADD Makefile pyproject.toml VERSION setup.py ./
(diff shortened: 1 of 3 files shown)


dominikschubert (approved):
LGTM, great investigation work :+1:

PR #11230, opened 2024-07-18
Fix: Set ci logs to default value

maxhoheiser:
## Motivation
This PR: https://github.com/localstack/localstack/pull/10890 wrongly sets the default value for `log_cli` to true, as discussed we want to keep ci logs cleaner this is why the decision has been made to set the default value to false, which is fix with the current PR.


simonrw (approved):
Great, thanks for fixing!

PR #11233, opened 2024-07-18
Group lambda event source mapping tests

joe4dev:
## Motivation

Group all Lambda ESM-related tests for easier selection with the upcoming test runner for ESM v2.

## Changes

* Move Lambda ESM tests into a separate `event_source_mapping` directory/Python package.
* Fix relative imports of Lambda function code in [fd973cc](https://github.com/localstack/localstack/pull/11233/commits/fd973ccfdfe63e9b20418d7bb92856cd1af624a1)

## Notes

Annoyingly, Git doesn't detect renaming of validation files after fixing their paths despite separating the rename into a dedicated commit [here](https://github.com/localstack/localstack/commit/9b2561f2b142f094fd8824eec109966438f0fe95). The make auxiliary updates make it quite cumbersome to perform simple test refactorings :(
-> We can ignore the history for validation files in such case as discussed in the core alignment sync 2024-07-19

Any change in `.test_durations` also skips the test selection
-> We can ignore updating these as discussed in the core alignment sync 2024-07-19


alexrashed (approved):
LGTM! 🧹 Makes sense to structure these tests. 💯

dominikschubert (approved):
LGTM! As discussed, it's fine if the history of the snapshot files is not 100% retrace-able :+1:

PR #11226, opened 2024-07-17
apigw next gen add circle ci pipeline

cloutierMat:
## Motivation

Now that we have a set of stable feature for the apigw Next Gen and all tests are now green :partying_face: . We are adding a new pipeline for apigw to help ensure we don't introduce any regression as we keep working on new features and extend the work to apigwv2.

PR #11198 enabled all tests and links to run https://app.circleci.com/jobs/github/localstack/localstack/227232 that was executed with the flag enabled.

## Changes

A new pipeline for circle-ci that sets `PROVIDER_OVERRIDE_APIGATEWAY=next_gen`

PR #11234, opened 2024-07-18
APIGW NG Refactor IntegrationRequest for Proxy Integration

bentsku:
## Motivation
~This PR is based on #11227 and #11226.~

As part of the header remapping work, it has been shown by @cloutierMat that `HTTP_PROXY` integration still supported Request Parameters Mappings, which isn't exactly documented.

To properly support this (and also Private API in the future), it made sense to always create an `IntegrationRequest` object to attach to the context so that the PROXY integration can use it: they don't need to render their own URI. 

I've also sneaked a little fix to a bug put into light by our tests: Lambda URL did not support multi value query string parameters. I've simplified the logic there and also wrote a specific test for it. 

## Changes
- refactor IntegrationRequest handler to always create an IntegrationRequest and attach it to the context
- refactor `AWS_PROXY` and especially `HTTP_PROXY` to make use of the new IntegrationRequest (AWS_PROXY still needs the Invocation Request, as it needs quite a lot of the original request data)
- fixed Lambda URL multi value query string

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -51,25 +51,49 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         integration: Integration = context.resource_method["methodIntegration"]
         integration_type = integration["type"]
 
-        if integration_type in (IntegrationType.AWS_PROXY, IntegrationType.HTTP_PROXY):
-            # `PROXY` types cannot use integration mapping templates
-            return
-
-        # find request template to raise UnsupportedMediaTypeError early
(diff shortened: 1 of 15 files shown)


cloutierMat (approved):
Awesome work! Really happy to see this shift! :rocket: 

Great fix in lambda as well!

bentsku:
@dfangl @joe4dev @dominikschubert I'll go ahead and merge the PR as we need it to continue building and finishing APIGW. 
As the Lambda fix feels very close to APIGW (the logic is so similar, and parts of the event also shows the Lambda URL to be behind an HTTP API from APIGW v2 (the `routeKey` and `stage` field 😄), I feel like this fix is safe.

However, if you have any comment, still feel free to review and I'll address them in a follow up PR tomorrow first thing.

Issue #11164, opened 2024-07-09
bug: CloudFormation and boolean parameters passed to a NestedStack


lukasz-mazur2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

It appears that there is an issue passing boolean valued to a nested stack:

Main template:
```
AWSTemplateFormatVersion: '2010-09-09'

Resources:
  TestMethod:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://common-modules.s3.localhost.localstack.cloud/child.yaml
      Parameters:
        ApiKeyRequired: true
```

Nested stack template:
```
AWSTemplateFormatVersion: '2010-09-09'
Parameters:
  RestApiId:
    Type: String
    Default: dummy-api-id
  ResourceId:
    Type: String
    Default: dummy-resource-id
  ApiKeyRequired:
    Type: String
    AllowedValues:
      - true
      - false

Resources:
  TestMethod:
    Type: AWS::ApiGateway::Method
    Properties:
      ApiKeyRequired:
        Ref: ApiKeyRequired
      AuthorizationType: AWS_IAM
      HttpMethod: GET
      ResourceId:
        Ref: ResourceId
      RestApiId:
        Ref: RestApiId
```

Outcome:
```
2024-07-09T13:58:06.748  INFO --- [et.reactor-0] l.request.internal.aws     : AWS cloudformation.CreateStack => 200
2024-07-09T13:58:06.752 DEBUG --- [et.reactor-1] rolo.gateway.wsgi          : POST localhost:4566/
2024-07-09T13:58:06.753 DEBUG --- [et.reactor-1] l.aws.protocol.serializer  : No accept header given. Using request's Content-Type (application/x-www-form-urlencoded; charset=utf-8) as preferred response Content-Type.
2024-07-09T13:58:06.754  INFO --- [et.reactor-1] l.request.internal.aws     : AWS cloudformation.DescribeStacks => 200
2024-07-09T13:58:06.780 DEBUG --- [functhread85] l.s.c.e.template_deployer  : Handling "Add" for resource "TestMethod" (1/1) type "AWS::ApiGateway::Method" (should_deploy=True)
2024-07-09T13:58:06.782 DEBUG --- [functhread85] l.s.c.e.template_deployer  : Error applying changes for CloudFormation stack "lm-test-3-TestMethod-f966bca4": Parameter validation failed:
[...]

PR #11236, opened 2024-07-18
refactor AWS::APIGW::Method resource provider

pinzon:
## Motivation
The issue #11164 mentions that the APIGW::Method resource is not able to deploy when a boolean parameter is passed from a parent stack due to being converted into a string. This is the correct behavior but it shows that our Resource provider is missing  another value type transformation. Instead of just adding it. I refactore the code into using the new utility for parameters and type conversions introduced in #11178.

## Changes
- Removed almost all the manual parameter selections and the type transformations

## Testing
- pipeline should be green

### DIFF
--- a/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_method.py
+++ b/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_method.py
@@ -130,15 +130,13 @@ class ApiGatewayMethodProvider(ResourceProvider[ApiGatewayMethodProperties]):
 
         responses = model.get("MethodResponses", [])
         for response in responses:
-            kwargs = util.convert_request_kwargs(
-                response, operation_model("PutMethodResponse").input_shape
-            )
-
             apigw.put_method_response(
                 restApiId=model.get("RestApiId"),
                 resourceId=model.get("ResourceId"),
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_method.py
```diff
@@ -100,75 +99,45 @@ def create(
...
 
             apigw.put_method_response(
-                restApiId=api_id,
-                resourceId=res_id,
-                httpMethod=model["HttpMethod"],
-                statusCode=str(response["StatusCode"]),
+                restApiId=model.get("RestApiId"),
+                resourceId=model.get("ResourceId"),
+                httpMethod=model.get("HttpMethod"),
                 **kwargs,
```
cloutierMat, line 141: nit: Some of the uses of the utils are done inline and some first create a `kwargs` variable. It doesn't seem like `kwargs` is used otherwise. Might be more consistent to always use inline? Unless I am missing something? :thinking:

cloutierMat (approved):
Awesome work @pinzon, thanks for tackling this!

It shows how powerful that new util is! The code is now much simpler. :rocket: 

just left a small nit!

PR #11246, opened 2024-07-23
fix typing of kwargs for lambda URL router

bentsku:
## Motivation
The Lambda URL router typing had a small issue with `kwargs`, where it was using a mapping for it instead of one type (we could also use a `TypedDict` with `Unpack`, see https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs). 

See also the discussion here: https://github.com/localstack/localstack/pull/11069#discussion_r1650648611 and the related Python PEP484: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

## Changes
- modify the typing used to the `kwargs` argument 

(This is related to https://github.com/localstack/localstack/pull/11245#issuecomment-2244657232)

### DIFF
--- a/localstack-core/localstack/services/lambda_/urlrouter.py
+++ b/localstack-core/localstack/services/lambda_/urlrouter.py
@@ -54,7 +54,11 @@ class FunctionUrlRouter:
         )
 
     def handle_lambda_url_invocation(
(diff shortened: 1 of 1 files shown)


joe4dev (approved):
Learned something 😮 

Thank you for clearly describing the context @bentsku 👏 💯

PR #11249, opened 2024-07-23
upgrade rolo to 0.7.0

thrau:
## Motivation

Upgrading to rolo 0.7.0 which includes support + fix on top of 0.6.0 for pydantic.

There was a bit of code refactoring for 0.7, which may cause breakage. Will follow up with rolo patch releases to fix.

## Changes

* pydantic models can now be used in routes

### DIFF
--- a/localstack-core/localstack/http/dispatcher.py
+++ b/localstack-core/localstack/http/dispatcher.py
@@ -1,9 +1,9 @@
 from json import JSONEncoder
 from typing import Type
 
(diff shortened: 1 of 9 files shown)


bentsku (approved):
LGTM! The changes in this PR looks right, and the full test suite is green now. 

It seems we still have issues upstream, but I'm not sure if it's a caching issue or not, it looks like it's using the `0.5.1` version, looking at logs. 

Hoping all goes well for the release 😄

PR #11250, opened 2024-07-23
APIGW NG fix _user_request_ format parsing

bentsku:
## Motivation
While implement more changes, I've spotted an issue with the `_user_request_` format style, there was a bad assumption from my side that the stage was before the path, but for some reason the format is very different. 

## Changes
- fix removing the stage from the path when using `_user_request_` style format


cloutierMat (approved):
Good catch! A bit of an oversight, glad you caught it! :rocket:

PR #11251, opened 2024-07-23
APIGW NG add routing exceptions

bentsku:
## Motivation
There was a TODO left regarding the Exception to be raised when we have a routing error. This one of the bad one in APIGW, where it will raise a 403 exception when you're trying to send a request to something you haven't defined. 

From the documentation:
> The gateway response for a missing authentication token error, including the cases when the client attempts to invoke an unsupported API method or resource.

In some raised exceptions, I've also forgot to add log statements to help the user to understand why it is raising an exception.

## Changes
- implement the `MissingAuthenticationToken` exception when we have a routing exception
- write an AWS validated test
- add missing log statements


cloutierMat (approved):
Great addition! Good logging! I still experience trauma from the `MissingAuthenticationToken` error! Glad to see we can help a bit in that regard now!

Issue #11186, opened 2024-07-11
bug: DynamoDB delete_expired_items not working properly "Attribute name is a reserved keyword"


cgaube:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the delete expired items endpoint `curl -X DELETE localhost:4566/_aws/dynamodb/expired`
and the function https://github.com/localstack/localstack/blob/b58b0f8ef6b82d38bb847b843e111e98d6c40acf/localstack-core/localstack/services/dynamodb/provider.py#L430 is triggered an error appear

```
 An error occurred when deleting expired items from table XXXXX: An error occurred (ValidationException) when calling the Scan operation: Invalid FilterExpression: Attribute name is a reserved keyword; reserved keyword: TTL
```

The endpoint returns a 200 with `{"ExpiredItems": 0}`   which is invalid

The issue I believe is because the scan function is improperly generated 
it should probably use the `#ttl` notation

`FilterExpression="#ttl <= :threshold",`

instead of `FilterExpression=f"{attribute_name} <= :threshold",` 

### Expected Behavior

Expired items are deleted from dynamo db

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```
awslocal dynamodb create-table \
    --table-name example-table \
    --attribute-definitions \
        AttributeName=id,AttributeType=S \
        AttributeName=ttl,AttributeType=N \
    --key-schema \
        AttributeName=id,KeyType=HASH \
    --billing-mode PAY_PER_REQUEST \
    --tags Key=Name,Value=example-table

awslocal dynamodb update-time-to-live \
    --table-name example-table \
    --time-to-live-specification Enabled=true,AttributeName=ttl

# Add items
# Add the expired item to the DynamoDB table
awslocal dynamodb put-item \
    --table-name example-table \
    --item '{
        "id": {"S": "item1"},
        "ttl": {"N": "1720713091"}
    }'

curl -X DELETE localhost:4566/_aws/dynamodb/expired
```

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

giograno:
[...]

PR #11254, opened 2024-07-23
fix: ttl with a reserved keyword as attribute

giograno:
## Motivation

In our TTL feature, we use a `Scan` operation to look for the expired keys to be deleted.
An expression with a reserved keyword (e.g., `ttl` might lead to errors).
This issue was raised in #11186.

## Changes
- Use expression attribute names (see [docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ExpressionAttributeNames.html)) in the `Scan` operation to avoid the mentioned issue.

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -97,6 +97,8 @@ class TestDynamoDB:
     @markers.aws.only_localstack
     def test_time_to_live_deletion(self, aws_client, ddb_test_table, cleanups):
         table_name = ddb_test_table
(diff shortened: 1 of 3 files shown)


tests/aws/services/dynamodb/test_dynamodb.py
alexrashed, line 1: nit: Nice, you changed the test to use a keyword where we know that it is reserved. It would be great if you could mention in a comment that the name of the attribute here is actually important for the test.

alexrashed (approved):
Awesome, thanks a lot for jumping on this! Really great issue report from @cgaube, nice and clean fix! 🚀

PR #11256, opened 2024-07-24
fix APIGW NG response enricher handler

bentsku:
## Motivation
Following #11237 and a last minute change + the new handling for routing with #11251, the default value for `context.integration` was not set. This PR handles the case where we fail early in the chain and the integration is not set yet

## Changes
- update the default value
- add a unit test to catch the regression

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -116,6 +116,7 @@ class RestApiInvocationContext(RequestContext):
         self.invocation_request = None
         self.resource = None
         self.resource_method = None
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
```diff
@@ -116,6 +116,7 @@ def __init__(self, request: Request):
         self.invocation_request = None
         self.resource = None
         self.resource_method = None
+        self.integration = None
```
cloutierMat, line 119: Oops! :eyes:

cloutierMat (approved):
Nice catch! Thanks for adding the extra test too~ :rocket:

PR #11257, opened 2024-07-24
add APIGW NG composite handler

bentsku:
## Motivation
As needed upstream, to allow greater flexibility, we would need a handler to modify the request after it has been parsed. 
We also need a small refactor in order to be able to act on the context from the `Endpoint`. 

We can merge this after the 3.6 release as it will break upstream dependencies. 

## Changes
- add a composite handler
- refactor the endpoint


cloutierMat (approved):
Awesome! I am glad we had that convo! This will be much more intuitive to use, I believe! :rocket:

PR #11261, opened 2024-07-24
fix flaky transcribe tests by pre-downloading dependencies

bentsku:
## Motivation
While running the pipeline to check for the release, `test_transcribe_happy_path` flaked again. It is also marked flaky on CircleCI. 

See failures: 
- https://app.circleci.com/pipelines/github/localstack/localstack/26973/workflows/543f0989-034b-4f92-bfbb-584da606b29b/jobs/230534/tests
- https://app.circleci.com/pipelines/github/localstack/localstack/26997/workflows/4031f23f-ef93-4557-b000-588789e68c6a/jobs/230767
- https://app.circleci.com/pipelines/github/localstack/localstack/26997/workflows/4031f23f-ef93-4557-b000-588789e68c6a/jobs/230769

We're using the same logic as in opensearch tests, where a check in the global AWS `conftest.py` checks if there are any tests in the test collection that would match the test parent (test file or test class), and if there is, we already trigger the async download of the dependencies early. When there is the test selection, this does not do a lot, but for full-runs it will help the dependencies being downloaded before the actually tests run.

To avoid this issue, I'm using a class fixture pre-downloading the dependencies with a longer timeout, but with a quick failure mode, to avoid waiting on the transcription job if it fails asynchronously

## Changes
- setup pre-downloading of Vosk, ffmpeg and the Vosk model used

## Testing

Running the pipeline a few times to be sure it's hardened:
- ✅ https://app.circleci.com/pipelines/github/localstack/localstack/27020/workflows/b35769d6-3dca-46bb-8c96-9b896193f270
-  ✅ https://app.circleci.com/pipelines/github/localstack/localstack/27020/workflows/393f701b-07f0-4a4a-8de5-36532cae3239
- ✅ https://app.circleci.com/pipelines/github/localstack/localstack/27020/workflows/6f510779-592f-43fc-b75c-416bd3f93ca1
- ✅ https://app.circleci.com/pipelines/github/localstack/localstack/27023/workflows/dddd5bec-1cdd-4bdb-abff-a4722088a632 (after PR comments addressed)
- ✅ https://app.circleci.com/pipelines/github/localstack/localstack/27023/workflows/f7ad2b5f-37d0-4d17-b992-e6d745bfbacc

### DIFF
--- a/tests/aws/services/transcribe/test_transcribe.py
+++ b/tests/aws/services/transcribe/test_transcribe.py
@@ -24,6 +24,7 @@ LOG = logging.getLogger(__name__)
 # Lock and event to ensure that the installation is executed before the tests
 INIT_LOCK = threading.Lock()
 installed = threading.Event()
+installation_errored = threading.Event()
 
 PRE_DOWNLOAD_LANGUAGE_CODE_MODELS = ["en-GB"]
 
@@ -40,19 +41,27 @@ def install_async():
         with INIT_LOCK:
             if installed.is_set():
                 return
-            LOG.info("installing Vosk default version")
-            vosk_package.install()
-            LOG.info("done installing Vosk default version")
-            LOG.info("installing ffmpeg default version")
-            ffmpeg_package.install()
-            LOG.info("done ffmpeg default version")
-            for language_code in PRE_DOWNLOAD_LANGUAGE_CODE_MODELS:
-                LOG.info("installing Vosk models used in test")
-                TranscribeProvider.download_model(language_code)
-                LOG.info("done installing Vosk models used in test")
-            installed.set()
+            try:
+                LOG.info("installing Vosk default version")
+                vosk_package.install()
+                LOG.info("done installing Vosk default version")
+                LOG.info("installing ffmpeg default version")
+                ffmpeg_package.install()
+                LOG.info("done ffmpeg default version")
+                LOG.info(
+                    "downloading Vosk models used in test: %s", PRE_DOWNLOAD_LANGUAGE_CODE_MODELS
+                )
+                for language_code in PRE_DOWNLOAD_LANGUAGE_CODE_MODELS:
+                    TranscribeProvider.download_model(language_code)
+                    LOG.info("done downloading Vosk model '%s'", language_code)
+                LOG.info("done downloading all Vosk models used in test")
+            except Exception:
+                LOG.exception("Error during installation of Transcribe dependencies")
+                installation_errored.set()
+            finally:
+                installed.set()
 
-    start_worker_thread(run_install)
+    start_worker_thread(run_install, name="transcribe-install-async")
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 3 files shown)


alexrashed (commented):
The skip is looking good, and the test is quite flaky. However, I will leave the approval to @sannya-singal and @ackdav since I think that this might be the only e2e test for transcribe. So I'll leave it to them if they want to skip it or fix it right away.

bentsku:
In this run, `test_transcribe_supported_media_formats[../../files/en-gb.amr-hello my name is]` flaked as well 😅

sannya-singal (commented):
I agree, fixing the tests should be worth it otherwise we need to skip all e2e tests and since there are few customers using transcribe service regularly, its good to have the these tests in place.

tests/aws/services/transcribe/test_transcribe.py
```diff
@@ -1,28 +1,120 @@
...
+                LOG.info("installing ffmpeg default version")
+                ffmpeg_package.install()
+                LOG.info("done ffmpeg default version")
+            except Exception:
+                LOG.exception("Error during installation of Vosk dependencies")
+                installation_errored.set()
+                # we also set the other event to quickly stop the polling
+                vosk_installed.set()
+            finally:
+                ffmpeg_installed.set()
```
alexrashed, line 86: The locking here shouldn't be necessary in my opinion, because the package installers already handle the locking.
(You could even have fine-grained locking control with the installers, see optional `install_lock` kwarg on the `PackageInstaller` class).

tests/aws/services/transcribe/test_transcribe.py
```diff
@@ -1,28 +1,120 @@
...
+                        language_code,
+                    )
+                LOG.info("done downloading all Vosk models used in test")
+            except Exception:
+                LOG.exception("Error during installation of Vosk dependencies")
+                installation_errored.set()
+                # we also set the other event to quickly stop the polling
+                ffmpeg_installed.set()
+            finally:
+                vosk_installed.set()
```
alexrashed, line 70: For another iteration:
This is actually an interesting use case, which could be interesting for other package installers as well.
If you have a service with multiple dependencies, we currently do not really execute them in parallel.
We could implement a native `async_install` and `await_install` functions in the `PackageInstaller` / `Package` abstraction, which would allow users of multiple package installers to easily implement a fork-join pattern to trigger them in parallel...

But again, something for another time 😛

alexrashed (approved):
The changes are looking good, but the lock handling should not be necessary in my opinion. 😉 

Thanks a lot for jumping on this, and the great team work to really get this stable! The changes are looking good and should ensure that the tests are going to be way more stable in the future! 🚀 💯

tests/aws/services/transcribe/test_transcribe.py
bentsku: Definitely, it's something I thought about as well, we could use a thread pool executor to dispatch all the `async_install` too, and wait on all the futures returned. Interesting ideas!

tests/aws/services/transcribe/test_transcribe.py
bentsku: I removed, will wait for the green pipeline and run it once again after just to be sure 👌

sannya-singal (approved):
Thanks Ben for an amazing work on pre downloading the dependencies and running them in parallel and already quickly addressing feedback from Alex on the lock handling🚀 Will work on upgrading the packages next and also optimising the installation a bit more 🙌 The changes LGTM!

Issue #11253, opened 2024-07-23
bug: API Gateway V1 (targeted to Lambda) gives 500 error with // in path


northvankiwiguy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With API Gateway V1, I have a `/orders` path that targets a Lambda function. When accessing the URL using the syntax:

`curl -v https://45szzn0od7.execute-api.localhost.localstack.cloud:4566/prod/orders` I see the correct response.
    
However, if there's an extra `/` in the path after `/prod` I see a 500 response:

`curl -v https://45szzn0od7.execute-api.localhost.localstack.cloud:4566/prod//orders`

I see the response message:

```
"__type": "InternalError", "message": "exception while calling apigateway with unknown operation: 308 Permanent Redirect: http://45szzn0od7.execute-api.localhost.localstack.cloud:4566/prod/orders"
```

I keep on hitting this bug because I have `API_URL` set to the base URL, and then call `$API_URL/orders`. If the `API_URL` contains a trailing `/`, it crashes.

I'm not sure if the Lambda integration is relevant or not (or whether other integrations will also see the problem). I can confirm that my Lambda function is _not_ invoked when `/prod//orders` is used.

### Expected Behavior

In the AWS service, both of these work correctly, regardless of whether the path is `/prod/orders` or `/prod//orders`.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-composed up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

API Gateway V1 created with CDK

```
const api = new apigateway.RestApi(this, 'example-api')
const ordersApi = api.root.addResource('orders')
const ordersLambdaInt = new apigateway.LambdaIntegration(orderLambda, { proxy: true })
ordersApi.addMethod('GET', ordersLambdaInt)
```    

I don't have the `awslocal` commands on hand, but I can figure them out if necessary.

### Environment

```markdown
- OS: MacOS Sonoma 4.5
- LocalStack: 
  LocalStack version: 3.5.1.dev
  LocalStack Docker image sha: (built from latest source)
  LocalStack build date: 2024-07-23
  LocalStack build git hash: a0a1ba090
```

northvankiwiguy:
Thanks for looking into this. Unfortunately, it's still failing, but with a different error:

```
[...]

PR #11268, opened 2024-07-25
fix APIGW path converter with prepended multiple slashes

bentsku:
## Motivation
As reported in #11253, our Edge Router had issue with `path` parameter starting with multiple slashes, as it would return a redirect which we shouldn't do. 

This PR introduces a new converter in the default `ROUTER`, which is the one we already use internally for the OP Router for AWS, and also internally for the new generation APIGW. 

This PR introduces the change and uses the new converter for APIGW routes, and fixes the behavior around it, with added integration test cases. 

## Changes
- migrate the `GreedyPathConverter` to `edge.py` as we need to use it in our default router, and refactor its usages
- add new tests for the router with the greedy path converter
- fix APIGW NG behavior with the new greedy path 
- fix the legacy APIGW, which was a bit more tricky, I'll run tests from upstream as well to be sure we're not breaking behavior

Note: failing tests are unrelated, and I'm running upstream full test suite to validate the changes which is green. Failing tests will be addressed with #11261

_fixes #11253_

### DIFF
--- a/localstack-core/localstack/aws/protocol/op_router.py
+++ b/localstack-core/localstack/aws/protocol/op_router.py
@@ -8,7 +8,6 @@ from werkzeug.exceptions import MethodNotAllowed, NotFound
 from werkzeug.routing import Map, MapAdapter
 
 from localstack.aws.protocol.routing import (
-    GreedyPathConverter,
     StrictMethodRule,
     path_param_regex,
     post_process_arg_name,
@@ -16,6 +15,7 @@ from localstack.aws.protocol.routing import (
 )
 from localstack.http import Request
 from localstack.http.request import get_raw_path
+from localstack.services.edge import GreedyPathConverter
 
 
 class _HttpOperation(NamedTuple):
--- a/localstack-core/localstack/aws/protocol/op_router.py
+++ b/localstack-core/localstack/aws/protocol/op_router.py
@@ -15,7 +15,7 @@ from localstack.aws.protocol.routing import (
 )
 from localstack.http import Request
 from localstack.http.request import get_raw_path
-from localstack.services.edge import GreedyPathConverter
+from localstack.http.router import GreedyPathConverter
 
 
 class _HttpOperation(NamedTuple):
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -702,7 +702,8 @@ def extract_path_params(path: str, extracted_path: str) -> Dict[str, str]:
 
 def extract_query_string_params(path: str) -> Tuple[str, Dict[str, str]]:
     parsed_path = urlparse.urlparse(path)
-    path = parsed_path.path
+    if not path.startswith("//"):
+        path = parsed_path.path
     parsed_query_string_params = urlparse.parse_qs(parsed_path.query)
 
     query_string_params = {}
--- a/localstack-core/localstack/services/edge.py
+++ b/localstack-core/localstack/services/edge.py
@@ -5,6 +5,8 @@ import subprocess
 import sys
 from typing import List, Optional, TypeVar
 
+from werkzeug.routing import PathConverter
+
 from localstack import config, constants
 from localstack.config import HostAndPort
 from localstack.constants import (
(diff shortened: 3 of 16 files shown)


localstack-core/localstack/aws/protocol/op_router.py
```diff
@@ -8,14 +8,14 @@
...
 from localstack.aws.protocol.routing import (
-    GreedyPathConverter,
     StrictMethodRule,
     path_param_regex,
     post_process_arg_name,
     transform_path_params_to_rule_vars,
 )
 from localstack.http import Request
 from localstack.http.request import get_raw_path
+from localstack.services.edge import GreedyPathConverter
```
thrau, line 18: i know i said "just add the greedy match router + converter to edge.py?", but really meant either copy&pasting it here, or move it to a more generic place and import it from there :-)

i don't like the `localstack.services.edge` import here.

i'd suggest moving the `GreedyPathConverter` to `localstack.http.router` for now, and perhaps also introduce it in rolo.

localstack-core/localstack/services/edge.py
```diff
@@ -23,7 +25,23 @@
...
+    regex = ".*?"
+
+    part_isolating = False
+    """From the werkzeug docs: If a custom converter can match a forward slash, /, it should have the
+    attribute part_isolating set to False. This will ensure that rules using the custom converter are
+    correctly matched."""
+
+
+ROUTER: Router[Handler] = Router(
+    dispatcher=handler_dispatcher(), converters={"greedy_path": GreedyPathConverter}
```
thrau, line 43: adding this here is :ok_hand:

thrau (commented):
Can't say much about the APIGW changes, just would like the converter to move somewhere else :-)

localstack-core/localstack/services/apigateway/helpers.py
```diff
@@ -702,7 +702,8 @@ def extract_path_params(path: str, extracted_path: str) -> Dict[str, str]:
 
 def extract_query_string_params(path: str) -> Tuple[str, Dict[str, str]]:
     parsed_path = urlparse.urlparse(path)
-    path = parsed_path.path
+    if not path.startswith("//"):
+        path = parsed_path.path
```
alexrashed, line 706: q: Could you explain why this is necessary / useful here?

alexrashed (approved):
Nice drill down, and a clean and well-tested change in API GW! 🏅 

I agree with @thrau that the `GreedyPathConverter` shouldn't move into `localstack.services.edge` but rather into `localstack.http.router`.
Besides that I only had a question on a small change in a helper (see comment).
So as soon as the question is answered, the converter is moved, and most importantly all tests are green (this is a tricky change which can break something totally unexpected), this is looking good to get merged! 🚀

localstack-core/localstack/services/apigateway/helpers.py
bentsku: Sure! This is always the same issue when using `RAW_URI`, and this helper should be removed once we remove the legacy implementation, it is doing a lot of parsing it shouldn't need to do (and sometimes it does it wrong).

If you pass a path only to `urlparse`, it can properly parse it and will return the schema/netloc etc to be `None`. However, if you pass it a path that starts with multiple slashes, it will remove the first slashes.
But if it does not start with multiple slashes, we still need to use `urlparse` because sometimes, the `RAW_URI` can be a full URL (often in combination with proxies).

Here's the comments I wrote about it in `get_raw_path`:
```python
        # werkzeug/flask request (already a string, and contains the query part)
        # we need to parse it, because the RAW_URI can contain a full URL if it is specified in the HTTP request
        raw_uri: str = request.environ.get("RAW_URI", "")
        if raw_uri.startswith("//"):
            # if the RAW_URI starts with double slashes, `urlparse` will fail to decode it as path only
            # it also means that we already only have the path, so we just need to remove the query string
            return raw_uri.split("?")[0]
        return urlparse(raw_uri or request.path).path
```

localstack-core/localstack/aws/protocol/op_router.py
bentsku: This is much nicer, thanks 👌 I didn't like it either, it felt very out of place. This looks better!

cloutierMat (approved):
Awesome! Super neat fix :rocket: 

Glad to see how simple it was to support with Next Gen also!

PR #11269, opened 2024-07-26
Add outages extension module to logger config

viren-nadkarni:
## Changes

The deprecation warning from the Outages extension was not being printed.

This PR updates the logger config setup.


alexrashed (changes requested):
I don't think that localstack can be responsible for the configuration of all potential extensions, shouldn't the extension just configure its own logging?
Maybe based on the logging config in LocalStack?

viren-nadkarni:
@alexrashed: yes, that's a perfectly fine alternative. I'll make the changes in the extension and close this PR.

PR #11276, opened 2024-07-29
fix breaking builds by avoiding setuptools==72.0.0

alexrashed:
## Motivation
With `setuptools==72.0.0`, the test command was removed after being deprecated for a long time.
This is causing quite some issues (see https://github.com/pypa/setuptools/issues/4519).
For `localstack-ext`, this hits JPype, where a PR has been created to fix the issue:
- https://github.com/jpype-project/jpype/pull/1206

For as long as this package has not been upgraded, or the `setuptools` version has been yanked, we need to find a workaround.
This PR creates and uses a pip constraints file in order to avoid `setuptools>=72.0.0`.
As soon as either JPype have been fixed or the `setuptools` version has been yanked, we can revert this commit / use the latest version of `setuptools` again.

## Changes
- Adds a `pip-constraints.txt` file declaring to avoid `setuptools>=72.0.0`.
- Sneaks in the `PIP_CONSTRAINT` environment variable to be used with `pip` everywhere.

## Testing
- Currently all pipelines are broken (at least those where the cache was invalidated today).
- If the pipelines are green with this PR, we can be quite sure that it works correctly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -151,10 +151,10 @@ RUN --mount=type=cache,target=/var/cache/apt \
 
 # upgrade python build tools
 RUN --mount=type=cache,target=/root/.cache \
-    (virtualenv .venv && . .venv/bin/activate && pip3 install --upgrade pip wheel setuptools)
+    (virtualenv .venv && . .venv/bin/activate && pip3 install --upgrade pip wheel setuptools==71.1.0)
 
 # add files necessary to install runtime dependencies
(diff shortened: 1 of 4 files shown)


alexrashed:
Closing this PR as `setuptools==74.0.0` has just been yanked: https://github.com/pypa/setuptools/issues/4519#issuecomment-2255940870

PR #11277, opened 2024-07-29
fix warnings for Dockerfiles

alexrashed:
## Motivation
While working on https://github.com/localstack/localstack/pull/11276, I stumbled over some warnings of Docker when building the images (over and over again):
```
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
```

## Changes
- Fixes deprecations in our `Dockerfile`s:
  - Have the same casing for `FROM` as for `AS`
  - Use `ENV <name>=<key>` instead of `ENV <name> <key>`


silv-io (approved):
It's nice that Docker gives some guidance for this! Helps in keeping it clean.

LGTM!

bentsku (approved):
LGTM, thanks for thinking about the S3 image too 😄

PR #11279, opened 2024-07-29
fix pro image detection in dev run script

thrau:
## Motivation

`localstack.dev.run` determined whether to use the pro or community image based on the source path the script was run from. 
with the multi-distribution, we're no longer running from `./localstack-ext`, but perhaps `./localstack-ext/localstack-pro-core` (or any other module therein).

## Changes

* the pro image is used correctly again when running `localstack.dev.run` from `localstack-ext/localstack-pro-core`

### DIFF
--- a/localstack-core/localstack/dev/run/__main__.py
+++ b/localstack-core/localstack/dev/run/__main__.py
@@ -237,6 +237,8 @@ def run(
         if pro is None:
             if os.getcwd().endswith("localstack-ext"):
                 pro = True
+            elif "localstack-pro-" in os.getcwd().split(os.pathsep)[-1]:
+                pro = True
             else:
                 pro = False
 


localstack-core/localstack/dev/run/__main__.py
```diff
@@ -237,6 +237,8 @@ def run(
         if pro is None:
             if os.getcwd().endswith("localstack-ext"):
                 pro = True
+            elif "localstack-pro-" in os.getcwd().split(os.pathsep)[-1]:
```
simonrw, line 240: Probably stupid question but is this correct for azure? I thought that was a different image so is that functionality implemented by `localstack.dev.run`?

simonrw (approved):
The change looks sensible, backwards compatible and a fairly safe bet, but are people cding into `localstack-ext/localstack-pro-core and running `l.d.r`? I always run it from the -ext checkout root. Perhaps I'm missing a use case?

thrau:
I may have shot too quickly with this PR. I had a completely unrelated issue that prevented the pro image to start up correctly when running into `localstack-ext`, so without checking i made the assumption that the script needs fixing. After your comment, I went back to check and indeed it works just fine. you're right that CDing into `localstack-pro-core` isn't really a use case, and even if, we'd probably want different startup procedures for different images.

theuniona (commented):
Yes

PR #7367, opened 2022-12-20
Add dev config flag for fixing lambda invoke on macOS in host mode

joe4dev:
Co-authored by @dominikschubert 

## Problem

The new lambda provider requires bi-directional communication (host <=> lambda container), which breaks host mode in macOS because the lambda container is not reachable via IPv4 in bridge mode.
Hence, invoking any lambda fails while the lambda container waits for an invocation (i.e., during `"blocking the credentials service"`) after a connection timeout with the error `Connection to 172.17.0.2 timed out.`

Enable new lambda provider: `PROVIDER_OVERRIDE_LAMBDA=asf`

Relevant test case: `tests.integration.awslambda.test_lambda.TestLambdaFeatures.test_invocation_type_request_response`

## Background

> Docker Desktop can’t route traffic to Linux containers. However, if you’re a Windows user, you can ping the Windows containers. (Source: Docker [I cannot ping my containers](https://docs.docker.com/desktop/networking/#i-cannot-ping-my-containers))

There is also a long discussion thread about this issue in the [moby project](https://github.com/moby/moby/issues/22753#issuecomment-222943352).

## LocalStack Workarounds

- Start LS in a container with bind-mounted code and run tests against it using `TEST_SKIP_LOCALSTACK_START=1`
    - ➖  Debugging doesn’t work
    - ➖ Tests that monkey-patch stuff don’t work
- Execute test in container `TEST_PATH="tests/integration/awslambda/test_lambda_common.py" make test-docker-mount`
    - ➖  cannot install moto using `pip install -e`
    ⇒ delete `-v $$PACKAGES_DIR/moto:/opt/code/localstack/.venv/lib/python3.10/site-packages/moto/` from test-docker-mount-code in the Makefile

Thanks, @dfangl for the hints 🙏 

## Potential macOS networking Workarounds (not tested)

- Blogpost summarizing solutions: https://itecnote.com/tecnote/r-how-to-ping-the-docker-container-from-the-host/
    - ➕ most comprehensive overview
- Docker Toolbox/VirtualBox or Vagrant/Ansible-based setup: https://stackoverflow.com/a/39217691
    - ➖ cumbersome to configure
    - ➖ doesn't support host volume mapping
- shim based fix:  https://github.com/AlmirKadric-Published/docker-tuntap-osx
    - ➖ depends on unmaintained [TunTap](https://tuntaposx.sourceforge.net/)

## Suggested Changes

[...]

PR #11280, opened 2024-07-29
Change default for LAMBDA_DEV_PORT_EXPOSE on macOS in host mode

joe4dev:
## Motivation

Lambda won't work in host mode on macOS without setting `LAMBDA_DEV_PORT_EXPOSE=1` (introduced in https://github.com/localstack/localstack/pull/7367), which is a bad developer experience.
So far, we opted for explicit configuration developer documentation [here](https://github.com/localstack/localstack/blob/master/docs/development-environment-setup/README.md#lambda) to unblock developers on macOS in host mode.
Several contributors hit this issue, and it was suggested that we set it by default if we detect LocalStack running in host mode on macOS. The Lambda service owners also discussed this in the last Lambda Service Owner Sync (2024-07-15).

## Changes

* [Unify helper utils for common operating systems](https://github.com/localstack/localstack/commit/d087e5bd64a9b728173053c10c845deadff78a8e)
* [Change default for Lambda port expose on macOS in host mode](https://github.com/localstack/localstack/commit/ae3a0c52eba2ba7ff3d4f45d0e3736eecdb2bbaf)

## Testing

Our test suite should capture any regressions. Anything else to consider? Is there a special environment where the assumption `not is_in_docker == host mode` does not apply (e.g., K8) @dfangl @simonrw?

## Discussion

Options considered:

- Option A: simply evaluate based on is_in_docker and the system platform, assuming `not is_in_docker == host mode`.
    - Mixing up config logic with production code makes config bugs more likely and leads to more complex config code
- Option B: introduce some explicit host mode LS devs should set for sensible default config
    - Could separate special host mode defaults from generic production defaults

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -380,6 +380,7 @@ OVERRIDE_IN_DOCKER = parse_boolean_env("OVERRIDE_IN_DOCKER")
 
 is_in_docker = in_docker()
 is_in_linux = is_linux()
+is_in_macos = is_macos()
 default_ip = "0.0.0.0" if is_in_docker else "127.0.0.1"
 
 # CLI specific: the configuration profile to load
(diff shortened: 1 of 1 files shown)


gregfurman:
Nice change! Tested locally running on host mode (without explicitly enabling the flag) on `master` and the `lambda-defaults-for-macos-hostmode` feature branch and the dev ports expose successfuly by default with these changes.

💻  I'm running `Darwin 23.3.0 arm64`

dfangl (approved):
is_in_docker will also be `True` if in kubernetes, it is really just a variable whether we are running in a container or not.

I think the taken approach is fine, but there are now some duplicated methods with utils/platform.py. I understand it is not that easy to use them, but it might be worth consolidating them anyway in the future.

joe4dev:
> is_in_docker will also be `True` if in kubernetes, it is really just a variable whether we are running in a container or not.
> 
> I think the taken approach is fine, but there are now some duplicated methods with utils/platform.py. I understand it is not that easy to use them, but it might be worth consolidating them anyway in the future.

Yeah, I know, and also small inconsistencies (e.g., `lower()`, `is_mac_os()`). I thought that was intentional because config should be self-contained for bootstrapping 🤷

simonrw (commented):
I think this change is sensible.

Regarding k8s behaviour: I think the same problem will be true for k8s as it is in Docker, but only when LS is being run in host mode. E.g. for `kind` the k8s "nodes" are Docker containers, therefore the same problem.

However we implicitly not really supporting that use case right now. We are focused on when LS is deployed in the k8s cluster as a pod, in which case there is no problem as all pods ports are accessible by other pods (with minor restrictions).

So in summary: I think we don't have to account for k8s behaviour specifically here.

PR #11281, opened 2024-07-29
Rework AssetDirectory to use pathlib.Path

giograno:
## Motivation
Rework of the `AssetDirectory` to use `pathlib.Path` instead of `str`.

## Changes
- use `pathlib.Path` to internally represent a path for `AssetDirectory`


thrau (approved):
LGTM! good api change :-) perhaps a unit test for the assetdirectory wouldn't hurt

Issue #11243, opened 2024-07-22
bug: AWS SES message tags with colon are not allowed (ses:feedback-id-a/b)


mnylensc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The newly introduced `ses:feedback-id-a` and `ses:feedback-id-b` message tags aren't allowed for AWS SES emulated by localstack.

See https://aws.amazon.com/about-aws/whats-new/2024/06/amazon-ses-custom-values-feedback-header/ and https://docs.aws.amazon.com/ses/latest/dg/monitor-using-event-publishing.html#event-publishing-fine-grained-feedback

Adding these against real AWS SES works.

The error observed by AWS Java SDK v2 is:

```
java.util.concurrent.CompletionException: software.amazon.awssdk.services.ses.model.SesException: Invalid tag name <ses:feedback-id-a>: only alphanumeric ASCII characters, '_', and '-' are allowed. (Service: Ses, Status Code: 400, Request ID: 66675e09-f776-45c1-a972-06db202d0ced)
```

### Expected Behavior

`ses:feedback-id-a` and `ses:feedback-id-b` should be supported as tags for `ses:SendEmail`

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

   ----

### Environment

```markdown
- OS: macOS
- LocalStack:
  LocalStack version: 3.5
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

PR #11284, opened 2024-07-30
SES: Add support for feedback campaign tags

viren-nadkarni:
## Background

AWS SES supports a special tag for sent emails that start with the prefix `ses:`. This is despite the character `:` not allowed in standard tag values.

https://docs.aws.amazon.com/ses/latest/dg/monitor-using-event-publishing.html#event-publishing-fine-grained-feedback

## Changes

This PR implements this special case in LocalStack. The snapshot tests are also updated.

## Related

Closes: https://github.com/localstack/localstack/issues/11243


sannya-singal (commented):
Thanks for tackling this 🚀 The changes LGTM!

Not super urgent: But maybe we could also consider adding the constraint for these message tags that we can only use up to two of these.

PR #11288, opened 2024-07-31
Add test for deleting an alias associated with a Lambda URL

gregfurman:
## Motivation
Records the AWS behaviour when a Lambda URL is created for an aliased version of a Lambda function; and then the alias is deleted post-creation of the URL.

The current behaviour is that the Lambda URL will remain. I believe a parity test for this edge-case is nesessary since the ARN of the function URL still points to the alias.

## Changes
- Single new test-case added called `test_lambda_url_persists_after_alias_delete` to test the above flow.

### DIFF
--- a/tests/aws/services/lambda_/test_lambda.py
+++ b/tests/aws/services/lambda_/test_lambda.py
@@ -1201,6 +1201,99 @@ class TestLambdaURL:
         assert result.status_code == 502
 
     @markers.aws.validated
+    def test_lambda_url_persists_after_alias_delete(
+        self, create_lambda_function, snapshot, aws_client
+    ):
+        snapshot.add_transformer(
+            snapshot.transform.key_value(
+                "FunctionUrl", "<function-url>", reference_replacement=False
+            )
+        )
+
+        function_name = f"test-fnurl-echo-{short_uid()}"
+        alias_name = f"alias-{short_uid()}"
+
+        exists_waiter = aws_client.lambda_.get_waiter(waiter_name="function_exists")
+        active_waiter = aws_client.lambda_.get_waiter("function_active_v2")
+
+        create_lambda_function(
+            func_name=function_name,
+            handler_file=TEST_LAMBDA_URL,
+            runtime=Runtime.nodejs20_x,
+        )
+
+        # Wait for function to exist and then create publish version and create alias
+        exists_waiter.wait(FunctionName=function_name)
+        fn_version_v1 = aws_client.lambda_.publish_version(FunctionName=function_name).get(
+            "Version"
+        )
+        aws_client.lambda_.create_alias(
+            FunctionName=function_name, FunctionVersion=fn_version_v1, Name=alias_name
+        )
+
+        # Create an aliased URL
+        exists_waiter.wait(FunctionName=function_name, Qualifier=alias_name)
+        url_config = aws_client.lambda_.create_function_url_config(
+            FunctionName=function_name,
+            AuthType="NONE",
+            InvokeMode=InvokeMode.BUFFERED,
+            Qualifier=alias_name,
+        )
+        snapshot.match("create_lambda_url_config", url_config)
+
+        aws_client.lambda_.add_permission(
+            FunctionName=function_name,
+            StatementId="urlPermission",
+            Action="lambda:InvokeFunctionUrl",
+            Principal="*",
+            FunctionUrlAuthType="NONE",
+            Qualifier=alias_name,
+        )
+
+        def _invoke_function() -> dict:
+            url = url_config["FunctionUrl"]
+            response = safe_requests.post(url, data="text", headers={"Content-Type": "text/plain"})
+            result = {"status_code": response.status_code}
+            if error_type := response.headers.get("x-amzn-ErrorType"):
+                return {
+                    **result,
+                    "headers": {
+                        "x-amzn-ErrorType": error_type,
+                    },
+                    "content": to_str(response.text),
+                }
+
+            return result
+
+        def _is_url_config_deleted():
+            try:
+                aws_client.lambda_.get_function_url_config(
+                    FunctionName=function_name, Qualifier=alias_name
+                )
+                return False
+            except aws_client.lambda_.exceptions.ResourceNotFoundException:
+                return True
+
+        # Wait for active lambda and then invoke
+        active_waiter.wait(FunctionName=function_name, Qualifier=alias_name)
+        snapshot.match("invoke_aliased_url_response", _invoke_function())
+
+        # Delete alias and then invoke aliased-URL
+        delete_alias_resp = aws_client.lambda_.delete_alias(
+            FunctionName=function_name, Name=alias_name
+        )
+        snapshot.match("delete_alias_resp", delete_alias_resp)
+
+        # TODO: The below will fail on LocalStack since URL deletions are async.
+        # When async URL deletion has been implemented, the below can be uncommented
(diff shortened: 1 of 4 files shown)


tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1146,6 +1146,75 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+        delete_alias_resp = aws_client.lambda_.delete_alias(
+            FunctionName=function_name, Name=alias_name
+        )
+        snapshot.match("delete_alias_resp", delete_alias_resp)
+
+        # Confirm the alias has deleted successfully
+        def _assert_alias_deleted():
+            assert (
+                len(
+                    aws_client.lambda_.list_aliases(
```
joe4dev, line 1201: style: I would assign the response to a variable for better readability and debuggability

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1146,6 +1146,75 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+                        FunctionName=function_name, FunctionVersion=fn_version_v1
+                    ).get("Aliases")
+                )
+                == 0
+            )
+
+        retry(_assert_alias_deleted, retries=15)
```
joe4dev, line 1208: Interesting, so `delete_alias` happens asynchronously, but there is no waiter (unlike for `create_alias` above)?

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1146,6 +1146,75 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+
+        get_url_config = aws_client.lambda_.get_function_url_config(
+            FunctionName=function_name, Qualifier=alias_name
+        )
+        snapshot.match("get_url_config", get_url_config)
+
+        assert get_url_config.get("FunctionArn") == url_config.get("FunctionArn")
+        assert get_url_config.get("FunctionUrl") == url_config.get("FunctionUrl")
```
joe4dev, line 1216: What happens if we invoke the function URL after the alias is deleted?

joe4dev (commented):
Thanks for adding coverage to that interesting edge case 👍 

Added some clarification questions, mainly what happens if we invoke the persisting URL?

tests/aws/services/lambda_/test_lambda.py
gregfurman: We get a `204` if successful -- the docs don't mention whether this is eventual or not. In truth, I wasn't sure whether this was async or not. Just wanted to be sure things actually did clear up to try minimise flakiness. Maybe I'm being pedantic here with this check.

tests/aws/services/lambda_/test_lambda.py
gregfurman: Just checked. We get a `403` forbidden. I believe the URL will be cleaned up asynchronously based on [the docs ](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html)(albeit the alias deletion is not the same as the function being deleted):

> When you delete a function that has a function URL, Lambda asynchronously deletes the function URL. If you immediately create a new function with the same name in the same account, it is possible that the original function URL will be mapped to the new function instead of deleted.

tests/aws/services/lambda_/test_lambda.py
joe4dev: I would try matching the AWS behavior in the test because async behavior is often harder to test and maintain. If a test uses retries, I assume the behavior is async (unless there's an appropriate note).

tests/aws/services/lambda_/test_lambda.py
gregfurman: Yeah 100%. Adding an invocation before and after the deletion 👍 

Perhaps it's not worth it but I'm thinking about how to test this asynchronous cleanup mechanism since the docs don't cover the above scenario. Maybe a sleep after delete, and then another test invoke?

joe4dev: Yeah, for a quick validation, I would just add a long sleep and then set a breakpoint afterward so I could check interactively what an invoke returns. Once you know the actual behavior, it's easier to write a reliable test case using a retry until the expected condition is met.

gregfurman: Just did a delayed test (waited 60 seconds after the alias was deleted) and it seems that the function URL gets cleaned up. Going to add a `retry` for this step.

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1153,6 +1153,99 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+        snapshot.match("invoke_aliased_url_response", _invoke_function())
+
+        # Delete alias and then invoke aliased-URL
+        delete_alias_resp = aws_client.lambda_.delete_alias(
+            FunctionName=function_name, Name=alias_name
+        )
+        snapshot.match("delete_alias_resp", delete_alias_resp)
+
+        # TODO: The below will fail on LocalStack since URL deletions are async.
+        # When async URL deletion has been configured, the below can be uncommented
```
joe4dev, line 1241: nit: implemented instead of configured could clarify that's an implementation parity issue (and not some configuration)

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1153,6 +1153,99 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+        # snapshot.match("invoke_deleted_alias_url_response_immediate", _invoke_function())
+
+        # Wait until URL config has deleted and then re-invoke aliased-URL
+        wait_until(_is_url_config_deleted, strategy="linear")
+        time.sleep(2)  # Wait some extra time to ensure cleanup on AWS
```
joe4dev, line 1246: Interesting, any idea why we need this in addition to the `wait_until`? Does the removal of permissions take longer?

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -1153,6 +1153,99 @@ def test_lambda_url_invocation_exception(self, create_lambda_function, snapshot,
...
+
+        # Delete alias and then invoke aliased-URL
+        delete_alias_resp = aws_client.lambda_.delete_alias(
+            FunctionName=function_name, Name=alias_name
+        )
+        snapshot.match("delete_alias_resp", delete_alias_resp)
+
+        # TODO: The below will fail on LocalStack since URL deletions are async.
+        # When async URL deletion has been configured, the below can be uncommented
+        # snapshot.match("invoke_deleted_alias_url_response_immediate", _invoke_function())
```
joe4dev, line 1242: Thanks for adding a test and comment to make this parity issue explicit 👍

joe4dev (approved):
LGTM 🚀 

Thank you for thinking about the special cases and discovering the async deletion behavior ✨

tests/aws/services/lambda_/test_lambda.py
gregfurman: Could have to do with AWS caching responses. I found that immediately calling the URL following the config being cleaned up returned `Forbidden` (as opposed to `null`).

That is, the mechanism for invoking and returning HTTP responses probably lags behind the endpoint being cleaned up.

tests/aws/services/lambda_/test_lambda.py
gregfurman: Fair! It's a bit misleading. Will change

PR #11291, opened 2024-07-31
Feature/EventStudio: patch put events

maxhoheiser:
## Motivation
To receive incoming events in their final form (after all transformers have been applied) a new proxy function is introduced in TargetSender. 
This proxy function is patched with the extension EventStduio in this PR: https://github.com/localstack/localstack-extension-event-studio/pull/26
In addition, the PR also introduces event-bus-name as a new field for the Event, required by the extension.

## Changes
- add proxy_send_event in TargetSender
- add additional field `event-bus-name` to Event


Morijarti (approved):
Awesome work on getting this done :)

joe4dev (approved):
Modifying the AWS event with `event-bus-name` is a bit hacky, so let's add a comment to clarify why this exists in the codebase because it is unrelated to AWS functionality.
-> makes it easier to consolidate in the future

PR #11294, opened 2024-08-01
upgrade ffmpeg to use version 7.0.1

sannya-singal:
## Motivation
As a follow up to PR: https://github.com/localstack/localstack/pull/11261, we decided to upgrade the `ffmpeg` package version. Currently we use an old release version `4.4.1`, for more information check out: https://www.johnvansickle.com/ffmpeg/old-releases.

## Changes
Upgrade the package to the latest release: `7.0.1`.

### DIFF
--- a/localstack-core/localstack/packages/ffmpeg.py
+++ b/localstack-core/localstack/packages/ffmpeg.py
@@ -6,19 +6,19 @@ from localstack.packages.core import ArchiveDownloadAndExtractInstaller
 from localstack.utils.platform import get_arch
 
 FFMPEG_STATIC_BIN_URL = (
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks for the upgrade! 💯 
In the future we could maybe think about an automated approach for all the installers... 🤔

sannya-singal:
> h for all the installers...

Yes we could actually run a cron job to check for all the latest stable versions of the installers and atleast list out the warnings for the deprecated ones/ the packages that use a version with say ~>1.5yr older release.

alexrashed:
Well, I think we should aim for a more generalized approach. 
A lot of installers directly reference specific versions, however, often it's easy to generalize the lookup for new versions, and maybe even build a mechanism for automated code updates.
But really just something to think about, it's definitely not necessary to build any kind of mechanisms for single packages right now.

PR #11296, opened 2024-08-01
Fixes for proper ARN generation / testing

dfangl:
## Motivation
With #10912 we introduced some fixes regarding returning proper ARNs in non-default partitions.

However, some have been missed, uncovered due to recently modified or added tests.

## Changes
* Use proper partition placeholders in tests or use the correct partition for manually generated ARNs in tests
* Events returns the correct partition in the generated ARN for root account principal ARNs on `put_permission` calls.

More fixes for events and resourcegroups are in getmoto/moto#7913

### DIFF
--- a/localstack-core/localstack/services/events/event_bus.py
+++ b/localstack-core/localstack/services/events/event_bus.py
@@ -107,7 +107,7 @@ class EventBusService:
         if condition and principal != "*":
             raise ValueError("Condition can only be set when principal is '*'")
         if principal != "*":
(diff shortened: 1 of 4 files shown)


maxhoheiser (approved):
LGTMT - nice that we now also support gov-cloud and china regions. I was also blocked by this for a P1 support ticket last.  week

I guess we have tests in place that rotate the region to also test against gov-cloud and china?

dfangl:
Yes, we have those tests, they do not yet all pass yet, but this is a big step towards it (the moto PR being another).

PR #11300, opened 2024-08-02
Fix/EventStudio: Remove event bus field before transformers

maxhoheiser:
## Motivation
This PR in EventStudio: https://github.com/localstack/localstack-extension-event-studio/pull/30 additionally collects events before transformers are applied, therefore there is no need to keep the additional field `event-bus-name` for later steps.

## Changes
- remove the additional field `event-bus-name` before input path or input transformers are applied


Morijarti (approved):
LGTM

PR #11301, opened 2024-08-02
Lambda DevX, Introducing Lambda Debug Mode

MEPalma:
## Motivation
The debugging of Lambda functions is currently hampered by the need to manually manage or disable execution timeouts, which often do not provide enough time to connect a remote debugger and perform the necessary evaluations. This issue is particularly relevant for services like Lambda and API Gateway. By default, Lambda limits execution to three seconds, though it can be configured for up to 15 minutes, while API Gateway can impose its own timeout on integrations.
To enhance the developer experience when debugging Lambda functions, this pull request (PR) proposes introducing a "Debug Mode." This mode would signal to the system that the deployment is in a debugging phase, allowing any provider to adjust relevant settings, such as timeouts, accordingly. The changes also suggest setting a default timeout value and implementing dynamic timeout adjustments that do not alter the system's static state. This ensures compatibility with persistence logic and maintains readability of timeout values during debugging, improving transparency and reducing potential distractions.
This initial effort will be followed by a more substantial configurable pattern designed to enable debugging across multiple Lambda functions simultaneously when in Debug Mode.
The new recommended command for starting LocalStack for the debug of lambda functions is now: 
`LAMBDA_DEBUG_MODE=1 LAMBDA_DOCKER_FLAGS='-p 19891:19891' localstack start`

_Note that API Gateway was found to not strictly enforce the evaluation of Lambda functions within its set integration timeouts. Instead, it ensures that the process is halted after the maximum Lambda execution time is reached. This behaviour has been maintained, with an extended timeout applied when in Debug Mode. However, enforcing timeouts in API Gateway should be considered in future efforts._

## Changes
- addition of the `LAMBDA_DEBUG_MODE` environment variable and related support constants and functions
- addition of automatic lambda evaluation timeout increase when in debug mode
- addition of automatic API Gateway timeout increase when in debug mode

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -429,9 +429,9 @@ if TMP_FOLDER.startswith("/var/folders/") and os.path.exists("/private%s" % TMP_
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
-# PUBLIC: 0 (default)
+# PUBLIC PREVIEW: 0 (default), 1 (preview)
 # When enabled it triggers specialised workflows for the debugging.
-DEBUG_MODE = is_env_true("DEBUG_MODE")
+LAMBDA_DEBUG_MODE = is_env_true("LAMBDA_DEBUG_MODE")
 
 # whether to enable debugpy
 DEVELOP = is_env_true("DEVELOP")
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -429,6 +429,10 @@ if TMP_FOLDER.startswith("/var/folders/") and os.path.exists("/private%s" % TMP_
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
+# PUBLIC: 0 (default)
+# When enabled it triggers specialised workflows for the debugging.
+DEBUG_MODE = is_env_true("DEBUG_MODE")
+
 # whether to enable debugpy
 DEVELOP = is_env_true("DEVELOP")
 
--- a/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
+++ b/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
@@ -23,6 +23,7 @@ from localstack.services.lambda_.invocation.runtime_executor import (
     RuntimeExecutor,
     get_runtime_executor,
 )
+from localstack.utils.debug_mode import DEFAULT_DEBUG_MODE_TIMEOUT_SECONDS, is_debug_mode
 from localstack.utils.strings import to_str
 
 STARTUP_TIMEOUT_SEC = config.LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT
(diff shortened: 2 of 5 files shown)


localstack-core/localstack/config.py
```diff
@@ -429,6 +429,10 @@ def in_docker():
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
+# PUBLIC: 0 (default)
+# When enabled it triggers specialised workflows for the debugging.
+DEBUG_MODE = is_env_true("DEBUG_MODE")
+
```
thrau, line 435: this may be potentially confusing to people, since we already have `DEBUG=1`. if this is specific to lambda, i would suggest `LAMBDA_DEBUG_MODE`, or if it's supposed to be a generic config, then we should discuss what exactly this config variable does and avoid potential overlap with `DEBUG=`

MEPalma: Thank you! As this targets debugging of lambda functions, I agree this is best if renamed.

localstack-core/localstack/config.py
```diff
@@ -429,6 +429,10 @@ def in_docker():
 LS_LOG = eval_log_type("LS_LOG")
 DEBUG = is_env_true("DEBUG") or LS_LOG in TRACE_LOG_LEVELS
 
+# PUBLIC: 0 (default)
```
joe4dev, line 432: Let's introduce this feature as public preview until we have more feedback about how it's working and the naming:

```
# PUBLIC PREVIEW: 0 (default), 1 (preview)
```

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -390,3 +391,11 @@ def _generate_trace_header(self):
             "utf-8"
         )  # TODO: segment doesn't actually exist at the moment
         return f"Root={root_trace_id};Parent={parent};Sampled={sampled}"
+
+    def _get_execution_timeout_seconds(self) -> int:
```
joe4dev, line 395: 👍

localstack-core/localstack/utils/debug_mode.py
```diff
@@ -0,0 +1,13 @@
+from typing import Final
+
+from localstack.config import DEBUG_MODE
+
+# Specifies the fault timeout value in seconds to be used by time restricted workflows when
+# Debug Mode is enabled. The value is set to one hour to ensure eventual termination of
+# long-running processes.
+DEFAULT_DEBUG_MODE_TIMEOUT_SECONDS: Final[int] = 3_600
```
joe4dev, line 8: style FYI: we're not consistently using `Final` in Lambda, but it cannot hurt

localstack-core/localstack/config.py
joe4dev: +1
Everyone associates debug mode with `DEBUG=1` (common terminology in support and customer interaction) and we should avoid naming overload.

I'm ok with `LAMBDA_DEBUG_MODE` as preview for now given it's only used in Lambda, but it makes sense to think about the naming here because we envision a LocalStack-wide debug mode that re-configures other services to work smoothly with interactive debugging (e.g., disabling timeouts on upstream services invoking a debug-enabled function).

Brainstorming some alternatives:
* `REMOTE_DEBUGGING` matching the [LS docs](https://docs.localstack.cloud/user-guide/lambda-tools/debugging/)
* `DEBUG_SETTINGS`
* `DEBUG_CONFIG`
* `DEBUGGING_MODE`
* `INTERACTIVE_DEBUG_MODE`
* `INTERACTIVE_DEBUGGING`
* `DEBUG_TIMEOUT_MODE` if it only affects timeouts

What do you think @dfangl @dominikschubert ?

joe4dev (approved):
I'm ok moving forward after re-naming the config flag and marking it as preview for now.

Let's continue the discussion about the naming, which might change once we have more clarify what it does at the end.

PR #11304, opened 2024-08-03
fix APIGW double slashes between stage and path

bentsku:
## Motivation
Follow up from #11268, the first fix only fixed one part of the issue where for greedy matching, but it did not address the original issue which was multiple slashes between the stage and the path. 

```
2024-08-02T11:22:49.052 DEBUG --- [et.reactor-5] l.s.apigateway.helpers     : No match found for path: '//orders' and method: 'GET'
2024-08-02T11:22:49.052  INFO --- [et.reactor-5] localstack.request.http    : GET /prod//orders => 404
```

## Changes
- add a new case in the existing test for hardcoded path
- rewrite a bit the test to have better test value, following a good comment from @cloutierMat that everything is "test" "test" and that is was a bit unclear
- fix the value for `pathParameters` in APIGW NG (if there are no values, it's `None` for AWS_PROXY)
- fix how we remove the leading slashes for a path in order to properly route a request, but keep them for the `requestContext.path` value which is stage + path

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -1503,11 +1503,11 @@ def get_event_request_context(invocation_context: ApiInvocationContext):
     resource_id = invocation_context.resource_id
 
     set_api_id_stage_invocation_path(invocation_context)
-    relative_path, query_string_params = extract_query_string_params(
-        path=invocation_context.path_with_query_string
-    )
     api_id = invocation_context.api_id
     stage = invocation_context.stage
+
+    full_path = invocation_context.raw_uri.removeprefix(f"/{stage}")
+    relative_path, query_string_params = extract_query_string_params(path=full_path)
 
     source_ip = invocation_context.auth_identity.get("sourceIp")
     integration_uri = integration_uri or ""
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/apigateway/helpers.py
```diff
@@ -1503,12 +1503,12 @@ def get_event_request_context(invocation_context: ApiInvocationContext):
...
     set_api_id_stage_invocation_path(invocation_context)
-    relative_path, query_string_params = extract_query_string_params(
-        path=invocation_context.path_with_query_string
-    )
     api_id = invocation_context.api_id
     stage = invocation_context.stage
 
+    full_path = invocation_context.raw_uri.removeprefix(f"/{stage}")
+    relative_path, query_string_params = extract_query_string_params(path=full_path)
+
```
cloutierMat, line 1511: This is probably where the failing test comes from. As the `raw_uri` contains `_user_request_`. That seems like a pretty bad naming for `path_with_query_string`, to also be responsible for removing the `_user_request_`... But we will be moving away from it soon! :rofl:

cloutierMat (commented):
Thanks for the good step forward in removing `/test/test/test` :pray: 

Overall it looks good to me. Good find to fix in the context.

Reguarding the test failure, we would lose a bit of parity in the `requestContext.path` to use `path_with_query_string`, but this will be fixed with next_gen.

Or we would need to use match with `PATH_REGEX_USER_REQUEST`? But that seems a bit dirty as well :face_with_diagonal_mouth:

cloutierMat (approved):
Great fix for the failing test :rocket:

PR #11309, opened 2024-08-05
Simplify publishing analytics of new LocalStack editions

silv-io:
## Motivation

With adding a new edition of LocalStack we need to add its marker to the analytics events

## Changes

- Generalizes the parsing of markers to give out `xyz` for any `.xyz-version` file in the static libs (`/usr/lib/localstack`)

### DIFF
--- a/localstack-core/localstack/utils/analytics/metadata.py
+++ b/localstack-core/localstack/utils/analytics/metadata.py
@@ -107,16 +107,18 @@ def get_machine_id() -> str:
 
 
 def get_localstack_edition() -> Optional[str]:
-    version_dir = "/usr/lib/localstack"
-
     # Generator expression to find the first hidden file ending with '-version'
     version_file = next(
-        (f for f in os.listdir(version_dir) if f.startswith(".") and f.endswith("-version")),
+        (
+            f
+            for f in os.listdir(config.dirs.static_libs)
+            if f.startswith(".") and f.endswith("-version")
+        ),
         None,
     )
 
     # Return the base name of the version file, or None if no file is found
-    return version_file.removesuffix("-version") if version_file else None
+    return version_file.removesuffix("-version").removeprefix(".") if version_file else None
 
 
 def is_license_activated() -> bool:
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -106,13 +106,15 @@ def get_machine_id() -> str:
...
-def get_localstack_edition() -> Optional[Literal["enterprise", "pro", "community"]]:
+def get_localstack_edition() -> Optional[Literal["enterprise", "pro", "community", "azure-alpha"]]:
     if os.path.exists("/usr/lib/localstack/.enterprise-version"):
         return "enterprise"
     elif os.path.exists("/usr/lib/localstack/.pro-version"):
         return "pro"
     elif os.path.exists("/usr/lib/localstack/.community-version"):
         return "community"
+    elif os.path.exists("/usr/lib/localstack/.azure-alpha-version"):
+        return "azure-alpha"
```
thrau, line 117: could we maybe generalize this to simply check for `/usr/lib/localstack/.<x>-version` and return` `<x>`?

thrau (commented):
thanks for adding! maybe we could generalize this?

localstack-core/localstack/utils/analytics/metadata.py
silv-io: Added this and a unit test for testing it.

Using `config.dirs.static_libs` (which should be `/usr/lib/localstack` in the container) now so that it can be tested more easily.

localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -106,15 +106,19 @@ def get_machine_id() -> str:
...
+def get_localstack_edition() -> Optional[str]:
+    # Generator expression to find the first hidden file ending with '-version'
+    version_file = next(
+        (
+            f
+            for f in os.listdir(config.dirs.static_libs)
+            if f.startswith(".") and f.endswith("-version")
+        ),
+        None,
+    )
```
thrau, line 118: i personally prefer the less clever way, but personal taste i guess :-)

```python
    version_file = None
    for f in os.listdir(config.dirs.static_libs):
        if f.startswith(".") and f.endswith("-version"):
            version_file = f
            break
```

thrau (approved):
nice thanks! :+1:  :shipit:

PR #11310, opened 2024-08-05
Automatically convert props in resource provider botocore calls

dominikschubert:
## Motivation

Follow-up to https://github.com/localstack/localstack/pull/11178

Resource properties are now automatically converted for any boto calls issues through the clients in resource providers are now automatically corrected according to the operation model.

## Changes

- The internal client factory used when creating resource providers now automatically registers a botocore hook that  performs parameter conversion according to the operation model on all requests.

## Todo

This PR still keeps the old utils around, so in most cases this is purely a fall back, which at least shouldn't break any existing functionality. Follow-up PRs will handle the refactoring of the util usage & cleaning up the utils.

## Testing

I've also already tested refactoring the covered utils into NO-OPs. Both Community and -ext need minor adjustments to some providers for that to work which I'll do in follow-up PRs. Seems to be working quite nicely though. :+1:

### DIFF
--- a/localstack-core/localstack/services/cloudformation/resource_provider.py
+++ b/localstack-core/localstack/services/cloudformation/resource_provider.py
@@ -110,7 +110,7 @@ ResourceProperties = TypeVar("ResourceProperties")
 
 def _handler_provide_client_params(event_name: str, params: dict, model: OperationModel, **kwargs):
     """
-    A botocore hook handler that will convert the passed parameters according to the corresponding operation model
+    A botocore hook handler that will try to convert the passed parameters according to the given operation model
     """
     return convert_request_kwargs(params, model.input_shape)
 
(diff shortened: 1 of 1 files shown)


pinzon (approved):
Great!!

simonrw (approved):
I love the elegance of this solution, but it is magical behaviour, and could be confusing. E.g. "why when I pass `key` does the boto call still work even when it requires `Key`?"...

I wonder if we can document this behaviour somewhere clearly, or (better): require opt in to its functionality. 

I know the user can go to the type definition of the client type, but they may go to the definition of the method instead which would not help.

Then again if manual conversion is already done then this functionality is a no op, and simple resources which can pass the input model become very straightforward, so probably win win so👍

Morijarti (approved):
I love the elegance of new solution.

PR #11316, opened 2024-08-06
fix Lambda custom aliased URL flaky test

gregfurman:
## Motivation
- Test case introduced in #11272 is flaky since `RevisionID` values are UUIDs and therefore not suited for comparison.

## Changes
- Compare the [`LastModified` (in ISO-8601 format) response values](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateFunctionCode.html#lambda-UpdateFunctionCode-response-LastModified) to assert that the second `UpdateFunctionCode` request occurs after the first.


joe4dev (approved):
Thanks for the quick flake fix 👍 

Timestamp format should allow for `>` comparison:
```
"LastModified": "2024-07-31T07:06:16.058+0000",
```

PR #11317, opened 2024-08-06
add APIGW NG workaround for AccessDenied exception

bentsku:
## Motivation
While testing against the real AWS, it seems the `AccessDenied` exception is special, and the `message` field in the default JSON response is capitalized, even though the default template for the `ACCESS_DENIED` GatewayResponse is the same as all others: `{"message":$context.error.messageString}`, so the CRUD layer is once again wrong for the default values. 

## Changes
- add a special case until we handle the `responseTemplate` where we will override the default value for `AccessDenied `
- add a unit test to prevent regression


cloutierMat (approved):
Nice catch! but What a nightmare :scream: 

Thanks for adding a test for this specific case

PR #11322, opened 2024-08-06
Upgrade pinned Python dependencies

pinzon:
## Motivation
Manual upgrade of pinned libs to force the inclusion of the newer rolo version

## Changes
- pinned libs upgrade

### DIFF
--- a/requirements-base-runtime.txt
+++ b/requirements-base-runtime.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --extra=base-runtime --output-file=requirements-base-runtime.txt --strip-extras --unsafe-package=distribute --unsafe-package=localstack-core --unsafe-package=pip --unsafe-package=setuptools pyproject.toml
 #
-attrs==24.1.0
+attrs==24.2.0
     # via localstack-twisted
 awscrt==0.21.2
     # via localstack-core (pyproject.toml)
@@ -23,7 +23,7 @@ cbor2==5.6.4
     # via localstack-core (pyproject.toml)
 certifi==2024.7.4
     # via requests
-cffi==1.16.0
+cffi==1.17.0
     # via cryptography
 charset-normalizer==3.3.2
     # via requests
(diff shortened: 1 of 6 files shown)


alexrashed (approved):
Thanks @pinzon to get the latest `rolo` upgrade in! The changes are lookin' good!
Just a few remarks to avoid any pitfalls and to make your life easier next time:
- Please be aware that dependencies are pinned for every single package. This means upgrading the dependencies here in the community project is not sufficient if you want to use this library in a downstream project (like our LocalStack Pro code). You have to explicitly upgrade the dependency pins there as well!
- You don't have to create these upgrade PRs manually, as a maintainer you can just [manually trigger our upgrade action](https://github.com/localstack/localstack/actions/workflows/upgrade-python-dependencies.yml) (which otherwise is automatically executed every Tuesday morning).
- If I can see that correctly, you are aiming at using the new features of the `rolo` upgrade in your PR. You don't have to upgrade the dependencies in an external PR, you can also do that directly in your PR. And you can just upgrade a single dependency:
  - Here you basically take a look in which extra this dependency is used, and then run the following command for the extra and all extras depending on it.
  - `rolo` is a `basic` dependency, which means you would have to run the following commands for all extras (since all of them depend on `basic`):
     ```
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra basic -o requirements-basic.txt pyproject.toml
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra base-runtime -o requirements-base-runtime.txt pyproject.toml
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra runtime -o requirementsruntime.txt pyproject.toml
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra test -o requirements-test.txt pyproject.toml
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra dev -o requirements-dev.txt pyproject.toml
      pip-compile --strip-extras --upgrade-package "rolo==0.7.1" --extra typehint -o requirements-typehint.txt pyproject.toml
     ```
    - I think it's favorable to upgrade a specific dependency in your PR directly (to link the upgrade to the feature where you want to use it). And if the dependency is upgraded to the specific version in the `master` branch in the meantime, this commit will just disappear in your PR.

Let me know if you have any questions, or if I can help anywhere. This is definitely something where we can improve the developer experience further, happy for any input!

PR #11327, opened 2024-08-07
improve X-Amz-Trace-Id to APIGW NG

bentsku:
## Motivation
As talked about in https://github.com/localstack/localstack/issues/11253#issuecomment-2271700499, we could use the same trace-id throughout the invocation, this was left as a TODO. I have a bit of time, so I implemented it quickly. 

It seems the Lambda integration is overwriting the trace header for now, so it is hard to test. It might be done with a regular AWS integration, to SQS for example. We can work on this as a follow-up. 
The implementation is very naive for now, as it does not take into account possible AWS callers already passing a trace-id header? My knowledge of X-Ray is not great, so we can build on that! 

## Changes
- implement a trace-id field to the context so that it is accessible throughout the invocation
- add a util to generate a trace and some unit tests around it
- adding the newly generated trace-id field where we would generate a random id before

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
@@ -15,6 +15,7 @@ from localstack.utils.time import timestamp
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
 from ..context import InvocationRequest, RestApiInvocationContext
 from ..header_utils import should_drop_header_from_invocation
+from ..helpers import generate_trace_id
 from ..moto_helpers import get_stage_variables
 from ..variables import ContextVariables, ContextVarsIdentity
 
@@ -43,6 +44,9 @@ class InvocationRequestParser(RestApiGatewayHandler):
         # then populate the stage variables
         context.stage_variables = self.fetch_stage_variables(context)
         LOG.debug("Initializing $stageVariables='%s'", context.stage_variables)
+
+        # TODO: improve the logic here, maybe there's already a X-Amz-Trace-Id header we should take from?
+        context.trace_id = f"Root={generate_trace_id()}"
 
     def create_invocation_request(self, context: RestApiInvocationContext) -> InvocationRequest:
         request = context.request
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
@@ -1,6 +1,8 @@
 import copy
 import logging
 import re
+import time
+from secrets import token_hex
 from typing import Type, TypedDict
 
 from moto.apigateway.models import RestAPI as MotoRestAPI
@@ -113,3 +115,12 @@ def validate_sub_dict_of_typed_dict(typed_dict: Type[TypedDict], obj: dict) -> b
     typed_dict_keys = {*typed_dict.__required_keys__, *typed_dict.__optional_keys__}
 
     return not bool(set(obj) - typed_dict_keys)
+
+
+def generate_trace_id():
+    """https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids"""
+    original_request_epoch = int(time.time())
+    timestamp_hex = hex(original_request_epoch)[2:]
+    version_number = "1"
+    unique_id = token_hex(12)
+    return f"{version_number}-{timestamp_hex}-{unique_id}"
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -88,6 +88,8 @@ class RestApiInvocationContext(RequestContext):
     """The region the REST API is living in."""
     account_id: Optional[str]
     """The account the REST API is living in."""
+    trace_id: Optional[str]
+    """The X-Ray trace ID for the request."""
     resource: Optional[Resource]
     """The resource the invocation matched"""
     resource_method: Optional[Method]
@@ -126,3 +128,4 @@ class RestApiInvocationContext(RequestContext):
         self.integration_request = None
         self.endpoint_response = None
         self.invocation_response = None
+        self.trace_id = None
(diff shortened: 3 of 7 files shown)


northvankiwiguy:
Thanks for implementing this so quickly! For my use case, it would be super helpful to pass the caller's X-Amzn-Trace-Id, if one is provided. This allows me to have part of my application running in LocalStack, while keeping the trace ID consistent with the rest of my application that has already generated the ID.

bentsku:
@northvankiwiguy thanks a lot, I'll update the PR with a test and this functionality. Thanks for the quick feedback!

localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
```diff
@@ -113,3 +115,27 @@ def validate_sub_dict_of_typed_dict(typed_dict: Type[TypedDict], obj: dict) -> b
...
+
+def parse_trace_id(trace_id: str) -> dict[str, str]:
+    split_trace = trace_id.split(";")
+    trace_values = {}
+    for trace_part in split_trace:
+        key_value = trace_part.split("=")
+        if len(key_value) == 2:
+            trace_values[key_value[0].capitalize()] = key_value[1]
+
+    return trace_values
```
cloutierMat, line 141: Question: Do you think we should put those method in a shared utils for more services to have access to? I feel like they will mostly be the same for everywhere?

cloutierMat (approved):
Thanks for tackling this!

Nice and clean implementation! Another proof of the strength of the NG provider!

I just left a question regarding the location of the helper methods!

localstack-core/localstack/services/apigateway/next_gen/execute_api/helpers.py
bentsku: That's a good point! Not sure where it should live for now 🤔 we might keep it here for now, and if we hear about more x-ray traces, move it somewhere else?

I believe lambda might have a similar implementation..

cloutierMat: I think Lambda is only using `long_uid()`, unless I missed it. 

I was thinking that it could make sense in `localstack-core/utils` maybe create a `trace.py` but not super convinced either.

I think it is fine with it here for now, but we should stay alert to prevent this behavior to be implemented many times in the code base!

northvankiwiguy:
Thanks for the bug fix! I'm glad you handle both Active and Passive cases.

However... right now I'm not able to see the HTTP header being passed to Lambda. I know that Lambda incorrectly overwrites the value, but I added `LOG.debug("Headers in Lambda: %s", context.request.headers)` in the Lambda service `provider.py`, and I get the following output (which doesn't include `X-Amzn-Trace-Id`).

There's a chance I'm looking in the wrong place, and I'll keep debugging it, but I thought I'd mention it in case you immediately knew the problem.

```
Accept-Encoding: identity
X-Amz-Invocation-Type: RequestResponse
User-Agent: Boto3/1.34.149 md/Botocore#1.34.149 md/awscrt#0.21.2 ua/2.0 os/macos#23.5.0 md/arch#x86_64 lang/python#3.12.4 md/pyimpl#CPython cfg/retry-mode#legacy Botocore/1.34.149
x-localstack-data: {"source_arn":"arn:aws:execute-api:us-west-2:000000000000:kvlt04h4ft/prod/GET/orders","service_principal":"apigateway"}
host: localhost:4566
X-Amz-Date: 20240808T142740Z
Authorization: AWS4-HMAC-SHA256 Credential=__internal_call__/20240808/us-west-2/lambda/aws4_request, SignedHeaders=host;x-amz-date;x-amz-invocation-type;x-localstack-data, Signature=136d245635dbc9dae16f42734841dae0bcefe4c0377c640123e4eaccd2829df5
amz-sdk-invocation-id: 4fccd077-753a-4c10-a106-af1cebb4d06f
amz-sdk-request: attempt=1
Content-Length: 1680
x-moto-account-id: 000000000000
```

I have your latest commit in my workspace: `improve X-Amz-Trace-Id for APIGW NG (#11327)`

bentsku:
Yes, it is currently not working for `AWS_PROXY` integration type. We will properly pass it in the `event` object passed to Lambda, but I'm not sure we have a proper way to pass it through our custom client for now (the `connect_to` client we use for internal requests between services). I'll see with the owner of this particular piece of code, but I'm not sure it will be available soon, sorry 😕 I'll come back to you with more info!

northvankiwiguy:
Thanks, I forgot I was using Proxy Integration. I double-checked and the header value is correctly being passed into the `event` object, and my application (using Fastify) is correctly seeing the header (in `request.headers['x-amzn-trace-id']`).

However, the Lambda functions `_X_AMZN_TRACE_ID` is being set incorrectly, but I realize this is because the Lambda service is not correctly respecting the existing header, and it goes ahead and generates a new Trace ID. Is this the problem you mentioned you were looking into, or should I open a separate issue for Lambda not supporting Active Tracing?

Again... thanks for adding the API Gateway feature so quickly!

bentsku:
It is somewhat similar, yes! There are 2 issues:
- our call to Lambda in the case of `AWS_PROXY` is done via an internal boto3 call, where we don't pass the `x-amzn-trace-id` header like you showed with your log statement, so even if Lambda was not overriding the trace-id, it would not get it
- but I did not focus too much on the point above during implementation because Lambda does not set `_X_AMZN_TRACE_ID` as you said, and it overrides anything that is passed. 

The point I was talking about above was the point 1, but we need to address point 2 first. It is still good to think about how to address passing the trace id along!

northvankiwiguy:
Thank you, that makes sense. For Point 2 (Lambda overriding the passed-in trace ID), will I need to open a new issue to get that fixed?

bentsku:
Yes, I think that would be good!

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
```diff
@@ -166,3 +169,15 @@ def fetch_stage_variables(context: RestApiInvocationContext) -> Optional[dict[st
...
+    @staticmethod
+    def populate_trace_id(headers: Headers) -> str:
+        incoming_trace = parse_trace_id(headers.get("x-amzn-trace-id", ""))
+        # parse_trace_id always return capitalized keys
+
+        trace = incoming_trace.get("Root", generate_trace_id())
+        incoming_parent = incoming_trace.get("Parent")
+        parent = incoming_parent or generate_trace_parent()
+        sampled = incoming_trace.get("Sampled", "1" if incoming_parent else "0")
+        # TODO: lineage? not sure what it related to
```
joe4dev, line 182: Lineage is related to detecting recursive Lambda loops (July 2023): https://aws.amazon.com/blogs/compute/detecting-and-stopping-recursive-loops-in-aws-lambda-functions/

They extended support for S3, SQS, and SNS in Aug 2024: https://aws.amazon.com/blogs/compute/aws-lambda-introduces-recursive-loop-detection-apis/

PR #11328, opened 2024-08-07
Provide regions to various inter-service communication clients

dfangl:
## Motivation
Until recently, we did not provide a proper region to various clients around LocalStack.
This worked somewhat well, especially with global services like IAM/STS.

However, with the recent advancement in partition support, the proper region has to be provided for those clients as well.

## Changes
* Provide correct regions for various internal clients

### DIFF
--- a/localstack-core/localstack/services/apigateway/provider.py
+++ b/localstack-core/localstack/services/apigateway/provider.py
@@ -428,7 +428,10 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
 
         # find matching hosted zone
         zone_id = None
-        route53 = connect_to().route53
+        # TODO check if this call is IAM enforced
+        route53 = connect_to(
+            region_name=context.region, aws_access_key_id=context.account_id
+        ).route53
         hosted_zones = route53.list_hosted_zones().get("HostedZones", [])
         hosted_zones = [hz for hz in hosted_zones if domain_name.endswith(hz["Name"].strip("."))]
         zone_id = hosted_zones[0]["Id"].replace("/hostedzone/", "") if hosted_zones else zone_id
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/apigateway/provider.py
```diff
@@ -428,7 +428,10 @@ def create_domain_name(
 
         # find matching hosted zone
         zone_id = None
-        route53 = connect_to().route53
+        # TODO check if this call is IAM enforced
+        route53 = connect_to(
+            region_name=context.region, aws_access_key_id=context.account_id
+        ).route53
```
bentsku, line 434: Yeah, this is pretty interesting. I'd say it might not get enforced, as I'm not sure we'd have any way to pass credentials here. 

I'm not too aware of the details of the domain name creation and logic, but from the documentation, it seems we're retrieving the `regionalHostedZoneId` but this is the doc:
> [distributionHostedZoneId](https://docs.aws.amazon.com/apigateway/latest/api/API_CreateDomainName.html#API_CreateDomainName_ResponseSyntax)
The region-agnostic Amazon Route 53 Hosted Zone ID of the edge-optimized endpoint. The valid value is Z2FDTNDATAQYW2 for all the regions. For more information, see Set up a Regional Custom Domain Name and AWS Regions and Endpoints for API Gateway.

See [the doc page](https://docs.aws.amazon.com/apigateway/latest/api/API_CreateDomainName.html).
So it is a bit weird 😄

bentsku (approved):
LGTM! Thanks for the fixes 😄

localstack-core/localstack/services/apigateway/provider.py
dfangl: It might be global (not sure if that lookup works though), but it is definitely only partition-global, so hitting the right partition should definitely be our goal here - might as well use the right region then :)

bentsku: Oh sorry, I meant this comment about the `# TODO check if this call is IAM enforced`, the change is definitely good! 😅

PR #11330, opened 2024-08-07
Cfn parameters support list

### LINKED ISSUE #11015: Unsupported stack parameter type: AWS::SSM::Parameter::Value<CommaDelimitedList>

### Feature description

I provide this parameter in my CFN:

```
  LambdaSubnetIds:
    Type: 'AWS::SSM::Parameter::Value<CommaDelimitedList>'
    Description: 'Subnet IDs for Lambda functions.'
```
    
 And upon creation of the stack i get:

```
 Unsupported stack parameter type: AWS::SSM::Parameter::Value<CommaDelimitedList>
```
### 🧑‍💻 Implementation

_No response_

### Anything else?

Please add this functionality so i can just reuse my CFN i have in Aws, but for localstack locally

cloutierMat:
## Motivation

fixes #11015

As mentioned on issue #11015 CLoudFormation lacked the support for `CommaDelimitedList` ssm parameters.

This allows to use SSM parameters to pass in list parameters to a cloudformation template.

## Changes

- As we already had support for `CommaDelimitedList` for Cfn parameters, the support was extended for ssm parameters
- A test was added to prevent regression


tests/aws/services/cloudformation/api/test_changesets.py
pinzon, line 1: I believe that this test could be a lot simpler and be located in another place like `tests/aws/services/cloudformation/test_template_engine.py` but it's just a nit

pinzon (approved):
LGTM, just one comment for your consideration

tests/aws/services/cloudformation/api/test_changesets.py
cloutierMat: Can do, I had placed it here to be with `test_create_change_set_with_ssm_parameter` but I agree that it could also be moved to `TestSsmParameters`. I kept a lot of the assertion from the original test. It is kinda hiding what I validating for. I will see to simplify it :wink:

cloutierMat: Thanks for the comment @pinzon 

I have moved the test to `tests/aws/services/cloudformation/test_template_engine.py:TestSsmParameters`, and simplified it using the `deploy_cfn_template` fixture. You are right that not much value was added with the extra validations that were already performed by other tests :+1:

PR #11331, opened 2024-08-08
Fix Lambda ESM lifecycle and stream poller exceptions

joe4dev:
## Motivation

The initial Lambda ESM v2 implementation implements only basic lifecycle support and doesn't handle exceptions in the streaming poller (both could happen in the test `test_disable_kinesis_event_source_mapping`).

## Changes

* Add missing CLI config flag for `LAMBDA_EVENT_SOURCE_MAPPING`
* Implement async state update upon `esm_worker` shutdown
* Introduce two error types `PipeInternalError` and `CustomerInvocationError` matching the AWS behavior
* Extract method `poll_events_from_shard` from `poll_events` in stream poller
* Extract method `get_records` from `poll_events` in stream poller and generalize
* Remove skip for test `test_disable_kinesis_event_source_mapping`
* Extract `_get_function_version_from_arn` and `_get_function_version` from Lambda provider into `provider_utils.py` to avoid cyclic dependency.

## Testing

Pipelines:

* `ext`: Full ext CI run (prevent Pipes regression)
* `multi`: Multi-accout/region pipeline

Builds:

* [13d7e32](https://github.com/localstack/localstack/pull/11331/commits/13d7e3257af0e6b0b2810b6221067c8e36af64e3)
  * ✅  [ext](https://github.com/localstack/localstack-ext/actions/runs/10366026786)
  * multi (trigger [here](https://app.circleci.com/pipelines/github/localstack/localstack?branch=fix-esm-lifecycle-and-stream-poller-exceptions) once previous run is done to avoid cancelling it) -> was green in follow-up PR [here](https://github.com/localstack/localstack/pull/11337).

## Limitations

* Edge cases (e.g., update during update, failure cases) and status reasons (especially for non-user-initiated reasons) for lifecycle updates still not implemented and tested
 * The control flow of `poll_events_from_shard` is still not ideal with the many returns and should be refactored further. Waiting for better test coverage and clarify around filtering conversions (only for filter or transform event for JSON?) and retry counter behavior (internal or re-fetch?)

## Follow Ups

* We discussed making this update synchronous from the Lambda provider (in a follow up to get this one out quickly).


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -81,101 +104,129 @@ def poll_events(self):
...
-                    abort_condition = "RecordAgeExpired"
-            matching_events = self.filter_events(polled_events)
-            # Don't trigger upon empty events
-            if len(matching_events) == 0:
-                # Update shard iterator upon if no records match the filter
-                self.shards[shard_id] = get_records_response["NextShardIterator"]
-                return
-            events = self.add_source_metadata(matching_events)
+            return get_records_response
+        except self.expired_iterator_exception_type() as e:
```
dfangl, line 204: The expired iterator exception types are all imported from our API, but this will be a boto client error. I would think that does not work, did you test this?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -81,101 +104,129 @@ def poll_events(self):
...
-    @abstractmethod
-    def initialize_shards(self) -> dict[str, str]:
-        """Returns a shard dict mapping from shard id -> shard iterator
-        The implementations for Kinesis and DynamoDB are similar but differ in various ways:
-        * Kinesis uses "StreamARN" and DynamoDB uses "StreamArn" as source parameter
-        * Kinesis uses "StreamStatus.ACTIVE" and DynamoDB uses "StreamStatus.ENABLED"
-        * Only Kinesis supports the additional StartingPosition called "AT_TIMESTAMP" using "StartingPositionTimestamp"
-        """
-        pass
+            self.shards = self.initialize_shards()
```
dfangl, line 210: As we discussed, we should really keep an eye on this and do more testing, as this might start the stream from the beginning if `TRIM_HORIZON` or `AT_...` is the iterator type.

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
```diff
@@ -113,12 +117,7 @@ def poller_loop(self, *args, **kwargs):
...
             self.current_state = EsmState.ENABLED
             self.state_transition_reason = self.user_state_reason
-            # TODO: idea to not update store state but query state from esm_worker upon querying, but async deletion needs to modify store?!
-            # Update store state !?
-            # function_version = LambdaProvider._get_function_version_from_arn(self.esm_config["FunctionArn"])
-            # state = lambda_stores[function_version.id.account][function_version.id.region]
-            # esm_update = {"State": EsmState.ENABLED}
-            # state.event_source_mappings[self.esm_config["UUID"]].update(esm_update)
+
         while not self._shutdown_event.is_set():
```
dfangl, line 121: Do we reset the shutdown event on `start` somewhere? Otherwise a restart after enabling a disabled ESM won't last for long.

dfangl (commented):
I have some questions which potentially break part of the code, but mostly LGTM!

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
joe4dev: That's correct and I skipped testing because it requires failure destinations to test this code path (implemented in the follow-up PR).

Fixed types in [8feb829](https://github.com/localstack/localstack/pull/11331/commits/8feb829ba1368d48d2495cbd61cfd99b70c1d9d3), also more readable:

```
except self.source_client.exceptions.ExpiredIteratorException as e:
```

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
joe4dev: agree, I added an explicit TODO clarifying the issue here 👍

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
joe4dev: Great catch 💯  Thank you for pointing this out.

Added the following to `start()` right before starting the poller thread:

```py
# Reset the shutdown event such that we don't stop immediately after a restart
self._shutdown_event.clear()
```

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -2117,7 +2066,8 @@ def update_event_source_mapping_v2(
 
         esm_worker = self.esm_workers[uuid]
         # Only apply update if the desired state differs
-        if enabled := request.get("Enabled") is not None:
```
joe4dev, line 2120: That was a stupid bug 🤦

dfangl (approved):
LGTM, thanks for addressing my comments!

PR #11338, opened 2024-08-10
CFn: better context on boto client errors

simonrw:
## Motivation

Client errors are often obtuse and don't show the resource that failed.

## Changes

- Wrap all internal client calls with logging to explain more
-  wrap resource provider action with extra logging to provide more context

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/parameters.py
+++ b/localstack-core/localstack/services/cloudformation/engine/parameters.py
@@ -18,10 +18,15 @@ Documentation extracted from AWS docs (https://docs.aws.amazon.com/AWSCloudForma
 
 """
 
+import logging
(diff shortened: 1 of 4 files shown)


pinzon (approved):
LGTM. Maybe in the future we could implement a boto hook that returns a better log by default.

simonrw:
> LGTM. Maybe in the future we could implement a boto hook that returns a better log by default.

I think that's a nice idea, however
* we would have to instrument both the client we pass to resource providers and the client CFn uses internally to e.g. resolve SSM parameters, and
* the `try: catch` catches the errors already, and writers of resource providers don't have to worry about this complexity making it transparent for most contributors of CFn outside of service owners.

Given these two items, I'm happy to keep the implementation as is, but I like your thinking!

PR #11344, opened 2024-08-12
fix APIGW CFN Stage with Stage Variables casing

bentsku:
## Motivation
As reported by a user, we had an issue with stage variables when declared via CloudFormation. 

Setting a stage variable to be `TestCasing` would save it as `testCasing`. As stage variables are case sensitive, this is quite an issue. 

I've tracked down the issue to be because we would call `keys_to_lower` on the parameters, which would change the casing.

In the meantime, I've added more tests around stage variables and their casing without CFN, and more validation about what the user was encoutering (the stage variable wouldn't be found, which would create an HTTP integration URI to not have a host). 

I've also found a small issue with how we would import OpenAPI/Swagger files for the method used for the integration for HTTP/HTTP_PROXY.

Resources:
- https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-variables

## Changes
- fix the CloudFormation `Stage` resource for APIGW, picking up the stage variables before the `keys_to_lower` call and using deep copy because of the nested dicts
- update a validated test to check the fix and casing
- update another test related to Stage Variables but not CFN, just to check behavior (also add a negative test for NextGen only)
- add better exception handler for `HTTP` & `HTTP_PROXY` NextGen integrations, can still be improved with negative testing, we just need to find how to trigger them

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -1257,12 +1257,10 @@ def import_api_from_openapi_spec(
                 case "AWS_PROXY":
                     # if the integration is AWS_PROXY with lambda, the only accepted integration method is POST
                     integration_method = "POST"
-                case "AWS":
+                case _:
                     integration_method = (
                         method_integration.get("httpMethod") or method_name
                     ).upper()
-                case _:
-                    integration_method = method_name
 
             connection_type = (
                 ConnectionType.INTERNET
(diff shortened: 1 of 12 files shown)


cloutierMat (approved):
Nice investigation with a clean fix!

Great job, I just left one nit, but not blocking! I am happy either way! :rocket:

PR #11345, opened 2024-08-12
CFn: support importing existing EC2 SSH keys

simonrw:
## Motivation

While exploring EC2 CloudFormation resources (via CDK) I found that when providing `PublicKeyMaterial` that the deployment failed, since this is not a valid parameter to `create_ssh_key`.

In this case, we should use `import_ssh_key`.

## Changes

* If `PublicKeyMaterial` is present in incoming resource definition, call `import_ssh_key` instead of `create_ssh_key`
* Add test to capture both behaviours


dominikschubert (approved):
LGTM! Thanks for the fix

PR #11349, opened 2024-08-13
APIGW NG: small cleanup 

bentsku:
## Motivation
Small cleanup when looking over the APIGW NG v1, removing some unused code already

## Changes
- refactor a bit the router instantiation so it makes more sense, we don't need to initialize everything if a router is passed
- remove the legacy handler


cloutierMat (approved):
LGTM :rocket:

PR #11350, opened 2024-08-14
APIGW NG: introduce new unified path style

bentsku:
## Motivation

In LocalStack API Gateway, we currently have a specific path-style format to call `execute-api`: `http(s)://<your-domain>:<port>/restapis/<api_id>/<stage>/_user_request_/<your-path>`.
The actual AWS format is this, with the host: `http(s)://<api-id>.execute-api.<domain>/<stage>/<your-path>`

There are multiple problems with our current "path style" format:
- it has `restapis` in the name but is shared with HTTP and WebSockets API in APIGW v2
- for HTTP APIs, <stage> can be optional. That format does not work with that case
- there is the custom _user_request_ between  <stage> and <your-path> , which looks weird and necessitate to re-organize the parts for the user

This PR introduces a new path format for API Gateway, available with the "Next Gen" provider:
`http(s)://<your-domain>:<port>/_aws/execute-api/<api-id>/<stage>/<your-path>`

This new format follows the same order as AWS, is clearer about the LocalStack-only part, is on the `_aws` namespace, and has `execute-api` in the name, which maps to the URL people are used to see.

It also sets up a deprecation path for the previous "path style" and log a message when people uses it with the new provider. We will also need to update the documentation accordingly.

Deprecation message:
```
2024-08-14T03:10:10.517 WARNING --- [PoolThread-twisted.internet.reactor-0] localstack.deprecations    : /restapis/<api_id>/<stage>/_user_request_ is deprecated (since 3.7.0) and will be removed in upcoming releases of LocalStack! Use /_aws/execute-api/<api_id>/<stage> instead.
```

## Changes
- register new routes for the APIGW NG router
- update the CORS handler to recognize this new router
- add deprecation on the `_user_request_` endpoint when using the NG provder
- update the tests to use this new endpoint too

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -104,6 +104,7 @@ class ApiGatewayEndpoint:
 class ApiGatewayRouter:
     router: Router[Handler]
     handler: ApiGatewayEndpoint
+    EXECUTE_API_INTERNAL_PATH = "/_aws/apigateway/execute-api"
 
     def __init__(self, router: Router[Handler] = None, handler: ApiGatewayEndpoint = None):
         self.router = router or ROUTER
@@ -134,7 +135,7 @@ class ApiGatewayRouter:
                 endpoint=self.handler,
                 strict_slashes=True,
             ),
-            # add the localstack-specific _user_request_ routes
+            # add the deprecated localstack-specific _user_request_ routes
             self.router.add(
                 path="/restapis/<api_id>/<stage>/_user_request_",
                 endpoint=self.handler,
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -134,14 +142,32 @@ def register_routes(self) -> None:
...
+                path=f"{self.EXECUTE_API_INTERNAL_PATH}/<api_id>/<stage>/",
+                endpoint=self.handler,
+                defaults={"path": ""},
+                strict_slashes=False,
+            ),
+            self.router.add(
+                path=f"{self.EXECUTE_API_INTERNAL_PATH}/<api_id>/<stage>/<greedy_path:path>",
                 endpoint=self.handler,
                 strict_slashes=True,
             ),
```
cloutierMat, line 173: A better strategy might be to match only `<api_id>` and `<api_id>/<greedy_path:path>` and have to do a bit more work when routing the request. It is added work on the rest api but would help us support HTTP api with less mess

bentsku: We had a lengthy discussion about this, and the TLDR; is this:

We're using the router to do as much parsing as it can, and if we were to remove the stage from the path, we would still need to manually parse it from every request by checking the first "resource" of the path and assigning it to the stage name.

For HTTP API, you can define a `$default` stage, so if you do not specify a stage name, it will fallback to the default. That leads to the registered routes being somewhat misleading, because if you target an HTTP API with a $default stage configured with the following path: `test/test-path`, then in our edge router endpoint, `test` will be assigned to the stage value.
The next check is to verify if that `test` is in the existing `stages`, and if not, assign the stage value to `$default`.

We don't use the `<greedy_path:path>` value in the rest of the invocation, as it is URL-decoded, and we need the raw value. In the rest of the logic, we first use `RAW_URI` which contains the entire path of the request, and we do manipulation to remove the stage name if the request is prefixed with it. The `stage` parameter here is only to help us get that first "resource" of the path, it being rightly the stage name or not is not really important, we do not know yet if the request is proper or not. 

We've also decided to model the edge router rules to the REST API format, as `stage` is mandatory. It will be a bit of a workaround to support all kind of API with the same routes, but if properly documented, should be alright. 

This Edge router endpoint can be seen as a "first pass" routing, trying to catch all API Gateway request and doing as much work as possible, and then it's up to us to properly route the request to the right API Gateway Resource/Route of the API.

cloutierMat (approved):
Thanks @bentsku for tackling this and be always ready for good discussion and clarification!

I really appreciate the switch to internal request using `_aws` :rocket:

PR #11353, opened 2024-08-14
fix test_import_with_stage_variables for multi account

bentsku:
## Motivation
Recently updated test `test_import_with_stage_variables` [broke the multi account pipeline](https://app.circleci.com/pipelines/github/localstack/localstack/27318/workflows/6163beb5-f542-4d1c-836e-7385d6ffe7bc/jobs/234706) because we used a lambda URL function in a non default region which doesn't have certificates. We however need HTTPS by default in the OpenAPI file because AWS does not accept HTTP only requests to lambda URL. 

There was a condition to replace the https scheme by http when running against LocalStack, but I've done a small mistake and did not assign the result 😅 
Anyway, this now works, you can validate the fix by running the same test with `TEST_AWS_REGION_NAME=ap-northeast-1` and see it is fixed.

## Changes
- assign the result of the `str.replace`

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_import.py
+++ b/tests/aws/services/apigateway/test_apigateway_import.py
@@ -753,8 +753,8 @@ class TestApiGatewayImportRestApi:
         )
         spec_file = load_file(OAS_30_STAGE_VARIABLES)
         if not is_aws_cloud():
-            # to make sure we return the test without https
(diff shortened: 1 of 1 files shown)


cloutierMat (approved):
Thanks for jumping on this do quickly! 🚀

PR #11354, opened 2024-08-14
Event filtering JSON detection for ESM v2 and Pipes

joe4dev:
## Motivation

A lot of filtering tests are currently failing for ESM v2 because JSON content detection is not implemented. The filtering behavior and content detection heuristics depend on the event source and might differ for ESM and Pipes.

## Changes

Implement conditional (i.e., only applies of filter contains data) decoding and re-encoding to make filtering tests pass in ESM v2. It's an ugly hack for now 😬 

## ESM v1 Implementation

The decoding and re-encoding is a hacky idea inspired by:
* `localstack-core/localstack/services/lambda_/event_source_listeners/sqs_event_source_listener.py`
* `localstack.services.lambda_.event_source_listeners.kinesis_event_source_listener.KinesisEventSourceListener._filter_records`

DynamoDB doesn't implement any of these hacks because the `dynamodb` data field is just taken "as is": `localstack.services.lambda_.event_source_listeners.dynamodb_event_source_listener.DynamoDBEventSourceListener._filter_records`

## Testing

Using the existing Lambda ESM test suite to reach parity with ESM v1 for now. We are missing big test coverage for many of these cases and this filtering pre-processing logic is hard to test because we cannot snapshot the event at the filtering stage.

Pipelines:

* `ext`: Full ext CI run (prevent Pipes regression)

Builds:

* [479879a](https://github.com/localstack/localstack/pull/11354/commits/479879ac5c8b9ba4599521ea630a009e55210ae1): [ext](https://github.com/localstack/localstack-ext/actions/runs/10388252395) ✅  (except for flaky kafka persistence test, rerun)
* [a5b2d6d](https://github.com/localstack/localstack/pull/11354/commits/a5b2d6d7b4eb74874df1d598a729e254e80950f0): [ext](https://github.com/localstack/localstack-ext/actions/runs/10388784196) ✅ 
* [1a6fc72](https://github.com/localstack/localstack/pull/11354/commits/1a6fc7239c23b460dcf0aa6c48c5017e0eeefefa): [ext](https://github.com/localstack/localstack-ext/actions/runs/10389105562) ✅  
* [4ca8e58](https://github.com/localstack/localstack/pull/11354/commits/4ca8e58356df825db2ca59458a89a0469228cb39): [ext](https://github.com/localstack/localstack-ext/actions/runs/10392173789) ✅ 

## Limitations

* One of the big limitations related to filtering is that non-matching events should forward the iterator using sequence number checkpointing. This is currently not implemented neither for the old nor the new implementation. Checkpointing the sequence number per shard should also be used to resume a disabled ESM (not implemented). This could lead to poor performance if many events are filtered out and repeatedly polled until they expire.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
@@ -10,7 +10,6 @@ from localstack.aws.api.pipes import (
     KinesisStreamStartPosition,
 )
 from localstack.services.lambda_.event_source_listeners.utils import (
-    has_data_filter_criteria,
     has_data_filter_criteria_parsed,
 )
 from localstack.services.lambda_.event_source_mapping.event_processor import (
@@ -170,7 +169,7 @@ class KinesisPoller(StreamPoller):
             return events
 
     def post_filter(self, events: list[dict]) -> list[dict]:
-        if has_data_filter_criteria(self.filter_patterns):
+        if has_data_filter_criteria_parsed(self.filter_patterns):
             # convert them back (HACK for fixing parity with v1 and getting regression tests passing)
             for event in events:
                 parsed_data = event.pop("data")
@@ -190,9 +189,9 @@ class KinesisPoller(StreamPoller):
         else:
             event["data"] = data
 
-    def parse_data(self, raw_data: str | dict) -> dict | str:
+    def parse_data(self, raw_data: str) -> dict | str:
         decoded_data = base64.b64decode(raw_data)
         return json.loads(decoded_data)
 
-    def encode_data(self, parsed_data: dict) -> dict | str | bytes:
-        return json.dumps(parsed_data).encode()
+    def encode_data(self, parsed_data: dict) -> str:
+        return base64.b64encode(json.dumps(parsed_data).encode()).decode()
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
@@ -1,4 +1,5 @@
 import base64
+import json
 import logging
 
 from botocore.client import BaseClient
@@ -99,6 +100,7 @@ class KinesisPoller(StreamPoller):
         for record in records:
             # TODO: consolidate with Kinesis event source listener:
             #  localstack.services.lambda_.event_source_listeners.kinesis_event_source_listener.KinesisEventSourceListener._create_lambda_event_payload
+            #  check `encryptionType` leading to serialization errors by Dotnet Lambdas
             sequence_number = record["SequenceNumber"]
             event = {
                 # TODO: add this metadata after filtering.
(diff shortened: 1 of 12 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
```diff
@@ -135,3 +142,57 @@ def get_sequence_number(self, record: dict) -> str:
...
+        if self.kinesis_namespace:
+            event["kinesis"]["data"] = data
+        else:
+            event["data"] = data
+
+    def parse_data(self, raw_data: str | dict) -> dict | str:
+        decoded_data = base64.b64decode(raw_data)
+        return json.loads(decoded_data)
+
+    def encode_data(self, parsed_data: dict) -> dict | str | bytes:
```
dfangl, line 197: How could that method return anything but "bytes"?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
```diff
@@ -135,3 +142,57 @@ def get_sequence_number(self, record: dict) -> str:
...
+        return json.dumps(parsed_data).encode()
```
dfangl, line 198: This is not the reverse operation of `parse_data`, there is no base64 happening here. Is this on purpose?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -135,14 +141,30 @@ def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
...
+        # convert to json for filtering (HACK for fixing parity with v1 and getting regression tests passing)
+        # localstack.services.lambda_.event_source_listeners.kinesis_event_source_listener.KinesisEventSourceListener._filter_records
+        # TODO: explore better abstraction for the entire filtering, including the set_data and get_data remapping
+        #  We need better clarify which transformations happen before and after filtering -> fix missing test coverage
+        parsed_events = self.pre_filter(polled_events)
+        # TODO: advance iterator past matching events!
+        #  We need to checkpoint the sequence number for each shard and then advance the shard iterator using
+        #  GetShardIterator with a given sequence number
+        #  https://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetShardIterator.html
+        #  Failing to do so kinda blocks the stream resulting in very high latency.
```
dfangl, line 157: Why would the shard iterator not advance if we received the messages? Even if none of them match, we still set the new shard iterator, right?

dfangl (approved):
LGTM, thanks for the fixing!

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
dfangl: For the record: It was not the same, but the BytesEncoder when dumping the event with json reencoded it anyway. Joel changed it though, since we should not make any changes to the events (even though that can still happen with the current approach) while filtering.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
joe4dev: fixed to `str`
(leftover from polymorphic version before I knew that's only a Kinesis thing)

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/kinesis_poller.py
joe4dev: Very good catch 💯 

Thanks for debugging session detecting the underlying `_parsed` suffix in the helper method etc :)

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
joe4dev: That only happens in (partial) error case. For example, if out of 10 events, the first five do not match the filter but others in the batch cause a (partial) error. We need to figure out whether AWS forwards the iterator passing the first non-matching events because they would be filtered out in a future retry anyway.

PR #11355, opened 2024-08-14
switch versioning to setuptools_scm

alexrashed:
## Motivation
TBD

## Changes
TBD

## Testing
TBD

## TODO
- [ ] TBD


bin/docker-helper.sh
```diff
@@ -192,16 +200,11 @@ function cmd-push() {
       docker push $TARGET_IMAGE_NAME:$MAJOR_VERSION.$MINOR_VERSION.$PATCH_VERSION-$PLATFORM
     }
 
-    if [ -n "$FORCE_VERSION_TAG_PUSH" ] && [ "$FORCE_VERSION_TAG_PUSH" -eq "1" ]; then
```
silv-io, line 195: Removing this means we cannot force a tag push anymore, no?

.circleci/config.yml
```diff
@@ -815,8 +815,7 @@ jobs:
           name: Publish a dev release
           command: |
             source .venv/bin/activate
-            bin/release-helper.sh set-ver $(bin/release-helper.sh next-dev-ver)
-            make publish
+            make dev-publish || echo "Not publishing a dev release"
```
alexrashed, line 818: I think we don't need an additional make target here, but can really just run `make publish` but either check if the version is already published, or ignore failures (because if it fails, it's basically always because the version already exists).

.github/workflows/tests-s3-image.yml
```diff
@@ -96,6 +96,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
alexrashed, line 100: Maybe add a comment here why this is necessary now...

bin/docker-helper.sh
```diff
@@ -153,6 +159,8 @@ function cmd-push() {
     _enforce_platform
     _set_version_defaults
 
+    echo "VERSION: $(_get_current_version)"
+
```
alexrashed, line 163: Oups, this is a leftover (from my commit) which should be removed.

bin/docker-helper.sh
alexrashed: Yes, that's true, but this was only introduced to allow proper testing of the function in the bats tests. Afaik this isn't used anywhere.

localstack-core/localstack/constants.py
```diff
@@ -1,10 +1,12 @@
...
+from localstack.version import __version_tuple__
 
-# temporary fix to avoid issues with missing version package in image
-# FIXME: remove this once all images contain the version package
-VERSION = __version__
+# Adapt the version from setuptools_scm to the version format used by LocalStack before
+if len(__version_tuple__) > 3:
+    VERSION = ".".join([str(i) for i in __version_tuple__[0:3]]) + ".dev"
+else:
+    VERSION = ".".join([str(i) for i in __version_tuple__[0:3]])
```
alexrashed, line 9: I don't think this is necessary anymore (since we use `local_scheme = "no-local-version"`).

Dockerfile.s3
```diff
@@ -13,7 +13,7 @@ RUN --mount=type=cache,target=/var/cache/apt \
         # Install dependencies to add additional repos
         apt-get install -y --no-install-recommends \
             # Runtime packages (groff-base is necessary for AWS CLI help)
-             ca-certificates curl make openssl
+             ca-certificates curl make openssl git
```
alexrashed, line 16: I think we might not need `git` anymore, because we use `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LOCALSTACK_CORE` instead now.

Dockerfile.s3
```diff
@@ -53,19 +53,19 @@ RUN --mount=type=cache,target=/var/cache/apt \
...
 ADD Makefile pyproject.toml requirements-base-runtime.txt ./
-# add the VERSION file to invalidate docker layers with version bumps
-ADD VERSION ./
 # add the localstack start scripts (necessary for the installation of the runtime dependencies, i.e. `pip install -e .`)
 ADD bin/localstack bin/localstack.bat bin/localstack-supervisor bin/
 
-# install dependencies to run the LocalStack Pro runtime and save which ones were installed
+# create the localstack-core directory for the version autogeneration from setuptools_scm
+RUN mkdir -p /opt/code/localstack/localstack-core/localstack
+
```
alexrashed, line 65: I think this is not necessary anymore.

alexrashed (commented):
This is quite an awesome iteration, @silv-io! We're nearly there! Just found a few nitpicks and leftovers we need to tackle before merging ;) 🚀

alexrashed (commented):
Awesome! Thanks, @silv-io for addressing the comments and questions! I think this is good to get looked at by another set of eyes, get a bit more testing, and then get merged! 🥳 💯

bin/docker-helper.sh
```diff
@@ -50,7 +49,15 @@ function _fail {
 }
 
 function _get_current_version() {
-    cat ${VERSION_FILE}
+    # setuptools_scm will be installed transparently if not available, Python3 is expected to be present
+    if ! python3 -m pip -qqq list | grep -F "setuptools_scm"; then
+      python3 -m pip install -qqq setuptools_scm > /dev/null 2>&1
+    fi
```
dfangl, line 55: I understand the reasoning for this, but I would like to discuss if this is something we can somehow avoid? Especially since developers might use `make docker-build` and use this script, and might not have the venv activated - it might lead to some modifications to the global / pyenv install users did not assume.

Same for the `release-helper.sh`

bin/docker-helper.sh
```diff
@@ -50,7 +49,15 @@ function _fail {
...
+    python3 -m setuptools_scm
+}
+
+function _is_release_commit() {
+    [[ $(_get_current_version) =~ ^[0-9]+.[0-9]+.[0-9]+$ ]]
```
dfangl, line 60:
```suggestion
    [[ $(_get_current_version) =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
```
Otherwise we match `0x0x0` as well. One could argue that never happens anyway usually, but why not match it properly :)

bin/docker-helper.sh
```diff
@@ -192,16 +202,11 @@ function cmd-push() {
...
-    if [ -n "$FORCE_VERSION_TAG_PUSH" ] && [ "$FORCE_VERSION_TAG_PUSH" -eq "1" ]; then
-      echo "Force-enabled version tag push."
-      _push
-    elif [ -n "$FORCE_VERSION_TAG_PUSH" ] && [ "$FORCE_VERSION_TAG_PUSH" -eq "0" ]; then
-      echo "Force-disabled version tag push. Not pushing any other tags."
-    elif (git diff HEAD^ ${VERSION_FILE} | tail -n 1 | grep -v '.dev'); then
-      echo "Pushing version tags, version has changed in last commit."
+    if _is_release_commit; then
+      echo "Pushing version tags, we're building the commit of a version tag."
       _push
```
dfangl, line 207: Not part of this commit, but I think we could rename that function to be something more expressive, like `_push_versioned_tags` or something.

bin/release-helper.sh
```diff
@@ -195,18 +173,14 @@ function cmd-git-commit-release() {
 
     echo $1 || verify_valid_version
 
-    for file in ${VERSION_FILE} ${VERSION_PY} ${DEPENDENCY_FILE}; do
-        git add "$file" || echo "did not add file '$file' to commit"
-    done
-    git commit -m "release version ${1}"
+    git add "${DEPENDENCY_FILE}"
+    git commit --allow-empty -m "release version ${1}"
```
dfangl, line 177: Maybe add a comment here why allow empty is necessary in community even though the git add above.

tests/bin/test.docker-helper.bats
```diff
@@ -99,48 +120,46 @@ setup_file() {
...
 @test "push pushes built image wo versions" {
   export IMAGE_NAME="localstack/test"
-  export MAIN_BRANCH="`git branch --show-current`"
+  export MAIN_BRANCH="main"
   export DOCKER_USERNAME=test
   export DOCKER_PASSWORD=test
   export PLATFORM=arm64
-  export FORCE_VERSION_TAG_PUSH=0
-  export VERSION_FILE="$BATS_TEST_DIRNAME/files/VERSION"
+  export TEST_SPECIFIC_VERSION="3.6.1.dev45+g15cb62bd5.d20240814"
```
dfangl, line 127: Comment by silvio during our call: We can remove the local part here, as it is not present anyway.

tests/bin/test.release-helper.bats
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@test "non-zero exit on unknown command" {
+  run bin/release-helper.sh unknown
+
+  [ "$status" -ne 0 ]
+}
+
+@test "get-ver prints a correct version" {
+  export TEST_SPECIFIC_VERSION="3.6.1.dev45+g15cb62bd5.d20240814"
```
dfangl, line 44: Maybe remove the local version here as well.

dfangl (approved):
Looks good, great push forward! I just had some minor nits regarding regexes and so on, but nothing major!

bin/docker-helper.sh
silv-io: I think we can avoid this specific issue by depending on the `venv` make target in the `docker-build` make targets

/cc @alexrashed

alexrashed: I don't think we should couple the docker-helper or the release-helper to the `Makefile`, that's even tighter than this coupling with `setuptools_scm`.
I would propose to keep track of this and tackle this in the coming days (in the next iteration), after some discussion and playing around with possible ways to circumvent this.

alexrashed: FYI: This is being tackled in https://github.com/localstack/localstack/pull/11441 by @silv-io! 🚀

Issue #11356, opened 2024-08-14
feature request: support sse-c in s3


jiachengdb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

It seems that localstack S3 doesn't support [SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html) encryption.

This repro shows that when uploading an object to localstack with sse-c headers, the object still uses the default SSE-S3 encryption as shown in the returned header `'x-amz-server-side-encryption': 'AES256'`.

```
$ podman run -d -p 4566:4566 localstack/localstack:latest
```

```
import boto3
from botocore.client import Config
import base64
import hashlib

# Localstack server details
localstack_url = "http://localhost:4566"
access_key = "test"
secret_key = "test"
bucket_name = "test-bucket"
object_name = "your-object.txt"
data = b"Hello, World!"  # The data to upload

# Create an S3 client
s3_client = boto3.client(
    's3',
    endpoint_url=localstack_url,
    aws_access_key_id=access_key,
    aws_secret_access_key=secret_key,
    config=Config(signature_version='s3v4')
)

# Generate a 256-bit (32 bytes) AES key for SSE-C
sse_customer_key=b'1234567890abcdef1234567890abcdef'
sse_customer_key_base64 = base64.b64encode(sse_customer_key).decode('utf-8')
sse_customer_key_md5 = base64.b64encode(hashlib.md5(sse_customer_key).digest()).decode('utf-8')

# Upload the object with SSE-C
s3_client.put_object(
    Bucket=bucket_name,
    Key=object_name,
    Body=data,
    SSECustomerAlgorithm='AES256',
    SSECustomerKey=sse_customer_key_base64,
    SSECustomerKeyMD5=sse_customer_key_md5
)

print(f"Object '{object_name}' has been uploaded to bucket '{bucket_name}' with SSE-C.")

response = s3_client.get_object(
    Bucket=bucket_name,
    Key=object_name,
    SSECustomerAlgorithm='AES256',
    SSECustomerKey=sse_customer_key_base64,
    SSECustomerKeyMD5=sse_customer_key_md5
)

print(response)
```

Program output:

```
Object 'your-object.txt' has been uploaded to bucket 'test-bucket' with SSE-C.
[...]

PR #11358, opened 2024-08-14
implement SSE-C encryption for S3

bentsku:
## Motivation
As reported in #11356, we did not support SSE-C for S3. This PR implements it, we still don't encrypt the files locally, but this is irrelevant for the usage, as AWS decrypts objects on the fly, and for now we don't really have a use-case for it.

I've implemented it following the documentation page: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html

And tried to think of most use-cases, as this is needed for the object-store Rust crate (see https://github.com/apache/arrow-rs/pull/6230) . 

I haven't implemented it for the pre-signed POST yet, this can be done in a follow-up. 

## Changes
- implement SSE-C parameter validation for `PutObject`, `GetObject`, `HeadObject`, `GetObjectAttributes`, `CopyObject`, `CreateMultipartUpload` and `UploadPart`. 
- add some tests around it, positive and negative
- properly return the right fields

Note: test failure unrelated, it's an OpenSearch failure, not blocking the merge

### DIFF
--- a/localstack-core/localstack/aws/api/s3/__init__.py
+++ b/localstack-core/localstack/aws/api/s3/__init__.py
@@ -942,6 +942,14 @@ class EntityTooLarge(ServiceException):
     ProposedSize: Optional[ProposedSize]
 
 
+class InvalidEncryptionAlgorithmError(ServiceException):
+    code: str = "InvalidEncryptionAlgorithmError"
+    sender_fault: bool = False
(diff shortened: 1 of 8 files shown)


thrau (approved):
nice! LGTM. congrats on the quick turnaround :100:

Issue #11313, opened 2024-08-06
bug: KMS "Perhaps the blob didn't come from LocalStack" error message when decrypting using public key generated by localstack


getglad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Localstack can generate a KMS `RSA_4096` key and export the public key. This key material can encrypt a payload on the client side using RSA OAEP. However, the KMS `decrypt` operation fails with the below message even when passing in the key id.

> operation error KMS: Decrypt, https response error StatusCode: 400, RequestID: 9f9efecf-6e08-447d-aef0-6c49dff32ed1, InvalidCiphertextException: LocalStack is unable to deserialize the ciphertext blob. Perhaps the blob didn't come from LocalStack

### Expected Behavior

Given the use of a key pair Localstack owns, I would expect the `decrypt` operation should be possible.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    `localstack start` (homebrew install)

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

It fails on the `decryptOutput, err := client.Decrypt(context.TODO(), decryptInput)` line below

```go
func TestKMSKeyFlow(t *testing.T) {
	// Generate a key
	publicKeyString, keyId := makeKey()

	// Encrypt payload using the public key
        pubKeyInterface, _ := x509.ParsePKIXPublicKey([]byte(publicKeyString))
	publicKey := pubKeyInterface.(*rsa.PublicKey)
	plaintext := []byte("Hello, World!")
	rsaCiphertext, err := crypto.EncryptRSAOAEP(publicKey, plaintext)
	if err != nil {
		t.Errorf("Failed to encrypt plaintext: %v", err)
	}

	// Decrypt the payload
        decryptedMsg := decryptMessage(string(rsaCiphertext), keyId)

	if string(plaintext) != string(decryptedMsg) {
		t.Errorf("Decrypted msg does not match original msg")
	}
}

func makeKey() (string, string) {
	client := kmsClient()
	// Create the RSA_4096 key``
	createKeyInput := &kms.CreateKeyInput{
		KeySpec:     types.KeySpec("RSA_4096"),
		KeyUsage:    types.KeyUsageType("ENCRYPT_DECRYPT"),
		Description: aws.String("RSA_4096 key for encryption and decryption"),
		Tags: []types.Tag{
			{
[...]

PR #11359, opened 2024-08-15
Init KMS support for offline encryption with RSA key

getglad:
## Motivation

Localstack can generate a KMS RSA_4096 key and export the public key. This key material can encrypt a payload on the client side using RSA OAEP. However, the KMS decrypt operation fails with the below message even when passing in the key id.

    operation error KMS: Decrypt, https response error StatusCode: 400, RequestID: 9f9efecf-6e08-447d-aef0-6c49dff32ed1, InvalidCiphertextException: LocalStack is unable to deserialize the ciphertext blob. Perhaps the blob didn't come from LocalStack

I brought this up as an issue, and @sannya-singal indicated that while this is an expected limitation, it could be changed. It wasn't exactly clear what exactly was on the backlog, I've tried to change it to do what I at least have noticed.

https://github.com/localstack/localstack/issues/11313

## Changes

Instead of `deserialize_ciphertext_blob` or die, the process now will support a flow that can look up the key and survive if the key ID isn't wrapped into the `ciphertext_blob`. If there is no `ciphertext` as a result of `deserialize_ciphertext_blob`, it assumes it is an RSA key and attempts to use the private key using a new `decrypt_rsa` path.

To avoid introducing breaking changes, I didn't attempt to change how "online" encryption (using `kms.encrypt`) is implemented.

I'm not positive that this approach makes _good_ assumptions, but having looked over what else is tested for in the library, it may be reasonable.

## Testing

I've added a new unit test for KMS. It passes, as do the other tests for KMS, and so does `make lint`.

This is my first `localstack` PR, and I'm having a hard time getting `make test` to run (it appears to hang), so its hard to tell w/o opening the PR how the rest of it runs.

## TODO

Happy to take feedback on the approach - gave it a good first effort based on what I could see

### DIFF
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -313,6 +313,20 @@ class KmsKey:
             self.crypto_key.key_material, ciphertext.ciphertext, ciphertext.iv, ciphertext.tag, aad
         )
 
+    def decrypt_rsa(self, encrypted: bytes) -> bytes:
+        private_key = crypto_serialization.load_der_private_key(
+            self.crypto_key.private_key, password=None, backend=default_backend()
+        )
+        decrypted = private_key.decrypt(
+            encrypted,
+            padding.OAEP(
+                mgf=padding.MGF1(algorithm=hashes.SHA256()),
+                algorithm=hashes.SHA256(),
+                label=None,
+            ),
+        )
+        return decrypted
+
     def sign(
         self, data: bytes, message_type: MessageType, signing_algorithm: SigningAlgorithmSpec
     ) -> bytes:
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -359,6 +359,10 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
 
         return context.account_id, context.region, key_id_or_arn
 
+    @staticmethod
+    def _is_rsa_spec(key_spec: str) -> bool:
+        return key_spec in [KeySpec.RSA_2048, KeySpec.RSA_3072, KeySpec.RSA_4096]
+
     #
     # Operation Handlers
     #
@@ -974,16 +978,25 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
         # Since LocalStack doesn't currently do asymmetrical encryption, there is a question of modeling here: we
         # currently expect data to be only encrypted with symmetric encryption, so having key_id inside. It might not
         # always be what customers expect.
-        try:
-            ciphertext = deserialize_ciphertext_blob(ciphertext_blob=ciphertext_blob)
-        except Exception:
-            raise InvalidCiphertextException(
-                "LocalStack is unable to deserialize the ciphertext blob. Perhaps the "
-                "blob didn't come from LocalStack"
-            )
-        account_id, region_name, key_id = self._parse_key_id(key_id or ciphertext.key_id, context)
+        if key_id:
+            account_id, region_name, key_id = self._parse_key_id(key_id, context)
+            try:
+                ciphertext = deserialize_ciphertext_blob(ciphertext_blob=ciphertext_blob)
+            except Exception:
+                ciphertext = None
+                pass
+        else:
+            try:
+                ciphertext = deserialize_ciphertext_blob(ciphertext_blob=ciphertext_blob)
+                account_id, region_name, key_id = self._parse_key_id(ciphertext.key_id, context)
+            except Exception:
+                raise InvalidCiphertextException(
+                    "LocalStack is unable to deserialize the ciphertext blob. Perhaps the "
+                    "blob didn't come from LocalStack"
+                )
+
         key = self._get_kms_key(account_id, region_name, key_id)
-        if key.metadata["KeyId"] != ciphertext.key_id:
+        if ciphertext and key.metadata["KeyId"] != ciphertext.key_id:
             raise IncorrectKeyException(
                 "The key ID in the request does not identify a CMK that can perform this operation."
             )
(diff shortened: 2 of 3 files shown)


sannya-singal:
Hey @getglad, thanks for contributing to Localstack 🙌  As a part of the process, can you please take a moment and sign the CLA as highlighted [here](https://github.com/localstack/localstack/pull/11359#issuecomment-2290543210)?

getglad:

localstack-core/localstack/utils/crypto.py
```diff
@@ -184,3 +186,16 @@ def decrypt(
...
+    private_key = serialization.load_der_private_key(key, password=None, backend=default_backend())
+    decrypted = private_key.decrypt(
+        encrypted,
+        padding.OAEP(
+            mgf=padding.MGF1(algorithm=hashes.SHA256()),
+            algorithm=hashes.SHA256(),
+            label=None,
+        ),
+    )
+    return decrypted
```
sannya-singal, line 201: Since this logic is centric to KMS I would recommend you to move this to the `kms/models.py` instead.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -992,7 +1000,10 @@ def decrypt(
         self._validate_key_state_not_pending_import(key)
 
         try:
-            plaintext = key.decrypt(ciphertext, encryption_context)
+            if not ciphertext:
+                plaintext = key.decrypt_rsa(ciphertext_blob)
```
sannya-singal, line 1004: This implementation is very specific to RSA asymmetric keys decrypt online and encrypt offline. It would be great if we can verify that only if the key is an RSA key then it should be passed to the `decrypt_rsa` function. Also, we can explicitly mention this in a short comment above here.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -992,7 +1000,10 @@ def decrypt(
...
+            else:
+                plaintext = key.decrypt(ciphertext or ciphertext_blob, encryption_context)
```
sannya-singal, line 1006: We should be using the `ciphertext` to decrypt here: 

`plaintext = key.decrypt(ciphertext, encryption_context)`

localstack-core/localstack/utils/crypto.py
```diff
@@ -184,3 +186,16 @@ def decrypt(
...
+
+def decrypt_rsa(key: bytes, encrypted: bytes) -> bytes:
+    private_key = serialization.load_der_private_key(key, password=None, backend=default_backend())
+    decrypted = private_key.decrypt(
+        encrypted,
+        padding.OAEP(
+            mgf=padding.MGF1(algorithm=hashes.SHA256()),
+            algorithm=hashes.SHA256(),
+            label=None,
+        ),
```
sannya-singal, line 199: Can we use the decrypt from crypto utils here?

localstack-core/localstack/utils/crypto.py
sannya-singal, line 201: I would recommend you to move `decrypt_rsa` to the `kms/models.py`

sannya-singal (changes requested):
Awesome 🚀 Thanks for adding these changes to KMS and increasing the parity of the service. I have few questions and comments, but its really great work 🎉

localstack-core/localstack/services/kms/provider.py
getglad: Makes sense - let me know what you think of the additional check + comments

localstack-core/localstack/services/kms/models.py
```diff
@@ -314,6 +329,9 @@ def decrypt(
             self.crypto_key.key_material, ciphertext.ciphertext, ciphertext.iv, ciphertext.tag, aad
         )
 
+    def decrypt_rsa(self, encrypted: bytes) -> bytes:
+        return _decrypt_rsa(self.crypto_key.private_key, encrypted)
+
```
sannya-singal, line 334: Any specific reason we keep the other helper function `_decrypt_rsa` within here now? 🤔 Considering there is no additional logic or abstraction we perform additionally before calling it we can move the entire thing here itself.

localstack-core/localstack/services/kms/provider.py
sannya-singal: I would suggest to use the RSA values from `KeySpec` rather than hardcoding it here. Also to keep the code really clean consider creating a helper function, so we can do something like this: 

```
def _validate_rsa_key(self, key_spec: KeySpec) -> bool: 
        # perform the logic here and return the boolean value 
```

For the comment, we can simplify it further: 

```
# TODO: Extend the implementation to handle additional encryption/decryption scenarios 
# beyond the current support for offline encryption and online decryption using RSA keys if key id exists in 
# parameters, where `ciphertext_blob` will not be deserializable.
```

localstack-core/localstack/services/kms/models.py
getglad: @sannya-singal good call - no I just got a bit of tunnel vision 👍

localstack-core/localstack/services/kms/provider.py
getglad: @sannya-singal - I implemented the helper method the way I think you have in mind. I didn't see the `KeySpec` enum earlier, so thanks for pointing that out. 👍 FWIW `key.crypto_key.key_spec` is a `str`

sannya-singal (approved):
Thanks for actively working on this PR and resolving the reviews @getglad 🚀 Once the pipeline is green and I will merge the PR.

sannya-singal:
@getglad can I ask you to rebase the PR?

getglad:
@sannya-singal - I rebased with the `master` branch and squashed down to 1 commit. 🤞

sannya-singal (commented):
Thanks @getglad 🎉 I have triggered the pipeline, once its green I will merge the PR.

dhudec:
This fix hardcodes SHA-256 for the OAEP padding and ignores the `EncryptionAlgorithm` parameter passed to the Decrypt API. This means `RSAES_OAEP_SHA_1` still fails — ciphertext encrypted with SHA-1 padding can't be decrypted when the implementation always uses SHA-256. Filed #13864 to track.

PR #11363, opened 2024-08-15
[Lambda DevX] Introducing Lambda Debug Mode Config

MEPalma:
## Motivation
Currently, Lambda Debug Mode supports debugging of only one Lambda function at a time, using a remote debugger connected via a port specified through Lambda Docker Flags during startup. 

The proposed changes introduce a new configuration feature for Lambda Debug Mode that enables users to assign specific debug ports to each Lambda function, as well as other parameters. This is achieved though a yaml file, such as:

```yaml
functions:
  arn:aws:lambda:eu-central-1:000000000000:function:foo:
    debug-port: 19891
  arn:aws:lambda:eu-central-1:000000000000:function:bar:
    debug-port: 19892
    enforce-timeouts: true
```

In this initial release, users can also disable the automatic timeout extensions that are normally applied by starting LocalStack with Lambda Debug Mode on.

The configuration can be passed to LocalStack through the new `LAMBDA_DEBUG_MODE_CONFIG_PATH` environment variable, which once set is only sourced and reflected in the running logic of the app is `LAMBDA_DEBUG_MODE` is set. 

## Changes
- Added a new environment variable `LAMBDA_DEBUG_MODE_CONFIG_PATH` which sources the yaml Lambda Debug Mode Config file
- Added courtesy query function for various debug parameters of lambda functions
- Added a new pydantic based specification of Lambda Debug Mode Config files, including mechanisms to parse and validate the file provided through `LAMBDA_DEBUG_MODE_CONFIG_PATH`, as well as useful error messaging logs for each of these steps
- Added a singleton lambda debug session object which allows the rest of the system to interact with details of the lambda debug configuration
- Added various non state affecting changes to configure timeouts and debug ports of lambda functions and containers
- Unit testing for base lambda debug mode config file parsing

## Testing
We do not yet have the necessary stack to test for lambda debugging integration tests. What follows is a guide to manually test a base scenario of this feature.

1. **Deploy Lambda Functions**
    - Deploy two Python Lambda functions, each with an evaluation timeout set to 1 second.
    - These functions are integrated with API Gateway, which has a 50ms timeout.
    - Example Lambda Function 1:
      ```python
      DEBUGGER_ENDPOINT = "0.0.0.0"
      DEBUGGER_PORT = 19891
      
      def handler_1(event, context):
          """Lambda handler that will get invoked by the LocalStack runtime"""
      
          # wait for the debugger to get attached
          import debugpy
          debugpy.listen((DEBUGGER_ENDPOINT, DEBUGGER_PORT))
          debugpy.wait_for_client()  # blocks execution until client is attached
      
          # print the incoming invocation event
          print(event)
          return {"body": "Hello from handler_1"}
      ```
    - Example Lambda Function 2:
      ```python
      DEBUGGER_ENDPOINT = "0.0.0.0"
      DEBUGGER_PORT = 19892
      
      def handler_2(event, context):
          """Lambda handler that will get invoked by the LocalStack runtime"""
      
          # wait for the debugger to get attached
          import debugpy
          debugpy.listen((DEBUGGER_ENDPOINT, DEBUGGER_PORT))
          debugpy.wait_for_client()  # blocks execution until client is attached
      
          # print the incoming invocation event
          print(event)
          return {"body": "Hello from handler_2"}
      ```
2. **Configure Debugging**
    - Use the **Lambda Debug Mode Config File** to set up the debugging environment:
        - Lambda Function 1 debugged on port 19891.
        - Lambda Function 2 debugged on port 19892.
    - Lambda Debug Mode Config file
    
    ```yaml
    functions:
      arn:aws:lambda:eu-central-1:000000000000:function:handler_1-handler_1:
        debug-port: 19891
      arn:aws:lambda:eu-central-1:000000000000:function:handler_2-handler_2:
        debug-port: 19892
    ```
    
3. **Enable Lambda Debug Mode**
    - Activate Lambda Debug Mode to adjust runtime evaluation parameters.
    - How:
        ```bash
        LAMBDA_DEBUG_MODE=1
        LAMBDA_DEBUG_MODE_PATH=/path/to/file/debug_config.yaml
        ```
        
    
4. **Debug**
    - Invoke Lambda Function 1 through ApiGateway
    - Verify the system hangs and the specified timeouts are not respected for ApiGateway nor Lambda
    - Connect remotely to the specified debug port
    - Verify that both Lambda functions can be debugged and upon termination the function returns the expected result
    - Repeat for Lambda Function 2


localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
```diff
@@ -0,0 +1,68 @@
...
+    LambdaDebugConfig,
+    LambdaDebugModeConfig,
+    load_lambda_debug_mode_config,
+)
+
+LOG = logging.getLogger(__name__)
+
+
+class _LambdaDebugModeSession:
+    _instance = None
```
joe4dev, line 16: @dfangl Does this match our common way of creating singletons?

tests/unit/lambda_debug_mode/test_config_parsing.py
```diff
@@ -0,0 +1,120 @@
...
+    null
+"""
+
+DEBUG_CONFIG_NULL_DEBUG_PORT = """
+functions:
+  arn:aws:lambda:eu-central-1:000000000000:function:functionname:
+    debug-port: null
+"""
+
+DEBUG_CONFIG_NULL_TIMEOUT_DISABLE = """
```
joe4dev, line 26: Should this contain a timeout disable flag?

localstack-core/localstack/config.py
```diff
@@ -433,6 +433,9 @@ def in_docker():
 # When enabled it triggers specialised workflows for the debugging.
 LAMBDA_DEBUG_MODE = is_env_true("LAMBDA_DEBUG_MODE")
```
joe4dev, line 434: Can we add `LAMBDA_DEBUG_MODE` to the `TRACKED_ENV_VAR` in `localstack-core/localstack/runtime/analytics.py` for telemetry insights?

localstack-core/localstack/config.py
```diff
@@ -433,6 +433,9 @@ def in_docker():
...
 
+# path to the lambda debug mode configuration file.
+LAMBDA_DEBUG_MODE_CONFIG_PATH = os.environ.get("LAMBDA_DEBUG_MODE_CONFIG_PATH")
```
joe4dev, line 437: Can we add all new environment variables to `CONFIG_ENV_VARS` in `config.py` at the bottom such that they work using the CLI?

joe4dev (approved):
Excellent work enabling multi-function debugging for Lambda 👏👏
Lovely validations and config test coverage 💯  

We should enable telemetry for the debug flag and add the new configs to the config list for CLI compatibility. Otherwise, nothing blocking a merge (just nits, discussions).

Given the lack of automated testing (for now), I highly recommend adding a `## Testing` section with manual instructions including suggested testing steps for reviewers, potentially linking to an example (the example you demonstrated in our Tue session would be very suitable).


I would appreciate an extra pair of 👀 from @dfangl on this one ✨

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+from yaml import Loader, MappingNode, MarkedYAMLError, SafeLoader
+
+from localstack.aws.api.lambda_ import Arn
+
+LOG = logging.getLogger(__name__)
+
+
+class LambdaDebugConfig(BaseModel):
+    debug_port: Optional[int] = Field(None, alias="debug-port")
+    timeout_disable: bool = Field(False, alias="timeout-disable")
```
joe4dev, line 17: I'm wondering about the naming: Would `enforce_timeouts: True|False` be clearer? What's your naming suggestion @dfangl @dominikschubert ?

Current behavior:
* `LAMBDA_DEBUG_MODE=0` -> never disable any timeout
* `LAMBDA_DEBUG_MODE=1`
  * `timeout-disable` not set or no debug config for Lambda function ARN available -> extend timeouts
  * `timeout-disable=False` -> extend timeouts
  * `timeout-disable=True` -> enforce timeouts as expected

Am I getting confused here or is this a logic error (the default value is `False`, which gets negated in `is_lambda_debug_timeout_for`) 🤔

sidenote: Technically, we're not fully disabling the timeouts, but extend them.


We need to document the behavior clearly after merging ...


PS: migrated from wrong follow-up PR

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+def _to_qualified_lambda_function_arn(lambda_arn: Arn) -> Arn:
+    # Returns the $LATEST qualified version of a structurally unqualified version of a lambda Arn iff this
+    # if detected to be structurally unqualified. Otherwise, it returns the given string.
+    if not lambda_arn:
+        return lambda_arn
+    lambda_arn_parts = lambda_arn.split(":")
+    lambda_arn_parts_len = len(lambda_arn_parts)
+
+    # The arn is qualified and with a non-empy qualifier.
+    is_qualified = lambda_arn_parts_len == 8
```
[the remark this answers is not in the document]
gregfurman: You could use the `api_utils.function_locators_from_arn` which will return a tuple of `(name, qualifier, account, region)` from a different schema of ARN (qualified, unqualified, full, partial, just function name, etc).

So your check could be:

```python
_, qualifier, *_ = api_utils.function_locators_from_arn(arn) 
is_arn_qualified = qualifier != None
# do stuff with the above ...
```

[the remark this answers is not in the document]
gregfurman: Keep in mind until the above PR is merged in, an invalid/unmatched ARN will return a single `None` (not a tuple) which could break some things.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+from pydantic import BaseModel, Field, ValidationError
+from yaml import Loader, MappingNode, MarkedYAMLError, SafeLoader
+
+from localstack.aws.api.lambda_ import Arn
+
+LOG = logging.getLogger(__name__)
+
+
+class LambdaDebugConfig(BaseModel):
+    debug_port: Optional[int] = Field(None, alias="debug-port")
```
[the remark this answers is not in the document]
MEPalma: only `debug-port` is recognised

localstack-core/localstack/services/lambda_/invocation/executor_endpoint.py
```diff
@@ -197,11 +197,16 @@ def invoke(self, payload: Dict[str, str]) -> InvocationResult:
             raise InvokeSendError(
                 f"Error while sending invocation {payload} to {invocation_url}. Error Code: {response.status_code}"
             )
+
+        # Set a reference future awaiting limit to ensure this process eventually ends,
+        # with timeout errors being handled by the lambda evaluator.
+        # The following logic selects which maximum waiting time to consider depending
+        # on whether the application is being debugged or not. Note however, that if
+        # debugging timeouts are disabled for the lambda function invoked at this endpoint,
+        # the lambda function will already enforce the expected timeouts.
```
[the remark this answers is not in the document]
MEPalma: correct -- I have adjusted this comment for clarity

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
dfangl: I agree with you here, it is confusing. For a variable called `timeout-disable` I would expect the flag being reversed. I personally prefer `enforce_timeout` in this case, but even if we reverse it, I am not a bit fan of `timeout-disable`, I think even `disable-timouts` would be clearer here.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode.py
```diff
@@ -0,0 +1,31 @@
...
+def lambda_debug_port_for(lambda_arn: Arn) -> Optional[int]:
+    if not is_lambda_debug_mode():
+        return None
+    debug_configuration = lambda_debug_mode_session.debug_config_for(lambda_arn=lambda_arn)
+    if debug_configuration is None:
+        return None
+    return debug_configuration.debug_port
+
+
+def is_lambda_debug_timeout_for(lambda_arn: Arn) -> bool:
```
dfangl, line 25: Would be nice to rename this function to be clearer, and maybe add docstrings to this entire module.

`is_lambda_debug_timeout_for` does not really make sense for me. Maybe `is_lambda_debug_timeout_enabled_for`?

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # The arn is qualified and with a non-empy qualifier.
+    is_qualified = lambda_arn_parts_len == 8
+    if is_qualified and lambda_arn_parts[-1]:
+        return lambda_arn
+
+    # The arn is not unqualified, but probably erroneous, pass the value upstream.
+    is_unqualified = lambda_arn_parts_len == 7
+    if not is_unqualified:
+        return lambda_arn
```
dfangl, line 145: As far as I can see, we just ignore invalid arns, for example those ending with `:` (so they have 8 parts, but are missing the qualifier, or totally invalid arns) and still set them in the internal config? Personally, I would prefer raising an error to silently accepting them, because they will never match anything.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
dfangl: No, we usually create singletons with a `.get()` function:

```
    @staticmethod
    @singleton_factory
    def get() -> "_LambdaDebugModeSession":
        """Returns a singleton instance of the debug mode session."""
        return _LambdaDebugModeSession()
```

I would prefer to avoid overriding `__new__` here. Instead ofusing the module global, we can always call `_LambdaDebugModeSession.get()`. Might be nice to rename it to avoid the leading `_` then, though.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+
+def validate_lambda_debug_mode_config(config: LambdaDebugModeConfig) -> None:
+    _validate_lambda_debug_mode_config(
+        validation_state=_LambdaDebugModeConfigValidationState(), config=config
+    )
+
+
+def _validate_lambda_debug_mode_config(
+    validation_state: _LambdaDebugModeConfigValidationState, config: LambdaDebugModeConfig
+):
```
dfangl, line 110: This function is called `validate_` but does actively change the values of the config (by replacing unqualified ARNs with qualified ARNs). It should be clear from the method that we are not only validating here, but also "canonicalizing". So maybe renaming it, or at least adding it in a docstring or something.

tests/unit/lambda_debug_mode/test_config_parsing.py
```diff
@@ -0,0 +1,121 @@
...
+functions:
+  arn:aws:lambda:eu-central-1:000000000000:function:functionname:
+    debug-port: null
+    enforce-timeout: null
```
dfangl, line 30: The flag is now called `enforce-timeouts`, not `enforce-timeout`. So we are not testing the `null` value here, but actually the default, as pydantic per default ignores extra fields.

dfangl (approved):
All in all, this is a great step towards improving DevX! I just had some minor nits, but nothing major!

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode.py
MEPalma: changed to `is_lambda_debug_timeout_enforced_for` so it's more consistent with the config flag `enforce_timeout`

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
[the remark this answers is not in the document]
MEPalma: I would choose to keep the logic of this function I have added. This logic should add the qualifier only to arns that contain all other parts. Although we don't parse or validate each part, for the majority of cases, we'd be adding or not a qualifier to correct arns. Unfortunately, `function_locators_from_arn` is not strict enough with identifying the parts. for example for `arn:aws:lambda:eu-central-1:000000000000:function` it selects `function` as the name, and null as the qualifier. Not only is it not correct (the function name is not actually given), but it also means that this logic would continue to produce `arn:aws:lambda:eu-central-1:000000000000:function:$LATEST`. Which seems a bit unfortunate. `_to_qualified_lambda_function_arn` would instead raise an UnknownLambdaArnFormat in this case, which would help the user getting the arn fixed (adding the name and/or the qualifier). `function_locators_from_arn` would is also too compliant with arns ending with empty qualifiers `arn:aws:lambda:eu-central-1:000000000000:function:functionname:`. Please do correct me if I have missing something! Otherwise, I think `_to_qualified_lambda_function_arn` is specific enough to this workflow that can be left as is. At the same time, I am not sure of the consequences of adding stricter checks to `function_locators_from_arn`.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+        return lambda_arn
+    lambda_arn_parts = lambda_arn.split(":")
+    lambda_arn_parts_len = len(lambda_arn_parts)
+
+    # The arn is qualified and with a non-empy qualifier.
+    is_qualified = lambda_arn_parts_len == 8
+    if is_qualified and lambda_arn_parts[-1]:
+        return lambda_arn
+
+    # The arn is not unqualified, but probably erroneous, pass the value upstream.
```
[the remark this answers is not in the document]
MEPalma: updated logic should be clearer to deduce now. The function attempts to annotate with a qualifier arns that are otherwise well structured with a non empty qualifier if one is given.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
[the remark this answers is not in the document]
gregfurman: Just to clarify, it's actually AWS that interprets
```
arn:<partition>:lambda:<region>:≤account-id>:function
```
as 
```
arn:<partition>:lambda:<region>:<account-id>:function:function
```

So the `function_locators_from_arn` is keeping parity 🫠 

I think what you're saying is fair, and I given how much effort the additional validations required, I am also quite pro tackling this via string splitting over regex (where possible).

With that said, would you consider adding the above to `api_utils`? There are a couple places where I believe it could be useful in the provider.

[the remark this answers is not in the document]
MEPalma: Thanks for the input! I'll reviews whether that "naming option" is something we should also support in this config file

PR #11366, opened 2024-08-16
Lambda: improve function name and qualifier validation

gregfurman:
## Motivation
Current validation checks for Lambda function names did not account for some invalid regex combinations as well as length limits for `FunctionName` differing between request types. This will add tests ensuring a wider range of invalid `FunctionName` and `Qualfier` combinations are caught.

## Changes
The Lambda Provider:
- Remove validation checks from `create_function` in favour of `api_utils.get_name_and_qualifier`
- Use improved qualifier validation and better account for invalid qualifier edge-case for `DeleteFunction`

The API utils `api_utils.py`:
- Conduct name validations in the `get_name_and_qualifier` -- which is ubiquitously used across the Lambda provider and should enact these checks.
- Account for operation type when running `validate_function_name` (accounting specifically for `CreateFunction`, `DeleteFunction`, `Invoke`, and `GetFunction`) with sensible defaults being used elsewhere.
- Add new `validate_qualifier` and `construct_validation_exception_message`

The Lambda API tests:
- Added new test `test_function_name_and_qualifier_validation` which accounts for a wide range of possible exception cases when attempting validation. 

## Testing

The below is a test table showing all scenarios tested for `CreateFunction`, `DeleteFunction` `GetFunction` and `Invoke` test types.

Some considerations:
- The "Negative Version Number" and "Incomplete ARN" test skips the `create_function` request.
- The "Function Name Too Long" test will only fail validation for `create_function` request -- with `ResourceNotFoundException` being thrown for othe types. None are skipped to ensure parity with AWS.

| Name | Function Name | Qualifier | Skipped Requests |
|------|---------------|-----------|------------------|
| Invalid Function Name | `my-function!` | N/A | None |
| Invalid Qualifier | `my-function` | `invalid!` | None |
| Invalid Account ID | `invalid-account:function:my-function` | N/A | None |
| Invalid Region in ARN | `arn:aws:lambda:invalid-region:{account_id}:function:my-function` | N/A | None |
| Non-Lambda ARN | `arn:aws:ec2:{region_name}:{account_id}:instance:i-1234567890abcdef0` | N/A | None |
| Function Name Too Long | `"a" * max_length` | N/A | None |
| Long Name, Invalid Region | `arn:aws:lambda:invalid-region:{account_id}:function:my-function["a" * max_length]` | N/A | None |
| Long ARN & Qualifier, Invalid Region | `arn:aws:lambda:invalid-region:{account_id}:function:my-function-["a" * max_length]` | `"a" * max_length` | None |
| Long Qualifier | `my-function` | `"a" * max_length` | None |
| Multiple Qualifiers in ARN | `arn:aws:lambda:{region_name}:{account_id}:function:my-function:1:2` | N/A | None |
| Negative Version Number | `my-function` | `-1` | `create_function` |
| Incomplete ARN | `arn:aws:lambda:{region_name}:{account_id}:function` | N/A | `create_function` |
| Partial ARN, Extra Qualifier | `function:my-function:$LATEST:extra` | N/A | None |
| Latest Ver, Add'l Qualifier | `arn:aws:lambda:{region_name}:{account_id}:function:my-function:$LATEST` | `1` | None |
| Lowercase Latest Qualifier | `arn:aws:lambda:{region_name}:{account_id}:function:my-function` | `$latest` | None |
| Missing Region in ARN | `arn:aws:lambda::{account_id}:function:my-function` | N/A | None |
| Missing Account ID in ARN | `arn:aws:lambda:{region_name}::function:my-function` | N/A | None |
| Misspelled Latest in ARN | `arn:aws:lambda:{region_name}:{account_id}:function:my-function:$LATES` | N/A | None |

#### LocalStack Pro: IAM Policies Enforced
- `ext`: Partial run against tests in `aws/tests/lambda_` and with `ENFORCE_IAM` enabled at [GitHub workflow](https://github.com/localstack/localstack-ext/actions/runs/10457344427)
- `ENV: ENFORCE_IAM=1`
#### LocalStack Community: Multi-Account and Multi-Region
- `multi`: Multi-region/account build at [CircleCI](https://app.circleci.com/pipelines/github/localstack/localstack/27405) pipeline.

## TODO

What's left to do:

- [x] Ensure a multi-account and multi-region test can run for this PR
- [ ] Runs against LocalStack Pro version with IAM enforcement enabled

## Potential future works
- Extend the operations being tested beyond `CreateFunction`, `DeleteFunction`, `GetFunction`, and `Invoke`
- Look into improving test-coverage and support for multi-account and multi-region operations for the provider.

### DIFF
--- a/localstack-core/localstack/services/lambda_/api_utils.py
+++ b/localstack-core/localstack/services/lambda_/api_utils.py
@@ -8,7 +8,7 @@ import re
 import string
 from typing import TYPE_CHECKING, Any, Optional, Tuple
 
-from localstack.aws.api import RequestContext
+from localstack.aws.api import CommonServiceException, RequestContext
 from localstack.aws.api import lambda_ as api_spec
 from localstack.aws.api.lambda_ import (
     AliasConfiguration,
@@ -27,6 +27,7 @@ from localstack.aws.api.lambda_ import (
     TracingConfig,
     VpcConfigResponse,
 )
+from localstack.services.lambda_.invocation import AccessDeniedException
 from localstack.services.lambda_.runtimes import ALL_RUNTIMES, VALID_LAYER_RUNTIMES, VALID_RUNTIMES
 from localstack.utils.aws.arns import ARN_PARTITION_REGEX, get_partition
 from localstack.utils.collections import merge_recursive
@@ -48,7 +49,7 @@ FULL_FN_ARN_PATTERN = re.compile(
     rf"{ARN_PARTITION_REGEX}:lambda:(?P<region_name>[^:]+):(?P<account_id>\d{{12}}):function:(?P<function_name>[^:]+)(:(?P<qualifier>.*))?$"
 )
 
-# Pattern for a full (both with and without qualifier) lambda function ARN
+# Pattern for a full (both with and without qualifier) lambda layer ARN
 LAYER_VERSION_ARN_PATTERN = re.compile(
     rf"{ARN_PARTITION_REGEX}:lambda:(?P<region_name>[^:]+):(?P<account_id>\d{{12}}):layer:(?P<layer_name>[^:]+)(:(?P<layer_version>\d+))?$"
 )
@@ -101,6 +102,65 @@ LAMBDA_DATE_FORMAT = "%Y-%m-%dT%H:%M:%S.%f+0000"
 
 # An unordered list of all Lambda CPU architectures supported by LocalStack.
 ARCHITECTURES = [Architecture.arm64, Architecture.x86_64]
+
+# ARN pattern returned in validation exception messages.
+# Some excpetions from AWS return a '\.' in the function name regex
+# pattern therefore we can sub this value in when appropriate.
+ARN_NAME_PATTERN_VALIDATION_TEMPLATE = "(arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{{2}}((-gov)|(-iso([a-z]?)))?-[a-z]+-\\d{{1}}:)?(\\d{{12}}:)?(function:)?([a-zA-Z0-9-_{0}]+)(:(\\$LATEST|[a-zA-Z0-9-_]+))?"
+
+
+def validate_function_name(function_name_or_arn: str, operation_type: str):
+    function_name, *_ = function_locators_from_arn(function_name_or_arn)
+    arn_name_pattern = ARN_NAME_PATTERN_VALIDATION_TEMPLATE.format("")
+    max_length = 170
+
+    match operation_type:
+        case "GetFunction" | "Invoke":
+            arn_name_pattern = ARN_NAME_PATTERN_VALIDATION_TEMPLATE.format(r"\.")
+        case "CreateFunction" if function_name == function_name_or_arn:  # only a function name
+            max_length = 64
+        case "CreateFunction" | "DeleteFunction":
+            max_length = 140
+
+    validations = []
+    if len(function_name_or_arn) > max_length:
+        constraint = f"Member must have length less than or equal to {max_length}"
+        validation_msg = f"Value '{function_name_or_arn}' at 'functionName' failed to satisfy constraint: {constraint}"
+        validations.append(validation_msg)
+
+    if not AWS_FUNCTION_NAME_REGEX.match(function_name_or_arn) or not function_name:
+        constraint = f"Member must satisfy regular expression pattern: {arn_name_pattern}"
+        validation_msg = f"Value '{function_name_or_arn}' at 'functionName' failed to satisfy constraint: {constraint}"
+        validations.append(validation_msg)
+
+    return validations
+
+
+def validate_qualifier(qualifier: str):
+    validations = []
+
+    if len(qualifier) > 128:
+        constraint = "Member must have length less than or equal to 128"
+        validation_msg = (
+            f"Value '{qualifier}' at 'qualifier' failed to satisfy constraint: {constraint}"
+        )
+        validations.append(validation_msg)
+
+    if not QUALIFIER_REGEX.match(qualifier):
+        constraint = "Member must satisfy regular expression pattern: (|[a-zA-Z0-9$_-]+)"
+        validation_msg = (
+            f"Value '{qualifier}' at 'qualifier' failed to satisfy constraint: {constraint}"
+        )
+        validations.append(validation_msg)
+
+    return validations
+
+
+def construct_validation_exception_message(validation_errors):
+    if validation_errors:
+        return f"{len(validation_errors)} validation error{'s' if len(validation_errors) > 1 else ''} detected: {'; '.join(validation_errors)}"
+
+    return None
 
 
 def map_function_url_config(model: "FunctionUrlConfig") -> api_spec.FunctionUrlConfig:
@@ -185,14 +245,22 @@ def get_function_name(function_arn_or_name: str, context: RequestContext) -> str
     return name
 
 
-def function_locators_from_arn(arn: str) -> tuple[str, str | None, str | None, str | None]:
+def function_locators_from_arn(arn: str) -> tuple[str | None, str | None, str | None, str | None]:
     """
     Takes a full or partial arn, or a name
 
     :param arn: Given arn (or name)
     :return: tuple with (name, qualifier, account, region). Qualifier and region are none if missing
     """
-    return FUNCTION_NAME_REGEX.match(arn).group("name", "qualifier", "account", "region")
+
+    if matched := FUNCTION_NAME_REGEX.match(arn):
+        name = matched.group("name")
+        qualifier = matched.group("qualifier")
+        account = matched.group("account")
+        region = matched.group("region")
+        return (name, qualifier, account, region)
+
+    return None, None, None, None
 
 
 def get_account_and_region(function_arn_or_name: str, context: RequestContext) -> Tuple[str, str]:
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/lambda_/api_utils.py
```diff
@@ -210,25 +278,57 @@ def get_name_and_qualifier(
...
+    # TODO: should this only run if operation type is unsupported?
+    if region and region != context.region:
+        raise ResourceNotFoundException(
+            f"Functions from '{region}' are not reachable in this region ('{context.region}')",
+            Type="User",
+        )
+
+    validation_errors = []
+    if function_name:
+        validation_errors.extend(validate_function_name(function_arn_or_name, operation_type))
```
dfangl, line 308: I think we are still missing cases here. If the function name is `None`, which is possible when you specify an completely invalid value for it, for example `*`, the checks never catch it, and it continues with the operation, leading to errors down the line. AWS validation catches that.

localstack-core/localstack/services/lambda_/api_utils.py
```diff
@@ -185,14 +246,21 @@ def get_function_name(function_arn_or_name: str, context: RequestContext) -> str
...
     """
-    return FUNCTION_NAME_REGEX.match(arn).group("name", "qualifier", "account", "region")
+
+    def _extract_named_groups(name, qualifier, account, region):
+        return name, qualifier, account, region
+
+    if matched := FUNCTION_NAME_REGEX.match(arn):
+        return _extract_named_groups(**matched.groupdict())
+
+    return 4 * (None,)
```
dfangl, line 263: Why do we need the internal function for this? Is there an advantage for this to using `matched.group(names...)`?

I would also prefer `None, None, None, None` over `4 * (None, )`, as I think it is more readable and clearer on the intent.

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -760,6 +760,191 @@ def test_function_arns(
...
+            },
+            {
+                "args": {
+                    "FunctionName": f"arn:aws:lambda:{region_name}:{account_id}:function:my-function:$LATES"
+                },
+                "SnapshotName": "misspelled_latest_in_arn",
+            },
+        ]
+
+        for test in tests:
```
dfangl, line 934: Could we also extract this into parametrization?

tests/aws/services/lambda_/test_lambda_api.validation.json
```diff
@@ -57,11 +57,26 @@
...
   "tests/aws/services/lambda_/test_lambda_api.py::TestLambdaFunction::test_function_arns": {
-    "last_validated_date": "2024-04-10T09:00:24+00:00"
+    "last_validated_date": "2024-08-13T13:09:05+00:00"
   },
   "tests/aws/services/lambda_/test_lambda_api.py::TestLambdaFunction::test_function_lifecycle": {
     "last_validated_date": "2024-04-10T08:59:22+00:00"
   },
+  "tests/aws/services/lambda_/test_lambda_api.py::TestLambdaFunction::test_function_name_and_qualifier_validation": {
+    "last_validated_date": "2024-08-14T15:25:00+00:00"
+  },
```
dfangl, line 67: Would be nice to clean this up before merging, I think it is a leftover :)

dfangl (changes requested):
Great first iteration! I found some issues with the validation, which would be nice to address before merging :)

tests/aws/services/lambda_/test_lambda_api.py
gregfurman: Yeah sure. I was concerned about the test snapshot naming being quite ugly though since the params are added to the snapshot naming, which is why I kept the test-table internal as opposed to parametrizing it.

Think it's a good suggestion though

dfangl: Take a look at the `ids` parameter of the parametrization, you could use the snapshot name in there, which should alleviate this concern.

localstack-core/localstack/services/lambda_/api_utils.py
gregfurman: > Why do we need the internal function for this? 

Previously I was doing qualifier and name validation within that internal function. It made it a lot cleaner since we could just unpack the dict args to that internal func and easily validate individual members. Since validation takes place elsewhere I agree this is unnecessary.

i.e 
```python
    def _extract_named_groups(name, qualifier, account, region):
        validate_name_and_qualifier(name, qualifier) # This would raise an exception
        return name, qualifier, account, region

    if matched := FUNCTION_NAME_REGEX.match(arn):
        return _extract_named_groups(**matched.groupdict())
```

> I would also prefer `None, None, None, None`

Fair. I think the current approach takes "pythonic" to the extreme lol. Will change.

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1290,6 +1259,13 @@ def delete_function(
         function_name, qualifier = api_utils.get_name_and_qualifier(
             function_name, qualifier, context
         )
+
+        if qualifier and qualifier.startswith("-"):
```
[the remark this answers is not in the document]
gregfurman: > Qualifiers do not need to start with a "-".

Precisely! If a qualifier/alias does start with a `-` then AWS interprets this as a user trying to delete multiple aliases and throws an error. See:

https://github.com/localstack/localstack/blob/2979749c47205e000aa3315e6a9ea9c0980f1ad2/localstack-core/localstack/services/lambda_/provider.py#L1263-L1267


> We should check if the qualifier is an alias, and then not allow it.

The `get_name_and_qualifier` checks this and assigns to the `qualifier` variable a qualifier derived from an ARN/function-name OR a qualifier passed in as a standalone parameter: 

See below how `qualifier` is re-assigned:
https://github.com/localstack/localstack/blob/2979749c47205e000aa3315e6a9ea9c0980f1ad2/localstack-core/localstack/services/lambda_/provider.py#L1259-L1261

See AWS and LocalStack output (with the prefix check):
```
$ aws lambda delete-function --function-name "hello-world" --qualifier -1
An error occurred (InvalidParameterValueException) when calling the DeleteFunction operation: Deletion of aliases is not currently supported.

$ aws lambda delete-function --function-name "hello-world:-1"               
An error occurred (InvalidParameterValueException) when calling the DeleteFunction operation: Deletion of aliases is not currently supported.

$ awslocal lambda delete-function --function-name "hello-world:-1"
An error occurred (InvalidParameterValueException) when calling the DeleteFunction operation: Deletion of aliases is not currently supported.

$ awslocal lambda delete-function --function-name "hello-world" --qualifier -1
An error occurred (InvalidParameterValueException) when calling the DeleteFunction operation: Deletion of aliases is not currently supported.
```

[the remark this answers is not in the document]
dfangl: I am aware of the qualfier checks and assesment, but this is not what I am talking about:

You cannot run `delete-function` with a qualifier pointing to an alias, only to a version:
```
$ aws lambda delete-function --function-name "hello-world" --qualifier test
An error occurred (InvalidParameterValueException) when calling the DeleteFunction operation: Deletion of aliases is not currently supported.
```

It does not matter if the alias starts with an `-`, and it does not interpret it as multiple aliases. you just cannot delete a function version by alias.

What is indeed valid, is something like this:

```
$ aws lambda delete-function --function-name "hello-world" --qualifier 1
```

This would delete only function version `1` of the `hello-world` function - not the entire function. However, the qualifier cannot be an alias, and an alias does not have to start with a dash. Instead of checking if it starts with a dash, you should use `api_utils.qualifier_is_alias` for the check if the qualifier is an alias.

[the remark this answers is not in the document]
gregfurman: A-ha! I see what you're saying. That makes sense.

localstack-core/localstack/services/lambda_/api_utils.py
gregfurman: Removed that `None` check and added a case for this

tests/aws/services/lambda_/test_lambda_api.py
gregfurman: Defined these in pytest fixtures 😄

tests/aws/services/lambda_/test_lambda_api.validation.json
gregfurman: Removed

localstack-core/localstack/services/lambda_/provider.py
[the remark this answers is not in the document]
dfangl: Could we get this change in as well, before I review? :)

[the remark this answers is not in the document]
gregfurman: Done! Thinking we can also probably get rid of some of these `TODOs` now above `delete_function`:


- TODO: does deleting the latest published version affect the next versions number?
- ~TODO: what happens when we call this with a qualifier and a fully qualified ARN? (+ conflicts?)~
- ~TODO: test different ARN patterns (shorthand ARN?)~
- TODO: test deleting across regions?
- TODO: test mismatch between context region and region in ARN
- ~TODO: test qualifier $LATEST, alias-name and version~

dfangl (approved):
LGTM now! Thanks for adding those changes!

FYI: We squash commits anyway on merging, so it is not necessary to force push for changes, and keep the commit count low. Personally, I find it is easier, especially for review, if you can see the changes done after a review in separate commits. Others might disagree with this, though :sweat_smile:

PR #11370, opened 2024-08-16
add support for ArchivePolicy in AWS::SNS::Topic

### LINKED ISSUE #11289: bug: Cloudformation validating ArchivePolicy property type from AWS::SNS::Topic wrong

### Current Behavior

When trying to create and deploy a SNS topic via CloudFormation where we configure `ArchivePolicy` property values (i.e. set `MessageRetentionPeriod`, see **Steps to Reproduce** for more details) it fails validation because it is expecting a string rather than a json like value (or equivalent in yaml).

Here is an example of the error when using LocalStack via the latest Docker image:
```
WARN --- [-functhread5] l.s.c.engine.entities      : CFn resource failed to deploy: SNS (Parameter validation failed:; Invalid type for parameter Attributes.ArchivePolicy, value: {'MessageRetentionPeriod': 30}, type: <class 'dict'>, valid types: <class 'str'>)
```

This works as expected when deploying to AWS.

### Expected Behavior

I would expect `ArchivePolicy` property to allow json like values as per [AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sns-topic.html)

With a docker-compose file

### Steps To Reproduce

Using the following files:

`template-example.yaml`
```
AWSTemplateFormatVersion: 2010-09-09

Resources:
  SNS:
    Type: AWS::SNS::Topic
    Properties:
      TopicName: example-topic
      FifoTopic: true
      ArchivePolicy:
        MessageRetentionPeriod: 30
```

`init-localstack.sh`
```
#!/bin/bash

set -e

awslocal cloudformation deploy \
    --template-file /template-example.yaml \
    --stack-name example-sns-stack \
    --region=eu-west-1
```

`docker-compose.yml`
```
version: '3.9'
services:
  localstack:
    image: localstack/localstack:latest
    environment:
      SERVICES: cloudformation,sns
      LS_LOG: info
    volumes:
      - "./init-localstack.sh:/etc/localstack/init/ready.d/init-localstack.sh"
      - "./template-example.yaml:/template-example.yaml"
```
And running `docker compose up localstack` you should see the error mentioned above.

### Environment

```markdown
- OS: macOS Sonoma 14.3.1
- LocalStack:
LocalStack version: 3.6.1.dev (latest)
LocalStack build date: 2024-07-30
LocalStack build git hash: 85b16b2d8
```

### Anything else?

_No response_

pinzon:
## Motivation
With this PR now an user can define an ArchivePolicy attribute without the stack deployment crashing. Fixes #11289

## Changes
- Resource provider now converts the dict into a json string as the service requires
-  The service provider now returns the attribute if it exists in the model
- A bit of refactoring in the Create Method op.

## Testing
- new cdk test

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -163,6 +163,8 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
                 moto_response["Attributes"][key] = getattr(moto_topic_model, attr)
             elif attr == "signature_version":
                 moto_response["Attributes"]["SignatureVersion"] = moto_topic_model.signature_version
+            elif attr == "archive_policy":
+                moto_response["Attributes"]["ArchivePolicy"] = moto_topic_model.archive_policy
         return moto_response
 
     def publish_batch(
--- a/localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
+++ b/localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
@@ -102,7 +102,7 @@ class SNSTopicProvider(ResourceProvider[SNSTopicProperties]):
         model["TopicArn"] = create_sns_response["TopicArn"]
 
         # now we add subscriptions if they exists
-        for subscription in model.get("Subscriptions", []):
+        for subscription in model.get("Subscription", []):
             sns.subscribe(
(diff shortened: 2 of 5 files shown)


bentsku (approved):
LGTM! Thanks for also fixing the SNS provider 💯

PR #11372, opened 2024-08-16
Apigw ng refactor deployment store

cloutierMat:
## Motivation

As discussed with @bentsku, we are changing the structure of internals and actives deployment dict in the store. This will facilitate access to the resources when only the api id is known. Notably when entirely deleting a rest api.

## Changes

Also fixed in this PR. When deleting a rest api, we were not deleting it's deployment from the store, leading to memory leaks.

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -78,8 +78,8 @@ class ApiGatewayEndpoint:
         self, context: RestApiInvocationContext, api_id: str, stage: str
     ):
         try:
-            deployment_id = self._global_store.active_deployments[(api_id, stage)]
-            frozen_deployment = self._global_store.internal_deployments[(api_id, deployment_id)]
+            deployment_id = self._global_store.active_deployments[api_id][stage]
+            frozen_deployment = self._global_store.internal_deployments[api_id][deployment_id]
 
         except KeyError:
             # TODO: find proper error when trying to hit an API with no deployment/stage linked
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -72,7 +72,7 @@ class ApiGatewayEndpoint:
         return context, response
 
     def is_rest_api(self, api_id: str, stage: str):
-        return (api_id, stage) in self._global_store.active_deployments
+        return (api := self._global_store.active_deployments[api_id]) and stage in api
 
     def populate_rest_api_invocation_context(
         self, context: RestApiInvocationContext, api_id: str, stage: str
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -72,7 +72,7 @@ class ApiGatewayEndpoint:
         return context, response
 
     def is_rest_api(self, api_id: str, stage: str):
-        return (api := self._global_store.active_deployments[api_id]) and stage in api
+        return stage in self._global_store.active_deployments.get(api_id)
 
     def populate_rest_api_invocation_context(
         self, context: RestApiInvocationContext, api_id: str, stage: str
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -72,14 +72,14 @@ def prepare_rest_api_invocation(
         return context, response
 
     def is_rest_api(self, api_id: str, stage: str):
-        return (api_id, stage) in self._global_store.active_deployments
+        return (api := self._global_store.active_deployments[api_id]) and stage in api
```
bentsku, line 75: I think this will raise a KeyError if the API does not exist 
```suggestion
        return stage in self._global_store.active_deployments.get(api_id, {})
```

bentsku (commented):
Just one comment that needs to be addressed, but then we can be good to merge 😄

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
cloutierMat: It shouldn't raise an error. Given that it was created as a defaultdict:
```python
active_deployments: dict[str, dict[str, str]] = CrossAccountAttribute(lambda: defaultdict(dict))
```
Both would return the same value. We could however simplify my code to:
```suggestion
        return stage in self._global_store.active_deployments[api_id]
```

cloutierMat: If you feel we shouldn't use a `defaultdict` in in this case, I am open to the conversation.

My reasoning for using one in the first place is the following
- It simplifies and improves resilience for  all accesses to the nested dict. `self._global_store.internal_deployments[api_id][deployment_id]` vs `self._global_store.internal_deployments.get(api_id, {}][deployment_id]`
- `KeyError` will still be raised if the deployment doesn't not exists.
- `active_deployment` should not be regarded as a source of truth with regard to the existence of an api. So having all apis defaulting to an empty dict should not have adverse effect. :thinking: 

Once again, happy to discuss this further if you have a different point of view :smile:

bentsku: Right yes, I had forgotten about this!

I think the `defaultdict` is fine, it simplifies our job not having to use `setdefault` when creating deployments/stages for the API id level of the dict.

I guess my point is very nit-picky and niche, but accessing the defaultdict with `[]` will create an entry in the dict, where `.get()` doesn't. This `internal_deployments` will be stored with persistence and cloud pods, so we will store empty dicts if the API doesn't exist, and if we want to list API with deployments, we will get those non-existing keys. 

I think using defaultdict to automatically create the first level in the provider is good, but using it this way here might be a bit tricky, because it doesn't raise `KeyError`, but it still creates an entry in the dict. 

Maybe a bit of a mix, something like `(api := self._global_store.active_deployments.get(api_id)) and stage in api` can also work, but using the brackets when we're not sure it exists and can be accessed outside the provider feels a bit weird to me. But like I said, this is really, really nit-picky 😄

bentsku: I think the defaultdict mostly simplifies this for us in the provider:
```python
# creating a new deployment
_global_store.internal_deployments.setdefault(api_id, {})[deployment_id] = deployment

# vs
_global_store.internal_deployments[api_id][deployment_id] = deployment
```

cloutierMat: I didn't think of that effect on the store. :thinking:. It is a very valid point. Looks like the `defaultdict` isn't the right solution here, after all. It would keep that "danger" in the future as well.

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -72,14 +72,14 @@ def prepare_rest_api_invocation(
         return context, response
 
     def is_rest_api(self, api_id: str, stage: str):
-        return (api_id, stage) in self._global_store.active_deployments
+        return stage in self._global_store.active_deployments.get(api_id)
```
bentsku, line 75: sorry again, this will raise `TypeError: argument of type 'NoneType' is not iterable`, sorry for the back and forth on this 😅

cloutierMat: :man_facepalming: This really is our day :rofl:

bentsku (approved):
LGTM! Thanks for taking care of this, really good changes! This is going to make our life easier 🙏

PR #11375, opened 2024-08-19
docs: Add workaround for amazon-ion issue in contributing docs

gregfurman:
## Motivation
- Workaround for issue encountered when compiling `amazon-ion==0.9.3` dependency using `gcc/cmake`.
- Updates to the C compiler now interpret some warnings as errors, so the dependency wheel will fail to build. In addition, there is no Mac OSX ARM64 build wheel available for this version on PyPi.

## Changes
- Added changes to `CONTRIBUTING.md` with workaround steps. 
- A user can either suppress the C warnings/errors OR remove `cmake` from their `PATH` variable causing pip to build the wheel using pure Python instead of C.

### DIFF
--- a/docs/development-environment-setup/README.md
+++ b/docs/development-environment-setup/README.md
@@ -58,6 +58,10 @@ Refer to our official [Dockerfile](https://github.com/localstack/localstack/blob
 * [JPype1](https://pypi.org/project/JPype1/) might require `g++` to fix a compile error on ARM Linux `gcc: fatal error: cannot execute ‘cc1plus’`
   * Used in EventBridge, EventBridge Pipes, and Lambda Event Source Mapping for a Java-based event ruler via the opt-in configuration `EVENT_RULE_ENGINE=java`
   * Introduced in [#10615](https://github.com/localstack/localstack/pull/10615)
+* [Amazon Ion Python](https://github.com/amazon-ion/ion-python) might fail to compile on more recent versions of `gcc` where the diagnostics [`incompatible-pointer-types`](https://gcc.gnu.org/pipermail/gcc-cvs/2023-December/394351.html) and [`int-conversion`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106416) were promoted from warnings to errors. This is more likely to affect Mac OSX ARM users since there are no build wheels available for [`amazon-ion==0.9.3`](https://pypi.org/project/amazon.ion/0.9.3/#files).
+  * Used by the Python driver for Amazon QLDB `pyqldb` and pinned to `amazon-ion==0.9.3` since later versions of `amazon-ion` have introduced [breaking changes](https://github.com/awslabs/amazon-qldb-driver-python/issues/164).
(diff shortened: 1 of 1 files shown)


HarshCasper (approved):

docs/development-environment-setup/README.md
```diff
@@ -58,6 +58,10 @@ Refer to our official [Dockerfile](https://github.com/localstack/localstack/blob
 * [JPype1](https://pypi.org/project/JPype1/) might require `g++` to fix a compile error on ARM Linux `gcc: fatal error: cannot execute ‘cc1plus’`
   * Used in EventBridge, EventBridge Pipes, and Lambda Event Source Mapping for a Java-based event ruler via the opt-in configuration `EVENT_RULE_ENGINE=java`
   * Introduced in [#10615](https://github.com/localstack/localstack/pull/10615)
+* [Amazon Ion Python](https://github.com/amazon-ion/ion-python) might fail to compile on more recent versions of `gcc` where the diagnostics [`incompatible-pointer-types`](https://gcc.gnu.org/pipermail/gcc-cvs/2023-December/394351.html) and [`int-conversion`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106416) were promoted from warnings to errors. This is more likely to affect Mac OSX ARM users since there are no build wheels available for [`amazon-ion==0.9.3`](https://pypi.org/project/amazon.ion/0.9.3/#files).
+  * Used by the Python driver for Amazon QLDB `pyqldb` and pinned to `amazon-ion==0.9.3` since later versions of `amazon-ion` have introduced [breaking changes](https://github.com/awslabs/amazon-qldb-driver-python/issues/164).
```
alexrashed, line 62: I think it's a bit counter-productive to mention exact versions here. Is this issue really only interesting for version `0.9.3`? If not, I would just mention the package and not the version (since this would be outdated as soon as `pyqldb` removes the dependency limit).

alexrashed (changes requested):
Thanks a lot for taking the time to adjust our contribution guides!
However, I don't think this is the right place.
`amazon-ion` is not a dependency of our community project, but in fact of a downstream-project. I think this should be added to the docs in the specific repo...

docs/development-environment-setup/README.md
gregfurman: > I think it's a bit counter-productive to mention exact versions here

Happy to make this doc change less specific. 

> However, I don't think this is the right place.

I agree it does feel odd to provide a workaround for a downstream issue However... this has now blocked two engineers when trying to setup local dev environments and I'm concerned not addressing it could continue to block new-starters and potential contributors.

Maybe a happier middle ground would be to extend the `Makefile` to include this workaround?

alexrashed: > I agree it does feel odd to provide a workaround for a downstream issue However... this has now blocked two engineers when trying to setup local dev environments and I'm concerned not addressing it could continue to block new-starters and potential contributors.

I fully agree that this should be documented, I'm just saying this is not the right place. I think this should be documented in the repo where it actually happens. 😉

gregfurman:
Closing since this is the incorrect repo 🫠

PR #11382, opened 2024-08-20
Refactor/Events: add localstack_only test markers for v1 only tests

maxhoheiser:
## Motivation
Some tests are only relevant for the v1 provider, since they are covered in separate tests or the functionality is not supported in v2 provider. 
They will be removed once the v1 provider is removed.
Some of these tests are not validated and it is not worth the effort to validate them, to clearly indicate that they are not to be validated the marker `localstack_only` with an explanation is added.

### DIFF
--- a/tests/aws/services/events/test_events.py
+++ b/tests/aws/services/events/test_events.py
@@ -101,8 +101,7 @@ class TestEvents:
         response = aws_client.events.put_events(Entries=entries)
         snapshot.match("put-events", response)
 
-    @markers.aws.only_localstack
-    # tests for legacy v1 provider delete once v1 provider is removed, v2 covered in separate tests
+    @markers.aws.validated
     @pytest.mark.skipif(
         is_old_provider(),
         reason="V1 provider does not support this feature",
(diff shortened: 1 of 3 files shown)


tests/aws/services/events/test_events.py
```diff
@@ -101,7 +101,8 @@ def test_put_events_without_source(self, snapshot, aws_client):
         response = aws_client.events.put_events(Entries=entries)
         snapshot.match("put-events", response)
 
-    @markers.aws.unknown
+    @markers.aws.only_localstack
+    # tests for legacy v1 provider delete once v1 provider is removed, v2 covered in separate tests
     @pytest.mark.skipif(
         is_old_provider(),
```
joe4dev, line 107: Why do we skip this test for the old provider if it is only intended for it?

joe4dev (approved):
LGTM after clarifying the contradiction for `test_put_event_without_detail`

tests/aws/services/events/test_events.py
maxhoheiser: correct, I validated the test

PR #11385, opened 2024-08-20
Fix SQS handling of empty Redrive Policy attribute

gregfurman:
## Motivation
Handle the case where setting a `RedrivePolicy` attribute to an empty string removes the `RedrivePolicy` from an SQS queue's attribute map. This should be identical to how LocalStack handles an empty `Policy` attribute.

Currently, this behavior is causing Terraform deletions to initially fail. Terraform attempts to remove a queue's `aws_sqs_queue_redrive_policy` resource by setting the `RedrivePolicy` attribute value to an empty string. However, LocalStack was not removing this empty value, leading to the failure.

## Changes
- Setting an SQS queue's `RedrivePolicy` to an empty string now deletes it from the internal `QueueAttributeMap`
- Added a test for this case at `TestSqsProvider::test_set_empty_redrive_policy`

## Testing
- Validated with Terraform & CLI against LocalStack and AWS

### DIFF
--- a/localstack-core/localstack/services/sqs/provider.py
+++ b/localstack-core/localstack/services/sqs/provider.py
@@ -1349,6 +1349,10 @@ class SqsProvider(SqsApi, ServiceLifecycleHook):
             del queue.attributes[QueueAttributeName.Policy]
 
         redrive_policy = queue.attributes.get(QueueAttributeName.RedrivePolicy)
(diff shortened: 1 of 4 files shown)


thrau (approved):
nice catch! welcome to the sqs contributors :grin: 

LGTM!

PR #11398, opened 2024-08-21
[Lambda DevX] Control Lambda Container Startup Timeout when In Lambda Debug Mode

MEPalma:
\* Depends on https://github.com/localstack/localstack/pull/11363/

## Motivation
Currently, when a Lambda function is executed in debug mode, the system increases the evaluation timeout, which works well if the debugger connects after the function handler is invoked. However, this approach fails in scenarios where the debugger needs to connect during the container's startup phase, specifically when the debugger connection code resides outside the Lambda function handler.

To address this limitation, these changes propose to automatically extend the Lambda container startup timeout when the system detects it is running in Lambda Debug Mode, and the target lambda function is being debugged on some remote port.

## Changes
- Add control logic for startup timeout value


localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -402,3 +403,11 @@ def _get_execution_timeout_seconds(self) -> int:
         if is_lambda_debug_timeout_for(self.function_version.qualified_arn):
             return DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS
         return self.function_version.config.timeout
+
+    def _get_startup_timeout_seconds(self) -> int:
+        # Returns the timeout value in seconds to be enforced during lambda container startups.
+        # This is the value defined through LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT or the LAMBDA
+        # DEBUG MODE default if this is enabled.
+        if is_lambda_debug_timeout_for(self.function_version.qualified_arn):
```
joe4dev, line 411: I think we should create a follow-up docs PR and list the implications of the Lambda debug mode.
We could also add a debug log here to make the config adjustment traceable at runtime (similar for other potentially implicit config updates) 🤔 Do you think logging is helpful or verbose clutter @dfangl @dominikschubert?

joe4dev (approved):
Thanks for considering the startup timeout as well 👍 (I just looked at the last commit)

Kudos to @dfangl for pointing it out.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_config.py
```diff
@@ -0,0 +1,198 @@
...
+from yaml import Loader, MappingNode, MarkedYAMLError, SafeLoader
+
+from localstack.aws.api.lambda_ import Arn
+
+LOG = logging.getLogger(__name__)
+
+
+class LambdaDebugConfig(BaseModel):
+    debug_port: Optional[int] = Field(None, alias="debug-port")
+    timeout_disable: bool = Field(False, alias="timeout-disable")
```
joe4dev, line 17: I'm wondering about the naming: Would `enforce_timeouts: True|False` be clearer? What's your naming suggestion @dfangl @dominikschubert ?

Current behavior:
* `LAMBDA_DEBUG_MODE=0` -> never disable any timeout
* `LAMBDA_DEBUG_MODE=1`
  * `timeout-disable` not set or no debug config for Lambda function ARN available -> extend timeouts
  * `timeout-disable=False` -> extend timeouts
  * `timeout-disable=True` -> enforce timeouts as expected

Am I getting confused here or is this a logic error (the default value is `False`, which gets negated in `is_lambda_debug_timeout_for`) 🤔

sidenote: Technically, we're not fully disabling the timeouts, but extend them.


We need to document the behavior clearly after merging ...

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
dfangl: I think it's a nice idea to see those changes live. However, currently users are using debug mode very often, which might lead to being too verbose. As long as the config changes are limited to a handful, I think it is fine, once it gets too much, maybe one consolidated line of changes?

dfangl (approved):
LGTM, thanks for adding this!

PR #11400, opened 2024-08-21
APIGW: add better validation for PutIntegration and PutIntegrationResponse

bentsku:
## Motivation
As reported in our Slack Community channel with https://localstack-community.slack.com/archives/CMAFN2KSP/p1724147479425329, we raised a bad error when using the wrong `httpMethod` with `PutIntegrationResponse`. This PR adds better exception handling for `PutIntegration` and `PutIntegrationResponse`

`api error InternalError: exception while calling apigateway.PutIntegrationResponse: 'dict' object has no attribute 'method_integration'`

## Changes
- add test for bad resourceId and bad httpMethod for those 2 operations
- improve the exceptions in the provider


cloutierMat (approved):
:rocket: Nice addition! 
Thanks for updating this!

PR #11401, opened 2024-08-21
Apigw ng http deployment

cloutierMat:
## Motivation

Updating spec for apigwv2, as the aws endpoint returns 201 and the spec specifies 200 for this call.


bentsku (approved):
Awesome, thanks! LGTM :shipit:

Issue #11389, opened 2024-08-21
bug: S3 object tags are not copied if versioning is enabled on a S3 bucket


kumarselvadgm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using LocalStack to copy S3 objects between buckets, the tags of the source object are not copied to the destination object. This issue occurs even when using commands that should copy tags by default. It appears to be a discrepancy in how LocalStack handles object tagging during copy operations compared to AWS S3.

### Expected Behavior

Tags should be copied from the source object to the destination object when using the `s3 cp` command, similar to the behavior observed in AWS S3.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

**Create an S3 Bucket:**
Ensure you have a bucket named test-bucket in LocalStack.
aws --endpoint-url http://localhost:4566 s3 mb s3://$BUCKET_NAME

**Enable versioning on S3 Bucket:**
aws  $AWS_CLI_DEFAULTS s3api put-bucket-versioning --bucket $BUCKET_NAME --versioning-configuration Status=Enabled

**Upload an Object with Tags:**
`aws --endpoint=http://localhost:4566 s3 cp <local-file> s3://test-bucket/object-key`

`aws --endpoint=http://localhost:4566 s3api put-object \
    --bucket $BUCKET_NAME \
    --key $OBJECT_KEY \
    --body $FILE_PATH \
    --metadata $METADATA_STRING \
    --tagging "$TAGS_STRING"`

**Sample Tags String**

> project=example-project-2ae8b1b0-49ef-47a4-81e8-0137a257d73a&env=test&object-type=application/json&object-id=object-key

**Sample MetaData String**

> object-type=application/json,object-id=object-key,organization-id=d568ce6e-e76a-4ff8-88b4-b196902f3297,user-id=a779cdb9-bd36-42f9-a478-9bf81fc3b65f,event-type=DIP_FTF_TEST-8aa03205-d169-4d80-b601-57047b49f26c,bundle-id=8e989591-f7fa-46a0-9675-4c2952f069ca

**Verify Tags on Source Object:**
`aws --endpoint=http://localhost:4566 s3api get-object-tagging --bucket test-bucket --key object-key`

**Copy the Object to a New Key:**
Use the s3 cp command to copy the object to a new key within the same bucket.

`aws --endpoint=http://localhost:4566 s3 cp s3://test-bucket/object-key s3://test-bucket/dup-object-key`

[...]

PR #11403, opened 2024-08-22
fix CopyObject not copying tag of current object

bentsku:
## Motivation
As reported in #11389, we did not copy the source object tags when the `TaggingDirective` was set to `COPY` or nothing, which default to `COPY`. 

This was due to a small oversight, we were using the parsed version id to create the tag key, but if you don't specify a version id, you get the current object (top of the stack, the one that was last added). Instead of using the parsed version id, we now use the source object version id directly. 

## Changes
- use the source object version id to create the object unique key for the tags
- add a new test regarding versioned bucket and tagging directive to confirm the change: it first copy an object without specifying the version id (to get the current one), then does the same against a specific version id which isn't current

_fixes #11389_


cloutierMat (approved):
LGTM! :tada:

Issue #11109, opened 2024-06-27
bug: stepfunction sends sqs Message without MessageAttributes


SchmitzChristian:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a step function send a sqs message it is always without MessageAttributes when changing from legacy provider to the new one.

Also I have to add the events service when changing to new provider, is this intentional?

### Expected Behavior

Step function send sqs with configured MessageAttributes

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

docker compose file:

    version: '3.8'
    services:
      localstack:
        image: localstack/localstack:3.5.0
        ports:
          - "4570:4566"
        volumes:
          - ./localstack/init-localstack.sh:/etc/localstack/init/ready.d/init-localstack.sh
          - /var/run/docker.sock:/var/run/docker.sock
          - /var/lib/localstack:/var/lib/localstack
        environment:
          - SERVICES=stepfunctions,sqs,events
          - START_WEB=0
          - DEBUG=1

init-localstack.sh:

    `#!/usr/bin/env bash

    awslocal sqs create-queue --queue-name test-token --region eu-central-1

    awslocal stepfunctions create-state-machine \
      --name test-state-machine \
      --region eu-central-1 \
      --role-arn arn:aws:iam::012345678901:role/DummyRole \
      --definition '{
      "StartAt": "first",
      "States": {
        "first": {
          "Type": "Task",
          "Resource": "arn:aws:states:::sqs:sendMessage.waitForTaskToken",
          "Parameters": {
            "QueueUrl": "http://localstack:4566/000000000000/test-token",
            "MessageBody": {"TaskToken.$": "$$.Task.Token"},
            "MessageAttributes": {
              "task_token": {
                "DataType": "String",
                "StringValue.$": "$$.Task.Token"
              },
              "test": {
                "DataType": "String",
                 "StringValue": "testMessageAttribute"
              }
            }
          },
          "ResultSelector": {
            "result.$": "States.StringToJson($.result)"
          },
          "End": true
        }
      }
    }'

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal stepfunctions start-execution --state-machine-arn "arn:aws:states:eu-central-1:000000000000:stateMachine:test-state-machine"
Note: start the esecution with awslocal does not work for me right now, with typescript it works...

[...]

PR #11405, opened 2024-08-22
Add MessageAttributes to SQS messages when executed via Step Function

gregfurman:
## Motivation
- The Step Functions' SQS provider did not fully support the `MessageAttributes` field during a `SendMessage` since the initial implementation used the incorrect field name provided by the AWS docs. 
- This PR fixes the fieldname and adds a parity test to ensure a `MessageAttributes` JSON object and `Md5OfMessageAttributes` hash is correctly returned.
- Raised by https://github.com/localstack/localstack/issues/11109

## Changes
- Renames `"MessageAttribute" => "MessageAttributes"`
- Normalises field `"MD5OfMessageAttributes" => "Md5OfMessageAttributes"` to ensure PascalCase for SFN output
- Add new stepfunction template for SQS message with attributes

## Testing
- New parity test `test_send_message_attributes`
- Manually validated the original issue input now passes following these changes https://github.com/localstack/localstack/issues/11109

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_sqs.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/service/state_task_service_sqs.py
@@ -29,7 +29,7 @@ _ERROR_NAME_AWS: Final[str] = "SQS.AmazonSQSException"
 _SUPPORTED_API_PARAM_BINDINGS: Final[dict[str, set[str]]] = {
     "sendmessage": {
         "DelaySeconds",
-        "MessageAttribute",
+        "MessageAttributes",
         "MessageBody",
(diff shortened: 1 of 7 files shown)


MEPalma (approved):
Thanks @gregfurman, LGTM!

[Aside from the misleading docs which lead to this error](https://docs.aws.amazon.com/step-functions/latest/dg/connect-sqs.html), an underlying problem appears to be our generic API Action key normaliser (upstream of where you added this normalisation step): here it is not able to recognise MD5 is an acronym.
We should now incorporate this ad-hoc change, as it has already been applied to other scenarios in the SFN interpreter. This decision is further supported by the frequent occurrence of such issues. However, moving forward, we can stay mindful of this limitation and, with future evidence, consider developing a generalisable solution.

PR #11406, opened 2024-08-22
CFn: show CDK stack events

simonrw:
## Motivation

CDK shows a nice progress indicator when deploying a stack, however our implementation does not do this. This is because we don't emit `*_IN_PROGRESS` events.

image

https://github.com/user-attachments/assets/beccc4ae-7a7f-4569-80af-7ae0a8814edc

## Changes

- Emit `*_IN_PROGRESS` events on resource operations
- Move the instanciation of the resource provider to outside the `ResourceProviderExecutor` 
  - this was needed because on update we need to provider the physical resource id of the resource, and our handy `extract_physical_resource_id_from_model_with_schema` function requires the schema

## Final result

_Running a CDK deploy with SQS queue and SSM parameter with artificial sleep for the SQS queue shows the in-progress event_

https://github.com/user-attachments/assets/13b9d3db-929d-4a5e-9f5b-c274b0477560


pinzon (approved):
Nice changes!

PR #11408, opened 2024-08-23
cloudwatch: check amount of datapoints, refactor database-locks

steffyP:
## Motivation
When requesting `get-metric-statistics` LocalStack didn't have any checks on the amount of datapoints. 
Therefore, it would accept large queries, e.g. with a timeframe of over a year and a period of 1 second, which would mean that the data is aggregated for each second in that timeframe. 

AWS only accepts up to 1440 datapoints, and raises an error if those are exceeded.

As LocalStack was running queries in a loop, with a lock on the database, it basically made the cloudwatch provider unresponsive for any other requests (as the lock wasn't released).

In this PR we are trying to address two issues:
1) don't allow more than 1440 datapoints in one request (also being in parity with AWS)
2) refactor the database locks that we currently have in place: by keeping sql queries to a limited amount, and only use the database lock for query-execution

## Changes
* added new aws-validated test for the amount of expected datapoints
* refactored code and database locks

### DIFF
--- a/localstack-core/localstack/services/cloudwatch/cloudwatch_database_helper.py
+++ b/localstack-core/localstack/services/cloudwatch/cloudwatch_database_helper.py
@@ -44,10 +44,9 @@ class CloudwatchDatabase:
             return
 
         mkdir(self.CLOUDWATCH_DATA_ROOT)
-        with self.DATABASE_LOCK:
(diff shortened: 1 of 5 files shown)


pinzon (approved):
Thank you for the refactoring and adding the validation. The more validations we add, the better the quality of our samples.

PR #11411, opened 2024-08-26
CloudWatch: Add missing response fields

viren-nadkarni:
## Background

CloudWatch has the `/_aws/cloudwatch/metrics/raw` endpoint that returns the raw metrics. It was noticed that the [v1](https://github.com/localstack/localstack/blame/0063912ad5e984692e83e615a7b38021bc15e9c2/localstack-core/localstack/services/cloudwatch/provider.py#L345C9-L370) and [v2](https://github.com/localstack/localstack/blob/0063912ad5e984692e83e615a7b38021bc15e9c2/localstack-core/localstack/services/cloudwatch/cloudwatch_database_helper.py#L424-L454) providers do not have the same API response spec. To be specific, the `account` and `region` fields were missing in the v1 provider.

## Changes

In order to keep the spec consistent and simplify #11025, this PR adds the missing fields to the v1 provider response.

### DIFF
--- a/localstack-core/localstack/services/cloudwatch/provider.py
+++ b/localstack-core/localstack/services/cloudwatch/provider.py
@@ -364,6 +364,8 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
                 "v": r.value,
                 "t": r.timestamp,
                 "d": [{"n": d.name, "v": d.value} for d in r.dimensions],
+                "account": account_id,
+                "region": region,
             }
             for r in result
         ]


steffyP (approved):
If it makes things easier, I am ok with the change 😄 Thanks for unifying 🙌 

For some additional context:
As already outlined in https://github.com/localstack/localstack/pull/11025#discussion_r1731016166: I think this endpoint especially for v1 is not really usable. 
For v2 I already added a comment to the code, that we should either refine how the endpoint could be used (include aggregated statistics, using filters, pagination etc), or deprecate it and remove it all together at some point. 

The endpoint was created years ago in response to a [customer request](https://github.com/localstack/localstack/issues/3508), but it has never been documented and in the state that is currently is, I also wouldn't recommend to do so. However, when trying to pin down certain issues, it can be helpful for debugging, so it can definitely be useful in certain cases. 

Anyhow, this is out of the scope for what you are trying to accomplish 😄

PR #11413, opened 2024-08-26
Updated parameter mapper typing

cloutierMat:
## Motivation

Improve re-usability of the helper method by expanding it's typing to accept all dict.


bentsku (approved):
LGTM! :shipit:

PR #11414, opened 2024-08-26
remove CFn legacy template deployer

pinzon:
## Motivation
Since the release of [v3.5.0](https://github.com/localstack/localstack/tree/v3.5.0), the new DAG-based CloudFormation template deployer has been the default option. Given the absence of issues or problems since then, we are confident enough to permanently remove the legacy template deployer.

## Changes
- Remove legacy template deployer
- Remove configuration option for its activation

## Testing
- pipeline should still be green.


simonrw (approved):
A naming suggestion but once that's implemented then 👍 

For context: this was decided as non-breaking as the implementation change should be transparent for users, and we have not had any push back since v3.5.0 in early June to indicate that this new implementation is a problem. 

We have not had to suggest to users that this flag is an option, and I can't find any evidence of users using this flag. Therefore I think it's safe to remove.

PR #11422, opened 2024-08-27
Fix CFn bug when loading custom resources

simonrw:
## Motivation

With https://github.com/localstack/localstack/pull/11406 we added support for `*_IN_PROGRESS` events however we broke custom resource support in -ext.

The issue is that we perform a custom type mapping in -ext that we do not do in this repository. This means that the tests passed in isolation, but the -ext pipeline broke.

## Changes

- Map the resource type with `get_resource_type`, which is overridden in -ext to support custom resources

## Testing

- Trigger a manual -ext pipeline against this branch name
  - manual CI run is 🟢 apart from unrelated issue likely caused by our recent code restructuring


pinzon (approved):
Thanks for fixing this.

PR #11423, opened 2024-08-27
apigw ng request parameter handler

cloutierMat:
## Motivation

Updated signature of `_retrieve_parameter_from_variables_and_static` to improve it's re-usability. Also added a check to prevent `_retrieve_parameter_from_context_variables` from returning a dict.


bentsku (approved):
LGTM! Thanks for improving it 👌

PR #11424, opened 2024-08-28
APIGW NG: update UpdateApi status code

bentsku:
## Motivation
We need to patch the spec to return the 201 status code for the `UpdateApi` operation. 
Note: it might be worth it to do a full pass on all update operations to find the others and do only one PR
Follow up from #11401

## Changes
- update the patch specs for apigwv2


cloutierMat (approved):
Agreed, we can look at that next and round them all up at once!

This is a bit annoying that they send `created` for an update ops :face_with_diagonal_mouth:

PR #11426, opened 2024-08-28
Modify certificate download URL to new endpoint

dfangl:
## Motivation
With recent certificate issues, we need to change to a new endpoint where certificates are served from.

Also, to alleviate load on that endpoint, we should cache it for 24 hours, instead of 6 hours.

## Changes
* Download certificate from new endpoint
* Increase caching duration for the certificate to 24h

### DIFF
--- a/localstack-core/localstack/constants.py
+++ b/localstack-core/localstack/constants.py
@@ -28,7 +28,7 @@ MAVEN_REPO_URL = "https://repo1.maven.org/maven2"
 ARTIFACTS_REPO = "https://github.com/localstack/localstack-artifacts"
 
 # Artifacts endpoint
-ARTIFACTS_ENDPOINT = "https://analytics.localstack.cloud"
+ASSETS_ENDPOINT = "https://assets.localstack.cloud"
 
 # host to bind to when starting the services
 BIND_HOST = "0.0.0.0"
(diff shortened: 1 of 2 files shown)


dominikschubert (approved):
LGTM

cgdallas:
Hi All,

Question regarding this PR - I've noticed this morning that I'm suddenly unable to access Localstack in the Brave browser via https://app.localstack.cloud/inst/default/status.  The error in the console mentions a certificate revocation error.

https://localhost.localstack.cloud:4566/_localstack/health net::ERR_CERT_REVOKED

I've tried rebuilding the Docker image from scratch but nothing seems to resolve the issue for me.  I came to the repo to see if anyone was reporting similar issues and noticed this PR opened so I was curious if this would resolve that?

simonrw (approved):

AntonFagerberg:
@cgdallas We ran into certificate issues with OpenSearch as well today.
The latest version `3.6.0` didn't work so I pulled the docker image tagged as `latest` and that seems to work fine. I'm guessing this issue may have caused the issues in OpenSearch as well.

Would be great with a "versioned" release containing this fix so we can avoid having `latest`! 🙏

paymand:
@AntonFagerberg https://github.com/localstack/localstack/releases/tag/v3.7.0

elliotdobson:
We've just found this breaking change too...

elliotdobson:
We've just found this breaking change too...

@dfangl you should have held off on merging localstack/localstack-artifacts#32 until such date that the changes in this PR were widespread. eg. in a years time or something.

elliotdobson:
But it looks like the most of the certificates for `localhost.localstack.cloud` generated over the last week have been revoked so perhaps that is why you had to move so quickly with a breaking change like this?

https://crt.sh/?q=localhost.localstack.cloud

Looks like there is some discussion in localstack/localstack#11428. Is there any announcement regarding this issue?

PR #11427, opened 2024-08-28
enable integration between EventsBridge and Stepfunctions

pinzon:
## Motivation

For some reason the integration between EventsBridge and StepFunctions service was not being used. With this PR the integration is used with certain changes so the execution name of the machine uses the event ID as it happens in AWS.

## Changes

- Enable integration
- Add Event ID to Execution Name

## Testing 
- New CDK test that asserts the integration. It basically combines EventBridge, StepFunctions and SQS. The test consists on putting an event in to the bus. A rule will execute the state machine with the event as input. And the machine will send the message to an SQS Queue.


macnev2013 (approved):
LGTM 🚀

PR #11434, opened 2024-08-30
env var to delete the cached cert at startup

giograno:
## Motivation
Manually deleting the cached certificate could be a cumbersome operation for customers.
We should have a way to make this easier.

## Changes
- Introduce a `REMOVE_SSL_CERT` env var.
- Added a hook that removes the cached certificate at startup.


alexrashed (approved):
Thanks for this small convenience fix! I tested it locally, and the hook is executed at the right time. 💯

PR #11435, opened 2024-08-30
fix S3 region for DNS resolving

bentsku:
## Motivation
Seems like we've had an issue with resolving the `us-west-2` subdomain. Switched it to `eu-central-1` to fix the pipeline for now, and we'll keep looking into the issue. 

## Changes
- switch the region in `test_bucket_does_not_exist`


alexrashed (approved):
LGTM! 🚀 Thanks for jumping on this!

bentsku:
Merged as the S3 only tests were green, showing it fixed the issue

alphachart (commented):
.

PR #11436, opened 2024-08-30
remove old ARN partition rewriter

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/11296 we have a better solution for adjusting the regions for non-default partitions now.
This PR removes the old rewriter.

## Changes
- Remove the old rewriter as it became obsolete.


dfangl (approved):
Good to get rid of that terrible hack!

PR #11438, opened 2024-08-30
fix logs events with state machine

pinzon:
TBD


pinzon:
Wrong approximation to the  error. Closing in favor of PR in PRO projects

PR #11441, opened 2024-09-02
Replace auto-install of setuptools_scm with error in helper scripts

silv-io:
## Motivation

With the introduction of `setuptools_scm` our helper scripts started to depend on it. First, we added an automatic install of this dependency. However, unknowing users of those scripts might then install those dependencies into their global python environment which they might not want.
Decoupling the installation from the script execution will solve this.

## Changes

- Change check for installation of setuptools_scm to use `pip show`,  so that both `setuptools-scm` and `setuptools_scm` can be detected (at some version, they switched the naming)
- Display an ERROR when the package is missing instead of quietly installing it in the active environment
- Change the workflows that use the helper script to install the newest versions of `setuptools` and `setuptools-scm` to avoid version incompatibilities.

### DIFF
--- a/.github/workflows/asf-updates.yml
+++ b/.github/workflows/asf-updates.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           python-version: '3.11'
 
+      - name: Install release helper dependencies
+        run: pip install --upgrade setuptools setuptools_scm
+
       - name: Cache LocalStack community dependencies (venv)
         uses: actions/cache@v4
         with:
--- a/bin/docker-helper.sh
+++ b/bin/docker-helper.sh
@@ -49,9 +49,9 @@ function _fail {
 }
 
 function _get_current_version() {
-    # setuptools_scm will be installed transparently if not available, Python3 is expected to be present
-    if ! python3 -m pip -qqq list | grep -F "setuptools_scm"; then
-      python3 -m pip install -qqq setuptools_scm > /dev/null 2>&1
+    # check if setuptools_scm is installed, if not prompt to install. python3 is expected to be present
+    if ! python3 -m pip -qqq show setuptools_scm > /dev/null ; then
+      echo "ERROR: setuptools_scm is not installed. Run 'pip install --upgrade setuptools setuptools_scm'" && exit 1
     fi
     python3 -m setuptools_scm
 }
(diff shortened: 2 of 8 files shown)


bin/release-helper.sh
```diff
@@ -42,9 +42,9 @@ function usage() {
 }
 
 function get_current_version() {
-    # setuptools_scm will be installed transparently if not available, Python3 is expected to be present
-    if ! python3 -m pip -qqq list | grep -F "setuptools_scm"; then
-      python3 -m pip install -qqq setuptools_scm > /dev/null 2>&1
+    # check if setuptools_scm is installed, if not prompt to install. python3 is expected to be present
+    if ! python3 -m pip -qqq show setuptools_scm > /dev/null ; then
+      echo "ERROR: setuptools_scm is not installed. Run 'pip install --upgrade setuptools setuptools_scm'" && exit 1
```
alexrashed, line 47: Would be great if you could add a test for this behavior to the bats tests.

bin/docker-helper.sh
```diff
@@ -49,9 +49,9 @@ function _fail {
 }
 
 function _get_current_version() {
-    # setuptools_scm will be installed transparently if not available, Python3 is expected to be present
-    if ! python3 -m pip -qqq list | grep -F "setuptools_scm"; then
-      python3 -m pip install -qqq setuptools_scm > /dev/null 2>&1
+    # check if setuptools_scm is installed, if not prompt to install. python3 is expected to be present
+    if ! python3 -m pip -qqq show setuptools_scm > /dev/null ; then
+      echo "ERROR: setuptools_scm is not installed. Run 'pip install --upgrade setuptools setuptools_scm'" && exit 1
```
alexrashed, line 54: Would be great if you could add a test for this behavior to the bats tests.

.github/workflows/asf-updates.yml
```diff
@@ -25,6 +25,9 @@ jobs:
         with:
           python-version: '3.11'
 
+      - name: Install release helper dependencies
+        run: pip install --upgrade setuptools setuptools_scm
+
```
alexrashed, line 30: nit: This actually wouldn't be necessary (we only use `set-dep-ver`, which doesn't need `setuptools_scm`), but I guess it's the safe thing to do, so we could leave it like this 😉

alexrashed (commented):
Thanks a lot for jumping on this, and explicitly following up on @dfangl's [comment](https://github.com/localstack/localstack/pull/11355#discussion_r1732866967) on #11355. 🦸🏽 
The changes are looking good, but I think it would really be good to make sure that we keep testing the scripts, now that we have a proper testing infra in place.

bin/release-helper.sh
silv-io: added this now and it is green in the pipeline :)

alexrashed (approved):
Awesome! Thanks for addressing the comments! This is looking great! 💯

PR #11443, opened 2024-09-02
cloudwatch: move TaggingService to the store

steffyP:
## Motivation
Currently, persistence for tags in cloudwatch is not working properly, because we use the `TaggingService` directly in the provider. 
With this PR we move the `TaggingService` to the cloudwatch-store (for the v2 implementation).
We are not adapting the v1 implementation as we would encourage customers to use the new default provider.

## Changes
* moved `TaggingService` to the cloudwatch store
* updated related snapshot test (and enhanced with check for empty tag)

### DIFF
--- a/localstack-core/localstack/services/cloudwatch/models.py
+++ b/localstack-core/localstack/services/cloudwatch/models.py
@@ -9,6 +9,7 @@ from localstack.services.stores import (
     LocalAttribute,
 )
 from localstack.utils.aws import arns
+from localstack.utils.tagging import TaggingService
 
 
 class LocalStackMetricAlarm:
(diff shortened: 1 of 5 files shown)


bentsku (approved):
LGTM! This change looks good! Adding a new parameter to the store won't break it, as the existing store is always merged with the freshly instantiated one, so this shouldn't break anything. Awesome 😄 

As a side-comment, it might be worth adding bit more validation on the tags (trying to tag/untag/list tags for a non-existent resource, but outside the scope of this particular PR. Also, it seems tagging is supported only for [alarms and Contributor Insights rules](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_TagResource.html). But yes, sorry, outside of the scope 😄 

Nice seeing the `TaggingService` being used more widely, like @joe4dev said, if all services are using it, we might be able to support `resourcetaggingapi` more widely.

PR #11444, opened 2024-09-02
make OPENAPI_VALIDATE_REQUEST opt-in

giograno:
## Motivation
We have seen the OpenAPI validation create issues with snapshot persistence.
With this PR we make the request for validation opt-in (was by default) upon further investigation.

## Changes
- `OPENAPI_VALIDATE_REQUEST` is now switched off by default.


alexrashed (approved):
Thanks for jumping on this to 💚ify the pipeline :)

PR #11446, opened 2024-09-02
Revert "Add OpenAPI schema for public REST API (#11025)"

giograno:
This reverts commit fcd110db077900ffc93530e284913176ace39070.

## Motivation
We have seen OpenAPI validation to create issues with snapshot-based persistence.


alexrashed (approved):
If this turns out to be necessary (if the last merge didn't fix the downstream pipelines), I think this is looking good. Thanks for jumping on this!
If it turns out the pipelines in fact were fixed, we can just close this PR.

PR #11452, opened 2024-09-03
Add reverted OpenAPI commit

giograno:
## Motivation
With https://github.com/localstack/localstack/pull/11025 we involuntarily broke the persistence snapshots in the downstream pipeline. The reasoning is explained in https://github.com/localstack/localstack-ext/pull/3407. The issue should be solved once such PR gets merged.

## Changes
- Re-add the reverted commit.


.github/workflows/tests-pro-integration.yml
alexrashed, line 1: q: With #11444, the request validation was (similar to the request validation) moved to an opt-in approach.
Would it make sense to activate the request validation here as well?
Or should the opt-in maybe be reverted to an opt-out?

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -0,0 +1,86 @@
...
+            try:
+                self.openapi.validate_response(
+                    WerkzeugOpenAPIRequest(context.request),
+                    WerkzeugOpenAPIResponse(response),
+                )
+            except OpenAPIError as exc:
+                LOG.error("Response validation failed for %s: $s", path, exc)
+                response.status_code = 500
+                response.set_json({"error": exc.__class__.__name__, "message": str(exc)})
+                chain.stop()
```
alexrashed, line 86: q: Here we are (other than in the request validator) way stricter. This can make sense (because this is deactivated by default), but:
- Is this on purpose? Do we really raise on every `OpenAPIError` here?
- If it's on purpose, it would make sense to explain the reasoning here in a comment :)

alexrashed (approved):
The revert, and adjusting the error handling based on recent feedback, is looking good! I just added two questions, but when they are answered, we can move forward with the merge. 💯

localstack-core/localstack/aws/handlers/validation.py
giograno: The idea was to be stricter, as you suggested, since errors here should be implementation errors on the LS side.
However, I don't yet have a precise idea of every possible error we might encounter here. Therefore, I followed the same logic of the request validation and only checked for `ResponseValidationError` at the moment. Thoughts?

Issue #11448, opened 2024-09-03
feature request: s3 eventbridge notifications to include the version-id when bucket version is enabled.


eoghanArtrya:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Currently the version-id of objects created in s3 are not included in the resulting eventbridge notifications. Eyeballing the s3/notifications.py can see it is a property on the S3EventNotificationContext.

### 🧑‍💻 Implementation

https://github.com/localstack/localstack/blob/a23e2fc70542af481fb3a0bd7042627ff50f0802/localstack-core/localstack/services/s3/notifications.py#L708-L721
Just add `"version-id": ctx.key_version_id` to the object?

https://github.com/localstack/localstack/blob/a23e2fc70542af481fb3a0bd7042627ff50f0802/localstack-core/localstack/services/s3/notifications.py#L99-L116

### Anything else?

I am willing to look further at this and open a PR but i do want to first confirm its not a premium feature or something? It does look to be completely omitted.

bentsku:
Hello @eoghanArtrya and thanks for the report!

I can tackle it this afternoon, as I can take the opportunity to also improve our handling of DeleteMarkers (currently we always send `Permanently Deleted` in EventBridge). I'll link the PR shortly!

PR #11453, opened 2024-09-03
fix S3 VersionId EventBridge and fix DeleteMarker notification

bentsku:
## Motivation
As reported in #11448, we did not add the `version-id` field in EventBridge S3 notifications, even though we did for other services.
This PR fixes that, and as the request was about bucket versioning, I made sure that we handled correctly `DeleteMarker`. 
(Quick primer on how versioning works in S3 and Delete Markers: https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html) 

It adds a test for EventBridge and SQS, with a flow adding 2 objects on the same key, then deleting the "top most" key to create a delete marker, then deleting a specific object version.
It then suspend versioning, delete a delete marker to verify behavior, and add a new object on top with a `null` version to check. 

Improved parity along the way for a few fields. 

## Changes
- implement logic around `DeleteMarker` creation
- fix the missing `version-id` field for EventBridge target
- also updated an existing test that was testing versioned bucket, to test regular behavior vs versioned in a specific test 

_fixes #11448_

### DIFF
--- a/localstack-core/localstack/services/s3/notifications.py
+++ b/localstack-core/localstack/services/s3/notifications.py
@@ -198,6 +198,7 @@ class S3EventNotificationContext:
         :return: S3EventNotificationContext
         """
         bucket_name = request_context.service_request["Bucket"]
+        event_type = EVENT_OPERATION_MAP.get(request_context.operation.wire_name, "")
 
         # TODO: test notification format when the concerned key is FakeDeleteMarker
         # it might not send notification, or s3:ObjectRemoved:DeleteMarkerCreated which we don't support yet
@@ -207,6 +208,7 @@ class S3EventNotificationContext:
             key_expiry = None
             storage_class = ""
         else:
+            # if event_type == Event.s3_LifecycleExpiration_Delete and obj
             etag = s3_object.etag.strip('"')
             key_size = s3_object.size
             key_expiry = s3_object.expires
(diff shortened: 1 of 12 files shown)


localstack-core/localstack/services/s3/notifications.py
```diff
@@ -421,14 +423,18 @@ def _get_event_payload(
...
 
         event_type = ctx.event_type.lower()
         if any(e in event_type for e in ("created", "restore")):
-            record["s3"]["object"]["size"] = ctx.key_size
-            record["s3"]["object"]["eTag"] = ctx.key_etag
+            if "deletemarker" not in event_type:
+                record["s3"]["object"]["eTag"] = ctx.key_etag
+                record["s3"]["object"]["size"] = ctx.key_size
+            else:
+                record["s3"]["object"]["eTag"] = "d41d8cd98f00b204e9800998ecf8427e"
```
sannya-singal, line 436: Can we please add a comment in the code here why we added the this top most key: similar to what you have mentioned in the PR description?

localstack-core/localstack/services/s3/notifications.py
```diff
@@ -207,14 +208,15 @@ def from_request_context_native(
             key_expiry = None
             storage_class = ""
         else:
+            # if event_type == Event.s3_LifecycleExpiration_Delete and obj
```
sannya-singal, line 211: nit: maybe a leftover during implementation?

sannya-singal (approved):
Awesome work Ben 🎉 Great implementation and cleanup work! Just minor comments for a leftover and adding a brief comment in the code. Changes LGTM! 🚀

PR #11456, opened 2024-09-03
lambda fix create self managed event source

cloutierMat:
## Motivation

This PR aims to fix support for self managed kafka event sources.

Lambda function event source mapping usually has a `EventSourceArn` as a AWS resource arn, but Kafka sources can be [self managed](https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html) in which case this value is not present in the request. 

## Changes

- Prevent `KeyError` by removing direct access to the key where necessary
- Added CRUD support for Kafka event source
- Added tests

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -8,7 +8,7 @@ import os
 import re
 import threading
 import time
-from typing import IO, Optional, Tuple
+from typing import IO, Any, Optional, Tuple
 
 from localstack import config
 from localstack.aws.api import RequestContext, ServiceException, handler
@@ -1805,9 +1805,10 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         # TODO: find a better abstraction to create these
         params = request.copy()
         params.pop("FunctionName")
-        batch_size = api_utils.validate_and_set_batch_size(
-            request["EventSourceArn"], request.get("BatchSize")
-        )
+        if not (service_type := self.get_source_type_from_request(request)):
+            raise InvalidParameterValueException("Unrecognized event source.")
+
+        batch_size = api_utils.validate_and_set_batch_size(service_type, request.get("BatchSize"))
         params["FunctionArn"] = fn_arn
         params["BatchSize"] = batch_size
         params["UUID"] = new_uuid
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1805,23 +1805,41 @@ def create_event_source_mapping_v1(
...
+        batch_size = api_utils.validate_and_set_batch_size(service_type, request.get("BatchSize"))
         params["FunctionArn"] = fn_arn
         params["BatchSize"] = batch_size
         params["UUID"] = new_uuid
         params["MaximumBatchingWindowInSeconds"] = request.get("MaximumBatchingWindowInSeconds", 0)
         params["LastModified"] = api_utils.generate_lambda_date()
         params["FunctionResponseTypes"] = request.get("FunctionResponseTypes", [])
         params["State"] = "Enabled"
-        if "sqs" in request["EventSourceArn"]:
+        if "sqs" in service_type:
```
joe4dev, line 1819: nit: maybe worth commenting that it covers `sqs` and `sqs-fifo`

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1805,23 +1805,41 @@ def create_event_source_mapping_v1(
...
             params["StateTransitionReason"] = "USER_INITIATED"
             if batch_size > 10 and request.get("MaximumBatchingWindowInSeconds", 0) == 0:
                 raise InvalidParameterValueException(
                     "Maximum batch window in seconds must be greater than 0 if maximum batch size is greater than 10",
                     Type="User",
                 )
+        elif service_type == "kafka":
+            params["StartingPosition"] = request.get("StartingPosition", "TRIM_HORIZON")
+            params["StateTransitionReason"] = "USER_INITIATED"
```
joe4dev, line 1828: Such state updates are unrelated to kafka and should ultimately handled differently. I'm okay with hardcoding these for ESM v1, but we should revisit this in ESM v2.
/cc @gregfurman

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1805,23 +1805,41 @@ def create_event_source_mapping_v1(
...
+            params["LastProcessingResult"] = "No records processed"
+            consumer_group = {"ConsumerGroupId": new_uuid}
+            if request.get("SelfManagedEventSource"):
```
joe4dev, line 1831: note for the future: might need some validation if both are provided, but good for now

joe4dev (approved):
Thank you very much @cloutierMat  for this ESM v1 fix and adding an aws-validated API test 👏 

Can we run the ext tests to capture potential regressions (e.g., `tests.aws.services.kafka.test_kafka.TestKafka.test_kafka_lambda_event_source_mapping`) before merging?

@gregfurman We should remember to run the full test suite against ESM v2, including API tests, and maybe consider moving them into the esm directory.

gregfurman (approved):
LGTM! 💯

cloutierMat:
@joe4dev The ext test were run in the companion pr. I also added an extra test for this use case.

Issue #11115, opened 2024-06-30
bug: SNS Message Size Limit Handling Discrepancy


nirf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

SNS has a maximum message size limit of 256KB, which includes both the message payload and message attributes. However, LocalStack only accounts for the message payload when calculating the message size. As a result, publishing a message with a total size of 256KB (excluding message attributes) does not fail on LocalStack but fails on AWS.

I noticed a similar implementation for SQS has been addressed: [SQS Implementation Reference](https://github.com/localstack/localstack/pull/7168/files).

The bug appears to be in the following file:

	•	File: [localstack/services/sns/provider.py](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/sns/provider.py)
	•	Line: 520

### Expected Behavior

The behavior when publishing a large message should be consistent between LocalStack and AWS. Specifically, the total message size should include both the message payload and the message attributes, as per AWS documentation.

Reference:
For more details on handling large message payloads with SNS, please refer to the AWS documentation: [Large Message Payloads with SNS](https://docs.aws.amazon.com/sns/latest/dg/large-message-payloads.html).

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

```javascript
const { SNSClient, PublishCommand } = require("@aws-sdk/client-sns");

const snsClient = new SNSClient({ region: "us-east-1" });

const message = "a".repeat(256 * 1024); // 256KB message payload

const params = {
  Message: message,
  TopicArn: "arn:aws:sns:us-east-1:123456789012:MyTopic",
  MessageAttributes: {
    "AttributeKey": {
      DataType: "String",
      StringValue: "localstackTest"
    }
  }
};

const run = async () => {
  try {
    const data = await snsClient.send(new PublishCommand(params));
    console.log("Success, message published. MessageID:", data.MessageId);
  } catch (err) {
    console.error("Error", err.stack);
  }
};

run()
  .then(() => {
    console.log("Run function executed successfully.");
  })
  .catch(err => {
    console.error("Run function encountered an error:", err.stack);
  });
```

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

PR #11458, opened 2024-09-04
fix SNS Publish and PublishBatch message size calculation and restriction

bentsku:
## Motivation
As reported by #11115, we did not take into account the `MessageAttributes` while calculating the max message size. This was a shortcut taken at the time, but we can now properly fix it. The implementation is inspired by SQS as reported in the GitHub issue.

The documentation specifies: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
> All parts of the message attribute, including name, type, and value, are included in the message size restriction, which is 256 KB.

I've taken the opportunity to also validate the max message size for `PublishBatch` as well, as it was easy to do so. 

## Changes
- take into account `MessageAttributes` when calculating the message size following the documentation, and validated the assumption
- add message size validation for `PublishBatch`
- improve the exiting validated test, and added one for `PublishBatch`

_fixes #11115_

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -520,8 +520,7 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
                 f"Invalid parameter: PhoneNumber Reason: {phone_number} is not valid to publish to"
             )
 
-        if len(message) > MAXIMUM_MESSAGE_LENGTH:
-            raise InvalidParameterException("Invalid parameter: Message too long")
+        validate_publish_size(message, message_attributes)
 
         # for compatibility reasons, AWS allows users to use either TargetArn or TopicArn for publishing to a topic
         # use any of them for topic validation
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/sns/provider.py
```diff
@@ -1035,6 +1045,29 @@ def get_next_page_token_from_arn(resource_arn: str) -> str:
...
 
+def _get_byte_size(payload: str | bytes) -> int:
+    # Calculate the real length of the byte object if the object is a string
+    return len(to_bytes(payload))
+
+
+def get_total_publish_size(
+    message_body: str, message_attributes: MessageAttributeMap | None
+) -> int:
+    size = len(to_bytes(message_body))
```
sannya-singal, line 1056: we can use the helper function `_get_byte_size` to calculate the message body size here.

localstack-core/localstack/services/sns/provider.py
```diff
@@ -1035,6 +1045,29 @@ def get_next_page_token_from_arn(resource_arn: str) -> str:
...
+    if message_attributes:
+        # https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+        # All parts of the message attribute, including name, type, and value, are included in the message size
+        # restriction, which is 256 KB.
+        # iterate over the Keys and Attributes, adding the length of the Key to the length of all Attributes values
+        # (DataType and StringValue or BinaryValue)
+        size += sum(
+            _get_byte_size(key) + sum(_get_byte_size(attr_value) for attr_value in attr.values())
+            for key, attr in message_attributes.items()
+        )
```
sannya-singal, line 1066: Awesome 🎉 We now consider all the parts to calculate the message size!

sannya-singal (approved):
Thanks for fixing the message size calculations 🎉 🙌 Nice set of changes, just a minor comment, rest LGTM!

localstack-core/localstack/services/sns/provider.py
bentsku: Ah right, oopsie 😄 good catch!

PR #11460, opened 2024-09-04
add G ruleset to Ruff (logging-format)

bentsku:
## Motivation
I've seen f-strings in newly created PR,  but f-string are eagerly executed even if the logging level would not have that statement printed out, and the default way of using the Python logger will defer execution/interpolation only if needed.

https://docs.python.org/3/howto/logging.html#optimization
https://docs.astral.sh/ruff/rules/#flake8-logging-log

Adding these new rules would prevent f-string to be used inside logging statements, something we often comment on during PR reviews. Using the linter to prevent it should alleviate and make review easier. 

We had a lot of those in tests, which shouldn't matter really as we're not aiming for that kind of performance, but it should still be best practice to do so. 

There's the downside that sometimes the readability suffers a bit, but `extra` could be used in those case. 
See https://docs.astral.sh/ruff/rules/logging-f-string/

## Changes
- add new `G` ruleset in Ruff to catch string interpolation in logging statements
- fix existing issues

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_import.py
+++ b/tests/aws/services/apigateway/test_apigateway_import.py
@@ -239,7 +239,7 @@ def apigateway_placeholder_authorizer_lambda_invocation_arn(
             try:
                 aws_client.lambda_.delete_function(FunctionName=arn)
             except Exception:
-                LOG.debug(f"Unable to delete function {arn=} in cleanup")
+                LOG.debug("Unable to delete function %s in cleanup", arn)
 
 
 @pytest.fixture
--- a/localstack-core/localstack/aws/handlers/service_plugin.py
+++ b/localstack-core/localstack/aws/handlers/service_plugin.py
@@ -64,8 +64,9 @@ class ServiceLoader(Handler):
                 request_router.add_skeleton(service_plugin.skeleton)
             else:
                 LOG.warning(
-                    f"found plugin for '{service_name}', "
-                    f"but cannot attach service plugin of type '{type(service_plugin)}'",
+                    "found plugin for '%s', but cannot attach service plugin of type '%s'",
+                    service_name,
+                    type(service_plugin),
                 )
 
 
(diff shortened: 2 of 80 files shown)


tests/aws/services/apigateway/test_apigateway_import.py
```diff
@@ -239,7 +239,7 @@ def _is_not_pending():
             try:
                 aws_client.lambda_.delete_function(FunctionName=arn)
             except Exception:
-                LOG.debug(f"Unable to delete function {arn=} in cleanup")
+                LOG.debug("Unable to delete function %s in cleanup", arn)
```
silv-io, line 242: This seems to not produce the exact same output. I think `{arn=}` will render to `arn=<arn>` and the change would only output `<arn>`.

Since, it's a debug message it is fine though, I think :)

silv-io (approved):
This inconsistency was something that always bothered me. Great that there is a rule for fixing it!

LGTM, maybe there is also a rule for having all non-LOG strings be f-strings (they are currently the fastest interpolation options afaik)? Then we would use only two variants for string interpolation.

alexrashed (approved):
Awesome! Thanks for jumping on this! We mentioned that often in PRs when reviewing, it's great to give devs early feedback by including these linting rules! 🧹

tests/aws/services/apigateway/test_apigateway_import.py
bentsku: Good point! I missed that, I did all of it manually! I guess it's alright, but good catch, thanks for checking everything 🙇

bentsku:
> LGTM, maybe there is also a rule for having all non-LOG strings be f-strings (they are currently the fastest interpolation options afaik)? Then we would use only two variants for string interpolation.

@silv-io I found this one: https://docs.astral.sh/ruff/rules/f-string/ (UP032) for removing `.format()` call with f-string. But I didn't find something for `%` yet. 

I totally agree we should look into adding more rules, there are pretty nice ones looking at it, like forbidding builtin overrides and such.

PR #11461, opened 2024-09-04
Simplify ACM patch

giograno:
## Motivation
This PR simplifies (a little) the moto patch we have in the ACM provider concerning the `DescribeCertificate` operation.
Internally, `moto` [uses](https://github.com/getmoto/moto/blob/fb51925d2be3a14d96abc12349be3ad332d18dc1/moto/acm/responses.py#L129-L136) the patched method to implement `ListCerficates`. This response for this operation contains an `InUse` key that `moto` was not returning. 
Therefore, to have the key in LocalStack we were patching the `describe` method to have it landed in `ListCertificates`.

We fixed this in the upstream PR: https://github.com/getmoto/moto/pull/8054

## Changes
- Simplify `moto` patch.


alexrashed (approved):
LGTM! Thanks for cleaning up after this came in with the moto upgrade! 🧹

PR #11463, opened 2024-09-04
Enable release branches

silv-io:


silv-io:
this branch is quite outdated now. We'll need to rewrite it when we revisit this topic either way, so closing it now

Issue #11391, opened 2024-08-21
bug: Can't create a CloudWatch Alarm with a colon `:` in the name


cj-christoph-gysin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Trying to create a CloudWatch Alarm with a colon `:` in the name:
```
$ awslocal cloudwatch put-metric-alarm --alarm-name "foo:bar" --evaluation-periods 1 --comparison-operator GreaterThanThreshold
```
 fails with:
```
localstack  | Traceback (most recent call last):
localstack  |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
localstack  |     handler(self, self.context, response)
localstack  |   File "/opt/code/localstack/localstack/aws/handlers/service.py", line 113, in __call__
localstack  |     handler(chain, context, response)
localstack  |   File "/opt/code/localstack/localstack/aws/handlers/service.py", line 83, in __call__
localstack  |     skeleton_response = self.skeleton.invoke(context)
localstack  |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack  |   File "/opt/code/localstack/localstack/aws/skeleton.py", line 154, in invoke
localstack  |     return self.dispatch_request(serializer, context, instance)
localstack  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack  |   File "/opt/code/localstack/localstack/aws/skeleton.py", line 168, in dispatch_request
localstack  |     result = handler(context, instance) or {}
localstack  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack  |   File "/opt/code/localstack/localstack/aws/forwarder.py", line 136, in _call
localstack  |     return handler(context, req)
localstack  |            ^^^^^^^^^^^^^^^^^^^^^
localstack  |   File "/opt/code/localstack/localstack/aws/skeleton.py", line 118, in __call__
localstack  |     return self.fn(*args, **kwargs)
localstack  |            ^^^^^^^^^^^^^^^^^^^^^^^^
localstack  |   File "/opt/code/localstack/localstack/aws/api/core.py", line 161, in operation_marker
localstack  |     return fn(*args, **kwargs)
localstack  |            ^^^^^^^^^^^^^^^^^^^
[...]

PR #11465, opened 2024-09-04
fix parsing issue for cloudwatch alarm names

steffyP:
## Motivation
Fixes https://github.com/localstack/localstack/issues/11391
The parsing of alarm-names does not work correctly when the alarm-name contains a colon `:` which is apparently a valid name for alarms.

Also checked the same for dashboard-names where this pattern is not allowed. (Only adding the fix for the default provider v2.)

## Changes
* fixed creation of `AlarmArn` which was the root cause (it would not parse the pattern and instead just return the alarm-name as it contains a colon)
* adapted one testcase to use alarm-name with a colon to make sure this is valid on AWS and passes on LocalStack
* added validation for Dashboard name and added test

### DIFF
--- a/localstack-core/localstack/utils/aws/arns.py
+++ b/localstack-core/localstack/utils/aws/arns.py
@@ -195,9 +195,10 @@ def dynamodb_stream_arn(
 
 
 def cloudwatch_alarm_arn(alarm_name: str, account_id: str, region_name: str) -> str:
-    pattern = "arn:%s:cloudwatch:%s:%s:alarm:%s"
     # format pattern directly as alarm_name can include ":" and this is not supported by the helper _resource_arn
-    return pattern % (get_partition(region_name), region_name, account_id, alarm_name)
+    return (
+        f"arn:{get_partition(region_name)}:cloudwatch:{region_name}:{account_id}:alarm:{alarm_name}"
+    )
 
 
 def cloudwatch_dashboard_arn(dashboard_name: str, account_id: str, region_name: str) -> str:
--- a/localstack-core/localstack/utils/aws/arns.py
+++ b/localstack-core/localstack/utils/aws/arns.py
@@ -196,12 +196,13 @@ def dynamodb_stream_arn(
 
 def cloudwatch_alarm_arn(alarm_name: str, account_id: str, region_name: str) -> str:
     pattern = "arn:%s:cloudwatch:%s:%s:alarm:%s"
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/utils/aws/arns.py
```diff
@@ -196,12 +196,13 @@ def dynamodb_stream_arn(
 
 def cloudwatch_alarm_arn(alarm_name: str, account_id: str, region_name: str) -> str:
     pattern = "arn:%s:cloudwatch:%s:%s:alarm:%s"
-    return _resource_arn(alarm_name, pattern, account_id=account_id, region_name=region_name)
+    # format pattern directly as alarm_name can include ":" and this is not supported by the helper _resource_arn
+    return pattern % (get_partition(region_name), region_name, account_id, alarm_name)
```
bentsku, line 200: Good catch and nice fix!

nit: you could also use f-strings here
```suggestion
    # format pattern directly as alarm_name can include ":" and this is not supported by the helper _resource_arn
    return f"arn:{get_partition(region_name)}:cloudwatch:{region_name}:{account_id}:alarm:{alarm_name}"
```

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1393,6 +1393,37 @@ def contains_receive_delete_metrics() -> int:
 
         snapshot.match("get_metric_data_2", response)
 
+    @markers.aws.validated
+    @pytest.mark.skipif(
+        condition=not is_aws_cloud(), reason="Old provider is not raising exception"
+    )
```
bentsku, line 1399: q: should this test still run for the new provider? is it missing `condition not is_aws_cloud() and is_old_provider()`?

bentsku (approved):
LGTM! Nice fix, clean and concise! Just 2 minor nits, but nothing blocking a merge 😄

sannya-singal (approved):
Thanks for fixing this Steffy 🚀 Nice work! Nothing in addition to what Ben already mentioned, changes LGTM!

tests/aws/services/cloudwatch/test_cloudwatch.py
steffyP: oops, copy-pasta 😅 great catch, thanks ❤️

localstack-core/localstack/utils/aws/arns.py
steffyP: sure, will do :)

PR #11469, opened 2024-09-05
Support quiet downloads

simonrw:
## Motivation

Sometimes the download process is very noisy. In particular it logs every megabyte which could be a lot of log spam for large files.

## Changes

* Add flag to the `download` function to disable logging debug messages


dfangl (approved):
LGTM, nice addition!

PR #11472, opened 2024-09-05
feat: make packaged terraform version user-configurable

the-wondersmith:
## Motivation

PR addresses #11473, specifically:

1) The pinned version in the terraform package is [>2 years old](https://github.com/hashicorp/terraform/releases/tag/v1.1.3)
2) Allowing users to specify the version of terraform they need installed simplifies IaC and helps eliminate skew on the user side of things (i.e. users wouldn't be forced to maintain two sets of terraform files because localstack is locked to `v1.1.3`)

## Changes
1) Updates the pinned version to the latest non-BSL version of terraform
  - note: the current pinned version of terraform (`v1.1.3`) uses the same license as the newly pinned version (`v1.5.7`)
    - ref:
      - [v1.1.3](https://github.com/hashicorp/terraform/blob/v1.1.3/LICENSE)
      - [v1.5.7](https://github.com/hashicorp/terraform/blob/v1.5.7/LICENSE)
      - [v1.6.0](https://github.com/hashicorp/terraform/blob/v1.6.0/LICENSE)

2) Consolidate the places where the terraform version is used or defined from 3 to one (i.e. single source of truth)
3) Allow users to specify their desired version of terraform via the `LS_TERRAFORM_VERSION` environment variable, falling back to `1.5.7` is the environment variable is not set

### DIFF
--- a/localstack-core/localstack/packages/terraform.py
+++ b/localstack-core/localstack/packages/terraform.py
@@ -7,7 +7,7 @@ from localstack.packages.core import ArchiveDownloadAndExtractInstaller
 from localstack.utils.files import chmod_r
 from localstack.utils.platform import get_arch
 
-TERRAFORM_VERSION = "1.1.3"
+TERRAFORM_VERSION = os.getenv("LS_TERRAFORM_VERSION", "1.5.7")
 TERRAFORM_URL_TEMPLATE = (
     "https://releases.hashicorp.com/terraform/{version}/terraform_{version}_{os}_{arch}.zip"
 )
@@ -15,10 +15,10 @@ TERRAFORM_URL_TEMPLATE = (
 
 class TerraformPackage(Package):
     def __init__(self):
-        super().__init__("Terraform", "1.1.3")
+        super().__init__("Terraform", TERRAFORM_VERSION)
 
     def get_versions(self) -> List[str]:
-        return ["1.1.3"]
+        return [TERRAFORM_VERSION]
 
     def _get_installer(self, version: str) -> PackageInstaller:
         return TerraformPackageInstaller("terraform", version)
--- a/localstack-core/localstack/packages/terraform.py
+++ b/localstack-core/localstack/packages/terraform.py
@@ -7,7 +7,7 @@ from localstack.packages.core import ArchiveDownloadAndExtractInstaller
 from localstack.utils.files import chmod_r
 from localstack.utils.platform import get_arch
 
-TERRAFORM_VERSION = os.getenv("LS_TERRAFORM_VERSION", "1.5.7")
+TERRAFORM_VERSION = os.getenv("TERRAFORM_VERSION", "1.5.7")
 TERRAFORM_URL_TEMPLATE = (
     "https://releases.hashicorp.com/terraform/{version}/terraform_{version}_{os}_{arch}.zip"
 )


the-wondersmith:

localstack-core/localstack/packages/terraform.py
```diff
@@ -7,18 +7,18 @@
 from localstack.utils.files import chmod_r
 from localstack.utils.platform import get_arch
 
-TERRAFORM_VERSION = "1.1.3"
+TERRAFORM_VERSION = os.getenv("LS_TERRAFORM_VERSION", "1.5.7")
```
alexrashed, line 10: We don't use prefixes for our environment variables anywhere else, so I think here it should also not be prefixed.
```suggestion
TERRAFORM_VERSION = os.getenv("TERRAFORM_VERSION", "1.5.7")
```

alexrashed (commented):
Thanks a lot for your contribution, @the-wondersmith!
I am just curious: How are you using this package installer?
As far as I can tell, in localstack, we actually only use this package in our tests. It won't be available in the image by default.

the-wondersmith:
> Thanks a lot for your contribution, @the-wondersmith! 

Of course! My pleasure 😁

> I am just curious: How are you using this package installer? As far as I can tell, in localstack, we actually only use this package in our tests. It won't be available in the image by default.

Looking over the codebase, I think you're probably right - it _does_ appear to only be used in the tests within localstack _itself_, and terraform isn't available in the image by default for sure. _However_, if you _happen_ to use the [`localstack_terraform_init`](https://github.com/localstack/localstack-extensions/blob/main/terraform-init/) extension, the extension actually [uses](https://github.com/localstack/localstack-extensions/blob/main/terraform-init/localstack_terraform_init/extension.py#L9) the `localstack.packages.terraform` module to handle installation of terraform (presumably as part of the `terraform_local` wrapping that it looks to be doing).

I'm sure you'll never guess who _happens_ to have a set of terraform files with a required terraform version newer than `1.1.3` 😅

localstack-core/localstack/packages/terraform.py
the-wondersmith: I prefixed it on the _assumption_ that terraform might actually pay attention to an environment variable like `TERRAFORM_VERSION`, but upon actually [checking it](https://developer.hashicorp.com/terraform/cli/config/environment-variables), it absolutely does not.

tl;dr good catch, great call

the-wondersmith:
@alexrashed I _have_ been trying to get the CI actions to pass, but GitHub refuses to let me add a label to the PR, I apologize. Would you be able to add the `semver: patch` label for me? (That _should_ be the right level, given the change)

the-wondersmith:
> ... Would you be able to add the `semver: patch` label for me?

@alexrashed thank you for handling that!

It _looks_ like everything is in the green? Is there a doc I can refer to for what the merge process / timeline looks like? If not, do you have any idea what the merge process looks like and how long it might take for the change to hit the `docker.io/localstack/localstack-pro:latest` image?

alexrashed (approved):
Thanks for the explanation and for addressing the comment! The changes are looking good, and I'll merge them right away.
The changes will be in `localstack/localstack:latest` right after the build pipeline triggered by the merge is done, so in a few minutes. 😉 

Thanks for contributing to LocalStack! 🥳

PR #11474, opened 2024-09-06
Add CLI version command prefix

gtsiolis:
## Motivation

Besides LocalStack releases, CLI is also released with semantic versioning.

This can be confusing to users, see relevant discussions[[1](https://localstack-cloud.slack.com/archives/C060ZRAULEL/p1725464854452179?thread_ts=1725445367.911869&cid=C060ZRAULEL)][[2](https://localstack-cloud.slack.com/archives/C02N36URU1G/p1725467328964489)][[3](https://localstack-cloud.slack.com/archives/C02N36URU1G/p1725556984925929)][...]. 
Cc @SimonWallner @HarshCasper @dfangl @inothnagel @whummer 

The CLI _version_ command outputs a semantic number, but does not specify if this is referring to LocalStack or CLI.

## Changes

This will add LocalStack CLI as a prefix before the version number and style the message to be more subtle and different from the rest of the output, so that the version command is more explicit and can be used later on in every command invocation for reference.

| BEFORE | AFTER |
|--------|--------|
| Frame 48096558 | Frame 48096557 |
| Frame 48096556 | Frame 48096555 |

### DIFF
--- a/localstack-core/localstack/cli/localstack.py
+++ b/localstack-core/localstack/cli/localstack.py
@@ -155,7 +155,13 @@ _click_format_option = click.option(
         "show_default": True,
     },
 )
-@click.version_option(VERSION, "--version", "-v", message="%(version)s")
+@click.version_option(
+    VERSION,
+    "--version",
+    "-v",
+    message="LocalStack CLI %(version)s",
+    help="Show the version of the LocalStack CLI and exit",
+)
 @click.option("-d", "--debug", is_flag=True, help="Enable CLI debugging mode")
 @click.option("-p", "--profile", type=str, help="Set the configuration profile")
 def localstack(debug, profile) -> None:


gtsiolis:
I have read the CLA Document and I hereby sign the CLA ❤️

gtsiolis:

alexrashed (commented):
The change itself is looking good, clean and simple.
However, I think this chance could cause quite a few issues considering how this command is being used by our users. A plain "version only" output can be very handy when you use the output of the command in scripts.
When just looking at the usage of the command in a quick GitHub code search, we can see that there are quite a lot of usages, some of which might break with this change:
https://github.com/search?q=%22localstack+--version%22&type=code

Is there a chance we could make this distinction between the core emulator runtime and the CLI versions a bit more clear without modifying the output here directly?

gtsiolis:
> A plain "version only" output can be very handy when you use the output of the command in scripts.

Great points! 
1. We could be missing an opportunity for a better developer experience, because of the confusion between versions, see linked discussions in the PR description.
2. From the GitHub results, I see most simply run the version command to print the version, right? The other references that include some version splitting[[1](https://github.com/Frank9912/localstack/blob/4d3cfd1bd3abf85b20653e72575af928f5c95e53/localstack/utils/bootstrap.py#L262)][[2](https://github.com/friscoMad/localstack/blob/2b16692bef50ad38242962ca001492738b451361/localstack/utils/bootstrap.py#L262)][...] that we had in our source code seem to be outdated forks that haven't been updated for 3+ years.

I'd like to challenge the belief that the version only output is proving more value, but leaving this decision to you and the rest of the engineering group! 🏓 

> Is there a chance we could make this distinction between the core emulator runtime and the CLI versions a bit more clear without modifying the output here directly?

I don't have a better idea for this—let's close the PR unless someone else want to pick it up to try this. 📕 


> [...] the failing pipeline

Force pushed a commit to fix this! Thanks for the hint, @alexrashed! 💛

SimonWallner:
lets take it to the next level, by providing _even moar_ information and help the user to achieve what they might have come for. 🚀 

This all is based on my assumption that me as a user typing `localstack version` I want to know the version of localstack, not the cli.

I understand that we don't know the version of localstack right away, but still we can tell the user about it and how to find out. something like this

```
> localstack version
cli version: 1.2.3.4 (+++ there is an update available for the cli and emulator +++)

- To find out the version of the localstack emulator <insert instructions>.
- To access a parser friendly shorter output of this command use the --short flag.
- To update the cli and the localstack emulator to the latest version run <insert instructions>
- To pin the localstack emulator to a specific version <insert instructions or link to docs>

command help:
  --short, only display a short version number that is easy to parse

**Comment by SimonWallner** [comment]
briefly looked over the search results @alexrashed posted. I'd guess people intent to get the emulator version most of the time instead of cli version.

**Comment by alexrashed** [comment]
> briefly looked over the search results @alexrashed posted. I'd guess people intent to get the emulator version most of the time instead of cli version.

I am not sure what you want to say by that. Do you mean "people are having a UX issue because they get something else than they expect"?
If that's the case: Sure, I am not saying this isn't a UX issue. It definitely is! And thanks, @gtsiolis for jumping on this! But that doesn't mean that we can just break an API (which it is) here.

**Comment by gtsiolis** [comment]
Thanks for your feedback, @SimonWallner and @alexrashed! 💛

I opened this as a minimal viable change (MVC) out of confusion around the output of the _version_ command, but it seems this could benefit from an alignment discussion, probably in a different medium (Slack, meeting, etc). 🛹 

Taking a bias for action and closing this PR to be respectful of everyone's time. 🏓

**Comment by gtsiolis** [comment]
Reopened this, following up from [relevant discussion](https://localstack-cloud.slack.com/archives/C0550DW38CT/p1730196035709929) (internal).

The intention was to include a link to https://localstack.cloud/releases (see [relevant discussion](https://localstack-cloud.slack.com/archives/C02LK3UT9DF/p1730210177491679?thread_ts=1730209010.584679&cid=C02LK3UT9DF) (internal)), but maybe this creates more confusion than help, especially when linking to our blog (see [relevant discussion](https://localstack-cloud.slack.com/archives/C02R6RPN6E4/p1730210493321839) (internal)). 💭 

So, I was thinking could we merge this _**as is**_ as a minimal viable change (MVC) iteration?

We can always come back to improve this more, or even consider making this output more dynamic, see [relevant discussion](https://localstack-cloud.slack.com/archives/C060ZRAULEL/p1725445367911869) (internal).

What do you think, @SimonWallner and @alexrashed?

**Comment by SimonWallner** [comment]
I don't see the harm in adding the additional info, especially the hint on how to find out the _actual_ LS version, but this should not block this PR. Once we get this change in, we can improve further.

**Comment by gtsiolis** [comment]
Thanks, @SimonWallner! I will keep this as is for simplicity for now. 💜 

@alexrashed Can I get a final approval from you as code owner before merging? 🏓

**Review by alexrashed — COMMENTED** [review]
Thanks for reopening the PR! We can definitely try to get this done with v4, but I think it would still be great (as outlined in my initial review, especially the GitHub search link) to have a migration path for those users where this change will _break_ the usage of `localstack --version`. Namely those who directly use the output of the command and expect the version string without a prefix in a script.
I am happy to also conclude that we should just mention this in the list of breaking changes, but I feel like this fact (that you are actively breaking current usages with this change) didn't get any attention in the discussion here yet.

**Comment by gtsiolis** [comment]
@alexrashed I'm probably wrong here or missing some technical context. I'd opt for closing this PR again, but before doing that, I scheduled up a coffee chat with you next week to get a better idea of the breaking changes here and catch up! 💜

**Comment by alexrashed** [comment]
I'm happy to discuss this in detail, but I can also show it here on public examples of the usage of the command in the community.
- For example [in this line in a community homebrew formula](https://github.com/gromgit/homebrew-core-mojave/blob/f8320cc47fbd7e435ec6fbfec5253bac5f9fa84c/Formula/localstack.rb#L217): Here `localstack --version` is explicitly being expected to only return a number which is then directly being used as the version of the homebrew formula.
- Or this simple [print statement here in this utils shell script](https://github.com/megaeth-labs/eigenda/blob/70614c0e4204ff74706a854963724dd1fe5cd8aa/versions.sh#L43) will show a weird output after this change.

My whole argument is that the CLI is not _only_ used by humans. Our users are developers, and developer's gotta code! 😉

**Comment by SimonWallner** [comment]
@alexrashed should we introduce something like a `localstack --version-short` then that contains the pure version string?

**Comment by SimonWallner** [comment]
@gtsiolis @alexrashed if we go with the additional `--version-short` then we should hint on it in the regular output
```
>> localstack --version
LocalStack CLI 3.1.4

use 'localstack --version-short' for a parsing friendly short version output
```

this makes it easier for users who actually run into an error. I think it s fine to accept that things can break going from v3 to v4. (with all the mitigations and help in place to soften the blow)

**Comment by alexrashed** [comment]
>  if we go with the additional --version-short then we should hint on it in the regular output

Sounds good, but looking at other commands in the CLI, it seems to be a repeating pattern (changing the format of the output). Would it make sense to reuse the `--format` option to control the output format of the command?
Here's for example the help of the "config show" command:
```
$ localstack config show --help
Usage: localstack config show [OPTIONS]

  ...

Options:
  -f, --format [table|plain|dict|json]
                                  The formatting style for the command output.
                                  [default: table]
  ...
```

But that's just an observation. I do agree that we can break certain things with major releases, especially if it resolves these common confusions. For me it is also okay to just plainly change the output, but we need to know that we are going to break things, and we should do our best to give our users a heads-up and a migration path.
For example, this command would work with both the old and the suggested version text on Linux with `sed` available: `localstack --version | sed 's/.*\s//'`

Another option to solve the problem at hand might also be to just make the help output more concrete? Something like this:
```diff
 $ localstack --help
 ...
 Options:
-  -v, --version       Show the version and exit.
+  -v, --version       Show the version of the CLI (not the runtime) and exit.
   ...
```

> >> localstack --version
> LocalStack CLI 3.1.4
> 
> use 'localstack --version-short' for a parsing friendly short version output

I think this could cause more issues (for example for users who want to use the output of the command in a script)...
A new option would however of course show up in the help output.

thrau:
> My whole argument is that the CLI is not only used by humans. Our users are developers, and developer's gotta code! 😉

appreciate you alex for thinking customer first, as so often, and always trying to maintain backwards compatibility :-)

> didn't get any attention in the discussion here yet

i also definitely agree with this point. when we propose changes like this, we should always make sure we're intentional about potential breaking APIs. so let's do that, and your initial research already really helped here!

i believe, in this instance, the benefit of improving the UX (specifically helping users disambiguate between CLI and localstack runtime version), doesn't outweigh he risk of breaking a couple of shell script. people are evidently and understandably confused about the difference between the CLI and the runtime. the simple change that george is proposing here would help fix that, by stating clearly to users "hey, you are using the CLI!"

for people who are using the output of `localstack --version` as input for another program, i would seriously question whether they are _intentionally_ asking for the CLI version and not the runtime version. we could of course ask people, but i'd take an educated guess saying they're actually interested in "what localstack version am i using" not "which version of the localstack CLI am i using that will start my localstack version". 

so in summary, while i agree we should be intentional about breaking APIs, this seems, based on your research, like a non-critical and also potentially misleading API, that would perhaps benefit from being broken. so let's intentionally break it to make it better! i trust any developer who uses the output of a `--version` command in another script to apply `cut`, `sed` or similar tools ;-)

gtsiolis:
Sorry for the noise, everyone! 💜

alexrashed:
FYI: The failing pipeline is unrelated and is addressed in https://github.com/localstack/localstack/pull/11808. I'll directly move forward and merge this PR into our release branch.

PR #11475, opened 2024-09-06
feat: pass context to functions patched in eventstudio

maxhoheiser:
## Motivation
Eventstudio patches the boot client to propagate tracing data (parent_id and trace_id) in the header and automatically add it to the context.
If a event is stored in the extension via the patched localstack function, the following is treated as a new span and the parent_id is updated.

## Changes
- add context to target process_evente and 

## Testing
see eventstudio PR: https://github.com/localstack/localstack-extension-event-studio/pull/70


Morijarti (approved):
LGTM

PR #11476, opened 2024-09-06
Feature/SNS/Eventstudio: pass context to patched function required in eventstudio

maxhoheiser:
## Motivation
Eventstudio patches the boot client to propagate tracing data (parent_id and trace_id) in the header and automatically adds it to the context.
If an event is stored in the extension via the patched localstack function, the following is treated as a new span and the parent_id is updated.

## Changes
- add context to publisher `publish_to_topic` and `publish_batch_to_topic` 

## Testing
see EventStudio PR: https://github.com/localstack/localstack-extension-event-studio/pull/70

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -77,6 +77,7 @@ from localstack.utils.aws.arns import (
 )
 from localstack.utils.collections import PaginatedList, select_from_typed_dict
 from localstack.utils.strings import short_uid, to_bytes, to_str
+from localstack.utils.tracing import get_trace_context
 
 # set up logger
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


Morijarti (approved):

Morijarti:
Good work of adding context to functions we need for our extension.

bentsku (commented):
Quick question regarding the async nature of SNS: how are you planning to pass the trace id down to the actual clients? As the patched function is only the dispatcher to the actual client call to invoke asynchronously the other clients? 

Could it be worth to attach the necessary data to the actual `SnsPublishContext`, which is forwarded all the way? However, it would probably be needed to be trimmed down to a small object which would contain the data and not the full context, as it might have a certain cost to be forwarded to every single invocation. 

(Like attaching a simple dict that can be mutated through the invocations, attached to the context instead of attached as multiple fields to the context. That way it could be contained, and also mutated without passing the whole `RequestContext` around?)

I'm not aware of the full context of the changes here, so I'm a bit curious 😄 seems like an interesting challenge, glad to see you tackle it!

bentsku (changes requested):
Thanks a lot for this PR! 

I feel like I am lacking context in how passing the `TraceContext` to the `publish_to_topic` call is going to help propagating the trace to the actual boto client calls. 

I think it might be worth to attach it to the actual `SnsPublishContext` so that you would have access to it all the way across the logic, when using the boto clients to connect to the application, except if I am missing a big part of the implementation? 

We can have a look together how that might look or if there is something I am not understanding, but I think it would make your life simpler, as you won't need to change much of the signature except the original `SnsPublishContext` creation, which can be created from a `RequestContext`.

maxhoheiser:
As agreed we are going with a threading.local approach

PR #11477, opened 2024-09-07
[Lambda DevX] Support for Config Hot Reloading

MEPalma:
## Motivation
Currently the configuration for Lambda Debug Mode is loaded on startup and cannot be changed until the next restart of LocalStack. This may not only be inconvenient whenever changes to the configuration may be required after startup, but also means that it is impossible for the user to provide configurations for Lambda functions for which the arn is unknown before startup. Such scenario may occur when using dynamic naming in cloud formation. Hence, these changes add initial support for the hot reloading of the Lambda Debug Mode configuration file. This is achieved by updating the config when changes are detected in the file, and stopping a lambda container after evaluation iff this is being debugged. Stopping the container after evaluation is encouraged by the difficulty/feasibility or reconfiguring the container (consider ports), and has the added benefit of unblocking the scenario of debugging the same lambda function more than once.

## Changes
- Added watch and reload logic to the Lambda Debug Mode session class
- Add environment stop after environment execution iff the lambda function being run is target of the debug session 

## Testing
As the development of a testing stack for Lambdas debugging is being developed, I refer to manual testing using the Lambda Debug Mode sampled provided here: https://github.com/localstack-samples/localstack-pro-samples/pull/253


localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
```diff
@@ -11,61 +14,135 @@
...
+        self._load_lambda_debug_mode_config()
+        self._initialised_event.set()
+
+        # Monitor for file changes in an endless loop: this logic should be started as a daemon.
+        while True:
+            time.sleep(1)
+            epoch_last_modified = self._config_file_epoch_last_modified_or_now()
+            if epoch_last_modified > epoch_last_loaded:
+                epoch_last_loaded = epoch_last_modified
+                self._load_lambda_debug_mode_config()
```
dfangl, line 134: I am not a fan of this endless loops, even though the thread is a daemon thread. I would prefer an orderly shutdown of the thread, by using an stopping event as loop condition, and shutting it down on shutdown of the lambda service.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
```diff
@@ -11,61 +14,135 @@
...
+        # thread will then take care of updating the configuration periodically and asynchronously.
+        # This may somewhat slow down the first upstream thread loading this module, but not
+        # future calls. On the other hand, avoiding this mechanism means that first Lambda calls
+        # occur with no Debug configuration.
+        self._initialised_event = Event()
+
+        self._watch_thread = Thread(
+            target=self._watch_logic, args=(), daemon=True, name="LambdaDebugModeConfigWatch"
+        )
+        TMP_THREADS.append(self._watch_thread)
```
dfangl, line 62: Appending it to `TMP_THREADS` accomplishes nothing here, as it is unstoppable.
Instead, as seen in the other comment, I would prefer shutting it down when the lambda service shuts down.

localstack-core/localstack/services/lambda_/invocation/assignment.py
```diff
@@ -75,7 +76,10 @@ def get_environment(
 
         try:
             yield execution_environment
-            execution_environment.release()
+            if is_lambda_debug_timeout_enabled_for(lambda_arn=function_version.qualified_arn):
+                self.stop_environment(execution_environment)
```
dfangl, line 80: Is this something the users expect? This would effectively stop the possibility of leaving the debugger attached for multiple invocations, which should be theoretically possible currently, right?

Does it make sense, maybe, to only stop the environment if something actually changed for it?

dfangl (changes requested):
I have a few comments, especially for the lifecycle of the lambda environment, and the ability to stop the watching thread.

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
MEPalma: changed to `SERVICE_PLUGINS.is_running("lambda")`

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
MEPalma: I removed this

localstack-core/localstack/services/lambda_/invocation/assignment.py
MEPalma: Leaving the debugger attached is definitely the behaviour we envision. However, this is not an experience we can offer consistently at the moment. I cannot get this to work for python for example. Currently it is impossible to debug a lambda function more than once.
I am of the idea of providing this behaviour to unblock, whilst we continue the work on integrating more tightly the Lambda Debug Mode and the debuggers. What do you think?

dfangl: That is fine by me, let's keep it as a temporary workaround!

localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode_session.py
```diff
@@ -17,55 +21,148 @@
...
+    def _watch_logic(self) -> None:
+        # TODO: consider relying on system calls (watchdog lib for cross-platform support)
+        #  instead of monitoring last modified dates.
+        # Run the first load and signal as initialised.
+        epoch_last_loaded: int = self._config_file_epoch_last_modified_or_now()
+        self._load_lambda_debug_mode_config()
+        self._initialised_event.set()
+
+        # Monitor for file changes whilst the lambda service is running.
+        while SERVICE_PLUGINS.is_running("lambda"):
```
dfangl, line 149: Not too happy with this here. could we have an internal shutdown / running state (for example a threading Event) which is checked here, and call `stop()` triggering the event in `on_before_stop`, just as we do the `ensure_running` now?

dfangl (approved):
Looks good now, just a minor comment regarding the shutdown, I am not too happy with the polling, let's just have an internal event which get's set `on_before_stop`.

PR #11483, opened 2024-09-09
Feature/SQS/Eventstudio: Add context to queue.put

maxhoheiser:
## Motivation
Eventstudio patches the boot client to propagate tracing data (parent_id and trace_id) in the header and automatically adds it to the context.
If an event is stored in the extension via the patched localstack function, the following is treated as a new span and the parent_id is updated.
Lambda uses a "fake" sqs client and no context is available for calling the models.put method, therefore the context is optional.

## Changes
- add context to put message for standard queue and fifo queue  in models `StandardQueue.put` and `FifoQueue.put` 

## Testing
see EventStudio PR: https://github.com/localstack/localstack-extension-event-studio/pull/70

### DIFF
--- a/localstack-core/localstack/services/sqs/models.py
+++ b/localstack-core/localstack/services/sqs/models.py
@@ -251,14 +251,14 @@ class MessageMoveTask:
         source_arn: str,
         destination_arn: str,
         max_number_of_messages_per_second: int = None,
-        context: RequestContext = None,
+        trace_context: TraceContext | None = None,
     ):
         self.task_id = long_uid()
         self.source_arn = source_arn
         self.destination_arn = destination_arn
         self.max_number_of_messages_per_second = max_number_of_messages_per_second
         self.cancel_event = threading.Event()
-        self.submitted_request_context = context
+        self.submitted_request_trace_context = trace_context
 
     def mark_started(self):
         self.started_timestamp = datetime.utcnow()
--- a/localstack-core/localstack/services/sqs/models.py
+++ b/localstack-core/localstack/services/sqs/models.py
@@ -507,6 +507,7 @@ class SqsQueue:
     def put(
         self,
         message: Message,
+        context: RequestContext,
         visibility_timeout: int = None,
         message_deduplication_id: str = None,
         message_group_id: str = None,
@@ -738,11 +739,12 @@ class StandardQueue(SqsQueue):
     def put(
         self,
         message: Message,
+        context: RequestContext,
         visibility_timeout: int = None,
         message_deduplication_id: str = None,
         message_group_id: str = None,
         delay_seconds: int = None,
-    ):
+    ):  # context required by eventstudio
         if message_deduplication_id:
             raise InvalidParameterValueException(
                 f"Value {message_deduplication_id} for parameter MessageDeduplicationId is invalid. Reason: The "
@@ -994,11 +996,12 @@ class FifoQueue(SqsQueue):
     def put(
         self,
         message: Message,
+        context: RequestContext,
         visibility_timeout: int = None,
         message_deduplication_id: str = None,
         message_group_id: str = None,
         delay_seconds: int = None,
-    ):
+    ):  # context required by eventstudio
         if delay_seconds:
             # in fifo queues, delay is only applied on queue level. However, explicitly setting delay_seconds=0 is valid
             raise InvalidParameterValueException(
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/sqs/models.py
```diff
@@ -244,13 +246,18 @@ class MessageMoveTask:
     cancel_event: threading.Event
 
     def __init__(
-        self, source_arn: str, destination_arn: str, max_number_of_messages_per_second: int = None
+        self,
+        source_arn: str,
+        destination_arn: str,
+        max_number_of_messages_per_second: int = None,
+        context: RequestContext = None,
```
gregfurman, line 253: The `MessageMoveTask` is an async task that is submitted upon receiving a `StartMessageMoveTask`. 

So let's just pass the `StartMessageMoveTask's` context to the submitted `MessageMoveTask` class

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -507,6 +507,7 @@ def _run(self, move_task: MessageMoveTask):
 
                 target_queue.put(
                     message=message.message,
+                    context=move_task.submitted_request_context,
```
gregfurman, line 510: The `MessageMoveTask` now stores the submittor `StartMessageMoveTask` request's context. So just extract it from and pass it around.

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -1510,6 +1512,7 @@ def start_message_move_task(
                 source_arn,
                 destination_arn,
                 max_number_of_messages_per_second,
+                context=context,
```
gregfurman, line 1515: Pass the `StartMessageMoveTask.RequestContext` to the `MessageMoveTask`

thrau (commented):
Generally speaking I don't think we should assume that every layer of a service can only be used in the context of a `RequestContext`. Handing down the RequestContext to lower level layers limits the use of these layers to a single request. This is a pretty strong assumption architecture wise that I find a bit hard to swallow.

Before we move forward, I want to make sure we've exhausted other options that don't involve changing lover level APIs in this way. Are there any?

Also @dfangl will this affect the use of SQS for lambda where we may not have a request context?

maxhoheiser:
> Generally speaking I don't think we should assume that every layer of a service can only be used in the context of a `RequestContext`. Handing down the RequestContext to lower level layers limits the use of these layers to a single request. This is a pretty strong assumption architecture wise that I find a bit hard to swallow.
> 
> Before we move forward, I want to make sure we've exhausted other options that don't involve changing lover level APIs in this way. For instance holding the RequestContext in a thread local? We can patch queues and message move tasks to in a way that accesses the thread local. Are there any others you can think of?
> 
> Also @dfangl will this affect the use of SQS for lambda where we may not have a request context?

@thrau 

As discussed in the demo today, we came up with a new option as suggested by @bentsku to keep additional overhead low.
We are creating a DTO trace_context that only contains the required references (for now parent_id and trace_id) and it will only be passed if any of the referenced values is not None.

This reduces the overhead when the eventstudio extension is active but it will not have any overhead when it is not active since no value is passed in this case, because both parent_id and trace_id will not be present in the context anyways.

Furthermore, trace_context will be an optional input parameter, thus allowing also to use the internal methods where the context is not available.

dfangl:
As long as the implementation does not depend on any context, it should be good to go from the Lambda side :+1:

thrau (commented):
i can see the current solution introduces a new `TraceContext` concept which is i suppose less information than the RequestContext carries, but it doesn't really address my original concern, which is around how intrusive the solution is and how it will scale (in terms of code changes).

even with this concept, it's very intrusive in the sense that we need to update _every_ relevant method and start passing this trace context along. we'd need to surgically implant this to every service where it's relevant. are we really planning to update every class and method like we're doing here? who will do and maintain this?

instrumentation like this is typically added from the outside through some level of inversion of control. we previously demoed a solution that is a bit brittle, but much less intrusive https://github.com/localstack/localstack/commit/fdab47bae64d1ecd8fd9f8d0f9a699292b13b678, i wonder whether we've considered doing something more akin to this?

joe4dev:
> As long as the implementation does not depend on any context, it should be good to go from the Lambda side 👍

For Lambda event invokes (i.e., event_manager.py) and pull-based control flow (e.g., Lambda Event Source Mappings or Pipes), I think there's no way around adjusting some internal APIs and passing the `TraceContext` as metadata along the data flow. It might also require custom trace parenting logic for scenarios such as fan-in (e.g., ESM batching) and metadata enrichment for retries.
/cc @gregfurman

PR #11485, opened 2024-09-09
Refactor markers report workflow

sannya-singal:


scripts/render_marker_report.py
```diff
@@ -66,7 +66,10 @@ def render_template(*, template: str, enriched_report: EnrichedReport) -> str:
 
 
 def create_test_entry(entry, *, code_owners: CodeOwners, commit_sha: str, github_repo: str):
-    rel_path = "".join(entry["file_path"].partition("tests/")[1:])
+    base_dir = github_repo.split("/")[-1]
+
+    rel_path = entry["file_path"].split(base_dir)[-1].removeprefix("/")
```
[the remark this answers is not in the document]
sannya-singal: Related PR: https://github.com/localstack/localstack-ext/pull/3110

.github/workflows/marker-report.yml
```diff
@@ -43,3 +68,54 @@ jobs:
...
+
+      - name: Print generated markdown
+        if: ${{ inputs.createIssue }}
+        run: |
+          cat ./target/MARKER_REPORT_ISSUE.md
+
+      - name: Upload generated markdown
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./target/MARKER_REPORT_ISSUE.md
```
dominikschubert, line 106:
```suggestion
      - name: Upload generated markdown
        if: ${{ inputs.createIssue }}
        uses: actions/upload-artifact@v4
        with:
          path: ./target/MARKER_REPORT_ISSUE.md
```

.github/workflows/marker-report.yml
```diff
@@ -43,3 +68,54 @@ jobs:
         run: |
           . ./.venv/bin/activate
           python -m pytest tests/aws/
+
+        # makes use of the marker report plugin localstack.testing.pytest.marker_report
+      - name: Generate marker report
```
dominikschubert, line 73: Seems a bit redundant that we're doing this twice now (see `Collect marker report` above). Maybe only execute `Collect marker report` if `createIssue` is false could make sense? WDYT?

dominikschubert (approved):
LGTM, I agree that continuing here with just unifying the two workflows makes sense. Added a few minor nits that should ideally be addressed before merging, but nothing inherently blocking

.github/workflows/marker-report.yml
sannya-singal: This is something we can definitely think about. I was wondering what if we need an updated version of markers report in tinybird and also update/create a gh issue, the suggested implementation would mean that we need to run the workflow twice? We could maybe use two input params here, one to create github issue that we already have `createIssue` and another to collect the marker report in tinybird, what do you suggest?

dominikschubert: Generally the tinybird upload is meant to be running constantly (every commit on master) so we have a history to track, while the issue report is a manual (or potentially scheduled) run. So I think generally defaulting to tinybird upload, except when intentionally running it to create an issue would be fine

PR #11487, opened 2024-09-09
fix: apigateway S3 integration method

cyrilhagege:
- Retreive S3 action according to integration method defined and not http method
- fallback to http method if integration method not defined

## Motivation
When using S3 integration with ApiGateway, use correct S3 action depending on integration httpMethod

## Changes
- Retreive S3 action according to integration method defined and not http method
- fallback to http method if integration method not defined
-

### DIFF
--- a/localstack-core/localstack/services/apigateway/integration.py
+++ b/localstack-core/localstack/services/apigateway/integration.py
@@ -718,7 +718,8 @@ class S3Integration(BackendIntegration):
 
         action = None
         invoke_args = {"Bucket": bucket, "Key": object_key}
-        match invocation_context.method:
+        integration_method = integration.get("httpMethod") or invocation_context.method
+        match integration_method:
             case HTTPMethod.GET:
                 action = s3.get_object
             case HTTPMethod.PUT:


cyrilhagege:
recheck

cyrilhagege:
recheck

localstack-core/localstack/services/apigateway/integration.py
```diff
@@ -718,7 +718,11 @@ def invoke(self, invocation_context: ApiInvocationContext):
 
         action = None
         invoke_args = {"Bucket": bucket, "Key": object_key}
-        match invocation_context.method:
+        integration_method = integration.get("httpMethod") or invocation_context.method
+        if integration_method == "ANY":
+            integration_method = invocation_context.method
+
+        match integration_method:
```
bentsku, line 725:
```suggestion
        if not (integration_method := integration.get("httpMethod")) or integration_method == "ANY":
            integration_method = invocation_context.method

        match integration_method:
```

bentsku (commented):
Hello @cyrilhagege and thanks for your contribution! 

We usually try to add test cases to validate the fix following our [contribution guidelines.](https://github.com/localstack/localstack/blob/master/docs/CONTRIBUTING.md)

However, I know for a fact this is the correct behavior, and we are currently in the process of reworking our API Gateway invocation logic. It is actually ready for use for REST APIs, and you can give it a try by starting LocalStack with `PROVIDER_OVERRIDE_APIGATEWAY=next_gen` (please note that if you a Pro user, this will limit functionality related to HTTP and WebSockets API, but if you use the community LocalStack image `localstack/localstack:<tag>`, then you will have all the functionality existing and more).
So if you encounter more issues with the S3 Integration, you can try using the new implementation that will become default in our next major release.

I only have a minor comment so that we could simplify the logic a bit, but we can also merge as-is, I don't really mind! Let me know if you'd like to address the comment or rather let it that way.

Thanks a lot again for your contribution!

localstack-core/localstack/services/apigateway/integration.py
cyrilhagege: Thanks I will update according to this comment. I also noticed a missing implementation of requestParameters for integration uri, can I post this fix inside this PR?

```
    def apply_integration_request_parameters(uri: str, integration: Dict[str, Any], invocation_context: ApiInvocationContext):
        groups = re.findall(r"({[a-zA-Z]+})", uri)
        request_parameters = integration.get("requestParameters", {})
        for group in groups:
            key = group.replace("{", "").replace("}", "")
            request_param_key = f"integration.request.path.{key}"
            associatedValueKey = request_parameters.get(request_param_key)
            if associatedValueKey:
                value = invocation_context.context.get(associatedValueKey.replace("context.", ""))
                if (value):
                    uri = uri.replace(group, value)
        return uri

    #...

    uri = apply_integration_request_parameters(uri, integration, invocation_context)


```

cyrilhagege: If that's ok for you you can go ahead and merge :)

cyrilhagege:
Any news on merging this PR?

localstack-core/localstack/services/apigateway/context.py
```diff
@@ -80,7 +81,14 @@ def __init__(
         self._path = path
         self.data = data
         self.headers = headers
-        self.context = {"requestId": short_uid()} if context is None else context
+        self.context = (
+            {
+                "requestId": short_uid(),
+                "requestTimeEpoch": str(time.time()).replace(".", ""),
```
bentsku, line 87: Is this the right format? What is the precision of the epoch on AWS?

bentsku (commented):
Thanks again for the change! The new fix you added looks fine, but isn't tested. Would it be possible to add a test case for it (in ` tests/aws/services/apigateway/test_apigateway_s3.py`)? 

I'll add that maybe using the new implementation (using `PROVIDER_OVERRIDE_APIGATEWAY=next_gen` would probably solve your issue without trying to improve the current implementation which has many flaws. Did you give it a try?

localstack-core/localstack/services/apigateway/integration.py
```diff
@@ -1118,3 +1126,19 @@ def invoke(self, invocation_context: ApiInvocationContext):
...
+    request_parameters = integration.get("requestParameters", {})
+    for group in groups:
+        key = group.replace("{", "").replace("}", "")
+        request_param_key = f"integration.request.path.{key}"
+        associatedValueKey = request_parameters.get(request_param_key)
+        if associatedValueKey:
+            value = invocation_context.context.get(associatedValueKey.replace("context.", ""))
+            if value:
+                uri = uri.replace(group, value)
+    return uri
```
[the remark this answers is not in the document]
cloutierMat: Thank you @cyrilhagege for submitting this pr.

I agree with @bentsku, that it would be best to have the feature tested. Also, consider looking into `localstack.utils.json.extract_jsonpath`. This could help with the point raised by Ben.

I also think that the 2 problems at hand here can be solved by using `PROVIDER_OVERRIDE_APIGATEWAY=next_gen`

PR #11488, opened 2024-09-09
fixup and validate test_serverless.py

steffyP:
## Motivation
This is part of the initiative of removing the `markers.aws.unknown` from the tests.
The `test_serverless.py` is running a few tests using the `serverless` plugin. 

In the future, we should move these tests to the [serverless repo](https://github.com/localstack/serverless-localstack), but we still need to setup a suitable framework, ideally also allowing running snapshot tests (this is out of scope of this PR).

There is one test left that needs fixing `test_apigateway_deployed` where the assertion doesn't hold for AWS. 

## Changes
* added scripts to the `package.json` to run deployment against AWS
* fixed some definitions in the `serverless.yml`
  * bucket-name
  * added deletion-policy to make sure the stack is properly cleaned-up
* adapted the test cases to use the proper function/queue/bucket names
* cleanup s3 bucket before attempting deleting it

### DIFF
--- a/tests/aws/test_serverless.py
+++ b/tests/aws/test_serverless.py
@@ -30,7 +30,11 @@ class TestServerless:
         mkypatch.undo()
 
     @pytest.fixture(scope="class")
-    def setup_and_teardown(self, aws_client, region_name, delenv):
+    def get_deployed_stage(self):
+        return "dev" if is_aws_cloud() else "local"
+
+    @pytest.fixture(scope="class")
+    def setup_and_teardown(self, aws_client, region_name, delenv, get_deployed_stage):
         if not is_aws_cloud():
             delenv("AWS_PROFILE", raising=False)
         base_dir = get_base_dir()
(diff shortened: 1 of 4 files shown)


tests/aws/test_serverless.py
```diff
@@ -43,22 +47,40 @@ def setup_and_teardown(self, aws_client, region_name, delenv):
...
+                aws_client.s3.delete_objects(
+                    Bucket=bucket_name,
+                    Delete={"Objects": objects},
+                )
             # TODO the cleanup still fails due to inability to find ECR service in community
-            run(["npm", "run", "undeploy", "--", f"--region={region_name}"], cwd=base_dir)
+            if is_aws_cloud():
+                run(["npm", "run", "undeploy-aws", "--", f"--region={region_name}"], cwd=base_dir)
+            else:
+                run(["npm", "run", "undeploy", "--", f"--region={region_name}"], cwd=base_dir)
```
sannya-singal, line 78: nit: maybe a personal preference, we can write something like the following since only one argument is changing in the above case: 

```
command = "undeploy-aws" if is_aws_cloud() else "undeploy"
run(["npm", "run", command, "--", f"--region={region_name}"], cwd=base_dir)
```
or maybe also try running: 

`run(["npm", "run", "undeploy-aws" if is_aws_cloud() else "undeploy", "--", f"--region={region_name}"], cwd=base_dir)`

tests/aws/test_serverless.py
```diff
@@ -97,11 +121,13 @@ def test_dynamodb_stream_handler_deployed(self, aws_client, setup_and_teardown):
         assert event_source_arn == resp["Table"]["LatestStreamArn"]
 
     @markers.skip_offline
-    @markers.aws.unknown
+    @markers.aws.validated
     @pytest.mark.skip(reason="flaky")
```
sannya-singal, line 125: Q: Is this test still flaky?

sannya-singal (approved):
Thanks for fixing this Steffy! 🎉 Nice to see more validated tests. I just have one question and one minor comment but nothing blocking the merge.

tests/aws/test_serverless.py
steffyP: Don't know for sure - I only ran it a few times locally. But as there hasn't been any changes in kinesis lately, and the test was set as flaky in June I assume it's still the case.

Issue #10826, opened 2024-05-15
bug: Nested Stack not deleted when parent stack in ROLLBACK_COMPLETE status is deleted


lukasz-mazur2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Template contents:
```
$ cat modules-test-templates/test-module-cloudformation-alb-v1.0.0.yaml
AWSTemplateFormatVersion: '2010-09-09'
Parameters:
  VPC:
    Type: AWS::EC2::VPC::Id
    Default: vpc-17c17d56
  SubnetDMZA:
    Type: AWS::EC2::Subnet::Id
    Default: subnet-879e3ee0
  SubnetDMZB:
    Type: AWS::EC2::Subnet::Id
    Default: subnet-c819287c
Resources:
  LoadBalancer:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://test-standard-common-modules.s3.localhost.localstack.cloud/templates/cloudformation-alb/1/0/cloudformation-alb-v1.0.0.yaml
      Parameters:
        ResourceNamePrefix: system-team-it
        VPC:
          Ref: VPC
        Subnets:
          Fn::Sub: ${SubnetDMZA}, ${SubnetDMZB}
        DomainName: localstack.test.cloud
Outputs:
  ALBARN:
    Value:
      Fn::GetAtt:
        - LoadBalancer
        - Outputs.ARN
  ALBDNSName:
    Value:
      Fn::GetAtt:
        - LoadBalancer
        - Outputs.DNSName
  ALBSecurityGroupID:
    Value:
      Fn::GetAtt:
        - LoadBalancer
        - Outputs.SecurityGroupID
  ALBListener:
    Value:
      Fn::GetAtt:
        - LoadBalancer
        - Outputs.Listener
```

Deployment:
```
$ awslocal cloudformation deploy --stack-name lm-test --template-file modules-test-templates/test-module-cloudformation-alb-v1.0.0.yaml

Waiting for changeset to be created..
Waiting for stack create/update to complete

Failed to create/update the stack. Run the following command
to fetch the list of events leading up to the failure
aws cloudformation describe-stack-events --stack-name lm-test
```
Deployed stakcs:
```
$ awslocal cloudformation list-stacks --query "StackSummaries[].StackId" | grep lm
    "arn:aws:cloudformation:eu-west-1:000000000000:stack/lm-test/c6f89e25",
[...]

PR #11489, opened 2024-09-09
fix deletion of failed nested stack in CFn

pinzon:
## Motivation
Addresses #10826. The resource_provider class only stores the new model properties if the create operation is successful. This can cause issues where a nested stack reaches a failed status after being in a pending state. The Id property is never stored, and if the delete method is called, the property required for deletion cannot be found.

## Changes
- Now the resource_provider stores the model properties 

## Testing
- New aws validated test based on issue triaged.

### DIFF
--- a/tests/aws/services/cloudformation/api/test_nested_stacks.py
+++ b/tests/aws/services/cloudformation/api/test_nested_stacks.py
@@ -3,6 +3,7 @@ import os
 import pytest
 from botocore.exceptions import ClientError
 
+from localstack import config
 from localstack.testing.aws.util import is_aws_cloud
 from localstack.testing.pytest import markers
 from localstack.utils.files import load_file
@@ -294,7 +295,7 @@ def test_nested_stacks_conditions(deploy_cfn_template, s3_create_bucket, aws_cli
 
 
 @markers.aws.validated
-def test_deletion_of_failed_nested_stack(s3_create_bucket, aws_client):
+def test_deletion_of_failed_nested_stack(s3_create_bucket, aws_client, region_name):
     """
     This test confirms that after deleting a stack parent with a failed nested stack. The nested stack is also deleted
     """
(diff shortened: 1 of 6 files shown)


tests/aws/services/cloudformation/api/test_nested_stacks.py
```diff
@@ -81,7 +82,7 @@ def test_nested_stack_output_refs(deploy_cfn_template, s3_create_bucket, aws_cli
     assert f"{nested_bucket_name}-suffix" == result.outputs["CustomOutput"]
 
 
-@pytest.mark.skip(reason="Nested stacks don't work properly")
+@pytest.mark.skip(reason="Nested rtacks don't work properly")
```
simonrw, line 85:
```suggestion
@pytest.mark.skip(reason="Nested stacks don't work properly")
```

tests/aws/services/cloudformation/api/test_nested_stacks.py
```diff
@@ -291,3 +292,59 @@ def test_nested_stacks_conditions(deploy_cfn_template, s3_create_bucket, aws_cli
...
+    wait_until(_status_is_complete, strategy="static")
+
+    aws_client.cloudformation.delete_stack(StackName=stack_name)
+
+    def _stack_and_child_deleted():
+        stacks = aws_client.cloudformation.describe_stacks()["Stacks"]
+        stacks_with_name = [s for s in stacks if stack_name in s["StackName"]]
+        assert len(stacks_with_name) == 0
+
+    retry(_stack_and_child_deleted, sleep=2 if is_aws_cloud() else 1)
```
simonrw, line 350: Could you use a [waiter](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/waiter/StackDeleteComplete.html) here? Also it might be clearer to `describe_stack(StackName=stack_name)` and catch the exception?

simonrw (approved):
Nice consistency change, thanks for resolving! A couple of minor suggestions, but LGTM!

tests/aws/services/cloudformation/api/test_nested_stacks.py
pinzon: An error from my part but thanks to it I found that we can delete the skip

tests/aws/services/cloudformation/api/test_nested_stacks.py
pinzon: Took a while but test is a lot better now. I'll merge once is green.

Issue #11362, opened 2024-08-15
bug: Unable to create AWS::Serverless::Function via Cloud Formation Template


Sutty100:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When trying to deploy a Lambda function via Cloud Formation the stack ends up with the state CREATE_FAILED and the Stack status reason is:
> Unable to resolve Ref for resource "ServerlessDeploymentApplication" (yet)

### Expected Behavior

The Cloud Formation Stack is successfully deployed and a Lambda function is created

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Using the Test Containers Local Stack Java Library

```
        localStackContainer = new LocalStackContainer(DockerImageName.parse("localstack/localstack").withTag("3.6.0"))
                .withServices(SQS, CLOUDFORMATION, DYNAMODB, LAMBDA, API_GATEWAY, S3, IAM, CLOUDWATCHLOGS);
        localStackContainer.start(); 
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    Cloud Formation file attempting to be deployed is

The Cloud Formation file looks like this

```
---
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: AWS SAM Template for spring-lambda Lambda Function.
Parameters:
  Service:
    Type: String
  Env:
    Type: String
  CodeBucket:
    Type: String
  CodeFile:
    Type: String



Resources:
  Javafunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: !Sub "${Service}-${Env}"
      AutoPublishAlias: LIVE
      DeploymentPreference:
        Type: AllAtOnce
      Handler: com.foo.bar.poc.lambda.snapstart.SnapstartStreamLambdaHandler::handleRequest
      Runtime: java21
      Timeout: 60
      MemorySize: 1024
      Description: !Ref CodeFile
      Tracing: Active
      CodeUri:
        Bucket: !Ref CodeBucket
        Key: !Ref CodeFile
      Role: arn:aws:iam::000000000000:role/spring-lambda-mock-lambda-role
      Environment:
        Variables:
          ENV: !Sub "${Env}"
          JAVA_TOOL_OPTIONS: -XX:+TieredCompilation -XX:TieredStopAtLevel=1
[...]

PR #11496, opened 2024-09-10
allow reference unsupported cfn resources

pinzon:
## Motivation
Addresses #11362. LocalStack already allows the deployment of templates with unsupported resource types. (Check `CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES` in [docs](https://docs.localstack.cloud/references/configuration/)). With these changes any reference to these types returns a mock value.

## Changes
- Before raising DependencyNotYetSatisfied the template deployer check if the resource type is supported

## Testing
- New `only_localstack` test.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
@@ -85,6 +85,14 @@ def get_attr_from_model_instance(
     resource_id: str,
     attribute_sub_name: Optional[str] = None,
 ) -> str:
+    if resource["PhysicalResourceId"] == MOCK_REFERENCE:
+        LOG.warning(
+            "Attribute '%s' requested from unsupported resource with id %s",
+            attribute_name,
+            resource_id,
+        )
+        return MOCK_REFERENCE
+
     properties = resource.get("Properties", {})
     # if there's no entry in VALID_GETATT_PROPERTIES for the resource type we still default to "open" and accept anything
     valid_atts = VALID_GETATT_PROPERTIES.get(resource_type)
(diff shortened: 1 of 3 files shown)


tests/aws/templates/cfn_ref_unsupported.yml
```diff
@@ -0,0 +1,10 @@
+Resources:
+  CodeDeployApplication:
+    Type: AWS::CodeDeploy::Application
```
dominikschubert, line 3: nit: I'd suggest using something that we're sure won't be implemented in the future here instead, e.g. `AWS::LocalStack::Unknown`

dominikschubert (commented):
Can you maybe explain a bit the decision behind preferring to patch this in resolve_refs_recursively, rather than actually setting the physical resource id on the resource after the skipped deployment? 

I'd also like to see at least one more test here to show that other access patterns besides a Ref in the Outputs section works as well, e.g. wrapped in an intrinsic function in a resource properties section.

tests/aws/templates/cfn_ref_unsupported.yml
pinzon: Thanks for the suggestion. I extended the test to use an intrinsic function an a development. Also moved the patch outside the recursive function.

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1270,6 +1285,8 @@ def apply_change(self, change: ChangeConfig, stack: Stack) -> None:
                 resource_provider, resource, resource_provider_payload
             )
         else:
+            resource["PhysicalResourceId"] = MOCK_REFERENCE
+            resource["Ref"] = MOCK_REFERENCE
```
dominikschubert, line 1289: Why do we need to set the Ref here? :thinking:  Is this somehow overriding the physical resource id lookup if set?

pinzon: After some triage I found that the ref is not necessary, only the PhysicalResourceId is. And the assignment is after the search for the provider returns none. In the code you sent me it appeared that the assigment was inside the `try_load_resource_provider` method but this is not possible due to the function only receiving the resource type string.

dominikschubert (approved):
LGTM, thanks for addressing the comments. Let's see where this approach will take us. I hope we won't run into to many issues where users might expect certain patterns that the mock reference obviously won't adhere to :+1:

PR #11497, opened 2024-09-10
introduce plugins to load openapi specs

giograno:
## Motivation
This PR is a small improvement over https://github.com/localstack/localstack/pull/11452.

## Changes
- Added a plux plugin which is responsible for loading a spec given a certain path;
- The validator can work now on multiple specs. For instance, pro would have its spec and plugin and the validation will be performed against all the specs the plugin manager can find.

### DIFF
--- a/localstack-core/localstack/aws/handlers/validation.py
+++ b/localstack-core/localstack/aws/handlers/validation.py
@@ -69,21 +69,21 @@ class OpenAPIRequestValidator(OpenAPIValidator):
         path = context.request.path
 
         if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):
-            try:
-                for openapi in self.open_apis:
+            for openapi in self.open_apis:
+                try:
                     openapi.validate_request(WerkzeugOpenAPIRequest(context.request))
-            except RequestValidationError as e:
-                # Note: in this handler we only check validation errors, e.g., wrong body, missing required in the body.
-                response.status_code = 400
-                response.set_json({"error": "Bad Request", "message": str(e)})
-                chain.stop()
-            except OpenAPIError:
-                # Other errors can be raised when validating a request against the OpenAPI specification.
-                #   The most common are: ServerNotFound, OperationNotFound, or PathNotFound.
-                #   We explicitly do not check any other error but RequestValidationError ones.
-                #   We shallow the exception to avoid excessive logging (e.g., a lot of ServerNotFound), as the only
-                #   purpose of this handler is to check for request validation errors.
-                pass
+                except RequestValidationError as e:
+                    # Note: in this handler we only check validation errors, e.g., wrong body, missing required.
+                    response.status_code = 400
+                    response.set_json({"error": "Bad Request", "message": str(e)})
+                    chain.stop()
+                except OpenAPIError:
+                    # Other errors can be raised when validating a request against the OpenAPI specification.
+                    #   The most common are: ServerNotFound, OperationNotFound, or PathNotFound.
+                    #   We explicitly do not check any other error but RequestValidationError ones.
+                    #   We shallow the exception to avoid excessive logging (e.g., a lot of ServerNotFound), as the only
+                    #   purpose of this handler is to check for request validation errors.
+                    pass
 
 
 class OpenAPIResponseValidator(OpenAPIValidator):
@@ -96,14 +96,14 @@ class OpenAPIResponseValidator(OpenAPIValidator):
         path = context.request.path
 
         if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):
-            try:
-                for openapi in self.open_apis:
+            for openapi in self.open_apis:
+                try:
                     openapi.validate_response(
                         WerkzeugOpenAPIRequest(context.request),
                         WerkzeugOpenAPIResponse(response),
                     )
-            except ResponseValidationError as exc:
-                LOG.error("Response validation failed for %s: $s", path, exc)
-                response.status_code = 500
-                response.set_json({"error": exc.__class__.__name__, "message": str(exc)})
-                chain.stop()
+                except ResponseValidationError as exc:
+                    LOG.error("Response validation failed for %s: $s", path, exc)
+                    response.status_code = 500
+                    response.set_json({"error": exc.__class__.__name__, "message": str(exc)})
+                    chain.stop()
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -22,19 +23,23 @@
 
 LOG = logging.getLogger(__name__)
 
-
-# TODO: replace with from importlib.resources.files when https://github.com/python/importlib_resources/issues/311 is
-#   resolved. Import from a namespace package is broken when installing in editable mode.
+# todo: replace with importlib.resources, which currently seems to have problems with plux
```
alexrashed, line 26: nit: It would be great if you could share a bit more details (maybe in the PR description) what it means that "plux has problems with it"?

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -22,19 +23,23 @@
...
 oas_path = os.path.join(os.path.dirname(__file__), "..", "..", "openapi.yaml")
 
 
+def get_oas_spec() -> dict:
```
alexrashed, line 30: I think we really should try to avoid monkey patches as much as possible. Is there another way to implement this extension mechanism you are planning? Can we get a cleaner interface on the OpenAPI "spec loader"?

alexrashed (commented):
While the change itself isn't looking bad, if the whole purpose is the extensibility, I think we should try to define a cleaner way for the extension.
What's the plan for merging multiple specs in the future? We need a mechanism for handling multiple specs here in Community as well. Maybe it would make sense to have multiple providers (maybe implemented in plux) which can provide their specs?

localstack-core/localstack/aws/handlers/validation.py
giograno: As I was trying to import `importlib.resources` or the backport (`importlib_resources`) in this module I was getting errors like `error importing module localstack.aws.handlers.proxy: MultiplexedPath only supports directories` when making the entrypoints. E.g., in [this build](https://github.com/localstack/localstack/actions/runs/10800929263/job/29959988036?pr=11497) (create community endpoints step). I did not try to dig further tbh. cc @thrau

localstack-core/localstack/aws/handlers/validation.py
giograno: Thanks for the comment. As we briefly discussed, I reworked this with proper plugins. Core now has its plugin and pro will define its own. The validators can now work with multiple `openapi` instances to perform the validation.

localstack-core/localstack/plugins.py
```diff
@@ -21,3 +26,17 @@ def delete_cached_certificate():
...
+
+    def __init__(self, spec_path: os.PathLike | str) -> None:
+        if isinstance(spec_path, str):
+            spec_path = Path(spec_path)
+        self.spec_path = spec_path
+        self.spec = {}
+
+    def load(self):
+        with self.spec_path.open("r") as f:
+            self.spec = yaml.safe_load(f)
```
alexrashed, line 42: I think this plugin spec here is a bit confusing. It is only used in the validator for now, so I think the plugin definition should be there. The _registration_ of the _specific plugin_ (in this case the `CoreOASPlugin`) could be here though, as it's the `plugins.py` next to the actual spec (but could also stay where it is right now).

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -51,7 +56,8 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
 
         if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):
             try:
-                self.openapi.validate_request(WerkzeugOpenAPIRequest(context.request))
+                for openapi in self.open_apis:
```
alexrashed, line 59: q: I am curious what is the runtime implication of this is. Since this is executed on every request to the APIs (if enabled), it is time critical. How long does a check with one big "merged" spec take? And how long does it take if you iterate over the specs and perform the validation for each and every spec registered?

I am afraid that the overhead in the handler performing lots of different validations could be slow... Since we currently don't enable the validator by default, it's fine for now, but this could become an issue in the future...

localstack-core/localstack/aws/handlers/validation.py
alexrashed, line 1: In my opinion, this handler and the plugin definition deserves quite a lot more documentation. It's a cool new feature where only a very small amount of our contributors know about. Please make sure that everyone can take a look at this file and knows what it does, and how one can interact with it (enable it, disable it, register OpenAPI Spec plugins,...). ;)

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -77,10 +83,11 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
 
         if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):
             try:
-                self.openapi.validate_response(
-                    WerkzeugOpenAPIRequest(context.request),
-                    WerkzeugOpenAPIResponse(response),
-                )
+                for openapi in self.open_apis:
+                    openapi.validate_response(
```
alexrashed, line 87: It would be great if we could cover the "multi-spec validation" in some kind of tests. Dynamic plugins for testing aren't straight forward, but a simple unit test would suffice and would be fairly simple I guess?

alexrashed (commented):
Nice to see that the OpenAPI validator gets pluggable! In general the code looks good, but I would like to discuss the runtime implications of the plugin mechanism, and had some other nitpicks :P Afterwards we're good to go!

localstack-core/localstack/aws/handlers/validation.py
giograno: That's a fair question. I quickly tried to get some sense of the overhead, and we are talking about ~2 milliseconds to validate a request against the core spec. I suspect that most of the overhead to validate a spec is due to how large the spec is and how deeply nested its objects are. Therefore, I assume (but I can be wrong here) that validating against one large spec would not have a big impact here. In addition, we are only validating "internal" (`_localstack` and `_aws`) requests and we currently have only two specs.

localstack-core/localstack/aws/handlers/validation.py
thrau: i second alex' comment here and think it can do with even a bit more doc. specifically i think it needs a more concrete example on how to use it. where do i put the openapi.yml? how do i correctly instantiate a new plugin? For some inspiration, maybe the [docs on the WebAppExtension](https://github.com/localstack/localstack/blob/40165545d1127f61fb11652ef2854ed6c569b85c/localstack-core/localstack/extensions/patterns/webapp.py#L24-L71) can help.

thrau (approved):
Nice! LGTM :+1: just a minor nit regarding documentation

PR #11499, opened 2024-09-11
[SFN] Bump Antlr4 to 4.13.2

MEPalma:
## Motivation
The Step Functions v2 interpreter relies on the Antlr4 project for parsing state machine derivations. The version of the this tool is not updated automatically given it's reliance on the version of the parser generator program. These changes update both the generator and runtime libraries, as well as the generated files


joe4dev (approved):
No regressions in the test suite, so LGTM 👍

PR #11502, opened 2024-09-12
Add DNS server host mode advice

joe4dev:
## Motivation

We are missing instructions on disabling the DNS server in host mode leading to undocumented errors for new contributors.

## Changes

* Add instructions on how to disable the DNS server in host mode
* Update and clarify outdated TODO comments about missing EventBridge target tests

### DIFF
--- a/docs/development-environment-setup/README.md
+++ b/docs/development-environment-setup/README.md
@@ -53,6 +53,11 @@ In host mode, additional dependencies (e.g., Java) are required for developing c
 The required dependencies vary depending on the service, [Configuration](https://docs.localstack.cloud/references/configuration/), operating system, and system architecture (i.e., x86 vs ARM).
 Refer to our official [Dockerfile](https://github.com/localstack/localstack/blob/master/Dockerfile) and our [package installer LPM](Concepts/index.md#packages-and-installers) for more details.
 
+#### Root Permissions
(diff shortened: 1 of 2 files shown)


HarshCasper (approved):

simonrw:
It would be nice to add a note that _in general_ we should not disable the DNS server if possible. We want to discourage our users from disabling the DNS server, and if we make this suggestion without an explanation, both contributors and new developers will start to think that disabling the DNS server is normal. This will then leak out in suggestions to community users, or support cases.

Perhaps we should add something like:



> #### Root Permissions
> 
> LocalStack runs its own [DNS server](https://docs.localstack.cloud/user-guide/tools/dns-server/) which listens for requests on port 53. This requires root permission. When LocalStack starts in host mode it runs the DNS server as `sudo`, so a prompt is triggered asking for the sudo password. This is annoying during local development, so to disable this functionality, use `DNS_ADDRESS=0`.
> 
> !! We don't recommend using this setting in general, and when running in a Docker container the sudo password is not required. !!


instead?

joe4dev:
Good point @simonrw 

Adding a "because" sub-sentence to the disclaimer/recommendation would make it clearer as to why not.

Feel free to adjust it accordingly.

PR #11503, opened 2024-09-12
ESM-v2: Filter irrelevant fields before logging with PipesLogger

gregfurman:
## Motivation
Small rework of `PipesLogger's` message filtering to strip invalid fields from logs. This means we don't have to implement any pipes/esm specific checks when generating and passing around error payloads.

Fields derived from downstream snapshots and [this article](https://aws.amazon.com/blogs/compute/introducing-logging-support-for-amazon-eventbridge-pipes/).

## Changes
- Uses hardcoded log-fields to filter out irrelevant data from being logged. 

## Testing
- Backwards compatible changes so shoudn't break anything. Nevertheless, this was tested against pipes suite in pro on my local.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pipe_loggers/pipe_logger.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pipe_loggers/pipe_logger.py
@@ -72,34 +72,38 @@ class PipeLogger(ABC):
         ```
         """
         # https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-logs.html#eb-pipes-logs-execution-data
-        execution_data_fields = [
+        execution_data_fields = {
             "payload",
             "awsRequest",
             "awsResponse",
-        ]
-        fields_to_include = [
+        }
+        fields_to_include = {
             "resourceArn",
             "timestamp",
             "executionId",
             "messageType",
             "logLevel",
-        ]
-        error_fields_to_include = [
+        }
+        error_fields_to_include = {
             "message",
             "httpStatusCode",
             "awsService",
             "requestId",
             "exceptionType",
             "resourceArn",
-        ]
+        }
 
         if self.include_execution_data == [IncludeExecutionDataOption.ALL]:
-            fields_to_include.extend(execution_data_fields)
+            fields_to_include.update(execution_data_fields)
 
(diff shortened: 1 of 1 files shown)


joe4dev (approved):
Elegant approach without introducing boolean flags 💯 

LGTM 👍 

Adding the link from the PR description and a comment clarifying this change in the code would be helpful.

localstack-core/localstack/services/lambda_/event_source_mapping/pipe_loggers/pipe_logger.py
```diff
@@ -55,14 +55,31 @@ def filter_message(self, message: dict) -> dict:
...
+            "awsService",
+            "requestId",
+            "exceptionType",
+            "resourceArn",
+        ]
+
+        if self.include_execution_data == [IncludeExecutionDataOption.ALL]:
+            fields_to_include.extend(execution_data_fields)
+
+        filtered_message = {key: message[key] for key in fields_to_include if key in message}
```
dfangl, line 99:
```suggestion
        filtered_message = {key: value for key, value in message.items() if key in fields_to_include}
```
Could we do it this way? I think this is more straightforward to read, as it is pure filtering. By also converting the fields_to_include lists to sets, we could even reduce time complexity by a bit (although it honestly does not matter in this case).

gregfurman: Making the above sets makes sense. Just actually did a quick `timeit` (thinking it wouldn't make much of a difference) and it turns out it very much does even with so few elements 🫠:
```python
>>> import timeit
>>> timeit.timeit(stmt='-1 in a', setup='a = list(range(10))', number=1000)
0.0001385409850627184
>>> timeit.timeit(stmt='-1 in a', setup='a = set(range(10))', number=1000)
4.804192576557398e-05
```

I opted for the above approach since we are going to be reworking this filtering API in the coming weeks in anycase. Happy to make this more readable though.

dfangl: You are right, it definitely makes sense on that order of magnitude, I was talking more about the LS context, where it is nothing compared to some other overhead we see :) Still, we should always strive to increase performance)

dfangl (approved):
Love it, thanks for including my suggestions!

PR #11504, opened 2024-09-12
Feature/Events/Eventstudio: Only propagate trace_context

maxhoheiser:
## Motivation
To keep the overhead minimal, especially important for SQS and SNS we want to only propagate the minimal required information, therefore we use a DTO trace_context that links to only the necessary data of the context and not propagate the complete context.


maxhoheiser:
As agreed we are going with a threading.local approach

PR #11506, opened 2024-09-12
APIGW: patch UpdateRoute spec

cloutierMat:
## Motivation

While working on the apigw v2 provider, another update route was discovered that returns `201` instead of `200` as per the specs.

## Changes

- Updating spec for apigatewayv2 `UpdatedRoute` operation


bentsku (approved):
LGTM! I think it is maybe all `Update*` that returns 201. We should dedicate some time to look at all of them somehow

cloutierMat:
Unfortunately, we have already validated at least one that returns `200`. `UpdateIntegration` does return `200`.

PR #11509, opened 2024-09-12
feat: serve swagger UI in LocalStack

giograno:
## Motivation
We recently introduced OpenAPI specs for different LocalStack packages (`localstack` and `localstack.pro.core` at the moment).
These specs document all the internal developer endpoints we serve in LocalStack.
To allow users to navigate those specs, we want to serve the SwaggerUI directly within LocalStack.

## Changes
- Build some logic around our OAS plugin system (https://github.com/localstack/localstack/pull/11497) that collects at runtime all the available specs and merges them.
- Adding a new endpoint (`http://localhost.localstack.cloud:4566/_localstack/swagger`) that serves built OAS spec with a SwaggerUI. The UI is rendered with `jinja2` and lives in the `localstack.static` package.


localstack-core/localstack/http/resources/swagger/templates/index.html
```diff
@@ -0,0 +1,19 @@
...
+    <link rel="icon" type="image/png" href="/static/favicon-16x16.png" sizes="16x16" />
+  </head>
+
+  <body>
+    <div id="swagger-ui"></div>
+    <script src="/static/swagger-ui-bundle.js" charset="UTF-8"> </script>
+    <script src="/static/swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
+    <script src="/_localstack/swagger-initializer.js" charset="UTF-8"> </script>
+  </body>
+</html>
```
viren-nadkarni, line 19: Consider using [swagger-ui-dist](https://www.npmjs.com/package/swagger-ui-dist) which will eliminate the need to dump the entire swagger UI code in our codebase

https://github.com/swagger-api/swagger-ui/blob/HEAD/docs/usage/installation.md#unpkg

giograno: Thanks 🙌 I overlooked this!

giograno:
@thrau I opted to create a `swagger` package in `localstack.http.resources`. We might also later move `internals.py` to `localstack.http.resources` if you agree (it requires some coordination with ext). What do you think?
Thanks to the suggestion of @viren-nadkarni we could get rid of the `static` package at all.

thrau:
> @thrau I opted to create a `swagger` package in `localstack.http.resources`. We might also later move `internals.py` to `localstack.http.resources` if you agree (it requires some coordination with ext). What do you think? Thanks to the suggestion of @viren-nadkarni we could get rid of the `static` package at all.

i think this is good for now! that said, the http package was originally really meant as the package for our http framework library. anything _concrete_ such as internal endpoints, should actually be part of some services/application package, which is currently only `localstack.services`. so i think ultimately we're missing something, but for now `localstack.http` seems fine to me, since your code is enabling more concrete things!

thrau (approved):
nice! looking forward to seeing this in action :-)

Issue #11510, opened 2024-09-13
Cannot update UsagePlan


artaban:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when updating stack with **UsagePlan** of **ApiGatewayV1**, following error is thrown
`Unexpected error processing resource type AWS::ApiGateway::UsagePlan: Exception: string indices must be integers, not 'str'

### Expected Behavior

the stack is updated

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

`cdklocal deploy` on stack with ApiGatewayV1

### Environment

```markdown
- OS: MacOS 14.6.1
- LocalStack:
  LocalStack version:3.7.3.dev14
  LocalStack build date: 2024-09-12
  LocalStack build git hash: 457a4875
```

artaban:
verified using latest docker image and is not happening anymore

PR #11511, opened 2024-09-13
fix APIGW CFN UsagePlan update operation with Tags

artaban:
Fix for the issue #11510 

## Changes
- Fixed Tags update in **ApiGatewayV1 UsagePlan**

## Testing
- **UsagePlan** tags are updated correctly

## TODO
- not sure if the test is working / testing the change. Has not been able to perform cross-check on my environment.

### DIFF
--- a/localstack-core/localstack/services/apigateway/provider.py
+++ b/localstack-core/localstack/services/apigateway/provider.py
@@ -119,6 +119,7 @@ from localstack.services.apigateway.router_asf import ApigatewayRouter, to_invoc
 from localstack.services.edge import ROUTER
 from localstack.services.moto import call_moto, call_moto_with_request
 from localstack.services.plugins import ServiceLifecycleHook
+from localstack.utils.aws.arns import get_partition
 from localstack.utils.collections import (
     DelSafeDict,
     PaginatedList,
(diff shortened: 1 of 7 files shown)


artaban:

bentsku (commented):
Hi @artaban and thanks a lot for the fix and contribution!

The fix looks good, but the test will fail as you would need to regenerate the snapshots taken against AWS. Also, I don't think that is where we need to test this, as the issue is related to CloudFormation, and the test case added is testing directly the operation `UpdateUsagePlan`. 

There is however a specific test for that in `tests.aws.services.cloudformation.resources.test_apigateway.test_update_usage_plan` that fits your need.

You can modify the CloudFormation template to add tags, and validate the fix there. The template lives in `tests/aws/templates/apigateway_usage_plan.yml`. 

If regenerating the snapshots against AWS is an issue, I'll update them once you update the test. Thank you again!

bentsku:
Hey @artaban, thanks a lot for adding it to the template! I've pushed some changes to validate the test against AWS and came across an issue in our implementation where we actually did not call `tag_resource` to update the tags. I've fixed it as our current implementation was not doing what it was supposed to do, I hope that's alright for you?

artaban:
Hi @bentsku, thank you for fixing this

bentsku (approved):
I'll wait for @cloutierMat to also approve the PR as I did some work, for a quick sanity check, but thank you so much for your contribution @artaban and digging quickly into the code base to fix your own issue! Welcome to the list of LocalStack's contributors!

cloutierMat (approved):
LGTM! Thanks @artaban for your contribution

PR #11505, opened 2024-09-12
add: Support for ESM v2 partial batch failure handling (Kinesis & DynamoDB)

gregfurman:
## Motivation
This PR adds support for `FunctionResponseTypes` to ESM v2, allowing for partial batch failures to be reported and handled. Failed items of a batch can now be retried in accordance with a `MaximumRetryAttempts` policy.

### Event Source Mapping: SQS
- Fixed failing test at `tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py::test_report_batch_item_failures_invalid_result_json_batch_fails` by propagating error and payload information to the raised `PartialBatchFailure`.

### Event Source Mapping: Kinesis & DynamoDB
- ESM v2 supports setting `FunctionResponseTypes` to `[ReportBatchItemFailures]` and handling partial batch failures when `batchItemFailures` is correctly returned by Lambda invocation via Kinesis or DynamoDB ESM.

### Partial Failure with ReportBatchItemFailures 
Both tests simulates a partial batch failure, capturing failure information in an `OnFailure` destination config using a DLQ:
- DynamoDB: Added `TestDynamoDBEventSourceMapping::test_dynamodb_report_batch_item_failures`
- Kinesis: Added `TestKinesisSource::test_kinesis_report_batch_item_failures` which simulates a batch failure and captures failure information in a DLQ

### Success and failure conditions
All success and failure cases outlined in the [ESM docs](https://docs.aws.amazon.com/lambda/latest/dg/services-ddb-batchfailurereporting.html#streams-batchfailurereporting-conditions) have test coverage (where conditions are identical for all Kinesis, DynamoDB, and SQS). These are covered in the following parametrized tests:
#### Total Batch Successes:
- `test_kinesis_report_batch_item_success_scenarios`
- `test_dynamodb_report_batch_item_success_scenarios`
#### Total Batch Failures:
- `test_kinesis_report_batch_item_failure_scenarios`
- `test_dynamodb_report_batch_item_failure_scenarios`

## TODO

What's left to do in follow-up PRs:

- [ ] Include a test case for when a sequence number is not present in a batch (should be a total failure)
- [ ] Figure out why Kinesis and DynamoDB have an extra invocation (seen in CloudWatch logs)

PR #11513, opened 2024-09-13
fix: Extraction of shard ID in ESMv2 StreamPoller

gregfurman:
## Motivation

The `shardId` of DynamoDB Streams events is incorrectly being set (when sending failed records to a failure destination config) since the `shard_id` extraction from `eventID` assumes the Kinesis format:
```python
# DynamoDB eventID example - "eventID": "f07f8ca4b0b26cb9c4e5e77e69f274ee"
shard_id = first_record["eventID"].split(":")[0] 
```

See [AWS ESM Kinesis example](https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-event-example):
```
"eventID": "shardId-<timestamp>:<sequence-number>"
```

See [AWS ESM DynamoDB Streams example](https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html#events-sample-dynamodb):
```
"eventID": "<uuid>"
```

Note: This unblocks the failing CI in https://github.com/localstack/localstack/pull/11505

## Changes
- Pass the `shard_id` of the current shard iteration instead of relying on extracting this value from a record's `eventID`

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -227,7 +227,7 @@ class StreamPoller(Poller):
             attempts_count=attempts,
             partner_resource_arn=self.partner_resource_arn,
         )
-        self.send_events_to_dlq(events, context=failure_context)
+        self.send_events_to_dlq(shard_id, events, context=failure_context)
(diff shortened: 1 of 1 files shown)


joe4dev (approved):
Thanks for fixing my bug 🙇‍♂️ 

Much cleaner than trying to peek into the first record 🙈

PR #11514, opened 2024-09-13
fixup if-evaluation for 'Create GH issue from template'

sannya-singal:
## Motivation
Fixes: https://github.com/localstack/localstack/pull/11485 
In favor of: https://github.com/localstack/localstack-ext/pull/3441

## Changes
As proposed by Steffy in the `-ext` PR: This PR tries to fix this issue by explicitly checking for `true` value.


dominikschubert (approved):
LGTM, thanks for jumping on the fix

PR #11518, opened 2024-09-14
APIGW NG: fix leftover REST issues

bentsku:
## Motivation

Fixing leftover APIGW NextGen tasks. Realized special case for `execute-api` style `AWS` integration `uri`, where we don't need to pass an authorization header, as those requests aren't signed by default and would fail with a bad service name. 

Also fixed some error handling related to Lambda, seemingly the `AWS` Lambda integration follows specific rules (as it already does with the `selectionPattern` specific behavior. 

## Changes
- only add mocked headers if the request targets a service operation
- remove the default exception message for `AuthorizerConfigurationError`
- add specific error handling for `AWS` lambda integration that is very specific to this one integration, not returning the `AccessDeniedError` unlike all other like `dynamodb`, `kinesis` and `sns`
- remove the `asynchronous` logic for `AWS_PROXY`, this is only possible for `AWS` Lambda (logical, as headers are not used to create the lambda invocation)
- added logic for `AWS_PROXY` which changes the status code raised if the error is `AccessDenied`
- fix `AWS` to properly add the `Action` query string parameter

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
@@ -174,6 +174,7 @@ class RestApiAwsIntegration(RestApiIntegration):
     def __init__(self):
         self._base_domain = config.internal_service_url()
         self._base_host = ""
+        self._service_names = get_service_catalog().service_names
 
     def invoke(self, context: RestApiInvocationContext) -> EndpointResponse:
         integration_req: IntegrationRequest = context.integration_request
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/aws.py
```diff
@@ -223,7 +228,7 @@ def invoke(self, context: RestApiInvocationContext) -> EndpointResponse:
         request_parameters = {
             "method": method,
             "url": url,
-            "params": integration_req["query_string_parameters"],
+            "params": query_params,
```
cloutierMat, line 231: Good catch! :eyes:

cloutierMat (approved):
All good fixes thanks for tackling these! :pray:

PR #11525, opened 2024-09-16
ESM-v2: Add persistence to Lambda event source mapping

gregfurman:
## Motivation
This PR aims to add persistence support to Event Source Mapping V2. 

## Changes
- Add to the Lambda provider's `on_after_state_load` method.
- Poll whether the mapped Lambda is in a ready state (`Active/Failed`). Log a warning if not ready and proceed.
- If ESM is in a disabled/disabling state on re-load, then set the `enabled` flag of the ESM worker to false. Otherwise, create the ESM as normal.
- Left a TODO about polling for a ready event source.

## Testing
- Downstream persistence tests for Lambda should pass.

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -326,9 +326,39 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                                 exc_info=True,
                             )
 
-                # Restore event source listeners
                 for esm in state.event_source_mappings.values():
-                    EventSourceListener.start_listeners_for_asf(esm, self.lambda_service)
+                    if config.LAMBDA_EVENT_SOURCE_MAPPING == "v2":
+                        # Restores event source workers
+                        function_arn = esm.get("FunctionArn")
+
+                        # TODO: How do we know the event source is up?
+                        # A basic poll to see if the mapped Lambda function is active/failed
+                        if not poll_condition(
+                            lambda: get_function_version_from_arn(function_arn).config.state.state
+                            in [State.Active, State.Failed],
+                            timeout=10,
+                        ):
+                            LOG.warning(
+                                "Creating ESM for Lambda that is not in running state: %s",
+                                function_arn,
+                            )
+
+                        function_version = get_function_version_from_arn(function_arn)
+                        function_role = function_version.config.role
+
+                        is_esm_enabled = esm.get("State", EsmState.DISABLED) not in (
+                            EsmState.DISABLED,
+                            EsmState.DISABLING,
+                        )
+                        esm_worker = EsmWorkerFactory(
+                            esm, function_role, is_esm_enabled
+                        ).get_esm_worker()
+
+                        # Note: a worker is created in the DISABLED state if not enabled
+                        esm_worker.create()
+                    else:
+                        # Restore event source listeners
+                        EventSourceListener.start_listeners_for_asf(esm, self.lambda_service)
 
     def on_after_init(self):
         self.router.register_routes()
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
@@ -86,6 +86,7 @@ class EsmWorker:
             with self._state_lock:
                 self.current_state = EsmState.DISABLED
                 self.state_transition_reason = self.user_state_reason
+            self.update_esm_state_in_store(EsmState.DISABLED)
 
     def start(self):
         with self._state_lock:


localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -326,9 +326,39 @@ def on_after_state_load(self):
...
+                        # A basic poll to see if the mapped Lambda function is active/failed
+                        if not poll_condition(
+                            lambda: get_function_version_from_arn(function_arn).config.state.state
+                            in [State.Active, State.Failed],
+                            timeout=10,
+                        ):
+                            LOG.warning(
+                                "Creating ESM for Lambda that is not in running state: %s",
+                                function_arn,
+                            )
```
dfangl, line 344: What would happen if the warning is printed? 10s might be rather short if LocalStack starts polling a new lambda image, but we also do not want to block here for two long. Will the first invocations just fail and be retried?

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -326,9 +326,39 @@ def on_after_state_load(self):
...
+
+                        function_version = get_function_version_from_arn(function_arn)
+                        function_role = function_version.config.role
+
+                        is_esm_enabled = esm.get("State", EsmState.DISABLED) not in (
+                            EsmState.DISABLED,
+                            EsmState.DISABLING,
```
dfangl, line 351: If the state is DISABLING when loading, we never set it to be disabled, even though the ESM worker is created correctly. That can lead to issues with the state later on, could we correct that?

dfangl (commented):
Looks good, just some questions regarding the waiting and restoring ESMs in the "disabling" state.

localstack-core/localstack/services/lambda_/provider.py
gregfurman: > What would happen if the warning is printed?

An ESM would be created pointing to a Lambda that is still in the `Pending` or `Inactive` state.

> 10s might be rather short if LocalStack starts polling a new lambda image, but we also do not want to block here for two long.

I am skeptical that we would encounter a polling issue here very often since there is a check that happens prior to ESMs being re-created:

https://github.com/localstack/localstack/blob/650483e43779c68f104009aa01b1d54f11c9cd6c/localstack-core/localstack/services/lambda_/provider.py#L292-L300

> Will the first invocations just fail and be retried?

I believe that the function invoke request in the `LambdaSender` (and configured in the `EsmWorker`) would return a `500` since the function is not in an active state. From the `EsmWorker` perspective, I don't think we retry invocations if there is a fail:

https://github.com/localstack/localstack/blob/650483e43779c68f104009aa01b1d54f11c9cd6c/localstack-core/localstack/services/lambda_/event_source_mapping/senders/lambda_sender.py#L45-L50

localstack-core/localstack/services/lambda_/provider.py
gregfurman: Ahh that is a good point. Let me just ensure we set the state to `DISABLED` if we encounter a `DISABLING` state on startup.

I took this approach since I assumed there was some async cleanups/actions that a worker would take prior to setting the state to `DISABLED`.

localstack-core/localstack/services/lambda_/provider.py
dfangl: Ah, I see, although an additional 5s won't make much of a difference, pulling the image can take quite some time. Personally, I am in favor of handling that issue later in the ESM worker itself, and retrying there, instead of blocking the on_state_restore call for too long. We could for example set the ESM to `Enabling` until the function is active, or something like that.

localstack-core/localstack/services/lambda_/provider.py
dfangl: The internal state of the EsmWorker is fine, but we never update the state in the store, right? So, the state in the store will be `DISABLING` indefinitely.

gregfurman: Ahh I see. Will change this in the store

localstack-core/localstack/services/lambda_/provider.py
gregfurman: Agree with the above. Happy to do this in a follow-up PR or should we introduce this more robust function active checker alongside these changes?

localstack-core/localstack/services/lambda_/provider.py
gregfurman: Gave this another look. 

Appears we do check if the state is `DISABLING` in the `poller_loop`, where that `update_esm_state_in_store` updates the ESM state store `state.event_source_mappings`

https://github.com/localstack/localstack/blob/9b3530f8d2a996a3eebf6b4389812beef8f63129/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py#L149-L153

dfangl: I see that, but since we set `self.enabled` to `False`, and the .create method has this branch: https://github.com/localstack/localstack/blob/9b3530f8d2a996a3eebf6b4389812beef8f63129/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py#L84-L88

The poller loop never gets executed, and therefore the code you mentioned never called.

gregfurman: A-ha! That makes sense. Will need to explicitly update this then

gregfurman: I've added a call to the `update_esm_state_in_store` method in the `EsmWorker` class. This way, we actually set the state to `DISABLED` in the store when `create()` is called:

https://github.com/localstack/localstack/blob/ecf24ecab704417abadebe6ca158de6e1f18f52e/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py#L151-L154

dfangl (approved):
LGTM! Let's tackle the waiting thing in a follow up :)

PR #11533, opened 2024-09-17
Feature/Eventstudio/Events: use thread local for trace_context propagation

maxhoheiser:
## Motivation
For EventStudio we introduced thread.local based propagation of trace_context, thus the context is not needed in the patched functions anymore.
Furthermore, a copy of a event is sent to target_sender since the same base event can be sent to multiple targets and modification further down the chain can happen. 
Also we handle capturing of transformation steps also inside the extension now, so the proxy_send_event is also no longer needed.

## Changes
- remove context input to patched methods
- decompose process_events
- add function to process single event from input events
- add proxy to capture event if no rule is attached
- add function to process individual rule

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -1130,7 +1130,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
 
                 target_sender = self._target_sender_store[target["Arn"]]
                 try:
-                    target_sender.process_event(event)
+                    target_sender.process_event(event.copy())
                 except Exception as e:
                     LOG.info(
                         "Unable to send event notification %s to target %s: %s",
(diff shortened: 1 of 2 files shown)


Morijarti (approved):
Nice work on decoupling things and extracting methods.

I've tested in in my locale and it works perfectly.

localstack-core/localstack/services/events/provider.py
```diff
@@ -1132,7 +1130,7 @@ def func(*args, **kwargs):
 
                 target_sender = self._target_sender_store[target["Arn"]]
                 try:
-                    target_sender.process_event(event, context)
+                    target_sender.process_event(event.copy())
```
joe4dev, line 1133: Can we add a comment in the code clarifying why and wherever `.copy()` is needed?
If EventStudio depends on specific behavior, we should explicitly add comments so that removing the particular code doesn't break It.

Issue #11319, opened 2024-08-06
bug: secret manager always returning base64 encoded string even in GoSDK


MathiasVandePol:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the Go aws SDK, It seems that localstack somehow always returns a base64 encoded secretBinary even though this should not happen for the GoSDK (and others besides Python?)

https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#API_GetSecretValue_ResponseSyntax

> The decrypted secret value, if the secret value was originally provided as binary data in the form of a byte array. When you retrieve a SecretBinary using the HTTP API, the Python SDK, or the AWS CLI, the value is Base64-encoded. Otherwise, it is not encoded.

### Expected Behavior

Return a non base64 encoded binary

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Below is some go code that works if the awsEndpoint is not overriden to localstack.

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
)

func main() {
	// $ touch secrets.json
	// $ cat > secrets.json << EOF
	// {
	//   "username": "admin",
	//   "password": "password"
	// }
	// EOF
	// $ awslocal secretsmanager create-secret --name xxx --secret-binary fileb://secrets.json

	secretName := "xxx"

	awsEndpoint := "http://localhost:4566"
	awsRegion := "us-east-1"

	awsCfg, err := config.LoadDefaultConfig(context.TODO(),
		config.WithRegion(awsRegion),
	)
	if err != nil {
		log.Fatal(err)
	}

	// Create Secrets Manager client
	svc := secretsmanager.NewFromConfig(awsCfg, func(o *secretsmanager.Options) {
		o.BaseEndpoint = aws.String(awsEndpoint)
	})

	input := &secretsmanager.GetSecretValueInput{
		SecretId: aws.String(secretName),
	}

	result, err := svc.GetSecretValue(context.TODO(), input)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%s", result.SecretBinary) // This should print the json but it prints a base64 encoded string
}

```

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
[...]

PR #11535, opened 2024-09-18
Fix: Make secret value decoded for specific usecases

faiyaz26:
## Motivation
With this PR I am trying to fix the bug mentioned on this issue : https://github.com/localstack/localstack/issues/11319

## Changes
I have fetched request from different SDK and HTTP Api, it seems the behaviour is return decoded value when requested from SDK. So added the change, to make the SecretBinary decoded all time.

## Testing
1. Added a test to fetch a secret when the SecretBinary should not be decoded
2. Added a test to fetch secret as http api request, where the SecretBinary should be decoded.

### DIFF
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -2397,7 +2397,7 @@ class TestSecretsManager:
         sm_snapshot.match("mismatch_version_id_and_stage", exc.value.response)
 
     @markers.aws.validated
-    def test_get_secret_value_to_fix_11319(
+    def test_get_secret_value(
         self, aws_client, aws_http_client_factory, region_name, create_secret, sm_snapshot
     ):
         """
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -2396,7 +2396,7 @@ class TestSecretsManager:
             )
         sm_snapshot.match("mismatch_version_id_and_stage", exc.value.response)
 
-    @markers.aws.unknown
+    @markers.aws.validated
     def test_get_secret_value(self, aws_client, create_secret, sm_snapshot):
         secret_name = short_uid()
         secret_string = b"footest"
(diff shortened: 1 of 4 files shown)


faiyaz26:

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -2395,6 +2396,48 @@ def test_get_secret_value_errors(self, aws_client, create_secret, sm_snapshot):
...
+
+        client = aws_http_client_factory(
+            "secretsmanager", region=region_name, signer_factory=SigV4Auth
+        )
+        parameters = {"SecretId": secret_name}
+
+        headers = {
+            "X-Amz-Target": "secretsmanager.GetSecretValue",
+            "Content-Type": "application/x-amz-json-1.1",
+        }
```
dominikschubert, line 2433: We should also add a test here to cover the previous case, basically the request that boto would usually send. This would make the particular difference between SDKs more obvious here.

dominikschubert (changes requested):
Hi @faiyaz26 ! Thanks for your contribution!

I had a few remarks regarding the usage of the snapshot and adding additional information in the code/test itself. Let me know if you need any clarifications for the review comments!

tests/aws/services/secretsmanager/test_secretsmanager.py
faiyaz26: there are two snapshot assertions in this one is testing in boto, another one for http request.

dominikschubert: Sorry, I actually misunderstood the issue of the PR initially. It seems the Go SDK was quite a red herring actually in the initial report. All the clients are broken currently and return the wrong values when used with SecretBinary!

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -2395,6 +2396,56 @@ def test_get_secret_value_errors(self, aws_client, create_secret, sm_snapshot):
             )
         sm_snapshot.match("mismatch_version_id_and_stage", exc.value.response)
 
+    @markers.aws.validated
+    def test_get_secret_value_to_fix_11319(
```
dominikschubert, line 2400:
```suggestion
    def test_get_binary_secret_value(
```
nit: Don't need to specify the issue in here

dominikschubert (approved):
LGTM! This was definitely a tricky issue because of the misleading AWS documentation and also the a bit misleading reported issue. Thanks for your contribution!

Issue #11534, opened 2024-09-17
bug: s3 PutObject hangs after requesting invalid range with GetObject


sameer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following sequence hangs on the last step:

1. Create a bucket `foo` (CreateBucket)
2. Upload a 500 byte object `bar` (PutObject)
3. Try to get an invalid range for `bar` i.e. `bytes=544-`, which returns InvalidRange as expected (GetObject)
4. Upload `bar` again with the same data, this times out (PutObject)

Step 3 causes step 4 to fail, but I'm not sure why.

### Expected Behavior

Step 4 should succeed.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

`SERVICES=s3 LS_LOG=trace localstack start `

^ I have confirmed that it still happens without the `SERVICES` env var set.

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
aws s3 mb s3://foo
dd if=/dev/zero of=bar bs=500 count=1 # make a 500 byte file
aws s3 cp bar s3://foo
aws s3api get-object --bucket foo --key bar --range 'bytes=544-' outfile # this fails, as expected
aws s3 cp bar s3://foo # this hangs and will eventually time out
```

### Environment

```markdown
- OS: macOS Sonoma 14.6.1
- LocalStack:
  LocalStack version: 3.7.3.dev28
  LocalStack build date: 2024-09-16
  LocalStack build git hash: dadf66461
```

### Anything else?

Based on the trace logs, the request never returns. I wonder if there's a deadlock somewhere.

bentsku:
Hey @sameer, thank you so much for your report! The linked PR will fix the issue, once the PR is merged, the `latest` version of LocalStack should contain the fix in the hour or so after. Thanks again!

PR #11536, opened 2024-09-18
fix S3 GetObject deadlock when using invalid Range

bentsku:
## Motivation
As reported in #11534, we entered a deadlock when calling `GetObject` with an invalid range, as we never released the lock that we purposefully do not release until the object has been released. However, there was an issue as we added validation after acquiring the lock, leading to LocalStack not returning the object, and the web server not calling `.close` on the S3 object. 

If in the future we need to add validation after acquiring the lock, we need to make sure to maybe wrap everything in a `try...except` block with the `return` in the `try` block, and release the lock in the `except` block. For now, this still seems safe to do how we are doing it, as if we have exceptions here, those should be fixed instead of putting tape over them, and a deadlock might not be the biggest issue. 

## Changes
- update the validation to happen before acquiring the lock, thus removing the issue, and added a small test case to avoid regression

_fixes #11534_

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -842,8 +842,20 @@ class S3Provider(S3Api, ServiceLifecycleHook):
 
         validate_failed_precondition(request, s3_object.last_modified, s3_object.etag)
 
+        range_header = request.get("Range")
+        part_number = request.get("PartNumber")
+        if range_header and part_number:
+            raise InvalidRequest("Cannot specify both Range header and partNumber query parameter")
+        range_data = None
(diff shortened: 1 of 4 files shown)


cloutierMat (approved):
Good fix @bentsku!

Thank you for taking the time for this expedited resolution!

Have you considered creating context using `with`? Not in the scope of this fix of course, but if implemented it would prevent other similar issues in the future.

cloutierMat:
I see, thanks for the clarification

bentsku:
We should maybe have better documentation around all of this, as this is quite implicit. Thanks for the comment and review, I'll think about improving it!

PR #11539, opened 2024-09-18
Fix: CFn: mappings with references

simonrw:
## Motivation

A customer presented an error which involved a `FindInMap` in a `Condition`. The `FindInMap` uses `Ref`s for its map and top level keys, which raised an error:

```
    return mappings[map_name][top_level_key][second_level_key]
           ~~~~~~~~^^^^^^^^^^
TypeError: unhashable type: 'dict'
```

## Changes

* Implement the fix by resolving the ref from the template parameters
* Improve error handling on invalid input


localstack-core/localstack/services/cloudformation/engine/template_utils.py
pinzon, line 1: It's good that our template deployer now supports many CloudFormation features. We should consider refactoring this function to improve readability and maintainability.

tests/aws/services/cloudformation/engine/test_mappings.py
```diff
@@ -153,3 +153,23 @@ def test_mapping_minimum_nesting_depth(self, aws_client, cleanups, snapshot):
...
+                "MapKey": "A",
+                "TopicName": topic_name,
+            },
+        )
+
+        topic_arn = stack.outputs.get("TopicArn")
+        assert topic_arn is not None
+
+        describe_res = aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
+        snapshot.match("topic-attributes", describe_res)
```
pinzon, line 175: Nit: These lines seem unnecessary for the test. Maybe a positive and a negative deployment could have more value in this situation .

pinzon (approved):
LGTM. Just some nits.

localstack-core/localstack/services/cloudformation/engine/template_utils.py
simonrw: Definitely, if we are going to do the same level of input valuation and error reporting that this one does

tests/aws/services/cloudformation/engine/test_mappings.py
simonrw: Good point, thanks!

PR #11549, opened 2024-09-19
fix logs filtering for state machine arn

pinzon:


pinzon:
The error is partially fix in our PRO version so I'm closing this PR in favor of one in the correct repo.

Issue #11494, opened 2024-09-10
bug: Stepfunctions fails with RuntimeError(The JSONPath $.Payload could not be found in the input....) on 'Catch' step


Relenna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Good afternoon, 
It looks like something wrong is going with stepfunctions and its `Catch` block. 
If `Catch` block leads to `Task` (at least), then execution of step function fails with:

`2024-09-10T13:38:36.152 ERROR --- [ad-19 (eval)] l.s.s.a.c.program.program  : Stepfunctions computation ended with exception 'RuntimeError(The JSONPath $.Payload could not be found in the input {"throwError": true, "error": {"Error": "Exception", "Cause": "{\"errorMessage\":\"Test exception\",\"errorType\":\"Exception\",\"requestId\":\"bc195d7e-2ee9-4c5c-b275-3398686663ad\",\"stackTrace\":[\"  File \\\"/var/task/test_lambda.py\\\", line 7, in lambda_handler\\n    raise Exception(f'Test exception')\\n\"]}"}})'.`

Step function example:
```
{
  "Comment": "A description of my state machine",
  "StartAt": "TestLambda",
  "States": {
    "TestLambda": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "OutputPath": "$.Payload",
      "Parameters": {
        "FunctionName": "arn:aws:lambda:us-east-1:000000000000:function:test-lambda",
        "Payload": {
          "throwError.$": "$.throwError"
        }
      },
      "Retry": [
        {
          "ErrorEquals": [
            "States.ALL"
          ],
          "IntervalSeconds": 1,
          "MaxAttempts": 3,
          "BackoffRate": 2
        }
      ],
      "End": true,
      "Catch": [
        {
          "ErrorEquals": [
            "States.ALL"
          ],
          "Next": "OnError",
          "ResultPath": "$.error"
        }
      ]
    },
    "OnError": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "OutputPath": "$.Payload",
      "Parameters": {
        "Payload.$": "$",
        "FunctionName": "arn:aws:lambda:us-east-1:000000000000:function:on-error"
      },
      "Retry": [
        {
          "ErrorEquals": [
            "States.ALL"
          ],
          "IntervalSeconds": 1,
          "MaxAttempts": 3,
          "BackoffRate": 2
[...]

PR #11552, opened 2024-09-20
[SFN] No Output Reductions for Successful Catch Blocks

MEPalma:
## Motivation
When an error in a state is successfully handled by a catch block, the error object containing the error message and cause string is directly passed as the state's output. This means that if the user defines output reduction logic, such as ResultPath and OutputPath, these are not applied. However, the current implementation of the SFN v2 interpreter mistakenly applies these reductions, which can lead to JsonPath errors and disrupt the state machine's logic, as seen in issue #11494.
These changes type the output handling for successful catch blocks so these can be detected and bypass output reductions.

## Changes
- Typed CatcherDecl outputs to be CatcherOutput objects
- Add bypass logic in the state's evaluation of the output
- Added relevant snapshot tests which test the use of OutputPath and ResultPath as output reductions

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
@@ -178,7 +178,7 @@ class CommonStateField(EvalComponent, ABC):
         output = env.stack[-1]
 
         # CatcherOutputs (i.e. outputs of Catch blocks) are never subjects of output normalisers,
-        # the entire value is instead passed by value as input ot the next state, or program output.
+        # the entire value is instead passed by value as input to the next state, or program output.
         if isinstance(output, CatcherOutput):
             env.inp = copy.deepcopy(output)
         else:
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
@@ -81,7 +81,6 @@ class CommonStateField(EvalComponent, ABC):
         self.comment = None
         self.input_path = InputPath(InputPath.DEFAULT_PATH)
         self.output_path = OutputPath(OutputPath.DEFAULT_PATH)
-        self.output_path = OutputPath(OutputPath.DEFAULT_PATH)
         self.state_entered_event_type = state_entered_event_type
         self.state_exited_event_type = state_exited_event_type
 
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+import copy
 import datetime
 import json
 import logging
@@ -14,6 +15,7 @@ from localstack.aws.api.stepfunctions import (
     StateEnteredEventDetails,
     StateExitedEventDetails,
 )
+from localstack.services.stepfunctions.asl.component.common.catch.catcher_decl import CatcherOutput
 from localstack.services.stepfunctions.asl.component.common.comment import Comment
 from localstack.services.stepfunctions.asl.component.common.error_name.failure_event import (
     FailureEvent,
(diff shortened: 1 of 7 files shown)


tests/aws/services/stepfunctions/templates/errorhandling/statemachines/task_service_lambda_invoke_catch_all_output_path.json5
```diff
@@ -0,0 +1,36 @@
...
+    "ProcessResult": {
+      "Type": "Pass",
+      "End": true
+    },
+    "HandleGeneralError": {
+      "Type": "Pass",
+      "Parameters": {
+        "InputValue.$": "$"
+      },
+      "End": true
```
gregfurman, line 33: I stand corrected but shouldn't we be trying to test that the reduction `"OutputPath": "$.Payload"` gets ignored here since (as per that original GitHub issue)?

```json
...
      "Catch": [
        {
          "ErrorEquals": [
            "States.ALL"
          ],
          "Next": "OnError",
          "ResultPath": "$.error"
        }
      ]
    },
    "OnError": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "OutputPath": "$.Payload", #
...
```

Am aware that the above `Catch` allows the error and input to be preserved (from https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html#input-output-resultpath-catch)

gregfurman (approved):
LGTM! Just a question on simplifying the statemachine and whether we are including all necessary test-cases for these changes.

tests/aws/services/stepfunctions/templates/errorhandling/statemachines/task_service_lambda_invoke_catch_all_output_path.json5
```diff
@@ -0,0 +1,36 @@
...
+          ],
+          "Next": "HandleGeneralError"
+        }
+      ],
+      "Next": "ProcessResult"
+    },
+    "ProcessResult": {
+      "Type": "Pass",
+      "End": true
+    },
```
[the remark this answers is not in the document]
MEPalma: Yes indeed from a state machine perspective we don't need `ProcessResult`. Here I want to make sure that the evaluation can continue to the "Next" state of the state machine with the expected memory state (the result of catch but not normalised). Basically checking we can leave the "error" state and continue as expected by sourcing "Next" with the right memory state.

tests/aws/services/stepfunctions/templates/errorhandling/statemachines/task_service_lambda_invoke_catch_all_output_path.json5
MEPalma: The test checks exactly for this. Previously the computation would fail as it tried to apply the `"OutputPath"` reduction on the error result (which had no "$.Payload" path). With these changes we can see how the evaluation instead continues correctly giving as output the expected output object `{"error"..., "cause"....}` and does not apply the reductions. We then have other tests in the test suite that instead check `*Path` operate as expected in other scenarios (e.g. with no errors)

localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
```diff
@@ -57,24 +59,31 @@ class CommonStateField(EvalComponent, ABC):
...
-        # A path that selects a portion of the state's output to be passed to the next state.
-        # If omitted, it has the value $ which designates the entire output.
-        self.output_path: OutputPath = OutputPath(OutputPath.DEFAULT_PATH)
-
-        self.state_entered_event_type: Final[HistoryEventType] = state_entered_event_type
-        self.state_exited_event_type: Final[Optional[HistoryEventType]] = state_exited_event_type
+        self.comment = None
+        self.input_path = InputPath(InputPath.DEFAULT_PATH)
+        self.output_path = OutputPath(OutputPath.DEFAULT_PATH)
+        self.output_path = OutputPath(OutputPath.DEFAULT_PATH)
```
derek-cbtw, line 84: Double assignment

localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
```diff
@@ -165,13 +174,20 @@ def _eval_body(self, env: Environment) -> None:
...
         output = env.stack[-1]
-        self._verify_size_quota(env=env, value=output)
 
-        # Filter the input onto the input.
-        if self.output_path:
-            self.output_path.eval(env)
+        # CatcherOutputs (i.e. outputs of Catch blocks) are never subjects of output normalisers,
+        # the entire value is instead passed by value as input ot the next state, or program output.
+        if isinstance(output, CatcherOutput):
+            env.inp = copy.deepcopy(output)
```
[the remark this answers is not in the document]
derek-cbtw: Thank you! I've been struggling for hours to understand what was going on here as I believe Step Functions Catch on custom errors is effectively broken due to this :)

localstack-core/localstack/services/stepfunctions/asl/component/state/state.py
gregfurman, line 84:
```suggestion
```

gregfurman:
```suggestion
```

tests/aws/services/stepfunctions/templates/errorhandling/statemachines/task_service_lambda_invoke_catch_all_output_path.json5
gregfurman: Ahh gotcha. That makes sense since. So we would expect the above to successfully process the reduction in an error case while ignoring the `{"error"..., "cause"....}` fields since that doesn't fall apart of teh payload we are wanting to access.

Issue #11560, opened 2024-09-22
bug: Second update of UsagePlan fails


artaban:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When performing **second** update of stack with `UsagePlan` it fails with exception with message **Id**.

### Expected Behavior

stack update could be performed many times in a row

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

`cdklocal deploy` performed twice in a row on stack with `UsagePlan` deployed

### Environment

```markdown
- OS: MacOS 15.0.0
- LocalStack:
  LocalStack version:3.7.3.dev37
  LocalStack build date: 2024-09-20
  LocalStack build git hash: 68e22de1
```

### Anything else?

The exception is thrown on `localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_usageplan.py:170`. The reason is that the `Id` is not present in `request.previous_state` after the first update.

PR #11561, opened 2024-09-22
bug: Second update of UsagePlan fails #11560

artaban:
## Motivation
Please take this PR as a guide for fixing the #11560. I'm not so deep into the localstack to find where the Id should have been added to the desired_state..

## Changes
- storing Id to the model of the UsagePlan update operation

What's left to do:
- implement it the right way :)
- write test


simonrw:
Hi @artaban , thank you for making this contribution. It just so happened I was doing some tidying up of our CFn resource providers and implemented the same fix you have in #11595. I think your approach fixes this problem as well, but I am trying to get my approach to become a pattern within resource providers.

Again, thank you but I will close this PR and merge mine instead, as this includes a fix for another resource provider with a similar issue. Please feel free to test the code changes once #11595 has been merged.

PR #11566, opened 2024-09-23
Update Dockerfile

sharkeshd:
Key Changes Made:
Removed Caching for apt-get: Cleaned up the apt-get cache to reduce image size.
Combined RUN Commands: Consolidated related RUN commands to minimize the number of layers.
Simplified Node.js Installation: Used the NodeSource setup script directly to avoid unnecessary dependencies.
Reduced Permissions Management: Simplified permission settings while still ensuring necessary access.
Avoided Repetitive Commands: Used COPY instead of ADD where applicable and minimized repetitive commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,201 +1,98 @@
-#
-# base: Stage which installs necessary runtime dependencies (OS packages, java,...)
-#
-FROM python:3.11.10-slim-bookworm@sha256:669bbd08353610485a94d5d0c976b4b6498c55280fe42c00f7581f85ee9f3121 AS base
+# Base stage: Install necessary runtime dependencies
+FROM python:3.11.10-slim-bookworm AS base
 ARG TARGETARCH
 
 # Install runtime OS package dependencies
-RUN --mount=type=cache,target=/var/cache/apt \
-    apt-get update && \
-        # Install dependencies to add additional repos
-        apt-get install -y --no-install-recommends \
-            # Runtime packages (groff-base is necessary for AWS CLI help)
-            ca-certificates curl gnupg git make openssl tar pixz zip unzip groff-base iputils-ping nss-passwords procps iproute2 xz-utils libatomic1 binutils && \
-        # patch for CVE-2024-45490, CVE-2024-45491, CVE-2024-45492
-        apt-get install --only-upgrade libexpat1
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates curl gnupg git make openssl tar pixz zip unzip groff-base iputils-ping \
+        && apt-get upgrade -y libexpat1 \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-# FIXME Node 18 actually shouldn't be necessary in Community, but we assume its presence in lots of tests
-# Install nodejs package from the dist release server. Note: we're installing from dist binaries, and not via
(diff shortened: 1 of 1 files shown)


thrau (commented):
hello @sharkeshd

could you please elaborate a bit on what these changes are aiming at exactly? the change set modifies a few things that we are doing on purpose (pinning the sha, using caches). some things look interesting (like simplifying the nodejs installer). but i'd prefer those to be separate contributions.

if you could explain what it is you're trying to achieve exactly, that would help us better direct you to make individual contributions for changes that make sense to localstack's dockerfile!

thanks

Dockerfile
```diff
@@ -1,201 +1,98 @@
...
+    curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && \
+    apt-get install -y nodejs && \
+    npm cache clean --force
+
+# Set workdir and create localstack user
+WORKDIR /opt/code/localstack
+RUN useradd -ms /bin/bash localstack && \
+    mkdir -p /var/lib/localstack /tmp/localstack && \
+    chmod 777 /var/lib/localstack /tmp/localstack && \
+    touch /tmp/localstack/.marker
```
alexrashed, line 28: This change here seems way cleaner, but it also does not perform some operations (chmod on `/root` and handling `/.npm`). They might be obsolete by now though?

Dockerfile
```diff
@@ -1,201 +1,98 @@
...
+# Upgrade Python build tools
+RUN virtualenv .venv && . .venv/bin/activate && pip install --upgrade pip wheel setuptools
 
-# add files necessary to install runtime dependencies
-ADD Makefile pyproject.toml requirements-runtime.txt ./
-# add the localstack start scripts (necessary for the installation of the runtime dependencies, i.e. `pip install -e .`)
-ADD bin/localstack bin/localstack.bat bin/localstack-supervisor bin/
+# Add necessary files and install runtime dependencies
+COPY Makefile pyproject.toml requirements-runtime.txt ./
+COPY bin/localstack bin/localstack.bat bin/localstack-supervisor bin/
```
alexrashed, line 54: +1, favoring add over copy [has been suggested by Docker due to its transparency / security / simplicity](https://www.docker.com/blog/docker-best-practices-understanding-the-differences-between-add-and-copy-instructions-in-dockerfiles/).

Dockerfile
```diff
@@ -1,201 +1,98 @@
...
-
-
-#
-# final stage: Builds upon base stage and copies resources from builder stages
-#
+# Final stage: Build LocalStack
 FROM base
 COPY --from=builder /opt/code/localstack/.venv /opt/code/localstack/.venv
-# The build version is set in the docker-helper.sh script to be the output of setuptools_scm
-ARG LOCALSTACK_BUILD_VERSION
```
alexrashed, line 136: ARG commands are only applicable in the context of a single image, so this cannot be removed here (because the arg is used with the `setuptools_scm` env var below).

alexrashed (changes requested):
Hi @sharkeshd!
Thanks a lot for your contribution! It actually contains some valid and useful enhancements to the `Dockerfile`. However, it is incredibly hard to ensure that this PR is safe and sound considering that the implications of each individual change itself are incredibly hard to predict. Maybe we can discuss the different enhancements here and then separate those deemed to be useful into individual PRs?

alexrashed:
I'll close this for now, please let us know if you want to pick it up again...

Issue #9337, opened 2023-10-12
bug: EC2 key pair fingerprint calculation


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

AWS uses different fingerprinting methods for key pairs based on their origin and types.

<https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html#how-ec2-key-fingerprints-are-calculated>

LocalStack currently uses MD5 for all cases.

### Expected Behavior

![image](https://github.com/localstack/localstack/assets/5170829/65d0d3c0-7bfd-4fcf-bbe2-8cc906e14262)

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

### Environment

```markdown
- OS: Ubuntu 22.04
- LocalStack: 2.3
```

iambary44:
> ### Is there an existing issue for this?
> 
> - [X] I have searched the existing issues
> 
> ### Current Behavior
> 
> AWS uses different fingerprinting methods for key pairs based on their origin and types.
> 
> 
> 
> <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html#how-ec2-key-fingerprints-are-calculated>
> 
> 
> 
> LocalStack currently uses MD5 for all cases.
> 
> ### Expected Behavior
> 
> ![image](https://github.com/localstack/localstack/assets/5170829/65d0d3c0-7bfd-4fcf-bbe2-8cc906e14262)
> 
> 
> 
> ### How are you starting LocalStack?
> 
> With a docker-compose file
> 
> ### Steps To Reproduce
> 
> #### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)
> 
> 
> 
>     docker run localstack/localstack
> 
> ### Environment
> 
> ```markdown
> - OS: Ubuntu 22.04
> 
> - LocalStack: 2.3
> ```
> 
> 
> ### Anything else?
> 
> _No response_

nifadyev:
Hey @viren-nadkarni . I made [PR](https://github.com/getmoto/moto/pull/8235) in `moto`, could you please take a look?

PR #11567, opened 2024-09-23
Fix EC2 key pair fingerprint calculation

Alansfit:
## Motivation
Fixes issue #9337: EC2 key pair fingerprint calculation.

LocalStack currently uses MD5 for all key types, but AWS uses different fingerprinting methods for key pairs based on their origin and types. This PR updates the fingerprint calculation logic to match AWS's methods.

## Changes
- Added `calculate_fingerprint` method to `EC2KeyPairProvider` class.
- Updated the `create` method to use the new `calculate_fingerprint` method for setting the `KeyFingerprint` property.

```python
def calculate_fingerprint(self, key_material: str, key_type: str) -> str:
    if key_type.lower() in ["ed25519", "ecdsa"]:
        return hashlib.sha256(key_material.encode()).hexdigest()
    else:
        return hashlib.md5(key_material.encode()).hexdigest()

<!-- Optional section: How to test these changes? -->
<!--
## Testing

-->

<!-- Optional section: What's left to do before it can be merged? -->
<!--
## TODO

What's left to do:

- [ ] ...
- [ ] ...
-->


Alansfit:

viren-nadkarni (changes requested):
Hi @Alansfit, thank you creating this PR!

Unfortunately this is not the right solution. The proposed changes will only alter the Cfn response. It will not update the fingerprint in the internal state. Retrieving the fingerprint through the EC2 REST API (CLI, boto, etc.) will continue to return the old, wrong value.

The fix should happen [here](https://github.com/getmoto/moto/blob/6c68326479024619214119c6d8835153e5f7804a/moto/ec2/models/key_pairs.py).

Alansfit:
Hi,

Thank you for pointing that out, I will rectify this.

PR #11568, opened 2024-09-23
minor: add ability to create tmp files inside specified dirs

eruditmorina:
## Motivation

I have a use case where I want to create temporary files inside a certain specified directory. This PR tries to extend the files utility functions so that temporary files and dirs can be created inside desired directories.

## Changes

- Added optional argument `dir` to the `new_tmp_file`, `new_tmp_dir` methods that defaults to `None` so that the default value can take place in that case.


eruditmorina:

thrau (approved):
LGTM

PR #11570, opened 2024-09-23
Add tags and operationId properties to all OpenAPI paths

giograno:
## Motivation
The end goal of our OpenAPI spec is to generate SDKs from them.
Having `operationIds` and `tags` for each path help generate better code. In particular:
- `operationId` is used for naming the generated functions;
- `tags` are used to organize the code into modules.

## Changes
- adding `operationId` and `tags` to all paths.

### DIFF
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -600,7 +600,7 @@ paths:
   /_localstack/diagnose:
     get:
       description: Get diagnostics report
-      operationId: get_diagnose
+      operationId: get_diagnostic
       tags: [internal]
       responses:
         '200':
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -254,7 +254,7 @@ paths:
   /_aws/dynamodb/expired:
     delete:
       description: Delete expired items from TTL-enabled DynamoDB tables
-      operationId: delete_expired_items
+      operationId: delete_ddb_expired_items
       tags: [aws]
       responses:
         '200':
@@ -555,7 +555,7 @@ paths:
           description: Current LocalStack configuration
     post:
       description: Configuration option to update with new value
-      operationId: update_configuration_option
+      operationId: update_config_option
       tags: [internal]
       requestBody:
         content:
@@ -702,7 +702,7 @@ paths:
   /_localstack/health:
     get:
       description: Get available LocalStack features and AWS services
-      operationId: get_ls_features_and_services
+      operationId: get_features_and_services
       tags: [internal]
       parameters:
       - allowEmptyValue: true
@@ -748,7 +748,7 @@ paths:
           description: ''
     post:
       description: Restart or terminate LocalStack session
-      operationId: restart_or_terminate_session
+      operationId: manage_session
       tags: [internal]
       requestBody:
         content:
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -273,7 +273,7 @@ paths:
   /_aws/events/rules/{rule_arn}/trigger:
     get:
       description: Trigger a scheduled EventBridge rule
-      operationId: trigger_event_bridge_Rule
+      operationId: trigger_event_bridge_rule
       tags: [aws]
       parameters:
       - description: EventBridge rule ARN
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/openapi.yaml
```diff
@@ -516,6 +544,8 @@ paths:
   /_localstack/config:
     get:
       description: Get current LocalStack configuration
+      operationId: get_localstack_config
+      tags: [internal]
```
viren-nadkarni, line 548: have you considered something like `localstack` instead of `internal`? The latter might suggest that it's not meant for public use.

furthermore, if the tag is set to `localstack`, you could set operation ID to `get_config`

localstack-core/localstack/openapi.yaml
```diff
@@ -524,6 +554,9 @@ paths:
                 type: object
           description: Current LocalStack configuration
     post:
+      description: Configuration option to update with new value
+      operationId: update_configuration_option
```
viren-nadkarni, line 558: for terseness

```suggestion
      operationId: update_config_option
```

localstack-core/localstack/openapi.yaml
```diff
@@ -568,6 +600,8 @@ paths:
   /_localstack/diagnose:
     get:
       description: Get diagnostics report
+      operationId: get_diagnose
```
viren-nadkarni, line 603: `get` and `diagnose` are both verbs.

maybe `get_diagnosis` or `get_diagnostics`?

localstack-core/localstack/openapi.yaml
```diff
@@ -668,6 +702,8 @@ paths:
   /_localstack/health:
     get:
       description: Get available LocalStack features and AWS services
+      operationId: get_ls_features_and_services
```
viren-nadkarni, line 705:
```suggestion
      operationId: get_features_and_services
```

`ls` is already implied via the tag

localstack-core/localstack/openapi.yaml
```diff
@@ -703,13 +739,17 @@ paths:
...
+      tags: [internal]
+      operationId: heath
       responses:
         '200':
           content:
             text/plain: {}
           description: ''
     post:
       description: Restart or terminate LocalStack session
+      operationId: restart_or_terminate_session
```
viren-nadkarni, line 751: this could be future-proof in case we support more operations

```suggestion
      operationId: manage_session
```

localstack-core/localstack/openapi.yaml
```diff
@@ -481,6 +507,8 @@ paths:
   /_aws/sqs/messages/{region}/{account_id}/{queue_name}:
     get:
       description: List SQS messages without side effects
+      operationId: list_sqs_queue_messages
```
viren-nadkarni, line 510: any idea if there is support for polymorphism and overloading? this operation and the previous are the same

if not, the operation ID could be set as follows to reduce confusion

`list_all_sqs_messages`
`list_sqs_messages`

localstack-core/localstack/openapi.yaml
```diff
@@ -252,6 +254,8 @@ paths:
   /_aws/dynamodb/expired:
     delete:
       description: Delete expired items from TTL-enabled DynamoDB tables
+      operationId: delete_expired_items
```
viren-nadkarni, line 257:
```suggestion
      operationId: delete_ddb_expired_items
```

or

```suggestion
      operationId: delete_dynamodb_expired_items
```

localstack-core/localstack/openapi.yaml
giograno: Good input, `localstack` fits much better (it is actually also what we used in our docs).

localstack-core/localstack/openapi.yaml
```diff
@@ -568,6 +600,8 @@ paths:
   /_localstack/diagnose:
     get:
       description: Get diagnostics report
+      operationId: get_diagnostic
```
viren-nadkarni, line 603: Nit

```suggestion
      operationId: get_diagnostics
```

localstack-core/localstack/openapi.yaml
```diff
@@ -703,13 +739,17 @@ paths:
                 type: object
           description: Available LocalStack features and AWS services
     head:
+      tags: [localstack]
+      operationId: heath
```
viren-nadkarni, line 743: Typo

```suggestion
      operationId: health
```

PR #11573, opened 2024-09-24
Fix EC2 key pair fingerprint calculation (#9337)

Alansfit:
## Motivation
This PR corrects the key pair fingerprint calculation for EC2.

## Changes
- Corrected the `key_pairs.py` to update the fingerprint calculation correctly.


.venv/Lib/site-packages/moto/ec2/models/key_pairs.py
alexrashed, line 1: Unfortunately, you can't just add something to the venv here. The venv contains installed dependencies. In this case it seems like you want to patch [moto](https://github.com/getmoto/moto). If you want to get this in, you have to file a PR over there. Once your changes are approved, merged, and released, they will automatically end up in LocalStack.

alexrashed (changes requested):
Hi @Alansfit!
Unfortunately, it seems like your PR has a lot of merge conflicts, and actually wants to patch something in moto. So this PR doesn't really belong here. I'll close this PR for now. Happy to answer any questions.

Issue #10773, opened 2024-05-06
bug: CreateFlowLogs with S3 log destination fails to discover bucket


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`EC2:CreateFlowLogs` when used with the S3 log destination type fails with following error:

```
$ awslocal ec2 create-flow-logs --resource-ids vpc-b996311b --resource-type VPC --log-destination-type s3 --log-destination arn:aws:s3:::flowlogs 
{
    "FlowLogIds": [],
    "Unsuccessful": [
        {
            "Error": {
                "Code": "400",
                "Message": "LogDestination: flowlogs does not exist."
            },
            "ResourceId": "vpc-b996311b"
        }
    ]
}
```

This is despite the S3 bucket existing:
```
$ awslocal s3 ls
2024-05-06 19:27:15 flowlogs
```

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

```
awslocal ec2 create-vpc --cidr-block 10.0.0.0/24 --query Vpc.VpcId --output text

awslocal s3 mb s3://flowlogs

awslocal ec2 create-flow-logs --resource-ids VPC --resource-type VPC --log-destination-type s3 --log-destination arn:aws:s3:::flowlogs 
```

### Environment

```markdown
- OS: jammy
- LocalStack: latest
```

### Anything else?

This occurs because in the Moto implementation of CreateFlowLogs, the S3 backend is directly accessed.

https://github.com/getmoto/moto/blob/dea4a98b64fe8b8a8d7b02e5446756d1b439af55/moto/ec2/models/flow_logs.py#L225-L231

On the other hand, the bucket lives in the S3 provider in LocalStack and the Moto backend is empty really.

arunsriniv:
I am also facing this issue. Is there any workaround for this issue or a fix is required?

PR #11574, opened 2024-09-24
fix EC2 FlowLogs with S3 destination

bentsku:
## Motivation
As reported in #10773, we had an issue since migrating S3 to LocalStack, as moto was directly accessing the bucket in memory to verify its existence.

It needs a fair amount of patching and overriding in the provider, but this seems like the safest bet. 

## Changes
- add 2 AWS validated tests, one positive and one negative
- implement `CreateFlowLogs` in the provider, and intercepted the request to remove the `s3` destination to avoid patching the whole method in `moto`
  - we now remove `s3` in order to pass the validation in `moto`, pre-validate ourselves, and put back the `s3` value after
  - we also validate the existence of the bucket, removing subfolders if there are any

_fixes #10773_

### DIFF
--- a/localstack-core/localstack/services/ec2/provider.py
+++ b/localstack-core/localstack/services/ec2/provider.py
@@ -525,8 +525,14 @@ class Ec2Provider(Ec2Api, ABC, ServiceLifecycleHook):
                     ],
                 )
 
-            response = call_moto_with_request(context, service_request)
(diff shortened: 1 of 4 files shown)


viren-nadkarni (approved):
LGTM, thanks for implementing this @bentsku !

PR #11577, opened 2024-09-25
fix(cors): append lambda logs header in the CORS headers list

webdev51:
## Motivation
During the implementation of the invoke function feature in the web app, lambda `invoke` call with `"LogType": "Tail"` option was setting a header `x-amz-log-type` causing a CORS error because this header is not in the CORS allowed headers list

## Changes
This PR adds the `x-amz-log-type` header to the list so the CORS error goes away


dfangl (approved):
Lgtm, thanks for adding it!

PR #11579, opened 2024-09-25
Feature/Events: Add internal log messages

maxhoheiser:
## Motivation
EventStudio open the black box eventbridge as an event router represents, therefore it is highly beneficial to surface internal steps that can be the cause for mismatches between expected behavior and actual behavior e.g. no rule matches the message. 
Therefore log messages are added at these specific instances.


Morijarti (approved):
Nice to see addition of structured logs.
Are we using similar logs in other parts of LS?

PR #11580, opened 2024-09-25
Add tests to test urlencoded IAM policies

dfangl:
## Motivation
During investigation of an IAM issue, I noticed our IAM implementation not returning exactly the same policies as submitted.

This happens, when parts of the policy are already urlencoded, and the response removes this encoding, returning the actual value.
This behavior however does not match AWS, which does return the policies as they were submitted.

A fix for this behavior is in getmoto/moto#8157, this PR only contains AWS validated tests for this behavior. The tests will fail, as long as the PR in moto is not merged and moto updated in LocalStack.

## Changes
* Add AWS validated tests to verify the policies being unchanged, even if they contain urlencoded parts

### DIFF
--- a/localstack-core/localstack/testing/snapshots/transformer_utility.py
+++ b/localstack-core/localstack/testing/snapshots/transformer_utility.py
@@ -345,6 +345,7 @@ class TransformerUtility:
             TransformerUtility.key_value("RoleName"),
             TransformerUtility.key_value("PolicyName"),
             TransformerUtility.key_value("PolicyId"),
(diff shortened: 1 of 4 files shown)


dominikschubert (approved):
LGTM! Tests should also now pass here after a rebase since we've upgraded our moto version since then :+1:

dfangl:
We need to wait for #11710 before rebasing and merging :+1:

pinzon (approved):
The PR you were waiting for is merged. Maybe is time to merge this one too :+1:

PR #11581, opened 2024-09-25
Add client factory to bypass LocalStack DNS server

simonrw:
## Motivation

We have the need for a boto client that bypasses the LocalStack DNS server. This client needs to be able to reach out to AWS (given valid user credentials).

## Changes

* Add new `ExternalBypassDnsClientFactory` client factory that ignores the LocalStack DNS server
* Subclass/specialise a few classes required to support a custom HTTP(s) connection factory
* The custom connection factory creates instances of a custom connection type, which temporarily replaces the host name connect to with the IP address of the same hostname, resolved _outside of LocalStack_.
* This replacement is done temporarily in the case of HTTPS because otherwise the certificate returned by AWS would not be valid for the IP address of the server

### DIFF
--- a/localstack-core/localstack/aws/connect.py
+++ b/localstack-core/localstack/aws/connect.py
@@ -668,12 +668,24 @@ class ExternalBypassDnsClientFactory(ExternalAwsClientFactory):
     Client factory that makes requests against AWS ensuring that DNS resolution is not affected by the LocalStack DNS
     server.
     """
+
+    def __init__(
+        self,
+        session: Session = None,
+        config: Config = None,
+    ):
+        super().__init__(use_ssl=True, verify=True, session=session, config=config)
+
     def _get_client_post_hook(self, client: BaseClient) -> BaseClient:
-        client._endpoint.http_session = CustomDNSSession()
+        client._endpoint.http_session = ExternalBypassDnsSession()
         return client
 
 
-class CustomHTTPConnectionCls(AWSHTTPConnection):
+class ExternalBypassDnsHTTPConnection(AWSHTTPConnection):
+    """
+    Connection class that bypasses the LocalStack DNS server for HTTP connections
+    """
+
     def _new_conn(self) -> socket:
         orig_host = self._dns_host
         try:
@@ -683,7 +695,10 @@ class CustomHTTPConnectionCls(AWSHTTPConnection):
             self._dns_host = orig_host
 
 
-class CustomHTTPSConnectionCls(AWSHTTPSConnection):
+class ExternalBypassDnsHTTPSConnection(AWSHTTPSConnection):
+    """
+    Connection class that bypasses the LocalStack DNS server for HTTPS connections
+    """
     def _new_conn(self) -> socket:
         orig_host = self._dns_host
         try:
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/aws/connect.py
```diff
@@ -5,18 +5,23 @@
...
 
+import functools
 import json
 import logging
 import re
 import threading
 from abc import ABC, abstractmethod
+from contextvars import ContextVar, copy_context
 from functools import lru_cache, partial
+from random import choice
```
pinzon, line 16: TIL `choice` :eyes:

localstack-core/localstack/aws/connect.py
```diff
@@ -631,8 +646,98 @@ def get_client(
...
+            "config": config,
+        }
+        return ServiceLevelClientFactory(
+            factory=self, client_creation_params=params, request_wrapper_clazz=ContextVarInjector
+        )
+
+
 connect_to = InternalClientFactory(use_ssl=localstack_config.DISTRIBUTED_MODE)
 connect_externally_to = ExternalClientFactory()
+connect_bypass_dns = ExternalBypassDnsClientFactory(use_ssl=True, verify=True)
```
pinzon, line 739: :partying_face:

alexrashed (commented):
The code itself is looking quite nice and clean, but it does perform some very low-level patches in `urllib3` and it needs a `ContextVar` which I'm both not a fan of to be honest.
I can see that there might not really be another option. I did a bit of research myself - and I'm sure you also did, so you might already see why none of them might work - maybe there is something we can try before moving forward with the current approach?

The things I came across when digging into `botocore` would be:
- There might be a way to dig into the endpoint resolving since this seems to be somewhat pluggable in botocore. This would allow us to modify the host to be looked up, which could then be "tainted" for the LocalStack DNS to use upstream instead. Unfortunately this would mean that the host header wouldn't be correct anymore...
- The best-practice in `requests` would be to use `HttpAdapters`, but it seems botocore moved away from `requests` 😢 
- There might be a way to replace the `urllib3` `HttpConnection` somehow:
  - `botocore` defines their own `Urllib3Session`: https://github.com/boto/botocore/blob/354c05fdd24c09fbdded587e2651a581de1c8487/botocore/httpsession.py#L275
  - This session registers custom `urllib3` connection pools: https://github.com/boto/botocore/blob/354c05fdd24c09fbdded587e2651a581de1c8487/botocore/httpsession.py#L302-L303
  - These pools define a custom `ConnectionCls` overwriting the default `urllib3` `HttpConnection`
  - The `HttpConnection.create_connection` call uses `self._dns_host` (the `host` property) to create the socket.
  - TL/DR: If we can replace the `HttpConnection` in a single client, and set `self._dns_host` with the already resolved IP we might have a chance to implement a bit of a cleaner hack (even though it's still a bit risky)?

simonrw:
> - There might be a way to replace the `urllib3` `HttpConnection` somehow:
>   - `botocore` defines their own `Urllib3Session`: https://github.com/boto/botocore/blob/354c05fdd24c09fbdded587e2651a581de1c8487/botocore/httpsession.py#L275
>   - This session registers custom `urllib3` connection pools: https://github.com/boto/botocore/blob/354c05fdd24c09fbdded587e2651a581de1c8487/botocore/httpsession.py#L302-L303
>   - These pools define a custom `ConnectionCls` overwriting the default `urllib3` `HttpConnection`
>   - The `HttpConnection.create_connection` call uses `self._dns_host` (the `host` property) to create the socket.
>   - TL/DR: If we can replace the `HttpConnection` in a single client, and set `self._dns_host` with the already resolved IP we might have a chance to implement a bit of a cleaner hack (even though it's still a bit risky)?

Thanks for the suggestions. I have managed to get a custom DNS resolver working for setting the `self._dns_host` via the upstream DNS resolver, however TLS verification fails because AWS does not present a certificate that includes the IP address in its SANs :( We could disable TLS verification, but I don't like that idea.

I think this implementation needs to operate on hostnames rather than IP addresses. Given that, I think our original solution is ok.

simonrw:
@alexrashed and I continued my spike to fully implement the DNS bypass in these changes

dfangl (approved):
Love the new approach, LGTM! Way better than patching and using context vars :)

localstack-core/localstack/aws/connect.py
```diff
@@ -631,9 +652,88 @@ def get_client(
...
+
+class ExternalBypassDnsHTTPConnectionPool(AWSHTTPConnectionPool):
+    ConnectionCls = ExternalBypassDnsHTTPConnection
+
+
+class ExternalBypassDnsHTTPSConnectionPool(AWSHTTPSConnectionPool):
+    ConnectionCls = ExternalBypassDnsHTTPSConnection
+
+
+class ExternalBypassDnsSession(URLLib3Session):
```
alexrashed, line 721: nit: Is there a chance to see the new session / client being tested here somehow? While this approach seems to be more stable, it still relies on some internals (like `_pool_classes_by_scheme`). Would be great to have a test which immediately informs us if this client won't work anymore (f.e. directly in the ASF update PRs where we bump the `botocore` version weekly).

alexrashed (approved):
Thanks a lot for addressing my comment, I feel like this solution - while still a bit intrusive - is less dangerous and more stable than the previous approach!
I think it would be great to have a test for this, but if this is already being waited for we could also wait for a later iteration, where the integration tests of usages of the client are implicitly testing this behavior.

localstack-core/localstack/aws/connect.py
cloutierMat: I have created a follow up ticket to make sure it doesn't fall off our radar. I should be good to tackle it this week. I will merge for now to help unblock further progress!

localstack-core/localstack/aws/connect.py
```diff
@@ -26,6 +37,7 @@
     INTERNAL_AWS_SECRET_ACCESS_KEY,
     MAX_POOL_CONNECTIONS,
 )
+from localstack.dns.server import get_fallback_dns_server
```
bblommers, line 40: This import is causing the CLI tests to fail, I think :eyes: 

https://github.com/localstack/localstack-ext/actions/runs/11251001003/job/31281083714?pr=3517

bblommers: NVM, I just saw @dfangl's to fix it: https://github.com/localstack/localstack/pull/11658 :raised_hands:

PR #11582, opened 2024-09-25
ESM-v2: Create hooks for downstream poller creation

gregfurman:
## Motivation
This PR allows for hooks to be defined downstream that can be used for defining an event source mapping. This hooks approach is currently due to the contraints of the factory pattern that `EsmWorkerFactory` and `EsmConfigFactory` currently employ.

## Changes
- Include `create_event_source_poller` and `set_event_source_config_defaults` pollers for creation of ESM workers and config defined downstream, respectively.
- The base `Poller` class, `EsmWorkerFactory`, and `EsmConfigFactory` now handle when a `source_arn` is not defined.

## Testing
- [ ] Running community test against downstream branch https://github.com/localstack/localstack/actions/runs/11071710532

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
@@ -1,3 +1,5 @@
+from typing import Callable, Optional, Type
+
 from localstack.aws.api.lambda_ import (
     EventSourceMappingConfiguration,
 )
@@ -12,6 +14,7 @@ from localstack.aws.api.pipes import (
     PipeTargetLambdaFunctionParameters,
     PipeTargetParameters,
 )
+from localstack.services.lambda_ import hooks as lambda_hooks
 from localstack.services.lambda_.event_source_mapping.esm_event_processor import (
     EsmEventProcessor,
 )
@@ -25,6 +28,7 @@ from localstack.services.lambda_.event_source_mapping.pipe_utils import (
 )
 from localstack.services.lambda_.event_source_mapping.pollers.dynamodb_poller import DynamoDBPoller
 from localstack.services.lambda_.event_source_mapping.pollers.kinesis_poller import KinesisPoller
+from localstack.services.lambda_.event_source_mapping.pollers.poller import Poller
 from localstack.services.lambda_.event_source_mapping.pollers.sqs_poller import SqsPoller
 from localstack.services.lambda_.event_source_mapping.senders.lambda_sender import LambdaSender
 from localstack.utils.aws.arns import parse_arn
@@ -58,7 +62,7 @@ class EsmWorkerFactory:
                 )
             ),
             target_client=lambda_client,
-            payload_dict=True,
+            payload_dict=True,  # TODO: This should be handled better since not all payloads in ESM are in the form { "Records" : List[Dict]}
         )
 
         # Logger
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
```diff
@@ -149,8 +157,23 @@ def get_esm_worker(self) -> EsmWorker:
                 processor=esm_processor,
             )
         else:
-            raise Exception(
-                f"Unsupported event source mapping source service {source_service}. Please upvote or create a feature request."
+            poller_holder = {"create_poller": None}
```
dfangl, line 160: It would be nice to add a type for that, so the hook signature can depend on it.

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
```diff
@@ -149,8 +157,23 @@ def get_esm_worker(self) -> EsmWorker:
...
+
+            create_poller: Optional[Callable[..., Type[Poller]]] = poller_holder.get(
+                "create_poller"
+            )
+            if not create_poller:
+                raise Exception(
+                    f"Unsupported event source mapping source service {source_service}. Please upvote or create a feature request."
+                )
+
+            poller: Type[Poller] = create_poller(
```
dfangl, line 173:
```suggestion
            poller: Poller = create_poller(
```

`create_poller` is most likely the type here, but the result should be `Poller`, right?

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
```diff
@@ -149,8 +157,23 @@ def get_esm_worker(self) -> EsmWorker:
...
             )
         else:
-            raise Exception(
-                f"Unsupported event source mapping source service {source_service}. Please upvote or create a feature request."
+            poller_holder = {"create_poller": None}
+            lambda_hooks.create_event_source_poller.run(
+                poller_holder, source_service, self.esm_config
+            )
+
+            create_poller: Optional[Callable[..., Type[Poller]]] = poller_holder.get(
```
dfangl, line 165: Here as well, is the return type really a callable returning the type of poller, or the instance of poller?

dfangl (approved):
LGTM in principal, would be nice to address the type comments I had first, then feel free to merge :)

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
gregfurman: `create_poller` is a function used to generate the `Poller`, where we expect the result to be a child instance of the base class `Poller`.

The way a child class of a parent `Parent` is "type-hinted" in Python is via the `Type[Parent]` syntax.

From the [Python docs](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#the-type-of-class-objects):
> Sometimes you want to talk about class objects that inherit from a given class. This can be spelled as type[C] (or, on Python 3.8 and lower, [typing.Type[C]](https://docs.python.org/3/library/typing.html#typing.Type)) where C is a class.

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
gregfurman: The return type is indeed a callable which we can use to construct the poller.

I opted for this approach since I did not want to pass the `source_arn`, `source_client`, and  `esm_processor` to the downstream hook.

Also, since we are using a factory pattern here, I prefer to keep the creation of the pollers in the driver function instead of downstream.

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
gregfurman: A type for the `create_poller` key-value or a type for the entire payload being passed to the hook?

i.e a `PollerFromHook()` class that gets some internal value mutated in the hook

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
dfangl: I think you misunderstood that documentation. We are not talking about the class itself, we are talking about instances of that class. The callable returned returns an instance of that class, not the class itself, so the `type[]` is not correct here.

For your type hints, the create_poller function would look like this:
```python
def _return_value(*args, **kwargs):
    return SubclassedPoller
```

But since the create_poller method (whatever it exactly is) creates an instance of the class, it is just `poller: Poller`.

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
gregfurman:
```suggestion
            create_poller: Optional[Callable[..., Poller]] = poller_holder.get(
```

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
gregfurman: Something like:

```python
class PollerHolder:
    def __init__(self):
        self.create_poller: Optional[Callable[..., Poller]] = None
```

That we pass in instead of the dict?

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
```diff
@@ -25,12 +28,19 @@
...
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.aws.client_types import ServicePrincipal
 
 
+class PollerHolder:
+    """Holds a `Callable` function `create_poller_fn` used to create a Poller. Useful when creating Pollers downstream via hooks."""
+
+    create_poller_fn: Callable[..., Poller] | None = None
+
+
```
gregfurman, line 43: @dfangl Is this what you had in mind?

PR #11585, opened 2024-09-25
rework conditional loading of the OAS plugins

giograno:
## Motivation
When [we introduced](https://github.com/localstack/localstack/pull/11025) the validation for the internal endpoints based on the OpenAPI specs, we made sure the correspondent plugins and specs wouldn't be loaded if at least one between `OPENAPI_VALIDATE_REQUEST` or `OPENAPI_VALIDATE_RESPONSE`.
The idea was to avoid this step, when unnecessary.

However, this is an issue if we turn these config variables when the container is already started, and the chain initialized.
In such a case, the calls to the validators fail because the instance variable `open_apis` hasn't been initialized (I came across this while writing some tests).

To support the later activation of the validation at runtime, we slightly reworked the loading logic.

## Changes
- We use the same old logic when instantiating the validator classes. However, when calling the handler we look for the `open_apis` attribute and eventually perform the loading.

### DIFF
--- a/localstack-core/localstack/aws/handlers/validation.py
+++ b/localstack-core/localstack/aws/handlers/validation.py
@@ -73,7 +73,10 @@ class OpenAPIValidator(Handler):
     open_apis: list["OpenAPI"]
 
     def __init__(self) -> None:
-        # avoid to load the specs if we don't have to perform any validation
+        self._load_specs()
+
+    def _load_specs(self) -> None:
+        """Load the openapi spec plugins iff at least one between request and response validation is set."""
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -92,6 +95,7 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
         if not config.OPENAPI_VALIDATE_REQUEST:
             return
 
+        hasattr(self, "open_apis") or self._load_specs()
```
viren-nadkarni, line 98: It might be cleaner to move the conditional into `_load_specs()` but yes, it will impact the performance due to the extra call

viren-nadkarni (approved):

PR #11586, opened 2024-09-26
Filter surrounding quotes from DNS_NAMES_TO_RESOLVE_UPSTREAM

dfangl:
## Motivation
Our documentation on skipping local DNS resolution https://docs.localstack.cloud/user-guide/tools/dns-server/#skip-localstack-dns-resolution uses surrounding quotes for the environment variable.
This is necessary for some shells to avoid expansion of the asterisks in the pattern, but when used in docker-compose it will lead to errors, as the value will be passed verbatim to the container - including the surrounding quotes.

Since the quotes are not part of the dns name, the pattern matching will fail, and the pattern will not be effective.

While we also want to fix the documentation, for ease of use it makes sense to just strip those surrounding quotes, as there should be no case where they are actually wanted.
I chose to strip them from every single name, to avoid cases like: `DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM='.*(ecr|lambda).*.amazonaws.com','second-domain.com'` not being properly stripped, even though that case seems unlikely.

## Changes
- LocalStack CLI will properly pass `DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM` to the container
- LocalStack will automatically strip surrounding quotes from the `DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM` config variable, to avoid user confusion


localstack-core/localstack/config.py
```diff
@@ -1177,6 +1177,7 @@ def use_custom_dns():
     "DNS_ADDRESS",
     "DNS_PORT",
     "DNS_LOCAL_NAME_PATTERNS",
+    "DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM",
```
simonrw, line 1180: Whoops

tests/bootstrap/test_dns_server.py
```diff
@@ -112,6 +112,47 @@ def test_skip_pattern(
...
+    local name resolution
+    """
+    ls_container = container_factory(
+        configurators=[
+            ContainerConfigurators.debug,
+            ContainerConfigurators.mount_docker_socket,
+            ContainerConfigurators.network(docker_network),
+            ContainerConfigurators.env_vars(
+                {
+                    "DNS_NAME_PATTERNS_TO_RESOLVE_UPSTREAM": r"'.*localhost.localstack.cloud'",
```
simonrw, line 133: To be complete, we should probably include a test with spaces and double quote characters to prevent regressions. And perhaps combinations? 

I don't know if this test should be parametrized over the "troublesome" character, or if we can combine them in one pattern for use here. I think a parametrized test makes more sense. It will slow down execution time a bit, but that is not a problem.

simonrw (approved):
Thanks for fixing this, we get this complaint surprisingly often.

tests/bootstrap/test_dns_server.py
dfangl: I parametrized the original test, less duplication, more covered cases :+1:

PR #11587, opened 2024-09-26
fix(cors): append lambda invocation headers in the expose headers list

webdev51:
## Motivation
After the implementation of https://github.com/localstack/localstack/pull/11577, CORS header was gone from lambda invocation implementation in the web app but a new problem came up where lambda invocation response was not returning full payload. The response should be returning 
```
export interface InvocationResponse {
    StatusCode?: Integer;
    FunctionError?: String;
    LogResult?: String;
    Payload?: _Blob;
    ExecutedVersion?: Version;
  }
```
but it was returning:
```
export interface InvocationResponse {
    StatusCode?: Integer;
    Payload?: _Blob;
  }
```
After some debugging, I found out the data was being returned in the response headers but not in the response. 

## Changes
This PR adds the required headers to the expose headers list so we get back the full payload.


dfangl (approved):
I think this is fine! We have to rethink our approach there in the future, though, as this list might grow significantly for other services.

PR #11589, opened 2024-09-26
fix APIGW test with hardcoded role name in CFN template

bentsku:
## Motivation
This PR fixes this issue: https://github.com/localstack/localstack/runs/30611307634
It seems we've had an APIGW test with a CFN template with a hardcoded role name set to `MyRole`, which most probably triggered a resource conflict issue

## Changes
- remove the role name attribute so that it is auto generated


steffyP (approved):
Thanks for jumping onto the fix 🙏 
LGTM!

PR #11591, opened 2024-09-26
CFn: improve error messages

simonrw:
## Motivation

When using our CloudFormation implementation, we often don't change errors when they are raised. This usually means the user is faced with horrible Python tracebacks that require our input to resolve.

A much better story is to parse and validate the template by converting the template into a meaningful graph of resources, with parameters, conditions and mappings resolved at template parse time, however we are not quite at that stage yet. The CFn implementation in LocalStack currently _works_, but is not pretty, and would take some effort to re-write at the moment, and at this stage it is not worth it.

The next best thing is to loose focus on parity with AWS and do what we can by improving error messages on obvious conditions. This PR tackles some of these things.

## Changes

* Add clarity when a mapping key cannot be found in a map, for both first and second level attributes
* Add clarity to parameters without default values when previous values are not set, and other similar situations

## TODO

More improvements...


simonrw:
The CI is green apart from one failing test with the arm tests: `tests/aws/services/cloudwatch/test_cloudwatch_metrics.py::TestSQSMetrics::test_alarm_number_of_messages_sent`. This is unrelated to this fix.

pinzon (approved):

dominikschubert (approved):
LGTM! Nice addition to improve clarity and allow for more insight into root causes. In terms of the tests, I think it might be good to separate these kind of tests a bit more (e.g. a separate class) but that's more of an optional nit and up for discussion at a later point.

PR #11595, opened 2024-09-27
CFn: fix unresolvable refs

### LINKED ISSUE #11560: bug: Second update of UsagePlan fails

simonrw:
## Motivation

With the previous template deployer implementation (supersceded in #10849) we iterated over the resources, and it was sometimes the case that we could not resolve a `Ref` (or `GetAtt` etc.) because the dependant resource had not been deployed yet.

However, with #10849 we deploy resources in the correct order, so we should not need to handle this case any more.

I added breakpoints in every position we raise a `DependencyNotYetSatisfied` exception, and uncovered the errors fixed in this PR.

## Changes

* Update `AWS::ApiGateway::UsagePlan` to correctly return the updated state from the `update` method
* Update `AWS::Lambda::CodeSigningConfig` to set the `CodeSigningConfigId` property

Note: this PR is not a complete fix, as we likely don't have test coverage of everything.

Closes #11560

### DIFF
--- a/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_usageplan.py
+++ b/localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_usageplan.py
@@ -210,6 +210,6 @@ class ApiGatewayUsagePlanProvider(ResourceProvider[ApiGatewayUsagePlanProperties
 
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
-            resource_model=model,
+            resource_model={**request.previous_state, **request.desired_state},
             custom_context=request.custom_context,
         )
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM for the APIGW `UpdatePlan` fix! Is it possible it is linked with https://github.com/localstack/localstack/pull/11560 and https://github.com/localstack/localstack/pull/11561? Or is it a different issue?

simonrw:
> LGTM for the APIGW `UpdatePlan` fix! Is it possible it is linked with #11560 and #11561? Or is it a different issue?

Thanks for pointing these out. This PR should resolve #11560 and supersede #11561.

simonrw:
Closes #11560

PR #11596, opened 2024-09-27
[SFN] Improve responsiveness on shutdown

MEPalma:
## Motivation
I recently noticed that LocalStack is unable to terminate responsively to user input when Step Functions workflows are evaluating callback tasks. Upon closer inspection, I found that some daemon threads were not properly configured. I also took the opportunity to introduce more conservative behaviour for certain infinite threads, explicitly syncing their lifetimes to that of the state machine. This last change could also improve the responsiveness of stop API actions.

## Changes
- Ensure proper configuration of daemon threads and TMP_THREADS where relevant
- Modify infinite loops to terminate when the associated state machine halts

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/program/program.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/program/program.py
@@ -72,7 +72,7 @@ class Program(EvalComponent):
     def eval(self, env: Environment) -> None:
         timeout = self.timeout_seconds.timeout_seconds if self.timeout_seconds else None
         env.next_state_name = self.start_at.start_at_name
-        worker_thread = threading.Thread(target=super().eval, args=(env,))
+        worker_thread = threading.Thread(target=super().eval, args=(env,), daemon=True)
         TMP_THREADS.append(worker_thread)
         worker_thread.start()
         worker_thread.join(timeout=timeout)
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/execute_state.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/execute_state.py
@@ -216,7 +216,7 @@ class ExecutionState(CommonStateField, abc.ABC):
         env.context_object_manager.context_object["State"]["RetryCount"] = 0
 
         # Attempt to evaluate the state's logic through until it's successful, caught, or retries have run out.
-        while True:
+        while env.is_running():
             try:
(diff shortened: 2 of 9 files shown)


localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/execute_state.py
```diff
@@ -213,7 +216,7 @@ def _eval_state(self, env: Environment) -> None:
         env.context_object_manager.context_object["State"]["RetryCount"] = 0
 
         # Attempt to evaluate the state's logic through until it's successful, caught, or retries have run out.
-        while True:
+        while env.is_running():
```
gregfurman, line 219: To clarify, is it deseriable to exit early when the program is in an error state?

```python
    def is_running(self) -> bool:
        return isinstance(self._program_state, ProgramRunning)
```

I'm wondering if there's a possibility of an early exit here since is there not a possibility of the program/env going into an error state (meaning `_program_state != ProgramRunning`) and this loop breaking prematurely since the condition is false?

gregfurman (approved):
LGTM! One nit (that can be done in a follow-up) and a clarification question on program state.

localstack-core/localstack/services/stepfunctions/asl/component/program/program.py
```diff
@@ -72,7 +72,7 @@ def _get_state(self, state_name: str) -> CommonStateField:
     def eval(self, env: Environment) -> None:
         timeout = self.timeout_seconds.timeout_seconds if self.timeout_seconds else None
         env.next_state_name = self.start_at.start_at_name
-        worker_thread = threading.Thread(target=super().eval, args=(env,))
+        worker_thread = threading.Thread(target=super().eval, args=(env,), daemon=True)
```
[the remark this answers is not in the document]
MEPalma: Yes good point, I first want to understand properly the internal guidelines on when to use or not this class. I would like future maintenance on FuncThread to be conceptually compatible with the idea of running parallel processes rather than background functions.

localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/execute_state.py
MEPalma: Yes this is the correct behaviour in this case.
The program state is changed only once after startup, and signals the program ending, terminal error, or stop request. All child threads should quite as soon as possible to allow the main thread (of the state machine execution) to return.

PR #11598, opened 2024-09-27
openapi: verify internal endpoints

giograno:
## Motivation

We recently introduced an OpenAPI spec for the developer endpoints of LocalStack. Additionally, our handler can also verify the requests and responses made to these endpoints.
Despite these tools at our disposal, we are not yet doing any sort of contract testing, therefore, we can't guarantee yet that our implementation matches the specs.

With this PR we start doing that for a set of LocalStack internal endpoints, for which we activate the OpenAPI validation.

## Changes
- Introducing a fixture to enable the validation for the request and the response of the developer endpoint.
- Minor fixes and improvements to the spec:
  - removed the required `feature` property for the health endpoint;
  - fixed the response of the `/_localstack/init/{stage}` endpoint;
  - created a component for the `/_aws/cloudwatch/metrics/raw` response.
- Enable the validation for the integration test suite testing the internal endpoints.

### Follow-up
I found out minor gaps between the specs and the implementation for which I'll follow up in separate PRs, with the help of the service owners.
- https://github.com/localstack/localstack/pull/11620
- https://github.com/localstack/localstack/pull/11621

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -336,7 +336,6 @@ jobs:
           GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DEBUG: 1
           DISABLE_BOTO_RETRIES: 1
-          OPENAPI_VALIDATE_RESPONSE: 1
           DNS_ADDRESS: 0
           LAMBDA_EXECUTOR: "local"
           LOCALSTACK_API_KEY: "test"
--- a/localstack-core/localstack/aws/handlers/validation.py
+++ b/localstack-core/localstack/aws/handlers/validation.py
@@ -138,7 +138,7 @@ class OpenAPIResponseValidator(OpenAPIValidator):
                     )
                     break
                 except ResponseValidationError as exc:
-                    LOG.error("Response validation failed for %s: $s", path, exc)
+                    LOG.error("Response validation failed for %s: %s", path, exc)
                     response.status_code = 500
                     response.set_json({"error": exc.__class__.__name__, "message": str(exc)})
                     chain.terminate()
--- a/localstack-core/localstack/aws/handlers/validation.py
+++ b/localstack-core/localstack/aws/handlers/validation.py
@@ -102,11 +102,14 @@ class OpenAPIRequestValidator(OpenAPIValidator):
             for openapi in self.open_apis:
                 try:
                     openapi.validate_request(WerkzeugOpenAPIRequest(context.request))
+                    # We stop the handler at the first succeeded validation, as the other spec might not even specify
+                    #   this path.
+                    break
                 except RequestValidationError as e:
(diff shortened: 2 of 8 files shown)


.github/workflows/tests-pro-integration.yml
```diff
@@ -336,7 +336,6 @@ jobs:
           GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DEBUG: 1
           DISABLE_BOTO_RETRIES: 1
```
alexrashed, line 338: q: This is an interesting change. You are disabling the global OpenAPI checks on every single request and instead just activate it selectively for certain tests. Could you elaborate a bit on the reasoning here? Why not just enabling it for all requests and responses for the whole execution of the test pipeline? Iirc the intention for the request validator was to even globally enable it by default for our users (in order to catch 4xx errors early without having to handle them in the actual endpoint implementations).

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -102,11 +102,14 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
...
                     openapi.validate_request(WerkzeugOpenAPIRequest(context.request))
+                    # We stop the handler at the first succeeded validation, as the other spec might not even specify
+                    #   this path.
+                    break
                 except RequestValidationError as e:
                     # Note: in this handler we only check validation errors, e.g., wrong body, missing required.
                     response.status_code = 400
                     response.set_json({"error": "Bad Request", "message": str(e)})
-                    chain.stop()
+                    chain.terminate()
```
alexrashed, line 112: q: You switched to terminate here in order to avoid the execution of the response handlers here. We only ever used this for the CORS handling to explicitly avoid any execution after the CORS handler. Why is this necessary here? This would circumvent f.e. the response logging.

alexrashed (commented):
Thanks @giograno for pushing the specs and building up measures for contract testing around our custom endpoints! 💯 
I posted two questions, afterwards we should be good to merge! 🚀

.github/workflows/tests-pro-integration.yml
giograno: Good question :) I enabled this with the PR introducing the validation and the pipeline was immediately green.
However, this was a false positive due to this typo: https://github.com/localstack/localstack/blob/e00a3623d5d7ddde2f5abbbe5db817bfe6235de8/localstack-core/localstack/aws/handlers/validation.py#L137.
Such a line was raising an exception and the handler was not stopping. As I fixed this, I saw that we have some gaps for some services, so I switched to a selective approach first. Globally enabling the verification is still the goal down the line.

localstack-core/localstack/aws/handlers/validation.py
giograno: No need actually. Thanks for pointing this out.

alexrashed (approved):
Thanks for addressing the questions! LGTM! 🚀

Issue #11600, opened 2024-09-28
feature request: IPv6 support for LocalStack gateway


laban-eilers-simplisafe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

LocalStack's gateway currently only supports listening on IPv4 addresses. It would be very useful to be able to use LocalStack in IPv6 environments, such as EKS in IPv6 mode, where pods don't get assigned IPv4 addresses.

Currently, the `GATEWAY_LISTEN` environment variable is used to configure the listening address and port. For example:

```
GATEWAY_LISTEN="0.0.0.0:4566"
```

I would like it to additionally support IPv6 format:

```
GATEWAY_LISTEN="[::]:4566"
```

### 🧑‍💻 Implementation

It looks like there's two places changes are needed:

* The `HostAndPort.parse()` method. It needs to do some additional string parsing to detect IPv6 addresses.
* The Twisted runtime server, which currently uses a `TCP4ServerEndpoint`, would need to be modified to use a `TCP6ServerEndpoint` for IPv6 hosts.

I haven't dug into what other aspects of localstack might need to be modified, but getting the main gateway server working would be a start.

simonrw:
@laban-eilers-simplisafe this functionality should exist based on your PR. If you still experience problems, feel free to open another issue. Thanks again for your contribution!

PR #11601, opened 2024-09-28
add IPv6 support to the runtime gateway

labaneilers:
## Motivation

Add support for configuring the gateway listener to listen on IPv6 addresses, so that LocalStack can be used in IPv6-only networks or environments (e.g. such as EKS in IPv6 mode).

For example, allow listening on all IPv6 interfaces:

```
GATEWAY_LISTEN="[::]"
```

Continue supporting the host + port format:

```
GATEWAY_LISTEN="[::]:4566"
```

Addresses https://github.com/localstack/localstack/issues/11600

**Note on the IPv6 address format**: `GATEWAY_LISTEN` expects IPv6 address to be enclosed in square brackets, since the existing format expected for `GATEWAY_LISTEN` is a subset of a URL (host and port), and square brackets are required for IPv6 addresses in URLs (to distinguish between the colons in an IPv6 address and the delimiter with the port number). This format is defined in https://www.rfc-editor.org/rfc/rfc6874.

## Changes

Updated:

* The HostAndPort.parse() method, which can now interpret an IPv6 address correctly. Added unit tests for this.
* The Twisted gateway, which will now use an IPv6 endpoint for an IPv6 host.

I also noticed a bug in the existing twisted implementation, in which it ignores any IPv4 host passed in via GATEWAY_LISTEN, and simply binds to all interfaces ("", which is interpreted as 0.0.0.0). I didn't fix this (given users might be depending on the current behavior), but the new IPv6 implementation respects the specified host.

## Testing

To test manually:

1. Build the LocalStack Docker image
2. Deploy it to an IPv6-only environment (such as a pod in an EKS cluster in IPv6-only mode)
3. Set the environment variable `GATEWAY_LISTEN="[::]"`
4. Use curl/wget to make a request to `http://[ipv6 address]]:4566` (e.g. `curl -v http://[2602:f95d:0:12:75cc::4]:4566`)
5. Validate that you got a 200 response, and the LocalStack logs show the request succeeded:

```
2024-09-28T17:50:29.073  INFO --- [et.reactor-0] localstack.request.http    : GET / => 200
```

## TODO

What's left to do:

- I didn't make any changes to the two other gateway implementations (hypercorn or werkzeug) as I'm not sure if they're still considered supported.
- There may be other services (e.g. emulated AWS interfaces) that may require additional changes to support IPv6, though I don't know enough about the LocalStack architecture to be sure.


localstack-core/localstack/runtime/server/twisted.py
```diff
@@ -33,8 +35,13 @@ def register(
...
         # add endpoint for each host/port combination
         for host_and_port in listen:
-            # TODO: interface = host?
-            endpoint = endpoints.TCP4ServerEndpoint(reactor, host_and_port.port)
+            if is_ipv6_address(host_and_port.host):
+                endpoint = endpoints.TCP6ServerEndpoint(
+                    reactor, host_and_port.port, interface=host_and_port.host
+                )
+            else:
+                # TODO: interface = host?
```
labaneilers, line 43: I preserved the existing behavior for IPv4 hosts here, even though it looks like a bug.

thrau:
thanks for the contribution @labaneilers . handing over this review to @simonrw

labaneilers:

labaneilers:
FYI: I don't think OSS contributors can add labels to a PR, but I would suggest this be labeled `semver:minor`.

labaneilers:
I think I've found a bug: UniqueHostAndPortsList will exclude an IPv6 address if GATEWAY_LISTEN includes two entries, and one of them is `0.0.0.0`, and they're on the same port. I'm working on a fix now.

labaneilers:
@localstack-bot recheck

labaneilers:
> @simonrw I think I've found a bug: UniqueHostAndPortsList will exclude an IPv6 address if GATEWAY_LISTEN includes two entries, and one of them is `0.0.0.0`, and they're on the same port. I'm working on a fix now, along with some new unit tests.

@simonrw I've fixed this, ready for review! Also took the opportunity to centralize the `is_ipv6_address()` function in the config module.

labaneilers:
and.... I've learned that in Twisted, you can't bind to both `0.0.0.0' and `::`. Binding to `::` appears to be dualstack (works for both IPv4 and IPv6). I'm going to rework my changes to `UniqueHostAndPortList` to make this work correctly.

localstack-core/localstack/config.py
```diff
@@ -584,40 +632,56 @@ class UniqueHostAndPortList(List[HostAndPort]):
...
+        - 0.0.0.0 "trumps" any other binding on IPv4 addresses only. IPv6 addresses
+          are not removed.
+        - Identical identical hosts and ports are de-duped
     """
 
-    def __init__(self, iterable=None):
-        super().__init__()
-        for item in iterable or []:
-            self.append(item)
+    def __init__(self, iterable: List[HostAndPort] | None = None):
```
labaneilers, line 644: I made a new algorithm for doing the dedupe, where the entire list is reconciled each time the list is modified, instead of having two separate sets of rules depending on the order items are added.

This also handles some additional corner cases, such as if the list contains multiple different IPs on the same port, and then a single `::` or `0.0.0.0` is added with the same port (which would previously result in a duplicate entry).

labaneilers:
I just figured out why the CLA assistant wasn't working- I had inadvertently signed the commit with my work identity instead of my personal identity, which I used to open the PR. I've just re-pushed with the correct identity.

localstack-core/localstack/config.py
```diff
@@ -584,40 +632,57 @@ class UniqueHostAndPortList(List[HostAndPort]):
...
+        """
+        if len(self) <= 1:
             return
 
-        # if 0.0.0.0:<port> already exists in the list, then do not add the new
-        # item
+        unique: List[HostAndPort] = list()
+
+        # Build a dictionary of hosts by port
+        hostsByPort: Dict[int, List[str]] = defaultdict(list)
```
simonrw, line 658: nit: we use snake case in LocalStack

```suggestion
        hosts_by_port: Dict[int, List[str]] = defaultdict(list)
```

localstack-core/localstack/config.py
```diff
@@ -584,40 +632,57 @@ class UniqueHostAndPortList(List[HostAndPort]):
...
-        if value.host == "0.0.0.0" and value.port in contained_ports:
-            for item in self:
-                if item.port == value.port:
-                    item.host = value.host
-            return
+            hostsByPort[item.port].append(item.host)
+
+        # For any given port, dedupe the hosts
+        for port, hosts in hostsByPort.items():
+            deduped_hosts = list(set(hosts))
```
simonrw, line 664: nit: `set`s can be iterated to the `list` is not required

tests/unit/test_config.py
```diff
@@ -203,6 +203,26 @@ def test_add_all_interfaces_value(self):
...
+        assert ports == [
+            HostAndPort("::", 42),
+        ]
+
+    def test_add_all_interfaces_value_mixed_ipv6_wins(self):
+        ports = config.UniqueHostAndPortList()
+        ports.append(HostAndPort("0.0.0.0", 42))
+        ports.append(HostAndPort("::", 42))
+        ports.append(HostAndPort("127.0.0.1", 42))
+        ports.append(HostAndPort("::1", 42))
```
simonrw, line 220: q: is it valid to listen on both `127.0.0.1` and `::1`?

simonrw (changes requested):
Thanks for making this change! Your test changes are comprehensive and this adds a good feature.

Unfortunately I don't have an environment to test this thoroughly, but since the feature is additive and you are finding benefit in using this then I'm happy to trust your tests.

Please can you let us know if you encounter any issues with this feature in the future!

Regarding your questions: the big unanswered one for me is whether services continue to work in an ipv6 only environment when we explicitly use ipv4. For example, lambda functions report back to the LocalStack container using the ipv4 address of the LS container.

One area that would be good to follow up with is the DNS server. For ipv4 we [use subnet matching to return a potentially different IP address for the LS container for the domain name `localhost.localstack.cloud`](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/dns/server.py#L190). See [my blog post](https://blog.localstack.cloud/2024-03-04-making-connecting-to-localstack-easier/#dynamic-name-resolution) for more information.

However we [don't do this for IPv6](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/dns/server.py#L192-L193). This would be a great follow up if you are interested?

---

Unfortunately this PR contains invalid syntax for Python 3.8 and 3.9, see the [failing test report](https://github.com/localstack/localstack/actions/runs/11149295358/job/31699601487?pr=11601#step:5:105). Could you fix this so we have a green test suite please? 

Once this has been done I will approve

tests/unit/test_config.py
labaneilers: It's nonsensical to bind to both, because `::1` includes `127.0.0.1`, and Twisted won't let you open listeners for both. This is why I made the logic remove `127.0.0.1` if `::1` has also been added.

labaneilers:
@simonrw I've updated everything with your feedback, and changed that union type to an old-school Python3.8-style `Union` type.

I'd be happy to look at that DNS change for my next contribution :)

labaneilers:
@simonrw oh, to answer this question:

> Regarding your questions: the big unanswered one for me is whether services continue to work in an ipv6 only environment when we explicitly use ipv4. For example, lambda functions report back to the LocalStack container using the ipv4 address of the LS container.

I'm not sure I can answer this well without knowing much more about the relationship of the Lambda is to the main LocalStack process. Is it a separate process running in the same container? How does the Lambda get the LocalStack process's IP?

Taking my best guess, that they're two processes in the same container: I think that in a true IPv6 only environment, the lambda wouldn't be able to connect to the main container if it's using an IPv4 address. However, in the environment I'm running in (EKS in IPv6 mode) containers still get an IPv4 address, but its just only routable from the same node (not part of the VPC). So I think the lambda callback would likely work in that case.

If you have any time, and are interested in helping me set up a test case, I can run it in an IPv6 EKS cluster and see if it works.

simonrw:
> I'm not sure I can answer this well without knowing much more about the relationship of the Lambda is to the main LocalStack process. Is it a separate process running in the same container? How does the Lambda get the LocalStack process's IP?

On non-k8s we typically run Lambda functions as [separate Docker containers](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/lambda_/invocation/docker_runtime_executor.py). 


> ... However, in the environment I'm running in (EKS in IPv6 mode) containers still get an IPv4 address, but its just only routable from the same node (not part of the VPC). So I think the lambda callback would likely work in that case.

So long as the inter-container networking is ipv4 then I think everything will work.

> If you have any time, and are interested in helping me set up a test case, I can run it in an IPv6 EKS cluster and see if it works.

Definitely. I can send you an example application which uses inter-container networking (more than just Lambda). I take it your workloads don't involve lambda?

I'm also interested in how you are using LocalStack. Particularly: how are you running LocalStack in your EKS environment? Are you using [our helm charts](https://github.com/localstack/helm-charts)? Have you found difficulties in running LocalStack on k8s? My team and I recently improved support for running LocalStack on k8s and I'd love to know any pain points or missing features.

labaneilers:
> I take it your workloads don't involve lambda?

Yeah- currently we aren't using LocalStack for Lambda. We use Lambda for other stuff in our infra, but don't have a need to emulate the API with LocalStack.

> I'm also interested in how you are using LocalStack. Particularly: how are you running LocalStack in your EKS environment? Are you using [our helm charts](https://github.com/localstack/helm-charts)? Have you found difficulties in running LocalStack on k8s? My team and I recently improved support for running LocalStack on k8s and I'd love to know any pain points or missing features.

We're not using the helm chart. As part of integration test setup, we spin up a single replica k8s deployment with a LocalStack container, along with a Service pointing at it. We build a custom LocalStack image, taking the base OSS image and adding some initialization scripts to it (e.g. adding Dynamo tables, SQS queues, etc). We then run other dependencies, the test subject app, or the test harness, in the same namespaces, using the service names to wire them all together.

The same tool that orchestrates all this in Kubernetes also can generate a docker-compose file and run the exact same stuff in Docker Desktop locally.

As far as missing features: we're generally pretty happy with the feature set, though I don't think we're using a lot of variety- mostly Dynamo, SQS, SNS, S3, and a few other APIs. IPv6 is the first feature we've really needed that was missing.

tests/unit/test_config.py
simonrw: I don't think this is covered by the tests. I can see tests that ensure that if `::` is added then this "trumps" everything, but I don't see a test covering adding `::` and `127.0.0.1`.

I have written this test and it fails:


```python
    def test_loopback_ipv6_only(self):
        ports = config.UniqueHostAndPortList()
        ports.append(HostAndPort("::1", 4566))
        ports.append(HostAndPort("127.0.0.1", 4566))

        assert ports == [
            HostAndPort("::", 4566),
        ]
```
Would it be possible to fix this, then I am happy with this PR

simonrw (approved):
Hi @labaneilers, thanks for making this contribution. From what I can see from my testing, LocalStack continues to work with IPv4 networking internally. Since this code change is additive and does not impact IPv4 networks, I am happy to merge this.

Thanks for this contribution, it's great to see wider support for IPv6 in general!

PR #11604, opened 2024-09-30
remove assigning a ServiceModel for `execute-api` APIGW routes

bentsku:
## Motivation
Due to legacy code when we didn't have the ASF framework and Edge routes, we would assign the service model for any "data plane" kind of routes (`execute-api`) for APIGW for example. 

Nowadays, if we have a service set in the context, and we did not match an edge route, we will forward the invocation to the skeleton/provider. This can lead to weird issues and confuse users with this kind of error:

```python
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 63, in __call__
    return self.parse_and_enrich(context)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 67, in parse_and_enrich
    operation, instance = parser.parse(context.request)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/protocol/parser.py", line 174, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/protocol/parser.py", line 567, in parse
    raise OperationNotFoundParserError(
localstack.aws.protocol.parser.OperationNotFoundParserError: Unable to find operation for request to service apigateway: POST /hello
2024-09-27T20:51:19.434  INFO --- [et.reactor-8] localstack.request.http    : POST /hello => 500; Request(b'', headers={'Host': 'https://apiid.execute-api.localhost.localstack.cloud:4566', 'User-Agent': 'curl/8.9.1', 'Accept': '*/*', 'Authorization': 'AWS4-HMAC-SHA256 Credential=injectedaccesskey/20160623/us-east-1/apigateway/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=1234', 'x-moto-account-id': '000000000000'}); Response(Bytes(2.187KB), headers={'Content-Type': 'application/json', 'X-Amzn-Errortype': 'InternalError', 'Content-Length': '2187', 'x-amzn-requestid': '3869759f-8b6b-4d2e-9acc-e07063de53b8', 'x-amz-request-id': '3869759f-8b6b-4d2e-9acc-e07063de53b8'})
```

This is not what we want: if the request goes to an `execute-api` route, it should never be forward to the provider, and instead just be caught by the router. We still have the "provider" loading issue, but this is unrelated to this particular problem. 

Also, it would then add a fake `Authorization` header populated with `apigateway` service, leading to even more confusion 😄 as there was a service model set when entering the `localstack.aws.handlers.auth.MissingAuthHeaderInjector` handler. 

Note: I believe we should maybe flag those request with the `execute-api` service identifier, as this is how AWS flags those with signing (for [IAM authorization](https://docs.aws.amazon.com/apigateway/latest/developerguide/permissions.html)):
> `execute-api` designates the underlying API execution component of API Gateway.

You need to sign your request with the `execute-api` service. I've tried replacing the rule above, but I believe it's going to try to load the `execute-api` service which doesn't exist, so it could maybe also be a specific IAM rule somewhere. Anyway, I thought it was worth mentioning!
See for the signature with `execute-api`:
- https://martinjt.me/2022/01/20/authenticating-with-aws-api-gateway-with-iam-credentials-using-javascript-and-node/
- https://medium.com/@jun711.g/how-to-secure-aws-api-gateway-requests-with-signature-version-4-using-aws-amplify-62d79f92966c

\cc @cloutierMat 

## Changes
- remove the legacy rule to assign the `apigateway` service to `execute-api` routes

### DIFF
--- a/localstack-core/localstack/aws/protocol/service_router.py
+++ b/localstack-core/localstack/aws/protocol/service_router.py
@@ -14,7 +14,7 @@ from localstack.aws.spec import (
     build_service_index_cache,
     load_service_index_cache,
 )
-from localstack.constants import LOCALHOST_HOSTNAME, PATH_USER_REQUEST, VERSION
+from localstack.constants import VERSION
 from localstack.http import Request
 from localstack.services.s3.utils import uses_host_addressing
 from localstack.services.sqs.utils import is_sqs_queue_url
@@ -178,13 +178,6 @@ def legacy_rules(request: Request) -> Optional[ServiceModelIdentifier]:
     path = request.path
     method = request.method
     host = hostname_from_url(request.host)
-
-    # API Gateway invocation URLs
-    # TODO: deprecated with #6040, where API GW user routes are served through the gateway directly
-    if ("/%s/" % PATH_USER_REQUEST) in request.path or (
-        host.endswith(LOCALHOST_HOSTNAME) and "execute-api" in host
-    ):
-        return ServiceModelIdentifier("apigateway")
 
     if ".lambda-url." in host:
         return ServiceModelIdentifier("lambda")
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -59,8 +59,7 @@ class ApiGatewayEndpoint:
             self.rest_gateway.process_with_context(context, response)
             return response
         else:
-            # TODO: return right response
-            return Response("Not authorized", status=403)
+            return self.create_not_found_response(api_id)
 
     def prepare_rest_api_invocation(
         self, request: Request, api_id: str, stage: str
@@ -100,6 +99,14 @@ class ApiGatewayEndpoint:
             # There appears to be in issue in Localstack, where setting "close" will result in "close, close"
             response.headers.set("Connection", "keep-alive")
         return response
+
+    @staticmethod
+    def create_not_found_response(api_id: str) -> Response:
+        not_found = Response(status=404)
+        not_found.set_json(
+            {"message": f"The API id '{api_id}' does not correspond to a deployed API Gateway API"}
+        )
+        return not_found
 
 
 class ApiGatewayRouter:
--- a/tests/aws/services/apigateway/test_apigateway_common.py
+++ b/tests/aws/services/apigateway/test_apigateway_common.py
@@ -1290,3 +1290,24 @@ class TestApigatewayRouting:
             "errorType": response.headers.get("x-amzn-ErrorType"),
         }
         snapshot.match("wrong-method", resp)
+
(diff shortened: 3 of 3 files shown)


localstack-core/localstack/aws/protocol/service_router.py
```diff
@@ -179,13 +179,6 @@ def legacy_rules(request: Request) -> Optional[ServiceModelIdentifier]:
     method = request.method
     host = hostname_from_url(request.host)
 
-    # API Gateway invocation URLs
-    # TODO: deprecated with #6040, where API GW user routes are served through the gateway directly
-    if ("/%s/" % PATH_USER_REQUEST) in request.path or (
-        host.endswith(LOCALHOST_HOSTNAME) and "execute-api" in host
-    ):
-        return ServiceModelIdentifier("apigateway")
-
```
alexrashed, line 188: Would it make sense to have a test which ensures that the handling of non-existing API GW routes works correctly?

alexrashed (approved):
Really good catch! This totally makes sense. I am a bit surprised that this issue wasn't raised earlier. Are there any known user reports from this? Why wasn't this being raised by anyone in the past? 🤔
But again, the change is clean and make sense. It would be great to have a test, but this isn't blocking a merge.
Thanks for jumping on this! 💯

localstack-core/localstack/aws/protocol/service_router.py
alexrashed: Yeah, that totally makes sense. And even this isn't properly being testable against AWS due to fundamental restrictions, I think a test would be super-valuable to (a) avoid any kind of regressions and most importantly (b) exactly define how the behavior should and does look like.

bentsku: Definitely agree! Will add them a bit later today

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -59,8 +59,7 @@ def __call__(self, request: Request, **kwargs: Unpack[RouteHostPathParameters])
             self.rest_gateway.process_with_context(context, response)
             return response
         else:
-            # TODO: return right response
-            return Response("Not authorized", status=403)
+            return self.create_not_found_response(api_id)
```
cloutierMat, line 62: Great addition!

cloutierMat (approved):
Good catch! I love the simple message that will help our user understand the root of the error, beyond a big unrelated python exception trace! :rocket:

PR #11593, opened 2024-09-27
Prototyping api/feature coverage

sannya-singal:
## Motivation
Extending https://github.com/localstack/localstack/pull/11592

PR #11592, opened 2024-09-27
Playground feature catalog

steffyP:

PR #11606, opened 2024-09-30
Feature catalog generate output, add generic ApiLimitation

steffyP:
## Motivation
This PR is building on https://github.com/localstack/localstack/pull/11593, which in turn already builds on https://github.com/localstack/localstack/pull/11592.
The PR is targeting the playground_feature_catalog branch (#11592).

## Changes
* includes the KMS feature catalog + script from https://github.com/localstack/localstack/pull/11593
* refactoring: move all feature-catalog related decorators into one package
* add another decorator `ApiLimitation` than could be used in a generic way for specific api operations
  * @sannya-singal also proposed decorators in #11593, but it requires generating new decorators for every api operation
  * with the generic approach, we could include limitations with one decorator
 * PoC for generating json-output using the feature-catalog classes, and collect api-operations that use those as decorator


localstack-core/localstack/services/kms/provider.py
```diff
@@ -368,6 +370,9 @@ def _is_rsa_spec(key_spec: str) -> bool:
     #
 
     @handler("CreateKey", expand=False)
+    @kms_feature.Provisioning
+    @kms_api.CreateKey
+    @ApiLimitation(limitation="Status 'Updating' is not supported.")
```
steffyP, line 375: testing both approaches here:
* the second one `ApiLimitation` would be more generic, eg can be reused for every operation, and we can provide the actual limitations as input (currently string, but we could change this to a list as well).
  * if there are a lot of limitations we could use constants that are defined in a separate file, to not pollute the code too much
  * tested this approach in the `playground.py` script, e.g. it will be added as `api_limitation`:
  ```
          "Provisioning": {
          "docs": {
            "general_docs": "Manage the creation and modification of the keys.",
            "implementation_status": "ImplementationStatus.FULLY_IMPLEMENTED",
            "support_type": "SupportStatus.SUPPORTED",
            "aws_docs_link": "https://docs.aws.amazon.com/kms/latest/developerguide/create-keys.html",
            "api_operations": [
              "CreateKey"
            ]
          },
          "operations": [
            "localstack.services.kms.provider.KmsProvider.create_key",
            "localstack.services.kms.provider.KmsProvider.update_key_description"
          ],
          "api_limitations": {
            "localstack.services.kms.provider.KmsProvider.create_key": "Status 'Updating' is not supported."
          }
        },
   ```

localstack-core/localstack/services/kms/provider.py
```diff
@@ -737,7 +756,7 @@ def _generate_data_key_pair(
             "PublicKey": crypto_key.public_key,
         }
 
-    @handler("GenerateDataKeyPair")
```
steffyP, line 740: i had to remove the empty handlers, as these somehow interfered with the tests in `asf_utils.py` - didn't investigate further, I am sure we could solve this and keeping the handler. On the other hand the handler doesn't do anything here :)

scripts/feature_catalog_playground/playground_kms_api_limitation.py
steffyP, line 1: moved the script and adapted the paths. Originally added by @sannya-singal in https://github.com/localstack/localstack/pull/11593

sannya-singal (approved):

sannya-singal:
Closing https://github.com/localstack/localstack/pull/11593 as the current branch pulls the related changes.

PR #11607, opened 2024-09-30
fix: mixed-content for the SwaggerUI

giograno:
## Motivation
With https://github.com/localstack/localstack/pull/11509 we started serving the SwaggerUI within LocalStack.
The Web UI is initialized with an OpenAPI spec, also served via LocalStack. However, the resolved URL always had a `http` protocol. This caused a mix-content issue (see below) when accessing the swagger URL (`localhost.localstack.cloud:4566/_localstack/swagger`) with `https`.

See the image below:

image

## Changes
- We now build the spec URL with the correct protocol based on the protocol of the request itself.

### DIFF
--- a/localstack-core/localstack/http/resources/swagger/endpoints.py
+++ b/localstack-core/localstack/http/resources/swagger/endpoints.py
@@ -1,22 +1,18 @@
 import os
 
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Thanks for jumping on this! Looks good! 🚀

PR #11609, opened 2024-09-30
fix APIGW NG test_cors_apigw_not_applied

bentsku:
## Motivation
Follow up from #11604, I've missed this one as we did not run the CI with the flag applied: 

Here's the failed run: https://app.circleci.com/pipelines/github/localstack/localstack/28439/workflows/ee18ae42-968c-4eda-96bf-fbae9ad40952/jobs/248146/tests

## Changes
- remove the special check for NextGen and always assert it is `404`

### DIFF
--- a/tests/integration/test_security.py
+++ b/tests/integration/test_security.py
@@ -178,10 +178,7 @@ class TestCSRF:
                 "Origin": "https://app.localstack.cloud",
             },
         )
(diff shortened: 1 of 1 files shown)


cloutierMat (approved):
Good catch! Thanks for fixing!

PR #11588, opened 2024-09-26
CFn: propagate lambda tags to function

simonrw:
## Motivation

In #11459 we find that tags are not propagated to lambda functions when deployed via CloudFormation.

This superscedes #11529.

Closes #11459 

## Changes

* Propagate the tags from the CFn resource payload to the `create_function` call.
* Add integration test capturing this

Note: the test is not a snapshot test because CFn adds default tags which we have to ignore, and the tags contain `:` characters, which the jsonpath-based snapshot library does not cope with.

PR #11614, opened 2024-10-01
CFn: improve lambda tag matching

simonrw:
## Motivation

PR #11588 implemented tag propagation for Lambda Functions using CFn. The test added was validated, but not snapshot tested because I did not know how to handle the automatic CloudFormation tags when matching snapshots.

A [very useful comment](https://github.com/localstack/localstack/pull/11588#issuecomment-2385454876) pointed out how to achieve this, so this PR upgrades the test to use snapshot validation of the created resource.

Thanks @bentsku!

## Changes

* Capture the `get_function` response in a snapshot, and correctly skip matching the CFn tags

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_lambda.py
+++ b/tests/aws/services/cloudformation/resources/test_lambda.py
@@ -487,9 +487,22 @@ def test_multiple_lambda_permissions_for_singlefn(deploy_cfn_template, snapshot,
 
 
 @markers.aws.validated
-def test_lambda_function_tags(deploy_cfn_template, aws_client):
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        # Added by CloudFormation
+        "$..Tags.'aws:cloudformation:logical-id'",
+        "$..Tags.'aws:cloudformation:stack-id'",
+        "$..Tags.'aws:cloudformation:stack-name'",
+    ]
+)
+def test_lambda_function_tags(deploy_cfn_template, aws_client, snapshot):
+    snapshot.add_transformer(snapshot.transform.lambda_api())
+    snapshot.add_transformer(snapshot.transform.key_value("CodeSha256"))
+
     function_name = f"fn-{short_uid()}"
     environment = f"dev-{short_uid()}"
+    snapshot.add_transformer(snapshot.transform.regex(environment, "<environment>"))
+
     deploy_cfn_template(
         template_path=os.path.join(
             os.path.dirname(__file__),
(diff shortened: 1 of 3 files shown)


tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
```diff
@@ -1529,5 +1529,65 @@
...
+          "TracingConfig": {
+            "Mode": "PassThrough"
+          },
+          "Version": "$LATEST"
+        },
+        "Tags": {
+          "Environment": "<environment>",
+          "aws:cloudformation:logical-id": "TestFunction",
+          "aws:cloudformation:stack-id": "arn:<partition>:cloudformation:<region>:111111111111:stack/stack-df418b45/<resource:3>",
+          "aws:cloudformation:stack-name": "stack-df418b45",
```
bentsku, line 1583: this needs to be transformed, as it wouldn't be runnable against AWS. 
adding `snapshot.transform.key_value("aws:cloudformation:stack-name")` would do the trick as it would replace the value both here and in the `stack-id` part

tests/aws/services/cloudformation/resources/test_lambda.py
```diff
@@ -487,9 +487,22 @@ def test_multiple_lambda_permissions_for_singlefn(deploy_cfn_template, snapshot,
...
+        "$..Tags.'aws:cloudformation:stack-name'",
+    ]
+)
+def test_lambda_function_tags(deploy_cfn_template, aws_client, snapshot):
+    snapshot.add_transformer(snapshot.transform.lambda_api())
+    snapshot.add_transformer(snapshot.transform.key_value("CodeSha256"))
+
     function_name = f"fn-{short_uid()}"
     environment = f"dev-{short_uid()}"
+    snapshot.add_transformer(snapshot.transform.regex(environment, "<environment>"))
```
bentsku, line 504: q: is this needed? Why can't this be a regular `key_value` transformer with the `Environment` key?

bentsku (commented):
Nice! Thanks for adding the snapshot! A view comments, and it should be good to merge 😄

tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
simonrw: Thanks for catching this, I missed this in my initial scan of the snapshot 👍

tests/aws/services/cloudformation/resources/test_lambda.py
simonrw: When the value is specific here (like a randomly generated ID that's not a standard part of the API) I like my transformations to be specific. Just a personal preference.

bentsku: I see. I'm not too fond of the `regex` transformer myself, as it acts on the raw JSON string, but I can see the point.

bentsku (approved):

PR #11616, opened 2024-10-01
Add configuration value for manually selecting runtime components

silv-io:
## Motivation

When loading multiple components in the python path and launching localstack, we currently just select the first component that was found. 
When working in a repository which has multiple exclusive components, however, this means that we need to manually remove one of the components from the python path to specify which component we want to run exactly.

Ideally, in a development environment we'd just want to set some configuration value in a run configuration to select the correct component.

/cc @dfangl 

## Changes

This PR adds a configuration value to manually select a single runtime components set to start: `RUNTIME_COMPONENTS`

Without setting it, the old behavior remains, i.e., all components are loaded and the first is chosen.

When specifying a value, however, that one gets loaded and chosen.

## Testing

In a repository with multiple defined components sets, put both in the python path, select first one and then the other. See if that causes different sets of components to be loaded.

### DIFF
--- a/localstack-core/localstack/runtime/runtime.py
+++ b/localstack-core/localstack/runtime/runtime.py
@@ -182,12 +182,10 @@ def create_from_environment() -> LocalstackRuntime:
         try:
             component = plugin_manager.load(config.RUNTIME_COMPONENTS)
             return LocalstackRuntime(component)
-        except ValueError as e:
-            LOG.error(
-                "Could not load runtime component %s: %s. Falling back to loading all components.",
-                config.RUNTIME_COMPONENTS,
-                e,
-            )
+        except Exception as e:
+            raise ValueError(
+                f"Could not load runtime components from config RUNTIME_COMPONENTS={config.RUNTIME_COMPONENTS}: {e}."
+            ) from e
     components = plugin_manager.load_all()
 
     if not components:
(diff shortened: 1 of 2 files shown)


thrau (commented):
nice! thanks for tackling this one

a few suggestions:
* let's call it `RUNTIME_COMPONENTS`
* if `RUNTIME_COMPONENTS` is set, i'd suggest simply trying to `plugin_manager.load(config.RUNTIME_COMPONENTS)` and fail if the plugin doesn't load, then proceed with the logic we have now where we load all components, etc)
* remove the TODO in the `create_from_environment` docstring

localstack-core/localstack/runtime/runtime.py
```diff
@@ -173,13 +173,21 @@ def create_from_environment() -> LocalstackRuntime:
...
+    if config.RUNTIME_COMPONENTS:
+        try:
+            component = plugin_manager.load(config.RUNTIME_COMPONENTS)
+            return LocalstackRuntime(component)
+        except ValueError as e:
+            LOG.error(
+                "Could not load runtime component %s: %s. Falling back to loading all components.",
+                config.RUNTIME_COMPONENTS,
+                e,
+            )
```
thrau, line 190: sorry i perhaps wasn't clear in the initial suggestions: i would actually raise an error here, something like:
i wouldn't silently fall back to something else if the user explicitly specifies something and expects it to load.

```suggestion
        except Exception as e:
            raise ValueError(
                "Could not load runtime components from config RUNTIME_COMPONENTS={config.RUNTIME_COMPONENTS}: {e}."
            ) from e
```

thrau (approved):
LGTM! i would just raise an exception instead of logging the error and falling back to a default

PR #11619, opened 2024-10-01
ESM-v2: Add optional close function to Poller

gregfurman:
## Motivation
Allows for a `Poller` to optionally close some internal components after the `poll_events` loop has exited.

## Changes
- Add a no-op `close()` method to the `Poller` base class.


joe4dev (approved):
LGTM as discussed 👍

PR #11620, opened 2024-10-01
Validate SQS dev endpoints

giograno:
## Motivation
As we introduced specs for the developer endpoints, we want to make sure they match the actual implementation.
For this purpose, we can activate request and response verification in our handler chain and test our endpoint against the OpenAPI spec.

This PR activates the validation for the SQS dev endpoints.

## Changes
- Enable the validation for the `TestSqsDeveloperEndpoints` suite;
- Minor changes to the spec to let the suite pass;

🔈  I'd ask the service owners to check if this PR is already comprehensive and eventually, if not, please help me fill all possible gaps.

### DIFF
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -8,7 +8,7 @@ info:
     checks, plugins, initialisation hooks, service introspection, and more.
   termsOfService: https://www.localstack.cloud/legal/tos
   title: LocalStack REST API for Community
-  version: 3.6.1.dev
+  version: latest
 externalDocs:
   description: LocalStack Documentation
   url: https://docs.localstack.cloud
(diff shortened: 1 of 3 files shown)


thrau (commented):
i have a question about this: since this endpoint is compatible with the query protocol that SQS uses, the schema doesn't really cover that. a strict validation would therefore probably break this endpoint, right? or is the schema only validated if the content type is application/json?

giograno:
Thanks for the comment @thrau! I am not a SQS expert so please confirm the following.

I noticed in our tests that we can use the dev endpoint as boto client URL (only for `ReceiveMessage`). That results in a `POST` call to `_aws/sqs/message`. We missed this part in our spec (we only had the `GET` operations). Our validation, however, was not failing since we let `OperationNotFound` errors slip through.

I added the `POST` operation for `_aws/sqs/message` with both the `x-www-form-urlencoded` and `json` options.

thrau (approved):
cool! if all tests pass then :shipit:

PR #11621, opened 2024-10-01
Validate SNS developer endpoints

giograno:
## Motivation
After introducing OpenAPI specs for our internal endpoints, we are now introducing stricter validations in our test pipeline.
This PR enables the validation of the requests and responses for the developer endpoints in the SNS test suite.
It also adds minor changes to the specs to reflect the actual implementation and make the suite green.

## Changes
- Enable Open API validation for the SNS tests;
- Minor adaptations to the spec.

### DIFF
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -232,6 +232,89 @@ components:
       - error
       - subscription_arn
       type: object
+    SNSPlatformEndpointMessage:
+      type: object
+      description: Message sent to a platform endpoint via SNS
+      additionalProperties: false
+      properties:
+        TargetArn:
+          type: string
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/openapi.yaml
```diff
@@ -514,8 +588,6 @@ paths:
       - $ref: '#/components/parameters/SnsPhoneNumber'
       responses:
         '204':
-          content:
-            text/plain: {}
```
viren-nadkarni, line 518: Q: is this because the body is empty? If so, is there another way to state the response mimetype?

viren-nadkarni (approved):
LGTM :heavy_check_mark: Thanks for expanding the test coverage!

PR #11624, opened 2024-10-02
Fix: CFn improperly formats account_id causing IAM engine to fail when using cdk

Morijarti:
## Motivation
- Currently CFn will inproperly format account id to int `0` instead of keeping it `000000000000` string

## Changes
- Removed casting of account_id to int

## Testing
- Running `cdklocal boostrap` with IAM enabled and then trying to deploy a stack will cause IAM to crash

## TODO

What's left to do:

- [ ] make sure that other account ids are not cast to int (when using cross account permissions in template)
- [ ] validate that boostraping and template deploying works with IAM enabled


pinzon (approved):
LGTM

PR #11629, opened 2024-10-03
Fix streaming poller performance for low shard counts in ESM v2

dfangl:
## Motivation
We currently wait 1 extra second after each iteration over all shards, since the poll_events method just returns at the end of the iterator.

Instead, we should just directly restart with the first shard in our shard list, without sleeping another second.

For single shard operations, like dynamodb streams, this effectively doubles the performance, and puts it on par with the old ESM implementation again.

## Changes
* Improve stream poller performance for lambda ESM v2 (the new default)

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -125,17 +125,17 @@ class StreamPoller(Poller):
             self.iterator_over_shards = iter(self.shards.items())
 
         current_shard_tuple = next(self.iterator_over_shards, None)
-        if current_shard_tuple:
-            try:
-                self.poll_events_from_shard(*current_shard_tuple)
-            # TODO: implement exponential back-off for errors in general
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -125,17 +125,17 @@ def poll_events(self):
...
-        if current_shard_tuple:
-            try:
-                self.poll_events_from_shard(*current_shard_tuple)
-            # TODO: implement exponential back-off for errors in general
-            except PipeInternalError:
-                # TODO: standardize logging
-                # Ignore and wait for the next polling interval, which will do retry
-                pass
-        else:
-            # Set current shard iterator to None to re-start round-robin at the first shard
```
joe4dev, line 137: that was the root cause 😢

joe4dev (approved):
LGTM, thank you @dfangl for helping out here 🙏 

Basically fixes a 2x performance regression for stream pollers 🩹

PR #11631, opened 2024-10-03
Moving OASPlugin definition

giograno:
## Motivation
We figured out, from our [CLI smoke tests](https://github.com/localstack/localstack-cli/actions/runs/11163193747/job/31029543533), that having the OAS base plugin in the handler would cause CLI import errors, this one not-installing runtime dependencies (e.g., `rolo`).

## Changes
- Moving the plugin definition in `plugins.py`.


alexrashed (approved):
Thanks a lot for jumping on this!🦸

PR #11633, opened 2024-10-03
fix Events v2 APIGW target test and update snapshots

bentsku:
## Motivation
When running a new test in Events v2 with APIGW NG enabled, I've encountered test failures due to bad replacement in the snapshots (it would replace the `HTTP` value across the whole snapshot string, leading to very weird error in `ResponseMetadata`), see the run here: https://app.circleci.com/pipelines/github/localstack/localstack/28547/workflows/b21fae5b-918f-4356-8a18-51b324914127/jobs/249483

I've refactored the snapshot logic a bit in order to delete less information from it, for us to be able to validate some assumptions in the `requestContext`, and have a clear idea of what is wrong/need to be fixed in the snapshot via the `skip_verify`. 

It showed that `events` will send specific headers to APIGW, like `X-Amz-Source-Arn` and `X-Amz-Source-Account` which might be used by the user's API Gateway integration. As a side-note, it also showed `events` is directly using the Java SDK to send/sign the requests 😄 

## Changes
- update the transformers to use v1 lambda proxy transform instead of v2
- added snapshot skips instead of transforming the whole problematic payload without replacement
- updated the Python runtime used in the lambda

\cc @cloutierMat

### DIFF
--- a/tests/aws/services/events/test_events_targets.py
+++ b/tests/aws/services/events/test_events_targets.py
@@ -7,9 +7,6 @@ import time
 
 import aws_cdk as cdk
 import pytest
-from localstack_snapshot.snapshots.transformer import (
-    JsonpathTransformer,
-)
 
 from localstack import config
 from localstack.aws.api.lambda_ import Runtime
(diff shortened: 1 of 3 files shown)


maxhoheiser (approved):
Great catch and thank you for fixing this looks not so easy to figure out what to replace exactly for the new APIGW 👍

Issue #11617, opened 2024-10-01
bug: SNS Topic does not respect Content-Type for Subscribed HTTP Endpoint


hop-along-polly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using this command
```
awslocal sns publish \
    --topic-arn arn:aws:sns:us-east-1:000000000000:local-events \
    --message '{ "clientId": "13471337" }' \
    --message-structure json
```

I am able to successfully publish a message to SNS running in Localstack. The HTTP endpoint I've subscribed to the `local-events` SNS topic receives the notification, but the message body is empty i.e. `{}`.

### Expected Behavior

I would expect the Request Body in my HTTP Endpoint to be `'{ "clientId": "13471337" }'` OR `{ "message": "13471337" }`

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

**docker-compose.yml** file
```yaml
version: '3.8'

services:
  my-api:
    build:
      context: .
    ports:
      - 3000:3000
    volumes:
      - ./src:/opt/my-api/src

  localstack:
    image: localstack/localstack
    container_name: localstack
    ports:
      - 4566:4566
    environment:
      - SERVICES=sns
      - AWS_DEFAULT_REGION=us-east-1
    volumes:
      - localstack-data:/var/lib/localstack
      # https://docs.localstack.cloud/references/init-hooks/
      - ./localstack-init-scripts/:/etc/localstack/init/ready.d/ # init.py script included below for reference.

volumes:
  localstack-data:
```

**localstack-init-scripts/init.py**
```python
import boto3
import requests

# SNS Client
sns = boto3.client('sns', endpoint_url='http://localhost:4566', region_name='us-east-1')

# Create Topic
topic_arn = sns.create_topic(Name='local-events')['TopicArn']
print('Topic Arn:', topic_arn)

#Subscribe HTTP endpoint to Topic
subscription_arn = sns.subscribe(
    TopicArn=topic_arn,
    Protocol='http',
    Endpoint='http://host.docker.internal:3000/events',
    # I've also tried Endpoint='http://my-api:3000/events',
    ReturnSubscriptionArn=True
)['SubscriptionArn']
print('Subscription Arn:', subscription_arn)

# Get the Subscription Token and Confirm the Subscription
[...]

PR #11634, opened 2024-10-03
SNS: implement Content-Type logic for http/https target

bentsku:
## Motivation
As reported with #11617, we did not properly use the configured Content-Type when sending `Notification` message type. 

I've tried to implement some CRUD validation around `Topic` attributes, but it needs a lot of it, and everything lives in moto. Either we move away from it as most of it is mocked only and implement it in LocalStack, or we need to implement the feature there.

The fix/feature implemented here should be pretty solid, the only issue is that we can accept any format and users might not understand straight away why it doesn't work when the `DeliveryPolicy` they passed would fail when trying to set it, against AWS. 

This page explains it pretty well: https://docs.aws.amazon.com/sns/latest/dg/sns-message-delivery-retries.html#creating-delivery-policy

## Changes
- for `http` and `https` targets, properly fetch the configured `Content-Type` to send.
- add a CRUD test around Topic Attributes as I was initially thinking of improving it, but this would need too much time, so it is skipped for now

_fixes #11617_

### DIFF
--- a/tests/aws/services/sns/test_sns.snapshot.json
+++ b/tests/aws/services/sns/test_sns.snapshot.json
@@ -4181,5 +4181,619 @@
         }
       }
     }
+  },
+  "tests/aws/services/sns/test_sns.py::TestSNSTopicCrud::test_topic_delivery_policy_crud": {
+    "recorded-date": "03-10-2024, 21:46:17",
+    "recorded-content": {
+      "get-topic-attrs": {
+        "Attributes": {
+          "ContentBasedDeduplication": "false",
+          "DeliveryPolicy": {
+            "http": {
+              "disableSubscriptionOverrides": false,
+              "defaultRequestPolicy": {
+                "headerContentType": "application/json"
+              }
+            }
+          },
+          "DisplayName": "TestTopic",
+          "EffectiveDeliveryPolicy": {
+            "http": {
+              "defaultHealthyRetryPolicy": {
+                "minDelayTarget": 20,
+                "maxDelayTarget": 20,
+                "numRetries": 3,
+                "numMaxDelayRetries": 0,
+                "numNoDelayRetries": 0,
+                "numMinDelayRetries": 0,
+                "backoffFunction": "linear"
+              },
+              "disableSubscriptionOverrides": false,
+              "defaultRequestPolicy": {
+                "headerContentType": "application/json"
+              }
+            }
+          },
+          "FifoTopic": "true",
+          "Owner": "111111111111",
+          "Policy": {
+            "Version": "2008-10-17",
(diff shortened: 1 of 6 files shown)


tests/aws/services/sns/test_sns.snapshot.json
```diff
@@ -4181,5 +4181,619 @@
...
+        "Timestamp": "date",
+        "TopicArn": "arn:<partition>:sns:<region>:111111111111:<resource:1>",
+        "Type": "Notification",
+        "UnsubscribeURL": "<unsubscribe-domain>/?Action=Unsubscribe&SubscriptionArn=arn:<partition>:sns:<region>:111111111111:<resource:1>:<resource:2>"
+      },
+      "http-message-headers": {
+        "Accept-Encoding": "gzip,deflate",
+        "Connection": "connection",
+        "Content-Length": "content--length",
+        "Content-Type": "application/json",
```
bentsku, line 4789: this is the important part of the fix: properly received content-type

tests/aws/services/sns/test_sns.snapshot.json
```diff
@@ -4181,5 +4181,619 @@
...
+        "Timestamp": "date",
+        "Token": "<token:1>",
+        "TopicArn": "arn:<partition>:sns:<region>:111111111111:<resource:1>",
+        "Type": "SubscriptionConfirmation"
+      },
+      "http-confirm-sub-headers": {
+        "Accept-Encoding": "gzip,deflate",
+        "Connection": "connection",
+        "Content-Length": "content--length",
+        "Content-Type": "text/plain; charset=UTF-8",
```
bentsku, line 4767: this shows that the `SubscriptionConfirmation` ignores the configured content-type

tests/aws/services/sns/test_sns.snapshot.json
```diff
@@ -4181,5 +4181,619 @@
...
+        "User-Agent": "Amazon Simple Notification Service Agent",
+        "X-Amz-Sns-Message-Id": "<uuid:1>",
+        "X-Amz-Sns-Message-Type": "SubscriptionConfirmation",
+        "X-Amz-Sns-Topic-Arn": "arn:<partition>:sns:<region>:111111111111:<resource:1>"
+      },
+      "http-message-headers-raw": {
+        "Accept-Encoding": "gzip,deflate",
+        "Connection": "connection",
+        "Content-Length": "content--length",
+        "Content-Type": "text/csv",
```
bentsku, line 4616: same, shows the fix

cloutierMat (approved):
:rocket: Nice fix. The testing is really great. Good attention to details to implement both subscription attributes and topic atributes!

Issue #11639, opened 2024-10-04
bug: Downloaded package archives are left in the docker image


kjagiello:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have noticed that the OpenJDK archive is left behind in the published docker image in `/tmp/localstack`.

### Expected Behavior

No archives left in `/tmp/localstack`.

### Steps To Reproduce

```sh
$ docker run --entrypoint bash -it localstack/localstack:3.8.0
root@40c78125f557:/opt/code/localstack# ls -lah /tmp/localstack
total 184M
drwxrwxrwx 1 root root  116 Oct  3 19:57 .
drwxrwxrwt 1 root root   50 Oct  3 19:57 ..
-rw-r--r-- 1 root root    0 Oct  3 17:38 .marker
-rw-r--r-- 1 root root 184M Oct  3 19:57 OpenJDK11U-jdk_aarch64_linux_hotspot_11.0.24_8.tar.gz
drwxr-xr-x 1 root root    0 Oct  3 19:57 state
```

### Environment

```markdown
- LocalStack:
  LocalStack version: 3.8.0
  LocalStack Docker image sha: sha256:f1db6c829fd3ab8db21007e91f62b6f76396722fcbaf6d223a8110f42482fd7e
  LocalStack build date: 2024-10-03
  LocalStack build git hash: 80b630c24
```

viren-nadkarni:
Resolved by #11640

PR #11640, opened 2024-10-04
Remove downloaded package archives after extracting them

kjagiello:
## Motivation

As mentioned in https://github.com/localstack/localstack/issues/11639, I have noticed that `python -m localstack.cli.lpm install` is leaving behind the downloaded package archives. This results in those files ending up in the published Docker image.

```
$ docker run --entrypoint bash -it localstack/localstack:3.8.0
root@40c78125f557:/opt/code/localstack# ls -lah /tmp/localstack
total 184M
drwxrwxrwx 1 root root  116 Oct  3 19:57 .
drwxrwxrwt 1 root root   50 Oct  3 19:57 ..
-rw-r--r-- 1 root root    0 Oct  3 17:38 .marker
-rw-r--r-- 1 root root 184M Oct  3 19:57 OpenJDK11U-jdk_aarch64_linux_hotspot_11.0.24_8.tar.gz
drwxr-xr-x 1 root root    0 Oct  3 19:57 state
```

## Changes

This PR solves the issue by removing the archive after extracting a package. This should result in a 184 MB reduction of the localstack docker image.

## Testing

1. Build the image: 
```sh
IMAGE_NAME="localstack/localstack" ./bin/docker-helper.sh build
```
3. Make sure that there are no archives in `/tmp/localstack`:
```sh
docker run --entrypoint ls -it localstack/localstack:latest -lah /tmp/localstack
```

### DIFF
--- a/localstack-core/localstack/packages/core.py
+++ b/localstack-core/localstack/packages/core.py
@@ -110,12 +110,14 @@ class ArchiveDownloadAndExtractInstaller(ExecutableInstaller):
         mkdir(target_directory)
         download_url = self._get_download_url()
         archive_name = os.path.basename(download_url)
+        archive_path = os.path.join(config.dirs.tmp, archive_name)
         download_and_extract(
             download_url,
             retries=3,
-            tmp_archive=os.path.join(config.dirs.tmp, archive_name),
+            tmp_archive=archive_path,
             target_dir=target_directory,
         )
+        rm_rf(archive_path)
         if self.extract_single_directory:
             dir_contents = os.listdir(target_directory)
             if len(dir_contents) != 1:


kjagiello:

alexrashed (commented):
Thanks a lot for your contribution, @kjagiello!
Your removal actually looks fine, but I think we should tackle the root cause here.
It seems like the new Java package installer (introduced with #11139) does not properly clean up after itself. /cc @viren-nadkarni 
I think we should actually remove the archive here after the extraction of the newly downloaded archive here: https://github.com/localstack/localstack/blob/958890a40ed4f5f1419a4cc69a85baa5c159c5cc/localstack-core/localstack/packages/core.py#L108-L128
@viren-nadkarni, what do you think?

viren-nadkarni:
Great catch @kjagiello !

I agree with @alexrashed , this should happen in the packaging module right after extraction is complete. This will fix the issue not just for java but also for other packages.

alexrashed:
@kjagiello, do you think you could adjust your PR such that it fixes the `ArchiveDownloadAndExtractInstaller` instead of the removing the leftover in the `Dockerfile`? Otherwise I can take over if you want.

kjagiello:
Haha that was actually the first approach I went to, then changed my mind, because I (wrongly) assumed it was intentional on your end. Should have stuck with my gut feeling 😄 I will adjust the PR shortly.

alexrashed (approved):
Awesome, thanks for jumping on the comment and adjusting the PR! The code is simple and clean. 💯 
@viren-nadkarni Anything from your side that we should maybe keep in mind here?

viren-nadkarni (approved):
LGTM, thanks for the fix :raised_hands:

PR #11648, opened 2024-10-07
ESM-v2: fix IAM service principal

gregfurman:
## Motivation
Addresses IAM regressions raised in new ESM v2.

## Changes
- Change the service principal from Pipes to Lambda.


dfangl (approved):
LGTM! The key point here is that there should be no IAM enforcement when lambda ESM invokes the lambda it is attached too! It happens within the service, and there is no permission for that.

PR #11651, opened 2024-10-07
APIGW: move current provider in legacy module

bentsku:
## Motivation
In order to validate a future switch, making the currently named "NextGen" provider default in the next release, this PR migrates the actually current provider in a `legacy` module in order to clearly separate the "old" and "new" code and help up see what parts of the code currently relies on the old implementation. 

## Changes
- move all code related to current provider in `legacy` module


localstack-core/localstack/utils/aws/client_types.py
alexrashed, line 1: This doesn't seem to be related to this PR, is it?

localstack-core/localstack/services/apigateway/helpers.py
alexrashed, line 1: Great to see that so much of the - a bit chaotic - "helpers" code is going to be removed with the legacy provider! 💯

alexrashed (commented):
Awesome! Really great to see that we're already preparing to sunset the old provider! 😛
From my side this looks great, but:
- I think a change for DMS slipped into this PR.
- It would be great if you could rebase the PR before merging, there was a lot going on in the last few days, and the pipeline was super fragile. I would like to avoid any new issues 😅

localstack-core/localstack/utils/aws/client_types.py
bentsku: Right yes, it's half related because DMS was actually using a utility from API Gateway which it shouldn't, but I'll file a different PR for that.

bentsku: PR for it: #11679

cloutierMat (approved):
Superb. I agree with Alex, the removal of so many helpers is great to see!

Thanks for addressing the comment! :pray:

alexrashed (approved):
Thanks for addressing the comments! Lookin' good! Nice preparation and cleanup!

PR #11658, opened 2024-10-09
Move dns server import inside resolving function to avoid import in CLI

dfangl:
## Motivation
With #11581, we introduced the option to bypass our own DNS for certain AWS clients.
However, importing the DNS server from the connect.py module leads to issues for older python versions in the CLI, as the CLI needs to be able to import connect.py, and the DNS server uses syntax only supported in later python versions.

## Changes
* Using python versions 3.8 and 3.9, using the CLI should be possible again without issues (this was never actually published, so it only affected development versions)


alexrashed (approved):
Awesome! Thanks so much for jumping on this! 🚀 🥳 🦸🏽

PR #11659, opened 2024-10-09
Bump moto-ext to 5.0.17.post1

viren-nadkarni:
## Summary

This PR bumps moto-ext to 5.0.17.post1

Contains:

- https://github.com/getmoto/moto/pull/8162
- https://github.com/getmoto/moto/pull/8220
- https://github.com/getmoto/moto/pull/8223
- https://github.com/getmoto/moto/pull/8216

## To do

- [x] Ext integration tests AWS / Ext Integration Tests # 1386
- [x] Randomised account ID and region run https://app.circleci.com/pipelines/github/localstack/localstack/28811/workflows/17343c89-d0bb-4290-a997-36d75adac361


bentsku:
I've pushed https://github.com/localstack/localstack/pull/11659/commits/bb272e0109f0fea6f56677a605293581f55e4017 to fix a regression in Moto introduced by https://github.com/getmoto/moto/pull/8202, I'll also open a PR in moto to fix it, but figured it'd be easier here, as we are planning to move away, so the more logic we have in the provider the better.

requirements-base-runtime.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.12
```
bentsku, line 2: I don't think this should be the case just yet 😄

bentsku:
I've rebased the PR following #11651, and added the last fix in the unit tests

requirements-base-runtime.txt
viren-nadkarni: I upgraded to Ubuntu 24.04 yesterday, and yes Python 3.12 makes more than just a cosmetic impact.

Reverted with 27dde674a :+1:

alexrashed: Pro-tip: Use `pyenv`. We do have `.python-version` files in our repos, which controls `pyenv` to automatically switch to the right version. :)

alexrashed (approved):
LGTM! Thanks for keeping moto(-ext) up to date, and thanks for the great team-work to fix incompatibilities! 💯

Issue #11553, opened 2024-09-20
feature request: Support for ValidateStateMachineDefinition since this now blocks upgrading AWS Teraform provider to 5.67.0


tonestrike:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

On 9/12/24, [version 5.67.0 of the terraform-provider-aws](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.67.0) was released which added support for plan time step function definition validation.

```
resource/aws_sfn_state_machine: Add plan-time validation of definition using the AWS Step Functions [Validation API](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ValidateStateMachineDefinition.html) (https://github.com/hashicorp/terraform-provider-aws/issues/39229)
```

Since this relies on the `ValidateStateMachineDefinition` action, upgrading to the latest Terraform AWS provider causes any existing Step Function definition defined in Terraform to break.

Lack of support for this operation is documented [here](https://docs.localstack.cloud/references/coverage/coverage_stepfunctions/). It would be great to not have to pin the AWS provider version in order to deploy local step functions using Terraform with Localstack.

MEPalma:
Thank you for taking the time to compile this report. I was able to replicate the behavior that resulted in the failure. We recently merged changes that introduced base support for ValidateStateMachineDefinition, which should unlock the workflow described. While more thorough definition validation logic for state machines is still being worked on, this initial implementation should address your immediate issue. Could you verify if this resolves your issue? If it does, feel free to close this issue. If you're still experiencing problems, please let us know and we'll investigate further. Thank you again for bringing this issue forward.

PR #11660, opened 2024-10-09
[SFN] Base Support for ValidateStateMachineDefinition

MEPalma:
## Motivation
Upgrading to the latest Terraform AWS provider is currently not possible because the Step Functions v2 provider lacks support for the [ValidateStateMachineDefinition](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ValidateStateMachineDefinition.html) API action. This pull request introduces basic support for this API action, reporting any currently detectable issues (for STANDARD and EXPRESS) in the diagnostics field and reflecting the validation outcome in the result field. Addresses: #11553

## Changes
- Added base support for ValidateStateMachineDefinition in the v2 provider
- Added base positive and negative snapshot tests (here I intentionally used simple pytest parametrisation to ease selective testing as done in other suites for the SFN provider)


tests/aws/services/stepfunctions/v2/test_sfn_api_validation.py
```diff
@@ -0,0 +1,70 @@
...
+    CallbackTemplates,
+)
+from tests.aws.services.stepfunctions.templates.validation.validation_templates import (
+    ValidationTemplate,
+)
+
+
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        "$..tracingConfiguration",
```
dominikschubert, line 17: Any particular reason for this here? I didn't see that key anywhere in the recorded snapshots below

dominikschubert (approved):
Thanks for immediately jumping on this and implementing basic support here. Definitely an area that will be interesting to see growing and aligning our validation with that done by AWS. :+1:

PR #11661, opened 2024-10-09
ESM v2: Allow target Lambda response payloads to not only be JSON

gregfurman:
## Motivation
This PR accounts for the case where an invoked Lambda with an Event Source Mapping is invalid JSON.

## Changes
- Checks if a payload from an invocation fails to be decoded into JSON, setting the payload value to `None` if an exception is raised.
- Add some testing for when no value is returned by a Lambda.

## TODO

What's left to do:

- [ ] Update EventSourceMapping configuration to relfect additional [EventSourceMappingArn](https://docs.aws.amazon.com/lambda/latest/api/API_CreateEventSourceMapping.html#lambda-CreateEventSourceMapping-response-EventSourceMappingArn) field -- since our current ESM parity tests cannot be extended to account for the above case.

### DIFF
--- a/tests/aws/services/lambda_/functions/lambda_none.js
+++ b/tests/aws/services/lambda_/functions/lambda_none.js
@@ -1,0 +1,3 @@
+exports.handler = async (event) => {
+   console.log(JSON.stringify(event))
+}
--- a/tests/aws/services/lambda_/functions/lambda_none.py
+++ b/tests/aws/services/lambda_/functions/lambda_none.py
@@ -1,0 +1,7 @@
+import json
+
+
+def handler(event, context):
+    # Just print the event that was passed to the Lambda and return nothing
+    print(json.dumps(event))
+    return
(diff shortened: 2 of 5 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/senders/lambda_sender.py
```diff
@@ -59,7 +59,16 @@ def send_events(self, events: list[dict] | dict) -> dict:
             InvocationType=invocation_type,
             **optional_qualifier,
         )
-        payload = json.load(invoke_result["Payload"])
+
+        try:
+            payload = json.load(invoke_result["Payload"])
+        except json.JSONDecodeError:
+            payload = None
```
gregfurman, line 66: Not sure if we should be setting this to `None` -- what if the payload is valid but just not JSON?

dfangl (approved):
LGTM! Let's tackle explicit testing in a follow up

PR #11662, opened 2024-10-09
Duplicate Flags in  Makefile

vignesh1507:
Duplicate Flags: In the `upgrade-pinned-dependencies` target, `--strip-extras` is specified twice in the `pip-compile` commands:

```bash
$(VENV_RUN); pip-compile --strip-extras --upgrade --strip-extras -o requirements-basic.txt pyproject.toml
```


bentsku (changes requested):
Hello and thanks for your contribution! Are you using an automated tool for grammar? It seems it was a bit too eager and changed things that do not make sense anymore. Could you revert the commented changes? 

I'll ping @silv-io for the change in the Makefile in particular!

silv-io (changes requested):
The change in the Make target looks good!

Before merging, I would just ask to remove all the other changes (seems like they were automated?) as they don't seem to be connected to the PR description and goal.

vignesh1507:
ok then shall i delete the files except the makefile one?

silv-io:
@vignesh1507 yes, please just remove the changes outside of the Makefile :)

vignesh1507:
Made a new PR (#11666 ) with only the makefile commit.

PR #11666, opened 2024-10-10
Remove duplicate --strip-extras flag in dependency pinning target

vignesh1507:
Duplicate Flags: In the `upgrade-pinned-dependencies` target, `--strip-extras` is specified twice in the `pip-compile` commands:

```bash
$(VENV_RUN); pip-compile --strip-extras --upgrade --strip-extras -o requirements-basic.txt pyproject.toml
```


silv-io (approved):
LGTM now!

PR #11667, opened 2024-10-10
Add LD_LIBRARY_PATH to JavaInstallerMixin

viren-nadkarni:
## Background

https://github.com/localstack/localstack-ext/pull/3419 removed the global default JRE installation. This caused certain tests to start failing for aarch64.

## Changes

This PR sets the `LD_LIBRARY_PATH` env var with Java libs and adds it to the `JavaInstallerMixin`.

Generally it is advisable to set `java.library.path` on the JVM's command line. But this approach should be a global fix. Besides this used to be set in the Dockerfile in the era of global JRE.

## Tests

This was detected with following failure:

```
FAILED tests/aws/services/qldb/test_qldb.py::TestQLDB::test_stream_journal - Exception: Timeout when waiting for KCL initialization.
```

Tested passing with following CI run:

https://github.com/localstack/localstack-ext/actions/runs/11272853744/job/31350949612 :green_circle:

### DIFF
--- a/localstack-core/localstack/packages/java.py
+++ b/localstack-core/localstack/packages/java.py
@@ -39,12 +39,13 @@ class JavaInstallerMixin:
         """
         return java_package.get_installer().get_java_home()
 
-    def get_java_env_vars(self, path: str = None) -> dict[str, str]:
+    def get_java_env_vars(self, path: str = None, ld_library_path: str = None) -> dict[str, str]:
         """
         Returns environment variables pointing to the Java installation. This is useful to build the environment where
         the application will run.
 
         :param path: If not specified, the value of PATH will be obtained from the environment
+        :param ld_library_path: If not specified, the value of LD_LIBRARY_PATH will be obtained from the environment
         :return: dict consisting of two items:
             - JAVA_HOME: path to JRE installation
             - PATH: the env path variable updated with JRE bin path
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/packages/java.py
```diff
@@ -56,6 +56,7 @@ def get_java_env_vars(self, path: str = None) -> dict[str, str]:
 
         return {
             "JAVA_HOME": java_home,
+            "LD_LIBRARY_PATH": f"{java_home}/lib:{java_home}/lib/server",
```
alexrashed, line 59: This would effectively overwrite an existing `LD_LIBRARY_PATH` wherever we use the Java mixin / this function. I think this should be handled the same way as for the path.

alexrashed (approved):
Nice catch! Thanks so much for digging into this! 🦸🏽 
I only added one comment, when this is addressed you can directly move forward and merge this PR! 💯

localstack-core/localstack/packages/java.py
viren-nadkarni: @alexrashed but shouldn't this be treated more like `JAVA_HOME` and localised to the Java environment to be used? An invalid exisiting `LD_LIBRARY_PATH` value might potentially poison the environment.

alexrashed: Sure, there definitely shouldn't be a faulty `LD_LIBRARY_PATH`, but couldn't it be necessary to have the shared libraries of Java as well as other libraries in the `LD_LIBRARY_PATH`?
For example in OpenSeach we add custom libraries for KNN to the `LD_LIBRARY_PATH` here:
https://github.com/localstack/localstack/blob/de272085dbdbbd594f797805e71ac2b2fd126749/localstack-core/localstack/services/opensearch/cluster.py#L479

I know that for opensearch we use the Java runtime from the package, but that would be an example where we would have to have multiple paths on the `LD_LIBRARY_PATH`.

PR #11668, opened 2024-10-10
There is a small issue in the html segment of diff_metrics_coverage.py

vignesh1507:
There is a small issue in the HTML generation section at the end of the `create_readable_report` function. The closing `>` for the last `fd.write("</body></html")` is missing. It should be `fd.write("</body></html>")`.


vignesh1507:
@ silv-io can you review this PR?

silv-io (approved):
Hi @vignesh1507, sorry for not getting back to you sooner, I didn't receive the ping because of the space between my user name and the @. 

The change looks good to me, so I'll merge it right away!

PR #11669, opened 2024-10-10
Update retrieve_legacy_data_from_circleci.py

vignesh1507:
Response Code Handling: In the `send_request_to_connection` function, the error message when the response fails should use parentheses: `print(f"connection failed: {res.getcode()}")` instead of `print(f"connection failed: {res.getcode}")`.

### DIFF
--- a/scripts/tinybird/retrieve_legacy_data_from_circleci.py
+++ b/scripts/tinybird/retrieve_legacy_data_from_circleci.py
@@ -34,7 +34,7 @@ def send_request_to_connection(conn, url):
         data = res.read()
         return data
     else:
-        print(f"connection failed: {res.getcode}")
+        print(f"connection failed: {res.getcode()}")
         return None
 
 


vignesh1507:
@ silv-io & @olleolleolle  can you review this PR?

scripts/tinybird/retrieve_legacy_data_from_circleci.py
```diff
@@ -34,7 +34,7 @@ def send_request_to_connection(conn, url):
         data = res.read()
         return data
     else:
-        print(f"connection failed: {res.getcode}")
+        print(f"connection failed: {res.getcode()}")
```
olleolleolle, line 37: This fixes the bug. Perhaps a local variable is useful for this value?

olleolleolle (commented):
Calling the method using `()` probably makes it work.

The PR title could reflect why we made this change (well, to fix a bug in error reporting).

silv-io:
@vignesh1507 can you please give the clarification asked for in the review? Afterwards we can merge this PR

alexrashed:
Since this script is not part of the actual LocalStack AWS Community Emulator, but an auxiliary utils script, and there hasn't been a response, I'll close this PR for now.

PR #11670, opened 2024-10-10
APIGW: improve parity for GetApiKeys and GetUsagePlanKeys

bentsku:
## Motivation
Based on #11659 to not have conflicts.

As there was a small regression with #11659, I've improved the existing test and added one regarding UsagePlanKeys, and improved the logic in the provider. 

Seems like current test failure is due to targeting another branch, I'll see when the bump is merged. 

## Changes
- improve existing API keys tests
- add new test for UsagePlanKeys
- fix provider behavior


cloutierMat (approved):
Awesome! Thanks @bentsku for fixing and being on top of the moto bump! :rocket:

Issue #11638, opened 2024-10-04
feature request: Support KMS DeriveSharedSecret operation


rbonestell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Our software and team rely heavily on the `DeriveSharedSecret` operation in AWS KMS to securely derive shared secrets as part of our encryption and data security protocols. Currently, LocalStack KMS does not support this operation which forces us to implement additional "local only" code to handle this functionality during local development and testing. This adds unnecessary complexity and divergence from production environments, as well as the potential for bugs.

https://docs.aws.amazon.com/kms/latest/APIReference/API_DeriveSharedSecret.html

We would greatly benefit from LocalStack supporting the `DeriveSharedSecret` operation in its KMS implementation, as this would allow us to align our local testing environments more closely with AWS production environments, reduce overhead in our development processes, and streamline our local testing workflows.

Thank you for considering this feature request, and we appreciate all the work done to make LocalStack such a valuable tool for AWS emulation!

sannya-singal:
Hey @rbonestell 👋 Thanks for creating a feature request. We now have an initial implementation of `DeriveSharedSecret` operation in KMS in the `latest` image. Please do not hesitate to contact us if you have any further questions or encounter new issues.

rbonestell:
I've gotten around to implementing `DeriveSharedSecret` using LocalStack 4 Pro and it is not behaving with parity to AWS KMS. When I upload key material for the following key params, decryption using LocalStack KMS results fail to validate the signed message tag after deriving the shared key. This demonstrates that the derived key result from LocalStack doesn't match that derived by AWS KMS.

```
{
  keyParams: {
    Origin: "EXTERNAL",
    KeySpec: "ECC_NIST_P256",
    KeyUsage: "KEY_AGREEMENT",
  },
}
```

When debugging the code which uses the AWS SDK to execute the `DeriveSharedSecret` for the same data, same public key, and the private key, which was created and key material uploaded in the same manner to both AWS and LocalStack KMS, the resulting derived shared secrets are completely different from this simple code:

```
const command = new DeriveSharedSecretCommand({ 
	KeyAgreementAlgorithm: KeyAgreementAlgorithmSpec.ECDH,
	KeyId: keyId,
	PublicKey: publicKey
 });
const response = await this.kmsClient.send(command);
```

PR #11672, opened 2024-10-11
kms: implement DeriveSharedSecret operation  

sannya-singal:
## Changes

Related feature request: https://github.com/localstack/localstack/issues/11638


localstack-core/localstack/services/kms/models.py
```diff
@@ -361,6 +365,27 @@ def verify(
...
+        if key_spec in [KeySpec.ECC_NIST_P256, KeySpec.ECC_SECG_P256K1]:
+            algorithm = hashes.SHA256()
+        elif key_spec == KeySpec.ECC_NIST_P384:
+            algorithm = hashes.SHA384()
+        elif key_spec == KeySpec.ECC_NIST_P521:
+            algorithm = hashes.SHA512()
+        else:
+            raise InvalidKeyUsageException(
+                f"{self.metadata['Arn']} key usage is {self.metadata['KeyUsage']} which is not valid for DeriveSharedSecret."
+            )
```
silv-io, line 379: This would be a nice place for a `match case` statement I think, but it's fine and easily legible like this as well

localstack-core/localstack/services/kms/models.py
```diff
@@ -616,14 +641,27 @@ def _get_key_usage(self, request_key_usage: str, key_spec: str) -> str:
...
+                KeySpec.ECC_NIST_P384,
+                KeySpec.ECC_NIST_P521,
+                KeySpec.ECC_SECG_P256K1,
+                KeySpec.SM2,
+            ]:
+                raise ValidationException(
+                    f"KeyUsage {request_key_usage} is not compatible with KeySpec {key_spec}"
+                )
+            else:
+                return request_key_usage
```
silv-io, line 664: This `elif` block is in the same indentation as a check for the key spec. The other such `elif` blocks which check the key usage are inside the topmost `if` statement. 

It might be perfectly valid like this. I just want to point it out as a sanity check :)

silv-io (approved):
LGTM! Just a comment and a question :)

localstack-core/localstack/services/kms/models.py
sannya-singal: The key spec for the first part makes sure to validate the HMAC keys and the recently added part will verify only the `KEY_AGREEMENT` usage's key specifications. In either case they are mutually exclusive. Does this answer your ques?

PR #11676, opened 2024-10-11
Test force deletion of deleted secrets, force delete secrets in CFN by default

dfangl:
## Motivation
AWS allows force-deletion of already deleted secrets, but LocalStack does not.

Also, CloudFormation will per default force-delete a secret, so they do not show up anymore after stack deletion.

The respective change for allowing force deletion of secrets already marked for deletion in moto is getmoto/moto#8223.

## Changes
* LocalStack will allow force deletion of secrets already marked for deletion
* Cloudformation will force delete secrets, instead of just marking them for deletion.


dfangl:
This PR previously contained changes now in #11680, as those require a moto change. I split it up to be able to merge this one faster, as they are only loosely related.

pinzon (approved):
another un-skipped test :partying_face:

PR #11679, opened 2024-10-11
add DMS ServicePrincipal

bentsku:
## Motivation
We need `dms` as a new Service Principal. 
Also, I've update the class of the `ServicePrincipal` to `StrEnum`

## Changes
- add DMS 
- use `StrEnum` instead of subclassing `str`, as done in #11176

### DIFF
--- a/localstack-core/localstack/utils/aws/client_types.py
+++ b/localstack-core/localstack/utils/aws/client_types.py
@@ -264,6 +264,7 @@ class ServicePrincipal(str):
     """
 
     apigateway = "apigateway"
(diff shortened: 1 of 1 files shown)


dfangl (approved):
LGTM!

bentsku:
@dfangl sorry for asking again for your review, I've updated the class of `ServicePrincipal` right after, just want to be sure this is okay 😄

dfangl (approved):
Using a StrEnum definitely makes sense as well, for typing, nice catch!

PR #11680, opened 2024-10-11
Test force deletion of secrets already marked for deletion

dfangl:
## Motivation
AWS allows force-deleting secrets which are already marked for deletion.
However, this is currently blocked in moto.
This PR contains an AWS validated test for this exact behavior.

The actual change to the behavior is in getmoto/moto#8223.

## Changes
* LocalStack now allows force-deletion of secrets already marked for deletion


dominikschubert (approved):
LGTM, great catch for that edge case

PR #11685, opened 2024-10-14
ESM v2: Test Lambda runtime where provided bootstrap returns nothing

gregfurman:
## Motivation
This PR tests how behaviou when a target Lambda function of an Event Source Mapping handles a Lambda runtime where nothing is returned by a Lambda invocation.

## Changes
- Adds a new parity test `test_kinesis_empty_provided` to snapshot how an ESM handles a Lambda runtime that does not send any data to the invocation API. 
   - This uses a custom `bootstrap`, that sends empty response to the runtime API on invocation
   - Custom `bootstrap` resides in `tests/aws/services/lambda_/functions/provided_bootstrap_empty`
- Skips new `EventSourceMappingArn` this field for tests where parity snapshots are not yet updated.
- Changes transformer settings in ESM's `conftest.py` to better cater for this case of multiple UUIDs being snapshot.

### DIFF
--- a/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
+++ b/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
@@ -35,6 +35,7 @@ TEST_LAMBDA_KINESIS_LOG = FUNCTIONS_PATH / "kinesis_log.py"
 TEST_LAMBDA_KINESIS_BATCH_ITEM_FAILURE = (
     FUNCTIONS_PATH / "lambda_report_batch_item_failures_kinesis.py"
 )
+TEST_LAMBDA_PROVIDED_BOOTSTRAP_EMPTY = FUNCTIONS_PATH / "provided_bootstrap_empty"
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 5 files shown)


tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
```diff
@@ -970,6 +974,71 @@ def _verify_messages_received():
...
+            StreamName=stream_name,
+        )
+
+        def _verify_invoke():
+            log_events = aws_client.logs.filter_log_events(
+                logGroupName=f"/aws/lambda/{function_name}",
+            )["events"]
+            assert len([e["message"] for e in log_events if e["message"].startswith("REPORT")]) == 1
+
+        retry(_verify_invoke, retries=30, sleep=5)
```
thrau, line 1037: i think this shows how tests could benefit from some internal endpoint that makes it easier detect lambda invocations :eyes: @dominikschubert @joe4dev

thrau (approved):
nice test! LGTM

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
joe4dev: indeed, we use many ways to assert Lambda invokes and these async assert patterns clutter dozens of test cases.

gregfurman: This would be brilliant. The current approach of using queues + searching logs seems like overkill

joe4dev: I created backlog items (for Lambda invocations and Lambda logs) and added them to the agenda of the next Lambda service owner sync 🚀

PR #11675, opened 2024-10-11
ESM v2: Add EventSourceMappingArn field

gregfurman:
## Motivation
This PR introduces increases parity with AWS by includng a new field in ESM responses.

## Changes
- Add new `EventSourceMappingArn` field to config:
  - Skips this field for tests where parity snapshots are not yet updated and for appropriate ESM v1 cases.
  - Changes transformer settings in ESM's `conftest.py` to better cater for this case of multiple UUIDs being snapshot.

PR #11686, opened 2024-10-14
ESM v2: Fix case where lambda is deleted prior to ESM resource

gregfurman:
## Motivation
This PR fixes the case where a Lambda is deleted prior to an Event Source Mapping raises an resource not found exception.

- Note: This PR relies on the changes for https://github.com/localstack/localstack/pull/11675 being merged in due to issues with AWS parity following a new field being added.

## Changes
- At create time, the state of the Lambda for which an ESM is created is now also stored within the ESM resource at `_state`. This is due to the underlying ESM state  potentially not being present at delete time of an ESM.  Stored at this lambda within the `event_source_mapping`  dictionary.
- An idempotent `delete_esm_in_store` is used to remove the ESM from the `event_source_mapping` state store. 

## Testing
- `test_event_source_mapping_lifecycle_delete_function` is created which tests the behaviour of deleting a Lambda function and then attempting to access and delete an ESM. To ensure the case is tested correctly, we monkey-patch the `delete_esm_in_store` at test time to ensure the correspondong Lambda has been deleted prior to the ESM deletion being asynchronously kicked-off.


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -5294,6 +5294,101 @@ def check_esm_active():
...
+                return True
+
+            def _delete_esm_in_store_with_delay(self):
+                wait_until(_assert_function_deleted)
+                self.delete_esm_in_store()
+
+            monkeypatch.setattr(EsmWorker, "delete_esm_in_store", _delete_esm_in_store_with_delay)
+
+        get_response_post_delete = aws_client.lambda_.get_event_source_mapping(UUID=uuid)
+        snapshot.match("get_response_post_delete", get_response_post_delete)
```
gregfurman, line 5387: FYI: Here we get the event source mapping AFTER a lambda has been deleted

dfangl: Would be interesting to see if the state would change if we would trigger some events, but the function is deleted. Not in scope for this PR though!

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -5294,6 +5294,101 @@ def check_esm_active():
...
+            def _assert_function_deleted():
+                with pytest.raises(aws_client.lambda_.exceptions.ResourceNotFoundException):
+                    aws_client.lambda_.get_function(FunctionName=function_name)
+                return True
+
+            def _delete_esm_in_store_with_delay(self):
+                wait_until(_assert_function_deleted)
+                self.delete_esm_in_store()
+
+            monkeypatch.setattr(EsmWorker, "delete_esm_in_store", _delete_esm_in_store_with_delay)
```
dfangl, line 5384: Why do we need this part? Could we just wait in the test? We do not delete the ESM anyway until later, right?

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -5294,6 +5294,101 @@ def check_esm_active():
...
+
+        uuid = create_response["UUID"]
+        cleanups.append(lambda: aws_client.lambda_.delete_event_source_mapping(UUID=uuid))
+        snapshot.match("create_response", create_response)
+
+        # the stream might not be active immediately(!)
+        def check_esm_active():
+            return aws_client.lambda_.get_event_source_mapping(UUID=uuid)["State"] != "Creating"
+
+        assert wait_until(check_esm_active)
```
dfangl, line 5362: Could we reuse `_await_event_source_mapping_enabled` here?

dfangl (commented):
Looks good, I just have some questions regarding the special case for ESM v2 in the test!

tests/aws/services/lambda_/test_lambda_api.py
gregfurman: Yeah I suppose we can simplify this test. I initially wrote it using monkeypatching to try and exactly re-create this race-condition but now that the ESM deletion code has changed it's no longer relevant.

dfangl (approved):

Issue #11599, opened 2024-09-28
bug: DynanmoStreams don't work with the SDK


toadzky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I can poll the shard for records indefinitely, even if I'm writing to the table, and no records will ever show up. I also tried enabling the kinesis configuration, but saw no results with the kinesis client either.

Based on looking through the code, the main problems seems to be that you don't actually use the built-in streams functionality in dynamodb-local. You intercept the calls and use an internal fake - which never actually spits out records to the sdk.

### Expected Behavior

Records actually show up as documented in the SDK. Ideally, you'd stop intercepting the calls and just use the dynamodb-local because it's more consistent and correct by far.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Testcontainers in Junit

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```kotlin
    var iterator: String = streams.getShardIterator { it.streamArn(streamArn).shardId(shardId).shardIteratorType(TRIM_HORIZON) }.shardIterator()

    val response = streams.getRecords { it.shardIterator(currentIterator) }
    iterator = response.nextShardIterator()

    if (response.records().isEmpty()) {
      log.debug { "No records found on stream. table=$table" }
      return
    }
    
    // do something with the records
```

### Environment

```markdown
- OS: macOS
- LocalStack Docker image sha: sha256:8569dadc82ddf0627569775dde517297351691ddbe28e60edd1f515b46f7ef2e
```

PR #11064, opened 2024-06-20
DynamoDB: Upgrade DDBLocal to v2

viren-nadkarni:
## Background

AWS released DDB Local v2 in July 2023.

<https://aws.amazon.com/about-aws/whats-new/2023/07/dynamodb-local-version-2-0/>

LocalStack currently uses DDB Local v1. The EOL for DDB Local v1 which is January 2025.

Among others, DDB Local v2 supports:
- Added support for table deletion protection and the `ReturnValuesOnConditionCheckFailure` parameter
- Support for configurable maximum throughput for on-demand tables, `ReturnValuesOnConditionCheckFailure`, `BatchExecuteStatement`, and `ExecuteTransactionRequest`

## Changes

This PR upgrades DDB Local to v2.

## To do

- [x] Restore certain skipped snapshot attributes (like `/Table/DeletionProtectionEnabled`)
- [x] Test on Mac OS M1 (verified on MacOS 14.6.1; openjdk 21.0.5)

## Related

Supersedes https://github.com/localstack/localstack/pull/9909

Depends on https://github.com/localstack/localstack/pull/11139, https://github.com/localstack/localstack/pull/11462

Closes https://github.com/localstack/localstack/issues/11038

PR #11688, opened 2024-10-15
implement new DDB streams provider using DynamoDB-Local

bentsku:
## Motivation
First commit is based on https://github.com/localstack/localstack/commit/6f294e8bbe8635661ab6e23e73914b99d17ee92b

As reported by some of our users and also in #11599, DynamoDB-local supports DynamoDB Streams out of the box. This PR remove the current logic we had to support the stream functionality (using Kinesis under the hood) and which had to often query DDB-local multiple times for one request to get the old and/or new images. 

This adds a considerable performance boost: we don't need to query ddb, and we reduce our internal usage of kinesis which reduces load on our gateway.

We only need to manipulate the headers + do some payload manipulation to modify the returned region, and update it before forwarding it to ddb local. 

These are quite big, so we are releasing them under a new provider for now behind a feature flag (the usual `PROVIDER_OVERRIDE_DYNAMODBSTREAMS=v2`). Question is, is it DynamoDB v2 or DynamoDBStreams v2? The 2 implementations are tied together, and DynamoDB is shorter, so I'd maybe be inclined to choose that one instead. 

Here's the successful run with the provider flag being forced: https://app.circleci.com/pipelines/github/localstack/localstack/28834/workflows/52d80e8f-6354-4764-b073-4f36f10fc121

## Changes
@dfangl changes:
- update the DynamoDB server to have a singleton method so that it can be shared between DDB and DDBStreams provider
- remove lot of related code to Streams in the DDB Provider
  - the modified provider is a copy-paste of the default one, without all the stream related code, it might be easier to do a diff between those 2 in order to review (`dynamodb/provider.py` <-> `dynamodb/v2/provider.py`)
- update all the code in the DynamoDB Streams provider to forward requests to ddb-local instead of accessing kinesis

@bentsku changes:
- migrated the code to be in 2 different providers and create the providers definitions
- updated the regex and added one in order to catch all possible cases with ddb streams
- update the tests to not have kinesis assumptions
- skip some fields because of ddb-local for ESM
- remove even more code in the DDB provider related to streams

## Limitations

This new implementation does not support `DynamoDB.EnableKinesisStreamingDestination` API action and logic about dispatching event to Kinesis directly yet. 

## TODO

What's left to do:

- [x] remove the commit https://github.com/localstack/localstack/pull/11688/commits/8431fa2771b4833e3e879b9098a91700d05306dd  forcing the usage of the new implementation!


thrau:
hey folks! i helped narrow down the reviewers to @dfangl and @viren-nadkarni :pray:

localstack-core/localstack/config.py
```diff
@@ -1077,6 +1077,15 @@ def populate_edge_configuration(
...
 
+# Whether the DynamoDBStreams native provider is enabled
+DDB_STREAMS_PROVIDER_V2 = os.environ.get("PROVIDER_OVERRIDE_DYNAMODBSTREAMS", "") == "v2"
+if DDB_STREAMS_PROVIDER_V2:
+    # in order to not have conflicts between the 2 implementations, as they are tightly coupled, we need to set DDB
+    # to be v2 as well
+    # TODO: to be decided if we make DDB Streams the "v2" or DDB, but the biggest change is in the streams themselves
+    if not os.environ.get("PROVIDER_OVERRIDE_DYNAMODB"):
+        os.environ["PROVIDER_OVERRIDE_DYNAMODB"] = "v2"
+
```
viren-nadkarni, line 1088: Should there be a reverse dependency? i.e. if DynamoDB v2 is used, should DynamoDBStreams v2 be used too?

localstack-core/localstack/services/dynamodbstreams/v2/provider.py
```diff
@@ -0,0 +1,74 @@
...
+        payload["ShardIterator"] = self.modify_stream_arn_for_ddb_local(
+            payload.get("ShardIterator", "")
+        )
+        return self.forward_request(context, payload)
+
+    @handler("GetShardIterator", expand=False)
+    def get_shard_iterator(
+        self, context: RequestContext, payload: GetShardIteratorInput
+    ) -> GetShardIteratorOutput:
+        payload["StreamArn"] = self.modify_stream_arn_for_ddb_local(payload.get("StreamArn", ""))
```
viren-nadkarni, line 69: Just a note: I'm not sure if modifying the request `payload` directly has side effects on the handler chain? At least for the request `context` we take precautions to restore the original state.

localstack-core/localstack/services/dynamodbstreams/v2/provider.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        DynamoDBProvider.prepare_request_headers(
+            context.request.headers, account_id=context.account_id, region_name=context.region
+        )
+        return self.server.proxy(context, service_request)
+
+    def modify_stream_arn_for_ddb_local(self, stream_arn: str) -> str:
+        parsed_arn = parse_arn(stream_arn)
+
+        return f"arn:aws:dynamodb:ddblocal:000000000000:{parsed_arn['resource']}"
```
viren-nadkarni, line 47: Any idea how well this plays when non-default account/region is used in the request context?

Would the ARN in the response returned by DDBLocal need to be updated too?

viren-nadkarni (approved):
This is a fantastic change :clap: Kudos!

I'm not sure I like the v1 provider code copied but it makes practical sense. We should deprecate the v1 provider at the earliest and make v2 the default to reduce any maintenance burden.

We also have the upgrade to DDBLocal v2 in the pipeline https://github.com/localstack/localstack/pull/11064. This will be a low-level change and should not cause confusion with regards to the `v2` marker.

localstack-core/localstack/config.py
bentsku: I think we could, yes. I thought it'd be easier to communicate about only one flag and the second one being "implicit" in a way, but may be better safe than sorry

localstack-core/localstack/services/dynamodbstreams/v2/provider.py
bentsku: Right yes, I normally never do that! Good catch!

localstack-core/localstack/services/dynamodbstreams/v2/provider.py
bentsku: No idea no, I'll give it a try quickly to see, thanks for pointing it out!

bentsku: The ARN returned by DDBLocal are being updated by the following service response handler: `localstack.services.dynamodb.utils.modify_ddblocal_arns`.

I've checked and the needed StreamArn is always the `000000000000` account id and `ddblocal` region, works properly with non default account and region 👍 thanks for checking!

PR #11691, opened 2024-10-15
logs: fix `put_metric_data client` for multi account/region

sannya-singal:
## Motivation
As an initiative to make the -ext multi-account/multi-region compatible this PR fixes functionality for logs service.
So whenever we use values other than `000000000000` for account ID or `us-east-1` for region, it should still create the consequent resources in these accounts and region.

## Changes
Add the region and account id from context to the `connect_to` client in `put_log_events` used to put the metrics data. 
This fixes: `tests.aws.services.logs.test_logs.TestCloudWatchLogsPro.test_json_metric_filters` in -ext for non-default credentials. 

## Testing
Set the environment variables
`TEST_AWS_ACCOUNT_ID=111111111111`, `TEST_AWS_ACCESS_KEY_ID=111111111111`, `TEST_AWS_REGION=us-west-1`


pinzon (approved):
LGTM :+1:

PR #11692, opened 2024-10-15
Feature: Decompose submit notification for s3 to be able to patch it in extension

maxhoheiser:
## Motivation
We need to be able to submit the threading.local context to the child thread where the notification is handled so we can pass trace context. We can do this by patching the `elf.executor.submit` method and only calling the patched method, for this we need to decompose the `send_notifications` method.


bentsku (commented):
I would like to have a bit more context regarding how the patching will be done, but the change looks alright! If it's possible to \cc me in the related PR, that would be great 😄

bentsku (approved):
LGTM!

PR #11693, opened 2024-10-15
refactor ref keys in conditional mappings and test

pinzon:
## Motivation
This PRs addresses the changes requested in #11671

## Changes
- refactoring of the reference search
- restore previous test and add a new one more focused on the refere

## Testing
- new test that asserts that Map name, top level key and second level key can be AWS:: refs


simonrw (approved):
That's great, thanks for making these changes!

PR #11694, opened 2024-10-15
remove key-existence checks against dict.keys() calls

bentsku:
## Motivation
While looking over some performance improvements for SQS, I've stumbled into a key existence checks like that `if key in dict.keys()`. 

Calling `dict.keys()` will create a dict view of the dict keys. In itself, this is not terrible, but it has a bit of a cost performance. 

Giving it a small benchmark:
<details><summary>Small benchmarking code</summary>
<p>

```python
import timeit


test_dict = {f"key{k}": k for k in range(1000)}
last_key = list(test_dict)[-1]


def check_in_keys():
    p = last_key in test_dict.keys()


def check_in_dict():
    p = last_key in test_dict


def check_not_in_keys():
    p = last_key not in test_dict.keys()


def check_not_in_dict():
    p = last_key not in test_dict
```

</p>
</details> 

Give the following results for a dictionary with 1000 entries:

- Check `in dict.keys()`: 0.04537 (+62%)
- Check `in dict` 0.02794

(Changing the dict size does not change the results too much, it's always at least +50%)

I thought about introducing a lint rule for it, I found the following `SIM118`, but it's a bit too eager and will also change `for` loops, which I believe sometimes makes it a bit more readable and I haven't touched yet. 

https://docs.astral.sh/ruff/rules/in-dict-keys/

## Changes
- remove occurrences of key-existence checks against `dict.keys()` in favor of `in dict`.

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -293,7 +293,7 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         if integration_type in {IntegrationType.AWS, IntegrationType.AWS_PROXY}:
             to_validate = ILLEGAL_INTEGRATION_REQUESTS_AWS
 
-        for header in headers.keys():
+        for header in headers:
             if header.lower() in to_validate:
                 LOG.debug(
                     "Execution failed due to configuration error: %s header already present", header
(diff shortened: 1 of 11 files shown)


viren-nadkarni (approved):
LTGM :rocket:

alexrashed (approved):
Nice drill down and performance evaluation! The changes are looking good!
But I would really encourage you to try and enable the linting rule to be honest. Why would you say it has a negative impact on readability?

bentsku:
> But I would really encourage you to try and enable the linting rule to be honest. Why would you say it has a negative impact on readability?

@alexrashed I gave the lint rule a try locally and the results of the formatting and errors were not really good to be honest, it modified lots of instance in list comprehensions and `for` loops, even though the rule itself says: "Checks for key-existence checks against dict.keys() calls.". Also, it raised a lot of errors for "unsafe" checks, when the linter cannot be 100% sure the object is a dict, as for example `Headers` has the `keys` method as well, but it might not behave like a `dict`, which we might need to ignore. 

Plus, having a clear loop of `for key in dict.keys()` helps a bit for readability and might make sense sometimes to be very specific, and this lint rule would prevent this somehow.

PR #11699, opened 2024-10-16
cfn: fix failing `test_mapping_ref_map_key` for multi-account/region

sannya-singal:
## Motivation
With the support to add the region ref in mappings inside the conditions for cloudformation in the PR:#11671 the test `tests.aws.services.cloudformation.engine.test_mappings.TestCloudFormationMappings.test_mapping_ref_map_key` was updated to use `AWS::Region`, but when the test runs against non-default region it would fail because of the missing mapping in the template with the following error: 

`localstack.services.cloudformation.engine.errors.TemplateError: Invalid reference: 'ap-southeast-2' could not be found in the 'MyMap' mapping: '['*********']'`

Related workflow: https://app.circleci.com/pipelines/github/localstack/localstack/28846/workflows/dc1a47d0-eaf1-46a7-8149-fc987e80351b/jobs/253009/parallel-runs/2

## Changes
This PR adds mappings for the non-default regions currently being used to run the MA/MR pipeline. 
(Alternatively added a static region in https://github.com/localstack/localstack/pull/11699/commits/680cc01ca948e18c9ba467c122e8cfa84aa05e72 but this would defeat the purpose of testing ref region in the mappings)

## Testing
Set the environment variable: `TEST_AWS_REGION=us-west-1`

### DIFF
--- a/tests/aws/services/cloudformation/engine/test_mappings.py
+++ b/tests/aws/services/cloudformation/engine/test_mappings.py
@@ -2,7 +2,6 @@ import os
 
 import pytest
 
-from localstack.constants import AWS_REGION_US_EAST_1
 from localstack.testing.pytest import markers
 from localstack.utils.files import load_file
 from localstack.utils.strings import short_uid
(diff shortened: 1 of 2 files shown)


sannya-singal:
Thanks for the review @simonrw 🙌 Yes, the regions list is compared to the ones we test with and no I didn't consider to  generate the region map with templating.

PR #11700, opened 2024-10-16
fixed s3 deletion from stack in cdk

macnev2013:
## Motivation
This PR aims to fix the S3 bucket cleaning when a stack contains more than one buckets (with objects in it). According to my theory due to late binding in the lambda function, the `cleanup_s3_bucket` is not able to get the reference to the actual variable.

## Changes
- updated the cleanup of s3 buckets in the cdc provisioner function.

### DIFF
--- a/localstack-core/localstack/testing/scenario/provisioning.py
+++ b/localstack-core/localstack/testing/scenario/provisioning.py
@@ -243,9 +243,8 @@ class InfraProvisioner:
                 ]
 
                 for s3_bucket in s3_buckets:
-                    bucket_to_delete = s3_bucket  # Capture current value
                     self.custom_cleanup_steps.append(
-                        lambda bucket=bucket_to_delete: cleanup_s3_bucket(
+                        lambda bucket=s3_bucket: cleanup_s3_bucket(
                             self.aws_client.s3, bucket, delete_bucket=False
                         )
                     )
--- a/localstack-core/localstack/testing/scenario/provisioning.py
+++ b/localstack-core/localstack/testing/scenario/provisioning.py
@@ -243,9 +243,10 @@ class InfraProvisioner:
                 ]
 
                 for s3_bucket in s3_buckets:
+                    bucket_to_delete = s3_bucket  # Capture current value
                     self.custom_cleanup_steps.append(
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/testing/scenario/provisioning.py
```diff
@@ -243,9 +243,10 @@ def provision(self, skip_deployment: Optional[bool] = False):
                 ]
 
                 for s3_bucket in s3_buckets:
+                    bucket_to_delete = s3_bucket  # Capture current value
```
dominikschubert, line 246: FYI: this line doesn't actually  do anything in regards to the root cause, it would still lead to the same issue

localstack-core/localstack/testing/scenario/provisioning.py
```diff
@@ -243,9 +243,10 @@ def provision(self, skip_deployment: Optional[bool] = False):
...
                     self.custom_cleanup_steps.append(
-                        lambda: cleanup_s3_bucket(
-                            self.aws_client.s3, s3_bucket, delete_bucket=False
+                        lambda bucket=bucket_to_delete: cleanup_s3_bucket(
+                            self.aws_client.s3, bucket, delete_bucket=False
```
dominikschubert, line 249:
```suggestion
                        lambda s3_bucket=s3_bucket: cleanup_s3_bucket(
                            self.aws_client.s3, s3_bucket, delete_bucket=False
```

this alone would be enough to bind the value here

dominikschubert (approved):
Great catch! And it is indeed because of the late binding here where the actual value of `s3_bucket` was only resolved at the time of the lambda function execution, rather than the definition.

localstack-core/localstack/testing/scenario/provisioning.py
macnev2013: Thanks for the suggestion. I've removed this line.

localstack-core/localstack/testing/scenario/provisioning.py
macnev2013: I've update the `lambda` to use the loop variable.

Issue #11696, opened 2024-10-16
bug: API Gateway does not persist correctly when you restart the localstack docker container


bdowson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a working api gateway created with localstack. When I restart the container and try to query the same url, I get this message:
`{"message": "The API id '0e0cf92f' does not correspond to a deployed API Gateway API"}`.

# Details:
First I create my API and confirm it works:
```
$ awslocal apigatewayv2 get-apis
{
    "Items": [
        {
            "ApiEndpoint": "http://0e0cf92f.execute-api.localhost.localstack.cloud:4566",
            "ApiId": "0e0cf92f",
            "ApiKeySelectionExpression": "$request.header.x-api-key",
            "CorsConfiguration": {
                "AllowHeaders": [
                    "*"
                ],
                "AllowMethods": [
                    "*"
                ],
                "AllowOrigins": [
                    "*"
                ],
                "ExposeHeaders": [
                    "*"
                ]
            },
            "CreatedDate": "2024-10-16T05:24:49.452000+00:00",
            "DisableExecuteApiEndpoint": false,
            "Name": "XpedigoAPI_v2",
            "ProtocolType": "HTTP",
            "RouteSelectionExpression": "$request.method $request.path",
            "Tags": {},
            "Version": "2024-09-25 01:18:37UTC"
        }
    ]
}
```
```
$ awslocal apigatewayv2 get-stages --api-id=0e0cf92f
{
    "Items": [
        {
            "CreatedDate": "2024-10-16T05:24:49.524619+00:00",
            "DefaultRouteSettings": {
                "DetailedMetricsEnabled": false
            },
            "DeploymentId": "4d3d207f",
            "LastUpdatedDate": "2024-10-16T05:24:49.524619+00:00",
            "RouteSettings": {},
            "StageName": "local",
            "StageVariables": {
                "baseurl": "alb-localstack-bdowson.ngrok.io",
                "env": "local"
            },
            "Tags": {}
        }
    ]
}
```
```
$ awslocal apigatewayv2 get-deployments --api-id=0e0cf92f
{
    "Items": [
        {
            "AutoDeployed": false,
            "CreatedDate": "2024-10-16T05:24:49.529068+00:00",
            "DeploymentId": "4d3d207f",
            "DeploymentStatus": "DEPLOYED"
        }
    ]
}
```
Confirm it works:
```
[...]

PR #11702, opened 2024-10-16
APIGW NG: fix storing store attribute instead of full store

bentsku:
## Motivation
As reported in #11696, we've had an issue with APIGW NG persistence. We need to store the actual full store variable (container) instead of some of its dict attributes, as those could be overwritten (replacing a nested dict value with another dict, thus losing the "sync" and reference). 

## Changes
- store the full account bundle instead of the store, and create a property to access the store itself


cloutierMat (approved):
Nice catch! Neat fix :rocket:

Issue #11387, opened 2024-08-20
bug: AWS_REGION not respected in startup script


DsMaccy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Description
AWS_REGION environment variable is overwritten and unusable in the docker-compose.  This is noticeable when trying to run the startup script.  When echo-ing the AWS_REGION, it gets overwritten to us-east-1.  Once the region is hard-coded to the appropriate region, it works fine

## localstack_setup.sh

```
#!/usr/bin/env bash

set -euo pipefail

# enable debug
# set -x

echo "configuring sqs"
echo "==================="
LOCALSTACK_HOST=localhost
echo $AWS_REGION  # NOTE: The environment variable is getting overridden as us-east-1.  Ignore this environment variable here
AWS_REGION=us-east-2

create_queue() {
    local QUEUE_NAME_TO_CREATE=$1
    awslocal --endpoint-url=http://${LOCALSTACK_HOST}:4566 sqs create-queue --queue-name ${QUEUE_NAME_TO_CREATE} --region $AWS_REGION --attributes VisibilityTimeout=30
}

create_queue "test"

```

## docker-compose.yaml
```
    image: localstack/localstack
    ports:
      - "4566:4566"
    environment:
      AWS_REGION: us-east-2 # This is ignored at point of the script, but I expect that this is a temporary bug
      SERVICES: sqs
    volumes:
      # Container startup script
      - ./local_setup_config/localstack_setup.sh:/etc/localstack/init/ready.d/script.sh
```

## Resources
1. Similar issue here which is marked as resolved: https://github.com/localstack/localstack/issues/2951
2. Startup script found here: https://stackoverflow.com/questions/68131349/automatically-create-sqs-queue-using-localstack-and-docker-compose

### Expected Behavior

The AWS_REGION should be respected in the docker compose.  If us-east-1 is needed for particular AWS resources like those that are region agnostic, this should be handled with either a separate environment variable or individually for those particular services.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)
1. Add the startup script provided 
2. chmod +x <script>
3. docker-compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
See startup script provided (main issue is the environment variable, not any particular command)

### Environment

```markdown
- OS: Windows 11 using WSL running Ubuntu 20.04
- LocalStack:
[...]

PR #11705, opened 2024-10-17
Init hooks: Remove default AWS credentials

viren-nadkarni:
> [!IMPORTANT]
> This is a breaking change designated for LocalStack v4.0.0

## Background

The old implementation of multi-accounts in LocalStack used environment variables that determined the account/region of the sandbox. This impacted various other features, most notably the init hooks where we used to inherit the same environment variables when running the startup script. The startup scripts could be written without the need for explicit AWS credentials.

Since then, LocalStack moved to a multi-accounts implementation based on the request context. It no longer requires a global env var setting of account/region. But to maintain backward compatibility with existing init hook startup scripts, we started setting default account/region values in its environment.

## Changes

With LocalStack v4, the default AWS credentials will no longer be set in the init hook execution environment.

Startup scripts will now be required to set explicit credentials.

## Related

Closes:

- https://github.com/localstack/localstack/issues/11387

### DIFF
--- a/localstack-core/localstack/runtime/init.py
+++ b/localstack-core/localstack/runtime/init.py
@@ -155,6 +155,8 @@ class InitScriptManager:
             for script in scripts:
                 LOG.debug("Running %s script %s", script.stage, script.path)
 
+                env_original = os.environ.copy()
+
                 try:
                     script.state = State.RUNNING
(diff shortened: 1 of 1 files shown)


dfangl (approved):
LGTM, nice to get rid of this!

PR #11706, opened 2024-10-17
Add special handling of bedrock-runtime service in service router

silv-io:


alexrashed (changes requested):
We have an automated test which basically tests every single possible endpoint.
For bedrock, this is currently disabled because it did not work.
Would be great to remove this exclusion to verify that your fix actually works.
Afterwards we should be good to go! 🚀

thrau (approved):
:rocket: :robot: LGTM!

alexrashed (approved):
Awesome! Thanks for enabling bedrock in the service router tests! This way we can be sure that (a) _every_ routing in the service really works, and (b) there are no regressions due to newly introduced services causing new routing ambiguities. :)

PR #11708, opened 2024-10-17
X-Ray trace header parsing and Lambda tracing propagation

joe4dev:
## Motivation

We do not parse incoming X-Ray [trace headers](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader). Sample trace header (with Lambda lineage):

```
X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Sampled=1;Lineage=a87bd80c:1|68fd508a:5|c512fbe3:2
```

## Changes

* Add trace context parser to LocalStack handler chain with an implementation for X-Ray trace header parsing
* Add a `trace_context` field to the LocalStack `RequestContext`
* Propage `trace_context` through the Lambda service (including encoding and decoding hacks for async invokes)
* Add aws-validated test case `test_xray_trace_propagation`
* Add `TraceHeader` helper class implementation based on the open source AWS Python X-Ray SDK (Apache2.0 license-attributed): https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/trace_header.py
* Add `TraceId` helper class implementation based on the open source AWS Python X-Ray SDK (Apache2.0 license-attributed): https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/traceid.py

## Testing

* ext integration tests:
  *  [d68d520](https://github.com/localstack/localstack/pull/11708/commits/d68d520f5b879bb41fa27322a7c33f76d9c8533c) https://github.com/localstack/localstack-ext/actions/runs/11407926273
  * [5a9693f](https://github.com/localstack/localstack/pull/11708/commits/5a9693f323d9dbf42f26dfba22ddc14cc6e9e6bd) https://github.com/localstack/localstack-ext/actions/runs/11436665260
  * [006b34b](https://github.com/localstack/localstack/pull/11708/commits/006b34b422c5136f1dac6fe5d265bdbf1d519516) https://github.com/localstack/localstack-ext/actions/runs/11456732880
    * re-run after persistence test fix: https://github.com/localstack/localstack-ext/actions/runs/11462610581

## Usage

Use `context.trace_context` within any provider to access the parsed X-Ray trace header. Check out the `TraceHeader` class for more details.

Create new TraceHeader: 

```python
TraceHeader()  # all fields are None !

h1 = TraceHeader().ensure_root_exists()
h1.to_header_str() == "Root=1-67126ec5-45ded1c6463bac5ef4be1d69"

h2 = TraceHeader(root="1-3152b799-8954dae64eda91bc9a23a7e8", parent="7fa8c0f79203be72", sampled=1)
h2.to_header_str() == "Root=1-3152b799-8954dae64eda91bc9a23a7e8;Parent=7fa8c0f79203be72;Sampled=1"
```


localstack-core/localstack/aws/handlers/tracing.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    A handler that parses trace context headers, including:
+    * AWS X-Ray trace header: https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
+    """
+
+    def __call__(self, chain: HandlerChain, context: RequestContext, response: Response):
+        # TODO: handle case-insensitive headers because HTTP headers are case-insensitive:
+        #  https://stackoverflow.com/a/5259004
+        # TODO: write test case for case-insensitive xray header (verified manually)
+        trace_header_str = context.request.headers.get("X-Amzn-Trace-Id")
```
joe4dev, line 17: Do we have some efficient CaseInsensitiveDict utility we could use here to retrieve headers?

localstack-core/localstack/services/lambda_/invocation/event_manager.py
```diff
@@ -57,6 +58,11 @@ class SQSInvocation:
     exception_retries: int = 0
 
     def encode(self) -> str:
+        # Encode TraceHeader
+        aws_trace_header_str = TraceHeader(
```
joe4dev, line 62: @dfangl not super happy about this encoding/decoding hack. Feel free to share your suggestion.

This doesn't go well if `aws_trace_header` is not of type `TraceHeader`

localstack-core/localstack/utils/xray/trace_header.py
```diff
@@ -0,0 +1,139 @@
+# Source: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/trace_header.py
```
joe4dev, line 1: @thrau Do we have some best practices around re-using open source code, especially from AWS sources?

joe4dev (commented):
Happy to hear your feedback about X-Ray trace propagation (presented in today's demo) :)

localstack-core/localstack/aws/api/core.py
```diff
@@ -106,6 +108,7 @@ def __init__(self, request=None) -> None:
         self.service_request = None
         self.service_response = None
         self.service_exception = None
+        self.trace_context = {}
```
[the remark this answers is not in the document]
thrau: this is fine :+1:

localstack-core/localstack/aws/handlers/tracing.py
thrau: `context.request.headers` (it's the werkzeug Header data structure) should already be case insensitive. so you could also call `.get("X-AMAZN-TrAcE-id")`

localstack-core/localstack/utils/xray/trace_header.py
thrau: we don't, but i think what you're doing here is fine! maybe we could be a bit more verbose:

```
# This file is part of LocalStack.
# It is adapted from aws-xray-sdk-python licensed under the Apache License 2.0.
# You may obtain a copy of the Apache License 2.0 at http://www.apache.org/licenses/LICENSE-2.0
# Original source: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/trace_header.py
# Modifications: Add optional lineage field for https://docs.aws.amazon.com/lambda/latest/dg/invocation-recursion.html
```

localstack-core/localstack/aws/handlers/tracing.py
joe4dev: thanks, it works 🥳

localstack-core/localstack/utils/xray/trace_header.py
joe4dev: Done 👍

localstack-core/localstack/aws/app.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, service_manager: ServiceManager = None) -> None:
                 handlers.add_region_from_header,
                 handlers.rewrite_region,
                 handlers.add_account_id,
+                handlers.parse_trace_context,
```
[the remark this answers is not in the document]
bentsku: I think this is good! At first, I thought it'd be nicer to have it before the edge routes, because we could also make use of this in API Gateway `execute-api` routes for example. However, we do not pass the `RequestContext` to the edge routes at the moment. (@thrau laments this design decision still today 😄) so there is no point in making it happen early for now. We can always move it later on if anything changes! But for now, this looks to me 👌

localstack-core/localstack/utils/xray/trace_header.py
```diff
@@ -0,0 +1,192 @@
...
+
+HEADER_DELIMITER = ";"
+
+
+def generate_random_id():
+    """
+    Generate a random 16-digit hex str.
+    This is used for generating segment/subsegment id.
+    """
+    return binascii.b2a_hex(os.urandom(8)).decode("utf-8")
```
bentsku, line 32: I know this is taken from AWS code, but I think we could use some Python utils to make it a bit nicer.

`from secrets import token_hex`

```suggestion
    return token_hex(8)
```

localstack-core/localstack/utils/xray/traceid.py
```diff
@@ -0,0 +1,38 @@
...
+
+    VERSION = "1"
+    DELIMITER = "-"
+
+    def __init__(self):
+        """
+        Generate a random trace id.
+        """
+        self.start_time = int(time.time())
+        self.__number = binascii.b2a_hex(os.urandom(12)).decode("utf-8")
```
bentsku, line 26: nit: not sure about the double `__` here, should we just use a single underscore?
And we can also use `token_hex` instead of `binascii` and `os.random`

localstack-core/localstack/utils/xray/traceid.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Convert TraceId object to a string.
+        """
+        return "%s%s%s%s%s" % (
+            TraceId.VERSION,
+            TraceId.DELIMITER,
+            format(self.start_time, "x"),
+            TraceId.DELIMITER,
+            self.__number,
+        )
```
bentsku, line 38: We should probably use f-strings here, so we could even avoid the call to `format`:
```suggestion
        return f"{TraceId.VERSION}{TraceId.DELIMITER}{self.start_time:x}{TraceId.DELIMITER}{self._number}"
```

localstack-core/localstack/aws/handlers/tracing.py
bentsku, line 1: This looks great 👌

bentsku (approved):
Awesome, this looks great! 🚀 I've only reviewed the parts that are not related to lambda, as I'm lacking a bit of context and knowledge there. 

The handler looks great, I only have a few nits in the code taken from the AWS Python X-Ray SDK, as I think we could simplify it a bit by using new available Python standard lib utils and f-strings. Great work! 🚀

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -343,11 +341,22 @@ def get_prefixed_logs(self) -> str:
 
     def invoke(self, invocation: Invocation) -> InvocationResult:
         assert self.status == RuntimeStatus.RUNNING
+        # Async/event invokes might miss an aws_trace_header, then we need to create a new root trace id.
+        aws_trace_header = (
+            invocation.trace_context.get("aws_trace_header") or TraceHeader().ensure_root_exists()
+        )
+        # The Lambda RIE requires a full tracing header including Root, Parent, and Samples. Otherwise, tracing fails
+        # with the warning "Subsegment ## handler discarded due to Lambda worker still initializing"
+        aws_trace_header.ensure_sampled_exists()
```
dominikschubert, line 350: so if I understand that right, for lambda it's always `Sampled=1` while other services will by default get a `Sampled=0`? :thinking:

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -343,11 +341,22 @@ def get_prefixed_logs(self) -> str:
...
+        # TODO: replace this random parent id with actual parent segment created within the Lambda provider using X-Ray
+        aws_trace_header.ensure_parent_exists()
+        # TODO: test and implement Active and PassThrough tracing and sampling decisions.
+        # TODO: implement Lambda lineage: https://docs.aws.amazon.com/lambda/latest/dg/invocation-recursion.html
```
dominikschubert, line 354: Thanks for leaving the TODOs here. Those seem like really interesting tasks for engineers getting started at LocalStack :wink:

localstack-core/localstack/aws/app.py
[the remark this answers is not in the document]
joe4dev: I see. Feel free to change later :)

localstack-core/localstack/services/lambda_/invocation/lambda_service.py
```diff
@@ -235,9 +236,12 @@ def invoke(
         :param account_id: Account id of the function
         :param invocation_type: Invocation Type
         :param client_context: Client Context, if applicable
+        :param trace_context: tracing information such as X-Ray header
         :param payload: Invocation payload
         :return: The invocation result
         """
+        # NOTE: consider making the trace_context mandatory once we update all usages (should be easier after v4.0)
```
dominikschubert, line 243: agree, wanted to point out the same here :raised_hands:

localstack-core/localstack/utils/xray/trace_header.py
joe4dev: nice suggestions ✨ This looks much cleaner 🤩 

I fully agree on a technical level. For licensing and maintenance reasons, I opted to minimize changes to the original AWS code here (it is easier to match the authoritative AWS implementation regarding header parsing). Do you think it's worth deviating here?

tests/aws/services/lambda_/test_lambda_integration_xray.py
```diff
@@ -35,3 +36,40 @@ def test_traceid_outside_handler(create_lambda_function, lambda_su_role, tracing
...
+
+    create_lambda_function(
+        func_name=fn_name,
+        handler_file=TEST_LAMBDA_XRAY_TRACEID,
+        runtime=Runtime.python3_12,
+    )
+
+    # add boto hook
+    root_trace_id = "1-3152b799-8954dae64eda91bc9a23a7e8"
+    xray_trace_header = TraceHeader(root=root_trace_id, parent="7fa8c0f79203be72", sampled=1)
```
dominikschubert, line 59: Does this present any issues when testing repeatedly against AWS with the same trace id / parent?

dominikschubert (approved):
LGTM, most points have already been discussed and the current state seems very well documented and thought through. There's a few things we still need to tackle in the future like the sampling strategies but I actually quite like that we leave that out for now and can delegate this in the future.

Great work on re-using existing components. It makes a lot of sense to stay close to the official tooling here to make sure our behavior isn't diverging from AWS.


In terms of adjusting the internalized code from the xray  lib, I'd generally prefer to stay closer to the original code if possible but as long as it doesn't impede porting new changes too much, it's of course also ok to modernize it a bit.

tests/aws/services/lambda_/test_lambda_integration_xray.py
joe4dev: nope, works fine.
AWS does not perform any magic validation trying to re-construct the implicit start timestamp encoded here :)

PS: I fixed the repeated AWS validation issues in other X-Ray tests in this ext PR (awaiting review, low prio): https://github.com/localstack/localstack-ext/pull/3596

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
joe4dev: no, the sampled field is missing altogether (i.e., not present rather than sampled=0) for other services if no incoming tracing header is present with the sampled field.

That's not ideal, but this depends on the (service-dependent) sampling strategies and other services. We're lacking parity tests here, so idk what's the best behavior. Just leaving it open for now ...

localstack-core/localstack/utils/xray/trace_header.py
bentsku: I suppose not, we could also keep the code as it is, if it makes it easier to port new changes as they come (not sure how often this happens?).

joe4dev: I'm also crying every time I see `%s%s%s%s%s` 😭 . Nevertheless, I tend to prefer keeping the original for full parity (it's easier to relate to the AWS repo). Yeah, I don't expect direct changes here, maybe indirect ones for other X-Ray parts from the same source repo.

PR #11709, opened 2024-10-17
Add workflow to dispatch the update of the OpenAPI spec

giograno:
## Motivation
We are building a separate repo to store LocalStack's OpenAPI specs.
This workflow is triggered when the spec changes and it dispatches a workflow on the destination repo that generates a new `latest` spec.

See https://github.com/localstack/localstack-ext/pull/3554 for more context.

### DIFF
--- a/.github/workflows/update_openapi_spec.yml
+++ b/.github/workflows/update_openapi_spec.yml
@@ -1,0 +1,23 @@
+name: Update OpenAPI Spec
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '**/*openapi.yaml'
+      - '**/*openapi.yml'
+  workflow_dispatch:
+
+jobs:
+  update-openapi-spec:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Dispatch update spec workflow
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.PRO_ACCESS_TOKEN }}
+          repository: localstack/openapi
+          event-type: openapi-update
+          client-payload: '{"ref": "${{ github.ref }}", "repo": "${{ github.repository }}"}'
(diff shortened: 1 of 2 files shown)


.github/workflows/update_openapi_spec.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Dispatch update spec workflow
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.PRO_ACCESS_TOKEN }}
+          repository: localstack/openapi
+          event-type: openapi-update
+          client-payload: '{"ref": "${{ github.ref }}", "repo": "${{ github.repository }}"}'
```
alexrashed, line 23: Q: Is this really about handing over the reg and repo here? This would mean that if you trigger this manually from another branch, the specs would be taken from another branch, updating the specs in the repo with a non-merged version.
Is this what we want? What would this be used for?

alexrashed (commented):
The workflow is looking good, but I would discuss the usage patterns for the client-payload config before merging this PR. I think this _could_ be problematic, but could maybe also be useful?

.github/workflows/update_openapi_spec.yml
giograno: Fair observation. I added the ref to the branch and the `workflow_dispatch` to be able to test the workflow from some spec changes sitting in a branch. 
In my view, this workflow should be dispatched only when a spec change is pushed on master. Mostly for testing purposes, it could be handy to have the chance to trigger it from another branch, but I can't come up with concrete examples right now. The spec won't be automatically updated anyway: a PR would still have to be approved and merged.

alexrashed: Ah, right that makes sense! The additional barrier with the PR is great and should definitely be enough!
Would you mind adding a comment to the `client-payload`? I feel like this isn't obvious and might make sense to add a comment here?

alexrashed (approved):
Thanks for answering the question! I think it would be great to add a small comment on the line discussed, but that's just a nitpick / not blocking a merge! 🚀

PR #11710, opened 2024-10-18
Bump moto-ext to 5.0.17.post2

viren-nadkarni:
## Summary

This PR bumps moto-ext to 5.0.17.post2

Contains:

- https://github.com/getmoto/moto/pull/8231 @cloutierMat 
- https://github.com/getmoto/moto/pull/8157 @dfangl 
- https://github.com/getmoto/moto/pull/8230 @cloutierMat 
- https://github.com/getmoto/moto/pull/8233 @viren-nadkarni 

## To do

- [x] Ext integration tests **AWS / Ext Integration Tests # 1550**
- [x] Randomised account ID and region run https://app.circleci.com/pipelines/github/localstack/localstack/28904/workflows/839481df-f2aa-4355-8978-145cf58dca61

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ runtime = [
     "json5>=0.9.11",
     "jsonpath-ng>=1.6.1",
     "jsonpath-rw>=1.4.0",
-    "moto-ext[all]==5.0.17.post1",
+    "moto-ext[all]==5.0.17.post2",
     "opensearch-py>=2.4.1",
     "pymongo>=4.2.0",
     "pyopenssl>=23.0.0",
--- a/requirements-basic.txt
+++ b/requirements-basic.txt
@@ -34,7 +34,7 @@ packaging==24.1
     # via build
 plux==1.11.1
     # via localstack-core (pyproject.toml)
-psutil==6.0.0
+psutil==6.1.0
     # via localstack-core (pyproject.toml)
 pycparser==2.22
     # via cffi
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -256,7 +256,7 @@ mdurl==0.1.2
     # via markdown-it-py
 more-itertools==10.5.0
     # via openapi-core
-moto-ext==5.0.17.post1
+moto-ext==5.0.17.post2
     # via localstack-core
 mpmath==1.3.0
     # via sympy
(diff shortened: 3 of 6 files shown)


viren-nadkarni:
@alexrashed I guess now the only way to test this bump in Ext is to merge this first, am I correct?

alexrashed:
@viren-nadkarni No, we should minimize the friction in our downstream projects as good as possible, so we should make sure they are compatible before merging this PR. This is why we should not merge this PR just now.

We can easily test compatibility with our companion branch detection. So you can just create a companion PR in our downstream repo which also bumps `moto-ext`. We should then check in the logs that we really are installing the right version. If this PR, testing against this branch here in community is green, we can be sure that we already have forward-compatibility in our downstream project.

alexrashed (approved):
The changes here are looking great, it seems the upgrade does not introduce any integration issues here in Community. 🥳 

As soon as we can guarantee the compatibility in our downstream projects, we can move forward with merging this PR.

So to answer your question @viren-nadkarni: No, I think we should minimize the friction in our downstream projects as good as possible, so we should make sure they are compatible before merging this PR. This is why we should not merge this PR just now.

We can easily test compatibility with our companion branch detection. So you can just create a companion PR in our downstream repo which also bumps `moto-ext`. We should then check in the logs that we really are installing the right version. If this PR, testing against this branch here in community is green, we can be sure that we already have forward-compatibility in our downstream project.

viren-nadkarni:
@alexrashed I'm still not quite sure about this.

As I understand, `pip-compile` only checks the package versions from PyPI. I can run `make upgrade-pinned-dependencies` in this repo which will pick up the published `moto-ext` just fine. However, running it in ext will have no effect, because there, `moto-ext` is an indirect dependency via `localstack-core` and the latter is not yet published.

Also (correct me if I'm wrong) the 'Upgrade pinned Python dependencies' action in Ext does not  have companion branch detection. But it wouldn't be useful anyway due to `pip-compile`'s dependence on published packages.

I think one solution is to add maybe add `moto-ext` as a temporary dependency to `localstack-ext`, just verify the CI, the rely on 'Upgrade pinned Python dependencies' after community is bumped.

alexrashed:
Right, that's true, that makes it a bit more complicated. However, you can just do that on your own instead of using the make target, right? So you could just modify the requirements files in a branch (with the same name) in the downstream repo, right?

viren-nadkarni:
Ext suite verified passing (with the exception of Cloud Pods and Persistence tests which are known to be timing out and failing)

See **AWS / Ext Integration Tests # 1550** 

Merging PR

PR #11713, opened 2024-10-18
Add note recommending pyenv in dev environment docs

silv-io:
## Motivation

Contributors might get confused by our current documentation stating Python 3.11+ is needed as Python 3.12 does not have full support at this time. 
For dev environments we already have a checked in `.python-version` file in the top level, which would get picked up by any python version manager (e.g., pyenv).
This can already serve as the single source of truth for this.

## Changes

- Remove direct reference to a Python version and instead point to use pyenv
- If contributors don't want to use pyenv they can also now just quickly have a look at the `.python-version` and handle the version themselves.


thrau (approved):
LGTM! good advice :-)

PR #11719, opened 2024-10-21
Utils: Allow proxy override in download()

bblommers:
## Motivation
Useful if we're running an emulator as a proxy - when downloading a file, we don't want to intercept our own request.

Passing a dict like `{"HTTPS_PROXY": ""}` should explicitly disable any proxy configuration.

### DIFF
--- a/localstack-core/localstack/utils/http.py
+++ b/localstack-core/localstack/utils/http.py
@@ -181,6 +181,7 @@ def download(
     timeout: float = None,
     request_headers: Optional[dict] = None,
     quiet: bool = False,
+    proxies: Optional[Dict[str, str]] = None,
 ) -> None:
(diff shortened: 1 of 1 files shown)


alexrashed (commented):
I'm not 100% sure what this exactly is for.
Do you think you could go into detail a bit on why this is necessary? Or maybe even add a test showcasing your issue?
The unit tests for this module can be found here: https://github.com/localstack/localstack/blob/master/tests/unit/utils/test_http_utils.py

bblommers:
@alexrashed This is to get around the problem where the Github URL returns a 404, what we talked about this morning :slightly_smiling_face: 

I haven't figured  out the root cause, but I presume it's because we set a `HTTPS_PROXY` somewhere in the emulator. And that makes sense when we want to call other services, but in this case we want to explicitly ignore any proxies, and access the download-url directly.

PR #11721, opened 2024-10-21
improve tests selection

cloutierMat:
## Motivation

This PR aims to expand on the functionality of the `Matcher` for tests selection. The changes included will allow us to use the `directory` method to add an argument for which directory should be executed.

## Changes

We can now enter the following rule to explicitly target test folders when files in `my_feature` modules are changed.

```
Matchers.glob("localstack/my_feature/**").directory(
    paths=["tests/integration/my_feature"]
)
```

### DIFF
--- a/localstack-core/localstack/testing/testselection/matching.py
+++ b/localstack-core/localstack/testing/testselection/matching.py
@@ -94,6 +94,10 @@ class Matcher:
         return lambda t: [t] if self.matching_func(t) else []
 
     def directory(self, paths: list[str] = None):
+        """Enables executing tests on a full directory if the file is matched.
+        By default, it will return the directory of the modified file.
+        If the argument `paths` is provided, it will instead return the provided list.
+        """
         return lambda t: (paths or [get_directory(t)]) if self.matching_func(t) else []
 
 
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/testing/testselection/matching.py
```diff
@@ -93,8 +93,8 @@ def service_tests(self, services: list[str]):
     def passthrough(self):
         return lambda t: [t] if self.matching_func(t) else []
 
-    def directory(self):
-        return lambda t: [get_directory(t)] if self.matching_func(t) else []
+    def directory(self, paths: list[str] = None):
+        return lambda t: (paths or [get_directory(t)]) if self.matching_func(t) else []
```
alexrashed, line 97: nits:
- Would it make sense to explain the behavior of this function quickly in a docstring?
- I think it would be great if this new behavior would be tested in the unit tests of this module:
https://github.com/localstack/localstack/blob/master/tests/unit/testing/testselection/test_matching.py

alexrashed (approved):
Nice enhancement to the matcher functionality! It's great to see this tooling being enhanced to match new requirements! 💯 
I added two nits, they shouldn't be blocking a merge if you need to move fast though.

localstack-core/localstack/testing/testselection/matching.py
cloutierMat: Agreed, I will write some tests and documentation for it.

PR #11726, opened 2024-10-22
add AWS json protocol support for SQS developer endpoints

thrau:
## Motivation

There was a long-standing TODO for the sqs developer endpoints `/_aws/sqs/messages` to also support the modern x-amz-json-1.0 protocol introduced in https://github.com/localstack/localstack/pull/8268. This PR adds this functionality.

The implementation is not pretty but the best I could come up with without reworking the entire serialization mechanism. Basically through the use of the `@aws_response_serializer` decorator the protocol used to serialize the response was hard-coded, but needed to be dynamic per-request.

It's still a bit awkward because [we documented](https://docs.localstack.cloud/user-guide/aws/sqs/#json-response) that when calling this endpoint with `application/json` content type, it will return a JSON serialization of the query protocol response. So basically we have three protocols now: query (XML), x-amz-json-1.0 (for modern clients), and our weird/wrong json protocol that just returns the XML responses as json. That needs to be addressed at some point but is technically a breaking change.

## Changes

* When making AWS SQS client requests in x-amz-json-1.0 format to the endpoint `localhost:4566/_aws/sqs/messages` , the endpoint now correctly returns the JSON response.

## TODO

What's left to do:

- [ ] Update OpenAPI spec to contain the new input and output shape for x-amz-json-1.0


localstack-core/localstack/services/sqs/provider.py
```diff
@@ -652,20 +690,14 @@ def list_messages(self, request: Request) -> ReceiveMessageResult:
         return self._get_and_serialize_messages(request, region, account_id, queue_name)
 
     @route("/_aws/sqs/messages/<region>/<account_id>/<queue_name>")
```
bentsku, line 692: q: not sure if it's normal that this method accept all HTTP methods?

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -621,29 +651,37 @@ class SqsDeveloperEndpoints:
...
-            raise CommonServiceException(
-                "InvalidRequest", "This endpoint only accepts ReceiveMessage calls"
-            )
 
-        if not request.values.get("QueueUrl"):
+        if "x-amz-" in request.mimetype or "x-www-form-urlencoded" in request.mimetype:
+            # only parse the request using a parser if it comes from an AWS client
+            protocol = get_sqs_protocol(request)
+            operation, service_request = create_parser(
+                load_service("sqs", protocol=protocol)
```
bentsku, line 667: nit: I've updated it, as we're moving away from the `sqs-query` service name to the "service name" + "protocol" way of creating/selecting serializer and parser.

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -612,6 +614,34 @@ def check_fifo_id(fifo_id, parameter):
         )
 
 
+def get_sqs_protocol(request: Request) -> Literal["query", "json"]:
+    content_type = request.headers.get("Content-Type")
+    return "json" if content_type == "application/x-amz-json-1.0" else "query"
+
+
+def sqs_auto_protocol_aws_response_serializer(service_name: str, operation: str):
```
bentsku, line 622: this is a bit verbose and copies the logic of `aws_response_serializer` to get the `request` out of it, I'd like to make it cleaner but not sure how 🤔 there's no way of getting the proper protocol before we're inside the method itself.

bentsku (commented):
I'll only comment as I've updated the PR a bit, let me know what you think @thrau! I've tried to remove the logic of the protocol outside of the method to avoid having nested functions, we still need it for the parser though. Happy to get feedback!

localstack-core/localstack/services/sqs/provider.py
gregfurman: Perhaps we make this the same as the `/_aws/sqs/messages` route?
```suggestion
    @route("/_aws/sqs/messages/<region>/<account_id>/<queue_name>", methods=["GET", "POST"])
```

thrau:
thanks for your changes @bentsku! merging this PR to unblock @webdev51 for the SQS resource browser improvements. we still need to add the x-amz-json format to the openapi spec so we can re-enable this check here https://github.com/localstack/localstack/blob/b5823f8367e30a29a26c20f7c8f0aa3989d0c9f4/tests/aws/services/sqs/test_sqs_backdoor.py#L29

Issue #5249, opened 2022-01-10
bug: service-linked role creation not matching AWS role name


estebane-frb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I run the following command to create a service-linked role:

```aws iam create-service-linked-role --aws-service-name ecs.amazonaws.com --endpoint-url=http://localhost:4566```

Here is the role that is created.
Two things to notice: 
1) "RoleName" is "service-linked-role-xxxxxxxx". The ECS service-linked role name when I call AWS is **AWSServiceRoleForECS**
2) Because the name is using a uuid concatenated to "service-linked-role", I can keep calling this endpoint and it will always create a new role, whereas in AWS, I would received an error saying role already exists.
```
{
    "Role": {
        "Path": "/",
        "RoleName": "service-linked-role-6226467e",
        "RoleId": "4trul9gogjlqtlz5r1jk",
        "Arn": "arn:aws:iam::000000000000:role/aws-service-role/ecs.amazonaws.com/service-linked-role-6226467e",
        "CreateDate": "2022-01-10T21:32:00.610000+00:00",
        "AssumeRolePolicyDocument": {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Effect": "Allow",
                    "Principal": {
                        "Service": "ecs.amazonaws.com"
                    },
                    "Action": "sts:AssumeRole"
                }
            ]
        },
        "MaxSessionDuration": 3600
    }
}
```

### Expected Behavior

1) Service linked roles should match the names that AWS returns (for the case above it should return "AWSServiceRoleForECS")
2) Service linked roles that already exist should return an error 

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    aws iam create-service-linked-role --aws-service-name ecs.amazonaws.com --endpoint-url=http://localhost:4566

### Environment

```markdown
- OS: Windows 10 and MacOS
- LocalStack: 0.13.3
```

lukqw:
Hi @estebane-frb and thanks for contributing to LocalStack!

I was able to replicate your issue and have implemented a test for this. 
Not sure yet how / when this will be implemented, but I'll keep you updated.
Best wishes.

estebane-frb:
[...]

PR #11731, opened 2024-10-22
Fix: Ensure service-linked role names match AWS and prevent duplicate…

Alansfit:
Fix: Ensure service-linked role names match AWS and prevent duplicates (#5249)

## Motivation
This PR addresses issue #5249 by ensuring that service-linked role names match AWS's naming convention and preventing the creation of duplicate roles.

## Changes
- Updated the `create_service_linked_role` method to set the role name to `AWSServiceRoleFor{aws_service_name.split('.')[0].capitalize()}`.
- Added a check to prevent the creation of duplicate roles by raising a `CommonServiceException` if a role with the same name already exists.

### DIFF
--- a/localstack-core/localstack/services/iam/provider.py
+++ b/localstack-core/localstack/services/iam/provider.py
@@ -303,46 +303,55 @@ class IamProvider(IamApi):
         for tag in tag_keys:
             profile.tags.pop(tag, None)
 
-    def create_service_linked_role(
-        self,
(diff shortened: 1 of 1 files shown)


pinzon (changes requested):
This is really good. It Could you add an integration test to avoid regressions in the future?

- we should have a test to confirm the name generation structure
- assert that the correct exception is returned

PR #11734, opened 2024-10-23
Remove legacy StepFunctions v1 provider

joe4dev:
Depends on CI job removal https://github.com/localstack/localstack/pull/11778

## Motivation

The new StepFunctions provider v2 has been the default implementation [since LocalStack 3.0.0](https://discuss.localstack.cloud/t/new-stepfunctions-implementation-in-localstack-3-0/593) and we announced in the blog post that we are planning to remove the legacy v1 implementation with the next major release.

## Changes

- [x] Remove legacy provider initialization providers.py 
- [x] Remove code  `localstack-core/localstack/services/stepfunctions/legacy` including the package installers not required anymore 🧹 
- [x]  Remove test suite only for legacy provider `tests/aws/services/stepfunctions/legacy`
- [x]  Add deprecation warning for `PROVIDER_OVERRIDE_STEPFUNCTIONS`

## Testing

Mostly regression test suite.

Deprecations could be tested manually.

## Known Regressions

There are no known regressions and only few tests were executed for the legacy StepFunctions provider (only the [legacy tests directory](https://github.com/localstack/localstack/pull/11734/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L450)).

These tests had a conditional v2 skip but are likely just new snapshot skips rather than regressions (thanks @MEPalma for the clarification):
- [ ]  `aws.services.stepfunctions.v2.scenarios.test_sfn_scenarios.TestFundamental.test_step_functions_calling_api_gateway`
    - `paths=["$..content-type"],  # *FIXME: v2 includes extra content-type fields in Header fields.*`
- [ ]  `aws.services.stepfunctions.v2.scenarios.test_sfn_scenarios.TestFundamental.test_wait_for_callback`
    - `paths=["$..MessageId"]`

## TODO

What's left to do:

- [ ] Update StepFunctions docs (near v4 release)


dominikschubert (approved):
LGTM, nice attention to detail with the explicit v2 provider as well. We could even leave this in until 5.0 for a longer phase-out but less code is always nice of course.


It's great that we can now finally remove the fairly hacky v1 provider and  SO much complex (and now unused) code :broom: 

The artifacts that we're hosting in the artifacts repo need to stay a while longer unfortunately to keep supporting older versions.

dominikschubert (approved):

PR #11740, opened 2024-10-24
Fix Java installer on Mac hosts

viren-nadkarni:
## Changes

This PR fixes a bug in the Java installer where it downloaded and installed the Linux JRE even on Mac hosts.

With this change, any Java application using Java 11+ will run correctly when LS runs in host mode on Mac.

Eclipse Temurin [does not support](https://adoptium.net/en-GB/supported-platforms/) Java 8 aarch64 for Mac, so some components (notably Big Data) will continue to be non-functional.

Other changes:
- Omit Thai locale generation. This will further slightly reduce the disk footprint
- Update the pinned fallback JRE versions to the latest

### DIFF
--- a/localstack-core/localstack/packages/java.py
+++ b/localstack-core/localstack/packages/java.py
@@ -77,6 +77,8 @@ class JavaPackageInstaller(ArchiveDownloadAndExtractInstaller):
         return os.path.join(install_dir, "bin", "java")
 
     def _get_download_url(self) -> str:
+        # Note: Eclipse Temurin does not provide Mac aarch64 Java 8 builds.
+        # See https://adoptium.net/en-GB/supported-platforms/
         try:
             LOG.debug("Determining the latest Java build version")
             return self._download_url_latest_release()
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/packages/java.py
```diff
@@ -77,6 +77,8 @@ def _get_install_marker_path(self, install_dir: str) -> str:
         return os.path.join(install_dir, "bin", "java")
 
     def _get_download_url(self) -> str:
+        # Note: Eclipse Temurin does not provide Mac aarch64 Java 8 builds.
+        # See https://adoptium.net/en-GB/supported-platforms/
```
alexrashed, line 81: Q: This would basically mean that the download will fail because the URL does not exist? If so, could we make the error a bit more transparent?

alexrashed (approved):
Nice catch that we don't need the TH locale anymore (not that opensearch is using the JRE it ships), and great to see the installer support Mac again! 🐞 🔨 💯

PR #11741, opened 2024-10-24
Fix yaml parsing in cloudformation deploy ui

Jesse-Bakker:
## Motivation

Yaml parsing in the cloudformation deploy ui failed because it tried to naively
parse yaml instead of using the cloudformation yaml parser. This would error
on tags like `!Ref`.

## Changes

Use the same parser as is used when not going through the deploy ui

## Testing

Use the deploy ui with a yaml cloudformation template using tags like `!Ref` to 
verify that it works


Jesse-Bakker:
I can't be bothered to ask my employer for permission to sign the CLA for a single-line fix

PR #11743, opened 2024-10-24
remove S3 legacy CI job

bentsku:
## Motivation
As we are planning to remove the S3 legacy implementation, we can already remove the S3 CI job, as I'd like to setup a job for the DynamoDB(Streams) v2 provider, and the CI service-specific job landscape is getting crowded.

## Changes
- remove the S3 legacy CI job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -461,36 +461,6 @@ jobs:
       - store_test_results:
           path: target/reports/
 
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Great to see that we clean up here as we really don't need the regression testing of a provider which will not even be contained in the next release anymore. 🧹 💯

PR #11746, opened 2024-10-24
remove s3 legacy provider

bentsku:
## Motivation

S3 v3 (fully native in LocalStack) has been the default since v3, and we deprecated and did not maintain the legacy v2 implementation and did not provide new features to it since then. 

For the v4 release, we are removing the implementation and the associated tests. 

## Changes
- remove all providers definition related to S3, as we now only support one default implementation
- remove all code associated with legacy
- remove all code related to the legacy implementation: skip markers and skip snapshot markers


alexrashed (approved):
Nice and clean deletion of the old S3 provider. 🧹 💯

Issue #11755, opened 2024-10-28
bug: API Gateway could not resolve models


InvokedLambda:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I do a post request to a rest resource of an API Gateway with a request validator, the reference model could not be resolved. I think this worked in the past, but I could not specify any version because I haven't run my local development env for a long time. But I can say it works on the aws cloud.

Error message:
```
2024-10-28T16:35:57.992 ERROR --- [t.reactor-11] l.aws.handlers.logging     : exception during call chain
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/local/lib/python3.11/http/client.py", line 1303, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/local/lib/python3.11/http/client.py", line 1349, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.11/http/client.py", line 1298, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/local/lib/python3.11/http/client.py", line 1058, in _send_output
    self.send(msg)
  File "/usr/local/lib/python3.11/http/client.py", line 996, in send
    self.connect()
  File "/usr/local/lib/python3.11/http/client.py", line 1468, in connect
    super().connect()
  File "/usr/local/lib/python3.11/http/client.py", line 962, in connect
    self.sock = self._create_connection(
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/socket.py", line 863, in create_connection
    raise exceptions[0]
  File "/usr/local/lib/python3.11/socket.py", line 848, in create_connection
    sock.connect(sa)
OSError: [Errno 99] Cannot assign requested address
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/referencing/_core.py", line 423, in get_or_retrieve
    resource = registry._retrieve(uri)
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/jsonschema/validators.py", line 113, in _warn_for_remote_retrieve
    with urlopen(request) as response:  # noqa: S310
[...]

Issue #9631, opened 2023-11-14
OpenAPI 'oneOf' not supported


bwa12:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello, 

I am using an OpenAPI yaml file which is describing my entire REST API. 

In this yaml file I am using the `oneOf` keyword to show that the `BaseEvent` object can be one of three schemas, like so:

```yaml
BaseEvent:
      oneOf:
        - $ref: "#/components/schemas/UnsettledEvent"
        - $ref: "#/components/schemas/SettledEvent"
        - $ref: "#/components/schemas/Game"
```

But when I call this endpoint with a valid JSON body I keep getting this error:

```json
{
  "__type": "InternalError",
  "message": "exception while calling apigateway with unknown operation: Unresolvable: http://apigateway.localhost.localstack.cloud:[port]/restapis/[localID]/models/UnsettledEvent"
}
```

Is there a way around this issue?

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

### Environment

```markdown
- OS: MacOS 13.5
- LocalStack: 2.3.0
```

bentsku:
Hello and thanks for your report! Could you confirm this works against AWS? Would it be possible to provide a small reproducible sample so that we could try it? No need for the integrations, we can replace them with `MOCK` type, but at least a working and importable OpenAPI file would be great. Thanks!

bwa12:
Hiya, 

Thanks for getting back to me! So for the url I hide these values just to be on the safe side 😅 

The full url I am seeing is `http://apigateway.localhost.localstack.cloud:4566/restapis/orc7susrxa/models/UnsettledEvent` which is all correct (port number and localstack id are both correct). 

I can confirm that this works against AWS! It is only an issue when running locally with localstack. 

The full OpenAPI yaml file is here:


```yaml
openapi: 3.0.3
info:
  title: Game Events API - OpenAPI 3.0
  description: |-
    An API to consume events from SBG games, as well as handling the redemption of prizes. Based on the OpenAPI 3.0 specification.
  version: 1.0.0
servers:
  - url: http://test4-url-to-be-confirmed
    description: Test4
  - url: http://staging-url-to-be-confirmed
    description: Staging
tags:
  - name: events
    description: Allows the creation of events
  - name: redemptions
    description: Allows the creation of prize redemptions
paths:
[...]

PR #11756, opened 2024-10-28
fix APIGW oneOf and add APIGW test for Model $ref resolving

bentsku:
## Motivation
The same issue as  #9631 was reported in #11755. I've implemented 2 tests to validate the behavior, and the main issue is that we were not properly replacing the `$ref` in lists, leading to the `jsonschema` library directly trying to reach the link. We could probably also remove the custom model resolving, and do a regex sub on the `https://apigateway.amazonaws.com` string and replace by the LocalStack edge, but still not sure it would work, as the first reported issue still had problems resolving it.

## Changes
- fix the logic to also resolve `$ref` inside JSON arrays
- add 2 tests related to `$ref` resolving, one regular and one with `oneOf` property

_fixes #11755 and fixes #9631_

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -261,6 +261,10 @@ class ModelResolver:
 
                 elif isinstance(value, dict):
                     _look_for_ref(value)
+                elif isinstance(value, list):
+                    for val in value:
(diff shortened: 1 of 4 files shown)


cloutierMat (approved):
Good dig and neat fix!

Thanks for adding good comprehensive tests!

PR #11757, opened 2024-10-28
Convert SQS test util into re-usable fixture sqs_collect_messages

joe4dev:
## Motivation

Collecting SQS messages is a common task in LocalStack testing and we don't have a re-usable fixture (across LS and LS-ext) yet. The fixture is used in a LS-ext PR for testing Pipes persistence.

## Changes

* Convert SQS helper util into re-usable fixture `sqs_collect_messages`

## Discussion

When should we use fixtures and when util/helper methods (passing in an aws_client)?

### DIFF
--- a/localstack-core/localstack/testing/pytest/fixtures.py
+++ b/localstack-core/localstack/testing/pytest/fixtures.py
@@ -6,7 +6,7 @@ import os
 import re
 import textwrap
 import time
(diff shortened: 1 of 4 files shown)


thrau (approved):
nice one! :100: it makes sense to me to have this as a re-usable fixture. i was looking briefly into making this into a boto waiter but found it to be too cumbersome, so just went for the fixture instead.

PR #11759, opened 2024-10-29
Remove legacy Lambda config for enabling the LocalStack SQS dependency

joe4dev:
## Motivation

@dfangl implemented a lightweight `internal_sqs_queue.py` (localstack-core/localstack/services/lambda_/invocation/internal_sqs_queue.py) for Lambda event/async invokes to improve performance and remove the internal dependency to LocalStack SQS. This Lambda-internal SQS queue is the default implementation and switching back to the LocalStack SQS queue was available as an internal (i.e., not communicated to customers) config flag called `LAMBDA_EVENTS_INTERNAL_SQS`.

## Changes

* Remove and deprecate the LocalStack-internal config option `LAMBDA_EVENTS_INTERNAL_SQS` for enabling the LocalStack SQS dependency (i.e., disabling the Lambda-internal SQS implementation).


joe4dev:
Should we consider this as a minor change (and merge into master right away) because it shouldn't break anything on the customer side and was never communicated?

dfangl (approved):
Thanks for taking care of this!

dfangl:
And yes, I would consider this a minor change, no one ever asked for the old version.

PR #11761, opened 2024-10-29
Feature/SNS: Decompose publish to topic

maxhoheiser:
## Motivation
In EventStudio we want to trace outgoing SNS notifications, when the are pushed to subscribers via the `PublishDispatcher.publish_to_topic`.
To optimize performance the publishing to subscribers is executed in subthreads.
For tracing to work we need to make the trace_context available in the sub-thread where need to wrap the `notifier.publish` function with the current `thrading.local` context.
Therfore, we need to decompose the execution in the subthread into a separate function.


bentsku (approved):
LGTM! Clean change 👍

PR #11764, opened 2024-10-29
APIGW NG: add LocalStack-only CORS handling for REST API

bentsku:
## Motivation
We currently have special logic in `localstack/aws/handlers/cors.py`, special for S3 and API Gateway who can self manage their CORS behavior.

In order to guess that we are going to target an edge route, we currently have custom logic in `should_enforce_self_managed_service`, which uses some heuristic to guess it and let the CORS handler do its thing. 

However, this does not catch every API Gateway routes, especially the ones related to custom domains. It would be good to move most of the CORS related logic for those "self-managed" services inside the service itself, like we have done for S3: meaning that if the config flag to disable it is enabled, we should let the service apply the global LocalStack CORS logic in its own way (as they most probably already have logic in place for it). 

We have the `DISABLE_CUSTOM_CORS_APIGATEWAY` flag to disable the CORS handling of API Gateway, and try to apply the default LocalStack CORS logic to API Gateway routes, allowing users to disregard and not modify their lambda for example for the `AWS_PROXY` (managing CORS for REST API can be quite a pain, see https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html). This specific handler will be executed only if the flag is on, and is meant to properly apply the CORS headers in all situations for the API Gateway edge routes. 

It also also allows us to have custom logic for the `OPTIONS` request, for example. 

## Changes
- add a custom CORS handler for APIGW NextGen
- we currently already have tests for the CORS handling when `config.DISABLE_CUSTOM_CORS_APIGATEWAY` is set, and the pipeline is green
- we also will have custom tests in the upstream repo

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway.py
@@ -32,15 +32,16 @@ class RestApiGateway(Gateway):
                 handlers.method_response_handler,
             ]
         )
-        self.response_handlers.extend(
-            [
-                handlers.response_enricher
-                # add composite response handlers?
-            ]
-        )
         self.exception_handlers.extend(
             [
                 handlers.gateway_exception_handler,
+            ]
+        )
+        self.response_handlers.extend(
+            [
+                handlers.response_enricher,
+                handlers.cors_response_enricher,
+                # add composite response handlers?
             ]
         )
 
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/cors.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/cors.py
@@ -1,0 +1,49 @@
+import logging
+from http import HTTPMethod
+
+from localstack import config
+from localstack.aws.handlers.cors import CorsEnforcer
+from localstack.aws.handlers.cors import CorsResponseEnricher as GlobalCorsResponseEnricher
+from localstack.http import Response
+
+from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
+from ..context import RestApiInvocationContext
+from ..gateway_response import MissingAuthTokenError
+
+LOG = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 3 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/gateway.py
```diff
@@ -32,15 +32,16 @@ def __init__(self):
                 handlers.method_response_handler,
             ]
         )
-        self.response_handlers.extend(
+        self.exception_handlers.extend(
```
bentsku, line 35: this was just a switch around, because exception handlers are executed before response handlers, as we've learned the hard way in the beginning 😄

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/cors.py
```diff
@@ -0,0 +1,49 @@
...
+    ):
+        """
+        This is a LocalStack only handler, to allow users to override API Gateway CORS configuration and just use the
+        default LocalStack configuration instead, to ease the usage and reduce production code changes.
+        """
+        if not config.DISABLE_CUSTOM_CORS_APIGATEWAY:
+            return
+
+        if not context.invocation_request:
+            return
```
cloutierMat, line 31: Question: What is the scenario in which this handler gets called without an invocation request?

cloutierMat (approved):
Another great example of the power of the handler chain! Everyday I am learning a bit more about cors! :pray:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/cors.py
bentsku: This is a very good question 😅 I've taken it from the HTTP CORS handler, probably if we encounter an unexpected exception in the `InvocationRequestParser`, we could maybe raise some kind of exception, and the response handlers are always executed, so we would avoid a weird attribute error, or trying to access `["headers"]` of `None` maybe? I think it's pretty unlikely, but better safe than sorry with the response handlers 😬

PR #11767, opened 2024-10-30
migrate to LOCALSTACK_AUTH_TOKEN

alexrashed:
## Motivation
With the upcoming release of LocalStack v4, we will push the deprecation of `LOCALSTACK_API_KEY` in favor of `LOCALSTACK_AUTH_TOKEN`.
This PR prepares this deprecation by migrating all usages of the API key in favor of the auth token.
This change does _not_ introduce any deprecation message yet. 

## Changes
- Use `LOCALSTACK_AUTH_TOKEN` instead of `LOCALSTACK_API_KEY` wherever possible.
- Avoid directing users to set a `LOCALSTACK_API_KEY` in case no credentials are set.

/cc @Pive01 @SimonWallner

### DIFF
--- a/localstack-core/localstack/utils/bootstrap.py
+++ b/localstack-core/localstack/utils/bootstrap.py
@@ -455,7 +455,7 @@ def get_docker_image_to_start():
     image_name = os.environ.get("IMAGE_NAME")
     if not image_name:
         image_name = constants.DOCKER_IMAGE_NAME
-        if is_api_key_configured():
+        if is_auth_token_configured():
             image_name = constants.DOCKER_IMAGE_NAME_PRO
     return image_name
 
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/utils/bootstrap.py
```diff
@@ -1358,10 +1358,11 @@ def in_ci():
     return False
 
 
-def is_api_key_configured() -> bool:
+def is_auth_token_configured() -> bool:
```
thrau, line 1361: i see you already made the changes in ext, but feel that leaving `is_api_key_configured` just in case would be useful. by creating code paths like this we leave a little remnant in there but at least decouple the dependency. anyway, just a nit

```python
def is_api_key_configured():
    return is_auth_token_configured()
```

thrau (approved):
nice cleanup! LGTM

PR #11768, opened 2024-10-31
fix S3 enabling versioning with a ObjectLockEnabledForBucket bucket

bentsku:
## Motivation
Raised from a support issue with a Crossplane integration. 

We would raise `InvalidBucketState: An Object Lock configuration is present on this bucket, so the versioning state cannot be changed` when setting up a bucket with both versioning and ObjectLock enabled. I've adapted a test case to reproduce this case, where enabling versioning on a bucket that already has it enabled via`ObjectLockEnabledForBucket` set to `True`, does not trigger an exception because it's effectively a null operation.

## Changes
- adapt a test to reproduce the exception
- modify the logic so that we only block changing the status to `Suspended`

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -2680,13 +2680,13 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                 message="The Versioning element must be specified",
             )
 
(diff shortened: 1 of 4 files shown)


cloutierMat (approved):
Clean fix! Nice test update! :rocket:

PR #11770, opened 2024-10-31
Fix typo `str.rsplit`

sbrugman:
## Motivation
Minor bug: `maxsplit` in `str.split` with negative numbers defaults to `maxsplit=0`, which is the default.
The author probably intended `maxsplit=2` as only the last two elements are required.

### DIFF
--- a/localstack-core/localstack/services/kms/utils.py
+++ b/localstack-core/localstack/services/kms/utils.py
@@ -14,7 +14,7 @@ def get_hash_algorithm(signing_algorithm: str) -> str:
     Return the hashing algorithm for a given signing algorithm.
     eg. "RSASSA_PSS_SHA_512" -> "SHA_512"
     """
-    return "_".join(signing_algorithm.rsplit(sep="_", maxsplit=-2)[-2:])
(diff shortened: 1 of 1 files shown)


sbrugman:

(of course I did not really read it)

sannya-singal (commented):
Thanks for raising the PR @sbrugman 🙌 I would really appreciate if you could read the CLA Document, it should be a quick read 😁 After that, perhaps you could update the signature comment above.🙂

sbrugman:
recheck

sannya-singal (commented):
Thanks for fixing this 🙌 I would really appreciate if you could also add a small test case around it to verify the results 🙂

sbrugman:
Feel free to add it. I've enabled "Allow edits by maintainers"

alexrashed:
@sbrugman Since this change without a proper test doesn't really give us confidence that it is correct and necessary, I'll close this PR for now. Feel free to reopen it again in case there's a need for it.
Sorry for the inconvenience, feel free to reach out again if this causes any actual issues (your code is equivalent as far as I can tell), then we can maybe craft a test together.

PR #11771, opened 2024-11-01
Delete unused test Cfn templates

viren-nadkarni:
This PR removes several unused Cfn test templates.


bentsku (commented):
Thanks for updating it! I've tried to look for the usage of the `template28.yaml` but couldn't find any. Is this template still used?

viren-nadkarni:
> Thanks for updating it! I've tried to look for the usage of the template28.yaml but couldn't find any. Is this template still used?

@bentsku You're right, it seems like most of these `template**.yaml` files are no longer used. I'll delete them and see if anything breaks

viren-nadkarni:
@bentsku Thanks for highlighting this. This PR now removes all the unused Cfn templates.

bentsku (approved):
🧹! LGTM, I think this is a nice cleanup, hard to keep track of those kind of dependencies when they all live in their own folder, and also pretty hard to know where it is used. 

None of them were used in upstream dependencies, right? It does not look like it 👌 

Maybe Dominik wants to chime in too just in case, but I think we lost the context of those files and they are not worth a lot anymore.

dominikschubert (approved):
LGTM, tried to find references for a few of the removed files but also couldn't find any, so seems good! Thanks for cleaning that up :broom:

PR #11772, opened 2024-11-01
Replicator: support generating specific VPC ids

simonrw:
## Motivation

For the Replicator project, we need the ability to create resources with fixed IDs. VPCs have a random ID that is not an input parameter on creation. E.g. `vpc-abcdefg0123456789'. We need the ability to generate knwon ids using the ID store.

## Changes

* add ability to generate known IDs, where the CIDR block is used as a unique key along with the account id and region
    * handle this case if a custom ID is _not_ provided via tags
* Create a unit test for this behaviour


localstack-core/localstack/services/ec2/patches.py
```diff
@@ -92,12 +127,21 @@ def ec2_create_vpc(
...
+        # try to generate a custom id from the id store
+        if not custom_id:
+            resource_identifier = VpcIdentifier(self.account_id, self.region_name, cidr_block)
+            id = localstack_id_manager.get_custom_id(resource_identifier)
+            if id:
+                custom_id = id
+
         # Check if custom id is unique
         if custom_id and custom_id in self.vpcs:
             raise InvalidVpcDuplicateCustomIdError(custom_id)
```
cloutierMat, line 139: Using the generate method will do all the work for you! It also looks at the `TAG_KEY_CUSTOM_ID`, and duplicate ids. It will not raise though :thinking: and instead will generate a new id... so maybe you will want to leave that part out?

```suggestion
        resource_identifier = VpcIdentifier(self.account_id, self.region_name, cidr_block)
        custom_id = resource_identifier.generate(tags=tags, existing_ids=self.vpcs.keys())
```

localstack-core/localstack/services/ec2/patches.py
```diff
@@ -9,11 +11,43 @@
...
 
 
+@localstack_id
+def generate_vpc_id(
+    resource_identifier: ResourceIdentifier,
+    existing_ids: ExistingIds = None,
+    tags: Tags = None,
+) -> str:
+    suffix = "".join(random.choice(string.ascii_letters) for _ in range(17))
+    return f"vpc-{suffix}"
```
cloutierMat, line 33: Here we could return `""`.  And maybe add a comment that moto does the generation?

cloutierMat (commented):
It looks pretty good. Glad you were able to make it work! :rocket: 

I think you are missing out on the power of the `generate` method. It does all of the conditional stuff for you!

The fix would probably be better if done in moto, modifying `create_vpc` and `random_vpc_id` directly. That would allow us to remove a lot of dirty patch work! :rofl: But this can also be tackled in batches once we have progressed more on a service rather than a resource by resource pr. Should we keep a tally?

localstack-core/localstack/services/ec2/patches.py
pinzon: Yes I think @cloutierMat merged recently a PR so moto resource can have the custom id generation. I think this utility is for our LocalStack store resources.

cloutierMat (approved):
Thanks for addressing the comments. :rocket:

PR #11773, opened 2024-11-01
Add context manager to id generator

cloutierMat:
## Motivation

Adding a context manager to allow for a streamlined usage of the id manager where a resource unsets after use.

### DIFF
--- a/localstack-core/localstack/utils/id_generator.py
+++ b/localstack-core/localstack/utils/id_generator.py
@@ -19,9 +19,10 @@ class LocalstackIdManager(MotoIdManager):
 
     @contextmanager
     def custom_id(self, resource_identifier: ResourceIdentifier, custom_id: str) -> None:
-        yield self.set_custom_id(resource_identifier, custom_id)
-
-        self.unset_custom_id(resource_identifier)
+        try:
+            yield self.set_custom_id(resource_identifier, custom_id)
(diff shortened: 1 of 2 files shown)


pinzon (approved):
LGTM

localstack-core/localstack/utils/id_generator.py
```diff
@@ -16,6 +17,12 @@ def set_custom_id_by_unique_identifier(self, unique_identifier: str, custom_id:
         with self._lock:
             self._custom_ids[unique_identifier] = custom_id
 
+    @contextmanager
+    def custom_id(self, resource_identifier: ResourceIdentifier, custom_id: str) -> None:
+        yield self.set_custom_id(resource_identifier, custom_id)
+
+        self.unset_custom_id(resource_identifier)
```
simonrw, line 24: We probably want to wrap this in a try/finally block to make sure that we reset the state if an error is thrown
Can we also cover this in the tests?

simonrw (approved):
Just a comment about catching any exceptions in the generator but otherwise LGTM!

localstack-core/localstack/utils/id_generator.py
cloutierMat: Good thinking! I will add that

PR #11783, opened 2024-11-05
APIGW: fix importing API with Cognito Authorizer

bentsku:
## Motivation
We got a report that import an API with the following configuration would fail with a cryptic error:
```python
localstack-test  | 2024-11-04T16:17:31.927  WARN --- [et.reactor-0] l.s.a.n.e.h.gateway_except : Non Gateway Exception raised: 'NoneType' object has no attribute 'startswith'
```

When the OpenAPI spec had the following:
```yaml
components:
  securitySchemes:
    EndpointAuthorizer:
      type: apiKey
      name: Authorization
      in: header
      x-amazon-apigateway-authtype: cognito_user_pools
      x-amazon-apigateway-authorizer:
        type: cognito_user_pools
        providerARNs:
          - arn:aws:cognito-idp:${AWS::Region}:${AWS::AccountId}:userpool/${CognitoUserPool}
```

This was because we did not populate the `identitySource` of the Authorizer when the authorizer type was `cognito_user_pools`. 

As a note: we greatly improve the CRUD validation of the provider, but the `ImportApi` logic is not very well validated, and will probably lead to issue down the invocation road. We should dedicate some time to properly rework the import logic. 

## Changes
- fix the logic to properly get the `identitySource` if the auth type is `cognito_user_pools`
- add a test to verify that the CRUD layer works properly

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -537,7 +537,7 @@ def import_api_from_openapi_spec(
                 name=security_scheme_name,
                 type=authorizer_type,
                 authorizerResultTtlInSeconds=aws_apigateway_authorizer.get(
-                    "authorizerResultTtlInSeconds", 300
+                    "authorizerResultTtlInSeconds", None
                 ),
             )
             if provider_arns := aws_apigateway_authorizer.get("providerARNs"):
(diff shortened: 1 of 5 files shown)


cloutierMat (approved):
Thanks for the fix! :rocket: And simple test to boot!

PR #11787, opened 2024-11-05
Fix: CFn: support tags with ESMs

simonrw:
## Motivation

Event Source Mappings are supported via CloudFormation, however two things are currently broken:

1. When assigning tags, the following error is received:

```
Invalid type for parameter Tags, value: [{'Key': 'component', 'Value': 'value'}], type: <class 'list'>, valid types: <class 'dict'>
```

This means that the tag format supported by CloudFormation is not directly compatible with the tag format used by `boto3`.

2. When performing `!GetAtt <ESM>.EventSourceMappingArn` the value returned is `None`

## Changes

* Translate tags from one format to another
* Set the `EventSourceMappingArn` property to handle the `GetAtt` case

### DIFF
--- a/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
+++ b/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
@@ -136,7 +136,6 @@ class LambdaEventSourceMappingProvider(ResourceProvider[LambdaEventSourceMapping
 
         response = lambda_client.create_event_source_mapping(**params)
         model["Id"] = response["UUID"]
-        model["EventSourceArn"] = response["EventSourceArn"]
         model["EventSourceMappingArn"] = response["EventSourceMappingArn"]
 
         return ProgressEvent(
--- a/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
+++ b/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
@@ -136,7 +136,8 @@ class LambdaEventSourceMappingProvider(ResourceProvider[LambdaEventSourceMapping
 
         response = lambda_client.create_event_source_mapping(**params)
         model["Id"] = response["UUID"]
-        model["EventSourceArn"] = model["EventSourceMappingArn"] = response["EventSourceMappingArn"]
+        model["EventSourceArn"] = response["EventSourceArn"]
+        model["EventSourceMappingArn"] = response["EventSourceMappingArn"]
 
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
--- a/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
+++ b/localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
@@ -1,6 +1,7 @@
 # LocalStack Resource Provider Scaffolding v2
 from __future__ import annotations
 
+import copy
 from pathlib import Path
 from typing import Optional, TypedDict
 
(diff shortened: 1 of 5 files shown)


gregfurman (commented):
LGTM! Think just fix that `EventSourceArn` assignment and I'll approve 👍

localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
```diff
@@ -126,8 +127,16 @@ def create(
...
+        params = copy.deepcopy(model)
+        if tags := params.get("Tags"):
+            transformed_tags = {}
+            for tag_definition in tags:
+                transformed_tags[tag_definition["Key"]] = tag_definition["Value"]
+            params["Tags"] = transformed_tags
+
+        response = lambda_client.create_event_source_mapping(**params)
         model["Id"] = response["UUID"]
+        model["EventSourceArn"] = model["EventSourceMappingArn"] = response["EventSourceMappingArn"]
```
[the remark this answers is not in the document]
simonrw: Thanks for the clarification!

localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
```diff
@@ -126,8 +127,17 @@ def create(
...
+        params = copy.deepcopy(model)
+        if tags := params.get("Tags"):
+            transformed_tags = {}
+            for tag_definition in tags:
+                transformed_tags[tag_definition["Key"]] = tag_definition["Value"]
+            params["Tags"] = transformed_tags
+
+        response = lambda_client.create_event_source_mapping(**params)
         model["Id"] = response["UUID"]
+        model["EventSourceArn"] = response["EventSourceArn"]
```
gregfurman, line 139: Just want to clarify that this value needs to be set? I see the CFn docs show us just being able to get the `Id` and `EventSourceMappingArn` attributes.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#aws-resource-lambda-eventsourcemapping-return-values

gregfurman (approved):

localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
```diff
@@ -126,8 +127,16 @@ def create(
         model = request.desired_state
         lambda_client = request.aws_client_factory.lambda_
 
-        response = lambda_client.create_event_source_mapping(**model)
+        params = copy.deepcopy(model)
+        if tags := params.get("Tags"):
+            transformed_tags = {}
+            for tag_definition in tags:
+                transformed_tags[tag_definition["Key"]] = tag_definition["Value"]
+            params["Tags"] = transformed_tags
```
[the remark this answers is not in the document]
simonrw: 😢 I know what you mean but sometimes that's the way things are

localstack-core/localstack/services/lambda_/resource_providers/aws_lambda_eventsourcemapping.py
simonrw: You are right, thanks 🙏 . I have just tested against AWS (never trust the docs, _especially_ for CFn!). On the other hand, there is a discrepancy between the specs we use to generate the resource providers and the CFn implementation. The specs do not mention `EventSourceMappingArn` as a property that can be accessed on the resource, however it is clearly there in the docs.

Again: _never_ **_NEVER_** trust the docs! (or the specs in this case... 😂 )

tests/aws/services/lambda_/event_source_mapping/test_cfn_resource.py
```diff
@@ -0,0 +1,54 @@
...
+    aws_client.sqs.send_message(
+        QueueUrl=queue_url,
+        MessageBody=json.dumps({"body": "something"}),
+    )
+
+    retry(
+        lambda: aws_client.s3.head_object(Bucket=stack.outputs["OutputBucketName"], Key=output_key),
+        retries=10,
+        sleep=5.0,
+    )
```
dominikschubert, line 54: nit: maybe overkill for this case where we technically only care about CRUD functionality

simonrw: True, but I always like to test the resources created actually work.

PR #11789, opened 2024-11-06
DynamoDB: Fix hangup when resetting state

viren-nadkarni:
## Background

This PR fixes DynamoDB provider hanging up when performing state reset using POST `/_localstack/state/reset`.

## Changes

This bug was caused in the underlying `Server` abstraction, where the `threading.Event` that track the runtime state where not correctly updated.

This was not a problem earlier because fresh server objects were being created after state resets. The DynamoDB server object was recently made a singleton in https://github.com/localstack/localstack/pull/11688

### DIFF
--- a/localstack-core/localstack/services/dynamodb/server.py
+++ b/localstack-core/localstack/services/dynamodb/server.py
@@ -62,7 +62,6 @@ class DynamodbServer(Server):
             self.db_path = None
 
         if self.db_path:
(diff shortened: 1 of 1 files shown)


giograno (approved):
Thanks for jumping on this very quickly and for the provided details!

I tried locally and can confirm this solves the issue 💯

PR #11793, opened 2024-11-06
Introduce CONTAINER_RUNTIME environment variable to control all container runtimes

dfangl:
## Motivation
Currently, we only use the CONTAINER_RUNTIME environment variable in pro.
However, it makes sense to have this variable a global override, including for the lambda runtime executor.

For this, we need to move this variable to community.

## Changes
* Update tracked variables
* Introduce CONTAINER_RUNTIME as environment variable in the open source project and make `LAMBDA_RUNTIME_EXECUTOR` fall back on it.


thrau (approved):
cool new development! :-)  looking forward to see this in action

PR #11794, opened 2024-11-06
Remove virtualenv dependency from docker build

dfangl:
## Motivation
We currently install `virtualenv` into its own layer for the only purpose to create a virtual environment.
However, we can do that just as easily with `python -m venv .venv`. Virtualenv is a nice tool, but its features do not seem to be needed in our docker build.

This change reduces the layers by 1, and should decrease the image size a tad, and we get one less unpinned dependency.

The change should not have any impact on our users, but users might depend on "virtualenv" being installed in init scripts (although unlikely), so I would like to release it with 4.0 as breaking change.

## Changes
* Remove virtualenv installation from docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,10 +78,6 @@ RUN chmod 777 . && \
     chmod 755 /root && \
     chmod -R 777 /.npm
 
-# install basic (global) tools to final image
-RUN --mount=type=cache,target=/root/.cache \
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
LGTM, nice catch! To be honest, I would actually not declare this as a major change, since we change the set of our dependencies all the time. ;)

Issue #2562, opened 2020-06-15
Simulate S3 BadDigest and InvalidDigest errors


charleswhchan:
# Type of request: This is a ...
[ ] bug report
[x] feature request

# Detailed description
As a Developer,
I want to be able to simulate BadDigest and InvalidDigest errors,
So that I can test the error handling logic

## Expected behavior
https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
BadDigest: The Content-MD5 you specified did not match what we received.
InvalidDigest: The Content-MD5 you specified is not valid.

## Actual behavior
localstack returns InvalidDigest when MD5 does not match payload.

whummer:
Thanks for reporting @charleswhchan . Not sure I fully understand your use case - can you please elaborate a bit? I think it should already be possible to pass the `Content-MD5` header on certain S3 API requests. See here for the code that checks the user-specified MD5 hash and compares it to the actual computed hash of the request payload: https://github.com/localstack/localstack/blob/a60eb05ef9f52e606bb08776368b872cc9f8e8cb/localstack/services/s3/s3_listener.py#L706-L714 Can you please provide a small, isolated example of what you're trying to achieve? Thanks!

charleswhchan:
Hi @whummer, thanks for looking into this.

What I am trying to do is write a test to write to S3 with localstack and produce conditions that would lead to BadDigest and InvalidDigest errors. This is so that I can check the error handling is working correctly in my code.

Based on my understanding of the AWS REST API.
In the code you listed above, when "actual != expected", localstack should return BadDigest as the checksum does not match.
Whereas if the content of "Context-MD5" is not a base64 encoded MD5, then the API should return InvalidDigest.

Does this help?

whummer:
Ok, thanks for the clarification @charleswhchan . This should be fixed in #2592 - can you please give it a try with the `latest` version of the Docker image. The code now distinguishes between `BadDigest` and `InvalidDigest`, as per the AWS REST API spec. Hope that helps - please report here if the behavior is still not fully correct. Thanks

charleswhchan:
Hi @whummer. 

I just tried the `latest` docker image ([link](https://hub.docker.com/layers/localstack/localstack/latest/images/sha256-fd7f9f1c1248038b5039fdfeb246e5cd1365ca4411d9c06d26992ea2bbd16ade?context=explore)) and it seems to broken in a different way -- it is now returning `BadDigest` all the time.

[...]

PR #11795, opened 2024-11-06
implement S3 MD5 checksum check for UploadPart and improve logic

bentsku:
## Motivation
As reported and re-opened in #2562, we've had a regression compared to 0.11 (yes!), as some behavior was untested and not caught during early migrations to ASF (S3 v2) and now v3. 

We did not make a difference between `InvalidDigest` and `BadDigest` (Invalid is raised when the given b64 encoded checksum is not a proper b64 encoded checksum, and Bad when the checksum does not match the item). 

Part of the MD5 behavior was re-added with #9064.

## Changes
- update the logic when comparing the ContentMD5
- check that the ContentMD5 is a proper b64-encoded MD5 hash
- add test for `UploadPart` MD5 behavior: funny enough, error messages are not the same between `PutObject` and `UploadPart` (not using the same values)

_fixes #2562_

### DIFF
--- a/localstack-core/localstack/aws/api/s3/__init__.py
+++ b/localstack-core/localstack/aws/api/s3/__init__.py
@@ -974,6 +974,14 @@ class ConditionalRequestConflict(ServiceException):
     Key: Optional[ObjectKey]
 
 
+class BadDigest(ServiceException):
+    code: str = "BadDigest"
+    sender_fault: bool = False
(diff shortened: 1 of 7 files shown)


cloutierMat (approved):
Awesome! That is an impressive level of parity with aws now! :rocket: 

Nice digging to get through all of those testing use case

PR #11796, opened 2024-11-06
add /_extension to well-known path prefixes to avoid s3 fallback handling

thrau:
## Motivation

Over time we've made `/_extension` a well-known path prefix, similar to `/_aws` or `/_localstack`, which still suffers from S3 xml errors when calling URLs within the `/_extension` submount that don't exist. This is a band aid fix for that.

## Changes

* calls to `/_extension/<path>` are now no longer handled by the s3 fallback handler

### DIFF
--- a/localstack-core/localstack/aws/protocol/service_router.py
+++ b/localstack-core/localstack/aws/protocol/service_router.py
@@ -171,6 +171,14 @@ def custom_path_addressing_rules(path: str) -> Optional[ServiceModelIdentifier]:
         return ServiceModelIdentifier("lambda")
 
 
+well_known_path_prefixes = (
+    "/_aws",
+    "/_localstack",
+    "/_pods",
+    "/_extension",
+)
+
+
 def legacy_rules(request: Request) -> Optional[ServiceModelIdentifier]:
     """
     *Legacy* rules which migrate routing logic which will become obsolete with the ASF Gateway.
@@ -192,8 +200,9 @@ def legacy_rules(request: Request) -> Optional[ServiceModelIdentifier]:
 
     # TODO Remove once fallback to S3 is disabled (after S3 ASF and Cors rework)
     # necessary for correct handling of cors for internal endpoints
-    if path.startswith("/_localstack") or path.startswith("/_pods") or path.startswith("/_aws"):
-        return None
+    for prefix in well_known_path_prefixes:
+        if path.startswith(prefix):
+            return None
 
     # TODO The remaining rules here are special S3 rules - needs to be discussed how these should be handled.
     #      Some are similar to other rules and not that greedy, others are nearly general fallbacks.


localstack-core/localstack/aws/protocol/service_router.py
```diff
@@ -192,8 +200,9 @@ def legacy_rules(request: Request) -> Optional[ServiceModelIdentifier]:
 
     # TODO Remove once fallback to S3 is disabled (after S3 ASF and Cors rework)
     # necessary for correct handling of cors for internal endpoints
-    if path.startswith("/_localstack") or path.startswith("/_pods") or path.startswith("/_aws"):
-        return None
+    for prefix in well_known_path_prefixes:
+        if path.startswith(prefix):
+            return None
```
bentsku, line 205: tiny nit:
```suggestion
    if any(path.startswith(prefix) for prefix in well_known_path_prefixes):
        return None
```

bentsku (approved):
Ah, good catch! Looking at this, I wonder if we often have the case that we have edge routes that actually need to have the service, especially since it does not have access to the request context. I think we'd be almost ready to move the service name parser after the edge routes to be honest, as the CORS detection does not rely on the services anymore (managed for S3 and APIGW, with some heuristics). 

It doesn't really change anything to this particular issue though, but it made me think that maybe we could have a way to register some base path in our edge router (similar to Flask Blueprints somehow) so that we could return some default values when hitting some base path, instead of continuing the chain and having those hardcoded rules. I can see the issues we could get with that, as it'd be maybe difficult to add dynamically add routes to those "blueprints". 

But yeah, maybe we should think about registering rules for those base endpoints, so that we could avoid those tricks? What do you think?

In any case, this is a good fix and good to merge! 👌 

edit: also, as a nit, I think it's not really a S3 fallback handler anymore since we removed the proxy listeners, just that those requests fit the routing rules of the S3 specs which are extremely greedy 😄

localstack-core/localstack/aws/protocol/service_router.py
thrau: i suppose this is highly subjective, but i don't perceive this as more readable. maybe it takes some getting used to :-)

thrau:
all great points @bentsku! thanks for greenlighting this, while also noticing some systemic issues we're not addressing here. definitely let's rework this soon!

localstack-core/localstack/aws/protocol/service_router.py
bentsku: Agreed yes, I've learned to really like `all` and `any`, but doing a quick test, seems like it's actually slower 😄 forget the nit 😄

thrau: always happy to have debates over style ;-)

PR #11797, opened 2024-11-06
Parity test for Eventbridge logs target

zaingz:
## Motivation
This PR adds parity tests for eventbridge logs target. 

## Changes
None in code. 

## Testing
```
PROVIDER_OVERRIDE_EVENTS=v2 DEBUG=1 make test TEST_PATH=tests/aws/services/events/test_events_targets.py::TestEventsTargetCloudWatchLogs::test_put_events_with_target_cloudwatch_logs
```

### DIFF
--- a/localstack-core/localstack/services/events/target.py
+++ b/localstack-core/localstack/services/events/target.py
@@ -181,20 +181,25 @@ class TargetSender(ABC):
             self._validate_input_transformer(input_transformer)
 
     def _initialize_client(self) -> BaseClient:
-        """Initializes internal boto client."""
+        """Initializes internal boto client.
+        If a role from a target is provided, the client will be initialized with the assumed role.
+        If no role is provided, the client will be initialized with the account ID and region.
+        In both cases event bridge is requested as service principal"""
         service_principal = ServicePrincipal.events
         role_arn = self.target.get("RoleArn")
-        if role_arn:
+        if role_arn:  # required for cross account
+            # assumed role sessions expire after 6 hours in AWS, currently no expiration in LocalStack
             client_factory = connect_to.with_assumed_role(
                 role_arn=role_arn,
                 service_principal=service_principal,
                 region_name=self.region,
             )
-            client = client_factory.get_client(self.service)
-            client.request_metadata(service_principal=service_principal, source_arn=self.rule_arn)
         else:
-            # Use connect_to to get the client
-            client = connect_to(region_name=self.region).get_client(self.service)
+            client_factory = connect_to(aws_access_key_id=self.account_id, region_name=self.region)
+        client = client_factory.get_client(self.service)
+        client = client.request_metadata(
+            service_principal=service_principal, source_arn=self.rule_arn
+        )
         return client
 
     def _validate_input_transformer(self, input_transformer: InputTransformer):
@@ -459,8 +464,7 @@ class LambdaTargetSender(TargetSender):
 class LogsTargetSender(TargetSender):
     def send_event(self, event):
         log_group_name = self.target["Arn"].split(":")[6]
-        log_stream_name = str(uuid.uuid4())
-
+        log_stream_name = str(uuid.uuid4())  # Unique log stream name
         self.client.create_log_stream(logGroupName=log_group_name, logStreamName=log_stream_name)
         self.client.put_log_events(
             logGroupName=log_group_name,
--- a/localstack-core/localstack/services/events/target.py
+++ b/localstack-core/localstack/services/events/target.py
@@ -181,26 +181,24 @@ class TargetSender(ABC):
             self._validate_input_transformer(input_transformer)
 
     def _initialize_client(self) -> BaseClient:
-        """Initializes internal boto client.
-        If a role from a target is provided, the client will be initialized with the assumed role.
-        If no role is provided, the client will be initialized with the account ID and region.
-        In both cases event bridge is requested as service principal"""
+        """Initializes internal boto client."""
         service_principal = ServicePrincipal.events
         role_arn = self.target.get("RoleArn")
-        if role_arn:  # required for cross account
-            # assumed role sessions expire after 6 hours in AWS, currently no expiration in LocalStack
+        if role_arn:
             client_factory = connect_to.with_assumed_role(
                 role_arn=role_arn,
                 service_principal=service_principal,
                 region_name=self.region,
             )
+            client = client_factory.get_client(self.service)
+            client.request_metadata(
+                service_principal=service_principal, source_arn=self.rule_arn
+            )
         else:
-            client_factory = connect_to(aws_access_key_id=self.account_id, region_name=self.region)
-        client = client_factory.get_client(self.service)
-        client = client.request_metadata(
-            service_principal=service_principal, source_arn=self.rule_arn
-        )
+            # Use connect_to to get the client
+            client = connect_to(region_name=self.region).get_client(self.service)
         return client
+
 
     def _validate_input_transformer(self, input_transformer: InputTransformer):
         if "InputTemplate" not in input_transformer:
@@ -465,6 +463,11 @@ class LogsTargetSender(TargetSender):
     def send_event(self, event):
         log_group_name = self.target["Arn"].split(":")[6]
         log_stream_name = str(uuid.uuid4())  # Unique log stream name
+
+        LOG.warn(f"log_group_name: {log_group_name}")
+        LOG.warn(f"log_stream_name: {log_stream_name}")
+        LOG.warn(f"event: {event}")
+
         self.client.create_log_stream(logGroupName=log_group_name, logStreamName=log_stream_name)
         self.client.put_log_events(
             logGroupName=log_group_name,
--- a/localstack-core/localstack/services/events/target.py
+++ b/localstack-core/localstack/services/events/target.py
@@ -191,14 +191,11 @@ class TargetSender(ABC):
                 region_name=self.region,
             )
             client = client_factory.get_client(self.service)
-            client.request_metadata(
-                service_principal=service_principal, source_arn=self.rule_arn
-            )
+            client.request_metadata(service_principal=service_principal, source_arn=self.rule_arn)
         else:
             # Use connect_to to get the client
             client = connect_to(region_name=self.region).get_client(self.service)
         return client
-
 
     def _validate_input_transformer(self, input_transformer: InputTransformer):
         if "InputTemplate" not in input_transformer:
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/events/target.py
```diff
@@ -181,25 +181,20 @@ def _validate_input(self, target: Target):
...
         role_arn = self.target.get("RoleArn")
-        if role_arn:  # required for cross account
-            # assumed role sessions expire after 6 hours in AWS, currently no expiration in LocalStack
+        if role_arn:
             client_factory = connect_to.with_assumed_role(
                 role_arn=role_arn,
                 service_principal=service_principal,
                 region_name=self.region,
             )
+            client = client_factory.get_client(self.service)
```
maxhoheiser, line 193: what is the reason behind reversing the behavior - the idea is that if a role_arn is provided it is used as the source arn and not the rule_arn as initially implemented.

localstack-core/localstack/services/events/target.py
```diff
@@ -181,25 +181,20 @@ def _validate_input(self, target: Target):
...
     def _initialize_client(self) -> BaseClient:
-        """Initializes internal boto client.
-        If a role from a target is provided, the client will be initialized with the assumed role.
-        If no role is provided, the client will be initialized with the account ID and region.
-        In both cases event bridge is requested as service principal"""
+        """Initializes internal boto client."""
         service_principal = ServicePrincipal.events
         role_arn = self.target.get("RoleArn")
-        if role_arn:  # required for cross account
-            # assumed role sessions expire after 6 hours in AWS, currently no expiration in LocalStack
```
maxhoheiser, line 191: we should leave this comment hin since it is something that is currently not parity with AWS and could lead to problems (very unlikely but might be)

localstack-core/localstack/services/events/target.py
```diff
@@ -181,25 +181,20 @@ def _validate_input(self, target: Target):
             self._validate_input_transformer(input_transformer)
 
     def _initialize_client(self) -> BaseClient:
-        """Initializes internal boto client.
```
maxhoheiser, line 184: Why remove the docstring - it makes it easier to understand why we initialize it with the role_arn as source

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,117 @@
...
+            policyName=policy_name, policyDocument=json.dumps(resource_policy)
+        )
+
+        # Wait for the resource policy to take effect (only needed in AWS)
+        if is_aws_cloud():
+            time.sleep(10)
+
+        # Step 4: Create a rule on the event bus
+        rule_name = f"test-rule-{short_uid()}"
+        event_pattern = {"source": ["test.source"], "detail-type": ["test.detail.type"]}
```
maxhoheiser, line 93: you can use the default pattern TEST_EVENT_PATTERN from test_events - makes it easier to avoid problems with not matching rules - and also use the default message from test_events

maxhoheiser (changes requested):
Hi - supper cool test thank you for adding it :) :

I have a few questions for your changes of the target client - added them as comments :)

localstack-core/localstack/services/events/target.py
zaingz: hey, thanks for catching these. I had a issue related to JAVA_PATH and to solve it I tried few things. I committed these changes accidentally. I have reverted them now.

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,116 @@
...
+        put_events_response = aws_client.events.put_events(Entries=[event_entry])
+        snapshot.match("put_events_response", put_events_response)
+        assert put_events_response["FailedEntryCount"] == 0
+
+        # Step 7: Verify that the event was logged to the log group
+        def get_log_events():
+            response = aws_client.logs.describe_log_streams(logGroupName=log_group_name)
+            log_streams = response.get("logStreams", [])
+            if not log_streams:
+                raise Exception("No log streams found")
```
gregfurman, line 131: nit: the common pattern when using the `retry` util function is normally using an `assert` lile:

```suggestion
            assert log_streams
```

could also optionally pass some info:
```suggestion
            assert log_streams, "No log streams found"
```

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,116 @@
...
+            log_stream_name = log_streams[0]["logStreamName"]
+            events_response = aws_client.logs.get_log_events(
+                logGroupName=log_group_name,
+                logStreamName=log_stream_name,
+            )
+            events = events_response.get("events", [])
+            if not events:
+                raise Exception("No log events found")
```
gregfurman, line 139: nit: Same `assert` comment as above 🙂

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,116 @@
...
+                {
+                    "Sid": "EventBridgePutLogEvents",
+                    "Effect": "Allow",
+                    "Principal": {"Service": "events.amazonaws.com"},
+                    "Action": ["logs:CreateLogStream", "logs:PutLogEvents"],
+                    "Resource": f"{log_group_arn}:*",
+                }
+            ],
+        }
+        policy_name = f"EventBridgePolicy-{short_uid()}"
```
gregfurman, line 82: I wonder if it's worth creating a fixture for this?

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,116 @@
...
+            if not events:
+                raise Exception("No log events found")
+            return events
+
+        events = retry(get_log_events, retries=5, sleep=5)
+        snapshot.match("log_events", events)
+
+        # Cleanup: delete the log group
+        aws_client.logs.delete_log_group(logGroupName=log_group_name)
+        aws_client.logs.delete_resource_policy(policyName=policy_name)
```
gregfurman, line 147: Perhaps rather use the `cleanup` fixture since these resources could end up hanging around if they're created and the test exits before this is reached.

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,116 @@
...
+    ):
+        snapshot.add_transformers_list(
+            [
+                snapshot.transform.key_value("EventId"),
+                snapshot.transform.key_value("RuleArn"),
+                snapshot.transform.key_value("EventBusArn"),
+            ]
+        )
+
+        # Step 1: Create an EventBridge event bus
```
gregfurman, line 59: Appreciate the comments here! Think a lot of the code is self-documenting though i.e `events_create_event_bus` signals the bus is being created.

For future reference, don't think you need to worry about adding comments/docs each time though (unless there's something particularly weedy that you think justifies the extra info 🧠).

If you think the extra docs are necessary, perhaps rather add a docstring to the test function that explains what the test does etc.

gregfurman (commented):
Nice one! Happy to approve once a `cleanup` fixture is used. Otherwise, a couple of nits/suggestions that you can feel free to ignore.

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,106 @@
 #   - Container (pro)
 #   - Redshift (pro)
 #   - Sagemaker (pro)
+class TestEventsTargetCloudWatchLogs:
```
joe4dev, line 40: nit: The first two TODOs above (L32-33) can be removed

localstack
joe4dev, line 1: This submodule looks like an accidental commit adding lots of reviewers :)

tests/aws/services/events/conftest.py
```diff
@@ -383,6 +383,56 @@ def _put_events_with_filter_to_sqs(
...
+            "policy_name": policy_name,
+        }
+
+    yield _create_log_group
+
+    for log_group in log_groups:
+        try:
+            aws_client.logs.delete_log_group(logGroupName=log_group)
+        except Exception:
+            pass
```
joe4dev, line 427: Would it make sense to log here or do we think logging just clutters the CI logs?

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,106 @@
...
+        }
+        policy_name = f"EventBridgePolicy-{short_uid()}"
+        aws_client.logs.put_resource_policy(
+            policyName=policy_name, policyDocument=json.dumps(resource_policy)
+        )
+
+        cleanups.append(lambda: aws_client.logs.delete_log_group(logGroupName=log_group_name))
+        cleanups.append(lambda: aws_client.logs.delete_resource_policy(policyName=policy_name))
+
+        if is_aws_cloud():
```
joe4dev, line 87: Can we add a comment indicating on what we wait?

tests/aws/services/events/test_events_targets.py
```diff
@@ -37,6 +37,106 @@
...
+                    "Resource": f"{log_group_arn}:*",
+                }
+            ],
+        }
+        policy_name = f"EventBridgePolicy-{short_uid()}"
+        aws_client.logs.put_resource_policy(
+            policyName=policy_name, policyDocument=json.dumps(resource_policy)
+        )
+
+        cleanups.append(lambda: aws_client.logs.delete_log_group(logGroupName=log_group_name))
```
joe4dev, line 84: Do we need these cleanups or can we rely on the fixture to clean up the log groups?

joe4dev (commented):
The main suggestion is to avoid committing the "localstack" submodule. I added some smaller suggestions, but overall it LGTM.

localstack-core/localstack/services/events/target.py
maxhoheiser: yes JAVA_PATH running it locally on mac is a problem
did you fix it?
I always hardcode my java home and change the ending to dylib

```
 java_home = "/usr/local/Cellar/openjdk/23/libexec/openjdk.jdk/Contents/Home"
    jvm_lib = Path(java_home) / "lib" / "server" / "libjvm.dylib"
```

tests/aws/services/events/test_events_targets.py
pinzon, line 84: This cleanup seems redundant when the fixuture is already going to do the cleaning

pinzon (commented):
I'm not sure why I was pinged as a reviewer. Regardless, everything seems in order with just some comments from @joe4dev that I recommend to address.

tests/aws/services/events/test_events_targets.py
zaingz: the fixture didn't clean these log groups therefore I have to manually do the cleanup

joe4dev (approved):
LGTM 👍 
Congratulations and thank you for your patience around the review comments 🙇 

Feel free to merge once the CI is 🟢 🚀

PR #11799, opened 2024-11-06
APIGW: fix requestOverride not being present in responseTemplates

bentsku:
## Motivation
As reported in our Community Slack, having a similar flow as the one added in this test in a VTL template did not work currently. 

The idea is to get some data from the incoming request, maybe transform it and set it in the `context.requestOverride`. 
Then, the response template can access it and transform it/do conditional logic on it. 

As we don't want to directly mutate the passed context, we deep copy it before sending it down for rendering (as we don't fully control the rendering logic and what is protected or not). 
We then need to attach the `requestOverride` data back to the context so that it is available in the `context` in the response template to execute logic on it. 

## Changes
- attach back the `requestOverride` to the context variables
- add a pretty complex test checking logic to parse b64 encoded data in the path to attach it to `requestOverride`, then parse the JSON back in the response template and execute logic on it


cloutierMat (approved):
Interesting use case for the request override! :eyes: Thanks for adding such a clear test as well!

PR #11801, opened 2024-11-07
remove APIGW CORS handler and update global CORS logic

bentsku:
## Motivation
TBD

## Changes
TBD


localstack-core/localstack/aws/handlers/cors.py
```diff
@@ -143,6 +143,15 @@ def _get_allowed_cors_origins() -> List[str]:
...
 
 
+def is_execute_api_call(context: RequestContext) -> bool:
+    path = context.request.path
+    return (
+        ".execute-api." in context.request.host
+        or (path.startswith("/restapis/") and f"/{PATH_USER_REQUEST}" in context.request.path)
+        or (path.startswith("/_aws/execute-api"))
+    )
+
```
bentsku, line 153: TLDR; just future idea, this is not very important for this PR or the review.

Note: this is extracted for now. We only have 2 services that self-manage CORS: S3 and API Gateway. S3 does not need "outside logic" to be applied thanks to the `preprocess_request` dynamic handler. 

API Gateway needs additional logic by patching this method.

In the future we might want to create an abstraction around it. I think "self-managed CORS" services could import maybe the `handlers.enforce_cors` Handler and call something like `register_self_managed_check(service_name, check_function)` or something instead of patching, similar to how services can register handlers in the chain right now via some predefined `CompositeHandler`. 

This way, we could have an internal map like `{"apigateway": [check_function, check_function2], "s3": check_function_s3}` that could be used to determine if the request should be self-managed or not, similar to what we have in `cors.should_enforce_self_managed_service` right now, but less hardcoded and with a better interface. 

Anyway, this is just some idea for extensibility and maintainability in the future, so need to over engineer this yet, we only have 2 services 😄

cloutierMat (commented):
awesome work!

I am a bit slow to understand the whole CORS thing, so pardon my stupic question! :smile: 

With that change, and removing the apigw cors handler, apigw_v1 is effectively no longer self managing cors? It only has a flag that allow a user to turn off CORS?

cloutierMat (approved):
Sorry, bad case of the Friday morning sleepies!

They are either managed by the user's apigw or the integration behind it, or if disabled, managed by Localstack!

Under that light! :+1: 

Awesome that the reorder of the service name parser allows us to remove that extra handler, that only duplicates what the LocalStack handler was already doing!

localstack-core/localstack/aws/handlers/cors.py
alexrashed: This sounds a lot like a plugin registry 😛 I would propose making the CORS handler pluggable with `plux` 😉

alexrashed (approved):
Absolutely great to see this change happening in order to enable the service name parser to be moved down the the AWS specific parts of the chain! 😍 🦸🏽
Thanks a lot for keeping an eye on this, and bringing this up again! 💯

localstack-core/localstack/aws/handlers/cors.py
bentsku: Agreed! I think as soon as another service needs to plugin with CORS, we need to make this happen. Right now, only one plugin would be there, but we need to keep that in mind 👍

PR #11802, opened 2024-11-07
Cfn/fix lambda update

cloutierMat:
## Motivation

_fixes #11365_

Updating lambda code through cloud formation was triggering a creation/deletion of a lambda function. This was only happening when the lambda name was not provided in the template.

As such when updating the code of a lambda authorizer, the name was changing and was no longer reachable by apigw. On AWS the code on a lambda function can change without the need to create a new deployment of a rest api.

## Changes

- Prevent replacing the function when the `functionName` is not provided
- Added Cfn Lambda tests to validate the behavior
- Added small apigw update test to validate that changes to the methods it depends on do not trigger the deployment

Out Of Scope
This pr does not address updating apigw resources such as authorizer, deployment and stages. More digging will be required to get right and this small pr does enough to fix the user issue. It seems Cfn does never recreate the deployment while the CDK seems to trigger it through some configuration (but I believe it does it by changing it's logical id, which should trigger a create/delete from Cfn?). The apigw test was added as have a a safeguard when we start implementing redeployment.


pinzon (approved):
Great PR! :+1:

Issue #11299, opened 2024-08-02
bug: event-ruler failing to install in Docker with EventBridgeV2


sujee81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm trying to test multi-accounts EventBridge integration using LocalStack. I have enabled EventBridgeV2 and when I send a message to source EventBus from the localstack web app, I get an error `Installation of event-ruler failed.`. This happens regardless if I test multi-accounts event bus or just the local bus. EventBridgeV1 works fine for local bus but does not deliver across accounts. Appreciate your help on looking at this issue.

Docker compose (WSL Ubuntu on Windows)
```
services:
  localstack:
    container_name: "localstack-main"
    image: localstack/localstack
    ports:
      - "127.0.0.1:4566:4566"
      - "127.0.0.1:4510-4559:4510-4559"
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
    environment:
      - PROVIDER_OVERRIDE_EVENTS=v2
```

Debug message (Copied from web app error dialog, not formatted nicely)
```
[...]

PR #11807, opened 2024-11-07
Switch default event rule engine from python to java

joe4dev:
## Motivation

Our current Python-native event rule engine has many parity limitations compared to the Java-based event rule engine offering perfect AWS parity. This engine is used for:
* [Amazon EventBridge event pattern](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html)
* [Lambda Event Source Mapping event filtering](https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html)
* [Event filtering in Amazon EventBridge Pipes](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-filtering.html)

This switch implies that we pick AWS parity over performance by default, but still offer users the option to choose.
Users behind enterprise proxies might also be interested in switching to the python-native event rule engine if they cannot download the JAR files from Maven central as reported in this issue https://github.com/localstack/localstack/issues/11299

## Changes

* Change default for `EVENT_RULE_ENGINE` from python to java
* Run full event pattern matching test suite for both event rule engines
* Remove redundant monkeypatching for tests that are now passing by default with the new default event rule engine
* Add feature flag to analytics such that we can track how many users switch the engine

## Limitations

The Java-based event rule engine has a performance penalty compared to the native Python implementation, especially for tail latency (spikes up to ~200ms).
For performance-sensitive users, we recommend opting in to the Python event rule engine if possible using `EVENT_RULE_ENGINE=python`.


joe4dev:
We decided against switching the default rule engine to avoid compatibility issues with the StepFunctions JPype integration. Therefore, we are closing this PR in favor of just adding the analytics in https://github.com/localstack/localstack/pull/11928

PR #11809, opened 2024-11-07
remove S3 persistence workaround

bentsku:
## Motivation
When implementing new features in S3 since `3.0.0`, we've taken extra care to not break persistence by using `getattr` around store attributes and have additional conditionals in place. 

As we're approaching a major release, I've collected the TODO related to this and removed the `getattr` blocks to simplify the logic. If people loaded earlier state and then saved it again after using it with a version with those calls, the state should not be broken anyway. 

## Changes
- remove TODOs and `getattr` around store operations, as v4 is probably going to break persistence in multiple ways with all the provider changes.


cloutierMat (approved):
LGTM!

The power of TODOs in display here

Issue #11803, opened 2024-11-07
LocalStack v4 release 🚀


dominikschubert:
Hello everyone! We are excited to announce that we are once more preparing a major release of LocalStack.
This release will introduce a lot of great new features, but will also clean up some legacy code and configurations.
This issue is meant to give you a heads-up on the upcoming changes such that you can already prepare your LocalStack setup accordingly.

## Tentative Timeline

- [x] Thursday, November 14, 2024: `latest` ships the v4 release candidate.
    -  :warning: We will merge the release branch `v4` into `master`, updating the `latest` Docker image with the release candidate of v4.
    - This merge will implement all breaking changes in v4, impacting your pipeline if you use the `latest` tag.
- [x]  Thursday, November 21, 2024 - Friday, November 22, 2024: `v4` release
    - We will publish the tagged release v4.0.0.
 

### Auth Tokens to replace CI Keys

Auth tokens have fully replaced legacy API and CI keys for both personal and CI use, increasing security and enabling more detailed access control.
This update includes:

1.  **Personal Developer Auth Token**: Remains unchanged for individual users, offering unique tokens for accessing the core cloud emulator and other platform features.
2.  **New CI Auth Token**: A team-shared token designed for use in CI and automated environments, allowing access to the core cloud emulator and other platform features.

Existing API and CI keys **will function until early 2025**, after which only Auth Tokens will be supported. Transitioning is straightforward. You can set the new token in the `LOCALSTACK_AUTH_TOKEN` variable. For a temporary period, legacy credentials can be used by configuring the new Auth Token in the `LOCALSTACK_API_KEY` variable to maintain backward compatibility.

### Migration of Lambda K8s Executor in Enterprise

We've updated our plan offerings and the Lambda Kubernetes Executor is now exclusively available with our Enterprise plan. If you're using any tier other than Enterprise, setting `LAMBDA_RUNTIME_EXECUTOR` ( or `lambda.executor` when using the Helm configuration) to `kubernetes` will not have any effect.

### Deprecation of non-prefixed environment variables

The LocalStack CLI is closely integrated with the runtime and needs to be able to determine which environment variables it should pass from the user's environment to the Docker container. This requirement restricts the CLI’s compatibility with new environment variables introduced in later LocalStack versions, often leading to issues that are challenging to debug. Using a well-known prefix  (`LOCALSTACK_`) will enable forwards-compatibility for newer configuration options.

[...]

PR #11810, opened 2024-11-07
deprecate env vars without prefix in CLI

alexrashed:
## Motivation
As announced in https://github.com/localstack/localstack/issues/11803, with LocalStack 4.0 we are deprecating the usage of non-prefixed environment variables in the CLI:

> The LocalStack CLI is closely integrated with the runtime and needs to be able to determine which environment variables it should pass from the user's environment to the Docker container. This requirement restricts the CLI’s compatibility with new environment variables introduced in later LocalStack versions, often leading to issues that are challenging to debug. Using a well-known prefix (`LOCALSTACK_`) will enable forwards-compatibility for newer configuration options. To enhance flexibility, users are encouraged to use a variable prefix (`LOCALSTACK_`) for more straightforward management of environment variables across different versions.

This PR adds a deprecation warning to inform users about this change.

## Changes
- Add a warning if a non-prefixed environment variable is loaded to be forwarded to the runtime container.
- Loads the env var "LOG" if "LS_LOG" is not present to allow setting "LOCALSTACK_LOG" instead of "LOCALSTACK_LS_LOG" in Docker mode.

## TODO
- [x] Add prefix for environment variables defined in a profile.
- [x] Check where the best place to show the warning is.
  - Unfortunately, the place for the warning right now isn't the best, but we can properly separate this in upcoming iterations.

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -431,8 +431,8 @@ VOLUME_DIR = os.environ.get("LOCALSTACK_VOLUME_DIR", "").strip() or TMP_FOLDER
 if TMP_FOLDER.startswith("/var/folders/") and os.path.exists("/private%s" % TMP_FOLDER):
     TMP_FOLDER = "/private%s" % TMP_FOLDER
 
-# whether to enable verbose debug logging
-LS_LOG = eval_log_type("LS_LOG")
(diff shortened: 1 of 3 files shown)


thrau (approved):
LGTM!

Issue #11537, opened 2024-09-18
bug: EventBridge pattern using suffix generates incorrect rule


arshsingh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm trying to create an EventBridge rule using the `suffix` matcher, with CDK:

```javascript
new Rule(this, "TestRule", {
  eventPattern: {
    source: ["aws.s3"],
    detailType: ["Object Created"],
    detail: {
      bucket: { name: ["test-s3-bucket"] },
      object: { key: [{ suffix: "/test.json" }] },
    },
  },
  targets: [...],
});
```

Creating the rule succeeds, but when there is an event it throws an error saying the rule is invalid (somthing like: suffix value must be a string, not an array). Upon further inspection it looks like the rule this ends up creating contains the pattern: `{ suffix: ["/test.json"] }`

However If I change this to `prefix`, it creates the rule correctly without the array: `{ prefix: "/test.json" }`

### Expected Behavior

In the rule created by cloudformation the value`"/test.json"` should not be inside an array

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

- create an evenbridge rule using CDK/cloudformation and use the `suffix` matcher in the event pattern
- call `awslocal events describe-rule --name {rule name here}`
- the EventPattern in the rule will have the suffix value inside an array

### Environment

```markdown
- OS: Mac OS 14
- LocalStack:
  LocalStack version: 3.7.2
  LocalStack Docker image sha: sha256:6121fe59d0c05ccc30bee5d4f24fce23c3e3148a4438551982a5bf60657a9e8d
  LocalStack build date: 2024-09-06
  LocalStack build git hash: 854016a0
```

u-minor:
Hi @pinzon, could you support `wildcard` and other keywords?

https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-pattern-operators.html#eb-filtering-wildcard-matching
https://github.com/aws/event-ruler?tab=readme-ov-file#wildcard-matching

pinzon:
The solution for this issue has been merged. Also includes more operations as requested. :+1:

PR #11811, opened 2024-11-07
fix transformation for matching operations in aws::events::rule

pinzon:
## Motivation
Adresses #11537. This PR extends the list of attributes that need to be converted in a EventPattern inside the resource provider for AWS::Events::Rule
Operations added are found in this [AWS doc](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-pattern-operators.htm)

## Changes
- simple addition of operations to the list of atts to be converted

## Testing
- new aws validated test.


maxhoheiser (approved):
Great addition - the failing tests kinesis-firehose-s3 prbly is a flake - feel free to merge once it passes :)

PR #11812, opened 2024-11-07
Remove cloudtrail tracking module

simonrw:
## Motivation

In 1e15828e7b589a0be0a887e04b80e3c3e9ceeb11 we introduced a module to help track AWS SDK calls from CloudFormation resource providers. We implemented this using CloudTrail in the `cloudtrail_tracking` module.

Since it is its own project stored in the main code base, it does not behave like a Python module, and as such causes weird behaviour. Most notably: errors when generating the entrypoints, but also during test collection.

## Changes

Remove the cloudtrail tracking module


dominikschubert (approved):
Thanks for cleaning this up. The warnings this caused has unfortunately led to some confusion when onboarding devs in the past, so it's good to extract this from the core codebase :+1:

simonrw:
The single test failure is unrelated, and a flake as it passes on ARM. I will merge this.

Issue #11597, opened 2024-09-27
bug: EventBridge v2 put events returns duplicated EventId in case matching more than one rule and target


artur-nogueira:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello

I apologise in advance if this is already an unknown issue, but I haven't found a related thread.

I found a small bug while using localstack (v 3.7.2) with EventBridge v2. I've noticed the response from PutEventsCommand for valid events includes the `EventId` for every matched rule and respective targets, when it should only return once even if multiple rules match that dispatched event. This is visible when we handle the SDK response to match the input events by the ID for example.
 
I think the code line in question is this one [link](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/events/provider.py#L1422).

### Expected Behavior

The expected behaviour of the Put Events response should return the respective ID of each event (while preserving order) only once regardless if there are multiple matched rules/targets.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

I run localstack version 3.7.2 using docker compose.

`docker-compose.yml`:
```
services:
  localstack:
    image: localstack/localstack:3.7.2
    container_name: test-localstack
    ports:
      - '4566:4566' # LocalStack Gateway
      - '4510-4559:4510-4559' # external services port range
    networks:
      - service-net
    environment:
      - DEBUG=1 # Debug level 1 if you want to logs, 0 if you want to disable
      - START_WEB=0 # Flag to control whether the Web UI should be started in Docker
      - DOCKER_HOST=unix:///var/run/docker.sock
      - DATA_DIR=/tmp/localstack/data #  Local directory for saving persistent data(Example: es storage)
      - DEFAULT_REGION=eu-west-1
      - PROVIDER_OVERRIDE_EVENTS=v2
    volumes:
      - '/var/run/docker.sock:/var/run/docker.sock'

networks:
  service-net:
```

Start container with `docker compose up`

For the following example, I'm using the CLI to create two event rules for the same detail type. And each event rule has two targets (a mock SQS queue). Then, I perform a put events which will return the same EventId 4 times in this case (2 rules * 2 targets).

1. Create a mock SQS to be used as rules target
`aws sqs create-queue --queue-name test-queue --endpoint-url http://localhost:4566`

2. Create two event rules
[...]

PR #11814, opened 2024-11-08
Added tests and fixed the issue of the multiple targets

zaingz:
## Motivation
Fixes EventBridge PutEvents response so it correctly returns event IDs only once per event, regardless of how many rules or targets match. Currently, the response includes duplicate EventIds for each matching target, which doesn't match AWS behavior and can cause issues when clients try to track events by their IDs.

Related issue: https://github.com/localstack/localstack/issues/11597

## Changes
- Modified `_process_entries` and related methods in events provider to track and return event IDs only once
- Added new helper method `_process_target` to handle target processing without duplicating event IDs
- Refactored event processing logic to maintain a single event result per input event
- Added comprehensive test case to verify the fix and prevent regressions

## Testing
1. Create an EventBridge rule with multiple targets
2. Send an event that matches the rule
3. Verify that:
   - PutEvents response contains each EventId only once
   - All targets still receive the event correctly
   - Event IDs are consistent between response and delivered messages

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -1938,24 +1938,27 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         via the target sender put to the defined archived."""
         processed_entries = []
         failed_entry_count = {"count": 0}
+
         for event in entries:
-            self._process_entry(event, processed_entries, failed_entry_count, context)
+            # Process each entry once and track any failures
+            entry_result = self._process_entry(event, failed_entry_count, context)
+            processed_entries.append(entry_result)
+
         return processed_entries, failed_entry_count["count"]
 
     def _process_entry(
         self,
         entry: PutEventsRequestEntry,
-        processed_entries: PutEventsResultEntryList,
         failed_entry_count: dict[str, int],
         context: RequestContext,
-    ) -> None:
+    ) -> dict:
+        """Process a single event entry and return its result."""
         event_bus_name_or_arn = entry.get("EventBusName", "default")
         event_bus_name = extract_event_bus_name(event_bus_name_or_arn)
         if event_failed_validation := validate_event(entry):
-            processed_entries.append(event_failed_validation)
             failed_entry_count["count"] += 1
             LOG.info(json.dumps(event_failed_validation))
-            return
+            return event_failed_validation
         region, account_id = extract_region_and_account_id(event_bus_name_or_arn, context)
         if encoded_trace_header := get_trace_header_encoded_region_account(
             entry, context.region, context.account_id, region, account_id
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/events/provider.py
```diff
@@ -1364,22 +1366,96 @@ def _process_entry(
...
+    ) -> bool:
+        """Process all rules for an event. Returns True if any processing error occurred."""
+        has_error = False
+
+        if not rules:
             LOG.info(
                 json.dumps(
                     {
                         "InfoCode": "InternalInfoEvents at process_rules",
-                        "InfoMessage": f"No rules attached to event_bus: {event_bus_name}",
```
maxhoheiser, line 1379: this logging here is especially relevant for EventStudio - I am fine with changing the naming to configured but we need the event bus name here.

localstack-core/localstack/services/events/provider.py
```diff
@@ -1364,22 +1366,96 @@ def _process_entry(
...
+                        "InfoMessage": "No rules configured for event bus",
                     }
                 )
             )
+            return has_error
+
+        for rule in rules:
+            event_pattern = rule.event_pattern
```
maxhoheiser, line 1409: we need to align these changes with EventStudio before we merge this - since we are patching specific methods and these changes here will break things.

localstack-core/localstack/services/events/provider.py
```diff
@@ -1364,22 +1366,96 @@ def _process_entry(
...
+        if has_processing_error:
+            failed_entry_count["count"] += 1
+            return {
+                "ErrorCode": "InternalException",
+                "ErrorMessage": "Error occurred while processing one or more targets",
+            }
+
+        return {"EventId": event_formatted["id"]}
+
+    def _process_rules_for_event(
```
maxhoheiser, line 1386: FYI we also need to be thoughtful with renaming methods since we patch specific methods in EventStudio and have to always adapt the changes there too (lets minimize renaming)

localstack-core/localstack/services/events/provider.py
```diff
@@ -1364,22 +1366,96 @@ def _process_entry(
...
                 )
             )
+            return has_error
+
+        for rule in rules:
+            event_pattern = rule.event_pattern
+            event_str = to_json_str(event_formatted)
+
+            if not matches_rule(event_str, event_pattern):
+                continue
```
maxhoheiser, line 1413: we need to keep all the logging messages previously since the are relevant for EventStudio internal failure detection.

```
LOG.info(
                json.dumps(
                    {
                        "InfoCode": "InternalInfoEvents at matches_rule",
                        "InfoMessage": f"No rules matched for formatted event: {event_formatted}",
                    }
                )
            )
```

maxhoheiser (changes requested):
Hi Zain - cool changes, some details we need to align before we can merge this in regards to EventStudio :)

localstack-core/localstack/services/events/provider.py
zaingz: This is a new internal method. used within `_process_entry`. Could this be a problem?

zaingz:
replaced by: https://github.com/localstack/localstack/pull/11880/files

PR #11816, opened 2024-11-08
Added EVENT_RULE_ENGINE flag based opt-in

zaingz:
## Motivation
This PR adds support for a Python-based event rule engine as an alternative to the Java implementation. 
This addresses installation issues some users face with the Java event ruler (GH-11299), particularly 
in environments with proxy setups or SSL certificate issues.

## Changes
- Created new utility `event_matcher.py` to centralize event matching logic
- Supports both string (EventBridge) and dict (ESM/Pipes) inputs
- Maintains Java engine as default but allows configuration via EVENT_RULE_ENGINE
- Updated references in:
  - events/provider.py
  - events/v1/provider.py
  - lambda_/event_source_mapping/pollers/poller.py
- Added unit test

## Testing
1. Run the new test suite:
```bash
pytest tests/unit/utils/test_event_matcher.py

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -92,7 +92,6 @@ from localstack.aws.api.events import Replay as ApiTypeReplay
 from localstack.aws.api.events import Rule as ApiTypeRule
 from localstack.services.events.archive import ArchiveService, ArchiveServiceDict
 from localstack.services.events.event_bus import EventBusService, EventBusServiceDict
-from localstack.utils.event_matcher import matches_event as matches_rule
 from localstack.services.events.models import (
     Archive,
     ArchiveDict,
@@ -132,6 +131,7 @@ from localstack.services.events.utils import (
 )
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.common import truncate
+from localstack.utils.event_matcher import matches_event as matches_rule
 from localstack.utils.strings import long_uid
 from localstack.utils.time import TIMESTAMP_FORMAT_TZ, timestamp
 
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -131,7 +131,7 @@ from localstack.services.events.utils import (
 )
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.utils.common import truncate
-from localstack.utils.event_matcher import matches_event as matches_rule
+from localstack.utils.event_matcher import matches_event
 from localstack.utils.strings import long_uid
 from localstack.utils.time import TIMESTAMP_FORMAT_TZ, timestamp
 
@@ -489,7 +489,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
         """
         try:
-            result = matches_rule(event, event_pattern)
+            result = matches_event(event_pattern, event)
         except InternalInvalidEventPatternException as e:
             raise InvalidEventPatternException(e.message) from e
 
@@ -1396,7 +1396,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
     ) -> None:
         event_pattern = rule.event_pattern
         event_str = to_json_str(event_formatted)
-        if matches_rule(event_str, event_pattern):
+        if matches_event(event_pattern, event_str):
             if not rule.targets:
                 LOG.info(
                     json.dumps(
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -92,7 +92,7 @@ from localstack.aws.api.events import Replay as ApiTypeReplay
 from localstack.aws.api.events import Rule as ApiTypeRule
 from localstack.services.events.archive import ArchiveService, ArchiveServiceDict
 from localstack.services.events.event_bus import EventBusService, EventBusServiceDict
-from localstack.services.events.event_ruler import matches_rule
+from localstack.utils.event_matcher import matches_event as matches_rule
 from localstack.services.events.models import (
     Archive,
     ArchiveDict,
--- a/localstack-core/localstack/services/lambda_/event_source_listeners/utils.py
+++ b/localstack-core/localstack/services/lambda_/event_source_listeners/utils.py
@@ -1,0 +1,236 @@
+import json
+import logging
+import re
+
+from localstack import config
+from localstack.aws.api.lambda_ import FilterCriteria
+from localstack.utils.event_matcher import matches_event
+from localstack.utils.strings import first_char_to_lower
+
+LOG = logging.getLogger(__name__)
+
+
+class InvalidEventPatternException(Exception):
+    reason: str
+
+    def __init__(self, reason=None, message=None) -> None:
+        self.reason = reason
+        self.message = message or f"Event pattern is not valid. Reason: {reason}"
+
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: localstack-core/localstack/utils/event_matcher.py [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+        # ESM/Pipes (dict-based):
+        >>> pattern = {"source": ["aws.ec2"]}
+        >>> event = {"source": "aws.ec2"}
+    """
+    if config.EVENT_RULE_ENGINE == "java":
+        # If inputs are already strings (EventBridge), use directly
+        if isinstance(event, str) and isinstance(event_pattern, str):
+            return matches_rule(event, event_pattern)
+        # Convert dicts (ESM/Pipes) to strings for Java engine
+        event_str = event if isinstance(event, str) else json.dumps(event)
```
joe4dev, line 37: In the long run, we could consider unifying the usages in ESM and Pipes to use strings. Hence, unifying everything in a single helper makes sense here 👍

localstack
joe4dev, line 1: Adding `localstack` as a submodule here might be accidental?

localstack-core/localstack/utils/event_matcher.py
```diff
@@ -0,0 +1,44 @@
+# localstack/utils/event_matcher.py
```
joe4dev, line 1: nit: file path comment seems redundant

localstack-core/localstack/utils/event_matcher.py
```diff
@@ -0,0 +1,44 @@
...
+import json
+from typing import Any, Dict, Union
+from localstack import config
+from localstack.services.events.event_ruler import matches_rule
+from localstack.services.events.v1.utils import matches_event as python_matches_event
```
joe4dev, line 6: Minor: We could consider grouping the Java and Python implementations because they are now spread across three different directories.

joe4dev (commented):
Nice centralization and addition of unit tests (which we probably under-utilize) 👍 

Feel free to add the change type label and self-assign the PR.

Can we trigger the CI run here by authenticating with CircleCI?

localstack-core/localstack/utils/event_matcher.py
```diff
@@ -0,0 +1,44 @@
...
+    Examples:
+        # EventBridge (string-based):
+        >>> pattern = '{"source": ["aws.ec2"]}'
+        >>> event = '{"source": "aws.ec2"}'
+        
+        # ESM/Pipes (dict-based):
+        >>> pattern = {"source": ["aws.ec2"]}
+        >>> event = {"source": "aws.ec2"}
+    """
+    if config.EVENT_RULE_ENGINE == "java":
```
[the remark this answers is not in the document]
zaingz: think the current approach (Python as default) might be more appropriate for LocalStack because:
- It reduces dependencies for basic usage
- Makes debugging easier
- Follows the principle of least surprise (Python users expect Python implementation by default)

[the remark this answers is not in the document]
joe4dev: I agree with all of these arguments.

The problem is that the current python implementation fails 19/64 test cases and causes many issues and surprises for users (based on support requests and GH issues). From a customer's perspective, we tend to favor AWS parity over performance here. Therefore, we have this open PR intending to switch the defaults LocalStack-wide https://github.com/localstack/localstack/pull/11807
An overview testing table is in our internal backlog item, "Switch default for event ruler to Java."

What do you think?

localstack-core/localstack/services/events/provider.py
```diff
@@ -1440,7 +1440,7 @@ def _process_rules(
             LOG.info(
                 json.dumps(
                     {
-                        "InfoCode": "InternalInfoEvents at matches_rule",
+                        "InfoCode": "InternalInfoEvents at matches_event",
```
joe4dev, line 1443: To avoid regressions in the EventStudio project, I think we should avoid this change.
Ultimately, there should be comments for every instrumentation and/or a list of instrumented places because it's currently impossible to keep track of them.

/cc @maxhoheiser @dominikschubert

localstack-core/localstack/utils/event_matcher.py
```diff
@@ -0,0 +1,53 @@
...
+
+    References:
+        - EventBridge Patterns: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
+        - EventBridge Pipes: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-filtering.html
+        - Event Source Mappings: https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html
+    """
+    if not event_pattern:
+        return True
+
+    if config.EVENT_RULE_ENGINE == "python":
```
joe4dev, line 41: We noticed some bigger issues related to the Java-based implementation concerning the LocalStack shutdown behavior and decided it's better to switch the default to Python here. Hence, I would suggest to flip the conditional back to your original version.

I'm sorry for this back and forth 😢

joe4dev (commented):
Could you please rebase to resolve the conflicts with master? Thank you 🙏

tests/aws/services/events/test_archive_and_replay.py
```diff
@@ -362,6 +363,10 @@ def test_delete_archive_error_unknown_archive(self, aws_client, snapshot):
 class TestReplay:
     @markers.aws.validated
     @pytest.mark.skipif(is_old_provider(), reason="not supported by the old provider")
+    @pytest.mark.skipif(
+        condition=is_v2_provider(),
```
joe4dev, line 367: This condition should be on the `EVENT_RULER` config because now the test gets always skipped

localstack-core/localstack/services/lambda_/event_source_listeners/utils.py
joe4dev, line 1: Can we move `utils.py` to the `event_source_mapping` directory? I think it's cleaner than re-introducing the legacy directory

joe4dev (commented):
LGTM after moving the restored utils file and fixing the test skip conditional 👏 

Let's wait for a full 🟢 test run of localstack-ext (see link in Notion ticket), then we're good to merge 🚀

tests/aws/services/events/test_archive_and_replay.py
joe4dev: ideally, this should be:

```
    @pytest.mark.skipif(
        condition=config.EVENT_RULE_ENGINE == "python",
        reason="Not supported with Python-based rule engine",
    )
```

tests/unit/utils/test_event_matcher.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _set_engine(engine: str):
+        monkeypatch.setattr(config, "EVENT_RULE_ENGINE", engine)
+
+    return _set_engine
+
+
+def test_matches_event_with_java_engine_strings(event_rule_engine):
+    """Test Java engine with string inputs (EventBridge case)"""
+    event_rule_engine("java")
```
joe4dev, line 33: @dominikschubert @dfangl Should we avoid any JPype-starting monkey-patched tests in the LS CI?
I'm somewhat worried this will blow up once we combine it with other usage (e.g., SFN tests)?

This shouldn't block the merge here. I'm happy to tackle this in a follow-up because we have several other usages.

tests/aws/services/events/test_archive_and_replay.py
joe4dev: (the reason currently contradicts the condition ;)

joe4dev (approved):
LGTM 🚀 

Ideally, fixing the test skip could clarify the motivation why we had to do so. At least, doing so in a follow up would be nice :)

PR #11828, opened 2024-11-11
Cloudwatch: add basic composite alarm support

tiurin:
## Motivation

Add basic support for [composite alarms](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Create_Composite_Alarm.html) evaluation in Cloudwatch. Currently composite alarms can be saved but there is no evaluation for them, therefore there is no way to get them triggered.

## Changes

Composite alarm rules will be evaluated now whenever a metric alarm state is changed.

This PR adds limited support for alarm rules:
- Alarms can be referred to by ARNs only (no alarm names support yet)
- It is assumed that only OR logic is applied for all child alarms
- Suppression logic for child alarms is not implemented
- Composite alarm actions only support a message to an SNS topic
- Only metric alarms can be child alarms, nesting of a composite ones is not supported yet

Example of supported rule: `ALARM("arn:<partition>:cloudwatch:<region>:111111111111:alarm:<simple-alarm-1-name>") OR ALARM("arn:<partition>:cloudwatch:<region>:111111111111:alarm:<simple-alarm-2-name>")`

## Testing
Snapshot tests have been recorded for several test cases to verify that alarm rule logic applies correctly.

### DIFF
--- a/localstack-core/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack-core/localstack/services/cloudwatch/provider_v2.py
@@ -845,9 +845,9 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
         alarms = list(store.alarms.values())
         composite_alarms = [a for a in alarms if isinstance(a, LocalStackCompositeAlarm)]
         for composite_alarm in composite_alarms:
-            self._evaluate_composite_alarm(composite_alarm, context, triggering_alarm)
+            self._evaluate_composite_alarm(context, composite_alarm, triggering_alarm)
 
-    def _evaluate_composite_alarm(self, composite_alarm, context, triggering_alarm):
+    def _evaluate_composite_alarm(self, context, composite_alarm, triggering_alarm):
         store = self.get_store(context.account_id, context.region)
         new_state_value = StateValue.OK
         alarm_rule = composite_alarm.alarm["AlarmRule"]
@@ -891,11 +891,14 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
         self._update_state(
             context, composite_alarm, new_state_value, state_reason, state_reason_data
         )
-        if not composite_alarm.alarm["ActionsEnabled"]:
-            return
-        if new_state_value == "OK":
+        if composite_alarm.alarm["ActionsEnabled"]:
+            self._run_composite_alarm_actions(context, composite_alarm, old_state_value, triggering_alarm)
+
+    def _run_composite_alarm_actions(self, context, composite_alarm, old_state_value, triggering_alarm):
+        new_state_value = composite_alarm.alarm["StateValue"]
+        if new_state_value == StateValue.OK:
             actions = composite_alarm.alarm["OKActions"]
-        elif new_state_value == "ALARM":
+        elif new_state_value == StateValue.ALARM:
             actions = composite_alarm.alarm["AlarmActions"]
         else:
             actions = composite_alarm.alarm["InsufficientDataActions"]
--- a/localstack-core/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack-core/localstack/services/cloudwatch/provider_v2.py
@@ -892,9 +892,13 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
             context, composite_alarm, new_state_value, state_reason, state_reason_data
         )
         if composite_alarm.alarm["ActionsEnabled"]:
-            self._run_composite_alarm_actions(context, composite_alarm, old_state_value, triggering_alarm)
+            self._run_composite_alarm_actions(
+                context, composite_alarm, old_state_value, triggering_alarm
+            )
 
-    def _run_composite_alarm_actions(self, context, composite_alarm, old_state_value, triggering_alarm):
+    def _run_composite_alarm_actions(
+        self, context, composite_alarm, old_state_value, triggering_alarm
+    ):
         new_state_value = composite_alarm.alarm["StateValue"]
         if new_state_value == StateValue.OK:
             actions = composite_alarm.alarm["OKActions"]
--- a/localstack-core/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack-core/localstack/services/cloudwatch/provider_v2.py
@@ -781,6 +781,7 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
         old_state_reason = alarm.alarm["StateReason"]
         store = self.get_store(context.account_id, context.region)
         current_time = datetime.datetime.now()
+        # version is not present in state reason data for composite alarm, hence the check
         if state_reason_data and isinstance(alarm, LocalStackMetricAlarm):
             state_reason_data["version"] = HISTORY_VERSION
         history_data = {
--- a/localstack-core/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack-core/localstack/services/cloudwatch/provider_v2.py
@@ -845,78 +845,78 @@ class CloudwatchProvider(CloudwatchApi, ServiceLifecycleHook):
         alarms = list(store.alarms.values())
         composite_alarms = [a for a in alarms if isinstance(a, LocalStackCompositeAlarm)]
         for composite_alarm in composite_alarms:
-            new_state_value = StateValue.OK
-            alarm_rule = composite_alarm.alarm["AlarmRule"]
-            # assuming that a rule consists only of ALARM evaluations of metric alarms, with OR logic applied
-            metric_alarm_arns = re.findall(
-                r'\("([^"]*)"\)', alarm_rule
-            )  # regexp for everything within (" ")
-            for metric_alarm_arn in metric_alarm_arns:
-                metric_alarm = store.alarms.get(metric_alarm_arn)
-                if metric_alarm.alarm["StateValue"] == StateValue.ALARM:
-                    triggering_alarm = metric_alarm
-                    new_state_value = StateValue.ALARM
-                    break
+            self._evaluate_composite_alarm(composite_alarm, context, triggering_alarm)
 
-            old_state_value = composite_alarm.alarm["StateValue"]
-
-            if old_state_value == new_state_value:
-                return
-
-            triggering_alarm_arn = triggering_alarm.alarm.get("AlarmArn")
-            triggering_alarm_state = triggering_alarm.alarm.get("StateValue")
-            triggering_alarm_state_change_timestamp = triggering_alarm.alarm.get(
-                "StateTransitionedTimestamp"
-            )
-            state_reason_formatted_timestamp = triggering_alarm_state_change_timestamp.strftime(
-                "%A %d %B, %Y %H:%M:%S %Z"
-            )
-            state_reason = (
-                f"{triggering_alarm_arn} "
-                f"transitioned to {triggering_alarm_state} "
-                f"at {state_reason_formatted_timestamp}"
-            )
-            state_reason_data = {
-                "triggeringAlarms": [
-                    {
-                        "arn": triggering_alarm_arn,
-                        "state": {
-                            "value": triggering_alarm_state,
-                            "timestamp": timestamp_millis(triggering_alarm_state_change_timestamp),
-                        },
-                    }
-                ]
-            }
-            self._update_state(
-                context, composite_alarm, new_state_value, state_reason, state_reason_data
-            )
-
-            if not composite_alarm.alarm["ActionsEnabled"]:
-                return
-            if new_state_value == "OK":
-                actions = composite_alarm.alarm["OKActions"]
-            elif new_state_value == "ALARM":
-                actions = composite_alarm.alarm["AlarmActions"]
+    def _evaluate_composite_alarm(self, composite_alarm, context, triggering_alarm):
+        store = self.get_store(context.account_id, context.region)
+        new_state_value = StateValue.OK
+        alarm_rule = composite_alarm.alarm["AlarmRule"]
+        # assuming that a rule consists only of ALARM evaluations of metric alarms, with OR logic applied
+        metric_alarm_arns = re.findall(
+            r'\("([^"]*)"\)', alarm_rule
+        )  # regexp for everything within (" ")
+        for metric_alarm_arn in metric_alarm_arns:
+            metric_alarm = store.alarms.get(metric_alarm_arn)
+            if metric_alarm.alarm["StateValue"] == StateValue.ALARM:
+                triggering_alarm = metric_alarm
+                new_state_value = StateValue.ALARM
+                break
+        old_state_value = composite_alarm.alarm["StateValue"]
+        if old_state_value == new_state_value:
+            return
+        triggering_alarm_arn = triggering_alarm.alarm.get("AlarmArn")
+        triggering_alarm_state = triggering_alarm.alarm.get("StateValue")
+        triggering_alarm_state_change_timestamp = triggering_alarm.alarm.get(
+            "StateTransitionedTimestamp"
+        )
+        state_reason_formatted_timestamp = triggering_alarm_state_change_timestamp.strftime(
+            "%A %d %B, %Y %H:%M:%S %Z"
+        )
+        state_reason = (
+            f"{triggering_alarm_arn} "
+            f"transitioned to {triggering_alarm_state} "
+            f"at {state_reason_formatted_timestamp}"
+        )
+        state_reason_data = {
+            "triggeringAlarms": [
+                {
+                    "arn": triggering_alarm_arn,
+                    "state": {
+                        "value": triggering_alarm_state,
+                        "timestamp": timestamp_millis(triggering_alarm_state_change_timestamp),
+                    },
+                }
+            ]
+        }
+        self._update_state(
+            context, composite_alarm, new_state_value, state_reason, state_reason_data
+        )
+        if not composite_alarm.alarm["ActionsEnabled"]:
+            return
+        if new_state_value == "OK":
+            actions = composite_alarm.alarm["OKActions"]
+        elif new_state_value == "ALARM":
+            actions = composite_alarm.alarm["AlarmActions"]
+        else:
+            actions = composite_alarm.alarm["InsufficientDataActions"]
+        for action in actions:
+            data = arns.parse_arn(action)
+            if data["service"] == "sns":
+                service = connect_to(
+                    region_name=data["region"], aws_access_key_id=data["account"]
+                ).sns
+                subject = f"""{new_state_value}: "{composite_alarm.alarm["AlarmName"]}" in {context.region}"""
+                message = create_message_response_update_composite_alarm_state_sns(
+                    composite_alarm, triggering_alarm, old_state_value
+                )
+                service.publish(TopicArn=action, Subject=subject, Message=message)
             else:
-                actions = composite_alarm.alarm["InsufficientDataActions"]
-            for action in actions:
-                data = arns.parse_arn(action)
-                if data["service"] == "sns":
-                    service = connect_to(
-                        region_name=data["region"], aws_access_key_id=data["account"]
-                    ).sns
-                    subject = f"""{new_state_value}: "{composite_alarm.alarm["AlarmName"]}" in {context.region}"""
-                    message = create_message_response_update_composite_alarm_state_sns(
-                        composite_alarm, triggering_alarm, old_state_value
-                    )
-                    service.publish(TopicArn=action, Subject=subject, Message=message)
-                else:
-                    # TODO: support other actions
-                    LOG.warning(
-                        "Action for service %s not implemented, action '%s' will not be triggered.",
-                        data["service"],
-                        action,
-                    )
+                # TODO: support other actions
+                LOG.warning(
+                    "Action for service %s not implemented, action '%s' will not be triggered.",
+                    data["service"],
+                    action,
+                )
 
 
 def create_metric_data_query_from_alarm(alarm: LocalStackMetricAlarm):
--- a/localstack-core/localstack/services/cloudwatch/provider_v2.py
+++ b/localstack-core/localstack/services/cloudwatch/provider_v2.py
@@ -81,6 +81,7 @@ from localstack.services.cloudwatch.models import (
     LocalStackAlarm,
     LocalStackDashboard,
     LocalStackMetricAlarm,
+    LocalStackCompositeAlarm,
     cloudwatch_stores,
 )
 from localstack.services.edge import ROUTER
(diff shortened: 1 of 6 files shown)


tiurin:

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -835,6 +831,84 @@ def _get_timestamp(input: dict):
             history = [h for h in history if (date := _get_timestamp(h)) and date <= end_date]
         return DescribeAlarmHistoryOutput(AlarmHistoryItems=history)
 
+    def _evaluate_composite_alarms(self, context: RequestContext, triggering_alarm):
+        store = self.get_store(context.account_id, context.region)
+        alarms = list(store.alarms.values())
+        composite_alarms = [a for a in alarms if isinstance(a, LocalStackCompositeAlarm)]
+        for composite_alarm in composite_alarms:
```
tiurin, line 838: I haven't added any parent-child relationship between alarms for the first iteration. Therefore, all composite alarms needs to be evaluated if any metric alarm changes its state, regardless of whether that metric alarm is a part of a composite one. It should be ok for the Localstack use case, usually there aren't many alarms running locally so evaluation is not a heavy process. However, it puts some limitations like not being able to include a composite alarm into another composite alarm. Should something similar be needed in the future it is likely a parent-child relationship should be implemented.

tiurin:
Haven't added PR labels as I seem to not have permissions.

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -835,6 +831,84 @@ def _get_timestamp(input: dict):
             history = [h for h in history if (date := _get_timestamp(h)) and date <= end_date]
         return DescribeAlarmHistoryOutput(AlarmHistoryItems=history)
 
+    def _evaluate_composite_alarms(self, context: RequestContext, triggering_alarm):
```
tiurin, line 834: I had some doubts about the length of this function and its effects on readability. From one side in this regard it is similar to `set_alarm_state` one for metric alarms. From the other side it is tempting to extract some smaller functions, for example for evaluating each individual composite alarm (the contents of outer for loop), or processing alarm actions. That would also mean propagating parameters through the call stack, which may hurt maintainability in the future.

What is localstack project's preference in this regard?

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -835,6 +831,84 @@ def _get_timestamp(input: dict):
...
+        store = self.get_store(context.account_id, context.region)
+        alarms = list(store.alarms.values())
+        composite_alarms = [a for a in alarms if isinstance(a, LocalStackCompositeAlarm)]
+        for composite_alarm in composite_alarms:
+            new_state_value = StateValue.OK
+            alarm_rule = composite_alarm.alarm["AlarmRule"]
+            # assuming that a rule consists only of ALARM evaluations of metric alarms, with OR logic applied
+            metric_alarm_arns = re.findall(
```
tiurin, line 842: Alarm rule parsing and further evaluation definitely deserves more checks, like check if parsed metric alarm ARM exists in the store and actually check for `OR` presence. However, I wanted to first get overall feedback on PR and approach and then work on these details.

tiurin: So there are some things that can break in real usage if unchecked, but I'd like to fix them after a general feedback on approach.

localstack-core/localstack/services/cloudwatch/provider_v2.py
tiurin: After a fresh look I extracted simpler methods as the original one was too error-prone, with several return statements that already brought an unwanted behavior.

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -445,21 +449,13 @@ def put_metric_alarm(self, context: RequestContext, request: PutMetricAlarmInput
...
-
-        LOG.warning(
-            "Composite Alarms configuration is not yet supported, alarm state will not be evaluated"
-        )
+        with _STORE_LOCK:
+            store = self.get_store(context.account_id, context.region)
+            composite_alarm = LocalStackCompositeAlarm(
+                context.account_id, context.region, {**request}
+            )
+            alarm_arn = composite_alarm.alarm["AlarmArn"]
```
tiurin, line 457: TODO: input validation, need to check whether alarms referenced in the rule exist in the store.

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+        )
+        composite_alarm_not_changed_by_second_trigger = composite_alarms_list["CompositeAlarms"][0]
+        # It is not possible to reuse the snapshot within the same test,
+        # therefore checking that alarm reason hasn't changed instead of recording new snapshot
+        state_reason_data = json.loads(
+            composite_alarm_not_changed_by_second_trigger["StateReasonData"]
+        )
+        triggering_alarms = state_reason_data["triggeringAlarms"]
+        assert len(triggering_alarms) == 1
+        assert triggering_alarms[0]["arn"] == alarm_2_arn
```
simonrw, line 1232: q: why do you think it's a problem to record new state?

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -835,6 +831,91 @@ def _get_timestamp(input: dict):
...
+        }
+        self._update_state(
+            context, composite_alarm, new_state_value, state_reason, state_reason_data
+        )
+        if composite_alarm.alarm["ActionsEnabled"]:
+            self._run_composite_alarm_actions(
+                context, composite_alarm, old_state_value, triggering_alarm
+            )
+
+    def _run_composite_alarm_actions(
```
simonrw, line 909: There is some duplication between this method and [the existing step in the `set_alarm_state` method](https://github.com/localstack/localstack/blob/385ed1508cfa91b6cc4859f89b57ec7385cff765/localstack-core/localstack/services/cloudwatch/provider_v2.py#L359). Do you think it's worth combining the implementations?

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -889,7 +970,7 @@ def create_message_response_update_state_lambda(
     return json.dumps(response, cls=JSONEncoder)
 
 
-def create_message_response_update_state_sns(alarm, old_state):
+def create_message_response_update_state_sns(alarm: LocalStackMetricAlarm, old_state):
```
simonrw, line 992: Good stuff, typing is great!

It looks like these states are `StateValue` instances

```suggestion
def create_message_response_update_state_sns(alarm: LocalStackMetricAlarm, old_state: StateValue):
```

simonrw (commented):
I have a couple of questions about the correct behaviour, but this is looking great!

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -454,8 +454,18 @@ def put_composite_alarm(self, context: RequestContext, request: PutCompositeAlar
...
             )
+
+            alarm_rule = composite_alarm.alarm["AlarmRule"]
+
+            alarms_from_rule = [alarm.strip() for alarm in alarm_rule.split("OR")]
+            for alarm in alarms_from_rule:
+                if not alarm.startswith("ALARM"):
+                    raise ValidationError(
+                        f"Error in alarm rule condition '{alarm}': Only ALARM expression is supported"
+                    )
```
simonrw, line 465: I think we should perhaps `LOG.warning` and ignore the rule rather than raising an error on unsupported actions. This becomes a known lack of parity rather than a user error.

... _except_ if this is actually a parity improvement with AWS, in which case we should have a snapshot to cover it. I think this would be better as an additional test to capture any validation errors thrown by CloudWatch.

localstack-core/localstack/services/cloudwatch/provider_v2.py
tiurin: Added several input validations in `put_composite_alarm`, namely:
- if each metric alarm condition starts with `ALARM`
- if there is a metric alarm in the store that corresponds to a referenced one

Likely some edge cases are still left out. For a more robust and extensible parsing likely a different approach needs to be taken, probably with recursive descent parsing.

localstack-core/localstack/services/cloudwatch/provider_v2.py
tiurin: added, see https://github.com/localstack/localstack/pull/11828#discussion_r1838458795

pinzon (changes requested):
This is a really good PR and now we're closer to provide full functionality to CW logs. Please resolve the comments from @simonrw  they're very insightful.

localstack-core/localstack/services/cloudwatch/provider_v2.py
```diff
@@ -344,6 +346,8 @@ def set_alarm_state(
                 state_reason_data,
             )
 
+            self._evaluate_composite_alarms(context, alarm)
```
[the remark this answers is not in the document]
tiurin: When child alarm changes state as a result of calculation in a scheduled run then this very `set_alarm_state` is [invoked](https://github.com/tiurin/localstack/blob/f2d214a038ff1c5de9c4e7afd27822f328cfd656/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py#L275). Therefore, all metric alarm state changes, manual or scheduled, go through `set_alarm_state` and would trigger composite alarm re-evaluation.

localstack-core/localstack/services/cloudwatch/provider_v2.py
tiurin: I did take inspiration from AWS response for `put-composite-alarm` when the alarm rule grammar has unsupported tokens.
For example, if I run the following command against AWS

```aws cloudwatch put-composite-alarm --alarm-name "composite-alarm" --alarm-rule "ALARRRRM(metric-alarm)"```

then the response is:

```An error occurred (ValidationError) when calling the PutCompositeAlarm operation: Error in AlarmRule [Unsupported token 'ALARRRRM' at char 0]```

I'm not sure if to qualify the validation I added as a parity improvement though. My motivation to add this validation was to make it very explicit to a Localstack user that as of now no expressions except `ALARM` one are supported. This way when they try to add a composite alarm that has expressions that are not supported yet by Localstack they will get a message and error response quite similar to what AWS provides. If there is no such validation a user could add a composite alarm with what is supported by AWS, let's say, `OK(metric-alarm)` in the rule. This rule would be saved and later be evaluated not as user expects, because in the current implementation every metric alarm is checked for ALARM state only.

So what I added is a _similar_ behaviour, not full parity.

Because of that, I didn't see how I can do a shapshot, as AWS would happily accept the `ALARM` token.

I think when a full alarm rule grammar is implemented many things will become simpler and cleaner, validation included. I just didn't venture into implementing full grammar parser in the first iteration, which lead to having to make several workarounds.

After my explanation, do you think it is worth to keep this validation? I'm absolutely good to change to `LOG.warning` as you suggest if that's the go-to approach in Localstack for lack of parity.

File: localstack-core/localstack/services/cloudwatch/provider_v2.py [hunk 10]
```diff
@@ -835,6 +831,91 @@ def _get_timestamp(input: dict):
             history = [h for h in history if (date := _get_timestamp(h)) and date <= end_date]
         return DescribeAlarmHistoryOutput(AlarmHistoryItems=history)
 
+    def _evaluate_composite_alarms(self, context: RequestContext, triggering_alarm):
+        store = self.get_store(context.account_id, context.region)
```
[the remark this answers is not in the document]
tiurin: `_evaluate_composite_alarms` is only called from `set_alarm_state`, which is locked by `STORE_LOCK`. So technically if a thread runs `set_alarm_state` another thread that wants to run same method will wait until lock is released, and `_evaluate_composite_alarms` invocations should not get into race conditions. In this regard it is similar to the already existing `_update_state` method that operates on store without explicitly acquiring lock.

That said, with `_STORE_LOCK` being a re-entrant lock, acquiring it again within the same thread won't hurt. And with more methods and increasing complexity of call stack combinations (for example, `_update_state` is now called first for metric alarm and then again for composite alarm) it may be better to lock every method that works with store with `_STORE_LOCK` to avoid race conditions in the future as cloudwatch provider parity improves.

So while technically not required here, I think it might be a good time to add those locks to all methods, including `_update_state`, for future maintainability purpose. What do you think?

localstack-core/localstack/services/cloudwatch/provider_v2.py
tiurin: Good catch, will add it here as per suggestion and to a new `create_message_response_update_composite_alarm_state_sns` one as well. 👍

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+
+        composite_alarm_rule = f'ALARM("{alarm_1_arn}") OR ALARM("{alarm_2_arn}")'
+
+        aws_client.cloudwatch.put_composite_alarm(
+            AlarmName=composite_alarm_name,
+            AlarmDescription=composite_alarm_description,
+            AlarmRule=composite_alarm_rule,
+            OKActions=[topic_arn_ok],
+            AlarmActions=[topic_arn_alarm],
+        )
```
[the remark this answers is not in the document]
tiurin: 👍 Makes sense! Adding it in a separate commit, together with the updated snapshot file.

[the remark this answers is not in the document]
tiurin: https://github.com/localstack/localstack/pull/11828/commits/3988f136b4741d640720a3e8a37898cd7ad4d370

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+            lambda: aws_client.sns.unsubscribe(
+                SubscriptionArn=subscription_alarm["SubscriptionArn"]
+            )
+        )
+        subscription_ok = aws_client.sns.subscribe(
+            TopicArn=topic_arn_ok, Protocol="sqs", Endpoint=arn_queue_ok
+        )
+        cleanups.append(
+            lambda: aws_client.sns.unsubscribe(SubscriptionArn=subscription_ok["SubscriptionArn"])
+        )
```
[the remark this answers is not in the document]
tiurin: Great suggestion, added a TODO comment for now but definitely will address either in this or follow-up PR.

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+        composite_alarm_rule = f'ALARM("{alarm_1_arn}") OR ALARM("{alarm_2_arn}")'
+
+        aws_client.cloudwatch.put_composite_alarm(
+            AlarmName=composite_alarm_name,
+            AlarmDescription=composite_alarm_description,
+            AlarmRule=composite_alarm_rule,
+            OKActions=[topic_arn_ok],
+            AlarmActions=[topic_arn_alarm],
+        )
+        cleanups.append(
```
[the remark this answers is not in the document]
tiurin: Great suggestion, added a TODO comment for now but definitely will address either in this or follow-up PR.

localstack-core/localstack/services/cloudwatch/provider_v2.py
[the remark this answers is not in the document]
simonrw: I can see that is what LocalStack currently does, but what does AWS do? Is there a way you can test this?

localstack-core/localstack/services/cloudwatch/provider_v2.py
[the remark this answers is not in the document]
simonrw: I suppose another option would be to pass the store into the `_evaluate_composite_alarms` method, since the lock is held at the call site (https://github.com/localstack/localstack/blob/1b3b2eb835484d47dbe729480e702284aa3daae4/localstack-core/localstack/services/cloudwatch/provider_v2.py#L316). I think re-locking is a fine thing to do, as you say the lock is re-entrant.

localstack-core/localstack/services/cloudwatch/provider_v2.py
simonrw: I agree with your assessment, I just wanted to raise the question.

localstack-core/localstack/services/cloudwatch/provider_v2.py
simonrw: I agree a line needs to be drawn for this first implementation. My take is: I would prefer to not fail on _valid_ input if it is not supported by LocalStack (yet). I think it is better to not break existing infrastructure (e.g. if a user tries to deploy their app to LocalStack, we should not block them just because we don't support a feature).

The only other option we have is to raise `NotImplementedError` which correctly conveys the message, however we treat that exception specially to return a response similar to "The API action '...' is not supported" which is not quite what we want. The _action_ is supported, just that this specific input is not.

Can we maybe switch to the `LOG` statement in this instance, and as you say: a proper rule parser will likely be a future step and we can properly validate the input at that stage.

tiurin: > I think it is better to not break existing infrastructure (e.g. if a user tries to deploy their app to LocalStack, we should not block them just because we don't support a feature).

Great point, now what you suggest makes a lot of sense to me! Also it made me think whether a partial rule implementation does more good than harm. 😄 I mean situations like when someone has a rule with AND in their infra and Localstack will treat it as if it was OR. Might be a rare case, yet it is probably better to not delay full alarm expression grammar support by a long time after (if) this gets merged.

Anyway, will now change raising error to logging as you suggest, thanks for explaining the details about parity concept!

localstack-core/localstack/services/cloudwatch/provider_v2.py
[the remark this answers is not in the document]
tiurin: Ah yeah, good option as well! One more method parameter vs one more context manager (and more indentation). I will address the rest of comments first and then get back to this one to try both options and see which one looks nicer.

tests/aws/services/cloudwatch/test_cloudwatch.py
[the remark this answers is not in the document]
tiurin: Fixed a test by changing alarm name to alarm ARN in the rule, according to scope, but probably it is better to just add support for alarm-names, especially after what @simonrw explained about parity [here](https://github.com/localstack/localstack/pull/11828#discussion_r1840341165). Using alarm names in rules is more human-readable and is likely used a lot. Will look into adding support for names tomorrow.

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+            lambda: aws_client.cloudwatch.delete_alarms(AlarmNames=[composite_alarm_name])
+        )
+
+        composite_alarms_list = aws_client.cloudwatch.describe_alarms(
+            AlarmNames=[composite_alarm_name], AlarmTypes=["CompositeAlarm"]
+        )
+        composite_alarm = composite_alarms_list["CompositeAlarms"][0]
+        assert composite_alarm["AlarmName"] == composite_alarm_name
+        assert composite_alarm["AlarmRule"] == composite_alarm_rule
+
```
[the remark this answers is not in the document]
tiurin: And for a good reason, adding snapshot here actually highlighted the lack of parity!

I tried to add shapshot but then running tests in Localstack mode against highlighted the following difference.

When composite alarm is created in AWS it has INSUFFICIENT_DATA state, but is immediately evaluated and transitioned to the corresponding calculated state (ALARM or OK), **with both child alarm added as triggering alarms**.
Initially I didn't add alarm evaluation to put_composite_alarm method. As a result, when created in Localstack, a composite alarm would be in INSUFFICIENT_DATA state until any of its child alarms changes state and triggers evaluation.
I tried to call evaluation from put_composite_alarm, but current Localstack alarm evaluation only allows to record one triggering alarm, so only works for state changes as a result of metric alarm change.

As mentioned in other comments, for better parity rule evaluation needs quite serious rewrite, this case included. I will add it as TODO to test.

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -2713,6 +2928,34 @@ def _sqs_messages_snapshot(expected_state, sqs_client, sqs_queue, snapshot, iden
     snapshot.match(f"{identifier}-sqs-msg", found_msg)
 
 
+def check_composite_alarm_message(
```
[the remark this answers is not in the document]
tiurin: I tried to do it but there are many arguments that are unique to each invocation. Could only spare `sqs_client`, which is not much.
However, I think I found a middle ground by [defining two inner functions](https://github.com/localstack/localstack/pull/11828/commits/5f3e279cd48a8287ebdc84877066f8f8762e428e), for OK and ALARM message respectively. These indeed require only 2 unique arguments in each call, the data for triggering alarm. WDYT?

tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -988,6 +988,221 @@ def test_set_alarm(self, sns_create_topic, sqs_create_queue, aws_client, cleanup
...
+
+        composite_alarms_list = aws_client.cloudwatch.describe_alarms(
+            AlarmNames=[composite_alarm_name], AlarmTypes=["CompositeAlarm"]
+        )
+        composite_alarm_in_alarm_when_alarm_2_in_alarm = composite_alarms_list["CompositeAlarms"][0]
+        snapshot.match(
+            "composite-alarm-in-alarm-when-alarm-2-is-in-alarm",
+            composite_alarm_in_alarm_when_alarm_2_in_alarm,
+        )
+
```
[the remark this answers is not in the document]
tiurin: Added a new test case [here](https://github.com/localstack/localstack/pull/11828/commits/5c22f5ee753393e89878aaf04e7f68991146f6c1).

tests/aws/services/cloudwatch/test_cloudwatch.py
tiurin: I thought that because composite alarm is unaffected it should be compared against previous snapshot, but now I can't remember why I thought it is a problem to record a new one. 🙂 🤷 I think I have a better understanding of snapshot testing by now and I changed this check to a new recorded snapshot comparison.

tests/aws/services/cloudwatch/test_cloudwatch.py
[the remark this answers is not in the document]
tiurin: Eventually reverted to the original test and moved validation from put alarm method to alarm evaluation, skipping and logging unsupported rules. This is the least intrusive approach which won't harm users who might have valid composite alarms in their infra - they should be able to deploy on Localstack regardless on whether Localstack can evaluate rules in their alarms or not.

localstack-core/localstack/services/cloudwatch/provider_v2.py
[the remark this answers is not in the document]
tiurin: Leaving this outside the scope of this PR, with [TODO added](https://github.com/tiurin/localstack/blob/6922402a8c8d8db99249522e05d7df778468f8a3/localstack-core/localstack/services/cloudwatch/provider_v2.py#L844).

pinzon (approved):

pinzon:
Hi @tiurin. Thank you for addressing the previous comments. I'm pleased to approve your pull request :heavy_check_mark:.

I noticed that the PR was labeled as a PATCH, but it should be classified as a MINOR since it introduces new functionality to the CloudWatch service. Given our current code freeze, we will merge this PR next week.

tiurin:
Thanks @pinzon and @simonrw, my pleasure to contribute! Yeah, good point on MINOR rather than PATCH. I will rebase my branch one more time later this week to make sure it builds well with all the latest changes to master.

PR #11830, opened 2024-11-12
feat(CloudControl): Introduces list operations for bunch of services so cloudcontrol can pick them up

Pive01:
## Motivation
In order to support the [Resource Overview](https://github.com/localstack/localstack-web/pull/1459/files) project we need some:
- CloudControl apis to list resources
- Endpoint to gather current usage

## Changes
 Introduce CloudFormation list-types operations to list supported types by cloudcontrol

Relative -ext PR https://github.com/localstack/localstack-ext/pull/3699
/cc @simonrw @dominikschubert


Pive01:

localstack-core/localstack/services/apigateway/resource_providers/aws_apigateway_resource.py
```diff
@@ -94,6 +96,30 @@ def read(
         """
         raise NotImplementedError
 
+    def list(
+        self,
+        request: ResourceRequest[ApiGatewayResourceProperties],
+    ) -> ProgressEvent[ApiGatewayResourceProperties]:
+        input_payload = json.loads(request.desired_state)
```
giograno, line 103: @simonrw I was playing with the resource endpoint and I came across this.
This is the only `list` operation so far that requires a desired state to be provided by `cloudcontrol`. When this does not happen, we pass an empty dictionary here, which of course fails to be JSON-loaded. From the rest of the code it looks like we could bound `Properties` to `dict`.

simonrw: What we should do to be in parity here is raise an exception if the required payload is not provided. I get the error message:

```
botocore.errorfactory.InvalidRequestException: An error occurred (InvalidRequestException) when calling the ListResources operation: Missing or invalid ResourceModel property in AWS::ApiGateway::Resource list handler request input.Required property:  (#: required key [RestApiId] not found)
```

I have just pushed a change that should make input handling better.

tests/integration/services/test_internal.py
```diff
@@ -61,3 +61,18 @@ def test_get(self):
         assert doc["machine_id"]
         assert doc["system"]
         assert type(doc["is_license_activated"]) == bool
+
+
+@pytest.mark.usefixtures("openapi_validate")
```
giograno, line 77: Kudos for adding a test 💯 !! This fixture won't have an effect till we define the endpoint spec in `openapi.yml`. I can take care of that in a second iteration!

simonrw (approved):

PR #11035, opened 2024-06-17
make next-gen APIGW provider default

bentsku:
## Motivation
This PR makes the APIGW NextProvider to be the default used. 

This PR also compiles all PRs done related to this work.

## Changes
- make the NextGen provider default

## NG Provider
- #10982
- #11043
- #11051
- #11057
- #11062
- #11065
- #11085
- #11096
- #11099
- #11107
- #11111
- #11114
- #11121
- #11129
- #11135
- #11138 
- #11140
- #11142
- #11143
- #11144
- #11148
- #11167
- #11183
- #11196
- #11198
- #11199
- #11219 
- #11226
- #11227
- #11234
- #11237 
- #11250
- #11251
- #11256
- #11257
- #11268
- #11317
- #11327
- #11344
- #11349
- #11350
- #11372
- #11401
- #11413
- #11518
- #11569
- #11702

## Follow-up?
- #11633

## TODO

What's left to do:
- [ ] move the code from the `next_gen` module back into default? TBD

Once this PR is merged:
- [ ] remove the NextGen CI step

PR #11834, opened 2024-11-12
Make EventBridge v2 provider default

joe4dev:
Depends on https://github.com/localstack/localstack/pull/11852

## Motivation

Upon v4, we plan to switch EventBridge's default provider from the current default v1 using a moto-fallback to the new LocalStack-native v2 provider. The new v2 provider offers better internal API integrations, API coverage, and generally improved AWS parity, especially for behavior.

Why minor?
We tend to ship new providers serving as drop-in replacements as minor and EventBridge v2 can be considered a drop-in replacement (i.e., a better LocalStack-native version of v1 without regressions requiring user changes).

## Changes

* Switch default provider from v1 to v2
* Add `legacy` as a possible fallback config in line with other service deprecations such as APIGatewayv2 (see PR change [here](https://github.com/localstack/localstack/pull/11035/files#diff-f1008346b11ae7b0c73caf77f80ebf3b83d732f1f4e89a8567be5126375304d5R35))

## Testing

`PROVIDER_OVERRIDE_EVENTS=legacy localstack start` should print a deprecation warning.

## Limitations (WIP)

* LS persistence
* [Connections](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-target-connection.html) and [API Destinations](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-api-destinations.html) are CRUD only

### DIFF
--- a/localstack-core/localstack/services/events/target.py
+++ b/localstack-core/localstack/services/events/target.py
@@ -19,6 +19,7 @@ from localstack.services.events.utils import (
     get_trace_header_encoded_region_account,
     to_json_str,
 )
+from localstack.utils import collections
 from localstack.utils.aws.arns import (
     extract_account_id_from_arn,
     extract_region_from_arn,
@@ -438,9 +439,13 @@ class FirehoseTargetSender(TargetSender):
 
 class KinesisTargetSender(TargetSender):
     def send_event(self, event):
-        partition_key_path = self.target["KinesisParameters"]["PartitionKeyPath"]
+        partition_key_path = collections.get_safe(
+            self.target,
+            "$.KinesisParameters.PartitionKeyPath",
+            default_value="$.id",
+        )
         stream_name = self.target["Arn"].split("/")[-1]
-        partition_key = event.get(partition_key_path, event["id"])
+        partition_key = collections.get_safe(event, partition_key_path, event["id"])
         self.client.put_record(
             StreamName=stream_name,
             Data=to_bytes(to_json_str(event)),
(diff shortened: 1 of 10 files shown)


tests/aws/services/events/helper_functions.py
```diff
@@ -7,14 +7,14 @@
 
 
 def is_v2_provider():
-    return os.environ.get("PROVIDER_OVERRIDE_EVENTS") == "v2" and not is_aws_cloud()
+    return (
+        os.environ.get("PROVIDER_OVERRIDE_EVENTS", "") not in ("v1", "legacy")
+        and not is_aws_cloud()
```
dominikschubert, line 12:
```suggestion
        and not is_aws_cloud()
```
nit: not a fan of that combination here

localstack-core/localstack/services/events/target.py
```diff
@@ -434,9 +439,13 @@ def send_event(self, event):
...
     def send_event(self, event):
-        partition_key_path = self.target["KinesisParameters"]["PartitionKeyPath"]
+        partition_key_path = collections.get_safe(
+            self.target,
+            "$.KinesisParameters.PartitionKeyPath",
+            default_value="$.id",
+        )
         stream_name = self.target["Arn"].split("/")[-1]
-        partition_key = event.get(partition_key_path, event["id"])
+        partition_key = collections.get_safe(event, partition_key_path, event["id"])
```
dfangl, line 448: It's fixed :D We should probably add a test in community as well to capture these regressions.

dfangl (approved):
I would have preferred a switch with less commented out checks, but let's see :)

PR #11836, opened 2024-11-12
[WIP] EventBridge Connections tests plus implementation

zaingz:


joe4dev:
Merging this after a tiny rebase (just an import error) because the pipeline was green and we need this missing feature for testing the default switch in https://github.com/localstack/localstack/pull/11834

PR #11837, opened 2024-11-13
Update Node.js release signer keys

viren-nadkarni:
## Background

When building the Docker images, Node.js binaries are downloaded and its signature is verified against [authorised release keys](https://github.com/nodejs/node?tab=readme-ov-file#release-keys).

For the latest 18.x build (Node.js v18.20.5) released on 2024-11-12, the signature verification fails because we use an outdated list of keys. This in turn causes the whole Docker build to [fail](https://app.circleci.com/pipelines/github/localstack/localstack/29407/workflows/ef2568f4-eed6-4151-9e6e-6cc181015e81/jobs/259311).

## Changes

This PR updates the release keys to the latest list.


dominikschubert (approved):
Thanks for jumping on this!

PR #11847, opened 2024-11-14
Add Bedrock configuration values to analytics

silv-io:
## Motivation

Bedrock is an upcoming new pro service which needs tracking in our analytics

## Changes

- add tracking for the pre-warming configuration to check how many people want to have their ollama container started on localstack startup
- add tracking for the default model used in ollama so that we know for which models to optimize our implementations


thrau (approved):
LGTM!

PR #11202, opened 2024-07-14
remove legacy runtime in localstack/service/infra.py

thrau:
## Motivation

This is a PR to remove the code that was deprecated by #10942.

I left `signal_supervisor_restart()` in the legacy module (it's still referenced in ext), but added a TODO.

Considering this a minor change since we never documented `LEGACY_RUNTIME` and have, to my knowledge, not needed it so far.

## Changes

* `infra.py` has been completely removed in favor of the new runtime framework

## Testing

One thing needs manual testing, namely the kill function of the health endpoint
* run `python -m localstack.dev.run -e DEBUG=1`
* observe that `curl -X POST -H 'Content-Type: application/json' localhost.localstack.cloud:4566/_localstack/health -d '{"action": "kill"}'` still works

PR #11848, opened 2024-11-14
fix usage analytics to properly register and not count if disabled

bentsku:
## Motivation
While looking to implement the usage analytics for the API Gateway service, I've noticed three things:
- we were not sending usage analytics anymore since #11202, as we removed the code to send it, but didn't register the hook back
- the usage counters were still counting even though analytics would be disabled, and would only skip sending the data over
- the usage counters are using lists that are growing unbounded, possibly using large amount of memory for very long lasting sessions (we're aware of 20+ hours sessions)

This PR mainly does 2 things:
- re-register the hook so that we send analytics over
- it does not count if analytics is disabled
- refactor the counter a bit to avoid growing a list, instead keeping the "set" entries and counting with `itertools.count` which is thread safe in CPython

I don't really like the boolean check for every call to `increment` and such, we could probably refactor our counters to be no-op classes if the analytics are disabled? But this at least prevents the memory leaks for users with `DISABLE_EVENTS=1` as a quick fix before v4, while re-enabling sending the analytics over. 

As a sanity check, I've run some small benchmarks with `timeit`:
Setting 1 000 000 values in the `state`:
Performance:
- 0.236s for the new counter
- 0.199s for the old counter with `append`

Memory:
- 384 bytes for the new counter
- 8448728 bytes for the old counter with `append`

## Changes
- register `aggregate_and_send ` with `@hooks.on_infra_shutdown(should_load=lambda: not config.DISABLE_EVENTS)`
- don't count if `DISABLE_EVENTS` is true
- small refactor of the `UsageSetCounter`
- small change: added the `count` aggregation method
- fix the `median` calculation, as it would have been incorrect without sorting the list first

## Follow-up

I'm following with #11854 which might contain more controversial changes, related to our counters. I believe the `UsageCounter` could be simplified into a really simple counter for less complex use case with no aggregation (like `lambda:hotreload`), and rename the current counter into a somewhat "timing" related counter with aggregation for more complex use cases. 

## Multithreading testing:

With this small test, we would always end up with the right amount of counting, just for sanity check. 
```python

iter_counter = UsageSetCounterIter("namespace")

def incr_iter(*args, **kwargs):
    iter_counter.record(random.choice(["test", "test2", "test3", "test4"]))

with ThreadPoolExecutor(max_workers=1000) as tp:
    p = tp.map(incr_iter, range(100000))

    for _ in p:
        pass

    print(sum(iter_counter.state.values()))
    > always prints out 100000
```

\cc @cloutierMat thanks for the help and pairing on this! 🙏

### DIFF
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -112,11 +112,14 @@ def aggregate() -> dict:
     return aggregated_payload
 
 
-@hooks.on_infra_shutdown(should_load=lambda: not config.DISABLE_EVENTS)
+@hooks.on_infra_shutdown()
 def aggregate_and_send():
     """
     Aggregates data from all registered usage trackers and immediately sends the aggregated result to the analytics service.
     """
+    if not config.DISABLE_EVENTS:
+        return
+
     metadata = EventMetadata(
         session_id=get_session_id(),
         client_time=str(datetime.datetime.now()),
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -37,7 +37,7 @@ class UsageSetCounter:
     def __init__(self, namespace: str):
         self.enabled = not config.DISABLE_EVENTS
         self.state = {}
-        self._counter = defaultdict(count)
+        self._counter = defaultdict(lambda: count(1))
         self.namespace = namespace
 
     def record(self, value: str):
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -1,8 +1,10 @@
 import datetime
 import math
+from collections import Counter
 from typing import Any
 
 from localstack import config
+from localstack.runtime import hooks
 from localstack.utils.analytics import get_session_id
 from localstack.utils.analytics.events import Event, EventMetadata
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/utils/analytics/usage.py
```diff
@@ -101,13 +112,11 @@ def aggregate() -> dict:
     return aggregated_payload
 
 
+@hooks.on_infra_shutdown(should_load=lambda: not config.DISABLE_EVENTS)
 def aggregate_and_send():
     """
     Aggregates data from all registered usage trackers and immediately sends the aggregated result to the analytics service.
     """
-    if config.DISABLE_EVENTS:
-        return
```
thrau, line 109: wow thanks for adding this, can't believe i missed this :sweat_smile: 

i personally probably would have kept the guard in the method and not use `should_load`, simply because it makes it a bit easier to debug in case something's up. it also allows us to change `config.DISABLE_EVENTS` after the runtime has initiated, though we really don't have a case for that now it would be a bit more consistent with being able to set certain config values at runtime via the config endpoint.

bentsku: I see, yes, I thought about this as well! Will update. 

Only thing is that with the current logic to avoid "counting" if not enabled, even if we update at runtime, we will still avoid counting for now, or I should update the code to use the config variable directly. We can always update later for that part. 👍

PR #11851, opened 2024-11-14
Fix non-validated events test for v2 provider

dominikschubert:
## Motivation

The PR addresses a subset of tests that were failing when testing against the v2 provider.

The main reason here is that they shouldn't even have run. The provider detection in the tests wasn't adapted, so the skip didn't evaluate to true.

The changes here should probably just be integrated into the "make events v2 the default provider" PR?

## Changes

- Fix `is_old_provider` and remove `is_v2_provider`. Keep in mind this is only technically valid when the default in  `providers.py` is already  switched to v2!
- Delete `test_scheduled_expression_events` as there's already other tests covering the same functionality executed for both provider versions.
-  `test_api_destinations`. untouched. can be kept for the v1 provider as a regression test but should be reimplemented for v2 tests
- `test_create_connection_validations` reworked into a validated snapshot test. Can be addressed during the work on the connections, but should be fairly low priority.

### DIFF
--- a/tests/aws/services/events/helper_functions.py
+++ b/tests/aws/services/events/helper_functions.py
@@ -2,19 +2,11 @@ import json
 import os
 from datetime import datetime, timedelta, timezone
 
-from localstack.testing.aws.util import is_aws_cloud
 from localstack.utils.sync import retry
 
 
-def is_v2_provider():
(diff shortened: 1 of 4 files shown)


joe4dev:
Closing this as I cherry-picked and integrated the changes into the default switch PR https://github.com/localstack/localstack/pull/11834
(not ideal for minimal change but it should do the job for quicker integration as suggested here)

PR #11852, opened 2024-11-14
Fix validated events test for v2 provider 

dominikschubert:
## Motivation

Addressing a few issues we observed with the new EventBridge v2 provider and the existing test suite.

## Changes

- `test_cfn_handle_events_rule` ... invalid legacy test => removed
- `test_cfn_handle_events_rule_without_name` ... invalid legacy test => removed
- `test_apigateway_to_eventbridge` ... fixed by not just ignoring the event if there's no rule configured
- validated test added for missing detail type => `test_put_event_without_detail_type`
- `test_eventbus_policy_statement` & `test_eventbus_policies` ... fixed via the `json.dumps` on the policy
- `test_invoke_lambda_eventbridge` ... fixed by using the proper `"detail-type"` for the event sent to EventBridge in Lambda EventBridge destinations. Adapted the behavior test to surface the raw event payload and remove the intermediate SQS queue

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -1986,6 +1986,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
                     rule, region, account_id, event_formatted, processed_entries, failed_entry_count
                 )
         else:
+            processed_entries.append({"EventId": event_formatted["id"]})
             LOG.info(
                 json.dumps(
                     {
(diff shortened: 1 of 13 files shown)


joe4dev:
Rebased and checked that `aws.services.lambda_.test_lambda_destinations.TestLambdaDestinationEventbridge.test_invoke_lambda_eventbridge` works locally, might have been a flake in the x86 CI 🤷‍♂️

dominikschubert:
> Rebased and checked that `aws.services.lambda_.test_lambda_destinations.TestLambdaDestinationEventbridge.test_invoke_lambda_eventbridge` works locally, might have been a flake in the x86 CI 🤷‍♂️

I've separated the success and failure case a bit better now to avoid the timing issue there

joe4dev (approved):
LGTM

PR #11853, opened 2024-11-14
EC2: implement determinstic subnet ID generation

simonrw:
## Motivation

In order to support the AWS replicator use case, we need to be able to generate predictable IDs for subnets.

We follow the pattern from the VPC replication (in #11772).

## Changes

* Add predictable ID generation

## Testing

Tests are implemented in our ext test suite


cloutierMat (approved):
:+1: LGTM

localstack-core/localstack/services/ec2/patches.py
```diff
@@ -30,6 +29,16 @@ def generate_vpc_id(
...
 
 
+@localstack_id
+def generate_subnet_id(
+    resource_identifier: ResourceIdentifier,
+    existing_ids: ExistingIds = None,
+    tags: Tags = None,
+) -> str:
+    # We return an empty string here to differentiate between when a custom ID was used, or when it was randomly generated by `moto`.
+    return ""
```
cloutierMat, line 39: Nit: Since it only returns an empty string and the signature is shared for all generate function, we could probably implement only one with the purpose of returning an empty string.

A single `match_or_blank()` could be reused across all resource with this specific need. (The naming ain't right, but I am too tired to come up with a good name right now! :rofl: )

Not blocking, we can definitely keep on improving with the next ones!

simonrw: I think that's a great idea! Since I will be working on another resource in the same file, I will implement this change there, thanks!

PR #11854, opened 2024-11-15
Split analytics UsageCounter and TimingStats

bentsku:
## Motivation
Follow up from #11848, with more changes to our counters.

This PR implements a simpler `UsageCounter` which does not need aggregation, as it only counts. 

To keep the current already implemented logic, which is more resembling the `timing` metric types of `statsd`, I've renamed it to `TimingStats` (naming TBD, it could also be NumericStats or anything?). 

The new `UsageCounter` is very memory efficient, and it relies on `itertools.count` which is thread safe in CPython, removing the need for an external locking mechanism. 

Statsd metrics types:
https://github.com/statsd/statsd/blob/master/docs/metric_types.md

## Changes
- refactor our counters to have a very simpler counter and more complex one with aggregations for more complex use cases
- modify the usage in lambda so that it uses the simple counter

\cc thanks again to @cloutierMat for the pairing 🙏

### DIFF
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -50,15 +50,47 @@ class UsageSetCounter:
 
 class UsageCounter:
     """
-    Use this counter to count numeric values and perform aggregations
-
-    Available aggregations: min, max, sum, mean, median
+    Use this counter to count numeric values
 
     Example:
-        my_feature_counter = UsageCounter("lambda:somefeature", aggregations=["min", "max", "sum"])
-        my_feature_counter.increment()  # equivalent to my_feature_counter.record_value(1)
-        my_feature_counter.record_value(3)
-        my_feature_counter.aggregate()  # returns {"min": 1, "max": 3, "sum": 4}
+        my__counter = UsageCounter("lambda:somefeature")
+        my_counter.increment()
+        my_counter.increment()
+        my_counter.aggregate()  # returns {"count": 2}
+    """
+
+    state: int
+    namespace: str
+
+    def __init__(self, namespace: str):
+        self.enabled = not config.DISABLE_EVENTS
+        self.state = 0
+        self._counter = count()
+        self.namespace = namespace
+        collector_registry[namespace] = self
+
+    def increment(self):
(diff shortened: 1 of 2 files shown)


thrau (approved):
LGTM, but i'd just like some confirmation from the original authors

bentsku:
@dominikschubert do you think we should go forward with this? I think we should decide on it before everybody starts implementing there, so it'd be easier to update. If you don't think this is necessary, we can also drop it for now and come back to it later.

localstack-core/localstack/utils/analytics/usage.py
```diff
@@ -50,15 +50,47 @@ def aggregate(self) -> dict:
...
+        my__counter = UsageCounter("lambda:somefeature")
+        my_counter.increment()
+        my_counter.increment()
+        my_counter.aggregate()  # returns {"count": 2}
+    """
+
+    state: int
+    namespace: str
+
+    def __init__(self, namespace: str):
```
bentsku, line 65: in order to not introduce a breaking change, we could still add `aggregations` in the signature of the class, even if it does nothing, in order to not break newly introduced usage in upstream dependencies

dominikschubert: Honestly probably not that important. Everyone is adapting the counters right now anyway, I'm fine with leaving it out here

localstack-core/localstack/utils/analytics/usage.py
```diff
@@ -50,15 +50,47 @@ def aggregate(self) -> dict:
...
+        #  when config.DISABLE_EVENTS is set
+        if self.enabled:
+            self.state = next(self._counter)
+
+    def aggregate(self) -> dict:
+        # TODO: should we just keep `count`? "sum" might need to be kept for historical data?
+        return {"count": self.state, "sum": self.state}
+
+
+class TimingStats:
```
dominikschubert, line 83: Not a fan of the name TBH. Not everything collected here has to be a "timing". I like your suggestion of `NumericStats` better :+1:

dominikschubert (approved):
Minor suggestions, but otherwise LGTM. Not terribly concerned about changing this as adapting any existing ones in open PRs should be trivial. :+1: 

Thanks so much for your work on improving these usage counters :rocket:

PR #11855, opened 2024-11-15
User-Agents analytics for developer endpoints

giograno:
## Motivation
As we are introducing SDKs to interact with LocalStack endpoints, we want to gather some analytics about their usage.
My idea was to use one `UsageSetCounter` for each endpoint and count the unique occurrences of the User-Agents.
A namespace would be `internal:<path>` and the record would happen in the open API request validation handler (even if the validation is not active).

Looking for feedback about this approach. For instance, does it make sense to log all agents? Would this generate too much data?

## Changes
- Adding analytics counters for the user agents of internal endpoints.


localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -45,6 +73,7 @@ class OpenAPIRequestValidator(OpenAPIValidator):
     """
 
     def __call__(self, chain: HandlerChain, context: RequestContext, response: Response):
+        self._record_usage(context)
```
bentsku, line 76: As this executed as the first step of the handler, this will be called for every request hitting LocalStack, even if not an internal route. I believe this should be put after the following condition:
`if path.startswith(f"{INTERNAL_RESOURCE_PATH}/") or path.startswith("/_aws/"):`

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -18,16 +19,43 @@
...
     open_apis: list["OpenAPI"]
 
     def __init__(self) -> None:
         self._load_specs()
 
+    def _record_usage(self, context: RequestContext) -> None:
+        if config.DISABLE_EVENTS:
+            return
+
+        request_path = context.request.path
```
bentsku, line 52: I'm a bit worried about the path part, as for some routes, we have dynamic path variables.

And API Gateway is technically part of the internal routes now, even though it serves a pretty different purpose (providing a path variant of the host endpoint) than our other internal routes which provide LocalStack only features. 
Also, API Gateway can have many different parts, which makes it even trickier, we will create a lot, lot of Counters .

bentsku (commented):
Nice, having a global way of doing this would be really great.

I have a few comments regarding adding this to all internal routes without discrimination, as API Gateway would get caught into it as well. Maybe adding it to the `_aws` namespace wasn't such a good idea after all 😕

I believe this PR is counting every requests coming to LocalStack.

localstack-core/localstack/aws/handlers/validation.py
giograno: Absolutely!

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -18,16 +19,52 @@
...
+from localstack.utils.analytics.usage import UsageSetCounter
 
 LOG = logging.getLogger(__name__)
 
 
+class UsageCollectorFactory:
+    _collector_registry: dict[str, Any] = {}
+    """Registry for the different paths."""
+
+    NAMESPACE_PREFIX = "internal"
```
bentsku, line 31: q: I wonder if we should rename this to something like related to user-agent, as it is what it will track in the end? as we might end up with a few entries, like `internal/_aws/sns`, `internal/_localstack/health`, `internal/_aws/iam` etc, etc. But this is gonna give us quite a lot of insights, being able to group them by service/feature, and count them!

So I'm not too sure about the naming, writing it down now looks alright to me 😅 naming is really not my forte

localstack-core/localstack/aws/handlers/validation.py
```diff
@@ -18,16 +19,52 @@
...
+            return
+        # Skip the endpoints for the new API Gateway implementation
+        if "execute-api" in request_path:
+            return
+        # We only record the first segment in the path after the _internal/ or _aws/ prefix, as a path can have
+        # potentially an infinite number of parameters.
+        recorded_path = request_path.split("/")[:2]
+        if len(recorded_path) < 2:
+            return
+        recorded_path = "/".join(recorded_path)
```
bentsku, line 64: I think this is a nice compromise for now, while still giving us a picture of which services/features are used in the internal endpoint but not too much granularity/cardinality (especially with dynamic path parameters).

We can still try to think of way to get the "matched" path, maybe with the help of the OpenAPI specs actually? No idea how we could get the "route" like `/_aws/events/rules/{rule_arn}/trigger` for example.

localstack-core/localstack/aws/handlers/validation.py
bentsku, line 1: Regarding the location of this analytics client, we also discussed if the OpenAPI validation was the best spot, and I don't really know. We could probably add a specific handler just for that in the response handlers, close to the `ServiceRequestCounter`. I'd say for now I'm okay with it as we're in a bit of a rush to get this in, but we could probably think of moving it? thoughts @thrau @dominikschubert ?

bentsku (commented):
This looks good to me! Just the question of the location of the code remains, to me this is not really blocking but should probably be addressed in a follow-up, as we are bit pressed for time. 

I think this solution is good for now, allowing us to still understand usage with not too much granularity. Very good idea to track at a global level 👌

localstack-core/localstack/aws/handlers/validation.py
giograno: Makes sense. I refactored this already and created a separate handler. It's much more clean this way!

localstack-core/localstack/aws/handlers/validation.py
giograno: I renamed the namespace as `agent:<path>`. It seems to be a bit more coherent also with the other namespaces we are adding around.

localstack-core/localstack/aws/handlers/validation.py
giograno: Try to get the matched path would be by far the best option, but given the time constraint, I don't see an easy solution we could adopt right now.

bentsku: Definitely agreed! 👍

bentsku (approved):
LGTM! Thanks for creating a dedicated handler, already living in `analytics.py`, this is clean 👌

thrau:
hey folks! i changed the milestone on this one, as i'd like to discuss this change more, and we won't get this in on time until 4.0. as ben pointed out, there's no real rush. will follow up post-v4 :-)

thrau:
sorry i never followed up again on this :sob:  perhaps we can get @vittoriopolverino to review this?

vittoriopolverino:
Hey @giograno ,
You should use the new counter implementation for these analytics. Please follow the guidelines outlined in the [Counter](https://www.notion.so/localstack/Counter-192fc2a234318097b596c35054bd55ae) and [Metric & Label Best Practices](https://www.notion.so/localstack/Metric-Label-Best-Practices-192fc2a2343180fb9c41d8d4aca279ee).

If you have any questions, let me know and we can sync up 💪🏼

alexrashed:
@giograno This one seems pretty old, but was approved at some point. Should it be closed, or should it be revived? :)

giograno:
I am going to close this and re-evaluate later with the new guidelines.

Issue #11817, opened 2024-11-08
bug: Malformed attributes to SNS publish-batch call produce 500 response


kylelyk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I get an 500 error from localstack when calling SnsAsyncClient.publishBatch on the Java AWS SDK with a malformed query. I had forgot to set the dataType field on the MessageAttributeValue. Real AWS gives back an error stating that the dataType field is required. Local stack returns a 500 with the stacktrace below:

```
Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack-core/localstack/aws/handlers/service.py", line 113, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack-core/localstack/aws/handlers/service.py", line 83, in __call__
    skeleton_response = self.skeleton.invoke(context)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(serializer, context, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 168, in dispatch_request
    result = handler(context, instance) or {}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/forwarder.py", line 133, in _call
    return handler(context, req)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/services/sns/provider.py", line 205, in publish_batch
    total_batch_size += get_total_publish_size(message_payload, message_attributes)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/services/sns/provider.py", line 1065, in get_total_publish_size
[...]

PR #11857, opened 2024-11-15
fix SNS empty MessageAttributes validation

bentsku:
## Motivation
As reported with #11817, we've had an issue in our validation of SNS attributes, with the order of the validation, and that directly accessed the `DataType` field. 

Validation is now updated with more test cases.

## Changes
- add few test cases for Publish and PublishBatch
- fix the behavior and change validation order

_fixes #11817_

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -199,14 +199,15 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
 
         total_batch_size = 0
         message_contexts = []
-        for entry in publish_batch_request_entries:
+        for entry_index, entry in enumerate(publish_batch_request_entries, start=1):
             message_payload = entry.get("Message")
             message_attributes = entry.get("MessageAttributes", {})
-            total_batch_size += get_total_publish_size(message_payload, message_attributes)
             if message_attributes:
                 # if a message contains non-valid message attributes
                 # will fail for the first non-valid message encountered, and raise ParameterValueInvalid
-                validate_message_attributes(message_attributes)
+                validate_message_attributes(message_attributes, position=entry_index)
+
+            total_batch_size += get_total_publish_size(message_payload, message_attributes)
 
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/sns/provider.py
```diff
@@ -943,6 +958,11 @@ def validate_message_attributes(message_attributes: MessageAttributeMap) -> None
             raise InvalidParameterValueException(
                 f"The message attribute '{attr_name}' has an invalid message attribute type, the set of supported type prefixes is Binary, Number, and String."
             )
+        if not any(attr_value.endswith("Value") for attr_value in attr):
+            raise InvalidParameterValueException(
+                f"The message attribute '{attr_name}' must contain non-empty message attribute value for message attribute type '{data_type}'."
+            )
```
cloutierMat, line 964: Question: It appears that only `StringValue` and `BinaryValue` are supported in the message attributes. Is there a reason to allow `<Any>Value`?

cloutierMat (approved):
Good fix and nice upgrade to validation and tests!

I only left one question, not blocking, but curious to know why the validation appears so cryptic! :rofl:

localstack-core/localstack/services/sns/provider.py
bentsku: Good point... I've tried to run a test with `BadValue` but I get a serialization error from the client 😭 
```python
prefix = 'MessageAttributes.entry.1.Value'

    def _serialize_type_structure(self, serialized, value, shape, prefix=''):
        members = shape.members
        for key, value in value.items():
>           member_shape = members[key]
E           KeyError: 'BadValue'
```

I'd need to get the raw request to validate... 

I guess then it will be a matter of raising the right exception. I think if I accept `<Any>Value`, then it will raise `InvalidParameterValueException(f"The message attribute '{attr_name}' with type '{data_type}' must use field '{value_key_data_type}'.")`. Not sure if it's the right behavior but at least we raise something. I guess I won't fight it too much, but it's very good point... 🤔

PR #11859, opened 2024-11-18
Switch Lambda debug path analytics to presense

joe4dev:
## Motivation

Paths might contain sensitive information, so we had better only track its presence.

## Changes

Switch to presence detection rather than tracking the path value.

## Questions

Could such a switch cause data analytics/quality challenges in Tinybird @ackdav ?


pinzon (approved):
LGTM, we just need to make sure @ackdav is aware of the change and reason.

PR #11860, opened 2024-11-18
add analytics to APIGW NextGen REST API handler chain

bentsku:
## Motivation
As part of our effort to improve our analytics, this PR implements usage counters for API Gateway.

We can now properly have usage data on APIGW invocations, with the integration type (`AWS_PROXY`, `MOCK` etc), and in case of the `AWS` integration, the service the integration is targeting (`dynamodb`, `sqs` etc..)

## Changes
- implement APIGW NextGen handler for analytics

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
@@ -1,0 +1,44 @@
+import logging
+
+from localstack.http import Response
+from localstack.utils.analytics.usage import UsageSetCounter
+
+from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
+from ..context import RestApiInvocationContext
+
+LOG = logging.getLogger(__name__)
+
+
+class IntegrationUsageCounter(RestApiGatewayHandler):
+    counter: UsageSetCounter
+
+    def __init__(self, counter: UsageSetCounter = None):
+        self.counter = counter or UsageSetCounter(namespace="apigateway:invokedrest")
+
+    def __call__(
+        self,
+        chain: RestApiGatewayHandlerChain,
+        context: RestApiInvocationContext,
+        response: Response,
+    ):
+        if context.integration:
+            invocation_type = context.integration["type"]
+            if invocation_type == "AWS":
+                service_name = self._get_aws_integration_service(context.integration.get("uri"))
+                invocation_type = f"{invocation_type}:{service_name}"
+        else:
+            # if the invocation does not have an integration attached, it probably failed before routing the request,
+            # hence we should count it as a NOT_FOUND invocation
+            invocation_type = "NOT_FOUND"
+
+        self.counter.record(invocation_type)
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
```diff
@@ -0,0 +1,44 @@
...
+    ):
+        if context.integration:
+            invocation_type = context.integration["type"]
+            if invocation_type == "AWS":
+                service_name = self._get_aws_integration_service(context.integration.get("uri"))
+                invocation_type = f"{invocation_type}:{service_name}"
+        else:
+            # if the invocation does not have an integration attached, it probably failed before routing the request,
+            # hence we should count it as a NOT_FOUND invocation
+            invocation_type = "NOT_FOUND"
```
cloutierMat, line 32: Question: I wonder if this is a relevant data point? On it's own, what intel can we gather from it, except for maybe knowing the total amount of requests sent to apigw? But is it relevant?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
```diff
@@ -0,0 +1,44 @@
...
+
+        self.counter.record(invocation_type)
+
+    @staticmethod
+    def _get_aws_integration_service(integration_uri: str) -> str:
+        if not integration_uri:
+            return "null"
+
+        if len(split_arn := integration_uri.split(":", maxsplit=5)) < 4:
+            return "null"
```
cloutierMat, line 42: Question: Same here. It seems we are we just capturing misconfigured Api? Is it relevant?

cloutierMat (approved):
Thanks for adding this! So awesome how clean it is to add this as a handler! :rocket: 

I have a couple questions as to the relevancy of some of the captured data point. If you feel those are important, let's merge. But if not, we might be better to eliminate some noise? :thinking:

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
bentsku: I think it would be good to know the total amount of requests sent to the endpoint. Not very relevant directly, but I think relevant if we do the sum of all types regardless of their value?

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
bentsku: The check is mostly to avoid exceptions, making sure we have a proper value to send

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
bentsku: Maybe @thrau could chime in too, I also see the point, so yeah I think it depends on expectations!

cloutierMat: Btw, I am OK either way! :+1: I was simply raising a question, to make sure we are deliberate with logging :wink:

PR #11861, opened 2024-11-18
[SFN] Fix access to selected items in Distributed Map state

gregfurman:
## Motivation
This PR fixes input passed to an `ItemSelector` not being correctly propogated to each iteration of a distributed map run. 

## Changes
- Instead of retrieving input data from the stack via a peek, items are instead popped off.
- Added `ItemSelector` tests for both a `DISTRIBUTED` and  `INLINE` map type.


dominikschubert (approved):
LGTM, nice catch. 

Kinda easy to miss. Would love if we had a bit more documentation around when to pop vs. peek the stack.

MEPalma (approved):
LGTM! 
It's true, the map run should consume it's input instead of just reading it. Not doing this meant that the iteration logic could not evaluate on the state's input (or it's reduction through InputPath), but on the ItemsPath;s result. Good catch, thank you @gregfurman!

PR #11862, opened 2024-11-18
add analytics for SNS internal routes

bentsku:
## Motivation
As part of our effort to improve our analytics, this PR implements usage counters for SNS internal routes.

We can now properly have usage data on SNS internal routes invocations, to understand their usage better. 

Usage is separated on what the internal route is targeting, be it Platform Endpoint messages, SMS messages or Subscription Tokens (useful for example for email subscriptions). 

## Changes
- implement analytics for SNS internal routes via a base class + decorator on routes

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -1,4 +1,5 @@
 import base64
+import functools
 import json
 import logging
 from typing import Dict, List
@@ -60,6 +61,7 @@ from localstack.services.edge import ROUTER
 from localstack.services.moto import call_moto
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.services.sns import constants as sns_constants
+from localstack.services.sns import usage
 from localstack.services.sns.certificate import SNS_SERVER_CERT
 from localstack.services.sns.filter import FilterPolicyValidator
 from localstack.services.sns.models import SnsMessage, SnsStore, SnsSubscription, sns_stores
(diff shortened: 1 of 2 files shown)


giograno:
FYI, we might have some overlap with https://github.com/localstack/localstack/pull/11855

bentsku:
@giograno yes, I think there might be some issue with #11855 because it will start counting every path for every internal route, even API Gateway which can technically have a lot of different path. I don't know how we can make it opt-in / opt-out, and maybe more granular?

localstack-core/localstack/services/sns/provider.py
```diff
@@ -1116,7 +1118,25 @@ def _format_messages(sent_messages: List[Dict[str, str]], validated_keys: List[s
...
+
+    def count_usage(self):
+        usage.internalapi.record(f"{self.resource_type}")
+
+
+def count_usage(f):
+    @functools.wraps(f)
+    def _wrapper(self, *args, **kwargs):
+        self.count_usage()
+        return f(self, *args, **kwargs)
```
thrau, line 1133: perhaps we don't need `self`?

```suggestion
    def _wrapper(*args, **kwargs):
        self.count_usage()
        return f(*args, **kwargs)
```

thrau (approved):
LGTM! one minor nit

i think it would be great to solve this more generically across endpoints moving forward, but good for now to get some initial readings!

localstack-core/localstack/services/sns/provider.py
bentsku: I think this will complain, we do need self passed, it failed before I added it 😄

bentsku: If I don't pass self, then I need to extract it from `args` right, otherwise `self.count_usage()` will raise an exception as it wouldn't be in scope?

thrau: oh ,whoops, didn't see that there, of course your right!

bentsku: I'm not too fond of this either tbh, for a one line change I could just call `self.count_usage()` inside the method itself... it's the same as decorating it 😬

PR #11863, opened 2024-11-18
re-add usage analytics collector_registry registration

bentsku:
## Motivation
Follow up from #11848, following the refactor and quick tests / copy-paste from scratch files, I've removed the registration of the counters 😬 

Note: can also be added to #11854 if approved. 

## Changes
- put back the registration


simonrw (approved):
Thanks for resolving this so quickly!

PR #11868, opened 2024-11-18
Track replicator usage in analytics

simonrw:
## Motivation

With the new replicator feature we should track its usage via our analytics.

## Changes

* Include `ENABLE_REPLICATOR` in the envars we track with analytics


cloutierMat (approved):
Thanks for adding!

PR #11876, opened 2024-11-19
Remove DNS_LOCAL_NAME_PATTERNS

simonrw:
## Motivation

We deprecated `DNS_LOCAL_NAME_PATTERNS` in LocalStack v3.0 as its name was confusing. We should now remove this variable.

## Changes

* Remove `DNS_LOCAL_NAME_PATTERNS` and all references to it


dfangl (commented):
I think we should keep the deprecating warning in, but rephrase it that it has been removed with 4.0, similar to what we do for other features as well. What do you think?

PR #11877, opened 2024-11-19
Use jpype-ext and set jpype interrupt=False

dfangl:
## Motivation
Currently, jpype1 only installs SIGINT signal handlers if `interrupt=False`. With https://github.com/localstack/jpype/pull/1, we (very crudely) added support for SIGTERM as well.

We should aim to upstream this changes ASAP, but for the 4.0 release, this should suffice to get proper shutdown handling in LocalStack even if a library is loaded using jpype.

## Changes
* Use jpype-ext, published with the changes here: https://github.com/localstack/jpype/pull/1
* Set `interrupt=False` when starting jpype


maxhoheiser (approved):
LGTMT - thanks for addressing this Daniel

PR #11879, opened 2024-11-19
Prevent usage of java rule engine in tests

dominikschubert:
## Motivation

We can't start both the JPype-based java rule engine as well as other JPype-based features in the same run, so we can't parametrize on this. If really necessary it would need to be added as a separate pipeline/test run.

## Changes

- Skip unit tests that would use the java rule engine
- Conditionally skip tests instead of monkeypatching java rule engine in integration tests


joe4dev (approved):
LGTM 👍

PR #11880, opened 2024-11-19
Put events fix for multiple targets

zaingz:
## Motivation
Fixes EventBridge PutEvents response so it correctly returns event IDs only once per event, regardless of how many rules or targets match. Currently, the response includes duplicate EventIds for each matching target, which doesn't match AWS behavior and can cause issues when clients try to track events by their IDs.

Related issue: https://github.com/localstack/localstack/issues/11597

## Changes
- Modified `_process_entries` and related methods in events provider to track and return event IDs only once
- Refactored event processing logic to maintain a single event result per input event
- Added test case to verify the fix and prevent regressions

## Testing
1. Create an EventBridge rule with multiple targets
2. Send an event that matches the rule
3. Verify that:
   - PutEvents response contains each EventId only once
   - All targets still receive the event correctly
   - Event IDs are consistent between response and delivered messages

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -145,7 +145,6 @@ from localstack.services.events.target import (
     TargetSenderDict,
     TargetSenderFactory,
 )
-from localstack.services.events.usage import rule_error, rule_invocation
 from localstack.services.events.utils import (
     extract_event_bus_name,
     extract_region_and_account_id,
@@ -1335,6 +1334,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
                 f"1 validation error detected: Value '{formatted_entries}' at 'entries' failed to satisfy constraint: Member must have length less than or equal to 10"
             )
         entries, failed_entry_count = self._process_entries(context, entries)
+
         response = PutEventsResponse(
             Entries=entries,
             FailedEntryCount=failed_entry_count,
@@ -1958,13 +1958,16 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
             failed_entry_count["count"] += 1
             LOG.info(json.dumps(event_failed_validation))
             return
+
         region, account_id = extract_region_and_account_id(event_bus_name_or_arn, context)
         if encoded_trace_header := get_trace_header_encoded_region_account(
             entry, context.region, context.account_id, region, account_id
         ):
             entry["TraceHeader"] = encoded_trace_header
+
         event_formatted = format_event(entry, region, account_id, event_bus_name)
         store = self.get_store(region, account_id)
+
         try:
             event_bus = self.get_event_bus(event_bus_name, store)
         except ResourceNotFoundException:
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/events/provider.py
```diff
@@ -2034,14 +2040,7 @@ def _process_rules(
                     target_sender = self._target_sender_store[target_arn]
                     try:
                         target_sender.process_event(event_formatted.copy())
-                        processed_entries.append({"EventId": event_formatted["id"]})
                     except Exception as error:
-                        processed_entries.append(
```
maxhoheiser, line 2039: how is this now handled, vaguely remember that AWS handles errors with targets this way - but I see that we don't have a test for this. If testing this I would do it via a missing IAM permission e.g. for EventBrdige to write to an SQS target because this should invoke a failed on send message and result in a similar error.

tests/aws/services/events/test_events.py
```diff
@@ -568,6 +568,137 @@ def test_create_connection_validations(self, aws_client):
...
+                snapshot.transform.regex(queue_arn_1, "<queue-1-arn>"),
+                snapshot.transform.regex(queue_arn_2, "<queue-2-arn>"),
+                snapshot.transform.regex(rule_name, "<rule-name>"),
+                *snapshot.transform.sqs_api(),
+                *snapshot.transform.sns_api(),
+            ]
+        )
+
+        rule_arn = f"arn:aws:events:{region}:{account_id}:rule/{rule_name}"
+        sqs_policy = {
```
maxhoheiser, line 599: there is a fixture that takes care of attaching the correct policy for you `create_sqs_events_target`
`queue_url, queue_arn = create_sqs_events_target()`

tests/aws/services/events/test_events.py
```diff
@@ -568,6 +568,137 @@ def test_create_connection_validations(self, aws_client):
...
+            QueueUrl=queue_url_1, Attributes={"Policy": json.dumps(sqs_policy)}
+        )
+        aws_client.sqs.set_queue_attributes(
+            QueueUrl=queue_url_2, Attributes={"Policy": json.dumps(sqs_policy)}
+        )
+
+        events_put_rule(
+            Name=rule_name,
+            EventPattern=json.dumps(
+                {"source": ["test.source"], "detail-type": ["test.detail.type"]}
```
maxhoheiser, line 623: it would make sense to use the constant pattern here too: `TEST_EVENT_PATTERN_NO_DETAIL` from `test_events.py` and similarly use `TEST_EVENT_PATTERN` as test event in line 647

tests/aws/services/events/test_events.py
```diff
@@ -568,6 +568,137 @@ def test_create_connection_validations(self, aws_client):
...
+                {"Id": target_id_1, "Arn": queue_arn_1},
+                {"Id": target_id_2, "Arn": queue_arn_2},
+            ],
+        )
+
+        assert (
+            target_response["FailedEntryCount"] == 0
+        ), f"Failed to add targets: {target_response.get('FailedEntries', [])}"
+
+        test_event = {
```
maxhoheiser, line 647: it would make sense to use the constant pattern and detail here:

```
[
            {
                "Source": TEST_EVENT_PATTERN_NO_DETAIL["source"][0],
                "DetailType": TEST_EVENT_PATTERN_NO_DETAIL["detail-type"][0],
                "Detail": json.dumps(EVENT_DETAIL),
            },
        ]
```

maxhoheiser (commented):
Nice changes - I am not 100% sure about AWS parity with removing the InternalException ErrorCode though.

localstack-core/localstack/services/events/provider.py
zaingz: Thanks for the feedback Max! After running the test against AWS, I discovered that AWS actually returns a successful EventId even when there are target permission failures. I've updated the code and tests to match this behavior.

Specifically:
- PutEvents returns a successful EventId and FailedEntryCount=0
- The failure to deliver to targets (due to missing permissions) doesn't affect the API response
- Added a snapshot test that verifies this behavior

maxhoheiser: Ok nice - yes it is a bit strange why they do it like this - they also silently fail if you put an event to a non-existent event bus (see line 1357)

maxhoheiser (approved):
Nice implementation to fix the problem 👍 (LGTMT when tests are green :) )

joe4dev (approved):
LGTM 👍

PR #11883, opened 2024-11-20
Add lambda ESM analytics

joe4dev:
## Motivation

Add usage analytics for Lambda ESM to learn more about potential errors we should tackle and usage of different event sources.

## Changes

* Add usage analytics for `invocation` (successful ones) and `error` (unexpected error; i.e., LS bug)
* Add unit test for existing set counter (also serves to exemplify usage)

### DIFF
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -1,5 +1,6 @@
 import datetime
 import math
+import threading
 from collections import defaultdict
 from itertools import count
 from typing import Any
@@ -77,6 +78,7 @@ class UsageMultiSetCounter:
 
     namespace: str
     _counters: dict[str, UsageSetCounter]
+    lock = threading.Lock()
 
     def __init__(self, namespace: str):
         self._counters = {}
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/utils/analytics/usage.py
```diff
@@ -49,6 +49,49 @@ def aggregate(self) -> dict:
...
+    namespace: str
+    _counters: dict[str, UsageSetCounter]
+
+    def __init__(self, namespace: str):
+        self._counters = {}
+        self.namespace = namespace
+
+    def record(self, key: str, value: str):
+        namespace = f"{self.namespace}:{key}"
+        set_counter = self._counters.setdefault(key, UsageSetCounter(namespace))
```
bentsku, line 85: I think using `setdefault` here might lead to an issue: when initializing a `UsageSetCounter`, it will register itself in the register with `collector_registry[namespace] = self`, overriding the namespace. 

I think we should probably not create the entry if `config.DISABLE_EVENTS` is enabled.

localstack-core/localstack/utils/analytics/usage.py
```diff
@@ -49,6 +49,59 @@ def aggregate(self) -> dict:
...
+
+    def record(self, key: str, value: str):
+        namespace = f"{self.namespace}:{key}"
+        # We cannot use setdefault here because Python always instantiates a new UsageSetCounter,
+        # which overwrites the collector_registry
+        if namespace in self._counters:
+            set_counter = self._counters[namespace]
+        else:
+            set_counter = UsageSetCounter(namespace)
+            self._counters[namespace] = set_counter
```
dfangl, line 93: This is not thread safe, we might unnecessarily instantiate some usagesetcounters with the same namespace here, which might override each other in an unprecedented way. This could for example happen in multiple event source mappings in multiple thread.

Since we cannot use setdefault, I'd rather just add a lock here.

dfangl (approved):
LGTM now! Thanks for addressing the thread safety issue!

thrau:
as discussed we've decided to postpone this for now and dive a bit deeper into the requirements first. i think it's worth delineating between usage analytics and observability, and perhaps understand more what we can learn from tracking things like errors. let's discuss post v4 :-)

PR #11884, opened 2024-11-20
Add EventBridge rule target analytics

joe4dev:
## Motivation

EventBridge v2 misses analytics to learn more about which targets are matched how frequently and how often they fail.

## Changes

* Add usage analytics for invocations and errors of rule targets

## Discussion

The error case currently does not distinguish between expected errors (i.e., AWS parity) and unexpected errors (i.e., not accounted), but I don't know what the AWS behavior here. Currently, we just ignore errors.

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -145,6 +145,7 @@ from localstack.services.events.target import (
     TargetSenderDict,
     TargetSenderFactory,
 )
+from localstack.services.events.usage import rule_error, rule_invocation
 from localstack.services.events.utils import (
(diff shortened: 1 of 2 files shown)


maxhoheiser (approved):
Nice feature

might be also and idea to track if an archive is added and if events are replayed from an archive

dfangl (approved):

joe4dev:
> might be also and idea to track if an archive is added and if events are replayed from an archive

@maxhoheiser AWS Request analytics are always captured and start_replay invokes `_process_entries`, so these counters should also capture rule matching for replays. Do you mean distinguishing replays from other rule matching or something else?

Issue #11843, opened 2024-11-13
bug: Appsync graphql errors when including multiple fields in query


Rohan0135:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running graphql queries via app sync against dynamodb, certain queries are showing inconsistent behaviour between localstack and aws. 

For queries where we have defined one or more fields in the @key , we are getting unexpected errors. 

When we have 2 fields (partition key and sort key)  defined we encounter the following error: 

`"An error occurred (ValidationException) when calling the Query operation: KeyConditionExpressions must only contain one condition per key",`

When we have 1 field  (just partition key) defined we get : 

`sortDirection is not supported for List operations without a Sort key defined`

### Expected Behavior

We expect both queries to return results i.e. data from dynamodb as it does with aws

 

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

Amplify graphql schema: 

```
type Order
  @model(
    timestamps: { createdAt: "created_at", updatedAt: "updated_at" }
    subscriptions: null
  )
  @key(name: "byRestaurantAndStatus",fields: ["restaurant_id","current_status"],queryField: "ordersByRestaurantAndStatus")
  @key(name: "byUser", fields: ["user_id"], queryField: "ordersByUser")

  ) {
  id: ID!
  user_id: ID
  restaurant_id: ID!
  current_status: String!
  total_price: Float!
  created_at: AWSDateTime!
  updated_at: AWSDateTime!
}
```

Query 1. 
```
query OrdersByUser {
    ordersByUser(user_id: "<id>", limit: 10) {
        nextToken
        items {
            id
            user_id
            restaurant_id
            current_status
     }
}
```

Query 2. 

```
ordersByRestaurantAndStatus(
        restaurant_id: "<id>"
        current_status: { eq: "<status>" }
        limit: 10
    ) {
        nextToken
        items {
            id
            user_id
            restaurant_id
            current_status
         }
}
```

### Environment

```markdown
- OS: MacOS 14.2.1
- LocalStack:
  LocalStack version: 3.8 
  LocalStack build git hash: 80b630c24f1a57a8bccac741a2e13b54f83e4c18
```

cloutierMat:
Hi @Rohan0135, thank you for your report. 

[...]

PR #11886, opened 2024-11-20
fix vtl resolver patch

cloutierMat:
## Motivation

While investigating the issue reported by #11843, I noticed that the default returned value of unfound variables differs between apigateway and appsync. We had this patch working for apigateway, but it would unfortunately fail some appsync functions that are expecting to receive a null value.

## Changes

- Augmented the signature of `prepare_namespace` to allow setting a `source`.
- Patch only if the source is apigateway
- Reimplemented the template used by the previous test reffered in the comment

_fixes #11843_


cloutierMat:
@simonrw I added you here as it directly impacts appsync :wink:

simonrw (approved):
Great change, thank you for making a clean implementation on being able to dispatch between AppSync and API Gateway behaviour!

bentsku (approved):
LGTM! Neat fix, thanks for updating and validating the test! Glad that it still caught a possible regression... 😅

PR #11888, opened 2024-11-21
fix java home for macos

macnev2013:
## Motivation
This PR resolves the issue with Java home retrieval on macOS. For the temurin Java distribution, the Java home directory on macOS resides in <install-dir>/Contents/Home. This discrepancy caused errors in big data services when running in host mode.

## Changes
Updated the logic for Java home retrieval on macOS to accommodate the temurin distribution structure.


viren-nadkarni (approved):
LGTM, thanks for improving the devx in Mac OS :heavy_check_mark:

maxhoheiser:
For me, it worked before the patch - and with the patch, it introduces a double nesting like so: `Contents/Home/Contents/Home`

PR #11891, opened 2024-11-21
Refactor EventBridge ArchiveServices to allow restoration without side effects

dfangl:
## Motivation
Currently, ArchiveServices have quite some side effects in its constructor.

This prevents us from avoiding those side effects when we want to restore those services from data in the store (for cloudpods or persistence).

With this PR, we split out the logic out of the constructor, to allow instantiation without side effects.

## Changes
* Remove boto calls from the ArchiveService constructor
* Refactor a bit to manually trigger rule/target registration


joe4dev (approved):
LGTM

PR #11892, opened 2024-11-21
fixed java lib path for macos

macnev2013:
## Motivation
This is a follow up PR to https://github.com/localstack/localstack/pull/11888#issuecomment-2490739229. @maxhoheiser [reported a bug](https://github.com/localstack/localstack/pull/11888#issuecomment-2490739229) where java lib path was not properly generated.

Thanks @maxhoheiser for reporting.

## Changes
- Updated java lib path for macos to not include `/Contents/Home` since it's already included in the java home.

### DIFF
--- a/localstack-core/localstack/packages/java.py
+++ b/localstack-core/localstack/packages/java.py
@@ -45,7 +45,7 @@ class JavaInstallerMixin:
         """
         if java_home := self.get_java_home():
             if is_mac_os():
-                return os.path.join(java_home, "Contents", "Home", "lib", "jli", "libjli.dylib")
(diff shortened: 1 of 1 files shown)


viren-nadkarni (commented):
Mac JRE builds seem to have a different file tree altogether. Have you checked if the paths on line 70 and 72 for example are valid with this change?

macnev2013:
Yes @viren-nadkarni, Since we've directly updated the JAVA_HOME, all the variables are updated accordingly.

e.g: For `ld_library_path` we are just adding {java_home}/lib (which already includes correct path for mac)

viren-nadkarni (approved):

PR #11896, opened 2024-11-21
Enable EventBridge API destinations and connections

dfangl:
## Motivation
In the current v2 provider, we do not support eventbridge API destinations.

As we do want this in 4.0.1, and I am unsure if we should do such a big code change, I staggered the commits in terms of escalating changes / complexity.

93822dd has the minimal changes necessary to make the feature work, we could cherry pick that into a separate PR.

After this commit, we increasingly increase parity, with more snapshots, creation of eventbridge secrets, bypass of secretmanager name assertions (since official EventBridge secrets contain a `!` in the name).

It might be too much for a release tomorrow, but I will let the reviewers judge that.

## Changes
* Allow persistence for connections and api destinations by moving them into the store (very simply, without dataclasses as for other models)
* Detect API destinations and forward events correctly
* Connections now correctly create secrets in secretsmanager for connection secrets
* Change snapshot transformers to add more assertions on the structure of ARNs
* Some minor testing improvements, to increase reliability of the tests (against AWS)
* Move the snapshots to an autouse fixture, to be able to use the `snapshot` name directly (honestly a preference, but I did quite some work in the tests)
* Add more snapshot on the event secrets
* Unskip test for api destinations
* Fix partitions in some ARNs

### DIFF
--- a/localstack-core/localstack/services/events/models.py
+++ b/localstack-core/localstack/services/events/models.py
@@ -9,6 +9,13 @@ from localstack.aws.api.events import (
     ArchiveName,
     ArchiveState,
     Arn,
+    ConnectionArn,
+    ConnectionAuthorizationType,
+    ConnectionDescription,
+    ConnectionName,
+    ConnectionState,
+    ConnectionStateReason,
+    CreateConnectionAuthRequestParameters,
     CreatedBy,
     EventBusName,
     EventPattern,
@@ -27,6 +34,7 @@ from localstack.aws.api.events import (
     RuleName,
     RuleState,
     ScheduleExpression,
+    SecretsManagerSecretArn,
     TagList,
     Target,
     TargetId,
@@ -223,6 +231,23 @@ class EventBus:
 EventBusDict = dict[EventBusName, EventBus]
 
 
+class Connection(TypedDict, total=False):
+    ConnectionArn: Optional[ConnectionArn]
+    Name: Optional[ConnectionName]
+    Description: Optional[ConnectionDescription]
+    ConnectionState: Optional[ConnectionState]
+    StateReason: Optional[ConnectionStateReason]
+    AuthorizationType: Optional[ConnectionAuthorizationType]
+    SecretArn: Optional[SecretsManagerSecretArn]
+    AuthParameters: Optional[CreateConnectionAuthRequestParameters]
+    CreationTime: Optional[Timestamp]
+    LastModifiedTime: Optional[Timestamp]
+    LastAuthorizedTime: Optional[Timestamp]
+
+
+ConnectionDict = dict[ConnectionName, Connection]
+
+
 class EventsStore(BaseStore):
     # Map of eventbus names to eventbus objects. The name MUST be unique per account and region (works with AccountRegionBundle)
     event_buses: EventBusDict = LocalAttribute(default=dict)
@@ -232,6 +257,9 @@ class EventsStore(BaseStore):
 
     # Map of replay names to replay objects. The name MUST be unique per account and region (works with AccountRegionBundle)
     replays: ReplayDict = LocalAttribute(default=dict)
+
+    # Map of connection names to connection objects.
+    connections: ConnectionDict = LocalAttribute(default=dict)
 
     # Maps resource ARN to tags
     TAGS: TaggingService = CrossRegionAttribute(default=TaggingService)
--- a/localstack-core/localstack/services/events/models.py
+++ b/localstack-core/localstack/services/events/models.py
@@ -5,18 +5,15 @@ from typing import Literal, Optional, TypeAlias, TypedDict
 
 from localstack.aws.api.core import ServiceException
 from localstack.aws.api.events import (
+    ApiDestinationName,
     ArchiveDescription,
     ArchiveName,
     ArchiveState,
     Arn,
-    ConnectionArn,
-    ConnectionAuthorizationType,
-    ConnectionDescription,
     ConnectionName,
-    ConnectionState,
-    ConnectionStateReason,
-    CreateConnectionAuthRequestParameters,
     CreatedBy,
+    DescribeApiDestinationResponse,
+    DescribeConnectionResponse,
     EventBusName,
     EventPattern,
     EventResourceList,
@@ -34,7 +31,6 @@ from localstack.aws.api.events import (
     RuleName,
     RuleState,
     ScheduleExpression,
-    SecretsManagerSecretArn,
     TagList,
     Target,
     TargetId,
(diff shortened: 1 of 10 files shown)


localstack-core/localstack/services/secretsmanager/provider.py
```diff
@@ -173,9 +173,20 @@ def create_secret(
...
+            tag_entry["Key"] == "BYPASS_SECRET_ID_VALIDATION"
+            for tag_entry in request.get("Tags", [])
+        ):
+            self._raise_if_invalid_secret_id(request["Name"])
+        else:
+            request["Tags"] = [
+                tag_entry
+                for tag_entry in request.get("Tags", [])
+                if tag_entry["Key"] != "BYPASS_SECRET_ID_VALIDATION"
+            ]
```
dfangl, line 187: We sadly need something to bypass the name validation here, as eventbridge connection secrets contain a `!`, which is not allowed. I used a tag here, but there are other options should we decide against it.

tests/aws/services/events/conftest.py
```diff
@@ -566,7 +566,12 @@ def _create_connection(auth_type_or_auth, auth_parameters=None):
                 AuthParameters=auth_parameters,
             )
 
-    return _create_connection
+    yield _create_connection
+
+    try:
+        aws_client.events.delete_connection(Name=connection_name)
+    except Exception as e:
+        LOG.debug("Error cleaning up connection: %s", e)
```
dominikschubert, line 574: Nice catch. Kinda reminding me of the fact that it  might be nice to revive the client-based cleanups at some point :grin:

localstack-core/localstack/services/secretsmanager/provider.py
dominikschubert: >  which is not allowed

I'm not sure I'm following here. Not allowed as in our SecretsManager service *should* support it, but doesn't? Or how would this otherwise work on AWS?

dominikschubert: Are AWS services able to create keys with names that we (as external users) can't?

tests/aws/services/events/test_events.py
```diff
@@ -1890,7 +1922,38 @@ def test_create_connection_name_validation(
...
+        snapshot.match("create-connection-invalid-name-error", e.value.response)
+
+    @markers.aws.validated
+    @pytest.mark.parametrize(
+        "auth_params", API_DESTINATION_AUTH_PARAMS, ids=["basic", "api-key", "oauth"]
+    )
+    @pytest.mark.skipif(
+        is_old_provider(),
+        reason="V1 provider does not support this feature",
+    )
```
dominikschubert, line 1934: @dfangl going above and beyond :rocket:

localstack-core/localstack/services/events/models.py
```diff
@@ -223,6 +227,10 @@ def arn(self) -> Arn:
 EventBusDict = dict[EventBusName, EventBus]
 
 
+ConnectionDict = dict[ConnectionName, DescribeConnectionResponse]
+ApiDestinationDict = dict[ApiDestinationName, DescribeApiDestinationResponse]
```
dominikschubert, line 231: Guess that's something for service owners to judge, but I don't see this blocking for this iteration.

localstack-core/localstack/services/events/target.py
```diff
@@ -390,6 +391,10 @@ class EventsTargetSender(TargetSender):
     def send_event(self, event):
         # TODO add validation and tests for eventbridge to eventbridge requires Detail, DetailType, and Source
         # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/events/client/put_events.html
+        target_arn = self.target["Arn"]
+        if ":api-destination/" in target_arn or ":destination/" in target_arn:
+            send_event_to_api_destination(target_arn, event, self.target.get("HttpParameters"))
```
dominikschubert, line 396: @maxhoheiser Might be something were instrumentation for EventStudio  might be a good idea in a follow up

localstack-core/localstack/services/events/provider.py
```diff
@@ -339,33 +338,112 @@ def _create_connection_arn(
...
+        self,
+        context: RequestContext,
+        name: str,
+        authorization_type: ConnectionAuthorizationType,
+        auth_parameters: CreateConnectionAuthRequestParameters,
+    ) -> str:
         """Create a standardized secret ARN."""
-        return f"arn:aws:secretsmanager:{context.region}:{context.account_id}:secret:events!connection/{name}/{connection_uuid}"
+        # TODO use service role as described here: https://docs.aws.amazon.com/eventbridge/latest/userguide/using-service-linked-roles-service-action-1.html
+        # not too important as it is created automatically on AWS anyway, with the right permissions
```
dominikschubert, line 383: even when using the  API or only via  the AWS Console?

localstack-core/localstack/services/events/provider.py
```diff
@@ -339,33 +338,112 @@ def _create_connection_arn(
         self, context: RequestContext, name: str, connection_uuid: str
     ) -> str:
         """Create a standardized connection ARN."""
-        return f"arn:aws:events:{context.region}:{context.account_id}:connection/{name}/{connection_uuid}"
+        return f"arn:{get_partition(context.region)}:events:{context.region}:{context.account_id}:connection/{name}/{connection_uuid}"
```
dominikschubert, line 341: nice catch here as well :+1:

dominikschubert (approved):
Nice! :trophy:  Think it's a good idea to move from the transient state in the provider to  directly using the entities from the store. Much simpler and less error-prone.

The somewhat weird workaround with the  secret names might be worth looking into later again. Maybe we can just detect  it as  an internal call via headers/request context and then allow arbitrary keys?  We can discuss this at a later  date though.

localstack-core/localstack/services/secretsmanager/provider.py
dfangl: It seems that way. The error message clearly states it should not be allowed, the aws console also does not allow it. So maybe aws services have some bypass for it.

tests/aws/services/events/test_events.snapshot.json
```diff
@@ -1754,7 +1754,7 @@
       ]
     }
   },
-    "tests/aws/services/events/test_events.py::TestEvents::test_create_connection_validations": {
+  "tests/aws/services/events/test_events.py::TestEvents::test_create_connection_validations": {
```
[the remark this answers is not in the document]
dfangl: Didn’t notice that, that is really interesting. Did someone manually mess with this file?

localstack-core/localstack/services/events/provider.py
dfangl: Should also be the case with the api.

localstack-core/localstack/services/events/models.py
dfangl: This is really a temporary shortcut to get some typing, I would also think this needs another iteration at some point.

joe4dev (approved):
LGTM 🚀  Thanks a lot @dfangl

PR #11898, opened 2024-11-21
initial kms key identifier

pinzon:
## Motivation
This PR implements a new class called Key Identifier that allows the setting of a custom ID for a new KMS Key

## Changes
- New class for custom Id

## Testing
- Testing done in another PR

### DIFF
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -7,7 +7,6 @@ import os
 import random
 import re
 import struct
-import uuid
 from collections import namedtuple
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


cloutierMat (commented):
I left a couple comments, but I am still not convinced this is a good solution here...

I appreciate how far you dug trying to make it work with the `description` :rocket:. But it would be hard to communicate to users and we might still hit a point where a user is trying to replicate many keys at once with no description?

We might want to use a different strategy here. Since this service does support tags. I feel like it would be safer in this case to simply add the tags at the time of replicating?

WDYT? @simonrw Do you have a better idea?

pinzon:
As discussed in the meeting. The KMS Key is a special case where no particular attribute can be used as an Identifier and the resource already accept tags. For setting the Id will limit ourselves to use the Tags.

PR #11903, opened 2024-11-22
Fix put_targets validation regressions

joe4dev:
## Motivation

`put_targets` does not validate invalid inputs for `target_id`, which is a regression compared to EventBridge v1.

## Changes

* Add validation exceptions for `put_targets`

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -146,6 +146,7 @@ from localstack.services.events.target import (
     TargetSenderFactory,
 )
 from localstack.services.events.utils import (
+    TARGET_ID_PATTERN,
     extract_event_bus_name,
     extract_region_and_account_id,
     format_event,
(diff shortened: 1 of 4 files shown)


dfangl (approved):
Looks good!

localstack-core/localstack/services/events/provider.py
```diff
@@ -1152,7 +1153,16 @@ def put_targets(
         failed_entries = rule_service.add_targets(targets)
         rule_arn = rule_service.arn
         rule_name = rule_service.rule.name
-        for target in targets:  # TODO only add successful targets
+        for index, target in enumerate(targets):  # TODO only add successful targets
```
bentsku, line 1156: came too late, but python tip: `enumerate` takes a `start` argument, avoiding this `index+1` issue

`for index, target in enumerate(targets, start=1):`

dfangl: That is really cool, didn't know that!

PR #11913, opened 2024-11-22
Fix linting issue after namespace package fix

dominikschubert:
## Motivation

Apologies. Was a bit too trigger happy and really didn't think that it would create a linting issue somewhere completely unrelated.

## Changes

- Fixes linting issues in `server.py` for whatever reason


cloutierMat (approved):
weird it didn't trigger the linter before :+1:

PR #11920, opened 2024-11-25
fix(providers): Removed cloudcontrol provider

Pive01:
## Motivation
Removed cloudcontrol provider for community as we moved it in pro

## Changes

Removed cloudcontrol  from the list of community providers

### DIFF
--- a/localstack-core/localstack/services/providers.py
+++ b/localstack-core/localstack/services/providers.py
@@ -42,14 +42,6 @@ def apigateway_legacy():
 
 
 @aws_provider()
-def cloudcontrol():
-    from localstack.services.cloudcontrol.provider import CloudControlProvider
-
-    provider = CloudControlProvider()
-    return Service.for_provider(provider)
-
-
-@aws_provider()
 def cloudformation():
     from localstack.services.cloudformation.provider import CloudformationProvider
 


silv-io (approved):
Is the code in `localstack.services.cloudcontrol.provider` also migrated to PRO?

If yes, then it's alright. Otherwise we'd have to take a look at that as well before merging

Pive01:
> Is the code in `localstack.services.cloudcontrol.provider` also migrated to PRO?
> 
> If yes, then it's alright. Otherwise we'd have to take a look at that as well before merging

Yes, everything has been moved [here](https://github.com/localstack/localstack-ext/pull/3699)

silv-io:
> > Is the code in `localstack.services.cloudcontrol.provider` also migrated to PRO?
> > If yes, then it's alright. Otherwise we'd have to take a look at that as well before merging
> 
> Yes, everything has been moved [here](https://github.com/localstack/localstack-ext/pull/3699)

In that PR I can only see it being added to pro, but where was that removed from community?

Pive01:
> > > Is the code in `localstack.services.cloudcontrol.provider` also migrated to PRO?
> > > If yes, then it's alright. Otherwise we'd have to take a look at that as well before merging
> > 
> > 
> > Yes, everything has been moved [here](https://github.com/localstack/localstack-ext/pull/3699)
> 
> In that PR I can only see it being added to pro, but where was that removed from community?

Was never added in the first time, initially was [added](https://github.com/localstack/localstack/pull/11830) but half way through we moved it

simonrw (approved):

Issue #11916, opened 2024-11-23
bug: Step Functions describe-activity incorrectly rejects ARN


peter-smith-phd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running `describe-activity` on an activity ARN that contains punctuation symbols in the name, an `Invalid arn` error is reported.

```
$ awslocal stepfunctions create-activity --name RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c
{
    "activityArn": "arn:aws:states:ca-west-1:000000000000:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c",
    "creationDate": 1732374466.340198
}

$ awslocal stepfunctions describe-activity --activity-arn arn:aws:states:ca-west-1:000000000000:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c

An error occurred (InvalidArn) when calling the DescribeActivity operation: Invalid arn: 'arn:aws:states:ca-west-1:000000000000:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c'
```

### Expected Behavior

The correct response (from AWS) is as follows:

```
$ aws stepfunctions create-activity --name RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c
{
    "activityArn": "arn:aws:states:ca-west-1:491085397214:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c",
    "creationDate": 1732373726.378
}

$ aws stepfunctions describe-activity --activity-arn arn:aws:states:ca-west-1:491085397214:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c
{
    "activityArn": "arn:aws:states:ca-west-1:491085397214:activity:RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c",
    "name": "RetryParallelStateIntegrationTest.2024-11-23-06-19-53.60e0ecac-0dbb-4e5e-bddb-c",
    "creationDate": 1732373726.378,
[...]

PR #11924, opened 2024-11-25
[SFN] Activity Name and Arn Validation Logic

MEPalma:
## Motivation
Currently, the SFN v2 provider applies faulty activity arn and name validation logics, which can result is `InvalidName` exceptions during activity creation and especially during `DescribeActivity` api actions. This is espiecially true for activity names that contain punctuations,  as described in https://github.com/localstack/localstack/issues/11916. This is due to the activity arn validation not being aligned with AWS SFN and the activity name validation (used on creation). These changes address both issues and add positive and negative snapshot tests for the creation deletion and description of activity names and arns.

## Changes
- positive and negative activity name/arn tests
- updated activity arn validation rule
- updated activity name validation logic to check for maximum name lenght

### DIFF
--- a/tests/aws/services/stepfunctions/v2/test_sfn_api_activities.py
+++ b/tests/aws/services/stepfunctions/v2/test_sfn_api_activities.py
@@ -6,16 +6,31 @@ from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
 
 
-@markers.snapshot.skip_snapshot_verify(paths=["$..tracingConfiguration"])
+@markers.snapshot.skip_snapshot_verify(paths=["$..encryptionConfiguration"])
 class TestSnfApiActivities:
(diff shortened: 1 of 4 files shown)


tests/aws/services/stepfunctions/v2/test_sfn_api_activities.py
```diff
@@ -48,13 +63,40 @@ def test_create_describe_delete_activity(
         sfn_snapshot.match("delete_activity_response_2", delete_activity_response_2)
 
     @markers.aws.validated
+    @pytest.mark.parametrize(
```
dominikschubert, line 66: praise: very thorough :+1:

dominikschubert (approved):
LGTM, thanks for fixing this so quickly :rocket:

PR #11925, opened 2024-11-25
Remove custom spec for AWS StepFunctions

dominikschubert:
## Motivation
With https://github.com/localstack/localstack/commit/907c3b1d90a864ef516ea1f6d962cf8fc162acb7 we added a custom spec for the StepFunctions API to make sure 2 things happen:
1. ASF has a spec to use for serializing response objects
2. AWS clients used in our tests can de-serialize the response into proper boto3 response objects

With the most recent botocore update in https://github.com/localstack/localstack/pull/11918 , this isn't necessary anymore as it now already includes the newest changes to the StepFunctions API.

## Changes

- Removes the custom spec file for StepFunctions. With the recently updated botocore spec, this is basically a NO-OP.


alexrashed (approved):
Nice! Thanks a lot for following up on this now that the new specs are contained in the new `botocore` package! 💯 🧹 🦸🏽

MEPalma (approved):
LGTM

PR #11928, opened 2024-11-26
Add analytics for event rule engine

joe4dev:
Supersedes https://github.com/localstack/localstack/pull/11807

## Motivation

We are missing analytics tracking for the event ruler and a plugins.py to make the package installable (if we decide to ship it). Currently, we don't see the need for shipping it as the Java-based event ruler is opt-in.

## Changes

* Add env var tracking for `EVENT_RULE_ENGINE`
* Make `event-ruler` package installable (in case we need to ship it to unblock users facing dynamic installation issues)


dominikschubert (approved):
LGTM, ideally not gonna be necessary though if we can phase out the java one :+1:

PR #11930, opened 2024-11-26
add variable to configure SNS sender

bentsku:
## Motivation
As reported by @HarshCasper, the sender of SNS email notifications was hardcoded to `admin@localstack.com` with no mean of changing that. We should replace this value by something like `no-reply@sns.localstack.cloud` or something similar, as AWS is sending those notifications with `no-reply@sns.amazonaws.com`. 

I'm a bit worried about changing this directly however, as some users might expected and test that emails are arriving from that sender. 

## Changes
- introduce `SNS_SES_SENDER_ADDRESS` config variable to make it configurable, and fallback to `admin@localstack.com` if not configured

## TODO

What's left to do:

- [x] [open a PR](https://github.com/localstack/docs/pull/1582) in the docs to document this variable and the default value

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -3003,6 +3003,49 @@ class TestSNSSubscriptionSES:
 
         retry(check_subscription, retries=PUBLICATION_RETRIES, sleep=PUBLICATION_TIMEOUT)
 
+    @markers.aws.only_localstack
+    def test_email_sender(
+        self,
+        sns_create_topic,
+        sns_subscription,
+        aws_client,
+        monkeypatch,
+    ):
+        # make sure to reset all received emails in SES
+        requests.delete("http://localhost:4566/_aws/ses")
+
+        topic_arn = sns_create_topic()["TopicArn"]
+        sns_subscription(
+            TopicArn=topic_arn,
+            Protocol="email",
+            Endpoint="localstack@yopmail.com",
+        )
+
+        aws_client.sns.publish(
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/sns/publisher.py
```diff
@@ -569,13 +569,16 @@ def _publish(self, context: SnsPublishContext, subscriber: SnsSubscription):
         region = extract_region_from_arn(subscriber["Endpoint"])
         ses_client = connect_to(aws_access_key_id=account_id, region_name=region).ses
         if endpoint := subscriber.get("Endpoint"):
+            # TODO: legacy value, replace by a more sane value in the future
+            #  no-reply@sns-localstack.cloud or similar
+            sender = config.SNS_SES_SENDER_ADDRESS or "admin@localstack.com"
```
alexrashed, line 574: **suggestion:** ‏Would be great to have a "localstack only" test for this newly introduced config value.

alexrashed (approved):
LGTM! 💯 Nice and clean, would be great to have a test though, but that could also be added in a later iteration (not blocking). 😉

localstack-core/localstack/services/sns/publisher.py
bentsku: Very good point, I've added the test now, tell me what you think 😄

alexrashed (approved):
Awesome! Thanks for addressing the comment right away and adding a test! 🚀 💯 :shipit:

PR #11932, opened 2024-11-26
Extend patch decorator to add methods to class types

giograno:
## Motivation
Our `patch` decorator easily allows us to patch a class method and extend its behavior.
However, it does not support the case where a given type does not implement a function and we want to add it.
Patch now would support the following use case:

```python
@patch(MyClass)
def new_method(self, *args):
    # some implementation
    ...

c = MyClass()
c.new_method()
```

If the class already implements a method with the given name, the patch won't have any effect.
Like our current implementation, a patch can be easily undone:

```python
new_method.patch.undo()
c.new_method() # raises AttributeError
```

## Changes
- Extending `patch` to support the use case of adding a new function to a class type;
- Unit tests to verify the behavior of the new use case.

### DIFF
--- a/localstack-core/localstack/utils/patch.py
+++ b/localstack-core/localstack/utils/patch.py
@@ -1,7 +1,7 @@
 import functools
 import inspect
 import types
-from typing import Any, Callable, List
+from typing import Any, Callable, List, Type
 
 
 def get_defining_object(method):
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/patch.py
```diff
@@ -111,6 +119,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.undo()
         return self
 
+    @staticmethod
+    def extend_class(target: Callable, fn: Callable):
```
thrau, line 123: nit: sorry i missed this, but the type hint should by `Type`, right?

thrau (approved):
this is much cleaner, thanks! just one minor nit that i missed, please just amend and merge :-)

PR #11934, opened 2024-11-26
Add UsageMultiSetCounter (up for discussion)

joe4dev:
Extracted from https://github.com/localstack/localstack/pull/11883

## Motivation

Our current `UsageSetCounter` can record events of different categories (i.e., set), but cannot record multi-dimensional data (e.g., source, target). This PR contains a possible implementation for a two-dimensional multi-set counter.

We decided not to move forward with this approach until the data requirements are clearer (e.g., allow for generic analysis/materialization rather than feature-specific solutions that don't scale). Therefore, this PR is parked for now and up for discussion.

## Discussion

The flat record structure (`pipes:invocation:aws:sqs`) is not ideal and would be cumbersome to analyze. 
Something like this might be more suitable, but still not ideal if the namespace is feature-specific:

```json
{"outer_namespace": {"inner_namespace": {"key": "count"}}}
```

## Changes

* Add `UsageMultiSetCounter`
* Add unit test `test_multi_set_counter`

### DIFF
--- a/localstack-core/localstack/utils/analytics/usage.py
+++ b/localstack-core/localstack/utils/analytics/usage.py
@@ -1,5 +1,6 @@
 import datetime
 import math
+import threading
 from collections import defaultdict
 from itertools import count
 from typing import Any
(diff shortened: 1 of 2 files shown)


joe4dev:
Superseeded by new structured metric counters https://github.com/localstack/localstack/pull/12230

PR #11939, opened 2024-11-26
Apigw fix openapi identity sources

cloutierMat:
## Motivation

Our current implementation of the OpenApi import did not add authorization scopes when creating a method with an authorizer. This was not an issue with Legacy apigw, as we did not handle the scopes properly.

Authorization scopes can be added to Cognito authorizers. They directly impact whether the api accepts an Identity token or an Access Token see [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-integrate-with-cognito.html). As Next Gen provider does enforce this, providing a greater compatibility with aws, user's configs would fail to authorize a request when an Access token is used to authenticate.

## Changes

- Get the scopes from the default authorizer or the method if specified
- Added tests

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -4,7 +4,7 @@ import hashlib
 import json
 import logging
 from datetime import datetime
-from typing import List, Optional, Union
+from typing import List, Optional, TypedDict, Union
 from urllib import parse as urlparse
 
 from jsonpatch import apply_patch
@@ -89,6 +89,11 @@ class OpenAPIExt:
     REQUEST_VALIDATOR = "x-amazon-apigateway-request-validator"
     REQUEST_VALIDATORS = "x-amazon-apigateway-request-validators"
     TAG_VALUE = "x-amazon-apigateway-tag-value"
+
+
+class AuthorizerConfig(TypedDict):
+    authorizer: Authorizer
+    authorization_scopes: Optional[list[str]]
 
 
 # TODO: make the CRUD operations in this file generic for the different model types (authorizes, validators, ...)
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/apigateway/helpers.py
```diff
@@ -815,7 +820,7 @@ def create_method_resource(child, method, method_schema):
         kwargs = {}
 
         if authorizer := get_authorizer(method_schema) or default_authorizer:
-            method_authorizer = authorizer or default_authorizer
```
cloutierMat, line 818: @bentsku This seemed to be a redundant operation... Do you see any reason to keep this? Was this a issue with the minifier/obfuscation tool?

bentsku: Seems like this was just being too safe probably? Don't think it's because of the minifier, as this is community and this code isn't minified. Probably just a mistake not caught in a review 😄

tests/aws/services/apigateway/test_apigateway_import.py
```diff
@@ -874,6 +879,13 @@ def test_import_with_cognito_auth_identity_source(
 
         rest_api_id = response["id"]
 
-        # assert that are no multiple authorizers
         authorizers = aws_client.apigateway.get_authorizers(restApiId=rest_api_id)
-        snapshot.match("get-authorizers", authorizers)
+        snapshot.match("get-authorizers", sorted(authorizers["items"], key=lambda x: x["name"]))
```
bentsku, line 883: very minor nit: could also use `itemgetter("name")` here like the `sorted` call below.

bentsku (approved):
LGTM! Nice find! I can't wait for us to rework the Import/Export functionality in the future, as this helper should at least be extracted in its own file, and reworked to be a bit clearer... and more tested 😄 nice and clean fix! 💯

tests/aws/services/apigateway/test_apigateway_import.py
bentsku: Honestly, don't have to change it, the CI is green, let's keep it!

PR #11941, opened 2024-11-27
implement S3 conditional write `IfMatch`

bentsku:
## Motivation
Depends on #11940

AWS just released a new Conditional Write feature with `IfMatch` (also called Compare-and-Swap) with https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-functionality-conditional-writes/

Documented at: https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html

This is a simple feature to implement now that we have everything is place already. Only issue is that we needed the latest ASF specs from the last botocore update. 

This is needed for https://github.com/apache/arrow-rs/issues/6799 and would unblock the following PR: https://github.com/apache/arrow-rs/pull/6802

Behavior is as expected by the documentation, except for `CompleteMultipartUpload` where executing an operation between a `CreateMultipartUpload` and `CompleteMultipartUpload` will trigger a 409 Conflict exception, and only a restart can unblock the situation (this was documented for `IfNoneMatch`, but not for `IfMatch`). 

## Changes
- implement the `IfMatch` parameter and behavior for `PutObject` and `CompleteMultipartUpload`


tests/aws/services/s3/test_s3_api.py
```diff
@@ -1869,3 +1873,252 @@ def test_put_object_if_none_match_versioned_bucket(self, s3_bucket, aws_client,
...
+
+        put_obj_2 = aws_client.s3.put_object(Bucket=s3_bucket, Key=key, Body="test")
+        snapshot.match("put-obj-during", put_obj_2)
+        # same ETag as first put
+        put_obj_etag_2 = put_obj_2["ETag"]
+        assert put_obj_etag_1 == put_obj_etag_2
+
+        # it seems that even if we overwrite the object with the same content, S3 will still reject the request if a
+        # write operation was done between creation and completion of the multipart upload, like the `Delete`
+        # counterpart of `IfNoneMatch`
```
dominikschubert, line 2005: question: because every write changes the ETag I assume?

dominikschubert (approved):
Kudos on immediately jumping on this new feature.
Gave the logic a quick sanity check and didn't find any obvious issues with it :+1: 

Nice attention to edge cases as well :rocket:

tests/aws/services/s3/test_s3_api.py
bentsku: almost, if the content of the object is the same, the ETag does not change, it's most of the time a MD5 hash of the object content. Which is why it's a bit surprising! I also gave a quick write up here: https://github.com/apache/arrow-rs/issues/6799#issuecomment-2502344533

This behavior was documented for `IfNoneMatch`, but not for `IfMatch` it seems, but I suspect it's because they use the same logic/primitive for both.

bentsku:
Thanks a lot for the review @dominikschubert 🙏 I've just pushed a new test because I thought of some edge case I wanted to test, as we never know... 😅 but things are all good, no code change. I'll merge once it's green again! Thanks again for the very quick review!

PR #11944, opened 2024-11-27
fix flaky SES test

bentsku:
## Motivation
Following #11930, we were sending SES emails via SNS, and it incremented the counter. However, we were snapshotting the initial value to `0.0`, which is then sensitive to test run orders and such. 

The test needs fixing in AWS, but I've manually run the update snapshot with only the initial part of the test to still validate the proper format of `GetSendQuota` from AWS. 

The behavior of the test is still correctly validated via asserts (not working in AWS, probably because of sandbox account).

## Changes
- transform the initial value of the usage counter to avoid sensitivity to test order


alexrashed (approved):
Nice catch! Thanks for jumping on this! 🚀

PR #11946, opened 2024-11-27
analytics for non-prefixed env var forwarding in CLI

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/11810 we deprecated the forwarding of not-explicitly-prefixed environment variables from the system environment to the runtime environment.
However, this is definitely a change which will take quite a while to reach all our users.
In order to get an idea where we are, this PR adds a simple analytics log which tracks the environment variables (only those marked as "known" by the CLI and only the name and not the content) which are affected by this deprecation.

## Changes
- Add a simple analytics log message to the deprecation check to help us gain knowledge on the adoption of env var prefix.

### DIFF
--- a/localstack-core/localstack/utils/bootstrap.py
+++ b/localstack-core/localstack/utils/bootstrap.py
@@ -513,6 +513,7 @@ class ContainerConfigurators:
         if config.LOADED_PROFILES:
             load_environment(profiles=",".join(config.LOADED_PROFILES), env=profile_env)
 
+        non_prefixed_env_vars = []
         for env_var in config.CONFIG_ENV_VARS:
             value = os.environ.get(env_var, None)
             if value is not None:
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/bootstrap.py
```diff
@@ -521,17 +522,29 @@ def config_env_vars(cfg: ContainerConfiguration):
...
+                # Show a deprecation warning for each individual env var collected above
+                LOG.warning(
+                    "Non-prefixed environment variable %(env_var)s is forwarded to the LocalStack container! "
+                    "Please use `LOCALSTACK_%(env_var)s` instead of %(env_var)s to explicitly mark this environment variable to be forwarded form the CLI to the LocalStack Runtime.",
+                    {"env_var": non_prefixed_env_var},
+                )
+
+            log.event(
+                event="non_prefixed_cli_env_vars", payload={"env_vars": non_prefixed_env_vars}
+            )
```
dominikschubert, line 546: thought: somewhat unrelated but IMO we should start also writing tests for the analytics events to make sure they work as expected :thinking:

dominikschubert (approved):
LGTM, good idea to track the usage of those variables, so we can estimate the effect of the removal in the future

localstack-core/localstack/utils/bootstrap.py
alexrashed: Definitely, I thought about adding this, but then I postponed this to a follow-up to introduce unit tests for multiple log events with some utilities around it (mocking / asserting).

PR #11954, opened 2024-11-28
Handle null value in SES OpenAPI spec

giograno:
## Motivation
This PR fixes a minor gap in our SES OpenAPI specs.
In particular, the `htlm_part` property in the response to `GET /_aws/ses` can be `null`. 
This issue was detected by @HarshCasper when using the SDK generated from the spec (as the generated Pydantic model does not expect a null value).

## Changes
- Change `html_part` to nullable in the specs.

### DIFF
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -158,7 +158,7 @@ components:
           additionalProperties: false
           properties:
             html_part:
-              type: string
+              type: [string, 'null']
             text_part:
               type: string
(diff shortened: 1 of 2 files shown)


viren-nadkarni (commented):
Thanks for fixing this.

Small ask: could you also update the class definition in `localstack.services.ses.models` before merging?

viren-nadkarni (approved):
Thanks @giograno

HarshCasper (approved):

Issue #11943, opened 2024-11-27
bug: ExceptionType is not acted upon/bubbling up to Catch/Retry definition


derek-cbtw:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When throwing a custom exception (python 3.12), SFN allows you to catch them by name for flow redirection.

Error being thrown:
`
{
--
2 |   | "resourceType": "lambda",
3 |   | "resource": "invoke",
4 |   | "error": "Exception",
5 |   | "cause": {
6 |   | "errorMessage": "Test",
7 |   | "errorType": "MoveableException",
8 |   | "requestId": "ad733d94-fda4-46ff-9e41-6091e79d7de5",
9 |   | "stackTrace": [
10 |   | "  File \"/var/lang/lib/python3.12/site-packages/aws_lambda_powertools/logging/logger.py\", line 447, in decorate\n    return lambda_handler(event, context, *args, **kwargs)\n",
11 |   | "  File \"/var/task/fetch_job.py\", line 25, in lambda_handler\n    status_code, data = processor.fetch(job_data=data, config=config)\n",
12 |   | "  File \"/var/task/processors/base_class.py\", line 23, in wrapper\n    return method(*args, **kwargs)\n",
13 |   | "  File \"/var/task/processors/base_class.py\", line 23, in wrapper\n    return method(*args, **kwargs)\n",
14 |   | "  File \"/var/task/processors/job_processor.py\", line 168, in fetch\n    raise MoveableException(\"Test\")\n"
15 |   | ]
16 |   | }
17 |   | }
`

Relevant definition:
`      
"Catch": [
      {
        "ErrorEquals": [
          "MoveableException"
        ],
        "Next": "MoveJob",
        "ResultPath": "$.fetch_job.output"
      },
        {
          "ErrorEquals": [
            "Exception"
          ],
          "Next": "NotifyFailure",
          "ResultPath": "$.failure.output"
        },
`

Actual flow:
![Execution-Events-LocalStack-11-27-2024_12_41_PM](https://github.com/user-attachments/assets/a953df77-464a-40d1-a05c-c9403859dd78)

**This flow is validated on live AWS environments.**

### Expected Behavior

When a known custom exception is thrown and defined in the SFN definition block, it should act accordingly.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Define a step function definition with simple catch/retry on a custom exception on a lambda invocation with a catch block mirroring the provided above.

[...]

PR #11957, opened 2024-11-28
StepFunctions: Fix ErrorName Value for Lambda Task Errors

MEPalma:
## Motivation
Currently, the SFN v2 interpreter incorrectly assigns the error name "Exception" to Lambda task failures that result in a LambdaFunctionErrorException, rather than utilising the exception's errorType value. This limitation prevents users from catching or retrying Lambda task failures using custom exception values, as reported in [this issue](https://github.com/localstack/localstack/issues/11943). These changes address the error name computation, correcting the behaviour, and introduce relevant tests for both legacy Lambda invocations and service task Lambda invocations.

## Changes
- revised error name computation for legacy task lambda states
- revised error name computation for service task lambda states
- added related snapshot tests

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task_lambda.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task_lambda.py
@@ -1,4 +1,5 @@
 import json
+import logging
 from typing import Union
 
 from botocore.exceptions import ClientError
@@ -40,6 +41,8 @@ from localstack.services.stepfunctions.asl.eval.event.event_detail import EventD
 from localstack.services.stepfunctions.asl.utils.encoding import to_json_str
 from localstack.services.stepfunctions.quotas import is_within_size_quota
 
+LOG = logging.getLogger(__name__)
+
 
 class StateTaskLambda(StateTask):
     resource: LambdaResource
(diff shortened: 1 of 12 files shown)


tests/aws/services/stepfunctions/v2/error_handling/test_task_lambda.py
```diff
@@ -53,6 +48,40 @@ def test_raise_exception(
             exec_input,
         )
 
+    @markers.aws.validated
+    def test_raise_custom_exception(
```
gregfurman, line 52: Should we add a JSONata test to this as well in a follow-up? Curious if the `$states.errorOutput` works here too.

localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task_lambda.py
```diff
@@ -60,12 +63,20 @@ def _from_error(self, env: Environment, ex: Exception) -> FailureEvent:
...
             cause = ex.payload
-        elif isinstance(ex, ClientError):
+            try:
+                cause_object = json.loads(cause)
+                error = cause_object["errorType"]
+            except Exception as ex:
+                LOG.warning(
+                    "Could not retrieve 'errorType' field from LambdaFunctionErrorException object: %s",
+                    ex,
+                )
```
gregfurman, line 78: nit: Could be worth unifying this function and the above to prevent duplication in both the `StateTaskLambda` and `StateTaskServiceLambda`

gregfurman (approved):
LGTM! Nothing that can't be done in a follow-uo

Issue #11950, opened 2024-11-27
bug: Step Functions $$.StateMachine.Name and $$.State.Name are not supported


peter-smith-phd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following state machine fails to parse correctly:

```
{
  "StartAt": "fail",
  "States": {
    "fail": {
      "Type": "Fail",
      "ErrorPath": "$$.State.Name",
      "CausePath": "$$.StateMachine.Name"
    }
  }
}
```

The incorrect output is:
```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: ASLParserException ['line 6:19, at "$$.State.Name", no viable alternative at input \'"ErrorPath":"$$.State.Name"\'', 'line 7:19, at "$$.StateMachine.Name", no viable alternative at input \'"CausePath":"$$.StateMachine.Name"\'']
```

### Expected Behavior

The ASL should parse correctly, and the state machine should return the state's name, and the state machine's name.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

   awslocal stepfunctions create-state-machine --name "Fail_errors" --definition file://file.asl.json --role-arn arn:aws:iam::000000000000:role/Admin

### Environment

```markdown
- OS: Amazon Linux
- LocalStack:
  LocalStack version: 4.0.3.dev22
  localstack-main | LocalStack build date: 2024-11-27
  localstack-main | LocalStack build git hash: 0e01610bb
```

peter-smith-phd:
I also see this code being problematic:
  
  ```
  {
    "Variable?": "$$.State.EnteredTime"
    "TimestampGreaterThanPath": "$.time"
    "Next": "pass1"
  }
```

MEPalma:
Hello @peter-smith-phd, thank you once again for the very report. I believe this issue should be resolved from the next latest version of localstack. Would you mind verifying this scenario works as expected? Thank you!

peter-smith-phd:
Thank you. I've confirmed that `$$.State.Name`, `$$.StateMachine.Name` and `$$.State.EnteredTime` are all supported now.

MEPalma:
[...]

PR #11958, opened 2024-11-28
StepFunctions: Support for ContextObject Paths Bindings for ErrorPath and CausePath

MEPalma:
## Motivation
The SFN v2 interpreter does not support ContextObject bindings for ErrorPath and CausePath fields, which leads to parse errors when these are defined: https://github.com/localstack/localstack/issues/11950. These changes add the relvant grammar and eval support.

## Changes
- updated ASL grammar to accept bindings to ContextObject paths for ErrorPath and CausePath fields
- added relevant preprocessing and evaluation logic
- added relevant snapshot test case

## Future work
As we accumulate more snapshot tests, a pattern is emerging in the evaluation logic that we should consider. We should explore introducing a reusable string evaluation expression into the grammar, capable of handling the parsing and evaluation of string literals, JSONPaths, context object paths, variables, and JSONata expressions. By incorporating preprocessing safeguards against invalid bindings, this approach could streamline the grammar, reduce its size, and eliminate some of the evaluation duplications introduced by these changes.


gregfurman (approved):
LGTM! Curious if we have any other productions that should support ContextObject path bindings 🤔

Issue #11630, opened 2024-10-03
EventBridge Retry Policy Implementation


utajum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

LocalStack should implement the EventBridge retry policy feature as described in the AWS documentation [1].

Currently, in LocalStack version 3.7.2, failed events are never retried, which differs from the behaviour of the actual AWS EventBridge service.

According to the AWS documentation, EventBridge has a retry policy for events that aren't successfully delivered to their targets. The default behaviour is to retry sending the event for 24 hours and up to 185 times with exponential backoff and jitter. 

Implementing this feature in LocalStack would provide a more accurate simulation of the AWS EventBridge service.

My setup
docker-compose:

```
  localstack:
    extra_hosts:
      - "host.docker.internal:host-gateway"
    image: localstack/localstack:3.7.2
    environment:
      DEBUG: ${DEBUG-}
      DOCKER_HOST: unix:///var/run/docker.sock
      SERVICES: s3,secretsmanager,events
      GATEWAY_LISTEN: "0.0.0.0:4567"
      AWS_ENDPOINT_URL: "http://localstack:4567"
      # PROVIDER_OVERRIDE_EVENTS: v2
```

If I use `PROVIDER_OVERRIDE_EVENTS: v2` EventBridge does not work

Setting `RetryPolicy` on `PutTargetsCommand` does nothing.

I have tested the same code with AWS EventBridge and failed attempts are queued and retried.

[1] https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-retry-policy.html

### Anything else?

https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-retry-policy.html

zaingz:
@utajum, thanks for raising the issue! I’ve been looking into the PROVIDER_OVERRIDE_EVENTS specifically but haven’t been able to reproduce the problem. Could you provide some additional context to help me investigate further?
	•	Does the issue still persist?
	•	Is it specific to version 3.7.2? It’s possible it’s resolved in the latest version.
	•	When you say, “If I use PROVIDER_OVERRIDE_EVENTS: v2 EventBridge does not work,” do you mean the flag isn’t being applied, or that retries aren’t functioning as expected?

utajum:
[...]

PR #11959, opened 2024-11-28
Added test and fix for wrong time format and serialization in events v2

zaingz:
## Motivation
Found through an issue: https://github.com/localstack/localstack/issues/11630
When using EventBridge v2, events containing datetime fields fail to be delivered due to JSON serialization error: `Object of type datetime is not JSON serializable`. This affects events with explicit Time fields as well as automatically generated timestamps.

## Changes
- Added `format_event_time` function to properly format datetime objects to ISO8601 strings
- Updated `format_event` to use the formatted time string
- Added test to verify datetime serialization and delivery behavior matches AWS

The fix ensures proper datetime handling in events while maintaining AWS parity by formatting all time values to ISO8601 strings (YYYY-MM-DDThh:mm:ssZ).

### DIFF
--- a/localstack-core/localstack/aws/api/events/__init__.py
+++ b/localstack-core/localstack/aws/api/events/__init__.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from enum import StrEnum
-from typing import Dict, List, Optional, TypedDict
+from typing import Dict, List, Optional, TypedDict, Union
 
 from localstack.aws.api import RequestContext, ServiceException, ServiceRequest, handler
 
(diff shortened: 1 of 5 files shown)


maxhoheiser (approved):
LGTMT - nice implementation - I belive we had a custom json encoder at one point for serialization

localstack-core/localstack/aws/api/events/__init__.py
```diff
@@ -919,7 +919,7 @@ class EventSource(TypedDict, total=False):
 
 
 EventSourceList = List[EventSource]
-EventTime = datetime
+EventTime = Union[datetime, str]
```
maxhoheiser, line 922: maybe use the newer recommended way `datetime | str`

localstack-core/localstack/services/events/utils.py
```diff
@@ -130,6 +130,11 @@ def get_event_time(event: PutEventsRequestEntry) -> EventTime:
     return event_time
 
 
+def format_event_time(event_time: datetime) -> str:
```
maxhoheiser, line 133: why would we need the same function 2 times? just use `event_time_to_time_string` ?

PR #11960, opened 2024-11-28
implement new native EventBridge EventRuleEngine

bentsku:
## Motivation
This PR copy-paste the SNS Filter Policy engine and modify it to be an `EventPatternEngine`. Implementations are close, but differs in quite a few ways. SNS has less features, maybe linked to complexity issues. They also allow way less complexity in rules. 

There a few differences with SNS in almost all methods, could maybe be consolidated by testing all the edge cases for SNS as well, but keeping the implementation different might just be a better idea, even if the base looks similar. Those are 2 different services and could evolve in different directions. 

Question: should it be behind a feature flag? ~Should probably be released after `4.0.3`~. It shouldn't be worse than the current implementation, but could probably have a few regression concerning validation (we might be stricter and wrong). 
edit: as the current Python implementation has a lot of flaws, merging this is most probably beneficial and should go in `4.0.3`, but I'll be on the lookout for any potential regression and jump on it next week if needed.

Note: 
In SNS, the filter policies are validated in a different step than the matching, which is why the current implementation is split into 2 classes. The validation (`EventPatternValidator`) could probably be consolidated into the `EventRuleEngine` for `events`. 

## Changes
- copy-paste the SNS filter policy engine to `events`, modify what was needed and add the missing implementation:
  - `wildcard` operator
  - `cidr` operator (should be ported back to SNS)
  - nested `anything-but` with `suffix`, `equals-ignore-case` and `wildcard` (untested before) 
    - `equals-ignore-case`, `wildcard`, `suffix` and `prefix` list values only when nested under `anything-but`
  - nested `equals-ignore-case` under `suffix` and `prefix` 
- modify the validation engine to be less strict than SNS, might still be too strict, we need to implement more tests in `events`
- add over 30 test cases for missing validation, new implementation has 114/114 🥳 

## Testing
All tests are passing, even the `_EXC` ones, I've added a few more as I had a few unanswered questions by doing the implementation.

Also testing downstream dependencies with a full run at #\2193 (seems to be failing for unrelated changes and #11961), new full run at #\2208 ✅ for AMD64, ARM timeout but surely unrelated

## TODO

What's left to do:

- [x] add the few tests marked with TODO in the implementation
- [x] maybe replace the logic in `matches_event` so that it used by Pipes, but this could be done in a follow up

### DIFF
--- a/localstack-core/localstack/services/events/event_rule_engine.py
+++ b/localstack-core/localstack/services/events/event_rule_engine.py
@@ -1,0 +1,515 @@
+import ipaddress
+import json
+import re
+import typing as t
+
+from localstack.aws.api.events import InvalidEventPatternException
+
+
+class EventRuleEngine:
+    def evaluate_pattern_on_event(self, filter_policy: dict, message_body: str):
+        try:
+            body = json.loads(message_body)
+            if not isinstance(body, dict):
+                return False
+        except json.JSONDecodeError:
+            # Filter policies for the message body assume that the message payload is a well-formed JSON object.
+            # See https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html
+            return False
+
+        return self._evaluate_nested_filter_policy_on_dict(filter_policy, payload=body)
+
+    def _evaluate_nested_filter_policy_on_dict(self, filter_policy, payload: dict) -> bool:
+        """
+        This method evaluates the filter policy against the JSON decoded payload.
+        Although it's not documented anywhere, AWS allows `.` in the fields name in the filter policy and the payload,
+        and will evaluate them. However, it's not JSONPath compatible:
+        Example:
+        Policy: `{"field1.field2": "value1"}`
+        This policy will match both `{"field1.field2": "value1"}` and  {"field1: {"field2": "value1"}}`, unlike JSONPath
+        for which `.` points to a child node.
+        This might show they are flattening the both dictionaries to a single level for an easier matching without
+        recursion.
+        :param filter_policy: a dict, starting at the FilterPolicy
+        :param payload: a dict, starting at the MessageBody
+        :return: True if the payload respect the filter policy, otherwise False
+        """
+        if not filter_policy:
+            return True
+
+        # TODO: maybe save/cache the flattened/expanded policy?
+        flat_policy_conditions = self.flatten_policy(filter_policy)
+        print(f"{flat_policy_conditions=}")
+        flat_payloads = self.flatten_payload(payload)
+        print(f"{flat_payloads=}")
+
+        return any(
+            all(
+                any(
+                    self._evaluate_condition(
+                        flat_payload.get(key), condition, field_exists=key in flat_payload
+                    )
+                    for condition in values
+                    for flat_payload in flat_payloads
+                )
+                for key, values in flat_policy.items()
+            )
+            for flat_policy in flat_policy_conditions
+        )
+
+    def _evaluate_condition(self, value, condition, field_exists: bool):
+        if not isinstance(condition, dict):
+            return field_exists and value == condition
+        elif (must_exist := condition.get("exists")) is not None:
+            # if must_exists is True then field_exists must be True
+            # if must_exists is False then fields_exists must be False
+            return must_exist == field_exists
+        elif value is None:
+            # the remaining conditions require the value to not be None
+            return False
+        elif anything_but := condition.get("anything-but"):
+            if isinstance(anything_but, dict):
+                if not_prefix := anything_but.get("prefix"):
+                    return not value.startswith(not_prefix)
+                elif not_suffix := anything_but.get("suffix"):
+                    return not value.endswith(not_suffix)
+                elif not_equal_ignore_case := anything_but.get("equals-ignore-case"):
+                    if isinstance(not_equal_ignore_case, str):
+                        return not_equal_ignore_case.lower() != value.lower()
+                    elif isinstance(not_equal_ignore_case, list):
+                        return all(value.lower() != v.lower() for v in not_equal_ignore_case)
+
+            elif isinstance(anything_but, list):
+                return value not in anything_but
+            else:
+                return value != anything_but
+
+        elif prefix := condition.get("prefix"):
+            if isinstance(prefix, dict):
+                if prefix_equal_ignore_case := prefix.get("equals-ignore-case"):
+                    return value.lower().startswith(prefix_equal_ignore_case.lower())
+
+            return value.startswith(prefix)
+
+        elif suffix := condition.get("suffix"):
+            if isinstance(suffix, dict):
+                if suffix_equal_ignore_case := suffix.get("equals-ignore-case"):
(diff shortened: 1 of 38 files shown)


tests/aws/services/events/test_events_patterns.py
```diff
@@ -21,40 +22,6 @@
 )
 COMPLEX_MULTI_KEY_EVENT = os.path.join(REQUEST_TEMPLATE_DIR, "complex_multi_key_event.json")
 
-SKIP_LABELS = [
```
joe4dev, line 24: 🤩 🥳

tests/aws/services/events/test_events_patterns.py
```diff
@@ -91,27 +58,38 @@ class TestEventPattern:
         ids=[t[1] for t in request_template_tuples],
     )
     @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(
+        paths=["$..MessageRaw"],  # AWS returns Java validation parts, we skip those
```
joe4dev, line 62: fair limitation 👍

tests/aws/services/events/test_events_patterns.py
```diff
@@ -91,27 +58,38 @@ class TestEventPattern:
...
+                boto_error["Error"]["Message"] = message.split("\n")[0]
+
+            return boto_error
 
         event = request_template["Event"]
         event_pattern = request_template["EventPattern"]
 
         if label.endswith("_EXC"):
-            with pytest.raises(Exception) as e:
+            with pytest.raises(ClientError) as e:
```
joe4dev, line 85: 👍

tests/unit/utils/test_event_matcher.py
```diff
@@ -68,9 +69,19 @@ def test_matches_event_non_matching_pattern():
...
+@pytest.mark.parametrize("engine", ("python", "java"))
+def test_matches_event_invalid_json(event_rule_engine, engine):
     """Test with invalid JSON strings"""
-    with pytest.raises(json.JSONDecodeError):
+    event_rule_engine(engine)
+
+    if engine == "java":
+        # this lets the exception bubble up to the provider, when AWS returns a proper exception, it should be fixed
+        exception_type = json.JSONDecodeError
+        pytest.skip("jpype conflict")
```
joe4dev, line 80: Unfortunately, we cannot risk re-introducing test coverage for the Java engine because it could break StepFunction tests using the new jsonata feature. Can we clarify that implication in a comment here because it's quite implicit knowledge and if anyone doesn't know it, re-enabling these tests risks 💥

joe4dev (approved):
Wow, that's amazing work @bentsku 👏 👏👏 
Thank you very much also for extending the test suite and even fixing the exception cases 😮 

LGTM, just added some small suggestions, but these are totally fine in a follow up. The main part where we need to be careful is not to re-enable the JPype engine here.

tests/unit/utils/test_event_matcher.py
bentsku: But this is only in the unit tests, right? I'll add the re-enabling lower down with a better comment, to be safe.

bentsku: I've moved the `event_rule_engine(engine)` after the skip to avoid enabling it

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -0,0 +1,566 @@
...
+                return False
+        else:
+            body = message_body
+
+        return self._evaluate_nested_event_pattern_on_dict(event_pattern, payload=body)
+
+    def _evaluate_nested_event_pattern_on_dict(self, event_pattern, payload: dict) -> bool:
+        """
+        This method evaluates the event pattern against the JSON decoded payload.
+        Although it's not documented anywhere, AWS allows `.` in the fields name in the event pattern and the payload,
```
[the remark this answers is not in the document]
bentsku: Yes, they also mention that people should not rely on this behavior to stay in the docs, right? I'll add a link.

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -0,0 +1,566 @@
+import ipaddress
+import json
+import re
+import typing as t
+
+from localstack.aws.api.events import InvalidEventPatternException
+
+
+class EventRuleEngine:
+    def evaluate_pattern_on_event(self, event_pattern: dict, message_body: str | dict):
```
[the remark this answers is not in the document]
bentsku: How does it work in ESM and Pipes if the pattern is invalid? I didn't really find it. 
The issue is that if the pattern is invalid, the whole matching logic will fail down the line, and it's part of the "compilation": it does not only validate, but compile the rules, unfolding the `$or` and such. I'll try to document that better

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -0,0 +1,566 @@
...
+
+
+class EventPatternValidator:
+    def __init__(self):
+        self.error_prefix = "Event pattern is not valid. Reason: "
+
+    def validate_event_pattern(self, event_pattern: str | dict) -> dict[str, t.Any]:
+        if isinstance(event_pattern, str):
+            try:
+                event_pattern = json.loads(event_pattern)
```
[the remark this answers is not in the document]
bentsku: Do we have a test for that? I don't think the string-based matching is currently implemented in this engine at all? it is expecting `dict` all the way. Could you link me to somewhere I could validate this? 

But yes a plain string is a valid JSON, however it will fail anyway because we are expecting and validating the event is a dict. I believe in Pipes, it should not use this engine if both body and pattern are string, this should only be used to evaluate dicts. 

Happy to jump on a call to discuss this

bentsku (commented):
Thank you @joe4dev for the thorough review! I've addressed all the comments, just the one about the string matching that I'm not sure how to address. Should we look at it quickly?

localstack-core/localstack/services/events/event_rule_engine.py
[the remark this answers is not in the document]
bentsku: I've renamed the `Validator` into `Compiler` but it's closer to what it really does, it translates the rules into something the engine can properly interprets. I guess that for ESM/Pipes, those rules would need to be evaluated only once, and could be cached then, instead of calling `matches_event` directly that way. I'm sure we could have some improvements, we should look at it in a follow up!

localstack-core/localstack/services/events/event_rule_engine.py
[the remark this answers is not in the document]
bentsku: after discussing this on a call, we came to the conclusion that what AWS is calling "plain string" is actually the `body` (or sometimes another field like `data` for Kinesis). You can provider valid JSON inside the body of your SQS message, and ESM/Pipes will automatically decode it for you before passing it down to the EventRuleEngine (or `event-ruler` for AWS), so that you can access nested fields.

This should be part of ESM/Pipes pollers (as it's different for every service) and not the responsibility of the engine, so this is resolved! ✅ thanks a lot @joe4dev for the quick call, we got a lot of great insights! 🙏

PR #11961, opened 2024-11-29
fix ESM analytics

bentsku:
## Motivation
It seems #11883 has broken downstream dependencies, related to Kafka not having the same format (sending a dict when the signature is `list[dict]`). Tests are failing and are not validated against AWS, so I suppose we're supposed to be able to accept `dict` directly to `process_events_batch` as `input_events`. 

## Changes
- add more safety around trying to access `first_event`


thrau (approved):
LGTM!

PR #11962, opened 2024-11-29
Cloudformation implement ssm param read

cloutierMat:
## Motivation

Implementation of `read` along with fixing `list` for ssm parameters. While fixing I realised that the generated specs were out of date. So I recreated them, highlighting an issue with the previous list operation.

## Changes

- update `create` method to return the `read` operation. (how can we test this?)
- added `read` operation
- updated `list` operation

### DIFF
--- a/localstack-core/localstack/services/ssm/resource_providers/aws_ssm_parameter.py
+++ b/localstack-core/localstack/services/ssm/resource_providers/aws_ssm_parameter.py
@@ -213,11 +213,10 @@ class SSMParameterProvider(ResourceProvider[SSMParameterProperties]):
         request: ResourceRequest[SSMParameterProperties],
     ) -> ProgressEvent[SSMParameterProperties]:
         resources = request.aws_client_factory.ssm.describe_parameters()
-        discoverable_properties = ["Name"]
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
             resource_models=[
-                SSMParameterProperties(**util.select_attributes(resource, discoverable_properties))
+                SSMParameterProperties(Name=resource["Name"])
                 for resource in resources["Parameters"]
             ],
         )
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/utils/aws/client_types.py
```diff
@@ -264,6 +264,7 @@ class ServicePrincipal(str):
     """
 
     apigateway = "apigateway"
+    cloudformation = "cloudformation"
```
cloutierMat, line 267: This is to be used with cloudcontrol. https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/resource-operations.html#resource-operations-permissions

localstack-core/localstack/services/ssm/resource_providers/aws_ssm_parameter.py
```diff
@@ -203,6 +223,7 @@ def list(
         return ProgressEvent(
             status=OperationStatus.SUCCESS,
             resource_models=[
-                SSMParameterProperties(Id=resource["Name"]) for resource in resources["Parameters"]
+                SSMParameterProperties(Name=resource["Name"])
+                for resource in resources["Parameters"]
```
cloutierMat, line 227: @Pive01 Will this create an issue with the resource overview? If so please let me know so that I don't break it for you! :wink:

cloutierMat: Actually, I was just watching your demo and realised that the ssm parameter didn't. So I realised that they are actually broken and this changes fixes it. We should probably try to push to merge this before the patch release!

Pive01: Thanks @cloutierMat , it does indeed fix an issue as the identifier was not set correctly...now is properly discovered :rocket:

dominikschubert (approved):
Thanks for spotting and addressing this so quickly @cloutierMat !

It also adds some pressure on a fairly old TODO item regarding an ASF-update-like CFn spec update automation :grin:

PR #11963, opened 2024-11-29
[SFN] Add usage analytics for Variable Workflow and JSONata features

gregfurman:
## Motivation
This PR adds analytics  the SFN provider to help ascertain usage of the new JSONata and Variable workflows. 

## Changes
- Add a static analyser that will check for:
  -  Any variable sampling or assigning of variables
  - The query language being used by the state machine definition
- Metrics have been added that will record whether a sfn uses JSONata or variable workflows, respectively. 

## Testing
- Added unit tests to ensure the static analysis is correct.

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
@@ -1,0 +1,56 @@
+import logging
+
+import localstack.services.stepfunctions.usage as UsageMetrics
+from localstack.services.stepfunctions.asl.antlr.runtime.ASLParser import ASLParser
+from localstack.services.stepfunctions.asl.component.common.query_language import (
+    QueryLanguageMode,
+)
+from localstack.services.stepfunctions.asl.static_analyser.static_analyser import StaticAnalyser
+
+LOG = logging.getLogger(__name__)
+
+
+class UsageMetricsStaticAnalyser(StaticAnalyser):
+    @staticmethod
+    def process(definition: str) -> "UsageMetricsStaticAnalyser":
+        analyser = UsageMetricsStaticAnalyser()
+        analyser.analyse(definition=definition)
+
+        try:
+            if analyser.has_jsonata:
+                UsageMetrics.jsonata_create_counter.increment()
+            else:
+                UsageMetrics.jsonpath_create_counter.increment()
+        except Exception as e:
+            LOG.warning(
+                "Failed to record metrics for StepFunctions QueryLanguage usage",
+                exc_info=e,
+            )
+
+        try:
+            if analyser.has_variable_sampling:
+                UsageMetrics.variables_create_counter.increment()
+        except Exception as e:
+            LOG.warning(
+                "Failed to record usage metrics for StepFunctions Variable Sampling usage",
+                exc_info=e,
+            )
+
+        return analyser
+
+    def __init__(self):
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/stepfunctions/usage.py
```diff
@@ -0,0 +1,20 @@
...
+jsonpath_create_counter = UsageCounter("stepfunctions:jsonpath:create")
+
+# Count of StepFunctions being created that use Variable Sampling or the Assign block
+variables_create_counter = UsageCounter("stepfunctions:variables:create")
+
+# Successful invocations (also including expected error cases in line with AWS behaviour)
+invocation_counter = UsageCounter("stepfunctions:invocation")
+
+# Unexpected errors that we do not account for
+error_counter = UsageCounter("stepfunctions:error")
```
gregfurman, line 20: Currently unused. Will be implemented in a follow-up

localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
```diff
@@ -0,0 +1,56 @@
...
+
+        return analyser
+
+    def __init__(self):
+        super().__init__()
+        self.has_jsonata: bool = False
+        self.has_variable_sampling = False
+
+    def visitQuery_language_decl(self, ctx: ASLParser.Query_language_declContext):
+        query_language_mode_int = ctx.children[-1].getSymbol().type
```
MEPalma, line 47: You can probably prune this logic if the flag is already set

MEPalma (approved):
Looking good from a SFN standpoint!

localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
```diff
@@ -0,0 +1,56 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class UsageMetricsStaticAnalyser(StaticAnalyser):
+    @staticmethod
+    def process(definition: str) -> "UsageMetricsStaticAnalyser":
+        analyser = UsageMetricsStaticAnalyser()
+        analyser.analyse(definition=definition)
+
+        try:
```
joe4dev, line 19: Do we need the try/catch here? `has_jsonata` is just a getter and we can expect the counters to work (we're not using try/catch anywhere else)
if we want to be as defensive, then we should rather cover the `analyse` step.

joe4dev (approved):
yeah data 👍 

I would skip the try/catch on the metric counter unless we have a good reason from the SFN analyser side.

localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
gregfurman: We didn't want the metrics collection to be the reason why SFN was failing. Will just surround the whole block in a try-catch rather

gregfurman: OK wrapped the whole process in a try-except

localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
gregfurman: Done!

PR #11966, opened 2024-11-29
fix: follow-up for null value in SES response

giograno:
## Motivation
This PR is a follow-up fix for https://github.com/localstack/localstack/pull/11954.
It turns out that our code generator needs the change in this PR to output the right Pydantic model, i.e.,

```python
class SesSentEmailBody(BaseModel):
    """
    SesSentEmailBody
    """ # noqa: E501
    html_part: Optional[StrictStr] = None
    text_part: StrictStr
```

Setting the type to `null` did not lead to the wanted generated code (i.e., `html_part` won't be optional).

## Changes
- Minor fix of the specs;

### DIFF
--- a/localstack-core/localstack/openapi.yaml
+++ b/localstack-core/localstack/openapi.yaml
@@ -158,11 +158,10 @@ components:
           additionalProperties: false
           properties:
             html_part:
-              type: [string, 'null']
+              type: string
             text_part:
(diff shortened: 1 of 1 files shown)


bentsku (commented):
LGTM! thanks for fixing this 👌

PR #11967, opened 2024-11-29
APIGW: fix parsing invalid JSON template in MOCK integration

cloutierMat:
## Motivation
When deploying a CDK stack with an APIGW REST API with "automatic" CORS management by APIGW, CDK will create an OPTIONS `method` with a MOCK integration. The problem is that this integration has a request template equal to `"application/json": "{statusCode: 200}"`, which is invalid JSON. 

From the documentation, nothing lets you think that invalid JSON would be accepted, as it needs the `statusCode` field. 
But it does, they don't JSON parse but use a custom token parser that we're tried to test the limit off. We're settling now on this basic implementation which still has flaws, but cover quite some ground already, as shown by the unit tests. We also updated an integration test to specifically cover the request template CDK was using. 

This custom parser will only be executed if the MOCK integration does not return valid JSON, meaning every existing use case will still work, and only people with very, very esoteric negative testing might be impacted by this new custom parser accepting data AWS wouldn't (but we already covered most of the bad negative cases). 

## Changes
- create a small (bad) parser to try to get and validate the status code from MOCK integration
- add a unit test with some weird cases
- update an integration test for the CDK use-case

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
@@ -78,6 +78,10 @@ class RestApiMockIntegration(RestApiIntegration):
                 assert k
                 assert v
 
+                if k == "statusCode":
+                    statuscode = int(v)
+                    continue
+
                 if (first_char := k[0]) in "[{":
                     raise Exception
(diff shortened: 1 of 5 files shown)


bentsku (approved):
LGTM! Let's unblock this for the demo!

PR #11968, opened 2024-11-30
apigw clean up the invalid json parser

cloutierMat:
## Motivation

In an attempt to move fast to fix the issue, with @bentsku. we created #11967.

The naming was done hastily and we will attempt to clean it up a little in this pr.

## Changes

- clean up naming
- simplify logic
- adding a few tests

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
@@ -66,45 +66,37 @@ class RestApiMockIntegration(RestApiIntegration):
         try:
             statuscode = ""
             matched = re.match(r"^\s*{(.+)}\s*$", body).group(1)
-            splits = [m.strip() for m in matched.split(",")]
+            pairs = [m.strip() for m in matched.split(",")]
             # TODO this is not right, but nested object would otherwise break the parsing
-            kvs = [s.split(":", maxsplit=1) for s in splits]
-            for kv in kvs:
-                assert len(kv) == 2
-                k, v = kv
-                k = k.strip()
-                v = v.strip()
+            key_values = [s.split(":", maxsplit=1) for s in pairs]
+            for key_value in key_values:
+                assert len(key_value) == 2
+                key, value = [el.strip() for el in key_value]
+                key = key or '""'
+                value = value or '""'
 
-                assert k
-                assert v
+                assert (leading_key_char := key[0]) not in "[{"
+                if leading_key_char in "'\"":
+                    assert len(key) >= 2
+                    assert key[-1] == leading_key_char
 
-                if k == "statusCode":
-                    statuscode = int(v)
+                if key == "statusCode":
+                    statuscode = int(value)
                     continue
 
-                if (first_char := k[0]) in "[{":
-                    raise Exception
-                if first_char in "'\"":
-                    assert len(k) > 2
-                    assert k[-1] == first_char
-                    k = k[1:-1]
-
-                if (v_first_char := v[0]) in "[{'\"":
-                    assert len(v) > 2
-                    if v_first_char == "{":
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
```diff
@@ -66,45 +66,37 @@ def parse_invalid_json(self, body: str) -> dict:
...
+                key = key or '""'
+                value = value or '""'
+
+                assert (leading_key_char := key[0]) not in "[{"
+                if leading_key_char in "'\"":
+                    assert len(key) >= 2
+                    assert key[-1] == leading_key_char
+
+                if key == "statusCode":
+                    statuscode = int(value)
```
bentsku, line 84: I think we had the mistake in our current implementation too, but here we directly check against the `stateCode` value without regards to single or double quotes it seems (except of the leftover second `if` check)

I think we need to still address that 🤔

tests/unit/services/apigateway/test_mock_integration.py
```diff
@@ -61,6 +61,9 @@ def test_custom_parser(self, create_default_context):
             "{statusCode: 200, foo}: [ba r]}",
             "{statusCode: 200, foo'}: [ba r]}",
             "{statusCode: 200, }foo: [ba r]}",
+            "{statusCode: 200, }foo: ''}",
+            '{statusCode: 200, " ": " "}',
+            '{statusCode: 200, "": ""}',
```
bentsku, line 66: we could also add a few cases with `statusCode` in single and double quotes, and maybe a few invalid with non closing quotes or something similar?

bentsku (commented):
Thanks a lot for cleaning up our... best? code! 😄 I think we still can improve the quote handling of the key, especially to match the `statusCode` value, but then we should be good to go! Thanks a lot again 🙏

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
cloutierMat: oh yes! I must have been tired when I wrote made the cleanup! The goal of moving it under the quotes assertions, was to solve this very issue! But it isn't removing the quotes anymore as we originally had :facepalm:

tests/unit/services/apigateway/test_mock_integration.py
cloutierMat: Good thinking! that is indeed the most important objective :rofl:

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
```diff
@@ -63,48 +63,46 @@ def parse_invalid_json(self, body: str) -> dict:
...
-                assert v
-
-                if k == "statusCode":
-                    statuscode = int(v)
+            key_values = [s.split(":", maxsplit=1) for s in pairs if s]
+            for key_value in key_values:
+                assert len(key_value) == 2
+                key, value = [el.strip() for el in key_value]
+                key = convert_null_value(key)
+                value = convert_null_value(value)
```
bentsku, line 82: nit: could probably be simplified with:
```suggestion
                key, value = [convert_null_value(el.strip()) for el in key_value]
```

I think we could even put the `strip()` call inside `convert_null_value`.

bentsku (approved):
LGTM! Just a minor simplification, but this is already much cleaner than before 🙏 Thanks a lot for addressing the comments and adding test cases! Really, this is such a weird case...

localstack-core/localstack/services/apigateway/next_gen/execute_api/integrations/mock.py
bentsku: Oops, missed it but we could probably add type hints for `convert_null_value(value: str) -> str:`

PR #11974, opened 2024-12-03
fix postgres install in pro test pipeline

alexrashed:
## Motivation
This PR fixes the [Community Integration Tests against Pro](https://github.com/localstack/localstack/actions/workflows/tests-pro-integration.yml) which are meant to inform us about changes that break our downstream project early.
This pipeline failed with the recent upgrade of the GitHub Runner to Ubuntu 24.04.

## Changes
- Removes the installation of `postgres` and related tools because they should not be necessary when running the Community Tests against Pro, and if they are needed they should be installed on demand by the respective package installer.

### DIFF
--- a/.github/workflows/tests-pro-integration.yml
+++ b/.github/workflows/tests-pro-integration.yml
@@ -234,8 +234,7 @@ jobs:
       - name: Install OS packages
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


dominikschubert (approved):
Great that we can remove this (and the previously added pinned version) :rocket:

PR #11975, opened 2024-12-03
upgrade outdated github action in ASF update workflow

alexrashed:
## Motivation
This PR upgrades the version of `mad9000/actions-find-and-replace-string`.
This action is somehow not updated by dependabot, which is why we are doing it manually to avoid issues in the future (GitHub is has deprecated Node 16 support a long time ago):
image

## Changes
- Upgrades `mad9000/actions-find-and-replace-string` to version 5


dominikschubert (approved):
LGTM!

Question: do we know why dependabot is not updating the version of this action?

dominikschubert (approved):

alexrashed:
That's a really good question, I tried to dig into this issue, but couldn't find anything.
I filed an issue in the action repo to follow up on this: https://github.com/mad9000/actions-find-and-replace-string/issues/14

PR #11977, opened 2024-12-03
[ESM] Handle Lambda TCP socket connection timeouts

gregfurman:
## Motivation
We need to handle the case where a synchronous Lambda invocation runs for an extended period of time -- ensuring the TCP socket remains open and alive.

See https://github.com/boto/boto3/issues/2424

## Changes
- Adds defaults for `connect_timeout`, `read_timeout`, `tcp_keepalive`, and `retries` in the client for the ESM lambda sender 
- Since a Lambda can run for 900s, we will only throw an exception once this timeout is reached.
- In addition, we specify `tcp_keepalive=True` which ensures the TCP connection does not drop prior to timeout.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
@@ -36,6 +36,10 @@ from localstack.services.lambda_.event_source_mapping.pollers.sqs_poller import 
 from localstack.services.lambda_.event_source_mapping.senders.lambda_sender import LambdaSender
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.aws.client_types import ServicePrincipal
+from localstack.utils.lambda_debug_mode.lambda_debug_mode import (
+    DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS,
+    is_lambda_debug_mode,
+)
 
 
 class PollerHolder:
@@ -57,12 +61,22 @@ class EsmWorkerFactory:
     def get_esm_worker(self) -> EsmWorker:
         # Sender (always Lambda)
         function_arn = self.esm_config["FunctionArn"]
+
+        if is_lambda_debug_mode():
+            timeout_seconds = DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS
+        else:
+            # 900s is the maximum amount of time a Lambda can run for.
+            lambda_max_timeout_seconds = 900
+            invoke_timeout_buffer_seconds = 5
+            timeout_seconds = lambda_max_timeout_seconds + invoke_timeout_buffer_seconds
+
         lambda_client = get_internal_client(
             arn=function_arn,  # Only the function_arn is necessary since the Lambda should be able to invoke itself
             client_config=botocore.config.Config(
-                retries={"max_attempts": 0, "total_max_attempts": 1},
-                read_timeout=900,  # 900s is the maximum amount of time a Lambda can run for
-                connect_timeout=900,
+                retries={
+                    "total_max_attempts": 1
+                },  # Disable retries, to prevent re-invoking the Lambda
+                read_timeout=timeout_seconds,
                 tcp_keepalive=True,
             ),
         )
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
@@ -1,5 +1,7 @@
 from typing import Callable
 
+import botocore.config
+
 from localstack.aws.api.lambda_ import (
     EventSourceMappingConfiguration,
     FunctionResponseType,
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker_factory.py
```diff
@@ -57,6 +59,12 @@ def get_esm_worker(self) -> EsmWorker:
         function_arn = self.esm_config["FunctionArn"]
         lambda_client = get_internal_client(
             arn=function_arn,  # Only the function_arn is necessary since the Lambda should be able to invoke itself
+            client_config=botocore.config.Config(
+                retries={"max_attempts": 0, "total_max_attempts": 1},
+                read_timeout=900,  # 900s is the maximum amount of time a Lambda can run for
+                connect_timeout=900,
```
dominikschubert, line 65: suggestion: I'd say we should make this configurable. While debugging for example it would be quite nice to not have it drop the connection.

WDYT @joe4dev @dfangl

gregfurman: Not sure if we should make the actual timeouts configurable -- since we set the Lambda debug limit to 3600 at `DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS` so allowing for a number higher than this seems confusing: https://github.com/localstack/localstack/blob/414a968e7404a9c190d519f9011ead6845d9622c/localstack-core/localstack/utils/lambda_debug_mode/lambda_debug_mode.py#L10

We could check if debug mode and pass in that timeout default timeout seconds as the timeout value, similar to how we configure the boto client's config here:
https://github.com/localstack/localstack/blob/414a968e7404a9c190d519f9011ead6845d9622c/localstack-core/localstack/services/lambda_/invocation/executor_endpoint.py#L208-L213

joe4dev: @gregfurman Adopting the debug mode is a great suggestion here 👍 
I suggest adding a buffer similar to the executor endpoint so that we don't terminate Lambdas running for exactly 900s. This buffer should account for potentially slow container startup and small LS processing delays.

gregfurman: OK awesome. Not sure the buffer is necessary here though since we set the `connect_timeout` to 900s -- giving our internal poller client up to 900s to establish a connectiom to the container (which IMO is more than enough time for the container to start up + processing delays).

Assuming we manage to establish a connection within that 900s timeframe, the synchronous invoke we triggered will then have an additional 900s from the `read_timeout` while awaiting for a response.

So, in actuality, this approach allow us a window of 1800s (or 30 minutes) from invoke -> response.

Perhaps I am misunderstanding these timeouts though, so some clarification here could be useful.

From the[ boto docs](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html):
> **connect_timeout** (float or int) – The time in seconds till a timeout exception is thrown when attempting to make a connection. The default is 60 seconds.
> 
> **read_timeout** (float or int) – The time in seconds till a timeout exception is thrown when attempting to read from a connection. The default is 60 seconds.

joe4dev: good point to distinguish these two:

`connect_timeout`: I think that can be low (e.g., `5s`) because it only involves connecting to our gateway running on the same machine.
`read_timeout`: I think that's the relevant one which includes LS gateway processing, container startup (up to `LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT` plus small Docker client overhead), and function execution time (up to 900s).

gregfurman: I don't think we shouldn't be retrying since that could cause unintended side-effects since retrying will cause the `invoke` request  to be sent multiple times.

Currently, the config I added `retries={"max_attempts": 0, "total_max_attempts": 1}` will disable retries on failing requests (albeit redundantly since `total_max_attempts` takes presedence over `max_attempts`).

gregfurman: According to the ESM docs, there are some cases where we'll want to do retries on failed invocation (with backoff). 

From [the docs](https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html#services-sqs-backoff-strategy):

> When an invocation fails, Lambda attempts to retry the invocation while implementing a backoff strategy. The backoff strategy differs slightly depending on whether Lambda encountered the failure due to an error in your function code, or due to throttling.
> 
> - If your function code caused the error, Lambda will stop processing and retrying the invocation. In the meantime, Lambda gradually backs off, reducing the amount of concurrency allocated to your Amazon SQS event source mapping. After your queue's visibility timeout runs out, the message will again reappear in the queue.
> 
> - If the invocation fails due to throttling, Lambda gradually backs off retries by reducing the amount of concurrency allocated to your Amazon SQS event source mapping. Lambda continues to retry the message until the message's timestamp exceeds your queue's visibility timeout, at which point Lambda drops the message.

We don't currently handle throttling and AFAIK our internal SQS implementation does not either. We probably should though, since it seems we could start leverage boto's `adaptive` rate limiting functionality.

joe4dev: yes, +1 on no retries here 👍 

+1 for extra error handling -> backlog

gregfurman: I'm adding this config to only the ESM worker when the client gets created. We can see how it performs and perhaps consider setting some of these values by default in the `get_internal_client` function?

gregfurman: Also, I removed setting the `connection_timeout` to allow it to default to 60s 🙂

joe4dev (approved):
Great catch and kudos for also considering the Lambda debug mode 💯 

Let's get some customer feedback here and consider the lessons learned in Pipes as well.

PR #11979, opened 2024-12-03
SNS: improve SNS Filter Policy engine and implement `cidr` operator

bentsku:
## Motivation
When implementing the EventRuleEngine with #11960, I've implemented different and new behaviors from SNS, and some felt like we just missed them in SNS and were better covered by the event test suite. 

This PR implements the [`cidr` operator](https://docs.aws.amazon.com/sns/latest/dg/string-value-matching.html#string-address-matching), and a missed edge case where one of the field of the payload is an empty array. Also adds a bit of validation around empty list passed to operators. 

Sorry @maxhoheiser and @joe4dev, I've sneaked in a new test case for the EventRuleEngine here as I wanted to verify a very edge case behavior in both engine when the value is an empty array, making the field not exists, which is an edge case we're also having, which is somewhat funny. I didn't want to open a new PR just for this tiny test case 😅 hope that's ok! no change in behavior as we were already right! For the anecdote, this is because I think `exists` can only be applied to "leaf" nodes, and an empty array will make the field not be a leaf anymore, as we're going "into" the array. Funny enough, we have the same edge case 😄 

## Changes
- adds the `cidr` operator and validation
- implements the empty array in payload edge case
- implements empty array in operator validation 
- adds new test cases for SNS
- add one test case for an edge case in `events`

### DIFF
--- a/localstack-core/localstack/services/sns/filter.py
+++ b/localstack-core/localstack/services/sns/filter.py
@@ -126,8 +126,11 @@ class SubscriptionFilter:
         elif numeric_condition := condition.get("numeric"):
             return self._evaluate_numeric_condition(numeric_condition, value)
         elif cidr := condition.get("cidr"):
-            ips = [str(ip) for ip in ipaddress.IPv4Network(cidr)]
-            return value in ips
+            try:
+                ip = ipaddress.ip_address(value)
+                return ip in ipaddress.ip_network(cidr)
+            except ValueError:
+                return False
 
         return False
 
@@ -424,12 +427,7 @@ class FilterPolicyValidator:
                     self._validate_numeric_condition(value)
 
                 elif operator == "cidr":
-                    try:
-                        ipaddress.IPv4Network(value)
-                    except ValueError:
-                        raise InvalidParameterException(
-                            f"{self.error_prefix}FilterPolicy: Malformed CIDR, one '/' required"
-                        )
+                    self._validate_cidr_condition(value)
 
                 else:
                     raise InvalidParameterException(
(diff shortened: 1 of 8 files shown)


localstack-core/localstack/services/sns/filter.py
```diff
@@ -124,6 +125,10 @@ def _evaluate_condition(self, value, condition, field_exists: bool):
             return equal_ignore_case.lower() == value.lower()
         elif numeric_condition := condition.get("numeric"):
             return self._evaluate_numeric_condition(numeric_condition, value)
+        elif cidr := condition.get("cidr"):
+            ips = [str(ip) for ip in ipaddress.IPv4Network(cidr)]
+            return value in ips
```
cloutierMat, line 130: That seems super inefficient. :thinking: If the mask is `10.0.0.0/8` that would create a huge list. :eyes: 

If we are sure it will always be an ipv4 address we can maybe do that `return ipaddress.IPv4Address(value) in ipaddress.IPv4Network(cidr)` and something like `return ipaddress.ip_address(value) in ipaddress.ip_network(cidr)` would cover ipv6 also, I believe

localstack-core/localstack/services/sns/filter.py
```diff
@@ -245,6 +250,8 @@ def _traverse(_object: dict, array=None, parent_key=None) -> list:
                     array = _traverse(values, array, _parent_key)
 
             elif isinstance(_object, list):
+                if not _object:
+                    return array
```
cloutierMat, line 254: Neat find!

cloutierMat (commented):
Great find on the empty array edge case. I am glad our recursive implementation allowed you to fix in such an simple way. Love the new addition to the test and the reusable method to get the sqs message. It cleans up a lot.

I am not fond of creating an array with a gazillion items in it to be discarded right away? Do you know if IPv6 can be used, as that could potentially create even greater arrays.

localstack-core/localstack/services/sns/filter.py
bentsku: Right, yes. Well, we can't be sure the `value` is a valid ipv4 or v6. Actually, looking at the docs, it seems it should work with ipv6, so this doesn't really work either. I'll update `events` as well to add IPv6 support. 

I'll need to wrap this in a try/except block probably, and do what you proposed.

bentsku: time to port back the fix to `events` 😂 thanks, always pushing further! 💪 always thankful for your reviews!

localstack-core/localstack/services/sns/filter.py
```diff
@@ -424,12 +427,7 @@ def _validate_rule(self, rule: t.Any) -> None:
                     self._validate_numeric_condition(value)
 
                 elif operator == "cidr":
-                    try:
-                        ipaddress.IPv4Network(value)
-                    except ValueError:
-                        raise InvalidParameterException(
-                            f"{self.error_prefix}FilterPolicy: Malformed CIDR, one '/' required"
-                        )
+                    self._validate_cidr_condition(value)
```
cloutierMat, line 430: Beauty!

cloutierMat (approved):
Great! Thanks for addressing the comment! Good new test additions as well! Super to get so much parity in the error messages as well

PR #11981, opened 2024-12-03
Events: EventRuleEngine: improve `cidr` to support ipv6 and better validation

bentsku:
## Motivation
After backporting `cidr` back to SNS, @cloutierMat made a really solid point in his review: 
https://github.com/localstack/localstack/pull/11979#discussion_r1868383162

After improving it for SNS, I'm backporting it back again to `events`

## Changes
- add `cidr` support for IPv6, and better validation


maxhoheiser (approved):
Great changes - I did not even know that IP addresses can be used as conditions for rules

PR #11984, opened 2024-12-04
fix SNS test with region replacement

bentsku:
## Motivation
New tests implemented with #11979 had quite the flaw, which was hardcoded region values in the event, even if not relevant to the test itself. 

This PR removes the region values. 

## Changes
- remove the region values in the test and make sort the snapshot as the queue is not FIFO

### DIFF
--- a/tests/aws/services/sns/test_sns_filter_policy.py
+++ b/tests/aws/services/sns/test_sns_filter_policy.py
@@ -1181,7 +1181,7 @@ class TestSNSFilterPolicyBody:
             "source": "aws.autoscaling",
             "account": "123456789012",
             "time": "2015-11-11T21:31:47Z",
-            "region": "us-east-1",
+            "region": "my-region",
             "resources": [],
             "detail": {
                 "eventVersion": "",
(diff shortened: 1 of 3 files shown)


tests/aws/services/sns/test_sns_filter_policy.py
```diff
@@ -1238,34 +1238,34 @@ def test_filter_policy_ip_address_condition(
                 "source": "test-source",
                 "detail-type": "test-detail-type",
                 "account": "123456789012",
-                "region": "us-east-2",
+                "region": "my-region",
```
alexrashed, line 1241: Just to add some context: The `us-east-2` has been replaced with `<my-region>` automatically by the snapshot fixture due to this default configuration:
https://github.com/localstack/localstack/blob/b182caed487c050ac48d4d75c43e6ca238b00485/tests/aws/conftest.py#L114

This means that by default any values equal to the currently used region when executing the test is replaced with `<my-region>`. Since this value was static and did not change even though the region of the test changed (in our MA/MR pipeline), the value was _not_ replaced with `<my-region>` but stayed the same and therefore failed on the matching.

alexrashed (approved):
Thanks for jumping on this! 💯 🦸🏽

PR #11988, opened 2024-12-04
enable localstack instance creation override for testing

cloutierMat:
## Motivation

To enable testing for the replicator, we need a mechanism by which we can start a localstack instance when running test against AWS. This change will allow us to spawn an instance while respecting the devs configuration without having to re:invent the wheel.

## Changes

- Look at env `TEST_FORCE_LOCALSTACK_START` to create instance even if `is_aws_cloud()` is true

### DIFF
--- a/localstack-core/localstack/testing/pytest/in_memory_localstack.py
+++ b/localstack-core/localstack/testing/pytest/in_memory_localstack.py
@@ -53,8 +53,12 @@ def pytest_runtestloop(session: Session):
 
     from localstack.testing.aws.util import is_aws_cloud
 
-    if is_env_true("TEST_SKIP_LOCALSTACK_START") or is_aws_cloud():
+    if is_env_true("TEST_SKIP_LOCALSTACK_START"):
         LOG.info("TEST_SKIP_LOCALSTACK_START is set, not starting localstack")
+        return
+
+    if is_aws_cloud() and not is_env_true("TEST_FORCE_LOCALSTACK_START"):
+        LOG.info("Test running against aws, not starting localstack")
         return
 
     from localstack.utils.common import safe_requests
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/testing/pytest/in_memory_localstack.py
```diff
@@ -53,10 +53,14 @@ def pytest_runtestloop(session: Session):
 
     from localstack.testing.aws.util import is_aws_cloud
 
-    if is_env_true("TEST_SKIP_LOCALSTACK_START") or is_aws_cloud():
+    if is_env_true("TEST_SKIP_LOCALSTACK_START"):
         LOG.info("TEST_SKIP_LOCALSTACK_START is set, not starting localstack")
```
dominikschubert, line 57: :+1: I like that this log statement is now also not mixing two concepts anymore

localstack-core/localstack/testing/pytest/in_memory_localstack.py
```diff
@@ -53,10 +53,14 @@ def pytest_runtestloop(session: Session):
...
         return
 
+    if is_aws_cloud() and not is_env_true("TEST_FORCE_LOCALSTACK_START"):
+        LOG.info("Test running against aws, not starting localstack")
+        return
```
dominikschubert, line 62: nit: while it's a bit more code, I'd argue it would be cleaner here as well to have a separate condition just for the `TEST_FORCE_LOCALSTACK_START` case to log *why* we're starting localstack even though we're targeting AWS

dominikschubert (approved):
LGTM, please add also add this to the testing docs though! 

=> https://github.com/localstack/localstack/blob/master/docs/testing/integration-tests/README.md

localstack-core/localstack/testing/pytest/in_memory_localstack.py
cloutierMat: Good idea! I will add that

HarshCasper (approved):

PR #11989, opened 2024-12-05
[ESM] Raise exception when setting DestinationConfig.OnSuccess

gregfurman:
## Motivation
Although the AWS docs indicates we can use a [`DestinationConfig` object's `OnSuccess`](https://docs.aws.amazon.com/lambda/latest/api/API_DestinationConfig.html#lambda-Type-DestinationConfig-OnSuccess) in ESM, this is actually not supported. This PR adds a parity test and raises an exception for the aforementioned case.

## Changes
- Added parity test with raised exception
- Raise exception in LS on ESM validation


joe4dev (approved):
LGTM, great catch 👏

PR #11990, opened 2024-12-05
[EventsBridge] Add validation for malformed Detail value

gregfurman:
## Motivation
This PR validates an event entry's `Detail` field to ensure it is a valid JSON object.

From the [`PutEventsRequestEntry` AWS docs](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEventsRequestEntry.html#eventbridge-Type-PutEventsRequestEntry-Detail):

> **Detail:** A valid JSON object. There is no other schema imposed. The JSON object may contain fields and nested sub-objects.

In addition to the [`PutEventsResultEntry` AWS docs](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEventsResultEntry.html#eventbridge-Type-PutEventsResultEntry-ErrorCode) for the error code:

> **MalformedDetail:** The JSON provided is not valid.

## Changes
- Add validation to `validate_event` that returns `MalformedDetail` error type response.
- Add negative test to validate malformed payloads return an error response.


tests/aws/services/events/test_events.py
```diff
@@ -152,6 +152,27 @@ def test_put_event_without_detail_type(self, snapshot, aws_client):
...
         snapshot.match("put-events", response)
 
+    @markers.aws.validated
+    @pytest.mark.skipif(
+        is_old_provider(),
+        reason="V1 provider does not support this feature",
+    )
+    @pytest.mark.parametrize(
+        "detail",
+        ["NotJSON", "[]", "{{}", json.dumps("NotJSON")],
```
maxhoheiser, line 162: nice also testing for plain string - was a bit surprised that this also results in malformed error but it is correct :)

maxhoheiser (approved):
Nice catch :)

PR #11991, opened 2024-12-05
[SFN] Correctly record failed EventsBridge PUT requests

gregfurman:
## Motivation
This PR fixes the construction of EventsBridge PUT failures -- as well as recording them to the execution history. Relies on changes in https://github.com/localstack/localstack/pull/11990

## Changes
- Remove skips on events tests.
- Adds another test to ensure we can handle the mixed case where some entries are malformed and some are valid.
- Pass the PUT response as the cause of the failure event when constructing failed events.


MEPalma (approved):
LGTM

PR #11992, opened 2024-12-05
Unskip fixed event filtering tests

joe4dev:
## Motivation

With @bentsku's new Python-based event filtering engine, we should be able to unskip tests that previously relied on the Java engine but were skipped to avoid JPype conflicts with StepFunctions JSONata.

## Changes

* Unskip event filtering ESM tests

## TODO

- [x] test_dynamodb_event_filter[numeric_filter] is fixed by https://github.com/localstack/localstack/pull/11994
- [x] Invalid filter detection for ESM should be fixed with a WIP PR by @gregfurman 
- [ ] test_dynamodb_event_filter[content_multiple_filters] is an ESM issue (as the skip comment says) and needs investigation -> SKIPPED for now, needs more investigation in a follow up
- [x] NICE-TO-HAVE: migrate `test_sqs_event_filter` to use `pytest.param` (Example in `aws.services.lambda_.event_source_mapping.test_lambda_integration_dynamodbstreams.TestDynamoDBEventSourceMapping.test_dynamodb_event_filter`) because it makes selective test execution and reporting so much easier and clearer.

### DIFF
--- a/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
+++ b/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
@@ -580,6 +580,7 @@ class TestDynamoDBEventSourceMapping:
                 {"eventName": ["INSERT"], "eventSource": ["aws:dynamodb"]},
                 1,
                 id="content_multiple_filters",
+                marks=pytest.skip(reason="Broken, needs investigation"),
             ),
             # Test content filter using the DynamoDB data type "S"
             pytest.param(
--- a/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
+++ b/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
@@ -672,8 +672,6 @@ class TestDynamoDBEventSourceMapping:
         Test assumption: The first item MUST always match the filter and the second item CAN match the filter.
         => This enables two-step testing (i.e., snapshots between inserts) but is unreliable and should be revised.
         """
-        if filter == {"eventName": ["INSERT"], "eventSource": ["aws:dynamodb"]}:
-            pytest.skip(reason="content_multiple_filters failing for ESM v2 (needs investigation)")
         function_name = f"lambda_func-{short_uid()}"
         table_name = f"test-table-{short_uid()}"
         max_retries = 50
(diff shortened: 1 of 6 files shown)


bentsku:
I'll create a PR shortly to fix the 2 use cases that are skipped now 👀 I didn't know about this, I would tried to run them before! darn it 😄

bentsku:
Update regarding `test_dynamodb_event_filter[content_multiple_filters]`: I don't think this failure is related to the Event Rule engine, but to ESM v2 directly as the skip indicated: 

This is the rule:
`{'eventName': ['INSERT'], 'eventSource': ['aws:dynamodb']}`

And this is the payload sent by Lambda to the EventRuleEngine:
```json
{
    "eventID": "455fabb3",
    "eventName": "INSERT",
    "dynamodb": {
      "ApproximateCreationDateTime": 1733416705.0,
      "Keys": {"id": {"S": "id_value"}},
      "NewImage": {"id": {"S": "id_value"}},
      "SequenceNumber": "49658361055172123998835686422808469640999919842681159682",
      "SizeBytes": 23,
      "StreamViewType": "NEW_AND_OLD_IMAGES"
    }
}
```

As you can see, the `eventSource` field is missing from the event, which is why the test is failing. I would say that this test will fail with the Java-based event ruler as well.

bentsku:
Update on all issues:

- I've opened #11994 to fix `test_dynamodb_event_filter[numeric_filter]`, which is now confirmed to work.
- `test_dynamodb_event_filter[content_multiple_filters]` is not related to the EventRuleEngine I believe, but to ESM not populating a field before passing it to the engine
- I was unable to make `test_sqs_event_filter[this is a test string]` fail, @joe4dev feel free to send it my way if you still have issues

All the other test failures seems to be related to the validation of the rule itself. 

Thanks again @joe4dev for raising this!

joe4dev:
> Update on all issues:
> 
>     * `test_dynamodb_event_filter[numeric_filter]`: I've opened [Events: fix `numeric` operator string handling and provider validation for TestEventPattern #11994](https://github.com/localstack/localstack/pull/11994) to fix it, which is now confirmed to work.
> 
>     * `test_dynamodb_event_filter[content_multiple_filters]` is not related to the EventRuleEngine I believe, but to ESM not populating a field before passing it to the engine
> 
>     * I was unable to make `test_sqs_event_filter[this is a test string]` fail, @joe4dev feel free to send it my way if you still have issues
> 
> 
> All the other test failures seems to be related to the validation of the rule itself.
> 
> Thanks again @joe4dev for raising this!

Wow, amazing quick updates @bentsku 🙇 I fully agree with your conclusion:
* `test_sqs_event_filter[this is a test string]` was a false alarm (confirmed by CI)
* `test_dynamodb_event_filter[content_multiple_filters]`  is an ESM issue (as the skip comment says)
* `test_dynamodb_event_filter[numeric_filter]` was the only problematic one. Thanks for the quick fix 🥇

dfangl (approved):
More test coverage is always good :D

Issue #11850, opened 2024-11-14
bug: Issue with SNS message parsing with C#


nascosto:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Note: I'm trying to migrate from localstack 2.3.2 to localstack 3.8.1. With 2.3.2, the following worked.

Parsing a message using `Amazon.SimpleNotificationService.Util.Message.ParseMessage` throws an `AmazonClientException("Signing certificate url is not from a recognised source.")`. Investigating this issue the error is being thrown by this method
```csharp
private static string ValidateCertUrl(string certUrl)
{
    Uri uri = new Uri(certUrl);
    if (uri.Scheme == "https" && certUrl.EndsWith(".pem", StringComparison.Ordinal) && CertUrlRegex().IsMatch(uri.Host))
    {
	    return certUrl;
    }
    throw new AmazonClientException("Signing certificate url is not from a recognised source.");
}
```
`CertUrlRegex` is the following regex: `^sns\\.[a-zA-Z0-9\\-]{3,}\\.amazonaws\\.com(\\.cn)?$`

The actual SigningCertURL with localstack 3.8.1 is http://localhost.localstack.cloud:4566/_aws/sns/SimpleNotificationService-6c6f63616c737461636b69736e696365.pem where I think in 2.x is was a dummy value that matched the validation needs above.

I'm wondering if there is any way around this.

### Expected Behavior

There is no error when calling ParseMessage.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    Testcontainers

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```csharp
string dockerHostUri = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
    "npipe://./pipe/docker_engine" :
    "unix:///var/run/docker.sock";
var localStackBuilder = new LocalStackBuilder()
    .WithImage("localstack/localstack:3.8.1")
    .WithCleanUp(true)
    .WithEnvironment("SERVICES", "sns,sqs")
    .WithEnvironment("DOCKER_HOST", dockerHostUri)
    .WithEnvironment("DEBUG", "1")
    .WithPortBinding(4566, 4566);
```

### Environment

```markdown
- OS: Windows 11
- LocalStack:
  LocalStack version: 3.8.1
  LocalStack Docker image sha:
[...]

PR #11993, opened 2024-12-05
make SNS `SigningCertURL` configurable even if not resolvable

bentsku:
## Motivation
As reported by #11850, some SDK will verify the `SigningCertURL` field host part even if not validating the messages. 

This PR introduces a config flag to allow the user to set the host of the `SigningCertURL` to pass those regex pattern, even if the host is then non-resolvable, it should at least unblock the user, as they confirmed this would be sufficient for their use case. 

## Changes
- add a config flag `SNS_CERT_URL_HOST` to allow the user the set the host to an arbitrary value
- create a test to validate SNS is properly using that value

_fixes #11850_

## TODO

What's left to do:

- [ ] create the docs PR to document this new config flag


cloutierMat (approved):
LGTM! Great solution for what looked like a pretty complex issue! :rocket:

PR #11994, opened 2024-12-05
Events: fix `numeric` operator string handling and provider validation for TestEventPattern

bentsku:
## Motivation
As reported in #11992, there might have been issue with the new engine rule engine.

I could only reproduce the `test_dynamodb_event_filter[numeric_filter]` failure, which was due to our bad handling of `numeric` logic with string values. 

While validating the fix and the handling of the new EventRuleEngine for `test_sqs_event_filter[this is a test string]`, the behavior was right, but the validation for `TestEventPattern` was wrong. 

I've also validated that `test_dynamodb_event_filter[numeric_filter]` now works with the fix 👍 

Then, when validating the DynamoDB fix, I've tried to copy-paste the input event from ESM, but it failed when calling `TestEventPattern` because the event was not following the right format, so I've added it in the provider to also validate this, as specified by the documentation here: https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_TestEventPattern.html (note, the `resources` field is not mandatory in reality). 

## Changes
- fix the handling of `numeric` if the value is a string (or more like the value is not a numeric value)
- fix the validation of the `TestEventPattern` provider operation
- add test cases to cover the `numeric` fix, and the provider validation

### DIFF
--- a/tests/aws/services/events/test_events_patterns.py
+++ b/tests/aws/services/events/test_events_patterns.py
@@ -251,7 +251,7 @@ class TestEventPattern:
         reason="V1 provider does not properly validate",
     )
     def test_invalid_event_payload(self, aws_client, snapshot):
-        # following fields are mandatory: `id`, `account`, `source`, `time`, `region`, `resources`, `detail-type`
+        # following fields are mandatory: `id`, `account`, `source`, `time`, `region`, `detail-type`
         event = {"testEvent": "value"}
         pattern = {"body": {"test2": [{"numeric": [">", 100]}]}}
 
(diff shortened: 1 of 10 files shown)


tests/aws/services/events/test_events_patterns.py
```diff
@@ -212,6 +213,39 @@ def test_invalid_json_event_pattern(self, aws_client, pattern, snapshot):
...
+            )
+        snapshot.match("plain-string-payload-exc", e.value.response)
+
+    @markers.aws.validated
+    @pytest.mark.skipif(
+        is_old_provider(),
+        reason="V1 provider does not properly validate",
+    )
+    def test_invalid_event_payload(self, aws_client, snapshot):
+        # following fields are mandatory: `id`, `account`, `source`, `time`, `region`, `resources`, `detail-type`
```
bentsku, line 238: note: will update the comment after the review, as `resources` is not mandatory unlike what the documentation says

joe4dev (approved):
Wow, amazing edge case coverage ✨ 💯

PR #11995, opened 2024-12-05
fix StringSplit regex escape

cloutierMat:
## Motivation

The `StringSplit` function allows to split on any of the character provided by the second argument. The current code was manually escaping each character, appending `\\` to it. While allowing for the most common special characters delimiter, the unfortunate side effect is that digits and letters would be incorrectly escaped.

## Changes

- update the `StringSplit` function to use `re.escape` instead for a safe escaping experience
- perform the `join` directly on the list comprehension for a small performance upgrade


MEPalma (approved):
re.escape is definately a better approach to this, thanks!

PR #12005, opened 2024-12-09
upgrade cryptography and amazon-kclpy

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/11829 and https://github.com/localstack/localstack/pull/11245 we introduced some dependency limits which we need to remove to get up-to-date again.

## Changes
- Remove the pin on `cryptography` and `amazon_kclpy`.
- Upgrade the client config for `amazon_kclpy` to enable "v2 compatibility mode" (to minimize the amount of necessary changes in this PR while fixing the failing tests).
- Extend the timeout for `test_firehose_kinesis_to_s3` as it was already flaky before.

## TODO
- [x] Fix failing tests:
  - [x] `test_firehose_kinesis_to_s3`
  - [x] `test_lambda_streams_batch_and_transactions`
  - [x] `test_firehose_stack_with_kinesis_as_source`
  - [x] `test_run_kcl`

## TODO in a follow-up
- Migrate the usage of KCL to v3 (since this PR just enables the v2 compat mode).

### DIFF
--- a/localstack-core/localstack/utils/kinesis/kclipy_helper.py
+++ b/localstack-core/localstack/utils/kinesis/kclipy_helper.py
@@ -95,11 +95,13 @@ def create_config_file(
 ):
     if not credentialsProvider:
         credentialsProvider = "DefaultCredentialsProvider"
+    # TODO properly migrate to v3 of KCL and remove the clientVersionConfig
     content = f"""
         executableName = {executableName}
         streamName = {streamName}
         applicationName = {applicationName}
         AWSCredentialsProvider = {credentialsProvider}
+        clientVersionConfig = CLIENT_VERSION_CONFIG_COMPATIBLE_WITH_2x
         kinesisCredentialsProvider = {credentialsProvider}
         dynamoDBCredentialsProvider = {credentialsProvider}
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/utils/kinesis/kclipy_helper.py
```diff
@@ -93,16 +93,17 @@ def create_config_file(
...
         executableName = {executableName}
         streamName = {streamName}
         applicationName = {applicationName}
         AWSCredentialsProvider = {credentialsProvider}
+        clientVersionConfig = CLIENT_VERSION_CONFIG_COMPATIBLE_WITH_2x
         kinesisCredentialsProvider = {credentialsProvider}
         dynamoDBCredentialsProvider = {credentialsProvider}
         cloudWatchCredentialsProvider = {credentialsProvider}
-        processingLanguage = python/3.10
+        processingLanguage = python/3.11
```
dfangl, line 106: Could we link this directly to `sys.version_info`? For example like this (needs sys import as well):
```suggestion
        processingLanguage = python/{sys.version_info.major}.{sys.version_info.minor}
```

PR #12006, opened 2024-12-09
Allow usage of additional runtimes with the Lambda SnapStart feature enabled

dfangl:
## Motivation
Recently, AWS announced support for additional runtimes for using SnapStart: https://aws.amazon.com/blogs/aws/aws-lambda-snapstart-for-python-and-net-functions-is-now-generally-available/

However, while investigating this change, it seems that over the API the feature is available for all runtimes, not just the announced ones.

Whether this is a temporary error, or a hint that additional runtimes could get supported in the future, I removed the runtime validation for snapstart creation and updates, and added aws-validated tests for all runtimes to show it is indeed possible to create them.

I did some more investigation, and it seems like SnapStart is even (somewhat) functioning, at least an initial initialization of the function potentially takes place (if you misconfigure the function, it errors out immediately, it does not do so without snapstart enabled).

## Changes
* SnapStart is now available for additional runtimes (CRUD only). As AWS does not raise an error for unsupported ones, we do not do so either.

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -36,7 +36,6 @@ from localstack.services.lambda_.provider import TAG_KEY_CUSTOM_URL
 from localstack.services.lambda_.runtimes import (
     ALL_RUNTIMES,
     DEPRECATED_RUNTIMES,
-    SNAP_START_SUPPORTED_RUNTIMES,
 )
 from localstack.testing.aws.lambda_utils import (
     _await_dynamodb_table_active,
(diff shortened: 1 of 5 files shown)


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -6599,20 +6592,15 @@ def test_snapstart_lifecycle(self, create_lambda_function, snapshot, aws_client,
         snapshot.match("get_function_response_version_1", get_function_response)
 
     @markers.aws.validated
-    @pytest.mark.parametrize("runtime", [Runtime.java21, Runtime.java17])
+    @markers.lambda_runtime_update
+    @markers.multiruntime(scenario="echo")
     def test_snapstart_update_function_configuration(
-        self, create_lambda_function, snapshot, aws_client, runtime
+        self, multiruntime_lambda, snapshot, aws_client
```
joe4dev, line 6598: `multiruntime_lambda` 🚀

tests/aws/services/lambda_/test_lambda_api.snapshot.json
```diff
@@ -12934,22 +12916,19 @@
...
           "Description": "",
-          "Environment": {
-            "Variables": {}
-          },
           "EphemeralStorage": {
             "Size": 512
           },
           "FunctionArn": "arn:<partition>:lambda:<region>:111111111111:function:<function-name:1>",
           "FunctionName": "<function-name:1>",
-          "Handler": "cloud.localstack.sample.LambdaHandlerWithLib",
```
joe4dev, line 12945: 🧹 🥳

PR #12010, opened 2024-12-10
CLI - Passthrough MSSQL_ACCEPT_EULA env var

bblommers:
## Motivation
The `MSSQL_ACCEPT_EULA` is required for users who want to start an MSSQL-database in RDS ([existing docs here](https://docs.localstack.cloud/user-guide/aws/rds/#microsoft-sql-server-engine)).

The same environment variable (with the same usecase) is now also used in other emulators, so it would be nice if users could use the LocalStack CLI with this variable enabled.


localstack-core/localstack/config.py
```diff
@@ -1321,6 +1321,7 @@ def use_custom_dns():
     "LS_LOG",
     "MAIN_CONTAINER_NAME",
     "MAIN_DOCKER_NETWORK",
+    "MSSQL_ACCEPT_EULA",
```
bentsku, line 1324: would it be possible to add this config variable in Pro instead? or does it not work with new emulators?

bentsku (commented):
Good catch! Just a question regarding where to add this variable, as this is related to Pro configuration, and we can also add variables passed to the CLI in Pro configuration.

localstack-core/localstack/config.py
bblommers: Ah, that should work as well! I didn't know that it was a Pro feature in AWS. I've opened a new PR there instead. :+1:

PR #12013, opened 2024-12-10
add validation before creating an s3 bucket for scenario testing

pinzon:
## Motivation
As discussed with @dominikschubert @bentsku S3 behaves differently when creating a Bucket in different regions. Being Idempotent when using region `us-east-1` and not idempotent in other Regions. This AWS behavior is causing issues with Multi-region test.

## Changes
- Check if intended bucket exists before trying to create it for Scenario test assets.

## Testing
- Since it's an extra validation, an all green checks should be enough.

### DIFF
--- a/localstack-core/localstack/testing/scenario/provisioning.py
+++ b/localstack-core/localstack/testing/scenario/provisioning.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Callable, ContextManager, Optional
 
 import aws_cdk as cdk
-from botocore.exceptions import WaiterError
+from botocore.exceptions import ClientError, WaiterError
 
 from localstack.config import is_env_true
 from localstack.testing.aws.util import is_aws_cloud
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/testing/scenario/provisioning.py
```diff
@@ -402,7 +402,12 @@ def _template_bucket_name(self):
         return f"localstack-testing-assets-{account_id}-{region}"
 
     def _create_bucket_if_not_exists(self, template_bucket_name: str):
-        create_s3_bucket(template_bucket_name, s3_client=self.aws_client.s3)
+        try:
+            self.aws_client.s3.head_bucket(Bucket=template_bucket_name)
+        except ClientError as exc:
+            if exc.response["Error"]["Message"] != "Not Found":
+                raise exc
```
dominikschubert, line 409:
```suggestion
        except ClientError:
            if exc.response["Error"]["Message"] != "Not Found":
                raise
```

Bit of a nit, feel free to ignore. Just a bit cleaner that way since you don't need the reference

dominikschubert: suggestion: Given the docs on HeadBucket, it might also be a bit cleaner to check the response code (404) instead

> If the bucket does not exist or you do not  have  permission  to
> access  it, the HEAD request returns a generic 400 Bad Request ,
> 403 Forbidden or 404 Not Found code. A message body is  not  in-
> cluded,  so you cannot determine the exception beyond these HTTP
> response codes.

dominikschubert (approved):
Minor nits, otherwise LGTM!

localstack-core/localstack/testing/scenario/provisioning.py
bentsku: I think you need to keep the reference to `exc` to be able to check on the status code or error message of it though! But just calling `raise` is cleaner 👍

PR #12015, opened 2024-12-10
Update help command output UX

gtsiolis:
## Motivation

Some UX updates in the help output.

⚠️ May contain non-optimized Python code. 🐍 

## Changes

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-12-11 at 00 42 23 | Screenshot 2024-12-11 at 00 42 17 |


alexrashed (commented):
I had a quick dive into `click` and how this works, and apparently this isn't super-straight forward to implement.
You modified the `LocalStackCliGroup`, which means it only applies to CLI groups which explicitly define to use this class.
However, there also are:
- Groups that do not explicitly define that this specific group class should be used.
- Commands, which are not groups, which show a similar help message, where we do not have a custom implementation yet.
- Other custom group classes (like the one for pods) which needs to be modified such that it inherits from this group class.

If you want we can try to take a quick dive together, but I am afraid this will take you some time to implement globally in the CLI...

gtsiolis:
@alexrashed Following up from a discussion a few weeks ago, feel free to take this over or close the PR. Happy to assist with more design specs if needed. 🏓 

Low priority! 🔅

alexrashed:
@gtsiolis Given that we are working on a new CLI experience, do you think we can close this PR for now?

gtsiolis:
> ... do you think we can close this PR for now?

Definitely. DONE.

gtsiolis:
Cc @silv-io @carole-lavillonniere for visibility. No action required.

PR #12017, opened 2024-12-11
Replicator implement cfn read for supported resources

cloutierMat:
## Motivation

Base implementation of `get` and `list` CloudFormation operation for `Vpc`, `Subnet`, `KmsKey`, and `secret`. This is in a continued effort to further the work with the cloudcontrol api.

## Changes

- Endpoint implementation
- Refresh of specs where needed


localstack-core/localstack/services/ec2/resource_providers/aws_ec2_subnet.py
```diff
@@ -41,6 +41,47 @@ class Tag(TypedDict):
 REPEATED_INVOCATION = "repeated_invocation"
 
 
+def generate_subnet_read_payload(
+    ec2_client, schema, subnet_ids: Optional[list[str]] = None
+) -> list[EC2SubnetProperties]:
```
pinzon, line 46: Suggestion: Seems like `generate_subnet_read_payload`  could be use it for both `list` and `read` like you did in other resource providers.

pinzon (approved):
Good PR :+1:  Just a minor suggestion

localstack-core/localstack/services/ec2/resource_providers/aws_ec2_subnet.py
cloutierMat: ah yes! That was the goal. Forgot to update that one! Thanks for the catch! :pray:

PR #12018, opened 2024-12-11
[ESM] Validate filter criteria and unskip tests

gregfurman:
## Motivation
Following the introduction of the new python-based filtering engine for events, we can unskip several tests and improve validations of filter criteria when creating/updating ESMs.

## Changes
- Unskips failing filter criteria tests.
- Adds `validate_event_pattern` to `event_matcher.py`.
- Adds `test_create_event_filter_criteria_validation` to ESM lifecycle tests.
- Fixes parity issue that includes an empty filter criteria object in returned ESM config.

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -5625,6 +5625,9 @@ class TestLambdaEventSourceMappings:
         )
 
         table_name = f"table-{short_uid()}"
+        # FIXME: Why is this not being automatically transformed?
+        snapshot.add_transformer(snapshot.transform.regex(table_name, "<table-name>"))
+
         dynamodb_create_table(table_name=table_name, partition_key="id")
         _await_dynamodb_table_active(aws_client.dynamodb, table_name)
         update_table_response = aws_client.dynamodb.update_table(
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -5610,6 +5610,64 @@ class TestLambdaEventSourceMappings:
         snapshot.match("error", response)
 
     @markers.aws.validated
+    def test_create_event_filter_criteria_validation(
+        self,
+        create_lambda_function,
+        lambda_su_role,
+        dynamodb_create_table,
+        snapshot,
+        aws_client,
+    ):
+        function_name = f"function-{short_uid()}"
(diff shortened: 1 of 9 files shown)


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -5607,6 +5607,67 @@ def test_create_event_source_validation(
...
+        function_name = f"function-{short_uid()}"
+        create_lambda_function(
+            handler_file=TEST_LAMBDA_PYTHON_ECHO,
+            func_name=function_name,
+            runtime=Runtime.python3_12,
+            role=lambda_su_role,
+        )
+
+        table_name = f"table-{short_uid()}"
+        # FIXME: Why is this not being automatically transformed?
```
joe4dev, line 5628: Do you expect DynamoDB transformers to apply automatically in the Lambda API tests?

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -5607,6 +5607,67 @@ def test_create_event_source_validation(
         response = e.value.response
         snapshot.match("error", response)
 
+    @markers.aws.validated
+    def test_create_event_filter_criteria_validation(
+        self,
+        create_lambda_function,
+        lambda_su_role,
+        dynamodb_create_table,
```
joe4dev, line 5615: Does this test need a relatively heavy-weight DynamoDB stream or would a more lightweight SQS queue be sufficient here (could also simplify the test case by removing `StartingPosition`)?

joe4dev (approved):
Nice work 👏 

We could consider using SQS to simplify and speed up testing, if DynamoDB is not required.

tests/aws/services/lambda_/test_lambda_api.py
gregfurman: Look at the `test_event_source_mapping_lifecycle_delete_function`. It resides in the same class but has it's `create_response` correctly transformed.

tests/aws/services/lambda_/test_lambda_api.py
gregfurman: Tbh I just copied and adpated the above `test_create_event_source_validation` test. Was the quickest solution here. Since we are only ever doing a create (and not reading anything) I think the improvement to SQS will probably be very minimal.

The `StartingPosition` is required unfortunately for creating an ESM with dynamodb streams.

PR #12020, opened 2024-12-11
fix APIGW test with hardcoded region

bentsku:
## Motivation
As reported by our pipeline here: https://app.circleci.com/pipelines/github/localstack/localstack/30176/workflows/23745dcd-aaf8-4add-84f5-ae041d88e5a0

We had a failure when the run was using the `us-west-1` region, and the automatic snapshot replacement would replace that value with the `<region>` placeholder. 

However, this hardcoded can be kept as is, because it is an actual available endpoint in AWS, and we can't be sure it's available in all regions. The workaround for this is to transform the value with a higher priority that the regular `<region>` transformer, to be sure we always replace the value in the snapshots. 

## Changes
- add a transformer with higher priority of the external AWS APIGW endpoint

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_import.py
+++ b/tests/aws/services/apigateway/test_apigateway_import.py
@@ -868,6 +868,13 @@ class TestApiGatewayImportRestApi:
                 ),
             ]
         )
+        snapshot.add_transformer(
+            snapshot.transform.regex(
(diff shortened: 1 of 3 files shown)


alexrashed (approved):
Awesome, thanks a lot for jumping on this! 🦸🏽 🧹

PR #12021, opened 2024-12-11
fix APIGW transformers to not do value replacement on port

bentsku:
## Motivation

@dfangl encountered a flake in an APIGW test which was quite unlikely, due to a port field (usually `443`) being present in a UUID and breaking the snapshot replacement. 

https://app.circleci.com/pipelines/github/localstack/localstack/30209/workflows/d7b445e5-43ba-4686-8d83-73a0c370d721/jobs/267451/tests

side-note: the content length changing between snapshots might mean we should transform it, it runs against AWS again when run but might change over time. We can always look into it in a follow up

## Changes
- do not reference replace the forwarded port field as it is not really important to do so, and update the snapshots that are concerned


cloutierMat (approved):
Awesome work on attacking the flakes once again! :pray: 

Curious for the content-length. maybe something to look at in the future but since they are all skipped anyway, it shouldn't be an issue for now.

PR #12023, opened 2024-12-12
APIGW: ParameterMapping skip if invalid input

bentsku:
## Motivation
We've got a report from a user with this stack trace:
```python
2024-12-11 15:56:52   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py", line 59, in map_integration_request

2024-12-11 15:56:52     if request_mapping.startswith("method.request."):

2024-12-11 15:56:52        ^^^^^^^^^^^^^^^^^^^^^^^^^^

2024-12-11 15:56:52 AttributeError: 'bool' object has no attribute 'startswith'
```

However, it seems the only way we can get to those conditions is by specifying an invalid `requestParameters` object with boolean values. It seems AWS does not accept that, but as we are lacking way to reproduce and to be sure we are not going too far with provider validation, this PR only skips the error for now. 

We need as a follow up to implement proper Parameter Mapping validation like we have done for HTTP APIs. 

I've already added a test for the AWS validation, but we need to implement the validation in a follow up. 

## Changes
- add a skipped test to verify that we cannot have any else than strings in the parameter mappings
- skip the parameter mapping if the values aren't string, to avoid non-caught exceptions

## TODO

What's left to do:

- [ ] in a follow-up, add the provider validation so that it's not possible to get in those conditions in the invocation layer

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/parameters_mapping.py
@@ -52,6 +52,15 @@ class ParametersMapper:
         case_sensitive_headers = build_multi_value_headers(invocation_request["headers"])
 
         for integration_mapping, request_mapping in request_parameters.items():
+            # TODO: remove this once the validation has been added to the provider, to avoid breaking
+            if not isinstance(integration_mapping, str) or not isinstance(request_mapping, str):
+                LOG.warning(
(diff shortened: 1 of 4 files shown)


cloutierMat (approved):
Thanks for adding the fix and a good test for the provider. 

That should unblock the user for now while we implement proper validation in the provider! :pray:

Issue #1573, opened 2019-09-18
New debug_java_port causing issues with lambda


a88zach:
We have several Kinesis streams set up with event notifications that trigger lambdas.  When the LAMBDA_EXECUTOR is set to 'docker' and we send a large number or records to Kinesis, multiple Lambdas will execute, but currently they are failing to create with error:

```
Exception: Lambda process returned error status code: 1. Output:
time="2019-09-18T17:28:11Z" level=error msg="error getting events from daemon: net/http: request canceled" 
Error response from daemon: driver failed programming external connectivity on endpoint busy_sanderson (a78e45a41d0ce5b0d8bab6ff54344fc34fe2b4969f55bf568e753121abc8759e): Bind for 0.0.0.0:38763 failed: port is already allocated
```

I believe this is due to the new debug_java_port introduced with this PR https://github.com/localstack/localstack/pull/1523

With this change, you are setting the port in the constructor by using a method known to have a race condition (https://github.com/localstack/localstack/blob/master/localstack/utils/common.py#L354)
(https://gist.github.com/gabrielfalcao/20e567e188f588b65ba2)

This should be changed to ensure unique ports are being issued across all lambdas

rmrbest:
@a88zach Thanks for the issue. Let me check to fix it. @whummer what do you think?

rmrbest:
@a88zach Do you put LAMDA_JAVA_OPTS as ennvironment variable in your docker run or docker-compose?

rmrbest:
I think that always set a get_free_tcp_port, even when LAMDA_JAVA_OPTS is not set. May be, set a port when LAMDA_JAVA_OPTS is added, so is for debug purpose. If does not exist(not set), normal behaviour, i mean, don't set a port.

a88zach:
Even with only setting the debug port when including java options, you're still going to run across this issue.  The main problem is that get_free_tcp_port is always going to return the same port as long as it is free.  With multiple dockers starting up at the same time you run into a race condition where  that function keeps giving out the same "free" port to multiple dockers b/c those dockers are still being initialized and have not taken the port yet

rmrbest:
@whummer Could we use a random port? I've a bit time that weekend to fix it.

whummer:
> I think that always set a get_free_tcp_port, even when debug_port is not set. May be, set a port when debug_port is added

I think this would be the right approach. Will push a fix asap @rmrbest

whummer:
[...]

PR #12026, opened 2024-12-12
Fix swagger endpoint from ephemeral instance

giograno:
## Motivation
The swagger endpoint is not functioning properly when used from an ephemeral instance (see [#1573](https://github.com/localstack/localstack-web/pull/1573) for more context).
We are relying on the `external_service_url` function to compute the path for the `openapi.yaml` file. This function uses the default 4566 port, leading to a fetch error.

## Changes
- Consider the special case of ephemeral instances when building the `openapi.yaml` path.

### DIFF
--- a/localstack-core/localstack/http/resources/swagger/endpoints.py
+++ b/localstack-core/localstack/http/resources/swagger/endpoints.py
@@ -7,10 +7,17 @@ from localstack.config import external_service_url
 from localstack.http import Response
 
 
+def _get_service_url(request: Request) -> str:
+    # special case for ephemeral instances
+    if "sandbox.localstack.cloud" in request.host:
+        return external_service_url(protocol="https", port=443)
+    return external_service_url(protocol=request.scheme)
+
+
 class SwaggerUIApi:
     @route("/_localstack/swagger", methods=["GET"])
     def server_swagger_ui(self, request: Request) -> Response:
-        init_path = f"{external_service_url(protocol=request.scheme)}/openapi.yaml"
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/http/resources/swagger/endpoints.py
```diff
@@ -7,10 +7,17 @@
 from localstack.http import Response
 
 
+def _get_service_url(request: Request) -> str:
+    # special case for ephemeral instances
+    if "sandbox.localstack.cloud" in request.host:
+        return external_service_url(protocol="https", port=443)
```
lukqw, line 13: **Thought**: This seems a bit ugly - I do wonder if there's something better we can use to use port 443.

As an idea: Could we set the port for the `external_service_url` call depending on the `request.scheme`? If it's HTTPS use 443 - if not 4566.

lukqw (commented):
Thanks for addressing this issue! I've added a comment which might improve the solution, but I'm probably missing lots of context.

localstack-core/localstack/http/resources/swagger/endpoints.py
giograno: If we do this, it messes up with the default `https://localhost.localstack.cloud:4566`. Ephemeral instances are the only case that comes to mind where we don't have a port but just default to 443.

thrau (approved):
also not a huge fan of hardcoding URLs and ports in this way, but i don't know of a better solution right now, so if this a stop-gap to unblock swagger in ephemeral instances, i guess we can go with this for now :+1:

PR #12027, opened 2024-12-12
apigw fix openapi request parameters

cloutierMat:
## Motivation

As reported by a user, if a YAML OpenApi specs passes a Boolean value for request parameter mapping, we would fail the invocation for that integration. 

While the value is invalid, AWS doesn't fail and instead just ignores the parameter when importing the api and adds a warning to the response.

## Changes

- Make sure that a boolean value cannot be added as an integration request parameters mapping
- adds warnings to the response of the import.

## Notes

If the flags `failsOnWarning` is provided, we should instead revert the changes when encountering a warning. But this is something we will need to tackle in a follow up as it would require much bigger changes to the import and put endpoints and the helpers method

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -24,6 +24,7 @@ from localstack.aws.api.apigateway import (
     IntegrationType,
     Model,
     NotFoundException,
+    PutRestApiRequest,
     RequestValidator,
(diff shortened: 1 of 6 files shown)


bentsku (approved):
LGTM! This was a fun one 😬 especially since JSON spec files are failing straight away, so it was hard to reproduce! 

Very nice find 🚀

PR #12032, opened 2024-12-13
fix `echo_http_server_post` returning double slash path

bentsku:
## Motivation
The `echo_http_server_post` has different behavior between AWS and LocalStack (where it will direct the request against `https://httpbin.org/post` in AWS and a local http server in LocalStack, but at this url: `http://localhost:<port>//post`).

This PR fixes the format for them to both return the `/post` path. 

I'll run a downstream test on this to verify everything is right. 

## Changes
- don't add a slash when there's already one returned by `http server.url_for("/")`.


alexrashed (approved):
Thanks a lot for jumping on this! 🦸🏽 💯

Issue #9809, opened 2023-12-05
bug: Step Function distributed maps throw pickle exception failing to enter map state, {"error": "States.Runtime", "cause": "cannot pickle '_thread.lock' object"}


JordanWeir:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When creating a step function using distributed maps on localstack pro, the execution stops before the first map task.  The localstack docker container shows an error indicating issues with pickling.

2023-12-05T16:45:58.482  WARN --- [d-159 (eval)] l.s.s.a.c.s.s.execute_stat : State Task encountered an unhandled exception that lead to a State.Runtime error.
[...]

PR #12033, opened 2024-12-13
StepFunctions: Fix Evaluation of Nested Map Runs

MEPalma:
## Motivation
Currently, the SFN v2 is unable to evaluate nested MapRun states, resulting in pickling runtime errors https://github.com/localstack/localstack/issues/9809. This was due to an unecessary deep copy of the map run program, which contains unserialisable objects such as locks. The deep copy is not considered necessary as EvalComponents are stateless by design, and can therefore be shared by multiple runtimes. These changes address such issue and add a relevant snapshot test.

## Changes
- remove unecessary program copy
- added a snapshot test for nested map runs


gregfurman (approved):
LGTM!

PR #12034, opened 2024-12-13
add read/list methods for ec2 security-groups resource provider

pinzon:
## Motivation
This PR adds the Operations for READ and LIST to the Resource Provider for AWS::Ec2::SecurityGroup

cc/ @cloutierMat  @simonrw 

## Changes
- Read and List operations
- Fix the setting of a custom ID in the Patches

## Testing
- These operations are use by CloudControl only. That service is not available in this repository.


localstack-core/localstack/services/ec2/patches.py
```diff
@@ -133,9 +127,9 @@ def ec2_create_security_group(
 
         if custom_id:
             # Remove the security group from the default dict and add it back with the custom id
-            self.groups[vpc_id].pop(result.group_id)
+            self.groups[result.vpc_id].pop(result.group_id)
             result.group_id = result.id = custom_id
-            self.groups[vpc_id][custom_id] = result
```
pinzon, line 138: This could crash if VpcId was null which is the case for auto assigning to the Default Vpc

localstack-core/localstack/services/ec2/patches.py
```diff
@@ -120,11 +119,6 @@ def ec2_create_security_group(
         # Extract tags and custom ID
         tags: dict[str, str] = tags or {}
         custom_id = tags.get(TAG_KEY_CUSTOM_ID)
-        vpc_id: str = kwargs["vpc_id"] if "vpc_id" in kwargs else args[2]
-
-        # Check if custom id is unique
-        if not force and custom_id in self.groups[vpc_id]:
-            raise InvalidSecurityGroupDuplicateCustomIdError(custom_id)
```
pinzon, line 127: This Validation is done already in the service.

viren-nadkarni (approved):

PR #12036, opened 2024-12-13
[ESM] Handle DynamoDB-local Invalid ShardID exception

gregfurman:
## Motivation
Doing a `GetRecords` operation against DynamoDB-local can raise an (undocumented) `Invalid ShardId in ShardIterator` exception. Since we poll a stream in ESM, encountering this sharding issue should necessitate a re-initialise of the shards we are polling.

## Changes
- Check whether ResourceNotFound exception is of type `Invalid ShardId in ShardIterator`, and trigger a re-sharding if encountered.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/dynamodb_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/dynamodb_poller.py
@@ -74,18 +74,6 @@ class DynamoDBPoller(StreamPoller):
             "eventVersion": "1.1",
         }
 
-    def get_records(self, shard_iterator: str) -> dict:
-        try:
-            return super().get_records(shard_iterator)
-        except self.source_client.exceptions.TrimmedDataAccessException as e:
-            LOG.debug(
-                "Attempted to iterate over trimmed record or expired shard iterator %s for stream %s, re-initializing shards",
-                shard_iterator,
-                self.source_arn,
-            )
-            self.shards = self.initialize_shards()
-            raise e
-
     def transform_into_events(self, records: list[dict], shard_id) -> list[dict]:
         events = []
         for record in records:
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/dynamodb_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/dynamodb_poller.py
@@ -74,6 +74,18 @@ class DynamoDBPoller(StreamPoller):
             "eventVersion": "1.1",
         }
 
+    def get_records(self, shard_iterator: str) -> dict:
+        try:
+            return super().get_records(shard_iterator)
+        except self.source_client.exceptions.TrimmedDataAccessException as e:
+            LOG.debug(
+                "Attempted to iterate over trimmed record or expired shard iterator %s for stream %s, re-initializing shards",
+                shard_iterator,
+                self.source_arn,
+            )
+            self.shards = self.initialize_shards()
+            raise e
+
     def transform_into_events(self, records: list[dict], shard_id) -> list[dict]:
         events = []
         for record in records:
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Thanks a lot for tackling this! 🙏 

I'll probably open a follow up soon with the `TrimmedDataAccessException` exception sometimes raised by DynamoDB-local and also considered to be a bug. 

I think this is a safe fix, and if anything goes wrong with Dynamodb-local, it will also flood the logs, so we should be good to go. Hopefully in the future we can rework the sharding behavior!

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -264,7 +264,10 @@ def get_records(self, shard_iterator: str) -> dict:
             )
             return get_records_response
         # TODO: test iterator expired with conditional error scenario (requires failure destinations)
-        except self.source_client.exceptions.ExpiredIteratorException as e:
+        except (
+            self.source_client.exceptions.ExpiredIteratorException,
+            self.source_client.exceptions.TrimmedDataAccessException,
```
bentsku, line 269: Ah, that's nice, I thought we'd need to do string matching again, but both `dynamodbstreams` and the `kinesis` clients implements the `TrimmedDataAccessException`? That's practical 😄

gregfurman: Kinesis doesn't raise this ://// will re-implement

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/dynamodb_poller.py
```diff
@@ -74,6 +74,18 @@ def extra_metadata(self) -> dict:
             "eventVersion": "1.1",
         }
 
+    def get_records(self, shard_iterator: str) -> dict:
+        try:
+            return super().get_records(shard_iterator)
+        except self.source_client.exceptions.TrimmedDataAccessException as e:
```
bentsku, line 80: I think this is not going to be caught because this is a subclass from `ClientError`, and the parent class is already catching that, and if not a "recognized" exception, raised `PipeInternalError from e`. Not sure if we can catch that somehow? Ugh, sorry for this one... 😓

bentsku: I guess we could try to catch `PipeInternalError`, verify if it is raised from the exception we want (not even sure how to do that 👀 ) and if not, re-raise?  Not great... thanks, DynamoDB-local 😅

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -264,7 +264,7 @@ def get_records(self, shard_iterator: str) -> dict:
             )
             return get_records_response
         # TODO: test iterator expired with conditional error scenario (requires failure destinations)
-        except self.source_client.exceptions.ExpiredIteratorException as e:
+        except self.source_client.exceptions.ExpiredIteratorExceptionas as e:
```
bentsku, line 267:
```suggestion
        except self.source_client.exceptions.ExpiredIteratorException as e:
```

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -283,12 +283,28 @@ def get_records(self, shard_iterator: str) -> dict:
...
+                    )
+                    self.initialize_shards()
+                else:
+                    LOG.warning(
+                        "Source stream %s does not exist: %s",
+                        self.source_arn,
+                        e,
+                    )
+                    raise CustomerInvocationError from e
+            elif "TrimmedDataAccessException" in str(e):
```
bentsku, line 300: thank you so much for adding this! 🙏 I'll tell the users next week about it for them to confirm to us it solves the issue and they don't need to run the command manually anymore! Thank you very much for going the extra mile here 🙇

bentsku (approved):
LGTM! Thanks a lot really for adding the `TrimmedDataAccessException` too 🙇

Issue #11255, opened 2024-07-23
bug: Step Functions CloudFormation "update" handler doesn't support "DefinitionS3Location" field


northvankiwiguy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using CloudFormation or CDK to deploy a Step Functions state machine, it is not possible to update the CloudFormation stack because the `DefinitionS3Location` directive is not supported by the Step Functions CloudFormation plugin.

When I run `cdklocal deploy` for the second time, I see the error:

```
localstack-main  | 2024-07-23T22:07:40.792  WARN --- [-functhread6] l.s.c.deployment_utils     : Unexpected error processing resource type AWS::StepFunctions::StateMachine: Exception: 'DefinitionString' - {'Name': 'StateMachineName', 'RoleArn': 'arn:aws:iam::000000000000:role/Admin', 'DefinitionS3Location': {'Bucket': 'assets', 'Key': 'state-machine-2.asl.json'}, 'Arn': 'arn:aws:states:us-east-1:000000000000:stateMachine:Buggy-StateMachine-8c880a39'} - status: None
```

The state machine definition is not correctly updated.

### Expected Behavior

A second (or successive) `cdklocal deploy` should update the state machine correctly without crashing.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
    docker-compose up
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
awslocal s3 mb s3://assets
awslocal s3 cp state-machine-1.asl.json s3://assets/state-machine-1.asl.json
awslocal s3 cp state-machine-2.asl.json s3://assets/state-machine-2.asl.json
awslocal cloudformation create-stack --stack-name Buggy --template-body file://cloudformation-1.json
```

The failure happens when executing this command:
```
awslocal cloudformation update-stack --stack-name Buggy --template-body file://cloudformation-2.json
```

Where `state-machine-N.asl.json` is:

```
{
  "StartAt": "SuccessN",     # set N to 1 or 2
  "States": {
    "SuccessN": {
      "Type": "Succeed"
    }
  }
}
```

and `cloudformation-N.json` is:

```{
  "Resources": {
    "StateMachine": {
      "Type": "AWS::StepFunctions::StateMachine",
      "Properties": {
        "Name": "StateMachineName",
[...]

PR #12044, opened 2024-12-16
CloudFormation, Fix Retrieval of State Machine Definitions On Stack Update

MEPalma:
## Motivation
Currently, CloudFormation encounters issues when updating state machines during StackUpdate operations, as it incorrectly assumes the state machine definition is always provided as a literal in the DefinitionString field. This assumption causes errors when using alternative methods for defining state machines, such as through DefaultS3Location https://github.com/localstack/localstack/issues/11255.
This update introduces tests for CloudFormation templates that create state machines using DefaultS3Location and ensures that during stack updates, the definition string is correctly extracted by sampling the available alternatives.

## Changes
- Snapshot tests for state machine creation and update through DefaultS3Location
- Adjusted the update routine to extract the state machine definition


pinzon (approved):
Thank you for addressing this. Looks Good. 
Just some nits in the test.

PR #12047, opened 2024-12-17
DynamoDB: Fix empty replicas list in response

viren-nadkarni:
## Changes

This PR fixes a regression caused by https://github.com/localstack/localstack/pull/11938 which would cause an empty list to be returned when no replicas are configured ([reference test run](https://app.circleci.com/pipelines/github/localstack/localstack/30313/workflows/a3d53cf9-5726-4dbb-8fee-5b90a7b60cf5/jobs/268546/tests)). The behaviour is AWS validated by the tests in the reference run.

## Tests

Randomised MA/MR tests: https://app.circleci.com/pipelines/github/localstack/localstack/30321/workflows/ade8da0d-d52a-453e-8c4f-efc418e5d8c3

### DIFF
--- a/localstack-core/localstack/services/dynamodb/provider.py
+++ b/localstack-core/localstack/services/dynamodb/provider.py
@@ -752,7 +752,8 @@ class DynamoDBProvider(DynamodbApi, ServiceLifecycleHook):
             if replica_region != context.region:
                 replica_description_list.append(replica_description)
 
-        table_description.update({"Replicas": replica_description_list})
+        if replica_description_list:
+            table_description.update({"Replicas": replica_description_list})
 
         # update only TableId and SSEDescription if present
         if table_definitions := store.table_definitions.get(table_name):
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/dynamodb/provider.py
```diff
@@ -752,7 +752,8 @@ def describe_table(
             if replica_region != context.region:
                 replica_description_list.append(replica_description)
 
-        table_description.update({"Replicas": replica_description_list})
+        if replica_description_list:
+            table_description.update({"Replicas": replica_description_list})
```
bentsku, line 756: nit: I think we could do `table_description["Replicas"] = replica_description_list` instead, but it was that way before so let's keep it, only a nit anyway 😄

bentsku (approved):
I'll review as I was, I think, by mistake pinged on the pipeline alert by Alex 😅 LGTM, I'll also merge the PR to unblock the pipeline, hope that's alright!

Issue #11703, opened 2024-10-17
bug: Update cbor2 requirement


hammer498:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently cbor is listed as required for >= 5.2
https://github.com/localstack/localstack/blob/master/pyproject.toml#L60 
but this commit
https://github.com/localstack/localstack/commit/b612563c223423fb808f7117ba4eb0043a49f7f1
relies on the existence of `cbor2._decoder` which appears first in cbor 5.5
https://github.com/agronholm/cbor2/tree/5.5.0/cbor2
notice its absence here
https://github.com/agronholm/cbor2/tree/5.4.6/cbor2

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal s3 mb s3://mybucket

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version:
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

PR #12051, opened 2024-12-18
bump minimum version of cbor

alexrashed:
## Motivation
Fixes https://github.com/localstack/localstack/issues/11703.
As described in the issue, we should update the minimum requirement for the `cbor` library, as we depend on some code which was introduced with `cbor2==5.5.0`.

## Changes
- Update the minimum version for `cbor`. No other changes are necessary, since in our dependency lock files we already pin the latest version (`5.6.5`).


silv-io (approved):
LGTM!

PR #12053, opened 2024-12-18
Improve type safety of `container_exists`

martinfrancois:
## Motivation

I added type annotations to the `container_exists` function to improve type safety, facilitate static type checking, and improve developer experience.  

## Changes
While the code's external behavior remains unchanged, these changes improve the readability and maintainability of the code.

### DIFF
--- a/tests/cli/test_cli.py
+++ b/tests/cli/test_cli.py
@@ -27,7 +27,7 @@ def runner():
     return CliRunner()
 
 
-def container_exists(client, container_name):
+def container_exists(client: ContainerClient, container_name: str) -> bool:
     try:
         container_id = client.get_container_id(container_name)
         return True if container_id else False


martinfrancois:

martinfrancois:
I unfortunately can't add a label to the PR, can someone please add the label `semver: patch`? Thanks in advance!

alexrashed (commented):
Thanks, @martinfrancois for the contribution!
Additional type hints are always great! But to be honest, I am not sure if your change on the return statement really is more readable. 😄 
This is obviously quite subjective, and I'm happy to discuss this further, but I don't really see a reason to explicitly change this line and given that there isn't much else to the PR, I am leaning towards closing this PR...

thrau (changes requested):
thanks for the contribution @martinfrancois! agree with @alexrashed here that i don't think the return condition is more readable than before. but i like that were adding type hints!

martinfrancois:
Hi @alexrashed and @thrau, thanks a lot for the review! I totally understand; the return condition is indeed a matter of preference. I now dropped the commit with the change to the return condition and left the one with the type safety improvements :)

thrau (approved):
thanks for including our review :-) type hints are always helpful :+1:

PR #12059, opened 2024-12-20
fix client factory for bypassing DNS server

bentsku:
## Motivation
It seems our tests broke due to maybe some change in how AWS was returning their DNS records, and it seems we return the full chain of resolution.

When doing the following request:
`request = dns.message.make_query(hostname, "A")`
We would get a full answer containing `CNAME` records:
`[<DNS sts.amazonaws.com. IN CNAME RRset: [<reg.sts-ge.amazonaws.com.>]>, <DNS reg.sts-ge.amazonaws.com. IN CNAME RRset: [<use1-default-r.sts-ge.amazonaws.com.>]>, <DNS use1-default-r.sts-ge.amazonaws.com. IN CNAME RRset: [<sts-global-default.us-east-1.amazonaws.com.>]>, <DNS sts-global-default.us-east-1.amazonaws.com. IN A RRset: [<209.54.177.164>]>]`

Which then lead to an issue when trying to access the `address` field:
`AttributeError: 'CNAME' object has no attribute 'address'`

I've introduced a filter step to make sure we're only trying to get `A` records before choosing. However, I'm not quite sure if this is expected behavior, I believe that might be a change in how AWS returns their DNS and introducing a few resolving steps in between. I don't know if we make the query better? Should we only get the last answer instead of filtering? Something like `list(response.answer[-1].items.keys())` instead? My knowledge of DNS stops here 😄 

Manual run for bootstraps tests: #/12434395225

## Changes
- filter the returned `list[RRSet]` to only contains `A` type before trying to access the `address` field.

### DIFF
--- a/localstack-core/localstack/aws/connect.py
+++ b/localstack-core/localstack/aws/connect.py
@@ -660,7 +660,11 @@ def resolve_dns_from_upstream(hostname: str) -> str:
     if len(response.answer) == 0:
         raise ValueError(f"No DNS response found for hostname '{hostname}'")
 
-    ip_addresses = list(response.answer[0].items.keys())
+    ip_addresses = []
+    for answer in response.answer:
+        if answer.match(dns.rdataclass.IN, dns.rdatatype.A, dns.rdatatype.TYPE0):
+            ip_addresses.extend(answer.items.keys())
+
     return choice(ip_addresses).address
 
 
--- a/localstack-core/localstack/aws/connect.py
+++ b/localstack-core/localstack/aws/connect.py
@@ -665,6 +665,9 @@ def resolve_dns_from_upstream(hostname: str) -> str:
         if answer.match(dns.rdataclass.IN, dns.rdatatype.A, dns.rdatatype.TYPE0):
             ip_addresses.extend(answer.items.keys())
 
+    if not ip_addresses:
+        raise ValueError(f"No DNS records of type 'A' found for hostname '{hostname}'")
+
     return choice(ip_addresses).address
 
 
--- a/localstack-core/localstack/aws/connect.py
+++ b/localstack-core/localstack/aws/connect.py
@@ -662,7 +662,7 @@ def resolve_dns_from_upstream(hostname: str) -> str:
 
     ip_addresses = []
     for answer in response.answer:
-        if answer.match(dns.rdataclass.IN, dns.rdatatype.A, dns.rdatatype.TYPE0):
+        if answer.match(dns.rdataclass.IN, dns.rdatatype.A, dns.rdatatype.NONE):
             ip_addresses.extend(answer.items.keys())
 
     if not ip_addresses:


localstack-core/localstack/aws/connect.py
```diff
@@ -660,7 +660,11 @@ def resolve_dns_from_upstream(hostname: str) -> str:
     if len(response.answer) == 0:
         raise ValueError(f"No DNS response found for hostname '{hostname}'")
 
-    ip_addresses = list(response.answer[0].items.keys())
+    ip_addresses = []
+    for answer in response.answer:
+        if answer.match(dns.rdataclass.IN, dns.rdatatype.A, dns.rdatatype.TYPE0):
```
simonrw, line 665: q: what is `TYPE0` in this context? The argument is called `covers` but I don't know what that means.

Does this mean "anything" as `0` is not a special value perhaps?

simonrw (approved):
It looks like the DNS response from 8.8.8.8 (the default upstream) is

```
; <<>> DiG 9.18.28 <<>> @8.8.8.8 sts.amazonaws.com A
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 29178
;; flags: qr rd ra; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;sts.amazonaws.com.		IN	A

;; ANSWER SECTION:
sts.amazonaws.com.	3365	IN	CNAME	reg.sts-ge.amazonaws.com.
reg.sts-ge.amazonaws.com. 2521	IN	CNAME	use1-default-r.sts-ge.amazonaws.com.
use1-default-r.sts-ge.amazonaws.com. 2199 IN CNAME sts-global-default.us-east-1.amazonaws.com.
sts-global-default.us-east-1.amazonaws.com. 12 IN A 54.239.29.25

;; Query time: 19 msec
;; SERVER: 8.8.8.8#53(8.8.8.8) (UDP)
;; WHEN: Fri Dec 20 16:47:56 GMT 2024
;; MSG SIZE  rcvd: 159
```

which is very strange, but consistent with the behaviour we see. Thanks for making this change!

localstack-core/localstack/aws/connect.py
bentsku: Good question. 😬 I hoped you'd be able to answer this, I've printed all the different `answer` objects and printed the `covers` field of them, and all of them returned `TYPE0`. Looking more into it, it seems related with `RRSIG` record types: https://simpledns.plus/help/rrsig-records

https://dnspython.readthedocs.io/en/latest/rdata-set-classes.html#dns.node.Node.find_rdataset

> covers, a dns.rdatatype.RdataType, the covered type. Usually this value is dns.rdatatype.NONE, but if the rdtype is dns.rdatatype.SIG or dns.rdatatype.RRSIG, then the covers value will be the rdata type the SIG/RRSIG covers. The library treats the SIG and RRSIG types as if they were a family of types, e.g. RRSIG(A), RRSIG(NS), RRSIG(SOA). This makes RRSIGs much easier to work with than if RRSIGs covering different rdata types were aggregated into a single RRSIG rdataset.

So I suppose it does not apply to type `A` records and can safely be put to `TYPE0` or `NONE` (both are equal to `0` in the Enum, so I've replaced `TYPE0` by `NONE` which is a bit more explicit?

PR #12060, opened 2024-12-20
Cfn support layer version read list

cloutierMat:
## Motivation

Implementation of the `read` and `list` operations for lambda layer versions.

The cloudformation schema were out of date so they were also updated as part of this pr.

## Changes

- Fix a bug in `get_layer_versions` where `store.layers.get(layer_name, {}).layer_versions.get(layer_version)` would fail when the lambda layer did not exists
- Updated the Primary Identifier from `Id` to `LayerVersionArn` with a validated test
- Implementation of `read` and `list`


pinzon (approved):
LGTM. Great to have the schema updated. :+1:

PR #12068, opened 2024-12-23
StepFunctions: Stateless evaluation of EvalComponents

MEPalma:
## Motivation
In the StepFunctions v2 interpreter, Component and EvalComponent objects must implement a stateless evaluation of their logic. This model is in place to ensure that components are not only reusable, but also usable in parallel. Such feature is useful especially in Map and Parallel states which apply the same logic multiple times and concurrently. 

By closing the following [issue](https://github.com/localstack/localstack/pull/12033), on one hand we did ensure nested distributed map states could be evaluated, but on the other hand we are now in a position to expose all implementation error related to the stateless fashion of evaluation components. This may lead to users experiencing inconsistencies in the evaluation of intrinsic functions or the retry logic within MapRun or parallel states.

These changes further ensure that the evaluation of all EvalComponents are stateless: heap scope strategies, caching of evaluation values in the component object, and shared member values; are moved to the scope of each eval call.

## Changes
- Revised the argument components of intrinsic functions to evaluate in a stateless manner
- Refactored the environment's heap strategy to manage values exclusively within the current scope. This is possible as there is no inherent need to access values in higher scopes and prevents components in different threads from overriding the same memory locations (such as for `MaxAttempts` and `BackoffRate`)
- Minor changes to the test suite: uniform lambda runtime fields


localstack-core/localstack/services/stepfunctions/asl/component/intrinsic/argument/argument.py
```diff
@@ -0,0 +1,92 @@
...
+
+from localstack.services.stepfunctions.asl.component.common.string.string_expression import (
+    StringVariableSample,
+)
+from localstack.services.stepfunctions.asl.component.eval_component import EvalComponent
+from localstack.services.stepfunctions.asl.eval.environment import Environment
+from localstack.services.stepfunctions.asl.utils.json_path import extract_json
+
+
+class Argument(EvalComponent, abc.ABC):
```
joe4dev, line 12: doc: A docstring explaining the motivation behind the argument abstraction would be very helpful here.

localstack-core/localstack/services/stepfunctions/asl/component/intrinsic/argument/argument.py
```diff
@@ -0,0 +1,92 @@
...
+    @abc.abstractmethod
+    def _eval_argument(self, env: Environment) -> Any: ...
+
+    def _eval_body(self, env: Environment) -> None:
+        argument = self._eval_argument(env=env)
+        env.stack.append(argument)
+
+
+class ArgumentLiteral(Argument):
```
joe4dev, line 21: These Argument* implementations seem identical apart from inheriting the new Argument abstraction.
I think they are easier to understand them grouped here rather than spread across individual files 👍

localstack-core/localstack/services/stepfunctions/asl/component/program/program.py
```diff
@@ -87,18 +87,11 @@ def eval(self, env: Environment) -> None:
     def _eval_body(self, env: Environment) -> None:
         try:
             while env.is_running():
-                # Store the heap values at this depth for garbage collection.
-                heap_values = set(env.heap.keys())
```
joe4dev, line 91: That's a very nice simplification ✨ 👏 

Do I understand correctly that we now have just one global heap instead of `heap_values` that only applied at a certain level?

tests/aws/services/stepfunctions/v2/callback/test_callback.py
```diff
@@ -467,7 +468,7 @@ def test_start_execution_sync_delegate_timeout(
         lambda_creation_response = create_lambda_function(
             func_name=function_name,
             handler_file=TT.LAMBDA_WAIT_60_SECONDS,
-            runtime="python3.9",
+            runtime=Runtime.python3_12,
```
joe4dev, line 471: Thanks for unifying these, potentially saving us some image-pulling time 🚄

joe4dev (approved):
The simplification of the heap memory model is amazing and greatly improves maintainability 💯 

The changes look good to me. Small suggestion: Adding documentation that motivates/explains the argument abstraction would improve understandability and maintainability.

localstack-core/localstack/services/stepfunctions/asl/component/program/program.py
MEPalma: Prior to these changes, the heap was shared globally, with scopes enforced through UUID prefixes attached to value identifiers. This approach worked because we never encountered a use case where accessing values from an outer scope was necessary. However, it led to a significant issue: UUID prefixes were assigned to each `EvalComponent` on demand but needed to remain fixed to allow consistent retrieval during iterative scenarios, such as retry counters in retry blocks. 

This setup was problematic since the same `EvalComponent` could execute on multiple threads (e.g., in distributed Map or Parallel operations). As a result, separate threads could inadvertently read and write each other’s data. 

Given that outer scope access wasn’t required, the changes instead introduce a new heap reference for each scope. In the implementation, we extend the concept of scope (related to memory access) to that of a frame, which also includes other SFN-related data such as execution events, logging, etc. Hence the updates made to `as_inner_frame_of`.

Issue #11997, opened 2024-12-05
bug: Step Functions incorrectly rejects valid time format in wait state


peter-smith-phd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following state machine is incorrectly rejected by `create-state-machine`

```
{
  "StartAt": "pass",
  "States": {
    "pass": {
      "Type": "Pass",
      "Next": "wait"
    },
    "wait": {
      "Type": "Wait",
      "Timestamp": "2024-12-05T21:29:29.243167252Z",
      "Next": "pass1"
    },
    "pass1": {
      "Type": "Pass",
      "Result": 1,
      "End": true
    }
  }
}

```

This gives the error:

```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: Error=ValueError Args=["time data '2024-12-05T21:29:29.243167252Z' does not match format '%Y-%m-%dT%H:%M:%SZ'"] in definition.
```

### Expected Behavior

The state machine should be accepted as valid.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal --region us-east-1 stepfunctions create-state-machine --name "Fail_errors" --definition file://file.asl.json --role-arn arn:aws:iam::000000000000:role/Admin

### Environment

```markdown
- OS:
- LocalStack:
  LocalStack version: 4.0.4.dev5
  LocalStack build date: 2024-12-03
  LocalStack build git hash: 78fdf458f
```

### Anything else?

Thanks Marco, you're the best!

MEPalma:
@peter-smith-phd Thank you for bringing this to our attention. I was able to replicate the issue with timestamp string literal validation and wanted to let you know that we’ve recently merged changes to address it.

We’ve also improved the error messages for both Timestamp and TimestampPath to better mimic those obtainable from AWS for consistency.

Could I kindly ask you to verify if this issue is now resolved on your end?

Thank you for helping the StepFunctions provider improve!

peter-smith-phd:
I'm aiming to get this issue verified in the week of Feb 10th. Thanks!

peter-smith-phd:
Thanks, this state machine definition now parses correctly.

PR #12074, opened 2024-12-27
StepFunctions: Improvements to Timestamp Validation

MEPalma:
## Motivation
The Step Functions (SFN) v2 interpreter has certain limitations in validating timestamp strings for the Timestamp field. This field can be set statically using string literals or dynamically, such as via JSONata expressions. While validation exists for TimestampPath fields, similar logic was not applied to Timestamp fields. Consequently, some timestamps failed validation during state machine creation, as described in https://github.com/localstack/localstack/issues/11997.

These changes enhance the validation logic to ensure both statically provided and dynamically computed timestamp values are checked for proper format and a valid date. Additionally, a series of positive and negative test cases have been added to verify timestamp handling.

It is worth mentioning that dynamically computed invalid timestamps via JSONata expressions result in a generic execution failure rather than a new evaluation failure. This discrepancy is limited to the error type, while the failure description correctly conveys the issue. Improvements to this behaviour are planned as we introduce validation patterns for JSONata expression outcomes.

## Changes
- Enhanced timestamp validation logic for both static and dynamically computed values
- Added positive and negative snapshot tests for `Timestamp*` fields


localstack-core/localstack/services/stepfunctions/asl/component/state/state_wait/wait_function/timestamp.py
```diff
@@ -60,31 +72,34 @@ def _create_failure_event(self, env: Environment, timestamp_str: str) -> Failure
             event_details=EventDetails(
                 executionFailedEventDetails=ExecutionFailedEventDetails(
                     error=StatesErrorNameType.StatesRuntime.to_name(),
-                    cause=f"The TimestampPath parameter does not reference a valid ISO-8601 extended offset date-time format string: {self.string.literal_value} == {timestamp_str}",
+                    cause="The Timestamp parameter does not reference a valid ISO-8601 "
+                    f"extended offset date-time format string: {self.string.literal_value} == {timestamp_str}",
```
tiurin, line 76: question: why do we need to print both literal value of object property and `timestamp_str` parameter? Also, what `==` here means?

tiurin (approved):
Great to see improved test coverage! 👍

localstack-core/localstack/services/stepfunctions/asl/component/state/state_wait/wait_function/timestamp.py
MEPalma: This error message is composed in an effort to match AWS's error message in these circumstances ([example](https://github.com/localstack/localstack/pull/12074/files#diff-eeb7e1991df56055fcb4c375135b4d5ce96eccd48aa0c94167d5262af3a3c101R25699)).
The message lists how the expression provided for sampling the timestamp `{self.string.literal_value}` evaluates to the invalid timestamp value `{timestamp_str}`. The error message may look like this: `The TimestampPath parameter does not reference a valid ISO-8601 extended offset date-time format string: $.TimestampValue == 2016-12-05 21:29:29Z"`

localstack-core/localstack/services/stepfunctions/asl/component/state/state_wait/wait_function/timestamp.py
```diff
@@ -28,30 +28,42 @@
...
 
     def __init__(self, string: StringExpression):
         self.string = string
-        # If it's a string literal, assert it encodes a timestamp
+        # If a string literal, assert it encodes a valid timestamp.
         if isinstance(string, StringLiteral):
-            parse_timestamp(string.literal_value)
+            timestamp = string.literal_value
+            if (
+                not self._is_valid_timestamp_pattern(timestamp)
```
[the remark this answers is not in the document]
MEPalma: Indeed, thank you for spotting this!

tiurin (approved):

Issue #12048, opened 2024-12-17
bug: Step Functions throw InvalidDefinition exception for a valid JSONata definition


tereschenkov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm using the following definition
```json
{
  "QueryLanguage": "JSONata",
  "Comment": "A description of my state machine",
  "StartAt": "Choice",
  "States": {
    "Choice": {
      "Type": "Choice",
      "Choices": [
        {
          "Next": "Fail",
          "Condition": "{% $states.input.status != \"success\" %}",
          "Output": {
            "execution_id": "{% $ExecutionId %}",
            "notification_type": "{% $states.input.status %}"
          }
        }
      ],
      "Default": "Success"
    },
    "Fail": {
      "Type": "Fail"
    },
    "Success": {
      "Type": "Succeed"
    }
  }
}
```

When I execute
```
awslocal stepfunctions create-state-machine --name "test" \
      --role-arn "arn:aws:iam::000000000000:role/step_function_role" \
      --definition file://test.json \
      --type STANDARD
```

the output is 
```
[...]

PR #12075, opened 2024-12-27
Step Functions, Support for Output Blocks in Choice Rules, Improvments to JSONata Choice Defaults

MEPalma:
## Motivation
The SFN v2 interpreter currently lacks support for `Output` block declarations within choice rules, as noted in https://github.com/localstack/localstack/issues/12048. Additionally, the current implementation of the choice state does not handle `Output` and `Assign` blocks as intended, since it always evaluates the default `Assign` and `Output` blocks before returning. 
A further issue lies in the use of a member variable to track the next state name, which has been identified as an antipattern and addressed in recent changes https://github.com/localstack/localstack/pull/12068.

## Changes
- Adds support for `Output` blocks in choice rules.
- Refines the evaluation logic for default `Assign` and `Output` blocks.
- Resolves the evaluation component antipattern. 
- Snapshot tests for `Output` blocks in choice rules
- Snapshot tests for default `Ouput` and `Assign` evaluation logic


localstack-core/localstack/services/stepfunctions/asl/component/state/state_choice/state_choice.py
```diff
@@ -55,8 +54,56 @@ def _eval_state(self, env: Environment) -> None:
...
+        # Filter the input onto the stack.
+        if self.input_path:
+            self.input_path.eval(env)
+        else:
+            env.stack.append(env.states.get_input())
+
+        # Exec the state's logic.
+        self._eval_state(env)
+
+        # Handle legacy output sequences if in JsonPath mode.
```
tiurin, line 94: question: Did you need to override base `_eval_body` method because of this legacy handling? If yes, is there any test that cover this?

localstack-core/localstack/services/stepfunctions/asl/component/state/state_choice/state_choice.py
```diff
@@ -55,8 +54,56 @@ def _eval_state(self, env: Environment) -> None:
...
+        )
+
+        # Filter the input onto the stack.
+        if self.input_path:
+            self.input_path.eval(env)
+        else:
+            env.stack.append(env.states.get_input())
+
+        # Exec the state's logic.
+        self._eval_state(env)
```
tiurin, line 92: question: in the base method this block is wrapped in a try-catch block that handles `NoSuchJsonPathError`. Why is it not needed here?

tiurin: Also, there is [another generic check in base method](https://github.com/localstack/localstack/blob/019199313b17e8d26d6cdb219a88d1a6b6025c30/localstack-core/localstack/services/stepfunctions/asl/component/state/state.py#L234-L235) that is not present here:
```
        if not isinstance(env.program_state(), ProgramRunning):
            return
```
Is there anything specific about Choice state that doesn't require the check?

tiurin (approved):
LGTM!
I'd appreciate a bit more information about the differences from base `_eval_body` method. It is not obvious why it needs to be overridden. Override that mostly repeats base method with slight differences adds additional maintenance effort.
As a suggestion to follow up later, maybe we could analyze if `super()._eval_body(env=env)` can be called from override instead of copying most parts of it?

localstack-core/localstack/services/stepfunctions/asl/component/state/state_choice/state_choice.py
MEPalma: I refactored the reusable parts of states' evaluation into: loading the state input, evaluating the state logic, and evaluation the state output. whilst the logical glue of these three steps is being reused in the _eval_body of CommonStateField.
We do have many snpahost tests for legacy (or JSONPath) choice states in `aws.services.stepfunctions.v2.scenarios`

tiurin: Great refactoring! 🎉 
Now all special cases are handled within their respective states and common method is clean and readable! 👏

tiurin (approved):

Issue #12066, opened 2024-12-23
bug: <title>


asomov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Creating AWS::KinesisFirehose::DeliveryStream without optional DeliveryStreamName raises an exception

```
localstack-main  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
localstack-main  |   File "/opt/code/localstack/localstack-core/localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py", line 416, in create
localstack-main  |     stream = firehose.describe_delivery_stream(DeliveryStreamName=model["DeliveryStreamName"])
localstack-main  |                                                                   ~~~~~^^^^^^^^^^^^^^^^^^^^^^
localstack-main  | KeyError: 'DeliveryStreamName'
```

### Expected Behavior

AWS::KinesisFirehose::DelCreating iveryStream without optional DeliveryStreamName should succedd and the name should be generated.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

start LocalStack
Create AWS::KinesisFirehose::DeliveryStream (with CloudFormation)

### Environment

```markdown
- OS: MacOS 15.2 (24C101)
- LocalStack:
  LocalStack version: 4.0.4.dev29
  LocalStack Docker image sha: 5f259341c488
  LocalStack build date:  2024-12-10
  LocalStack build git hash: c32c40104
```

### Anything else?

@Morijarti [the code in line 417](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/services/kinesisfirehose/resource_providers/aws_kinesisfirehose_deliverystream.py#L416) assumes that model has "DeliveryStreamName"
Probably the line should be chaged to this (because the name is generated in lines 398-402):
`        stream = firehose.describe_delivery_stream(DeliveryStreamName=attr["DeliveryStreamName"])`

@viren-nadkarni @alexrashed @steffyP [the tests assume that the stream creation should fail](https://github.com/localstack/localstack/blob/master/tests/aws/services/kinesis/test_kinesis.py#L115), but it shoud not

asomov:
[a proposal](https://github.com/localstack/localstack/pull/12078)

PR #12078, opened 2024-12-28
WIP - create Firehose streams without name

asomov:
## Motivation
Create Firehose streams without names with CloudFormation
It solves [this issue](https://github.com/localstack/localstack/issues/12066) 

## Changes
This is a work in progress to show a possible solution. 
- `provider_utils.generate_default_name()` - do not cut the name of the resource to be able to find it later
- add a few TODOs to raise a concern
- try to guess the stream name during REPEATED_INVOCATION (@Morijarti @alexrashed)
- use **addr** instead of `model`
`stream = firehose.describe_delivery_stream(DeliveryStreamName=`**attrs**`["DeliveryStreamName"])`
`stream = firehose.describe_delivery_stream(DeliveryStreamName=`**model**`["DeliveryStreamName"])`

## Testing
Unfortunately, I couls not run any tests. Neither with `make tests` nor within docker container
Any help is greatly appreciated!

## TODO
To finish the implementation to be able to create firehose streams wihout a name with CloudFormation (when the name is generated with stack name, name of the resorce and a random part)

### DIFF
--- a/localstack-core/localstack/services/cloudformation/deployment_utils.py
+++ b/localstack-core/localstack/services/cloudformation/deployment_utils.py
@@ -172,14 +172,14 @@ def select_parameters(*param_names):
 def is_none_or_empty_value(value):
     return not value or value == PLACEHOLDER_AWS_NO_VALUE
 
-
+# TODO: almost the same logic is in provider_utils.generate_default_name
 def generate_default_name(stack_name: str, logical_resource_id: str):
     random_id_part = short_uid()
     resource_id_part = logical_resource_id[:24]
     stack_name_part = stack_name[: 63 - 2 - (len(random_id_part) + len(resource_id_part))]
     return f"{stack_name_part}-{resource_id_part}-{random_id_part}"
 
-
+# TODO: almost the same logic is in provider_utils.generate_default_name_without_stack
 def generate_default_name_without_stack(logical_resource_id: str):
     random_id_part = short_uid()
     resource_id_part = logical_resource_id[: 63 - 1 - len(random_id_part)]
--- a/localstack-core/localstack/services/cloudformation/provider_utils.py
+++ b/localstack-core/localstack/services/cloudformation/provider_utils.py
@@ -17,14 +17,12 @@ from botocore.model import Shape, StructureShape
 
 
 def generate_default_name(stack_name: str, logical_resource_id: str):
-    random_id_part = str(uuid.uuid4())[0:8]
-    resource_id_part = logical_resource_id[:24]
-    stack_name_part = stack_name[: 63 - 2 - (len(random_id_part) + len(resource_id_part))]
-    return f"{stack_name_part}-{resource_id_part}-{random_id_part}"
+    random_id_part = str(uuid.uuid4())[0:8].upper()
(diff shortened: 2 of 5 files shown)


asomov:

asomov:
I do not know how to fix this:
`Error: Label error. Requires exactly 1 of: semver: patch, semver: minor, semver: major. Found: `

pinzon:
Hi @asomov. Good attempt but the changes seems excessive for what you want to achieve which is: **Allow the CFn to deploy a delivery stream without a Name defined in the template.** 

CFn allows the creation of many resources without a name even if the service API requires it. To do this it just defines a name in case is not defined. 

Here is an example of how is done in another CFn resource provider: https://github.com/localstack/localstack/blob/3b4cc72e1cd5b4477ba01c18a31d70fb2ff8905e/localstack-core/localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py#L62-L66

You would only need to do the same for the DeliveryStreamName attribute.

pinzon (changes requested):
Hi @asomov. Good attempt but the changes seems excessive for what you want to achieve which is: **Allow the CFn to deploy a delivery stream without a Name defined in the template.** 

CFn allows the creation of many resources without a name even if the service API requires it. To do this it just defines a name in case is not provided. 

Here is an example of how is done in another CFn resource provider: https://github.com/localstack/localstack/blob/3b4cc72e1cd5b4477ba01c18a31d70fb2ff8905e/localstack-core/localstack/services/cloudformation/resource_providers/aws_cloudformation_stack.py#L62-L66

You would only need to do the same for the DeliveryStreamName attribute.

asomov:
I agree with your comment, but I do understand how to call the same as this line for Firehose:
(this is used in your link in case of REPEATED_INVOCATION)
 stack = request.aws_client_factory.cloudformation.describe_stacks(StackName=model["Id"])["Stacks"][0]

pinzon:
Well the code would look like this:
```python
...
      parameters = [
            "DeliveryStreamName",
            "DeliveryStreamType",
            "S3DestinationConfiguration",
            "ElasticsearchDestinationConfiguration",
            "AmazonopensearchserviceDestinationConfiguration",
            "DeliveryStreamEncryptionConfigurationInput",
            "ExtendedS3DestinationConfiguration",
            "HttpEndpointDestinationConfiguration",
            "KinesisStreamSourceConfiguration",
            "RedshiftDestinationConfiguration",
            "SplunkDestinationConfiguration",
            "Tags",
        ]

        if not request.custom_context.get(REPEATED_INVOCATION):
            if not model.get("DeliveryStreamName"):
                model["DeliveryStreamName"] = util.generate_default_name(
                    request.stack_name, request.logical_resource_id
                )

            attrs = util.select_attributes(model, params=parameters)
            response = firehose.create_delivery_stream(**attrs)


...
```

We move the validation and generation of the `DeliveryStreamName` after the checking is the first time `create` is called.
Also we make sure that it's stored in the model. Then we put selection of the attributes (`attrs = util.select_attributes(model, params=parameters)`) before calling the creation.

asomov:
give up

PR #12083, opened 2024-12-31
DDB: parity fix for JS clients

giograno:
## Motivation
We got a report from a customer mentioning that an empty parameter list in an execute statement command does not raise any error against LocalStack when using the JS client.

For instance, the following snippet would raise a `ValidationException` error against AWS, but not against LocalStack.

```js
const params = {
  Statement: "SELECT * FROM Table", 
  Parameters: []
};
const command = new ExecuteStatementCommand(params);
await client.send(command)
```

## Changes
- Adding a check for an empty parameter list;
- Adding a snapshot test.

### DIFF
--- a/tests/aws/services/dynamodb/test_dynamodb.py
+++ b/tests/aws/services/dynamodb/test_dynamodb.py
@@ -9,6 +9,7 @@ import botocore.exceptions
 import pytest
 import requests
 from boto3.dynamodb.types import STRING
+from botocore.config import Config
 from botocore.exceptions import ClientError
 from localstack_snapshot.snapshots.transformer import SortingTransformer
 
@@ -748,8 +749,9 @@ class TestDynamoDB:
 
     @markers.aws.validated
     def test_dynamodb_execute_statement_empy_parameter(
-        self, dynamodb_create_table_with_parameters, snapshot, aws_client
+        self, dynamodb_create_table_with_parameters, snapshot, aws_client_factory
     ):
+        ddb_client = aws_client_factory(config=Config(parameter_validation=False)).dynamodb
         table_name = f"test_table_{short_uid()}"
         dynamodb_create_table_with_parameters(
             TableName=table_name,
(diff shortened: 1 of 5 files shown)


viren-nadkarni (commented):
The changes look great. I think this will need to be ported to DDB v2 provider too, did you run any checks there?

giograno:
> The changes look great. I think this will need to be ported to DDB v2 provider too, did you run any checks there?

The behavior was the same. I ported the fix over 👍

tests/aws/services/dynamodb/test_dynamodb.py
```diff
@@ -746,6 +746,34 @@ def test_dynamodb_batch_execute_statement(
...
+
+        aws_client.dynamodb.put_item(
+            TableName=table_name,
+            Item={"Artist": {"S": "The Queen"}, "SongTitle": {"S": "Bohemian Rhapsody"}},
+        )
+
+        statement = f"SELECT * FROM {table_name}"
+        with pytest.raises(Exception) as e:
+            aws_client.dynamodb.execute_statement(Statement=statement, Parameters=[])
+        snapshot.match("invalid-param-error", e.value)
```
bentsku, line 775: nit: It would be good to validate that the exception is a `ClientError` and is raised from the provider and not from boto. For this, you can disable validation of the client to get the proper exception raised in AWS too: we have a few examples in our tests:
https://github.com/localstack/localstack/blob/41ffa76891286fffa5ca72a7b501fd7ecbbd36c4/tests/aws/services/s3/test_s3_list_operations.py#L804

`ddb_client = aws_client_factory(config=Config(parameter_validation=False)).dynamodb` would allow you to pass incorrect input and still get the AWS error.

bentsku: Awesome, thanks a lot for addressing the comment at light's speed! ❤️🚀

giograno: Thank you for the nice suggestion 🚀 . I have never come across disabling parameter validation before.

viren-nadkarni (approved):
Thanks for fixing this @giograno !

PR #12086, opened 2025-01-02
Improve error message for a missing parameter value - provide its name

asomov:
## Motivation
Improve error message

## Changes
Provide the name of the parameter without a value

## Testing
Not tested - I do not know how to run tests

semver: patch


asomov:
What is this and how to fix it:
`Error: Label error. Requires exactly 1 of: semver: patch, semver: minor, semver: major. Found: `

pinzon (approved):
LGTM :+1:

simonrw (approved):
Thanks for this contribution, our error messages could definitely do with some work and this is an improvement. I have some minor suggestions about the wording but I think the change is fine as it is. I will merge this. Thanks again!🎉

Issue #11927, opened 2024-11-26
feature request: Step Function Lambda Invocations Missing IAM Policy Checks 


drauedo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

To implement IAM policy stream on StepFunctions in order to test IAM policies locally.

MEPalma:
Thank you for taking the time to compile this report. We have recently added support for `Credentials`. Could you verify if this resolves your issue? If it does, feel free to close this issue. If you're still experiencing problems, please let us know and we'll investigate further. Thank you again for bringing this issue forward.

kbond-rxss:
My issue was not cross account permission checks, but just the fact that when I run a Step function in LocalStack I expect IAM policy checks to happen from the step function and they do not appear to be checked.  For example, if a step function accesses an S3 bucket for which it does not have permissions, this is silently allowed, and nothing is written to the IAM policy stream.  Likewise, lambda invocations that will fail in AWS because of IAM issues succeed without error in LocalStack.

MEPalma:
Thank you for the clarification. I was able to replicate the issue and have recently implemented some changes to address this limitation. These updates should ensure that IAM policy checks are enforced more accurately during Step Function service integration operations. Could you please verify if this works as expected in the latest version of LocalStack Pro using the `ENFORCE_IAM=1` flag?

kbond-rxss:
I see checks for access to S3 objects that are referenced in the step function, but I still don’t see a lambda invocation policy check for when the step function invokes a lambda.  I have soft failures enabled so that the earlier failures don’t prevent the invocations, but I am still expecting to see the check shown in the policy stream.

From: MEPalma ***@***.***>
Date: Thursday, January 2, 2025 at 2:11 PM
To: localstack/localstack ***@***.***>
Cc: Bond, Kevin ***@***.***>, Manual ***@***.***>
Subject: [EXTERNAL] Re: [localstack/localstack] feature request: Step Function Lambda Invocations Missing IAM Policy Checks (Issue #11927)
This email came from a source outside of CoverMyMeds. Use caution when clicking on links or replying with confidential information.
________________________________
Thank you for the clarification. I was able to replicate the issue and have recently implemented some changes to address this limitation. These updates should ensure that IAM policy checks are enforced more accurately during Step Function service
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
This message came from outside your organization.

ZjQcmQRYFpfptBannerEnd

[...]

PR #12089, opened 2025-01-02
StepFunctions: Ensure Service Integrations Assume State Machine Role Credentials

MEPalma:
## Motivation
Currently, the SFN v2 interpreter creates boto clients for service integrations using user credentials if no explicit credentials are provided. This causes IAM policy checks to succeed incorrectly event when `ENFORCE_IAM=1` is set, as the checks were based on the user's permissions rather than those of the state machine's assumed role https://github.com/localstack/localstack/issues/11927. This fix ensures boto clients are created with the assumed role credentials of the state machine.

## Changes
- In Task states, `Credentials` values are always computed and defaulted to the state machine role arn if not manual override is given
- Boto clients are created through assume role logic, using the role arn in the credentials
- Related minor refactoring to improve readability (mainly around renaming ComputedCredentials -> StateCredentials)

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task.py
@@ -18,8 +18,8 @@ from localstack.services.stepfunctions.asl.component.state.state_execution.execu
     ExecutionState,
 )
 from localstack.services.stepfunctions.asl.component.state.state_execution.state_task.credentials import (
-    ComputedCredentials,
     Credentials,
+    StateCredentials,
 )
 from localstack.services.stepfunctions.asl.component.state.state_execution.state_task.service.resource import (
     Resource,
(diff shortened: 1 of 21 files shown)


localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task.py
```diff
@@ -69,13 +69,13 @@ def _eval_parameters(self, env: Environment) -> dict:
 
         return parameters
 
-    def _eval_credentials(self, env: Environment) -> ComputedCredentials:
+    def _eval_state_credentials(self, env: Environment) -> StateCredentials:
         if not self.credentials:
-            task_credentials = dict()
+            state_credentials = StateCredentials(role_arn=env.aws_execution_details.role_arn)
```
dominikschubert, line 74: nice :raised_hands:

localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task.py
```diff
@@ -69,13 +69,13 @@ def _eval_parameters(self, env: Environment) -> dict:
...
         else:
             self.credentials.eval(env=env)
```
dominikschubert, line 76: Could this fail? if so, what would/should be the behavior here then? :thinking: 

Also curious if (not here) a failed assume role would also lead to falling back to the role ARN potentially? I'd assume not, since that might be a security issue.

dominikschubert (approved):
LGTM, great job figuring this out so quickly.

localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_task/state_task.py
MEPalma: The computation of the user-defined credentials block can fail, we have some test cases to validate our handling of such failures `aws.services.stepfunctions.v2.credentials.test_credentials_base.TestCredentialsBase.test_invalid_credentials_field`
In the tandem PR, which adds tests for SFN IAM enforcement, I did not see this behaviour, instead the failure is simply passed upstream

Issue #11335, opened 2024-08-09
bug: Events.Rule is not filtering by EventPattern


vladshoob:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Given there are multiple `AWS::Events::Rule`s triggering the same target (lambda) only last rule is applied, without filtering down to rules satisfying `EventPattern` property.

### Expected Behavior

All rules are applied that are filtered down with EventPattern.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Given localstack is up with `docker compose up`
```
# docker-compose.yml

services:
  localstack:
    image: localstack/localstack:latest
    environment:
      - PROVIDER_OVERRIDE_EVENTS=v2
      - DEBUG=1
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./tmp/localstack}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

And stack
```
# cloudformation.yaml`
Resources:
  LogGroup:
    Type: AWS::Logs::LogGroup
    Properties:
      LogGroupName: /aws/lambda/AwsLambdaMinimalExample

  LambdaFunction:
    Type: AWS::Lambda::Function
    Properties:
      FunctionName: AwsLambdaMinimalExample
      Handler: index.handler
      Runtime: nodejs20.x
      Role: !GetAtt LambdaFunctionRole.Arn
      MemorySize: 256
      Code:
        ZipFile: |
          exports.handler = async (event) => {
            console.log(process.env);
            console.log('event', JSON.stringify(event));
          
            return {
              statusCode: 200,
              body: JSON.stringify(event),
            }
          }

  LambdaFunctionRole:
    Type: AWS::IAM::Role
    Properties:
      AssumeRolePolicyDocument:
        Version: '2012-10-17'
        Statement:
          - Effect: Allow
            Principal:
              Service:
                - lambda.amazonaws.com
            Action:
              - sts:AssumeRole
      Path: "/"
      Policies:
        - PolicyName: AppendToLogsPolicy
          PolicyDocument:
            Version: '2012-10-17'
            Statement:
              - Effect: Allow
                Action:
                  - logs:CreateLogGroup
                  - logs:CreateLogStream
                  - logs:PutLogEvents
                Resource: "*"

[...]

PR #12090, opened 2025-01-02
Bugfix/eventbridge/process to all matching rules

maxhoheiser:
## Motivation
Internally target_services are stored in a dictionary _target_servuce_store and identified via the target_arn, this leads to problems since the target_arn is not unique for targets globally and across rules. 
The target_id is unique across targets for a single rule, thus a globally unique id is a combination of rule_arn and target_id

This problem also solves the following github bug report: https://github.com/localstack/localstack/issues/11335

## Changes
Changing the key for _target_servuce_store  to `rule_arn-target_id`

## Testing
Adding test for multiple rules with single or multiple patterns invoking the same target.
Adding tests for uniqueness of target_id across targets of single rule and across multiple rules
Adding tests for uniqueness of target_arn across targets of single rule and across multiple rules

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -1772,7 +1772,7 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
         target_sender = TargetSenderFactory(
             target, rule_arn, rule_name, region, account_id
         ).get_target_sender()
-        self._target_sender_store[target_sender.arn] = target_sender
+        self._target_sender_store[target_sender.unique_id] = target_sender
         return target_sender
 
     def create_archive_service(
@@ -1835,11 +1835,11 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
     def _delete_target_sender(self, ids: TargetIdList, rule) -> None:
         for target_id in ids:
             if target := rule.targets.get(target_id):
-                target_arn = target["Arn"]
+                target_unique_id = f"{rule.arn}-{target_id}"
                 try:
-                    del self._target_sender_store[target_arn]
+                    del self._target_sender_store[target_unique_id]
                 except KeyError:
-                    LOG.error("Error deleting target service %s.", target_arn)
+                    LOG.error("Error deleting target service %s.", target["Arn"])
 
     def _get_limited_dict_and_next_token(
         self, input_dict: dict, next_token: NextToken | None, limit: LimitMax100 | None
@@ -1889,8 +1889,8 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
                         "resources": [rule.arn],
                         "detail": {},
                     }
-
-                target_sender = self._target_sender_store[target["Arn"]]
+                target_unique_id = f"{rule.arn}-{target['Id']}"
+                target_sender = self._target_sender_store[target_unique_id]
                 try:
                     target_sender.process_event(event.copy())
                 except Exception as e:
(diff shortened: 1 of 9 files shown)


tests/aws/services/events/test_events.py
```diff
@@ -1394,6 +1405,338 @@ def test_update_rule_with_targets(
...
+    @markers.aws.validated
+    def test_process_to_multiple_matching_rules_different_targets(
+        self,
+        events_create_event_bus,
+        sqs_create_queue,
+        sqs_get_queue_arn,
+        events_put_rule,
+        aws_client,
+    ):
+        """two rules with each two sqs targets, all 4 ques should receive the event"""
```
bentsku, line 1417:
```suggestion
        """two rules with each two sqs targets, all 4 queues should receive the event"""
```

tests/aws/services/events/test_events.py
```diff
@@ -1626,3 +1969,173 @@ def test_put_target_id_validation(
...
+                {"Id": target_id, "Arn": queue_arn, "InputPath": "$.doesnotexist"},
+            ],
+        )
+
+        snapshot.add_transformers_list(
+            [
+                snapshot.transform.regex(target_id, "target-id"),
+                snapshot.transform.regex(queue_arn, "target-arn"),
+            ]
+        )
```
bentsku, line 2135: nit: I think it'd be better to use `key_value` transformers here in order to have the reference counting, but it also work that way, just less explicit

localstack-core/localstack/services/events/provider.py
```diff
@@ -1835,11 +1835,11 @@ def _delete_rule_services(self, rules: RuleDict | Rule) -> None:
     def _delete_target_sender(self, ids: TargetIdList, rule) -> None:
```
bentsku, line 1835: I can see this is only used when deleting targets. However, from everything I can see in this PR, `Targets` are living inside a `Rule`. Shouldn't we delete all `TargetSender` when deleting a rule? As we're caching the clients inside the sender, it could lead to quite some memory being taken and never released. 

I think that we should delete all the `TargetSender` when deleting a rule. This is somewhat outside of the scope of this PR, but also somewhat in it, because you chose to go the "nested" route in the dictionary key, when in reality, it seems the structure is more `rule_arn` -> `target_id` -> `TargetSender` for the dict (`self._target_sender_store[rule_arn][target_id]`), so it should have another level of nesting, so that when you delete the `rule`, you can do `del self._target_sender_store[rule_arn]` instead of having to iterate over all target and create the "nested key" every time. Both works, you just need to manipulate the key every time.

bentsku (approved):
LGTM! This does fix the issue, and adds quite a big (massive!) amount of tests! It should not break persistence either because this is done manually when restoring. 

I have a nit concerning deleting the target senders when deleting rules and event buses, but this can be tackled in a follow up, as it's not exactly the scope of this PR and was already an issue before. 

Thanks a lot for fixing this! 💯

localstack-core/localstack/services/events/provider.py
maxhoheiser: I will add this in a followup PR - but very nice catch I did not think about this

maxhoheiser: So I like your approach for nesting targets under rules. Regarding deleting them - aws does not allow the deleting of a rule that still has targets - first all targets must be manually deleted - this takes care of the above mentioned problem.

PR #12093, opened 2025-01-02
Docker client: support lists for entrypoints

simonrw:
## Motivation

The Docker SDK supports passing the container entrypoint as a list, but we currently don't. This does not match the Docker API, nor the ECS API.

## Changes

Support either the list form as well as the string form.


dfangl (approved):
Looks good! Ideally we could have a small test :grin:

dfangl (approved):

Issue #11951, opened 2024-11-27
bug: Step Functions incorrect error message in get-execution-history


peter-smith-phd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With the following state machine, I see incorrect output for `get-execution-history`:

```
{
  "StartAt": "pass",
  "States": {
    "pass": {
      "Type": "Pass",
      "Next": "fail",
      "Parameters": {
        "Error": "error-value",
        "Cause": "cause-value"
      }
    },
    "fail": {
      "Type": "Fail",
      "ErrorPath": "$.ErrorX",
      "CausePath": "$.Cause"
    }
  }
}
```

The output of a call to `get-execution-history` has the incorrect entry (near the bottom):

```
"executionFailedEventDetails": {
                "error": "States.Runtime",
                "cause": "RuntimeError(The JSONPath $.ErrorX could not be found in the input {\"Error\": \"error-value\", \"Cause\": \"cause-value\"})"
            }
```

(the extra word `RuntimeError` is causing our test automation to fail.

### Expected Behavior

The output from `get-execution-history` should be:

```
"executionFailedEventDetails": {
                "error": "States.Runtime",
                "cause": "An error occurred while executing the state 'fail' (entered at the event id #4). The JSONPath '$.ErrorX' specified for the field 'ErrorPath' could not be found in the input '{\"Error\":\"error-value\",\"Cause\":\"cause-value\"}'"
            }
```

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose up

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

   awslocal --region us-east-1 stepfunctions create-state-machine --name "Fail_errors" --definition file://file.asl.json --role-arn arn:aws:iam::000000000000:role/Admin
   awslocal --region us-east-1 stepfunctions start-execution --state-machine-arn arn:aws:states:us-east-1:000000000000:stateMachine:Fail_errors
    awslocal --region us-east-1 stepfunctions get-execution-history --execution-arn <arn>

### Environment

```markdown
- OS: Amazon Linux
- LocalStack:
  LocalStack version: 4.0.3.dev22
  LocalStack build date: 2024-11-27
  LocalStack build git hash: 0e01610bb
```

MEPalma:
[...]

PR #12094, opened 2025-01-02
StepFunctions: Better Error Messages For JSONPath Operations

MEPalma:
## Motivation

Errors occurring during JSONPath operations are currently not reported accurately https://github.com/localstack/localstack/issues/11951. Additionally, error messages lack consistency across different JSONPath operations, such as those involving JSON templates and fields like `InputPath`, `OutputPath`, `HeartbeatSecondsPath`, etc.. Assembling these error messages is non-trivial, as it requires including the relevant field name. This update introduces logic to enhance the clarity and consistency of these error messages, as well as adding tests. A limitation of these changes are the execution history produced in the case of errors with `TimeoutSecondsPath` and `HeartbeatSecondsPath`, as these appear to be computed at a state level. However, the error messages are still corrcetly formatted and reported.

## Changes
- jsonpath extraction logic is revised to raise a NoSuchJsonPathError exception which is reported in the absence of refined formatting logics
- the environment now keeps track of the field names being evaluated, which can be sampled by other logics to compose error messages
- EvalComponent objects now update the field they are evaluating in the environment
- Fields with custom error reporting logic implement their custom behaviour
- Default error reports for NoSuchJsonPathError
- minor change to `next_state_name` to ensure the reference to the program state instance being manipulated is not lost between it's validation and the action call

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/component/common/jsonata/jsonata_template_binding.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/component/common/jsonata/jsonata_template_binding.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Final
+from typing import Final, Optional
 
 from localstack.services.stepfunctions.asl.component.common.jsonata.jsonata_template_value import (
     JSONataTemplateValue,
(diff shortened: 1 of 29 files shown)


gregfurman (approved):
LGTM! 

Noticed you didn't mention the `next_state` changes in the PR description. Any chance you can briefly add something on it? Diff look fine but just want to make sure we're aligned 🙂

PR #12097, opened 2025-01-02
fix template body when using CFN create-stack

pinzon:
## Motivation
As detected by a Customer, CloudFormation Get-Template always returns the stack template in JSON format even  when using a YAML template. This  is a mistake which the PR fixes

## Changes
- avoid transforming the the template body to JSON
- add test for the get-template operation after a create-stack call
- rename the previous test of the feature
- Remove old snapsshots.


simonrw (approved):
Thanks for fixing this, I'm surprised though that customers are relying on the format of the template but I suppose they may do their own parsing or something. I'm glad we have better parity here, thanks!

PR #12098, opened 2025-01-03
lambda layer enable deterministic layer version creation

cloutierMat:
## Motivation

As we are making progress with the replicator, we need to implement deterministic id generation for lambda layers in order to control the created version.

## Changes

- When defining the next version for a lambda layer, we will first verify if a static version was set. Defaulting to standard behavior
- added test

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -201,6 +201,7 @@ from localstack.services.lambda_.invocation.runtime_executor import get_runtime_
 from localstack.services.lambda_.lambda_utils import HINT_LOG
 from localstack.services.lambda_.layerfetcher.layer_fetcher import LayerFetcher
 from localstack.services.lambda_.provider_utils import (
+    LambdaLayerVersionIdentifier,
     get_function_version,
     get_function_version_from_arn,
 )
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -3524,8 +3525,10 @@ def publish_layer_version(
 
         layer = state.layers[layer_name]
         with layer.next_version_lock:
-            next_version = layer.next_version
-            layer.next_version += 1
+            next_version = LambdaLayerVersionIdentifier(
+                account_id=account, region=region, layer_name=layer_name
+            ).generate(next_version=layer.next_version)
+            layer.next_version = max(next_version, layer.next_version) + 1
```
joe4dev, line 3531: A comment might be helpful here to clarify the motivation of `max`, in particular to clarify when `next_version` != `layer.next_version`

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -6621,6 +6623,33 @@ def test_layer_policy_lifecycle(
             "get_layer_version_policy_postdeletes2", get_layer_version_policy_postdeletes2
         )
 
+    @markers.aws.only_localstack(reason="Deterministic id generation is LS only")
```
joe4dev, line 6626: A quick explanation why this is needed might be helpful. Our future selfs thank us :)

joe4dev (approved):
LGTM.

Small suggestion: Adding some clarifying comments could help to understand why this is necessary.

Issue #11970, opened 2024-12-02
bug: Event rule target fails to execute when SF with input transformers


ianloubser:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When there is an event rule setup that has a step function configured with target as well as input transformers, it consistently fails to invoke the step function due to a KeyError trying to access `id`.

Example event rule definition (edited to obfuscate actual properties/names):
```
FetchCustomerDataEventRule:
  Type: AWS::Events::Rule
  Properties:
    EventBusName:
      Fn::ImportValue: ${env:STAGE}-customer-event-bus
    EventPattern: '{ "detail": { "event": [ "KEY_FOR_EVENT_TYPE" ] } }'
    Name: TriggerTheStepFunction-${opt:stage}
    RoleArn: ${self:provider.role}
    Targets:
      - Arn:
          Ref: SomeStepFunctionArn${opt:stage}
        RoleArn: ${self:provider.role}
        Id: this-is-a-dummy-target
        InputTransformer:
          InputPathsMap:
            customerId: '$.detail.properties.customerId'
          InputTemplate: '{ "customerId": <customerId> }'
```

Error thrown:
```
{"ErrorCode": "TargetDeliveryFailure", "ErrorMessage": "Failed to deliver to target this-is-a-dummy-target: 'id'"}
```

If adding a dummy value for ID in input template it succeeds, ie:
```
InputTemplate: '{ "id": "dummy_value", "customerId": <customerId> }'
```

### Expected Behavior

Step function execution ID should fallback to a `longuid` when the input is transformed and original event ID is not available. Should not require mandatory fields in the input template for a target.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose up localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
```
const eventBridge = new EventBridge({ region: env.AWS_REGION, credentials: fromEnv(), logger: console })
...
eventBridge.putEvents({
    Entries: [
      {
        EventBusName: `${STAGE}-customer-event-bus`,
        Source: source,
        DetailType: eventType,
        Detail: JSON.stringify(eventData),
      },
    ],
  });
```

### Environment

```markdown
- OS: MacOS 14.7.1 (Apple M3 Chip)
- LocalStack:
  LocalStack version: 4.0.3
[...]

PR #12099, opened 2025-01-03
Bugfix/eventbridge/step functions target id input key

maxhoheiser:
## Motivation
Some target senders rely on specific fields from the event e.g. the event id, if an input transformer or input path is applied to the event, these keys could be removed, and the sending of the event will fail.

This PR fixes the problem addresses in this github bug report: https://github.com/localstack/localstack/issues/11970

## Changes
Optionally supply the input_event (no input transformer or input path applied for transformation) to the respective send_event method of Kinesis sender and Step Functions sender.

## Testing
Add test send event to target kinesis with input transformer enabled
Add test send event to target step functions with input transformer enabled

### DIFF
--- a/localstack-core/localstack/services/events/target.py
+++ b/localstack-core/localstack/services/events/target.py
@@ -168,11 +168,14 @@ class TargetSender(ABC):
         return self._client
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


alexrashed:

Issue #12065, opened 2024-12-23
bug: EventBus description is not assigned


gemyago:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Create EventBus with Description. Describe the created EventBus - description is NOT assigned.

### Expected Behavior

Create EventBus with Description. Describe the created EventBus - the description is ASSIGNED.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    awslocal events create-event-bus --name bus-with-description --description "Example description"
    awslocal events describe-event-bus --name bus-with-description

### Environment

```markdown
- OS: macOS 15.1.1
- LocalStack:
  LocalStack version: 4.0.4.dev63
  LocalStack Docker image sha: sha256:c121cc3b9bd3d39a918b093b238ce8bad68526a3a4d99b5776cb58e9ccb682ed
  LocalStack build date: 2024-12-23
  LocalStack build git hash: 96eec2edd
```

maxhoheiser:
Hi, this PR: https://github.com/localstack/localstack/pull/12100 fixes assigning the optional description to an event bus on create - thank you for bringing this to our attention.

Please let us know if you encounter any additional errors.

PR #12100, opened 2025-01-03
Bugfix/eventbridge/event bus desciption not asigned on create

maxhoheiser:
## Motivation
Event Bridge Buses can have a description, this description is returned on create and on describe event buses.

This PR fixes the issue raised in this github bug report: https://github.com/localstack/localstack/issues/12065

## Changes
Correctly store and handle the description of an event bridge bus on create.

## Testing
Add test for creating and describing event bridge bus with description


localstack-core/localstack/services/events/event_bus.py
```diff
@@ -45,6 +46,7 @@ def create_event_bus_service(
                 region,
                 account_id,
                 event_source_name,
+                description,
```
bentsku, line 49: as a note, I think it would be good to use keyword arguments here, it's much cleaner and readable (named parameters)

localstack-core/localstack/services/events/provider.py
```diff
@@ -963,6 +963,8 @@ def create_event_bus(
         response = CreateEventBusResponse(
             EventBusArn=event_bus_service.arn,
         )
+        if description := event_bus_service.event_bus.description:
+            response["Description"] = description
```
bentsku, line 967: in the case of the data being restored from persisted state, I'm not sure it's fully possible but the `description` field could be non-existent. Until the next minor release, it could be good to use `getattr(event_bus_service.event_bus, "description", None)` to be on the safe side. 

one example of a PR who removed such things: #11809 (and the PR implementing one: https://github.com/localstack/localstack/pull/11358, see this comment: https://github.com/localstack/localstack/pull/11358/files#r1717458334)

bentsku (commented):
Neat fix! This is great, and good thinking with the parametrization of the test!

I have a nit about accessing the model attribute like that, as we can restore persisted state that might not have the field yet. Using `getattr` can avoid weird errors popping up and not break existing users' setup. 

Once this is addressed, I think we merge the PR. Nice fix! 🚀

localstack-core/localstack/services/events/provider.py
maxhoheiser: fixed

localstack-core/localstack/services/events/provider.py
```diff
@@ -1945,6 +1949,8 @@ def _event_bus_to_api_type_event_bus(self, event_bus: EventBus) -> ApiTypeEventB
             "Name": event_bus.name,
             "Arn": event_bus.arn,
         }
+        if event_bus.description:
+            event_bus_api_type["Description"] = event_bus.description
```
bentsku, line 1953: There's still the access here, sorry I didn't put a comment 😅

bentsku (approved):
LGTM! A small comment for the `getattr` usage could be good, with a TODO, to know to revisit and remove once we switch for example major versions, but a quick search will do too! Thanks for addressing the comments!

maxhoheiser:
> LGTM! A small comment for the `getattr` usage could be good, with a TODO, to know to revisit and remove once we switch for example major versions, but a quick search will do too! Thanks for addressing the comments!

I will track it as a todo in the servicepage.

PR #12110, opened 2025-01-07
Ship StepFunction JSONata dependency by default

joe4dev:
## Motivation

A support case raised the issue that StepFunctions failed with an SSL error when downloading the JVM 11 dependency used by the JSONata feature. This blocks usage of the JSONata feature for customers behind a proxy in the default configuration.

Alternatively, we need to provide comprehensive documentation on configuring proxies with all possible LocalStack URLs downloaded dynamically.

## Changes

* Create an installable package plugin `jpype-jsonata` to make the JSONata package installable via lpm
* Remove unnecessary Jackson dependency (side effect from copying the event-ruler installer). Tested locally with the LocalStack dev run script and the SFN test `test_base_map_from_input`
* ATTEMPT 🚧 Change JRE version to unify with dynamodb-local
* Move the package installer to the top-level directory for easy discovery following our convention
* Fix some typos and document JPype limitation
* sneaky: clarify dev run help as a follow-up to https://github.com/localstack/localstack/pull/12092 (cc @simonrw )

## Implications

Shipping `jpype-jsonata` increases the size of the LocalStack image by ~60MB (477MB+60MB => +12%) because of the following dependencies:
* jsonata jar: 166kb
* JRE 11: 60MB. JPype depends on Java and the default JRE 11 is not re-using the same JRE 21 already shipped with dynamodb local (which depends on JRE 17+)

@viren-nadkarni @alexrashed @dominikschubert What do you think? Should we unify the JRE versions or live with the +60MB?
* JPype should work with JRE 21 based on my testing with JRE 23
* We could add `self.java_version = "21"` (🚧 implicit alignment) or change the default version (bigger blast radius)

### DIFF
--- a/localstack-core/localstack/dev/run/__main__.py
+++ b/localstack-core/localstack/dev/run/__main__.py
@@ -36,7 +36,7 @@ from .paths import HOST_PATH_MAPPINGS, HostPaths
     type=str,
     required=False,
     help="Overwrite the container image to be used (defaults to localstack/localstack or "
-    "localstack/localstack-pro.",
+    "localstack/localstack-pro).",
 )
 @click.option(
     "--volume-dir",
@@ -66,7 +66,7 @@ from .paths import HOST_PATH_MAPPINGS, HostPaths
     "--mount-source/--no-mount-source",
     is_flag=True,
     default=True,
-    help="Mount source files from localstack, localstack-ext, and moto into the container.",
+    help="Mount source files from localstack and localstack-ext. Use --local-packages for optional dependencies such as moto.",
 )
 @click.option(
     "--mount-dependencies/--no-mount-dependencies",
--- a/Dockerfile
+++ b/Dockerfile
@@ -153,6 +153,7 @@ RUN --mount=type=cache,target=/root/.cache \
     source .venv/bin/activate && \
     python -m localstack.cli.lpm install \
       lambda-runtime \
+      jpype-jsonata \
       dynamodb-local && \
     chown -R localstack:localstack /usr/lib/localstack && \
     chmod -R 777 /usr/lib/localstack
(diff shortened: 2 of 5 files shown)


localstack-core/localstack/dev/run/__main__.py
```diff
@@ -66,7 +66,7 @@
     "--mount-source/--no-mount-source",
     is_flag=True,
     default=True,
-    help="Mount source files from localstack, localstack-ext, and moto into the container.",
+    help="Mount source files from localstack and localstack-ext. Use --local-packages for optional dependencies such as moto.",
```
simonrw, line 69: Thank you!

viren-nadkarni:
12% increase is image size is substantial. 

>  JPype depends on Java and the default JRE 11 is not re-using the same JRE 21 already shipped with dynamodb local (which depends on JRE 17+)

I'm confused by this line. If JPype works with JRE 21, why is JRE 11 required?

joe4dev:
@viren-nadkarni 

> 12% increase is image size is substantial.

I fully agree and therefore trying out everything to avoid it 🤞 

> If JPype works with JRE 21, why is JRE 11 required?

I'm still experimenting 🚧 My worries are based on our experience with JPype, which only supports one running JVM instance. I'm trying to validate if that only applies to multiple JPype instances (which we knew before and blew up around the v4 release) or also in interaction with other JREs used by packages such as `dynamodb-local`.

joe4dev:
It seems that sharing the JRE 21 with `dynamodb-local` works  🟢  and we don't need to increase the image size significantly 🙌
Just waiting for the ext tests to validate ...

alexrashed:
> @viren-nadkarni @alexrashed @dominikschubert What do you think? Should we unify the JRE versions or live with the +60MB?

@joe4dev I fully agree with @viren-nadkarni that the additional image size is very substantial and should definitely be avoided if possible. I would propose in the short term to upgrade the Java version just for this package. A global upgrade should be handled in a different PR solely focusing on that one upgrade.
If it would speed up the availability of the LPM package, I would even propose to extract the Dockerfile modifications in question from the first iteration of the PR.

@viren-nadkarni What would you propose in the long run? How should the default version be handled? I remember this was a discussion when the Java installer has been introduced (https://github.com/localstack/localstack/pull/11139).

Issue #12054, opened 2024-12-18
bug: Dynamodb service has issue starting (JRE crush on M4 Pro)


srgg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

 awslocal dynamodb  list-tables 

triggers dynamo DB failure: 
 ```
localstack-test  | Ready.
localstack-test  | 2024-12-18T20:58:22.672 DEBUG --- [et.reactor-0] localstack.packages.api    : Installation of dynamodb-local 2 skipped (already installed).
localstack-test  | 2024-12-18T20:58:22.672 DEBUG --- [et.reactor-0] localstack.packages.api    : Performing runtime setup for already installed package.
localstack-test  | 2024-12-18T20:58:22.672 DEBUG --- [et.reactor-0] l.services.dynamodb.server : Starting DynamoDB Local: ['java', '-Xmx256m', '-javaagent:/usr/lib/localstack/dynamodb-local/2/ddb-local-loader-0.1.jar', '-Djava.library.path=/usr/lib/localstack/dynamodb-local/2/DynamoDBLocal_lib', '-jar', '/usr/lib/localstack/dynamodb-local/2/DynamoDBLocal.jar', '-port', '51453', '-dbPath', '/tmp/localstack/state/dynamodb']
localstack-test  | 2024-12-18T20:58:22.672 DEBUG --- [-functhread5] localstack.utils.run       : Executing command: ['java', '-Xmx256m', '-javaagent:/usr/lib/localstack/dynamodb-local/2/ddb-local-loader-0.1.jar', '-Djava.library.path=/usr/lib/localstack/dynamodb-local/2/DynamoDBLocal_lib', '-jar', '/usr/lib/localstack/dynamodb-local/2/DynamoDBLocal.jar', '-port', '51453', '-dbPath', '/tmp/localstack/state/dynamodb']
localstack-test  | 2024-12-18T20:58:22.685 DEBUG --- [-functhread5] l.services.dynamodb.server : #
localstack-test  | 2024-12-18T20:58:22.685 DEBUG --- [-functhread5] l.services.dynamodb.server : # A fatal error has been detected by the Java Runtime Environment:
localstack-test  | 2024-12-18T20:58:22.685 DEBUG --- [-functhread5] l.services.dynamodb.server : #
[...]

PR #12112, opened 2025-01-08
DynamoDB: Disable Scalable Vector Extensions in JRE

viren-nadkarni:
## Background

There have been several reports of JRE crashing with SIGILL on Apple Silicon M4. This crash is specific to JRE 17+ when running in a Docker container (i.e. when [prctl](https://manpages.ubuntu.com/manpages/jammy/en/man2/prctl.2.html) is not available) on aarch64.

In LocalStack, the bug affects DynamoDB Local.

## Changes

This PR adds the suggested workaround of disabling Scalable Vector Extensions (SVE) in JRE.

SVE disable option is passed to the JVM only when supported, otherwise it complains about unsupported VM option and exits.

## Tests

Due to lack of an M4 machine, it is not tested whether this resolves the issue.

## Related

See https://bugs.openjdk.org/browse/JDK-8345296

Might close - https://github.com/localstack/localstack/issues/12054

### DIFF
--- a/localstack-core/localstack/services/dynamodb/server.py
+++ b/localstack-core/localstack/services/dynamodb/server.py
@@ -1,5 +1,7 @@
+import contextlib
 import logging
 import os
+import subprocess
 import threading
 
 from localstack import config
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/dynamodb/server.py
```diff
@@ -142,9 +142,28 @@ def jar_path(self) -> str:
     def library_path(self) -> str:
         return f"{dynamodblocal_package.get_installed_dir()}/DynamoDBLocal_lib"
 
+    def _get_java_vm_options(self) -> list[str]:
+        dynamodblocal_installer = dynamodblocal_package.get_installer()
+
+        # Workaround for JVM SIGILL crash on Apple Silicon M4
+        # See https://bugs.openjdk.org/browse/JDK-8345296
+        # To be removed after Java is bumped to 17.0.15+ and 21.0.7+
```
viren-nadkarni, line 150: Eclipse Temurin haven't released the latest builds yet

https://github.com/adoptium/temurin17-binaries/releases

https://github.com/adoptium/temurin21-binaries/releases

alexrashed (approved):
Thanks a lot, @viren-nadkarni for digging into this! I think this could be a very good lead to fix the issue, even though we unfortunately lack access to the hardware to actually test it on Apple M4.
Definitely worth a try, and we can ask users in https://github.com/localstack/localstack/issues/12054 to verify the blind fix once it's shipped in the latest image.

PR #12113, opened 2025-01-08
fix readme run integration individual tests cmd

alvertogit:
## Motivation

While reading the [Integration tests README](https://github.com/localstack/localstack/blob/master/docs/testing/integration-tests/README.md) file and testing the example command `TEST_PATH="tests/integration/docker/test_docker.py::TestDockerClient" make test` locally I found out that it fails because the path is not correct.

https://github.com/localstack/localstack/blob/master/docs/testing/integration-tests/README.md#running-individual-tests

## Changes

Updated the integration individual tests example command with the corrected path to run the test.

### DIFF
--- a/docs/testing/integration-tests/README.md
+++ b/docs/testing/integration-tests/README.md
@@ -94,7 +94,7 @@ python -m pytest --log-cli-level=INFO tests/integration
 You can further specify the file and test class you want to run in the test path:
 
 ```bash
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Sorry for the late response, and thanks @alvertogit for the proactive contribution to update the docs! The change is looking good! Thanks for making sure that the next potential contributor won't run into this issue! 💯

PR #12114, opened 2025-01-08
Feature/scheduler/extend tests coverage

maxhoheiser:
## Motivation
A user reported a regression for `get_schedule` error code if the schedule was not found. To better understand this and also other discovered regressions and response errors from MOTO (used for implementing event bridge scheduler. And as a preparation for implementing a native localstack event bridge scheduler this PR adds AWS snapshot tests for the CRUD functionality of event bridge scheduler (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/scheduler.html) 

The regressions are fixed in moto is this PR: 

## Changes
Add AWS snapshot tests for CRUD for schedules and schedule groups

### DIFF
--- a/tests/aws/services/scheduler/test_scheduler.py
+++ b/tests/aws/services/scheduler/test_scheduler.py
@@ -58,3 +58,88 @@ def test_untag_resource(aws_client, events_scheduler_create_schedule_group, snap
     assert response["Tags"] == []
 
     snapshot.match("list-untagged-schedule", response)
+
+
+class TestScheduleGroupe:
+    @markers.aws.validated
+    @pytest.mark.parametrize("with_tags", [True, False])
+    @pytest.mark.parametrize("with_client_token", [True, False])
+    def test_create_schedule_group(
+        self, with_tags, with_client_token, events_scheduler_create_schedule_group, snapshot
+    ):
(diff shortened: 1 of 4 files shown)


tests/aws/services/scheduler/test_scheduler.py
```diff
@@ -167,3 +167,378 @@ def tests_create_schedule_with_valid_schedule_expression(
...
+            flexible_time_window["MaximumWindowInMinutes"] = 60
+
+        schedule_expression = "rate(1 minute)"
+        queue_url, queue_arn, role_arn = sqs_as_events_schedule_target()
+        target = {"Arn": queue_arn, "Input": '{"key": "value"}', "RoleArn": role_arn}
+
+        response = events_scheduler_create_schedule(
+            name, flexible_time_window, schedule_expression, target, **kwargs
+        )
+
```
bentsku: nit: I'm not sure we need to snapshot this call, only snapshotting one `create_schedule` call in a non parametrized test should be enough. Here, we are recording 32 calls that are identical

bentsku (commented):
Seems like lots of tests are still failing, might be worth putting the PR back in draft 😄 only had a comment unrelated to the failures that can already be addressed

alexrashed:

PR #12116, opened 2025-01-09
Fix SSM parameter handling with paths and queries by ARN

dfangl:
## Motivation
SSM does not properly return parameters stored when querying it by ARN, if the parameter has a path.

We have some logic to "normalize" the parameter name before passing the call to moto, and this logic is flawed when using an ARN to get the parameter, as the normalized name only contains the last segment of the path.

## Changes
* SSM now properly handles ARNs in its `get_parameter` and `get_parameters` operations, if the parameter contains a path.

### DIFF
--- a/localstack-core/localstack/services/ssm/provider.py
+++ b/localstack-core/localstack/services/ssm/provider.py
@@ -352,7 +352,12 @@ class SsmProvider(SsmApi, ABC):
     @staticmethod
     def _normalize_name(param_name: ParameterName, validate=False) -> ParameterName:
         if is_arn(param_name):
-            return extract_resource_from_arn(param_name).split("/")[-1]
+            resource_name = extract_resource_from_arn(param_name).replace("parameter/", "")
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/ssm/provider.py
```diff
@@ -352,7 +352,12 @@ def _has_secrets(names: ParameterNameList) -> Boolean:
     @staticmethod
     def _normalize_name(param_name: ParameterName, validate=False) -> ParameterName:
         if is_arn(param_name):
-            return extract_resource_from_arn(param_name).split("/")[-1]
+            resource_name = extract_resource_from_arn(param_name).replace("parameter/", "")
+            # if the parameter name is only the root path we want to look up without the leading slash.
+            # Otherwise, we add the leading slash
+            if "/" in resource_name:
```
simonrw, line 358: Might be overkill, but

```suggestion
            if "/" in resource_name and not resource_name.startswith("/"):
```

PR #12117, opened 2025-01-09
StepFunctions, improve the use of ItemsPath declarations in Distributed Map states

MEPalma:
## Motivation
In Step Functions, ItemsPath declarations within Distributed Map state are evaluated only if a JSON input is passed from a previous state in the workflow https://docs.aws.amazon.com/step-functions/latest/dg/input-output-itemspath.html. However, in the current implementation of the SFN v2 interpreter, ItemsPath declarations are always evaluated if declared. This can lead to evaluation failues if the provided path does not exist; AWS instead appears to simply skip the evaluation. These changes evaluate ItemsPath declarations only if this is the only input form for a Distributed Map state, or the Map state is inline. Relevant positive and negative snapshot tests are added.

## Changes
- improve the evaluation logic map state around ItemsPath
- snapshot tests for the scenarios described


localstack-core/localstack/services/stepfunctions/asl/component/state/state_execution/state_map/state_map.py
```diff
@@ -176,8 +179,13 @@ def _eval_execution(self, env: Environment) -> None:
         frame.stack = copy.deepcopy(env.stack)
 
         try:
-            if self.items_path:
-                self.items_path.eval(env=env)
+            # ItemsPath in DistributedMap states is only used if a JSONinput is passed from the previous state.
+            if (
+                not isinstance(self.iteration_component, DistributedIterationComponent)
```
joe4dev, line 184: sidenote: complexity is with instance check is probably something to keep an eye on in the future if this would become a common pattern, but it's good to learn about such special cases first :)

PR #12118, opened 2025-01-09
[ESM] Handle polling of batches exceeding SQS message limits

gregfurman:
## Motivation
This PR allows for SQS requests originating from event source mappings to circumvent message count validations -- allowing for up to `10 000` messages to be returned (or deleted) at a time. 

This is required since polling calls can only be made in batches of 10, making the collection of large batch sizes cumbersome and costly. (i.e 10K messages requires ~1K receive requests + ~1K delete requests).

## Changes
- Add custom `"x-localstack-sqs-override-message-count"` header which the SQS provider uses to override how many messages can be processed in a single call.
- Register event handlers to the AWS client in `SqsPoller`, allowing `ReceiveMessage` and `DeleteMessageBatch` calls to return/delete up to `10 000` records at a time.
- Extend SQS provider to handle custom `"x-localstack-sqs-override-message-count"` headers in boto3 requests.

## TODO

What's left to do:

- [ ] (Optional) Rebase against https://github.com/localstack/localstack/pull/12002 if merged in prior to these changes.

### DIFF
--- a/localstack-core/localstack/constants.py
+++ b/localstack-core/localstack/constants.py
@@ -11,6 +11,8 @@ HEADER_LOCALSTACK_REQUEST_URL = "x-localstack-request-url"
 HEADER_LOCALSTACK_AUTHORIZATION = "x-localstack-authorization"
 HEADER_LOCALSTACK_TARGET = "x-localstack-target"
 HEADER_AMZN_ERROR_TYPE = "X-Amzn-Errortype"
+# HTTP headers used to override internal SQS ReceiveMessage
+HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT = "x-localstack-sqs-override-message-count"
 
 # backend service ports, for services that are behind a proxy (counting down from 4566)
 DEFAULT_PORT_EDGE = 4566
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
@@ -8,6 +8,7 @@ from botocore.client import BaseClient
 from localstack.aws.api.pipes import PipeSourceSqsQueueParameters
 from localstack.aws.api.sqs import MessageSystemAttributeName
 from localstack.config import internal_service_url
+from localstack.constants import HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT
 from localstack.services.lambda_.event_source_mapping.event_processor import (
     EventProcessor,
     PartialBatchFailureError,
@@ -16,6 +17,7 @@ from localstack.services.lambda_.event_source_mapping.pollers.poller import (
     Poller,
     parse_batch_item_failures,
 )
+from localstack.services.lambda_.event_source_mapping.senders.sender_utils import batched
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.strings import first_char_to_lower
 
@@ -36,6 +38,7 @@ class SqsPoller(Poller):
     ):
         super().__init__(source_arn, source_parameters, source_client, processor)
         self.queue_url = get_queue_url(self.source_arn)
+        self._register_client_hooks()
 
     @property
     def sqs_queue_parameters(self) -> PipeSourceSqsQueueParameters:
@@ -45,6 +48,39 @@ class SqsPoller(Poller):
     def is_fifo_queue(self) -> bool:
         # Alternative heuristic: self.queue_url.endswith(".fifo"), but we need the call to get_queue_attributes for IAM
         return self.get_queue_attributes().get("FifoQueue", "false").lower() == "true"
+
+    def _register_client_hooks(self):
+        event_system = self.source_client.meta.events
+
+        def _handle_receive_message_override(params, context, **kwargs):
+            requested_count = params.get("MaxNumberOfMessages")
+            if not requested_count or requested_count <= DEFAULT_MAX_RECEIVE_COUNT:
+                return
+
+            # Allow overide parameter to be greater than default and less than maxiumum batch size.
+            override = min(requested_count, self.sqs_queue_parameters["BatchSize"])
+            context[HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = str(override)
+            params["MaxNumberOfMessages"] = min(requested_count, DEFAULT_MAX_RECEIVE_COUNT)
+
+        def _handle_delete_batch_override(params, context, **kwargs):
+            requested_count = len(params.get("Entries", []))
+            if not requested_count or requested_count <= DEFAULT_MAX_RECEIVE_COUNT:
+                return
+
+            override = min(requested_count, self.sqs_queue_parameters["BatchSize"])
+            context[HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = str(override)
+
+        def _handler_inject_header(params, context, **kwargs):
+            if override := context.pop(HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT, None):
+                params["headers"][HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = override
+
+        event_system.register(
+            "provide-client-params.sqs.ReceiveMessage", _handle_receive_message_override
+        )
+        event_system.register(
+            "provide-client-params.sqs.DeleteMessageBatch", _handle_delete_batch_override
+        )
+        event_system.register("before-call.sqs.*", _handler_inject_header)
 
     def get_queue_attributes(self) -> dict:
         """The API call to sqs:GetQueueAttributes is required for IAM policy streamsing."""
@@ -66,9 +102,9 @@ class SqsPoller(Poller):
         #  https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-sqs.html#pipes-sqs-scaling
         response = self.source_client.receive_message(
             QueueUrl=self.queue_url,
-            MaxNumberOfMessages=min(
-                self.sqs_queue_parameters["BatchSize"], DEFAULT_MAX_RECEIVE_COUNT
-            ),  # BatchSize cannot exceed 10
+            MaxNumberOfMessages=self.sqs_queue_parameters.get(
+                "BatchSize", DEFAULT_MAX_RECEIVE_COUNT
+            ),
             MessageAttributeNames=["All"],
             MessageSystemAttributeNames=[MessageSystemAttributeName.All],
         )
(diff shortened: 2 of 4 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -46,6 +49,39 @@ def is_fifo_queue(self) -> bool:
...
+    def _register_client_hooks(self):
+        event_system = self.source_client.meta.events
+
+        def _handle_receive_message_override(params, context, **kwargs):
+            requested_count = params.get("MaxNumberOfMessages")
+            if not requested_count or requested_count <= DEFAULT_MAX_RECEIVE_COUNT:
+                return
+
+            # Allow overide parameter to be greater than default and less than maxiumum batch size.
+            override = min(requested_count, self.sqs_queue_parameters["BatchSize"])
```
gregfurman, line 61: Useful for getting remaining records less than the batch size. i.e we need 100 records but `BatchSize` is 1k

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -46,6 +49,39 @@ def is_fifo_queue(self) -> bool:
...
+
+        def _handler_inject_header(params, context, **kwargs):
+            if override := context.pop(HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT, None):
+                params["headers"][HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = override
+
+        event_system.register(
+            "provide-client-params.sqs.ReceiveMessage", _handle_receive_message_override
+        )
+        event_system.register(
+            "provide-client-params.sqs.DeleteMessageBatch", _handle_delete_batch_override
```
gregfurman, line 81: Since we delete SQS messages after processing, this allows us to remove up to 10K entries at a time.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
alexrashed: I think nearly all comments like this can be great in-code comments :) If you want to tell a reviewer something, you most often also want to give this to anyone who wants to understand the code :)

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
alexrashed: Same as above: I think this could be a code comment instead of a PR comment :)

alexrashed (commented):
Super interesting topic and nice fix to avoid ESM basically DDOSing SQS 😄 
I posted a few questions, when these are answered, we should be good from my side! 💯 
I think a test would be great, but if that's not possible yet, we could still proceed.

FYI @dfangl @viren-nadkarni: I think this is pretty interesting when looking at #7240. The "internal request parameters" here could not really be used, because its pretty global while for this service it needs to be specific to the request. I feel like this PR could be good input on how to adjust the internal request metadata handling for these use cases. Or maybe there is already a way to use this infra?

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py
```diff
@@ -1560,6 +1560,63 @@ def test_duplicate_event_source_mappings(
             EventSourceArn=event_source_arn,
         )
 
+    @pytest.mark.skip(
+        reason="Flushing based on payload sizes not yet implemented so large payloads are causing issues."
+    )
```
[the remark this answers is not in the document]
gregfurman: To clarify, the blocker here is that we cannot do proper boundary testing since an event with 10K records surpasses the 6MB limit of what a Lambda function can synchronously ingest (see [Invocation Payload Quotas](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html#function-configuration-deployment-and-execution)).

> Is there any way to test this?

Yeah -- we can reduce the size of the batch we are sending. I'll parametrize this test to illustrate the override working and just skip the 10K case.

> When is this skip marker going to be removed?

The skip marker should be removed once https://github.com/localstack/localstack/pull/12002 is merged -- since size based batching will be introduced and payloads _should_ be split into chunks of up to 6MB each (that is, within Lambda quota bounds).

> Adding a test which is skipped by default is not optimal I would say...‏

Agree! Are we OK with multiple test cases where just a single one is (temporarily) skipped?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -36,6 +38,7 @@ def __init__(
     ):
         super().__init__(source_arn, source_parameters, source_client, processor)
         self.queue_url = get_queue_url(self.source_arn)
+        self._register_client_hooks()
```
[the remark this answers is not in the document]
gregfurman: > What is the impact of the registered client hooks

The hooks will add a `x-localstack-sqs-override-message-count` header to all `ReceiveMessage` and `DeleteMessageBatch` calls made via this client. If this client were to be re-used outside of this SQS poller, the expected parameter verification will no longer be used.

> who else uses the client which is handed in here (f.e. due to caching in the client factories)?

No other services should be using this customised source client -- where its exclusivelt created for the poller (outside of the caching mechanism).

> It's a bit dangerous that the client hooks here is registered on any source client handed in here.

OK I have a potential workaround. What about instead of automatically registering these headers, we extend the client to take in a custom parameter that is _only_ ever used in the poller. This parameter can then be converted into a header via a hook.

That way, if the cached-client is re-used, these headers are not added by default but instead require explicit setting. This approach will still require registering hooks but wont add this experimental value to every call.

Example usage:

```python
        response = self.source_client.receive_message(
            QueueUrl=self.queue_url,
            MessageAttributeNames=["All"],
            MessageSystemAttributeNames=[MessageSystemAttributeName.All],
            x_max_number_of_messages=10_000, # where this param will be converted to a header via a hook
        )
```

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -1232,8 +1233,11 @@ def receive_message(
 
         num = max_number_of_messages or 1
```
thrau, line 1234: how about moving the `num` assignment here from the condition block later? this could improve the structure and also allow us to retrieve all messages (with num=-1) using the header

```suggestion
        # override receieve count with value from custom header
        if override := extract_message_count_from_headers(context):
            num = override
        else:
            num = max_number_of_messages or 1
```

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -1232,8 +1233,11 @@ def receive_message(
...
 
-        # backdoor to get all messages
-        if num == -1:
+        # override receieve count with value from custom header
+        if override := extract_message_count_from_headers(context):
+            num = override
+        elif num == -1:
```
thrau, line 1239: ... and revert this for the above change to work
```suggestion
        # backdoor to get all messages
        if num == -1:
```

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -1882,3 +1890,12 @@ def message_filter_message_attributes(message: Message, names: Optional[MessageA
         message["MessageAttributes"] = {k: attributes[k] for k in matched}
     else:
         message.pop("MessageAttributes")
+
+
+def extract_message_count_from_headers(context: RequestContext) -> int:
```
thrau, line 1895: nit
```suggestion
def extract_message_count_from_headers(context: RequestContext) -> int | None:
```

localstack-core/localstack/constants.py
```diff
@@ -11,6 +11,8 @@
 HEADER_LOCALSTACK_AUTHORIZATION = "x-localstack-authorization"
 HEADER_LOCALSTACK_TARGET = "x-localstack-target"
 HEADER_AMZN_ERROR_TYPE = "X-Amzn-Errortype"
+# HTTP headers used to override internal SQS ReceiveMessage
+HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT = "x-localstack-sqs-override-message-count"
```
thrau, line 15: i think it would be better to move this into the SQS service module. having service-specific constants in the global constants file may clutter it over time.

thrau (approved):
nice changes, @gregfurman! the new parameters seems very cleanly added at the right levels of abstraction.

just added a few minor suggestions

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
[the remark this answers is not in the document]
gregfurman: @bentsku Do you think we should add the header as a param (like the example above) or continue with the current approach?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -46,6 +49,39 @@ def is_fifo_queue(self) -> bool:
...
+
+        def _handle_receive_message_override(params, context, **kwargs):
+            requested_count = params.get("MaxNumberOfMessages")
+            if not requested_count or requested_count <= DEFAULT_MAX_RECEIVE_COUNT:
+                return
+
+            # Allow overide parameter to be greater than default and less than maxiumum batch size.
+            override = min(requested_count, self.sqs_queue_parameters["BatchSize"])
+            context[HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = str(override)
+            params["MaxNumberOfMessages"] = min(requested_count, DEFAULT_MAX_RECEIVE_COUNT)
```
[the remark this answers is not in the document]
gregfurman: This is necessary since we want to change the `MaxNumberOfMessages` param back to a value below `10` -- otherwise validations fail.

joe4dev (commented):
Nothing blocking a merge from my side (once CI is 🟢  again). Great work 👏👏👏 
I'm just sharing my opinion about ideas around standardizing hook-based customization implementations (in the future).

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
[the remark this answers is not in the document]
gregfurman: Agree. Let's get this in and iterate on it

localstack-core/localstack/services/sqs/provider.py
gregfurman: The current logic allow us to only validate the parameter if no override (or -1) is provided.

I agree we should look into using the override header for this `-1` case but will rather do it in a follow-up PR.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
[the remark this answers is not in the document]
bentsku: Agreed as well, I like the parameter approach, as it also allows us to have a clearer way to add parameters, and be more explicit about it, I agree with Joel. But this works for now, and we can iterate on it. Sorry for the delay in reviewing! Thanks Joel 🙏

PR #12121, opened 2025-01-10
Add tests for KMS resource tagging and untagging

k-a-il:
## Motivation
The AWS-validated snapshot KMS tests for tagging and untagging keys are currently missing. We need to add these tests, collect their responses, and verify them against LocalStack.

## Changes

- AWS-validated tests for tagging and untagging keys have been added for the KMS service.

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -21,8 +21,8 @@ from localstack.utils.crypto import encrypt
 from localstack.utils.strings import short_uid, to_str
 
 
-def create_tags(*key_value_args):
-    return [{"TagKey": key, "TagValue": value} for key, value in key_value_args]
+def create_tags(**kwargs):
+    return [{"TagKey": key, "TagValue": value} for key, value in kwargs.items()]
 
 
 @pytest.fixture(scope="class")
@@ -116,7 +116,7 @@ class TestKMS:
             region_name=region_name, Description="test key 123", KeyUsage="ENCRYPT_DECRYPT"
         )["KeyId"]
 
-        tags = create_tags(("tag1", "value1"), ("tag2", "value2"))
+        tags = create_tags(tag1="value1", tag2="value2")
         kms_client.tag_resource(KeyId=key_id, Tags=tags)
 
         response = kms_client.list_resource_tags(KeyId=key_id)["Tags"]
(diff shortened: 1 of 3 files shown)


k-a-il:

tests/aws/services/kms/test_kms.py
```diff
@@ -21,6 +21,10 @@
 from localstack.utils.strings import short_uid, to_str
 
 
+def create_tags(**kwargs):
+    return [{"TagKey": key, "TagValue": value} for key, value in kwargs.items()]
```
bentsku, line 25: neat way of using Python named arguments de-structuring to create tags 👌

tests/aws/services/kms/test_kms.py
```diff
@@ -987,47 +1058,6 @@ def test_get_put_list_key_policies(self, kms_create_key, aws_client, account_id)
...
-
-        old_tag_one = _create_tag("one")
-        new_tag_one = _create_tag("one")
-        tag_two = _create_tag("two")
-        tag_three = _create_tag("three")
-
-        key_id = kms_create_key(Tags=[old_tag_one, tag_two])["KeyId"]
-        assert _are_tags_there([old_tag_one, tag_two], key_id) is True
-        # Going to rewrite one of the tags and then add a new one.
-        aws_client.kms.tag_resource(KeyId=key_id, Tags=[new_tag_one, tag_three])
```
bentsku, line 1024: nit: I think we're missing that case in the new tests: when the key already has tags, and calling `tag_resource` again on it and overwrite or adding new tags. I think we could test this scenario either by creating a key with existing tags, or calling `tag_resource` twice with different arguments. What do you think? I admit the existing test is really not clear, so it's easy to miss one check!

bentsku (commented):
Really nice! This looks really great, and I like the way the tests are structured 👌 well done! 

I think we're still missing one case from the deleted test (which is very not clear by the way, thanks for updating that one!). Once the comment is addressed, I believe we're good to merge! Can't wait to see the follow up 🚀 Congrats!

tests/aws/services/kms/test_kms.py
k-a-il: True, I'm not sure why I overlooked this part. Testing the overwriting aspect is especially important.

k-a-il: I’ve added a test that modifies an existing tag and adds additional tags to the key that’s already tagged.

bentsku (approved):
LGTM! Thanks for addressing the comment! Welcome to LocalStack's contributors 🥳 
The test failure is unrelated to the changes (it is in S3), so this is safe to merge.

PR #12122, opened 2025-01-10
Docker Utils: Utility method to create file in container

bblommers:
## Motivation
Adds a utility method to create a file in the docker container, by just passing the content of the file. I find myself reusing the same approach quite often, so IMO it would be much nicer to have it available as a utility.

#### Permissions
When running this inside a container, the file will be created by the `root` user. If the target container runs it's process as a custom (non-root) user, it may not have access to files created by `root`. 

The additional `chmod_mode` argument can be used to ensure the file is available to the custom user. (We could presumably also `chown` the file to the same effect, but I only used `chmod` so far to make files available in this scenario.)


dfangl (approved):
LGTM!

PR #12131, opened 2025-01-13
Add analytics on exception

simonrw:
## Motivation

When a user encounters an exception, we do not know about it. We can add anlytics to the error handling to capture this event, and gain more information about the failure.

## Changes

It would be nice to capture the failure traceback, however this would be a lot of data to send to our analytics backend. For now we just show the stringified exception.

* Rename some test files so they are included in the unit test suite
* Add an analytics logging event on exception


simonrw:
We basically already have this with the status code reporting

PR #12137, opened 2025-01-14
APIGW: fix OpenAPI import StatusCode string casting

bentsku:
## Motivation
We got a report that importing a specific OpenAPI file in API Gateway v1 would result in a type error in OpenTOFU:
> deserialization failed, failed to decode response body with invalid JSON, expected StatusCode to be of type string, got json.Number instead

This is because we were not casting as string the status code for the Method status code, and the IntegrationResponse.

When importing an API, we do so by creating the resources "in memory" directly and not via API calls, as this was how on the implementation was. This particular piece of code needs to be reworked fully, but this is out of scope of this PR.

## Changes
- add a test reproducing a similar exception
- cast imported status codes to string

### DIFF
--- a/tests/aws/services/apigateway/test_apigateway_import.snapshot.json
+++ b/tests/aws/services/apigateway/test_apigateway_import.snapshot.json
@@ -5192,7 +5192,7 @@
     }
   },
   "tests/aws/services/apigateway/test_apigateway_import.py::TestApiGatewayImportRestApi::test_import_with_integer_http_status_code": {
-    "recorded-date": "14-01-2025, 14:03:07",
+    "recorded-date": "14-01-2025, 14:09:57",
     "recorded-content": {
       "resources": {
         "items": [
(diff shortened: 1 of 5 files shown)


tests/aws/services/apigateway/test_apigateway_import.snapshot.json
```diff
@@ -5190,5 +5190,124 @@
...
+        "ResponseMetadata": {
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 200
+        }
+      },
+      "integration-response-test-get": {
+        "responseParameters": {
+          "method.response.header.Access-Control-Allow-Origin": "'*'"
+        },
+        "statusCode": "200",
```
bentsku, line 5305: before the fix, this would have been an `int` and fail with some typed languages trying to decode the response

simonrw (approved):
Thanks for improving the parity with this change, and yes types and yaml are not good friends!

PR #12154, opened 2025-01-21
EventBridge: fix empty conditions in event rule engine

bentsku:
## Motivation
Follow up from #12147, I've added more test cases related to the original issue raised by the user, when the condition is actually an empty string. As we're assigning using the walrus operator, we're testing for truthiness, and non-truthy values like `""` or `0` would not get matched as a proper condition and thus not be evaluated. 

I'd like to thank @groves again for the original PR and opening the issue. 

## Changes
- add new test cases covering the ones I could think of, nested prefix/suffix and other string matching, as well as the `0` value
- change the if condition + walrus to not check for truthiness but existence


localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -109,15 +109,15 @@ def _evaluate_condition(self, value, condition, field_exists: bool):
             else:
                 return self._evaluate_suffix(suffix, value)
 
-        elif equal_ignore_case := condition.get("equals-ignore-case"):
+        elif (equal_ignore_case := condition.get("equals-ignore-case")) is not None:
             return self._evaluate_equal_ignore_case(equal_ignore_case, value)
         elif numeric_condition := condition.get("numeric"):
```
joe4dev, line 114: The isinstance check should handle the numeric case, hence we don't need the None check here

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -109,15 +109,15 @@ def _evaluate_condition(self, value, condition, field_exists: bool):
...
             return self._evaluate_numeric_condition(numeric_condition, value)
 
         elif cidr := condition.get("cidr"):
```
joe4dev, line 117: I guess we'd get a value error, hence we don't need the None check here

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -97,7 +97,7 @@ def _evaluate_condition(self, value, condition, field_exists: bool):
             return False
         elif (prefix := condition.get("prefix")) is not None:
```
joe4dev, line 98: Here we had it 😃

localstack-core/localstack/services/events/event_rule_engine.py
```diff
@@ -147,7 +147,7 @@ def _evaluate_wildcard(condition: str, value: str) -> bool:
         return re.match(re.escape(condition).replace("\\*", ".+") + "$", value)
 
     @staticmethod
-    def _evaluate_numeric_condition(conditions, value) -> bool:
+    def _evaluate_numeric_condition(conditions: list, value) -> bool:
```
joe4dev, line 150: intentionally skipping the typings for `value` here? Is it `Any` because we check defensively with isinstance?

joe4dev (approved):
Great catch @bentsku 

Thank you for thoroughly following up these edge cases. What an impressive test suite we got here 😮 💯

localstack-core/localstack/services/events/event_rule_engine.py
bentsku: Yes, I was a bit conflicted about putting type hint and defensively checking, I could add `t.Any`, I think it makes a bit more sense than `int | float` which a check right after?

localstack-core/localstack/services/events/event_rule_engine.py
bentsku: added a comment 👍

localstack-core/localstack/services/events/event_rule_engine.py
bentsku: added a comment 👍

bentsku (commented):
Thanks a lot for the review @joe4dev! 🙏 I've added more comments and a small validation case. Thanks a lot for being thorough 🙇

localstack-core/localstack/services/events/event_rule_engine.py
joe4dev: yes 👍 

`int | float` would only cover the happy case, but the method is designed to handle `Any` (i.e., user input)

PR #12158, opened 2025-01-22
Remove deprecated Java-based event ruler

joe4dev:
## Motivation

With the new native event rule engine introduced in 4.0.3, we deprecated the Java-based event rule engine to mitigate runtime conflicts if used simultaneously with the StepFunctions JSONata feature.

## Changes

* Remove the feature flag `EVENT_RULE_ENGINE`
* Remove the Java-based event ruler implementation and all dependencies (i.e., LPM installer, plugin)

## Implications

* ~~Apart from the failing test `test_start_list_describe_canceled_replay` with the Python-based event rule engine, we are unaware of other regressions.~~ 👉  Turns out to be an unrelated provider bug, failing with a TypeError even with the Java engine. Clarified the skip message accordingly.
* Given the complexity of the event rule engine, we cannot rule out unknown regressions in edge cases (e.g., as recently fixed in https://github.com/localstack/localstack/pull/12154), but @bentsku was usually super quick to fix these quickly 🙇


bentsku:
I've tried running `test_start_list_describe_canceled_replay` and it seems the exception might not be due to the event rule engine, at least it's not the first exception:

```python
E             File "/Users/benjaminsimon/Projects/localstack/localstack/localstack-core/localstack/services/events/provider.py", line 1183, in start_replay
E               events_to_replay = archive_service.get_events(
E                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E             File "/Users/benjaminsimon/Projects/localstack/localstack/localstack-core/localstack/services/events/archive.py", line 133, in get_events
E               events_to_replay = self._filter_events_start_end_time(start_time, end_time)
E                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E             File "/Users/benjaminsimon/Projects/localstack/localstack/localstack-core/localstack/services/events/archive.py", line 182, in _filter_events_start_end_time
E               return [
E                      ^
E             File "/Users/benjaminsimon/Projects/localstack/localstack/localstack-core/localstack/services/events/archive.py", line 185, in <listcomp>
E               if event_start_time <= self._normalize_datetime(event["time"]) <= event_end_time
E                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E             File "/Users/benjaminsimon/Projects/localstack/localstack/localstack-core/localstack/services/events/archive.py", line 174, in _normalize_datetime
E               return dt.replace(second=0, microsecond=0)
E                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           TypeError: str.replace() takes no keyword arguments
```

Might be worth investigating somewhere else than here, but I think we might be good now with no tests failing due to the event rule engine?

joe4dev:
> Might be worth investigating somewhere else than here, but I think we might be good now with no tests failing due to the event rule engine?

Yes, I came to the same conclusion. Even using the Java engine, `test_start_list_describe_canceled_replay` fails with this `TypeError` provider bug. I clarified the skip reason accordingly.

bentsku (approved):
LGTM! Thanks for taking care of the cleanup 🧹 this looks great, and hopefully users will report issues that we can fix quickly if we do not already.

PR #12160, opened 2025-01-22
Fix legacy test that broke pipeline after eventbridge cron update

anisaoshafi:
## Motivation
After merging this [PR](https://github.com/localstack/localstack/pull/12156) on eventbridge schedule validation, this integration test broke the pipeline: `test_event_rule_creation_without_target`.
That PR actually brought more parity with AWS and broke an existing test that shouldn't have passed.
Now the test is aws_validated 🚀 

## Changes

- Fixed a cron schedule: in `events_rule_without_targets`.yaml. ` 'cron(0 1 * * * *)' `-->  `'cron(0 1 * * ? *)'`
- Test snapshot tested and aws validated.

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_events.py
+++ b/tests/aws/services/cloudformation/resources/test_events.py
@@ -154,10 +154,11 @@ def test_event_rule_to_logs(deploy_cfn_template, aws_client):
     assert message_token in log_events["events"][0]["message"]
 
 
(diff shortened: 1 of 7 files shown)


bentsku (approved):
LGTM! Thanks for jumping on this 🦸‍♀️ great fix, nice seeing another test being validated and now understanding why it failed before. The comment on the invalid CRON is also very useful because it was hard to understand why it would fail!

Also kudos for adding the snapshot test 🚀 we can have even more confidence in the test now!

joe4dev (approved):
Awesome work @anisaoshafi and super fast fix 👏👏👏
Feel free to add a conditional skip for the EventBridge v1 provider

```
    @pytest.mark.skipif(
        is_old_provider(),
        reason="V1 provider does not support this feature",
    )
```

Issue #12139, opened 2025-01-14
bug: Step function -> "Get S3 Object" Action's error handling not matching real AWS behaviour


lagarwal-uic:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I am using the attached state machine (simplified it for this issue reporting). It relies on a bucket named `test-bucket` (needs to be created to run the state machine).

When it is executed, I expected localstack to return exactly same response as AWS, particularly the `cause` field. However it is different as you can see below.

### AWS Response:

> aws stepfunctions describe-execution --execution-arn "arn:aws:states:eu-west-1:038610054328:execution:Leena-failing-s3:2a7e79d7-892d-4f3a-9669-8a5648a07750"
> {
>     "executionArn": "arn:aws:states:eu-west-1:038610054328:execution:Leena-failing-s3:2a7e79d7-892d-4f3a-9669-8a5648a07750",
>     "stateMachineArn": "arn:aws:states:eu-west-1:038610054328:stateMachine:Leena-failing-s3",
>     "name": "2a7e79d7-892d-4f3a-9669-8a5648a07750",
>     "status": "FAILED",
>     "startDate": "2025-01-14T19:30:54.206000+00:00",
>     "stopDate": "2025-01-14T19:30:54.442000+00:00",
>     "input": "{}",
>     "inputDetails": {
>         "included": true
>     },
>     "error": "S3.NoSuchKeyException",
>     "cause": "{\"errorMessage\":\"object not found\",\"statusCode\":404}",
>     "redriveCount": 0,
>     "redriveStatus": "REDRIVABLE"
> }

[failing-s3-step-function.json](https://github.com/user-attachments/files/18415609/failing-s3-step-function.json)

### Localstack Response:

> awslocal stepfunctions describe-execution --execution-arn "arn:aws:states:eu-west-1:000000000000:execution:TestSM:dc8b6ce9-09cb-4190-8ae3-49d60a45e0ec"
> {
>     "executionArn": "arn:aws:states:eu-west-1:000000000000:execution:TestSM:dc8b6ce9-09cb-4190-8ae3-49d60a45e0ec",
>     "stateMachineArn": "arn:aws:states:eu-west-1:000000000000:stateMachine:TestSM",
>     "name": "dc8b6ce9-09cb-4190-8ae3-49d60a45e0ec",
>     "status": "FAILED",
>     "startDate": 1736884055.929379,
>     "stopDate": 1736884056.021036,
[...]

PR #12161, opened 2025-01-22
StepFunctions: AWS-SDK Integration Retain Specific Error Names

MEPalma:
## Motivation
This PR addresses an issue in the Step Functions AWS-SDK integration where, upon integration failure, the error name is incorrectly cast to a fixed StatesError. This behavior prevents workflows from catching specific error types and implementing retry logic tailored to particular failures in the AWS-SDK integration: https://github.com/localstack/localstack/issues/12139

## Changes
- aws-sdk error decoration logic uses the computed exception name
- added relevant snapshot test


joe4dev (approved):
Nice catch and clear test case 👏

PR #12163, opened 2025-01-22
fix cfn template join with noValue

pinzon:
## Motivation
This PR fixes an issue encountered by a customer when using AWS::NoValue within a Fn::Join in CFn template
 
## Changes
Make  the template deployer ignore the AWS::NoValue when calculaing theh Fn:join result

## Test
- new aws validated test.


silv-io (approved):
LGTM!

PR #12164, opened 2025-01-22
fix pre-signed URL with JS SDK, enable test_presigned_url_v4_x_amz_in_qs

bentsku:
## Motivation
We had pipeline failures due to the `test_presigned_url_v4_x_amz_in_qs` test suddenly failing. This happened at the same time we've had a big update in S3 data integrity and checksums in our boto update.

I suspected a change in behavior in the Javascript SDK: what happens now is that the JS SDK is adding a checksum value for pre-signed URL, which in itself is quite a bad idea, as you cannot know in advance what the value of the object can be. 

So it seems AWS is just flat out ignoring the query string parameter. 

Pre-signed URLs are a funny concept: you pass some values that would normally be headers in a regular signed request as query string parameters, so that the URL can be shared. To not have to manually parse query string parameters in our S3 provider, our pre-signed URL handler is picking up the query string parameters and putting them back as headers and mutating the request before it being parsed, so that we can fully enjoy the ASF power. See #8918

But in that case, the query string parameter is important for the signature, if you try modifying it, S3 will raise an invalid signature exception. But you need to ignore it before passing it down to the provider. 

## Changes
- ignore any `x-amz-checksum-<>` header that is not in the signed headers, and do not pass it down the provider
- re-enable the test and add a few checks

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -7560,6 +7560,16 @@ class TestS3PresignedUrl:
         )
         assert response.status_code == 200
 
+        # assert that the checksum-crc-32 value is still validated and important for the signature
+        bad_presigned_url = presigned_url.replace("crc32=AAAAAA%3D%3D", "crc32=BBBBBB%3D%3D")
+        response = requests.put(
+            bad_presigned_url,
+            data=b"123456",
+            verify=False,
+            headers={"Content-MD5": "4QrcOUm6Wau+VuBX8g+IPg=="},
+        )
+        assert response.status_code == 403
+
         # verify that we properly saved the data
         head_object = aws_client.s3.head_object(
             Bucket=function_name, Key=object_key, ChecksumMode="ENABLED"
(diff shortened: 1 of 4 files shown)


tests/aws/services/s3/test_s3.py
```diff
@@ -7544,6 +7543,8 @@ def test_presigned_url_v4_x_amz_in_qs(
         # assert that the Javascript SDK hoists it in the URL, unlike Boto
         assert StorageClass.STANDARD in presigned_url
         assert "bar-complicated-no-random" in presigned_url
+        # the JS SDK also adds a default checksum now even for pre-signed URLs
```
silv-io, line 7546: Default checksum wow

silv-io (approved):
LGTM! Very weird in total

tests/aws/services/s3/test_s3.py
bentsku: Yeah I still don't understand how this made it into prod 😄

PR #12165, opened 2025-01-22
S3: fix exception from ObjectLock retention

bentsku:
## Motivation
Follow up from regenerating snapshots in #12145, one date that was set in the future actually was now in the past 😅 and triggered a new kind of exception. It had been skipped. This PR now implements this exception.

Also removed a few now useless snapshot markers that have been fixed in the meantime but not removed. 

Note: weird thing but S3 returns the error in the PST timezone, interesting 🤔 

## Changes
- fix the exception raised in the date was invalid
- remove a few now useless snapshot markers

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -9,6 +9,7 @@ from io import BytesIO
 from operator import itemgetter
 from typing import IO, Optional, Union
 from urllib import parse as urlparse
+from zoneinfo import ZoneInfo
 
 from localstack import config
 from localstack.aws.api import CommonServiceException, RequestContext, handler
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/s3/provider.py
```diff
@@ -2011,7 +2012,7 @@ def restore_object(
             return RestoreObjectOutput()
 
         restore_expiration_date = add_expiration_days_to_datetime(
-            datetime.datetime.utcnow(), restore_days
+            datetime.datetime.now(datetime.UTC), restore_days
```
bentsku, line 2015: `datetime.datetime.utcnow()` is now deprecated and should be replaced by `datetime.datetime.now(datetime.UTC)`

tests/aws/services/s3/test_s3.py
```diff
@@ -9409,10 +9377,20 @@ def test_s3_object_retention_exc(self, aws_client, s3_create_bucket, snapshot):
                 Bucket=s3_bucket_locked,
                 Key=object_key,
                 VersionId=version_id,
-                Retention={"Mode": "GOVERNANCE", "RetainUntilDate": datetime.datetime(2025, 1, 1)},
+                Retention={"Mode": "GOVERNANCE", "RetainUntilDate": datetime.datetime(2029, 1, 1)},
```
k-a-il, line 9380: nit: did you consider using a variable datetime based on the current date, like now() + 1 year?

k-a-il (approved):
Nice 👍

k-a-il:
> Note: weird thing but S3 returns the error in the PST timezone, interesting 🤔

yes, indeed, interesting decision from their side

PR #12166, opened 2025-01-22
S3: fix trailing slash logic

bentsku:
## Motivation
We've got a reported in our Community Slack channel about issue with trailing slashes, especially with short key name (single characters). The trailing slashes would be missing:
```bash
awslocal s3api put-object --bucket sample-bucket2 --key "u/" --content-type "application/x-directory"
awslocal s3api list-objects-v2 --bucket sample-bucket2
...
       {
           "Key": "u",
           "LastModified": "2025-01-22T11:46:17.000Z",
           "Size": 0,
           "StorageClass": "STANDARD"
       },
...
```

I wrote a test for it and I had a hunch it came from our custom logic to parse URI parameters. We had custom logic for S3 because our router strips trailing slashes by default. We would check if the key name is in the path, and then split on the key and check if the leftover characters in the path are all slashes, to keep them. 

However, to do so, we would partition the path on the parse `Key` parameter. Printing the result showed quite quickly what was the issue: 
```python
uri_params={'Bucket': 'test-bucket-ec3db341', 'Key': 'u'}
request.path.partition(uri_params["Key"])=('/test-b', 'u', 'cket-ec3db341/u/')
```

We would split/partition on the bucket name first while using path style. To avoid this, we need to start from the right side. That way, we properly get all the right side slashes.

## Changes
- add new tests for single characters with trailing slash
- partition the trailing slashes the path by the `Key`, but starting from the right

### DIFF
--- a/tests/aws/services/s3/test_s3.py
+++ b/tests/aws/services/s3/test_s3.py
@@ -532,6 +532,23 @@ class TestS3:
         snapshot.match("del-object-special-char", resp)
 
     @markers.aws.validated
+    def test_put_get_object_single_character_trailing_slash(self, s3_bucket, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.key_value("Name"))
+        single_chars = [
+            "a/",
+            "t/",
+            "u/",
+        ]
+        for char in single_chars:
+            resp = aws_client.s3.put_object(Bucket=s3_bucket, Key=char, Body=b"test")
+            snapshot.match(f"put-object-single-char-{char}", resp)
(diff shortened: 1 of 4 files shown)


tests/aws/services/s3/test_s3.py
alexrashed, line 1: **praise:** Nice and clean test! We are getting more and more edge cases covered by the S3 tests! 💯 ‏

alexrashed (approved):
Whoof, nice catch! I guess this was pretty tricky to pin down! Nice, clean, and simple fix, coming with a nice clean and simple AWS validated snapshot test! 💯

Issue #9782, opened 2023-11-30
Content encoding incorrect after recent AWS SDK for Java S3 changed to use sra auth


youngm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

No matter what value you specify for content encoding local stack stores a value of `aws-chunked`. Since AWS SDK for java 2.21.16. I believe the PR that broke this is https://github.com/aws/aws-sdk-java-v2/pull/4653.

Everything works fine when invoked against real S3 https://github.com/aws/aws-sdk-java-v2/issues/4725

### Expected Behavior

The content encoding specified should be the one stored with the S3 Object

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

I'm using TestContainers to launch localstack.
You need AWS SDK for Java v2 2.21.16 or newer.

```
    var putRequest =
        PutObjectRequest.builder()
            .bucket(TEST_BUCKET)
            .key("file")
            .contentEncoding("bogus")
            .build();

    s3Client.putObject(putRequest, RequestBody.empty());
    var response =
        s3Client.getObject(GetObjectRequest.builder().bucket(TEST_BUCKET).key("file").build());
    assertThat(response.response().contentEncoding()).isEqualTo("bogus");
```
This test fails with localstack but passes against real S3.

### Environment

```markdown
- OS: Windows
- LocalStack: 3.0.2
```

bentsku:
Hello @youngm and thanks for your report.

I'm currently unable to reproduce with the snippet you shared and the AWS Java SDK V2 version 2.21.24 and 2.21.34.
I'm not sure what is the difference, but maybe I am not creating the client in the same way and it does not trigger the `aws-chunked` content encoding. 

Could you share a full reproducible sample with maybe the `pom.xml` or something complete so that I could trigger the issue on my hand and fix it quickly?

Thank you!

bentsku:
I'm starting to believe that if you set a `ContentEncoding` parameter against the real S3, the SDK will not trigger the `aws-chunked` encoding of the request to not override the set `ContentEncoding` (as it is normally set in the `ContentEncoding` header). It might be a regression after all from the SDK, but it's not triggering for me.

youngm:
Thanks for digging into this @bentsku. I've done some more testing and found the failure only happens when using `url-connection-client` the default httpclient works fine. So, here are the facts as I see it:
* AWS S3 and Localstack both worked with httpclient and url-connection-client prior to 2.21.16
[...]

PR #12167, opened 2025-01-22
S3: fix ContentEncoding with aws-chunked

bentsku:
## Motivation
Following this discussion here: https://github.com/localstack/localstack/issues/9782#issuecomment-2595044611

We mishandled how the `aws-chunked` `Content-Enconding` header was saved on the object. If given, S3 should remove the header, [this was confirmed](https://github.com/aws/aws-sdk-java-v2/issues/5769#issuecomment-2594242699) by people at Amazon working on the SDK themselves.

## Changes
- remove the `aws-chunked` value in the object `ContentEncoding`

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -761,6 +761,11 @@ class S3Provider(S3Api, ServiceLifecycleHook):
             decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
             body = AwsChunkedDecoder(body, decoded_content_length, s3_object=s3_object)
 
+            if content_encoding := s3_object.system_metadata.pop("ContentEncoding", None):
+                encodings = [enc for enc in content_encoding.split(",") if enc != "aws-chunked"]
+                if encodings:
+                    s3_object.system_metadata["ContentEncoding"] = ",".join(encodings)
+
         with self._storage_backend.open(bucket_name, s3_object, mode="w") as s3_stored_object:
             # as we are inside the lock here, if multiple concurrent requests happen for the same object, it's the first
             # one to finish to succeed, and subsequent will raise exceptions. Once the first write finishes, we're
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/s3/provider.py
```diff
@@ -761,6 +761,12 @@ def put_object(
             decoded_content_length = int(headers.get("x-amz-decoded-content-length", 0))
             body = AwsChunkedDecoder(body, decoded_content_length, s3_object=s3_object)
 
+            # S3 removes the `aws-chunked` value from ContentEncoding
+            if content_encoding := s3_object.system_metadata.pop("ContentEncoding", None):
+                encodings = [enc for enc in content_encoding.split(",") if enc != "aws-chunked"]
+                if encodings:
+                    s3_object.system_metadata["ContentEncoding"] = ",".join(encodings)
```
alexrashed, line 768: **question (non-blocking):** ‏Content encoding is a very typical case for a middle ware (or gateway in terms of hops in HTTP traffic). A good example would be our GZIP decoding handler (I know, not for S3, but in general 😛).
Do you think a structure like this could also work here without adding too much overhead?

alexrashed (approved):
Awesome! Thanks for jumping on this! The solution is simple and the test is great! 💯 
I only added a comment / question with some food for thought.

localstack-core/localstack/services/s3/provider.py
bentsku: Good question! I think in our case, it makes it a bit tricky because we need to update some object values (checksum s and such) after reading the stream (because of trailing headers). We could probably rework it, I'm still not sure what to make of trailing headers and how to properly expose them once we finish reading the stream.

As we're only dealing with `aws-chunked` in `PutObject` and `UploadPart` (in slightly different ways), I think we're good for now to keep it here. If we were to see we need it more globally, then I'd be all for it. 

See the PR implementing it #8721.

PR #12172, opened 2025-01-23
[SQS] Add custom header for overriding WaitTimeSeconds

gregfurman:
## Motivation
This PR allows the `WaitTimeSeconds` parameter to be overridden via a custom header. Useful for increasing the amount of time an [SQS long polling request ](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html#sqs-long-polling) can collect for prior to returning.

## Changes
- New constant `HEADER_LOCALSTACK_SQS_OVERRIDE_WAIT_TIME_SECONDS` with `x-localstack-sqs-override-wait-time-seconds`.
- Add support for replacing value of `WaitTimeSeconds` parameter with value of `x-localstack-sqs-override-wait-time-seconds` header.

### DIFF
--- a/localstack-core/localstack/services/sqs/provider.py
+++ b/localstack-core/localstack/services/sqs/provider.py
@@ -79,6 +79,7 @@ from localstack.services.plugins import ServiceLifecycleHook
 from localstack.services.sqs import constants as sqs_constants
 from localstack.services.sqs.constants import (
     HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT,
+    HEADER_LOCALSTACK_SQS_OVERRIDE_WAIT_TIME_SECONDS,
 )
 from localstack.services.sqs.exceptions import InvalidParameterValueException
 from localstack.services.sqs.models import (
@@ -1234,9 +1235,17 @@ class SqsProvider(SqsApi, ServiceLifecycleHook):
         # TODO add support for message_system_attribute_names
         queue = self._resolve_queue(context, queue_url=queue_url)
 
-        if wait_time_seconds is None:
+        poll_empty_queue = False
+        if override := extract_wait_time_seconds_from_headers(context):
+            wait_time_seconds = override
+            poll_empty_queue = True
+        elif wait_time_seconds is None:
             wait_time_seconds = queue.wait_time_seconds
-
+        elif wait_time_seconds < 1 or wait_time_seconds > 20:
+            raise InvalidParameterValueException(
+                f"Value {wait_time_seconds} for parameter WaitTimeSeconds is invalid. "
+                f"Reason: Must be >= 0 and <= 20, if provided."
+            )
         num = max_number_of_messages or 1
 
         # override receive count with value from custom header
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/sqs/provider.py
```diff
@@ -1232,9 +1235,17 @@ def receive_message(
...
+            wait_time_seconds = override
+            poll_empty_queue = True
+        elif wait_time_seconds is None:
             wait_time_seconds = queue.wait_time_seconds
-
+        elif wait_time_seconds < 1 or wait_time_seconds > 20:
+            raise InvalidParameterValueException(
+                f"Value {wait_time_seconds} for parameter WaitTimeSeconds is invalid. "
+                f"Reason: Must be >= 0 and <= 20, if provided."
+            )
```
dfangl, line 1248: The error message would lead me to believe 0 is a valid value, but the condition does not. Did you verify this against AWS?

tests/aws/services/sqs/test_sqs_backdoor.py
```diff
@@ -361,3 +364,128 @@ def test_list_messages_with_queue_url_in_path(
...
+
+
+class TestSqsOverrideHeaders:
+    @markers.aws.only_localstack
+    def test_receive_message_override_max_number_of_messages(
+        self, sqs_create_queue, aws_client_factory
+    ):
+        # Create standalone boto3 client since registering hooks to the session-wide
+        # aws_client (from the fixture) will have side-effects.
+        aws_client = aws_client_factory().sqs
```
dfangl, line 376: Could you call this variable differently? I think the confusion potential with the `aws_client` fixture is quite high here!

tests/aws/services/sqs/test_sqs_backdoor.py
```diff
@@ -361,3 +364,128 @@ def test_list_messages_with_queue_url_in_path(
...
+
+        override_max_number_of_messages = 20
+
+        from localstack.services.sqs.constants import HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT
+        from localstack.services.sqs.provider import MAX_NUMBER_OF_MESSAGES
```
dfangl, line 381: Why not import them on the top of the test function? Or on a module level altogether?

tests/aws/services/sqs/test_sqs_backdoor.py
```diff
@@ -361,3 +364,128 @@ def test_list_messages_with_queue_url_in_path(
...
+
+        def _send_message():
+            aws_client.send_message(QueueUrl=queue_url, MessageBody=f"message-{short_uid()}")
+
+        # Populate with 9 messages (1 below the MaxNumberOfMessages threshold).
+        # This should cause long-polling to exit since MaxNumberOfMessages is met.
+        for _ in range(9):
+            _send_message()
+
+        Timer(25, _send_message).start()  # send message asynchronously after 1 second
```
dfangl, line 478: After 25 seconds, right?

tests/aws/services/sqs/test_sqs_backdoor.py
```diff
@@ -361,3 +364,128 @@ def test_list_messages_with_queue_url_in_path(
...
+
+        start_t = time.time()
```
dfangl, line 480: We should use `time.perf_counter` here, as `time.time` is not guaranteed to be monotonic. It is quite unlikely, but it could result in less than 25 seconds measured, if the system clock is altered in that timeframe.

dfangl (commented):
Just some minor questions, LGTM in general!

localstack-core/localstack/services/sqs/provider.py
gregfurman: Ahh good catch. I added a test for validation in this PR but didn't test the lower bound. Will change, thanks

https://github.com/localstack/localstack/blob/baa35efdb7f598b18dd0e715c1e190737f70efc7/tests/aws/services/sqs/test_sqs.py#L269-L280

dfangl (approved):
Thanks for addressing my feedback so quickly, LGTM!

Issue #12174, opened 2025-01-23
bug: SNS subscribe doesn't raise error when topic does not exist


coleman-c:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I attempt to subscribe to a topic which doesn't exist localstack succeeds and returns a subscription arn.

```bash
user@XYZ:~$ awslocal sns subscribe --topic-arn "arn:aws:sns:us-east-1:000000000000:topic-does-which-does-not-exist" --protocol sqs --notification-endpoint "arn:aws:sqs:us-east-1:000000000000:my-queue"
{
    "SubscriptionArn": "arn:aws:sns:us-east-1:000000000000:localstack-topic-does-not-exist:8c98427f-bfea-4b0f-88ab-4a377e113312"
}
```

Performing the same action against an aws account:

```bash
user@XYZ:~$ aws --profile my-aws-profile sns subscribe --topic-arn "arn:aws:sns:eu-west-1:123456789012:topic-does-which-does-not-exist" --protocol sqs --notification-endpoint "arn:aws:sqs:eu-west-1:123456789012:my-queue"

An error occurred (NotFound) when calling the Subscribe operation: Topic does not exist
```

### Expected Behavior

To return the topic does not exist error in line with the aws behavior.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```yml
services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME-localstack_main}"
    image: localstack/localstack:latest
    ports:
      - "4566:4566"           
    environment:
      - HOSTNAME_EXTERNAL=localstack
      - DEBUG=${LOCASTACK_DEBUG:-1}
      - LAMBDA_EXECUTOR=${LAMBDA_EXECUTOR-}
      - DOCKER_HOST=unix:///var/run/docker.sock
      - LS_LOG=trace
    volumes:
      - "${LOCALSTACK_DATA_DIR:-./.data/localstack/internal}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
    networks:
      local-development:
        aliases:     
          - sqs.us-east-1.localstack
          - sns.us-east-1.localstack
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```bash
user@XYZ:~$ awslocal  sqs create-queue --queue-name my-queue
{
[...]

PR #12175, opened 2025-01-23
SNS: fix Subscribe non existent topic exception

bentsku:
## Motivation
As reported in #12174, we did not raise an exception if the topic did not exist

## Changes
- add a test for raising an exception if the topic did not exist
- raise the exception

_fixes #12174_

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -651,6 +651,9 @@ class SnsProvider(SnsApi, ServiceLifecycleHook):
         parsed_topic_arn = parse_and_validate_topic_arn(topic_arn)
         store = self.get_store(account_id=parsed_topic_arn["account"], region_name=context.region)
 
+        if topic_arn not in store.topic_subscriptions:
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/sns/provider.py
```diff
@@ -651,6 +651,9 @@ def subscribe(
         parsed_topic_arn = parse_and_validate_topic_arn(topic_arn)
         store = self.get_store(account_id=parsed_topic_arn["account"], region_name=context.region)
 
+        if topic_arn not in store.topic_subscriptions:
```
anisaoshafi, line 654: **question**: does this error only happen for `subscribe` fnx?

bentsku: Very good question! This error should happen for every operation targeting a Topic somehow. I think we were already properly raising an exception for `Publish` for example, via the `_get_topic` method. As the `Subscribe` operation is quite complex, and does not really have the need for the `_get_topic` method, we check it a bit differently.

PR #12176, opened 2025-01-24
CFN/EC2: fix KeyError for EC2:Instance PublicIpAddress

bentsku:
## Motivation
We got a report from a support case that trying to deploy an EC2 instance with CDK and a basic stack would fail with a `KeyError`:
```python
2025-01-10 08:31:51 localstack-main  |   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/ec2/resource_providers/aws_ec2_instance.py", line 255, in create
2025-01-10 08:31:51 localstack-main  |     model["PublicIp"] = instance["PublicIpAddress"]
2025-01-10 08:31:51 localstack-main  |                         ~~~~~~~~^^^^^^^^^^^^^^^^^^^
2025-01-10 08:31:51 localstack-main  | KeyError: 'PublicIpAddress'
```

<details><summary>The Stack:</summary>
<p>

```python
from aws_cdk import (
    Stack,
    aws_ec2 as ec2,
    CfnOutput,
)
from constructs import Construct

class CdkStack(Stack):

    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
        super().__init__(scope, construct_id, **kwargs)

        # Create a VPC with outbound internet access
        vpc = ec2.Vpc(self, "Vpc", max_azs=3, vpc_name="my-vpc")

        # Create an ec2 instance
        instance = ec2.Instance(
            self,
             "Instance",
            vpc=vpc,
            instance_type=ec2.InstanceType("t3.nano"),
            instance_name="my-instance",
            machine_image=ec2.MachineImage.lookup(
                name="amazonlinux-2023"
            ),
            key_name="my-key"
        )
```

</p>
</details> 

(Note: To run on AWS, I had to update the stack to use `machine_image=ec2.MachineImage.latest_amazon_linux2(),` to find the proper AMI, and also create a key pair).
 
This is due because the `PublicIpAddress` response field is optional, and is not given to be returned depending on the configuration of the instance (and if the VPC of the instance is the default one or not, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-public-addresses).

This stack did not failed for me, possibly due to the different platform (macOS). 

I do not know how to add a test for this, because spawning real instance is pretty difficult as it is very platform dependent, and the fact that we do not return the `PublicIpAddress` seems to also be dependent on the platform. 

I've done a lot of testing on AWS, and with this stack, the instance does not have a Public Address, but it deploys correctly. However, trying to add the following to the stack and deploying again, or deploying from scratch:
```python
CfnOutput(self, "TestOutput", value=instance.instance_public_ip)
```
See https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Instance.html#instancepublicip

Will fail in AWS with the following message: `Attribute 'PublicIp' does not exist`, meaning that the field is actually not attached to the "model", and thus failing. 
I believe we can get close to this behavior by conditionally adding the field to the model. 

If you have an idea about the testing, please feel free to suggest it. 
 

## Changes
- only add the `PublicIpAddress` and `PublicDnsName` to the CFN `model` if those are present

### DIFF
--- a/localstack-core/localstack/services/ec2/resource_providers/aws_ec2_instance.py
+++ b/localstack-core/localstack/services/ec2/resource_providers/aws_ec2_instance.py
@@ -252,11 +252,20 @@ class EC2InstanceProvider(ResourceProvider[EC2InstanceProperties]):
                 resource_model=model,
                 custom_context=request.custom_context,
             )
-        model["PublicIp"] = instance["PublicIpAddress"]
-        model["PublicDnsName"] = instance["PublicDnsName"]
+
         model["PrivateIp"] = instance["PrivateIpAddress"]
         model["PrivateDnsName"] = instance["PrivateDnsName"]
         model["AvailabilityZone"] = instance["Placement"]["AvailabilityZone"]
+
+        # PublicIp is not guaranteed to be returned by the request:
+        # https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Instance.html#instancepublicip
+        # it says it is supposed to return an empty string, but trying to add an output with the value will result in
+        # an error: `Attribute 'PublicIp' does not exist`
+        if public_ip := instance.get("PublicIpAddress"):
+            model["PublicIp"] = public_ip
+
+        if public_dns_name := instance.get("PublicDnsName"):
+            model["PublicDnsName"] = public_dns_name
 
         return ProgressEvent(
             status=OperationStatus.SUCCESS,


macnev2013 (approved):
Thanks for the fix @bentsku! 🦸🏻  LGTM 🚀 

I tested this along with @bentsku using Java CDK, and here are the findings:  

### On AWS:
- If the instance does not have a `PublicIp`, CloudFormation rolls back the stack.  

### On LocalStack:
- When the `PublicIp` field is missing in the response, the CloudFormation output shows `undefined`.
- When the `PublicIp` field is present but empty (i.e., `""`), the CloudFormation output displays an empty string.

Issue #12179, opened 2025-01-24
SNS UnsubscribeConfirmation message is incorrectly signed


chringwer:
There is a typo in the message type so that incorrect fields are considered for signing the message:

"UnsubscriptionConfirmation" -> "UnsubscribeConfirmation" 

https://github.com/localstack/localstack/blob/7b5808198e0a29a3b3451c0566e3827032ec0e5a/localstack-core/localstack/services/sns/publisher.py#L927

PR #12181, opened 2025-01-24
SNS : fix Message Signature typo and add Lambda URL fixture

bentsku:
## Motivation
As reported by #12179, we were not calculating the message signature correctly for `UnsubscribeConfirmation` message for HTTP/HTTPS subscriptions.

I've created an `StrEnum` to avoid this kind of typo in the future.

I had also been talking about creating a Lambda URL fixture to allow forwarding received message to an HTTP endpoint to an SQS queue, so this was a good occasion to do it. 

This can be used as example for other services as well, the fixture could probably live in the global fixture file, once we see we can reuse it, as it can be dependent on what the HTTP endpoint needs to return. 
But I'm thinking of making this fixture to do more SNS related things, like being able to return different status code to enable retries and such, or auto-approve the subscription, so we could probably duplicate it in a more general way.

## Changes
- add a new lambda URL function fixture to forward input events to a queue, that allows exposing an accessible HTTP endpoint
- fix the typo and introduce an `StrEnum` to avoid future typos
- create a test to validate the behavior and the proper signature of messages for HTTP endpoints

_fixes #12179_

### DIFF
--- a/localstack-core/localstack/services/sns/models.py
+++ b/localstack-core/localstack/services/sns/models.py
@@ -1,6 +1,7 @@
 import itertools
 import time
 from dataclasses import dataclass, field
+from enum import StrEnum
 from typing import Dict, List, Literal, Optional, TypedDict, Union
 
 from localstack.aws.api.sns import (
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/sns/models.py
```diff
@@ -37,9 +38,15 @@ def get_next_sequence_number():
     return next(global_sns_message_sequence())
 
 
+class SnsMessageType(StrEnum):
+    notification = "Notification"
+    subscription_confirmation = "SubscriptionConfirmation"
+    unsubscribe_confirmation = "UnsubscribeConfirmation"
```
silv-io, line 44: **nit:** I think ASF would create such enums in the original CamelCase, but I don't think this matters a lot :D

silv-io (approved):
LGTM! well structured test, nice fixture and good to keep a more central hold on the constant strings 👍🏼

localstack-core/localstack/services/sns/models.py
bentsku: I was exactly thinking about that. Will update and merge 👍

PR #12183, opened 2025-01-25
S3: implement new data integrity for MultipartUpload

bentsku:
## Motivation

This PR implements the new data integrity checks for Multipart Upload (shortened MPU), namely the ability to set the `FULL_OBJECT` checksum type for a MPU, allowing the user to not keep track of individual checksum parts to calculate the final one, but instead send the full object checksum to S3. This can be done thanks to `CRC*` type checksums, which can linearize, meaning you can get the full object checksum from the individual parts' checksums (via a somewhat complicated function). 

The PR looks big (3k lines), but 70% of it is snapshots, because I had to cover a lot of different cases, operations' responses, and there was a lot of ground to cover that wasn't previously covered. 

There is a lot of additional validation and default values to check (different relationship between checksum algorithm and checksum type), new validation of the MPU total size at completion time, I've also ported the fix with different error message depending if the checksum value is valid or not.

The biggest part and challenge of this PR is to implement to logic to combine CRC checksums together in order to create the `FULL_OBJECT` checksum. More can be read about this in the following Stack Overflow post: https://stackoverflow.com/a/23126768
I had to look everywhere online to try to find someway to do that, and it was pretty hard to find. I found an old repository of Alibaba, with the function available in Python2 to combine CRC64 checksums, I had to adapt it a little to run with Python3 and fit our use case better. 

More resources to read about the new S3 data integrity push:
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
- https://docs.aws.amazon.com/sdkref/latest/guide/feature-dataintegrity.html

There's also a bit of cleanup done, removing some unused functions/code.

## Changes
- implement the new checksum types `FULL_OBJECT` and `COMPOSITE` for MPU
- implement the helpers to combine CRC checksums together
- clean up some dead code
- implement a lot of new tests covering as wide as I could think

### DIFF
--- a/localstack-core/localstack/services/s3/utils.py
+++ b/localstack-core/localstack/services/s3/utils.py
@@ -7,7 +7,7 @@ import re
 import zlib
 from enum import StrEnum
 from secrets import token_bytes
-from typing import IO, Any, Dict, Literal, NamedTuple, Optional, Protocol, Tuple, Union
+from typing import Any, Dict, Literal, NamedTuple, Optional, Protocol, Tuple, Union
 from urllib import parse as urlparser
 from zoneinfo import ZoneInfo
 
@@ -54,12 +54,12 @@ from localstack.aws.api.s3 import Type as GranteeType
 from localstack.aws.chain import HandlerChain
 from localstack.aws.connect import connect_to
 from localstack.http import Response
+from localstack.services.s3 import checksums
 from localstack.services.s3.constants import (
     ALL_USERS_ACL_GRANTEE,
     AUTHENTICATED_USERS_ACL_GRANTEE,
     CHECKSUM_ALGORITHMS,
     LOG_DELIVERY_ACL_GRANTEE,
-    S3_CHUNK_SIZE,
     S3_VIRTUAL_HOST_FORWARDED_HEADER,
     SIGNATURE_V2_PARAMS,
     SIGNATURE_V4_PARAMS,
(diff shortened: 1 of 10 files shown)


tests/unit/test_s3.py
bentsku, line 1: This file was testing botocore code, so it wasn't really useful

tests/aws/services/s3/test_s3.py
bentsku, line 1: Sorry for all the duplication in the tests, multipart uploads are super verbose and annoying to work with

localstack-core/localstack/services/s3/utils.py
```diff
@@ -390,40 +412,6 @@ def get_full_default_bucket_location(bucket_name: BucketName) -> str:
...
-    # TODO: is there a message if the header is missing?
-    checksum = request.get(key)
-    calculated_checksum = get_object_checksum_for_algorithm(checksum_algorithm, data)
-
-    if calculated_checksum != checksum:
-        raise InvalidRequest(
-            f"Value for x-amz-checksum-{checksum_algorithm.lower()} header is invalid."
-        )
-
-
```
bentsku, line 426: this code wasn't used anymore anywhere, since we're properly calculating checksums on streams instead of the full string

localstack-core/localstack/services/s3/utils.py
```diff
@@ -454,40 +442,6 @@ def base_64_content_md5_to_etag(content_md5: ContentMD5) -> str | None:
     return hex_digest
 
 
-def decode_aws_chunked_object(
```
bentsku, line 457: we did not use this code anymore, as `localstack.services.s3.codec.AwsChunkedDecoder` superseded it

k-a-il (approved):
Nice work! Code and the whole solution  is easy to understand thanks to comments :)

Issue #12186, opened 2025-01-25
bug: StepFunction throwing InvalidDefinition for "Version" MessageAttribute


Travis-Kirton:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Creating a StepFunction and it seems if you use a MessageAttribute with the key of `Version`, the step function fails.

_State machine JSON_
```
{
  "StartAt": "Delay",
  "Comment": "Publish to SNS with delay",
  "States": {
    "Delay": {
      "Type": "Wait",
      "Seconds": 5,
      "Next": "Publish to SNS"
    },
    "Publish to SNS": {
      "Type": "Task",
      "Resource": "arn:aws:states:::sns:publish",
      "Parameters": {
        "TopicArn": "${var.offer_sfn_sns_topic_arn}",
        "Message.$": "$.message",
        "MessageAttributes": {
          "Version": {
            "DataType": "String",
            "StringValue": "Hello World"
          }
        }
      },
      "End": true
    }
  }
}
```
_Command_
```
awslocal stepfunctions create-state-machine --name "test" \
      --role-arn "arn:aws:iam::000000000000:role/step_function_role" \
      --definition file://sfn_definition.json \
      --type STANDARD
```
_Output_
```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: ASLParserException ['line 17:10, at "Version", no viable alternative at input \'{"Version"\'', 'line 23:6, at "End", mismatched input \'"End"\' expecting {\'"Comment"\', \'"States"\', \'"StartAt"\', \'"Version"\', \'"TimeoutSeconds"\', \'"QueryLanguage"\'}', "line 25:2, at }, mismatched input '}' expecting <EOF>"]
```

When I change `Version` to `Version1` or even lowercase `version`, the StepFunction is created
```
{
    "stateMachineArn": "arn:aws:states:us-west-2:000000000000:stateMachine:test",
    "creationDate": "2025-01-25T17:43:17.180425+00:00"
}
```

### Expected Behavior

Expected behaviour is that there are no reserved MessageAttribute names and `Version` is allowed to be used.

### How are you starting LocalStack?

Running localstack (4.0.3) via docker-compose
```
localstack_main:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack:4.0.3
    ports:
      - 4566:4566            # LocalStack Gateway
      - 4510-4559:4510-4559  # external services port range
[...]

PR #12190, opened 2025-01-27
StepFunctions: Fix Parsing of Version Tokens as Soft Keywords

MEPalma:
## Motivation
This PR resolves an issue where the StepFunctions parser treated the `Version` token as a hard keyword across the entire state machine definition. As a result, users were unable to declare payload template bindings using a field named `Version` https://github.com/localstack/localstack/issues/12186.

## Changes
- The parser now recognizes "Version" as a soft keyword only when used in a top-level version declaration. In all other contexts (e.g., within state definitions or payload templates), `Version` behaves like a normal field name.
- Added and updated snapshot tests to verify that the `Version` keyword can be used in payload templates. We already have snapshot tests to verify the use of `Version` as a top layer declaration.


gregfurman (approved):
LGTM!

Issue #9284, opened 2023-10-04
bug: EventBridge Scheduler CreateSchedule operation not validating ScheduleExpression


KrazyK786:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When creating EventBridge Schedule with the `CreateSchedule` operation, passing an invalid `ScheduleExpression` is still successful

### Expected Behavior

Calling `CreateSchedule` with an invalid `ScheduleExpression` should return:

`ValidationException: Invalid Schedule Expression <expression>`

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack (via Testcontainers for Go)

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```go
import (
    "fmt"
    "time"

    "github.com/aws/aws-sdk-go/aws"
    "github.com/aws/aws-sdk-go/aws/session"
    "github.com/aws/aws-sdk-go/service/scheduler"
)

mySession := session.Must(session.NewSession())
svc := scheduler.New(mySession, aws.NewConfig().WithEndpoint("http://localhost:4566").WithRegion("us-east-2"))

interval := time.Hour * 24
cTime := time.Now()

_, err := svc.CreateSchedule(&scheduler.CreateScheduleInput{
    Name: aws.String("Test Schedule"),
    FlexibleTimeWindow: &scheduler.FlexibleTimeWindow{
	Mode: aws.String("OFF"),
    },
    ScheduleExpression:         aws.String(cTime.Add(interval).Format("2006-01-02T15:04:05")),  // invalid
    ScheduleExpressionTimezone: aws.String("America/Los_Angeles"),
    Target: &scheduler.Target{
	RoleArn: aws.String("arn:aws:iam::000000000000:root"),
	Arn:     aws.String("arn:aws:lambda:us-east-2:000000000000:function:test-lambda-function"),
    },
    ActionAfterCompletion: aws.String("DELETE"),
})
if err != nil {
    return err
}

actualSchedule, err := svc.GetSchedule(&scheduler.GetScheduleInput{
    Name: aws.String("Test Schedule"),
})
if err != nil {
    return err
}

fmt.Println("Expression: ", *actualSchedule.ScheduleExpression)  // prints: Expression:  2023-10-04T21:54:24
```

Running these same steps against AWS returns:
```bash
[...]

PR #12191, opened 2025-01-27
Fix: validate schedule_expression for EventBridge Scheduler

anisaoshafi:
## Motivation
Customer reported an issue for scheduler validation of schedule_expression  https://github.com/localstack/localstack/issues/9284.
The scheduler relies fully on moto for now, and validation is not implemented for `schedule_expression`.
Schedule_expression should support `at`, `cron` and `rate`.

## Changes
Patched the `create_schedule` method in order to validate `schedule_expression` .
Reused some REGEX rules from events service.

### DIFF
--- a/localstack-core/localstack/services/scheduler/provider.py
+++ b/localstack-core/localstack/services/scheduler/provider.py
@@ -1,10 +1,34 @@
 import logging
+import re
 
-from localstack.aws.api.scheduler import SchedulerApi
+from moto.scheduler.models import EventBridgeSchedulerBackend
+
+from localstack.aws.api.scheduler import SchedulerApi, ValidationException
+from localstack.services.events.rule import RULE_SCHEDULE_CRON_REGEX, RULE_SCHEDULE_RATE_REGEX
 from localstack.services.plugins import ServiceLifecycleHook
+from localstack.utils.patch import patch
 
 LOG = logging.getLogger(__name__)
+
+AT_REGEX = r"^at[(](0[1-9]|1\d|2[0-8]|29(?=-\d\d-(?!1[01345789]00|2[1235679]00)\d\d(?:[02468][048]|[13579][26]))|30(?!-02)|31(?=-0[13578]|-1[02]))-(0[1-9]|1[0-2])-([12]\d{3}) ([01]\d|2[0-3]):([0-5]\d):([0-5]\d)[)]$"
+RULE_SCHEDULE_AT_REGEX = re.compile(AT_REGEX)
 
 
 class SchedulerProvider(SchedulerApi, ServiceLifecycleHook):
     pass
+
+
+def _validate_schedule_expression(schedule_expression: str) -> None:
+    if not (
+        RULE_SCHEDULE_CRON_REGEX.match(schedule_expression)
+        or RULE_SCHEDULE_RATE_REGEX.match(schedule_expression)
+        or RULE_SCHEDULE_AT_REGEX.match(schedule_expression)
+    ):
+        raise ValidationException(f"Invalid Schedule Expression {schedule_expression}.")
+
+
+@patch(EventBridgeSchedulerBackend.create_schedule)
+def create_schedule(fn, self, **kwargs):
+    if schedule_expression := kwargs.get("schedule_expression"):
+        _validate_schedule_expression(schedule_expression)
+    return fn(self, **kwargs)
--- a/tests/aws/services/scheduler/test_scheduler.py
+++ b/tests/aws/services/scheduler/test_scheduler.py
@@ -1,4 +1,5 @@
 import pytest
+from botocore.exceptions import ClientError
 
 from localstack.testing.aws.util import in_default_partition
 from localstack.testing.pytest import markers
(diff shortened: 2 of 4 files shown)


tests/aws/services/scheduler/test_scheduler.py
```diff
@@ -58,3 +59,47 @@ def test_untag_resource(aws_client, events_scheduler_create_schedule_group, snap
...
+        "rate(10 minutess)",
+        "rate(foo minutes)",
+        "rate(-10 minutes)",
+        "rate( 10 minutes )",
+        " rate(10 minutes)",
+        "at(2021-12-31T23:59:59Z)",
+        "at(2021-12-31)",
+    ],
+)
+def tests_create_schedule_with_invalid_schedule_expression(
```
bentsku, line 87: nice test! 👌

localstack-core/localstack/services/scheduler/provider.py
```diff
@@ -1,10 +1,39 @@
...
 
+AT_REGEX = r"^at[(](0[1-9]|1\d|2[0-8]|29(?=-\d\d-(?!1[01345789]00|2[1235679]00)\d\d(?:[02468][048]|[13579][26]))|30(?!-02)|31(?=-0[13578]|-1[02]))-(0[1-9]|1[0-2])-([12]\d{3}) ([01]\d|2[0-3]):([0-5]\d):([0-5]\d)[)]$"
+RULE_SCHEDULE_AT_REGEX = re.compile(AT_REGEX)
+
 
 class SchedulerProvider(SchedulerApi, ServiceLifecycleHook):
     pass
+
+
+def _validate_schedule_expression(schedule_expression) -> None:
```
bentsku, line 21: nit: we could add a type hint for the `schedule_expression` argument

bentsku (commented):
LGTM! Nice port of your previous fix 👌 nice usage of the `patch` function. 

I only have a comment, it seems there code that cannot be executed, and that path is also untested. Once this is addressed, I believe the PR is ready to be merged! 🚀

localstack-core/localstack/services/scheduler/provider.py
```diff
@@ -1,10 +1,39 @@
...
+    if not schedule_expression:
+        raise ValidationException(
+            "Parameter(s) EventPattern or ScheduleExpression must be specified."
+        )
```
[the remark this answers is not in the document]
anisaoshafi: @bentsku sorry, I'm not following. Could you please rephrase this?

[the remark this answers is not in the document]
bentsku: Sure, sorry! 
Before calling the `_validate_schedule_expression`, you do this:
`if schedule_expression := kwargs.get("schedule_expression"):`. 

Meaning that the `schedule_expression` parameter for `_validate_schedule_expression` cannot be false-ish, thus `if not schedule_expression:` can never be true. 

Also, this specific exception message `"Parameter(s) EventPattern or ScheduleExpression must be specified."` is not in the snapshot, so I figured it was not tested.

Maybe we could add a small test case for this? Or remove this validation?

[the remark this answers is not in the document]
anisaoshafi: thanks, you're right. I'm going to double check it 🙏🏼

tests/aws/services/scheduler/test_scheduler.py
```diff
@@ -103,3 +103,46 @@ def tests_create_schedule_with_invalid_schedule_expression(
...
+            FlexibleTimeWindow={
+                "MaximumWindowInMinutes": 4,
+                "Mode": "FLEXIBLE",
+            },
+            Target={
+                "Arn": f"arn:aws:lambda:{region_name}:{account_id}:function:dummy",
+                "RoleArn": f"arn:aws:iam::{account_id}:role/role-name",
+            },
+        )
+    snapshot.match("empty-schedule-expression", e.value)
```
bentsku, line 148: `ParamValidationError` indicates that the error is coming from the client itself, and the request is not sent to AWS or LocalStack. We don't need to snapshot those kind of exception as they are dependent on the client only. 

To configure the client to not have validation, and send the invalid request against AWS, you can configure you client this way:
```python
from botocore.client import Config

def tests_create_schedule_with_empty_schedule_expression(
    aws_client_factory, region_name, account_id, snapshot
):
    scheduler_client = aws_client_factory(config=Config(parameter_validation=False)).scheduler
    rule_name = f"rule-{short_uid()}"

   with pytest.raises(ParamValidationError) as e:
        scheduler_client.create_schedule(
        [...]
```

This way, the client won't validate the request and will let you send the invalid request against AWS, and see the proper exception.

anisaoshafi: understood, thanks for sharing this. I'll keep it in mind for the future and will proceed removing these two dummy tests.

anisaoshafi:
> LGTM! Nice port of your previous fix 👌 nice usage of the `patch` function.
> 
> I only have a comment, it seems there code that cannot be executed, and that path is also untested. Once this is addressed, I believe the PR is ready to be merged! 🚀


@bentsku Thanks for your comments. Could you please double check this when you have a moment? Thanks 🙌🏼

bentsku (approved):
LGTM! Thanks for addressing the comments. 

I think we could have kept the tests when the `schedule_expression` was missing, but with the client parameter validation disabled to see what AWS returned, however it's not fully in the scope of the issue, so that's alright, it was just a nice to have as you had already added it 😄 and it's pretty niche as most of the SDK will validate the input. 

Anyway, very nice fix and parametrized tests! 🚀💯

anisaoshafi:
> LGTM! Thanks for addressing the comments.
> 
> I think we could have kept the tests when the `schedule_expression` was missing, but with the client parameter validation disabled to see what AWS returned, however it's not fully in the scope of the issue, so that's alright, it was just a nice to have as you had already added it 😄 and it's pretty niche as most of the SDK will validate the input.
> 
> Anyway, very nice fix and parametrized tests! 🚀💯

Sorry I deleted everything. I'll merge this as is and if I come back to eventbridge might add that test 🙏🏼

bentsku:
> Sorry I deleted everything. I'll merge this as is and if I come back to eventbridge might add that test 🙏🏼

No worries! You already went through the effort, that was all 😄 but this PR is good as it is, so let's get this merged! 🤝

ridgey-dev:
I now get the error:

Invalid Schedule Expression at(2025-01-31T15:59:42)

with value:

`at(2025-01-31T15:59:42)`

Looking at AWS documentation [https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html#one-time](url) it's valid, but looking at the regex in the PR it's not?

anisaoshafi:
Hi @ridgey-dev. Thanks for reporting this 🙏🏼 
I double checked with aws and the date you passed seems valid.
> it's expecting d-m-Y instead of Y-m-d?

No, it's "Y-m-d", I'll fix the validation rule since I closed it too much.

ridgey-dev:
Thanks for your quick answer. Looking forward to your fix! Our pipeline that runs the tests on LocalStack is now failing because of this.

anisaoshafi:
Hi @ridgey-dev, thank you for your patience. We have implemented changes that should resolve your reported issue. Kindly update to the latest LocalStack container image and let us know if your issue has been resolved.

ridgey-dev:
It's working again. Thanks for your quick fix!

PR #12192, opened 2025-01-27
add resource provider utility to convert LowerCamelCase to Pascal

pinzon:
## Motivation
This PR adds an utility to convert LowerCamelCase to PascalCase keys in a dictionary. Usefull for read methods in resource providers

## Changes
- new utility for resource providers

### DIFF
--- a/localstack-core/localstack/services/cloudformation/provider_utils.py
+++ b/localstack-core/localstack/services/cloudformation/provider_utils.py
@@ -52,6 +52,13 @@ def convert_pascalcase_to_lower_camelcase(item: str) -> str:
         return f"{item[0].lower()}{item[1:]}"
 
 
+def convert_lower_camelcase_to_pascalcase(item: str) -> str:
+    if len(item) <= 1:
+        return item.upper()
+    else:
+        return f"{item[0].upper()}{item[1:]}"
+
+
 def _recurse_properties(obj: dict | list, fn: Callable) -> dict | list:
     obj = fn(obj)
     if isinstance(obj, dict):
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/cloudformation/provider_utils.py
```diff
@@ -78,6 +85,18 @@ def _keys_pascalcase_to_lower_camelcase(obj):
     return _recurse_properties(model, _keys_pascalcase_to_lower_camelcase)
 
 
+def keys_lower_camelcase_to_pascalcase(model: dict) -> dict:
```
simonrw, line 88: nit: I notice this function is not used in the companion PR. Can you give an example of its usage in the PR description?

simonrw (changes requested):
I would like to see some unit tests that execute these functions before I would be happy with the implementation. In general some usage examples are required, either in docstrings or unit tests to demonstrate how the functions work.

In particular, I feel the `convert_lower_camelcase_to_pascalcase("my_value")` should equal `MyValue` however it equals `My_value`.

pinzon:
@simonrw I addressed the request for an unit test (which I should have done from the beginning). 

About this convertion: `my_value -> MyValue` Looks for me that is out the scope of this function. We already have a function that converts Pascal Case to Lower Camel Case, which is pretty similar to what I wrote in this PR. 

For what you want we should have one to convert Snake Case to Camel Case (either lower or pascal) first.

cloutierMat (approved):
LGTM! We still need @simonrw approval has he requested changes

simonrw (approved):

PR #12198, opened 2025-01-28
add product and edition to session analytics

ackdav:
## Motivation
This PR adds `product` and `edition` to the session's `ClientMetadata`. Given LocalStack's growing number of products, the distinction will help distinguish where a session is coming from. 

Both `product` and `edition` can be overwritten by environment variables to cover cases where a runtime is not necessarily given or correct (e.g. when packaging an extension).

## Changes
- adds `product` and `edition` to the session's `ClientMetadata`
- adds `mock_runtime` fixture. Analytics unit tests rely on client metadata - but `get_current_runtime()` [throws an error](https://github.com/localstack/localstack/blob/873d150c009f4bc6acaee6a4c48a57ab53c66f0e/localstack-core/localstack/runtime/current.py#L20) when not having an initialized runtime.

### DIFF
--- a/localstack-core/localstack/utils/analytics/metadata.py
+++ b/localstack-core/localstack/utils/analytics/metadata.py
@@ -5,7 +5,7 @@ import platform
 from typing import Optional
 
 from localstack import config
-from localstack.constants import VERSION
+from localstack import VERSION, ENV_PRO_ACTIVATED
 from localstack.runtime import hooks
 from localstack.utils.bootstrap import Container
 from localstack.utils.files import rm_rf
@@ -60,6 +60,8 @@ def read_client_metadata() -> ClientMetadata:
         is_ci=os.getenv("CI") is not None,
         is_docker=config.is_in_docker,
         is_testing=config.is_local_test_mode(),
+        product="aws",
+        edition="pro" if config.is_env_true(ENV_PRO_ACTIVATED) else "community"
     )
 
 
--- a/localstack-core/localstack/utils/analytics/metadata.py
+++ b/localstack-core/localstack/utils/analytics/metadata.py
@@ -5,8 +5,8 @@ import platform
 from typing import Optional
 
 from localstack import config
-from localstack.constants import ENV_PRO_ACTIVATED, VERSION
-from localstack.runtime import hooks
+from localstack.constants import VERSION
+from localstack.runtime import get_current_runtime, hooks
 from localstack.utils.bootstrap import Container
 from localstack.utils.files import rm_rf
 from localstack.utils.functions import call_safe
@@ -53,6 +53,8 @@ def get_version_string() -> str:
 
 
 def read_client_metadata() -> ClientMetadata:
+    runtime = get_current_runtime()
+
     return ClientMetadata(
         session_id=get_session_id(),
         machine_id=get_machine_id(),
@@ -62,8 +64,8 @@ def read_client_metadata() -> ClientMetadata:
         is_ci=os.getenv("CI") is not None,
         is_docker=config.is_in_docker,
         is_testing=config.is_local_test_mode(),
-        product="aws",
-        edition="pro" if config.is_env_true(ENV_PRO_ACTIVATED) else "community",
+        product=os.getenv("LOCALSTACK_TELEMETRY_PRODUCT") or runtime.components.name or "unknown",
+        edition=os.getenv("LOCALSTACK_TELEMETRY_EDITION") or get_localstack_edition(),
     )
 
 
--- a/localstack-core/localstack/utils/analytics/metadata.py
+++ b/localstack-core/localstack/utils/analytics/metadata.py
@@ -5,7 +5,7 @@ import platform
 from typing import Optional
 
 from localstack import config
-from localstack import VERSION, ENV_PRO_ACTIVATED
+from localstack.constants import ENV_PRO_ACTIVATED, VERSION
 from localstack.runtime import hooks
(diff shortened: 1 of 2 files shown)


vittoriopolverino:
@ackdav @thrau 
If we aim to align with the decisions in the [Top-level products](https://www.notion.so/Top-level-products-183fc2a234318056bb81d1f1d0f407cb?d=186fc2a234318093acfb001c7c29e318&pvs=4#184fc2a2343180b1aff8cdc86aaac204) document, where the edition seems to map 1:1 with the Docker image name used by customers, we might want to refactor `get_localstack_edition()` to something like this:

```python
def get_localstack_edition() -> Optional[Literal["localstack-enterprise", "localstack-pro", "localstack", "localstack-azure-alpha", "unknown"]]:
    image_name = get_docker_image_to_start()
    edition = image_name.removeprefix("localstack/")
    return edition if edition in {"localstack-enterprise", "localstack-pro", "localstack", "localstack-azure-alpha"} else "unknown"
```

Then, we could use this function to set the `edition` attribute in the `ClientMetadata` class.

❓ Are there any planned changes to the naming of the Docker images?

localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -60,6 +62,8 @@ def read_client_metadata() -> ClientMetadata:
         is_ci=os.getenv("CI") is not None,
         is_docker=config.is_in_docker,
         is_testing=config.is_local_test_mode(),
+        product="aws",
```
whummer, line 65: @ackdav As briefly discussed, perhaps we can add an environment variable override here, to allow specialized images (separate products like Azure, Snowflake) to define custom product strings.

Something like (simplified):
```suggestion
        product=os.getenv("LOCALSTACK_TELEMETRY_PRODUCT") or "aws",
```
Same for the edition, and also use the existing `get_localstack_edition()` function as the default.

tests/unit/utils/analytics/conftest.py
```diff
@@ -1,9 +1,26 @@
...
+class MockRuntime:
+    components = MockComponents()
+
+
+@pytest.fixture(autouse=True)
+def mock_runtime():
+    runtime = MockRuntime()
+    set_current_runtime(runtime)
+    yield
+    set_current_runtime(None)
```
alexrashed, line 26: **nitpick (not blocking):** This really just affects the analytics module, but it might make sense to play safe and really only set it if the runtime is really not set. Something like this:
```suggestion
    try:
        # don't do anything if a runtime is set
        get_current_runtime()
        yield
    except ValueError:
        # set a mock runtime if no runtime is set
        set_current_runtime(MockRuntime())
        yield
        set_current_runtime(None)
```

localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -60,6 +64,8 @@ def read_client_metadata() -> ClientMetadata:
         is_ci=os.getenv("CI") is not None,
         is_docker=config.is_in_docker,
         is_testing=config.is_local_test_mode(),
+        product=os.getenv("LOCALSTACK_TELEMETRY_PRODUCT") or runtime.components.name or "unknown",
+        edition=os.getenv("LOCALSTACK_TELEMETRY_EDITION") or get_localstack_edition(),
```
alexrashed, line 68: **praise:** Nice and simple enhancement! 💯

localstack-core/localstack/utils/analytics/metadata.py
```diff
@@ -60,6 +64,8 @@ def read_client_metadata() -> ClientMetadata:
         is_ci=os.getenv("CI") is not None,
         is_docker=config.is_in_docker,
         is_testing=config.is_local_test_mode(),
+        product=os.getenv("LOCALSTACK_TELEMETRY_PRODUCT") or runtime.components.name or "unknown",
```
alexrashed, line 67: **super nitpick (absolutely not blocking):** I think it might make sense to create a new function here as well, similar to `get_localstack_edition`. This gives a nice little abstraction layer which allows us to easily adjust the way we determine the product later on.

alexrashed (approved):
I think this is a nice small and concise change! 💯 🚀 
Maybe we could - here in the PR description - add some examples of what the current products and editions are? That would make it more concrete for me.

@vittoriopolverino: Concerning your question on the edition. The edition is right now defined by the "*-version" file in `/usr/lib/localstack/` added when building the Docker image. For example, for community this is happening here:
https://github.com/localstack/localstack/blob/1fca3886080e6d357a50eb06972acc762dc5d6d3/Dockerfile#L176

Does this clarify your question? Happy to discuss this further! :)

whummer (approved):
LGTM, modulo Alex' comments. 👍 Kudos for driving this topic forward @ackdav @vittoriopolverino ! 🚀

vittoriopolverino:
> I think this is a nice small and concise change! 💯 🚀 Maybe we could - here in the PR description - add some examples of what the current products and editions are? That would make it more concrete for me.
> 
> @vittoriopolverino: Concerning your question on the edition. The edition is right now defined by the "*-version" file in `/usr/lib/localstack/` added when building the Docker image. For example, for community this is happening here:
> 
> https://github.com/localstack/localstack/blob/1fca3886080e6d357a50eb06972acc762dc5d6d3/Dockerfile#L176
> 
> Does this clarify your question? Happy to discuss this further! :)

Yes, +1 on highlighting all the expected values for the different Docker images we are currently providing.

PR #12200, opened 2025-01-29
adjust --follow arg positioning to be compatible with docker and podman

cfeller:
## Motivation
As described in this issue:
https://github.com/localstack/localstack/issues/12201

localstack fails to log under podman.  Before this PR, starting logstack when using podman will reach the following error:

```
⠼ Starting LocalStack container2025-01-28T19:11:00.121 DEBUG --- [-log-printer] localstack.utils.run       : Executing command: ['docker', 'logs', '261b4f90dcc37e74bce5b97cf390a7b935fc2d1e0d12ecd41c8df632b7587e76', '--follow']          
────────────────────────────────────────────────────────────────────────────────── LocalStack Runtime Log (press CTRL-C to quit) ───────────────────────────────────────────────────────────────────────────────────
Error: no container with name or ID "--follow" found: no such container  
```
and logging will stop.

technically localstack will start, but no further console logs are produced after the above error.

In short, podman appears to be sensitive to the position of the `-f` or `--follow` argument.  Moving the argument before the container id is compatible with both docker and podman

## Changes
localstack will now start without generating a logging error under podman, allowing console logs to be followed.

### DIFF
--- a/localstack-core/localstack/utils/container_utils/docker_cmd_client.py
+++ b/localstack-core/localstack/utils/container_utils/docker_cmd_client.py
@@ -470,7 +470,7 @@ class CmdDockerClient(ContainerClient):
         self.inspect_container(container_name_or_id)  # guard to check whether container is there
 
         cmd = self._docker_cmd()
-        cmd += ["logs", container_name_or_id, "--follow"]
+        cmd += ["logs", "--follow", container_name_or_id]
 
         process: subprocess.Popen = run(
             cmd, asynchronous=True, outfile=subprocess.PIPE, stderr=subprocess.STDOUT


cfeller:

cfeller:
Corresponding issue created here:
https://github.com/localstack/localstack/issues/12201

cfeller:
Per the contributing [doc](https://github.com/localstack/localstack/blob/master/docs/CONTRIBUTING.md), it appears I need to add the `semver: patch` label, but I don't seem to have the permissions to do so.

dfangl (approved):
Very reasonable change, thank you for your contribution!
This change will be part of the released CLI with the next minor release, which is scheduled for tomorrow!

PR #12202, opened 2025-01-29
raise validationException if resource does not exist in CFn

pinzon:
## Motivation
This PR adds make the CFn provider return the correct exception type in case the `DescribeStackResource` is called for a resource that doesn't exist. The Serverless-cli tool would do this before re-deploying a stack and the users would be confuse about the LocalStack logs. 

## Changes
- Raise and return ValidationException when a resource doesn't exists

## Testing
- new aws validated test.


simonrw (approved):
Looks great, thanks for adding this extra parity!

PR #12203, opened 2025-01-29
fix: Explicitly stop ESM workers on LocalStack shutdown

gregfurman:
## Motivation

ESM workers were not being explicitly shut down when LocalStack was being closed. This PR fixes this by explicilty closing all workers on the `on_before_stop` hook prior to the Lambda service.

## Changes
- Loop over all ESM workers and signal them for graceful shutdown in `on_before_stop`
- Add a `stop_for_shutdown` method that signals to close the `poller_loop`


dfangl (approved):
I think stopping it as part of the service is preferable to using the global TMP_THREADS, nice work!

PR #12214, opened 2025-01-31
[SQS] Stop blocking Queue.get() call on LocalStack shutdown

gregfurman:
## Motivation
Blocking calls in the SQS provider's internal queue implementation cause LocalStack shutdown to hang, and in some cases time-out. 

This PR allows for these blocking `Queue.get` calls to exit early if LocalStack is shutting down and the `do_shutdown` method is called.

## Changes
- Add an `InterruptibleQueue` class that overwrites a `Queue` implementations `get` method to leverage a shutdown `Threading.Event`.
- `InterruptibleQueue` also has a `do_shutdown` method that a queue to be shutdown early if it is currently undergoing a blocking call.
- Add queue closing to the SQS provider's `on_before_stop` lifecycle hook (which happens after persistence) to ensure no hanging threads prevent LS from shutting down.

### DIFF
--- a/localstack-core/localstack/services/sqs/models.py
+++ b/localstack-core/localstack/services/sqs/models.py
@@ -301,7 +301,7 @@ class SqsQueue:
         self.permissions = set()
         self.mutex = threading.RLock()
 
-    def do_shutdown(self):
+    def shutdown(self):
         pass
 
     def default_attributes(self) -> QueueAttributeMap:
@@ -739,7 +739,7 @@ class StandardQueue(SqsQueue):
     def approx_number_of_messages(self):
         return self.visible.qsize()
 
-    def do_shutdown(self):
+    def shutdown(self):
         self.visible.shutdown()
 
     def put(
@@ -966,7 +966,7 @@ class FifoQueue(SqsQueue):
             n += len(message_group.messages)
         return n
 
-    def do_shutdown(self):
+    def shutdown(self):
         self.message_group_queue.shutdown()
 
     def get_message_group(self, message_group_id: str) -> MessageGroup:
--- a/localstack-core/localstack/services/sqs/models.py
+++ b/localstack-core/localstack/services/sqs/models.py
@@ -7,7 +7,7 @@ import re
 import threading
 import time
 from datetime import datetime
-from queue import Empty, PriorityQueue, Queue
+from queue import Empty
 from typing import Dict, Optional, Set
 
 from localstack import config
@@ -28,6 +28,7 @@ from localstack.services.sqs.exceptions import (
     InvalidParameterValueException,
     MissingRequiredParameterException,
 )
+from localstack.services.sqs.queue import InterruptiblePriorityQueue, InterruptibleQueue
 from localstack.services.sqs.utils import (
     decode_receipt_handle,
     encode_move_task_handle,
@@ -299,6 +300,9 @@ class SqsQueue:
 
         self.permissions = set()
         self.mutex = threading.RLock()
+
+    def do_shutdown(self):
+        pass
 
     def default_attributes(self) -> QueueAttributeMap:
         return {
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/sqs/models.py
```diff
@@ -300,6 +301,9 @@ def __init__(self, name: str, region: str, account_id: str, attributes=None, tag
         self.permissions = set()
         self.mutex = threading.RLock()
 
+    def do_shutdown(self):
```
dfangl, line 304: Usually we call `do_shutdown` from a method like `shutdown`, especially when inheriting from an implementation which has a more generic `shutdown` method.

Why did you choose this name here?

localstack-core/localstack/services/sqs/queue.py
```diff
@@ -0,0 +1,43 @@
...
+            return item
+
+    def shutdown(self):
+        self.shutdown_event.set()
+        with self.not_empty:
+            self.not_empty.notify_all()
+
+
+class InterruptiblePriorityQueue(PriorityQueue, InterruptibleQueue):
+    pass
```
dfangl, line 43: Nice!

localstack-core/localstack/services/sqs/queue.py
```diff
@@ -0,0 +1,43 @@
...
+                while not self._qsize() and not self.shutdown_event.is_set():
+                    remaining = endtime - time.time()
+                    if remaining <= 0.0:
+                        raise Empty
+                    self.not_empty.wait(remaining)
+            if self.shutdown_event.is_set():
+                raise Empty
+            item = self._get()
+            self.not_full.notify()
+            return item
```
dfangl, line 34: Would love a comment here to highlight the changes we made from the actual queue implementation, so basically highlighting the event changes.

localstack-core/localstack/services/sqs/queue.py
```diff
@@ -0,0 +1,43 @@
...
+import time
+from queue import Empty, PriorityQueue, Queue
+
+
+class InterruptibleQueue(Queue):
+    shutdown_event: threading.Event
+
+    def __init__(self, maxsize=0):
+        super().__init__(maxsize)
+        self.shutdown_event = threading.Event()
```
dfangl, line 11: Come to think about it, this could be a boolean as well, we never wait for it. WDYT?

dfangl (commented):
Already looks really nice, just some minor comments.
Could we test the shutdown behavior with a bootstrap test perhaps? Do you think that makes sense?

localstack-core/localstack/services/sqs/models.py
gregfurman: Later versions of the python `Queue` already have a shutdown implementation so thought to distinguish this method from that. Also saw some other classes using it so assumed it was standard naming practice.

Will rename it to just `shutdown` -- but we should be cognizant that `Queue.shutdown` will be introduced in python 3.13+.

localstack-core/localstack/services/sqs/queue.py
gregfurman: The python 3.13 implementation actually has a boolean flag instead of a thread (see [Queue.get](https://github.com/python/cpython/blob/808071b994370886a169cfb97cef1ca3837f89c1/Lib/queue.py#L195)). Suppose since we are locking on entry in anycase using a flag makes sense. Will change!

localstack-core/localstack/services/sqs/models.py
dfangl: This one is the sqs queue implementation though, not our inherited one, right?

gregfurman: This method is the SQS model yeah (which contains a `InterruptibleQueue` attribute)

gregfurman: Since the shutdown signal is made to the `SqsQueue` object, I was wanting to distinguish the `SqsQueue.shutdown` from the `SqsQueue.internal_queue.shutdown` since a `shutdown` call would otherwise look like:

```python
# StandardQueue(SqsQueue) 
def shutdown(self):
  self.queue.shutdown()
```

So the `shutdown` method is just us calling another interal `shutdown` method. Was hoping to distinguish these methods with the higher level abstraction calling a `do_shutdown` that (in turn) triggers the _internal_ queue's `shutdown`.

dfangl: Fair enough, but I think it is fine in this case!

gregfurman:
@dfangl Am currently running an -ext test just targeting the SQS persistence tests. Also, will add a boostrap test in a follow-up

PR #12215, opened 2025-01-31
APIGW: binary support for non-proxy integration

bentsku:
## Motivation
Follow up from #12199, is reviewable ~but will come out of draft once the base is merged to avoid the rebase review madness~.

This PR implements the full Binary media support for REST APIs. The AWS documentation will do a much better job explaining what this means exactly, but in big line, it allows you to define some behavior in how your payload in being pass through to your integration: it can be pass as is, converted to a safe UTF-8 string, or base64 decoded or encoded. 

There are precise tables on the behavior, but I tend to not trust the documentation so much, so I took it upon myself to write AWS validated tests, which was to be honest, quite a pain again 😅 

The behavior in itself is not too complicated, it is mapping some table to if conditions (which might be simplified, if you see it during the review, please say so, I didn't optimize the conditions). 

Some changes in behavior is how the `responseTemplates` VTL rendering actually needs the value to be a string, and so if it comes out of the "converter", it needs to be of the right type. We might want to add better log statements to help the user understand why it would fail for them. 

I've also sneaked some `TODO` removal, and a small change in `update_integration_response`, this API op really needs to be validated and fixed. 

Resources:
- https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html
- https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html

## Changes
- validate and remove snapshot skips for S3 AWS integration tests, now passing 🥳 
- add a good amount of binary handling tests in the S3 integration tests, as it's very easy to verify binary behavior with S3, and manipulate returned `Content-Type` headers
- add unit tests for the Convert functions 
- implement the logic for converting the request and response bodies
- ~modify the method~ (ported back to base PR) to check if a request if of "binary" type (AWS terms), meaning it matches the `binaryMediaTypes`
- update `UpdateIntegrationResponse` to support our test case
- remove some todos and update some of them

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -186,13 +186,18 @@ class IntegrationRequestHandler(RestApiGatewayHandler):
         if not template:
             return body, {}
 
+        try:
+            body_utf8 = to_str(body)
+        except UnicodeError:
+            raise InternalServerError("Internal server error")
+
         body, request_override = self._vtl_template.render_request(
             template=template,
             variables=MappingTemplateVariables(
                 context=context.context_variables,
                 stageVariables=context.stage_variables or {},
                 input=MappingTemplateInput(
-                    body=to_str(body),
+                    body=body_utf8,
                     params=MappingTemplateParams(
                         path=request.get("path_parameters"),
                         querystring=request.get("query_string_parameters", {}),
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
@@ -1,9 +1,10 @@
+import base64
 import logging
 from http import HTTPMethod
 
 from werkzeug.datastructures import Headers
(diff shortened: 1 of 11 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/integration_request.py
```diff
@@ -235,6 +243,39 @@ def get_request_template(integration: Integration, request: InvocationRequest) -
...
+            if not content_handling or content_handling == ContentHandlingStrategy.CONVERT_TO_TEXT:
+                body = body.decode(encoding="UTF-8", errors="replace")
+            else:
+                # it means we have CONVERT_TO_BINARY, so we need to try to decode the base64 string
+                try:
+                    body = base64.b64decode(body)
+                except ValueError:
+                    raise InternalServerError("Internal server error")
+
+        return body
```
cloutierMat, line 277: nice digging! There are a few hoops to jump through here as well :rocket:

cloutierMat (approved):
Awesome! :rocket: this adds great parity with aws!

The s3 tests just shows how far we have gone now! The logic of the `convert_body` methods is super well covered in unit tests as well :pray:

PR #12220, opened 2025-02-03
Changed: adjust validation for scheduler expression

anisaoshafi:
## Motivation
Before, the eventbridge scheduler passed no validation at all since all of the logic was defaulting to moto's logic.

We added some validation few days ago, however, seems like the rule for `at()` syntax for scheduled-expression was too limited, since the validation fails on localstack, but succeeds on aws.

This caused a blocker for this user: https://github.com/localstack/localstack/pull/12191#issuecomment-2626725425.

## Changes
Adjusted validation rule for schedule_expression with `at()` syntax.
Added a test to check the syntax works.


localstack-core/localstack/services/scheduler/provider.py
```diff
@@ -10,7 +10,9 @@
 
 LOG = logging.getLogger(__name__)
 
-AT_REGEX = r"^at[(](0[1-9]|1\d|2[0-8]|29(?=-\d\d-(?!1[01345789]00|2[1235679]00)\d\d(?:[02468][048]|[13579][26]))|30(?!-02)|31(?=-0[13578]|-1[02]))-(0[1-9]|1[0-2])-([12]\d{3}) ([01]\d|2[0-3]):([0-5]\d):([0-5]\d)[)]$"
+AT_REGEX = (
+    r"^at[(](19|20)\d\d-(0[1-9]|1[012])-([012]\d|3[01])T([01]\d|2[0-3]):([0-5]\d):([0-5]\d)[)]$"
+)
```
bentsku, line 15: question: how was the previous regex rule created, and how does the new one differ? How are we sure this is the right format, is there documentation online? Sorry, I don't know Scheduler so well and I wondered if there was any documentation or anything that helped you create this regex. Thanks! 🙏

bentsku (approved):
LGTM! I just have a small question, but this good to be merged to unblock the users! 🚀 thanks for jumping on this 🙏

localstack-core/localstack/services/scheduler/provider.py
anisaoshafi: We didn't have any test for what a correct schedule syntax with `at` syntax should look like.
Before, anything was allowed in scheduler because moto wasn't validating it.
I don't properly remember where I copied it from, sorry (stackoverflow or else), but apparently I didn't test the condition for at, I just tested the invalid ones, but not a valid one, and that ended up blocking a user. 

I should have added this test `tests_create_schedule_with_valid_schedule_expression` since that PR that introduced the validation for scheduler, but I struggled making a it work against aws.
This time I took the time to fix it, and turns out the issue I was getting was because the IAM role takes some time to propagate and because of that I was getting some permission issue for the scheduler. So we just have to wait some seconds for `is_aws_cloud`.

anisaoshafi: This is the aws [docs](https://docs.aws.amazon.com/scheduler/latest/UserGuide/schedule-types.html#one-time) for this.

PR #12221, opened 2025-02-03
restructure unit tests

bentsku:
## Motivation
In order to be easier to use test selection against our unit tests, and to fit a common format with our integration tests, this PR refactor our unit tests structure to both fit the rules, and be clearer. The AWS services unit tests were mixed up with common utils tests, docker client tests and such.

This way, we don't need to worry about file naming, and for new tests being added, it will be clearer which structure to follow in order to properly trigger test selection.

We're going from `tests/unit/test_<service>.py` to `tests/unit/services/<service>/test_<whatever>.py`

The `CODEOWNERS` have been manually updated, but our current generator is capable of handling the new format already and was already handling it, so it won't trigger issue. 

## Changes
- restructure the unit tests to follow the same format as our integration tests
- update the CODEOWNERS file manually to avoid pipeline failure, I'll run the generation manually to verify I haven't missed anything


tiurin (approved):
Nice improvement for tests maintainability @bentsku! Especially that some of the tests have already been under `services`. It's great that now we follow the same convention both inside unit tests and between unit and integration ones! 👏

alexrashed (approved):
Nice and clean change! It applies the same structure as we already know from the integration tests, and it also works with the current `CODEOWNERS` automation. :)

PR #12226, opened 2025-02-04
No sve dynamodb m4

tiurin:
## Motivation

Fixes https://github.com/localstack/localstack/issues/12054.

[Previous fix](https://github.com/localstack/localstack/pull/12112) was using a java command to check available VM options, and this command was crashing for exactly the same reason as the DynamoDB process.

## Changes

Added `-XX:UseSVE=0` to java VM options check and limited the check to arm64 architecture (otherwise check can't be run reliably - with option it will crash on arm64 and without option it will crash on m4 macs)

## Testing

Tested on M4 instance.

### DIFF
--- a/localstack-core/localstack/services/dynamodb/server.py
+++ b/localstack-core/localstack/services/dynamodb/server.py
@@ -14,6 +14,7 @@ from localstack.utils.common import TMP_THREADS, ShellCommandThread, get_free_tc
 from localstack.utils.functions import run_safe
 from localstack.utils.net import wait_for_port_closed
 from localstack.utils.objects import singleton_factory
+from localstack.utils.platform import Arch, get_arch
 from localstack.utils.run import FuncThread, run
 from localstack.utils.serving import Server
 from localstack.utils.sync import retry, synchronized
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/dynamodb/server.py
```diff
@@ -153,14 +154,21 @@ def _get_java_vm_options(self) -> list[str]:
...
+                        "-XX:UseSVE=0",
+                        "-XX:+UnlockDiagnosticVMOptions",
+                        "-XX:+PrintFlagsFinal",
+                        "-version",
+                    ],
+                    env_vars=dynamodblocal_installer.get_java_env_vars(),
+                    print_error=True,
+                )
+                # Check if Scalable Vector Extensions are support on this JVM and CPU. If so, disable it
+                if "UseSVE" in stdout:
```
viren-nadkarni, line 170: The intention of this check was to determine if `UseSVE` is a supported VM flag because passing unsupported VM flags makes JRE exit immediately.

Now, by passing this flag in what was intended to be a check renders it moot :slightly_smiling_face: So I think this whole block can be removed and the flag returned directly.

Please also verify that this doesn't cause issues in CPUs older than M4

tiurin: yeah, this is the only check needed indeed, will update 👍

tiurin: Also tested on M3 instance.

alexrashed (approved):
Awesome! Thanks a lot for digging into this, and for testing this in specifically provisioned M4 cloud instances! 💯 🦸🏽 
Thanks a lot also to @viren-nadkarni for the great upfront analysis and the first blind fix! It's a pity that this command also just fails with the same error! 😅 
The pipeline with the ARM64 integration tests also looks 💚 🥳

PR #12228, opened 2025-02-04
[SQS] Cancel pending and future tasks on shutdown

gregfurman:
## Motivation
Under high volumes of SQS receive/send/delete calls, the internal SQS `ThreadPoolExecutors` can execute tasks long after a service has completed their shutdown/stop lifecycle hooks.

That is, when we try stop the SQS service, **currently running** and **future pending** tasks are not cleared from the executor pool -- potentially being executed long after the service signalled to have stopped.

This PR ensures that all **enqueued** tasks are cancelled on shutdown.

## Changes
- Changed the `CloudWatchDispatcher.shutdown` and `MessageMoveTaskManager.close` methods to cancel all pending/future tasks that are currently enqueued for execution.

### DIFF
--- a/localstack-core/localstack/services/sqs/provider.py
+++ b/localstack-core/localstack/services/sqs/provider.py
@@ -77,7 +77,9 @@ from localstack.http import Request, route
 from localstack.services.edge import ROUTER
 from localstack.services.plugins import ServiceLifecycleHook
 from localstack.services.sqs import constants as sqs_constants
-from localstack.services.sqs.constants import HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT
+from localstack.services.sqs.constants import (
+    HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT,
+)
 from localstack.services.sqs.exceptions import InvalidParameterValueException
 from localstack.services.sqs.models import (
     FifoQueue,
@@ -192,7 +194,7 @@ class CloudwatchDispatcher:
         )
 
     def shutdown(self):
-        self.executor.shutdown(wait=False)
+        self.executor.shutdown(wait=True, cancel_futures=True)
 
     def dispatch_sqs_metric(
         self,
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/sqs/queue.py
```diff
@@ -12,6 +12,8 @@ def __init__(self, maxsize=0):
 
     def get(self, block=True, timeout=None):
         with self.not_empty:
+            if self.is_shutdown:
+                raise Empty
```
dfangl, line 16: Is there a specific reason why we need this at the beginning here as well?

localstack-core/localstack/services/sqs/provider.py
```diff
@@ -192,7 +194,7 @@ def __init__(self, num_thread: int = 3):
         )
 
     def shutdown(self):
-        self.executor.shutdown(wait=False)
+        self.executor.shutdown(wait=True, cancel_futures=True)
```
dfangl, line 197: Did you check if we get any issues when doing a wait=True here? I am fine with cancelling the futures, but this could block the rest of the shutdown quite early (depending on the shutdown order), in contrast to at the very end of the interpreter shutdown.

localstack-core/localstack/services/sqs/queue.py
gregfurman: Just allows for an earlier exit if we enter the shutdown loop in the time it takes between `shutdown()` being called and `get()` being called. Just a very minor optimisation

localstack-core/localstack/services/sqs/provider.py
gregfurman: I noticed these dispatched jobs were running while the SQS provider was already supposed to have shut down + while others were in the process of shutting down. 

The `wait=True` was to prevent this behaviour, meaning fewer exceptions being raised and no botocore retrying happening simultaneously to LS exiting.

If we're OK with the dispatched jobs running while other services are shutting down then then happy to make this false again.

localstack-core/localstack/services/sqs/queue.py
dfangl: We are for example also not raising exceptions on timeouts < 0 in this case here, but that might not be all too important.

localstack-core/localstack/services/sqs/provider.py
dfangl: It is fine, I am just worried about it blocking too long without a timeout after which we call it quits.

dfangl (approved):
I think this is innocent enough, LGTM!

PR #12229, opened 2025-02-05
Fix kubernetes dev run script

dfangl:
## Motivation
With recent refactorings to namespace packaging (#11190 ) and usage of setuptools_scm (#11355) the kubernetes dev run script broke.
It still mounted pro source code into `site-packages/localstack-ext` and not inside the `localstack` namespace package, and the entrypoint mounts did not take into account the new versioning scheme.

A main culprit was the implicit behavior of taking parts of the path from the host and using them inside the container, as this prevented mounting the pro source code in the `localstack/pro` directory.
Changing this required some refactoring.

## Changes
* The mount points are now generated in a separate method, and into a defined data structure. This provides more flexibility, as previous behavior was very implicit (taking the last part of the path on the host for mounting into the container for example). The config generation methods use the data and convert it into the respective format, without additional logic.
* Add `--mount-entrypoints` flag, to manually opt in to mounting the entrypoints. I did not yet implement a version detection to mount the entrypoints into the right `.egg-info` directory, so this has to be manually adjusted currently for the versions inside the docker image, if used.
* Add `CONTAINER_RUNTIME=kubernetes` and `lambda.executor=kubernetes` config to the value overrides if pro is enabled
* Fix some errors when mounting code into the community package (as the package installation of `localstack-core` differs in community from pro)

## Testing
Run `python -m localstack.dev.kubernetes --pro --write --output-dir=<config-dir>` and follow the instructions.

### DIFF
--- a/localstack-core/localstack/dev/kubernetes/__main__.py
+++ b/localstack-core/localstack/dev/kubernetes/__main__.py
@@ -70,6 +70,10 @@ def generate_mount_points(
         if pro:
             # Community entrypoints in pro image
             # TODO actual package version detection
+            print(
+                "WARNING: Package version detection is not implemented."
+                "You need to adapt the version in the .egg-info paths to match the package version installed in the used localstack-pro image."
+            )
             community_version = "4.1.1.dev14"
             pro_version = "4.1.1.dev16"
             egg_path = os.path.join(
--- a/localstack-core/localstack/dev/kubernetes/__main__.py
+++ b/localstack-core/localstack/dev/kubernetes/__main__.py
@@ -6,7 +6,9 @@ import yaml
 from localstack import version as localstack_version
 
 
-def generate_k8s_cluster_config(pro: bool = False, mount_moto: bool = False, port: int = 4566):
+def generate_k8s_cluster_config(
+    pro: bool = False, mount_moto: bool = False, port: int = 4566, mount_entrypoints: bool = False
+):
     volumes = []
     root_path = os.path.join(os.path.dirname(__file__), "..", "..", "..", "..")
     localstack_code_path = os.path.join(root_path, "localstack-core", "localstack")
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/plugins.py
```diff
@@ -19,6 +19,7 @@ def deprecation_warnings() -> None:
     LOG.debug("Checking for the usage of deprecated community features and configs...")
     from localstack.deprecations import log_deprecation_warnings
 
+    LOG.warning("Community change")
```
simonrw, line 22: q: what is this line doing?

localstack-core/localstack/dev/kubernetes/__main__.py
```diff
@@ -1,56 +1,148 @@
...
+            MountPoint(
+                name="localstack-pro",
+                host_path=os.path.normpath(pro_code_path),
+                node_path="/code/localstack-pro",
+                container_path=os.path.join(venv_path, "localstack", "pro", "core"),
+            )
         )
 
+    # entrypoints
+    if mount_entrypoints:
```
simonrw, line 69: Can we log a warning that this is currently not _really_ implemented, and may provide confusing results if the user expects this to work

simonrw: Also, note for the future: we implement this in `localstack.dev.run` so we can take the implementation from there.

See [`localstack.dev.run.configurators._list_files_in_container_image`](https://github.com/localstack/localstack/blob/ed8046a0d95762dceb94ac013671c1527d082afc/localstack-core/localstack/dev/run/configurators.py#L342)

simonrw (approved):
This change looks great, thanks!

localstack-core/localstack/plugins.py
dfangl: Testing the code mounting, that should not have made it into the PR, removing it :)

PR #12230, opened 2025-02-05
Add structured metrics instrumentation

vittoriopolverino:
## Motivation
The existing Counter implementation follows a parsing-based approach, relying on metric name patterns like label1:label2:* and inferring implicit relationships through JSON structures, dictionaries, or other nested data formats. Since these structures can vary depending on the feature or how each feature owner decides to implement them, querying, filtering, and aggregating metrics become inconsistent and complex. This lack of explicit structure makes it harder to standardize analysis across different use cases.

This refactor introduces a structured and extensible approach to metric collection, enforcing clear label standards and dimensional analysis while making all relationships and classifications explicit, rather than relying on naming conventions, positional placement, or nested JSON structures. By ensuring consistency, we can simplify querying, filtering, and aggregations, making analytics more scalable and reducing the need for custom metric handling.

## Changes
- Moved metrics to `metrics.py`, making them a standalone concept, not strictly tied to `usage.py`.
- Introduced `MetricRegistry` as a singleton for managing all registered metrics (Counters, Gauges, etc.).
- Introduced `Metric` as the base interface for metrics, enforcing the implementation of the `collect()` method in all metric types (e.g., Counters, Gauges).
- Updated published event name from `ls:usage` to `ls_metrics`
- Set a limit of 8 labels per metric to ensure labels are used effectively as dimensions for time-series analysis, filtering, and aggregations, preventing misuse or unnecessary complexity. A dedicated guide will follow to outline best practices.

### Usage Examples
**Basic Counter**
A Counter tracks the occurrences of an event. It can be used without labels for simple counting:
```
from localstack.utils.analytics.metrics import Counter

# Define a counter
http_responses = Counter(namespace="http", name="requests")

# Increment
http_responses.increment()
http_responses.increment(value=3)

# Reset
http_responses.reset()
```

**Counters with Labels**
Counters can include labels to categorize events more effectively:
```
from localstack.utils.analytics.metrics import Counter

# Define a counter with labels
chaos_invocations = Counter(namespace="chaos", name="invocations", labels=["operation"])

# Increment
chaos_invocations.labels(operation="fault").increment(value=3)
chaos_invocations.labels(operation="network-effect").increment(value=4)

# Reset
chaos_invocations.labels(operation="fault").reset()
chaos_invocations.labels(operation="network-effect").reset()
```

### Unit Tests (No External Services Required)  
- The test suite includes **unit tests** that validate metric collection and registry behavior **without requiring event publishing** to the analytics backend.  

### End-to-End Testing with the Analytics Backend
For a more in-depth test, follow these steps:

1. Run the Analytics Backend Locally
- Start the analytics backend to capture and process metric events.
- Ensure that environment variables are correctly set to send events to the data platform.

2. Start LocalStack Core with Analytics Enabled
- Start LocalStack Core with the appropriate environment variables to enable analytics and send events to the local backend:

```bash
ANALYTICS_API="http://localhost:8000/v1" \
DEBUG=1 \
DEBUG_ANALYTICS=1 \
python -m localstack.cli.main start --host
```

## TODO

What's left to do:

- [X] Instrument Chaos API - [PR](https://github.com/localstack/localstack-ext/pull/4076).
- [X] Complete internal documentation on best practices for metrics and label naming


vittoriopolverino:

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+                        "name": self._full_name,
+                        "value": value,
+                        # Example: If labels=["service", "status"], and values=("sqs", "error"),
+                        # it would generate: {"label_1": "sqs", "label_2": "error"}
+                        **label_dict,
+                        # Example: If labels=["service", "status"], it would generate:
+                        # {"label_1_des": "service", "label_2_des": "status"}
+                        **label_descriptions,
+                    }
+                )
```
vittoriopolverino, line 302: Currently, the namespace field is used to automatically prefix all metric names under that namespace. For example, with `Counter(namespace="chaos", name="invocations")`, the resulting metric name becomes `chaos_invocations`.

I'm considering adding namespace as a dedicated field in the payload to simplify filtering and high-level aggregations, instead of relying solely on the prefixed metric name.

Right now, filtering for broader categories (e.g., all chaos metrics) requires parsing the metric name. By making namespace explicit:

- It acts as an additional dimension for analysis.
- Filtering and grouping at a higher level (e.g., all chaos_api metrics) becomes easier.
- We avoid using one of the 5 available labels just to define a category.

```json
"metrics": [
    {
        "namespace": "chaos",
        "name": "chaos_invocations",
        "value": 20,
        "label_1": "fault",
        "label_2": "success",
        "label_1_des": "operation",
        "label_2_des": "status"
    }
]
```

@ackdav WDYT?

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+
+            return collected_data
+
+    def get_lock(self) -> threading.Lock:
+        """Provides access to the internal lock."""
+        return self._lock
+
+
+class LabeledCounter:
```
alexrashed, line 311: **suggestion:** This is a pretty interesting hierarchy model here. You have a `Counter` which has a multi-label data structure internally, which can handle none to multiple labels (all labels share a common lock). And you have the `LabeledCounter` here, which provides functionality to only work with a single label.

I would like to propose a different composition:
- A `Counter` only holds a single label.
- There is a `MultiLabelCounter` (<- replace this with a better name 😄) which uses the composite pattern to manage multiple `Counter` instances (one for each `label`). This class also implements `Metric`.

This way, your usecase would still be enabled, but the complexity of the `Counter` would be reduced and locks would only be used for a single label.
What do you think?

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+            ValueError: If the increment value is not positive.
+        """
+        self._counter.inc(label_key=self._labels, value=value)
+
+    def reset(self) -> None:
+        """Resets the labeled counter to zero."""
+        self._counter.reset(label_key=self._labels)
+
+
+@hooks.on_infra_start()
```
alexrashed, line 361: **suggestion:** ‏Awesome! This already integrates the new metrics module into the LocalStack emulator lifecycle!
However, it's not used yet. I think it would be great to actually see this module in action. This PR doesn't have to migrate all callers of the `usage.py` module, but it would be great to migrate some examples directly.
And if we plan to do a gradual migration to this new module, we should at least add a comment in the `usage.py` module which marks it as deprecated.

vittoriopolverino: > And if we plan to do a gradual migration to this new module, we should at least add a comment in the usage.py module which marks it as deprecated.

What’s the usual approach for marking modules or functions as deprecated?

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+from localstack.runtime import hooks
+from localstack.utils.analytics import get_session_id
+from localstack.utils.analytics.events import Event, EventMetadata
+from localstack.utils.analytics.publisher import AnalyticsClientPublisher
+
+LOG = logging.getLogger(__name__)
+
+# TODO: introduce some base abstraction for the counters after gather some initial experience working with it
+#  we could probably do intermediate aggregations over time to avoid unbounded counters for very long LS sessions
+#  for now, we can recommend to use config.DISABLE_EVENTS=
```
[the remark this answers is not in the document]
vittoriopolverino: You can ignore the comment, it’s a leftover from `usage.py`, which I’ll be removing soon

localstack-core/localstack/utils/analytics/metrics.py
alexrashed: We don't really have many examples of deprecations of internal APIs. You can just clearly mention in the docstrings of the components you are deprecating.

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+    _namespace: Optional[str]
+    _labels: Optional[List[str]]
+    __labels_normalization: Dict[str, str]
+
+    def __new__(
+        cls, name: str = "", labels: Optional[List[str]] = None, namespace: str = ""
+    ) -> Union[Counter, MockCounter]:
+        """Returns a real or mock instance based on the `DISABLE_EVENTS` config."""
+        if config.DISABLE_EVENTS:
+            return MockCounter()
```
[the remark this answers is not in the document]
vittoriopolverino: The idea is to keep methods clean without being cluttered with conditional checks and to prevent the instantiation of unnecessary data structures. To prevent interface divergence, I made `MockCounter` a child class of `BaseCounter`, ensuring that it follows the same interface and avoids potential inconsistencies.

Additionally, by design, the Counter will not have any additional methods beyond those it currently implements.

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+        """
+        Initializes a counter.
+
+        Args:
+            name (str): The metric name.
+            labels (Optional[List[str]]): List of labels (max 5). If not provided, the counter is unlabeled.
+            namespace (str): The namespace for the metric.
+        """
+        self._lock = threading.Lock()
+        self._values = defaultdict(int)
```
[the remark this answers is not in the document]
vittoriopolverino: I'm not really an expert in threading, but I initially avoided `itertools.count `because it's not thread-safe by definition. However, in CPython, it should still be safe to use because the GIL ensures that operations on itertools.count are effectively atomic, preventing race conditions in most cases. right? 🤔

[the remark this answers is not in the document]
alexrashed: Good point! Feel free to go either way, `itertools.count` was used before (and should do the job), but a custom lock handling would even be post-GIL safe (sounds to me like post-quantum security 😛).

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+    A thread-safe counter for tracking occurrences of an event.
+
+    Supports both:
+    - **Labeled counters** (via `.labels()`).
+    - **Unlabeled counters** (direct `.inc()` calls).
+
+    Attributes:
+        _lock (threading.Lock): Ensures thread-safe operations.
+        _values (defaultdict): Stores counter values, keyed by label tuples.
+        _name (Optional[str]): The metric name.
```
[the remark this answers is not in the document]
vittoriopolverino: The namespace is an optional prefix prepended to the metric name to provide logical grouping and if set, the full metric name follows the format `namespace_name`.

I've added this information [here](https://github.com/localstack/localstack/pull/12230/files#diff-c473699486ab7f1e12ed3fda938ba0f5961c88592dc7aecb32136559809292f4R158) and it will also be documented on Notion

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+            ValueError: If incrementing a labeled counter without labels.
+        """
+        if value <= 0:
+            raise ValueError("Increment value must be positive.")
+
+        if self._labels and label_key is None:
+            raise ValueError("This counter requires labels, use .labels() instead.")
+
+        # Use an empty tuple for non-labeled counters
+        key = label_key if label_key is not None else ()
```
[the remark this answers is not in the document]
vittoriopolverino: I'm afraid it's not possible to remove some of these checks in the `increment()` or `labels()` method when using a multi-label counter, as doing so could lead to one of the issues showed in these [tests](https://github.com/localstack/localstack/pull/12230/files#diff-f22bd96399693c592dfddc3b965c41f4ca0b9295696eec0458c33ed0985a14efR117-R151) 

I'm open to suggestions if you have any ideas on how to make this more efficient while keeping it correct! 🙏🏼 💪🏼

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+        # Increment the labeled counter
+        label_delegate.inc(5)  # Internally calls counter.inc(label_key=("GET",), value=5)
+
+        # Reset the labeled counter
+        label_delegate.reset()  # Internally calls counter.reset(label_key=("GET",))
+        ```
+
+    Attributes:
+        _counter (Counter): Reference to the main counter.
+        _labels (Tuple[Optional[str], ...]): The assigned label values.
```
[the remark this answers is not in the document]
vittoriopolverino: I made some changes and updated the naming of the data structures. 

For example:
```python
chaos_invocations = Counter(namespace="chaos", name="invocations", labels=["operation"])
chaos_invocations.labels(operation="fault").increment(value=3)
chaos_invocations.labels(operation="network-effect").increment(value=4)
```


`_labeled_counts` →  {('fault',): 3, ('network-effect',): 4})
`_client_defined_labels` → ("fault", ) or ("network-effect", )
`_fixed_labels` → ["label_1"]
`_client_defined_to_fixed_label_mapping` → {"operation": "label_1"}

My goal here is to want to keep label names fixed/static while letting users set their own meaning for each one. This keeps the schema consistent but still gives flexibility in naming.

When sending events to analytics, the payload follows this structure:

```json
{
  "events": [
    {
      "name": "ls_core",
      "metadata": {
        "session_id": "****",
        "client_time": "****"
      },
      "payload": {
        "metrics": [
          {
            "namespace": "chaos",
            "name": "chaos_invocations",
            "value": 3,
            "label_1": "fault",
            "label_1_des": "operation",
          },
          {
            "namespace": "chaos",
            "name": "chaos_invocations",
            "value": 4,
            "label_1": "network-effect",
            "label_1_des": "operation",
          }
        ]
      }
    }
  ]
}

```

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: You're right, I didn’t like the previous hierarchical structure either. Initially, I wanted to avoid adding complexity, but the new model makes more sense. Now there's:

Superclass: Metric
Base Class: BaseCounter(Metric)
Child Classes: MockCounter(BaseCounter), UnlabeledCounter(BaseCounter), MultiLabelCounter(BaseCounter)
I didn’t use a classic factory pattern because I didn’t want Counter to be instantiated with something like `CounterFactory.create()`. Instead, I opted for a factory-like [constructor](https://github.com/localstack/localstack/pull/12230/files#diff-c473699486ab7f1e12ed3fda938ba0f5961c88592dc7aecb32136559809292f4R375-R383), using __new__ to transparently return the correct counter type. WDYT?

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,424 @@
...
+            self._count = 0
+
+    def collect(self) -> List[Dict[str, Union[str, int]]]:
+        with self._mutex:
+            if self._count == 0:  # Exclude zero-value metrics
+                return []
+            return [{"namespace": self._namespace, "name": self._full_name, "value": self._count}]
+
+
+class MultiLabelCounter(BaseCounter):
```
alexrashed, line 263: **thought:** I might just not see it, but I think the patterns here do over-complicate the solution, while not really bringing the intended benefit.
- The `Counter.__new__` dunder implementation is a nice way to easily switch between implementations, but then the interface between them is somewhat different (as `MultiLabelCounter` is basically a composite but for most cases you need to define the specific component of the composite to use). I also think that this has a negative influence on typing integrations in IDEs.
- The `MultiLabelCounter` as its own implementation instead of just reusing multiple `UnlabeledCounter` instances broadens the locking across all labels.‏

An existing migration would be great to see how this framework is intended to be used, and if I am missing something here!

tests/unit/utils/analytics/conftest.py
```diff
@@ -1,9 +1,38 @@
...
+
+
+@pytest.fixture
+def counter() -> UnlabeledCounter:
+    return Counter(name="test_counter")
+
+
+@pytest.fixture
+def multilabel_counter() -> MultiLabelCounter:
+    return Counter(name="test_multilabel_counter", labels=["status"])
```
alexrashed, line 38: **note / suggestion (non-blocking):** ‏Coming back to the typing comments above: I think if you remove these fixtures, and directly use the Counter in your tests, you would see in your IDE that the types cannot be properly resolved. I would suggest you not to use fixtures in the tests for this case, because in your TDD process you really also want to get a feeling for the API and the usage of the module you are building

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def __new__(cls) -> MetricRegistry:
+        """
+        Ensures that only one instance of `MetricRegistry` exists.
+
+        :return: A instance of `MetricRegistry`.
+        """
+        if not cls._instance:
+            cls._instance = super(MetricRegistry, cls).__new__(cls)
+            cls._instance._registry = dict()  # Registry initialized here
```
alexrashed, line 42: **nitpick (non-blocking):** I think you might be mixing up the responsibilities of the `__new__` and the `__init__` dunder methods here. In my understanding, `__new__` is responsible of how the instance is created, while `__init__` is responsible of the initialization.‏

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,424 @@
...
+        return {
+            "metrics": [
+                metric
+                for metric_instance in self._registry.values()
+                for metric in metric_instance.collect()
+            ]
+        }
+
+
+def get_metric_registry() -> MetricRegistry:
```
alexrashed, line 78: **question:** What's the difference / benefits / downsides between this function,  and the singleton implementation in the `__new__` dunder, and a plain global variable?
Generally speaking, we usually go the pragmatic way of just declaring a global and reusing it (modules are only loaded once).
If this function is the only way the metrics registry should be retrieved, why not just setting a global with a single instance and using it everywhere instead? (This would allow you to remove this function and the `__new__` dunder).‏

alexrashed (commented):
Thanks a lot for acting on the comments from the previous reviews! A lot of the comments are resolved. I don't really see the necessity of the complex structure between the different `Counter` classes, but that shouldn't be a blocker.
I also added some more comments to discuss.

I feel like a lot of the complexity is coming from the need to improve the developer experience when using the new module, but then there are no real-world usages yet.
I think it is crucial to see this in action by migrating existing callers of the `usage.py` module. This will make it way more clearer how this framework performs in terms of developer experience. 😄

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: I’ve removed the method, and now the singleton is directly initialized and used within Counter instances or inside publish_metrics

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,424 @@
...
+            )
+
+        return collected_metrics
+
+    def collect(self) -> List[Dict[str, Union[str, int]]]:
+        with self._mutex:
+            return self.as_dict_list()
+
+
+class Counter:
```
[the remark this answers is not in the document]
vittoriopolverino: I agree that using a composite pattern could be a good approach to explore. In the meantime, I've leveraged overloading to enforce static type safety. 

it's now not possible for a simple counter to call .labels(), and a labeled counter cannot call .increment() without first specifying a valid label combination via .labels()

Thanks for bringing up this issue!

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: > An existing migration would be great to see how this framework is intended to be used, and if I am missing something here!

[PR]( https://github.com/localstack/localstack-ext/pull/4076)

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,318 @@
...
+
+    _namespace: Optional[str]
+    _name: str
+    _type: str
+
+    def __init__(self, name: str, namespace: Optional[str] = ""):
+        super(_CounterMetric, self).__init__()
+        if not name:
+            raise ValueError("Name is required and cannot be empty.")
+
```
alexrashed, line 160: **nitpick (non-blocking):** This here isn't really necessary, as the parameter is not optional.‏

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,318 @@
...
+    _namespace: Optional[str]
+    _name: str
+    _type: str
+    _unit: str
+    _labels: list[str]
+    _label_values: tuple[Optional[str], ...]
+    _counters_by_label_values: defaultdict[Tuple[str, ...], _Counter]
+
+    def __init__(self, name: str, labels: List[str] = list, namespace: Optional[str] = ""):
```
alexrashed, line 197: **nitpick (non-blocking):** I think the labels here shouldn't be optional, right? They need to be set, and they should not be empty.‏

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -0,0 +1,318 @@
...
+        if not name:
+            raise ValueError("Name is required and cannot be empty.")
```
alexrashed, line 199: **nitpick (non-blocking):** ‏I think this check here is not necessary, as the name is not optional. It might make sense to check if the list of labels is empty though?

alexrashed (approved):
I just added a few nitpick comments here, but from my perspective we are ready to roll!
Given that this is targeted for 4.3, and you might want to iterate over it one more time, we should be ready to merge this after the 4.2 release tomorrow! 🚀

vittoriopolverino:
FYI: I'm fixing unit-tests right now!

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: I'd still like to enforce that the metric name isn't empty to prevent `Counter(name="")`. 
Since _CounterMetric inherits from Metric, name validation is already enforced in the @name.setter. I should use this attribute instead of declaring a separate one in CounterMetric

```
@name.setter
    def name(self, value: str) -> None:
        """
        Validates and sets the full metric name.

        :raises ValueError: If the name is empty or invalid.
        """
        if not value or value.strip() == "":
            raise ValueError("Metric must have a valid name.")
        self._name = value
```

alexrashed: I think that is a very good point. But a setter here might not be the best option here, as you can't enforce that it is actually being set, you can only validate _when_ it is set.
So maybe this setter should be replaced with an init which enforces that name is set and not empty?
I think that would solve your problem here, since subclasses should actually always call the super init, which would then also enforce this in this class.

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: > they should not be empty.

Great observation! I noticed and issue here: the [condition](https://github.com/localstack/localstack/pull/12230/files#diff-c473699486ab7f1e12ed3fda938ba0f5961c88592dc7aecb32136559809292f4R290) `if not labels`: was causing _CounterMetric to be instantiated instead of _LabeledCounterMetric. I’ve updated the condition to always create _LabeledCounterMetric when labels are provided, even if empty, and added a validation in the constructor to enforce that the labels list cannot be empty

PR #12232, opened 2025-02-06
Core: Add py.typed, Remove docs from source dist

bblommers:
## Motivation
In order to actually use the type-hints, we need to add a `py.typed`-file - see [the mypy-docs](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages) for more info. We don't have to explicitly configure anything for this file to be part of both the wheel and source dist - we're already including almost all files by default.

Related: while building the source dist I also realized that it currently contains a lot of files/folders that we (IMHO) should not include, like the docs and CI workflows. Excluding these reduces the source dist size from 4.3MB to 3.4MB.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,10 @@
+exclude .github/**
+exclude .circleci/**
+exclude docs/**
 exclude tests/**
 exclude .test_durations
+exclude .gitignore
+exclude .pre-commit-config.yaml
+exclude .python-version
 include Makefile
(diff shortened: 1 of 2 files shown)


bblommers:
I've noticed that our CI pipeline/Makefile does currently not run MyPy - so all types is basically untested. :cry: 

I'd be happy to take a (time-boxed) stab at this as well, if necessary.

PR #12233, opened 2025-02-06
[ESM] Dynamically set SQS message override count

gregfurman:
## Motivation
Currently, event hooks used to override the number of SQS messages to delete/receive use static parameters from the `SqsPoller` class to determine override values. Since boto clients in the same session can be re-used, this approach could possibly cause data races if multiple clients register against the same event source with different `BatchSize` parameters.

Instead, this PR introduces event hooks that allow the override parameter to be dynamically passed in to each boto call (and injected into the headers) instead of trying to retrieve this information from the static ESM values.

## Changes
- Remove header injection implementation that relied on retrieving static values from the `SqsPoller.sqs_queue_parameters`.
- Introduce a `sqs_override_max_message_count` parameter to `ReceiveMessage` and `DeleteMessageBatch` requests to allow overriding message count on a per-boto-request basis.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
@@ -76,7 +76,9 @@ class SqsPoller(Poller):
         event_system.register(
             "provide-client-params.sqs.DeleteMessageBatch", handle_message_count_override
         )
-        event_system.register("before-call.sqs.*", handle_inject_headers)
+
+        event_system.register("before-call.sqs.ReceiveMessage", handle_inject_headers)
+        event_system.register("before-call.sqs.DeleteMessageBatch", handle_inject_headers)
 
     def get_queue_attributes(self) -> dict:
         """The API call to sqs:GetQueueAttributes is required for IAM policy streamsing."""
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
@@ -16,7 +16,6 @@ from localstack.services.lambda_.event_source_mapping.pollers.poller import (
     Poller,
     parse_batch_item_failures,
 )
-from localstack.services.lambda_.event_source_mapping.senders.sender_utils import batched
 from localstack.services.sqs.constants import HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT
 from localstack.utils.aws.arns import parse_arn
 from localstack.utils.strings import first_char_to_lower
@@ -29,6 +28,9 @@ DEFAULT_MAX_RECEIVE_COUNT = 10
 class SqsPoller(Poller):
     queue_url: str
 
+    batch_size: int
+    maximum_batching_window: int
+
     def __init__(
         self,
         source_arn: str,
@@ -38,6 +40,10 @@ class SqsPoller(Poller):
     ):
         super().__init__(source_arn, source_parameters, source_client, processor)
         self.queue_url = get_queue_url(self.source_arn)
+
+        self.batch_size = self.sqs_queue_parameters["BatchSize"]
+        self.maximum_batching_window = self.sqs_queue_parameters["MaximumBatchingWindowInSeconds"]
+
         self._register_client_hooks()
 
     @property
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -103,11 +98,11 @@ def poll_events(self) -> None:
...
         response = self.source_client.receive_message(
             QueueUrl=self.queue_url,
-            MaxNumberOfMessages=self.sqs_queue_parameters.get(
-                "BatchSize", DEFAULT_MAX_RECEIVE_COUNT
-            ),
+            MaxNumberOfMessages=min(self.batch_size, DEFAULT_MAX_RECEIVE_COUNT),
             MessageAttributeNames=["All"],
             MessageSystemAttributeNames=[MessageSystemAttributeName.All],
+            # Override how many messages we can receive per call
+            sqs_override_max_message_count=self.batch_size,
```
joe4dev, line 105: Nice 💯 

I love that we are explicit about this override on every occasion. Please highlight that pattern at an upcoming demo 🎥  and include an example in the demo notes 🙏 . That's super helpful for the wider core engineering org 🙇

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -103,11 +98,11 @@ def poll_events(self) -> None:
         #  https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-sqs.html#pipes-sqs-scaling
         response = self.source_client.receive_message(
             QueueUrl=self.queue_url,
-            MaxNumberOfMessages=self.sqs_queue_parameters.get(
-                "BatchSize", DEFAULT_MAX_RECEIVE_COUNT
-            ),
+            MaxNumberOfMessages=min(self.batch_size, DEFAULT_MAX_RECEIVE_COUNT),
```
joe4dev, line 101: Shouldn't that be `max` instead of `min` because `MaxNumberOfMessages` has a hard limit against AWS?

Any batch size >10 would currently break if someone uses this poller against a real AWS SQS queue, right?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -52,36 +58,25 @@ def is_fifo_queue(self) -> bool:
...
-
-            # Allow overide parameter to be greater than default and less than maximum batch size.
-            # Useful for getting remaining records less than the batch size. i.e we need 100 records but BatchSize is 1k.
-            override = min(requested_count, self.sqs_queue_parameters["BatchSize"])
-            context[HEADER_LOCALSTACK_SQS_OVERRIDE_MESSAGE_COUNT] = str(override)
-
-        def _handle_delete_batch_override(params, context, **kwargs):
-            requested_count = len(params.get("Entries", []))
+        def handle_message_count_override(params, context, **kwargs):
+            requested_count = params.pop("sqs_override_max_message_count", None)
```
joe4dev, line 62: That `pop` was previously missing to tidy up the params before making the boto3 calls 👍

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -52,36 +58,25 @@ def is_fifo_queue(self) -> bool:
...
-            "provide-client-params.sqs.ReceiveMessage", _handle_receive_message_override
+            "provide-client-params.sqs.ReceiveMessage", handle_message_count_override
         )
         # Since we delete SQS messages after processing, this allows us to remove up to 10K entries at a time.
         event_system.register(
-            "provide-client-params.sqs.DeleteMessageBatch", _handle_delete_batch_override
+            "provide-client-params.sqs.DeleteMessageBatch", handle_message_count_override
         )
-        event_system.register("before-call.sqs.*", _handler_inject_header)
+        event_system.register("before-call.sqs.*", handle_inject_headers)
```
joe4dev, line 79: nit: I think `*` is fine here and not worth changing. Just curious whether we can/should limit certain operations (e.g., `before-call.sqs.ReceiveMessage|DeleteMessageBatch`) as we adopt this pattern in other places.

joe4dev (approved):
Love the explicit clarity around overwrites here 💯 

My main suggestion is to double-check the `min` / `max` condition for `MaxNumberOfMessages` to enable our SQS poller to work against real AWS SQS queues.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
gregfurman: These changes effectively mean that the value of `MaxNumberOfMessages` would not be used (against LocalStack) while still preserving the `MaxNumberOfMessages` parameter incase we're polling against AWS.

> Any batch size >10 would currently break if someone uses this poller against a real AWS SQS queue, right?

The `min` ensures that that we wont encounter an issue since `MaxNumberOfMessages` will always be between 1-10:
- `BatchSize=10 => MaxNumberOfMessages=10`
- `BatchSize=1000 => MaxNumberOfMessages=10 # SQS provider uses override value`
- `BatchSize=7 => MaxNumberOfMessages=7`

The actual value that we're expecting to retrieve is in the `sqs_override_max_message_count` -- where this param takes presedence in the SQS provider over `MaxNumberOfMessages`

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
gregfurman: We previously were not passing in any custom params so the `pop` wasn't necessary 🙂

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
gregfurman: I wasn't sure whether regex was supported. If not, we can just do 2 register calls instead i.e

```python
event_system.register("before-call.sqs.ReceiveMessage", handle_inject_headers)
event_system.register("before-call.sqs.DeleteMessageBatch", handle_inject_headers)
```

gregfurman: The docs indicate that `register` just supports [wildcard matching](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html#wildcard-matching):

> Another aspect of Boto3’s event system is that it has the capability to do wildcard matching using the '*' notation.

Perhaps its a better idea to explicitly register to those calls 🤔

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
joe4dev: absolutely right 💯  all good 👍👍👍 

sorry for the confusion (I inferred max from DEFAULT_MAX_RECEIVE_COUNT, but we don't want to exceed that 🙈 😴).

PR #12234, opened 2025-02-06
[SQS] Add bootstrap test for graceful shutdown during long-poll

gregfurman:
## Motivation
Adds a bootstrap test to illustrate long-polling SQS calls shutting down gracefully.

## Changes
- Adds a `TestBlockingServiceShutdown.test_shutdown_during_sqs_long_poll` bootstrap test that will create some threaded long poll calls that exit early and without error when LS is being shutdown.


alexrashed:
@gregfurman this one seems pretty outdated, do you want to close or revive it?

PR #12240, opened 2025-02-10
[ESM] Provide SqsPoller with default parameters

gregfurman:
## Motivation
Provides default configuration values to the SqsPoller in the event that configuration is not set via the `PipeSourceSqsQueueParameters` dictionary.

## Changes
- The following defaults have been added to the `SqsPoller`
   - batch size defaults to `10`
   - batch window defaults to `0` -- which is short polling by default. Note: batch windows are not yet implemented in the SqsPoller so this value doesn't get used anywhere.


tiurin (approved):

silv-io (approved):
LGTM!

PR #12243, opened 2025-02-10
Workaround circleci test split cant be empty

tiurin:
## Motivation

some tests to make pipeline not fail when test selection is not empty but doesn't contain integration tests, found in https://github.com/localstack/localstack/pull/12234


alexrashed:
Closing this one as we are not on CircleCI anymore ;)

PR #12247, opened 2025-02-10
update rolo to 0.7.5

bentsku:
## Motivation
Updating the rolo dependency to contain the following fix: https://github.com/localstack/rolo/pull/26

## Changes
- update rolo from 0.7.4 to 0.7.5

### DIFF
--- a/requirements-base-runtime.txt
+++ b/requirements-base-runtime.txt
@@ -164,7 +164,7 @@ rfc3339-validator==0.1.4
     # via openapi-schema-validator
 rich==13.9.4
     # via localstack-core (pyproject.toml)
-rolo==0.7.4
+rolo==0.7.5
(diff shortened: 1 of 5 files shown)


bentsku:
@alexrashed is it voluntary that we do not have codeowners for those requirements files?

alexrashed:
This PR has been superseded by #12250, which contains the `rolo` upgrade.
Concerning codeowners on requirements: Yes, there are many parts of the codebase which doesn't have codeowners assigned. Who would you assign as codeowners for the requirements lock files?

bentsku:
> Who would you assign as codeowners for the requirements lock files?

Good question! 😄 I had realized there were for some downstream dependencies, so that's why I wondered. Thanks for closing it!

Issue #12255, opened 2025-02-12
bug: Deleting versions while listing them is broken since 3.0.0


philipp94831:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

We need to delete many object versions in S3. We use this snippet to delete all versions of objects for a given prefix
```java
        final S3Client s3Client = ...;
        final String bucketName = ...;
        final String prefix = ...;
        final ListObjectVersionsRequest request = ListObjectVersionsRequest.builder()
                .bucket(bucketName)
                .prefix(prefix)
                .build();
        final ListObjectVersionsIterable versionListing = s3Client.listObjectVersionsPaginator(request);
        versionListing.stream()
                .map(response -> response.versions().stream()
                        .map(s3Object -> ObjectIdentifier.builder()
                                .key(s3Object.key())
                                .versionId(s3Object.versionId())
                                .build())
                        .collect(Collectors.toList()))
                .forEach(keys -> {
                    if (!keys.isEmpty()) {
                        s3Client.deleteObjects(DeleteObjectsRequest.builder()
                                .bucket(bucketName)
                                .delete(builder -> builder
                                        .objects(keys))
                                .build());
                    }
                });
```

If there are more than 999 versions, which is the maximum page size for listing versions, the second `ListObjectVersionsResponse` is empty, although it should not be. Therefore, the remaining versions are not deleted. This is broken since localstack 3.0.0 but works in 2.3.2. We are using AWS SDK 2.29.52.

### Expected Behavior

Object versions are properly retrieved while deleting them.

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    Testcontainers Java

### Environment

```markdown
- OS: Ubuntu and Windows
- LocalStack:
  LocalStack version: 3.0.0
  LocalStack Docker image sha:
  LocalStack build date:
  LocalStack build git hash:
```

bentsku:
Hello @philipp94831 and thanks for your report!

At first I was unable to reproduce the issue, because I created 1250 versions across 50 different objects. But once doing it across one object, I can reproduce. Probably an issue in how we handle the markers. I'll start working on a fix and update you!

philipp94831:
@bentsku Thanks for fixing! Indeed in our case we had that many versions for a single object. Sorry for not specifying. I hope it didn't take you too long to figure that out

philipp94831:
[...]

PR #12259, opened 2025-02-13
fix ListObjectVersions pagination when version id marker is deleted

bentsku:
## Motivation
As reported with #12255, we've had an issue with `ListObjectVersions` and deletion.
The failing scenario is reproduced in the added test, but it can be summarized like that:

The list operation, when paginated and arriving at the max amount it can return, will return a `VersionIdMarker`, representing the last marker it returned (newer operations of S3 return the next one, which is a bit smarter). 

If you happen to delete this version, AWS will return the next value in line after this now deleted marker. Not sure exactly how they do it, if they encode the data in the version id, or keep track of deleted versions. I've tried to reverse engineer the VersionId base64 encoding, but didn't really get anything and didn't try very hard.

We didn't have anything to compare against as those values were randomly generated, and we didn't have a way to compare them against each other. I've introduced some ever increasing number to be the prefix of the b64 encoded value, so that we can lexicographically sort and compare them. 

## Changes
- introduce ever-increasing number byte prefix in the version id of S3 objects to allow comparison
- add the proper comparison in the pagination of `ListObjectVersions`
- add new test for the scenario of the version id marker being deleted, and pagination in general

## Side notes

- I'm packing the sequence number as 6 bytes, so a max of 281 474 976 710 656. Minus the start time in millisecond right now, it would give approx. 279735559077354 increases (meaning new versions) before we overflow. I think we're good 😄 (irrational fear of overflow).

- pagination is a mess in S3, not one operation implements it the same way, there's always a subtle difference 😅

_fixes #12255_

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -1854,6 +1854,13 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                     if version.version_id == version_id_marker:
                         version_key_marker_found = True
                         continue
+
+                    # it is possible that the version_id_marker related object has been deleted, in that case, start
+                    # as soon as the next version id is smaller than the version id marker (meaning this version was
+                    # next after the now-deleted version)
+                    elif version.version_id < version_id_marker:
+                        version_key_marker_found = True
+
                     elif not version_key_marker_found:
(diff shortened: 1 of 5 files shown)


k-a-il (approved):
LGTM 👍

PR #12260, opened 2025-02-13
skip flaky transcribe tests

alexrashed:
## Motivation
We have seen a bunch of flaky runs with the Transcribe tests in the past, and it seems that we still see issues.
This PR skips the last test which actually performs an actual transcription task. This should definitely be addressed fairly soon, but for the sake of the stability of our pipelines we are right now actively skipping all flaky tests.
/cc @sannya-singal 

## Changes
- Skips the last `transcribe` service integration tests which actually perform transcriptions.

### DIFF
--- a/tests/aws/services/transcribe/test_transcribe.py
+++ b/tests/aws/services/transcribe/test_transcribe.py
@@ -136,6 +136,7 @@ class TestTranscribe:
             "$..Error..Code",
         ]
     )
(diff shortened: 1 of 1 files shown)


ackdav (approved):

sannya-singal (approved):
Thanks for jumping on this 🦸 LGTM!

PR #12262, opened 2025-02-13
fix flaky transcribe tests

sannya-singal:
## Motivation

Fix flaky pipeline runs: 
- https://app.circleci.com/pipelines/github/localstack/localstack/31132/workflows/352699b7-785d-4672-a966-a5f073e5a6e6/jobs/277165/tests 
- https://app.circleci.com/pipelines/github/localstack/localstack/31131/workflows/2aaa3688-b9ae-4db9-b661-ba4693df7531/jobs/277148/tests

by making the downloading of vosk models more robust. 

## Changes
This PR: 
- adds provision to download the models from the vosk server and if it fails, it fallbacks to download the models from the [Hugging Face](https://huggingface.co/localstack/vosk-models). 
- removes skipping the flaky tests: `test_transcribe_happy_path`, `test_transcribe_supported_media_formats` and `test_transcribe_start_job`. 

Successful pipeline runs: https://app.circleci.com/pipelines/github/localstack/localstack?branch=test-transcribe

### DIFF
--- a/localstack-core/localstack/services/transcribe/provider.py
+++ b/localstack-core/localstack/services/transcribe/provider.py
@@ -30,7 +30,7 @@ from localstack.aws.api.transcribe import (
     TranscriptionJobSummary,
 )
 from localstack.aws.connect import connect_to
-from localstack.constants import ARTIFACTS_REPO
+from localstack.constants import HUGGING_FACE_ENDPOINT
 from localstack.packages.ffmpeg import ffmpeg_package
 from localstack.services.s3.utils import (
     get_bucket_and_key_from_presign_url,
@@ -45,9 +45,7 @@ from localstack.utils.threads import start_thread
 
 LOG = logging.getLogger(__name__)
 
-VOSK_MODEL_ARTIFACTS_URL = (
-    f"{ARTIFACTS_REPO}/raw/a1f6e430de71d383bc7477f130da052f8686deec/vosk/models/"
-)
+VOSK_MODELS_URL = f"{HUGGING_FACE_ENDPOINT}/vosk-models/resolve/main/"
 
 # Map of language codes to language models
 LANGUAGE_MODELS = {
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/transcribe/provider.py
```diff
@@ -235,11 +240,26 @@ def download_model(name: str):
...
+            download_urls = [VOSK_MODEL_ARTIFACTS_URL, MODEL_PRE_URL]
+
+            for url in download_urls:
+                if _try_download(url):
+                    # Download the models from the localstack artifacts if present else download from the Vosk URL
+                    LOG.info("Successfully downloaded model from %s", url)
+                    break
+                else:
+                    LOG.error("Failed to download model from %s", url)
+                    return
```
viren-nadkarni, line 262: Since the goal is to only improve the robustness of the CI, I would avoid changing the model URL and download logic which will affect production releases.

Instead, I would strongly suggest making it possible to override `MODEL_PRE_URL` using an environment configuration variable, and set it to our artifacts during CI test execution.

viren-nadkarni (changes requested):

localstack-core/localstack/services/transcribe/provider.py
sannya-singal: The model selection for now is based on the tests that we have for transcription, but its not exclusively intended to be used for CI. As a fallback if we do not have the model being used in our assets we could always download it from the source itself.

viren-nadkarni: Once the artifact URL is hardcoded and the Docker image shipped, it's a passive burden to maintain the artifacts over at https://github.com/localstack/localstack-artifacts.

In my opinion, we should either host all models ourself (at a proper location, which may not be the artifacts GitHub repo), or we defer to Vosk servers. The proposed change first queries the artifacts repo, then Vosk servers. For non-English models, the first HTTP call will always fail and is not good design.

By introducing a new environment config option that can override `MODEL_PRE_URL`, we can not only fix the network flakiness in CI tests which can use `en-XX` models, but also provide an option for certain users who may be behind corporate proxies and might prefer to host the models themselves.

sannya-singal: If we override the `MODEL_PRE_URL` using an env config option and set it to our artifacts during the CI test runs, we would deviate from the replicating the same environment otherwise used by the users because they would download these models from the traditional vosk servers. 

Currently there are 16 language models that are supported in the transcribe service in localstack each almost ~40MB, in my opinion we can also do the following: 
- For: `vosk-model-small-en-gb-0.15`, `vosk-model-small-en-in-0.4` and `vosk-model-small-en-us-0.15` i.e. `en-XX` models we always download from the artifacts/s3 assets and if the model is not `en-XX` we never check for these models in our self hosted location which would means that first HTTP would never fail. 
- add retry mechanism if the downloading of the models fails from the vosk server. 
- host the models say eg: in s3 assets

What do you think?

sannya-singal: From what I have seen so far is this error occurs 2 yearly, so hosting all these models given the size might not be the best option. Instead we could choose opting in for `en-XX` models only.

viren-nadkarni: > If we override the `MODEL_PRE_URL` using an env config option and set it to our artifacts during the CI test runs, we would deviate from the replicating the same environment otherwise used by the users because they would download these models from the traditional vosk servers.

That is acceptable. We only offer parity only for the AWS API, not transcription.

viren-nadkarni: Let's switch to a proper mirror. I've uploaded all small models to https://huggingface.co/localstack/vosk-models

The retry mechanism also sounds good, with Vosk servers as primary and Hugging Face as secondary. The override config option can be left as future work.

viren-nadkarni, line 262: This pattern is a little hard to follow. Can I suggest simplifying this as follows:

```suggestion
            for url in download_urls:
                try:
                    download(
                MODEL_PRE_URL + str(model_path.name) + ".zip", model_zip_path, verify_ssl=False
            )
                except Exception as exc:
                    LOG.warning(...)
                    continue
                break
```

viren-nadkarni (approved):
Looks great! The mirrored models should add resiliency to the Transcribe service startup and test executions.

Thanks for working on this :100:

PR #12264, opened 2025-02-13
[util] Create exponential backoff utility class

gregfurman:
## Motivation
This PR creates a re-usable component for calculating backoff durations. This was inspired by the exponential backoff algorithm from [Google's HTTP Client Library for Java](https://github.com/google/google-http-java-client/blob/da1aa993e90285ec18579f1553339b00e19b3ab5/google-http-client/src/main/java/com/google/api/client/util/ExponentialBackOff.java).

## Changes
- Adds an `ExponentialBackoff` utility class whose `next_backoff` method calculates and returns the next duration that should be waited for.

## Usage

The below shows this class being used with a maximum of 3 attempts (with no random jitter):

```python
import time

from localstack.utils.backoff import ExponentialBackoff

boff = ExponentialBackoff(max_retries=3, jitter=0)

time.wait(boff.next_backoff()) # waits 0.5s
time.wait(boff.next_backoff()) # waits 0.75s
time.wait(boff.next_backoff()) # waits 0.125s

# immediately returns since max_retries have been met
time.wait(boff.next_backoff())

# reset the backoff to the initial state
boff.reset()

# begin backing off again
time.wait(boff.next_backoff()) # waits 0.5s
# ...
```


tests/unit/utils/test_backoff.py
```diff
@@ -0,0 +1,137 @@
+import time
+import unittest
```
tiurin, line 2: suggestion: use pytest, as we [standardize on it](https://github.com/localstack/localstack/blob/7e58c741604b420993df42a3f7e9331d5fe010cb/docs/testing/integration-tests/README.md#L21) (even though the comment is in integration tests section, I see new unit tests using or implying pytest).

tiurin, line 2: suggestion: use pytest, as we [standardize on it](https://github.com/localstack/localstack/blob/7e58c741604b420993df42a3f7e9331d5fe010cb/docs/testing/integration-tests/README.md#L21) (even though the comment is in integration tests section, I see new unit tests using or implying pytest).

localstack-core/localstack/utils/backoff.py
```diff
@@ -0,0 +1,95 @@
...
+    randomization_factor: float = 0.5  # Factor to randomize backoff [0,1]
+    multiplier: float = 1.5  # Multiply interval by this factor each retry
+    max_interval: float = 60.0  # Maximum backoff interval in seconds
+    max_retries: int = -1  # Max retry attempts (-1 for unlimited)
+    max_time_elapsed: float = -1  # Max total time in seconds (-1 for unlimited)
+
+    def __post_init__(self):
+        self.retry_interval: float = 0
+        self.retries: int = 0
+        self.start_t: float = 0.0
```
tiurin, line 59: nit: even though the meaning is quite evident here, I'd suggest to not use magic letters in names. 
```suggestion
        self.start_time: float = 0.0
```

localstack-core/localstack/utils/backoff.py
```diff
@@ -0,0 +1,95 @@
...
+    | 7         | 5.692                | [2.846, 8.538]            |
+    | 8         | 8.538                | [4.269, 12.807]           |
+    | 9         | 12.807               | [6.403, 19.210]           |
+    | 10        | 19.210               | 0                         |
+
+    Note: The sequence stops at request #10 when `max_retries` or `max_time_elapsed` is exceeded
+    """
+
+    initial_interval: float = 0.5  # Initial backoff interval in seconds
+    randomization_factor: float = 0.5  # Factor to randomize backoff [0,1]
```
tiurin, line 50: suggestion: validate that factor doesn't go beyond [0,1] range

tiurin (commented):
Nice implementation!

Utility in itself is easy to understand, it would be great to see a companion example PR with its real-world usage, to understand the reason for this addition and how it fits the solution (e.g. to understand why we write utility ourselves and don't use some ready solution).

gregfurman:
@tiurin Thanks for the review!

> [...] it would be great to see a companion example PR with its real-world usage, to understand the reason for this addition and how it fits the solution 

I've linked a companion PR showing its usage 🙂 

> (e.g. to understand why we write utility ourselves and don't use some ready solution).

From what I read, large + well maintained external retry libraries provide us with retry decorators (see https://tenacity.readthedocs.io/en/latest/ or https://github.com/litl/backoff) which would be far more clunky to integrate than the current class-based approach.

tiurin:
> I've linked a companion PR showing its usage 🙂

Thank you! ❤️ 

> From what I've read, most large + well maintained external retry libraries provide us with retry function decorators (see https://tenacity.readthedocs.io/en/latest/ or https://github.com/litl/backoff) which would be far more difficult to integrate into our current implementations than the class-based approach.

Indeed, in linked PR case of a loop with a complex exit condition a separate backoff mechanism is better suited! 👍 It's not too easy to wrap something with a simple retry there, nor needed probably.
Also, for `backoff` one the maintainer seems to be inactive, so importing it might not be the best solution, so tenacity looks like the main candidate, and indeed it is not well suited.

For the reference, we also have our own collection of `retry/wait` functions here (used mostly in tests but also in some ext production code as well), with `wait_until` having a notion of exponential backoff: https://github.com/localstack/localstack/blob/958890a40ed4f5f1419a4cc69a85baa5c159c5cc/localstack-core/localstack/utils/sync.py#L16

I do agree a separate class that you suggest is easier to adjust to different use cases as its only job is to keep track of numbers. 👍

tests/unit/utils/test_backoff.py
gregfurman: Tbh we don't really even need to use a testing framework here. I'll remove `unittest` and just use assertions

gregfurman:
> For the reference, we also have our own collection of retry/wait functions here (used mostly in tests but also in some ext production code as well), with wait_until having a notion of exponential backoff: 

Good reference here. Those functions wont allow us much flexability and could cause blocking waits + shutdowns since the `wait_until` (or equivalent) doesn't provide us the ability to cancel early.

Also, I prefer the utility-based approach since it makes no assumptions about where the backoff is going to be called from. 

That is, the `boff` object exists purely to provide some caller with what backoff _should_ be after a certain number of invocations as opposed to actually waiting on that calculated duration.

localstack-core/localstack/utils/backoff.py
```diff
@@ -0,0 +1,97 @@
...
+    | 8         | 8.538                | [4.269, 12.807]           |
+    | 9         | 12.807               | [6.403, 19.210]           |
+    | 10        | 19.210               | 0                         |
+
+    Note: The sequence stops at request #10 when `max_retries` or `max_time_elapsed` is exceeded
+    """
+
+    initial_interval: float = Field(0.5, title="Initial backoff interval in seconds", gt=0)
+    randomization_factor: float = Field(0.5, title="Factor to randomize backoff", ge=0, le=1)
+    multiplier: float = Field(1.5, title="Multiply interval by this factor each retry", gt=0)
```
tiurin, line 53:
```suggestion
    multiplier: float = Field(1.5, title="Multiply interval by this factor each retry", gt=1)
```
I think shrinking intervals are not expected from exponential backoff, should we limit this with validation? Thanks for pydantic btw! ❤️

tiurin (approved):
Great work @gregfurman! I just have one last validation comment, leaving it up to you to address in this or in comanion PR as you see fit. 👍

PR #12270, opened 2025-02-13
fix ListObjectVersions pagination comparison

bentsku:
## Motivation
Follow up from #12259, I've introduced flakiness in the tests and bad behavior (also reported). I saw the failed run here: https://github.com/localstack/localstack/actions/runs/13314341024/job/37184566141 and realized my mistake instantly 🤦‍♂️ 

I incorrectly assumed the b64 encoded data could be sorted lexicographically, but this was wrong. The increasing values are cycling from upper case, lower case, digits, back to upper case. However, there is one comparison that fails: `z` to `0`. 

This is because if we look at the proper Unicode code point for the characters, we go from `y`, `z`, `0`, `1` to -> `49`,  `48`, `122`, `121`. So we break the order at this one character. (if they had started with digits, things would have been fine 😝)

I've introduced a small utility function to properly compare version between each other, maybe using hexadecimal characters could have been fine, but who knows what people does with `VersionId` field 😅 keeping the format as close as possible to AWS might be preferable.

Also introduced unit tests for it, and an integration test to validate the paginator behavior. 

## Changes
- fix the `VersionId` comparison to properly look at the raw data instead of the encoded one
- add unit test for behavior
- add integration test

_fixes #12255_

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -282,6 +282,7 @@ from localstack.services.s3.utils import (
     get_system_metadata_from_request,
     get_unique_key_id,
     is_bucket_name_valid,
+    is_version_older_than_other,
     parse_copy_source_range_header,
     parse_post_object_tagging_xml,
     parse_range_header,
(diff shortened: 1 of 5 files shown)


cloutierMat (approved):
Awesome! Thanks for the super proactive fix. :pray: for keeping the pipeline :green_heart:!

The naming `is_version_older_than_other` feels a bit off to me, but it might just be a Friday morning vibe! :rofl:

PR #12271, opened 2025-02-14
Return correct IP address when LS container runs inside K8s cluster

RobertLucian:
## Motivation

The `get_endpoint_for_network` function is used to determine what the IP address of the main container is (whether it's running on Docker or Kubernetes, or in host mode).

We check if the k8s pod adds an env of the `LOCALSTACK_PORT_4566_TCP_ADDR` kind when port 4566 is open. Somehow when attempting to try if the env var associated with port 53 exists or not, it doesn't show up, despite the fact that the deployment has port 53 (tcp/udp) exposed. Anyways, this doesn't seem to be a problem as we make use of the `dns_server.is_server_running` function to determine if the DNS server is running (at the very least). 

## Changes

What's left to do:

- [ ] Change the helm chart / k8s operator codebases to export an env var (i.e. `LOCALSTACK_K8S_POD_IP`) through the use of the `status.podIP` metadata instead. Use `LOCALSTACK_K8S_POD_IP` instead to determine if we're running inside a K8s pod.

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -290,6 +290,13 @@ def ping(host):
     )
 
 
+def tcp_port4566_open():
+    """
+    Returns True if this runs inside a Kubernetes pod and if the sidecar has port 4566 (TCP) open.
+    """
+    return os.getenv("LOCALSTACK_PORT_4566_TCP_ADDR") is not None
+
+
 def in_docker():
     """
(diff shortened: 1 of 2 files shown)


RobertLucian:
@dfangl closing this one down as I've brought the necessary stop-gap alternative solution in https://github.com/localstack/localstack-ext/pull/4038.

PR #12281, opened 2025-02-18
[ESM] Add backoff between Stream Poller retries

gregfurman:
## Motivation
This PR leverages the backoff util to ensure, when retrying, we wait between retries so as to not spam the processing loop.

Companion PR to https://github.com/localstack/localstack/pull/12264

## Changes
- Checks `has_record_expired` every iteration to allow for an expired record to break processing.
- Use `backoff`utility, in conjunction with a `threading.Event`, to exponentially back-off between processing retries.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import threading
 from abc import abstractmethod
 from datetime import datetime
 from typing import Iterator
@@ -28,6 +29,7 @@ from localstack.services.lambda_.event_source_mapping.pollers.poller import (
 )
 from localstack.services.lambda_.event_source_mapping.pollers.sqs_poller import get_queue_url
 from localstack.utils.aws.arns import parse_arn, s3_bucket_name
+from localstack.utils.backoff import ExponentialBackoff
 from localstack.utils.strings import long_uid
 
 LOG = logging.getLogger(__name__)
@@ -47,6 +49,9 @@ class StreamPoller(Poller):
     # The ARN of the processor (e.g., Pipe ARN)
     partner_resource_arn: str | None
 
+    # Used for backing-off between retries and breaking the retry loop
+    _is_shutdown: threading.Event
+
     def __init__(
         self,
         source_arn: str,
@@ -61,6 +66,8 @@ class StreamPoller(Poller):
         self.esm_uuid = esm_uuid
         self.shards = {}
         self.iterator_over_shards = None
+
+        self._is_shutdown = threading.Event()
 
     @abstractmethod
     def transform_into_events(self, records: list[dict], shard_id) -> list[dict]:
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -187,9 +210,26 @@ def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
...
+                    abort_condition = "RecordAgeExpired"
+
+                if attempts > 0:
+                    # TODO: Should we always backoff (with jitter) before processing since we may not want multiple pollers
+                    # all starting up and polling simultaneously
+                    # For example: 500 persisted ESMs starting up and requesting concurrently could flood gateway
+                    self._is_shutdown.wait(boff.next_backoff())
+
                 self.processor.process_events_batch(events)
+                boff.reset()
```
gregfurman, line 231: @tiurin An example of how we can use the new backoff utility 👆👀

tiurin (approved):

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -187,9 +210,26 @@ def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
...
+            not abort_condition
+            and not self.max_retries_exceeded(attempts)
+            and not self._is_shutdown.is_set()
+        ):
             try:
+                if self.has_record_expired(polled_events[-1]):
+                    abort_condition = "RecordAgeExpired"
+
+                if attempts > 0:
+                    # TODO: Should we always backoff (with jitter) before processing since we may not want multiple pollers
```
joe4dev, line 225: Starting with backoff after a retry seems a good solution to start with for now, not impacting any best-case responsiveness

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -104,12 +111,29 @@ def format_datetime(self, time: datetime) -> str:
...
         return events
 
+    def has_record_expired(self, event: dict):
+        # Check MaximumRecordAgeInSeconds
+        if maximum_record_age_in_seconds := self.stream_parameters.get("MaximumRecordAgeInSeconds"):
+            arrival_timestamp_of_last_event = event.get("approximateArrivalTimestamp")
+            if not arrival_timestamp_of_last_event:
+                return False
+
+            now = get_current_time().timestamp()
```
joe4dev, line 130: Are we handling timezones correctly here?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -104,12 +111,29 @@ def format_datetime(self, time: datetime) -> str:
...
+    def close(self):
+        self._is_shutdown.set()
+
     def pre_filter(self, events: list[dict]) -> list[dict]:
         return events
 
     def post_filter(self, events: list[dict]) -> list[dict]:
         return events
 
+    def has_record_expired(self, event: dict):
```
joe4dev, line 123: That's an untested feature for which we should ideally add a test (which is a bit annoying because of the 60s minimum https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-maximumrecordageinseconds). Can we at least add a backlog item for this?

joe4dev (approved):
Nice application of the backoff utility 👍 

It would be great to highlight the motivation on why we are doing this from a customer perspective (as discussed and raised by @dfangl ) in future PR descriptions.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
gregfurman: I'm happy to revert these changes and make a follow-up PR with a test instead?

dfangl: This is blocking for one of my PRs (as the logs are so massive they cannot reasonably be printed due to the retries), would be nice to get this in a separate PR, as there seem to be some confusion still.

gregfurman: Done!

dfangl (approved):
Looks good, thanks for the fix! No more 200k log lines after the function was deleted :)

Issue #12231, opened 2025-02-05
bug: Elastic Search 6.8 - JAVA_HOME doesn't exist


daveodwyer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When creating a new elastic search 6.8 domain within the localstack container like so:

```
awslocal es create-elasticsearch-domain --domain-name my-domain-name --elasticsearch-version 6.8
```

You get a response suggesting the domain is processing:

```
{
    "DomainStatus": {
        "DomainId": "000000000000/my-domain-name",
        "DomainName": "my-domain-name",
        "ARN": "arn:aws:es:us-east-1:000000000000:domain/my-domain-name",
        "Created": true,
        "Deleted": false,
        "Endpoint": "localstack:4566/es/us-east-1/my-domain-name",
        "Processing": true,
        "UpgradeProcessing": false,
        "ElasticsearchVersion": "6.8",
        "ElasticsearchClusterConfig": {
            "InstanceType": "m3.medium.elasticsearch",
            "InstanceCount": 1,
            "DedicatedMasterEnabled": true,
            "ZoneAwarenessEnabled": false,
            "DedicatedMasterType": "m3.medium.elasticsearch",
            "DedicatedMasterCount": 1,
            "WarmEnabled": false,
            "ColdStorageOptions": {
                "Enabled": false
            }
        },
        "EBSOptions": {
            "EBSEnabled": true,
            "VolumeType": "gp2",
            "VolumeSize": 10,
            "Iops": 0
        },
        "AccessPolicies": "",
        "SnapshotOptions": {
            "AutomatedSnapshotStartHour": 0
        },
        "CognitoOptions": {
            "Enabled": false
        },
        "EncryptionAtRestOptions": {
            "Enabled": false
        },
        "NodeToNodeEncryptionOptions": {
            "Enabled": false
        },
        "AdvancedOptions": {
            "override_main_response_version": "false",
            "rest.action.multi.allow_explicit_index": "true"
        },
        "ServiceSoftwareOptions": {
            "CurrentVersion": "",
            "NewVersion": "",
            "UpdateAvailable": false,
            "Cancellable": false,
            "UpdateStatus": "COMPLETED",
            "Description": "There is no software update available for this domain.",
            "AutomatedUpdateDate": 0.0,
            "OptionalDeployment": true
        },
        "DomainEndpointOptions": {
            "EnforceHTTPS": false,
[...]

PR #12286, opened 2025-02-19
ElasticSearch: Fix missing JRE in 5.x and 6.x

viren-nadkarni:
## Background

In https://github.com/localstack/localstack/pull/11462, we changed the ElasticSearch/OpenSearch installers to use the [bundled JRE](https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html#jvm-version). However, based on https://github.com/localstack/localstack/issues/12231, it was identified that ElasticSearch 5.x and 6.x builds [do not](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/zip-targz.html) have a bundled JRE and instead require a separate JRE installation.

## Changes

This PR introduces a specialised package installer for ElasticSearch 5.x and 6.x which manages the Java dependency through standardised approach.

## Tests

Tested manually locally. On the CI, automated tests run only against the latest versions, which was also why this went undetected.

## Related

Closes: https://github.com/localstack/localstack/issues/12231

### DIFF
--- a/localstack-core/localstack/services/opensearch/cluster.py
+++ b/localstack-core/localstack/services/opensearch/cluster.py
@@ -676,13 +676,24 @@ class ElasticsearchCluster(OpensearchCluster):
         settings = {
             "http.port": self.port,
             "http.publish_port": self.port,
-            "transport.port": "0",
             "network.host": self.host,
             "http.compression": "false",
             "path.data": f'"{dirs.data}"',
             "path.repo": f'"{dirs.backup}"',
-            "discovery.type": "single-node",
         }
+
+        # This config option was renamed between 6.7 and 6.8, yet not documented as a breaking change
+        # See https://github.com/elastic/elasticsearch/blob/f220abaf/build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java#L1349-L1353
+        if self.version.startswith("Elasticsearch_5.") or (
+            self.version.startswith("Elasticsearch_6.") and self.version != "Elasticsearch_6.8"
+        ):
+            settings["transport.tcp.port"] = "0"
+        else:
+            settings["transport.port"] = "0"
+
+        # `discovery.type` had a different meaning in 5.x
+        if not self.version.startswith("Elasticsearch_5."):
+            settings["discovery.type"] = "single-node"
 
         if os.path.exists(os.path.join(dirs.mods, "x-pack-ml")):
             settings["xpack.ml.enabled"] = "false"
--- a/localstack-core/localstack/services/opensearch/cluster.py
+++ b/localstack-core/localstack/services/opensearch/cluster.py
@@ -16,7 +16,6 @@ from localstack.aws.api.opensearch import (
 )
 from localstack.http.client import SimpleRequestsClient
 from localstack.http.proxy import ProxyHandler
-from localstack.packages import InstallTarget
 from localstack.services.edge import ROUTER
 from localstack.services.opensearch import versions
 from localstack.services.opensearch.packages import elasticsearch_package, opensearch_package
@@ -670,7 +669,7 @@ class ElasticsearchCluster(OpensearchCluster):
         return constants.OS_USER_OPENSEARCH
 
     def _ensure_installed(self):
-        elasticsearch_package.install(self.version, target=InstallTarget.VAR_LIBS)
+        elasticsearch_package.install(self.version)
 
     def _base_settings(self, dirs) -> CommandSettings:
         settings = {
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/opensearch/cluster.py
```diff
@@ -669,7 +670,7 @@ def os_user(self):
         return constants.OS_USER_OPENSEARCH
 
     def _ensure_installed(self):
-        elasticsearch_package.install(self.version)
+        elasticsearch_package.install(self.version, target=InstallTarget.VAR_LIBS)
```
viren-nadkarni, line 673: `target` uses `InstallTarget.VAR_LIBS` as a default anyway, but it is made explicit here because `get_java_env_vars()` method below needs it.

localstack-core/localstack/services/opensearch/cluster.py
```diff
@@ -669,28 +670,41 @@ def os_user(self):
...
+            self.version.startswith("Elasticsearch_6.") and self.version != "Elasticsearch_6.8"
+        ):
+            settings["transport.tcp.port"] = "0"
+        else:
+            settings["transport.port"] = "0"
+
+        # `discovery.type` had a different meaning in 5.x
+        if not self.version.startswith("Elasticsearch_5."):
+            settings["discovery.type"] = "single-node"
+
```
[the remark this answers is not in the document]
kenny-alvizuris-cko: Thanks for the fix @viren-nadkarni willing to give it a try once is released, I'm facing the same issue

localstack-core/localstack/services/opensearch/cluster.py
alexrashed: **note:** Unfortunately, this is not necessarily the case... If the installer is used by `lpm`, the `InstallTarget.STATIC_LIBS` will be used. So this will actually break when the package is installed at Docker build time. But this can easily be fixed by just making the `get_java_env_vars` independent of it (see comment below) 😉

localstack-core/localstack/services/opensearch/packages.py
```diff
@@ -233,6 +236,12 @@ class ElasticsearchPackageInstaller(PackageInstaller):
     def __init__(self, version: str):
         super().__init__("elasticsearch", version)
 
+    def get_java_env_vars(self, target: InstallTarget) -> dict[str, str]:
+        install_dir = self._get_install_dir(target)
```
alexrashed, line 240: **issue:** This is the root of the issue outlined in the comment above. Here you really shouldn't use the target where the package is "about to be installed if it's not yet installed", but you should use [`get_installed_dir`](https://github.com/localstack/localstack/blob/3026d5cb618c4d64d6be18ff0aa7619954a2143b/localstack-core/localstack/packages/api.py#L116-L125) instead, which dynamically detects the installation directory where the package _has already been_ installed.‏

localstack-core/localstack/services/opensearch/packages.py
```diff
@@ -42,6 +43,8 @@ def __init__(self, default_version: str = OPENSEARCH_DEFAULT_VERSION):
 
     def _get_installer(self, version: str) -> PackageInstaller:
         if version in versions._prefixed_elasticsearch_install_versions:
+            if version.startswith("Elasticsearch_5.") or version.startswith("Elasticsearch_6."):
+                return ElasticsearchLegacyPackageInstaller(version)
```
alexrashed, line 47: **praise:** Great catch! It appears that there are [still ~1,5 years left until versions < 6.8 are EOL](https://aws.amazon.com/blogs/big-data/amazon-opensearch-service-announces-standard-and-extended-support-dates-for-elasticsearch-and-opensearch-versions/)‏.

alexrashed (commented):
Thanks a lot for jumping on this! Perfect implementation on the package installer to dynamically select the right installer for the version! 💯 
It really only needs a small fix on the installer (see comments), then we are good to go! 🚀

localstack-core/localstack/services/opensearch/packages.py
viren-nadkarni: Yes that's right. Added some notes in 194e4ba70 for posterity

localstack-core/localstack/services/opensearch/packages.py
viren-nadkarni: Brilliant suggestion, this will simplify everything. Fixed in 7eb88688c

localstack-core/localstack/services/opensearch/packages.py
```diff
@@ -336,5 +345,27 @@ def get_elasticsearch_install_version(self) -> str:
         return versions.get_install_version(self.version)
 
 
+class ElasticsearchLegacyPackageInstaller(ElasticsearchPackageInstaller):
```
alexrashed, line 348: **question:** ‏Would it be possible to use the `JavaInstallerMixin` here? 🤔

alexrashed (approved):
Awesome! Thanks for acting on the comment! Nice and clean! 🚀 💯

PR #12288, opened 2025-02-19
refactoring of removal of AWS::NoValue in reference list for CFn

pinzon:
## Motivation
As discovered by a customer. CFn should remove the AWS::NoValue from list of references when pasing such list as as an Attribute Value for a resource definition.

Example. In the following code, CFn should remove the AWS::NoValue from the list of RouteTableIds before attempting to create the resource.

```yaml
S3VpcEndpoint:
    Type: AWS::EC2::VPCEndpoint
    Properties:
      RouteTableIds:
        - Fn::If:
            - cNewVpc
            - Ref: PrivateSubnet1RouteTable
            - Ref: AWS::NoValue
        - Fn::If:
            - cNewVpc
            - Ref: PrivateSubnet2RouteTable
            - Ref: AWS::NoValue
``` 

The previous solution (#12163) only managed to fix this for `Fn::Join` list parameter. With the new changes, the fix is applied to any referenced list.

## Changes
- Temployer utils removes the AWS::NoValue from the lists.
- The Fn:Join procces now uses the cleaned result instead of filtering it itself

## No testing
- Since it's a refactoring a green pipeline should be enough.
- Most of the resources that directly accept a list as attribute are available only in PRO

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
@@ -755,7 +755,7 @@ def _resolve_refs_recursively(
 
         # remove _aws_no_value_ from resulting references
         clean_list = []
-        for i in range(len(value)):
+        for item in value:
             temp_value = resolve_refs_recursively(
                 account_id,
                 region_name,
@@ -764,7 +764,7 @@ def _resolve_refs_recursively(
                 mappings,
                 conditions,
                 parameters,
-                value[i],
+                item,
             )
             if not (isinstance(temp_value, str) and temp_value == PLACEHOLDER_AWS_NO_VALUE):
                 clean_list.append(temp_value)
--- a/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
@@ -398,19 +398,18 @@ def _resolve_refs_recursively(
                     join_values,
                 )
 
-            join_values = [
-                resolve_refs_recursively(
-                    account_id,
-                    region_name,
-                    stack_name,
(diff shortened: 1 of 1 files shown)


simonrw:
You mention in your PR description:

> Since it's a refactoring a green pipeline should be enough.

Only if we have examples where we use this functionality in CFn templates, right?

> Most of the resources that directly accept a list as attribute are available only in PRO

Let's make sure we do a manual run of the pro tests to be sure

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -756,8 +753,10 @@ def _resolve_refs_recursively(
                     {inner_list[0]: inner_list[1]},
                 )
 
+        # remove _aws_no_value_ from resulting references
+        clean_list = []
         for i in range(len(value)):
```
simonrw, line 758: nit: this no longer needs to be a list over the indices

```suggestion
        for item in value:
```

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -756,8 +753,10 @@ def _resolve_refs_recursively(
                     {inner_list[0]: inner_list[1]},
                 )
 
+        # remove _aws_no_value_ from resulting references
+        clean_list = []
```
simonrw, line 757: Thank you for writing the longer (but easier to understand) loop manually building up a new list, rather than some horribly complex list comprehension!

simonrw (approved):
Thanks for proving the change is valid in -ext. This change looks good to me, a nice simplification stripping out the values in the list type before it makes it to the intrinsic function call!

PR #12292, opened 2025-02-20
Logging mask credentials

cloutierMat:
## Motivation

This pr introduces a logging filter that can be used to mask sensitive values in our http logger. As we log every incoming HTTP request raw body, we can user this filter to find sensitive keys in a json string and mask their value.

## Changes

- Base class to allow for value masking
- Test including an example subclass

### DIFF
--- a/localstack-core/localstack/logging/format.py
+++ b/localstack-core/localstack/logging/format.py
@@ -72,15 +72,15 @@ class MaskSensitiveInputFilter(logging.Filter):
     b'{"my_key": "sensitive_value"}' would become b'{"my_key": "******"}'.
     """
 
-    patterns: Dict[str, re.Pattern]
+    patterns: list[tuple[re.Pattern[bytes], bytes]]
 
     def __init__(self, sensitive_keys: list[str]):
         super(MaskSensitiveInputFilter, self).__init__()
-        self.sensitive_keys = sensitive_keys
 
-        self.patterns = {
-            key: re.compile(to_bytes(rf'"{key}":\s*"[^"]+"')) for key in self.sensitive_keys
-        }
+        self.patterns = [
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/logging/format.py
```diff
@@ -61,6 +63,36 @@ def _get_compressed_logger_name(self, name):
...
+        }
+
+    def filter(self, record):
+        if record.input and isinstance(record.input, bytes):
+            record.input = self.mask_sensitive_msg(record.input)
+        return True
+
+    def mask_sensitive_msg(self, message: bytes) -> bytes:
+        for key, pattern in self.patterns.items():
+            message = re.sub(pattern, to_bytes(f'"{key}": "******"'), message)
```
bentsku, line 92: question/nit: should we actually save the pattern and the replacement value already in `self.patterns`?
Maybe we do not need a dict, and could use a list of tuple?
So that we would need to convert the replacement value every time. It's quite a minor optimisation, but I'm not seeing the value of a dict directly here as we never access it by key. 
Something like:
```python
self.patterns = [(re.compile(to_bytes(rf'"{key}":\s*"[^"]+"')), to_bytes(f'"{key}": "******"')) for key in self.sensitive_keys]
...
for pattern, repl_value in self.patterns.items():
    message = re.sub(pattern, repl_value, message)
```

bentsku (approved):
LGTM! Nice test, pretty neat 👌

I only have one comment regarding the `dict` usage for storing the pattern + key, I wonder if we could save the replacement value already at init time. Very minor, not really blocking!

localstack-core/localstack/logging/format.py
cloutierMat: you are totally right :eyes:. I missed that simple fact! :rofl: I will update :wink:

PR #12293, opened 2025-02-20
Lamba: fix unhandled error when SubnetIds is invalid

bentsku:
## Motivation
While working with Lambda and during a support case, I realized we didn't handle the case when the SubnetIds would be invalid. 

While this PR does not implement complet behavior (we only validate the first subnet id in the list), it at least makes sure that we don't return `InternalException`. 

This was the error:
```python
  File "<>/localstack/localstack-core/localstack/services/lambda_/provider.py", line 1007, in create_function
    vpc_config=self._build_vpc_config(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "<>/localstack/localstack-core/localstack/services/lambda_/provider.py", line 491, in _build_vpc_config
    vpc_id=self._resolve_vpc_id(account_id, region_name, subnet_ids[0]),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<>/localstack/localstack-core/localstack/services/lambda_/provider.py", line 473, in _resolve_vpc_id
    ).ec2.describe_subnets(SubnetIds=[subnet_id])["Subnets"][0]["VpcId"]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<>/.venv/lib/python3.11/site-packages/botocore/client.py", line 569, in _api_call
  File "<>/.venv/lib/python3.11/site-packages/botocore/client.py", line 1023, in _make_api_call
botocore.exceptions.ClientError: An error occurred (InvalidSubnetID.NotFound) when calling the DescribeSubnets operation (reached max retries: 0): The subnet ID 'bad-format-1f2656ac' does not exist
, headers={'Content-Type': 'application/json', 'X-Amzn-Errortype': 'InternalError', 'Content-Length': '3057', 'x-amzn-requestid': 'aa043e2f-52d3-4289-8646-ad4ee10ba812', 'x-amz-request-id': 'aa043e2f-52d3-4289-8646-ad4ee10ba812'})
```

And we would return an `InternalError` to the client. 

This PR also implements a skipped test to validate the `SecurityGroups`. I did not implement the logic to actually validate as it is not currently checked, and am not sure if it could lead to issues with users. 

## Changes
- add 2 tests for validating `SubnetIds` and `SecurityGroups`
- implement regex validation for the `SubnetIds[0]` of the list
- add a try...except block around the `connect_to` call in order to surface the proper exception to the client

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_api.py
+++ b/tests/aws/services/lambda_/test_lambda_api.py
@@ -1265,6 +1265,112 @@ class TestLambdaFunction:
         )
 
     @markers.aws.validated
+    def test_invalid_vpc_config_subnet(
+        self, create_lambda_function, lambda_su_role, snapshot, aws_client, cleanups
+    ):
+        """
+        Test invalid "VpcConfig.SubnetIds" Property on the Lambda Function
+        """
+        non_existent_subnet_id = f"subnet-{short_uid()}"
+        wrong_format_subnet_id = f"bad-format-{short_uid()}"
+
+        # AWS validates the Security Group first, so we need a valid one to test SubnetsIds
+        security_groups = aws_client.ec2.describe_security_groups(MaxResults=5)["SecurityGroups"]
+        security_group_id = security_groups[0]["GroupId"]
+
+        snapshot.add_transformer(snapshot.transform.regex(non_existent_subnet_id, "<subnet_id_1>"))
+        snapshot.add_transformer(snapshot.transform.regex(wrong_format_subnet_id, "<subnet_id_2>"))
+
+        zip_file_bytes = create_lambda_archive(load_file(TEST_LAMBDA_PYTHON_ECHO), get_content=True)
+
+        with pytest.raises(ClientError) as e:
+            aws_client.lambda_.create_function(
+                FunctionName=f"fn-{short_uid()}",
+                Handler="index.handler",
+                Code={"ZipFile": zip_file_bytes},
+                PackageType="Zip",
+                Role=lambda_su_role,
+                Runtime=Runtime.python3_12,
+                VpcConfig={
+                    "SubnetIds": [non_existent_subnet_id],
+                    "SecurityGroupIds": [security_group_id],
+                },
+            )
+
+        snapshot.match("create-response-non-existent-subnet-id", e.value.response)
+
+        with pytest.raises(ClientError) as e:
+            aws_client.lambda_.create_function(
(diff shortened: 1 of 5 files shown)


tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -1264,6 +1264,112 @@ def test_vpc_config(
             "delete_vpcconfig_get_function_response", delete_vpcconfig_get_function_response
         )
 
+    @markers.aws.validated
+    def test_invalid_vpc_config_subnet(
+        self, create_lambda_function, lambda_su_role, snapshot, aws_client, cleanups
```
joe4dev, line 1269: nit: create_lambda_function fixture unused

tests/aws/services/lambda_/test_lambda_api.py
```diff
@@ -1264,6 +1264,112 @@ def test_vpc_config(
...
+    ):
+        """
+        Test invalid "VpcConfig.SubnetIds" Property on the Lambda Function
+        """
+        non_existent_subnet_id = f"subnet-{short_uid()}"
+        wrong_format_subnet_id = f"bad-format-{short_uid()}"
+
+        # AWS validates the Security Group first, so we need a valid one to test SubnetsIds
+        security_groups = aws_client.ec2.describe_security_groups(MaxResults=5)["SecurityGroups"]
```
joe4dev, line 1278: I guess this just assumes that there is at least one valid security group (probably a sensible assumption)

joe4dev (approved):
Thank you for the fix @bentsku 👏👏

tests/aws/services/lambda_/test_lambda_api.py
bentsku: Yes, I think I've seen other tests doing this assumption with the default VPC, so figured I could as it'd be easier than to create the full VPC. Hopefully this is "safe" to do 😅

tests/aws/services/lambda_/test_lambda_api.py
bentsku: Oh no I'm so sorry, I thought I had push the changes but I forgot! 😭 there was `create_lambda_function` and `cleanups` being unused...

joe4dev: No worries at all. It was just a tiny nit ☺️

Issue #12295, opened 2025-02-20
bug: API Gateway Lambda Invoke Deletes Stage Name from Request


hubbcaps:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

With Localstack's implementation of API Gateway V1 REST API with a Lambda Integration, any request forwarded to the lambda will strip out the Stage Name from the request path to the lambda.

### Expected Behavior

Within AWS, if you create a base path mapping for a custom domain using the API to leave the values of base path and stage blank, AWS automatically looks for the stage name as the first part of the path and properly forwards to a stage. This should be emulated in Localstack. Even without base path mapping, localstack is currently stripping the request URL if hitting the execute URL directly.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Start up localstack using Docker Compose:

```
version: "3.8"

services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack-pro
    ports:
      - "127.0.0.1:4566:4566"            # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"  # external services port range
    environment:
      # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
      - DEBUG=1
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN- }
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
    env_file:
      - .env
```

Use Terraform to setup an API Gateway integrating with a lambda based on a Hot Reload directory, base path mapping should have blank stage name and blank base path so that API Gateway will try to determine the stage base on the request path:

```
resource "aws_api_gateway_rest_api" "xxx-local" {
  name = "xxx-local"

  endpoint_configuration {
    types = ["REGIONAL"]
  }

  tags = {
    "_custom_id_" = "local"
  }
}

resource "aws_api_gateway_resource" "xxx-local" {
  parent_id   = aws_api_gateway_rest_api.xxx-local.root_resource_id
  rest_api_id = aws_api_gateway_rest_api.xxx-local.id
  path_part   = "{proxy+}"
}

resource "aws_api_gateway_method" "xxx-local" {
  rest_api_id      = aws_api_gateway_rest_api.xxx-local.id
[...]

PR #12296, opened 2025-02-21
APIGW: fix AWS_PROXY payload with Custom Domains

bentsku:
## Motivation
As we got a report in #12295, we had an issue if our path handling when working with API Gateway Custom Domains and REST APIs.

This PR is a small fix to manage the Lambda Input payload `path` value, which is quite frankly weird, even for AWS.

See this comment: https://github.com/localstack/localstack/issues/12295#issuecomment-2673084002
- No mapping, no custom domain
  - `path`: `/your-path` -> stage is stripped
  - `requestContext.path`: `/<stage>/your-path`
- Mapping configured with no basePath and no stage:
  - `path`: `/<stage>/your-path` -> stage is not stripped
  - `requestContext.path`: `/<stage>/your-path`
- Mapping configured with basePath and no stage
  - `path`: `/<basePath>/<stage>/example/v1` -> nothing is stripped
  - `requestContext.path`: `/<basePath>/<stage>/example/v1`
- Mapping configured with basePath and a stage
  - `path`: `/<basePath>/example/v1` -> note that they strip the stage in that case when the stage is specified
  - `requestContext.path`: `/<basePath>/example/v1` -> the stage is even stripped here? which is not following "the rules"

So this PR implements the behavior that if a `basePath` is configured and present (meaning you're invoking from a Custom Domain), you need to use the `requestContext.path`, because our internal context `invocation_request["path"]` is used for routing, it should only have the resource path values.

I think there might be somewhat of an issue in the lambda payload construction in AWS, and even in the `requestContext`, I think there might be some bug in their implementation 😅 

## Changes
- add a small change in how we handle the `path` field for the `AWS_PROXY` payload

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/parse.py
@@ -134,7 +134,6 @@ class InvocationRequestParser(RestApiGatewayHandler):
         domain_prefix = domain_name.split(".")[0]
         now = datetime.datetime.now()
 
-        # TODO: verify which values needs to explicitly have None set
         context_variables = ContextVariables(
             accountId=context.account_id,
             apiId=context.api_id,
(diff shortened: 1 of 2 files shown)


cloutierMat (approved):
:rocket: 

Nice investigation. I guess it would be too simple if the values were consistent! :rofl: 

Thanks for the fix

PR #12297, opened 2025-02-21
[ESM] Validate event sources existence

eLRuLL:
## Motivation

This PR allows for the ESM to validate the existence of different event source resources

## Changes

- Adds check for sqs existence in the event source validation
- Adds check for kinesis streams existence in the event source validation
- Adds check for dynamodb streams existence in the event source validation

## Testing
- Created the `test_esm_with_not_existing_sqs_queue` test inside the `test_lambda_integration_sqs.py` tests

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -9,6 +9,8 @@ import threading
 import time
 from typing import IO, Any, Optional, Tuple
 
+from botocore.exceptions import ClientError
+
 from localstack import config
 from localstack.aws.api import RequestContext, ServiceException, handler
 from localstack.aws.api.lambda_ import (
@@ -160,6 +162,7 @@ from localstack.services.lambda_.event_source_mapping.esm_worker import (
 from localstack.services.lambda_.event_source_mapping.esm_worker_factory import (
     EsmWorkerFactory,
 )
+from localstack.services.lambda_.event_source_mapping.pipe_utils import get_internal_client
 from localstack.services.lambda_.invocation import AccessDeniedException
 from localstack.services.lambda_.invocation.execution_environment import (
     EnvironmentStartupTimeoutException,
@@ -224,6 +227,7 @@ from localstack.utils.aws.arns import (
     lambda_event_source_mapping_arn,
     parse_arn,
 )
+from localstack.utils.aws.client_types import ServicePrincipal
 from localstack.utils.bootstrap import is_api_enabled
 from localstack.utils.collections import PaginatedList
 from localstack.utils.event_matcher import validate_event_pattern
@@ -1866,6 +1870,52 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
     # =======================================
     # ======= EVENT SOURCE MAPPINGS =========
     # =======================================
+    def check_service_resource_exists(
+        self, service: str, resource_arn: str, function_arn: str, function_role_arn: str
+    ):
+        """
+        Check if the service resource exists and if the function has access to it.
+
+        Raises:
+            InvalidParameterValueException: If the service resource does not exist or the function does not have access to it.
+        """
+        arn = parse_arn(resource_arn)
+        source_client = get_internal_client(
+            arn=resource_arn,
+            role_arn=function_role_arn,
+            service_principal=ServicePrincipal.lambda_,
+            source_arn=function_arn,
+        )
+        if service in ["sqs", "sqs-fifo"]:
+            try:
+                source_client.get_queue_attributes(QueueUrl=arn["resource"])
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "AWS.SimpleQueueService.NonExistentQueue":
+                    raise InvalidParameterValueException(
+                        f"Error occurred while ReceiveMessage. SQS Error Code: {e.response['Error']['Code']}. SQS Error Message: {e.response['Error']['Message']}",
+                        Type="User",
+                    )
+                raise e
+        elif service in ["kinesis"]:
+            try:
+                source_client.describe_stream(StreamARN=resource_arn)
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                    raise InvalidParameterValueException(
+                        f"Stream not found: {resource_arn}",
+                        Type="User",
+                    )
+                raise e
+        elif service in ["dynamodb"]:
+            try:
+                source_client.describe_stream(StreamArn=resource_arn)
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                    raise InvalidParameterValueException(
+                        f"Stream not found: {resource_arn}",
+                        Type="User",
+                    )
+                raise e
 
     @handler("CreateEventSourceMapping", expand=False)
     def create_event_source_mapping(
@@ -1881,14 +1931,14 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         request: CreateEventSourceMappingRequest,
     ) -> EventSourceMappingConfiguration:
         # Validations
-        function_arn, function_name, state = self.validate_event_source_mapping(context, request)
+        function_arn, function_name, state, function_version, function_role = (
+            self.validate_event_source_mapping(context, request)
+        )
 
         esm_config = EsmConfigFactory(request, context, function_arn).get_esm_config()
 
         # Copy esm_config to avoid a race condition with potential async update in the store
         state.event_source_mappings[esm_config["UUID"]] = esm_config.copy()
-        function_version = get_function_version_from_arn(function_arn)
-        function_role = function_version.config.role
         enabled = request.get("Enabled", True)
         # TODO: check for potential async race condition update -> think about locking
         esm_worker = EsmWorkerFactory(esm_config, function_role, enabled).get_esm_worker()
@@ -1962,6 +2012,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         fn = state.functions.get(function_name)
         if not fn:
             raise InvalidParameterValueException("Function does not exist", Type="User")
+
         if qualifier:
             # make sure the function version/alias exists
             if api_utils.qualifier_is_alias(qualifier):
@@ -1981,6 +2032,11 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         else:
             fn_arn = api_utils.unqualified_lambda_arn(function_name, account, region)
 
+        function_version = get_function_version_from_arn(fn_arn)
+        function_role = function_version.config.role
+
+        if source_arn := request.get("EventSourceArn"):
+            self.check_service_resource_exists(service, source_arn, fn_arn, function_role)
         # Check we are validating a CreateEventSourceMapping request
         if is_create_esm_request:
 
@@ -2026,7 +2082,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                             f"existing mapping with UUID {uuid}",
                             Type="User",
                         )
-        return fn_arn, function_name, state
+        return fn_arn, function_name, state, function_version, function_role
 
     @handler("UpdateEventSourceMapping", expand=False)
     def update_event_source_mapping(
@@ -2065,7 +2121,9 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         temp_params = {}  # values only set for the returned response, not saved internally (e.g. transient state)
 
         # Validate the newly updated ESM object. We ignore the output here since we only care whether an Exception is raised.
-        function_arn, _, _ = self.validate_event_source_mapping(context, event_source_mapping)
+        function_arn, _, _, function_version, function_role = self.validate_event_source_mapping(
+            context, event_source_mapping
+        )
 
         # remove the FunctionName field
         event_source_mapping.pop("FunctionName", None)
@@ -2090,8 +2148,6 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         state.event_source_mappings[uuid] = event_source_mapping
 
         # TODO: Currently, we re-create the entire ESM worker. Look into approach with better performance.
-        function_version = get_function_version_from_arn(function_arn)
-        function_role = function_version.config.role
         worker_factory = EsmWorkerFactory(
             event_source_mapping, function_role, request.get("Enabled", esm_worker.enabled)
         )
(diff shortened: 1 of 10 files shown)


eLRuLL:

eLRuLL:
recheck

bentsku (changes requested):
Hello @eLRuLL and thanks a lot for your contribution!

Nice extra validation of Event Source Mappings! 🚀 this will help users by failing early in the process, and provide more clarity. 

It seems it broke the behavior of Event Source Mappings: there are 43 broken integration tests, and another one in bootstrap that also seems related to the change: it looks like it is not possible to create new valid Event Source Mapping of certain type anymore. 

I also have comments about the access to other services, we usually tend to manage inter-service communication via the `connect_to` utility, so that we have clear boundaries between services. 

The tests also do not use the `snapshot` fixture and it would be great to add it, as it provides a little bit extra parity at no cost. 

Thanks again for the contribution!

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1921,6 +1928,44 @@ def validate_event_source_mapping(self, context, request):
                     Type="User",
                 )
 
+        def check_service_resource_exists(service, resource_arn):
```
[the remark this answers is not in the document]
eLRuLL: I agree, moving it out, but in the same provider

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1921,6 +1928,44 @@ def validate_event_source_mapping(self, context, request):
...
+                queue = sqs_store.queues.get(resource_arn)
+                if not queue:
+                    raise InvalidParameterValueException(
+                        "The resource that you specified for the EventSourceArn parameter doesn't exist.",
+                        Type="User",
+                    )
+            elif service in ["kinesis"]:
+                account_id = extract_account_id_from_arn(resource_arn)
+                region_name = extract_region_from_arn(resource_arn)
+                kinesis = connect_to(aws_access_key_id=account_id, region_name=region_name).kinesis
```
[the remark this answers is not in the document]
eLRuLL: nice, yep I'll replace that with the `get_internal_client` util

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
```diff
@@ -93,6 +93,30 @@ def _get_lambda_logs_event(function_name, expected_num_events, retries=30):
...
+        with pytest.raises(ClientError) as e:
+            aws_client.lambda_.create_event_source_mapping(
+                FunctionName=function_name,
+                BatchSize=1,
+                StartingPosition="TRIM_HORIZON",
+                EventSourceArn=not_existing_dynamodb_stream_arn,
+                MaximumBatchingWindowInSeconds=1,
+                MaximumRetryAttempts=1,
+            )
+        e.match(InvalidParameterValueException.code)
```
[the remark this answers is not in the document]
eLRuLL: oh ok, I followed those steps, and I was confirming it worked with aws, but I didn't understand I had to use the snapshot fixture, but it makes sense.

I added it now, please let me know if they look ok now

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
...
     # ======= EVENT SOURCE MAPPINGS =========
     # =======================================
+    def check_service_resource_exists(self, service, resource_arn):
+        arn = parse_arn(resource_arn)
+        source_client = get_internal_client(resource_arn)
+        if service in ["sqs", "sqs-fifo"]:
+            try:
+                source_client.get_queue_url(QueueName=arn["resource"])
+            except ClientError as e:
+                if e.response["Error"]["QueryErrorCode"] == "QueueDoesNotExist":
```
bentsku, line 1879: question: why use the `QueryErrorCode` here? Since the error message returned by Lambda returns the SQS error code as `AWS.SimpleQueueService.NonExistentQueue`?

Should we use the regular `Code` one, and also return that in the error message? As it is exactly what is returned by the SQS service, we might not need to hardcode it here, and directly return what SQS is returning.

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
     # =======================================
     # ======= EVENT SOURCE MAPPINGS =========
     # =======================================
+    def check_service_resource_exists(self, service, resource_arn):
+        arn = parse_arn(resource_arn)
+        source_client = get_internal_client(resource_arn)
+        if service in ["sqs", "sqs-fifo"]:
+            try:
+                source_client.get_queue_url(QueueName=arn["resource"])
```
bentsku, line 1877: We should try to use the same operation that AWS is using to verify the queue existence. 

Looking at the documentation, when setting up a queue based ESM, AWS advises to use the `AWSLambdaSQSQueueExecutionRole` role: https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-configure.html


The role has the following permissions:
```json
{
  "Version" : "2012-10-17",
  "Statement" : [
    {
      "Effect" : "Allow",
      "Action" : [
        "sqs:ReceiveMessage",
        "sqs:DeleteMessage",
        "sqs:GetQueueAttributes",
        "logs:CreateLogGroup",
        "logs:CreateLogStream",
        "logs:PutLogEvents"
      ],
      "Resource" : "*"
    }
  ]
}
```

It does not have the `sqs:GetQueueUrl`. (see the possible permissions here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-api-permissions-reference.html)

So we can guess that AWS is probably using `GetQueueAttributes` instead to verify the existence of the queue.

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
     # =======================================
     # ======= EVENT SOURCE MAPPINGS =========
     # =======================================
+    def check_service_resource_exists(self, service, resource_arn):
+        arn = parse_arn(resource_arn)
+        source_client = get_internal_client(resource_arn)
```
bentsku, line 1874: We would need to create the internal client with the same kind of parameters that we create it in the actual event source mapping:
```python
source_client = get_internal_client(
    arn=resource_arn,
    role_arn=<configured-lambda-role-arn>,
    service_principal=ServicePrincipal.lambda_,
    source_arn=<lambda-function-arn>,
)
```

This way, we properly pass required metadata about the origin and the original caller of the inter-service request. You will need the lambda function version role, which is fetched a bit later after validation, so you might need to do a bit of refactoring, and the lambda function ARN.

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
     # =======================================
     # ======= EVENT SOURCE MAPPINGS =========
     # =======================================
+    def check_service_resource_exists(self, service, resource_arn):
```
bentsku, line 1872: nit: we should add type hint and a docstring (to express what exceptions it can raise) to the function
```suggestion
    def check_service_resource_exists(self, service: str, resource_arn: str) -> None:
```

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
...
+        elif service in ["dynamodb"]:
+            try:
+                source_client.describe_stream(StreamArn=resource_arn)
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                    raise InvalidParameterValueException(
+                        f"Stream not found: {resource_arn}",
+                        Type="User",
+                    )
+                raise e
```
bentsku, line 1904: nit: it looks like both case are exactly the same, they probably could be merged into one case

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1962,6 +1998,8 @@ def validate_event_source_mapping(self, context, request):
         fn = state.functions.get(function_name)
         if not fn:
             raise InvalidParameterValueException("Function does not exist", Type="User")
+
+        self.check_service_resource_exists(service, request["EventSourceArn"])
```
bentsku, line 2002: the `EventSourceArn` parameter of the incoming request is not a required parameter: it is not given when the event source is an external Apache Kafka cluster. This would raise a `KeyError`, so it is probably safer to do a check before.
```suggestion
        if source_arn := request.get("EventSourceArn"):
            self.check_service_resource_exists(service, source_arn)
```

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_dynamodbstreams.py
```diff
@@ -93,6 +93,30 @@ def _get_lambda_logs_event(function_name, expected_num_events, retries=30):
...
+        with pytest.raises(ClientError) as e:
+            aws_client.lambda_.create_event_source_mapping(
+                FunctionName=function_name,
+                BatchSize=1,
+                StartingPosition="TRIM_HORIZON",
+                EventSourceArn=not_existing_dynamodb_stream_arn,
+                MaximumBatchingWindowInSeconds=1,
+                MaximumRetryAttempts=1,
+            )
+        snapshot.match("error", e.value.response)
```
bentsku, line 118: very nice use of snapshots! this is really nice, we now have full parity of the exception message, and we can be confident we're in line with AWS 💯

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
...
+            except ClientError as e:
+                if e.response["Error"]["QueryErrorCode"] == "QueueDoesNotExist":
+                    raise InvalidParameterValueException(
+                        "Error occurred while ReceiveMessage. SQS Error Code: AWS.SimpleQueueService.NonExistentQueue. SQS Error Message: The specified queue does not exist.",
+                        Type="User",
+                    )
+                raise e
+        elif service in ["kinesis"]:
+            try:
+                source_client.describe_stream(StreamARN=resource_arn)
```
bentsku, line 1887: Here, this is the right operation 🚀 : https://docs.aws.amazon.com/lambda/latest/dg/services-kinesis-create.html#services-kinesis-eventsourcemapping

The permission needed is `kinesis:DescribeStream`

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1869,39 @@ def update_alias(
...
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                    raise InvalidParameterValueException(
+                        f"Stream not found: {resource_arn}",
+                        Type="User",
+                    )
+                raise e
+        elif service in ["dynamodb"]:
+            try:
+                source_client.describe_stream(StreamArn=resource_arn)
```
bentsku, line 1897: Same here, this is the right operation! https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AWSLambdaDynamoDBExecutionRole.html

bentsku (changes requested):
Thanks for jumping on the comment to fix them! This is really nice, and no tests are broken now! Nice 🚀 

Very nice use of the snapshots too! 

I just have a few comments: one regarding the creation of the internal client, that misses a bit of internal metadata in order to properly identify the request internally, so we can properly verify permission when `LOCALSTACK_ENFORCE_IAM` is enabled. 

Another comment is regarding the SQS operation used to verify the queue existence. It seems AWS is using a different operation (this needed a bit of investigation which is not too clear on AWS side sadly!)

We could also re-use the exception raised by SQS to return the error message in order to rely on SQS sending us the right error message, as it seems passed as-is by AWS too. 

I believe we will be good to go once this is addressed! Thanks again, this is looking great!

localstack-core/localstack/services/lambda_/provider.py
eLRuLL: mostly because it was shorter tbh, but yes, reformatting this

localstack-core/localstack/services/lambda_/provider.py
eLRuLL: very nice analysis on how to determine what method/action to use, somehow I was not looking it that way but it makes total sense, updating it

localstack-core/localstack/services/lambda_/provider.py
eLRuLL: I think it's better to keep them separated to make it more explicit that kinesis and dynamodb are handled differently, and yes they both use the `describe_stream` method, but they are not actually the same method as you can see the method parameters are different (`StreamARN` and `StreamArn`)

bentsku: Fait enough, good catch! This is better 👍

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1870,52 @@ def update_alias(
...
     # =======================================
+    def check_service_resource_exists(
+        self, service: str, resource_arn: str, function_arn: str, function_role_arn: str
+    ):
+        """
+        Check if the service resource exists and if the function has access to it.
+
+        Raises:
+            InvalidParameterValueException: If the service resource does not exist or the function does not have access to it.
+        """
```
bentsku, line 1881: thanks for the doc string, very neat!

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1866,6 +1870,52 @@ def update_alias(
...
+            try:
+                source_client.get_queue_attributes(QueueUrl=arn["resource"])
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "AWS.SimpleQueueService.NonExistentQueue":
+                    raise InvalidParameterValueException(
+                        f"Error occurred while ReceiveMessage. SQS Error Code: {e.response['Error']['Code']}. SQS Error Message: {e.response['Error']['Message']}",
+                        Type="User",
+                    )
+                raise e
+        elif service in ["kinesis"]:
```
bentsku, line 1899: nit: I missed it before but this could be `elif service == "kinesis"` as I don't think we would check other kind of services that way. Not blocking however! and same for `dynamodb`.

bentsku (approved):
LGTM! Thanks a lot for addressing the comments, this looks great and very streamlined. 

I'd like to get @joe4dev approval on this as he is one of the owners of Lambda and I wouldn't want to miss something, but to me this looks good to go! 🚀

joe4dev (approved):
Thank you for improving the DevX around Lambda Event Source mappings and dilligently addressing @bentsku comments 👏👏👏
It's so much clearer to catch issues early rather than allowing for creating broken ESMs 🚀

k-a-il (approved):
Appreciate your contribution to the Lambda ESM validation!

localstack-core/localstack/services/lambda_/provider.py
eLRuLL: oh sorry, I got here late, your team was so quick to merge it 😅

PR #12300, opened 2025-02-24
Fix failing lambda error reporting test

dfangl:
## Motivation
With recent updates to `public.ecr.aws/lambda/python:3.12` it seems AWS ships a newer Lambda RIC client with those images than they use in AWS.

This leads to a disparity between the stacktrace returned by aws, and the one in LS.

As the stack trace line numbers are not really important to get _exactly_ right, for me that is a reasonable compromise for now (other option would be pinning the images).

For now I have skipped the stack trace check - we should check in the future to unskip it once AWS incorporates the changes in their environment as well.

## Changes
* Update the test to python 3.13 (sadly this also has the change, but it is good to keep anyway)
* Skip snapshot verify on the stackTrace

### DIFF
--- a/tests/aws/services/lambda_/test_lambda.py
+++ b/tests/aws/services/lambda_/test_lambda.py
@@ -1777,6 +1777,9 @@ class TestLambdaFeatures:
 
 class TestLambdaErrors:
     @markers.aws.validated
+    # TODO it seems like the used lambda images have a newer version of the RIC than AWS in production
+    # remove this skip once they have caught up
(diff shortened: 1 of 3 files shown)


joe4dev (approved):
Thank you @dfangl for unblocking this line change.

This unblocks https://github.com/localstack/localstack/pull/12002
/cc @gregfurman 


💡 Transforming the line could be an alternative to guard against future similar changes, which circumvents setting our future selves a reminder to re-snapshot this :)

PR #12302, opened 2025-02-24
Feature/EventBridge: Clearn up rule-processing logic

maxhoheiser:
## Motivation
Only process rules that match the given event and its conditions.

## Changes
Move matching of condition outside of process_rule logic.


alexrashed:

PR #12304, opened 2025-02-25
[ESM] Add backoff between poll events calls

gregfurman:
## Motivation
As a part of performance improvements to ESM, this PR introduces backoff in poll-frequency when an event source either returns no records or raises an exception on data retrieval.

By delaying polling on empty data sources and errors, we expect to reduce load on the LocalStack gateway and make more resources available for successful processing events.

## Changes
- Small refactor of `poll_events` to better facilitate changing the poll-frequency duration to include backoff.
- New `EmptyPollResultsException` that is raised when a poll request to an event source returns no data. This is used to signal a poll-miss scenario and allows us to decrease poller frequency for empty resources.
- Add an `empty_boff` that returns backoff (up to 10s) when an event source does not return any data (i.e a `EmptyPollResultsException` is raised).
- Add an `error_boff` that returns backoff (up to 60s) when a poll request to an event source raises an exception (other than `EmptyPollResultsException`).
- Note: All backoffs are reset upon a successful poll.

## Benchmarking

Some benchmarks tests were run to assess the extent to which including backoff would improve gateway processing latency of LocalStack. Note, that each experiment was run twice and aggregated appropriately.

We tested exponential backoff with a multiplication factor `x1.5` (default) per iteration.

The below table shows request count, throughput, and latency information about each experiment.
Experiment | # Requests | Average (ms) | Min (ms) | Max (ms) | RPS
-- | -- | -- | -- | -- | --
Baseline | 48096 | 514.275 | 2 | 2247 | 172.56
With Backoff (x1.5) | 111858 | 237.655 | 1 | 1553 | 373.045

The below table shows the percentiles of request-response latency for each experiment.
Experiment | P(50) (ms) | P(90) (ms) | P(95) (ms) | P(99) (ms) | P(100) (ms)
-- | -- | -- | -- | -- | --
Baseline | 480 | 830 | 935 | 1200 | 1800
With Backoff (x1.5) | 225 | 415 | 485 | 665 | 1400

The below table shows factor-level changes comparing backoff experiments to the baseline.
  | P(50) (ms) | RPS
-- | -- | --
With Backoff (x1.5) | -0.531 | +1.162

<details>
<summary><h3>Expand for Baseline Performance Chart(s)</h3></summary>

#### Baseline Performance Chart 1/2
Performance measurements over 5m run for baseline:
image

#### Baseline Performance Chart 2/2
Performance measurements over 5m run for baseline:
image
</details>

<details>

<summary><h3>Expand for Performance With Backoff Chart(s)</h3></summary>

#### With Backoff (x1.5) Performance Chart 1/2
Performance measurements over 5m baseline for run with backoff:
total_requests_per_second_1740565012 795

#### With Backoff (x1.5) Performance Chart 2/2
Performance measurements over 5m baseline for run with backoff:
total_requests_per_second_1740565698 292

</details>

### No. of Requests

Adding backoff on poll-misses saw significant improvements in latency and throughput in our proxy measurement:

- x1.5 backoff: ~2.2x improvement compared to baseline

### RPS

- Baseline Avg: 172.56 RPS
- With Backoff (x1.5) Avg: 373.05 RPS

### P(50)

- Baseline: 480 ms
- With Backoff (x1.5): 225 ms

### P(95)

- Baseline: 935 ms
- With Backoff (x1.5): 485 ms

### Main Findings & Conclusions:

- RPS and P(50) is less variable/noisy over time when backoff is added.
- P(50) + P(95) graphs: The initial performance hit from overlapping requests lessened slightly when compared to baseline -- likely due to backoff on poll-misses reducing the likelihood of request-clustering.
- While adding backoff on poll-misses showed performance gains, without tracking e2e throughput it is unclear how more aggressive backoff will affect the user experience

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
@@ -5,12 +5,24 @@ from enum import StrEnum
 from localstack.aws.api.lambda_ import (
     EventSourceMappingConfiguration,
 )
-from localstack.services.lambda_.event_source_mapping.pollers.poller import Poller
+from localstack.config import (
+    LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_EMPTY_POLL_SEC,
+    LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC,
+    LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC,
+)
+from localstack.services.lambda_.event_source_mapping.pollers.poller import (
+    EmptyPollResultsException,
+    Poller,
+)
 from localstack.services.lambda_.invocation.models import LambdaStore, lambda_stores
 from localstack.services.lambda_.provider_utils import get_function_version_from_arn
+from localstack.utils.backoff import ExponentialBackoff
 from localstack.utils.threads import FuncThread
 
-POLL_INTERVAL_SEC: float = 1
+POLL_INTERVAL_SEC: float = LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC
+MAX_BACKOFF_POLL_EMPTY_SEC: float = LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_EMPTY_POLL_SEC
+MAX_BACKOFF_POLL_ERROR_SEC: float = LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC
+
 
 LOG = logging.getLogger(__name__)
 
@@ -133,13 +145,31 @@ class EsmWorker:
             self.update_esm_state_in_store(EsmState.ENABLED)
             self.state_transition_reason = self.user_state_reason
 
+        error_boff = ExponentialBackoff(initial_interval=2, max_interval=MAX_BACKOFF_POLL_ERROR_SEC)
+        empty_boff = ExponentialBackoff(initial_interval=1, max_interval=MAX_BACKOFF_POLL_EMPTY_SEC)
+
+        poll_interval_duration = POLL_INTERVAL_SEC
+
         while not self._shutdown_event.is_set():
             try:
-                self.poller.poll_events()
                 # TODO: update state transition reason?
-                # Wait for next short-polling interval
-                # MAYBE: read the poller interval from self.poller if we need the flexibility
-                self._shutdown_event.wait(POLL_INTERVAL_SEC)
+                self.poller.poll_events()
+
+                # If no exception encountered, reset the backoff
+                error_boff.reset()
+                empty_boff.reset()
+
+                # Set the poll frequency back to the default
+                poll_interval_duration = POLL_INTERVAL_SEC
+            except EmptyPollResultsException as miss_ex:
+                # If the event source is empty, backoff
+                poll_miss_boff_duration = empty_boff.next_backoff()
+                LOG.debug(
+                    "The event source %s is empty. Backing off for for %s seconds until next request.",
+                    miss_ex.source_arn,
+                    poll_miss_boff_duration,
+                )
+                poll_interval_duration = empty_boff.next_backoff()
             except Exception as e:
                 LOG.error(
                     "Error while polling messages for event source %s: %s",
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
```diff
@@ -148,9 +172,10 @@ def poller_loop(self, *args, **kwargs):
                     e,
                     exc_info=LOG.isEnabledFor(logging.DEBUG),
                 )
-                # TODO: implement some backoff here and stop poller upon continuous errors
```
joe4dev, line 151: strictly speaking, we could parity test when pollers are shut down at AWS after continous errors. However, I think the current solution with the up to 60s max error backoff is good enough (or maybe even preferrable in a dev environment to make the error visible).

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
```diff
@@ -133,13 +140,30 @@ def poller_loop(self, *args, **kwargs):
             self.update_esm_state_in_store(EsmState.ENABLED)
             self.state_transition_reason = self.user_state_reason
 
+        error_boff = ExponentialBackoff(initial_interval=2, max_interval=MAX_BACKOFF_POLL_ERROR_SEC)
```
joe4dev, line 143: super nit: any particular reason why one of the two parameters is a constant and the other not :)

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
```diff
@@ -148,9 +172,10 @@ def poller_loop(self, *args, **kwargs):
                     e,
                     exc_info=LOG.isEnabledFor(logging.DEBUG),
                 )
-                # TODO: implement some backoff here and stop poller upon continuous errors
                 # Wait some time between retries to avoid running into the problem right again
-                self._shutdown_event.wait(2)
+                poll_interval_duration = error_boff.next_backoff()
```
joe4dev, line 176: Much nicer than static waits 👏👏👏

joe4dev (approved):
Wow, these results look very promising 🤩 Thorough PR documentation 👏 

Great and clean changes 💯 


Generally, we should remember to trigger the ext Pipes tests upon ESM poller changes (to be on the safe side shortly before release). These changes should be safe for Pipes because the `EmptyPollResultsException` inherits from `Exception`.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
```diff
@@ -131,6 +132,8 @@ def poll_events(self) -> None:
                     e,
                     exc_info=LOG.isEnabledFor(logging.DEBUG),
                 )
+        else:
+            raise EmptyPollResultsException(service="sqs", source_arn=self.source_arn)
```
tiurin, line 136: nit: for easier maintenance in the future I'd suggest to invert condition to have fewer nested blocks, e.g.:
```python
messages = response.get("Messages")
if not messages:
    raise EmptyPollResultsException(service="sqs", source_arn=self.source_arn)
LOG.debug("Polled %d events from %s", len(messages), self.source_arn)
    try:
...
```
(aware that inverted conditions may also be harder to read, so not insisting on the change, up to you)

tiurin: Just noticed that you actually do inverted condition in stream poller, would be nice to unify, wdyt?

tiurin (approved):
Fantastic improvement! And thanks for exhaustive PR description, made it very easy to review! 👍 

Beyond this PR, I'd like to discuss in the future whether such polling results are exception or polling methods deserve a return value, let's discuss outside PR!

localstack-core/localstack/services/lambda_/event_source_mapping/esm_worker.py
gregfurman: I actually wanted to make these values configurable via environment variables (see this commit https://github.com/localstack/localstack/pull/12304/commits/451ce151b33c1ee678121498bb377eecb6b67c49)

Didn't think to make the initial interval configurable though 🫠

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
gregfurman: Yeah happy with that. Perhaps in a follow-up since I don't think we have time for a CI re-run

joe4dev:
> Beyond this PR, I'd like to discuss in the future whether such polling results are exception or polling methods deserve a return value, let's discuss outside PR!

That's a very good point @tiurin to keep in mind. @gregfurman brought up this idea as well a while ago proposing a typed return structure, which could enable passing metadata such as errors or trace context. I think it's a good idea and would most likely be needed for efficient tracing.

PR #12307, opened 2025-02-26
[ESM] Remove warning when worker shutdown is graceful

gregfurman:
## Motivation
Removes an unnecessary log on shutdown with event source mappings.

## Changes
- Adds a flag to check if shutdown was graceful before logging an invalid state error.


dfangl (approved):
It's hacky, but seems to be the best solution for now.

PR #12312, opened 2025-02-27
[ESM] Correctly enable exponential backoff inside stream poller

gregfurman:
## Motivation
A follow-up PR to https://github.com/localstack/localstack/issues/12281 that correctly initialises the exponential backoff object.

## Changes
- Fixes incorrectly passing the `attempt` counter to the boff object. Instead, the `MaximumRetryAttempts` attribute of the ESM config is used.


dfangl:
I would classify this as a "minor" change, since it effectively activates the changes in #12281, which was also minor.

dfangl (approved):
Let's fix it!

gregfurman:
@dfangl This minor label job is still failing. Let me rebase and push

PR #12313, opened 2025-02-27
Add Lambda@Edge service principal

joe4dev:
## Motivation

We are missing a service principal definition for `edgelambda.amazonaws.com` used at Lambda@Edge https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-permissions.html

## Changes

* Add `edgelambda` to the list of service principals

### DIFF
--- a/localstack-core/localstack/utils/aws/client_types.py
+++ b/localstack-core/localstack/utils/aws/client_types.py
@@ -272,6 +272,7 @@ class ServicePrincipal(str):
     apigateway = "apigateway"
     cloudformation = "cloudformation"
     dms = "dms"
+    edgelambda = "edgelambda"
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM! I wonder if we could try to find a list somewhere of all possible principals 😄

joe4dev:
> LGTM! I wonder if we could try to find a list somewhere of all possible principals 😄

That might be a starting point (unofficial list): https://gist.github.com/shortjared/4c1e3fe52bdfa47522cfe5b41e5d6f22
I guess the question is should we add them proactively (makes it easier to add) or reactively (the list reflects what's used) :)

PR #6672, opened 2022-08-15
add caching for service catalog indexes

thrau:
This PR adds caching of the indexes created in the `ServiceCatalog`.

## Problem

The `ServiceCatalog` is what powers the `ServiceNameParser` by providing indexes that allow fast lookup of service indicators to services (e.g., finding the service by its signing name or endpoint prefix). Creating the indexes involves loading all 300-something services, and iterating over all 11k-something operations. This takes around 1500 ms on my machine. This happens the first time an HTTP call is made that triggers the `ServiceNameParser`, making the first call take at least 1500ms.

## Solution

@alexrashed raised some concerns in the past about shipping an index pre-built with the distribution, which was a valid point since that would involve an error-prone build step. So instead I opportunistically create the cache and store it in the localstack volume directory cache. I restructured the service catalog a bit to separate concerns (providing access to the index, vs actually holding the data). I added both the localstack and botocore version identifier to the file name so we always re-load the cache on new versions.

## Result

Loading the cache takes 8ms vs. creating it lazily which takes about 1500ms. This has a dramatic effect on cold starts, specifically the first invocation of a service now only loads the service model, which is much faster (previously we would also load *all* service models to build the index).

Here's me executing `awslocal opensearch list-domains`. "process" is the end-to-end server latency (timer around `Gateway.process`, and `by_signing_name` is what's being executed to look up the service):

Without caching:
```
# first invocation
Execution of "by_signing_name" took 1458.67ms
Execution of "process" took 1562.29ms
# second invocation
Execution of "by_signing_name" took 0.00ms
Execution of "process" took 23.25ms
```

With caching:
```
# first invocation
Execution of "by_signing_name" took 0.01ms
Execution of "process" took 127.29ms
# second invocation
Execution of "by_signing_name" took 0.00ms
Execution of "process" took 22.81ms
```

I also tried cold-starting several different services in a row. There's only a measurable effect for the first one. 
But a pretty dramatic one: 10x speedup for the first cold start.

### Drawbacks/Limitations

The index obviously has to be computed at some point. Previously we would do that on the first invocation of any HTTP request for *every* localstack run. Now we do it *once* when localstack starts up for the very first time. Any subsequent localstack runs will have no startup delay. This means that this PR has no effect on ephemeral instances of localstack (e.g., in CI environments). For that we would have to ship the index with the distribution.

PR #12314, opened 2025-02-27
add service catalog cache loading from static var

alexrashed:
## Motivation
With the release of LocalStack 4.1.0, we had issues with the smoke tests of the binary build of our CLI.
It turned out that the initial loading of the service catalog took more than 2 seconds (which is the default connection timeout of `localstack wait`). https://github.com/localstack/localstack-cli/pull/33 introduced a workaround for this, by triggering the service catalog cache creation with a call prior to the `localstack wait` command.
However, this issue has a user impact, as this cache will always be created when localstack is started with a new `botocore` version (even if the localstack volume dir is properly configured).
In fact, this was already discussed with @thrau when he introduced the cache in the first place (https://github.com/localstack/localstack/pull/6672).
This PR solves this issue by generating the service catalog index cache at docker build time and stores it directly in the Docker image (~650kb uncompressed). 

## Changes
- Moves `get_service_catalog` from the `service_router` to the `spec` module.
- Introduces a new lookup directory (highest prio): The static libs directory (usually `/usr/lib/localstack/`).
- Adds a new `main` to the spec which allows to easily generate a `ServiceCatalog` index cache file to the static libs dir.
- Adds an invocation of this command to the Dockerfile

## Testing
- [ ] Update our downstream dependencies to also generate the service catalog as part of their Docker build process.

### DIFF
--- a/localstack-core/localstack/aws/spec.py
+++ b/localstack-core/localstack/aws/spec.py
@@ -2,14 +2,20 @@ import dataclasses
 import json
 import logging
 import os
+import sys
 from collections import defaultdict
 from functools import cached_property, lru_cache
 from typing import Dict, Generator, List, Literal, NamedTuple, Optional, Tuple
 
+import botocore
 import jsonpatch
 from botocore.exceptions import UnknownServiceError
 from botocore.loaders import Loader, instance_cache
 from botocore.model import OperationModel, ServiceModel
+
+from localstack import config
+from localstack.constants import VERSION
+from localstack.utils.objects import singleton_factory
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


localstack-core/localstack/aws/spec.py
```diff
@@ -303,5 +309,61 @@ def save_service_index_cache(index: LazyServiceCatalogIndex, file_path: str) ->
...
+def get_service_catalog() -> ServiceCatalog:
+    """Loads the ServiceCatalog (which contains all the service specs), and potentially re-uses a cached index."""
+
+    try:
+        catalog_file_name = _get_catalog_filename()
+        static_catalog_file = os.path.join(config.dirs.static_libs, catalog_file_name)
+
+        # try to load or load/build/save the service catalog index from the static libs
+        index = None
+        if os.path.exists(static_catalog_file):
```
silv-io, line 333: If the current localstack version or botocore version is different, then the static cache should be invalid, no?

This could happen when mounting a different localstack version into the container with the container dev mode I think

silv-io (approved):
LGTM! Just a small question about cache validity

localstack-core/localstack/aws/spec.py
silv-io: nvm, it would look up a different directory then because the filename is generated based on the current version. was confused at first

alexrashed: Also the filename changed ;)

PR #12319, opened 2025-02-28
[ESM] Fix constantly triggering SQS back-off

gregfurman:
## Motivation
This fixes an issue that causes our SQS poller to always trigger exponential backoff between poller calls instead of only when a response is empty or exception is raised. Currently, this bug is slowing down performance and should be merged ASAP.

### Why did this happen?

GitHub's automatic rebase (which did not raise a merge conflict) caused our `if-else` conditional to change into a `for-else`. Since the `else` block was where we raised a backoff exception, this has caused our poller to _always_ raise an exception and signals the ESM worker to constantly back-off.

See: 

https://github.com/localstack/localstack/blob/ceb6140362a2dcabd2e687344a399ccbeba6dab9/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py#L196

And:

https://github.com/localstack/localstack/blob/8859590c9974262a222c486ce7aaaeea151cd70f/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py#L194

## Changes
- Moves the backoff exception from always being raised to instead be within the conditional check for no records being found.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
@@ -160,7 +160,6 @@ class SqsPoller(Poller):
 
         messages = response.get("Messages", [])
         if not messages:
-            # TODO: Consider this case triggering longer wait-times (with backoff) between poll_events calls in the outer-loop.
             raise EmptyPollResultsException(service="sqs", source_arn=self.source_arn)
 
         LOG.debug("Polled %d events from %s", len(messages), self.source_arn)
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py
@@ -161,7 +161,7 @@ class SqsPoller(Poller):
         messages = response.get("Messages", [])
         if not messages:
             # TODO: Consider this case triggering longer wait-times (with backoff) between poll_events calls in the outer-loop.
-            return
+            raise EmptyPollResultsException(service="sqs", source_arn=self.source_arn)
 
         LOG.debug("Polled %d events from %s", len(messages), self.source_arn)
         # TODO: implement invocation payload size quota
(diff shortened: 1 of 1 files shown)


gregfurman:
@joe4dev There is a flaky test (both against AWS and LocalStack) that can be explained by this line:

> If you're using a batch window and your SQS queue contains very low traffic, Lambda might wait for up to 20 seconds before invoking your function. This is true even if you set a batch window lower than 20 seconds. 

https://github.com/localstack/localstack/blob/4285fb18cf8e2b79298634fd714177c83b984843/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py#L392-L394

Basically, we cannot assume that an ESM will always invoke a target Lambda with a very precise timeframe -- where AWS (and now LocalStack) dynamically adjusts this frequency based on traffic to the data source.

I think the behavioural changes should remain, but we should keep an eye out for tests that assume invocations happen at precise intervals.

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py
```diff
@@ -412,19 +412,26 @@ def test_report_batch_item_failures(
         "get_destination_queue_url", aws_client.sqs.get_queue_url(QueueName=destination_queue_name)
     )
 
-    # timeout in seconds, used for both the lambda and the queue visibility timeout.
-    # increase to 10 if testing against AWS fails.
+    # If an SQS queue is not receiving a lot of traffic, Lambda can take up to 20s between invocations.
+    # See AWS docs https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html.
     retry_timeout = 8
```
joe4dev, line 417: We could consider reducing the retry timeout (e.g., 6s) to mitigate the change of simultaneous processing). Having the assert `visibility_timeout > retry_timeout` would be safer

gregfurman: Separated changes into https://github.com/localstack/localstack/pull/12323

PR #12324, opened 2025-03-03
APIGW: add validation for AWS ARN in PutIntegration

bentsku:
## Motivation
We've got a report from a user with an unhandled error in the invocation logic:
```python
localstack | File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/apigateway/next_gen/execute_api/helpers.py", line 114, in get_lambda_function_arn_from_invocation_uri
localstack | return uri.split("functions/")[1].removesuffix("/invocations")
localstack | ~~~~~~~~~~~~~~~~~~~~~~~^^^
localstack | IndexError: list index out of range
```

This is probably because the Integration URI passed was directly a Lambda ARN, and it needs to be an API Gateway invocation ARN:
Correct format:
`arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:123456789012:function:HelloWorld/invocations`
Incorrect format:
`arn:aws:lambda:us-east-1:123456789012:function:HelloWorld`

This PR implements a bit of validation to avoid this. 

Also had to move a bit of validation already done in moto in the provider just to keep proper validation order (with `integrationHttpMethod`)

## Changes
- add a test around the URI format for AWS/AWS_PROXY integration
- add a bit of validation for `PutIntegration`

### DIFF
--- a/localstack-core/localstack/services/apigateway/legacy/provider.py
+++ b/localstack-core/localstack/services/apigateway/legacy/provider.py
@@ -122,7 +122,7 @@ from localstack.services.apigateway.patches import apply_patches
 from localstack.services.edge import ROUTER
 from localstack.services.moto import call_moto, call_moto_with_request
 from localstack.services.plugins import ServiceLifecycleHook
-from localstack.utils.aws.arns import get_partition
+from localstack.utils.aws.arns import InvalidArnException, get_partition, parse_arn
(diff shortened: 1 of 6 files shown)


cloutierMat (approved):
Great provider validation upgrade, and super clean tests, as usual! :rocket:

PR #12338, opened 2025-03-04
ESM: fix CreateESM SQS validation

bentsku:
## Motivation
Follow up from #12297, we did not pass the `QueueUrl` to `GetQueueAttributes` but rather the `QueueArn`. This actually still worked somehow because SQS took it as the `QueueName` and was able to resolve it. 
However, when working with cross-account, this did not work because we were not able to resolve to the right account id. 

I thought of doing an internal `connect_to().sqs.get_queue_url(QueueName=name, QueueOwnerAwsAccountId=queue_account)`, to get the proper `QueueUrl`, but if the queue did not exist, it would have raised without triggering an IAM check for `GetQueueAttributes`, so the IAM parity would not be there... so I resolved to build a dummy value instead, as it is not very important and we only need SQS to be able to parse it. 

Upstream run: ✅  13658265278

## Changes
- build a dummy queue url from the queue ARN to pass to `GetQueueAttributes`.

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -1892,11 +1892,17 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
         )
         if service in ["sqs", "sqs-fifo"]:
             try:
-                source_client.get_queue_attributes(QueueUrl=arn["resource"])
+                # AWS uses `GetQueueAttributes` internally to verify the queue existence, but we need the `QueueUrl`
+                # which is not given directly. We build out a dummy `QueueUrl` which can be parsed by SQS to return
+                # the right value
+                queue_name = arn["resource"].split("/")[-1]
+                queue_url = f"http://sqs.{arn['region']}.domain/{arn['account']}/{queue_name}"
+                source_client.get_queue_attributes(QueueUrl=queue_url)
             except ClientError as e:
-                if e.response["Error"]["Code"] == "AWS.SimpleQueueService.NonExistentQueue":
+                error_code = e.response["Error"]["Code"]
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1892,11 +1892,17 @@ def check_service_resource_exists(
         )
         if service in ["sqs", "sqs-fifo"]:
             try:
-                source_client.get_queue_attributes(QueueUrl=arn["resource"])
+                # AWS uses `GetQueueAttributes` internally to verify the queue existence, but we need the `QueueUrl`
+                # which is not given directly. We build out a dummy `QueueUrl` which can be parsed by SQS to return
+                # the right value
+                queue_name = arn["resource"].split("/")[-1]
+                queue_url = f"http://sqs.{arn['region']}.domain/{arn['account']}/{queue_name}"
```
joe4dev, line 1899: Interesting hack. I wonder how AWS does it internally because [GetQueueUrl](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueUrl.html) is not a required IAM permission: https://repost.aws/knowledge-center/lambda-sqs-event-source

joe4dev (approved):
Thank you for coming up with this innovative workaround @bentsku 🧠 

what a coincidenc that it worked in the first place 😅

localstack-core/localstack/services/lambda_/provider.py
bentsku: Yes, I really wonder too.. maybe just rebuilding the queue manually like us, but it seems… weird 😅 I don’t know if there would be any other operation that sqs:getQueueAttributes permission would give access to. Thanks for the review!

Issue #10723, opened 2024-04-25
enhancement request: RotateKeyOnDemand


k-wall:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Enhancement description

I am writing an application which needs to test scenarios when keys are rotated.   In the application's tests (which run against LocalStack) I wish to use RotateKeyOnDemand, but this API is not implemented.

Could an implementation of RotateKeyOnDemand be considered for a later release?

k-wall:
The issue still exists and it is still important to the project I contribute to.
https://docs.localstack.cloud/references/coverage/coverage_kms/

k-wall:
https://github.com/localstack/localstack/releases/tag/v4.3.0 says:

> LocalStack now supports on-demand key rotation for KMS symmetric keys without imported key material, allowing up to 10 rotations per key that don’t affect automatic rotation schedules.

PR #12103, opened 2025-01-05
fix kms api call response on method get-key-rotation-status

pureiboi:
## Motivation
This PR will resolve issue https://github.com/localstack/localstack/issues/12080

## Changes

The response payload as per AWS cli, and request payload used as per issue 

```

{
    "KeyRotationEnabled": true,
    "KeyId": "46d57f31-b654-4cba-bfb5-fa5847724beb",
    "RotationPeriodInDays": 120,
    "NextRotationDate": "2025-05-06T03:16:36.190688+08:00"
}

```

image

PR #12342, opened 2025-03-05
KMS: on-demand key rotation

agseco:
## Motivation

Motivated by this enhancement request: https://github.com/localstack/localstack/issues/10723

## Changes

Initial implementation of the [on-demand key rotation feature in KMS](https://docs.aws.amazon.com/kms/latest/developerguide/rotating-keys-on-demand.html).

## TODO

What's left to do:

- [ ] Preserve old key material, as specified in AWS's docs
- [ ] Consider multi-region keys, if necessary

### DIFF
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -92,6 +92,8 @@ from localstack.aws.api.kms import (
     ReEncryptResponse,
     ReplicateKeyRequest,
     ReplicateKeyResponse,
+    RotateKeyOnDemandRequest,
+    RotateKeyOnDemandResponse,
     ScheduleKeyDeletionRequest,
     ScheduleKeyDeletionResponse,
     SignRequest,
@@ -1333,6 +1335,26 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
         )
         kwargs = {"NextMarker": next_token, "Truncated": True} if next_token else {}
         return ListResourceTagsResponse(Tags=page, **kwargs)
+
+    @handler("RotateKeyOnDemand", expand=False)
+    def rotate_key_on_demand(
+        self, context: RequestContext, request: RotateKeyOnDemandRequest
+    ) -> RotateKeyOnDemandResponse:
+        account_id, region_name, key_id = self._parse_key_id(request["KeyId"], context)
+        key = self._get_kms_key(account_id, region_name, key_id, any_key_state_allowed=True)
+
+        if key.metadata["KeySpec"] != KeySpec.SYMMETRIC_DEFAULT:
+            raise UnsupportedOperationException(
+                "On-demand key rotation is supported only on symmetric encryption KMS keys"
+            )
+
+        print(f"Before: {key.crypto_key.key_material}")
+        key.crypto_key = KmsCryptoKey(KeySpec.SYMMETRIC_DEFAULT)
+        print(f"After: {key.crypto_key.key_material}")
+
+        return RotateKeyOnDemandResponse(
+            KeyId=key_id,
+        )
 
     @handler("TagResource", expand=False)
     def tag_resource(self, context: RequestContext, request: TagResourceRequest) -> None:
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -1099,6 +1099,14 @@ class TestKMS:
         assert aws_client.kms.get_key_rotation_status(KeyId=key_id)["KeyRotationEnabled"] is False
 
     @markers.aws.validated
+    def test_rotate_key_on_demand_succeeds_given_symmetric_key(self, kms_key, aws_client):
+        raise NotImplementedError
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_returns_error_given_non_symmetric_key(self, kms_key, aws_client):
+        raise NotImplementedError
+
+    @markers.aws.validated
     @pytest.mark.parametrize("rotation_period_in_days", [90, 180])
     def test_key_enable_rotation_status(
         self,


agseco:

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,19 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
         assert aws_client.kms.get_key_rotation_status(KeyId=key_id)["KeyRotationEnabled"] is False
 
+    @markers.aws.validated
+    def test_rotate_key_on_demand_succeeds_given_symmetric_key(self, kms_key, aws_client):
+        key_id = kms_key["KeyId"]
+        response = aws_client.kms.rotate_key_on_demand(KeyId=key_id)
+        assert response["KeyId"] == key_id
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_returns_error_given_non_symmetric_key(self, kms_create_key, aws_client):
```
agseco, line 1108: Here I'm testing just a single occurrence of non-symmetrical key, I have considered it's not strictly required to test them all, but could be done if you deem it necessary :)

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,19 @@ def test_key_rotation_status(self, kms_key, aws_client):
         aws_client.kms.disable_key_rotation(KeyId=key_id)
         assert aws_client.kms.get_key_rotation_status(KeyId=key_id)["KeyRotationEnabled"] is False
 
+    @markers.aws.validated
+    def test_rotate_key_on_demand_succeeds_given_symmetric_key(self, kms_key, aws_client):
+        key_id = kms_key["KeyId"]
+        response = aws_client.kms.rotate_key_on_demand(KeyId=key_id)
```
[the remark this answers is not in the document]
agseco: What I can also do is to ensure that the the original rotation period remains unchanged, as specified in AWS's docs.

For this #12103 will come handy @k-a-il 😄

tests/aws/services/kms/test_kms.py
agseco, line 1108: Here I'm just testing a single non-symmetrical key. At this stage I think it's not strictly required to test them all (RSA, HMAC, etc..). Let me know if you think differently.

tests/aws/services/kms/test_kms.py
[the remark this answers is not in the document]
k-a-il: It's important to check that the key material's status is different before and after the change. However, we want to avoid accessing the internal store directly. 
Focus on validating the observable effects of the change rather than the underlying data itself.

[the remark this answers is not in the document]
agseco: Thanks for the feedback @k-a-il.

I have implemented it, also I have divided this test cases in two:

1. Key with disabled automatic rotation (default behaviour)
2. Key with enabled automatic rotation

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1334,6 +1336,28 @@ def list_resource_tags(
...
+        key = self._get_kms_key(account_id, region_name, key_id, any_key_state_allowed=True)
+
+        if key.metadata["KeySpec"] != KeySpec.SYMMETRIC_DEFAULT:
+            raise UnsupportedOperationException(
+                "On-demand key rotation is supported only on symmetric encryption KMS keys"
+            )
+        if key.metadata["Origin"] == OriginType.EXTERNAL:
+            raise UnsupportedOperationException(
+                "On-demand key rotation is not supported for keys with imported key material"
+            )
```
agseco, line 1353: Is it important to provide the very same error messages that AWS provides?

In affirmative case, it'd be an impediment for me  sinceI don't have an AWS account to test it.

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,95 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_raises_error_given_key_that_does_not_exist(
+        self, kms_create_key, aws_client
+    ):
+        key_id = "1234abcd-12ab-34cd-56ef-1234567890ab"
+
+        with pytest.raises(ClientError) as exc:
+            aws_client.kms.rotate_key_on_demand(KeyId=key_id)
+        assert exc.match("NotFoundException")
```
agseco, line 1168: This test may not strictly required, but I added to comply with the "contract" specified in [the docs](https://docs.aws.amazon.com/kms/latest/APIReference/API_RotateKeyOnDemand.html#API_RotateKeyOnDemand_Errors).

tests/aws/services/kms/test_kms.py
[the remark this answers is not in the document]
agseco: This just came to my mind: in order to verify in the test that the key material has effectively changed, would it make sense to try to encrypt something before and after and compare the results?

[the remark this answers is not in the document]
agseco: Just pushed it.

localstack-core/localstack/services/kms/models.py
```diff
@@ -249,6 +249,7 @@ class KmsKey:
     is_key_rotation_enabled: bool
     rotation_period_in_days: int
     next_rotation_date: datetime.datetime
+    on_demand_rotation_start_date: datetime.datetime
```
agseco, line 252: I had to add this new attribute, as AWS returns this attribute in `GetKeyRotationStatusResponse` when `RotateKeyOnDemand` action is executed.

tests/aws/services/kms/test_kms.snapshot.json
```diff
@@ -2035,5 +2035,125 @@
...
+  },
+  "tests/aws/services/kms/test_kms.py::TestKMS::test_rotate_key_on_demand_raises_error_given_non_symmetric_key": {
+    "recorded-date": "08-03-2025, 09:27:44",
+    "recorded-content": {
+      "error-response": {
+        "Error": {
+          "Code": "UnsupportedOperationException",
+          "Message": ""
+        },
+        "message": "",
```
agseco, line 2135: I temporarily added `"message": "",` because I didn't know how to solve this conflict; AWS doesn't return it in the response but LocalStack does.

It seems to me it's not something to do with my implementation, but rather some core behavior and would require more time and effort investment to solve it.

Probably you came across similar issues and we can easily solve it in the tech debrief session.

localstack-core/localstack/services/kms/provider.py
agseco: Solved.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1258,7 +1260,10 @@ def get_key_rotation_status(
             KeyId=key_id,
             KeyRotationEnabled=key.is_key_rotation_enabled,
             NextRotationDate=key.next_rotation_date,
-            RotationPeriodInDays=key.rotation_period_in_days,
+            OnDemandRotationStartDate=key.on_demand_rotation_start_date,
+            RotationPeriodInDays=key.rotation_period_in_days
+            if key.is_key_rotation_enabled
+            else None,
```
agseco, line 1266: I was forced to had to do this since the model sets a default value `rotation_period_in_days` but `GetKeyRotationStatus` **should only return this attribute in the response when the automatic rotation is enabled**.

localstack-core/localstack/services/kms/models.py
k-a-il: I see, that makes sense. However, I'm not sure whether `on_demand_rotation_start_date` is always returned or only when the on-demand rotation is in progress. I believe it's the latter because the documentation states: 
[OnDemandRotationStartDate](https://docs.aws.amazon.com/kms/latest/APIReference/API_GetKeyRotationStatus.html#API_GetKeyRotationStatus_ResponseSyntax)
> identifies the date and time when an in-progress on-demand rotation was initiated. 

Have you tested the AWS API to check whether this status is still returned even after the rotation is completed?

localstack-core/localstack/services/kms/provider.py
k-a-il: > The model sets a default value for rotation_period_in_days but GetKeyRotationStatus should only return this attribute in the response when the automatic rotation is enabled.

Indeed, this looks like a bug. Thank you for addressing this

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,114 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+        key_id = kms_create_key(KeyUsage="ENCRYPT_DECRYPT", KeySpec="RSA_4096")["KeyId"]
+        aws_client.kms.disable_key(KeyId=key_id)
+
+        with pytest.raises(ClientError) as e:
+            aws_client.kms.rotate_key_on_demand(KeyId=key_id)
+        snapshot.match("error-response", e.value.response)
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_raises_error_given_key_that_does_not_exist(
+        self, kms_create_key, aws_client, snapshot
```
k-a-il, line 1181: `kms_create_key` fixture is not being used

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,114 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+            KeyId=key_id,
+            Plaintext=base64.b64encode(message),
+            EncryptionAlgorithm="SYMMETRIC_DEFAULT",
+        )["CiphertextBlob"]
+
+        assert ciphertext_before != ciphertext_after
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_with_symmetric_key_and_automatic_rotation_disabled_preserves_automatic_rotation_schedule_attributes(
+        self, kms_key, aws_client, snapshot
```
k-a-il, line 1125: Why did you decide to use here `kms_key` fixture instead of using `kms_create_key` as in the test above? I'm concerned that if the fixture is later modified to create non-symmetric keys, this test may start failing

k-a-il (changes requested):
Congratulations on your first contribution! Great work, and thank you for adding integration tests validated against AWS.

localstack-core/localstack/services/kms/models.py
bentsku: Adding on to this, I believe we could probably create a new attribute in the store itself instead, to keep track on all "ongoing" rotations. That way, we could also re-use it for `ListKeyRotations` without having to iterate over all existing keys to verify for their status.

tests/aws/services/kms/test_kms.snapshot.json
bentsku: Do you mean you modified the snapshot manually? It would be good to revert this, then. 
We can manually skip some fields that are missing (or returned by LocalStack and not AWS) by using the @markers.snapshot.skip_snapshot_verify` marker instead. 

It indeed looks like a deeper issue with our serializer, something maybe a bit more tricky to address.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1258,7 +1260,10 @@ def get_key_rotation_status(
             KeyId=key_id,
             KeyRotationEnabled=key.is_key_rotation_enabled,
             NextRotationDate=key.next_rotation_date,
-            RotationPeriodInDays=key.rotation_period_in_days,
+            OnDemandRotationStartDate=key.on_demand_rotation_start_date,
```
bentsku, line 1263: Is the `OnDemandRotationStartDate` returned when there was no on-demand rotation initiated? this is probably linked with the comment under.

localstack-core/localstack/services/kms/provider.py
bentsku: Manually returning `None` is not usually the best way to indicates that a field is not returned. It will depends on the "protocol" on the service (REST-JSON, Query, REST-XML, there are a few) and its serializer. In the case of KMS, it prevents the field being returned in the response, but for the S3 service for example it will return `null` instead. 

But it would be better to manually add the field if present. An example:
```python
response =  GetKeyRotationStatusResponse(
            KeyId=key_id,
            KeyRotationEnabled=key.is_key_rotation_enabled,
            NextRotationDate=key.next_rotation_date,
            OnDemandRotationStartDate=key.on_demand_rotation_start_date,
            RotationPeriodInDays=key.rotation_period_in_days
            if key.is_key_rotation_enabled
            else None,
        )
if key.is_key_rotation_enabled:
    response["RotationPeriodInDays"] = key.rotation_period_in_days
if key.on_demand_rotation_start_date is not None:
    response["OnDemandRotationStartDate"] = key.on_demand_rotation_start_date

return response
```

That way, it is a bit more explicit that the fields are not returned by AWS if not present.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1334,6 +1339,29 @@ def list_resource_tags(
         kwargs = {"NextMarker": next_token, "Truncated": True} if next_token else {}
         return ListResourceTagsResponse(Tags=page, **kwargs)
 
+    @handler("RotateKeyOnDemand", expand=False)
+    def rotate_key_on_demand(
```
bentsku, line 1343: it would be good to leave a TODO in the code indicating that we are not storing the previous key material, or in some way indicates that that feature is yet supported. 

Calling `RotateKeyOnDemand` in our implementation will prevent the user to decrypt any previously encoded material.

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,114 @@ def test_key_rotation_status(self, kms_key, aws_client):
         aws_client.kms.disable_key_rotation(KeyId=key_id)
         assert aws_client.kms.get_key_rotation_status(KeyId=key_id)["KeyRotationEnabled"] is False
 
+    @markers.aws.validated
+    def test_rotate_key_on_demand_modifies_key_material(self, kms_create_key, aws_client, snapshot):
```
bentsku, line 1102: praise: this is a nice test for validating that the key material changed 👍

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,114 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+
+        assert ciphertext_before != ciphertext_after
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_with_symmetric_key_and_automatic_rotation_disabled_preserves_automatic_rotation_schedule_attributes(
+        self, kms_key, aws_client, snapshot
+    ):
+        key_id = kms_key["KeyId"]
+
+        rotation_status_response_before = aws_client.kms.get_key_rotation_status(KeyId=key_id)
```
bentsku, line 1129: nit: if we were to snapshot this `rotation_status_response_before`, we would see what are the default values when no scheduled or ongoing key rotation would be, and we would not need to do the `assert` under as the snapshot library would do it for us.

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1098,114 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+        snapshot.match("rotation-status-response-after-rotation", rotation_status_response_after)
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_with_symmetric_key_and_automatic_rotation_enabled_preserves_automatic_rotation_schedule_attributes(
+        self, kms_key, aws_client, snapshot
+    ):
+        key_id = kms_key["KeyId"]
+
+        aws_client.kms.enable_key_rotation(KeyId=key_id)
+        rotation_status_response_before = aws_client.kms.get_key_rotation_status(KeyId=key_id)
```
bentsku, line 1148: nit: same as above, we should snapshot this to avoid all the manual assert under

bentsku (changes requested):
Nice work! This is great, thanks for adding a new operation and improving parity! Great set of snapshot tests. 

I have a few comments about the new attribute on the key that could maybe be on the store itself, another about returning `None` values from the provider, and maybe we could add a bit more snapshot match in order to reduce the manual `assert`. 

Thanks again for the very thorough testing! 🚀

tests/aws/services/kms/test_kms.snapshot.json
agseco: Yes did intentionally to better illustrate it in the PR, it was in plans to revert it at some point 😬 

I already reverted it and solved the issue with `@markers.snapshot.skip_snapshot_verify`, thanks for the recommendation!

localstack-core/localstack/services/kms/models.py
```diff
@@ -597,6 +599,9 @@ def _update_key_rotation_date(self) -> None:
                 days=self.rotation_period_in_days
             )
 
+    def _update_on_demand_rotation_start_date(self) -> None:
+        self.on_demand_rotation_start_date = datetime.datetime.now()
+
```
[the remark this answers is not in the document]
agseco: Perhaps I'm biased by DDD, but it made sense to me from a _domain modelling_ point of view.

localstack-core/localstack/services/kms/models.py
agseco: Good catch, AWS explicitly states _in progress_:

> Identifies the date and time that an in progress on-demand rotation was initiated.

I tested it and this is the response right after triggering the rotation:

```
{
    "KeyRotationEnabled": false,
    "KeyId": "arn:aws:kms:eu-central-1:443370710705:key/6c4f6c59-1dff-45ff-827e-9ccb4937a03e",
    "OnDemandRotationStartDate": "2025-03-11T10:42:02.816000+01:00"
}
```

And this a while after:

{
    "KeyRotationEnabled": false,
    "KeyId": "arn:aws:kms:eu-central-1:443370710705:key/6c4f6c59-1dff-45ff-827e-9ccb4937a03e"
}

I guess we should remove it from the response because our implementation is synchronous, but as you point out to internally keep track of it somehow.

Also we will have to ignore `OnDemandRotationStartDate` in the tests, because I added this attribute when one of the tests was failing from the snapshot.

agseco: As agreed in the call, we will:

1. Create a new dictionary in the store to store the on demand rotation date
2. Return the `OnDemandRotationStartDate` in the `GetKeyRotationStatusResponse` only if the previous one is less than x amount of time old

This we will simulate the asynchronous behavior of AWS.

bentsku: I've just read the documentation again, and I got confused by the `ListKeyRotations` command. I initially thought we would be able to list all currently on-going rotations, which lead to some confusing things said from me! I'm really sorry about that. 

Storing the `rotations` on the key model itself should be good, as the `ListKeyRotations` operation is scoped to one key only. That way, you can keep track of the previously done `rotations`. Again, really sorry about the confusion here!

tests/aws/services/kms/test_kms.py
agseco: Inconsistent usage by accident 😅

localstack-core/localstack/services/kms/provider.py
agseco: To be followed up [in this other thread](https://github.com/localstack/localstack/pull/12342#discussion_r1986051876).

localstack-core/localstack/services/kms/models.py
```diff
@@ -693,6 +693,9 @@ def _get_key_usage(self, request_key_usage: str, key_spec: str) -> str:
         else:
             return request_key_usage or "ENCRYPT_DECRYPT"
 
+    def rotate_key_on_demand(self):
+        self.crypto_key = KmsCryptoKey(KeySpec.SYMMETRIC_DEFAULT)
```
agseco, line 697: I moved the on demand key rotation operation to the model; the reason is that it seems that further business logic may be added here, so it the way is already "paved" in this sense.

Let me know if it didn't make sense to you :)

localstack-core/localstack/services/kms/provider.py
```diff
@@ -369,6 +371,30 @@ def _parse_key_id(key_id_or_arn: str, context: RequestContext) -> Tuple[str, str
...
+        ):
+            return True
+        else:
+            return False
+
+    @staticmethod
+    def _on_demand_rotation_timestamp(account_id: str, region_name: str, key_id: str):
+        store = kms_stores[account_id][region_name]
+        return store.on_demand_rotations[key_id]
+
```
agseco, line 397: Here is where the _business logic_ that simulates async execution of the on demand rotation. Since we agreed to keep the rotation date in the store (`KmsStore`), I felt this was a natural place for helper methods.

While I was implementing it, I continued to feel that it may suit better in the model, e.g. using a new attribute in the model to store the rotation date (`on_demand_rotation_start_date`, as it was before).

tests/aws/services/kms/test_kms.py
```diff
@@ -1098,6 +1099,129 @@ def test_key_rotation_status(self, kms_key, aws_client):
...
+        ciphertext_after = aws_client.kms.encrypt(
+            KeyId=key_id,
+            Plaintext=base64.b64encode(message),
+            EncryptionAlgorithm="SYMMETRIC_DEFAULT",
+        )["CiphertextBlob"]
+
+        assert ciphertext_before != ciphertext_after
+
+    @markers.aws.validated
+    def test_rotate_key_on_demand_should_return_rotation_start_date_only_while_its_in_progress(
```
agseco, line 1125: This new test tests the asynchronous behavior simulation.

localstack-core/localstack/services/kms/provider.py
agseco: Also I decided that `2.5s` is a good amount of time, mainly because it may match better AWS's actual behavior.

tests/aws/services/kms/test_kms.py
agseco: Thanks :)

localstack-core/localstack/services/kms/models.py
```diff
@@ -284,6 +284,7 @@ def __init__(
         self.crypto_key = KmsCryptoKey(self.metadata.get("KeySpec"), custom_key_material)
         self.rotation_period_in_days = 365
         self.next_rotation_date = None
+        self.on_demand_rotation_start_date = None
```
agseco, line 287: For simplicity current it's limited to tracking a single on demand rotation.

As part of implementing `ListKeyRotations`, the model can be evolved to track many.

localstack-core/localstack/services/kms/models.py
```diff
@@ -693,6 +694,21 @@ def _get_key_usage(self, request_key_usage: str, key_spec: str) -> str:
...
+    def rotate_key_on_demand(self):
+        self.on_demand_rotation_start_date = datetime.datetime.now()
+        self.crypto_key = KmsCryptoKey(KeySpec.SYMMETRIC_DEFAULT)
+
+    def on_demand_rotation_in_progress(self):
+        if self.on_demand_rotation_start_date is None:
+            return False
+
+        if datetime.datetime.now() < self.on_demand_rotation_start_date + datetime.timedelta(
+            seconds=2.5
```
agseco, line 706: I thought that 2.5s is a good amount of time instead of 1s, mainly because it may match better AWS's actual behavior.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1334,6 +1343,29 @@ def list_resource_tags(
...
+    @handler("RotateKeyOnDemand", expand=False)
+    # TODO: keep trak of key rotations as AWS does and return them in the ListKeyRotations operation
+    def rotate_key_on_demand(
+        self, context: RequestContext, request: RotateKeyOnDemandRequest
+    ) -> RotateKeyOnDemandResponse:
+        account_id, region_name, key_id = self._parse_key_id(request["KeyId"], context)
+        key = self._get_kms_key(account_id, region_name, key_id, any_key_state_allowed=True)
+
+        if key.metadata["KeyState"] == KeyState.Disabled:
+            raise DisabledException(f"{key.metadata['Arn']} is disabled.")
```
bentsku, line 1355: question: wouldn't this be caught already by not setting `any_key_state_allowed`, and having `_get_kms_key`
raise the disabled exception? Or am I missing a case? Was there a specific reason to set `any_key_state_allowed` to `True` ?

localstack-core/localstack/services/kms/models.py
k-a-il: After some discussion we've decided not to replicate AWS behavior for `on_demand_rotation_start_date`, which means LocalStack will not return this field. Since we don't consider it essential to emulate and it's only used in the `get_key_rotation_status` function, please remove support for it from your code, along with all related tests.
To fix tests you can use `poll_condition` to wait until `OnDemandRotationStartDate` is no longer in the response, just like you did in the `test_rotate_key_on_demand_should_return_rotation_start_date_only_while_its_in_progress`.

localstack-core/localstack/services/kms/provider.py
agseco: Good catch... No reason at all it was a mistake, thanks for pointing it out. Fixed!

localstack-core/localstack/services/kms/models.py
agseco: Both issues fixed 🚀

bentsku (approved):
LGTM! Thanks for addressing the comments so quickly 🚀

k-a-il (approved):
Thank you for your contribution!

PR #12351, opened 2025-03-06
Implement IAM service specific credentials CRUD

dfangl:
## Motivation
IAM users can have service specific credentials assigned to them, they can be used for AWS CodeCommit (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_ssh-keys.html) and Amazon Keyspaces (for Apache Cassandra) (https://docs.aws.amazon.com/keyspaces/latest/devguide/programmatic.credentials.ssc.html).

For now, we need to have an implementation of the CRUD part of this functionality.

This initally depended on localstack/localstack-snapshot#10, but the last commit removes that dependency to avoid this PR being blocked by it.

## Changes
* LocalStack now supports service specific credential create, list, update, reset and delete operations.

### DIFF
--- a/localstack-core/localstack/services/iam/iam_patches.py
+++ b/localstack-core/localstack/services/iam/iam_patches.py
@@ -7,6 +7,7 @@ from moto.iam.models import (
     IAMBackend,
     InlinePolicy,
     Policy,
+    User,
 )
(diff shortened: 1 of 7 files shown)


pinzon (approved):
LGTM :+1:

PR #12352, opened 2025-03-07
fix redshift tests if not in docker

alexrashed:
## Motivation
Redshift in LocalStack Pro now fails when it cannot install a database (fail fast).
This breaks two RedShift CloudFormation tests, when not executed against LocalStack in the Docker container.
This PR adds markers to avoid the execution of these tests if executed outside of Docker.

## Changes
- Add `only_in_docker` marker to:
  - `test_redshift_cluster (tests.aws.services.cloudformation.resources.test_redshift)`
  - `test_create_clusters (tests.aws.services.redshift.test_redshift.TestRedshift)`

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_redshift.py
+++ b/tests/aws/services/cloudformation/resources/test_redshift.py
@@ -3,6 +3,8 @@ import os
 from localstack.testing.pytest import markers
 
 
+# only runs in Docker when run against Pro (since it needs postgres on the system)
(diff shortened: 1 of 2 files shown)


bentsku (approved):
LGTM! Thanks for jumping on this 🦸

alexrashed:
FYI: Ignoring an unrelated issue with a firehose tset in the pipeline.

PR #12355, opened 2025-03-07
Cloud Formation: [POC] Modeling Stack Operations and Evaluating ChangeSet Descriptions

MEPalma:
## Motivation
In an effort to address the current limitations in the handling of stack updates and the ChangeSet workflow (describe, update, etc.), as well as other workflows such as stack creation, deletion, and rollback, this pull request introduces a new design approach to representing operations on stack templates. This design enables:
- Implementation of updates to stack templates, and creation/deletion operations.
- Logic that computes update graphs for different template versions.
- A visitor pattern for extension logic to operate on these update graphs.
- The ability to derive detailed ChangeSet descriptions via the new design model.

## Changes
- ChangeSetModel: A POC class that, given one or two CloudFormation templates, produces an update graph. This graph can represent both template versions, and it can be acted upon via a visitor mechanism.
- ChangeSetModelVisitor: A POC abstract class to navigate update graphs through the visitor pattern, enabling extensions and custom operations on the graph.
- ChangeSetModelDescriber: A POC class extending ChangeSetModelVisitor to traverse update graphs and compute the outputs necessary for the ChangeSetDescribe operations.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -305,7 +305,13 @@ class ChangeSetModel:
     def _visit_resource(
         self, resource_name: str, before_resource: Maybe[dict], after_resource: Maybe[dict]
     ) -> NodeResource:
-        # TODO: fix add/delete/unchanged logic, needs minor rework of node types being update informants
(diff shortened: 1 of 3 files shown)


simonrw (approved):
Thanks @MEPalma for introducing this structure, this will make operating on CFn templates much easier in the future!

PR #12357, opened 2025-03-07
CFn: provider cleanup

simonrw:
## Motivation

The provider needs cleaning up before we start integrating the new update process.

## Changes

* Remove `temp_stack` concept from `create_change_set`
* Restructure the `create_change_set` method to more closely follow the approach that AWS does


simonrw:
Closing this in favour of the rewrites in the new engine

PR #12362, opened 2025-03-11
Use per-file-target-version to define Python 3.9 for CLI code

silv-io:
## Motivation

As in dependent repositories, we have an issue with the ASF updates since the latest `ruff` update. The definition of the Python 3.9 target version in linting/formatting leads to syntax errors when enabling preview features like we do in the ASF update action.

To fix that, we need to define our versions correctly depending on if the code runs in the CLI or in the runtime (container)

## Changes

- Set target version to `py311` matching our `.python-version`
- Set `per-file-target-version` to `py39` for all code accessed in the CLI matching our minimum required version
- Apply changes that are made necessary by upgrading the target version to `py311`
- Make code that can be reached by the CLI `py39` compatible (remove match statement from `usage.py`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "localstack-core"
 authors = [
-        { name = "LocalStack Contributors", email = "info@localstack.cloud" }
+    { name = "LocalStack Contributors", email = "info@localstack.cloud" }
 ]
 description = "The core library and runtime of LocalStack"
 requires-python = ">=3.9"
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -190,7 +190,18 @@ exclude = [
...
+
+[tool.ruff.per-file-target-version]
+# Only allow minimum version for code used in the CLI
+"localstack-core/localstack/cli/**" = "py39"
+"localstack-core/localstack/packages/**" = "py39"
+"localstack-core/localstack/config.py" = "py39"
+"localstack-core/localstack/constants.py" = "py39"
+"localstack-core/localstack/utils/analytics/**" = "py39"
+"localstack-core/localstack/utils/bootstrap.py" = "py39"
+"localstack-core/localstack/utils/json.py" = "py39"
```
silv-io, line 203: I'm unsure whether this set of files is correct or maybe even too much (e.g., in the case of the analytics code where I even had to remove a `match` statement)

PR #12367, opened 2025-03-11
chore: improve test snapshot

giograno:
## Motivation
When snapshotting an exception, the best practice is to snapshot the actual response of the exception, instead of the `ExceptionInfo` object. In this way we can capture the proper code, message, and status code.
In addition, the test becomes more stable (e.g., [here](https://github.com/localstack/localstack/actions/runs/13791616449/job/38572719566?pr=12365) the test fails for a different traceback level).

## Changes
- Improve the snapshot as described.


silv-io (approved):
LGTM! thank you for the quick turnaround 🚀

PR #12368, opened 2025-03-11
Fix Invalid noqa Usage in Step Functions TestState Preprocessor

MEPalma:
## Motivation
We are getting some linter warning in the TestState preprocessor for Step Functions due to the misuse of `# noqa`.

## Changes
The changes move the noqa motivations to their separate comments.


silv-io (approved):
Thanks for fixing these! LGTM

PR #12369, opened 2025-03-11
Step Functions: Lazy Initialization of JVM for JSONata Evaluation 

MEPalma:
## Motivation
The current implementation starts the JVM during module import, which interferes with unrelated processes like generating entrypoints—especially on Windows machines, where it may break the entrypoint generation process entirely. This PR defers the startup of the JVM until the JSONata evaluation feature is actually needed first, ensuring that other processes remain unaffected by premature JVM initialization.

## Changes
- Delay JVM instantiation until the first JSONata evaluation call, then cache the function reference to ensure fast, persistent access throughout the application’s runtime.

\* Known Limitation: The first invocation in a state with tight timeouts might trigger a timeout error in the evaluation of the state machine due to the JVM installation overhead. This edge case is rare, does not result in abrupt terminations (the state machine will terminate gracefully), and does not affect future executions. Future refinements to the installer behavior may address this further.

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
@@ -86,11 +86,16 @@ class _JSONataJVMBridge:
             raise JSONataException("UNKNOWN", str(ex))
 
 
-# Final reference to the java evaluation function.
-_eval_jsonata: Final[Callable[[JSONataExpression], Any]] = _JSONataJVMBridge.get().eval_jsonata
+# Lazy initialization of the `eval_jsonata` function pointer.
+# This ensures the JVM is only started when JSONata functionality is needed.
+_eval_jsonata: Optional[Callable[[JSONataExpression], Any]] = None
 
 
 def eval_jsonata_expression(jsonata_expression: JSONataExpression) -> Any:
+    global _eval_jsonata
+    if _eval_jsonata is None:
+        # Initialize _eval_jsonata only when invoked for the first time using the Singleton pattern.
+        _eval_jsonata = _JSONataJVMBridge.get().eval_jsonata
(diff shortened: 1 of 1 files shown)


alexrashed:
I'll let @joe4dev and @gregfurman do the actual review, but I can verify that this solves the issue with the entrypoint generation.
`make entrypoints | grep JVM` fails on this PR (as no log with "JVM" is written) while there are lots of errors logged on the current `master`.

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
```diff
@@ -86,11 +86,16 @@ def eval_jsonata(self, jsonata_expression: JSONataExpression) -> Any:
...
+# Lazy initialization of the `eval_jsonata` function pointer.
+# This ensures the JVM is only started when JSONata functionality is needed.
+_eval_jsonata: Optional[Callable[[JSONataExpression], Any]] = None
 
 
 def eval_jsonata_expression(jsonata_expression: JSONataExpression) -> Any:
+    global _eval_jsonata
+    if _eval_jsonata is None:
+        # Initialize _eval_jsonata only when invoked for the first time using the Singleton pattern.
+        _eval_jsonata = _JSONataJVMBridge.get().eval_jsonata
```
gregfurman, line 98: Should we not be locking before assigning this global variable? I'm concerned about a race condition.

gregfurman (approved):
LGTM! Just one small concern about the global variable assign being racey.

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
MEPalma: The singleton class pattern and the installer are thread-safe, and the installation can only occur once, so we should be okay to keep this as it

Issue #12115, opened 2025-01-08
bug: Binary media support in API Gateway broken [lambda proxy integration] in 4.0.0+ (new native apigw provider) 


petr-pokorny-1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

As documented [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html), apigateway should return png image as binary stream when aws lambda passes base64 encoded image through proxy integration.

### Expected Behavior

4.0.0+ versions of localstack returns base64 string passed by lambda without converting it to binary stream. Older versions were work fine.

### How are you starting LocalStack?

With a docker-compose file

#### docker-compose

```yml
services:
  localstack:
    container_name: "${LOCALSTACK_CONTAINER_NAME:-localstack-issues}"
    image: "localstack/localstack-pro:${LOCALSTACK_VERSION:-latest}"
    ports:
      - "127.0.0.1:4566:4566"
      - "127.0.0.1:4510-4559:4510-4559"
      - "127.0.0.1:53:53"
      - "127.0.0.1:443:443"
    environment:
      - DOCKER_HOST=unix:///var/run/docker.sock
      - DEBUG=${DEBUG:-1}
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN-}
    volumes:
      - "./.localstack:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

#### lambda

```js
exports.handler = async (event) => {
    // Base64-encoded PNG image (example: 1x1 pixel transparent PNG)
    const imageBase64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8/wIAAgkBAyMAlYwAAAAASUVORK5CYII=";

    return {
        statusCode: 200,
        headers: {
            "Content-Type": "image/png",
            "Cache-Control": "no-cache"
        },
        body: imageBase64,
        isBase64Encoded: true
    };
};
```

#### deploy script

```bash
#!/bin/bash

set -e

# Set AWS CLI LocalStack endpoint
export AWS_ACCESS_KEY_ID=test
export AWS_SECRET_ACCESS_KEY=test
export AWS_DEFAULT_REGION=us-east-1
LOCALSTACK_URL=http://localhost:4566

# create lambda zip
zip function.zip lambda.js

[...]

PR #12371, opened 2025-03-11
fix APIGW binary media types

bentsku:
## Motivation
As reported by https://github.com/localstack/localstack/issues/12115#issuecomment-2715763423, we had an issue with binary media types and CDK, which led to finding out Moto was not considering `binaryMediaTypes` in the `CreateRestAPI` operation

## Changes
- add `binaryMediaTypes` to the REST API model


cloutierMat (approved):
Awesome! thanks for fixing, Surprising we didn't already have that implemented! :rofl:

PR #12373, opened 2025-03-12
chore: improve remaining tests snapshots

giograno:
## Motivation
As we fixed a snapshot leading to a test failure in https://github.com/localstack/localstack/pull/12367, I thought it was worth briefly going all over the repository and doing the same.

## Changes
- Improve snapshots by matching the response instead of the exception info object.


silv-io (approved):
Awesome! thanks for handling these

PR #12375, opened 2025-03-12
CFn updates: create new v2 provider for new engine

simonrw:
## Motivation

In order to make improvements to the CFn engine, we have decided to add a new provider to update the functionality. It is useful to be able to update the provider methods for specific operations (typically around change sets) to change the template preprocessing etc.

## Changes

Add a new provider under the name `engine-v2`. This can be enabled with `PROVIDER_OVERRIDE_CLOUDFORMATION=engine-v2`. There is no implementation for now.


MEPalma (approved):
Awesome, thanks!

PR #12382, opened 2025-03-13
CFn: switch to new usage counters

simonrw:
## Motivation

We are standardising on our metrics counters. This PR updates CloudFormation to use the new API rather than our own (non-standardised) `UsageSetCounter` API.

## Changes

- Use the new counters
## Testing

1. Start LocalStack with `DISABLE_EVENTS=0`
2. Deploy the following template with `awslocal cloudformation deploy --template-file <filename> --stack-name mystack`:

```yaml
Resources:
  MyParameter:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: my-parameter

  NonSupportedResource:
    Type: AWS::Logs::ResourcePolicy
    Properties:
      PolicyName: "MyResourcePolicy"
      PolicyDocument: "{ \"Version\": \"2012-10-17\", \"Statement\": [ { \"Sid\": \"Route53LogsToCloudWatchLogs\", \"Effect\": \"Allow\", \"Principal\": { \"Service\": [ \"route53.amazonaws.com\" ] }, \"Action\":\"logs:PutLogEvents\", \"Resource\": \"logArn\" } ] }"
```

3. put a breakpoint in `localstack.utils.analytics.metrics.publish_metrics`) and evaluate the `collected_metrics` variable. It should contain:

```json
{
  "metrics": [
    {
      "label_1": "latency",
      "label_1_value": 0,
      "name": "network_effect",
      "namespace": "chaos",
      "type": "counter",
      "value": 8
    },
    {
      "label_1": "resource_type",
      "label_1_value": "AWS::SSM::Parameter",
      "name": "resourcetype",
      "namespace": "cloudformation",
      "type": "counter",
      "value": 1
    },
    {
      "label_1": "missing_resource_type",
      "label_1_value": "AWS::Logs::ResourcePolicy",
      "name": "missingresourcetypes",
      "namespace": "cloudformation",
      "type": "counter",
      "value": 1
    }
  ]
}
```
(the first event is unrelated)

Or... evaluate the analytics events once they have reached the platform.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/usage.py
+++ b/localstack-core/localstack/services/cloudformation/usage.py
@@ -2,10 +2,6 @@ from localstack.utils.analytics.metrics import Counter
 
 COUNTER_NAMESPACE = "cloudformation"
 
-# XXX the name is resourcetype to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
-resource_types = Counter(namespace=COUNTER_NAMESPACE, name="resourcetype", labels=["resource_type"])
-
-# XXX the name is missingresourcetypes to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
-missing_resource_types = Counter(
-    namespace=COUNTER_NAMESPACE, name="missingresourcetypes", labels=["missing_resource_type"]
+resources = Counter(
+    namespace=COUNTER_NAMESPACE, name="resources", labels=["resource_type", "missing"]
 )
--- a/localstack-core/localstack/services/cloudformation/usage.py
+++ b/localstack-core/localstack/services/cloudformation/usage.py
@@ -1,4 +1,11 @@
-from localstack.utils.analytics.usage import UsageSetCounter
+from localstack.utils.analytics.metrics import Counter
 
-resource_type = UsageSetCounter("cloudformation:resourcetype")
-missing_resource_types = UsageSetCounter("cloudformation:missingresourcetypes")
+COUNTER_NAMESPACE = "cloudformation"
+
+# XXX the name is resourcetype to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
+resource_types = Counter(namespace=COUNTER_NAMESPACE, name="resourcetype", labels=["resource_type"])
+
+# XXX the name is missingresourcetypes to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
+missing_resource_types = Counter(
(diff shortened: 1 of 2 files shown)


pinzon (approved):
LGTM :+1:

localstack-core/localstack/services/cloudformation/usage.py
```diff
@@ -1,4 +1,11 @@
...
-resource_type = UsageSetCounter("cloudformation:resourcetype")
-missing_resource_types = UsageSetCounter("cloudformation:missingresourcetypes")
+COUNTER_NAMESPACE = "cloudformation"
+
+# XXX the name is resourcetype to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
+resource_types = Counter(namespace=COUNTER_NAMESPACE, name="resourcetype", labels=["resource_type"])
+
+# XXX the name is missingresourcetypes to maintain backwards compatibility with the previous counters (when using UsageSetCounter)
+missing_resource_types = Counter(
+    namespace=COUNTER_NAMESPACE, name="missingresourcetypes", labels=["missing_resource_type"]
```
vittoriopolverino, line 10: It might be a bit redundant to mention "missing" both in the metric name and the label. Since the metric is already called `missingresourcetypes`, using `resource_type` as the label (like in the resource_types counter) should work well and would also help maintain consistency within the same namespace! WDYT?

vittoriopolverino (commented):
**question:** would it make sense for you to unify the two counters into one, where we choose a metric name that clearly describes the CloudFormation behaviour/operation being measured? This way, we could use two labels: one to specify the resource type and another to indicate whether the resource is missing or not.

simonrw:
> question: would it make sense for you to unify the two counters into one, where we choose a metric name that clearly describes the CloudFormation behaviour/operation being measured? This way, we could use two labels: one to specify the resource type and another to indicate whether the resource is missing or not.

Thanks for the suggestion. I'm hoping to try to keep consistency with the metrics we already have (I don't know if this new implementation _will_ be consistent). I don't know if this is going to cause problems merging the two schemas later on. 

I can see the value in recording either supported or not, as currently we record all resources attempted, of which a subset are not supported. This way we capture each resource type once. Thanks for the suggestion!

vittoriopolverino:
> I don't know if this is going to cause problems merging the two schemas later on.

Great point. That’s exactly the main concern. Do you think the resource and missing resource schemas might diverge over time because they don’t share common characteristics? If so, I’d keep the metrics separate. Otherwise, I’d merge them.

PR #12383, opened 2025-03-13
APIGW: migrate to new Counter type for REST API analytics

bentsku:
## Motivation
This PR replaces the analytics Counter used for tracking usage of `execute-api` routes for API Gateway REST APIs.

## Changes
- replace the previous UsageSetCounter with the new Counter with labels

## Testing

To test the changes:

- run a LocalStack in host mode with this branch checked out
- run the `tests.aws.services.apigateway.test_apigateway_http.test_http_integration_method` integration test with `TEST_SKIP_LOCALSTACK_START=1` set, to avoid trying to start LocalStack with the test
shutdown LocalStack
- I've tried without sending the events but only printing the collected metrics and this is what I got:

```json
{
  "metrics": [
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 3,
      "type": "counter",
      "label_1_value": "HTTP",
      "label_1": "invocation_type"
    },
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 3,
      "type": "counter",
      "label_1_value": "HTTP_PROXY",
      "label_1": "invocation_type"
    }
  ]
}
```

Now, with more tests:
- `apigateway.test_apigateway_eventbridge.test_apigateway_to_eventbridge`
- `apigateway.test_apigateway_lambda.test_lambda_aws_proxy_integration`
- `apigateway.test_apigateway_sqs.test_sqs_aws_integration`

```json
{
  "metrics": [
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 11,
      "type": "counter",
      "label_1_value": "AWS_PROXY",
      "label_1": "invocation_type"
    },
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 1,
      "type": "counter",
      "label_1_value": "AWS:sqs",
      "label_1": "invocation_type"
    },
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 1,
      "type": "counter",
      "label_1_value": "AWS:events",
      "label_1": "invocation_type"
    }
  ]
}
```

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
@@ -1,7 +1,7 @@
 import logging
 
 from localstack.http import Response
-from localstack.utils.analytics.metrics import Counter, _LabeledCounterMetric
+from localstack.utils.analytics.metrics import Counter
 
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
 from ..context import RestApiInvocationContext
@@ -10,9 +10,7 @@ LOG = logging.getLogger(__name__)
 
 
 class IntegrationUsageCounter(RestApiGatewayHandler):
-    counter: _LabeledCounterMetric
-
-    def __init__(self, counter: _LabeledCounterMetric = None):
+    def __init__(self, counter: Counter = None):
         self.counter = counter or Counter(
             namespace="apigateway", name="rest_api_execute", labels=["invocation_type"]
         )
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
@@ -1,7 +1,7 @@
 import logging
 
 from localstack.http import Response
-from localstack.utils.analytics.usage import UsageSetCounter
+from localstack.utils.analytics.metrics import Counter, _LabeledCounterMetric
 
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
 from ..context import RestApiInvocationContext
@@ -10,10 +10,12 @@ LOG = logging.getLogger(__name__)
 
 
 class IntegrationUsageCounter(RestApiGatewayHandler):
-    counter: UsageSetCounter
+    counter: _LabeledCounterMetric
 
-    def __init__(self, counter: UsageSetCounter = None):
-        self.counter = counter or UsageSetCounter(namespace="apigateway:invokedrest")
+    def __init__(self, counter: _LabeledCounterMetric = None):
+        self.counter = counter or Counter(
+            namespace="apigateway", name="rest_api_execute", labels=["invocation_type"]
+        )
 
     def __call__(
         self,
@@ -31,7 +33,7 @@ class IntegrationUsageCounter(RestApiGatewayHandler):
             # hence we should count it as a NOT_FOUND invocation
             invocation_type = "NOT_FOUND"
 
-        self.counter.record(invocation_type)
+        self.counter.labels(invocation_type=invocation_type).increment()
 
     @staticmethod
     def _get_aws_integration_service(integration_uri: str) -> str:
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
@@ -1,7 +1,7 @@
 import logging
 
 from localstack.http import Response
-from localstack.utils.analytics.metrics import Counter
+from localstack.utils.analytics.metrics import Counter, LabeledCounterMetric
 
 from ..api import RestApiGatewayHandler, RestApiGatewayHandlerChain
 from ..context import RestApiInvocationContext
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
```diff
@@ -10,10 +10,12 @@
...
 class IntegrationUsageCounter(RestApiGatewayHandler):
-    counter: UsageSetCounter
+    counter: LabeledCounterMetric
 
-    def __init__(self, counter: UsageSetCounter = None):
-        self.counter = counter or UsageSetCounter(namespace="apigateway:invokedrest")
+    def __init__(self, counter: LabeledCounterMetric = None):
+        self.counter = counter or Counter(
+            namespace="apigateway", name="rest_api_execute", labels=["invocation_type"]
+        )
```
vittoriopolverino, line 18: **not-blocking**: I see that invocation_type includes two pieces of information: the invocation type itself and the target service being invoked via API Gateway, correct?
If you find it useful, we could also define a counter like this:

```
Counter(
    namespace="apigateway",
    name="rest_api_execute",
    labels=["invocation_type", "target_service"]
)

```
This would structure the data as:
```
 "metrics": [
    {
      "namespace": "apigateway",
      "name": "rest_api_execute",
      "value": 3,
      "type": "counter",
      "label_1": "invocation_type",
      "label_1_value": "AWS_PROXY",
      "label_2": "target_service",
      "label_2_value": "stepfunctions",
    }
]
```

vittoriopolverino (approved):
LGTM 🥳  LGTM! I just added a non-blocking comment in case you’d like to consider separating execution type and target service. This would simplify filtering and aggregation of invocation types by service.

localstack-core/localstack/services/apigateway/next_gen/execute_api/handlers/analytics.py
bentsku: The issue I have is that the "target service" is only relevant when the `invocation_type` is of `AWS`. However, all others: `AWS_PROXY, `MOCK`, `HTTP` and `HTTP_PROXY` do not have a target service, and I think the label value cannot be empty, right?

I think it's worth keeping them all separated, but as you said the grouping by `AWS:` prefix would be trickier. But I don't think the aggregation by service would be an issue because it could only be prefixed by `AWS:` and not any kind of `invocation_type`.

vittoriopolverino: Evaluating both options, using a placeholder or applying logic to extract the value after the AWS: prefix, I’m leaning toward the first option. Otherwise, we’d be introducing custom service-specific parsing rules, which I’d like to avoid by leveraging labels instead 🤔  WDYT?

bentsku: I see, that's also an idea. But I'm not sure I see what it could bring us except the ability to group by the `AWS` invocation type? Is there any other benefit to it?

vittoriopolverino: I see that AWS_PROXY also has a target service
Screenshot 2025-03-18 at 12 14 00

bentsku: > ....or, another way could be splitting the counter into two, one for AWS and another for AWS_PROXY, MOCK, HTTP, and HTTP_PROXY.

I think this would then lead to custom logic to join the 2 counters to get the total invocations, I'm not sure it would be worth. 

> But again, if aggregating by service when the invocation is AWS isn’t necessary, then we can just leave everything as it is.

But aggregating by service could be done currently, right, just by selecting the `AWS:<service>` value? The only we cannot do is aggregating all services together under one `AWS:` prefix, which I don't think is necessary.

Also, this value is somewhat already calculated by itself mentally, it's the total of invocations minus the sum of others.

vittoriopolverino: > I think this would then lead to custom logic to join the 2 counters to get the total invocations, I'm not sure it would be worth.


in this case, you can simply use a filter on `namespace = 'apigateway'` (no joins) to get the total invocations.

bentsku: I think the main thing is that I'm seeing `AWS:sqs` to be of same value/level as `HTTP_PROXY`, and not to be a subtype of the `AWS` invocation type. I think seeing things that way, it might make more sense?

I'm not hard set on it, I'd also say this is not a very used invocation type either, so modifying the "majority" to include the `n/a` value when `AWS:<x>` might be minority feels a bit wrong. But I can also see the point, as we're encoding it in the "type". But I'd argue it is a type in itself? Not entirely sure really what is the best way 🤔

vittoriopolverino: > But aggregating by service could be done currently, right, just by selecting the AWS:<service> value? The only thing we cannot do is aggregating all services together under one AWS: prefix, which I don't think is necessary.

Yes, that’s correct, it’s not that it’s difficult to do, but having the value explicitly in a label allows us to directly use a dropdown filter in a dashboard. This way, you can easily select a specific target service or view all services for that metric.

On the other hand, if we have to parse a label value because the information isn’t already in a separate label, we’d need to create a custom parsing query specifically for API Gateway. Ideally, we’d like to avoid these custom logic workarounds and instead be explicit by using labels.

From a data usability, dashboarding, and schema generalization perspective, this approach makes sense for us but I understand that these are details that don’t directly impact you, so I don’t want to compromise the usability of the counter on the codebase side either 😃

vittoriopolverino: Conversation resolved after a quick huddle 💪🏼 🥳 🚀

PR #12386, opened 2025-03-13
make counter classes public

vittoriopolverino:
## Motivation
@bentsku noticed this behavior while working on this [PR](https://github.com/localstack/localstack/pull/12383), thanks for the heads-up! 🚀 
Before this change, the classes `_CounterMetric` and `_LabeledCounterMetric` were "protected", which made it inappropriate to use them directly for static typing. This caused type hinting issues, particularly in cases where a Counter instance was assigned to a class attribute: 
```python
class TestClass:
    counter: Counter

    def __init__(self, counter: Counter):
        self.counter = counter or Counter(name="test", labels=["l1"])

        self.counter.labels(l1="test_value").increment()
```
Screenshot 2025-03-13 at 15 38 49

## Changes
The `CounterMetric` and `LabeledCounterMetric` classes have been made public, allowing clients to use them directly for typing. The use of TypeVar and [Protocol](https://peps.python.org/pep-0544/) was also considered, but it may not be necessary to rely on them in this specific case.
```python
class TestClass:
    counter: LabeledCounterMetric

    def __init__(self, counter: LabeledCounterMetric):
        self.counter = counter or Counter(name="test", labels=["l1"])

        self.counter.labels(l1="test_value").increment()
```

### DIFF
--- a/localstack-core/localstack/utils/analytics/metrics.py
+++ b/localstack-core/localstack/utils/analytics/metrics.py
@@ -140,7 +140,7 @@ class BaseCounter:
 class CounterMetric(Metric, BaseCounter):
     """
     A thread-safe counter for tracking occurrences of an event without labels.
-    This class is intended for type hinting only and should not be instantiated directly; use the Counter class instead.
+    This class should not be instantiated directly, use the Counter class instead.
     """
 
     _namespace: Optional[str]
@@ -175,7 +175,7 @@ class CounterMetric(Metric, BaseCounter):
 class LabeledCounterMetric(Metric):
     """
     A labeled counter that tracks occurrences of an event across different label combinations.
-    This class is intended for type hinting only and should not be instantiated directly; use the Counter class instead.
+    This class should not be instantiated directly, use the Counter class instead.
     """
 
     _namespace: Optional[str]
--- a/localstack-core/localstack/utils/analytics/metrics.py
+++ b/localstack-core/localstack/utils/analytics/metrics.py
@@ -260,7 +260,7 @@ class Counter:
     A factory class for creating counter instances.
 
     This class provides a flexible way to create either a simple counter
-    (`_CounterMetric`) or a labeled counter (`_LabeledCounterMetric`) based on
+    (`CounterMetric`) or a labeled counter (`LabeledCounterMetric`) based on
     whether labels are provided.
     """
 
--- a/localstack-core/localstack/utils/analytics/metrics.py
+++ b/localstack-core/localstack/utils/analytics/metrics.py
@@ -99,16 +99,17 @@ class Metric(ABC):
         pass
 
 
-class _Counter:
+class BaseCounter:
     """
     A thread-safe counter for any kind of tracking.
+    This class should not be instantiated directly, use the Counter class instead.
     """
 
     _mutex: threading.Lock
     _count: int
 
     def __init__(self):
-        super(_Counter, self).__init__()
+        super(BaseCounter, self).__init__()
         self._mutex = threading.Lock()
         self._count = 0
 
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -260,26 +260,26 @@ class Counter:
     A factory class for creating counter instances.
 
     This class provides a flexible way to create either a simple counter
-    (`_SimpleCounter`) or a labeled counter (`_LabeledCounter`) based on
+    (`_CounterMetric`) or a labeled counter (`_LabeledCounterMetric`) based on
```
alexrashed, line 263:
```suggestion
    (`CounterMetric`) or a labeled counter (`LabeledCounterMetric`) based on
```

localstack-core/localstack/utils/analytics/metrics.py
```diff
@@ -136,7 +136,7 @@ def reset(self) -> None:
             self._count = 0
 
 
-class _CounterMetric(Metric, _Counter):
+class CounterMetric(Metric, _Counter):
```
alexrashed, line 139: **question:** ‏`_Counter` is still private as to the convention with the leading underscore. Would that mean that `increment` and `reset` are still considered private functions in the IDE integration / for type suggestions?

alexrashed, line 139: **question:** ‏Making them public could mean that they are used directly instead of via the `Counter` init, Is that something that is fine, or should this be mentioned explicitly in the docstring maybe?

alexrashed (approved):
LGTM! Type hints are an important point of devx! Thanks for addressing this! Since all current usages are via the `Counter` init, and the types are not yet used, this change should be safe 🚀

localstack-core/localstack/utils/analytics/metrics.py
vittoriopolverino: Actually, this is something I’d like to avoid (having the classes used directly instead of the Counter). Do you think specifying this in the docstring would be sufficient?

alexrashed: That depends, you can make it absolutely clear by creating abstract classes, where the actual implementations are private, but I don't think that would be necessary. You can add it to the docstring, and you can guide core engineers in the PRs when they migrate to the new metrics client. I think that should be enough for now. :)

vittoriopolverino: The IDE suggestions seem to work the same as before, showing both increment and reset as public methods. It does feel odd to see a public method suggested from a class that is, by convention, private.

Making `_Counter` public would require renaming it to avoid conflicts with the factory class, and it would also implicitly allow direct usage, something I’d like to avoid.

Would making it public while explicitly stating in the docstring that it should not be used directly be an acceptable solution? It feels like the factory class approach is introducing some ambiguity in multiple areas 🤔

alexrashed: Well, that's totally yours to judge. As mentioned in the initial PR on the new metrics framework, I don't really see the need for the `Counter` class which "redirects to the right class" in the first place, but it's fair if you want to keep it this way. Also, as mentioned in the other comment, you can very much guide the usage of this framework, and - together with comments in the docstring - can drive the usage. Once it's used in a specific way all over the place it will be used in this way moving forward for new implementations as well usually.

If you really want to keep it like that you can still make the interface explicit and abstract while having a private class with the actual implementation (as outlined in my other comment).
That would bloat up the amount of classes, but would allow you to have everything you want:
- Create public abstract base classes defining the interfaces for both the `CounterMetric`, the `Counter`, and the `LabeledCounterMetric`.
- Create private implementations of each of the abstract classes.
- Use the public abstract classes as type hints for the Counter "factory", but the private implementations when creating them.

vittoriopolverino: Ok, agreed. In a separate PR over the next few days, I’ll either remove the factory class or implement abstract classes to define the interfaces. In the meantime, I’m adding comments in the docstrings to avoid blocking the other incoming PRs. This upcoming change won’t impact how the data is sent.

Additionally, introducing that change should be fairly straightforward for me and won’t require involving other engineers, as they’re already handling the migration from the old counter to the new one.

Thanks again for the feedback! 💪🏼 🙏🏼

PR #12387, opened 2025-03-13
Match IAM service linked role naming with AWS

dfangl:
## Motivation
Currently, our service linked roles will be named `r-<short-id>`.
This does not match the AWS naming, as it is usually named `AWSServiceRoleFor<service>`.

Also, the service linked roles will not have proper policies attached, rendering them unusable with IAM enforcement.

This PR will increase parity by defining both predefined names for service linked roles (for services LS supports), and a heuristic to get the name for those services which do not.

Also, we will properly reject requests for service linked roles with a suffix for services which do not support it, and attach the correct policies to the service linked roles.

This supersedes #11731

## Changes
* Service linked roles are now named correctly, and have the correct policies attached
* Creating service linked roles with a suffix for services not supporting it will fail with the right error message

### DIFF
--- a/localstack-core/localstack/services/iam/provider.py
+++ b/localstack-core/localstack/services/iam/provider.py
@@ -4,6 +4,7 @@ import logging
 import random
 import re
 import string
+import uuid
 from datetime import datetime
 from typing import Any, Dict, List, TypeVar
 from urllib.parse import quote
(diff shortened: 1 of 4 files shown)


pinzon (approved):
Great. More Parity!! :partying_face:

PR #12389, opened 2025-03-13
Step Functions: Migrate Usage Metrics to New Counter Standards

MEPalma:
## Motivation
This PR migrates our step functions usage metrics to the new Counter standards. It increases transparency by accurately reflecting the query language usage within each state machine. Previously, JSONata was always counted as used—even when combined with JSONPath. Since interleaved usage of both modes is critical for testing the shared memory model and understanding user reliance on this mechanism, the updated metrics now provide a clearer picture.

## Changes
-  Migrated step functions usage metrics to the new Counter standards
- Consolidated tracking of language features into a single Counter per state machine on creation
- Updated query language detection logic to differentiate between exclusive and interleaved usage of JSONPath and JSONata

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/usage.py
+++ b/localstack-core/localstack/services/stepfunctions/usage.py
@@ -6,5 +6,7 @@ from localstack.utils.analytics.metrics import Counter
 
 # Initialize a counter to record the usage of language features for each state machine.
 language_features_counter = Counter(
-    namespace="stepfunctions", name="language_features_used", labels=["query_language", "variables"]
+    namespace="stepfunctions",
+    name="language_features_used",
+    labels=["query_language", "uses_variables"],
 )
--- a/localstack-core/localstack/services/stepfunctions/usage.py
+++ b/localstack-core/localstack/services/stepfunctions/usage.py
@@ -2,19 +2,9 @@
 Usage reporting for StepFunctions service
 """
 
-from localstack.utils.analytics.usage import UsageCounter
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/stepfunctions/usage.py
```diff
@@ -2,19 +2,9 @@
...
-variables_create_counter = UsageCounter("stepfunctions:variables:create")
-
-# Successful invocations (also including expected error cases in line with AWS behaviour)
-invocation_counter = UsageCounter("stepfunctions:invocation")
-
-# Unexpected errors that we do not account for
-error_counter = UsageCounter("stepfunctions:error")
+# Initialize a counter to record the usage of language features for each state machine.
+language_features_counter = Counter(
+    namespace="stepfunctions", name="language_features_used", labels=["query_language", "variables"]
```
vittoriopolverino, line 9: not-blocking: for the label variables, we might consider naming it are_variables_used or uses_variables, which makes the intent even clearer. What do you think?

vittoriopolverino:
Could you please provide some guidance on how to trigger these StepFunctions "functions" locally? This would allow me to test the changes end-to-end, send events to Tinybird, and give you a chance to run queries on the test data before merging the PR. That way, we’ll have time to adjust the counter and refine the labels if needed 💪🏼

vittoriopolverino (approved):
LGTM! 👍 I'm happy to see that you’ve condensed the three events into a single one, that was exactly the goal of the counter refactoring 🚀

Issue #12144, opened 2025-01-20
bug: Secrets Manager is not properly using --rotation-lambda-arn and --rotation-rules


gzoller:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

(There were similar previous issues--but they were all closed as fixed.)

I create a secret key in Secrets Manager, set up a lambda to trigger on secret rotation, including all the appropriate IAM, etc. I can manually invoke the lambda and it works fine. My start up script includes as its last command:

```
aws secretsmanager rotate-secret \
    --region $AWS_DEFAULT_REGION \
    --endpoint-url $AWS_ENDPOINT_URL \
    --secret-id MySecretKey \
    --rotation-lambda-arn $ROTATION_LAMBDA_ARN \
    --rotation-rules AutomaticallyAfterDays=30 
```
This completes successfully with a 200 in the logs and appropriate/expected output in the lambda's logs.

However... if, after this script is done, I immediately issue this command:
```
aws secretsmanager rotate-secret \
    --region $AWS_DEFAULT_REGION \
    --endpoint-url $AWS_ENDPOINT_URL \
    --secret-id MySecretKey
```
Then I see this in the logs:
```
2025-01-20T04:06:19.586  INFO --- [et.reactor-0] localstack.request.aws     : AWS secretsmanager.RotateSecret => 400 (ResourceNotFoundException)
```
And there's zero output in a well-logged lambda, indicating it was not called. The only difference is the last 2 params, and if I manually call that command again (the first one) it works.

If I do a aws secretsmanager describe-secret...  I get appropriate output:
```
{
    "ARN": "arn:aws:secretsmanager:us-east-1:000000000000:secret:MySecretKey-MVFUZI",
    "Name": "MySecretKey",
    "RotationEnabled": true,
    "RotationLambdaARN": "arn:aws:lambda:us-east-1:000000000000:function:RotateSecretFunction",
    "RotationRules": {
        "AutomaticallyAfterDays": 30
    },
    "LastRotatedDate": "2025-01-19T22:17:41-06:00",
    "LastChangedDate": "2025-01-19T22:17:41.089000-06:00",
    "LastAccessedDate": "2025-01-19T18:00:00-06:00",
    "VersionIdsToStages": {
        "03ee0074-71e7-4129-a4ae-a694d2055401": [
            "AWSPREVIOUS"
        ],
        "d5c6f636-aa06-4813-a849-dfdfb371cf6f": [
            "AWSPENDING",
[...]

PR #12391, opened 2025-03-16
Secret Manager: Solve the issue for rotate secret after sub-sequent r…

mabuaisha:
## Motivation

This PR is added in order to address the issue raised by our community [12144](https://github.com/localstack/localstack/issues/12144). The issue happens when making subsequent rotate-secret calls without specifying RotationLambdaARN and RotationRules cause ResourceNotFoundException 

## Changes

- Read from rotation configurations for a given secret from the secret store when they are not provided by user
- Reading from secret store (current rotation configuration) before rotation happened helped us to fetch these configuration in order to invoke the target lambda function 

## Testing

In order to test this you can run localstack locally and run the following command

# Create role for lambda function
```
awslocal iam create-role --role-name lambda-rotation-role --assume-role-policy-document '{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": { "Service": "lambda.amazonaws.com" },
      "Action": "sts:AssumeRole"
    }
  ]
}'
```

#### Attach policy to role
```
awslocal iam put-role-policy --role-name lambda-rotation-role --policy-name secret-rotation-policy --policy-document '{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "secretsmanager:GetSecretValue",
        "secretsmanager:PutSecretValue"
      ],
      "Resource": "*"
    }
  ]
}'
```

#### Create lambda function
```bash 
awslocal lambda create-function --function-name rotate-secret-function --runtime python3.9 --role arn:aws:iam::000000000000:role/lambda-rotation-role --handler rotate_function.lambda_handler --timeout 30 --zip-file fileb://rotate_function.zip
```

#### Create secret
```
bash
awslocal secretsmanager create-secret --name MySecretKey --secret-string "initialValue"
```

#### First rotation 
```bash
awslocal secretsmanager rotate-secret --secret-id MySecretKey --rotation-lambda-arn arn:aws:lambda:us-east-1:000000000000:function:rotate-secret-function --rotation-rules AutomaticallyAfterDays=1
```

#### Second rotation
```bash 
awslocal secretsmanager rotate-secret --secret-id MySecretKey
```

### DIFF
--- a/localstack-core/localstack/services/secretsmanager/provider.py
+++ b/localstack-core/localstack/services/secretsmanager/provider.py
@@ -729,11 +729,15 @@ def backend_rotate_secret(
     if not self._is_valid_identifier(secret_id):
         raise SecretNotFoundException()
 
-    if self.secrets[secret_id].is_deleted():
+    secret = self.secrets[secret_id]
+    if secret.is_deleted():
         raise InvalidRequestException(
             "An error occurred (InvalidRequestException) when calling the RotateSecret operation: You tried to \
             perform the operation on a secret that's currently marked deleted."
         )
+    # Resolve rotation_lambda_arn and fallback to previous value if its missing
+    # from the current request
+    rotation_lambda_arn = rotation_lambda_arn or secret.rotation_lambda_arn
 
     if rotation_lambda_arn:
         if len(rotation_lambda_arn) > 2048:
@@ -746,14 +750,16 @@ def backend_rotate_secret(
             if rotation_period < 1 or rotation_period > 1000:
                 msg = "RotationRules.AutomaticallyAfterDays must be within 1-1000."
                 raise InvalidParameterException(msg)
+    else:
+        # Resolve auto_rotate_after_days and fallback to previous value
+        # if its missing from the current request
+        rotation_period = secret.auto_rotate_after_days
 
     try:
         lm_client = connect_to(region_name=self.region_name).lambda_
         lm_client.get_function(FunctionName=rotation_lambda_arn)
     except Exception:
         raise ResourceNotFoundException("Lambda does not exist or could not be accessed")
-
-    secret = self.secrets[secret_id]
 
     # The rotation function must end with the versions of the secret in
     # one of two states:
@@ -782,7 +788,7 @@ def backend_rotate_secret(
         pass
 
     secret.rotation_lambda_arn = rotation_lambda_arn
-    secret.auto_rotate_after_days = rotation_rules.get(rotation_days, 0)
+    secret.auto_rotate_after_days = rotation_period
     if secret.auto_rotate_after_days > 0:
         wait_interval_s = int(rotation_period) * 86400
         secret.next_rotation_date = int(time.time()) + wait_interval_s
--- a/localstack-core/localstack/services/secretsmanager/provider.py
+++ b/localstack-core/localstack/services/secretsmanager/provider.py
@@ -738,22 +738,25 @@ def backend_rotate_secret(
     # Resolve rotation_lambda_arn and fallback to previous value if its missing
     # from the current request
     rotation_lambda_arn = rotation_lambda_arn or secret.rotation_lambda_arn
-    rotation_period = 0
+    if not rotation_lambda_arn:
+        raise InvalidRequestException(
+            "No Lambda rotation function ARN is associated with this secret."
+        )
+
     if rotation_lambda_arn:
         if len(rotation_lambda_arn) > 2048:
             msg = "RotationLambdaARN must <= 2048 characters long."
             raise InvalidParameterException(msg)
 
+    # In case rotation_period is not provided, resolve auto_rotate_after_days
+    # and fallback to previous value if its missing from the current request.
+    rotation_period = secret.auto_rotate_after_days or 0
     if rotation_rules:
         if rotation_days in rotation_rules:
             rotation_period = rotation_rules[rotation_days]
             if rotation_period < 1 or rotation_period > 1000:
                 msg = "RotationRules.AutomaticallyAfterDays must be within 1-1000."
                 raise InvalidParameterException(msg)
-    else:
-        # Resolve auto_rotate_after_days and fallback to previous value
-        # if its missing from the current request
-        rotation_period = secret.auto_rotate_after_days
 
     try:
         lm_client = connect_to(region_name=self.region_name).lambda_
@@ -788,7 +791,7 @@ def backend_rotate_secret(
         pass
 
     secret.rotation_lambda_arn = rotation_lambda_arn
-    secret.auto_rotate_after_days = rotation_period or 0
+    secret.auto_rotate_after_days = rotation_period
     if secret.auto_rotate_after_days > 0:
         wait_interval_s = int(rotation_period) * 86400
         secret.next_rotation_date = int(time.time()) + wait_interval_s
--- a/localstack-core/localstack/services/secretsmanager/provider.py
+++ b/localstack-core/localstack/services/secretsmanager/provider.py
@@ -738,7 +738,7 @@ def backend_rotate_secret(
     # Resolve rotation_lambda_arn and fallback to previous value if its missing
     # from the current request
     rotation_lambda_arn = rotation_lambda_arn or secret.rotation_lambda_arn
-
+    rotation_period = 0
     if rotation_lambda_arn:
         if len(rotation_lambda_arn) > 2048:
             msg = "RotationLambdaARN must <= 2048 characters long."
@@ -788,7 +788,7 @@ def backend_rotate_secret(
         pass
 
     secret.rotation_lambda_arn = rotation_lambda_arn
-    secret.auto_rotate_after_days = rotation_period
+    secret.auto_rotate_after_days = rotation_period or 0
     if secret.auto_rotate_after_days > 0:
         wait_interval_s = int(rotation_period) * 86400
         secret.next_rotation_date = int(time.time()) + wait_interval_s
--- a/tests/aws/services/secretsmanager/functions/lambda_rotate_secret.py
+++ b/tests/aws/services/secretsmanager/functions/lambda_rotate_secret.py
@@ -224,3 +224,14 @@ def finish_secret(service_client, arn, token):
         token,
         arn,
     )
+    if "AWSPENDING" in metadata["VersionIdsToStages"].get(token, []):
+        service_client.update_secret_version_stage(
+            SecretId=arn,
+            VersionStage="AWSPENDING",
+            RemoveFromVersionId=token,
+        )
+        logger.info(
+            "finishSecret: Successfully removed AWSPENDING stage from version %s for secret %s.",
+            token,
+            arn,
+        )
--- a/tests/aws/services/secretsmanager/test_secretsmanager.py
+++ b/tests/aws/services/secretsmanager/test_secretsmanager.py
@@ -5,7 +5,7 @@ import random
 import uuid
 from datetime import datetime
 from math import isclose
-from typing import Optional
+from typing import Any, Optional
 
 import pytest
 import requests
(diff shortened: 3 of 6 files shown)


mabuaisha:

tests/aws/services/secretsmanager/test_secretsmanager.snapshot.json
```diff
@@ -4586,5 +4582,339 @@
...
+            ]
+          }
+        ],
+        "ResponseMetadata": {
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 200
+        }
+      }
+    }
+  },
```
simonrw, line 4641: nit: I don't think this is a valid test snapshot. I can't see where this function was run with the parameter `False` in the code

tests/aws/services/secretsmanager/test_secretsmanager.validation.json
```diff
@@ -101,14 +101,23 @@
...
+  "tests/aws/services/secretsmanager/test_secretsmanager.py::TestSecretsManager::test_rotate_secret_multiple_times_with_lambda_success[False]": {
+    "last_validated_date": "2025-03-16T07:42:50+00:00"
+  },
+  "tests/aws/services/secretsmanager/test_secretsmanager.py::TestSecretsManager::test_rotate_secret_multiple_times_with_lambda_success[True]": {
+    "last_validated_date": "2025-03-16T07:43:08+00:00"
+  },
   "tests/aws/services/secretsmanager/test_secretsmanager.py::TestSecretsManager::test_rotate_secret_with_lambda_success": {
     "last_validated_date": "2024-03-15T08:12:22+00:00"
   },
+  "tests/aws/services/secretsmanager/test_secretsmanager.py::TestSecretsManager::test_rotate_secret_with_lambda_success[False]": {
```
simonrw, line 113: As per my previous comment, can we remove this validation if it is not a currently tracked test combination?

tests/aws/services/secretsmanager/test_secretsmanager.snapshot.json
mabuaisha: you are right, I think I missed this while I was doing some refactoring and forgot to change it back again. It should not be False, it should be `None` indeed, I can update this

tests/aws/services/secretsmanager/test_secretsmanager.validation.json
mabuaisha: Same as I mentioned before, you are right. This should be `None` not False. I will update this

localstack-core/localstack/services/secretsmanager/provider.py
```diff
@@ -746,15 +750,17 @@ def backend_rotate_secret(
             if rotation_period < 1 or rotation_period > 1000:
                 msg = "RotationRules.AutomaticallyAfterDays must be within 1-1000."
                 raise InvalidParameterException(msg)
+    else:
+        # Resolve auto_rotate_after_days and fallback to previous value
+        # if its missing from the current request
+        rotation_period = secret.auto_rotate_after_days
```
macnev2013, line 756: this could cause 'rotation_period' variable to be referenced before assignment [here](https://github.com/mabuaisha/localstack/blob/secret-manager-rotation-secret-issue/localstack-core/localstack/services/secretsmanager/provider.py#L791)

localstack-core/localstack/services/secretsmanager/provider.py
```diff
@@ -729,11 +729,15 @@ def backend_rotate_secret(
...
-    if self.secrets[secret_id].is_deleted():
+    secret = self.secrets[secret_id]
+    if secret.is_deleted():
         raise InvalidRequestException(
             "An error occurred (InvalidRequestException) when calling the RotateSecret operation: You tried to \
             perform the operation on a secret that's currently marked deleted."
         )
+    # Resolve rotation_lambda_arn and fallback to previous value if its missing
+    # from the current request
+    rotation_lambda_arn = rotation_lambda_arn or secret.rotation_lambda_arn
```
macnev2013, line 740: What if the secret is being rotated first time and we haven't specified the `rotation_lambda_arn`?

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -103,6 +103,81 @@ def _is_secret_deleted():
...
+        )
+
+        sm_snapshot.match(
+            f"list_secret_versions_rotated_1_{snapshot_key_suffix}", list_secret_versions_1
+        )
+
+        # As a result of the Lambda invocations. current version should be
+        # pointed to `AWSCURRENT` & previous version to `AWSPREVIOUS`
+        assert response["VersionIdsToStages"][initial_secret_version] == ["AWSPREVIOUS"]
+        assert response["VersionIdsToStages"][rot_res["VersionId"]] == ["AWSCURRENT"]
```
macnev2013, line 179: suggestion: we usually don't use assertion in the helper methods.

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -533,58 +608,65 @@ def test_rotate_secret_with_lambda_success(
...
+    @markers.aws.validated
+    def test_rotate_secret_multiple_times_with_lambda_success(
+        self,
+        sm_snapshot,
+        secret_name,
+        create_secret,
+        create_lambda_function,
+        aws_client,
+        set_rotation_on_second_run,
+    ):
```
macnev2013, line 641: It would be worth testing - what happens if the `function_arn` is not provided for the first time rotation.

macnev2013 (commented):
Thanks for the PR @mabuaisha 🚀 

The PR looks good to me overall - I just have some minor concerns/suggestion.

tests/aws/services/secretsmanager/functions/lambda_rotate_secret.py
```diff
@@ -224,3 +224,14 @@ def finish_secret(service_client, arn, token):
         token,
         arn,
     )
+    if "AWSPENDING" in metadata["VersionIdsToStages"].get(token, []):
+        service_client.update_secret_version_stage(
+            SecretId=arn,
+            VersionStage="AWSPENDING",
+            RemoveFromVersionId=token,
+        )
+        logger.info(
```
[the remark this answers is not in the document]
mabuaisha: The reason why I choose the `info` is because in the same function for `AWSCURRENT` and others log levels  we are also using `info` so I thought this was added in purpose, so I stick to use info in this case

localstack-core/localstack/services/secretsmanager/provider.py
```diff
@@ -782,7 +788,7 @@ def backend_rotate_secret(
         pass
 
     secret.rotation_lambda_arn = rotation_lambda_arn
-    secret.auto_rotate_after_days = rotation_rules.get(rotation_days, 0)
+    secret.auto_rotate_after_days = rotation_period
```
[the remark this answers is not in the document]
mabuaisha: I added this in purpose without `or 0` because usually the default value for https://github.com/localstack/moto/blob/localstack/moto/secretsmanager/models.py#L118 based on this logic usually is 0 in case its not set. 

For extra check, we can still also fallback to 0
`secret.auto_rotate_after_days = rotation_period or 0 `

I will update this

localstack-core/localstack/services/secretsmanager/provider.py
mabuaisha: Yeah thats right, I will init `rotation_period = 0`

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -103,6 +103,81 @@ def _is_secret_deleted():
...
+    ) -> (str, str):
+        cre_res = create_secret(
+            Name=secret_name,
+            SecretString="my_secret",
+            Description="testing rotation of secrets",
+        )
+
+        sm_snapshot.add_transformer(
+            sm_snapshot.transform.key_value("RotationLambdaARN", "lambda-arn")
+        )
```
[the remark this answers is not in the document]
mabuaisha: Definitely, I can move it there to make it a generic transformer since `RotationLambdaARN ` could be user later on with multiple tests. define it there could be useful later on

tests/aws/services/secretsmanager/test_secretsmanager.py
mabuaisha: Is it a pattern you followed when you write tests ? or you are talking in general ?. The reason why I add this method is to avoid duplication for multiple tests and I  thought usually this could be re-used only within multiple tests for assertions. However, I'm open to change this if you think this could create confusion or violates any sort of patterns/convention you usually followed

localstack-core/localstack/services/secretsmanager/provider.py
mabuaisha: Actually, this was an existing behavior even before I introduce my fix. In case for the first time rotation we did not specify `rotation_lambda_arn` then it will raise error, check this code here 
```

    try:
        lm_client = connect_to(region_name=self.region_name).lambda_
        lm_client.get_function(FunctionName=rotation_lambda_arn)
    except Exception:
        raise ResourceNotFoundException("Lambda does not exist or could not be accessed")
```

This will raise error whenever we invoked it from the cli

`An error occurred (ResourceNotFoundException) when calling the RotateSecret operation: Lambda does not exist or could not be accessed`

mabuaisha:
@macnev2013  Thanks a lot for taking the time to review my PR. I replied to your comments and pushed changes as a result of your comments. Let me know if need to discuss this further

localstack-core/localstack/services/secretsmanager/provider.py
simonrw: Is it always `0` or can we move the `rotation_period = secret.auto_rotate_after_days` above the `if` block and have _that_ as the default, then remove the `else` branch?

localstack-core/localstack/services/secretsmanager/provider.py
simonrw: What error message does AWS return in this case? Does it match the error we raise in this circumstance? I suspect not.

tests/aws/services/secretsmanager/test_secretsmanager.py
simonrw: nit: we normally prefer two things:
1. assertions in the test body, not in helper methods
2. fixtures over helper methods

I personally find it clearer for the assertion failure to occur in the test body so it's clear what's being tested at first glance in the test code itself, rather than having to go diving into helper methods. It also makes the helper methods more flexible, as the assertions may not be consistent from test to test.

For the second item, we often create fixtures that look like:

```python
def my_fixture(aws_client, snapshot):  # other fixtures
    def inner(name: str, *args, **kwargs):  # variables taken from the test body
        # do something with both arguments and fixtures, e.g.:
        return aws_client.sts.get_caller_identity()
    
    return inner
```

then in the test we only need to include the fixture and use the "inner" function, which simplifies the usage of that fixture:

```python
def test_foo(my_fixture):
    identity = my_fixture("name")
```

See how we could call AWS methods without passing in the `aws_client`.

These are just preferences and I'm _not_ asking you to refactor the helper methods into fixtures, but perhaps you could consider moving the assertions to the test bodies?

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -533,85 +628,91 @@ def test_rotate_secret_with_lambda_success(
...
-            )
-        sm_snapshot.match("rotate_secret_invalid_arn_exc", e.value.response)
+        self._setup_invalid_rotation_secret(
+            invalid_arn, "rotate_secret_invalid_arn_exc", sm_snapshot, secret_name, aws_client
+        )
+        describe_secret = aws_client.secretsmanager.describe_secret(SecretId=secret_name)
+        sm_snapshot.match("describe_secret", describe_secret)
+        assert "RotationEnabled" not in describe_secret
+        assert "RotationRules" not in describe_secret
+        assert "RotationLambdaARN" not in describe_secret
```
simonrw, line 706: nit: these assertions are not required giving we are matching the snapshot above

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -533,85 +628,91 @@ def test_rotate_secret_with_lambda_success(
...
 
+    @markers.snapshot.skip_snapshot_verify(paths=["$..Error", "$..Message"])
+    @markers.aws.validated
+    def test_first_rotate_secret_with_missing_lambda_arn(
```
simonrw, line 710: ~This test is a start, however it does not execute the rotation itself, therefore testing the error message given by AWS if a lambda function arn is not specified. This will help clarify the error handling [here](https://github.com/localstack/localstack/pull/12391/files#r2017059049).~

Here is a demonstration of the confusion that extracting these helper methods causes. I (incorrectly) believed that you hadn't tried to rotate the lambda function, but when I looked in the snapshot I saw that the `_setup_invalid_rotation_secret` method tries to rotate a secret without specifying a function ARN.

I would find it much clearer if the test was more like:


* create secret (using our `create_secret` fixture)
* try to rotate secret, wrapped in `pytest.raises(...)`
* capturing a snapshot of the error message
* (optional) a describe secret call, however this is covered by other more specific tests

tests/aws/services/secretsmanager/test_secretsmanager.snapshot.json
```diff
@@ -3788,19 +3786,18 @@
         },
         "VersionIdsToStages": {
           "<version_uuid:1>": [
-            "AWSPREVIOUS"
+            "AWSCURRENT"
```
simonrw, line 3789: q: do you understand why this changed?

tests/aws/services/secretsmanager/test_secretsmanager.py
```diff
@@ -533,85 +628,91 @@ def test_rotate_secret_with_lambda_success(
...
+        self._setup_invalid_rotation_secret(
+            invalid_arn, "rotate_secret_invalid_arn_exc", sm_snapshot, secret_name, aws_client
+        )
+        describe_secret = aws_client.secretsmanager.describe_secret(SecretId=secret_name)
+        sm_snapshot.match("describe_secret", describe_secret)
+        assert "RotationEnabled" not in describe_secret
+        assert "RotationRules" not in describe_secret
+        assert "RotationLambdaARN" not in describe_secret
 
+    @markers.snapshot.skip_snapshot_verify(paths=["$..Error", "$..Message"])
```
simonrw, line 708: issue: you have skipped the key part of this test => the error message. Can you remove this verify and implement the correct behaviour?

tests/aws/services/secretsmanager/functions/lambda_rotate_secret.py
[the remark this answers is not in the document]
simonrw: This is only in the lambda function body though right @macnev2013?

simonrw (commented):
I hope have added some clarity and context to some of the points from @macnev2013 particularly around the use of helper methods and assertions.

I have an issue with `test_first_rotate_secret_with_missing_lambda_arn`: you are skipping verification of the two fields that _matter_ for this test. It is specifically around capturing the error that is presented, but you have skipped the wording.

I think this may cause confusion to our users who get an invalid response. I would like to see this implemented correctly.

tests/aws/services/secretsmanager/test_secretsmanager.py
mabuaisha: Sure thing, I will remove the check it. The only thing that need to be updated is to add a check that raise an error `InvalidRequestException` that match AWS one

tests/aws/services/secretsmanager/test_secretsmanager.snapshot.json
mabuaisha: My assumption for this. The list is not deterministic as I noticed that sometimes version1 could be the previous version and version 2 is the current version and vice versa. Thats the reason why sometimes version 1 could be previous and sometimes it could be current. The same thing for version 2 as well

localstack-core/localstack/services/secretsmanager/provider.py
mabuaisha: you are right. I will update the code because I need to add a check earlier to raise an error that match AWS one

mabuaisha: The error mainly look like this from AWS

```
        "Error": {
          "Code": "InvalidRequestException",
          "Message": "No Lambda rotation function ARN is associated with this secret."
        },
```

localstack-core/localstack/services/secretsmanager/provider.py
mabuaisha: I can remove the `else` branch and make it the default one. However, I will fallback to `0` like this

    rotation_period = secret.auto_rotate_after_days or 0

That way even if previous value for for `auto_rotate_after_days` that is part of secret is not integer, we are make sure always if the value is not set the default one is `0`

tests/aws/services/secretsmanager/test_secretsmanager.py
mabuaisha: Make sense, I think you raised a valid point here. I will do some refactoring to this

tests/aws/services/secretsmanager/test_secretsmanager.py
mabuaisha: Now I have been more lean toward refactor this since it created this confusion. Thanks for raising this

simonrw (approved):

Issue #12390, opened 2025-03-14
bug: EventBridge PutEvents call with a payload of 40KB crashes LocalStack


ffeytons:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When sending a payload of >= 40KB in the LocalStack EventBridge using the `PutEvents` API, the LocalStack Docker container increases its consumption of RAM by multiple GB and ends up being killed by the Docker Desktop engine.
Tried with OrbStack (alternative to Docker Desktop), similar issue: the container increases its consumption of RAM by multiple GB and OrbStack indicates in the end that the "python" process has been killed. At that moment, LocalStack outputs the following log: `LocalStack supervisor: localstack process (PID 13) returned with exit code -9`

Sending a much smaller payload (e.g. 2-3KB) works fine.

I did not test with in-between sizes for the payload, so it might be that the minimum to crash LocalStack is lower than 40KB.

To be noted:
* I did not have this behaviour with the previous major version of LocalStack (v3.x)
* I tried with `LOCALSTACK_PROVIDER_OVERRIDE_EVENTS=legacy` --> same result
* I tried with `LOCALSTACK_PROVIDER_OVERRIDE_EVENTS=v1` --> same result

### Expected Behavior

LocalStack EventBridge is hable to handle a payload of up to 256KB (the limit of AWS EventBridge)

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)
    // Multiple services written in Node.js and PHP, they all have the same issue

    // Node.js - AWS SDK
    await eventBridgeClient.send(new PutEventsCommand(event));
    
    // PHP - AWS SDK 
    $this->eventBridgeClient->putEvents([
            'Entries' => ...
        ]);

#### Steps to reproduce
1. Send an EventBridge message with the `PutEvents` API, and include in the `Detail` object of the `Entries` a payload >= 40KB
2. The API call to LocalStack times out after some moment (PHP client error: `Error executing "PutEvents" on "http://host.docker.internal:4566"; AWS HTTP error: cURL error 52: Empty reply from server (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://host.docker.internal:4566`)
3. The message is never processed by the LocalStack EventBridge
[...]

PR #12394, opened 2025-03-17
events: fix Event Ruler to not evaluate the full payload

bentsku:
## Motivation
As reported by #12390, when evaluate large payload for `PutEvents`, LocalStack would fail. I could reproduce the issue and saw LocalStack consumes as much as 20GB of memory while trying to evaluate the payload given by the user.
This payload was very complex, with lots of nested dictionaries and list. List are "exploding" the complexity, creating "or" conditions and new "branches" to evaluate. 

This is because we're using recursion to evaluate EventPattern against payload, and we would try to get the event payload in a flat shape to be able to be evaluated against the Pattern (due to how AWS treats the `.` notation). 

I've added some conditions checks to only evaluate the payload when relevant to the Event Pattern, and to not evaluate "branches" of the payload. This fixes the issue and the memory usage, as payloads now stay "simple" and without the bounds. 

I've also added some small unit tests because the doc strings were getting a bit too big to explain all the behavior. 

\cc @cloutierMat as we worked on this together for SNS

_fixes #12390_

See https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-pattern.html

Reading again, this should have been considered from the start with the following (my bad!):
> EventBridge ignores the fields in the event that aren't included in the event pattern.

## Changes
- add a check to not evaluate the payload "branch" if not relevant to the pattern

We will need to port this fix to the SNS Filter policy engine as well, I'll tackle that in a follow up.

### DIFF
--- a/localstack-core/localstack/services/events/event_rule_engine.py
+++ b/localstack-core/localstack/services/events/event_rule_engine.py
@@ -43,7 +43,7 @@ class EventRuleEngine:
 
         # TODO: maybe save/cache the flattened/expanded pattern?
         flat_pattern_conditions = self.flatten_pattern(event_pattern)
-        flat_payloads = self.flatten_payload(payload)
+        flat_payloads = self.flatten_payload(payload, flat_pattern_conditions)
 
         return any(
             all(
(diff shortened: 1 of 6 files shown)


maxhoheiser (approved):
LGTMT - nice and quick fix for the large payload problem ! 👍

PR #12395, opened 2025-03-17
SNS: fix Filter Policy engine to not evaluate full complex payload

bentsku:
## Motivation
Follow up from #12394. I'm porting the fix to SNS as well as it suffered from the same issue.

Complex payload would explode in complexity, and we were evaluating them in their entirety instead of only evaluate what is needed evaluate the policy itself. AWS is doing the same thing.

## Changes
- only evaluate the payload if relevant to the filter policy
- add AWS validated integration test for it and adapt unit tests

### DIFF
--- a/localstack-core/localstack/services/sns/filter.py
+++ b/localstack-core/localstack/services/sns/filter.py
@@ -58,7 +58,7 @@ class SubscriptionFilter:
 
         # TODO: maybe save/cache the flattened/expanded policy?
         flat_policy_conditions = self.flatten_policy(filter_policy)
-        flat_payloads = self.flatten_payload(payload)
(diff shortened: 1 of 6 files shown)


cloutierMat (approved):
Great solution! Given that the complexity of the filter is limited, this will greatly limit the possible paths followed when flattening the payload.

Question: have you tried to test with a highly complex filter and highly matching payload to see the worst case scenario? I assume it will still be manageable given the 150 max complexity, it won't be as crazy as before though! :eyes:

cloutierMat:
Indeed, I don't think it is worth investigating more for now! Pure curiosity! :wink:

Issue #12330, opened 2025-03-03
bug: S3 crc64 checksum mismatches for multi-part uploads


rumblefrog:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

```
upload failed: ./40mb.json to s3://repro-bucket/40mb.json An error occurred (BadDigest) when calling the CompleteMultipartUpload operation: The crc64nvme you specified did not match the calculated checksum.
```

### Expected Behavior

It uploads the file without any errors. I also don't believe it should be using crc64 for multi-part as it's not a composite checksum.

### How are you starting LocalStack?

With a `docker run` command

#### Minimal Reproduce

```
aws --version
aws-cli/2.24.15 Python/3.12.9 Darwin/23.5.0 source/arm64
```

```bash
#!/bin/bash

# Grab 40 MB file here: https://examplefile.com/code/json/40-mb-json or any files that's large enough to warrant multi-part.


# Start localstack docker container if it does not exist
docker rm -f localstack || true
docker run -d -p 4566:4566 --name localstack localstack/localstack:4.2.0

# Create s3 bucket
AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar  aws --endpoint-url=http://localhost:4566 --region=us-east-1 s3 mb s3://repro-bucket

AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=bar  aws --endpoint-url=http://localhost:4566 s3 cp ./40mb.json s3://repro-bucket/40mb.json
```

```
❯ ./repro.sh
localstack
40f2d38222ac9fbebc7450788a38a8544f280b9566db9ae89f5743cd9a2d8254
make_bucket: repro-bucket
upload failed: ./40mb.json to s3://repro-bucket/40mb.json An error occurred (BadDigest) when calling the CompleteMultipartUpload operation: The crc64nvme you specified did not match the calculated checksum.
```

### Environment

```markdown
- OS: MacOS Sonoma
- LocalStack:
  LocalStack version: 4.2.0
  LocalStack Docker image sha: 8a2777169ecd245da1e05383275489d0d26f89f4d120d20504e5cdf6d8e0a88b
  LocalStack build date: 2025-02-27
  LocalStack build git hash: 2144203e0
```

### Anything else?

Related https://github.com/localstack/localstack/issues/12187

bentsku:
Hello @rumblefrog and thanks for the report!

[...]

PR #12401, opened 2025-03-17
S3: fix checksum behavior and add new AWS behavior

bentsku:
## Motivation
TBD: the PR is going to fail because I need to update some snapshots and I'd rather wait for CI to know which one 😄

This PR fixes the behavior reported by #12330, which was due to a combination of parameters that was previously untested (default behavior of the AWS CLI v2, different from v1). 

It also updates the snapshots because AWS has finally decided to actually follow their documentation and truly set `CRC64NVME` to be the default checksum if none is specified by the client. I had an explicit test for that with a raw HTTP client that still returned no default checksum which surprised me at the time.

It seems it finally got deployed, so this new behavior is now added. 

I've also added a big amount of tests because AWS seems to be.. somewhat weird about `CopyObject` and returning default `CRC64NVME` checksums, so I've added tests to verify behavior across all checksum algorithms. 

Note: it seems that when SSE-C is used with Multipart uploads, it disables the automatic checksum calculation, but not for PutObject.

I've also sneak a tiny parity fix for `HeadObject` with range requests, it now returns the range. 

Upstream run: ✅  #\13919255105

_fixes #12330_

## Changes
- update the validation behavior of S3 multipart in case no checksum are specified at creation (you don't have to specify a checksum value when completing, this also might be new?)
- separate the default underlying object checksum algorithm from the multipart checksum algorithm, as S3 now properly differentiate between the 2
- return the `ChecksumType` with `HeadObject` (new, saw with the ASF update and the snapshots)
- start calculating default checksum value every time if it isn't specified for any new objects, like AWS
- return the `ContentRange` for `HeadObject` like `GetObject` (recently added by Botocore)


k-a-il (approved):
Nice! Great test for checksums after the object is copied 😄 🚀

PR #12405, opened 2025-03-18
Cloud Formation: [POC] Scoping Mechanism, Base Support for Parameters, Dynamic Parameters, Conditions, Intrinsic Functions, and Type Divergence

MEPalma:
## Motivation
Currently, the POC Update Graph approach to modeling CloudFormation templates does not support CloudFormation Parameters or the Ref intrinsic function. Additionally, due to the lack of built-in positional identifiers, the ChangeSetEntity abstraction does not support workflows that need to track the asynchronous computation of template entities, nor does it enable memoization strategies.
These changes introduce a scoping mechanism that assigns each ChangeSetEntity a unique identifier representing its logical position (or scope) within the template. They also add support for modeling changes in Parameters, Dynamic Parameters, and Ref operations within change sets. Furthermore, unit tests have been added to cover basic scenarios in this context.

The POC Update Graph approach to modeling CloudFormation templates does not support CloudFormation Conditions. This gap prevents condition-driven workflows from being accurately tracked or updated within the existing logic. In addition, there is no support for certain intrinsic functions (such as Fn::If, Fn::Not, and Fn::Equals). Furthermore, the modeling layer does not currently handle scenarios where changes in a template result in a new subtype or intrinsic function (for example, switching from one intrinsic function to a different one). This limitation meant that strict type changes to certain template bindings was not supported. Finally, caching mechanisms within the describer visitor are needed. A number of minor issues with this pipeline were also addressed. Unit tests have been added to validate these scenarios.

## Changes
- Added initial scoping logic to the Update Graph building logic
- Added base support for Parameter and Dynamic Parameters processing in the Update Graph building logic
- Added base support for Ref intrinsic functions in the Update Graph building logic
- Added base support for change set describes involving Parameters, Dynamic Parameters, and Ref
- Added base support for modeling CloudFormation Conditions in the Update Graph building logic
- Added modeling support for new intrinsic functions: Fn::If, Fn::Not, and Fn::Equals
- Enabled propagation of condition changes throughout the update graph, including condition resource statements and references in intrinsic functions
- Introduced the NodeDivergence concept, which models changes in subtype values (e.g., switching from one intrinsic function to another)
- Implemented caching in the describer visitor to improve performance and clarity in describing template changes
- Added support for describing change set changes involving conditions and the newly added intrinsic functions
- Created unit tests covering conditions, new intrinsic functions, and the NodeDivergence mechanism
- Several other minors


simonrw (approved):
Thanks for implementing this, I'm excited to get these changes working!

PR #12409, opened 2025-03-19
Make Lambda reusable for CloudFront Lambda@Edge

joe4dev:
## Motivation

CloudFront Lambda@Edge requires some changes in Lambda to be able to re-use the existing Lambda infrastructure.

## Changes

* Make the Lambda `Function` object optional in `counting_service::get_invocation_lease` such that we can re-use the counting service for Lambda@Edge, and skip reserved and provisioned concurrency
* Track version state in the version manager (i.e., fix unused self.state field)
* Refactor version manager such that the optional `function` is only used for the counting service


dfangl (approved):
LGTM! As discussed a bit hacky, but the best we can do for now!

PR #12410, opened 2025-03-19
transcribe: add speaker diarization test

sannya-singal:


tests/aws/services/transcribe/test_transcribe.py
```diff
@@ -403,3 +405,39 @@ def test_transcribe_start_job_same_name(
...
+        self, transcribe_create_job, media_file, max_speakers, aws_client, snapshot
+    ):
+        file_path = os.path.join(BASEDIR, media_file)
+        settings = {"Settings": {"MaxSpeakerLabels": max_speakers, "ShowSpeakerLabels": True}}
+
+        if max_speakers < 2:
+            with pytest.raises(ParamValidationError) as e:
+                transcribe_create_job(audio_file=file_path, params=settings)
+            snapshot.match("err_speaker_labels_diarization", e.value)
+            return
```
viren-nadkarni, line 428: It's a good practice to make a test only test a certain behaviour. It would be cleaner to assert this exception in a separate test.

tests/aws/services/transcribe/test_transcribe.py
```diff
@@ -403,3 +405,39 @@ def test_transcribe_start_job_same_name(
...
+
+        def _is_transcription_done():
+            transcription_status = aws_client.transcribe.get_transcription_job(
+                TranscriptionJobName=job_name
+            )
+            assert transcription_status["TranscriptionJob"]["TranscriptionJobStatus"] == "COMPLETED"
+
+        retry(_is_transcription_done, retries=50, sleep=2)
+
+        resp = aws_client.transcribe.get_transcription_job(TranscriptionJobName=job_name)
```
viren-nadkarni, line 440: You could eliminate the extra `GetTranscriptionJob` call as follows:

```suggestion
        def _is_transcription_done():
            r = aws_client.transcribe.get_transcription_job(
                TranscriptionJobName=job_name
            )
            assert r["TranscriptionJob"]["TranscriptionJobStatus"] == "COMPLETED"
            return r

        resp = retry(_is_transcription_done, retries=50, sleep=2)
```

tests/aws/services/transcribe/test_transcribe.py
```diff
@@ -403,3 +405,39 @@ def test_transcribe_start_job_same_name(
...
+        response = requests.get(resp["TranscriptionJob"]["Transcript"]["TranscriptFileUri"])
```
viren-nadkarni, line 441: A nice addon is to do this when using `requests`

```suggestion
        response = requests.get(resp["TranscriptionJob"]["Transcript"]["TranscriptFileUri"])
        response.raise_for_status()
```

sannya-singal:
Thanks for the review 🙌 I have slightly different viewpoint here, although I agree that we won't be able to reproduce the same timing and other numerical values without the actual models being used by AWS Transcribe, but in my opinion this test could be really helpful in kicking off the speaker diarization implementation in future, where we could atleast know the currently missing attributes and functionality considering the logged AWS responses, also making it easier to understand and implement the feature of atleast having `speaker_label` and the associated `transcripts` in the response and having other mocked values. 

This could be later modified according to the then implementation and skip the snapshot of certain functionality.

viren-nadkarni (approved):
Thanks for addressing the comments. Looks good to merge :+1:

PR #12412, opened 2025-03-19
Migrate legacy usage counters to new analytic counters

joe4dev:
## Motivation

Migrate Lambda and Lambda Event Source Mapping counters to our new standard.

## Changes

* Remove legacy counters
* Add hotreload counter for tracking how many Lambda functions are created with hot reloading enabled
* Add function counter for tracking Lambda functions by:
  * operation: invoke
  * runtime: AWS Lambda runtime such as python3.12 or `n/a` if `package_type` is `Image`
  * invocation_type: Event | RequestResponse
  * status: trying out a typed definition for distinguishing different error types
  * package_type: Zip | Image
* Add esm counter for tracking Lambda Event Source Mappings by:
  * status: success | partial_batch_failure_error | batch_failure_error | unhandled_error
  * source: e.g., aws:sqs

## Testing

* hotreload: `aws.services.lambda_.test_lambda_developer_tools.TestHotReloading.test_hot_reloading`
  * Manuel: Create a Lambda function with hot reloading enabled: https://docs.localstack.cloud/user-guide/lambda-tools/hot-reloading/#hot-reloading-for-python-lambdas
* function:
  * invocation_type:
    * `aws.services.lambda_.test_lambda.TestLambdaFeatures.test_invocation_type_request_response`
    * `aws.services.lambda_.test_lambda.TestLambdaFeatures.test_invocation_type_event`
  * runtime: `aws.services.lambda_.test_lambda_common.TestLambdaRuntimesCommon.test_echo_invoke`
  * status: many scenarios in `test_lambda.py`
* esm: `tests/aws/services/lambda_/event_source_mapping`

## Questions

a) hotreload>operation: Should we use the operations label (e.g., create) if we envison that at some point we might want to track other operations (e.g., reload). Unfortunately, that's a bit tricky in this case because we would need to do that in our custom [Lambda RIE](https://github.com/localstack/lambda-runtime-init/) Golang codebase.
b) function>operation: Should we add an operations label (e.g., invoke) if we envison to track create as well? If so, how to treat the invocation type label, which only applies for the invoke operation?
c) function>status: What do you think about using typed definitions for such a status label? Are we too fine-grained here?


vittoriopolverino:
> a) hotreload>operation: Should we use the operations label (e.g., create) if we envison that at some point we might want to track other operations (e.g., reload). Unfortunately, that's a bit tricky in this case because we would need to do that in our custom [Lambda RIE](https://github.com/localstack/lambda-runtime-init/) Golang codebase.

I’d suggest yes if `create` and `reload`, or other operations in general will share common characteristics and, in the future you want to track them in a more fine-grained way with additional labels/dimensions.


> b) function>operation: Should we add an operations label (e.g., invoke) if we envison to track create as well? If so, how to treat the invocation type label, which only applies for the invoke operation?

That’s a tough one, and it ties back to question (a) 🤣 

If `invocation_type` is the only label that isn’t shared across the different operations you want to track, we could compromise by using a placeholder like `'n/a'` for create events, for example.

However, if you anticipate that different operations will diverge and require operation-specific labels, then I’d suggest creating two separate metrics instead of using an operation label, even if the two metrics share some common characteristics.


> c) function>status: What do you think about using typed definitions for such a status label? Are we too fine-grained here?
What’s the cardinality here? How many different status values are possible? Are we talking hundreds, thousands, or something else?


**Additional Note:**
If it helps with your decision, I totally understand that it’s often difficult to anticipate how the measurement of a feature/service might evolve over time. That’s why I’m considering introducing a schema versioning mechanism to help mitigate this issue—allowing us to apply corrections to label usage (e.g., deprecations, additions, modifications, etc.).

That said, I wouldn’t want this to be over-relied upon or used as a fallback for every case. Ideally, we should still aim to anticipate future needs as much as possible.

If you’d like, we can have a quick chat and run some queries together to test different combinations or possibilities. That might give me a better understanding of your intentions and help refine the approach 💪🏼

tiurin:
> However, grouping them might be less convenient, in that case, we would need to introduce a general error attribute to group all error events != 'success'.

@vittoriopolverino to confirm, do you suggest here to have 2 separate labels - one for general operation result, i.e. success/error, and another one for specific reason for error?

tiurin:
@joe4dev @vittoriopolverino I added a draft PR for EventBridge Pipes: https://github.com/localstack/localstack-ext/pull/4224
Since Pipes and ESM share a part of infrastructure, namely pollers, we need to coordinate both changes and maybe unify enums like status, wdyt?

vittoriopolverino:
> I added a draft PR for EventBridge Pipes: https://github.com/localstack/localstack-ext/pull/4224
Since Pipes and ESM share a part of infrastructure, namely pollers, we need to coordinate both changes and maybe unify enums like status, wdyt?

From a data modeling perspective, if the enums have the same meaning in both EventBridge Pipes and ESM, it would make sense to ensure consistency across them.

That way, you can use a single filter for status across both Pipes and ESM, instead of needing two separate filters to represent the same status. (Not sure if this kind of cross-service analysis is something you’d ever do, but in any case, if we can ensure consistency, it’s better)

joe4dev:
> > to confirm, do you suggest here to have 2 separate labels - one for general operation result, i.e. success/error, and another one for specific reason for error?
> 
> Yes 👍🏼
> 
>     * Advantage: Having both successful and failed invocations in a single metric, since they share more common characteristics than differences.
> 
>     * Disadvantage: For successful invocations, the error reason label will be empty, but that’s an acceptable trade-off in this case.

One of the main question for me is what do we categorize as "error":
* AWS API error responses can be expected errors and part of the intended application behavior (e.g., partial batch failure handling in Lambda ESM because an application entry fails and then gets retried). These errors are expected AWS behavior and from a LocalStack perspective not actionable errors because we correctly implement AWS error parity.
* Unhandled errors are actionable for LocalStack because they indicate bugs in LocalStack and can guide maintenance effort to certain code paths (e.g., unhandled_state_error).
* Some errors can be user-side errors (e.g., pending_state_error: invoke a Lambda function that is not yet active)
* For some errors, we have insufficient information to categorize them and they could be user-side or LocalStack bugs (e.g., invocation_error or failed_state_error)

@vittoriopolverino @tiurin What is your understanding/suggestion regarding what should be considered an "error"?

vittoriopolverino:
> What is your understanding/suggestion regarding what should be considered an "error"?

It looks like you’ve already found the answer while writing this message! 😃 If an error is expected (part of intended application behavior), doesn’t require intervention, and isn’t something we need to act on, then maybe we need a way to differentiate errors.

We have two options:
1️⃣ Use a single status label with three categories: success, actionable_error, and non_actionable_error. The downside is that this makes it harder to filter all errors under a single error status.
2️⃣ Introduce an additional label, like is_error_actionable (true/false), while keeping status=error. This keeps all errors grouped but adds a second error-specific label alongside error_details.

Which approach makes the most sense really depends on the query patterns you anticipate using the most in the future. I don’t think there’s a strictly right or wrong choice between the two, it’s more about what will make querying and analysis more efficient for your use case

tiurin:
> From a data modeling perspective, if there are enums having the same meaning in both EventBridge Pipes and ESM, it would make sense to ensure consistency across them.

Pipes is a more broad case of ESM, moreover they have some shared code, so if we can align both well we will avoid "if pipes" conditions for counters.

tiurin (approved):
ESM part ✅ 🚀

vittoriopolverino (approved):
LGTM 🥳  Fantastic work, really appreciate the deep thinking that went into using labels effectively. Some interesting insights emerged while applying the new counter and structuring the labels 🚀

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1037,6 +1042,13 @@ def create_function(
             )
             fn.versions["$LATEST"] = version
             state.functions[function_name] = fn
+        function_counter.labels(
+            operation=FunctionOperation.create,
+            runtime=runtime,
```
joe4dev, line 1047: @dfangl : Detect `None` and add "n/a" for `package_type=Image`

Suggestion: add `or` "n/a"

dfangl (commented):
Don't we need a similar `or "n/a"` for the invokes in the function counters as well, not just the creation?

joe4dev:
> Don't we need a similar `or "n/a"` for the invokes in the function counters as well, not just the creation?

absolutely spot on @dfangl  💯 Thank you! 

I added the `or "n/a"` for invokes in the `lambda_service` (sync) and `event_manager` (async)  👍

gregfurman (approved):
LGTM. I think we should consider adding a counter for cold vs. warm starts since it appears to have an influence on performance of ESM. Not necessary in this PR though 🙂

PR #12322, opened 2025-03-02
Make SQS reject messages with empty body (#12317)

juanrh:
## Motivation
To fix #12317 

## Changes
Before this change localstack accepts SQS messages with empty body. 
After this changes, localstack rejects SQS messages with empty body with an error message "The request must contain the parameter MessageBody"

## Testing
Added a test case to an existing suite for SQS

Manual testing

```bash
make start

(.venv) (.venv) juanrh@barelinux:~/git/localstack$ QUEUE_URL=$(awslocal sqs create-queue --queue-name Test | jq -r .QueueUrl)
(.venv) (.venv) juanrh@barelinux:~/git/localstack$ $AWS  sqs send-message --queue-url $QUEUE_URL --message-body 'asdas'
{
    "MD5OfMessageBody": "0aa1ea9a5a04b78d4581dd6d17742627",
    "MessageId": "0087c1cd-019b-41d0-beb6-eeb043e76590"
}
(.venv) (.venv) juanrh@barelinux:~/git/localstack$ $AWS  sqs send-message --queue-url $QUEUE_URL --message-body ''

An error occurred (InvalidParameterValue) when calling the SendMessage operation: The request must contain the parameter MessageBody.
(.venv) (.venv) juanrh@barelinux:~/git/localstack$ 
```

I also did run `make format` and `make lint`

PR #12413, opened 2025-03-19
APIGW: fix rendering of $input.body when empty

bentsku:
## Motivation
We broke an upstream integration test following #12322, because API Gateway would forward an empty body to the SQS integration, and SQS disallows that.

This is because API Gateway will replace an empty body with `{}` when using `$input.body` in a VTL template.

I've added a few tests validating the assumption, and @dfangl added the SQS regression test to make sure we don't break upstream anymore. He also added some changes in the retry mechanism for deleting REST APIs, and I've updated it to use the Boto config in both creation and deletion of REST APIs. 

note: I've also tried with a request template set with `text/html` and the empty body is still `{}` so it's not dependent on the content-type. 

## Changes
- return a `{}` string if the body is empty for `$input.body` (so that it can be passed to function like `$util.urlEncode`)
- add a test with multiple cases
- add an SQS AWS integration test with empty body

### DIFF
--- a/localstack-core/localstack/testing/pytest/fixtures.py
+++ b/localstack-core/localstack/testing/pytest/fixtures.py
@@ -17,7 +17,6 @@ from _pytest.config import Config
 from _pytest.nodes import Item
 from botocore.exceptions import ClientError
 from botocore.regions import EndpointResolver
-from mypy_boto3_apigateway import APIGatewayClient
 from pytest_httpserver import HTTPServer
 from werkzeug import Request, Response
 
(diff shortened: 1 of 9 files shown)


localstack-core/localstack/testing/pytest/fixtures.py
```diff
@@ -2054,30 +2054,33 @@ class SampleStore(BaseStore):
...
         return api_id, response.get("name"), response.get("rootResourceId")
 
     yield _create_apigateway_function
 
-    for rest_api_id, region_name in rest_apis:
-        apigateway_client = aws_client_factory(region_name=region_name).apigateway
+    for rest_api_id, _client_region_name in rest_apis:
+        apigateway_client = aws_client_factory(
+            region_name=_client_region_name,
+            config=retry_boto_config,
```
cloutierMat, line 2082: good catch!

tests/aws/services/apigateway/test_apigateway_common.snapshot.json
```diff
@@ -1391,5 +1391,40 @@
       "to-string": "{foo=bar}",
       "list-to-string": "{list=[{\"foo\":\"bar\"}]}"
     }
+  },
+  "tests/aws/services/apigateway/test_apigateway_common.py::TestApiGatewayCommon::test_input_body_formatting": {
+    "recorded-date": "19-03-2025, 17:03:40",
+    "recorded-content": {
+      "response-raw-body": {
+        "empty-body": "b'{}'",
```
cloutierMat, line 1399: hmmm! interesting! nice find

tests/aws/services/apigateway/test_apigateway_common.snapshot.json
```diff
@@ -1391,5 +1391,40 @@
...
+      },
+      "response-str-body-all": {
+        "empty-body": "b'Action=SendMessage&MessageBody=Action=SendMessage&MessageBody={}'",
+        "json-body": "b'Action=SendMessage&MessageBody=Action=SendMessage&MessageBody={\"some\": \"value\"}'",
+        "str-body": "b'Action=SendMessage&MessageBody=Action=SendMessage&MessageBody=some raw data'"
+      },
+      "response-body-attr-access": {
+        "empty-body": "b'{}'",
+        "json-body": "b'{}'",
+        "str-body": "b'{}'"
```
cloutierMat, line 1421: I was trying to expect what would come out  of it... not what I thought! :rofl:

cloutierMat (approved):
Awesome! A clean one line fix for 300 lines of testing! :rocket: 

Thanks for deep dive and addressing this issue in VTL, it looks like we will never be out of surprise from it! :rofl:

PR #12415, opened 2025-03-19
[ESM] Add configurable poll frequency and log shard info

gregfurman:
## Motivation
In order to address performance shortcomings, we should allow for a user to configure their poller frequency as well as their maximum backoff on error/empty polls.

## Changes
- `StreamPollers` no longer back-off when a `GetRecords` call is empty
- `StreamPollers` back-off when no shards have been initialised.
- More debug logs have been added to show shard count.
- Adds the following config varibales:
   - `LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC` -> set all poller's per-second frequency.
   - `LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC` -> set maximum backoff on poller exception.
   - `LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_EMPTY_POLL_SEC` -> set maximum backoff on empty polls.

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -1071,6 +1071,24 @@ LAMBDA_INIT_POST_INVOKE_WAIT_MS = os.environ.get("LAMBDA_INIT_POST_INVOKE_WAIT_M
 # DEV: sbx_user1051 (default when not provided) Alternative system user or empty string to skip dropping privileges.
 LAMBDA_INIT_USER = os.environ.get("LAMBDA_INIT_USER")
 
+# INTERNAL: 1 (default)
+# The duration (in seconds) to wait between each poll call to an event source.
+LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC = float(
+    os.environ.get("LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC") or 1
+)
+
+# INTERNAL: 60 (default)
+# Maximum duration (in seconds) to wait between retries when an event source poll fails.
+LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC = float(
+    os.environ.get("LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC") or 60
+)
+
+# INTERNAL: 10 (default)
+# Maximum duration (in seconds) to wait between polls when an event source returns empty results.
+LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_EMPTY_POLL_SEC = float(
+    os.environ.get("LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_EMPTY_POLL_SEC") or 10
+)
+
 # Adding Stepfunctions default port
 LOCAL_PORT_STEPFUNCTIONS = int(os.environ.get("LOCAL_PORT_STEPFUNCTIONS") or 8083)
 # Stepfunctions lambda endpoint override
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -135,6 +135,12 @@ class StreamPoller(Poller):
         if not self.shards:
             self.shards = self.initialize_shards()
 
+        if not self.shards:
+            LOG.debug("No shards found for %s.", self.source_arn)
+            raise EmptyPollResultsException(service=self.event_source(), source_arn=self.source_arn)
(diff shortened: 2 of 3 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -164,7 +170,7 @@ def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
         records = get_records_response.get("Records", [])
         if not records:
             self.shards[shard_id] = get_records_response["NextShardIterator"]
-            raise EmptyPollResultsException(service=self.event_source(), source_arn=self.source_arn)
```
joe4dev, line 167: that gonna help a lot with responsiveness; good we caught this accidential delay introducer 😬

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -135,6 +135,12 @@ def poll_events(self):
         if not self.shards:
             self.shards = self.initialize_shards()
 
+        if not self.shards:
+            LOG.debug("No shards found for %s.", self.source_arn)
+            raise EmptyPollResultsException(service=self.event_source(), source_arn=self.source_arn)
+        else:
+            LOG.debug("Event source %s has %d shards.", self.source_arn, len(self.shards))
```
joe4dev, line 142: 👍

localstack-core/localstack/config.py
```diff
@@ -1071,6 +1071,24 @@ def populate_edge_configuration(
 # DEV: sbx_user1051 (default when not provided) Alternative system user or empty string to skip dropping privileges.
 LAMBDA_INIT_USER = os.environ.get("LAMBDA_INIT_USER")
 
+# INTERNAL: 1 (default)
+# The duration (in seconds) to wait between each poll call to an event source.
+LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC = float(
```
joe4dev, line 1076: 👍

localstack-core/localstack/config.py
```diff
@@ -1071,6 +1071,24 @@ def populate_edge_configuration(
...
+    os.environ.get("LAMBDA_EVENT_SOURCE_MAPPING_POLL_INTERVAL_SEC") or 1
+)
+
+# INTERNAL: 60 (default)
+# Maximum duration (in seconds) to wait between retries when an event source poll fails.
+LAMBDA_EVENT_SOURCE_MAPPING_MAX_BACKOFF_ON_ERROR_SEC = float(
```
joe4dev, line 1082: Thanks for marking this as internal 👍 

I understand that it can be helpful for quick experimentation with some customers, but we should be generally cautious making everything configurable, which introduces extra complexity. These are candidates to remove if there is no strong need in favor of choosing sensible defaults.

joe4dev (approved):

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
dfangl: This will effectively prevent backoff when querying empty streams, right?

dfangl (approved):
Looks good, just some minor comments

gregfurman:
@dfangl Unfortunately we can't reliably backoff when a `GetRecords` call is empty since this could be expected behabiour. I'll make a note of this as a comment but here's the relevant snippet from [GetRecords returns an empty records array even when there is data in the stream](https://docs.aws.amazon.com/streams/latest/dev/troubleshooting-consumers.html#getrecords-returns-empty):

> Consuming, or getting records is a pull model. Developers are expected to call [GetRecords](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetRecords.html) in a continuous loop with no back-offs. Every call to GetRecords also returns a ShardIterator value, which must be used in the next iteration of the loop.

Which goes on to say:
> [...] An empty Records element is returned under two conditions:
> 1. There is no more data currently in the shard.
> 2. There is no data near the part of the shard pointed to by the ShardIterator

And we can't reliably tell whether this is (1) or (2) which makes this back-off incorrect if we're doing pull-based retrieval.

dfangl:
@gregfurman Thank you for linking the docs! Yes - the empty get records call is something we also discussed yesterday. I would like to bring up two points for this one:
1. Against AWS this perfectly makes sense - however, did we check if kinesis-mock also has this behavior? Are we fixing a potential problem in theory, or did we actually experience the issue? To be clear, I am fine with removing this for now, and the ESM should really work against AWS as well, just asking out of interest.
2. This will again increase the number of requests against the gateway. It is fine, since it is not higher than a couple of weeks ago, but do we have data about what improvements we actually lose this way?

Issue #12419, opened 2025-03-20
bug: SNS Subscribe RawMessageDelivery=True only works first time


dngferreira:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

SNS subscrition with `RawMessageDelivery=True` works on first time but fails on following executions with 
> An error occurred (InvalidParameter) when calling the Subscribe operation: Invalid parameter: Attributes Reason: Subscription already exists with different attributes

### Expected Behavior

It should second+ executions should be a no-op since the attributes are the same.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker run localstack/localstack

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

> awslocal sns create-topic --name localstack-topic                                            
  awslocal sqs create-queue --queue-name localstack-queue
  awslocal sns subscribe   --topic-arn arn:aws:sns:us-east-1:112233445566:localstack-topic    --protocol sqs    --notification-endpoint arn:aws:sqs:us-east-1:112233445566:localstack-queue    --return-subscription-arn  --attributes RawMessageDelivery=True
  awslocal sns subscribe   --topic-arn arn:aws:sns:us-east-1:112233445566:localstack-topic    --protocol sqs    --notification-endpoint arn:aws:sqs:us-east-1:112233445566:localstack-queue    --return-subscription-arn  --attributes RawMessageDelivery=True

### Environment

```markdown
- OS: macos
- LocalStack:
  LocalStack version: 4.2.0
  LocalStack Docker image sha: sha256-2bb3f448d2a264f73ef4c4042a6374159d8e4d05f8fdd22a13b066bb17257e90
  LocalStack build date:
  LocalStack build git hash:
```

### Anything else?

caused on version 3.4.0 by PR #10575 changing the values to lower case and [SNS Provider ](https://github.com/localstack/localstack/blob/7fdfe017467612019db8a51458a80cd3ab74145a/localstack-core/localstack/services/sns/provider.py#L718) not taking it in consideration that the value is now stored as lower case.

PR #12420, opened 2025-03-20
SNS: fix Subscribe idempotency due to RawMessageDelivery casing

bentsku:
## Motivation
As reported by #12419, we had an issue with Subscribe idempotency due to `RawMessageDelivery` casing, value like `True` and `true` are consider equal as they should be lower-cased before comparison. 

This PR fixes the issue by lower-casing the attribute before comparison

## Changes
- adapt a test to cover the casing issue
- lower case the attribute before comparison

### DIFF
--- a/localstack-core/localstack/services/sns/provider.py
+++ b/localstack-core/localstack/services/sns/provider.py
@@ -1,4 +1,5 @@
 import base64
+import copy
 import functools
 import json
 import logging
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/sns/provider.py
```diff
@@ -696,26 +697,29 @@ def subscribe(
...
+            for attr_name, attr_value in sub_attributes.items():
                 validate_subscription_attribute(
                     attribute_name=attr_name,
                     attribute_value=attr_value,
                     topic_arn=topic_arn,
                     endpoint=endpoint,
                     is_subscribe_call=True,
                 )
+                if raw_msg_delivery := sub_attributes.get("RawMessageDelivery"):
+                    sub_attributes["RawMessageDelivery"] = raw_msg_delivery.lower()
```
k-a-il, line 711: nit: Interesting, is this the only boolean attribute that the user can provide?

k-a-il (approved):
Nice fix!

localstack-core/localstack/services/sns/provider.py
bentsku: Yes, most of the others are JSON strings or enum: https://docs.aws.amazon.com/sns/latest/api/API_SetSubscriptionAttributes.html

PR #12421, opened 2025-03-20
Cloud Formation: [POC] Support for Update Graph Modeling of: Conditions, Intrinsic Functions, and Type Divergences

MEPalma:
\* Depends on https://github.com/localstack/localstack/pull/12405

## Motivation
Currently, the POC Update Graph approach to modeling CloudFormation templates does not support CloudFormation Conditions. This gap prevents condition-driven workflows from being accurately tracked or updated within the existing logic. In addition, there is no support for certain intrinsic functions (such as Fn::If, Fn::Not, and Fn::Equals). Furthermore, the modeling layer does not currently handle scenarios where changes in a template result in a new subtype or intrinsic function (for example, switching from one intrinsic function to a different one). This limitation meant that strict type changes to certain template bindings was not supported. Finally, caching mechanisms within the describer visitor are needed. A number of minor issues with this pipeline were also addressed. Unit tests have been added to validate these scenarios.

## Changes
- Added base support for modeling CloudFormation Conditions in the Update Graph building logic
- Added modeling support for new intrinsic functions: Fn::If, Fn::Not, and Fn::Equals
- Enabled propagation of condition changes throughout the update graph, including condition resource statements and references in intrinsic functions
- Introduced the NodeDivergence concept, which models changes in subtype values (e.g., switching from one intrinsic function to another)
- Implemented caching in the describer visitor to improve performance and clarity in describing template changes
- Added support for describing change set changes involving conditions and the newly added intrinsic functions
- Created unit tests covering conditions, new intrinsic functions, and the NodeDivergence mechanism
- Several other minors

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -768,8 +768,7 @@ class ChangeSetModel:
     def _visit_parameters(
         self, scope: Scope, before_parameters: Maybe[dict], after_parameters: Maybe[dict]
     ) -> NodeParameters:
-        # FIXME: sampling should be able scope
-        node_parameters = self._visited_scopes.get(ParametersKey)
+        node_parameters = self._visited_scopes.get(scope)
         if isinstance(node_parameters, NodeParameters):
             return node_parameters
         parameter_names: list[str] = self._safe_keys_of(before_parameters, after_parameters)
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -80,53 +113,123 @@ class ChangeSetTerminal(ChangeSetEntity, abc.ABC): ...
...
+        conditions: NodeConditions,
+        resources: NodeResources,
+    ):
+        super().__init__(scope=scope, change_type=change_type)
+        self.parameters = parameters
+        self.conditions = conditions
         self.resources = resources
 
 
+class NodeDivergence(ChangeSetNode):
```
simonrw, line 134: note for the future: it will be nice to get some docstrings indicating what these resources represent (possibly with an example from a template)

PR #12432, opened 2025-03-24
CloudFormation: [POC] Support Update Graph Modeling of Mappings and FindInMap

MEPalma:
…indinmap

## Motivation
Currently, the POC Update Graph approach to modeling CloudFormation templates does not support CloudFormation Mappings. In addition, intrinsic functions that depend on mapping lookups (Fn::FindInMap) is also unsupported in the modeling layer, which restricted the accuracy of the update graph and impacted downstream processes relying on this data. This PR introduces initial support for Mappings and Fn::FindInMap modeling, ensuring that such constructs are correctly parsed, modeled, and propagated through the update graph. The changes also add support for change set describe workflows involving maps. Relevant initial unit tests have been included to validate these enhancements.

## Changes
- Added support for modeling CloudFormation Mappings within the Update Graph
- Implemented modeling support for the intrinsic function Fn::FindInMap
- Added support for Update Graph descriptions of changes in mappings through Fn::FindInMap

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -113,6 +113,7 @@ class ChangeSetTerminal(ChangeSetEntity, abc.ABC): ...
 
 
 class NodeTemplate(ChangeSetNode):
+    mappings: Final[NodeMappings]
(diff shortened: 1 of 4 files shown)


simonrw (approved):
Thanks for adding mapping support

PR #12437, opened 2025-03-25
[ESM] Support Stream Poller batching

gregfurman:
## Motivation
In order to reduce pressure on the LocalStack internal gateway, events being collected by ESM/Pipes pollers should be properly collected and and sent onto a target as a batch.

## Changes
- Uses new `Batcher` utility that allows us to collect and batch records based on an ESM's `MaximumBatchingWindowInSeconds` and `BatchSize` configuration.
- Adds a `shard_batcher` dictionary to the `StreamPoller` that implements a batcher for each shard -- removing those batchers that do not correspond to shards each poll_event iteration.

## Performance

We ran some performance tests comparing baseline LocalStack with implemented **Batching** and **Parallelised** invocations.

### Batching

- Proper batching was implemented such that we only invoke a function when a `BatchSize` is met or a `MaximumBatchingWindowInSeconds` is reached

### Parallelised

- Polling of each shard was done in parallel
- `ParallelizationFactor` allowed for batches to be split by partition key and sent onto a target lambdas concurrently

### Experiments

- Batch Size = 100
- Batch Window = 20s
- No. Streams/Functions/ESMs = 20/20/20
- No. Partition Keys = 2
- No. Shards per Stream = 2

<details>
<summary>Expand for graph of average latency per `Lambda::Invoke` and `Kinesis::GetRecords`</summary>

image

</details>

### Results

Batching showed improved lambda invocation performance against baseline LocalStack -- with the average latency of `Invoke` both:
- Improving over baseline and parallel and;
- Seemingly not increasing as more locust clients were added -- indicating that batching assists the scalability of adding multiple producer clients to LocalStack.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -202,6 +202,7 @@ class StreamPoller(Poller):
         for batch in batched(collected_records, self.stream_parameters.get("BatchSize")):
             # This could potentially lead to data loss if forward_events_to_target raises an exception after a flush
             # which would otherwise be solved with checkpointing.
+            # TODO: Implement checkpointing, leasing, etc. from https://docs.aws.amazon.com/streams/latest/dev/kcl-concepts.html
             self.forward_events_to_target(shard_id, next_shard_iterator, batch)
 
     def forward_events_to_target(self, shard_id, next_shard_iterator, records):
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -80,8 +80,8 @@ class StreamPoller(Poller):
 
         self.shard_batcher = defaultdict(
             lambda: Batcher(
-                max_count=self.stream_parameters.get("BatchSize"),
-                max_window=self.stream_parameters.get("MaximumBatchingWindowInSeconds"),
+                max_count=self.stream_parameters.get("BatchSize", 100),
+                max_window=self.stream_parameters.get("MaximumBatchingWindowInSeconds", 0),
             )
         )
 
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -2,6 +2,7 @@ import json
 import logging
 import threading
 from abc import abstractmethod
+from collections import defaultdict
 from datetime import datetime
 from typing import Iterator
 
@@ -29,8 +30,12 @@ from localstack.services.lambda_.event_source_mapping.pollers.poller import (
     get_batch_item_failures,
 )
 from localstack.services.lambda_.event_source_mapping.pollers.sqs_poller import get_queue_url
+from localstack.services.lambda_.event_source_mapping.senders.sender_utils import (
+    batched,
+)
 from localstack.utils.aws.arns import parse_arn, s3_bucket_name
 from localstack.utils.backoff import ExponentialBackoff
+from localstack.utils.batch_policy import Batcher
 from localstack.utils.strings import long_uid
 
 LOG = logging.getLogger(__name__)
@@ -53,6 +58,9 @@ class StreamPoller(Poller):
     # Used for backing-off between retries and breaking the retry loop
     _is_shutdown: threading.Event
 
+    # Collects and flushes a batch of records based on a batching policy
+    shard_batcher: dict[str, Batcher[dict]]
+
     def __init__(
         self,
         source_arn: str,
@@ -69,6 +77,13 @@ class StreamPoller(Poller):
         self.iterator_over_shards = None
 
         self._is_shutdown = threading.Event()
+
+        self.shard_batcher = defaultdict(
+            lambda: Batcher(
+                max_count=self.stream_parameters.get("BatchSize"),
+                max_window=self.stream_parameters.get("MaximumBatchingWindowInSeconds"),
+            )
+        )
 
     @abstractmethod
     def transform_into_events(self, records: list[dict], shard_id) -> list[dict]:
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -140,6 +155,9 @@ def poll_events(self):
             raise EmptyPollResultsException(service=self.event_source(), source_arn=self.source_arn)
         else:
             LOG.debug("Event source %s has %d shards.", self.source_arn, len(self.shards))
+            # Remove all shard batchers without corresponding shards
+            for shard_id in self.shard_batcher.keys() - self.shards.keys():
```
joe4dev, line 159: Do we need to log and synchronize the shard batchers upon every poll?
Wouldn't it be sufficient to do that after after shard (re-)initialization?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -165,18 +183,31 @@ def poll_events(self):
...
+        if not should_flush:
+            self.shards[shard_id] = next_shard_iterator
             return
 
+        # Retrieve and drain all events in batcher
+        collected_records = self.shard_batcher[shard_id].flush()
+        # If there is overflow (i.e 1k BatchSize and 1.2K returned in flush), further split up the batch.
+        for batch in batched(collected_records, self.stream_parameters.get("BatchSize")):
+            # This could potentially lead to data loss if forward_events_to_target raises an exception after a flush
+            # which would otherwise be solved with checkpointing.
```
joe4dev, line 204: Can we add a TODO with an actionable suggestion on how to fix this (i.e., implement checkpointing)?

💡 The [KCL concepts](https://docs.aws.amazon.com/streams/latest/dev/kcl-concepts.html) page could give some hints/insights on how to design a robust stream poller.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -165,18 +183,31 @@ def poll_events(self):
...
+            self.forward_events_to_target(shard_id, next_shard_iterator, batch)
+
+    def forward_events_to_target(self, shard_id, next_shard_iterator, records):
```
joe4dev, line 207: Good refactoring 👍

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -165,18 +183,31 @@ def poll_events(self):
             pass
 
     def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
-        abort_condition = None
         get_records_response = self.get_records(shard_iterator)
```
joe4dev, line 186: 💡  Follow up (out of scope): Moving forward, we might need to think about a way how call `get_records` multiple times and take the `IteratorAge` into consideration without waiting for the inter-poll interval.

joe4dev (approved):
Adding support for `MaximumBatchingWindowInSeconds` is a great addition that can help to reduce unnecessary Lambda invocations with small batch sizes.

Nice work 👏👏 . I like how clean it looks with the batcher abstraction. I still think we need to re-think polling around shard management to enable more frequent or adaptive `get_records` calls. This PR is a good step in the right direction 📈.

The benchmark results are not very clear to me. What does improving over the baseline mean? fewer Lambda invokes due to batching?

My main question is around frequent logging and shard batch synchronization.

Can we trigger an ext run to ensure all Pipe tests are 🟢 ?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -165,18 +183,31 @@ def poll_events(self):
...
+        # However, we still need to check if batcher should be triggered due to time-based batching.
+        should_flush = self.shard_batcher[shard_id].add(records)
+        if not should_flush:
+            self.shards[shard_id] = next_shard_iterator
             return
 
+        # Retrieve and drain all events in batcher
+        collected_records = self.shard_batcher[shard_id].flush()
+        # If there is overflow (i.e 1k BatchSize and 1.2K returned in flush), further split up the batch.
+        for batch in batched(collected_records, self.stream_parameters.get("BatchSize")):
```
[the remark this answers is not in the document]
gregfurman: There is actually functionality for this -- I just thought using a `while` here looked strange and didn't want to add in a generator as a part of the first iteration of the batcher 😄

If you do a partial batch flush with a while loop, it'll achieve the same thing:

```python
        while batch := self.shard_batcher[shard_id].flush(partial=True):
```

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
gregfurman: Not necessarily but since the re-initialization method is abstract, and defined in the child classes, I thought that this would be the cleanest way of syncing up the batcher.

We could add this after every `self.initialize_shard()` call as well? Else we can change the `initialize_shard` from being `abstract` to always do this re-sync (and log).

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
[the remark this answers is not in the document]
tiurin: thanks, yeah, would be great to iterate on that - would be nice if `Batcher` could handle such use cases in an obvious way without extra functions. I agree on `while` being not too obvious, but both `batcher` and `batched` makes it a bit difficult to read if you don't know the difference upfront!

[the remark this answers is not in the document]
gregfurman: We can create a generator for this then. Next time!

gregfurman:
> The benchmark results are not very clear to me. What does improving over the baseline mean? fewer Lambda invokes due to batching?

If you look at the `operation="Invoke"` part of the figure, the processing time of each `Lambda::invoke` request is consistently less than HEAD of master. There are both fewer invokes as well as each invoke completing faster. Does that explain it a bit better?

PR #12441, opened 2025-03-26
Fix flaky lambda test event retry reserved concurrency

joe4dev:
## Motivation

The test `test_reserved_concurrency_async_queue` is currently skipped due to flakiness.

## Changes

* Add `aws_client_no_retry` fixture and use it in the testcase
* Re-design test case to use SQS-based notifications instead of unpredictable sleep
* Add `requests_id`-based assertions

## Testing

Temporarily using a botocore configuration with increased retries makes the retry problem reproducible causing the test to fail. This shows that we need an `aws_client_no_retry` to mitigate potential flakiness and reduce unnecessary retries.

To reproduce:
1. Ensure that retries are not disabled by configuration: `TEST_DISABLE_RETRIES_AND_TIMEOUTS=0`
2. Use the following config:

```py
retry_delay_config = botocore.config.Config(retries={"max_attempts": 10})
retry_delay_client = aws_client_factory(config=retry_delay_config)
with pytest.raises(aws_client.lambda_.exceptions.TooManyRequestsException) as e:
    retry_delay_client.lambda_.invoke(
        FunctionName=fn_arn, InvocationType="RequestResponse"
    )
```

### DIFF
--- a/localstack-core/localstack/testing/pytest/fixtures.py
+++ b/localstack-core/localstack/testing/pytest/fixtures.py
@@ -67,6 +67,19 @@ if TYPE_CHECKING:
     from mypy_boto3_sqs.type_defs import MessageTypeDef
 
 
+@pytest.fixture(scope="session")
+def aws_client_no_retry(aws_client_factory):
+    """
+    This fixture can be used to obtain Boto clients with disabled retries for testing.
+    botocore docs: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#configuring-a-retry-mode
+
+    Use this client when testing exceptions (i.e., with pytest.raises(...)) or expected errors (e.g., status code 500)
+    to avoid unnecessary retries and mitigate test flakiness if the tested error condition is time-bound.
+    """
+    no_retry_config = botocore.config.Config(retries={"max_attempts": 1})
+    return aws_client_factory(config=no_retry_config)
+
+
 @pytest.fixture(scope="class")
 def aws_http_client_factory(aws_session):
     """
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/testing/pytest/fixtures.py
```diff
@@ -67,6 +67,19 @@
     from mypy_boto3_sqs.type_defs import MessageTypeDef
 
 
+@pytest.fixture(scope="session")
+def aws_client_no_retry(aws_client_factory):
+    """
+    This fixture can be used to obtain Boto clients with disabled retries for testing.
+    botocore docs: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#configuring-a-retry-mode
+
+    Use this client when testing exceptions (i.e., with pytest.raises(...)) or expected errors (e.g., status code 500)
```
dfangl, line 76: Maybe we can clarify here, that this is mostly needed when exceptions are tested which have retries? So all listed in here: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#legacy-retry-mode (or matching the http status codes also listed there).
For a "ResourceNotFound" exception for example, this is not necessary

dfangl (approved):
LGTM, nice fix of the test!

localstack-core/localstack/testing/pytest/fixtures.py
joe4dev: yeah, we are indeed using the legacy retry mode, which makes things even worse 😬 .

I wanted to keep it simple and long-term given there are no adverse effects of using the no_retry client with a non-retrying exception. Nevertheless, I added the clarification for most accurate and actionable advice at the current status.

PR #12443, opened 2025-03-26
CloudFormation: POC Support for Modeling of Outputs Blocks in the Update Graph, Improved Handling of Intrinsic Function Types

MEPalma:
## Motivation
The current strategy for modeling CloudFormation templates using the Update Graph does not support changes in Outputs blocks. Additionally, the ChangeSetModel class implements a somewhat tedious approach to visiting and modeling intrinsic function types, due to their natural representation as bindings within dictionaries/objects.
This PR introduces support for modeling changes in Outputs blocks within the Update Graph, along with temporary logic to generate descriptions of those changes. The new functionality is covered by several new unit test cases.
Furthermore, this PR improves the parsing and modeling of intrinsic functions by making more extensive use of the recently introduced TypeDivergence node class.
Several additional issues related to intrinsic parsing and the describe logic are also addressed.

## Changes
- Added support for modeling changes in Outputs blocks within the Update Graph.
- Introduced temporary logic for describing Outputs changes.
- Added new unit tests covering Outputs diff scenarios.
- Refactored intrinsic function parsing using the TypeDivergence node class.
- Simplified modeling of intrinsic function types in ChangeSetModel.
- Fixed issues in intrinsic parsing and describe logic.


pinzon (commented):
:+1:

tests/unit/services/cloudformation/test_change_set_describe_details.py
```diff
@@ -1222,3 +1245,192 @@ def test_condition_update_production_remove_resource(self):
...
+                "NewParam": {
+                    "Type": "AWS::SSM::Parameter",
+                    "Properties": {"Name": "param-name", "Type": "String", "Value": "value-1"},
+                },
+            },
+            "Outputs": {"NewParamName": {"Value": {"Ref": "NewParam"}}},
+        }
+        outputs_unit = self.debug_outputs(t1, t2)
+        assert not outputs_unit.before_context
+        assert outputs_unit.after_context == [{"Name": "NewParamName", "Value": "NewParam"}]
```
simonrw, line 1273: This is not correct. The value should be the `Name` of the parameter, which is autogenerated by CloudFormation. See the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html#aws-resource-ssm-parameter-return-values). So it should be `param-name`.

tests/unit/services/cloudformation/test_change_set_describe_details.py
```diff
@@ -1222,3 +1245,192 @@ def test_condition_update_production_remove_resource(self):
...
+            "Resources": {
+                "UnrelatedParam": {
+                    "Type": "AWS::SSM::Parameter",
+                    "Properties": {"Name": "unrelated-param", "Type": "String", "Value": "foo"},
+                }
+            }
+        }
+        outputs_unit = self.debug_outputs(t1, t2)
+        assert outputs_unit.before_context == [
+            {"Name": "FeatureToggleName", "Value": "FeatureToggle"}
```
simonrw, line 1303:
```suggestion
            {"Name": "FeatureToggleName", "Value": "app-feature-toggle"}
```

simonrw (changes requested):
The test assertions are not correct. This is the advantage of using snapshot tests.

MEPalma:
In this PR, the outputs are evaluated solely by the describer. Since the POC doesn’t yet include executor logic, we still want to validate that Outputs blocks are being parsed correctly and mapped to the intended resources. Given that, the describer currently reports the entity name as a proxy for this verification.

I’d like to keep this logic within the describer for now and plan to move it to the template processor in that very spike, where it can be overridden appropriately by the executor. Does that sound reasonable to you?

simonrw (approved):
Ok fair thanks for the explanation @MEPalma . I am happy there is a path forward, but perhaps can we at least add a comment explaining?

PR #12444, opened 2025-03-26
Make event invoke function update test more predicatable

joe4dev:
Depends on: https://github.com/localstack/localstack/pull/12441

## Motivation

The test `test_async_invoke_queue_upon_function_update` has been flaky in CircleCI (e.g., [old example](https://app.circleci.com/pipelines/github/localstack/localstack/27256/workflows/07361ef1-6096-46f0-a007-aef7d094e7ae/jobs/233909)) and locally (assertion error `14 != 15`).

## Changes

* Migrate test to new SQS-based lambda notifier infrastructure introduced in https://github.com/localstack/localstack/pull/12441
  * This allows us to wait until the first invoke is actually being processed rather than using a flaky sleep.
  * This should speeds up the test because the SQS message collection uses long polling
* Migrate to label-based snapshots for easier snapshot skip

## TODO

- [ ] Fix buggy LocalStack behavior not respecting the SQS message deletion and therefore causing a snapshot key issue (guarded via assert). It seems that the SQS message does not get deleted properly although the delete_message call was successful 🤔 .
  * Increasing the visibility timeout didn't help `sqs_create_queue(Attributes={"VisibilityTimeout": "90"})` (the default 20s should be ok as long as the cold start doesn't take more than 10s)


alexrashed:
@joe4dev This one seems pretty outdated... Do you want to revive it, or close it? :)

joe4dev:
I think its still a valuable improvement, but I don't have capacity to fix the blocker atm (might be part of Lambda Gold later this year).

joe4dev:
Closing as migrated to https://github.com/localstack/localstack-core/pull/23

Issue #12448, opened 2025-03-28
bug: `content-length-range` not respected


paciops:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The condition `content-length-range` is not respect in localstack S3.

Lunching localstack with `docker run --rm -p 4566:4566 -e localstack/localstack:s3-latest` and this test in node.js https://gist.github.com/paciops/c69e6d7dee4b0454758b2bf45d0c5bad shows how it is possible to upload files with size out of range.
Test log: 
```bash
Bucket created
▶ File Upload
  ✖ should not upload a oversize file with POST (47.503439ms)
    AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
    
    204 !== 400
    
        at TestContext.<anonymous> (/home/test/integration/gist.ts:87:5)
        at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
        at async Test.run (node:internal/test_runner/test:935:9)
        at async Promise.all (index 0)
        at async Suite.run (node:internal/test_runner/test:1320:7)
        at async startSubtestAfterBootstrap (node:internal/test_runner/harness:296:3) {
      generatedMessage: true,
      code: 'ERR_ASSERTION',
      actual: 204,
      expected: 400,
      operator: 'strictEqual'
    }

  ✖ should not upload a undersize file with POST (74.76786ms)
    AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
    
    204 !== 400
    
        at TestContext.<anonymous> (/home/test/integration/gist.ts:107:5)
        at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
        at async Test.run (node:internal/test_runner/test:935:9)
        at async Suite.processPendingSubtests (node:internal/test_runner/test:633:7) {
      generatedMessage: true,
      code: 'ERR_ASSERTION',
      actual: 204,
      expected: 400,
      operator: 'strictEqual'
    }

✖ File Upload (407.059509ms)
ℹ tests 2
ℹ suites 1
ℹ pass 0
ℹ fail 2
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1505.248712

```

### Expected Behavior

Responses should be 400 and 204

### How are you starting LocalStack?

docker run --rm -p 4566:4566 -e localstack/localstack:s3-latest

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```
[...]

PR #12449, opened 2025-03-28
S3: fix casing of PreSignedPost validation

bentsku:
## Motivation
As reported by #12448, we had an issue with the JS AWS SDK v3 accepting some conditions that shouldn't have been validated. This is because we skipped validating the request if the request did not have a `policy` field, lower-cased. But the JS SDK sends it as `Policy`. 

## Changes
- update the logic to be case insensitive while validating certain fields of the Post request

_fixes #12448_


k-a-il (approved):
Nice fix :) all keys used to query the dictionary were lower cased

PR #12455, opened 2025-03-31
Admin: Improve license metadata

bblommers:
## Motivation
With [PEP 639](https://peps.python.org/pep-0639/), the latest setuptools (>= 77.0) throws a warning when building projects with a deprecated license-format. This PR updates the format to the latest recommended format, removing all warnings.

The [Python Packaging docs](https://packaging.python.org/en/latest/specifications/core-metadata/#classifier-multiple-use) now also explicitly states that this is the recommended format:

| The use of License :: classifiers is deprecated as of Metadata 2.4, use License-Expression instead. See [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dynamic = ["version"]
 classifiers = [
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.11",
-    "License :: OSI Approved :: Apache Software License",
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Awesome! Thanks a lot for jumping on the deprecation warning and migrating to the new expression before it actually hits us! 💯

silv-io (approved):

Issue #12377, opened 2025-03-12
LocalStack does not detect subnet tags with Terraform


jose-skoolscout:
### Description:

**Problem Description:**  
I am using **LocalStack** version `4.2` to perform AWS infrastructure testing with **Terraform**. However, when creating subnets in LocalStack, the **tags** are not correctly applied to the subnets, even though the tags are properly defined in Terraform.

---

### Steps to Reproduce:

1. I created a VPC in LocalStack.
2. I tried to create private subnets with tags using Terraform.
3. I used the following Terraform code to define the private subnets:

   ```hcl
   resource "aws_subnet" "private" {
     count             = var.az_count
     cidr_block        = cidrsubnet(aws_vpc.main.cidr_block, 8, count.index)
     availability_zone = data.aws_availability_zones.available.names[count.index]
     vpc_id            = aws_vpc.main.id
     tags = {
       Domain      = "example.com"
       Environment = "shared"
       Application = "example-com"
       Name        = "ts-sub-net-priv-prod"
     }
   }

### Output of the created subnets
```json private_subnet_ids = [
  {
    "arn": "arn:aws:ec2:us-east-1:000000000000:subnet/subnet-ad9e8d62f9513fef3",
    "assign_ipv6_address_on_creation": false,
    "availability_zone": "us-east-1a",
    "availability_zone_id": "use1-az6",
    "cidr_block": "172.17.0.0/24",
    "customer_owned_ipv4_pool": "",
    "enable_dns64": false,
    "enable_lni_at_device_index": 0,
    "enable_resource_name_dns_a_record_on_launch": false,
    "enable_resource_name_dns_aaaa_record_on_launch": false,
    "id": "subnet-ad9e8d62f9513fef3",
    "ipv6_cidr_block": "",
    "ipv6_cidr_block_association_id": "",
    "ipv6_native": false,
    "map_customer_owned_ip_on_launch": false,
    "map_public_ip_on_launch": false,
    "outpost_arn": "",
    "owner_id": "000000000000",
    "private_dns_hostname_type_on_launch": "ip-name",
    "tags": {},
    "tags_all": {},
    "timeouts": null,
    "vpc_id": "vpc-2accbbae3126d386c"
  },
  {
    "arn": "arn:aws:ec2:us-east-1:000000000000:subnet/subnet-9d6e838804f9b548a",
    "assign_ipv6_address_on_creation": false,
    "availability_zone": "us-east-1b",
    "availability_zone_id": "use1-az1",
[...]

PR #12459, opened 2025-03-31
Fix subnet tags going missing

jw2:
## Motivation
#11853 introduced a regression whereby the structure of the tags dict was altered before passing through to moto.  This was reported in #12377 and is fixed by this PR.

Whilst fixing this PR, I noticed that the tags were also lost when creating a subnet, SG or VPC with a custom ID.

## Changes
* Pass the original tags dict through to moto on subnet creation, rather than the modified one.
* Because tags are not stored within the subnet, VPC or SG objects within moto, but instead in a separate tags dict, the tags needed to be reindexed in that dict when using a custom id.

### DIFF
--- a/localstack-core/localstack/services/ec2/patches.py
+++ b/localstack-core/localstack/services/ec2/patches.py
@@ -1,5 +1,4 @@
 import logging
-from re import sub
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


jw2:
recheck

PR #12461, opened 2025-03-31
Step Functions: Migrate v2 Test Suite to no_retry aws_client Fixture

MEPalma:
## Motivation
To minimize unnecessary retries of boto actions, a `_no_retry` variant of the `aws_client` boto client factory fixture was recently introduced in https://github.com/localstack/localstack/pull/12441. 
These changes aim to migrate the SFN v2 test suite to use this new fixture. By doing so, we eliminate redundant retries that occur during negative snapshot tests, but also in potential flaky tests in the test suite.

## Changes
- Migrate v2 test suite from `aws_client` boto client factory to `aws_client_no_retry`


joe4dev (changes requested):
Good idea to pro-actively adopt the new `aws_client_no_retry` 🚀 

However, we should avoid replacing all clients and use `aws_client_no_retry` only "when testing exceptions (i.e., with pytest.raises(...)) or expected errors (e.g., status code 500)" https://github.com/localstack/localstack/pull/12441/files#diff-2776b578d9cfe9e8eaed135a0f652801d1e00d4bfee6eed855d6d097fc33f8c0R76
Retries actually help to reduce flakiness, especially when running tests against AWS.

All credits go to @dfangl for noticing this :)

MEPalma:
Thank you! I have updated these changes to only use the `_no_retry` variant in negative scenarios.
I understand that AWS can be flaky in positive scenarios, but I’d also consider that negative scenarios may not always be immune to flakiness—something that may be interesting to reflect on for our testing approach. 
My intention was to ensure that the SFN service is exact on the first shot, given that LocalStack has no reason to behave otherwise, unlike AWS. Moving forward, I’d like the test suite to adopt this approach for LocalStack, but consider disabling the one-shot behavior when testing against AWS.

tests/aws/services/stepfunctions/v2/test_sfn_api_logs.py
```diff
@@ -142,11 +142,8 @@ def test_invalid_logging_configuration(
 
         sm_name = f"statemachine_{short_uid()}"
 
-        stepfunctions_client = aws_client_factory(
-            config=Config(parameter_validation=False)
```
joe4dev, line 146: nice catch that parameter validation disabling didn't seem necessary here 👍

joe4dev (approved):
This change highlights the extensive test coverage of exception cases for StepFunctions 🙌  ✨ 

> I’d also consider that negative scenarios may not always be immune to flakiness—something that may be interesting to reflect on for our testing approach.

Very good point and something to observe while updating snapshots. Given this process is currently manual, I don't think it's a problem for now. Re-running the entire test using `pytest-rerunfailures` could be a suitable alternative here to a conditional client configuration.

PR #12465, opened 2025-04-01
Fix lambda timeout race condition

dfangl:
## Motivation
We recently discovered that a race condition is possible, killing in-progress invocations of lambda functions if the environment is scheduled for removal due to inactivity at the same time.

This will lead to different errors (depending on the phase of the execution), and has to be fixed.

Also, the error messages we return to the caller in case of internal errors during invocations do not convey the actual error - making it harder to trace the error, especially for internal calls.
Here we fix this, to clearly communicate the error to the caller.

Old error message:
```
Internal error while executing lambda
```

New error message (containing request id and function name + qualifier):

```
[fd2d870e-827c-44a2-b365-2eb7efc38e86] Internal error while executing lambda test-function-28f7fd59:None. Caused by AssignmentException: Could not start new environment: NoSuchImage:public.ecr.aws/lambda/python:3.12
```

The error logging in various places has also been improved to log the exception type as well as the message.

## Changes
* Fix race condition during environment timeout
* Improve error messages

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
+++ b/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
@@ -275,6 +275,9 @@ class ExecutionEnvironment:
             self.id,
             self.function_version.qualified_arn,
         )
+        # The stop() method allows to interrupt invocations (on purpose), which might cancel running invocations
+        # which we should not do when the keepalive timer passed.
+        # The new TIMING_OUT state prevents this race condition
         with self.status_lock:
             if self.status != RuntimeStatus.READY:
                 LOG.debug(
--- a/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
+++ b/localstack-core/localstack/services/lambda_/invocation/execution_environment.py
@@ -37,10 +37,11 @@ class RuntimeStatus(Enum):
     INACTIVE = auto()
     STARTING = auto()
     READY = auto()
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -274,6 +275,14 @@ def keepalive_passed(self) -> None:
             self.id,
             self.function_version.qualified_arn,
         )
+        with self.status_lock:
+            if self.status != RuntimeStatus.READY:
+                LOG.debug(
+                    "Keepalive timer passed, but current runtime status is %s. Aborting keepalive stop.",
+                    self.status,
+                )
+                return
```
joe4dev, line 284: Great catch ✨

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -274,6 +275,14 @@ def keepalive_passed(self) -> None:
             self.id,
             self.function_version.qualified_arn,
         )
+        with self.status_lock:
```
joe4dev, line 278: docs: Would it make sense to add some explanation about the motivation why this is needed to guard against the race condition?

dfangl: Agreed, added!

PR #12467, opened 2025-04-02
KMS: fix RSA PSS signing issue for salt length

sannya-singal:
## Motivation
When using an RSA key to sign with PSS algorithm, currently the salt length is set to be `MAX_LENGTH`, which is not in accordance with the ones required by RFC 4055 making making it difficult to verify such signatures across platforms: https://datatracker.ietf.org/doc/html/rfc4055#section-3.3. 

Closes https://github.com/localstack/localstack/issues/9602

## Changes
This PR: 
- sets the salt length appropriately for the signing algorithm being used.
- adds an AWS validated test to validate the changes to sign with an aws client and verify the signature using `cryptography` library.

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -10,7 +10,7 @@ import pytest
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from cryptography.hazmat.primitives import hashes, hmac, serialization
-from cryptography.hazmat.primitives.asymmetric import ec, padding
+from cryptography.hazmat.primitives.asymmetric import ec, padding, utils
 from cryptography.hazmat.primitives.serialization import load_der_public_key
 
 from localstack.services.kms.models import IV_LEN, Ciphertext, _serialize_ciphertext_blob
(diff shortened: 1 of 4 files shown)


silv-io (approved):
LGTM! First I got a bit confused because Section 3.3 in the RFC doesn't mention the recommended salt length. However, 3.1 does:

```
saltLength

         The saltLength field is the octet length of the salt.  For a
         given hashAlgorithm, the recommended value of saltLength is the
         number of octets in the hash value.  Unlike the other fields of
         type RSASSA-PSS-params, saltLength does not need to be fixed
         for a given RSA key pair; a different value could be used for
         each RSASSA-PSS signature generated.
```

So using the digest (hash) length here is indeed more correct.

Awesome work on the test as well! :)

tests/aws/services/kms/test_kms.py
```diff
@@ -724,6 +745,40 @@ def test_sign_verify(self, kms_create_key, snapshot, key_spec, sign_algo, aws_cl
...
+        [
+            ("RSA_2048", "RSASSA_PSS_SHA_256"),
+            ("RSA_2048", "RSASSA_PSS_SHA_384"),
+            ("RSA_2048", "RSASSA_PSS_SHA_512"),
+            ("RSA_4096", "RSASSA_PKCS1_V1_5_SHA_256"),
+            ("RSA_4096", "RSASSA_PKCS1_V1_5_SHA_512"),
+            ("ECC_NIST_P256", "ECDSA_SHA_256"),
+            ("ECC_NIST_P384", "ECDSA_SHA_384"),
+            ("ECC_SECG_P256K1", "ECDSA_SHA_256"),
+        ],
```
k-a-il, line 760: nice testing

tests/aws/services/kms/test_kms.snapshot.json
```diff
@@ -2152,5 +2152,37 @@
...
+    "recorded-date": "02-04-2025, 06:06:54",
+    "recorded-content": {}
+  },
+  "tests/aws/services/kms/test_kms.py::TestKMS::test_verify_salt_length[RSA_2048-RSASSA_PSS_SHA_512]": {
+    "recorded-date": "02-04-2025, 06:06:57",
+    "recorded-content": {}
+  },
+  "tests/aws/services/kms/test_kms.py::TestKMS::test_verify_salt_length[RSA_4096-RSASSA_PKCS1_V1_5_SHA_256]": {
+    "recorded-date": "02-04-2025, 06:06:59",
+    "recorded-content": {}
```
k-a-il, line 2170: The snapshots are empty—did you forget to add them?

k-a-il (changes requested):

tests/aws/services/kms/test_kms.snapshot.json
sannya-singal: No, these are just run against AWS. We are not snapshotting any responses for this test.

k-a-il: Does not it make sense to snapshot some responses? for example response from `kms.sign`?

sannya-singal: Not sure if we should add this because firstly the signature changes on every run of thus test which means we will have to transform the response which does not make sense, and secondly we already snapshot sign response in `test_sign_verify`.

k-a-il: Alright, I was thinking we could snapshot the sign algorithm, for example. But if that doesn’t make sense, we can leave it out for now

k-a-il (approved):
Nice work! thank you for doing that

PR #12470, opened 2025-04-02
Reapply reduce requests necessary for log publishing from lambda to cloudwatch logs

dfangl:
## Motivation
With #10294 we had to revert increased performance for lambda cloudwatch logs, reduced requests necessary for every log storage from 3 to (on average) 1.

The root cause for this revert was the `.splitlines()` method used to split the logs, as it also splits on several other characters, like a carriage return, unlike the old log method.

Since the runtime interface clients patch the loggers (e.g. here: https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/2ea9a34b5c1b15621df132eb2d377ac54b330c8f/src/LogPatch.js#L82-L88), we need to leave carriage returns as is for now.

In the long term, to further increase logging parity, we need to:
1. Properly log the `INIT_START` message (this should allow for proper snapshotting of logs as well then, as the count should align)
2. Use the file descriptor based logging mechanism (counterpart to this: https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/2ea9a34b5c1b15621df132eb2d377ac54b330c8f/src/LogPatch.js#L102-L120). This should allow proper log line delimiters for logging as well, and avoid the replacement of newlines by carriage returns in some RIC patches.

The test added mostly assures normal carriage returns are not split into multiple messages - the main cause for the failing ext test.
In the future, with the improvements above, we need to add more tests for parity for different logging mechanisms.

## Changes
* Reduced load on LocalStack for every invocation, due to less requests.
* Log events now have (correctly) trailing newlines, instead of stripping them out.

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/logs.py
+++ b/localstack-core/localstack/services/lambda_/invocation/logs.py
@@ -1,13 +1,13 @@
 import dataclasses
 import logging
 import threading
+import time
 from queue import Queue
 from typing import Optional, Union
 
 from localstack.aws.connect import connect_to
 from localstack.utils.aws.client_types import ServicePrincipal
 from localstack.utils.bootstrap import is_api_enabled
-from localstack.utils.cloudwatch.cloudwatch_util import store_cloudwatch_logs
 from localstack.utils.threads import FuncThread
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


dfangl:
Set the semver label to minor, as the changed log format (keeping newline endings) can be quite visible to the user, even though it is technically a parity improvement.

dfangl:
Ext run: https://github.com/localstack/localstack-ext/actions/runs/14221427642

localstack-core/localstack/services/lambda_/invocation/logs.py
```diff
@@ -50,10 +50,34 @@ def run_log_loop(self, *args, **kwargs) -> None:
             log_item = self.log_queue.get()
             if log_item is QUEUE_SHUTDOWN:
                 return
+            # we need to split by newline - but keep the newlines in the strings
+            # strips empty lines, as they are not accepted by cloudwatch
+            logs = [line + "\n" for line in log_item.logs.split("\n") if line]
```
joe4dev, line 55: Do we have a test for the empty lines case?

joe4dev, line 55: Following [this discussion](https://github.com/localstack/localstack/pull/12470/files#diff-2a8c6622efa5103d5807b4619b06e1f9492dc879f3474eee5fdfdf1e96f10a16R505), do we have an intuition where these replacements came from?

```
log_output = log_output.replace("\\x1b", "\n\\x1b")
log_output = log_output.replace("\x1b", "\n\x1b")
```

I just wanted to add the reference in case we need to add them reactively because we removed them without knowing where they came from 😄

joe4dev (approved):
LGTM 🚀 

Thanks for re-introducing this optimization and pointing out the next steps. I created a backlog item to capture these improvement suggestions.

Ext run is 🟢

localstack-core/localstack/services/lambda_/invocation/logs.py
dfangl: No, still no idea. Since those are ANSI escape sequences, I still think they stem from the time were we captured the lambda logs directly from docker running inside a full terminal instance.

dfangl: No, but cloud watch does not accept empty lines. The logic matches the one in `store_cloudwatch_logs` though, so we at least do not get worse in terms of parity. We should write tests for this though.

PR #12471, opened 2025-04-02
Add new volume dir mount option, rename VolumeBind to BindMount

dfangl:
## Motivation
This PR refactors the usage of `get_host_path_for_path_in_docker`. 
This utility is very specific, assuming:

1. The volume dir is mounted as bind mount
2. We are using docker (or something API compatible, like podman) for spawning containers

To overcome these restrictions in the long term, I added a new type of mount for our container clients, the `VolumeDirMount`.

This mount should convey the intent to make the contents of a subpath of `/var/lib/localstack` (as path inside the LS container) available to the target container - however this happens.

This shifts the responsibility for this from the services to the docker client - which can then decide how it is achieved.
Currently this will continue to be `get_host_path_for_path_in_docker`, but in the future support for named volumes and kubernetes is thinkable as well.

From here on out, we should not use `get_host_path_for_path_in_docker` outside tests, and very specific, to be considered, usecases.

Also, `VolumeBind` is renamed to `BindMount`, which is a clearer name in my opinion. I did a usage search for this class within or organization, and could find any usages outside of localstack and our ext repository.

## Changes
* No user facing changes
* Introduce new VolumeDirMount to mount subpaths of /var/lib/localstack into other containers
* Rename VolumeBind to BindMount

### DIFF
--- a/localstack-core/localstack/utils/container_utils/container_client.py
+++ b/localstack-core/localstack/utils/container_utils/container_client.py
@@ -27,6 +27,7 @@ from typing import (
 import dotenv
 
 from localstack import config
+from localstack.constants import DEFAULT_VOLUME_DIR
 from localstack.utils.collections import HashableList, ensure_list
 from localstack.utils.files import TMP_FILES, chmod_r, rm_rf, save_file
 from localstack.utils.no_exit_argument_parser import NoExitArgumentParser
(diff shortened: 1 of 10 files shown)


localstack-core/localstack/utils/container_utils/container_client.py
```diff
@@ -395,8 +396,14 @@ def to_str(self) -> str:
 
         return ":".join(args)
 
+    def to_dict(self):
```
simonrw, line 399: nit: can we add types to this? I feel the name is a bit strange since it returns a `tuple[str, dict[str, str]]` but the name is `to_dict`

simonrw (approved):
This is a great first step towards a more flexible API for volumes for containers in general. Thanks!

PR #12476, opened 2025-04-03
Admin: Use cryptography instead of OpenSSL

bblommers:
## Motivation
The [PyOpenSSL](https://pypi.org/project/pyOpenSSL/) module explicitly suggests that users should use the standard `cryptography` module instead.

As far as I can tell, all SSL-related functions within LocalStack (Pro) now use `cryptography`, so this both simplifies the implementation, and reduces the mental load for developers.

I was hoping to be able to remove the dependency as well, just to make LocalStack's footprint a little smaller, but unfortunately `twisted` still uses it. :cry:

### DIFF
--- a/localstack-core/localstack/utils/crypto.py
+++ b/localstack-core/localstack/utils/crypto.py
@@ -1,11 +1,16 @@
 import io
+import ipaddress
 import logging
 import os
 import re
 import threading
+from datetime import datetime, timedelta, timezone
 from typing import Tuple
 
+from cryptography import x509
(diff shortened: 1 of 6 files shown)


bblommers:
This is dependent on https://github.com/getmoto/moto/pull/8761 being part of Moto-ext - will revisit this once that's merged.

bblommers:
The outstanding issue here is that Moto doesn't play nice with this new way of creating certificates.

Closing because it doesn't really solve a customer problem, and I don't have the time to look into why Moto is acting up.

Issue #12424, opened 2025-03-22
bug: LoggingConfig.LogFormat of Lambda function not applied via CloudFormation or SAM


kakakakakku:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Currently, AWS Lambda supports `LoggingConfig.LogFormat` setting to choose the log format from `Text` or `JSON`.

- https://docs.aws.amazon.com/lambda/latest/api/API_LoggingConfig.html
- https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/create-function.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html
- https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html

Although LocalStack supports `LoggingConfig.LogFormat`, it is not supported when deploying with AWS CloudFormation or AWS SAM.

- AWS CLI: OK🙆‍♂️
- AWS CloudFormation: NG🙅
- AWS SAM: NG🙅

### Expected Behavior

`LoggingConfig.LogFormat` should also be supported in AWS CloudFormation and AWS SAM.

### How are you starting LocalStack?

With the `localstack` script

## AWS CLI

Check `LoggingConfig.LogFormat` attributes👀
`LogFormat` set `JSON`.

```sh
$ awslocal lambda create-function \
  --function-name function-cli \
  --runtime python3.12 \
  --role arn:aws:iam::000000000000:role/function-role \
  --handler app.lambda_handler \
  --zip-file fileb://app.zip \
  --logging-config LogFormat=JSON

$ awslocal lambda get-function-configuration --function-name function-cli
{
    "FunctionName": "function-cli",
    "FunctionArn": "arn:aws:lambda:ap-northeast-1:000000000000:function:function-cli",
    "Runtime": "python3.12",
    "Role": "arn:aws:iam::000000000000:role/function-role",
    "Handler": "app.lambda_handler",
    "CodeSize": 219,
    "Description": "",
    "Timeout": 3,
    "MemorySize": 128,
    "LastModified": "2025-03-22T03:27:01.941729+0000",
    "CodeSha256": "OLn3OOQr7kkPDqZFzMn0oYcoe8QMIB0pG1HTCzwwMe8=",
    "Version": "$LATEST",
    "TracingConfig": {
        "Mode": "PassThrough"
    },
[...]

PR #12480, opened 2025-04-04
CFn: Add Lambda Function LoggingConfig

anisaoshafi:
## Motivation
Customer reported in https://github.com/localstack/localstack/issues/12424 that `LoggingConfig` is not set properly via Cloudformation.

## Changes
- Added `LoggingConfig` param in `LambdaFunctionProvider` _kwargs_ used to  create and update lambda in: `lambda_/resource_providers/aws_lambda_function.schema.json`
- Updated schema specification (& manually tackled custom logic) via [scaffolding](https://www.notion.so/localstack/Writing-a-Resource-Provider-d83bbd2d17ec4bff90d193f232a4e945): `python -m localstack.services.cloudformation.scaffolding generate --resource-type "AWS::Lambda::Function" ...`
- Added aws validated test `test_lambda_logging_config`

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
+++ b/tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
@@ -1608,7 +1608,7 @@
     }
   },
   "tests/aws/services/cloudformation/resources/test_lambda.py::test_lambda_logging_config": {
-    "recorded-date": "03-04-2025, 17:34:50",
+    "recorded-date": "08-04-2025, 12:10:56",
     "recorded-content": {
       "stack_resource_descriptions": {
         "StackResources": [
@@ -1616,8 +1616,8 @@
             "DriftInformation": {
               "StackResourceDriftStatus": "NOT_CHECKED"
             },
-            "LogicalResourceId": "LambdaFunction",
-            "PhysicalResourceId": "<function-name>",
+            "LogicalResourceId": "logical-resource-id",
+            "PhysicalResourceId": "physical-resource-id",
             "ResourceStatus": "CREATE_COMPLETE",
             "ResourceType": "AWS::Lambda::Function",
             "StackId": "arn:<partition>:cloudformation:<region>:111111111111:stack/<stack-name:1>/<resource:1>",
@@ -1628,8 +1628,8 @@
             "DriftInformation": {
               "StackResourceDriftStatus": "NOT_CHECKED"
             },
-            "LogicalResourceId": "MyFnServiceRole",
-            "PhysicalResourceId": "<stack-name:1>-MyFnServiceRole-IVwvF4i3Fcqy",
+            "LogicalResourceId": "logical-resource-id",
+            "PhysicalResourceId": "physical-resource-id",
             "ResourceStatus": "CREATE_COMPLETE",
             "ResourceType": "AWS::IAM::Role",
             "StackId": "arn:<partition>:cloudformation:<region>:111111111111:stack/<stack-name:1>/<resource:1>",
@@ -1640,8 +1640,8 @@
             "DriftInformation": {
               "StackResourceDriftStatus": "NOT_CHECKED"
             },
-            "LogicalResourceId": "Version",
-            "PhysicalResourceId": "arn:<partition>:lambda:<region>:111111111111:function:<resource:2>",
+            "LogicalResourceId": "logical-resource-id",
+            "PhysicalResourceId": "physical-resource-id",
             "ResourceStatus": "CREATE_COMPLETE",
             "ResourceType": "AWS::Lambda::Version",
             "StackId": "arn:<partition>:cloudformation:<region>:111111111111:stack/<stack-name:1>/<resource:1>",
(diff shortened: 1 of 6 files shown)


tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
```diff
@@ -1606,5 +1606,60 @@
...
+            "StackId": "arn:<partition>:cloudformation:<region>:111111111111:stack/<stack-name:1>/<resource:1>",
+            "StackName": "<stack-name:1>",
+            "Timestamp": "timestamp"
+          },
+          {
+            "DriftInformation": {
+              "StackResourceDriftStatus": "NOT_CHECKED"
+            },
+            "LogicalResourceId": "MyFnServiceRole",
+            "PhysicalResourceId": "<stack-name:1>-MyFnServiceRole-IVwvF4i3Fcqy",
```
joe4dev, line 1632: `IVwvF4i3Fcqy` looks like a random uuid. Does this test pass when validated against AWS without updating the snapshot? A transformer for the Lambda role could help here and potentially resolve the TODO in `skip_snapshot_verify`.

joe4dev: I see that's a common issue in Lambda CloudFormation tests; not nice but feel free to skip :)

tests/aws/templates/cfn_lambda_logging_config.yaml
```diff
@@ -0,0 +1,53 @@
...
+              statusCode: 200,
+              body: "Hello, World!" 
+            };
+          };
+      Role:
+        Fn::GetAtt:
+          - MyFnServiceRole
+          - Arn
+      Handler: index.handler
+      Runtime: nodejs18.x
```
joe4dev, line 42: nit: unifying the runtimes (e.g., Python 3.12) could potentially speed up test execution if we don't need to pull many different runtimes. I wouldn't recommend `nodejs18.x`, which will get deprecated later this year (~sept).

tests/aws/templates/cfn_lambda_logging_config.yaml
```diff
@@ -0,0 +1,53 @@
...
+      LoggingConfig:
```
joe4dev, line 43: Maintainability: Do we need a dedicated tests for each sub-feature (e.g., LoggingConfig) or is there an existing test case we can extend?

Imagine we have a minimal case for testing defaults and a "reference" case that attempts to cover as much as possible of the API surface.

joe4dev (approved):
The fix LGTM.

My main question concerns maintainability on whether we need a test case for each sub-feature. Also happy to hear the thoughts of the CloudFormation team on this one :)

tests/aws/templates/cfn_lambda_logging_config.yaml
anisaoshafi: TBH, I thought about this and what is the preference/convention in general at LS. My initial approach was to adapt an existing test, then decided to create a dedicated one (because I wasn't sure which test was the right one to add this new condition too, and if I did find it, what if someone removed that test for some reason - though this case still happen). 
Good point. It's probably better to adopt an existing one, since having so many tests is great, but each of them will take its time in the pipeline, so better to reuse what we have 💯

simonrw (approved):
For the CFn changes specifically, thank you for re-running the scaffolding. I'm sorry if the process was painful for you. We don't currently separate the autogenerated files from the main code so I assume you had to merge the new and old versions.

It's nice that this is at least possible, and the implementation seems straight forward, thank you!

tests/aws/services/cloudformation/resources/test_lambda.snapshot.json
anisaoshafi: I didn't like it either, and now I understand a bit better why that happened. Now it's fixed by setting `reference_replacement=False` to that property. Thanks for pointing it out!

tests/aws/templates/cfn_lambda_logging_config.yaml
anisaoshafi: Good catch, updated that 💯

anisaoshafi:
> For the CFn changes specifically, thank you for re-running the scaffolding. I'm sorry if the process was painful for you. We don't currently separate the autogenerated files from the main code so I assume you had to merge the new and old versions.
> 
> It's nice that this is at least possible, and the implementation seems straight forward, thank you!

We don't currently separate the autogenerated files from the main code so I assume you had to merge the new and old versions.

@simonrw Thanks for the review and the documentation in [Writing a Resource Provider](https://www.notion.so/localstack/Writing-a-Resource-Provider-d83bbd2d17ec4bff90d193f232a4e945), that was clear to follow 🎉 
There were few manual changes so I  had to merge the new and old versions as you correctly assumed. Great to hear we're aware of this limitation for the future 🚀

tests/aws/templates/cfn_lambda_logging_config.yaml
joe4dev: good choice 👏

PR #12481, opened 2025-04-04
feat: propagate x-ray trace id to event bridge targets

maxhoheiser:
## Motivation
Currently, when using X-Ray tracing with EventBridge, the trace context is not properly propagated to the target services. This makes it difficult to trace end-to-end requests across multiple AWS services in LocalStack. By propagating the X-Ray trace ID from the original request to EventBridge targets, we can provide a complete trace of the request flow, which is essential for debugging and monitoring distributed applications.
We can make use of the already correctly insturmented internal botocore client to propagate existing X-Ray trace ID headers.
For now only Lambda and API Gateway support X-Ray trace ID propagation.

[X-Ray trace header concept](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader)

## Changes
- Set new X-Ray trace ID or new parent if present by creating a new Span
- Modified the EventBridge target processing to include X-Ray trace ID when invoking target services with not already instrumented botocore client
- Ensured compatibility with the existing trace header encoding mechanism

## Testing
- `test_xray_trace_propagation_events_lambda`
- `test_xray_trace_propagation_events_api_gateway`
- `test_xray_trace_propagation_events_events`

we need to skip a schedule rate test `tests_schedule_rate_custom_input_target_sqs` since the check of strict execution time of 60 seconds seams to be flaky


localstack-core/localstack/services/events/provider.py
```diff
@@ -1809,6 +1810,10 @@ def _process_entry(
             return
 
         region, account_id = extract_region_and_account_id(event_bus_name_or_arn, context)
+
+        trace_id = populate_trace_id(context.request.headers)
```
bentsku, line 1814: You can replace that with the context trace, see the main review comment.

```suggestion
        trace_id = context.trace_context["aws_trace_header"]
```

localstack-core/localstack/services/events/provider.py
```diff
@@ -1809,6 +1810,10 @@ def _process_entry(
...
+
+        # TODO check interference with x-ray trace header
```
bentsku, line 1816: question: how could those clash? could you quickly explain to me, I'm not familiar with this 😅

bentsku (changes requested):
This is a great idea, this is going to help a lot!

I'm really sorry the API Gateway implementation was available "externally", this was implemented before @joe4dev implemented a really great way to access this data directly in the `RequestContext` with this PR: #11708.

For context, API Gateway logic happens outside of a regular service provider, which is why it needs custom logic, as the edge routes don't have access to the `RequestContext`.

But `events` can access the `RequestContext.trace_context` which already contains everything done here, and Lambda is a good example on how to use it.
You can access the X-Ray header via `context.trace_context["aws_trace_header"]`: I believe it should always be generated. 

Additionally, do you think it's possible to add a test for this, maybe by re-using existing API Gateway tests? This feels a bit tricky as there are many steps along the way, events to API Gateway (to maybe lambda..., but not mandatory). But if we were to send an X-Ray header to the `events.put_events` call, some parts of it should be the same from the APIGW invocation for example.
By re-using the test `test_xray_trace_propagation` but adapt it to `events`, and using some parts of `TestApiGatewayCommon.test_invocation_trace_id` with a MOCK integration... we could test that we always receive a trace id, generated if not provided, and parts of it similar if provided? 

Once this is addressed, I believe this is good to go 🚀

localstack-core/localstack/services/events/provider.py
maxhoheiser: I am piggybacking on the trace header for intermediary storing region and account for cross-region cross-account events

localstack-core/localstack/services/events/target.py
```diff
@@ -566,17 +579,29 @@ def _validate_input(self, target: Target):
...
 class LambdaTargetSender(TargetSender):
     def send_event(self, event):
+        # instrument boto client to add x-ray trace header
+        trace_header_str = get_trace_header_str_from_segment()
+
+        def add_xray_header(request, **kwargs):
+            request.headers["X-Amzn-Trace-Id"] = trace_header_str
+
+        event_name = "before-send.lambda.*"
+        self.client.meta.events.register(event_name, add_xray_header)
```
bentsku, line 589: do you know what's the impact of registering and deregistering? I believe maybe @dfangl or @gregfurman had some issues with this at some point, doesn't it have quite a high performance cost?

Also, if the `invoke()` call fails, we won't deregister here, this should be wrapped in a `try...finally` probably? 

I think generally, we might want to have proper instrumentation similar to the IAM instrumentation of the client, to have a way to add the X-Ray header to all `connect_to` client? We have the same issue in API Gateway to forward the X-Ray header to the `Invoke` call of the `AWS_PROXY` implementation. But this is out of scope, it's just that we're seeing this happen more and more often now 🤔 

Until we have a more general approach, maybe we could take inspiration from @gregfurman implementation for the internal SQS header:
https://github.com/localstack/localstack/blob/97bc2c77eacdfb2f2ac8e8a2db5b512a526d4054/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/sqs_poller.py#L73

Not sure if that's the right way for our use-case here. Open to all suggestions! if @joe4dev also has ideas?

localstack-core/localstack/services/events/utils.py
```diff
@@ -293,3 +295,29 @@ def is_nested_in_string(template: str, match: re.Match[str]) -> bool:
...
+    segment.parent_id = trace_header.parent
+    if trace_header.sampled == "1":
+        segment.sampled = True
+    elif trace_header.sampled == "0":
+        segment.sampled = False
+
+
+def get_trace_header_str_from_segment():
+    # Get the current segment
+    segment = xray_recorder.current_segment()
```
bentsku, line 312: question: what if we import the `xray_recorder` in the code somewhere else and start using it? wouldn't this start failing weirdly? as `xray_recorder` is a global variable?

tests/aws/services/events/test_x_ray_trace_propagation.py
```diff
@@ -0,0 +1,284 @@
...
+
+from localstack.testing.aws.util import is_aws_cloud
+from tests.aws.services.events.helper_functions import is_old_provider
+from tests.aws.services.events.test_events import TEST_EVENT_DETAIL, TEST_EVENT_PATTERN
+from tests.aws.services.lambda_.test_lambda import TEST_LAMBDA_XRAY_TRACEID
+
+# currently only API Gateway v2 and Lambda support X-Ray tracing
+
+
+@markers.aws.unknown
```
bentsku, line 32:
```suggestion
@markers.aws.validated
```

tests/aws/services/events/test_x_ray_trace_propagation.py
bentsku, line 1: Nice for the tests! Happy to see them validated now, thanks for doing it!

bentsku (changes requested):
Nice, this is looking nice and it's great to see the tests are green! 🚀 

I have some concerns about using the `xray_recorder` global variable here. Would this start falling if something else was using it? Is it the right solution and what issue is it solving?
Should the `TraceHeader` be extended instead to allow generating "child"/"segment" from the existing `TraceHeader` in the context? 

Also have some concerns about registering/deregistering for every call of the botocore client, I'm not sure what can be done to not affect all the clients, but also be maybe more lightweight if needed, depending on the performance cost. 

Sorry to delay this implementation, just want to make sure we're not introducing memory leaks/performance issues with this.

Thanks!

bentsku (commented):
LGTM! Thanks for improving the client logic. 

I think we will need to find a global solution for the propagation of traces between services, as we're dealing it case per case for now.

This update also required a lot of function signature changes. Either using thread local context / context var, or passing down a "context" object holding the data you need in the interim, so you just need to update some kind of dictionary/dataclass and wouldn't need to update the function signature anymore. But still not optimal 😄 

But as we need this right now, I think it's an acceptable way of going until we find a better way. Maybe we could have made it a bit more flexible by using `**kwargs`, but I think this should work!

We just need to update the signature of the upstream implementation of `ECSTargetSender` and we're good to go! thanks for pushing through this one! 🚀

bentsku (approved):
LGTM! Just waiting to see upstream tests being green, had some issues with the run but seems to be going now 🚀

PR #12482, opened 2025-04-04
KMS: add ability to decrypt data with all rotated keys

sannya-singal:
## Motivation
Add ability to decrypt data that was encrypted before the rotation event by preserving the history of the key material on `RotateKeyOnDemand`. 

Closes: https://github.com/localstack/localstack/issues/10723

## Changes
Store all the previous crypto keys with a maximum of 10 times per KMS key (checkout the [AWS Documentation](https://docs.aws.amazon.com/kms/latest/APIReference/API_RotateKeyOnDemand.html)) to ensure the decryption of the data that was encrypted before the rotation of the key.

Validate the implementation using AWS validated snapshot tests.

### DIFF
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -705,7 +705,7 @@ class KmsKey:
             return request_key_usage or "ENCRYPT_DECRYPT"
 
     def rotate_key_on_demand(self):
-        if len(self.previous_keys) == ON_DEMAND_ROTATION_LIMIT:
+        if len(self.previous_keys) >= ON_DEMAND_ROTATION_LIMIT:
             raise LimitExceededException(
                 f"The on-demand rotations limit has been reached for the given keyId. "
                 f"No more on-demand rotations can be performed for this key: {self.metadata['Arn']}"
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -12,7 +12,7 @@ from collections import namedtuple
 from dataclasses import dataclass
 from typing import Dict, Optional, Tuple
 
-from cryptography.exceptions import InvalidSignature, UnsupportedAlgorithm
+from cryptography.exceptions import InvalidSignature, InvalidTag, UnsupportedAlgorithm
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, hmac
 from cryptography.hazmat.primitives import serialization as crypto_serialization
@@ -29,6 +29,7 @@ from localstack.aws.api.kms import (
     CreateGrantRequest,
     CreateKeyRequest,
     EncryptionContextType,
+    InvalidCiphertextException,
     InvalidKeyUsageException,
     KeyMetadata,
     KeySpec,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/kms/models.py
```diff
@@ -694,6 +705,12 @@ def _get_key_usage(self, request_key_usage: str, key_spec: str) -> str:
             return request_key_usage or "ENCRYPT_DECRYPT"
 
     def rotate_key_on_demand(self):
+        if len(self.previous_keys) >= ON_DEMAND_ROTATION_LIMIT:
```
k-a-il, line 708: Why is the condition "greater than or equal to" instead of just "equal to"? In this case, the invalid scenario only happens when the length is exactly 10.

k-a-il:
Nice work! It's great to see the GitHub issues being handled :) Could you please explain in the PR description or comments how you determine which key material was used to encrypt the value?

localstack-core/localstack/services/kms/models.py
sannya-singal: Using `>= 10` is defensive — it prevents issues not just when the length is exactly 10, but also if, due to bugs or unexpected behaviour, it ever exceeds 10. It's a safeguard for future changes or misuse.

localstack-core/localstack/services/kms/models.py
```diff
@@ -319,9 +324,15 @@ def decrypt(
...
-        return decrypt(
-            self.crypto_key.key_material, ciphertext.ciphertext, ciphertext.iv, ciphertext.tag, aad
-        )
+        keys_to_try = [self.crypto_key.key_material] + self.previous_keys
+
+        for key in keys_to_try:
+            try:
+                return decrypt(key, ciphertext.ciphertext, ciphertext.iv, ciphertext.tag, aad)
+            except (InvalidTag, InvalidSignature):
+                continue
```
sannya-singal, line 333: Here the data can be decrypted only with the key which encrypted it. 
During decryption, we first attempt with the current key material. If decryption fails, we try previous key versions until we find the one that correctly matches the ciphertext.

k-a-il: >During decryption, we first attempt with the current key material. If decryption fails, we try previous key versions until we find the one that correctly matches the ciphertext.

Ok, but why does decryption actually fail here? When does`InvalidSignature` get thrown, and how does it know the key material’s wrong—what field is it checking?

k-a-il: What is the difference between `InvalidTag` and `InvalidSignature` being thrown in this case?

sannya-singal: When data is encrypted using a specific key, it can only be decrypted using that exact key. Attempting to decrypt the data with a different key will result in an error—commonly `InvalidSignature` or `InvalidTag`, depending on the algorithm being used.

To handle key rotation gracefully, especially when accessing older data, the decryption process must attempt to use not just the current key but also all previous key materials. This ensures backward compatibility and allows successful decryption of data encrypted before the most recent key rotation.

k-a-il (approved):

localstack-core/localstack/services/kms/models.py
sannya-singal: Relevant doc: https://cryptography.io/en/latest/hazmat/primitives/aead/

PR #12483, opened 2025-04-04
[ESM] Re-initialize shards when NextShardIterator value is empty

gregfurman:
## Motivation

We do not currently cater for the case of a next shard iterator value not being present in a `GetRecords` response. According to the [AWS doc](https://docs.aws.amazon.com/streams/latest/dev/troubleshooting-consumers.html#getrecords-returns-empty):

> In a production scenario, the only time the continuous loop should be exited is when the NextShardIterator value is `NULL`. When NextShardIterator is `NULL`, it means that the current shard has been closed and the ShardIteratorvalue would otherwise point past the last record.

In addition, DynamoDB Local can occasionally not return a `NextShardIterator` in its response. This seems to occur when trying to read from a closed shard or when using an expired iterator. Hence, an exception is raised since the StreamPoller always expects this field in a `GetRecords` response -- raising the below `KeyError` exception:

```
2025-03-20T16:02:18.318 ERROR --- [functhread17] l.s.l.e.esm_worker         : Error while polling messages for event source arn:aws:dynamodb:us-east-1:000000000000:table/test_table/stream/2025-03-19T22:20:57.833: 'NextShardIterator'
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/lambda_/event_source_mapping/esm_worker.py", line 155, in poller_loop
    self.poller.poll_events()
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py", line 155, in poll_events
    self.poll_events_from_shard(*current_shard_tuple)
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py", line 166, in poll_events_from_shard
    self.shards[shard_id] = get_records_response["NextShardIterator"]
```

## Changes
- Re-initialize shards when a `NextShardIterator` is not returned or is `None` in a `GetRecords` response.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -154,7 +154,6 @@ class StreamPoller(Poller):
             LOG.debug("No shards found for %s.", self.source_arn)
             raise EmptyPollResultsException(service=self.event_source(), source_arn=self.source_arn)
         else:
-            LOG.debug("Event source %s has %d shards.", self.source_arn, len(self.shards))
             # Remove all shard batchers without corresponding shards
             for shard_id in self.shard_batcher.keys() - self.shards.keys():
                 self.shard_batcher.pop(shard_id, None)
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -185,7 +185,10 @@ class StreamPoller(Poller):
     def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
         get_records_response = self.get_records(shard_iterator)
         records: list[dict] = get_records_response.get("Records", [])
-        next_shard_iterator = get_records_response["NextShardIterator"]
+        if not (next_shard_iterator := get_records_response.get("NextShardIterator")):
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -185,7 +185,10 @@ def poll_events(self):
     def poll_events_from_shard(self, shard_id: str, shard_iterator: str):
         get_records_response = self.get_records(shard_iterator)
         records: list[dict] = get_records_response.get("Records", [])
-        next_shard_iterator = get_records_response["NextShardIterator"]
+        if not (next_shard_iterator := get_records_response.get("NextShardIterator")):
+            # If the next shard iterator is None, we can assume the shard is closed or
```
joe4dev, line 189: Can we add that great docs link here: https://docs.aws.amazon.com/streams/latest/dev/troubleshooting-consumers.html#getrecords-returns-empty

joe4dev (approved):
Good catch and excellent link 👍 

General limitation (unrelated to this change):
Do we clarify in the code and docs that resharding might lead to undefined/untested behavior?
I guess we could lose data with `LATEST` and potentially re-process records with `AT_TIMESTAMP` and `TRIM_RECORDS` here 😬

PR #12493, opened 2025-04-07
add base config to mount MockConfigFile and testing support

MEPalma:
## Motivation
The current SFN v2 interpreter lacks support for testing state machines using mocked service responses; a feature initially introduced by AWS Step Functions Local: https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-test-sm-exec.html. To align with LocalStack’s goal of serving as a drop-in replacement for Step Functions Local, this update introduces support for mock-based state machine execution via a configuration file (MockConfigFile.json).

Specifically, this change allows users to specify the mock configuration through the `SFN_MOCK_CONFIG` environment variable, mirroring the behavior of SFN Local. The only distinction is when using LocalStack inside a Docker container, where the environment variable must be prefixed with `LOCALSTACK_`.

Additionally, due to the unique requirements of this mocking approach, the changes include initial support for handling temporary paths to `MockConfigFile.json` within test scenarios.

## Changes
- Added new SFN_MOCK_CONFIG configuration variable for mounting MockConfigFile in the LocalStack runtime
- Added initial testing support for testing mocked state machines

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -1094,6 +1094,9 @@ LOCAL_PORT_STEPFUNCTIONS = int(os.environ.get("LOCAL_PORT_STEPFUNCTIONS") or 808
 # Stepfunctions lambda endpoint override
 STEPFUNCTIONS_LAMBDA_ENDPOINT = os.environ.get("STEPFUNCTIONS_LAMBDA_ENDPOINT", "").strip()
 
+# Specifies the path to the mock configuration file for Step Functions, commonly named MockConfigFile.json.
+SFN_MOCK_CONFIG = os.environ.get("SFN_MOCK_CONFIG")
+
 # path prefix for windows volume mounting
 WINDOWS_DOCKER_MOUNT_PREFIX = os.environ.get("WINDOWS_DOCKER_MOUNT_PREFIX", "/host_mnt")
 
--- a/localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
+++ b/localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
@@ -1,0 +1,43 @@
+import json
+import logging
+import os
+from typing import Optional, Tuple
+
+from localstack import config
+
+LOG = logging.getLogger(__name__)
+
+_SFN_MOCK_CONFIG_CACHE: Optional[Tuple[int, dict]] = None
+
+
+def load_sfn_mock_config_file() -> Optional[dict]:
+    configuration_file_path = config.SFN_MOCK_CONFIG
+    if not configuration_file_path:
+        return None
+
+    global _SFN_MOCK_CONFIG_CACHE
+
+    try:
+        modified_time = os.path.getmtime(configuration_file_path)
+        modified_epoch = int(modified_time)
+    except Exception as ex:
+        LOG.warning(
+            "Unable to access the step functions mock configuration file at '%s' due to %s",
+            configuration_file_path,
+            ex,
+        )
+        return None
+
+    if _SFN_MOCK_CONFIG_CACHE is None or _SFN_MOCK_CONFIG_CACHE[0] < modified_epoch:
+        with open(configuration_file_path, "r") as df:
+            try:
+                mock_config = json.load(df)
+            except Exception as ex:
+                LOG.warning(
+                    "Unable to load step functions mock configuration file at '%s' due to %s",
+                    configuration_file_path,
+                    ex,
+                )
+                return None
+        _SFN_MOCK_CONFIG_CACHE = (modified_epoch, mock_config)
+    return _SFN_MOCK_CONFIG_CACHE[1]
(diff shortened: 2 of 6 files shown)


localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -0,0 +1,43 @@
...
+
+_SFN_MOCK_CONFIG_CACHE: Optional[Tuple[int, dict]] = None
+
+
+def load_sfn_mock_config_file() -> Optional[dict]:
+    configuration_file_path = config.SFN_MOCK_CONFIG
+    if not configuration_file_path:
+        return None
+
+    global _SFN_MOCK_CONFIG_CACHE
```
gregfurman, line 18: **suggestion**: Can you use `functools.lru_cache` instead of a global variable? i.e

```python
@lru_cache(maxsize=1)
def _retrieve_sfn_mock_config(configuration_file_path: str, modified_time: Optional[int]) -> Tuple[Optional[int], Optional[Dict]]:
    if not configuration_file_path:
        return None, None
    
        with open(configuration_file_path, "r") as df:
            try:
                mock_config = json.load(df)
            except Exception as ex:
                LOG.warning(
                    "Unable to load step functions mock configuration file at '%s' due to %s",
                    configuration_file_path,
                    ex,
                )
                return None
    
    return modified_time, mock_config
```

Since we cache on the modified epoch as a key, if it changes then the function will execute and re-load in the file.

tests/aws/services/stepfunctions/v2/mocking/test_mock_config_file.py
```diff
@@ -0,0 +1,28 @@
+from localstack import config
+from localstack.services.stepfunctions.mocking.mock_config import load_sfn_mock_config_file
+from localstack.testing.pytest import markers
+from tests.aws.services.stepfunctions.mocked_responses.mocked_response_loader import (
+    MockedResponseLoader,
+)
+
+
+class TestMockConfigFile:
```
gregfurman, line 9: Probably not relevant to this PR but I wonder if it'd be worth snapshotting against [stepfunctions local](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html) and making sure our behaviour is the same. Unless, that is, the feature we're adding is different to what is expected?

gregfurman (approved):
LGTM! Left some questions. Nothing blocking. Feel free to merge when ready.

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
MEPalma: Thanks! Yes, although modified_time needs to be passed and isn’t used directly within the retrieval function, this approach may simplify things overall. I’ll switch to using lru_cache and add documentation to clarify the role of modified_time as a cache key.

tests/aws/services/stepfunctions/v2/mocking/test_mock_config_file.py
MEPalma: Yes, this is definitely an options that is being evaluated

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -0,0 +1,43 @@
...
+        modified_epoch = int(modified_time)
+    except Exception as ex:
+        LOG.warning(
+            "Unable to access the step functions mock configuration file at '%s' due to %s",
+            configuration_file_path,
+            ex,
+        )
+        return None
+
+    if _SFN_MOCK_CONFIG_CACHE is None or _SFN_MOCK_CONFIG_CACHE[0] < modified_epoch:
```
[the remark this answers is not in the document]
anisaoshafi: True 🤦🏼‍♀️ thanks a lot for clarifying that bit. Now I understand why that varibale was GLOBAL.

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -0,0 +1,69 @@
...
+            The last modified time of the file, in epoch seconds. This value is used
+            as part of the cache key to ensure the cache is refreshed when the file is updated.
+
+    Returns:
+        Optional[dict]:
+            The parsed configuration as a dictionary if the file is successfully loaded,
+            or `None` if an error occurs during reading or parsing.
+
+    Notes:
+        - The `modified_epoch` argument is not used inside the function logic, but is
```
anisaoshafi, line 36: Thanks for explaining this 💡

anisaoshafi (approved):
LGTM 🚀  
Thanks for adding the comments in __retrieve_sfn_mock_config_, much clearer. 

TIL about **lru_cache**!

localstack-core/localstack/config.py
```diff
@@ -1359,6 +1362,7 @@ def use_custom_dns():
     "SQS_ENDPOINT_STRATEGY",
     "SQS_DISABLE_CLOUDWATCH_METRICS",
     "SQS_CLOUDWATCH_METRICS_REPORT_INTERVAL",
+    "STEPFUNCTIONS_LAMBDA_ENDPOINT",
```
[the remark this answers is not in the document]
MEPalma: I'm removing it, it slipped through during conflict resolution!

PR #12494, opened 2025-04-07
EC2: generate security group ids using id manager concept

simonrw:
## Motivation

The PR fixing tag behaviour in LocalStack via https://github.com/localstack/localstack/pull/12459 broke the [LocalStack replicator](https://docs.localstack.cloud/user-guide/tools/aws-replicator/) functionality as the tags that were replicated no longer matched.

## Changes

- Use the moto id generator functionality to generate a security group id from either the ID cache (moto/localstack pro concept) or custom tags

## Testing

- expanded on existing tests to included creating resources with the id manager and ensure it will not break if vpc id are provided


viren-nadkarni (approved):
Looks good, thanks for fixing this :+1:

PR #12500, opened 2025-04-08
Add profile as an option to every click command

jw2:
## Motivation
We've had feedback that mandating the `--profile` option be specified at the top level, before any actions, is overly restrictive and causes headaches when wrapping the CLI in custom developer tools.  The following currently works:
```bash
localstack --profile=custom start
```
but the following:
```bash
localstack start --profile=custom
```
fails with:
> Usage: localstack start [OPTIONS]
> Try 'localstack start -h' for help.
> 
> Error: No such option: --profile

## Changes
`--profile` can now be specified at any point.  This is achieved by removing any occurrences of  `--profile` options from the command line arguments (`sys.argv`) before click gets invoked.

The profile option was already being read and used before click was invoked, and click performed no further processing of it.  The only functionality that click was actually performing w.r.t. the profile option was:
1. ensuring it was listed in the help
2. ensuring that it can only be provided as a top-level option (which we don't want).

`--profile` no longer makes it through to click, so 2. never happens.

There is no change to the processing of `-p` because that option is used in at least one subcommand.  Therefore it would break those options if we were to strip it out like `--profile`.  Ideally, we would remove the ability to specify a profile using `-p` as there is a pre-existing issue whereby if a port mapping command was specified, e.g. `localstart start -p 80:8080`, then localstack would start using a profile called `80:8080`.  However, this would break back-compatibility.

There is another top-level switch `--debug` and this remains top-level only.  The reasoning behind this decision is:
* There is no known requirement to allow this to be specified anywhere.
* The fix would be more difficult - either it would need to be added to every possible command, or further effort to extract the processing outside of click like `--profile`.

### DIFF
--- a/localstack-core/localstack/cli/localstack.py
+++ b/localstack-core/localstack/cli/localstack.py
@@ -4,7 +4,6 @@ import os
 import sys
 import traceback
 from typing import Dict, List, Optional, Tuple, TypedDict
-import functools
 
 import click
 import requests
@@ -145,13 +144,6 @@ _click_format_option = click.option(
     help="The formatting style for the command output.",
 )
 
-# Common global options for all commands.
-def common_options(func):
-    @click.option("-p", "--profile", type=str, help="Set the configuration profile", hidden=True)
-    @functools.wraps(func)
-    def wrapper(*args, **kwargs):
-        return func(*args, **kwargs)
-    return wrapper
 
 @click.group(
     name="localstack",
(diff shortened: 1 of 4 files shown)


thrau (approved):
I think this is a decent solution for now. Clean implementation and well documented :100: Thank you Jim!

jw2:
I've reworked this PR as follows.
1. Daniel pointed out I could use https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args to replace the manual parsing I'd written.
2. There was an issue with stripping out -p parameters from the arguments when these could be genuine non-profile arguments to subcommands, e.g. `localstack start -p 80:8080` to pass a port range to docker.  As such, I've changed the PR so that only `--profile` can appear anywhere on the command line.

Note the updated PR description which discussed the existing bugged behaviour of `-p`.  One other option might be to deprecate use of `-p` for specifying a profile.

@alexrashed could you give this another check please seeing as the PR is substantially changed?

PR #12501, opened 2025-04-08
Scheduled GitHub Action to generate artifacts with feature catalog files

k-a-il:
## Motivation

This GH action will add a GH action which based on schedule produces artifacts with all `features.yml` files (also known as feature catalog files) found in the repository. This GH action re-uses this [GH workflow](https://github.com/localstack/meta/pull/17). 

This GitHub Action will be disabled until mid-May, when we start with writing feature files for other services

## Changes

- Adds a new workflow which generates an artifact with `features.yml` files

### DIFF
--- a/.github/workflows/create_artifact_with_features_files.yml
+++ b/.github/workflows/create_artifact_with_features_files.yml
@@ -1,0 +1,14 @@
+name: AWS / Archive feature files
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
Nice and clean usage of the shared workflow introduced in `localstack/meta`! 💯

PR #12505, opened 2025-04-09
Add dry_run support to GenerateDataKeyPair/GenerateDataKeyPairWithoutPlaintext.

macpak:
## Motivation
I am adding support for the _dry_run_ parameter to the **GenerateDataKeyPair** and **GenerateDataKeyPairWithoutPlaintext** operations to align with AWS behavior and enhance validation capabilities. 

Reference: [AWS KMS Dry Run Documentation](https://docs.aws.amazon.com/kms/latest/developerguide/testing-permissions.html)

## Changes
Before these changes, the **GenerateDataKeyPair** and **GenerateDataKeyPairWithoutPlaintext** operations in KMS did not take the _dry_run_ parameter into account. This PR updates both operations to properly respect the _dry_run_ flag, aligning their behavior with AWS expectations.

Since IAM permission checks are not yet implemented in KMS, the primary purpose of this functionality is to determine whether a request would succeed (in which case a **DryRunOperationException** is raised when `dry_run=True`) or fail due to validation errors, rather than to enforce or simulate IAM permissions

## Testing
You can verify the behavior by executing the following CLI commands:
```
awslocal kms generate-data-key-pair \
  --key-id <key_id> \
  --key-pair-spec RSA_2048 \
  --dry-run

awslocal kms generate-data-key-pair-without-plaintext \
  --key-id <key_id> \
  --key-pair-spec RSA_2048 \
  --dry-run
```

When the request is valid, these commands should return a **DryRunOperationException**. If the input is invalid, a validation error will be returned instead.

### DIFF
--- a/localstack-core/localstack/services/kms/exceptions.py
+++ b/localstack-core/localstack/services/kms/exceptions.py
@@ -15,6 +15,7 @@ class TagException(CommonServiceException):
     def __init__(self, message=None):
         super().__init__("TagException", status_code=400, message=message)
 
+
 class DryRunOperationException(CommonServiceException):
     def __init__(self, message="The request would have succeeded, but the DryRun option is set."):
         super().__init__("DryRunOperationException", status_code=400, message=message)
--- a/localstack-core/localstack/services/kms/exceptions.py
+++ b/localstack-core/localstack/services/kms/exceptions.py
@@ -14,8 +14,3 @@ class AccessDeniedException(CommonServiceException):
 class TagException(CommonServiceException):
     def __init__(self, message=None):
         super().__init__("TagException", status_code=400, message=message)
-
-
-class DryRunOperationException(CommonServiceException):
-    def __init__(self, message="The request would have succeeded, but the DryRun option is set."):
-        super().__init__("DryRunOperationException", status_code=400, message=message)
--- a/localstack-core/localstack/services/kms/exceptions.py
+++ b/localstack-core/localstack/services/kms/exceptions.py
@@ -14,3 +14,7 @@ class AccessDeniedException(CommonServiceException):
 class TagException(CommonServiceException):
     def __init__(self, message=None):
         super().__init__("TagException", status_code=400, message=message)
+
+class DryRunOperationException(CommonServiceException):
+    def __init__(self, message="The request would succeed if DryRun wasn’t specified."):
+        super().__init__("DryRunOperationException", status_code=400, message=message)
--- a/localstack-core/localstack/services/kms/exceptions.py
+++ b/localstack-core/localstack/services/kms/exceptions.py
@@ -16,5 +16,5 @@ class TagException(CommonServiceException):
         super().__init__("TagException", status_code=400, message=message)
 
 class DryRunOperationException(CommonServiceException):
-    def __init__(self, message="The request would succeed if DryRun wasn’t specified."):
+    def __init__(self, message="The request would have succeeded, but the DryRun option is set."):
         super().__init__("DryRunOperationException", status_code=400, message=message)
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -197,7 +197,8 @@ class KmsCryptoKey:
             raise ValidationException(
                 f"1 validation error detected: Value '{key_spec}' at 'keySpec' "
                 f"failed to satisfy constraint: Member must satisfy enum value set: "
-                f"{valid_specs}"
+                f"[RSA_2048, ECC_NIST_P384, ECC_NIST_P256, ECC_NIST_P521, HMAC_384, RSA_3072, "
+                f"ECC_SECG_P256K1, RSA_4096, SYMMETRIC_DEFAULT, HMAC_256, HMAC_224, HMAC_512]"
             )
 
     def __init__(self, key_spec: str, key_material: Optional[bytes] = None):
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -47,7 +47,6 @@ from localstack.aws.api.kms import (
     ReplicateKeyRequest,
     SigningAlgorithmSpec,
     TagList,
-    UnsupportedOperationException,
 )
 from localstack.constants import TAG_KEY_CUSTOM_ID
 from localstack.services.kms.exceptions import TagException, ValidationException
@@ -215,7 +214,7 @@ class KmsCryptoKey:
 
         if key_spec == "SYMMETRIC_DEFAULT":
             return
-       
+
         if key_spec.startswith("RSA"):
             key_size = RSA_CRYPTO_KEY_LENGTHS.get(key_spec)
             key = rsa.generate_private_key(public_exponent=65537, key_size=key_size)
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -178,6 +178,32 @@ class KmsCryptoKey:
     key_material: bytes
     key_spec: str
 
+    @staticmethod
+    def assert_valid(key_spec: str):
+        """
+        Validates that the given key_spec is supported in the current context.
+
+        LocalStack supports additional key specs beyond what AWS officially allows.
+        For example, AWS's GenerateDataKeyPair operation supports only the following:
+            - RSA_2048, RSA_3072, RSA_4096
+            - ECC_NIST_P256, ECC_NIST_P384, ECC_NIST_P521, ECC_SECG_P256K1, SM2
+
+        Raises:
+            ValidationException: If key_spec is not in the supported list.
+        """
+        valid_specs = (
+            list(RSA_CRYPTO_KEY_LENGTHS.keys()) +
+            list(ECC_CURVES.keys()) +
+            list(HMAC_RANGE_KEY_LENGTHS.keys())
+        )
+
+        if key_spec not in valid_specs:
+            raise ValidationException(
+                f"1 validation error detected: Value '{key_spec}' at 'keySpec' "
+                f"failed to satisfy constraint: Member must satisfy enum value set: "
+                f"{valid_specs}"
+            )
+
     def __init__(self, key_spec: str, key_material: Optional[bytes] = None):
         self.private_key = None
         self.public_key = None
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -47,6 +47,7 @@ from localstack.aws.api.kms import (
     ReplicateKeyRequest,
     SigningAlgorithmSpec,
     TagList,
+    UnsupportedOperationException,
 )
 from localstack.constants import TAG_KEY_CUSTOM_ID
 from localstack.services.kms.exceptions import TagException, ValidationException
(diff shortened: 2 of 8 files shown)


macpak:

localstack-core/localstack/services/kms/exceptions.py
```diff
@@ -14,3 +14,8 @@ def __init__(self, message: str):
 class TagException(CommonServiceException):
     def __init__(self, message=None):
         super().__init__("TagException", status_code=400, message=message)
+
+
+class DryRunOperationException(CommonServiceException):
+    def __init__(self, message="The request would have succeeded, but the DryRun option is set."):
+        super().__init__("DryRunOperationException", status_code=400, message=message)
```
alexrashed, line 21: **suggestion:** Actually, for KMS most of the exceptions are contained in the service specs and are therefore also contained in our generated API stub. The benefit here is that this is weekly updated, so we can be sure that the operation is still being thrown, and the code and status code kept up to date.
So in my opinion you should remove this section here and [directly reuse the `DryRunOperationException` in the API module](https://github.com/localstack/localstack/blob/b58b0f8ef6b82d38bb847b843e111e98d6c40acf/localstack-core/localstack/aws/api/kms/__init__.py#L332)‏.

localstack-core/localstack/services/kms/utils.py
```diff
@@ -58,3 +60,27 @@ def validate_tag(tag_position: int, tag: Tag) -> None:
...
+        func (Callable[..., T]): The function to be executed.
+        dry_run (bool): Flag indicating whether the execution is a dry run.
+        *args: Positional arguments to pass to the function.
+        **kwargs: Keyword arguments to pass to the function.
+
+    Returns:
+        T: The result of the function call if `dry_run` is False.
+
+    Raises:
+        DryRunOperationException: If `dry_run` is True.
```
alexrashed, line 82: **suggestion:** This is a docstring in the google format, while we use reST in our codebase. Would you mind re-formatting this to a reStructuredText formatted docstring? Same for other docstrings in the PR :)

tests/aws/services/kms/test_kms.py
```diff
@@ -1942,3 +1942,51 @@ def test_encryption_context_generate_data_key_pair_without_plaintext(
...
+
+        with pytest.raises(ClientError) as exc:
+            aws_client.kms.generate_data_key_pair_without_plaintext(
+                KeyId=key_id, KeyPairSpec="RSA_2048", DryRun=True
+            )
+
+        err = exc.value.response
+        assert err["Error"]["Code"] == "DryRunOperationException"
+
+        # Optional: match the exception with snapshot
```
alexrashed, line 1991: **nitpick (non-blocking):** Actually, I would say it's the other way around 😄 We very much favor snapshot validation compared to plain selective asserts, since for snapshot matching we have:
- A clear tracking when it was validated (see `test_kms.validation.json`).
- A way bigger surface of testing: The snapshot matching validates the HTTP status code, the code, and the message. Your assert only covers the code.

So in my opinion you could remove the manual assert and the comment. Same for the integration test above by the way. :)
```suggestion
```

localstack-core/localstack/services/kms/models.py
```diff
@@ -175,6 +175,32 @@ class KmsCryptoKey:
     key_material: bytes
     key_spec: str
 
+    @staticmethod
+    def assert_valid(key_spec: str):
```
alexrashed, line 179: **issue:** Having this function as static method on the model is introduces a few issues here. The vast majority of validations are currently done in the init dunder function of the `KmsCryptoKey` class. These validations need to be considered when checking the dry run. So the code in this static function should move there as well. But this would mean that you need to restructure your code a bit... Let me know if you need any help with that :)

localstack-core/localstack/services/kms/utils.py
```diff
@@ -58,3 +60,27 @@ def validate_tag(tag_position: int, tag: Tag) -> None:
 
     if tag_key.lower().startswith("aws:"):
         raise TagException("Tags beginning with aws: are reserved")
+
+
+def execute_dry_run_capable(func: Callable[..., T], dry_run: bool, *args, **kwargs) -> T:
```
alexrashed, line 65: **thought (non-blocking):** Pretty nice encapsulation here. When looking at this code, this very much looks like a classic Python Decorator to me. Maybe we could (in another iteration) check if we could rewrite this to a Python decorator and really just decorate operations which should support the DryRun feature with it. That comes with more complexity though when thinking about state handling and is definitely out of scope for this PR :)‏

tests/aws/services/kms/test_kms.py
```diff
@@ -1942,3 +1942,51 @@ def test_encryption_context_generate_data_key_pair_without_plaintext(
...
+        with pytest.raises(ClientError) as exc:
+            aws_client.kms.generate_data_key_pair(KeyId=key_id, KeyPairSpec="RSA_2048", DryRun=True)
+
+        err = exc.value.response
+        assert err["Error"]["Code"] == "DryRunOperationException"
+
+        # Optional: match the exception with snapshot
+        snapshot.match("dryrun_exception", err)
+
+    @markers.aws.validated
```
alexrashed, line 1969: **praise:** Nice that the tests are AWS validated from the start! I just ran the test against AWS KMS and it works just fine! 💯 ‏

tests/unit/services/kms/test_kms.py
```diff
@@ -1,8 +1,170 @@
...
+    result = execute_dry_run_capable(lambda: 1 + 1, dry_run=False)
+    assert result == 2
+
+
+def test_execute_dry_run_capable_raises_when_dry():
+    with pytest.raises(DryRunOperationException):
+        execute_dry_run_capable(lambda: "should not run", dry_run=True)
+
+
+@pytest.mark.parametrize(
```
alexrashed, line 30: **question:** The tests here and below are interesting. We usually don't really perform unit tests on the providers themselves, as these are pretty big units and we very much favor complete integration tests (as you very successfully already wrote!). What do you think is the additional value these unit tests here bring compared to the integration tests?

alexrashed (changes requested):
Thanks a lot for the great contribution! In principal the code is looking great! I just found an issue (only parts of the validations are considered for the dry run), and I posted some more nitpicks and questions.
Once they have been addressed we are good to go! 🚀 💯

localstack-core/localstack/services/kms/models.py
macpak: You're absolutely right, but let me share my thought process. The idea behind `dry_run` is to separate validation from execution. In the case of `KmsCryptoKey`, the validation and core business logic are tightly coupled, so I needed to find a way to split them. Since this impacts an additional operation that I wasn't originally planning to modify, I decided to leave the existing logic intact and introduce a separate method.

That said, if we want to approach this more diligently, I'm happy to refactor it properly. What I had in mind is that while we currently use `key_spec` as a plain string, it's not just any string — it comes with constraints. So we could introduce a `KeySpec` value type, which would encapsulate those constraints and provide built-in validation. That way, instead of passing key_spec as a raw string, we'd require a properly constructed value type, effectively separating validation from usage.

By the way, this was one the questions I posted on Slack earlier about whether I need to refactor beyond the initial scope.

alexrashed: I'm sorry if there was a misunderstanding, I did not make that connection. In fact, I might still now understand the full issue here...
I can see that the interconnection between the business logic and the validation logic is not perfect here, but is that something we can ignore in this iteration? Can't you just integrate your validation logic into the init dunder and instantiate the `KmsCryptoKey` here? This would mean that the business logic (generating the cryptographic primitives), but they are not returned or stored anywhere.

macpak: Sure. Let me work on that.

localstack-core/localstack/services/kms/utils.py
macpak: Thanks. I initially tried using the decorator approach, but—perhaps due to my limited Python knowledge—I ran into an issue: the `dry_run` parameter had to be included in the decorated method's signature, even though it was only used within the decorator itself. This meant someone could accidentally (thinking it's not used anywhere)  remove `dry_run` from the method, breaking the decorator’s logic.

localstack-core/localstack/services/kms/models.py
macpak: I’ve moved the assertion logic into the `__init__ `method and also cleaned up a big the validation logic there.

tests/aws/services/kms/test_kms.py
macpak: Fixed

localstack-core/localstack/services/kms/utils.py
macpak: Fixed.

localstack-core/localstack/services/kms/exceptions.py
macpak: Fixed.

tests/unit/services/kms/test_kms.py
macpak: From my perspective, unit tests are especially helpful during the development phase. Your code seems well-suited for a TDD approach, which makes unit testing even more effective. I find that I can write unit tests quickly to validate changes, rather than implementing the logic first and then relying on integration tests, which are usually time-consuming. In my experience, integration tests can grow to the point where they slow down the development process (if a developer wants to validate each of the change with the integration tests) —sometimes taking 20–30 minutes to run. That kind of delay isn’t perfect. Unit tests, on the other hand, typically run in seconds or, at most, a few minutes, making them a far more efficient tool during development. Still, it's based on the solutions I've worked with. 

Of course, unit tests have their downsides—they're still code and require maintenance like any other part of the codebase. But if you asked me whether I'd rather have them or not, I’d definitely choose to have them :)

macpak (commented):
Addressed the review comments and applied the suggested fixes.

localstack-core/localstack/services/kms/models.py
```diff
@@ -195,22 +220,11 @@ def __init__(self, key_spec: str, key_material: Optional[bytes] = None):
...
-                )
             minimum_length, maximum_length = HMAC_RANGE_KEY_LENGTHS.get(key_spec)
             self.key_material = key_material or os.urandom(
                 random.randint(minimum_length, maximum_length)
             )
             return
-        else:
-            # We do not support SM2 - asymmetric keys both suitable for ENCRYPT_DECRYPT and SIGN_VERIFY,
-            # but only used in China AWS regions.
-            raise UnsupportedOperationException(f"KeySpec {key_spec} is not supported")
```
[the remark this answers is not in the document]
macpak: Yes, that was a deliberate decision. I made the change because, based on how the AWS CLI behaves, when an unknown **key-pair-spec** is provided, it returns a **ValidationException**. In the original code, we would return a 500 error—for example, if someone used an invalid **key-pair-spec** like **RSA_20** or something completely off like **blabla**. My changes address both of those cases.
image


However, when I look at the [AWS KMS API documentation](https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKeyPair.html#API_GenerateDataKeyPair_Errors), it states that the service returns an **UnsupportedOperationException**. So it seems there's a significant difference between the documentation and the actual behavior.

I chose to follow a more empirical approach based on observed behavior, though I’ll likely need to verify how the direct API responds to be sure the AWS CLI isn’t masking / transforming the original error responses.

localstack-core/localstack/services/kms/models.py
macpak: Thanks for the suggestion! I use this method outside of the `KmsCryptoKey` class (`_generate_data_key_pair` method), so making it internal wouldn’t work in this case. I need to maintain a clear separation between the validation logic and the actual creation of the key — this is essential for supporting dry-run operations. For that reason, I believe it makes sense to keep the method as-is.

That said, if it were only used within the `__init__ `method, your suggestion would be absolutely perfect.

I also have another idea for improving this further — specifically by extracting `key_spec` into its own type with built-in validation logic. Happy to share an example of that if you're interested!

tests/unit/services/kms/test_kms.py
macpak: Ok, I'll leave the decision to you—if you think the maintenance costs outweigh the benefits of keeping them, I'm happy to remove them.

localstack-core/localstack/services/kms/models.py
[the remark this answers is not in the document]
alexrashed: It makes sense to follow the parity tests here, the AWS documentation is often off.
But in this case, the code comment in the two lines above indicates that this was a very deliberate decision here to raise a 500 because LocalStack does not a specific key spec type ( `SM2`).
If you use `SM2` with your solution, it would return a validation exception, where on AWS it would just work, right? But in my opinion the current solution (an HTTP 500) or maybe even a 501 would be more suitable here than reusing the `ValidationException`.

What do you think about this special case?

localstack-core/localstack/services/kms/models.py
alexrashed: Oh, thanks for the clarification, I didn't see the outside usage. This definitely indicates - as you mentioned - that there are multiple mixed responsibilities in this class, which we can clean up in a later iteration. 💯

alexrashed (commented):
Thanks a lot for addressing the comments once more! The validation is looking good, and the reasoning on the `KmsCryptoKey` is sound. I really only want to check that we don't introduce an issue with the switch from the `UnsupportedOperationException` to the `ValidationException` in some cases. ;)

localstack-core/localstack/services/kms/models.py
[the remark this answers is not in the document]
macpak: I’ve refactored the method, but the validation still behaves a bit differently from the original. We now explicitly raise an `UnsupportedOperationException`, whereas the original code could throw a 500 error due to "indirect validation logic" within some if branches. I’m not entirely sure if that was the intended behavior or just a bug—I’m leaning toward the latter, especially since there are no tests covering that case.
For example, a 500 could be triggered in the original code with an input like:
`key_Spec == "RSA_blabla"`.

macpak (commented):
The remaining comment has been addressed.

tests/unit/services/kms/test_kms.py
alexrashed: I would say we leave them for now, it's an interesting pattern which might become more interesting in the future for specific provider operations which can be considered as independent units / which are not too much interconnected. :)

localstack-core/localstack/services/kms/models.py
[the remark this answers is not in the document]
alexrashed: I think this definitely was a bug, an invalid key spec should raise a 4xx, while an unsupported operation should raise a 5xx! 💯

alexrashed (approved):
Thanks a lot for addressing the comments, really following up on each and every one! I'll go ahead and directly merge this one! 🤩 
Thanks for the contribution! 💯

PR #10054, opened 2024-01-11
Revert "patch HTTPServer fixture to spawn daemon thread"

bentsku:
Reverts localstack/localstack#10052

PR #12507, opened 2025-04-09
unpin pytest-httpserver, fix patches

alexrashed:
## Motivation
In the past we were facing issues with our integration tests timing out due to shutdowns being blocked. It turns out that some of the `pytest-httpserver` threads were not properly shut down. @bentsku fixed this by patching the `HTTPServer` class to spawn a deamon thread to make sure that it cannot block the shutdown of the interpreter with https://github.com/localstack/localstack/pull/10052.
This patch broke with the latest release, as the `start` method was modified (which is why we did not upgrade to the latest version 1.1.2, but introduced an upper limit with https://github.com/localstack/localstack/pull/12303/commits/01a97ba1fa8619da6266815f88c542f7ea952011).

This PR updates the patch and updates the dependency declaration for `pytest-httpserver`. I also filed https://github.com/csernazs/pytest-httpserver/pull/411 which would allow us to remove the patch completely if it gets merged upstream.

## Changes
- Unpins `pytest-httpserver` in the `pyproject.toml` and in the `requirement*.txt` lock files.
- Updates a patch introduced by @bentsku with https://github.com/localstack/localstack/pull/10052 for `pytest-httpserver==1.1.2`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,8 +108,7 @@ test = [
     "pluggy>=1.3.0",
     "pytest>=7.4.2",
     "pytest-split>=0.8.0",
-    # TODO fix issues with pytest-httpserver==1.1.2, remove upper boundary
-    "pytest-httpserver>=1.0.1,<1.1.2",
+    "pytest-httpserver>=1.1.2",
     "pytest-rerunfailures>=12.0",
     "pytest-tinybird>=0.2.0",
     "aws-cdk-lib>=2.88.0",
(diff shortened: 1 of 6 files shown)


bentsku (approved):
LGTM! Thanks for addressing the issue, and glad to see the actual fix was approved and merged upstream! 🥳 soon we can update it and remove this all together! I must have an old PR lying around removing it 👀 #10054 but this is so old and has conflict now 😄

PR #12508, opened 2025-04-09
Support IAM transitive session tagging

dfangl:
## Motivation
This PR adds IAM transitive session tagging support for the STS `assume_role` operation.
By now storing both session tags, and transitive session tags in the session store, we can make use of them for IAM enforcement.

The session tag store for a session will now include both new tags and transitive tags from the assuming session.
This leads to proper aggregation of transitive tags over role chaining operations.

## Changes
* Store transitive session tags when creating new sessions
* Introduce some input validation for tags and transitive tags

### DIFF
--- a/localstack-core/localstack/services/sts/models.py
+++ b/localstack-core/localstack/services/sts/models.py
@@ -1,9 +1,16 @@
+from typing import TypedDict
+
 from localstack.services.stores import AccountRegionBundle, BaseStore, CrossRegionAttribute
+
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/sts/models.py
pinzon, line 1: More typing :partying_face:

pinzon (approved):
LGTM :+1:

Issue #12399, opened 2025-03-17
bug: long startup/execution times for lambdas in step function map


thomaspage:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

lambdas inside of a map in a step function take an extremely long time to start (> 10 minutes), and once they do start the execution takes even longer. i have tried reducing the concurrency to see if this helps, and while it improves the start time (< 5 minutes) it does not improve the execution time. this issue seems to appear in localstack v3.0.0 and higher (latest test on v4.2.0). i was previously manually setting the v2 executor and had no issues until upgrading. i need to upgrade localstack because i need to upgrade boto3 and using upgraded boto3 on lower localstack versions breaks

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker-compose

```
version: '2.1'
services:
  localstack:
    container_name: ${COMPOSE_PROJECT_NAME}-localstack
    image: 490606849374.dkr.ecr.us-east-1.amazonaws.com/gamification/localstack:${LOCALSTACK_VERSION}
    ports:
      - 4566                              # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559"   # external services port range
    environment:
      - DEBUG=${DEBUG:-1}
      - DOCKER_HOST=unix:///var/run/docker.sock
      - LAMBDA_RUNTIME_ENVIRONMENT_TIMEOUT=300 # 5 minutes
    volumes:
      - ./volume:/var/lib/localstack
      - /var/run/docker.sock:/var/run/docker.sock

```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

stepfunction cloudformation
```
---
Resources:
  GamificationInsightsAggregationStateMachine:
    Type: AWS::StepFunctions::StateMachine
    DependsOn: GamificationAggregatesTable
    Properties:
      StateMachineName: gamification-insights-aggregation-state-machine
      LoggingConfiguration:
        Level: ALL
        Destinations:
          - CloudWatchLogsLogGroup:
              LogGroupArn:
                Fn::GetAtt:
                  [GamificationInsightsAggregationStateMachineLogGroup, Arn]
      DefinitionString: |
        {
          "StartAt": "GetDatesList",
          "States": {
            "GetDatesList": {
              "Type": "Task",
              "Resource": "${GamificationAggregationListDatesInRangeLambdaArn}",
[...]

PR #12512, opened 2025-04-10
Step Functions: Increase Retry Attempts on Service Integrations for Resilience Against Transient Network Errors

MEPalma:
## Motivation
To temporarily reduce the impact of transient network instability in concurrent Lambda executions, this PR increases the total number of retry attempts for the boto client from 1 to 5. Due to occasional “connection refused” errors, the previous setting would immediately trigger retry workflows in the state machine. This could significantly increase the runtime of the program if large backoff rates or wait settings were used in case of failures https://github.com/localstack/localstack/issues/12399. This change does not affect the semantics of evaluation. Catch and Retry blocks still function as intended, but it makes the state machine slightly more resilient against non-service errors such as transient connection failures.

## Changes
- increase the total_max_attempts of service integrations


joe4dev (approved):
Looks like a sensible approach to temporarily work around a known issue.

Any idea how we can debug this to discover the root cause? 🤔

PR #12515, opened 2025-04-11
Added GH action to build community image

k-a-il:
## Motivation

Adds a GitHub Action for building the community Docker image, as part of the CircleCI to GitHub Actions migration

## Changes

- Added new file `build-image/action.yml`

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -21,7 +21,7 @@ runs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.11'
+        python-version: 'localstack/.python-version'
 
     - name: Install docker helper dependencies
       shell: bash
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -21,7 +21,7 @@ runs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 'localstack/.python-version'
+        python-version-file: 'localstack/.python-version'
 
     - name: Install docker helper dependencies
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/build-image/action.yml
```diff
@@ -0,0 +1,62 @@
...
+    description: "Name of the artifact containing the built docker image"
+    value: ${{ steps.image-artifact-name.outputs.image-artifact-name }}
+runs:
+  using: "composite"
+  # This GH Action requires localstack repo in 'localstack' dir + full git history (fetch-depth: 0)
+  steps:
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
```
silv-io, line 24: We could try using the [caching functionality](https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies) here.

We could also use the [`python-version-file`](https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies) input here

silv-io (commented):
The action looks good to me in principle.
I think we should do a quick sanity check and use it before merging it. 
I've added some suggestions for how we could handle python caching here. It's only for the setuptools install, though.

We could also look into using the [Docker GHA cache](https://docs.docker.com/build/cache/backends/gha/) in the helper script. This is a stretch goal, however :)

.github/actions/build-image/action.yml
k-a-il: Cool, will check docs

silv-io (approved):
Changes look good to me! This [run](https://github.com/k-a-il/localstack/actions/runs/14408694707) indicates that everything works fine 

Further caching for the docker build step can happen in a next step

PR #12517, opened 2025-04-11
add AVP to list of CFN composite quirks

bentsku:
## Motivation
While implement AVP resource providers, it appeared that we needed to add custom composites refs for some of its resources.

The values are... quite random. Both `Policy` and `IdentitySource` follow the "own resource id"|"policyStoreId" but `PolicyTemplate` is reversed, and has "PolicyStoreId"|"own resource id" order 🤷 

## Changes
- add `AWS::VerifiedPermissions::IdentitySource`, `AWS::VerifiedPermissions::Policy`, `AWS::VerifiedPermissions::PolicyTemplate` to the list of quirks


pinzon (approved):

simonrw (approved):
Thanks for adding these and working out another gap in the cfn specs!

PR #12521, opened 2025-04-14
Add error handling if lambda logs are not received from the environment

dfangl:
## Motivation
Currently, if the lambda environment, for currently unknown reasons, does not post logs to LS for an invocation, we error the invocation while trying to print logs.

We should avoid failing the entire invocation, but clearly log a warning, and ideally log the container execution logs as well, to be able to determine the root cause of the missing logs.

## Changes
* Prevent invocation from failing when we cannot print logs to the console
* Add additional logging to allow investigation of the root cause in the future

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/version_manager.py
+++ b/localstack-core/localstack/services/lambda_/invocation/version_manager.py
@@ -238,13 +238,20 @@ class LambdaVersionManager:
             )
         # TODO: consider using the same prefix logging as in error case for execution environment.
         #   possibly as separate named logger.
-        LOG.debug("Got logs for invocation '%s'", invocation.request_id)
-        for log_line in invocation_result.logs.splitlines():
-            LOG.debug(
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/invocation/version_manager.py
```diff
@@ -260,7 +267,8 @@ def store_logs(
             self.log_handler.add_logs(log_item)
         else:
             LOG.warning(
-                "Received no logs from invocation with id %s for lambda %s",
+                "Received no logs from invocation with id %s for lambda %s. Execution environment logs: \n%s",
                 invocation_result.request_id,
                 self.function_arn,
+                execution_env.get_prefixed_logs(),
```
joe4dev, line 273: Good idea to present actionable output 👍 

Hopefully, not too long 🤞

joe4dev (approved):
More resilient against missing logs 🩹 🚀

PR #12522, opened 2025-04-14
Lambda: fix transient connection errors on first container invoke with retry logic

MEPalma:
## Motivation
This change improves the robustness of the container invocation logic by addressing a timing issue observed during Step Functions integration tests https://github.com/localstack/localstack/pull/12512, where transient ConnectionErrors occurred despite the container signaling readiness via `startup_future`. Although the startup flag may be marked as FINISHED, this does not guarantee that the container is immediately ready to accept incoming connections. To handle this transitional state, the _perform_invoke method has been updated to automatically retry on requests.exceptions.ConnectionError up to `_INVOCATION_CONNECTION_ERROR_MAX_RETRY` times, using exponential backoff defined by `_INVOCATION_CONNECTION_ERROR_BACKOFF_FACTOR_SECONDS`. 

Alternative solutions were considered: 
- awaiting the `startup_future` within the `status_read` logic was explored, but we currently lack a reliable mechanism to verify full container readiness without issuing an actual invocation
- similarly, deferring the `startup_future` resolution until the first successful invoke was also evaluated, but since transient network errors may occur beyond the initial request, a retry mechanism on every invoke call was deemed both simpler and more robust. 
Although this retry logic is primarily expected to benefit the first invocation following container startup, it applies to all invokes to ensure resilience under intermittent network conditions.

## Changes
- Implemented retry logic in _perform_invoke for requests.exceptions.ConnectionError
- Configured retries using _INVOCATION_CONNECTION_ERROR_MAX_RETRY with exponential backoff via _INVOCATION_CONNECTION_ERROR_BACKOFF_FACTOR_SECONDS
- Reverted Step Functions' `total_max_attempts` back to 1, as this issue appears now resolved


localstack-core/localstack/services/lambda_/invocation/executor_endpoint.py
```diff
@@ -214,3 +218,66 @@ def invoke(self, payload: Dict[str, str]) -> InvocationResult:
...
+            requests.exceptions.ConnectionError: If all retry attempts fail to connect.
+        """
+        backoff = None
+        last_exception = None
+        max_retry_on_connection_error = 5
+
+        for attempt_count in range(max_retry_on_connection_error + 1):  # 1 initial + n retries
+            try:
+                response = requests.post(url=invocation_url, json=payload, proxies=proxies)
+                response.raise_for_status()
```
dfangl, line 253: The response is checked on line 200 as well - this will prevent the other error handling, and leak request errors to the callers (instead of the expected InvokeSendError). We should not raise for the status here, I think the connection error should still be raised, right?

MEPalma: Yes indeed I could confirm; I removed this raise in the invoke routine

PR #12525, opened 2025-04-15
Step Functions: Surface Support for Mocked Responses

MEPalma:
## Motivation
These changes introduce base support for executing AWS Step Functions state machines in a mock-enabled mode, designed to streamline testing and simulation workflows. The primary motivation is to enable users to execute state machines using mock configuration files, allowing simulated task execution by sampling predefined responses from a mock oracle. This capability is particularly valuable for validating control flow logic and response handling in a deterministic manner without invoking actual downstream services [AWS Docs](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-test-sm-exec.html)

The changes in this PR surface support for all critical components necessary to enable this workflow. Future changes will refine and expand on each aspect introduced in this PR

## Changes
- Parsing mock configuration files that define the oracle of mocked responses
- Selecting specific test cases through standard state machine execution APIs by appending #test-case-name to the execution ARN
- Introducing a mock mode in the environment
- Sampling the mocked response at runtime based on configuration and execution context
- Injecting the mocked response in place of real task logic
- Support for Return and Throw response types
- Snapshot testing utils to compare mocked executions against real AWS executions


localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -1,69 +1,214 @@
...
+
+from localstack.services.stepfunctions.mocking.mock_config_file import (
+    RawMockConfig,
+    RawResponseModel,
+    RawTestCase,
+    _load_sfn_raw_mock_config,
+)
+
+
+class MockedResponse(abc.ABC):
```
gregfurman, line 12: Nice one on modelling all these requests/responses 😄

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -1,69 +1,214 @@
...
+        try:
+            range_value = int(range_part)
+            return range_value, range_value
+        except Exception:
+            raise RuntimeError(
+                f"Unknown mocked response retry range value '{range_part}', not a valid integer"
+            )
+    elif len(definition_parts) == 2:
+        range_part_start = definition_parts[0]
+        range_part_end = definition_parts[1]
```
gregfurman, line 113: Should we be checking that `range_part_start > range_part_end`?

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
```diff
@@ -1,69 +1,214 @@
...
+            state_name = state_mocked_response.state_name
+            if state_name in self.state_mocked_responses:
+                raise RuntimeError(
+                    f"Duplicate definition of state '{state_name}' for test case '{test_case_name}'"
+                )
+            self.state_mocked_responses[state_name] = state_mocked_response
+
+
+def _parse_mocked_response_range(string_definition: str) -> tuple[int, int]:
+    definition_parts = string_definition.strip().split("-")
```
gregfurman, line 101: Do we need to validate these values at all? Like are there valid bounds?

localstack-core/localstack/services/stepfunctions/mocking/mock_config_file.py
```diff
@@ -0,0 +1,150 @@
+import logging
+import os
+from functools import lru_cache
+from typing import Dict, Final, Optional
+
+from pydantic import BaseModel, RootModel, model_validator
```
gregfurman, line 6: pydantic ❤️

localstack-core/localstack/services/stepfunctions/mocking/mock_config_file.py
```diff
@@ -0,0 +1,150 @@
...
+_THROW_KEY: Final[str] = "Throw"
+
+
+class RawReturnResponse(BaseModel):
+    """
+    Represents a return response.
+    Accepts any fields.
+    """
+
+    model_config = {"extra": "allow", "frozen": True}
```
gregfurman, line 22: nit: Should we be using a `pydantic.dataclass` here instead? It contains a `frozen` attribute out-the-box so am wondering about the `BaseModel` approach

localstack-core/localstack/services/stepfunctions/mocking/mock_config_file.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    model_config = {"frozen": True}
+
+    StateMachines: Dict[str, RawStateMachine]
+    MockedResponses: Dict[str, Dict[str, RawResponseModel]]
+
+
+@lru_cache(maxsize=1)
+def _read_sfn_raw_mock_config(file_path: str, modified_epoch: int) -> Optional[RawMockConfig]:  # noqa
```
gregfurman, line 93: question: this will only cache a single call. I imagine we wont be reading in multiple mock files?

gregfurman (approved):
LGTM! Some very small nits/questions. Otherwise, I think this is good to go 👍

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
MEPalma: We check when trying to construct the mock responses from these values `localstack.services.stepfunctions.mocking.mock_config.StateMockedResponses.__init__`

localstack-core/localstack/services/stepfunctions/mocking/mock_config.py
MEPalma: Yes indeed, we do this after parsing the values and constructing the StateMockedResponses object for the runtime later `localstack.services.stepfunctions.mocking.mock_config.StateMockedResponses.__init__`

localstack-core/localstack/services/stepfunctions/mocking/mock_config_file.py
MEPalma: correct, for now we'll be loading one file only

localstack-core/localstack/services/stepfunctions/mocking/mock_config_file.py
MEPalma: Definitely a good point you raise! In this case we still need a few BaseModel features, I can see:
- extra="allow" on RawReturnResponse
- the root model on RawTestCase
- the model_validator in RawResponseModel
Pydantic’s dataclass doesn’t support any of those as far as I can see, and we'd also lose the model_dump() (although this may not be an issue by itself if we were using records)

PR #12531, opened 2025-04-16
[ESM] Support discarding events exceeding MaxRecordAgeInSeconds

gregfurman:
## Motivation
To further improve ESM parity, this PR adds support for expiring/discarding events based on their age.

## Changes
- Refactor `StreamPoller` processing loop to account for discarding records based on event age.
- Add `bisect_events_by_record_age` function to bisect retrieved records in a stream based on their age.

## Testing
- `test_kinesis_maximum_record_age_exceeded` -- tests an expired event being ingested prior to invocation as well as an event expiring while retrying.
- `test_kinesis_maximum_record_age_exceeded_discard_records` -- tests discarding an expired record within a batch of valid records.

## TODO

What's left to do:

- [ ] Add changes in pipes.
- [ ] Potentially add validations in a sibling PR.
- [ ] Re-evaluate [ESM behaviour docs](https://docs.localstack.cloud/user-guide/aws/lambda/#behaviour-coverage) since we were (incorrectly) saying this behaviour was supported when in actuality it wasn't.

### DIFF
--- a/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
+++ b/localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
@@ -2,6 +2,7 @@ import json
 import logging
 import threading
 from abc import abstractmethod
+from bisect import bisect_left
 from collections import defaultdict
 from datetime import datetime
 from typing import Iterator
@@ -209,16 +210,7 @@ class StreamPoller(Poller):
 
     def forward_events_to_target(self, shard_id, next_shard_iterator, records):
         polled_events = self.transform_into_events(records, shard_id)
-
         abort_condition = None
-        # Check MaximumRecordAgeInSeconds
-        if maximum_record_age_in_seconds := self.stream_parameters.get("MaximumRecordAgeInSeconds"):
-            arrival_timestamp_of_last_event = polled_events[-1]["approximateArrivalTimestamp"]
-            now = get_current_time().timestamp()
-            record_age_in_seconds = now - arrival_timestamp_of_last_event
-            if record_age_in_seconds > maximum_record_age_in_seconds:
-                abort_condition = "RecordAgeExpired"
-
         # TODO: implement format detection behavior (e.g., for JSON body):
         #  https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-event-filtering.html
         #  Check whether we need poller-specific filter-preprocessing here without modifying the actual event!
@@ -250,16 +242,26 @@ class StreamPoller(Poller):
         # TODO: handle partial batch failure (see poller.py:parse_batch_item_failures)
         # TODO: think about how to avoid starvation of other shards if one shard runs into infinite retries
         attempts = 0
+        discarded_events_for_dlq = []
         error_payload = {}
 
         max_retries = self.stream_parameters.get("MaximumRetryAttempts", -1)
+        max_record_age = max(
+            self.stream_parameters.get("MaximumRecordAgeInSeconds", -1), 0
+        )  # Disable check if -1
         # NOTE: max_retries == 0 means exponential backoff is disabled
         boff = ExponentialBackoff(max_retries=max_retries)
-        while (
-            not abort_condition
-            and not self.max_retries_exceeded(attempts)
-            and not self._is_shutdown.is_set()
-        ):
+        while not abort_condition and events and not self._is_shutdown.is_set():
+            if self.max_retries_exceeded(attempts):
+                abort_condition = "RetryAttemptsExhausted"
+                break
+
+            if max_record_age:
+                events, expired_events = self.bisect_events_by_record_age(max_record_age, events)
+                if expired_events:
+                    discarded_events_for_dlq.extend(expired_events)
+                    continue
+
             try:
                 if attempts > 0:
                     # TODO: Should we always backoff (with jitter) before processing since we may not want multiple pollers
@@ -269,10 +271,8 @@ class StreamPoller(Poller):
 
                 self.processor.process_events_batch(events)
                 boff.reset()
-
-                # Update shard iterator if execution is successful
-                self.shards[shard_id] = next_shard_iterator
-                return
+                # Stop processing if execution is successful
+                break
             except PartialBatchFailureError as ex:
                 # TODO: add tests for partial batch failure scenarios
                 if (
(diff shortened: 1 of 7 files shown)


tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
```diff
@@ -1054,6 +1056,226 @@ def _verify_messages_received():
...
+        event_source_mapping_uuid = create_event_source_mapping_response["UUID"]
+        _await_event_source_mapping_enabled(aws_client.lambda_, event_source_mapping_uuid)
+
+        def _verify_failure_received():
+            result = aws_client.sqs.receive_message(QueueUrl=queue_event_source_mapping)
+            assert result.get("Messages")
+            return result
+
+        sleep = 15 if is_aws_cloud() else 5
+        record_age_exceeded_payload = retry(
```
joe4dev, line 1150: FYI: Could consider re-using `localstack.testing.pytest.fixtures.sqs_collect_messages` (potentially 7->2 LOC). One downside would be that `sqs_collect_messages` does not (yet) support `sleep_before` :)

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
```diff
@@ -1054,6 +1056,226 @@ def _verify_messages_received():
...
+        _await_event_source_mapping_enabled(aws_client.lambda_, event_source_mapping_uuid)
+
+        def _verify_failure_received():
+            result = aws_client.sqs.receive_message(QueueUrl=dead_letter_queue)
+            assert result.get("Messages")
+            return result
+
+        batches = []
+
+        def _verify_events_received(expected: int):
```
joe4dev, line 1244: Could the `sqs_collect_messages` fixture replace/simplify this?

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -250,16 +242,26 @@ def forward_events_to_target(self, shard_id, next_shard_iterator, records):
         # TODO: handle partial batch failure (see poller.py:parse_batch_item_failures)
```
joe4dev, line 242: The new test covers this TODO question/scenario of a record expiring at the source, right?
`# TODO: A retry should probably re-trigger fetching the record from the stream again?!`

joe4dev (approved):
Excellent work @gregfurman 
These are well-designed complex test cases 🙇 
Thank you for improving the horrifically complex control flow 🤯 
Good follow up items 👏 

I added small suggestions, but nothing blocking a merge after a 🟢 ext run from Pipes 🚀

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
gregfurman: Unfortunately that wont simplify things much since we need to extract the SQS message body -- which could contain a list of events.

Thinking that maybe the Lambda function code could be modified to ingest 1 message per event (as opposed to a dump of all evets as 1 list), which would work better with the `sqs_collect_messages` fixture.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -479,6 +484,17 @@ def bisect_events(
 
         return events, []
 
+    def bisect_events_by_record_age(
+        self, maximum_record_age: int, events: list[dict]
+    ) -> tuple[list[dict], list[dict]]:
+        """Splits events into [valid_events], [expired_events] based on record age.
+        Where:
+          - Events with approximate arrival time >= cutoff_timestamp are valid.
```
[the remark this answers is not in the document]
gregfurman: That's true 🤦 WIll change!

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
joe4dev: Good point with that inner event unpacking 👍 . Given how often we use that pattern, it feels we should have a fixture alike `sqs_collect_events` based on `_verify_events_received`. That would allow us to keep the Lambda code simple and carry as much information as possible for snapshotting in case we need it (rather than discarding this info in the Lambda code). wdyt?

gregfurman: 100% agree. Let me add in a follow-up? It should be backwards compatible in anycase

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -269,10 +271,8 @@ def forward_events_to_target(self, shard_id, next_shard_iterator, records):
 
                 self.processor.process_events_batch(events)
                 boff.reset()
-
-                # Update shard iterator if execution is successful
-                self.shards[shard_id] = next_shard_iterator
-                return
+                # Stop processing if execution is successful
```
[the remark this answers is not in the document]
gregfurman: Yeah exactly! Will add a comment 😄

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
```diff
@@ -327,15 +327,20 @@ def forward_events_to_target(self, shard_id, next_shard_iterator, records):
...
-        self.send_events_to_dlq(shard_id, events, context=failure_context)
+        if abort_condition:
+            failure_context = self.processor.generate_event_failure_context(
+                abort_condition=abort_condition,
+                error=error_payload,
+                attempts_count=attempts,
+                partner_resource_arn=self.partner_resource_arn,
+            )
+            self.send_events_to_dlq(shard_id, events, context=failure_context)
         # Update shard iterator if the execution failed but the events are sent to a DLQ
```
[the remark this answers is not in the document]
gregfurman: Tbh I think we should investigate whether these DLQ sends can happen mid-batch processing or if they're just relegated to occurring after-the-fact.

Like. the records check sends to a DLQ but doesn't inhibit processing at all.

localstack-core/localstack/services/lambda_/event_source_mapping/pollers/stream_poller.py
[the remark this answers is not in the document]
gregfurman: Well... technically this doesn't check presedence wrt to filtering and batching windows being defined simultaneously

gregfurman: We can't delete records from a kinesis/ddb stream -- so don't think re-triggering a fetch is ever necessary.

With that said... I imagine we could consider doing another fetch if we wanted to exclude certain records (i.e instead of the bisect + discard functionality, we trigger a `get_records` and point the starting position/timestamp to after the last expired record/timestamp).

I'll remove the comment though since I think we have coverage on this though.

tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_kinesis.py
gregfurman: I've given this fixture another look. I think it's actually kinda broken since the `timeout` is only ever decremented by the `interval` duration -- so if you set a timeout to `timeout=100` and `interval=0.1`, we'd expect ~1000 polls.

If the condition (in this case `ReceiveMessage` calls) has long polling (i.e `WaitTimeSeconds=5`) we'd expect a failure to take `5000s + 100s`:

i.e
```python
    remaining = 0
    if timeout is not None:
        remaining = timeout

    while not condition(): # What if this takes 5s?
        if timeout is not None:
            remaining -= interval

            if remaining <= 0:
                return False

        time.sleep(interval)

    return True

```

joe4dev: Good catch 🕵️ 

I guess we should be using some monotonically increasing counter instead 💡

PR #12533, opened 2025-04-16
CloudFormation Engine V2: Improve delta computation of properties, conditional resolution, and physical resources ref

MEPalma:
## Motivation
This PR resolves several issues in the Engine V2 provider related to template delta handling and reference evaluation. Specifically, it addresses:
- Misinterpretation of deltas in templates where property definitions are empty or missing
- Faulty comparison logic in preprocessor delta conditions, where the absence of a condition was incorrectly treated as a match
- Limited handling of reference evaluation when a physical ID changes during updates
- Additional minor issues and inconsistencies

## Changes
- Fixed misinterpretation of deltas in templates with empty/missing properties
- Corrected logic for comparing preprocessor delta conditions
- Improved handling of Ref evaluation across physical ID changes
- Fixed the lack of describe and execution pruning for unchanged resources
- Addressed various minor issues

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -769,7 +769,10 @@ class ChangeSetModel:
             before_properties=before_properties,
             after_properties=after_properties,
         )
-        change_type = change_type.for_child(properties.change_type)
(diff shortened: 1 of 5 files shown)


simonrw (approved):
Nice one thanks for fixing my hacks and the tests!

PR #12534, opened 2025-04-17
revert removal of get_resource_type method

cloutierMat:
## Motivation

Changes from #12511 broke custom resources plugin loading in upstream repo.

## Changes

- Bring back usage of the method


dominikschubert (approved):
Thanks for unblocking this. We can look into another attempt at removing this at a later stage but since custom resources can be named arbitrarily with a `Custom::` prefix this might be tricky

PR #12537, opened 2025-04-17
Cloud Formation v2 Engine: Support for Default fields in Parameters

MEPalma:
\* Extends https://github.com/localstack/localstack/pull/12535

## Motivation
Currently, the CFn v2 engine only supports CFn parameters if set through dynamic assignment as api action arguments. Moreover, the internal representation of Parameter definition is limited to a generic set of bindings.

## Changes
- Rework the internal representation of Parameters to include the supported types of Dynamic, Default, and Type
- Add support for parameter resolution through Default fields
- Add base boundary tests for the update of templates about Default and Dynamic values of Parameters


tests/aws/services/cloudformation/v2/test_change_set_parameters.py
```diff
@@ -0,0 +1,128 @@
+import pytest
+from localstack_snapshot.snapshots.transformer import RegexTransformer
+
+from localstack.services.cloudformation.v2.utils import is_v2_engine
+from localstack.testing.pytest import markers
+from localstack.utils.strings import long_uid
+
+
+@pytest.mark.skipif(condition=not is_v2_engine(), reason="Requires the V2 engine")
```
simonrw, line 9: A minor ease of use feature

```suggestion
@pytest.mark.skipif(condition=not is_v2_engine() and not is_aws_cloud(), reason="Requires the V2 engine")
```

simonrw (approved):
A nice comprehensive set of tests, and a clean implementation, thank you!

tests/aws/services/cloudformation/v2/test_change_set_parameters.py
MEPalma: I changed this in other test files too

PR #12539, opened 2025-04-19
APIGW: validate REST API custom id tag

bentsku:
## Motivation
When setting a custom id for a REST API and using the `_custom_id_` tag, using an hyphen will make the API fail because the routing won't work, as the regex explicitly doesn't match hyphens: `<regex('[^-]+'):api_id>`, to avoid matching on the `-vpce` part. 
It would fail silently, because it wouldn't be able to find the API. 

We now validate the custom id to not allow `-` character in it. 

Note: maybe another solution would be to adapt the regex to allow hyphens, but didn't want to take the risk to change it, and this has been the behavior since this has been introduced. 

## Changes
- add validation and adapt existing test


cloutierMat (approved):
I totally agree with the decision to to not accept apigw ids with `-` in it. :+1: 

Thanks for adding the tests also!

PR #12545, opened 2025-04-22
Migrate full-run integration tests from CircleCI to GH Actions except docker push and push of all artifacts

k-a-il:
## Motivation

This PR is part of initiative to migrate from CircleCI to GithubActions. It migrates part of full-run tests to GithubActions. This PR does not include merged artifact uploads or Docker push to registry jobs; they will be added in a separate PR.

## Changes

- Removed parallel coverage runs from service tests that execute on a single runner
- re-uses Github Action to build community image
- introduces Github Action to prepare test environment and load docker image from artifacts
- caching of pip dependencies is done using `actions/setup-python@v5` 

## Testing

- Pipeline testing acceptance tests run: https://github.com/k-a-il/localstack/actions/runs/14730738450
- Full integration tests run (tests in TestDockerClient were skipped for now): https://github.com/k-a-il/localstack/actions/runs/14795627190


.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+  schedule:
+    - cron: 0 2 * * MON-FRI
+  push:
+    paths-ignore:
+      - 'CODEOWNERS'
+      - 'README.md'
+      - '.gitignore'
+      - '.git-blame-ignore-revs'
+    branches-ignore:
+      - master
```
silv-io, line 13: **q:** This is temporary until we switch to GHA completely, right?

if yes:
**suggestion:** put a todo here so that this is clear in the meantime

.github/actions/load-localstack-docker-from-artifacts/action.yml
```diff
@@ -0,0 +1,29 @@
...
+      uses: actions/download-artifact@v4
+      with:
+        name: localstack-docker-image-${{ inputs.platform }}
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version-file: '.python-version'
+        cache: 'pip'
+        cache-dependency-path: 'requirements-dev.txt'
```
silv-io, line 21: **q:** why do we use requirements-dev.txt as the cache dependency path here?

.github/actions/load-localstack-docker-from-artifacts/action.yml
silv-io, line 1: **praise:** I really like this and the other composite action. It makes the integration test much more concise and less noisy

.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+      onlyAcceptanceTests:
+        description: "Run only acceptance tests"
+        default: false
+        required: false
+        type: boolean
+
+env:
+  PYTEST_LOGLEVEL: ${{ inputs.PYTEST_LOGLEVEL || 'WARNING' }}
+  IMAGE_NAME: "localstack/localstack"
+  TINYBIRD_DATASOURCE: "community_tests"
```
silv-io, line 45: **thought:** we should maybe already start thinking about how we will structure the names for tinybird. I guess this can have the name `community_tests_integration` (or `community_tests_mamr` in the case of MA/MR)

.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+          name: test-selection
+          path: |
+            dist/testselection/test-selection.txt
+          retention-days: 1
+
+      - name: List files
+        shell: bash
+        run: |
+          ls -a
+          ls -a dist/testselection/
```
silv-io, line 125: **nit:** This seems to be a leftover from debugging

.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+        with:
+          name: test-selection
+          path: dist/testselection/
+
+      - name: Run Integration Tests
+        timeout-minutes: 120
+        env:
+          # add the GitHub API token to avoid rate limit issues
+          GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PYTEST_ARGS: "${{ env.TINYBIRD_PYTEST_ARGS }}${{ env.TESTSELECTION_PYTEST_ARGS }} --splits 2 --group ${{ matrix.group }} --store-durations --clean-durations --ignore=tests/unit/ --ignore=tests/bootstrap"
```
silv-io, line 233: **q:** was there no possibility to determine the amount of splits from the matrix group? Or would it be too verbose?

.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: test-selection
+          path: dist/testselection/
+
+      - name: List files
+        shell: bash
+        run: |
+          ls -a
+          ls -a dist/testselection/
```
silv-io, line 367: **change:** We don't need test selection for the acceptance tests. They should always run in full.

.github/workflows/tests-integration.yml
```diff
@@ -0,0 +1,627 @@
...
+        if: success() || failure()
+        with:
+          name: test-results-acceptance-${{ env.PLATFORM }}-${{ matrix.group }}
+          include-hidden-files: true
+          path: |
+            target/pytest-junit-acceptance-${{ env.PLATFORM }}-${{ matrix.group }}.xml
+            target/.coverage.acceptance-${{ env.PLATFORM }}-${{ matrix.group }}
+          retention-days: 30
+
+  test-cloudwatch-v1:
```
silv-io, line 395: **thought:** These alternative provider test runs all seem similar enough so that we could put them in a matrix together actually. Would be a nice-to-have but we don't really need to.

Makefile
```diff
@@ -91,7 +91,7 @@ start:             		  ## Manually start the local infrastructure for testing
 	($(VENV_RUN); exec bin/localstack start --host)
 
 docker-run-tests:		  ## Initializes the test environment and runs the tests in a docker container
-	docker run -e LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC=1 --entrypoint= -v `pwd`/.git:/opt/code/localstack/.git -v `pwd`/requirements-test.txt:/opt/code/localstack/requirements-test.txt -v `pwd`/tests/:/opt/code/localstack/tests/ -v `pwd`/target/:/opt/code/localstack/target/ -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/localstack:/var/lib/localstack \
+	docker run -e LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC=1 --entrypoint= -v `pwd`/.git:/opt/code/localstack/.git -v `pwd`/requirements-test.txt:/opt/code/localstack/requirements-test.txt -v `pwd`/tests/:/opt/code/localstack/tests/ -v `pwd`/dist/:/opt/code/localstack/dist/ -v `pwd`/target/:/opt/code/localstack/target/ -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/localstack:/var/lib/localstack \
```
silv-io, line 94: **praise:** I can only imagine how annoying this fix was to find, kudos!

silv-io (commented):
LGTM in principle! Some awesome ideas in there with the composite actions and more 🥳 

I will give the final approval after having looked at the full test runs.

.github/actions/load-localstack-docker-from-artifacts/action.yml
k-a-il: A cache key is constructed based on the file `requirements-dev.txt`. This requirements file was used in CircleCI configuration as well.

.github/workflows/tests-integration.yml
k-a-il: Removed

.github/workflows/tests-integration.yml
k-a-il: That's a cool idea, though I'm not sure it's feasible since the environment variables differ between tests.

.github/workflows/tests-integration.yml
silv-io: Alright, this seems to verbose then. We can leave that out 👍🏼

.github/workflows/tests-integration.yml
k-a-il: Fixed 👍

.github/workflows/aws-tests.yml
```diff
@@ -0,0 +1,676 @@
...
+      - name: Archive Test Results
+        uses: actions/upload-artifact@v4
+        if: success() || failure()
+        with:
+          name: test-results-integration-${{ env.PLATFORM }}-${{ matrix.group }}
+          include-hidden-files: true
+          path: |
+            target/pytest-junit-integration-${{ env.PLATFORM }}-${{ matrix.group }}.xml
+            target/.coverage.integration-${{ env.PLATFORM }}-${{ matrix.group }}
+          retention-days: 30
```
silv-io, line 296: **nit:** Just noticed that we will also need to add this here, like we have in ext: https://github.com/EnricoMi/publish-unit-test-result-action so that we can quickly browse the results in the Summary. 

We can either add this now or in a follow-up PR. And we'll need that for all test jobs

silv-io (approved):
LGTM! It's very easily readable which I think is very important for pipelines like these 🚀

Only the publishing of the results to the GHA summary would be left now. (Sorry that I came up with that just now :P) However, we can also add that in a follow up.

.github/workflows/aws-tests.yml
k-a-il: I will add this in the follow-up PR

PR #12546, opened 2025-04-22
CFn: remove get_resource_type

simonrw:
## Motivation

In a [previous PR](https://github.com/localstack/localstack/pull/12511) I incorrectly removed the `get_resource_type` function from CloudFormation. This broke CI for our private repository so the function was replaced in [#12534](https://github.com/localstack/localstack/pull/12534).

This function _is_ suitable for removal since it's not used in the private codebase, and it's API is a little strange.

## Changes

Swap `get_resource_type` for `standardise_resource_type` which takes the resource type, and _if_ it is a custom resource, standardise the name. Otherwise return the name unaffected.

## Testing

* Community testing should be enough as it's a core component of the CloudFormation engine
* I will perform a manual run of our private CI to ensure we don't break that code base


simonrw:
Closing this PR since we are not using this function in the v2 provider

PR #12548, opened 2025-04-23
Manually update ASF APIs and all dependencies

alexrashed:
## Motivation
With the latest release of `awscli==1.39.0` / `botocore==1.38.0` / `boto3==1.38.0` the [ASF update action started to fail with dependency resolution issues](https://github.com/localstack/localstack/actions/runs/14612933999):
```
ERROR: Cannot install boto3==1.38.0, botocore==1.38.0, localstack-core (pyproject.toml), moto-ext and moto-ext[all]==5.1.3.post1 because these package versions have conflicting dependencies.
```
It turns out that this is due to the fact that there was an update of a transitive dependency of `awscli`: `docutils` was upgraded to version `0.19.0`. This does not work with the way how we currently update the package as it does not consider transitive dependencies of `awscli`:
https://github.com/localstack/localstack/blob/77f30dcd2ed68017e4972d81d5f69a86bcab2b5f/.github/workflows/asf-updates.yml#L86

This was now manually fixed by removing the existing `requirements-*.txt` files and regenerating them.

## Changes
- Upgrade `boto3` / `botocore` to `1.39.0`.
- Regenerate all `requirements-*.txt` files.
- Generate the updated ASF APIs - the following services were updated:
  - `ec2` 
  - `events`
  - `resource_groups`
- Updates the `events` provider signature after the ASF specs were updated (added `kms_key_identifier`). /cc @maxhoheiser @bentsku

### DIFF
--- a/localstack-core/localstack/aws/api/ec2/__init__.py
+++ b/localstack-core/localstack/aws/api/ec2/__init__.py
@@ -3171,6 +3171,7 @@ class ServiceConnectivityType(StrEnum):
 class ServiceManaged(StrEnum):
     alb = "alb"
     nlb = "nlb"
+    rnat = "rnat"
 
 
 class ServiceState(StrEnum):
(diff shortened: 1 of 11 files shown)


bentsku (approved):
LGTM! Nice investigation 👀 interesting how the error message by pip seemed really off from the real issue! 

Small nit, in the PR description you're mentioning we are updating `botocore` and `boto3` to `1.39.0`, but this is the `awscli` version, those packages are updated to `1.38.0`. 
edit: updated it myself 😄 

Thanks a lot for unblocking the ASF Update workflow! 🙏

PR #12549, opened 2025-04-23
APIGW: fix Host regex to allow hyphen and remove restriction

bentsku:
## Motivation
Follow-up from #12539, sorry for back and forth on this.

A failure on our upstream pipeline put into light the fact that the `-` character would still work when using "path based routing" for the API.
In order to not break existing users' setup, I'm removing the previous restriction, and instead updating the regex to allow for dashes in the REST API id, as this is actually a simple fix (making the quantifier be lazy to not match the next regex group). 

## Changes
- add a `only_localstack` test showing that the routing works with hyphen
- remove previous restriction


cloutierMat (approved):
Gotta love those simple fix when they finally appear! Thanks for the follow-up and the added test

PR #12550, opened 2025-04-23
Step Functions: Decrease LocalStack Sampling Delays to Speed Up Tests Suite

MEPalma:
## Motivation
The Step Functions v2 integration suite currently comprises ~900 tests that all poll for state-machine completion. Because we use identical polling logic, without any environment-specific callbacks, for both AWS and LocalStack, every test incurs the same wait time regardless of backend. 
In practice, LocalStack completes executions far more quickly than AWS, so these uniform delays become needless overhead when running tests locally. As the test count continues to grow (especially with the addition of many fast mock-response tests), overall runtime will only climb higher, slowing down feedback for developers.

With these changes I cut the sampling wait periods down to `0.2` seconds for LocalStack runs, which trims the overall test suite runtime by about 1.5 folds locally.

## Changes
- Reduce sampling interval to 0.2 seconds for runs against LocalStack


joe4dev (approved):
Neat optimization and cleanly standardized ✨

PR #12559, opened 2025-04-25
 [Kinesis] add Scala kinesis-mock build behind feature flag 

gregfurman:
## Motivation
We've observed the `NodeJS` build of kinesis mock to encounter performance issues at higher volumes of requests -- namely with larger payload sizes.

This PR adds the original Scala build as an alternative engine for our Kinesis server behind the `KINESIS_MOCK_PROVIDER_ENGINE` flag.

### Notes
- Relevant upstream issue: https://github.com/etspaceman/kinesis-mock/issues/1082
- Ext run with flag enabled: https://github.com/localstack/localstack-ext/actions/runs/14646658785

### Functionality
- Splits up the `KinesisMockServer` to include a `KinesisMockScalaServer` and `KinesisMockNodeServer`
- `KinesisMockPackage` can now either use the `KinesisMockScalaPackageInstaller` (downloads from GitHub) or `KinesisMockNodePackageInstaller` (downloads from NPM)

### Config
- Adds a `KINESIS_MOCK_PROVIDER_ENGINE` flag for switching to Scala build of Kinesis Mock. Valid values are `node` or `scala` -- where empty or invalid values will always default to `node`.
- `KINESIS_MOCK_MAXIMUM_HEAP_SIZE` sets the maximum Java heap size corresponding to the '-Xmx<size>' flag
- `KINESIS_MOCK_INITIAL_HEAP_SIZE` sets the initial Java heap size corresponding to the '-Xms<size>' flag

## Testing
- Creates a `TestKinesisMockScala(TestKinesis)` subclass that runs all tests in `tests/aws/services/kinesis/test_kinesis.py::TestKinesis` but with the configuration value `monkeypatched` to use the Scala build of Kinesis mock instead of NodeJS.

### DIFF
--- a/localstack-core/localstack/services/kinesis/kinesis_mock_server.py
+++ b/localstack-core/localstack/services/kinesis/kinesis_mock_server.py
@@ -2,29 +2,20 @@ import logging
 import os
 import threading
 from abc import abstractmethod
-from enum import StrEnum
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 from localstack import config
-from localstack.services.kinesis.packages import kinesismock_node_package, kinesismock_scala_package
+from localstack.services.kinesis.packages import (
+    KinesisMockEngine,
+    kinesismock_package,
+    kinesismock_scala_package,
+)
 from localstack.utils.common import TMP_THREADS, ShellCommandThread, get_free_tcp_port, mkdir
 from localstack.utils.run import FuncThread
 from localstack.utils.serving import Server
 
 LOG = logging.getLogger(__name__)
-
-
-class KinesisMockEngine(StrEnum):
-    NODE = "node"
-    SCALA = "scala"
-
-    @classmethod
-    def _missing_(cls, value: str | Any) -> str:
-        # default to 'node' if invalid enum
-        if not isinstance(value, str):
-            return cls(cls.NODE)
-        return cls.__members__.get(value.upper(), cls.NODE)
 
 
 class KinesisMockServer(Server):
@@ -229,8 +220,8 @@ class KinesisServerManager:
             )
 
         # Otherwise, install the NodeJS version (default)
-        kinesismock_node_package.install()
-        kinesis_mock_path = Path(kinesismock_node_package.get_installer().get_executable_path())
+        kinesismock_package.install()
+        kinesis_mock_path = Path(kinesismock_package.get_installer().get_executable_path())
 
         return KinesisMockNodeServer(
             port=port,
--- a/localstack-core/localstack/services/kinesis/kinesis_mock_server.py
+++ b/localstack-core/localstack/services/kinesis/kinesis_mock_server.py
@@ -2,16 +2,29 @@ import logging
 import os
 import threading
 from abc import abstractmethod
+from enum import StrEnum
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from localstack import config
-from localstack.services.kinesis.packages import KinesisMockEngine, kinesismock_package
+from localstack.services.kinesis.packages import kinesismock_node_package, kinesismock_scala_package
 from localstack.utils.common import TMP_THREADS, ShellCommandThread, get_free_tcp_port, mkdir
 from localstack.utils.run import FuncThread
 from localstack.utils.serving import Server
 
 LOG = logging.getLogger(__name__)
+
+
+class KinesisMockEngine(StrEnum):
+    NODE = "node"
+    SCALA = "scala"
+
+    @classmethod
+    def _missing_(cls, value: str | Any) -> str:
+        # default to 'node' if invalid enum
+        if not isinstance(value, str):
+            return cls(cls.NODE)
+        return cls.__members__.get(value.upper(), cls.NODE)
 
 
 class KinesisMockServer(Server):
@@ -109,7 +122,7 @@ class KinesisMockScalaServer(KinesisMockServer):
     @property
     def _environment_variables(self) -> Dict:
         default_env_vars = super()._environment_variables
-        kinesis_mock_installer = kinesismock_package.get_installer()
+        kinesis_mock_installer = kinesismock_scala_package.get_installer()
         return {
             **default_env_vars,
             **kinesis_mock_installer.get_java_env_vars(),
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/kinesis/kinesis_mock_server.py
```diff
@@ -91,13 +86,64 @@ def _create_shell_command(self) -> Tuple[List, Dict]:
...
+    def _get_java_vm_options(self) -> list[str]:
+        return [
+            f"-Xms{config.KINESIS_MOCK_INITIAL_HEAP_SIZE}",
+            f"-Xmx{config.KINESIS_MOCK_MAXIMUM_HEAP_SIZE}",
+            "-XX:+UseG1GC",
+            "-XX:MaxGCPauseMillis=500",
+            "-XX:+UseGCOverheadLimit",
+            "-XX:+ExplicitGCInvokesConcurrent",
+            "-XX:+HeapDumpOnOutOfMemoryError",
+            "-XX:+ExitOnOutOfMemoryError",
```
dfangl, line 127: Do we really need to set all these flags? Do they improve performance in any way? Especially the heap dump seems like something our customers do not want, and G1GC is generally the default GC in all recent java versions.

localstack-core/localstack/services/kinesis/packages.py
```diff
@@ -1,28 +1,77 @@
...
+    def __init__(
+        self,
+        default_version: str = _KINESIS_MOCK_VERSION,
+    ):
         super().__init__(name="Kinesis Mock", default_version=default_version)
 
+    @cached_property
+    def engine(self) -> KinesisMockEngine:
+        return KinesisMockEngine(config.KINESIS_MOCK_PROVIDER_ENGINE)
+
```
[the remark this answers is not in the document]
gregfurman: To determine which Kinesis mock backend to run:

https://github.com/localstack/localstack/blob/4eed69f17a5c41be9b8e86f3ba7d6741c6be2f74/localstack-core/localstack/services/kinesis/kinesis_mock_server.py#L208-L225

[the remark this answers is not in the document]
gregfurman: Yeah sure happy to have 2 packages 👍

localstack-core/localstack/services/kinesis/kinesis_mock_server.py
gregfurman: The `MaxGCPauseMillis` was for performance reasons and the `ExitOnOutOfMemoryError` was because I didn't want the server to stall when encountering an OOM.

Will remove:
- `UseG1GC` since I was unaware it was the default (as per your comment)
- `UseGCOverheadLimit` since it does nothing with `G1GC`
- `HeapDumpOnOutOfMemoryError` since it'll spam the logs quite a bit

localstack-core/localstack/services/kinesis/packages.py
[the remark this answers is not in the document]
gregfurman: @dfangl How should we go about configuring the plugin for the LPM? 

Should we have a seperate `kinesis-mock` and `kinesis-mock-scala` plugin or this implementation be unified to cater for both node + scala builds i.e

```python
@package(name="kinesis-mock")
def kinesismock_package() -> Package:
    from localstack.services.kinesis.packages import kinesismock_package

    return kinesismock_package

@package(name="kinesis-mock-scala")
def kinesismock_scala_package() -> Package:
    from localstack.services.kinesis.packages import kinesismock_scala_package
    
    return kinesismock_scala_package
```

OR

```python
@package(name="kinesis-mock")
def kinesismock_package() -> Package:
    from localstack.services.kinesis.packages import (
        KinesisMockEngine,
        kinesismock_package,
        kinesismock_scala_package,
    )

    if KinesisMockEngine(config.KINESIS_MOCK_PROVIDER_ENGINE) == KinesisMockEngine.SCALA:
        return kinesismock_scala_package

    return kinesismock_package
```

localstack-core/localstack/services/kinesis/packages.py
```diff
@@ -1,28 +1,82 @@
...
+class KinesisMockEngine(StrEnum):
+    NODE = "node"
+    SCALA = "scala"
+
+    @classmethod
+    def _missing_(cls, value: str | Any) -> str:
+        # default to 'node' if invalid enum
+        if not isinstance(value, str):
+            return cls(cls.NODE)
+        return cls.__members__.get(value.upper(), cls.NODE)
```
dfangl, line 22: Nit: Not entirely happy with this fallback logic here - it is effectively unused, as we only ever check if it is scala anyway, and we would silently accept obviously wrong values. Also, if we ever switch the default, we have to do it in multiple places. Nothing to hold back the merge, but I think we could potentially make this easier.

localstack-core/localstack/services/kinesis/plugins.py
```diff
@@ -1,8 +1,16 @@
...
 def kinesismock_package() -> Package:
-    from localstack.services.kinesis.packages import kinesismock_package
+    from localstack.services.kinesis.packages import (
+        KinesisMockEngine,
+        kinesismock_package,
+        kinesismock_scala_package,
+    )
+
+    if KinesisMockEngine(config.KINESIS_MOCK_PROVIDER_ENGINE) == KinesisMockEngine.SCALA:
+        return kinesismock_scala_package
```
dfangl, line 14: Q: These package plugins are mostly used for `lpm`, right? Would it make sense to have two different packages here, instead of relying on the environment variable when calling `lpm`?

dfangl (approved):
LGTM! Way cleaner than before in my opinion :+1:

localstack-core/localstack/services/kinesis/packages.py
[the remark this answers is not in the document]
gregfurman: @dfangl FYI I spoke to Alex re: the above and we went with the second approach ☝️

localstack-core/localstack/services/kinesis/packages.py
gregfurman: Yeah I suppose this is unideal but I wanted to make sure (since we're feature-flagging this strategy) that nothing gets broken if incorrect values are used. Definitely erring on the defensive side.

What have we done in past for situations like this?

localstack-core/localstack/services/kinesis/plugins.py
dfangl: Makes sense, thanks for the clarification! I missed the comment above it seems!

localstack-core/localstack/services/kinesis/packages.py
dfangl: Usually we would fallback, but at least with a warning in the log.

Issue #12551, opened 2025-04-23
bug: ECS Does Not Propagate Container Labels


the-wondersmith:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When a task definition is registered with `dockerLabels` for the container(s), the labels are dropped entirely when the task is launched. FTR, this *also* includes the additional labels that ECS/LocalStack set internally (`cluster-arn` and `task-id`).

### Expected Behavior

The specified labels (and any internally set values) should be passed all the way through from LocalStack to Docker and set on the running container.

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

1) Follow the [official `ECS` guide](https://docs.localstack.cloud/user-guide/aws/ecs/), using the MRR task definition below in place of the one in the guide.

### MRR Task Definition

```json
{
  "family": "someTask",
  "containerDefinitions": [
    {
      "name": "task",
      "image": "alpine",
      "cpu": 10,
      "memory": 10,
      "command": [
        "sh",
        "-c",
        "while true; do sleep 1; done"
      ],
      "essential": true,
      "dockerLabels": {
        "some.label": "some-value"
      },
      "logConfiguration": {
        "logDriver": "awslogs",
        "options": {
          "awslogs-create-group": "true",
          "awslogs-group": "someloggroup",
          "awslogs-stream-prefix": "someprefix",
          "awslogs-region": "us-east-1"
        }
      }
    }
  ],
}
```

2) Verify that the `dockerLabels` weren't dropped in ECS:

```bash
awslocal ecs describe-task-definition --task-definition "$(awslocal ecs list-task-definitions --query 'taskDefinitionArns[0]' --output text)" --query 'taskDefinition.containerDefinitions[0].dockerLabels'
```

3) Verify that the labels are not actually present on the running container:

```bash
docker container inspect '{CONTAINER ID}'
```

### Environment

```markdown
- OS: macOS 14.7.4
- LocalStack:
  LocalStack version: `4.0.3`
  LocalStack Docker image sha: `340ad3af995e44242ba634614588569979c7052b9ffffcdf4d4eb78248b082ce`
  LocalStack build date:
  LocalStack build git hash:
```

### Anything else?

The problem appears to be coming from two distinct places. First, user-provided labels are **_overwritten_** (not merged) in [`SdkDockerClient.create_container`](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/utils/container_utils/docker_sdk_client.py#L732).

[...]

PR #12565, opened 2025-04-29
fix: docker label propagation

the-wondersmith:
## Motivation

See [changes](#changes)

## Changes

PR addresses the public half of the bug surfaced in #12551

## Testing

PR includes additional test that covers updated parse-and-merge logic

### DIFF
--- a/localstack-core/localstack/utils/container_utils/container_client.py
+++ b/localstack-core/localstack/utils/container_utils/container_client.py
@@ -33,6 +33,14 @@ from localstack.utils.files import TMP_FILES, chmod_r, rm_rf, save_file
 from localstack.utils.no_exit_argument_parser import NoExitArgumentParser
 from localstack.utils.strings import short_uid
 
+try:
+    from typing import Self
+except ImportError:
+    from typing import TypeVar
+
+    Self = TypeVar("Self")
+
+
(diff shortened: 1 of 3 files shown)


the-wondersmith:
@alexrashed saw that you added this to the 4.5 milestone, just wanted to check in on what that specifically means for this 🙂

alexrashed:
The milestone was added because we try to match the PRs with upcoming releases before we create a release (in this case I added 4.5 before creating the 4.4 release).
The PR was a bit stale since then because your new test is failing in CI.
Would you mind fixing that? Afterwards @dfangl and @dominikschubert could review the PR. :)

the-wondersmith:
> `{ ... }`. The PR was a bit stale since then because your new test is failing in CI. Would you mind fixing that? `{ ... }`

Went to do this and realized something *else* screwy might be going on. Still 99.999% sure it's ultimately to do with the way the underlying functions are wanting to do in-place update instead of explicit creation, but need to chase the thread a bit just to make double-triple-extra sure. Gonna do a bit more digging before I re-up the PR 🤔

alexrashed:
Okay, thanks for the heads up! In this case I am closing the PR for now. Feel free to re-open it / a new one once you have a clear solution.

PR #12567, opened 2025-04-30
Switch to using `kclpy-ext`

silv-io:
## Motivation

We repeatedly run into issues of kclpy wheels not being able to build inside of our pipelines (in the Docker build). This is because kclpy needs to download JARs from maven and there is no built wheel including those JARs available in PyPI. 

However, we can just create and publish that wheel once ourselves and then use our own published package/wheel instead - therefore not relying on maven anymore.

## Changes

- Switch from `amazon-kclpy` to `kclpy-ext`, which we build and publish in the https://github.com/localstack/amazon-kinesis-client-python repo. 

## Testing

The existing kinesis test suite should handle this sufficiently. After all, this should be the exact same package, just with a pre-built wheel available

<!--
## TODO

What's left to do:

- [ ] Are we ok with using the same JARs throughout the lifecycle of one version? Currently, we theoretically could receive updates to the JARs on every build.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.11.7
+    rev: v0.11.8
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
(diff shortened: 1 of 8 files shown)


dfangl (approved):
Great work! Looks good, and should positively impact our build failures :+1:

PR #12573, opened 2025-05-01
remove usage analytics module 

vittoriopolverino:
## Motivation
The` usage.py` module has been deprecated and is no longer in use. All services and features that previously relied on it have been fully migrated to the new [`metrics.py`](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/utils/analytics/metrics.py) module.

Given that the transition is now complete, we can safely remove the `usage.py` module along with its associated unit tests.


localstack-core/localstack/utils/diagnose.py
```diff
@@ -153,4 +153,4 @@ def get_host_kernel_version() -> str:
 
 
 def get_usage():
-    return usage.aggregate()
+    return MetricRegistry().collect()
```
vittoriopolverino, line 156: **note:** tested by calling the following endpoints
- curl http://localhost:4566/_localstack/usage
- curl http://localhost:4566/_localstack/diagnose

alexrashed (approved):
Nice cleanup! Thanks for driving the migration and taking care of removing the old code afterwards! 💯

PR #12576, opened 2025-05-02
CFn v2: Implement stack deletion

simonrw:
## Motivation

Quite a few resource tests fail because the deletion of a stack is not correctly implemented. The tests assert that resources are deleted, but they aren't.

## Changes

- Improve parity of SSM exceptions when getting non-existent resources
- Implement deletion by creating a "hidden" change set, and performing the execution by passing `None` in for the `after` template and parameters

## Testing

In addition to the included test, run the test `tests/aws/services/cloudformation/resources/test_events.py::test_cfn_event_bus_resource` which fails on the master branch.


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -89,10 +89,9 @@ def visit_node_depends_on(self, node_depends_on: NodeDependsOn) -> PreprocEntity
         if array_identifiers_delta.after:
             depends_on_resource_logical_ids.update(array_identifiers_delta.after)
         for depends_on_resource_logical_id in depends_on_resource_logical_ids:
-            node_resource = self._get_node_resource_for(
-                resource_name=depends_on_resource_logical_id, node_template=self._node_template
-            )
-            self.visit_node_resource(node_resource)
+            for node_resource in self._node_template.resources.resources:
```
MEPalma: Why are we avoiding `_get_node_resource_for`? It looks like the user can define `depends_on` values that don't match a resource in the template

MEPalma (approved):
Thanks! This is definitely one of the key pieces for the v2 engine; great to see so many test cases unblocked now.

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
simonrw: I removed this function because it visits the resource, and was triggering a double deployment. Perhaps this is more a failure of the node caching aspect? Visiting a node that has already been visited should be a no-op right?

simonrw: After discussion with @MEPalma I replaced this removal of `_get_node_resource_for` with the original function, but using `self.visit` instead of `self.visit_node_resource` as `self.visit` implements the expected caching behaviour to prevent double-deployment of resources.

PR #12581, opened 2025-05-05
Step Functions: Add Telemetry for SFN_MOCK_CONFIG Usage

MEPalma:
## Motivation
The mocked service integrations feature for Step Functions testing was recently introduced, but no telemetry was added to track its usage. These changes address that gap by recording usage of the associated environment variable, `SFN_MOCK_CONFIG`, without capturing the file path it points to.

## Changes
- added `SFN_MOCK_CONFIG` to analytics' `PRESENCE_ENV_VAR` list


localstack-core/localstack/services/stepfunctions/usage.py
```diff
@@ -10,3 +10,8 @@
     name="language_features_used",
     labels=["query_language", "uses_variables"],
 )
+
+# Initialize a counter to record the use of each execution type.
+execution_type_counter = Counter(
+    namespace="stepfunctions", name="execution_type", labels=["is_mock_test_case"]
```
joe4dev, line 16: naming: Isn't `execution` sufficient because the `_type` sounds like an aspect that a label should cover. See "Metric & Label Best Practices" in Notion.

localstack-core/localstack/services/stepfunctions/provider.py
```diff
@@ -789,6 +790,10 @@ def start_execution(
             state_machine_arn_parts[1] if len(state_machine_arn_parts) == 2 else None
         )
 
+        # Count metrics about the execution type.
+        is_mock_test_case: bool = mock_test_case_name is not None
+        UsageMetrics.execution_type_counter.labels(is_mock_test_case=is_mock_test_case).increment()
```
joe4dev, line 795: question: Are we missing other entry points, such as `start_sync_execution`? Are they worth tracking?

localstack-core/localstack/services/stepfunctions/usage.py
joe4dev: suggestion/idea: Following the Lambda example in `localstack-core/localstack/services/lambda_/analytics.py:function_counter`, what about the following focusing on the main operation(s) rather than having feature-specific counters (e.g., JSONata, is_mock_test_case, workflow_type, invocation_type, ...):

```py
execution_counter = Counter(
    namespace="stepfunctions"
    name="execution",
    labels=[
        "is_mock_test_case",
        "workflow_type",  # standard | express
        "invocation_type",  # async | sync
    ],
)
```

Is it worthwhile (i.e., can we derive something actionable) from capturing more metadata linked to executions?
In Lambda, we even use an `operation` field to capture create and invoke activities separately. Feel free to weigh in on what's the best model for StepFunctions here.

sidenote: The dimension `status` might be interesting to learn about unexpected/unhandled errors (i.e., try/catch), the more metadata we collect about executions. However, if all executions happen through the API, unhandled exceptions should be tracked generically by our ASF-based telemetry (including actionable stack traces in DEBUG mode).

joe4dev: What is your recommendation/guidance from the data side @vittoriopolverino, especially regarding the trade-off: counters for features/parameters (e.g., JSONata, is_mock_test_case, workflow_type, invocation_type) vs. for operations (e.g., execute, create) with features as labels?

joe4dev (commented):
I added some questions and discussion items regarding the optimal place of instrumentation and the trade-off regarding feature/parameter vs. operation tracking.

MEPalma:
Thanks @joe4dev  for sharing your thoughts! 
The objective here was to quickly introduce a basic mechanism to track usage of this new feature in time for the upcoming release. Hence, optimality of this solution was not a goal and aimed to minimize changes ahead of the release. That said, I agree that implementing a more robust and optimal tracking strategy is worth exploring whilst introducing a new counter. 
For now, I’m opening a separate PR that only tracks the use of the environment variable, and I’ll continue the broader discussion in this thread.

MEPalma:
https://github.com/localstack/localstack/pull/12584

PR #12586, opened 2025-05-06
APIGW: fix binaryMediaTypes when importing/updating REST APIs

bentsku:
## Motivation
We've recently received a report that binary support for APIGW wasn't working when deploying a REST API through CloudFormation and using the `Body` attribute to use an OpenAPI spec file to define the API. 

The OpenAPI spec file didn't define binary media types via the `x-amazon-apigateway-binary-media-types`, but it did with the [CloudFormation parameter `BinaryMediaTypes`](https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-binarymediatypes).

After testing against AWS, it seems the behavior is defined by the `merge` mode when calling `PutRestApi` for an existing API, and it will be merged by default with the custom extension when using CloudFormation. 

## Changes
- update support for binary media types when used with the CFN parameter and the `Body` parameter
- update test cases to recreate the scenario (both CFN parameter + body)
- add test case around `PutRestApi` depending on the merge mode
- refresh tests using the same OpenAPI spec file
- update export logic to return the binary media types extension

### DIFF
--- a/localstack-core/localstack/services/apigateway/helpers.py
+++ b/localstack-core/localstack/services/apigateway/helpers.py
@@ -951,11 +951,11 @@ def import_api_from_openapi_spec(
 
     # binary types
     if mode == "merge":
-        existing_modes = rest_api.binaryMediaTypes or []
+        existing_binary_media_types = rest_api.binaryMediaTypes or []
     else:
-        existing_modes = []
+        existing_binary_media_types = []
 
-    rest_api.binaryMediaTypes = existing_modes + resolved_schema.get(
+    rest_api.binaryMediaTypes = existing_binary_media_types + resolved_schema.get(
         OpenAPIExt.BINARY_MEDIA_TYPES, []
     )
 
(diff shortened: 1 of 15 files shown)


localstack-core/localstack/services/apigateway/helpers.py
```diff
@@ -948,7 +950,14 @@ def create_method_resource(child, method, method_schema):
         get_or_create_path(base_path + path, base_path=base_path)
 
     # binary types
-    rest_api.binaryMediaTypes = resolved_schema.get(OpenAPIExt.BINARY_MEDIA_TYPES, [])
+    if mode == "merge":
+        existing_modes = rest_api.binaryMediaTypes or []
```
cloutierMat, line 954: nit: The name `existing_modes` is confusing me here. Would we be better to use something like `existing_binary_media_types`?

cloutierMat (approved):
Thanks for the addition and the super thorough testing and investigation of behaviours for the put modes! 
:rocket:

localstack-core/localstack/services/apigateway/helpers.py
bentsku: oh yeah thanks a lot, fixed that 🤦‍♂️

PR #12588, opened 2025-05-07
Transcribe: Enable MyPy

bblommers:
## Motivation
Part of a broader push to enforce typing support everywhere (see #12532). One of the areas that I wanted to focus next on was adding correct types to the `BaseStore`/`AccountRegionBundle`/etc, because it is such a fundamental concept that all emulators would benefit from improved type support.

However, before improving types there, it makes sense to first ensure that the usage of `BaseStore` is typed and validated. Only then can we be improve the types in `BaseStore` and be sure they are correct.

Hence this PR: enforcing type checks in one of the smaller services, for the sole purpose of checking that `BaseStore` is used/typed correctly.

## History
- Optional API arguments (`param: X | None = None`) were added in #12614 
- Proper types for `context.account_id`/`region_name` were set in #12617 

An earlier version of this PR tackled everything at once, but with the above PR's already in place, this PR has been simplified to only change the Transcribe-specific types.

### DIFF
--- a/localstack-core/localstack/services/transcribe/provider.py
+++ b/localstack-core/localstack/services/transcribe/provider.py
@@ -5,7 +5,7 @@ import threading
 import wave
 from functools import cache
 from pathlib import Path
-from typing import Tuple
+from typing import Any, Tuple
 from zipfile import ZipFile
 
 from localstack import config
@@ -102,16 +102,16 @@ _DL_LOCK = threading.Lock()
 
 class TranscribeProvider(TranscribeApi):
     def get_transcription_job(
-        self, context: RequestContext, transcription_job_name: TranscriptionJobName, **kwargs
+        self, context: RequestContext, transcription_job_name: TranscriptionJobName, **kwargs: Any
     ) -> GetTranscriptionJobResponse:
         store = transcribe_stores[context.account_id][context.region]
 
         if job := store.transcription_jobs.get(transcription_job_name):
             # fetch output key and output bucket
             output_bucket, output_key = get_bucket_and_key_from_presign_url(
-                job["Transcript"]["TranscriptFileUri"]
+                job["Transcript"]["TranscriptFileUri"]  # type: ignore[index,arg-type]
             )
-            job["Transcript"]["TranscriptFileUri"] = connect_to().s3.generate_presigned_url(
+            job["Transcript"]["TranscriptFileUri"] = connect_to().s3.generate_presigned_url(  # type: ignore[index]
                 "get_object",
                 Params={"Bucket": output_bucket, "Key": output_key},
                 ExpiresIn=60 * 15,
@@ -128,13 +128,13 @@ class TranscribeProvider(TranscribeApi):
         # Install and configure vosk
         vosk_package.install()
 
-        from vosk import SetLogLevel  # noqa
+        from vosk import SetLogLevel  # type: ignore[import-not-found]  # noqa
 
         # Suppress Vosk logging
         SetLogLevel(-1)
 
     @handler("StartTranscriptionJob", expand=False)
-    def start_transcription_job(
+    def start_transcription_job(  # type: ignore[override]
         self,
         context: RequestContext,
         request: StartTranscriptionJobRequest,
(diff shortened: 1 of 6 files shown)


alexrashed:
@sannya-singal As a service owner for transcribe: What do you think? I personally are very much in favor or increasing the enforcement of type hints and checks, but it's really you as a service owner who has to decide if you want to introduce this for your service. 😃
@bblommers kudos for driving this effort, onboarding the first AWS service is a huge step towards starting a "mypy pandemic" in the good way / i.e. making typing enforcement infectious 😄

sannya-singal (commented):
Awesome work 🚀 Enabling MyPy for the Transcribe module is a great step towards our broader goals of strengthening type checks across the codebase, and I really like the incremental approach you are taking here 🎉

Kudos as well for integrating this with the first AWS service 🙌  Having type checking enabled at the service level will also empower code owners to enforce typing in their codebase, also making it easier to catch issues early and improve maintainability. 

Looks great from my side 🚀 thanks for pushing this forward! 👍

sannya-singal (approved):

tests/unit/aws/api/test_asf_providers.py
```diff
@@ -11,4 +13,8 @@
     collect_implemented_provider_operations(),
 )
 def test_provider_signatures(sub_class: type, base_class: type, method_name: str):
+    if "TranscribeApi" in str(base_class):
+        raise SkipTest(
+            "Transcribe Signatures are different on purpose, as optional args are marked as such"
+        )
```
alexrashed, line 19: **suggestion:** Unfortunately, this one is problematic. This test makes sure that the provider implementations don't drift from the upstream API / their superclasses.
While we did reduce the coupling between the generated API and the providers implementation with adding the `kwargs` everywhere, we were still monitoring and updating the signatures based on this tests' results.
An example for an ASF Update PR with changed providers would be https://github.com/localstack/localstack/pull/12490.

The root of this problem is actually that we do generate the code with wrong typing. In this case I think we should just update the code generation and all its subclasses. It might even make sense to implement a script (with something like `libcst`) that does this automatically to help us keep our implementations up-to-date automatically (as part of the ASF Update action directly).

What do you think?

alexrashed (changes requested):
Thanks again for this push towards better and enforced typing! 💯 
I just stumbled upon the exclusion of the updated provider from the unit test which worries me a bit. I think it would be great to quickly discuss how we want to move forward here before we merge this PR.

localstack-core/localstack/services/transcribe/provider.py
```diff
@@ -192,13 +192,13 @@ def start_transcription_job(
     def list_transcription_jobs(
         self,
         context: RequestContext,
-        status: TranscriptionJobStatus = None,
-        job_name_contains: TranscriptionJobName = None,
-        next_token: NextToken = None,
-        max_results: MaxResults = None,
-        **kwargs,
+        status: Optional[TranscriptionJobStatus] = None,
```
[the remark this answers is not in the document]
bblommers: I explicitly used `Optional` here because it's not supported in Python 3.9, so if the CLI ever explicitly loads this module in Python 3.9 it would fall over.

I don't know how likely that is though - I'm happy to do either way :shrug:

tests/unit/aws/api/test_asf_providers.py
bblommers: Updating the generated code makes a lot of sense IMO. Considering the scope of changing signatures over the entire codebase, we could split it up in three parts:

1. Change this test to view `X=None` and `X|None = None` as equal, as a stop gap measure while we have both versions in place
2. Change the ASF update action to update all the generated code to `X | None`
3. Change the providers themselves using `libcst`

We could do all of this in one PR, but that would be massive, and very difficult to review.

Any objections to the stepwise approach @alexrashed?

(I'll mark this PR as draft for now, until we have a solution.)

alexrashed: This is perfect. let's move forwards with (1) here and then we can tackle the other two in follow ups! 💯

sannya-singal (approved):
LGTM 🚀 Thanks for incorporating the feedback! I will defer to @alexrashed for a re-review since he requested the changes.

alexrashed (approved):
Thanks a lot for addressing the comments, and sorry this was blocked for so long!
I guess a rebase would be good before merging this to make sure that the pipeline is still green / the new checks don't fail. Thanks for the push towards more type enforcement! :100:

PR #12590, opened 2025-05-07
Docker registry customisation

simonrw:
## Motivation

Some customers work in airgapped environments, and require customisation of the Docker image registries used. This PR enables a flexible method for adding customisation to the Docker image pulling strategy.

## Changes

* Introduce the `RegistryResolverStrategy` protocol to support custom image resolving
* Add an implementation `HardCodedResolver` which is a no-op
* Use this resolver when pulling images, or creating containers


dfangl (approved):
Good and extensible approach :+1:

PR #12591, opened 2025-05-07
Defer CDK imports in infra provisioning

dfangl:
## Motivation
Currently, we import the `localstack.testing.scenario.provisioning` module on every pytest run, which in turns imports the aws_cdk python bindings.

This leads to a heavy increase in startup time, especially when debugging (tested in Pycharm).

By deferring the import to the first fixture usage, instead of on all imports, that increased delay will only happen for tests actually requesting the fixture, instead of all tests.

For unrelated tests, this can reduce the start time of the test (on my system):
* When debugging, from 46 to 7 seconds
* Without debugging, from 7 to 1 seconds

CI runtime should only marginally be affected, only if test selection selects tests without a single usage of the infra provisioner.

I also fixed the mypy precommit errors happening on my system, by ignoring the plux import completely, instead of just ignoring import-untyped errors.

## Changes
* Defer infra provisioner imports to actual fixture usages
* Ignore plux type errors completely to prevent mypy precommit errors

### DIFF
--- a/localstack-core/localstack/packages/api.py
+++ b/localstack-core/localstack/packages/api.py
@@ -8,7 +8,7 @@ from inspect import getmodule
 from threading import RLock
 from typing import Any, Callable, Generic, List, Optional, ParamSpec, TypeVar
 
-from plux import Plugin, PluginManager, PluginSpec  # type: ignore[import-untyped]
+from plux import Plugin, PluginManager, PluginSpec  # type: ignore
 
 from localstack import config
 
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Great catch! Thanks a lot for jumping on this! Absolutely great to fix this issue, speeding up our own development cycles / improving our own DevX! 💯 🦸🏽 
I added @bblommers as reviewer to quickly give his sentiment on the mypy change / issue.

bblommers (approved):

bblommers:
LGTM! As long as `plux` doesn't have any types, the specific type of ignore doesn't really matter IMO

PR #12592, opened 2025-05-07
Fix provisioned concurrency set on Lambda alias

dfangl:
## Motivation
Currently, when setting provisioned concurrency on a lambda alias pointing to a version, and then invoking it, there will still be another container spawned up on demand.

This happens due to an error in the lookup of the provisioned concurrency config in the counting service, which issues leases for environments.

## Changes
* Provisioned concurrency environments of a config set on an alias should now work as expected

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/counting_service.py
+++ b/localstack-core/localstack/services/lambda_/invocation/counting_service.py
@@ -157,11 +157,14 @@ class CountingService:
                     function_version.id.qualifier
                 )
                 if not provisioned_concurrency_config:
+                    # check if any aliases point to the current version, and check the provisioned concurrency config
+                    # for them. There can be only one config for a version, not matter if defined on the alias or version itself.
                     for alias in function.aliases.values():
                         if alias.function_version == function_version.id.qualifier:
                             provisioned_concurrency_config = (
                                 function.provisioned_concurrency_configs.get(alias.name)
                             )
+                            break
                 if provisioned_concurrency_config:
                     available_provisioned_concurrency = (
                         provisioned_concurrency_config.provisioned_concurrent_executions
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/lambda_/invocation/counting_service.py
```diff
@@ -156,6 +156,12 @@ def get_invocation_lease(
                 provisioned_concurrency_config = function.provisioned_concurrency_configs.get(
                     function_version.id.qualifier
                 )
```
joe4dev, line 158: docs: Would a docstring help to clarify our findings?

```
# Look up a potential provisioned concurrency config for aliases because a function version shares provisioned concurrency with aliases. However, the config only applies to a single function version or alias on the CRUD level.
```

joe4dev (approved):
Thank you for the fix @dfangl 👏👏👏 
Good catch 🕵️

gregfurman (approved):
LGTM 👍

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -2539,6 +2539,59 @@ def test_provisioned_concurrency(self, create_lambda_function, snapshot, aws_cli
         result2 = json.load(invoke_result2["Payload"])
         assert result2 == "on-demand"
 
+    @markers.aws.validated
+    def test_provisioned_concurrency_on_alias(self, create_lambda_function, snapshot, aws_client):
```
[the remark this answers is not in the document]
dfangl: There will be an additional container brought up for the last invocation, but in general, confirming that we hit the right invocation type environment should be sufficient, otherwise we cannot validate it against AWS as well.

localstack-core/localstack/services/lambda_/invocation/counting_service.py
dfangl: I added some comments on it!

PR #12594, opened 2025-05-07
Add CloudFormation Lambda Version Provisioned Concurrency

joe4dev:
Depends on https://github.com/localstack/localstack/pull/12592

## Motivation

Testing for a support issue revealed an auxiliary issue that provisioned concurrency is not implemented for the CloudFormation resource [AWS::Lambda::Version](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-version.html).

## Changes

* Add support for provisioned concurrency config in the Lambda version CloudFormation resource provider
* Add aws-validated test case `test_lambda_version_provisioned_concurrency`. I opted for creating a separate test case for provisioned concurrency to ensure both paths work independently, especially given that provisioned concurrency includes different waiters (takes much longer). Feel free to suggest if combining them would be better.
* Improve test case `test_lambda_version` to validate actual behavior for provisioned concurrency (fixed in https://github.com/localstack/localstack/pull/12592)

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_lambda.py
+++ b/tests/aws/services/cloudformation/resources/test_lambda.py
@@ -239,6 +239,12 @@ def test_lambda_alias(deploy_cfn_template, snapshot, aws_client):
         parameters={"FunctionName": function_name, "AliasName": alias_name},
     )
 
+    invoke_result = aws_client.lambda_.invoke(
+        FunctionName=function_name, Qualifier=alias_name, Payload=b"{}"
+    )
+    assert "FunctionError" not in invoke_result
+    snapshot.match("invoke_result", invoke_result)
+
     role_arn = aws_client.lambda_.get_function(FunctionName=function_name)["Configuration"]["Role"]
     snapshot.add_transformer(
         snapshot.transform.regex(role_arn.partition("role/")[-1], "<role-name>"), priority=-1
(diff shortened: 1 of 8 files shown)


tests/aws/services/cloudformation/resources/test_lambda.py
```diff
@@ -239,6 +239,12 @@ def test_lambda_alias(deploy_cfn_template, snapshot, aws_client):
         parameters={"FunctionName": function_name, "AliasName": alias_name},
     )
 
+    invoke_result = aws_client.lambda_.invoke(
+        FunctionName=function_name, Qualifier=alias_name, Payload=b"{}"
+    )
+    assert "FunctionError" not in invoke_result
```
simonrw, line 245: q: do we still need this assertion of we are snapshotting the result?

simonrw (approved):
The change looks sensible to me. Just a question about an assertion but lgtm!

tests/aws/services/cloudformation/resources/test_lambda.py
joe4dev: The assertion is not strictly when it comes to AWS<->LS parity. However, we commonly clarify the intention for Lambda invokes (see Lambda test suite) because it's too/so easy to overlook a snapshot update, which contains a failing Lambda response (happened twice here).

PR #12598, opened 2025-05-08
fix put-metric-alarm test failure rate

pinzon:
## Motivation
As discussed internally, the test_put_metric_alarm test currently has a failure rate of around 14%, which makes it frustrating during full test suite runs. The root cause appears to be that the alarm-triggering value can vary in LocalStack—sometimes evaluating to 21 instead of the expected average 21.5 (which is consistently correct on AWS). This is likely due to timing differences or slight inconsistencies in how LocalStack handles metric ingestion and evaluation windows.

## Changes
- Increase the threshold of the CW Alarm to make the Alarm triggering value consistent between LS and AWS.

## Testing
- updated snapshots


tests/aws/services/cloudwatch/test_cloudwatch.py
```diff
@@ -1338,8 +1338,8 @@ def test_put_metric_alarm(
         retry(
             _sqs_messages_snapshot,
             retries=60,
-            sleep=3 if is_aws_cloud() else 1,
-            sleep_before=5 if is_aws_cloud() else 0,
+            sleep=3,
```
tiurin: Here retries are applied to receiving messages from SQS. Increasing sleep won't change the resulting alarm. Currently alarm has period of 30 seconds, 60 retries wit sleep 1 covers it more than enough. Locally test succeeds exactly after 30 seconds.

tiurin: In fact, this made me think that we can reduce period to a minimum value of 10 and therefore shave 20 seconds off test execution!

tiurin: I've pushed changes to the branch - reverted sleeps and reduced period to 10. Test now consistently executes successfully locally in just over 10 seconds. If you're happy with those changes please merge the PR! Otherwise happy to revert if there is any consideration.

pinzon:
The changing of the period is a better change than the edition of  retry. Thanks @tiurin. :+1:

PR #12601, opened 2025-05-09
CFn v2: better handle deploy errors

simonrw:
## Motivation

Currently when a deploy error occurs, the stack does not get put into an error state, so the deploy process loops until it times out. If an exception occurs during the deploy process we should stop the deployment ASAP and mark it as failed so that our development and investigation can continue quicker, and this better reflects AWS state.

## Changes

Catch an exception from the deploy process and mark the stack as failed.

## Testing

Uncomment a skipped test from the `tests/aws/services/cloudformation/v2` directory that fails _during deployment time_, and marvel at how quickly an error is thrown, whereas before it would take the wait time of the stack deployment to fail.


pinzon (commented):
:partying_face:

PR #12606, opened 2025-05-12
Autogenerate templates

simonrw:
## Motivation

When testing CFn deployments, we need valid templates. These are sometimes annoying to write. 

We have access to the [specifications for all resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html), can we generate valid templates?

Moreover, can we generate _permutations_ of templates so we can test our update process in the V2 engine?

## Changes

TBD


simonrw:
This work was a poc. I'll close this as it's not for merging

PR #12609, opened 2025-05-12
Added pagination and filtering for s3 list buckets operation

### LINKED ISSUE #12585: bug: aws java sdk v1 seems to not be able to retrieve region from localstack S3 bucket

- [x] I have searched the existing issues

### Current Behavior

Hello,

I'm writing this issue to ask you about about a problem i encounter when using localstack bucket.
I have a scala script i want to use on AWS later, and I need to read a file from a S3 bucket. In my local local version i'm using a localstack container. 

However, when running my code, I get an error telling me that the bucket does not exist.

Any ideas about whay I could have done wrong with localstack for that to happen ?

### Expected Behavior

I expect the listBuckets method of AmazonS3 to return a Bucket of region that is not null

With a docker-compose file

### Steps To Reproduce

# How I create the localstack elems:

I run a container from image `localstack/localstack:latest`, with the following env variables : 
```
AWS_DEFAULT_REGION: eu-west-1
AWS_ACCESS_KEY_ID: admin
AWS_SECRET_ACCESS_KEY: password
```
Then, once the container is running, i create a bucket with the following command :
`awslocal s3 mb s3://my-bucket --region eu-west-1`

# How I (try to) access my bucket

In my scala script, I create a AmazonS3 client the following way :

```
private val s3: AmazonS3 = AmazonS3ClientBuilder
    .standard()
    .withEndpointConfiguration(
      new EndpointConfiguration(endpoint, region)
    )
    .withCredentials(
      new AWSStaticCredentialsProvider(
        new BasicAWSCredentials(accessKeyId, secretAccessKey)
      )
    )
    .build()

// endpoint is "http://localhost:4566"
// region is "eu-west-1"
// accessKeyId is "admin"
// secretAccessKey is "password"
```

Then, when run s3.listBuckets() i have the following line : 
[S3Bucket [name=my-bucket, creationDate=Tue May 06 11:22:21 CEST 2025, owner=S3Owner [name=webfile,id=75aa57f09aa0c8caeab4f8c24e99d10f8e7faeebf76c078efc7c6caea54ba06a], region=null]]

Moreover, running s3.getObject("my-bucket","/path/to/file") tells me that NoSuchBucket

### Environment

```markdown
# Versions

aws-java-sdk-s3 : 1.12.782 (can't upgrade to sdkv2 because bounded by scala version)
localstack image : latest
awslocal cli : latest, installed yesterday (May 5th 2025)
```

### Anything else?

_No response_

bryansan-local:
## Motivation
Right now, the list buckets operation is ignoring the `Prefix`, `MaxBuckets`, `BucketRegion` and `ContinuationToken` which results in the operation always returning all buckets.

Closes #12585

## Changes
- Process correctly `Prefix`, `MaxBuckets`, `BucketRegion` and `ContinuationToken` to filter and paginate accordingly.
- Added integration tests + parity validations for these scenarios

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -617,10 +617,6 @@ class S3Provider(S3Api, ServiceLifecycleHook):
             buckets.append(output_bucket)
             count += 1
 
-        # Weird behavior of AWS that we need to reproduce to keep parity
-        if len(buckets) == 0 and not prefix:
-            del owner["DisplayName"]
-
         return ListBucketsOutput(
             Owner=owner, Buckets=buckets, Prefix=prefix, ContinuationToken=next_continuation_token
         )
--- a/tests/aws/services/s3/test_s3_list_operations.py
+++ b/tests/aws/services/s3/test_s3_list_operations.py
@@ -101,6 +101,10 @@ class TestS3ListBuckets:
         snapshot.match("list-objects-with-empty-continuation-token", response)
 
     @markers.aws.validated
+    # In some regions AWS returns the Owner display name (us-west-2) but in some it doesnt (eu-central-1)
+    @markers.snapshot.skip_snapshot_verify(
+        paths=["$.list-objects-by-bucket-region-empty..Owner.DisplayName"]
+    )
     def test_list_buckets_by_bucket_region(
         self, s3_create_bucket, s3_create_bucket_with_client, aws_client_factory, snapshot
     ):
@@ -109,22 +113,26 @@ class TestS3ListBuckets:
         snapshot.add_transformer(snapshot.transform.regex(region_us_west_2, "<region>"))
 
         s3_create_bucket()
+        client_us_west_2 = aws_client_factory(region_name=region_us_west_2).s3
+
+        bucket_name = f"test-bucket-{short_uid()}"
+        s3_create_bucket_with_client(
+            client_us_west_2,
+            Bucket=bucket_name,
+            CreateBucketConfiguration={"LocationConstraint": region_us_west_2},
+        )
 
         region_eu_central_1 = "eu-central-1"
         client_eu_central_1 = aws_client_factory(region_name=region_eu_central_1).s3
-        response = client_eu_central_1.list_buckets(BucketRegion=region_eu_central_1, MaxBuckets=1)
+        response = client_eu_central_1.list_buckets(
+            BucketRegion=region_eu_central_1, Prefix=bucket_name
+        )
         assert len(response["Buckets"]) == 0
 
         snapshot.match("list-objects-by-bucket-region-empty", response)
 
-        client_us_west_2 = aws_client_factory(region_name=region_us_west_2).s3
-        s3_create_bucket_with_client(
-            client_us_west_2,
-            Bucket=f"test-bucket-{short_uid()}",
-            CreateBucketConfiguration={"LocationConstraint": region_us_west_2},
-        )
-
-        response = client_us_west_2.list_buckets(BucketRegion=region_us_west_2, MaxBuckets=1)
+        response = client_us_west_2.list_buckets(BucketRegion=region_us_west_2, Prefix=bucket_name)
+        assert len(response["Buckets"]) == 1
 
         returned_bucket = response["Buckets"][0]
         assert returned_bucket["BucketRegion"] == region_us_west_2
--- a/tests/aws/services/s3/test_s3_list_operations.snapshot.json
+++ b/tests/aws/services/s3/test_s3_list_operations.snapshot.json
@@ -3087,7 +3087,7 @@
     }
   },
   "tests/aws/services/s3/test_s3_list_operations.py::TestS3ListBuckets::test_list_buckets_with_max_buckets": {
-    "recorded-date": "13-05-2025, 17:26:52",
+    "recorded-date": "14-05-2025, 09:10:49",
     "recorded-content": {
       "list-objects-with-max-buckets": {
         "Buckets": [
@@ -3110,7 +3110,7 @@
     }
   },
   "tests/aws/services/s3/test_s3_list_operations.py::TestS3ListBuckets::test_list_buckets_with_continuation_token": {
-    "recorded-date": "13-05-2025, 17:27:02",
+    "recorded-date": "14-05-2025, 09:10:59",
     "recorded-content": {
       "list-objects-with-continuation": {
         "Buckets": [
@@ -3133,7 +3133,7 @@
     }
   },
   "tests/aws/services/s3/test_s3_list_operations.py::TestS3ListBuckets::test_list_buckets_when_continuation_token_is_empty": {
-    "recorded-date": "13-05-2025, 17:26:55",
+    "recorded-date": "14-05-2025, 09:10:50",
     "recorded-content": {
       "list-objects-with-empty-continuation-token": {
         "Buckets": [
@@ -3156,7 +3156,7 @@
     }
   },
   "tests/aws/services/s3/test_s3_list_operations.py::TestS3ListBuckets::test_list_buckets_by_prefix_with_case_sensitivity": {
-    "recorded-date": "13-05-2025, 17:26:49",
+    "recorded-date": "14-05-2025, 09:10:46",
     "recorded-content": {
       "list-objects-by-prefix-empty": {
         "Buckets": [],
@@ -3191,13 +3191,14 @@
     }
   },
   "tests/aws/services/s3/test_s3_list_operations.py::TestS3ListBuckets::test_list_buckets_by_bucket_region": {
-    "recorded-date": "13-05-2025, 17:26:59",
+    "recorded-date": "14-05-2025, 09:10:54",
     "recorded-content": {
       "list-objects-by-bucket-region-empty": {
         "Buckets": [],
         "Owner": {
           "ID": "<owner-id>"
         },
+        "Prefix": "<bucket-name:1>",
         "ResponseMetadata": {
           "HTTPHeaders": {},
           "HTTPStatusCode": 200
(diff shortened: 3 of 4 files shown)


tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +48,87 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
+        snapshot.match("list-objects-with-max-buckets", response)
+
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..ContinuationToken"])
+    def test_list_buckets_by_bucket_region(
+        self, s3_create_bucket, s3_create_bucket_with_client, aws_client_factory, snapshot
+    ):
+        region_us_west_2 = "us-west-2"
+        snapshot.add_transformer(snapshot.transform.s3_api())
+        snapshot.add_transformer(RegexTransformer(region_us_west_2, "<region>"))
```
bentsku, line 93: nit: you can avoid the import of the `RegexTransformer` class by using `snapshot.transform.regex(param1, param2)` instead

```suggestion
        snapshot.add_transformer(snapshot.transform.regex(region_us_west_2, "<region>"))
```

tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +48,87 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
+        response = aws_client.s3.list_buckets(Prefix=bucket_name)
+        assert len(response["Buckets"]) == 1
+
+        returned_bucket = response["Buckets"][0]
+        assert returned_bucket["Name"] == bucket_name
+
+        snapshot.match("list-objects-by-prefix", response)
+
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..ContinuationToken"])
```
bentsku, line 74: for values that will change every time you run the test against AWS, you will need to transform the value instead of ignoring it. The best practice when writing and running tests against AWS is to be able to re-run the test against AWS and for it to pass without updating the snapshot. 

Here, you would need to add this line, for example: `snapshot.add_transformer(snapshot.transform.key_value("ContinuationToken"))`

tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +48,87 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
+            Bucket=f"test-bucket-{short_uid()}",
+            CreateBucketConfiguration={"LocationConstraint": region_us_west_2},
+        )
+
+        response = client_us_west_2.list_buckets(BucketRegion=region_us_west_2)
+
+        returned_bucket = response["Buckets"][0]
+        assert returned_bucket["BucketRegion"] == region_us_west_2
+
+        snapshot.match("list-objects-by-bucket-region", response)
```
bentsku, line 109: nit: might be good to add a case where we try listing buckets in another region, and for it to return an empty result?

tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +48,87 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
+
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..ContinuationToken"])
+    def test_list_buckets_with_continuation_token(self, s3_create_bucket, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.s3_api())
+
+        s3_create_bucket()
+        s3_create_bucket()
+        response = aws_client.s3.list_buckets(MaxBuckets=1)
+        assert "ContinuationToken" in response
```
bentsku, line 119: nit: here might be a good candidate for a snapshot test, it would assert the presence of the field without needing the manual `assert`? but this also makes the intent of the test clear, so this is not blocking 😄

tests/aws/services/s3/test_s3_list_operations.snapshot.json
```diff
@@ -3085,5 +3085,95 @@
...
+    "recorded-content": {
+      "list-objects-with-max-buckets": {
+        "Buckets": [
+          {
+            "BucketRegion": "<region>",
+            "CreationDate": "datetime",
+            "Name": "<bucket-name:1>"
+          }
+        ],
+        "ContinuationToken": "19q8mEPF1r8_Ra_NOYw.O189_XK05XnbMmjtCCZLBl0eWIkOP77iU8.A3RGmSDt2L9aUSBfiqSgPNGQAlwpl7zQ--",
```
bentsku, line 3100: as written in another comment, this needs to be transformed via a transformer

localstack-core/localstack/services/s3/provider.py
```diff
@@ -581,14 +581,51 @@ def list_buckets(
         bucket_region: BucketRegion = None,
         **kwargs,
     ) -> ListBucketsOutput:
-        # TODO add support for max_buckets, continuation_token, prefix, and bucket_region
+        if continuation_token == "":
+            raise InvalidArgument(
+                "The continuation token provided is incorrect",
+                ArgumentName="continuation-token",
+            )
```
bentsku, line 588: it seems we do not have a snapshot test covering this, it would be great to add a snapshot validated test for it

tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +48,87 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
 
 
+class TestS3ListBuckets:
+    @markers.aws.validated
+    def test_list_buckets_by_prefix_with_case_sensitivity(
+        self, s3_create_bucket, aws_client, snapshot
+    ):
+        snapshot.add_transformer(snapshot.transform.s3_api())
+
+        bucket_name = f"test.bucket.{short_uid()}"
```
bentsku, line 58: small nit: dots `.` are not well supported in S3, because it creates URL that have subdomains, and do not have SSL certs covering them. Might be better to use `-` instead.

bentsku (commented):
Awesome, great work, this is a nice addition! 

I have a few comments, one blocking regarding the snapshot transformations of fields that are changing between every runs. 

There's also a case that seems to be missing: when the continuation token provided is an empty string. 

Once these are addressed, we're good to go 🚀 nice work!

tests/aws/services/s3/test_s3_list_operations.py
bentsku: I think we could then update the test to use both the `BucketRegion` and the `Prefix` to make sure to return only the bucket you want, and then use the wrong region and the `Prefix`? 

I believe the issue you're mentioning could also be happening in this case if anything creates a bucket in `us-west-2`.

I think those list operations on global resources like this one are indeed tricky for parallelization, and most of those tests would fail when asserting the length of the list response.

tests/aws/services/s3/test_s3_list_operations.py
bentsku: Makes sense, work for me either way 👍

bentsku (commented):
Thanks for addressing the comments! This looks good 👌 I've answered the `BucketRegion` negative case question, and after that we should be good to go!
Seems like the test failures are unrelated, and linked to the Transcribe service?

localstack-core/localstack/services/s3/provider.py
```diff
@@ -581,14 +581,49 @@ def list_buckets(
...
+                Name=bucket.name,
+                CreationDate=bucket.creation_date,
+                BucketRegion=bucket.bucket_region,
+            )
+            buckets.append(output_bucket)
+            count += 1
+
+        # Weird behavior of AWS that we need to reproduce to keep parity
+        if len(buckets) == 0 and not prefix:
+            del owner["DisplayName"]
```
bentsku, line 622: question: are you sure about this? I believe this is region dependent, and is somewhat of a parity gap we have: some region have the `DisplayName`, and some don't. `us-east-1` has it. 

You don't have to address this parity gap in this PR, you can leave it, and add a snapshot skip for this particular response this way:
`@markers.aws.snapshot.skip_snapshot_verify(paths=["$.list-objects-by-bucket-region-empty..DisplayName"]` should do it, and it only ignore for the one specific snapshot key that is a problem (the `ListBuckets` call in the other region than `us-east-1`) 

Side-note: it would also a bit problematic to address it depending on region, because we run and generate our snapshots against `us-east-1` usually, but we run our pipeline against random regions (so we would have different behaviors). So this is a safe issue to ignore and snapshot skip 😄

bentsku: Follow-up: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Owner.html

It seems the `DisplayName` is going to be fully deleted in the near future, so that's also a great news 😄 we can delete it everywhere we have it, and regenerate all S3 snapshots. But let's keep it outside the scope of this PR, this is fine 😄

tests/aws/services/s3/test_s3_list_operations.py
```diff
@@ -46,6 +47,120 @@ def assert_timestamp_is_iso8061_s3_format(timestamp: str):
...
+        s3_create_bucket()
+        s3_create_bucket()
+
+        response = aws_client.s3.list_buckets(ContinuationToken="", MaxBuckets=1)
+        assert len(response["Buckets"]) == 1
+
+        snapshot.match("list-objects-with-empty-continuation-token", response)
+
+    @markers.aws.validated
+    # In some regions AWS returns the Owner display name (us-west-2) but in some it doesnt (eu-central-1)
```
bentsku, line 104: nice comment, thanks for adding it! 👌

bentsku (approved):
LGTM! This is great, thanks for taking the time to address the comments, this looks awesome! 💯 

Also, thank you for taking the time to properly transform the region field, making sure the tests are always passing even with non-default regions. 

Let's ship it! :shipit:

localstack-core/localstack/services/s3/provider.py
bentsku: This is good, thanks for checking this! 👍

PR #12614, opened 2025-05-13
ASF: Mark optional params as such (X | None)

bblommers:
## Motivation
As discussed in #12588 - if we want to enable MyPy on the entire codebase, one of the first steps would be to make sure that the generated API has the correct types.

## Changes
 - Marks optional parameters as such (`X | None = None` instead of `X = None`)
 - Updates the `test_asf_providers.py::test_provider_signatures` to handle all variations of Optional/Union params
 - Updates the API itself


alexrashed (approved):
Small and concise change and directly as discussed in the other linked PRs / conversations!
Would it make sense to maybe also update the APIs directly here in the PR in a separate commit and in our downstream project right after merging this one, instead of mixing it with the weekly automatically scheduled execution of the update next Monday? :)

alexrashed (approved):
🤩

MEPalma (approved):
The extra annotation for SFN is definitely welcome, as is the possibility to use MyPy; thank you!

Just curious, are there plans/or would it be feasible for us to distinguish between `Optional[X]` and `NotRequired[X]`? I find that distinction quite useful.

silv-io (approved):

bblommers:
> Just curious, are there plans/or would it be feasible for us to distinguish between `Optional[X]` and `NotRequired[X]`? I find that distinction quite useful.

@MEPalma I don't think I've ever used `NotRequired` actually!

Do you find the distinction useful when invoking a `method(arg: NotRequired)` to better understand the signature, or is it easier during the implementation of that method? As far as I can tell, there is no practical difference for the implementation - so I'm inclined to stick to `Optional`, just because it's simpler to read/understand.

pinzon (commented):
:+1:

Issue #12602, opened 2025-05-09
feature request: Add EC2 support for GetSecurityGroupsForVpc API operation


gregfurman:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

LocalStack does not currently support the EC2 API's [GetSecurityGroupsForVpc](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetSecurityGroupsForVpc.html) operation.

### 🧑‍💻 Implementation

This operation is effectively just syntatic sugar for a [DescribeSecurityGroups](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSecurityGroups.html) operation filtering for `vpc-id`.

i.e
```python
response = aws_client.ec2.describe_security_groups(
    Filters=[
        {
            'Name': 'vpc-id',
            'Values': [
                'vpc-abcd1234',
            ]
        },
    ]
)
```

iamramtin:
I'd like to take this on. Could you assign it to me please?

viren-nadkarni:
Fixed with #12615

PR #12615, opened 2025-05-13
Add EC2 support for GetSecurityGroupsForVpc API operation (#12602)

iamramtin:
## Motivation
This PR adds support for the EC2 API's `GetSecurityGroupsForVpc` operation, which retrieves security groups associated with a specified VPC. This operation provides a convenient way to filter security groups by VPC ID, without having to use the more generic filters syntax of `DescribeSecurityGroups`.

## Changes

- Added implementation for the `GetSecurityGroupsForVpc` EC2 API operation
- Built the implementation to leverage the existing `DescribeSecurityGroups` functionality with a VPC filter
- Added integration tests to verify both normal operation and edge cases
- Ensured proper validation for empty VPC IDs
- Made sure the response format matches AWS API specifications

## Testing
You can test this new functionality in the following way:

```
# Create a VPC and capture its ID
VPC_ID=$(awslocal ec2 create-vpc --cidr-block 10.0.0.0/16 --query Vpc.VpcId --output text)

# Create a security group in the VPC
awslocal ec2 create-security-group --group-name my-sg --description "My security group" --vpc-id $VPC_ID

# Filter security groups by VPC ID using describe-security-groups
awslocal ec2 describe-security-groups --filters Name=vpc-id,Values=$VPC_ID

# Use the new get-security-groups-for-vpc operation (should do the same as above)
awslocal ec2 get-security-groups-for-vpc --vpc-id $VPC_ID
```

Additionally, the PR includes integration tests that verify regular functionality and edge cases.


iamramtin:

iamramtin:
How can I add label `semver: minor`?

localstack-core/localstack/services/ec2/provider.py
```diff
@@ -539,6 +542,33 @@ def create_flow_logs(
...
         return response
 
+    @handler("GetSecurityGroupsForVpc", expand=False)
+    def get_security_groups_for_vpc(
+        self,
+        context: RequestContext,
+        get_security_groups_for_vpc_request: GetSecurityGroupsForVpcRequest,
+    ) -> GetSecurityGroupsForVpcResult:
+        vpc_id = get_security_groups_for_vpc_request.get("VpcId")
+        if vpc_id:
```
viren-nadkarni, line 552: `VpcId` is a [required parameter](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetSecurityGroupsForVpc.html). The conditional must be removed.

localstack-core/localstack/services/ec2/provider.py
```diff
@@ -539,6 +542,33 @@ def create_flow_logs(
...
+            next_token = get_security_groups_for_vpc_request.get("NextToken", None)
```
viren-nadkarni, line 553: `NextToken` is used for pagination, which is not implemented for this operation. A null value must be returned.

tests/aws/services/ec2/test_ec2.py
```diff
@@ -482,6 +482,87 @@ def test_create_vpc_with_custom_id(self, aws_client, create_vpc):
...
+                    "Tags": [
+                        {"Key": "Name", "Value": "test-vpc"},
+                    ],
+                }
+            ],
+        )
+        vpc_id: str = vpc["Vpc"]["VpcId"]
+
+        # Create security groups in the VPC
+        sg1_response = aws_client.ec2.create_security_group(
```
viren-nadkarni, line 502: Please use the `ec2_create_security_group` fixture

tests/aws/services/ec2/test_ec2.py
```diff
@@ -482,6 +482,87 @@ def test_create_vpc_with_custom_id(self, aws_client, create_vpc):
         assert e.value.response["ResponseMetadata"]["HTTPStatusCode"] == 400
         assert e.value.response["Error"]["Code"] == "InvalidVpc.DuplicateCustomId"
 
+    @markers.aws.only_localstack
```
viren-nadkarni, line 485: Ideally we want to verify both tests against AWS with snapshots. Take a look [here](https://blog.localstack.cloud/2022-08-04-parity-explained/) for more details. This is both for responses and exceptions. You can find several examples in other tests.

viren-nadkarni (commented):
Hi @iamramtin, thanks for raising the PR! I have added my review comments below. I will also add the semver label.

tests/aws/services/ec2/test_ec2.py
iamramtin: I tried to use the `ec2_create_security_group` fixture as requested, but I ran into an issue with writing the parity tests because it uses `GroupName` instead of `GroupId` in the `authorize_security_group_ingress` call, which fails when security groups with the same name exist in different VPCs.

I believe the error occurs because AWS requires using `GroupId` to be unique across security groups across VPCs, whereas `GroupName` only needs to be unique within a single VPC.

As a part of this PR, I've updated it to use `GroupId` instead of `GroupName`:

```
# Current implementation:
aws_client.ec2.authorize_security_group_ingress(
    GroupName=kwargs["GroupName"],
    IpPermissions=permissions,
)

# Suggested fix:
aws_client.ec2.authorize_security_group_ingress(
    GroupId=security_group["GroupId"],
    IpPermissions=permissions,
)
```

Please let me know if this is the right approach, otherwise I can revert it and create the security groups directly instead of the fixture (which works as well).

iamramtin: Nvm the above, I see a commit was pushed yesterday that resolves this

tests/aws/services/ec2/test_ec2.py
```diff
@@ -694,6 +694,84 @@ def _create_security_group() -> dict:
...
         assert e.value.response["Error"]["Code"] == "InvalidSecurityGroupId.DuplicateCustomId"
 
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            "$..Tags",
+            "$..SecurityGroupForVpcs..Description",
+            "$..SecurityGroupForVpcs..GroupId",
+            "$..SecurityGroupForVpcs..GroupName",
+        ]
+    )
```
viren-nadkarni, line 704: Generally paths are skipped only if they are not implemented in LocalStack. Why are these skipped?

iamramtin:
@viren-nadkarni thanks for your review. I took a closer look at the other snapshot tests in the codebase and reworked my approach. I've removed the assertions and now rely on the snapshot framework for validation similar to other parity tests.

I've also added some FIXME/TODOs to highlight potential issues / inconsistencies I came across. Let me know if this looks good, or if you'd prefer any further adjustments to the testing approach.

viren-nadkarni (changes requested):
Hi @iamramtin, the implementation currently breaks when a security group has tags. Could you please fix it and include the necessary test cases?

```
$ awslocal ec2 create-vpc --cidr-block 10.1.0.0/16 --query Vpc.VpcId
"vpc-596131631cff9cea1"

$ awslocal ec2 create-security-group --group-name hello --description world --vpc vpc-596131631cff9cea1 --tag-specifications "ResourceType=security-group,Tags=[{Key=colour,Value=blue}]"{
    "GroupId": "sg-cbf52384c65ddc0b6",
    "Tags": [
        {
            "Key": "colour",
            "Value": "blue"
        }
    ],
    "SecurityGroupArn": "arn:aws:ec2:eu-central-1:000000000000:security-group/sg-cbf52384c65ddc0b6"
}

$ awslocal ec2 get-security-groups-for-vpc --vpc-id vpc-596131631cff9cea1

An error occurred (InternalError) when calling the GetSecurityGroupsForVpc operation (reached max retries: 4): exception while calling ec2.GetSecurityGroupsForVpc: Traceback (most recent call last):
  File "/home/viren/.virtualenvs/ls_venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/handlers/service.py", line 113, in __call__
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/handlers/service.py", line 83, in __call__
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/skeleton.py", line 154, in invoke
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/skeleton.py", line 168, in dispatch_request
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/forwarder.py", line 133, in _call
    return handler(context, req)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/skeleton.py", line 118, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/viren/repo/localstack/localstack-core/localstack/aws/api/core.py", line 163, in operation_marker
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/viren/repo/localstack/localstack-core/localstack/services/ec2/provider.py", line 556, in get_security_groups_for_vpc
    sgs = [
          ^
  File "/home/viren/repo/localstack/localstack-core/localstack/services/ec2/provider.py", line 563, in <listcomp>
    Tags=[{"Key": k, "Value": v} for k, v in (sg.get_tags() or {}).items()],
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'items'
```

tests/aws/services/ec2/test_ec2.py
```diff
@@ -719,10 +722,21 @@ def test_get_security_groups_for_vpc(
...
+        snapshot.add_transformer(snapshot.transform.key_value("Value"))
 
         # Create VPC for testing
         vpc: dict = create_vpc(
             cidr_block="10.0.0.0/16",
+            tag_specifications=[
+                {
+                    "ResourceType": "vpc",
+                    "Tags": [
+                        {"Key": "Name", "Value": vpc_name},
```
viren-nadkarni, line 735: Try to avoid the value `Name` in `Key` attribute because it seems to mess up the snapshots.

```suggestion
                        {"Key": "SomethingElse", "Value": vpc_name},
```

tests/aws/services/ec2/test_ec2.py
```diff
@@ -719,10 +722,21 @@ def test_get_security_groups_for_vpc(
         snapshot.add_transformer(snapshot.transform.key_value("VpcId"))
         snapshot.add_transformer(snapshot.transform.key_value("AssociationId"))
         snapshot.add_transformer(snapshot.transform.key_value("DhcpOptionsId"))
+        snapshot.add_transformer(snapshot.transform.key_value("Key"))
+        snapshot.add_transformer(snapshot.transform.key_value("Value"))
```
viren-nadkarni, line 726: Only the attributes that return randomised values must be transformed. Tags are not random, they are same as the ones requested and must not be transformed.

localstack-core/localstack/services/ec2/provider.py
```diff
@@ -560,7 +560,14 @@ def get_security_groups_for_vpc(
                 GroupName=sg.name,
                 OwnerId=context.account_id,
                 PrimaryVpcId=sg.vpc_id,
-                Tags=[{"Key": k, "Value": v} for k, v in (sg.get_tags() or {}).items()],
+                Tags=[
+                    {
+                        "Key": tag.get("key") or tag.get("Key"),
+                        "Value": tag.get("value") or tag.get("Value"),
+                    }
+                    for tag in (sg.get_tags() or [])
```
viren-nadkarni, line 568: `get_tags()` [returns a list](https://github.com/getmoto/moto/blob/master/moto/ec2/models/core.py#L9-L13) so the conditional is redundant

```suggestion
                    for tag in sg.get_tags()
```

localstack-core/localstack/services/ec2/provider.py
```diff
@@ -560,7 +560,14 @@ def get_security_groups_for_vpc(
...
+                    if ("key" in tag or "Key" in tag) and ("value" in tag or "Value" in tag)
```
viren-nadkarni, line 569: Moto uses `key` and `value` fields (lowercase first characters).

Could you explain the need for capitalised field names?

viren-nadkarni (changes requested):
Thanks for bearing with me. If you could clarify these last concerns, we can merge this.

localstack-core/localstack/services/ec2/provider.py
iamramtin: I saw in other implementations the `Key` and `Value` fields were being processed with uppercase letters but those might not have been for Moto functionality... I have confirmed this only needs lower case, so I've updated the process in the provider file

tests/aws/services/ec2/test_ec2.py
iamramtin: I was using `short_uid` to generate the tags so they were random and that's why I was transforming them, but since that's not necessary I've just hardcoded `Key` and `Value` and thus removed the transform, let me know if that's okay

tests/aws/services/ec2/test_ec2.py
iamramtin: I've opted to just remove this tag since another test tag is already being created so it's not actually necessary

viren-nadkarni (approved):
Looks good, thanks for the contribution @iamramtin!

PR #12617, opened 2025-05-14
Core: Add type hints to aws/core.py

bblommers:
## Motivation
The main reason for wanting types in this file is to make sure that `context.account_id` and `context.region` are always a string. 

The handler chain will always set defaults for these variables, so we know they are set at runtime. Explicitly marking them as `str` makes it more obvious for service developers that they can trust the variables are set, and they don't have to manually assert that they are not None.

## Changes
Most of the changes are self-explanatory, but the more spicier ones:

### Installing `botocore` and `rolo` during linting

CircleCI now installs all `typehint`-dependencies during installation.

In our local venv we should have `requirements-typehints` installed, which means we have all the types - but the precommit-hook runs in an custom venv without those dependencies, so it will complain that they are missing. There are workarounds, for instance [by creating a custom hook](https://github.com/pre-commit/pre-commit/issues/266) for `mypy`, but explicitly installing the dependencies we need during the `pre-commit` step is the easiest solution for now.

### RequestContext API change
The `request`-parameter in `RoloRequestContext` is mandatory, and we do always supply it - so I've changed it to a mandatory parameter, and changed all the places where we create a `RequestContext` to add the `request` as an `__init__`-parameter.

Note that this class is never explicitly created in the AWS Pro repo, so we don't have to worry much about breaking changes. It is instantiated in Another emulator, but breaking changes are less important there, so I'll change that afterwards.


.circleci/config.yml
```diff
@@ -164,15 +164,16 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: python-requirements-{{ checksum "requirements-dev.txt" }}
+          key: python-requirements-{{ checksum "requirements-typehint.txt" }}
       - run:
           name: Setup environment
           command: |
+            make install-dev-types
```
alexrashed, line 171: @k-a-il @silv-io Now that the CircleCI pipeline is being migrated to GitHub Actions I think this would mean that we also have to change the make target here, right?
https://github.com/localstack/localstack/blob/9b6ef35c76a0685f8061e09f3145141c12368ad6/.github/actions/setup-tests-env/action.yml#L15

alexrashed (commented):
Awesome! This is another great push for MyPy type enforcements!
From my perspective the changes to the typing are great, I would just love to have @k-a-il or @silv-io approve the change / verify that this would not break the GitHub workflow which is currently being implemented :)

.circleci/config.yml
silv-io: Yes, would be great to add this in this PR.

On the other hand I think we should also be cautious with this switch because the typehint dependencies take waaaay longer to install in my experience (except when using `uv` :D)

alexrashed: Well, is this a blocker then? What does "waaaay longer" mean here?

silv-io: Yes :) 

- I'll add a suggestion in my review
- Then, we can manually run the `AWS / Build, Test, Push` pipeline on this PR to check if it's green. (We've still got some unrelated test failures around the Docker SDK there. These will not be blocking this PR)
- I've checked the CircleCI Pipeline and it looks like it will add 30s of installation time. Seems like caching works a bit better here than on-device, so the impact is not as big. The trade-off is fine in this case.

silv-io (changes requested):
The changes in CircleCI look fine to me. The increase in installation time is not too big and completely justified by the extra functionality we're gaining. 

Before merging this, I'd love to have these features in the GitHub Actions pipeline as well. To do this, we'd need to change these two actions to install the typehint dependency set instead of only the dev dependencies:

https://github.com/localstack/localstack/blob/940aa0a23638ff91b0af9a6adec404a0f4b986c7/.github/actions/setup-tests-env/action.yml#L6-L15

https://github.com/localstack/localstack/blob/549fd9b66127355b6a137753450f1de2163d7f3e/.github/actions/load-localstack-docker-from-artifacts/action.yml#L16-L21

/cc @k-a-il

bblommers:
@silv-io Thanks for checking! I've also changed the GH pipelines to now install `requirements-typehints`.

I've also changed the pipeline to use the `load-localstack-docker-from-artifacts`/`setup-tests-env` actions from the current branch, instead of from master - otherwise we'd never be able to test changes to those actions.

k-a-il (approved):
Nice, thank you for fixing paths to actions 👍 LGTM

k-a-il:
I just merged my fix for the integration tests `TestDockerClient` and `TestHooks`. If you rebase your branch, the tests `AWS / MA/MR tests / Run integration tests` will pass.

silv-io (approved):
LGTM! Thank you for the changes

bblommers:
Appreciate the review and feedback @silv-io and @k-a-il :pray:

PR #12610, opened 2025-05-12
Add tinybird reporting gha

silv-io:

PR #12627, opened 2025-05-16
Restrict tinybird reporting to master branch only

silv-io:
## Motivation

To not swarm the tinybird workspace with branch executions, we restrict the tinybird reporting to `master` only. 

I messed up by not re-activating it before merging in #12610 - sorry 😅 

## Changes

- Re-add guard for sending data to tinybird only on master


k-a-il (approved):
Quick fix - nice :)

Issue #12621, opened 2025-05-15
feature request: Support of $context.requestOverride.status|querystring  in apigateway


pieronatan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

LocalStack does not currently support overriding:

- `responseOverride.status` to set HTTP response status
- `requestOverride.querystring` to inject query parameters into the integration request

Both work correctly in AWS API Gateway when using Velocity templates (VTL), but fail in LocalStack.

---

#### Override response status code (VTL):

```vtl
#set($statusCode = $context.authorizer.statusCode)
#if($statusCode == 200)
    ##     Mapping the parameters from the request to the backend
    #foreach($param in $input.params().querystring.keySet())
        #set($context.requestOverride.querystring[$param]= $input.params().querystring.get($param))
        #if($foreach.hasNext) #end
    #end
    {
        "statusCode": 200,
        "message": "Valid scope found"
    }
#else
    #set($context.responseOverride.status = $statusCode)
    {
        "error": $context.authorizer.error
        "statusCode": $statusCode,
    }
#end
```
In **AWS** → returns HTTP 200.  
In **LocalStack** → returns HTTP 200, or fails with:

```
2025-05-15T04:12:58.145  WARN --- [et.reactor-5] l.s.a.n.e.h.gateway_except : Non Gateway Exception raised: line 5, column 13: expected assignment in set directive, got: ($context.requestOverride.querystrin ... ...
 Traceback (most recent call last):
   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/airspeed/operators.py", line 350, in require_next_element
     element = element_spec(self.filename, self._full_text, self.end)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/airspeed/operators.py", line 289, in __init__
     self.parse()
   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/airspeed/operators.py", line 1077, in parse
     (var_name,) = self.identity_match(self.START)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/opt/code/localstack/.venv/lib/python3.11/site-packages/airspeed/operators.py", line 306, in identity_match
     raise NoMatch()
 airspeed.operators.NoMatch
 
 During handling of the above exception, another exception occurred:
 
[...]

PR #12628, opened 2025-05-16
Apigw/add support for response override in request

cloutierMat:
## Motivation

This is a partial solve for #12621. 

This pr will allow to use `responseOverride` in a request template and have it's value passed to the context of the response template and possibly affecting the response.

There are 2 issues described by the user. I will work on the VTL template update in a follow up.

## Changes

- Move responsibility of creating the overrides context variable to the parser. Those values are accessible in the context and that will prevent needing to check for their existence everywhere in the code
- Pass the full return of the context variables from the request mapping to properly update the context that is passed down
- Added validated test to confirm the behavior

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
@@ -16,6 +16,11 @@ from .context import InvocationRequest, RestApiInvocationContext
 from .handlers.resource_router import RestAPIResourceRouter
 from .header_utils import build_multi_value_headers
 from .template_mapping import dict_to_string
+from .variables import (
+    ContextVariableOverrides,
+    ContextVarsRequestOverride,
+    ContextVarsResponseOverride,
+)
 
 # TODO: we probably need to write and populate those logs as part of the handler chain itself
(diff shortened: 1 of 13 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
```diff
@@ -159,6 +159,7 @@ def create_test_invocation_context(
         requestOverride=ContextVarsRequestOverride(header={}, path={}, querystring={}),
         responseOverride=ContextVarsResponseOverride(header={}, status=0),
     )
+    invocation_context.trace_id = parse_handler.populate_trace_id({})
```
cloutierMat, line 162: 🤔 ??

bentsku: it had been deleted in the previous commit, I've just added it back 😄 https://github.com/localstack/localstack/pull/12628/commits/bb5dbe1c81d73dcc24345f8b7990f5f27336a0a2

bentsku (approved):
LGTM! This looks good, quite a really weird case by AWS 👀

It's also funny that we can add user-defined fields to the `responseOverride`, acting like a context stash for user-defined values... 

That must have been a fun one 😄 thanks for added parity and jumping on this! 🚀

PR #12629, opened 2025-05-16
fix: correct recognition of lambda path

thetric:
## Motivation

The AWS Lambda SDK invokes `/2015-03-31/functions` to create a new Lambda function (see https://docs.aws.amazon.com/lambda/latest/api/API_CreateFunction.html) but the handler previously did not recognise it as a Lambda route due to a trailing slash in the path.

## Changes

Localstack now recognises `/2015-03-31/functions` as a Lambda endpoint

### DIFF
--- a/localstack-core/localstack/aws/protocol/service_router.py
+++ b/localstack-core/localstack/aws/protocol/service_router.py
@@ -155,7 +155,7 @@ def custom_path_addressing_rules(path: str) -> Optional[ServiceModelIdentifier]:
     if is_sqs_queue_url(path):
         return ServiceModelIdentifier("sqs", protocol="query")
 
-    if path.startswith("/2015-03-31/functions/"):
+    if path.startswith("/2015-03-31/functions"):
         return ServiceModelIdentifier("lambda")
 
 


alexrashed (commented):
Thanks a lot, @thetric for the contribution! I would love to hand over the review for this change to our experts on Lambda to make sure that this change (a) really fixes the issue you see, and (b) ensure that this really is in parity with AWS.
@joe4dev @dfangl @gregfurman could you maybe take over here?

dfangl:
Hi @thetric ! Could you please elaborate what exact issues this solves? Which SDK is affected from this issue?

thetric:
I'm trying to create a new Lambda function with the AWS SDK Java v2. The Lambda client is configured to use an IP address (because it's running in our CI via Docker-in-Docker). With the 4.4 release of Localstack, this request fails with an empty HTTP 200 response and the following log:

> l.aws.handlers.service     : no service set in context, skipping request parsing

Searching the Localstack code I found https://github.com/localstack/localstack/blob/master/localstack-core/localstack/aws/protocol/service_router.py#L158. The request parsing only recognises Lambda requests if the path starts with `/2015-03-31/functions/` but not all Lambda requests contain the trailing slash in the path, such as: https://docs.aws.amazon.com/lambda/latest/api/API_CreateFunction.html: `POST /2015-03-31/functions`

thetric:
Rebased onto master. Had to manually rebase + force push, because the "Sync fork" action seems to create a merge commit instead

Issue #12631, opened 2025-05-17
feature request: Implement ListRuleNamesByTarget


etiago:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

Currently LocalStack does not implement the [ListRuleNamesByTarget](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_ListRuleNamesByTarget.html) API call.

I have an implementation ready and this issue serves to track it.

### 🧑‍💻 Implementation

Since this is a small feature, I already have an initial implementation ready in a PR that I'll link with this issue.

bentsku:
Hello @etiago and thanks for the contribution! As the linked PR has been merged, I'll close the issue 👍 feel free to comment if there's anything. Thanks!

PR #12632, opened 2025-05-17
Feature: implement list rule names by target

etiago:
## Motivation
I'm raising this PR to increase LocalStack's coverage of AWS APIs. Currently ListRuleNamesByTarget is not implemented.

It fixes issue #12631 .

## Changes
This PR makes it so that instead of returning `NotImplementedError`, LocalStack will start returning the actual list of EventBus rules for a given target.

## Testing
The changes are covered by snapshot/parity tests.

The happy path can also be tested manually by spinning up LocalStack and following these steps:
```bash
awslocal sqs create-queue --queue-name test-sqs-queue
QUEUE_ARN=$(awslocal sqs get-queue-attributes \
  --queue-url http://sqs.eu-west-2.localhost.localstack.cloud:4566/000000000000/test-sqs-queue \
  --attribute-names QueueArn \
  --query 'Attributes.QueueArn' \
  --output text)
awslocal events create-event-bus --name test-event-bus
awslocal events put-rule \
  --name rule-to-sqs \
  --event-bus-name test-event-bus \
  --event-pattern '{"source": ["test-source"]}'
RULE_ARN=$(aws events describe-rule \
  --name rule-to-sqs \
  --event-bus-name test-event-bus \
  --query 'Arn' \
  --output text)
awslocal events put-targets \
  --rule rule-to-sqs \
  --event-bus-name test-event-bus \
  --targets "Id"="1","Arn"="$QUEUE_ARN"

# Verify that the correct target ARN returns the rule:
awslocal events list-rule-names-by-target --target-arn $QUEUE_ARN --event-bus-name test-event-bus

# Verify that a non-existant target ARN returns an empty rule list:
awslocal events list-rule-names-by-target --target-arn "non-existant-arn" --event-bus-name test-event-bus
```

### DIFF
--- a/tests/aws/services/events/test_events.py
+++ b/tests/aws/services/events/test_events.py
@@ -1760,13 +1760,11 @@ class TestEventRule:
         queue_name = f"queue-{short_uid()}"
         queue_url = sqs_create_queue(QueueName=queue_name)
         queue_arn = sqs_get_queue_arn(queue_url)
-        snapshot.add_transformer(snapshot.transform.regex(queue_arn, "<queue-arn>"))
 
         # Create an event bus if using custom bus
         if bus_name == "custom":
             bus_name = f"bus-{short_uid()}"
             events_create_event_bus(Name=bus_name)
-            snapshot.add_transformer(snapshot.transform.regex(bus_name, "<bus-name>"))
 
         # Create multiple rules targeting the same SQS queue
         rule_prefix = f"rule-{short_uid()}-"
@@ -1785,7 +1783,6 @@ class TestEventRule:
 
             # Add the SQS queue as a target for this rule
             target_id = f"target-{i}"
-            snapshot.add_transformer(snapshot.transform.regex(target_id, f"<target-id-{i}>"))
             aws_client.events.put_targets(
                 Rule=rule_name,
                 EventBusName=bus_name,
@@ -1809,38 +1806,6 @@ class TestEventRule:
         # The response should contain all rules that target our queue
         snapshot.match("list_rule_names_by_target", response)
 
-        # Verify the response contains the expected rule names
-        assert set(response.get("RuleNames", [])) == set(rule_names)
-        assert other_rule not in response.get("RuleNames", [])
-
-        # Clean up targets and rules
-        for rule_name in rule_names + [other_rule]:
-            try:
-                # Remove targets first (only for rules that have targets)
-                if rule_name in rule_names:  # other_rule has no targets
-                    aws_client.events.remove_targets(
-                        Rule=rule_name,
-                        EventBusName=bus_name,
-                        Ids=[f"target-{rule_names.index(rule_name)}"],
-                    )
-                # Delete rule
-                aws_client.events.delete_rule(
-                    Name=rule_name,
-                    EventBusName=bus_name,
-                    Force=True,
-                )
-            except aws_client.events.exceptions.ResourceNotFoundException:
-                # Ignore if rule or event bus was already deleted
-                pass
-
-        # Delete custom event bus if we created one
-        if bus_name != "default":
-            try:
-                aws_client.events.delete_event_bus(Name=bus_name)
-            except aws_client.events.exceptions.ResourceNotFoundException:
-                # Ignore if event bus was already deleted
-                pass
-
     @markers.aws.validated
     @pytest.mark.parametrize("bus_name", ["custom", "default"])
     def test_list_rule_names_by_target_with_limit(
@@ -1859,13 +1824,11 @@ class TestEventRule:
         queue_name = f"queue-{short_uid()}"
         queue_url = sqs_create_queue(QueueName=queue_name)
         queue_arn = sqs_get_queue_arn(queue_url)
-        snapshot.add_transformer(snapshot.transform.regex(queue_arn, "<queue-arn>"))
 
         # Create an event bus if using custom bus
         if bus_name == "custom":
             bus_name = f"bus-{short_uid()}"
             events_create_event_bus(Name=bus_name)
-            snapshot.add_transformer(snapshot.transform.regex(bus_name, "<bus-name>"))
 
         # Create multiple rules targeting the same SQS queue
         rule_prefix = f"rule-{short_uid()}-"
@@ -1884,7 +1847,6 @@ class TestEventRule:
 
             # Add the SQS queue as a target for this rule
             target_id = f"target-{i}"
-            snapshot.add_transformer(snapshot.transform.regex(target_id, f"<target-id-{i}>"))
             aws_client.events.put_targets(
                 Rule=rule_name,
                 EventBusName=bus_name,
@@ -1902,17 +1864,19 @@ class TestEventRule:
             Limit=2,
         )
         # Store the original NextToken value before replacing it for snapshot comparison
-        original_next_token = response["NextToken"]
-        # Transform NextToken to a fixed value for snapshot comparison
-        if "NextToken" in response:
-            response["NextToken"] = "<next-token>"
+        next_token = response["NextToken"]
+        snapshot.add_transformer(
+            snapshot.transform.jsonpath(
+                jsonpath="$..NextToken",
+                value_replacement="<next-token>",
+                reference_replacement=True,
+            )
+        )
+
         snapshot.match("first_page", response)
         all_rule_names.extend(response["RuleNames"])
-        assert len(response["RuleNames"]) == 2
-        assert "NextToken" in response
 
         # Second page
-        next_token = original_next_token
         response = aws_client.events.list_rule_names_by_target(
             TargetArn=queue_arn,
             EventBusName=bus_name,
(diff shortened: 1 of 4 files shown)


etiago:

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+                # Ignore if rule or event bus was already deleted
+                pass
+
+        # Delete custom event bus if we created one
+        if bus_name != "default":
+            try:
+                aws_client.events.delete_event_bus(Name=bus_name)
+            except aws_client.events.exceptions.ResourceNotFoundException:
+                # Ignore if event bus was already deleted
+                pass
```
dominikschubert, line 1842: :bulb: A general pattern we try to follow, is to be very defensive about cleanups to avoid leaking resources. This is particularly important when running tests against AWS, where they might continue incurring cost when leaked.

In that spirit we typically avoid cleanup code like this:

```python
# creating resource_1
# creating dependent resource_2
...
delete_resource_2(...)
delete_resource_1(...)
...
```
even when wrapped in try/catch blocks.

Instead we have a pytest fixture called `cleanups` which is a simple list of callables that are called in reverse order of insertion when the test has finished executing.

Given my example above, this would correspond to:
```python
# creating resource_1
...
cleanups.append(lambda: delete_resource_1(...))
 
...

# creating dependent resource_2
...
cleanups.append(lambda: delete_resource_2(...))
```

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+
+    @markers.aws.validated
+    @pytest.mark.parametrize("bus_name", ["custom", "default"])
+    def test_list_rule_names_by_target_no_matches(
```
dominikschubert, line 1980: question: Is there a particular reason this deviates from the cleanup structure in your other tests?

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+        )
+        # Store the original NextToken value before replacing it for snapshot comparison
+        original_next_token = response["NextToken"]
+        # Transform NextToken to a fixed value for snapshot comparison
+        if "NextToken" in response:
+            response["NextToken"] = "<next-token>"
+        snapshot.match("second_page", response)
+        all_rule_names.extend(response["RuleNames"])
+        assert len(response["RuleNames"]) == 2
+        assert "NextToken" in response
```
dominikschubert, line 1930: nit: While asserts are ok to leave in, we typically try to avoid it for testing API responses as it's more brittle than the snapshots and while technically redundant, it does come with an additional maintenance overhead, particularly when rewriting / refactoring a test. For parity tests a good guideline is to keep them mostly for asserting behavior or if you want to be extremely explicit in what edge case the test case is covering.

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+        aws_client,
+        snapshot,
+        clean_up,
+    ):
+        """Test the ListRuleNamesByTarget API with pagination to verify it correctly handles limits and next tokens."""
+        # Create an SQS queue to use as a target
+        queue_name = f"queue-{short_uid()}"
+        queue_url = sqs_create_queue(QueueName=queue_name)
+        queue_arn = sqs_get_queue_arn(queue_url)
+        snapshot.add_transformer(snapshot.transform.regex(queue_arn, "<queue-arn>"))
```
dominikschubert, line 1862: This doesn't actually do anything here since we're not capturing any response with a queue ARN
```suggestion
```

Similar with the other test cases

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+            rule_names.append(rule_name)
+            events_put_rule(
+                Name=rule_name,
+                EventBusName=bus_name,
+                EventPattern=json.dumps({"source": [f"source-{i}"]}),
+            )
+
+            # Add the SQS queue as a target for this rule
+            target_id = f"target-{i}"
+            snapshot.add_transformer(snapshot.transform.regex(target_id, f"<target-id-{i}>"))
```
dominikschubert, line 1887: Like with the SQS ARN and EventBus name, we don't capture target ids in the response
```suggestion
```

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+        queue_name = f"queue-{short_uid()}"
+        queue_url = sqs_create_queue(QueueName=queue_name)
+        queue_arn = sqs_get_queue_arn(queue_url)
+        snapshot.add_transformer(snapshot.transform.regex(queue_arn, "<queue-arn>"))
+
+        # Create an event bus if using custom bus
+        if bus_name == "custom":
+            bus_name = f"bus-{short_uid()}"
+            events_create_event_bus(Name=bus_name)
+            snapshot.add_transformer(snapshot.transform.regex(bus_name, "<bus-name>"))
```
dominikschubert, line 1868:
```suggestion
```
Same case as with the SQS ARN, we're not actually capturing a bus name anywhere in the collected snapshots here.

tests/aws/services/events/test_events.py
```diff
@@ -1742,6 +1742,293 @@ def test_process_pattern_to_single_matching_rules_single_target(
...
+        response = aws_client.events.list_rule_names_by_target(
+            TargetArn=queue_arn,
+            EventBusName=bus_name,
+            Limit=2,
+        )
+        # Store the original NextToken value before replacing it for snapshot comparison
+        original_next_token = response["NextToken"]
+        # Transform NextToken to a fixed value for snapshot comparison
+        if "NextToken" in response:
+            response["NextToken"] = "<next-token>"
```
dominikschubert, line 1908: Q: Typically we'd solve this with a transformer. Any particular reason to avoid doing that here, as well as in the other cases?

dominikschubert (changes requested):
Hey @etiago ! Great job on the first iteration here and thanks for tackling this. :rocket:
Showing great attention already to cleaning up resources and covering different edge cases :clap:  

There's a few minor nits and things we should address before it being ready to merge, particularly for the tests to make sure they follow our best practices for parity testing.
The comments are not added exhaustively, so please extrapolate accordingly and adapt similar cases to the ones commented one. 

Let me know if you have any questions :+1:

tests/aws/services/events/test_events.py
etiago: It's a good point... I just had a habit of asserting at the end of tests, but you're right, since we have the snapshots which do a 1:1 validation, the asserts add no value and I see what you mean about the maintenance cost going forwards.

Will remove these.

tests/aws/services/events/test_events.py
etiago: No reason... I was stuck thinking of the `regex` transformer which wouldn't work here since the `NextToken` is different each time.

But since you mentioned this, I now spotted we have a `jsonpath` transformer which makes this much simpler, since I can now just have a single
```python
        snapshot.add_transformer(
            snapshot.transform.jsonpath(
                jsonpath="$..NextToken",
                value_replacement="<next-token>",
                reference_replacement=True,
            )
        )
```

And this replaces all instances of `NextToken` and keeps them numbered when verifying multiple times.

I regenerated the snapshot and it's all good now.

tests/aws/services/events/test_events.py
etiago: No reason...

To be completely honest, looks like I "forgot" to do it for this test.

In hindsight, after reading your other comment above about cleanups, I now realise that this test is actually the correct one in that as long as resources are created using the existing fixtures, those resources are already tracked and automatically cleaned up at the end of the test.

tests/aws/services/events/test_events.py
etiago: This makes sense.

Having looked through my test again, and also looking through some of the existing tests, I think I don't even need to do this via `cleanups`.

Best as I can tell, looking at the `events_create_event_bus` fixture [here](https://github.com/localstack/localstack/blob/master/localstack-core/localstack/testing/pytest/fixtures.py#L2437) for example, `events_create_event_bus` actually yields a function which when I call it to create event buses, it also adds them to a list to keep track of them. And on tear down, it cycles through that list, cleaning them up.

So I think I can safely remove the whole cleanup code and as long as I use the existing fixtures to create resources, my tests don't need to explicitly worry with cleaningup.

bblommers (approved):

dominikschubert (approved):
Thanks for addressing the issues and great work on identifying that fixtures are doing their own cleanup already, reducing the need for manual cleanup!

Congratulations on the first contribution to LocalStack :partying_face:

PR #12634, opened 2025-05-19
ffmpeg: Update build source to use BtbN GitHub Releases

sannya-singal:
## Motivation
The current `ffmpeg` installer in LocalStack downloads binaries from [johnvansickle.com](https://johnvansickle.com/ffmpeg/), a privately hosted site that has exhibited timeout issues, likely due to bandwidth limitations or rate throttling.

To address this, we are switching to [BtbN/FFmpeg-Builds](https://github.com/BtbN/FFmpeg-Builds) — a well-maintained, GitHub-hosted source for `ffmpeg` binaries with:
- Actively maintained nightly and tagged releases
- LGPL-compliant builds suitable for LocalStack’s hybrid licensing model

## Changes
- Uses the latest build of a specific major version and replaces `ffmpeg` download URL with BtbN-hosted static LGPL `7.1` builds. 
- Resolves architecture dynamically based on the host (e.g., linux64, linuxarm64)
- Updates the structure handling to match BtbN’s layout (`bin/ffmpeg`, `bin/ffprobe`)

## Testing
Successful installation tested in local setup-
Screenshot 2025-05-16 at 3 14 22 PM

Successful installation tested in CI-
https://app.circleci.com/pipelines/github/localstack/localstack/32720/workflows/4f5455cf-dddb-4d27-b11b-75c52c3c604a

### DIFF
--- a/localstack-core/localstack/packages/ffmpeg.py
+++ b/localstack-core/localstack/packages/ffmpeg.py
@@ -5,20 +5,22 @@ from localstack.packages import Package
 from localstack.packages.core import ArchiveDownloadAndExtractInstaller
 from localstack.utils.platform import get_arch
 
-FFMPEG_STATIC_BIN_URL = (
-    "https://www.johnvansickle.com/ffmpeg/releases/ffmpeg-{version}-{arch}-static.tar.xz"
-)
+# Mapping LocalStack architecture to BtbN's naming convention
+ARCH_MAPPING = {"amd64": "linux64", "arm64": "linuxarm64"}
+
+# Download URL template for ffmpeg 7.1 LGPL builds from BtbN GitHub Releases
+FFMPEG_STATIC_BIN_URL = "https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-n{version}-latest-{arch}-lgpl-{version}.tar.xz"
 
 
 class FfmpegPackage(Package["FfmpegPackageInstaller"]):
     def __init__(self) -> None:
-        super().__init__(name="ffmpeg", default_version="7.0.1")
+        super().__init__(name="ffmpeg", default_version="7.1")
 
     def _get_installer(self, version: str) -> "FfmpegPackageInstaller":
         return FfmpegPackageInstaller(version)
 
     def get_versions(self) -> List[str]:
-        return ["7.0.1"]
+        return ["7.1"]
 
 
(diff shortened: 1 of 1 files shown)


k-a-il (approved):
Great, I hope this source proves to be more reliable 🚀

viren-nadkarni (commented):
Thanks for jumping on this!

I was wondering about the [retention policy](https://github.com/BtbN/FFmpeg-Builds?tab=readme-ov-file#release-retention-policy):

> The last build of each month is kept for two years.
> The last 14 daily builds are kept.

Have you verified that if there is a new ffmpeg release (7.2), would our pinning of 7.1 and use of `latest` marker in the URL not cause the link to break? Would it continue to point to the 'last build of the month' for as long as the build is stored?

The link will break beyond two years, I'm guessing that's acceptable lifetime of a LS release anyway.

I would also like @alexrashed to take a look at this.

sannya-singal:
> Thanks for jumping on this!
> 
> I was wondering about the [retention policy](https://github.com/BtbN/FFmpeg-Builds?tab=readme-ov-file#release-retention-policy):
> 
> > The last build of each month is kept for two years.
> > The last 14 daily builds are kept.
> 
> Have you verified that if there is a new ffmpeg release (7.2), would our pinning of 7.1 and use of `latest` marker in the URL not cause the link to break? Would it continue to point to the 'last build of the month' for as long as the build is stored?
> 
> The link will break beyond two years, I'm guessing that's acceptable lifetime of a LS release anyway.
> 
> I would also like @alexrashed to take a look at this.

Great points!

Regarding your concern: It would not break immediately. The URL targets a 7.1-specific build, which will continue to receive nightly rebuilds until 7.2 becomes the new upstream release. Once 7.2 is out, n7.1-latest would no longer be updated, but the last 7.1 monthly build would remain available for 2 years, but will not be referenced by latest anymore in such cases. 

After that, we would need to either: Refresh the pinned version (e.g., to 7.2) or follow the approach already described in Notion concerning the mirror the binary internally (e.g., to S3 or artifacts) if long-term retention is required. 

I already synced with @alexrashed about the current approach in Notion, but I’m happy to wait for his review, though that likely means a ~3-week timeline since he’s on vacation 😛

Let me know how you'd like to proceed! 🙂

sannya-singal:
Marking this PR as a draft to prevent any accidental merges until we reach a conclusion on this.

viren-nadkarni (approved):

PR #12638, opened 2025-05-19
Record validated tests duration

tiurin:
## Motivation

Have more insights into how long does it take to run a parity test against AWS. Have more insights for each test execution phase.

## Changes

- Add a pytest hook that collects duration of each test execution phase - setup, call, teardown and writes the result into `*.validation.json`, along with already existing last validation date.
- Remove the old hook and some unused code.
- Re-validate one existing test to demonstrate the example. It is especially interesting to see how teardown can be almost half of total duration in some cases. This is the case when one sees test passed in PyCharm but the wheels are still spinning for a while:
  image

## Considerations

Machines on which tests are executed can vary significantly in processing power. However, in case of AWS validated tests significant amount of time is spent in I/O waiting for AWS endpoint responses. That can also be affected by geographic location. A quick test on a small sample showed under 2% difference on the test in this PR on 3 different machines between Spain and South Africa (all tests were executed against us-east-1 region). Repetitive runs in different times of day on one machine were also similarly stable.

In any case, durations are meant to give _an idea_ about how long the test runs rather than exact numbers.

### DIFF
--- a/localstack-core/localstack/testing/pytest/validation_tracking.py
+++ b/localstack-core/localstack/testing/pytest/validation_tracking.py
@@ -12,9 +12,8 @@ from pathlib import Path
 from typing import Dict, Optional
 
 import pytest
-from _pytest.reports import TestReport
-from _pytest.stash import StashKey
 from pluggy import Result
+from pytest import StashKey, TestReport
 
 from localstack.testing.aws.util import is_aws_cloud
 


tests/aws/services/lambda_/event_source_mapping/test_cfn_resource.validation.json
```diff
@@ -1,5 +1,11 @@
 {
   "tests/aws/services/lambda_/event_source_mapping/test_cfn_resource.py::test_adding_tags": {
-    "last_validated_date": "2024-11-06T11:55:29+00:00"
+    "last_validated_date": "2025-05-19T08:31:57+00:00",
+    "durations_by_phase": {
```
bblommers, line 4:
```suggestion
    "durations_in_sec": {
```
Just a suggestion. Having the units makes it easier to read IMO - and the fact that this is by phase is already obvious from the content of the dict.

Love the initiative btw! Very useful to include this data in the test suite.

anisaoshafi (approved):
Looks good to me, tried it out and works like a charm ✨ 
Thanks for tackling this Misha. 👏🏼

localstack-core/localstack/testing/pytest/validation_tracking.py
```diff
@@ -29,65 +41,76 @@ def find_snapshot_for_item(item: pytest.Item) -> Optional[dict]:
...
-        fd.seek(0)
-        json.dump(content, fd, indent=2, sort_keys=True)
-        fd.write("\n")  # add trailing newline for linter and Git compliance
+        timestamp = datetime.datetime.now(tz=datetime.timezone.utc)
+        test_execution_data["last_validated_date"] = timestamp.isoformat(timespec="seconds")
 
+        durations_by_phase = item.stash[durations_key]
+        test_execution_data["durations_in_seconds"] = durations_by_phase
 
-# TODO: we should skip if we're updating snapshots
```
joe4dev, line 66: question: Are we intentionally removing this TODO?
If so, it might be worth motivating why. I'm in favor of having some mechanism (kinda snapshot police ;) to identify tests that have not been validated after the initial snapshot recording. Tests not following the order: snapshot first -> validate *AFTER* often suffer from transformer issues.

Theoretically, we could compare the `last_validated_date` and `recorded-date` (i.e., basically resolves this TODO), but practically this could be imprecise because the timestamps are generated separately and formatted differently 😿 

TLDR: I think we should have some data-driven way to identify such issues, but we can probably keep the current practice of updating the validation upon snapshotting because a) it's still a valid aws-validated run and b) it's hard (and probably not worth) to change retroactively.

localstack-core/localstack/testing/pytest/validation_tracking.py
```diff
@@ -29,65 +41,76 @@ def find_snapshot_for_item(item: pytest.Item) -> Optional[dict]:
...
-        return
+        # For json.dump sorted test entries enable consistent diffs.
+        # But test execution data is more readable in insert order for each step (setup, call, teardown).
+        # Hence, not using global sort_keys=True for json.dump but rather additionally sorting top-level dict only.
+        content = dict(sorted(content.items()))
 
-    record_passed_validation(item)
-
-
-# this is a sort of utility used for retroactively creating validation files in accordance with existing snapshot files
```
joe4dev, line 79: question: I guess we don't need that anymore because we did this initially and cannot reconstruct timings anyway, right?

joe4dev (approved):
Love it ❤️ Thank you @tiurin for the making this happen 🚀 

I can confirm that time reporting works as expected, both in localstack and localstack-ext. I also tested that failing tests or non-AWS test executions are ignored.

localstack-core/localstack/testing/pytest/validation_tracking.py
tiurin: Yes, intentionally removing it as this is hardly actionable from a TODO in the code. I added an issue to snapshot backlog board instead and tagged you as a co-reporter. Thanks for brining the context, indeed I don't think the order is followed universally, I myself often skip AWS validated run, relying on snapshot recording session...

localstack-core/localstack/testing/pytest/validation_tracking.py
tiurin: Yes, correct, that was for the initial validation files creation.

I haven't seen a test that would only have a snapshot file without a validation one. Maybe there are some, but I think those would be quite old ones and should definitely be re-validated properly. :)

Removing commented out code to reduce the noise. If ever needed, it can be found through git history.

dominikschubert (approved):
Gave it a quick glance and just had some minor suggestions. 

Thanks for tackling this! One somewhat immediate improvement I could think of is having a pytest flag that would allow you to print statistics of these execution times or at least collect them for each known pytest node id (i.e. for each test). We could use this to track how far along we are in collecting execution data over the whole test suite as well. :thinking: 

Just some thoughts though, nothing immediately actionable besides the import comment :+1:

docs/testing/parity-testing/README.md
```diff
@@ -248,3 +248,12 @@ localstack.testing.snapshots.transformer: Replacing regex '000000000000' with '1
 localstack.testing.snapshots.transformer: Replacing regex 'us-east-1' with '<region>'
 localstack.testing.snapshots.transformer: Replacing '1ad533b5-ac54-4354-a273-3ea885f0d59d' in snapshot with '<uuid:1>'
 ```
+
+### Test duration recording
+
+When a test runs successfully against AWS, its last validation date and duration are recorded in a corresponding ***.validation.json** file.
+The validation date is recorded precisely, while test durations can vary between runs.
+For example, test setup time may differ depending on whether a test runs in isolation or as part of a class test suite with class-level fixtures.
```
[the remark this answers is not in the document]
tiurin: One thing I thought about was to get test collection name via `item.session.config.args`, e.g. `['aws/services/lambda_/event_source_mapping']`. It can be used as a part of key, or a unique property, so that test duration is updated if the test has been validated within the same test collection.
However, such usage means we should record durations for each new test collection, which is confusing. Or, record only for a predefined collections, e.g. only for individual runs, or only for class runs. Which is also confusing and can be opaque ("why my durations haven't been updated?"). Also, test ordering might come into play for collections. Plus, need to sanitize args as they may contain full path and reveal local setup details. Quite hard to factor in due to many unknown details and their unknown impact.

I'd bet now on simplicity, see if durations actually flip a lot (somewhat good sign, means tests are re-validated, hehe) and learn how we can adapt a format if needed.

tiurin (commented):
> One somewhat immediate improvement I could think of is having a pytest flag that would allow you to print statistics of these execution times or at least collect them for each known pytest node id (i.e. for each test). We could use this to track how far along we are in collecting execution data over the whole test suite as well. 

I think this call for reporting rather than printing, maybe a dashboard? I'll think about it.

PR #12640, opened 2025-05-19
Added retries to the container state check in TestDocker

k-a-il:
## Motivation

I missed fixing one of the flaky tests in the `TestDockerClient` class in the PR #12625 that was merged today. Like `test_is_container_running`, test `test_running_container_names` fails due to a race condition:

```
FAILED tests/integration/docker_utils/test_docker.py::TestDockerClient::test_running_container_names[SdkDockerClient] - AssertionError: assert 'lst_test_4337619e' not in ['lst_test_4337619e', 'sleepy_tu']
```
We need to wait until the container is fully stopped. I also added a retry to the container start step, just to be safe.

## Changes

- Added function `_assert_container_state` which is used in `test_running_container_names` and `test_is_container_running`
- Added retries to `test_running_container_names`

## Testing

Tested here (transcribe tests are failing but it is not connected to changes in this PR, tests changed in this PR are passing): https://github.com/localstack/localstack/actions/runs/15117166475

### DIFF
--- a/tests/integration/docker_utils/test_docker.py
+++ b/tests/integration/docker_utils/test_docker.py
@@ -70,6 +70,10 @@ def _is_podman_test() -> bool:
     return os.getenv("DOCKER_CMD") == "podman"
 
 
+def _assert_container_state(docker_client: ContainerClient, name: str, is_running: bool):
+    assert docker_client.is_container_running(name) == is_running
+
+
 @pytest.fixture
 def dummy_container(create_container):
     """Returns a container that is created but not started"""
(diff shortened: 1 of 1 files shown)


tests/integration/docker_utils/test_docker.py
```diff
@@ -1273,22 +1277,39 @@ def test_run_container_non_existent_image(self, docker_client: ContainerClient):
     def test_running_container_names(self, docker_client: ContainerClient, dummy_container):
         docker_client.start_container(dummy_container.container_id)
         name = dummy_container.container_name
-        assert name in docker_client.get_running_container_names()
+        retry(
```
silv-io, line 1280: **comment:** This works because the retry catches the AssertionError, which doesn't happen in, e.g., `poll_condition`. Was a bit confused at first :D

silv-io (approved):
LGTM!

Issue #7405, opened 2022-12-30
feature request: DynamoDB Streams for global tables


viren-nadkarni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

It must be possible to access DynamoDB Streams in all regions where a global DynamoDB table is replicated.

Currently this is not supported.

```
# Create a table
$ awslocal dynamodb create-table \
    --table-name global01 \
    --key-schema AttributeName=id,KeyType=HASH \
    --attribute-definitions AttributeName=id,AttributeType=S \
    --billing-mode PAY_PER_REQUEST \
    --region ap-south-1

# Create replicas
$ awslocal dynamodb update-table \
    --table-name global01 \
    --replica-updates '[{"Create": {"RegionName": "eu-central-1"}}]' \
    --region ap-south-1


# Configure streams
$ awslocal dynamodb update-table \
    --table-name global01 \
    --stream-specification '{"StreamEnabled": true, "StreamViewType": "NEW_IMAGE"}' \
    --region ap-south-1

# Ensure table is replicated
$ awslocal dynamodb describe-table --table-name global01 --query 'Table.StreamSpecification' --region eu-central-1
{
    "StreamEnabled": true,
    "StreamViewType": "NEW_IMAGE"
}

$ awslocal dynamodb describe-table --table-name global01 --query 'Table.StreamSpecification' --region ap-south-1
{
    "StreamEnabled": true,
    "StreamViewType": "NEW_IMAGE"
}

# List streams works in original region...
$ awslocal dynamodbstreams list-streams --region ap-south-1
{
    "Streams": [
        {
            "StreamArn": "arn:aws:dynamodb:ap-south-1:000000000000:table/global01/stream/2022-12-30T11:52:55.566",
            "TableName": "global01",
            "StreamLabel": "2022-12-30T11:52:55.566"
        }
    ]
}

# ... but no in replicated regions
$ awslocal dynamodbstreams list-streams --region eu-central-1
{
    "Streams": []
}
```

deepak2431:
Can I take up this issue?

anekar3416:
Are Kinesis streams supported for Replicated tables?

HarshCasper:
Due to the forthcoming changes in our distribution and development model ([read more](https://blog.localstack.cloud/the-road-ahead-for-localstack/)), this issue tracker is no longer being maintained, and this issue is now being closed. If you'd like us to take a look or need further support, please reach out through one of our support channels:

[...]

PR #12641, opened 2025-05-19
DDB Global table: add logic for streams

giograno:
## Motivation
In LocalStack, we support global DynamoDB tables by keeping track of the region of the original table and its replicas. Whenever we receive a request for a replica table, we redirect the request to the original region. Therefore, we do not really replicate any data, unlike AWS.

Global tables also come with associated streams. According to the [docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/V2globaltables_HowItWorks.html#global-tables-streams), _"each global table produces an independent stream based on all its writes, regardless of the origination point for those writes"_.
However, we did not implement any redirect logic for DynamoDB Streams, as the one I just described for DDB.

## Changes
- Implementing the redirect logic for the DDBStreams of global tables.
- Adding a dictionary to the provider class, mapping the emitted shard iterators to the original regions (in case of a global table). This allows us to get the records that are kept in the original region.
- Adding a validated test with snapshots.

Changes have been implemented to both v1 and v2 providers.

### DIFF
--- a/localstack-core/localstack/services/dynamodb/provider.py
+++ b/localstack-core/localstack/services/dynamodb/provider.py
@@ -805,11 +805,6 @@ class DynamoDBProvider(DynamodbApi, ServiceLifecycleHook):
                 # Terraform depends on this parity for update operations
                 gsi["ProvisionedThroughput"] = default_values | gsi.get("ProvisionedThroughput", {})
 
-        if global_table_region != context.region:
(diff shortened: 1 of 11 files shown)


viren-nadkarni:
@giograno Would this fix https://github.com/localstack/localstack/issues/7405? Could you run a quick check?

PR #12653, opened 2025-05-21
Resolve non-subdomain host prefixes to LocalStack

joe4dev:
Depends: on public DNS entries for the dash-prefixed domains deployed (e.g., `data-localhost.localstack.cloud`)

## Motivation

The botocore specs include 562 operations with prefixed endpoints such as `api.` or `data-` and non-subdomains (i.e., `-` dash-prefixed operations) require special consideration.
These operations are defined in the botocore specs with a `hostPrefix`. Example for [DiscoverInstances](https://docs.aws.amazon.com/cloud-map/latest/api/API_DiscoverInstances.html) in CloudMap: `"endpoint":{"hostPrefix":"data-"}`

Example of a failing AWS CLI command due to missing DNS resolution:

```bash
aws --profile localstack servicediscovery discover-instances --namespace-name my-namespace --service-name my-service

Could not connect to the endpoint URL: "https://data-localhost.localstack.cloud:4566/"
```

To fix this problem, we need to:
a) deploy public DNS entries for these currently 11 prefixes
b) add them to `NAME_PATTERNS_POINTING_TO_LOCALSTACK` to ensure they get resolved to LocalStack by our [DNS server](https://docs.localstack.cloud/user-guide/tools/dns-server/)

## Changes

* Add non-subdomain prefixes to `NAME_PATTERNS_POINTING_TO_LOCALSTACK` such that these are resolved to LocalStack.
* Add a unit test to catch potential future additions of host prefixes by comparing the static list to the botocore specs
* Extend the bootstrap test `test_resolve_localstack_host` to cover the `hostPrefix` API operation scenario
* Add an integration test for a `hostPrefix` operation for Lambda in `test_lambda_host_prefix_api_operation`

## Discussion

Why a static list?

I think it's clearer (and faster) to maintain a static list such that we have full control and easier visibility into which name patterns get resolved to LocalStack. It also helps detecting changes, which require manual action by deploying a new DNS entry into our public DNS.

### DIFF
--- a/tests/unit/test_dns_server.py
+++ b/tests/unit/test_dns_server.py
@@ -6,9 +6,11 @@ import pytest
 
 from localstack import config
 from localstack.aws.spec import iterate_service_operations
+from localstack.constants import LOCALHOST_HOSTNAME
 from localstack.dns.models import AliasTarget, RecordType, SOARecord, TargetRecord
 from localstack.dns.server import (
     HOST_PREFIXES_NO_SUBDOMAIN,
+    NAME_PATTERNS_POINTING_TO_LOCALSTACK,
     DnsServer,
     add_resolv_entry,
     get_fallback_dns_server,
@@ -481,4 +483,9 @@ class TestDnsUtils:
                 unique_prefixes.add(operation.endpoint["hostPrefix"])
 
         non_dot_unique_prefixes = [prefix for prefix in unique_prefixes if not prefix.endswith(".")]
+        # Intermediary validation to easily summarize all differences
         assert set(HOST_PREFIXES_NO_SUBDOMAIN) == set(non_dot_unique_prefixes)
+
+        # Real validation of NAME_PATTERNS_POINTING_TO_LOCALSTACK
+        for host_prefix in non_dot_unique_prefixes:
+            assert f"{host_prefix}{LOCALHOST_HOSTNAME}" in NAME_PATTERNS_POINTING_TO_LOCALSTACK


tests/unit/test_dns_server.py
```diff
@@ -460,3 +466,19 @@ def test_no_resolv_conf_overwriting_on_host(self, tmp_path: Path, monkeypatch):
 
         assert "nameserver 127.0.0.1" not in new_contents.splitlines()
         assert "nameserver 127.0.0.11" in new_contents.splitlines()
+
+    def test_host_prefix_no_subdomain(
```
simonrw, line 470: Thank you for adding this test step, this will be great for preventing regressions and making sure we handle new services!

tests/unit/test_dns_server.py
```diff
@@ -460,3 +466,19 @@ def test_no_resolv_conf_overwriting_on_host(self, tmp_path: Path, monkeypatch):
...
+        1) Add the new entry to `HOST_PREFIXES_NO_SUBDOMAIN` to reflect any changes
+        2) IMPORTANT: Add a public DNS entry for the given host prefix!
+        """
+        unique_prefixes = set()
+        for service_model, operation in iterate_service_operations():
+            if operation.endpoint and operation.endpoint.get("hostPrefix"):
+                unique_prefixes.add(operation.endpoint["hostPrefix"])
+
+        non_dot_unique_prefixes = [prefix for prefix in unique_prefixes if not prefix.endswith(".")]
+        assert set(HOST_PREFIXES_NO_SUBDOMAIN) == set(non_dot_unique_prefixes)
```
simonrw, line 484: **suggestion(minor)**: I think the real test that would give me confidence is that if everything in `non_dot_unique_prefixes` is also included in `NAME_PATTERNS_POINTING_TO_LOCALSTACK` as this is the complete list of hostnames that target LocalStack. There may be some subsequent manipulation of `HOST_PREFIXES_NO_SUBDOMAIN` that affect the DNS behaviour.

simonrw (approved):
Thanks for adding this implementation. It has been a problem that crops up every now and again but has never been a high priority. Now we can handle all AWS operations with their strange domain name prefixes!

tests/unit/test_dns_server.py
joe4dev: Good point 🎯 . Thanks for raising this @simonrw 👍 

I added a "real validation" of `NAME_PATTERNS_POINTING_TO_LOCALSTACK` in addition to the set comparison, which allows for easy one-glance diffs.

PR #12654, opened 2025-05-22
Include reserved_concurrent_executions in get_function response

k-a-il:
## Motivation

This PR fixes a bug reported by a user. The issue was that when a Lambda function was created in Localstack using Terraform, the `reserved_concurrent_executions` value always reset to -1 in tfstate after apply.

While trying to reproduce the issue, I saw that the `reserved_concurrent_executions` value was saved correctly in the created lambda, but  `reserved_concurrent_executions` wasn’t returned in the [`AWSLambda::GetFunction` response.](https://docs.aws.amazon.com/lambda/latest/api/API_GetFunction.html) Because the response didn’t include the value, [Terraform set the value to default -1 ](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#reserved_concurrent_executions-1) and saved that in the tfstate file.

## Changes

Added code to return the function's `reserved_concurrent_executions` value if it is set

## Testing

- Added new integration test to test that after `reserved_concurrent_executions` value is updated it is also being returned in response of `AWSLambda::GetFunction`

### DIFF
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -1543,7 +1543,7 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
             )
         concurrency = None
         if fn.reserved_concurrent_executions:
-            concurrency = GetFunctionConcurrencyResponse(
+            concurrency = Concurrency(
                 ReservedConcurrentExecutions=fn.reserved_concurrent_executions
             )
 
@@ -1552,8 +1552,8 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                 version, return_qualified_arn=bool(qualifier), alias_name=alias_name
             ),
             Code=code_location,  # TODO
-            **additional_fields,
             Concurrency=concurrency,
+            **additional_fields,
         )
 
     def get_function_configuration(
--- a/localstack-core/localstack/services/lambda_/provider.py
+++ b/localstack-core/localstack/services/lambda_/provider.py
@@ -1541,6 +1541,11 @@ class LambdaProvider(LambdaApi, ServiceLifecycleHook):
                 RepositoryType=image.repository_type,
                 ResolvedImageUri=image.resolved_image_uri,
             )
+        concurrency = None
+        if fn.reserved_concurrent_executions:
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1541,14 +1541,19 @@ def get_function(
...
+            )
 
         return GetFunctionResponse(
             Configuration=api_utils.map_config_out(
                 version, return_qualified_arn=bool(qualifier), alias_name=alias_name
             ),
             Code=code_location,  # TODO
             **additional_fields,
-            # Concurrency={},  # TODO
+            Concurrency=concurrency,
```
dfangl, line 1556: Could we move this one line up? I think we generally do **fields in the end in lambda.

tests/aws/services/lambda_/test_lambda.py
```diff
@@ -2286,6 +2286,42 @@ def test_lambda_concurrency_crud(self, snapshot, create_lambda_function, aws_cli
...
+        assert (
+            updated_concurrency_result["ReservedConcurrentExecutions"] == new_reserved_concurrency
+        )
+
+        function_concurrency_info = aws_client.lambda_.get_function(FunctionName=func_name)
+        assert function_concurrency_info["Concurrency"] is not None
+        assert (
+            function_concurrency_info["Concurrency"]["ReservedConcurrentExecutions"]
+            == new_reserved_concurrency
+        )
```
dfangl, line 2316: Could we snapshot this as well?

localstack-core/localstack/services/lambda_/provider.py
```diff
@@ -1541,14 +1541,19 @@ def get_function(
                 RepositoryType=image.repository_type,
                 ResolvedImageUri=image.resolved_image_uri,
             )
+        concurrency = None
+        if fn.reserved_concurrent_executions:
+            concurrency = GetFunctionConcurrencyResponse(
```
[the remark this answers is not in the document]
k-a-il: yes, my mistake, will fix

dfangl (approved):
Thanks for addressing my comments! LGTM!

PR #12656, opened 2025-05-22
add small fixes/improvements to Firehose.CreateDeliveryStream

pinzon:
## Motivation
This PR adds small fixes for issues reported by @kostas-localstack internally.

1.- The operation allows the user to attempt to create a second DeliveryStream with a name already in use.
2.- The `DirectPut` is the default value for `DeliveryStreamType` when is not defined.
3.- When creating a DeliveryStream with Kinesis as destination and it fails. The resource is still created.

## Changes
- Add assertion that the name is not in use before attempting to create a DeliveryStream.
- Add `DirectPut` as default value.
- Move the storage of the resource to the end of the operation.


viren-nadkarni (approved):
Changes look good but I would have loved to see some AWS validated tests!

PR #12657, opened 2025-05-23
re-introduce eager service loading

bentsku:
## Motivation
It seems we deleted eager service loading behavior with #11202 by accident.

Now with the new runtime framework, we can simply eager load services via hooks. 

I've also added a proper bootstrap test to avoid deleting the behavior again without realizing it. 

I updated the eager loading logic to not eager load all services if set, as it will most likely just wait forever.

My biggest question is where this hook declaration should live? 

## Changes
- re-add eager loading via hooks
- add bootstrap tests
- add eager loading logic to not eager load all possible services


pinzon (commented):
Interesting how we lost a feature and no one noticed :sweat_smile:

tests/bootstrap/test_service_loading.py
```diff
@@ -0,0 +1,141 @@
...
+    assert services
+    assert all(services.get(key) == "disabled" for key in services.keys())
+
+
+def test_eager_and_strict_service_loading(
+    container_factory: ContainerFactory,
+    wait_for_localstack_ready,
+    aws_client_factory,
+):
+    # this is undocumented behavior, to allow eager loading of specific services while not restricting services loading
```
jeremygiberson, line 115: FYI I arrived at this PR because I was specifically trying to figure out how to eager load specific services in a specific order without loading everything or preventing services from ever being enabled. Lucky issue searching turned up this PR.

My use case is building integration tests that have isolated localstack containers and building/deploying images to ecr for lambda as well as spinning up RDS. I ended up getting tripped up because depending on whether RDS or ECR was started first, ECR might be 4510 or 4511. 

All that is to say, at least one person (me) finds this behavior useful and would appreciate the behavior being documented! 

Looking forward to see this PR land either way.

tests/bootstrap/test_service_loading.py
```diff
@@ -0,0 +1,141 @@
...
+    services = response.json().get("services")
+
+    # sqs should be running now
+    assert services.get("sqs") == "running"
+    assert services.get("s3") == "available"
+    assert services.get("sns") == "available"
+    assert services.get("cloudwatch") == "disabled"
+
+
+def test_eager_service_loading(
```
dominikschubert, line 77: nice 👏  thanks for adding these.

Might also be good to add a case for `{"EAGER_SERVICE_LOADING": "0", "SERVICES": "s3,sqs,sns"}` just to make sure we have no regression there.

tests/bootstrap/test_service_loading.py
dominikschubert: @jeremygiberson please keep in mind though that we still don't technically guarantee that these port assignments would be deterministic.

bentsku:
> As far as I understood, this would be a major change though, since it would break how users expect EAGER_SERVICE_LOADING=1 to work in isolation. 🤔 Did you find any occurence where we communicated this differently?

@dominikschubert this is fair. Knowing that we lost the feature for almost a year now and had a major release since, I thought that this would be a fair change that we could document going forward. I had a ~PR~ branch* ready in the documentation that `EAGER_SERVICE_LOADING=1` would need to be used in conjunction with `SERVICES` (didn't open the PR yet as I wasn't sure about this PR). 
Technically not a "breaking" change since using `EAGER_SERVICE_LOADING=1` without `SERVICES` would do nothing, like it is doing right now... 😄 

I'm not sure we ever communicated about `EAGER_SERVICE_LOADING` in isolation, at least I'm not aware of such communication, I think we implemented eager loading after the new implementation of lazy loading of services, so it was done when people were still using the `SERVICES` variable. I'm not sure we ever officially supported eager loading in isolation, and if it ever really worked. 

I tried to start LocalStack with `EAGER_SERVICE_LOADING=1` and no `SERVICES` while working on this change, and it never succeeded to start for me, having so many services starting at once is... incredibly expensive, some services will start Postgres or start downloading their dependencies, one after the other. 

I'm also fine to leave the behavior as it was before, and prepare this change for `5.0`.

tests/bootstrap/test_service_loading.py
bentsku: I think this would be covered by the test above: `test_strict_service_loading`. `STRICT_SERVICE_LOADING` is set to `1` by default in LocalStack, and `EAGER_SERVICE_LOADING` to 0. 

If something were to change with eager loading, I believe the test above would start failing as we're asserting that the services are `available` and not `running`? But we could make that explicit as well, here we are depending on default values.

dominikschubert:
@bentsku Even though it's extremely expensive, I'd argue we should keep the behavior for now since someone might depend on it. We can discuss next week and onward if we want to change that behavior and if we want to do so via a major or minor release. Protecting users from potentially unintended mistakes is of course great, but the use case of trying to make sure everything is up and running is a valid one as well :man_shrugging: , so we need to consider that in a bit more detail. 

In the meantime we can also investigate why it's never actually starting up when eagerly loading all services. I'd argue that's a bug to address separately :lady_beetle:

bentsku:
@dominikschubert got it, I've updated the PR now 👍

I see it is as half a bug and half a feature, as starting each individual service will take a long time, maybe I didn't have a lot of patience either 😄 but not sure we have users using every single service we have, but that's another question! We can keep this discussion offline.  Thanks for the review 🙏

dominikschubert (approved):
LGTM, thanks for getting this regression fixed and addressing my earlier concerns :+1:

thrau:
just reading the title i thought we're moving away from lazy loading. before i had all the context my heart rate went up significantly, lol.

but this is a great PR, thanks for fixing this @bentsku! the implementation is really clean, i think the hook you added is one of the cleanest ways to do this :+1:

PR #12661, opened 2025-05-26
EC2: fix describe-availability-zones filtering

anisaoshafi:
## Motivation
Customer reported some issue trying to run a tf script where they were getting this error: `Error: no matching EC2 Availability Zone found`.
The reason for that is that this operation implementation in the ec2 provider.py filters doesn't work against moto, resulting into an empty list.

image

More context in the notion [ticket](https://www.notion.so/localstack/Oleg-Tyaglo-No-matching-EC2-Availability-Zone-found-1fcfc2a23431817abe14c63939ee03f7?d=173fc2a234318207919e83a394a8a046).

## Changes
Change how filters are passed for the call done to moto.
Added aws validated tests.
`test_describe_availability_zones_filter_with_zone_names` fails against LS with the old code, and passes now.


anisaoshafi:
FYI: The ci check failure is unrelated to these changes, due to: test_sqs_event_source_mapping_batch_size_override.

PR #12662, opened 2025-05-26
CloudFormation v2 Engine: Base Support for Fn::Transform

MEPalma:
\* Extends https://github.com/localstack/localstack/pull/12660

## Motivation
The introduction of the CloudFormation v2 engine laid the foundation for a redesigned engine capable of accurately determining update requirements between CloudFormation deployments, while also enabling parallel execution during updates. However, the current implementation offers no support for `Fn::Transform`.

## Changes
- Add base support for `Fn::Transform`
- Port v1 tests for `Fn::Transform`


tests/aws/services/cloudformation/v2/ported_from_v1/engine/test_mappings.py
```diff
@@ -249,6 +249,7 @@ def test_mapping_ref_map_key(self, deploy_cfn_template, aws_client, map_key, sho
 
             aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
 
+    # @pytest.mark.skip(reason="CFNV2:Mappings")
```
simonrw, line 252:
```suggestion
```

tests/aws/services/cloudformation/v2/ported_from_v1/test_template_engine.validation.json
```diff
@@ -0,0 +1,107 @@
+{
+  "tests/aws/services/cloudformation/test_template_engine.py::TestImportValues::test_cfn_with_exports": {
```
simonrw, line 2: Did you just copy the validations over? These paths are wrong. I see the snapshots have not been updated as well. I don't mind not revalidating (though it would be nice!) but they should at least be consistent so revalidating in the future will change rather than add new validation items.

simonrw (approved):
Just a comment about validations, if we are not going to revalidate then we should at least copy the snapshot/validation files correctly

tests/aws/services/cloudformation/v2/ported_from_v1/test_template_engine.validation.json
MEPalma: Thanks, indeed these should be ported like in https://github.com/localstack/localstack/pull/12660

Issue #12670, opened 2025-05-28
bug: SNS cross region topic subscription fails (CloudFormation)


codenem:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Starting version `4.1.0`, subscribing to a topic from a different region fails with the following:
```
An error occurred (NotFound) when calling the Subscribe operation: Topic does not exist
```
I think the bug has been introduced by https://github.com/localstack/localstack/pull/12175, with the intention of raising the error for legit cases.

The code might be pretty self explanatory of the bug. If I understand correctly, [here](https://github.com/localstack/localstack/pull/12175/files#diff-b8d395ba7fcc226f2dd13eeaf575a66dd0f93e2fceff296ccf1d56ffc719cd49R652) we should check the store corresponding to the region of the topic ARN we want to subscribe **to**, not the store of the region we are subscribing **from**.

### Expected Behavior

Subscribing to a topic from a different region succeeds.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```yaml
services:
  localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
    image: localstack/localstack-pro:4.1.0 # required for Pro
    ports:
      - "127.0.0.1:4566:4566" # LocalStack Gateway
      - "127.0.0.1:4510-4559:4510-4559" # external services port range
      - "127.0.0.1:443:443" # LocalStack HTTPS Gateway (Pro)
    environment:
      # Activate LocalStack Pro: https://docs.localstack.cloud/getting-started/auth-token/
      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN:?} # required for Pro
      # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
      - DEBUG=${DEBUG:-0}
      - PERSISTENCE=${PERSISTENCE:-1}
      - ENFORCE_IAM=0
      - SNAPSHOT_LOAD_STRATEGY=ON_STARTUP
      - SNAPSHOT_SAVE_STRATEGY=ON_SHUTDOWN
    volumes:
      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```bash
[...]

PR #12673, opened 2025-05-28
SNS: validate cross-region behavior

bentsku:
## Motivation
We've got a report with #12670 that we didn't support cross-region subscribe calls. It seems it is not allowed by AWS, but we didn't have test coverage over it. 

This PR adds more AWS validated tests testing cross-region behavior for operations that would allow cross-account access, as well as a cross-region delivery scenario. 

## Changes
- add tests around cross-region access in SNS
- add test for `DeleteTopic` idempotency
- fix behavior in provider, to raise proper error message in case of cross-region access

### DIFF
--- a/tests/aws/services/sns/test_sns.py
+++ b/tests/aws/services/sns/test_sns.py
@@ -192,9 +192,6 @@ class TestSNSTopicCrud:
             aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
         snapshot.match("topic-not-exists", e.value.response)
 
-        delete_topic = aws_client.sns.delete_topic(TopicArn=topic_arn)
-        snapshot.match("delete-topic-not-exists", delete_topic)
-
     @markers.aws.validated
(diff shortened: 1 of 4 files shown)


tests/aws/services/sns/test_sns.py
```diff
@@ -4427,6 +4446,127 @@ def get_messages_from_queues(message_content: str):
...
+        snapshot.match("get-topic-attrs", e.value.response)
+
+        with pytest.raises(ClientError) as e:
+            sns_region2_client.publish(TopicArn=topic_arn, Message="hello world")
+        snapshot.match("cross-region-publish-forbidden", e.value.response)
+
+        with pytest.raises(ClientError) as e:
+            sns_region2_client.subscribe(
+                TopicArn=topic_arn, Protocol="email", Endpoint="devil@hell.com"
+            )
```
baermat, line 4488: nice email :smile:

bentsku: I copy pasted it from another test, I hadn't even realized 😂

PR #12678, opened 2025-05-28
S3: add heuristic for S3 requests from IaC with no prefix

bentsku:
## Motivation
TBD

## Changes
TBD


bentsku:
Closing as it has significant drawback, but will keep in mind if the need is there

PR #12683, opened 2025-05-29
CI: update CFn test name

simonrw:
## Motivation

While migrating to GitHub Actions, the name for the CloudFormation test suite was incorrectly set to "CLoudFront".

## Changes

* Update the test name to reflect the test nature

PR #12686, opened 2025-05-29
GH Actions: fix order of path filters for push and pull-request events

k-a-il:
## Motivation

Thanks to [this PR](https://github.com/localstack/localstack/pull/12683), it was discovered that the order of path filters for push and pull_request triggers was incorrect. As a result, the `aws-main.yml` workflow was not triggered when `aws-tests.yml` was updated in the PR.

## Changes

- Fix path ordering not to trigger a workflow for all `.github` files except files in the `actions` folder, `aws_main.yml`, and `aws_tests.yml`

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -6,6 +6,7 @@ on:
   push:
     paths:
       - '**'
+      - '!.github/**'
       - '.github/actions/**'
       - '.github/workflows/aws-main.yml'
       - '.github/workflows/aws-tests.yml'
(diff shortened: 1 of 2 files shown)


simonrw:
This seems like a very strange fix for this error. Could you explain more how the old evaluation logic worked, and how this change fixes it?

simonrw (approved):

k-a-il:
> This seems like a very strange fix for this error. Could you explain more how the old evaluation logic worked, and how this change fixes it?

Yes, indeed. The problem is that order of path patterns matter and it was applied in the incorrect order. Now, we firstly exclude all files in the `.github` folder and after that include only some we are interested in, such as `aws_main.yml`. [Reference to GH docs.](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore)

PR #12689, opened 2025-05-29
add testing for Fn::Transform in CFn v2

pinzon:
## Motivation
This PR adds a collection of tests to validate the behavior of intrinsic transformation in a template. The tests are due to be unskipped when the feature is implemented in another PR.

## Testing
-  New AWS validated tests to confirm that the function works with change sets.


simonrw (commented):
I think it would be nice to add some tests that cover _changing_ `Fn::Transform` usages, so it would exist in the first template, then the arguments change for the second template. WDYT?

simonrw (approved):
Thanks for capturing these tests, I'm excited to see what the implementation will be in the future!

I am not super keen on _adding_ skips to the test suite, I would have expected this PR to be purely additive, however we are going to overhaul it so maybe starting from scratch is a good idea.

PR #12694, opened 2025-05-30
APIGW: implement Canary Deployments CRUD logic

bentsku:
## Motivation
We've got a request to support Canary Deployments CRUD logic as it is needed in some Terraform deployments. 

This was on a radar for a while, and there was no support for it in Moto. 
This PR only implements the CRUD support for it, a follow-up PR will implement the actually canary deployments logic during invocation, with the ability to direct requests to either deployment. The test for it is already in this PR just to verify how it worked in AWS (with `test_invoking_canary_deployment`). 

I'm conscious that this implementation is... not great. There are so many patches, the update logic is really complex. Those JSON Patches really don't make this easy. We should probably rework some kind of framework to easily handle validation + application of those. 

At first, I didn't want to duplicate the code, and I couldn't insert it from the child class inside `update_stage`. I had done in the legacy provider, but I don't want to break existing things and have partial things implemented so I moved it back to the next gen/current provider. This would have made the diff simpler, sorry!

The logic implemented is documented in AWS here: https://docs.aws.amazon.com/apigateway/latest/developerguide/canary-release.html

There was also a bug in Moto where it would update your stage deployment id to the one of the canary setting if you were to set one due to greedy check. 

## Changes
- update operations related to canary settings: `CreateStage`, `UpdateStage`, `CreateDeployment`
- add a lot of tests around the behavior to make sure we're getting things somewhat right

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/provider.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/provider.py
@@ -174,12 +174,13 @@ class ApigatewayNextGenProvider(ApigatewayProvider):
                 }
                 default_settings.update(canary_settings)
                 moto_stage.canary_settings = default_settings
+            else:
+                moto_stage.canary_settings = None
 
             if variables:
                 moto_stage.variables = variables
 
-            if stage_description:
-                moto_stage.description = stage_description
+            moto_stage.description = stage_description
 
             if cache_cluster_enabled is not None:
                 moto_stage.cache_cluster_enabled = cache_cluster_enabled
--- a/localstack-core/localstack/services/apigateway/next_gen/provider.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/provider.py
@@ -1,5 +1,6 @@
 from localstack.aws.api import CommonServiceException, RequestContext, handler
 from localstack.aws.api.apigateway import (
+    BadRequestException,
     CacheClusterSize,
     CreateStageRequest,
     Deployment,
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/apigateway/next_gen/provider.py
```diff
@@ -66,13 +75,35 @@ def delete_rest_api(self, context: RequestContext, rest_api_id: String, **kwargs
 
     @handler("CreateStage", expand=False)
     def create_stage(self, context: RequestContext, request: CreateStageRequest) -> Stage:
-        response = super().create_stage(context, request)
+        # TODO: we need to internalize Stages and Deployments in LocalStack, we have a lot of split logic
```
bentsku, line 78: note: most of this code is already taken from the parent class, sorry it does not make a clean diff

localstack-core/localstack/services/apigateway/next_gen/provider.py
```diff
@@ -84,20 +115,121 @@ def update_stage(
         patch_operations: ListOfPatchOperation = None,
         **kwargs,
     ) -> Stage:
-        response = super().update_stage(
-            context, rest_api_id, stage_name, patch_operations, **kwargs
-        )
+        moto_rest_api = get_moto_rest_api(context, rest_api_id)
```
bentsku, line 118: note: most of this code is already taken from the parent class, sorry it does not make a clean diff

PR #12695, opened 2025-05-31
Apigw canary logic

bentsku:
## Motivation
Based on #12694 

This is the actual implementation of canary dispatching between deployments. 

Not a lot to add, the code is pretty descriptive. I wasn't sure where to add the dispatching logic, but I think getting this in the router somewhat makes sense, as it is where we are deciding which deployment to use. One of the limitation is that we only know the account id and region from the active deployment itself. 

We could also add a handler first in the chain to fetch the stage config and update the deployment.

Also added a small fix in `UpdateIntegrationResponse`, this will be a good revamp at some point 😅 

## Changes
- enable the test
- implement the logic in the invocation layer, adding some fields in the context to support the use case

### DIFF
--- a/localstack-core/localstack/services/apigateway/legacy/provider.py
+++ b/localstack-core/localstack/services/apigateway/legacy/provider.py
@@ -622,7 +622,7 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
                 param = param.replace("~1", "/")
                 if op == "remove":
                     integration_response.response_templates.pop(param)
-                elif op == "add":
+                elif op in ("add", "replace"):
                     integration_response.response_templates[param] = value
 
             elif "/contentHandling" in path and op == "replace":
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/context.py
@@ -5,7 +5,7 @@ from rolo import Request
 from rolo.gateway import RequestContext
 from werkzeug.datastructures import Headers
 
-from localstack.aws.api.apigateway import Integration, Method, Resource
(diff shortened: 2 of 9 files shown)


localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -622,7 +622,7 @@ def update_integration_response(
                 param = param.replace("~1", "/")
                 if op == "remove":
                     integration_response.response_templates.pop(param)
-                elif op == "add":
+                elif op in ("add", "replace"):
```
cloutierMat, line 625: question: What is this fixing? Is it a leftover from the previous pr?

cloutierMat (approved):
Nice and clean. Too bad we still need more moto helpers here, but thank you for the constant effort to keep it organized

localstack-core/localstack/services/apigateway/legacy/provider.py
bentsku: Somewhat, this is because now that the test for the actual logic has been enabled, the test couldn't work because the `update_integration_response` method was not properly updating...
Here in `TestCanaryDeployments.test_invoking_canary_deployment`:
```python
        aws_client.apigateway.update_integration_response(
            restApiId=api_id,
            resourceId=resource_id,
            httpMethod="GET",
            statusCode="200",
            patchOperations=[
                {
                    "op": "replace",
                    "path": "/responseTemplates/application~1json",
                    "value": json.dumps(
                        {
                            "statusCode": 200,
                            "message": "canary deployment",
                            "variable": "$stageVariables.testVar",
                            "nonExistingDefault": "$stageVariables.noStageVar",
                            "nonOverridden": "$stageVariables.defaultVar",
                            "isCanary": "$context.isCanaryRequest",
                        }
                    ),
                }
            ],
        )
```

So I just made that work for now 😅

PR #12704, opened 2025-06-03
refactor: reduce max number of allowed labels in counters from 8 to 6 [DAT-146]

vittoriopolverino:
## Motivation
- After reviewing label usage over the past few months, we found that no existing metric uses more than 5 labels.
- Limiting counters to 6 labels help to avoid overly complex metrics that try to capture too many different aspects.
- Fewer labels simplify analytics query and makes dashboards easier to build/read.
- A limit of 6 is sufficiently high for current needs, but this threshold can be revisited in the future if necessary.


vittoriopolverino:
@joe4dev not sure who to tag for review. I used to loop in Alex, but he's on leave. Hope you don’t mind! The change is quite minimal 🙏🏼

ackdav (approved):
:shipit:

PR #12706, opened 2025-06-03
CloudFormation v2 Engine: Base Support for Fn::And Fn::Or and Condition Bindings

MEPalma:
## Motivation
The introduction of the CloudFormation v2 engine laid the foundation for a redesigned engine capable of accurately determining update requirements between CloudFormation deployments, while also enabling parallel execution during updates. However, the current implementation does not support `Fn::And`, `Fn::Or`, nor `{"Condition": "<identifer>"}` bindings.

## Changes
- Add support for update modeling/describe/execution of `Fn::And`, `Fn::Or`, and condition bindings
- Unblock related tests ported from v1
- Unrelated maintenance on ported tests from v1

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -219,8 +219,8 @@ jobs:
 
       - name: Push ${{ env.PLATFORM_NAME_AMD64 }} Docker Image
         env:
-          DOCKER_USERNAME: ${{ secrets.DOCKERHUB_PUSH_USERNAME }}
-          DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_PUSH_TOKEN }}
+          DOCKER_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
           PLATFORM: ${{ env.PLATFORM_NAME_AMD64 }}
         run: |
           # Push to Docker Hub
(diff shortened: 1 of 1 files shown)


tests/aws/services/cloudformation/v2/ported_from_v1/engine/test_mappings.py
```diff
@@ -249,7 +249,6 @@ def test_mapping_ref_map_key(self, deploy_cfn_template, aws_client, map_key, sho
 
             aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
 
-    # @pytest.mark.skip(reason="CFNV2:Mappings")
```
simonrw, line 252: 🎉

tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_events.py
```diff
@@ -18,7 +18,9 @@
 )
 
 
-@pytest.mark.skip(reason="CFNV2:ReferenceDotSyntax")
+@pytest.mark.skip(
+    reason="CFNV2:Other resource name conflict with another test case resource in this suite"
+)
```
simonrw, line 23: I guess because the deletion is not implemented yet, so we are leaving resources lying around.

tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_kinesis.py
```diff
@@ -169,7 +169,7 @@ def test_dynamodb_stream_response_with_cf(deploy_cfn_template, aws_client, snaps
     snapshot.add_transformer(snapshot.transform.key_value("TableName"))
 
 
-@pytest.mark.skip(reason="CFNV2:ReferenceDotSyntax")
+@pytest.mark.skip(reason="CFNV2:Other")
```
simonrw: **nit**: it would be nice when classifying skips as "Other" to add a little more detail (as you do above) on the actual failure, so we can look for common patterns.

simonrw (approved):
Nice work!

PR #12709, opened 2025-06-04
Add missing switch for forcing ARM tests on workflow dispatch

silv-io:
## Motivation

The GHA workflow only runs ARM tests on master and the dependency upgrade PRs, or when `input.forceARMTests` is set to true. However, we do not have that input. This PR adds that

## Changes

- Add `forceARMTests` input to both the main pipeline as well as the test sub-pipeline
- Fix minor typo in Acceptance test job name

## Testing

Test run with force ARM tests:
https://github.com/localstack/localstack/actions/runs/15441275724

## TODO

What's left to do:

- [ ] Fix workflow summary reporting after merging this (#12707)
- [ ] ...

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -36,6 +36,11 @@ on:
         required: false
         type: boolean
         default: false
+      forceARMTests:
+        description: 'Run the ARM tests'
+        required: false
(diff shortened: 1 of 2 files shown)


silv-io:
Merging because the tests are not affected by this

PR #12710, opened 2025-06-04
GH Actions:change test selection parameter

k-a-il:
## Motivation

This PR corrects the condition for applying test selection. Test selection should be enabled for pull requests, while it remains disabled by default for manual and scheduled runs.

## Changes

- Changed condition of parameter `disableTestSelection` which is propagated to `aws-tests.yml`

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -134,9 +134,6 @@ class TestKMS:
         assert response["KeyId"] == key_id
         assert f":{region_name}:" in response["Arn"]
         assert f":{account_id}:" in response["Arn"]
-        assert f":{region_name}:" in response["Arn"]
-        assert f":{account_id}:" in response["Arn"]
-        assert response["KeyId"] != "tst"
 
     @markers.aws.validated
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-main.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       # default "disableCaching" to `false` if it's a push or schedule event
       disableCaching: ${{ inputs.disableCaching == true }}
       # default "disableTestSelection" to `true` if it's a push or schedule event
-      disableTestSelection: ${{ inputs.enableTestSelection != true }}
+      disableTestSelection: ${{ (inputs.enableTestSelection != '' && inputs.enableTestSelection) || github.event_name == 'push' }}
```
k-a-il, line 79: `inputs.enableTestSelection` will be an empty string if the workflow is triggered automatically. I think this deifinition version `(inputs.enableTestSelection != '' && inputs.enableTestSelection)` makes our intentions clearer: use the input value if it's defined, or use the push event as a fallback.

silv-io (approved):
LGTM!

Issue #12711, opened 2025-06-04
feature request: Mocking for worflows started with StartSyncExecution in Step Functions


carole-lavillonniere:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Description

The current Step Functions mocking feature is limited to execution started with the `StartExecution` API (standard and express workflows). Example:
```
LOCALSTACK_SFN_MOCK_CONFIG=/tmp/MockConfigFile.json \
  localstack start --volume ./MockConfigFile.json:/tmp/MockConfigFile.json
```

Express workflows started with `StartSyncExecution` ignore the mocking configuration.

## Feature Request

Extend the Step Functions mocking capability to support `StartSyncExecution`, enabling consistent testing and simulation across both Standard and Express state machines.

HarshCasper:
Due to the forthcoming changes in our distribution and development model ([read more](https://blog.localstack.cloud/the-road-ahead-for-localstack/)), this issue tracker is no longer being maintained, and this issue is now being closed. If you'd like us to take a look or need further support, please reach out through one of our support channels:

-   💬 [GitHub Discussions](https://github.com/orgs/localstack/discussions)
-   📖 [Help & Support](https://docs.localstack.cloud/aws/getting-started/help-support/)

PR #12712, opened 2025-06-04
Step Functions: Mock feature for StartSyncExecution

carole-lavillonniere:
## Motivation
Extend the Step Functions mocking capability to support `StartSyncExecution`, enabling consistent testing across both Standard and Express state machines https://github.com/localstack/localstack/issues/12711

## Changes
- Apply mock config (`LOCALSTACK_SFN_MOCK_CONFIG`) when starting a state machine via `StartSyncExecution`
- New parity test `tests/aws/services/stepfunctions/v2/mocking/test_base_scenarios.py::TestBaseScenarios::test_lambda_service_invoke` to verify that the mocked service integration mechanism also works when using `StartSyncExecution`

## Testing
1. Create an express state machine
2. Create a mock config to mock the state machine response
3. Start localstack with the mock configuration
4. Start the state machine synchronously with `start-execution`
5. Verify that the execution response is the expected one

## TODO
- Add a parity test `tests/aws/services/stepfunctions/v2/mocking/test_base_callbacks.py::TestBaseScenarios::test_sqs_wait_for_task_token`  to verify that the callback mechanism (sync2) returns correctly when executed via `StartSyncExecution`

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/provider.py
+++ b/localstack-core/localstack/services/stepfunctions/provider.py
@@ -443,7 +443,7 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
             logging_configuration=state_machine_logging_configuration
         )
 
-        # CreateStateMachine is an idempotent API. Subsequent requests won’t create a duplicate resource if it was
+        # CreateStateMachine is an idempotent API. Subsequent requests won't create a duplicate resource if it was
         # already created.
         idem_state_machine: Optional[StateMachineRevision] = self._idempotent_revision(
             context=context,
@@ -656,7 +656,7 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
                     stateMachineAliasArn=state_machine_alias_arn, creationDate=alias.create_date
                 )
             else:
-                # CreateStateMachineAlias is an idempotent API. Idempotent requests won’t create duplicate resources.
+                # CreateStateMachineAlias is an idempotent API. Idempotent requests won't create duplicate resources.
                 raise ConflictException(
                     "Failed to create alias because an alias with the same name and a "
                     "different routing configuration already exists."
@@ -772,6 +772,33 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
                     raise TaskDoesNotExist()
         raise InvalidToken("Invalid token")
 
+    @staticmethod
+    def _extract_base_arn(state_machine_arn: str) -> str:
+        """Extract the base ARN from a state machine ARN by removing any test case suffix."""
+        return state_machine_arn.split("#")[0]
+
+    @staticmethod
+    def _get_mock_test_case(
+        state_machine_arn: str, state_machine_name: str
+    ) -> Optional[MockTestCase]:
+        """Extract and load a mock test case from a state machine ARN if present."""
+        parts = state_machine_arn.split("#")
+        if len(parts) != 2:
+            return None
+
+        mock_test_case_name = parts[1]
+        mock_test_case = load_mock_test_case_for(
+            state_machine_name=state_machine_name, test_case_name=mock_test_case_name
+        )
+        if mock_test_case is None:
+            raise InvalidName(
+                f"Invalid mock test case name '{mock_test_case_name}' "
+                f"for state machine '{state_machine_name}'."
+                "Either the test case is not defined or the mock configuration file "
+                "could not be loaded. See logs for details."
+            )
+        return mock_test_case
+
     def start_execution(
         self,
         context: RequestContext,
@@ -783,21 +810,16 @@ class StepFunctionsProvider(StepfunctionsApi, ServiceLifecycleHook):
     ) -> StartExecutionOutput:
         self._validate_state_machine_arn(state_machine_arn)
 
-        state_machine_arn_parts = state_machine_arn.split("#")
-        state_machine_arn = state_machine_arn_parts[0]
-        mock_test_case_name = (
-            state_machine_arn_parts[1] if len(state_machine_arn_parts) == 2 else None
-        )
-
+        base_arn = self._extract_base_arn(state_machine_arn)
         store = self.get_store(context=context)
 
-        alias: Optional[Alias] = store.aliases.get(state_machine_arn)
+        alias: Optional[Alias] = store.aliases.get(base_arn)
         alias_sample_state_machine_version_arn = alias.sample() if alias is not None else None
         unsafe_state_machine: Optional[StateMachineInstance] = store.state_machines.get(
-            alias_sample_state_machine_version_arn or state_machine_arn
+            alias_sample_state_machine_version_arn or base_arn
         )
         if not unsafe_state_machine:
-            self._raise_state_machine_does_not_exist(state_machine_arn)
+            self._raise_state_machine_does_not_exist(base_arn)
 
         # Update event change parameters about the state machine and should not affect those about this execution.
         state_machine_clone = copy.deepcopy(unsafe_state_machine)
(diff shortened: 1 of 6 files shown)


carole-lavillonniere:

localstack-core/localstack/services/stepfunctions/provider.py
```diff
@@ -772,6 +772,33 @@ def send_task_failure(
                     raise TaskDoesNotExist()
         raise InvalidToken("Invalid token")
 
+    @staticmethod
+    def _extract_base_arn(state_machine_arn: str) -> str:
```
MEPalma: There are multiple types of ARNs used in Step Functions, so the term “base ARN” might be a bit misleading. It might be clearer to move the logic for stripping the mock test case name into a different function, perhaps in _get_mock_test_case below. That way, we can both reuse the logic and make its purpose more immediately clear, returning the actual state machine ARN along with the mock test case name, if present. What do you think?

localstack-core/localstack/services/stepfunctions/provider.py
```diff
@@ -889,11 +899,13 @@ def start_sync_execution(
         **kwargs,
     ) -> StartSyncExecutionOutput:
         self._validate_state_machine_arn(state_machine_arn)
+
+        base_arn = self._extract_base_arn(state_machine_arn)
```
MEPalma: Following from the comment above, this could be reduced to a single statement with the one on line 943 (e.g. `base_arn, mock_test_case = <util>(state_machine_arn)`

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
```diff
@@ -507,6 +509,27 @@ def launch_and_record_mocked_execution(
     return execution_arn
 
 
+def launch_and_record_mocked_sync_execution(
```
MEPalma, line 512: Thank you for adding this utility!

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
```diff
@@ -587,10 +611,18 @@ def create_and_record_mocked_execution(
         sfn_snapshot,
         definition,
         state_machine_name=state_machine_name,
+        state_machine_type=state_machine_type,
     )
-    execution_arn = launch_and_record_mocked_execution(
-        target_aws_client, sfn_snapshot, state_machine_arn, execution_input, test_name
-    )
+    if state_machine_type == StateMachineType.EXPRESS:
```
MEPalma: There are STANDARD and EXPRESS state machines. LocalStack wants to enable the mock feature for both. However, when using the `StartSyncExecution` API action, only EXPRESS state machines can be executed, otherwise an error is raised. This utility seems to be limited to recording the executions through StartExecution only, as it invokes the asynchronous await logic. 
What do you think about reverting the changes in this function, and adding the snapshotting features for mocked executions through StartSyncExecution in your launch_and_record_mocked_sync_execution? Also considering create_and_record_express_sync_execution. This could also allow us to test for failures if a STANDARD with mocking through StartSyncExecution.

MEPalma (commented):
Thank you for contributing to this feature! I’m sure users relying on mocked service integrations will find it valuable. I’ve left a few comments suggesting some possible refactoring options for your consideration.

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
carole-lavillonniere: True! This is a use case I had forgotten.

Addressed here 78574bbaf1155674a563992560b620d90dc6e29f, let me know what you think.

I added a `create_and_record_mocked_sync_execution` that always uses `StateMachineType.EXPRESS`, and left the argument `state_machine_type` for `create_and_record_mocked_execution` to make it possible to test the combination of `StateMachineType.EXPRESS` and `StartExecution` in the future.

Again, just to make sure we are on the same page, this is my understanding of how it works:

| Workflow Type | Asynchronous (`StartExecution`) | Synchronous (`StartSyncExecution`) |
|---------------|-------------------------------|-----------------------------------|
| **Standard**  | ✅ Supported                   | ❌ Not supported                   |
| **Express**   | ✅ Supported                   | ✅ Supported                       |

localstack-core/localstack/services/stepfunctions/provider.py
carole-lavillonniere: I considered your suggestion first but decided for 2 functions instead of 1 to respect the single responsibility principle, since extracting the ARN and getting the mock test case seemed to be two unrelated tasks. I am not too opinionated about this though :)

Hence, I gave a try to your suggestion just now, and realised it's not super straightforward because:
- we need the extracted ARN to get the `state_machine_clone` (line 825)
- we need the `state_machine_clone.name` to get the `mock_test_case_name` (line 867)

-> we have a circular dependency.

How would you suggest to solve this? Can I get the state machine name some other way than through the clone?

As an alternative I would suggest to simply rename `_extract_base_arn` to `_get_state_machine_arn`.

Let me know what you think.

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
```diff
@@ -579,6 +602,32 @@ def create_and_record_mocked_execution(
...
+    state_machine_arn = create_state_machine_with_iam_role(
+        target_aws_client,
+        create_state_machine_iam_role,
+        create_state_machine,
+        sfn_snapshot,
+        definition,
+        state_machine_name=state_machine_name,
+        state_machine_type=state_machine_type,
+    )
+    execution_arn = launch_and_record_mocked_sync_execution(
```
MEPalma: Judging from the name of this utility seems like it should run executions asynchronously using `launch_and_record_mocked_execution`. A `sync` variant of this utility `create_and_record_mocked_sync_execution` would then use `launch_and_record_mocked_sync_execution`. Should we revert this change to use `launch_and_record_mocked_execution`?

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
```diff
@@ -579,6 +602,32 @@ def create_and_record_mocked_execution(
...
+        target_aws_client, sfn_snapshot, state_machine_arn, execution_input, test_name
+    )
+    return execution_arn
+
+
+def create_and_record_mocked_sync_execution(
```
MEPalma, line 622: Thank you!

localstack-core/localstack/services/stepfunctions/provider.py
MEPalma: Thank you for sharing your view on this! Semantically, this function removes the mock test case portion if one exists. The reason why I don't think referring to the output as a 'base arn' is ideal is because other types of arns in stepfuctions also append text to state machine arns (such as execution arns). I wouldn't want maintainers to assume they may rely on this function to strip those portions from other types of arns too. Renaming it in the context of the mock feature seems good!

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
MEPalma: Thank you, I left a new comment with a minor. 
Regarding the use of STANDARD and EXPRESS state machine, yes I would agree [I see `StartSyncExecution` should only be available for `EXPRESS` workflows](https://docs.aws.amazon.com/step-functions/latest/apireference/API_StartSyncExecution.html).

localstack-core/localstack/services/stepfunctions/provider.py
carole-lavillonniere: Renamed the function here fbabb132e793d8979ba92e9b37c354695579bc9e

localstack-core/localstack/testing/pytest/stepfunctions/utils.py
carole-lavillonniere: Absolutely! Wrong function call :see_no_evil: 3e9ac67272f62dcf6fd14ff917abad839ca553cb

PR #12713, opened 2025-06-04
test(esm/sqs): Skip flaky test_report_batch_item_failures test

gregfurman:
## Motivation
This test makes a faulty assumption that a `ReceiveMessages` call will always return a value. Skip for now but re-visit this test after release.

## Changes
- Skips flaky `tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py/test_report_batch_item_failures`

### DIFF
--- a/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py
+++ b/tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py
@@ -415,6 +415,9 @@ def test_sqs_queue_as_lambda_dead_letter_queue(
 
 
 @markers.aws.validated
+@pytest.mark.skip(
+    reason="Flaky as an SQS queue will not always return messages in a ReceiveMessages call."
+)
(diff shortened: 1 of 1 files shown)


tests/aws/services/lambda_/event_source_mapping/test_lambda_integration_sqs.py
```diff
@@ -538,6 +541,7 @@ def test_report_batch_item_failures(
     # now wait for the first invocation result which is expected to have processed message 1 we wait half the retry
     # interval to wait long enough for the message to appear, but short enough to check that the DLQ is empty after
     # the first attempt.
+    # FIXME: We cannot assume that the queue will always return a message in the given time-interval.
```
joe4dev, line 544: Maybe, we can use the collect sqs messages fixture doing retries. Good catch 👍

joe4dev (approved):
Good decision to skip 👍

PR #12716, opened 2025-06-04
Events: fix PutEvents triggering Scheduled rules

bentsku:
## Motivation
As reported in #12714, we would trigger scheduled rules when a call to `PutEvents` would be done (in this issue, the call would be internal from `SSM.PutParameter`. 

We need to skip triggering scheduled rules when processing entries. 

I validated that the user shared sample now works with this fix. 

## Changes
- add a skip when processing rules
- add a test validating that AWS does not trigger the scheduled rules when calling `PutEvents`

### DIFF
--- a/localstack-core/localstack/services/events/provider.py
+++ b/localstack-core/localstack/services/events/provider.py
@@ -1896,6 +1896,10 @@ class EventsProvider(EventsApi, ServiceLifecycleHook):
 
         if configured_rules := list(event_bus.rules.values()):
             for rule in configured_rules:
+                if rule.schedule_expression:
+                    # we do not want to execute Scheduled Rules on PutEvents
+                    continue
+
                 self._process_rules(rule, region, account_id, event_formatted, trace_header)
         else:
             LOG.info(
--- a/tests/aws/services/events/test_events_schedule.py
+++ b/tests/aws/services/events/test_events_schedule.py
@@ -6,6 +6,7 @@ import pytest
 from botocore.exceptions import ClientError
 
 from localstack.testing.aws.eventbus_utils import trigger_scheduled_rule
+from localstack.testing.aws.util import is_aws_cloud
 from localstack.testing.pytest import markers
 from localstack.testing.snapshots.transformer_utility import TransformerUtility
 from localstack.utils.strings import short_uid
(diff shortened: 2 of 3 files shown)


Garethp:
Thanks for getting a solution so quickly! I just had a quick question. I might be misunderstanding how the process event works here:

```py

        if configured_rules := list(event_bus.rules.values()):
            for rule in configured_rules:
                if rule.schedule_expression:
                    # we do not want to execute Scheduled Rules on PutEvents
                    continue

                self._process_rules(rule, region, account_id, event_formatted, trace_header)
        else:
            LOG.info(
                json.dumps(
                    {
                        "InfoCode": "InternalInfoEvents at process_rules",
                        "InfoMessage": f"No rules attached to event_bus: {event_bus_name}",
                    }
                )
            )
```

Does this mean that the filtering of **which** targets to trigger off of an event isn't happening down the line, or is there something special around scheduled rules that makes it difficult to filter out when doing that later?

bentsku:
I think the issue stems from the fact that, from my understanding, `Rules` can be of two "types", regular or scheduled, without taking targets into account yet. The `_process_entry` function is only called when `PutEvents` is called. I think scheduled rules cannot be triggered from a `PutEvents` call, but they are stored as Rules, mixed up with regular ones. 

So we shouldn't consider them when coming from a `PutEvents` call. Or did I misunderstood something? Is the question about where the fix lives? 

I also think the "process_rules" function suffers from a bit of naming issue, as it process one rule only, but it will then dispatch the event to all targets. 
I was a bit overly careful with the fix as well to not disturb logic related to the EventStudio. 

But maybe I misunderstood the question, so please feel free to ask more 😄

Garethp:
Ah @bentsku, thanks for explaining that! It was a misunderstanding on my half. I thought that either there were more types of rules in EventBridge or that schedule rules would also be caught by the filtering that matches rules to targets matching their patterns. 

> Is the question about where the fix lives?

I think that was it, I was curious why the fix was here and how Schedule Rules were different from other rules. I imagined that routing Rules of different types would be handled in the same place, so I expecting to see the fix in the same area of code that, for example, prevented an SSM PutEvent from triggering a lambda that would be the target of an EC2 state change event.

> I also think the "process_rules" function suffers from a bit of naming issue, as it process one rule only, but it will then dispatch the event to all targets.

So if I understand correctly, the event gets dispatched to all targets but the targets filters out the rule to make sure it matches? So an SSM PutEvent would be sent to a target that's listening for EC2 state changes, but the target itself would see that it doesn't need to do anything with that event? And working off of that, if so then Scheduled Targets aren't doing filtering based on the events that they're recieving, so they were being triggered by any event coming that way? Is that somewhat accurate or am I still off the mark on how EventBridge works?

Garethp:
> Nice fix - interesting that we did not notice that sooner it seams to be quite critical for scheduled rules (speaks also to how few people use Eventbridge buses for scheduling things)

It could also be the case that where this was being used it wasn't being used in ways that would be super noticable. I only noticed it because one of our integration tests that flips some SSM Feature Flags triggered a scheduled job that fired not-quite valid data into an SQS Queue with a 5 minute visibility timeout, locking the message group up for that period of time and causing failured further down the line of integration tests. There's another failure in our tests that I'm hoping this would fix, but other use-cases of Scheduled Rules we use are often ones that could have been triggered at any point without failing, so they'd run silently.

bentsku:
> I imagined that routing Rules of different types would be handled in the same place, so I expecting to see the fix in the same area of code that, for example, prevented an SSM PutEvent from triggering a lambda that would be the target of an EC2 state change event.

Ah, I see! Yes, that part is handled right down the line in `_process_rules`. It is true that the fix could have been there as well, with an early return. 

> So if I understand correctly, the event gets dispatched to all targets but the targets filters out the rule to make sure it matches?

I believe the filtering is happening at the rule level (https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rules.html), via the `EventPattern`. It seems a rule can either have a `ScheduleExpression` and be triggered on a Schulz, or an `EventPattern` to match against incoming `PutEvents` calls (even internal ones like a SSM `PutParameter`, which will send a message on the EventBus). 

And for how it worked internally for us, it seems the rules with a Schedule Expression did not have an Event Pattern, and no event pattern somewhat means that there is no filtering. At least that's how I see the issue 😄

Issue #4638, opened 2021-09-22
bug: Lambda Layers API's return 404


shield007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I'm starting localstack via docker-compose. I have init scripts that create a dummy layer. It used the following command:

```
awslocal lambda publish-layer-version --layer-name dummy-layer --description 'Dummy Layer' --license-info MIT --content 'S3Bucket=temp-workspace,S3Key=dummy-layer.zip' --compatible-runtimes python3.6 python3.7 python3.8 nodejs14.x
```

This gives the following response:
```
An error occurred (404) when calling the PublishLayerVersion operation: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.</p>
```

All other requests, give a simlar response:

```/opt/code/localstack # awslocal lambda list-layer-versions --layer-name dummy-layer

An error occurred (404) when calling the ListLayerVersions operation: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.</p>```

Sometimes things do work, most the time they do not. So it would seem to me that something has crashed in the localstack server. It's possiblly when recreating the localstack docker container, but either way I'd expect that to work.


### Expected Behavior

Layers upload and can be listed etc...

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

Here is the localstack part of the docker-compose file:

```  
localstack:
    container_name: "${LOCALSTACK_DOCKER_NAME-localstack_main}"
    image: localstack/localstack
    ports:
      - "443:443"
      - "4566:4566"
      - "4571:4571"
    environment:
      - SERVICES=s3,lambda,stepfunctions,cloudformation,sts,ssm,apigateway,sns,iam,events,kms,sqs,ecr
      - DEBUG=1
#      - DATA_DIR=/tmp/localstack/data
      - LAMBDA_EXECUTOR=docker-reuse
      - LOCALSTACK_API_KEY=${LOCALSTACK_API_KEY- }
      - KINESIS_ERROR_PROBABILITY=${KINESIS_ERROR_PROBABILITY- }
[...]

PR #12717, opened 2025-06-05
refactor: metrics instrumentation framework typing and structure

vittoriopolverino:
## Motivation
follow-up to PR #12386 (thanks @alexrashed for the inputs)

The refactoring simplifies the class hierarchy and removes the overloaded factory pattern, which is increasingly becoming a code smell, as Counter and LabeledCounter evolve with diverging behavior and only partially overlapping interfaces.

The previous design made it increasingly difficult to enforce both static and dynamic typing guarantees, leading to ambiguity around the correct usage of the Counter and weakening the overall abstraction. The new structure clearly separates responsibilities, improves type safety at both runtime and development time, and lays a more solid foundation for extensibility.

Additionally, the monolithic `metrics.py` module will be split into smaller, focused modules.

## Testing
`pytest tests/unit/utils/analytics/test_metrics.py`

## TODO
- introduce metrics schema versioning draft PR #12732 

<br>

**note:** the corresponding updates to the classes using the Counter in the ext package have been addressed in PR [#4638](https://github.com/localstack/localstack-ext/pull/4638)


localstack-core/localstack/utils/analytics/metrics/type.py
```diff
@@ -0,0 +1,72 @@
...
+        return {
+            "namespace": self.namespace,
+            "name": self.name,
+            "value": self.value,
+            "type": self.type,
+        }
+
+
+@dataclass(frozen=True)
+class LabeledCounterPayload:
```
vittoriopolverino, line 32: note: `LabeledCounter` has its own payload as it's diverging from the base `Counter` (e.g. it includes labels, which the base does not). Also, once schema versioning is introduced, it will likely apply only to LabeledCounter, as the base Counter payload should remain stable and not require schema changes (still TBD)

pinzon (approved):
LGTM

localstack-core/localstack/utils/analytics/metrics/hook.py
vittoriopolverino, line 1: **thought:** I think I’m going to rename this module, "hook" doesn’t feel like the most appropriate name 🤔

localstack-core/localstack/services/cloudformation/resource_provider.py
```diff
@@ -19,7 +19,7 @@
 
 from localstack import config
 from localstack.aws.connect import InternalClientFactory, ServiceLevelClientFactory
-from localstack.services.cloudformation import usage
+from localstack.services.cloudformation import analytics as usage
```
alexrashed: **nit:** Since you renamed the module from analytics to usage, would it make sense to also change the name here, i.e. avoid the `as usage`?

localstack-core/localstack/services/stepfunctions/asl/static_analyser/usage_metrics_static_analyser.py
```diff
@@ -3,7 +3,7 @@
 import logging
 from typing import Final
 
-import localstack.services.stepfunctions.usage as UsageMetrics
+import localstack.services.stepfunctions.analytics as UsageMetrics
```
alexrashed: **nit:** I know this was already the case, but the casing of the import is a bit strange here. Would it make sense to just import `analytics` here and use it as such without renaming it?

localstack-core/localstack/utils/analytics/metrics/interface.py
```diff
@@ -0,0 +1,40 @@
...
+    @property
+    def namespace(self) -> str:
+        return self._namespace
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @abstractmethod
+    def collect(self) -> list[Union[CounterPayload, LabeledCounterPayload]]:
```
alexrashed: **nit:** You could use a prototype here which defines that the returned class needs to have an `as_dict` method. That would decouple the parent interface from the implementations.

localstack-core/localstack/utils/analytics/metrics/type.py
```diff
@@ -0,0 +1,72 @@
...
+@dataclass
+class MetricPayload:
+    """
+    A data object storing the value of all metrics collected during the execution of the application.
+    """
+
+    _payload: list[Union[CounterPayload, LabeledCounterPayload]]
+
+    @property
+    def payload(self) -> list[Union[CounterPayload, LabeledCounterPayload]]:
```
alexrashed: **nit:** Similar to the comment in the `Metric` class, you could use a prototype here to decouple the API from its implementations.

localstack-core/localstack/utils/analytics/metrics/interface.py
alexrashed, line 1: **nit:** I am a bit confused by the naming and the content of the different modules in the `metrics` module. Maybe the term `api` would fit better? Similar to `localstack.packages.api`?

localstack-core/localstack/utils/analytics/metrics/type.py
alexrashed, line 1: **suggestion:** I would suggest to avoid a separate `type` module which contains all the types, but rather add the types next to the classes that use them. That should be easily possible once the parent interface does not reference its implementations (I guess the cyclic import was the initial motivation for creating the `type` module).

alexrashed (commented):
Nice and clean refactoring to move away from the factory pattern. As you explained in your PR description, and as discussed in the previous PRs, I think this totally makes sense.
I added a few comments / questions / suggestions, afterwards this should be good to go! 💯
Thanks for continuously improving the analytics framework!

localstack-core/localstack/services/cloudformation/resource_provider.py
vittoriopolverino: Absolutely, that makes total sense. I initially avoided it out of caution, in case I missed some spots during the renaming 😄 
 I'll do it

localstack-core/localstack/utils/analytics/metrics/interface.py
vittoriopolverino: api instead of interface sounds good 🚀 

**question:** Are there any other modules whose naming you find unclear or confusing?

localstack-core/localstack/utils/analytics/metrics/type.py
vittoriopolverino: Thanks for the suggestion! I’ll give it a try 🚀

localstack-core/localstack/utils/analytics/metrics/interface.py
alexrashed: No, I think now the modules are pretty clear! 💯

PR #12723, opened 2025-06-06
Fix publishing credentials

silv-io:
## Motivation

The DockerHub user secrets were wrongly assumed to contain `PUSH` in them, compared to the `PULL` tokens. However, a look at the s3 image pipeline shows that this is not the case.

Also, the twine credentials are missing in the dev release publishing step.

## Changes

- Adapt to use correct secret names
- Adapt to use twine username and password for dev release publish


k-a-il (approved):
Thank you for handling this!

PR #12726, opened 2025-06-06
fix dependencies installation for dev release publish

bentsku:
## Motivation
Follow up from #12725, we were using `make install-basic` before building the entry points, which lead to issue while building, short snippet:

```bash
running plugins
error importing module localstack.aws.api.sts: No module named 'botocore'
error importing module localstack.services.opensearch.resource_providers.aws_elasticsearch_domain: No module named 'botocore'
error importing module localstack.aws.api.core: No module named 'botocore'
error importing module localstack.services.stepfunctions.asl.component.state.state_choice.comparison.comparison: No module named 'antlr4'
error importing module localstack.services.opensearch.cluster: No module named 'werkzeug'
error importing module localstack.services.sqs.exceptions: No module named 'botocore'
```

By using `install-runtime`, we'll make sure we have the right dependencies before building the entry points

## Changes
- switch the make target before publishing

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -262,7 +262,7 @@ jobs:
           if git describe --exact-match --tags >/dev/null 2>&1; then
             echo "not publishing a dev release as this is a tagged commit"
           else
-            make install-basic publish || echo "dev release failed (maybe it is already published)"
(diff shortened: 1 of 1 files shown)


cloutierMat (approved):
:rocket: Thank you for the deep dive, debug and super quick fix! :tada:

PR #12731, opened 2025-06-09
Implement support for CDK Bootstrap stacks

simonrw:
## Motivation

Currently CDK bootstrap fails for a variety of reasons. Without this,
users who use CDK cannot test their applications with our new engine.

## Changes

* Handle dynamic parameter lookup of SSM parameters
* Set resource status on success/failure of a deployment
* Handle the case where an Fn::Join is called without a list as a second argument, in the special case of an empty string, e.g. `Fn::Join: ["", ""] as the CDK does this
* Add helper `find_stack_v2` function for finding v2 stacks
* Implement `describe_stack_resources` to support our CDK bootstrap tests
* Update the cdk bootstrap test skip reason as our lack of deletion
  causes test failures

## Testing

I have not added any integration tests for this since its functionality is covered by our existing `TestCdkInit.test_cdk_bootstrap` (parametrized) tests. Unfortunately since we don't support deletions yet (see https://github.com/localstack/localstack/pull/12576), the tests fail when run together. Each test individually passes, so I've updated the skip reason to reflect the updated nature of the failure.

To test manually, unskip the tests, and run

```
pytest tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_cdk.py::TestCdkInit::test_cdk_bootstrap[10]
```

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -4,8 +4,13 @@ import uuid
 from dataclasses import dataclass
 from typing import Final, Optional
 
-from localstack.aws.api.cloudformation import ChangeAction, StackStatus
+from localstack.aws.api.cloudformation import (
+    ChangeAction,
+    ResourceStatus,
+    StackStatus,
+)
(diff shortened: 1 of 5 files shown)


pinzon (approved):
LGTM

PR #12732, opened 2025-06-09
feat: metrics schema version

vittoriopolverino:


localstack-core/localstack/utils/analytics/metrics/counter.py
```diff
@@ -17,13 +17,15 @@ class CounterPayload:
...
     type: str
+    schema_version: int
 
     def as_dict(self) -> dict[str, Any]:
         return {
             "namespace": self.namespace,
             "name": self.name,
             "value": self.value,
             "type": self.type,
+            "schema_version": self.schema_version,
```
vittoriopolverino, line 28: note: for now, I’m adding `schema_version` only to the counter implementations. We can reassess later whether this field becomes relevant for other metric types as well. if that’s the case, we can move its definition up to the base `Metric` class.

localstack-core/localstack/utils/analytics/metrics/counter.py
```diff
@@ -132,15 +151,25 @@ class LabeledCounter(Metric):
     """
 
     _type: str
+    _schema_version: int
```
vittoriopolverino: **note:** for now, I’m adding `schema_version` only to the counter implementations. If it proves to be relevant for other metric types as well, we’ll move its definition up to the base `Metric` class.

localstack-core/localstack/utils/analytics/metrics/counter.py
```diff
@@ -98,12 +102,23 @@ class Counter(Metric, ThreadSafeCounter):
...
 
-    def __init__(self, namespace: str, name: str):
+    def __init__(self, namespace: str, name: str, schema_version: int = 1):
```
vittoriopolverino, line 106: **note:** default is set to 1 and automatically applied to all existing counters.

localstack-core/localstack/utils/analytics/metrics/counter.py
```diff
@@ -98,12 +102,23 @@ class Counter(Metric, ThreadSafeCounter):
...
         Metric.__init__(self, namespace=namespace, name=name)
         ThreadSafeCounter.__init__(self)
 
+        if schema_version is None:
+            raise ValueError("Labeled metrics require an explicit schema version.")
```
vittoriopolverino: **thought:** It shouldn’t ever happen given the default is set to 1, 

**question:** should we remove the check, or simply keep it as an extra safeguard?

localstack-core/localstack/utils/analytics/metrics/counter.py
```diff
@@ -98,12 +102,23 @@ class Counter(Metric, ThreadSafeCounter):
...
+
+        if not isinstance(schema_version, int):
+            raise TypeError("Schema version must be an integer.")
+
+        if schema_version <= 0:
+            raise ValueError("Schema version must be greater than zero.")
```
vittoriopolverino: **note:** naive approach. At the moment, we can't verify whether the `schema_version` is being incremented sequentially. However, once approved, we’ll explicitly document how `schema_version` should be used going forward.

alexrashed: +1
I don't really think this is going to be a problem, since it is pretty obvious that a schema is evolved by incrementing this value. As you mentioned, I don't think anything else will be necessary here.

localstack-core/localstack/utils/analytics/metrics/counter.py
alexrashed: **suggestion:** I would suggest to move it to the `Metric` class directly because:
- You are already adding it to all known implementations.
- You could already get rid a few lines of code (the code checking the parameter value and type).
- Schema versioning are a common practice and it makes sense to introduce this early for all metrics.

alexrashed (approved):
This is looking good, and in my opinion you should just define this on the base `Metric` class from the beginning. However, I'll totally leave this up to you! Thanks for continuously improving the metrics framework! 💯

localstack-core/localstack/utils/analytics/metrics/counter.py
alexrashed: Since this check is only executed when instantiated it should not have a bigger runtime impact.
And by the way: This can still happen if a caller explicitly sets it to None. The default will only be used if the parameter is not set, not if the parameter is set but None:
```
>>> from localstack.utils.analytics.metrics.counter import Counter
>>> Counter("namespace", "name", None)
  File "<stdin>", line 1, in <module>
  File "/home/localstack/Repos/localstack/localstack-core/localstack/utils/analytics/metrics/counter.py", line 112, in __init__
    raise ValueError("An explicit schema_version is required for Counter metrics")
ValueError: An explicit schema_version is required for Counter metrics
```

localstack-core/localstack/utils/analytics/metrics/counter.py
vittoriopolverino: agree, wil do 👍🏼

localstack-core/localstack/utils/analytics/metrics/counter.py
vittoriopolverino: Right, I guess there was a reason I added it in the first place 😄
I'll add unit tests as well.

alexrashed:
FYI: The three failing tests are unrelated / are also failing on `master` right now and will be resolved with https://github.com/localstack/localstack/pull/12767.

PR #12736, opened 2025-06-10
WIP: feat(lambda): Use Lambda Runtime Emulator Interface v2

gregfurman:
## Motivation
Currently a WIP. This PR was created to trigger a CI/CD run with the new Lambda RIE v2 enabled by default.

## Related

Part of DRG-1


tests/aws/services/lambda_/test_lambda.snapshot.json
```diff
@@ -3692,38 +3730,14 @@
...
-            "AWS_EXECUTION_ENV": "AWS_Lambda_rapid",
-            "AWS_LAMBDA_INITIALIZATION_TYPE": "on-demand",
-            "_LAMBDA_RUNTIME_LOAD_TIME": "<runtime-load-time:1>"
-          }
+          "errorMessage": "[Errno 13] Permission denied: '/proc/1/environ'",
+          "errorType": "PermissionError",
+          "requestId": "<uuid:2>",
+          "stackTrace": [
+            "  File \"/var/task/handler.py\", line 3, in handler\n    with open(f\"/proc/{pid}/environ\", mode=\"rt\") as f:\n"
+          ]
```
gregfurman, line 3740: @dfangl @joe4dev Seems like this test now expects a permissions error 🤔 Any pointers on how we should be configuring the user permissions when accessing the lambda FS? We currently set the following:

https://github.com/localstack/lambda-runtime-init/blob/9be92e099e99c9973f2e1730d1af6d4ffd769bb2/cmd/localstack/main.go#L155-L169

joe4dev: Can we parity-test/inspect this, or does the permission prohibit that?
If not, @dfangl might have a better guess.  Maybe, try 600 🤷‍♂️

dfangl: I think the tests makes no longer sense then - it was mean to introspect the environment of the init process, if that is no longer possible in AWS, there is no real point (although we could try to hide it in our containers as well, I don't necessarily see this as a priority)

alexrashed:
@gregfurman This PR seems pretty outdated, do you want to revive or close it? :)

alexrashed:
Since this PR is outdated quite a bit, it can be migrated at any time in the future, and we are closing down operations in this repo, I am closing this PR.

PR #12740, opened 2025-06-11
Publish service provider mapping on startup

silv-io:
## Motivation

We have the need to track which providers are used for different AWS services. We can do this by a startup hook which basically sends the service config on startup.

/cc @mmaureenliu 

## Changes

- add a hook which takes the service config and creates a mapping from service to provider plugin name and publishes it to the analytics backend

### DIFF
--- a/localstack-core/localstack/utils/analytics/service_providers.py
+++ b/localstack-core/localstack/utils/analytics/service_providers.py
@@ -6,25 +6,14 @@ def publish_provider_assignment():
     """
     Publishes the service provider assignment to the analytics service.
     """
-    from datetime import datetime
 
     from localstack.config import SERVICE_PROVIDER_CONFIG
     from localstack.services.plugins import SERVICE_PLUGINS
-    from localstack.utils.analytics import get_session_id
-    from localstack.utils.analytics.events import Event, EventMetadata
-    from localstack.utils.analytics.publisher import AnalyticsClientPublisher
+    from localstack.utils.analytics import log
 
     provider_assignment = {
         service: f"localstack.aws.provider/{service}:{SERVICE_PROVIDER_CONFIG[service]}"
         for service in SERVICE_PLUGINS.list_available()
     }
-    metadata = EventMetadata(
-        session_id=get_session_id(),
-        client_time=str(datetime.now()),
-    )
 
-    event = Event(
-        name="ls_service_provider_assignment", metadata=metadata, payload=provider_assignment
-    )
-
-    AnalyticsClientPublisher().publish([event])
+    log.event("ls_service_provider_assignment", provider_assignment)


localstack-core/localstack/aws/handlers/analytics.py
```diff
@@ -42,13 +42,15 @@ def __call__(self, chain: HandlerChain, context: RequestContext, response: Respo
         err_type = self._get_err_type(context, response) if response.status_code >= 400 else None
         service_name = context.operation.service_model.service_name
         operation_name = context.operation.name
+        provider = config.SERVICE_PROVIDER_CONFIG.get_provider(service_name)
```
mmaureenliu: this line seems to suggest that there's a 1:1 relationship of service_name to provider, which I think is not true - did I misread the function or the 1:1 relationship actually is true in the community version?

silv-io: In a running instance the relation is 1:1. Only one provider can be active at once. The active provider is defined in the config

thrau (commented):
Thanks for taking a stab at this! My high-level feedback is: if we include the service provider in each service request, we'll also need to adapt the analytics backend, specifically the materialized view that stores the AWS service request. We can do that, but this creates additional work, is quite a substantial change. Also, since the set of providers doesn't (because it currently cannot) change during the runtime of LocalStack, we'll be sending redundant data for each request.

It might be more efficient to capture the configured providers _once_ at startup, similar to how we log environment variables. This would avoid redundancy and keep the data leaner. Happy to provide more guidance if needed, but suggest to sync up with @vittoriopolverino on this

mmaureenliu:
> Thanks for taking a stab at this! My high-level feedback is: if we include the service provider in each service request, we'll also need to adapt the analytics backend, specifically the materialized view that stores the AWS service request. We can do that, but this creates additional work, is quite a substantial change. Also, since the set of providers doesn't (because it currently cannot) change during the runtime of LocalStack, we'll be sending redundant data for each request.
> 
> It might be more efficient to capture the configured providers _once_ at startup, similar to how we log environment variables. This would avoid redundancy and keep the data leaner. Happy to provide more guidance if needed, but suggest to sync up with @vittoriopolverino on this

Just want some clarification on the "capture the configured providers once at startup" suggestion - do you mean at each session start, send a map of provider to service to the data backend, and let the data backend figure out the provider for each service within the session?

vittoriopolverino:
> suggest to sync up with @vittoriopolverino on this

Happy to support on this. +1 to what Thomas already pointed out

localstack-core/localstack/utils/analytics/service_providers.py
```diff
@@ -0,0 +1,30 @@
...
+    metadata = EventMetadata(
+        session_id=get_session_id(),
+        client_time=str(datetime.now()),
+    )
+
+    event = Event(
+        name="ls_service_provider_assignment", metadata=metadata, payload=provider_assignment
+    )
+
+    AnalyticsClientPublisher().publish([event])
```
thrau: we should use analytics.log.event event here:

```python
from localstack.utils import analytics

@hooks.on_runtime_ready()
def publish_provider_assignment():
    ...
    analytics.log.event("ls_service_provider_assignment", provider_assignment)
```

vittoriopolverino: ~~**note:** just to add more context, this way we should receive the event only after the `session` event. This is important for us due to some payload enrichment logic in the analytics-backend.~~ 
I hadn’t realized Thomas had already explained it above

silv-io: Awesome, thanks. This was very confusing to me at first. Will fix it when I'm back next week

vittoriopolverino:
FYI @silv-io  I’ve opened a [PR](https://github.com/localstack/analytics-backend/pull/127) so that we’re ready to receive and materialize this new event.

Issue #12739, opened 2025-06-11
feature request: Implement Put, Get, Delete, List S3 BucketMetricsConfiguration


cristiangiann:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

Implement the following S3 operations related to CloudWatch metrics configuration:
- [PutBucketMetricsConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketMetricsConfiguration.html)
- [GetBucketMetricsConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketMetricsConfiguration.html)
- [DeleteBucketMetricsConfiguration](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucketMetricsConfiguration.html)
- [ListBucketMetricsConfigurations](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBucketMetricsConfigurations.html)

Don’t need to build the actual metrics logic or connect to CloudWatch — just focus on the basic create, read, update, and delete operations for managing the configurations.

PR #12741, opened 2025-06-11
Implement Put, Get, Delete, List S3 MetricsConfiguration API

cristiangiann:
## Motivation
Feature request: Implement Put, Get, Delete, List S3 BucketMetricsConfiguration: https://github.com/localstack/localstack/issues/12739

## Changes
Implement Put, Get, Delete, List S3 BucketMetricsConfiguration APIs. These APIs will be available in LocalStack.

- **PutBucketMetricsConfiguration**: Add or update one BucketMetricsConfiguration. The limit to existing BucketMetricsConfiguration per bucket is 1000.
- **GetBucketMetricsConfiguration**: Get one BucketMetricsConfiguration. Return empty response if it does not exist.
- **DeleteBucketMetricsConfiguration**: Delete one BucketMetricsConfiguration.
- **ListBucketMetricsConfiguration**: Fetch all BucketMetricsConfigurations, paginated through 100-item pages.

### DIFF
--- a/localstack-core/localstack/services/s3/models.py
+++ b/localstack-core/localstack/services/s3/models.py
@@ -37,6 +37,8 @@ from localstack.aws.api.s3 import (
     LoggingEnabled,
     Metadata,
     MethodNotAllowed,
+    MetricsConfiguration,
+    MetricsId,
     MultipartUploadId,
     NoSuchKey,
     NoSuchVersion,
@@ -115,6 +117,7 @@ class S3Bucket:
     intelligent_tiering_configurations: dict[IntelligentTieringId, IntelligentTieringConfiguration]
     analytics_configurations: dict[AnalyticsId, AnalyticsConfiguration]
     inventory_configurations: dict[InventoryId, InventoryConfiguration]
+    metric_configurations: dict[MetricsId, MetricsConfiguration]
     object_lock_default_retention: Optional[DefaultRetention]
     replication: ReplicationConfiguration
     owner: Owner
@@ -154,6 +157,7 @@ class S3Bucket:
         self.intelligent_tiering_configurations = {}
         self.analytics_configurations = {}
         self.inventory_configurations = {}
+        self.metric_configurations = {}
         self.object_lock_default_retention = {}
         self.replication = None
         self.acl = acl
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -80,6 +80,7 @@ from localstack.aws.api.s3 import (
     GetBucketLifecycleConfigurationOutput,
     GetBucketLocationOutput,
     GetBucketLoggingOutput,
+    GetBucketMetricsConfigurationOutput,
     GetBucketOwnershipControlsOutput,
     GetBucketPolicyOutput,
     GetBucketPolicyStatusOutput,
@@ -126,6 +127,7 @@ from localstack.aws.api.s3 import (
     ListBucketAnalyticsConfigurationsOutput,
     ListBucketIntelligentTieringConfigurationsOutput,
     ListBucketInventoryConfigurationsOutput,
+    ListBucketMetricsConfigurationsOutput,
     ListBucketsOutput,
     ListMultipartUploadsOutput,
     ListObjectsOutput,
@@ -138,6 +140,8 @@ from localstack.aws.api.s3 import (
     MaxParts,
     MaxUploads,
     MethodNotAllowed,
+    MetricsConfiguration,
+    MetricsId,
     MissingSecurityHeader,
     MpuObjectSize,
     MultipartUpload,
@@ -240,6 +244,7 @@ from localstack.services.s3.exceptions import (
     MalformedXML,
     NoSuchConfiguration,
     NoSuchObjectLockConfiguration,
+    TooManyConfigurations,
     UnexpectedContent,
 )
 from localstack.services.s3.models import (
@@ -4432,6 +4437,143 @@ class S3Provider(S3Api, ServiceLifecycleHook):
             response["Location"] = response["LocationHeader"]
 
         return response
+
+    def put_bucket_metrics_configuration(
+        self,
+        context: RequestContext,
+        bucket: BucketName,
+        id: MetricsId,
+        metrics_configuration: MetricsConfiguration,
+        expected_bucket_owner: AccountId = None,
+        **kwargs,
+    ) -> None:
+        """
+        Update or add a new metrics configuration. If the provided `id` already exists, its associated configuration
+        will be overwritten. The total number of metric configurations is limited to 1000. If this limit is exceeded,
+        an error is raised unless the `is` already exists.
+
+        :param context: The request context.
+        :param bucket: The name of the bucket associated with the metrics configuration.
+        :param id: Identifies the metrics configuration being added or updated.
+        :param metrics_configuration: A new or updated configuration associated with the given metrics identifier.
+        :param expected_bucket_owner: The expected account ID of the bucket owner.
+        :return: None
+        :raises TooManyConfigurations: If the total number of metrics configurations exceeds 1000 AND the provided
+            `metrics_id` does not already exist.
+        """
+        store, s3_bucket = self._get_cross_account_bucket(
+            context, bucket, expected_bucket_owner=expected_bucket_owner
+        )
+
+        if (
+            len(s3_bucket.metric_configurations) >= 1000
+            and id not in s3_bucket.metric_configurations
+        ):
+            raise TooManyConfigurations("Too many metrics configurations")
+        s3_bucket.metric_configurations[id] = metrics_configuration
+
+    def get_bucket_metrics_configuration(
+        self,
+        context: RequestContext,
+        bucket: BucketName,
+        id: MetricsId,
+        expected_bucket_owner: AccountId = None,
+        **kwargs,
+    ) -> GetBucketMetricsConfigurationOutput:
+        """
+        Retrieve the metrics configuration associated with a given metrics identifier.
+
+        :param context: The request context.
+        :param bucket: The name of the bucket associated with the metrics configuration.
+        :param id: The unique identifier of the metrics configuration to retrieve.
+        :param expected_bucket_owner: The expected account ID of the bucket owner.
+        :return: The metrics configuration associated with the given metrics identifier.
+        :raises NoSuchConfiguration: If the provided metrics configuration does not exist.
+        """
+        store, s3_bucket = self._get_cross_account_bucket(
+            context, bucket, expected_bucket_owner=expected_bucket_owner
+        )
+
+        metric_config = s3_bucket.metric_configurations.get(id)
+        if not metric_config:
+            raise NoSuchConfiguration("The specified configuration does not exist.")
+        return GetBucketMetricsConfigurationOutput(MetricsConfiguration=metric_config)
+
+    def list_bucket_metrics_configurations(
+        self,
+        context: RequestContext,
+        bucket: BucketName,
+        continuation_token: Token = None,
+        expected_bucket_owner: AccountId = None,
+        **kwargs,
+    ) -> ListBucketMetricsConfigurationsOutput:
+        """
+        Lists the metric configurations available, allowing for pagination using a continuation token to retrieve more
+        results.
+
+        :param context: The request context.
+        :param bucket: The name of the bucket associated with the metrics configuration.
+        :param continuation_token: An optional continuation token to retrieve the next set of results in case there are
+            more results than the default limit. Provided as a base64-encoded string value.
+        :param expected_bucket_owner: The expected account ID of the bucket owner.
+        :return: A list of metric configurations and an optional continuation token for fetching subsequent data, if
+            applicable.
+        """
+        store, s3_bucket = self._get_cross_account_bucket(
+            context, bucket, expected_bucket_owner=expected_bucket_owner
+        )
+
+        metrics_configurations: list[MetricsConfiguration] = []
+        next_continuation_token = None
+
+        decoded_continuation_token = (
+            to_str(base64.urlsafe_b64decode(continuation_token.encode()))
+            if continuation_token
+            else None
+        )
+
+        for metric in sorted(s3_bucket.metric_configurations.values(), key=lambda r: r["Id"]):
+            if continuation_token and metric["Id"] < decoded_continuation_token:
+                continue
+
+            if len(metrics_configurations) >= 100:
+                next_continuation_token = to_str(base64.urlsafe_b64encode(metric["Id"].encode()))
+                break
+
+            metrics_configurations.append(metric)
+
+        return ListBucketMetricsConfigurationsOutput(
+            IsTruncated=next_continuation_token is not None,
+            ContinuationToken=continuation_token,
+            NextContinuationToken=next_continuation_token,
+            MetricsConfigurationList=metrics_configurations,
+        )
+
+    def delete_bucket_metrics_configuration(
+        self,
+        context: RequestContext,
+        bucket: BucketName,
+        id: MetricsId,
+        expected_bucket_owner: AccountId = None,
+        **kwargs,
+    ) -> None:
+        """
+        Removes a specific metrics configuration identified by its metrics ID.
+
+        :param context: The request context.
+        :param bucket: The name of the bucket associated with the metrics configuration.
+        :param id: The unique identifier of the metrics configuration to delete.
+        :param expected_bucket_owner: The expected account ID of the bucket owner.
+        :return: None
+        :raises NoSuchConfiguration: If the provided metrics configuration does not exist.
+        """
+        store, s3_bucket = self._get_cross_account_bucket(
+            context, bucket, expected_bucket_owner=expected_bucket_owner
+        )
+
+        deleted_config = s3_bucket.metric_configurations.pop(id, None)
+        if not deleted_config:
+            raise NoSuchConfiguration("The specified configuration does not exist.")
 
 
 def generate_version_id(bucket_versioning_status: str) -> str | None:
--- a/tests/aws/services/s3/test_s3_api.py
+++ b/tests/aws/services/s3/test_s3_api.py
@@ -2180,3 +2180,164 @@ class TestS3ObjectWritePrecondition:
             IfMatch=multipart_etag,
         )
         snapshot.match("complete-multipart-if-match-overwrite-multipart", complete_multipart_1)
+
+
+class TestS3MetricsConfiguration:
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            # PutBucketMetricsConfiguration should return 204, but we return 200
+            "$.put_bucket_metrics_configuration.ResponseMetadata.HTTPStatusCode",
+        ]
+    )
+    def test_put_bucket_metrics_configuration(self, s3_bucket, aws_client, snapshot):
+        snapshot.add_transformer([snapshot.transform.key_value("Id")])
+
+        metric_id = short_uid()
+        metrics_config = {"Id": metric_id, "Filter": {"Prefix": "logs/"}}
+
+        put_result = aws_client.s3.put_bucket_metrics_configuration(
+            Bucket=s3_bucket, Id=metric_id, MetricsConfiguration=metrics_config
+        )
+        snapshot.match("put_bucket_metrics_configuration", put_result)
+
+        get_result = aws_client.s3.get_bucket_metrics_configuration(Bucket=s3_bucket, Id=metric_id)
+        snapshot.match("get_bucket_metrics_configuration", get_result)
+
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            # PutBucketMetricsConfiguration should return 204, but we return 200
+            "$.overwrite_bucket_metrics_configuration.ResponseMetadata.HTTPStatusCode",
+        ]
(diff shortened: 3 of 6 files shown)


cristiangiann:

cristiangiann:
Sorry, I missed the label before pushing Publish PR button. I think I cannot modify it after publishing, probably I don't have permissions.
Suggested label: semver: minor

localstack-core/localstack/services/s3/models.py
```diff
@@ -4,7 +4,7 @@
 from collections import defaultdict
 from datetime import datetime
 from secrets import token_urlsafe
-from typing import Literal, NamedTuple, Optional, Union
+from typing import List, Literal, NamedTuple, Optional, Tuple, Union
```
bentsku: nit: these type aliases are [deprecated](https://docs.python.org/3/library/typing.html#deprecated-aliases), please use `list[<type>]` and `tuple[<type>]` instead

localstack-core/localstack/services/s3/models.py
```diff
@@ -244,6 +250,90 @@ def get_object(
 
         return s3_object
 
+    def put_metrics_configuration(
```
bentsku: Due to how we are serializing our state in LocalStack, we try to put as little logic as possible in the `models` themselves. 

This logic should live in the Provider itself instead, and in the API operation. I can see your implementation is more in line with OOP principles, but we have constraints that prevents us to add this logic in the `S3Bucket` model. 

Would you mind moving this over the provider itself instead? Thanks!

localstack-core/localstack/services/s3/models.py
```diff
@@ -244,6 +250,90 @@ def get_object(
...
+        self,
+        metrics_id: MetricsId,
+        metrics_configuration: MetricsConfiguration,
+    ) -> None:
+        """
```
bentsku: note: great docstrings! 🚀

localstack-core/localstack/services/s3/provider.py
```diff
@@ -4423,6 +4428,98 @@ def post_object(
...
+        expected_bucket_owner: AccountId = None,
+        **kwargs,
+    ) -> None:
+        store, s3_bucket = self._get_cross_account_bucket(context, bucket)
+
+        if not s3_bucket:
+            raise NoSuchBucket(bucket)
+
+        if expected_bucket_owner and s3_bucket.bucket_account_id != expected_bucket_owner:
+            raise AccessDenied("Access Denied")
```
bentsku: question: could you please explain how these cases could happen? it looks like `_get_cross_account_bucket` would take care of that, and the first error raised here does not follow the proper format.

localstack-core/localstack/services/s3/provider.py
```diff
@@ -4423,6 +4428,98 @@ def post_object(
...
+        if not metrics_configuration.get("Id"):
+            raise InvalidRequest("Missing Id in MetricsConfiguration")
+
+        if filter := metrics_configuration.get("Filter"):
+            if not any([filter.get("Prefix"), filter.get("Tag"), filter.get("And")]):
+                raise InvalidRequest("Filter must have at least one of Prefix, Tag, or And")
+        try:
+            s3_bucket.put_metrics_configuration(id, metrics_configuration)
+        except InvalidArgument:
+            raise TooManyConfigurations("Too many metrics configurations")
```
bentsku: question: I do not see those exception messages (`InvalidRequest `) in the snapshots. Do we have tests covering those cases? We do not have to add those tests as part of this PR, but in this case, we might refrain from adding validation that hasn't been tested against AWS

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
             IfMatch=multipart_etag,
         )
         snapshot.match("complete-multipart-if-match-overwrite-multipart", complete_multipart_1)
+
+
+class TestS3MetricsConfiguration:
+    @markers.aws.validated
+    def test_put_bucket_metrics_configuration(self, s3_bucket, aws_client, snapshot):
+        metric_id = str(uuid.uuid4())
```
bentsku: nit: no need to import and use `str(uuid.uuid4())`, we have the `short_uid()` util method for this use case

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+        metrics_config = {"Id": metric_id, "Filter": {"Prefix": "logs/"}}
+
+        put_result = aws_client.s3.put_bucket_metrics_configuration(
+            Bucket=s3_bucket, Id=metric_id, MetricsConfiguration=metrics_config
+        )
+        snapshot.match("put_bucket_metrics_configuration", put_result)
+
+        result = aws_client.s3.get_bucket_metrics_configuration(Bucket=s3_bucket, Id=metric_id)
+        assert result["MetricsConfiguration"] == metrics_config
```
bentsku: question: why are we not snapshotting the response from `get_bucket_metrics_configuration`? 

In this case, we'd try to avoid manual `assert` and rely on the snapshot library which gives us greater confidence in our AWS parity

tests/aws/services/s3/test_s3_api.py
bentsku, line 1: it seems those tests haven't been ran against AWS, could you confirm? it looks like we are missing a file attesting this: `test_s3_api.validation.json`

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+        for metrics_config in metrics_configs.values():
+            aws_client.s3.put_bucket_metrics_configuration(
+                Bucket=s3_bucket, Id=metrics_config["Id"], MetricsConfiguration=metrics_config
+            )
+
+        result_configs = aws_client.s3.list_bucket_metrics_configurations(Bucket=s3_bucket)
+        assert len(result_configs["MetricsConfigurationList"]) == 2
+        for result_config in result_configs["MetricsConfigurationList"]:
+            assert result_config["Id"] in metrics_configs
+            assert result_config == metrics_configs[result_config["Id"]]
```
bentsku: thanks to the snapshot match, you do not need all those asserts, they are already being take care of with the `list_bucket_metrics_configurations` snapshot match. 👍

localstack-core/localstack/services/s3/models.py
```diff
@@ -244,6 +250,90 @@ def get_object(
...
+        is exceeded, an error is raised unless the `metrics_id` already exists.
+
+        :param metrics_id: Identifies the metrics configuration being added or updated.
+        :param metrics_configuration: A new or updated configuration associated with the given metrics identifier.
+        :return: None
+        :raises InvalidArgument: If the total number of metrics configurations exceeds 1000 AND the provided
+            `metrics_id` does not already exist.
+        """
+        # Overwrite existing configuration if metrics_id already exists
+        if len(self.metric_configurations) >= 1000 and metrics_id not in self.metric_configurations.keys():
```
bentsku: nit: no need to call `.keys()` to do a presence check in a dictionary, you can do `metrics_id not in self.metric_configurations`

localstack-core/localstack/services/s3/models.py
```diff
@@ -244,6 +250,90 @@ def get_object(
...
+        self,
+        metrics_id: MetricsId,
+    ) -> Optional[MetricsConfiguration]:
+        """
+        Retrieve the metrics configuration associated with a given metrics identifier.
+
+        :param metrics_id: The unique identifier of the metrics configuration to retrieve.
+        :return: The metrics configuration associated with the given metrics ID, or None if no configuration is found.
+        """
+        return self.metric_configurations.get(metrics_id, None)
```
bentsku: nit: no need to set `None` as the second argument to `.get`, it is the default value

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+            metrics_configs[metric_id] = {"Id": metric_id, "Filter": {"Prefix": "logs/prefix"}}
+
+        for metrics_config in metrics_configs.values():
+            aws_client.s3.put_bucket_metrics_configuration(
+                Bucket=s3_bucket, Id=metrics_config["Id"], MetricsConfiguration=metrics_config
+            )
+
+        result_configs_page_1 = aws_client.s3.list_bucket_metrics_configurations(Bucket=s3_bucket)
+        assert len(result_configs_page_1["MetricsConfigurationList"]) == 100
+        assert result_configs_page_1["NextContinuationToken"]
```
bentsku, line 2276: here, I can see us not wanting to snapshot and manually assert, I agree with that, as it would make the snapshot really long for no real benefit 👍 this is nice!

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+
+        result_configs_page_2 = aws_client.s3.list_bucket_metrics_configurations(
+            Bucket=s3_bucket, ContinuationToken=result_configs_page_1["NextContinuationToken"]
+        )
+        assert len(result_configs_page_2["MetricsConfigurationList"]) == 2
+        for result_config in result_configs_page_2["MetricsConfigurationList"]:
+            assert result_config["Id"] in metrics_configs
+            assert result_config == metrics_configs[result_config["Id"]]
```
bentsku: here we might want to check that we are properly returning the 2 last values to see if the pagination is working properly? 
snapshotting the 2 last ones might be a good idea 👍 
if you have issue with values changing between runs, you can use Snapshot Transformers: https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing#using-transformers

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+        assert result["MetricsConfiguration"] == metrics_config
+        snapshot.match("get_bucket_metrics_configuration", result)
+
+    @markers.aws.validated
+    def test_get_bucket_metrics_configuration_not_exist(self, s3_bucket, aws_client, snapshot):
+        result = aws_client.s3.get_bucket_metrics_configuration(
+            Bucket=s3_bucket, Id="does-not-exist"
+        )
+        assert not result["MetricsConfiguration"]
+        snapshot.match("get_bucket_metrics_configuration", result)
```
bentsku: I thought, from would AWS usually do, that this would raise an Exception. You can find this pattern by looking for `pytest.raises(ClientError) as e:` in this test file. 

I believe this should raise a `NoSuchConfigurationError`.

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2181,122 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
+        )
+
+        delete_result = aws_client.s3.delete_bucket_metrics_configuration(
+            Bucket=s3_bucket, Id=metric_id
+        )
+        snapshot.match("delete_bucket_metrics_configuration", delete_result)
+
+        get_result = aws_client.s3.get_bucket_metrics_configuration(Bucket=s3_bucket, Id=metric_id)
+
+        assert not get_result["MetricsConfiguration"]
```
bentsku: question: could we try to delete the same configuration twice to see if AWS raises an error or just accept it? 
also, the `get_bucket_metrics_configuration` should fail here

cristiangiann:
I published an updated commit fixing a leftover from the last `make lint`

localstack-core/localstack/services/s3/models.py
```diff
@@ -244,6 +250,93 @@ def get_object(
...
+        for metric in sorted(self.metric_configurations.values(), key=lambda r: r["Id"]):
+            if continuation_token and metric["Id"] < decoded_continuation_token:
+                continue
+
+            if len(metrics_configs) >= 100:
+                next_continuation_token = to_str(base64.urlsafe_b64encode(metric["Id"].encode()))
+                break
+
+            metrics_configs.append(metric)
+
```
bentsku: kudos for going the extra mile and implementing pagination! 🚀

localstack-core/localstack/aws/api/s3/__init__.py
```diff
@@ -1025,6 +1025,12 @@ class BadDigest(ServiceException):
     CalculatedDigest: Optional[ContentMD5]
 
 
+class TooManyConfigurations(ServiceException):
+    code: str = "TooManyConfigurations"
+    sender_fault: bool = False
+    status_code: int = 400
+
```
bentsku: this [file is autogenerated](https://github.com/localstack/localstack/blob/master/docs/localstack-concepts/README.md), and this change will be overwritten on our next automated update.

You can either define it in s3/exceptions.py, or, as it used only once here, manually raise a CommonServiceException:

```python
# when you need to raise this exception
raise CommonServiceException(
    code="TooManyConfigurations",
    message="Too many metrics configurations",
)
```

tests/aws/services/s3/test_s3_api.py
cristiangiann: I thought to match the snapshots only for the methods that the test is actually testing (example: `test_put_bucket_metrics_configuration` is only checking that `put_bucket_metrics_configuration` API response matches the expected value, while `get_bucket_metrics_configuration` response is checked in a dedicated test). I can add it also for the other assertions

localstack-core/localstack/services/s3/models.py
cristiangiann: Just a clarification: Should I remove the methods put/get/list/delete methods from the model, and handle everything in the provider?

bentsku: Yes, please, it would align more with how the rest of the codebase is structured 👍

localstack-core/localstack/services/s3/provider.py
cristiangiann: You're right, NoSuchBucket error is never thrown here because of `_get_cross_account_bucket ` logic, that is handling this case. `expected_bucket_owner `, instead, was not passed to `_get_cross_account_bucket `, hence the owner check was not performed.

I'm adding `expected_bucket_owner` as parameter for `_get_cross_account_bucket`, and removing the exceptions from here

localstack-core/localstack/services/s3/provider.py
cristiangiann: There is no test as of now. I will remove this piece and leave filter validation for a following PR.

localstack-core/localstack/services/s3/provider.py
bentsku: Ah, good catch! 🚀

tests/aws/services/s3/test_s3_api.py
cristiangiann: You are right, by properly running the tests against AWS I could easily spot this issue. I tried to run them before committing the changes but I did not realise I was still using my local implementation

tests/aws/services/s3/test_s3_api.py
cristiangiann: Sure, I'm adding a dedicated test

tests/aws/services/s3/test_s3_api.py
cristiangiann: Correct, I run it before the commit but with a wrong configuration. I'm now able to validate it with AWS

tests/aws/services/s3/test_s3_api.py
```diff
@@ -2180,3 +2180,162 @@ def test_multipart_if_match_etag(self, s3_bucket, aws_client, snapshot):
...
         snapshot.match("complete-multipart-if-match-overwrite-multipart", complete_multipart_1)
+
+
+class TestS3MetricsConfiguration:
+    @markers.aws.validated
+    def test_put_bucket_metrics_configuration(self, s3_bucket, aws_client, snapshot):
+        snapshot.add_transformer(
+            [
+                snapshot.transform.key_value("Id"),
+                snapshot.transform.key_value("HTTPStatusCode", reference_replacement=False),
```
bentsku: nit: as we would still like to properly validate the Status Code, it might better here to use a "snapshot skip" flag. 
Usually, we use the transformers for things we know will change between runs, like randomly generated ID, or things that by design are going to be different than AWS, like endpoints, host values, etc.

When things are not in parity with AWS, we tend to use a pytest marker to indicate that something is not like AWS, and we need to fix it later. It uses JSONPath to indicate which parts of the snapshot to skip. 

In this case, you could use the marker like so:
```python
        paths=[
            # PutBucketMetricsConfiguration should return 204, but we return 200
            "$.put_bucket_metrics_configuration.ResponseMetadata.HTTPStatusCode",
        ]
    )
    @markers.aws.validated
    def test_put_bucket_metrics_configuration(self, s3_bucket, aws_client, snapshot):
```

You can specify the "snapshot key" as the first part of the path, the one you specify as the first argument of `snapshot.match`, then the fields you want to skip. 
You can add this snapshot skip to the tests having the issue. 

You can then remove the transformer, and regenerate the snapshots 😄

bentsku (commented):
Very nice, this looks really great! We're almost good to go 🚀 thank you for addressing the comments really fast. 
The implementation is very nice, and ready to go! 

I only have one comment left regarding the issue you mentioned about the status code not being in line with AWS for `PutBucketMetricsConfiguration`. A ["snapshot skip" might be preferable](https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing#skipping-verification-of-snapshot-test) here instead of the transformation. 

Once this is addressed, we're ready! 🚀

tests/aws/services/s3/test_s3_api.py
cristiangiann: Something interesting is happening, and I'm not able to root cause it, at least at this time of the night. I tried to add the `skip_snapshot_verify` decorator, as you suggested, but my implementation does not work as expected, both in `test_put_bucket_metrics_configuration` and in `test_overwrite_bucket_metrics_configuration` tests.

For example:
```
    @markers.aws.validated
        paths=[
            # PutBucketMetricsConfiguration should return 204, but we return 200
            "$.put_bucket_metrics_configuration.ResponseMetadata.HTTPStatusCode",
        ]
    )
    def test_put_bucket_metrics_configuration(self, s3_bucket, aws_client, snapshot):
        snapshot.add_transformer([snapshot.transform.key_value("Id")])

        metric_id = short_uid()
        metrics_config = {"Id": metric_id, "Filter": {"Prefix": "logs/"}}

        put_result = aws_client.s3.put_bucket_metrics_configuration(
            Bucket=s3_bucket, Id=metric_id, MetricsConfiguration=metrics_config
        )
        snapshot.match("put_bucket_metrics_configuration", put_result)

        get_result = aws_client.s3.get_bucket_metrics_configuration(Bucket=s3_bucket, Id=metric_id)
        snapshot.match("get_bucket_metrics_configuration", get_result)
```

After updating again the snapshot, this test runs successfully locally, but fails when run against AWS with the following error:
```
_____TestS3MetricsConfiguration.test_put_bucket_metrics_configuration _____
>> match key: put_bucket_metrics_configuration
	(~) /ResponseMetadata/HTTPStatusCode 200 → 204 ... (expected → actual)

	["$..ResponseMetadata.HTTPStatusCode"]
```

I triple checked the path, which seems correct, and also tried some different ones. The test seems to work if I use the same path while adding a transformer with jsonpath, through `snapshot.transform.jsonpath`. I compared the logic with other usages of the skipper method in the codebase, but I was not able to find significant differences. I searched online, without success.

Are you able to spot any mistake?

This is how the updated snapshot looks like:
```
  "tests/aws/services/s3/test_s3_api.py::TestS3MetricsConfiguration::test_put_bucket_metrics_configuration": {
    "recorded-date": "13-06-2025, 00:01:27",
    "recorded-content": {
      "put_bucket_metrics_configuration": {
        "ResponseMetadata": {
          "HTTPHeaders": {},
          "HTTPStatusCode": 200
        }
      },
      "get_bucket_metrics_configuration": {
        "MetricsConfiguration": {
          "Filter": {
            "Prefix": "logs/"
          },
          "Id": "<id:1>"
        },
        "ResponseMetadata": {
          "HTTPHeaders": {},
          "HTTPStatusCode": 200
        }
      }
    }
  },
```

bentsku: This looks like you updated the snapshot against LocalStack, so that when you run it again against AWS, the snapshot verification fails because it expects the (wrong) status code LocalStack is returning. The "snapshot skip" only works against LocalStack, which is expected. 

When you run the test against LocalStack, make sure you're not using the `--snapshot-update` flag or the `SNAPSHOT_UPDATE=1` environment variable. Sorry for the inconvenience!

cristiangiann: Totally right, this is now fixed. I expected that AWS tests would have skipped it as well, but it totally makes sense.

Thank you for your patience!

bentsku (approved):
LGTM! Thanks a lot for addressing the comments so fast and for all the updates! This looks great, and is going to be a nice addition in S3. 🚀 

Thank you also for going the extra mile, and already adding pagination and some additional validation!

bentsku:
@cristiangiann The fix for the pipeline have been merged, so if you could rebase your branch on top of `master`, this would allow us to run the tests, would it be okay?

cristiangiann:
Sorry @bentsku, I just noticed your message. I just rebased it!

PR #12742, opened 2025-06-11
CloudFormation v2 Engine: Base Support for Serverless Global Transforms

MEPalma:
## Motivation
The introduction of the CloudFormation v2 engine laid the foundation for a redesigned engine capable of accurately determining update requirements between CloudFormation deployments, while also enabling parallel execution during updates. However, the current implementation does not support `AWS::Serverless` global transforms.

## Changes
- modeling support for updates to transform blocks
- introduce new transform visitor to inductively apply global transforms to both versions of the template
- port v1 logic for serverless global transforms
- minor refactor of provider and stack entities
- remove relevant CFNV2 test skips

## Future Work
- expand global transform support to macros (port from v1) and language extensions (in transform or preproc visitor class)
- investigate the behavior for updates to transform blocks
- other minors, see in-code TODO notes

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -1,7 +1,7 @@
 import copy
 import logging
 from datetime import datetime, timezone
-from typing import Any
+from typing import Any, Optional
 
 from localstack.aws.api import RequestContext, handler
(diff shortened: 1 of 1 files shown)


simonrw (approved):
We pair reviewed an earlier version of this. Nice clean changes thanks!

PR #12748, opened 2025-06-12
Update test durations automatically based on latest durations from master

k-a-il:
## Motivation

While switching from CircleCI to GitHub Actions, I saw that the test durations file hadn’t been updated in about a year. Since then, we’ve added a lot of new tests. This pull request adds a workflow that downloads test durations from the latest scheduled run on the master branch, combines them, and creates a PR with the updated test durations file. New workflow is scheduled to run every month.

## Changes

- Added a workflow to download and merge the test durations file, and create a pull request with the updates

## Testing

Tested in a forked repostiory.
- PR created automatically: https://github.com/k-a-il/localstack/pull/5
- Merged test durations uploaded as [artifact](https://github.com/k-a-il/localstack/actions/runs/15628284152)

### DIFF
--- a/.github/workflows/update-test-durations.yml
+++ b/.github/workflows/update-test-durations.yml
@@ -31,7 +31,7 @@ jobs:
             latest_workflow_id=$(curl -s https://api.github.com/repos/localstack/localstack/actions/workflows \
               | jq '.workflows[] | select(.name=="AWS / Build, Test, Push").id')
             latest_run_id=$(curl -s \
-              https://api.github.com/repos/localstack/localstack/actions/workflows/159258776/runs?branch=master&status=success&per_page=30 \
+              "https://api.github.com/repos/localstack/localstack/actions/workflows/${latest_workflow_id}/runs?branch=master&status=success&per_page=30" \
                 | jq '[.workflow_runs[] | select(.event == "schedule")][0].id')
             echo "Latest run: https://github.com/localstack/localstack/actions/runs/${latest_run_id}"
             echo "AWS_MAIN_LATEST_SCHEDULED_RUN_ID=${latest_run_id}" >> $GITHUB_ENV
@@ -43,7 +43,7 @@ jobs:
             path: artifacts-test-durations
             merge-multiple: true
             run-id: ${{ env.AWS_MAIN_LATEST_SCHEDULED_RUN_ID }}
-            github-token: ${{ secrets.GH_UPDATE_TEST_DURATIONS_TOKEN }}  # PAT with access to artifacts from GH Actions
+            github-token: ${{ secrets.GITHUB_TOKEN }}  # PAT with access to artifacts from GH Actions
 
         - name: Merge test durations files
           shell: bash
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Nice! This is really cool! The `update-test-durations.yml` might be something we want to consider generalizing and moving to our `meta` repository in the future to use it in other repositories as well 🤩

.github/workflows/update-test-durations.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v4
+          with:
+            path: localstack
+
+        - name: Latest run-id from community repository
+          run: |
+            latest_workflow_id=$(curl -s https://api.github.com/repos/localstack/localstack/actions/workflows \
+              | jq '.workflows[] | select(.name=="AWS / Build, Test, Push").id')
```
silv-io: suggestion: I think you can use the `aws-test.yml` filename here as well. This might make it more stable in regards to name changes

.github/workflows/update-test-durations.yml
```diff
@@ -0,0 +1,75 @@
...
+            latest_run_id=$(curl -s \
+              "https://api.github.com/repos/localstack/localstack/actions/workflows/${latest_workflow_id}/runs?branch=master&status=success&per_page=30" \
```
silv-io, line 34: q: What happens if the latest workflow doesn't have a successful attempt?

.github/workflows/update-test-durations.yml
```diff
@@ -0,0 +1,75 @@
...
+            title: "[Testing] Update test durations"
+            body: "This PR includes an updated `.test_durations` file, generated based on latest test durations from master"
+            branch: "test-durations-auto-updates"
+            author: "LocalStack Bot <localstack-bot@users.noreply.github.com>"
+            committer: "LocalStack Bot <localstack-bot@users.noreply.github.com>"
+            commit-message: "CI: update .test_durations to latest version"
+            path: localstack
+            add-paths: .test_durations
+            labels: "semver: patch, area: testing, area: ci"
+            token: ${{ secrets.GITHUB_TOKEN }}
```
silv-io: suggestion: You could use `PRO_ACCESS_TOKEN` here so that it's using the PAT of @localstack-bot. This way the PR author should be the bot instead of `github-actions`

silv-io (approved):
LGTM! Just some minor non-blocking suggestions :)

.github/workflows/update-test-durations.yml
k-a-il: yes, I can filter on the file path 👍

.github/workflows/update-test-durations.yml
k-a-il: Good catch, will change

.github/workflows/update-test-durations.yml
k-a-il: It will use the test durations from the most recent successful run on the master branch. If the latest run on master has failed, it will take test durations from the previous day's run

PR #12749, opened 2025-06-12
CI: correct test selection propagation for service-specific tests and run them conditionally

k-a-il:
## Motivation

This PR addresses two current issues with test selections, specifically:
- missing propagation of test selection files to service-specific tests
- issue with unnecessary job startup if no service-specific code was updated

## Changes

- Add `TESTSELECTION_PYTEST_ARGS` to `PYTEST_ARGS` for provider specific tests
- Add a step to identify changed services and use its output to conditionally trigger service-specific jobs

## Testing

- Test that Integration tests and DynamoDB-v2 tests are triggered https://github.com/localstack/localstack/pull/12751

### DIFF
--- a/.github/workflows/aws-tests.yml
+++ b/.github/workflows/aws-tests.yml
@@ -627,7 +627,7 @@ jobs:
 
   test-cloudwatch-v1:
     name: Test CloudWatch V1
-    if: ${{ !inputs.onlyAcceptanceTests && needs.test-preflight.outputs.cloudwatch-v1 == 'true' }}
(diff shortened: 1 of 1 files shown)


bentsku (commented):
I have a question regarding what happens if you try for example to do a change only in one service, which would conflict with the hardcoded `TEST_PATH` given in the specific job? Do we have any way to test for this?

Maybe another PR targeting this branch, with only a small change in some services, to see what happens in those alternative providers job? Do you think we could try that?

PR #12760, opened 2025-06-16
Bump moto-ext to 5.1.6.post2

viren-nadkarni:
## Summary

This PR bumps moto-ext to 5.1.6.post2

Contains:

- https://github.com/getmoto/moto/pull/8960

Moto response serialiser is being reworked upstream where now the tags field in IAM resources is always returned. AWS does not return this field if no tags are set. This is causing snapshot tests to fail, because of which this PR marks the tags attribute to be skipped. This is expected to be resolved in future Moto releases. See https://github.com/getmoto/moto/pull/8997.

## To do

- [x] Ext compatibility https://github.com/localstack/localstack-ext/actions/runs/15845021789
- [x] Multi-account/region compatibility https://github.com/localstack/localstack/actions/runs/15843314419

### DIFF
--- a/localstack-core/localstack/services/iam/provider.py
+++ b/localstack-core/localstack/services/iam/provider.py
@@ -335,7 +335,7 @@ class IamProvider(IamApi):
         backend = get_iam_backend(context)
         profile = backend.get_instance_profile(instance_profile_name)
         response = ListInstanceProfileTagsResponse()
-        response["Tags"] = [Tag(Key=k, Value=v) for k, v in profile.tags.items()]
+        response["Tags"] = profile.tags
         return response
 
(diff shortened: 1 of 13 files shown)


alexrashed (commented):
Awesome! Thanks a lot for keeping moto(-ext) up to date! 💯 
The changes are looking good to me, but I'll leave the final review to @dfangl since the only changes to a service provider is in IAM.

dfangl (approved):
I am not really happy with the snapshot ignores - this is strictly speaking, at least in part, a parity degradation after all, and we should follow up on this, by not including tags if they are not set at all.
For now this is fine for me though, to not hold back the merge :+1:

PR #12761, opened 2025-06-16
CloudFormation v2 Engine: Base Support for Global Macros

MEPalma:
## Motivation
The introduction of the CloudFormation v2 engine laid the foundation for a redesigned engine capable of accurately determining update requirements between CloudFormation deployments, while also enabling parallel execution during updates. However, the current implementation does not support global macros.

## Changes
- added integrations to v1 utils for the execution of global macros
- resolved issues with the computation of stack parameters, which resulted in sometimes malformed parameter objects during the transformation step (see compatibility requirements with v1 utils)
- resolved an issue with the normalization of transform parameters, which did not support some definition format 
- added change set update test using macros (see in-code limitations)
- updated the CFNV2 test annotations

## Future work
- unblock v1 tests by porting the create stack provider logic to use the v2 engine
- investigate some issue for which in our test setup a delete request for the template carrying the macro is registered before later events
- add further change set lifecycle tests involving global transformations

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/transformers.py
+++ b/localstack-core/localstack/services/cloudformation/engine/transformers.py
@@ -266,9 +266,6 @@ def execute_macro(
             message="Template format error: unsupported structure.. Rollback requested by user.",
         )
 
-    if result.get("fragment") is None:
-        pass
-
     return result.get("fragment")
 
 


simonrw (approved):
Amazing stuff, thanks for adding this!

PR #12767, opened 2025-06-17
Fix failing Ruby tests due to internal RUBYLIB variable changed by AWS

joe4dev:
## Motivation

Ruby transparent endpoint injection tests fail due to a change in the `RUBYLIB` variable. The official AWS Docker images do not match what we observe in AWS Lambda.

## Changes

* Ignore internal (not documented in Lambda [environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html)) `RUBYLIB` variable changed by AWS

### DIFF
--- a/tests/aws/services/lambda_/test_lambda_common.snapshot.json
+++ b/tests/aws/services/lambda_/test_lambda_common.snapshot.json
@@ -64,7 +64,7 @@
     "recorded-content": {}
   },
   "tests/aws/services/lambda_/test_lambda_common.py::TestLambdaRuntimesCommon::test_introspection_invoke[python3.8]": {
-    "recorded-date": "31-03-2025, 12:18:00",
+    "recorded-date": "17-06-2025, 09:51:26",
     "recorded-content": {
       "create_function_result": {
         "Architectures": [
@@ -205,7 +205,7 @@
     }
   },
   "tests/aws/services/lambda_/test_lambda_common.py::TestLambdaRuntimesCommon::test_introspection_invoke[java11]": {
-    "recorded-date": "31-03-2025, 12:18:12",
+    "recorded-date": "17-06-2025, 09:51:40",
     "recorded-content": {
       "create_function_result": {
         "Architectures": [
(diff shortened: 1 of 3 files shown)


tests/aws/services/lambda_/test_lambda_common.snapshot.json
```diff
@@ -549,21 +549,21 @@
           "AWS_LAMBDA_INITIALIZATION_TYPE": "on-demand",
           "AWS_LAMBDA_LOG_GROUP_NAME": "/aws/lambda/<function-name:1>",
           "AWS_LAMBDA_LOG_STREAM_NAME": "<log-stream-name:1>",
-          "AWS_LAMBDA_RUNTIME_API": "127.0.0.1:9001",
+          "AWS_LAMBDA_RUNTIME_API": "169.254.100.1:9001",
```
gregfurman, line 552: Should we not be transforming this IP address? The same applies to a couple of these other variables:

```
AWS_XRAY_DAEMON_ADDRESS
_AWS_XRAY_DAEMON_ADDRESS
```

gregfurman (approved):
LGTM! 

We should _probably_ look at improving our transformations of these envars going forward (especially since we don't set them). I'm concerned we're a bit too at the mercy of random variables changing and breaking our pipelines 🤦

gregfurman:
FYI, think this is the commit https://github.com/aws/aws-lambda-base-images/commit/8c4c65947e31ea270e9fb35aa99a070f74675d9f that introduced these changes. Seems like new layers were being added and Dockerfiles updated?

Could be worth setting up some internal monitoring for when these branches are pushed to 👀

tests/aws/services/lambda_/test_lambda_common.snapshot.json
joe4dev: I don't think we should transform these unless the value becomes dynamic. They seem stable, and having them raw allows us to detect upstream changes that may have other implications.

Example: https://github.com/localstack/localstack/blob/b4aed3e7e28df7b061ee4cbb471129900728d88c/tests/aws/services/lambda_/test_lambda_common.py#L146

joe4dev:
> LGTM!
> 
> We should _probably_ look at improving our transformations of these envars going forward (especially since we don't set them). I'm concerned we're a bit too at the mercy of random variables changing and breaking our pipelines 🤦

Good point. I added an item to our next Lambda service owner sync 👍

PR #12768, opened 2025-06-17
Added update functionality for AWS::SNS::Topic CFn Resource Provider

michalisfot:
## Motivation
This PR implements comprehensive update support for the AWS::SNS::Topic CloudFormation resource, addressing the core requirement to enable updates for DisplayName, TopicName, and Tags properties. The implementation follows AWS CloudFormation's update semantics, including resource replacement when immutable properties change.

### Enhanced Resource Provider (`aws_sns_topic.py`)

- __Implemented complete `update()` method__ with support for:

  - __In-place updates__: `DisplayName` attribute changes
  - __Resource replacement__: `TopicName` changes (recreates topic with new name)
  - __Tag management__: Add/remove/update tags with proper diffing
  - __State preservation__: Maintains subscriptions and existing tags during topic replacement

- __Key features__:

  - Proper ARN handling and validation
  - Graceful error handling for subscription/tag preservation
  - Follows AWS update semantics (replacement vs. in-place)
  - Resource cleanup (deletes old topic after successful replacement)

### Comprehensive Test Coverage (`test_sns.py`)

- __Created `test_sns_topic_update_attributes()`__ with in-place updates (no resource replacement):

  - __DisplayName updates__: Verifies `DisplayName` attribute changes are applied correctly
  - __Tag updates__: Tests adding, modifying, and removing tags
  - __Same topic preservation__: Ensures TopicArn remains unchanged during in-place updates
  - __Snapshot validation__: Captures before/after state for attribute and tag changes

- __Created `test_sns_topic_update_name()`__ with resource replacement:

  - __Tag preservation validation__: Verifies existing tags are preserved during replacement
  - __Subscription preservation validation__: Ensures subscriptions are migrated to new topic
  - __ARN format validation__: Builds and validates new ARN using proper AWS format
  - __Resource replacement verification__: Confirms old topic deletion and new topic creation
  - __Comprehensive snapshots__: Captures before/after state for all operations

## Testing
- On AWS:
  - `test_sns_topic_update_attributes` test succeeds but the `test_sns.snapshot.json` might need changes.
  - `test_sns_topic_update_name` test fails on AWS. Logs were not helpful due to very high verbosity even with `DEBUG=0`.
- Locally both test pass. 

## TODO
- Test against AWS and find out why `test_sns_topic_update_name` test fails on AWS but not locally.


michalisfot:

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,138 @@ def update(
           - sns:ListSubscriptionsByTopic
           - sns:GetDataProtectionPolicy
           - sns:PutDataProtectionPolicy
+          - sns:CreateTopic (for TopicName changes)
+          - sns:DeleteTopic (for TopicName changes)
```
simonrw: These docstrings are autogenerated by our scaffolding. It's interesting that there is a gap in the [specification](https://github.com/localstack/localstack/blob/958890a40ed4f5f1419a4cc69a85baa5c159c5cc/localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.schema.json#L131-L144), however not very surprising. 

We don't have a path for patching our specifications (like e.g. `cfn-lint`) so we expect discrepencies.

**suggestion**: can we make a note that this was not in the original spec?

```suggestion
          # NOT IN SPEC:
          - sns:CreateTopic (for TopicName changes)
          - sns:DeleteTopic (for TopicName changes)
```

tests/aws/services/cloudformation/resources/test_sns.snapshot.json
```diff
@@ -134,5 +134,349 @@
         }
       }
     }
+  },
+  "tests/aws/services/cloudformation/resources/test_sns.py::test_sns_topic_update_attributes": {
```
simonrw, line 138: We aim to [parity test](https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing) everything in LocalStack if possible. This means running our tests against AWS and capturing the snapshots of the _AWS_ responses. 

When done correctly, this process also generates/updates the `test_sns.validation.json` file. I don't see any changes to that file in this PR, so can I ask how you generated the changes to this file?

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../../templates/sns_topic_update.yaml"
+        ),
+    )
+
+    initial_topic_arn = stack.outputs["TopicArn"]
+
+    initial_attrs = aws_client.sns.get_topic_attributes(TopicArn=initial_topic_arn)
+    snapshot.match("initial-topic-attributes", initial_attrs)
+    assert initial_attrs["Attributes"]["DisplayName"] == "Initial Display Name"
```
simonrw: These assertions are not required with our snapshot library, since it already asserts fields are equal

```suggestion
```

tests/aws/templates/sns_topic_update.yaml
```diff
@@ -0,0 +1,28 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Parameters:
+  TopicName:
+    Type: String
+    Default: sns-topic-simple
```
simonrw: **suggestion**: I would prefer _not_ setting the defaults in the template, since it keeps all context in the body of the test, namely it's clear what's changed from one deployment to the next.

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+@markers.aws.unknown
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        "$..Attributes.DeliveryPolicy",
+        "$..Attributes.EffectiveDeliveryPolicy",
+        "$..Attributes.Policy.Statement..Action",
+        "$..Subscriptions..SubscriptionArn",
+    ]
+)
+def test_sns_topic_update_name(deploy_cfn_template, aws_client, snapshot):
```
simonrw, line 274: **issue:** ‏ this test fails when I run it against AWS. I get

```
        # Assert subscription was preserved with same endpoint and protocol
>       assert len(new_subscriptions["Subscriptions"]) == 1
E       assert 0 == 1
E        +  where 0 = len([])
```

Ideally these tests should be validated against AWS and should run both there and against LocalStack uneffected.

See our parity [testing documentation](https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing) for more details.

The issue is that updating the topic via CFn does *NOT* migrate the subscriptions if they are created outside of CloudFormation. (I have not tested it yet, but I think if you were to create an `AWS::SNS::Subscription` CloudFormation resource, then the subscription would be updated to subscribe to the new topic.)

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+        template_path=os.path.join(
+            os.path.dirname(__file__), "../../../templates/sns_topic_update.yaml"
+        ),
+    )
+
+    topic_arn = stack.outputs["TopicArn"]
+
+    initial_attrs = aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
+    snapshot.match("initial-topic-attributes", initial_attrs)
+    assert initial_attrs["Attributes"]["DisplayName"] == "Initial Display Name"
```
simonrw:
```suggestion
```

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+    initial_tags = aws_client.sns.list_tags_for_resource(ResourceArn=topic_arn)
+    snapshot.match("initial-topic-tags", initial_tags)
+
+    tag_dict = {tag["Key"]: tag["Value"] for tag in initial_tags["Tags"]}
+    assert tag_dict["Environment"] == "test"
+    assert tag_dict["Project"] == "localstack"
```
simonrw:
```suggestion
```

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+    deploy_cfn_template(
+        parameters={
+            "TopicName": topic_name,
+            "DisplayName": "Updated Display Name",
+            "Environment": "production",  # tag
+            "Project": "backend",  # tag
```
simonrw, line 245: **praise**: thanks for adding these clarifying comments, it really helps!

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+    snapshot.match("updated-topic-attributes", updated_attrs)
+
+    updated_tags = aws_client.sns.list_tags_for_resource(ResourceArn=topic_arn)
+    snapshot.match("updated-topic-tags", updated_tags)
+
+    assert updated_attrs["Attributes"]["DisplayName"] == "Updated Display Name"
+
+    updated_tag_dict = {tag["Key"]: tag["Value"] for tag in updated_tags["Tags"]}
+    assert updated_tag_dict["Environment"] == "production"
+    assert updated_tag_dict["Project"] == "backend"
```
simonrw:
```suggestion
```

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+@markers.aws.unknown
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        "$..Attributes.DeliveryPolicy",
+        "$..Attributes.EffectiveDeliveryPolicy",
+        "$..Attributes.Policy.Statement..Action",  # SNS:Receive is added by moto but not returned in AWS
+    ]
+)
+def test_sns_topic_update_attributes(deploy_cfn_template, aws_client, snapshot):
```
simonrw, line 204: In general the shape of this test looks fine. 

**suggestion:** perhaps consider validating the subscriptions as you do in the other test? Changing these attributes (that don't require replacement) should mean a subscription remains "attached", right?

**nit:** if a Python object (e.g. dictionary) is compared with `snapshot.match` you don't need to assert the individual fields of that object

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+        ),
+        stack_name=stack.stack_name,
+        is_update=True,
+    )
+
+    # Build the expected new ARN using the format: arn:aws:sns:REGION:ACCOUNT_ID:TOPIC_NAME
+    arn_parts = initial_topic_arn.split(":")
+    region = arn_parts[3]
+    account_id = arn_parts[4]
+    new_topic_arn = f"arn:aws:sns:{region}:{account_id}:{new_topic_name}"
```
simonrw: **suggestion**: this new Arn is available in `stack.outputs["TopicArn"]`. You will have to reassign the `stack` variable to be the result of calling `deploy_cfn_template` a second time.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,138 @@ def update(
...
         """
-        raise NotImplementedError
+        desired_state = request.desired_state
+        previous_state = request.previous_state
+        sns = request.aws_client_factory.sns
+
+        current_topic_arn = desired_state.get("TopicArn") or previous_state.get("TopicArn")
```
simonrw: **issue:** this is incorrect, the topic arn will _never_ be present in the `desired_state` as it's a property that cannot be specified by the user. It is also guaranteed to be there, however I don't mind the explicit `.get` then verify pattern.

```suggestion
        current_topic_arn = previous_state.get("TopicArn")
```

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,138 @@ def update(
...
+        if not current_topic_arn:
+            raise ValueError("TopicArn not found in desired_state or previous_state")
```
simonrw:
```suggestion
            raise ValueError("TopicArn not found previous_state")
```

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,138 @@ def update(
...
+
+        # Check if TopicName has changed (requires recreation)
+        desired_topic_name = desired_state.get("TopicName")
+        previous_topic_name = previous_state.get("TopicName")
```
simonrw, line 188: **suggestion:** the topic name should _always_ be present in the `previous_state`, so you can simplify some of the logic below if you ensure this.

```suggestion
        previous_topic_name = previous_state.get("TopicName")
        if not previous_topic_name:
            raise ValueError("Previous topic name not present")
```

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,138 @@ def update(
...
+                    account_id=request.account_id,
+                    region_name=request.region_name,
+                    desired_state=request.previous_state,
+                    logical_resource_id=request.logical_resource_id,
+                    resource_type=request.logical_resource_id,
+                    logger=request.logger,
+                    custom_context=request.custom_context,
+                    action=request.action,
+                )
+                self.delete(delete_request)
```
simonrw, line 258: I'm not sure how happy I am to directly call these methods. 

_Context:_ the behaviour of the resource provider executor explicitly invokes these methods multiple times in a loop as part of the deployment process. The basic shape looks like:

```python
while True:
    result = provider.create(payload)
    if result == "SUCCESS":
        break
    time.sleep(5)
```

and the methods of a resource provider _should_ look like:

```python
def create(payload):
    if is_first_time():
        create_resource()

    if resource_is_ready():
        return "SUCCESS"
    else:
        return "PENDING"
```

The SNS topic resource provider does not follow this pattern, so in _this case_ the implementation is fine, however with a different resource that _does_ implement this pattern, the resource provider executor could get stuck in a loop since the resource would be created, `resource_is_ready()` would return `False`, the `create` method would return `PENDING` and the status would never be checked.

For this resource provider I think it's ok because it is _not_ implemented with this approach in mind, but i might consider implementing this in a way that we continue to get status updates (perhaps returning early with the event so the outer polling loop could continue to drive the execution).

simonrw (changes requested):
Hi @michalisfot, thanks for this contribution 🎉 . I have a few major issues however:
* I'm not sure you captured the snapshots against AWS, particularly since one of the tests does not pass in its current form, but also there are no changes to the `test_sns.validation.json` file.
* Likely because of this first aspect, the behaviour in the test is incorrect. Subscriptions (at least created outside of CloudFormation) are not updated to target the new topic.


I have a few suggestions and nitpicks as well.

As an extension (i.e. a future PR) it would be nice to
* cover what happens to subscriptions when the topic is _not_ recreated
* include CloudFormation created `AWS::SNS::Subscription` resources and make sure they are updated correctly

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: Alright, I'll remove them.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
michalisfot: Yes, I'll add a comment

tests/aws/services/cloudformation/resources/test_sns.snapshot.json
michalisfot: The reason there is no changes to teh `test_sns.validation.json` file is that the snapshot was created manually by me. I realized ater submitting the PR that this is not the correct way of testing which is also mentioned on the testing documentation. Testing against AWS first I guess will fix that.

tests/aws/templates/sns_topic_update.yaml
michalisfot: Noted, I will remove it

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: So theoretically if the subscriptions were a part of the CF template this would work? I can try moving them in the template and test again.

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: Yes, I can align with the other test. Subscriptions should remain intact.

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: Got it, will do that.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
michalisfot: My idea was to avoid redundancy but I was missing this context. So what is the proposed change in this case?

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: I played around with the transformers but I couldn't make it work because of the extra AWS specific tags that are created along with the resource. The "Tags" attribute in the snapshot is a list of dicts and I couldn't find a way to exempt the AWS tags from the snapshot verification. So I falled back to checking them manually.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
simonrw: You have marked this comment as resolved, but not made any changes or explained why you have not made any changes. Can you go through your thought process here?

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
simonrw: > For this resource provider I think it's ok because it is not implemented with this approach in mind, but i might consider implementing this in a way that we continue to get status updates (perhaps returning early with the event so the outer polling loop could continue to drive the execution).


I am ok with this implementation in this resource provider for now as I stated above. 

I think the real implementation would involve extracting common functionality to helper methods. In fact the `read` method is an example of this: it's common to see resource providers call `self.read` at the end of the `create` and `update` methods to get the state of the resource from the infrastructure. The `delete` method is different, in that it is a re-entrant function call that needs to be driven from the outside.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
michalisfot: I'm not sure I follow. I followed the exact suggestion. The previous state was this:
```
        # Check if TopicName has changed (requires recreation)
        desired_topic_name = desired_state.get("TopicName")
        previous_topic_name = previous_state.get("TopicName")

        if desired_topic_name and previous_topic_name and desired_topic_name != previous_topic_name:
        ...
```

The check now is simplified to this: `if desired_topic_name != previous_topic_name:`

Was the change suggestion in regards to something else?

simonrw: Ok that's my misreading of this diff. I didn't see the update below. Thanks!

PR #12769, opened 2025-06-17
feat(handler): Add custom header to all AWS responses from LocalStack

gregfurman:
## Motivation
This PR adds a custom header to all service responses -- allowing us to distinguish whether a request orignates from LocalStack or AWS.

## Changes
- Adds a `LOCALSTACK_RESPONSE_HEADER_ENABLED` environment variable that is opt-out by default.
- When `LOCALSTACK_RESPONSE_HEADER_ENABLED` is truthy, a handler is added to the response chain that adds a `x-localstack: true` header to all responses from LocalStack gateway.

## Notes

The presence of the `x-localstack` header is used by the AWS Toolkit to detect LocalStack: https://github.com/aws/aws-toolkit-vscode/blob/c39a8f0f509701b29b13bb3ca37351faddb6f6ea/packages/core/src/auth/utils.ts#L894

### DIFF
--- a/tests/aws/conftest.py
+++ b/tests/aws/conftest.py
@@ -8,7 +8,6 @@ from localstack_snapshot.snapshots.transformer import RegexTransformer
 
 from localstack import config as localstack_config
 from localstack import constants
-from localstack.testing.pytest import markers
 from localstack.testing.snapshots.transformer_utility import (
     SNAPSHOT_BASIC_TRANSFORMER,
     SNAPSHOT_BASIC_TRANSFORMER_NEW,
@@ -108,32 +107,6 @@ def infrastructure_setup(cdk_template_path, aws_client):
     return _infrastructure_setup
 
 
-@pytest.hookimpl
-def pytest_runtest_setup(item):
-    """Modifies all tests using the snapshot fixture to skip validating 'x-localstack' header"""
-
-    # If snapshot fixture is not used, we don't need to skip anything.
-    if "snapshot" not in item.fixturenames:
-        return
-
-    # If snapshot fixture is used AND no skip_snapshot_verify marker is used, mark test to skip x-localstack.
-    is_localstack_header_path = f"$..{constants.HEADER_LOCALSTACK_IDENTIFIER}"
-    if not (skip_markers := list(item.iter_markers(name="skip_snapshot_verify"))):
-        item.add_marker(markers.snapshot.skip_snapshot_verify(paths=[is_localstack_header_path]))
-        return
-
-    # Otherwise, dynamically inject a path to the custom header inside all skip_snapshot_verify markers paths.
-    # If a path parameter is None, it's assumed that all paths are skipped.
-    for mark in skip_markers:
-        paths = mark.kwargs.get("paths")
-        if paths is None and len(mark.args) > 0:
-            paths = mark.args[0]
-
-        if paths and is_localstack_header_path not in paths:
-            mark.kwargs["paths"] = paths + [is_localstack_header_path]
-            continue
-
-
 @pytest.fixture(scope="function")
 def snapshot(request, _snapshot_session: SnapshotSession, account_id, region_name):
     # Overwrite utility with our own => Will be refactored in the future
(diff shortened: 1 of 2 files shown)


tiurin (approved):
Reviewed and tested env variable in a pairing session.

localstack-core/localstack/config.py
```diff
@@ -1245,6 +1245,9 @@ def use_custom_dns():
 # This flag enables `connect_to` to be in-memory only and not do networking calls
 IN_MEMORY_CLIENT = is_env_true("IN_MEMORY_CLIENT")
 
+# This flag enables all responses from LocalStack to contain a `x-localstack` HTTP header.
```
anisaoshafi, line 1248: _dismiss_: would rephrase this that flag "disables" this header from being set in response. default: True.

just a heads up to add the env var in public docs: https://docs.localstack.cloud/references/configuration/

tests/aws/conftest.py
```diff
@@ -107,6 +108,26 @@ def _infrastructure_setup(
     return _infrastructure_setup
 
 
+@pytest.hookimpl
+def pytest_runtest_setup(item):
```
anisaoshafi: would be nice to add a small high-level docstring/comment what this fixture was added for, so one can understand it in the code without having to open the PR where it was introduced for additional context.

e.g: """ fixture that skips header `x-localstack` from snapshot evaluation"""

tests/aws/conftest.py
```diff
@@ -107,6 +108,26 @@ def _infrastructure_setup(
...
+
+    # If snapshot fixture is used AND no skip_snapshot_verify marker is used, mark test to skip x-localstack.
+    is_localstack_header_path = f"$..{constants.HEADER_LOCALSTACK_IDENTIFIER}"
+    if not (skip_markers := list(item.iter_markers(name="skip_snapshot_verify"))):
+        item.add_marker(markers.snapshot.skip_snapshot_verify(paths=[is_localstack_header_path]))
+        return
+
+    # Otherwise, dynamically inject a path to the custom header inside all skip_snapshot_verify markers paths.
+    # If a path parameter is None, it's assumed that all paths are skipped.
+    for mark in skip_markers:
```
anisaoshafi: question: there are some snapshot tests [example](https://github.com/localstack/localstack-ext/blob/76ff58d34575580dbb7b5843855a668ce66f55c5/localstack-pro-core/tests/aws/services/cloudfront/test_cloudfront.py#L581C1-L588C34) where `path` parameter is not explicitly set, but just the list with the item is passed. Is that covered by these conditions? 

`
        [
            "$..Invalidation.Status",  # In real AWS it can be still in progress
            "$..Location",  # not currently supported
        ]
    )
`

anisaoshafi (approved):
nice and clean, good job @gregfurman 👏🏼 
I liked the logic for skipping the x-localstack header check from snapshot matching.

tests/aws/conftest.py
gregfurman: Huh that is a good question 🤔 Let me see if the `kwargs` accounts for this

gregfurman: OK damn. You're correct. This is not accounted for. Nice catch!

gregfurman: Update on this... looks like paths are _not_ skipped when added as args as opposed to kwargs 💀 

That means the test you linked _actually_ skips all paths.

tests/aws/conftest.py
```diff
@@ -118,6 +118,9 @@ def snapshot(request, _snapshot_session: SnapshotSession, account_id, region_nam
         RegexTransformer(f"arn:{get_partition(region_name)}:", "arn:<partition>:"), priority=2
     )
 
+    # Removes the 'x-localstack' header from all responses
+    _snapshot_session.add_transformer(_snapshot_session.transform.remove_key("x-localstack"))
```
dfangl, line 122: Nit: We could reference the constant for the header here

dfangl (approved):
LGTM!

tests/aws/conftest.py
gregfurman: Yeah fair. I originally did this but realised it's more explicit to add the string literal over referencing the constant. Will add in a follow-up.

PR #12771, opened 2025-06-18
refactored archive installer

macnev2013:
## Motivation

Refactors the `ArchiveDownloadAndExtractInstaller` class to improve modularity and adds checksum verification  support for enhanced security and reliability when downloading packages.

## Changes
- extracted reusable methods from `_install` method
  - `_download_and_extract_archive` - downloads and extracts archive
  - `_handle_single_directory_extraction` - handles logic for extracting all files in a single root directory
- improves cleanup with proper `try/finally` block
- Added checksum URLs for `FfmpegPackageInstaller` and `TerraformPackageInstaller`

- **Added checksum verification**: New `localstack.utils.checksum` module with:
  - Support for multiple checksum formats (Standard, BSD, Apache BSD)
  - Automatic algorithm detection (MD5, SHA1, SHA256, SHA512)
  - Comprehensive error handling with custom exceptions
  - Integration with `download_and_extract` function


localstack-core/localstack/utils/cheksum.py
alexrashed, line 1: **suggestion:** Seems like there's a little typo in this filename. `cheksum` -> `checksum` :)

localstack-core/localstack/packages/core.py
```diff
@@ -63,21 +64,47 @@ def _install(self, target: InstallTarget) -> None:
 
 
 class ArchiveDownloadAndExtractInstaller(ExecutableInstaller):
-    def __init__(self, name: str, version: str, extract_single_directory: bool = False):
+    def __init__(
+        self,
+        name: str,
+        version: str,
+        extract_single_directory: bool = False,
+        verify_archive_integrity: bool = False,
```
alexrashed: **suggestion:** I would suggest that the implementations of the installer (the subclasses) decide this based on the fact if `_get_checksum_url` and `_get_checksum_algo` is implemented.
i.e. if these functions return something, the verification should be executed.
What do you think?

localstack-core/localstack/packages/core.py
```diff
@@ -107,29 +134,83 @@ def get_executable_path(self) -> str | None:
...
+        if not os.path.isdir(target_subdir):
+            return
+        os.rename(target_subdir, f"{target_directory}.backup")
+        rm_rf(target_directory)
+        os.rename(f"{target_directory}.backup", target_directory)
+
+    def _verify_checksum(self, archive_path: str) -> None:
+        if not self.verify_archive_integrity:
+            LOG.debug("Skipping integrity verification (verify_integrity=False)")
+            return
```
alexrashed: **suggestion:** Based on the comment above (in the init dunder), I would suggest to have a try catch here only catching `NotImplemented` instead of introducing a member called `verify_archive_integrity`.

localstack-core/localstack/packages/core.py
```diff
@@ -107,29 +134,83 @@ def get_executable_path(self) -> str | None:
...
-            os.rename(target_subdir, f"{target_directory}.backup")
-            rm_rf(target_directory)
-            os.rename(f"{target_directory}.backup", target_directory)
+        try:
+            download_and_extract(
+                download_url,
+                retries=3,
+                tmp_archive=archive_path,
+                target_dir=target_directory,
+            )
```
alexrashed: **nitpick / question:** Would it make sense to extend the `download_and_extract` functionality with the verification feature? This way every archive download would benefit from your cool new feature to verify the integrity of the download.
This would also help implementing the best flow: Right now its (1) download, (2) extract, (3) check hash
In the best case it would be: (1) download, (2) check hash, (3) extract

localstack-core/localstack/packages/core.py
```diff
@@ -107,29 +134,83 @@ def get_executable_path(self) -> str | None:
...
+            self._verify_checksum(archive_path)
+            self._handle_single_directory_extraction(target_directory)
+        finally:
+            rm_rf(archive_path)
```
alexrashed, line 170: **suggestion:** If the hash check stays here (and the flow stays with (1) download, (2) extract, (3) check hash), then you have to delete the target folder here as well (as it might be extracted, but later on the hash check fails).

localstack-core/localstack/packages/core.py
alexrashed, line 1: **nit / suggestion:** I know this is being improved for a downstream project, but it would be really cool to see this being used in this project here as well. This would visualize the usage of this cool new feature, and would transitively test it as well. For example the `FfmpegPackageInstaller`?

localstack-core/localstack/packages/core.py
```diff
@@ -107,29 +134,83 @@ def get_executable_path(self) -> str | None:
...
+                    f"Could not parse checksum from {checksum_name}. "
+                    "Please check the checksum file format."
+                )
+
+            if not verify_file_integrity(
+                algorithm=self._get_checksum_algo(),
+                file_path=archive_path,
+                expected_checksum=expected_checksum,
+            ):
+                raise PackageException(
```
alexrashed: **nit:** I think the checksum related exceptions should be specific exceptions / a subclass of `PackageException` defined in this module.

localstack-core/localstack/packages/core.py
```diff
@@ -107,29 +134,83 @@ def get_executable_path(self) -> str | None:
...
+                    f"Checksum verification failed for {os.path.basename(archive_path)}. "
+                    "The downloaded file may be corrupted or tampered with."
+                )
+
+            LOG.info("Archive integrity verified successfully")
```
alexrashed: **nit:** I don't think it's necessary to log this on INFO.
```suggestion
            LOG.debug("Archive integrity verified successfully")
```

alexrashed (commented):
Thanks a lot for this great push! In my opinion, this is really a big step towards introducing integrity checks on a broader scope, which will enable us in the future to use external sources without lowering the trust, and we can avoid download issues / detect MITM attacks! 🚀 

Everything is looking good, I just added a few comments, questions, and nitpicks. Happy to discuss them directly :)

localstack-core/localstack/packages/core.py
macnev2013: Definitely - That's a great idea! I've added checksum verification to `FfmpegPackageInstaller` and `TerraformPackageInstaller`. I've update the PR with this implementation.

localstack-core/localstack/packages/core.py
macnev2013: Thanks for pointing this out. I've moved the hash check into the `download_and_extract` method.

localstack-core/localstack/utils/cheksum.py
macnev2013: Good catch! Fixed the typo. Thanks for spotting that!

localstack-core/localstack/packages/core.py
macnev2013: Great suggestion! You're right that it would be cleaner to detect this automatically based on whether the subclass implements these methods. I've updated it to make detect the verification based the method implemention.

localstack-core/localstack/packages/core.py
macnev2013: Updated this based on the implementation.

localstack-core/localstack/packages/core.py
macnev2013: Agreed. I've moved verification to `download_and_extract` to make integrity checks available across all the archives.

localstack-core/localstack/packages/core.py
macnev2013: With to the updated implementation, this is taken care of by `download_and_extract` method.

localstack-core/localstack/packages/core.py
macnev2013: Good catch. I've updated `LOG.info` to `LOG.debug`

localstack-core/localstack/utils/checksum.py
```diff
@@ -0,0 +1,349 @@
...
+    if checksum_length == 32:
+        algorithm = "md5"
+    elif checksum_length == 40:
+        algorithm = "sha1"
+    elif checksum_length == 64:
+        algorithm = "sha256"
+    elif checksum_length == 128:
+        algorithm = "sha512"
+    else:
+        raise ChecksumException(f"Unsupported checksum length: {checksum_length}")
```
alexrashed, line 292: **nit:** This seems a bit brittle and redundant. What if there are two different hash algorithms with the same size? Shouldn't the checksum parsers return the algorithm together with the hashes? Definitely not blocking a merge though :)

localstack-core/localstack/utils/checksum.py
```diff
@@ -0,0 +1,349 @@
...
+        LOG.error(
+            "Checksum mismatch for %s: calculated %s, expected %s",
+            file_path,
+            calculated_checksum,
+            expected_checksum,
+        )
+        raise ChecksumException(
+            f"Checksum mismatch for {file_path}: calculated {calculated_checksum}, expected {expected_checksum}"
+        )
+    LOG.info("Checksum verification successful for %s", file_path)
```
alexrashed: **nit:** I don't think this has to be logged on INFO level.

tests/unit/test_checksum.py
alexrashed, line 1: **praise:** Nice! Unit tests are perfect for this component! 💯

localstack-core/localstack/packages/ffmpeg.py
alexrashed, line 1: **praise:** Really nice to see how easy it is now to add the checksum verification to the package installers now! 💯

localstack-core/localstack/utils/archives.py
```diff
@@ -176,7 +177,19 @@ def download_and_extract(
...
+    """
+    Download and extract an archive to a target directory with optional checksum verification.
+
+    :param archive_url: URL of the archive to download
+    :param target_dir: Directory to extract the archive contents to
+    :param retries: Number of download retries (default: 0)
+    :param sleep: Sleep time between retries in seconds (default: 3)
+    :param tmp_archive: Optional path for the temporary archive file
+    :param checksum_url: Optional URL of the checksum file for verification
+    :raises Exception: If download fails, checksum verification fails, or archive format is unsupported
```
alexrashed: **nit:** I feel this comment is a bit redundant. I don't think there is any additional information to the method signature. What do you think?

localstack-core/localstack/utils/checksum.py
alexrashed, line 1: **nit:** I think this file contains quite a few comments that are redundant (because they are defined in the super method or because they do not add additional information to the method signature). While I am a big fan of comments where they add necessary information, they can also be distracting / cause an overload if they are really just there "to be there". 😛

localstack-core/localstack/utils/checksum.py
macnev2013: Good point. Changing this would require updating a lot of other places. Would be great if we can tackle this in a follow-up PR.

localstack-core/localstack/utils/checksum.py
macnev2013: Changed this to debug.

localstack-core/localstack/utils/archives.py
macnev2013: I've added some a bit more information about the optional checksum. I think it makes more sense to keep it now.

localstack-core/localstack/utils/checksum.py
macnev2013: Yep. Agreed. There was indeed a lot of duplication. I've cleaned it. Thanks for pointing out.

PR #12773, opened 2025-06-18
APIGW: add negative tests for IntegrationResponse and improve error handling in PutIntegrationResponse and GetIntegrationResponse

ArthurAkh:
## Motivation

Add negative test coverage for the `IntegrationResponse` operation in API Gateway v1, to ensure LocalStack properly mimics AWS behavior when invalid inputs are provided. This improves parity and robustness of the emulated service.

## Changes
- Added **negative test coverage** for `PutIntegrationResponse` and `GetIntegrationResponse` operations:
  - `test_integration_response_wrong_api`
  - `test_integration_response_wrong_resource`
  - `test_integration_response_wrong_method`
  - `test_integration_response_wrong_invalid_statuscode`
  - `test_integration_response_invalid_responsetemplates` (skipped for now due to LocalStack/AWS parity gap)
  - `test_integration_response_invalid_integration`
  - `test_integration_response_wrong_status_code`

- Updated `put_integration_response` and `get_integration_response` in `provider.py` to:
  - Validate `statusCode` using regex (`[1-5]\d\d`), raise `ValidationException` for invalid values
  - Raise appropriate `NotFoundException` messages for missing resources, methods, or integration responses

## TODO

What's left to do:

- [ ] Implement validation logic for `responseTemplates` in `put_integration_response` to ensure LocalStack raises appropriate exceptions (e.g., `SerializationException`, `BadRequestException`) when invalid values are provided
- [ ] Enable the currently skipped test `test_integration_response_invalid_responsetemplates` once LocalStack behavior matches AWS

### DIFF
--- a/localstack-core/localstack/aws/api/apigateway/__init__.py
+++ b/localstack-core/localstack/aws/api/apigateway/__init__.py
@@ -202,6 +202,12 @@ class LimitExceededException(ServiceException):
     retryAfterSeconds: Optional[String]
 
 
+class ValidationException(ServiceException):
+    code: str = "ValidationException"
+    sender_fault: bool = False
+    status_code: int = 400
+
+
 class NotFoundException(ServiceException):
     code: str = "NotFoundException"
     sender_fault: bool = False
--- a/localstack-core/localstack/aws/api/apigateway/__init__.py
+++ b/localstack-core/localstack/aws/api/apigateway/__init__.py
@@ -202,12 +202,6 @@ class LimitExceededException(ServiceException):
     retryAfterSeconds: Optional[String]
 
 
-class ValidationException(ServiceException):
-    code: str = "ValidationException"
-    sender_fault: bool = False
-    status_code: int = 400
-
-
 class NotFoundException(ServiceException):
     code: str = "NotFoundException"
     sender_fault: bool = False
--- a/localstack-core/localstack/services/apigateway/legacy/provider.py
+++ b/localstack-core/localstack/services/apigateway/legacy/provider.py
@@ -89,6 +89,7 @@ from localstack.aws.api.apigateway import (
     UsagePlan,
     UsagePlanKeys,
     UsagePlans,
+    ValidationException,
     VpcLink,
     VpcLinks,
 )
@@ -2059,6 +2060,28 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         status_code: StatusCode,
         **kwargs,
     ) -> IntegrationResponse:
+        if not re.fullmatch(r"[1-5]\d\d", status_code):
+            raise ValidationException(
+                "1 validation error detected: Value 'wrong' at 'statusCode' failed to satisfy constraint: Member must satisfy regular expression pattern: [1-5]\\d\\d"
+            )
+
+        try:
+            # get integration response doesn't return the right exception compared to AWS
+            moto_rest_api = get_moto_rest_api(context, rest_api_id)
+        except NotFoundException:
+            raise NotFoundException("Invalid Resource identifier specified")
+        moto_resource = moto_rest_api.resources.get(resource_id)
+        if not moto_resource:
+            raise NotFoundException("Invalid Resource identifier specified")
+
+        moto_method = moto_resource.resource_methods.get(http_method)
+        if not moto_method:
+            raise NotFoundException("Invalid Method identifier specified")
+
+        integration_response = moto_method.method_integration.integration_responses.get(status_code)
+        if not integration_response:
+            raise NotFoundException("Invalid Integration Response identifier specified")
+
         response: IntegrationResponse = call_moto(context)
         remove_empty_attributes_from_integration_response(response)
         # moto does not return selectionPattern is set to an empty string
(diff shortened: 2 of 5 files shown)


localstack-core/localstack/aws/api/apigateway/__init__.py
```diff
@@ -202,6 +202,12 @@ class LimitExceededException(ServiceException):
     retryAfterSeconds: Optional[String]
 
 
+class ValidationException(ServiceException):
+    code: str = "ValidationException"
+    sender_fault: bool = False
+    status_code: int = 400
+
+
```
bentsku: This file is auto-generated, and this change will be overwritten by our next ASF update (see the last one with #12785). 

If you need to define a new one, we can do it in a new `exceptions.py` file living in the `apigateway` module if there are several, or if only a few are missing, we can define them in the `provider.py` file.

When you use it only once, you can also use the `CommonServiceException` like this example:
https://github.com/localstack/localstack/blob/3d56935cbb820982c67c49b0e135f88d519c18d6/localstack-core/localstack/services/apigateway/legacy/provider.py#L2483-L2487

As we see that there are lots of `ValidationException`, you could define it like so:
```python
class ValidationException(CommonServiceException):
    def __init__(self, message=None):
        super().__init__("ValidationException", status_code=400, message=message)
```

localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -2059,17 +2060,37 @@ def get_integration_response(
...
+            raise NotFoundException("Invalid Resource identifier specified")
+
+        if not (moto_method := moto_resource.resource_methods.get(http_method)):
+            raise NotFoundException("Invalid Method identifier specified")
+
+        integration_response = None
+        if integration_responses := moto_method.method_integration.integration_responses:
+            integration_response = integration_responses.get(status_code)
+            if not integration_response:
+                raise NotFoundException("Invalid Integration Response identifier specified")
```
bentsku: If I understood well, if there are no `integration_responses` (the variable is set to `None`), we should fail, right?

So maybe we could update the logic to somewhat be one `if` with 2 checks:
```python
        if not moto_method.method_integration.integration_responses or not (integration_response := integration_responses.get(status_code)):
            raise NotFoundException("Invalid Integration Response identifier specified")
```

that way, you don't need to define the `integration_response = None`? Would this work? 
Maybe it would even be, not sure what `moto_method.method_integration` is set to when there is no integration defined, but I believe it'd be `None`.

```python
        if not moto_method.method_integration or not moto_method.method_integration.integration_responses or not (integration_response := integration_responses.get(status_code)):
            raise NotFoundException("Invalid Integration Response identifier specified")
```

This would also be a good test case to add: create a method, and try calling `GetIntegrationResponse` on a resource + method that does not have an integration set. Maybe this could be added?

tests/aws/services/apigateway/test_apigateway_api.py
bentsku, line 1: Great set of tests! 🚀

bentsku (changes requested):
Awesome work! This is looking great! 🚀 

I only have a few comments, the only blocking-one related to the modified auto-generated file, which is maybe lacking a disclaimer, as it's a common pitfall.

I also have a comment regarding the new changes required to make the other tests outside of this PR work, and maybe a suggestion or a follow-up we could do to write a test to trigger the scenario that was making your PR fail, but in a clear way and not with `ImportAPI` calls. It could help to avoid future regressions, and make the failing scenario clearer. See the comment related to that 😄 

All in all, great work, this is looking really nice, and this was not an easy task as we're directly manipulating objects and the flow is not very clear. Congrats for navigating this! 🚀

localstack-core/localstack/services/apigateway/legacy/provider.py
ArthurAkh: > If I understood well, if there are no integration_responses (the variable is set to None), we should fail, right?

Well, we will fail, but not right away, and not with that error `Invalid Integration Response identifier specified`! Instead we will fail inside of moto, as it's handled by moto, when we jump into `call_moto(context)`.
But you're right, maybe we can do something about that. So, right now:

- for example I've run the test `test_import_and_validate_rest_api(...)` line 352 inside of /localstack/localstack/tests/aws/services/apigateway/test_apigateway_import.py, with your suggested change of doing one long if statement, and it fails because we are raising that `Invalid Integration Response identifier specified` instead of `Invalid Response status code specified`

- but maybe instead I should go look into that test more precisely, and figure out if I can detect that error and raise it myself (`Invalid Response status code specified`)?

(Here, I was setting integration_response to None on line 2078, just because I reference it inside an if statement lower in the code.)

> This would also be a good test case to add: create a method, and try calling GetIntegrationResponse on a resource + method that does not have an integration set. Maybe this could be added?

Okay I'll look into that. You mean add that test in this PR, right?

bentsku: Yes, I think adding the test in this PR so that it's easier.

Now I understand a bit better!
Probably that long statement needs to be split into smaller one, and if there first one fail it raises the `Invalid Response status code`, and if it can go further it raises the `Invalid Integration Response identifier`, that one or the other way around 😄 

Being able to trigger the `Invalid Response status code specified` in a simpler test might be nice to try to understand when to raise that issue :smile:

localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -2059,17 +2059,40 @@ def get_integration_response(
...
-        if "selectionPattern" not in response:
-            moto_rest_api = get_moto_rest_api(context, rest_api_id)
-            moto_resource = moto_rest_api.resources.get(resource_id)
-            method_integration = moto_resource.resource_methods[http_method].method_integration
-            integration_response = method_integration.integration_responses[status_code]
-            if integration_response.selection_pattern is not None:
-                response["selectionPattern"] = integration_response.selection_pattern
+        if (
+            "selectionPattern" not in response
+            and integration_response
```
bentsku, line 2092: nit: this line could probably be removed because we are failing early if it not true (in line 2082)
`and integration_response`

But let's not block the merge, we can remove it in a follow-up PR!

bentsku (approved):
LGTM! Congrats, I know this was a very tricky one, wrangling moto internals and adding the validation in the right order! The CI is green and I've checked that our dependencies tests are also green, so this is good to merge! 🚀 awesome work!

PR #12774, opened 2025-06-18
APIGW: fix VTL $input.path and $input.json

bentsku:
## Motivation
I've noticed while working on other kind of VTL template that we were not handling `$input.path` and `$input.json` return values correctly when the body would be empty, or the JSONPath would not point to the correct value.

There were several reasons:
- the default `extract_jsonpath` helper method from LocalStack builds the result from a list, and return an empty list if there are no result. Because of this, we can't differentiate if we're actually fetching an empty list via the JSONPath, or we actually did not find a match. We now properly return `None` and verify behavior
- we would return an empty dict by default if there were no body, but this isn't the right behavior from AWS. We'd need new tests with request templates again to validate, I've checked it with the AWS console and multiple scenarios and external tests that this is the right behavior

## Changes
- use special `extract_jsonpath` for VTL rendering to stop returning empty list when no match
- stop returning an empty dict when there is no body when calling `input.path` or `input.body`
- add tests case
- update unit tests with finding from AWS console

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/template_mapping.py
@@ -23,6 +23,7 @@ from urllib.parse import quote_plus, unquote_plus
 
 import airspeed
 from airspeed.operators import dict_to_string
+from jsonpath_rw import parse
 
 from localstack import config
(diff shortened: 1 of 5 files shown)


cloutierMat (approved):
Awesome! :rocket: Thanks for the added test and the good pr comment! 

The fix is looking good, we are starting to iron out a lot of those edge cases :tada:

PR #12777, opened 2025-06-19
Sns topic update functionality

michalisfot:
Related PR: https://github.com/localstack/localstack/pull/12768

## Motivation
This PR implements basic update support for the AWS::SNS::Topic CloudFormation resource, addressing the core requirement to enable updates for DisplayName, TopicName, and Tags properties. The implementation follows AWS CloudFormation's update semantics, including resource replacement when immutable properties change.

### Enhanced Resource Provider (`aws_sns_topic.py`)

- __Implemented complete `update()` method__ with support for:

  - __In-place updates__: `DisplayName` attribute changes
  - __Resource replacement__: `TopicName` changes (recreates topic with new name)
  - __Tag management__: Add/remove/update tags with proper diffing
  - __State preservation__: Maintains subscriptions and existing tags during topic replacement

- __Key features__:

  - Proper ARN handling and validation
  - Graceful error handling for subscription/tag preservation
  - Follows AWS update semantics (replacement vs. in-place)
  - Resource cleanup (deletes old topic after successful replacement)

### Test Coverage (`test_sns.py`)

- __Created `test_sns_topic_update_attributes()`__ with in-place updates (no resource replacement):

  - __DisplayName updates__: Verifies `DisplayName` attribute changes are applied correctly
  - __Tag updates__: Tests adding, modifying, and removing tags
  - __Same topic preservation__: Ensures TopicArn remains unchanged during in-place updates
  - __Snapshot validation__: Captures before/after state for attribute and tag changes

- __Created `test_sns_topic_update_name()`__ with resource replacement:

  - __Tag preservation validation__: Verifies existing tags are preserved during replacement
  - __Subscription preservation validation__: Ensures subscriptions are migrated to new topic
  - __ARN format validation__: Builds and validates new ARN using proper AWS format
  - __Resource replacement verification__: Confirms old topic deletion and new topic creation
  - __Comprehensive snapshots__: Captures before/after state for all operations

## Testing
- On AWS and ocally both test pass. 

## TODO
- ~~Test against AWS and find out why `test_sns_topic_update_name` test fails on AWS but not locally.~~
- Find a proper way to validate the snapshot against the tags rather than manually checking each tag

### 1: Complete Core Properties 

1. __Archive Policy Support__ 
2. __Access Policy Support__
3. __KMS Key Management__ 
4. __FIFO Topic Properties__
5. __ContentBasedDeduplication__

### 2: Advanced Update Capabilities 

6. __Subscription Management__ 
7. __Data Protection Policy__ 
8. __Delivery Policy Configuration__ 

### 3: Robustness & Edge Cases 

9. __Error Handling Enhancement__ - Improve rollback scenarios for failed updates
11. __Performance Optimizations__


michalisfot:
recheck

michalisfot:
recheck

michalisfot:
I have read the CLA Document and I hereby sign the CLA.

alexrashed:
recheck

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,141 @@ def update(
...
+        previous_state = request.previous_state
+        sns = request.aws_client_factory.sns
+
+        current_topic_arn = previous_state.get("TopicArn")
+        if not current_topic_arn:
+            raise ValueError("TopicArn not found in previous_state")
+
+        # Check if TopicName has changed (requires recreation)
+        desired_topic_name = desired_state.get("TopicName")
+        previous_topic_name = previous_state.get("TopicName")
```
simonrw, line 188: See comment here: https://github.com/localstack/localstack/pull/12768/files#r2163728008

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
```diff
@@ -172,8 +172,141 @@ def update(
...
+                    account_id=request.account_id,
+                    region_name=request.region_name,
+                    desired_state=request.previous_state,
+                    logical_resource_id=request.logical_resource_id,
+                    resource_type=request.logical_resource_id,
+                    logger=request.logger,
+                    custom_context=request.custom_context,
+                    action=request.action,
+                )
+                self.delete(delete_request)
```
simonrw, line 258: See comment here: https://github.com/localstack/localstack/pull/12768/files#r2163737069

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+@markers.aws.validated
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        "$..Attributes.DeliveryPolicy",
+        "$..Attributes.EffectiveDeliveryPolicy",
+        "$..Attributes.Policy.Statement..Action",  # SNS:Receive is added by moto but not returned in AWS
+    ]
+)
+def test_sns_topic_update_attributes(deploy_cfn_template, aws_client, snapshot):
```
simonrw, line 260: In general the shape of this test looks fine. 

**suggestion:** perhaps consider validating the subscriptions as you do in the other test? Changing these attributes (that don't require replacement) should mean a subscription remains "attached", right?

**nit:** if a Python object (e.g. dictionary) is compared with `snapshot.match` you don't need to assert the individual fields of that object

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+    topic_arn = stack.outputs["TopicArn"]
+
+    initial_attrs = aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
+    snapshot.match("initial-topic-attributes", initial_attrs)
+
+    initial_tags = aws_client.sns.list_tags_for_resource(ResourceArn=topic_arn)
+    tag_dict = {tag["Key"]: tag["Value"] for tag in initial_tags["Tags"]}
+    assert tag_dict["Environment"] == "test"
+    assert tag_dict["Project"] == "localstack"
```
simonrw, line 291:
```suggestion
    snapshot.match("initial-tags", initial_tags)
```

We may need to add specific non-supported tags to the `skip_snapshot_verify` decorator at the top of the test function, as CFn adds its own tags to resources and we don't support that yet.

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+        is_update=True,
+    )
+
+    updated_attrs = aws_client.sns.get_topic_attributes(TopicArn=topic_arn)
+    snapshot.match("updated-topic-attributes", updated_attrs)
+
+    updated_tags = aws_client.sns.list_tags_for_resource(ResourceArn=topic_arn)
+    updated_tag_dict = {tag["Key"]: tag["Value"] for tag in updated_tags["Tags"]}
+    assert updated_tag_dict["Environment"] == "production"
+    assert updated_tag_dict["Project"] == "backend"
```
simonrw, line 316:
```suggestion
    snapshot.match("updated-tags", updated_tags)
```

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+    # Subscriptions should be preserved
+    snapshot.match("new-subscriptions", initial_subscriptions)
```
simonrw: **issue:** you are not refetching the subscriptions here, you are asserting the contents of the old subscriptions response. Can you update this snapshot to fetch the new subscriptions?

```suggestion
    snapshot.match("new-subscriptions", initial_subscriptions)
```

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+    initial_topic_arn = stack.outputs["TopicArn"]
+
+    initial_attrs = aws_client.sns.get_topic_attributes(TopicArn=initial_topic_arn)
+    snapshot.match("initial-topic-attributes", initial_attrs)
+
+    # Store initial tags to verify they are preserved
+    initial_tags = aws_client.sns.list_tags_for_resource(ResourceArn=initial_topic_arn)
+    initial_tag_dict = {tag["Key"]: tag["Value"] for tag in initial_tags["Tags"]}
+    assert initial_tag_dict["Environment"] == "test"
+    assert initial_tag_dict["Project"] == "localstack"
```
simonrw, line 363: See the same snapshot comment from above.

tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -191,3 +191,166 @@ def test_sns_subscription_region(
...
+
+    # Verify tags were preserved and updated
+    new_tags = aws_client.sns.list_tags_for_resource(ResourceArn=new_topic_arn)
+    new_tag_dict = {tag["Key"]: tag["Value"] for tag in new_tags["Tags"]}
+
+    # Assert tags were preserved (Project tag should still exist)
+    assert "Project" in new_tag_dict
+    assert new_tag_dict["Project"] == initial_tag_dict["Project"]  # Should be "localstack"
+    # Assert Environment tag was updated
+    assert new_tag_dict["Environment"] == "production"
```
simonrw, line 401: **suggestion:** use snapshot matching here

tests/aws/services/cloudformation/resources/test_sns.snapshot.json
```diff
@@ -134,5 +134,427 @@
...
+          "SubscriptionsDeleted": "0",
+          "SubscriptionsPending": "0",
+          "TopicArn": "<topic-arn:1>"
+        },
+        "ResponseMetadata": {
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 200
+        }
+      },
+      "initial-topic-tags": {
```
simonrw: These keys are not in the test anywhere, perhaps you need to regenerate your snapshots (after addressing my comments above).

tests/aws/services/cloudformation/resources/test_lambda.validation.json
simonrw, line 1: **issue:** this change is unrelated to this PR, can we undo this change?

simonrw (changes requested):
Thanks for posting the updates to this PR. I have left some follow ups on the [previous PR](https://github.com/localstack/localstack/pull/12768) and added some comments here. In general I am ok with this PR. I think we have captured a good part of the update implementation for SNS topics, including using CFn to manage the subscriptions.

I would be interested to capture the behaviour you had previously in a follow up where the subscriptions are created outside of CFn, however I think this use case is less common and therefore not as high a priority as the changes you've added here.

Thanks very much!

(I have requested changes again as I still have a couple of outstanding comments that have not been addressed. These are only minor though, I just would like your take on them before I approve).

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
michalisfot: Answered on the other PR.

localstack-core/localstack/services/sns/resource_providers/aws_sns_topic.py
michalisfot: Resolving as we are aligned on the current implementation.

tests/aws/services/cloudformation/resources/test_lambda.validation.json
michalisfot: Ah yes, it got away while testing. I'll roll it back.

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: That was my initial plan. I couldn't make it work though because of how the Tags are formatted. I couldn't find a way to skip the verification based on the Key. Also, AWS creates the tags in a random order and this also affects the validation. My idea was to temporarily check the tags manually until we find a proper solution (which might be already there but I missed it). I checked other examples but the tags there were formatted differently e.g. in Lambda snapshots tags are all in a single dictionary (not a list of dicts like bellow).

E.g.:
```
        "Tags": [
          {
            "Key": "Project",
            "Value": "localstack"
          },
          {
            "Key": "aws:cloudformation:stack-name",
            "Value": "replaced-value"
          },
          {
            "Key": "Environment",
            "Value": "test"
          },
          {
            "Key": "aws:cloudformation:logical-id",
            "Value": "TestTopic"
          },
          {
            "Key": "aws:cloudformation:stack-id",
            "Value": "replaced-value"
          }
        ],
```

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: https://github.com/localstack/localstack/pull/12777#discussion_r2164683943

tests/aws/services/cloudformation/resources/test_sns.py
michalisfot: This was done on purpose. I though that if the initial_subscriptions match the new subscriptions on the snapshot it means they are preserved. Do you mean we should do something like this?

```
    new_subscriptions = aws_client.sns.list_subscriptions_by_topic(TopicArn=topic_arn)
    snapshot.match("new-subscriptions", new_subscriptions)
```

tests/aws/services/cloudformation/resources/test_sns.py
simonrw: I think that's fair enough for now. Thanks for explaining.

tests/aws/services/cloudformation/resources/test_sns.py
simonrw: Yes this is what's required. You have already snapshotted the initial subscriptions, and the value of the variable does not change between your two `snapshot.match` calls. Therefore you are snapshotting the exact same information twice.

The snapshot library takes care of what the expected value should be. Provided you have captured the snapshot against AWS and the tests pass on LocalStack, we have confidence that the responses and behaviour match.

So you first get the subscriptions for the topic on create, then update the topic and re-fetch the subscriptions after the update. Whether they match is irrelevant, the fact that the snapshots are the same means that LocalStack does the same as what AWS does.

simonrw:
Hi @michalisfot , I've just merged in the master branch from LocalStack to resolve the merge conflicts. Provided the CI passes, I will approve and merge this work, thank you very much!

PR #12786, opened 2025-06-23
Fix workflows for forked repositories

silv-io:

### DIFF
--- a/.github/workflows/aws-tests.yml
+++ b/.github/workflows/aws-tests.yml
@@ -132,7 +132,7 @@ env:
   CI_COMMIT_SHA: ${{ github.sha }}
   CI_JOB_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}/attempts/${{ github.run_attempt }}
   # report to tinybird if executed on master
-  TINYBIRD_PYTEST_ARGS: "${{ github.ref == 'refs/heads/master' && '--report-to-tinybird ' || '' }}"
+  TINYBIRD_PYTEST_ARGS: "${{ github.repository == 'localstack/localstack' && github.ref == 'refs/heads/master' && '--report-to-tinybird ' || '' }}"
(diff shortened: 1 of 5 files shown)


alexrashed (commented):
Nice! Thanks for jumping on this! 💯 
Since this is skipping some of the tests / the execution of the tests on some of the environments, what are the implications of this? What do we need to be aware of when we pull changes from these private forks to our public repo here?
/cc @joe4dev

silv-io:
> Nice! Thanks for jumping on this! 💯 Since this is skipping some of the tests / the execution of the tests on some of the environments, what are the implications of this? What do we need to be aware of when we pull changes from these private forks to our public repo here? /cc @joe4dev

@alexrashed The implications are that we won't be able to catch ARM edgecases in a neutral CI environment. However, this is a risk we're already taking with regular PR work as there we don't run the ARM tests per default as well.

I suggest that while working on the forked repositories the contributors should run the affected tests locally on an ARM device. Also, I assume any changes which will make it upstream will also do so by a PR onto this repository. On that PR we will then need to run the ARM tests manually to make sure everything's fine before merging.

alexrashed:
> I suggest that while working on the forked repositories the contributors should run the affected tests locally on an ARM device.

I am not sure how feasible this is, but I guess it's the only way for now. Do you have an idea when the ARM runners will be enabled in private repos?

> Also, I assume any changes which will make it upstream will also do so by a PR onto this repository. On that PR we will then need to run the ARM tests manually to make sure everything's fine before merging.

That's a very good point. @joe4dev let's make sure that we leave a bit of a buffer and make sure that we explicitly run the ARM tests on any PRs which pull changes from private forks into the public repo.

alexrashed (approved):

silv-io:
> I am not sure how feasible this is, but I guess it's the only way for now. Do you have an idea when the ARM runners will be enabled in private repos?

[In a discussion thread an official account mentioned H2 2025](https://github.com/orgs/community/discussions/148648#discussioncomment-11920086), so hopefully soon :)

Issue #12783, opened 2025-06-21
bug: IfMatch Succeeds when it should fail


da1rren:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Hey Guys,

Despite providing an eTag that doesn't match the object the request still succeeds.

### Expected Behavior

The expected response would be for the localstack s3 api to throw a 412 Precondition Failed.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

docker compose up

```
services:
  localstack:
    container_name: localstack
    image: localstack/localstack:latest
    ports:
      - "127.0.0.1:4566:4566" # LocalStack Gateway
      - "127.0.0.1:4510:4510" # LocalStack Health endpoint (optional)
    environment:
      - SERVICES=s3,kms
      - DOCKER_HOST=unix:///var/run/docker.sock
    volumes:
      - ./src/scripts/:/etc/localstack/init/ready.d
      - /var/run/docker.sock:/var/run/docker.sock # Mount Docker socket for LocalStack
    healthcheck:
      test: [ "CMD", "curl", "-f", "http://localhost:4510/_localstack/health" ]
      interval: 30s
      timeout: 10s
      retries: 5
```

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```
#!/bin/bash
awslocal s3 mb s3://securemint --region eu-west-1
KEY_ID=`awslocal kms create-key  --region eu-west-1 --query "KeyMetadata.KeyId" --output text`
awslocal kms create-alias --region eu-west-1 --alias-name alias/securemint --target-key-id $KEY_ID
```

#### Golang Example
I'm using the following aws packages:
```
	github.com/aws/aws-sdk-go-v2 v1.36.5
	github.com/aws/aws-sdk-go-v2/config v1.29.17
	github.com/aws/aws-sdk-go-v2/service/kms v1.41.1
	github.com/aws/aws-sdk-go-v2/service/s3 v1.81.0
```

I've written the following test case:
```
func TestIfMatchPrecondition(t *testing.T) {
	cfg, err := config.LoadDefaultConfig(context.TODO())
	assert.Nil(t, err)

	s3Client := s3.NewFromConfig(cfg, func(o *s3.Options) {
		o.UsePathStyle = true
	})

	putResp, err := s3Client.PutObject(context.TODO(), &s3.PutObjectInput{
[...]

Issue #12184, opened 2025-01-25
feature request: S3 Directory Bucket support


avinassh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

Does Localstack support concept of directory buckets? I have my own custom client (for reasons), the same client works well with directory buckets on AWS, but not on localstack. There are subtle differences between S3 Express (which has directory buckets) and S3 Standard, so I am not sure how do I configure localstack to support directory buckets

bentsku:
Hello @avinassh! LocalStack does not currently support Directory Buckets. It's on our radar, but I can't give you an ETA yet. Thanks for the feature request!

avinassh:
@bentsku thanks for confirming! I look forward to the feature in localstack :)

HarshCasper:
Due to the forthcoming changes in our distribution and development model ([read more](https://blog.localstack.cloud/the-road-ahead-for-localstack/)), this issue tracker is no longer being maintained. We've migrated this issue to our internal tracker and will continue to work on it there. If you'd like to follow up on the status or expedite this, feel free to reach out through one of our support channels:

-   💬 [GitHub Discussions](https://github.com/orgs/localstack/discussions)
-   📖 [Help & Support](https://docs.localstack.cloud/aws/getting-started/help-support/)

PR #12787, opened 2025-06-23
S3: add error for DeleteObject preconditions for regular buckets

bentsku:
## Motivation
We got a report with #12783 that we did not support `IfMatch` header for `DeleteObject`. This is because this is only supported for Directory Bucket (S3 Express One Zone), which we don't support yet (requested at #12184). 

In the reported issue, the bucket was not created as a Directory Bucket, which would raise an exception in S3 when trying to use preconditions headers for `DeleteObject`. 

This PR adds that parity + a log message to indicate that directory buckets are not yet supported by LocalStack, to avoid users not understanding why such feature would just fail silently. 

## Changes
- add test for DeleteObject preconditions when the bucket is not a directory bucket
- add the same validation in the provider

## Side note regarding Directory Buckets

It seems AWS is adding more feature again only to Directory Buckets, as seen by #12785 and https://aws.amazon.com/about-aws/whats-new/2025/06/amazon-s3-express-one-zone-atomic-renaming-objects-api/

### DIFF
--- a/localstack-core/localstack/services/s3/provider.py
+++ b/localstack-core/localstack/services/s3/provider.py
@@ -1195,6 +1195,24 @@ class S3Provider(S3Api, ServiceLifecycleHook):
                 ArgumentName="x-amz-bypass-governance-retention",
             )
 
+        # TODO: this is only supported for Directory Buckets
+        non_supported_precondition = None
+        if if_match:
+            non_supported_precondition = "If-Match"
+        if if_match_size:
+            non_supported_precondition = "x-amz-if-match-size"
+        if if_match_last_modified_time:
+            non_supported_precondition = "x-amz-if-match-last-modified-time"
+        if non_supported_precondition:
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/s3/provider.py
```diff
@@ -1195,6 +1195,24 @@ def delete_object(
...
+            non_supported_precondition = "If-Match"
+        if if_match_size:
+            non_supported_precondition = "x-amz-if-match-size"
+        if if_match_last_modified_time:
+            non_supported_precondition = "x-amz-if-match-last-modified-time"
+        if non_supported_precondition:
+            LOG.warning(
+                "DeleteObject Preconditions is only supported for Directory Buckets. "
+                "LocalStack does not support Directory Buckets yet."
+            )
```
k-a-il, line 1210: Nice, this will improve user experience

PR #12793, opened 2025-06-24
Update STS session storage model to be more flexible

dfangl:
## Motivation
This PR contains no behavioral changes, just some refactoring of the STS session storage.
Instead of storing tags exclusively, I adapted naming and added additional properties to the session storage.

This will allow us to store more information per session, which is needed for advanced IAM functionality downstream.

## Changes
* Rename SessionTagConfig to SessionConfig to reflect storage of other attributes of a session
* Some renames in local variables
* Add `iam_context` field to allow storage of iam context variables


pinzon (approved):
Thank you for the refactoring. It looks promising :+1:

PR #12795, opened 2025-06-24
Update lambda runtime init, fix legacy ENV format in dockerfile

dfangl:
## Motivation
Our current go lambda init binary contains several vulnerabilities reported by scanners, including one designated as HIGH.
Even though they are not exploitable in LocalStack, we need to fix it.

This PR contains no behavioral changes:
- The lambda init is just a rebuild using a newer go version
- The ENV variable change is changing the deprecated syntax(https://github.com/docker/cli/pull/2743) to the proper one to avoid warnings with newer docker versions.

## Changes
- Update lambda init (rebuild without changes)
- Correct syntax of ENV instruction in dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
   && test ! $(which python3.9)
 
 SHELL [ "/bin/bash", "-c" ]
-ENV LANG C.UTF-8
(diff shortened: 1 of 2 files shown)


gregfurman (approved):
LGTM once CI passes 👍

alexrashed (approved):
🦸🏽

PR #12796, opened 2025-06-24
CFNv2: parameter resolution and presentation

simonrw:
## Motivation

We currently don't support the resolution of SSM parameters in the template `Parameters` when performing template describing, only executing. This does not match AWS behaviour, which represents the resolved values as well as the given parameters.

## Changes

* Move SSM parameter resolution to the preprocessing step as it's performed during `create_change_set`, so we should do it before describing and executing
* Add significant test for resolving SSM parameters
* Remove skip from cdk bootstrap since it's no longer failing
* Add structured type to parameters when describing so we can represent the original value (e.g. SSM parameter name) and resolved value (the SSM parameter value)

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -499,6 +499,18 @@ INTRINSIC_FUNCTIONS: Final[set[str]] = {
 }
 
 
+class ResolvedParameter:
+    value: str
+    resolved_value: str | None
+
+    def __init__(self, value: str, resolved_value: str | None = None):
+        self.value = value
+        self.resolved_value = resolved_value
+
+    def resolve(self) -> str:
+        return self.resolved_value or self.value
+
+
 class ChangeSetModel:
     # TODO: should this instead be generalised to work on "Stack" objects instead of just "Template"s?
 
@@ -1398,7 +1410,7 @@ class ChangeSetModel:
 
     @staticmethod
     def _is_terminal(value: Any) -> bool:
-        return type(value) in {int, float, bool, str, None, NothingType}
+        return type(value) in {int, float, bool, str, None, NothingType, ResolvedParameter}
 
     @staticmethod
     def _is_object(value: Any) -> bool:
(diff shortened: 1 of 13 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -1382,7 +1394,7 @@ def _type_name_of(value: Maybe[Any]) -> str:
 
     @staticmethod
     def _is_terminal(value: Any) -> bool:
-        return type(value) in {int, float, bool, str, None, NothingType}
+        return type(value) in {int, float, bool, str, None, NothingType, ResolvedParameter}
```
MEPalma, line 1413: Can you help me understand why we need this in the modeler?

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
```diff
@@ -304,6 +308,10 @@ def _resolve_reference(self, logical_id: str) -> PreprocEntityDelta:
         node_parameter = self._get_node_parameter_if_exists(parameter_name=logical_id)
         if isinstance(node_parameter, NodeParameter):
             parameter_delta = self.visit(node_parameter)
+            if isinstance(parameter_delta.before, ResolvedParameter):
```
MEPalma, line 343: Why do we need to first wrap the relevant values in a ResolvedParameter class and only evaluate them here?

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
```diff
@@ -908,6 +916,35 @@ def visit_node_parameter(self, node_parameter: NodeParameter) -> PreprocEntityDe
...
+                        )
+                    # we don't need to resolve the parameter here since it was done in the previous deployment
+                    after = before
+                else:
+                    resolved_value = resolve_ssm_parameter(
+                        account_id=self._change_set.account_id,
+                        region_name=self._change_set.region_name,
+                        stack_parameter_value=after,
+                    )
+                    after = ResolvedParameter(value=after, resolved_value=resolved_value)
```
MEPalma, line 966: why are we not resolving the resolved parameter here (this visitor function), but later?

simonrw:
Closing in favour of https://github.com/localstack/localstack/pull/12926 and an (as yet unwritten) implementation PR

PR #12798, opened 2025-06-25
CFn telemetry: capture actions on resources

simonrw:
## Motivation

We want to know which operations are being used with our resources. This change means that we capture the operation/action the user is performing on the resource. We want to see which resources are most common to be updated.

## Changes

When we load a resource provider, we capture which operation we are applying to that resource provider.

Note: for v1 and v2 we have to handle the case of both `str` and `ChangeAction`. This can be simplified when we remove the v1, *HOWEVER* we must not change the names of the event actions from `create`, `update` and `delete` or else we won't be able to map metrics captured from different times.


localstack-core/localstack/services/cloudformation/resource_provider.py
```diff
@@ -559,7 +560,7 @@ def execute_action(
                 raise NotImplementedError(change_type)  # TODO: change error type
 
     @staticmethod
-    def try_load_resource_provider(resource_type: str) -> ResourceProvider | None:
+    def try_load_resource_provider(action: str, resource_type: str) -> ResourceProvider | None:
```
dominikschubert: nit (but mostly just a comment): I get the intent but it does feel a bit weird passing it in here like this just for analytics purposes.

localstack-core/localstack/services/cloudformation/analytics.py
```diff
@@ -1,7 +1,57 @@
+import enum
+from typing import Self
+
+from localstack.aws.api.cloudformation import ChangeAction
 from localstack.utils.analytics.metrics import LabeledCounter
 
 COUNTER_NAMESPACE = "cloudformation"
+COUNTER_VERSION = 2
```
dominikschubert, line 8: nice, already coming in useful here!

dominikschubert (approved):
Thanks a lot for improving the resource provider usage analytics. This will help us a lot in prioritizing work on the resource providers :clap:

localstack-core/localstack/services/cloudformation/resource_provider.py
simonrw: I thought that when I wrote it. I might have a think about how to reorganise this information. I wonder if I can capture the same intent from the caller scope (where the `action` is available).

PR #12799, opened 2025-06-25
Add In_WSL function to check whether the host runs on WSL

paolosalvatori:
## Motivation
 On Windows 11 + WSL, you have a layered setup:

1. **Windows 11 Host:** Your primary operating system.
2. **WSL 2:** A lightweight virtual machine running a Linux kernel. Docker Desktop runs its daemon inside this WSL 2 VM.
3. **Docker Container:** Your Azurite container is running *inside* the Docker daemon's environment, which itself is in the WSL 2 VM.

The IP address of the Azurite container under `NetworkSettings.Networks.bridge.IPAddress`, let's say `172.17.02`, is the IP address of your Azurite container within the Docker bridge network.

This IP is only directly routable from other containers on the same Docker bridge network or from the Docker host itself (which, in this case, is the WSL 2 VM). Now, usually Docker on Windows is installed on a separate distro. You can check this by running the following command from the Windows command-prompt:

```bash
wsl --list --running
```

This should return something like this:

```bash
Windows Subsystem for Linux Distributions:
Ubuntu-24.04 (Default)
docker-desktop
```

As you can see, there are two distros: `Ubuntu-24.04 (Default)` running WSL, and `docker-desktop`. So far so good? Perfect. When running localhost in host mode in the WSL 2 distro's shell and not inside another Docker container, cannot use the IP address of the Azurite container. Full stop.

So I extended the `localstack.config` class with a function to understand whether the host is running on WSL:

```python
def is_wsl() -> bool:
    return platform.system().lower() == "linux" and os.environ.get("WSL_DISTRO_NAME") is not None
```

## Changes
I extended the `localstack.config` class with a function to understand whether the host is running on WSL:

```python
def is_wsl() -> bool:
    return platform.system().lower() == "linux" and os.environ.get("WSL_DISTRO_NAME") is not None
```

## Testing
On WSL the `is_wsl() ` returns `true`.


paolosalvatori:

localstack.sln
alexrashed, line 1: **issue:** I do not thing this is something that should be part of this PR?

localstack-core/localstack/config.py
```diff
@@ -276,6 +276,8 @@ def is_macos() -> bool:
 def is_windows() -> bool:
     return platform.system().lower() == "windows"
 
+def is_wsl() -> bool:
+    return platform.system().lower() == "linux" and os.environ.get("WSL_DISTRO_NAME") is not None
```
alexrashed, line 281: **question:** I do think that this is _only_ going to work in host mode of LocalStack, because `WSL_DISTRO_NAME` is not going to be set in the LocalStack Docker container when it is being started. So this really can only be used in the CLI, and at the same time, the CLI is not the only way to start LocalStack.
- How exactly do you intend this to be used?
- Are you sure you will be able to solve your issue with this function in the setting you intend to use it?
- Is there a more generic way to solve this Docker networking issue you are facing (what if a similar setup is used in other system scenarios than W11+WSL)?

In general, the function itself is looking good, a simple unit test could be useful though?

thrau (changes requested):
hey @paolosalvatori! thanks for being a WSL champion :-) we have few users on windows, but this may change with the growing azure team.

just two minor things:
* do we need the solutions file checked into the repo? we normally don't check in environment-specific files
* a more descriptive PR title for the commit history would be great ;-)

paolosalvatori:
> hey @paolosalvatori! thanks for being a WSL champion :-) we have few users on windows, but this may change with the growing azure team.
> 
> just two minor things:
> 
> * do we need the solutions file checked into the repo? we normally don't check in environment-specific files
> * a more descriptive PR title for the commit history would be great ;-)

I think I filed the PR from PyCharm to make an experiment, and I forgot to change the title to something meaningful. I think PyCharm turned the name of the branch `paolo_in_wsl` into the PR name 😄 I usually use very descriptive titles in my PRs.


Regarding the .sln file, I had the same issue on the other PR I opened on `localstack-ext`: I originally followed the instructions from [Getting Started – Setting up a Development Environment for LocalStack Core ](https://www.notion.so/localstack/Getting-Started-Setting-up-a-Development-Environment-for-LocalStack-Core-bb806ef9236e416d902667c43f88e027) on Notion and initially used Visual studio Code, then I switched to PyCharm. The `.sln` file was added by Visual studio Code. I removed the `.sln` file from the PR and added `*.sln` to `.gitignore`.

localstack.sln
paolosalvatori: I originally followed the instructions from [Getting Started – Setting up a Development Environment for LocalStack Core ](https://www.notion.so/localstack/Getting-Started-Setting-up-a-Development-Environment-for-LocalStack-Core-bb806ef9236e416d902667c43f88e027) on Notion and initially used Visual studio Code, then I switched to PyCharm. The `.sln` file was added by Visual studio Code. I removed the `.sln` file from the PR and added `*.sln` to `.gitignore`.

localstack-core/localstack/config.py
paolosalvatori: Yes, this is working only when localstack runs on WSL in host mode. As I explained on this [PR](https://github.com/localstack/localstack-ext/pull/4728), this is required to understand when localstack runs in host mode on WSL. In this case, localstack needs to use a different IP address to invoke the underlying Docker container. This line is from `proxy_request` method of the `AzuriteWrapper` in my PR](https://github.com/localstack/localstack-ext/pull/4728);

```python
host = "127.0.0.1" if is_in_wsl and not is_in_docker else self.docker_host
```

There is no way to solve the issue as this happens only when localstack runs in host mode on WSL. Please let me know if this answer your question. 

P.S. I can't close the other PR until I close this one, as the code of the other PR uses the code of this PR 😄

alexrashed (approved):
Thanks for jumping on the comments and answering the questions! The code is simple and lookin' good :)

paolosalvatori:
Thanks @alexrashed 🙂 @thrau when you have time, please give your approval, so I can unblock the other PR on `localstack-ext` repo as linting test fails saying that `is_in_wsl` attribute is not in `localstack.config` module 🙂 Thanks! 🙏

thomaschaaf:
@paolosalvatori I believe you mentioned the wrong person on your request. 🙃

paolosalvatori:
> @paolosalvatori I believe you mentioned the wrong person on your request. 🙃

Yep, wrong Thomas 🙂 I trusted GitHub when it proposed your alias instead of Thomas Rausch one 🙂

alexrashed:
@paolosalvatori You already have the approval from my side (@thrau needs to approve still though).
But you also do have linting errors here in the project.
Please make sure to fix those and configure the pre-commit hook (to check this before pushing a commit) to avoid issues like this in the future.

paolosalvatori:
@alexrashed thanks for the heads up. I ran `pre-commit install` on both `localstack` and `localstack-ext` projects and I fixed the linting issue affecting this PR. I know that you approved the PR 🙏 but Thomas approval is pending. He requested a change, so the PR cannot be merged until he approves it.

PR #12807, opened 2025-06-26
CloudFormation v2 Engine: Runtime Caching Support

MEPalma:
## Motivation
The introduction of the CloudFormation v2 engine laid the foundation for a redesigned engine capable of accurately determining update requirements between CloudFormation deployments, while also enabling parallel execution during updates. However, the new engine does not store or cache the evaluation of previous change set operations. This leads to  issues, such as the need to reevaluate global macros from earlier template versions. This process fails if the macro is no longer available, or it can result in corrupted states if the macro has been modified.
The absence of caching also forces full reevaluation of both template versions during a chain of change set updates, even when this information could simply be read from cache.
This PR introduces support for caching, resolving issues related to global transformations. It also adds utility functions for downstream processors to take advantage of the cache, applies these utilities in several scenarios within the preprocessor, and includes a negative test case for the global transformation scenario.

## Changes
- Added UpdateModel encapsulation for change set model visitors to handle runtime caching values
- Added logics to load and save the runtime caches
- Added utility function to utilize the cache within change set model visitors
- Ported a number of preprocessing visitors to using the cache
- Added related negative test for global transforms (see description above)
- Other related/minors

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
@@ -147,6 +147,22 @@ class ChangeSetNode(ChangeSetEntity, abc.ABC): ...
 class ChangeSetTerminal(ChangeSetEntity, abc.ABC): ...
 
 
+class UpdateModel:
+    # TODO: may be expanded to keep track of other runtime values such as resolved_parameters.
+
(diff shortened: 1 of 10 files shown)


simonrw (approved):
This PR was pair-reviewed with @MEPalma .

Issue #12530, opened 2025-04-16
bug: KMS decrypt fails for symmetric decryption when `KeyId` is provided


cj-christoph-gysin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When decrypting data encrypted using symmetric encryption, the keyId is included in the ciphertext blob and does not need to be provided. However, when providing the keyId, localstack fails with an error:

```
2025-04-16T08:48:25.069  INFO --- [et.reactor-2] localstack.request.aws     : AWS kms.Decrypt => 500 (InternalError)
2025-04-16T08:48:25.089 ERROR --- [et.reactor-1] l.aws.handlers.logging     : exception during call chain
Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/rolo/gateway/chain.py", line 166, in handle
    handler(self, self.context, response)
  File "/opt/code/localstack/localstack-core/localstack/aws/handlers/service.py", line 113, in __call__
    handler(chain, context, response)
  File "/opt/code/localstack/localstack-core/localstack/aws/handlers/service.py", line 83, in __call__
    skeleton_response = self.skeleton.invoke(context)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 154, in invoke
    return self.dispatch_request(serializer, context, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 168, in dispatch_request
    result = handler(context, instance) or {}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/aws/skeleton.py", line 118, in __call__
    return self.fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/services/kms/provider.py", line 1022, in decrypt
    plaintext = key.decrypt(ciphertext, encryption_context)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/localstack-core/localstack/services/kms/models.py", line 334, in decrypt
[...]

PR #12809, opened 2025-06-27
KMS #12530: Error handling for invalid blob  issue

pbansal2:
## Motivation
User reported a issue https://github.com/localstack/localstack/issues/12530 that causes a attribute error in Localstack. The error doesn't guide or provide clue to user on the problem. User is trying to decrypt the ciphertext blob by providing KeyId using Lambda function.
This PR fixes how LocalStack handles corrupted blobs during decryption when KeyId is provided by user. The fix ensures LocalStack's response matches AWS behavior.

## Changes
Localstack will mimic AWS behavior i.e. it will generate same response that AWS would generate. It will help user to validate the blob text that need to be decrypted using Localstack.

## Testing
Add a snaphshot test case in existing test method test_encrypt_decrypt_encryption_context() because AWS response is same. Newl response from AWS is added to test_kms.snapshot.json.

### DIFF
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -1043,7 +1043,8 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
             account_id, region_name, key_id = self._parse_key_id(key_id, context)
             try:
                 ciphertext = deserialize_ciphertext_blob(ciphertext_blob=ciphertext_blob)
-            except Exception:
+            except Exception as e:
+                logging.error("Error deserializing ciphertext blob: %s", e)
                 ciphertext = None
                 pass
         else:
@@ -1072,6 +1073,9 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
             if self._is_rsa_spec(key.crypto_key.key_spec) and not ciphertext:
                 plaintext = key.decrypt_rsa(ciphertext_blob)
             else:
+                # if symmetric encryption then ciphertext must not be None
+                if ciphertext is None:
+                    raise InvalidCiphertextException()
                 plaintext = key.decrypt(ciphertext, encryption_context)
         except InvalidTag:
             raise InvalidCiphertextException()
--- a/tests/aws/services/kms/test_kms.snapshot.json
+++ b/tests/aws/services/kms/test_kms.snapshot.json
@@ -1565,7 +1565,7 @@
     }
   },
   "tests/aws/services/kms/test_kms.py::TestKMS::test_encrypt_decrypt_encryption_context": {
-    "recorded-date": "11-05-2023, 22:46:49",
+    "recorded-date": "08-07-2025, 05:53:27",
     "recorded-content": {
       "encrypt_response": {
         "CiphertextBlob": "ciphertext-blob",
@@ -1579,6 +1579,7 @@
       "decrypt_response_with_encryption_context": {
         "EncryptionAlgorithm": "SYMMETRIC_DEFAULT",
         "KeyId": "<key-id:1>",
+        "KeyMaterialId": "e2333676b9bf055cb0caa2bec3957d7f3e60b7545a3706314e397746cd26122e",
         "Plaintext": "plaintext",
         "ResponseMetadata": {
           "HTTPHeaders": {},
@@ -1586,6 +1587,16 @@
         }
       },
       "decrypt_response_without_encryption_context": {
+        "Error": {
+          "Code": "InvalidCiphertextException",
+          "Message": ""
+        },
+        "ResponseMetadata": {
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 400
+        }
+      },
+      "decrypt_response_with_invalid_ciphertext": {
         "Error": {
           "Code": "InvalidCiphertextException",
           "Message": ""
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -14,6 +14,7 @@ from cryptography.hazmat.primitives.asymmetric import ec, padding, utils
 from cryptography.hazmat.primitives.serialization import load_der_public_key
 
 from localstack.services.kms.models import (
+    HEADER_LEN,
     IV_LEN,
     ON_DEMAND_ROTATION_LIMIT,
     Ciphertext,
@@ -1787,7 +1788,7 @@ class TestKMS:
         snapshot.match("invalid-plaintext-size-encrypt", e.value.response)
 
     @markers.aws.validated
-    @markers.snapshot.skip_snapshot_verify(paths=["$..message"])
+    @markers.snapshot.skip_snapshot_verify(paths=["$..message", "$..KeyMaterialId"])
     def test_encrypt_decrypt_encryption_context(self, kms_create_key, snapshot, aws_client):
         key_id = kms_create_key()["KeyId"]
         message = b"test message 123 !%$@ 1234567890"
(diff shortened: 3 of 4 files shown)


pbansal2:

tests/aws/services/kms/test_kms.snapshot.json
simonrw, line 1: **question**: when I re-run the tests against AWS, I find that a `KeyMaterialId` field in the `decrypt_response_with_encryption_context` block of this snapshot is added, which is not captured in this PR. Can you explain why? Did you execute these tests against AWS, or manually update the snapshot file?

You can see more details about our parity testing approach here: https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing

localstack-core/localstack/services/kms/provider.py
simonrw, line 1: **question**: do you think it's worth logging an error message on line 1047 where we catch and silence any exceptions?

pbansal2: Yes, agree. I should have done it. We should log the error before line 1047 to avoid hiding potential bugs caused by catching all exceptions without logging.

tests/aws/services/kms/test_kms.snapshot.json
pbansal2: I executed the tests against AWS that generated the new block (`decrypt_response_with_invalid_ciphertext`) in `test_kms.snapshot.json`. I don't expect any changes in response for `decrypt_response_with_encryption_context `with my changes in test.
Let me test this again to double check.

pbansal2: @simonrw I am not seeing `KeyMaterialId` or any other changes in block `decrypt_response_with_encryption_context` in `test_kms_snapshot.json`
I am testing with following command after setting required env variables to update the snapshot and to execute the against the AWS. I can also see keys created in my AWS account.
`TEST_PATH="tests/aws/services/kms/test_kms.py::TestKMS::test_encrypt_decrypt_encryption_context" make test`
This is updated snapshot block after running above test:
```
diff --git a/tests/aws/services/kms/test_kms.snapshot.json b/tests/aws/services/kms/test_kms.snapshot.json
index 17ebf79f2..e33911e7e 100644
--- a/tests/aws/services/kms/test_kms.snapshot.json
+++ b/tests/aws/services/kms/test_kms.snapshot.json
@@ -1565,7 +1565,7 @@
     }
   },
   "tests/aws/services/kms/test_kms.py::TestKMS::test_encrypt_decrypt_encryption_context": {
-    "recorded-date": "11-05-2023, 22:46:49",
+    "recorded-date": "01-07-2025, 17:01:36",
     "recorded-content": {
       "encrypt_response": {
         "CiphertextBlob": "ciphertext-blob",
@@ -1594,6 +1594,16 @@
           "HTTPHeaders": {},
           "HTTPStatusCode": 400
         }
+      },
+      "decrypt_response_with_invalid_ciphertext": {
+        "Error": {
+          "Code": "InvalidCiphertextException",
+          "Message": ""
+        },
+          "HTTPHeaders": {},
+          "HTTPStatusCode": 400
+        }
       }
     }
   },

```
could you please suggest anything wrong with these steps test against AWS?

simonrw: If you look at the [How to write parity tests](https://github.com/localstack/localstack/tree/master/docs/testing/parity-testing#how-to-write-parity-tests) section of our docs, you should find the instructions on how to run the tests against AWS.

Specifically:

> Run the test against AWS: use the parameter `--snapshot-update` (or the environment variable `SNAPSHOT_UPDATE=1`) and set the environment variable as `TEST_TARGET=AWS_CLOUD`.

You have run the tests with your AWS credentials, however you are not configuring the LocalStack test suite to target AWS.

> I am testing with following command after setting required env variables to update the snapshot and to execute the against the AWS. I can also see keys created in my AWS account.
TEST_PATH="tests/aws/services/kms/test_kms.py::TestKMS::test_encrypt_decrypt_encryption_context" make test

I can't really explain what is going on here unfortunately. You are clearly 
* updating the snapshot with `SNAPSHOT_UPDATE=1` since it has changed, and
* targeting AWS with `TEST_TARGET=AWS_CLOUD` since you say the keys are created (but they should only be temporary, they should be scheduled for deletion after the test completes, right?)

but when I re-run the test with the same configuration, I find the `KeyMaterialId` property has been added

```diff
diff --git i/tests/aws/services/kms/test_kms.snapshot.json w/tests/aws/services/kms/test_kms.snapshot.json
index e97a3097e6d7..33e987b371d7 100644
--- i/tests/aws/services/kms/test_kms.snapshot.json
+++ w/tests/aws/services/kms/test_kms.snapshot.json
@@ -1565,7 +1565,7 @@
     }
   },
   "tests/aws/services/kms/test_kms.py::TestKMS::test_encrypt_decrypt_encryption_context": {
-    "recorded-date": "27-06-2025, 08:31:49",
+    "recorded-date": "04-07-2025, 10:40:02",
     "recorded-content": {
       "encrypt_response": {
         "CiphertextBlob": "ciphertext-blob",
@@ -1579,6 +1579,7 @@
       "decrypt_response_with_encryption_context": {
         "EncryptionAlgorithm": "SYMMETRIC_DEFAULT",
         "KeyId": "<key-id:1>",
+        "KeyMaterialId": "4ba2a196f27b13af15cce4730e67b632eff7b53b68f987b7331d209fddc0b103",
         "Plaintext": "plaintext",
         "ResponseMetadata": {
           "HTTPHeaders": {},

```

> I executed the tests against AWS that generated the new block (decrypt_response_with_invalid_ciphertext) in test_kms.snapshot.json. I don't expect any changes in response for decrypt_response_with_encryption_context with my changes in test.

If you are executing the test against AWS and capturing snapshots, then it captures _all_ `snapshot.match` values, not just the one that you have changed. So it's not unexpected that even though you have not changed the code that contributes to the `decrypt_response_with_encryption_context` snapshot content, that it may change. 

This is a challenge we face regularly, you see the last time this snapshot was executed was 2023! We regularly find the AWS responses have changed since the last time the snapshot was captured.

When I re-run the entire kms test suite against AWS I see
* other snapshot responses have additional keys (also relating to key material so likely related)
* some tests fail which is interesting, they need revisiting

localstack-core/localstack/services/kms/provider.py
simonrw: Do you think this change should be part of this PR or a separate one?

pbansal2: I have made this change in this PR only.

tests/aws/services/kms/test_kms.snapshot.json
pbansal2: > targeting AWS with TEST_TARGET=AWS_CLOUD since you say the keys are created (but they should only be temporary, they should be scheduled for deletion after the test completes, right?)

Yes, they were scheduled for deletion right after test was finished.

> If you are executing the test against AWS and capturing snapshots, then it captures all snapshot.match values, not just the one that you have changed. So it's not unexpected that even though you have not changed the code that contributes to the decrypt_response_with_encryption_context snapshot content, that it may change.

Understood. Thanks!
 I also tried running the entire KMS test suite but couldn't see `keymaterialId` in `decrypt_response_with_encryption_context` or any other test function in KMS suite.
I can also see few tests failing in KMS suite.

pbansal2: I tested two other setups as well to repro it. It seems some issue in my Localstack set up probably.

1. Outside Localstack: I can see`keymaterialId` in the response of `decrypt()` from AWS.
2. Within Localstack using `awslocal` CLI: Same issue as KMS test suite where there is no` keymaterialId` from decrypt  with encryption context.

pbansal2: @simonrw It seems that earlier I added the fix and test the changes on Localstack version 4.3 where the issue was raised by user. However, probably I didn't test it again on `master` before commiting my changes.

I have tested again on master and able to see `keyMaterialId` in response for `decrypt_response_with_encryption_context`. I have made this change in this PR now. Please review it once. Thank you!

PR #12810, opened 2025-06-27
SQS: add automatic propagation of AWSTraceHeader attribute when sending message

bentsku:
## Motivation
While reading about SQS and X-Ray, I had a hunch we didn't implement the automatic passing of X-Ray headers to the Attributes. 
See https://docs.aws.amazon.com/xray/latest/devguide/xray-services-sqs.html

It might have been the case in the past, due to this following function:
https://github.com/localstack/localstack/blob/5b0180cbf277936f83334c7a61e1f63f18e9c01e/localstack-core/localstack/utils/aws/aws_responses.py#L189-L196

But this was lacking now.
I've added a test to verify the behavior in AWS. 

## Changes
- add automatic propagation from incoming request headers to `AWSTraceHeader` Attribute


thrau (approved):
LGTM! thanks for adding this!

PR #12824, opened 2025-07-02
LDM as provisioner of debug-enabled containers

MEPalma:
## Motivation
The current LDM implementation depends heavily on the Lambda service’s container provisioning logic, injecting LDM-specific parameter overrides throughout various parts of the service to support debug-enabled containers. This fragmented integration makes LDM difficult to maintain and limits control over the lifecycle of debug-enabled containers. This refactor centralizes LDM responsibilities by transforming it into a dedicated provisioner for debug-enabled containers. As a result, the Lambda service is relieved of most LDM-related logic. Additionally, these changes ensure that debug-enabled containers persist across multiple invocations, allowing the debug client to remain connected or reconnect after disconnections. 

## Changes
- LDM as a debug-enabled container provisioner
- removed LDM fragments in the lambda provider
- introduce new ExecutionEnvironment subtype for container-enabled execution environments 
- avoid container restarts between debug invokes, for compatibility python lambda functions must include a revised debug server logic:

```python
def handler(event, context):
    print(event)
    return event

def wait_for_debug_client(port: int=19891, timeout: int=3600):
    import time, threading
    import sys, glob
    sys.path.append(glob.glob(".venv/lib/python*/site-packages")[0])
    import debugpy

    if not hasattr(wait_for_debug_client, "_debugpy_listening"):
        wait_for_debug_client._debugpy_listening = False

    if not wait_for_debug_client._debugpy_listening:
        try:
            debugpy.listen(("0.0.0.0", port))
            wait_for_debug_client._debugpy_listening = True
            print(f"debugpy is now listening on port {port}")
        except RuntimeError as e:
            print(f"debugpy.listen() failed or already active: {e}")

    if not debugpy.is_client_connected():
        print("Waiting for client to attach debugger...")

        def cancel_wait():
            time.sleep(timeout)
            print("Canceling debug wait task after timeout...")
            debugpy.wait_for_client.cancel()

        threading.Thread(target=cancel_wait, daemon=True).start()
        debugpy.wait_for_client()
    else:
        print("Debugger already attached.")

wait_for_debug_client()

```
- several other related changes


localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm_utils.py
```diff
@@ -0,0 +1,42 @@
...
+          - arn:aws:lambda:eu-central-1:000000000000:function:functionname: ->
+              exception UnknownLambdaArnFormat
+
+          - arn:aws:lambda:eu-central-1:000000000000:function ->
+              exception UnknownLambdaArnFormat
+    """
+
+    if not lambda_arn:
+        return lambda_arn
+    lambda_arn_parts = lambda_arn.split(":")
```
gregfurman, line 26: nit: Should we not specifically be checking if the service is `lambda`? Perhaps we should use `api_utils.function_locators_from_arn` or `localstack.utils.aws.arns.parse_arn`.

PR #12826, opened 2025-07-02
APIGW: add support for new endpoint configuration

ArthurAkh:
## Motivation

AWS recently introduced the `ipAddressType` field for API Gateway Rest APIs, defaulting to `"ipv4"`.  
LocalStack needs to mimic this behavior to maintain parity.

Related docs:
- [AWS Blog Post](https://aws.amazon.com/blogs/aws/amazon-api-gateway-now-supports-dual-stack-ipv4-and-ipv6-endpoints/)
- [API Gateway IP Address Type](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-ip-address-type.html)

## Changes

- Added support for returning `endpointConfiguration.ipAddressType` in `CreateRestApi`.
- Added handling of the `replace` operation for `types` and `ipAddressTypes` in `UpdateRestApi`.
- Added validation logic.
- Updated relevant snapshot tests.

## TODO

What's left to do:

- [ ] Currently a skip `"$..endpointConfiguration.ipAddressType"` was added in the `test_create_execute_api_vpc_endpoint`, because regenerating its snapshot wasn't working.

### DIFF
--- a/localstack-core/localstack/services/apigateway/legacy/provider.py
+++ b/localstack-core/localstack/services/apigateway/legacy/provider.py
@@ -251,7 +251,9 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
         ip_address_type = endpoint_configuration.get("ipAddressType")
 
         if not types:
-            raise BadRequestException("REGIONAL Configuration and EDGE Configuration cannot be both DISABLED.")
+            raise BadRequestException(
+                "REGIONAL Configuration and EDGE Configuration cannot be both DISABLED."
+            )
         elif len(types) > 1:
             raise BadRequestException("Cannot create an api with multiple Endpoint Types.")
         endpoint_type = types[0]
@@ -405,7 +407,11 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
                     raise BadRequestException(
                         "Invalid patch operation specified. Must be 'add'|'remove'|'replace'"
                     )
-                if patch_op.get("value") not in (EndpointType.REGIONAL, EndpointType.EDGE, EndpointType.PRIVATE):
+                if patch_op.get("value") not in (
+                    EndpointType.REGIONAL,
+                    EndpointType.EDGE,
+                    EndpointType.PRIVATE,
+                ):
                     raise BadRequestException(
                         "Invalid EndpointTypes specified. Valid options are REGIONAL,EDGE,PRIVATE"
                     )
@@ -424,7 +430,10 @@ class ApigatewayProvider(ApigatewayApi, ServiceLifecycleHook):
                     raise BadRequestException(
                         "Invalid patch operation specified. Must be one of: [replace]"
                     )
-                if (ipAddressType := patch_op.get("value")) not in (IpAddressType.ipv4, IpAddressType.dualstack):
+                if (ipAddressType := patch_op.get("value")) not in (
+                    IpAddressType.ipv4,
+                    IpAddressType.dualstack,
+                ):
                     raise BadRequestException("ipAddressType must be either ipv4 or dualstack.")
                 if (
                     rest_api.endpoint_configuration["types"] == [EndpointType.PRIVATE]


localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -359,6 +389,51 @@ def update_rest_api(
                     )
                 patch_op["value"] = value
 
+            elif patch_op_path.startswith("/endpointConfiguration/types"):
+                if patch_op["op"] not in ("add", "remove", "move", "test", "replace", "copy"):
+                    raise CommonServiceException(
+                        code="ValidationException",
+                        message=f"1 validation error detected: Value '{patch_op['op']}' at 'updateRestApiInput.patchOperations.1.member.op' failed to satisfy constraint: Member must satisfy enum value set: [add, remove, move, test, replace, copy]",
+                    )
```
bentsku: nit: I believe this validation is for all types, and could be move outside of the `if patch_op_path` check, and be done very early, almost after the `for patch_op in patch_operations:` line. What do you think?

localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -244,8 +244,37 @@ def test_invoke_method(
 
     @handler("CreateRestApi", expand=False)
     def create_rest_api(self, context: RequestContext, request: CreateRestApiRequest) -> RestApi:
+        endpoint_configuration = request.get("endpointConfiguration", {})
+        types = endpoint_configuration.get("types", ["EDGE"])
+        ip_address_type = endpoint_configuration.get("ipAddressType")
+        if len(types) > 1:
+            raise BadRequestException("Cannot create an api with multiple Endpoint Types.")
```
bentsku: I believe that maybe the length could actually not be different than 1. This could help us simplify the code a little bit.
you'll need to add the test case where `types=[]` in the tests, bit I've checked it and it raises an exception

```python
if not types:
    raise BadRequestException("REGIONAL Configuration and EDGE Configuration cannot be both DISABLED.")
    
elif len(types) > 1:
    raise BadRequestException("Cannot create an api with multiple Endpoint Types.")

endpoint_type = types[0]
# and then you can use `endpoint_type` directly and do not have to do the check against lists, like so

# ...

if endpoint_type not in (EndpointType.PRIVATE, EndpointType.EDGE, EndpointType.REGIONAL):
# etc etc
```

This way, you would have to wrap every value in `[]` and could treat it as single value. What do you think?

localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -244,8 +244,37 @@ def test_invoke_method(
 
     @handler("CreateRestApi", expand=False)
     def create_rest_api(self, context: RequestContext, request: CreateRestApiRequest) -> RestApi:
+        endpoint_configuration = request.get("endpointConfiguration", {})
+        types = endpoint_configuration.get("types", ["EDGE"])
```
bentsku: It seems we have an `StrEnum` we can import from `localstack.aws.api.apigateway`: `EndpointType` which you can access to have those hardcoded string values. 

You can try replacing all hardcoded values from this PR with the `StrEnum` instead. 
```suggestion
        types = endpoint_configuration.get("types", [EndpointType.EDGE])
```

localstack-core/localstack/services/apigateway/legacy/provider.py
ArthurAkh: Ah yes, of course! Perfect, I moved this check just after the `for patch_op in patch_operations:` line. I even considered putting it into a separate loop, to save doing unnecessary patch operations, if it happens that one of the operations is invalid and we end up raising the exception anyways. What do you think?

localstack-core/localstack/services/apigateway/legacy/provider.py
```diff
@@ -244,8 +246,43 @@ def test_invoke_method(
...
+        endpoint_type = types[0]
+
+        error_messages = []
+        if endpoint_type not in (EndpointType.PRIVATE, EndpointType.EDGE, EndpointType.REGIONAL):
+            error_messages.append(
+                f"Value '[{endpoint_type}]' at 'createRestApiInput.endpointConfiguration.types' failed to satisfy constraint: Member must satisfy constraint: [Member must satisfy enum value set: [PRIVATE, EDGE, REGIONAL]]",
+            )
+        elif not ip_address_type:
+            if endpoint_type in (EndpointType.EDGE, EndpointType.REGIONAL):
+                ip_address_type = IpAddressType.ipv4
```
bentsku, line 268: praise: very nice to use the `StrEnum` for the `IpAdressType` as well!

bentsku (approved):
LGTM! This is great, thanks for addressing the comments super fast! This is a great contribution 😄

Regarding the comment for the separate loop, I think this is already good to go! This could be part of another push to improve our `update_*` operations at some point. 

Nice work!

PR #12829, opened 2025-07-03
Add tag push trigger to main workflow

silv-io:
## Motivation

We use an empty commit + a version tag to mark the release of new versions in PyPI and DockerHub. The empty commit triggered the run in CircleCI
Since switching to GitHub Actions for our CI, the empty commit does not do that anymore because of the path specifiers. However, we can use tag pushes instead to trigger the CI + push jobs.

[In a separate repo](https://github.com/silv-io/gha-trigger-testing/actions/runs/16269110767/job/45931870915), I've checked how this would behave:

Having both `tags` and `paths` patterns in a workflow does not interfere with each other. That means that pushed tags will trigger the CI regardless of which files were changed in the commit that the tag points to. 

Furthermore, a tag push event still has the same name of `push` in `github.event_name`, so we don't need to change that part in the pipeline.

The `github.ref` however changes to `refs/tags/v...` which is why we need to check for `startsWith( github.ref, refs/tags/v' )` to determine if a triggered action is on a tagged release.

## Changes

- add a tag push trigger on all version tags (so `v<major>.<minor>.<patch>` where major, minor and hotfix can be arbitrary integers)
- Adapt all workflows that are called by the main workflow or are the main workflow to include checks for tag refs in addition to master branch checks.

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -190,7 +190,7 @@ jobs:
     name: "Push images"
     runs-on: ubuntu-latest
     # push image on master, target branch not set, and the dependent steps were either successful or skipped
-    if: github.ref == 'refs/heads/master' && !failure() && !cancelled() && github.repository == 'localstack/localstack'
+    if: ( github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') ) && !failure() && !cancelled() && github.repository == 'localstack/localstack'
     needs:
       # all tests need to be successful for the image to be pushed
       - test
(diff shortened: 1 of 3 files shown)


alexrashed:
Good point! Tag protection rules are in place since last week, only our release automation (no one else) can push tags in this repository.

silv-io:
> Do you think it would also be possible to adjust our release workflow (which is in our pro repo) such that this empty commit is not even created in the first place anymore?

I was thinking the same thing. I think we can just remove `--allow-empty` from the release helper function. I'll think it through and then if I don't find anything I'll adjust it.

PR #12831, opened 2025-07-04
fix delete method of AWS::SNS::TopicPolicy CFn resource

thrau:
## Motivation

This was a short detour on my main quest to get the CloudTrail scenario test working. The test uses a CDK construct to configure CloudTrail delivery events to SNS, which also creates a topic policy for SNS. I noticed the stacks weren’t deleting properly because the TopicPolicy resource was stuck in a retry loop.

I also raised https://github.com/getmoto/moto/pull/9041 to get rid of the `SNS:Receive` action in the default policy that moto creates incorrectly.

## Changes

* Deleting a topic policy via cloudformation now correctly resets the topic's policy to the default policy

### DIFF
--- a/localstack-core/localstack/services/sns/models.py
+++ b/localstack-core/localstack/services/sns/models.py
@@ -11,6 +11,7 @@ from localstack.aws.api.sns import (
     topicARN,
 )
 from localstack.services.stores import AccountRegionBundle, BaseStore, LocalAttribute
+from localstack.utils.aws.arns import parse_arn
 from localstack.utils.objects import singleton_factory
 from localstack.utils.strings import long_uid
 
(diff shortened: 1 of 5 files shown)


tests/aws/services/cloudformation/resources/test_sns.py
```diff
@@ -152,6 +153,61 @@ def test_sns_topic_with_attributes(infrastructure_setup, aws_client, snapshot):
         snapshot.match("topic-archive-policy", response["Attributes"]["ArchivePolicy"])
 
 
+@markers.aws.validated
+@markers.snapshot.skip_snapshot_verify(
+    paths=[
+        "$..Statement..Action",  # TODO: see https://github.com/getmoto/moto/pull/9041
```
bentsku, line 159: thanks a lot for fixing this 😄 this is gonna remove a few snapshot skips!

localstack-core/localstack/services/sns/resource_providers/aws_sns_topicpolicy.py
```diff
@@ -96,14 +97,17 @@ def delete(
...
+                    AttributeValue=json.dumps(create_default_sns_topic_policy(topic_arn)),
                 )
+
             except ClientError as err:
                 if "NotFound" not in err.response["Error"]["Code"]:
                     raise
 
         return ProgressEvent(
-            status=OperationStatus.IN_PROGRESS,
+            status=OperationStatus.SUCCESS,
```
bentsku, line 110: question: I suppose this is the main fix, right? the Delete operation would never finish as it never returned `SUCCESS`?

localstack-core/localstack/services/sns/models.py
```diff
@@ -25,6 +26,38 @@
 SnsMessageProtocols = Literal[SnsProtocols, SnsApplicationPlatforms]
 
 
+def create_default_sns_topic_policy(topic_arn: str) -> dict:
```
bentsku, line 29: nice addition! I'm always torn with this kind of utils in `models.py`, but it makes sense. It's going to nicely be used when we start creating topics in LocalStack directly 👀

bentsku:
As a side note, it will hopefully be possible in the future to skip list item like the troublesome `SNS:Receive` with this PR: https://github.com/localstack/localstack-snapshot/pull/16

Right now, it's not possible to skip individual list item

localstack-core/localstack/services/sns/resource_providers/aws_sns_topicpolicy.py
thrau: this + the default policy, because sending an empty attribute value to moto raises and error here: https://github.com/getmoto/moto/blob/338cb256a4ab938f24b5b29ec079ed26a9d0c093/moto/sns/models.py#L137-L139

localstack-core/localstack/services/sns/models.py
thrau: happy to put it elsewhere. what would be a good place? should we create a utils module, or should it just go into the provider? i could also move it to the resource provider directly

bentsku: I think it is good for now! If we internalize, we’ll create a util folder and move things around! Sorry for the delay 😄

PR #12832, opened 2025-07-04
Update action names for AWS prefix

alexrashed:
## Motivation
Super small change aligning some of our GitHub action names and their filenames.
This helps grouping / structuring the workflows we have (since we already have quite a few).

## Changes
- Rename OpenAPI workflow introduced with #11709 to use `-` as separator in the filename
- Rename `tests-s3-image.yml` to `aws-tests-s3-image.yml`
- Rename "AWS Integration Tests" and "S3 Image Integration Tests" to prefix them properly with "AWS /".


k-a-il (approved):
LGTM 🚀

PR #12848, opened 2025-07-08
Docker: Add 'log_handler'-argument to pull()

bblommers:
## Motivation
Some of the images that we're downloading to support our emulation are quite large, so I've been considering a way to show more info to the user about what's happening in the background.

This PR adds an additional argument, `log_handler` to the `docker_client.pull(..)` method, allowing service developers to inspect (or log) the exact output of the `pull`-command.

### Example invocation
For example, to stream the logs to user (probably the most common usecase for this):
```python
def some_action():
    ...
    def _process(line: str):
        LOG.debug(line)

    docker_client.pull_image("alpine", log_handler=_process)
```


dfangl (approved):
I think this is a reasonable change, LGTM! Thanks for adding a test right away!

PR #12851, opened 2025-07-10
LDM as provisioner of debug-enabled containers

MEPalma:
## Motivation
The current LDM implementation depends heavily on the Lambda service’s container provisioning logic, injecting LDM-specific parameter overrides throughout various parts of the service to support debug-enabled containers. This fragmented integration makes LDM difficult to maintain and limits control over the lifecycle of debug-enabled containers. This refactor centralizes LDM responsibilities by transforming it into a dedicated provisioner for debug-enabled containers. As a result, the Lambda service is relieved of most LDM-related logic. Additionally, these changes ensure that debug-enabled containers persist across multiple invocations, allowing the debug client to remain connected or reconnect after disconnections. 

## Changes
- LDM as a debug-enabled container provisioner
- removed LDM fragments in the lambda provider
- introduce new ExecutionEnvironment subtype for container-enabled execution environments 
- avoid container restarts between debug invokes, for compatibility python lambda functions must include a revised debug server logic:

```python
def handler(event, context):
    print(event)
    return event

def wait_for_debug_client(port: int=19891, timeout: int=3600):
    import time, threading
    import sys, glob
    sys.path.append(glob.glob(".venv/lib/python*/site-packages")[0])
    import debugpy

    if not hasattr(wait_for_debug_client, "_debugpy_listening"):
        wait_for_debug_client._debugpy_listening = False

    if not wait_for_debug_client._debugpy_listening:
        try:
            debugpy.listen(("0.0.0.0", port))
            wait_for_debug_client._debugpy_listening = True
            print(f"debugpy is now listening on port {port}")
        except RuntimeError as e:
            print(f"debugpy.listen() failed or already active: {e}")

    if not debugpy.is_client_connected():
        print("Waiting for client to attach debugger...")

        def cancel_wait():
            time.sleep(timeout)
            print("Canceling debug wait task after timeout...")
            debugpy.wait_for_client.cancel()

        threading.Thread(target=cancel_wait, daemon=True).start()
        debugpy.wait_for_client()
    else:
        print("Debugger already attached.")

wait_for_debug_client()

```
- several other related changes

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/version_manager.py
+++ b/localstack-core/localstack/services/lambda_/invocation/version_manager.py
@@ -192,6 +192,8 @@ class LambdaVersionManager:
             LOG.warning(message)
             raise ServiceException(message)
 
+        # If the environment has debugging enabled, route the invocation there;
+        # debug environments bypass Lambda service quotas.
         if debug_execution_environment := LDM.get_execution_environment(
             qualified_lambda_arn=self.function_version.qualified_arn,
             user_agent=invocation.user_agent,
--- a/localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
+++ b/localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
@@ -27,9 +27,9 @@ from localstack.utils.strings import long_uid
 
 LOG = logging.getLogger(__name__)
 
-# Specifies the fault timeout value in seconds to be used by time restricted workflows when
-# Debug Mode is enabled. The value is set to one hour to ensure eventual termination of
-# long-running processes.
+# Specifies the default timeout value in seconds to be used by time restricted workflows
+# when Debug Mode is enabled. The value is set to one hour to ensure eventual termination
+# of long-running processes.
 DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS: int = 3_600
 LDM_ENV_VAR_DEBUG_PORT: str = "LDM_DEBUG_PORT"
 
@@ -124,6 +124,9 @@ class LambdaDebugTarget:
                 lambda_function_debug_config=self.lambda_function_debug_config,
                 on_timeout=self._on_execution_environment_timeout,
             )
+            # FIXME: this log should take place after RuntimeStatus.READY, however the debug-enabled
+            #        docker container will not notify LS about it starting up until the user has
+            #        connected a debug client. Future work should resolve this notification issue.
             LOG.info(
                 "LDM is ready for debugger connections for '%s' on port %i.",
                 self.lambda_function_debug_config.qualified_lambda_arn,
@@ -164,7 +167,10 @@ class LambdaDebugTarget:
     def _on_execution_environment_timeout(
         self, version_manager_id: str, environment_id: str
     ) -> None:
-        # TODO: add support
+        # This function is run by the ExecutionEnvironment when the
+        # release of on-demand container times-out whilst waiting for
+        # invokes. However, DebugEnabledExecutionEnvironment are
+        # provisioned-concurrency ExecutionEnvironments.
         LOG.warning(
             "Lambda Debug Mode function '%s' timed out ('%s', '%s')",
             self._lambda_qualified_arn,


localstack-core/localstack/services/lambda_/invocation/assignment.py
```diff
@@ -79,10 +75,7 @@ def get_environment(
 
         try:
             yield execution_environment
-            if is_lambda_debug_timeout_enabled_for(lambda_arn=function_version.qualified_arn):
-                self.stop_environment(execution_environment)
```
joe4dev, line 83: praise: Nice simplification and quality of life improvement to not stop the LDM-enabled container anymore 👍

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
```diff
@@ -139,7 +135,7 @@ def get_environment_variables(self) -> Dict[str, str]:
             # AWS_LAMBDA_DOTNET_PREJIT
             "TZ": ":UTC",
             # 2) Public AWS RIE interface: https://github.com/aws/aws-lambda-runtime-interface-emulator
-            "AWS_LAMBDA_FUNCTION_TIMEOUT": self._get_execution_timeout_seconds(),
```
joe4dev, line 142: question: Is a debug-enabled Lambda function killed pre-maturely by the Golang code if we don't extend this timeout anymore?

Background: `AWS_LAMBDA_FUNCTION_TIMEOUT` is used by the Golang Lambda RIE [here](https://github.com/localstack/lambda-runtime-init/blob/9be92e099e99c9973f2e1730d1af6d4ffd769bb2/cmd/localstack/main.go#L221).

localstack-core/localstack/services/lambda_/invocation/version_manager.py
```diff
@@ -191,6 +192,28 @@ def invoke(self, *, invocation: Invocation) -> InvocationResult:
             LOG.warning(message)
             raise ServiceException(message)
 
+        if debug_execution_environment := LDM.get_execution_environment(
```
joe4dev, line 197: docs(nit): It might be worth adding some docstring here, clarifying that debug-enabled execution environments don't consider Lambda quotas

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
```diff
@@ -0,0 +1,369 @@
...
+)
+from localstack.services.lambda_.lambda_debug_mode.ldm_config_file import parse_ldm_config
+from localstack.services.lambda_.provider_utils import get_function_version
+from localstack.utils.aws.arns import parse_arn
+from localstack.utils.objects import singleton_factory
+from localstack.utils.strings import long_uid
+
+LOG = logging.getLogger(__name__)
+
+# Specifies the fault timeout value in seconds to be used by time restricted workflows when
```
joe4dev: typo (nit): fault -> default

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
joe4dev: I see that LDM overwrites this later [here](https://github.com/localstack/localstack/pull/12851/files#diff-af99e5d85f77b8cbd24995f21c3794d42e5bdf09804e9ef4812bfbfbe4e3c5bfR77).

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
```diff
@@ -0,0 +1,369 @@
...
+        self._lambda_qualified_arn = lambda_function_debug_config.qualified_lambda_arn
+        self.lambda_function_debug_config = lambda_function_debug_config
+        self._debug_execution_environment = None
+
+    def start_debug_enabled_execution_environment(self):
+        # Attempt to create the debug environment now if the function exists.
+        with self._mutex:
+            if self._debug_execution_environment is not None:
+                return
+            self.stop_debug_enabled_execution_environment()
```
joe4dev, line 103: question: Why are we stopping the environment here? Can this happen given the `return` above if there is an existing environment?

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
```diff
@@ -0,0 +1,369 @@
...
+            except ResourceNotFoundException:
+                # The lambda function has not being created yet.
+                return
+
+            self._debug_execution_environment = DebugEnabledExecutionEnvironment(
+                function_version=function_version,
+                lambda_function_debug_config=self.lambda_function_debug_config,
+                on_timeout=self._on_execution_environment_timeout,
+            )
+            LOG.info(
```
joe4dev, line 130: question: Shouldn't that info log happen after starting the environment? Like `status == RuntimeStatus.READY`?

localstack-core/localstack/services/lambda_/invocation/lambda_models.py
```diff
@@ -66,6 +66,7 @@ class Invocation:
     # = invocation_id
     request_id: str
     trace_context: dict
+    user_agent: Optional[str] = None
```
joe4dev, line 69: FYI: The `Invocation` is used in the event manager for async invokes. The optional new field should be ok, given it initializes with a default

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm_config_file.py
```diff
@@ -0,0 +1,178 @@
+from __future__ import annotations
```
joe4dev, line 1: FYI: refactoring where code is moved from https://github.com/localstack/localstack/pull/12851/files#diff-2266754b538bc83024fa9e051519928adf58f50124f471c186512e25731e8736L1

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
```diff
@@ -0,0 +1,369 @@
...
+    def stop_debug_enabled_execution_environment(self):
+        with self._mutex:
+            if environment := self._debug_execution_environment:
+                environment.stop()
+            self._debug_execution_environment = None
+
+    def get_execution_environment(self) -> DebugEnabledExecutionEnvironment:
+        # TODO: add support for concurrent invokes, such as invoke object queuing, new container spinup
+        with self._mutex:
+            # TODO: move this start-up logic to lambda function creation.
```
joe4dev, line 151: question (regression): Does this limitation introduce a regression such that LDM is not applied to Lambda functions created or updated AFTER adding an LDM configuration?

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
```diff
@@ -0,0 +1,369 @@
...
+                raise TooManyRequestsException(
+                    "Rate Exceeded.",
+                    Reason="SingleLeaseEnforcement",
+                    Type="User",
+                )
+
+    def _on_execution_environment_timeout(
+        self, version_manager_id: str, environment_id: str
+    ) -> None:
+        # TODO: add support
```
joe4dev: question: What does "add support" include (e.g., raising aws-validated timeout error)?
Does this happen if `DEFAULT_LAMBDA_DEBUG_MODE_TIMEOUT_SECONDS` (i.e., 1h) is exceeded or `enforce-timeout` is disabled and then running into a short function timeout?

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
joe4dev: Testing showed that this is not the case.

Brainstorming the list of change points potentially to consider (in the future):
* CreateFunction: Marco pointed out that function creation is handled [here](https://github.com/localstack/localstack/pull/12851/files#diff-d7cfc1abddc47b8f3fdbfd708f278960ac057febe9d4eb469aab1b373ad80441R1072)
* UpdateFunction: doesn't change ARN, should be fine
* UpdateFunctionConfiguration: doesn't change ARN, should be fine
* [PublishVersion](https://docs.aws.amazon.com/lambda/latest/api/API_PublishVersion.html): creates a new function version, which might require spawning a debug-enabled container if the new function version ARN matches a debug config
* [CreateAlias](https://docs.aws.amazon.com/lambda/latest/api/API_CreateAlias.html): creates a new function ARN via the alias `Name`, which could match a debug config

localstack-core/localstack/services/lambda_/invocation/execution_environment.py
MEPalma: Exactly, like you pointed out, this value continues to be refined by the LDM through the debug-enabled execution environment

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
MEPalma: thank you, yes this is indeed redundant, I removed this statement

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
MEPalma: In principle yes, I would like this Log to occur after the container is READY. However, due to the debug layer, the docker container will not notify LS until the user has connected the debug client. I think this is something we should refine soon: I added a todo.

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
MEPalma: Thank you for this list, it's very helpful. I'll refer to this to refine the revision of this logic we already implemented in  follow-up changes to these.

localstack-core/localstack/services/lambda_/lambda_debug_mode/ldm.py
MEPalma: This functionality should not be applicable for debug-enabled containers. It is used in the release of on-demand containers, however these are `provisioned-concurrency`. I revised the comment to be more informative on the situation.

PR #12853, opened 2025-07-11
KMS: Add list-key-rotations flag

demaj:
## Motivation

Adding support for [ListKeyRotations](https://docs.aws.amazon.com/kms/latest/developerguide/list-rotations.html). Mentioned in #12342.

## Changes

Implementation of [list-key-rotations](https://docs.aws.amazon.com/kms/latest/developerguide/list-rotations.html) feature for keys that may have multiple key materials associated with them.

As per AWS's, list-key-rotation is only allowed for:
- symmetric keys
- Imported keys from single-region-only

## TODO

What's left to do:

- [ ] Add support for imported EXTERNAL origin single-region keys

### DIFF
--- a/localstack-core/localstack/services/kms/models.py
+++ b/localstack-core/localstack/services/kms/models.py
@@ -277,6 +277,16 @@ class KmsCryptoKey:
         )
 
 
+@dataclass
+class KeyRotationEntry:
+    key_id: str
+    key_material: bytes
+    key_material_id: str
+    key_material_state: str
+    rotation_date: datetime.datetime
+    rotation_type: str | None = None
+
+
 class KmsKey:
     metadata: KeyMetadata
     crypto_key: KmsCryptoKey
@@ -285,7 +295,7 @@ class KmsKey:
     is_key_rotation_enabled: bool
     rotation_period_in_days: int
     next_rotation_date: datetime.datetime
-    previous_keys = [str]
+    key_rotations: list[KeyRotationEntry]
 
     def __init__(
         self,
@@ -294,7 +304,7 @@ class KmsKey:
         region: str = None,
     ):
         create_key_request = create_key_request or CreateKeyRequest()
-        self.previous_keys = []
+        self.key_rotations = []
 
         # Please keep in mind that tags of a key could be present in the request, they are not a part of metadata. At
         # least in the sense of DescribeKey not returning them with the rest of the metadata. Instead, tags are more
@@ -322,6 +332,16 @@ class KmsKey:
         self.crypto_key = KmsCryptoKey(self.metadata.get("KeySpec"), custom_key_material)
         self.rotation_period_in_days = 365
         self.next_rotation_date = None
+
+        if self._supports_rotation():
+            initial_rotation = KeyRotationEntry(
+                key_id=self.metadata["Arn"],
+                rotation_date=datetime.datetime.now(),
+                key_material_state="CURRENT",
+                key_material_id=long_uid(),
+                key_material=bytes(self.crypto_key.key_material),
+            )
+            self.key_rotations.append(initial_rotation)
 
     def calculate_and_set_arn(self, account_id, region):
         self.metadata["Arn"] = kms_key_arn(self.metadata.get("KeyId"), account_id, region)
@@ -357,7 +377,11 @@ class KmsKey:
         self, ciphertext: Ciphertext, encryption_context: EncryptionContextType = None
     ) -> bytes:
         aad = _serialize_encryption_context(encryption_context=encryption_context)
-        keys_to_try = [self.crypto_key.key_material] + self.previous_keys
+        keys_to_try = [self.crypto_key.key_material]
+
+        for rotation in sorted(self.key_rotations, key=lambda r: r.rotation_date, reverse=True):
+            if rotation.key_material not in keys_to_try:
+                keys_to_try.append(rotation.key_material)
 
         for key in keys_to_try:
             try:
@@ -738,13 +762,39 @@ class KmsKey:
             return request_key_usage or "ENCRYPT_DECRYPT"
 
     def rotate_key_on_demand(self):
-        if len(self.previous_keys) >= ON_DEMAND_ROTATION_LIMIT:
+        on_demand_count = sum(
+            1 for rotation in self.key_rotations if rotation.rotation_type == "ON_DEMAND"
+        )
+
+        if on_demand_count >= ON_DEMAND_ROTATION_LIMIT:
             raise LimitExceededException(
                 f"The on-demand rotations limit has been reached for the given keyId. "
                 f"No more on-demand rotations can be performed for this key: {self.metadata['Arn']}"
             )
-        self.previous_keys.append(self.crypto_key.key_material)
-        self.crypto_key = KmsCryptoKey(KeySpec.SYMMETRIC_DEFAULT)
+
+        for rotation in self.key_rotations:
+            rotation.key_material_state = "NON_CURRENT"
+
+        new_key_material = os.urandom(SYMMETRIC_DEFAULT_MATERIAL_LENGTH)
+
+        rotation_entry = KeyRotationEntry(
+            key_id=self.metadata["Arn"],
+            rotation_date=datetime.datetime.now(),
+            rotation_type="ON_DEMAND",
+            key_material_state="CURRENT",
+            key_material_id=long_uid(),
+            key_material=new_key_material,
+        )
+        self.key_rotations.append(rotation_entry)
+
+        self.crypto_key = KmsCryptoKey(KeySpec.SYMMETRIC_DEFAULT, new_key_material)
+
+    def _supports_rotation(self) -> bool:
+        return (
+            self.metadata.get("KeySpec") == KeySpec.SYMMETRIC_DEFAULT
+            and self.metadata.get("Origin") == OriginType.AWS_KMS
+            and self.metadata.get("KeyUsage") == KeyUsageType.ENCRYPT_DECRYPT
+        )
 
 
 class KmsGrant:
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -60,6 +60,7 @@ from localstack.aws.api.kms import (
     GrantTokenType,
     ImportKeyMaterialResponse,
     ImportType,
+    IncludeKeyMaterial,
     IncorrectKeyException,
     InvalidCiphertextException,
     InvalidGrantIdException,
@@ -78,6 +79,8 @@ from localstack.aws.api.kms import (
     ListGrantsResponse,
     ListKeyPoliciesRequest,
     ListKeyPoliciesResponse,
+    ListKeyRotationsRequest,
+    ListKeyRotationsResponse,
     ListKeysRequest,
     ListKeysResponse,
     ListResourceTagsRequest,
@@ -1354,6 +1357,62 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
             key.rotation_period_in_days = request.get("RotationPeriodInDays")
         key._update_key_rotation_date()
 
+    @handler("ListKeyRotations", expand=False)
+    def list_key_rotations(
+        self, context: RequestContext, request: ListKeyRotationsRequest
+    ) -> ListKeyRotationsResponse:
+        # https://docs.aws.amazon.com/kms/latest/developerguide/key-state.html
+        # "If the KMS key has imported key material or is in a custom key store: UnsupportedOperationException."
+        # We do not model that here, though.
+        key = self._get_kms_key(
+            context.account_id,
+            context.region,
+            request.get("KeyId"),
+            disabled_key_allowed=True,
+            pending_deletion_key_allowed=True,
+        )
+
+        self._validate_key_supports_rotation(key)
+
+        include_key_material = request.get("IncludeKeyMaterial")
+        if include_key_material:
+            if include_key_material not in ["ALL_KEY_MATERIAL", "ROTATIONS_ONLY"]:
+                raise ValidationException(
+                    f"1 validation error detected: Value '{include_key_material}' at 'includeKeyMaterial' "
+                    f"failed to satisfy constraint: Member must satisfy enum value set: [ALL_KEY_MATERIAL, ROTATIONS_ONLY]"
+                )
+
+        rotation_history = []
+        for rotation in key.key_rotations:
+            rotation_entry = {
+                "KeyId": rotation.key_id,
+                "KeyMaterialId": rotation.key_material_id,
+                "KeyMaterialState": rotation.key_material_state,
+                "RotationDate": rotation.rotation_date,
+                "RotationType": rotation.rotation_type,
+            }
+
+            if include_key_material == IncludeKeyMaterial.ALL_KEY_MATERIAL:
+                rotation_history.append(rotation_entry)
+            else:
+                if rotation.rotation_type in ["AUTOMATIC", "ON_DEMAND"]:
+                    rotation_history.append(rotation_entry)
+
+        rotation_history = sorted(
+            rotation_history,
+            key=lambda x: x.get("RotationDate", datetime.datetime.min),
+            reverse=True,
+        )
+
+        rotations_list = PaginatedList(rotation_history)
+        page, next_token = rotations_list.get_page(
+            lambda rotation: rotation.get("RotationDate"),
+            next_token=request.get("Marker"),
+            page_size=request.get("Limit", 100),
+        )
+        kwargs = {"NextMarker": next_token, "Truncated": True} if next_token else {}
+        return ListKeyRotationsResponse(Rotations=page, **kwargs)
+
     @handler("ListKeyPolicies", expand=False)
     def list_key_policies(
         self, context: RequestContext, request: ListKeyPoliciesRequest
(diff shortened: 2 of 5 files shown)


demaj:

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+
+        include_key_material = request.get("IncludeKeyMaterial")
+        if include_key_material is not None:
+            if include_key_material not in ["ALL_KEY_MATERIAL", "ROTATIONS_ONLY"]:
+                raise ValidationException(
+                    f"1 validation error detected: Value '{include_key_material}' at 'includeKeyMaterial' "
+                    f"failed to satisfy constraint: Member must satisfy enum value set: [ALL_KEY_MATERIAL, ROTATIONS_ONLY]"
+                )
+
+            if key.metadata["KeySpec"] != KeySpec.SYMMETRIC_DEFAULT:
```
demaj, line 1384: As per AWS docs, for asymmetric keys when `IncludeKeyMaterial` is not provided we simply return the `Rotations` as empty `[]`.
See here: https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-compare.html#key-type-table

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+                    rotation_entry["ExpirationModel"] = rotation.expiration_model
+                if rotation.import_state is not None:
+                    rotation_entry["ImportState"] = rotation.import_state
+                if rotation.valid_to is not None:
+                    rotation_entry["ValidTo"] = rotation.valid_to
+
+                if include_key_material == IncludeKeyMaterial.ALL_KEY_MATERIAL:
+                    rotation_history.append(rotation_entry)
+                else:  # Default ROTATIONS_ONLY
+                    if rotation.rotation_type in ["AUTOMATIC", "ON_DEMAND"]:
```
demaj, line 1416: A KMS key on its creation doesn't have any rotation_type.

localstack-core/localstack/services/kms/models.py
```diff
@@ -323,6 +337,15 @@ def __init__(
         self.rotation_period_in_days = 365
         self.next_rotation_date = None
 
+        if self.supports_rotation():
+            initial_rotation = KeyRotationEntry(
+                key_id=self.metadata["Arn"],
+                key_material_state="CURRENT",
+                key_material_id=long_uid(),  # FIXME: a more appropriate KMS output
```
demaj, line 344: Currently relying on uuid, open to suggestions.

tests/aws/services/kms/test_kms.snapshot.json
demaj, line 1: I'm getting `"Truncated": false` here. Not entirely sure why it's being included, I do not see it on my awscli output; there is a discrepancy between AWS Docs and AWS CLI.
Removed them manually from this file (I know we're not supposed to manually edit it). Let me know if I should re-add it and make changes on the function's response payload.

localstack-core/localstack/services/kms/models.py
```diff
@@ -277,6 +277,20 @@ def key(self) -> RSAPrivateKey | EllipticCurvePrivateKey:
...
+    key_id: str
+    key_material: bytes
+    key_material_id: str
+    key_material_state: str
+    rotation_date: datetime.datetime | None = None
+    rotation_type: str | None = None
+    key_material_description: str | None = None
+    expiration_model: str | None = None
+    import_state: str | None = None
+    valid_to: str | None = None
```
sannya-singal, line 291: Did you consider using the predefined class in `__init__.py` file which already includes this type `RotationsListEntry`?

localstack-core/localstack/services/kms/models.py
```diff
@@ -323,6 +337,15 @@ def __init__(
         self.rotation_period_in_days = 365
         self.next_rotation_date = None
 
+        if self.supports_rotation():
+            initial_rotation = KeyRotationEntry(
+                key_id=self.metadata["Arn"],
+                key_material_state="CURRENT",
```
sannya-singal, line 343: Instead of using hardcoded values, import the class defined in `__init__.py` file which already includes the types you have defined here. 
The file is automatically generated, you can find more details here: https://github.com/localstack/localstack/blob/main/docs/localstack-concepts/README.md

localstack-core/localstack/services/kms/models.py
sannya-singal: `key_material_id` should follow the constraints as mentioned in AWS documentation here: https://docs.aws.amazon.com/kms/latest/APIReference/API_RotationsListEntry.html#KMS-Type-RotationsListEntry-KeyMaterialId. 

Hint: try playing around generating random bytes from `os`, you could find several references for it in the codebase 😉 and then try to convert it to a 64-character hexadecimal string.

localstack-core/localstack/services/kms/models.py
```diff
@@ -738,13 +765,39 @@ def _get_key_usage(self, request_key_usage: str, key_spec: str) -> str:
             return request_key_usage or "ENCRYPT_DECRYPT"
 
     def rotate_key_on_demand(self):
-        if len(self.previous_keys) >= ON_DEMAND_ROTATION_LIMIT:
+        on_demand_count = sum(
+            1 for rotation in self.key_rotations if rotation.rotation_type == "ON_DEMAND"
+        )
+
+        if on_demand_count >= ON_DEMAND_ROTATION_LIMIT:
```
sannya-singal, line 772: What happens if the rotation type is `AUTOMATIC` here?

localstack-core/localstack/services/kms/provider.py
sannya-singal: question: From the test `test_list_key_rotations_unsupported_key_types` seems like it always raises an error for asymmetric keys, could you explain this a bit more here why it should be `[]`?

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+            context.account_id,
+            context.region,
+            request.get("KeyId"),
+            disabled_key_allowed=True,
+            pending_deletion_key_allowed=True,
+        )
+
+        include_key_material = request.get("IncludeKeyMaterial")
+        if include_key_material is not None:
+            if include_key_material not in ["ALL_KEY_MATERIAL", "ROTATIONS_ONLY"]:
```
sannya-singal, line 1378: `ALL_KEY_MATERIAL` and `ROTATIONS_ONLY` should be used from the `IncludeKeyMaterial` class in `__init__.py` file which already includes the types you've defined here. 

(Note: This is applicable in other changes made to this PR, I am not commenting in every place. Let me know if there is any confusion around this, happy to help and explain more 🙂)

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+
+            if key.metadata["KeySpec"] != KeySpec.SYMMETRIC_DEFAULT:
+                raise ValidationException(
+                    "Input parameter includeKeyMaterial can only be used with symmetric keys."
+                )
+
+            self._validate_key_supports_include_key_material(key)
+
+        rotation_history = []
+        if key.supports_rotation() or include_key_material is not None:
```
sannya-singal, line 1392: Why do we validate similar checks twice here: `_validate_key_supports_include_key_material()` and `supports_rotation()`?

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+                if rotation.rotation_date is not None:
+                    rotation_entry["RotationDate"] = rotation.rotation_date
+                if rotation.rotation_type is not None:
+                    rotation_entry["RotationType"] = rotation.rotation_type
+                if rotation.expiration_model is not None:
+                    rotation_entry["ExpirationModel"] = rotation.expiration_model
+                if rotation.import_state is not None:
+                    rotation_entry["ImportState"] = rotation.import_state
+                if rotation.valid_to is not None:
+                    rotation_entry["ValidTo"] = rotation.valid_to
```
sannya-singal, line 1411: Instead of so many conditional statements can we simply add the values to a var of type `RotationsListEntry` and then append the var to the list which needs to be returned?

So something like 
```
RotationsListEntry(
                KeyId=key_id,
                ValidTo=...,
            )
```

localstack-core/localstack/services/kms/provider.py
sannya-singal: I suggest its worth writing a test case to validate the behavior in case of automatically rotated keys, what do you think?

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1354,6 +1357,74 @@ def enable_key_rotation(
...
+                    if rotation.rotation_type in ["AUTOMATIC", "ON_DEMAND"]:
+                        rotation_history.append(rotation_entry)
+
+        rotations_list = PaginatedList(rotation_history)
+        page, next_token = rotations_list.get_page(
+            lambda rotation: rotation.get("RotationDate"),
+            next_token=request.get("Marker"),
+            page_size=request.get("Limit", 100),
+        )
+        kwargs = {"NextMarker": next_token, "Truncated": True} if next_token else {}
```
sannya-singal, line 1425: I recommend here either we remove pagination for this PR and leave a TODO or add a test to validate this implementation.

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1605,6 +1676,20 @@ def _validate_plaintext_key_type_based(
                 f"Algorithm {encryption_algorithm} and key spec {key.metadata['KeySpec']} cannot encrypt data larger than {max_size_bytes} bytes."
             )
 
+    def _validate_key_supports_include_key_material(self, key: KmsKey):
+        if key.metadata["Origin"] != OriginType.AWS_KMS:
+            raise UnsupportedOperationException(
+                f"{key.metadata['Arn']} origin is {key.metadata.get('Origin')} which is not valid for this operation."
+            )
```
sannya-singal, line 1683: What happens if the KMS key is imported here, i.e. `Origin` is `EXTERNAL`? 
It should still allow to list key rotations: but rather we can throw a 5xx instead of 4xx and say that its not yet implemented in localstack.

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
             aws_client.kms.rotate_key_on_demand(KeyId=key_id)
```
sannya-singal, line 1465: Awesome set of tests 🚀

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
         snapshot.match("error-response", e.value.response)
 
+    @markers.aws.validated
+    def test_list_key_rotations_after_on_demand_rotation(
+        self, kms_create_key, aws_client, snapshot
+    ):
+        key_id = kms_create_key(KeyUsage="ENCRYPT_DECRYPT", KeySpec="SYMMETRIC_DEFAULT")["KeyId"]
```
sannya-singal, line 1472: nit: simply use `kms_create_key()["KeyId"]` as by default an `ENCRYPT_DECRYPT` with `SYMMETRIC_DEFAULT` gets created 🙂

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+        )
+
+        response_initial_rotation = aws_client.kms.list_key_rotations(KeyId=key_id)
+        snapshot.match("initial-rotations", response_initial_rotation)
+
+        aws_client.kms.rotate_key_on_demand(KeyId=key_id)
+
+        def _rotation_completed():
+            status = aws_client.kms.get_key_rotation_status(KeyId=key_id)
+            return "OnDemandRotationStartDate" not in status
```
sannya-singal, line 1485: Given that this utility is used across multiple tests, you should move it to a common helper function for reuse.

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+        snapshot.match("after-on-demand-rotation", response_after_on_demand_rotation)
+
+        assert (
+            len(response_after_on_demand_rotation["Rotations"])
+            == len(response_initial_rotation["Rotations"]) + 1
+        )
+
+    @markers.aws.validated
+    def test_list_key_rotations_different_key_states(self, kms_create_key, aws_client, snapshot):
+        key_id = kms_create_key(KeyUsage="ENCRYPT_DECRYPT", KeySpec="SYMMETRIC_DEFAULT")["KeyId"]
```
sannya-singal, line 1499: Since `schedule_key_deletion` is used explicitly in this test, you should rather not use a fixture here since it also does a cleanup later and use `aws_client.kms.create_key`

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+        for rotation in rotations:
+            assert "KeyId" in rotation
+            assert (
+                "RotationDate" in rotation or rotation.get("RotationDate") is None
+            )  # None in initial
+            assert (
+                "RotationType" in rotation or rotation.get("RotationType") is None
+            )  # None in initial
+            assert "KeyMaterialId" in rotation
+            assert "KeyMaterialState" in rotation
```
sannya-singal, line 1556: The test suggests that it would validate the `list_key_rotations` content and it is doing so in snapshot `multiple-rotations`, why do we explicitly need the assert here then since we capture the exact same thing already or is there something i am missing here?

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+
+    @markers.aws.validated
+    @pytest.mark.parametrize(
+        "key_spec,key_usage,expected_error",
+        [
+            ("RSA_2048", "SIGN_VERIFY", "ValidationException"),
+            ("RSA_4096", "ENCRYPT_DECRYPT", "ValidationException"),
+            ("ECC_NIST_P256", "SIGN_VERIFY", "ValidationException"),
+            ("HMAC_256", "GENERATE_VERIFY_MAC", "ValidationException"),
```
sannya-singal, line 1565: Can you add an id to these parameterised tests to make it more readable?

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+        error_response = e.value.response
+        assert error_response["Error"]["Code"] == expected_error
+        snapshot.match(f"{key_spec.lower()}-{key_usage.lower()}-error", error_response)
+
+    @markers.aws.validated
+    def test_list_key_rotations_include_key_material(self, kms_create_key, aws_client, snapshot):
+        key_id = kms_create_key(KeyUsage="ENCRYPT_DECRYPT", KeySpec="SYMMETRIC_DEFAULT")["KeyId"]
+
+        snapshot.add_transformer(
+            snapshot.transform.key_value("KeyMaterialId", reference_replacement=False)
```
sannya-singal, line 1586: Instead of adding it in every test, we should move the transformer to [transformer_utility.py](https://github.com/localstack/localstack/blob/main/localstack-core/localstack/testing/snapshots/transformer_utility.py).

tests/aws/services/kms/test_kms.py
```diff
@@ -1465,6 +1465,159 @@ def test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material(
...
+        snapshot.match("rotations-only", response_rotations_only)
+
+        assert len(response_rotations_only["Rotations"]) < len(response_all_material["Rotations"])
+        assert all(
+            r["RotationType"] in ["AUTOMATIC", "ON_DEMAND"]
+            for r in response_rotations_only["Rotations"]
+        )
+
+        with pytest.raises(ClientError) as e:
+            aws_client.kms.list_key_rotations(KeyId=key_id, IncludeKeyMaterial="INVALID_VALUE")
```
sannya-singal, line 1618: Since this usecase doesn't really belong to the intent of the test named `test_list_key_rotations_include_key_material`, we should write a separate test for this part.

tests/aws/services/kms/test_kms.snapshot.json
sannya-singal: Rightly said 👍 I wouldn't manually make any changes in snapshot. 

What i suggest here is to simple skip verification of `Truncated` in the test and add a TODO in there explaining the same discrepancy.

tests/aws/services/kms/test_kms.snapshot.json
```diff
@@ -2286,5 +2286,256 @@
...
+    "recorded-content": {}
+  },
+  "tests/aws/services/kms/test_kms.py::TestKMS::test_list_key_rotations_unsupported_key_types[ECC_NIST_P256-SIGN_VERIFY-UnsupportedOperationException]": {
+    "recorded-date": "16-07-2025, 15:59:31",
+    "recorded-content": {}
+  },
+  "tests/aws/services/kms/test_kms.py::TestKMS::test_list_key_rotations_unsupported_key_types[HMAC_256-GENERATE_VERIFY_MAC-UnsupportedOperationException]": {
+    "recorded-date": "16-07-2025, 15:59:32",
+    "recorded-content": {}
+  },
```
sannya-singal, line 2376: Are these the leftovers? Because i cant see related tests for `UnsupportedOperationException` specifically?

localstack-core/localstack/services/kms/provider.py
demaj: Based on the AWS Docs, and the behavior of the aws-cli, we return `[]` for unsupported KMS keys. But, whenever the attribute `IncludeKeyMaterial` is provided, it raises an error. See here: https://docs.aws.amazon.com/kms/latest/developerguide/key-state.html#key-state-table
In the `test_list_key_rotations_unsupported_key_types` we provide the `IncludeKeyMaterial` parameter.

localstack-core/localstack/services/kms/provider.py
demaj: I agree, it makes sense to add one. I'll do it. Thanks!

localstack-core/localstack/services/kms/models.py
demaj: I don't know how I missed it. That is a valid point, I will refactor and use the classes from `__init__.py`
Thank you!

sannya-singal:
Hi @demaj! Thanks again for your contribution. Would you like to continue working on this PR? If you need any help to move it forward, please let me know 🙂

demaj:
Hi @sannya-singal, sincere apologies, I was not able to actively work on this. I will try to find some time during this week, is it okay with you? I would love to wrap this PR up, and have my first contribution to Localstack! Thank you! 😊

sannya-singal:
No worries at all 🙂 Glad want to wrap it up, please go ahead and work on it 🎉 Excited for your first contribution! 🚀

alexrashed:
Thanks, @demaj for your contribution. It doesn't look like this is still being worked on, so I'll close this PR for now.

Issue #12854, opened 2025-07-11
bug: setting `DISABLE_CUSTOM_CORS_S3=1` decodes S3 objects with `Content-Encoding: gzip`


davidjb:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When the `DISABLE_CUSTOM_CORS_S3=1` env variable is set, this affects LocalStack's handling of S3 `PutObject` calls with a `Content-Encoding: gzip`. Attempting to put an object like this (see commands below) results in a checksum error:

> An error occurred (BadDigest) when calling the PutObject operation: The CRC64NVME you specified did not match the calculated checksum.

The reason for this appears to be the object has seemingly been decoded ahead of being stored in LocalStack. The reason this appears to be the case is that when disabling payload verification via an SDK and putting the object, the object _is_ stored in LocalStack successfully but like so from `HeadObject` response:

```
AcceptRanges: bytes
ContentEncoding: identity
ContentLength: 6
ContentType: text/plain
ETag: '"605f1ecb0bf8fc6fdef2a4e4983e7478"'
LastModified: '2025-07-11T00:19:01+00:00'
Metadata: {}
ServerSideEncryption: AES256
```

The original text file is `hello` (6 bytes) and when gzipped is 35 bytes. Running `GetObject` on this object returns the decoded text and not the gzipped content which was sent and should have been stored.

The `ContentEncoding: identity` is non-standard behaviour - production AWS (and LocalStack itself when `DISABLE_CUSTOM_CORS_S3=0`) stores this with `ContentEncoding: gzip` and the object remains encoded when stored/retrieved.

This doesn't appear to affect either `br` or `deflate` codings - these are stored correctly and don't fail checksum validation on `PutObject` when `DISABLE_CUSTOM_CORS_S3=1` is set.

### Expected Behavior

Behaves like production AWS (and LocalStack itself when `DISABLE_CUSTOM_CORS_S3=0`):  an object put with `Content-Encoding: gzip` is stored as-is and when retrieved, is returned with the same `gzip` encoding.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

```yaml
services:
  localstack:
    image: localstack/localstack:4.6.0
    ports:
      - 4566:4566
    environment:
      - DEBUG=1
      - DISABLE_CUSTOM_CORS_S3=1
```

[...]

PR #12855, opened 2025-07-11
S3: fix CORS handler + GZIP handling

bentsku:
## Motivation
As reported by #12854, we had an issue in S3 with GZIP handling when disabling S3 CORS. 

This is because the GZIP handler depends on the service name being set in the context. However, we were only attaching the S3 service name if S3 CORS was not disabled.
This is a consequence of the ServiceNameParser being moved further down the chain with #11800

As a quick fix and bandaid, I've moved the logic around to *always* try to determine if it's an S3 request, even if S3 CORS is disabled.

However, we're muddling a bit the responsibility of the "CORS" handler with a mini service name parser. I'd suggest to get this first PR as a quick fix for the issue, because the first part will always need to be executed anyway.

As a follow-up, we could either split this in 2 handlers: one "S3 Detection" and one "S3 CORS", however the order will need to be respected closely, and the ordering of handlers inside a `CompositeHandler` is not something that is defined. It is rather implicit depending on the order you are appending your handlers, but it could be modified somewhere else. 

My problem is the naming of the "S3CORSHandler", which has more responsibilities than just CORS

So for now I'm proposing to do it this way to quickly solve the issue, which does not really have a cost anyway, but we'll need to address the naming + responsibility issue soon 👀 

## Changes
- always execute the S3 detection in the S3 CORS handler
- add a small test validating the fix

_fixes #12854_

### DIFF
--- a/localstack-core/localstack/services/s3/cors.py
+++ b/localstack-core/localstack/services/s3/cors.py
@@ -97,12 +97,6 @@ class S3CorsHandler(Handler):
         https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html
         https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManageCorsUsing.html
         """
-
-        # this is used with the new ASF S3 provider
-        # although, we could use it to pre-parse the request and set the context to move the service name parser
-        if config.DISABLE_CUSTOM_CORS_S3:
-            return
-
         request = context.request
         is_s3, bucket_name = self.pre_parse_s3_request(context.request)
 
@@ -111,7 +105,13 @@ class S3CorsHandler(Handler):
             return
 
         # set the service so that the regular CORS enforcer knows it needs to ignore this request
+        # we always want to set the service early, because the `ContentDecoder` is very early in the chain and
+        # depends on S3
         context.service = self._service
+
+        if config.DISABLE_CUSTOM_CORS_S3:
+            # we do not apply S3 specific headers if this config flag is set
+            return
 
         is_options_request = request.method == "OPTIONS"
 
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/s3/cors.py
```diff
@@ -111,8 +105,14 @@ def handle_cors(self, chain: HandlerChain, context: RequestContext, response: Re
             return
 
         # set the service so that the regular CORS enforcer knows it needs to ignore this request
+        # we always want to set the service early, because the `ContentDecoder` is very early in the chain and
+        # depends on S3
         context.service = self._service
 
+        if config.DISABLE_CUSTOM_CORS_S3:
```
alexrashed, line 112: **thought:** I'll just ramble around here, just to make sure that we are on the same page when it comes to the (performance) implications of this PR. Please correct me if any of this is wrong 😅 

So this config is `false` by default. This means in the vast majority of cases, we already already did execute the minimum S3 service detection, and for this (default) case, this PR does not have any implications.

However, for users who actively set `DISABLE_CUSTOM_CORS_S3` in their configuration, this PR will change the behavior such that this simplified service routing heuristic is now being executed.

If this is the case, I would say that this change is absolutely reasonable, and the implications are pretty much minimal.

alexrashed (approved):
Nice catch! I added a comment just to verify that I did grasp the implications of this PR correctly, but in my opinion they are minimal and this PR is good to go.

On the broader topic: I think the problem here is just that this is a conceptional problem we are having with our one-dimensional handler chain, in combination with multiple different routing mechanisms (internal endpoints, dynamic edge routes, service routes), and a _clean_ solution to this issue would mean that we would have to have a generalized routing which then invokes service specific handler chains... 🤔

localstack-core/localstack/services/s3/cors.py
bentsku: Exactly, this is exactly it. Thanks for making it clearer 👌

PR #12857, opened 2025-07-11
Increase AWS CLI lower bound version to shorten resolution time

silv-io:
## Motivation

In some circumstances, when other packages with large dependencies install this package, they can have a large resolution tree because of the large amount of awscli versions we allow in the `pyproject.toml` in the runtime extras.

## Changes

- Up the lower bound for the `awscli` dependency in the `runtime` extra, to cut down the resolution tree for dependent projects. This is a safe upgrade because we know for both this project and the dependent one that a working resolution includes the latest version `1.41.3`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ base-runtime = [
 runtime = [
     "localstack-core[base-runtime]",
     # pinned / updated by ASF update action
-    "awscli>=1.37.0",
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM, I believe this is a safe change, and something, as you mentioned, that we should probably automate. Thanks for taking of it 🧹

PR #12864, opened 2025-07-15
CFNv2: support validation of CAPABILITY_AUTO_EXPAND

simonrw:
## Motivation

When creating a stack with transforms, the `CAPABILITY_AUTO_EXPAND` capability is required. This is not currently validated with the new engine

## Changes

* Validate auto expand capability on create and update stacks

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -477,9 +477,8 @@ class CloudformationProviderV2(CloudformationProvider):
         template_body = api_utils.extract_template_body(request)
         structured_template = template_preparer.parse_template(template_body)
 
-        if (
-            "CAPABILITY_AUTO_EXPAND" not in request.get("Capabilities", [])
-            and "Transform" in structured_template.keys()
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -476,6 +477,14 @@ def create_stack(self, context: RequestContext, request: CreateStackInput) -> Cr
         template_body = api_utils.extract_template_body(request)
         structured_template = template_preparer.parse_template(template_body)
 
+        if (
+            "CAPABILITY_AUTO_EXPAND" not in request.get("Capabilities", [])
+            and "Transform" in structured_template.keys()
```
pinzon: **Question**:  we certain the capability is only meant for Global Transformation, or can it be a limiting factor when using Local Transformations and macros?

simonrw: It is also required for macros as well. This is the implementation in the old provider, so if we are not in parity there we are not in parity either, and have not been in parity for a long time!

PR #12865, opened 2025-07-15
CFNv2: Implement get_template

simonrw:
## Motivation

The new provider does not support the `GetTemplate` operation. This PR implements an initial version.

## Changes

* Store a copy of the original (parsed but not transformed) template on the stack
* Implement the provider method basically copying the v1 implementation
* Unskip 3 tests

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -850,8 +850,9 @@ class CloudformationProviderV2(CloudformationProvider):
         after_template = structured_template
 
         previous_update_model = None
-        if previous_change_set := find_change_set_v2(state, stack.change_set_id):
-            previous_update_model = previous_change_set.update_model
+        if stack.change_set_id:
+            if previous_change_set := find_change_set_v2(state, stack.change_set_id):
+                previous_update_model = previous_change_set.update_model
 
         change_set = ChangeSet(
             stack,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -589,6 +592,49 @@ def describe_stack_events(
...
+                "Conditions",
+                "Mappings",
+            ]:
+                copy_template.pop(key, None)
+            for key in ["Parameters", "Outputs"]:
+                if key in copy_template and not copy_template[key]:
+                    copy_template.pop(key)
+            for resource in copy_template.get("Resources", {}).values():
+                resource.pop("LogicalResourceId", None)
+            template_body = json.dumps(copy_template)
```
pinzon: **comment**: Not a fan of this. I believe this is one of the points we discussed in the CFn syncs long ago where we wanted to have an actual attribute of stack that stores the processed template and not  version of stack state converted into a template.

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -589,6 +592,49 @@ def describe_stack_events(
...
+        else:
+            template_body = stack.template_body
+
+        return GetTemplateOutput(
+            TemplateBody=template_body,
+            StagesAvailable=[TemplateStage.Original, TemplateStage.Processed],
```
pinzon, line 662: **question**: Are stages: Original and Processed always available?

simonrw: This is in parity with the old provider, but I will check. Thanks for flagging

simonrw: It seems both options are always available and there are no other options so I think we are good here.

localstack-core/localstack/services/cloudformation/v2/provider.py
simonrw: You have a fair point. I will see about storing the transformed template after we have performed the transform. Thanks

PR #12866, opened 2025-07-15
APIGW: update CFN related snapshots

bentsku:
## Motivation
Follow up from #12826, we took extra care with @ArthurAkh to make sure we would not break upstream pipelines, but the test selection did not trigger on CloudFormation and Scenario tests, making us miss some outdated snapshots (https://github.com/localstack/localstack/runs/46031151305)

This PR regenerated those outdated snapshots

## Changes
- regenerate CFN / Scenario tests related to APIGW
- update snapshot skip related to DynamoDB

### DIFF
--- a/tests/aws/services/cloudformation/resources/test_apigateway.py
+++ b/tests/aws/services/cloudformation/resources/test_apigateway.py
@@ -329,11 +329,8 @@ def test_cfn_deploy_apigateway_integration(deploy_cfn_template, snapshot, aws_cl
     paths=[
         "$.resources.items..resourceMethods.GET",  # TODO: after importing, AWS returns them empty?
         # TODO: missing from LS response
-        "$.get-stage.createdDate",
-        "$.get-stage.lastUpdatedDate",
         "$.get-stage.methodSettings",
         "$.get-stage.tags",
-        "$..endpointConfiguration.ipAddressType",
     ]
 )
 def test_cfn_deploy_apigateway_from_s3_swagger(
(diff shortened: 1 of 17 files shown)


tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_sam.snapshot.json
```diff
@@ -56,7 +57,7 @@
           "Architectures": [
             "x86_64"
           ],
-          "CodeSha256": "+xvKfGS3ENINs/yK7dLJgId2fDM+vv9OP03rJ9mLflU=",
+          "CodeSha256": "W479VjWcFpBg+yx255glPq1ZLEq5WjlmjJi7CmxLFio=",
```
cloutierMat, line 60: nit (non-blocking): I know this is unrelated to this change, but it might be worth transforming this value?

bentsku: agreed, it also troubled me but I realized only after regenerating all of them 🤦‍♂️ I figured this could be tackled as a follow-up. Good catch!

tests/aws/services/cloudformation/resources/test_apigateway.py
```diff
@@ -329,11 +329,8 @@ def test_cfn_deploy_apigateway_integration(deploy_cfn_template, snapshot, aws_cl
     paths=[
         "$.resources.items..resourceMethods.GET",  # TODO: after importing, AWS returns them empty?
         # TODO: missing from LS response
-        "$.get-stage.createdDate",
-        "$.get-stage.lastUpdatedDate",
         "$.get-stage.methodSettings",
         "$.get-stage.tags",
-        "$..endpointConfiguration.ipAddressType",
```
cloutierMat, line 336: :partying_face:

alexrashed (approved):
Thanks so much for jumping on this! 💯 🦸🏽

PR #12867, opened 2025-07-16
Pin aws cli version instead of having a lower bound

silv-io:


.github/workflows/asf-updates.yml
```diff
@@ -80,13 +80,21 @@ jobs:
           bin/release-helper.sh set-dep-ver botocore "==$BOTOCORE_VERSION"
           bin/release-helper.sh set-dep-ver boto3 "==$BOTOCORE_VERSION"
 
+          # determine awscli version that works with $BOTOCORE_VERSION
+          uv venv /tmp/awscli-ver-venv && source /tmp/awscli-ver-venv/bin/activate && uv pip install "botocore==$BOTOCORE_VERSION" awscli && export AWSCLI_VERSION=$(uv pip list --format=json | jq -r '.[] | select(.name=="awscli") | .version') && deactivate
```
alexrashed: **nit:** Since the "run" step in a GitHub action is basically a bash script, I think i would be more readable to split this command into multiple lines instead of chaining it. What do you think?
```suggestion
          uv venv /tmp/awscli-ver-venv
          source /tmp/awscli-ver-venv/bin/activate
          uv pip install "botocore==$BOTOCORE_VERSION" awscli
          export AWSCLI_VERSION=$(uv pip list --format=json | jq -r '.[] | select(.name=="awscli") | .version')
          deactivate
```

silv-io: Yeah makes sense :)

PR #12870, opened 2025-07-17
CFn: correctly set physical resource id for VpnGatewayAttachment

simonrw:
## Motivation

We don't currently set the physical resource id when creating a VPC gateway attachment. This means the new CFn provider fails to deploy the resource.

## Changes

* Correctly set the physical resource id for either the vpn gateway id  or internet gateway id, whichever is provided
* Validate that the physical resource id is set


tests/aws/templates/vpc_gateway_attachment.yml
pinzon, line 1: **Nit:** maybe consider removing the commented out resources

PR #12875, opened 2025-07-18
Introduce jsonpickle serialization backend

giograno:
## Motivation
This PR introduces a new pair of encoder/decoder based on [`jsonpickle`](https://github.com/jsonpickle/jsonpickle).
This is the first step towards a better serialization backend that avoids tight coupling to the code and focuses on serializing data instead.

`jsonpickle` is well suited for this since:
- It serializes instance variables using standard JSON and includes class metadata (e.g., fully qualified class names) to enable rehydration;
- If the original class is no longer importable (e.g., refactoring), `jsonpickle` fails gracefully instead of raising an exception. This allows us to implement custom recovery or migration strategies
- It supports extensibility via custom handlers, enabling fine-grained control over serialization for specific types, similar to pickle’s `__reduce__` and `save_reduce` mechanisms

## Changes
- Introducing a `JsonEncoder` and a `JsonDecoder`;
- The `localstack.state.pickle` `dumps/dump/loads/load` methods get the encoder and the decoder via utility functions, instead of fixed `PickleEncoder/Decoder`.
- Introducing a `ENABLE_JSON_EXPERIMENTAL_BACKEND` variable to switch to the `jsonpickle` codec.

The companion [pro PR](https://github.com/localstack/localstack-pro/pull/4873):
- Uses the `get_default_encoder` and `get_default_decoder` in all the `StateVisitor` implementations to select the codec at runtime;
- Introduces all the `jsonpickle` custom handlers;
- Runs all our persistence test suite with the new `ENABLE_JSON_EXPERIMENTAL_BACKEND` flag enabled.

Implements: PNX-45

### DIFF
--- a/localstack-core/localstack/config.py
+++ b/localstack-core/localstack/config.py
@@ -1254,8 +1254,9 @@ IN_MEMORY_CLIENT = is_env_true("IN_MEMORY_CLIENT")
 # This flag enables all responses from LocalStack to contain a `x-localstack` HTTP header.
 LOCALSTACK_RESPONSE_HEADER_ENABLED = is_env_not_false("LOCALSTACK_RESPONSE_HEADER_ENABLED")
 
-# Enable experimental JSON backend for serialization (`dill` is the default one)
-ENABLE_JSON_EXPERIMENTAL_BACKEND = is_env_true("ENABLE_JSON_EXPERIMENTAL_BACKEND")
+# Serialization backend for the LocalStack internal state (`dill` is used by default`).
+# `jsonpickle` enables the new experimental backend.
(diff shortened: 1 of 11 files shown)


thrau (approved):
Looks pretty good to me!

Perhaps an improvement we could consider (non-blocking) is to rework `ENABLE_JSON_EXPERIMENTAL_BACKEND: bool` to a `STATE_SERIALIZATION_BACKEND: str` variable (name tbd), and in `get_default_encoder/decoder`, do a case match on `STATE_SERIALIZATION_BACKEND`. would perhaps be a bit more flexible. but again not needed for this PR necessarliy.

PR #12876, opened 2025-07-18
add implementation of DeleteChangeSet for CFnV2

pinzon:
## Motivation
This PR implements the DeleteChangeSet for Cloudformation v2 provider.

## Changes
- Operation implementation 
## Testing
- Manual testing with `tests.aws.services.cloudformation.api.test_changesets.test_delete_change_set_exception`.
- The logs pertaining to errors while deletion should be gone.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -473,9 +473,9 @@ class CloudformationProviderV2(CloudformationProvider):
                 "StackName must be specified if ChangeSetName is not specified as an ARN."
             )
 
-        state.stacks_v2.get(change_set.stack.stack_id).change_set_ids.remove(
-            change_set.change_set_id
-        )
+        if not change_set:
+            return DeleteChangeSetOutput()
+
         change_set.stack.change_set_ids.remove(change_set.change_set_id)
         state.change_sets.pop(change_set.change_set_id)
 


localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -84,7 +85,7 @@ def is_changeset_arn(change_set_name_or_id: str) -> bool:
 
 class StackNotFoundError(ValidationError):
     def __init__(self, stack_name: str):
-        super().__init__(f"Stack with id {stack_name} does not exist")
+        super().__init__(f"Stack [{stack_name}] does not exist")
```
simonrw, line 88: This is interesting I have seen both forms. I don't know if AWS have changed the format or it is actually different

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -448,6 +449,30 @@ def describe_change_set(
...
+            change_set = state.change_sets.get(change_set_name)
+        elif not is_changeset_arn(change_set_name) and stack_name:
+            change_set = find_change_set_v2(state, change_set_name, stack_name)
+        else:
+            raise ValidationError(
+                "StackName must be specified if ChangeSetName is not specified as an ARN."
+            )
+
+        change_set.stack.change_set_ids.remove(change_set.change_set_id)
+        state.change_sets.pop(change_set.change_set_id)
```
simonrw, line 480: We also need to remove the change set from stack.change_set_ids I think

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -448,6 +454,33 @@ def describe_change_set(
...
+        self,
+        context: RequestContext,
+        change_set_name: ChangeSetNameOrId,
+        stack_name: StackNameOrId = None,
+        **kwargs,
+    ) -> DeleteChangeSetOutput:
+        state = get_cloudformation_store(context.account_id, context.region)
+
+        if is_changeset_arn(change_set_name):
+            change_set = state.change_sets.get(change_set_name)
```
simonrw, line 468: We need to early return if the change set cannot be found

```suggestion
            change_set = state.change_sets.get(change_set_name)
            if change_set is None:
                # AWS does not raise an error if it cannot find the change set
                return
```

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -448,6 +454,33 @@ def describe_change_set(
...
+        elif not is_changeset_arn(change_set_name) and stack_name:
+            change_set = find_change_set_v2(state, change_set_name, stack_name)
```
simonrw, line 470: We need to early return if the change set cannot be found

```suggestion
            change_set = find_change_set_v2(state, change_set_name, stack_name)
            if change_set is None:
                # AWS does not raise an error if it cannot find the change set
                return
```

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -448,6 +454,33 @@ def describe_change_set(
...
+        else:
+            raise ValidationError(
+                "StackName must be specified if ChangeSetName is not specified as an ARN."
+            )
+
+        state.stacks_v2.get(change_set.stack.stack_id).change_set_ids.remove(
+            change_set.change_set_id
+        )
```
simonrw: This should not be needed

```suggestion
```

localstack-core/localstack/services/cloudformation/v2/provider.py
simonrw: This comment is actually an error and I did not read the code properly

simonrw (approved):
Pair-reviewed with @pinzon

PR #12880, opened 2025-07-21
Stop pushing empty release commit for community

silv-io:
## Motivation

With the removal of CircleCI and #12829 we removed the need of having an empty commit to trigger the CI on releases of `localstack-core` and `localstack/localstack`. 

We can now proceed to trigger the pipeline purely via the tag on the latest commit in the `main` branch.

## Changes

- remove `--allow-empty` in the release commit

## Testing

- running `git commit -m "test"` on a clean working tree just skips the creation of a commit. Subsequently running `git tag` will then put it on the latest commit in the branch. 

## TODO

What's left to do:

- [ ] Decide whether we should remove `--allow-empty` for the development iteration branch as well or not. In theory this commit is also unnecessary
- [ ] ...

### DIFF
--- a/bin/release-helper.sh
+++ b/bin/release-helper.sh
@@ -175,8 +175,7 @@ function cmd-git-commit-release() {
     echo $1 || verify_valid_version
 
     git add "${DEPENDENCY_FILE}"
-    # allow empty commit here as the community version might not have any changes, but we still need a commit for a tag
-    git commit --allow-empty -m "release version ${1}"
+    git commit -m "release version ${1}"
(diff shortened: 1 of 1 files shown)


alexrashed:
@silv-io What's the status of this PR?

silv-io:
@alexrashed I never got to thoroughly test this and it was just an optional minor improvement we could've made after switching to triggering the release via tags. Also, since the way we're releasing will change considerably in the coming months, I think we can close this 👍🏼

PR #12885, opened 2025-07-21
Add proper validation for STS role arn and session name parameters

dfangl:
## Motivation
We currently do not validate the role arn or the role session name in LocalStack.

This can lead to weird issues when parsing the ARN, like empty account ids or key errors.

With this PR, we make sure the role arn and session name will satisfy some (incredibly basic) requirements, the same as AWS imposes.

## Changes
* Better validation of role arn and role session name for the assume role call

### DIFF
--- a/localstack-core/localstack/services/sts/provider.py
+++ b/localstack-core/localstack/services/sts/provider.py
@@ -1,6 +1,7 @@
 import logging
+import re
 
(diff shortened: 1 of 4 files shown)


pinzon (approved):
Great more parity :+1:

Issue #10921, opened 2024-05-29
bug: importKeyMaterial fails with "Unsupported padding, requested wrapping algorithm:'RSA_AES_KEY_WRAP_SHA_256'"


SzyszkowskaAgnieszka:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently when I follow the AWS instruction to importKeyMaterial for asymmetric RSA key using localstack and aws-cli, I'm getting the following exception:
An error occurred (KMSInvalidStateException) when calling the ImportKeyMaterial operation: Unsupported padding, requested wrapping algorithm:'RSA_AES_KEY_WRAP_SHA_256'

### Expected Behavior

It is possible to import key material for RSA asymmetric key to localstack.

### How are you starting LocalStack?

With the `localstack` script

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    localstack start -d

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

    aws kms create-key --origin EXTERNAL --key-spec RSA_4096 --profile localstack

    aws kms get-parameters-for-import --key-id b0f79e39-4636-44c3-a6c1-0967b8eb1a8c --wrapping-algorithm RSA_AES_KEY_WRAP_SHA_256 --wrapping-key-spec RSA_4096 --profile localstack

    aws kms import-key-material --key-id b0f79e39-4636-44c3-a6c1-0967b8eb1a8c --encrypted-key-material fileb://EncryptedKeyMaterial.bin --import-token fileb://ImportToken.bin --expiration-model KEY_MATERIAL_DOES_NOT_EXPIRE --profile localstack   

I'm following this instruction: AWS KMS documentation: https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys-create-cmk.html for all openssl commands.

### Environment

```markdown
- OS: MacOs 14.5 (23F79)
- LocalStack:
   LocalStack version: 3.4.1.dev
   LocalStack Docker image sha: sha256:cf62a941b36ba7cc49c27e1a9cc01cc17c9512a3ca179b65b1fdcea64ba5b46c
   LocalStack build date: 2024-05-29
   LocalStack build git hash: df6f0a42e
- aws-cli: 2.13.7
```

### Anything else?

Is this wrapping algorithm even supported by localstack? There are no community tests to prove it, but also there is no information in the documentation that it's not supported.

nunofernandes:
[...]

PR #12888, opened 2025-07-22
KMS: Add support for `RSA_AES_KEY_WRAP_SHA_256` in import key material

sannya-singal:
## Motivation
This PR adds support for the wrapping algorithm `RSA_AES_KEY_WRAP_SHA_256` in the `ImportKeyMaterial`, which raises:

`KMSInvalidStateException: An error occurred (KMSInvalidStateException) when calling the ImportKeyMaterial operation: Unsupported padding, requested wrapping algorithm:'RSA_AES_KEY_WRAP_SHA_256'`

This algorithm is defined by AWS as a hybrid encryption scheme, that combines encrypting the key material with an AES symmetric key that is generated, and encrypting the AES symmetric key with the public wrapping key and the `RSAES_OAEP_SHA_256` wrapping algorithm.

## Changes
- Added support for `RSA_AES_KEY_WRAP_SHA_256` in the `import_key_material` method.
- Refactored decryption of wrapped key material into a new, centralized helper: `_decrypt_wrapped_key_material(...) ` now handles both RSA-only and hybrid RSA+AES key wrapping formats.
- Added AWS validated snapshot test: `test_import_key_rsa_aes_wrap_sha256` to validate the implementation. 

## References:
- [AWS official documentation for Importing Key Material](https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys-conceptual.html)
- [RFC 5649: AES Key Wrap with Padding](https://datatracker.ietf.org/doc/html/rfc5649)

Closes: https://github.com/localstack/localstack/issues/10921

### DIFF
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -1561,32 +1561,35 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
         algo = import_state.wrapping_algo
         decrypt_key = import_state.key.crypto_key.key
 
-        if algo == AlgorithmSpec.RSAES_PKCS1_V1_5:
-            padding_scheme = padding.PKCS1v15()
-        elif algo == AlgorithmSpec.RSAES_OAEP_SHA_1:
-            padding_scheme = padding.OAEP(padding.MGF1(hashes.SHA1()), hashes.SHA1(), None)
-        elif algo == AlgorithmSpec.RSAES_OAEP_SHA_256:
-            padding_scheme = padding.OAEP(padding.MGF1(hashes.SHA256()), hashes.SHA256(), None)
-        elif algo == AlgorithmSpec.RSA_AES_KEY_WRAP_SHA_256:
-            rsa_key_size_bytes = decrypt_key.key_size // 8
-            wrapped_aes_key = encrypted_key_material[:rsa_key_size_bytes]
-            wrapped_key_material = encrypted_key_material[rsa_key_size_bytes:]
+        match algo:
+            case AlgorithmSpec.RSAES_PKCS1_V1_5:
+                padding_scheme = padding.PKCS1v15()
+            case AlgorithmSpec.RSAES_OAEP_SHA_1:
+                padding_scheme = padding.OAEP(padding.MGF1(hashes.SHA1()), hashes.SHA1(), None)
+            case AlgorithmSpec.RSAES_OAEP_SHA_256:
+                padding_scheme = padding.OAEP(padding.MGF1(hashes.SHA256()), hashes.SHA256(), None)
+            case AlgorithmSpec.RSA_AES_KEY_WRAP_SHA_256:
+                rsa_key_size_bytes = decrypt_key.key_size // 8
+                wrapped_aes_key = encrypted_key_material[:rsa_key_size_bytes]
+                wrapped_key_material = encrypted_key_material[rsa_key_size_bytes:]
 
-            aes_key = decrypt_key.decrypt(
-                wrapped_aes_key,
-                padding.OAEP(
-                    mgf=padding.MGF1(algorithm=hashes.SHA256()),
-                    algorithm=hashes.SHA256(),
-                    label=None,
-                ),
-            )
-            return keywrap.aes_key_unwrap_with_padding(
-                aes_key, wrapped_key_material, default_backend()
-            )
-        else:
-            raise KMSInvalidStateException(
-                f"Unsupported padding, requested wrapping algorithm: '{algo}'"
-            )
+                aes_key = decrypt_key.decrypt(
+                    wrapped_aes_key,
+                    padding.OAEP(
+                        mgf=padding.MGF1(algorithm=hashes.SHA256()),
+                        algorithm=hashes.SHA256(),
+                        label=None,
+                    ),
+                )
+
+                return keywrap.aes_key_unwrap_with_padding(
+                    aes_key, wrapped_key_material, default_backend()
+                )
+
+            case _:
+                raise KMSInvalidStateException(
+                    f"Unsupported padding, requested wrapping algorithm: '{algo}'"
+                )
 
         return decrypt_key.decrypt(encrypted_key_material, padding_scheme)
 
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -1564,10 +1564,13 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
         match algo:
             case AlgorithmSpec.RSAES_PKCS1_V1_5:
                 padding_scheme = padding.PKCS1v15()
+                return decrypt_key.decrypt(encrypted_key_material, padding_scheme)
             case AlgorithmSpec.RSAES_OAEP_SHA_1:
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/services/kms/provider.py
```diff
@@ -1564,6 +1553,43 @@ def _validate_grant_request(self, data: Dict):
...
+            )
+            return keywrap.aes_key_unwrap_with_padding(
+                aes_key, wrapped_key_material, default_backend()
+            )
+        else:
+            raise KMSInvalidStateException(
+                f"Unsupported padding, requested wrapping algorithm: '{algo}'"
+            )
+
+        return decrypt_key.decrypt(encrypted_key_material, padding_scheme)
```
k-a-il: comment:  this code is somewhat harder to read. I propose to use match-case statement instead of ifs, it will significantly improve readability
```
        match algo:
            case AlgorithmSpec.RSAES_PKCS1_V1_5:
                return decrypt_key.decrypt(
                    encrypted_key_material,
                    padding.PKCS1v15()
                )
            case AlgorithmSpec.RSA_AES_KEY_WRAP_SHA_256:
                rsa_key_size_bytes = decrypt_key.key_size // 8
                wrapped_aes_key = encrypted_key_material[:rsa_key_size_bytes]
                wrapped_key_material = encrypted_key_material[rsa_key_size_bytes:]

                aes_key = decrypt_key.decrypt(
                    wrapped_aes_key,
                    padding.OAEP(
                        mgf=padding.MGF1(algorithm=hashes.SHA256()),
                        algorithm=hashes.SHA256(),
                        label=None
                    )
                )
...
```

localstack-core/localstack/services/kms/provider.py
```diff
@@ -1564,6 +1553,43 @@ def _validate_grant_request(self, data: Dict):
...
+                    label=None,
+                ),
+            )
+            return keywrap.aes_key_unwrap_with_padding(
+                aes_key, wrapped_key_material, default_backend()
+            )
+        else:
+            raise KMSInvalidStateException(
+                f"Unsupported padding, requested wrapping algorithm: '{algo}'"
+            )
```
k-a-il: comment: I propose not to raise an exception in this helper method, as it breaks the flow. Since this method is just responsible for decrypting the key, it should return None if decryption fails. Then we can raise an exception in the main `import_key_material` method instead

localstack-core/localstack/services/kms/provider.py
sannya-singal: I agree 💯 match case would improve the readability. But I think keeping the common `decrypt(...)` call outside the match block (similar to how it was with if-else) to avoid repeating it makes more sense to me as it keeps things a bit cleaner and readable with assigning `padding_scheme` as per the algo specification.

localstack-core/localstack/services/kms/provider.py
sannya-singal: Thanks for the suggestion 🙏 A few thoughts here: I would need a bit more context on what exactly is meant by 

> breaking the flow

Could you please explain here what kind of break this causes? 

Just to note, for the point: 

> it should return None if decryption fails. 

We are not raising an exception when decryption fails but rather for unsupported algo specs. Thats why I would prefer to keep this exception here. Returning `None` could cause to silently miss this failure here in `_decrypt_wrapped_key_material` and can be misleading as it would mean that the response to decryption is `None` which ideally is not possible in this context, and since this method is already tightly scoped to handle this, I think it makes sense for it to explicitly raise if something unsupported is passed since this is expected. 
I guess this also comes down to individual preference, I find explicit exceptions in small and focused helper functions easier to reason rather than returning `None` and checking later in every function that utilises this helper function.

k-a-il: >Could you please explain here what kind of break this causes?

It makes the code harder to follow. It's clearer if the main function returns the exceptions

> I guess this also comes down to individual preference, I find explicit exceptions in small and focused helper functions easier to reason rather than returning None and checking later in every function that utilises this helper function.

Right now, this function is only used in one place. But if we use it somewhere else later, the 'algorithm not supported' error might not be expected as well

localstack-core/localstack/services/kms/provider.py
k-a-il: I propose to use match case because logic for selecting padding is mixed with decryption and unwrapping logic.

sannya-singal: I have already implemented the match case 👍

localstack-core/localstack/services/kms/provider.py
sannya-singal: > Right now, this function is only used in one place. But if we use it somewhere else later, the 'algorithm not supported' error might not be expected as well

Yes, that was a general sentiment. 

For this case I strongly feel returning `None` can be misleading and cause more confusion - without any information on why it returned `None`, even adhering to the best practices followed in our code base and the fail fast principle.

localstack-core/localstack/services/kms/provider.py
k-a-il: I also meant that the decrypted value should be returned directly, as shown in the example above.

sannya-singal: And still return `aes_key_unwrap_with_padding` for `RSA_AES_KEY_WRAP_SHA_256`?

sannya-singal: I am not fully convinced yet 😄 But happy to make the changes 🙂👍

k-a-il: What I meant was to return results directly from each match case. Full example:
```

match algo:
            case AlgorithmSpec.RSAES_PKCS1_V1_5:
                return decrypt_key.decrypt(encrypted_key_material, padding.PKCS1v15())
            case AlgorithmSpec.RSAES_OAEP_SHA_1:
                return decrypt_key.decrypt(encrypted_key_material, padding.OAEP(padding.MGF1(hashes.SHA1()), hashes.SHA1(), None))
            case AlgorithmSpec.RSAES_OAEP_SHA_256:
                return decrypt_key.decrypt(encrypted_key_material, padding.OAEP(padding.MGF1(hashes.SHA256()), hashes.SHA256(), None))
            case AlgorithmSpec.RSA_AES_KEY_WRAP_SHA_256:
                rsa_key_size_bytes = decrypt_key.key_size // 8
                wrapped_aes_key = encrypted_key_material[:rsa_key_size_bytes]
                wrapped_key_material = encrypted_key_material[rsa_key_size_bytes:]

                aes_key = decrypt_key.decrypt(
                    wrapped_aes_key,
                    padding.OAEP(
                        mgf=padding.MGF1(algorithm=hashes.SHA256()),
                        algorithm=hashes.SHA256(),
                        label=None,
                    ),
                )

                return keywrap.aes_key_unwrap_with_padding(
                    aes_key, wrapped_key_material, default_backend()
                )

            case _:
                raise KMSInvalidStateException(
                    f"Unsupported padding, requested wrapping algorithm: '{algo}'"
                )
```

sannya-singal: I have already made the changes 👍

localstack-core/localstack/services/kms/provider.py
k-a-il: I understand your point, but raising an AWS-specific exception here mixes responsibilities between decryption and API error handling.

That said, I’m fine to leave it here and move on.

PR #12891, opened 2025-07-22
Step Functions: variable-reference parsing to skip string and regex literals

MEPalma:
## Motivation
Currently, the SFN v2 interpreter’s logic for extracting variable references from JSONata expressions can misinterpret dollar signs inside quoted strings or regex literals as variable markers. That leads to spurious validation errors at runtime reporting illegal uses of “$” even when they occur only within "/…/" or "…", ’…’ literals.

## Changes
- Updates the variable-reference regex to first consume any JSONata regex literals (`/…/`) and string literals (`"…"` or `'…'`), including support for escaped slashes and escaped quotes
- Retains the original rules for matching `$$`, `$identifier(.prop…)*` and lone `$`, but only after skipping over consumed literals
- Ensures that only actual variable references are returned by findall
- Add related positive and negative snapshot tests

### DIFF
--- a/localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
+++ b/localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
@@ -17,13 +17,19 @@ VariableReference = str
 VariableDeclarations = str
 
 
-# TODO: move the extraction logic to a formal ANTLR-base
-#       parser, as done with legacy Intrinsic Functions.
+# TODO: move the extraction logic to a formal ANTLR-base parser, as done with legacy
+#       Intrinsic Functions in package localstack.services.stepfunctions.asl.antlr
+#       with grammars ASLIntrinsicLexer and ASLIntrinsicParser, later used by upstream
+#       logics such as in:
+#       localstack.services.stepfunctions.asl.parse.intrinsic.preprocessor.Preprocessor
 _PATTERN_VARIABLE_REFERENCE = re.compile(
-    # 1) Non-capturing branch for JSONata regex literal /.../, allowing escaped \/
+    # 1) Non-capturing branch for JSONata regex literal
+    #    /.../ (slash delimited), allowing escaped slashes \/
     r"(?:\/(?:\\.|[^\\/])*\/[a-zA-Z]*)"
     r"|"
-    # 2) Non-capturing branch for JSONata string literal: "..." or '...', allowing escapes
+    # 2) Non-capturing branch for JSONata string literal:
+    #    "..." (double quotes) or '...' (single quotes),
+    #    allowing escapes
     r"(?:\"(?:\\.|[^\"\\])*\"|\'(?:\\.|[^\'\\])*\')"
     r"|"
     # 3) Capturing branch for $$, $identifier[.prop…], or lone $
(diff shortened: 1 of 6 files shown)


tests/aws/services/stepfunctions/v2/evaluate_jsonata/test_base_evaluate_expressions.snapshot.json
```diff
@@ -1383,13 +1608,13 @@
                     "Date": "date",
                     "X-Amz-Executed-Version": "$LATEST",
                     "x-amzn-Remapped-Content-Length": "0",
-                    "x-amzn-RequestId": "<uuid:1>",
+                    "x-amzn-RequestId": "x-amzn-RequestId",
```
joe4dev, line 1611: sidenote(nit): Shoudn't this transformer do a reference replacement to maintain parity information that the RequestId should match across these 3 fields? That transformer was probably introduced somewhere else.

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
```diff
@@ -16,9 +16,20 @@
 VariableReference = str
 VariableDeclarations = str
 
-_PATTERN_VARIABLE_REFERENCE: Final[re.Pattern] = re.compile(
-    r"\$\$|\$[a-zA-Z0-9_$]+(?:\.[a-zA-Z0-9_][a-zA-Z0-9_$]*)*|\$"
+
+# TODO: move the extraction logic to a formal ANTLR-base
+#       parser, as done with legacy Intrinsic Functions.
```
joe4dev: docs(nit): Can we add a link to the example of legacy Intrinsic Functions as a starting point on how such an ANTL-based implementation could look like?

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
```diff
@@ -16,9 +16,20 @@
...
+_PATTERN_VARIABLE_REFERENCE = re.compile(
+    # 1) Non-capturing branch for JSONata regex literal /.../, allowing escaped \/
```
joe4dev: praise: constructive comments to distinguish consuming JSONata literals from matching rules 👏 
The complexity of the regex is becoming increasingly difficult to manage/understand, hence such clarifying comments are essential.

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
```diff
@@ -16,9 +16,20 @@
...
+    r"(?:\/(?:\\.|[^\\/])*\/[a-zA-Z]*)"
+    r"|"
+    # 2) Non-capturing branch for JSONata string literal: "..." or '...', allowing escapes
```
joe4dev: supernit: clarifying the literal as noun can help to avoid confusion of trying to look for `...` because literals are often presented in double or single quotes ✨ . Suggestion:

> Non-capturing branch for JSONata string literal: "..." (double quotes) or '...' (single quotes), allowing escapes

Same above for slash

localstack-core/localstack/services/stepfunctions/asl/jsonata/jsonata.py
MEPalma: The entire SFN test suite uses the `sfn_snapshot` fixture, which adds a set of SFN-related transformers. Unfortunately this also adds the transformer you highlighted here; changing this would mean updating all the snapshots for this service.

PR #12892, opened 2025-07-22
CFnV2: add stack events when there are errors during preprocessing

simonrw:
## Motivation

While the provider is still new with edge cases, it's frustrating to find out the cause of the error if it occurs during preprocessing.

With this change, a stack event is captured during resource deployment time so at least the error message is shown in the stack events.

## Changes

* Add stack event on resource preprocessing failure

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -177,7 +177,19 @@ class ChangeSetModelExecutor(ChangeSetModelPreproc):
         Overrides the default preprocessing for NodeResource objects by annotating the
         `after` delta with the physical resource ID, if side effects resulted in an update.
         """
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -109,6 +110,14 @@ class ChangeType(enum.Enum):
...
         return self.value
 
+    def to_change_action(self) -> ChangeAction:
+        # Convert this change type into the change action used throughout the CFn API
+        return {
+            ChangeType.CREATED: ChangeAction.Add,
+            ChangeType.MODIFIED: ChangeAction.Modify,
+            ChangeType.REMOVED: ChangeAction.Remove,
+        }.get(self, ChangeAction.Add)
+
```
pinzon, line 120: **comment**: Nice! :+1:

PR #12902, opened 2025-07-24
feat: add the pro activation to the tracked environment variables

kittylon:
## Motivation

Tracking the `ACTIVATE_PRO` will enable us to identify sessions that use the pro image with an invalid token and fall back to the community image.

**Context:**
Some users set an invalid token using the pro image and set the `ACTIVATE_PRO=0` variable.
Without this, it appears that someone started a session with an invalid token using the pro image, which is misleading, as they would fall back to the community in this situation.

## What changes does this PR make?
Adds `ACTIVATE_PRO` to the list of tracked environment variables that are sent to the analytics backend, so that it's ingested into TinyBird.


kittylon:

valentinarmo (approved):

PR #12905, opened 2025-07-24
handle unsupported resource in CFn v2

pinzon:
## Motivation
- This PR adds to the CFn v2 the capability to handle  unsupported resources in a better way.  

## Changes
- if `CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES` is set to true the engine will ignore nonexistent resource providers and return `unknown` when a property  is requested.
- Removal of duplicated code.

## Testing
- Unskipped test.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -22,6 +22,7 @@ from localstack.services.cloudformation.engine.v2.change_set_model import (
     is_nothing,
 )
 from localstack.services.cloudformation.engine.v2.change_set_model_preproc import (
+    MOCKED_REFERENCE,
     ChangeSetModelPreproc,
     PreprocEntityDelta,
     PreprocOutput,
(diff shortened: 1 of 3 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -389,22 +388,27 @@ def _execute_resource_action(
...
-                stack.set_resource_status(
-                    logical_resource_id=logical_resource_id,
-                    # TODO,
-                    physical_resource_id="",
-                    resource_type=resource_type,
-                    status=ResourceStatus.CREATE_FAILED
-                    if action == ChangeAction.Add
-                    else ResourceStatus.UPDATE_FAILED,
-                    resource_status_reason=reason,
-                )
```
pinzon, line 402: The processing of a failed event is handled at a level above. So these lines just add repetition to that

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -366,18 +366,9 @@ def _execute_resource_action(
         )
         resource_provider = resource_provider_executor.try_load_resource_provider(resource_type)
         track_resource_operation(action, resource_type, missing=resource_provider is not None)
-        if resource_provider is None:
-            log_not_available_message(
-                resource_type,
-                f'No resource provider found for "{resource_type}"',
-            )
-            if not config.CFN_IGNORE_UNSUPPORTED_RESOURCE_TYPES:
-                raise NoResourceProvider
```
pinzon, line 375: @simonrw I believe that raising an error would confuse and alert users. Having the deployment fail and explaining why seems to me like a better experience.

PR #12906, opened 2025-07-24
add support for Exports/Imports in CFn v2

pinzon:
## Motivation
This PR enables the functionality of Exports/Imports in CFn v2

## Changes
- Implement ListExports operation
- ChangeSetModelExecutor stores output values that also define an Export name.
- ChangeSetPreproc is able to resolve imports

## Testing
- Unskipped tests

## Todo
- 2 exports cannot use the same name in the same template
- An export cannot be registered if the name already exists in the exports store.


localstack-core/localstack/services/cloudformation/stores.py
```diff
@@ -50,6 +50,18 @@ def exports(self):
                 output_keys[export_name] = stack.stack_id
         return exports
 
+    def exports_v2(self) -> dict:
+        stacks = self.stacks_v2.values()
+        exports = {}
+        for stack in stacks:
+            for export_name, export_value in stack.resolved_exports.items():
```
simonrw, line 61: **issue**: surely we have to filter for active stacks?

localstack-core/localstack/services/cloudformation/stores.py
```diff
@@ -50,6 +50,18 @@ def exports(self):
                 output_keys[export_name] = stack.stack_id
         return exports
 
+    def exports_v2(self) -> dict:
```
simonrw: **question**: what is the reason to make this a method rather than a property like we had before?

localstack-core/localstack/services/cloudformation/stores.py
```diff
@@ -132,3 +144,8 @@ def exports_map(account_id: str, region_name: str):
     for export in store.exports:
         result[export["Name"]] = export
     return result
+
+
+def exports_map_v2(account_id: str, region_name: str):
```
simonrw: **nit**: I'm not a fan of having this separate free-floating function to access store state. We don't have it anywhere else, however it is not the end of the world, and could be cleaned up later.

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
```diff
@@ -887,6 +890,29 @@ def _compute_fn_base_64(string) -> Any:
         )
         return delta
 
+    def visit_node_intrinsic_function_fn_import_value(
+        self, node_intrinsic_function: NodeIntrinsicFunction
+    ) -> PreprocEntityDelta:
+        def _compute_fn_import_value(string) -> Any:
```
simonrw:
```suggestion
        def _compute_fn_import_value(string) -> str:
```

localstack-core/localstack/services/cloudformation/stores.py
simonrw:
```suggestion
    def exports_v2(self) -> dict[str, Export]:
```

localstack-core/localstack/services/cloudformation/stores.py
```diff
@@ -50,6 +50,18 @@ def exports(self):
...
+    def exports_v2(self) -> dict:
+        stacks = self.stacks_v2.values()
+        exports = {}
+        for stack in stacks:
+            for export_name, export_value in stack.resolved_exports.items():
+                exports[export_name] = {
+                    "ExportingStackId": stack.stack_id,
+                    "Name": export_name,
+                    "Value": export_value,
+                }
```
simonrw: **suggestion**: I have been trying to use the types from the API dir when building up values like this


```suggestion
                exports[export_name] = Export(
                    ExportingStackId=stack.stack_id,
                    Name=export_name,
                    Value=export_value,
                )
```

localstack-core/localstack/services/cloudformation/stores.py
simonrw:
```suggestion
def exports_map_v2(account_id: str, region_name: str) -> dict[str, Export]:
```

simonrw (changes requested):
Thanks for adding these changes. The blocker is filtering the stacks by status. Perhaps we can add a test for this case as well, i.e. create two stacks with exports, delete one stack then call `list_exports`?

pinzon:
Thanks for the comments. after implementing the changes the feature code feels much better :+1:

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
```diff
@@ -901,6 +904,29 @@ def _compute_fn_base_64(string) -> Any:
...
+            return exports.get(string)["Value"]
+
+        arguments_delta = self.visit(node_intrinsic_function.arguments)
+        delta = self._cached_apply(
+            scope=node_intrinsic_function.scope,
+            arguments_delta=arguments_delta,
+            resolver=_compute_fn_import_value,
+        )
+        return delta
+
```
simonrw: **issue**: I think this lookup happens only in the executor. For example when describing a change set, the import is not retrieved. 

I can reproduce this by
* creating a change set including an `ImportValue`, e.g.
```yaml
Resources:
  MyFoo:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: !ImportValue MyGlobalExportValue
```
* describing the change set with `aws describe-change-set ... --include-property-values`
On AWS the `Value` field is `KNOWN_AFTER_APPLY`, but on LocalStack this code crashes because the export is fetched (but doesn't exist).

Can we add a test to cover this behaviour and implement the correct behaviour in the executor only.

localstack-core/localstack/services/cloudformation/stores.py
```diff
@@ -23,12 +23,15 @@ class CloudFormationStore(BaseStore):
     # maps macro ID to macros
     macros: dict[str, dict] = LocalAttribute(default=dict)
 
-    # exports: dict[str, str]
     @property
-    def exports(self):
+    def exports(self) -> Exports:
```
simonrw: **issue**: in general I would prefer to have an `exports` dict in the store, and on successful stack deploy update this state (like we do with other state fields on the stack). Then we can explicitly manage the lifecycle of exports (i.e. make updating the exports field atomic, rather than risk either inconsistent state when a stack is deploying, or bugs in partial stack failures). I don't like this dynamic exports property in general (I know we had it before), but I realise that will be more work.

I would also prefer to keep the v2 and v1 state separate, rather than merging them in the way you have here. It's then clear what behaviour needs to be removed when we remove the old provider (and global search/replace of `exports_v2` is going to be super easy compared to decomposing this function, WDYT?

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
pinzon: I added a test and it passes with the functionality in the Preproc. maybe I'm not understanding correctly the issue.

tests/aws/services/cloudformation/test_change_sets_exports_imports.py
```diff
@@ -0,0 +1,75 @@
...
+from localstack.aws.api.cloudformation import ChangeSetType
+from localstack.services.cloudformation.v2.utils import is_v2_engine
+from localstack.testing.aws.util import is_aws_cloud
+from localstack.testing.pytest import markers
+from localstack.utils.strings import short_uid
+
+
+@pytest.mark.skipif(
+    condition=not is_v2_engine() and not is_aws_cloud(), reason="Requires the V2 engine"
+)
```
simonrw:
```suggestion
@skip_if_v1_provider(reason="Not implemented in the V1 provider")
)
```

tests/aws/services/cloudformation/api/test_transformers.py
```diff
@@ -13,6 +13,8 @@
 from localstack.utils.functions import call_safe
 from localstack.utils.strings import short_uid, to_bytes
 
+skip_if_v2_provider(reason="Fn transform not implemented yet")
```
simonrw: **issue**: this does not work with the markers. They should be used as decorators. 

If you want to skip the entire file you need to set `pytestmark` to this variable, e.g.

```python
pytestmark = skip_if_v2_provider(reason="Fn::Transform not implemented yet")
```

tests/aws/services/cloudformation/test_change_sets_exports_imports.py
simonrw, line 1: **nit**: the other files in this directory are named `test_change_set_*`, could you rename the file?

tests/aws/services/cloudformation/test_change_sets_exports_imports.py
```diff
@@ -0,0 +1,75 @@
...
+        export_name = f"b-{short_uid()}"
+        deploy_cfn_template(
+            template_path=os.path.join(
+                os.path.dirname(__file__), "../../templates/cfn_function_export.yml"
+            ),
+            parameters={"BucketExportName": export_name},
+        )
+
+        empty_stack_name = f"stack-{short_uid()}"
+        cleanup_stacks([empty_stack_name])
```
simonrw: **question**: what is this block doing? I think it's safe to assume the stack does not exist based on the randomness of `short_uid`

tests/aws/services/cloudformation/test_change_sets_exports_imports.py
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+
+        change_set_name = f"change-set-{short_uid()}"
+        aws_client.cloudformation.create_change_set(
+            StackName=empty_stack_name,
+            ChangeSetName=change_set_name,
+            TemplateBody=json.dumps(change_set_body),
+            ChangeSetType=ChangeSetType.CREATE,
+        )
```
simonrw: **nit**: we need to clean up the change set so we don't leave it lying around

```suggestion
        aws_client.cloudformation.create_change_set(
            StackName=empty_stack_name,
            ChangeSetName=change_set_name,
            TemplateBody=json.dumps(change_set_body),
            ChangeSetType=ChangeSetType.CREATE,
        )
        cleanups.append(lambda: aws_client.cloudformation.delete_change_set(...))
```

tests/aws/services/cloudformation/test_change_sets_exports_imports.py
```diff
@@ -0,0 +1,75 @@
...
+        aws_client.cloudformation.get_waiter("change_set_create_complete").wait(
+            ChangeSetName=change_set_name, StackName=empty_stack_name, WaiterConfig={"Delay": 2}
+        )
+
+        describe = aws_client.cloudformation.describe_change_set(
+            StackName=empty_stack_name, ChangeSetName=change_set_name, IncludePropertyValues=True
+        )
+
```
simonrw: **suggestion**: so we can get more API coverage?

```suggestion
        describe = aws_client.cloudformation.describe_change_set(
            StackName=empty_stack_name, ChangeSetName=change_set_name, IncludePropertyValues=False
        )

        describe_with_prop_values = aws_client.cloudformation.describe_change_set(
            StackName=empty_stack_name, ChangeSetName=change_set_name, IncludePropertyValues=True
        )
```

tests/aws/services/cloudformation/test_template_engine.py
```diff
@@ -756,6 +756,8 @@ def test_snipped_scope(
         snapshot.match("original_template", original_template)
         snapshot.match("processed_template", processed_template)
 
+    skip_if_v2_provider(reason="Fn:transform not implemented")
```
simonrw: See the comment above about using this as a decorator

simonrw:
@pinzon I have added a commit that includes both scenarios I was thinking of. Now we have coverage of:
* create stack with export, create change set using export, describe change set, execute change set, describe stack
* create change set using (missing) export, describe change set, execute change set (which fails), describe failed stack
* create change set using (missing) export, describe change set, create stack with export, execute change set, describe (successful) stack using export
I had to add some minor additional skips that I think are ok for now, and merged `main` into this branch to pick up https://github.com/localstack/localstack/pull/12689 that I merged this morning. If this PR is green I will approve and merge, thanks!

simonrw (approved):

PR #12909, opened 2025-07-25
CFNv2: Support stack sets

simonrw:
## Motivation

The CloudFormation V1 provider had basic support for stack sets, with a single test covering the functionality. This functionality should be ported over to the new provider.

## Changes

* Implement stack sets in parity with the old provider (i.e. unskip the test)
* Implement instance and stack deletion when deleting a stack set (beyond parity)
* Add additional tests for edge cases of creating stack instances for non-existent stacks
* Add test covering deleting a non-existent stack set

## Testing

Unfortunately there is manual set up required for testing stack sets. See the docs here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacksets-prereqs-self-managed.html so this change does not test/assert the behaviour of deployed resources though they were manually verified.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -7,6 +7,7 @@ from typing import Any, Optional
 
 from localstack.aws.api import RequestContext, handler
 from localstack.aws.api.cloudformation import (
+    CallAs,
     Changes,
     ChangeSetNameOrId,
     ChangeSetNotFoundException,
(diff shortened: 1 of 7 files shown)


localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -667,6 +721,195 @@ def describe_stack_resources(
...
+                stack_set.stack_instances.append(stack_instance)
+
+                stacks_to_await.append((stack_name, account, region))
+
+        # wait for completion of stack
+        for stack_name, account_id, region_name in stacks_to_await:
+            client = connect_to(
+                aws_access_key_id=account_id, region_name=region_name
+            ).cloudformation
+            client.get_waiter("stack_create_complete").wait(StackName=stack_name)
```
pinzon, line 787: **comment:** is this correct? Maybe a better approach would be to put this code in a separate thread and update the stack set operation status depending on the result of the waits.

pinzon (approved):
LGTM. Just a minor comment.

localstack-core/localstack/services/cloudformation/v2/provider.py
simonrw: I think you are right but I think you have two suggestions here:
1. run each stack instance concurrently in a background thread
2. don't assume success and handle the error
I don't think point 1. is worth tackling at this stage, however point 2 is very relevant. I would like to tackle this in a follow up since this PR does not support successfully deploying stack instances. See the [note here](https://github.com/localstack/localstack/pull/12909/files#diff-a648c06c242bacf1b1fcdf32c010c21ba487a32bb682d04aa9130476c301a7f5R43).

I have started a PR to create a validated stack sets test. I think the step after **that** is to handle error cases. Thanks!

PR #12911, opened 2025-07-25
implement string replacement feature in CFn v2

pinzon:
## Motivation
This PR implements the capability to replace AWS URLs in CFn templates for CFn v2.

## Changes
- Override `visit_terminal_value_created` to replace strings if necessary

## Testing
- more unskipped tests

## Notes
- Possibly needs to be implemented also in `visit_terminal_value_updated`.
- Consider moving the code to the Preproc.

### DIFF
--- a/tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_lambda.py
+++ b/tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_lambda.py
@@ -8,8 +8,7 @@ from localstack_snapshot.snapshots.transformer import SortingTransformer
 
 from localstack import config
 from localstack.aws.api.lambda_ import InvocationType, Runtime, State
-from localstack.services.cloudformation.v2.utils import is_v2_engine
-from localstack.testing.aws.util import in_default_partition, is_aws_cloud
+from localstack.testing.aws.util import in_default_partition
 from localstack.testing.pytest import markers
 from localstack.utils.aws.arns import get_partition
 from localstack.utils.common import short_uid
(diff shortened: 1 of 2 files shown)


tests/aws/services/cloudformation/v2/ported_from_v1/resources/test_lambda.py
```diff
@@ -19,10 +18,10 @@
...
 from localstack.utils.testutil import create_lambda_archive, get_lambda_log_events
 
-pytestmark = pytest.mark.skipif(
-    condition=not is_v2_engine() and not is_aws_cloud(),
-    reason="Only targeting the new engine",
-)
+# pytestmark = pytest.mark.skipif(
+#     condition=not is_v2_engine() and not is_aws_cloud(),
+#     reason="Only targeting the new engine",
+# )
```
simonrw: **question**: we need to keep this right? 

```suggestion
pytestmark = pytest.mark.skipif(
    condition=not is_v2_engine() and not is_aws_cloud(),
    reason="Only targeting the new engine",
)
```

simonrw (changes requested):
Thanks for adding this, I think your assessment is correct we need to update `visit_terminal_value_modified` but I _think_ we only need to transform the `after` value. 

Can you try to create a test that exercises this so we can find out for sure? I think it will require modifying the URLs in the lambda test, which should not be too difficult

simonrw:
@pinzon I added some updates namely expanding on the lambda test to cover the update phase, and then needed to update the code in `visit_terminal_value_unchanged` and `visit_terminal_value_modified`

PR #12871, opened 2025-07-17
CFNv2: minor parity assessment and triaging

simonrw:
## Motivation

While triaging the outstanding work for the CFn v2 provider project, I have made some minor changes (typically minor fixes) to gain us more parity with the v1 provider. 

## Changes

* Recursively fetch nested property values (mostly for nested stacks)
* Validate the stack name is provided in `describe_stack_events`
* Don't assume a previous update model is present, for the case of calling `update_stack` after `create_stack` (or another `update_stack`)
* Handle stacks with no changes
* Correctly model stack change based on additional template properties
* Update tests that have been updated in the v1 directory
* Unskip 22 tests and rescope 13 tests
* Explicitly skip the v2 directory in the main integration test step
    * the markers in the individual files are clearly not enough to correctly skip the tests, and running these tests just slows down the main pipeline run

PR #12912, opened 2025-07-25
CFNv2: implement describe-stack-resource

simonrw:
## Motivation

The current V2 CloudFormation provider does not support the describe_stack_resource operation. In order to reach parity with the v1 provider it must do.

## Changes

* Provide typing for the NodeResource in various places
* Rework the way the resolved resources are propagated to provide more type safety and structure
* Add ChangeSetModelValidator for validations we _know_ happen before `create_change_set` returns (i.e. during the modeling phase). This can be a place to hook other validations as well 
* Validate logical resource ids are valid
* Update the StackNotFoundError to handle yet another case, this time through an explicit message override
* Implement DescribeStackResource
* Add tests for describing non-existent things
* Add test validating that stack parameters are given
* Unskip 9 tests

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -18,6 +18,7 @@ from localstack.services.cloudformation.engine.v2.change_set_model import (
     NodeDependsOn,
     NodeOutput,
     NodeParameter,
+    NodeParameters,
     NodeResource,
     is_nothing,
 )
@@ -68,6 +69,10 @@ class ChangeSetModelExecutor(ChangeSetModelPreproc):
         return ChangeSetModelExecutorResult(
             resources=self.resources, parameters=self.resolved_parameters, outputs=self.outputs
         )
+
+    def visit_node_parameters(self, node_parameters: NodeParameters) -> PreprocEntityDelta:
+        delta = super().visit_node_parameters(node_parameters)
+        return delta
 
     def visit_node_parameter(self, node_parameter: NodeParameter) -> PreprocEntityDelta:
         delta = super().visit_node_parameter(node_parameter)
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -2,6 +2,7 @@ import copy
 import logging
 import uuid
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from typing import Final, Optional
 
 from localstack import config
@@ -414,7 +415,6 @@ class ChangeSetModelExecutor(ChangeSetModelPreproc):
                 message=f"Resource type {resource_type} not supported",
             )
 
-        self.resources.setdefault(logical_resource_id, {"Properties": {}})
         match event.status:
             case OperationStatus.SUCCESS:
                 # merge the resources state with the external state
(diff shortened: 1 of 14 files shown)


.github/workflows/tests-pro-integration.yml
```diff
@@ -339,7 +339,7 @@ jobs:
           AWS_DEFAULT_REGION: "us-east-1"
           JUNIT_REPORTS_FILE: "pytest-junit-community-${{ matrix.group }}.xml"
           TEST_PATH: "../../localstack/tests/aws/"  # TODO: run tests in tests/integration
-          PYTEST_ARGS: "${{ env.TINYBIRD_PYTEST_ARGS }}${{ env.TESTSELECTION_PYTEST_ARGS }}--splits ${{ strategy.job-total }} --group ${{ matrix.group }} --durations-path ../../localstack/.test_durations --store-durations"
+          PYTEST_ARGS: "${{ env.TINYBIRD_PYTEST_ARGS }}${{ env.TESTSELECTION_PYTEST_ARGS }}--splits ${{ strategy.job-total }} --group ${{ matrix.group }} --durations-path ../../localstack/.test_durations --store-durations --ignore ../../localstack/tests/aws/services/cloudformation/v2"
```
alexrashed, line 342: FYI: As far as I can tell, this change really just migrates the ignore introduced in https://github.com/localstack/localstack/pull/12871 to the tests executed against Pro.

alexrashed (approved):
Nice to see the tests being unskipped! The implementation, the snapshots, and the tiny pipeline change are looking good! 💯

baermat (approved):
SQS changes are looking good :+1:

tests/aws/services/cloudformation/v2/ported_from_v1/api/test_stacks.py
```diff
@@ -1061,3 +1060,13 @@ def test_stack_resource_not_found(deploy_cfn_template, aws_client, snapshot):
 
     snapshot.add_transformer(snapshot.transform.regex(stack.stack_name, "<stack-name>"))
     snapshot.match("Error", ex.value.response)
+
+
+@markers.aws.validated
```
pinzon, line 1065: **comment**: can't believe we didn't have this validation tested. :sweat_smile: 
Edit: This test should be moved too.

tests/aws/services/cloudformation/v2/ported_from_v1/api/test_resources.py
pinzon, line 1: **Changes**: The tests look great but I'd prefer if we move them to the original CFn test folder and conditionally skip them if the old engine doesn't support them.

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_validator.py
pinzon, line 1: **comment**: Nice :tada:

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -422,18 +422,24 @@ def _execute_resource_action(
...
+                    LogicalResourceId=logical_resource_id,
+                    Type=resource_type,
+                    LastUpdatedTimestamp=datetime.now(timezone.utc),
+                    ResourceStatus=ResourceStatus(f"{status_from_action}_COMPLETE"),
+                    PhysicalResourceId=physical_resource_id,
+                )
+                # TODO: do we actually need this line?
+                resolved_resource.update(extra_resource_properties)
+
+                self.resources[logical_resource_id] = resolved_resource
```
pinzon, line 443: **question/comment**: Does this mean that we only store the resource description when the event is successful? Is that correct? If not, we could move the storage one level above,

pinzon (changes requested):
Looks very good :+1: . The only blocker is the location of the new tests.

tests/aws/services/cloudformation/v2/ported_from_v1/api/test_resources.py
simonrw: At some point in the near future I want to re-merge the tests again so they we don't have divergence. And yes these are tests that don't work with the old engine so would require explicit skipping. Is it ok to move the tests at that point? I might work on this tomorrow if this PR is approved

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
simonrw: You make an excellent point. However we don't have much test coverage of this situation. It should be fairly easy to trigger: set up a resource definition with deliberately invalid data. Then the stack will be left in a bad state, _HOWEVER_ we don't yet model that state and behaviour. I think that's a pre-requisite. I would feel nervous to implement something as fundamental as that without test coverage. I would love it if at the end of this initiative we have correct rollback behaviour but I'm not sure we will yet. 

Perhaps I can work on this in a follow up?

PR #12919, opened 2025-07-28
improve Patch utility: add idempotency

bentsku:
## Motivation
While using the Patch utility, we realize we had to manually check `if patch.is_applied` before applying it again because it did not check itself if it was already applied. 

This PR adds this check internally to avoid applying the patch twice.

It also adds a test for it, and some more patches for some issues we realized and should keep track of. 

I've added a single lock for the `Patch` class as we're mutating the global class variable `applied_patches`, and it feels like once we start adding more checks for multiple patches mutating the same function, this kind of global lock could be useful. 

Linearizing the Patching mechanism over one single lock sounds a bit safer than having a single lock per Patch, and will probably end up using a little bit less memory. 

## Changes
- add an internal check for `Patch.apply` and `Patch.undo` to verify if the patch was already applied, locked with a single lock for the Patch class
- add xfailed tests for the Patch class over multiple patching of the same function

### DIFF
--- a/localstack-core/localstack/utils/patch.py
+++ b/localstack-core/localstack/utils/patch.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+import threading
 import types
 from typing import Any, Callable, List, Type
 
@@ -72,6 +73,7 @@ def create_patch_proxy(target: Callable, new: Callable):
 
 class Patch:
     applied_patches: List["Patch"] = []
+    _lock: threading.RLock = threading.RLock()
     """Bookkeeping for patches that are applied. You can use this to debug patches. For instance,
     you could write something like::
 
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/patch.py
```diff
@@ -72,6 +73,7 @@ def proxy(*args, **kwargs):
 
 class Patch:
     applied_patches: List["Patch"] = []
+    _lock: threading.RLock = threading.RLock()
```
thrau: q: do we really need thread safety in the `Patch` object? is it likely that multiple threads will call `patch.apply()` on the _same_ `patch` object? or is this to guard the `applied_patches` data structure?

bentsku: Good question 😄 I mentioned it in the PR description, but can maybe go a bit further:

> I've added a single lock for the Patch class as we're mutating the global class variable `applied_patches`, and it feels like once we start adding more checks for multiple patches mutating the same function, this kind of global lock could be useful.
>
> Linearizing the Patching mechanism over one single lock sounds a bit safer than having a single lock per Patch, and will probably end up using a little bit less memory.

I was a bit more worried about the case when we start applying different patches to the same function. I also somewhat wanted to protect the `applied_patches` datastructure as the operation is not fully atomic, but to be honest, reading it once more, the most "dangerous" place is maybe `Patch.function` which will call `get_defining_target` which can change when multiple patches are applied. I think it's fine to remove the lock for now, the only calls are `append` and `remove`, and no logic is actually applied to `applied_patches` (we might just end up with duplicate patches, or maybe a `ValueError` if we double call `undo` from 2 different threads, but that seems unlikely). We can think about adding it again once we fix the failing tests and think about how to support multiple patching 👍 thanks for raising this

PR #12920, opened 2025-07-28
CFNv2: validate stack sets test

simonrw:
## Motivation

With https://github.com/localstack/localstack/pull/12909 we introduced stack set support, however the test was not enough and did not actually test the resources in the stack instances were actually created.

I originally didn't think this was possible and required manual setup, however in a classic shower thought I realised I could ensure the "manual" setup could be done via a CFn template.

## Changes

* Validate the stack instance resources are created both via an API snapshot and through service level operations
* Introduce a pre-deploy setup step that ensures the correct administration and execution roles are present


alexrashed:
@simonrw This PR seems a bit outdated, do you want to revive or close it? :)

PR #12923, opened 2025-07-29
fix coveralls upload

alexrashed:
## Motivation
@simonrw just discovered that we are currently not properly uploading our coverage measurements to [Coveralls](https://coveralls.io/github/localstack/localstack).
In the [latest main run](https://github.com/localstack/localstack/actions/runs/16589279437/job/46927136067), we can see that the "Report coverage statistics" step is executed, but there is an error reported (but swallowed):
```
Error running coveralls: Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.
Traceback (most recent call last):
  File "/home/runner/work/localstack/localstack/.venv/lib/python3.11/site-packages/coveralls/cli.py", line 65, in main
    coverallz = Coveralls(
                ^^^^^^^^^^
  File "/home/runner/work/localstack/localstack/.venv/lib/python3.11/site-packages/coveralls/api.py", line 47, in __init__
    self.ensure_token()
  File "/home/runner/work/localstack/localstack/.venv/lib/python3.11/site-packages/coveralls/api.py", line 54, in ensure_token
    raise CoverallsException(
coveralls.exception.CoverallsException: Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.
```
This PR tries to fix this issue by explicitly setting the token on this workflow step.

## Changes
- Explicitly set the `GITHUB_TOKEN` when executing `coveralls` to fix the upload.

## Testing
- [ ] Trigger the execution of the main pipeline on this branch and check whether the coverage data is being uploaded:
  - https://github.com/localstack/localstack/actions/runs/16594368620

PR #12708, opened 2025-06-04
Remove CircleCI pipeline, set GitHub Actions as default

silv-io:
## Motivation

This is the final PR in the initiative to move from CircleCI to GitHub Actions. 
It removes the CircleCI Pipeline and activates pushing of the images from the GitHub Actions pipeline.

## Changes

- Remove CircleCI
- Push Images in GitHub Actions Pipeline

## TODO

What's left to do:

- [ ] ...
- [ ] ...

PR #12924, opened 2025-07-29
avoid coverage reporting when only executing test subset

alexrashed:
## Motivation
With https://github.com/localstack/localstack/pull/12923 we fixed the upload of our coverage data, which didn't work since we moved from CircleCI to GitHub actions (implemented with https://github.com/localstack/localstack/pull/12708).

Unfortunately, we now publish the coverage data also for runs where we only execute a small subset of our tests (i.e. only the acceptance tests).
This PR aims at fixing this issue.

## Changes
- Only execute the reporting step in the main pipeline if `onlyAcceptanceTests` and `enableTestSelection` are `false`, i.e. we also run our integration tests and do not select the tests to execute (but execute all of them).

## Testing
- [x] Check if a partial test run skip the coverage upload.
  - https://github.com/localstack/localstack/actions/runs/16596600205
- [x] Check if a full test run uploads the coverage data.
  - https://github.com/localstack/localstack/actions/runs/16599353262

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -96,6 +96,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - test
+    if: ${{ inputs.onlyAcceptanceTests == false && inputs.enableTestSelection == false }}
     steps:
         - name: Checkout
           uses: actions/checkout@v4


silv-io (approved):
LGTM if the action confirms it! Thank you for the fix :)

PR #12925, opened 2025-07-29
pre-commit: validate json files on commit

simonrw:
## Motivation

Sometimes I have to update snapshot files manually. For example when moving tests from one file to another. I regularly mess up the syntax so the tests fail once CI has run.

We have a series of pre-commit hooks that are run before code is committed. As part of the "push left" initiative (of which LocalStack itself is part!) we should be able to catch these checks before committing code, and _definitely_ before pushing code up to GitHub.

## Changes

- Add the `check-json` hook in our pre-commit configuration file

-->

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,7 @@ repos:
     hooks:
       - id: end-of-file-fixer
       - id: trailing-whitespace
+      - id: check-json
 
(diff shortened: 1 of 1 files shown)


alexrashed (approved):
I think this is a nice convenience feature which should not have any impact on performance (since pre-commit only checks files in the commit). 💯

PR #12927, opened 2025-07-29
Ec2/fix/vpc endpoint dns entries

cloutierMat:
## Motivation

This pr improves on the dns entries returned by moto so they both follow the structure of AWS DNS entries and include localstack hostname for services to be able to register the endpoint.

In the future, if we implement networking in ec2, those endpoints should instead be routed from ec2 to the appropriate service, but since this isn't implemented, for the time being, letting the individual services do their own routing is favorable.

## Changes

- flip the service name around from `com.amazonws.<region>.<service>` to `<service>.<region>.amazonaws.com`, replacing `amazonaws.com` with LocalStack host and port.
- add test for the structure

### DIFF
--- a/localstack-core/localstack/testing/pytest/fixtures.py
+++ b/localstack-core/localstack/testing/pytest/fixtures.py
@@ -6,7 +6,7 @@ import os
 import re
 import textwrap
 import time
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Unpack
 
 import botocore.auth
 import botocore.config
@@ -21,7 +21,7 @@ from pytest_httpserver import HTTPServer
 from werkzeug import Request, Response
 
 from localstack import config
-from localstack.aws.api.ec2 import CreateSecurityGroupRequest
+from localstack.aws.api.ec2 import CreateSecurityGroupRequest, CreateVpcEndpointRequest, VpcEndpoint
 from localstack.aws.connect import ServiceLevelClientFactory
 from localstack.services.stores import (
     AccountRegionBundle,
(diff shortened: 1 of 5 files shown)


localstack-core/localstack/testing/pytest/fixtures.py
```diff
@@ -2046,6 +2046,45 @@ def factory(ports=None, ip_protocol: str = "tcp", **kwargs):
...
+            return _endpoint_details
+
+        return retry(_check_available, retries=30, sleep=5 if is_aws_cloud() else 1)
+
+    yield _create
+
+    try:
+        aws_client.ec2.delete_vpc_endpoints(VpcEndpointIds=vpc_endpoints)
+    except Exception as e:
+        LOG.error("Error cleaning up VPC endpoint: %s, %s", vpc_endpoints, e)
```
viren-nadkarni, line 2072: You could also consider logging `unsuccessful` endpoints from the response of `DeleteVpcEndpoints` operation

tests/aws/services/ec2/test_ec2.py
```diff
@@ -411,6 +413,51 @@ def test_describe_vpc_endpoints_with_filter(self, aws_client, region_name):
...
+            SubnetIds=[subnet_id],
+        )
+
+        # LS only returns one dns entry
+        vpc_endpoint["DnsEntries"] = vpc_endpoint["DnsEntries"][:1]
+        snapshot.match(
+            "dns-suffix", vpc_endpoint["DnsEntries"][0]["DnsName"].split(".")[0].split("-")[-1]
+        )
+        snapshot.match("network-interface-id", vpc_endpoint["NetworkInterfaceIds"][0])
+        snapshot.match("available-endpoint", vpc_endpoint)
```
viren-nadkarni: Isn't it redundant to snapshot match `dns-suffix` and `network-interface-id` separately if the entire response is also matched anyway?

viren-nadkarni (approved):
Looks good, thanks for implementing this :100:

localstack-core/localstack/testing/pytest/fixtures.py
cloutierMat: I found it super useful against aws to have the logs here. As those endpoints are pricey if they don't clean up properly! :rofl:

tests/aws/services/ec2/test_ec2.py
cloutierMat: For `network-interface-id`, I was running into issues transforming list items. I could arguably use a `json_path` transformers. :thinking: Actually I will do that, it will be a bit cleaner.

For `dns-suffix` it allows matching the full shape of the returned dns name, `"<vpc-endpoint-id:1>-<dns-suffix:1>.execute-api.<region>.vpce.<host:1>"`, as just transforming the whole `DnsName` we would lose testing of what this pr introduced.

PR #6444, opened 2022-07-13
Rework RegionBackend

viren-nadkarni:
### Summary

This PR introduces Stores, which supercede RegionBackends.

Stores provide in-memory storage for all provider data. They are analogous to the LocalStack's RegionBackend or Moto's BackendDict.

The terminology used, *Store*, is to avoid confusion with *Backend* which is used by Moto. The usage pattern is similar to Moto's where a singleton tracks all the stores for a service (see example below with `sqs_stores`)

This rework is influenced by Moto v4's approach of using nested dicts to delimit data by AWS account IDs and regions. The usage pattern is inspired by SQLAlchemy and its use of descriptor protocols.

#### Problem with RegionBackend

RegionBackend relies on [voodoo magic](https://github.com/localstack/localstack/blob/c58ae33/localstack/services/generic_proxy.py#L372) to support multi-accounts. It is complex and breaks serialisation, causing compatibiltiy issues with persistence and cloud pods.

#### Usage pattern

```py
# localstack.services.sqs.models
class SqsStore(BaseStore):
    queues = LocalAttribute(default=dict)  # Dict[str, str]
    DELETED = CrossRegionAttribute(default=dict)  # type: Dict[str, float]

sqs_stores = AccountRegionBundle('sqs', SqsStore)

# Usage
sqs_stores['111122223333']  # -> RegionStore
sqs_stores['11112222aaaa']  # -> ValueError! invalid Account ID
sqs_stores['111122223333']['ap-south-1']  # -> SqsStore
sqs_stores['111122223333']['zz-north-1']  # -> ValueError! invalid region
```

#### Followup in separate PRs
- RegionBackends for all providers will be migrated to Stores
- Compatibility with persistence and cloudpods
- Remove RegionBackend code
- Remove feature flag and make multi-accounts default (depends on https://github.com/localstack/localstack/pull/6430)

PR #12931, opened 2025-07-30
Add schemas to stores

giograno:
## Motivation
This PR introduces a heuristic to extract a schema representation for a given store class.

### Context
Service providers implemented in LocalStack store their data (if they are not `moto` based or rely on 3rd party services) in `stores` (see https://github.com/localstack/localstack/pull/6444 for more details). 

LocalStack serializes these classes to implement persistence. Therefore, we should treat them as public APIs and avoid breaking them. As stores evolve within the LocalStack codebase, a few issues can occur when deserializing a previous state:
- A class gets renamed, and we are not able to load that module and class anymore;
- One attribute in the store is removed. Therefore, the information stored there is lost.

Unfortunately, we don't have much visibility on how the structure of the stores changes over time.
For this reason, this PR introduces a heuristic to extract a schema representation from a store definition.
It allows us to reason about store changes. For instance:
- If a type can't be loaded anymore, we can compare an old and a new schema version to understand which new type should be used instead.
- By comparing the schema across two versions, we can detect if an attribute has been removed and come up with a migration path.

## Changes
- Introducing a new `localstack.state.schema` module that returns a schema definition from a `BaseStore` subclass. To achieve this goal, we heavily rely on the type hints of the store attributes. The docstring of the `StoreSchemaBuilder` class reports a few examples.
- A few unit tests.

Closes PNX-46.

### DIFF
--- a/localstack-core/localstack/state/schema.py
+++ b/localstack-core/localstack/state/schema.py
@@ -42,13 +42,13 @@ class StoreSchema(typing.TypedDict):
     attributes: AttributeSchema
 
 
-def get_fully_qualified_name(obj: type) -> str:
+def get_fully_qualified_name(obj: type, separator: str = "::") -> str:
     """Get the fully qualified name of a type"""
     try:
         module = getattr(obj, "__module__", None)
         qualname = getattr(obj, "__qualname__", None)
         if module and qualname:
-            return f"{module}.{qualname}"
+            return f"{module}{separator}{qualname}"
         return getattr(obj, "__name__", str(obj))
     except Exception as e:
         LOG.debug("Unable to compute the FQN for '%s': %s", obj, e)
--- a/tests/unit/state/test_schema.py
+++ b/tests/unit/state/test_schema.py
@@ -71,20 +71,20 @@ def test_simple_store():
     build = StoreSchemaBuilder(MyStore)
     schema = build.build_schema()
     assert schema["attributes"]["field1"] == {
-        "LS/TYPE": "builtins.dict",
-        "LS/ARGS": ["builtins.str", "builtins.str"],
+        "LS/TYPE": "builtins::dict",
+        "LS/ARGS": ["builtins::str", "builtins::str"],
     }
     assert schema["attributes"]["field2"] == {
-        "LS/TYPE": "builtins.list",
-        "LS/ARGS": ["builtins.str"],
+        "LS/TYPE": "builtins::list",
+        "LS/ARGS": ["builtins::str"],
     }
     assert schema["attributes"]["field3"] == {
-        "LS/TYPE": "types.UnionType",
-        "LS/ARGS": ["builtins.str", "builtins.int"],
+        "LS/TYPE": "types::UnionType",
+        "LS/ARGS": ["builtins::str", "builtins::int"],
     }
     assert schema["attributes"]["field4"] == {
-        "LS/TYPE": "builtins.tuple",
-        "LS/ARGS": ["builtins.str", "builtins.int"],
+        "LS/TYPE": "builtins::tuple",
+        "LS/ARGS": ["builtins::str", "builtins::int"],
     }
 
 
@@ -95,13 +95,13 @@ def test_nested_attributes():
     build = StoreSchemaBuilder(MyStore)
     schema = build.build_schema()
     assert schema["attributes"]["field1"] == {
-        "LS/TYPE": "builtins.dict",
+        "LS/TYPE": "builtins::dict",
         "LS/ARGS": [
-            "builtins.str",
+            "builtins::str",
             {
-                "LS/TYPE": "builtins.list",
+                "LS/TYPE": "builtins::list",
                 "LS/ARGS": [
-                    {"LS/TYPE": "builtins.dict", "LS/ARGS": ["builtins.str", "builtins.str"]}
+                    {"LS/TYPE": "builtins::dict", "LS/ARGS": ["builtins::str", "builtins::str"]}
                 ],
             },
         ],
@@ -125,6 +125,6 @@ def test_additional_classes(monkeypatch):
     assert schema["attributes"]["move_tasks"]["LS/ARGS"][1] == mmt_fqn
     assert mmt_fqn in schema["additional_classes"]
     assert schema["additional_classes"][mmt_fqn]["source_arn"] == {
-        "LS/TYPE": "types.UnionType",
-        "LS/ARGS": ["builtins.str", "builtins.NoneType"],
+        "LS/TYPE": "types::UnionType",
+        "LS/ARGS": ["builtins::str", "builtins::NoneType"],
     }


localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+    attributes: AttributeSchema
+
+
+def get_fully_qualified_name(obj: type) -> str:
+    """Get the fully qualified name of a type"""
+    try:
+        module = getattr(obj, "__module__", None)
+        qualname = getattr(obj, "__qualname__", None)
+        if module and qualname:
+            return f"{module}.{qualname}"
```
viren-nadkarni: You could use a different separator than `.` incase there is a need to reverse the operation

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+    if not module:
+        return False
+    if any(str(module).startswith(prefix) for prefix in module_prefixes):
+        return True
+    return False
+
+
+class StoreSchemaBuilder:
+    """
+    This class build a schema for a ``BaseStore`` class by recursively parsing its type hints.
```
viren-nadkarni:
```suggestion
    This class builds a schema for a ``BaseStore`` class by recursively parsing its type hints.
```

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+                  "builtins.str",
+                  "builtins.NoneType"
+                ]
+              }
+            }
+          }
+        }
+    """
+
+    skip_attributes = ["_account_id", "_global", "_region_name", "_service_name", "_universal"]
```
viren-nadkarni: Is there a way to make a SSOT for this because it's being duplicated with `localstack.services.stores`?

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+                    _hint[TAG_ARGS] = [self._serialize_hint(args[0])]
+                return _hint
+            case types.UnionType | typing.Union | typing.Tuple | builtins.tuple:
+                _hint = {TAG_TYPE: get_fully_qualified_name(origin)}
+                args = typing.get_args(type_hint)
+                if args:
+                    _hint[TAG_ARGS] = [self._serialize_hint(_arg) for _arg in args]
+                return _hint
+            case _:
+                # A few things that can end up here: generics, Union, or Literal. See ``get_origin`` for more.
```
viren-nadkarni: `Union` appears to have been addressed in the previous case

```suggestion
                # A few things that can end up here: generics, or Literal. See ``get_origin`` for more.
```

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+            if is_internal_class(type_hint):
+                # We add in advance the FQN of this class to the keys of ``additional_classes`` in case ``type_hint``
+                #   fields point recursively to ``type_hint`` itself.
+                self.schema["additional_classes"][fqn] = {}
+                self.schema["additional_classes"][fqn] = self._attribute_schema(type_hint, [])
+            return fqn
+
+        match origin:
+            case builtins.dict:
+                _hint = {TAG_TYPE: get_fully_qualified_name(origin)}
```
viren-nadkarni: Nitpick: I would suggest avoiding `_` variables if there is no conflict in the current namespace

tests/unit/state/test_schema.py
```diff
@@ -0,0 +1,130 @@
...
+from localstack.services.sts.models import STSStore
+from localstack.services.transcribe.models import TranscribeStore
+from localstack.state import schema as ls_schema
+from localstack.state.schema import StoreSchemaBuilder, get_fully_qualified_name
+
+
+def test_smoke_schema_dumps():
+    """
+    This test extracts the schema for all the stores in the repo and dumps it.
+    It's just a smoke test to make sure
```
viren-nadkarni: Incomplete comment?

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+                if args:
+                    _hint[TAG_ARGS] = [self._serialize_hint(args[0])]
+                return _hint
+            case types.UnionType | typing.Union | typing.Tuple | builtins.tuple:
+                _hint = {TAG_TYPE: get_fully_qualified_name(origin)}
+                args = typing.get_args(type_hint)
+                if args:
+                    _hint[TAG_ARGS] = [self._serialize_hint(_arg) for _arg in args]
+                return _hint
+            case _:
```
viren-nadkarni, line 234: Any thoughts how cases where stores have member functions e.g. `SqsStore.expire_deleted()` be handled? How could such callables be represented in the schema as changes within them can also affect store compatibility?

localstack-core/localstack/state/schema.py
```diff
@@ -0,0 +1,229 @@
...
+
+AttributeSchema = dict[AttributeName, SerializedHint]
+"""Maps an attribute name its serialized hints"""
+
+AdditionalClasses = dict[FQN, AttributeSchema]
+"""Maps the a FQN of a class to its Attribute Schema"""
+
+TAG_TYPE = "LS/TYPE"
+TAG_ARGS = "LS/ARGS"
+"""Tags for subscribed types and their args. See ``StoreSchemaBuilder`` for examples."""
```
viren-nadkarni: I think this can be simplified. Can `TAG_TYPE` and `TAG_ARGS` be statically defined instead of constants? This will also allow defining `SerializedHint` recursively.

viren-nadkarni (commented):
This is a good foundation to start with. This approach will build the schema out of how stores are declared, but unfortunately not what goes in them. Arguably that's a quality of weakly typed languages. One could say an attribute is of a certain type, but keep a different type in there, and this cannot be detected without running an exhaustive static type check.

Overall this tackles one aspect of schema comparison -- the store side. The other aspect is state side. I imagine each persisted state will have its own derivable schema which describes the actual data in it. This could then be checked against the store schema for load compatibility.

I'm curious to hear @thrau's thoughts.

localstack-core/localstack/state/schema.py
giograno: Fixed in https://github.com/localstack/localstack/pull/12931/commits/21a884cd6383a2309c3527a2c19191d8dbb3e151

tests/unit/state/test_schema.py
giograno: Fixed in https://github.com/localstack/localstack/pull/12931/commits/21a884cd6383a2309c3527a2c19191d8dbb3e151

localstack-core/localstack/state/schema.py
giograno: Nice suggestion! Fixed in https://github.com/localstack/localstack/pull/12931/commits/f9fb24bc170c42329790129b567594c8b1eb40fc by fetching the `BaseStore` annotations.

localstack-core/localstack/state/schema.py
giograno: Nice observation. My idea was to not include these functions in the schema definition, as they won't be serialized by a JSON schema backend. Our goal with the project is indeed to serialize only data and not code anymore.

localstack-core/localstack/state/schema.py
giograno: Using `::` not (see https://github.com/localstack/localstack/pull/12931/commits/607c522bf8fd3395233e06f0a822a08aa9cdedaa)

localstack-core/localstack/state/schema.py
giograno: I am not sure I fully understood the suggestion. In https://github.com/localstack/localstack/pull/12931/commits/7e0ba598f664fd06847ff122c1954270af70256c I made the `SerializedHint` recursive and switched to a `StrEnum` for the tags.

thrau (commented):
As discussed in Today's meeting with Gio and Viren, we've come to the agreement that the direction of generating a schema from stores is the correct approach, but using some existing standard (like avro, protobuf, ...) for the schema definition would be better, so we can leverage the ecosystem around the tools.

giograno:
Good learning here. Closing for further investigation with the other standards mentioned above.

PR #12932, opened 2025-07-30
KMS: return `NotImplementedError` for rotation of imported keys

sannya-singal:
## Motivation
After the launch of [this feature](https://aws.amazon.com/blogs/security/how-to-use-on-demand-rotation-for-aws-kms-imported-keys/), AWS introduced support for on demand rotation of imported keys i.e. keys with `Origin` set to `EXTERNAL`. 

At the moment, LocalStack does not support this feature and we return a `400` error with `UnsupportedOperationException` even though the request itself is valid.

## Changes
Raise appropriate error response with error code `501` when rotation of keys is done with imported keys. 

Please note, this PR does not aim to fix existing tests and other logic in the code which revolves around assuming that external keys are unsupported, this should be tackled in a separate PR.

### DIFF
--- a/localstack-core/localstack/services/kms/provider.py
+++ b/localstack-core/localstack/services/kms/provider.py
@@ -1408,9 +1408,7 @@ class KmsProvider(KmsApi, ServiceLifecycleHook):
         if key.metadata["KeySpec"] != KeySpec.SYMMETRIC_DEFAULT:
             raise UnsupportedOperationException()
         if key.metadata["Origin"] == OriginType.EXTERNAL:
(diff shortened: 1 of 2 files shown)


k-a-il (approved):
LGTM 🚀

sannya-singal:
The pipeline failed: https://github.com/localstack/localstack/actions/runs/16620811305/job/47024507676?pr=12932 because `test_rotate_key_on_demand_raises_error_given_key_with_imported_key_material` should be fixed/removed with the introduction of the new feature by AWS. I have skipped it for now.

PR #12934, opened 2025-07-30
CFnv2: implement SSM parameter resolving

simonrw:
## Motivation

The CloudFormation v2 engine does not yet have the ability to resolve external SSM parameters wheh used in the `Parameters` position. 

## Changes

* Introduce the concept of an `EngineParameter` which tracks the default, given and resolved values for the parameter
* Resolve SSM parameters in the provider since it blocks change set creation on failure
* Remove custom handling in the executor as this is too late in the deployment process
* Hoist the fetching of the resource type for events up a level where we have access to the value rather than trying to read it from the template
* Make `_process_event` require keyword arguments
* Start to migrate away from `stack.describe_details` and put the logic into the provider where it belongs
* Add validation that parameters are fulfilled i.e. either provided, from defaults or external
* Don't set the stack template until change set execution has finished, in parity with AWS
* Set the template on the change set object in parity with AWS
* Add testing for resolivng SSM parameters and changing the value between stack deploys
* Add testing for resolving SSM parameters that don't exist
* Unskip `TestImports` as these now pass with #12906

## TODO (follow ups)

* Add additional "external" types, e.g. `AWS::SSM::Parameter::Value<StringList>`


localstack-core/localstack/services/cloudformation/v2/entities.py
```diff
@@ -200,7 +216,7 @@ def describe_details(self) -> ApiStack:
             "Tags": [],
             "NotificationARNs": [],
             "Capabilities": self.capabilities,
-            "Parameters": self.parameters,
+            # "Parameters": self.resolved_parameters,
```
pinzon:
```suggestion
```

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -679,11 +744,68 @@ def describe_stacks(
...
-            raise StackNotFoundError(stack_name)
-        # TODO: move describe_details method to provider
-        return DescribeStacksOutput(Stacks=[stack.describe_details()])
+        if stack_name:
+            stack = find_stack_v2(state, stack_name)
+            if not stack:
+                raise StackNotFoundError(stack_name)
+            stacks = [stack]
+        else:
+            raise RuntimeError("TODO")
```
pinzon: So it's either try to find one or return all. [See here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/describe_stacks.html)

```suggestion
            stacks = state.stacks_v2.values()
```

simonrw: That's great, thanks!

Issue #12889, opened 2025-07-22
bug: Unable to write to HTTP response from Golang Lambda with Function URL


sno6:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have deployed a Golang Lambda that is set up to be a streaming function URL. When hitting the deployed URL my code will complete, but the client will receive the following error instead of the data the URL wrote:

```
{"__type": "InternalError", "message": "exception while calling lambda with unknown operation: Extra data: line 1 column 222 (char 221)"}
```

FWIW I have similar code running in production on AWS Lambda so I strongly believe the problem is on Localstack side.

I have a minimal reprod repo for y'all here: https://github.com/sno6/localstack-lambdaurl-issue

### Expected Behavior

I expected `cURL` to receive the data that the lambda wrote to the response writer. Also, I expect that using the -N flag would result in the response streaming in to the client.

### How are you starting LocalStack?

With the `localstack` script

### Steps To Reproduce

1. Clone https://github.com/sno6/localstack-lambdaurl-issue
2. Run `make build`
3. Deploy the stack using cdklocal.
4. `curl` the deployed function url.

### Environment

```markdown
- OS: Mac OSX
- LocalStack:
  LocalStack version: LocalStack CLI 4.2.0
```

pinzon:
Thanks for reporting this issue. It's been some months since the solution has been merged.

PR #12935, opened 2025-07-30
support ResponseStream mode for functionUrl

pinzon:
## Motivation
This Pr implements the capability to handle LambdaURL invocations of type StreamResponse [See here for more info](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html)

## Changes
- AWS::Lambda::Url  accepts the InvokeMode parameter to create a LambdaUrl
- UrlRouter can handle results with streaming separator
- UrlRouter can handle results with simple strings

## Testing
- New AWS validated test


pinzon:
Closes: #12889

PR #12937, opened 2025-07-31
Apigw/enable vpce routing

cloutierMat:
## Motivation

depends on: #12927 

This pr adds a route to apigw to allow invoking an ApiGateway API through it's vpc endpoint using the `x-apigw-api-id` headers, as describe in [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-api-test-invoke-url.html#apigateway-private-api-public-dns).

At the moment we are not validating the vpc endpoint exists, so any "valid" VPC endpoint like url targeted at `.vpce.execute-api` will attempt to execute the api provided with the `x-apigw-api-id`. Note this is already the case for the other path style of VPC endpoint and as such falls outside the scope of the current PR.

## Changes

- expand testing to include targeting with `host` headers (already supported by default, but now tested)
- expand testing to include targeting with host headers
- register routes

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -145,7 +145,7 @@ class ApiGatewayEndpoint:
         )
         return not_found
 
-    def vpe_endpoint_handler(
+    def vpc_endpoint_handler(
         self, request: Request, **kwargs: Unpack[RouteHostPathParameters]
     ) -> Response:
         # TODO validate the vpc endpoint exists in the account/region before routing
@@ -228,21 +228,21 @@ class ApiGatewayRouter:
             self.router.add(
                 path="/",
                 host=vpce_host_pattern,
-                endpoint=self.handler.vpe_endpoint_handler,
+                endpoint=self.handler.vpc_endpoint_handler,
                 defaults={"path": "", "stage": None},
                 strict_slashes=True,
             ),
             self.router.add(
                 path="/<stage>/",
                 host=vpce_host_pattern,
-                endpoint=self.handler.vpe_endpoint_handler,
+                endpoint=self.handler.vpc_endpoint_handler,
                 defaults={"path": ""},
                 strict_slashes=False,
             ),
             self.router.add(
                 path="/<stage>/<greedy_path:path>",
                 host=vpce_host_pattern,
-                endpoint=self.handler.vpe_endpoint_handler,
+                endpoint=self.handler.vpc_endpoint_handler,
                 strict_slashes=True,
             ),
         ]
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
@@ -33,6 +33,7 @@ class RouteHostPathParameters(TypedDict, total=False):
     server: str | None
     stage: str | None
     vpce_suffix: str | None
+    vpce_dns: str | None
 
 
 class ApiGatewayEndpoint:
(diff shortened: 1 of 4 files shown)


tests/aws/services/apigateway/test_apigateway_integrations.validation.json
```diff
@@ -12,7 +12,13 @@
     "last_validated_date": "2024-12-11T15:28:54+00:00"
   },
   "tests/aws/services/apigateway/test_apigateway_integrations.py::test_create_execute_api_vpc_endpoint": {
-    "last_validated_date": "2024-04-15T23:07:07+00:00"
+    "last_validated_date": "2025-07-30T17:57:02+00:00",
+    "durations_in_seconds": {
+      "setup": 12.89,
+      "call": 1064.8,
```
bentsku, line 18: painful 😭

tests/aws/services/apigateway/test_apigateway_integrations.py
```diff
@@ -1064,14 +1069,61 @@ def handler(event, context):
...
+    # LocalStack
+    #   url: http://localhost.localstack.cloud:4566/{stage}
+    #   host: {rest-api-id}-{vpce-id}.execute-api.localhost.localstack.cloud
+    retry(lambda: _invoke_api(endpoint, {"host": host_header}), retries=15, sleep=poll_sleep)
+
+    # invoke Lambda and assert result
+    # AWS
+    #   url: https://{public-dns-hostname}.execute-api.{region}.vpce.amazonaws.com/{stage}
+    #   x-apigw-api-id: {rest-api-id}
+    # LocalStack Not a clue
```
bentsku: why "Not a clue"? 😄 same in the comment above

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -144,6 +145,13 @@ def create_not_found_response(api_id: str) -> Response:
         )
         return not_found
 
+    def vpc_endpoint_handler(
```
bentsku, line 148: I think this is a good way forward, as the shape of the URL is very different and doesn't have some of the parameters we're always expecting in a route, like `api_id`. Thanks for implementing it this way 🚀 and minimal changes with upstream dependencies 👍

localstack-core/localstack/services/apigateway/next_gen/execute_api/router.py
```diff
@@ -214,6 +225,26 @@ def register_routes(self) -> None:
                 endpoint=self.handler,
                 strict_slashes=True,
             ),
+            self.router.add(
```
bentsku, line 228: we have so many routes 😅

tests/aws/services/apigateway/test_apigateway_integrations.py
```diff
@@ -1012,21 +1023,15 @@ def _check_available():
...
     def handler(event, context):
         import requests
-        headers = {{"content-type": "application/json", "host": "{host_header}"}}
-        result = requests.post("{endpoint}", headers=headers)
-        return {{"content": result.content.decode("utf-8"), "code": result.status_code}}
+        url = event["url"]
+        headers = event["headers"]
+
+        result = requests.post(url, headers=headers)
+        return {"content": result.content.decode("utf-8"), "code": result.status_code}
```
bentsku, line 1034: thanks for making the test more modular! this is really nice

tests/aws/services/apigateway/test_apigateway_integrations.py
cloutierMat: oops! Placeholder not removed :stuck_out_tongue_closed_eyes:

cloutierMat:
> I suppose it is only because it is a stacked PR

:100: Thank you for the early review! :rocket:

PR #12938, opened 2025-07-31
Allow stack traces in HTTP responses only if `INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE` is enabled

sannya-singal:
## Motivation
As a part of the initiative for improving Localstack error messages, this PR aims to conditionally remove stack traces from HTTP responses. A new environment variable `INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE` is introduced, which when enabled, allows stack traces to be included in the HTTP response. 
This is expected to be always enabled during test runs, where it improves LocalStack developers devX, unless its explicitly disabled. This flag is for internal use only. The stack traces wouldn't appear in the HTTP responses for users unlike before when `DEBUG` is enabled. 

**Current Behaviour**: 

```
sannyasingal@Sannyas-MacBook-Pro ~ % awslocal organizations describe-organization

An error occurred (InternalError) when calling the DescribeOrganization operation (reached max retries: 4): exception while calling organizations.DescribeOrganization: Traceback (most recent call last):
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 112, in __call__
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/handlers/service.py", line 82, in __call__
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 154, in invoke
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 168, in dispatch_request
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/forwarder.py", line 133, in _call
    return handler(context, req)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/aws/skeleton.py", line 118, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/code/localstack/.venv/lib/python3.11/site-packages/localstack/pro/core/services/organizations/provider.py", line 304, in describe_organization
    a = org[0]
        ~~~^^^
TypeError: 'NoneType' object is not subscriptable

```

**Expected Behaviour**: 

```
sannyasingal@Sannyas-MacBook-Pro ~ % awslocal organizations describe-organization

An error occurred (InternalError) when calling the DescribeOrganization operation (reached max retries: 4): exception while calling organizations.DescribeOrganization: 'NoneType' object is not subscriptable

```

## Changes
This PR excludes stack traces from HTTP responses when `DEBUG=1` mode is enabled (for `DEBUG=0` we already do not log stack trace in the response) and rather allow stack traces when `INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE` is enabled. 
Related PR in pro https://github.com/localstack/localstack-pro/pull/5003

## Testing

The responses to:
- local test runs by enabling and disabling `INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE` have been tested locally by deliberately introducing an error by assigning `a = request["KeyId"][100]` in `describe_key` handler and run `awslocal kms describe-key --key-id=411cff4c-e1d0-49cc-9ec7-73ac91049d86` command ensuring expected stack trace behaviour for both cases. 
- CI run with a failing operation https://github.com/localstack/localstack/actions/runs/16820674492/job/47646766927?pr=12938 ensuring expected stack trace behaviour when enabled. 

## TODO

- [x] test CI run with a failing operation

### DIFF
--- a/localstack-core/localstack/aws/handlers/service.py
+++ b/localstack-core/localstack/aws/handlers/service.py
@@ -1,7 +1,6 @@
 """A set of common handlers to parse and route AWS service requests."""
 
 import logging
-import traceback
 from collections import defaultdict
 from typing import Any
 
@@ -183,13 +182,6 @@ class ServiceExceptionSerializer(ExceptionHandler):
         elif not isinstance(exception, ServiceException):
             if not self.handle_internal_failures:
                 return
-
-            if config.DEBUG:
-                exception = "".join(
-                    traceback.format_exception(
-                        type(exception), value=exception, tb=exception.__traceback__
-                    )
-                )
 
             # wrap exception for serialization
             if operation:
--- a/localstack-core/localstack/aws/handlers/service.py
+++ b/localstack-core/localstack/aws/handlers/service.py
@@ -1,14 +1,13 @@
 """A set of common handlers to parse and route AWS service requests."""
 
 import logging
-import os
 import traceback
 from collections import defaultdict
 from typing import Any
 
 from botocore.model import OperationModel, ServiceModel
 
-from localstack import config, constants
+from localstack import config
 from localstack.http import Response
 from localstack.utils.coverage_docs import get_coverage_link_for_service
 
@@ -185,10 +184,7 @@ class ServiceExceptionSerializer(ExceptionHandler):
             if not self.handle_internal_failures:
                 return
 
-            if (
-                os.environ.get(constants.INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE)
-                in constants.TRUE_STRINGS
-            ):
+            if config.INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE:
                 exception = "".join(
                     traceback.format_exception(
                         type(exception), value=exception, tb=exception.__traceback__
--- a/localstack-core/localstack/aws/handlers/service.py
+++ b/localstack-core/localstack/aws/handlers/service.py
@@ -1,12 +1,14 @@
 """A set of common handlers to parse and route AWS service requests."""
 
 import logging
+import os
+import traceback
 from collections import defaultdict
 from typing import Any
 
 from botocore.model import OperationModel, ServiceModel
 
-from localstack import config
+from localstack import config, constants
 from localstack.http import Response
 from localstack.utils.coverage_docs import get_coverage_link_for_service
 
@@ -182,6 +184,16 @@ class ServiceExceptionSerializer(ExceptionHandler):
         elif not isinstance(exception, ServiceException):
             if not self.handle_internal_failures:
                 return
+
+            if (
+                os.environ.get(constants.INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE)
+                in constants.TRUE_STRINGS
+            ):
+                exception = "".join(
+                    traceback.format_exception(
+                        type(exception), value=exception, tb=exception.__traceback__
+                    )
+                )
 
             # wrap exception for serialization
             if operation:
--- a/localstack-core/localstack/testing/pytest/in_memory_localstack.py
+++ b/localstack-core/localstack/testing/pytest/in_memory_localstack.py
@@ -71,7 +71,7 @@ def pytest_runtestloop(session: Session):
 
     # configure
     os.environ[ENV_INTERNAL_TEST_RUN] = "1"
-    os.environ[INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE] = "1"
+    os.environ[INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE] = "0"
 
     safe_requests.verify_ssl = False
 
--- a/localstack-core/localstack/testing/pytest/in_memory_localstack.py
+++ b/localstack-core/localstack/testing/pytest/in_memory_localstack.py
@@ -71,7 +71,7 @@ def pytest_runtestloop(session: Session):
 
     # configure
     os.environ[ENV_INTERNAL_TEST_RUN] = "1"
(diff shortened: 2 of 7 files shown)


thrau (commented):
general question: if we remove this, how does the devx look like for localstack developers? the stacktrace is added when `DEBUG=1` is set so the stack traces appear in test failures directly, which has helped (at least me) in the past to find issues very quickly.

bentsku:
> general question: if we remove this, how does the devx look like for localstack developers? the stacktrace is added when DEBUG=1 is set so the stack traces appear in test failures directly, which has helped (at least me) in the past to find issues very quickly.

That's a good point: it returns the error directly in the client call, which allows the test annotation to show the error directly and not have to go through the logs of the running LocalStack instance. 

I think there's a fine line with the customer running with `DEBUG=1` and what we show. But returning stack traces in error is IMHO a behavior change that is a bit more intrusing that just the debug level.
Maybe this should be a different flag, and not only rely on the logging level, but with another environment variable that would be set to true while running tests? Something that we could directly automatically enable via a fixture? That way, we'd keep the internal DevX while still prevent customers to see those stack traces.

thrau:
that's fair. so perhaps instead of removing the behavior (returning stack traces) all together, we could tie it to our tests being run. for example `os.environ[constants.ENV_INTERNAL_TEST_RUN]`, will be set for integration tests and we could use that instead of `config.DEBUG`

localstack-core/localstack/aws/handlers/service.py
```diff
@@ -184,7 +185,10 @@ def create_exception_response(self, exception: Exception, context: RequestContex
             if not self.handle_internal_failures:
                 return
 
-            if config.DEBUG:
+            if (
+                os.environ.get(constants.INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE)
+                in constants.TRUE_STRINGS
+            ):
```
bentsku: question: was there a specify reason that would prevent us from adding `INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE` to config like other configuration variables? It would more easily discoverable, and we could still patch it in the fixture with `monkeypatch.setattr(config, "INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE", True)`

But maybe I am missing something?

bentsku (commented):
I have a question regarding adding the flag to our `config.py` file instead of directly manipulating environment variables, not sure if I miss some context. Maybe it is because it is internal only?

For the other "test only" variables like `LOCALSTACK_INTERNAL_TEST_RUN` and `LOCALSTACK_INTERNAL_TEST_COLLECT_METRIC`, they don't change LocalStack behavior, only the tests behavior, which is maybe why they are not added to `config.py`. 

Here we're directly manipulating LocalStack behavior, so I think we might be better if it is a proper config flag, what do you think? 

As a side-note, it could also allow us to keep the current behavior until the next release (where we would only flip the flag) to not modify behavior suddenly, but to be honest I'm not sure it is worth it, I don't see what it would bring to our users, so I think we can change it in `latest` 😄

k-a-il (approved):
LGTM 👍

tests/conftest.py
```diff
@@ -87,3 +89,13 @@ def secondary_aws_client(secondary_aws_client_factory):
...
+
+
+@pytest.fixture(autouse=True)
+def enable_stack_trace_for_tests(monkeypatch):
+    """
+    Ensure stack traces are enabled in HTTP responses during test sessions.
+
+    This is useful for debugging purposes.
+    """
+    monkeypatch.setenv(INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE, "1")
```
bentsku: nit: we could create a `session` scope fixture too with `MonkeyPatch` directly: https://github.com/pytest-dev/pytest/issues/1872#issuecomment-375108891

See here where we do it for a `class` fixture:
https://github.com/localstack/localstack/blob/4464cd24b7691f632e7c8a93175d379a852d33c7/tests/aws/scenario/bookstore/test_bookstore.py#L50-L60

localstack-core/localstack/aws/handlers/service.py
sannya-singal: Thanks for raising this 🙌  There was no strong reason not to add it to `config.py`, the flag was introduced mainly for internal use so took this approach. But I agree that making it a proper config flag would improve discoverability also since this flag affects the LocalStack behaviour. I will go ahead and move it to `config.py` with a default of `0`, and patch it to `1` in the fixture and also make relevant changes in pro 👍

localstack-core/localstack/testing/pytest/in_memory_localstack.py
```diff
@@ -71,6 +71,8 @@ def pytest_runtestloop(session: Session):
 
     # configure
     os.environ[ENV_INTERNAL_TEST_RUN] = "1"
+    localstack_config.INCLUDE_STACK_TRACES_IN_HTTP_RESPONSE = True
```
bentsku, line 74: question: do we need this if we have the fixture? I guess it cannot hurt, so I'm fine leaving it like that 👍

sannya-singal: Right, the fixture should be enough, I just added it here as a safeguard. But happy to drop it if we want to clean it up in future.

PR #12939, opened 2025-07-31
fix partial coverage reporting

alexrashed:
## Motivation
With #12923 we fixed the coverage upload, and with #12924 we tried to fix the issue of reporting partial results.
Unfortunately, debugging job level conditionals in GitHub actions is notoriously hard / [there isn't really a way to get proper debug information](https://github.com/orgs/community/discussions/20640) on how the conditions are evaluated.

## Changes
- Fixes the conditional for uploading the coverage report by switching to "not explicitly true" for the test selection and aceptancfe tests and considering the push event.
- Fixes a conditional to forward the `enableTestSelection` correctly on the workflow call.

## Testing
- [x] [Test full run is reported](https://github.com/localstack/localstack/actions/runs/16645199459)
- [x] [Test run triggered by a push event](https://github.com/localstack/localstack/actions/runs/16647895400)
- [x] [Test only acceptance test run is not reported](https://github.com/localstack/localstack/actions/runs/16645189125)
- [x] [Test run with enabled test selection is not reported](https://github.com/localstack/localstack/actions/runs/16645179453)

### DIFF
--- a/.github/workflows/aws-main.yml
+++ b/.github/workflows/aws-main.yml
@@ -4,33 +4,6 @@ on:
   schedule:
     - cron: 0 2 * * MON-FRI
   push:
-    paths:
-      - '**'
-      - '!.github/**'
-      - '.github/actions/**'
-      - '.github/workflows/aws-main.yml'
-      - '.github/workflows/aws-tests.yml'
-      - '!CODEOWNERS'
-      - '!README.md'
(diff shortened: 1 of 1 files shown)


silv-io (approved):
LGTM! Conditionals in GHA are indeed very painful

Issue #12881, opened 2025-07-21
bug: Fifo Queue SendMessage DelaySeconds


bly2k:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have a fifo queue named "Test-delay.fifo" where DelaySeconds=60.

I am using a 3rd party library (.NET) that puts messages into this queue. I cannot modify the library code but I ran into a problem, debugged it, and noticed this behavior. The library calls SendMessage passing in a message with DelaySeconds = 0. On Localstack, it appears that even though the queue DelaySeconds = 60, it appears the SendMessage DelaySeconds = 0 overrides it and makes the message immediately visible. 

This does not happen on AWS SQS. So above code on AWS SQS just ignores the SendMessage DelaySeconds parameter (0) and uses the queue DelaySeconds (60).

I did some testing and the issue occurs for both SendMessage and SendMessageBatch.

### Expected Behavior

SendMessage API call with DelaySeconds parameter to fifo queues probably should ignore the DelaySeconds parameter? This is what I am seeing when going to AWS SQS.

### How are you starting LocalStack?

With a docker-compose file

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

    docker compose up

Reproduce (C#):

            //first create fifo queue Test-delay.fifo, set DelaySeconds = 60

            var sendRequest = new SendMessageRequest
            {
                QueueUrl = "http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/Test-delay.fifo",  // this is a fifo queue with DelaySeconds = 60
                MessageBody = "{ \"Message\": \"Hello\" }",
                MessageGroupId = "TestGroup",
                MessageDeduplicationId = Guid.NewGuid().ToString(),
                MessageAttributes = new Dictionary<string, MessageAttributeValue>
                {
                },
                DelaySeconds = 0  //causes immediate visibility in LocalStack but not in AWS SQS
            };

            await sqsClient.SendMessageAsync(sendRequest);

            //after above call, the message is immediately visible without delay (for example ReceiveMessage API) in LocalStack. This does not happen in AWS SQS.

### Environment

```markdown
- OS: Windows/WSL
- LocalStack:
  LocalStack version: 4.6.1.dev16
  LocalStack Docker image sha:
  LocalStack build date: 2025-07-11
  LocalStack build git hash: aa264061b
```

kn1ves:
I was able to reproduce this using the following:

[...]

PR #12941, opened 2025-07-31
SQS: fix DelaySeconds behavior for FIFO queues on message level

kn1ves:
## Motivation
As reported in [#12881](https://github.com/localstack/localstack/issues/12881) when a FIFO SQS queue has a default DelaySeconds value (e.g. 60 seconds), LocalStack honors an explicit SendMessage call with DelaySeconds=0 and makes the message immediately visible.

This behavior differs from AWS SQS, where a DelaySeconds=0 parameter is ignored and the queue’s default delay is applied instead. 

To ensure behavior parity with AWS SQS—and prevent surprises when migrating from AWS to LocalStack—LocalStack should ignore explicit DelaySeconds=0 for FIFO queues and respect the queue’s default settings.

## Changes
- Modified the put method in FifoQueue to always use the queue level delay message setting
- Add test to verify this behavior


kn1ves:
Not letting me add labels myself, if someone with access can please add semver:patch and aws:sqs please

tests/aws/services/sqs/test_sqs.py
```diff
@@ -2119,6 +2119,36 @@ def test_fifo_queue_send_message_with_delay_on_queue_works(
...
+            QueueName=f"queue-{short_uid()}.fifo",
+            Attributes={
+                "FifoQueue": "true",
+                "ContentBasedDeduplication": "true",
+                "DelaySeconds": str(delay_seconds),
+            },
+        )
+
+        send_result = aws_sqs_client.send_message(QueueUrl=queue_url, MessageBody="message-1", MessageGroupId="1", DelaySeconds=0)
+        snapshot.match("send_message_result", send_result)
```
gregfurman: suggestion: As a sanity check, can we add a test that sends a message to SQS with the largest possible delay (900s) and ensure the message arrives after 2s?

tests/aws/services/sqs/test_sqs.py
```diff
@@ -2119,6 +2119,36 @@ def test_fifo_queue_send_message_with_delay_on_queue_works(
...
+
+        response_initial_receive = aws_sqs_client.receive_message(QueueUrl=queue_url, WaitTimeSeconds=1)
+        snapshot.match("receive_message_initial_result", response_initial_receive)
+        assert response_initial_receive.get("Messages", []) == []
```
gregfurman: suggestion(non-blocking): We could also check the SQS message's `ApproximateFirstReceiveTimestamp` and `SentTimestamp` values and assert the delta is longer than (or equal to)  2s (see [MessageSystemAttributeNames](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html#SQS-ReceiveMessage-request-MessageSystemAttributeNames).

Probably a bit extra though so feel free to ignore 🙂

kn1ves:

tests/aws/services/sqs/test_sqs.py
kn1ves: For a fifo queue? If yes, FIFO queues will not let you send a request with anything other than 0 as the delay seconds, you'll get an exception.

gregfurman: > FIFO queues will not let you send a request with anything other than 0 as the delay seconds, you'll get an exception.

If that's the case then we should probably add a test for this, since we want LocalStack to emulate error response behaviour as well.

You can see an example of how we snapshot error responses here:

https://github.com/localstack/localstack/blob/fcffe33c004bac69a339f506c17930b508a24cc7/tests/aws/services/sqs/test_sqs.py#L308-L313

We should probably raise the appropriate exception in the `SqsProvider` that ensures the `DelaySeconds` parameter is always `0` when sending messages to a FIFO queue? This should probably be included in our message checks:

https://github.com/localstack/localstack/blob/82de91e30b0185a6792dd5047168b29d69bbb1f9/localstack-core/localstack/services/sqs/provider.py#L1200-L1217

Would you have capacity for this addition? 🤔

kn1ves: > > FIFO queues will not let you send a request with anything other than 0 as the delay seconds, you'll get an exception.
> 
> If that's the case then we should probably add a test for this, since we want LocalStack to emulate error response behaviour as well.
> 
Hmm, isn't that scenario covered here?
https://github.com/localstack/localstack/blob/fcffe33c004bac69a339f506c17930b508a24cc7/tests/aws/services/sqs/test_sqs.py#L2074-L2089
> 
> We should probably raise the appropriate exception in the `SqsProvider` that ensures the `DelaySeconds` parameter is always `0` when sending messages to a FIFO queue? This should probably be included in our message checks:
> 
> https://github.com/localstack/localstack/blob/82de91e30b0185a6792dd5047168b29d69bbb1f9/localstack-core/localstack/services/sqs/provider.py#L1200-L1217
> 
> Would you have capacity for this addition? 🤔
> >
Wouldn't that introduce parity between the behaviors of LocalStack and AWS?

AWS allows you to send a message to a FIFO queue with delay seconds as 0 (just 0 -- no other value -- AWS just allows you to send the message without any error and uses the delay defined on the queue)

Also the check where a FIFO queue can either have a null or 0 as the `DelaySeconds` is here:
https://github.com/localstack/localstack/blob/fcffe33c004bac69a339f506c17930b508a24cc7/localstack-core/localstack/services/sqs/models.py#L1029-L1042

gregfurman: > Hmm, isn't that scenario covered here?

Ahh you're 100% correct. Good spot ☝️  

> We should probably raise the appropriate exception in the SqsProvider that ensures the DelaySeconds parameter is always 0 when sending messages to a FIFO queue? This should probably be included in our message checks:

To clarify, I wasn't suggesting we overwrite the value. I was suggesting that we assert DelaySeconds is valid (prior to you pointing out that the case has already been covered) 🙂

gregfurman:
Also, it seems there are merge conflicts 🤔 Once they're resolved I'll merge 👍

kn1ves:
> Also, it seems there are merge conflicts 🤔 Once they're resolved I'll merge 👍

Resolved conflicts -- also ran the linter+formatter which I forgot previously. Thanks for the review :)

PR #12943, opened 2025-07-31
Add handling of tags in docker-helper

silv-io:
## Motivation

With the latest release we still had issues with pushing docker images on pipeline runs triggered by the tag push.

This is because the docker helper assumes that only the `main` branch should be allowed for pushing the images. Since tags are not bound to a branch, a tag-triggered run will not have a branch ref. 

We need to add extra provisions for allowing tagged commits to also push images. 

/cc @dfangl (and thanks for providing the patch files)

## Changes

- Allow tagged commits to also successfully push images
- Add bats test to assure this is possible

## Testing

To avoid further issues we'll do a run in a forked repository where we push into a different DockerHub account.

## TODO

What's left to do:

- [x] Thorough testing:
   - [x] Test that regular pushes to main will still work: https://github.com/silv-io/localstack/actions/runs/16909810881
   - [x] Test that tag pushes to main will start working: https://github.com/silv-io/localstack/actions/runs/16909892613 -> see [here](https://hub.docker.com/r/agihi/localstack/tags)
- [ ] ...

### DIFF
--- a/bin/docker-helper.sh
+++ b/bin/docker-helper.sh
@@ -80,6 +80,13 @@ function _enforce_tagged_or_main_branch() {
     echo "Current git branch: '$CURRENT_BRANCH'"
     echo "Current tag: '$CURRENT_TAG'"
     test -n "$CURRENT_TAG" || test "$CURRENT_BRANCH" == "$MAIN_BRANCH" || _fail "Current branch ($CURRENT_BRANCH) is not $MAIN_BRANCH and current tag ($CURRENT_TAG) is not set."
+
+    # if we're not on the main branch but have a tag, ensure it's a version tag like v1.2.3
+    if [[ "$CURRENT_BRANCH" != "$MAIN_BRANCH" && -n "$CURRENT_TAG" ]]; then
+      if ! [[ "$CURRENT_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+        _fail "tag '$CURRENT_TAG' is not a version tag of the shape ^v[0-9]+\.[0-9]+\.[0-9]+$"
+      fi
+    fi
 }
 
 function _enforce_no_fork() {
--- a/tests/bin/test.docker-helper.bats
+++ b/tests/bin/test.docker-helper.bats
@@ -148,6 +148,19 @@ setup_file() {
   [[ "$output" =~ "docker push $IMAGE_NAME:stable-$PLATFORM" ]]
 }
 
+@test "push fails on malformed tag" {
+  export MAIN_BRANCH="non-existing-branch"
+  export IMAGE_NAME="localstack/test"
+  export DOCKER_USERNAME=test
+  export DOCKER_PASSWORD=test
+  export PLATFORM=amd64
+  export TEST_TAG=not-a-version
+  run bin/docker-helper.sh push
+  [ "$status" -ne 0 ]
+  [[ "$output" =~ "tag '$TEST_TAG' is not a version tag" ]]
+}
+
+
 @test "push fails without PLATFORM" {
   export IMAGE_NAME="localstack/test"
   export MAIN_BRANCH="main"


tests/bin/test.docker-helper.bats
```diff
@@ -117,12 +120,26 @@ setup_file() {
 @test "push fails on non-default branch" {
   export MAIN_BRANCH="non-existing-branch"
   export IMAGE_NAME="localstack/test"
+  export DOCKER_USERNAME=test
+  export DOCKER_PASSWORD=test
```
alexrashed, line 129: **question:** Why is this necessary now?

tests/bin/test.docker-helper.bats
```diff
@@ -117,12 +120,26 @@ setup_file() {
...
   run bin/docker-helper.sh push
   [ "$status" -ne 0 ]
   [[ "$output" =~ "is not non-existing-branch" ]]
 }
 
+@test "push succeeds on tag without main branch" {
+  export MAIN_BRANCH="non-existing-branch"
+  export IMAGE_NAME="localstack/test"
+  export DOCKER_USERNAME=test
+  export DOCKER_PASSWORD=test
```
alexrashed, line 140: **question:** Same as above, why is this necessary?

tests/bin/test.docker-helper.bats
```diff
@@ -117,12 +120,26 @@ setup_file() {
...
+  export PLATFORM=amd64
+  export TEST_TAG=v1.0.0
+  run bin/docker-helper.sh push
+  echo "$output"
+  [ "$status" -eq 0 ]
```
alexrashed, line 144: **suggestion:** Would be great to also check that the proper docker tags here are being pushed.
And does it make sense to also test other scenarios in addition to this one new test?

tests/bin/test.docker-helper.bats
```diff
@@ -117,12 +120,26 @@ setup_file() {
...
 
+@test "push succeeds on tag without main branch" {
+  export MAIN_BRANCH="non-existing-branch"
+  export IMAGE_NAME="localstack/test"
+  export DOCKER_USERNAME=test
+  export DOCKER_PASSWORD=test
+  export PLATFORM=amd64
+  export TEST_TAG=v1.0.0
+  run bin/docker-helper.sh push
+  echo "$output"
```
alexrashed: **suggestion:** Seems like a leftover from testing?

tests/bin/test.docker-helper.bats
```diff
@@ -13,6 +13,9 @@ setup_file() {
     "branch")
       echo "main"
       ;;
+    "describe")
+            echo "$TEST_TAG"
```
alexrashed: **nit:** The indentation here is a bit off:
```suggestion
      echo "$TEST_TAG"
```

bin/docker-helper.sh
```diff
@@ -65,15 +65,21 @@ function _get_current_branch() {
     git branch --show-current
 }
 
+function _get_current_tag() {
+    git describe --tags --exact-match 2> /dev/null || true
```
alexrashed, line 69: **question:** Would it make sense to check that this really is a release tag? Is there a scenario where this can slip through and enable the push for something we do not want?

alexrashed (commented):
Thanks for following up on this! 💯 
Since this is quite a critical change this needs to be bullet proof, so I do have a few questions and some suggestions which I would love to discuss before we merge this.

tests/bin/test.docker-helper.bats
silv-io: To have a successful run of push, those values need to be set. Previously we didn't set them because the failure of the branch enforcement happens before the failure of the missing docker credentials. 

I believe this is an error. The only possible failure in this test should be the branch/tag enforcement.

tests/bin/test.docker-helper.bats
silv-io: Same as above, we don't want potential other errors polluting the output if we want to debug this test when something goes wrong.

tests/bin/test.docker-helper.bats
silv-io: true, will remove it.

tests/bin/test.docker-helper.bats
silv-io: Makes sense, will add this. 

We can add another test to test it failing if the docker tag is malformed or not a release tag.

tests/bin/test.docker-helper.bats
alexrashed: Thanks! Great catch! 💯

bin/docker-helper.sh
dfangl: I think this also leaves a bigger question: Should we allow any tags which are not release tags in the repository at all? This would be a case for the repository protection rules, then.

alexrashed: That's a good point! But since a few weeks we have branch protection rules in place in the main emulator repos. Only the release action / bot is allowed to create any tags.
Are you suggesting to make the rule even stricter such that the release action is only allowed to push version compliant tag names?

dfangl: It would make sense in my opinion, even though it is unlikely that the release action starts to push random tags. I think a ruleset:
1. Allowing `v\d+\.\d+\.\d+` (not sure of the regex flavor for github, please don't quote me on it) tags only for the release action / bot
2. Disallowing all other tag pushes
Would make sense. We could also add another check in the docker helper, to feel more safe, but I don't think it is necessary.

silv-io: I just want to add here that the pipeline which will use the docker-helper to push the images (aws-main.yml) can only be triggered by tags of this shape:
```
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

tests/bin/test.docker-helper.bats
silv-io: I've added such a test now and also tested the docker tags in this test

bin/docker-helper.sh
silv-io: I've also added functionality to the docker helper now to refuse using tags without this shape and a test for it

alexrashed: **thought:** That's awesome, thanks! Great improvement and cleanup! One follow up question: What happens if the commit is tagged twice?

alexrashed (approved):
Awesome! This is now really looking great and is nicely tested in the bats tests! 🧹 

It would be great if you could really verify that the changes we made here during the review iterations are working in your fork setup as well, just in case! 💚

bin/docker-helper.sh
silv-io: I'd assume the workflow would be triggered once for each tag. I can explore that

alexrashed: Thanks! Would also be very interesting how this function behaves for the two tags. :)

PR #12950, opened 2025-08-05
Replicator/support x account kms

cloutierMat:
## Motivation

This pr enables describing a kms key using the alias arn. This is part of solving a support issue with the replicator, where the user is querying a kms key x-account with it's alias.

## Changes

- the `key_id` returned by the regex will now return `alias/<alias_name>` if an alias arn is found.
- added test to validate the behavior

### DIFF
--- a/tests/aws/services/kms/test_kms.py
+++ b/tests/aws/services/kms/test_kms.py
@@ -2006,6 +2006,22 @@ class TestKMS:
             )
         snapshot.match("response-invalid-public-key", e.value.response)
 
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..CurrentKeyMaterialId"])
+    def test_describe_with_alias_arn(self, kms_create_key, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.key_value("CurrentKeyMaterialId"))
+
+        alias_name = f"alias/{short_uid()}"
+        created_key = kms_create_key(Description="test - description")
+        snapshot.match("created-key", created_key)
+
+        aws_client.kms.create_alias(TargetKeyId=created_key["KeyId"], AliasName=alias_name)
+        key_arn = created_key["Arn"]
+        alias_arn = key_arn.rsplit(":", maxsplit=1)[0] + f":{alias_name}"
+
+        describe_response = aws_client.kms.describe_key(KeyId=alias_arn)
+        snapshot.match("describe-key", describe_response)
+
 
 class TestKMSMultiAccounts:
     @markers.aws.needs_fixing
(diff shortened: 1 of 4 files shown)


tests/aws/services/kms/test_kms.py
```diff
@@ -2006,6 +2006,22 @@ def test_derive_shared_secret(self, kms_create_key, aws_client, snapshot):
...
+    def test_describe_with_alias_arn(self, kms_create_key, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.key_value("CurrentKeyMaterialId"))
+
+        alias_name = f"alias/{short_uid()}"
+        created_key = kms_create_key(Description="test - description")
+        snapshot.match("created-key", created_key)
+
+        aws_client.kms.create_alias(TargetKeyId=created_key["KeyId"], AliasName=alias_name)
+        key_arn = created_key["Arn"]
+        alias_arn = key_arn.rsplit(":", maxsplit=1)[0] + f":{alias_name}"
```
sannya-singal: Instead of manually creating the alias ARN here, we should use the kms operations to retrieve it, something like: 

```suggestion
        aliases = aws_client.kms.list_aliases()["Aliases"]
        alias_entry = next(a for a in aliases if a["AliasName"] == alias_name)
        alias_arn = alias_entry["AliasArn"]
```

tests/aws/services/kms/test_kms.py
```diff
@@ -2006,6 +2006,22 @@ def test_derive_shared_secret(self, kms_create_key, aws_client, snapshot):
...
         snapshot.match("response-invalid-public-key", e.value.response)
 
+    @markers.aws.validated
+    @markers.snapshot.skip_snapshot_verify(paths=["$..CurrentKeyMaterialId"])
+    def test_describe_with_alias_arn(self, kms_create_key, aws_client, snapshot):
+        snapshot.add_transformer(snapshot.transform.key_value("CurrentKeyMaterialId"))
+
+        alias_name = f"alias/{short_uid()}"
+        created_key = kms_create_key(Description="test - description")
+        snapshot.match("created-key", created_key)
```
sannya-singal, line 2016: Is there a reason we capture the create key response here? I think we can remove the snapshot since we cover this usage in a lot of other tests already, wdyt?

sannya-singal (commented):
Thanks for tackling this issue 🚀 🎉 I have just two minor comments, rest LGTM 🙌

tests/aws/services/kms/test_kms.py
cloutierMat: I knew there had to be an alias arn somewhere :rofl: :+1:

tests/aws/services/kms/test_kms.py
cloutierMat: There are actually no snapshots of `describe_key` called with an alias name or an alias arn, and that will come in relevant in the future if x-account is properly implemented since the arn returned is different then. Having the `created-key` snapshot will confirm the `AWSAccountId`/`Arn` consistency

sannya-singal (approved):
Thanks for addressing the comments, changes LGTM 🙂🚀

PR #12953, opened 2025-08-05
Add pyupgrade rules to ruff

giograno:
## Motivation
This PR introduces [Pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up) to our ruff configuration.
[pyupgrade](https://github.com/asottile/pyupgrade) is a tool that automatically upgrades Python syntax to newer versions. 

It introduces the same set of rules of [pro#4989](https://github.com/localstack/localstack-pro/pull/4989).

All the rules are automatically applied by the `make format`. They are also part of the pre-commit hook (after a `make init-precommit`).

## Changes
- Adding the `PU` rules to our `ruff` configuration. A few rules have been ignored as `ruff` does not yet provide a fix that guarantees the program's behavior to stay the same.
- Setting py39 rules for two paths (`localstack-core/localstack/utils/**` and `localstack-core/localstack/testing/pytest/**`) as they are imported from the CLI test.
- Ignore a few rules for the `__init__` files. The rationale is the same as the companion PR.
- Ignore the `PU` rules for `tests/aws/services/lambda_/functions/**. These modules are used by Lambda tests that run with a bunch of different runtimes.

I added a few self-contained commits to fix minor stuff I encountered, before committing the rules and all the formatting changes:
- https://github.com/localstack/localstack/pull/12953/commits/18363fbae8cf7a3a68fe6eb8ef81c6ca66c2fd6f that to convert all the `typing.Union` outside of type annotations (e.g., right-hand side of an assignment), as they can't be safely fixed by `ruff`.
- https://github.com/localstack/localstack/pull/12953/commits/fe94564646ea48ec00c7749177fc140b737fc45d removing two instances of `typing.ContextManager` in favor of `contextlib.AbstractContextManager`. This is an unsafe fix at the moment.
- I did find some type annotations with modules (`json` and `datetime`). I fixed them in https://github.com/localstack/localstack/pull/12953/commits/4631eca119f07c761dc52d4e7341f612e99c42b9 and https://github.com/localstack/localstack/pull/12953/commits/20dca66d59d279a0ff03279cf755cc94666f1f6f.
- https://github.com/localstack/localstack/pull/12953/commits/6925499d6fc827fa0ed441275c3d1c78487f97e7 to prevent the transformation of `Optional[RLock] `to `RLock | None`. In our Python version, `RLock` is not a type but a factory function. More info in the issue attached to the commit.
- In https://github.com/localstack/localstack/pull/12953/commits/8f07b53db1c065733300d8f26b8a8e79c97d25d2 I specifically avoided fixes for types used as test input for `convert_to_typed_dict`.


alexrashed (approved):
Awesome! From my perspective this is looking great! The changeset contains some very subtle changes which directly demonstrate why this kind of linting is a good idea to integrate! 💯 
The only question that I still have is the same as for the PR in the downstream project:
How should we best merge this in order to cause the minimum amount of friction for other developers? Would be great if you could outline the different steps / the timeline and the implications (i.e. the steps contributors should take), before moving forward.

giograno:
Thanks @alexrashed! My plan would be the following.
- Give a shout-out today to the entire team. Everyone should be aware that they will need to rebase their open PRs after this gets merged. Otherwise, `main` could likely (temporarily) break.
- Keeping an eye on this PR, keep rebasing and formatting all the new changes landing on `main` throughout he day.
- Merge the PR at a time in the day when few engineers are working (e.g., late CET evening or early CET morning).

silv-io (approved):
LGTM!

k-a-il (approved):
LGTM 🚀

PR #12959, opened 2025-08-06
Sfn jsonpath dictionaries

andrewmaye:
## Motivation

This is just raised to help support a bug I raised. It is currently demonstrating some behavioural differences between AWS and localstack with using JSONPath in step functions

https://github.com/localstack/localstack/issues/12960

## Changes

This PR currently only contains a new integration/snapshot test to demonstrate differences.

### DIFF
--- a/tests/aws/services/stepfunctions/v2/base/test_base.py
+++ b/tests/aws/services/stepfunctions/v2/base/test_base.py
@@ -396,7 +396,6 @@ class TestSnfBase:
             create_state_machine,
             sfn_snapshot,
             definition,
-            definition,
             exec_input,
         )
 


tests/aws/services/stepfunctions/v2/base/test_base.py
```diff
@@ -469,3 +469,30 @@ def test_json_path_array_wildcard_or_slice_with_no_input(
             definition,
             exec_input,
         )
+
+    @markers.aws.validated
+    @pytest.mark.parametrize(
+        "json_path_string",
+        ["$[*]", "$[0]", "$[0:]", "$[:0]", "$[0:1]", "$[1]", "$[1:]", "$[:1]", "$[1:2]"],
+    )
```
gregfurman, line 477: Have given this a look and it seems to be an issue with how our JSONPath dependency interprets dict values.

Looks like all slicing from the`0` index fails. Since they will currently fail against LocalStack, as per your bug ticket, could you skip these cases for now?

```python
$[*]
$[0]
$[0:]
$[0:1]
$[:1]
```

I recommend using a parametrize approach where the faulty cases are marked as skipped with a reason, but feel free to skip all tests if you so please 🙂 

Also, if we could ref your original issue as a comment here as to _why_ these are skipped that'd be brilliant.

i.e

https://github.com/localstack/localstack/blob/4f9792f57dfc01950e7c4b77ec68352e35011609/tests/aws/services/stepfunctions/v2/services/test_apigetway_task_service.py#L286-L289

alexrashed:
I guess it's been a bit too long 😅 I will close this PR for now, happy to revive it if @andrewmaye decides to revive it.

PR #12963, opened 2025-08-06
add java 24 support in java installer for trino

macnev2013:
## Motivation
This PR adds support for Java version 24, which is required for upgrading Trino server from version 389 to 476. Additionally, this PR refactors the module installation commands to improve code maintainability and readability.

## Changes
- Addition of new java version `24`
- Refactored application for more readable module installation command generation
- Addition of `jdk.incubator.vector` java module for java 24 for trino

## Testing
- Existing tests are passing which signifies the existing Java installations are not affected.
- java 24 installer is used in Trino in `pro` which was tested locally.


alexrashed (approved):
Nice and clean! 💯 
I think we should generally start upgrading the Java version of basically all our backends. This here is a great start, and will pave the way for an upgrade of the backend of Athena! ⬆️

PR #12966, opened 2025-08-06
add support for Fn::Tranform in CFnV2

pinzon:
## Motivation
This PR implements the capability to resolve FnTransform in the new CFnV2 engine in multiple locations of a template. 

## Changes

- Add properties (parent, jsonpath) to Scope class
- NodeResources and NodeProperties define a possible fn::transformation
- New NodeIntrinsicFunctionFnTransform class to model a Fn::Transform
- _visit_intrinsic_function can return a NodeIntrinsicFunctionFnTransform
- Refactor _excute_macro and AWS::Include into transformer
- Add events when a transformation fails
- Add template body character limit assertion
- Transformer adds some extra overrides due to missing resources during transformations

## Testing
- Removed `skip_if_v2_provider` in related tests

## Notes
This PR only implements the resolution of the transformations in specific locations. More locations are going to be needed in the future


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -597,12 +618,30 @@ def _visit_intrinsic_function(
...
-        node_intrinsic_function = NodeIntrinsicFunction(
-            scope=scope,
-            change_type=change_type,
-            intrinsic_function=intrinsic_function,
-            arguments=arguments,
-        )
+
+        if intrinsic_function == FnTransform:
+            if scope.count(FnTransform) > 1:
+                raise RuntimeError("Nested Fn::Transforms are bad")
```
simonrw: "are bad"! I like that error message!

But seriously... let's change that before we merge

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -597,12 +618,30 @@ def _visit_intrinsic_function(
...
+
+            path = "$" + ".".join(scope.split("/")[:-1])
```
simonrw, line 683: I do quite like the idea of making this implementation a helper on the `Scope` class, e.g.

```python
class Scope(str):
    # ...
    def to_jsonpath(self) -> str:
        return "$" + ".".join(self.split("/"))
```

perhaps even adding a `.parent` property if we know we need to traverse up the tree

```python
class Scope(str):
    # ...
    @property
    def parent(self) -> Self:
        return Scope(self._SEPARATOR.join(self.split(self._SEPARATOR)[:-1]))
```

or something

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -299,9 +300,18 @@ def _setup_change_set_model(
...
+                change_set_model_transform.transform()
+            )
+        except FailedTransformationException as e:
+            change_set.status = ChangeSetStatus.FAILED
+            change_set.status_reason = e.message
+            change_set.stack.set_stack_status(
+                status=StackStatus.ROLLBACK_IN_PROGRESS, reason=e.message
+            )
+            change_set.stack.set_stack_status(status=StackStatus.CREATE_FAILED)
+            return
```
pinzon, line 314: @simonrw This is the section I'm not liking that much

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -427,6 +469,26 @@ def __init__(
...
 
+class NodeIntrinsicFunctionFnTransform(NodeIntrinsicFunction):
+    def __init__(
+        self,
+        scope: Scope,
+        change_type: ChangeType,
+        intrinsic_function: str,
+        arguments: ChangeSetEntity,
+        before_siblings: list[Any],
+        after_siblings: list[Any],
```
simonrw, line 480: **question**: I'm slightly confused why the `arguments` can be a `ChangeSetEntity` but the siblings have to be separate values. When I try to change it I end up re-visiting the transform and hitting the "Invalid: Fn::Transforms cannot be nested inside another Fn::Transform" case. I understand why (the `Fn::Transform` is included in the siblings) and I don't think it's worth it but perhaps we can try to switch to this type (i.e. make these a single `siblings: ChangeSetEntity` property) in a follow up

localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -611,12 +673,32 @@ def _visit_intrinsic_function(
...
-            arguments=arguments,
-        )
+
+        if intrinsic_function == FnTransform:
+            if scope.count(FnTransform) > 1:
+                raise RuntimeError(
+                    "Invalid: Fn::Transforms cannot be nested inside another Fn::Transform"
+                )
+
+            path = "$" + ".".join(scope.split("/")[:-1])
```
simonrw, line 683: **suggestion**: could this be replaced with 


```suggestion
            path = scope.parent.jsonpath
```

localstack-core/localstack/services/cloudformation/v2/provider.py
simonrw: I think you are right, it's not general enough (e.g. it hard codes `CREATE_FAILED`) however I think as we get more testing around transforms in general this will work itself out.

tests/aws/services/cloudformation/test_change_set_fn_transform.py
```diff
@@ -2,13 +2,14 @@
...
 from localstack_snapshot.snapshots.transformer import RegexTransformer
+from tests.aws.services.cloudformation.conftest import skip_if_v1_provider
 
 from localstack.aws.api.lambda_ import Runtime
 from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
 
 
-@pytest.mark.skip(reason="Not implemented with either provider")
+@skip_if_v1_provider("change sets")
```
simonrw, line 12: 🎉

tests/aws/services/cloudformation/test_template_engine.py
```diff
@@ -830,7 +825,7 @@ def test_attribute_uses_macro(self, deploy_cfn_template, create_lambda_function,
         assert "test-" in resulting_value
 
     @markers.aws.validated
-    @pytest.mark.skip(reason="Fn::Transform does not support array of transformations")
+    @skip_if_v1_provider("V1 is unable to resolve fn::transform with lists")
```
simonrw, line 828: 🎉

Issue #5018, opened 2021-11-26
bug: DynamoDB Stream Lambdas fail because no "SequenceNumber" field in event


toadzky:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a rust lambda running using the `provided.al2` runtime and the `docker` lambda executor. When I update my dynamo table, the lambda invokes but the event has an incorrect structure and the lambda throws an error parsing the event payload. The error from the lambda container is 

```
Error: Error("missing field `SequenceNumber`", line: 1, column: 1022)
```

Here's a copy of the `AWS_LAMBDA_EVENT_BODY` set by localstack: 

```json

{
    "Records": [
        {
            "eventID": "79ac67bf",
            "eventVersion": "1.1",
            "dynamodb": {
                "ApproximateCreationDateTime": 1637948167.023251,
                "SizeBytes": 722,
                "Keys": {
                    ...
                },
                "NewImage": {
                    ....
                },
                "StreamViewType": "NEW_AND_OLD_IMAGES"
            },
            "awsRegion": "test-region",
            "eventSource": "aws:dynamodb",
            "eventName": "INSERT",
            "eventSourceARN": "arn:aws:dynamodb:test-region:000000000000:table/k4JV6xP4"
        }
    ]
}
```

### Expected Behavior

The lambda should invoke with a correctly structured event, as [defined by AWS](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_StreamRecord.html). Here's an example from [the official tutorial](url) (highlighting by me): 

```json
{
    "Records": [
        {
            "eventID": "7de3041dd709b024af6f29e4fa13d34c",
            "eventName": "INSERT",
            "eventVersion": "1.1",
            "eventSource": "aws:dynamodb",
            "awsRegion": "region",
            "dynamodb": {
                "ApproximateCreationDateTime": 1479499740,
                "Keys": {
                    ...
                },
                "NewImage": {
                    ...
                },
                *"SequenceNumber": "13021600000000001596893679",*
                "SizeBytes": 112,
                "StreamViewType": "NEW_IMAGE"
            },
            "eventSourceARN": "arn:aws:dynamodb:region:123456789012:table/BarkTable/stream/2016-11-16T20:42:48.104"
        }
    ]
}
```

### How are you starting LocalStack?

Custom (please describe below)

### Steps To Reproduce

[...]

PR #12968, opened 2025-08-07
Re-enabling ruff rule E721

giograno:
## Motivation
Following up on our efforts for a better linting (e.g., #12953 and [#5018](https://github.com/localstack/localstack-pro/pull/5018)), this PR re-enables the rule E721 and adjusts the code to be compliant.


alexrashed (approved):
Thanks for jumping on this and getting rid of one more linting exclusion! 💯 🧹

PR #12969, opened 2025-08-07
Re-enabling ruff rule C408

giograno:
## Motivation
Along the lines of #12968.
`ruff` considers fixes for C408 as potentially unsafe as they could remove comments ([docs](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/)).

## Changes
- Remove C408 from the list of unsafe rules;
- Run `ruff check --select C408 . --fix --unsafe-fixes` on the codebase.
- Verify that no comment has been dropped.


alexrashed (approved):
Awesome! Thanks for jumping on yet another linting exception! 🧹

PR #12971, opened 2025-08-07
CFNV2: fix EC2 VPCEndpoint test

simonrw:
## Motivation

One of the CFNV2 tests that was failing seems relatively straightforward: it was specifically to do with the vpc endpoint resource. This should be an easy fix I thought, probably something in the provider I thought...

No it was as straightforward as: the target property for a `GetAtt` is represented as a `datetime` internally. We assume that non-strings are an error.

On AWS we believe CFn values _are_ likely strings throughout, but we have the ability to keep the `datetime` type around to be more useful for users.

## Changes

* Treat `datetime`s as acceptable values when it comes to performing a `GetAtt` on a resource
* Unskip 1 test

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
@@ -4,7 +4,6 @@ import base64
 import copy
 import re
 from collections.abc import Callable
-from datetime import datetime
 from typing import Any, Final, Generic, TypeVar
 
 from botocore.exceptions import ClientError


pinzon (commented):
Are we sure that the resource provider isn't at fault here? It's seems more likely than adding value types to the supported types.

simonrw:
> Are we sure that the resource provider isn't at fault here? It's seems more likely than adding value types to the supported types.

As in you think we should convert to a string in the resource provider?

pinzon:
Yes :+1:

simonrw:
> Yes 👍

I think you are right, the schema for the resource specifies that the value is a `string` as well 👍

pinzon (approved):
Thanks for tackling this :+1:

PR #12972, opened 2025-08-07
SES: fix infinite loop with SNS event destination

bentsku:
## Motivation
While testing an SES/SNS sample, we encountered an issue where sending an email with SES with an SNS event destination configured that would have another SES subscriber, it would trigger an infinite loop. 

There were 2 problems:
- SNS was sending the `send` and `delivery` notification even if you didn't configure it
- it was sending for event destinations that exists for all `ConfigurationSet`, even if you did not specify a `ConfigurationSetName`. SES only sends event notification if you pass a `ConfigurationSetName` because `EventDestinations` are inherently linked to a configuration set, see https://docs.aws.amazon.com/ses/latest/dg/monitor-sending-using-event-publishing-setup.html

> The following steps required for setting up event publishing are covered in the topics below:
>
> - You must create a configuration set using the Amazon SES console or API.
> - Add one or more event destinations (CloudWatch, Firehose, Pinpoint, or SNS) to the configuration set, and configure parameters unique to the event destination.
> - **When you send an email, you specify which configuration set to use that contains your event destination.**

There is still one somewhat big issue because Moto only accepts one `EventDestination` per configuration set, even if AWS accepts multiple, Moto will overwrite it every time you call `CreateConfigurationSetEventDestination`, but this is out of scope. 

## Changes
- refactor a bit the SNS sending to centralize in one point and not have to modify the code 3 times, and for it to be easy to extend. We will probably need to refactor it again if we want to extend the `SNSEmitter` and add new kind of destinations
- add a new LocalStack only test because we cannot really look into received emails when validating against AWS
- write some TODOs about behavior problem of Event Destinations

### DIFF
--- a/tests/aws/services/ses/test_ses.py
+++ b/tests/aws/services/ses/test_ses.py
@@ -12,7 +12,7 @@ from localstack.services.ses.provider import EMAILS, EMAILS_ENDPOINT
 from localstack.testing.aws.util import is_aws_cloud
 from localstack.testing.pytest import markers
 from localstack.utils.strings import short_uid
-from localstack.utils.sync import retry
+from localstack.utils.sync import poll_condition, retry
 
 SAMPLE_TEMPLATE = {
     "TemplateName": "hello-world",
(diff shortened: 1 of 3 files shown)


tests/aws/services/ses/test_ses.py
```diff
@@ -918,6 +918,79 @@ def test_special_tags_send_email(self, tag_name, tag_value, aws_client):
...
+    ):
+        """
+        Validates that configure Event Destinations and sending an email does not trigger an infinite loop of sending
+        SNS notifications that sends an email that would trigger SNS.
+        """
+
+        sender_email_address, recipient_email_address = setup_email_addresses()
+        config_set_name = f"config-set-{short_uid()}"
+
+        EMAILS.clear()
```
viren-nadkarni: I would ask to use the REST endpoint for SES email introspection because tests are not guaranteed to run in the same process as LocalStack.

- GET `/_aws/ses`
- DELETE `/_aws/ses`

viren-nadkarni (changes requested):
Just one request. Other affected tests also need updating but it's out of scope.

tests/aws/services/ses/test_ses.py
bentsku: Sure, makes sense 👍 will update

viren-nadkarni: Pinging @giograno: Are there test fixtures for the [LocalStack SDK](https://docs.localstack.cloud/aws/tooling/localstack-sdks/python-sdk/#ses)?

bentsku: Yeah that actually would be super great to use 👍 I think we can do a second pass to update all usage of `EMAILS` and use the SDK instead 👀

viren-nadkarni (approved):
Thanks, looks good :100:

tests/aws/services/ses/test_ses.py
giograno: @viren-nadkarni, we don't have it atm, but we should totally do it. Happy to take this :D

PR #12976, opened 2025-08-08
add transformation function with key skip

baermat:
## Motivation
Current transformation from pascal case to lower camel case doesn't let us skip keys. In the [create_codepipeline](http://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codepipeline/client/create_pipeline.html) boto call, the "configuration" of action declarations is listed as
```
'configuration': {
  'string': 'string'
  },
```
In the [AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-codepipeline-pipeline-actiondeclaration.html), it is listed as JSON, and uses pascal case. The current implementation transforms all keys, therefore also wrongly transforming the configuration values (configuration itself is and should be transformed).

this code would be used [in the codepipeline resource provider](https://github.com/localstack/localstack-pro/pull/4805/files#diff-afa3f90d6eaa051a7b51948d12f341f6f82a8f3e42cc388eccd71991b6257ec1)

## Changes
- add new function for this transformation that allows the skipping of provided keys.

## TODO

I would be happy about any input on how to proceed here. Mainly, should this be done one a more general level in the recurse_properites function? Options that come to mind are.
a) We leave it as it is.
b) The current implementation uses `recurse_properties`. This could be available for all functions that use it by implementing this logic in `recursed_properties` (I didn't want to touch a potentially huge surface to be honest.)
c) Contain it for this special case: Leave as is but move it somewhere else more specific to codepipeline.

### DIFF
--- a/tests/unit/services/cloudformation/test_provider_utils.py
+++ b/tests/unit/services/cloudformation/test_provider_utils.py
@@ -135,3 +135,68 @@ class TestDictUtils:
                 }
             ],
         }
+
+    def test_lower_camelcase_to_pascalcase_skip_keys(self):
+        original_dict = {
+            "Stages": [
+                {
+                    "Actions": [
+                        {
+                            "Actiontypeid": {
+                                "Category": "Source",
+                                "Owner": "AWS",
+                                "Provider": "S3",
+                                "Version": "1",
+                            },
+                            "Configuration": {
+                                "S3bucket": "localstack-codepipeline-source-86a13a88",
+                                "S3objectkey": "source-key",
+                                "Subconfig": {"Subconfig1": "Foo", "Subconfig2": "bar"},
+                            },
+                            "Inputartifacts": [],
+                            "Name": "S3Source",
+                            "Namespace": "S3SourceVariables",
+                            "Outputartifacts": [{"Name": "Artifact_Source_S3Source"}],
+                            "Rolearn": "arn:aws:iam::096845016391:role/EcrPipelineStack-MyPipelineSourceS3SourceCodePipeli-YOoRQUZQe6WU",
+                            "Runorder": 1,
+                        }
+                    ],
+                    "Name": "Source",
+                }
+            ]
+        }
+        target_dict = {
+            "stages": [
+                {
+                    "actions": [
+                        {
+                            "actiontypeid": {
+                                "category": "Source",
+                                "owner": "AWS",
+                                "provider": "S3",
+                                "version": "1",
+                            },
+                            # The excluded key itself is transformed
+                            # Its values are not
+                            # Recursion stops, items at lower levels are not transformed as well
+                            "configuration": {
+                                "S3bucket": "localstack-codepipeline-source-86a13a88",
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/cloudformation/provider_utils.py
```diff
@@ -85,6 +85,27 @@ def _keys_pascalcase_to_lower_camelcase(obj):
     return _recurse_properties(model, _keys_pascalcase_to_lower_camelcase)
 
 
+def keys_pascalcase_to_lower_camelcase_skip_keys(model: dict, exclude_keys: set = None) -> dict:
```
simonrw: We already have `keys_pascalcase_to_lower_camelcase` which is a simplified implementation, so presumably

```python
def keys_pascalcase_to_lower_camelcase(model: dict) -> dict:
    return keys_pascalcase_to_lower_camelcase_skip_keys(model)
```

It might be nice to make this change

simonrw (approved):
I am happy to have this implementation as it is but I've raised a comment about refactoring the existing function to use it, which brings me on to my main suggestion: can we add a unit test for this covering the casing example so we have a clear usage example for future use? _If_ we refactor the existing function to use this  one then we have reasonable test coverage from the existing integration tests, but a unit test can serve as documentation for helper functions like this.

Another point is: @dominikschubert implemented in #11310 a type-converting boto client factory that's given to every resource provider. Is that not working for this situation? Is it because this property is not interpreted by `boto3` and instead passed on "as is"?

localstack-core/localstack/services/cloudformation/provider_utils.py
simonrw: Maybe something like this?

```python
def test_lower_camelcase_to_pascalcase_skip_keys(self):
    original_dict = {
        "EventBusName": "my-event-bus",
        "Targets": [
            {
                "Id": "AnId",
            }
        ],
    }

    converted_dict = utils.keys_pascalcase_to_lower_camelcase_skip_keys(original_dict, {"id"})
    assert converted_dict == {
        "eventBusName": "my-event-bus",
        "targets": [
            {
                "id": "AnId",
            }
        ],
    }
```
but ideally with a more complex structure showing off where it does and doesn't work?

pinzon:
@simonrw I can answer this point:
> Another point is: @dominikschubert implemented in https://github.com/localstack/localstack/pull/11310 a type-converting boto client factory that's given to every resource provider. Is that not working for this situation? Is it because this property is not interpreted by boto3 and instead passed on "as is"?

The issue with the auto-type-converting (lets's call it that for now) is that only works with the operation arguments directly. This doesn't work in this case where all the CFn resource definition is wrapped inside the "pipeline" argument.

localstack-core/localstack/services/cloudformation/provider_utils.py
baermat: Does the implemented test/function refactoring is what you wanted to see? just making sure I understood correctly

PR #12978, opened 2025-08-08
add(utils/archives): Support .whl files in download_and_extract

gregfurman:
## Motivation

Python wheels are considered valid `.zip` archives and can be extracted using the same functionality. This change allows us to extend the `ArchiveDownloadAndExtractInstaller` to download and unzip python wheels.

## Changes
- Python `.whl` files are unzipped when doing a `download_and_extract` whereas previously an exception was raised.


joe4dev (approved):
LGTM 📦

PR #12979, opened 2025-08-08
Improve remove docker container function to take into account stopped containers

bryansan-local:
## Motivation
Right now, this remove container method doesn't take into account stopped containers when you ask it to check for existence of it

Related PR for more context about why bother with this change -> https://github.com/localstack/localstack-pro/pull/5037

## Changes
* I have introduced a new method to return all containers including stopped ones (`get_all_container_names`) I didn't touch the old method (`get_running_container_names`) to keep compatibility and not break any code already using it.
* I changed the remove container method to use this new method ^ so it takes into account stopped containers when you ask it to check for existence. This is kinda a breaking change BUT in my opinion this is an improvement to the remove operation.

### DIFF
--- a/localstack-core/localstack/utils/container_utils/docker_cmd_client.py
+++ b/localstack-core/localstack/utils/container_utils/docker_cmd_client.py
@@ -924,7 +924,7 @@ class CmdDockerClient(ContainerClient):
         )
 
     def _check_output_and_raise_no_such_container_error(
-        self, container_name_or_id: str, output: str, error: str | None = None
+        self, container_name_or_id: str, output: str, error: Optional[str] = None
     ):
         """
         Check the given client invocation output and raise a `NoSuchContainer` exception if it


localstack-core/localstack/utils/container_utils/container_client.py
alexrashed, line 1: **suggestion:** This module is incredibly important and well-tested, and we basically try to have tests for every change whenever possible. Could you please add a test for this edge case to the test for this module?
The tests are done in this class here:
https://github.com/localstack/localstack/blob/34650cc976d9e16bc1087378a900af9ee17172c6/tests/integration/docker_utils/test_docker.py#L131
When adding a new test, it will automatically be executed for both - the SDK and the CMD client. :)

dfangl (commented):
LGTM, but I agree with Alex on the tests, would be nice to get one!

localstack-core/localstack/utils/container_utils/container_client.py
bryansan-local: @alexrashed @dfangl I have added tests covering these edge scenarios. I also took the initiative to expand beyond that since I didn't see tests specific for the `get_running_container_names` and `remove_container` methods. They were "kinda" covered by other tests but I am a firm believer we need to have dedicated ones.

From these dedicated ones I found a "bug" in the `remove_container` method. The sdk client throws an exception if the container doesnt exist and the `force` parameter is `false` but the cmd client doesn't do that. it swallows the error no matter what you pass in the `force` parameter. I have made the required changes to make it consistent with the sdk client. Let me know if you agree with that.

See the new tests `test_remove_container_should_throw_exception_when_container_doesnt_exist_and_not_forcing` and `test_remove_container_should_work_even_when_container_doesnt_exist_because_of_forcing` which covers that scenario where the container doesnt exist + the force parameter value

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -915,13 +922,20 @@ def _check_and_raise_no_such_container_error(
         Check the given client invocation error and raise a `NoSuchContainer` exception if it
         represents a `no such container` exception from Docker or Podman.
         """
-
-        # consider different error messages for Docker/Podman
-        error_messages = ("No such container", "no container with name or ID")
         process_stdout_lower = to_str(error.stdout).lower()
```
alexrashed: **nit / suggestion:** This function basically boils down to this:
```
self. _check_output_and_raise_no_such_container_error(container_name_or_id, to_str(error.stdout))
```

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -37,6 +37,9 @@
 
 LOG = logging.getLogger(__name__)
 
+# consider different error messages for Docker/Podman
+container_not_found_error_messages = ("No such container", "no container with name or ID")
```
alexrashed: **nit / suggestion:** Maybe this one should be prefixed with an underscore, as it's just an internal member, not part of the API?

localstack-core/localstack/utils/container_utils/container_client.py
alexrashed: Awesome! Thanks a lot for the detailed new tests! Really great that you also fixed a bug along the way, I totally agree with the testing principle and the force parameter semantics you outlined.

alexrashed (approved):
Nice! Thank you so much for addressing the comments and even fixing a bug along the way!
I think it would be great to get a final review from @dfangl as owner of this module, but then we should be good to go! 💯

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
bryansan: Hope you like the new version. I didnt unified them before as I'm not really a fan of merging them into one since they receive different parameters (subprocess.CalledProcessError vs str) but I used your idea to come up with this version. Let me know if you really want to merge them as one or if this one is good enough

alexrashed: Nice, that was exactly my intention, LGTM!

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -266,17 +266,21 @@ def commit(
...
         cmd = self._docker_cmd() + ["rm"]
         if force:
             cmd.append("-f")
         cmd.append(container_name)
         LOG.debug("Removing container with cmd %s", cmd)
         try:
-            run(cmd)
+            output = run(cmd)
+            # When the container does not exist, the output could have the error message without any exception
+            if isinstance(output, str) and not force:
```
dfangl, line 279: Does the output have the error message if force is set? I am fine with checking this, just wondering

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
```diff
@@ -266,17 +266,21 @@ def commit(
...
+                self._check_output_and_raise_no_such_container_error(container_name, output=output)
```
dfangl, line 280: Very nice catch! It seems the docker CLI does not exit with a non-zero exit code in this case, weird!

dfangl (approved):
LGTM! Nice tests, and very good catch with the remove container function! Thanks for adding all those sweet new tests!

localstack-core/localstack/utils/container_utils/docker_cmd_client.py
bryansan-local: Yes, for the CMD one the error is in the output without any exception involved even when force = True

PR #12980, opened 2025-08-08
CFNV2: improve parity with describe operations

simonrw:
- **Only set change set id on deployment completion**
- **Unskip and update skip**
- **Support propagating stack description**
- **Increase parity with describe stacks and change sets**
- **Remove parity check for notification arns**
- **Better parity with outputs**


pinzon (approved):
LGTM :+1:

Issue #10000, opened 2024-01-05
bug: Kinesis Subscriber returns SubscribeToShardEvent with invalid sequence number of 0


fs-jasonc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## Behavior 
When implementing a `Subscriber<SubscribeToShardEventStream>` using the AWS Java SDK, starting from a timestamp and inserting records, `onNext` returns a `SubscribeToShardEvent` with a `continuationSequenceNumber` of 0. The returned record sequenceNumbers are like:
```
49658026119490182770032252969867129256605616914593865986
49658026119490182770032252969868338182425231543768572162
49658026119490182770032252969869547108244846172943278338
```
Creating a new subscriber using this sequence number of 0 gives an error:
`Unable to find record with provided SequenceNumber SequenceNumber(0) in stream`

Localstack says:
```
localstack      | 2024-01-05T11:29:24.910  INFO --- [functhread10] l.s.k.kinesis_mock_server  : kinesis.mock.ResourceNotFoundException: Unable to find record with provided SequenceNumber SequenceNumber(0) in stream Some(state-store-test-stream-8e44ffd7-b8ea-4a8e-b78a-d64d3d95fb4e)
```

### Expected Behavior

The continuationSequenceNumber should be the highest sequence number reached, not 0 (a non-existent record).

### How are you starting LocalStack?

With a docker-compose file

### Steps To Reproduce

1. Create a Kinesis Stream
2. Put records to it
3. Subscribe to it. Read records (note 0 continuationSequenceNumber)
5. Put more records
6. Subscribe to it again, using position of AFTER_SEQUENCE_NUMBER of 0. Get error due to invalid sequence number.

### Environment

```markdown
- OS: Ubuntu
- LocalStack: 3.0.5
- Java AWS SDK: 2.21.37
```

## Cause
`ContinuationSequenceNumber` is simply set to 0 and never updated when records are added:

https://github.com/localstack/localstack/blob/master/localstack/services/kinesis/provider.py#L136

In previous versions (eg 0.14.5) which worked, I think the number was derived from the records.

## Versions
This wasn't in v0.14.5, but is in v3.0.5 and v2.3.2

mdemille:
This still needs a fix. See the above issue for some additional information, including logs and a minimal example to demonstrate the problem.

evanbarn2:
Still needs a fix. I think the problem description above is clear. Same happens on 2.8, 3.6, 3.8. (MacOs).
[...]

PR #12981, opened 2025-08-08
Fix ContinuationSequenceNumber in SubscribeToShard

NigelRook:
## Motivation
Currently, when using SubscribeToShard, SubscribeToShardEvents will always have a ContinuationSequenceNumber of 0.

https://github.com/localstack/localstack/issues/10000

## Changes
When using Kinesis SubscribeToShard, ContinuationSequenceNumber will now match the sequence number of the last record sent, rather than 0

## Testing
Tested locally with a WIP consumer

### DIFF
--- a/tests/aws/services/kinesis/test_kinesis.py
+++ b/tests/aws/services/kinesis/test_kinesis.py
@@ -175,7 +175,15 @@ class TestKinesis:
         snapshot.match("One_consumer_by_describe_stream", consumer_description_by_arn)
 
     @markers.aws.validated
-    @markers.snapshot.skip_snapshot_verify(paths=["$..Records..EncryptionType"])
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            "$..Events..EncryptionType",
+            # TODO: We set this to the final sequence_number of returned records which is technically not correct,
+            # but will suffice as a checkpoint for now. See AWS docs:
+            # https://docs.aws.amazon.com/kinesis/latest/APIReference/API_SubscribeToShardEvent.html#API_SubscribeToShardEvent_Contents
+            "$..Events..ContinuationSequenceNumber",
+        ]
+    )
     def test_subscribe_to_shard(
         self,
         kinesis_create_stream,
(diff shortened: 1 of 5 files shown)


NigelRook:

NigelRook:
recheck

gregfurman:
Hey @NigelRook! Thanks for the PR 🚀  This approach will 100% close that bug ticket and unblock users, so am fine with it going in. Would it be alright if I pushed some [parity tests](https://github.com/localstack/localstack/blob/main/docs/testing/parity-testing/README.md) to this feature branch? You'll retain full authorship of the contribution since we squash commits in anycase 🙂 

Just to note, there is a minor distinction we should draw between the value of `ContinuationSequenceNumber` and an actual record's `SequenceNumber`. From the [AWS docs](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_SubscribeToShardEvent.html#API_SubscribeToShardEvent_Contents) (bolding is my own):

> Use `ContinuationSequenceNumber` for checkpointing because it captures your shard progress **even when no data is written to the shard.**

So a snapshot test against AWS would actually fail -- since `ContinuationSequenceNumber` is a logical checkpoint marker that may not correspond to an actual data record. Implementing this behvaiour is definitely out-of-scope so we can explore it later if need be.

NigelRook:
Totally fine with you pushing improvements and/or tests in here

tests/aws/services/kinesis/test_kinesis.py
```diff
@@ -175,7 +175,15 @@ def test_stream_consumers(
         snapshot.match("One_consumer_by_describe_stream", consumer_description_by_arn)
 
     @markers.aws.validated
-    @markers.snapshot.skip_snapshot_verify(paths=["$..Records..EncryptionType"])
+    @markers.snapshot.skip_snapshot_verify(
+        paths=[
+            "$..Events..EncryptionType",
+            # TODO: We set this to the final sequence_number of returned records which is technically not correct,
```
joe4dev, line 181: praise: thank you for clarifying this limitation in a comment

tests/aws/services/kinesis/test_kinesis.py
```diff
@@ -210,21 +218,26 @@ def test_subscribe_to_shard(
         # put records
         num_records = 5
         msg = "Hello world"
-        for i in range(num_records):
-            aws_client.kinesis.put_records(
-                StreamName=stream_name, Records=[{"Data": f"{msg}_{i}", "PartitionKey": "1"}]
-            )
+        aws_client.kinesis.put_records(
+            StreamName=stream_name,
+            Records=[{"Data": f"{msg}_{i}", "PartitionKey": "1"} for i in range(num_records)],
```
joe4dev, line 223: praise: neat improvement ✨

PR #12989, opened 2025-08-11
Log stack traces only in debug mode

sannya-singal:
## Motivation
As a part of the initiative for improving Localstack error messages and stack traces, this PR aims to ensure that stack traces are logged when debug mode is enabled. Currently, in some use cases stack traces are logged without checking if debug mode is enabled or not. This causes excessive log noise even when `DEBUG=0`. 

## Changes
This PR cleans up logging across different use cases to ensure that stack traces are only logged when debug mode is enabled.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/template_deployer.py
@@ -1479,10 +1479,11 @@ class TemplateDeployer:
                         # correct order yet.
                         continue
                     case OperationStatus.FAILED:
-                        LOG.exception(
+                        LOG.error(
                             "Failed to delete resource with id %s. Reason: %s",
                             resource_id,
                             event.message or "unknown",
+                            exc_info=LOG.isEnabledFor(logging.DEBUG),
                         )
                     case OperationStatus.IN_PROGRESS:
                         # the resource provider executor should not return this state, so
@@ -1494,10 +1495,11 @@ class TemplateDeployer:
                         raise Exception(f"Use of unsupported status found: {other_status}")
 
             except Exception as e:
-                LOG.exception(
+                LOG.error(
                     "Failed to delete resource with id %s. Final exception: %s",
                     resource_id,
                     e,
+                    exc_info=LOG.isEnabledFor(logging.DEBUG),
                 )
 
         # update status
--- a/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
+++ b/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
@@ -69,7 +69,10 @@ class AlarmScheduler:
         schedule_period = evaluation_periods * period
 
         def on_error(e):
-            LOG.exception("Error executing scheduled alarm", exc_info=e)
+            if LOG.isEnabledFor(logging.DEBUG):
+                LOG.exception("Error executing scheduled alarm", exc_info=e)
+            else:
+                LOG.error("Error executing scheduled alarm")
 
         task = self.scheduler.schedule(
             func=calculate_alarm_state,
--- a/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
+++ b/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
@@ -69,10 +69,7 @@ class AlarmScheduler:
         schedule_period = evaluation_periods * period
 
         def on_error(e):
-            if LOG.isEnabledFor(logging.DEBUG):
-                LOG.exception("Error executing scheduled alarm", exc_info=e)
-            else:
-                LOG.error("Error executing scheduled alarm")
+            LOG.error("Error executing scheduled alarm", exc_info=LOG.isEnabledFor(logging.DEBUG))
 
         task = self.scheduler.schedule(
             func=calculate_alarm_state,
--- a/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
+++ b/localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
@@ -69,7 +69,10 @@ class AlarmScheduler:
         schedule_period = evaluation_periods * period
 
         def on_error(e):
-            LOG.error("Error executing scheduled alarm", exc_info=LOG.isEnabledFor(logging.DEBUG))
+            if LOG.isEnabledFor(logging.DEBUG):
+                LOG.exception("Error executing scheduled alarm", exc_info=e)
+            else:
+                LOG.error("Error executing scheduled alarm")
 
         task = self.scheduler.schedule(
             func=calculate_alarm_state,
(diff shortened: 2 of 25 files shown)


localstack-core/localstack/services/s3/notifications.py
```diff
@@ -403,7 +403,11 @@ def _verify_target(self, target_arn: str, verification_ctx: BucketVerificationCo
                 QueueName=arn_data["resource"], QueueOwnerAWSAccountId=arn_data["account"]
             )["QueueUrl"]
         except ClientError:
-            LOG.exception("Could not validate the notification destination %s", target_arn)
+            LOG.exception(
+                "Could not validate the notification destination %s",
+                target_arn,
+                exc_info=LOG.isEnabledFor(logging.DEBUG),
+            )
```
bentsku: nit: this stayed `LOG.exception` 

This is very nice, would it be possible to do the following so that we still get why it failed? Sorry it's on me, the current exception logging wasn't that great for this one:
```python
except ClientError as e:
    code = response["Error"]["Code"]
    LOG.error(
        "Could not validate the notification destination %s: %s",
        target_arn,
        code,
    )
```

So that the user would still get an indication as to why it failed, as this is a Boto error, we can provider at least the name of the exception like "AccessDenied" for example. 

I had started reviewing the rest but saw you put back on draft 👍 will do the rest later

sannya-singal: Super sorry for the early ping 🙏 There are more incoming changes. Sure 💯  I will fix this one 👍

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
```diff
@@ -1479,10 +1479,11 @@ def delete_stack(self):
                         # correct order yet.
                         continue
                     case OperationStatus.FAILED:
-                        LOG.exception(
+                        LOG.error(
```
simonrw, line 1482: **suggestion**: it might be nice to extract a helper method to our `LOG` class that we can use so we don't forget to set this behaviour in the future. E.g.

```python
from localstack.logging import get_logger

LOG = get_logger(__name__)

LOG.error("message")
```

which in turn is implemented like
```python
class LocalStackLogger(logging.Logger):
    def _log(self, level, *args, **kwargs):
        kwargs.pop("exc_info", None)
        super()._log(level, *args, **kwargs, exc_info=self.isEnabledFor(logging.DEBUG))

    def exception(self, *args, **kwargs):
        """
        Protect against using this method which logs the exception by default
        """
        return self.error(*args, **kwargs)
```

WDYT? Or we could write a linting rule that operates over our AST and catches usages of the `LOG.error(...)` without handling the `exc_info` field properly, or protects us against usages of `LOG.exception`?

simonrw (approved):
LGTM for CFn

localstack-core/localstack/services/cloudformation/engine/template_deployer.py
bentsku: I definitely agree about the linting rule to protect about the `exception` usage, this is the main issue we have currently.

For the logger above, I'm not sure, or we'd instead need to override `exception` with a conditional `exc_info` instead of the default keyword `exc_info=True` that it has currently, so we'd be protected without changing behavior for the rest. 
Also, not sure what happens when adding `exc_info` outside of an exception handler (`except`), but not anything good I suppose. 

I think we'd still need to be deliberate about when we want to log stack traces or not, and always printing stack traces might be too much, it might be an exception we're actually expecting so we would not need it.

edit: it seems Ruff does not support custom linting rules though: https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-flake8, so we might need to look into that.

sannya-singal: I agree, the lint rule for `exception` seems like a nice way here, but as Ben mentioned Ruff seems not to support custom linting rules. 
On the `LocalStackLogger` idea, I see the point there but adding `exc_info` outside of an except block might not be the best option here.

localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
```diff
@@ -69,7 +69,10 @@ def schedule_metric_alarm(self, alarm_arn: str) -> None:
         schedule_period = evaluation_periods * period
 
         def on_error(e):
-            LOG.exception("Error executing scheduled alarm", exc_info=e)
+            if LOG.isEnabledFor(logging.DEBUG):
+                LOG.exception("Error executing scheduled alarm", exc_info=e)
+            else:
+                LOG.error("Error executing scheduled alarm")
```
bentsku, line 75: nit: I think the previous implementation was wrong, `e` is an `Exception` here and `exc_info` takes a boolean.

This can be replaced by the regular one we've been using, no need for the `if` check. 
```python
LOG.error("Error executing scheduled alarm", exc_info=LOG.isEnabledFor(logging.DEBUG))
```

It's not blocking, but it sends the wrong message that the `Exception` object needs to be based to `exc_info`

localstack-core/localstack/services/sqs/query_api.py
```diff
@@ -166,7 +166,7 @@ def handle_request(request: Request, region: str) -> Response:
         op = service.operation_model(service.operation_names[0])
         return serializer.serialize_error_to_response(e, op, request.headers, request_id)
     except Exception as e:
-        LOG.exception("exception")
+        LOG.error("exception", exc_info=LOG.isEnabledFor(logging.DEBUG))
```
bentsku, line 169: nit: a bit out of scope of this PR so no need to address this, but this logging message is not going to be very useful is debug is not activated, it will just log `"exception"`. It seems we're properly handling the exception anyway and returning a proper serialized error, so it could maybe be a `debug` logging statement somehow or contain more info, like `LOG.error("Internal Server exception while executing SQS Query operation: '%s'", e, exc_info=LOG.isEnabledFor(logging.DEBUG))` ?

bentsku (approved):
Awesome, nice work! 🚀 

This is quite a lot of logging statements that got updated! This is good to go, I only have 2 small nits, but this can go as is. Kudos for churning through all of them! 

The pipeline failure are fixed with #13004, so you can rebase and we should be good to go 👍

localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
sannya-singal: As per my understanding from the official docs which mention that `exc_info` can accept exception instances, [here](https://docs.python.org/3/library/logging.html), I thought we were doing that intentionally. But if it seems incorrect, I am happy to update this to the simpler version 🙂

localstack-core/localstack/services/sqs/query_api.py
sannya-singal: Agreed 💯 Adding more info would definitely make it more useful 👍 As you mentioned, I will keep it as is for this PR since it is a bit out of scope for this one, but I will create a follow up PR for this 🙂

sannya-singal: Follow PR: https://github.com/localstack/localstack/pull/13007

localstack-core/localstack/services/cloudwatch/alarm_scheduler.py
bentsku: Ah, good catch! Let's leave it like that then 😄 this is good!

PR #12991, opened 2025-08-11
LocalStack SDK for integration tests

giograno:
## Motivation
In https://github.com/localstack/localstack/pull/12972 @viren-nadkarni suggested using the LocalStack SDK for our own integration tests.

## Changes
This PR:
- adds `localstack-sdk-python` to our test dependencies;
- adds a fixture that returns the SDK client;
- reworks the test introduced by @bentsku in #12972 to use the SDK.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,7 @@ test = [
     "aws-cdk-lib>=2.88.0",
     "websocket-client>=1.7.0",
     "localstack-snapshot>=0.1.1",
+    "localstack-sdk-python"
 ]
 
 # for developing localstack
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.12.7
+    rev: v0.12.8
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -116,6 +116,7 @@ test = [
     "aws-cdk-lib>=2.88.0",
     "websocket-client>=1.7.0",
     "localstack-snapshot>=0.1.1",
+    "localstack-sdk-python"
```
alexrashed, line 119: **issue:** This effectively implements a circular dependency:
- The SDK is directly generated from the OpenAPI specs.
- The OpenAPI specs are released with every release of LocalStack Pro.
- The SDK is defined as a dependency of LocalStack Pro's test dependencies here.

In my opinion, this could make it very much more tedious to implement changes when changing the API endpoint or adding new ones.

But what exactly is the _motivation_ behind using the SDK here directly?
- Is it to save some lines of code? Maybe we should then restructure the code or the API to make sure the API is very easily usable.
- Or is it to verify that the SDK works? I think that should be verified directly in the repo of the SDK. If you want to do regular integration tests, you could run a scheduled integration test pipeline there, ensuring that the latest SDK still integrates well with latest?

alexrashed (changes requested):
While it is intriguing to use the SDK directly here to test the features of LocalStack that are exposed via the SDK, I think we have to be very careful with dependencies like this.
Let's discuss this a bit more in depth before introducing this dependency! :)

PR #12997, opened 2025-08-12
CFNV2: Unskip remaining Resolve tests

simonrw:
## Motivation

We have some currently skipped tests relating to "Resolve". We should have mainly implemented resolving with #12934 so these remaining test skips should be relatively minor.

## Changes

* Extract helper function for rendering `EngineParameter`s
* Handle comma delimited lists for SSM parameter parameters using a regex to match whether they are a list type or not, or otherwise treat them as a string
* Unskip more tests


pinzon (approved):
:tada:

PR #13000, opened 2025-08-12
CFNV2: handle AWS::NoValue

simonrw:
## Motivation

We currently model the `AWS::NoValue` intrinsic value by returning `None`, however this is not the right approach. We already perform lots of filtering in e.g. `visit_node_properties` for `Nothing` values, and `AWS::NoValue` has a similar place as returning `Nothing`.

## Changes

* Shortcut the `Ref` resolving if the arguments are `AWS::NoValue` and return `Nothing` in its place
* Add validated test specifically ensuring the correct behaviour (passes on v1 and v2)

### DIFF
--- a/tests/aws/services/cloudformation/engine/test_references.py
+++ b/tests/aws/services/cloudformation/engine/test_references.py
@@ -122,3 +122,28 @@ def test_resolve_transitive_placeholders_in_strings(deploy_cfn_template, aws_cli
         snapshot.transform.regex(r"/cdk-bootstrap/(\w+)/", "/cdk-bootstrap/.../")
     )
     snapshot.match("tags", tags)
+
+
+@markers.aws.validated
+@pytest.mark.parametrize("parameter_value", ["yes", "no"])
+def test_aws_novalue(deploy_cfn_template, parameter_value):
+    """
+    Test that AWS::NoValue is correctly executed in the CFn engine
+    """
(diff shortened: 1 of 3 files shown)


pinzon (approved):
Great change.

simonrw:
> Great change 👍 Nit: it would be nice to see a test for the case when AWS::NoValue is used in a Fn::Sub.

If I try to use an `AWS::NoValue` in a `Sub` expression, e.g.

```yaml
Resources:
  AnotherParameter:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value:
        Fn::Sub:
          - "My great value ${foo}"
          - foo: !Ref AWS::NoValue

```

I get the following error:

```
Template error: every value of the context object of every Fn::Sub object must be a string or a function that returns a string
```

simonrw:
> Great change 👍 Nit: it would be nice to see a test for the case when AWS::NoValue is used in a Fn::Sub.

If I try to use an `AWS::NoValue` in a `Sub` expression, e.g.

```yaml
Resources:
  AnotherParameter:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value:
        Fn::Sub:
          - "My great value ${foo}"
          - foo: !Ref AWS::NoValue

```

I get the following error:

```
Template error: every value of the context object of every Fn::Sub object must be a string or a function that returns a string
```

PR #13001, opened 2025-08-13
fix port subrange from subclass

cloutierMat:
## Motivation

This PR aims to fix `PortRange.subrange` to return an instance of a subclass instead of harcoding `PortRange`.

While working on Elasticache and MemoryDb in container, I realised it was not possible to create more than 1 Redis/Valkey container at once. This was due to `PortRange` not validating that the port is available on the host. Using `_DockerPortRange` would fix the issue, however the `subrange` method would return an instance of `PortRange` instead of `_DockerPortRange`.

## Changes

- use `self.__class__` to retrieve the instance's class
- Add simple unit test

### DIFF
--- a/localstack-core/localstack/utils/net.py
+++ b/localstack-core/localstack/utils/net.py
@@ -343,7 +343,7 @@ class PortRange:
             raise ValueError(f"end not in range ({end} < {self.end})")
 
         # ensures that we return an instance of a subclass
-        port_range = self.__class__(start, end)
+        port_range = type(self)(start, end)
         port_range._ports_cache = self._ports_cache
         port_range._ports_lock = self._ports_lock
         return port_range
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/utils/net.py
```diff
@@ -342,7 +342,8 @@ def subrange(self, start: int = None, end: int = None) -> "PortRange":
         if end > self.end:
             raise ValueError(f"end not in range ({end} < {self.end})")
 
-        port_range = PortRange(start, end)
+        # ensures that we return an instance of a subclass
+        port_range = self.__class__(start, end)
```
dfangl:
```suggestion
        port_range = type(self)(start, end)
```
**suggestion**: I think this is more idiomatic than accessing the dunder class method

PR #13002, opened 2025-08-13
Improve handling of LDM shutdowns during invocations

MEPalma:
## Motivation
The current Lambda version manager invocation logic for LDM does not handle cases correctly when an LDM configuration is removed during invocation. These changes improve the process to ensure such scenarios are handled more gracefully.

## Changes
- Handle canceled errors during LDM invokes

### DIFF
--- a/localstack-core/localstack/services/lambda_/invocation/version_manager.py
+++ b/localstack-core/localstack/services/lambda_/invocation/version_manager.py
@@ -213,12 +213,16 @@ class LambdaVersionManager:
                 self.store_logs(
                     invocation_result=invocation_result, execution_env=ldm_execution_environment
                 )
-            except CancelledError:
+            except CancelledError as e:
                 # Timeouts for invocation futures are managed by LDM, a cancelled error here is
                 # aligned with the debug container terminating whilst debugging/invocation.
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/services/lambda_/invocation/version_manager.py
```diff
@@ -212,6 +213,16 @@ def invoke(self, *, invocation: Invocation) -> InvocationResult:
                 self.store_logs(
                     invocation_result=invocation_result, execution_env=ldm_execution_environment
                 )
+            except CancelledError:
+                # Timeouts for invocation futures are managed by LDM, a cancelled error here is
```
joe4dev, line 217: Should we add a debug log with exception details (stack trace only in debug mode following https://github.com/localstack/localstack/pull/12989), in case we need to troubleshoot cancellations?

dfangl, line 217: We set a timeout, right? it's just a very long one, but it still might have actually run out. Maybe we can include this in the error message?

dfangl (approved):
LGTM, makes sense!

PR #13003, opened 2025-08-13
Create new markers for requiring in process execution or docker

dfangl:
## Motivation
We currently have no way of marking tests as requiring docker or requiring to run in the same process as the LS instance (usually for monkeypatching / pytest http server usage).

This makes certain test scenarios harder, for example testing against a running LS instance, or testing against kubernetes.

With this PR, new markers are introduced:

- `requires_docker`: This marker should be set if a test requires the docker daemon (or a compatible implementation, like podman with docker socket support) to run.
- `requires_in_process`: This marker should be set if a test needs to run inside the same process (or network namespace) as LocalStack. Usually used for tests with `monkeypatch` or `pytest-httpserver`, with the latter hopefully fixed in the future.
- `requires_in_container`: For better distinction between this marker and `requires_docker` this is a renamed `only_in_docker` - it requires LocalStack to run in a container (usually for system level dependency installation), no matter if it is a docker container or kubernetes pod.

## Changes
- Rename `only_in_docker` to `requires_in_container`
- Add markers `requires_docker` and `requires_in_process`
- Add markers to appropriate tests

### DIFF
--- a/localstack-core/localstack/testing/pytest/marking.py
+++ b/localstack-core/localstack/testing/pytest/marking.py
@@ -58,18 +58,23 @@ class Markers:
 
     # test selection
     acceptance_test = pytest.mark.acceptance_test
+    """This test is an acceptance test"""
     skip_offline = pytest.mark.skip_offline
+    """Test is skipped if offline, as it requires some sort of internet connection to run"""
     only_on_amd64 = pytest.mark.only_on_amd64
+    """Test requires ability of the system to execute amd64 binaries"""
     only_on_arm64 = pytest.mark.only_on_arm64
+    """Test requires ability of the system to execute arm64 binaries"""
     resource_heavy = pytest.mark.resource_heavy
-    # Tests require LocalStack to be run inside a container
+    """Test is very resource heavy, and might be skipped in CI"""
     requires_in_container = pytest.mark.requires_in_container
-    # The test and the LS instance have to be run in the same process
+    """Test requires LocalStack to run inside a container"""
     requires_in_process = pytest.mark.requires_in_process
-    # The test requires docker or a compatible container engine - will not work on kubernetes
+    """The test and the LS instance have to be run in the same process"""
     requires_docker = pytest.mark.requires_docker
-    # Tests to execute when updating snapshots for a new Lambda runtime
+    """The test requires docker or a compatible container engine - will not work on kubernetes"""
     lambda_runtime_update = pytest.mark.lambda_runtime_update
+    """Tests to execute when updating snapshots for a new Lambda runtime"""
 
 
 # pytest plugin


localstack-core/localstack/testing/pytest/marking.py
```diff
@@ -62,7 +62,12 @@ class Markers:
     only_on_amd64 = pytest.mark.only_on_amd64
     only_on_arm64 = pytest.mark.only_on_arm64
     resource_heavy = pytest.mark.resource_heavy
-    only_in_docker = pytest.mark.only_in_docker
+    # Tests require LocalStack to be run inside a container
+    requires_in_container = pytest.mark.requires_in_container
```
simonrw, line 70: **suggestion**: is there a way to add a docstring for these? So we can hover over the definition to get this comment? 
```suggestion
    requires_in_container = pytest.mark.requires_in_container
    """Tests require LocalStack to be run inside a container"""
```

This would be good to add to the other markers as well

localstack-core/localstack/testing/pytest/marking.py
```diff
@@ -144,8 +149,8 @@ def filter_by_markers(config: "Config", items: list[pytest.Item]):
     for item in items:
         if is_offline and "skip_offline" in item.keywords:
             item.add_marker(skip_offline)
-        if not is_in_docker and "only_in_docker" in item.keywords:
-            item.add_marker(only_in_docker)
+        if not is_in_docker and "requires_in_container" in item.keywords:
+            item.add_marker(requires_in_container)
         if is_in_ci and not is_amd64 and "only_on_amd64" in item.keywords:
             item.add_marker(only_on_amd64)
         if is_in_ci and not is_arm64 and "only_on_arm64" in item.keywords:
```
simonrw, line 161: **suggestion**: could we do something like:

```python
requires_in_process = pytest.mark.skip(
    reason="..."
)

# ...

if localstack_config.is_env_true("TEST_SKIP_LOCALSTACK_START") and "requires_in_process" in item.keywords:
    item.add_marker(requires_in_process)
```

?

simonrw (approved):
These are a great step forwards to enabling "external" testing for LocalStack, and gives us a TODO list for what we need to update, great!

Just a couple of questions/suggestions

localstack-core/localstack/testing/pytest/marking.py
dfangl: Good point, will add them!

localstack-core/localstack/testing/pytest/marking.py
dfangl: Yes we can, but I am not entirely confident we should, it might get confusing to have these skips. In the k8s pipeline we manually skip them for now, and there was no other usecase until now. I'm happy with adding it, but I think this can wait until we actually encounter a problem this will solve, WDYT?

localstack-core/localstack/testing/pytest/marking.py
dfangl: Added them, and the IDE shows them as expected: 
image

localstack-core/localstack/testing/pytest/marking.py
simonrw: Makes sense. You are right, we don't want to (silently) skip these specific tests without warning the user and the CI will opt in. If we execute tests against a running container and look at the failures, we will clearly see the new marker which will explain why the test fails easily. I'm ok with this 👍

simonrw (commented):
I think a few tests are missing, from my recent investigation into parallelising the CFn tests, I found these three tests require in-process access:

* tests/aws/services/cloudformation/api/test_stacks.py::TestStacksApi::test_create_stack_with_custom_id (requires in memory access to the provider, however it's not super clear. It uses the `set_resource_custom_id` fixture)
* tests/aws/services/cloudformation/resources/test_apigateway.py::test_cfn_apigateway_swagger_import (uses `echo_http_server_post`)
* tests/aws/services/cloudformation/resources/test_lambda.py::test_lambda_cfn_run_with_non_empty_string_replacement_deny_list (uses `monkeypatch`)

Have we verified these markers in this community PR?

dfangl:
They have been verified as part of our `Compile list of services that we think won’t work` backlog task. I did not test the cloudformation tests personally, and you are right, we don't execute them in the pipeline.
I'd vote for moving this to a follow up PR, where we enable the community tests and tag the remaining missing ones, if you are ok with it!

PR #13005, opened 2025-08-13
Migrate from jpype1-ext to jpype1

dfangl:
## Motivation
With the addition of jpype1 last year, we had the issue that jpype would not pass control back to the python process on SIGTERM, but only on SIGINT.

This lead to issues with shutdown, which we fixed in our fork `jpype1-ext`. However, we have since contributed this fix upstream, and jpype1 1.6.0 should have this fix in by default, which was released recently.

With the effort to use python 3.13 as well, this provides a good opportunity to switch back, as `jpype1` provides wheels for python3.13, while `jpype1-ext` currently does not (it would be fairly easy to add, but unnecessary as we want to deprecate the fork anyway).

## Changes
* Move from `jpype1-ext` to `jpype`
* General dependency updates that came in with `make upgrade-pinned-dependencies`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,8 +89,8 @@ runtime = [
     "aws-sam-translator>=1.15.1",
     "crontab>=0.22.6",
     "cryptography>=41.0.5",
-    # allow Python programs full access to Java class libraries. Used for opt-in event ruler.
-    "jpype1-ext>=0.0.1",
+    # allow Python programs full access to Java class libraries. Used for stepfunctions jsonata support
+    "jpype1>=1.6.0",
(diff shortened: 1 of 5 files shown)


dfangl:
Pro run: https://github.com/localstack/localstack-pro/actions/runs/17040740976

alexrashed (approved):
Awesome! Really great that you proactively clean up, and discard the fork now after the patch has been released! 💯 🧹 
Looking forward to Python 3.13! 🤩

dfangl:
Just did some manual testing and SIGTERM handling works as expected :tada:

PR #13006, opened 2025-08-13
introduce `get-release-version` in docker helper

eruditmorina:
## Motivation

- I want to be able to get the release version if `HEAD` is a release tag via `docker-helper.sh`.

## Changes

- Introduce `get-release-version` command which checks if commit is a release, and if so, outputs the version. Otherwise just a message telling that the commit is not a release.

### DIFF
--- a/bin/docker-helper.sh
+++ b/bin/docker-helper.sh
@@ -54,8 +54,7 @@ function _fail {
 function _get_current_version() {
     # check if setuptools_scm is installed, if not prompt to install. python3 is expected to be present
     if ! python3 -m pip -qqq show setuptools_scm > /dev/null ; then
-      echo "ERROR: setuptools_scm is not installed. Run 'pip install --upgrade setuptools setuptools_scm'" >&2
-      exit 1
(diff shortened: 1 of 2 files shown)


eruditmorina:
> Why do you need to source the helper?

@alexrashed Currently we can make use of `_get_current_version` to determine the projects version and `_is_release_commit` to be used to check whether we need to generate release changelog notes. Since both those functions are not exposed, I am proposing the ability to source the script. Another alternative would be to expose these as commands explicitly while being cautious about other functions. Or not relying on `docker-helper.sh` for this use-case at all, but this would speed-up a more straightforward solution. Happy to discuss more!

bin/docker-helper.sh
```diff
@@ -275,6 +278,14 @@ function cmd-push-manifests() {
     fi
 }
 
+function cmd-get-release-version() {
+    if _is_release_commit; then
+        _get_current_version
+    else
+        echo "Not a release commit."
```
alexrashed: **suggestion:** I think this one here should return a non-zero exit code to clearly indicate that this isn't a release commit. That would also make it easier to work with, otherwise you would always have to check the output when using it, right?

PR #13007, opened 2025-08-14
sqs: fix exception log message for sqs query execution

sannya-singal:
## Motivation
This PR aims to improve the log message for executing SQS query operation. Previously, the error handling in the SQS request handler logged errors like this:
`LOG.error("exception", exc_info=LOG.isEnabledFor(logging.DEBUG))`
while this ensured stack traces were only shown in debug mode, the actual log message 

> exception

 was not really useful. 

## Changes
This PR improves the logging message to include more context in the log ensuring that it carries useful information even when stack traces are suppressed when debug mode is disabled.


bentsku (approved):
LGTM, thanks a lot for following up 🚀

PR #13010, opened 2025-08-14
CFNV2: fix remaining transform tests

simonrw:
## Motivation

With #12966 a couple of transform tests were left unskipped

## Changes

* Handle Fn::Transform in the resource position
* Mark the stack as modified if a transform is present (see `test_update_stack_with_same_template_withoutchange_transformation`!)
* Skip two transform tests
* Remove two skipped tests that have no need to exist any more

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -226,7 +226,10 @@ class CloudformationProviderV2(CloudformationProvider):
             given_value = parameters.get(name)
             default_value = parameter.get("Default")
             resolved_parameter = EngineParameter(
-                type_=parameter["Type"], given_value=given_value, default_value=default_value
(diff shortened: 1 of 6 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model.py
```diff
@@ -152,7 +155,7 @@ def to_change_action(self) -> ChangeAction:
 
 class ChangeSetEntity(abc.ABC):
     scope: Final[Scope]
-    change_type: Final[ChangeType]
+    change_type: ChangeType
```
pinzon, line 158: Is this necessary?

simonrw: Yes because we overwrite it in the provider and the type checker complains if we leave the final in

PR #13012, opened 2025-08-15
CFNV2: finish URL replacements for API Gateway

simonrw:
## Motivation

One remaining case where we replace API Gateway urls with LocalStack URLs remains.

Note: this is needed for APIGWv1 because the URL is not an output of the API so we have to adjust it in CFn :( 

> [!NOTE]
> This forms a stack on top of #13011  

## Changes

* Handle URL replacements in Join and Sub

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
@@ -18,6 +18,7 @@ from localstack.services.cloudformation.deployment_utils import log_not_availabl
 from localstack.services.cloudformation.engine.template_deployer import REGEX_OUTPUT_APIGATEWAY
 from localstack.services.cloudformation.engine.v2.change_set_model import (
     NodeDependsOn,
+    NodeIntrinsicFunction,
     NodeOutput,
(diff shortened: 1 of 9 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -585,6 +593,15 @@ def _replace_url_outputs_if_required(value: str) -> str:
 
         return value
 
+    def _replace_url_outputs_in_delta_if_required(
+        self, delta: PreprocEntityDelta
+    ) -> PreprocEntityDelta:
+        if not is_nothing(delta.before) and isinstance(delta.before, str):
+            delta.before = self._replace_url_outputs_if_required(delta.before)
+        if not is_nothing(delta.after) and isinstance(delta.after, str):
+            delta.after = self._replace_url_outputs_if_required(delta.after)
```
pinzon:
```suggestion
        if isinstance(delta.before, str):
            delta.before = self._replace_url_outputs_if_required(delta.before)
        if isinstance(delta.after, str):
            delta.after = self._replace_url_outputs_if_required(delta.after)
```

PR #13017, opened 2025-08-18
extend the Encoder/Decoder state API to be type-aware

thrau:
## Motivation

Currently our decoders serve as adapter only to dill pickle. Because they type information is stored in the pickle, you can de-serialize a pickle and get the correct type without having to specify it at deserialization. Unfortunately this process breaks down when you've moved or renamed a class that you want to deserialize into.

So to be able to de-couple code and data serialization, and separate reader/writer schema, we need the decoder to be type-aware. The Encoder has to be type aware for certain operations as well, for example if we pass a TypedDict to encode, we won't know which type to encode since python doesn't retain type information for typed dicts. So this is just to make the encoding API a more flexible.

## Changes

* state Encoder and Decoder APIs now optionally take a expected_type argument that allows an underlying implementation to be specific about the python type to serialize or deserialize into

### DIFF
--- a/localstack-core/localstack/state/core.py
+++ b/localstack-core/localstack/state/core.py
@@ -97,42 +97,48 @@ class AssetDirectory:
 
 
 class Encoder:
-    def encodes(self, obj: Any) -> bytes:
+    def encodes(self, obj: Any, py_type: type = None) -> bytes:
         """
         Encode an object into bytes.
 
         :param obj: the object to encode
+        :param py_type: the type of the object. needed by some encoders that don't have implicit type knowledge.
         :return: the encoded object
         """
         b = io.BytesIO()
         self.encode(obj, b)
         return b.getvalue()
 
-    def encode(self, obj: Any, file: IO[bytes]):
+    def encode(self, obj: Any, file: IO[bytes], py_type: type = None):
         """
         Encode an object into bytes.
 
         :param obj: the object to encode
+        :param py_type: the type of the object. needed by some encoders that don't have implicit type knowledge.
         :param file: the file to write the encoded data into
         """
         raise NotImplementedError
 
 
 class Decoder:
-    def decodes(self, data: bytes) -> Any:
+    def decodes(self, data: bytes, py_type: type = None) -> Any:
         """
         Decode a previously encoded object.
 
         :param data: the encoded object to decode
+        :param py_type: the type that is expected as return type. Needed by some decoders that don't have implicit
+        type knowledge.
         :return: the decoded object
         """
-        return self.decode(io.BytesIO(data))
+        return self.decode(io.BytesIO(data), py_type)
 
-    def decode(self, file: IO[bytes]) -> Any:
+    def decode(self, file: IO[bytes], py_type: type = None) -> Any:
         """
         Decode a previously encoded object.
 
         :param file: the io object containing the object to decode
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/state/core.py
```diff
@@ -97,42 +97,48 @@ def __str__(self) -> str:
 
 
 class Encoder:
-    def encodes(self, obj: Any) -> bytes:
+    def encodes(self, obj: Any, py_type: type = None) -> bytes:
```
giograno, line 100:
```suggestion
    def encodes(self, obj: Any, py_type: type | None = None) -> bytes:
```
small nit (applies everywhere)

localstack-core/localstack/state/core.py
```diff
@@ -97,42 +97,48 @@ def __str__(self) -> str:
...
         :param data: the encoded object to decode
+        :param py_type: the type that is expected as return type. Needed by some decoders that don't have implicit
+        type knowledge.
         :return: the decoded object
         """
-        return self.decode(io.BytesIO(data))
+        return self.decode(io.BytesIO(data), py_type)
 
-    def decode(self, file: IO[bytes]) -> Any:
+    def decode(self, file: IO[bytes], py_type: type = None) -> Any:
```
giograno, line 135: Following up on our discussion from yesterday, I believe we need to extend the decoder APIs and provide the writer schema. 

The `avro` `DatumReader` needs both the writer and the reader schema (see [here](https://github.com/apache/avro/blob/cd655b20fda31e774cc8549f7e1155ddd9d9b9e6/lang/py/avro/io.py#L602-L606)), and our initial prototype was wrongly providing the reader as the writer. If only the writer schema is provided, `avro` [assumes](https://github.com/apache/avro/blob/cd655b20fda31e774cc8549f7e1155ddd9d9b9e6/lang/py/avro/io.py#L636-L638) the reader schema is the same.
Maybe we could also get along with a `**kwargs` for the initial iteration.

thrau: good callout! i think this is not the right level of abstraction to do this. adding **kwargs just obfuscates the fact that it at this point has become a leaky abstraction.

my suggestion would be to find a different way to hide this. for example, we can make sure in the encoding step that the writer schema is placed into the same file, and we read it first, and then read the data.

giograno: Good point! We can definitely save everything in a single file. `avro` supports this [out of the box](https://github.com/apache/avro/blob/cd655b20fda31e774cc8549f7e1155ddd9d9b9e6/lang/py/avro/datafile.py#L146-L165). I could easily implement it.

Since you mentioned leaky abstractions, couldn't we get rid of `py_type` as well?

thrau: how would we derive the reader schema?

ad: i don't think i really fully understand. could you share an example?

giograno: For the encoder, we can just `type(obj)`. For the encoder, we can read the type from the writer schema we stored in the binary data. The store class might be renamed, though, and we'd need to provide an alias.

giograno (approved):
My main comment was about the writer schema in the decoder signature.
As we discussed, we have been able to store it along with the serialized data in the `avro` implementation (see companion PR), so that's not valid anymore.

PR #13018, opened 2025-08-18
CFNV2: provide logging for deferred actions and cleanup

simonrw:
## Motivation

While debugging I found it useful to log the deffered actions as they happen along with _why_ the action was being run. 

There are also some tests that now pass that haven't been unskipped.

> [!NOTE]
> This PR is part of a stack and builds on #13012

## Changes

* Allow the logging of deferred actions at DEBUG level
* Override select in the describer to propagate known after apply status if the index cannot be accessed
* unskip three more tests 🎉

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_describer.py
+++ b/localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_describer.py
@@ -115,22 +115,23 @@ class ChangeSetModelDescriber(ChangeSetModelPreproc):
     def visit_node_intrinsic_function_fn_select(
         self, node_intrinsic_function: NodeIntrinsicFunction
     ):
+        # TODO: should this not _ALWAYS_ return CHANGESET_KNOWN_AFTER_APPLY?
         arguments_delta = self.visit(node_intrinsic_function.arguments)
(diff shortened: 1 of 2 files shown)


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_preproc.py
```diff
@@ -630,8 +630,12 @@ def _compute_fn_or(args: list[bool]):
     def visit_node_intrinsic_function_fn_not(
         self, node_intrinsic_function: NodeIntrinsicFunction
     ) -> PreprocEntityDelta:
-        def _compute_fn_not(arg: bool) -> bool:
-            return not arg
+        def _compute_fn_not(arg: list[bool] | bool) -> bool:
+            # Is the argument ever a lone boolean?
```
pinzon, line 634: [the docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/intrinsic-function-reference-conditions.html#intrinsic-function-reference-conditions-not) define the argument of a Fn::Not is an array. But let's keep the try/except just in case.

pinzon (approved):
LGTM

Issue #13008, opened 2025-08-14
feature request: SNS Standard topic supports passing "MessageGroupId" for Fair Queues feature


johnlemon93:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature description

Hi team,

Recently, AWS released an update to allow forwarding MessageGroupId from standard SNS topics to SQS standard queues, enabling the Fair Queue feature, as per their article [here](https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-sns-standard-topics-sqs-fair-queues/).

Currently, Localstack SNS does not allow that param to be passed when publishing a message to standard topics. Can we please fix that?

Thanks!

bentsku:
Hello @johnlemon93 and thanks for the report! 

We've introduced support for passing the `MessageGroupId` to non-FIFO topic now, and it should be available in the `latest` image. Could you run `docker pull localstack/localstack:latest` and see if it now works for you? 


Just a small thing I've realized when working with AWS and implementing the feature: for FIFO topics targeting regular and Fair Queues (non-FIFO queues), AWS will **not** forward the `MessageGroupId`, at least for now. This might be a remnant of SNS FIFO Topics adding support for regular Queues subscribers, but wanted to share just in case 😄 

I'll close the issue, feel free to comment again if you still have issues and I'll re-open 👍

PR #13019, opened 2025-08-18
SNS: support passing `MessageGroupId` to non-FIFO topic

bentsku:
## Motivation

Related to #13008 -> https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-sns-standard-topics-sqs-fair-queues/

Since adding support for SQS Fair Queues, we can now pass a `MessageGroupId` to non-FIFO topic.

This PR implements that behavior in SNS, removing the previously raised exception, and properly passing the attribute. 

This also fixes the behavior in SQS itself following #12930, because we were not adding the `MessageGroupId` to the `Attributes` of the received Message, as this only done for FIFO queue. I've updated the SQS test to verify this

## Changes
- remove raising an exception for non-FIFO topics
- remove the check for FIFO topic before passing the attribute as kwarg to `SQS.SendMessage`
- fix the behavior in SQS to return the `MessageGroupId` as `Attributes` of the message
- update SQS test

### DIFF
--- a/localstack-core/localstack/services/sqs/models.py
+++ b/localstack-core/localstack/services/sqs/models.py
@@ -771,7 +771,11 @@ class StandardQueue(SqsQueue):
                 f"request includes a parameter that is not valid for this queue type."
             )
 
-        standard_message = SqsMessage(time.time(), message)
+        standard_message = SqsMessage(
+            time.time(),
+            message,
+            message_group_id=message_group_id,
+        )
 
         if visibility_timeout is not None:
             standard_message.visibility_timeout = visibility_timeout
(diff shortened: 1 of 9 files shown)


localstack-core/localstack/services/sqs/models.py
```diff
@@ -771,7 +771,11 @@ def put(
                 f"request includes a parameter that is not valid for this queue type."
             )
 
-        standard_message = SqsMessage(time.time(), message)
+        standard_message = SqsMessage(
+            time.time(),
+            message,
+            message_group_id=message_group_id,
+        )
```
thrau, line 778: thanks for catching and fixing this!

PR #13021, opened 2025-08-18
CFNV2: SPIKE rework transforms to not use visitors

simonrw:
## Motivation

TBD

## Changes

TBD

> [!NOTE]
> This PR is part of a stack and builds on #13018


simonrw:
Spike did not work out

Issue #12999, opened 2025-08-12
bug: S3 getObjectAttributes does not return checksum when checksum is not first attribute in request


haus:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When sending a get-object-attributes request, the order of attributes seems to matter in the request. If the checksum attribute is first in the list of requested object attributes, it is returned in the response, otherwise it is not.

### Expected Behavior

When calling AWS apis the order does not matter, so even if checksum is not first it should still be returned in results. 

### How are you starting LocalStack?

Custom (please describe below)

#### How are you starting localstack (e.g., `bin/localstack` command, arguments, or `docker-compose.yml`)

Using testcontainers (with docker), though i verified the same behavior when using `bin/localstack`.

#### Client commands (e.g., AWS SDK code snippet, or sequence of "awslocal" commands)

```java
package localstack.reproducer;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.Test;
import org.testcontainers.containers.localstack.LocalStackContainer;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;
import org.testcontainers.utility.DockerImageName;
import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.s3.S3Client;
import software.amazon.awssdk.services.s3.model.GetObjectAttributesRequest;
import software.amazon.awssdk.services.s3.model.GetObjectAttributesResponse;
import software.amazon.awssdk.services.s3.model.ObjectAttributes;
import software.amazon.awssdk.services.s3.model.PutObjectRequest;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.StandardOpenOption;
import java.util.Base64;

import static org.apache.commons.codec.digest.DigestUtils.sha256;
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotNull;
import static org.junit.jupiter.api.Assertions.fail;

@Testcontainers
public class TestTest {
    @Container
    public static final LocalStackContainer localstack = new LocalStackContainer(
        DockerImageName.parse("localstack/localstack:4.7"))
[...]

PR #13022, opened 2025-08-18
S3/ASF parser: fix parsing header `list` shape type with multivalue headers

bentsku:
## Motivation
We got a report with #12999 that sending a `GetObjectAttributes` request would fail depending on the order of the fields being passed.

While debugging and printing out low level HTTP request on the Java side, I realized that Java was sending the request with multiple headers for the same key, like some others SDK (Go in particular)

```
GET /test-key?attributes= HTTP/1.1
[...]
Amz-Sdk-Invocation-Id: 5cb65548-6947-4436-b731-ee6d618bc91c
Amz-Sdk-Request: attempt=1; max=3
Authorization: ---
X-Amz-Date: 20220316T000334Z
X-Amz-Object-Attributes: Checksum
X-Amz-Object-Attributes: ObjectParts
```

When receiving headers that way, our web server will return them as a list of values, accessible via `Headers.getlist()`.

Because the parser was calling `Headers.get()`, it would only get the first value, and in the user sample above, would not even return the second requested attribute. 

This PR adapts the parser to use `getlist` in the header shape type is `list`, and keep the old automatic splitting on `,` that is still required. 

## Changes
- add a parser unit test (thanks a lot for adding the `headers` override! This was so much easier to write a test for it!!!)
- fix the parsing logic to handle multivalue headers
- add an S3 test around multivalue headers (`requests` concatenates multivalue headers by default but `urllib3` does not-
- add an order S3 test because my first intuition was related to the order of returned fields

## Testing

If you remove the parser fix, the tests now fail.
I also validated that the user-provided sample in the linked issue now works 👍

### DIFF
--- a/localstack-core/localstack/aws/protocol/parser.py
+++ b/localstack-core/localstack/aws/protocol/parser.py
@@ -234,13 +234,23 @@ class RequestParser(abc.ABC):
         if location is not None:
             if location == "header":
                 header_name = shape.serialization.get("name")
-                payload = request.headers.get(header_name)
-                if payload and shape.type_name == "list":
+                if shape.type_name == "list":
                     # headers may contain a comma separated list of values (e.g., the ObjectAttributes member in
                     # s3.GetObjectAttributes), so we prepare it here for the handler, which will be `_parse_list`.
                     # Header lists can contain optional whitespace, so we strip it
                     # https://www.rfc-editor.org/rfc/rfc9110.html#name-lists-rule-abnf-extension
-                    payload = [value.strip() for value in payload.split(",")]
+                    # It can also directly contain a list of headers
+                    payload = request.headers.getlist(header_name)
+                    if len(payload) == 1:
+                        header_values = payload[0].split(",")
+                        payload = [value.strip() for value in header_values]
+
+                    # we set the payload to None in case the list in empty, as `getlist` returns an empty list
+                    if not payload:
+                        payload = None
+                else:
+                    payload = request.headers.get(header_name)
+
             elif location == "headers":
                 payload = self._parse_header_map(shape, request.headers)
                 # shapes with the location trait "headers" only contain strings and are not further processed
--- a/localstack-core/localstack/aws/protocol/parser.py
+++ b/localstack-core/localstack/aws/protocol/parser.py
@@ -240,14 +240,12 @@ class RequestParser(abc.ABC):
                     # Header lists can contain optional whitespace, so we strip it
                     # https://www.rfc-editor.org/rfc/rfc9110.html#name-lists-rule-abnf-extension
                     # It can also directly contain a list of headers
+                    # See https://datatracker.ietf.org/doc/html/rfc2616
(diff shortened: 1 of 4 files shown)


localstack-core/localstack/aws/protocol/parser.py
```diff
@@ -234,13 +234,23 @@ def _parse_shape(
...
                 header_name = shape.serialization.get("name")
-                payload = request.headers.get(header_name)
-                if payload and shape.type_name == "list":
+                if shape.type_name == "list":
                     # headers may contain a comma separated list of values (e.g., the ObjectAttributes member in
                     # s3.GetObjectAttributes), so we prepare it here for the handler, which will be `_parse_list`.
                     # Header lists can contain optional whitespace, so we strip it
                     # https://www.rfc-editor.org/rfc/rfc9110.html#name-lists-rule-abnf-extension
-                    payload = [value.strip() for value in payload.split(",")]
+                    # It can also directly contain a list of headers
```
alexrashed, line 242: **thought (non-blocking):** Is it also possible that there are multiple headers with the same key, some contain a single value, and some contain multiple values (comma-separated)?

bentsku: I also thought of this one 👀 I was thinking about splitting all the individual values in the list over `,` at first. It's a bit tricky, because I know for example `Set-Cookie` (which was one of the first case I had with multi values headers in https://github.com/localstack/rolo/pull/29) can contain commas in the value, which is why they are sending it a multiple values with the same key instead of a comma-separated single header.  

The [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616) states the following:

> Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].
   It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma. The order in which header fields with the same
   field-name are received is therefore significant to the
   interpretation of the combined field value, and thus a proxy MUST NOT
   change the order of these field values when a message is forwarded.

So it's not super clear either. I'll give it a try, and depending on the result will update or merge.

bentsku: I've tried with `curl`, which treats multi value headers properly but couldn't get it to run against AWS because `GetObjectAttributes` needs to be signed, so I've also tried with [`awscurl`](https://github.com/okigan/awscurl) but it doesn't manage multi value headers properly and just overwrite with the last one specified 😕

But reading the RFC, I'm thinking of actually joining all returned values from `getlist` over `,`, then splitting them like the previous behavior, following this particular part: `It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma.`

I'll update, I think this is probably never gonna happen, but it might simplify the code a little bit :+1:

PR #13026, opened 2025-08-19
IaC: AWS and CloudFormation features support catalog

k-a-il:
## Motivation

Create initial structure for AWS and CloudFormation catalogs.

## Changes

- Adds `AwsServicesCatalog` and `CloudFormationCatalog`
- Logic will be added in next PRs


localstack-core/localstack/utils/catalog/aws_catalog.py
```diff
@@ -0,0 +1,30 @@
...
+from localstack.utils.objects import singleton_factory
+
+LOG = logging.getLogger(__name__)
+
+
+class AwsServicesCatalog:
+    @staticmethod
+    @singleton_factory
+    def get() -> "AwsServicesCatalog":
+        return AwsServicesCatalog()
```
bentsku, line 16: nit: I commented on the wrong PR, but I think this might be better encapsulated in a factory function outside of the class, so it'd be easier to change the factory behavior to return a different class. Here, we're stuck with `AwsServicesCatalog` with maybe no way to extend it from Pro
Sorry, this might have been my suggestion 😕

k-a-il: no problem, good point :)

k-a-il:
Closing this PR because it is part of https://github.com/localstack/localstack/pull/13027

PR #13027, opened 2025-08-19
IaC: Add AWS operations and CFN resources catalog

k-a-il:
## Motivation

This PR introduces a plugin AWS license catalog which will be used to customize error messages for unimplemented AWS services or operations.

Downloading and loading of the license catalog is not implemented in this PR and will be completed in the next PR.

## Changes

- Add new license catalog plugin

### DIFF
--- a/localstack-core/localstack/aws/handlers/service.py
+++ b/localstack-core/localstack/aws/handlers/service.py
@@ -9,8 +9,8 @@ from botocore.model import OperationModel, ServiceModel
 
 from localstack import config
 from localstack.http import Response
-from localstack.utils.coverage_docs import get_coverage_link_for_service
 
+from ...utils.coverage_docs import get_coverage_link_for_service
 from ..api import CommonServiceException, RequestContext, ServiceException
 from ..api.core import ServiceOperation
 from ..chain import CompositeResponseHandler, ExceptionHandler, Handler, HandlerChain
@@ -176,8 +176,8 @@ class ServiceExceptionSerializer(ExceptionHandler):
             action_name = operation.name
             exception_message: str | None = exception.args[0] if exception.args else None
             message = exception_message or get_coverage_link_for_service(service_name, action_name)
-            LOG.info(message)
             error = CommonServiceException("InternalFailure", message, status_code=501)
+            LOG.info(message)
             context.service_exception = error
 
         elif not isinstance(exception, ServiceException):
--- a/localstack-core/localstack/utils/catalog/catalog.py
+++ b/localstack-core/localstack/utils/catalog/catalog.py
@@ -1,0 +1,135 @@
+import logging
+from abc import abstractmethod
+
+from plux import Plugin
+
+from localstack.services.cloudformation.resource_provider import (
+    plugin_manager as cfn_plugin_manager,
+)
+from localstack.utils.catalog.catalog_loader import RemoteCatalogLoader
+from localstack.utils.catalog.common import (
+    AwsServiceOperationsSupportInLatest,
+    AwsServicesSupportInLatest,
+    AwsServiceSupportAtRuntime,
+    CloudFormationResourcesSupportAtRuntime,
+    CloudFormationResourcesSupportInLatest,
+    LocalstackEmulatorType,
+)
+
+ServiceName = str
+ServiceOperations = set[str]
+ProviderName = str
+CfnResourceName = str
+CfnResourceMethodName = str
+AwsServicesSupportStatus = (
+    AwsServiceSupportAtRuntime | AwsServicesSupportInLatest | AwsServiceOperationsSupportInLatest
+)
+CfnResourceSupportStatus = (
+    CloudFormationResourcesSupportInLatest | CloudFormationResourcesSupportAtRuntime
+)
+
+LOG = logging.getLogger(__name__)
+
+
+class CatalogPlugin(Plugin):
+    namespace = "localstack.utils.catalog"
+
+    @staticmethod
+    def get_cfn_resources_catalog(cloudformation_resources: dict):
+        cfn_resources_catalog = {}
+        for emulator_type, resources in cloudformation_resources.items():
+            for resource_name, resource in resources.items():
+                cfn_resources_catalog[emulator_type] = {resource_name: set(resource.methods)}
+        return cfn_resources_catalog
+
+    @staticmethod
+    def get_aws_services_at_runtime():
+        from localstack.services.plugins import SERVICE_PLUGINS
+
+        return SERVICE_PLUGINS.list_available()
+
+    @abstractmethod
+    def get_aws_service_status(
+        self, service_name: str, operation_name: str | None = None
+    ) -> AwsServicesSupportStatus | None:
+        pass
+
+    @abstractmethod
+    def get_cloudformation_resource_status(
+        self, resource_name: str, service_name: str
+    ) -> CfnResourceSupportStatus | AwsServicesSupportInLatest | None:
+        pass
+
+
+class AwsCatalogRuntimePlugin(CatalogPlugin):
+    name = "aws-catalog-runtime-only"
+
+    def get_aws_service_status(
+        self, service_name: str, operation_name: str | None = None
+    ) -> AwsServicesSupportStatus | None:
+        return None
+
+    def get_cloudformation_resource_status(
+        self, resource_name: str, service_name: str
+    ) -> CfnResourceSupportStatus | AwsServicesSupportInLatest | None:
+        return None
+
+
+class AwsCatalogRemoteStatePlugin(CatalogPlugin):
+    name = "aws-catalog-remote-state"
+    current_emulator_type: LocalstackEmulatorType = LocalstackEmulatorType.COMMUNITY
+    services_in_latest: dict[ServiceName, dict[LocalstackEmulatorType, ServiceOperations]] = {}
+    services_at_runtime: set[ServiceName] = set()
+    cfn_resources_in_latest: dict[
+        LocalstackEmulatorType, dict[CfnResourceName, set[CfnResourceMethodName]]
+    ] = {}
+    cfn_resources_at_runtime: set[CfnResourceName] = set()
+
+    def __init__(self, remote_catalog_loader: RemoteCatalogLoader | None = None) -> None:
+        catalog_loader = remote_catalog_loader or RemoteCatalogLoader()
+        remote_catalog = catalog_loader.get_remote_catalog()
+        for service_name, emulators in remote_catalog.services.items():
+            for emulator_type, service_provider in emulators.items():
+                self.services_in_latest.setdefault(service_name, {})[emulator_type] = set(
+                    service_provider.operations
+                )
+
+        self.cfn_resources_in_latest = AwsCatalogRemoteStatePlugin.get_cfn_resources_catalog(
+            remote_catalog.cloudformation_resources
+        )
+        self.cfn_resources_at_runtime = set(cfn_plugin_manager.list_names())
+        self.services_at_runtime = AwsCatalogRemoteStatePlugin.get_aws_services_at_runtime()
+
+    def get_aws_service_status(
+        self, service_name: str, operation_name: str | None = None
+    ) -> AwsServicesSupportStatus | None:
+        if not self.services_in_latest:
+            return None
+        if service_name not in self.services_in_latest:
+            return AwsServicesSupportInLatest.NOT_SUPPORTED
+        if self.current_emulator_type not in self.services_in_latest[service_name]:
+            return AwsServicesSupportInLatest.SUPPORTED_WITH_LICENSE_UPGRADE
+        if not operation_name:
+            return AwsServicesSupportInLatest.SUPPORTED
+        if operation_name in self.services_in_latest[service_name][self.current_emulator_type]:
+            return AwsServiceOperationsSupportInLatest.SUPPORTED
+        for emulator_type in self.services_in_latest[service_name]:
+            if emulator_type is self.current_emulator_type:
+                continue
+            if operation_name in self.services_in_latest[service_name][emulator_type]:
+                return AwsServiceOperationsSupportInLatest.SUPPORTED_WITH_LICENSE_UPGRADE
+        return AwsServiceOperationsSupportInLatest.NOT_SUPPORTED
+
+    def get_cloudformation_resource_status(
+        self, resource_name: str, service_name: str
+    ) -> CfnResourceSupportStatus | AwsServicesSupportInLatest | None:
+        if resource_name in self.cfn_resources_at_runtime:
+            return CloudFormationResourcesSupportAtRuntime.AVAILABLE
+        if service_name in self.services_at_runtime:
+            if resource_name in self.cfn_resources_in_latest[self.current_emulator_type]:
+                return CloudFormationResourcesSupportInLatest.SUPPORTED
+            else:
+                return CloudFormationResourcesSupportInLatest.NOT_SUPPORTED
+        if service_name in self.services_in_latest:
+            return self.get_aws_service_status(service_name, operation_name=None)
+        return AwsServicesSupportInLatest.NOT_SUPPORTED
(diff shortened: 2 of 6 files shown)


localstack-core/localstack/aws/skeleton.py
bentsku, line 1: I wonder if we should still catch exceptions in the `skeleton` instead of letting them bubbling up to the proper exception handlers. 

This doesn't need to be acted upon in this PR in particular, but it feels a bit weird to have the logic duplicated.

localstack-core/localstack/aws/catalog_exceptions.py
```diff
@@ -0,0 +1,65 @@
...
+            message = f"Sorry, the {operation_name} operation on the {service_name} service is not supported with your LocalStack license. Please refer to {_DOCS_COVERAGE_URL} for more details."
+            error_code = 5
+        super().__init__(message, error_code)
+
+
+def map_catalog_availability_to_exception(
+    service_name: str,
+    operation_name: str | None,
+    support_status: AwsServicesSupportInLatest | AwsServiceOperationsSupportInLatest,
+) -> AwsServiceAvailabilityException:
```
bentsku: question: how will we extend this with more possible values in the Pro version? we might want to aim for it to be extended transparently, as we won't be able to plug into the `Skeleton.on_not_implemented_error` function directly to modify which function is being called. 

We could probably patch it, and execute some code before deferring the logic to this function? maybe there are better ways. Patching might do, we're also using it for our AWS routing.

localstack-core/localstack/aws/skeleton.py
```diff
@@ -213,13 +214,31 @@ def on_not_implemented_error(
...
+            message = exception_message
+            error = CommonServiceException("InternalFailure", message, status_code=501)
+            # record event
+            analytics.log.event(
+                "services_notimplemented", payload={"s": service_name, "a": action_name}
+            )
+        else:
+            support_in_latest = AwsServicesCatalog.get().get_support_status(
+                service_name, action_name
+            )
```
bentsku: rethinking this, we might want to encapsulate it instead in a factory function like `get_licensed_environment()` so we could decide which class to return (community vs pro)? 

Something like:
```python
@singleton_factory
def get_aws_catalog() -> AwsServicesCatalog:
    return AwsServicesCatalog()
```

So that we're free to introduce more logic in the factory to return what we need (be it a plugin or any other mechanism that we could extend from Pro)

bentsku (commented):
This is nice 👌 super nice to see how easy it is to plug into this! this is only a pre-review as I got the notification, I'm still unsure how we are going to extend the behavior from Pro. Might be good to try it already to see how we could do it, and adapt the code in consequence? You might be on it already 😄

k-a-il:
> This is nice 👌 super nice to see how easy it is to plug into this! this is only a pre-review as I got the notification, I'm still unsure how we are going to extend the behavior from Pro. Might be good to try it already to see how we could do it, and adapt the code in consequence? You might be on it already 😄

Thanks, yes, I will extend this in pro tomorrow :) I also want to see how the whole integration will work before merging this PR

localstack-core/localstack/utils/catalog/aws_catalog.py
```diff
@@ -1,30 +1,10 @@
...
-    @staticmethod
-    def _build_catalog() -> dict[str, dict[str, dict]]:
-        raise NotImplementedError()
-
-    def get_support_status(
-        self, service_name: str, operation_name: str | None = None
-    ) -> AwsServicesSupportInLatest | AwsServiceOperationsSupportInLatest:
-        # TODO: firstly checks if service is supported, after that if operation is supported
-        raise NotImplementedError()
+        return AwsCatalog
```
bentsku: question: would it be easier to return a class instance here, so you don't need to do `CATALOG_PLUGIN_MANAGER.load(plugin_name).load()()` but can return `CATALOG_PLUGIN_MANAGER.load(plugin_name).load()` instead? or was there a specific reason for it?

localstack-core/localstack/aws/skeleton.py
k-a-il: Will be addressed in this PR https://github.com/localstack/localstack/pull/13068

localstack-core/localstack/aws/handlers/service.py
```diff
@@ -175,9 +177,20 @@ def create_exception_response(self, exception: Exception, context: RequestContex
...
+                )
+                error = CommonServiceException("InternalFailure", message, status_code=501)
+            else:
+                support_in_latest = get_aws_catalog().get_aws_service_status(
+                    service_name, action_name
+                )
+                error = map_catalog_availability_to_exception(
+                    service_name, action_name, support_in_latest
+                )
+                message = error.message
```
silv-io: question: I'm a bit confused by this. We only add the catalog info if there is no exception message already? Is this there to leave it open to service implementers to give custom messages in `NotImplementedError`s?

localstack-core/localstack/utils/catalog/catalog.py
```diff
@@ -0,0 +1,140 @@
...
+        cfn_resources_catalog = {}
+        for emulator_type in list(LocalstackEmulatorType):
+            if emulator_type in cloudformation_resources:
+                for resource_name in cloudformation_resources[emulator_type].keys():
+                    cfn_resources_catalog[emulator_type] = {
+                        resource_name: set(
+                            cloudformation_resources[emulator_type][resource_name]["methods"]
+                        )
+                    }
+        return cfn_resources_catalog
```
silv-io: question: Is this to make sure that we only allow `community` and `pro` as the types?

Otherwise we could just simplify to:


```suggestion
    def get_cfn_resources_catalog(cloudformation_resources: dict):
        cfn_resources_catalog = {}
        for emulator_type, emulator in cloudformation_resources.items():
            for resource_name, resource in emulator.items():
                cfn_resources_catalog[emulator_type] = {
                    resource_name: set(
                            resource["methods"]
                    )
                }
        return cfn_resources_catalog
```

and lose a bit of complexity here

localstack-core/localstack/utils/catalog/catalog.py
```diff
@@ -0,0 +1,140 @@
...
+            for provider_name in SERVICE_PLUGINS.plugin_manager.list_names()
+        }
+
+    def __init__(self, remote_catalog_loader: RemoteCatalogLoader | None = None) -> None:
+        catalog_loader = remote_catalog_loader or RemoteCatalogLoader()
+        remote_catalog = catalog_loader.get_remote_catalog()
+        for service_name in remote_catalog.services.keys():
+            service = remote_catalog.services[service_name]
+            for emulator_type in list(LocalstackEmulatorType):
+                if emulator_type in service:
```
silv-io: same here with the emulator type

localstack-core/localstack/utils/catalog/plugins.py
```diff
@@ -0,0 +1,13 @@
...
+from localstack.utils.catalog.catalog import CatalogPlugin
+from localstack.utils.objects import singleton_factory
+
+
+@singleton_factory
+def get_aws_catalog() -> CatalogPlugin:
+    plugin_manager = PluginManager(CatalogPlugin.namespace)
+    plugin_name = "aws_catalog_pro"
+    if not plugin_manager.exists(plugin_name):
+        plugin_name = "aws_catalog"
```
silv-io: nit: I think usually we named plugins with `-` not `_`

localstack-core/localstack/utils/catalog/loader.py
```diff
@@ -0,0 +1,20 @@
...
+from pydantic import BaseModel
+
+LICENSE_CATALOG_PATH = ""
+
+
+class AwsRemoteCatalog(BaseModel):
+    schema_version: str
+    localstack: dict[str, str]
+    services: dict[str, dict]
+    cloudformation_resources: dict[str, dict]
```
silv-io: nit: We could make the typing here more explicit, as we assume it to be so in subsequent checks. (e.g., direct references to `"methods"`)

silv-io (approved):
LGTM! It's all very clean and easily understandable :) 

I basically only have super minor nitpicks and some conceptual questions

localstack-core/localstack/aws/handlers/service.py
bentsku: Yes, exactly, like if you want to manually raise `NotImplemented("Memcached is not a support backend yet")`, then we shouldn't override that message I think

localstack-core/localstack/utils/catalog/loader.py
bentsku: agreed, maybe some TypedDict/other kind of model to indicate the expected schema? but it would also change access to those dicts

localstack-core/localstack/aws/catalog_exceptions.py
bentsku, line 1: neat and clean 👌

localstack-core/localstack/utils/catalog/catalog.py
```diff
@@ -0,0 +1,140 @@
...
+            return None
+        if service_name not in self.services_in_latest:
+            return AwsServicesSupportInLatest.NOT_SUPPORTED
+        if self.current_emulator_type not in self.services_in_latest[service_name]:
+            return AwsServicesSupportInLatest.SUPPORTED_WITH_LICENSE_UPGRADE
+        if not operation_name:
+            return AwsServicesSupportInLatest.SUPPORTED
+        if operation_name in self.services_in_latest[service_name][self.current_emulator_type]:
+            return AwsServiceOperationsSupportInLatest.SUPPORTED
+        for emulator_type in self.services_in_latest[service_name].keys():
```
bentsku: nit: to verify if something is in a dictionary, it's better to not call `keys()` which would create a dict view. 
Prefer `for emulator_type in self.services_in_latest[service_name]` instead

localstack-core/localstack/utils/catalog/catalog.py
```diff
@@ -0,0 +1,140 @@
...
+            for emulator_type in list(LocalstackEmulatorType):
+                if resource_name in self.cfn_resources_in_latest[emulator_type]:
+                    if emulator_type is self.current_emulator_type:
+                        return CloudFormationResourcesSupportInLatest.SUPPORTED
+                    else:
+                        return CloudFormationResourcesSupportInLatest.SUPPORTED_WITH_LICENSE_UPGRADE
+            return CloudFormationResourcesSupportInLatest.NOT_SUPPORTED
+        if service_name in self.services_in_latest:
+            return self.get_aws_service_status(service_name, operation_name=None)
+        return AwsServiceOperationsSupportInLatest.NOT_SUPPORTED
```
bentsku: question: should this value be `AwsServicesSupportInLatest.NOT_SUPPORTED` instead? As it does not relate to `Operations`?

tests/unit/utils/test_catalog.py
bentsku, line 1: nice set of test! 🚀

localstack-core/localstack/utils/catalog/loader.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class RemoteCatalogLoader:
+    def get_remote_catalog(self) -> AwsRemoteCatalog:
+        with open(LICENSE_CATALOG_PATH) as f:
+            return AwsRemoteCatalog(
+                **json.load(f)
+            )  # TODO: catch exceptions and fallback to default catalog
```
bentsku: I believe we need to catch the exception right now before merging, right? because right now if we hit a `501`, we'll try calling `get_aws_catalog()` which will trigger the loader which will fail and everything will fail? it could also be caught when instantiating the `RemoteCatalogLoader` in `AwsCatalogPlugin.__init__`, but I think it needs to be done otherwise we'd break things.

localstack-core/localstack/utils/catalog/aws_catalog.py
k-a-il: We discussed this together and this was resolved

bentsku: Oops yes sorry this was a leftover comment! Forgot to remove it 😅

localstack-core/localstack/aws/handlers/service.py
k-a-il: yes, I think we should still support custom messages as mentioned above

localstack-core/localstack/utils/catalog/catalog.py
k-a-il: True, thanks, this is a mistake 👍

localstack-core/localstack/utils/catalog/catalog.py
k-a-il: My bad, will fix

localstack-core/localstack/utils/catalog/loader.py
k-a-il: yeah, makes sense, will add this 👍

localstack-core/localstack/utils/catalog/catalog.py
k-a-il: > Is this to make sure that we only allow community and pro as the types?

Yes, initially the idea was to allow only pro and community emulators defined in the enum. But after completing the catalog in pro I think this is unnecessary complexity, will adapt code

localstack-core/localstack/utils/catalog/loader.py
k-a-il: Thanks, I've added new plugin with runtime state only and fallback to it in case initialization of catalogs with remote state fails

k-a-il:
I removed integration to service and skeleton to merge this PR faster

localstack-core/localstack/utils/catalog/catalog.py
```diff
@@ -0,0 +1,135 @@
...
+            for emulator_type, service_provider in emulators.items():
+                self.services_in_latest.setdefault(service_name, {})[emulator_type] = set(
+                    service_provider.operations
+                )
+
+        self.cfn_resources_in_latest = AwsCatalogRemoteStatePlugin.get_cfn_resources_catalog(
+            remote_catalog.cloudformation_resources
+        )
+        self.cfn_resources_at_runtime = set(cfn_plugin_manager.list_names())
+        self.services_at_runtime = AwsCatalogRemoteStatePlugin.get_aws_services_at_runtime()
```
bentsku: question: would it make more sense to call those 2 methods via the `self` instance, even if they are static methods?

localstack-core/localstack/utils/catalog/plugins.py
```diff
@@ -0,0 +1,24 @@
...
+        plugin_name = "aws-catalog-remote-state-with-license"
+        if not plugin_manager.exists(plugin_name):
+            plugin_name = "aws-catalog-remote-state"
+        return plugin_manager.load(plugin_name)
+    except Exception as e:
+        LOG.debug(
+            "Failed to load catalog plugin with the latest LocalStack services support data, falling back to catalog without remote state: %s",
+            e,
+        )
+        return plugin_manager.load("aws-catalog-runtime-only")
```
bentsku: question: should we maybe have a different fallback for the Pro version? 

Currently, if I understood well, the Runtime-only fallback doesn't do anything in Community, because it cannot access any licensing data, and won't provide additional features at all. I can see how this works for the 501 error, it will just fallback to what we had before. 

For CloudFormation, I'm not sure what the behavior will be if we return `None`, but I'm sure we can handle it, so for now I don't see a big issue. 

However, for the Pro "runtime fallback", we will have additional data even without the remote catalog: for example, if the service is in another tier, etc. So there we should probably share code for the runtime part between the 2 catalogs (remote and runtime-only). Not sure if I'm making sense here 😄

bentsku: something I missed: I believe we should maybe raise a specific exception we would have subclassed ourselves in case we have a failure in the `RemoteCatalogLoader`, so that we could probably guard only against this? because here, if we have any kind of exceptions when instantiating the Catalog, we would fallback, even if the exception is something we should fix, like a Key error or something?

k-a-il: yes, this is actually true.It makes sense for CloudFormation, I will add different fallback to runtime catalog for pro.

k-a-il: Fixed 👍

k-a-il: Notes from discussion: agreed that for now it is ok to fall back to the default runtime catalog in case any exception is thrown during catalog initialization.

PR #13030, opened 2025-08-19
APIGW: fix TestInvokeMethod path logic

bentsku:
## Motivation
We got a report that the `TestInvokeMethod` would not work when the `pathWithQueryString` was not provided. After testing in AWS, it showed that this parameter was not required and this was an assumption that stayed from the previous implementation. 

This PR implements more tests around this parameter and the behavior of AWS when it is not provided. 

It also showed AWS is not sending the same `log` response when the integration is of type `MOCK`. 

I took the opportunity to refactor a bit the utilities we have around `TestInvokeMethod`, because it is quite tricky to snapshot match as it returns a big string with newlines that we would like to snapshot match more precisely. 

## Changes
- fix the behavior of `TestInvokeMethod` when `pathWithQueryString` is not provided: it will fallback to the `Resource` `path` value
- fix the behavior when it cannot extract the path parameters from the provided path: it won't fail but silently set the path parameters to an empty map
- add a new log template for `MOCK` integration, this should probably be revisited at some point
- improve the existing tests and add more cases
- refactor the test utilities
- remove overly verbose stack trace logging when debug is enabled when we don't match a route in API Gateway REST API router

### DIFF
--- a/localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
+++ b/localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
@@ -45,6 +45,19 @@ TEST_INVOKE_TEMPLATE = """Execution log for request {request_id}
 {formatted_date} : Method completed with status: {method_response_status}
 """
 
+TEST_INVOKE_TEMPLATE_MOCK = """Execution log for request {request_id}
+{formatted_date} : Starting execution for request: {request_id}
+{formatted_date} : HTTP Method: {request_method}, Resource Path: {resource_path}
+{formatted_date} : Method request path: {method_request_path_parameters}
+{formatted_date} : Method request query string: {method_request_query_string}
+{formatted_date} : Method request headers: {method_request_headers}
+{formatted_date} : Method request body before transformations: {method_request_body}
+{formatted_date} : Method response body after transformations: {method_response_body}
+{formatted_date} : Method response headers: {method_response_headers}
+{formatted_date} : Successfully completed execution
+{formatted_date} : Method completed with status: {method_response_status}
+"""
+
 
 def _dump_headers(headers: Headers) -> str:
     if not headers:
@@ -93,6 +106,29 @@ def log_template(invocation_context: RestApiInvocationContext, response_headers:
     )
 
 
+def log_mock_template(
+    invocation_context: RestApiInvocationContext, response_headers: Headers
+) -> str:
+    formatted_date = datetime.datetime.now(tz=datetime.UTC).strftime("%a %b %d %H:%M:%S %Z %Y")
+    request = invocation_context.invocation_request
+    context_var = invocation_context.context_variables
+    method_resp = invocation_context.invocation_response
+
+    return TEST_INVOKE_TEMPLATE_MOCK.format(
+        formatted_date=formatted_date,
+        request_id=context_var["requestId"],
+        resource_path=request["path"],
+        request_method=request["http_method"],
+        method_request_path_parameters=dict_to_string(request["path_parameters"]),
+        method_request_query_string=dict_to_string(request["query_string_parameters"]),
+        method_request_headers=_dump_headers(request.get("headers")),
+        method_request_body=to_str(request.get("body", "")),
+        method_response_status=method_resp.get("status_code"),
+        method_response_body=to_str(method_resp.get("body", "")),
+        method_response_headers=_dump_headers(response_headers),
+    )
+
+
 def create_test_chain() -> HandlerChain[RestApiInvocationContext]:
     return HandlerChain(
         request_handlers=[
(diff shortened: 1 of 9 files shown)


localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
```diff
@@ -114,13 +150,16 @@ def create_test_invocation_context(
...
+    resource = deployment.rest_api.resources[test_request["resourceId"]]
+    resource_path = resource["path"]
 
     # we do not need a true HTTP request for the context, as we are skipping all the parsing steps and using the
     # provider data
     invocation_context = RestApiInvocationContext(
         request=Request(method=http_method),
     )
-    path_query = test_request.get("pathWithQueryString", "/").split("?")
+    test_request_path = test_request.get("pathWithQueryString") or resource_path
```
bentsku, line 165: this single line is effectively the fix for it

localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
```diff
@@ -140,9 +179,15 @@ def create_test_invocation_context(
...
         body=to_bytes(test_request.get("body") or ""),
     )
+
     invocation_context.invocation_request = invocation_request
+    try:
+        # this is AWS behavior, it will accept any value for the `pathWithQueryString`, even if it doesn't match
+        # the expected format. It will just fall back to no value if it cannot parse the path parameters out of it
+        _, path_parameters = RestAPIResourceRouter(deployment).match(invocation_context)
+    except Exception:
+        path_parameters = {}
```
bentsku: this is also part of the fix, due to the ability to pass invalid values, but it was not strictly needed for the sample to work

localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
```diff
@@ -140,9 +179,15 @@ def create_test_invocation_context(
         # TODO: handle multiValueHeaders
         body=to_bytes(test_request.get("body") or ""),
     )
+
     invocation_context.invocation_request = invocation_request
+    try:
+        # this is AWS behavior, it will accept any value for the `pathWithQueryString`, even if it doesn't match
+        # the expected format. It will just fall back to no value if it cannot parse the path parameters out of it
+        _, path_parameters = RestAPIResourceRouter(deployment).match(invocation_context)
+    except Exception:
```
joe4dev: Are we ok swalling the exception here, or are there scenarios where debug logs would be necessary to understand a bug in `.match`?

localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
```diff
@@ -45,6 +45,19 @@
 {formatted_date} : Method completed with status: {method_response_status}
 """
 
+TEST_INVOKE_TEMPLATE_MOCK = """Execution log for request {request_id}
```
joe4dev, line 52: praise: neat detail considering these verbose mock logs ✨

tests/aws/services/apigateway/conftest.py
```diff
@@ -238,3 +240,60 @@ def _import_apigateway_function(*args, **kwargs):
...
+    )
+
+    def _transform_log(_log: str) -> dict[str, str]:
+        return {f"line{index:02d}": line for index, line in enumerate(_log.split("\n"))}
+
+    counter = count(start=1)
+
+    # we want to do very precise matching on the log, and splitting on new lines will help in case the snapshot
+    # fails
+    # the snapshot library does not allow to ignore an array index as the last node, so we need to put it in a dict
```
joe4dev, line 289: FYI: @tiurin snapshot library limitation to consider tracking

tests/aws/services/apigateway/test_apigateway_api.py
```diff
@@ -2718,60 +2759,124 @@ def invoke_method(api_id, resource_id, path_with_query_string, method, body=""):
...
             invoke_method,
             retries=10,
             sleep=5,
-            api_id=api_id,
-            resource_id=resource_id,
+            api_id=rest_api_id,
+            target_resource_id=resource_id,
+            method="GET",
+        )
+        # assert "HTTP Method: GET, Resource Path: /pets/123" in response["log"]
```
joe4dev: nit: Is this comment intentional (same below)?

localstack-core/localstack/services/apigateway/next_gen/execute_api/test_invoke.py
bentsku: Yes, I thought we're fine, the routing step is already done and we're using it only to extract the path parameters. As the input can be anything, any error could surface. I was thinking about adding debug logs, but I wasn't sure it would be very useful. But maybe just a warning for user could still be useful somehow 🤔 I'll add something

tests/aws/services/apigateway/conftest.py
bentsku: This was copy-pasted from the previous utility, I actually fixed that issue with https://github.com/localstack/localstack-snapshot/pull/16 but we don't yet have a release of `localstack-snapshot` 😅 soon I'll update these!

tests/aws/services/apigateway/test_apigateway_api.py
bentsku: Ah, good catch! I'll update them, I think I had the case once that the snapshots were not exactly what I was expecting from AWS, so those made sure that we got what we wanted. Will update them 👍

PR #13031, opened 2025-08-19
 Enable B017 ruff linting rule

giograno:
## Motivation
After https://github.com/localstack/localstack/pull/12953, we started to enable all the ignored `ruff` rules.
With this PR, we re-enable [B017](https://docs.astral.sh/ruff/rules/assert-raises-exception/) and fix the code accordingly.

## Changes
- Enable B017;
- Format the code.

### DIFF
--- a/localstack-core/localstack/services/cloudformation/v2/provider.py
+++ b/localstack-core/localstack/services/cloudformation/v2/provider.py
@@ -369,9 +369,8 @@ class CloudformationProviderV2(CloudformationProvider):
     def create_change_set(
         self, context: RequestContext, request: CreateChangeSetInput
     ) -> CreateChangeSetOutput:
-        try:
-            stack_name = request["StackName"]
-        except KeyError:
+        stack_name = request.get("StackName")
+        if not stack_name:
             # TODO: proper exception
             raise ValidationError("StackName must be specified")
         try:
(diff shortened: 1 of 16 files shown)


tests/aws/services/cloudformation/api/test_changesets.py
```diff
@@ -377,7 +377,7 @@ def test_create_change_set_missing_stackname(aws_client):
     template_path = os.path.join(
         os.path.dirname(__file__), "../../../templates/sns_topic_simple.yaml"
     )
-    with pytest.raises(ClientError):
+    with pytest.raises(ParamValidationError):
```
bentsku: interesting find! This is a client-side validation, so probably not really worth to catch/validate, or we'd need to disable the validation with something like `cfn_client = aws_client_factory(config=Config(parameter_validation=False)).cloudformation` 

Nice value from the linter 🚀

alexrashed: Wow, great catch and a great example why this rule really makes sense! @giograno would you mind either fixing the test or adding a TODO here?

giograno: I disabled `boto` parameter validation and raised the right exception in the provider (V2 already implemented it). I agree that there isn't much value in catching the client-side validation error.

localstack-core/localstack/services/cloudformation/v2/provider.py
alexrashed, line 1: **praise:** Nice! You didn't even fix the test, you also fixed an actual bug in the providers! 💯

PR #13033, opened 2025-08-19
CFNV2: Implement Fn::Length for language extensions

simonrw:
## Motivation

We need to support the language extensions transform to match parity with the V1 provider. We start by implementing the `Fn::Length` transform to demonstrate general integration with the v1 implementation. We will implement `Fn::ForEach` in a follow up as this is a more complex change.

## Changes

* Separate local and global transforms in the `ChangeSetModelTransformer.transform` method
* **TEMPORARILY** repurpose `resolve_refs_recursively` to enable the feature, with the knowledge that the requirements for transforms are much lower, and we can potentially re-purpose the `ChangeSetVisitor` or similar to re-write this usage after the new provider is released
    * This allows us to repurpose the v1 implementation of the language extensions macro which we know is in parity with AWS and has been validated by customers
* Unskip the `Fn::Length` test

### DIFF
--- a/localstack-core/localstack/services/cloudformation/engine/transformers.py
+++ b/localstack-core/localstack/services/cloudformation/engine/transformers.py
@@ -14,6 +14,7 @@ from samtranslator.translator.transform import transform as transform_sam
 
 from localstack.aws.api import CommonServiceException
 from localstack.aws.connect import connect_to
+from localstack.services.cloudformation.engine.parameters import StackParameter
 from localstack.services.cloudformation.engine.policy_loader import create_policy_loader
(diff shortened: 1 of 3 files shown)


pinzon (approved):
:+1:

bentsku:
Just a comment: this looks like it was not squashed merged but rebased merged:
Screenshot 2025-08-21 at 00 35 22

simonrw:
Ah yes thanks @bentsku I guess GH considers the setting global? I had rebase merged some other PRs recently

bentsku:
Yeah, I guess it remembers the last value used, not super clear 😅

PR #13037, opened 2025-08-20
upgrade to Python 3.13 and Debian Trixie

dfangl:
## Motivation
This PR updates the base image to debian trixie, and the python version to python 3.13.

This has several advantages:

1. More modern python version
2. Less CVEs detected due to up-to-date dependencies (not all fixes get backported)
3. Generally more up-to-date software

However, this might also lead to certain breakages:

1. Newer debian version will deprecate certain commands, like `apt-key`. For more information, see here: https://www.debian.org/releases/trixie/release-notes/issues.en.html#obsolescence-and-deprecation

2. Removal of certain, already deprecated, python functionality. For more information on this, see here:
Python 3.12: https://docs.python.org/3.12/whatsnew/3.12.html#removed
Python 3.13: https://docs.python.org/3.13/whatsnew/3.13.html#removed-modules-and-apis

## Changes
* Upgrade to python 3.13 and debian trixie
* Init script calls will now have the script name (correctly) as first argument
* Increased unit test timeouts


alexrashed:
This run is already looking pretty good, I triggered a full run to see if there's anything there are any issues in tests that are currently skipped (+ARM): https://github.com/localstack/localstack/actions/runs/17149129206

alexrashed (approved):
Thanks a lot for the great team work on upgrading Python and Debian! This will really help us to keep up to date with the language and the OS, allowing us to use all those cool new language features, and get rid of a lot of CVEs in the image! 💯

alexrashed:
FYI: [Here's a full run (including ARM)](https://github.com/localstack/localstack/actions/runs/17824391565) just to make sure we don't break something unexpected.

PR #13038, opened 2025-08-20
Raise DockerNotAvailable exception if port checking is impossible

joe4dev:
## Motivation

The port checking utility retries upon not being able to find a bindable because we always return `False` upon exception. If Docker (or container client) is not available, retries (without backoff) won't help (unless it's a temporary issue; e.g., Docker starting).

## Changes

Raise `DockerNotAvailable` exception for client code to better distinguish error cases.
⚠️ Breaking change: The current behavior is to return `False`

## Testing

Stop Docker and try feature that uses `container_ports_can_be_bound`

This change might affect the localstack-pro and K8 pipelines, which must succeed before proceeding.
Both pipelines are 🟢🟢

### DIFF
--- a/localstack-core/localstack/utils/docker_utils.py
+++ b/localstack-core/localstack/utils/docker_utils.py
@@ -9,6 +9,7 @@ from localstack.constants import DEFAULT_VOLUME_DIR, DOCKER_IMAGE_NAME
 from localstack.utils.collections import ensure_list
 from localstack.utils.container_utils.container_client import (
     ContainerClient,
+    DockerNotAvailable,
     PortMappings,
     VolumeInfo,
 )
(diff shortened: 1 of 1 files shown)


localstack-core/localstack/utils/docker_utils.py
```diff
@@ -153,10 +154,14 @@ def container_ports_can_be_bound(
             ports=port_mappings,
             remove=True,
         )
+    except DockerNotAvailable as e:
+        LOG.warning("Cannot perform port check because Docker is not available.")
```
joe4dev, line 158: Not sure what's the best logging strategy if the exception is caught upstream as well 🤔

localstack-core/localstack/utils/docker_utils.py
```diff
@@ -153,10 +154,14 @@ def container_ports_can_be_bound(
             ports=port_mappings,
             remove=True,
         )
+    except DockerNotAvailable as e:
```
joe4dev, line 157: Moving forward, we might want to consider generalizing this exception for other container platforms such as K8

dfangl (approved):
LGTM! All tests pass, and it will make it easier for users to spot obvious errors!

PR #13040, opened 2025-08-20
CFNV2: correct state modelling of failed stacks

simonrw:
## Motivation

While implementing the fix for `test_deletion_of_failed_nested_stack` I found that our modelling of stack and change set state needed some work, particularly around deletion. 

> [!NOTE]
> This PR is part of a stack building in #13034

> [!IMPORTANT]
> This is the final PR to unskip all CloudFormation tests for the new engine in this repo 🎉

### Service changes

* Allow the execution to contain a `failure_message`. If this value is present it means the deployment was a failure. Otherwise it was successful
* Create a specific exception for triggering rollback
    * this is not specifically handled at the moment, but it may come in handy in the future
    * It is used to terminate the deployment process so no further resources are deployed after failure
* If the stack fails then model the rollback process
* Only assign the physical resource id from resources that succeed to deploy
* Signal that an operation may be "part of a replacement"
    * when performing replacements, two actual operations occur:
        * creation of new resource
        * deletion of old resource
    * before this change, after an update the _deleted_ state (which is only partial, with no physical resource id in particular) was being stored on the stack causing further bugs
    * now with this change we only update the stack state if the deletion is _not_ part of a replacement, so we don't clobber the new resources' state
* Previously we were not setting failed resources' state on the stack which caused failures to look up the resource properties in future operations. This is now fixed
* Model the resource status in the preproc (and associated types)
* In the transformer, override `visit_node_resource` to skip failures to visit and just move on to the properties to coninue traversing the graph
* Better handling of state propagation from `Executor` -> `ChangeSet`/`Stack` after `execute_change_set`
* Correctly set stack status after `create_stack` on failure
* Increase API parity in minor ways

### Test changes

* 🎉Remove the `skip_if_v2_provider` decorator as we don't need it any more 🎉
* Validate `test_create_and_then_update_refreshes_template_metadata` test
    * the previous version was broken
    * the v1 engine does not support this new version of the test 😂
* Support `test_deletion_of_failed_nested_stack`
* Improve the `test_sns_topic_fifo_without_suffix_fails` test to move the classification from `needs_fixing` to `validated`

## Follow ups

* Differentiate between failures to update and failure to create
* Unify the post deploy actions from `execute_change_set` with those of `create_stack` and `update_stack`

### DIFF
--- a/tests/aws/services/cloudformation/api/test_changesets.py
+++ b/tests/aws/services/cloudformation/api/test_changesets.py
@@ -633,7 +633,7 @@ def test_create_and_then_remove_non_supported_resource_change_set(deploy_cfn_tem
     )
 
 
-@skip_if_v1_provider("Requires the v1 engine")
+@skip_if_v1_provider("Unsupported in v1 engine")
 @markers.aws.validated
 def test_create_and_then_update_refreshes_template_metadata(
     aws_client,


localstack-core/localstack/services/cloudformation/engine/v2/change_set_model_executor.py
```diff
@@ -83,10 +94,23 @@ def __init__(self, change_set: ChangeSet):
...
         if self._deferred_actions:
-            self._change_set.stack.set_stack_status(StackStatus.UPDATE_COMPLETE_CLEANUP_IN_PROGRESS)
+            if failure_message:
+                # TODO: differentiate between update and create
+                self._change_set.stack.set_stack_status(StackStatus.ROLLBACK_IN_PROGRESS)
+            else:
+                # TODO: correct status
+                self._change_set.stack.set_stack_status(
+                    StackStatus.UPDATE_COMPLETE_CLEANUP_IN_PROGRESS
+                )
```
pinzon, line 113: Finally rollback support :tada:

tests/aws/services/cloudformation/api/test_changesets.py
```diff
@@ -633,79 +633,49 @@ def test_create_and_then_remove_non_supported_resource_change_set(deploy_cfn_tem
     )
 
 
-@skip_if_v2_provider("Test", reason="test needs fixing")
-@markers.aws.needs_fixing
+@skip_if_v1_provider("Requires the v1 engine")
```
pinzon:
```suggestion
@skip_if_v1_provider("Unsupported in v1 engine")
```

tests/aws/services/cloudformation/conftest.py
```diff
@@ -12,14 +12,6 @@
 SKIP_TYPE_RE = re.compile(r"^CFNV2\((?P<reason>[^\)]+)\)")
 
 
-def skip_if_v2_provider(*types: str, reason: str = ""):
```
pinzon, line 15: :tada:

localstack-core/localstack/services/cloudformation/v2/provider.py
```diff
@@ -582,20 +583,15 @@ def _run(*args):
...
-                    e,
+                    result.failure_message,
                     exc_info=LOG.isEnabledFor(logging.DEBUG) and config.CFN_VERBOSE_ERRORS,
                 )
-                new_stack_status = StackStatus.UPDATE_FAILED
-                if change_set.change_set_type == ChangeSetType.CREATE:
-                    new_stack_status = StackStatus.CREATE_FAILED
-
-                change_set.stack.set_stack_status(new_stack_status)
+                # stack status is taken care of in the executor
```
pinzon, line 592: **Important:** We should be more constant on which entity is in charge of updating the stack status. As it is now, sometimes it's the provider, other times it's the executor. This could cause confusion when working in the codebase.

simonrw: You are absolutely right I'm glad you picked up on that. A good cleanup opportunity

PR #13042, opened 2025-08-21
Fix cross-arch Docker build

joe4dev:
## Motivation

The `PLATFORM` argument doesn't work for cross-arch builds (e.g., on ARM macs) because without the `--platform` argument, Docker buildkit simply picks the native arguments. In the current implementation, `make PLATFORM=amd64 docker-build` results in arm64 image being built on my M1 MacBook.

## Changes

* Make `PLATFORM` configurable in the `Makefile`
* Use explicit `linux/$PLATFORM` flag for the Docker build

## Testing

`make PLATFORM=amd64 docker-build`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ PYTEST_LOGLEVEL ?= warning
 
 uname_m := $(shell uname -m)
 ifeq ($(uname_m),x86_64)
-platform = amd64
+PLATFORM ?= amd64
 else
-platform = arm64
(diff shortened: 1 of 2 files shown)


alexrashed (approved):
Thanks a lot for jumping on this! As far as I can see, this only has an effect on local cross-arch builds (since in our CI we use dedicated runners to avoid cross-compilation). The changes are looking good! 💯

PR #13060, opened 2025-08-26
Use file URI scheme for hotreload presigned URL

joe4dev:
## Motivation

The presigned URL field is currently used to return the host path for hotreloaded Lambda functions. We currently return an arbitrary string: `Code location: ` followed by the host path.

Example:
* Before: `Code location: /Users/joe/Downloads/my-function`
* After: `file:///Users/joe/Downloads/my-function`

## Changes

Adopt the `file://` URI scheme

* 👍 easier to parse
* 👍 returning a valid URI seems better than an arbitrary string (still no valid URL, but file URI is clickable in some terminals)
* 👎  The meta-information "code location" gets lost


gregfurman (approved):
LGTM

PR #12957, opened 2025-08-05
increase sqs message size limit to 1MiB

SethPyle376:
## Motivation
This increases the SQS message and batch size limit to reflect the change AWS recently made. https://aws.amazon.com/about-aws/whats-new/2025/08/amazon-sqs-max-payload-size-1mib/

fixes #12956 

## Changes
Increases the size limit for SQS messages.

PR #13062, opened 2025-08-27
Fix SQS tests failing due to missing snapshot update after #12957

dfangl:
## Motivation
With #12957, the SQS maximum message size was changed to 1 MiB.
However, some snapshots were missed due to the test selection not taking into account cloudformation tests, which fails our pipelines.

## Changes
* Update snapshots for failing SQS cloudformation tests


baermat (approved):
Since I was part of the initial sqs PR, I took the liberty of reviewing this :) Thanks for tackling this, LGTM!

